PR #554, opened 2022-02-07
Update CONTRIBUTING.md

abidlabs:


CONTRIBUTING.md
```diff
@@ -5,18 +5,25 @@ Prequisites:
 * Python 3.7+
 * Node 16.0+ (optional for backend-only changes, but needed for any frontend changes)
 
-More than 30 awesome developers have contributed to the `gradio` library, and we'd be thrilled if you would like be the next `gradio` contributor! You can start by forking or cloning the
-repo (https://github.com/gradio-app/gradio.git) and creating your own branch to work from.
+More than 30 awesome developers have contributed to the `gradio` library, and we'd be thrilled if you would like be the next `gradio` contributor! You can start by forking or cloning the repo (https://github.com/gradio-app/gradio.git) and creating your own branch to work from.
 
-### To install the local version of Gradio
+### Install Gradio locally from the `master` branch
```
FarukOzderim, line 10: Is there a reason for moving this part to here?

CONTRIBUTING.md
```diff
@@ -5,18 +5,25 @@ Prequisites:
...
 
-### To install the local development version of Gradio
+* Build the front end
+
+```
+bash scripts/build_frontend.sh
+```
+
+
+### To install development and testing requirements
```
FarukOzderim, line 26: Might drop "To" here as well

FarukOzderim:
LGTM other than the mentioned points.

CONTRIBUTING.md
abidlabs: Yes, if installed in a fresh environment, Gradio will not run when installed from `master` if the frontend is not built. You will be able to launch an interface, but the GUI will be blank unless this step is performed.

CONTRIBUTING.md
abidlabs: Ah thanks

PR #555, opened 2022-02-07
Fixing guides url 

aliabd:
# fixing the guides url so gradio.app/guides/ will work 

Currently only gradio.app/guides.html works which is inconsistent with /docs/ /getting_started/ and all the other links on the page. We will still also support gradio.app/guides.html


abidlabs:
LGTM

PR #564, opened 2022-02-08
Chatbot hotfix

aliabid94:
Hotfix involving css parameter and output state, which were causing issues


frontend/public/index.html
```diff
@@ -8,6 +8,7 @@
   <link rel='stylesheet' href='build/themes.css'>
 
   <link rel="stylesheet" href="build/global.css">
+  <style></style>
```
abidlabs, line 11: Why?

pngwn (approved):

abidlabs:
Looks good to me

frontend/public/index.html
aliabid94: The custom css will be dumped into this element

abidlabs (approved):

frontend/public/index.html
abidlabs: Gotcha

aliabid94: actually might be better to create a new element dunamically, you're right

PR #565, opened 2022-02-08
Guides fixes

abidlabs:
# Description

- Added a contributing document to make it easier for others to contribute guides
- Some minor fixes to various guides


aliabid94:
its confusing to call the file readme.md because there is already a readme_template.md in the same folder that generates the main readme file. Perhaps call it contributing.md instead, that's a standard name

abidlabs:
> its confusing to call the file readme.md because there is already a readme_template.md in the same folder that generates the main readme file. Perhaps call it contributing.md instead, that's a standard name

But `readme.md` gets displayed automatically when you visit the `guides` folder on GitHub (https://github.com/gradio-app/gradio/tree/master/guides). Does `contributing.md`?

I think it would make more sense to move `render_readme.md` somewhere else, since it's not really a Guide.

guides/using_flagging.md
```diff
@@ -3,31 +3,35 @@
...
 
-* `allow_flagging`: 
-    * This parameter can be set to either `"manual"`, `"auto"`, or `"never"`.                 
+* `allow_flagging`: this parameter can be set to either `"manual"`, `"auto"`, or `"never"`.                 
     * `manual`: users will see a button to flag, and events are only flagged when it's clicked.
     * `auto`: users will not see a button to flag, but every event will be flagged automatically. 
     * `never`: users will not see a button to flag, and no event will be flagged. 
-* `flagging_options`:
-    * This parameter takes a list of strings.
+* `flagging_options`: this *optional* parameter takes a list of strings.
```
aliabd, line 24: what does optional mean here? all of 4 of these parameters are optional

abidlabs: As in, it can be set to None. We should actually add that if it is set to None, then no options will be provided, and flagging will be more of a binary thing. Let me update it.

aliabd: Hmm, not sure that's relevant, since its None by default. Why would a user want to set it to None?

abidlabs: We still need to mention that it can be either `None` or a list of strings. Otherwise, the specification for the parameter is incomplete. Take a look at how I've written it now, I think it's pretty clear?

aliabd:
Pretty sure it can only be called readme to display the way we want. Also, I think we readme_template shouldn't be in this directory at all. It's not a guide, so it should be pulled from somewhere else.

abidlabs:
Reorganized readme files as we discussed. @aliabd if it looks good, pls go ahead and merge

abidlabs:
Thanks that's correct

website/homepage/render_html.py
```diff
@@ -43,7 +43,7 @@ def render_index():
 
 guides = []
 for guide in sorted(os.listdir(GRADIO_GUIDES_DIR)):
-    if "readme" in guide.lower():
+    if guide.lower() in ["getting_started.md", "readme.md"]:
```
aliabd, line 46: @abidlabs why did you add `getting_started.md` here? was that intentional? It stops `/getting_started` from getting rendered. Note it already gets skipped in the main guides page if that's what you meant

abidlabs: Ah yes it looks like that was a mistake on my part

PR #571, opened 2022-02-09
Update requirements.in

abidlabs:
To include `black>=22.1.0`. Because the `test/requirements.txt` are generated after these requirements are installed.


FarukOzderim:
Going to update this in another PR

PR #589, opened 2022-02-09
use `ui` for frontend

pngwn:
This PR replace the existing `frontend` directory with the `ui` directory.

- All recent changes to `frontend` synced/ ported to `ui` (including themes)
- `ui/packages/app` builds its output to the template dir (`gradio/templates/frontend`)
- CI updated to build/ use the `ui` dir during the build instead of frontend.
- server updated to serve the `assets` path correctly (as the new build uses an `assets` dir not `build`)

### DIFF
--- a/scripts/build_frontend.sh
+++ b/scripts/build_frontend.sh
@@ -5,8 +5,8 @@ if [ -z "$(ls | grep CONTRIBUTING.md)" ]; then
 else
   echo "Building the frontend"
   cd frontend
-  npm install
-  npm run build
+  pnpm i --frozen-lockfile
+  pnpm build
 fi
 
 
--- a/scripts/build_frontend.sh
+++ b/scripts/build_frontend.sh
@@ -4,7 +4,7 @@ if [ -z "$(ls | grep CONTRIBUTING.md)" ]; then
   exit -1
 else
   echo "Building the frontend"
-  cd frontend
+  cd ui
   pnpm i --frozen-lockfile
   pnpm build
 fi
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,9 +31,10 @@ jobs:
       - run:
           name: Build frontend
           command: |
-            cd frontend
-            npm install
-            npm run build
+            npm i pnpm -g
+            cd ui
+            pnpm i --frozen-lockfile
+            pnpm build
       - run:
           command: |
             mkdir screenshots
--- a/gradio/routes.py
+++ b/gradio/routes.py
@@ -25,10 +25,8 @@ from gradio import encryptor, queueing, utils
 from gradio.process_examples import load_from_cache, process_example
 
 STATIC_TEMPLATE_LIB = pkg_resources.resource_filename("gradio", "templates/")
-STATIC_PATH_LIB = pkg_resources.resource_filename(
-    "gradio", "templates/frontend/static")
-BUILD_PATH_LIB = pkg_resources.resource_filename(
-    "gradio", "templates/frontend/assets")
+STATIC_PATH_LIB = pkg_resources.resource_filename("gradio", "templates/frontend/static")
+BUILD_PATH_LIB = pkg_resources.resource_filename("gradio", "templates/frontend/assets")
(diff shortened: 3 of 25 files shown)


abidlabs:
Will give this a spin now!

aliabid94:
Love it. Can we remove the frontend folder now?

scripts/build_frontend.sh
```diff
@@ -5,8 +5,8 @@ if [ -z "$(ls | grep CONTRIBUTING.md)" ]; then
 else
   echo "Building the frontend"
   cd frontend
```
aliabid94, line 7: should this be "cd ui"?

aliabid94:
Pushed some updates to the scripts folder to correspond to new ui folder. When I built for testing, a package-lock.json file was created that wasn't in git. Do we want to gitignore this file?

scripts/build_frontend.sh
pngwn: Ah, yeah it should. I didn't actually know what this script was for.

aliabid94:
For some reason, after I made my small changes (removing console.logs, changing scripts), some of the tailwind css classes are not being compiled to index.css. This is why 
image
is failing, also happens locally. What could be causing this? @pngwn ?

pngwn:
This seems very strange. Your changes shouldn't have impacted this at all. I'll take a look as soon as I can.

abidlabs:
In addition to what @aliabid94 pointed out, I noticed a couple of other things while testing out all of the different components on this branch (these issues may not specifically be a result of this migration, so feel free to move to a different issue if that's the case)

1. The `TimeSeries` component has two issues:
* If the CSV columns do not match what is specified, the time series component becomes blank without any sort of helpful error message. Try running the following code with the attached [timeseries.csv](https://github.com/gradio-app/gradio/files/8038426/timeseries.csv)

```python
import gradio as gr

gr.Interface(
    lambda x:x,
    gr.inputs.Timeseries(x="time", y=["wind1", "temperature", "precipitation"], optional=True),
    gr.outputs.Timeseries(x="time", y=["wind1", "temperature", "precipitation"], label="Timeseries"),
).launch()
```

* Even when the time series component is correctly specified, pressing the CLEAR button causes this weirdness where another blank time series component appears under the first:

image

Code (use the same csv above):
```python
import gradio as gr

iface = gr.Interface(
    lambda x:x,
    gr.inputs.Timeseries(x="time", y=["wind", "temperature", "precipitation"], optional=True),
    gr.outputs.Timeseries(x="time", y=["wind", "temperature", "precipitation"], label="Timeseries"),
).launch()
```

2. The `Image` component in the cropper mode does not clear properly:

image

Code:
```python
import gradio as gr

iface = gr.Interface(
    lambda x:x,
    gr.inputs.Image(label="Image w/ Cropper", tool="select", optional=True),
    "image",
).launch()
```

3. The length of the bars in the output `Label` component no longer seem to correspond to the percentages:

image

```python
import requests
import tensorflow as tf

import gradio as gr

inception_net = tf.keras.applications.MobileNetV2()  # load the model

# Download human-readable labels for ImageNet.
response = requests.get("https://git.io/JJkYN")
labels = response.text.split("\n")


def classify_image(inp):
    inp = inp.reshape((-1, 224, 224, 3))
    inp = tf.keras.applications.mobilenet_v2.preprocess_input(inp)
    prediction = inception_net.predict(inp).flatten()
    return {labels[i]: float(prediction[i]) for i in range(1000)}


image = gr.inputs.Image(shape=(224, 224))
label = gr.outputs.Label(num_top_classes=3)

gr.Interface(
    fn=classify_image, inputs=image, outputs=label).launch()
```

4. I got a `"Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'toFixed')"` when running a modified version of the kitchen sink demo, see below (had to save the python file as a .txt for GitHub to let me attach it) 

[modified-kitchen-sink.txt](https://github.com/gradio-app/gradio/files/8038530/modified-kitchen-sink.txt)


I'll continue testing all of the components and let you know if I find anything else!

pngwn:
#596

PR #590, opened 2022-02-10
Gradio Blocks [WIP]

aliabid94:
Gradio Blocks Branch

dashboard_demo

See accompanying issue: #690


abidlabs:
Can you add the Python code needed to generate this `Page`?

aliabid94:
No backend yet, this is from a hardcoded config in the frontend. Will add a basic backend today and add a demo to the demo folder.

pngwn:
Nice work on this. Will leave a small list of concerns / things we need to think about here. We can break them into issues later.

- I don't think Blocks + App should be separate. They can just be the same thing. The backend can generate a similar shaped config for both. We will just need to apply some styling tweaks to the default in/out interface so it looks like what we have today. Or we can express that interface in terms of some Blocks template.
- I think we should treat all component exactly the same with no distinction between layout style and page style components. The JSON config should be able to represent that (`Node { children:  Node[] }` ). Then we can use a recursive dynamic component to render the tree without need special handling for any components.
- We need to think carefully about how we are storing state and the different kinds of state that we can store. There are times where we need to process state into something more useful for the component and keep hold of that for performance reasons but send the original state to the API. We need a clearer mechanism for this.
- We also need to think about 'events' and how that is going to work.
- We need to ensure we are decoupling standalone component from our business logic. Tabs, Rows, etc should not need to know anything about how gradio works or how the components are being composed together.

gradio/blocks.py
```diff
@@ -0,0 +1,124 @@
...
+        if context.block is not None:
+            context.block.children.append(self)
+        if context.root_block is not None:
+            context.root_block.blocks.append(self)
+        self._id = context.id
+        self.events = []
+        context.id += 1
+
+    def click(self, fn, inputs, outputs):
+        if not isinstance(inputs, list):
```
FarukOzderim, line 17: we might drop support for singular elements to remove duality

abidlabs:
Some small refactoring/renaming suggestions to make using blocks easier. 

1. Can we somehow avoid having to write `gr.` in front of every component? For example, by having a `from gradio.blocks import *` at the top allows us to import all of the components? I've seen this e.g. in Keras tutorials where you import all of the layers. at once. It would more readable:

```python
with Blocks() as blocks:
	Markdown("""
	# Detect Disease From Scan
	With this model you can lorem ipsum
	- ipsum 1
	- ipsum 2
	""")
	disease = inputs.CheckboxGroup(["Covid", "Malaria", "Lung Cancer"], 
                          label="Disease to Scan For")

	with Tab("X-ray"):	
		with Row():
			xray_scan = inputs.Image()
			xray_results = outputs.JSON()
		xray_run = Button("Run")
		xray_run.click(xray_model, inputs=xray_scan, outputs=xray_results)

	with Tab("CT Scan"):
		with Row():
			ct_scan = inputs.Image()
			ct_results = outputs.JSON()
		ct_run = Button("Run")
		ct_run.click(ct_model, inputs=ct_scan, outputs=ct_results)
		
	overall_probability = outputs.Textbox()

	blocks.launch()
```

2. It should be possible to treat ANY output component like a static component by simply dumping an appropriate value to it. For reference, many Streamlit apps are just people dumping dataframes or graphs and seeing Streamlit render it in a pretty way. Like this:

```python
df = ...

with Blocks() as block:
  outputs.Dataframe(df)
```

I think in order to do this, we would have to refactor the arguments for the `Output` components so that instead of the first argument being label, the first argument could be some sort of default value. So perhaps we should save this for `gradio 3.x`  but I think this will be very convenient.

A non-breaking alternative that is a bit more cumbersome may be:

```python
df = ...

with Blocks() as block:
  outputs.Dataframe(value=df)
```

3. If we implement (2), we could upgrade `Markdown` to a proper `Output` component instead of it being a Static component like it is now.

4. Thinking from the perspective of current Gradio users, I think it would be much easier for them to be able to provide Gradio Interfaces as blocks. What I mean is something like this:

```python
with Blocks() as blocks:
	with Tab("X-ray"):	
		Interface(xray_model, inputs=xray_scan, outputs=xray_results)

	with Tab("CT Scan"):
		Interface(ct_model, inputs=ct_scan, outputs=ct_results)
		
	blocks.launch()
```

Many people just want to be able to have multiple interfaces in one place and this would allow them to do that without having to learn event handling, etc.

I'll turn these feature requests into actual issues, but pasting them here so we can discuss them all easily

FarukOzderim:
@aliabid94 btw did you realize tests started to failed after this [commit](https://github.com/gradio-app/gradio/pull/590/commits/188757c1abd64a69a27ee926e30a890fcd87bc7b)

FarukOzderim (commented):
It seems this review did not appear because it requires submit altogether.
image

abidlabs:
Now I can see your comments @FarukOzderim!

PR #601, opened 2022-02-11
Anchor tags on all headers 

### LINKED ISSUE #592: Add automatic anchor links to the different sections of getting started and guides automatically
Many times when I'm responding to issues, I want to link to a specific section in our getting started page. For example, the section related to "State"

image

We could manually add anchor links, but I'm sure there is a way to automatically turn any header tag into an anchorable link. Sometimes I've seen a little symbol next to the heading that you can click on to get the anchor link for that section, which would be super convenient.

Applies to the getting started as well as the guides.

aliabd:
All headers from the markdown will include an anchor tag (visible when hovering) 

* every header will get an id during the conversion to html
* on the js an anchor tag will be added linking to that id
* there was regex that adds an id for h3 headers, not sure why. I removed it as it's not longer going to do anything. But it used underscores instead of dashes so if these links are used anywhere they will not anchor the page (not a big deal because they will still link to the right url) 

Closes #592

https://user-images.githubusercontent.com/9021060/153637316-e553a977-8048-4c2a-9659-6ec1f82e04f7.mov


aliabid94:
Looks great. Please use tailwind styling.

aliabid94 (approved):
LGTM

Issue #256, opened 2021-08-12
optional input components


AK391:
input components could be made optional with a flag to avoid having all fields being populated to run the app

AK391:
similar to https://github.com/gradio-app/gradio/issues/236 closing

PR #603, opened 2022-02-12
Optional-Input-Clarification

FarukOzderim:
- when the component is optional, label it as optional

Fixes: #256

### DIFF
--- a/gradio/component.py
+++ b/gradio/component.py
@@ -1,6 +1,6 @@
 import os
 import shutil
-from typing import Any, Dict
+from typing import Any, Dict, Optional
 
 from gradio import processing_utils
 
@@ -49,7 +49,7 @@ class Component:
 
     def save_flagged_file(
         self, dir: str, label: str, data: Any, encryption_key: bool
-    ) -> str:
+    ) -> Optional[str]:
         """
         Saved flagged data (e.g. image or audio) as a file and returns filepath
         """
--- a/gradio/inputs.py
+++ b/gradio/inputs.py
@@ -30,11 +30,15 @@ class InputComponent(Component):
     Input Component. All input components subclass this.
     """
 
-    def __init__(self, label: str, requires_permissions: bool = False):
+    def __init__(
+        self, label: str, requires_permissions: bool = False, optional: bool = False
+    ):
         """
         Constructs an input component.
         """
         self.set_interpret_parameters()
+        if optional is True:
+            label = InputComponent.label_as_optional(label)
         super().__init__(label, requires_permissions)
 
     def preprocess(self, x: Any) -> Any:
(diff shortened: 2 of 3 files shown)


abidlabs:
Thanks for adding the screenshot @FarukOzderim!

abidlabs:
I tested it and noticed an issue that would happen if the component was not given a `label`, so I refactored the logic for optional and moved it to the `get_config_file` function

abidlabs:
Specifically the issue was that previously, if an input component doesn't have a label, it is assigned one based on the name of the corresponding parameter in the `fn`.

However now, if an input component is marked as optional, its name would get set to "None (optional)" and it would not get one assigned based on the name of the corresponding parameter in the `fn`. 

Run this code to verify:
```python
import gradio as gr

def identity(x):
    return x

io = gr.Interface(identity, gr.inputs.Number(optional=True), "text")
io.launch()
```

Should be fixed now.

abidlabs (approved):

FarukOzderim:
@abidlabs thanks for the updates.
I think using this instead of the latter would be better, what do you say?

```
from typing Union
List[Union[str, Number, bool]] 
```

```
List[str | Number | bool]
```

abidlabs:
Either way is fine, but I personally find the latter more readable, especially in nested situations.

It was introduced specifically in Python 3.10 to simplify syntax (https://www.python.org/dev/peps/pep-0604/#simplified-syntax) so I've imported it from `__future__` and started using it throughout the codebase

aliabid94:
the " (optional)" panel header tag should be added in the frontend, not be added to the label from the backend. There are other places where the label is used, for example as headers of the CSV file if flagging is used. They will all be affected by the label change which seems unnecessary. Makes more sense to have the frontend add a small optional tag if the component is optional.

abidlabs:
That makes sense. @FarukOzderim is this something you can do in the frontend? Or should someone else take it on?

FarukOzderim:
@abidlabs I would be happy to look into it with some guidance, however I do not know where to edit after looking into the frontend a little bit. 
@aliabid94 can you guide me on slack about this issue, understanding the frontend structure could be nice for me.

abidlabs:
@FarukOzderim it looks like you should be able to make the change here: 

https://github.com/gradio-app/gradio/blob/6d973aa939e0523c18fa233c9cce280c9d980d05/ui/packages/app/src/Interface.svelte#L203

Something along the lines of 
```
<div class="panel-header mb-1.5">{input_component.label}{#if input_component.optional} (optional){/if}</div>
```

FarukOzderim:
Updated the PR, this is how it looks now.

image

FarukOzderim:
This is how it looks when it is not optional as well:

abidlabs:
LGTM!

PR #607, opened 2022-02-14
Dataframe fixes

### LINKED ISSUE #588: Dataframe input loading strange default data (and other issues)
**Describe the bug**

The `dataframe` input loads very specific input data by default (the first column are the names "Frank", "Beatrice", "Simone", then there are two columns of numbers, as shown in the screenshot below)

image

This happens whether or not the `default` parameter is passed in. If this parameter is passed in, it is ignored, which can lead to issues, as can be seen by going to the "Working with DataFrames and Graphs" section in the getting started: https://gradio.app/getting_started/

The correct behavior should be to load the `default` dataframe if it is provided. Otherwise, instantiate with an empty dataframe.

**To Reproduce**

Minimal example:

```python
import gradio as gr

iface = gr.Interface(
    lambda x:x,
    gr.inputs.Dataframe(
        headers=["Name", "Jan Sales", "Feb Sales", "Mar Sales"],
        default=[["Jon", 12, 14, 18], ["Alice", 14, 17, 2], ["Sana", 8, 9.5, 12]],
    ),
    "dataframe",
)
iface.launch()
```

pngwn:
- default data is now respected
- Backspace and Delete can clear cells
- Heading label has been removed

Closes #588

### DIFF
--- a/ui/packages/app/src/components/input/DataFrame/DataFrame.svelte
+++ b/ui/packages/app/src/components/input/DataFrame/DataFrame.svelte
@@ -2,14 +2,15 @@
 	import { tick } from "svelte";
 
 	export let headers: Array<string> = [];
-	export let values: Array<Array<string | number>> = [
-		["Frank", 32, "Male"],
-		["Beatrice", 99, "Female"],
-		["Simone", 999, "Male"]
-	];
+	export let values: Array<Array<string | number>> = [["", "", ""]];
+	export let default_data: Array<Array<string | number>> = [];
+
+	if ($$props.default) values = $$props.default;
 
 	export let setValue: (val: typeof values) => typeof values;
 	export let editable = true;
+
+	$: console.log($$props);
 
 	let id = 0;
 	let editing: boolean | number = false;
--- a/ui/packages/app/src/components/input/DataFrame/DataFrame.svelte
+++ b/ui/packages/app/src/components/input/DataFrame/DataFrame.svelte
@@ -1,7 +1,6 @@
 <script lang="ts">
 	import { tick } from "svelte";
 
-	export let label = "Title";
 	export let headers: Array<string> = [];
 	export let values: Array<Array<string | number>> = [
 		["Frank", 32, "Male"],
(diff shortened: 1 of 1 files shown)


ui/packages/app/src/components/input/DataFrame/DataFrame.svelte
```diff
@@ -1,17 +1,17 @@
...
-	];
+	export let values: Array<Array<string | number>> = [["", "", ""]];
+	export let default_data: Array<Array<string | number>> = [];
+
+	if ($$props.default) values = $$props.default;
 
 	export let setValue: (val: typeof values) => typeof values;
 	export let editable = true;
 
+	$: console.log($$props);
```
aliabid94, line 13: remove?

ui/packages/app/src/components/input/DataFrame/DataFrame.svelte
```diff
@@ -87,6 +87,7 @@
 		id: number
 	) {
 		let is_data;
+		console.log(event.key, data);
```
aliabid94, line 90: remove?

aliabid94 (approved):
LGTM

aliabid94:
Do you know why this only ran the static checks, not the entire circleci build @pngwn @FarukOzderim ?

FarukOzderim:
Maybe since only ui folder has changed :D? I checked other commits only affecting the ui folder but they have circleci tests 😕 
Probably a bug on circleci side, maybe it was down at the time of the commit.

FarukOzderim:
While checking $$ syntax I came across this, is this an important point @pngwn?

`$$props references all props that are passed to a component, including ones that are not declared with export. It is not generally recommended, as it is difficult for Svelte to optimise. But it can be useful in rare cases – for example, when you don't know at compile time what props might be passed to a component.`

pngwn:
@FarukOzderim Yes it is. In general the more specific we are with references, the more svelte can optimise the code. But things like `$$props` are useful when debugging to take a look at all props being passed around.

They have been removed now.

pngwn:
Going to merge this and create a new issues for removing rows/ cols.

PR #617, opened 2022-02-15
fixes PermissionError for non-writable flagging path

kollmats:
# Description

Launching the interface crashes if the flagging directory is read-only. This happens even if flagging is set to never. 

It looks like one of the if conditions was expecting allow_flagging to be a boolean even though it is now a string, so I've fixed that and added a test case for it.

Fixes: # (issue)

# Checklist:


abidlabs:
Thank you @kollmats! Tested locally + LGTM

PR #633, opened 2022-02-16
tweaks + tests for 18n

pngwn:
This PR does the following:

- Moves langs into the `src` of the app, these are source files rather than static assets.
- Convert `i18n.js` to typescript
- Removes the requirement to manually add the lang to the i18n file. We can glob import all of the translation files so adding translations is even easier.
- Added a few tests to ensure languages have the correct code format and that the translations exist.
  - These aren't perfect right now, it only looks at the shape of the country code, we can be stricter here.
  - We can add more 18n tests in general but this is a good starting point.
  
This PR will conflict with some in flight PRs. I'm happy for us to merge those first and I can rebase this afterwards, or we can merge this and I'll take care of the conflicts on those branches.


ui/packages/app/src/lang/README.md
```diff
@@ -7,8 +7,7 @@ To add your language, do the following steps:
 1. Create a new json file in this directory
 2. Name the file after the language code (Here's a list: http://4umi.com/web/html/languagecodes.php)
 3. Please provide clear and complete translations. Take a look at the [`en.json`](https://github.com/gradio-app/gradio/blob/master/ui/packages/app/public/lang/en.json) file for the corresponding English text.
-4. Lastly, the json file will need to be imported and added into the [`i18n.js` file](https://github.com/gradio-app/gradio/blob/master/ui/packages/app/src/i18n.js)
 
 **Example:** take a look at this [pull request which adds support for Urdu](https://github.com/gradio-app/gradio/pull/621/files)
```
abidlabs, line 11: Would remove this for now, as the example PR would no longer be valid

pngwn:
@abidlabs sounds good.

dawoodkhan82 (approved):
Looks great

PR #652, opened 2022-02-17
fix: dark mode

pngwn:
Fixes dark mode.


abidlabs:
Noticed this issue in the kitchen sink: when you scroll down past the initially visible part of the interface, the dark background changes to a white background

https://user-images.githubusercontent.com/1778297/154535233-ce85158f-66af-48a9-9e08-7675fbbdc933.mp4

:

pngwn:
hmm, i thought if fixed this.

abidlabs:
I see this when I run `/demo/kitichen_sink/run.py` changing the theme to "dark"

pngwn:
Okay, I think that should be fixed now.

abidlabs:
LGTM. 

Now the footer is directly under the Interface, which looks much nicer imo, but it broke the selenium tests. Let me update the test images and we should be good to go.

PR #685, opened 2022-02-20
Add Polish translation

niemiaszek:
Translate interface to Polish


abidlabs:
Thank you so much for the contribution @niemiaszek!

PR #689, opened 2022-02-21
Persian Language Support

HalflingWizard:
# Description

As a Persian user of Gradio, I thought it would be nice if it also supported my language. This helps Persian (Farsi) speakers use gradio with more ease. also, it would be my little contribution to this great project.

I simply used JSON files in `ui\packages\app\src\lang` directory as a blueprint and Created `fa.json` according to the guidelines in `README.md`.

### DIFF
--- a/ui/packages/app/src/lang/fa.json
+++ b/ui/packages/app/src/lang/fa.json
@@ -1,0 +1,18 @@
+{
+	"interface": {
+		"submit": "ثبت",
+		"clear": "پاک کردن",
+		"interpret": "تحلیل",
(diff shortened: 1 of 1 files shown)


abidlabs:
Thank you so much for the contribution, @HalflingWizard! We actually had another contributor open a PR for Farsi yesterday: https://github.com/gradio-app/gradio/pull/684

Since that was opened earlier, I merged it in. However, if you feel like anything should be translated differently, feel free to discuss on that PR and we can go from there. متشکرم

Issue #645, opened 2022-02-17
Delay interface re-run with "live=True"


yesdeepakmittal:
Please add an option to delay the re-run of an application when we use `live=True`. 

What do I want? I want to implement `live=True` into **[Using State](https://gradio.app/getting_started/#using-state)** example mentioned into documentation. Currently, if I use `live=True`, my application is re-run after entering every character. I want to give users some time so that they can successfully enter the input and then **Using State** will work.

abidlabs:
Can you give a little more context into what kind of machine learning application you are building?

How often would you prefer it to be rerun? Every X characters?

abidlabs:
Yup. On the Python side, `live` could take an alternative `float` value which would the amount of time after which to call the function again.

pngwn:
I think this would probably be a good default, at least for text inputs, with a sensible value that we can figure out with a little testing. We can also make it configurable via a param as you suggested.

abidlabs:
Hi @yesdeepakmittal, we experimented with adding a debounce, but it caused other issues with live interfaces. I don't think we'll add debounce to the components since we've found that trying to get this to work well varies significantly depending on the model and the rest of the GUI.

@pngwn feel free to reopen if you reached a different conclusion.

PR #692, opened 2022-02-21
Refresh rate

abidlabs:
# Description

For demos in which the machine learning model has a large inference time, `live=True` causes issues as it reruns the model every time the input changes (e.g. on every keystroke for a text input model). A solution to this is add a parameter that delays the re-running of the model based on a user-specified `refresh_rate`.

This PR adds this functionality. Currently, the backend is implemented. The `Interface` class now takes a `refresh_rate` parameter, which is saved to the config file. By default, this is set to 0, which means that the output is refreshed every single time the input is changed at all. This is the current behavior, so this does not break any interfaces which currently have `live=True`. 

If this is provided as some value > 0, then the Interface should start a timer and make sure that at least `refresh_rate` seconds have passed before rerunning the input. @pngwn @dawoodkhan82 @aliabid94 can one of you take the front end for this?

Fixes: #645

### DIFF
--- a/gradio/interface.py
+++ b/gradio/interface.py
@@ -629,6 +629,7 @@ class Interface:
         favicon_path: Optional[str] = None,
         ssl_keyfile: Optional[str] = None,
         ssl_certfile: Optional[str] = None,
+        ssl_keyfile_password: Optional[str] = None,
     ) -> Tuple[flask.Flask, str, str]:
         """
         Launches the webserver that serves the UI for the interface.
@@ -653,6 +654,7 @@ class Interface:
         favicon_path (str): If a path to a file (.png, .gif, or .ico) is provided, it will be used as the favicon for the web page.
         ssl_keyfile (str): If a path to a file is provided, will use this as the private key file to create a local server running on https.
         ssl_certfile (str): If a path to a file is provided, will use this as the signed certificate for https. Needs to be provided if ssl_keyfile is provided.
+        ssl_keyfile_password (str): If a password is provided, will use this with the ssl certificate for https.
         Returns:
         app (flask.Flask): Flask app object
         path_to_local_server (str): Locally accessible link
@@ -694,7 +696,8 @@ class Interface:
             cache_interface_examples(self)
 
         server_port, path_to_local_server, app, server = networking.start_server(
-            self, server_name, server_port, ssl_keyfile, ssl_certfile
+            self, server_name, server_port, ssl_keyfile, ssl_certfile, 
+            ssl_keyfile_password
         )
 
         self.local_url = path_to_local_server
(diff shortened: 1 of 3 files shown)


FarukOzderim:
@abidlabs @pngwn instead of having a refresh_rate wouldn't it be better to have a timer start after the typing has stopped? And I think it would be better if we do not provide the duration to the user.

abidlabs:
Keep in mind that we want to support all the different kinds of interfaces (textbox, sketchpad, microphone, webcam, etc.) where the user might want a live interface, yet model inference might take some time. The `refresh_rate` can provide a generic solution to this. For example, in the case of a livestreaming microphone, the model could rerun and refresh the output every `refresh_rate` seconds.

> wouldn't it be better to have a timer start after the typing has stopped

That's interesting. Maybe that would make sense for a textbox interface, but that would probably would not be consistent with e.g. how we implement a livestreaming microphone where we would want the output to refreshed every X seconds regardless of user input. I think for the sake consistency across components, we should go with refreshing every X seconds if the input has changed.  

>  And I think it would be better if we do not provide the duration to the user.

You mean do not provide this parameter to the user? We don't know ahead of time how long a model inference is going to take, so in my opinion, it would make sense for the user who has that knowledge to be able set this value. For example, for this [real-time sketching demo](https://huggingface.co/spaces/abidlabs/Draw), it makes sense to refresh after every stroke, because the model runs so quickly. Yet for a slower model, it probably would make more sense to add a refresh_rate >0.

aliabid94:
I don't understand the need for this. If live=True, the function will rerun when
a) the user input has changed AND
b) the previous submission is complete.
So if for example, there's a text based model that takes 20 seconds to run, it will run when the user types the first letter, and then if the user keeps typing, it will run 20 seconds later when the previous run is complete. Why isn't that enough?

abidlabs:
Oh really? I thought there was a request sent every time the input changes in the front end. If it’s how you are describing @aliabid94 then I agree this is not needed

FarukOzderim:
> I don't understand the need for this. If live=True, the function will rerun when a) the user input has changed AND b) the previous submission is complete. So if for example, there's a text based model that takes 20 seconds to run, it will run when the user types the first letter, and then if the user keeps typing, it will run 20 seconds later when the previous run is complete. Why isn't that enough?

@aliabid94 I think it would not be enough due to the waiting 20 sec. As a user I would like it to run when I stop typing or finished my paragraph. Waiting 20 sec after I typed my first character does not seem like a nice UX to me. I think users would prefer smt like this.
> @abidlabs @pngwn instead of having a refresh_rate wouldn't it be better to have a timer start after the typing has stopped? And I think it would be better if we do not provide the duration to the user.

abidlabs:
Ah interesting. @FarukOzderim what you're proposing (and actually what the original author asked for as well) is to start the model inference X number of seconds *after* the input has stopped changing. The goal here is to provide a better user experience, not so much as to prevent a bunch of inference requests happening in parallel like I or @aliabid94 assumed.

I suppose this is reasonable and can be useful in some cases. Not a high priority, but we could add support for it.

FarukOzderim:
ccing @pngwn as well on this topic, since I remember him talking about this in a meeting about blocks.

aliabid94:
I feel like like in this case, they should just have a submit button, e.g. live=False. Seems like a very small use case where you'd want to wait a specific amount of time and they can just press Submit in that case.

PR #710, opened 2022-02-22
Switched from `json` to `orjson`

### LINKED ISSUE #371: BUG: DataFrame output cannot handle null values
#### Result:
pandas DataFrame doesn't display if more than 10 cells are to be shown initially

#### Expectation: 
The DataFrame will be displayed

#### Versions:
gradio 2.2.15
Python 3.9.6

#### Reproducible Example

```
import gradio as gr
import pandas as pd
import plotly.express as px

def show_pens(alpha):
    df_pens = pd.read_csv('https://raw.githubusercontent.com/mwaskom/seaborn-data/master/penguins.csv')
    print(df_pens)
    return df_pens.iloc[:5, :3]

iface = gr.Interface(
    fn=show_pens, 
    inputs=['text'],
    outputs=[gr.outputs.Dataframe()],
    description="Table of Palmer Penguins"
)

iface.launch()
```

### Screenshot of failing output
Screen Shot 2021-11-16 at 7 59 47 AM

### Console error screenshot

Screen Shot 2021-11-16 at 8 05 56 AM

### Working example

Change the return statement to  ` return df_pens.iloc[:3, :3]`

### Screenshot of working output

Screen Shot 2021-11-16 at 8 00 24 AM

abidlabs:
Switched FastAPI from using `json` to `orjson` to render the responses.

**Why did we go from a built-in library to a 3rd-party library**? Understandable question...

- [orjson](https://github.com/ijl/orjson) is a well-maintained library that is faster than `json`, which means everything runs a tad bit faster, especially for large responses (e.g. images, big data frames, etc.)
-  More importantly, it correctly serializes things like `numpy` arrays and `pandas` dataframes, which `json` does not always handle correctly. This allows us to close #371 with a general fix

### DIFF
--- a/gradio.egg-info/requires.txt
+++ b/gradio.egg-info/requires.txt
@@ -5,6 +5,7 @@ ffmpy
 markdown-it-py[linkify,plugins]
 matplotlib
 numpy
+orjson
 pandas
 paramiko
 pillow
--- a/gradio/routes.py
+++ b/gradio/routes.py
@@ -9,8 +9,9 @@ import posixpath
 import secrets
 import traceback
 import urllib
-from typing import List, Optional, Type
+from typing import Any, List, Optional, Type
 
+import orjson
 import pkg_resources
 import uvicorn
 from fastapi import Depends, FastAPI, HTTPException, Request, status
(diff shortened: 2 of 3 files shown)


FarukOzderim:
An example benchmark about json libraries, orjson looks good 🦸 

https://github.com/ultrajson/ultrajson#benchmarks

FarukOzderim:
You need to install gradio locally for packages in setup py to appear in requires.txt I presume?

abidlabs:
Whoops just added to `requires.txt`, thanks @FarukOzderim. 

I don't need to add it anywhere else, do I?

FarukOzderim:
> Whoops just added to `requires.txt`, thanks @FarukOzderim.
> 
> I don't need to add it anywhere else, do I?

Yeah no need.

FarukOzderim:
@abidlabs did you see my review at top 😺 ?

Issue #701, opened 2022-02-22
File Uploading Issue


samarth70:
Hi, I am unable to take examples and use them as input. Though if I upload the file manually the app is working fine.

Link to the app- https://huggingface.co/spaces/Sa-m/manifesto

Error-
![image](https://user-images.githubusercontent.com/44370096/155064876-7bdc4ecf-2d4e-4097-91bd-e4e3f861e152.png)

abidlabs:
Thanks @samarth70. This is most likely the same issue as #694, which we are working on resolving soon!

Issue #697, opened 2022-02-21
Multiple Files input not working


arturcuringa:
Using `gr.inputs.File(file_count="multiple", label=None)` for multiple files input results in the interface allowing only one file and not processing it correctly.

**To Reproduce**
1. Set gr.inputs.File(file_count="multiple", label=None)
2. Start the gradio server 
3. Drag and drop multiple files on the file box/Manually Select a file
4. See that the files are not attached

**Device information:**
 - OS: Ubuntu 20.04
 - Browser Firefox, Chrome
 - gradio==2.8.1

**Additional context**
This issue does not occurs in version 2.7.5

abidlabs:
Thanks @arturcuringa. Can confirm I can reproduce the problem. We'll fix it soon!

abidlabs:
Should be fixed now by #714!

Issue #694, opened 2022-02-21
Cannot load example for files.


lukewys:
**Describe the bug**
In current version, gradio.inputs.File accepts input as:

`File accepts the List of JSON objects with filename as 'name' property and base64 data as 'data' property as type List[Dict[name: str, data: str]]`

However, for example files, it somehow does not work, probably because the example files only provide an str filepath.

**To Reproduce**
In colab demos (https://colab.research.google.com/drive/18abcC7R4NtzA-CcsI1QoJ4ZTuvBvccdY) under https://gradio.app/docs/#i_file, the file input UI will crash after clicked on example files:

before:

![image](https://user-images.githubusercontent.com/28220671/155012004-1d72c770-0500-4cfa-9f98-3f93bd349caa.png)

After:

![image](https://user-images.githubusercontent.com/28220671/155012043-f3ef9879-53d7-479d-8a62-84e0d07c5386.png)

A similar problem can be found here: https://huggingface.co/spaces/lukewys/midi-ddsp, as you click on example mid file, the upload frame ("drop file here or upload") crashes.

abidlabs:
Thanks for reporting this issue @lukewys. Can confirm that I can reproduce this with the code below. We'll fix it soon!

```python
from zipfile import ZipFile

import gradio as gr


def zip_two_files(file1, file2):
    with ZipFile("tmp.zip", "w") as zipObj:
        zipObj.write(file1.name, "file1")
        zipObj.write(file2.name, "file2")
    return "tmp.zip"


iface = gr.Interface(
    zip_two_files,
    ["file", "file"],
    "file",
    examples=[
        ["files/titanic.csv", "files/titanic.csv"],
    ],
)

iface.launch()

```

abidlabs:
Should be fixed now by #714!

lukewys:
Oh, I see, I need to wait until 2.9.0 comes out. Huge thanks for fixing the bug

abidlabs:
Feel free to [install Gradio from `master`](https://github.com/gradio-app/gradio/blob/master/CONTRIBUTING.md) if you need it immediately, but we should be releasing a version out on pypi within the next day or so.

lukewys:
Huge thanks!

PR #714, opened 2022-02-23
fix file upload issues

aliabid94:
Fixes: #701, #697 , #694 

Fix upload file examples, multiple file upload. 

File example issue fix was that there is no size given for example files.


pngwn (approved):
lgtm!

pngwn:
Frontend test flake was my bad, will fix that later. Not sure about the CircleCI test.

abidlabs:
The backend test was failing because one of HF's models was down and we were testing an integration with it. It's up now, so it looks like we're good to merge, right @pngwn?

Issue #709, opened 2022-02-22
Visual feedback for flagging


abidlabs:
When you click the flagging button, there is no visual feedback that you've clicked flag. Makes it hard to tell if you've successfully completed the action or not. Would be nice to have.

abidlabs:
Same for the interpretation button

Issue #682, opened 2022-02-19
Flag options and screenshots is not working with v2.8.1


kingabzpro:
**Describe the bug**
Flag button is showing but without dropdown option and screenshot is not even showing. 

**To Reproduce**
I have use these options 👇

```python
def main():
	iface = gr.Interface(
	                     gen, 
	                     gr.inputs.Image(shape=(640,480),type='numpy'), 
	                     "image", 
	                     title = title, 
	                     flagging_options=["incorrect", "worst","ambiguous"], 
	                     allow_flagging = "manual",
	                     flagging_callback=hf_writer,
	                     description = description, 
	                     article = article, 
	                     examples = examples, 
	                     theme ="peach",
	                     thumbnail=thumbnail, 
	                     allow_screenshot=True
	                     )

	iface.launch(enable_queue=True)

if __name__ == '__main__':
	main()
```
Here is the project: https://huggingface.co/spaces/kingabzpro/savtadepth where I have enabled both custom flag options and screen shot in `app/savta_app.py` file. 

**Screenshots**
![image](https://user-images.githubusercontent.com/36753484/154812094-ee094e03-1211-49a6-96ab-eced1dd7af0e.png)

**Device information (please complete the following information):**
 - OS: HF Spaces
 - Browser: Edge
 - Gradio version: 2.8.1

abidlabs:
Just FYI we intentionally deprecated the screenshot button because it was quite buggy and hard to support all of the edge cases.

But the flagging options should be supported, it's weird that they are not.

abidlabs:
Should be fixed now @kingabzpro with the latest version on `master`. Feel free to build from source if you need it immediately, but we will be releasing a version out on pypi within the next couple of days.

PR #715, opened 2022-02-23
Restore flagging options

aliabid94:
# Description

Restores functionality to specify option when flagging. Also flashes the flagging button red when flagging as click feedback.

Fixes:  #709, #682


abidlabs:
image

LGTM.

Works for me. Would be good to add an integration test for this later.

PR #716, opened 2022-02-23
handle input number values correctly

### LINKED ISSUE #667: Gradio Number field taking string input as well
By default Gradio takes float value in `gradio.inputs.Number()` and it does not allow us to enter string characters into the field. But if we enter a number and then try to enter the string, it accepts it as well. 

image

---

Please consider this documentation example: 
image

Please tell me how to enforce this condition to allow numbers only?

--- 

Another example
image

pngwn:
Use svelte value bindings instead of listeners as this will handle all number values correctly.

fixes #667


abidlabs:
Cool, tested it and works pretty well. Handles lots of interesting cases:

image

LGTM

PR #717, opened 2022-02-23
Add Code Snippet Part to the Bug Report

FarukOzderim:
Make reproducing easier with bug reports.


pngwn:
Maybe we could do this instead: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms

This way we can guarantee (as much as possible) that the correct information is provided, as these fields can be marked as required.

You can see an example of the issues forms here: https://github.com/sveltejs/svelte/issues/new?assignees=&labels=triage%3A+bug&template=bug_report.yml

FarukOzderim:
Hmm, I am not sure we need it but I like it, it is a long run investment so let me do it!

FarukOzderim:
Done!

abidlabs:
Is there a way I can see what this bug report template looks like? Can you add a screenshot @FarukOzderim?

(meta haha)

FarukOzderim:
Well we cannot do it on this repo :D

I tried it on the fork but there is no issue segment on forks :DD

Another repo [here](https://github.com/FarukOzderim/bug_report/issues/new?assignees=&labels=bug&template=bug_report.yml)

And the screenshot:

image

abidlabs:
Cool. Made some minor tweaks to the wording. 

The only part that I find confusing is the severity dropdown, what is the difference between `critical` and `blocker`. Do we even need this dropdown? I think we can just drop it

```
  - type: dropdown
    id: severity
    attributes:
      label: Severity
      description: Select the severity of this issue
      options:
        - annoyance
        - critical
        - blocker
    validations:
      required: true
```

FarukOzderim:
> Cool. Made some minor tweaks to the wording.
> 
> The only part that I find confusing is the severity dropdown, what is the difference between `critical` and `blocker`. Do we even need self-reported severity? Not sure how useful it will be for us, so I suggest we just drop it
> 
> ```
>   - type: dropdown
>     id: severity
>     attributes:
>       label: Severity
>       description: Select the severity of this issue
>       options:
>         - annoyance
>         - critical
>         - blocker
>     validations:
>       required: true
> ```

Critical would be a critical error,
Blocker would be an error blocking a development.

I think it is nice to have to decrease our effort on prioritizing bugs, but we would not trust it %100 since it is self reported as well.

abidlabs:
LGTM, feel free to merge

PR #718, opened 2022-02-23
Cancel tests with updates

FarukOzderim:
# Description

Cancel the previous commits test when a newer commit is added. Motivated by this [tweet](https://mobile.twitter.com/sebastienlorber/status/1495776331287646209).


abidlabs:
I might be missing something basic... I understand why we are updating `.github/workflows/ui.yml`. But why are we updating `.github/ISSUE_TEMPLATE/bug_report.yml` with the concurrency lines?

FarukOzderim:
> I might be missing something basic... I understand why we are updating `.github/workflows/ui.yml`. But why are we updating `.github/ISSUE_TEMPLATE/bug_report.yml` with the concurrency lines?

It was an error on my part, reverted the commit but forgot the push :D, thanks for catching it!

abidlabs:
Sounds good. LGTM, feel free to merge!

FarukOzderim:
Whoops I was waiting an approval from @pngwn forgot that, could you look into the changes even if it's merged? :D

pngwn:
Not sure this is really as clear cut as the tweet suggests, you lose the ability to map a commit back to a CI failure with this approach. Our CI takes seconds rather than minutes or hours and this change will have very little impact. I appreciate the motivation to save resources but depending what that means I am unconvinced that there would be significant impact:
 - If resources means $$$: CI is free for public repos
 - If resources means energy: this will have next to no impact, I would be very surprised if even long running CI has significant energy consumption other than in the most exceptional of cases. I don't want to get into this too deeply, but... information on the energy consumption of datacentres is hard to come by but estimates put it at single digit percentages of global energy consumption, CI automation makes up a tiny fraction of datacentre utilisation, our CI is yet again a tiny fraction of that. Even if every CI in existence implemented this change, the effects would be negligible. 
 
I don't particularly think the change needs to be reverted, happy for it to stay in; we can address any problems if they come up. But I don't think the benefits claimed in the tweet are true in the vast majority of cases.

abidlabs:
> lose the ability to map a commit back to a CI failure

Is this the case? Perhaps I'm misunderstanding, but any time that there is a CI failure, you can still map it to a commit. What this does is cancel the CI tests that are running if another commit happens right after it. The difference is that some (short-lived) commits may not have an associated test status

abidlabs:
Oh wait I see your point @pngwn. If there is a failure, it could map to any of the successive commits that were grouped together before the CI test fully ran.

abidlabs:
In almost all likelihood, these successive commits should be tiny alterations of each other, so they can be easily analyzed together. But yeah, if it any problems do come up, we can rethink this

pngwn:
You won't know which of the previous commits is actually responsible for the failure. Imagine the following scenario:

- make changes for fix/ feature
- push
- ci starts
- forgot to format so you format, changing many files
- push
- previous Ci run is cancelled, new CI run starts
- CI fails

You have two significant commits and you don't know which one is responsible for the failure. The second commit is unlikely to have caused it (it is just formatting after all) so you look through your new feature to understand how it could have broken the tests. But autoformatting can generate malformed code and could be responsible.

There are other scenarios but this is one i have actually experienced myself. Like I say not a huge issue, we can wait and see.

abidlabs:
Makes sense, thanks for explaining

FarukOzderim:
> You won't know which of the previous commits is actually responsible for the failure. Imagine the following scenario:
> 
>     * make changes for fix/ feature
> 
>     * push
> 
>     * ci starts
> 
>     * forgot to format so you format, changing many files
> 
>     * push
> 
>     * previous Ci run is cancelled, new CI run starts
> 
>     * CI fails
> 
> 
> You have two significant commits and you don't know which one is responsible for the failure. The second commit is unlikely to have caused it (it is just formatting after all) so you look through your new feature to understand how it could have broken the tests. But autoformatting can generate malformed code and could be responsible.
> 
> There are other scenarios but this is one i have actually experienced myself. Like I say not a huge issue, we can wait and see.

@pngwn if only we had precommit hooks :D

Btw this brought an idea to me. What if we do not have a precommit hook but instead have a user prompt reminding the user to format the codebase before the commit or push?

pngwn:
This was just one example, there are others that don't involve formatting. I don't think the complexity is worth it, CI will tell you what is wrong and it is a quick fix.

FarukOzderim:
Yeah that is correct when the CI is fast :)

Issue #693, opened 2022-02-21
Video component is flipped horizontally


charlesollion:
**Describe the bug**
Videoplayer (and video) is flipped horizontally since 2.8 release.

**To Reproduce**
Build any video input or output component, without webcam

**Screenshots**
![videoflipped](https://user-images.githubusercontent.com/10595147/155003686-ae61d142-0fd7-422e-8a82-2c307c91e282.png)

**Device information (please complete the following information):**
 - OS: Linux
 - Browser Chrome
 - Gradio version  2.8+

**Additional context**
The problem comes from the -1 scale introduced for webcam processing:
```css
<style lang="postcss" global>
	video {
		-webkit-transform: scaleX(-1);
		transform: scaleX(-1);
	}
</style>

PR #726, opened 2022-02-23
Issue 693: Video component is flipped horizontally

dawoodkhan82:
# Description

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

Fixes: # (issue)
https://github.com/gradio-app/gradio/issues/693

# Checklist:

### DIFF
--- a/ui/packages/app/src/components/utils/Webcam.svelte
+++ b/ui/packages/app/src/components/utils/Webcam.svelte
@@ -66,10 +66,3 @@
 		/>
 	</button>
 </div>
-
-<style lang="postcss" global>
-	video {
-		-webkit-transform: scaleX(-1);
-		transform: scaleX(-1);
-	}
-</style>
--- a/ui/packages/app/src/components/utils/Webcam.svelte
+++ b/ui/packages/app/src/components/utils/Webcam.svelte
@@ -66,3 +66,10 @@
 		/>
 	</button>
 </div>
+
+<style lang="postcss">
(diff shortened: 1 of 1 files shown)


pngwn:
Do we want to keep the webcam flipped but prevent this from affecting other video elements? If so we can just remove the `global` attribute on line 70. By default svelte will scope CSS to the component but global removes that behaviour.

ui/packages/app/src/components/utils/Webcam.svelte
```diff
@@ -66,10 +66,3 @@
 		/>
 	</button>
 </div>
-
-<style lang="postcss" global>
-	video {
-		-webkit-transform: scaleX(-1);
-		transform: scaleX(-1);
-	}
-</style>
```
abidlabs, line 75: I thought we needed these lines because the webcam was flipped (#627)?

dawoodkhan82:
> @dawoodkhan82 Do we want to keep the webcam flipped but prevent this from affecting other video elements? If so we can just remove the `global` attribute on line 70. By default svelte will scope CSS to the component but global removes that behaviour.

done

ui/packages/app/src/components/utils/Webcam.svelte
dawoodkhan82: added back, but removed global, as per @pngwn suggestion

pngwn (approved):

PR #746, opened 2022-02-27
updated scripts to use pnpm and removed outdated readme

abidlabs:


pngwn (approved):
Only question I have on this is do we have another script that runs `pnpm i` separately? We could be duplicating the `pnpm i` runs. Not a huge issue as the installation is very fast.

FarukOzderim:
> Only question I have on this is do we have another script that runs `pnpm i` separately? We could be duplicating the `pnpm i` runs. Not a huge issue as the installation is very fast.

We could use **build_frontend.sh** script inside other scripts to remove duplication. But not that important.

abidlabs:
> Only question I have on this is do we have another script that runs `pnpm i` separately? We could be duplicating the `pnpm i` runs. Not a huge issue as the installation is very fast.

Gotcha. We don't have another script and it seems overkill to create one just for `pnpm i`. On my laptop, `pnpm i` takes less than 1 second if the installation is already up to date

PR #752, opened 2022-02-28
Update first launch commands to deliver the usage of ";" and app, pat…

### LINKED ISSUE #751: Update Launch Getting Started Guides
Usage of `iface.launch()` prints out `(<fastapi.applications.FastAPI at 0x7fef0dbbad10>,` in Google Colab and it is not fancy.

Let's update the first deliveries of `launch` in the getting_started guides so that people can adopt those usages.

FarukOzderim:
Fixes #751

### DIFF
--- a/demo/hello_world/run.py
+++ b/demo/hello_world/run.py
@@ -7,4 +7,4 @@ def greet(name):
 
 iface = gr.Interface(fn=greet, inputs="text", outputs="text")
 if __name__ == "__main__":
-    iface.launch()
+    iface.launch();
--- a/demo/hello_world_2/run.py
+++ b/demo/hello_world_2/run.py
@@ -11,4 +11,5 @@ iface = gr.Interface(
     outputs="text",
 )
 if __name__ == "__main__":
-    iface.launch()
+    app, path_to_local_server, share_url = iface.launch()
+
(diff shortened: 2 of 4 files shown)


FarukOzderim:
@abidlabs We might also want to update our colab links in the guides with the `;` usage, who owns those colab links?

abidlabs:
If we are going to return the parameters, we need to explain them too. My suggestion would be to just put a semicolon in all 3 examples

FarukOzderim:
I want people to learn what launch returns and use it if they need it, and just checked we even don't share that info in docs 😨. I will put a note inside the guide.

https://gradio.app/docs/#launch

abidlabs:
I have some small nits. Can I leave them here? 
* It's better to just have this line `app, path_to_local_server, share_url = iface.launch()` once, say in `hello_world_2` where the return values are explained. I would remove from `hello_world_3`
* For the sake of readability, I suggest replacing `path_to_local_server` with `local_url`, so the line looks like: `app, local_url, share_url = iface.launch()`
* The return values are listed but are not really explained. Suggest something like this:

**Sidenote**: The `Interface.launch()` method returns 3 values: (1) `app`, which is the FastAPI application that is powering the Gradio demo, (2) `local_url`, which is a `str` path to the local address of the server, and (3) `share_url`, the `str` path to the public address for this demo (if `share=True`, more [on this later]())

* Note that we haven't discussed `share` yet, so it's better to link to it.

Issue #723, opened 2022-02-23
[Blocks] Implement other event triggers besides `click` & support event triggers across multiple components


abidlabs:
Currently, the only event trigger we support is `click`. 

1. Can we add support additional event triggers? Some of these will be component specific, but something like an "on change" event would be useful for many different components. For example, in my Block below, I would like to run the function any time the input sliders change value

```python
block = gr.Blocks()

with block:
    gr.Markdown("Let's do some kinematics! Choose the speed and angle to see the trajectory.")
    
    with gr.Row():
        speed = gr.inputs.Slider(1, 30, default=25)
        angle = gr.inputs.Slider(0, 90, default=45)
    output = gr.outputs.Image(type="plot")
    btn = gr.Button("Run")
    btn.click(plot, [speed, angle], output)

block.launch()
```

2. It would be nice if I could easily assign a trigger to "if any of these two sliders change" without having to write two separate event trigger lines

3. It would be _really_ nice if I could just say something like "if anything changes in this Block" re-run all of the function(s). I would assume that many of our users want this functionality for their blocks. Could we have a parameter in `gr.Blocks()` like `live=True`. In other words, if you run `gr.Blocks(live=True)`, then you have an automatically responsive Block in which if any of the input components change, all of the functions are re-run and the outputs automatically refresh.

pngwn:
**Proposed Component Events**

| Component | Event | When? | Notes |
|---|---|--|--|
| **Audio** | `change`|  audio source changes |  should this be `on:data`? |
|  | `clear`| audio source is reset to nothing | |
| **Button** | `click` | button is clicked | |
| **Carousel** | `change` | current item changes | |
| **ChatBot** | `message` | chatbot posts a message | |
| **Checkbox** | `change` | checkbox is checked/ unchecked | |
| **CheckboxGroup** | `change` | checkbox is checked/ unchecked | |
| **Column** |  | | static layout component (nyi) | 
| **DataFrame** | `change` | data changes | should this be `on:data`? |
| **Dropdown** | `change` | selected option changes | |
| **File** | `change` | file data changes | should this be `on:data`? |
|  | `clear`| video source is reset to no value | |
| **Grid** |  | | static layout component (nyi) | 
| **HighlightedText** | `change` | content is passed in | output only |
| **HTML** | `change` | content is passed in | output only |
| **JSON** | `change` | content is passed in | output only |
| **Label** | `change` | content is passed in | output only |
| **Number** | `change` | number value changes | |
[...]

PR #758, opened 2022-03-01
Blocks Events

FarukOzderim:
Add change and save events on backend

#723


pngwn:
I don't think `save` makes sense for every component. Component events (other than `change`) should be implemented on a per component basis and not globally.

FarukOzderim:
@pngwn thanks for your comment, I realized the incorrect docstrings and updated them .D

Well the events are defined by components
ie:
´´´python
ct_run = gr.Button("Run")
ct_run.click(ct_model, inputs=[disease, ct_scan], outputs=ct_results)
´´´

Would you like there to be a control mechanism allowing save event just on certain components?

FarukOzderim:
Well actually we have some discussion here, @abidlabs , @aliabid94 , @pngwn.

We want to enable events just for some components, 
- we could write the supported components into the event function docstring. And throw warnings or errors if they are not respected.
- or we could write functions for every component, but we have 10+ component and will have multiple events in the future, I think this increases complexity of the design, not challenging but feels like a big duplication to me.

Not so different for me, whatever you guys prefer in terms of API. 1st is stronger for the UX, but not so different imo.

aliabid94 (approved):
LGTM. Could consolidate some shared code b/w save and change methods but this code will probably go through a lot of changes soon so can do later too.

abidlabs:
I thought in our conversation earlier today, we decided that each component was going to have its own set of events? Or are we planning on doing that, _but also_ defining `change()` and `save()` as events that all components support? If so, I understand `change()` but what does `save()` do?

(Sorry if these questions have already been discussed)

FarukOzderim:
> Well actually we have some discussion here, @abidlabs , @aliabid94 , @pngwn.
> 
> We want to enable events just for some components,
> 
>     * we could write the supported components into the event function docstring. And throw warnings or errors if they are not respected.
> 
>     * or we could write functions for every component, but we have 10+ components and will have multiple events in the future, I think this increases complexity of the design, not challenging but feels like a big duplication to me.
> 
> 
> Not so different for me, whatever you guys prefer in terms of API. 1st is stronger for the UX, but not so different imo.


Abubakar actually it is a discussion now.
Agree with Ali, we can decide to allow an event just for some components with a warning or event functions in the specific component level after trying them out

abidlabs:
To (1) avoid code duplication and (2) get static checks for supported events, could we do something like this?

* Define a generic `set_event_trigger()` method in the `Block` class:

```python
    def set_event_trigger(
        self, event: str, fn: str, inputs: List["Component"], outputs: List["Component"]
    ) -> None:
        """
        Adds change event to the component's dependencies.
        Whenever the component changes the function is triggered.
        Parameters:
            fn: function name
            inputs: input list
            outputs: output list
        Returns: None
        """
        if not isinstance(inputs, list):
            inputs = [inputs]
        if not isinstance(outputs, list):
            outputs = [outputs]
        Context.root_block.fns.append(fn)
        Context.root_block.dependencies.append(
            {
                "targets": [self._id],
                "trigger": "change",
            }
        )
```

* For each component, define supported events as:

```python
def click(self, event: str, fn: str, inputs: List["Component"], outputs: List["Component"]):
    self.set_event_trigger("click", fn=fn, inputs=inputs, outputs=outputs)
```

FarukOzderim:
Yeah it looks like a great design, it would work if every event will be defined similarly.

FarukOzderim:
@aliabid94 regarding support for singular elements, I think we should drop it as it creates a duality, WDYT?
https://github.com/gradio-app/gradio/pull/590#discussion_r807258246
```
if not isinstance(inputs, list):
            inputs = [inputs]
        if not isinstance(outputs, list):
            outputs = [outputs]
```

abidlabs:
I strongly think that we should keep the API as human-friendly as possible, so that if the user has only 1 input or 1 output, they can pass it in without wrapping it inside a list, i.e.

`gr.Interface(fn=predict, inputs="mic", outputs="audio")` should work

At this point, this is basically a convention in machine learning libraries. For example, in `keras`, you can pass in a single input to your `Model`, or you can pass in a list of inputs if it takes in multiple inputs: https://machinelearningmastery.com/keras-functional-api-deep-learning/

FarukOzderim:
> I strongly think that we should keep the API as human-friendly as possible, so that if the user has only 1 input or 1 output, they can pass it in without wrapping it inside a list, i.e.
> 
> `gr.Interface(fn=predict, inputs="mic", outputs="audio")` should work
> 
> At this point, this is basically a convention in machine learning libraries. For example, in `keras`, you can pass in a single input to your `Model`, or you can pass in a list of inputs if it takes in multiple inputs: https://machinelearningmastery.com/keras-functional-api-deep-learning/

It's not that handy but one of the disadvantages is duality, which should be avoided with APIs and languages. 
Another drawback is it is naming, a single element is not a plural(inputs), and requires further documentation/guide.

Though it is not a major point. So whatever you guys prefer. 😸

pngwn:
@FarukOzderim While that is true in prionciple, in this instance we would be forcing all users to put single inputs into a list even though the common case is to pass a single input rather than a list. So i think the discrepancy is worth it here due to the facxt that it makes the common case really intuitive but still supports the more complex case.

FarukOzderim:
Merging the PR for now, will adopt Abubakar's design suggestion after the component work is finished.

Issue #759, opened 2022-03-01
Loading from Spaces is Broken


abidlabs:
### Describe the bug

Getting this error when trying to use `gr.Interface.load()` with a Space

```
AttributeError                            Traceback (most recent call last)
c:\users\islam\dev\gradio-repos\gradio\gradio\external.py in get_spaces_interface(model_name, api_key, alias)
    299     try:
--> 300         config = json.loads(result.group(1))
    301     except AttributeError:

AttributeError: 'NoneType' object has no attribute 'group'
```

### Reproduction

import gradio as gr

gr.Interface.load("spaces/abidlabs/english_to_spanish")

### System Info

```shell
Gradio 2.8.5
```

### Severity

critical

abidlabs:
Closed via #760

PR #760, opened 2022-03-01
[hotfix] fixed loading from spaces

abidlabs:
Spaces changed how they were rendering Gradio demos, so this broke our regex for extracting the config for the demo. This hotfixes it, though perhaps we should identify a more robust solution for extracting the config.

### DIFF
--- a/gradio/external.py
+++ b/gradio/external.py
@@ -294,7 +294,7 @@ def get_spaces_interface(model_name, api_key, alias):
 
     r = requests.get(iframe_url)
     result = re.search(
-        "window.gradio_config = (.*?);[\s]*</script>", r.text
+        r"window.gradio_config = (.*?);[\s]*</script>", r.text
     )  # some basic regex to extract the config
     try:
         config = json.loads(result.group(1))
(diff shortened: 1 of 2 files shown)


gradio/inputs.py
```diff
@@ -123,6 +123,7 @@ def __init__(
         numeric: Optional[bool] = False,
         type: Optional[str] = "str",
         label: Optional[str] = None,
+        optional: bool = False,
```
abidlabs, line 126: But it turns out a secondary reason was that every input component includes an `optional` property in its config because the base Component class has the `optional` attribute. However, when we are recreating the input component from its config, some input components don't accept the `optional` parameter. As a fix, I added the `optional` attribute to every input component, even though in many cases it has no effect (which is documented). 

Is there a better way to do this? @aliabid94?

gradio/external.py
```diff
@@ -294,7 +294,7 @@ def get_spaces_interface(model_name, api_key, alias):
 
     r = requests.get(iframe_url)
     result = re.search(
-        "window.gradio_config = (.*?);</script>", r.text
+        "window.gradio_config = (.*?);[\s]*</script>", r.text
```
abidlabs, line 297: This was the main reason we were seeing #759

Issue #639, opened 2022-02-16
[Blocks] Allow any output component to be used to display statically 


abidlabs:
It should be possible to treat ANY output component like a static component by simply dumping an appropriate value to it. Like this:

```python
with Blocks() as block:
  outputs.Dataframe(df)
```

FarukOzderim:
I think we should reconsider this with Inputs and Outputs merging into Components.

Inputs and Outputs both can have default values.

abidlabs:
In addition to default values, components should be able to display suitable variables statically for convenience. For example, we should be able to use `Image` to display an image easily, `Dataframe` to display a pandas DataFrame, `Audio` to display audio files, etc.

As an example of why this is useful, see this website: https://erlj.notion.site/Neural-Instrument-Cloning-from-very-few-samples-2cf41d8b630842ee8c7eb55036a1bfd6 -- we want people to be able to put stuff like this together with Blocks

FarukOzderim:
@pngwn I think this will be the case when a component is not used as either an input or output in any event.
I think you are handling this currently as well, assigning it to you :)

abidlabs:
Right now, this does not work. The components don't display the `default_value` parameter: 

For example:

```python
import gradio as gr

b = gr.Blocks()

with b:
    gr.Textbox("abc")
    
b.launch()
```

produces an empty textbox instead of one with "abc" in it

![image](https://user-images.githubusercontent.com/1778297/159591234-831b4599-fc17-4c48-85f0-081f7d751246.png)

abidlabs:
This should be closed, right @pngwn?

pngwn:
In principle, yeah. They might not be as 'static' as we'd like in some cases but i think the core functionality is there.

abidlabs:
Ok sounds good. Closing this general issue and we can reopen specific ones if needed

Issue #731, opened 2022-02-24
[Blocks] String shortcuts for components


aliabd:
Can we support string shortcuts for components like in `Interface`? 

shortcuts like "sketchpad" save from having to define 4 parameters (`gr.inputs.Image(image_mode='L', source="canvas", shape=(28, 28), invert_colors=True)`)

Not currently supported in blocks.

FarukOzderim:
I think using strings for components names are not legit and hacky.
We can have these pre-prepared components in a class and import from there.

```python
from gradio.components.Templates import Sketchpad
inputs = [Sketchpad]
```
or
```python
from gradio import Templates
inputs = [Templates.Sketchpad]
```

abidlabs:
String identifiers are  pretty common practice in parameters the way we have them in `inputs` and `outputs`. For example, see https://keras.io/api/optimizers/ (`"adam"` and `Adam()` both work). I actually think they are part of the convenience of high level APIs like Keras and Gradio.

That being said, they are not very convenient in the Blocks context, since you can't just have a string by itself the way you could have a Class being instantiated. I think the approach you are suggesting @FarukOzderim make sense to me. I would prefer the last one, if possible, so that all of the components live in the same namespace. I think there is a small typo and it should read:

```python
from gradio.component import Sketchpad
inputs = Sketchpad()
```

The only I don't love is having to import from a separate module

FarukOzderim:
About this issue let's give the final decision

- for blocks we want to have Template components that are implemented by us. 
- What about Interface? Would you want to drop support shortcuts within Interface and move to Templates usage as well? Or still support shortcuts within Interface?

abidlabs:
> for blocks we want to have Template components that are implemented by us.

To clarify, how would this work? Would it be that we pass in `Template("sketchpad")` and it renders the `Image` component with the "sketchpad" parameters? If so, that seems reasonable to me.

> What about Interface? Would you want to drop support shortcuts within Interface and move to Templates usage as well? Or still support shortcuts within Interface?

We should still support shortcuts within Interface: (a) because they make Interface easy to use for common use cases (b) backwards compatibility

FarukOzderim:
Ok then;

- We will still support shortcuts within interface.
- We will create custom components with necessary parameters and etc for the shortcuts and they will work like this:
```
from gradio import Sketchpad
inputs = Sketchpad()
```

abidlabs:
[...]

PR #765, opened 2022-03-02
Blocks components

FarukOzderim:
Combine all input output and static components under components. A huge refactoring, detailed review would be great!

Fixes: #639 
- create general structure
- move Textbox into component.py


abidlabs:
This generally looks like a promising approach to me! 

However, specifically for the `Textbox`, there are some changes that need to be made. The reason for that is the `type` is deprecated for the `inputs.Textbox` but not for the `outputs.Textbox` where it is actually quite important. 

So what we should do is to accept the `type` parameter but have its behavior mimic the behavior in `outputs.Textbox` constructor. 

I'll add specific comments in the code to show what I mean

gradio/inputs.py
```diff
@@ -108,151 +108,6 @@ def get_template_context(self):
         }
 
 
-class Textbox(InputComponent):
```
abidlabs, line 111: Instead of deleting the class completely, I would do this so that you can throw a warning appropriately:

```python
class Textbox(InputComponent):
    """
    Component creates a textbox for user to enter input. Provides a string as an argument to the wrapped function.
    Input type: str
    Demos: hello_world, diff_texts
    """

    def __init__(
        self,
        lines: int = 1,
        placeholder: Optional[str] = None,
        default: str = "",
        numeric: Optional[bool] = False,
        type: Optional[str] = "str",
        label: Optional[str] = None,
        optional: bool = False,
    ):
        """
        Parameters:
        lines (int): number of line rows to provide in textarea.
        placeholder (str): placeholder hint to provide behind textarea.
        default (str): default text to provide in textarea.
        numeric (bool): DEPRECATED. Whether the input should be parsed as a number instead of a string.
        type (str): DEPRECATED. Type of value to be returned by component. "str" returns a string, "number" returns a float value. Use Number component in place of number type.
        label (str): component name in interface.
        optional (bool): this parameter is ignored.
        """
        if numeric or type == "number":
            warnings.warn(
                "The 'numeric' type has been deprecated and this parameter has no effect. Use the Number input component instead.",
                DeprecationWarning,
            )
        return components.Textbox(lines=lines, 
                                  placeholder=placeholder, 
                                  default=default, 
                                  label=label,
                                  optional=optional)
```

gradio/outputs.py
```diff
@@ -49,50 +49,6 @@ def deserialize(self, x):
         return x
 
 
-class Textbox(OutputComponent):
```
abidlabs, line 52: If you delete this, and someone does the following import:

`from gradio.components import Textbox` 

Does this work fine?

gradio/component.py
```diff
@@ -95,3 +111,233 @@ def get_all_shortcut_implementations(cls):
...
+        label (str): component name in interface.
+        numeric (bool): DEPRECATED.
+        type (str): DEPRECATED.
+        optional (bool): DEPRECATED
+        """
+        self.lines = lines
+        self.placeholder = placeholder
+        self.default = default
+        self.type = "str"
+        if numeric:
```
abidlabs, line 225: Remove this here

gradio/component.py
```diff
@@ -95,3 +111,233 @@ def get_all_shortcut_implementations(cls):
...
+            warnings.warn(
+                "The 'numeric' type has been deprecated. Use the Number component instead.",
+                DeprecationWarning,
+            )
+        if type:
+            warnings.warn(
```
abidlabs, line 231: Remove this here. Because we actually _do need_ to store the `type` parameter

Instead, this should be:

```python
        self.type = type
```

gradio/component.py
```diff
@@ -95,3 +111,233 @@ def get_all_shortcut_implementations(cls):
...
+        }
+
+    @classmethod
+    def get_shortcut_implementations(cls):
+        return {
+            "text": {},
+            "textbox": {"lines": 7},
+        }
+
+    def preprocess(self, x: str) -> str | float:
```
abidlabs, line 254: Now that we are deprecating the input type, this can just be:

```python
    def preprocess(self, x: str) -> str:
        """
        Parameters:
        x (str): text input
        """
        return x
```

gradio/component.py
```diff
@@ -95,3 +111,233 @@ def get_all_shortcut_implementations(cls):
...
+        result = []
+        for token, score in zip(tokens, scores):
+            result.append((token, score))
+            result.append((self.interpretation_separator, 0))
+        return result
+
+    def generate_sample(self) -> str:
+        return "Hello World"
+
+    def postprocess(self, y: str):
```
abidlabs, line 335: This is the main problem. The Textbox postprocess *needs* to be type aware.

```python
    def postprocess(self, y):
        """
        Parameters:
        y (str): text output
        Returns:
        (Union[str, number]): output value
        """
        if self.type == "str" or self.type == "auto":
            return str(y)
        elif self.type == "number":
            return y
        else:
            raise ValueError(
                "Unknown type: " + self.type + ". Please choose from: 'str', 'number'"
            )
```

gradio/inputs.py
abidlabs: See below for why we shouldn't throw the warning inside the `components.Textbox` class itself

gradio/outputs.py
abidlabs: If not, we should do something similar to what I have `inputs.Textbox` above where we construct one class from the other.

But if these imports all work, then I agree it would be better to avoid it

aliabid94:
I put these thoughts in gradio-technical channel  but probably more appropriate here:

~ ~ ~
- All of our output components should also be able to be displayed in "static" mode, e.g. image or dataframe data passed to the component directly from the app code that never changes.
- Most components (but not all) will have an input mode, output mode, and static mode, e.g. Image component has an input mode where you can drag and drop an image with editor, an output mode that only displays an image result, and a static mode that always displays the same constant image. A radio button, on the other hand, would only have an input mode. The input mode may look different from the output and static modes (which look the same)
- In Blocks, we can infer whether a component is A) input, B) output, C) both, or D) static if, A) it is an input to a dependency, B) it is an output to a dependency, C) it is both A and B, or D) if it is not in any dependency.
- Ideally it would be nice if the first argument in gr.Component() syntax is the default / static value.  That way we can quickly load static images, dataframes and markdown with `gr.Image('path/to/image.jpg')` or `gr.Markdown('# Hello World')` , instead of `gr.Markdown(default='#Hello World')`.
This is stuff we've probably discussed before but just wanted to make sure we're on the same page.

~ ~ ~

aliabid94:
Also, how do we keep backward compatibility with any code that may have specified Component arguments without keywords? (this is something [we can disable](https://www.python.org/dev/peps/pep-3102/) if we want). So for example, if someone uses `gr.inputs.Image((200,200), "RGB")` instead of `gr.inputs.Image(shape=(200,200), color_mode="RGB")`. Abubakar's suggestion of keeping inputs.Textbox and having that call components.Textbox would solve this issue.

gradio/component.py
FarukOzderim: I am planning to keep breaking changes to the minimum. I don't think we need number type in the Textbox, because we will have a Number component both for the input and the output. People can use the Number component if they need it.

FarukOzderim: And having it for the output case makes the design complexier and feels not that meaningful to me when we have a Number component.

gradio/outputs.py
FarukOzderim: I am not in favor of having different initilizations for the input or output, I think that makes combining them meaningless and would prefer having a NewComponent type for blocks instead of this.
People will be able to reference Textbox through components, inputs and outputs. And use it both as an input or output.

However if are going to create code duplication by having different classes for input and output let's have NewComponents and don't break anything that Interface uses.

If we are going to combine components input and output should merge into one.


These are my thoughts, WDYT?

FarukOzderim:
I am not in favor of having different initilizations for the input or output, I think that makes combining them meaningless and would prefer having a NewComponent type for blocks instead of this.
People will be able to reference Textbox through components, inputs and outputs. And use it both as an input or output.

However if are going to create code duplication by having different classes for input and output let's have NewComponents and don't break anything that Interface uses.

If we are going to combine components input and output should merge into one.

These are my thoughts, WDYT?

FarukOzderim:
> Also, how do we keep backward compatibility with any code that may have specified Component arguments without keywords? (this is something [we can disable](https://www.python.org/dev/peps/pep-3102/) if we want). So for example, if someone uses `gr.inputs.Image((200,200), "RGB")` instead of `gr.inputs.Image(shape=(200,200), color_mode="RGB")`. Abubakar's suggestion of keeping inputs.Textbox and having that call components.Textbox would solve this issue.

We are changing our component structure & design. I think it is ok to move to only keyword parameters except `default` parameter.

abidlabs:
> I am not in favor of having different initilizations for the input or output, I think that makes combining them meaningless

I don't think it makes it meaningless. We still avoid significant code duplication and create a single component that is the *preferred* way of instantiating components in the future. However, what @aliabid94 is saying is how we can preserve backwards compatibility and reduce the amount of code that break. We will update our docs to use the new `components.py` module, but there's a lot of existing code / tutorials / examples etc. 

Here's how it would like for the `Textbox` class (roughly).  

`inputs.py`

```py
class Textbox(InputComponent):
    """
    Component creates a textbox for user to enter input. Provides a string as an argument to the wrapped function.
    Input type: str
    Demos: hello_world, diff_texts
    """

    def __init__(
        self,
        lines: int = 1,
        placeholder: Optional[str] = None,
        default: str = "",
        numeric: Optional[bool] = False,
        type: Optional[str] = "str",
        label: Optional[str] = None,
        optional: bool = False,
    ):
        if numeric or type == "number":
            warnings.warn(
                "The 'numeric' type has been deprecated and this parameter has no effect. Use the Number input component instead.",
                DeprecationWarning,
            )
        return components.Textbox(lines=lines, 
                                  placeholder=placeholder, 
                                  default=default, 
                                  label=label,
                                  optional=optional)
```

`outputs.py`

class Textbox(OutputComponent):
    """
    Component creates a textbox to render output text or number.
    Output type: Union[str, float, int]
    Demos: hello_world, sentence_builder
    """

    def __init__(self, type: str = "auto", label: Optional[str] = None):
        """
        Parameters:
        type (str): Type of value to be passed to component. "str" expects a string, "number" expects a float value, "auto" detects return type.
        label (str): component name in interface.
        """
        return components.Textbox( 
                                  type=type, 
                                  label=label)


`components.py`

The full code for the components class

This way we achieve merged components with minimal code duplication _and_ no code breakage.

gradio/component.py
abidlabs: We don't have a `Number` output component right now. We could create one but we don't need to break existing code that uses `outputs.Textbox(type="number")`

FarukOzderim:
> > We are changing our component structure & design. I think it is ok to move to only keyword parameters except default parameter. We can warn the users about this.
> 
> Why do this? We should avoid unnecessary code breakage as much as possible

Mainly to support Ali's suggestion.
While changing our design, it is a good idea to move into keyword only parameters to block any problems that could generate with positional parameters.

FarukOzderim:
With an offline chat we decided on:

- We'll keep inputs.py and outputs.py for backwards compatibility but they will show deprecation warnings and will not be documented in favor of the new components.py classes

FarukOzderim:
With the latest commit xray demo is working, you can take a look there.

From now on I will stop supporting component_type in the config, with the new component design.

gradio/inputs.py
```diff
@@ -19,12 +19,39 @@
 from ffmpy import FFmpeg
 
 from gradio import processing_utils, test_data
-from gradio.component import Component
+from gradio.component import Component, Textbox
```
abidlabs, line 22: Suggest importing components as:

`from gradio import components`

And then having something like:

`class Textbox(components.Textbox)` for readability?

abidlabs:
Very small nit. Can we rename `component.py` to be `components.py` (the same way `inputs` and `outputs` and `strings` and `routes` are all pluralized)?

It also reads better to have:

`from gradio.components import Textbox, Image`, etc

FarukOzderim:
> Very small nit. Can we rename `component.py` to be `components.py` (the same way `inputs` and `outputs` and `strings` and `routes` are all pluralized)?
> 
> It also reads better to have:
> 
> `from gradio.components import Textbox, Image`, etc

That's actually is imported like that, could change name of the file as well :)

gradio/__init__.py
```diff
@@ -1,5 +1,8 @@
 import pkg_resources
 
+import gradio.component as components
```
FarukOzderim, line 3: Here it is

gradio/inputs.py
FarukOzderim: Well actually importing the related class and using it directly is corresponding with PEP-8 as far as I know :)

abidlabs:
> That's actually is imported like that, could change name of the file as well :)

Yup I think refactoring it to be `components.py` would be nice

gradio/inputs.py
abidlabs: Oh I see

abidlabs:
Regarding components like `Image` who have both an input `type` and and output `type`, how should we handle the parameter naming? The simplest approach would be to have `Image` take both an `input_type` and `output_type` parameter. 

Some thoughts:
* For the sake of consistency, should we name all of the parameters, even for components that only take one `type`, either `input_type` or `output_type`? So `components.Textbox` would take an `output_type` instead of `type`
  * Disadvantage: longer parameter name
  * Advantage: consistent API (no need to look up what the parameter is called) + the purpose of parameter is clearer  
* Of course, we shouldn't rename the parameters in `inputs.Textbox` or `outputs.Textbox` for reverse compatibility

Cc @FarukOzderim @aliabid94 @pngwn

FarukOzderim:
> Regarding components like `Image` who have both an input `type` and and output `type`, how should we handle the parameter naming? The simplest approach would be to have `Image` take both an `input_type` and `output_type` parameter.
> 
> Some thoughts:
> 
>     * For the sake of consistency, should we name all of the parameters, even for components that only take one `type`, either `input_type` or `output_type`? So `components.Textbox` would take an `output_type` instead of `type`
>       
>       * Disadvantage: longer parameter name
>       * Advantage: consistent API (no need to look up what the parameter is called) + the purpose of parameter is clearer
> 
>     * Of course, we shouldn't rename the parameters in `inputs.Textbox` or `outputs.Textbox` for reverse compatibility
> 
> 
> Cc @FarukOzderim @aliabid94 @pngwn

I feel like having different types for input and output could complexify it. And would not be so different for having inputs and outputs independently.
Why not combine them together and take the same type for both input and output? `type=pil `means input_type and output_type is pil

FarukOzderim:
Yeah!

Btw I think having just output_type would be enough, and type would work as input_type, but it works both ways!

gradio/components.py
```diff
@@ -0,0 +1,2412 @@
...
+    Component creates a textbox for user to enter string input or display string output. Provides a string as an argument to the wrapped function.
+    Input type: str
+    Output type: str
+
+    Demos: hello_world, diff_texts, sentence_builder
+    """
+
+    def __init__(
+        self,
+        default: str = "",
```
abidlabs, line 214: Rename from `default` to `default_value` as per discussion

(this change applies to all components)

abidlabs:
You've probably seen this @FarukOzderim but the CircleCI tests aren't running because of circular imports

gradio/components.py
```diff
@@ -0,0 +1,2412 @@
...
+        return y
+
+    def deserialize(self, x):
+        """
+        Convert from serialized output (e.g. base64 representation) from a call() to the interface to a human-readable version of the output (path of an image, etc.)
+        """
+        return x
+
+
+class Dropdown(Radio):
```
abidlabs, line 883: Nice to make `Dropdown` a subclass of `Radio`!

gradio/components.py
```diff
@@ -0,0 +1,2412 @@
...
+    def __init__(
+        self,
+        default=None,
+        *,
+        shape: Tuple[int, int] = None,
+        image_mode: str = "RGB",
+        invert_colors: bool = False,
+        source: str = "upload",
+        tool: str = "editor",
+        type: str = "numpy",
```
abidlabs, line 931: Needs `output_type`

gradio/components.py
```diff
@@ -0,0 +1,2412 @@
...
+
+    def __init__(
+        self,
+        default: Optional[float] = None,
+        *,
+        minimum: float = 0,
+        maximum: float = 100,
+        step: Optional[float] = None,
+        label: Optional[str] = None,
+        **kwargs,
```
abidlabs, line 498: Why do we have `**kwargs` in all of the components? Seem to have no effect?

abidlabs:
Ok so as I've been reviewing the code, I'm realizing the approach of using `type` to refer to both input type and output type by default is not going to work @FarukOzderim. 

The reason is the "auto" parameter value we use for the output type for many of the components (e.g. `Image`, `Label`, and many others). "auto" means that the Output component's type should be determined based on inspecting the value that is provided to it. For example, if a "pil" image is provided, then it understands it should apply a postprocessing function that converts "pil" images to base64. This is super handy is it allows users to basically "not worry" about output types and only provide "type" to output components if they want to restrict the output component in some way. That is why for all of the output components that support the "auto" type, it is the default.

Now "auto" can be an output type, but there is no way to make it an "input" type since we cannot know _a priori_ what format a user expects the input data to look like for their prediction function. They have to tell us. 

If `type` refers to both "input_type" and "output_type", this leads us to several problems...

* Can `type` be "auto"? As the code is currently written, it will not work because the preprocessing functions do not allow `type` to be auto (example: `components.py, line 1057`), while the postprocessing functions (example:  `components.py, line 1179`) do. 
* What should the default value of the "output type" be? As it is currently proposed, it will be the same as the "input type." However, this will break lots of existing examples, which relies on the output component being inferred automatically and not necessarily being the same as the input. (Consider for example a method that takes in a PIL image, modifies it, saves it, and returns the file path.

Let me know you wanna sync about this @FarukOzderim. Happy to review the rest of the code, but we should get this sorted out first.

aliabid94:
How about getting rid of output_type? It can be "auto" always so we determine it ourselves, and then we just have kwarg "type" which refers to the desired input type.

abidlabs:
Huh that actually could work. I skimmed through the output docs and we should always be able to tell what the right output type is by looking at the returned value

The only exception is the Video output which has an interesting  type parameter. But that parameter should really be called something else, like “output_format”

FarukOzderim:
> You've probably seen this @FarukOzderim but the CircleCI tests aren't running because of circular imports

The complete version is on the `gradio-dash-fe` branch. We shall continue branching or updating from that branch.

gradio/components.py
FarukOzderim: I am planning to do it after this PR is completed,
But we need to decide on should we have default_value on each component? Because some components feel like not eligible to having a default_value. A few examples:

- json
- highlightedtext
- markdown
- buttown

@aliabid94 @abidlabs

gradio/components.py
FarukOzderim: They are optional parameters and they are sent to `Component` class. They are on every component for the design. And we can make use of them in `Component `or a particular component class whenever we wish.

FarukOzderim:
> > You've probably seen this @FarukOzderim but the CircleCI tests aren't running because of circular imports
> 
> The complete version is on the `gradio-dash-fe` branch. We shall continue branching or updating from that branch.

Never mind it, I am going to merge that branch here.

FarukOzderim:
> How about getting rid of output_type? It can be "auto" always so we determine it ourselves, and then we just have kwarg "type" which refers to the desired input type.

On which components would you like to have this feature, not necessary to have it on all of them I suppose?

gradio/components.py
abidlabs: `JSON`, `HighlightedText`, and `Markdown` should have a "default value" parameter because that is how we will use them statically. 

E.g.

```
Markdown("""
# My cool demo
Try my **cool demo** below
""")
```

I think the "default value" of a button can just the text that is displayed on it? Or we could have a separate parameter, I think either is fine.

abidlabs: Quick comment: what about just dropping "default" and calling the parameter `value`?

FarukOzderim: image value?
video value?
textbox value?

Does not feel intuitive to me, better use default or default_value

abidlabs:
It should be all of the components whose output forms currently have a `type` parameter. It looks like that is:
* Audio
* Video
* Image
* Label
* Dataframe

We no longer need this parameter and instead should just use the "auto" logic to infer what the output type is.

gradio/components.py
abidlabs: Ok `default_value` is fine too

aliabid94:
default_value seems like a good name. And agree with removing type for the components @abidlabs listed above

FarukOzderim (commented):
Thanks for the updates @abidlabs , great catch on my missing removals :D

gradio/external.py
```diff
@@ -4,7 +4,7 @@
 
 import requests
 
-from gradio import inputs, outputs
+from gradio import components
```
FarukOzderim, line 7: Should not we test gradio.inputs and gradio.outputs as well? We are still supporting those components.

gradio/interpretation.py
```diff
@@ -198,7 +198,7 @@ def quantify_difference_in_label(interface, original_output, perturbed_output):
             score = diff(original_label, perturbed_label)
         return score
 
-    elif isinstance(output_component, Textbox):
+    elif isinstance(output_component, (Textbox, Number)):
```
FarukOzderim, line 201: I think only having Number here would be enough, because we only support Text with Textbox and Number with Number.

gradio/components.py
```diff
@@ -1567,9 +1567,6 @@ def postprocess(self, y):
     def deserialize(self, x):
         return processing_utils.decode_base64_to_file(x).name
 
-    def restore_flagged(self, dir, data, encryption_key):
```
FarukOzderim, line 1571: Why did we remove this function? I remember it being in the output component of `Audio`.

gradio/components.py
```diff
@@ -52,9 +52,7 @@ def get_template_context(self):
         """
         :return: a dictionary with context variables for the javascript file associated with the context
         """
-        return {
-            "label": self.label,
-        }
+        return {"name": self.__class__.__name__.lower(), "label": self.label}
```
FarukOzderim, line 55: Why did we add name to get_template_context? There was no in Input or Output.

gradio/external.py
abidlabs: You're right, we should still test `gradio.inputs` and `gradio.outputs`. All I'm doing here is just updating `external.py` to use `components` instead of `inputs` and `outputs`

gradio/interpretation.py
abidlabs: Makes sense.

gradio/components.py
abidlabs: Yes but both `InputComponent` and `OutputComponent` had this line:

https://github.com/gradio-app/gradio/blob/1a4302ce883986a5be377c74369ac0fdc45ce307/gradio/inputs.py#L107

And the parent `Component` class would add it:

https://github.com/gradio-app/gradio/blob/1a4302ce883986a5be377c74369ac0fdc45ce307/gradio/component.py#L25

gradio/components.py
abidlabs: Hmmm it looks like we have a discrepancy in how the `inputs.Audio` and `outputs.Audio` are built from flagged. `inputs.Audio` doesn't have this line. Let me look into this some more.

Thanks for catching this

test/test_inputs.py
```diff
@@ -315,16 +311,6 @@ def test_in_interface(self):
         self.assertEqual(iface.process([["a", "c"]])[0], ["a|c"])
         self.assertEqual(iface.process([[]])[0], [""])
         checkboxes_input = gr.inputs.CheckboxGroup(["a", "b", "c"], type="index")
-        iface = gr.Interface(
```
FarukOzderim, line 330: Hmm, why did we remove these checks?

gradio/components.py
FarukOzderim: What is the latest situation about this issue? I see you removed this function from several components in this [commit](https://github.com/gradio-app/gradio/pull/765/commits/f2a5ecdca6d5ad24b150ff1c154c4cbbf1da1af1). What was its functionality and why remove it?

FarukOzderim:
Btw I think we should remove support for text shortctus and create prebuilt components for that.
Related discussion #731

gradio/components.py
abidlabs: Yeah I'm actually still looking into it. This function is designed to take flagged data and load it back into the interface. It seems like we were handling this differently for inputs and outputs, which was causing some things to break. I can look into it some more when I work on the output tests

test/test_inputs.py
abidlabs: Because we no longer have interpretation for textbox output, as per this comment: https://github.com/gradio-app/gradio/pull/765#discussion_r830451228

FarukOzderim:
I am merging this PR to work with smaller PRs in the future, please highlight anything if not discussed yet.

PR #769, opened 2022-03-03
Rewrite state to be backend-based

abidlabs:
To support live streaming use cases, we are working on a new backend-based state system. The rationale is that live streaming, etc. requires storing the audio/video recording recorded "so far" in the state and we don't want this to get sent back and forth from the front end back to the backend each time.

Some requirements:

1. `State` should store the state in the backend rather than sending it back and forth each time
2. `State` should still allow the funciton to be run normally (outside of a Gradio interface)
3. In `Blocks`, we may have many intermediate hidden variables, can we support this too with `State`?
4. If we can preserve backwards compatibility with the old way of doing `State`, that would be great


gradio/routes.py
```diff
@@ -210,8 +211,15 @@ def api_docs(request: Request):
 @app.post("/api/predict/", dependencies=[Depends(login_check)])
 async def predict(request: Request, username: str = Depends(get_current_user)):
     body = await request.json()
+    session_hash = body.get("session_hash", None)  # TODO(aliabid94): send from frontend
```
abidlabs, line 214: @aliabid94 can you have the frontend send a hash back here?

FarukOzderim: @abidlabs instead of getting smt from the request body it is better to use validations in FastAPI.

abidlabs: Thanks for the tip @FarukOzderim. I've updated to use pydantic models for this route.

There are a few other routes were we should make this change. Opened #777 to address this

abidlabs:
@aliabid94 feel free to merge in if it looks good to you

Issue #552, opened 2022-02-07
API Related Issue


samarth70:
The post request shows Internal Server error when default example is given on interface. Otherwise post request works.How can it be fixed?

abidlabs:
Sorry I don't understand the issue. Can you please provide a code example and more details on what is not working?

samarth70:
This is model's payload and response.
![image](https://user-images.githubusercontent.com/44370096/152912413-f0609497-a7a8-491b-9476-2d29e607d3f8.png)

This is when a default input is given .According to this the api is working.
![image](https://user-images.githubusercontent.com/44370096/152912606-216f9ee6-44d0-4c01-9316-f0df4bef7692.png)
 

**But** when I add my custom inputs as examples in interface.It shows Internal Server error.
![image](https://user-images.githubusercontent.com/44370096/152912868-622687fe-9be6-4c27-a55c-f25f3bc2e374.png)

abidlabs:
@aliabd can you look into this?

PR #787, opened 2022-03-07
API Docs Fixes

aliabd:
# Description
A few changes and fixes to api docs as requested by users:
* Sample inputs will now be serialized correctly if pulled from examples. Fixes: #552 before and after below.
Screen Shot 2022-03-07 at 4 45 53 PM
Screen Shot 2022-03-07 at 4 41 09 PM

* New related methods section (originally requested by @nateraw) that quickly guides users on conversions like image file to base64. Only implemented (and shows) for images now (in either input or output) but can be easily expanded if such functions exist for other components. 
 
Screen Shot 2022-03-08 at 4 28 03 AM

 * Added copy button and styling that matches gradio guides.

### DIFF
--- a/gradio/inputs.py
+++ b/gradio/inputs.py
@@ -374,7 +374,7 @@ class Slider(InputComponent):
         if step is None:
             difference = maximum - minimum
             power = math.floor(math.log10(difference) - 2)
-            step = 10**power
+            step = 10 ** power
         self.step = step
         self.default = minimum if default is None else default
         self.test_input = self.default
--- a/gradio/inputs.py
+++ b/gradio/inputs.py
@@ -374,7 +374,7 @@ class Slider(InputComponent):
         if step is None:
             difference = maximum - minimum
             power = math.floor(math.log10(difference) - 2)
-            step = 10 ** power
+            step = 10**power
         self.step = step
         self.default = minimum if default is None else default
         self.test_input = self.default
--- a/gradio/processing_utils.py
+++ b/gradio/processing_utils.py
@@ -336,7 +336,7 @@ def _convert(image, dtype, force_copy=False, uniform=False):
             Output image array. Has the same kind as `a`.
         """
         kind = a.dtype.kind
-        if n > m and a.max() < 2**m:
+        if n > m and a.max() < 2 ** m:
             return a.astype(_dtype_bits(kind, m))
         elif n == m:
             return a.copy() if copy else a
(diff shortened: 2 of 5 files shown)


abidlabs:
Looks good @aliabd! My only confusion is around the "Related Methods" section. As far as I understand, the purpose of this section is to shed some light on how we are preprocessing and postprocessing the data, but I'm not sure how helpful just having the method headers there is. 

image

Some suggestions to improve it:
* Either provide the full body of code for the methods or link to them on GitHub
* Remove `import gradio` -- that's implied and it's not like someone is going to run this code (unless I'm misunderstanding the purpose of this section)
* Rename from "Related methods" to "Preprocessing and postprocessing functions"

aliabd:
> My only confusion is around the "Related Methods" section. As far as I understand, the purpose of this section is to shed some light on how we are preprocessing and postprocessing the data, but I'm not sure how helpful just having the method headers there is.

You are misunderstanding it. The purpose of this section is to help users convert their inputs to the type required by the API, and their outputs from the type returned from the API. In the example of Image, I had gotten a few messages asking for help converting images to the correct base64 formats and from those formats to a file. I kept sending the processing_utils.py file  because I realized that's what we do as well in pre and post processing. 

So basically I'm exposing the methods so that they can use them. In my experience with users, almost everyone using the api of an image model is converting to and from base64. which they can easily do with these functions if they import gradio. Let me know if this makes sense, and if you have suggestions on renaming/expanding so that it's clearer.

abidlabs:
Got it that makes sense. I take back all 3 of my comments then, but then we should make this clearer on the page. 

* You could do this through inline comments:

```python
# To convert your image file into the format required by the API
import gradio as gr
gr.processing_utils.encode_url_or_file_to_base64(path)

# To convert the base 64 image returned by the API to an image file
...
```

* It would be nice to have syntax highlighting here if that's easy to get

aliabd:
Thanks for the suggestions! Made some changes: 
* Added syntax highlighting and inline comments, as well as expanded to audio, video, and file. They all use the same 2 functions, which are general functions that can detect the filetype by extension. The only difference currently is in the comment. But in the future we can probably add more component specific relevant methods like cropping for images, etc 

Screen Shot 2022-03-10 at 3 07 10 AM

* I also discovered a bug with examples being pulled as sample input when the examples parameter is pointing to a directory instead of a list of examples. To avoid too much complexity I decided to only render a sample input from examples if they are a list (otherwise, a generated sample is used)

PR #797, opened 2022-03-09
Website Tracker Slackbot

aliabd:
Building a slackbot to send a notification when the website is no longer tracking master

Has not been tested

### DIFF
--- a/website/reload_website.sh
+++ b/website/reload_website.sh
@@ -9,7 +9,7 @@ if grep -q "Already up to date." /tmp/git_changes.txt; then
 else
     if grep -q "error" /tmp/git_changes.txt; then
         LOGS=`cat /tmp/git_changes.txt`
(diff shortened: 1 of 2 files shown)


abidlabs:
Awesome

aliabid94:
It will only send a notification if the docker steps fail. This is necessarily the same as tracking master. For example, when earlier, the git pull step failed so it was not tracking master, this would not have caught that.

aliabid94:
also of course, move the slack url webhook to an os variable

aliabd:
Tested significantly in devserver. but could not test the `tail -n 25 /var/mail/ubuntu` command as its only on the gradio instance. 

@aliabid94 I can catch git pull errors now and read the webhook URL from an os variable (deleted old one)

aliabd:
This is ready to be merged. On the gradio machine we'll install `jo` for json formatting and the cron job will point to the new shell script.

Issue #191, opened 2021-05-25
Request for streaming microphone input component


m-ali-awan:
Once we record by microphone, I am not able to listen anything.
I am trying to have a real-time prediction with fastai trained model.

This is my code:
`
labels=learn.dls.vocab
def predict_audio(audio):
    at=AudioTensor.create(audio)
    pred,pred_idx,probs = learn.predict(at)
    return {labels[i]: float(probs[i]) for i in range(len(labels))}
`
`   
gr_interface = gr.Interface(fn=[
                                predict_audio], inputs=gr.inputs.Audio(source="microphone"),
                            outputs=gr.outputs.Label(num_top_classes=2),
                            title="Audio Classification:",
                            description='{}'.format(labels),
                            interpretation="default")
gr_interface.launch(debug=True)
`

Thanks for any help.

aliabid94:
can you link or provide the code for the entire file? also which platform & browser?

m-ali-awan:
Thanks, for replying.
I am using Colab pro, and fastai.
Please help me out to handle this:

`AssertionError: Expected an input of type in 
  - <class 'pandas.core.series.Series'>
  - <class 'str'>
  - <class 'fastaudio.core.signal.AudioTensor'>
 but got <class 'tempfile._TemporaryFileWrapper'>`

m-ali-awan:
As if I use "numpy" mode, it returns weird shape (samples,2), which is not like librosa.load, nor torch.load.
So, how to properly have the input.

Thanks.

m-ali-awan:
This is the colab notebook  link:

https://colab.research.google.com/drive/13fYBSQmuJpkwWVpRBbcav1pcs3gu_Dl3?usp=sharing

m-ali-awan:
You can see the Gradio part:

Thanks a lot..

m-ali-awan:
Also when using "numpy" mode:
numpy array may be as it is not normalized or due to some other reason;
It gives weird error:
  File "/usr/local/lib/python3.7/dist-packages/fastaudio/augment/signal.py", line 54, in _tfm_pad_signal
    c, x = sig.shape
ValueError: too many values to unpack (expected 2)

Kindly, help me out with this Audio Part.
Thanks a lot...

m-ali-awan:
I have got the solution:

Actually for file model:
we have to grab the path by input_file.name,
Then specifically for fastai model do path=Path(str( input_file.name)), 
And then pass it to the learner.

[...]

PR #800, opened 2022-03-09
Streaming audio

aliabid94:
Adding live mode for audio streaming, see demo/streaming_stt for an example.


abidlabs:
Closes #191

gradio/routes.py
```diff
@@ -252,10 +253,13 @@ def api_docs(request: Request):
 async def predict(body: PredictBody, username: str = Depends(get_current_user)):
     if app.launchable.stateful:
         session_hash = body.session_hash
-        state = app.state_holder.get(
-            (session_hash, "state"), app.launchable.state_default
-        )
-        body.state = state
+        if body.cleared:
+            body.state = None
```
abidlabs, line 257: Should be `body.state = app.launchable.state_default`

demo/streaming_stt/run.py
```diff
@@ -0,0 +1,43 @@
+from deepspeech import Model
```
abidlabs, line 1: Suggestion: replace this with a ASR model from `transformers`. I can do this tomorrow

abidlabs:
@aliabid94 I don't think the streaming Audio component works correctly. I tried running the following code:

```python
import gradio as gr
import time

def transcribe(speech):
    time.sleep(2)
    return speech

gr.Interface(transcribe, 
             gr.inputs.Audio(source="microphone", type="filepath"), 
             "text",  
             live=True).launch()
```

This should print the path to audio files that are all at least 2 seconds long, because the backend function takes 2 seconds to complete. But if you play any of the audio files that are printed, they only last less than a second. What I'm afraid is happening is that the frontend always sends the ~0.5 seconds of audio to the backend, even if the inference function takes longer than that to complete. Let me know if I'm missing something

abidlabs (approved):

PR #807, opened 2022-03-11
Gradio dash fe

### LINKED ISSUE #636: [Blocks] implement core architecture to support blocks (frontend)
Rough notes/ checklist that I'm working through:

- [ ] Single internal API for blocks and interfaces to reduce code duplication + maintenance (this is being worked on in the backend to make it simpler, I'lld do something temporary on the FE if i get to it first to allow this).
- [ ] Treat all components exactly the same and render them recursively, including layout components such as tabs/rows. May require changes to JSON config will patch on the FE until changes are synced.
- [ ] Figure out internal state storage mechanism. State provided by user vs state required by components vs state sent to the API. Sometimes these are the same, sometimes they aren't.
- [ ] Mechanism for listening to and attaching triggers to events.
- [ ] Ensure components are entirely decoupled form the app. They should know nothing about gradio and deal with simple inputs + outputs (via events). 
- [ ] Define a consistent even interface for simplicity (probably start with `change` but could add a `save` or `click` event). At some point it might make sense to offer complete flexibility here but we'll start with the basics.

### LINKED ISSUE #639: [Blocks] Allow any output component to be used to display statically

### LINKED ISSUE #723: [Blocks] Implement other event triggers besides `click` & support event triggers across multiple components

### LINKED ISSUE #736: [Blocks] Input/Output Component Limitation
I would like to build a workflow where the output of a model is the input of another (two interfaces in series), but I can't pass `gr.outputs.<etc>` as an input to something (`submit.click(predict, inputs=gr.outputs.<etc>, outputs=predictions)`)

This is a major limitation. Flow I want is:
 
* User submits first input
* model returns first output 
* second model returns second output using first output as input 

Note that not all outputs can be inputs

### LINKED ISSUE #786: Combine Inputs and Outputs under Components
We will Combine Inputs, Outputs and Statics under Components while supporting the old api usage with workarounds.

Related PR and discussion: #765

### LINKED ISSUE #790: Overwriting input image with output image ?
Hi, how can we show the output image at the place of input image, this means after model inference overwriting the resulted image (OUTPUT) with the input image instead of showing in parallel ?

Thanks

pngwn:
- all components are just 'components', static/ dynamic, input/output doesn't matter they can all be input/output techincally, it juist depends what functionality the component has. If it emits an event then it can be a trigger for a predict fn, if it accepts a value then it can be an output. or of course it can be both.
static/ dynamic is automatically detected based on whether or not the component is used as an input (heuristic might need tweaking)
- Render code is much simplified, we simply recursively walk the tree rendering components, layout components like tabs are treated exactly like any other so we don't need to special case certain types of components. We may need to think about how we design our component with this in mind.
- We can attach arbitrary events to arbitrary components, we do not need to limit ourselves to a small set of events across all components, we can decide component by component. This has the benefit of supporting custom component more comprehensively than before. Users could create custom events from their own components that they could then use as event triggers, etc.
- Components are now loaded on demand based on what is passed down in the config. This should speed things up considerably but also allow us to freely add new components without worrying too much

Closes #636. Closes #723. Closes #736. Closes #639. Closes #786. Closes #790.


FarukOzderim:
Just merged this branch on `Blocks-Components` to get updates from here.

PR #811, opened 2022-03-11
Adding custom css kwarg directly to a component [WIP]

aliabid94:
# Directly adding custom CSS to components

Users should be able to add css directly to components without having to know class names of elements. Adding a css kwarg to all components will make this possible in most basic cases:

e.g.

A red button:
```python
gr.Button(css={"background-color": "pink", "color": "red"})
```

A custom layout resembling the current Interface
```python
with gr.Row(css={"gap": "2rem"}):
  with gr.Column(css={
    "background-color": "whitesmoke",
    "padding": "2rem",
    "border-radius": "1rem"
  }):
    # components
    with gr.Row():
      clear = gr.Button()
      submit = gr.Button(css={"background-color": "orange"})
  # ...
```

CSS variables can also be passed in at any component level or at the top Blocks element:
```python
app = Blocks(css={"--primary-color": "red"})

with app:
   gr.Button(css={"--secondary-color": "blue"})
```

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -7,8 +7,9 @@ from gradio.routes import PredictBody
 
 
 class Block:
-    def __init__(self):
+    def __init__(self, css=None):
         self._id = Context.id
+        self.css = css
         Context.id += 1
         if Context.block is not None:
             Context.block.children.append(self)


pngwn:
I don't have a strong objection to this but one thing we need to figure out is what passing style declarations means for components hwere we can't realistically figure what what we should be applyin the style to.

For a button there is only one place where the padding can realistically appear but for complex component that numerous nested elements that can all have some kind of padding, what do we do and mroe importantly, how do we effectively communicate those subteleties to users?

My personal opinion is that this kind of interface, which is common in some front end libraries can often cause more harm than good. Partly because the solution doesn't scale particularly well (a user can't just learn a 'pattern' and apply it everywhere) and partly because it exposes implementation details.

Would it be better for components to define an explicit interface aside from the CSS variables that user could make use of? That way we aren't manually passing declaratrions down but have an explicit style API on a per component basis?

Lastly:
Would it be possible to assign a custom CSS id selector to elements in the same manner? An id would be ideal for disambiguation when embedding multiple Gradios into a single page.

pngwn (approved):

PR #822, opened 2022-03-15
New guides

aliabd:
Added two new guides:

* Adding Markdown and Other Text
* Using the API Docs

Fixes: #568


abidlabs:
The "Adding Markdown and Other Text" guide generally looks good, but some issues/suggestions:

* "Adding Markdown And Other Text" --> rename title to something clearer like "Adding Rich Descriptions to Your Demo"
* The demo you chose is a nice one to highlight, but I don't think it is working right. When I tried just putting "abc", I got all this old messages that showed up including some with profanity and some that stretched the interface so wide that it messed up the layout:

image

* Also, we shouldn't embed user demos since they might change. I would [fork this Space](https://huggingface.co/spaces/osanseviero/repo_duplicator), fix the issues mentioned previously, and embed yours instead.
* The parameters should be lower case
* The image is nice and very helpful
* The code example currently uses HTML, which makes it hard to read. Can you instead write it in Markdown? 
* You *might* want to include some common markdown commands directly in the article so it has more content. This could also be the time to mention that URLs are automatically converted to links (you don't even need to use Markdown here)

The "API Docs" is very sparse and not very useful as a guide as it stands.
* You can add more explanation around each of the sections. Perhaps describe why each section is useful and give an example?
* Maybe include a screengrab of your terminal as you call a prediction using the API
* For the example that you embed (which is a nice idea), use a Space that you've generated or one of us has generated instead of a user-submitted one so that it doesn't break

aliabid94:
LOL at the chat history

aliabd:
I made the changes @abidlabs can you take another look :)

abidlabs:
LGTM!

PR #826, opened 2022-03-16
Component label fixes

### LINKED ISSUE #825: A few small component label issues
### Describe the bug

As I was creating demos for the course, I noticed some small issues related to the labels of the components:
* The label of an input component is sometimes `null` (by default, it should be "Input")
* The default input component label ("Input") and output component label ("Output") should both be lower-cased to be consistent e.g. with parameters, which are typically lower case

### Reproduction

For one such example, run:

```python
gr.Interface.load("spaces/abidlabs/remove-bg", inputs="webcam", title="Remove your webcam background!").launch()
```

### Screenshot

image

### Logs

_No response_

### System Info

```shell
2.8.10
```

### Severity

annoyance

abidlabs:
Closes #825

### DIFF
--- a/gradio/flagging.py
+++ b/gradio/flagging.py
@@ -330,7 +330,7 @@ class HuggingFaceDatasetSaver(FlaggingCallback):
                 for i, component in enumerate(interface.input_components):
                     component_label = interface.config["input_components"][i][
                         "label"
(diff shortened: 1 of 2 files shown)


abidlabs:
Fixes #826

Issue #824, opened 2022-03-16
Interface doesn't work right


MohamedAliRashad:
### Describe the bug

After i upgraded gradio to it's latest version. The interface doesn't take the whole area of the website.

### Reproduction

```python
inputs = [gr.inputs.Image(type='pil', label="Original Image")]
outputs = [
           gr.outputs.Image(type="numpy",label="sal-1"),
           gr.outputs.Image(type="numpy",label="sal-2"),
           gr.outputs.Image(type="numpy",label="sal-3"),
           ]

title = "Salient Application Testing"
article = "<p style='text-align: center'><a href='https://arxiv.org/abs/2005.09007'>U^2-Net: Going Deeper with Nested U-Structure for Salient Object Detection</a> | <a href='https://github.com/xuebinqin/U-2-Net'>Github Repo</a></p>"

iface = gr.Interface(
    fn=infer,
    inputs=inputs,
    outputs=outputs,
    title=title,
    article=article,
    allow_screenshot=False,
    theme="dark",
)
iface.launch(share=False, server_name="0.0.0.0", server_port=8880)

```

### Screenshot

![Screenshot from 2022-03-16 16-05-26](https://user-images.githubusercontent.com/26205298/158608305-7ee0edf2-3757-403b-9c95-841f33beab8b.png)

### System Info

```shell
`gradio.__version__ = 2.8.10`
```

### Severity

annoyance

abidlabs:
Hi @MohamedAliRashad thanks for opening the issue. Can you please explain what is the issue with the output here?

aliabid94:
The dark theme doesn't come to the bottom. Will fix!

PR #828, opened 2022-03-17
Full height gradio app

aliabid94:
We had an issue with dark theme: in any demo that extended past the bottom of the page (.e.g kitchen sink), the background color did not continue past browser height. @pngwn fixed this in https://github.com/gradio-app/gradio/pull/652 by setting `height: 100%` to `min-height: 100%` for the root element. However, due to issues with how browsers implement min-height, this introduced  new issue: if the dark mode app does not reach the bottom of the page, neither does the background color. Basically `min-height: 100%` does not work if no ancestor element has an explicitly defined height (not percentage or viewport based). Switched implementation to use flex instead, should work now.

Fixes: #824


pngwn (approved):
lgtm

PR #843, opened 2022-03-21
Blocks components V2

FarukOzderim:
# Description

- refactor, default -> default_value
- refactor output types into output_type and make them auto
Regarding this

1. video is not a type but format, did not update it as I see no need for it.
2. refactored type to output_type for the related components, just updated uses in the init function. However it seems there are usages of type within processing_utils, please check whether if it is relevant.


abidlabs:
@FarukOzderim a high level comment: we should change the parameter name in `components.py`, but we should *not* change the parameter name in `inputs.py` and `outputs.py` to preserve backwards compatability

(The purpose of the `inputs` and `outputs` module is basically just backwards comptability)

gradio/components.py
```diff
@@ -235,11 +235,11 @@ def __init__(
                 "The 'type' parameter has been deprecated. Use the Number component instead if you need it.",
                 DeprecationWarning,
             )
-        default = str(default)
+        default_value = str(default_value)
         self.lines = lines
         self.placeholder = placeholder
-        self.default = default
-        self.test_input = default
+        self.default = default_value
```
abidlabs, line 241: Nit: `self.default_value = default_value`

gradio/components.py
```diff
@@ -931,15 +937,13 @@ def __init__(
     ):
         """
         Parameters:
-        default(str): IGNORED
+        default_value(str): IGNORED
```
abidlabs, line 940: This is no longer ignored. We should say in the docstring something like, "default image showed by the input component. If None, no image is shown"

gradio/components.py
```diff
@@ -931,15 +937,13 @@ def __init__(
...
         tool (str): Tools used for editing. "editor" allows a full screen editor, "select" provides a cropping and zoom tool.
-        type (str): #TODO:(Faruk) combine the descriptions below
-            input:  Type of value to be returned by component. "numpy" returns a numpy array with shape (width, height, 3) and values from 0 to 255, "pil" returns a PIL image object, "file" returns a temporary file object whose path can be retrieved by file_obj.name, "filepath" returns the path directly.
-            output: Type of value to be passed to component. "numpy" expects a numpy array with shape (width, height, 3), "pil" expects a PIL image object, "file" expects a file path to the saved image or a remote URL, "plot" expects a matplotlib.pyplot object, "auto" detects return type.
+        type (str): Input Type of value to be returned by component. "numpy" returns a numpy array with shape (width, height, 3) and values from 0 to 255, "pil" returns a PIL image object, "file" returns a temporary file object whose path can be retrieved by file_obj.name, "filepath" returns the path directly.
```
abidlabs, line 946: Slight modification: "the format the image is converted to before being passed into the prediction function. "numpy" converts the image to a numpy array with shape (width, height, 3) and values from 0 to 255, "pil" converts the image to a PIL image object, "file" produces a temporary file object whose path can be retrieved by file_obj.name, "filepath" returns the path directly."

gradio/components.py
```diff
@@ -952,6 +956,7 @@ def __init__(
             self.type = type
 
         self.type = type
+        self.output_type = "auto"
```
abidlabs, line 959: Do we need this? Can we just remove this line (see my comment below as well)

gradio/components.py
```diff
@@ -1173,7 +1178,7 @@ def postprocess(self, y):
         Returns:
         (str): base64 url data
         """
-        if self.type == "auto":
+        if self.output_type == "auto":
```
abidlabs, line 1181: If we remove `self.output_type`, we can just rewrite this to be:

          if isinstance(y, np.ndarray):
              dtype = "numpy"
          elif isinstance(y, PIL.Image.Image):
              ...

gradio/components.py
```diff
@@ -1342,13 +1347,15 @@ def __init__(
     ):
         """
         Parameters:
+        default_value (str): IGNORED
         source (str): Source of audio. "upload" creates a box where user can drop an audio file, "microphone" creates a microphone input.
-        type (str): Type of value to be returned by component. "numpy" returns a 2-set tuple with an integer sample_rate and the data numpy.array of shape (samples, 2), "file" returns a temporary file object whose path can be retrieved by file_obj.name, "filepath" returns the path directly.
+        type (str): Input type. Type of value to be returned by component. "numpy" returns a 2-set tuple with an integer sample_rate and the data numpy.array of shape (samples, 2), "file" returns a temporary file object whose path can be retrieved by file_obj.name, "filepath" returns the path directly.
```
abidlabs, line 1352: The same comments I gave in `Image` basically apply here.

gradio/components.py
```diff
@@ -1342,13 +1347,15 @@ def __init__(
...
         label (str): component name in interface.
         """
         self.source = source
         requires_permissions = source == "microphone"
         self.type = type
+        self.output_type = "auto"
```
abidlabs, line 1358: The same comments I gave in `Image` basically apply here.

gradio/external.py
```diff
@@ -49,7 +49,7 @@ def encode_to_base64(r: requests.Response) -> str:
     pipelines = {
         "audio-classification": {
             # example model: https://hf.co/ehcalabres/wav2vec2-lg-xlsr-en-speech-emotion-recognition
-            "inputs": components.Audio(label="Input", source="upload", type="filepath"),
+            "inputs": components.Audio(source="upload", type="filepath", label="Input"),
```
abidlabs, line 52: Out of curiosity, why do we need to reorder these?

abidlabs:
Left some comments above. Happy to review again after you've addressed them

abidlabs:
> @FarukOzderim a high level comment: we should change the parameter name in components.py, but we should not change the parameter name in inputs.py and outputs.py to preserve backwards compatability
> 
> (The purpose of the inputs and outputs module is basically just backwards comptability)

You shouldn't have to change any of the code for any of the demos. They should all work as is. (Separately, we will update the demos to use `components` instead of `inputs` or `outputs`, but I will create a PR for that separately)

gradio/external.py
FarukOzderim: It probably happened automatically with refactoring.

gradio/components.py
FarukOzderim: Actually it currently is ignored and not used in the component. Let me implement this in an individual commit.

FarukOzderim:
> @FarukOzderim a high level comment: we should change the parameter name in `components.py`, but we should _not_ change the parameter name in `inputs.py` and `outputs.py` to preserve backwards compatability
> 
> (The purpose of the `inputs` and `outputs` module is basically just backwards comptability)

This seems to happened because of inheritance and refactor, will fix!

gradio/components.py
FarukOzderim: I actually wanted to keep it that way as we may want to have changes in the future 
- where we could not support `auto` type in some components
- want to enable users to declare a specific output_type

gradio/components.py
FarukOzderim: Gave a reply at upper comment

demo/form_graph/run.py
```diff
@@ -28,14 +28,12 @@ def plot_forecast(final_year, companies, noise, show_legend, point_style):
     plot_forecast,
     [
         gr.inputs.Radio([2025, 2030, 2035, 2040], label="Project to:"),
-        gr.inputs.CheckboxGroup(
-            ["Google", "Microsoft", "Gradio"], label="Company Selection"
-        ),
-        gr.inputs.Slider(1, 100, label="Noise Level"),
+        gr.inputs.CheckboxGroup(["Google", "Microsoft", "Gradio"],  label="Company Selection"),
+        gr.inputs.Slider(1, label="Noise Level"),
```
abidlabs, line 32: Why did we drop the 2nd parameter in `Slider`?

demo/fraud_detector/run.py
```diff
@@ -23,10 +23,8 @@ def fraud_detector(card_activity, categories, sensitivity):
     fraud_detector,
     [
         gr.inputs.Timeseries(x="time", y=["retail", "food", "other"]),
-        gr.inputs.CheckboxGroup(
-            ["retail", "food", "other"], default=["retail", "food", "other"]
-        ),
-        gr.inputs.Slider(1, 3),
+        gr.inputs.CheckboxGroup(["retail", "food", "other"],  default=["retail", "food", "other"]),
+        gr.inputs.Slider(1),
```
abidlabs, line 27: Same, why did we drop the 2nd parameter in Slider?

abidlabs:
@FarukOzderim noticing small issues with the changes in the demos. I caught the Slider issues above, but I might miss other things. Since the demos should not change, so I have reverted the demo run files in this commit. Also makes the PR cleaner and easier to review

gradio/components.py
```diff
@@ -1974,22 +1978,21 @@ class Label(Component):
...
-        default(str): IGNORED
+        default_value(str): IGNORED
         num_top_classes (int): number of most confident classes to show.
-        type (str): Type of value to be passed to component. "value" expects a single out label, "confidences" expects a dictionary mapping labels to confidence scores, "auto" detects return type.
         label (str): component name in interface.
         """
         self.num_top_classes = num_top_classes
-        self.type = type
+        self.output_type = "auto"
+        self.type = "auto"
```
abidlabs, line 1995: We can remove this line?

gradio/inputs.py
```diff
@@ -43,17 +41,17 @@ def __init__(
             DeprecationWarning,
         )
         super().__init__(
+            default_value=default,
```
abidlabs, line 44: **Important**: in all of the components in `inputs.py` and `outputs.py`, the parameter should still be called `default` for backwards compatibility.

gradio/components.py
FarukOzderim: not sure about how to handle default image input, opened an issue and added a TODO

gradio/inputs.py
FarukOzderim: Well it is default in the inputs.py, it is super init here

FarukOzderim:
Ready to merge!

gradio/inputs.py
abidlabs: Oh you are right my bad, GitHub's code preview skipped over a few lines :)

PR #844, opened 2022-03-21
Blocks-Backend-Events

### LINKED ISSUE #723: [Blocks] Implement other event triggers besides `click` & support event triggers across multiple components

FarukOzderim:
All events are implemented except `Tabs`. `Tabs` is not a component in backend.

Closes #723

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -76,7 +76,15 @@ class Column(BlockContext):
 
 
 class Tabs(BlockContext):
-    pass
+    def change(self, fn: Callable, inputs: List[Component], outputs: List[Component]):
+        """
+        Parameters:
+            fn: Callable function
+            inputs: List of inputs
+            outputs: List of outputs
+        Returns: None
+        """
+        self.set_event_trigger("change", fn, inputs, outputs)
 
 
 class TabItem(BlockContext):
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -19,77 +19,33 @@ class Block:
             Context.root_block.blocks[self._id] = self
         self.events = []
 
-    def click(self, fn, inputs, outputs):
-        if not isinstance(inputs, list):
-            inputs = [inputs]
-        if not isinstance(outputs, list):
-            outputs = [outputs]
-        Context.root_block.fns.append(fn)
-        Context.root_block.dependencies.append(
-            {
-                "id": len(Context.root_block.dependencies),
-                "targets": [self._id],
-                "trigger": "click",
-                "inputs": [block._id for block in inputs],
-                "outputs": [block._id for block in outputs],
-            }
-        )
-
-    def change(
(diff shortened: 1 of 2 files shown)


pngwn:
@FarukOzderim I think `Tabs` _should_ be the same as everything else. It isn't any different functionally from rows, columns or carousel. Grid will probably fall under a similar category.

abidlabs:
I agree with @pngwn, tab change could be a useful event to track. @FarukOzderim since `Tab` is a `BlockContext` which is a `Block`, it should also support the `set_event_trigger()` already, right?

My only additional suggestion here is to have more descriptive docstrings for the event triggers of. Instead :

```
"""
Parameters:
    fn: function name
    inputs: List of inputs
    outputs: List of outputs
Returns: None
"""
```

```
"""
Creates a trigger to call a function when this component is clicked
Parameters:
    fn: the function to call
    inputs: the input components whose values should be used as the inputs to the function
    outputs: the output components whose values should be updated based on the function return values
Returns: None
"""
```


Otherwise, LGTM.

gradio/blocks.py
```diff
@@ -19,77 +19,33 @@ def __init__(self):
...
-                "outputs": [block._id for block in outputs],
-            }
-        )
-
-    def change(
-        self, fn: str, inputs: List["Component"], outputs: List["Component"]
+    def set_event_trigger(
+        self,
+        event_name: str,
+        fn: str,
```
abidlabs, line 25: Type of `fn` should be `Callable`

FarukOzderim:
I wanted to clarify meaning of tabs below, but now it is clarified!
https://github.com/gradio-app/gradio/issues/723#issuecomment-1073978532

FarukOzderim:
Ready to merge!

PR #848, opened 2022-03-21
Started updating demos to use the new `gradio.components` syntax

abidlabs:
* Updating all of the demos in the `demos/` folder to use the new `gradio.components` imports and syntax
* Creating new demos using `Blocks`
* Also doing some minor cleanups as I encounter issues
* As discussed with @pngwn, will add a script to quickly create a config for any particular demo from its name


gradio/__init__.py
```diff
@@ -1,9 +1,32 @@
 import pkg_resources
 
-import gradio.components as components
-import gradio.inputs as inputs
-import gradio.outputs as outputs
 from gradio.blocks import Blocks, Column, Row, TabItem, Tabs
+from gradio.components import (
```
FarukOzderim, line 4: @abidlabs are you sure about this? I think it is more proper to import components from `gr.components`. I am against importing all the components in the `init`. There are many components and a single component is not a critical element of gradio library, they need to referenced from `gr.components`

If people want to have shorter references they can import specific components via
`from gradio.components import Textbox, Json...`

abidlabs: Hmm I really think the components and layouts both should be very easily imported since they are fundamental to building blocks. For me, `gr.Textbox` is a lot friendlier than `gr.components.Textbox`. 

There are basically 3 ways to make it friendly that I considered.

1. Doing `from gradio.components import Textbox, Image...` - the problem with this is that you have to explicitly import every component, which is annoying. Plus some of the components have common names (like `Image`), which could conflict with other classes in other libraries (like `from PIL import Image`)
2. Doing `from gradio.components import *` - bad practice, plus the same conflict issue
3. Making the components accessible as `gr.Textbox`, etc. I think this is way to go. The only potential problem is that if we have conflicts within the `gr.` namespace, which is unlikely.

gradio/__init__.py
```diff
@@ -1,9 +1,32 @@
 import pkg_resources
 
-import gradio.components as components
-import gradio.inputs as inputs
```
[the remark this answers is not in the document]
abidlabs: To be honest, I don't understand the effects of these lines. I thought importing a class in `__init__.py` just makes it accessible under `gradio.` namespace. But `inputs` and `components` are already accessible as `gradio.inputs` and `gradio.components`? 

But if it is causing issues in your IDE, I can bring them back

gradio/__init__.py
FarukOzderim: Well I still think the same, might hear thoughts from @aliabid94 and @pngwn if it will help.

gradio/__init__.py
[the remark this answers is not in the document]
abidlabs: Sounds good

gradio/__init__.py
aliabid94: We need to be able to support `gr.Checkbox()` or `gr.Textbox()`. It is way too much extra typing to have to include gr.components every single time. This is how streamlit does it as well. There are tradeoffs to be made between user-friendliness and proper explicitness in building APIs and gradio definitely needs to lean towards user-friendliness.

abidlabs:
Note: I've changed the non-keyword argument for the `CheckboxGroup`, `Radio`, and `Dropdown` components to be the _choices_ instead of the default selected values. We should not require using a keyword to create a simple radio or dropdown or checkbox group. Now, this works as expected:

```python
gr.Radio(["add", "subtract", "multiply", "divide"])
```

aliabid94:
LGTM. What's the purpose of write_config.py?

abidlabs:
> LGTM. What's the purpose of write_config.py?

Thanks. @pngwn needs it for setting up integration tests

PR #849, opened 2022-03-22
Interfaces from Blocks

aliabid94:
# Making Interfaces from Blocks. 

Interface is implemented as a Blocks config. Frontend will use blocks only. Most Interface features are implemented, try with demo/calculator.py for example.


FarukOzderim:
You might want to have destination branch as [Blocks-Components-v2](https://github.com/gradio-app/gradio/tree/Blocks-Components-v2) to show less diff, but let's merge it to Blocks-Dev when we are going to merge. It will have cleaner PR change history that way.

demo/xray_blocks/run.py
```diff
@@ -8,35 +8,38 @@
...
-                xray_scan = gr.inputs.Image()
-                xray_results = gr.outputs.JSON()
-                output_textbox = gr.outputs.Textbox()
-                input_textbox = gr.inputs.Textbox(default_value="Hello This Is a Input Textbox")
-            xray_run = gr.Button("Run")
+                xray_scan = gr.components.Image()
+                xray_results = gr.components.JSON()
+                output_textbox = gr.components.Textbox()
+                input_textbox = gr.components.Textbox(default_value="Hello This Is a Input Textbox")
+            xray_run = gr.Button("Run", css={
```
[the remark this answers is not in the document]
FarukOzderim: It seems you made the changes with a single commit 😞, it is a good practice to have singular changes in a commit, and giving commit messages meaningful descriptions 😄.
Why don't you move Interfaces as blocks parts to another branch and target there with this PR?

gradio/components.py
```diff
@@ -2343,13 +2393,63 @@ def postprocess(self, y):
...
+    def __init__(
+        self,
+        default_value: str = "",
+        *,
+        label: Optional[str] = None,
+        css: Dict = {},
+        **kwargs,
+    ):
+        super().__init__(label=label, css=css, **kwargs)
+        self.value = default_value
```
FarukOzderim, line 2425: These changes do not seem to be related to this PR as well, better have these kind of changes in another PR or commit, in the future.

gradio/interface.py
```diff
@@ -569,6 +643,20 @@ def process_api(self, data: Dict[str, Any], username: str = None) -> Dict[str, A
             "flag_index": flag_index,
         }
 
+    def process_api(self, data: Dict[str, Any], username: str = None) -> Dict[str, Any]:
+        raw_input = data["data"]
+        fn_index = data["fn_index"]
+        if fn_index == 0:  # submit
```
FarukOzderim, line 649: can make use of enums here

gradio/interface.py
```diff
@@ -569,6 +643,20 @@ def process_api(self, data: Dict[str, Any], username: str = None) -> Dict[str, A
...
+            prediction, durations = self.process(raw_input)
+            return {"data": prediction}
+        elif fn_index == 1:  # clear
+            return {
+                "data": [None]
+                * (len(self.input_components) + len(self.output_components))
+            }
+        elif fn_index == 2: # flag
```
FarukOzderim, line 657: Is flag not supported yet?

gradio/interface.py
```diff
@@ -476,7 +477,80 @@ def __repr__(self):
...
+            id = len(components) + 1
+            components.append(
+                {
+                    "id": len(components) + 1,
+                    "type": component.__class__.__name__.lower(),
+                    "props": component.get_template_context(),
+                }
+            )
+            layout_context = {"id": id}
+            if isinstance(component, BlockContext):
```
FarukOzderim, line 494: Why is there a check of `BlockContext`?

gradio/interface.py
```diff
@@ -476,7 +477,80 @@ def __repr__(self):
...
+
+        if self.title:
+            add_component(layout, Markdown("<h1>" + self.title + "</h1>"))
+        if self.description:
+            add_component(layout, Markdown(self.description))
+        panel_row = add_component(layout, Row())
+        input_panel = add_component(panel_row, Column())
+        input_ids = []
+        for component in self.input_components:
+            input_id = add_component(input_panel, component)
```
FarukOzderim, line 507: better have this for clarity
input_id = add_component(input_panel, component)["id]"

FarukOzderim: similar thing for output_id as well

gradio/interface.py
```diff
@@ -476,7 +477,80 @@ def __repr__(self):
...
+                "outputs": input_ids + output_ids,
+            }
+        )
+        dependencies.append(
+            {
+                "id": 2,
+                "trigger": "click",
+                "targets": [flag_btn["id"]],
+                "inputs": input_ids + output_ids,
+                "outputs": [],
```
FarukOzderim, line 544: how will flag button work? completely in frontend?

gradio/interface.py
aliabid94: to add the "children" field to the config. BlockContexts have children, other Blocks don't.

gradio/interface.py
aliabid94: Still needs to be implemented in backend, but now handled from the /api/predict call, need to connect flagging handler to it.

gradio/interface.py
aliabid94: correct

abidlabs:
Btw @aliabid94 I tried testing this with a simple `Interface`, and it's not working for me:

```python
gr.Interface(lambda x:x, gr.components.Textbox(), gr.components.Textbox()).launch()
```

The clear button works, but the submit button has no effect (and throws no error except printing "boo" in the console):

image

aliabid94:
weird I'll take a look

aliabid94:
The textbox component seems to not be sending the data on submit. Seems like a problem with the frontend textbox component not related to this PR, @pngwn can you take a look? @abidlabs try a demo with any other component.

abidlabs:
Ok yeah it looks like the other components are working fine

abidlabs:
In that case, everything else LGTM. Should we go ahead and merge this, so that we can get work off of this for #848?

We can handle the textbox issue (and other issues that we find) in separate, smaller PRs

PR #866, opened 2022-03-24
add test for output to input

### LINKED ISSUE #790: Overwriting input image with output image ?

pngwn:
Closes #790.

This already worked, I've just added a browser test to confirm. I can't really write a good test for images so i used text instead but it applies to all components.


abidlabs:
LGTM!

PR #869, opened 2022-03-24
fix default_value 

pngwn:
A feww issue with this one:

- The value that is passed via the config is called default not default_value could we change this to match? default is a keyword in javascript which makes using it tricky.
- Not every component seems to pass down the default value, we need to add this behaviour to everything and make sure it is as consistent as possible across components. HTML and JSON and Timeseries don't seem to pass down a default (unless I'm adding the default value incorrectly).


abidlabs:
> Not every component seems to pass down the default value, we need to add this behaviour to everything and make sure it is as consistent as possible across components. HTML and JSON and Timeseries don't seem to pass down a default (unless I'm adding the default value incorrectly).

This should be fixed in #871. @FarukOzderim has added default value to the config of all of the components.

>  The value that is passed via the config is called default not default_value could we change this to match? default is a keyword in javascript which makes using it tricky.

This has also been fixed in the same PR #871. So if I'm reading your js code correctly in this PR, this means that this PR needs to update the property name from `default` to `default_value`? @aliabid94 if you can also take a look to confirm (and fix, if needed), that would be great so that we can components fully working statically.

aliabid94:
I've renamed default to default_value in the frontend and backend components in this PR.

FarukOzderim:
> I've renamed default to default_value in the frontend and backend components in this PR.

@aliabid94 we actually did backend part in this PR, did you see it?
https://github.com/gradio-app/gradio/pull/871

abidlabs:
Yeah there was some unnecessary duplication in this PR, but fortunately no
merge conflict!

PR #871, opened 2022-03-24
backend_default_value_refactoring

FarukOzderim:
- refactor default to default_value


abidlabs:
Made some minor tweaks, but otherwise, this LGTM. @aliabid94 can you also take a look to confirm?

PR #872, opened 2022-03-24
[Spaces] Cleaner hf.space URLs

julien-c:
The `/gradioiframe/...` URLs were never really meant to be user facing so we switched to a new URL path on the `hf.space` proxy

#### Context: 
URLs are now slightly cleaner-looking, also there was not really anything Gradio specific on Spaces side for this URL, so the new path is more logical. 

Finally moving from huggingface.co to hf.space for this will be more future-proof for when the new Spaces infra is up (cc @cbensimon @christophe-rannou @pierrci)

#### Note:

we redirect the old links to the new so existing links that could live "in the wild" should still work for quite some time. (cc @osanseviero)


abidlabs:
Looks great, thanks for making the updates @julien-c!

Issue #887, opened 2022-03-28
[`blocks-dev`] Some components throw JS error


abidlabs:
### Describe the bug

When using the following components, they throw a javascript error so nothing shows up:
* `HighlightedText` 
* `Carousel`

### Reproduction

Run the following demos for each of the above components, respectively:

`demo/diff_texts/run.py`
`demo/disease_report/run.py`

It might also be worth testing the kitchen sink demo to make sure there aren't other components that have the same problem.

### Logs

```shell
index.84342a61.js:3 
        
       Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'length')
    at R (index.84342a61.js:3:3273)
    at Yr (vendor.906e418a.js:1:6936)
    at new U (index.84342a61.js:3:3624)
    at V (index.84342a61.js:3:3716)
    at Yr (vendor.906e418a.js:1:7136)
    at new X (index.84342a61.js:3:4447)
    at Bl (index.9cf03fca.js:8:5868)
    at Yr (vendor.906e418a.js:1:7136)
    at new st (index.9cf03fca.js:8:7419)
    at tt (index.9cf03fca.js:8:4729)
```

### System Info

```shell
blocks-dev
```

### Severity

blocker

abidlabs:
Closed via #901

PR #901, opened 2022-03-29
Interface fix

pngwn:
Some small tweaks to fix a few errors i came across with interfaces.

### DIFF
--- a/ui/packages/app/src/Render.svelte
+++ b/ui/packages/app/src/Render.svelte
@@ -42,7 +42,7 @@
 	{theme}
 	{root}
 >
-	{#if children.length}
+	{#if children && children.length}
 		{#each children as { component, id, props, children, has_modes }}
 			<svelte:self
 				{component}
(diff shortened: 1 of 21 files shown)


abidlabs:
Thanks @pngwn! What errors does this PR fix? That way, I can test them and link potentially related issues

pngwn:
It was runtime errors when running the app. A few of those `cannnot get length of undefined` issues. It was just related to components that don't have children. I also fixed the `HighlightedText` one was a regression (named colours working as part of the color_map #724 ), I think the other was #887. The carousel needs some other work, including on the backend.


The `default_value` changes were a pre-emptive fix because changing the `value` prop to `default_value` on the frontend will break event triggers.

pngwn:
In terms of testing, the kitchen_sink should work (apart from the carousel although it shouldn't error).

abidlabs:
Tested and LGTM!

Issue #906, opened 2022-03-30
Live mode doesn't automatically generate output for examples


NimaBoscarino:
### Describe the bug

When I have some examples and have set `live=True` for the interface, clicking on the examples doesn't automatically run the function and generate the output. Using inputs, such as sliders, does trigger the live reload properly.

### Reproduction

Here's a Hugging Face space with the described behaviour: https://huggingface.co/spaces/NimaBoscarino/kornia-image-enhancement

### Screenshot

https://www.loom.com/share/3fdf4588509e4cf5b005feeae8748f94

### System Info

```shell
Gradio version 2.9.0
```

### Severity

annoyance

abidlabs:
Thanks, we'll fix this!

abidlabs:
This should be working now in the latest beta release of Gradio (`gradio==2.9.0b2`)

![image](https://user-images.githubusercontent.com/1778297/164075083-0ab94bc5-d302-4777-b71a-2f5999d3ee5c.png)

Thanks for creating the issue @NimaBoscarino!

PR #915, opened 2022-03-31
Restore Interpretation, Live, Auth, Queueing

aliabid94:
Interpretation, Auth, Live restored in format of blocks. Also included:
- Setting False to a Row or Column hides that panel of components.

Note: 
- There is currently no "blocker" from submitting while a request is in progress. Also applies to changes in live mode.
- Currently no loading animation on submit, or updates when queued.
- @pngwn some issues:
  - for some components, on:change events seem to fire with the "pre-change" value of the component. For example, with the Radio element in calculator_live/run.py, if you change from 'add' to 'subtract', it will trigger the on:change but use the value 'add'.  
  - with debouncing, it is possible that I type in an input field and click submit very quickly, and the submission does not take in the latest value of input field.


abidlabs:
Fixed the formatting, but frontend "functional-tests" are failing @aliabid94

abidlabs:
Would be good to close #906 as part of this if possible

abidlabs:
Generally looks good to me. However, there are some issues that I noticed:
* Clear button does not work when `live=True`
* Clear button does not work when `interpet="default"` 
* In general, the clear button seems to be a bit buggy actually -- doesn't work for image examples like `demo\image_classifier\run.py` for example
* Interpretation doesn't seem to work for textbox. See `demo\gender_sentence_default_interpretation\run.py`
* Multiple tests are not passing because presumably we changed the config -- pls take a look

aliabid94:
Fixed above issues

Issue #690, opened 2022-02-21
🙋 Community Feedback on Blocks


abidlabs:
#### What is Blocks?

Blocks allows you to build web-based demos in a flexible way using the `gradio` library. Blocks is a more low-level and flexible alternative to the core `Interface` class. Specifically, Blocks support:
* grouping together related demos using Tabs
* flexible positioning of components using Columns, Rows, etc. and the ability to add explanatory Markdown or Text at any point
* flexible data flows (e.g. any number of input and output components can be displayed on screen, and functions can define data flow between any subset of inputs and outputs)
* flexible event triggers (e.g. button clicks, element changes, etc.)

Here's an example Block:

![Recording 2022-02-22 at 11 54 03](https://user-images.githubusercontent.com/1778297/155180109-d84ecc00-b3f8-4282-b103-4918aac931ad.gif)
 
If you've used Gradio before, you'll find that Blocks still allows you to build web apps:
* entirely in Python
* using the standard Python interpreter, which means it will run in Jupyter notebooks, colab notebooks, or any Python IDE

#### Getting Started with Blocks
* Blocks is just released with Gradio 3.0. Check here for the announcement. 
* To use it, please install the latest release of Gradio: `pip install gradio`
* TODO: Add [Blocks Introduction Guide Link]

You can also use Blocks in Spaces! See an example [here](https://huggingface.co/spaces/aliabd/blocks-image-audio).

----------

#### *Giving feedback*: In the comments below, we'd love to hear from you:

* Did Blocks allow you to build something cool? Share a screenshot or a link to a Blocks Demo!
* Does Blocks *not* support your use case? Let us know too so that we can support it!
* Did you find anything confusing or hard within the syntax or usage of Blocks? Let's improve it together!

nateraw:
Really looking forward to this feature! In colab, I'm running into an issue. [Here's a notebook.](https://gist.github.com/nateraw/147d1bd360263022ac3639ddd961fec0)

`AttributeError: 'Blocks' object has no attribute 'requires_permissions'`

Looks like some env variable I'm supposed to set/isn't set properly?

abidlabs:
Hi folks! I just built a `Block` to plot projectile trajectories when a user specifies the speed and angle of the projectile. 

**Screenshot**

![image](https://user-images.githubusercontent.com/1778297/155348930-bbe7977b-f6d0-4261-9772-34bbe6367067.png)

**Code**

```python
import gradio as gr
[...]

PR #919, opened 2022-04-02
Allow `Blocks` instances to be used like a `Block` in other `Blocks`

abidlabs:
This PR does a number of things:

- Handle Markdown better: cleans up indentation from markdown blocks so that if every line is indented a certain amount, that is removed uniformly. Before this, any indented Markdown would show up as codeblocks (try running `demo/xray_blocks/run.py` to see the difference)
- Allow running multiple Gradio servers from one Python session / jupyter notebook / colab notebook
- Allow using Blocks/Interfaces in other Blocks:

So that we can do things like this:

```python
with gr.Blocks() as bl:
    with gr.Tabs():
        with gr.TabItem("Text"):
            gr.Interface(lambda x:x, "textbox", "textbox")
        with gr.TabItem("Image"):
            gr.Interface(lambda x:x, "image", "image")
    bl.launch()
```

- Creates the `TabbedInterface` abstraction to make the above easier. Just a rough draft, feedback on naming and API welcome here

```python
io = gr.Interface(lambda x:x, "textbox", "textbox")
io2 = gr.Interface(lambda x:x, "image", "image")            

t = gr.TabbedInterface([io, io2])
t.launch()
```


gradio/interface.py
```diff
@@ -773,6 +769,39 @@ def integrate(self, comet_ml=None, wandb=None, mlflow=None) -> None:
...
+                            num_shap=interface.num_shap,
+                            title=interface.title,
+                            description=interface.description,
+                            article=interface.article,
+                            allow_flagging=interface.allow_flagging,
+                            flagging_options=interface.flagging_options,
+                            flagging_dir=interface.flagging_dir,
+                        )
+
+
```
aliabid94, line 804: This seems not the right way to do this. Why can't you just call interface.render() instead of creating a new Interface? 
Also I'm a little skeptical we should even create TabbedInterface, instead of allowing users to naturally explore Blocks as they want more customizability.

gradio/blocks.py
```diff
@@ -12,7 +12,9 @@
 
 
 class Block:
-    def __init__(self):
+    def __init__(self, without_rendering=False):
```
aliabid94, line 15: is it possible that instead of using "without_rendering", it just checks to see if its within a BlockContext and only calls render if that's the case?

aliabid94 (changes requested):

gradio/interface.py
abidlabs: > Why can't you just call interface.render()

There's a few Interface kwargs that don't make sense to have within this context. For example, `theme`. So I just decided to pick the kwargs that make sense to include here.

> Also I'm a little skeptical we should even create TabbedInterface, instead of allowing users to naturally explore Blocks as they want more customizability

Open to discussion about this. We should keep in mind that we've heard directly from several users that they want something like this rather than having to delve into the lower-level Blocks API. E.g. https://github.com/gradio-app/gradio/issues/690#issuecomment-1050115607

gradio/blocks.py
abidlabs: That doesn't work. Consider the following example:

```python
with gr.Blocks() as bl:
    with gr.Tabs():
        with gr.TabItem("Text"):
            gr.Interface(lambda x:x, "textbox", "textbox")
        with gr.TabItem("Image"):
            gr.Interface(lambda x:x, "image", "image")
    bl.launch()
```

Inside the `Interface` class, you create `Component` objects once, and then render them later. 

Create components: https://github.com/gradio-app/gradio/blob/5c44ae353672026984df7cdb9157ff9fc6cb1d4a/gradio/interface.py#L182

Render: https://github.com/gradio-app/gradio/blob/5c44ae353672026984df7cdb9157ff9fc6cb1d4a/gradio/interface.py#L496

In this case, both statements are inside a `BlockContext` (the outer level `Blocks` `bl`), so the component would end up getting rendered twice. That's why we need to explicitly escape the context.

Issue #450, opened 2022-01-05
[Blocks] Need for multiple separate interfaces


ejohb:
**Is your feature request related to a problem? Please describe.**

I need to host a model that takes one of two inputs (e.g. text _or_ an image, but not both). Because the inputs are so different, they each need their own data types, outputs, examples, pull-downs, etc.

I don't seem to be able to achieve this in a non-hacky way. I've tried using `gr.Parallel` but that requires unified input types. 

Plus using a single, two-input interface/function means all the examples and their arguments, etc. get run together, one of the outputs is always empty, and it's just confusing to use.

**Describe the solution you'd like**

Ideally there would be some kind of `gr.Stack` or `gr.Sequence` container object that would simply stack two separate interfaces, or display them in a tabbed interface, etc.

abidlabs:
Hi @ejohb thanks for creating this issue. Your use case makes a lot of sense. We do have "stacked" interfaces on our roadmap towards the end of January, so we'll keep you updated on this. 

In the meantime, the best workaround I would suggest is to have two inputs to your function, an image and a textbox, but for the `gr.inputs.Image`, set the `optional` parameter to `True`. Then, in your function, you can check if the image parameter is None, or if the textbox is empty, and handle the two cases accordingly 

(cc @aliabd)

ejohb:
> Hi @ejohb thanks for creating this issue. Your use case makes a lot of sense. We do have "stacked" interfaces on our roadmap towards the end of January, so we'll keep you updated on this.
> 
> In the meantime, the best workaround I would suggest is to have two inputs to your function, an image and a textbox, but for the `gr.inputs.Image`, set the `optional` parameter to `True`. Then, in your function, you can check if the image parameter is None, or if the textbox is empty, and handle the two cases accordingly
> 
> (cc @aliabd)

Wonderful. Thanks very much for your reply.

abidlabs:
I'm actually going to keep this open until we ship the feature :)

JuanCarlosMartinezSevilla:
Hi everyone! 

I'm currently trying to use Gradio in a project and I'm facing the same problem, I can't launch different interfaces in the same page, without reloading with F5. 

I guess for me the solution will be to add something like an event (on_click) and depending the component clicked launch the wanted interface, inside the same page. Or something like the "live" feature that updates the output when a change is made but with the interfaces visibility.

[...]

PR #924, opened 2022-04-05
`TabbedInterface` Class

abidlabs:
This PR creates higher-level abstractions using the `Blocks` API to cover common use cases that we have heard from users. I'd like to propose that we create the following abstractions to start (feedback 100% welcome):

- `TabbedInterface(interface_list: List[Interface], tab_names: Optional[List[str]])`: user passes in a list of interfaces, and then we create a tabbed interface in which they can navigate through the different interfaces and run them individually. (this one has already been implemented, but ofc we can change the name/parameters/etc, and we need to add tests/demos)

Addresses: https://github.com/gradio-app/gradio/issues/690#issuecomment-1050115607 #450 #286 

image

- `MultiStepInterface(funciton_list: List[Callable], components_list: List[Component])`: takes in a list of functions, as well as a list of components (the length of the latter should be one more than the length of the former), and then creates a demo that can be run sequentially, so that the first component is the input to the first function, which outputs to the second component, which is the input to the second function, etc.
 
- `UnifiedInterface(fn: Callable, component: Component | List[Component])`: designed for functions in which the input and output are the same, it creates a single full page-width component (or list of components) and then running the Interface simply updates the component. 

Should we create all of these? Some of these? None of these? Other abstractions?


FarukOzderim:
The only abstraction I support is `TabbedInterface`, as others are not very generalistic and might be counterproductive to have them. I think it would be better if they are written with the Blocks API, as we do want to make it common as well :D.

abidlabs:
> others are not very generalistic

Yeah, as I was writing out `MultiStepInterface`, I got the same sense actually.. Are there other abstractions that we should support?

FarukOzderim:
Not any I can think of right now, lets open an issue if there is any in the future.

gradio/interface.py
```diff
@@ -769,6 +788,20 @@ def integrate(self, comet_ml=None, wandb=None, mlflow=None) -> None:
...
+        self, interface_list: List[Interface], tab_names: Optional[List[str]] = None
+    ):
+        if tab_names is None:
+            tab_names = ["Tab {}".format(i) for i in range(len(interface_list))]
+        super().__init__()
+        with self:
+            with Tabs():
+                for (interface, tab_name) in zip(interface_list, tab_names):
+                    with TabItem(label=tab_name):
+                        interface.render_basic_interface()
```
FarukOzderim, line 802: We could do `Interface.__init__(interface)` as well right @aliabid94?

abidlabs: Didn't work for me: `TypeError: __init__() missing 1 required positional argument: 'fn'`

FarukOzderim: Oh it actually should be `Block.__init__ ` for init to register in Context.

```
                    for component in self.input_components:
                        Block.__init__(component)
```

abidlabs: No error with `Block.__init__`, but the Interfaces inside the tabs don't show up:

image

So let's just stick to `render_basic_interface()`

gradio/interface.py
```diff
@@ -769,6 +788,20 @@ def integrate(self, comet_ml=None, wandb=None, mlflow=None) -> None:
             utils.integration_analytics(data)
 
 
+class TabbedInterface(Blocks):
+    def __init__(
```
aliabid94, line 792: Instead can the API be tuples of (interface: Interface, tabname: string)? Would prefer that over parallel arrays

gradio/interface.py
aliabid94: why can't we run Interface.render() using the Blocks render method()

aliabid94 (approved):

gradio/interface.py
abidlabs: It doesn't render the component either. Let's leave it like this right now since it doesn't cause any issues and gives us control to make sure only the `Interface` kwargs that are relevant inside a `TabbedInterface` are included

gradio/interface.py
abidlabs: Ok I have no strong opinion, so I'll make the change

abidlabs: Actually wait, it's more the way I have it because tab names can be optional

abidlabs: Unless you have a strong reason for the other way, this seems more convenient for users.

FarukOzderim (approved):
Good work!

gradio/utils.py
```diff
@@ -286,3 +287,45 @@ def get_default_args(func: Callable) -> Dict[str, Any]:
         v.default if v.default is not inspect.Parameter.empty else None
         for v in signature.parameters.values()
     ]
+
+
+def assert_configs_are_equivalent_besides_ids(config1, config2):
```
FarukOzderim, line 292: beautiful addition!

PR #929, opened 2022-04-06
static mode for textbox

### LINKED ISSUE #704: [Blocks] Component labels don't show in Blocks
Is this intentional? It makes hard to designate what each input and output is for. For example, I'd like to say that one slider is for speed and the other for angle:

image

Code to reproduce:

```python
import gradio as gr
import numpy as np
import random
import matplotlib.pyplot as plt

def generate(s, a):
    x = np.arange(-100,100,1)
    y = -x**2
    plt.plot(x, y)
    return plt

block = gr.Blocks()

with block:
    gr.static.Markdown("Let's do some kinematics. Plot the Trajectory of a Parabola based on its speed and angle!")

    with gr.Row():
        speed = gr.inputs.Slider(label="Speed")
        angle = gr.inputs.Slider(label="Angle")
    
    btn = gr.Button(label="Run")
    btn.click(generate, [speed, angle], output)
    
block.launch()
```

### LINKED ISSUE #881: [`blocks-dev`] Default value of Checkbox does not show up

### LINKED ISSUE #897: [`blocks-dev`] Problems with Textbox input and output
### Describe the bug

I was running this demo: `demo\gpt_j\run.py` and noticed some issues:
* The input textbox size does not get set by the `lines` parameter
* The output textbox size does not dynamically change size according the output (see screenshot)
* The examples don't show up at all (although clicking on them does work interestingly)

image

### Reproduction

`demo\gpt_j\run.py`

### Screenshot

_No response_

### Logs

_No response_

### System Info

```shell
blocks-dev
```

### Severity

annoyance

pngwn:
- Add static mode for textbox + Dropdown. 
- Ensure Checkbox group respects default (initial) values.
- Ensure Checkbox respect default (initial values).
- Remove duplicate title from Checkbox.

Closes #704 (ish). Closes #897. Closes #881.


ui/packages/app/src/components/Checkbox/Checkbox.svelte
```diff
@@ -8,6 +8,8 @@
 	export let mode: "static" | "dynamic";
 
 	if (default_value) value = default_value;
+
+	$: console.log($$props);
```
pngwn, line 12: remove

ui/packages/form/src/StaticTextbox.svelte
```diff
@@ -0,0 +1,68 @@
+<script lang="ts">
```
pngwn, line 1: delete file

pngwn:
Was that the behaviour previously? I thought the `lines` parameter controlled that?

Are you saying the the textbox height should scale to the content dynamically, as the user types?

abidlabs:
> Was that the behaviour previously? I thought the lines parameter controlled that?

It used to be that `inputs.Textbox` had a `lines` parameter, but the number of lines in the `outputs.Textbox` was set automatically based on the content.

> Are you saying the the textbox height should scale to the content dynamically, as the user types?

I'm wondering whether we can easily recreate the original behavior for the sake of backwards compatibility. The intended behavior would that for "static" textboxes, the number of lines is determined by the content.

If this is not possible or annoying to do because we have merged inputs and outputs, then that's fine, I don't think it's a huge deal

pngwn:
@abidlabs SHould all be adressed now.

- `lines` is respected for input components
- `output` and `static` components will update their size based on the content.

This isn't possible to test with interfaces atm (because every interface output is dynamic) but I've created a `blocks_inputs` demo you can test with.

abidlabs:
LGTM!

abidlabs (approved):

FarukOzderim:
@abidlabs @pngwn I think having input textbox lines dynamically changing would be great as well. It would make it much more less cumbersome, WDYT?

pngwn:
Guess it depends how picky users are about how the component is laid out. We could use the autoheight logic when a `lines` value isn't provided?

FarukOzderim:
> Guess it depends how picky users are about how the component is laid out. We could use the autoheight logic when a `lines` value isn't provided?

Yeah I thought the same thing at first, then later thought what about respecting the line size initially and let the line size increase when input does not fit in.
Btw the default `lines` value is 1 currently, but we could change it.

PR #930, opened 2022-04-06
allow audio components to take a string value

### LINKED ISSUE #896: [`blocks-dev`] Audio output doesn't seem to play audio
### Describe the bug

I'm running a demo which produces an audio output (`demo\generate_tone\run.py`), but all I see is a grayed audio player (see screenshot).

image

I've verified that the Interface returns a valid base64 audio file, so the problem must be in the Audio component itself

### Reproduction

`demo\generate_tone\run.py`

### Screenshot

_No response_

### Logs

_No response_

### System Info

```shell
`blocks-dev`
```

### Severity

critical

pngwn:
Audio previosuly expect the value to be an object rather thana string. Audio can now take a string.

Closes #896.

### DIFF
--- a/ui/packages/app/src/components/Audio/Audio.svelte
+++ b/ui/packages/app/src/components/Audio/Audio.svelte
@@ -1,33 +1,39 @@
 <script lang="ts">
 	import { Audio } from "@gradio/audio";
 	import type { FileData } from "@gradio/upload";
+	import { normalise_file } from "@gradio/upload";
 
 	export let mode: "static" | "dynamic";
 	export let value: null | FileData | string = null;
-	export let theme: string;
+	export let default_value: null | FileData | string = null;
 	export let style: string | null;
 	export let name: string;
 	export let source: "microphone" | "upload";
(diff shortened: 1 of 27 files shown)


abidlabs:
The demos work, but I'm still getting issues with the examples that I provided later in the issue. For example

```python
bl = gr.Blocks()

with bl:
    gr.Audio("test.mp3")
    bl.launch()
```

produces a blank webpage (nothing is shown), instead of an embedded audio file that can be played:

image

```python
bl = gr.Blocks()

with bl:
    gr.Video("test.mp4")
    bl.launch()
```

produces an embedded video, but it cannot be played:

image

Attaching the test files so you can reproduce. [test.zip](https://github.com/gradio-app/gradio/files/8429059/test.zip)

pngwn:
There are some inconsistencies with how file data is passed around, I'm going to address that more comprehensively to avoid these issues (that is the source of all of these issues).

abidlabs:
Yayy! Everything works LGTM.

abidlabs (approved):

PR #931, opened 2022-04-06
fix file upload text

### LINKED ISSUE #900: [`blocks-dev`] file input is broken
### Describe the bug

Tried running these two demos:
* `demo\zip_two_files\run.py`
* `demo\zip_to_json\run.py`

But they were unrunnable because the file input is broken:
* Before you select the file, it says "drop an audio file" instead of "drop a file"
* After you drop the file, the file takes over the whole screen making you unable to click submit (see screenshot)

### Reproduction

* `demo\zip_two_files\run.py`
* `demo\zip_to_json\run.py`

### Screenshot

image

### Logs

_No response_

### System Info

```shell
blocks-dev
```

### Severity

blocker

pngwn:
Closes #900

### DIFF
--- a/ui/packages/file/src/FileUpload.svelte
+++ b/ui/packages/file/src/FileUpload.svelte
@@ -1,5 +1,5 @@
 <script lang="ts">
-	import { createEventDispatcher } from "svelte";
+	import { createEventDispatcher, tick } from "svelte";
 	import { Upload, ModifyUpload } from "@gradio/upload";
 	import type { FileData } from "@gradio/upload";
 	import { Block, BlockLabel } from "@gradio/atoms";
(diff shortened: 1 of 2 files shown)


pngwn:
Not sure if this is the only thing that needs fixing, everything else seemed to work.

abidlabs:
(The other demo, `demo\zip_to_json\run.py` works great)

pngwn:
@abidlabs I have cleaned up the file display a little (since it looked horrific) and added an explicit download link. This won't be final until victor has worked his magic but makes things a little more usabl for now. I've also added the cursor for upload inputs.

Screenshot 2022-04-07 at 19 02 21

abidlabs:
✅

abidlabs (approved):

PR #932, opened 2022-04-06
remove debounce from form inputs

### LINKED ISSUE #863: [`blocks-dev`] Gradio Interface is unresponsive for the first ~1 sec after page loads
### Describe the bug

I'm seeing a weird bug in which, if you click the "submit" button within the first second or so after a webpage with a Gradio Interface is loaded, nothing happens. You don't see any output. There are no errors either. Perhaps the event triggers take some time to attach? But it makes the user think that the Interface doesn't work.

### Reproduction

Run a simple interface like this and quickly type something in to see if anything runs

```py
import gradio as gr

gr.Interface(lambda x:x, "text", "text").launch()
```

### Screenshot

_No response_

### Logs

```shell
No error logs in Python or JS
```

### System Info

```shell
blocks-dev
```

### Severity

annoyance

### LINKED ISSUE #923: [blocks-dev] Some issues with new components
### Describe the bug

- for some components, on:change events seem to fire with the "pre-change" value of the component. For example, with the Radio element in calculator_live/run.py, if you change from 'add' to 'subtract', it will trigger the on:change but use the value 'add'.
- with debouncing, it is possible that I type in an input field and click submit very quickly, and the submission does not take in the latest value of input field.

### Reproduction

Run demo/calculator_live/run.py to reproduce on:change issue with Dropdowns
Run demo/hello_world/run.py to reproduce debouncing issue

### Screenshot

_No response_

### Logs

_No response_

### System Info

```shell
latest blocks-dev
```

### Severity

critical

pngwn:
Closes #923. Closes #863.

@abidlabs @aliabid94  I wasn't able to fully reproduce both of those issues but I _think_ this is the right fix. If you could test and confirm, I'd appreciate it.

### DIFF
--- a/gradio/templates/frontend/index.html
+++ b/gradio/templates/frontend/index.html
@@ -45,8 +45,8 @@
 		</script>
 		<script src="https://cdnjs.cloudflare.com/ajax/libs/iframe-resizer/4.3.1/iframeResizer.contentWindow.min.js"></script>
 		<title>Gradio</title>
-		<script type="module" crossorigin src="./assets/index.18732393.js"></script>
-		<link rel="stylesheet" href="./assets/index.ff8c1728.css">
+		<script type="module" crossorigin src="./assets/index.54d43414.js"></script>
+		<link rel="stylesheet" href="./assets/index.e209fdc1.css">
 	</head>
 
 	<body style="height: 100%; margin: 0; padding: 0">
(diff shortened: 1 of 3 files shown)


abidlabs:
LGTM

Confirmed that both of the issues have been fixed and both demos work beautifully!

abidlabs (approved):

PR #944, opened 2022-04-07
examples, cache_examples fix

aliabid94:
Fix examples issues. Examples supported for:
- Textbox
- Number
- Slider
- Checkbox
- CheckboxGroup
- Radio
- Dropdown
- Image
- Video
- Audio
- File
- Dataframe

There were issues previously with file-type examples, e.g. Audio. Now these components use the URL instead of loading the data for examples, (much faster than converting the entire file to base64 and then sending to frontend).

Also cache_examples should work now. Functionality slightly different, now both input and output will automatically load on click if examples are cached. 

Issues found in components not related to this PR:
- Checkbox, CheckboxGroup, and Dataframe do not show updated values in the frontend when a function outputs to them.


pngwn:
#929 should fix checkbox and checkbox group not updating correctly. I'll be looking at Dataframe today or tomorrow.

demo/image_classifier/run.py
```diff
@@ -17,15 +18,13 @@ def classify_image(inp):
     return {labels[i]: float(prediction[i]) for i in range(1000)}
 
 
-image = gr.Image(shape=(224, 224))
-label = gr.Label(num_top_classes=3)
+image = gr.inputs.Image(shape=(224, 224))
```
FarukOzderim, line 21: why use deprecated components?

gradio/blocks.py
```diff
@@ -36,6 +36,7 @@ def set_event_trigger(
         inputs: List[Component],
         outputs: List[Component],
         preprocess=True,
+        postprocess=True,
```
FarukOzderim, line 39: I've missed preprocess in events, what is it for, let's add them to docstring?

demo/image_classifier/run.py
aliabid94: Whoops. Fixed

aliabid94:
Okay removed examples_dir and made FileData.name the ultimate source, but the concatenation  to prefix `root + "file/"` does need to be done in the frontend because of cases where the frontend is not being served from the url as the backend (e.g. dev mode, embedded gradio instances), the frontend needs to create the full path to the resource.

aliabid94:
added logic to normalise_file

pngwn (approved):
This looks great.

FarukOzderim:
Doing some changes to the branch for deleting big files that entered to this branch with #955

pngwn:
I think that failure is flake, rerunning to check.

FarukOzderim:
@aliabid94 did you rebase? Abubakar's commit "Added another demo and tweaks.." still lives in branch.

FarukOzderim:
If it is not resolved with a rebase I will edit this branch a little bit and remove that commit when I sit to the pc.

FarukOzderim:
@aliabid94 if you have not rebased pls rebase like I mentioned. And if you have, please remove this commit from this branch. (Added another demo and tweaked some of the older ones). If you can't I can take a look in the next week as well.

> @aliabid94 can you rebase and resolve conflicts in this branch, as some big files entered to this branch with #955, but removed from other branches.
> 
> `git rebase blocks-dev`
> 
> These files create conflict: image

abidlabs:
So as discussed, let's close this @aliabid94 in favor of #971

PR #947, opened 2022-04-07
Blocks analytics

### LINKED ISSUE #705: [Blocks] Blocks-specifics Analytics
Our segment analytics should differentiate based on what kind of `Launchable()` is launched: `Block` vs. `Interface` vs. others in the future.  

In addition, are there other, Block-specific analytics we want to measure?
* What kind of components are being used?
* What kind of static components are being used?
* More error logging? I anticipate that a lot more errors will happen with Blocks compared to regular Interfaces.

abidlabs:
Added launch analytics to `Blocks`. Didn't bother with initiate analytics because we don't use them at all

Closes #705

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -158,11 +158,13 @@ class TabItem(BlockContext):
 
 
 class Blocks(Launchable, BlockContext):
-    def __init__(self, 
-                 theme: str = "default", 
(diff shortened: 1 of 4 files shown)


abidlabs:
Made the relevant changes in `gradio-api-server` as well

dawoodkhan82 (approved):
looks good. lgtm

PR #955, opened 2022-04-07
Added another demo and tweaked some of the older ones

abidlabs:
* New demo is called `/demo/fake_gan` and looks something like this:

image


aliabid94 (approved):
LGTM

PR #966, opened 2022-04-08
fix i18n and some tweaks

### LINKED ISSUE #933: [blocks-dev] i18n is broken in blocks-dev
### Describe the bug

a bunch of the i18n we added is broken. i know cos i broke it.

### Reproduction

-

### Screenshot

-

### Logs

```shell
-
```

### System Info

```shell
-
```

### Severity

critical

pngwn:
Fix i18n and a few tweaks/ cleans ups to props

Closes #933.


pngwn:
I don't actually know how to test i18n properly.

abidlabs:
For testing, you can change your browser language: https://support.google.com/chrome/answer/173424?hl=en&co=GENIE.Platform%3DDesktop ("Change the language of your Chrome browser" section)

I'll also test it out now

abidlabs:
The components look good:

image

However, the buttons are not being translated. This is probably because we hardcode the value of the buttons on the frontend when we create them.

pngwn:
@abidlabs could we change what we are hardcoding in the backend to be the language item key rather than the english string, basically just the lowercase english string?

Here is the list:

```
"submit": "Submit",
"clear": "Clear",
"interpret": "Interpret",
"flag": "Flag",
"examples": "Examples"
```

abidlabs:
Yup definitely, let me push that now

abidlabs:
Actually thinking about it again, it makes more sense to update the keys in the internationalization files to be the cased versions, right? Because it doesn't make sense to "translate" from lower case to upper case (if their locale is English)

pngwn:
Actually if you could do `interface.[word]` that would be ideal because if a match isn't found in the i18n lib it just passes through the provided string. Otherwise I"ll need to jump through some hoops to make this work.

```
$_("interface.submit") -> `Submit`
$_("Some String") -> "Some String"
```

We can swap to Cased words in the i18n files.

abidlabs (approved):

PR #968, opened 2022-04-08
Cleaning up `Launchable()`

abidlabs:
Now that `Interface` is written entirely in terms of `Blocks` (and all future abstractions should be as well), we no longer need a separate `Launchables` class. 

This PR will: 
* move `launch()` inside `Blocks`
* I've also been playing around with `Blocks` in jupyter notebooks and of the things that bothers me is having to relaunch Blocks and waiting 2 seconds for it to restart every time I make a change. I'm going to try to see if we can just "update the config" if you relaunch a `Blocks` that is running rather than having to start a whole new server.  For people who use jupyter notebooks, this will be a much nicer experience, I think.


aliabid94:
Lgtm

PR #971, opened 2022-04-09
Fix #944

FarukOzderim:
fixes #944, 
soft reset and remove a commit, then commit the further commits again


FarukOzderim:
Hehe I faked to be Ali in the previous commit .D

PR #975, opened 2022-04-11
New Blocks Demo: neural instrument cloning

abidlabs:
Working on a really complete, sophisticated Blocks demo inspired by this report: https://erlj.notion.site/Neural-Instrument-Cloning-from-very-few-samples-2cf41d8b630842ee8c7eb55036a1bfd6

Also fixed some bugs I noticed while writing it


aliabid94:
What is the size cutoff we’re allowing for binary files in the respositort? Lgtm otherwise?

aliabid94:
Might be confusing to have interfaces within tabbedinterfaces within blocks.

abidlabs:
> What is the size cutoff we’re allowing for binary files in the respositort? Lgtm otherwise.

Hmm how about 1 MB? I'll put a different audio file so it's under the limit (cc @FarukOzderim -- have thoughts on this?)

PR #977, opened 2022-04-12
State and variables

aliabid94:
Created new component for storing state called a Variable. There can be multiple variables, all stored in the backend. Original state functionality in interfaces still supported through the "state" string shortcut. Can be tested through chatbot demo or simple_squares demo. Also created flashcards app demo, though the broken dataframe component makes the app not functional.


FarukOzderim:
Questions: 
1. What is the purpose of `variable`?
2. Will backend just send updates of the variables or send them every time?

gradio.egg-info/requires.txt
```diff
@@ -1,5 +1,5 @@
-aiohttp
 analytics-python
+aiohttp
```
FarukOzderim, line 2: What are these unnecessary changes for?

gradio/blocks.py
```diff
@@ -208,29 +208,46 @@ def render(self):
         self._id = Context.id
         Context.id += 1
 
-    def process_api(self, data: Dict[str, Any], username: str = None) -> Dict[str, Any]:
+    def process_api(
+        self,
+        data: Dict[str, Any],
+        username: str = None,
+        state: Optional[Dict[int, any]] = None,
```
FarukOzderim, line 215: let's create a docstring and explain state inside it?

gradio/blocks.py
```diff
@@ -208,29 +208,46 @@ def render(self):
...
-                if predictions[i] is not None
-                else None
-                for i, output_id in enumerate(dependency["outputs"])
-            ]
+            output = []
+            for i, output_id in enumerate(dependency["outputs"]):
+                block = self.blocks[output_id]
+                if getattr(block, "stateful", False):
+                    state[output_id] = predictions[i]
+                    output.append(None)
```
FarukOzderim, line 242: why not append it to output?

gradio/blocks.py
```diff
@@ -208,29 +208,46 @@ def render(self):
...
+                else:
+                    output.append(
+                        block.postprocess(predictions[i])
+                        if predictions[i] is not None
+                        else None
+                    )
+            else:
+                output = predictions
```
FarukOzderim, line 250: why do we override `output` after the loop is finished?

gradio/interface.py
```diff
@@ -178,27 +179,19 @@ def __init__(
         if not isinstance(outputs, list):
             outputs = [outputs]
 
+        if "state" in inputs or "state" in outputs:
+            if len([i for i in inputs if i == "state"]) != 1 or len([o for o in outputs if o == "state"]) != 1:
```
FarukOzderim, line 183: shall we break this check for readibility?

FarukOzderim, line 183:
```suggestion
            state_input_count = len([i for i in inputs if i == "state"])
            state_output_count = len([o for o in outputs if o == "state"])
            if state_input_count  != 1 or state_output_count  != 1:
```

FarukOzderim (changes requested):
Good work!

aliabid94:
> 



> Questions:
> 
> 1. What is the purpose of `variable` and this PR? Was there no multiple state component support previously?
> 2. Will backend just send updates of the `variables `or send `variable `data every time?

1. There was a State component, but it was only possible to have one State component with an interface. It did not work with Blocks at all. Variables have replaced State.
2. Variable data is stored completely in the backend, unlike all other components. It is never passed to the frontend. 

The main purposes of variable are:
- Allow users to be able to store state between interactions without having them displayed in a frontend component
- Be able to store large binary data, e.g. model state, between interactions that may not be JSON serializable easily
- Bring statefulness to Blocks

gradio/blocks.py
aliabid94: We don't want to send the state to the frontend. Variables are not displayed in the frontend, and may be massive or binary data. We have to append to preprocess because it is needed to run the function, but variable data is never sent to the frontend.

gradio/blocks.py
aliabid94: Whoops typo - the return of this function should be output. Output is not overridden after the loop, its the 'else' clause where there is no postprocesssing, so the output is just the predictions.

gradio/interface.py
aliabid94: Done.

demo/flashcards/run.py
```diff
@@ -0,0 +1,90 @@
+import gradio as gr
```
abidlabs, line 1: Suggest naming this folder `blocks_flashcards`

demo/simple_squares/run.py
```diff
@@ -0,0 +1,16 @@
+import gradio as gr
```
abidlabs, line 1: Suggest naming this folder `blocks_simple_squares` to be consistent with naming of all of the blocks demos.

gradio/interface.py
```diff
@@ -178,27 +179,21 @@ def __init__(
         if not isinstance(outputs, list):
             outputs = [outputs]
 
+        if "state" in inputs or "state" in outputs:
```
abidlabs, line 182: Will this work if someone passes in a `gr.Variable()` instead of `"state`"

gradio/routes.py
```diff
@@ -308,15 +320,13 @@ def get_types(cls_set: List[Type], component: str):
...
+        " component. Please see the getting started for more information."
     )
 
 
 def set_state(*args):
     raise DeprecationWarning(
-        "This function is deprecated. To create stateful demos, pass 'state'"
-        "as both an input and output component. Please see the getting started"
-        "guide for more information."
+        "This function is deprecated. To create stateful demos, use the Variable"
```
abidlabs, line 330: I think we should leave the old warnings. If someone is using `gr.set_state()` and `gr.get_state()`, they are probably trying to create an `Interface`, not a general `Blocks`

abidlabs:
Left some minor comments, but overall LGTM.

gradio/interface.py
aliabid94: If they pass the same variable, yes. Meaning
```
var = gr.Variable()
gr.Interface(fn, ["text", var], ["text", var])
```
will work, but
```
gr.Interface(fn, ["text", gr.Variable()], ["text", gr.Variable()])
```
will not map to the same variable. Using the string "state" automatically creates the same Variable for both input and output.

demo/simple_squares/run.py
aliabid94: Why are we naming them like this? I'll change for now to be consistent but not a fan of this naming.

FarukOzderim: I think one of the main advantages would be promoting and showing examples for the Blocks release.

gradio/blocks.py
FarukOzderim: As far as I can see the else comes after the for loop and it enters to the `else` scope every time.
```
for loops also have an else clause which most of us are unfamiliar with. The else clause executes after the loop completes normally.
```

PR #981, opened 2022-04-12
ensure dataframe updates as expected

### LINKED ISSUE #890: [`blocks-dev`] Dataframe input and output both behave incorrectly
### Describe the bug

Noticing some issues with the Dataframe component.

When used as an input, none of the rows show up by default. Even clicking "new row" only produces a tiny narrow row that can't really be used to input any data since it is so narrow. The column headers seem to be completely unaligned with the cells.

When used as an output, the "new column" and "new row" button still show up, though they should be disabled for an output component.

### Reproduction

demo\filter_records\run.py

### Screenshot

image

### Logs

_No response_

### System Info

```shell
blocks-dev
```

### Severity

blocker

### LINKED ISSUE #921: Dataframe input does not refresh as intended
### Describe the bug

When using a dataframe as input, the Gradio app output is locked to the input received when run for the first time. Subsequent user updates to the dataframe is not reflected in the Gradio app output. In addition, when "Clear" is selected to clear the outputs, the Gradio app behaves as if an empty dataframe is passed. The app has to be relaunched to have new input processed correctly.

Reporting both instances in the same issue as I suspect they both have to do with how Gradio handles dataframe input under the hood.

### Reproduction

Both of these are reproducible using the `filter_records.ipynb` demo at https://colab.research.google.com/drive/1mSrwzeNRMQc0FBiUY9Yx_rwqyX1X-7Cf#scrollTo=RN-8fjhplcnX.  Code reproduced below for convenience:

```
import gradio as gr

def filter_records(records, gender):
    return records[records["gender"] == gender]

iface = gr.Interface(
    filter_records,
    [
        gr.inputs.Dataframe(
            headers=["name", "age", "gender"],
            datatype=["str", "number", "str"],
            row_count=5,
        ),
        gr.inputs.Dropdown(["M", "F", "O"]),
    ],
    "dataframe",
    description="Enter gender as 'M', 'F', or 'O' for other.",
)

iface.test_launch()

iface.launch()
```

Including reproduction steps in the screenshot section below.

### Screenshot

1. Launch the app, key in the following table, and run the app. The output will correctly filter to show only the first row.

| name | age | gender |
|--------|-----|----------|
| A        | 1    | M          |
| B        | 1    | F          |
| C        | 1    | O          |

image

2. Change `gender` in rows 2 and 3 to "M", and submit the new input. All rows should be displayed as expected output, but the output did not change.

image

3. Select "Clear" then "Submit". The output is now blank, as if no rows are given as input.

image

### Logs

```shell
No logs generated
```

### System Info

```shell
Gradio version: 2.9.1, running on Google Colab
OS: MS Windows
Browser: Edge
```

### Severity

blocker

pngwn:
Fixes #890. Fixes #921.

(I think these are basically duplicates)

### DIFF
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -2038,18 +2038,16 @@ class Dataframe(Component):
             "number": 0,
             "bool": False,
             "date": "01/01/1970",
-        }        
+        }
         column_dtypes = (
             [datatype] * self.col_count if isinstance(datatype, str) else datatype
         )
         self.test_input = [
             [default_values[c] for c in column_dtypes] for _ in range(row_count)
-        ]        
+        ]
         self.default_value = (
-            default_value
-            if default_value is not None
-            else self.test_input
-        )        
+            default_value if default_value is not None else self.test_input
+        )
         self.max_rows = max_rows
         self.max_cols = max_cols
         self.overflow_row_behaviour = overflow_row_behaviour
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -2033,23 +2033,23 @@ class Dataframe(Component):
         self.col_width = col_width
         self.type = type
         self.output_type = "auto"
-        self.default_value = (
-            default_value
-            if default_value is not None
-            else [[None for _ in range(self.col_count)] for _ in range(self.row_count)]
-        )
-        sample_values = {
-            "str": "abc",
-            "number": 786,
-            "bool": True,
-            "date": "02/08/1993",
-        }
+        default_values = {
+            "str": "",
+            "number": 0,
+            "bool": False,
+            "date": "01/01/1970",
+        }        
         column_dtypes = (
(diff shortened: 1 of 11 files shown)


pngwn:
Looking at the tests, it is a bit weird.

abidlabs:
Were you able to figure out the tests @pngwn? It seems like it's affecting all of the open PRs

aliabid94:
btw we want to replace the "null"s with empty space right?

abidlabs:
Yeah we should. That's a fix in the backend from `None` to `""`

abidlabs:
Created #987 based off of this branch

abidlabs:
LGTM!

abidlabs (approved):

abidlabs:
@pngwn would you like to merge `dataframe-nulls` into this branch and then merge both into `blocks-dev`?

PR #985, opened 2022-04-12
custom-components

### LINKED ISSUE #731: [Blocks] String shortcuts for components

FarukOzderim:
- create template components
- changes in PKG and requires comes from scripts/install_gradio.sh

closes #731


gradio/__init__.py
```diff
@@ -4,6 +4,7 @@
 import gradio.inputs as inputs
 import gradio.outputs as outputs
 import gradio.processing_utils
+import gradio.templates as Templates
```
abidlabs, line 7: This has the effect of allowing us to import `gr.templates.Sketchpad` as `gr.Templates.Sketchpad`. Why is that useful? 

Instead I suggest making the templates first order components, just like components:

`from gradio.templates import Sketchpad, Webcam, ...`

gradio.egg-info/PKG-INFO
```diff
@@ -1,11 +1,14 @@
...
 Home-page: https://github.com/gradio-app/gradio-UI
 Author: Abubakar Abid, Ali Abid, Ali Abdalla, Dawood Khan, Ahsen Khaliq, Pete Allen, Ömer Faruk Özdemir
 Author-email: team@gradio.app
 License: Apache License 2.0
-Description: UNKNOWN
 Keywords: machine learning,visualization,reproducibility
 Platform: UNKNOWN
+License-File: LICENSE
+
+UNKNOWN
```
abidlabs, line 13: Why UNKNOWN on a line by itself? That seems off

demo/blocks_templates/run.py
```diff
@@ -0,0 +1,11 @@
+import gradio as gr
+user_db = {"admin": "admin", "foo": "bar"}
```
abidlabs, line 2: Not needed

demo/blocks_templates/run.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+def greet(name):
+    return "Hello " + name + "!!"
+
+
+demo = gr.Interface(fn=greet, inputs=gr.Templates.Textbox(), outputs=gr.Templates.Textbox())
```
abidlabs, line 9: I suggest using a more interesting template, such as `Webcam` for this demo.

Also, I don't see a template called `Textbox`

gradio/templates.py
```diff
@@ -0,0 +1,106 @@
+from gradio.components import Audio, Dataframe, File, Image, Textbox, Video
+
+
+class HugeTextbox(Textbox):
```
abidlabs, line 4: "Huge" is an informal word, suggest `LargeTextbox` instead.

abidlabs:
A general point I want to make:

Now that we have templates, we should remove the `get_shortcut_implementations()` method to avoid defining parameters twice. We can simply adapt `get_component_instance()` to search through both component names and template names.

But we have an internal discrepancy, which is that `gr.Textbox()` creates a textbox with 1 line, whereas `"textbox"` creates a textbox with 7 lines. I think we should resolve this discrepancy by:
* letting "textbox" only create a textbox with 1 line (breaks backwards compatibility, but in a very small way, and we are changing the design of components anyways)
* creating a `Textarea` template and "textarea" string shortcut that creates textboxes with 7 lines (as mentioned in my inline comment)

Also, we should create a separate template for `Mic` that is the same as `Microphone` for backwards compatibility

demo/blocks_templates/run.py
FarukOzderim: My bad, it was going to be HugeTextbox, will update accordingly.

gradio.egg-info/PKG-INFO
FarukOzderim: I don't have any idea, it just updated after local gradio installation, shall I remove it?

gradio/__init__.py
FarukOzderim: we do not import components from gradio.components, we import them from gradio.

it is benefical because it feels like a class and enables usage of this kind:
```
from gradio import Templates
Templates.TextArea()
```
We could also go with your suggestion however I like this one more.

FarukOzderim:
> A general point I want to make:
> 
> Now that we have templates, we should remove the `get_shortcut_implementations()` method to avoid defining parameters twice. We can simply adapt `get_component_instance()` to search through both component names and template names.
> 
> But we have an internal discrepancy, which is that `gr.Textbox()` creates a textbox with 1 line, whereas `"textbox"` creates a textbox with 7 lines. I think we should resolve this discrepancy by:
> 
>     * letting "textbox" only create a textbox with 1 line (breaks backwards compatibility, but in a very small way, and we are changing the design of components anyways)
> 
>     * creating a `Textarea` template and "textarea" string shortcut that creates textboxes with 7 lines (as mentioned in my inline comment)
> 
> 
> Also, we should create a separate template for `Mic` that is the same as `Microphone` for backwards compatibility

The only reason we support string shortcuts is for backwards compatibility. So I wouldn't consider this as a discrepancy and would not hurt to leave it like this. That's why I dropped some of the shortcuts like `text` and `mic`.
Btw we can drop string shortcuts whenever we want in the future.

FarukOzderim:
Good to merge if everything is ok!

gradio.egg-info/PKG-INFO
abidlabs: I think so yeah

abidlabs: I think so yeah

gradio/__init__.py
abidlabs: > it is benefical because it feels like a class and enables usage of this kind:

I don't think this is standard usage. You shouldn't import a Python module as a class

> we do not import components from gradio.components, we import them from gradio.

Yeah, let's do the same with templates. So that we can do `gr.Sketchpad`, `gr.Webcam`, etc.

abidlabs: > it is benefical because it feels like a class and enables usage of this kind:

I don't think this is standard usage. You shouldn't import a Python module as a class

> we do not import components from gradio.components, we import them from gradio.

Yeah, let's do the same with templates. So that we can do `gr.Sketchpad`, `gr.Webcam`, etc.

gradio/blocks.py
```diff
@@ -242,7 +249,11 @@ def get_config_file(self):
             config["components"].append(
                 {
                     "id": _id,
-                    "type": block.__class__.__name__.lower(),
+                    "type": (
```
abidlabs, line 252: This could be `block.get_block_name()` given the method that was just added to the `Blocks` class

gradio.egg-info/PKG-INFO
abidlabs: I think so yeah

gradio/__init__.py
abidlabs: > it is benefical because it feels like a class and enables usage of this kind:

I don't think this is standard usage. You shouldn't import a Python module as a class

> we do not import components from gradio.components, we import them from gradio.

Yeah, let's do the same with templates. So that we can do gr.Sketchpad, gr.Webcam, etc.

abidlabs:
Sorry GitHub was being weird for me, and I think some of my comments above got duplicated.

gradio/blocks.py
FarukOzderim: good catch!

FarukOzderim: well I actually mainly created the function to use here, but it seems I forgot, thx for the careful review.

FarukOzderim:
Good to go! Updating get_shortcut_implementations in another branch.

gradio/routes.py
```diff
@@ -175,8 +174,8 @@ def api_docs(request: Request):
         outputs = [type(out) for out in app.blocks.output_components]
         input_types_doc, input_types = get_types(inputs, "input")
         output_types_doc, output_types = get_types(outputs, "output")
-        input_names = [type(inp).__name__ for inp in app.blocks.input_components]
-        output_names = [type(out).__name__ for out in app.blocks.output_components]
+        input_names = [inp.get_block_name() for inp in app.blocks.input_components]
+        output_names = [out.get_block_name() for out in app.blocks.output_components]
```
abidlabs, line 178: Nice

abidlabs:
Yeah, so this looks good. Though I still think we should change:

`import gradio.templates as Templates`

to 

`from gradio.templates import Webcam, Sketchpad, ..`

FarukOzderim:
> Yeah, so this looks good. Though I still think we should change:
> 
> `import gradio.templates as Templates`
> 
> to
> 
> `from gradio.templates import Webcam, Sketchpad, ..`

Adding @aliabid94 and @pngwn as it seems we cannot reach to an agreement :)

I am not in favor of importing templates via
```
import gradio as gr
gr.Webcam()
```
Related [comment](https://github.com/gradio-app/gradio/issues/731#issuecomment-1096717600).

I prefer these 
```
from gradio import Templates

sketchpad = Templates.Sketchpad()
```


```
from gradio.Templates import Sketchpad

sketchpad = Sketchpad()
```

Btw I am okay with `templates `instead of `Templates `as well, but I think `Templates ` look way more nicer.

aliabid94:
Sorry for the late comment on this. I do not understand the purpose here. Are we trying to encourage the use of these Template components? I think it's going to be *very* confusing for a user to have both Components and Templates in the same namespace. When should they use gr.Image(type="pil"), and when should they use gr.PIL()? And if they use gr.PIL, how do they find the documentation for the other constructor arguments - do they go to gr.Image, or is the documentation duplicated entirely for this gr.PIL class?

Using naming format gr.Template.PIL creates a distinction between Templates and regular Components, but having gr.Template.PIL() is such a long name, it does not save enough time compared to gr.Image(source="pil") to be worth it. It also just adds a lot of duplicate documentation.

I really think having these Templates is going to create much more confusion down the line. If we want to support string shortcuts, then they should remain strings. Creating extended subclasses, when in most cases the extension only sets a single keyword argument is just making two ways to do the same thing, which only adds confusion. It's true that the original string shortcuts also introduced two ways to do the same thing, but because the strings were **short** and they **could not be configured like classes**, it was clear when to use the string shortcut and when the use the Component class. 

If we want to keep string shortcuts, I would be *maybe* be okay with something like gr.get("webcam").

Happy to discuss in person.

FarukOzderim:
Hmm, I actually agree with some of your points. Though the reason we have these templates is because we want to support previous use of string shortcuts in Blocks(which is I am against it actually, we should drop the shortcuts and make people use components like you mentioned because of the points you gave above). Other than we can't use string shortcuts in blocks, and it is better to have a class instead of `gr.get("webcam")`.


Btw having shortcuts for `Webcam`, `Microphone` and `Sketchpad` are hand but `PIL` and etc. are not.

FarukOzderim:
Btw for shortcutting in Blocks we can make use of this [function](https://github.com/gradio-app/gradio/pull/995/files#diff-8b617cd56845dba0483334e849e314582a743a488a7e2fe616cd3f23596e066fR124-R144) in the other PR:
```

from gr.components import get_component_shortcut

mic = get_component_shortcut("mic")
text= get_component_shortcut("textbox")
dropdown= get_component_shortcut("dropdown")
json = get_component_shortcut("json")



```

abidlabs:
I'm persuaded as well.

Ok so let's just stick to using the `get_component_shortcut()` method instead of defining `templates`.

@FarukOzderim we can still keep the demos here, let's just update them to use `get_component_shortcut()` and we can focus on other things.

Issue #821, opened 2022-03-15
Allow users to change arbitrary component inputs


pngwn:
We chatted briefly about this offline, but how should we handle multiple inputs (and possibly outputs for a component). The work in #807 partially addresses this issue but this slide/dropdown example would not be possible currently as we do not have a way to have multiple inputs/ouputs and inconsistent inputs/outputs:

```py
from gradio.blocks import *

def dropdown_values(val):
    if val < 50:
       return ["fail", "repeat"]
    else:
       return ["succeed", "repeat"]
  
def fn(val, action):
     return "Your score is {} and you decided to {}".format(val, action)


with Blocks() as block:

  s = Slider(0, 100)
  d = Dropdown([]) 
  t = Textbox()

  s.select(fn=dropdown_values, outputs=d)
  d.select(fn=fn, inputs=[s, d], outputs=t)

  block.launch()
```

@aliabid94 Mentioned recreating a new config but I'm concerned about that approach for a number of reasons.

If we could gather our thoughts here and figure out how we should proceed.

aliabid94:
I was thinking something the lines of this:
```python
from gradio.blocks import *

def dropdown_values(val):
    if val < 50:
       opts = ["fail", "repeat"]
    else:
       opts = ["succeed", "repeat"]
    return Dropdown.update(choices=opts)
  
def fn(val, action):
     return "Your score is {} and you decided to {}".format(val, action)


with Blocks() as block:

  s = Slider(0, 100)
  d = Dropdown(chocies=[]) 
  t = Textbox()

  s.change(fn=dropdown_values, outputs=d)
  drop.select(fn=fn, inputs=[s, d], outputs=t)

  block.launch()
```

So if you return normal values from a function, that becomes the new value represented by the Block. But if you return a Block.update(), the config of the Block gets updated instead. Block.update takes the same args as the constructor for the Block itself. 

It's possible to update both the config and the value of a component, e.g.

```python
from gradio.blocks import *

def dropdown_values(val):
    if val < 50:
       opts = ["fail", "repeat"]
    else:
       opts = ["succeed", "repeat"]
    return "repeat", Dropdown.update(choices=opts)
  
with Blocks() as block:

  s = Slider(0, 100)
  d = Dropdown(chocies=[]) 

  s.change(fn=dropdown_values, outputs=[d, d])

  block.launch()
```

aliabid94:
Block.update being a class level method implemented for every Block/

pngwn:
[...]

PR #986, opened 2022-04-12
update-components

FarukOzderim:
- implement update function
- create an example demo

backend part of #821

### DIFF
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -209,6 +209,17 @@ class Component(Block):
         """
         return x
 
+    @staticmethod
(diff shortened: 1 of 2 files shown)


abidlabs:
LGTM

gradio/components.py
```diff
@@ -209,6 +209,17 @@ def deserialize(self, x):
         """
         return x
 
+    @staticmethod
```
abidlabs, line 212: Oh yep static method is the right call here!

abidlabs (approved):

PR #992, opened 2022-04-13
added interactive parameter to components

### LINKED ISSUE #925: Adding an `interactive` parameter so that `Interface` outputs are not treated as inputs
### Describe the bug

As discussed previously, output components in an `Interface` are treated like `inputs`, presumably because they are actually an input to the flagging button.

In this case, @pngwn suggested adding a parameter such as `interactive`, which can be `None`, `True`, or `False`. If `True` or `False`, it overrides the frontend's determination of what is an input and what is an output.

cc @aliabid94 

### Reproduction

Any of the Interface demos

### Screenshot

_No response_

### Logs

_No response_

### System Info

```shell
blocks-dev
```

### Severity

critical

abidlabs:
All `Component` classes now have an `interactive` parameter, which can be `True`, `False`, or (by default), `None`.

In the `Interface` class, input components have `interactive` set to `None`, but output components set to `False`

Closes #925


pngwn:
@abidlabs I have implemented the frontend for this _kindof_. There are some bugs/ inconsistencies with the static version of components but the mechanism is implemented and can be tested. We can create new issues for any bugs we find.

pngwn:
Hmm test failing, I'll have to take a look at that a little later.

pngwn:
@abidlabs Should be good now. Made some additional tweaks to ensure everything is rendering as expected, even when empty.


We'll need to cleanupo the components at some point, the line between what stays in the app and what goes into the separate component is a little blurry atm but we can sort that out later.

pngwn:
@abidlabs Should be good now. Made some additional tweaks to ensure everything is rendering as expected, even when empty.


We'll need to cleanupo the components at some point, the line between what stays in the app and what goes into the separate component is a little blurry atm but we can sort that out later.

PR #995, opened 2022-04-13
Refactor component shortcuts

FarukOzderim:
will be merged after #985


gradio/components.py
```diff
@@ -131,12 +124,25 @@ def restore_flagged_file(
         return {"name": file, "data": data}
 
     @classmethod
-    def get_all_shortcut_implementations(cls):
-        shortcuts = {}
+    def get_component_shortcut(cls, str_shortcut: str) -> (bool, Optional[Component]):
```
abidlabs, line 127: It's redundant for this function to return two values. It should just return a component or `None`

gradio/components.py
```diff
@@ -2958,10 +2851,29 @@ def get_template_context(self):
...
+    """
+    Creates a component, where class name equals to str_shortcut.
+
+    @param str_shortcut: string shortcut of a component
+    @return:
+        True, found_class or
+        False, None
+    """
+    does_exist, component = Component.get_component_shortcut(str_shortcut)
+    if not does_exist:
```
abidlabs, line 2864: Following my previous comment, here, you can just check: `if component is None`

abidlabs:
Besides the nit above, looks good to me

Issue #572, opened 2022-02-09
Redesign the Queue


FarukOzderim:
There were some errors related to the queueing when tests were divided into subfolders, #551 .

get_status function for the queue had some bugs which can be seen [here](https://app.circleci.com/pipelines/github/gradio-app/gradio/1444/workflows/c0992a04-a11f-48a1-a211-593202beb8ea/jobs/1445).

Furthermore @aliabd's local machine were giving errors because queue was starting from the index -1.

After the order of the test subfolders were changed another error has appeared like [this](https://app.circleci.com/pipelines/github/gradio-app/gradio/1471/workflows/75a8420e-9e7a-4674-a1db-be61bffc192c/jobs/1472).

Building on top of these I would suggest we move queueing from the database to a list variable, which would accelerate things as database accesses slow us down.

abidlabs:
Makes sense, if we can avoid file-based databases, it would probably simplify things. @aliabid94 can you add your thoughts here? Why did we use a file database for something as simple as just keeping a queue?

FarukOzderim:
Comment about queueing from @cbensimon, we shall take this into account while refactoring queueing.

```
I agree (even if the experience is slightly degraded when the processing is really fast (<< 1s) because the queue only refreshes every second)

In future Gradio versions, I think that we should make the queue totally transparent to the developer / user (by improving the internals of the backend in a way or another, whether using WebSockets, or Server Side Events, or any other technique)
To me the queue is an implementation detail and we shouldn't make the developer responsible for making this decision.
```

cbensimon:
> Makes sense, if we can avoid file-based databases, it would probably simplify things. @aliabid94 can you add your thoughts here? Why did we use a file database for something as simple as just keeping a queue?

Totally agree, I don't think that a DB was needed for the queue

FarukOzderim:
We should also think of communication of queuing with autoscaling. We will probably collaborate with @cbensimon 🤗

FarukOzderim:
I feel like this could stay in low priority as it could need some effort, dropping the milestone.

pngwn:
I do think websockets would be the way to go in the future for the queue. I think this was discussed briefly somewhere (may have been on slack).

FarukOzderim:
There was a nice feature suggestion from @ronvoluted.

Let's have a endpoint to get queue information from the Gradio server in the new design.

aliabid94:
[...]

PR #997, opened 2022-04-14
Progress indicator bar

aliabid94:
Created a progress indicator that can be attached to an event listener API call. Still a little work to show ETA and queue position, but mostly functional, try out demo/calculator.py. I think there will be opinions on how this is implemented so thought to open PR now to get feedback.


gradio/components.py
```diff
@@ -382,25 +382,43 @@ def deserialize(self, x):
         """
         return x
 
-    def change(self, fn: Callable, inputs: List[Component], outputs: List[Component]):
+    def change(
+        self,
+        fn: Callable,
+        inputs: List[Component],
+        outputs: List[Component],
+        status_tracker: Optional[StatusTracker] = None,
```
FarukOzderim, line 390: A question here, why do we provide status_tracker to user?

gradio/blocks.py
```diff
@@ -57,6 +58,7 @@ def set_event_trigger(
             preprocess: whether to run the preprocess methods of components
             postprocess: whether to run the postprocess methods of components
             no_target: if True, sets "targets" to [], used for Blocks "load" event
+            status: StatusTracker to visualize function progress
```
FarukOzderim, line 61:
```suggestion
            status_tracker: StatusTracker to visualize function progress
```

gradio/blocks.py
```diff
@@ -65,14 +67,17 @@ def set_event_trigger(
...
 
-        Context.root_block.fns.append((fn, preprocess, postprocess))
+        Context.root_block.fns.append(BlockFunction(fn, preprocess, postprocess))
         Context.root_block.dependencies.append(
             {
                 "targets": [self._id] if not no_target else [],
                 "trigger": event_name,
                 "inputs": [block._id for block in inputs],
                 "outputs": [block._id for block in outputs],
                 "queue": queue,
```
FarukOzderim, line 77: how queue is used within frontend? is it currently used?

gradio/blocks.py
```diff
@@ -46,6 +46,7 @@ def set_event_trigger(
         postprocess: bool = True,
         queue=False,
```
FarukOzderim, line 47: let's add queue to docstring

FarukOzderim (commented):
Beautiful design!

FarukOzderim:
How do we handle parallel events or actions in the frontend?
Do we wait until the previous event's processing is completed before sending another event to Backend server? I feel like that would be the cleanest, but might decrease responsiveness..

FarukOzderim:
Dropping Queueing issue here for design reference: #572
While designing queueing we need to take autoscaling into account. We might not design it yet, but may brainstorm.

demo/calculator/run.py
```diff
@@ -1,7 +1,8 @@
 import gradio as gr
-
+import time, random
 
 def calculator(num1, operation, num2):
+    time.sleep(5 + random.random() * 3)
```
abidlabs, line 5: Revert back before merging. You can use `demo/fake_gan` to test this out btw which already includes a short time sleep

gradio/components.py
abidlabs: I think the reason is so that users can create a status tracker for arbitrary events in Blocks

@aliabid94 it would be great to see an example of this in one of the Blocks demos

abidlabs:
Nice design, LGTM.

It would be great to see an example of a Blocks demo that uses Status Tracker for some or all of the events.

pngwn (commented):
Left a few notes but implementation looks good 👍 

I think we need to discuss how this should work, though. In my opinion we should be handling concurrent requests whereas this implementation will only handle one at a time. We also have awareness on which components will be populated as a result of the API request and should be outputting the loader visuals into the affected components. (This is also the basis on which @gary149 is deisgning as far as I'm aware as well, so it would be good to get that mechanism implemented as part of this and we can tweak the design later).

I think we should be tracking the loading status across components. So a map of some sort, keyed by the function ID which will set a prop on the relevant components of 'idle', 'pending' or 'failed'. Each component can then decide what to do for each loading state. We have all of the information for this functionality. 

The current visuals will work well for most components (it just needs to be adapter to fill the component width + height as an overlay).

Happy to go over this in-person if needed.

pngwn:
@aliabid94 Scratch some of that last comment, i misread the implementation at first.

gradio/components.py
aliabid94: Sure I'll make another demo

gradio/blocks.py
aliabid94: yes queueing should work as expected

ui/packages/app/src/Blocks.svelte
```diff
@@ -112,37 +114,49 @@
...
+	};
 
 	async function handle_mount({ detail }) {
 		await tick();
-		dependencies.forEach(({ targets, trigger, inputs, outputs, queue }, i) => {
-			const target_instances: [number, Instance][] = targets.map((t) => [
-				t,
-				instance_map[t]
-			]);
+		dependencies.forEach((dependency, i) => {
```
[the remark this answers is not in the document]
aliabid94: whoops, was testing something, undone

Issue #988, opened 2022-04-13
Problem for image examples when using absolute path.


JefferyChiang:
### Describe the bug

My code is work in 2.4.6 version. Recently I updated to 2.7.5.2, and found that the images didnt show in the examples.
After a while, I figure out that is the issue about the absolute path.
In gradio/ app.py, it will force to join the app.cwd and path, so the final path is wrong.
`@app.get("/file/{path:path}", dependencies=[Depends(login_check)])
def file(path):
    print(app.cwd, path)
    config.app = app
    if app.interface.encrypt and isinstance(
        app.interface.examples, str) and path.startswith(
            app.interface.examples):
        with open(safe_join(app.cwd, path), "rb") as encrypted_file:
            encrypted_data = encrypted_file.read()
        file_data = encryptor.decrypt(
            app.interface.encryption_key, encrypted_data)
        return FileResponse(
            io.BytesIO(file_data), attachment_filename=os.path.basename(path))
    else:
        return FileResponse(safe_join(app.cwd, path))`

After modifying the `return FileResponse(safe_join(app.cwd, path))` to` return FileResponse(path))`, it works properly.
Is the images path in examples are strict to use relative path? Because I didnt find any description of this in the Gradio Docs.

### Reproduction

Below is my code for gr.interface:
`iface = gr.Interface(fn = AD_UI,
                     inputs=[
                             gr.inputs.Radio(["Predict"], default = 'Predict'),
                             gr.inputs.Image(shape=(config.load_size, config.load_size), optional=True, type='numpy'),
                             ],
                     outputs=[
                              gr.outputs.Image(type="auto", label='Result'),
                              ],
                     examples=[['Predict','D:\\AnomalyDetection\\AD_toolbox\\datasets\\tile\\test\\good\\000.png'],
                               ['Predict','D:\\AnomalyDetection\\AD_toolbox\\datasets\\bottle\\test\\good\\000.png'],
                               ],
                     examples_per_page = 5,
                     layout = 'vertical',
                     theme = 'default',
                     css=css,
                     allow_flagging = 'never',
                     live = False,
                     )`

### System Info

```shell
python 3.7.7
gradio 2.7.5.2
chrome
windows 7
```

### Severity

annoyance

abidlabs:
Thanks for catching this @JefferyChiang, would you like to open a PR so that we can credit you for the fix?

PR #1004, opened 2022-04-15
Fixed image input for absolute path

JefferyChiang:
Changed the File function in route.py to make image input available for absolute path.
I tested the absolute, relative path and same folder for image, and it worked properly.

# Description

Please include: 
* relevant motivation
* a summary of the change
* which issue is fixed. 

Fixes: #988  (issue)

# Checklist:

### DIFF
--- a/gradio/routes.py
+++ b/gradio/routes.py
@@ -212,7 +212,7 @@ def file(path):
             io.BytesIO(file_data), attachment_filename=os.path.basename(path)
         )
     else:
-        return FileResponse(safe_join(app.cwd, path))
+        return FileResponse(app.cwd, path)
 
 
 @app.get("/api", response_class=HTMLResponse)  # Needed for Spaces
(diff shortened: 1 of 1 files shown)


abidlabs:
Hi @JefferyChiang thanks for the contribution! 

Upon a closer look at this, this introduces a potential security concern in that it could allow users to access files on the demo creator's computer that the creator does not want to provide access to. We actually introduced `safe_join` for this very reason, see https://github.com/gradio-app/gradio/security/advisories/GHSA-rhq2-3vr9-6mcr

However, there is an improvement that we can do, which is to support absolute paths, as long as they are _within_ the working directly, `app.cwd`. 

We could use an approach like this to make sure that filepath is within `app.cwd`: https://stackoverflow.com/a/34236245/5209347

If you can make this update and address this security concern, happy to review and merge.

abidlabs:
Ready to merge, thanks so much @JefferyChiang!

JefferyChiang:
Thanks! I really learned a lot from your reply!

Issue #980, opened 2022-04-12
Textbox Autoheight Logic


FarukOzderim:
Continuation of halted discussion at #929 

I propose having `lines=1` by default, always having autoheight logic for `textbox` component to increase its size with the incoming input. What are your thoughts @abidlabs @aliabid94 @pngwn?

```
@abidlabs @pngwn I think having input textbox lines dynamically changing would be great as well. It would make it much more less cumbersome, WDYT?

Guess it depends how picky users are about how the component is laid out. We could use the autoheight logic when a lines value isn't provided?

Yeah I thought the same thing at first, then later thought what about respecting the line size initially and let the line size increase when input does not fit in.
Btw the default lines value is 1 currently, but we could change it.
```

abidlabs:
I don't have a strong opinion on this. I can see how increasing the input textbox height can be useful in some cases, but in other cases (imagine a large text that is copy-pasted), then it could lead to the textbox becoming very large and crowding out the other components.

FarukOzderim:
> I don't have a strong opinion on this. I can see how increasing the input textbox height can be useful in some cases, but in other cases (imagine a large text that is copy-pasted), then it could lead to the textbox becoming very large and crowding out the other components.

Is large text a use case though?
If it is then we shan't enable it by default or make it closable by parameters. 
Then how does this sound?

- If lines is not provided we make it autoheight
- If autoheight is True, we make it autoheight
- otherwise autoHeight is False

pngwn:
We could also set a max pretty easily.

abidlabs:
I assume that should be `autoheight=False` above. 

Otherwise, seems reasonable. 

I think we can wait until after 3.0 as this doesn't seem be urgent.

FarukOzderim:
Would like to include in 3.0 if it is a short work from frontend perspective. Otherwise can wait in low priority.

pngwn:
Frontend logic is already there really (for outputs), just needs hooking up to whatever param we want to use to control it. I think there should be a default max_height/ max_lines otherwise weird stuff might happen. Probably around 10.

FarukOzderim:
Ok then, I will update accordingly.

abidlabs:
Sorry one quick thing:

If we have a `lines` parameter and a `max_lines` parameter, do we need `autoheight`? We can just allow a user to block auto-adjusting by setting `max_lines` equal to `lines`? We can cutdown on one unnecessary parameter and prevent parameter settings from clashing

[...]

PR #1009, opened 2022-04-15
textbox-autoheight

FarukOzderim:
- add max-lines to textbox

Backend side of #980


gradio/components.py
```diff
@@ -247,6 +247,7 @@ def __init__(
         default_value: str = "",
         *,
         lines: int = 1,
+        max_lines: int = 100,
```
abidlabs, line 250: How about we leave `max_lines` as `None` by default, which allows the Textbox to get arbitrarily large. And set its type to be `Optional[int]`

abidlabs:
Looks good, see my comment above -- note that we will also need to fix a bunch of tests that currently test the config produced by the Textbox

gradio/components.py
FarukOzderim: I had the same intention, my arbitrarily large number was 100, what is your arbitrarily large number? Frontend will have its own max value, so we can give a default value of arbitrarily large number.

abidlabs: Why does there need to be a max value set in the frontend? I would set the default for `max_lines` to be `None`, meaning that there is no max -- the textbox just keeps getting longer.

FarukOzderim: @pngwn I think we would not let Textbox grow infinitely large, that's what I understood from our discussions and how I feel. 
What are your thoughts on this @pngwn?

pngwn: I think allowing the textbox to grow infinitely large is a poor default, we can easily set a default in the frontend though without there being one in the back.

FarukOzderim: What about allowing 1 million max_lines if the user prefers. I think we would have a max for that as well right?

FarukOzderim: Btw this design assigns a max value(100) when the user does not specify it, we could increase it if we wish. I don't see a need for passing None to frontend and frontend selecting that max value.

pngwn: @FarukOzderim We dont' need to pass anything to the front end particularly, the absence of any preference can trigger a default. I don't think UX defaults should live in the python library particularly.

20 is a better default. We don't want to a hard maximum because people may want longer 'prose'.

PR #1022, opened 2022-04-19
backend-async-support

FarukOzderim:
- update reqs
- create skeleton for Http
- write simple tests for Http

This PR is a Http wrapper suggestion. Let's hear your ideas. When we reach an agreement all network requests will be replaced by this design. You can suggest functionalities for the design as well.

### DIFF
--- a/gradio/utils.py
+++ b/gradio/utils.py
@@ -361,7 +361,8 @@ class Http:
     A class that shares an AsyncClient which sends and handles network requests.
     Further different functionalities can be gathered within this class.
     - Client configurations
-    - Error types and error handling
+    - Response validation
+    - Support error types and error handling
     - Client stop/restart
     """
 
--- a/gradio/utils.py
+++ b/gradio/utils.py
@@ -12,12 +12,14 @@ import random
 import warnings
 from copy import deepcopy
 from distutils.version import StrictVersion
-from typing import TYPE_CHECKING, Any, Callable, Dict, List
+from typing import TYPE_CHECKING, Any, Callable, Dict, Optional
 
 import aiohttp
 import analytics
 import pkg_resources
 import requests
+from httpx import AsyncClient
+from pydantic import BaseModel
 
 import gradio
 
@@ -352,3 +354,48 @@ def format_ner_list(input_string: str, ner_groups: Dict[str : str | int]):
 
     output.append((input_string[end:], None))
     return output
+
+
+class Http:
+    """
+    A class that shares an AsyncClient which sends and handles network requests.
+    Further different functionalities can be gathered within this class.
+    - Client configurations
+    - Error types and error handling
+    - Client stop/restart
+    """
+
+    client = AsyncClient()
+
+    @classmethod
+    async def request(
+        cls,
+        method: str,
+        url: str,
+        json: dict = None,
+        validation_model: Optional[BaseModel] = None,
+        **kwargs,
+    ) -> (int, dict):
+        """
+        Wrapper function for AsyncClient request.
+        Parameters:
+            method: GET, POST etc.
+            url: URL
+            json: json payload
+            validation_model: Validation model
+            kwargs: any additional kwargs
+        Returns:
+        """
+        # TODO: Might add exception handling
+        response = await cls.client.request(
+            method=method.lower(), url=url, json=json, **kwargs
+        )
+        if response.status_code == 200:
+            data = response.json()
+            if validation_model is not None:
+                # TODO: Do validation and if it fails raise an error
+                raise NotImplementedError("Validation handling is not implemented yet")
+        else:
+            # TODO: Raise error or retry a few times
+            return response.status_code, response.json()
+        return response.status_code, data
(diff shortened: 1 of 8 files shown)


pngwn:
@FarukOzderim Could you explain the rationale for this proposal? As it stands the wrapper doesn't do much.

FarukOzderim:
@pngwn That's correct. It is mainly a skeleton right now.

- It will mainly enable making network requests with a single line of code instead of [this](https://www.python-httpx.org/async/).
- It will share the client and will not open a new one at each request
- And can have any other extra functionality we want in the future.
```
    """
    A class that shares an AsyncClient which sends and handles network requests.
    Further different functionalities can be gathered within this class.
    - Client configurations
    - Response validation
    - Support error types and error handling
    - Client stop/restart
    """
```

abidlabs:
@FarukOzderim can we use the `aiohttp` library as it is already a dependency of `gradio` and we are already using it to make async requests, e.g. here:

https://github.com/gradio-app/gradio/blob/f97cbbd0d7ed1d3a58b9b1ec647158a8aaf9e7c9/gradio/utils.py#L106

FarukOzderim:
Yeah I saw aiohttp and I am planning to replace them with httpx in the future.

I am planning to use `httpx` for two reasons.
- A library from the same family fastapi/starlette/uvicorn/httpx
- Has more advanced features

FarukOzderim:
> > Yeah I saw aiohttp and I am planning to replace them with httpx in the future.
> 
> Sounds good
> 
> Now getting back to the wrapper, I'm not sure if we need to create a whole class for this. For example, looking at how we might async test the fastapi routes (https://fastapi.tiangolo.com/advanced/async-tests/), it seems like a quick 2 liner using `httpx` is sufficient:
> 
> ```python
> import pytest
> from httpx import AsyncClient
> 
> from .main import app
> 
> @pytest.mark.anyio
> async def test_root():
>     async with AsyncClient(app=app, base_url="http://test") as ac:
>         response = await ac.get("/")
>     assert response.status_code == 200
>     assert response.json() == {"message": "Tomato"}
> ```
> 
> Can you share what the equivalent would look like with the abstraction?
> 
> If we just want to share an `AsyncClient`, we can use a fixture, correct?

The point of these tests are just for demonstration purposes. The functionalities are not for tests but for the library.

However you are correct about testing the routes and sharing a fixture.

gradio/utils.py
```diff
@@ -352,3 +354,49 @@ def format_ner_list(input_string: str, ner_groups: Dict[str : str | int]):
...
+    - Support error types and error handling
+    - Client stop/restart
+    """
+
+    client = AsyncClient()
+
+    @classmethod
+    async def request(
+        cls,
+        method: str,
```
hibestil, line 374: We can use aiohttp's hdrs.METH_ALL type instead of str. 
`
from aiohttp import  hdrs
 ....
async def request(
        cls,
        method: hdrs.METH_ALL, ... ):
...
`

gradio/utils.py
```diff
@@ -352,3 +354,49 @@ def format_ner_list(input_string: str, ner_groups: Dict[str : str | int]):
...
+    - Client configurations
+    - Response validation
+    - Support error types and error handling
+    - Client stop/restart
+    """
+
+    client = AsyncClient()
+
+    @classmethod
+    async def request(
```
hibestil, line 372: I think Http naming is so generic to do just making requests. You plan to do some validations (validation model/function). So there will be request related functions such as validate and parse. Creating a class for requests may be helpful for future improvements. 

This can be psoudo-implementated like the belowing.
```python
class Request:
    def __await__():
        return self.__run().__await__()
    def __init__(self,
                method: hdrs.METH_ALL,
                url: str,
                *,
                validation_model: Union[BaseModel, Type] = None,
                error_class: Type[Error] = Error,
                **kwargs):
        # Inıt request params.
        self.request = _create_request(...)
        # TODO: Init request related stuff...

    async def __run() -> Request:
        """
        A function to make request, validate and serialize
        """
        self._response = None
        self._validated_data = None
        self._exception = None

        try:
            # make client call and get response
            # Parse response
            # handle http errors
            # validate response ( with model and/or function)
            ...
        except Exception as e:
            self._exception = ...
        return self

# Making a request
response = await Request(hdrs.METH_PUT,
                         url=url,
                         json=payload,
                         timeout=timeout)
```

WDYT?

gradio/utils.py
FarukOzderim: Thx for the suggestion!

I actually thought about using `hdrs` however since we only need "get" and "post" I thought there is no need to add aiohttp to the requirements. Planning to only have `httpx` withing requirements.

gradio/utils.py
FarukOzderim: @hibestil I think it looks very beautiful, would you like to implement this suggestion?

FarukOzderim: Note that this branch has diverged from the main a lot, you might want to merge main or create another branch and move the work in this branch to there.

FarukOzderim:
Closed in favor of #1595

PR #1023, opened 2022-04-19
fixed failing test on main

abidlabs:
Because of the ways the files were reorganized in `blocks-dev`, one of the tests that loaded an image from a URL from the GitHub repo failed. This hotfixes it.


FarukOzderim (commented):
Good job!

PR #1024, opened 2022-04-19
Use smaller ASR model in external test

abidlabs:
The `facebook/wav2vec2-large` takes a long time to load which causes our `external` tests to fail sometimes. This replaces it with `facebook/wav2vec2-base-960h`, which is more reliable (I had made this change in `blocks-dev`, but it didn't get ported over)


FarukOzderim (approved):
👍

Issue #1025, opened 2022-04-19
`ModuleNotFoundError: No module named 'gradio.test_data'` in colab notebook


abidlabs:
### Describe the bug

I tested the latest version of the library (by releasing 2.9.0b from `main`) in colab and it throws this error:

```
[/usr/local/lib/python3.7/dist-packages/gradio/components.py](https://localhost:8080/#) in <module>()
     23 from gradio import processing_utils
     24 from gradio.blocks import Block
---> 25 from gradio.test_data import media_data
     26 
     27 

ModuleNotFoundError: No module named 'gradio.test_data'
```

### Reproduction

!pip install --quiet gradio==2.9.0b0

import gradio as gr

gr.Interface(lambda x:x, "textbox", "textbox").launch()

### System Info

```shell
2.9.0b
```

### Severity

blocker

FarukOzderim:
Will investigate this issue further, it works perfectly fine when I import folder inside google colab.

https://user-images.githubusercontent.com/44067394/164151329-dcc1fdad-4d90-470e-8656-23caf99f2d1e.png

FarukOzderim:
We could try importing test_data within __init__.

PR #1027, opened 2022-04-19
Fixing import issues so that the package successfully installs on colab notebooks

abidlabs:
The pre-release `gradio==2.9.0b0` was created from the new main branch, but whenever I try to test it on colab, I get this error: #1025

This PR attempts to fix that. Closes: #1025


abidlabs:
Released `2.9.0b2` with these changes, let's see if it works in colab...

abidlabs:
This seems to be working now! (Ignore the weird footer duplication error)

image

abidlabs:
Thanks

FarukOzderim:
This actually is a very strange error and everything seems fine from project structure. Did you find any similar errors on internet? We would prefer having `media_data` under `test_data` if possible.

test/test_components.py
```diff
@@ -11,7 +11,7 @@
 import PIL
 
 import gradio as gr
-from gradio.test_data import media_data
+from gradio import media_data
```
FarukOzderim, line 14: did you need to apply these changes since you did this in test_data/__init__.py
`
import gradio.media_data as media_data`

FarukOzderim:
I surprised this error happens with a package, because it does not happen with the folders inside colab.

image

abidlabs:
Yeah @FarukOzderim. I tried various things including modifying the `test_data/init.py` file to import `media_data` in different ways, but nothing worked. And I could not find anything useful online either.

You can test this yourself by installing `!pip install gradio==2.9.0b1` in a colab notebook and then running `import gradio as gr`

I agree with you that this is not an ideal solution, since `media_data` belongs under `test_data` if possible

PR #1035, opened 2022-04-20
Model3D Examples fixes

dawoodkhan82:
# Description

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

Fixes: # (issue)

# Checklist:


abidlabs:
Looks like we just need to update the tests and we should be good to go. Let me push them now

abidlabs:
Actually I think the output Model3d component does *not* work right now. I tried a very simple function in Python:

```py
import gradio as gr

gr.Interface(lambda x:x, "model3d", "model3d").launch()
```

image

The input model3d component works just fine, but when I try to pass in an object, it does not get rendered at all in the output model3d component. It instead throws a `TypeError: Cannot read properties of null (reading '0')` error

PR #1038, opened 2022-04-20
inputless-interfaces

### LINKED ISSUE #838: Allow input-less and output-less interfaces
Sometimes, I want to create an interface without an input (such as for GANs) or without outputs (for crowd-sourced dataset generation). Internal Slack discussion https://huggingface.slack.com/archives/C02QZLG8GMN/p1647601809063159

Allowign something like `gr.Interface(fn=fn, outputs="image")`

FarukOzderim:
- enable inputless or outputless interfaces

Closes #838 

This is an outputless interface, I think it does not make sense in the Interface context:
image

This is an inputless interface, spent some time trying to make a single column layout, but it was not very direct to hack it and required a bit effort and was going to make interface design more and unnecessarily complex. I think this does the job. If anyone has more functionality or custom needs s/he can move to `Blocks`.
image

### DIFF
--- a/demo/no_input/run.py
+++ b/demo/no_input/run.py
@@ -1,0 +1,16 @@
+import gradio as gr
+import random
+
+sentence_list = [
+    "Good morning!",
+    "Prayers are with you, have a safe day!",
+    "I love you!"
+]
+
+
+def random_sentence(a):
+    return sentence_list[random.randint(0, 2)]
+
+
+demo = gr.Interface(fn=random_sentence, outputs="text")
+demo.launch()
--- a/demo/no_input/run.py
+++ b/demo/no_input/run.py
@@ -8,7 +8,7 @@ sentence_list = [
 ]
 
 
-def random_sentence(a):
+def random_sentence():
     return sentence_list[random.randint(0, 2)]
 
 
--- a/gradio/interface.py
+++ b/gradio/interface.py
@@ -172,6 +172,11 @@ class Interface(Blocks):
         """
         super().__init__(analytics_enabled=analytics_enabled, mode="interface")
 
+        if inputs is None:
+            inputs = []
+        if outputs is None:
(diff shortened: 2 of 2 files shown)


osanseviero:
> This is an outputless interface, I think it does not make sense in the Interface context:

I think it's useful if you want to allow a user to write a text, then make an audio recording, and click "Flag" to create a crowd sourced dataset.

pngwn:
I think we want to generate new blocks configs for this case + for the `unified_interface` option we discussed previously, rather than just having the same config without components.

FarukOzderim:
Brainstorming:

Shall we have UnifiedInterface instead of supporting different layouts within the Interface?

abidlabs:
> Shall we have UnifiedInterface working both as an input and output instead of supporting different layouts within the Interface? I think it would be better that way.

Can you clarify what you mean here @FarukOzderim?

FarukOzderim:
> > Shall we have UnifiedInterface working both as an input and output instead of supporting different layouts within the Interface? I think it would be better that way.
> 
> Can you clarify what you mean here @FarukOzderim?

Sure, I say we let the inputless or outputless interface just like this.
And create another Interface called `UnifiedInterface`, having a single column -- single component layout where that component either is used as an input or output or both. That way our interface design would be cleaner and we would have an interface for these requirements.

abidlabs:
Ok, I see what you're saying. If we create a separate class to handle the single-column/single-component class, let's
* make it a subclass of `Interface` if possible to avoid code duplication
* choose a good name that covers the 3 use cases (no input, no output, same input/output): I think `UnifiedInterface` only covers the 3rd one, but I can't think of a better name right now :D 

Also let's keep #838 open until we have this in, because I think this class is what users will want to use for this use case.

abidlabs (approved):

Issue #669, opened 2022-02-18
omit null values in config


pngwn:
Do not send null values in the config.

FarukOzderim:
To remember what was it about, is it the issue where `None` values in the config cause problem for frontend. Is it solvable in frontend or do we need to remove `None` values from config?

cc: @aliabid94 @pngwn

abidlabs:
I think this was saying that if any component property is `None`, it should not be included in the config, instead of it being passed in as "null"

Issue #982, opened 2022-04-12
[Blocks] Improve default labels for form components


abidlabs:
For the components in which we are requiring labels, we could improve the default labels. For example, right now, the default label for `Textbox` is `None` which gets rendered as `"null"` by the frontend and looks ugly. We could at least make the label be `"Textbox"` instead

PR #1040, opened 2022-04-20
added better default labels to form components

abidlabs:
This PR adds better default labels to form components. Instead of "null", the label now shows the name of the component. For example, the default label for a slider is "Slider": 

image

Of course, these labels can be overridden by the user to say whatever they want. In the `Interface` class, the labels continue to be based off of the parameter names (for the input components) and are "output" or "output X" for the output components.

As part of this work, I also removed any prop whose value is "null" from the config. This could potentially lead to some problems, so it would be great if you could review this @pngwn. For example, I noticed that the `interactive` prop could take `true`, `false`, or `null`, and so I adapted the logic to not depend on `null`.

Also I included some other minor cleanups that I noticed going through the components.py file. 

Closes: #669 #982


pngwn (approved):
This is all working fine, the three `interactive` cases (explicit `true`/`false` and implicit autodetection) are working perfectly and the default values look great!

abidlabs:
Yay thanks @pngwn!

ui/packages/app/src/components/Checkbox/Checkbox.svelte
```diff
@@ -4,7 +4,7 @@
 	export let value: boolean = false;
 	export let default_value: boolean = false;
 	export let style: string = "";
-	export let label: string;
```
FarukOzderim, line 7: Question: How does the label is realized when the user provides one?

test/test_utils.py
```diff
@@ -164,5 +165,24 @@ def test_format_ner_list_empty(self):
         self.assertEqual(format_ner_list(string, groups), result)
 
 
+class TestDeleteNone(unittest.TestCase):
```
FarukOzderim, line 168: Good practice to write a test for it, great!

gradio/components.py
```diff
@@ -2371,7 +2369,11 @@ def get_template_context(self):
         return {"default_value": self.default_value, **super().get_template_context()}
 
 
+############################
```
FarukOzderim, line 2372: Better borders 👍

FarukOzderim (commented):
LGTM, Good work!

ui/packages/app/src/components/Checkbox/Checkbox.svelte
FarukOzderim: @pngwn the default label is assigned like this:
`export let label: string = "Checkbox";`

I was wondering how do we assign the label given by the user

pngwn: These `export let` statement essentially marks the interface for the component. They are passed in from above : `<Checkbox label="some_label" />`. Same way arguments are marked on a function and passed in by the caller.

PR #1041, opened 2022-04-20
UI design update

gary149:


abidlabs:
LGTM, looked through a bunch of demos and they look so nice @gary149!

One small caveat, I had to resolve a merge conflict so that I could check with everything with the latest code on `main`: https://github.com/gradio-app/gradio/pull/1041/commits/20b3f75de27c70adfb694687d99cd1ccc6a182af

@pngwn if you can just confirm that the merge looks reasonable

PR #1046, opened 2022-04-20
[hotfix] HighlightedText

aliabid94:
Fix highlighted text component. Issue was not using $: in svelte component to recheck logic on value change.


abidlabs:
LGTM! Merging so that I can use it for the demo I'm working on

PR #1054, opened 2022-04-21
Form layout

pngwn:
Implements form groupings when there is more than 1 adjacent form. Part of #1041.

Run `demo/fake_gan_complex` to see this in action in a complex flow of components (with multiple foprms groups of different sizes). 
Run `demo/kitchen_sink` to see all forms together.

@gary149 Few implementation notes here. I added a new class to the `components` layer of tailwind to apply the new spacing. We had to switch to margins instead of `gap` because you can't switch off `gap` for certain elements. The lobotomised owl selector is pretty standard for child component spacing though and gives us the flexibility we need. Let me know what you think. 

We _do_ have some styling in the global css and some in the components themselves but since those styles are applied to components directly we still remain in complete control, so I think it should be ok.

**Screenshots**

Screenshot 2022-04-21 at 12 30 46
Screenshot 2022-04-21 at 12 46 27


ui/packages/app/src/Render.svelte
```diff
@@ -50,6 +94,7 @@
 >
 	{#if children && children.length}
 		{#each children as { component, id, props, children, has_modes }}
+			<!-- {@const type = get_form_type(i)} -->
```
pngwn, line 97: remove

ui/packages/form/src/CheckboxGroup.svelte
```diff
@@ -22,7 +23,10 @@
 </script>
 
 <fieldset
-	class="gr-box overflow-hidden border-solid border border-gray-200 gr-panel"
+	class="form gr-box overflow-hidden border-solid border border-gray-200 gr-panel"
+	class:!rounded-none={form_position === "mid"}
+	class:!rounded-b-none={form_position === "first"}
+	class:!rounded-t-none={form_position === "last"}
 >
```
pngwn, line 30: We should create a blocks variant for this fieldset, and the label should be wrapped in a `legend` tag.

pngwn: Actually, i think we could use `svelte:element` here. I'll create an issue to remind me. Low priority.

ui/packages/form/src/Dropdown.svelte
```diff
@@ -7,13 +7,14 @@
...
 	export let disabled: boolean = false;
+	export let form_position: "first" | "last" | "mid" | "single" = "single";
 
 	const dispatch = createEventDispatcher<{ change: string }>();
 
 	$: dispatch("change", value);
 </script>
 
-<Block>
+<Block {form_position}>
```
pngwn, line 17: We might be able to move this Block wrapping to the app and prevent some repetition by doing it in the `each` when we render the components. It would clean up the components quite a bit.

ui/packages/theme/src/global.css
```diff
@@ -2,6 +2,16 @@
...
+@layer components {
+	.gr-gap > * + * {
+		margin-top: 0.5rem;
+	}
+	.gr-form-gap > .form + .form {
+		margin-top: 0;
+		border-top: none;
+	}
+}
+
```
pngwn, line 14: @gary149 This is the new tailwind style i mentioned.

ui/packages/form/src/CheckboxGroup.svelte
pngwn: #1055

ui/packages/form/src/Dropdown.svelte
pngwn: #1055

ui/packages/atoms/src/Block.svelte
```diff
@@ -14,6 +15,11 @@
 <div
 	class="gr-box overflow-hidden {styles[variant]} {styles[color]}"
 	class:gr-panel={padding}
+	class:form={form_position}
+	class:!rounded-none={form_position === "mid"}
+	class:!rounded-b-none={form_position === "first"}
+	class:!rounded-t-none={form_position === "last"}
+	{form_position}
```
pngwn, line 22: remove

ui/packages/theme/src/global.css
gary149: @apply mt-0 border-t-0
For the same reasons.

ui/packages/theme/src/global.css
```diff
@@ -2,6 +2,16 @@
 @tailwind components;
 @tailwind utilities;
 
+@layer components {
+	.gr-gap > * + * {
+		margin-top: 0.5rem;
```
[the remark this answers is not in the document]
pngwn:
```suggestion
		@apply  mt-2;
```

ui/packages/theme/src/global.css
```diff
@@ -2,6 +2,16 @@
...
+	}
+	.gr-form-gap > .form + .form {
+		margin-top: 0;
+		border-top: none;
+	}
```
pngwn, line 12:
```suggestion
	.gr-form-gap > .form + .form {
		@apply mt-0 border-t-0;
	}
```

gary149 (approved):

pngwn:
@abidlabs I think it would be good to have a conversation around the narrative for interface + blocks and where those boundaries like. We _could_ say to users that for layouts like this then they should move to blocks. That said I think your proposed 'section' component would be really nice. It could take a title, aribtrary (and optional) markdown and a list of components. It could be collapsible, configurable via a flag.

PR #1056, opened 2022-04-21
remove-queue-from-events

FarukOzderim:
- enable_queue exists just under Blocks.launch()


abidlabs:
Seems reasonable to me, will let @aliabid94 review this one

aliabid94:
LGTM. I do wonder if there will be cases where a Blocks will have one event that takes a long time, and one that is instantaneous, and bc of queueing, the quick one has to wait behind the long one. We can wait and see if we have users with that problem

FarukOzderim:
Thanks for the quick review :)

Issue #680, opened 2022-02-19
[Blocks] Allowing users to adapt the parameters of one component based on another component


abidlabs:
A VERY common request. Here is the original discussion post:

> My interface is taking 2 inputs from the user. One using **Slider** & another using **Dropdown**. I want to change the options of dropdown depending upon the value selected using Slider. Then both of these values will be input to the calling function. How can I implement that functionality?

_Originally posted by @yesdeepakmittal in https://github.com/gradio-app/gradio/discussions/679_

abidlabs:
The way it might look for the user:

```python
from gradio.blocks import *

def dropdown_values(val):
    if val < 50:
       return ["fail", "repeat"]
    else:
       return ["succeed", "repeat"]
  
def fn(val, action):
     return "Your score is {} and you decided to {}".format(val, action)


with Blocks() as block:

  s = Slider(0, 100)
  d = Dropdown([]) 
  t = Textbox()

  s.select(fn=dropdown_values, outputs=d)
  d.select(fn=fn, inputs=[s, d], outputs=t)

  block.launch()
```

Basically every component (e.g. the Dropdown) should be able to be treated as an input or output

FarukOzderim:
I think this should not be in our priority, this should be in the 2nd next release :)

pngwn:
@abidlabs is this the same as #736 ?

abidlabs:
Yes it is, let me close this one. Thanks

PR #1057, opened 2022-04-21
spaces-defaults

### LINKED ISSUE #674: Cache examples by default on Spaces
Examples should be cached by default on spaces. Much better user experience. 

Ideally the way to do it would be by reading from an env variable that spaces can provide. 

We could also potentially cache them by default everywhere, but to avoid  slowing down the interface launch we'd want to cache as the first users click them instead of during the launch event.

FarukOzderim:
- enable_queue and cache_examples are activated in spaces in default option

fixes #674 #680

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -397,7 +397,10 @@ class Blocks(BlockContext):
         server_port (int): will start gradio app on this port (if available). Can be set by environment variable GRADIO_SERVER_PORT.
         server_name (str): to make app accessible on local network, set this to "0.0.0.0". Can be set by environment variable GRADIO_SERVER_NAME.
         show_tips (bool): if True, will occasionally show tips about new Gradio features
-        enable_queue (bool): if True, inference requests will be served through a queue instead of with parallel threads. Required for longer inference times (> 1min) to prevent timeout.
+        enable_queue (Optional[bool]):
+            if True, inference requests will be served through a queue instead of with parallel threads. Required for longer inference times (> 1min) to prevent timeout.
+            The default option in HuggingFace Spaces is True.
+            The default option elsewhere is False.
         width (int): The width in pixels of the iframe element containing the interface (used if inline=True)
         height (int): The height in pixels of the iframe element containing the interface (used if inline=True)
         encrypt (bool): If True, flagged data will be encrypted by key provided by creator at launch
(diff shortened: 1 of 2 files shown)


gradio/blocks.py
```diff
@@ -433,7 +433,10 @@ def launch(
                 getpass.getpass("Enter key for encryption: ")
             )
 
-        self.enable_queue = enable_queue
+        if self.is_space and enable_queue is None:
+            self.enable_queue = True
+        else:
+            self.enable_queue = enable_queue
```
abidlabs, line 439: What happens if `enable_queue` is not provided and you are *not* running in Spaces? Should explicitly make the default `False`

```suggestion
            self.enable_queue = enable_queue or False
```

gradio/interface.py
```diff
@@ -109,7 +109,7 @@ def __init__(
         outputs: str | Component | List[str | Component] = None,
         verbose: bool = False,
         examples: Optional[List[Any] | List[List[Any]] | str] = None,
-        cache_examples: bool = False,
+        cache_examples: Optional[bool] = None,
```
abidlabs, line 112: Same as above, let's update the docstring

gradio/interface.py
```diff
@@ -288,7 +288,6 @@ def clean_html(raw_html):
 
         self.thumbnail = thumbnail
         theme = theme if theme is not None else os.getenv("GRADIO_THEME", "default")
-        self.is_space = True if os.getenv("SYSTEM") == "spaces" else False
```
abidlabs, line 291: Good catch of duplication

gradio/interface.py
```diff
@@ -487,8 +486,11 @@ def clean_html(raw_html):
             if component.label is None:
                 component.label = "output_" + str(i)
 
-        self.cache_examples = cache_examples
-        if cache_examples:
+        if self.is_space and cache_examples is None:
+            self.cache_examples = True
+        else:
+            self.cache_examples = cache_examples
```
abidlabs, line 492: Same as above, consider

```suggestion
            self.cache_examples = cache_examples or False
```

abidlabs:
Looks good! Minor points above

gradio/interface.py
FarukOzderim: Wow it cache_examples was missing in the docstring :D

FarukOzderim:
@abidlabs thanks for the careful review!

PR #1060, opened 2022-04-21
Running, testing, and fixing demos

abidlabs:
Fixes include:
* For examples, using paths relative to the demo file, so that the demo can be run from anywhere
* Fixing parameter names (`default_value` instead of `default`)
* Making demos test more parameters/features (e.g. `placeholder`)


FarukOzderim (approved):
I like this usage a lot!
`os.path.join(os.path.dirname(__file__),"sax.wav"),`

Would you like to use this standard for the file usages within tests? That would enable running individual test functions or classes from IDE or different locations other the repo folder.

abidlabs:
> I like this usage a lot!
os.path.join(os.path.dirname(__file__),"sax.wav"),

> Would you like to use this standard for the file usages within tests? That would enable running individual test functions or classes from IDE or different locations other the repo folder.

Good idea -- can do that in a future PR

PR #1077, opened 2022-04-25
Scroll to output

### LINKED ISSUE #1055: Block wrapper handling
~~Two~~Three main points here:

- [x] We are wrapping _almost_ every component with `Block` inside the component istelf. This is okay but there is a lot of repetition. It might be able to do this at the app level, when we render the components from the config. This would simplify the components a little but would require a little refactoring. One to investigate rather than implement blindly.
- [x] `Radio` and `CheckboxGroup` use a `fieldset` element as their wrapper rather than the `Block` component, because they are a collection of form elements that exist together. This results in some duplication of code/classes, however small, to get consistent styling. We could pass a prop to `Blocks` and use `svelte:element` to switch between a `div` and a `fieldset`.
- [ ] Related to the final point, the form components that use a fieldset should use a `legend` element for their 'title'/'label' for accessibility reasons.

pngwn:
Implements a loader per component and scrolls the relevant output component into view when an API request has been triggered.

The behaviour has been implemented such that if there are multiple outputs for a given trigger then the component that is closest to the top of the page will be the target of the scroll-into-view functionality.

Additionally, the components have been refactored so that the `Block` wrapping is done in the APP rather than the component package. This allows for greater flexibility with app-specific styling and cleans up the components somewhat.

Not every component has a laoder as the standard laoder design doesn't work for smaller components.

In order to test, try the `fake_gan`, `chatbot`, and `kitchen_sink` demos. 

Workbench probably looks quite messy atm. Will clean that up later.

Closes #1055.


pngwn:
@gary149 @abidlabs This can be reviewed, test failures are just due to some code changes which I'll fix shortly.

abidlabs:
Very cooool! Love the loading indicator and the scrolling. 

As you mentioned, we don't have a loading indicator for the smaller components, such as `Textbox`. Are you planning on adding them @pngwn?

abidlabs:
cc @aliabid94 as well

pngwn:
@abidlabs Yes, but need to chat with @gary149 about those first. I think it is okay to merge this when we're happy and add the others in a follow up. I have some idea for _some_ form inputs but need to think about 'empty' styles + laoders for others.

abidlabs:
Whoops my bad @pngwn I pushed some changes to this branch when I thought I was working on a different branch. It's just small changes to make sure the kitchen sink and 2 other demos work when run from any path.

If you'd like, I can revert and move to a new branch?

abidlabs:
> @abidlabs Yes, but need to chat with @gary149 about those first. I think it is okay to merge this when we're happy and add the others in a follow up. I have some idea for some form inputs but need to think about 'empty' styles + laoders for others.

Sounds good. We should keep an issue related to the loaders open so that we make sure to get it in before releasing 3.0

pngwn:
Main reason to merge is that there a bunch of changes (refactors) in this branch that will be useful for other features + potentially lots of conflicts with other work that would happen in the meantime. Might take a few days to finalise some of the loaders.

The chatbot bloader in particular can't happen until the 'custom js' branch is in (and that will need some tweaks to work for that purpose).

pngwn:
And dw about the extra commit.

Issue #946, opened 2022-04-07
[blocks] optional labels for certain inputs


pngwn:
For most form elements, labels are required for accessibility reasons but for _outputs_ the case is less clear. They certainly need useful descriptions but since they are not interactive form elements, the accesibility considerations are more around readability.

We need  to figure out how/ where we can make certain labels optional without compromising end user usabilty or accessibility.

Continuation of #704.

abidlabs:
Hi @pngwn I was looking into this and it looks like labels needed to be provided for accessibility reasons, but they don't necessarily need to be _visible_:

> A label for a form control helps everyone better understand its purpose. In some cases, the purpose may be clear enough from the context when the content is rendered visually. The label can be hidden visually, though it still needs to be provided within the code to support other forms of presentation and interaction, such as for screen reader and speech input users. The method used in this tutorial to hide an element visually but make it available for assistive technologies is explained in [Note on hiding element](https://www.w3.org/WAI/tutorials/forms/labels/#note-on-hiding-elements).

Source: https://www.w3.org/WAI/tutorials/forms/labels/

I think it would be nice if people can make the labels invisible because sometimes they clutter up the UI or look out of place. What do you think about adding a "hide_label" boolean parameter to each of the form elements? False by default, but can be set to True, which makes the label invisible.

pngwn:
Works for me, tailwind has a class for that: https://tailwindcss.com/docs/screen-readers

PR #1080, opened 2022-04-25
Backend for optional labels

abidlabs:
Added a boolean parameter, `show_label` to every `Component`, based on our discussion in #946. Note that I've added it to every component instead of just the form components, to provide a more unified API across our components. 

We still need to add frontend support: if `show_label` is `True`, then the label should not be visible. However, we should still make it available for screen readers for accessibility purposes. @pngwn feel free to push to this branch direct.y.

Closes (the backend part of) #946


abidlabs:
I still need to fix the tests but since that is a very manual process, want to make sure we are all good with this prop

pngwn:
Then i think that needs to be changed as well. We can't do anything meanginful with `label`s for layout components for example or buttons. I think that made sense with the interface usecase but with the flexibility of blocks if becomes more problematic.

In the case of `Button`, we may well set the `label` as the button text but in that case we cannnot hide it for any reason either (we can only hide the whole button). This is why i think it should be a little more granular.

Could we not create a 'LabelledComponent' subclass that some components could extend from? Surely that would allow us to keep things clean while still having more granularity.

abidlabs:
That should work! Will implement

abidlabs:
@pngwn the backend is ready. I've created an `IOComponent` class that has methods and props specific to those components that can be used as inputs or outputs.

Feel free to merge or push to this branch.

Issue #841, opened 2022-03-18
HF Callback not working with Urdu Text


kingabzpro:
### Describe the bug

The call back works fine with vision and chatbot but with ASR it's generating error and I cant seem to figure it out. 

### Reproduction

Link to app: https://huggingface.co/spaces/kingabzpro/Urdu-ASR-SOTA
Link to dataset: https://huggingface.co/datasets/kingabzpro/Urdu-ASR-flags/tree/main

### Screenshot

![Uploading image.png…]()

### Logs

```shell
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/home/user/.local/lib/python3.8/site-packages/uvicorn/protocols/http/h11_impl.py", line 366, in run_asgi
    result = await app(self.scope, self.receive, self.send)
  File "/home/user/.local/lib/python3.8/site-packages/uvicorn/middleware/proxy_headers.py", line 75, in __call__
    return await self.app(scope, receive, send)
  File "/home/user/.local/lib/python3.8/site-packages/fastapi/applications.py", line 261, in __call__
    await super().__call__(scope, receive, send)
  File "/home/user/.local/lib/python3.8/site-packages/starlette/applications.py", line 112, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/home/user/.local/lib/python3.8/site-packages/starlette/middleware/errors.py", line 181, in __call__
    raise exc
  File "/home/user/.local/lib/python3.8/site-packages/starlette/middleware/errors.py", line 159, in __call__
    await self.app(scope, receive, _send)
  File "/home/user/.local/lib/python3.8/site-packages/starlette/middleware/cors.py", line 92, in __call__
    await self.simple_response(scope, receive, send, request_headers=headers)
  File "/home/user/.local/lib/python3.8/site-packages/starlette/middleware/cors.py", line 147, in simple_response
    await self.app(scope, receive, send)
  File "/home/user/.local/lib/python3.8/site-packages/starlette/exceptions.py", line 82, in __call__
    raise exc
  File "/home/user/.local/lib/python3.8/site-packages/starlette/exceptions.py", line 71, in __call__
    await self.app(scope, receive, sender)
  File "/home/user/.local/lib/python3.8/site-packages/fastapi/middleware/asyncexitstack.py", line 21, in __call__
    raise e
  File "/home/user/.local/lib/python3.8/site-packages/fastapi/middleware/asyncexitstack.py", line 18, in __call__
    await self.app(scope, receive, send)
[...]

PR #1081, opened 2022-04-25
Flagging fixes

abidlabs:
This PR fixes a few things currently broken with flagging:

* Only show flagging button if `allow_flagging == "manual"` (Closes: #1086)
* Fixes flagging audio files if they are loaded from examples (Closes: #841)

### DIFF
--- a/gradio/interface.py
+++ b/gradio/interface.py
@@ -548,6 +548,11 @@ class Interface(Blocks):
                     with Row():
                         if self.allow_flagging == "manual":
                             flag_btn = Button("Flag")
+                            flag_btn._click_no_preprocess(
+                                lambda *flag_data: self.flagging_callback.flag(flag_data),
+                                inputs=self.input_components + self.output_components,
+                                outputs=[],
+                            )                            
                         if self.interpretation:
                             interpretation_btn = Button("Interpret")
             submit_fn = (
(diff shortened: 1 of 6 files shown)


abidlabs:
Ready to be reviewed @FarukOzderim @aliabid94

gradio/interface.py
```diff
@@ -546,7 +546,15 @@ def clean_html(raw_html):
                     for component in self.output_components:
                         component.render()
                     with Row():
-                        flag_btn = Button("Flag")
+                        if self.allow_flagging == "manual":
+                            flag_btn = Button("Flag")
+                            flag_btn._click_no_preprocess(
+                                lambda *flag_data: self.flagging_callback.flag(
+                                    flag_data
```
FarukOzderim, line 553: Not sure about this, but wouldn't this be the correct version:

```suggestion
                                    *flag_data
```

gradio/routes.py
```diff
@@ -56,30 +56,6 @@ def render(self, content: Any) -> bytes:
 ###########
 
 
-class PredictBody(BaseModel):
```
FarukOzderim, line 59: Why did you remove these validations?

Better update them instead of removing.

FarukOzderim (commented):
LGMT other than the above points

gradio/routes.py
abidlabs: Good point, will update

gradio/interface.py
abidlabs: Nope `flagging_callback.flag()` takes a list for its first parameter, which is what `flag_data` is. `*flag_data` would unravel the list into separate parameters, which is not what we want here.

abidlabs:
Thanks @FarukOzderim for the review. Have added the response validation back

PR #1082, opened 2022-04-25
Custom JS calls

aliabid94:
Allows users to wrap JS functions in addition to regular python functions. "Clear" now uses this functionality.

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -60,6 +60,7 @@ class Block:
         preprocess: bool = True,
         postprocess: bool = True,
         queue=False,
+        js: bool = False,
         no_target: bool = False,
         status_tracker: Optional[StatusTracker] = None,
     ) -> None:
@@ -91,6 +92,7 @@ class Block:
                 "inputs": [block._id for block in inputs],
                 "outputs": [block._id for block in outputs],
                 "queue": queue,
+                "js": fn if js else False,
                 "status_tracker": status_tracker._id
                 if status_tracker is not None
                 else None,
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -277,6 +277,7 @@ class Textbox(Component):
         self.max_lines = max_lines
         self.placeholder = placeholder
         self.default_value = default_value
+        self.cleared_value = ""
         self.test_input = default_value
         self.interpret_by_tokens = True
         super().__init__(label=label, css=css, **kwargs)
@@ -398,6 +399,7 @@ class Textbox(Component):
         inputs: List[Component],
         outputs: List[Component],
         status_tracker: Optional[StatusTracker] = None,
+        _js=False,
     ):
         """
         Parameters:
@@ -408,7 +410,7 @@ class Textbox(Component):
         Returns: None
         """
         self.set_event_trigger(
-            "change", fn, inputs, outputs, status_tracker=status_tracker
+            "change", fn, inputs, outputs, status_tracker=status_tracker, js=_js
         )
 
     def submit(
@@ -417,6 +419,7 @@ class Textbox(Component):
         inputs: List[Component],
         outputs: List[Component],
         status_tracker: Optional[StatusTracker] = None,
+        _js=False,
     ):
         """
         Parameters:
@@ -427,7 +430,7 @@ class Textbox(Component):
         Returns: None
         """
         self.set_event_trigger(
-            "submit", fn, inputs, outputs, status_tracker=status_tracker
+            "submit", fn, inputs, outputs, status_tracker=status_tracker, js=_js
         )
 
 
@@ -546,6 +549,7 @@ class Number(Component):
         inputs: List[Component],
         outputs: List[Component],
         status_tracker: Optional[StatusTracker] = None,
+        _js=False,
     ):
         """
         Parameters:
@@ -556,7 +560,7 @@ class Number(Component):
         Returns: None
         """
         self.set_event_trigger(
-            "change", fn, inputs, outputs, status_tracker=status_tracker
+            "change", fn, inputs, outputs, status_tracker=status_tracker, js=_js
         )
 
     def submit(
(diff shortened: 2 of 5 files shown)


gradio/blocks.py
```diff
@@ -59,6 +59,7 @@ def set_event_trigger(
         outputs: List[Component],
         preprocess: bool = True,
         postprocess: bool = True,
+        js: bool = False,
```
abidlabs, line 62: Add to docstring

gradio/blocks.py
```diff
@@ -88,6 +89,7 @@ def set_event_trigger(
                 "trigger": event_name,
                 "inputs": [block._id for block in inputs],
                 "outputs": [block._id for block in outputs],
+                "js": fn if js else False,
```
abidlabs, line 92: Interesting, so this is either the serialized javascript function or False? 

Does line 85 need to be changed in the case of a javascript function?

gradio/components.py
```diff
@@ -277,6 +277,7 @@ def __init__(
         self.max_lines = max_lines
         self.placeholder = placeholder
         self.default_value = default_value
+        self.cleared_value = ""
```
abidlabs, line 280: Should `cleared_value` be a property in the backend? I feel like it would be cleaner if it was defined as a prop in the frontend.

gradio/interface.py
```diff
@@ -571,23 +572,18 @@ def clean_html(raw_html):
...
             clear_btn.click(
-                (
-                    lambda: [
-                        component.default_value
-                        if hasattr(component, "default_value")
-                        else None
-                        for component in self.input_components + self.output_components
-                    ]
-                    + [True]
+                f"""() => {json.dumps(
```
abidlabs, line 575: Why did you change the logic to go from `default_value` to `cleared_value`? How will this impact components like sliders that have a default value but not a clear value?

ui/packages/app/src/Blocks.svelte
```diff
@@ -29,6 +29,7 @@
 		inputs: Array<number>;
 		outputs: Array<number>;
 		queue: boolean;
+		js: string | null;
```
abidlabs, line 32: Should this be `string | bool` as it can be a string or False?

abidlabs:
Very nice. I tested it and it works well. Just left a few small comments above

gradio/interface.py
abidlabs: Actually it seems to work for slider (resets to default value) but I don't know why lol

abidlabs (approved):

ui/packages/app/src/api.ts
```diff
@@ -17,10 +17,22 @@ export const fn = async (
...
 	action: string,
+	js: string | false,
 	data: Record<string, unknown>,
 	queue: boolean,
 	queue_callback: (pos: number | null, is_initial?: boolean) => void
 ) => {
+	if (js !== false) {
+		console.log(js);
+		try {
+			var jsfn = eval(js);
```
pngwn, line 28: We should do this up front when we are parsing the config rather than doing ti everytime we make a call. The sooner we can treat this as a normal function the easier our lives will be.

Additionally, [`new Function()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function) is preferable to `eval` as it is faster and has safer scoping rules.

Something like this shoudl give the same result

```js
const jsfn = new Function(`return ${js}`)();
```

pngwn (commented):
This is looking great.

In addition to the comment I left about evaluating the code when processing the config and adjusting the implementation, this implementation has one important omission: the ability to update state.

The function needs to be passed the data from the input components, as it currently is, but it should also be passed the current state of the oputput components. In addition, the return value (if there is one) should become the new state for the output components.

So to clarify, i think the following needs to happen:

- function has two parameters
  - The value of the input components (same as API payload)
  - The current state of the output components (as single item or array, depending on quantity)
- If there is a return value then that should be used to set the state of the output components immediately.
- If there is no return value then no local state will be changed. This means we will not be able to assign an undefined value to a single output but i think that is a reasonabel tradeoff.

Something like this:

```js
// input is a single textbox, output is a chatbot
function js(data, state) {
  return [...state, [data, ""]]
}
```
This would immediately update the state of the chatbot

In python we'd then do something like `tail(history).push(repsonse)` rather than adding both the message _and_ the response.

Question I have regarding state updates is what happens if the returned value doesn't come in the expected shape. Say we have a list iof outputs but the returned value is a single item. Or what if the list of outputs and the list of return values is not equal? We could technically validate some of these values too if needed but i think we should do that later, if we decide it makes sense.

pngwn:
Okay, I've slightly misread this. This allows an alternative implementation to the API request. I think this needs to be _in addition_ to the API request.

pngwn:
> if we need the state of the output component for whatever reason, can't we simply include the output component as one of the input components? I don't think we don't need to pass in the state of the output as a special parameter

We could do, i just think it is confusing and annoying. In my mind the main (only?) usecase for this is updating component state prior to or instead of an API request. Passing the current state just seems like a sensible convenience rather than making people add the output as an input (which will then require them to set the output component to `interactive=false` manually). 

Almost every state management mechanism passes both the event payload and the existing state by default because people typically need the ability to compute state based on the current state and the state of the payload.

pngwn:
Another note, we need to ensure that any API calls are sent off _before_ we manipulate state via JS otehrwise we could end up with nasty race conditions. This could be controllable via an option.

gradio/interface.py
aliabid94: Because some components, like a textbox, should clear when clear is pressed, not reset to the default value. For components like Slider, the cleared_value is the default value.

gradio/blocks.py
aliabid94: Now it's just a string function.

ui/packages/app/src/api.ts
aliabid94: Done. Instead of combining input data in one list called data and output data in one arg called state, I spread the args to match the style used in python - the output state values come right after input arguments.

aliabid94:
All requested changes made @pngwn, take a look at blocks_js_methods/run.py for an example.

Issue #784, opened 2022-03-07
[Blocks] Handling visibility and/or gating for multi-step interfaces


abidlabs:
We've heard a lot of demand for multi-step interfaces, i.e. where the user first specifies one input, then function runs, and then the user picks a second input, and so on. As part of these interfaces, we need to give the user control over what components are visible at any point. 

**Motivation:** 

As @Lastly mentions in #690:

> [In a multi-step demo, we want to] ensure that step 1 is completed before letting the user engage with step 2.

If we simply hide the step 2 components until step 1 is complete, this would solve @Lastly's issue.

Furthermore, @prakharrathi25 mentions that what components are shown could be input-dependent:

> for example there can be a checkbox which lets me decide which components I want to show at a time:

> ```python
> if checkbox1 == True: 
>     show_text_input("Name?")
> ...
> ```

**Implementation suggestion**

* One super simple idea to make this work would be to hide a component if its value is set to **None**. We already do this with output components with the `Interface` class. If a function returns `None` for any output component, it is automatically hidden. But (1) Will this be possible after we merge input and output components? (2) Will it be annoying for users who have multiple (more than 2) steps?

abidlabs:
This is now possible! With #1089

PR #1089, opened 2022-04-26
Update component config

aliabid94:
Allow users to update component config parameter with `gr.update()` or with component `gr.Textbox.update()`. Changed backend to provide type hints and validation so users can't pass bad parameters. 

Valid formats:
- `gr.update(lines=5)`
- `gr.Textbox.update(lines=5)`
- `gr.Textbox.update(lines=5, value='hello')`

**See blocks_update/run.py for example**
Recording 2022-05-09 at 14 31 45

Also:
- Added support for visibility changes.
- Renamed get_template_context to get_config.


abidlabs:
Can we have the behavior that setting the `value` of a component to `None` hides the component? This is already the case with some components (like `Label` I think) and it would be good to have uniform behavior across the components and provide a simple way of addressing #784

abidlabs:
> Can we have the behavior that setting the value of a component to None hides the component? This is already the case with some components (like Label I think) and it would be good to have uniform behavior across the components and provide a simple way of addressing https://github.com/gradio-app/gradio/issues/784

I talked about this some more with @aliabid94 and we felt it would be better to actually just have a property for every component called `visible` that you could change using this approach

gradio/components.py
```diff
@@ -75,8 +56,10 @@ def __init__(
         *,
         label: Optional[str] = None,
         show_label: bool = True,
-        requires_permissions: bool = False,
         interactive: Optional[bool] = None,
+        css: Optional[Dict] = None,
+        visible: bool = True,
```
abidlabs, line 61: Is this needed? Since the parent `Block` class already defines a `visible` parameter int he constructor

gradio/components.py
```diff
@@ -444,15 +433,46 @@ def __init__(
...
+    @staticmethod
+    def update(
+        value: Optional[Any] = None,
+        lines: Optional[int] = None,
+        max_lines: Optional[int] = None,
+        placeholder: Optional[str] = None,
+        label: Optional[str] = None,
+        show_label: Optional[bool] = None,
+        css: Optional[Dict] = None,
+        visible: Optional[bool] = None,
```
abidlabs, line 464: You don't have `interactive` as a parameter in any of the `update()` methods -- is that intentional?

gradio/components.py
```diff
@@ -444,15 +433,46 @@ def __init__(
...
             "max_lines": self.max_lines,
             "placeholder": self.placeholder,
             "default_value": self.default_value,
-            **IOComponent.get_template_context(self),
+            **IOComponent.get_config(self),
+        }
+
+    @staticmethod
+    def update(
+        value: Optional[Any] = None,
```
abidlabs, line 457: There is a discrepancy between the name of the parameter in the constructor `default_value` and `value`. Can we set both to be `value`?

gradio/components.py
```diff
@@ -898,13 +1023,40 @@ def __init__(
...
         return {
             "choices": self.choices,
             "default_value": self.default_value,
-            **IOComponent.get_template_context(self),
+            **IOComponent.get_config(self),
+        }
+
+    @staticmethod
+    def update(
+        value: Optional[Any] = None,
```
abidlabs, line 1045: I assume `value` updates `default_selected` but this is not clear. I think we should just change the name of the first parameter in every component to be `value`

abidlabs:
Tests need to be updated + see my comments above.

Other than that, works amazing!

abidlabs (approved):

pngwn (approved):
Looks great 👍

gradio/components.py
aliabid94: Yes, because we want type hints for everything.

gradio/components.py
FarukOzderim: I think update method should just follow/respect component's parameters.
Some has `default_value`, some has `choices `and some has no default_value. Furthermore IMO name of default_value is another topic, and not necessary for now to change.

abidlabs: Yeah parameters in `update()` should be consistent with constructor parameters.

But we should settle `default_value` vs `value` now I think so that we don't have to worry about backwards compatibility. Imo `value` is a better name than `default_value` if you consider the ways this parameter is used:
* constructing a component in `interactive` mode (e.g. input): here either `value` or `default_value` make sense as this is the original value of the component. E.g. the [text in a HTML textbox](https://www.w3schools.com/jsref/prop_text_value.asp) is called "value"
* constructing a component in  `non-interactive` mode (e.g. output): here `default_value` doesn't make sense, as this value cannot be changed -- it is simply the value that is displayed
* updating a component: again, `default_value` is not ideal because you are updating this value

Not to mention `value` is a shorter name!

demo/blocks_update/run.py
```diff
@@ -1,17 +1,44 @@
...
+            choices=species_map[species], value=species_map[species][1]
+        ), gr.update(visible=True)
+
+    species.change(filter_species, species, [animal, details_col])
+
+    def filter_weight(animal):
+        if animal in ("Elephant", "Shark", "Giraffe"):
+            return gr.update(maximum=100)
+        else:
+            return gr.update(maximum=20)
```
FarukOzderim, line 37: I am not very fond of `gr.update` and I don't think it is intutive. WDYT of just having `gr.Dropdown.update` syntax? This would decrease the error occurences, and would make the code more intuitive. Furthermore it would make following/respecting component's field names easier.

gradio/blocks.py
```diff
@@ -179,23 +181,30 @@ def __init__(
...
+        }
+
+    @staticmethod
+    def update(
+        variant: Optional[str] = None,
+        css: Optional[Dict] = None,
+        visible: Optional[bool] = None,
+    ):
+        return {
+            "variant": variant,
```
FarukOzderim, line 200: Do we send `None` values to the Frontend, if I remember correctly we were removing them from the dicts. @abidlabs I remember like you had some PR about this in the past.

gradio/components.py
```diff
@@ -2305,13 +2715,18 @@ def __init__(
...
+        show_label: bool = True,
         css: Optional[Dict] = None,
+        visible: bool = True,
         **kwargs,
     ):
         """
         Parameters:
         default (str): IGNORED
-        label (str): component name in interface.
+        label (Optional[str]): component name in interface.
```
FarukOzderim, line 2726: We deprecated the component, but we are still updating it :D 

Shall we move deprecated components under another file?

FarukOzderim (approved):
Good work!

gradio/components.py
FarukOzderim: My main problem is `value` does not map to initial value or default value in my mind. That might just for me, if you think it will translate into the initial value in the community, it can be way to go.

And in all 3 cases `default_value` makes sense for me, 
2. non-interactive has just default value
3. we are updating the default_value of the component

demo/blocks_update/run.py
```diff
@@ -1,17 +1,44 @@
...
+    }
 
-    s.change(fn=update_dropdown_choices, inputs=[s], outputs=[d])
-    block.launch()
+    def filter_species(species):
+        return gr.Dropdown.update(
+            choices=species_map[species], value=species_map[species][1]
+        ), gr.update(visible=True)
+
+    species.change(filter_species, species, [animal, details_col])
```
FarukOzderim, line 31: I have an idea which will unlock more opportunities with the update functionality, but might make things a little bit more complex from the development POV.

what if we send components in to the function and update component's field over there?
```

    def filter_species(species, dropdown, column):
        return (dropdown.update(
            choices=species_map[species], value=species_map[species][1]
        )
        , column.update(visible=True))
      
       species.update_on_change(filter_species, [species, animal, details_col], [animal, details_col])
```

We would need to distinguish between sending values inside the component and sending the component itself. But this would enable better intution, and fetching a Component's fields and updating component's fields according to the previous fields.

gradio/blocks.py
aliabid94: removed them with utils.delete_none

demo/blocks_update/run.py
aliabid94: This is a pretty drastic change, and I'm not sure if it makes the API cleaner. I think there are pros/cons to it but maybe discussion for after 3.0?

gradio/components.py
aliabid94: Removed.

demo/blocks_update/run.py
aliabid94: I think it's nice for a quick shortcut. It will fail if any invalid argument is passed to it because gr.update passes to gr.Dropdown.update right away.

FarukOzderim:
Created the different approach suggestion in #1211 you could drop your thoughts there. @aliabid94

adamwuyu:
Hi @aliabid94, how do you realize click "fish" and more UI element shows up? I really need this kind of interaction, can you give me a hint or piece of code? Thanks a lot.

Issue #1093, opened 2022-04-26
Article parameter not rendered on Interface beta 


radames:
### Describe the bug

I know with Blocks there's a different and better approach to build interactive and non-interactive elements (text, descriptions, etc). Are we going the be backwards compatible?  Right now working on `2.9.0b8` the property `article` is not rendered on the [`Interface Class`](https://github.com/gradio-app/gradio/blob/910db9959b478c620fbbee84112ec0f97d9655f5/gradio/interface.py#L511-L519)

### Reproduction

```python

input = gr.Image()
output = gr.Image()

gr.Interface(
    description="## Description",
    article="## Article",
    fn=classify_image,
    inputs=input,
    outputs=output,
    ).launch()

```

### System Info

```shell
Gradio 2.9.0b8
```

### Severity

annoyance

abidlabs:
Ahh it looks like we just missed this parameter when we were rewriting `Interface`. But no worries, it should be an easy fix. @aliabid94 would you be able to take a look?

abidlabs:
Yes the goal is to be fully backwards compatible @radames!

Issue #960, opened 2022-04-08
[Blocks] Support Blocks components as Plug and Use components


FarukOzderim:
The design of **Blocks** does not support multiple **Blocks** defined outside the main **Blocks** scope. But this is needed for complex **Blocks** components as it will enable users to package their modules under smaller **Blocks**

We can potentially support it with a new design,
- either a workaround function that passes every component, event and layouts inside the smaller Blocks to Parent Blocks
- or have smaller Blocks as a component inside the Bigger Blocks with its config (this will need changes within frontend)

**Usage example**:
```
import gradio as gr

import random

xray_model = lambda diseases, img: {disease: random.random() for disease in diseases}
ct_model = lambda diseases, img: {disease: 0.1 for disease in diseases}

with gr.Blocks() as demo_1:
    disease = gr.CheckboxGroup(
        choices=["Covid", "Malaria", "Lung Cancer"], label="Disease to Scan For"
    )
    with gr.Tabs():
        with gr.TabItem("Blocks2"):
            with gr.Row():
                xray_scan = gr.Image()
                xray_results = gr.JSON()
            xray_run = gr.Button("Run")
            xray_run.click(
                xray_model, inputs=[disease, xray_scan], outputs=[xray_results]
            )

with gr.Blocks() as demo_2:
     with gr.Row():
            ct_scan = gr.Image()
            ct_results = gr.JSON()
            ct_run = gr.Button("Run")
            ct_run.click(ct_model, inputs=[disease, ct_scan], outputs=ct_results)


with gr.Blocks() as demo:
    gr.Markdown(
        """
# Detect Disease From Scan
With this model you can lorem ipsum
- ipsum 1
- ipsum 2
"""
    )
    disease = gr.CheckboxGroup(
        choices=["Covid", "Malaria", "Lung Cancer"], label="Disease to Scan For"
    )

    with gr.Tabs():
        with gr.TabItem("X-ray"):
            demo_1.render()
        with gr.TabItem("CT Scan"):
            demo_2.render()
    overall_probability = gr.Textbox()

if __name__ == "__main__":
    print(demo.get_config_file())
    demo.launch()

```

abidlabs:
The problem is with the `render()` function. It just needs to be tweaked -- I can take a look into this

FarukOzderim:
This functionality is not supported with render yet, just shared an example outline, there could be a different function called x as well.

abidlabs:
[...]

PR #1102, opened 2022-04-27
Blocks rendering fix

abidlabs:
* Fixed Blocks so that it can be used inside other blocks as plug-and-play components (Closes: #960)

* Added a demo to illustrate usage: `demo\blocks_plug\run.py`

* Also adds `article` parameter back into `Interface` (Closes: #1093)


demo/blocks_plug/run.py
```diff
@@ -0,0 +1,30 @@
+import gradio as gr
+
+text_identity, text_input, text_output = gr.Blocks(), gr.Blocks(), gr.Blocks()
```
FarukOzderim, line 3: I think we can rename inner demos better to clarify the message.
```suggestion
identity_demo, input_demo, output_demo= gr.Blocks(), gr.Blocks(), gr.Blocks()
```

gradio/blocks.py
```diff
@@ -19,24 +19,43 @@
...
         self.events = []
 
+    def unrender(self):
+        """
+        Removes self from BlockContext if it has been rendered (otherwise does nothing).
+        Only deletes the first occurrence of self in BlockContext. Removes from the
+        layout and collection of Blocks, but does not delete any event triggers.
+        """
+        self._id = Context.id
+        Context.id += 1
```
FarukOzderim, line 46: Why increase the id within removal?

FarukOzderim (commented):
LGTM, would be better if @aliabid94 takes a look as well.

gradio/blocks.py
```diff
@@ -19,24 +19,43 @@
...
         """
-        self._id = Context.id
-        Context.id += 1
         if Context.block is not None:
             Context.block.children.append(self)
         if Context.root_block is not None:
             Context.root_block.blocks[self._id] = self
         self.events = []
 
+    def unrender(self):
```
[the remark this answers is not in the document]
FarukOzderim: What would happen if a component is used in 2 different interfaces.

textbox = gr.Textbox()
with blocks_1:
  demo_1 = Interface(inputs = textbox)
  demo_2 = Interface(inputs = textbox)

aliabid94:
Discussed with @abidlabs earlier, but instead of unrender method, get_component_shortcuts should just create Components with render=False in cunstructor. Should remove the need for unrender. 
LGTM otherwise.

demo/blocks_plug/run.py
abidlabs: Sounds good, will do

gradio/blocks.py
abidlabs: My bad

gradio/interface.py
```diff
@@ -198,8 +198,8 @@ def __init__(
             inputs[inputs.index("state")] = state_variable
             outputs[outputs.index("state")] = state_variable
 
-        self.input_components = [get_component_instance(i) for i in inputs]
-        self.output_components = [get_component_instance(o) for o in outputs]
+        self.input_components = [get_component_instance(i).unrender() for i in inputs]
```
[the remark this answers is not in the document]
abidlabs: Not sure I follow what you mean

gradio/blocks.py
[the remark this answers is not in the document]
abidlabs: I tested the following code:

```py
textbox = gr.Textbox()

with gr.Blocks() as blocks:
    demo_1 = gr.Interface(lambda x:x, inputs=textbox, outputs="text")
    demo_2 = gr.Interface(lambda x:x, inputs=textbox, outputs="number")

    blocks.launch()
```

And it works fine (although the input textboxes are linked in an interesting way):  

image

abidlabs:
I discussed some more with @aliabid94 and we will actually keep `unrender()`. We need to handle cases like this:

```py
import gradio as gr

blocks = gr.Blocks()
with blocks:    
    gr.Interface(
      fn=lambda x: x,
      inputs=gr.Image(),
      outputs=gr.Image())
blocks.launch() 
```

gradio/interface.py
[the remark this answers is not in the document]
FarukOzderim: oh I misspelled as render, it will be unrender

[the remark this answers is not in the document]
FarukOzderim: some of the components inside input/output are just string shortcuts("text"), and we shall not `unrender` them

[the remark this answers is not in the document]
FarukOzderim: Well now I read the discussions below, this point is not correct at all.

gradio/components.py
```diff
@@ -3100,7 +3099,7 @@ def __init__(
         label (str): component name
```
FarukOzderim, line 3099: need to update docstring with the label's removal

gradio/components.py
```diff
@@ -3193,11 +3192,10 @@ def __init__(
         components: List[Component],
         samples: List[List[Any]],
         type: str = "values",
-        label: Optional[str] = None,
         css: Optional[Dict] = None,
         **kwargs,
     ):
```
FarukOzderim, line 3197: not related to this PR but, Dataset has no docstring, let's add one?

gradio/components.py
```diff
@@ -3274,11 +3272,10 @@ def __init__(
         self,
         component: Component,
         *,
-        label: Optional[str] = None,
         css: Optional[Dict] = None,
         **kwargs,
     ):
```
FarukOzderim, line 3277: not related to this PR but, Interpretation has no docstring, let's add one?

Issue #1007, opened 2022-04-15
Add integer type for `gradio.Number`


joowon-dm-snu:
**Is your feature request related to a problem? Please describe.**

Code below raise `TypeError: can't multiply sequence by non-int of type 'float'`
```python
def main(n_samples):
    ...
    uc = model.get_learned_conditioning(n_samples * [""])
    ...

iface = gr.Interface(
    fn=main,
    inputs=[
        gr.inputs.Number(default=200),
    ]
    ...
```

For solving that issue, I needed to change `def main` like below 

```python
def main(n_samples):
    ...
    n_samples = int(n_samples)
    uc = model.get_learned_conditioning(n_samples * [""])
    ...
```

**Describe the solution you'd like**
```python
iface = gr.Interface(
    fn=main,
    inputs=[
        gr.inputs.Number(default=200, integer=True),
    ]
    ...
```

**Additional context**

FarukOzderim:
Hello @joowon-dm-snu, thanks for reporting.
If you check the Number component, it supports float not int.

abidlabs:
I think what @joowon-dm-snu is asking here is for us to have a parameter that automatically preprocesses the Number input into an integer. That's not a bad idea -- why did we close this?

FarukOzderim:
Sorry my bad, misunderstood the issue.

I feel negative about this, what would happen when integer conversion is unsuccessful, throw an error?
Wouldn't it be better for user to convert to int when he/she is going to use it?

joowon-dm-snu:
Issue was for feature-request not for error report :) Thanks @abidlabs .

@FarukOzderim is there any possibility that `float -> int` conversion throw error?
I needed to change `argparse.ArgumentParser()` styled [code](https://github.com/CompVis/latent-diffusion/blob/main/scripts/txt2img.py) to `gradio` and face this issue that I need to all integer parameters by using code below. 
```python
def foo(..., ddim_steps, n_iter, H, W, n_samples, ...):
    ...
    ddim_steps = int(ddim_steps)
    n_iter= int(n_iter)
    H= int(H)
    W = int(W)
    n_samples = int(n_samples)
    ...
```
I still fill this is pretty redundant and hope `gradio` provide more easier approach for this.
Anyway its just feature proposal so skip it if this is not appropriate for future implements :)

FarukOzderim:
Thanks for the suggestion.

For example it can throw an error when it is converting 3.14 to integer

abidlabs:
[...]

PR #1015, opened 2022-04-18
Website Design Changes

abidlabs:
Changes to gradio.app
* New landing page design 
* New gallery page showing popular Gradio Spaces in different categories
* New docs: 
  * #675
  * #948
  * #949
  * #950
  * #951
  * #952
  * #953

PR #1125, opened 2022-04-28
Add precision to Number, backend only

freddyaboulton:
# Description

Adds an `integer` flag to `Number` so that only integers are allowed in the front end. Opening a draft before going further to see if this is the right approach.

Does it make sense to introduce a new component class for `Integer` as opposed to adding a parameter to `Number`? I think we'll have to also modify the `get_interpretation_neighbors` to only generate integers if `integer=True` so it might justify it's own class. 

 
integer_demo

Closes: #1007 

# Checklist:


ui/packages/app/src/components/Integer/Integer.svelte
```diff
@@ -0,0 +1,33 @@
+<script lang="ts">
```
freddyaboulton, line 1: This pattern may not be sustainable if we also want to introduce more numeric types in the future like natural number. Pretty new to svelte so this is the first thing I thought of to get the desired behavior.

gradio/components.py
```diff
@@ -462,7 +467,7 @@ def preprocess(self, x: float | None) -> Optional[float]:
         """
         if x is None:
             return None
-        return float(x)
+        return int(x) if self.integer else float(x)
```
FarukOzderim, line 470: Let's update the docstring and type hint.

gradio/components.py
```diff
@@ -472,7 +477,7 @@ def preprocess_example(self, x: float | None) -> float | None:
         if x is None:
             return None
         else:
-            return float(x)
+            return int(x) if self.integer else float(x)
```
FarukOzderim, line 480: Let's update the docstring and type hint.

gradio/components.py
```diff
@@ -523,7 +528,7 @@ def postprocess(self, y: float | None):
         if y is None:
             return None
         else:
-            return float(y)
+            return int(y) if self.integer else float(y)
```
FarukOzderim, line 531: Let's update the docstring and type hint.

gradio/components.py
```diff
@@ -448,12 +449,16 @@ def __init__(
         self.default_value = float(default_value) if default_value is not None else None
         self.test_input = self.default_value if self.default_value is not None else 1
         self.interpret_by_tokens = False
+        self.integer = integer
         super().__init__(label=label, css=css, **kwargs)
 
     def get_template_context(self):
         return {"default_value": self.default_value, **super().get_template_context()}
```
FarukOzderim, line 456: lets add integer into the to context dictionary

ui/packages/app/src/components/Integer/Integer.svelte
FarukOzderim: I don't think we need and Integer component in frontend, we can make use of `Number.integer` using its `get_template_context` and adopt the Number component's input filtering in Frontend according to that.
cc: @abidlabs @aliabid94

FarukOzderim (changes requested):
Good work!

aliabid94:
I think it would be better to just add a kwarg to Number, such as `type="integer"`. Otherwise there would be too much duplicated code

ui/packages/app/src/components/Integer/Integer.svelte
FarukOzderim: Instead of creating a new frontend component, we can just utilize `get_template_context` function.

freddyaboulton: Thank you for the pointer. Will look into this!

abidlabs:
Very cool, thanks for opening the PR @freddyaboulton! It would be great to have a way of having a way of integer-specific inputs. I agree with @aliabid94 that this is better done through a parameter in the existing `Number` component as opposed to creating an entirely new component.

> Does it make sense to introduce a new component class for Integer as opposed to adding a parameter to Number? I think we'll have to also modify the get_interpretation_neighbors to only generate integers if integer=True so it might justify it's own class.

We can modify `get_interpretation_neighbors()` to check the `self.type` and modify behavior accordingly

gradio/components.py
```diff
@@ -489,14 +498,21 @@ def set_interpret_parameters(
...
-        x = float(x)
+    def get_interpretation_neighbors(self, x: float | int) -> Tuple[List[float], Dict]:
+        x = int(x) if self.integer else float(x)
         if self.interpretation_delta_type == "percent":
             delta = 1.0 * self.interpretation_delta * x / 100
         elif self.interpretation_delta_type == "absolute":
             delta = self.interpretation_delta
         else:
             delta = self.interpretation_delta
+        if self.integer and math.floor(delta) != delta:
```
freddyaboulton, line 509: I thought it was important to prevent users from trying to generate invalid neighbors if type="integer"

gradio/components.py
freddyaboulton: Done!

gradio/components.py
freddyaboulton: Done!

gradio/components.py
freddyaboulton: Done!

pngwn:
Instead of `type='integer/float'`, what about a `precision` (can bikeshed the name) kwarg instead that toojk a number. That would cover this case but also allow more more flexibility with floats. `0` would be an int but you'd also be able to pass any arbitrary non-negative number to clamp values using fixed-point notation.

@abidlabs @aliabid94 @freddyaboulton

freddyaboulton:
@pngwn Cool idea! I guess it's a product question. Originally I thought users would be more interested in excluding certain kinds of numbers in their app, (like negatives, floats) and so the current design would be better suited for that (`type='nonnegative'`) but users probably care more about properly formatting numeric input in the app. 

I like your idea. Before moving forward, I'd like to hear from other members of the team to make sure we're aligned. One clarifying point -  if we move forward I'm thinking we can set the default of `precision` to `None` (no rounding) to preserve the current behavior?

pngwn:
I don't think negatives and floats are in the same category and it wouldn't make sense to flag them via the same kwarg because you might want negative floats or negative ints. The we'd need to pass multiple options to the same kwarg. min/max values would be more suited to that task imo.

Be good to get the thoughts of others though.

pngwn:
@freddyaboulton 
> One clarifying point - if we move forward I'm thinking we can set the default of precision to None (no rounding) to preserve the current behavior?

Yes, I think that would make sense.

abidlabs:
Thanks for thinking this through with us, @freddyaboulton. I do think it makes sense to have two different (sets of) parameters: `precision`, which is essentially used to round input to a certain decimal point or to an integer, and then `min`/`max`, which is used to restrict the input range and is enforced by the frontend. 

Let's start with the `precision` parameter since it is just a backend change and would address the concerns in #1007. Summarizing the discussion above, it has the following behavior:
* `precision=None` (default): does not round the input number at all
* `precision=0`: rounds the float to the nearest `int` and changes the type to `int`
* `precision=n` for a positive integer `n`: rounds the float to the specified number of decimal points and leaves it as a float

freddyaboulton:
Thank you for the summary @abidlabs ! Sorry for the delay, I just implemented the `precision` parameter as described above and backed out the front-end changes I had made.

gradio/components.py
```diff
@@ -652,6 +655,26 @@ def __init__(
...
         )
 
+    @staticmethod
+    def round_to_precision(num: float | int | None, precision: int | None):
+        """Round to a given precision.
+        If precision is None, no rounding happens. If 0, num is converted to int.
+
+        Parameters:
+        num (float | int): Number to round.
+        precision (Optional[int]): Precision to round to.
```
FarukOzderim, line 665: Let's have definition and docstring the same
```suggestion
        precision (int | None): Precision to round to.
```

gradio/components.py
```diff
@@ -652,6 +655,26 @@ def __init__(
             **kwargs,
         )
 
+    @staticmethod
+    def round_to_precision(num: float | int | None, precision: int | None):
+        """Round to a given precision.
```
FarukOzderim, line 660: Let's follow the codebase
```suggestion
        """
        Round to a given precision.
```

gradio/components.py
```diff
@@ -652,6 +655,26 @@ def __init__(
...
+        precision (Optional[int]): Precision to round to.
+        Returns:
+        (float | int): rounded number
+        """
+        if num is None:
+            return None
+        if precision is None:
+            return float(num)
+        elif precision == 0:
+            return int(round(float(num), precision))
```
FarukOzderim, line 674:
```suggestion
            return int(round(num, precision))
```

gradio/components.py
```diff
@@ -652,6 +655,26 @@ def __init__(
...
+        else:
+            return round(float(num), precision)
```
FarukOzderim, line 676:
```suggestion
            return round(num, precision)
```

gradio/components.py
```diff
@@ -674,26 +697,26 @@ def update(
             "__type__": "update",
         }
 
-    def preprocess(self, x: float | None) -> Optional[float]:
+    def preprocess(self, x: int | float | None) -> int | float | None:
         """
         Parameters:
         x (string): numeric input as a string
```
FarukOzderim, line 703:
```suggestion
        x (int | float | None): numeric input as a string
```

gradio/components.py
```diff
@@ -710,14 +733,21 @@ def set_interpret_parameters(
...
+        x = self.round_to_precision(x, self.precision)
         if self.interpretation_delta_type == "percent":
             delta = 1.0 * self.interpretation_delta * x / 100
         elif self.interpretation_delta_type == "absolute":
             delta = self.interpretation_delta
         else:
             delta = self.interpretation_delta
+        if self.precision == 0 and math.floor(delta) != delta:
+            raise ValueError(
+                f"Delta value {delta} is not an . Cannot generate valid set of neighbors. "
```
FarukOzderim, line 746: typo probably:
`
f"Delta value {delta} is not an .`

FarukOzderim (changes requested):
Good work with the code and the tests!

FarukOzderim:
@abidlabs @aliabid94 I think we can add this feature to 3.0?

gradio/components.py
```diff
@@ -638,8 +639,10 @@ def __init__(
         label (Optional[str]): component name in interface.
         show_label (bool): if True, will display label.
         visible (bool): If False, component will be hidden.
+        integer (bool): Whether the field should only contain integer values. Defaults to False.
```
abidlabs, line 642: This parameter should be `precision` instead of `integer`

abidlabs:
Looks great @freddyaboulton -- @FarukOzderim and I pointed out a few minor changes but we should be good to merge after those are in.

Note that we also have a big PR that we are trying to merge in (#1015) and depending on if that gets merged in first, we might have some conflicts (should be pretty minor, mostly just documentation).

gradio/components.py
freddyaboulton: Thank you for pointing this out!

freddyaboulton:
Thank you for the reivews @FarukOzderim and @abidlabs ! I addressed those changes. Will keep an eye on #1015 and rebase if necessary.

abidlabs (approved):

abidlabs:
LGTM, ready to merge. @FarukOzderim can you approve and merge it in?

abidlabs:
Thank you for the nice PR and willingness to adapt the solution as we thought about it together @freddyaboulton!

Issue #1120, opened 2022-04-28
Queue is not working on spaces


FarukOzderim:
### Describe the bug

Report from @osanseviero:

I have this demo with two interfaces within a block, but I think it is dying after 60s (I don't see anything else in logs). I already have enable_queue in the block launch method. Should I have done anything else? Queing might not be enabled in blocks yet.
This same space with Interface was working alright (without using blocks)

### Reproduction

https://huggingface.co/spaces/osanseviero/tortoisse-tts

### System Info

```shell
Spaces
```

### Severity

critical

FarukOzderim:
Testing it locally with printouts, it successfuly runs the loop inside `queue_thread` however the interface does not load.
![image](https://user-images.githubusercontent.com/44067394/165955616-bc7c9ffb-57f5-4f5d-8a5f-80034cc82a50.png)

FarukOzderim:
Btw when enable_queue=True, I cannot exit python with ctrl+C command, the background thread continues to work.

PR #1136, opened 2022-04-29
Refactor queuing

FarukOzderim:
There was not much to dig around with #1120, I created a suggestion for queuing instead of trying to debug it, since we are going to refactor it soon.

Migrated to async wait as well, since we are now on a async framework.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report_template.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report_template.yml
@@ -6,6 +6,13 @@ body:
     attributes:
       value: |
         Thanks for taking the time to fill out this bug report!
+  - type: checkboxes
+    attributes:
+      label: Is there an existing issue for this?
+      description: Please search to see if an issue already exists for the issue you encountered.
+      options:
+        - label: I have searched the existing issues
+          required: true
   - type: textarea
     id: bug-description
     attributes:
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -6,7 +6,7 @@ labels: ''
 assignees: ''
 
 ---
-- [] I have searched to see if a similar issue already exists.
+- [ ] I have searched to see if a similar issue already exists.
 
 
 **Is your feature request related to a problem? Please describe.**  
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -6,12 +6,14 @@ labels: ''
 assignees: ''
 
 ---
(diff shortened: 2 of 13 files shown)


abidlabs:
Nice, I prefer a non-file-based approach to store the queue. Let me know when you want a review!

cc @aliabid94

aliabid94:
Is the idea that the queue will now be stored in memory? I worry about OOM errors if working with large input data amd large queues. 
I did have some thoughts though on refactoring queueing, including:
- Refactoring the API so the user does not send input data until it is their "turn" in line. This way, we don't have to store data for possibly hundreds of inputs waiting in queue. 
- Currently, even if a user closes the browser, their queued data is run. Would be nice to have some heartbeat check to see if a user has closed the browser (could use the status check that is sent every second right now), and dequeue if a user has left.
- We could possibly use celery with sqlite backend if we want a dedicated query manager.

abidlabs:
> Is the idea that the queue will now be stored in memory? I worry about OOM errors if working with large input data amd large queues.

Ah this is a good point. We have gotten some really large queue sizes (>1000) and for image files or video files, this could go beyond RAM. 

But then wouldn't this fix that:

 > Refactoring the API so the user does not send input data until it is their "turn" in line. This way, we don't have to store data for possibly hundreds of inputs waiting in queue.

abidlabs:
> Currently, even if a user closes the browser, their queued data is run. Would be nice to have some heartbeat check to see if a user has closed the browser (could use the status check that is sent every second right now), and dequeue if a user has left.

+1

FarukOzderim:
> Is the idea that the queue will now be stored in memory? I worry about OOM errors if working with large input data amd large queues. I did have some thoughts though on refactoring queueing, including:
> 
>     * Refactoring the API so the user does not send input data until it is their "turn" in line. This way, we don't have to store data for possibly hundreds of inputs waiting in queue.
> 
>     * Currently, even if a user closes the browser, their queued data is run. Would be nice to have some heartbeat check to see if a user has closed the browser (could use the status check that is sent every second right now), and dequeue if a user has left.
> 
>     * We could possibly use celery with sqlite backend if we want a dedicated query manager.

Very good points!

> Refactoring the API so the user does not send input data until it is their "turn" in line. This way, we don't have to store data for possibly hundreds of inputs waiting in queue.

I think the point above solve RAM problems and need for a heartbeat check. When the users turn has come in the queue, if the user is not online(browser tab is open), we remove the user from the queue.

1. @aliabid94, would you like to integrate Frontend and Backend with this approach?
2. I think this refactoring is not very critical for 3.0 release, we could either have before or after the release, WDYT?

FarukOzderim:
Related #572

FarukOzderim:
Continued with #1489

PR #1137, opened 2022-04-29
Restore queue in Blocks

aliabid94:
Restore broken queue in blocks. Fixed by storing entire API call in queue database (including session_hash, fn_index), not just component data payload.

Fixes: #1120


osanseviero:
Thank you for this PR! :fire: :fire: 

It would be great if we could prioritize merging+releasing this PR as it blocks many cool use cases/demos

PR #1143, opened 2022-04-30
networking tweak

abidlabs:
Exposes a method to make it easier to set up a `FastAPI` server with a `Blocks` instance without necessarily launching the `FastAPI` server.


aliabid94:
LGTM

PR #1145, opened 2022-05-02
html tweaks

### LINKED ISSUE #1032: Set links in a Gradio app to open in a new tab by default
Based on feedback from @osanseviero @julien-c @radames, we should have links in Gradio apps (in the description, article, etc.) to open up in a new tab by default

A handy trick to make this easy (h/t @radames):

```
<head>
   <base target="_blank">
</head>
```

pngwn:
- all links now open in a new tab
- analytics is enabled in the svelte file, removes the annoying additional text during dev
- the config mustache templates are injected by a vite plugin to avoid syntax errors in the index file

Closes #1032

### DIFF
--- a/ui/packages/app/index.html
+++ b/ui/packages/app/index.html
@@ -11,7 +11,7 @@
 	"
 >
 	<head>
-		<base target="_blank">
+		<base target="_blank" />
 		<meta charset="utf-8" />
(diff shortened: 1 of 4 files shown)


abidlabs:
Good stuff

PR #1146, opened 2022-05-02
move styling for interface panels to frontend

### LINKED ISSUE #999: handle custom styling for interfaces in the frontend rather than in the python lib
At the minute we are injecting custom CSS to create the interface design. Ideally we would not need to do this and would use either some new layout components (`Panel`, etc.) or custom CSS properties to modify the components we are using.

The current approach makes the dev experience very tedious, when modifying the interface styling.

pngwn:
Adds a `variant` kwarg to `Column`. Closes #999.

@gary149 If you want to tweak the 'panel' styling for interfaces, feel feree to push to this branch.

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -166,7 +166,7 @@ class Column(BlockContext):
     ):
         """
         css: Css rules to apply to block.
-        variant: row type, 'default' or 'panel'
+        variant: column type, 'default' (no background) or 'panel' (gray background color and rounded corners)
         """
         self.variant = variant
         super().__init__(visible, css)
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -158,15 +158,23 @@ class Row(BlockContext):
 
 
 class Column(BlockContext):
-    def __init__(self, visible: bool = True, css: Optional[Dict[str, str]] = None):
+    def __init__(
+        self,
+        visible: bool = True,
+        css: Optional[Dict[str, str]] = None,
+        variant: str = "default",
+    ):
         """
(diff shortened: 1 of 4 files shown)


pngwn:
Don't actually know what I'm doing re: the python, so would appreciate a quick review @abidlabs @FarukOzderim @aliabid94 when you get time.

gradio/blocks.py
```diff
@@ -158,15 +158,23 @@ def get_template_context(self):
...
-    def __init__(self, visible: bool = True, css: Optional[Dict[str, str]] = None):
+    def __init__(
+        self,
+        visible: bool = True,
+        css: Optional[Dict[str, str]] = None,
+        variant: str = "default",
+    ):
         """
         css: Css rules to apply to block.
+        variant: row type, 'default' or 'panel'
```
abidlabs, line 169:
```suggestion
        variant: column type, 'default' or 'panel'
```

gradio/blocks.py
```diff
@@ -158,15 +158,23 @@ def get_template_context(self):
 
 
 class Column(BlockContext):
-    def __init__(self, visible: bool = True, css: Optional[Dict[str, str]] = None):
+    def __init__(
+        self,
+        visible: bool = True,
+        css: Optional[Dict[str, str]] = None,
+        variant: str = "default",
```
abidlabs, line 165: We've been using the parameter "type" for most of our components, so I suggest doing that here, at least in the user facing API

```suggestion
        variant: type = "default",
```

abidlabs:
Perfect, just updated the docstring!

Issue #1053, opened 2022-04-21
UI Improvements


pngwn:
This is a list of the remaining work on the UI and UX to keep track and aid comms.

## Loading
- [ ] add a loader for the first load of the entire Block @pngwn

## Layout
- [x] Layout: group form input #1054 
- [x] Mobile: Scroll to results #1077

### HighlightedText
- [x] Colors #1119
- [x] Higlight category on legend mouseover #1119
- [x] Show user-provided label (see `demo\diff_texts\run.py`) #1204

### Textbox
- [x] Show placeholder (see `demo\sentiment_analysis\run.py`) #1135
- [x] Textbox size does not adjust to fit the size of the output text (I know this has been brought up in other places), but here is an example (`demo\gpt_j\run.py`) #1153

### JSON
- [x] copy content #1205

### Loading state
- [x] Implement UI #1077
- [x] Loading animation:  https://play.tailwindcss.com/UDwhpFwrmQ?file=css  #1077

### Image
- [x] Sketch has some bugs #1215

### Button
- [x] CSS scoping Vs. component
- [x] Integrate secondary/primary on the Gradio side #1173

### Tabs
- [x] Responsivness

### Forms
- [x] Checkbox/radio focus with tab css

### Carrousel 
- [x] Center image
- [x] Label overflow text

### Table @pngwn
- [x] #962
- [x] #868
- [x] #631
- [x] #945
- [x] #745
- [x] #753
- [x] #1065

### Chart
- [x] Use tailwind colors #1141

## Examples
- [x] Update `DataSet` component #1121

## Typography
- [x] How to implement

## Fonts
- [x] Choose between system and source sans pro
- [x] Performance (loading, too many fonts)

created issue for post 3.0: #1224

## Theming
- [x] #741

abidlabs:
I've noticed some other UI issues as well @pngwn as I've been testing the various demos:

- [ ] `HighlightedText` does not show the `label` (see `demo\sentiment_analysis\run.py`)
- [ ] `Textbox` does not show the `placeholder` (see `demo\sentiment_analysis\run.py`)

pngwn:
I've added these issue to list. The `DataSet` issue will be handled when victor is finished looking at the examples.

I'm going to file a separate issue for the table and add it to the tracking issue for the `DataFrame`.

PR #1149, opened 2022-05-03
Timeseries improvements

### LINKED ISSUE #1088: Support custom colors for Timeseries
A user on discord posted:

> Is there a way to add more colors to legends in plots (in this case for a time series)? I have up to 8 lines to be plotted at once, but color scheme I am using  appears to support only 6 distinct colors: 
> 
https://huggingface.co/spaces/emilylearning/causing_gender_pronouns

I propose an additional argument should be passed to Timeseries, `colors`. Someththing like this:

```py
gr.Timeseries(colors=[x, y, z])
```

The list could either be named colors (`["red", "blue", "green"]`) for which we would choose a nice shade for you, hex codes (`#000000`), or rgb values (`rgb(0, 0, 0)` or `rgba(0, 0,0,0.5)`).

### LINKED ISSUE #1141: Time series with decimal numbers

It seems currently time series component only works with integer values.

Add support for decimal values.

**Additional context**

pngwn:
Made a few changes to the `Timeseries` component:

- Add a `colors` kwarg (`List[str]`) so that users can pass in custom colors. This supports certain  named colors (for which we will select a nice shade from our palette) as well arbitrary hex or rgb values.
- Use the color palette introduced for `HighlightedText` for automatically selecting colors
- Support decimal values as plot values
- Support decimal ranges (for example a chart that has x or y axes that only span 0-1)
- Ensured it works with negative values.
- Cleaned up the UI a little: Added some spacing, rounded some borders, fixed alignment of the legend.

Closes #1088. Closes #1141. Relates to #1053.

Some screenshots.

`0` to `<1` ranges with decimal values. Custom colors.

Screenshot 2022-05-03 at 13 14 49

'normal' range with decimal values. Custom colors.

Screenshot 2022-05-03 at 13 15 15

Negative values.

Screenshot 2022-05-03 at 13 15 34


abidlabs:
Sweeeeet!

abidlabs:
@pngwn I realized right after merging this that `Chatbot` and `HighlightedText` take a `color_map` parameter, whereas `Timeseries` takes in a `colors` parameter.

Let's rename this parameter to `color_map` as well for consistency?

pngwn:
Mainly because it isn't a map, it is a list. That is technically true for chatbot as well but in my head it is an implicit user/bot map. But I think you're right, will make a PR.

PR #1152, opened 2022-05-03
use color palette for chatbot

pngwn:
use color palette for chatbot colors.


dawoodkhan82:
Looks good. So we're only supporting a fixed number of colors?

pngwn:
It falls back to whatever was passed in we don't match the name, so it will still support hex + rgb colours, but we can document these 'nice shades' and allow people to use them if they want to.

abidlabs:
Very nice colors! 
image

PR #1153, opened 2022-05-03
max_lines + autoheight for textbox

### LINKED ISSUE #980: Textbox Autoheight Logic

pngwn:
- implement max_lines
- fix autoheight bug
- cleanup

Closes #980. Related #1053.

### DIFF
--- a/demo/gpt_j/run.py
+++ b/demo/gpt_j/run.py
@@ -10,7 +10,7 @@ examples = [
 
 demo = gr.Interface.load(
     "huggingface/EleutherAI/gpt-j-6B",
-    inputs=gr.Textbox(lines=5, label="Input Text"),
+    inputs=gr.Textbox(lines=5, max_lines=6, label="Input Text"),
     title=title,
     examples=examples,
 )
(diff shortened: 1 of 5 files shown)


abidlabs:
Looks good! One thing I noticed is that if a `Textbox` is used as an output (or generally has `interactive=False`), then there is no way to scroll the textbox if the text inside exceeds the `max_lines` amount

abidlabs (approved):

pngwn:
Ah good point. I'll have a think about that, we may be better off reverting to using a plain div rather than a disabled textbox for static output.

PR #1154, opened 2022-05-03
add select event for tabitems

### LINKED ISSUE #1123: implement `change` event for individual `TabItem`s
Implemented on the backend, not the front.

Note: maybe change the name to `show`, `load`, `select`.

@abidlabs any thoughts on the name?

pngwn:
@abidlabs I went for `select` can you check that my change to the python method is correct please. I have only half tested this, couldn't think of a good test case.

Closes #1123

### DIFF
--- a/demo/blocks_xray/run.py
+++ b/demo/blocks_xray/run.py
@@ -27,7 +27,7 @@ With this model you can lorem ipsum
     )
 
     with gr.Tabs():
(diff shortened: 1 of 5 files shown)


abidlabs:
Looks great and works great!

abidlabs (approved):

PR #1155, opened 2022-05-03
Allow enabling queue per event listener

aliabid94:
Needed for examples


abidlabs:
Haven’t tested it but code changes look good. If you’ve tested it, go ahead and merge

abidlabs (approved):

Issue #1029, opened 2022-04-19
Interfaces are no longer mobile-responsive & column widths are not equal


abidlabs:
### Describe the bug

Gradio Interfaces used to be mobile-responsive -- for narrow screen widths, the output would appear underneath the input, like this: https://huggingface.co/spaces/awacke1/HealthSummary

![image](https://user-images.githubusercontent.com/1778297/164065286-2c4fb7d8-7012-4f0a-87cb-e67881c41258.png)

However, this is not the case anymore, see:

![image](https://user-images.githubusercontent.com/1778297/164065419-97e94607-f8d0-4800-8f16-068ccb725d15.png)

### Reproduction

Run any demo, in this case I ran: `demo/outbreak_forecast/run.py`

### System Info

```shell
2.9.0b2
```

### Severity

annoyance

abidlabs:
Also, I've noticed that column widths for inputs and outputs are not always equal. If the output is a large image, for example, it will take up most of the width of the page, leaving the input column very narrow. Here's an example:

![image](https://user-images.githubusercontent.com/1778297/164070149-f064bb2c-f8bf-4b45-88b3-7df6648c2405.png)

Notebook to reproduce (make sure to install `gradio==2.9.0b2`: https://colab.research.google.com/drive/1MJ5Kz67ghYprPxJBjOal66VFU_CK2lqG#scrollTo=e200MmBU2aLT

abidlabs:
The unequal column widths are also an issue in the `blocks_xray` demo:

<img width="969" alt="image" src="https://user-images.githubusercontent.com/1778297/164504701-36aee051-70dd-4f1d-a9cd-487c1306e8c3.png">

abidlabs:
Closed via #1165 :)

PR #1165, opened 2022-05-05
enable flex grow for gr-box

radames:
# Description

For individual elements on a row, use `flex-1` so it grows and shrinks as needed

Before:
image

After:

image


radames:
thanks @pngwn I've added @gary149, but we came to that conclusion together in call this morning!

abidlabs:
Looks great to me too. I think this would close #1029 too.

pngwn:
I'll check #1029 and close that.

pngwn:
Thanks @radames !

PR #1168, opened 2022-05-06
Small fixes

abidlabs:
Small fixes that I noticed while writing the guides:

* better handling of null inputs or null outputs
* Added some alternative cases for component classes (e.g. `gr.DataFrame` is the same as `gr.Dataframe`)


abidlabs:
I messed up git, closing this branch and will open up a separate PR

Issue #1110, opened 2022-04-28
Rename (or accept both names) Dataframe to DataFrame


osanseviero:
**Is your feature request related to a problem? Please describe.**
https://gradio.app/docs/#o_dataframe uses `Dataframe` but in Pandas it's called `DataFrame`, which can lead to easily committing an error. 

**Describe the solution you'd like**
Allow `outputs.DataFrame` too.

PR #1171, opened 2022-05-06
A few small fixes

abidlabs:
Fixes a few small things I noticed as I was building demos / writing the `Blocks` guide

* allow markdown to be set dynamically (i.e. used as an output component)
* provide aliases for components that have a 2-word name in case people use the wrong case (i.e. `gr.DataFrame` is now equivalent to `gr.Dataframe`) -- Closes: #1110
* allows you to set `inputs=None` or `outputs=None` to an event trigger
* removed a flaky test


gradio/__init__.py
```diff
@@ -30,6 +34,7 @@
     Slider,
     StatusTracker,
     Textbox,
+    TimeSeries,
```
aliabid94, line 37: why twuce, typo?

gradio/interface.py
```diff
@@ -166,12 +166,12 @@ def __init__(
         )
 
         self.interface_type = self.InterfaceTypes.STANDARD
-        if inputs is None and outputs is None:
+        if (inputs is None or inputs == []) and (outputs is None or outputs == []):
```
aliabid94, line 169: since you call super.init() first, you could just check if self.inputs == [].

aliabid94 (approved):
LGTM, small comments

gradio/__init__.py
abidlabs: Different spelling (or casing actually)

abidlabs: TimeSeries vs Timeseries

abidlabs: That is what you're referring to, right?

gradio/interface.py
abidlabs: Where in `Blocks()` do I assign `self.inputs`?

PR #1185, opened 2022-05-07
1183 component height

### LINKED ISSUE #1183: Components getting cut off vertically

### Describe the bug

In some Blocks, the components don't show the full height, instead getting cut off vertically. E.g. look at the Textbox here:

image

@AK391 may be able to provide other examples

Usually just an annoyance, but sometimes this gets to the point where the component is unusable. 

### Reproduction

https://huggingface.co/spaces/abidlabs/cutoff

### Screenshot

_No response_

### Logs

_No response_

### System Info

```shell
2.9b24
```

### Severity

critical

### LINKED ISSUE #1184: form groups broken when in a row

### Describe the bug

Form groups are broken when in a row, rather than a column. This just hadn't been accounted for.

https://github.com/gradio-app/gradio/issues/690#issuecomment-1119351175

### Reproduction

-

### Screenshot

_No response_

### Logs

_No response_

### System Info

```shell
-
```

### Severity

critical

pngwn:
This was more involved that I thought it would be. Closes #1183. Closes #1184.

### Notes for #1183

In #1165 we changed some flex behaviour but this didn't work in all contexts, most notably it forced everything to grow regardless of its parent. This worked well for one case but broke some others (including the default interface layout). 

Most components are now aware of their parent and selectively apply `flex-1`.

I've also made some other tweaks as I found a few additional bugs when fixing this, all relating to layout.

**Interface**
<details>
 <summary>before</summary>
Screenshot 2022-05-07 at 10 16 38
</details>

<details>
 <summary>after</summary>
Screenshot 2022-05-07 at 13 45 39
</details>

**Custom Layout**

<details>
<summary>before</summary>
Screenshot 2022-05-07 at 12 26 04
</details>

<details>
<summary>after</summary>
Screenshot 2022-05-07 at 11 42 42
</details>

### Notes for #1184

We had some custom logic to detect when form components were adjacent to one another, removing spacing and top  + bottom border radiuses where necessary. However, this logic assumed that form elements always appeared in a column (one on top of the other). Side-by-side form elements _also_ applied the exact same logic leading to very broken layouts. With the above change making components aware of their parent, we can easily change the layout behaviour depending on whether the parent is a column or a row.

- If adjacent forms are in a column, they will be grouped 'stacked' (one on top of the other with top and bottom border radiuses modified).
- - if adjacent forms are in a row, they will grouped alongside one another with left + right border radiuses modified.

We might need to modify this if we change how layouts work.

<details>
<summary>before</summary>
Screenshot 2022-05-07 at 14 05 27

</details>

<details>
<summary>after</summary>
Screenshot 2022-05-07 at 14 06 39

</details>

### DIFF
--- a/demo/blocks_layout/run.py
+++ b/demo/blocks_layout/run.py
@@ -1,4 +1,3 @@
-from matplotlib import interactive
 import gradio as gr
 
 
--- a/demo/blocks_layout/run.py
+++ b/demo/blocks_layout/run.py
@@ -1,0 +1,23 @@
+from matplotlib import interactive
+import gradio as gr
+
+
+demo = gr.Blocks()
+
+with demo:
+    with gr.Row():
+
+        gr.Image(interactive=True)
+        gr.Image()
+    with gr.Row():
+        with gr.Row():
+            gr.Image(interactive=True)
+            gr.Image()
+            with gr.Column():
+                gr.Image(interactive=True)
+                gr.Image()
+    gr.Image()
+
+
+if __name__ == "__main__":
+    demo.launch()
--- a/demo/blocks_layout/run.py
+++ b/demo/blocks_layout/run.py
@@ -6,17 +6,26 @@ demo = gr.Blocks()
 
 with demo:
     with gr.Row():
-
         gr.Image(interactive=True)
         gr.Image()
     with gr.Row():
+        gr.Textbox(label="Text")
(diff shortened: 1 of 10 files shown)


pngwn:
Oh, @gary149, yet another use case for container queries. This was actually very annoying to implement.

gary149 (approved):

PR #1186, opened 2022-05-07
962 dataframe

### LINKED ISSUE #745: Dataframe Component examples are not working
### Describe the bug

Examples paramter for Dataframe component is not showing up in the app. Even for tax calculator app in the docs page is not showing dataframe examples.

### Reproduction

https://colab.research.google.com/drive/1DV1J45I-pkbwPEh0e6xVOkPgtLaXn3Zx

### Screenshot

image

### Logs

_No response_

### System Info

```shell
2.8.5
```

### Severity

annoyance

### LINKED ISSUE #753: Dataframe "New column" generating error when submitting
### Describe the bug

Hi,

I tried using the "New column" feature to populate custom data in the Dataframe component, but it seems to be giving me an error. When adding a new column and submitting, I get the following error : 

`ValueError: 3 columns passed, passed data had 4 columns`

I tested this feature in the Colab example in the documentation and Gradio also flags an error when submitting.

### Reproduction

```python
import gradio as gr


def filter_records(records, gender):
    return records[records["gender"] == gender]


iface = gr.Interface(
    filter_records,
    [
        gr.inputs.Dataframe(
            headers=["name", "age", "gender"],
            datatype=["str", "number", "str"],
            row_count=5,
        ),
        gr.inputs.Dropdown(["M", "F", "O"]),
    ],
    "dataframe",
    description="Enter gender as 'M', 'F', or 'O' for other.",
)

iface.test_launch()

iface.launch()

```

### Screenshot

image

### Logs

```shell
  File "/usr/local/lib/python3.7/dist-packages/gradio/routes.py", line 232, in predict
    app.interface.process, raw_input
  File "/usr/local/lib/python3.7/dist-packages/starlette/concurrency.py", line 39, in run_in_threadpool
    return await anyio.to_thread.run_sync(func, *args)
  File "/usr/local/lib/python3.7/dist-packages/anyio/to_thread.py", line 29, in run_sync
    limiter=limiter)
  File "/usr/local/lib/python3.7/dist-packages/anyio/_backends/_asyncio.py", line 818, in run_sync_in_worker_thread
    return await future
  File "/usr/local/lib/python3.7/dist-packages/anyio/_backends/_asyncio.py", line 754, in run
  File "/usr/local/lib/python3.7/dist-packages/gradio/interface.py", line 561, in process
    for i, input_component in enumerate(self.input_components)
  File "/usr/local/lib/python3.7/dist-packages/gradio/interface.py", line 561, in <listcomp>
    for i, input_component in enumerate(self.input_components)
  File "/usr/local/lib/python3.7/dist-packages/gradio/inputs.py", line 1470, in preprocess
    return pd.DataFrame(x, columns=self.headers)
  File "/usr/local/lib/python3.7/dist-packages/pandas/core/frame.py", line 700, in __init__
    dtype,
  File "/usr/local/lib/python3.7/dist-packages/pandas/core/internals/construction.py", line 483, in nested_data_to_arrays
    arrays, columns = to_arrays(data, columns, dtype=dtype)
  File "/usr/local/lib/python3.7/dist-packages/pandas/core/internals/construction.py", line 807, in to_arrays
    content, columns = _finalize_columns_and_data(arr, columns, dtype)
  File "/usr/local/lib/python3.7/dist-packages/pandas/core/internals/construction.py", line 909, in _finalize_columns_and_data
    raise ValueError(err) from err
ValueError: 3 columns passed, passed data had 4 columns
```

### System Info

```shell
Gradio version : 2.8.1
OS : macOS Monterey 12.2.1
Browser : Google Chrome Version 98.0.4758.109
```

### Severity

blocker

### LINKED ISSUE #868: Improve usability of dataframe input
See https://huggingface.co/spaces/osanseviero/accuracy_metric

Modifying the input of a cell requires double-clicking on it. I would love to be able to just click and add my input. There were also a couple of dev experience improvements I would love to see

Three different things that would be great:
* Single click in dataframe
* Allow making the number of columns or rows fixed, since for some use cases you don't want users creating new rows.
* With using `datatype="number"`, I would love if users can only write numbers and not strings. I know this modifies what is passed to the interface, but modifying the user input type as well would be great

### LINKED ISSUE #945: DataFrame support from Files
Gradio's file input takes file and doesn't do any preview of the file, and [DataFrame](https://gradio.app/docs/#i_dataframe) input doesn't really support any dataframe creation from files like a regular data science workflow with `pandas_dataframe.head()` would have in a colab. It would be better UX for data scientists to input a CSV/excel and have a head or tail of a dataset with as much columns/rows that they choose like you do with `df.head(5)` or `df.tail(5)` in pandas. A workaround we did with Keras working group is below and it's too much hassle. 
Ekran Resmi 2022-04-07 16 40 27
If you upload a file directly it looks like this which applies for CSV files as well, nothing informative about what model does, the features etc. 
unknown

Pinging @aliabid94 and @abidlabs here 🙌🏻

### LINKED ISSUE #962: Dataframe Improvements
There are a number of issues with the Dataframe component as it exist today and we need to do some work to fix the outstanding but also improve the usability for humans. 

We can use this issue to keep track of the issues that have been reported and come up with a design that addresses the usability issues. I'll start with a simple proposal and we can discuss from there.

## python API Changes

Today the dataframe API looks like this:

```py
Dataframe(
  headers=None, 
  row_count=3, 
  col_count=3, 
  default_value=None, 
  datatype="str", 
  label=None, 
  col_width=None, 
  type="pandas", 
  optional=False
)
```

### modifying column width

Proposal: _`col_width` should be removed._

I think allowing users to set the column width is not a good idea. The whole purpose of gradio is to generate high quality web apps to share and showcase models. This should work across device sizes and screen widths. Tables automatically adapt column widths to accomodate their content, providing an API that will almost definitely break the UI is not ['pit of success'](https://blog.codinghorror.com/falling-into-the-pit-of-success/) stuff. 

### fixed column and row count

Proposal: _`col_count` and `col_width` should take either a `number` or a tuple of `(number, "fixed"|"dynamic")._

in #868 @osanseviero wrote:
> Allow making the number of columns or rows fixed, since for some use cases you don't want users creating new rows.

We do not currently have a mechanism to prevent end-users from creating new columns and rows. I think we have two options here:

- **create `col_fixed` and `row_fixed` boolean kwargs.**
  This adds additional options to dataframe when it already has a lot of kwargs, it could start to get overwhelming if we keep adding to the API, but it is simple and would work.
- **allow `row_count` and `col_count` to take _either_ a number (e.g. `3`) or a tuple of `(number, "fixed"|"dynamic")`**
  This does expand the API for `*_count` but I quite like it as it binds two highly related options together. This would look like `col_count=(3, "fixed")`. `col_count=3` would essentially be shorthand for `col_count=(3, "dynamic")`. 
  
_Note: We could rename these kwargs to `col`, and `row`_
  
I propose the second option (tuple) but I do not feel strongly about it.

## conflicts and confusements with kwargs relating to col and row quantity

Proposal: _`headers`, `col_count`, `row_count`, `default_value` should be validated to ensure there are no conflicts_.

More specifically: _Any combination of kwargs that can set the column count must always equal the same number of columns. Any kwargs that can set the row count must not result in provided data being hidden._

`headers` and `col_count` can conflict; `default_value` and `headers` can conflict (kinda); `default_value`, `col_count`, `row_count` and `default_value` can conflict.

This is easiest to explain with examples. 

This is confusing but not _necessarily_ an issue:

```py
Dataframe(
  col_count=3,
  headers=["One", "Two"]
)
```

However this is just wrong and will lead to unexpected behaviour:

```py
Dataframe(
  col_count=2,
  headers=["One", "Two", "Three"]
)
```

What should happen here:

```py
Dataframe(
  headers=["One", "Two", "Three"]
  default_value=[[1, 2], [3, 4]]
)
```

And here:

```py
Dataframe(
  default_value=[[1, 2, 3, 4], [5, 6, 7, 8]],
  col_count=2
)
```

We need to figure out simple rules to validate datafram inputs that affect the ciolumns + widths, or decide how to normalise. 

Some possible rules aimed at removing ambiguity:

- if `headers` and `col_count` are provided, the length of `headers` must be equal to `col_count`.
- if `headers` and `default_value` are provided, the length of each piece of column data in `default_value` should match the length of `headers`.
- if `default_value` and `col_count` are provided, the length of each piece of column data in `default_value` should be equal to col_count.
- if `default_value` and `row_count` are provided, the length of the row data in `default_value` should be equal to or less than the `row_count`. (This isn't essential but would lead to weird behaviour).

The obvious counter to this is that we could add additional values to `default_value` or `headers` to 'fill in the gaps' but I think the API will be far easier to reason about for users if we have clear rules. It will allow us to easily detect errors and provide helpful messages to users. Trying to guess what users want without being explicit is how perl happened.

This validation would happen at python time, and we could provide error messages like: 

```
`col_count` is 3 but you passed 2 headers. Set `col_count` to 2 or add 1 header, even if it is an empty string.
```

```
`col_count` is 2 but you passed 4 headers. Set `col_count` to 4 or remove 2 headers.
```

```
default_value contains data for 4 columns but col_count is set to 2. Set col_count to 4 or remove some data from `default_value`
```

### proposed python API

```py
Dataframe(
  headers=None,             # validated
  row_count=(3, "dynamic")  # validated
  col_count=(3, "dynamic")  # validated
  default_value=None,       # validated
  datatype="str", 
  label=None, 
  type="pandas", 
  optional=False
)
```

### make cells easier to interact with

in #868 @osanseviero said:
> Modifying the input of a cell requires double-clicking on it. I would love to be able to just click and add my input. There were also a couple of dev experience improvements I would love to see

I'm not certain about this. 

The current behaviour mimics how most spreadsheets work but users of spreadhseets freequently move around the spreadsheet befopre editing. Our dataframe is not a powertool but a quick user entry tool, so perhaps ease of data entry is more important than ease of cell navigation. 

If we change `click` behavioour, we also need to change keyboard behaviour for parity of usability. Essentially this feature request is to remove the different 'states' from the dataframe, so that it is essentially 'edit only', rather than having view/ edit modes as without click triggering that state it would be impossible to get to. Static or output dataframes would still have this behaviour.

@omarespejel Could you add some mroe details about how you would like to interact with the dataframe. Not just click but how would you like to change to a different cell, how would that work for keyboard users who do not or cannot use a mouse?

### better inputs when the datatype is given

in #868 @osanseviero said:
> With using datatype="number", I would love if users can only write numbers and not strings. I know this modifies what is passed to the interface, but modifying the user input type as well would be great.

Currently everything is treated as a string by the frontend, even when we know the datatype. I think we can improve this significantly.

- `number` fields could use a number input which will only allow number entry
- `boolean` fields could use a toggle or checkbox 
-  `date` fields could use a date input
- `string` fields will use the current textbox functionality

But I think we can go further if we expanded the datatype kwarg. 

- enums/ unions would render a dropdown or autocompleting dropdown thing.
- we could support prefix + suffixes for currencies + measurements. `datatype=[()]
- it might be possible to support custom validators in the future as long as they are regex based.

Be good to get your thoughts @gary149

### populate dataframe fom file

@merveenoyan create issued #945 discussing uploading csv/ tsv files into the dataframe.

I think this makes perfect sense. We are alrady doing this for the timeseries, adapting it for the dataframe should be straightforward.

@merveenoyan could you clarify the first part of that issue. Are you saying it would be good for the dataframe to accept different values in the python library (i.e. the `default_value` kwarg)? or that It would be good to be able to modify what is displayed after a user uploads the file via the UI (i.e. only showing the first/ last 5 rows, etc.)?

### row + col creation and deletion

#631 

Row and column creation and deletion needs some work. Deletion isn't currently possible.

Would love to get people's thoughs on what good creation and deletion might look like, are there other datatables you have seen in the wild that do this well while remaining very compact?

Another for @gary149

### Redesign

Just putting this here for posterity. Things are being redesigned.

## bugs

We have bugs:
 - [x] #921
 - [x] #890
 - [ ] #753 
 - [ ] #745
 - [ ] #1065

Issues relating to features for tracking purposes:
- [ ] #868
- [ ] #945
- [ ] #631 

---

Let me know if I have missed anything and would be good to get people's thoughts on this.

cc @abidlabs @aliabid94 @dawoodkhan82 @aliabd @FarukOzderim

### LINKED ISSUE #1065: Dataframe unfocuses when typing
### Describe the bug

Dataframe component "unfocuses" the cell as soon as you type a single character. Try using tax_calculator demo for example

### Reproduction

-

### Screenshot

-

### Logs

```shell
-
```

### System Info

```shell
-
```

### Severity

critical

pngwn:
This is the refreshed dataframe and captures most the changes that have been discussed.

Short version: i changed the dataframe. Try it by running `demo/blocks_outputs`, I added a few dataframes there.

**API**

- `col_width` has been removed (#962)
- `row_count` can now take a `int` or a tuple of (int, 'fixed' | 'dynamic') (#962)
- `col_count` can now take a `int` or a tuple of (int, 'fixed' | 'dynamic') (#962)
- A small amount of validation is done to check `col_count` and the length `headers` are the same. We could probably do more validation but I think this is enough to start with.

**UX**

- You can drop a valid TSV or CSV onto the dataframe to populate it. Valid basically means not malformed + it needes headers. There isn't any error handling right now.
  - When you drag the table fades slightly and the border goes green.
  - If you have set `col_count=(n, 'fixed')` then that is how many columns will show up. Data is trimmed or blanks are filled in as needed.
  - If you set `col_count=(n, 'fixed')` then that is how many rows will show up. Data is trimmed or blanks are filled in as needed.
- Clicking or Tabbing a cell 'selects' it, from here you can move around the cell with arrow keys or tab. Shift + tab goes backwards. Tabbing out of the table carries on the normal tab flow of the page, ditto for shift tab.
- When selecting a cell you can type any printable key to erase current content and start editing. 
- When selecting a cell you can press return to _continue_ editing a cell
- When selecting a cell you can press Shift + Return to create a new row at your current position. This allows you create Rows in between rows.
- When editing a cell, clicking outside of the cell, pressing escape, return or tab will cancel editing of that cell.
- Single clicking a cell selects it, double clicking a cell edits it (without clearing it).
- If you set If you set `col_count=(n, 'fixed')` the "New Column" button is hidden and the action prevented.
- If you set `row_count=(n, 'fixed')` the "New Row" button is hidden and the action is prevents.

**Bugs**

- many bugs have been fixed
- many new bugs were created
- they were then fixed
- there are probably more bugs

The only thing that hasn't been done is removing rows + cols. There is an open issue for that which we will get to later, it will need some UX work but I don't think it needs to block 3.0

Closes #753. Closes #868. Closes #1065. Closes #945. Closes #962. Closes #745.

(need to double check #745 is fixed by this)


abidlabs:
Wow took me a while to review all of the fixes and new features. Great work, especially love the drag-and-drop csv!

There are a couple of minor issues that I noticed -- for example, dataframe examples don't adjust the number of rows in the dataframe UI when clicked, which can cause problems if the number of rows in the example < the number of rows in the dataframe UI. But these are pretty minor things we can address post 3.0 launch, overall LGTM.

Pushed a few tweaks to make everything work, and will go ahead and merge so that we can release a new version.

abidlabs (approved):

PR #1187, opened 2022-05-08
WIP: Add dark mode support

### LINKED ISSUE #741: [Blocks] Support dark mode in the browser
Blocks currently doesn't support dark mode, as can be seen by Space: https://huggingface.co/spaces/chinhon/multilingual_headline_writer

(screenshot below)

image

gary149:
This PR adds support for dark mode:

image

**Please note:**

- It uses the class strategy [to manually toggle dark mode](https://tailwindcss.com/docs/dark-mode#toggling-dark-mode-manually). It doesn't mean the default can't be prefers-color-scheme but this also lets the user toggle dark/default mode by himself (we did the same on the HF website).
- There is an issue with icons, we want to use `currentColor` but it might require inline svgs.
- We want different colors for dark mode in `ui/packages/theme/src/colors.ts`. It's not done by now.

Closes #741


cbensimon:
> let's also use this as an opportunity to improve/speed up the way we pass the `theme` from hf.co, ok? cc @cbensimon

Yes, totally agree @julien-c

cbensimon:
Context (for Spaces) :

We [currently](https://github.com/gradio-app/gradio/blob/9c1cef57f67694de29135fc6ea6f581572e7a324/ui/packages/app/src/main.ts#L89-L101) check if the `__dark-theme` URL query parameter is `true` to enable the darktheme for Gradio on Spaces.
We still need to support this (backward compatibility).

As @julien-c mentioned, in order to speed-up Space page loads,
we need to support reading a new `__theme` query parameter that can take three values : `light`, `dark` and `system`
If the value is `system`, we want to set the theme according the the system theme (`prefers-color-scheme`)

@pngwn what do you think of this ?

pngwn:
@cbensimon Are those query params set depending on what is set in the hub? Would light/ dark ever be set if the hub doesn't have a toggle?

I think those three values would give us what we need on the gradio side to keep things in sync with the surrouding context.

pngwn:
I can add this logic to the PR here, falling back to system for non-space gradio apps. There will also be some work required on the hub, could that be ready for Gradio 3.0 release on wednesday?

julien-c:
on the hub side i think it's just going to be a super super small PR. @cbensimon will link it to here when it's ready

cbensimon:
> @cbensimon Are those query params set depending on what is set in the hub? Would light/ dark ever be set if the hub doesn't have a toggle?

The `__theme` query parameter value will be the theme value from the hub (user theme setting or `light` by default if not logged-in).

The old (current) `__dark-theme` value is the theme "at runtime" (after DOM is loaded)

pngwn:
- resolve conflicts
- converted icons to components, used currentColor for strokes/fills, gave icons their own packages
- implemented dark mode switcher based on query params/ system settings (`__theme` query param -> `__dark-mode` query param => system)
- various tweaks.

abidlabs:
Tested with a bunch of blocks and interface demos. Really good stuff guys!

abidlabs (approved):

Issue #727, opened 2022-02-23
Gradio Async Support


FarukOzderim:
Blocks and interface should support async functions.

abidlabs:
Moving to the 3.x milestone since we are aiming to release this soon @FarukOzderim

wbqtac:
hi there, wish to check, does the gradio support async call now?

freddyaboulton:
Yes!

PR #1190, opened 2022-05-09
async-function-support

FarukOzderim:
- support async functions with Blocks
- refactor app a bit

Related: #727 

This took some time to correctly work on Interface, it seems lambda functions are always wicked :D


FarukOzderim:
You can try these:


```
import asyncio
import gradio as gr


async def async_func(*args):
    await asyncio.sleep(5)
    return "Hello after 5 second"


gr.Interface(async_func, "text", "text").launch()
```


```
import asyncio
import gradio as gr
with gr.Blocks() as demo:
    text = gr.components.Textbox()
    button = gr.components.Button()
    button.click(async_func, text, text)
    demo.launch()

```

gradio/routes.py
```diff
@@ -78,199 +78,228 @@ class PredictBody(BaseModel):
...
-
-    @app.get("/login_check")
-    @app.get("/login_check/")
-    def login_check(user: str = Depends(get_current_user)):
-        if app.auth is None or not (user is None):
-            return
-        raise HTTPException(
-            status_code=status.HTTP_401_UNAUTHORIZED, detail="Not authenticated"
-        )
+class App(FastAPI):
```
FarukOzderim, line 81: No functional change here, just tidied things a little bit into OOP.

FarukOzderim:
There are some problems, not completed yet.

gradio/routes.py
abidlabs: Nice, thanks!

abidlabs (approved):

FarukOzderim:
> Looking great @FarukOzderim!
> 
> I would only that we add a test in `test_blocks` with an async function to ensure everything works as expected.

Great idea!

Issue #785, opened 2022-03-07
Potentially add more configurations to the footer / make it expand into a menu?


abidlabs:
We need to add a footer to Blocks or Interface made from Blocks. We could potentially use this opportunity to build out the Gradio footer a little more. It could provide users with:

* ability to toggle between supported languages 
* ability to switch between dark/light mode
* ability to generate an embed link (maybe?)
* ability to record a GIF / screencaste (maybe?)
* ability to view the API docs (already implemented)

To support all this without taking up too much real estate, we should consider redesigning the footer so that these options appear in a menu when you click on the footer.

abidlabs:
@dawoodkhan82 based on our discussion today, I would just add a simple "built with Gradio" type footer for now. 

We can add other menu options like the view API (which needs to be rewritten anyways with the new endpoints) and i18n toggle later after the more detailed redesign.

abidlabs:
Updating the issue since it is no longer a blocker for merging `blocks-dev` into `main`, but it still has useful suggestions for the future

freddyaboulton:
I think we should close. This is over two years old without activity and no one has asked for this in that time hehe.

abidlabs:
Reopening in light of https://github.com/gradio-app/gradio/pull/10254

PR #1199, opened 2022-05-09
Restored /api/predict/ endpoint for Interfaces

abidlabs:
As discussed, this brings back the ability to send requests to `/api/predict` to any Interface so that people can continue using Gradio Spaces as api endpoints. 

This does not add the "view API" link in the footer -- we'll leave that for after 3.0 as part of #785


gradio/interface.py
```diff
@@ -587,6 +574,26 @@ def clean_html(raw_html):
...
+            if self.allow_flagging == "manual":
+                if self.interface_type in [
+                    self.InterfaceTypes.STANDARD,
+                    self.InterfaceTypes.OUTPUT_ONLY,
+                ]:
+                    flag_btn.click(
+                        lambda *flag_data: self.flagging_callback.flag(flag_data),
+                        inputs=self.input_components + self.output_components,
+                        outputs=[],
+                        _preprocess=False,
```
aliabid94, line 586: queue=False skipped here

abidlabs: Good catch

gradio/interface.py
```diff
@@ -450,9 +450,14 @@ def clean_html(raw_html):
             cache_interface_examples(self)
 
         if self.allow_flagging != "never":
```
aliabid94, line 452: you could make a variable called `all_components = self.input_components + (self.output_components if self.interface_type =...)`
and then use `all_components` everywhere instead of repeating so much flagging logic

aliabid94 (approved):

gradio/interface.py
abidlabs: Yeah ok

PR #1202, opened 2022-05-10
move init script into index.html

### LINKED ISSUE #1028: Footer appears twice (once above the interface) when `share=True`
### Describe the bug

When `share=True` in `launch()`, the footer appears twice:

image

### Reproduction

import gradio as gr

gr.Interface(lambda x:x, "textbox", "textbox").launch(share=True)

### Screenshot

_No response_

### Logs

_No response_

### System Info

```shell
2.9.0b2 (the latest on main)
```

### Severity

annoyance

pngwn:
We currently have an issue where the scripts are evaluated even when they are redirected. I haven't had the time to debug this properly (and we do need to stop it from happening) but this is a quick fix that should prevent the app from initialising twice (causing duplicated content) as we do not redirect the HTML file as far as I am aware.

Will create a follow up to look more closely at the redirect issue post 3.0.

This will need to be published before it can really be tested.

Closes #1028.

### DIFF
--- a/ui/packages/app/index.html
+++ b/ui/packages/app/index.html
@@ -74,5 +74,26 @@
 			id="root"
 			style="display: flex; flex-direction: column; flex-grow: 1"
 		></div>
+
+		<script type="module">
+			if (window.gradio_mode == "app") {
+				get_config().then((config) => {
+					window.launchGradio(config, "#root");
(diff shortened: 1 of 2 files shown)


abidlabs:
Let me test!

PR #1204, opened 2022-05-10
hltext-label

pngwn:
- show HighlightText label
- impropve colors
- tweaks


pngwn:
Colors aren't 100% there but I'll leave that to victor.

PR #1208, opened 2022-05-10
Sharing fix for assets

aliabid94:
When uploading to pypi, will create a second template html that links to the CDN for assets directly called share.html. When `share=True`, this file is used as the template.

Closes: #1028


abidlabs:
Have you tested this with a pypi release or would you like me to?

abidlabs:
It works!! https://colab.research.google.com/drive/1pUTiKv4fO04-qdy-dwJAPExIb6ibl5aE

image

abidlabs (approved):

abidlabs:
I don't know why the tests are timing out on CircleCI, they work fine locally. Any ideas @aliabid94?

abidlabs:
Works! To test, try `gradio==2.9b48`

PR #1215, opened 2022-05-11
fix placeholder prompt

pngwn:
Ensures that the Sketch placeholder text only appears when there is nothing on the canvas.


abidlabs:
Looks good!

Issue #820, opened 2022-03-15
Video Example Guide Not Working


FarukOzderim:
### Describe the bug

https://colab.research.google.com/drive/1hWvmAfh_2TVxK5D0_EsJuGYTDqMQTjrx#scrollTo=e21CGmzAaOq8

This example does not work in colab or gradio public link.

### Reproduction

https://colab.research.google.com/drive/1hWvmAfh_2TVxK5D0_EsJuGYTDqMQTjrx#scrollTo=e21CGmzAaOq8

### Screenshot

![image](https://user-images.githubusercontent.com/44067394/158328802-b2042c62-573a-4464-afe2-0518d4b02c5c.png)

### System Info

```shell
Google colab
Gradio Version 2.8.10
Firefox
```

### Severity

critical

aliabid94:
Webcam and Microphone never worked on colab because the iframe doesn't allow media access, but we used to print a warning for that when you ran in colab. Will take a look what happened to it.

FarukOzderim:
I see, try it gradio link as well please, because the video did not work gradio public link.

abidlabs:
Not sure what the original issue was exactly, but when I'm testing it now with `gradio==2.9.0b2`, the input video records fine from the webcam, but no output video shows up. There is no Python error, so I'm assuming the issue is with the Video output component. 

I tried changing `playable_video` to `video`, but it did not make any difference.

yvrjsharma:
Hi, the issues that I have identified might be a bit related to this hence mentioning it here.
1. In the Docs both the Input and Output Video demo/examples are same or they only talk about inputting a video file but not about gr.outputs.Video() function. You might want to look into it.
2. Using the Public URL The input video is recording for me as well , however the sound doesn't get captured. I have tried couple time with inbuilt mic in my laptop and and with a mic enabled headset too.
Apologies if I am not suppose to raise these issues in this thread. Please feel free to point me in the right direction.

FarukOzderim:
@yvrjsharma thanks for reporting these. 
Yeah I think these are related to this issue. Btw we are redesigning our docs with the planned 3.0 release.

PR #1218, opened 2022-05-11
ensure webcam video value is propogated correctly

pngwn:
This fixes the issue with the webcam videos not being capture as data correctly.

I've also removed all flips from webcam related inputs. So they aren't mirrored anymore. This removes any inconsistency in how they displayed. We _actually_ want the opposite, mirror all webcam inputs but it is far more complex than just flipping an element, so I've put that off for now. We might be able to get it done very soon but it isn't a priority over other things.

Closes: #820


abidlabs:
LGTM!

abidlabs (approved):

PR #1234, opened 2022-05-12
update-shortcut-syntax

FarukOzderim:
- fix&update gr.component
- create a demo introducing shortcuts within Blocks

gr.component()'s functionality was changed and lost its main purpose, it was returning classes instead of objects, return that functionality back.


radames (commented):
It looks cool to me! thanks for the new demo, I guess it now makes more sense.

demo/blocks_component_shortcut/run.py
```diff
@@ -0,0 +1,25 @@
...
+    2. gr.templates.Template() or gr.Template()
+    All the templates are listed in gradio/templates.py
+    """
+    with gr.Row():
+        with gr.Column():
+            text1 = gr.component("textarea")
+        with gr.Column():
+            text2 = gr.TextArea()
+        with gr.Column():
+            text3 = gr.templates.TextArea()
```
pngwn, line 21: Would just having these components at the top level achieve the same layout?

PR #1243, opened 2022-05-13
Button default name

FarukOzderim:
# Description

Add a default button name since it was not looking good in default.

Before:
image

Now:
image


abidlabs:
Looks good, I think this is a good change. The other alternative would be just to make a Button's value a required parameter, but this works as well. Let's get feedback from @gary149 or @pngwn as well.

pngwn:
I think "Run" is a reasonable default, although the sole puprose of this button is to render some text, I don't think asking users to pass in a value is unreasonable. I'd be happy with either the default or making it required.

gary149:
> Looks good, I think this is a good change. The other alternative would be just to make a Button's value a required parameter, but this works as well. Let's get feedback from @gary149 or @pngwn as well.

I think it makes more sense as a required param (so buttons always get the right wording), but no super strong opinion on it.

pngwn:
Lets make it required then, seems reasonable.

FarukOzderim:
If there is no strong opinion on this I would suggest we continue with default values.
Since we are supporting component shortcuts in the Blocks as well, and component shortcuts are initializing components with default values.

Check [demo](https://github.com/gradio-app/gradio/blob/main/demo/blocks_component_shortcut/run.py)

pngwn:
The string shortcut is a good point. Lets merge this.

pngwn (approved):

Issue #1241, opened 2022-05-13
[fake_gan_complex] Markdown with Interface input (no Blocks) bug


radames:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

`gr.Markdown` component raises  `'requires_permissions'` when used inside `Interface`  inputs without  `Blocks`

https://colab.research.google.com/drive/1sU2aU61roXO0mxTyO2Dk2ObaRHtMuRId?usp=sharing

`AttributeError: 'Markdown' object has no attribute 'requires_permissions'` 

### Reproduction

https://colab.research.google.com/drive/1sU2aU61roXO0mxTyO2Dk2ObaRHtMuRId?usp=sharing

https://github.com/gradio-app/gradio/blob/main/demo/fake_gan_complex/run.py

```python


demo = gr.Interface(
    fn=fake_gan,
    inputs=[
        gr.Image(label="Initial Image (optional)"),
        gr.Markdown("**Parameters**"),
        gr.Slider(25, minimum=0, maximum=50, label="TV_scale (for smoothness)"),
        gr.Slider(25, minimum=0, maximum=50, label="Range_Scale (out of range RBG)"),
        gr.Number(label="Respacing"),
        gr.Markdown("**Parameters Two**"),
        gr.Slider(25, minimum=0, maximum=50, label="Range_Scale (out of range RBG)"),
        gr.Number(label="Respacing"),
        gr.Markdown("**Parameters Three**"),
        gr.Textbox(label="Respacing"),
    ],
    outputs=gr.Image(label="Generated Image"),
    title="FD-GAN",
    description="This is a fake demo of a GAN. In reality, the images are randomly chosen from Unsplash.",
)
```

vs no Markdown

```python

demo = gr.Interface(
    fn=fake_gan,
    inputs=[
        gr.Image(label="Initial Image (optional)"),
        gr.Slider(25, minimum=0, maximum=50, label="TV_scale (for smoothness)"),
        gr.Slider(25, minimum=0, maximum=50, label="Range_Scale (out of range RBG)"),
        gr.Number(label="Respacing"),
        gr.Slider(25, minimum=0, maximum=50, label="Range_Scale (out of range RBG)"),
        gr.Number(label="Respacing"),
        gr.Textbox(label="Respacing"),
    ],
    outputs=gr.Image(label="Generated Image"),
    title="FD-GAN",
    description="This is a fake demo of a GAN. In reality, the images are randomly chosen from Unsplash.",
[...]

Issue #1248, opened 2022-05-13
[3.0b1 issue] gradio.Interface.load() doesn’t load Spaces that use the gr.Interface but are on 3.0


apolinario:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

If you try to use `gradio.Interface.load()` on a Spaces that uses the `gr.Interface` but has already updated to 3.0b1 (or previous pre-releases), the interface does not load properly throwing an error

### Reproduction

I downgraded the Spaces that were on 3.0 back to 2.9 since I got this errors and they were already deployed in production, but any Spaces I own that I upgraded to 3.0 when I tried `gradio.Interface.load()` it broke. If you can't replicate I can upgrade again to help

### System Info

```shell
Gradio==3.0b1
```

### Severity

blocker

PR #1260, opened 2022-05-14
Fixes to components

abidlabs:
* added style components for dall-e mini demo
* added support for Markdown as an IO component (closes: #1241)
* brings back basic implementation of carousel for backwards-compatibility with gradio 2.x for `gr.Interface.load()`
* small cleanups

Started working on this, but will move to a dedicated PR as it is quite complex:
* fixes gr.Interface.load for 3.0 demos (#1248)

### DIFF
--- a/gradio/interface.py
+++ b/gradio/interface.py
@@ -206,10 +206,12 @@ class Interface(Blocks):
 
         self.input_components = [get_component_instance(i).unrender() for i in inputs]
         self.output_components = [get_component_instance(o).unrender() for o in outputs]
-        
+
         for component in self.input_components + self.output_components:
-            if not(isinstance(component, IOComponent)):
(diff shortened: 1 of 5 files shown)


gradio/interface.py
```diff
@@ -206,6 +207,12 @@ def __init__(
         self.input_components = [get_component_instance(i).unrender() for i in inputs]
         self.output_components = [get_component_instance(o).unrender() for o in outputs]
 
+        for component in self.input_components + self.output_components:
```
aliabid94, line 210: or gr.Variable

aliabid94 (approved):
LGTM

gradio/interface.py
abidlabs: Good catch thanks

PR #1261, opened 2022-05-15
maybe fix auth in iframes

### LINKED ISSUE #1259: Auth is broken on Gradio 3.0

### Describe the bug

`auth` doesn't seem to be working in 3.0

### Reproduction

Check `demo\hello_login`

### Screenshot

_No response_

### Logs

_No response_

### System Info

```shell
3.0b5
```

### Severity

blocker

pngwn:
- close the loader when the login form loads
- use a different mechanism to open link in a new tab that _only_ affects links (a tags) rather than form actions as well.


pngwn:
Difficult to test since the real issue is only present in spaces/ iframes.

abidlabs:
Actually @aliabid94 would you be able to review this?

aliabid94:
ok will test

gradio/templates/frontend/index.html
```diff
@@ -11,7 +11,7 @@
 	"
 >
 	<head>
-		<base target="_blank" />
+		<!-- <base target="_blank" /> -->
```
aliabid94, line 14: Remove?

ui/packages/app/src/Blocks.svelte
```diff
@@ -149,6 +149,14 @@
 
 	async function handle_mount() {
 		await tick();
+
```
aliabid94, line 152: why?

aliabid94:
I didn't quite understand how this fixed the problem, but auth works locally when I tested it at least. LGTM otherwise

ui/packages/app/src/Blocks.svelte
pngwn: Because we want to open them all in a new tab (including any links that are generated via markdown, which is why we need to do it this way), this was previously being done with the `base` element but that _also_ affects form actions which we don't want because of the login form in iframes.

pngwn:
Yeah sorry, I didn't elaborate.

A while ago we added the base element with a `target="_blank"` to open all links in a new tab. Unfortiunately this also affects form `action`s. I think that is what is break auth in iframes because, opening the form action link in a new tab causes it to hang and fail. We want the form action to open the url in the same window (the iframe). I don't know for certain that it will fix it but I think it will

gradio/templates/frontend/index.html
pngwn, line 14:
```suggestion
```

abidlabs (dismissed):
Unfortunately, does not seem to work on Hugging Face Spaces. 

I released `gradio==3.0b8` with the changes in this branch, and it works locally and with share=True. However, it doesn't seem to work on Spaces. Here's a sample Space I bult to test: https://huggingface.co/spaces/abidlabs/hello-login

abidlabs (approved):

PR #1262, opened 2022-05-15
Remove disabled checkboxes

aliabid94:


abidlabs:
@aliabid94 do you mean to remove the disabled versions completely? I don't think we should do that because there might be certain situations where users want that. 

I think we should just make the defaults (for the form components) be the interactive versions.

pngwn:
Proposed alternative here: https://github.com/gradio-app/gradio/pull/1264

Apologies for alternative PR without discussion but worried about time.

PR #1264, opened 2022-05-15
Static forms

pngwn:
Alternative to #1262.

- If a component has is not an input or ouput it will be dynamic, for a better dev workflow. 
- If a component is used as an output but never an input it will be static (I know that non-intercative form inputs can be a little confusing but being able to interact and nothing ever happen is worse imo, we may be able to improve the UX in the future)

So broadly:

- Nothing - interactive
- Input - interactive
- Input + output - interactive
- output - static
- explicit`interactive=bool` are always respected


aliabid94:
LGTM!

aliabid94 (approved):

PR #1265, opened 2022-05-15
Cdn assets

pngwn:
This PR introduces a new dedicated frontend build for the CDN.

This was relatively problematic. Vite supports a custopm base URL for preloading assets and for CSS files, and normal static `import x from y` statements use their own url to resolve dependencies but dynamic imports (`import()`), which we use for code splitting the many components we have, use the requesters URL which in our case is almost always wrong when share=True. I have fixed this by patching `import()` to use a custom `fix_import` which uses `import.meta.url` (the url of the file itself) to resolve the correct url fopr the module. This _should_ resolve all of our issues, I have tested it as best I can locally.

This doesn't require any backend changes, we just copy accross the `cdn/index.html` to `frontend/share.html` so we can utilise the current logic when `share=True`. the HTML file requests the correct assets + the CDN bundle takes care of the rest.

In order to utilise this new build we need to build twice when uploading to pypi, once for local and once for the cdn. The CDN version needs a `GRADIO_VERSION` env var passing. `pnpm build` will still build the local version but `pnpm build:cdn` will build the cdn version.

I have updated the `upload_to_pypi` script to make it work (i think).


aliabid94 (approved):
LGTM!

ui/packages/app/vite.config.js
```diff
@@ -1,19 +1,27 @@
...
 import sveltePreprocess from "svelte-preprocess";
+import path from "path";
+import fs from "fs";
 
 // this is dupe config, gonna try fix this
 import tailwind from "tailwindcss";
 import nested from "tailwindcss/nesting/index.js";
 
+const GRADIO_VERSION = process.env.GRADIO_VERSION;
+const CDN_URL = `https://gradio.s3-us-west-2.amazonaws.com/${GRADIO_VERSION}/`;
```
julien-c, line 12: FWIW S3 is not really a CDN ie. is pretty slow especially when you're far from the region (here us-west) but it'd be pretty trivial to hook a Cloudfront distribution in front of it if needed to speed things up @pngwn

pngwn: Yeah, I agree. After 3.0 I'd like to rethink our CDN stuff. It would be cool to have a nice `cdn.gradio.app` with `/` resolving to the js entrypoint (which is currently broken). Bare url would redirect to latest. `cdn.gradio.app/3/` redirecting to latest 3.x, `cdn.gradio.app/3.1/` redirecting to latest `3.1.x` etc. For users embedding gradio via CDN, needing to manually track versions is frustrating and people will get out of date quickly.

But I also want to change how we do releases (+ asset deployments). I'd like us to have everything in CI with the asset deployments IaCed with TF or something (i assume we use TF here but havent checked).

pngwn: Use of the phrase 'CDN' is very flexible here. :D But using a real CDN is the aim.

PR #1270, opened 2022-05-16
Add embedded demos to website

aliabid94:
Supports embedding all demos referenced in docstrings or guides into the website

See: http://3.139.217.3/getting_started/ (used for development and may be down)

Some frontend library changes were made to support this (@pngwn):
- Added "important: '.gradio-container'" to tailwind config and attach 'gradio-container' to blocks container element. This will prepend all tailwind css with .gradio-container to prevent leaking
- Disabled scrolling and dark mode when in website mode
- Modified cdn logic to support website cdn


abidlabs:
Really good stuff guys! Some issues that I noticed as I was playing around with the dev server link:

* The "try demo" button doesn't appear next to `Interface` (it used to appear before)
* The embedded components don't appear for some components (see `Label`, `Plot`, and a few others)
* When do you click on the button, the Code part overflows if the code is long. Same for the demo if the demo is too tall. Look at the demos for Number for example
* When you try interacting with the demos, they don't actually seem to work. For example, if you try to upload an image, it doesn't show up. Nothing happens when you click an example image, or when you click the submit button

pngwn (approved):
lgtm on the lib side. 

I think we can probably clean some of these configs up + use an alternative method for scoping styles in the future but this looks good for now.

Issue #806, opened 2022-03-11
Development Server for Gradio - Real-Time Updates in Code


ovuruska:
What about live server? Every change inside the root directory restarts server.
It will significantly improve development speed. In addition to that, people can benefit from Gradio not only a production tool but also a development tool.

create-react-app shows wonderful realization of that feature. When  a pre-specified file in the project is modified, React DOM is rerendered.

abidlabs:
It is possible to set up Gradio to do live development. Take a look at the discussion here: https://github.com/gradio-app/gradio/discussions/643

pngwn:
I think this is a request to have this supported out of the box, via some kind of command. I think the issue currently is that gradio is just a library and has little opinion on how your run the code. That would need to change. 

I do think there is some value in this, it is something I have wanted myself but probably isn't a high priority right now. People don't only interact with gradio locally, so this kind of feature would only benefit some of our users.

FarukOzderim:
Regarding this we should also make our launch of our Interface more beautiful or proper.

- We should support running application with uvicorn, ie: not launch just prepare for launch.
- I also think it would be better if we launched application with `uvicorn.launch` command instead of running in a thread as well.

PR #1278, opened 2022-05-16
Reload mode on

FarukOzderim:
# Description
I've been playing around with Gradio and uvicorn, trying to create a reload mode with nice UX. Been unsuccessful with doing it inside Gradio, as uvicorn does not allow it. 
This PR creates a simple reloading example, we can build reload mode from here, I think it might be nice to add it within 2 weeks of Gradio 3.0 Release sprint, because users will be able to play around with Gradio much more easily.

Related: #806

```
WARNING:  StatReload detected file change in 'user_mode.py'. Reloading...
WARNING:  The --reload flag should not be used in production on Windows.
INFO:     Started server process [20520]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
```


abidlabs:
@FarukOzderim this is great, and I agree we need to get reloading working. However, this looks like a WIP -- we wouldn't want users going down this route for reloading, so we shouldn't release demos around it. Instead, we can use this as an internal starting point to eventually getting reloading working & then release a proper wrapper around it.

abidlabs:
Continued @FarukOzderim's work to enable proper reloading for any gradio script. 

Usage is very simple. To run a file called "file.py" normally (not in live reload mode), we do `$ python file.py`. To get it to run in "live reload" mode, now you can do:

`$ gradio file.py`

This watches for any changes in `file.py` and restarts the server if it detects such changes. Note that by default this looks in `file.py` to find a gradio blocks / interface called `demo`. You can change the name of the object by passing in a second parameter:

`$ gradio file.py iface`

abidlabs:
Together with #1321, we now have a lot easier development workflow for Blocks

Kludex:
As mentioned on https://github.com/encode/uvicorn/discussions/1482#discussioncomment-2799159, it's not uvicorn's fault per se. 🙏

abidlabs:
@aliabid94 I added a flag in the config for any Blocks called `dev_mode`. This will be `true` if running in developer mode, and `false` when running regularly. 

Is this sufficient for you to refresh the page if the backend server "changes" or "goes down" etc.?

aliabid94:
No, I would also like an app_id exposed, maybe at route /app_id, that can be a random hash that changes when the app restarts. That way I can check if the app_id has changed periodically. Just checking if it's down is not a good solution.

ui/packages/app/src/main.ts
```diff
@@ -55,8 +55,25 @@ interface Config {
...
+		.then((response) => response.text())
+		.then((response) => {
+			if (app_id === null) {
+				app_id = response;
+			} else if (app_id != response) {
+				location.reload();
+			}
+		})
+		.finally(() => {
+			setTimeout(() => reload_check(root), 1000);
```
abidlabs, line 73:
```suggestion
			setTimeout(() => reload_check(root), 250);
```

gradio/blocks.py
```diff
@@ -228,6 +229,8 @@ def __init__(
         self.is_space = True if os.getenv("SYSTEM") == "spaces" else False
         self.favicon_path = None
         self.auth = None
+        self.dev_mode = True
```
aliabid94, line 232: why True by default?

aliabid94 (approved):
LGTM. I think we're going to have to figure out a way to prevent long steps in reload (e.g. re-importing every time, loading a model, etc), like `st.cache` from streamlit

gradio/blocks.py
abidlabs: The reason is that `uvicorn` bypasses the standard `launch()` method. So if you're using `uvicorn` to launch in debug mode, then `dev_mode` stays True, otherwise changes to `False`.

abidlabs:
> LGTM. I think we're going to have to figure out a way to prevent long steps in reload (e.g. re-importing every time, loading a model, etc), like st.cache from streamlit

That's a good point. I don't know how much control we'll have over that because uvicorn just reruns the whole file, but we can research this some more later

Thanks for the helpful review!

FarukOzderim:
LGTM as well. Very beautiful wrap-up, thx&congrats @abidlabs!

FarukOzderim:
> Continued @FarukOzderim's work to enable proper reloading for any gradio script.
> 
> Usage is very simple. To run a script called "file.py" normally (not in live reload mode), we do `$ python file.py`. To get it to run in "live reload" mode, now you can do:
> 
> `$ gradio file.py`
> 
> This watches for any changes in `file.py` and restarts the uvicorn server if it detects such changes. Note that by default this looks in `file.py` to find a gradio blocks / interface called `demo`. You can change the name of the object by passing in a second parameter. E.g.:
> 
> `$ gradio file.py iface`

Did we document this somewhere, if not lets open an issue for this.

FarukOzderim:
@abidlabs btw I cannot see an alternative to this [example](https://github.com/gradio-app/gradio/blob/a0011c0330045d4b2a6ed06af48a341731d1b338/demo/reload/dev_mode.py), let's not remove this feature and have it somewhere in the repo. Developer reload mode is also important.

gradio/dev.py
```diff
@@ -0,0 +1,35 @@
...
+        raise ValueError("No file specified.")
+    if len(args) == 1:
+        demo_name = "demo"
+    else:
+        demo_name = args[1]
+
+    path = args[0]
+    path = os.path.normpath(path)
+    if not (path == os.path.basename(path)):
+        raise ValueError("Must provide file name in the current directory")
```
FarukOzderim, line 22: I can make it work like this in the repo folder. But cannot make it work in different folders.
```
 uvicorn dev_mode:app --reload --reload-dir "gradio"
←[32mINFO←[0m:     Will watch for changes in these directories: ['F:\\SecondaryDownloads\\git_repos\\gradio\\gradio']
←[32mINFO←[0m:     Uvicorn running on ←[1mhttp://127.0.0.1:8000←[0m (Press CTRL+C to quit)
←[32mINFO←[0m:     Started reloader process [←[36m←[1m23928←[0m] using ←[36m←[1mstatreload←[0m
←[33mWARNING←[0m:  The --reload flag should not be used in production on Windows.
←[32mINFO←[0m:     Started server process [←[36m19512←[0m]
←[32mINFO←[0m:     Waiting for application startup.
←[32mINFO←[0m:     Application startup complete.
←[33mWARNING←[0m:  StatReload detected file change in 'gradio\components.py'. Reloading...
←[33mWARNING←[0m:  The --reload flag should not be used in production on Windows.
←[32mINFO←[0m:     Started server process [←[36m7704←[0m]
←[32mINFO←[0m:     Waiting for application startup.
←[32mINFO←[0m:     Application startup complete.
←[33mWARNING←[0m:  StatReload detected file change in 'gradio\components.py'. Reloading...
```

FarukOzderim: Gonna update

PR #1284, opened 2022-05-16
Dm

pngwn:
- makes dark mode work again by adding loads of selectors to increase specificty (this is really bad but will fix it some other time)
- ensure the `gradio-container` and `dark` classes are on different element to avoid specificity nightmares (`.gradio-container.dark` etc)
- final dark mode tweaks to ensure text is the correct color.


abidlabs:
Tested, looks great!

abidlabs (approved):

PR #1305, opened 2022-05-17
New big gifs

AK391:
added newer gifs for sections as discussed to make text look bigger


abidlabs:
LGTM, thank you @AK391!

PR #1308, opened 2022-05-17
Remove image editor

pngwn:
- Removes the image editor complete
- It is now possible to clear the image when in mode select


abidlabs:
Very clean!

abidlabs (approved):

PR #1310, opened 2022-05-17
fixed guides

abidlabs:
* Removed outdated API guide
* Organized guides so that quickstart, advanced Interface, and blocks guides appear first
* Removed emojis from Guides sidebar


aliabd:
LGTM

PR #1321, opened 2022-05-18
Jupyter extension for easier development with Blocks

abidlabs:
Adds a jupyter extension `gradio` and associated magic command `%%blocks` that makes it easier to do iterative development with blocks, as shown in the GIF below:

Recording 2022-05-17 at 23 15 54 (1)

### DIFF
--- a/gradio/ipython_ext.py
+++ b/gradio/ipython_ext.py
@@ -1,0 +1,74 @@
+try:
+    from IPython.core.magic import register_cell_magic
+except ImportError:
+    pass
+
+import gradio
+
+demo = gradio.Blocks()
+
+def load_ipython_extension(ipython):
+    @register_cell_magic
+    def blocks(line, cell):
+        if "imp" in line:
+            from gradio.components import (
+                HTML,
+                JSON,
+                Audio,
+                Button,
+                Carousel,
+                Chatbot,
+                Checkbox,
+                Checkboxgroup,
+                CheckboxGroup,
+                DataFrame,
+                Dataframe,
+                Dropdown,
+                File,
+                Gallery,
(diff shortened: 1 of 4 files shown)


gradio/ipython_ext.py
```diff
@@ -0,0 +1,75 @@
...
+
+import gradio
+
+
+def load_ipython_extension(ipython):
+    demo = gradio.Blocks()
+
+    @register_cell_magic
+    def blocks(line, cell):
+        if "imp" in line:
```
aliabid94, line 14: what is this checking?

gradio/ipython_ext.py
```diff
@@ -0,0 +1,75 @@
...
+            from gradio.components import (
```
aliabid94, line 15: why do we have to import everything like this? is there a more maintainable way?

gradio/ipython_ext.py
abidlabs: If you do %%blocks imp then it becomes even easier to do development as all of the top-level gradio classes become imported: 

image

gradio/ipython_ext.py
abidlabs: >Can we import everything from init.py?

Hmm I think that will create a circular import

But I'll use wildcards, I think that's pretty maintainable

abidlabs: Actually nvm: 
image

FarukOzderim:
@abidlabs  What does this PR mainly provide, could you elaborate? Does it not create a new server from scratch but update the blocks app?

FarukOzderim:
Also lets create an issue to add this to docs if we have not.

PR #1324, opened 2022-05-18
Small bugs

pngwn:
- Set the default value of Gallery to `None` for some reason this had been changed recently but an empty list is not the same as a null value. A predict function could return an empty list and we need to handle that slightly differently to `None` (which is translated to `null` in js). We display the placeholder for `null` false but not for empty lists of images.
- Adds the submit event listener to the `Textbox` when the lines are equal to 1. For some reason this got removed (it was probably me).

### DIFF
--- a/ui/packages/app/src/components/Gallery/Gallery.svelte
+++ b/ui/packages/app/src/components/Gallery/Gallery.svelte
@@ -4,7 +4,7 @@
 	import { tick } from "svelte";
 	import type { LoadingStatus } from "../StatusTracker/types";
 	import StatusTracker from "../StatusTracker/StatusTracker.svelte";
-	import ImageIcon from "./ImageIcon.svelte";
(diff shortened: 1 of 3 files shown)


ui/packages/app/src/components/Gallery/Gallery.svelte
```diff
@@ -82,16 +82,18 @@
 
 	let height = 0;
 	let window_height = 0;
+
+	$: console.log(value);
```
aliabid94, line 86: remove

aliabid94 (approved):
LGTM

aliabid94 (approved):
lgtm

ui/packages/app/src/components/Gallery/Gallery.svelte
abidlabs, line 86:
```suggestion
```

PR #1334, opened 2022-05-18
File Cached Examples Not Working Issue

dawoodkhan82:
# Description

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

Closes: # (issue)

# Checklist:


gradio/components.py
```diff
@@ -2236,9 +2236,14 @@ def save_flagged(self, dir, label, data, encryption_key):
...
-            dir, label, None if data is None else data[0]["data"], encryption_key
-        )
+        if isinstance(data, dict):
+            return self.save_flagged_file(
+                dir, label, data["data"], encryption_key, data["name"]
+            )
+        else:
+            return self.save_flagged_file(
+                dir, label, None if data is None else data[0]["data"], encryption_key
+            )
```
dawoodkhan82, line 2247: @abidlabs Do only the file tests `test_as_component_as_output` pass in a data array instead of a dict here? If so I can update the tests and remove this check.

abidlabs: I think so yeah. I would test with single files and multiple files and if everything works, go ahead and update the tests and remove the check

abidlabs:
LGTM!

abidlabs:
The only thing is that model3D demo seems to have been renamed. Renaming it back so that it is referenced correctly in the docstrings

PR #1341, opened 2022-05-19
Add processing diagram

aliabd:
Adds a processing diagram to docs so users can get a better mental model of what's going on.

Screenshot below

Screen Shot 2022-05-19 at 6 29 17 PM


abidlabs:
LGTM

abidlabs (approved):

Issue #1338, opened 2022-05-19
Model3D seems to be broken


thomwolf:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

Hi all, I have some problem to load a 3D model file locally.
When I run this
```
import gradio as gr

with gr.Blocks() as demo:
  gr.Model3D("absolute_path_to_model_file.gltf")

demo.launch()
````
The model visualizer doesn't show anything but when I drop the file directly in the interface it shows it.

I include a very simple gltf file (a box) in this post for debugging

### Reproduction

Here is the content of a ver ysimple GLTF file (a cube) to debug. You can save it as `box.gltf` for instance.

```
{
    "asset": {
        "generator": "COLLADA2GLTF",
        "version": "2.0"
    },
    "scene": 0,
    "scenes": [
        {
            "nodes": [
                0
            ]
        }
    ],
    "nodes": [
        {
            "children": [
                1
            ],
            "matrix": [
                1.0,
                0.0,
                0.0,
                0.0,
                0.0,
                0.0,
                -1.0,
                0.0,
                0.0,
                1.0,
                0.0,
                0.0,
                0.0,
                0.0,
                0.0,
                1.0
            ]
        },
        {
            "mesh": 0
        }
    ],
    "meshes": [
        {
            "primitives": [
                {
                    "attributes": {
                        "NORMAL": 1,
                        "POSITION": 2
                    },
                    "indices": 0,
                    "mode": 4,
                    "material": 0
                }
            ],
            "name": "Mesh"
        }
    ],
    "accessors": [
        {
            "bufferView": 0,
            "byteOffset": 0,
            "componentType": 5123,
            "count": 36,
            "max": [
                23
            ],
            "min": [
                0
            ],
            "type": "SCALAR"
        },
        {
            "bufferView": 1,
            "byteOffset": 0,
            "componentType": 5126,
            "count": 24,
            "max": [
                1.0,
                1.0,
                1.0
            ],
            "min": [
                -1.0,
                -1.0,
                -1.0
            ],
            "type": "VEC3"
        },
        {
            "bufferView": 1,
            "byteOffset": 288,
            "componentType": 5126,
            "count": 24,
            "max": [
                0.5,
                0.5,
                0.5
            ],
            "min": [
                -0.5,
                -0.5,
                -0.5
            ],
            "type": "VEC3"
        }
    ],
    "materials": [
        {
            "pbrMetallicRoughness": {
[...]

PR #1342, opened 2022-05-19
Model3d fix

abidlabs:
Started fixing the model3D issue in #1338 and then realized that we actually weren't handling `value` correctly. This PR goes through every single component and makes sure the values get passed in correctly so that components can always be used in "static" mode.

Closes: #1338


abidlabs:
Thanks @dawoodkhan82, merging now

PR #1363, opened 2022-05-23
Mute examples videos

gary149:
Videos examples autoplay on mouseover but maybe we want to mute them by default? (I got jump scared)
No super strong opinion on it open to discussions.
example: https://huggingface.co/spaces/radames/edit-video-by-editing-text

### DIFF
--- a/ui/packages/app/src/components/Dataset/ExampleComponents/Video.svelte
+++ b/ui/packages/app/src/components/Dataset/ExampleComponents/Video.svelte
@@ -1,9 +1,19 @@
 <script lang="ts">
 	import { playable } from "../../utils/helpers";
+	import { onMount } from "svelte";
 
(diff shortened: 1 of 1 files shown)


abidlabs:
Seems reasonable -- @aliabid94 do you have any objections or should I go ahead and merge?

radames:
I've added a commit to this, with some hacks to make it work on Safari/IOS

aliabid94:
what were the hacks to get around? @pngwn might have thoughts on how to resolve those issues, but LGTM otherwise.

aliabid94 (approved):

gary149:
> I've added a commit to this, with some hacks to make it work on Safari/IOS

Nice catch I'll let @pngwn check that.

PR #1370, opened 2022-05-24
reload-dev-mode

FarukOzderim:
- add dev reload mode
- support reload mode from different directories as well

Took some time to do it right but it is completed now. Retrieve developer-reload-mode feature that was removed in #1278 and add support for different directories.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,6 +40,11 @@ bash scripts/create_test_requirements.sh
 bash scripts/install_test_requirements.sh
 ```
 
+* You can run gradio scripts in developer reload mode which will follow the changes in the gradio folder
+```
+gradio-dev app.py
+```
+
 * Run the tests
 
 ```
--- a/gradio/reload.py
+++ b/gradio/reload.py
@@ -7,7 +7,7 @@ import sys
 from gradio import networking
 
 
-def main():
+def user():
     args = sys.argv[1:]
     if len(args) == 0:
         raise ValueError("No file specified.")
@@ -32,4 +32,32 @@ def main():
     )
     os.system(
         f"uvicorn {filename}:{demo_name}.app --reload --port {port} --log-level warning"
+    )
+
+
+def dev():
+    args = sys.argv[1:]
+    if len(args) == 0:
+        raise ValueError("No file specified.")
+    if len(args) == 1:
+        demo_name = "demo"
+    else:
+        demo_name = args[1]
+
+    path = args[0]
+    path = os.path.normpath(path)
+    if not (path == os.path.basename(path)):
+        raise ValueError("Must provide file name in the current directory")
+    filename = os.path.splitext(path)[0]
+
+    port = networking.get_first_available_port(
+        networking.INITIAL_PORT_VALUE,
+        networking.INITIAL_PORT_VALUE + networking.TRY_NUM_PORTS,
+    )
+
+    print(
+        f"\nLaunching in *reload mode* on: http://{networking.LOCALHOST_NAME}:{port} (Press CTRL+C to quit)\n"
+    )
+    os.system(
+        f"uvicorn {filename}:{demo_name}.app --reload --port {port} --log-level warning --reload-dir gradio"
     )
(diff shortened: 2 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -40,6 +40,11 @@ bash scripts/create_test_requirements.sh
 bash scripts/install_test_requirements.sh
 ```
 
+* You can run gradio scripts in developer reload mode which will follow the changes in the gradio folder
```
abidlabs, line 43:
```suggestion
* You can run gradio scripts in developer reload mode which will watch for changes in the `gradio` folder and restart the app if changes are made.
```

gradio/reload.py
```diff
@@ -0,0 +1,54 @@
...
+    args = sys.argv[1:]
+    if len(args) == 0:
+        raise ValueError("No file specified.")
+    if len(args) == 1:
+        demo_name = "demo"
+    else:
+        demo_name = args[1]
+
+    original_path = args[0]
+    path = original_path.replace("/", ".")
```
abidlabs, line 26: Needs support for windows paths as well. For example, I get an error if I try:

`$ gradio .\demo\calculator\run.py`

```
  File "c:\users\islam\appdata\local\programs\python\python39\lib\site-packages\uvicorn\importer.py", line 21, in import_from_string
    module = importlib.import_module(module_str)
  File "c:\users\islam\appdata\local\programs\python\python39\lib\importlib\__init__.py", line 122, in import_module
    raise TypeError(msg.format(name))
TypeError: the 'package' argument is required to perform a relative import for '.\\demo\\calculator\\run'
```

Also seems like leading "." might be necessary on Windows? I get an error without it:

`$ gradio demo\calculator\run.py`

```
ERROR:    Error loading ASGI app. Could not import module "demo\calculator\run".
```

The easiest way might be to convert Windows path to *NIX format.

setup.py
```diff
@@ -39,7 +39,7 @@
         "uvicorn",
         "Jinja2"
     ],
-    entry_points = {
-        'console_scripts': ['gradio=gradio.dev:main'],
+    entry_points={
+        'console_scripts': ['gradio=gradio.reload:user', 'gradio-dev=gradio.reload:dev']
```
abidlabs, line 43: We shouldn't create a separate entry point. Instead this should be a flag:

`$gradio --dev`

gradio/reload.py
```diff
@@ -0,0 +1,54 @@
...
+        )
+        os.system(
+            f"uvicorn {filename}:{demo_name}.app --reload --port {port} --log-level warning"
+        )
+    elif mode == "dev":
+        print(
+            f"\nLaunching in *developer reload mode* on: http://{networking.LOCALHOST_NAME}:{port} (Press CTRL+C to quit)\n"
+        )
+        os.system(
+            f"uvicorn {filename}:{demo_name}.app --reload --port {port} --log-level warning --reload-dir gradio --reload-dir {os.path.dirname(original_path)}"
```
abidlabs, line 45: This only works if you are running the script from the same root folder where the `gradio` folder is located. 

It should be possible to get the full path to where the gradio folder is using `inspect(gradio)`, e.g. see https://stackoverflow.com/questions/247770/how-to-retrieve-a-modules-path

abidlabs:
Great addition! Pls see comments above

gradio/reload.py
FarukOzderim: Hmm, when I tried in windows autocomplete was providing demo/calculator/run.py paths and it was working.
The problem with NX format is the char's were different than how it looks in the string and that's why I was not able to replace characters, will take a look.

setup.py
FarukOzderim: Different endpoints feels like a nice decoupling for me, and would allow auto complete for `gradio-dev` as well, I prefer this way. Might ask @aliabid94's thought as well.

```
1. gradio-dev demo/folder/run.py
2. gradio --dev demo/folder/run.py
3. gradio demo/folder/run.py
```

FarukOzderim: Btw united gradio and gradio-dev commands, from now on only `gradio demo/folder/run.py` exists

gradio/reload.py
FarukOzderim: Done!

abidlabs:
LGTM, very nice!!

Issue #1354, opened 2022-05-21
Markdown visible=False is still visible


andfanilo:
### Describe the bug

`gr.Markdown("Text", visible=False)` still display on screen.

Wrapping it in `with gr.Group(visible=False)` does not help either 😕 

Ultimately I hope to be able to run a callback event like:

```python
def hide(checked: bool):
    return gr.Markdown.update(visible=not checked)
```

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```python
import gradio as gr

def main() -> gr.Blocks:
    with gr.Blocks() as demo:
        gr.Markdown("I should be hidden :/", visible=False)

    return demo

if __name__ == "__main__":
    demo = main()
    demo.launch()
```

### Screenshot

![image](https://user-images.githubusercontent.com/5351877/169659373-0dc62b08-23cd-4102-894c-2a340fbd4adc.png)

### Logs

```shell
-
```

### System Info

```shell
Gradio 3.0.3
```

### Severity

annoying

abidlabs:
Thank you @andfanilo! Can reproduce the issue, and it seems to be affecting other components as well:

<img width="775" alt="image" src="https://user-images.githubusercontent.com/1778297/169668219-e86f3a29-f310-4b8b-874c-00a05d1c1651.png">

@aliabid94 do you know about this?

PR #1373, opened 2022-05-24
Allow users to return dicts for values

aliabid94:
Allow user to return dicts instead of lists for event listener functions. Keys are component objects themselves, values are new values. Check blocks_form for example.

Also, elem_id and visible were broken for some components, fixed those. Closes: #1354

### DIFF
--- a/demo/blocks_form/run.py
+++ b/demo/blocks_form/run.py
@@ -1,0 +1,32 @@
+import gradio as gr
+
+with gr.Blocks() as demo:
+    error_box = gr.Textbox(label="Error", visible=False)
+
+    name_box = gr.Textbox(label="Name")
+    age_box = gr.Number(label="Age")
+    symptoms_box = gr.CheckboxGroup(["Cough", "Fever", "Runny Nose"])
+    submit_btn = gr.Button("Submit")
+
+    diagnosis_box = gr.Textbox(label="Diagnosis")
+    patient_summary_box = gr.Textbox(label="Patient Summary")
+
+    def submit(name, age, symptoms):
+        if len(name) == 0:
+            return {error_box: gr.update(value="Enter name", visible=True)}
+        if age < 0 or age > 200:
+            return {error_box: gr.update(value="Enter valid age", visible=True)}
+        return {
+            diagnosis_box: "covid" if "Cough" in symptoms else "flu",
+            patient_summary_box: f"{name}, {age} y/o",
+        }
+
+    submit_btn.click(
+        submit,
+        [name_box, age_box, symptoms_box],
+        [error_box, diagnosis_box, patient_summary_box],
+    )
+
+
+if __name__ == "__main__":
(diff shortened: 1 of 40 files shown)


abidlabs:
The update() part is solid. Code and demo both look great. We should be writing tests, so it would be great if you can add a test for it.

For the missing props that you fix, I tested them and most of them working fine. However, the `visible` property in `Markdown` doesn't seem to be working. See here:

image

Updated the description to say that this closes the relevant issue.

aliabid94:
thanks for the testing, fixed markdown.

demo/blocks_form/run.py
```diff
@@ -0,0 +1,32 @@
...
+        return {
+            diagnosis_box: "covid" if "Cough" in symptoms else "flu",
+            patient_summary_box: f"{name}, {age} y/o",
+        }
+
+    submit_btn.click(
+        submit,
+        [name_box, age_box, symptoms_box],
+        [error_box, diagnosis_box, patient_summary_box],
+    )
```
FarukOzderim, line 28: I like the new way of update a lot, it is very intuitive and clear! Good job @aliabid94!

But I still have some concerns about we `gr.update`, it does the job but it does not link to the specific component, and it does not have any function signature that users could follow.

WDYT of these? I prefer 1.

1. Instead of gr.update, let's have `gr.{Component}.update`, ie. `gr.Textbox.update(value="Enter name", visible=True)`
2. (Or, have `input.update((value="Enter name", visible=True)`, in this case it would be like this:
   ```
def submit(name, age, symptoms, error_box):
        if len(name) == 0:
            return {error_box: error_box.update(value="Enter name", visible=True)}
        if age < 0 or age > 200:
            return {error_box: error_box.update(value="Enter valid age", visible=True)}
        return {
            diagnosis_box: "covid" if "Cough" in symptoms else "flu",
            patient_summary_box: f"{name}, {age} y/o",
        }

    submit_btn.click(
        submit,
        [name_box, age_box, symptoms_box, error_box],
        [error_box, diagnosis_box, patient_summary_box],
    )
```

abidlabs: > Instead of gr.update, let's have gr.{Component}.update, ie. gr.Textbox.update(value="Enter name", visible=True)

We actually do have this @FarukOzderim! `gr.update()` is a catch all, but every component defines its own `update()` which includes type hints. 

@aliabid94 let's use `Textbox.update()`, etc. in our examples

FarukOzderim: Great!

FarukOzderim: Let's update this [blocks_update](https://github.com/gradio-app/gradio/blob/main/demo/blocks_update/run.py)as well

FarukOzderim: @aliabid94 reminder in case you missed these messages.

PR #1376, opened 2022-05-25
test-big-file-diff

FarukOzderim:
- add big file for testing


FarukOzderim:
Added [sizewatcher](https://github.com/adobe/sizewatcher.git) to the repo according to its README, however the CI check does not trigger, any idea on why is it like that @pngwn, @aliabid94?

pngwn:
@FarukOzderim The `sizewatcher` workflow does not have any events which is how github actions know when to run a workflow.

Add this to the top of the file and it should work (example [here](https://github.com/gradio-app/gradio/blob/main/.github/workflows/ui.yml#L3-L11)):

```yaml
on: [pull_request]
```

This tells github to run this workflow on every pull request event. That is PR creation + all pushes to that branch in the future.

pngwn:
(if you add this workflow in this PR, it should start working immediately)

FarukOzderim:
Thx, it worked. However there is some problem in the CI, not sure what causes it, will open an issue on the sizewatcher's repo.
It did not warn even if the change was >1MB

```
Run npx @adobe/sizewatcher
npx: installed 49 in 4.036s
Checking out git branches...
Comparing changes from 'main' to 'test-big-file-diff'

Calculating size changes for
- git
fatal: ambiguous argument 'main..a86e45bac7b68838283058579351323c85bdc320': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'

Sizewatcher measured the following changes:

  'main' => 'test-big-file-diff' (sha a86e45bac7b68838283058579351323c85bdc320)

+ ❌  git: -71.8% (160 MB => 45.1 MB)



Done. Cleaning up...
```

PR #1381, opened 2022-05-25
fix default variable value

aliabid94:
Deepcopy default variable value. Otherwise separate sessions will refer to the same value for the variable if the variable is not primitive.


FarukOzderim (approved):
Good job!

abidlabs:
LGTM. Ideally we turn these demos into tests to prevent regression...

FarukOzderim:
Agree with @abidlabs, whenever we find cases similar to this, we should add necessary tests for checking the case always in the future. Though if it is an error that could not appear again in the future and testing would require complex setup, we could pass it 😸

PR #1383, opened 2022-05-25
Minor cosmetic changes

aliabid94:
Introduced some minor cosmetic changes:
- Footer stays at the bottom, is grayed out except on hover
- Button variant is "secondary" (gray) by default
- Font was never loaded properly, fell back on system fonts because index.html loaded IBM Plex, and tailwind.config required Source Sans Pro. Made both refer to IMB Plex (though we could also point both to Source Sans Pro if desired)
- Introduced minor gradient to form buttons (radio / checkboxgroup), and inner shadow to input text. This was done because the UI looked quite "flat", especially when using Blocks instead of Interface since there was no gray panel background when using Blocks. These gradients/shadows were introduced in the original Huggingface theme created by victor/julien in gradio 2.x, but were removed in Gradio 3.0 design. I believe they add a lot of visual cues to what is clickable / enterable and also help breakup the white monotony.
-  Made submit button slightly darker orange. Open to discussion on this.

BEFORE:
Screen Shot 2022-05-25 at 1 54 39 PM

AFTER:
Screen Shot 2022-05-25 at 1 55 13 PM

BEFORE GIF:
Recording 2022-05-25 at 13 58 02

AFTER GIF:
Recording 2022-05-25 at 13 58 49

Further improvements:
- I would like to suggest we also override the default slider to use a custom slider that can be styled with colors instead of using the browser default. Same for radio and checkboxes

### DIFF
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -3497,7 +3497,7 @@ class Button(Clickable, Component):
         self,
         value: str = "Run",
         *,
-        variant: str = "primary",
+        variant: str = "secondary",
         visible: bool = True,
         elem_id: Optional[str] = None,
         **kwargs,
--- a/gradio/interface.py
+++ b/gradio/interface.py
@@ -465,14 +465,14 @@ class Interface(Blocks):
                                 self.InterfaceTypes.STANDARD,
                                 self.InterfaceTypes.INPUT_ONLY,
                             ]:
-                                clear_btn = Button("Clear", variant="secondary")
+                                clear_btn = Button("Clear")
                                 if not self.live:
-                                    submit_btn = Button("Submit")
+                                    submit_btn = Button("Submit", variant="primary")
                             elif self.interface_type == self.InterfaceTypes.UNIFIED:
-                                clear_btn = Button("Clear", variant="secondary")
-                                submit_btn = Button("Submit")
+                                clear_btn = Button("Clear")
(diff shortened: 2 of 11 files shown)


abidlabs:
All of this looks good to me @aliabid94! 

The only design decision I disagree with is choosing IBM Plex. I prefer Source Sans Pro, especially on the buttons. IBM Plex doesn't look quite right on the buttons to me.

aliabid94:
Changed to Source Sans

pngwn:
All looks good to me, except the gradients. I don't think it was ever difficult to work out what was interactive as these are pretty standard UI elements, so I don't think the gradients help in that regard.

That said I intensely dislike gradients, so I'll leave this one to @gary149.

aliabid94:
another reason for gradients is that everything looks so monotonously white in my opinion in Blocks mode (like the first screenshot). The buttons on the HF website also use gradients to add a little visual character.

pngwn:
I don't really think they're noticeable enough to break up the white much. They don't add a lot of interest imo.

aliabid94:
Agreed it doesn't do that much to break up whiteness, but we do use gradients with gr.Buttons already so there's some consistency with that. Will let @gary149 make the call.

aliabid94:
Will merge for now, @gary149 let us know when you have thoughts on this.

PR #1392, opened 2022-05-27
Reload fix

abidlabs:


abidlabs:
@FarukOzderim see Slack for the issue. I had to change `original_path` to `os.path.abspath(original_path)` to get it to work.

Everything else is just because of beta releases and me printing helpful messages to the terminal in case similar issues happen up again in the future.

FarukOzderim (approved):
LGTM!, added some tweaks, as [] was not looking good.

Current Output:
`
Watching: [f:\secondarydownloads\git_repos\gradio\gradio] [F:\SecondaryDownloads\git_repos\gradio\demo\blocks_component_shortcut]
`

FarukOzderim:
Didn't know we could do beta releases on other branches than main, should document releasing and do it more frequently in cases like these.

abidlabs:
Thanks @FarukOzderim! Did you push your tweaks? I don't see them

FarukOzderim:
Oh shoot, it seems like I forgot, it seems the mind is not very careful in the time between tahajjud and fajr :D

FarukOzderim:
Btw why did we add `log-level warning`?
By default uvicorn provides watched paths:
```

INFO:     Will watch for changes in these directories: ['F:\\SecondaryDownloads\\git_repos\\gradio\\demo\\blocks_component_shortcut']
INFO:     Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
INFO:     Started reloader process [18252] using statreload
```

FarukOzderim:
@aliabid94 @abidlabs is OOO, can you merge and release to fix the bug?

abidlabs:
> Btw why did we add log-level warning? Did you not like ←[32mINFO←[0m:s in the output?

It was printing every HTTP request that was being made. And the frontend in "reload mode" makes frontend requests every 250ms to see if it should refresh the page.

julien-c:
> the frontend in "reload mode" makes frontend requests every 250ms to see if it should refresh the page.

A suggestion could be to use Server-sent events for this, they're pretty easy to do in FastAPI/ASGI

abidlabs:
> A suggestion could be to use Server-sent events for this, they're pretty easy to do in FastAPI/ASGI

@aliabid94 if you can take a look!

aliabid94:
Will merge and release to fix the issue, and take a look at server side events.

PR #1394, opened 2022-05-27
[Guide] Developer Mode

abidlabs:
Adds a Guide that covers:
* Reload Mode
* Blocks Magic


guides/developing_quickly_with_reload_mode.md
```diff
@@ -0,0 +1,91 @@
...
+If you are building Gradio Blocks using a Python IDE, your file of code (let's name it `app.py`) probably looks something like this: 
+
+```python
+import gradio as gr
+
+with gr.Blocks() as demo:
+    gr.Markdown("Hello World")
+    with gr.Tabs():
+        with gr.TabItem("Tab 1")
+            ...
```
FarukOzderim, line 24: Instead of having `...` wouldn't it be better to have a complete example?

guides/developing_quickly_with_reload_mode.md
```diff
@@ -0,0 +1,91 @@
...
+Watching...
+
+WARNING:  The --reload flag should not be used in production on Windows.
+```
+
+The important part here is the line that says `Watching...` What's happening here is that Gradio will be observing the directory where `app.py` file lives, and if the file changes, it will automatically rerun the file for you. So you can focus on writing your code, and your Gradio demo will refresh automatically 🥳
+
+⚠️ Now, there are a couple of important things to keep in mind when use the reload mode:
+
+* Because Gradio launches the demo for you in debug mode, you should *not* `launch()` your Gradio Interface or Gradio Blocks in the main block in your code. Instead, like we did above, you can `launch()` your demo within an `if __name__ == "__main__":` context, so that your code will continue to run normally if you decide to do `python app.py` as well:
```
FarukOzderim, line 50: I think we can remove this part?

1. I don't think this is correct, because both `python` and `gradio` works when demo.launch is in `__main__` or outside `__main__`. Does it not work for you for some cases?
2. The message is not very clear

guides/developing_quickly_with_reload_mode.md
```diff
@@ -0,0 +1,91 @@
...
+
+```py
+if __name__ == "__main__":
+    demo.launch()
+```
+
+* Gradio specifically looks for a Gradio Blocks/Interface demo called `demo` in your code. If you have named your demo something else, you can pass that as the 2nd parameter in your code, like this: `gradio app.py my_demo`
+
+As a small aside, this auto-reloading happens if you change your `app.py` source code, OR the Gradio source code. Meaning that this can be useful if you decide to [contribute to Gradio itself](https://github.com/gradio-app/gradio/blob/main/CONTRIBUTING.md) ✅
```
FarukOzderim, line 59:
```suggestion
As a small aside, this auto-reloading happens if you change your `app.py` source code or the Gradio source code. Meaning that this can be useful if you decide to [contribute to Gradio itself](https://github.com/gradio-app/gradio/blob/main/CONTRIBUTING.md) ✅
```

guides/developing_quickly_with_reload_mode.md
```diff
@@ -0,0 +1,91 @@
...
+
+### Jupyter Notebook Magic 🔮
```
FarukOzderim, line 61: 2 suggestions
1. Let's add a colab link to demonstrate
2. Let's add a GIF to show it immediately(that way what is happening will be more clear)

FarukOzderim (commented):
LGTM other than the mentioned points, good work!

abidlabs:
Thanks for all of the great feedback! Will update tomorrow

guides/developing_quickly_with_reload_mode.md
abidlabs: Makes sense, I've changed it.

guides/developing_quickly_with_reload_mode.md
abidlabs: Interesting, so I just tested it "kinda" works. For me, I  get the following message repeated twice:

```py
Launching in *reload mode* on: http://127.0.0.1:7861 (Press CTRL+C to quit)

Watching: 'c:\users\islam\dev\gradio-repos\gradio\gradio', 'C:\Users\islam\Desktop'

WARNING:  The --reload flag should not be used in production on Windows.
Running on local URL:  http://127.0.0.1:7862/
```

Basically what is happening is that two Gradio Interfaces are launched, but only the 2nd one is actually running in debug mode.

abidlabs: For simpliclity, I'll remove this part of the Guide and hopefully my example is enough to show people what the "correct usage" should be

guides/developing_quickly_with_reload_mode.md
abidlabs: Done

guides/developing_faster_with_reload_mode.md
```diff
@@ -0,0 +1,102 @@
...
+out = gr.Textbox()
+
+inp.change(fn=lambda x: f"Welcome, {x}!", 
+           inputs=inp, 
+           outputs=out)
+```
+
+Notice that:
+
+* You do not need to put the boiler plate `with gr.Blocks() as demo:` and `demo.launch()` code — Gradio does that for you automatically!
```
Seedmanc, line 83: But how do I pass the params that used to be set up in those calls?

PR #1361, opened 2022-05-23
Getting Interface.load() working for 2.x and 3.x models and Spaces

abidlabs:
Took waaay longer than I thought, but there were a lot of cases to cover. We can now load any model, 2.x space, or 3.x space using `gr.Interface.load()` or basically its alias `gr.Blocks.load()`. The basic idea here is:

- When loading a `model`, it creates an `Interface` based on the type of `model`
- When loading a 2.x Space, it creates an `Interface` based on the config (very easy mapping as the config is just basically the keyword arguments needed to instantiate the `Interface`)
- When loading a 3.x Space, it reads the config and recreates the Blocks and components locally. Of course the backend functions are a little different -- instead of calling the local function, it instead calls the remote /api/predict appropriately

Touches a lot of the backend code, so happy to explain anything

You can test in different scenarios:

- Loading a **model** from the Hub

```py
gr.Interface.load("models/gpt2", title="GPT 2", description="a completion model").launch()
```

```py
gr.Interface.load("models/facebook/wav2vec2-base-960h", inputs="mic").launch()
```

- Loading a **Gradio 2.x Space** from the Hub

```py
gr.Interface.load("spaces/abidlabs/test5").launch()
```

- Loading a **Gradio 3.x Space** from the Hub

```py
gr.Interface.load("spaces/gradio/pictionary").launch()
```

- Loading a **Gradio 3.x Space** from the Hub within a Blocks

```py
with gr.Blocks() as demo:
    with gr.Tabs():
        with gr.TabItem("Pictionary"):
            gr.Interface.load("spaces/gradio/pictionary")

demo.launch()
```

- Creating a Blocks that loads both a **Gradio 3.x Space** and a **Gradio 2.x Space**

```py
with gr.Blocks() as demo:
    with gr.Tabs():
        with gr.TabItem("Pictionary"):
            gr.Interface.load("spaces/gradio/pictionary")
        with gr.TabItem("Hello World"):
            gr.Interface.load("spaces/abidlabs/test5")

demo.launch()
```

 
Closes: #1248 cc @apolinario

PR #1397, opened 2022-05-27
API Docs with support for Blocks

### LINKED ISSUE #1390: Bring back API Docs
Bringing back API docs with support for both blocks and interfaces:

Backend changes:
- Adding an `api_name` parameter to event listeners. For blocks, only events with this parameter defined will be documented. For interfaces, the predict function will be documented. 
- Support the `/api/[api_name]` endpoint for events with `api_name` defined 
- Send a json with all the info required to document the endpoint to `/api/`
- Pull the documentation for components including type, samples, syntax etc from docstrings 

Frontend changes:
- Rebuild api_docs.html in svelte

aliabd:
Fixes #1390 

Backend work: 
* Added new parameter to events and trigger: `api_name`. Only events with this parameter defined will be documented. 
* Interface will have the predict function documented by default. 
* Added support for `/api/predict/{api_name}` as an endpoint (routes to `/api/predict/` with correct fn_index) 
* Added missing input/output types to component docstrings 
* JSON sent from backend to `/api/`:
```python
 docs = {
                "auth": # True or false,
                "local_login_url":  # local login url,
                "local_api_url": # local api url ,
                "endpoints": [
                     {
                     "name": # api_name,
                     "url": # /predict/{api_name}/,
                     "len_inputs": # how many inputs,
                     "len_outputs": # how many outputs,
                     "inputs":  [
                                          {
                                                   "name": # name of input component,
                                                   "type": # input type docstring,
                                                   "sample": # sample input
                                          }, ....
                                       ]
                     "outputs":  [
                                          {
                                                   "name": # name of output component,
                                                   "type": # output return type docstring,
                                          }, ....
                                       ],
                     "len_components": # number of components,
                     "components" : # inputs + outputs from above
                     }, ...
                ],
                "len_endpoints": len(endpoints)
            }
```


apolinario:
Just expanding a bit the comment @abidlabs mentioned: 
I feel with the new `gr.Interface.load()` *not* creating a function for Blocks Spaces, we ended up losing an amazing functionality  (turning any Spaces into a callable function on Python); I feel the spirit fo this feature should be maintained as one of the use-cases of the API

In this other comment I suggest how this could potentially work: https://github.com/gradio-app/gradio/pull/1361#issuecomment-1138264240

abidlabs:
We'll need to add the API doc link to the footer as well. Can we design an enhanced footer as part of this that can do things like dark mode, language, etc. https://github.com/gradio-app/gradio/issues/785

aliabd:
Looks great @aliabid94 

A few notes:
* CircleCI is failing because the config has changed, so this file will need to be updated: `/gradio/test_data/blocks_configs.py`
* The return payload in the docs is missing `average_duration` which is also returned 

I think we should merge this for now to bring back api docs support then start a new PR to work on some of the nice to haves like: 
* Code blocks for using the API in python, JS and cURL 
* Showing related methods to the user like converting image files to the correct base64 url format 
* Better design

aliabd:
@aliabid94 it doesn't seem like you're raising an error for reserved `/api/` routes, like `/api/queue/`. I was able to define `api_name="queue"`

aliabid94:
@aliabd actually /api/queue is not a reserved path because the endpoints are actually `/api/queue/push/` and `/api/queue/status/`

PR #1398, opened 2022-05-27
Fix broken link in README.md

astrocronopio:
Introduction to Blocks in line 372: The previous link does not exist in the repo, new link points to a gradio.app website page about the topic.


abidlabs:
Thank you for the fix @astrocronopio! One small thing: instead of editing the `README.md` directly, would you actually be able to make this change here: `/guides/getting_started.md`

Then run `python render_readme.py` to update the README. That way, it gets fixed everywhere. 

Appreciate the contribution!

PR #1426, opened 2022-05-31
Fix/combine adjacent

mfumanelli:
By using the HighlightedText function with combine_adjacent, text belonging to the same categories is merged and spaces are removed.

I have opened an issue with more details and will tag this pull request in it.

Closes: [#1427](https://github.com/gradio-app/gradio/issues/1427)

# Checklist:

- [ x] I have performed a self-review of my own code
- [ x] My code follows the style guidelines of this project


abidlabs:
cc @aliabid94 on this. My sense is that `HighlightedText` could take in a `delimeter` parameter, which gets inserted between every token for convenience.

abidlabs:
Yes, confirmed with @aliabid94 as well, a more flexible approach makes sense. Specifically, we should have:
* a `adjacent_separator` parameter in the `HighlightedText` constructor that is by default `""` (for backwards compatibility)
* we should use this `adjacent_separator` the way you have above to separate adjacent tokens that share the same label

@mfumanelli would you be able to make the change?

mfumanelli:
Yes @abidlabs, I can implement it.
So you want to create a parameter at the level of HighlightedText which is only used at the end when using the combine_adjacent parameter, correct?
Another question I have is, currently if you set combine_adjacent = True this happens:
Suppose we have the phrase "I live in New York", and we are looking for tokens belonging to the class "location". The application will return "I live in", written without spaces, a space, and then "New York" written without spaces. Is this behavior to be maintained? Or would it be better, by default, not to add even that space between two different classes?

abidlabs:
Thanks for the clarifications @mfumanelli. 

> So you want to create a parameter at the level of HighlightedText which is only used at the end when using the combine_adjacent parameter, correct?

Yup exactly

>  Is this behavior to be maintained? Or would it be better, by default, not to add even that space between two different classes?

I don't see a space being inserted in the backend between two different classes. It's possible that the frontend adds this space or some padding to separate different classes. As part of this PR, I wouldn't worry about it -- you can just make the backend changes.

Thank you!

mfumanelli:
Perfect, I added the parameter in the docstring too! Let me know if that is OK. Thanks 😊

abidlabs:
LGTM, thank you for the contribution @mfumanelli!

PR #1433, opened 2022-05-31
set iframe policy to enable camera and audio

### LINKED ISSUE #1406: Set IFrame feature policy for Colab environment
### Describe the bug

Related to this discussion https://github.com/gradio-app/gradio/discussions/1389 

There are two issues here,

1. Gradio generated `IFrame` doesn't have the right feature policies such as `allow="camera;microphone;`

https://github.com/gradio-app/gradio/blob/2eaf61cf7652a3ebc70ea877c33b4a66e0e0e1ce/gradio/blocks.py#L686-L697

It would be easy to add the IFrame `extras=['allow="camera;microphone;"']` parameter, however Collab is stuck to IPython 5.5.0 which doesn't support the `extras` parameter.

https://github.com/googlecolab/colabtools/issues/1582 IPython 5.50

Screen Shot 2022-05-27 at 10 45 21
ps. in the picture the `extra` is working because one can upgrade the IPython `pip install ipython -U` on colab

### Solution
One Solution is to use `IPython.display.HTML` instead of `IFrame` 

```python
 display(
    HTML(f'<iframe src="{self.share_url,}" width="{self.width}" height="{self.height}" allow="camera;microphone;">')
)
```

### Reproduction

https://colab.research.google.com/drive/1dqNIMngC1ZS7HmJvyDrsYdsBjH3P0Fpe

### Screenshot

_No response_

### Logs

_No response_

### System Info

```shell
gradio==3.0.6
```

### Severity

annoying

radames:
This PR fixes #1406. I also enable iframe features via local jupyter dev.

Notes:
* While `IPython.display.IFrame` is a better method to generate the HTML, temporary `IPython.display.HTML` address the fact google colab is using an outdated IPython https://github.com/googlecolab/colabtools/issues/1582

* The `<div>` wrapping `<iframe>` is a hack to disable the  IPython`HTML` [iframe warning](https://github.com/ipython/ipython/blob/c97f6d2ebb9a2eee5bc6531ecaf96dcb0649dadd/IPython/core/display.py#L416-L419)

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -683,17 +683,21 @@ class Blocks(BlockContext):
                     "click the link to access the interface in a new tab."
                 )
             try:
-                from IPython.display import IFrame, display  # type: ignore
+                from IPython.display import HTML, display  # type: ignore
 
                 if share:
                     while not networking.url_ok(self.share_url):
                         time.sleep(1)
                     display(
-                        IFrame(self.share_url, width=self.width, height=self.height)
+                        HTML(
+                            f'<div><iframe src="{self.share_url}" width="{self.width}" height="{self.height}" allow="autoplay; camera; microphone;" frameborder="0" allowfullscreen></iframe></div>'
+                        )
                     )
                 else:
                     display(
-                        IFrame(self.local_url, width=self.width, height=self.height)
+                        HTML(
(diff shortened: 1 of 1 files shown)


abidlabs (approved):

abidlabs (approved):
Works great!!!

PR #1434, opened 2022-05-31
set camera icon to be always text-white

radames:
Quick CSS fix for camera icon, set it to be always `text-white` as icon always has a dark background

**Before:**
Screen Shot 2022-05-31 at 13 45 56
Screen Shot 2022-05-31 at 13 46 04

**After:**

Screen Shot 2022-05-31 at 13 45 02
Screen Shot 2022-05-31 at 13 45 16


abidlabs:
Great, thank you @radames!

Issue #1417, opened 2022-05-30
Gradio 3.0 cannot load interfaces from Gradio 2


osanseviero:
### Describe the bug

Loading an interface of a Space that has <3 from a Space with Gradio 3.0 breaks

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

https://huggingface.co/spaces/osanseviero/mix_match_gradio

### Logs

```shell
Fetching interface from: https://huggingface.co/spaces/mrm8488/GPT-J-6B
/home/user/.local/lib/python3.8/site-packages/gradio/deprecation.py:40: UserWarning: `optional` parameter is deprecated, and it has no effect
  warnings.warn(value)
/home/user/.local/lib/python3.8/site-packages/gradio/deprecation.py:43: UserWarning: You have unused kwarg parameters in Textbox, please remove them: {'default': ''}
  warnings.warn(
/home/user/.local/lib/python3.8/site-packages/gradio/interface.py:282: UserWarning: Currently, only the 'default' theme is supported.
  warnings.warn("Currently, only the 'default' theme is supported.")
Fetching interface from: https://huggingface.co/spaces/akhaliq/T0pp
Traceback (most recent call last):
  File "app.py", line 4, in <module>
    iface2 = gr.Interface.load("spaces/akhaliq/T0pp")
  File "/home/user/.local/lib/python3.8/site-packages/gradio/interface.py", line 90, in load
    return super().load(name=name, src=src, api_key=api_key, alias=alias, **kwargs)
  File "/home/user/.local/lib/python3.8/site-packages/gradio/blocks.py", line 518, in load
    return external.load_blocks_from_repo(name, src, api_key, alias, **kwargs)
  File "/home/user/.local/lib/python3.8/site-packages/gradio/external.py", line 34, in load_blocks_from_repo
    blocks: gradio.Blocks = factory_methods[src](name, api_key, alias, **kwargs)
  File "/home/user/.local/lib/python3.8/site-packages/gradio/external.py", line 309, in get_spaces
    return get_spaces_blocks(model_name, config)
  File "/home/user/.local/lib/python3.8/site-packages/gradio/external.py", line 349, in get_spaces_blocks
    return gradio.Blocks.from_config(config, fns)
  File "/home/user/.local/lib/python3.8/site-packages/gradio/blocks.py", line 311, in from_config
    event_method(fn=fn, **dependency)
[...]

PR #1436, opened 2022-06-01
Fixes `gr.Interface.load()`

abidlabs:
As identified in #1417 and in other places, there was an issue with `gr.Interface.load()` not handling the `Dataset` component correctly, which was making it unable to load Spaces that were running 3.x.  This should fix it.

Now, there is a separate issue in the example with #1417, which is that when 3.x Interfaces are loaded, they are created as regular Blocks and not Interfaces, so they cannot be used with Parallel/Series. Will work on that next, along with making Blocks callable, like what @apolinario has suggested.


aliabid94:
LGTM

PR #1437, opened 2022-06-01
Callable blocks

### LINKED ISSUE #1417: Gradio 3.0 cannot load interfaces from Gradio 2

abidlabs:
This PR allows `Blocks` to be used as regular functions, just like `Interfaces`. This feature will be much nicer to use once we have `api_names` (PR #1397), but this PR sets up the basic functionality in place. 

* Here is example usage:

```py
import gradio as gr
import asyncio

demo = gr.Interface.load("spaces/gradio/question-answering")
demo("Pakistan, officially the Islamic Republic of Pakistan, is a country in South Asia.", "Where is Pakistan?")
```

Notice that calling the function returns an **async coroutine** so you must **await** it or **async.run** it. The reason for this is now we support async functions inside Blocks, so everything is treated asynchronously. This is also nice because you can call multiple api requests concurrently (yay parallelization!)

* Works with images, etc. as well:

```py
import gradio as gr

io = gr.Interface.load("spaces/gradio/Echocardiogram-Segmentation")
io("lion.jpg")
```

* How about calling non-Interface Blocks? You can specify the `fn_index` -- here's example usage:

```py
import gradio as gr

io = gr.Interface.load("spaces/abidlabs/blocks-dev-test")
```

This is a super simple Space that has 2 functions, one that doubles the word, and one that triples the word. You can call each of the functions by doing:

```py
io("a", fn_index=0)
>> "aa"
io("a", fn_index=1)
>> "aaa"
```

Right now, this requires you to know the `fn_index` for each function. This will be nicer when we can use the `api_name` instead. cc @apolinario 

-------

This PR also does two other things:

1. @dawoodkhan82 and I discovered a bug with image uploads. It fixes that in `Upload.svelte`

2. It addresses #1417 by making somewhat of an assumption (that people will only use `gradio.mix` if they load an `Interface`, and that for an `Interface`, the first blocks function is the prediction function). Although this may not always be True, the use case in #1417 is important enough such that we should support it (I've added warnings to be safe). You can test the code sample in #1417 to make sure it works:

```py
import gradio as gr

iface1 = gr.Interface.load("spaces/mrm8488/GPT-J-6B")
iface2 = gr.Interface.load("spaces/akhaliq/T0pp")

iface3 = gr.mix.Parallel(
  iface1, iface2, 
  examples = [
    ['Which country will win the 2002 World Cup?'],
    ["A is the son's of B's uncle. What is the family relationship between A and B?"],
    ["In 2030, "],
  ],
  theme="default",
  css=".footer{display:none !important}")
  
iface3.launch()
```

cc @osanseviero 

Closes: #1417


apolinario:
I have installed this branch locally and got this issue while testing it with Dalle-Mini:
```
dallemini = gr.Interface.load("spaces/dalle-mini/dalle-mini")
```

```
Fetching interface from: https://huggingface.co/spaces/dalle-mini/dalle-mini
  File "/mnt/c/Users/poli/Documents/GitHub/gradio/gradio/external.py", line 303, in get_spaces
    config = json.loads(result.group(1))
AttributeError: 'NoneType' object has no attribute 'group'

During handling of the above exception, another exception occurred:

  File "/mnt/c/Users/poli/Documents/GitHub/gradio/app.py", line 3, in <module>
    dallemini = gr.Interface.load("spaces/dalle-mini/dalle-mini")
  File "/mnt/c/Users/poli/Documents/GitHub/gradio/gradio/interface.py", line 90, in load
  File "/mnt/c/Users/poli/Documents/GitHub/gradio/gradio/blocks.py", line 582, in load
  File "/mnt/c/Users/poli/Documents/GitHub/gradio/gradio/external.py", line 34, in load_blocks_from_repo
  File "/mnt/c/Users/poli/Documents/GitHub/gradio/gradio/external.py", line 305, in get_spaces
    raise ValueError("Could not load the Space: {}".format(model_name))
ValueError: Could not load the Space: dalle-mini/dalle-mini
```

abidlabs:
Ok so the reason that `gr.Interface.load("spaces/dalle-mini/dalle-mini")` doesn't work is that `dalle-mini` isn't actually a proper Gradio app.

It just uses the Gradio components on the frontend and then makes an API request in javascript. As a result, it can't really be used with `gr.Interface.load`

abidlabs:
Ok so the reason that `gr.Interface.load("spaces/dalle-mini/dalle-mini")` doesn't work is that `dalle-mini` isn't actually a proper Gradio app.

It just uses the Gradio components on the frontend and then makes an API request in javascript. As a result, it can't really be used with `gr.Interface.load`

abidlabs:
Added a string representation for Blocks that looks like this: 

image

abidlabs:
Also checked loading and using the top Spaces of the Week and they seemed to work well. So ready for review @aliabid94!

abidlabs:
So if you are not in an async function, you will need to do something like this:

```py
import gradio as gr

io = gr.Interface.load("spaces/abidlabs/blocks-dev-test")

asyncio.run(io("a", fn_index=0)())
```

abidlabs:
After discussing with @aliabid94, decided to make the calls synchronous instead of asynchronous. This drops support for async functions when using `__call__` (almost certainly a very small use case), but makes the API much more user-friendly.

gradio/blocks.py
```diff
@@ -310,8 +312,91 @@ def iterate_over_children(children_list):
                     event_method = getattr(original_mapping[target], trigger)
                     event_method(fn=fn, **dependency)
 
+            # Allows some use of Interface-specific methods with loaded Spaces
```
aliabid94, line 315: Instead of implementing Interface specific things in Blocks, can you overwrite __call__ in Interface?

aliabid94: You can call super().__call__ where necessary

aliabid94 (commented):
LGTM after addressing/fixing comments

gradio/blocks.py
abidlabs: This code is in `from_config()` not `__call__()`

abidlabs: In this case, I don't think it's actually appropriate to put it in Interface. Because it's actually a `Blocks` that gets created, not an `Interface`. This additional hack simply lets the `Blocks` be used in some ways that `Interfaces` are used -- it's more or less an experimental thing, not necessarily something we want to support fully right now.

gradio/blocks.py
```diff
@@ -310,8 +312,91 @@ def iterate_over_children(children_list):
...
+                else:
+                    serialized_input = block.serialize(params[i], True)
+                    serialized_params.append(serialized_input)
+        else:
+            serialized_params = params
+
+        processed_input = self.preprocess_data(fn_index, serialized_params, None)
+
+        if inspect.iscoroutinefunction(block_fn.fn):
+            raise ValueError(
```
FarukOzderim, line 352: @abidlabs why not run async function in `asyncio.run()`, I think we should support async functions as well, or this feature is incomplete.

abidlabs: The problem is that this approach wouldn't work inside of a jupyter notebook. You get this error: https://stackoverflow.com/questions/55409641/asyncio-run-cannot-be-called-from-a-running-event-loop

I wasn't sure how to fix it myself, so I left this case as unsupported for now

FarukOzderim: hmm, I see, solved smt like [this](https://github.com/FarukOzderim/Memory-Vault/blob/master/src/__init__.py#L17-L41) in the past, it might help

PR #1421, opened 2022-05-30
add-logos-to-repo

FarukOzderim:
- only encode is missing

PR #1440, opened 2022-06-01
alternative-tech-stack-look

FarukOzderim:
- positioning tweaks

Pete and Victor suggests having open source stack below in the readme. Though I quite like the previous one because of
- the positioning and layout at the beginning
- emphasizing our open source stack in the beginning
I think it can start a culture in open source community 🐱. Nevertheless I think both could work. WDYT @abidlabs and @aliabd?

> Also, I don't really get those logos it might be confusing for users. They should be at the end.

> The tech isn't the point though, OSS or not, they shouldn't be so prominent. I think it makes sense to put them below the information about the library.

see slack [thread](https://huggingface.slack.com/archives/C02QZLG8GMN/p1653921182885029)

## Old
image

## New
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,16 +3,6 @@
 </p>
 <p align="center">
     <em>Build & share delightful machine learning apps easily</em>
-<br>
-<br>
-<a href="https://hf.co/"><img src="website/homepage/src/assets/img/logos/huggingface_mini.svg" alt="huggingface" height=40></a>
-<a href="https://www.python.org/"><img src="website/homepage/src/assets/img/logos/python.svg" alt="python" height=40></a>
-<a href="https://fastapi.tiangolo.com/"><img src="website/homepage/src/assets/img/logos/fastapi.svg" alt="fastapi" height=40></a>
-<a href="https://github.com/encode/"><img src="website/homepage/src/assets/img/logos/encode.svg" alt="encode" height=40></a>
-<a href="https://svelte.dev"><img src="website/homepage/src/assets/img/logos/svelte.svg" alt="svelte" height=40></a>
-<a href="https://vitejs.dev/"><img src="website/homepage/src/assets/img/logos/vite.svg" alt="vite" height=40></a>
-<a href="https://pnpm.io/"><img src="website/homepage/src/assets/img/logos/pnpm.svg" alt="pnpm" height=40></a>
-<a href="https://tailwindcss.com/"><img src="website/homepage/src/assets/img/logos/tailwind.svg" alt="tailwind" height=40></a>
 </p>
 
 <p align="center">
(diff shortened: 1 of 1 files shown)


abidlabs:
Also `README.md` should not be edited directly. Instead, edit `readme_template.md` and run `python render_readme.py` to make the changes to the readme. Otherwise, these edits will get overridden next time `render_readme.py` is run

FarukOzderim:
> Hmmm I don't love it being located here right before the quickstart. As someone reading through the readme, it's not clear what these logos represent. A user might think, _should I use Gradio if I'm already using one of these libraries_? I also don't think that we should put more content _before_ the quickstart as we want users to get to the quickstart as quickly as possible.
> 
> My suggestion would be to put this before or after the System Requirements section and say something like: "Gradio is built using many wonderful open-source libraries:" and then have the logos

1. How do you feel about this design btw?
image

2. The only concern I have about this is our Quickstart is way too long(like a few pages), and it would not have visibility under it.
3. A bit different context, why do we have that long Quickstart in the README? We already provide it in the webpage. WDYT about showing a trailer/short_version of the quick_start in README, then forward users to the original getting_started? I don't think we actually want people in README too long for this, should really forward them to original quickstart for a better UX, both before and after the Quickstart trailer.

abidlabs:
> 1. How do you feel about this design btw?

It's not clear what the logos represent in this design either

> 2. The only concern I have about this is our Quickstart is way too long(like a few pages), and it would not have visibility under it.

Is there a particular reason why we are pushing for this to have visibility? I don't see it as a particularly strong value to users to know what libraries Gradio is built with, so I think it's fine if appears at the bottom. 

> 3. why do we have that long Quickstart in the README

Some users may not click through to the website, so it's cool for them to know what can be done with Gradio directly from the readme.

FarukOzderim:
> > 3. why do we have that long Quickstart in the README
> 
> Some users may stumble onto the readme but not click through to the website, so it's cool for them to know what can be done with Gradio directly from the readme.

OK, got you about 1 and 2. But I don't think our README is good in terms of UX we really should direct users to webpage after a quick intro about quickstart IMO.
> We already provide it in the webpage. WDYT about showing a trailer/short_version of the quick_start in README, then forward users to the original getting_started? I don't think we actually want people in README too long for this, should really forward them to original quickstart for a better UX, both before and after the Quickstart trailer.

FarukOzderim:
OK then, this is the final look, please LMK if you approve @abidlabs. Made system reqs shorter as well.

image

abidlabs:
LGTM! Just note that you should make these changes in "readme_template.md" and then run "python render_readme.py". Otherwise, they'll get overridden when "render_readme.py" is run

FarukOzderim:
Made changes in #1421, closing this PR.

Issue #1442, opened 2022-06-01
Allow increase of max parallel requests


pcuenca:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
The maximum number of threads being processed in parallel is set to `40` (see below for the reason why). This creates a contention queue at the server, even if the payload function is capable of more parallelism, and despite the server `backlog` being set to 2048 by default. This is happening in our dalle-mini space, where requests are being sent to a backend server that runs behind a load balancer: under high-load situations requests wait at the gradio server even though our server could absorb them.

**Describe the solution you'd like**  
- Create a mechanism to increase the limit.
- (Optionally) Create an option, possibly in `launch`, to set the desired limit.

**Additional context**  
The limit happens when requesting predictions using `await run_in_threadpool(block_fn.fn, *processed_input)` in `blocks.py`. This goes to the Starlette server, which in turn uses `anyio` to abstract threading operations. Inspecting the configuration of the default `CapacityLimiter` shows that the limit is set to 40:
```
anyio.to_thread.current_default_thread_limiter().total_tokens
```

I can work on a pull request to address this improvement.

PR #1443, opened 2022-06-01
Increase max parallel requests

pcuenca:
Please, see #1442.

So far I haven't found a better way to deal with this than monkey-patching
`anyio.to_thread.run_sync` to use a custom `CapacityLimiter`. I did try
to obtain the predictions inside a context handler instead:

```
async with self.limiter:
    predictions = await run_in_threadpool(block_fn.fn, *processed_input)
```

However, that approach didn't seem to work.

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

Closes: #1442

# Checklist:

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -12,7 +12,6 @@ from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple
 
 import anyio
 from anyio import CapacityLimiter
-from fastapi.concurrency import run_in_threadpool
 
 from gradio import encryptor, external, networking, queueing, routes, strings, utils
 from gradio.context import Context
@@ -254,16 +253,6 @@ class Blocks(BlockContext):
         self.auth = None
         self.dev_mode = True
         self.app_id = random.getrandbits(64)
-        self.thread_limit_set = False
-
-    async def set_thread_limit(self):
-        if self.thread_limit_set:
-            return
-        self.thread_limit_set = True
-        if self.max_threads is None:
-            return
-        limiter = CapacityLimiter(total_tokens=self.max_threads)
-        anyio.to_thread.run_sync = partial(anyio.to_thread.run_sync, limiter=limiter)
 
     @classmethod
     def from_config(cls, config: dict, fns: List[Callable]) -> Blocks:
@@ -333,6 +322,11 @@ class Blocks(BlockContext):
         if Context.block is not None:
             Context.block.children.extend(self.children)
 
+    async def create_limiter(self, max_threads: Optional[int]):
+        return (
+            None if max_threads is None else CapacityLimiter(total_tokens=max_threads)
+        )
+
     async def process_api(
         self,
         data: PredictBody,
@@ -352,8 +346,6 @@ class Blocks(BlockContext):
         block_fn = self.fns[fn_index]
         dependency = self.dependencies[fn_index]
 
-        await self.set_thread_limit()
-
         if block_fn.preprocess:
             processed_input = []
             for i, input_id in enumerate(dependency["inputs"]):
@@ -368,7 +360,9 @@ class Blocks(BlockContext):
         if inspect.iscoroutinefunction(block_fn.fn):
             predictions = await block_fn.fn(*processed_input)
         else:
-            predictions = await run_in_threadpool(block_fn.fn, *processed_input)
+            predictions = await anyio.to_thread.run_sync(
+                block_fn.fn, *processed_input, limiter=self.limiter
+            )
         duration = time.time() - start
         block_fn.total_runtime += duration
         block_fn.total_runs += 1
@@ -619,7 +613,7 @@ class Blocks(BlockContext):
         else:
             self.enable_queue = enable_queue or False
 
-        self.max_threads = max_threads
+        self.limiter = anyio.run(self.create_limiter, max_threads)
         self.config = self.get_config_file()
         self.share = share
         self.encrypt = encrypt
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -7,12 +7,12 @@ import random
 import sys
 import time
 import webbrowser
-from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple
-from anyio import CapacityLimiter
-import anyio
-
-from fastapi.concurrency import run_in_threadpool
 from functools import partial
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple
+
+import anyio
+from anyio import CapacityLimiter
+from fastapi.concurrency import run_in_threadpool
 
 from gradio import encryptor, external, networking, queueing, routes, strings, utils
 from gradio.context import Context
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -8,8 +8,11 @@ import sys
 import time
 import webbrowser
 from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple
+from anyio import CapacityLimiter
+import anyio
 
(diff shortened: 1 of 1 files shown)


gradio/blocks.py
```diff
@@ -251,6 +254,17 @@ def __init__(
         self.auth = None
         self.dev_mode = True
         self.app_id = random.getrandbits(64)
+        self.thread_limit_set = False
+
+    async def set_thread_limit(self):
+        if self.thread_limit_set:
```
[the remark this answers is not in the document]
pcuenca: Yes, I did not consider that case, I thought apps were just meant to invoke `launch` once. Should we clear the flag whenever `launch` is called?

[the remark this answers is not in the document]
FarukOzderim: Yes please, in the google colab launch can be invoked several times in different cells, I think we don't need a flag like this(?), and should set the `max_threads` if it is provided always with launch.

gradio/blocks.py
```diff
@@ -339,6 +353,8 @@ async def process_api(
         block_fn = self.fns[fn_index]
         dependency = self.dependencies[fn_index]
 
+        await self.set_thread_limit()
```
FarukOzderim, line 356: Why do we `set_thread_limit` in `process_api`, let's do it within launch?
If the running async function in a sync function is a problem we can handle it with anyio or asyncio, let me know if this is the case.

gradio/blocks.py
```diff
@@ -251,6 +254,17 @@ def __init__(
...
+
+    async def set_thread_limit(self):
+        if self.thread_limit_set:
+            return
+        self.thread_limit_set = True
+        if self.max_threads is None:
+            return
+        limiter = CapacityLimiter(total_tokens=self.max_threads)
+        prev_run_in_thread = anyio.to_thread.run_sync
+        anyio.to_thread.run_sync = partial(prev_run_in_thread, limiter=limiter)
```
FarukOzderim, line 267: Let me clarify this, as far as I understand whenever `anyio.to_thread.run_sync` is gets called, it will be called with our custom limiter? 

If so this would be more readable imo:
```suggestion
        limiter = CapacityLimiter(total_tokens=self.max_threads)
        anyio.to_thread.run_sync = partial(anyio.to_thread.run_sync, limiter=limiter)
```

FarukOzderim: But would this still hold if anyio is imported in another file again, ie `starlette` library? I am not sure this is a solution for what we're trying to do, did you try it out?

Maybe we could use this instead of `run_in_threadpool`
`await to_thread.run_sync(time.sleep, 5, limiter=limiter)
`

gradio/blocks.py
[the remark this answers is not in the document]
pcuenca: Sure, I'll take a look later!

gradio/blocks.py
pcuenca: > I am not sure this is a solution for what we're trying to do, did you try it out?.
I did try it out and it worked in my case. I may have overlooked other situations though.

> Maybe we could use this instead of `run_in_threadpool` `await to_thread.run_sync(time.sleep, 5, limiter=limiter)`
Sorry, I don't follow. `run_in_threadpool` goes inside starlette code and I didn't want to touch that. Why the sleep?

gradio/blocks.py
pcuenca: I agree, it's better in `launch`, I'll take a look. Also, we may have to store the original `run_sync` if we need to support multiple `launch` invocations instead of nesting a lot of partials. This patching approach does not feel right, do you have any other suggestions?

gradio/blocks.py
FarukOzderim: Sry, my bad, missed adding this explanation:
It seems you can [spawn new threads ](https://anyio.readthedocs.io/en/stable/threads.html)in anyio, and we use `run_in_threadpool` in process_api. We might change `run_in_threadpool` with the above.

pcuenca: An alternative would be to use our own `run_in_threadpool` instead of using starlette's, and provide the limiter as an additional argument.

FarukOzderim: See [process_api](https://github.com/gradio-app/gradio/blob/main/gradio/blocks.py#L353-L356)

FarukOzderim: > I did try it out and it worked in my case. I may have overlooked other situations though.
If it works, this design is ok as well!

pcuenca: > Sry, my bad, missed adding this explanation:
> It seems you can [spawn new threads ](https://anyio.readthedocs.io/en/stable/threads.html)in anyio, and we use `run_in_threadpool` in process_api. We might change `run_in_threadpool` with the above.

Oh, you mean replace the call with our own? I'm totally fine with it!

pcuenca: I'll do some testing later, in a few hours' time.

gradio/blocks.py
FarukOzderim: This [could](https://github.com/gradio-app/gradio/pull/1443#discussion_r887802872) work!

pcuenca: I just committed dfa1c56d8bbf33636cc5c878fd4bc774d2a69859.

I'll try to setup the test environment to see why circleci is failing.

gradio/blocks.py
```diff
@@ -320,6 +322,11 @@ def render(self):
         if Context.block is not None:
             Context.block.children.extend(self.children)
 
+    async def create_limiter(self, max_threads: Optional[int]):
```
FarukOzderim, line 325: Hmm, why is this a async function?

FarukOzderim:
Thank you for the good work @pcuenca!
LGTM except the last comment, might be good for you to test max_thread>40 as well. Would be happy if you could share your results too!

gradio/blocks.py
pcuenca: I got an error when I tried to do it inline in `launch`, or using a regular function. `CapacityLimiter` refuses to run outside an async context, I think it attempts to retrieve the async library currently in use, and fails:

```
Exception has occurred: AsyncLibraryNotFoundError
unknown async library, or not in async context
```

FarukOzderim: Ok then I will resolve conflicts 👍

pcuenca:
> Thank you for the good work @pcuenca!
> LGTM except the last comment, might be good for you to test max_thread>40 as well. Would be happy if you could share your results too!

Sure, I tested the following `max_thread` values using client code that performs 64 parallel requests:
- `100`: all the parallel requests are accepted by the backend. Since the payload function just sleeps for 5s, all responses are received after 5s.
- `10`: we observe that the backend receives the 64 requests 10 at a time.
- `None`: the default behavior is seen: the backend handles `40` requests at a time.

Are there any other results you are interested in?

Thanks a lot for your help!

pcuenca:
@FarukOzderim regarding the tests, `TestSeries.test_with_external` is failing locally for me. We are receiving an error response from one of the spaces: `result = {'error': 'Could not complete request to HuggingFace API, Error 429'}`.

I also got an error in https://huggingface.co/spaces/abidlabs/image-classifier when I submitted a random image.

FarukOzderim:
That's great thanks a lot! Tests in Circleci are passing, so no problem about tests!

It seems blocks.py has changed a bit since you forked, and it is not possible to resolve with easily.
Planning to add the changes in this PR to another PR, but don't want to lose your contributions. Would you like to update your main, diverge to a branch and apply the changes again?
image

pcuenca:
> Would you like to update your main, diverge to a branch and apply the changes again?

Sure!

FarukOzderim:
Never mind conflicts actually, I will resolve them in a moment!

PR #1444, opened 2022-06-01
distribute gradio as a web component

pngwn:
This change modifies gradio to be a web component. This resolves all CSS scoping issues, simplifies embedding in websites + blogs, and provides a clean standardised API that we can extend in the future (via attributes and events), to allow users greater control and cleaner integrations with gradio widgets.

A bunch of changes + plugins were required to get this working because Vite and Tailwind _really_ don't seem to like custom elements, I also had to move things around a little bit to make sure CSS is correctly applied. There is slightly different behaviour in dev + build to work around the fact that modifying Vite's dev server is far more complex than modifying the build itself, custom elements also pose problems with dev-time reloading, that situation has been avoided entirely.

I have also added another build + test app, so we can test the CDN build to ensure it is working before publishing + deploying the static assets. It isn't a perfect representation of the remote build but it is close enough. I'd like to improve that in the future.

Happy to go through this on a call if that makes reviewing easier.


aliabid94 (approved):
LGTM!

aliabid94:
@aliabd let's migrate the website to these once this PR is in.

abidlabs:
I tested this and it works great when you have a single gradio app on your page. However, with multiple apps, it crashes / loads indefinitely. Specifically, I tried this:

```html
<!DOCTYPE html>
<html lang="en">
	<head>
		<script type="module" src="https://gradio.s3-us-west-2.amazonaws.com/3.0.9b20/gradio.js"></script>
	</head>
	<body>
		<gradio-app space="nateraw/animegan-v2-for-videos"></gradio-app>
		<gradio-app space="gradio/Echocardiogram-Segmentation"></gradio-app>
	</body>
</html>
```

and I got this loading indicator running indefinitely:

image

pngwn:
@abidlabs Should be fixed now. Was due to loading status conflicts. I just need to check the default heights for loaders, I think they could cause issues when embedding in certain contexts.

Issue #1371, opened 2022-05-24
Interface.launch(share=False) silently creates a public link


EdAbati:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  

A public link is always automatically created when _launching_ a Gradio `Interface` inside a Google Colab notebook . This happens also when setting `share=False`.
I may be wrong here, but I couldn't find any part in the docs or in the logs or print statements, that flags this behaviour. The only place I saw it mentioned is: https://github.com/gradio-app/gradio/blob/e4b5d4c1b798f90467fae25507040e49f5e804bd/gradio/blocks.py#L506

I found another side effect, when executing the following in Colab:
``` python
demo = gr.Interface(fn=predict, inputs="text", outputs="text")
demo.launch(share=False)
```

Here `demo.share_url` gets populated (e..g.`'https://12345.gradio.app'`) while `demo.share` is NOT changed but it is still set to `False`.

**Describe the solution you'd like** 

- add a note in the [docs of Interface.launch](https://gradio.app/docs/#launch-header) about this behaviour
- (more permissive behaviour) add an extra print statement to tell the user that the shareable link is always created (for example after `'Colab notebook detected. To show errors in colab notebook, set debug=True in launch()'`)
- (stricter behaviour) raise an Exception if a user set `share=False` in a Colab notebook
- set `self.share` accordingly when it is changed automatically 

If you agree with the request and the approach, I can make a PR for this.
Thank you :)

abidlabs:
Hi @EdAbati this is a really good point. Thanks for creating the issue and proposing various solutions. I agree with all of these steps, and would be in favor of raising an `Exception` if a user set `share=False` in a Colab notebook. 

We should change `share` so that it is `None` by default, which defaults to `True` in colab notebooks and `False` elsewhere. If a user decides to set `share=False` explicitly in a colab, then they should get the Exception.

If you could kindly create the PR, I would be happy to review and merge.

EdAbati:
Hey @abidlabs , sorry no, not yet. These days I played a bit with Gradio for the 'Blocks Party'!
I can have a look at this tomorrow or the day after if it is ok :)

abidlabs:
Yes of course, glad you joined the Blocks Party!

PR #1449, opened 2022-06-01
Changed share argument default in .launch method

EdAbati:
# Description

This addresses what discussed in #1371.

Closes: #1371

# Checklist:

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -252,6 +252,18 @@ class Blocks(BlockContext):
         self.dev_mode = True
         self.app_id = random.getrandbits(64)
 
+    @property
+    def share(self):
+        return self._share
+
+    @share.setter
+    def share(self, value: Optional[bool]):
+        # If share is not provided, it is set to True when running in Google Colab, or False otherwise
+        if value is None:
+            self._share = True if utils.colab_check() else False
+        else:
+            self._share = value
+
     @classmethod
     def from_config(cls, config: dict, fns: List[Callable]) -> Blocks:
         """Factory method that creates a Blocks from a config and list of functions."""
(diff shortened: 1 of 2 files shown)


gradio/blocks.py
```diff
@@ -634,10 +639,12 @@ def launch(
         utils.launch_counter()
 
         # If running in a colab or not able to access localhost,
-        # automatically create a shareable link.
-        is_colab = utils.colab_check()
+        # a shareable link must be created.
         if is_colab or (_frontend and not networking.url_ok(self.local_url)):
-            share = True
+            if not self.share:
```
abidlabs, line 644: We should only raise this exception if a user explicitly sets `share=False` in colab. It looks like this exception will be raised even if they call `.launch()` without passing in an explicit value for `None`

abidlabs:
This approach looks perfect, thanks @EdAbati!

EdAbati:
Thanks @abidlabs for the super quick feedback :)
I will add some unit tests then

abidlabs:
Awesome thanks!

EdAbati:
Hi @abidlabs , I made a small change by adding a 'setter' to `Blocks.share`. I think it is a bit nicer and also easier to test.
I also added then the unit tests about that.

Let me know if I missed anything. :)

EdAbati:
Hi @abidlabs , not sure if you saw it already but 1 test fails. This is because of `test_xray`: `demo.get_config_file` seems to be affected by the unit tests I added. 
In `XRAY_CONFIG` dictionary, the `"id"` in the `components` fields are increasing the more tests I add, for example if I comment out 1 test the `"id"` of the 1st component is `2`, with 2 tests the id of the 1st component is `3`.
I tried to run the unit tests sequentially manually and they are all successful.

Do you know why the tests are not indipendent from each other? What am I missing?

abidlabs:
Ah interesting, good catch. This lack of independence is because there is a single global Context that manages all of the IDs that are defined in a Python session. Let me fix the test so that this is no longer an issue.

abidlabs:
LGTM, thank you @EdAbati for the nice contribution and tests!

EdAbati:
Thank you!

PR #1459, opened 2022-06-03
Fix cache examples for audio, video and image

dawoodkhan82:
# Description
Fixes issue where clicking on cached examples on and video and audio demos would not work.
Context:
https://huggingface.slack.com/archives/C02Q9UFS8AW/p1653432160679459

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

Closes: # (issue)

# Checklist:


dawoodkhan82:
@abidlabs Can you verify this change makes sense? This fixes cache examples, because the frontend expects base64 representations of the examples to display in the component; not the file path. This is the same fix for the Image component in PR #1455

abidlabs:
Both this PR and #1455 look reasonable to me, though I haven't tested extensively. @aliabid94 can you review them as well?

aliabid94:
We actually don't want to send the entire audio and video as a base64 because that would be a massive amount of data being processed and sent at once - we instead want the frontend to be able to buffer chunks of the source audio/video at a time. We used to have frontend support for URLs and need to bring that back.

abidlabs:
@dawoodkhan82 looks like there is a failing test here: `TestImage.test_component_functions`

There's also a failing `external.py` test but you can ignore that, as that seems to be a rate limited HF Space

dawoodkhan82:
> @dawoodkhan82 looks like there is a failing test here: `TestImage.test_component_functions`
> 
> There's also a failing `external.py` test but you can ignore that, as that seems to be a rate limited HF Space

fixed

abidlabs:
LGTM!

abidlabs (approved):

PR #1468, opened 2022-06-06
[Guide] Embedding Gradio Space on another website

abidlabs:
A guide on how to embed gradio demos on another website (as I've gotten a lot of questions on this front) using web components and iframes.

cc @pngwn @aliabid94 @douwekiela


guides/embedding_gradio_demos.md
```diff
@@ -0,0 +1,87 @@
...
+<gradio-app space="abidlabs/pytorch-image-classifier"> </gradio-app>
+```
+
+* Instead of <code>abidlabs/pytorch-image-classifier</code>, you should, of course, put the path to your Space
+
+That's it! Here's what it looks like in action:
+
+<gradio-app space="abidlabs/pytorch-image-classifier"> </gradio-app>
+
+Note: you can provide an *initial* height for your embedded demo by using the `height` parameter. This height is used to render the page appropriately until the demo loads, at which point the embedded demo takes up the height it needs for the demo. This would look like:
```
pngwn, line 50: Is it worth pointing our that the default `initial_height` is 300px?

guides/embedding_gradio_demos.md
```diff
@@ -0,0 +1,87 @@
...
+
+```html
+<gradio-app space="abidlabs/pytorch-image-classifier" height="500px"> </gradio-app>
```
pngwn, line 53: This is actually `initial_height` not `height`(https://github.com/gradio-app/gradio/blob/main/ui/packages/app/src/main.ts#L229). I didn't want to use `height` in case we actually add that later. We could change it though if it is a nice API.

pngwn (approved):
Few comments but looks good otherwise.

guides/embedding_gradio_demos.md
abidlabs, line 50:
```suggestion
Note: you can provide an *initial* height for your embedded demo by using the `initial_height` parameter. This height is used to render the page appropriately until the demo loads, at which point the embedded demo takes up the height it needs for the demo. This would look like:
```

guides/embedding_gradio_demos.md
abidlabs, line 53:
```suggestion
<gradio-app space="abidlabs/pytorch-image-classifier" initial_height="500px"> </gradio-app>
```

PR #1477, opened 2022-06-07
ensure form group styles correctly apply

pngwn:
Form group styles were broken, i think this fixes them. Although I need to take a look at that `Block` styling because the current approach is not maintainable.


ui/packages/atoms/src/Block.svelte
```diff
@@ -60,8 +60,13 @@
...
 		(typeof style.rounded === "boolean" && style.rounded);
 
-	$: rounded_style = get_styles({ rounded: rounded }, ["rounded"]).classes;
+	$: rounded_style =
+		typeof style.rounded === "boolean"
+			? get_styles({ rounded: rounded }, ["rounded"]).classes
+			: "";
 	$: visible_style = get_styles({ visible }, ["visible"]).classes;
+
+	$: console.log(explicit_call);
```
aliabid94, line 69: Remove.

aliabid94 (approved):
LGTM

abidlabs:
I'll just remove the `console.log()` and merge into `main` so that I can release a new version with all of these changes in

Issue #1016, opened 2022-04-19
Is there a way to run gradio with python 3.6?


hn2:
I am usingp ython 3.6 for other applications. Is there a way to run gradio with this version?
Right now I am getting:
SyntaxError: future feature annotations is not defined
when trying to:
import gradio as gr

abidlabs:
Nope unfortunately not as far as I know. We extensively use features that are only available in Python 3.7+

simon-mo:
I just ran into this as well. Would it make sense to make it no longer pip installable? Ideally, we can add `python_requires` in `setup.py` to avoid pip installation.

https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#python-requires

Happy to send a PR, please lmk if this is the right approach from your point of view.

abidlabs:
That sounds perfect @simon-mo and a PR would be much appreciated.

PR #1480, opened 2022-06-07
Lower bound Python version to 3.7

simon-mo:
# Description

Gradio no longer works on Python 3.6 becauses it calls `from __future__ import annotations`. Using [python_requires](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#python-requires) can help the package to be no longer pip installable on 3.6. 

Closes: #1016

# Checklist:


abidlabs (approved):
LGTM

Issue #1494, opened 2022-06-08
Remove loading indicator when using gr.Image streaming mode


nateraw:
**Is your feature request related to a problem? Please describe.**  

As mentioned in #1490 , the loading indicator gets in the way when you run the following:

```
import gradio as gr

gr.Interface(lambda x: x, gr.Image(source='webcam', streaming=True), "image", live=True).launch()
```

**Describe the solution you'd like**  

Maybe we just show each individual output image as they come in. 

**Additional context**  

Removing this loading indicator could get tricky though, especially if folks use really hefty models. They might be confused as to why its taking so long

nateraw:
Hey there @aliabd I just pip installed from source and built the frontend - this issue still exists there. Should we reopen this, or might I have done something wrong?

abidlabs:
Same for me, reopening the issue!

<img width="962" alt="image" src="https://user-images.githubusercontent.com/1778297/173713477-a61fefc8-b700-4ca2-bcc2-e7f7bc5f8ff2.png">

abidlabs:
Hi @nateraw, this should be fixed now! Thanks for bringing this to our attention, and we'll be releasing 3.1 soon :)

nateraw:
Just gave it a go - works for me as well! Looking forward to the release so I can put together a demo with it :)

hsonthach:
Hello there! It appears that the issue has resurfaced in the latest version (3.36.0). I recently tested the streaming feature, but unfortunately, I had to revert back to version 3.32.0 in order to resolve i

abidlabs:
Thanks @hsonthach I can confirm the issue, reopening

abidlabs:
We should fix this, its pretty annoying

Yash-Divekar:
Hey @abidlabs , I am new to Open source and I loved this repository. Upon reading some code I have an assumption ( Might be a wrong assumption) that this issue might be occurring due to ' encode_plot_to_base64 ' in ' processing_utils.py ' . I am not really that knowledgeable but I feel saving every frame in a video will create a lag for video. Can we just skip this saving for videos in some way to reduce this lag.

![Screenshot 2023-10-01 173638](https://github.com/gradio-app/gradio/assets/133482969/43e13ada-9669-4dd0-a191-6b73660e8cb9)

abidlabs:
Also there's an issue with the upload progress bar showing up again and again, see `stream_frames` demo:

[...]

Issue #1479, opened 2022-06-07
Stop automatic scrolling


abidlabs:
### Discussed in https://github.com/gradio-app/gradio/discussions/1456

<div type='discussions-op-text'>

<sup>Originally posted by **yuto51** June  2, 2022</sup>
From the version 3, gradio seems automatically scroll the current position on browser to the output element. How can I stop this behavior?

In case of sketch recognition in [this tutorial](https://huggingface.co/course/chapter9/1?fw=pt), when you zoom in enough to line the sketch area and answer area vertically, current position in the brower is automatically moved to the answer area just after you finished drawing.
I'd like to stop this automatic scroll behavior.
![zoom_in](https://user-images.githubusercontent.com/40570762/171778848-c1ea8213-a2dc-41ba-86c2-ecff171e7afe.png)

</div>

PR #1499, opened 2022-06-09
Scrolling by default only in Interface, remove progress animation on streaming

aliabid94:
This PR 
- changes scrolling so it only occurs by default when in a non-live Interface, though can be added to any Dependency in Blocks via `scroll_to_output` kwarg. 
- Also removes progress animation when streaming.
- Also fixed a bug where the status loader would be set to opacity 0 instead of disappearing (was there a reason for this @pngwn?) It caused a weird gap at the top of the output panel because it was still taking up space.

Fixes: #1479
Fixes: #1494.

See demo/blocks_scroll/run.py for example.


aliabid94:
Btw @pngwn seems like there's some issue with scrolling as it is. I didn't change any of the logic of scrolling itself, but if you try running demo/blocks_scroll, you'll see it's not scrolling correctly.

pngwn:
> Also fixed a bug where the status loader would be set to opacity 0 instead of disappearing (was there a reason for this @pngwn?) It caused a weird gap at the top of the output panel because it was still taking up space.

This isn't a bug. We can't scroll to an element that doesn't exist, the scroll into view works by scrolling to the status tracker which always fills the component and is always on screen, removing it prevents it from working in certain cases/ introduces race conditions.

The scrolling is broken because position relative got removed from the `Block` in certain circumstances. Will push a fix.

pngwn:
We also need to disable scrolling when embedding via a web component, I think we should probably only enable it for interfaces in spaces, everything else should require an option to be passed (even interfaces).

Issue #1465, opened 2022-06-06
Error in prediction function leads to indefinite running


osanseviero:
### Describe the bug

I have an error in the prediction function, but instead of showing ERROR, it just keeps running

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Example 

```
def predict(place, hugging_secret):
  # errors out
  print([]["elem"])

gr.Interface(
    predict,
    inputs=[
        gr.inputs.Textbox(label="Address or place name"),
        gr.inputs.Textbox(label="Hugging Secret"),
    ],
    outputs="text"
).launch(debug=True)
```

### Logs

```shell
No logs apart from the usual error
```

### System Info

```shell
Colab, 3.0.11
```

### Severity

serious, but I can work around it

PR #1500, opened 2022-06-09
Show error on error

aliabid94:
Show error on error

Fixes: #1465


abidlabs:
The ERROR message strangely leaves the labels of the output components. Is this supposed to happen:

Before the error is thrown:

image

When the error is thrown:

image

I think it would be better (if possible) to associate the error with the specific output components that are being affected instead of every component on the page.

So when I run this code:

```py
with gr.Blocks() as demo:
    a = gr.Number()
    b = gr.Number()
    c = gr.Number()
    a.change(lambda x:1/x, a, b)
    
demo.launch()
```

I would expect only the 2nd textbox to "error out"

aliabid94:
Ugh a z-index issue. lemme fix.

aliabid94:
Ok fixed, labels were z-40 so I set status_tracker to be z-50. Though I didn't understand the existing logic @pngwn, for some reason it was set to be z-10, but z-50 if cover_all was true (but this didn't even work bc the z-10 overwrote the z-50). Not sure what the intention of this logic was but I removed it.

pngwn (approved):

Issue #1497, opened 2022-06-08
username and password fields not visible in gradio == 3.0.13, are visible in <=3.0.12


charlesfrye:
### Describe the bug

The username and password entry fields are present, but not visible. They are also working: if you inspect element and enter text, you can access the app. This makes auth unusable in 3.0.13.

I have observed this bug affecting mobile and desktop simultaneously.

This is distinct to #1491, which was reported in an earlier version of gradio.

I have not checked whether this is different in dark or light mode.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Colab reproduction: https://colab.research.google.com/drive/1ipN50CRosPAN74SiHfcJppHD4fL1ijkm?usp=sharing

This issue also affects a more complex app that I run locally.

### Here's the login page in the broken case, with one of the elements inspected, and the login elements are invisible

![image](https://user-images.githubusercontent.com/10442975/172727281-b4f5c884-0355-4bc6-b3c5-6d0f201a6f16.png)

### Here's the login page for the same code with gradio decremented to 3.0.12, which looks as expected
![image](https://user-images.githubusercontent.com/10442975/172727563-1ebd552d-67c5-467c-af73-0ba4bccfbe26.png)

### Logs

```shell
/usr/local/lib/python3.7/dist-packages/gradio/deprecation.py:40: UserWarning: `optional` parameter is deprecated, and it has no effect
  warnings.warn(value)
Colab notebook detected. This cell will run indefinitely so that you can see errors and logs. To turn off, set debug=False in launch().
Running on public URL: https://56919.gradio.app

This share link expires in 72 hours. For free permanent hosting, check out Spaces (https://huggingface.co/spaces)
```

### System Info

```shell
gradio 3.0.13 fails, gradio 3.0.12 works
Chrome Version Version 101.0.4951.64 (Official Build) (x86_64)
also observed on Android Chrome
```

### Severity

blocking upgrade to latest gradio version

abidlabs:
Thanks @charlesfrye -- really good to know. It looks like some of our CSS may have gotten broken in the latest upgrade.

hpeiyan:
> 

@abidlabs Hi, there. I have setting Interface's parmas, but it's not work on share client.

examples = [[f"examples/{img}"] for img in imgs]
css = ".output-image, .input-image, .image-preview {height: 200px !important}"
iface = gr.Interface(sepia, inputs=gr.inputs.Image(type="pil", label="input image"),
[...]

Issue #1491, opened 2022-06-08
auth_message does not show on login page


dboshardy:
### Describe the bug

According to the docs, `auth_message` should "If provided, HTML message provided on login page."

However, when I add this when launching the app, nothing shows on the login page.  
```python
 demo.launch(auth_message="This is my custom auth message.", auth=('admin', '123'))
```

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Add `auth_message` as named param to `.launch()`

### Screenshot

<img width="842" alt="Screen Shot 2022-06-08 at 10 50 56 AM" src="https://user-images.githubusercontent.com/2131537/172661456-235b407f-fa4f-4ae6-9bd6-19369ce76ed8.png">

### Logs

```shell
Running on local URL:  http://127.0.0.1:7861/

To create a public link, set `share=True` in `launch()`.
```

### System Info

```shell
gradio==3.0.9
Chrome
macOS Monterey 12.2.1
```

### Severity

annoying

abidlabs:
Thanks @dboshardy I think this broke in our 3.0 release. We'll work on bringing this back. 

cc @aliabid94

PR #1501, opened 2022-06-09
Fix Login

aliabid94:
Restoring login (currently loader blocks login view), and auth message.

Fixes: #1491 
Fixes: #1497


abidlabs:
LGTM!

Issue #654, opened 2022-02-17
Adjust width / height of image preview in the Image Component?


Adeel-Intizar:
Hello, 
How can i adjust the width / height of image input in the browser instead of jupyter notebook?

Thank you

abidlabs:
Thanks @Adeel-Intizar for creating this issue. Would you kindly add more details about what you are looking to achieve?

Adeel-Intizar:
I created an interface of an object detection model, but the size of input and output image in the browser is small (height especially), so i was wondering if gradio supports the feature where i can increase / decrease the width/height of the input image size in the browser, i know it supports it in jupyter

abidlabs:
Ah got it. Can you try setting `layout="vertical"` in the `Interface()` class.?

So like: `gr.Interface(... , layout="vertical").launch()`

It will make the input and output both full width. Can you let me know if that helps?

Adeel-Intizar:
sure, I'll check it

Adeel-Intizar:
yes, it sets the with to the fullscreen, but what i actually want to do is increase the height in horizontal layout

Adeel-Intizar:
waiting for this feature... Thanks

abidlabs:
Is this something we can facilitate in Blocks?

Adeel-Intizar:
there is already an option for width and height in gr.Interface().launch(width=..., height=...), but currently it works for only jupyter notebooks, it'd be best if it can be effective in browsers as well

rhjohnstone:
Should this solution still work? It has no effect on my app.

abidlabs:
cc @aliabid94 @pngwn is this still the right way to solve this issue in 3.0?

hpeiyan:
> cc @aliabid94 @pngwn is this still the right way to solve this issue in 3.0?

It's not work in 3.0.11 or 3.0.13 version

hpeiyan:
> 

have you solved this problem?

hpeiyan:
> Should this solution still work? It has no effect on my app.

have you solved this problem?

Maxlinn:
For those who still concerned about the image size of the gallery, i come across one request to shrink the preview images inside the gallery, becuase the inputs are in low resolution, by default gradio stretch them to higher resolution and causes them to blur.

I tried `gallery.style(height="64px")` but i does nothing, also `"64", 64` has no effect.

[...]

PR #1503, opened 2022-06-09
Allow styling height for Image and Video, also dynamic logic change

aliabid94:
Allows users to style height and width for images and video through style parameter (Fixes: #654)

Also, change to dynamic logic: if a Component is not part of a dependency, but it does have a default value, then it is dynamic, not static. Needed when a user wants to place a static image / audio / video, they can just write
`gr.Image("lion.jpg")` as a static lion image appears instead of an editable lion.


abidlabs:
Just to confirm, you mean the opposite of this sentence, right?

> Also, change to dynamic logic: if a Component is not part of a dependency, but it does have a default value, then it is dynamic, not static. 

In this case, a component would be static, not dynamic?

aliabid94:
Yep I got that backward

pngwn (approved):

abidlabs:
Great, merging!

Issue #1496, opened 2022-06-08
Flagging UX redesign


charlesfrye:
### Describe the bug

I love the flagging feature as a way to collect quick feedback from users about model behavior. It becomes even more useful when that feedback is non-binary and custom, as when using `flagging_options`. Serious model evaluation and retraining workflows require feedback of at least this granularity.

But it seems that in `gradio>=3`, the `flagging_options` parameter is not being respected. I originally ran into this on my own app, but I was able to recreate it for the calculator example in the [flagging guide](https://gradio.app/using_flagging/). See the reproduction section.

Specifically, the dropdown arrow does not appear on the "Flag" button, and so there is no way to select an option. Instead, the button behaves exactly as if the `flagging_options` argument is not set. The `flag` column still appears in the `log.csv`, but it is always empty.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Reproducing Colab Notebook: https://colab.research.google.com/drive/1Xf05scd6t-R20hP1tyLg7Q8IbWUCSegQ?usp=sharing

### On gradio website (or on Colab in gradio<3), we see the flagging options
![image](https://user-images.githubusercontent.com/10442975/172723851-b17325d1-8912-4db4-aab5-bf254422a378.png)

### Running on Colab (with gradio>=3), there are no flagging options
![image](https://user-images.githubusercontent.com/10442975/172722524-ec96c92a-0926-4fdc-bbbd-f82f46cd564a.png)

### Logs

```shell
/usr/local/lib/python3.7/dist-packages/gradio/deprecation.py:40: UserWarning: `optional` parameter is deprecated, and it has no effect
  warnings.warn(value)
Colab notebook detected. This cell will run indefinitely so that you can see errors and logs. To turn off, set debug=False in launch().
Running on public URL: https://58371.gradio.app

This share link expires in 72 hours. For free permanent hosting, check out Spaces (https://huggingface.co/spaces)
```

### System Info

```shell
carefully checked in gradio `3.0.0` and `3.0.13`, observed in other 3.0.x versions
in Chrome Version 101.0.4951.64 (Official Build) (x86_64)
run on Colab and on a local Linux machine
```

### Severity

blocking upgrade to latest gradio version

abidlabs:
Thanks @charlesfrye, we need to bring flagging options back in!

abidlabs:
[...]

PR #1504, opened 2022-06-09
Restore flag options

aliabid94:
Bring back flag options

Fixes: #1496


abidlabs:
I really like this approach from a UI perspective! Let's see how our users feel about it.

image


In the long run, we'll want a more general solution to allow people to "correct" outputs (e.g. correct the output image).

abidlabs (approved):

PR #1511, opened 2022-06-09
rework visible, attach to all components

### LINKED ISSUE #1510: visible kwarg not attached to many Components
make visible kwarg work for all components

aliabid94:
Makes the `visible=` kwarg work for every component. Changes implementation to be applied at the Render element level, instead of on a per component level.

fixes #1510


aliabid94:
Done.

Issue #1507, opened 2022-06-09
The default value of a checkbox does not show up in the ui


ags-dy:
### Describe the bug

When adding a checkbox input, if the default value is set to True, it does not show up on the ui as ticked box.

When running the function, the default value is returned properly to the callback function, but the ui does not reflect the value that is passed to the callback.

If the checkbox button is clicked manually to turn the check on or off, the display matches the value that is passed to the function. 

This shows the state of the ui after running the code below.  

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```
import gradio as gr

def func3(bool_val1):
    print("inputs")
    print(f"val1: {bool_val1}")
    return str(bool_val1)

demo = gr.Interface(func3, gr.Checkbox(label="bool1", value=True), "text")
demo.launch(inbrowser=True)
```

### Screenshot

![image](https://user-images.githubusercontent.com/38684178/172897049-0473aa4a-3651-4c0f-b77f-c8fca6bd72f0.png)

### Logs

```shell
Running on local URL:  http://127.0.0.1:7860/

To create a public link, set `share=True` in `launch()`.
inputs
val1: True
```

### System Info

```shell
Gradio Version 3.0.13
Ubuntu 20.04
```

### Severity

serious, but I can work around it

abidlabs:
Confirming I can reproduce this issue. From your example, it's clear that the checkbox is not checked as it should be. However, the output still prints True, meaning that the internal state of the checkbox is checked. 

@pngwn is this something you can take a look at?

aliabid94:
I'll take a look

abidlabs:
Fixed in `main`, thanks @aliabid94! We'll release a version of `gradio` to pypi today or tomorrow that should have this in.

PR #1513, opened 2022-06-09
Fix checkbox default value

aliabid94:
Fixes: #1507 

test:
```
with gr.Blocks() as demo:
  cbox = gr.Checkbox(value=True)
  textbox = gr.Textbox("Hey There")
  cbox.change(lambda v: gr.update(visible=v), cbox, textbox)
```


abidlabs:
LGTM!

abidlabs (approved):

Issue #1469, opened 2022-06-07
Allow running an input-less interface when loading


osanseviero:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
I would prefer to not expect users to have to click "Generate" in an input-less interface

**Describe the solution you'd like**  
If I set `live=True`, I would expect an input-less interface tu automatically run.

PR #1515, opened 2022-06-10
`live=True` for input-less Interfaces

abidlabs:
* Implements `live=True` for input-less interfaces (closes: #1469)

Test code:

```py
import gradio as gr

def func3():
    return "https://images.unsplash.com/photo-1507003211169-0a1dd7228f2d?ixlib=rb-1.2.1&ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&auto=format&fit=crop&w=387&q=80"

demo = gr.Interface(func3, None, "image", live=True)
demo.launch()
```

Or a more sophisticated example:

```py
import gradio as gr

def func3():
    return "https://images.unsplash.com/photo-1507003211169-0a1dd7228f2d?ixlib=rb-1.2.1&ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&auto=format&fit=crop&w=387&q=80"

def f2(x):
    return x

demo = gr.Interface(func3, None, "image", live=True)
demo2 = gr.Interface(f2, "text", "text")


gr.TabbedInterface([demo, demo2]).launch()
```

### DIFF
--- a/gradio/interface.py
+++ b/gradio/interface.py
@@ -518,7 +518,9 @@ class Interface(Blocks):
                         if isinstance(component, Streamable):
                             if component.streaming:
                                 component.stream(
-                                    submit_fn, self.input_components, self.output_components
+                                    submit_fn,
+                                    self.input_components,
+                                    self.output_components,
                                 )
                                 continue
                             else:
--- a/gradio/interface.py
+++ b/gradio/interface.py
@@ -499,7 +499,8 @@ class Interface(Blocks):
                         with Row().style(mobile_collapse=False):
                             if self.interface_type == self.InterfaceTypes.OUTPUT_ONLY:
                                 clear_btn = Button("Clear")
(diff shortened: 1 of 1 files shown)


FarukOzderim (approved):
LGTM 👍 

One minor comment, are we sure we want to get rid of the Run button completely? Because output could be probabilistic and users might want to use the button several times for the same input.

abidlabs:
Good point! I'll bring it back

abidlabs:
"Generate" button is back. Here's an example with a probabilistic output that can be used to test:

```py
import gradio as gr
import random

def func3():
    if random.random()>0.5:
        return "Hi"
    return "Hello"

def f2(x):
    return x

demo = gr.Interface(func3, None, "text", live=True)
demo2 = gr.Interface(f2, "text", "text")


gr.TabbedInterface([demo, demo2]).launch()
```

Issue #1487, opened 2022-06-08
CSS doesn't seem to be working in gradio 3.0.13


hysts:
### Describe the bug

CSS doesn't seem to be working in gradio 3.0.13.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

gradio 3.0.13 https://huggingface.co/spaces/hysts/_sample_001
gradio 3.0.11 https://huggingface.co/spaces/hysts/_sample_002

### Logs

```shell
Running on local URL:  http://localhost:7860/

To create a public link, set `share=True` in `launch()`.
```

### System Info

```shell
gradio 3.0.13
```

### Severity

annoying

abidlabs:
Yup @pngwn this looks like it might be related to web components, since we introduced that in 3.0.12.

PR #1540, opened 2022-06-11
Combines 2 PRs for testing

abidlabs:
1. Mount the user provided css to the page. (#1532)

Closes: #1487

2. Set configurable min height (#1532)


abidlabs:
Tested with `gradio=3.0.13b15` and it looks great!

PR #1547, opened 2022-06-13
Handle async funcs within sync scopes

FarukOzderim:
# Description

1. Creates utility functions for running async coroutines within sync scopes. It was quite troublesome to create proper solutions for this, finally resolved.
2. Gives async support to using demo("input") programmatically. Which was introduced in a recent PR recently that I can't find right now 🐱


abidlabs:
@FarukOzderim is there any way we can avoid adding another dependency to `gradio`? Especially as `fsspec` has < 500 stars.

I took a quick look at the `fsspec.asyn.sync()` function and it doesn't seem to be too complex, can we just implement it as a utility function ourselves? https://filesystem-spec.readthedocs.io/en/latest/_modules/fsspec/asyn.html

FarukOzderim:
> 

I think this is a needed dependency, why implement a complex functionality that deals with async, threading and multiple event loops when there is a library handling it? I just played around for a bit, and I don't think it will be easy to get it safe&right ourselves and adding the dependency causes no problems, we should add a dependency when we need it.

abidlabs:
> Why do you want to avoid adding another dependency?

For all of the standard reasons: security, increased installation time, dependency conflict, etc. I'm not totally against adding a new dependency, but since it looks like we are using a single function, purely written in Python, from this dependency (or am I mistaken about that?), I think we should at least consider implementing it ourselves

FarukOzderim:
> I think we should at least consider implementing it ourselves
Agreed, this sounds good but
> I think this is a needed dependency, why implement a complex functionality that deals with async, threading and multiple event loops when there is a library handling it? I just played around for a bit, and I don't think it will be easy or quick to get it safe&right ourselves
I personally think that repeating complex processes is unnecessary, and resource consuming.
Open to suggestions about this or you can also take it if you think it'd help a lot or see a clean&easy way of implementing it.

aliabid94:
there was a discussion in hf tech slack channel today actually about importing vs reimplementing libraries, seems like the general idea was just reimplement to keep a library lightweight if only using a small part of it.

FarukOzderim:
I don't think we're currently a lightweight project(considering our dependencies and tarball size), and reimplementing a complex library handling multiple event loops + multithreading + async coroutines does not look like a good approach to me 😸

FarukOzderim:
Let's merge this PR for now to unblock it, and remove the fsspec dependency if we implement the functionality ourselves in the future?

abidlabs:
Just taking a look into the PR now @FarukOzderim!

gradio.egg-info/PKG-INFO
```diff
@@ -13,15 +13,36 @@ Description-Content-Type: text/markdown
...
-
-[![CircleCI](https://circleci.com/gh/gradio-app/gradio.svg?style=svg)](https://circleci.com/gh/gradio-app/gradio)  [![PyPI version](https://badge.fury.io/py/gradio.svg)](https://badge.fury.io/py/gradio)  [![codecov](https://codecov.io/gh/gradio-app/gradio/branch/master/graph/badge.svg?token=NNVPX9KEGS)](https://codecov.io/gh/gradio-app/gradio) [![PyPI - Downloads](https://img.shields.io/pypi/dm/gradio)](https://pypi.org/project/gradio/) [![Twitter Follow](https://img.shields.io/twitter/follow/gradio.svg?style=social&label=Follow)](https://twitter.com/gradio)
-
-#  Gradio: Build Machine Learning Web Apps — in Python
-
-Gradio (pronounced GRAY-dee-oh) is an open-source Python library that is used to build machine learning and data science demos and web applications. 
-
-With Gradio, you can quickly create a beautiful user interface around your machine learning models or data science workflow and let people "try it out" by dragging-and-dropping in their own images, pasting text, recording their own voice, and interacting with your demo, all through the browser.  
+    "guides/getting_started.md", AND THEN RUN: "python render_readme.py" -->
```
abidlabs, line 16: Why are these changes here?

abidlabs: Interesting ok

abidlabs (approved):

Issue #1535, opened 2022-06-10
Warning: TypeError: 'NoneType' object is not callable


malay5:
### Describe the bug

I am new to gradio and am trying to learn it. I went to the tutorial, which provided a link for a simple hello world project on google colab. I opened it and ran it, but am getting a 

`Traceback (most recent call last):
  File "/usr/lib/python3.7/asyncio/events.py", line 88, in _run
    self._context.run(self._callback, *self._args)
TypeError: 'NoneType' object is not callable`

This is the link for the code: https://colab.research.google.com/drive/18ODkJvyxHutTN0P5APWyGFO_xwNcgHDZ?usp=sharing

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

https://colab.research.google.com/drive/18ODkJvyxHutTN0P5APWyGFO_xwNcgHDZ?usp=sharing

### Logs

```shell
(<gradio.routes.App at 0x7f039d320510>,
 'http://127.0.0.1:7862/',
 'https://55491.gradio.app')Exception in callback None(<Task finishe...> result=None>)
handle: <Handle>
Traceback (most recent call last):
  File "/usr/lib/python3.7/asyncio/events.py", line 88, in _run
    self._context.run(self._callback, *self._args)
TypeError: 'NoneType' object is not callable
```

### System Info

```shell
Gradio version 3.0.14

On google colab
```

### Severity

blocking all usage of gradio

abidlabs:
I've fixed the main issue, so the code in the colab notebook should work now @malay5.

That being said, I'll keep this issue open until we can resolve the warning.

abidlabs:
Gone now!

endomorphosis:
I also saw this bug when I was using the API on my local machine, is this fix currently in the pip packages?

endomorphosis:
I installed 3.2 and I still got this error, when trying to call the API

endomorphosis:
Traceback (most recent call last):
  File "/home/barberb/anaconda3/envs/ldo/lib/python3.8/site-packages/gradio/routes.py", line 259, in run_predict
    output = await app.blocks.process_api(
  File "/home/barberb/anaconda3/envs/ldo/lib/python3.8/site-packages/gradio/blocks.py", line 687, in process_api
    predictions, duration = await self.call_function(fn_index, inputs)
  File "/home/barberb/anaconda3/envs/ldo/lib/python3.8/site-packages/gradio/blocks.py", line 604, in call_function
[...]

PR #1549, opened 2022-06-13
Eliminates Warning in `launch()`

abidlabs:
There was a warning that appeared every time that someone `launched()` an Interface. Traced it down to this line here:

```py
utils.synchronize_async(self.create_limiter, max_threads)
``` 

This function was adding a `None` callback, which was triggering a warning. I've removed that callback since we don't ever use that functionality.

Closes: #1535


FarukOzderim:
Closing the PR since this was solved #1547

PR #1559, opened 2022-06-14
Skip tests that return 429 from HF Hub API

abidlabs:
Modifies the tests that depend on HF Hub API to be skipped if the tests return a 429 (too many requests)

### DIFF
--- a/gradio/external.py
+++ b/gradio/external.py
@@ -13,6 +13,11 @@ import gradio
 from gradio import components, utils
 
 
+class TooManyRequestsError(Exception):
+    """Raised when the Hugging Face API returns a 429 status code."""
+    pass
+
+
 def load_blocks_from_repo(name, src=None, api_key=None, alias=None, **kwargs):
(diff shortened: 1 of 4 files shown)


gradio/external.py
```diff
@@ -334,6 +340,10 @@ def fn(*data):
                     try:
                         output = result["data"]
                     except KeyError:
```
aliabid94, line 342: nope, if result.status_code == 429

aliabid94 (dismissed):

gradio/external.py
abidlabs: Unfortunately, the HF Hub API seems to return a 500 error with the following message: "Could not complete request to HuggingFace API, Error 429"

abidlabs: So I don't think this is possible

aliabid94: ok if HF API doesn't support this, nvm its fine

aliabid94: I would still check for a `response.status_code == 500 and 429 in result["error"], and id

PR #1560, opened 2022-06-14
Fix images for pypi guide

abidlabs:
The images on our Pypi page were broken because we were using relative paths to images

Based on [this answer](https://stackoverflow.com/questions/41983209/how-do-i-add-images-to-a-pypi-readme-that-works-on-github), it seems that Pypi respects absolute links only.

Changed the way that the description is generated to have absolute paths to images.


setup.py
```diff
@@ -3,11 +3,14 @@
...
     from distutils.core import setup
 from pathlib import Path
+import re
 
 this_directory = Path(__file__).parent
 long_description = (this_directory / "README.md").read_text(encoding='utf8')
 
-print(long_description)
+# Replace relative paths to images with absolute paths
+long_description = re.sub("website/homepage/src/assets/img/", "https://raw.githubusercontent.com/gradio-app/gradio/main/website/homepage/src/assets/img/", long_description)
```
aliabid94, line 12: In case we move around assets or the img folder, maybe replace at a higher level, e.g.`re.sub("website/homepage", "https://raw.githubusercontent.com/gradio-app/gradio/main/website/homepage"`

aliabid94 (approved):
LGTM

Issue #1544, opened 2022-06-11
Plotly plots do not work with gradio>3.0.13


duerrsimon:
### Describe the bug

Plotly plots do not render anymore. 

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Standard example or even empty plot does not work and works unchanged in gradio 3.0.11

### Logs

```shell
`Uncaught (in promise) Error: DOM element provided is null or undefined
    getGraphDiv https://gradio.s3-us-west-2.amazonaws.com/3.0.13/assets/index.9a7d92ae.js:113
    newPlot https://gradio.s3-us-west-2.amazonaws.com/3.0.13/assets/index.9a7d92ae.js:113
    1 https://gradio.s3-us-west-2.amazonaws.com/3.0.13/assets/index.9a7d92ae.js:472
    flush https://gradio.s3-us-west-2.amazonaws.com/3.0.13/assets/index.e8a1a070.js:530`
```

### System Info

```shell
gradio 3.0.13 and upwards
```

### Severity

blocking upgrade to latest gradio version

abidlabs:
Thanks for the heads up Simon! @dawoodkhan82 can you look into this?

PR #1565, opened 2022-06-14
Fix plotly rendering in the Plot component

Ian-GL:
# Description

Please include: 
The issue was reported on #1544 . Seems like something is making Svelte not work properly with some document functions like `getElementById`. Anyhow, it also seems like the more Svelte-like way of doing DOM manipulations is via bindings, so I changed the plot div to use bindings and it now works.

plotly-working

Unfortunately, the Bokeh plots also face the same problem and this solution does not work for that case. Bokeh internally uses the div id to render with `embed_item` as it is done [here](https://github.com/gradio-app/gradio/blob/d455baca4ed982846e17b5f27b5a9429b9944f1a/ui/packages/plot/src/Plot.svelte#L58). To solve that I think either a bigger debugging to see what is messing up with the document functions or another way to render Bokeh (maybe using [components](https://docs.bokeh.org/en/latest/docs/reference/embed.html#bokeh.embed.components)?) is needed.

Closes: #1544 

# Checklist:

### DIFF
--- a/ui/packages/plot/src/Plot.svelte
+++ b/ui/packages/plot/src/Plot.svelte
@@ -14,10 +14,13 @@
 	import Plotly from "plotly.js-dist-min";
 	import { Plot as PlotIcon } from "@gradio/icons";
 	
-	import { afterUpdate, onMount } from "svelte";
+	import { afterUpdate } from "svelte";
 
 	export let value: null | string;
 	export let theme: string;
+
+	// Plotly
+	let plotDiv;
 
 	// Bokeh
 	let bokehLoaded = false
@@ -50,7 +53,6 @@
 	afterUpdate(() => {
 		if (value && value["type"] == "plotly") {
 			let plotObj = JSON.parse(value["plot"]);
-			let plotDiv = document.getElementById("plotlyDiv");
 			Plotly.newPlot(plotDiv, plotObj["data"], plotObj["layout"]);
 		} else if (value && value["type"] == "bokeh") {
 			document.getElementById("bokehDiv").innerHTML = "";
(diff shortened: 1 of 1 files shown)


ui/packages/plot/src/Plot.svelte
```diff
@@ -14,11 +14,14 @@
 	import Plotly from "plotly.js-dist-min";
 	import { Plot as PlotIcon } from "@gradio/icons";
 	
-	import { afterUpdate, onMount } from "svelte";
```
Ian-GL, line 17: `onMount` was unused

abidlabs:
Thank you @Ian-GL for opening this PR and fixing the plotly issue! @dawoodkhan82 or @pngwn would you be able to review?

Ian-GL:
@abidlabs - You're welcome. After some digging I found that the reason `getElementById` does not work is the use of a shadow DOM. For the plotly part this wasn't a big issue, however, for Bokeh it is since bokehJS [uses the function](https://github.com/bokeh/bokeh/blob/e8f01bf661459d7bbf20e28cf5e2babc5c06318d/bokehjs/src/lib/embed/dom.ts#L9) in `embed_item`. Curious to hear what you would think is the best path forward here. Maybe even contribute to bokeh so that `embed_item` can accept an HTML element and not just an id.

pngwn (approved):
LGTM.

pngwn:
Regarding Bokeh, the way these libraries work (Bokeh in particular) is really not fun at all. I'l look into those issues a little more closely when I have time. Ideally we would be able to instantiate them in a way that makes sense, and include the library without using a load of CDN scripts but the way they are packaged and distributed does not make this particularly simple.

Issue #1566, opened 2022-06-15
`gr.update()` doesn't support updating parameters that affect the preprocessing/postprocessing in backend


NimaBoscarino:
### Describe the bug

When `gr.Radio.update(choices=choices)` is called on a radio component that has `type="index"`, the radio component breaks because the underlying `self.choices` (server-side) isn't updated. So when the call to `return self.choices.index(x)` [happens here](https://github.com/gradio-app/gradio/blob/d455baca4ed982846e17b5f27b5a9429b9944f1a/gradio/components.py#L1173), an error gets thrown.

Using `type="value"` isn't affected by this. (In general, I don't think any of the underlying `self.ETC` values are actually updated, but I haven't dug too deep into the code)

~~I intend on spending some time trying to find a fix for it!~~

>[ I think the solution here needs to be that we pass in the updated component configs from the frontend into the backend when we call the /api/predict endpoint.](https://github.com/gradio-app/gradio/pull/1568#issuecomment-1157008364) - @abidlabs 

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

https://colab.research.google.com/drive/1SgBexhut4rFFg0i6BPo_o9ZBcnmrZ2yH#scrollTo=0EGvW7wzCGfW

### Logs

```shell
File "/Users/nima/.pyenv/versions/3.10.4/envs/sandbox/lib/python3.10/site-packages/gradio/components.py", line 1173, in preprocess
    return self.choices.index(x)
ValueError: 'A' is not in list
```

### System Info

```shell
Gradio 3.0.17, MacOS 12.4.
```

### Severity

annoying

abidlabs:
Thanks @NimaBoscarino for the detailed issue and for looking into a solution!

NimaBoscarino:
Un-assigning myself since I don't think I'll have time work on this in the next few weeks, so it's up for grabs!

abidlabs:
Finally closed via #5240!

AndreiSedyshev:
This is still not fixed by #5240 

This change does not work:
https://github.com/gradio-app/gradio/blob/da05e59a53bbad15e5755a47f46685da18e1031e/demo/variable_outputs/run.py

AndreiSedyshev:
> Hi @AndreiSedyshev you'll need to install from main, which you can do by following the instructions here: https://gradio.app/docs/main

Tremendous, I apologize I wasn't installing it correctly. Love this change, thank you so much

[...]

PR #1568, opened 2022-06-15
Update blocks in memory after prediction

NimaBoscarino:
# Description 

The dependencies are being updated in the front-end, but not on the server-side. This causes some issues with certain components, such as the Radio component (see #1566).

This PR adds a small step for each non-stateful dependency in `postprocess_data`, so that any prediction of type `update` will have all of its attributes copied over to the relevant block. I've tested this manually with a few components.

Note that this PR is being opened as a draft to get feedback the approach. Let me know if this seems like an appropriate solution!

Also, the only tests that I can see that hit `postprocess_data` are:

```
test.test_external.TestLoadInterface.test_numerical_to_label_space
test.test_routes.TestRoutes.test_predict_route
test.test_routes.TestRoutes.test_predict_route_without_fn_index
test.test_routes.TestRoutes.test_state
```

But these all do it through either `.load` or the API routes. Any recommendations on good ways to test this? I was thinking of doing it in `test.test_blocks.TestBlocks`.

@abidlabs @aliabd

Closes: #1566

# Checklist:

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -507,6 +507,13 @@ class Blocks(BlockContext):
                                 if prediction_value["value"] is not None
                                 else None
                             )
+                        if prediction_value["__type__"] == "update":
+                            keys_to_update = [
+                                key for key in prediction_value if key != "__type__"
+                            ]
+                            for key in keys_to_update:
+                                setattr(block, key, prediction_value[key])
+
                         output_value = prediction_value
                     else:
                         output_value = (


abidlabs:
Hmmm I don't think this will work. The problem is that this will update the component's values for *all users* of the app, since all of the users share a single backend. So if user 1 presses the Run button, it will update the state of the component in the backend, but that will also affect user 2.

I think the solution here needs to be that we pass in the updated component configs from the frontend into the backend when we call the `/api/predict` endpoint. cc-ing @aliabid94 @pngwn for their thoughts as well

abidlabs:
Here's an example where this does not work:

```py
import gradio as gr


def set_radio_options():
    choices = ["A", "B", "C"]

    return (
        gr.Radio.update(choices=choices)
    )


def set_text(input_val):
    return gr.Text.update(input_val)


demo = gr.Blocks()

with demo:
    button = gr.Button()
    radio = gr.Radio(choices=["D", "E", "F"], type="index")  # changing to type="value" works, but index doesn't
    text = gr.Text()
    

    button.click(
        fn=set_radio_options,
        inputs=[],
        outputs=[radio]
    )

    radio.change(
        set_text,
        inputs=[radio],
        outputs=[text]
    )

demo.launch()
```

If you just run the demo in a single browser tab, then there are no problems. However, if you open a second tab, and after clicking Run in the first tab, try to use the demo in the 2nd tab without clicking Run, you'll see an Error.

NimaBoscarino:
Aw, dang! I had a feeling that it couldn't be such an easy fix :sim

NimaBoscarino:
Aw, dang! I had a feeling that it couldn't be such an easy fix 🥲 I might have some more time later this week to take another shot at it, or early next week.

NimaBoscarino:
Closing since I don't think I'll have time to work on it anytime soon.

pngwn:
Going to repoen this to track + for visibility. We are ok with issues staying open even if we aren't working on them currently.

pngwn:
it is a PR not an issue. my bad,

PR #1570, opened 2022-06-15
fix autoscroll

pngwn:
Finishes off what was previously started.

Allow autoscroll _only_ when it is explicitly set on the event (as it is in Interface) _and_ when the user has not set `autoscroll` to `"false"` when embedding as a web component. This ensures that 'apps' using interface will always scroll but embedded gradio interfaces won't unless the user explicitly asks them to. This is important for the UX of embedded interfaces and blocks such as on blogs or docs.


abidlabs:
@aliabid94 would you be able to review this PR? Can't follow the frontend changes

PR #1571, opened 2022-06-15
Wrap dataframe

pngwn:
Adds the ability to enable text-wrapping on the Dataframe.

- Adds `wrap: Optional[bool] = False` as a kwarg to the dataframe.

**Screenshot:**
Screenshot 2022-06-15 at 15 18 08


abidlabs:
LGTM! Works great, both for input and output dataframes

Issue #1574, opened 2022-06-15
Support modifying interactive in component update


ID56:
- [X] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
Right now a component can be created as _non-editable_ with the `interactive=False` argument. However, it is not possible to modify this interactivity later on at runtime.

We may want to use some behavior, where depending on certain conditions, a component (like Textbox, Number, Slider, etc) will be frozen / unfrozen.

At the moment, it is technically possible to do this if the visibility of the box is updated. Instead of freezing a component, we just make it not visible, and thus no one can interact with it. But disappearing components is perhaps not an ideal approach to this sort of problem.

**Describe the solution you'd like**  
Ideally, it would be great if the following was supported:

```python
SomeComponent.update(interactive=False) # or True
```

**Additional context**  
This feature will be helpful for implementing the third test of the [7 GUI Benchmark for Gradio](https://github.com/gradio-app/7guis-gradio), `Flight Booker`.

abidlabs:
Fixed in `main`, will be out in `gradio==3.0.18` shortly!

PR #1575, opened 2022-06-16
Allows updating `interactive` property for components

abidlabs:
Right now, you cannot update a component's "interactivity" using the standard `gr.component.update()` method: some components like `Number` actually accept `interactive` as a parameter to their `update()` method, but it has no effect, while others like `Textbox` don't accept it at all and error out.

In this PR, I'm working on fixing that. Note that the `interactive` parameter is a little different than others because it is used indirectly to set a component's `mode` property, which is what actually reflects whether a component is static or dynamic. Because "mode" is only used in the frontend, we shouldn't ask backend users to change "mode" but instead update the "mode" ourselves based on what users pass in for `interactive`. 

So I've done that in this draft PR for the `Number` component. Does this look like a reasonable approach @pngwn? If so, I'll implement it for the other components as well. 

You can test with code like this:

```py
import gradio as gr

with gr.Blocks() as demo:
    a = gr.Checkbox()
    c = gr.Textbox()
    b = gr.Number()
    a.change(lambda x:(x, gr.Number.update(interactive=x)), a, [c, b])
    
demo.launch()
```

Closes: #1574

### DIFF
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -337,6 +337,7 @@ class Textbox(Changeable, Submittable, IOComponent):
         label: Optional[str] = None,
         show_label: Optional[bool] = None,
         visible: Optional[bool] = None,
+        interactive: Optional[bool] = None,
     ):
         return {
             "lines": lines,
@@ -546,7 +547,7 @@ class Number(Changeable, Submittable, IOComponent):
             "show_label": show_label,
             "visible": visible,
             "value": value,
-            "interactive": interactive,
+            "mode": "dynamic" if interactive else "static",
             "__type__": "update",
         }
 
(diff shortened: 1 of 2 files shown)


pngwn:
Yeah, this should work fine. The only thing preventing this is returning the key from `update`, everything else should already be in place (although it will need testing, there could be some edge cases).

Just need to make sure we only do it for components that actually have an interactive and non-interactive mode.

abidlabs:
Implemented and tested with the following components:
* Textbox
* Number
* Slider
* Checkbox
* Radio
* CheckboxGroup
* Dropdown
* Image
* Video
* Audio
* File
* Dataframe
* Timeseries

abidlabs:
Ready for a final review @pngwn or @aliabid94!

aliabid94:
LGTM!

miguelwon:
Hi!
Is this working with Buttons? I would to change `interactive` parameter.

pngwn:
@miguelwon Yes, this should work with `gr.Button`.

PR #1576, opened 2022-06-16
Fix visible

### LINKED ISSUE #1573: If the visibility of a component is updated, it stops responding to its own events
### Describe the bug

If the _visibility_ of any component is updated (i.e. `component.update(visible=bool_value)` ), that component stops responding to its events.

I've tried both `.update(visible=bool_value)` and `.update(value=some_value)`, and it seems the problem occurs only when visibility is updated.

I have also checked that it is not a problem with the event (i.e. it can occur for both `change` and `submit`). It will be clearer with the minimal example I've included below.

### Scenario
Let there be an input `Textbox` **a** and another `Textbox` **o**, such that **o** always displays `"Hello {a}"`.

We also have a `Number` component **s**, which turns **a**'s visibility off or on, depending on whether it is zero or non-zero respectively.

### Code
```python
import gradio as gr

def switch(s: int):
    """Textbox visibility is turned off when s = 0, and is on otherwise.""" 
    return gr.Textbox.update(visible=bool(s))

with gr.Blocks() as demo:
    with gr.Column():
        a = gr.Textbox("world", placeholder="Write something...", label="Input: a")
        o = gr.Textbox("", label="Output: o")
        s = gr.Number(1, label="Switch a on/off")

        # updating a should update o
        a.change(lambda t: f"Hello {t}", a, o)

        # a's visibility depends on s
        s.change(switch, s, a)

demo.launch(share=False, debug=True)
```

### Steps to reproduce
1. First, we can write some stuff in **a**, and verify that the outputs are occurring in **o**
2. Next, we enter 0 in the number component **s**. This should set **a**'s visibility to `False`, so it should vanish
3. We can enter 1 (or anything apart from 0) in **s**, to turn **a's** visibility back to `True`
4. Now if we repeat step 1 (entering inputs in **a**), we will no longer observe any changes in **o**. This is because **a** no longer triggers its `change` events.

### Screenshot

Working okay:
image

Turned off `Textbox` a's visibility with the switch s:
image

Made a visible again with the switch, but now it no longer responds to events:
image

### Logs

```shell
No stack trace.
```

### System Info

```shell
Gradio Version: 3.0.17, OS: Ubuntu 20.04 LTS, Browser: Firefox 100.0.2.
```

### Severity

blocking all usage of gradio

aliabid94:
Earlier I had tried to be clever earlier by removing visible from every component and just rendering on condition in Render.svelte, but this broke event listeners as per #1573 . Fixes #1573 by adding visible prop back to all components.


ui/packages/app/src/Blocks.svelte
```diff
@@ -265,6 +265,7 @@
 				target_instances.forEach(([id, { instance }]: [number, Component]) => {
 					if (handled_dependencies[i]?.includes(id) || !instance) return;
 					instance?.$on(trigger, () => {
+						console.log("go");
```
Ian-GL, line 268: This log was not removed after debugging :)

pngwn:
Hiding + showing was a little cleaner, there might be a way to fix the event listeners. I'll take a quick look, see how simple it might be.

aliabid94:
Cool. I actually had another idea that I thought could be pretty clean @pngwn - something like this in Render.svelte
```
{#if props.visible === false}
<div class="hide-next"></div>
{/if}
<svelte:component
this={component}
...
>
   ...
</svelte:component>
```
and CSS:
```
.hide-next + * {
  @apply hidden;
}
```
This way the Component is always rendered so no issues with event listeners. However, we have to make sure every component is a single element, because only the element directly after hide-next is hidden.

aliabid94:
@pngwn I'm going to merge this for now to unblock, but I'll keep the issue open if you have ideas for a better implementation

PR #1577, opened 2022-06-16
Remove query parameter from API docs url

### LINKED ISSUE #1572: Documented API should not include URL parameters
Shows wrong source URL on hf spaces

aliabid94:
The API url displayed in the api docs is incorrent, should not include url parameter. Removed. Fixes #1572


abidlabs (approved):
Tested, LGTM

PR #1585, opened 2022-06-17
Hidden fix [hotfix]

aliabid94:
hidden must be !hidden to overwrite and block or flex. hidden conflicts with flex, !hidden gives !important to hidden to it overrides flex


abidlabs:
Looks fine to me, @pngwn should probably look at it later!

abidlabs (approved):

PR #1586, opened 2022-06-17
Add wrapper on Render component and control visibility via CSS

Ian-GL:
# Description

Taking inspiration from the discussion on PR 1576, I thought that a similar solution to[ this proposed one](https://github.com/gradio-app/gradio/pull/1576#issuecomment-1157953873) was to make a wrapper in the Render component and manage the visibility there with a class, with the difference that in this case it will hide anything inside the wrapper without depending on the proximity of the rendered component.

The changes so far only include the Render file and the CSS class. Wanted to get your opinion on this before changing the component files.

@aliabid94 @pngwn 

Closes: # (issue)

# Checklist:


abidlabs:
@pngwn would you be able to review this PR?

pngwn:
As I mentioned in [this comment](https://github.com/gradio-app/gradio/pull/1576#issuecomment-1157426883) issue, I'd like to hold off on this until I've had a chance to take a look at fixing the listeners. Removing the element from the DOM is far cleaner and has other benefits and I'd like to avoid too much churn (adding this and then changing it again). 

I should get around to looking into it later this week.

pngwn:
I resolved the conflicts here and tried to get this working, and in most cases it does with a few CSS tweaks but due to how we are handling the styling of certain child elements (form groups are one example), implementing a wrapper won't work unless we change some of out foundational styles. Using `display: contents` also doesn't work particularly well and introduces new (although quite interesting) breakages.

I'm going to close this PR for now, but thanks for the time you've put into this one @Ian-GL!

PR #1599, opened 2022-06-17
ensure iframe can detect dom changes within shadow root

pngwn:
My understanding of this issue is that the iFrameResizer uses a mutation observer to check if the DOM within the iframe has changed (after which it will update the size of the iframe). However, as we are now using custom elements, and more specifically, the shadow dom, the mutation observer cannot observe the changes happening within the shadow DOM, they are hidden.

This PR adds a custom event that fires whenever the DOM within the Shadow DOM changes. The parent HTML page listens for this event and changes a random attribute which the iFrameResizer _will_ pick up. It seems to work in my testing but we'll need to release a beta to check if it actually works on spaces.


aliabid94 (approved):

abidlabs:
Sweet, I'll merge in the branch and release 3.0.19

Issue #1404, opened 2022-05-27
 'Button' object has no attribute 'postprocess'


mizoru:
### Describe the bug

I'm trying to make the interface of my demo available in 2 languages, but when I return `gr.Button.update(value='...')` I get this error:
```Traceback (most recent call last):
  File "/usr/local/lib/python3.7/dist-packages/gradio/routes.py", line 281, in predict
    output = await app.blocks.process_api(body, username, session_state)
  File "/usr/local/lib/python3.7/dist-packages/gradio/blocks.py", line 399, in process_api
    if prediction_value["value"] is not None
AttributeError: 'Button' object has no attribute 'postprocess'```

Other interface components update just fine.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```python
import gradio as gr
from transformers import pipeline, AutoModelForCTC, Wav2Vec2Processor, Wav2Vec2ProcessorWithLM

MODELS = {
    "Tatar": {"model_id": "sammy786/wav2vec2-xlsr-tatar", "has_lm": False},
    "Chuvash": {"model_id": "sammy786/wav2vec2-xlsr-chuvash", "has_lm": False},
    "Bashkir": {"model_id": "AigizK/wav2vec2-large-xls-r-300m-bashkir-cv7_opt", "has_lm": True},
    "Erzya": {"model_id": "DrishtiSharma/wav2vec2-large-xls-r-300m-myv-v1", "has_lm": True}
    }
    
# CACHED_MODELS_BY_ID = {}
    
LANGUAGES = list(MODELS.keys())

def run(input_file, language, decoding_type):

    #logger.info(f"Running ASR {language}-{model_size}-{decoding_type} for {input_file}")

    model = MODELS.get(language, None)

    
    if decoding_type == "LM" and not model["has_lm"]:
        history.append({
            "error_message": f"LM not available for {language} language :("
        })
    else:

        # model_instance = AutoModelForCTC.from_pretrained(model["model_id"])
        model_instance = CACHED_MODELS_BY_ID.get(model["model_id"], None)
        if model_instance is None:
            model_instance = AutoModelForCTC.from_pretrained(model["model_id"])
            CACHED_MODELS_BY_ID[model["model_id"]] = model_instance

        if decoding_type == "LM":
            processor = Wav2Vec2ProcessorWithLM.from_pretrained(model["model_id"])
            asr = pipeline("automatic-speech-recognition", model=model_instance, tokenizer=processor.tokenizer, 
[...]

PR #1603, opened 2022-06-18
Makes Button value gettable/settable

abidlabs:
Makes `gr.Button` a subclass of `IOComponent` instead of `Component` which gives it default preprocessing/postprocessing methods. This allows the value (label) of the button to be used as an input to a function or set via the output of a function.

Here's a quick example:

```py
with gr.Blocks() as demo:
    gr.Markdown("Counter:")
    a = gr.Button("1", variant="primary")
    a.click(lambda x: str(int(x) + 1), a, a)
    
demo.launch()
```

Closes: #1404

### DIFF
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -2690,6 +2690,73 @@ class Variable(IOComponent):
         return self
 
 
+class Button(Clickable, IOComponent):
+    """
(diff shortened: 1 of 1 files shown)


aliabid94:
LGTM

Issue #1408, opened 2022-05-28
Cannot set non-JSON serializable value as the initial value of Variable


hysts:
### Describe the bug

Non-JSON serializable value cannot be set as the initial value of Variable.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```python
import gradio as gr
import numpy as np

with gr.Blocks() as demo:
    x = gr.Variable(value=np.arange(3))
demo.launch()
```

### System Info

```shell
gradio version 3.0.5
```

### Severity

serious, but I can work around it

FarukOzderim:
Thx for the report @hysts, please also provide the stacktrace/logs next time.
It seems we can support python arrays but not numpy arrays. 

@aliabid94 you had made some changes with the Variable, WDYT about this issue? It seems it cannot create a TemplateResponse.
```

Traceback (most recent call last):
  File "F:\SecondaryDownloads\git_repos\gradio\venv\lib\site-packages\uvicorn\protocols\http\h11_impl.py", line 366, in run_asgi
    result = await app(self.scope, self.receive, self.send)
  File "F:\SecondaryDownloads\git_repos\gradio\venv\lib\site-packages\uvicorn\middleware\proxy_headers.py", line 75, in __call__
    return await self.app(scope, receive, send)
  File "F:\SecondaryDownloads\git_repos\gradio\venv\lib\site-packages\fastapi\applications.py", line 261, in __call__
    await super().__call__(scope, receive, send)
  File "F:\SecondaryDownloads\git_repos\gradio\venv\lib\site-packages\starlette\applications.py", line 112, in __call__
    await self.middleware_stack(scope, receive, send)
  File "F:\SecondaryDownloads\git_repos\gradio\venv\lib\site-packages\starlette\middleware\errors.py", line 181, in __call__
    raise exc
  File "F:\SecondaryDownloads\git_repos\gradio\venv\lib\site-packages\starlette\middleware\errors.py", line 159, in __call__
    await self.app(scope, receive, _send)
  File "F:\SecondaryDownloads\git_repos\gradio\venv\lib\site-packages\starlette\middleware\cors.py", line 84, in __call__
    await self.app(scope, receive, send)
  File "F:\SecondaryDownloads\git_repos\gradio\venv\lib\site-packages\starlette\exceptions.py", line 82, in __call__
    raise exc
[...]

Issue #1399, opened 2022-05-27
The use of 'label' for Interface Outputs No Longer Works


fmegahed:
### Describe the bug

Hello,

I may have missed this in updated docs, but your old examples/demos for using scikit-learn type models with Gradio does not seem to work. 

**Link to your Old Titanic Example:** https://colab.research.google.com/drive/1xOU3sDHs7yZjuBosbQ8Zb2oc4BegfSFX?usp=sharing  

**Resulting App with No Output Shown:** https://15047.gradio.app/

Any suggestions?

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Links are provided in the bug description

### System Info

```shell
Google Colab
```

### Severity

blocking all usage of gradio

FarukOzderim:
I can confirm this it takes forever it to run, could it be an error about dataframe?

![image](https://user-images.githubusercontent.com/44067394/171172902-e8bb7908-78bc-4856-addd-fe44d9acb919.png)

Btw we don't have a run button in the interface, but it runs with changes in the input, is it expected @abidlabs and @aliabid94?

FarukOzderim:
Running it locally, I got the stacktrace, it is similar to #1408 numpy not json parsable.
```

TypeError: Type is not JSON serializable: numpy.float64
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "F:\SecondaryDownloads\git_repos\gradio\venv\lib\site-packages\uvicorn\protocols\http\h11_impl.py", line 366, in run_asgi
    result = await app(self.scope, self.receive, self.send)
  File "F:\SecondaryDownloads\git_repos\gradio\venv\lib\site-packages\uvicorn\middleware\proxy_headers.py", line 75, in __call__
    return await self.app(scope, receive, send)
  File "F:\SecondaryDownloads\git_repos\gradio\venv\lib\site-packages\fastapi\applications.py", line 261, in __call__
    await super().__call__(scope, receive, send)
  File "F:\SecondaryDownloads\git_repos\gradio\venv\lib\site-packages\starlette\applications.py", line 112, in __call__
    await self.middleware_stack(scope, receive, send)
  File "F:\SecondaryDownloads\git_repos\gradio\venv\lib\site-packages\starlette\middleware\errors.py", line 181, in __call__
    raise exc
  File "F:\SecondaryDownloads\git_repos\gradio\venv\lib\site-packages\starlette\middleware\errors.py", line 159, in __call__
[...]

PR #1606, opened 2022-06-18
Fixes json serialization issues

abidlabs:
Fixes json serialization so it can handle numpy arrays.

Closes: #1399
Closes: #1408

### DIFF
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -2684,7 +2684,7 @@ class Variable(IOComponent):
             raise ValueError(
                 f"Cannot set initial value because type: {type(value)} is"
                 "not JSON serializable. Instead, you can set the initial "
-                "value to be None, and in your Python function, change it"
+                "value to be None, and in your Python function, change it "
                 "to be the your desired value when the function is "
                 "run initially."
             )
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -2682,7 +2682,7 @@ class Variable(IOComponent):
             orjson.dumps(self.value, option=orjson.OPT_SERIALIZE_NUMPY)
         except TypeError:
             raise ValueError(
-                f"Cannot set initial value because type: {type(value)} is"
+                f"Cannot set initial value because type: {type(value)} is "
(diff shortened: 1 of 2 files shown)


gradio/components.py
```diff
@@ -2676,13 +2677,25 @@ def __init__(
...
         """
-        self.value = value
+        self.value = orjson.loads(value) if isinstance(value, str) else value
+        try:
+            orjson.dumps(self.value, option=orjson.OPT_SERIALIZE_NUMPY)
+        except TypeError:
+            raise ValueError(
+                f"Cannot set initial value because type: {type(value)} is"
+                "not JSON serializable. Instead, you can set the initial "
+                "value to be None, and in your Python function, change it"
```
Kludex, line 2687:
```suggestion
                "value to be None, and in your Python function, change it "
```

Kludex (commented):
👀

aliabid94:
orjson was not added to requirements.txt. Do we need to add a new library? Is it not sufficient to test if a value is a numpy.*, and use .tolist() or .item() to get the python native equivalent?

FarukOzderim:
We've been already using `orjson `in routes.py and there were no problems about it.

> orjson was not added to requirements.txt. Do we need to add a new library? Is it not sufficient to test if a value is a numpy.*, and use .tolist() or .item() to get the python native equivalent?

gradio/components.py
```diff
@@ -2676,13 +2677,25 @@ def __init__(
         Parameters:
         value (Any): the initial value of the state.
         """
-        self.value = value
+        self.value = orjson.loads(value) if isinstance(value, str) else value
+        try:
+            orjson.dumps(self.value, option=orjson.OPT_SERIALIZE_NUMPY)
+        except TypeError:
+            raise ValueError(
+                f"Cannot set initial value because type: {type(value)} is"
```
abidlabs, line 2685:
```suggestion
                f"Cannot set initial value because type: {type(value)} is "
```

Issue #1388, opened 2022-05-26
Return type for empty input


datasith:
### No documentation exists for what specifically to return as an **output** when an interface's **input** is empty.

This normally wouldn't be a problem, except, when the `Clear` button is pressed, the function we defined for the interface is called. The question is then, what should we return in such case, which we can identify by inspecting the `input` and capturing the case where `type(input) == None`.

The `output` parameters allows for a few types, and in my case, I'm using a `dictionary` with labels (`string`) for the `key`'s and confidence scores as `value`'s. Naturally, I tried to return an empty dictionary `{}` in the case that the input was empty. This results in the error:

```
Traceback (most recent call last):
  File "/home/ubuntu/.local/lib/python3.10/site-packages/gradio/routes.py", line 280, in predict
    output = await app.blocks.process_api(body, username, session_state)
  File "/home/ubuntu/.local/lib/python3.10/site-packages/gradio/blocks.py", line 310, in process_api
    block.postprocess(prediction_value)
  File "/home/ubuntu/.local/lib/python3.10/site-packages/gradio/components.py", line 2650, in postprocess
    "label": sorted_pred[0][0],
IndexError: list index out of range
```

It took me a while to figure this one out, and I trust me that I tried various types of outputs, so I'm sharing in hopes that it's documented. Although this is specific to *image classification*, it might be true for other tasks as well.

```python
def classify_image_filepath(inp):
    if inp is None:
        return ''
    ...

interface = gr.Interface(fn=classify_image_filepath, 
                         ...
                         outputs=gr.Label(num_top_classes=6),
                         examples=samples)
```

Yes, that's it, we should just return an empty string. HTH!

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Explained in the description of the bug.

### Screenshot

<img width="1646" alt="Screen Shot 2022-05-26 at 11 28 51 AM" src="https://user-images.githubusercontent.com/7302561/170552838-b3f23fdd-30df-4b46-8907-bb1d219c2d6a.png">

### System Info

```shell
gradio==3.0.5 

[Amazon EC2]
Distributor ID: Ubuntu
Description:    Ubuntu 22.04 LTS
Release:        22.04
Codename:       jammy
```

### Severity

annoying

datasith:
[...]

Issue #1380, opened 2022-05-25
Better Documentation for `Model3D`


MohamedAliRashad:
When i read the docs for `Model3D` class i didn't know what file format i should upload or provide in my function for it to work.

Maybe, some elaboration in this part with an example will be more than enough

PR #1607, opened 2022-06-18
Misc fixes

abidlabs:
Does a few small misc fixes:
* Creates `gr.Json` as an alias of `gr.JSON`
* Fixes some typing and documentation (Closes: #1380)
* Allows `Label` to accept `{}` as equivalent to `None` (Closes: #1388)


aliabid94 (approved):
LGTM

PR #1609, opened 2022-06-20
Attempt to find workaround to fix bokeh plots

Ian-GL:
# Description

The Bokeh plots are currently broken. The reason for this is that bokehJS internally uses `getElementById` to get the container of the plot and render it. Since Gradio UI is using the shadow DOM to render, this step fails.

I have tried here a workaround where I added a new hidden div in the index.html file to use as a helper to render the plot, once it is rendered, then the content is appended to the actual div that should have the plot. This part is all working, unfortunately although I can see the div with the expected content, the plot is still not showing.

Closes: # (issue)

# Checklist:

### DIFF
--- a/ui/packages/app/index.html
+++ b/ui/packages/app/index.html
@@ -59,8 +59,11 @@
 			><div
 				id="root"
 				style="display: flex; flex-direction: column; flex-grow: 1"
-			></div
-		></gradio-app>
+			></div>
+			<div id="bokeh-helper">
+				<div id="bokeh-plot-helper"></div>
+			</div>
+		</gradio-app>
 
 		<script>
 			const ce = document.getElementsByTagName("gradio-app");
@@ -68,7 +71,6 @@
 				ce[0].addEventListener("domchange", () => {
 					document.body.style.padding = "0";
 				});
-				console.log("hello");
 				document.body.style.padding = "0";
 			}
 		</script>
--- a/ui/packages/plot/src/Plot.svelte
+++ b/ui/packages/plot/src/Plot.svelte
@@ -23,6 +23,7 @@
 	let plotDiv;
 
 	// Bokeh
+	let bokehDiv;
 	let bokehLoaded = false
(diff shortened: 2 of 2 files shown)


pngwn:
I'm not a big fan of this approach. Going to try to fix it upstream.

pngwn:
This change was merged in recently as a result of an issue I filed, but I'm not sure when it will be released: https://github.com/bokeh/bokeh/pull/12186

Ian-GL:
@pngwn thanks for following that closely! I went ahead and saw that is included in the dev release `3.0.0dev8`, which is already [available on CDN](https://cdnjs.com/libraries/bokeh) and pypi. I will start working on a PR using that version and once it works we can just wait until `3.0.0` is released.

abidlabs:
What if we were to check the version of bokeh and printing a warning saying that `gradio` requires `3.0.0dev8` or higher? Would that allow us to merge this PR in? cc @dawoodkhan82

dawoodkhan82:
> What if we were to check the version of bokeh and printing a warning saying that `gradio` requires `3.0.0dev8` or higher? Would that allow us to merge this PR in? cc @dawoodkhan82

That would work. @Ian-GL Is this something you can add to this PR?

Ian-GL:
@abidlabs , @dawoodkhan82 , that would need to be another PR. The changes here were just to find a workaround using Bokeh 2.4, to use `3.0.0dev8` the changes needed are to upgrade the CDN versions. I tried it but using Bokeh `3.0.0dev8` raised some errors but from the Python side.

I'll open another PR at the end of the week using the latest Bokeh version (even if it's dev), hopefully it now works, if not I could use some help solving the Python-side issue.

I think this PR can be closed given that it is not useful since the way to go is to use a newer bokeh version.

abidlabs:
Thank you very much @Ian-GL for looking into this! Sounds good, will go ahead and close this PR & open a new issue to fix bokeh plots.

Issue #1587, opened 2022-06-17
Gallery component doesn't display default images


NimaBoscarino:
### Describe the bug

The initial images passed as values to a Gallery component are not displayed, either in a Block or an Interface. If the Gallery values are updated, then the Gallery renders the images.

I dug back to previous commits and isolated it to this one (https://github.com/gradio-app/gradio/commit/74703898c636c4e2a233880feac3eb239b4f8a2b) 🤔 
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

https://colab.research.google.com/drive/1KEFsoOxWw_dp9w--fkh5lKITJgK4VRXK?usp=sharing

### Logs

```shell
N/A
```

### System Info

```shell
Gradio 3.0.18, macOS 12.4 and Google Colab
```

### Severity

annoying

aliabid94:
Thanks for raising this. Will take a look.

PR #1614, opened 2022-06-20
Restore get_config function for Gallery component

Ian-GL:
# Description

The `get_config` function from the Gallery component had been erased, causing the default images set to not be shown.

Closes: #1587 

# Checklist:

### DIFF
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -3156,6 +3156,12 @@ class Gallery(IOComponent):
         }
         return updated_config
 
+    def get_config(self):
+        return {
+            "value": self.value,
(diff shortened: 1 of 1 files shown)


abidlabs:
Looks great, thank you @Ian-GL.

Btw @pngwn @aliabid94 I noticed that the Gallery component no longer shows images in a 3x3 grid by default. I think that was a sensible default that help people understand the purpose of the Gallery component. Instead, it shows images in a full-width vertical grid, like this:

image

Consider changing that back?

pngwn:
Yeah, I need to take a look at the gallery, that is a bug not a feature.

PR #1619, opened 2022-06-22
Release 3.0.20

abidlabs:


FarukOzderim:
Is [version](https://github.com/gradio-app/gradio/blob/51c8c34486bfddca5948e46e498de44e21ab6496/setup.py#L17
) in setup.py important? @aliabid94 @abidlabs 

Can we delete it or should we add it to [wiki](https://github.com/gradio-app/gradio/wiki/How-to-deploy-Gradio-to-Pypi-using-GitHub-actions)

abidlabs:
So the GitHub action actually updates the version number in `setup.py` automatically before pushing to pypi

FarukOzderim:
But it does not push that to the github repo?

abidlabs:
Yeah agreed, that would def be better, will create an issue and look at it later

pngwn:
Rather than changing and committing `setup.py`, can we not just load in the version defined in `version.txt`?

FarukOzderim:
Agreed.

Either that or have both
- version.txt
- setup.py

Update documented in the wiki page

abidlabs:
Yes, that makes sense @pngwn. I'm not sure why we have a separate `version.txt` file. @aliabid94 do you know? Can we just get rid of it and use the `setup.py`'s version directly?

pngwn:
I guess it is easier to read in a simple txt file for various purposes. No opinion on how we do it though really, but having just the one will be simpler.

PR #1621, opened 2022-06-22
Mask image

pngwn:
to follow.


abidlabs:
Some thoughts here @pngwn:
* Overall, the data flow makes sense. It complicates the preprocessing function for the `Image` class, but I don't see any way around this. I've implemented the backend so now the demo works. @aliabid94 could you take a look as well?
* Right now, `sketch` only works if `source` is `upload`. It should be compatible with source being `webcam` as well, although this does not work for me right now. I don't think it makes any sense if source is `canvas` so we can catch and throw that error. WDYT?
* For the UI component itself, some suggested changes:
  * Can we reduce the stroke width? So that it is a similar size as the stroke width of the canvas?
  * Can we make the stroke white (at least by default)? That's what I typically see in inpainting demos: [example 1](https://www.google.com/search?q=inpainting+demo+python&oq=inpa&aqs=edge.0.69i59j69i57j0i433i512j0i131i433i512j0i433i512j0i131i433i512l4.600j0j1&sourceid=chrome&ie=UTF-8), [example 2](https://www.geeks3d.com/20180425/nvidia-deep-learning-based-image-inpainting-demo-is-impressive/)
  * There doesn't seem to be any way for you to clear the image once you have uploaded it. Pressing the X only clears the strokes
  * It would be useful if you could provide a default image using the `value` parameter. This parameter doesn't seem to do anything if `tool` is `sketch`

abidlabs:
LGTM!

abidlabs (approved):

Issue #1605, opened 2022-06-18
Gradio 3 does not allowing stacking multiple functions 


osanseviero:
### Describe the bug

As per https://discuss.huggingface.co/t/space-is-displaying-infinitely-loading-while-status-is-running/18603, Gradio <3 allowed something like

```py
model_iface = gr.Interface(
    fn=[your_function_1, your_function_2],
    inputs=gr.inputs.Textbox(),
    outputs=gr.outputs.Label(num_top_classes=10),
    examples=[["first example"],
              ["second example"]]
)
```
The logic above can be replicated in some sense with `gr.Parallel`, but this might confuse users that want to upgrade.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

-

### Logs

```shell
-
```

### System Info

```shell
3
```

### Severity

annoying

osanseviero:
Note: I think this is likely very low impact so it might be ok :) I was just reporting it

abidlabs:
Thanks @osanseviero you are completely right. This is actually one of the few things that we deprecated from Gradio 2 to 3 because it was rarely used and complicates the internal logic quite a bit. I'll add a deprecation notice.

AlvinKimata:
You can try using the `TabbedInterface` class provided by Gradio. It takes in a list of interfaces as input and renders the interfaces in multiple tabs. For example: 

```python
app = gr.TabbedInterface(interface_list=[image_interface, video_interface], 
                         tab_names = ["Image inference", "Video inference"])

if __name__ == '__main__':
    app.launch()
```

PR #1623, opened 2022-06-23
Deprecation error if multiple functions are passed into `fn` parameter in `Interface`

abidlabs:
Raises deprecation warning and rewrite gr.mix.Parallel to not use stack of functions.

Closes: #1605

### DIFF
--- a/gradio/interface.py
+++ b/gradio/interface.py
@@ -106,7 +106,7 @@ class Interface(Blocks):
 
     def __init__(
         self,
-        fn: Callable | List[Callable],
+        fn: Callable,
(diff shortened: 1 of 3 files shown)


FarukOzderim (approved):

FarukOzderim:
Good work, though some tests are throwing deprecation 🐱

abidlabs:
Thanks @FarukOzderim would you be able to take another look? Rewrote `gr.mix.Parallel` to not require the use of stacked functions

FarukOzderim (approved):
good work, LGMT!

Issue #1539, opened 2022-06-10
Upscale/downscale example images to be consistent


osanseviero:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
At https://huggingface.co/spaces/keras-io/conv_autoencoder, the example images look tiny, but if clicked they look ok. Should we upscale here so all apps feel consistent?

abidlabs:
Hmm interesting, it's probably because the original images are only 28x28. I think it's reasonable to upscale them -- wdyt @aliabid94?

osanseviero:
Yes, it's because of the original size, but I think it would be nice to be consistent regardless of specific example size

aliabid94:
Yep makes sense, will fix.

PR #1627, opened 2022-06-23
Fix example image thumbnail size

dawoodkhan82:
# Description
Issue #1539

Before:
Screen Shot 2022-06-23 at 6 10 36 PM

After:
Screen Shot 2022-06-23 at 6 10 19 PM

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

Closes: # (issue)
Issue #1539

# Checklist:


dawoodkhan82:
@aliabid94 does this change make sense? Since we only want to resize the example thumbnail image; and not actually modify the image itself.

ui/packages/app/src/components/Dataset/ExampleComponents/Image.svelte
```diff
@@ -4,4 +4,7 @@
 </script>
 
 <!-- svelte-ignore a11y-missing-attribute -->
-<img class="gr-sample-image" src={samples_dir + value} />
+<img
+	class="gr-sample-image object-contain h-20 w-20"
```
dawoodkhan82, line 8: @aliabid94 does this change make sense? Since we only want to resize the example thumbnail image; and not actually modify the image itself.

aliabid94: I think it makes sense, just test it with different image sizes. Also let's add these classes to the tokens.css file where all the other classes are applied to `gr-sample-image`, not in this file

demo/image_mod/run.py
```diff
@@ -1,12 +1,16 @@
...
 
 
 def image_mod(image):
     return image.rotate(45)
 
 
 demo = gr.Interface(image_mod, gr.inputs.Image(type="pil"), "image", 
-    flagging_options=["blurry", "incorrect", "other"])
+    flagging_options=["blurry", "incorrect", "other"], examples=[
+        os.path.join(os.path.dirname(__file__), "images/cheetah1.jpg"),
```
aliabid94, line 11: is `os.path.join(os.path.dirname())`  necessary? cant you just do `images/cheetah1.jpg`.

demo/image_mod/run.py
```diff
@@ -1,12 +1,16 @@
...
+        os.path.join(os.path.dirname(__file__), "images/lion.jpg")
+        ], cache_examples=True)
```
aliabid94, line 13: pls remove cache_examples

aliabid94 (dismissed):

demo/image_mod/run.py
abidlabs: This allows the demo (run.py) file to be run from anywhere

freddyaboulton (approved):
@dawoodkhan82 Looks good to me! I was able to verify the thumbnails are all the same size.

However, I think we should add an image to `image_mode/images` that's deliberately not the same size as the others so we can catch any regressions in the future. Right now the cheetah and lion are the same size.

dawoodkhan82:
> @dawoodkhan82 Looks good to me! I was able to verify the thumbnails are all the same size.
> 
> However, I think we should add an image to `image_mode/images` that's deliberately not the same size as the others so we can catch any regressions in the future. Right now the cheetah and lion are the same size.

Agreed, added one

Issue #1629, opened 2022-06-24
Deploy to pypi GitHub action does not push changed files to `main`


abidlabs:
See discussion:

> But it does not push that to the github repo? It is not a big problem but creates inconsistency.

_Originally posted by @FarukOzderim in https://github.com/gradio-app/gradio/issues/1619#issuecomment-1165254000_

PR #1631, opened 2022-06-25
Read version.txt

abidlabs:
In `setup.py`, this reads the version from `version.txt` to have a single source of truth.

Closes: #1629


pngwn (approved):
lgtm!

Issue #1604, opened 2022-06-18
Loading interface from Space does not work (sometimes)


osanseviero:
### Describe the bug

Trying

```python
iface = gr.Interface.load("spaces/hysts/TADNE")
```
Fails. Also reported few weeks ago in the forum https://discuss.huggingface.co/t/gr-interface-load-not-working/18500/2.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```py
import gradio as gr
iface = gr.Interface.load("spaces/hysts/TADNE")
```

### Logs

```shell
The error I got is different from what the person got in the forum

Fetching interface from: https://huggingface.co/spaces/hysts/TADNE
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-8-456895495868> in <module>()
----> 1 iface = gr.Interface.load("spaces/hysts/TADNE")

5 frames
/usr/local/lib/python3.7/dist-packages/gradio/interface.py in load(cls, name, src, api_key, alias, **kwargs)
     88         (gradio.Interface): a Gradio Interface object for the given model
     89         """
---> 90         return super().load(name=name, src=src, api_key=api_key, alias=alias, **kwargs)
     91 
     92     @classmethod

/usr/local/lib/python3.7/dist-packages/gradio/blocks.py in load(self_or_cls, fn, inputs, outputs, name, src, api_key, alias, **kwargs)
    660             if outputs is not None:
    661                 kwargs["outputs"] = outputs
--> 662             return external.load_blocks_from_repo(name, src, api_key, alias, **kwargs)
    663         else:
    664             self_or_cls.set_event_trigger(

/usr/local/lib/python3.7/dist-packages/gradio/external.py in load_blocks_from_repo(name, src, api_key, alias, **kwargs)
     38         factory_methods.keys()
     39     )
---> 40     blocks: gradio.Blocks = factory_methods[src](name, api_key, alias, **kwargs)
     41     return blocks
     42 

/usr/local/lib/python3.7/dist-packages/gradio/external.py in get_spaces(model_name, api_key, alias, **kwargs)
    313         return get_spaces_interface(model_name, config, alias, **kwargs)
    314     else:  # Create a Blocks for Gradio 3.x Spaces
--> 315         return get_spaces_blocks(model_name, config)
    316 
    317 

/usr/local/lib/python3.7/dist-packages/gradio/external.py in get_spaces_blocks(model_name, config)
    358         else:
[...]

PR #1636, opened 2022-06-26
fixed load from spaces + added test

abidlabs:
Fixes: #1604, which is was happening because we changed the config for a Blocks to include an optional `documentation` parameter for events. 

Also adds a test to ensure that the loading `from_config` works to prevent this kind of error in the future.


abidlabs:

Issue #1156, opened 2022-05-03
[Blocks] Expose Examples as Block Component


radames:
At the moment, examples are part of the `Interface` Class, as well as the `cache_examples` feature. With `Blocks` users can compose custom layouts, positioning inputs and outputs arbitrarily using `Columns`, `Rows`, etc.

This request is to expose examples as a `Block` component, thus one can position and layout it as needed.

Here is one snippet idea:

```python
with blocks:
....
    with gr.Row() 
        gr.Examples(examples=examples, cache_examples=False)
....
```

or maybe

```python
with blocks:
....
    with gr.Row() 
        examples = gr.Examples(examples=examples, cache_examples=False)

    button.click(predict, inputs, outputs, examples)
....
```

PR #1639, opened 2022-06-26
Exposing examples as a component for Blocks

abidlabs:
We now have `gr.Examples()` as a component that can be used within Blocks. See example: `demo\blocks_inputs\run.py`

Usage:

* Case 1: Only `inputs` is provided, in which case clicking on the examples will populate the inputs components

```py
with blocks:
....
    with gr.Row() 
        textbox = gr.Textbox()
        examples = gr.Examples(examples=examples, inputs=textbox)
        ....

    button.click(predict, textbox, label)
....
```

* Case 2: Both `inputs` and `outputs` as well as the prediction `fn` is provided (required for `cache_examples=True`), in which case, if the examples are cached, clicking on the examples will populate the inputs components and the output components.

```py
with blocks:
....
    with gr.Row() 
        textbox = gr.Textbox()
        examples = gr.Examples(examples=examples, inputs=textbox, outputs=label, fn=predict, cache_examples=True)
        ....

    button.click(predict, textbox, label)
....
```

This approach is more wordy than the one proposed by @radames in #1156 but I think more logical since it ties the examples clearly to the input (and potentially the output components). Open to feedback here.

Original API proposed by @radames:

```py
with blocks:
....
    with gr.Row() 
        examples = gr.Examples(examples=examples, cache_examples=False)

    button.click(predict, inputs, outputs, examples)
....
```

* Problem: in this approach, what would happen if multiple event triggers are passed the same `examples`?

Closes: #1156

**Note:** This PR also does a lot of other cleanups in the `Interface` class. A lot of the bloat is removed because we now only accept single functions in the `fn` parameter of `Interface`. I've tested a bunch of demos and they work fine, but would appreciate a careful review to make sure I didn't miss anything!


radames:
hi @abidlabs ,

Your proposal approach looks good to me. It's also enables users to create multiple examples for different predict functions on the same screen, right?   For example, this [Space](https://huggingface.co/spaces/CVPR/Text2Human/tree/main) has two sections, and would benefit of separated examples for each step.  

```python 
with blocks:
....
    with gr.Row() 
        textbox = gr.Textbox()
        outbox =  gr.Textbox()
        examples = gr.Examples(examples=examples, inputs=textbox, outputs=outbox, fn=predict, cache_examples=True)
        ....
    with gr.Row() 
        data = gr.DataFrame()
        examples2 = gr.Examples(examples=examples2, inputs=outbox, outputs=data, fn=predict2, cache_examples=True)
        ....

    button.click(predict, textbox, label)
.... 
````

**Comment by abidlabs** [comment]
Yes, that is now possible @radames. See example here: `demo\blocks_inputs\run.py`

File: gradio/examples.py [hunk 1]
```diff
@@ -0,0 +1,191 @@
...
+        fn: (Callable | None): optionally, provide the function to run to generate the outputs corresponding to the examples. Required if `cache` is True.
+        cache_examples (bool): if True, caches examples for fast runtime. If True, then `fn` and `outputs` need to be provided
+        examples_per_page (int): how many examples to show per page (this parameter currently has no effect)
+        """
+        if not isinstance(inputs, list):
+            inputs = [inputs]
+        if not isinstance(outputs, list):
+            outputs = [outputs]
+
+        if examples is None or (
```
freddyaboulton, line 50: `examples` cannot be None according to the type hint and when I set it to None the front-end hangs to load up the app in the blocks_input demo:

image

If `None` is an allowed input we should figure out why the front-end can't display it. But if it's not (and I don't think it should be allowed btw) we should raise an appropriate error to the user?

freddyaboulton: This is the repro:

```python
import gradio as gr
import os


def combine(a, b):
    return a + " " + b


def mirror(x):
    return x


with gr.Blocks() as demo:
    txt_3 = gr.Textbox("", label="Output")
    btn = gr.Button("Submit")

    with gr.Row():
        im = gr.Image()
        im_2 = gr.Image()

    btn = gr.Button("Mirror Image")

    gr.Markdown("## Text Examples")
    gr.Examples([["hi", "Adam"], ["hello", "Eve"]], [txt, txt_2], txt_3, combine, cache_examples=True)
    gr.Examples(
        examples=None,
        inputs=im,
        cache_examples=False)

if __name__ == "__main__":
    demo.launch()
```

freddyaboulton (commented):
@abidlabs This is awesome! Taking a closer look now but one thing that stood out to me is that we're giving lots of power to the user and that increases the chance of misusing examples. I wonder if we should invest some time in making sure incorrect combinations of parameters are caught and informative error messages are displayed.

For example, 

- when you set `cache_examples` to True but don't provide a `fn` you get a `TypeError: 'NoneType' object is not callable` error. 
```python
    gr.Examples(
        examples=[os.path.join(os.path.dirname(__file__), "lion.jpg")],
        inputs=im,
        outputs=im_2,
        cache_examples=True)
```

- when you provide inputs, fn but not outputs and cache_examples is True:
```python
    gr.Examples(
        examples=[os.path.join(os.path.dirname(__file__), "lion.jpg")],
        inputs=im,
        fn=mirror,
        cache_examples=True)
```
- If the size of the examples list doesn't match the api of the fn you get a `IndexError: list index out of range`
```python
    gr.Examples([["hi"], ["hello", "Eve"]], [txt, txt_2], txt_3, combine, cache_examples=True)
```
- There's a weird bug when `examples` is None that I commented below. 

Some of these may be rare cases and not worth catching in the library but some may be worth guiding the user on. Curious what your thoughts on the matter are.

gradio/examples.py
```diff
@@ -0,0 +1,191 @@
...
+        self.cache_examples = cache_examples
+        self.examples_per_page = examples_per_page
+        self.cached_folder = os.path.join(CACHED_FOLDER, str(dataset._id))
+        self.cached_file = os.path.join(self.cached_folder, "log.csv")
+
+        if cache_examples:
+            self.cache_interface_examples()
+
+        def load_example(example_id):
+            processed_examples = [
```
freddyaboulton, line 117: The examples don't change during the lifetime of the app. I think we can compute `processed_examples` once and avoid a bunch of repetitive work?

gradio/mix.py
```diff
@@ -60,49 +54,43 @@ class Series(gradio.Interface):
...
+                        for i, input_component in enumerate(interface.input_components)
                     ]
 
                 # run all of predictions sequentially
-                predictions = []
-                for predict_fn in io.predict:
-                    prediction = predict_fn(*data)
-                    predictions.append(prediction)
-                data = predictions
+                data = interface.run_prediction(data)
```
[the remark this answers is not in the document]
abidlabs: Hmm yeah I was confused by this, let me play around with it some more and see if we can fix it.

abidlabs:
Series should be fixed now, thanks for the suggestion @freddyaboulton!

gradio/examples.py
```diff
@@ -0,0 +1,200 @@
...
+                        None,
+                    )
+            examples = exampleset
+        else:
+            raise ValueError(
+                "The parameter `examples` must either be a directory or a nested "
+                "list, where each sublist represents a set of inputs."
+            )
+
+        if cache_examples and (fn is None or outputs is None):
```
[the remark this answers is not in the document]
abidlabs: Ah good catch thanks!

abidlabs:
Thanks for the thorough feedback @freddyaboulton!

PR #1645, opened 2022-06-27
fix interactivity

### LINKED ISSUE #1569: Dropdown is disabled when created with blocks
### Describe the bug

Hi,
Tried creating gradio app with blocks, Dropdown was disabled after launch.

- [x] I have searched the existing issues

### Reproduction

```
import gradio as gr

demo = gr.Blocks()

with demo:
    with gr.Tabs():
        with gr.TabItem("Record Video"):
        
            with gr.Row():
                inp1=gr.inputs.Video(source="webcam",optional=False,label='Capture Video')
                inp2=gr.inputs.Video(source="webcam",optional=False,label='Capture Video')
            with gr.Row():
                subject_id = gr.inputs.Dropdown(["cat", "dog", "bird"])
    
        
demo.launch()
```

### Screenshot

image
image

### Logs

```shell
NA
```

### System Info

```shell
Browser: Chrome
Gradio Version: 3.0.17
```

### Severity

blocking all usage of gradio

pngwn:
Closes #1569.

Previously there were some issues with interactivity auto-detection. It was actually working as intended for the most part but it was a little confusing. The important factor here was that components needed to have no default_value in order to appear as dynamic. As soon as you gave them a default_value, the logic determined that if they have no inputs or outputs they are static. This makes sense in general but is slightly confusing when developing the application before all events have been hooked up.

This PR is more of a proposal. It essentially means that components which act only as an output will rename static, everything else is dynamic regardless of whether there is a default value. It is less magical but much easier to understand. _Most_ of the time the component will be dynamic, users will always have the `interactive=False` kwarg for those times when they need to go against the very simple auto behaviour.

### DIFF
--- a/ui/packages/app/src/Blocks.svelte
+++ b/ui/packages/app/src/Blocks.svelte
@@ -118,8 +118,8 @@
 	const dynamic_ids = components.reduce((acc, { id, props }) => {
 		const is_input = is_dep(id, "inputs", dependencies);
 		const is_output = is_dep(id, "outputs", dependencies);
-
-		if (!is_input && !is_output && !props.value) acc.add(id); // default dynamic
+		console.log(props.value, is_input, is_output);
+		if (!is_input && !is_output) acc.add(id); // default dynamic
 		if (is_input) acc.add(id);
 
 		return acc;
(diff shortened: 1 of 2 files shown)


aliabid94:
Hmm I see the point of reducing "magic", but I also think placing a static image as `gr.Image("cat.jpg")` is quite intuitive. Don't feel strongly either way though, so feel free to merge. Also, feel free to chime in @freddyaboulton if you think one way is more intuituve.

pngwn:
Yeah, I agree. Personally I like the autodetection logic. I'll make an alternative PR that tackles any bugs with the current implementation and we can decide which to keep.

pngwn:
@aliabid94 Alternative retaining the magic is here: #1685

pngwn:
Closing in favour of #1685

PR #1661, opened 2022-06-29
fix overflow

### LINKED ISSUE #1618: dataframe display exceed the window size
## Describe the problem

Hi, I want to display multiply columns at each row and I found that the Dataframe output will exceed the window size. And I have searched relative issues and I couldn't find any solutions to fix the dataframe output width.

## Additional context

Code:

```
import pandas as pd
import gradio as gr

def example_fn(text):
    d = {'c1': ["This is col 1"], 'c2': ["This is col 2"], 'c3': ["This is col 3"], 
         'c4': ["This is col 4"], 'c5': ["This is col 5"], 
         'c6': ["This is col 6"], 'c7': ["This is col 7"]}
    df = pd.DataFrame(data=d)
    return df

demo = gr.Interface(
    fn = example_fn,
    inputs = [gr.Textbox(label="Input", lines=3, )
             ],
    outputs = [gr.DataFrame()],
    title="Test",
    allow_flagging=False,
)

demo.launch()
```

Image:
image

Version:
python: 3.7.10
gradio: 3.0.11

pngwn:
Just an overflow issue.

Closes #1618.

### DIFF
--- a/ui/packages/app/src/components/Column/Column.svelte
+++ b/ui/packages/app/src/components/Column/Column.svelte
@@ -13,7 +13,7 @@
 	class:bg-gray-50={variant === "panel"}
 	class:p-2={variant === "panel"}
 	class:rounded-lg={variant === "panel"}
-	class="flex flex-col gr-gap gr-form-gap relative col w-full {create_classes(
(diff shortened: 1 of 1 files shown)


freddyaboulton (approved):
Thanks @pngwn ! 

Confirmed it fixes the bug

image

PR #1668, opened 2022-06-29
Undo use of orjson for Variable default value

aliabid94:
The variable default value will instead be deepcopied, because in case it is an Object, we don't want different sessions to refer to the same object.


gradio/routes.py
```diff
@@ -40,7 +40,7 @@ class ORJSONResponse(JSONResponse):
     media_type = "application/json"
 
     def render(self, content: Any) -> bytes:
-        return orjson.dumps(content, option=orjson.OPT_SERIALIZE_NUMPY)
+        return orjson.dumps(content)
```
freddyaboulton, line 43: Don't we still need `OPT_SERIALIZE_NUMPY` in case the variable value is numpy? 

https://github.com/gradio-app/gradio/pull/1606/

aliabid94: You're right, just fixed.

gradio/components.py
```diff
@@ -2723,25 +2722,13 @@ def __init__(
...
+        self.value = deepcopy(value)
         self.stateful = True
         IOComponent.__init__(self, **kwargs)
 
     def get_config(self):
         return {
-            "value": orjson.dumps(self.value, option=orjson.OPT_SERIALIZE_NUMPY).decode(
-                "utf-8"
-            ),
+            "value": self.value,
```
freddyaboulton, line 2731: We may need to revert this change as well? This breaks this fix made last week https://github.com/gradio-app/gradio/issues/1408

gradio/components.py
```diff
@@ -2723,25 +2722,13 @@ def __init__(
...
-            orjson.dumps(self.value, option=orjson.OPT_SERIALIZE_NUMPY)
-        except TypeError:
-            raise ValueError(
-                f"Cannot set initial value because type: {type(value)} is "
-                "not JSON serializable. Instead, you can set the initial "
-                "value to be None, and in your Python function, change it "
-                "to be the your desired value when the function is "
-                "run initially."
-            )
+        self.value = deepcopy(value)
```
freddyaboulton, line 2725: Might be better to dump before load

```python
orjson.loads(orjson.dumps(value))
```

image

freddyaboulton (changes requested):

gradio/components.py
freddyaboulton: That way we don't break the numpy serialization merged before.

aliabid94: I got rid of get_config so the value of Variable is never passed to the frontend, so I don't think this is necessary correct?

freddyaboulton (approved):
Thanks @aliabid94 ! Confirmed the fix works on your test code. Thanks for putting up the fix and addressing comments so quickly.

FarukOzderim:
Let me just ping @abidlabs in case, since I remember he added the orjson dumping.

PR #1669, opened 2022-06-29
Use requirements.txt to specify dependencies. Delete egg-info

### LINKED ISSUE #1648: List python runtime dependencies in requirements.txt as opposed to setup.py

As a gradio developer, I think it's frustrating that the `gradio-egg.info` folder is tracked in version control. Whenever I build gradio locally, there are changes in the `gradio-egg.info` directory that are not related to my PR.

I see that `gradio-egg.info` is only included in the repo so that we can
1. cache the env in circleci 
2. install all dependencies in the `install_test_requirements.sh` script

I'd like to specify all dependencies in a `requirements.txt` file. That will allow us to still cache the circle-ci env. I don't think we need to install the gradio dependencies again in `install_test_requirements.sh` since that's done in `install_gradio.sh` but we can still install from `requirements.txt` if necessary. 

This will also allow us to remove `gradio-egg.info` from version control. 

**Additional context**

freddyaboulton:
# Description

Fixes #1648

# Checklist:


gradio/__init__.py
```diff
@@ -67,5 +67,5 @@
     Webcam,
 )
 
-current_pkg_version = pkg_resources.require("gradio")[0].version
+current_pkg_version = version("gradio")
```
freddyaboulton, line 70: Use pkg_resources is discouraged by setuptools

image

.circleci/config.yml
```diff
@@ -23,13 +23,14 @@ jobs:
           command: |
             python3 -m venv venv
             . venv/bin/activate
+            bash scripts/install_gradio.sh
```
freddyaboulton, line 26: So that gradio is installed in circleci worker. We were doing this implicitly by having the egg-info in version control.

aliabid94 (approved):
Beautiful!

FarukOzderim (approved):
This looks great! It was definitely a headache, Didn't know that we could remove gradio.egg, thought it was necessary as a pip package.

Where do we keep the egg info now? Was it not necessary at all?

dawoodkhan82 (approved):

abidlabs:
Egg info is not necessary as eggs are pretty much an abandoned format: https://stackoverflow.com/a/68897551/5209347

PR #1680, opened 2022-07-01
Fix versioning in requirements

aliabid94:
httpx was recently added as a dependency, and now we cannot build gradio from source because of a version conflict of the library h11, which is a dependency of both httpx (requires `h11>=0.11,<0.13`) and fastapi (requires `h11>=0.8`). Specified version `h11=0.12.0` manually to satisfy both but I'm assuming there's probably a better way to manage our dependencies.


setup.py
```diff
@@ -47,6 +47,7 @@
         "uvicorn",
         "Jinja2",
         "fsspec",
+        "h11==0.12.0",
```
freddyaboulton, line 50: @aliabid94 I can't repro the original problem on main. I did the following:

```bash
conda create -n test python=3.9 pip -y
conda activate test
python setup.py sdist
pip install dist/gradio-3.0.22.tar.gz
```

And the install succeeded. How can I reproduce?

aliabid94: Here it is in a google colab. 
https://colab.research.google.com/drive/1sRcgYQoJ5IOaHRO-QRm2-26CWIcg1qLb?usp=sharing
The error might be flaky depending on the order that the dependencies are installed.

freddyaboulton: Thank you for the link @aliabid94 !

freddyaboulton (approved):
@aliabid94 This looks good to me! But we should make the requirement as flexible as possible. I think `h11<0.13,>=0.11` will work right?

FarukOzderim (approved):
Did not know that this kind of versioning conflict could appear with `reqs.txt`. This is a easy fix short run, but will not fix for the long run.

Shall we migrate to poetry for version controlling when we find the time for it?

FarukOzderim:
Also we could use `pip-tools` like we do in test requirements. But it would create very specific versions for each package, and people could use gradio just in virtual environments, most of their packages would get reinstalled. So that's why we use pip-tools just in test-requirements.

Issue #1446, opened 2022-06-01
File component: list multiple files and allow for download


dawoodkhan82:
As of now, when `file_count` is set to multiple; the file component does not list all uploaded files and does not allow each to be downloaded.

- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  

**Describe the solution you'd like**  

**Additional context**

PR #1681, opened 2022-07-01
File component: list multiple files and allow for download #1446

dawoodkhan82:
# Description
List each file individually and allow for download in a scroll view.

Before:
Screen Shot 2022-07-08 at 9 12 40 AM

After:
Screen Shot 2022-07-11 at 2 38 44 PM

Closes: #1446

# Checklist:


abidlabs:
Not a lot of documentation on the issue or PR @dawoodkhan82?

dawoodkhan82:
@abidlabs I think the multiple file component could use a legit re-design (kinda basic right now), but this pr fixes the issue where the files weren't listed nor able to download.

demo/zip_two_files/run.py
```diff
@@ -13,7 +13,7 @@ def zip_two_files(file1, file2):
 
 demo = gr.Interface(
     zip_two_files,
-    ["file", "file"],
+    [gr.File(file_count="multiple"), gr.File(file_count="multiple")],
```
abidlabs, line 16: Why did we change both of the input components here to accept multiple files each? We should either have a single component that can accept multiple files (and change the name of this demo) or just change it back to two file components that each accept a single file

ui/pnpm-lock.yaml
```diff
@@ -41,7 +41,7 @@ importers:
       '@tailwindcss/forms': 0.5.0_tailwindcss@3.0.23
       '@testing-library/dom': 8.11.3
       '@testing-library/svelte': 3.1.0_svelte@3.47.0
-      '@testing-library/user-event': 13.5.0_gzufz4q333be4gqfrvipwvqt6a
+      '@testing-library/user-event': 13.5.0_@testing-library+dom@8.11.3
```
abidlabs, line 44: Why do we have all these changes in `ui/pnpm-lock.yaml`?

demo/zip_two_files/run.py
dawoodkhan82: Pushed this by accident. Changed this for testing purposes

dawoodkhan82:
> Added some thoughts above. Also, I notice that in the original component, the columns are distributed evenly:
> 
> image
> 
> In the redesigned version, the columns are all bunched towards the left:
> 
> image
> 
> Can we fix that so that the columns are distributed evenly again?
> 
> Would be great if you can give this a review as well @pngwn since I can't really comment on the frontend changes

fixed. see new screenshot

ui/pnpm-lock.yaml
dawoodkhan82: fixed

pngwn (approved):
lgtm!

Zengyf-CVer:
@dawoodkhan82 
Can you provide the code for this example?

dawoodkhan82:
> @dawoodkhan82 Can you provide the code for this example?

To use the multiple file input and out components you can specify `gradio.File(file_count="multiple")`

Zengyf-CVer:
@dawoodkhan82 
Thank you very much!

PR #1684, opened 2022-07-01
add markdown and html support to dataframe

### LINKED ISSUE #1364: Dataframe allow Markdown or HTML content

Simple example here: https://huggingface.co/spaces/Gradio-Blocks/Leaderboard. Currently it's not possible to create hyperlinks inside a Dataframe table. I can see other use cases beyond `<a>` elements, maybe images or simple video previews?  I'm thinking that would be handy to preview rich content Dataframes outputs. I understand the performance issues with large tables, but I guess it needs lazy loading anyway.

pngwn:
Allows dataframe cells to contain markdown or html.

- markdown will be processed in the same way that markdown in the `Markdown` component is (using `MarkdownIt`)
- Supports both default values and generated values (via predict function)
- If the table is interactive, the 'editable' value is the resulting HTML. This feels like a niche usecase but it is supported. We could optionally disable editing for html + markdown fields. Allowing the original source markdown to be the editable value is technically possible but introduces a degree of complexity I did not want to introduce in this PR, it would require significant changes to the implementation.

Closes #1364.

cc @radames

### DIFF
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -2356,7 +2356,7 @@ class Dataframe(Changeable, IOComponent):
     Demos: filter_records, matrix_transpose, tax_calculator
     """
 
-    md = None
+    markdown_parser = None
 
     def __init__(
         self,
@@ -2572,13 +2572,13 @@ class Dataframe(Changeable, IOComponent):
         if "markdown" not in datatype:
             return data
 
-        if cls.md is None:
-            cls.md = MarkdownIt()
+        if cls.markdown_parser is None:
+            cls.markdown_parser = MarkdownIt()
 
         for i in range(len(data)):
             for j in range(len(data[i])):
                 if datatype[j] == "markdown":
-                    data[i][j] = Dataframe.md.render(data[i][j])
+                    data[i][j] = Dataframe.markdown_parser.render(data[i][j])
 
         return data
 
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -2356,6 +2356,8 @@ class Dataframe(Changeable, IOComponent):
     Demos: filter_records, matrix_transpose, tax_calculator
     """
 
+    md = None
+
     def __init__(
         self,
         value: Optional[List[List[Any]]] = None,
@@ -2412,6 +2414,8 @@ class Dataframe(Changeable, IOComponent):
             "number": 0,
             "bool": False,
             "date": "01/01/1970",
+            "markdown": "",
+            "html": "",
         }
         column_dtypes = (
             [datatype] * self.col_count[0] if isinstance(datatype, str) else datatype
@@ -2419,7 +2423,10 @@ class Dataframe(Changeable, IOComponent):
         self.test_input = [
             [values[c] for c in column_dtypes] for _ in range(self.row_count[0])
         ]
+
         self.value = value if value is not None else self.test_input
+        self.value = self.__process_markdown(self.value, datatype)
+
         self.max_rows = max_rows
         self.max_cols = max_cols
         self.overflow_row_behaviour = overflow_row_behaviour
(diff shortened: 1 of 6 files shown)


pngwn:
Just to add, this can be tested via the `demo/blocks_outputs` demo. The fn is at the top, the dataframes are at the bottom (for predict fn + default value testing).

pngwn:
Another thing, I didn't update it here (although I can) but we don't need a new instance of `MarkdownIt` for each class instance of `Markdown`. That can be a static class variable.

gradio/components.py
```diff
@@ -2550,10 +2565,28 @@ def __validate_headers(headers: List[str] | None, col_count: int):
                 )
             )
 
+    @staticmethod
+    def __process_markdown(data: List[List[Any]], datatype: str | List[str]):
+        if (
```
freddyaboulton, line 2570: @pngwn I think I figured out the bug.

In the demo `datatype` is `['markdown', 'markdown', 'html']`. That's always not equal to the string "markdown" so the `or` evaluates to True.

A more verbose but easier to understand implementation would be

```python
        if (
            isinstance(datatype, list) and "markdown" not in datatype
        ):
            return data
        if isinstance(datatype, str) and datatype != "markdown":
            return data
```

Also `isinstance` is better that `type(....) is ...` in my opinion

freddyaboulton: It can render the markdown now

image

gradio/components.py
```diff
@@ -2521,15 +2528,23 @@ def postprocess(self, y):
             return y
         if isinstance(y, str):
             y = pd.read_csv(str)
```
freddyaboulton, line 2530: Unrelated to your PR but this is a bug right? It should be `y = pd.read_csv(y)` right?

gradio/components.py
```diff
@@ -2550,10 +2565,28 @@ def __validate_headers(headers: List[str] | None, col_count: int):
                 )
             )
 
+    @staticmethod
```
freddyaboulton, line 2568: nitpick but I think a `classmethod` might be more appropriate here since we modify `md` in this method.

freddyaboulton (approved):
@pngwn This looks good to me once we fix the `if` statement in `process_markdown` . Thank you!

gradio/components.py
pngwn: It has to be. I'll fix in this PR.

gradio/components.py
pngwn: Happy to change, google guides me.

gradio/components.py
freddyaboulton: Yes! Changes the api to this

```python
    @classmethod
    def __process_markdown(cls, data: List[List[Any]], datatype: str | List[str]):
```

But the usage in `preprocess` will be the same so no need to change anything there.

pngwn: That's nice, cleaner way to access the class. ty!

pngwn:
As i discussed with @freddyaboulton in slack, I've switched to normalising the `datatype` upfront, so it is always a list of strings matching the column count. This simplifies the `process_markdown` logic significantly.

@freddyaboulton Feel free to take another glance + check that the previous example you tried works correctly.

I also fixed the bug you caught.

gradio/components.py
```diff
@@ -2356,6 +2356,8 @@ class Dataframe(Changeable, IOComponent):
     Demos: filter_records, matrix_transpose, tax_calculator
     """
 
+    md = None
```
hibestil, line 2359: IMO, It'd be better for understandability if we rename this variable to `markdown` or `has_markdown`. WDYT @pngwn ?

pngwn: Good point, this is actually the markdown parser instance. I'll rename it accordingly.

PR #1685, opened 2022-07-01
detect all types of null default value

### LINKED ISSUE #1569: Dropdown is disabled when created with blocks

pngwn:
ALternative to #1645.

- This PR checks for all types of 'null' default value (i.e. the value we get when no default value is provided), fixing the interactive/ non-interactive autodetection when no default value is given.
- I had to modify the preprocessing for Radio (which Dropdown inherits) as it was selecting the first value for the user even when none was given. I can kind of understand the reasoning but if we do this we have no way of knowing whether or a not a default value is being passed or not. I don't think we should be selecting a value for the user if they have not provided a valid one.

Closes #1569.

### DIFF
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -1233,9 +1233,7 @@ class Radio(Changeable, IOComponent):
         Returns:
         (str): string of choice
         """
-        return (
-            y if y is not None else self.choices[0] if len(self.choices) > 0 else None
-        )
+        return y if y is not None else None
 
     def deserialize(self, x):
         """
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -1233,7 +1233,7 @@ class Radio(Changeable, IOComponent):
         Returns:
         (str): string of choice
         """
-        return y if y is not None else None
+        return y
 
     def deserialize(self, x):
         """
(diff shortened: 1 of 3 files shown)


gradio/components.py
```diff
@@ -1233,9 +1233,7 @@ def postprocess(self, y):
         Returns:
         (str): string of choice
         """
-        return (
-            y if y is not None else self.choices[0] if len(self.choices) > 0 else None
-        )
+        return y if y is not None else None
```
aliabid94, line 1236: why not just `return y`

aliabid94:
makes sense to me

aliabid94 (approved):

gradio/components.py
pngwn: why indeed, will fix.

PR #1686, opened 2022-07-01
1220 mirror webcam

### LINKED ISSUE #1220: Mirroring for webcam inputs
- Mirror webcam's by default with necessary preprocessing.
- kwarg to disable mirror

pngwn:
Closes #1220 and completes the steps noted in that issue.

On pause for a moment because I've just realised I need to build custom controls for the video in order to implement this fully for `Video`. It is mostly working.

The python code can be reviewed but marking as a draft so it doesn't get merged.

I've done the best I can with `PIL` and `FFMPEG`, knowing nothing about them. If there are better alternative implementations then I'm all ears.


gradio/components.py
```diff
@@ -1773,11 +1783,24 @@ def preprocess(self, x: Dict[str, str] | None) -> str | None:
...
         file_name = file.name
         uploaded_format = file_name.split(".")[-1].lower()
+
         if self.format is not None and uploaded_format != self.format:
             output_file_name = file_name[0 : file_name.rindex(".") + 1] + self.format
             ff = FFmpeg(inputs={file_name: None}, outputs={output_file_name: None})
             ff.run()
             return output_file_name
+        elif self.source == "webcam" and self.mirror_webcam is True:
+            output_file_name = (
```
freddyaboulton, line 1793: I think path manipulation is better with pathlib

```python
import pathlib
path = pathlib.Path(file_name)
output_file_name = str(path.with_stem(f"{path.stem}_flip"))
```

freddyaboulton (commented):
@pngwn Python looks good to me except that maybe we can make the path manipulation easier by using pathlib

Issue #1653, opened 2022-06-28
assert_configs_are_equivalent_besides_ids does not check documentation field


freddyaboulton:
### Describe the bug

`assert_configs_are_equivalent_besides_ids` does not check all fields in the config, e.g. `documentation`, `api_name`

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Repro. I would expect the assertion to fail in this case since `documentation` is different.

```python
import copy
from gradio.utils import assert_configs_are_equivalent_besides_ids

config1 = {'version': '3.0.20\n', 'mode': 'blocks', 'dev_mode': True,
           'components': [{'id': 1, 'type': 'textbox',
            'props': {'lines': 1, 'max_lines': 20, 'placeholder': 'What is your name?', 'value': '',
                      'show_label': True, 'name': 'textbox', 'visible': True, 'style': {}}},
                      {'id': 2, 'type': 'textbox', 'props': {'lines': 1, 'max_lines': 20, 'value': '', 'show_label': True, 'name': 'textbox', 'visible': True, 'style': {}}}, {'id': 3, 'type': 'image', 'props': {'image_mode': 'RGB', 'source': 'upload', 'tool': 'editor', 'streaming': False, 'show_label': True, 'name': 'image', 'visible': True, 'style': {'height': 54, 'width': 240}}}], 'theme': 'default', 'css': None, 'enable_queue': False, 'layout': {'id': 0, 'children': [{'id': 1}, {'id': 2}, {'id': 3}]},
           'dependencies': [
               {'targets': [1], 'trigger': 'submit', 'inputs': [1], 'outputs': [2],
                'backend_fn': True, 'js': None, 'status_tracker': None, 'queue': None, 'api_name': 'greet',
                 'scroll_to_output': False, 'show_progress': True, 'documentation': [['(str): text'], ['(str | None): text']]
               }
           ]}

config2 = copy.deepcopy(config1)
config2['dependencies'][0]['documentation'] = None

assert assert_configs_are_equivalent_besides_ids(config1, config2)
```

### Logs

```shell
N/a
```

### System Info

```shell
Latest main
```

### Severity

annoying

PR #1700, opened 2022-07-04
Fix `assert_configs_are_equivalent_besides_ids()`

abidlabs:
Fixes the `assert_configs_are_equivalent_besides_ids()` to check for other fields in the `dependencies`. Adds a test as well to catch the mistake @freddyaboulton pointed out.

Fixes: #1653


gradio/utils.py
```diff
@@ -225,13 +224,16 @@ def same_children_recursive(children1, chidren2, mapping):
     same_children_recursive(children1, children2, mapping)
 
     for d1, d2 in zip(config1["dependencies"], config2["dependencies"]):
-        for t1, t2 in zip(d1["targets"], d2["targets"]):
+        for t1, t2 in zip(d1.pop("targets"), d2.pop("targets")):
             assert mapping[t1] == t2, "{} does not match {}".format(d1, d2)
```
FarukOzderim, line 228: Shall we use f"" strings over here and below?

gradio/utils.py
```diff
@@ -225,13 +224,16 @@ def same_children_recursive(children1, chidren2, mapping):
...
-        assert d1["trigger"] == d2["trigger"], "{} does not match {}".format(d1, d2)
-        for i1, i2 in zip(d1["inputs"], d2["inputs"]):
+        for i1, i2 in zip(d1.pop("inputs"), d2.pop("inputs")):
             assert mapping[i1] == i2, "{} does not match {}".format(d1, d2)
-        for o1, o2 in zip(d1["outputs"], d2["outputs"]):
+        for o1, o2 in zip(d1.pop("outputs"), d2.pop("outputs")):
             assert mapping[o1] == o2, "{} does not match {}".format(d1, d2)
+        # status tracker is ignored
```
FarukOzderim, line 233: Why do we ignore status tracker?

test/test_blocks.py
```diff
@@ -86,7 +87,11 @@ def fake_func():
 
         config = demo.get_config_file()
         config.pop("version")  # remove version key
```
FarukOzderim, line 89: hmm, instead of popping `version` in individual tests, wouldn't it be better to do it in the `assert_configs_are_equivalent_besides_ids`?

FarukOzderim (commented):
LGTM except a few points, good job!

test/test_blocks.py
freddyaboulton: Yea I think it will be better if the logic for ignoring fields is in `assert_configs_are_equivalent..` as opposed to the calling code. Maybe we can add a `ignore_fields: List[str]` parameter to the function that gives uses the ability to specify what they want to ignore. I can see use cases where you would want to ignore the version key and others where you wouldn't. 

I don't think this is necessary to merge this PR though. If anything I can file an issue to implement in the future.

gradio/utils.py
```diff
@@ -38,7 +38,7 @@
 
 def version_check():
     try:
-        current_pkg_version = pkg_resources.require("gradio")[0].version
```
freddyaboulton, line 41: We need to back these changes out due to importlib not being available on 3.7 right?

gradio/utils.py
```diff
@@ -225,13 +224,16 @@ def same_children_recursive(children1, chidren2, mapping):
     same_children_recursive(children1, children2, mapping)
 
     for d1, d2 in zip(config1["dependencies"], config2["dependencies"]):
-        for t1, t2 in zip(d1["targets"], d2["targets"]):
+        for t1, t2 in zip(d1.pop("targets"), d2.pop("targets")):
```
freddyaboulton, line 227: I think it would be helpful to add a comment about why popping is necessary here.

test/test_utils.py
```diff
@@ -130,20 +116,99 @@ def test_get_ip_without_internet(self, mock_get):
...
-            assert_configs_are_equivalent_besides_ids(XRAY_CONFIG, XRAY_CONFIG)
+            assert_configs_are_equivalent_besides_ids(
+                copy.deepcopy(XRAY_CONFIG), copy.deepcopy(XRAY_CONFIG)
+            )
         )
 
     def test_equivalent_configs(self):
         self.assertTrue(
-            assert_configs_are_equivalent_besides_ids(XRAY_CONFIG, XRAY_CONFIG_DIFF_IDS)
+            assert_configs_are_equivalent_besides_ids(
```
freddyaboulton, line 126: `test_equivalent_configs` and `test_same_configs` are identical right? If so we should delete one.

test/test_utils.py
```diff
@@ -130,20 +116,99 @@ def test_get_ip_without_internet(self, mock_get):
 class TestAssertConfigsEquivalent(unittest.TestCase):
     def test_same_configs(self):
         self.assertTrue(
-            assert_configs_are_equivalent_besides_ids(XRAY_CONFIG, XRAY_CONFIG)
+            assert_configs_are_equivalent_besides_ids(
+                copy.deepcopy(XRAY_CONFIG), copy.deepcopy(XRAY_CONFIG)
```
freddyaboulton, line 120: I don't think this copy is necessary since the util performs a copy.

freddyaboulton (approved):
@abidlabs Thank you for the fix! I left some minor nits that could be addressed prior to merging but are not blocking in my opinion.

gradio/utils.py
FarukOzderim: oh, that's correct 😄

gradio/utils.py
abidlabs: Sg will update

gradio/utils.py
abidlabs: The reason we ignore the status tracker is because it is going to have different IDs between different blocks, just like input and output components. In the case of inputs and outputs, we keep a mapping dictionary to make sure all of the dependencies in the blocks are equivalent (the same input/output combinations are used in each event). However, status tracker ids are never used for anything else, so we can just delete them. I'll add a comment to explain in the code

test/test_utils.py
abidlabs: Yeah I'll remove these copy.deepcopies

test/test_blocks.py
abidlabs: Sounds good! Removed the line and added a parameter `root_keys` which specifies which keys to test for equivlance

gradio/utils.py
abidlabs: The reason is so that we can test the remaining items dictionaries for equivalence later on when we do: `assert d1 == d2`

test/test_utils.py
abidlabs: Not quite, `test_equivalent_configs` is testing between `XRAY_CONFIG` and `XRAY_CONFIG_DIFF_IDS`. The configs have the same components, but different IDs

abidlabs:
Thanks @FarukOzderim and @freddyaboulton for all of the feedback! Should be addressed now

PR #1703, opened 2022-07-04
Allows setting custom <title>

### LINKED ISSUE #1691: Allow users to set the title in the browser tab (the <head> <title>) for Interfaces

When I share my app, I want the title bar to display the app name, rather than "Gradio", as it currently does.

Users in professional and other highly polished contexts, like me, want to control branding and style.

If I bring my own hosting I can control the URL and `Interface.launch` already includes a `favicon_path` (thanks!), but AFAICT the title can't be changed, except perhaps by writing custom css.

The "title" kwarg to an `Interface` sets the text in an h1 header at the top -- the simplest thing would be to duplicate that information in the `<head>` section's `<title>` element.

**Additional context**  
Here's a screenshot with the Gradio name in the title position:
image

abidlabs:
Allows users to provide a custom title which shows up on the browser tab. For an `Interface`, simply uses the existing `title` kwarg. For Blocks, there is a new `title` kwarg that has been added to the constructor.

Closes #1691

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -226,6 +226,7 @@ class Blocks(BlockContext):
         theme: str = "default",
         analytics_enabled: Optional[bool] = None,
         mode: str = "blocks",
+        title: str = "Gradio",
         css: Optional[str] = None,
         **kwargs,
     ):
@@ -272,6 +273,7 @@ class Blocks(BlockContext):
         self.auth = None
         self.dev_mode = True
         self.app_id = random.getrandbits(64)
+        self.title = title
 
     @property
     def share(self):
@@ -573,6 +575,7 @@ class Blocks(BlockContext):
             "components": [],
             "theme": self.theme,
             "css": self.css,
+            "title": self.title or "Gradio",
             "enable_queue": getattr(
                 self, "enable_queue", False
             ),  # attribute set at launch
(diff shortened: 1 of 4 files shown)


gradio/blocks.py
```diff
@@ -226,6 +226,7 @@ def __init__(
         theme: str = "default",
         analytics_enabled: Optional[bool] = None,
         mode: str = "blocks",
+        title: str = "Gradio",
```
FarukOzderim, line 229: Missing title in docstrings

gradio/interface.py
```diff
@@ -153,7 +153,11 @@ def __init__(
         analytics_enabled (bool | None): Whether to allow basic telemetry. If None, will use GRADIO_ANALYTICS_ENABLED environment variable if defined, or default to True.
```
FarukOzderim, line 153: cannot comment on line 143 but title's docstring seems outdated to me

FarukOzderim (commented):
good job!

freddyaboulton (approved):
Awesome fix @abidlabs ! Confirmed it worked on the demos.

Should we add a custom title to one of the demos, like `blocks_style.py` ?

abidlabs:
Thanks guys for the review! Will fix both and merge.

PR #1705, opened 2022-07-04
test hotfix

abidlabs:
Tests were failing because we removed the .egg file, which means that pkg_resources cannot fetch the version info. This should fix it.


FarukOzderim (commented):
good job!

PR #1709, opened 2022-07-05
allow users to update the current tab from python SDK

### LINKED ISSUE #1652: Allow currently selected `TabItem` to be set + updated
Currently there is no way to initialise or update the `Tabs` component to show a specific `TabItem` it always defaults to the first.

We need to implement the following:

- Add `selected` kwarg to `Tabs`
- Add `id` kwargs to `TabItem`
- Add an `update` method to `Tabs` that accepts a `selected` kwarg.
- Use the provided `TabItem` `id` as the actual id in the frontend
- Use the provided `selected` `id` passed to `Tabs` as the selected tab.

pngwn:
Closes #1652 and implement steps noted in the issue.

To test run `demo/blocks_plug`.

### DIFF
--- a/gradio/layouts.py
+++ b/gradio/layouts.py
@@ -96,7 +96,7 @@ class Tabs(BlockContext):
     Tabs context.
     """
 
-    def __init__(self, selected, **kwargs):
+    def __init__(self, selected=False, **kwargs):
         super().__init__(**kwargs)
         self.selected = selected
 
@@ -128,7 +128,7 @@ class TabItem(BlockContext):
     components defined within the TabItem will be rendered within a tab.
     """
 
-    def __init__(self, label, id, **kwargs):
+    def __init__(self, label, id=False, **kwargs):
         super().__init__(**kwargs)
         self.label = label
         self.id = id
--- a/gradio/layouts.py
+++ b/gradio/layouts.py
@@ -96,6 +96,21 @@ class Tabs(BlockContext):
     Tabs context.
     """
 
+    def __init__(self, selected, **kwargs):
+        super().__init__(**kwargs)
+        self.selected = selected
+
+    def get_config(self):
+        return {"selected": self.selected, **super().get_config()}
+
+    def update(
+        selected: Optional[int | str] = None,
+    ):
+        return {
+            "selected": selected,
+            "__type__": "update",
+        }
+
     def change(self, fn: Callable, inputs: List[Component], outputs: List[Component]):
         """
         Parameters:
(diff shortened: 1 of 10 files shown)


gradio/layouts.py
```diff
@@ -96,6 +96,21 @@ class Tabs(BlockContext):
     Tabs context.
     """
 
+    def __init__(self, selected=False, **kwargs):
```
abidlabs, line 99: Doesn't seem pythonic to have selected default to `False`. Can we have it default to `None` instead?

```suggestion
    def __init__(self, selected: Optional[int | str] = None, **kwargs):
```

gradio/layouts.py
```diff
@@ -96,6 +96,21 @@ class Tabs(BlockContext):
...
+        super().__init__(**kwargs)
+        self.selected = selected
+
+    def get_config(self):
+        return {"selected": self.selected, **super().get_config()}
+
+    def update(
+        selected: Optional[int | str] = None,
```
abidlabs, line 107: This typing is also correct if we go with the suggestion above.

gradio/layouts.py
abidlabs: The `selected` parameter needs documentation

gradio/layouts.py
```diff
@@ -113,12 +128,13 @@ class TabItem(BlockContext):
     components defined within the TabItem will be rendered within a tab.
     """
 
-    def __init__(self, label, **kwargs):
+    def __init__(self, label, id=False, **kwargs):
```
abidlabs, line 131: Let's use `None` instead of `False` 

```suggestion
    def __init__(self, label:str, id:Optional[int | str] = None, **kwargs):
```

abidlabs: Let's document these parameters

ui/packages/tabs/src/TabItem.svelte
```diff
@@ -5,12 +5,14 @@
 	export let elem_id: string = "";
 	export let visible: boolean = true;
 	export let name: string;
+	export let id: string | number | object | false = {};
```
abidlabs, line 8: Presumably this would no longer include "false" as an acceptable type if we make the changes above.

abidlabs:
Tested and works great! Just see my notes on typing

freddyaboulton:
@pngwn This is great! I confirmed the `blocks_plug` demo opens up to the specified tab. However, when I click the submit button on the first tab, it takes me to the second tab. That's counterintuitive and feels like a regression compared to `main`. I think we should fix that prior to merge? Same thing happens with the clear button. 

tab_input_change

pngwn:
@freddyaboulton Can you check that behaviour again now to confirm the fix. I think tests should be good now too.

PR #1712, opened 2022-07-05
Run integration tests on all prs

freddyaboulton:
# Description

Run integration tests on all prs as opposed to just those that change the ui

# Checklist:

### DIFF
--- a/.github/workflows/ui.yml
+++ b/.github/workflows/ui.yml
@@ -5,7 +5,6 @@ on:
     branches:
       - "main"
   pull_request:
-    types: [opened, synchronize]
 
 defaults:
   run:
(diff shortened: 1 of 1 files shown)


.github/workflows/ui.yml
```diff
@@ -4,11 +4,8 @@ on:
   push:
     branches:
       - "main"
-    paths:
-      - "ui/**"
   pull_request:
-    paths:
-      - "ui/**"
+    types: [opened, synchronize]
```
FarukOzderim, line 8: why do specify these two types?

freddyaboulton: So that this action runs when the pull request is opened and synchronized. In this case, synchronized means when  commits are pushed to the pr branch, the base branch is updated.

FarukOzderim: yeah, I understand, but doesn't CI work on all types, when we do not specify them?

PR #1716, opened 2022-07-06
allow data frame to change number of columns

### LINKED ISSUE #1198: Dataframe with headers throws error when new column is added

### Describe the bug

If a `gr.Dataframe` is created with defined `headers`, then when a new column is created, it throws an error, since the backend tries to apply the original headers to the new dataframe which is of a different size

### Reproduction

Minimum example:

```py
gr.Interface(lambda x:x, 
             gr.Dataframe(headers=["a", "b", "c"]), 
             "dataframe").launch()
```

### Screenshot

_No response_

### Logs

_No response_

### System Info

```shell
gradio==2.9b25
```

### Severity

critical

### LINKED ISSUE #1430: Dynamic Dataset Size for Dataframe Component
is there any way we can dynamically handle consuming of the dataframe in the background without limiting users to upload a certain size? 

Like when we create the dataframe component we have to give a list to headers which should match the number of columns in a dataset in a CSV/TSV file you're dragging to UI. This is not good for below reasons:
- one thing in tabular workflows is that we have so many feature engineering going on, we add modify datasets by adding one hot encoded features or eliminate features according to shap or importance, it would be reaaaalllly essential to have such feature.
- not so many people know about the number of columns of their dataset in a CSV file beforehand.

maybe ping @pngwn  for this 🙂
thought: We can change the size through UI, so it should be fine for CSV/TSV file read to have.

pngwn:
Closes #1198. Closes #1430.

This PR changes the preprocess signature for Dataframe so we have access to the current headers rather than only the initial ones. I'm not sure if there is an impact outside of what I've tested, there may also be some test failures that I may not understand.


gradio/components.py
```diff
@@ -2500,21 +2500,21 @@ def update(
...
         """
         if self.type == "pandas":
-            if self.headers:
-                return pd.DataFrame(x, columns=self.headers)
+            if x["headers"]:
+                return pd.DataFrame(x["data"], columns=x["headers"])
             else:
-                return pd.DataFrame(x)
+                return pd.DataFrame(x["data"])
         if self.col_count[0] == 1:
```
freddyaboulton, line 2512: I don't think it's correct to use the `self.col_count` defined when the component is created. If the headers are dynamic so is the `col_count`? 

In this repro, the function is only given access to the first column which is not the expected behavior imo

```python
import gradio as gr
import pandas as pd
import numpy as np

def return_empty(x):
    return np.atleast_1d(x.astype(float).mean(axis=0))


gr.Interface(return_empty,
             gr.Dataframe(type="numpy", headers=["a"]),
             gr.Dataframe(headers=["mean"])).launch()
```

gradio/components.py
```diff
@@ -2500,21 +2500,21 @@ def update(
     def preprocess(self, x: List[List[str | Number | bool]]):
```
freddyaboulton, line 2500: nit: Let's change the type hint here as well.

gradio/components.py
pngwn: What was this logic ever even doing? This feels wrong regardless of the changes in this PR. Can we remove? @abidlabs @aliabid94

test/test_components.py
```diff
@@ -1046,7 +1049,7 @@ def test_component_functions(self):
         )
         dataframe_input = gr.Dataframe()
         output = dataframe_input.preprocess(x_data)
-        self.assertEqual(output[1][1], 24)
+        self.assertEqual(output.iloc[1]["Age"], 24)
```
freddyaboulton, line 1052: Left some comments explaining why on slack but `self.assertEqual(output["Age"][1], 24)`  makes the diff easier to explain and justify. Not blocking though

gradio/components.py
```diff
@@ -2521,26 +2525,22 @@ def update(
...
             Dataframe in requested format
         """
         if self.type == "pandas":
-            if self.headers:
-                return pd.DataFrame(x, columns=self.headers)
+            if x.get("headers") is not None:
+                return pd.DataFrame(x["data"], columns=x.get("headers"))
             else:
-                return pd.DataFrame(x)
-        if self.col_count[0] == 1:
```
freddyaboulton, line 2538: Awesome, make sense to delete `self.col_count`

test/test_components.py
```diff
@@ -1140,7 +1143,10 @@ def check_odd(array):
             return array % 2 == 0
 
         iface = gr.Interface(check_odd, "numpy", "numpy")
-        self.assertEqual(iface.process([[2, 3, 4]])[0], {"data": [[True, False, True]]})
+        print(iface.process([{"data": [[2, 3, 4]]}]))
```
freddyaboulton, line 1146: nit: Think we should remove this print statement prior to merge

demo/blocks_plug/run.py
```diff
@@ -1,36 +1,19 @@
 import gradio as gr
 
+import gradio as gr
```
freddyaboulton, line 3: nit: We're re-importing gradio here and we're not using pandas

gradio/components.py
```diff
@@ -2389,7 +2389,11 @@ def style(
         )
 
 
-@document()
+class DataframeData(TypedDict):
```
freddyaboulton, line 2392: Nice

demo/blocks_plug/run.py
freddyaboulton: Actually, should we move this to another demo? I think `blocks_plug` is a nice demo for how to set named tabs and don't want to lose it unless there's a good reason for it.

freddyaboulton (approved):
Thanks for this feature @pngwn ! Python looks good to me. Left some minor nits (I think we should get rid of the print statements) but otherwise nothing blocking merge.

test/test_components.py
pngwn: ops

demo/blocks_plug/run.py
pngwn: Yeah something weird happened after a merge and I'm not sure what i did when resolving the conflict. I'll check this against main to see what is what and create a new demo for the tab changing if necessary.

pngwn: reinstated, no idea what happened.

Issue #1362, opened 2022-05-23
Exclude inputs from examples


jbdel:
**Is your feature request related to a problem? Please describe.**  
By default _**examples**_ shows a columns for every input, even if they are empty
Even
`examples = [[i, None, None, None, None] for i in glob.glob("./images/*")]
`
or 
`examples = [[i] for i in glob.glob("./images/*")]
`
Still show empty columns for inputs at index 1,2,3 and 4

**Describe the solution you'd like**  
Sometimes we have multiple inputs, but just want examples of one input, say images. We don't want a example column for the input "how many images to generate" or "top_k" samples.

abidlabs:
Makes sense, and shouldn't be too hard to support! @dawoodkhan82 would you be able to take a look at this as part of your examples fixes?

sciai-ai:
A) from a ui standpoint it showed redundant inputs on the screen

B) these empty inputs actually correspond to gr.text or gr.slider abd have nothing to do with examples

abidlabs:
It feels to me that both from a UI point of view and user intention, it would be better to exclude these inputs from showing up. The same thing was brought up in #1473

jbdel:
> I actually think it should stay as it is, since it shows that input can or will be blank. That's an information users can make use of. ![image](https://user-images.githubusercontent.com/44067394/171200382-ca371084-e4c4-4380-8c3e-abb35dc42240.png)

I think its pretty straightforward for the user to infer that inputs will stay as it is if not mentioned in an example. Moreover, in your example, the inputs "max_length" or "num_return_sequence" will be indeed be blank, likely causing the model to crash or trigger error. 

Also, i'm not suggesting this to be the default behavior, just to add an 'exclude' option that the programmer could use.

rnwang04:
In my case, I have two inputs : image and radio, and i just want to provide some images as examples.
But if I set the radio value to None, corresponding image can't be used and it's really inconvenient.
'exclude' option may help.

PR #1722, opened 2022-07-06
Allow excluding input components in examples

abidlabs:
As pointed out in #1362, for demos with many examples, it is often desirable to exclude certain input components from being part of examples. This PR implements that functionality.

If *all* example values corresponding to a particular component are `None`, then that column is not displayed as part of the `Examples` object. Clicking on the examples has no effect on that particular input component.  

Test with: `demo/fake_gan/run.py`

Fixes: #1362


freddyaboulton (approved):
Looks good to me @abidlabs !

Issue #1717, opened 2022-07-06
Miscellaneous formatting improvements to website after PR 1578


freddyaboulton:
- [x] I have searched to see if a similar issue already exists.

In the review of #1578, we noticed some miscellaneous things we'd like to improve about the new website design:

- [x] No sidebars in the guides

- [x] **kwargs in Series/Parallel is shown as required when it is not
![image](https://user-images.githubusercontent.com/41651716/177609864-f10d2acb-23de-4684-8436-118dabfd27c8.png)

- [x] No example Number/Text (other input component) in the docs
### Old website
![image](https://user-images.githubusercontent.com/41651716/177611117-ffa2f3d4-8665-49c3-a5a2-68924bad2471.png)
### Current website locally
![image](https://user-images.githubusercontent.com/41651716/177611177-ed6d26ec-93c1-470d-93b4-683a95eff201.png)

- [x] No example output HighlightedText component in the docs
### Old Website
![image](https://user-images.githubusercontent.com/41651716/177613303-d43b336e-624e-4f04-a149-3f63fec1f9d7.png)
### Current website locally
![image](https://user-images.githubusercontent.com/41651716/177613264-1310669d-e569-4b65-bf17-3179dbc62ac4.png)

- [x] Missing docstring descriptions for parameters of `Row`, `Column`
![image](https://user-images.githubusercontent.com/41651716/177613958-26b8caae-27ce-4eb8-8151-aa71ee1a7563.png)

- [x] Missing description for `color_map` in `Chatbot`
### Old website
![image](https://user-images.githubusercontent.com/41651716/177614151-c2aac094-96f6-470e-9db1-4850c5896351.png)
### Current website locally
![image](https://user-images.githubusercontent.com/41651716/177614111-d93b7ec7-044b-4cfd-b7ed-078176bd9c00.png)

- [x] The text on "Developing faster with reload mode" guide is not aligned
![image](https://user-images.githubusercontent.com/41651716/177614680-51f2fd4b-8e1c-4ff2-b46a-37cef73538bd.png)

aliabd:
Thanks for putting this together @freddyaboulton, will start checking them off today. 

Just a note that all the embedded example components are currently missing on the docs (not just Textbox and HighlightedText). We removed them because we had hardcoded the config, and they wouldn't update if anything changed. I'm going to open a PR soon to launch and embed them the same way we do for demos.

freddyaboulton:
Thats awesome @aliabd !

[...]

PR #1726, opened 2022-07-07
Embedding Components on Docs

aliabd:
Embeds components on the docs page so readers can immediately see what they look like and interact with them. 

The old way we did this was by hardcoding the configs, which meant they wouldn't change if the library did. This launches them the same way as demos. For most of the input components, the code is just this (to launch a demo with only the component):

```python
import gradio as gr
with gr.Blocks() as demo:
    gr.[COMPONENT]()
demo.launch()
```

But some require parameters (like CheckboxGroup, etc) or some are output only. All the code to launch the demos is in `/demo/components_demos/run.py`, and modifying this will change the embedded component.


website/demos/run_demos.py
```diff
@@ -22,19 +22,34 @@
...
-    assert "demo.launch()" in filedata, demo_name + " has no demo.launch()\n" + filedata 
+    if demo_name.endswith("_component"):
+        if os.path.exists(demo_file):
+            demo_file = os.path.join(demo_folder, "run.py")
+            with open(demo_file, "r") as file:
+                filedata = file.read()
+            assert "demo.launch()" in filedata, demo_name + " has no demo.launch()\n" + filedata
+        else:
+            os.mkdir(demo_folder)
+            filedata = component_launch_code.format(component_name=demo_name[:-10])
```
freddyaboulton, line 43: I think it would be better if we defined a `SUFFIX="_component"` constant and then used it where it's needed.

Instead of `demo_name.endwith("_component")` we can do `demo_name.endswith(SUFFIX)` and `.format(component_name=demo_name[:-len(SUFFIX)]` as opposed to `format(component_name=demo_name[:-10])`

freddyaboulton (approved):
Thank you so much for adding this back in @aliabd ! Already made a note of this in https://github.com/gradio-app/gradio/issues/1717 !

website/demos/run_demos.py
aliabd: Agreed, thanks.

aliabd:
Wonder what people think about the `_component` folders being in `/demos`, is it going to confuse users?

website/demos/run_demos.py
abidlabs: What will happen if we create a new component that requires constructor parameters? The demo will not work properly for sure, but will it break the website build?

aliabd:
@abidlabs there are a few of us who use the dev server now, I think when you checked someone else was testing a different change and checked out out of this branch. My mistake for linking to it as that's not how it should be used. 

https://user-images.githubusercontent.com/9021060/177973616-39d164b2-34a9-4a43-bb9a-9591d7abe348.mov

website/demos/run_demos.py
aliabd: To be clear, there isn't a component right now with required parameters. For example, `CheckboxGroup` should get `choices` but they're not required in the constructor, they default to `None`

So

```python
import gradio as gr
with gr.Blocks() as demo:
     component = gr.CheckboxGroup()
demo.launch()
```

doesn't break the website, it just shows up like this:

Screen Shot 2022-07-08 at 5 50 52 PM

If we create a component with required parameters in the constructor, but don't override with a `_component` demo, it will break the website build.

abidlabs: Ok got it, I think we should try catch this to be more robust but otherwise LGTM!

abidlabs:
LGTM!

aliabd:
As suggested by @aliabid94, moved all the `_component` files to one file: `components_demos.py`

website/demos/run_demos.py
```diff
@@ -22,22 +22,39 @@
 def launch_demo(demo_folder):
     subprocess.check_call([f"cd {demo_folder} && python run2.py"], shell=True)
 
+SUFFIX = "_component"
```
aliabid94, line 25: better name, COMPONENT_SUFFIX?

website/demos/run_demos.py
```diff
@@ -22,22 +22,39 @@
...
+        demo_folder = os.path.join(GRADIO_DEMO_DIR, demo_name)
+        demo_file = os.path.join(demo_folder, "run.py")
+        demo_2_file = os.path.join(demo_folder, "run2.py")
+        with open(demo_file, "r") as file:
+            filedata = file.read()
+        assert "demo.launch()" in filedata, demo_name + " has no demo.launch()\n" + filedata
+        filedata = filedata.replace(f"demo.launch()", f"demo.launch(server_port={port}, _frontend=False)")
+        with open(demo_2_file, "w") as file:
+            file.write(filedata)
+        demo_thread = threading.Thread(target=launch_demo, args=(demo_folder,))
```
aliabid94, line 52: some duplicated code across this if/else branches

aliabid94 (approved):
LGTM beside small nits

website/homepage/src/docs/template.html
```diff
@@ -245,6 +245,19 @@ <h3 class="text-3xl font-light my-4" id="block-layouts">Block Layouts</h3>
       addCopyButtons(navigator.clipboard);
     };
 
+
+    // Remove built-with-gradio footers and extra space from embedded components
```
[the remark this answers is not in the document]
aliabd: will do this in a separate PR

Issue #1644, opened 2022-06-27
Add information on how to install python test dependencies for M1 development


freddyaboulton:
- [x] I have searched to see if a similar issue already exists. 

Similar issue here https://github.com/gradio-app/gradio/issues/1369 but that was closed as there is no problem installing the gradio pip package on M1 mac. This issue is related to installing the test dependencies in M1 mac. 

**Is your feature request related to a problem? Please describe.**  
As a Gradio developer, I wish the contributing guide provided some information on how to fix some of the common problems that will arise when installing the test dependencies on an M1 mac, e.g. `tokenizers` does not publish a wheel for osx arm architectures. 

**Describe the solution you'd like**  
Update development guide to include information on how to install test dependencies on M1 mac. Alternatively can we remove some dependencies on `requirements.in`, e.g. `tensorflow` ? 

**Additional context**

abidlabs:
I'm in favor of removing `tensorflow`! If `pytorch` is easier to run on M1, let's switch to that.

FarukOzderim:
@hibestil had problems with test requirements as far as I know, he might help if he has solved them as well.

PR #1730, opened 2022-07-08
remove `tensorflow` from testing dependency

abidlabs:
Removes `tensorflow` from our testing dependencies as it was causing issues on Macs with M1s. I don't think it should be needed if `torch` is present.

Closes: #1644


abidlabs:
The tests pass on CI which confirms that `tensorflow` isn't a required dependency. But @pngwn let me know if this facilitates testing on Macs with M1s, or whether we need to shave off more dependencies.

pngwn:
It is still trying to install tensorflow for me locally, is there something i need to do to make it work? I'm just using the script in `/scripts/`.

pngwn:
I installed manually and it worked fine (the instructions that @abidlabs gave me directly). I'm going to merge this so I can carry on with the DF fix.

PR #1735, opened 2022-07-08
Mark tests as flaky and run in separate job

freddyaboulton:
# Description

Mark flaky tests and run them in a separate job. My hope is that we make the "not flaky" tests as required. We should still look into failures in the flaky tests but not prevent people from merging because of them. 

Curious what people think. Certainly we can also make some modifications to `test_external.py` so that all exceptions are ignored but I think it's still valuable to run the tests in separate jobs. 

Tests properly selected:
image

image

# Checklist:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -64,7 +64,7 @@ jobs:
       - run:
           command: |
             . venv/bin/activate
-            coverage run -m pytest -m << parameters.test-type >>
+            coverage run -m pytest -m "<< parameters.test-type >>"
             coverage xml
       - codecov/upload:
           file: "coverage.xml"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,15 +3,6 @@ orbs:
   codecov: codecov/codecov@3.2.2
   node: circleci/node@5.0.2
 
-workflows:
-  backend:
-    jobs:
-      - build:
-          matrix:
-            parameters:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -2,8 +2,12 @@ version: 2.1
 orbs:
   codecov: codecov/codecov@3.2.2
   node: circleci/node@5.0.2
+
 jobs:
-  build:
+  test:
```
freddyaboulton, line 7: Changing the name to test so that the intent is clearer also goes better with the concept of "flaky"

abidlabs:
Very nice @freddyaboulton! Will look deeply into this later today but this definitely sounds like an improvement

abidlabs:
This looks great @freddyaboulton! If we can get the CircleCI badge on our README to reflect that status of the non-flaky tests, that would be perfect.

image

abidlabs (approved):

dawoodkhan82 (approved):
Looks good. I agree with @abidlabs about the circleci badge, if possible.

freddyaboulton:
@abidlabs @dawoodkhan82 Going to merge. Can't fix the circle ci badge unfortunately. The overall status is determined by all workflows in a branch. I know github actions let's you create a [badge per workflow](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/adding-a-workflow-status-badge) so that could be a pro to migrating the backend tests to github actions.

Issue #1665, opened 2022-06-29
Require CI to pass in order to merge a PR


freddyaboulton:
- [ ] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
Right now the CI checks are not required for merging a PR because they often flake. I was chatting with @pngwn about this and we agree this isn't great for the developer culture because it leads developers to not trust the tests and assume any failures are flakes. This also discourages first time contributors because a test that may or may not have anything to do with their change will fail and they may not have the experience with the repo to know if it's a legitimate issue with their code.

**Describe the solution you'd like**  
We should make the checks required prior to merge but we should make sure they are not flakey prior to do so. This issue tracks work related to improving test suite reliability in order to make the tests required. 

Adding the brainstorming tag so that we can discuss others' thoughts on this issue as well as other requirements needed to make the test suite required for merging.

**Additional context**

pngwn:
I don't think this issue is very clear, the title is discussing required checks but the body is mostly about making the test suites more robust. 

I don't think there will be many objections to either but I'm not certain what this issue is for. Is it to discuss whether or not we should have required checks for PR merges, or to discuss + track progress towards making the test suites more reliable?

freddyaboulton:
Yea good point. I think the test suite should be reliable prior to making it a required check but you’re right that those are two separate concepts. I’ll rename this issue to make it clear it tracks making the test suite more reliable since that’s what the body mostly talks about. I’ll file a separate issue to track discussion around required checks.

abidlabs:
I think we've managed to fix this now, but on a related note @freddyaboulton I've noticed tests always show as "failing" when new contributors are merging in a fork of their repo, since they can't deploy on Spaces, e.g. https://github.com/gradio-app/gradio/pull/2234

Is it possible to change the deploy job to not run in these cases?

freddyaboulton:
Yea I think we can program the job to exit early with a success if it's running from a fork

freddyaboulton:
Closing as the test suite has not flaked in a while

PR #1742, opened 2022-07-08
Updated `atomicwrites` version

abidlabs:
To get the tests to pass!

### DIFF
--- a/test/requirements.txt
+++ b/test/requirements.txt
@@ -16,7 +16,7 @@ astunparse==1.6.3
     # via tensorflow
 asyncio==3.4.3
     # via -r requirements.in
-atomicwrites==1.4.0
+atomicwrites==1.4.1
     # via pytest
 attrs==21.4.0
     # via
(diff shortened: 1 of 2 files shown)


abidlabs:
This fix raises the question, why do we have hardcoded versions for our testing requirements in `test\requirements.txt`? As opposed to just pip installing `test\requirements.in`? That would avoid issues like this.

cc @FarukOzderim @freddyaboulton

freddyaboulton:
Fair point but having constantly changing dependency versions being installed may increase flakiness and make it harder to compare runs.

It's much more likely that a new version will break something than a maintainer yanking an existing version in a rage quit 😂

freddyaboulton (approved):
Thank you for fixing @abidlabs and for going through the trouble of updating a space that was causing failures!

I think this is good to get the CI passing again but we need a more sustainable solution. Text-to-speech models may fail next week or something like that. 

Please put any ideas you may have on the matter here: https://github.com/gradio-app/gradio/issues/1665

I think #1735 can help in that regard.

PR #1744, opened 2022-07-08
Fixes multiple people using `HuggingFaceDatasetSaver` simultaneously

abidlabs:
If multiple people were simultaneously flagging a demo using `HuggingFaceDatasetSaver`, it would error out because we didn't pull the dataset repo before changing it and trying to push the changes. This fixes and that and closes: #1727

### DIFF
--- a/gradio/flagging.py
+++ b/gradio/flagging.py
@@ -279,7 +279,7 @@ class HuggingFaceDatasetSaver(FlaggingCallback):
     ) -> int:
         is_new = not os.path.exists(self.log_file)
         infos = {"flagged": {"features": {}}}
-        
+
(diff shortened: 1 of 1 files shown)


abidlabs:
I released `gradio==3.0.23.dev1` to be able to test this version on Spaces. See here for Space: https://huggingface.co/spaces/abidlabs/HF-Callback

You can test this by opening up the app on two separate tabs and flagging data from both of them (without refreshing either page).

abidlabs:
Thanks for the test @freddyaboulton, let me fix the concurrency issue. 

Agreed, it would be nice to have a separate json callback

PR #1769, opened 2022-07-12
Run all demos in website, not just components

freddyaboulton:
# Description

This PR makes it so that all demos tagged in docstrings are run on the website. Previously, we only ran demos for components if they were not listed in the guides. As we add more apis to our docs I think it will be important to automatically run the demos beyond just the components.

### Screenshot of demos run prior to this change
Screen Shot 2022-07-12 at 12 11 51 PM

### Screenshot of demos after this change
Note that `interface_series` and `interface_parallel` are now in the list

image

# Checklist:

### DIFF
--- a/gradio/inputs.py
+++ b/gradio/inputs.py
@@ -42,7 +42,6 @@ class Number(components.Number):
     """
     Component creates a field for user to enter numeric input. Provides a number as an argument to the wrapped function.
     Input type: float
-    Demos: tax_calculator, titanic_survival
     """
 
     def __init__(
@@ -68,7 +67,6 @@ class Slider(components.Slider):
     """
     Component creates a slider that ranges from `minimum` to `maximum`. Provides number as an argument to the wrapped function.
     Input type: float
-    Demos: sentence_builder, generate_tone, titanic_survival
     """
 
     def __init__(
@@ -108,7 +106,6 @@ class Checkbox(components.Checkbox):
     """
     Component creates a checkbox that can be set to `True` or `False`. Provides a boolean as an argument to the wrapped function.
     Input type: bool
-    Demos: sentence_builder, titanic_survival
     """
 
     def __init__(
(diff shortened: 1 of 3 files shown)


gradio/inputs.py
```diff
@@ -42,7 +42,6 @@ class Number(components.Number):
     """
     Component creates a field for user to enter numeric input. Provides a number as an argument to the wrapped function.
     Input type: float
-    Demos: tax_calculator, titanic_survival
```
freddyaboulton, line 45: Removing the demos from these deprecated files. We should not be running demos for deprecated components.

website/demos/map_demos.py
```diff
@@ -13,11 +14,16 @@
 port = 7860
 
 demos_to_run = []
-GRADIO_COMPONENTS_FILE = os.path.join(GRADIO_DIR, "gradio", "components.py")
-with open(GRADIO_COMPONENTS_FILE) as comp_file:
-    comp_text = comp_file.read()
-for demostr in re.findall(r'Demos:(.*)', comp_text):
-  demos_to_run += re.findall(r'([a-zA-Z0-9_]+)', demostr)
+for root, _, files in os.walk(pathlib.Path(GRADIO_DIR) / "gradio"):
```
freddyaboulton, line 17: Just applying the same logic as before but on all `.py` files in the repo

website/demos/map_demos.py
```diff
@@ -13,11 +14,16 @@
...
+    for filename in files:
```
freddyaboulton, line 18: Was able to confirm the demos still build correctly on a local docker compose up

image

aliabid94 (approved):
Makes sense to me. Something we could implement in the future that's even cleaner is to search all the documentation returned from `generate_documentation()` to check demos within the docstrings only.

freddyaboulton:
Completely agree that would be much cleaner @aliabid94 Filed https://github.com/gradio-app/gradio/issues/1783 to track that. Thank you!

Issue #1688, opened 2022-07-01
Add guide (or modify existing guides or demos) to show how to set default values on components


freddyaboulton:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
As a gradio user, I wish the guides were clearer on how to set default values on components. This is documented on the docs page but some users may not now to look there.

From the discord:

<img width="634" alt="image" src="https://user-images.githubusercontent.com/41651716/176954825-2b63f6f5-e7cd-43db-a183-942be8b034a0.png">

<img width="591" alt="image" src="https://user-images.githubusercontent.com/41651716/176954703-f1eb73af-2559-4e45-b3cd-ac5dcccfd0b1.png">

**Describe the solution you'd like**  
Modify existing guides and or demos (or write a new guide) on how to set default values for components. 

**Additional context**

PR #1788, opened 2022-07-13
Add info to docs on how to set default values for components

freddyaboulton:
# Description

Fixes: #1688

Went about this by setting `value` via keyword arg in the examples and adding a section to the blocks guide about default values

# Checklist:


demo/blocks_inputs/run.py
```diff
@@ -12,14 +12,14 @@ def mirror(x):
     txt = gr.Textbox(label="Input", lines=2)
     txt_2 = gr.Textbox(label="Input 2")
     txt_3 = gr.Textbox("", label="Output")
-    btn = gr.Button("Submit")
```
freddyaboulton, line 15: In the original issue, the discord user said he couldn't find demos using `value` despite it being used in demos already. 

I figured that was because we didn't set it explicitly with a keyword argument so I went through some demos and set value via keyword arg.

demo/image_mod_default_image/run.py
```diff
@@ -0,0 +1,18 @@
+import gradio as gr
```
freddyaboulton, line 1: Adding another demo that sets the default value on the image since there wasn't an example doing that already and that was mentioned on the discord.

demo/blocks_inputs/run.py
abidlabs: Good idea

abidlabs:
LGTM!

abidlabs (approved):

FarukOzderim (approved):
LGTM! 

You might want to change demo/image_mod_default_image/screenshot.png with a 10KB file instead of 1MB with `git soft reset` to not increase git size.

PR #1794, opened 2022-07-14
Fix Request Tests

freddyaboulton:
# Description

Requests unit tests were failing because `https://reqres.in/api/users` is unavailable. 

This PR mocks out the calls to that api to make the tests pass and not have to rely on external services

# Checklist:


freddyaboulton:
Lol looks like the service is back up again. I still think it's worthwhile to remove dependencies on external services from unit tests so I'll keep this PR open to hear what people think!

test/test_utils.py
```diff
@@ -269,8 +269,20 @@ async def test_get(self):
         assert client_response.is_valid() is True
         assert validated_data["Host"] == "headers.jsontest.com"
```
abidlabs, line 270: Let's mock this request to "headers.jsontest.com" as well?

abidlabs:
Thanks for fixing this @freddyaboulton! Agreed that this is definitely the better approach. I noticed a couple of things:
* There are some Requests that are not mocked. See the tests in lines 263, 323, 347. Let's mock these as well?
* We currently define the same `MagicMock` across various functions. Would be good to apply the mock at a class level to avoid repeating code: https://docs.python.org/3/library/unittest.mock-examples.html#applying-the-same-patch-to-every-test-method

abidlabs:
Thanks for fixing this @freddyaboulton! Agreed that this is definitely the better approach. I noticed a couple of things:

* There are some Requests that are not mocked. See the tests in lines 263, 323, 347. Let's mock these as well?
* We currently define the same MagicMock across various functions. Would be good to apply the mock at a class level to avoid repeating code: https://docs.python.org/3/library/unittest.mock-examples.html#applying-the-same-patch-to-every-test-method

abidlabs:
Thanks for fixing this @freddyaboulton! Agreed that this is definitely the better approach. I noticed a couple of things:

* There are some Requests that are not mocked. See the tests in lines 263, 323, 347. Let's mock these as well?
* We currently define the same MagicMock across various functions. Would be good to apply the mock at a class level to avoid repeating code: https://docs.python.org/3/library/unittest.mock-examples.html#applying-the-same-patch-to-every-test-method

abidlabs:
Very clean, thanks for addressing the suggestions @freddyaboulton!

abidlabs (approved):

FarukOzderim:
Sounds good! actually there is a similar library for this which I like, it mocks the request call and you don't mock your code at all. I would suggest we use it, but this works as well

The library is [aioresponses](https://github.com/pnuckowski/aioresponses) and there is a pytest [plugin ](https://github.com/pheanex/pytest-aioresponses)as well.

PR #1796, opened 2022-07-14
Remove usage of deprecated gr.inputs and gr.outputs from website

### LINKED ISSUE #1786: Remove usage of deprecated inputs and outputs from docs and guides

As a gradio user, I don't want to see usage of the deprecated `gr.inputs` and `gr.outputs` in the examples and guides.

Here is an example from the "Hugging face integrations" guide

image

Remove usage of deprecated `gr.inputs` and `gr.outputs` guides. 

**Additional context**

freddyaboulton:
# Description

Fixes #1786

# Checklist:

### DIFF
--- a/guides/advanced_interface_features.md
+++ b/guides/advanced_interface_features.md
@@ -68,7 +68,7 @@ gr.Interface.load("huggingface/gpt2").launch();
 
 ```python
 gr.Interface.load("huggingface/EleutherAI/gpt-j-6B", 
-    inputs=gr.inputs.Textbox(lines=5, label="Input Text")  # customizes the input component
+    inputs=gr.Textbox(lines=5, label="Input Text")  # customizes the input component
 ).launch()
 ```
 
--- a/demo/image_mod/run.py
+++ b/demo/image_mod/run.py
@@ -6,7 +6,7 @@ def image_mod(image):
     return image.rotate(45)
 
 
-demo = gr.Interface(image_mod, gr.inputs.Image(type="pil"), "image", 
+demo = gr.Interface(image_mod, gr.Image(type="pil"), "image",
     flagging_options=["blurry", "incorrect", "other"], examples=[
         os.path.join(os.path.dirname(__file__), "images/cheetah1.jpg"),
         os.path.join(os.path.dirname(__file__), "images/lion.jpg"),
--- a/guides/Gradio_and_ONNX_on_Hugging_Face.md
+++ b/guides/Gradio_and_ONNX_on_Hugging_Face.md
@@ -129,7 +129,7 @@ def inference(img):
 title = "EfficientNet-Lite4"
 description = "EfficientNet-Lite 4 is the largest variant and most accurate of the set of EfficientNet-Lite model. It is an integer-only quantized model that produces the highest accuracy of all of the EfficientNet models. It achieves 80.4% ImageNet top-1 accuracy, while still running in real-time (e.g. 30ms/image) on a Pixel 4 CPU."
 examples = [['catonnx.jpg']]
-gr.Interface(inference, gr.inputs.Image(type="filepath"), "label", title=title, description=description, examples=examples).launch()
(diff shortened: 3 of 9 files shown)


demo/chatbot_demo/run.py
```diff
@@ -17,11 +17,12 @@ def chat(message, history):
     return history, history
 
 
+state = gr.Variable(value=[])
```
freddyaboulton, line 20: using "state" is technically deprecated so I'm using `gr.Variable` here.

It makes the demo more verbose but it's more aligned with our current api. What do people think? Also had to modify the chatbot and real-time speech recognition guides.

guides/advanced_interface_features.md
```diff
@@ -123,7 +123,13 @@ This sets up a queue of workers to handle the predictions and return the respons
...
 
-Another type of data persistence Gradio supports is session **state**, where data persists across multiple submits within a page load. However, data is *not* shared between different users of your model. To store data in a session state, you need to do three things: (1) Pass in an extra parameter into your function, which represents the state of the interface. (2) At the end of the function, return the updated value of the state as an extra return value (3) Add the `'state'` input and `'state'` output components when creating your `Interface`. See the chatbot example below: 
+Another type of data persistence Gradio supports is session **state**, where data persists across multiple submits within a page load. However, data is *not* shared between different users of your model. To store data in a session state, you need to do three things:
+
+1. Pass in an extra parameter into your function, which represents the state of the interface.
```
freddyaboulton, line 128: Unrelated to the original purpose of this PR but I think using a numbered list is easier to read here:
Screen Shot 2022-07-14 at 4 00 39 PM

abidlabs: Agreed thanks @freddyaboulton!

guides/creating_a_chatbot.md
```diff
@@ -41,7 +41,7 @@ Next, you will need to define a function that takes in the *user input* as well
 In the case of our pretrained model, it will look like this:
 
 ```python
-def predict(input, history=[]):
+def predict(input, history):
```
abidlabs, line 44: I think we need this to set the original value of the state

demo/chatbot_demo/run.py
abidlabs: I think you can still use "state" shortcut in Interfaces. Is this demo not working for you as is @freddyaboulton? Because I just tested it on my end and it is working for me!

cc @aliabid94 as well

abidlabs: Actually I think "state" shortcut is not deprecated in `Interfaces`. Is this demo not working for you as is @freddyaboulton? Because I just tested it on my end and it is working for me!

cc @aliabid94 as well

guides/advanced_interface_features.md
```diff
@@ -123,7 +123,13 @@ This sets up a queue of workers to handle the predictions and return the respons
...
+2. At the end of the function, return the updated value of the state as an extra return value.
+3. Create a `gr.Variable` and pass it as an input and output component of the `Interface`.
```
abidlabs, line 130: Let's resolve the issue about the string shortcut. I think it is an acceptable approach within `Interface`. 

We've deprected "gr.State" but the "state" shortcut should still work, it just maps to "gr.Variable" instead

demo/chatbot_demo/run.py
```diff
@@ -17,11 +17,12 @@ def chat(message, history):
...
 chatbot = gr.Chatbot(color_map=("green", "pink")).style()
```
abidlabs, line 21: On the other hand, `color_map` is not part of the constructor for `gr.Chatbot`, it's been moved to `style()`

abidlabs:
Great, thanks for the fix @freddyaboulton! Have some considerations around "state" and whether we should really deprecate it, but otherwise LGTM.

cc @aliabid94

freddyaboulton:
Thanks for the review @abidlabs ! Should be good for another look now - removed all the changes related to the "state" string shortcut.

FarukOzderim (approved):
Thanks for removing deprecated parts from the codebase :)

PR #1799, opened 2022-07-15
Add a guide on how to use apps like functions

### LINKED ISSUE #1795: Add a guide on how to use blocks like regular python functions

As a gradio user, I wish there was a guide explaining how to use use a blocks app like a regular python function. Right now, it's only mentioned briefly for interfaces in the "Advanced Interfaces" [guide](https://gradio.app/advanced_interface_features/#loading-hugging-face-models-and-spaces) but there is no mention that this works for Blocks.

Some of our users are asking about this on discord

image

A guide describing how to use blocks apps like regular python functions and how this relates to loading from hugging face spaces.

We should also add the `load` method to the documentation. 

**Additional context**

freddyaboulton:
# Description

Fixes #1795

# Checklist:


gradio/blocks.py
```diff
@@ -671,21 +671,19 @@ def load(
...
+        Class method: loads a demo from a Hugging Face Spaces repo and creates it locally and returns a block instance.
+
+
+        Instance method: adds an event for when the demo loads in the browser and returns None.
         Parameters:
-            fn: Callable function
-            inputs: input list
-            outputs: output list
-        Returns: None
+            name: Class Method - the name of the model (e.g. "gpt2"), can include the `src` as prefix (e.g. "models/gpt2")
```
freddyaboulton, line 680: I think adding the load method to the docs page will increase it's visibility and help with the original problem this issue tracks, which is that users don't know they can load Blocks demos.

Only tricky thing is that this is a class and instance method, and the current documentation building code does not naturally display two separate parameters tables. So I decided to combine them.

image

guides/using_blocks_like_functions.md
```diff
@@ -0,0 +1,72 @@
+<script type="module" src="https://gradio.s3-us-west-2.amazonaws.com/3.0.24/gradio.js"></script>
```
freddyaboulton, line 1: I debated whether or not to include this in an already existing guide or add it to its own guide.

The two other guides where this could go are:
1. Introduction to Blocks
2. Using Hugging Face Integrations

I don't think this is an introductory concept, so it'd be out of place in the introduction to blocks guide.

It could go in the integrations guide but there's a lot already happening there so I'm afraid this content would get lost.

Ultimately I felt this was a concept worth going into depth so I added it to its own guide.

guides/using_hugging_face_integrations.md
```diff
@@ -112,6 +112,8 @@ The code above generates the following demo.
 
 As you can see, the demo looks the same, but it uses a webcam input instead of user-uploaded images.
 
+You can learn more about this feature, and how to use it with the new Blocks API in the [Treating Gradio apps like functions guide](/using_blocks_like_functions/)
```
freddyaboulton, line 115: This guide already covers material similar to the proposed "Using apps like functions" guide so I debated whether or not we even need to add another guide.

I decided that we did because:
1. The material in this guide only covers Interfaces, which makes users think it only applies to Interfaces and demos built with Interfaces. 
2. It doesn't cover the full depth of the concept - just how you can change the input type. I think the concept that apps are functions that could be used anywhere is worth highlighting.

gradio/blocks.py
abidlabs: 👍

guides/using_apps_like_functions.md
```diff
@@ -0,0 +1,72 @@
+<script type="module" src="https://gradio.s3-us-west-2.amazonaws.com/3.0.24/gradio.js"></script>
+
+# Using Gradio Apps Like Functions
```
abidlabs, line 3: How about:
```suggestion
# Using Gradio Blocks Like Functions
```

guides/using_apps_like_functions.md
```diff
@@ -0,0 +1,72 @@
...
+
+Docs: Blocks
+
+**Prerequisite**: This Guide builds on the Blocks Introduction. Make sure to [read that guide first](/introduction_to_blocks).
+
+## Introduction
+
+Did you know that apart from being a full-stack machine learning demo, a Gradio app is also a regular-old python function!?
```
abidlabs, line 11:
```suggestion
Did you know that apart from being a full-stack machine learning demo, a Gradio Blocks app is also a regular-old python function!?
```

guides/using_apps_like_functions.md
```diff
@@ -0,0 +1,72 @@
...
+
+This means that if you have a gradio app called `demo`, you can use `demo` like you would any python function.
```
abidlabs, line 13:
```suggestion
This means that if you have a gradio Blocks (or Interface) called `demo`, you can use `demo()` like you would any python function.
```

abidlabs:
LGTM @freddyaboulton, but I'm wondering if we can use a more complex Space than a string concatenator to make the Guide more interesting to readers? 

For example here is a translation model Space: https://huggingface.co/spaces/abidlabs/english2german (though this particular Space is an Interface, would be easy to create a version of it that is a regular Blocks)

FarukOzderim (approved):

freddyaboulton:
@abidlabs thank you for the review! 

I hear you about using a more interesting use case in the guide. I reworked it to use an english-to-german translation app written in the blocks api. It then shows how to load that app into another app so that users can generate text in both english and german with gpt.

Let me know if I should address anything else prior to merging!

abidlabs (approved):
That's all I got! Thanks for making the change @freddyaboulton!

Issue #1801, opened 2022-07-16
Version 3.0.25 incompatibility problems in Google Colab


rafaeldepablo:
### Describe the bug

After installing the version (3.0.25) in a Google colab notebook, the import stop to working returning the following error: "ImportError: cannot import name 'TypedDict' from 'typing' (/usr/lib/python3.7/typing.py)"

Installing older version 3.0.0. everything works normally again

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

in a google colab

!pip install --quiet gradio
import gradio as gr

### Screenshot

![Captura de pantalla de 2022-07-16 07-10-00](https://user-images.githubusercontent.com/57199787/179340402-4af94193-d88d-43ae-9249-e002b4fff81d.png)

### Logs

```shell
ImportError                               Traceback (most recent call last)
<ipython-input-2-43eca54f7d45> in <module>()
----> 1 import gradio as gr

1 frames
/usr/local/lib/python3.7/dist-packages/gradio/components.py in <module>()
     17 from copy import deepcopy
     18 from types import ModuleType
---> 19 from typing import Any, Callable, Dict, List, Optional, Tuple, TypedDict
     20 
     21 import matplotlib.figure

ImportError: cannot import name 'TypedDict' from 'typing' (/usr/lib/python3.7/typing.py)

---------------------------------------------------------------------------
NOTE: If your import is failing due to a missing package, you can
manually install dependencies using either !pip or !apt.

To view examples of installing some common dependencies, click the
"Open Examples" button below.
---------------------------------------------------------------------------
```

### System Info

```shell
problem 3.0.25
```

### Severity

annoying

abidlabs:
Thanks for bringing this to our attention so quickly @rafaeldepablo, this should be fixed now as of `gradio==3.0.26`!

PR #1803, opened 2022-07-16
Fix missing TypedDict on python 3.7

Ryan-Qiyu-Jiang:
Attempt to import TypedDict from typing_extensions
if running on python 3.7.
Remove unused importlib.metadata.

# Description
Fixes import issue on 3.7 as TypedDict was introduced in python 3.8
Should get gradio working on default colab again.
This change attempts to import TypedDict from typing_extensions when required so 3.7 users can add the new dependency as required.
This doesn't require 3.8+ users to install typing_extensions.

@pngwn @abidlabs 

Closes: #1801

### DIFF
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -12,11 +12,17 @@ import operator
 import os
 import pathlib
 import shutil
+import sys
 import tempfile
 import warnings
 from copy import deepcopy
 from types import ModuleType
-from typing import Any, Callable, Dict, List, Optional, Tuple, TypedDict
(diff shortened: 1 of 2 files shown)


abidlabs:
Thank you @Ryan-Qiyu-Jiang! This looks good but I think we also need to add typing-extension to our `requirements.txt`, right?

abidlabs:
I can do that on my end

abidlabs:
I can do that on my end

Issue #1802, opened 2022-07-16
cant import gradio


Sujith-Sunilan:
### Describe the bug

import gradio as gr 
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
[<ipython-input-14-fd1c832c29f5>](https://localhost:8080/#) in <module>()
----> 1 import gradio

1 frames
[/usr/local/lib/python3.7/dist-packages/gradio/components.py](https://localhost:8080/#) in <module>()
     17 from copy import deepcopy
     18 from types import ModuleType
---> 19 from typing import Any, Callable, Dict, List, Optional, Tuple, TypedDict
     20 
     21 import matplotlib.figure

ImportError: cannot import name 'TypedDict' from 'typing' (/usr/lib/python3.7/typing.py)

---------------------------------------------------------------------------
NOTE: If your import is failing due to a missing package, you can
manually install dependencies using either !pip or !apt.

To view examples of installing some common dependencies, click the
"Open Examples" button below.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

can't import gradio

### Logs

```shell
can't import gradio
```

### System Info

```shell
NameError                                 Traceback (most recent call last)
<ipython-input-16-8f51e41751d4> in <module>()
----> 1 gradio.__version__

NameError: name 'gradio' is not defined
```

### Severity

annoying

abidlabs:
Thanks for bringing this to our attention so quickly @Sujith-Sunilan, this should be fixed now as of `gradio==3.0.26`!

PR #1805, opened 2022-07-16
installs typing extensions, releases new package

abidlabs:
Hotfix for #1801, #1802


freddyaboulton (approved):
Thanks @abidlabs for jumping on this so quickly !

I'm able to import gr.Examples in a colab notebook:

image


We should be good to release. 

FYI, I think it may be best to do 

```python
import typing TYPE_CHECKING

if TYPE_CHECKING:
     from typing TypedDict
```

Like we do in other parts of the library, like `blocks.py`. That way we don't need the extra dependency. We can do that next week!

abidlabs:
Thank you so much for the suggestion @freddyaboulton -- agreed! Will file an issue to track this and fix next week.

Issue #1676, opened 2022-06-30
Add a flagging callback to save json files to a hugging face dataset


freddyaboulton:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
As a gradio user, I would like to save flagged instances as separate json files to a hugging face dataset.

@chrisemezue wrote this awesome [space](chrisjay/mnist-adversarial) but he had to use his own flagging function to save to a hf dataset to json because our callback only saves to a csv file. He needed to save to json so that it would work better with concurrent usage of his app.

As an aside, he also needed to do some custom work in the flagging callback on his app, but I think saving to json is a common enough need that should merit it's own class in the core library especially considering that it can help with multiple users.

**Describe the solution you'd like**  
A different flagging callback or an extension of the existing `HuggingFaceDatasetSaver` to save to json. 

**Additional context**

chrisemezue:
With pleasurere!

PR #1821, opened 2022-07-18
Add a flagging callback to save json files to a hugging face dataset

chrisemezue:
# Description

Based on issue #1676 I have created the `HuggingFaceDatasetJSONSaver` class which saves the files as JSONL.

Specifically, for each flagged sample:
- I create a unique ID (a hash of random numbers and strings) and create a folder with the name of the ID. In the code I call the new folder `folder_name`.
- Save the files (images, audio) inside `folder_name`
- Save the other details (output, numbers, etc) in a `metadata.jsonl` file inside the `folder_name` folder.

### Advantages of this:
- The major advantage is that we bypass the need to read and write to one CSV. Where the advantage of this is useful is if there are three users on their devices simultaneously flagging a sample. With CSV there would be an error because there can't be more than one simultaneous edit to a CSV file. But this way I propose enables parallel flagging.

- no additional dependencies are required. I tried my best to make sure I leverage functions from the original `HuggingFaceDatasetSaver`.  

Closes: #1676 

# Checklist:

- [ X] I have performed a self-review of my own code


abidlabs:
Hi @chrisemezue,

This is very cool, two quick high-level questions:

1) Can you run the formatter on your code? That way, the CI won't complain about the formatting. The easiest way is to run this script:

```
bash scripts/format_backend.sh
```

2) Just to confirm, is this way of saving data into a HuggingFace Dataset (having folders for each sample) compatible with the Dataset previewer on the Hub?

osanseviero:
I think the title of this PR is a bit misleading. `HuggingFaceDatasetSaver` will still not work in parallel afaik. For me it's a bit confusing if this is saving a `data.csv` or a json based on the name (`HuggingFaceDatasetJSONSaver`). If it's a csv, should we directly update `HuggingFaceDatasetJSONSaver`?

chrisemezue:
> I think the title of this PR is a bit misleading. `HuggingFaceDatasetSaver` will still not work in parallel afaik. For me it's a bit confusing if this is saving a `data.csv` or a json based on the name (`HuggingFaceDatasetJSONSaver`). I see `log_file` uses `csv` but that does not seem to be used anywhere. If it's a csv, should we directly update `HuggingFaceDatasetJSONSaver`?

Thanks @osanseviero for your feedback 
- I will remove the `self.log_file` and other redundant variables not used.
-  the class `HuggingFaceDatasetJSONSaver` is just saving the flagged samples to a jsonl format instead of csv. Is there a better name you suggest?
- I changed the name of PR to the name of [its issue](https://github.com/gradio-app/gradio/issues/1676). If you have a better easier to understand suggestion I will love to hear.

osanseviero:
Hey @chrisemezue, thank you! I was a bit confused by the mentions of csvs, but now that you mention it's a `json` then it's great! Thanks!

osanseviero:
Please let us know whenever this is ready for review :)

chrisemezue:
@osanseviero I am done now. Ready for review.

gradio/flagging.py
```diff
@@ -5,6 +5,8 @@
 import io
 import json
 import os
+import random
```
osanseviero, line 8: There are formatting errors. See how at bottom of https://github.com/gradio-app/gradio/pull/1821 it shows some automatic errors linking to https://app.circleci.com/pipelines/github/gradio-app/gradio/4649/workflows/70a7b34f-f35d-424b-9220-f01aa924f77f/jobs/5281, which  says 

```
Imports are incorrectly sorted and/or formatted.
```

Please follow the instructions of https://github.com/gradio-app/gradio/blob/main/CONTRIBUTING.md#submitting-prs to make sure things are formatted properly

gradio/flagging.py
```diff
@@ -344,3 +346,170 @@ def flag(
         self.repo.push_to_hub(commit_message="Flagged sample #{}".format(line_count))
 
         return line_count
+
+
+class HuggingFaceDatasetJSONSaver(FlaggingCallback):
+    """
+    A FlaggingCallback that saves flagged data to a HuggingFace dataset in JSONL format.
```
osanseviero, line 353: Maybe add a line explaining why this is useful (e.g. saving the metadata json is useful for images/audio/other formats + it can be loaded out of the box with `datasets.load_dataset()`)

gradio/flagging.py
```diff
@@ -344,3 +346,170 @@ def flag(
...
+        self.hf_foken = hf_foken
+        self.dataset_name = dataset_name
+        self.organization_name = organization
+        self.dataset_private = private
+        self.verbose = verbose
+
+    def setup(self, components: List[Component], flagging_dir: str):
+        """
+        Params:
+        flagging_dir (str): local directory where the dataset is cloned,
```
osanseviero, line 386: Should a docstring for `components` also be added?

gradio/flagging.py
```diff
@@ -344,3 +346,170 @@ def flag(
...
+        self.infos_file = os.path.join(self.dataset_dir, "dataset_infos.json")
+
+    def flag(
+        self,
+        flag_data: List[Any],
+        flag_option: Optional[str] = None,
+        flag_index: Optional[int] = None,
+        username: Optional[str] = None,
+    ) -> int:
+        self.repo.git_pull(lfs=True)
```
osanseviero, line 423: I would also add `rebase=True` to be safer when multiple people are flagging at same time.

osanseviero, line 423: I don't think `lfs=True` is really needed. As per https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/repository.py#L1025-L1029, this is only useful if the lfs files were not pulled in first clone, but they should be there.

gradio/flagging.py
```diff
@@ -344,3 +346,170 @@ def flag(
...
+
+        # Generate unique folder for the flagged sample
+        unique_name = self.get_unique_name()  # unique name for folder
+        folder_name = os.path.join(
+            self.dataset_dir, unique_name
+        )  # unique folder for specific example
+        os.makedirs(folder_name, exist_ok=True)
```
osanseviero, line 430: If the folder is unique, should `exist_ok=True` really be used?

gradio/flagging.py
```diff
@@ -344,3 +346,170 @@ def flag(
...
+        if is_new:
+            json.dump(infos, open(self.infos_file, "w"))
+
+        self.repo.push_to_hub(commit_message="Flagged sample {}".format(unique_name))
+        return unique_name
+
+    def get_unique_name(self):
+        return "".join(
+            [random.choice(string.ascii_letters + string.digits) for n in range(32)]
+        )
```
osanseviero, line 511: @abidlabs anything we could do here to make this really unique?

gradio/flagging.py
```diff
@@ -344,3 +346,170 @@ def flag(
...
+
+        infos = {"flagged": {"features": {}}}
+
+        # File previews for certain input and output types
+        file_preview_types = {
+            gr.inputs.Audio: "Audio",
+            gr.outputs.Audio: "Audio",
+            gr.inputs.Image: "Image",
+            gr.outputs.Image: "Image",
+        }
```
osanseviero, line 443: I **think** you might be able to do this.

```suggestion
        file_preview_types = {
            gr.Audio: "Audio",
            gr.Image: "Image",
        }
```

osanseviero (commented):
Thanks a lot for this! This is very cool! I left some minor comments 🤗 

I would love to see an example output of using this flagging callback (a small dataset, since https://huggingface.co/datasets/chrisjay/crowd-speech-africa has too many files and it does not load :()).

cc @lhoestq

gradio/flagging.py
lhoestq: You can use `uuid` instead of generating this name randomly yourself btw

gradio/flagging.py
```diff
@@ -344,3 +346,170 @@ def flag(
...
+        unique_name = self.get_unique_name()  # unique name for folder
+        folder_name = os.path.join(
+            self.dataset_dir, unique_name
+        )  # unique folder for specific example
+        os.makedirs(folder_name, exist_ok=True)
+
+        # Now uses the existence of `dataset_infos.json` to determine if new
+        is_new = not os.path.exists(self.infos_file)
+
+        infos = {"flagged": {"features": {}}}
```
lhoestq, line 435: Do yo really need the dataset_infos.json file ?

gradio/flagging.py
chrisemezue: thanks. done

gradio/flagging.py
chrisemezue: i just put it as safeguard. but you are right, it is redundant. will remove it.

gradio/flagging.py
chrisemezue: yh makes sense. Thanks for the reminder

gradio/flagging.py
chrisemezue: is this better?

gradio/flagging.py
chrisemezue: @lhoestq I think yes
- I think the `dataset_infos.json` file is used by the datasets viewer.
- I do not know of a better way to evaluate `is_new` (besides using the existence of a file that can only be created once during the creation of the dataset instance). The previous HF DatasetSaver [here](https://github.com/gradio-app/gradio/blob/main/gradio/flagging.py#L309) uses the existence of the `data.csv` file to determine `is_new`.

gradio/flagging.py
chrisemezue: I remember coming across UUID while looking for random ids on Python. I saw this drawback about them and that was made me not to use them ([source](https://www.geeksforgeeks.org/generating-random-ids-using-uuid-python/)):
>Drawback :
>This way includes the used of MAC address of computer, and hence can compromise the privacy, even though it provides uniqueness.


is this something of concern or not @lhoestq ?

gradio/flagging.py
chrisemezue: all right. I have added it.

chrisemezue:
> 

@osanseviero [here](https://huggingface.co/datasets/chrisjay/test-mnist-2) is an example of a small dataset with this flagging callback.

gradio/flagging.py
chrisemezue: for some reason I am having issues running the `bash scripts/run_circleci.sh` command. It's a problem with my Docker. I am trying to work on it.
- In the meantime, is it possible to understand what import is incorrectly formatted in my case. I went to [the traceback](https://app.circleci.com/pipelines/github/gradio-app/gradio/4649/workflows/70a7b34f-f35d-424b-9220-f01aa924f77f/jobs/5281) but it does not really help much in showing where the error is.

freddyaboulton: @chrisemezue `bash scripts/ format_backend.sh` will automatically sort the imports for you!

chrisemezue: oh..I understand now. Thank you. Done

gradio/flagging.py
abidlabs: You can use `uuid.uuid4()` to avoid this issue. As mentioned later in the article, "This function guarantees the random no. and doesn’t compromise with privacy."

gradio/flagging.py
```diff
@@ -344,3 +346,175 @@ def flag(
...
+        components List[Component]: list of components for flagging
+        flagging_dir (str): local directory where the dataset is cloned,
+        updated, and pushed from.
+        """
+        try:
+            import huggingface_hub
+        except (ImportError, ModuleNotFoundError):
+            raise ImportError(
+                "Package `huggingface_hub` not found is needed "
+                "for HuggingFaceDatasetSaver. Try 'pip install huggingface_hub'."
```
abidlabs, line 401:
```suggestion
                "for HuggingFaceDatasetJSONSaver. Try 'pip install huggingface_hub'."
```

gradio/flagging.py
abidlabs: Yes, I believe dataset_infos.json file is used by the datasets viewer, so this is good to have!

gradio/flagging.py
```diff
@@ -277,7 +279,7 @@ def flag(
         flag_index: Optional[int] = None,
         username: Optional[str] = None,
     ) -> int:
-        self.repo.git_pull(lfs=True)
+        self.repo.git_pull()
```
abidlabs, line 282: Why did we get rid of `lfs=True`?

gradio/flagging.py
```diff
@@ -344,3 +346,175 @@ def flag(
...
+        )  # unique folder for specific example
+        os.makedirs(folder_name)
+
+        # Now uses the existence of `dataset_infos.json` to determine if new
+        is_new = not os.path.exists(self.infos_file)
+
+        infos = {"flagged": {"features": {}}}
+
+        # File previews for certain input and output types
+        file_preview_types = {
```
abidlabs, line 445: This is all good. But I would move the logic that is repeated between `HuggingFaceDatasetSaver` and `HuggingFaceDatasetJSONSaver` to a separate function that is shared between both to avoid duplicating code

gradio/flagging.py
```diff
@@ -344,3 +346,175 @@ def flag(
...
+        self.infos_file = os.path.join(self.dataset_dir, "dataset_infos.json")
+
+    def flag(
+        self,
+        flag_data: List[Any],
+        flag_option: Optional[str] = None,
+        flag_index: Optional[int] = None,
+        username: Optional[str] = None,
+    ) -> int:
+        self.repo.git_pull(rebase=True)
```
abidlabs, line 430: Why is `rebase=True` here? Should it the case in the `HuggingFaceDatasetSaver.flag()` method as well?

gradio/flagging.py
```diff
@@ -344,3 +346,175 @@ def flag(
...
+
+            csv_data.append(filepath)
+        headers.append("flag")
+        csv_data.append(flag_option if flag_option is not None else "")
+
+        # Creates metadata dict from row data and dumps it
+        metadata_dict = {
+            header: _csv_data for header, _csv_data in zip(headers, csv_data)
+        }
+        self.dump_json(metadata_dict, os.path.join(folder_name, "metadata.jsonl"))
```
abidlabs, line 505: Is it standard to dump to a `.jsonl` file instead of a `.json` file?

abidlabs:
Hi @chrisemezue this looks really good! I left some suggestions / clarification questions in the PR, but once these are addressed, we should be good to merge

abidlabs:
Pushed some changes which should fix the tests. As discussed over Slack, we just have a couple of minor fixes, and then we should be good to merge!

abidlabs (approved):

abidlabs:
Thanks so much @chrisemezue for making the PR and addressing the suggestions! And thanks all for reviewing. 

LGTM -- will merge in after the tests run

osanseviero (approved):
This looks good! Thanks for working on this!

abidlabs:
I'll resolve the conflicts and fix the last few suggestions you made @osanseviero, so that we can get this merged in. 

Thanks a bunch @chrisemezue!

PR #1822, opened 2022-07-19
Some cleanups to the docs page

abidlabs:
Small changes to the wording in various classes to make the documentation easier to follow

### DIFF
--- a/website/homepage/src/docs/obj_doc_template.html
+++ b/website/homepage/src/docs/obj_doc_template.html
@@ -38,7 +38,7 @@
   {% else %}
   <div class="codeblock"><pre><code class="lang-python">{{ parent }}.<span>{{ obj["name"] }}(</span><!--
 -->{% for param in obj["parameters"] %}<!--
-  -->{% if "positional" in param["kind"] and "default" not in param %}<!--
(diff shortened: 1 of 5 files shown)


website/homepage/src/docs/obj_doc_template.html
```diff
@@ -38,7 +38,7 @@
   {% else %}
   <div class="codeblock"><pre><code class="lang-python">{{ parent }}.<span>{{ obj["name"] }}(</span><!--
 -->{% for param in obj["parameters"] %}<!--
-  -->{% if "positional" in param["kind"] and "default" not in param %}<!--
+  -->{% if "positional" in param["kind"] and "default" not in param and param["name"] != "self" %}<!--
```
abidlabs, line 41: Removes "self" from appearing in the list of parameters for every function in the docs page

abidlabs: Ah yeah that looks weird. I think including all of the parameters is too much (especially for classes like `Interface`). How about we always include the first parameter (even if it has a default argument) + include all parameters that are required (aka don't have defaults)

abidlabs: Will discuss that separately

Issue #1323, opened 2022-05-18
The way to start or stop the Image component for wabcam streaming


JefferyChiang:
- [X] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.** 

related to #220 
I tried the Image streaming function for webcam, and it can performed real time inference correctly.
I'm wondering is there a proper way to trigger start or stop the real time inference?

**Describe the solution you'd like**  

Though I can use another Tab to stop the inference thread, I think it will be more easy to have a button to trigger.

**Additional context**  
Here is my code for testing the Image stream function.

```
import gradio as gr

def image_mod(image):
    return image.rotate(45)
    
block = gr.Blocks()
with block:
    with gr.Tabs():
        with gr.TabItem("Train"):
            train = gr.Button("Train")
        with gr.TabItem("Inference"):
            with gr.Row():
                input_image = gr.Image(type='pil',label="Input Image", source="webcam",
                                 streaming=True)
                output_image = gr.Image(label="Output Image")
                input_image.change(image_mod,
                           inputs=input_image,
                           outputs=output_image)
block.launch(show_error=True, inbrowser=True)
```

abidlabs:
I'll close as this is a duplicate of #3642 and that issue has a bit more activity. We're working on this!

Issue #1332, opened 2022-05-18
Audio Streaming broken


mahnerak:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

It seems the new Audio component does not support audio streaming in live mode. The stream is sent only after you click stop.
But it does work on `gradio==2.9`.

### Reproduction

Install `gradio>3.0`.
Run any realtime audio app (say, https://huggingface.co/spaces/YSU/aspram-realtime/blob/main/app.py). It will send the data after you click stop instead of continuously sending chunks.

### System Info

```shell
gradio==3.0
```

### Severity

critical

tango4j:
I am having the same issue (audio streaming is not activated when it is running on local machine). Also, gradio.inputs.Audio() does not have argument "streaming" unlike described in the docs.

abidlabs:
Thanks @mahnerak and @tango4j for pointing this out! We are looking into this

aliabid94:
ahh I see, can you guys try using `gr.Audio(source="microphone", streaming=True)` (not `gr.inputs.Audio`) for the input, and let me know if that works? Will figure out the right way to bring it back to the old `gr.inputs.Audio` in the meantime.

mahnerak:
The problem is there even when you're not using streaming=True.
You can use just live mode on which will rerun gradio pipeline every T secs.

Starting 3.0 you should click stop to make gradio send the request to the backend:

https://user-images.githubusercontent.com/1367529/169415102-cc1200b2-d742-43f1-b0cf-cb968957c4b8.mov


In 2.9 it works as expected. You can note pulsating icon of processing while I'm still recording the audio:

https://user-images.githubusercontent.com/1367529/169415125-446b2bb6-781b-4bf0-b9c4-4b9660f57446.mov

BTW, You can access exactly the same gradio app here:
https://huggingface.co/spaces/YSU/aspram-realtime
I've changed gradio version to 2.9 (in README, yaml on the top) to make it work.


Hope this helps.

aliabid94:
Fixing now, thanks for the thorough documentation!

abidlabs:
Hi @mahnerak and @tango4j, this should be fixed now! We'll release `gradio==3.1` soon, and you can try the following example as a simple streaming speech to text example:

```py
from transformers import pipeline
import gradio as gr
import time

p = pipeline("automatic-speech-recognition")

def transcribe(audio, state=""):
    time.sleep(2)
    text = p(audio)["text"]
[...]

PR #1828, opened 2022-07-19
Fix streaming issues

aliabid94:
Hide status loading animation on stream, fix echo in stream on last audio chunk.

Check demo/stream_audio for example.

Still not 100% if audio streaming is being done the correct way (basically stopping and starting the MediaRecorder at 500ms intervals) but seems to work.


abidlabs:
Does this close #1323, #1332, and #1494?

freddyaboulton:
Thanks @aliabid94 ! 

Confirmed this fixes the audio/image streaming example. The only thing is that I noticed a very faint flicker in the audio component during flicker (you can kind of see it at second 7 in this gif) but I think that's ok

stream_audio

abidlabs:
Yup also tested this with the examples from our real time ASR guide:

```py
import gradio as gr
import time

p = pipeline("automatic-speech-recognition")

def transcribe(audio, state=""):
    time.sleep(2)
    text = p(audio)["text"]
    state += text + " "
    return state, state

gr.Interface(
    fn=transcribe, 
    inputs=[
        gr.Audio(source="microphone", type="filepath", streaming=True), 
        "state"
    ],
    outputs=[
        "textbox",
        "state"
    ],
    live=True).launch()
``` 

and the examples work great! I think we can fix the flicker & clean up any other issues we fine in a future PR, but for the sake of releasing 3.1, let's go ahead and merge this in.

PR #1834, opened 2022-07-19
Adding flagging callbacks to the docs

### LINKED ISSUE #1673: Add missing classes and functions to docs page

As a gradio user, I can't find all the functionality gradio has to offer documented in the Docs page. 

For example, I can't find `gr.update` (https://github.com/gradio-app/gradio/issues/1647), the different flagging callbacks, and some components like `Interpretation` on the docs page. 

Some of these concepts, like the flagging callbacks, are covered in the guides but I think they should also be in the docs page. This will make it easier for developers and users to find all the public apis in one place as opposed to having to click through different guides and skim through the text to the part they need. 

Document all user-facing functionality on the Docs page. We may need to reorganize the Docs page too, like adding more subheadings beyond `Interface` and `Components`.

abidlabs:
* Partially closes #1673

Here's what the flagging section looks like on the website:

image

* Also moves example usage above the parameter table


freddyaboulton (approved):
Thank you @abidlabs !! Looks good to me

abidlabs:
Thanks @freddyaboulton for the review! Will look into these points

abidlabs:
Thanks for the review @freddyaboulton! Fixed the points above, so that we no longer need to have a `show_parameter_table` flag and now we simply loop over the `FlaggingCallback()` classes that have a `document()` decorator.

Was getting weird behavior in the docs when deleting `__init__`s so I left them in there. We can look into that in the future but low priority imo.

freddyaboulton:
Thank you @abidlabs ! Very cool

PR #1840, opened 2022-07-20
Fix Underscore Bug in Demo Names

aliabd:
When a demo in a guide has two underscores in the title (for example `hello_world_2.py` in `getting_started.md`) the markdown to HTML generator converts what's between the underscores into italic, which breaks the demo logic. This is causing demos to go missing on guides. 

This was fixed but lost in one of the big website PRs. This PR brings back the fix which is just to replace the underscores with a special token during the HTML rendering process.


abidlabs:
Can confirm that the web component scripts show up correctly, e.g.:

```html
<gradio-app src='/demo/hello_world_3' />
```

abidlabs (approved):

PR #1843, opened 2022-07-20
Fix the types

pngwn:
# Description

Fixes the types.

Sadly I did have to change some code, so it would be good to test with some more complex blocks and something that uses examples. I also needed to rewrite how we load `Bokeh`, so if we could test out whatever Bokeh examples we have that would be good.

I have tested these things but would appreciate a second set of eyes.


abidlabs:
This is great, thanks Pete! I couldn't look at the files directly, but here are the demos that I tested:

* sepia_filter 👍 
* image_classifier_2 👍 
* calculator_live 👍 
* main_note -- includes examples👍
* sentence_builder -- includes examples 👍
* titanic_survival -- includes examples 👍
* outbreak_forecast ❌ 

The `outbreak_forecast/run.py` demo includes matplotlib, plotly, and bokeh plots. **None** of them are working for me. I get the following javascript error:

```
Uncaught (in promise) ReferenceError: theme is not defined
    at Object.c (index.0091c959.js:181:23)
    at Object.p (index.0091c959.js:289:15)
    at update (index.c616cb56.js:610:36)
    at flush (index.c616cb56.js:577:13)
```

Just to confirm, I tested the `stock_forecast/run.py` demo and it's also throwing the same error.

freddyaboulton:
I get the same behavior as @abidlabs in `stock_forecast`. I have to set `cache_examples=False` to even run `outbreak_forecast`. 

Additionally I tested:

*blocks_plug 👍 
* image_classifier_interpretation 👍 
* blocks_update 👍 
* blocks_calculator 👍 
* color_picker 👍 
* digit_classifier 👍 
* zip_two_files 👍 
* zip_to_json 👍 
* text_analysis 👍 
* hello_login  ❌  I see the following upon opening the app

```
index.781fb532.js:6826 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading '$set')
    at instance$2 (index.781fb532.js:6826:31)
    at init (index.781fb532.js:1066:11)
    at new Login (index.781fb532.js:6840:3)
    at mount_app (index.781fb532.js:9903:5)
    at GradioApp.connectedCallback (index.781fb532.js:9968:9)
```

pngwn:
I've fixed the issue with `outbreak_forecast` but I can't reproduce the issue with `hello_login`. However, it looks like auth is broken again? I'm not getting any login screen at all (it goes straight to the app) and the config does not contain `detail`, `auth_required` or `auth_message` even though at least two of these should be present in the config for following demo:

```py

if __name__ == "__main__":
    demo.launch(
        auth=lambda u, p: user_db.get(u) == p,
        auth_message="This is a welcome message")
```

I also cannot reproduce @freddyaboulton's separate issue with `outbreak_forecast`.

pngwn:
I think I'm not seeing this because I am not seeing the login screen, maybe this is specific to me though. I'll see why that might be.

pngwn:
Oka, never mind, I can see the login now. It must have been a caching issue.

Just pushed a fix for the issue you were seeing @freddyaboulton.

abidlabs:
Retested several demos, including both `hello_login` and `outbreak_forecast` and everything works! (Besides bokeh plots, but that's a separate issue). Good stuff @pngwn!

abidlabs (approved):

PR #1845, opened 2022-07-20
async-interface-fix

FarukOzderim:
- support async in interface with workarounds
- make mix.Parallel async and concurrent

# Description
Challenged going through the async chain, was going to use utils.synchronize_async but realized I don't need that a lot, because the chain's endpoint arrived at a function that I can make async :D

This is not a clean solution, and I would like to refactor Interface and Interpretation cleanly before doing this. However it seems this is blocking Ray team, so I wanted to unblock this part. There'll be a lot of failing tests I assume, will fix them tomorrow if the PR is good for you guys.

Btw the current code structure definitely needs refactoring, the process was a bit mind-blowing. Interpretation and mix was recursively tied to Interface :D

Testing different scenarios with Interface, Mix.parallel and Interpretation would be nice, as they were quite affected from this PR.


abidlabs:
Awesome, will review later today!

abidlabs:
@FarukOzderim I think we should test a lot of different scenarios/paths to make sure everything works:

- [ ] `Interface` with regular Python function that returns a single value
- [ ] `Interface` with regular Python function that returns multiple values
- [ ] `Interface` with coroutine that returns single value
- [ ] `Interface` with coroutine that returns multiple values
- [ ] `Interface` with function with interpretation
- [ ] `Interface` with coroutine with interpretation
- [ ] `Parallel` with two `Interfaces` with functions
- [ ] `Parallel` with two `Interfaces` with coroutines
- [ ] `Parallel` with two `Interfaces`, one with function, one with coroutine

I started testing but ran into an issue:

FarukOzderim:
It was a bit hastily put, was missing a few parts, fixed those and tested the scenarios above. Mix and Interface works like a charm but Interpretation does not work, can you take a look on Interpretation @abidlabs?

Example tests:
**1**
```

import asyncio
import gradio as gr


async def sl(x):
    await asyncio.sleep(5)
    return "hello" + x + "hello", "hello" + x + "hello"


def sl_sync(x):
    return "hello" + x + "hello", "hello" + x + "hello"


gr.Interface(sl, "text", ["text", "text"], interpretation="default").launch()
```

**2**
```
import gradio as gr
import asyncio


async def sl(x):
    await asyncio.sleep(5)
    return "hello" + x + "hello"


interfaces = [gr.Interface.load("spaces/onnx/GPT-2"), gr.Interface(sl, "text", "text")]
gr.mix.Parallel(*interfaces, title="example-title", description="example-description").launch()

```

abidlabs:
Ok I think I have a fix for interpretation. Just pushed. Here's my code to test:

```py
import gradio as gr

async def sl(x, y):
    return max(len(y) for y in x.split(" "))

gr.Interface(sl, ["text", "text"], "label", interpretation="default").launch()
```

Haven't tested extensively, so please do confirm that interpretation is actually fixed!

freddyaboulton:
@FarukOzderim Awesome PR! The code looks good to me but I'm still reviewing. One thing that stood out to me is that I think `Series` should support a mix of sync and async function as well right?

This does not produce the correct behavior

```python
import gradio as gr

async def say_hello(name):
    return f"Hello {name}!"

get_name = gr.Interface(lambda name: name, inputs="textbox", outputs="textbox")
prepend_hello = gr.Interface(say_hello, inputs="textbox", outputs="textbox")
append_nice = gr.Interface(lambda greeting: f"{greeting} Nice to meet you!",
                           inputs="textbox", outputs=gr.Textbox(label="Greeting"))
demo = gr.Series(get_name, prepend_hello, append_nice)

if __name__ == "__main__":
    demo.launch()
```

image


Which I think is not what users would expect given that the prepend_hello interface works fine on its own.

FarukOzderim:
@freddyaboulton good catch! Will revisit, thx!

FarukOzderim:
1. Fixed the bug in series
2. Fixed tests other than in test_components.py

What should we do with the tests in test_components.py(there are a lot of tests over there)? Comment them out until we refactor Interface, or fix them now?
I won't be able to take a look at them soon. Would you like to take a look @abidlabs or @freddyaboulton?

freddyaboulton:
@FarukOzderim Thanks for the fixes! Can you file an issue so that we don't forget to fix the test? Looks like Series with external interfaces broke?

FarukOzderim:
> @FarukOzderim Thanks for the fixes! Can you file an issue so that we don't forget to fix the test? Looks like Series with external interfaces broke?

I think it is a flake, let's file an issue if it persists 👍

abidlabs:
I looked into the tests, and they are failing pretty consistently for me. Let me file an issue and look more deeply into the reason why

Issue #1809, opened 2022-07-17
The examples run only once, then they begin to generate error


kmkarakaya:
### Describe the bug

When I deploy an app on Spaces, the examples run only once, then if I click one of them they begin to generate error. 

After building an app on spaces, each example run perfectly ONLY for ONCE. When I click any of them for the second or more times on the screen I only see ERROR message.

However, I can enter a new prompt manually and it works fine.

If I re-build the app, again the examples run only ONCE. 

I checked the log but there seems no problem.

As far as I understood, **there is a problem to fetch the cached example and its result.**

Thank you in advanced

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

kmkarakaya/Auto_Review_Generation_in_Turkish

### Screenshot

![chrome_unlSniKJN9](https://user-images.githubusercontent.com/41159849/179402232-0c7b3591-71f6-4e63-b43b-b7c3a12c9717.gif)

### Logs

```shell
2022-07-17 13:40:02.394081: W tensorflow/stream_executor/platform/default/dso_loader.cc:64] Could not load dynamic library 'libcudart.so.11.0'; dlerror: libcudart.so.11.0: cannot open shared object file: No such file or directory
2022-07-17 13:40:02.394233: I tensorflow/stream_executor/cuda/cudart_stub.cc:29] Ignore above cudart dlerror if you do not have a GPU set up on your machine.
2022-07-17 13:40:03.758419: W tensorflow/stream_executor/platform/default/dso_loader.cc:64] Could not load dynamic library 'libcuda.so.1'; dlerror: libcuda.so.1: cannot open shared object file: No such file or directory
2022-07-17 13:40:03.758556: W tensorflow/stream_executor/cuda/cuda_driver.cc:269] failed call to cuInit: UNKNOWN ERROR (303)
2022-07-17 13:40:03.758695: I tensorflow/stream_executor/cuda/cuda_diagnostics.cc:156] kernel driver does not appear to be running on this host (s-kmkarakaya-auto--review--generation--in--turkish-ece09-bmmkbs): /proc/driver/nvidia/version does not exist
2022-07-17 13:40:03.759272: I tensorflow/core/platform/cpu_feature_guard.cc:193] This TensorFlow binary is optimized with oneAPI Deep Neural Network Library (oneDNN) to use the following CPU instructions in performance-critical operations:  AVX2 AVX512F FMA
To enable them in other operations, rebuild TensorFlow with the appropriate compiler flags.

[...]

PR #1850, opened 2022-07-21
Fixes cached examples

abidlabs:
Fixes: #1809


gradio/examples.py
```diff
@@ -153,9 +153,9 @@ def __init__(
             self.cache_interface_examples()
 
         def load_example(example_id):
-            processed_example = self.processed_examples[example_id]
+            processed_input = self.processed_examples[example_id]
             if cache_examples:
-                processed_example += self.load_from_cache(example_id)
```
abidlabs, line 158: The cached output was being appended again and again to the input, causing the bug

gradio/flagging.py
```diff
@@ -163,7 +163,7 @@ def replace_flag_at_index(file_content):
         if self.encryption_key:
             output = io.StringIO()
             if not is_new:
-                with open(log_filepath, "rb") as csvfile:
+                with open(log_filepath, "rb", encoding="utf-8") as csvfile:
```
abidlabs, line 166: Separate problem (on Windows) was that the special characters in Turkish were not being encoded correctly

aliabid94 (changes requested):
I got this error when running the demo/calculator/run.py and clicking on any example:
```
File "/Users/aliabid/projects/gradio/gradio/examples.py", line 159, in load_example
    if len(processed_example) == 1:
UnboundLocalError: local variable 'processed_example' referenced before assignment
```

freddyaboulton (commented):
Thanks @abidlabs ! Looks good to me apart from the bug pointed out by @aliabid94 . 

Once that's fixed I think this is good to merge. I tested `image_classifier`, `gpt_j`, and the original space mentioned in the issue and they work with `cache_examples=True`

abidlabs:
Yikes good catch @aliabid94. Will fix and retest

abidlabs:
Ok fixed the issue @aliabid94 and @freddyaboulton. If you get a chance to make another pass, would be great. Otherwise, will go ahead and merge in shortly

freddyaboulton (approved):
Thank you @abidlabs !

abidlabs:
Thanks @freddyaboulton!

PR #1865, opened 2022-07-22
Fixes the twitter image preview for Guides 

abidlabs:
Don't know of any way to test except to deploy :D

### DIFF
--- a/website/homepage/src/guides/template.html
+++ b/website/homepage/src/guides/template.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
-    {% with title=pretty_name, url="https://gradio.app/" + name, image="/assets/img/meta-image.png", description="A Step-by-Step Gradio Tutorial" %}
+    {% with title=pretty_name, url="https://gradio.app/" + name, image="https://www.gradio.app/assets/img/meta-image.png", description="A Step-by-Step Gradio Tutorial" %}
       {% include "templates/meta.html" %}
     {% endwith %}
     <link rel="stylesheet" href="/style.css">
--- a/website/homepage/src/templates/meta.html
+++ b/website/homepage/src/templates/meta.html
@@ -12,7 +12,7 @@
 <meta name="twitter:creator" content="@Gradio">
 <meta name="twitter:title" content="{{ title }}">
 <meta name="twitter:description" content="{{ description }}.">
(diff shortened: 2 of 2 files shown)


website/homepage/src/guides/template.html
```diff
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
-    {% with title=pretty_name, url="https://gradio.app/" + name, image="/assets/img/meta-image.png", description="A Step-by-Step Gradio Tutorial" %}
+    {% with title=pretty_name, url="https://gradio.app/" + name, image="https://www.gradio.app/assets/img/meta-image.png", description="A Step-by-Step Gradio Tutorial" %}
```
freddyaboulton, line 4: I think we also need to modify the `meta.html` template otherwise the twitter path is going to be messed up?

```
  | <meta property="og:image" content="https://www.gradio.app/assets/img/meta-image.png" />
  | <meta name="twitter:card" content="summary_large_image">
  | <meta name="twitter:creator" content="@Gradio">
  | <meta name="twitter:title" content="Using Blocks Like Functions">
  | <meta name="twitter:description" content="A Step-by-Step Gradio Tutorial.">
  | <meta name="twitter:image" content="https://gradio.app/https://www.gradio.app/assets/img/meta-image.png">
```

I'm actually not sure what the original problem is or how to test. @aliabd Do you know? 

Is it the double-slash that gets created in the twitter meta tag in the template?

In any event, maybe just changing the template to be
```
<meta name="twitter:image" content="{{ image }}">
```

Is the right thing to do.

abidlabs: My bad, yeah you're right. 

I think Twitter actually looks at 

`<meta property="og:image" content="https://www.gradio.app/assets/img/meta-image.png" />`

not 

`  | <meta name="twitter:image" content="https://gradio.app/https://www.gradio.app/assets/img/meta-image.png">`

At least according to this: https://developer.twitter.com/en/docs/twitter-for-websites/cards/guides/getting-started

abidlabs: Will fix after lunch

abidlabs:
I think the easiest way to check would be to just deploy this. @freddyaboulton does this look all right to you?

freddyaboulton (approved):
Looks good to me! Thank you @abidlabs !

PR #1870, opened 2022-07-23
Create a ColorPicker demo

Zengyf-CVer:
# Description
I created a demo that randomly generates dark or light colors based on the ColorPicker component. The following is a demonstration of this demo:
https://huggingface.co/spaces/Zengyf-CVer/Color_Switch

# Checklist:

### DIFF
--- a/demo/color_picker/color_switch.py
+++ b/demo/color_picker/color_switch.py
@@ -1,0 +1,59 @@
+import gradio as gr
+import cv2
+import numpy as np
+import random
+
+
+# Convert decimal color to hexadecimal color
+def RGB_to_Hex(rgb):
+    color = "#"
+    for i in rgb:
+        num = int(i)
+        color += str(hex(num))[-2:].replace("x", "0").upper()
+    return color
+
+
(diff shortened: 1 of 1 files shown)


abidlabs:
Hi @Zengyf-CVer, thanks for the contribution! Happy to add the demo to the `gradio` repo, but I'd like to make some suggestions:

* First, we have each demo in a separate folder in a file called `run.py`. It's important that each demo follows this pattern so that we can load them correctly on the website. Can you move the demo to a separate folder? I would suggest naming the folder something like `color_generator` and putting your code in a `run.py` file inside of the folder
* Second, please add this demo to the `ColorPicker` component's documentation here:

https://github.com/gradio-app/gradio/blob/4ef87e03d0ab1b2606b14e0bf4bd2177b036337d/gradio/components.py#L2897

* As far as the demo itself, it looks good. I would suggest also including a Textbox output that shows the color that is generated as a hexadecimal string

Zengyf-CVer:
@abidlabs Thanks for your suggestion, I will seriously revise it.

PR #1878, opened 2022-07-25
fix link

abidlabs:
As pointed out by @AK391, on some systems such as Repl.it, the Spaces link doesn't show up correctly because of a trailing closing parenthesis. This should fix that


abidlabs:
I've also overridden it on the Gradio API Server, so it will show up correctly for all older versions of Gradio (as long as they have an Internet connection). See colab:

image

freddyaboulton (approved):

Issue #1763, opened 2022-07-11
Apply different formatting for the types in the component docstrings


freddyaboulton:
As a gradio user, I wish the types used in the `preprocessing`, `postprocessing` and `examples` sections of the docstrings were rendered differently than the surrounding text to make them stand out.

Here is what these sections currently look like:
![image](https://user-images.githubusercontent.com/41651716/178369662-a7c2fdf9-2652-4df0-b071-0b58e22d584e.png)

We currently bracket the types in these sections in the hopes of one day applying different formatting to them.

Related issue: #1678 

### This originally came up during PR review
For consistency, let's bracket the type. We'll probably want to italicize the types or render them a little differently on the docs page
```suggestion
    Examples: {float} or {int} representing the value to populate the component.
```

_Originally posted by @abidlabs in https://github.com/gradio-app/gradio/pull/1761#discussion_r918388736_

PR #1879, opened 2022-07-25
Make supported events and other sections of component doc stand out

### LINKED ISSUE #1678: Make supported events for each component stand out in the docs page

Similar to https://github.com/gradio-app/gradio/issues/1368 but smaller in scope

As a gradio user reading the docs, I wish the supported events for each component stood out from the surrounding text so I could more easily see which events a component supports.

image

Up for debate. One thing we could do is put the supported events in their own line and make the font slightly bigger. Something like the following where "Supported Events" is in a different color.

### **Supported Events**: _change()_

Might be worthwhile to apply this same principle to preprocessing and postprocessing parts of the docs. 

**Additional context**

freddyaboulton:
# Description

Fixes #1678

I think changing the font color for the supported events (preprocessing/postprocessing) sections + line break is sufficient

image

# Checklist:

### DIFF
--- a/website/homepage/src/docs/obj_doc_template.html
+++ b/website/homepage/src/docs/obj_doc_template.html
@@ -50,14 +50,15 @@
     </div>
   {% endif %}
   <p class="mt-8 mb-2 text-lg">{{ obj["description"] }}</p>
+  <br>
   {% if is_component %}
-  <p class="mb-2 text-lg text-gray-500">As input, {{ obj["tags"]["preprocessing"] }}</p>
-  <p class="mb-2 text-lg text-gray-500">As output, {{ obj["tags"]["postprocessing"] }}</p>
+  <p class="mb-2 text-lg text-gray-500"> <span class="text-orange-500">As input: </span> {{ obj["tags"]["preprocessing"] }}</p>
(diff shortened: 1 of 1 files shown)


website/homepage/src/docs/obj_doc_template.html
```diff
@@ -50,14 +50,15 @@
     </div>
   {% endif %}
   <p class="mt-8 mb-2 text-lg">{{ obj["description"] }}</p>
+  <br>
```
freddyaboulton, line 53: I think the line break helps with readability

abidlabs:
This LGTM.

One thing thing that would be nice (either in this PR or in a separate PR) would be to transform anything inside curly brackets to monospace to reflect the fact that it is a Python type. (An alternative that might be easier on the eyes would be to transform them to italics). But as it stands, the curly brackets are kind of mysterious!

image

abidlabs (approved):

freddyaboulton:
@abidlabs Thank you for the review! I agree the curly braces look strange and we should definitely apply different formatting.

That's being tracked by #1763 so I'll leave it for another PR!

PR #1880, opened 2022-07-25
Adds multistep blocks to docs

abidlabs:
Adds a multistep demo to the new documentation around Blocks. I think this is sufficient to explain how a multistep block works, but if people feel like more explanation is needed, let me know!


gradio/blocks.py
```diff
@@ -262,7 +262,7 @@ def update(name):
             btn.click(fn=update, inputs=inp, outputs=out)
 
         demo.launch()
-    Demos: blocks_hello, blocks_flipper, blocks_speech_text_length, generate_english_german
+    Demos: blocks_hello, blocks_flipper, blocks_speech_text_sentiment, generate_english_german
```
freddyaboulton, line 265: Looks good to me but I actually can't run this demo locally?

I get 
```python
 raise ValueError("We expect a numpy ndarray as input")
```

Had to do

```python
asr(speech[1].astype(float))
```

freddyaboulton (approved):
@abidlabs Agree we don't need a whole guide but the demo you added doesn't work for me locally? Get an exception from transformers.

gradio/blocks.py
abidlabs: Interesting, when using `tensorflow` as a backend for `pipeline`, it works. Let me see if it's a pytorch/tensorflow difference

abidlabs: Wait I'm actually confused by this -- `speech` should be a filepath. What does `speech[1].astype(float)`

abidlabs: I also tested on Spaces, and the code is running fine as it was written originally: https://huggingface.co/spaces/abidlabs/asr-test-pt

Is there something different on your local version of the demo?

freddyaboulton: @abidlabs could be a difference in my transformers version.

I can’t check cause I’m at the airport now. Feel free to merge, as long as the demo works on the website were good!

abidlabs: Thanks, safe travels!

PR #1881, opened 2022-07-25
Queue refactor backend fe

pngwn:
New queueing frontend


FarukOzderim (commented):
Let's not forget this here :D, better to commit a different file for testing?

demo/fake_gan_2/run.py
```diff
@@ -8,7 +8,7 @@
 
 
 def fake_gan(*args):
-    time.sleep(1)
+    time.sleep(60)
```
FarukOzderim, line 11: Let's not forget this here :D, better to commit a different file for testing?

pngwn: I think it is okay for now, I'll need to redo some of this work soon anyway. Can you let me know when you are up to date with main and I'll create another new branch? Because I'll have too many conflict due to the recent frontend refactor.

PR #1882, opened 2022-07-25
Add example of using component update method 

### LINKED ISSUE #1359: Add component update to docs

In workflows with multiple steps it might be necessary to make the user choose based on output for model 1 for subsequent models. 
Currently, the radio and dropdown fields only support changing the active value but not all the values. 

Like updating the textbox i'd like to be able to do the following:
```
import gradio as gr

def update(name):
    return [f'Hello {name}',f'Bye {name}', f'Good morning {name}' ]

demo = gr.Blocks()

with demo:
    inp = gr.Textbox(placeholder="What is your name?")
    out = gr.Radio(label="Greeting to use")
    
    inp.change(fn=update, 
               inputs=inp, 
               outputs=out)

demo.launch()
```

freddyaboulton:
# Description

Fixes #1359 

image

# Checklist:

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -208,7 +208,7 @@ def update(**kwargs) -> dict:
     Note that your editor's autocompletion will suggest proper parameters
     if you use the update method on the component.
 
-    Demos: blocks_update, blocks_essay, blocks_essay_update
+    Demos: blocks_essay, blocks_update, blocks_essay_update
 
     Parameters:
         kwargs: Key-word arguments used to update the component's properties.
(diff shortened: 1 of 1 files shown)


gradio/blocks.py
```diff
@@ -205,8 +205,10 @@ def update(**kwargs) -> dict:
     Updates component parameters.
     This is a shorthand for using the update method on a component.
     For example, rather than using gr.Number.update(...) you can just use gr.update(...).
+    Note that your editor's autocompletion will suggest proper parameters
+    if you use the update method on the component.
 
-    Demos: blocks_update, blocks_essay_update
+    Demos: blocks_update, blocks_essay, blocks_essay_update
```
freddyaboulton, line 211: blocks_essay uses the component method

image

FarukOzderim, line 211: Good job!

Would prefer if the default example is Component.update version.
```suggestion
    Demos: blocks_essay, blocks_update, blocks_essay_update
```

PR #1884, opened 2022-07-26
Better navigation and search in docs

### LINKED ISSUE #1674: Improve the searchability of the docs page

As a gradio user, I find it hard to navigate the docs page to the section I need. 

For example, to find the documentation for `Series`, I either `ctrl + F` for "Series" (which brings up false positives like the Timeseries component) or click on `Combining Interfaces` (only because I know that's where it is) and then scroll down.

I think this makes the docs hard to use and it also makes important information harder to find.

It's definitely up for debate. One thing I thought of:

- Add more granularity to the sidebar. I'm used to having a drop down for each subheading of the sidebar that tells you what's under that section with a clickable link (from the pandas docs):

image

**Additional context**

aliabd:
- More complete nav bar with sub-headers 
- Nav bar expands based on scrolling/position in screen
- Search bar to find classes quicker (for instance ctrl+f "parallel" will take you to some parameters in interface.launch() before gradio.Parallel)

Fixes #1674

https://user-images.githubusercontent.com/9021060/180898036-dd6a91b8-7f2b-47ce-a36c-96057fd87cda.mov

(code needs some refactoring before review)


abidlabs:
Sweeet!

abidlabs:
This is awesome @aliabd!! I tested it and it looks great to me. 

A couple of minor points:
* CTRL+K didn't work for me (on Windows). It looks like Edge/Chrome uses CTRL+K to open up search with Google, so that's what popped up for me instead. _However_, Hugging Face docs somehow overrides this and CTRL+K does work for those docs, so there might be a way to do override the default CTRL+K
* I can imagine some people wanting to search through parameter names, not just class names. It would be cool if the search could also search through parameters.

I'm really loving our documentation now!

aliabd:
@abidlabs thanks for the feedback! 

There is a way to override defaults, though I thought it would just mess with users flow. I guess it's not a big deal though so will implement.

> I can imagine some people wanting to search through parameter names, not just class names. It would be cool if the search could also search through parameters.

This is possible, though might get complicated. Also the more granular we get the closer we get to ctrl+f, which isn't ideal because too many things surface for the user. Will look into it though

freddyaboulton:
@aliabd This is awesome! Worked perfectly for me locally.

This will deliver immediate value - so exciting!!

Searching by parameters could be useful but I'm worried users would get a lot of false positives (like every component would come up if you search for "value"). Worth looking into but I think it'd be ok to merge without it!

abidlabs:
Yes, I'm fine with merging as well!

abidlabs (approved):

abidlabs:
Works great for me now @aliabd!

abidlabs (approved):

abidlabs:
Took another look at this and everything looks great! (Include the esc icon 😃!)

aliabid94:
LGTM left some small comments. I also do wonder about searching parameter names, etc., what's the standard in other docs searches?

aliabd:
@aliabid94 it doesn't look like you left any comments. Am I missing something?

The standard is to search through everything and not just parameter names (examples of this are [tailwind](https://tailwindcss.com/docs), and even [huggingface](https://huggingface.co/docs/transformers/)). I do wanna add that later on but I think it's gonna take a bit of redesign and this should be shipped first.

aliabid94 (approved):

PR #1889, opened 2022-07-26
Add estimation parameters to queue

FarukOzderim:
This took way longer than expected, it seems I cannot focus well. Mind is too foggy, could not test it out whether if it works, sorry!


gradio/event_queue.py
```diff
@@ -14,8 +14,12 @@
 
 class Estimation(BaseModel):
     msg: Optional[str] = "estimation"
+    rank: Optional[int] = -1  # waiting duration for the xth rank:
```
aliabid94, line 17: I think we can remove these comments. The purpose of the variables should be evident from their names, and the logic to calculate them can be found in the code itself.

FarukOzderim: Added because the meaning was not very clear before, and we had talked with Pete about it. Wanted to make it clear and they do no harm. Makes it clear for auto-scaling as well.

FarukOzderim:
Tested it out and works well, see the new Estimation message:

image
{
    "msg": "estimation",
    "rank": 0,
    "queue_size": 1,
    "avg_process_time": 1.0,
    "avg_concurrent_process_time": 1.0,
    "queue_duration": 1
}

PR #1891, opened 2022-07-26
ensure falsey values are set to null for images

### LINKED ISSUE #1838: Image component in a tab cannot be clicked
### Describe the bug

I'm seeing a weird bug where if a `gr.Image` is inside a `gr.Tab` that is not selected by default, then it cannot be clicked to upload an image -- until it is cleared (even though no image is loaded). It's a bit hard to describe, so look at the reproduction below:

### Reproduction

```py
import gradio as gr

demo = gr.Blocks()

with demo:
    with gr.Tabs():
        with gr.TabItem("Flip Text"):
            gr.Textbox()
        with gr.TabItem("Flip Image"):
            image_input = gr.Image()
            image_output = gr.Image()
            image_button = gr.Button("Flip")

    image_button.click(lambda x:x, inputs=image_input, outputs=image_output)

demo.launch()
```

Here, `image_input` should be clickable, but it is not (until you press the X to clear it):

image

### Screenshot

_No response_

### Logs

```shell
N/A
```

### System Info

```shell
3.1
```

### Severity

annoying

pngwn:
Closes #1838.

### DIFF
--- a/ui/packages/app/src/components/Image/Image.svelte
+++ b/ui/packages/app/src/components/Image/Image.svelte
@@ -27,6 +27,10 @@
 
 	$: value, dispatch("change");
 	let dragging: boolean;
+
+	$: console.log($$props);
+
+	$: value = !value ? null : value;
 </script>
 
 <Block
--- a/ui/packages/app/src/components/Image/Image.svelte
+++ b/ui/packages/app/src/components/Image/Image.svelte
@@ -28,8 +28,6 @@
 	$: value, dispatch("change");
 	let dragging: boolean;
 
-	$: console.log($$props);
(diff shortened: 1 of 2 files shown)


ui/packages/app/src/components/Image/Image.svelte
```diff
@@ -27,6 +27,8 @@
 
 	$: value, dispatch("change");
 	let dragging: boolean;
+
+	$: value = !value ? null : value;
```
aliabid94, line 31: how about `value = value || null`?

aliabid94 (approved):
lgtm

PR #1892, opened 2022-07-26
Add Events to docs page

### LINKED ISSUE #1368: Add events section to docs

1. We do have list supported events within docs, however we do not have a doc for events. 
image
- [ ] Shall we document a generic event or one example event like `change`?
- [ ] Shall we document every event [signature](https://github.com/gradio-app/gradio/blob/main/gradio/events.py), because they have different signatures?
- [ ] We could also explain what do event do, though they are quite self-explanatory with their names. 
- [ ] We could also create a matrix for components/events they support.
- [ ] Could link to the abstract classes for [events](https://github.com/gradio-app/gradio/blob/main/gradio/events.py) to explain their function signatures. 
- [ ] We could explain each abstract class and its signature in the docs as well. We could even create event-component matrix with these abstract classes. This would help community to add their own components or events.

- [ ] Alternatively we could just link from Blocks Docs to the Blocks Guide Events [section](https://gradio.app/introduction_to_blocks/#events). Though this guide is outdated and misses js within events.

WDYT @aliabd , @abidlabs?

freddyaboulton:
# Description

Fixes #1368

# Checklist:


gradio/events.py
```diff
@@ -35,8 +39,7 @@ def change(
             status_tracker: StatusTracker to visualize function progress
             scroll_to_output: If True, will scroll to output component on completion
             show_progress: If True, will show progress animation while pending
-            _js: Optional frontend js method to run before running 'fn'. Input arguments for js method are values of input and outputs components, return should be a list of values for output component.
```
freddyaboulton, line 38: Removing the underscore params from the docstring because our doc building code does not allow documenting them.

website/homepage/src/docs/event_template.html
```diff
@@ -0,0 +1,60 @@
+{% for obj in event["fns"] %}
```
freddyaboulton, line 1: This is basically obj_doc_template but with all the component-specific logic removed and with some minor 

I decided to add a new template for events because events are different from our other apis. Users are not intended to use the `Clickable` class directly, instead they use the `click` method on the associated component. I felt the existing template did not accommodate this difference.

So what this template does is treat the methods of each Event class as if they were a top-level function.

gradio/documentation.py
```diff
@@ -103,7 +103,7 @@ def document_fn(fn):
 def document_cls(cls):
     doc_str = inspect.getdoc(cls)
     if doc_str is None:
-        return "", {}
+        return "", {}, ""
```
freddyaboulton, line 106: So that this function always returns a tuple with three elements.

website/homepage/src/docs/template.html
```diff
@@ -157,6 +164,25 @@ <h3 class="text-3xl font-light my-4" id="update-section-header">Update</h3>
...
+           Events give you control of when your functions are run in the app. For example,
+           on a Button, the `click` event will trigger the function when the button is clicked. Each component
+           and many layouts have specific events that they support. You can find the events that a component supports
+           by going to its documentation.
+          </p>
+        </div>
+        {% for event in docs["events"] %}
+            {% include "docs/event_template.html" %}
+        {% endfor %}
+        <!-- Document the load event with the events as well-->
```
freddyaboulton, line 181: I think the `load` event should go with the other events as well. I'm worried users will not know of the load event if it's only listed in the `Blocks` docs.

gradio/events.py
FarukOzderim: I think we still need description of _js in the function for users who check the function to understand what it is clearly. Can we find a better solution? How does describing it just after docstring as a comment line?

Valid for postprocess and preprocess as well.

gradio/events.py
```diff
@@ -35,8 +39,7 @@ def change(
...
-        Returns: None
+            queue: If True, will place the request on the queue, if the queue exists
         """
```
FarukOzderim, line 43:
```suggestion
        """
        #_js: Optional frontend js method to run before running 'fn'. Input arguments for js method are values of input and outputs components, return should be a list of values for output component.
```

gradio/events.py
freddyaboulton: Sounds good to me!

freddyaboulton:
@FarukOzderim Added docstrings for _js and the other underscore parameters. This should be good for another look!

pngwn:
Is there a reason we aren't documenting events on a per component basis? I could be wrong but this looks like a global 'Events' section whereas the events are specific to each component. I can understand there may be a concern about duplication but I think it would be better for users, as they would typically be interested in the events for a component rather than events in an of themselves.

FarukOzderim:
> Is there a reason we aren't documenting events on a per component basis? I could be wrong but this looks like a global 'Events' section whereas the events are specific to each component. I can understand there may be a concern about duplication but I think it would be better for users, as they would typically be interested in the events for a component rather than events in an of themselves.

Agree on this, I think we should add a matrix showing which component supports which events at the events section, and give example usage like, `textbox.change...`

gradio/components.py
```diff
@@ -3729,8 +3821,16 @@ def postprocess(self, y: List[Tuple[str, str]]) -> List[Tuple[str, str]]:
     def style(
         self,
         rounded: Optional[bool | Tuple[bool, bool, bool, bool]] = None,
-        color_map: Optional[Dict[str, str]] = None,
+        color_map: Optional[List[str, str]] = None,
```
freddyaboulton, line 3824: `blocks_style` demo uses a list

website/homepage/src/docs/obj_doc_template.html
```diff
@@ -128,11 +128,11 @@ <h4 class="mt-4 p-3 font-semibold">Example Usage</h4>
     <h4 class="mt-4 p-3 font-semibold">Methods</h4>
       <div class="flex flex-col gap-8 pl-12">
         {% for fn in obj["fns"] %}
-          {% with obj=fn, is_class=False, parent=parent + "." + obj["name"] %}
+          {% with obj=fn, is_component=False, parent=parent + "." + obj["name"] %}
             {% include "docs/obj_doc_template.html" %}
           {% endwith %}
         {% endfor %}
-      <div class="ml-12">
+      <div class="ml-12"> </div>
```
freddyaboulton, line 135: Pre-existing bug. Need to close the div here otherwise the components are all nested within the `textbox` div and the current-nav-link logic gets thrown off.

freddyaboulton:
@FarukOzderim @pngwn and everyone else, this should be good for another review!

abidlabs:
Hey @freddyaboulton this looks great! One small thing is that we can improve a few of the docstrings

* For parameters like `rounded` in `style()`, it says, "starting from top and proceeding clock-wise", but does it start from the "top left" or "top right"?
* For the `container` parameter in `style()`, it says "If true, will place the component in a container." but what  does the container look like here? Does it just provide some additional padding? Is it colored a particular color?
* For the `style()`, the current docstring ("Style the component") could be reworded as: "This method can be used to changes the appearance of the component"
* For all of the events(), perhaps we should clarify that this is only for Blocks. So something like, "This event is triggered when the user presses the Enter key while the component (e.g. a textbox) is focused. This method can be used when this component is in a Gradio Blocks."

abidlabs (approved):

freddyaboulton:
Thank you for the thorough review @abidlabs ! 

Agree on all the points and I've made the corresponding changes. Merging now!

PR #1893, opened 2022-07-26
Always return headers from postprocess

### LINKED ISSUE #1887: Dataframe headers overwritten with lists/arrays
### Describe the bug

When `Dataframe` outputs data which is a Python list/numpy array, the header names get overwritten with numeric indicies.

Maybe this is not a bug but caused by me not using it properly. However, I did not find relevant information in the documentation.

### Reproduction

```
import gradio as gr

interface = gr.Interface(
    fn=lambda: [
        ["x1", "y1"],
        ["x2", "y2"],
        ["x3", "y3"],
    ],
    inputs=None,
    outputs=gr.DataFrame(
        col_count=2,
        row_count=3,
        type="array",
        headers=["A", "B"],
        interactive=False,
    )
)
interface.launch(debug=True)
```

### Screenshot

Before:

image

After:

image

### Logs

```shell
No logs available.
```

### System Info

```shell
gradio 3.1.1, Fedora 36, Firefox 102.0
```

### Severity

pngwn:
Closes #1887.

This PR ensures that Dataframe's postprocess method always returns headers along with the data, this means the frontend does not need to decide what to do with the headers. I've also moved the automatic generation of headers (when none are given, we use the list index + 1) to the backend.

I have also added some additional logic to generate a new compound set of headers for the following case:

- users set headers: `["A", "B"]`
- User returns new dataframe as a list that has more than 2 columns

In this case the existing headers will be used for the first two columns, any subsequent headers will be the list index +1. So in the above case the initial headers would be: `["A", "B"]` and after returning the new list `["A", "B", "3", "4"]`.

This PR modifies the API payload a little but doesn't change anything user facing.

### DIFF
--- a/demo/fake_gan_2/run.py
+++ b/demo/fake_gan_2/run.py
@@ -38,4 +38,4 @@ demo = gr.Interface(
 )
 
 if __name__ == "__main__":
-    demo.launch()
+    demo.launch(enable_queue=True)
--- a/demo/fake_gan_2/run.py
+++ b/demo/fake_gan_2/run.py
@@ -1,41 +1,17 @@
-# This demo needs to be run from the repo folder.
-# python demo/fake_gan/run.py
-import os
-import random
-import time
-
 import gradio as gr
-
-
-def fake_gan(*args):
-    time.sleep(1)
-    image = random.choice(
-        [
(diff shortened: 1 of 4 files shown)


freddyaboulton (commented):
@pngwn I ran `fraud_detector` and for some reason `sensitivity` is no longer displayed in the examples

image


Edit: maybe this isn't associated with this PR? I see the same behavior on main but not on the latest released version of gradio.

Python changes look good to me and I can test that this fixes the original bug fix. Would be good to add some python unit tests for the "returns less than original number of headers" and "returns more than the original number of headers" cases.

abidlabs:
Added unit tests as @freddyaboulton suggested. So now this should be good to merge!

PR #1896, opened 2022-07-26
Suggestion: add more info on `update()`

abidlabs:
Reading through the docs, I think we need to add a little more explanation in this section: https://gradio.app/docs/#update-section-header

This is just a proposal of what we could add. Feel free to edit @freddyaboulton 

This code hasn't been tested as I'm traveling right now :)


freddyaboulton (commented):
@abidlabs Thanks for the PR! Agree with the new prose explaining the update function.

I'm just going to move your example usage to the docstring and merge

abidlabs:
Perfect, thanks @freddyaboulton!

Issue #1740, opened 2022-07-08
Write a guide on how data flows through an app


freddyaboulton:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
As a gradio user, I wish there was a guide on how data flows through the components of my app. 

This will help me understand how the value in a component eventually makes it to my prediction function and how the output of my prediction function is turned into what a user of my app will see. 

Having this understanding will help me chain together components properly and avoid bugs when developing my app.

**Describe the solution you'd like**  
We should write a guide that documents how data is preprocessed/postprocessed by each of the components in an app and passed from component to component. 

One thing that may confuse new users is that returning a filepath in a function means the whole file is displayed in the UI. 

**Additional context**  
We have this diagram in the docs but we should go into greater depth

![image](https://user-images.githubusercontent.com/41651716/178072094-2f2a8555-fff8-4269-8941-2648c019ebf4.png)

PR #1897, opened 2022-07-26
How data flows through an app

abidlabs:
I added some more details on how data is preprocesses/postprocessed in one of the early Guides. Haven't been able to test locally b/c of travels, so I would appreciate if someone can do a quick review. 

Feedback welcome!

Fixes: #1740


freddyaboulton (approved):
Thank you @abidlabs ! This is useful material for users. I was able to build locally and it looks good

image

abidlabs:
Thanks for confirming @freddyaboulton!

PR #1899, opened 2022-07-27
Embedding the spaces on the landing page using web components

aliabd:
Removes the iframes on the landing page, replacing them with web component embeddings. Much better for styling and alignment. Both the spaces (sdk_version) and the landing page (gradio.js) are now pinned to the same version of gradio (v3.1.1). We can update this for major releases but this will insure the demos never break. Also added copy buttons to the codeblocks like we have in docs and guides.


aliabid94 (approved):
LGTM

PR #1900, opened 2022-07-27
Fix default value in docs for objects

aliabd:
Issue is the default value in docs parameter tables gets thrown off for objects that aren't builtin types (eg `flagging_callback = CSVLogger()`). This fixes that by detecting them and using the class name instead.


abidlabs:
Nice catch!

abidlabs (approved):

PR #1905, opened 2022-07-28
Referencing guides in docstrings and other fixes

aliabd:
Fixes to guides in docs and in the gallery page: 
* Tie guides to docs in docstrings and not metadata of guide (similar to demos) 
* Reference more guides in docs
* Fix missing guides section in Interface and Blocks
* Remove 'pinned' metatag that was used to order guides in the gallery page 
* Fix space before comma in tags


freddyaboulton (commented):
@aliabd This is a great idea! I have a couple of suggestions for improvement. Once those are in, I think this should be good to merge.

### Fix flagging guide links
One thing that jumped out is that the guides in the flagging section look weird. Right under the class, it says "No guide yet, contribute a guide" but then right under that text there is a link to a guide.

image

I don't think this is caused by this PR but might as well fix it while we're fixing the guides.

### Add Controlling Layout to Blocks and Rows

We should link the controlling layout guide to Columns and Rows components as they are used in that guide.

aliabd:
@freddyaboulton great catches! Will fix both right away

aliabid94 (approved):
lgtm, much cleaner now.

PR #1909, opened 2022-07-29
Queue tweaks

FarukOzderim:
# Description
Some tweaks on the Estimation payload after offline discussions with Ali. Will be merged after #1910


aliabid94:
LGTM

PR #1912, opened 2022-07-30
Draft: Release new queue beta

aliabid94:


.circleci/config.yml
```diff
@@ -58,7 +58,7 @@ jobs:
       - run:
           command: |
             . venv/bin/activate
-            python -m  flake8 --ignore=E731,E501,E722,W503,E126,F401,E203,F403 gradio test
+            python -m  flake8 --ignore=E731,E501,E722,W503,E126,F401,E203,F403,F541 gradio test
```
aliabid94, line 61: why did we add this ignore rule?

gradio/blocks.py
```diff
@@ -748,6 +750,11 @@ def launch(
         ssl_certfile: Optional[str] = None,
         ssl_keyfile_password: Optional[str] = None,
         quiet: bool = False,
```
aliabid94, line 752: I'm afraid that adding all these configurables to the launch method will lead to expanding our argument list endlessly. Here's what I suggest instead: create a method called `configure_queue` to blocks, and add the parameters there. This method will modify the Queue class directly. Two benefits to this approach:
- We don't end up with this ever-expanding argument list to `launch`
- All the queue configuration attributes are grouped together in one spot.

gradio/blocks.py
```diff
@@ -997,6 +1026,10 @@ def close(self, verbose: bool = True) -> None:
         Closes the Interface that was launched and frees the port.
         """
         try:
+            from gradio.event_queue import Queue
```
aliabid94, line 1029: why importing here when we already have a reference to event_queue imported at the top?

gradio/event_queue.py
```diff
@@ -0,0 +1,276 @@
...
+import fastapi
+from fastapi import WebSocketDisconnect
+from pydantic import BaseModel
+
+from gradio.utils import Request, run_coro_in_background
+
+
+class Estimation(BaseModel):
+    msg: Optional[str] = "estimation"
+    rank: Optional[int] = -1  # waiting duration for the xth rank:
```
aliabid94, line 17: as we discussed, let's make the variable names extra clear and remove the comments

gradio/event_queue.py
```diff
@@ -0,0 +1,276 @@
...
+    @classmethod
+    def clean_job(cls, event: Event):
+        cls.ACTIVE_JOBS[cls.ACTIVE_JOBS.index(event)] = None
+
+    @classmethod
+    async def clean_event(cls, event: Event) -> None:
+        # Using mutex to avoid editing a list in use
+        async with cls.LOCK:
+            # TODO: Might log number of cleaned events in the future
+            try:  # TODO: Might search with event hash to speed up
```
aliabid94, line 121: insstead of using try / catch, can you just just use an if statement? If event in queue...

gradio/event_queue.py
```diff
@@ -0,0 +1,276 @@
...
+                for event in cls.EVENT_QUEUE[: cls.DATA_GATHERING_STARTS_AT]
+            ]
+        )
+
+    @classmethod
+    async def gather_event_data(cls, event: Event) -> None:
+        """
+        Gather data for the event
+
+        Args:
```
aliabid94, line 143: Here and elsewhere throughout the code, can you use `Parameters` instead of `Args` to be consistent with docstrings throughout the library

.circleci/config.yml
FarukOzderim: Cant remember it now, but was probably smt needed and cannot be bypassed.

gradio/blocks.py
FarukOzderim: I like it!

gradio/event_queue.py
FarukOzderim: yeah I've changed them in #1909

gradio/event_queue.py
FarukOzderim: Will do, but I actually don't like the current status, different standards in the codebase and I don't know which standard we're following. Will follow up this is in another issue.

PR #1918, opened 2022-08-01
Scripts become much easier to read and understand

sheikhartin:
### Description

Most of the scripts in the `scripts` folder were written in a complicated way, and there is no convincing reason for this when they can be made simpler (both in terms of reading and understanding) with a few relatively small changes. They also had bugs and I didn't just simplify...

This condition with a strange logic:

```bash
if [ -z "$(ls | grep CONTRIBUTING.md)" ]; then
  echo "Please run the script from repo directory"
  exit -1
else
  ...
fi
```
Replaced by this in short:

```bash
cd "$(dirname ${0})/.."
```
A comment for these line of codes in `gh_action_pypi.sh` to avoid misunderstanding:

```bash
old_version=$(grep -Po "(?<=version=\")[^\"]+(?=\")" setup.py)
read -r new_version < gradio/version.txt
sed -i "s/version=\"$old_version\"/version=\"$new_version\"/g" setup.py
```

The above code actually does NOTHING! (Except the latter) The first one probably wants to get the previous version from `setup.py`, but `setup.py` itself reads the version from `gradio/version.txt`, so it won't give you anything to read and replace until it is executed.

But don't worry, I wrote a better one (in `upload_to_pypi.sh`):

```bash
old_version="$(cat gradio/version.txt)"
read -p "Current version is ${old_version}. What is the new version? " new_version
echo "So you want to release version ${new_version}. Updating gradio/version.txt..."
echo "${new_version}" > gradio/version.txt
```

### Checklist

I wanted to thank the Gradio team (and all contributors) for simplifying the problem of deploying machine learning, data science, etc. applications... This project is very useful and valuable, especially for those who are less interested in the web and front-end :heart:

### DIFF
--- a/scripts/build_frontend.sh
+++ b/scripts/build_frontend.sh
@@ -1,10 +1,11 @@
 #!/bin/bash
-if [ -z "$(ls | grep CONTRIBUTING.md)" ]; then
-  echo "Please run the script from repo directory"
-  exit -1
-else
-  echo "Building the frontend"
-  cd ui
-  pnpm i --frozen-lockfile
-  pnpm build
-fi
+
+cd "$(dirname ${0})/.."
+source scripts/helpers.sh
+
+pnpm_required
+
+echo "Building the frontend..."
+cd ui
+pnpm i --frozen-lockfile
+pnpm build
--- a/scripts/helpers.sh
+++ b/scripts/helpers.sh
@@ -26,7 +26,8 @@ function program_required() {
 #   None
 #######################################
 function pip_required() {
-  program_required "pip" "https://pip.pypa.io/en/stable/installing/"
+  program_required "pip" "https://pip.pypa.io/en/stable/installation/"
+
 }
 
 #######################################
@@ -41,6 +42,17 @@ function npm_required() {
 }
 
 #######################################
+# Check for the PNPM program.
+# Arguments:
+#   None
+# Outputs:
+#   None
+#######################################
+function pnpm_required() {
+  program_required "pnpm" "https://pnpm.io/6.x/installation"
+}
+
+#######################################
 # Check for the CircleCI program.
 # Arguments:
 #   None
@@ -49,17 +61,6 @@ function npm_required() {
 #######################################
 function circleci_required() {
   program_required "circleci" "https://circleci.com/docs/1.0/local-cli/"
-}
-
-#######################################
-# Check for the PNPM program.
-# Arguments:
-#   None
-# Outputs:
-#   None
-#######################################
-function pnpm_required() {
-  program_required "pnpm" "https://pnpm.js.org/en/latest/installation/"
 }
 
 #######################################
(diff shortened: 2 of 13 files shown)


abidlabs:
Thanks for the kind words and the contribution @sheikhartin! Reviewing your PR now

scripts/build_frontend.sh
```diff
@@ -1,10 +1,11 @@
...
-  echo "Please run the script from repo directory"
-  exit -1
-else
-  echo "Building the frontend"
-  cd ui
-  pnpm i --frozen-lockfile
-  pnpm build
-fi
+
+cd "$(dirname ${0})/.."
```
abidlabs, line 3: Thanks!

scripts/helpers.sh
```diff
@@ -0,0 +1,85 @@
...
+
+#######################################
+# Check for the PNPM program.
+# Arguments:
+#   None
+# Outputs:
+#   None
+#######################################
+function pnpm_required() {
+  program_required "pnpm" "https://pnpm.js.org/en/latest/installation/"
```
abidlabs, line 62: We currently support pnpm 6, so let's use this URL: https://pnpm.io/6.x/installation

scripts/helpers.sh
```diff
@@ -0,0 +1,85 @@
...
+
+#######################################
+# Check for the PIP program.
+# Arguments:
+#   None
+# Outputs:
+#   None
+#######################################
+function pip_required() {
+  program_required "pip" "https://pip.pypa.io/en/stable/installing/"
```
abidlabs, line 29:
```suggestion
  program_required "pip" "https://pip.pypa.io/en/stable/installation/"
```

scripts/run_all_tests.sh
```diff
@@ -1,8 +1,8 @@
...
-if [ -z "$(ls | grep CONTRIBUTING.md)" ]; then
-  echo "Please run the script from repo directory"
-  exit -1
-else
-  echo "Running the tests"
-  python -m pytest --cov=gradio --durations=20 --durations-min=1 test
-fi
+
+cd "$(dirname ${0})/.."
+# source scripts/create_test_requirements.sh
```
abidlabs, line 4: I don't think we want to create the test requirements each time you run the tests. It unnecessarily slows the process

scripts/run_all_tests.sh
```diff
@@ -1,8 +1,8 @@
...
+# source scripts/install_test_requirements.sh
```
abidlabs, line 5: I would probably just remove this as well

abidlabs:
Thank you for the contribution @sheikhartin! I have made some tweaks above, but overall these changes are a welcome improvement to the scripts! 

cc @aliabid94 @FarukOzderim

abidlabs (approved):

abidlabs:
LGTM, thanks for making the changes @sheikhartin! Always nice to have clean scripts :)

PR #1919, opened 2022-08-01
Fix broken API page

aliabid94:
When we changed the documentation style, the view API page for Interfaces got broken. Should be fixed now.

### DIFF
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -177,7 +177,6 @@ class IOComponent(Component):
                 "is_example": True,
             }
 
-    # Input Functionalities
     def preprocess(self, x: Any) -> Any:
         """
         Any preprocessing needed to be performed on function input.
@@ -237,7 +236,6 @@ class IOComponent(Component):
         """
         pass
 
-    # Output Functionalities
     def postprocess(self, y):
         """
         Any postprocessing needed to be performed on function output.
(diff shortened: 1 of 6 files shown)


abidlabs:
What exactly was broken previously @aliabid94?

gradio/documentation.py
```diff
@@ -177,3 +176,16 @@ def generate_documentation():
                 )
             documentation[mode].append(cls_documentation)
     return documentation
+
+
+def document_component_api(component_cls, target):
```
abidlabs, line 181: Can you give some explanation, what's happening here?

aliabid94: Added documentation. Uses

aliabid94:
> What exactly was broken previously @aliabid94?

Previously the app documentation used the docstring to get the types expected by the API. We removed types from the docstring since it was redundant with the type hint. Now "view api" page uses the same type-hint based logic to generate the api page.

gradio/components.py
```diff
@@ -802,9 +782,7 @@ def get_interpretation_scores(
     def generate_sample(self) -> float:
         return self.maximum
 
-        # Output Functionalities
-
-    def postprocess(self, y: float | None) -> float:
+    def postprocess(self, y: float | None) -> float | None:
```
abidlabs, line 785: `postprocess()` always produces a float
```suggestion
    def postprocess(self, y: float | None) -> float:
```

abidlabs:
Fixed some typing in some of the components' `preprocess()` and `postprocess()` functions

abidlabs:
Otherwise, LGTM! Nice fix

abidlabs (approved):

PR #1921, opened 2022-08-01
Fixes Download button does not respect the filepath returned by the function

### LINKED ISSUE #1689: Download button does not respect the filepath returned by the function
### Describe the bug

In the `zip_two_files` demo:

```python
import os
from zipfile import ZipFile

import gradio as gr


def zip_two_files(file1, file2):
    with ZipFile("tmp.zip", "w") as zipObj:
        zipObj.write(file1.name, "file1")
        zipObj.write(file2.name, "file2")
    return "tmp.zip"


demo = gr.Interface(
    zip_two_files,
    ["file", "file"],
    "file",
    examples=[
        [os.path.join(os.path.dirname(__file__),"files/titanic.csv"), 
         os.path.join(os.path.dirname(__file__),"files/titanic.csv")],
    ],
)

if __name__ == "__main__":
    demo.launch()
```

When you download the file, it is renamed `download.zip` by default on Chrome.

### Download version 3.0.22
file_download_bug

That's not the case on version 3.0.12. The file is called `tmp` because that's what the prediction function returns. You can see the same behavior on this space: https://huggingface.co/spaces/radames/dpt-depth-estimation-3d-obj Was this change intentional?

### Download version 3.0.12
file_download_correct

### Reproduction

Run the `zip_two_files` demo on 3.0.22 and download the file.

### Screenshot

_No response_

### Logs

```shell
-
```

### System Info

```shell
3.0.22
```

### Severity

annoying

dawoodkhan82:
# Description
Fixes #1689

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

Closes: # (issue)
#1689

# Checklist:

### DIFF
--- a/ui/packages/file/src/File.svelte
+++ b/ui/packages/file/src/File.svelte
@@ -50,7 +50,7 @@
 				<div class="file-size w-3/12 p-2 hover:underline">
 					<a
 						href={download_files(value)}
-						download
+						download={display_file_name(value)}
 						class="text-indigo-600 hover:underline dark:text-indigo-300"
 						>Download</a
 					>


freddyaboulton (approved):
Thanks @dawoodkhan82 - was able to verify that this fixed the issue!


download_filepath_respected

freddyaboulton:
Might be good to write a unit test/integration test around this? All of the reports about this make me think this is important to users and we should be careful about accidentally breaking this in the future.

abidlabs (approved):
LGTM @dawoodkhan82!

PR #1926, opened 2022-08-01
Fix dark mode with apps smaller than window size

### LINKED ISSUE #1772: Background color (and app) does not extend to entire window in 3.0.24
### Describe the bug

Previously, a Gradio app in its own window took up the entire window vertically, with the "built with gradio" and other footer information at the bottom and the title at the top. Similarly, the app took up all horizontal space.

In 3.0.24, the app instead takes up all horizontal space but not all vertical space. In dark mode, this looks particularly bad, since the browser fills the rest of the space with a light background (see screenshots below).

The desired behavior is in all versions I've used up to 3.0.17 and in 3.0.21.

Looking at the styling, it appears that the `min-h-screen` class is not applied to the div that contains the app. Removing that class from the app's div in an earlier version gives the same undesired behavior.

While I would prefer the footer to return to its previous position, it would also be acceptable if the area that is currently filled with light color was instead filled with the same background color as the app.

### Reproduction

I've uploaded images to Docker Hub for repro if desired.

You can run the server with `docker run -p$OPENPORT:11700 $image_name`

Repository on Docker Hub [here](https://hub.docker.com/repository/docker/fullstackdeeplearning/app-gradio).

Image with bug is tagged `3.0.24`, [here](https://hub.docker.com/layers/252509533/fullstackdeeplearning/app-gradio/3.0.24/images/sha256-7ae8611b5b7e0c156a4b8cf0406de9960026df3289ab5bb21a9e2937316ffd40).

Image without bug is tagged `3.0.21`, [here](https://hub.docker.com/layers/252517260/fullstackdeeplearning/app-gradio/3.0.21/images/sha256-7f63f60e046156bee47af7da16dfa4341e46a5a84e95f23559244dee930cfdb7?context=repo).

## 3.0.24 - not as desired

image

## 3.0.21 - as desired

image

## 3.0.21 without `min-h-screen` - not as desired

image

### Logs

```shell
n/a
```

### System Info

```shell
server was gradio 3.0.24 and 3.0.21 running on (dockerized and bare metal) linux, client was a chrome browser running on OS X
```

### Severity

aliabid94:
These is white-space at the bottom of any dark mode app thats not long enough for the whole page (most apps). 

Screen Shot 2022-08-01 at 4 00 40 PM

Looks very gross. Fixed now. 
Screen Shot 2022-08-01 at 4 02 41 PM

Also added the dark bg color to the body, so on chrome when you scroll past the body, it doesn't show white.

Fixes #1772

### DIFF
--- a/ui/packages/app/index.html
+++ b/ui/packages/app/index.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en" style="margin: 0; padding: 0">
+<html lang="en" style="margin: 0; padding: 0; min-height: 100%">
 	<head>
 		<meta charset="utf-8" />
 		<meta
@@ -54,14 +54,8 @@
 		<script src="https://cdnjs.cloudflare.com/ajax/libs/iframe-resizer/4.3.1/iframeResizer.contentWindow.min.js"></script>
 	</head>
 
-	<body style="width: 100%; margin: 0; padding: 0">
-		<gradio-app
-			><div
-				id="root"
-				style="display: flex; flex-direction: column; flex-grow: 1"
-			></div
-		></gradio-app>
-
+	<body style="width: 100%; margin: 0; padding: 0; height: 100%">
+		<gradio-app></gradio-app>
 		<script>
 			const ce = document.getElementsByTagName("gradio-app");
 			if (ce[0]) {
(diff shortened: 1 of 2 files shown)


pngwn (approved):
lgtm

abidlabs:
@aliabid94 it looks like this closes: #1772, right?

aliabid94:
> @aliabid94 it looks like this closes: #1772, right?

Yep

PR #1933, opened 2022-08-02
Fix broken frontend when iframed

aliabid94:
I had removed the "#root" div element earlier, but that is still being used for mounting within iframes. Restored


freddyaboulton (approved):
Looks good to me @aliabid94 ! Thanks for the fixes!

PR #1935, opened 2022-08-02
Allow Users to set random value on sliders on app load

### LINKED ISSUE #1849: Allow to randomize the input in a slider component

For demos of generative models, the input seed/slider can have a huge effect. Something that would be very useful in the slider is if we could randomize the initial value of the slider for each user. E.g. by setting `value="random"` then every  time someone opens the app they get a different value in the slider. 

**Additional context**  
Internal discussion: https://huggingface.slack.com/archives/C03HBN1C8CW/p1658396924819859

freddyaboulton:
# Description

Fixes #1849

To test, run `demo/blocks_random_slider`, `demo/interface_random_slider`, and `demo/kitchen_sink_random`

kitchen_sink_random

# Checklist:


pngwn:
I'm not sure about this API. Making strings have magical values doesn't seem particularly intuitive and doesn't really offer a huge amount of flexibility. Would be not allow the value to take a function that returns a number instead? This way we could support random values within custom ranges, random values with a seed number, etc.

pngwn:
I don't generally like having two APIs for the same thing, also the string shortcuts have caused issues as well as solved them (for example the conversation yesterday). Curious to hear others thoughts though.

I would expect the function to run every time the app is loaded, yes. So every user would get a new random value (rather than every user having the same random value).

abidlabs:
Unless I'm misunderstanding, functions wouldn't work. Any Python function would just get evaluated once, when the config is generated and all users would see the same random number. I suppose we could pass in a *javascript* function, which would get evaluated every time the page is refreshed, but that seems very messy.

I definitely see the case for flexibility, but I believe the vast, vast majority of the time, the desired behavior is either a fixed value or a uniformly random value. Additionally, I find the syntax in this PR very intuitive and readable:

```py
slider_1 = gr.Slider(value="random", minimum=100, maximum=200)
````

So I think this approach is nice syntactic sugar. And if a user wants a more flexible initial value, then they can use the other approach @freddyaboulton outlined in the issue: https://github.com/gradio-app/gradio/issues/1849#issuecomment-1191538451

**Comment by pngwn** [comment]
Why can't the function we pass be used as a callback that is evaluated at runtime? I'm really not a huge fan of stringly typed values, especially given that typehints don't really offer much granularity in that regard. Happy for it to go in as is, if we think it will work well over time.

**Comment by freddyaboulton** [comment]
@abidlabs @pngwn Just added the `randomize` parameter! This should be good for another look. Thank you for the feedback so far.

File: gradio/components.py [hunk 1]
```diff
@@ -664,21 +665,22 @@ class Slider(Changeable, IOComponent):
...
-    Demos: sentence_builder, generate_tone, titanic_survival
+    Demos: sentence_builder, generate_tone, titanic_survival, interface_random_slider, blocks_random_slider
     """
 
     def __init__(
         self,
         minimum: float = 0,
         maximum: float = 100,
-        value: Optional[float] = None,
+        value: Optional[float | str] = None,
```
abidlabs, line 675:
```suggestion
        value: Optional[float] = None,
```

gradio/components.py
```diff
@@ -692,6 +694,7 @@ def __init__(
             interactive: if True, slider will be adjustable; if False, adjusting will be disabled. If not provided, this is inferred based on whether the component is used as an input or output.
             visible: If False, component will be hidden.
             elem_id: An optional string that is assigned as the id of this component in the HTML DOM. Can be used for targeting CSS styles.
+            randomize: If True, the value of the slider when the app loads is taken uniformly at random from the range given by the minimum and maximum.
```
abidlabs, line 697: Can we update the docstring for `value` parameter to be:

"default value for slider. Ignored if `randomize=True`

gradio/components.py
```diff
@@ -723,6 +729,12 @@ def get_config(self):
             **IOComponent.get_config(self),
         }
 
+    def get_random_value(self):
+        return random.randrange(int(self.minimum), int(self.maximum))
```
abidlabs, line 733: Is there a reason we are restricting to integers? If the minimum is 0 and the maximum is 1, for example, a float would make more sense.

I think the right thing to do here is to do a `random.choice()` from amongst the possible values of the slider. Something like `value = random.choice(range(self.minimum, self.maximum, self.step))`

test/test_blocks.py
```diff
@@ -197,5 +197,25 @@ def test_integration_mlflow(self):
...
+        )
+        gr.Slider(
+            randomize=True, minimum=10, maximum=23.2, label="Random Slider (Input 2)"
+        )
+    for component in demo.blocks.values():
+        if "Non-random" in component.label:
+            assert not component.randomize
+        else:
+            assert component.randomize
+    assert all([dep["trigger"] == "load" for dep in demo.config["dependencies"]])
```
abidlabs, line 216: I think we should also confirm the existence of 2 `load()` dependencies

gradio/blocks.py
```diff
@@ -166,6 +168,12 @@ def get_config(self):
             "style": self._style,
         }
 
+    def get_random_value_on_load(self):
```
abidlabs, line 171: Thoughts on moving this to `IOComponent` since that should be subclassed by any component that has the concept of a value. (Note that `Block` is more general and includes as children classes like `Row` and `Tabs`)

gradio/blocks.py
```diff
@@ -1046,3 +1056,17 @@ def block_thread(
             self.server.close()
             if self.enable_queue:
                 queueing.close()
+
+    def configure_random_values(self):
+        """Add a load event for every component whose initial value should be randomized."""
+        for component in Context.root_block.blocks.values():
```
abidlabs, line 1062: So this approach works, but I think there's a better way that doesn't involve iterating over all of the blocks (many of whom don't need to know about the concept of randomization). 

What if we move this logic to the `render()` method of the `IOComponent` class? The `IOComponent.render()` would first call `super().render()`  and include this logic: if `component.randomize`, then look up the parent block (using `Context.root_block`) and create an event trigger

abidlabs:
Thanks @freddyaboulton for implementing this! I like the `randomize` parameter approach and it's cool that we can do entirely from the backend. 

I had some suggestions on the implementation side, see my comments above! Happy to discuss over a Huddle

gradio/components.py
freddyaboulton: Yea good point! This is trickier than I thought hehe.

The number of possible values can be huge so I don't think generating all possible values just to pick one is the best approach.

We might be better off generating a random multiple of step from the set of all possible multiples allowed by the min and max.

Something like this:

```python
def get_random_value(self,):
        # Goal is to generate a possible value of the slider
        # without generating all possible values
        # We randomly pick a step among all possible steps and add  to the minimum
        n_steps = int((self.maximum - self.minimum) / self.step)
        step = random.randint(0, n_steps)
        value = self.minimum + step * self.step
        return value
```

test/test_blocks.py
freddyaboulton: Agreed

gradio/blocks.py
freddyaboulton: I like it! Much more targeted change since it only applies to components that can take on values. 

Will give this a shot and let you know if I have questions!

aliabid94:
I also agree with @pngwn that this feels like a specialized implementation for something that should be more general. How about this: if the default value of any Component is set to a function / Callable, that function is called on load, separately for every session. E.g.

Slider
```python
gr.Slider(lambda: return random.randint(1,5), 1, 5)
```

Textbox
```
wordbank = [...]
gr.Textbox(lambda: return random.choice(wordbank), lines=3)
```

Feel ambivalent about adding syntactic sugar for the Slider specifically, I'm not sure how common of a usecase it is.

abidlabs:
@aliabid94 -- we discussed this over a huddle and we decided on a different approach: having a `randomize` parameter for components like `Slider()` (and maybe also `Dropdown()` and `Radio()`) which, if set to `True`, sets a uniformly random initial value to the component every time the demo web page is loaded. 

This avoids muddling the typing while also keeping the syntax pretty simple and readable.

aliabid94:
I think I'm okay with the `random` parameter if it's a common use case, I just think the root problem is not solved, in case the user wants even slightly customized logic of setting the initial value. Perhaps that can be another issue.

gradio/blocks.py
aliabid94: If we supported custom functions, it would be a simple change to swap out randomize to swap out `if component.randomize` for `if component.value isinstance Callable`. Or even support both, such as

```python
load_fn = None
if component.randomize:
  load_fn = component.get_random_value_on_load
elif component.value isinstance Callable:
  load_fn = component.value
if load_fn:
self.set_event_trigger(
      "load",
      load_fn,
      None,
      component,
      no_target=True,
      queue=False,
  )
```

aliabid94:
I very much agree with @pngwn's initial suggestion and I commented in the code how it would be quite easy to implement.

pngwn:
For posterity: Also mentioned this in the huddle and not relevant to this PR right now, but the concept of state would also solve this problem as you could initialise state to any value you wanted in any way you wanted (at runtime).

demo/kitchen_sink_random/run.py
```diff
@@ -0,0 +1,102 @@
...
+            )
+        ),
+        gr.Gallery(
+            value=lambda: [os.path.join(img_dir, img) for img in os.listdir(img_dir)]
+        ),
+        gr.Chatbot(
+            value=lambda: random.choice([[("hello", "hi!")], [("bye", "goodbye!")]])
+        ),
+        gr.Model3D(
+            value=lambda: random.choice(
```
freddyaboulton, line 91: For some reason Model3D is not being randomized on the front-end even though it is on the back-end.

My guess is that the model3d component is not being cleared?

```
/Users/freddy/sources/gradio/demo/kitchen_sink_random/../model3D/files/Bunny.obj
/Users/freddy/sources/gradio/demo/kitchen_sink_random/../model3D/files/Bunny.obj
/Users/freddy/sources/gradio/demo/kitchen_sink_random/../model3D/files/face.obj
/Users/freddy/sources/gradio/demo/kitchen_sink_random/../model3D/files/Bunny.obj
/Users/freddy/sources/gradio/demo/kitchen_sink_random/../model3D/files/face.obj
/Users/freddy/sources/gradio/demo/kitchen_sink_random/../model3D/files/Fox.gltf
```

Don't think that should hold up this PR though.

gradio/blocks.py
freddyaboulton: This is now implemented @aliabid94 @abidlabs

gradio/components.py
```diff
@@ -318,7 +336,8 @@ def __init__(
         self.lines = lines
         self.max_lines = max_lines
         self.placeholder = placeholder
-        self.value = self.postprocess(value)
+        load_fn, initial_value = self.get_load_fn_and_initial_value(value)
```
freddyaboulton, line 339: You'll see this pattern for getting the load_fn repeated in every IOComponent because the logic for setting the value is done in each component as opposed to the IOComponent parent class.

freddyaboulton:
@abidlabs @pngwn @aliabid94 This should be good for another pass. Still need to update the docstrings and such but hopefully this api is something we can all agree to 😄  

In short, every IOComponent can accept a callable as it's value, if value is a callable that function will set the value of the component whenever the app loads. Only for Slider did I implement the `randomize` flag since that's the only component so far where we see the need to provide a more concise syntax.

I implemented a random version of the kitchen sink where you can see this in action.

kitchen_sink_random

gradio/components.py
```diff
@@ -2906,8 +2973,11 @@ def __init__(
             visible: If False, component will be hidden.
             elem_id: An optional string that is assigned as the id of this component in the HTML DOM. Can be used for targeting CSS styles.
         """
-        Component.__init__(self, visible=visible, elem_id=elem_id, **kwargs)
-        self.value = value
+        load_fn, initial_value = self.get_load_fn_and_initial_value(value)
+        self.value = initial_value
+        IOComponent.__init__(
```
freddyaboulton, line 2978: We agreed Button should be an IOComponent now

abidlabs:
Looking at this now @freddyaboulton, great demos!

demo/kitchen_sink_random/constants.py
```diff
@@ -0,0 +1,72 @@
...
+]
+
+highlighted_text = "Does Chicago have any Pakistani restaurants"
+
+
+def random_model3d():
+    model_3d = random.choice(
+        [os.path.join(model3d_dir, model) for model in os.listdir(model3d_dir) if model != "source.txt"]
+    )
+    print(model_3d)
```
abidlabs, line 71:
```suggestion
```

gradio/blocks.py
```diff
@@ -1047,3 +1047,19 @@ def block_thread(
             self.server.close()
             if self.enable_queue:
                 queueing.close()
+
+    def attach_load_events(self):
+        """Add a load event for every component whose initial value should be randomized."""
+        from gradio.components import IOComponent
```
abidlabs, line 1053: Can you move the import to the top of the file? I'm guessing you might be experiencing circular imports though. 

Did you try `from gradio import components` and then check `isinstance(component, components.IOComponent)`

test/test_blocks.py
```diff
@@ -197,5 +198,55 @@ def test_integration_mlflow(self):
...
+        subclass = classes_to_check.pop()
+        children = subclass.__subclasses__()
+
+        if children:
+            classes_to_check.extend(children)
+        if "value" in inspect.signature(subclass).parameters:
+            subclasses.append(subclass)
+
+    interface = gr.Interface(
+        lambda x: x,
```
abidlabs, line 240: The # of parameters and return values in this lambda don't match the inputs and outputs. How about:
```suggestion
        lambda *args: None
```

test/test_blocks.py
```diff
@@ -197,5 +198,55 @@ def test_integration_mlflow(self):
...
+        inputs=[comp(value=lambda: None) for comp in subclasses],
+        outputs=None,
+    )
+
+    dependencies_on_load = [
+        dep for dep in interface.config["dependencies"] if dep["trigger"] == "load"
+    ]
+    assert len(dependencies_on_load) == len(subclasses)
```
abidlabs, line 248: Nice

test/test_blocks.py
```diff
@@ -197,5 +198,55 @@ def test_integration_mlflow(self):
...
+                assert component.attach_load_event
+    dependencies_on_load = [
+        dep["trigger"] == "load" for dep in demo.config["dependencies"]
+    ]
+    assert all(dependencies_on_load)
+    assert len(dependencies_on_load) == 2
+    assert not any([dep["queue"] for dep in demo.config["dependencies"]])
+
+
+def test_io_components_attach_load_events_when_value_is_fn():
```
abidlabs, line 226: Nice test

test/test_components.py
```diff
@@ -1829,5 +1829,20 @@ def test_gallery_save_and_restore_flagged(my_uuid, tmp_path):
...
+    slider = gr.Slider(minimum=-5, maximum=10, randomize=True)
+    assert slider.attach_load_event
+    assert slider.value == 7
+    assert slider.load_fn() == 7
+
+
+@patch("random.randint", return_value=3)
+def test_slider_rounds_when_using_default_randomizer(mock_randint):
+    slider = gr.Slider(minimum=0, maximum=1, randomize=True, step=0.1)
+    # 0 + 0.1 * 3 = 0.30000000000000004 in python so we test we round to the number of decimals of step
```
abidlabs, line 1843: Not sure I understand what this comment is saying

abidlabs:
Awesome @freddyaboulton, this is really good stuff! Left a couple of small points above, but basically ready to ship

abidlabs (approved):

gradio/blocks.py
freddyaboulton: Awesome, yea importing `components` works so I'll push that up!

test/test_blocks.py
freddyaboulton: Good call

PR #1949, opened 2022-08-04
Fix bug where you cannot cache examples with Interface.load

### LINKED ISSUE #1643: HF API + Spaces give error + uninformative error message
### Describe the bug

There is [this space](https://huggingface.co/spaces/osanseviero/testtumor) which uses the API under the hood and crashes with a super weird error. This works in a Colab. I concluded it is because cache_examples=True in Spaces, and the API is probably not ready so it of course crashes. If I set cache_examples=False it works fine, but it's not clear I should do that from the error message. I think here we should either improve the error message or not cache when using the API

### Reproduction

See code in https://huggingface.co/spaces/osanseviero/testtumor

### Screenshot

_No response_

### Logs

```shell
Fetching model from: https://huggingface.co/SerdarHelli/ThyroidTumorClassificationModel
/home/user/.local/lib/python3.8/site-packages/gradio/interface.py:286: UserWarning: Currently, only the 'default' theme is supported.
Cache at /home/user/app/gradio_cached_examples/log.csv not found. Caching now in 'gradio_cached_examples/' directory.
  File "app.py", line 23, in <module>
    gr.Interface.load("huggingface/SerdarHelli/ThyroidTumorClassificationModel",title=title,description=description,article=article,examples=examples,cache_examples=True).launch()
  File "/home/user/.local/lib/python3.8/site-packages/gradio/interface.py", line 90, in load
  File "/home/user/.local/lib/python3.8/site-packages/gradio/blocks.py", line 662, in load
  File "/home/user/.local/lib/python3.8/site-packages/gradio/external.py", line 40, in load_blocks_from_repo
  File "/home/user/.local/lib/python3.8/site-packages/gradio/external.py", line 294, in get_models_interface
    interface = gradio.Interface(**kwargs)
  File "/home/user/.local/lib/python3.8/site-packages/gradio/interface.py", line 420, in __init__
    cache_interface_examples(self)
  File "/home/user/.local/lib/python3.8/site-packages/gradio/process_examples.py", line 51, in cache_interface_examples
    raise e
  File "/home/user/.local/lib/python3.8/site-packages/gradio/process_examples.py", line 47, in cache_interface_examples
    prediction = process_example(interface, example_id)
  File "/home/user/.local/lib/python3.8/site-packages/gradio/process_examples.py", line 29, in process_example
    prediction = interface.process(raw_input)
  File "/home/user/.local/lib/python3.8/site-packages/gradio/interface.py", line 748, in process
    predictions = self.run_prediction(processed_input)
  File "/home/user/.local/lib/python3.8/site-packages/gradio/interface.py", line 712, in run_prediction
    prediction = predict_fn(*processed_input)
  File "/home/user/.local/lib/python3.8/site-packages/gradio/external.py", line 259, in query_huggingface_api
    data = pipeline["preprocess"](*params)
  File "/home/user/.local/lib/python3.8/site-packages/gradio/external.py", line 131, in <lambda>
    i.split(",")[1]
```
```


### System Info

```shell
Gradio 3.20, Spaces
```

### Severity

annoying

freddyaboulton:
# Description
My solution is to set api_mode=True in the interface during intialization as opposed to after initialization.

This is done via a kwarg called `_api_mode` that is not documented in the constructor since we don't want users to directly set `api_mode`. 

Outside `external.py`, `Interface.api_mode` will still be set to False by default.

Fixes #1643

How to test:
Unit test: `pytest test/test_external.py::test_interface_load_cache_examples`

Run the following demos with cache_examples=True

```
Model3D/run
fake_gan
color_pickers
sts_or_tts
image_classifer_interface_load
```

# Checklist:


demo/image_classifier_interface_load/run.py
```diff
@@ -0,0 +1,29 @@
...
+
+images = ["cheetah1.jpeg", "cheetah1.jpg", "lion.jpg"]
+
+
+img_classifier = gr.Interface.load(
+    "models/google/vit-base-patch16-224", examples=images, cache_examples=True
+)
+
+
+def func(img, text):
```
freddyaboulton, line 11: Demo to test using interface as function still works and that cache examples=True works now

test/test_external.py
```diff
@@ -229,5 +231,15 @@ def test_text_to_text_model_from_pipeline(self):
         self.assertIsNotNone(output)
 
 
+def test_interface_load_cache_examples(tmp_path):
```
freddyaboulton, line 234: Ideally this test would not make an actual request but the way the external.py module is set up makes it really hard to mock out the request. `query_huggingface_api` is defined inside a function which makes it impossible to mock.

gradio/deprecation.py
```diff
@@ -38,6 +38,8 @@ def check_deprecated_parameters(cls: str, **kwargs) -> None:
             kwargs.pop(key)
             # Interestingly, using DeprecationWarning causes warning to not appear.
             warnings.warn(value)
+    if "_api_mode" in kwargs:
```
freddyaboulton, line 41: So that we do not get warnings about un-used kwargs. I also don't want to use `api_mode` since that would raise a warning about `api_mode` being deprecated so I call the parameter `_api_mode`.

abidlabs:
Ready for a review @freddyaboulton?

abidlabs:
Taking a look!

gradio/interface.py
```diff
@@ -274,7 +274,7 @@ def __init__(
         else:
             raise ValueError("Invalid value for parameter: interpretation")
 
-        self.api_mode = False
+        self.api_mode = kwargs.get("_api_mode", False)
```
abidlabs, line 277: Would it be better to include `_api_mode` as an explicit but undocumented parameter in `Interface` with a default value of `False`? Better to have all of the parameters and their defaults in one place

gradio/interface.py
```diff
@@ -668,7 +668,7 @@ def run_prediction(
         if prediction is None or len(self.output_components) == 1:
             prediction = [prediction]
 
-        if self.api_mode:  # Deerialize the input
+        if self.api_mode:  # Deserialize the input
```
abidlabs, line 671: Good catch ;)

abidlabs:
Nice fix @freddyaboulton and test! This LGTM -- only one question on whether to explicitly define the `_api_mode` parameter.

abidlabs (approved):

gradio/interface.py
freddyaboulton: Yea it's better to be explicit. I just didn't want to add something resembling `api_mode` to the signature since it was removed previously. Just pushed this up and will merge! 

Thank you for the review.

PR #1950, opened 2022-08-04
implement queue frontend

pngwn:
This works for the most part although there is some strangeness around the status indicator that I've been able to fully resolve. I think some of the smaller PRs going into the base branch may have implications for the frontend, so it would be good to get those in and retest this.


aliabid94:
LGTM, some nits but will merge and fix from release-new-queue-beta branch

Issue #1941, opened 2022-08-03
TabbedInterface will ignore all style setup of block api


HuangChiEn:
#### python version == 3.8.8

There are 2 way to setup the style of each component in the block API. 
1. Feed the CSS string in the block API and setup `elem_id` of component according to CSS template.
> ``` 
> # my code snippet
> css_def = '''
> #im_id {height: 30rem !important; width: 40% !important;}
> #txt_id {width: 40% !important;}
> #sld_id {width: 20% !important;}
> '''
> blk_cntr = gr.Blocks(css=css_def) 

2. call style method in the corresponding component
> `image_input = gr.Image(label="Uploaded image").style(height='24')`

However, all of aforementioned settings will be fail, while i further warp the block API into the TabbedInterface.

```
    def layout_func(blk_cntr):
        with blk_cntr:
            ... all of component i want to setup
            # e.g. Both not work while apply TabbedInterface...QQ
            # method 1.
            # image_input = gr.Image(label="Uploaded image", elem_id='im_id')
            # method 2.
            # image_input = gr.Image(label="Uploaded image").style(height='24')
            
    
    css_def = '''
    #im_id {height: 30rem !important; width: 40% !important;}
    #txt_id {width: 40% !important;}
    #sld_id {width: 20% !important;}
    blk_cntr = gr.Blocks(css=css_def) 
    layout_func(blk_cntr)
    ....
    # for convenience, i just use the same block API (blk_cntr) 
    demo = gr.TabbedInterface([blk_cntr, blk_cntr], ["Text-to-speech", "Speech-to-text"])
    demo.launch(share=True, enable_queue=True)
```

Any suggestion will be appreciated !!

abidlabs:
Hi @HuangChiEn I'm looking into this. Would you be able to provide a complete code snippet? Will make it easier to reproduce

abidlabs:
Hi @HuangChiEn, this should now be fixed due to #1951. Please take a look at https://github.com/gradio-app/gradio/pull/1951 for more information, but the summary is that you will have to pass in a `css` parameter to `TabbedInterface` with your desired css, like this:

```py
import gradio as gr

def test(text1):
    return [(text1, "good"), ("\n", None), (text1, "good")]

demo = gr.Interface(
    test,
    "text",
    gr.HighlightedText(elem_id="htext"),
)

def test(text1):
    return [(text1, "good"), ("\n", None), (text1, "good")]

demo2 = gr.Interface(
    test,
    "text",
[...]

Issue #1839, opened 2022-07-20
Can't add newline when using Highlighted Context


dyxohjl666:
The newlines don’t seem to work when I try to split the multiple results. I realized it maybe due to escaping but have no idea how to solve the problem.
What can I do for that?  Or I have to switch to a html output?

aliabid94:
Thanks for letting us know about this. Will take a look!

aliabid94:
Let me know if this works for you, or if you have any questions!

dyxohjl666:
Thanks for your reply! 
It looks quite perfect now when I apply a single Interface component.
But I met the same problem when add the Interface into TabbedInterface. It seems like that the CSS is invalid when in TabbedInterface. I'm confused about that.

dyxohjl666:
I generate the Interface like this:
```
import gradio as gr

parse_str = gr.Interface(
    fn=pred_str,
    inputs=gr.Textbox(
        label="String",
        lines=5,
    ),
    outputs=gr.HighlightedText(
        label="The Result of Parsing",
        combine_adjacent=True,
        adjacent_separator=" ",
    )
)

parse_file = gr.Interface(
    fn=pred_file,
    inputs="file",
    outputs=gr.HighlightedText(
        elem_id="htext",
        label="The Result of Parsing",
        combine_adjacent=True,
        adjacent_separator=" ",
    ),
    css="#htext span {white-space: pre-line}",
)
demo = gr.TabbedInterface([parse_str, parse_file], ["Parsing reference string", "Parsing reference for PDF"])

parse_file.launch(share=True)
```

aliabid94:
Ahh hmm seems like a bug where custom CSS isn't passed up in a TabbedInterface. @abidlabs can you take a look?

freddyaboulton:
This is a dupe of #1683. Going to keep both open so as to not lose all the discussion and add to 3.2 milestone.

HuangChiEn:
> Hi @dyxohjl666, this is now fixed. See #1951 for more information, but the summary is that you will have to pass in a `css` parameter to `TabbedInterface` with your desired css, like this:
> 
> ```python
> import gradio as gr
> 
> def test(text1):
>     return [(text1, "good"), ("\n", None), (text1, "good")]
> 
> demo = gr.Interface(
>     test,
>     "text",
>     gr.HighlightedText(elem_id="htext"),
> )
> 
> def test(text1):
>     return [(text1, "good"), ("\n", None), (text1, "good")]
[...]

Issue #1683, opened 2022-07-01
Honour line breaks in HighlightedText


versae:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
When using `HighlightedText`, line breaks are usually ignored. This is particularly annoying for showing GPT generated text, and even more so for poetry.

**Describe the solution you'd like**  
An option to replace `\n` with `<br>` so the resulting HTML is rendered correctly. If I try to write the replace myself it gets escaped and it never renders the actual line break but the characters in the string `<br>`.

**Additional context**  
Try with the next text
```
Caminante, son tus huellas
el camino y nada más;
Caminante, no hay camino,
se hace camino al andar.
```
In the demo here: https://huggingface.co/spaces/bertin-project/bertin-gpt-j-6B (press the `Generar` button)

abidlabs:
Thanks for raising the issue @versae! @aliabid94 is there a workaround for this already or do we need to add support for new lines specifically?

abidlabs:
Hi @versae, you can do this by passing in a little bit of css to your `Interface` or `Blocks`, like this:

```py
import gradio as gr

def test(text1):
    return [(text1, "good")]

demo = gr.Interface(
    test,
    "text",
    gr.HighlightedText(elem_id="htext"),
    css="#htext span {white-space: pre}",
)
demo.launch()
```

See #1951 and #1839 for more info.

versae:
Interesting. Does this work with the Blocks API as well?

abidlabs:
Yup, you can pass `css` into `Blocks` as well.

versae:
Awesome!

versae:
OK, I finally had the chance to try this but now the lines are not breaking at the length of their container. Any idea why?
![image](https://user-images.githubusercontent.com/173537/190135881-de2c1848-40c8-4f64-bf9a-032babc04c6d.png)

aliabid94:
hmm will take a look

PR #1951, opened 2022-08-05
Fixes to `TabbedInterface`

abidlabs:
Adds a few properties to the `TabbedInterface`, most notably support for `css` passed into the constructor. 

I considered whether the `TabbedInterface` should accept its own `css` parameter, or whether it should simply "concatenate" the css stylesheets from its list of Interfaces, and I went with the former for a couple of reasons:

- There might be conflicts between Interface css style sheets. For example, one Interface might set a certain background color, while another Interface might set another. If we were to simply concatenate, both of the Interfaces would end up with the same background color. This forces the user to set proper CSS with element IDs instead of being surprised by eventual behavior.
- This is consistent with how Interfaces behave within Blocks (or even how Blocks behave within Blocks). If an Interface (or Blocks) is defined inside of a Blocks, it is rendered, but its `css` parameter is completely ignored. 

We can change this behavior by automatically scoping css (better, but a lot of work) or concatenating css (easy but will lead to surprises), but we should apply that behavior consistently within Blocks.

Provides a solution for #1839 and #1941


abidlabs:
For example, for #1839, the user can do something like this:

```py
import gradio as gr

def test(text1):
    return [(text1, "good"), ("\n", None), (text1, "good")]

demo = gr.Interface(
    test,
    "text",
    gr.HighlightedText(elem_id="htext"),
)

def test(text1):
    return [(text1, "good"), ("\n", None), (text1, "good")]

demo2 = gr.Interface(
    test,
    "text",
    gr.HighlightedText(elem_id="htext2"),
)

gr.TabbedInterface([demo, demo2], css="#htext span {white-space: pre} #htext2 span {white-space: pre}").launch()
```

gradio/interface.py
```diff
@@ -735,18 +736,35 @@ class TabbedInterface(Blocks):
...
+        title: str = "Gradio",
+        css: Optional[str] = None,
     ):
         """
         Parameters:
             interface_list: a list of interfaces to be rendered in tabs.
             tab_names: a list of tab names. If None, the tab names will be "Tab 1", "Tab 2", etc.
+            theme: which theme to use - right now, only "default" is supported.
+            analytics_enabled: whether to allow basic telemetry. If None, will use GRADIO_ANALYTICS_ENABLED environment variable or default to True.
+            mode: a human-friendly name for the kind of Blocks interface being created.
```
freddyaboulton, line 753: We should remove mode from the docstring since it's always "tabbed_interface"

gradio/interface.py
```diff
@@ -735,18 +736,35 @@ class TabbedInterface(Blocks):
...
+            title: The tab title to display when this is opened in a browser window.
+            css: custom css or path to custom css file to apply to entire Blocks
         Returns:
             a Gradio Tabbed Interface for the given interfaces
         """
+        super().__init__(
+            theme=theme,
+            analytics_enabled=analytics_enabled,
+            mode="tabbed_interface",
+            title=title,
```
freddyaboulton, line 763: The UI won't display the title so maybe we shouldn't add it to the init until it can? I feel like it will confuse users.

```python
import gradio as gr

def test(text1):
    return [(text1, "good"), ("\n", None), (text1, "good")]

demo = gr.Interface(
    test,
    "text",
    gr.HighlightedText(elem_id="htext"),
    title="Tab 1",
)

def test(text1):
    return [(text1, "good"), ("\n", None), (text1, "good")]

demo2 = gr.Interface(
    test,
    "text",
    gr.HighlightedText(elem_id="htext2"),
    title="Tab 2",
)

gr.TabbedInterface([demo, demo2], css="#htext span {white-space: pre} #htext2 span {white-space: pre}",
                   title="Tabbed Interface").launch()
```

image


The title is set properly in the config though so I think it's just a front-end issue

freddyaboulton (approved):

I agree that concatenating css is not the best approach.


Fixes: #1941, Fixes: #1683, Fixes: #1839

gradio/interface.py
abidlabs: My bad, thanks

gradio/interface.py
abidlabs: Ah so the title here just sets the Tab title. (It's more similar to a Blocks title as opposed to the Interface title). But I agree it might lead to confusion so will remove.

abidlabs:
Thanks @freddyaboulton for the review! Agree with all of the suggestions.

I'll merge this in and respond to the issues individually.

HuangChiEn:
Thanks for you'll quick reply. However, may i reopen #1941 due to the CSS argument is not supported in gradio==3.1.3 version.
image

Will it be released it in the next version (3.1.4) or i should build it from source code to catch up the updated commit [f46e95d](https://github.com/gradio-app/gradio/commit/f46e95d7e0b6d901e65d3c85cd24c56c7554abf4) ?

Phillweston:
TabbedInterface Error: the 'clear' and 'submit' of the second tab will operate the first tab

See also: #8425

emptystack1024:
solution：
See also: https://github.com/gradio-app/gradio/issues/8425
Thanks, @ Phillweston

Issue #1948, opened 2022-08-04
gr.Examples with image and text don't make a proper table when used with Blocks


apolinario:
### Describe the bug

If using blocks, when I have this code block
```python
input_img = gr.Image()
input_prompt = gr.Textbox()
gr.Examples([["kitty.jpg", "a kitten running around"]],[input_img,input_prompt])
```
I get only the first element, so just the image
![image](https://user-images.githubusercontent.com/788417/182839895-489d9fbb-6298-4848-93c6-b215fc4e76fb.png)

However, if I do a `gr.Examples` using gr.Interfaces it works as it was supposed to: 

![image](https://user-images.githubusercontent.com/788417/182839681-f817faee-6ea2-422e-84fd-2940db4f4167.png)

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

With gr.Blocks(), where you can see it failing: 
```python
import gradio as gr
       
with gr.Blocks() as demo:
            with gr.Row():
                with gr.Column():
                    input_img = gr.Image(label="Original", type="file", interactive=False, value="kitty.jpg")
                    input_prompt = gr.Textbox()
                    gr.Examples([["kitty.jpg", "a kitten running around"]],[input_img,input_prompt])
                
demo.launch()
```
![image](https://user-images.githubusercontent.com/788417/182839895-489d9fbb-6298-4848-93c6-b215fc4e76fb.png)

With gr.Interface(), where you can see it working:
```python
import gradio as gr 

def predict():
    pass

image = gr.inputs.Image()
text = gr.inputs.Textbox()

gr.Interface(
    fn=predict,
    inputs=[image, text],
    outputs=gr.outputs.Image(),
    examples=[['kitty.jpg','a kitten running around']]).launch()
```
![image](https://user-images.githubusercontent.com/788417/182839681-f817faee-6ea2-422e-84fd-2940db4f4167.png)

### System Info

```shell
Gradio 3.1.3
```

### Severity

annoying

dawoodkhan82:
@apolinario thanks for reporting. I can confirm this is an issue and can reproduce it. @aliabid94 Ik you worked on examples, can you take a look?

abidlabs:
I can take this, as I'm also looking into #1875

abidlabs:
Hi @apolinario, looks like the issue actually stems from the fact that some of the components in your Blocks demo have labels while others don't. If both the `Image` and `Textbox` components have labels (or neither do), then the issue does not occur.

[...]

Issue #1875, opened 2022-07-25
Issue with Example helper component when passing str path to directory of examples


shivalikasingh95:
Hi,

When I'm trying to pass a string path to the directory of samples to the "Example" helper component, I'm getting a file not found error as shown in the screenshot labelled "Error 1 colab screenshot" below.

I tried fixing the above error by fixing the path being passed to 'processing_utils.encode_file_to_base64()'  in line 1637 of gradio/components.py
However, after fixing that I was getting another error as shown in screenshot labelled "Error 2'. 

I was able to fix all the errors and get the desired output by commenting out lines 92-103 of gradio/examples.py as highlighted in screenshot labelled "Problem Fix".
Since, I'm new to gradio I'm not sure if this is the relevant fix. 

Please let me know if you guys are able to reproduce this issue and if the fix suggested on top is right. I'd be happy to submit a PR for the same. If there's an alternative fix that can be done for this issue, I'd be happy to take that up as well.

**Error 1 colab screenshot**:

![Screenshot from 2022-07-28 00-59-02](https://user-images.githubusercontent.com/73357305/181356237-ac72940d-918c-417e-991c-03f749c4aa72.png)

**Error 1 Stack trace:**
![Screenshot from 2022-07-25 07-53-47](https://user-images.githubusercontent.com/73357305/180681083-924c528f-0e45-4322-ac1f-dc766ae35356.png)

**Error 2:**
![Screenshot from 2022-07-25 08-37-31](https://user-images.githubusercontent.com/73357305/180691342-085c9b6c-c911-41a9-ad05-87884b381c35.png)

**Problem Fix:** Commenting lines 92-103 of gradio/examples.py 
![Screenshot from 2022-07-25 08-31-44](https://user-images.githubusercontent.com/73357305/180692228-f1de0d9d-080d-477d-ab14-7bd19ed48ad2.png)

**The following code snippet should help in reproducing Error 1:**

` import gradio as gr

  demo = gr.Blocks()
  input_image = gr.Image(type='filepath',label="Input Image", show_label=True)
  output_label = gr.Label(label="Model Output", show_label=True)
  
  with demo:
         gr.Examples("examples", [input_image], output_label) #where "examples" is path of directory containing sample images.
  
  demo.launch()`

**System Info:**
gradio version: 3.1.1
OS: Ubuntu 22.04 LTS
Browser: Chrome

shivalikasingh95:
Hi any updates on this issue?

[...]

PR #1952, opened 2022-08-05
Fixes to `gr.Examples`

abidlabs:
This PR fixes some issues with `gr.Examples`:

* The Examples dataset component would not appear correctly if only some of the input components had labels. This is now fixed, with tests to ensure the correct behavior. (Fixes: #1948)
* When examples were read from a log file, they were not loaded correctly because of relative file imports. This is now fixed, with tests to ensure the correct behavior. (Fixes: #1875)
* Additional tests are added for `gr.Examples`


freddyaboulton:
@abidlabs Is this ready for review? I see it's still Draft and the tests are failing

abidlabs:
Ready now for a review @freddyaboulton, thank you!

gradio/examples.py
```diff
@@ -225,6 +222,7 @@ def load_from_cache(self, example_id: int) -> List[Any]:
         example = examples[example_id + 1]  # +1 to adjust for header
         output = []
         for component, cell in zip(self.outputs, example):
+            print("self.cached_folder", self.cached_folder, "cell", cell)
```
abidlabs, line 225:
```suggestion
```

abidlabs:
As discussed with @freddyaboulton, I don't love the fact that we change directories to read in the examples. Basically the issue is that paths of samples in the log.csv file are relative to the folder containing the log.csv file (call this the _log path_). So we basically have 3 options:

1. Change directories when loading the samples
1. Append the log path to all paths when we read in the csv file (not obvious to know which samples in the log file paths vs. strings, etc.)
1. Change the preprocess_example to also accept in a directory

I feel like changing directories is the best option, but definitely open to discussing (maybe #3 is better)

abidlabs:
We can make #1 more robust by using context managers to wrap the change in directory, e.g. https://dev.to/teckert/changing-directory-with-a-python-context-manager-2bj8

abidlabs:
I've implemented the changing of the working directory with context managers instead, I think it's more readable and robust now.

freddyaboulton (approved):
Thank you for the fix, using the context manager approach and adding the unit tests @abidlabs ! 

This is good to merge

abidlabs:
Thanks for your review and suggestions @freddyaboulton! Will fix and merge

Issue #1962, opened 2022-08-05
Microphone and Audio not working 


aliabd:
### Describe the bug

Recording from microphone or clicking an audio example doesn't seem to be working in the latest version of gradio. 

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

This space: https://huggingface.co/spaces/gradio/same-person-or-different with the latest version of gradio. 

### Logs

```shell
Traceback (most recent call last):
  File "/Users/aliabdalla/anaconda3/envs/new/lib/python3.8/site-packages/gradio/routes.py", line 255, in run_predict
    output = await app.blocks.process_api(
  File "/Users/aliabdalla/anaconda3/envs/new/lib/python3.8/site-packages/gradio/blocks.py", line 630, in process_api
    predictions, duration = await self.call_function(fn_index, processed_input)
  File "/Users/aliabdalla/anaconda3/envs/new/lib/python3.8/site-packages/gradio/blocks.py", line 545, in call_function
    prediction = await anyio.to_thread.run_sync(
  File "/Users/aliabdalla/anaconda3/envs/new/lib/python3.8/site-packages/anyio/to_thread.py", line 31, in run_sync
    return await get_asynclib().run_sync_in_worker_thread(
  File "/Users/aliabdalla/anaconda3/envs/new/lib/python3.8/site-packages/anyio/_backends/_asyncio.py", line 937, in run_sync_in_worker_thread
    return await future
  File "/Users/aliabdalla/anaconda3/envs/new/lib/python3.8/site-packages/anyio/_backends/_asyncio.py", line 867, in run
    result = context.run(func, *args)
  File "/Users/aliabdalla/anaconda3/envs/new/lib/python3.8/site-packages/gradio/interface.py", line 489, in <lambda>
    lambda *args: self.run_prediction(args)[0]
  File "/Users/aliabdalla/anaconda3/envs/new/lib/python3.8/site-packages/gradio/interface.py", line 669, in run_prediction
    prediction = self.fn(*processed_input)
  File "app.py", line 52, in similarity_fn
    wav1, _ = apply_effects_file(path1, EFFECTS)
  File "/Users/aliabdalla/anaconda3/envs/new/lib/python3.8/site-packages/torchaudio/sox_effects/sox_effects.py", line 285, in apply_effects_file
    raise RuntimeError("Failed to load audio from {}".format(path))
[...]

PR #1963, opened 2022-08-05
Remove broken audio demo from landing page

aliabd:
Removes the audio demo from the landing page since it isn't working. We should bring it back when #1962 is resolved or this [space](https://huggingface.co/spaces/gradio/same-person-or-different) works.


abidlabs:
LGTM @aliabd

abidlabs (approved):

Issue #1194, opened 2022-05-09
Better error message when format of `examples` is incompatible with `inputs` 


merveenoyan:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

Hello the greatest OS devs on 🌍 👋

I'm playing with the dataframe component and saw that examples are only passed as lists, so I converted the dataframe into list and managed to break the interface. It applies for any list. 
Broken interface looks like below:
![Ekran Resmi 2022-05-09 15 05 18](https://user-images.githubusercontent.com/53175384/167416036-ab3df9e9-9540-43b5-8aa8-7beb5b313cfa.png)
I removed the example and it worked as usual. 

### Reproduction

You can run this [kaggle notebook ](https://www.kaggle.com/code/unofficialmerve/notebookf190d598a4/edit)or simply doing:

```
import gradio as gr
df = gr.Dataframe(interactive=True, headers=["1","2","3"], type = "pandas")
outcome = gr.outputs.Dataframe(headers = ["Diabetes"])
iface = gr.Interface(inference_function, inputs = df, examples = [1,2,3,4,5,6,7,8],   outputs = outcome)
iface.launch(share=True, debug=True)
```
should work. 

### Logs

There's no logs even when I set `debug = True.`

### System Info

```shell
gradio==2.9b25
```

### Severity

annoyance

abidlabs:
@merveenoyan not sure I understand the issue:

```py
import gradio as gr
df = gr.Dataframe(interactive=True, headers=["1","2","3"], type = "pandas")
outcome = gr.outputs.Dataframe(headers = ["Diabetes"])
iface = gr.Interface(inference_function, inputs = df, examples = [1,2,3,4,5,6,7,8],   outputs = outcome)
iface.launch(share=True, debug=True)
```

Here, the examples that are provided don't line up with the headers that were provided, so there should be an error that is thrown. What is the behavior that you were expecting?

merveenoyan:
@abidlabs I'd rather have an error thrown, there's nothing that gives me a signal that this shouldn't be the way to create this Block. (like we do in interface)

abidlabs:
Agreed, we should catch this in the backend and throw a better error message

taimaruuu:
would love to help resolve this issue!

t-vi:
[...]

Issue #1721, opened 2022-07-06
Issues with inpainting image input


abidlabs:
### Describe the bug

This demo uses the new image input (`gr.Image(tool="sketch", label="Input",type="numpy")`): https://huggingface.co/spaces/akhaliq/lama

However, when examples are added to the demo, clicking on the examples does not populate the inpainting image input. Instead, you get a blank gray rectangle there:

![image](https://user-images.githubusercontent.com/1778297/177636336-9a688d62-cdfe-4ec0-8238-eb55dc4f50c2.png)

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

https://huggingface.co/spaces/akhaliq/lama

### Logs

```shell
No errors
```

### System Info

```shell
3.0.22
```

### Severity

serious, but I can work around it

drscotthawley:
Finding that manual mask drawing is not supported for almost any uploaded image for the LAMA model https://huggingface.co/spaces/akhaliq/lama

  Only the default image works.  This is using Brave Browser (Chromium... [Version 1.37.116 Chromium: 100.0.4896.127 (Official Build) (64-bit)](https://brave.com/latest/) ).

For square uploaded images, I notice that the white circle showing the mask drawing appears about 150 pixels down and to the right of the mouse pointer. 

For most images, however, be they square or rectangular, simply no mask drawing is possible.

abidlabs:
Thanks @drscotthawley, we'll take a look!

AK391:
this might be related to setting a default image using gr.Image(value="image.png"), after removing that works on the images shown above
<img width="804" alt="Screen Shot 2022-07-06 at 10 43 18 PM" src="https://user-images.githubusercontent.com/81195143/177679233-4e238902-32fa-4ab2-a65b-287f01153470.png">

pngwn:
This is a scaling issue with very large images, as I understand it. The image _and UI_ is downscaled when the image is very large. We need to scale the image independent of the UI and scale the painted region to the size of the image in order for it to appear correctly.

pngwn:
Pretty sure I'm wrong, these images aren't even that big. Investigating.

pngwn:
The default image issue also affected using the sketch as an output, the simple reason for this is we simple didn't support that use case, which is a huge oversight and a bug. This is easily addressed.

[...]

Issue #1325, opened 2022-05-18
Try It Code Snippets in Predict API Docs (Copy-Pasteable)


nbroad1881:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

The copy-paste code snippets in the "Try it (live demo)" section of the Predict API Docs are unusable because of the formatting.

Here are the python, curl, and js snippets copy-pasted (from https://hf.space/embed/akhaliq/FaceMesh/api)

(I deleted a lot of characters from the image string to make it easier to look at)
```python
import requests




r = requests.post(url='https://hf.space/embed/akhaliq/FaceMesh/+/api/predict/

',

 

json={"data":

 ["data:image/jpeg;base64,/9j/4AAQSkZJRgABAQAAAQABAAD/2wBDAAUDBAQEAwUEBAQFBQUGBwwIBwcHBw8LCwkMEQ8SEhEPERETFhwXExQaFRERGCEYGh0dHx8fExciJCIeJBweHx7/2wBDAQUFBQcGBw4ICA4eFBEUHh4eHh4eHh4eHh4eH"

]})

r.json()
```

```sh
curl -X POST 

https://hf.space/embed/akhaliq/FaceMesh/+/api/predict/ -H 'Content-Type: application/json' -d '{"data": ["data:image/jpeg;base64,/9j/4AAQSkZJRgABAQAAAQABAAD/2wBDAAUDBAQEAwUEBAQFBQUGBwwIBwcHBw8LCwkMEQ8SEhEPERETFh"

]}'
```

```javascript
fetch('https://hf.space/embed/akhaliq/FaceMesh/+/api/predict/', { method: "POST", body: JSON.stringify({"data":[ "data:image/jpeg;base64,/9j/4AAQSkZJRgABAQAAAQABAAD/2wBDAAUDBAQEAwUEBAQFBQUGBwwIBwcHBw8LCwkMEQ8SEhEPERETFhwXExQaFRERGCEYGh0dHx8fExciJCIeJBweHx7/2wBDAQUFBQcGBw4ICA4eFBEUHh4eHh4eHh4eHh4eHh4eHh4eHh4eHh4eHh4eHh4eHh4eHh4eHh4eHh4eHh4eHh4eHh7/wAARCAIAAgADASIAAhEBAxEB/8QAHAAAAQUBAQEAAAAAAAAAAAAA"

[...]

PR #1967, opened 2022-08-07
Cleaning up the way data is processed for components

abidlabs:
There is a lot of redundant data processing code for each `Component` (more details below), which is problematic because: (1) bugs are introduced when some parts of the code are updated, but other parts are not (2) it's hard to understand how the data is preprocessed / postprocessed, thereby making it harder to understand the existing codebase as well as to contribute new components.

This PR eliminates the following redundancies and along the ways, fixes: #1194, #1721 

1. `preprocess_example()` being a special case of `postprocess()` for most components -- both functions takes a Python object (e.g. PIL image or path to image file) and serializes it (e.g. to base64 representation) so that it can be viewed in the browser. However, for file-based components, we do not serialize examples, but rather just include a reference to them, so this special case needs to be handled. 
2. `serialize()` and `deserialize()` being part of `postprocess()` and `preprocess()` respectively.

### DIFF
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -201,12 +201,6 @@ class IOComponent(Component):
         """
         return x
 
-    def preprocess_example(self, x: Any) -> Any:
-        """
-        Any preprocessing needed to be performed on an example before being passed to the main function.
-        """
-        return x
-
     def set_interpret_parameters(self):
         """
         Set any parameters for interpretation.
@@ -405,15 +399,6 @@ class Textbox(Changeable, Submittable, IOComponent):
             called_directly: if True, the interface was called(), otherwise, it is being used via the GUI
         """
         return x
-
-    def preprocess_example(self, x: str | None) -> Any:
-        """
-        Any preprocessing needed to be performed on an example before being passed to the main function.
-        """
-        if x is None:
-            return None
-        else:
-            return str(x)
 
     def set_interpret_parameters(
         self, separator: str = " ", replacement: Optional[str] = None
--- a/gradio/examples.py
+++ b/gradio/examples.py
@@ -15,7 +15,7 @@ from gradio.flagging import CSVLogger
 
 if TYPE_CHECKING:  # Only import for type checking (to avoid circular imports).
     from gradio import Interface
-    from gradio.components import Component
+    from gradio.components import IOComponent
 
 CACHED_FOLDER = "gradio_cached_examples"
 
@@ -37,8 +37,8 @@ class Examples:
     def __init__(
         self,
         examples: List[Any] | List[List[Any]] | str,
-        inputs: Component | List[Component],
-        outputs: Optional[Component | List[Component]] = None,
+        inputs: IOComponent | List[IOComponent],
+        outputs: Optional[IOComponent | List[IOComponent]] = None,
         fn: Optional[Callable] = None,
         cache_examples: bool = False,
         examples_per_page: int = 10,
@@ -143,7 +143,7 @@ class Examples:
 
         self.processed_examples = [
             [
-                component.preprocess_example(sample)
+                component.postprocess(sample)
                 for component, sample in zip(inputs_with_examples, example)
             ]
             for example in non_none_examples


abidlabs:
3. We might also standardize the return types of all of the components to be in the format {`data`:data, ...}

demo/interface_random_slider/run.py
```diff
@@ -1,7 +1,7 @@
 import gradio as gr
 
 
-def func(slider_1, slider_2):
+def func(slider_1, slider_2, *args):
```
abidlabs, line 4: Fixed this demo.

demo/outbreak_forecast/run.py
```diff
@@ -42,15 +39,10 @@ def outbreak(plot_type, r, month, countries, social_distancing):
                    yaxis_title="Days Since Day 0")
         return fig
     else:
-        source = ColumnDataSource(df)
-        p = bk.figure(title="Outbreak in " + month, x_axis_label="Cases", y_axis_label="Days Since Day 0")
-        for country in countries:
-            p.line(x='day', y=country, line_width=2, source=source)
-        item_text = json_item(p, "plotDiv")
-        return item_text
+        raise ValueError("A plot type must be selected")
```
abidlabs, line 42: Removed `bokeh` from this demo as we don't support it yet.

gradio/blocks.py
```diff
@@ -361,6 +363,7 @@ def __init__(
         self.auth = None
         self.dev_mode = True
         self.app_id = random.getrandbits(64)
+        self.temp_dir = tempfile.TemporaryDirectory().name
```
abidlabs, line 366: One big change is that we now don't serialize output files (audio, videos, etc.) to base64 format to send them to the frontend. Instead, we save the files to a temporary directory and simply return the file link (linking is faster than serialization and this approach consolidates code). The `self.temp_dir` the location for these temporary files for a given Blocks.

gradio/components.py
```diff
@@ -3450,7 +2970,7 @@ def style(
 
 
 @document("change", "style")
-class JSON(Changeable, IOComponent):
+class JSON(Changeable, IOComponent, JSONSerializable):
```
abidlabs, line 2973: This introduces a small change in behavior. Previously, flagged JSON was simply stringified and stored in a CSV. But JSON should not be in a CSV. Instead, this saves the JSON to a file and links to the file in the CSV, similar to how we handle images, videos, etc.

gradio/components.py
```diff
@@ -2686,6 +2286,8 @@ def postprocess(
             JSON object with key 'headers' for list of header names, 'data' for 2D array of string or numeric data
         """
         if y is None:
+            return self.postprocess(self.test_input)
```
abidlabs, line 2289: Not totally sure if this makes sense, but this is kinda what we were doing before

gradio/examples.py
```diff
@@ -217,40 +220,35 @@ async def cache_interface_examples(self) -> None:
...
-            for i, example in enumerate(example_set)
-        ]
-        processed_input = [
-            input_component.preprocess(raw_input[i])
-            for i, input_component in enumerate(self.inputs)
-        ]
+        processed_input = self.processed_examples[example_id]
+        if not self._api_mode:
+            processed_input = [
+                input_component.preprocess(processed_input[i])
```
freddyaboulton, line 237: I think `enumerate(self.inputs)` assumes all inputs have examples provided, which is not always true. I think this is why `fake_gan` is failing on spaces.

demo/kitchen_sink/run.py
```diff
@@ -155,6 +154,7 @@ def fn(
     cache_examples=False,
     description="Try out all the components!",
     article="Learn more about [Gradio](http://gradio.app)",
```
aliabid94, line 156: `cache_examples` is set to both True above and False here, causing an error

gradio/serializing.py
```diff
@@ -0,0 +1,142 @@
...
+
+class SimpleSerializable(Serializable):
+    def serialize(
+        self, x: Any, load_dir: str = "", encryption_key: bytes | None = None
+    ) -> Any:
+        """
+        Convert data from human-readable format to serialized format. For SimpleSerializable components, this is a no-op.
+        Parameters:
+            x: Input to interface
+            called_directly: if True, the component is part of an Interface/Blocks was called as a function, otherwise, it is being used via the GUI
```
aliabid94, line 38: called_directly not a parameter

gradio/serializing.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+class JSONSerializable(Serializable):
+    def serialize(
+        self, x: str, load_dir: str = "", encryption_key: bytes | None = None
+    ) -> str:
+        """
+        Convert from a a human-friendly version (string path to json file) to a
+        serialized representation (json string)
+        """
```
aliabid94, line 124: comment not necessary, explained in docstring

aliabid94 (approved):
Great stuff, makes code a lot cleaner.

gradio/components.py
```diff
@@ -2211,32 +1882,13 @@ def get_interpretation_scores(
...
-
     def generate_sample(self):
         return deepcopy(media_data.BASE64_AUDIO)
 
-    def postprocess(self, y: Tuple[int, np.array] | str) -> str:
+    def postprocess(self, y: Tuple[int, np.array] | str | None) -> str | None:
         """
         Parameters:
-            y: audio data in requested format
+            y: audio data in either of the following formats: a tuple of (sample_rate, data), or a string of the path to an audio file, or None.
```
freddyaboulton, line 1891: I can't run the `sst_or_tts` demo locally (suspect it will fail on spaces too). 

I am getting this error `OSError: [Errno 63] File name too long: 'data:audio/flac;base64,....`

I think the problem is the output is not being converted to a file. I think the `isinstance(y, tuple)` check is not actually what we want (don't yet understand everything in this PR fully to propose the right fix here)

gradio/components.py
```diff
@@ -3136,20 +2707,14 @@ def postprocess(self, y: str | None) -> str | None:
...
-
 
 ############################
 # Only Output Components
 ############################
 
 
 @document("change", "style")
-class Label(Changeable, IOComponent):
+class Label(Changeable, IOComponent, JSONSerializable):
```
freddyaboulton, line 2717: I think the front-end is not able to properly display the label json file returned by the back-end now.

I had to make the some modifications to `image_classifier_interface_load` (it was broken on main) but in tab 1, the filename is displayed instead of the contents and in tab 2 I get an error.

### Tab 1 Error
image

### Tab 2 Error
```
n run_sync
    return await get_asynclib().run_sync_in_worker_thread(
  File "/Users/freddy/miniconda3/envs/gradio/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 937, in run_sync_in_worker_thread
    return await future
  File "/Users/freddy/miniconda3/envs/gradio/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 867, in run
  File "/Users/freddy/sources/gradio/gradio/external.py", line 259, in query_huggingface_api
    data = pipeline["preprocess"](*params)
  File "/Users/freddy/sources/gradio/gradio/external.py", line 131, in <lambda>
    i.split(",")[1]
```

Modifications here:
```python
import gradio as gr
import pathlib

current_dir = pathlib.Path(__file__).parent

images = [str(current_dir / "cheetah1.jpeg"), str(current_dir / "cheetah1.jpg"), str(current_dir / "lion.jpg")]


img_classifier = gr.Interface.load(
    "models/google/vit-base-patch16-224", examples=images, cache_examples=False
)


def func(img, text):
    return img_classifier(img), text


using_img_classifier_as_function = gr.Interface(
    func,
    [gr.Image(type="filepath"), "text"],
    ["label", "text"],
    examples=[
        [str(current_dir / "cheetah1.jpeg"), None],
        [str(current_dir / "cheetah1.jpg"), "cheetah"],
        [str(current_dir / "lion.jpg"), "lion"],
    ],
    cache_examples=False,
)
demo = gr.TabbedInterface([using_img_classifier_as_function, img_classifier])

if __name__ == "__main__":
    demo.launch()
```

freddyaboulton: Actually, I don't think the front-end I think the problem since this doesn't happen in other demos I've tried with `Label` output (`sentiment_analysis`, `gender_sentence_default_interpretation`)

gradio/examples.py
abidlabs: Ah yes, good catch. That should `self.inputs_with_examples` I believe. Will fix

demo/kitchen_sink/run.py
abidlabs: Whoops thanks

gradio/serializing.py
abidlabs: Ah yes it's now removed. Good catch

gradio/components.py
abidlabs: Weird will take a look, thanks!

abidlabs:
Thanks guys for the review! Will go through these and fix these tonight, please let me know if you find any other failing demos

abidlabs:
Hmm okay it looks like this breaks `Interface.load` for models loaded from the Hub. Will need to dive into this more deeply tomorrow

demo/fake_gan/run.py
```diff
@@ -40,12 +39,12 @@ def fake_gan(count, *args):
     title="FD-GAN",
     description="This is a fake demo of a GAN. In reality, the images are randomly chosen from Unsplash.",
     examples=[
-        [2, cheetah, 12, None, None, None],
-        [1, cheetah, 2, None, None, None],
-        [4, cheetah, 42, None, None, None],
-        [5, cheetah, 23, None, None, None],
-        [4, cheetah, 11, None, None, None],
-        [3, cheetah, 1, None, None, None],
+        [2, cheetah, None, 12, None, None],
```
abidlabs, line 42: To be able to test non-consecutive optional examples

abidlabs:
Hi @freddyaboulton and @aliabid94, thanks for the initial review of the PR. I fixed the `Interface.load()` issue with the inference API and individually tested with all of the pipelines that we support. That issue should be fixed now. 

If you could take one more pass before I merge this in tomorrow, would really appreciate it!

demo/fake_gan/run.py
freddyaboulton, line 42: Maybe this broke fake_gan?

fake_gan_error

gradio/components.py
```diff
@@ -1533,6 +1294,9 @@ def format_image(
                 + ". Please choose from: 'numpy', 'pil', 'filepath'."
             )
 
+    def generate_sample(self):
```
freddyaboulton, line 1297: Why do we need this method? As far as I can tell, it's only called outside components.py in test code?

abidlabs: Ah yeah good point. This used to be called from the view API page in Gradio 2.x when we would generate a mock request for the user in the "Try It" section. It looks like we might bring it back (#1325) so I'd say let's leave this method here

demo/fake_gan/run.py
abidlabs: Let me take a look

gradio/serializing.py
```diff
@@ -0,0 +1,170 @@
...
+
+import os
+from abc import ABC, abstractclassmethod
+from typing import Any, Dict
+
+from gradio import processing_utils
+
+
+class Serializable(ABC):
+    @abstractclassmethod
```
freddyaboulton, line 11: abstractclassmethod is deprecated since 3.3. Ok to keep as is from a functional stand-point but maybe not a best practice.

image

demo/fake_gan/run.py
abidlabs: Good catch! Pushed a fix

gradio/serializing.py
abidlabs: That should actually be @abstractmethod anyways, will fix

gradio/components.py
```diff
@@ -1977,7 +1679,7 @@ def style(
 
 
 @document("change", "clear", "play", "pause", "stop", "stream", "style")
-class Audio(Changeable, Clearable, Playable, Streamable, IOComponent):
+class Audio(Changeable, Clearable, Playable, Streamable, IOComponent, FileSerializable):
```
freddyaboulton, line 1682: sst_or_tts is broken when you cache examples on the `tts_demo` interface

### Changes to demo
```python
import gradio as gr

title = "GPT-J-6B"

tts_examples = [
    "I love learning machine learning",
    "How do you do?",
]

tts_demo = gr.Interface.load(
    "huggingface/facebook/fastspeech2-en-ljspeech",
    title=None,
    examples=tts_examples,
    description="Give me something to say!",
    cache_examples=True
)

stt_demo = gr.Interface.load(
    "huggingface/facebook/wav2vec2-base-960h",
    title=None,
    inputs="mic",
    description="Let me try to guess what you're saying!",
)

demo = gr.TabbedInterface([tts_demo, stt_demo], ["Text-to-speech", "Speech-to-text"])

if __name__ == "__main__":
    demo.launch()
```

### exception
```
❯ python demo/sst_or_tts/run.py
Fetching model from: https://huggingface.co/facebook/fastspeech2-en-ljspeech
Caching examples at: '/Users/freddy/sources/gradio/gradio_cached_examples/51/log.csv'
  File "demo/sst_or_tts/run.py", line 10, in <module>
    tts_demo = gr.Interface.load(
  File "/Users/freddy/sources/gradio/gradio/interface.py", line 109, in load
  File "/Users/freddy/sources/gradio/gradio/blocks.py", line 759, in load
  File "/Users/freddy/sources/gradio/gradio/external.py", line 41, in load_blocks_from_repo
  File "/Users/freddy/sources/gradio/gradio/external.py", line 306, in get_models_interface
    interface = gradio.Interface(**kwargs)
  File "/Users/freddy/sources/gradio/gradio/interface.py", line 595, in __init__
    self.examples_handler = Examples(
  File "/Users/freddy/sources/gradio/gradio/examples.py", line 51, in create_examples
    utils.synchronize_async(examples_obj.create)
  File "/Users/freddy/sources/gradio/gradio/utils.py", line 362, in synchronize_async
    return fsspec.asyn.sync(fsspec.asyn.get_loop(), func, *args, **kwargs)
  File "/Users/freddy/miniconda3/envs/gradio/lib/python3.8/site-packages/fsspec/asyn.py", line 96, in sync
    raise return_result
  File "/Users/freddy/miniconda3/envs/gradio/lib/python3.8/site-packages/fsspec/asyn.py", line 53, in _runner
    result[0] = await coro
  File "/Users/freddy/sources/gradio/gradio/examples.py", line 194, in create
    await self.cache_interface_examples()
  File "/Users/freddy/sources/gradio/gradio/examples.py", line 231, in cache_interface_examples
    raise e
  File "/Users/freddy/sources/gradio/gradio/examples.py", line 228, in cache_interface_examples
    cache_logger.flag(prediction)
  File "/Users/freddy/sources/gradio/gradio/flagging.py", line 163, in flag
    component.deserialize(
  File "/Users/freddy/sources/gradio/gradio/serializing.py", line 129, in deserialize
    if x.get("is_file", False):
AttributeError: 'str' object has no attribute 'get'
```

freddyaboulton: I think the problem is that the external.py module is still returning base64 representation of the audio?

abidlabs: Ah okay I see, I think I missed this workflow: caching examples in Interfaces loaded from external.py

ui/packages/model3D/src/Model3DUpload.svelte
```diff
@@ -80,8 +80,8 @@
 		if (!value) return;
 
 		let url: string;
-		if (value.is_example) {
```
freddyaboulton, line 83: Might still need `is_example`? The model3D demo does properly display the input model3d file anymore
model_3d_upload_weird

gradio/components.py
```diff
@@ -10,13 +10,12 @@
 import numbers
```
freddyaboulton, line 10: nit-pick: Let's delete `save_flagged`, `restore_flagged` from the Carousel as those methods are deprecated?

gradio/serializing.py
freddyaboulton: I think we should make them static since they don't need a reference to the object inheriting these methods? In any event, thanks for making the fix!

freddyaboulton:
@abidlabs This is a heroic PR! So far I've just been finding little things that are broken from the refactor but I really like the approach  you've taken so I think it's just a matter of making sure we don't accidentally break anything with the PR!

About to take off will continue reviewing later this afternoon!

abidlabs:
Thank you for reviewing so thoroughly @freddyaboulton!

gradio/interface.py
```diff
@@ -650,75 +635,26 @@ def __repr__(self):
             repr += "\n|-{}".format(str(component))
         return repr
 
-    async def submit_func(self, *args):
-
-        prediction = await self.run_prediction(args)
-        # TODO(faruk): We don't have tuple or array clearence in Blocks, can we remove this and have one standart?
-        return prediction[0] if len(self.output_components) == 1 else prediction
-
     async def run_prediction(
```
FarukOzderim, line 638: This looks great! It looks like we don't need this function anymore, thx to the cleanups. Let's use this [function ](https://github.com/gradio-app/gradio/blob/main/gradio/blocks.py#L549)inside Blocks?

abidlabs: Good catch! Just to confirm, you mean use `Blocks.call_function` instead, right?

gradio/components.py
abidlabs: Fixed as long as `cache=False`. Still looking into `cache=True`

ui/packages/model3D/src/Model3DUpload.svelte
abidlabs: Now fixed, thanks for catching!

gradio/components.py
abidlabs: Should be fixed now!

abidlabs:
Great find @freddyaboulton and thanks for the tireless reviews! Will merge it in and reopen #1689 to fix afterwards

Issue #1833, opened 2022-07-19
Gradio dataframe not taking 'date' datatype


mbjamil:
### Describe the bug

I am trying to output a dataframe which has datetime values, gradio is automatically converting it into 'seconds'. When I check gradio documentation on dataframes, it says to pass the 'date' as a 'datatype'.

But when i do it, i get the following error:

TypeError: __init__() got an unexpected keyword argument 'datatype'

documentation link: https://gradio.app/docs/#dataframe

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

I am working on a company project, i cant share all the code. But i am doing something like this:

with demo:
    gr.Markdown(
        """
    # Welcome 
    """
    )
    list=[]
    for i in df['id'].unique():
        list.append(i)
    input = gr.Dropdown(list)
    btn = gr.Button("run")
    output = [gr.outputs.Dataframe(label='title1',datatype='date'),
                             gr.outputs.Dataframe(label='title2'),
                             gr.outputs.Dataframe(label='title3'),
                             gr.outputs.Dataframe(label='title4'),
                             gr.outputs.Dataframe(label='title5'),
                             gr.outputs.Dataframe(label='title6'),
                             gr.outputs.Dataframe(label='title7'),
                             gr.outputs.Dataframe(label='title8'),
                             gr.outputs.Dataframe(label='title9'),
                             gr.outputs.Dataframe(label='title10'),
                             gr.outputs.Dataframe(label='title11'),
                             gr.outputs.Dataframe(label='title12'),
                             gr.outputs.Dataframe(label='title13')]
    btn.click(fn=function, inputs=input, outputs=output)

demo.launch()

### Logs

```shell
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-91-c12e18d64ded> in <module>
     68     input = gr.Dropdown(list)
     69     btn = gr.Button("Submit")
---> 70     output = [gr.outputs.Dataframe(),]

TypeError: __init__() got an unexpected keyword argument 'datatype'
```

### System Info

```shell
Gradio version: 3.0.20
Intel(R) Core(TM) i7-10510U CPU @ 1.80GHz   2.30 GHz
```

### Severity

annoying

freddyaboulton:
@mbjamil glad we solved the original issue! I think the formatting of `How the actual time looks in the dataframe:` may have gotten messed up in the email. Do you mind sharing again?

mbjamil:
[...]

PR #1979, opened 2022-08-08
Prevent Dataframe post-processing from changing dtypes

freddyaboulton:
# Description
Calling `to_list()` would change the dtypes of the data in the dataframe. Using `to_dict` preserves the dtypes. 

In my opinion, the `datatype` parameter is confusing. It should not accept the value `date` because it's not clear what dateformat we would use in the conversion. We should be letting users control the format of their data from their predict function by applying as little post-processing as possible.

To test, run `demo/dataframe_datatype`

Fixes: #1833

# Checklist:

### DIFF
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -18,7 +18,7 @@ import uuid
 import warnings
 from copy import deepcopy
 from types import ModuleType
-from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional, Tuple
+from typing import TYPE_CHECKING, Any, Callable, Dict, Iterable, List, Optional, Tuple
 
 if TYPE_CHECKING:
     from typing import TypedDict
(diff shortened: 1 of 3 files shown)


gradio/components.py
```diff
@@ -2535,7 +2535,7 @@ def __init__(
             headers: List of str header names. If None, no headers are shown.
             row_count: Limit number of rows for input and decide whether user can create new rows. The first element of the tuple is an `int`, the row count; the second should be 'fixed' or 'dynamic', the new row behaviour. If an `int` is passed the rows default to 'dynamic'
             col_count: Limit number of columns for input and decide whether user can create new columns. The first element of the tuple is an `int`, the number of columns; the second should be 'fixed' or 'dynamic', the new column behaviour. If an `int` is passed the columns default to 'dynamic'
-            datatype: Datatype of values in sheet. Can be provided per column as a list of strings, or for the entire sheet as a single string. Valid datatypes are "str", "number", "bool", and "date".
+            datatype: Datatype of values in sheet. Can be provided per column as a list of strings, or for the entire sheet as a single string. Valid datatypes are "str", "number", "bool", "date", and "markdown".
```
freddyaboulton, line 2538: I feel like we should only accept "markdown" as a value since that's the only team we really change the user data?

abidlabs: I think it's useful to have different data types for the Dataframe component when it is used as an input component. For example, we could preprocess the data in different ways depending on what the column type is (I actually thought we did that, but I don't see it in the code), or we might do same basic validation in the future

abidlabs:
Hi @freddyaboulton the changes look good to me. Strangely enough, I had to run the example you provided in the issue to reproduce this behavior:

```py
import gradio as gr
import pandas as pd


def make_dataframe(n_periods):
    return pd.DataFrame({"date_1": pd.date_range("2021-01-01", periods=n_periods),
                         "date_2": pd.date_range("2022-02-15", periods=n_periods)})


gr.Interface(make_dataframe, 
             gr.Number(precision=0), 
             gr.Dataframe(datatype="date")).launch()

```

as the demo that you created as part of this PR worked fine even before you made the changes!

abidlabs (approved):

freddyaboulton:
@abidlabs I've been using pandas for a while but it still surprises me 😂 

Looks like the problem only surfaces if the only columns are date columns. 

image

Added a unit test for this - will merge once ci passes.

Thank you for the review and great catch!

PR #1981, opened 2022-08-09
Queue fixes

aliabid94:
This fixes the frontend and modifies the backend for the Queue:
Backend:
- Calculates rank and ETA on first estimation
- Cleans code to use None for null values, not -1
- Updated rank_eta calculation to factor in whether there is a thread available
Frontend:
- Changed logic so that an updated value for status does not affect ETA calculation


gradio/event_queue.py
```diff
@@ -116,6 +115,7 @@ async def start_processing(cls) -> None:
     @classmethod
     def push(cls, event: Event):
         cls.EVENT_QUEUE.append(event)
+        return len(cls.EVENT_QUEUE) - 1
```
FarukOzderim, line 119: let's add a return type hint

gradio/event_queue.py
```diff
@@ -195,9 +195,13 @@ async def send_estimation(
...
-        estimation.rank_eta = round(
-            (estimation.rank + 1) * cls.AVG_CONCURRENT_PROCESS_TIME + cls.AVG_PROCESS_TIME
-        )
+
+        if cls.AVG_CONCURRENT_PROCESS_TIME is not None:
+            estimation.rank_eta = (
+                estimation.rank * cls.AVG_CONCURRENT_PROCESS_TIME + cls.AVG_PROCESS_TIME
+            )
+            if None not in cls.ACTIVE_JOBS:
+                estimation.rank_eta += cls.AVG_CONCURRENT_PROCESS_TIME
```
FarukOzderim, line 204: Good perspective, a comment would be nice though, had hard time to understand what it does.
```suggestion
                # Add estimated amount of time for a thread to get empty 
                estimation.rank_eta += cls.AVG_CONCURRENT_PROCESS_TIME 
```

gradio/event_queue.py
```diff
@@ -246,7 +250,7 @@ async def process_event(cls, event: Event) -> None:
             json=event.data,
         )
         end_time = time.time()
-        cls.update_estimation(duration=round(end_time - begin_time, 3))
+        cls.update_estimation(end_time - begin_time)
```
FarukOzderim, line 253: why did you drop the round?
I did add that to not send floats with 10 digits after the "." to the frontend. 

- decreases package size
- makes it more readable with few digits

gradio/routes.py
```diff
@@ -279,15 +279,15 @@ async def predict(
...
             e_hash = await event.get_message()
             if e_hash is None:
                 return
             print(f"Event Body Received: {e_hash}")
             event.hash = e_hash["hash"]
-            Queue.push(event)
-            print("An event Joined the queue")
+            rank = Queue.push(event)
+            estimation = Queue.get_estimation()
+            await Queue.send_estimation(event, estimation, rank)
```
FarukOzderim, line 290: Good idea!

FarukOzderim (commented):
A few minor points. Good work!

FarukOzderim:
Should dawoodkhan82 take a quick look at changeis in the frontend @aliabid94?

FarukOzderim:
Can you bump version to 3.1.4b3 as well @aliabid94?

test/test_event_queue.py
```diff
@@ -19,7 +19,9 @@ async def wait(data):
             button.click(wait, [text], [text])
         app, local_url, _ = demo.launch(prevent_thread_lock=True, enable_queue=True)
         client = TestClient(app)
-        with client.websocket_connect("/queue/join") as websocket:
+        with client.websocket_connect("/queue/join") as _:  # websocket
+            """#Unable to make this part work, seems like there is an issue with thread acquire and exiting the scope
+            websocket.send_json({"hash": "0001"})
```
FarukOzderim, line 24: Commented the tests since they don't work well. Only tests connecting ws to the endpoint from now on.

We need to test the queue in frontend, will open an issue.

FarukOzderim:
Btw tests in frontend also fails @aliabid94 😸

gradio/event_queue.py
aliabid94: rounding should be done at the last step in the front end. Rounding at every step in a calculation introduces compounding errors and is unnecessary calculation. for example:
rounding done at the last step: 0.44 + 0.44 = 0.88. round(0.88, 1) => 0.9
rounding done at every step: round(0.44, 1) + round(0.44, 1) => 0.8

gradio/event_queue.py
aliabid94: Makes sense

gradio/event_queue.py
```diff
@@ -13,11 +13,11 @@
 
 class Estimation(BaseModel):
     msg: Optional[str] = "estimation"
-    rank: Optional[int] = -1
+    rank: Optional[int] = None
```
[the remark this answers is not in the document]
aliabid94: Yep the frontend handles it fine now.

PR #1992, opened 2022-08-10
Plot Component Guide

dawoodkhan82:
# Description
Guide for creating plot component. Using matplotlib, plotly, bokeh. Need to add more cool plot demos.

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

Closes: # (issue)
#1993

# Checklist:

### DIFF
--- a/guides/5)other_tutorials/how_to_use_plot_component.md
+++ b/guides/5)other_tutorials/how_to_use_plot_component.md
@@ -1,0 +1,103 @@
+# How to Use the Plot Component
+
+Related spaces: https://huggingface.co/spaces/dawood/Plot
+Tags: PLOT, IMAGE
+
+## Introduction
+
+Matplotlib, Plotly, and Bokeh are the most popular plotting libraries used in machine learning. Using `gradio`, you can easily build a demo using any of these libraries and share it with anyone. The Gradio plot component accepts any of the following: a `matplotlib.figure.Figure`, a `plotly.graph_objects._figure.Figure`, or a `dict` corresponding to a bokeh plot (json_item format).
+
+This guide will show you how to build a demo for your plot model in a few lines of code; like the one below.
+
+<gradio-app space="dawood/Plot"> </gradio-app>
+
+### Prerequisites
+
+Make sure you have the `gradio` Python package already [installed](/getting_started).
+
+
+## Taking a Look at the Code
+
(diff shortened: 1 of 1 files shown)


guides/5)other_tutorials/how_to_use_plot_component.md
```diff
@@ -0,0 +1,103 @@
...
+* `fn`: the prediction function that is used when the user clicks submit. In our case this is the `outbreak` function.
+* `inputs`: a list of inputs including a dropdown menu for selecting which plotting library should be used; as well as other inputs used for plotting the output.
+* `outputs`: a plot output component.
+* `examples`: list of valid inputs from the 5 input components.
+* `cache_examples`: saves the predicted output for the examples, to save time on inference.
+
+
+## Exploring mode complex Plot Demos:
+
+TODO
```
dawoodkhan82, line 96: @abidlabs @AK391 Have you seen any cool plot demos on spaces I can link here?

abidlabs:
@dawoodkhan82 I think this Guide is a really good opportunity to walk users through how to create plots with `matplotlib` and `plotly` (I would leave out `bokeh` to avoid confusion, since we don't currently support it. We can add a note saying it will be supported in the future).

As it's currently structured, however, this Guide, only provides a single (relatively complex) example of using both together. I would avoid that and instead break this Guide into 2 sections, for `matplotlib` plots and `plotly` plots separately (since a user is probably looking for how to use one of those frameworks).

In each section, I would provide:
* Some background of the library (`matplotlib` vs. `plotly`) -- what is the advantage / disadvantage of each library
* What object should be returned for that library (e.g. a `matplotlib.pyplot.Figure` object)
* A **simple** example showing that library being used with Gradio

At the end, you could have the `plot_outbreak` demo combining both of the plotting frameworks together.

dawoodkhan82:
> 

makes sense. Will update

freddyaboulton:
Thanks for starting the guide @dawoodkhan82 ! I know plotting isn't strictly related to data science but what are your thoughts on creating a Data Science "track" to the guides and adding this guide to that? Would help onboard potential data scientists.

dawoodkhan82:
@freddyaboulton Yup makes sense, I assume we'll need more guides when we do the additional integrations; like skops.

Issue #1999, opened 2022-08-11
Some `async`-related tests are failing


abidlabs:
### Describe the bug

Some async-related tests are failing on `main`. Specifically:
* `test/test_mix.py:33`
* `test/test_external.py:237`
* `test/test_examples.py:65`
* `test/test_examples.py:78`

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

https://app.circleci.com/pipelines/github/gradio-app/gradio/5035/workflows/e80ec70c-8fd0-46fc-bea6-47b7e0bd97fc/jobs/6793

### Logs

```shell
See Circle CI
```

### System Info

```shell
Circle CI
```

### Severity

serious, but I can work around it

PR #2000, opened 2022-08-11
Fix async tests

abidlabs:
Closes: #1999 

Note: commented out one of the tests because we think it will get fixed by #1967

### DIFF
--- a/gradio/examples.py
+++ b/gradio/examples.py
@@ -7,6 +7,7 @@ import csv
 import inspect
 import os
 import shutil
+import warnings
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable, List, Optional, Tuple
 
@@ -44,6 +45,7 @@ def create_examples(
         fn=fn,
         cache_examples=cache_examples,
         examples_per_page=examples_per_page,
+        _initiated_directly=False,
     )
     utils.synchronize_async(examples_obj.create)
     return examples_obj
@@ -69,6 +71,7 @@ class Examples:
         fn: Optional[Callable] = None,
         cache_examples: bool = False,
         examples_per_page: int = 10,
+        _initiated_directly=True,
     ):
         """
         Parameters:
@@ -79,6 +82,11 @@ class Examples:
             cache_examples: if True, caches examples for fast runtime. If True, then `fn` and `outputs` need to be provided
             examples_per_page: how many examples to show per page (this parameter currently has no effect)
         """
+        if _initiated_directly:
+            raise warnings.warn(
+                "Please use gr.Examples(...) instead of gr.examples.Examples(...) to create the Examples.",
+            )
+
         if cache_examples and (fn is None or outputs is None):
             raise ValueError("If caching examples, `fn` and `outputs` must be provided")
 
@@ -122,7 +130,8 @@ class Examples:
 
         else:
             raise ValueError(
-                "The parameter `examples` must either be a directory or a nested "
+                "The parameter `examples` must either be a string directory or a list"
+                "(if there is only 1 input component) or (more generally), a nested "
                 "list, where each sublist represents a set of inputs."
             )
 
--- a/gradio/examples.py
+++ b/gradio/examples.py
@@ -28,6 +28,27 @@ LOG_FILE = "log.csv"
 set_documentation_group("component-helpers")
 
 
+def create_example(
+    examples: List[Any] | List[List[Any]] | str,
+    inputs: Component | List[Component],
+    outputs: Optional[Component | List[Component]] = None,
+    fn: Optional[Callable] = None,
+    cache_examples: bool = False,
+    examples_per_page: int = 10,
+):
+    """Top-level coroutine function that creates Examples. Provided for backwards compatibility, e.g. so that gr.Examples can be used ."""
+    examples = Examples(
+        examples=examples,
+        inputs=inputs,
+        outputs=outputs,
+        fn=fn,
+        cache_examples=cache_examples,
+        examples_per_page=examples_per_page,
+    )
+    # TODO: run Examples.create synchronously
+    # TODO: export this function as gr.Examples for backwards compatibility
+
+
 @document()
 class Examples:
     """
@@ -150,7 +171,7 @@ class Examples:
         self.cached_file = os.path.join(self.cached_folder, "log.csv")
         self.cache_examples = cache_examples
 
-    async def setup(self) -> Examples:
+    async def create(self) -> Examples:
         """Caches the examples if self.cache_examples is True and creates the Dataset
         component to hold the examples"""
         if self.cache_examples:
--- a/gradio/examples.py
+++ b/gradio/examples.py
@@ -240,9 +240,7 @@ class Examples:
         if inspect.iscoroutinefunction(self.fn):
             predictions = await self.fn(*processed_input)
         else:
-            predictions = await anyio.to_thread.run_sync(
-                self.fn, *processed_input, limiter=self.limiter
-            )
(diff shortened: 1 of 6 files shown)


gradio/examples.py
```diff
@@ -198,7 +228,12 @@ def process_example(self, example_id: int) -> Tuple[List[Any], List[float]]:
             input_component.preprocess(raw_input[i])
             for i, input_component in enumerate(self.inputs)
         ]
-        predictions = self.fn(*processed_input)
+        if inspect.iscoroutinefunction(self.fn):
+            predictions = await self.fn(*processed_input)
+        else:
+            predictions = await anyio.to_thread.run_sync(
+                self.fn, *processed_input, limiter=self.limiter
+            )
```
FarukOzderim, line 236: It looks like Example class is not linked to Blocks class, but could we make use of the function in Blocks?

`async def call_function(self, fn_index, processed_input):
`

gradio/examples.py
```diff
@@ -25,6 +28,27 @@
...
+    """Top-level synchronous function that creates Examples. Provided for backwards compatibility, i.e. so that gr.Examples(...) can be used to create the Examples component."""
+    examples_obj = Examples(
+        examples=examples,
+        inputs=inputs,
+        outputs=outputs,
+        fn=fn,
+        cache_examples=cache_examples,
+        examples_per_page=examples_per_page,
+    )
+    utils.synchronize_async(examples_obj.create)
```
FarukOzderim, line 48: If this does not need to be sync, we could also run it in the background with 

            utils.run_coro_in_background(fnc, parameters)

FarukOzderim: Actually, let's use `run_coro_in_background`

abidlabs: What if the examples take a while to cache. I think this would lead to the Interface launching without the examples caching first?

gradio/examples.py
abidlabs: I agree it's similar code, but it's not possible to use `call_function()` as is. We would have to rewrite it to accept a function as an argument, which I think is overkill for this situation

test/test_mix.py
```diff
@@ -23,17 +23,17 @@ async def test_in_interface(self):
...
-        io2 = gr.Interface.load("spaces/abidlabs/image-classifier")
-        series = mix.Series(io1, io2)
-        try:
-            output = series("gradio/test_data/lion.jpg")
-            assert output["lion"] > 0.5
-        except TooManyRequestsError:
-            pass
+    # @pytest.mark.asyncio
+    # @pytest.mark.flaky
+    # async def test_with_external(self):
```
freddyaboulton, line 28: How come this test was failing? And how come #1967 will fix it if it has nothing to do with sync vs async?

test/test_examples.py
```diff
@@ -59,20 +61,22 @@ def test_some_headers(self):
 
 
 class TestProcessExamples:
```
freddyaboulton, line 63: Think it would be useful to test Examples when the inference function is a coroutine?

gradio/examples.py
```diff
@@ -55,6 +82,11 @@ def __init__(
             cache_examples: if True, caches examples for fast runtime. If True, then `fn` and `outputs` need to be provided
             examples_per_page: how many examples to show per page (this parameter currently has no effect)
         """
+        if _initiated_directly:
+            raise warnings.warn(
```
freddyaboulton, line 86: I think we should raise an exception. The demo would be broken if they used `gr.examples.Examples` right?

gradio/examples.py
freddyaboulton: Yea agreed we should run this synchronously - like we do now

test/test_mix.py
abidlabs: So this test is failing because the serialization/deserialization logic is not correct anymore in `Series`. 

As part of #1967, I'm actually rewriting the serialization logic, so this will either be fixed automatically or I'll take another look at it and fix it. 

We felt it would be unnecessary effort to fix it only to change the way we do serialization.

gradio/examples.py
```diff
@@ -198,7 +237,12 @@ def process_example(self, example_id: int) -> Tuple[List[Any], List[float]]:
             input_component.preprocess(raw_input[i])
             for i, input_component in enumerate(self.inputs)
         ]
-        predictions = self.fn(*processed_input)
+        if inspect.iscoroutinefunction(self.fn):
+            predictions = await self.fn(*processed_input)
+        else:
+            predictions = await anyio.to_thread.run_sync(
+                self.fn, *processed_input, limiter=self.limiter
```
freddyaboulton, line 244: self.limiter isn't defined anywhere which will lead to exception when running a blocks api app.

```
  File "/Users/freddy/sources/gradio/gradio/examples.py", line 220, in cache_interface_examples
    prediction = await self.process_example(example_id)
  File "/Users/freddy/sources/gradio/gradio/examples.py", line 244, in process_example
    self.fn, *processed_input, limiter=self.limiter
AttributeError: 'Examples' object has no attribute 'limiter'
```

This is what I get running `blocks_inputs`

freddyaboulton (commented):
@abidlabs Thanks for the fix.

I think once we fix the issue with `self.limiter` this will be good to merge!

test/test_mix.py
freddyaboulton: Fair point! We should try to always keep the ci on main passing though. What if someone decides to cut a release tomorrow (to test the new queue for example) and they're not aware that some of the code is broken on main?

gradio/examples.py
FarukOzderim: Oh shoot, I was actually going to say 

Actually, let's use synchronize_async :D

gradio/examples.py
FarukOzderim: This is because limiter is defined on Blocks. Btw since this is just examples, we can remove limiter support from it, not need IMO.

```suggestion
                self.fn, *processed_input
```

freddyaboulton: Agreed!

gradio/examples.py
abidlabs: Not necessarily. They can use `ex = gr.examples.Example` as long as they remember to run `await ex.create()` afterwards. Basically just what the top-level function does.

test/test_examples.py
abidlabs: Yup sounds good, I'll add a test

gradio/examples.py
abidlabs: Thanks, will remove

test/test_mix.py
abidlabs: Ah I agree! But if you guys think it's all right, we can make an exception in this case, because I think it's quite unlikely a user will run into this edge case (it only happens when you use `Series` with `external`) in the time it takes for us to fix it.

abidlabs:
Thanks for the review @freddyaboulton and @FarukOzderim! I've made the changes. I'll let the tests run and will merge shortly afterwards

gradio/examples.py
freddyaboulton: Actually, I just realized `raise` will actually try to raise an exception. I think you mean `warnings.warn(...)` only.

Yea that's a good point! Let's just warn instead of raising an exception.

test/test_mix.py
FarukOzderim: Yeah, being flexible is alright. We don't have a upcoming stable release soon as well.

gradio/examples.py
abidlabs: Ah good catch, will fix (sorry missed this earlier)

PR #2003, opened 2022-08-11
Convert video to playable format if ffmpeg installed

### LINKED ISSUE #1970: Add information about available video codecs

The docstring of `gr.Video` only mentions about file extensions, but a video file must be encoded with specific codecs to be playable in browsers as written in [this issue comment](https://github.com/gradio-app/gradio/issues/1508#issuecomment-1154545730).
I've seen couple of times that people (including myself) fell into this pitfall and got confused why the video didn't play as expected. See [this comment](https://huggingface.co/spaces/Yuliang/ICON/discussions/6#62e8fd3e4210d3fe69eb3637) for example.

It'd be helpful if there was description about available video codecs in the docstring of `gr.Video`.

### LINKED ISSUE #1972: Issues when loading/outputting video
### Describe the bug

I'm having weird situations in which calling an inference function directly with a given video works well (I can download and play it), but just wrapping it with a `gr.Interface` outputs an invalid video (even if I download the video, I cannot play it, it gives an error in the demo). I saw @nateraw also had to do some hacks to make it work in a previous demo https://huggingface.co/spaces/nateraw/yolov6/blob/main/app.py#L33.

Adding this before returning makes it work

```python
out_file = tempfile.NamedTemporaryFile(suffix="out.mp4", delete=False)
subprocess.run(f"ffmpeg -y -loglevel quiet -stats -i {output_fname} -c:v libx264 {out_file.name}".split())
return out_file.name
```

### Reproduction

https://huggingface.co/spaces/nateraw/yolov6/blob/main/app.py#L58 without those two lines

### Screenshot

_No response_

### Logs

```shell
None
```

### System Info

```shell
Gradio 3.1.0 and above
```

### Severity

### LINKED ISSUE #1990: Error displaying video generated by open-cv
### Describe the bug

Hello,
I tried the following code to write frames into video:
```
    fourcc = cv2.VideoWriter_fourcc(*'mp4v')
    vid_out = cv2.VideoWriter(args.output, fourcc, fps, (w, h))
    ret = True
    while ret:
        ret, frame = cap.read()
        if ret:
                frame = inference_frame(frame)
                if len(frame) == 0:
                    continue
                detected += 1
                if args.output is not None:
                    vid_out.write(frame)
    cap.release()
    vid_out.release()

```
but this code gives an error result on website.
捕获
I tried different codecs, i.e. avc1, h264, avc1, VP90, theo, and only VP90 works but it is super slow.

And also, the uploaded video is mirrored for some reason.
捕获

Last, is there any way to change the 'content' in browser tags? Now it is displaying the same thing as the "title".
捕获2

Any solutions to those problems?
Thanks!

### Reproduction

The code to reproduce the problem is attached in description.

### Screenshot

_No response_

### Logs

```shell
No logs although set debug = True
```

### System Info

```shell
3.1.1
```

### Severity

freddyaboulton:
# Description
Fixes #1970 
Fixes #1990 
Fixes #1972 

In summary, I added `video_is_playable` util function to determine if a video can be played in the browser and `convert_video_to_playable_mp4` in the event that the video is not playable. The rest are just unit tests. 

I also added `white_noise_vid_not_playable` which is a reproducer of the original problem listed in one of the issues. 

# Checklist:


test/test_processing_utils.py
```diff
@@ -131,5 +133,31 @@ def test_subclass_conversion(self):
             assert y.dtype == x.dtype
 
 
+def test_video_has_playable_codecs(test_file_dir):
```
freddyaboulton, line 136: I added some files so that we can test positive and negative cases. I know increasing the repo size is a concern but I think testing the logic is worth adding the couple hundred kb

gradio/components.py
```diff
@@ -1919,7 +1921,7 @@ def restore_flagged(self, dir, data, encryption_key):
     def generate_sample(self):
         return deepcopy(media_data.BASE64_VIDEO)
 
-    def postprocess(self, y: str) -> str:
+    def postprocess(self, y: str) -> Optional[Dict[str, str]]:
```
freddyaboulton, line 1924: Old type hint was incorrect

gradio/components.py
```diff
@@ -1929,6 +1931,14 @@ def postprocess(self, y: str) -> str:
         if y is None:
             return None
         returned_format = y.split(".")[-1].lower()
```
freddyaboulton, line 1933: Unreleated to this pr but I feel like we should use a NamedTemporaryFile here like we do for Image?

abidlabs: #1967 takes care of this

gradio/processing_utils.py
```diff
@@ -518,3 +522,61 @@ def _scale(a, n, m, copy=True):
...
+
+
+def ffmpeg_installed() -> bool:
+    return shutil.which("ffmpeg") is not None
+
+
+def video_is_playable(video_filepath: str) -> bool:
+    """Determines if a video is playable in the browser.
+
+    A video is playable if it has a playable container and codec.
```
abidlabs, line 534: This is true all across major browsers?

gradio/processing_utils.py
```diff
@@ -518,3 +522,61 @@ def _scale(a, n, m, copy=True):
...
+        )
+        output = json.loads(output.stdout)
+        video_codec = output["streams"][0]["codec_name"]
+        return (container, video_codec) in [
+            (".mp4", "h264"),
+            (".ogg", "theora"),
+            (".webm", "vp9"),
+        ]
+    # If anything goes wrong, assume the video can be played to not convert downstream
+    except Exception:
```
abidlabs, line 563: Nit: can we narrow the scope of this Exception?

abidlabs, line 563: Nit: can we narrow the scope of this Exception?

gradio/processing_utils.py
```diff
@@ -518,3 +522,61 @@ def _scale(a, n, m, copy=True):
...
+    """Determines if a video is playable in the browser.
+
+    A video is playable if it has a playable container and codec.
+        .mp4 -> h264
+        .webm -> vp9
+        .ogg -> theora
+    """
+    try:
+        container = pathlib.Path(video_filepath).suffix.lower()
+        output = subprocess.run(
```
abidlabs, line 541: Instead of using `subprocess`, could we use `ffmpeg.probe` for cleaner code and better error handling: https://kkroening.github.io/ffmpeg-python/#ffmpeg.probe

gradio/processing_utils.py
```diff
@@ -518,3 +522,61 @@ def _scale(a, n, m, copy=True):
...
+        with tempfile.NamedTemporaryFile() as tmp_file:
+            shutil.copy2(video_path, tmp_file.name)
+            # ffmpeg will automatically use h264 codec (playable in browser) when converting to mp4
+            ff = FFmpeg(
+                inputs={str(tmp_file.name): None},
+                outputs={str(output_path): None},
+                global_options="-y -loglevel quiet",
+            )
+            ff.run()
+    except:
```
abidlabs, line 580: Same nit: can we narrow the scope to `ffmpeg.Error`

gradio/processing_utils.py
```diff
@@ -518,3 +522,61 @@ def _scale(a, n, m, copy=True):
...
+        ]
+    # If anything goes wrong, assume the video can be played to not convert downstream
+    except Exception:
+        return True
+
+
+def convert_video_to_playable_mp4(video_path: str) -> str:
+    """Convert the video to mp4. If something goes wrong return the original video."""
+    try:
+        output_path = pathlib.Path(video_path).with_suffix(".mp4")
```
abidlabs, line 570: I don't understand why we create this `output_path` variable and then copy the original file to it. 

1. Shouldn't we just pass as the input to `FFmpeg` the original `video_path`?
2. If there's an exception in `FFmpeg`, it looks like we would return the original video file but with the ".mp4" suffix. Wouldn't this cause issues if the video is in a different format?

gradio/processing_utils.py
freddyaboulton: Yea good call!

abidlabs:
I tried running the `white_noise_vid_not_playable` demo, and it isn't working for me. I see a completely black output:

image

Note that downloading the file works fine.

No Python logs except for the warning: `UserWarning: Video does not have browser-compatible container or codec. Converting to mp4`

On the JS side, I see: "Uncaught (in promise) DOMException: The element has no supported sources."

Tested on Chrome and Edge browsers on Windows.

gradio/processing_utils.py
freddyaboulton: 1. The converted video needs to have an extension that's playable in the browser (`mp4`, `ogg`, `webm`). If not, the browser won't play it (even if it has a playable video codec) so this function has to rename the extension (I arbitrarily picked mp4).
2. If there's an exception, I think we'll return the original video which is what's tested with this unit [test](https://github.com/gradio-app/gradio/pull/2003/files#diff-c7fbfc627571585283ab24f28191f1e640428c3d91159faf0c3e393ab7759f03R164)

gradio/processing_utils.py
freddyaboulton: I believe so! Will confirm

gradio/processing_utils.py
abidlabs: Oh sorry I misread the code, yeah this is completely fine!

freddyaboulton:
@abidlabs Thanks for reviewing/running the demo! Let me dig into why it didn't work for you

abidlabs:
Working great now @freddyaboulton! 

image

abidlabs (approved):

gradio/processing_utils.py
lazarusking: I got a ```video_codec= output["streams"][0]["codec_name"]
 ``` 
The narrowed scope of the exception broke my application, it worked when I reverted back to `except Exception:`

gradio/processing_utils.py
lazarusking: I got a ```video_codec= output["streams"][0]["codec_name"]
 ``` 
The narrowed scope of the exception broke my application, it worked when I reverted back to `except Exception:`

abidlabs: Ah yikes that's good to know. I think it would be still be better to have a scope that includes `IndexError` as well, but WDYT @freddyaboulton?

freddyaboulton: @lazarusking sorry to hear that! I originally used a wide exception because analyzing video streams with ffmpeg felt really unpredictable. 

Can you provide a reproducer so we can dig in further? I’m curious why ffmpeg can’t find any video streams in your file, yet the video can be played in the browser if comment this out.

lazarusking: [My ffmpeg demo](https://github.com/lazarusking/gradio-ffmpeg)
It basically takes a file and converts it to the chosen format, allowing you the option to play it as video, audio or just as a file. 
In the older gradio version, playing an audio as a video still works(when you toggle to video display option) but the new one doesn't find any video stream for such a case and just breaks.
It didn't account for "if anything goes wrong in this case".

freddyaboulton: Thanks - I can put up a PR to also catch IndexError and KeyError!

lazarusking: Thank you!

lukasfolle:
In my [hf space](https://huggingface.co/spaces/lfolle/LDM-SyntheticChestX-Ray), the video plays nicely on Chrome and Firefox, but not on Safari. I use the .ogg video container and the theora codec. 

Normally, I would expect a conversion to mp4, but this is not happening..
Any ideas @freddyaboulton @abidlabs?

freddyaboulton:
Hi @lukasfolle ! Thanks for asking. Very cool space and I am also unable to see the video on my mac when using safari. 

The conversion to mp4 may not be happening because .ogg is a playable video container but it has spotty support on safari: https://caniuse.com/?search=ogg 

Maybe you can "force" the mp4 conversion by giving your file an mp4 extension and using a different video codec? 

I think just changing the extension here to `.mp4` should be enough?

https://huggingface.co/spaces/lfolle/LDM-SyntheticChestX-Ray/blob/main/create_video.py#L19

lukasfolle:
Hi @freddyaboulton,

Thanks for your reply. I now switched to mp4 (h264) and it seems to only work on edge, firefox now shows a black window for me.
Any idea whats going wrong here?

Issue #1937, opened 2022-08-02
[Integration Tests] Update a demo's config.json as part of integration test


dawoodkhan82:
### Describe the bug

The config.json for each demo used by the integration test needs to be manually updated by running `write_config.py`. This should be done as part of each integration test programatically. 

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

N/A

### Logs

```shell
N/A
```

### System Info

```shell
N/A
```

### Severity

annoying

PR #2005, opened 2022-08-12
[Integration Tests] Update a demo's config.json as part of integration test

dawoodkhan82:
# Description
When `pnpm test:browser` or `pnpm test:browser:full` is run from the `ui` the configs for each demo in the integration tests will be re-created.

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

Closes: # (issue)
#1937

# Checklist:


demo/kitchen_sink/run.py
```diff
@@ -73,7 +73,7 @@ def fn(
             (".", 0),
         ]
         + [(f"test", x / 10) for x in range(-10, 10)],  # HighlightedText
-        json.loads(JSONOBJ),  # JSON
+        json.loads(JSONOBJ),  # JSONåå
```
abidlabs, line 76: Weird symbols showing up at the end for me?

abidlabs:
@dawoodkhan82 this approach looks good, but the functional tests can't be run because `gradio` is not installed on the machine. Perhaps you have to do `pnpm exec pip install -r requirements.txt` first?

abidlabs:
Actually probably better to add the steps to install `gradio` to this workflow directly: https://github.com/gradio-app/gradio/blob/21f9da9380b66420a598e79eb0b41feceba9077d/.github/workflows/ui.yml#L56

demo/kitchen_sink/run.py
dawoodkhan82: whoops good catch

.github/workflows/ui.yml
```diff
@@ -53,6 +53,7 @@ jobs:
           node-version: 16
           cache: pnpm
           cache-dependency-path: ui/pnpm-lock.yaml
+      - run: pnpm exec pip install -r ../requirements.txt
```
abidlabs, line 56: I think this should work, right?
```suggestion
      - run: pip install -r ../requirements.txt
```

abidlabs:
Awesome @dawoodkhan82! Just one suggestion for simplification but otherwise LGTM

abidlabs (approved):

PR #2012, opened 2022-08-15
Deploy All Demos to Spaces on PRs

freddyaboulton:
# Description
The goal of this PR is to increase test coverage and make it easier to review PRs by deploying all* of the demos to a HF space. By increasing the frequency with which we run our demos, my hope is that devs and reviewers will catch breaking changes prior to opening up a PR for review.

This PR accomplishes that in two parts:

### 1. Combining all demos into one demo
* Rather than publishing a separate space for each demo, I decided to deploy one space that contains all demos. The reason for this is that there are 70+ demos already and I think it will be easier to review if all the demos are in one page as opposed to 70 different pages. Also I think the number of spaces will explode really quickly with this approach.
* To accomplish this, I created `all_demos` demo. This script programmatically combines all demos into a tabbed demo where each tab is a different demo.
* Prior to running `all_demos`, the `copy_demos.py` script should be run to copy the demos to the `all_demos` directory. All the demos need to be in the `all_demos` directory for deployment to spaces to work. Right now I'm not actually running all_demos, just the ones that don't have a `requirements.txt` file (arbitrary decision).

### 2. Github Action that runs on PR commits
* This action will build a wheel of the current commit and upload it to s3 at `s3://gradio-builds/<pr-number>/*.whl`. The wheel has to be available over the internet so that spaces can install it. I think s3 is easier than pypi and it won't pollute the gradio pypi page.
* Then we copy all the demos to `all_demos` with `copy_demos` and then upload to spaces with the `upload_to_space` script

### Miscellaneous changes
* Some demos were missing the `if __name__ == '__main__'` block so I added them
* Had to make a small front-end change to `Tabs` in order for all the tabs to wrap around

### Notes for discussion
1. Should we run a subset of demos or all demos?
2. Need to fix the caching examples issue.
3. Chatbot seems to be broken

# Checklist:


abidlabs:
So cool!

demo/image_classifier_interface_load/run.py
```diff
@@ -1,10 +1,13 @@
...
 
-images = ["cheetah1.jpeg", "cheetah1.jpg", "lion.jpg"]
+current_dir = pathlib.Path(__file__)
+
+images = [current_dir / "cheetah1.jpeg", current_dir / "cheetah1.jpg", current_dir / "lion.jpg"]
 
 
 img_classifier = gr.Interface.load(
-    "models/google/vit-base-patch16-224", examples=images, cache_examples=True
+    "models/google/vit-base-patch16-224", examples=images, cache_examples=False
```
abidlabs, line 10: Why do we need to set `cache_examples=False`?

scripts/copy_demos.py
```diff
@@ -0,0 +1,59 @@
...
+    all_demos = []
+    for p in os.listdir(source_dir):
+        full_dir = os.path.join(source_dir, p)
+        is_dir = os.path.isdir(full_dir)
+        no_requirements = not os.path.exists(os.path.join(full_dir, "requirements.txt"))
+        has_app = os.path.exists(os.path.join(full_dir, "run.py"))
+        if is_dir and no_requirements and has_app:
+            all_demos.append(p)
+
+    to_exclude = [
```
abidlabs, line 18: Why are we excluding these ones?

FarukOzderim:
Looks fantastic! Though isn't all the demos a bit overboard. It looks too crowded, could be hard to do manual testing for every big release. Would it be better if we select the demos we want to test, or is it good like this?

scripts/copy_demos.py
abidlabs: If there's some sort of underlying issue affecting all of these demos, it might also affect new demos that we create in the future and break this script. So perhaps better to select a few demos to create instead.

abidlabs:
@freddyaboulton this looks great. I had some clarification questions, but otherwise the PR LGTM.

My thoughts:

> Should we run a subset of demos or all demos?

Agreeing with @FarukOzderim and continuing my point above, launching all of the demos (except for a few that we exclude manually) might be overkill and lead to a flaky script. Instead, how about we select a ~10 demos and launch them manually. For example, we could include `calculator`, `kitchen_sink`, `interface_random_slider`, `blocks_layout`, `blocks_essay`, `sst_or_tts`, and a few others.

@aliabd we could choose the same ones for the Recipes section

> Need to fix the caching examples issue.

Yes, let's figure out which one it is. Perhaps it's fixed by #1967 

> Chatbot seems to be broken

cc @dawoodkhan82

freddyaboulton:
@abidlabs @FarukOzderim Thank you for the review! 

Yea sounds good re: running a subset of demos. I will take the ones you mentioned and some other ones that test `Series`, `Parallel` and `Examples` as well as image and video demos. Will cap it at around ~20 though! I will make sure all of those demos look good and can cache examples.

scripts/copy_demos.py
freddyaboulton: There was a lot of iteration here so I forgot exactly why each was excluded 😅 Will select a hard-coded list to avoid flakiness.

FarukOzderim:
Btw should we merge the kitchen sink in the webpage @aliabd was working on recently with this and embed it in on the webpage? I could not find the PR, but I would make updating the kitchen sink easier.

aliabid94:
Beautiful work.

aliabid94 (approved):

scripts/copy_demos.py
```diff
@@ -0,0 +1,61 @@
...
+    source_dir = pathlib.Path(pathlib.Path(__file__).parent, "..", "demo")
+    dest_dir = pathlib.Path(
+        pathlib.Path(__file__).parent, "..", "demo", "all_demos", "demos"
+    )
+    copy_all_demos(source_dir, dest_dir)
+    reqs_file_path = pathlib.Path(
+        pathlib.Path(__file__).parent, "..", "demo", "all_demos", "requirements.txt"
+    )
+    requirements = f"""
+    {args.gradio_version}
```
FarukOzderim, line 54: I recently experienced this,
In hugging face spaces defining gradio version in requirements.txt file does not make it work. You need to define it in README.md, see [here](https://huggingface.co/spaces/farukozderim/test-new-queue/blob/main/README.md#L7).

Also what do you think of using packages in test/requirements.txt instead of defining each manually?

freddyaboulton: It'll work as long as the version in requirements.txt matches the latest stable version (what's listed in README by default). Currently working on a fix so that this whole process still works when you change `version.txt` in a release PR.

freddyaboulton:
Thank you everyone on the reviews and suggestions!!!

scripts/copy_demos.py
freddyaboulton: @FarukOzderim Regarding packages in `test/requirements.txt` - If we were running all demos I think that may be a more appropriate solution but since we're only running a subset, I think manually specifying the minimal set of required packages is better (faster build with fewer packages).

Issue #1811, opened 2022-07-17
Flag can not create an appropriate CSV file when a textbox / text has a comma  


kmkarakaya:
### Describe the bug

Flag can not create an appropriate CSV file when the interface have a textbox / text which includes a comma .
The CSV file columns are incorrectly created assuming the comma in the text box is a separator.
See the attached screenshots.

Thank you!
 

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

kmkarakaya/Auto_Review_Generation_in_Turkish

### Screenshot

![image](https://user-images.githubusercontent.com/41159849/179424201-69235b8c-dcb1-4fd8-a9ad-88529d67d5b1.png)
![image](https://user-images.githubusercontent.com/41159849/179424304-7e14215f-6bd2-4f2b-b4d8-fdc376653b3d.png)
![image](https://user-images.githubusercontent.com/41159849/179424350-10e04b25-7fa0-4bba-bf1b-0bc24449bb4c.png)

### Logs

```shell
'Prompt'	'Generated Text'	'flag'	'username'	'timestamp'	
'Tatil iÃ§in rezervasyon yaptÄ±rdÄ±m'	'Tatil iÃ§in rezervasyon yaptÄ±rdÄ±m. Ä°ptal Edilen Tur	 tur ile tatil iÃ§in 3 ay sonra otel tur turu iptali iÃ§in 2 adet rezervasyon yaptÄ±m. 2 gÃ¼n Ã¶ncesinden tatil iptal edildi. Ancak iptal etmek istedik. Fakat Ã¼cret'	''	''	'2022-07-17 20:35:02.542229'
'Bir hafta Ã¶nce aldÄ±ÄŸÄ±m cep telefonu'	'Bir hafta Ã¶nce aldÄ±ÄŸÄ±m cep telefonu sÃ¼rekli meÅŸgul	 arama yapamÄ±yorum. HiÃ§bir cevap alamadÄ±m. MÃ¼ÅŸteri hizmetleri ile ilgili ne bir bilgi alamÄ±yorum. Ã‡aÄŸrÄ± merkezi de ne kadar mail yoluyla dÃ¶nÃ¼ÅŸ yapÄ±lmadÄ±. BÃ¶yle bir ÅŸey gÃ¶rmedim.'	''	''	'2022-07-17 20:36:11.325220'
[...]

PR #2017, opened 2022-08-15
Sanitize for CSV

abidlabs:
I went back and forth on how to best implement this, but I decided on the following approach:
* If any unsafe characters are detected in the cell value, prepend it with a single quotation mark, which prevents it from being run as a formula in Excel / LibreOffice
* Otherwise, the value in unchanged

I'm hoping that for 99% of the users, this means they can read the flagged data like a regular CSV file and not worry about any changes introduced by this quotation.

The reason I went with this approach:
* instead of quoting everything with double and single quotes ([the 2nd solution here](https://owasp.org/www-community/attacks/CSV_Injection)) is because this approach allows the data to be read in a regular CSV format (99% of the time)
* instead of not sanitizing anything: was to prevent CSV injection (which this PR should do, in fact better than how things were before) 
* instead of leaving things as they were before on `main`: was to fix issues like #1811 

Fixes: #1811

### DIFF
--- a/demo/kitchen_sink/run.py
+++ b/demo/kitchen_sink/run.py
@@ -151,7 +151,6 @@ demo = gr.Interface(
     * 3,
     theme="default",
     title="Kitchen Sink",
-    cache_examples=False,
     description="Try out all the components!",
     article="Learn more about [Gradio](http://gradio.app)",
     cache_examples=True
--- a/demo/kitchen_sink/run.py
+++ b/demo/kitchen_sink/run.py
@@ -153,7 +153,7 @@ demo = gr.Interface(
     title="Kitchen Sink",
     description="Try out all the components!",
     article="Learn more about [Gradio](http://gradio.app)",
-    cache_examples=True
+    cache_examples=False
 )
 
 if __name__ == "__main__":
--- a/gradio/utils.py
+++ b/gradio/utils.py
@@ -612,3 +612,9 @@ def set_directory(path: Path):
 
 def strip_invalid_filename_characters(filename: str) -> str:
     return "".join([char for char in filename if char.isalnum() or char in "._- "])
+
+
+def sanitize_for_csv(value: str) -> str:
+    value = value.replace("\"", "\"\"")
+    return "\"'" + value + "\""
+
--- a/gradio/utils.py
+++ b/gradio/utils.py
@@ -15,6 +15,7 @@ from contextlib import contextmanager
 from copy import deepcopy
 from distutils.version import StrictVersion
 from enum import Enum
+from numbers import Number
 from pathlib import Path
 from typing import (
     TYPE_CHECKING,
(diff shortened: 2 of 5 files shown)


demo/kitchen_sink/run.py
```diff
@@ -151,10 +151,9 @@ def fn(
     * 3,
     theme="default",
     title="Kitchen Sink",
-    cache_examples=False,
     description="Try out all the components!",
     article="Learn more about [Gradio](http://gradio.app)",
-    cache_examples=True
```
freddyaboulton, line 157: How come we need to not cache examples?

gradio/utils.py
```diff
@@ -612,3 +613,37 @@ def set_directory(path: Path):
...
+    values: List[str | Number] | List[List[str | Number]],
+) -> List[str | Number] | List[List[str | Number]]:
+    """
+    Sanitizes a list of values (or a list of list of values) that is being written to a
+    CSV file to prevent CSV injection attacks.
+    """
+    sanitized_values = []
+    for value in values:
+        if isinstance(value, list):
+            sanitized_value = sanitize_list_for_csv(value)
```
freddyaboulton, line 644: I think a list comprehension would be better here than the recursion

```python
sanitized_value = [sanitize_value_for_csv(v) for v in value]
```

abidlabs:
@freddyaboulton @aliabid94 @FarukOzderim I'm going to merge this into the other PR so that I don't duplicate efforts, but it would be great if you can provide your feedback here!

demo/kitchen_sink/run.py
abidlabs: Do we want to cache examples? I'm fine either way, but it's just a demo, right?

freddyaboulton:
@abidlabs Lol I was about to approve! Looks good, was able to handle everything I could think of to try to break it. Good unit tests as well!

What I tried

```python
import gradio as gr

gr.Interface(lambda s: s, gr.Text(label="@invalid"), gr.Text(label="=hack"),
            flagging_options=["@very bad", ",escape this please", "\ngood"]).launch()
```

Was able to load the csv well afterwards:
image

abidlabs:
Niiice thanks for the review @freddyaboulton!

demo/kitchen_sink/run.py
freddyaboulton: Was thinking it could help catch accidental breaking changes when it comes to caching examples since this demo will run on spaces for prs.

gradio/utils.py
abidlabs: Thanks, updated!

demo/kitchen_sink/run.py
abidlabs: Ah yeah makes sense. Will remove the cached examples argument so it caches on Spaces

PR #2026, opened 2022-08-17
Update Queue API, documentation

aliabid94:
Made Queue API a little clearer, added docstrings and documentation to site. Remove print statements. Improved estimation logic (no need to store duration history as a list)

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -10,7 +10,17 @@ import time
 import warnings
 import webbrowser
 from types import ModuleType
-from typing import TYPE_CHECKING, Any, AnyStr, Callable, Dict, List, Optional, Tuple, Literal
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    AnyStr,
+    Callable,
+    Dict,
+    List,
+    Literal,
+    Optional,
+    Tuple,
+)
 
 import anyio
 import requests
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -10,16 +10,7 @@ import time
 import warnings
 import webbrowser
 from types import ModuleType
-from typing import (
-    TYPE_CHECKING,
-    Any,
-    AnyStr,
-    Callable,
-    Dict,
-    List,
-    Optional,
-    Tuple,
-)
+from typing import TYPE_CHECKING, Any, AnyStr, Callable, Dict, List, Optional, Tuple
 
 import anyio
 import requests
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -17,7 +17,6 @@ from typing import (
     Callable,
     Dict,
     List,
-    Literal,
     Optional,
     Tuple,
 )
@@ -783,7 +782,7 @@ class Blocks(BlockContext):
     def queue(
         self,
         concurrency_count: int = 1,
-        status_update_rate: float | Literal["auto"] = "auto",
+        status_update_rate: float | str = "auto",
         client_position_to_load_data: int = 30,
         default_enabled: bool = True,
     ):
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -7,6 +7,7 @@ import os
 import random
 import sys
 import time
+import warnings
 import webbrowser
 from types import ModuleType
 from typing import TYPE_CHECKING, Any, AnyStr, Callable, Dict, List, Optional, Tuple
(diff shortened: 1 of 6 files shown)


gradio/blocks.py
```diff
@@ -768,28 +769,32 @@ def clear(self):
         self.children = []
         return self
 
-    def configure_queue(
+    @document()
+    def queue(
```
abidlabs, line 773: Very clean, love it!

gradio/blocks.py
```diff
@@ -768,28 +769,32 @@ def clear(self):
...
-            update_intervals: Queue will send estimations to the clients in intervals=update_intervals when live_queue_updates==false
-            duration_history_size: Queue duration estimation calculation window size.
+            concurrency_count: Number of worker threads that will be processing requests concurrently.
+            status_update_rate: If "auto", Queue will send status estimations to all clients whenever a job is finished. Otherwise Queue will send status at regular intervals set by this parameter as the number of seconds.
+            client_position_to_load_data: Once a client's position in Queue is less that this value, the Queue will collect the input data from the client. You may make this smaller if clients can send large volumes of data, such as video, since the queued data is stored in memory.
+            default_enabled: If True, all event listeners will use queueing by default.
+        Example:
+            demo = gr.Interface(gr.Textbox(), gr.Image(), image_generator)
```
abidlabs, line 788: use string shortcuts in Interface for simplicity

gradio/blocks.py
```diff
@@ -768,28 +769,32 @@ def clear(self):
...
-    def configure_queue(
+    @document()
+    def queue(
         self,
-        live_queue_updates: bool = True,
         concurrency_count: int = 1,
-        data_gathering_start: int = 30,
-        update_intervals: int = 5,
-        duration_history_size: int = 100,
+        status_update_rate: float | str = "auto",
```
abidlabs, line 776: Better to use `float | Literal["auto"]` for the type hint

gradio/event_queue.py
```diff
@@ -31,31 +31,29 @@ class Queue:
...
-        live_queue_updates: bool,
-        queue_concurrency_count: int,
+        live_updates: bool,
+        concurrency_count: int,
         data_gathering_start: int,
         update_intervals: int,
-        duration_history_size: int,
     ):
         """
         See Blocks.configure_queue() docstring for the explanation of parameters.
```
abidlabs, line 51:
```suggestion
        See Blocks.queue() docstring for the explanation of parameters.
```

guides/1)getting_started/2)key_features.md
```diff
@@ -119,22 +119,31 @@ with gr.Interface(css="body {background-color: red}") as demo:
...
+demo = gr.Interface(...).queue()
+demo.launch()
 ```
 
 With `Blocks`:
 ```python
 with gr.Blocks() as demo:
     #...
+demo.queue()
 demo.launch(enable_queue=True)
```
abidlabs, line 135:
```suggestion
demo.launch()
```

abidlabs:
Small nits, but overall LGTM! Very clean

gradio/blocks.py
FarukOzderim: I think we should not use string shortcuts cuz we removed them from our docs and we just support backwards compatibility.

FarukOzderim (approved):
Documentation and naming for queue function is miles ahead of mine congrats 🎉
The only point I could make is since we change duration history total and count and calculation why don't we take the average or calculate estimations function index by function index, there was an issue around it and could be great if you want to handle it right now which would make our etas more precise.

FarukOzderim:
By the way if we don't use last x number of durations for calculating ETA the calculation could get ugly when the queue size gets very big because when the queue size is Big there's a lot of communication going on which will slow the processes of events.

gradio/blocks.py
aliabid94: Agree with @FarukOzderim.

aliabid94:
Let's bring in function specific ETA's later.

aliabid94:
> By the way if we don't use last x number of durations for calculating ETA the calculation could get ugly when the queue size gets very big because when the queue size is Big there's a lot of communication going on which will slow the processes of events.

I don't understand why things would be any different with "last x" calculation?

gradio/blocks.py
aliabid94: not supported in 3.7

PR #2027, opened 2022-08-17
Fix loading status on website

### LINKED ISSUE #1960: No loading icon or processing message on embedded demos
### Describe the bug

When demos are embedded as web componenets, there is no loading/processing feedback on a submit. 

### Reproduction

See demos embedded in https://gradio.app

### Screenshot

https://user-images.githubusercontent.com/9021060/183219540-16384984-29ac-47cc-8b0f-aa77811037a4.mov

### Logs

```shell
No logs
```

### System Info

```shell
3.1.1
```

### Severity

annoying

aliabd:
Fixes #1960


abidlabs:
Thanks for the fix, @aliabd! I'm testing it locally (by running the `scripts/launch_website.sh`), and I don't see the issue fixed (despite a hard refresh):

image

Let me know if there's anything else I should do to test it.

(As an aside, there seem to be typing issues caught by the static checks)

aliabd:
@abidlabs Since this is a change to the frontend its pretty complicated to use the scripts to load the website locally (without docker). There are a few steps in the Dockerfile missing in those scripts, especially [this line](https://github.com/gradio-app/gradio/blob/5fa94e5f4c0f2b012e05592a2618d7bad5182473/website/homepage/Dockerfile#L41).

Also I just realized that the landing page is built with index.js loaded from a pinned version of gradio, because of this [PR](https://github.com/gradio-app/gradio/pull/1899). I just pushed a commit that makes it use the generated index file like all the other demos, and I think we should instead update the referenced spaces often so they always run the same gradio version. 

Anyway, just built the website on gradio dev server and everything looks good, try it out here: http://54.189.115.49/


https://user-images.githubusercontent.com/9021060/185043366-9324b9cd-c2ba-4eef-b7ee-d9aceeeab4d4.mov



BTW this is the first time I've tested the website... How I've been testing these changes was by embedding multiple spaces directly onto `ui/packages/app/index.html` using `<gradio-app space="">`, building the frontend, and launching any demo. Also playing around with `window.__gradio_mode__ = "app";` vs `window.__gradio_mode__ = "website";`, which now are the same with regards to loading status. 

Let me know if this all makes sense... I'm happy to huddle on this too. 

Will look into the typing issues.

aliabd:
To be clear, here's a before and after showing what used to happen when multiple interfaces were embedded on the same page:

https://user-images.githubusercontent.com/9021060/185044537-f9a2b1a8-5481-4b97-ad6f-5bcb5511671c.mov




https://user-images.githubusercontent.com/9021060/185044570-852f5ab9-1a2c-4121-a5cf-fdde9c37f0a5.mov


To avoid this we used `window.__gradio_mode__ = "website"` as a flag to remove the loading status altogether from the wbesite

abidlabs:
Awesome, checked on the dev server and confirmed that it fixes the issue. LGTM! 

Some small things (that we could do in a separate PR if it's easier):
* Noticed that clearing the sketchpad causes the loading to appear. It would be great if clear was instantaneous
* We can bring demo #4 (the speech one) back
* Still having the typing issue, which is actually maybe not a typing issue, but rather a missing function?

```js
Error: Property 'update' does not exist on type 'LoadingStatusCollection'. 
		if (queue && ["predict", "interpret"].includes(action)) {
			loading_status.update(fn_index as number, "pending", null, null);
```

abidlabs (approved):

aliabd:
@abidlabs its a typing issue, still working on it

aliabd:
@abidlabs I fixed the typing, but @aliabid94 should review as I'm not sure if they're all actually `Function`

aliabid94 (approved):
LGTM

Issue #2019, opened 2022-08-16
Gracefully Scaling Down on Spaces With the new Queue


omerXfaruq:
**Is your feature request related to a problem? Please describe.**  
Scaling up works alright however Scaling down is a pain point right now. When we tested scaling down, clients lost connection to the server and queue info stopped getting updated.  
<img width="758" alt="image" src="https://user-images.githubusercontent.com/44067394/184917082-34e16681-3b81-4b8c-bfd6-c4fe3706c5cc.png">

**Describe the solution you'd like**  
There are 3 ways we can solve this.
1. If autoscale can scale down(kill an instance) when queue-size==0 (easiest one) (Christophe Rannou will let us know about this)
2. Take a look at what Gradio do at the cancel signal(I think this is same as ctrl+c) and wait until to Queue to finish when we receive the signal.
3. Add timestamp to the each event and let client join into another queue whenever the ws connection is lost. And the server would put the event to a rank according to its timestamp 
(This would be useful if mobile connections' shaky internet connections will cause ws connection to get lost, might not ever happen as well)

I would prefer 1>3>2

abidlabs:
I don't remember what this issue was about? Does anyone? Or should we close this?

freddyaboulton:
Don't remember either

abidlabs:
Ok let's go ahead and close this

PR #2036, opened 2022-08-18
Queue maximum length

aliabid94:
Added queue parameter `max_size` to limit queue size. 

(For now, based only on number of requests, not time. We can add maximum eta logic if requested easily).


apolinario:
It is not working right now. Whenever it is present, it just keeps 'executing' without actually processing. As soon as I remove the `max_size` setting the app works again. Regardless of the `max_size` value

image

This is a minimal replication I have tested:
```py
import gradio as gr

block = gr.Blocks()

def run(text):
  return text

with block: 
  text = gr.Textbox(label="Enter your prompt", show_label=False, max_lines=1)
  textout = gr.Textbox(show_label=False, max_lines=1)
  btn = gr.Button("Run")
  text.submit(run, inputs=text, outputs=textout)
  btn.click(run, inputs=text, outputs=textout)

block.queue(concurrency_count=2, max_size=2).launch(enable_queue=True)
```
All was tested on Spaces

apolinario:
Actually, this is happening no matter what I do `3.5.0b1` - simply `3.5.0b1` seems broken and never processes requests

ui/packages/app/src/api.ts
```diff
@@ -142,6 +142,11 @@ export const fn =
 					case "send_data":
 						send_message(fn_index, payload);
 						break;
+					case "queue_full":
+						alert("Queue full. Try again later.");
```
apolinario, line 146: Instead of an alert, do you think when the feature gets matured, this could be sent as an error to the output field requested? With this same message instead of the generic "ERROR"

apolinario: Actually, could we let this message be set up by the user? If not I would suggest I would change it to something like: 
"This application is too busy! Try again soon."

gradio/routes.py
```diff
@@ -283,6 +283,8 @@ async def join_queue(websocket: WebSocket):
                 return
             event.hash = e_hash["hash"]
             rank = Queue.push(event)
+            if rank is None:
+                await event.send_message({"msg": "queue_full"})
```
FarukOzderim, line 287: let's disconnect the ws here? No need to continue running it IMO. Btw we will be handling ws disconnections in Frontend in the future(#2019), so it is ok if you want to always dc from frontend and handle in backend.

`await event.disconnect()
`we could even disconnect with a code if you can utilize that in frontend
`await event.disconnect(429)`

FarukOzderim (commented):
LGTM except a few points, good work!

ui/packages/app/src/api.ts
aliabid94: Done

gradio/routes.py
aliabid94: Done

gradio/event_queue.py
```diff
@@ -112,6 +115,8 @@ def push(cls, event: Event) -> int:
         Returns:
             rank of submitted Event
         """
+        if cls.MAX_SIZE is not None and len(cls.EVENT_QUEUE) >= cls.MAX_SIZE:
```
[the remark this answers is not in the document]
aliabid94: Done

FarukOzderim (approved):

PR #2038, opened 2022-08-18
Max thread tweaks

FarukOzderim:
# Description
1. move creation of limiter to fastapi endpoint since run_coro_in_background works only in there
2. make max_threads compatible with queue concurrency count
3. add warning to run_coro_in_background

Tested limiter creation with putting a print into `Blocks.predict_api` and tested both queue and nonqueue demos, both works alright!

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -332,6 +332,7 @@ class Blocks(BlockContext):
         self.encrypt = False
         self.share = False
         self.enable_queue = None
+        self.max_threads = 40
         if css is not None and os.path.exists(css):
             with open(css) as css_file:
                 self.css = css_file.read()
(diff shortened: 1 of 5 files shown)


gradio/blocks.py
```diff
@@ -881,14 +884,15 @@ def reverse(text):
...
             )
-        if self.is_space and self.enable_queue is None:
-            self.enable_queue = True
-        if self.enable_queue is None:
-            self.enable_queue = False
-        utils.run_coro_in_background(self.create_limiter, max_threads)
+        if self.is_space:
+            self.enable_queue = enable_queue is not False
+        else:
+            self.enable_queue = enable_queue is True
```
FarukOzderim, line 890: This logic felt incorrect, hence changed it.

FarukOzderim: Now that I think about it, it was not incorrect :D

FarukOzderim:
Merging this for now, please LMK about your reviews.

gradio/blocks.py
freddyaboulton: You're going to undo this change right?

freddyaboulton (commented):
@FarukOzderim Looks good to me! Thank you for adding the warning to the docstring too!

The one thing is that I think we should revert the change around `is_space and enable_queue` ?

gradio/blocks.py
FarukOzderim: Nope, since it works like this as well :)

freddyaboulton: You're right! I think the old way was easier to understand though 🙈

PR #2041, opened 2022-08-18
dataset-fixes

abidlabs:
Noticed some small issues with the Dataset component


freddyaboulton (approved):
@abidlabs Looks good to me! 

Should we add a label parameter to `gr.Examples`? Users don't normally interact with the Dataset component directly.

abidlabs:
Good suggestion @freddyaboulton. Will add and merge

PR #2044, opened 2022-08-19
Reset components to original state by setting value to None

### LINKED ISSUE #1672: Make it easier to reset components to their original state when the app first launches
### Describe the bug

On occasions, one wants to set back a component back to initial state. In my situation, I have a 2-steps app, where one predicts a label for a text, then the second explains how the label got obtained. When the user clears the app or starts typing new content, I want to move the component back to the initial look (with the image).

1. Setting `value` back to `None` fails to do this and nothing happens.
2. I can trick the component with `value` set to empty list `[]`, but this removes the image and changes the size.

### Reproduction

```{python}
demo = gr.Blocks()

with demo:
    text = gr.Textbox()
    submit_btn = gr.Button(value="Submit")
    explanation = gr.HighlightedText()
    samples = gr.Dataset(
        components = [text],
        samples = [["Example of a text. This should reset highlighted text but does not."]]
    )
    submit_btn.click(
      fn = lambda text: gr.HighlightedText.update(value=[("Fake output", 1)]), inputs=[text], outputs=[explanation]
    )
    samples.click(
      fn = lambda sample: gr.Textbox.update(value=sample[0]), inputs=[samples], outputs=[text]
    )
    samples.click(
      fn = lambda: gr.HighlightedText.update(value=None), inputs=[], outputs=[explanation]
    )

demo.launch()
```

### Screenshot

At launch :
Capture d’écran 2022-06-30 à 16 45 09
Resetting `value` to `None` (as in snippet) does not do anything.

If instead you reset `value` to `[]` :
Capture d’écran 2022-06-30 à 16 45 29

### Logs

```shell
[no error in the logs]
```

### System Info

```shell
Gradio 3.0.20
MacOs Monterey 12.4
Brave (Chromium 1.39.111)
```

### Severity

annoying

freddyaboulton:
# Description

Fixes #1672

Previously, it was unclear how to reset components to their original state. As stated in the original issue, the "reset" value depended on the component, e.g. `""` for textbox and `[]` for HightlightedText, and `None` would confusingly not do anything.

This PR makes it so that `None` will always reset the component to its unset state. The reason `None` would originally work is that the `delete_none` util method would delete the `value` field from the update dict if it was None. 

To test:
- `python demo/kitchen_sink_random/reset_component_demo.py` and hit the `Reset` button
- unit tests

# Checklist:


gradio/blocks.py
```diff
@@ -600,10 +600,8 @@ def postprocess_data(self, fn_index, predictions, state):
                             )
                         prediction_value = delete_none(prediction_value)
                         if "value" in prediction_value:
-                            prediction_value["value"] = (
-                                block.postprocess(prediction_value["value"])
-                                if prediction_value["value"] is not None
-                                else None
+                            prediction_value["value"] = block.postprocess(
```
freddyaboulton, line 603: All components can now postprocess `None`

gradio/components.py
```diff
@@ -3944,7 +3944,7 @@ def postprocess(self, y: List[Tuple[str, str]]) -> List[Tuple[str, str]]:
         Returns:
             List of tuples representing the message and response
         """
-        return y
+        return [] if y is None else y
```
freddyaboulton, line 3947: To reset the chatbot

gradio/test_data/blocks_configs.py
```diff
@@ -61,13 +61,20 @@
                 "show_label": True,
                 "name": "image",
                 "visible": True,
+                "value": None,
```
freddyaboulton, line 64: None is no longer filtered out from the config.

demo/kitchen_sink_random/reset_component_demo.py
```diff
@@ -0,0 +1,18 @@
+import gradio as gr
+from run import demo
+
+
+no_reset = [gr.Button, gr.Variable]
```
freddyaboulton, line 5: Not sure if it's appropriate to reset these components with None

gradio/utils.py
```diff
@@ -279,6 +279,8 @@ def delete_none(_dict):
     """
     if isinstance(_dict, dict):
         for key, value in list(_dict.items()):
+            if key == "value":
```
abidlabs, line 282: Suggest adding something in the docstring to mention the fact that we exclude keys whose name is "value" and the rationale (in case someone is reading the code later wonders why)

abidlabs:
Nice fix @freddyaboulton! And great tests and demo. You might consider putting the demo in its own separate folder and in a file named `run.py`. There are a few places in the codebase where we assume demos are all in files called `run.py`, for example if we try to reference their code in the website. (Nit.)

abidlabs (approved):

freddyaboulton:

PR #2046, opened 2022-08-19
Hotfix: Fix clear button with queue

### LINKED ISSUE #2045: Interface Clear button is broken on spaces
### Describe the bug

The _js parameter does not seem to work on spaces. This means the clear button doesn't work as well.

This is in gradio 3.1.5. I think it was a recent change because that was not the case as of this Monday since this is not the case in the demos deployed for #2012 

This also doesn't happen locally.

### Clear button working on Monday (PR #2012)
interface_working_previous

### Clear button broken 3.1.5 (PR #2033)
interface_broken_3 1 5

### Screenshot

Logs - noticed we log a lot of things to the console?

image

### Logs

```shell
-
```

### System Info

```shell
Spaces
```

### Severity

freddyaboulton:
# Description

Fixes #2045 

This is a hotfix. I think the root cause is that the front-end is not processing events without a backend_fn when the queue is enabled. The backend is not getting any requests to add the queue button to the queue.

Bypassing the queue for the clear button seems to fix it. I think that should have been the case all along but there might be a better fix. Maybe @aliabid94 knows. 

### Testing
On main, run `calculator` with queue enabled. The clear button will not work. On this branch, it will. 

# Checklist:


aliabid94:
Oh yikes, good catch @freddyaboulton !

freddyaboulton:
This will fix the clear button but not the "general problem". `blocks_js_methods` is still broken because the "Reverse sentence" and "Foo bar." buttons don't do anything. Interestingly, the `verb` is still reversed because that one does use a backend function.

aliabid94:
Actually, I think we can fix this in `blocks.py` line 154, such as
```python
{
   ...,
   "queue": queue and fn is not None,
}
```

I think we can update to a new gradio version too since Clear is broken on all Spaces currently.

freddyaboulton:
Awesome, will look into that fix @aliabid94 !!

should we yank 3.1.5? Broken Clear button seems like a pretty big regression for a minor version bump.

aliabid94:
I think bumping to 3.1.6 asap is the best move, let's discuss yanking in the slack channel

aliabd (approved):
LGTM, tested locally and it works

aliabid94 (approved):

Issue #956, opened 2022-04-08
Give Descriptive Error Messages


FarukOzderim:
We need to give more descriptive error messages to users, and maybe give out a try again message to let them try again.

Context, a question coming from @ronvoluted:

> When the output gives an error due to backend issues (as opposed to bad input), how often is it that submitting a second time is enough to make it work? If it's very often, changing "ERROR" to "Try again" could prevent some users from bouncing

radames:
I came across this. What's the best way to validate inputs? and/or to avoid running the prediction function without triggering queuing and loading, raising errors that communicate to the UI

```python

def speech_to_text(audio_file_path):
    if(audio_file_path == None):
        raise Exception("No file path provided")


...
        btn.click(speech_to_text, audio, text)
```

FarukOzderim:
Well you could direct it to your outputs. I don't think there currently is a method for raising errors in UI. But thx for sharing we shall consider user made errors within UI as well.

abidlabs:
It would be awesome if people can provide their own custom error messages for events. We could accept an optional (str) "error_message" parameter in an event listener that would be displayed instead of the default error.

abidlabs:
Feedback from @apolinario on this issue: a global replacement for ERROR doesn't always work. You might want to throw different errors depending on what happened (OOM vs. NSFW image generated)

> I would like to specify my own error message if the error is thrown by me - in this case would be "Error: no avaliable GPUs, try again!"

> An advanced version of this would allow people to specificy text messages from different potential errors that can happen "internally" on their models

FarukOzderim:
A bit Related: We have exception types in our Request [class](https://github.com/gradio-app/gradio/blob/main/gradio/utils.py#L418). And can define error types similarly and use them. Also we can set event-exception type while we are sending a Request from Queue to fastapi server and make use of it.

FarukOzderim:
I think this issue should stay open since it is not just User defined errors, but we should also give proper errors for the errors happened in frontend or backend.

abidlabs:
> I think this issue should stay open since it is not just User defined errors, but we should also give proper errors for the errors happened in frontend or backend.

I think we can close this issue actually because the best place for such errors would be the logs (on Spaces) or your Python console (if you're running locally). A typical end user shouldn't see Python errors by default. If a model creator wants to show particular errors to the end user, they can use `gr.Error()` via #2047

FarukOzderim:
OK, though let me remind the original question
```
[...]

PR #2047, opened 2022-08-19
Queue error fix & custom errors

aliabid94:
If there's an error within the function when queueing is enabled, it would not show ERROR. now it does. (also shows error message if `show_error=True`).

Also fixed submit button color, gradient was weird on hover.

Fixes: #956


gradio/exceptions.py
```diff
@@ -0,0 +1,4 @@
+class Error(Exception):
```
abidlabs, line 1: I think it's good to add a `__str__()` method, following this tutorial: https://docs.python.org/3.3/tutorial/errors.html#user-defined-exceptions

guides/1)getting_started/2)key_features.md
```diff
@@ -11,6 +11,10 @@ $demo_calculator
 
 You can load a large dataset into the examples to browse and interact with the dataset through Gradio. The examples will be automatically paginated (you can configure this through the `examples_per_page` argument of `Interface`).
 
+## Errors
+
+You wish to pass custom error messages to the user. To do so, raise a `gr.Error()` to display an error message. See what happens if you try to divide by zero in the the calculator demo above.
```
abidlabs, line 16: Describe what the error message looks like

abidlabs:
This is really awesome @aliabid94! I left some very small nits as comments above. A couple more things:

* IIRC `show_error=True` is supposed to print the error in the console. That's not happening if a `gr.Error()` is raised. I guess if the error is being flashed as a modal, you don't really need this, but I think it would be good to preserve the behavior. Try running `fake_gan2` for example:

image

* Overall, I like the design of the error, but maybe it's a little *too* bold and covers too much of the demo? Maybe we can get @gary149's feedback on this.

image

* This PR closes #956 so I've updated the PR body to reflect that.

Nice work!

Issue #1767, opened 2022-07-12
TimeSeries examples not properly displayed in UI


freddyaboulton:
### Describe the bug

When running an app with a `TimeSeries` example, the time series data is not displayed in the examples dataset and when you click on the example, the time series component in the UI collapses.

The part about collapsing is strange since that doesn't happen when you manually upload a time series dataset.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Run the `demo/fraud_detector/run.py` script. 

## TimeSeries collapsing when selecting example
![fraud_example_not_displayed](https://user-images.githubusercontent.com/41651716/178526057-61366aeb-033b-4835-a8bb-6d5e452d2307.gif)

## TimeSeries displayed properly on manual upload
![fraud_example_displayed_upload](https://user-images.githubusercontent.com/41651716/178526156-7a60cfe4-6098-460e-991b-6d829cd7d9ce.gif)

### Logs

```shell
-
```

### System Info

```shell
Gradio 3.0.24
```

### Severity

annoying

PR #2064, opened 2022-08-23
Fix TimeSeries examples not properly displayed in UI

dawoodkhan82:
# Description
Fix TimeSeries Example

https://user-images.githubusercontent.com/12725292/186215020-ae8ff107-ce5e-4f5d-8d6d-703dae033616.mov

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

Closes: #1767

# Checklist:


gradio/components.py
```diff
@@ -2519,8 +2519,7 @@ def postprocess(self, y: str | pd.DataFrame) -> Dict:
         if y is None:
             return None
         if isinstance(y, str):
-            y = pd.read_csv(y)
-            return {"headers": y.columns.values.tolist(), "data": y.values.tolist()}
```
pngwn, line 2523: All return values have to have headers now.

pngwn: Oh maybe this is only true for Dataframe, ideally we'd make the interfaces consistent since they are very similar. What is the reason for this change, how does it address the issue?

dawoodkhan82: Removed this change, and instead changed the frontend to convert the timeseries obj to string (to fix examples).

freddyaboulton (approved):
Thanks for the fix @dawoodkhan82 ! Confirm it works locally for me. 

Code change makes sense to me but defer to you since you are more familiar with it!

ui/packages/app/src/components/TimeSeries/TimeSeries.svelte
```diff
@@ -70,10 +70,22 @@
 		reader.readAsText(blob);
 	}
 
+	function dict_to_string(dict: any) {
```
pngwn, line 73: Can we add a better type for this. Looks like this should be something like:

```ts
interface TimeSeriesDict {
  data: Array<Array<unknown>>;
  headers: Array<string>;
}
```

ui/packages/app/src/components/TimeSeries/TimeSeries.svelte
```diff
@@ -70,10 +70,22 @@
...
+		_value = dict["headers"].join(",");
```
pngwn, line 74:
```suggestion
		_value = dict.headers.join(",");
```

ui/packages/app/src/components/TimeSeries/TimeSeries.svelte
```diff
@@ -70,10 +70,22 @@
...
+
+		dict["data"].forEach((x: Array<any>) => {
+			_value = _value + "\n";
+			_value = _value + x.join(",");
+		});
```
pngwn, line 79:
```suggestion
		dict.data.forEach((x: Array<unknown>) => {
			_value = _value + "\n";
			_value = _value + x.join(",");
		});
```

pngwn (dismissed):

ui/packages/app/src/components/TimeSeries/TimeSeries.svelte
pngwn: Actually, we already have some compatible types here: https://github.com/gradio-app/gradio/blob/cbaa090146010a41feec5338192c6992c517eeec/ui/packages/app/src/components/TimeSeries/TimeSeries.svelte#L25-L32

We might need to  add some conditional checks in the `dict_to_string` function but they should work otherwise.

abidlabs:
LGTM, works for me locally. It looks like @pngwn's comments were resolved, so let's go ahead and merge this @dawoodkhan82?

abidlabs (approved):

abidlabs:
I'll go ahead and merge this in

PR #2067, opened 2022-08-23
Add data science demos to landing page

freddyaboulton:
# Description
Fixes #2048 

Added two demos:
- [Time series forecasting with Prophet](https://huggingface.co/spaces/gradio/timeseries-forecasting-with-prophet)
- [XGBoost Classification with Explainability](https://huggingface.co/spaces/gradio/xgboost-income-prediction-with-explainability)

Forecasting, Prophet, Tabular datasets, and Prediction Explainability are all concepts that have come up in our conversations with data scientists (both internal and external) and these demos cover those concepts. 

My hope is that putting them in the front page will catch the eyes of prospective data science users. The xgboost example is a sneak peek of what we want to be able to automate for skops models.

image

image

# Checklist:

- [x] I have performed a self-review of my own code
- [x] My code follows the style guidelines of this project
- [x] I have commented my code in hard-to-understand areas
- [x] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #2079, opened 2022-08-24
Hide time series + xgboost demos by default

freddyaboulton:
# Description

Fix a mistake I made in #2067 where I didn't hide the new demos by default

# Checklist:


aliabd:
The link for the source on time series forecasting is broken, missing the `/spaces/`.


Also I would recommend adding a code block like the original demos, and linking to the code there instead, something like below, but definitely not a big deal.
Screen Shot 2022-08-24 at 11 14 57 AM

aliabd (approved):

freddyaboulton:
Thanks for pointing out the broken link @aliabd ! Will fix in the space + make the other changes we discussed offline (replacing `source` with `code` and linking to the code itself)

PR #2082, opened 2022-08-25
Updated our analytics on creation of Blocks/Interface

abidlabs:
Three quick changes to our analytics on creation of Blocks/Interface:

* We actually fire the analytics when Blocks are created (previously, it was only when Interface was created)
* We record whether it was a Blocks or Interface that was created (using "mode")
* We also record which Gradio version they are running

I've also updated the `gradio-api-server` running to accept these analytics. We should keep in mind that we'll see an artificial bump in the "initiation" analytics as a result of this. For retrospective analysis, we should stick to launch analytics, which correctly record whenever Blocks are launched

### DIFF
--- a/demo/all_demos/run.py
+++ b/demo/all_demos/run.py
@@ -5,6 +5,7 @@ import sys
 import copy
 import pathlib
 
+os.environ["GRADIO_ANALYTICS_ENABLED"] = "False"
 
 demo_dir = pathlib.Path(__file__).parent / "demos"
 
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -4,6 +4,7 @@ import copy
 import getpass
 import inspect
 import os
+import pkgutil
 import random
 import sys
 import tempfile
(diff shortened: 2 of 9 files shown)


dawoodkhan82:
@abidlabs wont adding tests with analytics enabled, inflate the analytics?

dawoodkhan82:
@abidlabs the tests with analytics_enabled are mocked right?

freddyaboulton (approved):
Looks good to me @abidlabs ! I think we're mocking the requests so we should be good to go.

That being said, I don't think we're setting `analytics_enabled` to False in the `all_demos` demo that's being deployed to spaces so we should do that! I can do that too, up to you!

dawoodkhan82 (approved):

abidlabs:
Thanks for the review, guys! Will update `all_demos`, fix the tests, and then merge in.

krrishdholakia:
Hey! @abidlabs, if i enable analytics where do I go to see the analytics for my space (if that's even possible?)

Issue #2080, opened 2022-08-24
Can't load Image Classification model from hub and call it from another interface


freddyaboulton:
### Describe the bug

If you can an image classification model loaded from the hub from another interface, the interface will display a filename as a class name.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```python
import gradio as gr

tts_demo = gr.Interface.load(
    "huggingface/facebook/fastspeech2-en-ljspeech",
    title=None,
    description="Give me something to say!",
)

def call_text_to_speech_in_fn(text):
    return tts_demo(text)


tts_from_func = gr.Interface(call_text_to_speech_in_fn, "text", "audio")
tts_from_func.launch()
```

### Screenshot

![image](https://user-images.githubusercontent.com/41651716/186525116-d3eea5d8-cbc3-4826-956e-896403b36a49.png)

### Logs

```shell
-
```

### System Info

```shell
main
```

### Severity

blocking upgrade to latest gradio version

freddyaboulton:
@abidlabs Assigning you since you said you could fix during our huddle but happy to help as well!

PR #2083, opened 2022-08-25
`Label` component now accepts file paths to `.json` files 

abidlabs:
This PR adds support for filepaths to the `Label` component. Specifically, if you pass in a filepath to a ".json" file that exists, the Label component will read it and display its contents instead of just displaying the string filepath.

This fixes: #2080


aliabid94 (approved):
LGTM

abidlabs:
Thanks for the review!

Issue #2061, opened 2022-08-23
Pressing Enter doesn't start the execution if max_lines > 1


apolinario:
### Describe the bug

Not sure if it is a bug or not, but if `max_lines` > 1 pressing enter goes to a newline instead of triggering execution. While I can understand why that is great when the field is a "textbox", I also feel like few services have Shift+Enter for new line and Enter for submit. 

This is particularly prominent when `lines=1` but `max_lines > 1`, as in one hand we'd like to have multi-line in case the prompt is too long
![image](https://user-images.githubusercontent.com/788417/186125009-97ee3a91-e197-4a44-8024-aceb0332c43f.png)

But we still want Enter to submit instead of going to a new line
![image](https://user-images.githubusercontent.com/788417/186125197-2a74ff08-b152-4008-b2ab-69800b0e660a.png)

Not sure if needs changing. Maybe a solution would be changing the "enter for submit" only if `lines > 1` not if `max_lines > 1`. Not sure, a bit of brainstorming. 

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

No need

### Logs

```shell
No logs
```

### System Info

```shell
Gradio 3.1.7
```

### Severity

annoying

abidlabs:
I like the shift+enter

abidlabs:
Agree that this is confusing. What do you think about keeping enter for creating a new line, but accepting CMD/CTRL + Enter for submit? That's what GitHub uses (though I don't know how widely used this is)

apolinario:
I think the convention I observe is: 
- If it is a multi-line text-field already from the get-go, Enter creates a new line, CTRL+Enter submits (like GitHub) 
- If it is a single-line text-field that can be expanded to multi-line (Discord, Slack), Enter submits and CTRL+Enter creates a new line 

So in my example, where I have one line that can be expanded, imo the expectation is Enter to submit and CTRL+Enter for new line

Right now, Gradio works like this: 
if `lines = 1` and `max_lines = 1`, then enter submits
if `lines > 1` or `max_lines > 1`, then enter gives a new line and there is no way to submit 

What I would suggest is: 
if `lines = 1` and `max_lines > 1`, keep enter submitting and CTRL+Enter for new line (same Discord/Slack expected behavior)
if `lines > 1`, then it is visually a textarea and not an input field anymore, so then the logic of Enter for submit and CTRL+Enter for newline makes sense 

But I agree all of this is very confusing

[...]

PR #2084, opened 2022-08-25
Textbox Shift + Enter key Refactor

dawoodkhan82:
# Description
- if lines = 1 and max_lines > 1, keep that Enter submits and adds that Shift+Enter creates a new line (same Discord/Slack expected behavior the user only see one line, so they expect enter to submit imo)
- if lines > 1, then it is visually a textarea and not an input field anymore, so then the logic of Enter goes to the next line, and Shift+Enter submits makes sense to me
- if lines == 1 and max_lines == 1 then enter submits

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

Closes: #2061

# Checklist:


dawoodkhan82:
Some issues:
Max_lines doesnt seem to currently work
Ctrl key doesn't currently work with these changes, I can work on adding that if needed

abidlabs:
I think it's fine (actually better) to stick to just Shift+Enter because it's more clear when there's a single way to do it + shift is cross-platform whereas ctrl/cmd isn't

abidlabs:
Maybe we can tackle #2103 as part of this PR as well @dawoodkhan82?

dawoodkhan82:
> Maybe we can tackle #2103 as part of this PR as well @dawoodkhan82?

Will make a separate pr for this. Want to get this merged

dawoodkhan82:
@abidlabs can you confirm this works as expected, so I can merge

abidlabs:
Works perfectly, thanks for the clear explanation in the PR!

abidlabs (approved):

Issue #2077, opened 2022-08-24
Model3D preview shows grey square on spaces


freddyaboulton:
### Describe the bug

When uploading a Model3D file on spaces (from examples or manually) the preview modal doesn't show the Model3D object.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Go to the Model3D demo here:

https://huggingface.co/spaces/gradio-pr-deploys/pr-2064-all-demos

### Screenshot

![model3d_upload](https://user-images.githubusercontent.com/41651716/186450910-551f77b9-501d-466a-b903-c222cc0050a0.gif)

### Logs

```shell
Uncaught (in promise) TypeError: Cannot set properties of undefined (setting 'active')
    at e.disableAttributeByIndex (index.9f8deacc.js:30:345996)
    at e.unbindAllAttributes (index.9f8deacc.js:30:371147)
    at e.wipeCaches (index.9f8deacc.js:30:354673)
    at t.dispose (index.9f8deacc.js:30:274055)
    at addNewModel (index.9f8deacc.js:611:10)
    at Upload.handle_upload (index.9f8deacc.js:588:3)
```

```
Unable to load from file//tmp/tmpd7j3mvz9/Duckgmo4snro.glb: Error status: 500  - Unable to load file//tmp/tmpd7j3mvz9/Duckgmo4snro.glb
```
```


### System Info

```shell
Spaces
```

### Severity

blocking upgrade to latest gradio version

freddyaboulton:
The same thing happens in a more minimal space https://huggingface.co/spaces/freddyaboulton/Model3D so I think it's an issue with changing the logic for how we load files as opposed to the `all_demos` space

PR #2086, opened 2022-08-25
Fix issues related to demos in Spaces

abidlabs:
This PR:

* Modifies the `all_demos` demo so that it will still run if some of the demos in it error out. This is important as some demos are loaded from HF API and can result in 429 timeouts (or have other errors), which would previously prevent the whole demo from running. Now, the overall demo still loads, but individual tabs have a clear ❌ to show that the particular demo did not load. Clicking on the tab shows the error message, e.g.

image

* Fixes: #2077. Basically the problem is that Spaces seems to automatically convert double slashes ("//") into single slashes ("/"), which is a problem when trying to supply paths that begin with "/" via URLs. So I've changed the "/file/" route to be "/file="

`all_demos` is now working on Spaces, confirm here: https://huggingface.co/spaces/gradio-pr-deploys/pr-2086-all-demos


demo/audio_debugger/run.py
```diff
@@ -0,0 +1,18 @@
+import gradio as gr
+import subprocess
+import os
+
+audio_file = os.path.join(os.path.dirname(__file__), "cantina.wav")
```
aliabid94, line 5: Why do we have these `os.path.join(os.path.dirname())` in all our demos?  You should be able to just put the filename, and users who look at demos for examples will think that they have to do this.

aliabid94 (approved):
LGTM. Tested a few demos, and worked fine

demo/audio_debugger/run.py
abidlabs: Yeah that's a good point. But you need this to be able to run the demo files from any location

PR #2094, opened 2022-08-26
Show error on Connection drops

aliabid94:
A Websocket connection can drop for several reasons, e.g. a computer going to sleep, or on phone devices, if an android switches tab, or an iphone goes to sleep. Previously we did not display an error message for this. letting the broken websocket show a "processing" animation indefinitely, making users wait forever. Now there is an error on a broken connection,

Works for both Queue websockets and regular rest api (in case internet connection breaks). PR also includes general queue code cleanup.


demo/fake_gan_2/run.py
```diff
@@ -10,14 +10,15 @@
 def fake_gan(desc):
     if desc == "NSFW":
         raise gr.Error("NSFW - banned content.")
-    time.sleep(5)
+    if desc == "":
+        raise ValueError("Too short")
+    time.sleep(12)
```
abidlabs, line 15: I would remove this. Good for testing, not great for a demo
```suggestion
    time.sleep(12)
```

demo/fake_gan_2/run.py
```diff
@@ -10,14 +10,15 @@
...
     image = random.choice(
         [
-            "https://images.unsplash.com/photo-1507003211169-0a1dd7228f2d?ixlib=rb-1.2.1&ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&auto=format&fit=crop&w=387&q=80",
-            "https://images.unsplash.com/photo-1554151228-14d9def656e4?ixlib=rb-1.2.1&ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&auto=format&fit=crop&w=386&q=80",
-            "https://images.unsplash.com/photo-1542909168-82c3e7fdca5c?ixlib=rb-1.2.1&ixid=MnwxMjA3fDB8MHxzZWFyY2h8MXx8aHVtYW4lMjBmYWNlfGVufDB8fDB8fA%3D%3D&w=1000&q=80",
-            "https://images.unsplash.com/photo-1546456073-92b9f0a8d413?ixlib=rb-1.2.1&ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&auto=format&fit=crop&w=387&q=80",
-            "https://images.unsplash.com/photo-1601412436009-d964bd02edbc?ixlib=rb-1.2.1&ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&auto=format&fit=crop&w=464&q=80",
+            "files/cheetah1.jpg",
```
abidlabs, line 18: This is fine, but this syntax makes the files only accessible from the demo directory. Would be better to adopt a more flexible format like in other demos

gradio/event_queue.py
```diff
@@ -120,19 +122,13 @@ def push(cls, event: Event) -> int | None:
...
-        async with cls.DELETE_LOCK:
-            # TODO: Might log number of cleaned events in the future
-            try:  # TODO: Might search with event hash to speed up
+        if event in cls.EVENT_QUEUE:
+            async with cls.DELETE_LOCK:
                 cls.EVENT_QUEUE.remove(event)
-            except ValueError as er:
-                print(f"{er}, {event}")
+        elif event in cls.ACTIVE_JOBS:
+            cls.ACTIVE_JOBS[cls.ACTIVE_JOBS.index(event)] = None
```
abidlabs, line 131: Should we still have an `try`/`else` clause for other errors that might happen if removing the event? Or was the only source of errors?

abidlabs:
One issue I noticed with the queue is that if you run a demo with `queue()` enabled from the terminal, then you cannot kill the python process with the standard keyboard interrupt (ctrl+c). Don't think that should block this PR from merging though

abidlabs:
Very cool @aliabid94! To be completely honest, there were parts that I couldn't fully reason through (some more documentation would help as I commented), but the overall logic makes sense to me.

abidlabs:
Now I tried testing it locally. I opened up two tabs and ran the `fake_gan_2` demo, ran both of them, and then pressed the power button on my computer to put it to sleep. When I woke my computer a minute later, I found that both of the demos said "error" instead of showing a picture. Is this expected behavior? 

Tab 1:

image

Tab 2:

image

aliabid94:
Yes that's the expected behaviour. The websocket connection broke when you put your computer to sleep.

abidlabs:
Following up from our conversation, can you remind me how I can go about testing the new connection error introduced in this PR?

gradio/event_queue.py
aliabid94: I don't see why there should be a try/catch on something like this

abidlabs:
Ok that’s fine

gradio/event_queue.py
aliabid94: Don't see how any other error could happen

abidlabs:
LGTM @aliabid94, thanks for the fix!

abidlabs (approved):

Issue #1162, opened 2022-05-05
Be able to adjust column widths


aliabid94:
We use gr.Row() and gr.Column() currently to configure the layout using flexbox. Currently this implementation is unintuitive for some users, so opening this issue for discussion. 

Take a look at this image below:
![Screen Shot 2022-05-04 at 5 43 52 PM](https://user-images.githubusercontent.com/7870876/166849035-a077a3b0-e26c-4687-b236-49ab3a422d7a.png)

The current code to render this would be: 

```python
import gradio as gr

with gr.Blocks() as demo:
    gr.Markdown("# Welcome")
    gr.CheckboxGroup(["Option 1", "Option 2", "Option 3"])
    with gr.Row():
        gr.Image("cheetah1.jpg")
        gr.Image("lion.jpg")
    with gr.Row():
        btn1 = gr.Button("Load")
        btn2 = gr.Button("Save")
        with gr.Column():
            btn3 = gr.Button("Delete")
            btn4 = gr.Button("Close")
            
demo.launch()
```

where gr.Row() will lay out every child component in a flex row, and gr.Column() will lay out all child components in a flex column.

aliabid94:
@abidlabs had suggested an alternative API, where everything is laid out vertically, unless placed in adjacent columns. His suggested API for the same demo would be:

```python
import gradio as gr

with gr.Blocks() as demo:
    gr.Markdown("# Welcome")
    gr.CheckboxGroup(["Option 1", "Option 2", "Option 3"])
    with gr.Row():
        with gr.Column():
            gr.Image("cheetah1.jpg")
        with gr.Column():
            gr.Image("lion.jpg")
    with gr.Row():
        with gr.Column():
            btn1 = gr.Button("Load")
        with gr.Column():
            btn2 = gr.Button("Save")
        with gr.Column():
            btn3 = gr.Button("Delete")
            btn4 = gr.Button("Close")
            
demo.launch()
```

abidlabs:
Slightly tweaking the API above, we can actually get rid of `gr.Row()` altogether:

```py
import gradio as gr

with gr.Blocks() as demo:
    gr.Markdown("# Welcome")
    gr.CheckboxGroup(["Option 1", "Option 2", "Option 3"])
    col1, col2 = gr.Columns(2)
    with col1:
        gr.Image("cheetah1.jpg")
    with col2:
        gr.Image("lion.jpg")
      
    col1, col2, col3 = gr.Columns(3)
    with col1:
[...]

PR #2097, opened 2022-08-27
Refactoring Layout: Adding column widths, forms, and more.

aliabid94:
Lots of changes related to layout, sorry for the large PR. Fixes: #1162 

# Column width manipulation

Can now set width of columns using `scale` (relative widths) and `min_width`

# Replacing mobile collapse for flex based wrapping

Collapsing previously had issues because collapsing was defined by screen width, not element width. For example, a nested row would not collapse even if it had limited space, simply because the screen was large. Neither would a row in an embedded space. Now this works  

# Simplifying "form" visual generation

Combing adjacent form elements into a unified form visual was complex and a bit hacky. Previously there was a lot of logic to calculate which components were "first", "middle" or "last" and render their borders appropriately. This had a lot of css and js code, and had bugs (e.g.  if a "first" component was hidden, the second component had a missing border)

Simplified significantly. Now in the backend, adjacent form components are automatically combined into a gr.Form, and in the frontend we simply render a Form layout, which inherits its flex-direction.

# Automatic parent generation (for Forms and Tab)

To do the previous change (combining adjacent form components to be within a Form layout), I had to implement automatic parent generation. That means 
```python
with gr.Row():
  t1 = gr.Textbox()
  n1 = gr.Number()
```
automatically was converted to:
```python
with gr.Row():
  with gr.Form():
    t1 = gr.Textbox()
    n1 = gr.Number()
```

it was a simple change to extend this to tabs, so now:
```python
with gr.Tab("item1"):
  t1 = gr.Textbox()
with gr.Tab("item2"):
  t2 = gr.Textbox()
```
automatically adds the parent `Tabs` if missing, and is converted to:
```python
with gr.Tabs():
  with gr.Tab("item1"):
    t1 = gr.Textbox()
  with gr.Tab("item2"):
    t2 = gr.Textbox()
```

saving a lot of nesting. Note I also renamed `TabItem` to `Tab`, and have updated documentation. 

I have NOT implemented this for columns yet, but we could also apply this to columns, so
```python
with gr.Column():
  t1 = gr.Textbox()
with gr.Column():
  t2 = gr.Textbox()
```
automatically generates
```python
with gr.Row():
  with gr.Column():
    t1 = gr.Textbox()
  with gr.Column():
    t2 = gr.Textbox()
```

## Other

- Added "component-" prefix to autogenerated element ids since previous ids were only numbers, which is an invalid attribute and could not be used with CSS.


abidlabs:
Thanks for fixing this issue @aliabid94! What about a slight modification of this API so that for width, you can accept a _ratio_ or a _factor_, similar to how flex works. So for example, if two elements are in a row, and the first has width=3, and the second has width=1, the first one will take 75% of the available width, while the second will take 25% of the available width (by default, elements could have a width factor of 1)? Tbh I'm not sure if this is a better approach, but I wanted to bring it up as the approach proposed in this PR could lead to weird conflicts (for example, if the percentages add up to more than 100%, etc.)

abidlabs:
Wow this is good stuff @aliabid94!

pngwn (dismissed):
Looks good other than one point:

> Note I also renamed TabItem to Tab, and have updated documentation.

This is a pretty major breaking change and will need to be reverted unless we want to release  4.0. Alternatively, we can deprecate `TabItem` and add `Tab` in addition to `TabItem` rather than replacing it.

aliabid94:
Sorry should have been more clear, Tab has been added in addition to TabItem, which has been removed form the documentation but preserved for backwards compatibility.

gradio/blocks.py
```diff
@@ -425,6 +425,7 @@ def get_block_instance(id: int) -> Block:
             block_config["props"].pop("type", None)
             block_config["props"].pop("name", None)
             style = block_config["props"].pop("style", None)
+            print("!!!", block_config)
```
abidlabs, line 428: Remove
```suggestion
            print("!!!", block_config)
```

gradio/blocks.py
```diff
@@ -693,6 +694,35 @@ def get_config_file(self):
...
+                if not expected_parent or isinstance(block, expected_parent):
+                    pseudo_parent = None
+                    children.append(child)
+                else:
+                    if pseudo_parent is not None and isinstance(
+                        pseudo_parent, expected_parent
+                    ):
+                        pseudo_parent.children.append(child)
+                    else:
+                        pseudo_parent = expected_parent(render=False)
```
abidlabs, line 714: So `render()` is useful after all ;)

gradio/layouts.py
```diff
@@ -55,49 +47,65 @@ def update(
...
-
+            warnings.warn("mobile_collapse is no longer supported.", DeprecationWarning)
         return self
 
 
 @document()
 class Column(BlockContext):
     """
-    Column is a layout element within Blocks that renders all children vertically.
+    Column is a layout element within Blocks that renders all children vertically. The widths of columns can be set through its arguments.
```
abidlabs, line 68:
```suggestion
    Column is a layout element within Blocks that renders all children vertically. The widths of columns can be set through the `scale` and `min_width` parameters.
```

abidlabs:
LGTM! Some nits above

abidlabs (approved):

demo/blocks_plug/run.py
```diff
@@ -22,13 +22,12 @@ def change_tab():
...
 with gr.Blocks() as demo:
     gr.Markdown("Three demos in one!")
-    with gr.Tabs(selected=1) as tabs:
-        with gr.TabItem("Text Identity", id=0):
-            identity_demo.render()
-        with gr.TabItem("Text Input", id=1):
-            input_demo.render()
-        with gr.TabItem("Text Static", id=2):
-            output_demo.render()
+    with gr.Tab("Text Identity", id=0):
```
freddyaboulton, line 25: This breaks the demo since the `tabs` variable is no longer defined!

demo/blocks_plug/run.py
```diff
@@ -22,13 +22,12 @@ def change_tab():
 
 with gr.Blocks() as demo:
```
freddyaboulton, line 23: I'm not sure if the changes to Tabs are truly backwards compatible.

Changing the demo back to what it was before breaks (shows red `Error`)

image

```
index.9e34751b.js:6412 TypeError: Cannot read properties of undefined (reading 'type')
    at walk_layout (index.9e34751b.js:6385:57)
    at index.9e34751b.js:6394:45
    at Array.map (<anonymous>)
    at walk_layout (index.9e34751b.js:6394:36)
    at async Promise.all (:7860/index 1)
    at async walk_layout (index.9e34751b.js:6394:4)
    at async Promise.all (:7860/index 0)
    at async walk_layout (index.9e34751b.js:6394:4)
    at async Promise.all (:7860/index 0)
    at async walk_layout (index.9e34751b.js:6394:4)
```

For reference, I just made the following modification to this code block

```python
with gr.Blocks() as demo:
    gr.Markdown("Three demos in one!")
    with gr.Tabs(selected=1) as tabs:
        with gr.Tab("Text Identity", id=0):
            identity_demo.render()
        with gr.Tab("Text Input", id=1):
            input_demo.render()
        with gr.Tab("Text Static", id=2):
            output_demo.render()
    btn = gr.Button("Change tab")
    btn.click(inputs=None, outputs=tabs, fn=change_tab)
```

demo/blocks_plug/run.py
aliabid94: Good catch!

demo/blocks_plug/run.py
aliabid94: Took me forever to figure out what was going on here! Turns out Interfaces embedded in Blocks broke, finally fixed in by moving the expected_parent logic int the BlockContext.

Issue #2074, opened 2022-08-24
Where is the CSS style file location while lunch the demo, and how could i locate the image resource by CSS url ?


HuangChiEn:
First, thanks for you'll release this amazing repo for fast building ML GUI.
I notice that customize CSS allow us insert arbitrary image as background. However, `{background-image: url(...) }` requires the user to indicate the source path. However, I have tried several combinations to locate the source path, which does did not work.
```
# suppose i have an image under /workspace/data/ex.jpg, the demo.py lunch the demo in /workspace folder
# what i have tried : 
css_def = " #exm_id { background-image: url('data/ex.jpg') }"  
#css_def = " #exm_id { background-image: url('/workspace/data/ex.jpg') }"  
#css_def = " #exm_id { background-image: url('/data/ex.jpg') }" 

demo = block(css=css_def)
with demo:
    ...

if __name__ == "__main__":
    # a bit complex setup..
    demo.lunch(server='0.0.0.0', port='4444')  # in docker
    # docker run -it -p 4444:5555 dkr_img bash
    #     and i have open the browser to view this demo website "xx.xx.xx.xx:5555" 
```

I wonder where is the CSS style file location, since most of path apply the relative path of CSS style file to locate the source image.

Any suggestion about CSS is also appreciated !!

abidlabs:
Hi @HuangChiEn can you try putting the prefix “file/“ before the relative or absolute path to your image URL?

abidlabs:
Here is a working example:

```py
import gradio as gr

with gr.Blocks(css="#t{background-image: url('file/lion.jpg') }") as demo:
    gr.Textbox(elem_id="t")

demo.launch()
```

Which produces this:

<img width="721" alt="image" src="https://user-images.githubusercontent.com/1778297/187062198-551d8d6d-9f78-4038-9a19-619d3f69d4f6.png">

I'll keep this issue open until we document this better

HuangChiEn:
> Here is a working example:
> 
> ```python
> import gradio as gr
> 
> with gr.Blocks(css="#t{background-image: url('file/lion.jpg') }") as demo:
>     gr.Textbox(elem_id="t")
> 
> demo.launch()
> ```
> 
> Which produces this:
> 
> <img alt="image" width="721" src="https://user-images.githubusercontent.com/1778297/187062198-551d8d6d-9f78-4038-9a19-619d3f69d4f6.png">
> 
> I'll keep this issue open until we document this better

@abidlabs 
[...]

PR #2099, opened 2022-08-28
Accept deprecated `file` route as well

abidlabs:
This PR does two things:

* As part of #2086, I changed the route corresponding to serving files from `/file/` to `/file=`. However, I realized that this might break demos if people are providing file paths themselves, for example as part of custom css, as in #2074. So this PR restores the deprecated file path as well. 

* Added documentation about providing files as part of CSS to address questions that have popped up about it (including #2074). So this fixes: #2074

### DIFF
--- a/gradio/routes.py
+++ b/gradio/routes.py
@@ -236,6 +236,10 @@ class App(FastAPI):
                     f"it is not in any of {app.blocks.temp_dirs}"
                 )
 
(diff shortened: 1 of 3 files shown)


abidlabs:
I've gotten a verbal review from @aliabid94 so I'm going to merge this in so that I can make some tweaks to #2073 and avoid conflicts. Feel free to provide feedback even after the PR is closed!

Issue #730, opened 2022-02-24
[Blocks] `State` throws a js error 


aliabd:
### Describe the bug

Using `State` with `Chatbot` throws an error.

### Reproduction

```python 
import gradio as gr


def chat(message, history):
    history = history or []
    history.append((message, message))
    return history, history


block = gr.Blocks()


with block:
    chatbot = gr.outputs.Chatbot()
    message = gr.inputs.Textbox()
    submit = gr.Button("SEND")
    submit.click(chat, inputs=[message, gr.inputs.State()], outputs=[chatbot, gr.outputs.State()])


block.launch()
```

Throws this console error:

```
index.800943b3.js:14 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'component')
    at Array.uu (index.800943b3.js:14:7188)
    at kn (index.800943b3.js:14:9671)
    at du (index.800943b3.js:14:10040)
    at Qs (vendor.ba9f2ee9.js:1:6986)
    at new nt (index.800943b3.js:14:11214)
    at hu (index.800943b3.js:14:11379)
    at Qs (vendor.ba9f2ee9.js:1:6986)
    at new bu (index.800943b3.js:14:12689)
    at window.launchGradio (index.800943b3.js:14:15525)
    at index.800943b3.js:14:15863
uu @ index.800943b3.js:14
kn @ index.800943b3.js:14
du @ index.800943b3.js:14
Qs @ vendor.ba9f2ee9.js:1
nt @ index.800943b3.js:14
hu @ index.800943b3.js:14
Qs @ vendor.ba9f2ee9.js:1
bu @ index.800943b3.js:14
window.launchGradio @ index.800943b3.js:14
(anonymous) @ index.800943b3.js:14
Promise.then (async)
(anonymous) @ index.800943b3.js:14
```
Nothing is rendered on the page. 

Replacing the block launch with 
```python
gr.Interface(chat, [message, gr.inputs.State()], [chatbot, gr.outputs.State()]).launch()
```
works as expected

### System Info

```shell
gradio v2.8.0b20
```

### Severity

blocker

abidlabs:
Different error, but still an issue. 

Try running `demo/chatbot/run.py` to reproduce

abidlabs:
Closed via #977

satpalsr:
[...]

PR #2100, opened 2022-08-28
Improvements to `State`

abidlabs:
This PR makes some enhancements to session state. First, it introduces `gr.State()` as an alias of `gr.Variable()`, unifying the naming we use in `Blocks` and `Interface`. To use state in `Interface`, one can do either of these approaches:

* gr.Interface(..., "state", "state")
* gr.Interface(..., `gr.State()`, `gr.State()`) (previously, this would **not** work)

In Blocks, the same approach works as before but using `gr.State()` instead of `gr.Variable()`.

Some additional tests have been added. The demos and documentation have also been updated to use `gr.State()`. For backwards compatibility, `gr.Variable()` is still accepted.

Fixes: #730

### DIFF
--- a/demo/hangman/run.py
+++ b/demo/hangman/run.py
@@ -4,7 +4,7 @@ import random
 secret_word = "gradio"
 
 with gr.Blocks() as demo:    
-    used_letters_var = gr.Variable([])
+    used_letters_var = gr.State([])
     with gr.Row() as row:
         with gr.Column():
             input_letter = gr.Textbox(label="Enter letter")
--- a/demo/kitchen_sink_random/run.py
+++ b/demo/kitchen_sink_random/run.py
@@ -17,7 +17,7 @@ from constants import (
 
 
 demo = gr.Interface(
-    lambda x: x,
+    lambda *args: args[0],
     inputs=[
         gr.Textbox(value=lambda: datetime.now(), label="Current Time"),
         gr.Number(value=lambda: random.random(), label="Ranom Percentage"),
@@ -91,7 +91,9 @@ demo = gr.Interface(
         gr.Plot(value=random_plot),
         gr.Markdown(value=lambda: f"### {random.choice(['Hello', 'Hi', 'Goodbye!'])}"),
     ],
-    outputs=None,
+    outputs=[
+        gr.State(value=lambda: random.choice(string.ascii_lowercase))
+    ],
 )
 
 if __name__ == "__main__":
--- a/demo/kitchen_sink_random/run.py
+++ b/demo/kitchen_sink_random/run.py
@@ -60,7 +60,7 @@ demo = gr.Interface(
             )
         ),
         gr.Timeseries(value=lambda: os.path.join(file_dir, "time.csv")),
-        gr.Variable(value=lambda: random.choice(string.ascii_lowercase)),
+        gr.State(value=lambda: random.choice(string.ascii_lowercase)),
         gr.Button(value=lambda: random.choice(["Run", "Go", "predict"])),
         gr.ColorPicker(value=lambda: random.choice(["#000000", "#ff0000", "#0000FF"])),
         gr.Label(value=lambda: random.choice(["Pedestrian", "Car", "Cyclist"])),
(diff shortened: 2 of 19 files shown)


gradio/components.py
```diff
@@ -2570,6 +2570,13 @@ def style(self):
         return self
 
 
+class Variable(State):
```
freddyaboulton, line 2573: I think we need to fix the config for Variable to display `state` as the type. As is, it shows Variable and it causes demos that use `Variable` to break.

demo/hangman/run.py
```diff
@@ -4,7 +4,7 @@
 secret_word = "gradio"
 
 with gr.Blocks() as demo:    
-    used_letters_var = gr.Variable([])
+    used_letters_var = gr.State([])
```
freddyaboulton, line 7: Can we keep one of these demos, using `gr.Variable` so that we can make sure both can be properly used?

gradio/interface.py
```diff
@@ -213,17 +213,30 @@ def __init__(
...
+            pass
+        elif len(state_input_indexes) != 1 or len(state_output_indexes) != 1:
+            raise ValueError(
+                "If using 'state', there must be exactly one state input and one state output."
+            )
+        else:
+            state_input_index = state_input_indexes[0]
+            state_output_index = state_output_indexes[0]
+            if inputs[state_input_index] == "state":
+                default = utils.get_default_args(fn)[state_input_index]
```
freddyaboulton, line 233: Unrelated, but the type hint for `get_default_args` is wrong (Dict instead of List)

guides/3)building_with_blocks/3)state_in_blocks.md
```diff
@@ -19,11 +19,11 @@ $demo_hangman
...
 Let's see how we do each of the 3 steps listed above in this game:
 
-1. We store the used letters in `used_letters_var`. In the constructor of `Variable`, we set the initial value of this to `[]`, an empty list. 
+1. We store the used letters in `used_letters_var`. In the constructor of `State`, we set the initial value of this to `[]`, an empty list. 
 2. In `btn.click()`, we have a reference to `used_letters_var` in both the inputs and outputs.
-3. In `guess_letter`, we pass the value of this `Variable` to `used_letters`, and then return an updated value of this `Variable` in the return statement.
+3. In `guess_letter`, we pass the value of this `State` to `used_letters`, and then return an updated value of this `State` in the return statement.
 
-With more complex apps, you will likely have many Variables storing session state in a single Blocks app.
+With more complex apps, you will likely have many State variables storing session state in a single Blocks app.
```
freddyaboulton, line 26: In the `interface_state.md` guide should we mention that interfaces only support a single state variable? And point them to blocks for more complex state?

freddyaboulton (commented):
@abidlabs Looks good to me except the change in the config for gr.Variable I note in one of the comments

demo/kitchen_sink_random/run.py
```diff
@@ -60,7 +60,7 @@
             )
         ),
         gr.Timeseries(value=lambda: os.path.join(file_dir, "time.csv")),
-        gr.Variable(value=lambda: random.choice(string.ascii_lowercase)),
+        gr.State(value=lambda: random.choice(string.ascii_lowercase)),
```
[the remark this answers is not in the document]
abidlabs: Ah thank you! Will fix

demo/hangman/run.py
abidlabs: Hmm good point. Rather than using demos for testing (and showing things that we'll be deprecating), how about making it a test instead?

gradio/interface.py
abidlabs: Fixed

guides/3)building_with_blocks/3)state_in_blocks.md
abidlabs: Sure will make the addition

demo/hangman/run.py
freddyaboulton: Makes sense to me!

abidlabs: Done

gradio/components.py
abidlabs: Good catch!

abidlabs:
Thanks for the catches @freddyaboulton! Have updated the PR and will merge after the tests pass

PR #2112, opened 2022-08-29
Partial fix to HF Spaces auth issue

aliabid94:
Currently, auth doesn't work on Spaces. This is for two reasons:
1. Queueing is automatically enabled in Spaces, and currently queueing isn't supported with auth
2. The HF Infra seems to block POST requests to /login, causing a timeout

This fixes problem 1. If auth is enabled, then queueing isn't automatically enabled for Spaces. Need HF infra to fix 2.


abidlabs:
Is there a blocker on merging this in @aliabid94?

abidlabs:
Closing, thanks to #3049

Issue #2117, opened 2022-08-30
Ability to pass elem_id to Examples


ankrgyl:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  

I'm trying to write custom CSS for an examples gallery, but unlike other components, it does not accept an `elem_id` argument.

**Describe the solution you'd like**  
Ideally, I could pass in `elem_id` like I can with other components.

**Additional context**  
I'm able to work around this by just using the `gr-samples-table` class, so it's not urgent or blocking by any means.

abidlabs:
Should be an easy fix, we'll get it in soon, thanks for the heads up @ankrgyl!

ankrgyl:
Happy to post this as a separate issue, but kind of related, it'd be useful for our use case to be able to "scope" multiple components' CSS. E.g. pass an `elem_id` into `gr.Row()` and `gr.Column()`. Apologies as always if I'm missing something obvious!

abidlabs:
It will be possible to apply CSS to `gr.Row()` and `gr.Column()` soon via #2097. Thanks!

abidlabs:
Closing as #2097 has been merged in!

abidlabs:
(and should be out on pypi in a day or two)

PR #2119, opened 2022-08-30
Added `elem_id` to `Examples`

abidlabs:
Quickly fixes: #2117 so that we can get it in 3.2


gradio/examples.py
```diff
@@ -186,6 +189,7 @@ def __init__(
             samples=non_none_examples,
             type="index",
             label=label,
+            elem_id=elem_id,
```
freddyaboulton, line 192: This change looks good to me but we need to rebase prior to merging in! Thanks @abidlabs !

abidlabs:
Actually let me just make a new PR with some other things too

Issue #2015, opened 2022-08-15
Create Gradio Demo from skops pipeline


freddyaboulton:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
(skops)[https://huggingface.co/blog/skops] makes it easy to host a scikit learn model on the model hub and tag it with a model card that includes important metadata, like sklearn version, input data types, and example inputs. 

It would be really helpful to users to automatically create gradio demos from scikit learn models on the hub.

**Describe the solution you'd like**

The existing api for loading transformers pipelines on the hub should work for skops pipelines.

```python
gr.Interface.load("models/<user-name>/<sklearn-pipeline-name>").launch(...)
```

**Additional context**  
Out of scope - "rich" ui that depends on the input types, e.g. drop-down for categorical. This information is not tracked by skops in the metadata yet.

merveenoyan:
x-posting


I would like a huggingface transformers-like integration for gradio where you can call the model from the Hub and build an interface in one line of code, like below:

interface = gr.Interface.load("huggingface/scikit-learn/tabular-playground")

What I have in mind is dataframe component in, dataframe component out UI.

Wanted to open discussion for requirements to have this and what to do (might be cool if you could somehow put links on to-do's for previous transformers integration). @freddyaboulton @abidlabs (might be interested)

The UI looks like this: https://huggingface.co/spaces/scikit-learn/tabular-playground (dataframe in, dataframe out)
However, I couldn't manage to pass examples like we used to do as list of lists, where in each inner list is a row (an example) I opened an issue (https://github.com/gradio-app/gradio/issues/2089 ).

We can automatically receive headers and example inputs from config.json files of skops repositories and construct dataframe component by passing those.

Also pinging @adrinjalali and @BenjaminBossan here.

adrinjalali:
I think there are at least two usecases here. One is that we want predictions for a single input, in which case @freddyaboulton's suggestion makes sense, and another one is to have a csv or a data file for which we'd like the output. To me, most tabular usecases don't want to have the output for a single input. They need the output for a batch, and if they need the output for a single input, it's not the user wanting to get that, it's a program sending API calls.

I'm not sure how these components work, but I think having a table with inputs and having the output in the same or a different table would make sense.

merveenoyan:
[...]

PR #2126, opened 2022-08-30
Skops integration: Load tabular classification and regression models from the hub

freddyaboulton:
# Description

Ability to load tabular classification and regression models from the hub and turn it into a demo.

Closes: #2015

### Tabular Regression
```python
import gradio as gr

gr.Interface.load("models/skops-ci/test-3255bd22-bb75-4641-8655-824cd25d140f").launch()
```
Users can edit the dataframe to get new predictions

tabular_regression_demo

### Tabular classification
```python
import gradio as gr

gr.Interface.load("models/scikit-learn/tabular-playground").launch()
```

image

### Demo with missing data in the input widget
```python
import gradio as gr

gr.Interface.load("models/demo-org/tabular-playground").launch()
```

image

### What happens when api request fails
image

# Checklist:

### DIFF
--- a/gradio/external.py
+++ b/gradio/external.py
@@ -59,7 +59,13 @@ def get_tabular_examples(model_name) -> Dict[str, List[float]]:
             "(?:^|[\r\n])---[\n\r]+([\\S\\s]*?)[\n\r]+---([\n\r]|$)", readme.text
         )
         example_yaml = next(yaml.safe_load_all(readme.text[: yaml_regex.span()[-1]]))
-        example_data = example_yaml["widget"]["structuredData"]
+        example_data = example_yaml.get("widget", {}).get("structuredData", {})
+    if not example_data:
+        raise ValueError(
+            f"No example data found in README.md of {model_name} - Cannot build gradio demo. "
+            "See the README.md here: https://huggingface.co/scikit-learn/tabular-playground/blob/main/README.md "
+            "for a reference on how to provide example data to your model."
+        )
     return example_data
 
 
@@ -67,11 +73,16 @@ def cols_to_rows(
     example_data: Dict[str, List[float]]
 ) -> Tuple[List[str], List[List[float]]]:
     headers = list(example_data.keys())
+    n_rows = max(len(example_data[header] or []) for header in headers)
     data = []
-    for row_index in range(len(example_data[headers[0]])):
+    for row_index in range(n_rows):
         row_data = []
         for header in headers:
-            row_data.append(example_data[header][row_index])
+            col = example_data[header] or []
+            if row_index >= len(col):
+                row_data.append(None)
+            else:
+                row_data.append(col[row_index])
         data.append(row_data)
     return headers, data
 
@@ -306,6 +317,7 @@ def get_models_interface(model_name, api_key, alias, **kwargs):
     if p in ["tabular-classification", "tabular-regression"]:
         example_data = get_tabular_examples(model_name)
         col_names, example_data = cols_to_rows(example_data)
+        example_data = [[example_data]] if example_data else None
 
         pipelines[p] = {
             "inputs": components.Dataframe(
@@ -322,7 +334,7 @@ def get_models_interface(model_name, api_key, alias, **kwargs):
                 "headers": ["prediction"],
                 "data": [[pred] for pred in json.loads(r.text)],
             },
-            "examples": [[example_data]],
+            "examples": example_data,
         }
 
     if p is None or not (p in pipelines):
--- a/gradio/external.py
+++ b/gradio/external.py
@@ -52,7 +52,6 @@ def load_blocks_from_repo(name, src=None, api_key=None, alias=None, **kwargs):
 
 
 def get_tabular_examples(model_name) -> Dict[str, List[float]]:
-    breakpoint()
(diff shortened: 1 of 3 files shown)


BenjaminBossan:
For skops models, instead of reading the sample data from the README, it would also be possible to read the `config.json`, which also contains the sample data (or is there any reason not to @adrinjalali?). That way, you can get rid of the regex and pyyaml dependency.

adrinjalali:
as long as `skops` is concerned, we only deal with the `config.json` file, so we don't have pyyaml and regex work in it. Those are only in the README file since the widget expects them.

freddyaboulton:
Good point about the config.json vs README @BenjaminBossan @adrinjalali ! Since reading the input data from the README will work for models that were uploaded with and without skops, e.g. (https://huggingface.co/julien-c/wine-quality), I think I will stick with that.

gradio/external.py
```diff
@@ -42,6 +49,53 @@ def load_blocks_from_repo(name, src=None, api_key=None, alias=None, **kwargs):
     return blocks
 
 
+def get_tabular_examples(model_name) -> Dict[str, List[float]]:
+    readme = requests.get(f"https://huggingface.co/{model_name}/resolve/main/README.md")
```
freddyaboulton, line 53: We can either get the example data from the README or the `config.json` but the `config.json` will only have the example data if the model was uploaded with skops. 

I think it would be better if gradio could create a demo for any tabular model and not just those created with skops. Downside is that it introduces a pyyaml dependency.

In the future, once the skops config json file contains richer metadata about feature types (categorical vs null) etc we can read from the config.json if it's present.

test/test_external.py
```diff
@@ -242,5 +243,64 @@ def test_interface_load_cache_examples(tmp_path):
...
+    if not any(readme_examples.values()):
+        assert not any([c for c in config["components"] if c["type"] == "dataset"])
+    else:
+        dataset = next(c for c in config["components"] if c["type"] == "dataset")
+        assert dataset["props"]["samples"] == [
+            [utils.delete_none(cols_to_rows(readme_examples)[1])]
+        ]
+
+
+@pytest.mark.parametrize(
```
freddyaboulton, line 286: Should cover most of the weirdness from malformed READMEs. Haven't actually come across a repo with bad README data but it's possible.

freddyaboulton:
@BenjaminBossan @adrinjalali @merveenoyan Can't officially tag you as a reviewer but feel free to give this a look when you get a chance!

gradio/external.py
```diff
@@ -42,6 +49,53 @@ def load_blocks_from_repo(name, src=None, api_key=None, alias=None, **kwargs):
...
+        warnings.warn(f"Cannot load examples from README for {model_name}", UserWarning)
+        example_data = {}
+    else:
+        yaml_regex = re.search(
+            "(?:^|[\r\n])---[\n\r]+([\\S\\s]*?)[\n\r]+---([\n\r]|$)", readme.text
+        )
+        example_yaml = next(yaml.safe_load_all(readme.text[: yaml_regex.span()[-1]]))
+        example_data = example_yaml.get("widget", {}).get("structuredData", {})
+    if not example_data:
+        raise ValueError(
```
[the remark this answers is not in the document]
merveenoyan: Regardless of the feature names being provided, each feature has it's own value range or feature type anyway, so it doesn't make sense even if you provide everything. What would make sense would be people calling it and loading the interface with dynamic dataframe and still provide an example themselves in the interface.

test/test_external.py
```diff
@@ -242,5 +243,64 @@ def test_interface_load_cache_examples(tmp_path):
         )
 
 
+def test_cols_to_rows():
+    assert cols_to_rows({"a": [1, 2, "NaN"], "b": [1, "NaN", 3]}) == (
```
merveenoyan, line 247: Just a small question, if there's a cell left empty, how is it handled? Do you impute "NaN" directly? (how is it sent to inference?)

merveenoyan (commented):
Not much of a feedback but to help you out with perspective 🙂

test/test_external.py
freddyaboulton: Right now it would be left empty but that doesn't work for the inference API so I'll replace with "NaN" instead! Thank you

gradio/external.py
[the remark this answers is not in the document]
freddyaboulton: Heads up: ended up filing the issue we talked about https://github.com/gradio-app/gradio/issues/2155 . Once this is fixed it may be possible to show an empty dataframe and have users type in all the values themselves.

BenjaminBossan (approved):
LGTM overall. I know too little about what we can expect from the input data to judge whether `cols_to_rows` and `rows_to_cols` cover all edge cases, but I couldn't spot anything obviously incorrect.

gradio/external.py
merveenoyan: it was something we didn't want to have in model card specifically 
(cc @adrinjalali is working on having dtypes atm) 
Maybe you could check for both? @freddyaboulton

merveenoyan (approved):
Thanks a lot for this PR, I will make sure this is well adopted 🙂❤️

merveenoyan:
@freddyaboulton when would this get merged? I'm planning to do a blog post on skops want to include gradio :)

Issue #2089, opened 2022-08-26
Giving examples for dataframe component as a `pandas.DataFrame`


merveenoyan:
I’m trying to give examples to dataframe component as list of lists like we do with other things, however I couldn’t succeed. when I look at the documentation, it doesn’t exist in the example as well.  The example is not parsed correctly.

The list of lists is [[header1_row_1 header2_row1 ..] [header2_row1, header2_row2 ...]] where headers are features and rows are values. Each inner list is a sample (see example below)

```
(Pdb) examples
[['material_7', 'material_8', 5, 8, 154.02, 14, 6, 16.637, 20.719, 12.824, 16.067, 15.181, 18.546, 19.402, 643.086, 6, 19.532, 11.017, 15.639, 16.709, 10.057, 20.201, 11.106, 'C'], ['material_7', 'material_8', 5, 8, 108.73, 4, 7, 16.207, 20.058, 11.898, 13.871, 14.266, 15.734, 16.886, 642.533, 9, 18.128, 11.866, 17.891, 20.302, 'NaN', 18.148, 10.221, 'C'], ['material_7', 'material_6', 6, 9, 99.84, 6, 7, 17.17, 20.858, 10.968, 16.448, 15.6, 14.637, 13.86, 673.545, 6, 'NaN', 10.064, 16.287, 17.445, 12.117, 20.659, 11.999, 'E']]
```
It also gets each input and divides the string and treats it weirdly (see screenshot below) (it says `m a t ..` for first example as you can see, where it's `material_7`)

<img width="1332" alt="Ekran Resmi 2022-08-26 14 08 38" src="https://user-images.githubusercontent.com/53175384/186900197-891f0236-e730-4c82-baf0-5f832ec8850c.png">

Also on Spaces, I get "list index out of range" for the above reason.
Can you tell me what am I doing wrong?

freddyaboulton:
Hi @merveenoyan !

I think the follwoing should work

```python
import gradio as gr
import pandas as pd


examples = [[[[7, 8, 9], [10, 11, 12]]],
            [[[5, 6, 7], [8, 9, 10]]]]

def get_max(df: pd.DataFrame) -> float:
    return df.max().max()


gr.Interface(get_max, gr.DataFrame(type="pandas", headers=["a", "b", "c"]), gr.Number(),
             examples=examples).launch()
```

but I agree we should make it possible to pass a dataframe directly and fix the examples are displayed in the UI.

merveenoyan:
[...]

PR #2128, opened 2022-08-30
Better processing of example data prior to creating dataset component

freddyaboulton:
# Description
The goal of this PR is to add some processing to the example data passed in by users to either:

1. Accept more data formats as valid examples for a given component, e.g. pandas dataframes for dataframes
2. Improve how the data is displayed in the UI, e.g. only display the filename for a file as opposed to the full path
3. Validate the data matches the expected format by the UI, e.g. an image example must be a path to an image file.

Points 1 and 2 follow my discussion with @abidlabs [here](https://huggingface.slack.com/archives/C02SPHC1KD1/p1661808604873899)

Point 3 is **not implemented** since I did not discuss that with anyone prior  but I think will improve the developer experience. Unfortunately, I don't think we'll be able to validate file-path based components due to the ambiguity between absolute paths and relative paths when reading from a csv file: #2129

In terms of design, we opted to add an `as_example` method that will be called by the Dataset component on init. Didn't want to couple the component-specific example logic with the dataset component implementation in case the developer is defining their own component. We might have to re-work how we do examples when we get to community components though.

You can test the original dataframe issue here:

```python
import gradio as gr
import pandas as pd


examples = [[pd.DataFrame([[7, 8, 9], [10, 11, 12]])],
             [pd.DataFrame([[5, 6, 7], [8, 9, 10]])]]


def get_max(df: pd.DataFrame, *args) -> float:
    return df.max().max()


gr.Interface(get_max, [gr.DataFrame(type="pandas"), gr.Number()], gr.Number(),
             examples=examples).launch()
```

Closes: #2089

# Checklist:

### DIFF
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -10,6 +10,7 @@ import math
 import numbers
 import operator
 import os
+import pathlib
 import random
 import tempfile
 import uuid
@@ -1538,9 +1539,9 @@ class Image(Editable, Clearable, Changeable, Streamable, IOComponent, ImgSeriali
         )
 
     def as_example(self, input_data):
-        is_str = isinstance(input_data, str)
-        is_file = Path(input_data).is_file()
-        if not all([is_str, is_file]):
+        is_valid = isinstance(input_data, str) or isinstance(input_data, pathlib.Path)
+        exists = Path(input_data).is_file()
+        if not all([is_valid, exists]):
             raise ValueError(
                 f"Example data for an image component must be a valid file, received {input_data}"
             )
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -1538,6 +1538,12 @@ class Image(Editable, Clearable, Changeable, Streamable, IOComponent, ImgSeriali
         )
 
     def as_example(self, input_data):
+        is_str = isinstance(input_data, str)
+        is_file = Path(input_data).is_file()
+        if not all([is_str, is_file]):
+            raise ValueError(
+                f"Example data for an image component must be a valid file, received {input_data}"
+            )
         return input_data
 
 
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -10,7 +10,6 @@ import math
 import numbers
 import operator
 import os
-import pathlib
 import random
 import tempfile
 import uuid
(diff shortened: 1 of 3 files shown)


gradio/components.py
```diff
@@ -210,6 +211,10 @@ def get_load_fn_and_initial_value(value):
             load_fn = None
         return load_fn, initial_value
 
+    @abstractmethod
+    def as_example(self, input_data):
```
freddyaboulton, line 215: Marking this as an abstract method so we don't forget to add it for new components. The trade-off is that there will be some repeated 

```python
    def as_example(self, input_data):
        return input_data
```

code but I think it's worth it for the "safety" of knowing a someone had to type out an implementation as opposed to just forgetting and inheriting from IO component.

gradio/components.py
```diff
@@ -1506,6 +1532,15 @@ def stream(
             _postprocess=_postprocess,
         )
 
+    def as_example(self, input_data):
```
freddyaboulton, line 1535: We can add this validation to other components to improve the developer experience, e.g. Audio, Video, CheckboxGroup. We can also get really detailed if we want, e.g. Image files must end in png or jpg

freddyaboulton:
@abidlabs Tagging you for review even though it's still draft just cause I want to make sure this is on the right path

abidlabs:
This looks great to me @freddyaboulton! I actually think it's fine to make `as_example()` a regular method (not an abstract method) defined in the `IOComponent` class. The default implementation should be the identity function.

Yes, we have to remember to implement it for all of the `IOComponent` subclasses, but the same is true for `preprocess()`, `postprocess()`, or any data-processing method. And given that most components' `as_example()` method is simply the identity function, this will save a lot of repeated code.

abidlabs:
I'll look into the filepath problem and the issue you created

gradio/examples.py
```diff
@@ -6,10 +6,9 @@
 import csv
 import inspect
 import os
-import shutil
```
freddyaboulton, line 9: Noticed not used

freddyaboulton:
@abidlabs This should be good for review now. Removed the abstract method and added tests. We can customize the `as_example` for different components in the future, e.g. accepting PIL images as mentioned [here](https://huggingface.slack.com/archives/C02QZLG8GMN/p1661930679607899)

abidlabs:
LGTM @freddyaboulton, tested and works great!

abidlabs (approved):

Issue #1561, opened 2022-06-14
Request Image shows in full_width and specified Height.


hpeiyan:
- [ x ] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  

Image shows in full_width and specified Height.

**Describe the solution you'd like**  

Hope the feature can be code like below, after setting full_width and specified height, the image shows in specified height and auto put in the center.

```python
gr.Image().style(full_width=True, height=60)
```

**Additional context**  
there's Image's style feature that setting specified width and height, such as 
```python
components.Image(type="pil", label="image").style(height=260, width=300)
```
but, the image shows in left of container.

<img width="402" alt="image" src="https://user-images.githubusercontent.com/12880598/173490515-9fa1ae2c-12da-43b0-8aea-eddc417025d7.png">

aliabid94:
Will take a look, thanks for opening.

agn-7:
@hpeiyan 
Doesn't work:
```
AttributeError: 'Image' object has no attribute 'style'. Did you mean: 'scale'?
```

PR #2133, opened 2022-08-31
Image size fix

aliabid94:
Height used to not work bc of typo. Fixed now.

Fixes: #1561


abidlabs:
LGTM @aliabid94, will go ahead and merge since failing tests have to do with Spaces infra

PR #2145, opened 2022-08-31
Respect original filename when cached example files are downloaded

### LINKED ISSUE #2140: When caching the output of a file component in examples, the filename is not respected
### Describe the bug

Run Model3D with cache examples - you can do it with the version hosted on spaces here: https://huggingface.co/spaces/gradio-pr-deploys/pr-2128-all-demos

Click on an example and download the resulting file, you will see that the downloaded file does not have a complete name, e.g. `_home_user_app_gradio_cached_e (2)`

### Reproduction

Stated above^

### Screenshot

model_3d_download

### Logs

```shell
-
```

### System Info

```shell
-
```

### Severity

freddyaboulton:
# Description

Fixes #2140

The downloaded filename is the same as what is returned from the prediction function now but you'll note that the extension for the file in the Model3D demo is messed up. 

This is because when the output is cached, we return a base64 representation of the file as opposed to serving the file itself. The data streams for Model3D files have `application/x-tgif` and `application/octet-stream` mime types and I think Chrome can't guess the file extension for those mime types. 

To get around this we'd have to serve the file for examples too but decided to hold off and hear other's thoughts on the matter.

# Checklist:

### DIFF
--- a/demo/model3D/run.py
+++ b/demo/model3D/run.py
@@ -1,4 +1,3 @@
-import time
 import gradio as gr
 import os
 
--- a/gradio/serializing.py
+++ b/gradio/serializing.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import os
 from abc import ABC, abstractmethod
+from pathlib import Path
 from typing import Any, Dict
 
 from gradio import processing_utils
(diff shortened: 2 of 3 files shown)


abidlabs:
Just merged 3.2 in here so the tests pass

abidlabs:
LGTM, but there is a deeper problem that you point out, which is the for cached file examples, we first serialize them to base64 (thereby losing the extension information), and then the browser has to convert them to a file for download. We really should avoid this, but for now, let's file an issue and take a look later @freddyaboulton?

abidlabs (approved):

freddyaboulton:
@abidlabs Thanks for the review! Filed #2187

PR #2147, opened 2022-08-31
Fix roundedness and form borders

aliabid94:


abidlabs:
LGTM @aliabid94! Merging so we can release 3.2 soon

Issue #2156, opened 2022-09-01
`gr.Audio()` does not maintain value after being set to invisible in Gradio 3.2


oconnoob:
### Describe the bug

When working with Gradio 3.1.6, I would switch between file and microphone audio sources, setting the deselected component to be invisible. E.g.:

Step 1 - select "Record Audio":
![image](https://user-images.githubusercontent.com/46252299/187986321-5400ef0f-d796-42c4-b18c-a1fc413663bc.png)

Step 2 - record Audio:
![image](https://user-images.githubusercontent.com/46252299/187986813-18c4dd82-06a8-4bfd-b3ee-9b8bae70be1a.png)

Step 3 - switch to "Audio File"
- This sets the `mic_data` component (`gr.Audio()`) to invisible, and the previously invisible `file_data` component (`gr.Audio()`) to be visible. 
![image](https://user-images.githubusercontent.com/46252299/187986843-17054550-c77c-4e85-b081-d04ac752642a.png)

Step 4 - switch back to "Record Audio":
- This does the reverse of the Step 3, switching visibility as appropriate
**- The recorded audio file is still present as the value of the `mic_data` component**
![image](https://user-images.githubusercontent.com/46252299/187986878-fceacd33-1267-47f1-b1d1-78ac8c94f143.png)

Now, if I upgrade to `gradio==3.2` from `gradio==3.1.6`, **Step 4 now has a different result**. Namely, the audio file does not persist, and this is the result of switching back to "Record Audio" in Step 4 **without deleting the previously recorded audio.**

![image](https://user-images.githubusercontent.com/46252299/187987313-38d8e9b8-88e4-4fe1-af1c-635d74ea4eed.png)

Note that `gr.State()` was not working in 3.1.6, so I was temporarily using global variables, although think that I've done everything right updating to `gr.State()`s and therefore the problem is not stemming from any mistake there.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Simply running the above steps in the GUI will reproduce the result, selecting either the broken or working commit to see the corresponding behavior:

### BROKEN (Gradio 3.2):
Commit: 90b6c88d33e4efcd43aaef3a17eee71fc7dc328f

### WORKING (Gradio 3.1.6):
Commit: cf9607fbbe0af33b0bfc8e1a51a250c08aff123f

### Reproduce:
```bash
git clone https://github.com/AssemblyAI-Examples/audio-intelligence-dashboard.git
[...]

PR #2157, opened 2022-09-01
Fix bugs with gr.update

### LINKED ISSUE #2154: 3.2's gr.update(visible=X) causes value to be reset
### Describe the bug

Every time I use gr.update(visible=X) on an element, its value becomes nothing. sliders get reset to zero, gr.Radio elements lose selection, ...

Here's how elements look in 3.2 after they are shown (they are invisible initially, and are shown when SD Upscale option is picked)  
firefox_Ws0SSiTDv8

Here's same with gradio downgraded to 3.1.5:
firefox_49CiIPMs9c

### Reproduction

https://github.com/AUTOMATIC1111/stable-diffusion-webui

### Screenshot

see description

Also here is js error log for below
firefox_CFKKtTSH6f

### Logs

```shell
no errors in python

same unrelated errors in javascript for both versions
```

### System Info

```shell
3.2 gradio, Windows 10, Firefox and Chrome
```

### Severity

freddyaboulton:
# Description

Fixes #2154 #2156

During #2044, I made it so that `delete_none` would keep `None` in the `value` key. This was to let users set the value of the component to `null` in the front-end which would reset the component value. The problem is that `None` is the default of the `value` key in the `update` component method, so the value is always reset, even when users are not trying to reset the value!!

The solution I went for was to distinguish the "don't update the value" and "set the value to null" use cases. For the former, we use an enum for the default value of the update method and for the latter, the user can return `None` as before.

The enum I chose is `_Keywords.NO_VALUE`. It's not exposed at the top-level gradio package and it's marked "private" by the leading underscore so it should not be autocompleted in IDES. So it should mean that the user did not pass in a value for the `"value"` key.

To test:

```python
import gradio as gr

with gr.Blocks() as demo:
    radio = gr.Radio(label="algorithm", choices=["a", "b", "c"], value="a")
    slider = gr.Slider(label="threshold", value=0, minimum=-1, maximum=1, visible=False, interactive=True)
    style = gr.Dropdown(label="style", choices=["modern", "classical", "cartoon"], value="modern", visible=False, interactive=True)

    def make_visible(algorithm):
        is_visible = algorithm == "c"
        return {slider: gr.update(visible=is_visible), style: gr.update(visible=is_visible)}
    radio.change(make_visible, inputs=radio, outputs=[slider, style])

demo.launch()
```
The slider value and drop down should be persisted when clicking different radio buttons.

I also added `blocks_update` to `all_demos` - slider value should be persisted as well.

# Checklist:

### DIFF
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -61,8 +61,7 @@ set_documentation_group("component")
 
 
 class _Keywords(Enum):
-    # Set value in front-end to null
-    NO_VALUE = "NO_VALUE"
+    NO_VALUE = "NO_VALUE"  # Used as a sentinel to determine if nothing is provided as a argument for `value` in `Component.update()`
 
 
 class Component(Block):
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -62,7 +62,7 @@ set_documentation_group("component")
 
 class Keywords(Enum):
     # Set value in front-end to null
-    VOID = "VOID"
+    NO_VALUE = "NO_VALUE"
 
 
 class Component(Block):
@@ -292,7 +292,7 @@ class Textbox(Changeable, Submittable, IOComponent, SimpleSerializable, FormComp
 
     @staticmethod
     def update(
-        value: Optional[str] = None,
+        value: Optional[str] = Keywords.NO_VALUE,
         lines: Optional[int] = None,
         max_lines: Optional[int] = None,
         placeholder: Optional[str] = None,
@@ -471,7 +471,7 @@ class Number(Changeable, Submittable, IOComponent, SimpleSerializable, FormCompo
 
     @staticmethod
     def update(
-        value: Optional[float] = None,
+        value: Optional[float] = Keywords.NO_VALUE,
         label: Optional[str] = None,
         show_label: Optional[bool] = None,
         interactive: Optional[bool] = None,
@@ -643,7 +643,7 @@ class Slider(Changeable, IOComponent, SimpleSerializable, FormComponent):
 
     @staticmethod
     def update(
-        value: Optional[float] = None,
+        value: Optional[float] = Keywords.NO_VALUE,
         minimum: Optional[float] = None,
         maximum: Optional[float] = None,
         step: Optional[float] = None,
@@ -769,7 +769,7 @@ class Checkbox(Changeable, IOComponent, SimpleSerializable, FormComponent):
 
     @staticmethod
     def update(
-        value: Optional[bool] = None,
+        value: Optional[bool] = Keywords.NO_VALUE,
         label: Optional[str] = None,
         show_label: Optional[bool] = None,
         interactive: Optional[bool] = None,
@@ -867,7 +867,7 @@ class CheckboxGroup(Changeable, IOComponent, SimpleSerializable, FormComponent):
 
     @staticmethod
     def update(
-        value: Optional[List[str]] = None,
+        value: Optional[List[str]] = Keywords.NO_VALUE,
         choices: Optional[List[str]] = None,
         label: Optional[str] = None,
         show_label: Optional[bool] = None,
@@ -1030,7 +1030,7 @@ class Radio(Changeable, IOComponent, SimpleSerializable, FormComponent):
 
     @staticmethod
     def update(
-        value: Optional[Any] = None,
+        value: Optional[Any] = Keywords.NO_VALUE,
         choices: Optional[List[str]] = None,
         label: Optional[str] = None,
         show_label: Optional[bool] = None,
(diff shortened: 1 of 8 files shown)


gradio/test_data/blocks_configs.py
```diff
@@ -59,7 +58,6 @@
             "id": 29,
             "type": "json",
             "props": {
-                "value": None,
```
freddyaboulton, line 62: Backing out changes from #2044

abidlabs:
Just wanted to confirm a few things:

1. This would be a breaking change, right, since users can no longer use `None` to reset a component. But I guess it would not too breaking since this was just introduced in #2044? 

2. Now to reset the value of a component, you pass in `gr.VOID` as the parameter to `gr.update(value=)`? an alternative might be `gr.RESET` if the purpose of the component is to reset a value.

3. Finally, I wanted to ask whether you can also return `gr.VOID` from a function directly, instead of within a `gr.update(value=)`. My preliminary experiments suggest that this doesn't work although I haven't tested thoroughly yet. I think we should have the same behavior if a user: (a) returns a value `x` from a function or (b) returns `gr.update(value=x)` from a function

abidlabs:
Also cc @aliabid94 for his feedback on this API

freddyaboulton:
@abidlabs Thanks for reviewing! Good questions!

1. Yea I guess this is a breaking change but #2044 was also technically a breaking change 🙈 
2. Yea I thought about `gr.RESET` but this doesn't reset to the original value, but rather it makes the value `null` in the front-end which was the goal of #1672 . We can also do `gr.NULL` ?
3.  Right now the logic only works for updates. We can do it so you can return `gr.VOID` from a function but that would mean modifying all the component post process functions to replace `VOID` with None.

williamberrios:
Hi @freddyaboulton. It seems that when the output is only slider, the example doesn't work as expected

`import gradio as gr

with gr.Blocks() as demo:
    radio = gr.Radio(label="algorithm", choices=["a", "b", "c"], value="a")
    slider = gr.Slider(label="threshold", value=0, minimum=-1, maximum=1, visible=False, interactive=True)

    def make_visible(algorithm):
        is_visible = algorithm == "c"
        return {slider: gr.update(visible=is_visible)}
    radio.change(make_visible, inputs=radio, outputs=[slider])

demo.launch(share = True)`

abidlabs:
Hey @freddyaboulton could we use sentinels to distinguish between the user passing in `None` for the `value` parameter versus the user leaving the `value` parameter empty? Here's how numpy does it: https://stackoverflow.com/a/60157942/5209347

If this works, it has several advantages: (1) it's not a breaking change (2) it does not introduce any key words / additional syntax that a Gradio user needs to know

freddyaboulton:
Hi @williamberrios ! Thanks for the find. That's a separate issue (not introduced in this PR). I filed an issue for that here: https://github.com/gradio-app/gradio/issues/2191

freddyaboulton:
@abidlabs Thank you for the suggestion about "sentinel values". I think it does work - I implemented it by just setting the default value of `value` in the `update` methods to be the Keyword enum I introduced in this PR. 

This should be good for another look!

abidlabs:
Will take a look now!

gradio/components.py
```diff
@@ -59,6 +60,11 @@ class DataframeData(TypedDict):
 set_documentation_group("component")
 
 
+class _Keywords(Enum):
+    # Set value in front-end to null
+    NO_VALUE = "NO_VALUE"
```
abidlabs, line 65:
```suggestion
class _Keywords(Enum):
    NO_VALUE = "NO_VALUE"  # Used as a sentinel to determine if nothing is provided as a argument for `value` in `Component.update()`
```

test/test_blocks.py
```diff
@@ -266,5 +266,20 @@ def test_blocks_do_not_filter_none_values_from_updates(io_components):
     )
 
 
+def test_blocks_does_not_replace_keyword_literal():
+
+    with gr.Blocks() as demo:
+        text = gr.Textbox()
+        btn = gr.Button(value="Reset")
+        btn.click(
+            lambda: gr.update(value="VOID"),
```
abidlabs, line 275:
```suggestion
            lambda: gr.update(value="NO_VALUE"),
```

test/test_blocks.py
```diff
@@ -266,5 +266,20 @@ def test_blocks_do_not_filter_none_values_from_updates(io_components):
...
+            inputs=[],
+            outputs=text,
+        )
+
+    output = demo.postprocess_data(0, gr.update(value="VOID"), state=None)
+    assert output[0]["value"] == "VOID"
```
abidlabs, line 281:
```suggestion
    output = demo.postprocess_data(0, gr.update(value="NO_VALUE"), state=None)
    assert output[0]["value"] == "NO_VALUE"
```

abidlabs:
Some minor comments, but LGTM! Nice fix

abidlabs (approved):

PR #2170, opened 2022-09-02
Keep embedded PngInfo metadata

cobryan05:
Maintain any text-only PngInfo metadata on a PIL image when dragging UI images between components.

# Description

Please include: 
* PNG images can have metadata attached to them. This data could, for example, be added during image generation and store the generation parameters. However, if dragging an annotated image between UI elements in Gradio, the conversion to base64 would lose the metadata
* This change maintains the metadata when dragging PNGs between UI elements

Closes: # (issue)

# Checklist:

- [x] New and existing unit tests pass locally with my changes  (well, nothing *new* breaks)


abidlabs:
Thank you so much for this contribution @cobryan05! Png metadata is very cool, and I'm glad to see that we now preserve it.

Issue #344, opened 2021-11-09
Option to add path to the url


claree007:
When the app is launched locally, it opens at http://localhost:7860. I would like to have an option to add a path to the url. So if I specify "myapp" as the custom path, the app should launch at http://localhost:7860/myapp and for "myapp/gradio", the app should launch at http://localhost:7860/myapp/gradio.

The need for this that the server on which our gradio app will be deployed has got rules where only urls starting with whitelisted paths are allowed. For example, if "myapp" is in the whitelisted paths, then urls like http://server-ip:port/myapp and http://server-ip:port/myapp/foobar will work but http://server-ip:port/somethingelse will not go through.

abidlabs:
Looking into this!

abidlabs:
Working on this now as part of #440!

abidlabs:
After looking into this some more, I think it's going to be a fair bit of effort since we will have to modify every single route in `app.py` to take a base URL prefix, along with every template that we use. Given that this seems to be a very niche use case (correct me if I'm wrong), we will push off for now. But PRs are welcome!

claree007:
I agree that only a handlful of the users might need this feature. Sure, I will open a PR for this.

abidlabs:
Thank you!

abidlabs:
Any updates on this @claree007?

claree007:
I found a way of doing what I wanted using voila. I put my gradio app inside a notebook and opened it using voila which has an option to configure the path where the UI is displayed.

abidlabs:
Interesting, thanks for letting us know @claree007!

MrNocTV:
> After looking into this some more, I think it's going to be a fair bit of effort since we will have to modify every single route in `app.py` to take a base URL prefix, along with every template that we use. Given that this seems to be a very niche use case (correct me if I'm wrong), we will push off for now. But PRs are welcome!

I think it's not that simple. I cant deploy gradio app on a K8S environment simply because the ingress path not working. 
Because the app always starts on path `/`, if I have a path in my url , let say `abc.com/myapp`and route to the service, it will return `404`. If I rewrite the path to `/`, it just wont work because I need a lot of ingress with different path. You get my point?

abidlabs:
I see -- let's open this up as a feature request

bottledmind:
Hey everyone.
Any progress here?

abidlabs:
[...]

PR #2171, opened 2022-09-03
Url path

bottledmind:
# Description

I cannot install frontend, because of pnpm. I'm getting a bunch of errors, that seem to be the same for my code and your main without an installed frontend. So, I need someone's help with running the tests on my version (and main to compare. There was an IndexError, idk if it's from frontend).
Once the tests are passing and you're satisfied with my solution, I'm gonna write some tests for my code

Please include: 
* relevant motivation 
    - check the issue below. I personally had deployment issues
* a summary of the change - 
    - added server_url_path parameter to `.launch()` method
    - check the provided path for a collision with reserved by Gradio and FastAPI
* which issue is fixed
    - https://github.com/gradio-app/gradio/issues/344
    - None

Closes: #344

# Checklist:

- [x] I have commented my code in hard-to-understand areas (I believe there are none)

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -893,6 +893,7 @@ class Blocks(BlockContext):
         show_error: bool = False,
         server_name: Optional[str] = None,
         server_port: Optional[int] = None,
+        server_url_path: str = "/",
         show_tips: bool = False,
         height: int = 500,
         width: int = 900,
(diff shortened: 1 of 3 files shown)


bottledmind:
Oh, you actually have CI

bottledmind:
Well, can someone explain me whats the difference between CI black and local black?
image

freddyaboulton:
Hi @bottledmind ! You probably have a different version of black installed locally. The dependencies on CI are installed via bash scripts/install_test_requirements.sh

abidlabs:
Hi @bottledmind, thanks for making this PR! We'd love to get a solution to this, but I think the right way to do it is to add a URL prefix, since we'd need to adjust all of the URL endpoints (such as `api/predict` and `/login'), not just the main Gradio app landing page ('/')

FastAPI supports URL prefixes in a couple of different ways, it looks like:
* Sub-applications:  https://fastapi.tiangolo.com/advanced/sub-applications/
* API Routers: https://fastapi.tiangolo.com/tutorial/bigger-applications/#another-module-with-apirouter

I think the first (sub-application) approach should work quite well. 

Secondly, we will need to make sure that in the frontend, all of the paths are updated to include the prefix as well so that they target the right API endpoints.

cc @aliabid94

bottledmind:
> Secondly, we will need to make sure that in the frontend, all of the paths are updated to include the prefix as well so that they target the right API endpoints.

If so, I think I have to close this PR since I haven't written a single JS app or even script in my life

abidlabs:
Hi @bottledmind yes, the frontend will need adjustment, otherwise it will send requests to the original API endpoint. I will close the PR for now then and we'll work on this later!

Issue #382, opened 2021-11-19
Method url_ok check is not reaching host if server_name is 0.0.0.0


cansik:
In the documentation it is mentioned that it is possible to set the `server_name` to `0.0.0.0` to host it on any network interface (LAN):
> server_name (str) - to make app accessible on local network set to "0.0.0.0".

This is problematic, because the server URL for the local flask server is generated with the `server_name`. Which means `path_to_local_server` is set to `http://0.0.0.0:7860/`. Later, the `url_ok()` method checks if this URL is reachable and if not, creates a public proxy for the server. And because [http://0.0.0.0:7860/](http://0.0.0.0:7860/) is not a reachable IP address (at least under Windows), this won't work.

Maybe it makes sense to check if the local Flask server is bound to `0.0.0.0` and replace it with `localhost` (which still is not a very nice solution, but for now fixes it for me):

```python
path_to_local_server = "http://{}:{}/".format(server_name, port)

if server_name == "0.0.0.0":
    path_to_local_server = path_to_local_server.replace(server_name, "localhost")
```

abidlabs:
Interesting... let me look into this. Thanks for the suggestion @cansik

abidlabs:
Looking into this issue (sorry for the delay @cansik). I have a question -- why are you setting the server name to "0.0.0.0" exactly? Doesn't your code work if you just leave it as "localhost"?

cansik:
As far as I know, the server name in gradio defines the local network-interface the server is listening on. So if it is localhost, only `127.0.0.1` is a valid IP to reach the sever. To bind to all network-interfaces, it is common (as well in flask) to bind to `0.0.0.0`. This allows me to reach the server over all ip addresses (`127.0.0.1`, `192.168.1.30`, whater-public-ip, ..) that are assigned to my network interfaces. This is not a valid ip address, but it is a valid host address ([wiki/0.0.0.0](https://en.wikipedia.org/wiki/0.0.0.0)).

So, we are allowed to bind to a server to `0.0.0.0`, but on some host systems (Windows), this "ip" can not be reached because it is not valid, leading to the explained problem in the issue.

abidlabs:
That makes sense! So we should handle `0.0.0.0` as a special case. Let me patch this

cansik:
I could create a PR if you want, maybe it makes sense to handle other special addresses as well. This one is only for IPv4 interfaces.

abidlabs:
[...]

PR #2176, opened 2022-09-04
Fixed misleading log when server_name is '0.0.0.0'

lamhoangtung:
# Description

Assuming I have this script:
```python
import gradio as gr
demo = gr.Blocks(title="Test")
demo.launch(server_name='0.0.0.0', server_port=4567)
```
Currently, running this with gradio 3.2 produce this log:
```console
Running on local URL:  http://localhost:4567/

```
But this is misleading to user since `localhost` is often associated  to `127.0.0.1` and the log should say this instead:
```console
Running on local URL:  http://0.0.0.0:4567/
``` 
This small PR basically fixed this misleading log

# Checklist:

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -1025,7 +1025,11 @@ class Blocks(BlockContext):
                 else:
                     print(strings.en["COLAB_DEBUG_FALSE"])
         else:
-            print(strings.en["RUNNING_LOCALLY_SEPARATED"].format(self.protocol, self.server_name, self.server_port))
+            print(
+                strings.en["RUNNING_LOCALLY_SEPARATED"].format(
+                    self.protocol, self.server_name, self.server_port
+                )
+            )
         if is_colab and self.requires_permissions:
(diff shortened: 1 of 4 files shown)


abidlabs:
Thanks for the PR @lamhoangtung! I agree with the point about what should be printed, but removing the `url_host_name` isn't a good idea because of the issue on Windows machine described in #382. 

But we can modify what we print to be clearer. Let me push a fix here and merge it in.

lamhoangtung:
That made a lot more sense to me why it was localhost before. Glad that I will be fixed soon. Thanks @abidlabs

abidlabs:
I've made some changes. Would you be able to try them out @lamhoangtung? If they work for you as well, I'll go ahead and merge these in

lamhoangtung (commented):
LGTM. Working fine on my local machine as well @abidlabs

abidlabs:
Thanks @lamhoangtung, merging. Will issue a patch release to pypi sometime next week!

Issue #2132, opened 2022-08-31
Variable's state leaked between sessions


ankrgyl:
### Describe the bug

We've observed that when our app is hosted on HuggingFace, a `gradio.Variable()` value is somehow getting leaked across sessions. My assumption is that a `Varaible()` is supposed to be isolated within a session.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

[Here](https://huggingface.co/spaces/impira/docquery/blob/67ef8151d95765a3ae43b37540f64fa26efb3ab8/app.py) is the code, the variable in question is `document`. When run locally, it's fine, but on huggingface spaces, state is bleeding across sessions.

This video walks through the repro, and shows the difference between the same code running locally vs. hosted on HF Spaces: https://www.loom.com/share/d820fb46fce24a4b9dcc6676fc4f0ab2. Feel free to play with the space here: https://huggingface.co/spaces/impira/docquery.

### Logs

```shell
N/A
```

### System Info

```shell
3.1.7 (running on HuggingFace Spaces)
```

### Severity

blocking all usage of gradio

ankrgyl:
To clarify -- although I recorded this with two browser windows on the same machine, we observed this across two different users at Impira testing our space.

abidlabs:
@ankrgyl thanks for raising this issue! A very strange and problematic bug, we'll look into it and figure out what's happening

freddyaboulton:
@ankrgyl out of curiosity, can you try launching with enable_queue=False and see if the problem persists?

ankrgyl:
Yes I'll try that now (it's rebuilding).

In the meantime, looking at the network tab, I noticed that this is the request that is sent (and ultimately generates the bad result).
<img width="573" alt="Screen Shot 2022-08-30 at 5 52 24 PM" src="https://user-images.githubusercontent.com/565363/187569230-f99357c7-77a2-4bb1-ae1c-f2505a9ffa30.png">

That middle `null` is suspiciously the `document` variable in the `inputs` for the task:

```
336     question.submit(
337         fn=process_question,
338         inputs=[question, document, model],
339         outputs=[image, output, output_text],
340     )
```

Just sharing in case it's a useful clue...

ankrgyl:
That seems to have broken it completely (on hugging face, not locally): https://www.loom.com/share/6fa87104cf214069895b48ebf2a6f01e.

[...]

Issue #776, opened 2022-03-04
[Blocks] Is it possible to support iterative updating of an output component?


abidlabs:
Not huge priority, but we've heard this request in the past, and just heard this from a user at Hugging Face, who wanted to update the images in an image output component iteratively. 

From the user: 
> I couldn't work out how to repeatedly update an image in a loop. It is probably best to refer to the example Space I linked, any help would be much appreciated, as I would prefer if it was a Gradio App!

Space they linked: https://huggingface.co/spaces/edbeeching/atari_live_model

pngwn:
Is this supposed to be like a fake gif? Rendering a series of images one after the other with a small delay between them?

abidlabs:
Yeah somewhat like that. The basic idea is to render a sequence of images (or more generally any output) _as they are returned_ by a function instead of returning a GIF all at once at the end. You could imagine a loop that returns a 100 images, but it takes a while to generate each one. In Python, you have [generators](https://realpython.com/introduction-to-python-generators/), which can yield multiple items instead of a function which returns a single item. We could potentially use that in the backend. In the frontend, the interface would have to make repeated calls to the backend and update the output component each time. 

The other common use case where we've herad that we've seen is generating GANs, which is a slow an sequential process, and one can return "intermediate" images as a cool visualization.

pngwn:
Starting to feel that we should be setting up a websocket connection with the backend rather than polling, it wouldn't make stuff like this much easier for realtime communication between client + server.

abidlabs:
Potentially related: https://discuss.huggingface.co/t/can-we-print-run-time-outputs/15810/2

kuprel:
Would love to use this for the min-dalle space:  https://huggingface.co/spaces/kuprel/min-dalle

Here's how that model works with progressive outputs: https://replicate.com/kuprel/min-dalle

freddyaboulton:
We should also be mindful to support all components (e.g. Text) and not just images. Comment from @NotNANtoN : 

Comment from https://github.com/gradio-app/gradio/issues/1973#issuecomment-1208338917

altryne:
Given how much gradio is being used now with the explosion of stable diffusion, I would love for this to be included in future releases. 
Seeing a loading progress bar is one thing, but seeing your image getting updated is a whole another experience (Midjourney bot has something similar)

abidlabs:
Yess absolutely, we'll prioritize this next week!

abidlabs:
[...]

PR #2188, opened 2022-09-06
Support for iterative outputs (#2162)

abidlabs:
This PR adds support for **iterative outputs** in `gradio`. What this means is that users can now define functions that return (or rather, `yield`) a series of outputs, which are displayed in the output component in real time, as they are returned. This is motivated by the need to display RL agents in real time, the intermediate steps of image generation models, etc. In order to use this functionality, a user simply uses the `yield` keyword in Python to define a generator function. See #776 for more implementation details.

Example:

```py
import gradio as gr
import numpy as np
import time

def fake_diffusion(steps):
    for i in range(steps):
        time.sleep(1)
        yield np.random.random((200, 200, 3))

io = gr.Interface(fake_diffusion, gr.Slider(1, 10, 3), gr.Image())
```

As part of this PR, I also fixed the session leakage issue reported in #2132. I am also releasing a beta version which can be used to test iterative outputs on Spaces, etc.

You can use these new demos to check this PR
* `demo\fake_diffusion\run.py`
* `demo\fake_diffusion_with_gif\run.py`
* `demo\count_generator\run.py`

I would appreciate a review of this PR (especially the frontend part). _However_, this PR is not ready to be merged because:
* There are no tests for iterative outputs
* We need to throughly test the new queue before making it default. Specifically, there are rough edges around using the queuing with `share=True`, as well as on jupyter notebooks with interactive sessions (running a cell the first time works, but subsequent runs do not)

Closes: #776 
Closes: #2132

Recording 2022-09-02 at 03 25 51


abidlabs:
What's left to do is to support session state and add tests, which I'll work on today. And potentially add some indicator in the UI to show that the process is iterating, as suggested by @aliabid94. @aliabid94 would you like to add that to this PR directly?

Issue #2190, opened 2022-09-06
Add manual trigger for releases to pypi


abidlabs:
For beta releases, it's fairly cumbersome to have to merge to a release branch to release to pypi. It would be great to add a manual trigger so that we can deploy to pypi by only specifying the version number.

See: https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/

PR #2192, opened 2022-09-06
Add manual trigger to deploy to pypi

abidlabs:
This creates a "run workflow" button for the "deploy to pypi" action, similar to this:

image

I haven't been able to fully test the deployment on my fork because secrets, etc. don't transfer over to the fork, but I can confirm that the button appears. I'll do a complete test once this is merged in.

Also removes the whole "merge to a branch starting with 'release'" to deploy a test branch as it's quite cumbersome and we won't need it if this works

Fixes: #2190


.github/workflows/deploy-pypi.yml
```diff
@@ -9,17 +9,17 @@
...
     branches:
       - main
-      - release**
     paths:
       - 'gradio/version.txt'
 
 jobs:
   deploy:
-    if: github.event.pull_request.merged == true
+    if: github.event.pull_request.merged == true || github.event.action == ''
```
freddyaboulton, line 22: What happens if you run this without updating the version.txt file? 

My guess is that the pypi upload would fail (since there already is a package with the current version) but I'm worried we're going to overwrite the front-end assets in the cdn.

abidlabs: Ah nice catch. I will just have it terminate if an existing version is found on aws

freddyaboulton: Sounds good! Might be good to just terminate as soon as the job starts if there's an existing version on pypi.

pypi was a json api here: https://pypi.org/pypi/gradio/json

and theres a `releases` key that has all the known releases. We could write a script to check the current `version.txt` with the pypi release history.

###Example output of api:

image

abidlabs: Should be fixed now

abidlabs: Good point about pypi, that's better than checking S3. Let me fix that up

freddyaboulton (approved):
Thanks @abidlabs ! Only pending change is checking whether or not the version exists via pypi in the github action.

abidlabs:
Added the pypi check near the beginning of the GitHub action. Thanks for the suggestion @freddyaboulton, and this should be good to go now.

PR #2193, opened 2022-09-06
Make queue per app

aliabid94:
There used to be a single class-based view. This made multiple queued apps in a notebook impossible. Refactored to tie a queue to an app to fix this.


gradio/event_queue.py
```diff
@@ -22,175 +22,146 @@ class Estimation(BaseModel):
...
-    @classmethod
-    def resume(cls):
-        cls.STOP = False
-
-    @classmethod
-    def get_active_worker_count(cls) -> int:
+        self.event_queue = []
+        self.events_pending_reconnection = []
+        self.stopped = False
+        self.max_thread_count = concurrency_count
```
abidlabs, line 36: This used to be 1. Maybe you mean:
```suggestion
        self.max_thread_count = concurrency_count or 1
```

abidlabs:
Awesome LGTM! Tested and works great in jupyter notebooks :)

abidlabs (approved):

Issue #1203, opened 2022-05-10
No Output In Browser


stuartthomson:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the bug

I see nothing in my browser when following 'Quick Start' section of the readme.

1. Create environment with python 3.7
2. `pip install gradio`
3. Run the example script

In the browser (Chrome and Edge tested) I see nothing. There errors in the console logs that say:
```
Failed to load module script: Expected a JavaScript module script but the server responded with a MIME type of "text/plain". Strict MIME type checking is enforced for module scripts per HTML spec.
```

### Reproduction

```python
import gradio as gr

def greet(name):
    return "Hello " + name + "!!"

gr.Interface(fn=greet, inputs="text", outputs="text").launch(debug=True)
```

### Screenshot

![image](https://user-images.githubusercontent.com/1838353/167616100-3ba63855-eb5c-443f-9c4d-7476917a2a69.png)

### Logs

```shell
C:\tools\miniconda3\envs\gradio\lib\site-packages\paramiko\transport.py:236: CryptographyDeprecationWarning: Blowfish has been deprecated
  "class": algorithms.Blowfish,
Running on local URL:  http://127.0.0.1:7864/

To create a public link, set `share=True` in `launch()`.
```

### System Info

```shell
Gradio: 2.9.4
Python: 3.7.13
OS: Windows 10
Browsers: Chrome, Edge
```

### Severity

blocker

biubiubiiu:
Met the same problem. I am using gradio 2.9.4 with Python 3.9.12 on Windows 10. Downgrading to 2.9 beta version doesn't work for me.

abidlabs:
Hi @stuartthomson and @biubiubiiu can you check if this issue is resolved using `gradio==3.0.2`? We've just released it today and it includes many fixes

leexinhao:
> Hi @stuartthomson and @biubiubiiu can you check if this issue is resolved using `gradio==3.0.2`? We've just released it today and it includes many fixes

I met the same problem using `gradio==3.0.2`, my os is windows11 and I 've tried Python3.9.7, Python3.7.13, Python3.7.0.

daspartho:
I'd like to work on this. I'll link a PR below.

PR #2200, opened 2022-09-07
use correct MIME type for js-script file

daspartho:
Fixes: #1203 

This makes the server use the proper MIME type for the js-script file.

I made the changes as suggested in the issue. @abidlabs could you please review the changes i made?


abidlabs:
Hi @daspartho this seems reasonable. Were you able to reproduce the original issue and can you confirm that this fixes it?

daspartho:
I reproduced the original issue and was able to resolve it with this.

abidlabs:
Great, code changes look reasonable to me. Let me format it and then merge it in!

Issue #938, opened 2022-04-06
Implementation of a collapsible accordion (for advanced options for e.g.)


apolinario:
**Is your feature request related to a problem? Please describe.**
Sometimes way too many options and parameters are exposed to the user and the UI gets cluttered.

**Describe the solution you'd like**
There could be an accordion component, similar to Streamlit's `st.expander` https://docs.streamlit.io/library/api-reference/layout/st.expander that enables you to hide certain options inside of it

abidlabs:
Agreed, this could be helpful for UIs with many inputs and parameters. cc @aliabid94 @pngwn @AK391

abidlabs:
This would also be useful for general Blocks to collapse certain sections of a long Block. For example, I'm trying to implement this in Blocks:

![image](https://user-images.githubusercontent.com/1778297/162820290-18caaa18-1c64-447e-abed-5d2aee346030.png)

PR #2208, opened 2022-09-08
Add accordion component

aliabid94:
Allows users to wrap elements in an accordion. It's been requested several times and I thought this might be useful for the more complex SD demos being created right now.

Fixes: #938 

Use:
```python
with gr.Accordion("open up"):
   # components here
```

Recording 2022-09-07 at 17 04 33

### DIFF
--- a/gradio/layouts.py
+++ b/gradio/layouts.py
@@ -296,3 +296,42 @@ class Box(BlockContext):
 class Form(BlockContext):
     def get_config(self):
         return {"type": "form", **super().get_config()}
+
+
+@document()
+class Accordion(BlockContext):
+    """
+    Accordion is a layout element which can be toggled to show/hide the contained content.
+    Example:
+        with gradio.Accordion("See Details"):
+            gr.Markdown("lorem ipsum")
+    """
+
+    def __init__(self, label, *, open: bool = True, **kwargs):
+        """
+        Parameters:
+            label: name of accordion section.
+            open: An optional string that is assigned as the id of this component in the HTML DOM. Can be used for targeting CSS styles.
+        """
+        self.label = label
+        self.open = open
+        super().__init__(**kwargs)
+
+    def get_config(self):
+        return {
+            "type": "accordion",
+            "open": self.open,
+            "label": self.label,
+            **super().get_config(),
+        }
+
+    @staticmethod
+    def update(
+        open: Optional[bool] = None,
+        visible: Optional[bool] = None,
+    ):
+        return {
+            "visible": visible,
+            "open": open,
+            "__type__": "update",
+        }
--- a/gradio/layouts.py
+++ b/gradio/layouts.py
@@ -307,15 +307,18 @@ class Accordion(BlockContext):
             gr.Markdown("lorem ipsum")
     """
 
-    def __init__(self, label, *, open: bool = True, **kwargs):
+    def __init__(
+        self, label, *, open: bool = True, elem_id: Optional[str] = None, **kwargs
+    ):
         """
         Parameters:
             label: name of accordion section.
-            open: An optional string that is assigned as the id of this component in the HTML DOM. Can be used for targeting CSS styles.
+            open: if True, accordion is open by default.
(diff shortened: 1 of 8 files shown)


abidlabs:
Yesss very cool @aliabid94!

gradio/layouts.py
```diff
@@ -296,3 +296,42 @@ def style(
...
+    Example:
+        with gradio.Accordion("See Details"):
+            gr.Markdown("lorem ipsum")
+    """
+
+    def __init__(self, label, *, open: bool = True, **kwargs):
+        """
+        Parameters:
+            label: name of accordion section.
+            open: An optional string that is assigned as the id of this component in the HTML DOM. Can be used for targeting CSS styles.
```
abidlabs, line 314: * Fix docstring for `open`

abidlabs: `open` is a Python keyword so syntax highlighters annoyingly highlight it. WDYT about naming this `closed` instead?

abidlabs:
Would it be possible to support `gr.Accordion()` in `gr.Interface()`? The reason I ask is because a lot of `gr.Interface()` users would probably love to have a section of collapsible inputs as well. I can imagine something like the following API:

```py
gr.Interface(
  image_generator,
  [gr.Textbox(),
   gr.Accordion("Advanced Options", [
     gr.Slider(),
     gr.Slider(),
    ]
  ]
  gr.Image()
]
```

Otherwise, LGTM @aliabid94!

abidlabs:
One other thing: the label of the Accordion is much bigger, darker, and bolder than the label of the other components as you can see here:

image

I think it stands out *too* much. What about making it the same font style as the title of tabs:

image

aliabid94:
ok made the label smaller and not bold. Also added mention to guides. Adding to Interface will be a lot of work, let's make a separate issue for that. Let's get this in for 3.3 @abidlabs

guides/3)building_with_blocks/2)controlling_layout.md
```diff
@@ -31,7 +31,7 @@ $demo_rows_and_columns
 
 See how the first column has two Textboxes arranged vertically. The second column has an Image and Button arranged vertically. Notice how the relative widths of the two columns is set by the `scale` parameter. The column with twice the `scale` value takes up twice the width.
 
-## Tabs
+## Tabs amd Accordions
```
abidlabs, line 34:
```suggestion
## Tabs and Accordions
```

guides/3)building_with_blocks/2)controlling_layout.md
```diff
@@ -40,6 +40,8 @@ For example:
 $code_blocks_flipper
 $demo_blocks_flipper
 
+Also note the Accordion that can be opened and closed - another layout element to selectively show content.
```
abidlabs, line 43:
```suggestion
Also note the `gradio.Accordion('label')` in this example. The Accordion is a layout that can be toggled open or close. Like `Tabs`, it is another layout element that can selectively hide or show content. Any components that are defined inside of a `with gradio.Accordion('label'):` will be hidden or shown when the accordion's toggle icon is clicked.
```

abidlabs:
Ok sounds good! LGTM

abidlabs (approved):

aliabid94:
oh whoops merged before reading comments, will make another pr for them

pngwn:
I missed this PR but I don't understand why this component was added instead of just adding some options to `gr.Group`, those were the conversations we had in the past and wouldn't have significantly increased API surface area like a new component does.

abidlabs:
Agree with communicating on this more, but from my understanding `gr.Group()` is designed to change the padding/display of children. `gr.Accordion()` has a pretty orthogonal functionality (toggling the children), so I think it's clearer to have a separate class for this.

pngwn:
I may have meant `gr.Box` which is mostly visual and could easily have taken some additional arguments.

There are also some issues with the implementation, I'll take a look at them when I get time.

aliabid94:
Oh sorry yeah. Makes sense to incorporate into `gr.Box`

Issue #1983, opened 2022-08-09
Write Tests for Queueing


FarukOzderim:
**Is your feature request related to a problem? Please describe.**  
We need to write tests for the new event queue in the Frontend.

**Additional context**  
Websocket testing tooling is seems to not work well in Python, and we're unable to test it in Backend.

Testing a single event case and multiple events case would be nice.

abidlabs:
We should write unit tests for `queueing.py` as well!

hibestil:
@abidlabs I'd like to write unit tests for 'queueing.py' if it is still unassigned. Please let me know

FarukOzderim:
@abidlabs due to limitations of tooling we cannot test functionality of the queue in Backend. That's why there needs to be queue functionality tests in Frontend IMO. If that is not possible, then some manual testing will be necessary, and spaces demos has queue activated by default, automatic spaces demo deployments will help 👍 

> Websocket testing tooling is seems to not work well in Python, and we're unable to test it in Backend.

PR #2214, opened 2022-09-08
Tests for Queue

abidlabs:
Adds unit tests for queuing. In some cases, it was hard to test behavior and not implementation, so any feedback on making the tests better is welcome. But I think it's important to have some unit tests for queuing in place as we work on improving it to make sure we don't inadvertently break anything.

Fixes: #1983

Some additional notes:
* `event_queue.py` has been renamed to `queue.py`


test/test_routes.py
```diff
@@ -12,6 +14,11 @@
 os.environ["GRADIO_ANALYTICS_ENABLED"] = "False"
 
 
+class AsyncMock(MagicMock):
```
freddyaboulton, line 17: This class isn't used anywhere

gradio/queue.py
```diff
@@ -184,6 +184,7 @@ async def send_estimation(
         client_awake = await self.send_message(event, estimation.dict())
         if not client_awake:
             await self.clean_event(event)
+        return estimation
```
freddyaboulton, line 187: This messes up the type-hint since we actually return something now

freddyaboulton (approved):
Thank you for adding the tests @abidlabs ! I think these back end tests are important even though it's hard to capture the behavior of the queue absent the front-end.

We can revisit this when we have expanded integration tests in place.

Issue #2081, opened 2022-08-24
Enable queuing by default everywhere


abidlabs:
Given that the new queue offers a better experience for users (for example, by allowing inference requests to exceed 60 seconds), it would be great if we can enable queueing by default everywhere, just like it is enabled on Hugging Face Spaces

abidlabs:
Duplicate of #2969

PR #2215, opened 2022-09-08
Queue on by default

abidlabs:
We've made the queueing mechanism a lot better over the past few weeks, and I think it's ready for primetime! This PR makes queuing on by default. It can still be disabled if someone sets `.queue(default_enabled=False)` or -- using the deprecated approach -- sets `.launch(enable_queue=False)`.

Since we just released `gradio==3.3`, I think this is the right time to turn on the switch as it will allow us a reasonable amount of time to test the queue before releasing it to pypi. 

cc @aliabid94 

Fixes: #2081


abidlabs:
Discussed with @aliabid94 and there are a few changes we should make here:
* Set the number of workers to be 40 at default
* Set a maximum queue size

Also should probably wait until we've resolved the #2374 and #2399 before merging this, so changing this to draft status.

abidlabs:
Given that we want Colab demos to run locally by default (see #2455), I think we should enable queue by default _unless_ running in a Colab notebook. WDYT @aliabid94?

abidlabs:
Let's close this issue until we have some clarity around how we will handle colab notebooks.

Issue #2216, opened 2022-09-09
Have separate queues and ETAs for each event


abidlabs:
Right now, all events are added to the same queue. However, a single Blocks might have some events that are much faster than others. Instead of queuing them in one queue, we should have separate queues for different events (we would need to have some mechanism for how to delegate workers to different queues). We should also then have separate ETAs for each queue.

abidlabs:
I'm planning on implementing this using [Fair-share scheduling](https://en.wikipedia.org/wiki/Fair-share_scheduling), which means that the workers will basically rotate between events in a round-robin fashion, spending roughly an equal time on each event. 

So let's say there are 3 events: A, B, C. The `rotation_time` parameter (which is a single parameter for the entire Queue object) defines how long the workers spend on each event.

So let's say the queue looks like this: A, B, A, A, B, C, B, A, C, A, B and for simplicity assume one worker, and assume `rotation_time` is 5 seconds (a reasonable default I think). Let's say event A takes 0.5 seconds to compute (e.g. a "clear button" or some other quick event) event B takes 2 seconds (e.g. like a "flag" button), and event C takes 15 seconds (e.g. like an image generation). 

1. The worker will start by working on event A, complete that in 0.5 seconds, and then will proceed to the next event A, complete that in 0.5 seconds, and then eventually complete all of the event A's.
2. Then the worker will move to event B, complete that in 2 seconds, move to the next event B, then the next event B. After 3 event B's, then the worker will notice that more than 5 seconds have elapsed, and then 
3. Move to event C, and will complete that in 15 seconds. Since more than 5 seconds have elapsed, the worker will
4. Move back to event A. If no new event A's have arrived, the worker will move to event B

... and so on

If there are multiple workers, then each worker will independently proceed in this fashion.

FarukOzderim:
This looks great ideally but I am not sure this is really needed in Gradio.

1. This will make Queue design more complex. I am not sure how easy it will be to context-switch in Python scope, and how efficient it'll be. And it will probably take a significant time to develop this feature.
2. Long duration events will have quite a lot of trouble finishing when there is a high traffic.
3. The queue location will not be respected like before.

Due to these reasons, I think it could be better to provide these 2 options instead.
1. Letting events bypass the queue(this is already supported, might need more documentation and marketing)
[...]

Issue #2016, opened 2022-08-15
Write an Advanced Guide about Queue, Concurrency, Batching in Gradio


FarukOzderim:
With initiatives in #1598 we are very flexible and allow quite a lot of options to our users. Writing a Guide on this could be really helpful. We could also talk about technical side of each of the options.

Possible overall layout:

1. General information about fastapi server and how sync and async functions are run in the server.(sync in threadpool, async in event-loop, see [here](https://fastapi.tiangolo.com/async/))
1. Using the default launch(without queue) for low traffic.(also note that queue is enabled by default in spaces)
2. Increasing number of threads in the starlette server, see [here](https://github.com/gradio-app/gradio/blob/d7e5251c5ec5d581d7fdf6d0391f4a783e4f1171/gradio/blocks.py#L838), this is useful especially for remote servers with high throughput and parallelization.
**example: demo/concurrency-without-queue**
3. Using the default queue, which will run every request sequentially.
4. Configuring the queue parameters, especially concurrency_count. It depends on the server's concurrency optimality, which would depend on GPU power, ML Model size, and etc. See [here](https://github.com/gradio-app/gradio/blob/d7e5251c5ec5d581d7fdf6d0391f4a783e4f1171/gradio/blocks.py#L771)
**example: demo/concurrency-with-queue**

PR #2218, opened 2022-09-09
Batching

abidlabs:
Ignore this giant thread and go all the way to the bottom: https://github.com/gradio-app/gradio/pull/2218#issuecomment-1258823942

⬇️ ⬇️ ⬇️ ⬇️ ⬇️ ⬇️ ⬇️


aliabid94:
makes sense. Should `batch_size` be tied to the event listener instead of the queue? do batch functions usually support a single batch size? what happens if there's less than that (e.g. 5 data points in the batch for a batch_fn that takes 8 batches when the timeout goes off)

abidlabs:
No I don't think batched functions typically support a single batch size. Usually, you use the largest batch size that you expect to fit in memory. So there would be no real problem if a batch of size 5 samples were to get fed into a batch_fn where we've set that batch_size to be 8)

Would you agree with that @apolinario @pcuenca?

pcuenca:
Yes, except when using JAX, where a multiple of 8 is often required for parallelization and batch sizes are therefore fixed, usually. But this could be handled by the user in the `batch_fn`, I guess. I don't think it's necessary to add more complexity to the queue by introducing a `fixed_size` parameter or something like that, what do you think?

freddyaboulton:
@abidlabs This looks great! I agree with @aliabid94 that it may be better to tie the batch fn with the event listener as opposed to the queue. With this approach, it seems only one function can be batched per demo.

aliabid94:
Agree with @FarukOzderim, why do we need a separate batch fn? And with @freddyaboulton that we shouldn't add another API endpoint. Here's the syntax I have in mind:
`btn.click(fn, inputs, outputs, batch=True, max_batch_size=8)`
and everything should automatically be handled from there.

abidlabs:
Thanks for all of the suggestions guys! This makes a lot of sense, we don't actually need a separate `batch_fn`. If batch mode is enabled, we _require_ the function to work with a batch of inputs, which also avoids us creating a separate API endpoint. So the new proposed design is:

## User API
- To enable "batch mode", the user creates an event trigger that looks like this: `btn.click(fn, inputs, outputs, batch=True)`. 
- To configure the batch processing, the user can also pass in two optional parameters: `max_batch_size=8`, `batch_timeout=1`
- We also add these parameters (`default_batch=False`, `default_max_batch_size=8`, `default_batch_timeout=1`) to `queue()`, which applies them to all of the backend functions for Blocks. The main advantage of this would be making it easier to get batching to work with `Interface` (**open to other suggestions here**)

## Backend changes
- The queue will have a new `process_batch_events()` method to handle a batch of events, but this is called from the regular 
 `/api/{api_name}` endpoint if `batch=True` for that particular function

abidlabs:
> I'm not sure if adding those parameters to the queue method is the best because I don't think it's a good idea to apply the same batching behavior to all backend functions. For Interfaces, for example, I don't think we want to batch the flagging or at least not the same way we batch the main prediction function.

> What if we control the batching behavior for interfaces via the constructor?

I had originally avoided that because (a) I didn't want to add too many parameters to the `Interface` constructor, and (b) I thought that these parameters only made sense if queueing was enabled

But now that I think about it, if batch mode is enabled, we require the user to pass in a `fn` that can handle batch inputs, so it would make sense to put it in the same class as the `fn`, which in this case would the be `Interface` constructor. And strictly speaking, we don't need queuing to enable batch mode. You could pass in a batch of inputs through http requests as well

So yeah I think this is a better approach, thanks @freddyaboulton

abidlabs:
Putting some thoughts in writing...

When `batch=True`, the following should happen:

* If a user calls the Interface/Blocks as a function, it should accept a batch of size up to `max_batch_size` (`batch_timeout` is irrelevant here)
* If a user calls the `/api/predict` endpoint, it should also accept a batch of data (the API docs will need to be updated to reflect this) (`batch_timeout` is irrelevant here)
* If a user launches() the UI without queueing, the UI will directly call `/api/predict` with a batch size of 1
* If a user launches() the UI with queueing, the queue should wait until `batch_timeout` seconds pass or batch is filled before internally calling `/api/predict` with the batch

abidlabs:
**All right folks, this is finally ready for review!** 

### User API
* To enable "batch mode", the user creates an event trigger that looks like this: `btn.click(fn, inputs, outputs, batch=True)`.
* To configure the batch processing, the user can also pass an optional parameter: `max_batch_size=8`, 
* Note that I got rid of the `batch_timeout` parameter as it adds unnecessary complexity for v1 of queuing. And for viral demos for long queues, it shouldn't make any difference.
* When using Interface, a full example might look like this:

```py
import gradio as gr
import time

def test_batch(batch):
    rv = []
    for sample in batch:
        x, y = sample
        rv.append([y, x])
    time.sleep(3)
    return rv


interface = gr.Interface(test_batch, ["textbox", "textbox"], ["textbox", "textbox"], batch=True, max_batch_size=16)
interface.queue()
interface.launch()
```

gradio/queue.py
```diff
@@ -211,70 +240,91 @@ def get_estimation(self) -> Estimation:
             queue_eta=self.queue_duration,
         )
 
-    async def call_prediction(self, event: Event):
+    async def call_prediction(self, events: List[Event], batch: bool):
+        data = events[0].data
+        if batch:
+            data["data"] = [event.data["data"] for event in events]
```
freddyaboulton, line 246: nit: but my editor is complaining about the types here. Event.data is supposed to be `PredictBody` but it's treated as a dict.

gradio/events.py
```diff
@@ -183,13 +203,15 @@ def edit(
         built-in editor. This method can be used when this component is in a Gradio Blocks.
 
         Parameters:
-            fn: Callable function
-            inputs: List of inputs
-            outputs: List of outputs
+            fn: Callable function. When batch is False, each parameter of the function corresponds to one input component, and each return value corresponds to one output component. But if batch is True, the function must accept a *single* parameter, which is a *nested* list of inputs, where the outer list represents the batch, and the inner list represents the inputs to underlying prediction. For example, if the prediction takes two inputs, the input to the function would be [[input1, input2], [input1, input2], [input1, input2]]. Similarly, the function must return a single *nested* list of outputs. For example, if the prediction has one output, the output of the function should be [[output1], [output1], [output1]].
```
freddyaboulton, line 206: Having the function expect a list of lists and return a list of lists is a natural api but it may get a bit clunky. For example, for a gallery it'd be a list[list[list[image]]] . Not sure if there's a better alternative.

gradio/queue.py
```diff
@@ -211,70 +240,91 @@ def get_estimation(self) -> Estimation:
...
-                await self.send_message(
-                    event,
-                    {
-                        "msg": "process_completed",
-                        "output": response.json,
-                        "success": response.status == 200,
-                    },
-                )
+                for e, event in enumerate(awake_events):
+                    output = copy.deepcopy(response.json)
```
freddyaboulton, line 306: I'm wondering if doing this deep copy for every event will be too wasteful. Imagine if it's a diffusion model with a gallery of 8 images and a batch size of 8. We'd be copying over the 64 images for element in the batch.

gradio/blocks.py
```diff
@@ -685,16 +696,84 @@ async def call_function(self, fn_index, processed_input, iterator=None):
             "iterator": iterator,
         }
 
-    def postprocess_data(self, fn_index, predictions, state):
+    async def call_batch_function(
```
freddyaboulton, line 699: Do we need a separate `call_batch_function` method?

Provided `processed_input` is the processed input for the batch isn't the following true?
```
call_batch_function(fn_index=0, processed_input_batch=batch) == call_function(fn_index, processed_input=batch, iterator=None)
```

ui/packages/app/src/api.ts
```diff
@@ -129,7 +129,7 @@ export const fn =
 			}
 			const WS_ENDPOINT = `${ws_protocol}//${ws_host}${ws_path}queue/join`;
 
-			var websocket = new WebSocket(WS_ENDPOINT);
+			var websocket = new WebSocket(WS_ENDPOINT + `?fn=${fn_index}`);
```
freddyaboulton, line 132: nit: I think naming the parameter `fn_index` would be more consistent with the rest of the routes/codebase.

gradio/events.py
pcuenca: This is a tough one, I can't think of a better solution. I'm not sure I follow the three nested lists in your example, wouldn't it be `[[b1i1, b1i2], [b2i1, b2i2]]` for a gallery of 2 images and a batch size of 2?

gradio/queue.py
abidlabs: Yeah let me rewrite this

ui/packages/app/src/api.ts
abidlabs: Sorry I must be missing something but isn't the name already `fn_index`?

gradio/queue.py
```diff
@@ -81,10 +98,24 @@ async def start_processing(self) -> None:
...
+                batch_size = self.blocks_dependencies[event_fn_index]["max_batch_size"]
+
+                if batch:
+                    while len(events) < batch_size:
+                        fn_indices = [ev.fn_index for ev in self.event_queue]
+                        try:
+                            index = fn_indices.index(event_fn_index)
+                            events.append(self.event_queue.pop(index))
+                        except ValueError:
+                            break
```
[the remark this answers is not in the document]
abidlabs: @FarukOzderim didn't you suggest not having separate queues for separate events because it would add unnecessary complexity in #2216? :D

I agree some of the "fill batch" logic could be eliminated by having separate queues. However, I think it would be replaced with similarly complex logic with the workers having to pick events from different queues, which is why we closed #2216. If I had to pick between the two, I think this approach is simpler

In addition, I'll rewrite this code to make the "fill batch" logic faster (especially since it's a blocking operation)

gradio/blocks.py
abidlabs: The main difference is that we pass in the input data as a list in `call_batch_function`: `block_fn.fn(processed_input_batch)` and we unpack the list in `call_function`: `block_fn.fn(*processed_input)`. 

I suppose this could be consolidated, though I'll wait until we finalize the API for batched functions

gradio/queue.py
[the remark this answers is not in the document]
FarukOzderim: @abidlabs did I do that? haha :D

The part that I was against was scheduling and complexity coming with it. One of the alternatives I suggested over there aligns with this current suggestion:

> 
Allocating special thread slots to events, a thread that will just serve to a single event. (OS already does scheduling between threads and this will move the complexity to the OS instead of Gradio. Honestly, It feels unnecessary to implement a scheduling in Gradio when OS already schedules the threads.

My view point in both suggestions is complexity, I think handling batched event in a different queue concurrently (without getting into any scheduling complexity) is more simpler and easier to implement&maintain. It is better to make use of easiness of Python coroutines. Though at the same time, it is not a very big difference. Let me know how these look to you @abidlabs @freddyaboulton.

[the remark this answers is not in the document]
FarukOzderim: Actually, I think you are right, a single queue would be more simpler. Seeing searching for the next events part carried me to the suggestions above. Smt like this would be quite simple. There would be a wait list for each batched event.

```
batch_wait_list = []
...


while loop
...
event = queue.pop()

if event.batch:
  batch_wait_list.append(event)
  if batch_wait_duration > time_threshold:
    await process_batch(batch_wait_list);
  if len(batch_wait_list) < batch_size:
    continue
  else:
    await process_batch(batch_wait_list);

else:
  await process_event(event)
```

gradio/blocks.py
abidlabs: Removed

gradio/queue.py
```diff
@@ -81,10 +98,24 @@ async def start_processing(self) -> None:
 
             # Using mutex to avoid editing a list in use
             async with self.delete_lock:
-                event = self.event_queue.pop(0)
-
-            self.active_jobs[self.active_jobs.index(None)] = event
-            run_coro_in_background(self.process_event, event)
+                first_event = self.event_queue.pop(0)
```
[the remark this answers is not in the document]
abidlabs: Split into separate method and added tests, thanks for the suggestion!

gradio/queue.py
abidlabs: Fixed, thanks

gradio/queue.py
abidlabs: Rewrote to use a single copy

gradio/queue.py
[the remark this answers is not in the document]
abidlabs: Rewrote the code to be more efficient. I want to avoid using `time_threshold` for v1 of the queue.

gradio/queue.py
```diff
@@ -69,6 +85,26 @@ def get_active_worker_count(self) -> int:
...
+
+        event_fn_index = first_event.fn_index
+        batch = self.blocks_dependencies[event_fn_index]["batch"]
+
+        if batch:
+            batch_size = self.blocks_dependencies[event_fn_index]["max_batch_size"]
+            rest_of_batch = [
+                event for event in self.event_queue if event.fn_index == event_fn_index
+            ][: batch_size - 1]
+            events.extend(rest_of_batch)
```
FarukOzderim, line 103: I strongly think that this should not be handled like this, there are 2 advantages 
1. This will search the whole queue, not efficient
2. It will not respect the queue rank for other event types

I suggest this approach, keep a batch_list for every batched event, and save batchy events
```

batch_wait_list = []
...


while loop
...
event = queue.pop()

if event.batch:
  batch_wait_list.append(event)
  if batch_wait_duration > time_threshold:
    await process_batch(batch_wait_list);
  if len(batch_wait_list) < batch_size:
    continue
  else:
    await process_batch(batch_wait_list);

else:
  await process_event(event)
```

abidlabs: Thanks @FarukOzderim for the feedback, what do you mean by "It will not respect the queue rank for other event types"?

FarukOzderim: well, when the batched event gets popped, it finds the other several batched events irrespective of their rank. Even if they are in the end of the queue.

abidlabs: @FarukOzderim I may be misunderstanding, but the approach you are proposing will not work because it assigns all batched events to a single `batch_wait_list`. But you can only process events in a batch if they share the same underlying function (`fn_index`). 

If you are suggesting that we have separate `batch_wait_lists` for each event, then it effectively becomes the same approach as what I have already implemented, except that it is unclear to me in your implementation how to select between different `batch_wait_lists` when a worker is free.

> well, when the batched event gets popped, it finds the other several batched events irrespective of their rank. Even if they are in the end of the queue.

This is not accurate, as what happens is `rest_of_batch` takes the _first_ `max_batch_size` events in the event queue that share the same function index. So it does respect the order of the queue, but it prioritizes events that share the same function index in order to facilitate batching.

FarukOzderim: > well, when the batched event gets popped, it finds the other several batched events irrespective of their rank. Even if they are in the end of the queue.
This is irrespective of their rank compared to unbatched events.
Extreme example, assume there are 2 batched events in the queue,
1st one is in the 1st rank
2nd one is in the 1000st rank
In this case rank of events in 2-999 are not respected.

> If you are suggesting that we have separate batch_wait_lists for each event, then it effectively becomes the same approach as what I have already implemented, except that it is unclear to me in your implementation how to select between different batch_wait_lists when a worker is free.

It is not the same thing, first of all with your approach there is a O(n) list copy-search everytime a batched event is popped, which will hurt the performance considerably. The difference lies in performance and execution time. Current version executes with the first batched event, my suggestion executes with the last batch event. (considering the batch-size) 
Selecting a batch_list is quite trivial actually, let me make the pseudo code more detailed.

```

batch_wait_list_list = [ [] * fn_index_count ]

...


while loop
...
event = queue.pop()

if event.batch:
  batch_wait_list = batch_wait_list_list[event.fn_index]
  batch_wait_list.append(event)
  if batch_wait_duration > time_threshold: #Never mind threshold for now, it can also be tracked with a list
    await process_batch(batch_wait_list);
  if len(batch_wait_list) < batch_size:
    continue
  else:
    await process_batch(batch_wait_list);

else:
  await process_event(event)
```

abidlabs: Ah that's much clearer now @FarukOzderim. However, this approach has one problem in my mind, which is that it could result in an arbitrarily long execution time for a user whose request is at the beginning of the queue, if their batch never "fills up". Imagine the following queue (I'm just writing the `fn_index` for each event), in which event `1` is a batch event:

`[1, 2, 2, 2, 2, 2 ...]`

If we get an influx of event `2`s, event `1` will never get processed. That is why I prefer the approach of going from the beginning of the queue, because it will never result in arbitrarily long wait times for people at the beginning of the queue. If a user is in position $i$, and the runtime for a batch is $t$, we can guarantee that the user has to wait less than $i \cdot t$.

Now in theory this could overcome with an event-specific `batch_wait_duration`, but I would really like to avoid adding that hyperparameter at this time, because if it's improperly set, it could result in a *worse* performance with batching than without.

That being said, I agree with your point about the look-up time being $O(n)$ and how that could be expensive if the queue size is large. Let me benchmark it and if it is expensive, then I'll use a different structure that will allow $O(1)$ lookups.

FarukOzderim: @abidlabs IMO the solution to that problem is `batch_wait_duration_threshold` and compromising on other parts is not a good idea. I don't think setting this threshold as 60 secs is a problem. If batch is being used than letting people wait is just expected.

Nonetheless if you are too determined about this just start a search from the beginning of the queue, instead of copying and searching all the queue over here
```
rest_of_batch = [
                event for event in self.event_queue if event.fn_index == event_fn_index
            ][: batch_size - 1]
```

abidlabs:
Before anyone does a review, I realized that this PR still needs to:

* add support for examples 
* handle the case if some of the examples in the batch error out
Fun stuff 

:D

gradio/queue.py
abidlabs: @FarukOzderim I profiled this code and the latency is not a big concern. Sample code:

```py
import time
import random

queue = [{'index': random.randint(0, 5)} for _ in range(10000)]

t = time.time()
rest_of_batch = [e for e in queue if e['index'] == 3][:30]
print(time.time() - t)
```

You can see that the latency is ~1 ms even for a queue size of 10k items.

abidlabs:
All righty, this is finally ready for another review! @freddyaboulton would you be able to take a look? I've merged in your changes as well as added support for examples in batch mode and error handling.

abidlabs:
Merged in `main`, the biggest sources of conflicts were between this branch and the branch that added the ability to cancel events. I've resolved the conflicts, but it would be good to ensure that canceling continues to work as expected. Also you can now cancel some of the events within a batch and the rest of the events within the batch will continue to work as expected. Here's how you can test it out:

Run the following:

```py
import gradio as gr
import time

def trim_words(words, lens):
    trimmed_words = []
    time.sleep(5)
    for w, l in zip(words, lens):
        trimmed_words.append(w[:l])        
    return [trimmed_words]

with gr.Blocks() as demo:
    with gr.Row():
        word = gr.Textbox(label="word", value="abc")
        leng = gr.Number(label="leng", precision=0, value=1)
        output = gr.Textbox(label="Output")
    with gr.Row():
        run = gr.Button()
        canc = gr.Button("Cancel", variant="stop")

    event = run.click(trim_words, [word, leng], output, batch=True, max_batch_size=16)
    canc.click(None, None, None, cancels=event)

demo.queue()
demo.launch()
```

Open up 4 separate tabs. Then:
1. Run all 4 of the tabs
2. Go the 2nd tab, and cancel the event
3. Tabs 3 and 4 should continue to run and complete together

gradio/queue.py
freddyaboulton: Calling `remove` may be inefficient cause I think python has to move all the elements of the list whenever you remove something but I suspect it would only be noticed for very large batches (which are rare)

gradio/examples.py
```diff
@@ -5,22 +5,19 @@
...
 
 from gradio import utils
-from gradio.blocks import convert_component_dict_to_list, postprocess_update_dict
 from gradio.components import Dataset
 from gradio.context import Context
 from gradio.documentation import document, set_documentation_group
 from gradio.flagging import CSVLogger
 
 if TYPE_CHECKING:  # Only import for type checking (to avoid circular imports).
+    from gradio.blocks import Blocks
```
freddyaboulton, line 20: This import is not used anywhere

gradio/examples.py
```diff
@@ -5,22 +5,19 @@
...
 import ast
 import csv
-import inspect
 import os
 import warnings
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Callable, List, Optional
-
-import anyio
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Optional
```
freddyaboulton, line 11: Dict is not used

gradio/blocks.py
```diff
@@ -768,34 +821,61 @@ async def process_api(
...
+            if inspect.isasyncgenfunction(block_fn.fn) or inspect.isgeneratorfunction(
+                block_fn.fn
+            ):
+                raise ValueError("Gradio does not support generators in batch mode.")
+            if not all(x == batch_size for x in batch_sizes):
+                raise ValueError(
+                    f"All inputs to a batch function must have the same length but instead have sizes: {batch_sizes}."
+                )
+            if batch_size > max_batch_size:
+                raise ValueError(
```
freddyaboulton, line 854: We should raise an error when you try to batch a function and the queue is not enabled.

gradio/routes.py
```diff
@@ -313,6 +289,9 @@ async def run_predict(
                     content={"error": str(error) if show_error else None},
                     status_code=500,
                 )
+
+            if not (body.batched) and batch:
```
freddyaboulton, line 293: This is confusing because shouldn't the predict body always match what's in the config? I think we can get rid of `body.batched`?

freddyaboulton (commented):
Looks good @abidlabs ! Awesome PR.

Unit tests and code looks good to me. I tried with some examples locally and wasn't able to break anything.

I also tested and cancelling does seem to work as expected. If the batch is already being processed, there's no way to remove the cancelled event from the batch but the frontend will disconnect the queue so the result won't be shown. 

Before merge I'd like to:

- [ ] Make sure gr.Blocks.Load works with batched functions
- [ ] Run the benchmark script on a space with a batched fn to make sure it's ok

Also for some reason stream_frames and stream_audio are broken on the deployed space for this PR?


image

Is it introduced in this PR? Don't see it for PR 2488: https://huggingface.co/spaces/gradio-pr-deploys/pr-2488-all-demos

CHANGELOG.md
```diff
@@ -133,6 +133,7 @@ No changes to highlight.
 
 * When an `Image` component is set to `source="upload"`, it is now possible to drag and drop and image to replace a previously uploaded image by [@pngwn](https://github.com/pngwn) in [PR 1711](https://github.com/gradio-app/gradio/issues/1711)
 * The `gr.Dataset` component now accepts `HTML` and `Markdown` components by [@abidlabs](https://github.com/abidlabs) in [PR 2437](https://github.com/gradio-app/gradio/pull/2437)
+* Gradio now supports batch functions by [@abidlabs](https://github.com/abidlabs) in [PR 2218](https://github.com/gradio-app/gradio/pull/2218)
```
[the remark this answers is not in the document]
abidlabs: Whoops thanks

abidlabs:
Thanks for the detailed review @freddyaboulton! Let me go through the issues you identified and fix this up.

Another thing we need to do is add batched functions to our guides and show example usage. But I'll do that as part of a separate PR (maybe one tackling #2016) so that we don't keep this open for too much longer and accumulate more conflicts.

freddyaboulton:
@abidlabs Good news I just did some benchmarking on batching and I think the results are favorable!

## What I did
I created this space: https://huggingface.co/spaces/gradio/queue-batch-benchmark which is a copy of our existing queue benchmark but I added a batching. I also created this space: https://huggingface.co/spaces/gradio/queue-batch-benchmark-max-size-50 which uses a max_batch_size=20 (I titled the space incorrectly lol) for each function.

## Results

### Main branch (no batching)
| Input type | Average time to complete prediction over 1000 requests |
| ------------- | ------------- |
| Audio  | 1.5452740753398222 |
| Image  | 1.5123367877716714 |
| Text | 1.5666406706197937 |
| Video |2.029747503848115|

### This PR (batch_size 4)
| Input type | Average time to complete prediction over 1000 requests |
| ------------- | ------------- |
| Audio  |1.165496392960244 |
| Image  |0.8552168062177755 |
| Text |0.841041189251524 |
| Video |2.532181569765199|

### This PR (max_batch_size 20)
| Input type | Average time to complete prediction over 1000 requests |
| ------------- | ------------- |
| Audio  |1.2939001665276997|
| Image  |0.9907303559825795|
| Text |0.993362974653057|
| Video |3.0297954572785284|

I think it's interesting that a higher max_batch_size did not uniformly bring down the latency.

I tried again, only sending requests for the image function:

| Max batch size | Average time to complete prediction over 1000 requests |
| ------------- | ------------- |
|4  |0.9002170345783234|
|20 |1.3551465120315551|

Using a batch_size of 20 is still faster than no batching but not faster than a batch size of 4 🤔 

I think this is evidence that we can optimize the batch logic (maybe gathering data for each event in the batch is adding overhead?) but we can do so in a future PR. IMO this can go out and we can get user feedback.

## Other thing
gr.Blocks.load does not work if the space using batching. I tried with the queue benchmark and you'll see that the websocket returns "['foo']" and then we take the first element of it to return "[". 

image

Other than that (and the comments I already made) I think this is good to merge! Feel free to merge without fixing `gr.Blocks.load` and we can do that in a separate PR.

abidlabs:
Amaaazing thank you so much @freddyaboulton for the benchmarking! Have a few meetings, but then I'll parse through this more fully and address the comments above

freddyaboulton:
Also, just noticed that a demo without batching built off this PR is showing higher latency than the same demo built off main.

` python benchmark_script_gather.py wss://spaces.huggingface.tech/gradio/queue-benchmark/queue/join --batch_size 50`
```
{'fn_index': ['audio', 'image', 'text', 'video'], 'duration': [1.724048194614982, 1.6302076853891285, 1.5274697332470506, 2.1359671726822853]}
```

`python benchmark_script_gather.py wss://spaces.huggingface.tech/gradio/queue-benchmark-batch-branch/queue/join --batch_size 50`

```
{'fn_index': ['audio', 'image', 'text', 'video'], 'duration': [2.0668310448826555, 1.888936443934365, 1.9982213218865446, 2.88502193372184]}
```

abidlabs:
> Also, just noticed that a demo without batching built off this PR is showing higher latency than the same demo built off main. Would be good to look into this as well!

How are you running `--batch_size 50` with the demo built off main?

freddyaboulton:
My bad for confusing terminology. There are two batches here: —batch_size 50 means send 50 requests to the space at once. The other batch size is max_batch_size in the event listener. I’ll edit my previous comment to make that clearer.

freddyaboulton:
I’m also going to run the two benchmark apps locally (not hosted on spaces) and run the benchmark to make sure it’s not something weird with my internet connection.

abidlabs:
Thanks @freddyaboulton this is super helpful

freddyaboulton:
Actually disregard that last comment about not-batching being slower in this branch. I can't reproduce locally (while I still reproduce the fact batching is faster than not batching). I did a factory reset on https://huggingface.co/spaces/gradio/queue-benchmark-batch-branch and the times are now pretty much in line with https://huggingface.co/spaces/gradio/queue-benchmark 🤷‍♂️ . False alarm I think.

gradio/queue.py
FarukOzderim: @abidlabs I think you don't take into remove overhead into account.

Here is the time comparison between these two versions, there is ~50x speedup in my runs.
This would unnecessarily slow down demos with short inference times significantly. But would not affect long inference times that much. Since there is no loss but some gain in the second version I recommend going with V2.

Time of the V1: 0.0877530574798584
Time of the V2: 0.0018298625946044922
Speedup: 47.95609120521173


```

import time

queue = [1, 2, 3, 4, 5] * 1000

start = time.time()

while queue:
    element = queue.pop()
    rest_of_batch = [e for e in queue if e == element][:5]
    [queue.remove(e) for e in rest_of_batch]

end = time.time()
v1_time = end-start

print(f"Time of the V1: {v1_time}")

queue = [1, 2, 3, 4, 5] * 1000
batch_wait_list = [[] for i in range(5)]

start = time.time()

while queue:
    element = queue.pop()
    batch_wait_list[element-1].append(element)
    if len(batch_wait_list[element-1]) == 5:
        batch_wait_list[element-1] = []
end = time.time()
v2_time = end-start
print(f"Time of the V2: {v2_time}")
print(f"Speedup: {v1_time/v2_time}")

```

gradio/blocks.py
```diff
@@ -668,23 +639,64 @@ def render(self):
...
-            for i, input_id in enumerate(dependency["inputs"]):
-                block = self.blocks[input_id]
-                if getattr(block, "stateful", False):
-                    processed_input.append(state.get(input_id))
-                else:
-                    processed_input.append(block.preprocess(raw_input[i]))
-        else:
-            processed_input = raw_input
-        return processed_input
+        if inspect.isasyncgenfunction(block_fn.fn):
```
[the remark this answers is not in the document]
abidlabs: We do not, which is why the function returns `False`. Or did you mean something else?

gradio/dataclasses.py
```diff
@@ -0,0 +1,15 @@
+from typing import Any, List, Optional
+
+from pydantic import BaseModel
+
+
+class PredictBody(BaseModel):
+    session_hash: Optional[str]
+    data: List[Any]
+    fn_index: Optional[int]
+    batched: Optional[bool] = False
```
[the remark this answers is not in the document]
abidlabs: Actually it's `True` if it's called from the queue with a batched of data (see line 256 in the new queue.py). I'll add a comment to explain the point of this parameter

gradio/routes.py
abidlabs: See above

gradio/blocks.py
abidlabs: Added error and tests to confirm the error is called

gradio/queue.py
abidlabs: Hey @FarukOzderim thanks for providing this profiling code. Agreed I misunderstood the part about `.remove()` taking an unnecessary amount of time. I still feel like the absolute difference is not great (we're talking about 80 ms to clear a queue of 5,000 events) and I'd like to get this PR in without significant additional refactoring, so let me merge this in as is, but I agree the way you propose would be better, so I'll revisit in a future PR (or feel free to make a PR and I'll review & merge)

abidlabs:
All righty, this is finally ready to be merged. I've added documentation about how to use batched functions as well. There are further optimizations that can be added, like @FarukOzderim pointed out, but I'll save that for a future PR (probably when I write the Guide on batched functions and concurrency)

@freddyaboulton I didn't understand your question about `asyncgenfunctions`, but otherwise everything else has been addressed.

I'll merge in in a couple of hours -- thanks all!

Issue #2217, opened 2022-09-09
`gr.HighlightedText()` problems when `entities` not sorted by start character


oconnoob:
### Describe the bug

# Problem (N.B. Temporary solution below)
If the `entities` of a `gr.HighlightedText()` component are not sorted by start, you get a repeating text issue

```python
d = {
  'text': 'Four score and seven years ago our fathers brought forth on this continent a new nation conceived in liberty and dedicated to the proposition that all men are created equal.\n\nFour score and seven years ago our fathers brought forth on this continent a new nation conceived in liberty and dedicated to the proposition that all men are created equal.',
  'entities': [
    {
      'entity': 1,
      'start': 101,
      'end': 108
    },
    {
      'entity': 1,
      'start': 276,
      'end': 283
    },
    {
      'entity': 1,
      'start': 77,
      'end': 87
    },
    {
      'entity': 1,
      'start': 252,
      'end': 262
    }
  ]
}

with gr.Blocks(css="body {background-color: red}") as demo:
    gr.HighlightedText(d)
    
demo.launch()
```
# Output:
![image](https://user-images.githubusercontent.com/46252299/189216487-c0e372bb-9875-488e-86c4-77b02ea939d9.png)

# Temporary Solution
Apparently if the starting characters of the entities are out of order, this issue will happen. Upon rearranging the elements of the `'entities'` list using `d['entities'] = sorted(d['entities'], key= lambda x: x['start'])` I get the appropriate output.

# Remaining Issues - Line Breaks
In addition, the line breaks present in the text are not rendered by the `HighlightedText()` component.

Thanks!

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```python
d = {
  'text': 'Four score and seven years ago our fathers brought forth on this continent a new nation conceived in liberty and dedicated to the proposition that all men are created equal.\n\nFour score and seven years ago our fathers brought forth on this continent a new nation conceived in liberty and dedicated to the proposition that all men are created equal.',
  'entities': [
    {
      'entity': 1,
      'start': 101,
      'end': 108
    },
    {
      'entity': 1,
      'start': 276,
      'end': 283
    },
    {
      'entity': 1,
      'start': 77,
      'end': 87
    },
    {
      'entity': 1,
      'start': 252,
      'end': 262
    }
  ]
}

with gr.Blocks(css="body {background-color: red}") as demo:
    gr.HighlightedText(d)
    
demo.launch()
```

### Logs

```shell
N/A
[...]

Issue #559, opened 2022-02-08
Nested spans with`HighlightedText`


JulesBelveze:
**Is your feature request related to a problem? Please describe.**

Hey all,
I am trying to build a Gradio app with a named entity recognition model which supports nested spans (e.g. in "JFK Street" the model would return "JFK" as a person and "JFK Street" as a location). However (to the best of my knowledge) this isn't currently possible when using `HighlightedText` as it requires outputs to be of the following type: `List[Tuple[str, Union[float, str]]]`.

**Describe the solution you'd like**

It would be great that tokens can have multiple labels. An idea could be to allow outputs of the following type: `List[Tuple[str, List[Union[float, str]]]]`

Thanks for the super cool library 😺

abidlabs:
Hi! We've now made it possible for Gradio users to create their own custom components -- meaning that you can write some Python and JavaScript (Svelte), and publish it as a Gradio component. You can use it in your own Gradio apps, or share it so that anyone can use it in their Gradio apps. Here are some examples of custom Gradio components:

* A "Rich Textbox" that [allows you to write bold/italics/colored text](https://huggingface.co/spaces/abidlabs/gradio_rich_textbox)
* A "Folium Map Viewer" component that [allows you to use interactive maps](https://huggingface.co/spaces/freddyaboulton/gradio_folium)

You can see the source code for those components by clicking the "Files" icon and then clicking "src". The complete source code for the backend and frontend is visible. **In particular, its very fast if you want to build off an existing component**. We've put together a Guide: https://www.gradio.app/guides/five-minute-guide, and we're happy to help if you'd like to put together a custom component for this issue

PR #2221, opened 2022-09-09
Fix `HighlightedText` component

abidlabs:
Quick fix for #2217 and a corresponding test


aliabid94 (approved):
LGTM. I didn't know about this alternate API of returning entity sections, it is glitchy if there are overlapping entity regions. perhaps worth revisiting at some point how to handle those cases.

abidlabs:
That's true (and why I added the comment in the docstring about why intervals should not overlap). There's actually a separate issue open about how to deal with overlapping intervals (#559) so we should think about that at some point (not high priority imo)

PR #2226, opened 2022-09-10
Update package metadata

### LINKED ISSUE #1662: Use static config file to configure setuptools as opposed to setup.py

As a gradio developer, I think we should adopt the best practice put forth in [PEP 621](https://peps.python.org/pep-0621/) and declare project metadata statically in a pyproject.toml (setup.cfg would be ok too). 

We should still keep setup.py for editable installs. 

Use pyproject.toml (or setup.cfg) to define project metadata

**Additional context**  
[PEP 621](https://peps.python.org/pep-0621/)

ofek:
## Description

Fixes #1662 

Hello there! The Python packaging ecosystem has standardized on the interface for build backends ([PEP 517](https://peps.python.org/pep-0517/)/[PEP 660](https://peps.python.org/pep-0660/)) and the format for metadata declaration ([PEP 621](https://peps.python.org/pep-0621/)/[PEP 631](https://peps.python.org/pep-0631/)). As a result, the execution of `setup.py` files is now [deprecated](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html).

So, I'm spending my free time updating important projects so that they are modernized and set an example for others 😄

## Summary of changes

This implements [PEP 621](https://peps.python.org/pep-0621/), obviating the need for `setup.py` and `MANIFEST.in`. The build backend `hatchling` (of which I am a maintainer in the [PyPA](https://github.com/pypa)) is now used as that is the default in the official [Python packaging tutorial](https://packaging.python.org/en/latest/tutorials/packaging-projects/). Hatchling is available on all the major distribution channels such as [Debian](https://salsa.debian.org/python-team/packages/hatchling), [Fedora](https://src.fedoraproject.org/rpms/python-hatchling), [Arch Linux](https://archlinux.org/packages/community/any/python-hatchling/), [conda-forge](https://anaconda.org/conda-forge/hatchling), [Nixpkgs](https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/python-modules/hatchling/default.nix), [Alpine Linux](https://github.com/alpinelinux/aports/blob/master/community/py3-hatchling/APKBUILD), [FreeBSD](https://github.com/freebsd/freebsd-ports/blob/main/devel/py-hatchling/Makefile)/[OpenBSD](https://github.com/openbsd/ports/blob/master/devel/py-hatchling/Makefile), [Gentoo Linux](https://github.com/gentoo/gentoo/tree/master/dev-python/hatchling), [MacPorts](https://github.com/macports/macports-ports/blob/master/python/py-hatchling/Portfile), [OpenEmbedded](https://github.com/openembedded/openembedded-core/commit/846e806181f1349be29cbce78c5041735dfd7e6f), [Spack](https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/py-hatchling/package.py), [MSYS2](https://github.com/msys2/MINGW-packages/blob/master/mingw-w64-python-hatchling/PKGBUILD), etc.

## Notes

- All readme image paths are fixed (see current https://pypi.org/project/gradio/)
- The source distributions on PyPI are erroneously shipping a build artifact `*.egg-info` from `python setup.py develop`; this is now fixed
- The binary distributions (wheels) on PyPI are unintentionally shipping the `test` directory ever since https://github.com/gradio-app/gradio/pull/1817 which clashes with other installed packages; this is now fixed and the wheel is ~1mb smaller

## Future

- We could ship optional wheels compiled with Mypyc, as Black [recently did](https://ichard26.github.io/blog/2022/05/31/compiling-black-with-mypyc-part-1/)

# Checklist:


.circleci/config.yml
```diff
@@ -24,10 +24,14 @@ jobs:
           command: |
             sudo apt update
             sudo apt install ffmpeg -y
+      - run:
+          name: Upgrade virtual environment tools
+          command: |
+            python -m pip install --upgrade pip virtualenv
```
ofek, line 30: PEP 660 (`pip install -e`) landed in https://pip.pypa.io/en/stable/news/#v21-3

abidlabs:
Thank you @ofek for the contribution. To be honest, I am not very familiar with `hatchling` so I will need to look into this and make sure this does not break any of our ci/deployment scripts. One question for you -- will this PR still allow us editable installs (`pip install -e .`) if `setup.py` is removed? 

What are your thoughts @freddyaboulton @aliabid94 @dawoodkhan82?

ofek:
> will this PR still allow us editable installs (`pip install -e .`) if `setup.py` is removed?

Yup! Since [v21-3](https://pip.pypa.io/en/stable/news/#v21-3)

freddyaboulton:
Thank you for the contribution @ofek ! I had filed an issue to track work similar to this https://github.com/gradio-app/gradio/issues/1662 and I'm glad you beat us to it.

I don't know that much about hatchling either - I was thinking we'd use setuptools when I filed #1662 - but I'll check this out this week so I can see how it works and if this PR accidentally breaks anything in our current build process.

ofek:
Happy to help! Please let me know any feedback you may have for Hatchling. As I mentioned to someone else earlier today, some of the discussion about making that the new default starts here https://github.com/pypa/packaging.python.org/pull/1031#issuecomment-1124977628

I'll quote Henry who expressed the common sentiment:

> Which backend should be the "only" backend? `tools.setuptools` is still experimental and there are a lot of older pre-PEP 621 copies of it around, and it's large and full of legacy cruft still, causing slower builds and harder error messages - it can be powerful, but it seems to be an overkill for newcomers and simple Python only projects. Flit is lite and fast, with simple error messages, but a crippled file include/exclude system that won't include `LICENSE.md` by default but will include `__pycache__` folders. Hatchling seems to be a nice balance, with reasonable defaults & nice error messages. For a nice newcomer experience, I think Hatchling is likely best.

Essentially. I did my best to learn the best and worst parts of other backends in order to implement Hatching, which is why it Just Works ™️ while at the same time being simpler to configure and easier to extend.

pyproject.toml
```diff
@@ -0,0 +1,55 @@
+[build-system]
```
freddyaboulton, line 1: @ofek There's a `gradio/templates` directory that's created when the front-end code is built. We need to include it in the source and wheel distribution. Can you please make the change? Thank you!

ofek: Done

freddyaboulton:
Thanks @ofek ! 😄 

@abidlabs I tested it out today and I think everything is working. I uploaded version 3.3b1 so you can test it out. See it [here](https://pypi.org/project/gradio/3.3b1/). This is what I tested:

- [ ] Editable install works as expected
- [ ] `gradio` command still works
- [ ] `%blocks` jupyter cell magic still works. See the colab below
- [ ] Running on spaces still spaces works. See [here](https://huggingface.co/spaces/gradio-pr-deploys/pr-2252-all-demos)
- [ ] Running on colab still works. See [here](https://colab.research.google.com/drive/11_P6t7XPwf8OSF1AJSIUl4CuS8jeRc08#scrollTo=vo5_olatsyC1).
- [ ] The docs can still be built and the demos work as expected
image
- [ ] Changes to the front-end show up in the editable install.


Anything else we should test? Otherwise looks good to me. CC @aliabd @aliabid94

aliabid94:
LGTM, thanks @ofek! thanks for testing @freddyaboulton

abidlabs:
Awesome @freddyaboulton! Very comprehensive testing, LGTM

freddyaboulton (approved):

PR #2234, opened 2022-09-12
Bugfix: null dereference resizing empty textbox

cobryan05:
Fixes "TypeError: Cannot read properties of null (reading 'trim')" when resizing an empty textbox. 'value' is an Optional parameter so must be null checked before empty checked.

This causes https://github.com/sd-webui/stable-diffusion-webui/issues/656 . It prevents the UI from appearing at all if the input prompt is defaulted to multiline. The fix is a simple null check.

# Checklist:

- [x ] I have performed a self-review of my own code
- [ x] My code follows the style guidelines of this project
- [x ] New and existing unit tests pass locally with my changes

### DIFF
--- a/ui/packages/form/src/Textbox.svelte
+++ b/ui/packages/form/src/Textbox.svelte
@@ -77,7 +77,7 @@
 		el.style.overflowY = "scroll";
 		el.addEventListener("input", resize);
 
-		if (!value.trim()) return;
+		if (!value || !value.trim()) return;
 		resize({ target: el });
 
 		return {


cobryan05:
*bump*

cobryan05:
Reproducing was even simpler than I thought 😆 .
```
import gradio as gr

with gr.Blocks() as demo:
    box = gr.Textbox(value=None)
demo.launch()
```

Not sure exactly how 'None' is getting there in webui, since it should be being set to '' ... but TextBox's value *is* marked as 

value: Optional[str | Callable] = "",

so it shouldn't crash if 'None' is in there

abidlabs:
We can always map null to "" in the backend for `Textbox`, but I think given that other components like `Number` and `Image` support being null, I think `Textbox` should as well @pngwn?

abidlabs:
Thank you for the PR @cobryan05. We discussed this and figured it's better to fix this via the backend. We've merged in a fix here: #2264, so this should now work on `main`. We'll also release a version to pypi by the end of the week.

Issue #2195, opened 2022-09-07
Direct support for partial functions


Sreerag-ibtl:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
Currently, partial functions can only be used when the object is patched with its __ name __ attribute. Support for partial functions enables one to use functions with more arguments than the number of input components specified.

**Describe the solution you'd like**  
I do not have a clear picture. But I found that [__ name __](https://github.com/gradio-app/gradio/blob/5360ea002ea79ca6016ff61842317f2227570015/gradio/interface.py#L291) is only available for function objects, not partial functions. Maybe that's the only blocking thing here.

**Additional context**  
I currently use a partial function by explicitly setting partial_function.__ name __.

abidlabs:
Hi @Sreerag-ibtl can you please provide a full code example with partial functions so we can understand what you'd like to support?

Sreerag-ibtl:
@abidlabs Here's a [colab](https://colab.research.google.com/drive/1Ilg-wRVcZKavM6uHicHvs2mMcxWUX8Tj?usp=sharing) notebook for your reference.

PR #2235, opened 2022-09-12
Quick fix for partial functions

abidlabs:
Partial functions don't have the `__name__` attribute, so were causing a problem with Interface/Blocks. This fixes: #2195 and adds a test to prevent regressions in the future.

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -248,7 +248,9 @@ class BlockFunction:
     def __str__(self):
         return str(
             {
-                "fn": self.fn.__name__ if self.fn is not None else None,
+                "fn": getattr(self.fn, "__name__", "fn")
+                if self.fn is not None
+                else None,
                 "preprocess": self.preprocess,
                 "postprocess": self.postprocess,
             }
--- a/gradio/interface.py
+++ b/gradio/interface.py
@@ -288,7 +288,7 @@ class Interface(Blocks):
         self.api_mode = _api_mode
         self.fn = fn
         self.fn_durations = [0, 0]
-        self.__name__ = fn.__name__
+        self.__name__ = getattr(fn, "__name__", "fn")
         self.live = live
         self.title = title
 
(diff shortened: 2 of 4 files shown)


freddyaboulton (approved):
Looks good to me @abidlabs ! I might also check the `fn` name in the config is set to `fn`

dawoodkhan82 (approved):

cobryan05:
I was running into an issue last night with partial functions and integrating the new iterative output.  My partial async functions were failing the `inspect.iscoroutinefunction` calls in blocks.py, since the 'iscouroutinefunction' wasn't unwrapping the partial. Is there any chance this change might also fix that issue?

freddyaboulton:
@cobryan05 You have an async generator defined in your app? The iterative output feature [does not support](https://github.com/gradio-app/gradio/blob/main/gradio/blocks.py#L631) async generators (rationale [here](https://github.com/gradio-app/gradio/pull/2162#issuecomment-1237495624)). I can look into it but right now the expectation is for async generators to not work regardless of whether they are partial functions or not.

cobryan05:
> @cobryan05 You have an async generator defined in your app?
I'm probably getting terminology mixed up. I have a function that yields.  

I tried lots of combinations of 'async' and type hints and such while trying to get it to work before I figured out it was the partial.

But basically, I have a function that has a yield statement, but since I have it wrapped in 'partial' the 'inspect' routines don't think it is going to be a function that yields. 

I just saw 'partial' in the name here and thought "Hey, I was just dealing with an issue like that, I wonder if their fix is the 'better' fix for the issue" :)

cobryan05:
Awesome, sounds like exactly what I needed! Will try it out tonight :)

abidlabs:
Thanks a bunch for the reviews @freddyaboulton and @dawoodkhan82! I've added a test to check the config and will merge this PR once the CI finishes running.

PR #2242, opened 2022-09-12
Fix postprocess when value is a dict with only one key

### LINKED ISSUE #2191: Returning a dict from a function doesn't work as expected when the dict only has one key
### Describe the bug

When you return a dictionary from a function, the component may not be updated as expected if the dictionary has one key.

This is because the dict will get converted into a tuple of lists as opposed to a list with one element.

The following code handles dict outputs and you can see that it converts the predictions into a list

```python
        if type(predictions) is dict and len(predictions) > 0:
            keys_are_blocks = [isinstance(key, Block) for key in predictions.keys()]
            if all(keys_are_blocks):
                reordered_predictions = [skip() for _ in dependency["outputs"]]
                for component, value in predictions.items():
                    if component._id not in dependency["outputs"]:
                        return ValueError(
                            f"Returned component {component} not specified as output of function."
                        )
                    output_index = dependency["outputs"].index(component._id)
                    reordered_predictions[output_index] = value
                predictions = reordered_predictions
```
Then we convert that to a tuple:

```python
        if len(dependency["outputs"]) == 1:
            predictions = (predictions,)
```

### Reproduction

```python
import gradio as gr
import random
import string

with gr.Blocks() as demo:
    num = gr.Number()
    text = gr.Textbox()
    update = gr.Button(value="update")

    def update_values():
        return {text: random.choice(string.ascii_lowercase)}

    update.click(update_values, inputs=None, outputs=[text])

demo.launch()
```    

### Screenshot

You can see that the text displays `['v']` as opposed to just the character "v"

image

### Logs

```shell
-
```

### System Info

```shell
3.2
```

### Severity

annoying

freddyaboulton:
# Description
Fixes #2191 

# Checklist:

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -671,12 +671,7 @@ class Blocks(BlockContext):
                         )
                     output_index = dependency["outputs"].index(component._id)
                     reordered_predictions[output_index] = value
-                # Unwrap list if only one output component to prevent
-                # wrapping the list into a tuple in the if len(reordered_predictions) == 1
-                # check
-                if len(reordered_predictions) == 1:
-                    reordered_predictions = reordered_predictions[0]
-                predictions = reordered_predictions
+                predictions = utils.resolve_singleton(reordered_predictions)
             elif any(keys_are_blocks):
                 raise ValueError(
                     "Returned dictionary included some keys as Components. Either all keys must be Components to assign Component values, or return a List of values to assign output values in order."
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -671,6 +671,11 @@ class Blocks(BlockContext):
                         )
                     output_index = dependency["outputs"].index(component._id)
                     reordered_predictions[output_index] = value
+                # Unwrap list if only one output component to prevent
+                # wrapping the list into a tuple in the if len(reordered_predictions) == 1
+                # check
+                if len(reordered_predictions) == 1:
+                    reordered_predictions = reordered_predictions[0]
                 predictions = reordered_predictions
             elif any(keys_are_blocks):
                 raise ValueError(
--- a/test/test_blocks.py
+++ b/test/test_blocks.py
@@ -287,15 +287,18 @@ def test_blocks_returns_correct_output_dict_single_key():
         update = gr.Button(value="update")
 
         def update_values():
-            return {output: output.update(value=42)}
+            return {num2: gr.Number.update(value=42)}
 
         update.click(update_values, inputs=[num], outputs=[num2])
 
-    output = demo.postprocess_data(0, {num2: num2.update(value=42)}, state=None)
+    output = demo.postprocess_data(0, {num2: gr.Number.update(value=42)}, state=None)
     assert output[0]["value"] == 42
 
     output = demo.postprocess_data(0, {num2: 23}, state=None)
     assert output[0] == 23
+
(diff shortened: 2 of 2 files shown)


test/test_blocks.py
```diff
@@ -279,6 +279,25 @@ def test_blocks_does_not_replace_keyword_literal():
...
 
+def test_blocks_returns_correct_output_dict_single_key():
+
+    with gr.Blocks() as demo:
+        num = gr.Number()
+        num2 = gr.Number()
+        update = gr.Button(value="update")
+
+        def update_values():
+            return {output: output.update(value=42)}
```
abidlabs, line 290: I think this should be as follows:
```suggestion
        def update_values(inp):
            return {num2: gr.update(value=42)}
```

test/test_blocks.py
```diff
@@ -279,6 +279,25 @@ def test_blocks_does_not_replace_keyword_literal():
...
+
+        update.click(update_values, inputs=[num], outputs=[num2])
+
+    output = demo.postprocess_data(0, {num2: num2.update(value=42)}, state=None)
```
abidlabs, line 294:
```suggestion
    output = demo.postprocess_data(0, {num2: gr.Number.update(value=42)}, state=None)
```

gradio/blocks.py
```diff
@@ -671,6 +671,11 @@ def postprocess_data(self, fn_index, predictions, state):
                         )
                     output_index = dependency["outputs"].index(component._id)
                     reordered_predictions[output_index] = value
+                # Unwrap list if only one output component to prevent
+                # wrapping the list into a tuple in the if len(reordered_predictions) == 1
+                # check
```
abidlabs, line 676: I would remove this long inline comment
```suggestion
```

gradio/blocks.py
```diff
@@ -671,6 +671,11 @@ def postprocess_data(self, fn_index, predictions, state):
...
+                if len(reordered_predictions) == 1:
+                    reordered_predictions = reordered_predictions[0]
```
abidlabs, line 678: We've used this utils function in many other places in the code
```suggestion
                    reordered_predictions = utils.resolve_singleton(reordered_predictions)
```

gradio/blocks.py
```diff
@@ -671,6 +671,11 @@ def postprocess_data(self, fn_index, predictions, state):
...
                 predictions = reordered_predictions
```
abidlabs, line 679: We've used this utils function in many other places in the code
```suggestion
                    predictions = utils.resolve_singleton(reordered_predictions)
```

abidlabs:
LGTM, some small suggestions above. Also noticed that returning an empty dictionary `{}` as the output of a function results in an error. Instead, it should have no effect.

We could replace line 662: `if type(predictions) is dict and len(predictions) > 0:`, with the following:

```py
        if type(predictions) is dict:
             if len(predictions) == 0:
                 return []
```

to fix that

abidlabs (approved):

test/test_blocks.py
freddyaboulton: Yep you're right it's technically a class method.  Both work and I sometimes inadvertently use it as an instance method cause of the faster autocomplete. Will make the change!

freddyaboulton:
@abidlabs If we make that change, then you can't return an empty dictionary for JSON output. I'll keep that part of the code as is for now.

Issue #2246, opened 2022-09-12
visible attribute has no effect on gradio.Group()


shahbuland:
### Describe the bug

- Setting visible = false in gr.Group (i.e. with gr.Group(visible = False)) does nothing
- Subbing in gr.Column instead of gr.Group works as expected with regards to visibility
- gr.update(visible = False) also has no effect on gr.Group

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```python
import gradio as gr

with gr.Blocks() as demo:
    with gr.Group(visible = True):
        gr.Textbox("A")
    with gr.Group(visible = False):
        gr.Textbox("B")
```

### Screenshot

[Running the demo provided in reproduction code](https://cdn.discordapp.com/attachments/738882678711123980/1019013634636124210/unknown.png)

### Logs

```shell
No errors so no logs
```

### System Info

```shell
Browser = Chrome
gradio version = 3.3
```

### Severity

serious, but I can work around it

aliabid94:
will take a look!

lcrh:
The buggy behavior has reappeared as of '3.34.0'

![image](https://github.com/gradio-app/gradio/assets/14020730/5bb8e6b0-30a4-4451-9d7c-6322dbddb022)

taesiri:
Any update on this?

taesiri:
We have a fix for this here #4567

PR #2247, opened 2022-09-13
Attach visible kwarg to all components

aliabid94:
There were some components missing a `visible=` argument. Attached to all now.

Fixes: #2246


gradio/layouts.py
```diff
@@ -128,15 +132,19 @@ class Tabs(BlockContext):
 
     def __init__(
         self,
+        *,
         selected: Optional[int | str] = None,
+        visible: bool = True,
```
freddyaboulton, line 137: I think we also need to modify the update method of Tabs class so that users can update the visibility in the demo

gradio/layouts.py
```diff
@@ -167,15 +175,20 @@ class TabItem(BlockContext):
     def __init__(
         self,
         label: str,
+        *,
         id: Optional[int | str] = None,
+        visible: bool = True,
```
freddyaboulton, line 180: Why do we need `visible` for the TabItem, isn't that what `selected` in the `Tabs` class is for?

freddyaboulton (commented):
@aliabid94 Thanks for the PR! I think this looks good although I'm confused as to why we're adding `visible` to the individual tabs. In any event, might be good to write some unit tests for this change of behavior. 

To check that the visibility of all components can be controlled in the constructor, I'm thinking we can do this:

```python
def test_all_components_visible(io_components):
    component_classes = [component(visible=False) for component in io_components]
    component_classes += [gr.Row(visible=False), gr.Column(visible=False), gr.Group(visible=False),
                          gr.Tabs(visible=False), gr.Box(visible=False), gr.Accordion(label="foo", visible=False),
                          gr.TabItem(label="tab", visible=False)]
    for component in component_classes:
        assert component.get_config().get("visible") is False
```

Might also be good to write a demo/unit test that tests whether the visibility of the layout components can be updated as well.

For example, in the demo below, the `show_tab` button won't work because you can't control the visibility of tabs via update.

```python
import gradio as gr

with gr.Blocks() as demo:
    with gr.Row(visible=False) as row:
        with gr.Tabs(visible=False) as tabs:
            box = gr.Box(visible=False)
            accordion = gr.Accordion(label="algorithm", visible=False)
            with gr.Group(visible=False) as group:
                gr.Textbox(value="Visible!")


    show_row = gr.Button(value="Show row")
    show_row.click(lambda: row.update(visible=True), inputs=None, outputs=row)
    show_tab = gr.Button(value="Show tab")
    show_tab.click(lambda: tabs.update(visisble=True), inputs=None, outputs=tabs)
    show_contents = gr.Button(value="Show contents")
    show_contents.click(lambda: [box.update(visible=True), accordion.update(visible=True), group.update(visible=True)],
                        inputs=None, outputs=[box, accordion, group])

demo.launch()
```

gradio/layouts.py
freddyaboulton: Hmm on second thought, I think I can see a case for needing to show an additional tab based on input defined elsewhere in the app!

aliabid94:
Thanks for the feedback @freddyaboulton! Removed visible from tabitem, it was a little confusing on how to handle that.

PR #2249, opened 2022-09-13
Delete spaces created on PRs after a certain amount of time

### LINKED ISSUE #2245: Delete PR Spaces automatically

We should delete the spaces created automatically during PR reviews after the PR is merged or after some given amount of time has passed after the PR has been merged. Keeping around the spaces post merge for a bit may help keep a history of the demos in case we realize something is broken.

freddyaboulton:
# Description

Fixes #2245

# Checklist:

### DIFF
--- a/scripts/delete_old_spaces.py
+++ b/scripts/delete_old_spaces.py
@@ -41,7 +41,7 @@ if __name__ == "__main__":
     args = parser.parse_args()
     api = HfApi()
 
-    to_delete = get_spaces_to_delete(args.n_days, args.org_name)
+    to_delete = get_spaces_to_delete(args.n_days, args.org_name, api_client=api)
     for space in to_delete:
         print(f"Deleting {space}")
         delete_space(space, args.hf_token, api_client=api)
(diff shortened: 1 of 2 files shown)


scripts/delete_old_spaces.py
```diff
@@ -0,0 +1,47 @@
...
+
+def delete_space(space_id: str, hf_token: str, api_client: Optional[HfApi] = None):
+    api_client = api_client or HfApi()
+    api_client.delete_repo(repo_id=space_id, token=hf_token, repo_type="space")
+
+
+def get_spaces_to_delete(
+    n_days: int, org_name: str, api_client: Optional[HfApi] = None
+):
+    api_client = api_client or HfApi()
```
abidlabs, line 16: Out of curiosity, why pass in an `api_client` as a parameter like this? We can just define a single instance of `HfApi()` outside of the function?

abidlabs (approved):

abidlabs:
LGTM @freddyaboulton! I haven't tested the code, but I did go through and inspect it and it looks ✅

scripts/delete_old_spaces.py
freddyaboulton: I think it makes the function more general in the event we want to reuse them in other scripts in the future. We do currently define a single instance of api_client and pass it to all the functions.

freddyaboulton:

PR #2250, opened 2022-09-13
Show view api button even if there are no named api routes in the app

### LINKED ISSUE #2160: Show "view API" button with a helpful message even with no named API routes
It has been pretty confusing for our users when they see some Gradio demos with the "view API" link and others without it. The reason this happens is that when there are no named API routes, we do not show the "view API" link. This is the case in Blocks (by default if there is no `api_name` attached to any event listener), while in Interface, there is always at least one named API.

To avoid this confusion, I would propose that we always show the "view API" link, but if there is no named API then clicking on the "view API" link includes a message that says "this demo has no named API routes" and link to the documentation for named API routes.

Related discussion: https://huggingface.slack.com/archives/C02QNREU24S/p1662068541775709

freddyaboulton:
# Description

Fixes #2160

### To test

- [x] Case of no named api routes: run `demo/calculator_blocks/run.py`

Should see the "To add api docs" page

image

- [x] Case of named api routes: run modify `demo/calculator_blocks/run.py`   and add an api_name to `submit_btn.click` 

Should see the actual api docs

image

- [x] Case of suppressing the api page when there are named api routes: Set `show_api` to False in the above modified `calculator_blocks` demo

image

You can also do this with an interface demo like `calculator`

# Checklist:

### DIFF
--- a/ui/packages/app/src/ApiDocs.svelte
+++ b/ui/packages/app/src/ApiDocs.svelte
@@ -16,99 +16,121 @@
 </script>
 
 <div>
-	<h2 class="text-3xl text-center mb-6">
-		API Docs for
-		<span class="italic text-amber-500">
-			{root}
-		</span>
-	</h2>
-	<div class="flex flex-col gap-6">
-		{#each dependencies as dependency, d}
-			{#if dependency.documentation}
-				<div
-					class="bg-gray-50 border border-gray-100 dark:bg-gray-800 dark:border-gray-700 p-6 rounded"
-				>
-					<h3 class="text-3xl text-amber-500 font-semibold mb-2">
-						POST /api/{dependency.api_name}
-					</h3>
-					<div class="mb-6">
-						Full URL: <span class="underline"
-							>{root}api/{dependency.api_name}</span
-						>
-						<button
-							class="ml-1 px-2 py-0.5 rounded bg-gray-200 dark:bg-gray-700"
-							on:click={() => {
-								navigator.clipboard.writeText(
-									root + "api/" + dependency.api_name
-								);
-								just_copied = d;
-								setTimeout(() => {
-									just_copied = -1;
-								}, 500);
-							}}
-						>
-							{#if just_copied === d}copied!{:else}copy{/if}
-						</button>
-					</div>
-					<h4 class="text-2xl mt-6 mb-4">Input Payload</h4>
+	{#if dependencies.some((d) => "documentation" in d)}
+		<h2 class="text-3xl text-center mb-6">
+			API Docs for
+			<span class="italic text-amber-500">
+				{root}
+			</span>
+		</h2>
+		<div class="flex flex-col gap-6">
+			{#each dependencies as dependency, d}
+				{#if dependency.documentation}
 					<div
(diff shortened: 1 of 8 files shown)


ui/packages/app/src/ApiDocs.svelte
```diff
@@ -16,99 +16,121 @@
...
 					</div>
-				</div>
-			{/if}
-		{/each}
-	</div>
+				{/if}
+			{/each}
+		</div>
+	{:else}
+		<h2 class="text-3xl text-center mb-6">
```
freddyaboulton, line 117: Big diff but I just added an if else, where the else clause is the case of no named api routes.

ui/packages/app/src/ApiDocs.svelte
```diff
@@ -16,99 +16,121 @@
...
+			There are no named API Routes for
+			<span class="italic text-amber-500">
+				{root}
+			</span>
+		</h2>
+		<div>
+			To expose an api endpoint of your app in the documentation, set the <span
```
freddyaboulton, line 124: Happy to change the text/styling as well

freddyaboulton:
Thanks a really good point @aliabid94 ! 

I wanted to target the "empty state" of the api docs at developers since we get lots of questions from developers asking why the api docs are "missing" from their app. 

That being said, maybe it will be better if developers can disable the documentation via a parameter of the Blocks `launch` method. That way, when they see the "empty state" api docs page, they can decide to either name some of the end points or disable the documentation entirely.

abidlabs:
Agreed on adding a `show_api` parameter to `launch()` to allow developers to hide this page if they want. 

For the message itself, I think we should point developers to the `api_name` parameter in the Guides: https://gradio.app/sharing_your_app/#api-page

So I would suggest the message to read:

--

To expose an API endpoint of your app in these API docs, set the _api_name_ parameter of the event listener. For more information, see the ["API Page" section in the Guides](https://gradio.app/sharing_your_app/#api-page). To hide this API page, set  to _show_api=False_ in the `.launch()` method.

ui/packages/app/src/Blocks.svelte
```diff
@@ -35,6 +35,7 @@
 	export let target: HTMLElement;
 	export let id: number = 0;
 	export let autoscroll: boolean = false;
+	export let show_api: boolean = true;
```
freddyaboulton, line 38: I think this is backwards compatible. I tested by manually removing `show_api` from the config and this parameter is still set to true. I think that covers us in the case you try to load a config from a demo created in a previous version that doesn't have `show_api`. 

Does this sound right to you @aliabid94 ?

freddyaboulton:
@aliabid94 @abidlabs I added the `show_api` parameter to the Blocks.launch method and modified the text for the "no named routes" page as suggested.

This should be good for another look!

ui/packages/app/src/Blocks.svelte
aliabid94: sounds good to me

aliabid94 (approved):
LGTM

abidlabs:
Same LGTM!

PR #2264, opened 2022-09-14
Skip value key in delete_none only during updates

freddyaboulton:
# Description

Fixes issue where None values are not filtered out of the config. Initially we modified `delete_none` in #2157 to not delete the `value` key so that the value of a component could be set to `None` in `gr.update`. However, we should still filter out `None` from the config outside of `gr.update`.

To test, the following demo should work without breaking the `blocks_reset` and `blocks_update` demo on spaces.

```python
import gradio as gr

with gr.Blocks() as demo:
    box = gr.Textbox(value=None)
demo.launch() 
```

Should fix the problem mentioned in #2234

# Checklist:

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -707,7 +707,9 @@ class Blocks(BlockContext):
                             == components._Keywords.NO_VALUE
                         ):
                             prediction_value.pop("value")
-                        prediction_value = delete_none(prediction_value)
+                        prediction_value = delete_none(
+                            prediction_value, skip_value=True
+                        )
                         if "value" in prediction_value:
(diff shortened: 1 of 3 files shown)


abidlabs:
LGTM, thanks @freddyaboulton! 

Confirmed that this works too:

```py
import gradio as gr

with gr.Blocks() as demo:
    box = gr.Textbox(value="hi")
    btn = gr.Button()
    
    btn.click(lambda :gr.Textbox.update(value=None), None, box)
demo.launch() 
```

abidlabs (approved):

freddyaboulton:
Thank you for the review @abidlabs !

PR #2274, opened 2022-09-15
Do not deploy pr to spaces on prs originating from forks

freddyaboulton:
To prevent the job from failing on the PR

### DIFF
--- a/.github/workflows/deploy-pypi.yml
+++ b/.github/workflows/deploy-pypi.yml
@@ -19,7 +19,7 @@ on:
 
 jobs:
   deploy:
-    if: (github.event.pull_request.merged == true || github.event.action == '')
+    if: github.event.pull_request.merged == true || github.event.action == ''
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-pypi.yml
```diff
@@ -19,7 +19,7 @@ on:
 
 jobs:
   deploy:
-    if: github.event.pull_request.merged == true || github.event.action == ''
```
abidlabs, line 22: Was this not working before?

abidlabs:
Very cool, LGTM

abidlabs (approved):

.github/workflows/deploy-pypi.yml
freddyaboulton: It was! I can get rid of the parens, It's a remnant of an earlier commit

Issue #2091, opened 2022-08-26
Add add ability to add captions to image in a Gallery 


abidlabs:
We have a feature request to be able to display some sort of label/caption for each image in a Gallery. Specifically the request is to: "display numbers in a gr.Gallery, to label images? e.g. in my [space](https://huggingface.co/spaces/sasha/PromptBook), I would like to have numbers alongside the grid to help people flag them"

I think this is a good idea as it takes on the most common use case of the `Carousel` component, which was to display images and text together, and we can properly deprecate that component.

hwasiti:
This is a great addition to the gallery component. 
Can you consider adding an argument to make the labels visible/invisible? In some use cases it would be better to not cover the images with the labels and only show the label below the image when clicked and enlarged.

PR #2284, opened 2022-09-16
Add captions to galleries

aliabid94:
Adds option to have caption for gallery images. Return tuples of (img, string caption) instead of images instead. Does break the returned API for gallery but I think that's okay. 

See demo/fake_gan for example.

Fixes: #2091


abidlabs:
Very cool, a LOT of people have asked for this! In fact, we might want to show a preview in the #gradio-discuss channel and get some quick feedback from folks there. 

As far as the PR itself:
* it's not showing up here correctly, do you know why? https://huggingface.co/spaces/gradio-pr-deploys/pr-2284-all-demos
* I really do NOT think we should change the internal API for any component until 4.0. We avoided changing the API for the `Image` component for this very reason. Even though it's an internal API, I believe it will break anyone using `gr.Interface.load()` and loading any demo that has a gallery from an older version of Gradio. Is there any way to implement this so that both APIs are supported?

aliabid94:
Ok I made it so that the frontend still sends the same payload as before if no captions are provided, but if captions are provided then it uses a new API. That should keep gr.Interface.load working right?

aliabid94:
Ok so when the fake_gan demo uses captions, it breaks in the all_demos space, but when I remove them, it works. I'm assuming this has to do with captions using the new API, and the all_demos space using gr.Interface.load? @abidlabs @freddyaboulton.

abidlabs:
But `all_demos` doesn't use `gr.Interface.load()`. 

It does use `.render()`, but I don't think that should be causing the issue: see https://github.com/gradio-app/gradio/blob/99833d506ef88f9452e516ef78db98edae8798f6/demo/all_demos/run.py#L33

abidlabs:
Light mode looks good @aliabid94:

image

Labels are a little hard to read in dark mode though (font should be white):

image

Looking into the `all_demo` issue now

gradio/components.py
```diff
@@ -3264,17 +3264,25 @@ def get_config(self):
             **IOComponent.get_config(self),
         }
 
-    def postprocess(self, y: List[np.ndarray | PIL.Image | str] | None) -> List[str]:
+    def postprocess(
```
freddyaboulton, line 3267: If we're going to add captions to the Gallery, I wonder if we can make it so that users can add captions to individual gr.Image components? 

My mental model is that `gr.Gallery` is a shorthand for `List[gr.Image]`, with the front-end doing all the magic to make sure each is properly displayed. Would be cool if this feature followed that paradigm.

pngwn:
Looks good, definitely a useful features. Two points of feedback:

- Could we do this in a way that doesn't cover the image? Captions typically appear below the image for this reason.
- This isn't currently accessible but captions for images are directly supported by the spec with `figure` `img` and `figcaption`. Accessible markup would look something like this:

```html
<figure>
  <img src="image.png" alt="put image description here" />
  <figcaption>Figure 1: The caption</figcaption>
</figure>
```

Alternatively we could set the caption as the `alt` text of the `img`, since we don't have any, and duplicate it in the markup exactly as we are doing now.

pngwn:
Thinking bout it i think the use of `figure` is probably wrong here and the second accessibility option should be used, this isn't semantically a 'figure'. Setting the `alt` text of the `img` to the provided caption value is probably preferable.

gradio/components.py
abidlabs: Just fwiw It might not be as critical to add captions to images because people can get away with using the `label` for the `Image` component, which they can't do for individual images in a gallery.

freddyaboulton: Fair point but I'm worried that differences between the Gallery and Image api will make our api harder to understand and reason about. Also, adding a caption via a label for a single image would look different than how captions would look for the gallery which might be weird for users.

abidlabs:
Ok so the problem we are seeing in the Spaces demo happens because of example caching. I see the issue and will push a fix

gradio/components.py
abidlabs: I do think we should support captions for Images, but it could be part of a separate PR if needed. One other consideration is that we might want to open up the `Gallery` component for other non-Image components in the  future, specifically `Video` (we've heard requests for this a lot). If we do that, at that point, we could add captions to `Video` as well

freddyaboulton: Sounds good @abidlabs ! I'm ok with doing labels for images later - I'll file an issue to track that

nateraw:
What if the label here is longer, like a text prompt from a text-to-image model? Maybe something like 120 characters?

Will this feature handle texts of that length without any text being hidden/doing anything wonky? Very relevant to something I've been working on -- just spent a couple hours trying to hack a solution around it before seeing this PR 😅 .

aliabid94:
Updated styling of captions as seen below @gary149. 

Recording 2022-09-23 at 02 05 00

> What if the label here is longer, like a text prompt from a text-to-image model?

@nateraw it truncates labels that are too long with ellipses (see first image in gallery above) - however you can hover over an image and will see the full label as alt text. 

Ready to be merged I believe @abidlabs @pngwn

pngwn (approved):
Looks great!

ui/packages/app/src/components/Gallery/Gallery.svelte
```diff
@@ -148,22 +165,31 @@
...
 						>
 							<img
 								alt=""
 								class="h-full w-full overflow-hidden object-contain"
 								src={image}
 							/>
+							{#if caption}
+								<div class="bottom-0 absolute z-[5] flex justify-end w-full">
+									<div
+										class="bg-gray-50 dark:bg-gray-700 dark:text-gray-200 border-t border-l dark:border-gray-600 font-semibold px-3 py-1 rounded-tl-lg group-hover:opacity-50 max-w-full truncate"
```
gary149, line 187: maybe in size "text-sm"?

aliabid94: was already sm, made it xs

gradio/components.py
```diff
@@ -3270,17 +3270,25 @@ def get_config(self):
...
+        self,
+        y: List[np.ndarray | PIL.Image | str]
+        | List[Tuple[np.ndarray | PIL.Image | str], str]
```
abidlabs, line 3276:
```suggestion
        | List[Tuple[np.ndarray | PIL.Image | str, str]]
```

gradio/components.py
```diff
@@ -3270,17 +3270,25 @@ def get_config(self):
...
+        | None,
+    ) -> List[str]:
         """
         Parameters:
-            y: list of images
+            y: list of images, or list of (image, caption) tuples
         Returns:
-            list of base64 url data for images
+            list of [base64 image data, caption] list pairs
```
abidlabs, line 3283:
```suggestion
            list of base64 data or (base64 image data, caption) pairs
```

gradio/components.py
```diff
@@ -3216,7 +3216,7 @@ class Gallery(IOComponent):
     """
     Used to display a list of images as a gallery that can be scrolled through.
     Preprocessing: this component does *not* accept input.
-    Postprocessing: expects a list of images in any format, {List[numpy.array | PIL.Image | str]}, and displays them.
+    Postprocessing: expects a list of images in any format, {List[numpy.array | PIL.Image | str]}, or list of (image, caption) tuples and displays them.
```
abidlabs, line 3219:
```suggestion
    Postprocessing: expects a list of images in any format, {List[numpy.array | PIL.Image | str]}, or a {List} of (image, {str} caption) tuples and displays them.
```

gradio/components.py
```diff
@@ -3323,15 +3334,28 @@ def deserialize(
             return None
         gallery_path = os.path.join(save_dir, str(uuid.uuid4()))
         for img_data in x:
-            ImgSerializable.deserialize(self, img_data, gallery_path)
+            if isinstance(img_data, list) or isinstance(img_data, tuple):
+                img_data, caption = img_data
+                prefix = f"[{utils.make_valid_filename(caption)}]-"
```
abidlabs, line 3339: For consistency, can we use the `str(uuid.uuid4())` approach from line 3335?

gradio/utils.py
```diff
@@ -680,3 +681,8 @@ def validate_url(possible_url: str) -> bool:
 
 def is_update(val):
     return type(val) is dict and "update" in val.get("__type__", "")
+
+
+def make_valid_filename(name):
```
abidlabs, line 686: We already have `utils.strip_invalid_filename_characters()`. Can we use that insetad?

gradio/components.py
abidlabs, line 3339: I don't love this approach. It's a bit flaky and the user loses any ability to input special characters in their caption. I would instead create a `captions.json` in the same folder as the images, and read/write from their.

abidlabs:
@aliabid94 this looks pretty good. I tested passing in:

* images without captions
* images whose caption was `None`
* images with captions
* mix of such images

And in cases, it worked perfectly. I do have some suggestions for the internal implementation, though, see inline comments.

freddyaboulton:
Awesome pr @aliabid94 ! Just flagging some minor points here that I noticed while testing. 

1. Did the behavior of the gallery "click to zoom" change in this pr? Seems like the images are cropped too much when I click on any on the results of `fake_gan`  on this branch vs main.

### this branch
image

### main
Screen Shot 2022-09-23 at 6 30 59 PM


2. This one is a bit more hard to explain so please bear with me lol. Let's say you run fake_gan with `cache_examples=True`. If you click on a gallery with two images and zoom into the second image, you can't click on a row with less than two images. If you do, the whole app freezes. This is what gets put to the console

```
index.994c3b81.js:313 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading '0')
    at Object.p (index.994c3b81.js:313:149)
    at Object.p (index.994c3b81.js:115:16)
    at Object.p (index.994c3b81.js:798:41)
    at update_slot_base (index.dec5a3df.js:126:14)
    at Object.p (Block.4f047baf.js:53:6)
    at Object.p (Block.4f047baf.js:122:21)
    at update (index.dec5a3df.js:615:36)
    at flush (index.dec5a3df.js:582:13)
```

Gif showing behavior: 

fake_gan_freezes

I see the same on spaces right now:
fake_gan_freezes_spaces

3. One other minor point is that the label text on light mode (spaces) is a bit hard to read

Other than that looks good though! I also tried to break it from the python side like @abidlabs and couldn't 😄

gradio/components.py
aliabid94: You're right, I'll store captions in a json. But I'll preserve the captions in the filename too, I think it makes it easier to view

aliabid94: Done.

aliabid94:
Great catches @freddyaboulton! I cropped the images to go above the gallery preview (and label if there is one). What the other problem actually was that if you selected an image, the gallery kept that index selected. If the images changed, it tried to select the image at that same index in the new batch of images. If there were fewer images, the index did not exist and the app broke. Fixed now.
Screen Shot 2022-09-26 at 12 01 45 PM

btw @pngwn I've noticed often that gradio is somewhat brittle, in that if there's any javascript error, the whole app completely stops working. Do you know why that is? all the event listeners also stop working too.

freddyaboulton (approved):
Thanks for the PR @aliabid94 !

gradio/components.py
```diff
@@ -3322,16 +3333,43 @@ def deserialize(
         if x is None:
             return None
         gallery_path = os.path.join(save_dir, str(uuid.uuid4()))
+        captions = {}
         for img_data in x:
-            ImgSerializable.deserialize(self, img_data, gallery_path)
+            if isinstance(img_data, list) or isinstance(img_data, tuple):
+                img_data, caption = img_data
+                prefix = f"[{utils.strip_invalid_filename_characters(caption)}]-"
```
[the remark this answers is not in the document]
aliabid94: It’s not needed but I think it helps the app creator keep track of flagged files more easily through the filename as well. I think it’s worth keeping

hwasiti:
Can we give a choice to show labels only when an image is clicked and expanded? this in a way labels don't cover the images.

PR #2293, opened 2022-09-19
Promotes `postprocess` and `preprocess` to documented parameters

abidlabs:
Makes `postprocess` and `preprocess` official and documented parameters for event triggers, since we've started to see questions that could be resolved by using these parameters (e.g. see #2283)

Also does the following:
* Noticed that some event triggers were missing some parameters, so this adds other missing parameters as well. 
* Deprecates `status_tracker` parameter
* Snuck in the documentation for the `grid` parameter in `Gallery.style` before I forget

Closes: #2283


gradio/events.py
```diff
@@ -366,21 +413,27 @@ def stream(
...
+            scroll_to_output=scroll_to_output,
+            show_progress=show_progress,
+            status_tracker=status_tracker,
             js=_js,
-            preprocess=_preprocess,
-            postprocess=_postprocess,
+            preprocess=preprocess,
+            postprocess=postprocess,
             queue=queue,
             show_progress=False,
```
aliabd, line 438: show_progress is defined twice

aliabd:
@abidlabs this doesn't build locally, see my comment on events.py

```
  File "/Users/aliabdalla/Desktop/dev/gradio/website/homepage/build.py", line 4, in <module>
    from src import index, guides, docs, demos
  File "/Users/aliabdalla/Desktop/dev/gradio/website/homepage/src/docs/__init__.py", line 2, in <module>
    from gradio.documentation import generate_documentation, document_cls
  File "/Users/aliabdalla/Desktop/dev/gradio/gradio/__init__.py", line 3, in <module>
    import gradio.components as components
  File "/Users/aliabdalla/Desktop/dev/gradio/gradio/components.py", line 41, in <module>
    from gradio.events import (
  File "/Users/aliabdalla/Desktop/dev/gradio/gradio/events.py", line 425
    self.set_event_trigger(
        ^
SyntaxError: keyword argument repeated: show_progress
```

abidlabs:
Thanks for catching that @aliabd! Will fix

abidlabs:
Should be fixed now @aliabd

aliabd (approved):
Not sure if this is related to the changes at all, but I can't get the flag button on `/demo/blocks_flag` to do anything. Can you take a look? 

Otherwise LGTM

gradio/components.py
```diff
@@ -3291,14 +3291,15 @@ def postprocess(self, y: List[np.ndarray | PIL.Image | str] | None) -> List[str]
...
-        grid: Optional[int | Tuple[int, int, int, int, int, int]] = None,
+        grid: Optional[int | Tuple] = None,
         height: Optional[str] = None,
         container: Optional[bool] = None,
     ):
         """
         This method can be used to change the appearance of the gallery component.
         Parameters:
             rounded: If True, will round the corners. If a tuple, will round corners according to the values in the tuple, starting from top left and proceeding clock-wise.
+            grid: Represents the number of images that should be shown in one row, for each of the six standard screen sizes (<576px, <768px, <992px, <1200px, <1400px, >1400px). if fewer that 6 are given then the last will be used for all subsequent breakpoints
```
freddyaboulton, line 3302: 👀

gradio/events.py
```diff
@@ -199,10 +208,13 @@ def clear(
         inputs: List[Component],
         outputs: List[Component],
         api_name: AnyStr = None,
+        status_tracker: Optional[StatusTracker] = None,
+        scroll_to_output: bool = False,
```
freddyaboulton, line 212: Was there a reason these parameters were omitted originally or was this just simply an oversight on our part?

gradio/events.py
```diff
@@ -326,20 +365,26 @@ def stop(
             inputs: List of inputs
             outputs: List of outputs
             api_name: Defining this parameter exposes the endpoint in the api docs
+            status_tracker: StatusTracker to visualize function progress
```
freddyaboulton, line 368: Do we have any examples of setting `status_tracker` ? Wondering if it will confuse users.

freddyaboulton (approved):
@abidlabs This looks good to me! Left a question about the new parameters were adding to the event triggers but nothing blocking.

gradio/events.py
abidlabs: I think an oversight. @aliabid94 do you know?

gradio/events.py
abidlabs: Yeah good point, @aliabid94 maybe it would be good to have an example or demo with `StatusTracker`, since I'm not sure how it's supposed to be used myself

aliabid94: No there's not really a good demo for it, I want to reconsider even having this argument.

aliabid94: blocks_xray does use it I believe

abidlabs:
Synced with @aliabid94 and confirmed that `StatusTracker` is deprecated. Removed it from the user-facing APIs but kept it internally as part of the config for backwards compatibility with `gr.Interface.load()`

abidlabs:
Thank you @aliabd and @freddyaboulton for the review! Merging in

Issue #1316, opened 2022-05-17
Queue upstream when loading apps via gr.Interface.load()


apolinario:
If **app A** uses `gr.Interface.load()` to load an **app B** that contains a `your_app.launch(enable_queue=True)`, the queue does not get respected when the **app B** is executed from the **app A**. So if there are 3 **app A** users, and all trigger **app B** at the same time, **app B** runs 3x in parallel, regardless if `enable_queue` was set to `True` on **app B**.

This implies in two things:
1. Any person can bypass the **app B** queue by using **app A** 
2. The host machine of **app B** may OOM if multiple users are running it from **app A**, as no queue is in place

This can currently be mitigated by adding a `enable_queue=True` to **app A** however this two shortcomings:
1. **app A** developer has to have the goodwill to include the queuing function there (including using `gr.Interface.load()` privately via localhost - which can be an easy way to bypass any queue), if they don't, **app B** gets resource drained outside of their control
2. This makes **app A** inefficient. Suppose **app A** not only loads **app B** but also loads **apps C and D** with `gr.Interface.load()`. If `enable_queue` is on **app A**, now when you call apps B, C, D from it, they get into the same queue, even though each could have very different internal queues (this happens with [MindsEye Lite](https://huggingface.co/spaces/multimodalart/mindseye-lite))

Suggested solution: 
- Add the **app A** users that are calling **app B** to the same queue as native **app B** users

If that gets implemented make sure to de-register from **app B**'s queue the users that leave/give up on **app A** too, otherwise an infinite queue could arise

apolinario:
@abidlabs, this may be important to include when bringing `gr.Interface.load()` to 3.x

abidlabs:
Thanks for the catch and the heads up @apolinario! Will implement appropriately

FarukOzderim:
We need to check if the loaded space is a queued space and create a ws connection from Backend.

abidlabs:
Hi @apolinario, we are planning on making queuing on by default everywhere (#2215) -- and it already has been on by default on Spaces for quite some time. 

I'm wondering whether that makes this issue redundant as it basically is the second approach that you've mentioned (**App A** will have queuing on by default, whether or not the developer has the good will to do this). The other point about efficiency is still there, but I wonder whether that could be remedied in a different way -- for example by having separate queues for different events.

apolinario:
[...]

PR #2294, opened 2022-09-19
Respect Upstream Queue when loading interfaces/blocks from Spaces

freddyaboulton:
# Description

Closes: #1316

The approach taken is to open up a websocket connection to the space for each prediction request sent to the loaded space. The main limitation of this approach is that the loaded app doesn't display information about the request's position in the original app's queue. However, there are a couple of reasons why I went with this approach:

1. We can't send updates to the front-end if the front-end isn't running, e.g. when using the interface or app as a function.
2. The app that's loading the interface may have its own queue, and if so, it doesn't make sense to display another app's queue. Imagine you're a user of an app with a queue that's loading another app with a queue. You'd see your position decrease until your request is being processed, but then you'd see your queue position jump back up again when the upstream queue gets the request sent from the downstream. I think that will be confusing to users who have no idea about how the downstream app they are using is implemented. In short, passing updates from the upstream queue to the downstream app would only make sense for `gr.Interface.load().launch` workflows but not more complex uses of `gr.Interface.load`. 

How to test,

Launch the following app:

```python
import gradio as gr

io = gr.Interface.load("spaces/freddyaboulton/saymyname")
print(io("foo"))
io.launch(enable_queue=True)
```

Then go to this space: https://huggingface.co/spaces/freddyaboulton/saymyname

Launch three simultaneous requests but make sure the first two are on the HF space. On the app running locally, it should take around ~15 seconds to complete the request.

respect_upstream_spaces

# Checklist:

### DIFF
--- a/gradio/external.py
+++ b/gradio/external.py
@@ -11,14 +11,15 @@ import operator
 import re
 import warnings
 from copy import deepcopy
-from typing import TYPE_CHECKING, Callable, Dict, List, Tuple
-import websockets
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Tuple
 
 import requests
+import websockets
 import yaml
+from packaging import version
 
 import gradio
-from gradio import components, utils
+from gradio import components, exceptions, utils
 
 if TYPE_CHECKING:
     from gradio.components import DataframeData
@@ -418,18 +419,33 @@ def get_spaces(model_name, api_key, alias, **kwargs):
         return get_spaces_blocks(model_name, config)
 
 
+async def get_pred_from_ws(
+    websocket: websockets.WebSocketClientProtocol, data: str
+) -> Dict[str, Any]:
+    completed = False
+    while not completed:
+        foo = await websocket.recv()
+        resp = json.loads(foo)
+        if resp["msg"] == "queue_full":
+            raise exceptions.Error("Queue is full! Please try again.")
+        elif resp["msg"] == "send_data":
+            await websocket.send(data)
+        completed = resp["msg"] == "process_completed"
+    return resp["output"]
+
+
 def get_ws_fn(ws_url):
-    async def get_pred_from_ws(data):
+    async def ws_fn(data):
         async with websockets.connect(ws_url) as websocket:
-            completed = False
-            while not completed:
-                resp = json.loads(await websocket.recv())
-                print(resp)
-                if resp['msg'] == "send_data":
-                    await websocket.send(data)
-                completed = resp['msg'] == 'process_completed'
-            return resp["output"]
-    return get_pred_from_ws
+            return await get_pred_from_ws(websocket, data)
+
+    return ws_fn
+
+
+def use_websocket(config):
+    return config["enable_queue"] and version.parse(
+        config["version"]
+    ) >= version.Version("3.1.5")
 
 
 def get_spaces_blocks(model_name, config):
@@ -446,6 +462,9 @@ def get_spaces_blocks(model_name, config):
     headers = {"Content-Type": "application/json"}
     ws_url = "wss://spaces.huggingface.tech/{}/queue/join".format(model_name)
 
+    ws_fn = get_ws_fn(ws_url)
+    connect_to_websocket = use_websocket(config)
+
     fns = []
     for d, dependency in enumerate(config["dependencies"]):
         if dependency["backend_fn"]:
@@ -453,8 +472,8 @@ def get_spaces_blocks(model_name, config):
             def get_fn(outputs, fn_index):
                 def fn(*data):
                     data = json.dumps({"data": data, "fn_index": fn_index})
-                    if config['enable_queue']:
-                        result = utils.synchronize_async(get_ws_fn(ws_url), data)
+                    if connect_to_websocket:
+                        result = utils.synchronize_async(ws_fn, data)
                         output = result["data"]
                     else:
                         response = requests.post(api_url, headers=headers, data=data)
--- a/gradio/external.py
+++ b/gradio/external.py
@@ -442,10 +442,13 @@ def get_ws_fn(ws_url):
     return ws_fn
 
 
-def use_websocket(config):
-    return config["enable_queue"] and version.parse(
-        config["version"]
+def use_websocket(config, dependency):
+    queue_enabled = config.get("enable_queue", False)
+    queue_uses_websocket = version.parse(
+        config.get("version", "2.0")
     ) >= version.Version("3.2")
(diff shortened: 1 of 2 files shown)


gradio/external.py
```diff
@@ -417,6 +419,35 @@ def get_spaces(model_name, api_key, alias, **kwargs):
         return get_spaces_blocks(model_name, config)
 
 
+async def get_pred_from_ws(
```
freddyaboulton, line 422: Split this out into it's own function as opposed to defining it in `get_ws_fn` so that it's easier to test. It's tricky mocking async context generators and I sunk way too much time into it.

gradio/external.py
```diff
@@ -417,6 +419,35 @@ def get_spaces(model_name, api_key, alias, **kwargs):
...
+    websocket: websockets.WebSocketClientProtocol, data: str
+) -> Dict[str, Any]:
+    completed = False
+    while not completed:
+        msg = await websocket.recv()
+        resp = json.loads(msg)
+        if resp["msg"] == "queue_full":
+            raise exceptions.Error("Queue is full! Please try again.")
```
freddyaboulton, line 430: I think raising an error when the queue is full is the right thing to do since that's what the web client does but curious what other's thoughts are.

abidlabs: Seems like a good option to me!

gradio/external.py
```diff
@@ -429,6 +460,10 @@ def streamline_config(config: dict) -> dict:
     config = streamline_config(config)
     api_url = "https://hf.space/embed/{}/api/predict/".format(model_name)
     headers = {"Content-Type": "application/json"}
+    ws_url = "wss://spaces.huggingface.tech/{}/queue/join".format(model_name)
+
+    ws_fn = get_ws_fn(ws_url)
+    connect_to_websocket = use_websocket(config)
 
     fns = []
     for d, dependency in enumerate(config["dependencies"]):
```
abidlabs, line 469: Also consider whether the specific function itself enables queue or not (see my comment to the PR)

abidlabs:
Thanks @freddyaboulton this looks great! Tested both with the downstream app enabling queue explicitly and with the downstream app not enabling queuing, like this:

```py
import gradio as gr

io = gr.Interface.load("spaces/freddyaboulton/saymyname")
print(io("foo"))
io.launch()
```

In both cases, the upstream queue is respected. Also tested with some upstream apps that don't have queue.

The one wrinkle that we should address is that in a Blocks demo, the upstream app may enable queuing for _some_ functions, but not all. This can happen by enabling the queue by default, but then disabling for some specific functions, or vice versa. The current implementation will only look at the default queuing value in the upstream app. Instead, when we iterate through the dependencies, it would be good to check for that specific function, queuing is enabled.

gradio/external.py
```diff
@@ -417,6 +419,35 @@ def get_spaces(model_name, api_key, alias, **kwargs):
...
+def get_ws_fn(ws_url):
+    async def ws_fn(data):
+        async with websockets.connect(ws_url, open_timeout=10) as websocket:
+            return await get_pred_from_ws(websocket, data)
+
+    return ws_fn
+
+
+def use_websocket(config):
+    return config["enable_queue"] and version.parse(
```
abidlabs, line 446: I believe older versions of Gradio did not include the `version` in the config, and so it would be good to check for that first so that a `KeyError` is not thrown

gradio/external.py
```diff
@@ -429,6 +460,10 @@ def streamline_config(config: dict) -> dict:
     config = streamline_config(config)
     api_url = "https://hf.space/embed/{}/api/predict/".format(model_name)
     headers = {"Content-Type": "application/json"}
+    ws_url = "wss://spaces.huggingface.tech/{}/queue/join".format(model_name)
```
abidlabs, line 463: I didn't know this was the URL? Why is it on a completely different domain than the `api_url`?

test/test_external.py
```diff
@@ -331,5 +339,61 @@ def test_can_load_tabular_model_with_different_widget_data(hypothetical_readme):
...
+async def test_get_pred_from_ws_raises_if_queue_full():
+    mock_ws = AsyncMock(name="ws")
+    messages = [json.dumps({"msg": "queue_full"})]
+    mock_ws.recv.side_effect = messages
+    data = json.dumps({"data": ["foo"], "fn_index": "foo"})
+    with pytest.raises(gradio.Error, match="Queue is full!"):
+        await get_pred_from_ws(mock_ws, data)
+
+
+@pytest.mark.skipif(sys.version_info < (3, 8), reason="Async mocks don't work for 3.7")
```
abidlabs, line 388: I'm confused, as there's as `AsyncMock` in the previous tests which are not skipped?

gradio/external.py
freddyaboulton: Good point!

test/test_external.py
```diff
@@ -331,5 +339,61 @@ def test_can_load_tabular_model_with_different_widget_data(hypothetical_readme):
...
+def test_respect_queue_when_load_from_config():
+    with unittest.mock.patch("websockets.connect"):
+        with unittest.mock.patch(
+            "gradio.external.get_pred_from_ws", return_value={"data": ["foo"]}
+        ):
+            interface = gr.Interface.load("spaces/multimodalart/saymyname")
```
abidlabs, line 394: I think you should use your own Space in the test in case `multimodalart` decides to change theirs. This test should also be marked as `flaky`.

test/test_external.py
freddyaboulton: The problem is that `websockets.connect` returns an async context manager and mock doesn't really work with async context managers in 3.7. I can make the `reason` a bit more precise.

abidlabs: Got it thanks for clarifying! Good to know

gradio/external.py
freddyaboulton: Not sure. My guess is that the spaces team restricts the ports/paths with which you can connect to the underlying machine.

abidlabs:
Good stuff! Left some more comments, mostly nits. The main thing is to respect upstream queue per-function as opposed to the entire app, as I mentioned above.

freddyaboulton:
Thanks for the review @abidlabs and good catch about honoring the queue per function. Made that change - this should be good for another look!

abidlabs:
Tested and LGTM @freddyaboulton! This is awesome

abidlabs (approved):

FarukOzderim:
@freddyaboulton this looks pretty good! I think adding some tests to the queue would be great using a similar approach.

PR #2308, opened 2022-09-20
Fix Web Tracker Script

aliabd:
Fixes the issue with the web tracker script, which is spamming slack with success messages despite not actually reloading the website. 

Problem was that I miswrote the if condition. This should work fine now and I've already started testing it on the gradio machine so we can see until this gets merged.


aliabid94 (approved):
LGTM

Issue #2305, opened 2022-09-20
Example caching doesn't work with `gr.update()`


abidlabs:
### Describe the bug

If a function returns a `gr.update()`, it will not work correctly with example caching. Here are a couple of failure modes:

(1) The caching does not happen at all, and an Exception is thrown

Example:

```py
import gradio as gr

io = gr.Interface(lambda x: gr.update(visible=False), "textbox", "html", examples=["hi"], cache_examples=True).launch()
```

Error:

```
AttributeError: 'dict' object has no attribute 'startswith'
```

(2) Caching happens without error, but when the example is clicked, the update does not happen as expected:

Example:

```py
import gradio as gr

io = gr.Interface(lambda x: gr.update(visible=False), "textbox", "textbox", examples=["hi"], cache_examples=True).launch()
```

Error:

<img width="626" alt="image" src="https://user-images.githubusercontent.com/1778297/191339448-705c3cc2-905d-4e88-8ce4-719b4306e13d.png">

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

See above

### Screenshot

See above

### Logs

```shell
See above
```

### System Info

```shell
gradio==3.3.1
```

### Severity

serious, but I can work around it

Issue #1798, opened 2022-07-15
Standardize return types in the backend


freddyaboulton:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
As a gradio user and developer, I wish the return type of the predict route was more predictable.

Right now it is very flexible.

This is the return type:

```python
return {
    "data": output,
    "duration": duration,
    "average_duration": block_fn.total_runtime / block_fn.total_runs,
}
```

The problem is that `data` can be a list of pretty much anything: text, numbers, interpretation output, filepaths, dataframes, updates. 

This makes the return type unpredictable - which makes it hard to maintain the front-end and also may be putting a burden to our users who hit the api manually.

**Describe the solution you'd like**  
Standardize the return types of the `data` key in the return type.

This is a sketch of how it could go: 

```python
class Update(BaseModel):
    label: str
    show_label: bool
    visible: bool
    value: float | str | DataframeData
    minimum : float | None
   <more fields>

class PredictResponse(BaseModel):
   data: List[Update]
   duration: float
   average_duration: float
```

**Additional context**

pngwn:
This is exactly what I had in mind @freddyaboulton.

abidlabs:
To clarify, would we have a single `Update` class with all of the parameters across all components? Or would we have separate `Update`'s for each component?

abidlabs:
Along similar lines, when the backend function returns an update, it can either be:
* a dictionary created by `gr.update()`
* a single value returned by the function

We should standardize this (convert everything to a dictionary) before being passed into the frontend

freddyaboulton:
This got done in the v4 branch!

PR #2309, opened 2022-09-20
Supports `gr.update()` in example caching

abidlabs:
Previously, we couldn't handle `gr.update()` in flagging or example caching, as described in #2305. This PR fixes that.

Fixes: #2305

I'm still writing some tests but ready to be reviewed

### DIFF
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -205,7 +205,7 @@ class IOComponent(Component, Serializable):
         if interactive is not None:
             config["mode"] = "dynamic" if interactive else "static"
         return config
-    
+
     @staticmethod
     def get_load_fn_and_initial_value(value):
         if callable(value):
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -59,7 +59,7 @@ from gradio.serializing import (
 set_documentation_group("component")
 
 
-class _Keywords(Enum):
+class _Keywords(str, Enum):
     NO_VALUE = "NO_VALUE"  # Used as a sentinel to determine if nothing is provided as a argument for `value` in `Component.update()`
     FINISHED_ITERATING = "FINISHED_ITERATING"  # Used to skip processing of a component's value (needed for generators + state)
 
@@ -205,7 +205,7 @@ class IOComponent(Component, Serializable):
         if interactive is not None:
             config["mode"] = "dynamic" if interactive else "static"
         return config
-
+    
     @staticmethod
     def get_load_fn_and_initial_value(value):
         if callable(value):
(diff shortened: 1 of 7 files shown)


gradio/components.py
```diff
@@ -59,7 +59,7 @@ class DataframeData(TypedDict):
 set_documentation_group("component")
 
 
-class _Keywords(Enum):
+class _Keywords(str, Enum):
```
abidlabs, line 62: Makes this enum jsonserializable

gradio/examples.py
```diff
@@ -189,9 +191,8 @@ def __init__(
         ]
         if cache_examples:
             for ex in non_none_examples:
-                if (
-                    len([sample for sample in ex if sample is not None])
-                    != self.inputs_with_examples
+                if len([sample for sample in ex if sample is not None]) != len(
```
abidlabs, line 194: Noticed a bug here

gradio/flagging.py
```diff
@@ -206,15 +206,18 @@ def flag(
                         component.label or f"component {idx}"
                     ),
                 )
-                csv_data.append(
-                    component.deserialize(
-                        sample,
-                        save_dir=save_dir,
-                        encryption_key=self.encryption_key,
+                if utils.is_update(sample):
```
freddyaboulton, line 209: I think it's strange we treat updates differently from returning a value from a function. At the end of the day, everything is a UI update and I think our api would be easier to maintain if we embraced that. This is similar to #1798. Definitely out of scope for this PR but something I'd like to add to the list of "nice to haves" for  4.0 😂

gradio/examples.py
```diff
@@ -278,10 +277,15 @@ async def predict_example(self, example_id: int) -> List[Any]:
         if len(self.outputs) == 1:
             predictions = [predictions]
         if not self._api_mode:
-            predictions = [
-                output_component.postprocess(predictions[i])
-                for i, output_component in enumerate(self.outputs)
-            ]
+            predictions_ = []
+            for i, output_component in enumerate(self.outputs):
```
freddyaboulton, line 281: Just realized this won't work if you return a dictionary of updates. I think we should file but not fix immediately since I think it's low priority.

freddyaboulton (approved):
This looks good to me @abidlabs ! Thanks for the fix.

gradio/examples.py
abidlabs: Thanks for the catch here as well. Fixed now + added a test

abidlabs:
Thank you so much @freddyaboulton for the comprehensive review. I've patched all of the holes now and this should be all set to merge after the tests pass

PR #2310, opened 2022-09-20
Make buttons respect full_width = False [hotfix]

aliabid94:
Fixes `Button.style(full_width=False)`

### DIFF
--- a/ui/packages/button/src/Button.svelte
+++ b/ui/packages/button/src/Button.svelte
@@ -19,6 +19,7 @@
 <button
 	on:click
 	class:!hidden={!visible}
+	class:!grow-0={style.full_width === false}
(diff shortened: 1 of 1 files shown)


abidlabs:
Tested this @aliabid94 and can confirm that it works, as in the following example:

```py

```
now produces this:

image

instead of this:

image

However, a couple of points:

1. It does not seem to work if the parent element is a `Column` instead of a `Row`. Take a look at this for example:

image

2. We should clarify in the documentation for `full_width` that the default behavior is that `full_width=True`. The current documentation seems to imply otherwise: https://github.com/gradio-app/gradio/blob/eaae71fdd1035490afca66f2e4f59800b74d03bd/gradio/components.py#L2677

Issue #1976, opened 2022-08-08
Provide a way to determine the selected image in a `Gallery`


pcuenca:
- [x] I have searched to see if a similar issue already exists.
Possibly related to #1379.

**Is your feature request related to a problem? Please describe.**  
The use case is: generate a grid of images, using for example a text to image model, then select one of the results for further processing.

**Describe the solution you'd like**  
I thought about two potential solutions:
- Update `value` to reflect the index of the currently selected image, so it can be checked as a response to some user action (click a button, for instance).
- Adopt `Gallery.click` and send that event whenever an image is clicked.

This sounds similar to the discussion in #1379, except for the upload ability. I think that exposing the selection would be useful even if upload is not yet supported.

abidlabs:
Thanks for creating this issue @pcuenca! Agreed that this would be a useful addition to Gallery, and is a lower hanging fruit than making Gallery support full upload capabilities.

abidlabs:
Another example of where this component would be useful is here: https://huggingface.co/spaces/NimaBoscarino/playlist-generator, where a person could simply click on one of the song pictures to get the lyrics

yuvalkirstain:
Hi, I am also interested in this feature. Specifically for my use case, it would have been cool if we could highlight the selected image in the gallery and act in the backend after a user selects an image.

temirov:
I want to display the images from a folder, and then allow a deletion of selected ones. The display part works but I can't find a way to know which image was clicked at. Resorting to custom JS would defeat the purpose of using Gradio in the first place :-)

MohamedAliRashad:
@abidlabs 
I can use something like this.
I want the user to be able to go through all the images in the gallery and if he wants to procecss a certain image further he can click on `gr.Button` and it will trigger a function that will read the selected image in gallery and make some processing on it.

Is there a way to do this ?

abidlabs:
Yes, please take a look at https://github.com/gradio-app/gradio/pull/3399

wwdok:
Hi, @abidlabs My gallery stores a list of images , I want when I click a image, I can get the image path just like that passed into the argument `value` of `gr.Gallery`， I print out all attributes of `evt`,
```
# the function triggered by Gallery.select()
def get_select_value(evt: gr.SelectData):
    print(f"==>> evt.selected: {evt.selected}") # True
    print(f"==>> evt._data: {evt._data}") # {'index': 1, 'value': None}
[...]

Issue #1379, opened 2022-05-25
Allow `Gallery` to be used as in input component and upload multiple images


MohamedAliRashad:
- [x] I have searched to see if a similar issue already exists.

I found a similar question [here](https://github.com/gradio-app/gradio/issues/713) but when i tried it. i only was able to upload one image.
What i am looking for is for the `gr.Image` class to have a flag called `multiple_images`. when it's `true` it allows user to upload more than one image and preview it before submission. (it delivers the images as a list of paths to my function).

aliabid94:
For now, using gr.File with file_count="multiple" would be the way to go, @dawoodkhan82 that works right? Will look into batch inputs for Image soon.

pngwn:
Gallery feels like the right way to do this. The only problem is that we currently always match inputs + outputs. What I mean by that is the inputs that a component accepts (when it is being populated as an output) and the values that a component provides (when it is used as an input) are always the same. It would be easy enough to change the interface for the Gallery to accept `{ images: List[string], selected: int }` but ideally when used as an input the gallery wouldn't send the index of the currently selected image but the image itself.

We could inroduce optional `input_value`(the value passed in either as the default value or the result of a predict function) and `output_value`(the value passed to a predict function) to work around issues like this, rather than simply using `value` for all cases. This wouldn't have any bearing on existing components, we could check for `input_value` or `output_value` first and fall back to `value` if they did not exist. We would need to consider how this would interact with other features, such as `js` functions.

abidlabs:
As discussed, moving to 3.x milestone @dawoodkhan82

abidlabs:
Renaming to reflect the new issue. In particular, it would be nice if the Gallery component could allow you to upload a bunch of images, and you could pass in the images, along with some of them selected into your function.

abidlabs:
A good example of where this component would be useful is here: https://huggingface.co/spaces/NimaBoscarino/playlist-generator, where a person could simply click on one of the song pictures to get the lyrics

diegoakel:
This would be super useful!

colemanhindes:
If you use the File option, then on mobile it doesn't take the user to their photos, it takes them to their files, which are often empty

abidlabs:
This is now possible via the `.select()` method and using the `File` component as mentioned earlier. You can now upload a bunch of images, and then as select them to preview them individually. Here's a code snippet to show how that would work:

```py
import gradio as gr

def preview(files, sd: gr.SelectData):
[...]

PR #2315, opened 2022-09-21
Gallery events WIP

abidlabs:
Colab with @pngwn to add `click()` and `change()` events to `Gallery`.

* The `click()` event should fire when an image is clicked, so that it can be used for some downstream function
* The `change()` event should fire when the Gallery loads images, allowing it to trigger downstream actions

Fixes: #1976, fixes: #1379 which should be marked as duplicates

### DIFF
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -3205,8 +3205,8 @@ class HTML(Changeable, IOComponent, SimpleSerializable):
         return self
 
 
-@document("style")
-class Gallery(IOComponent):
+@document("style", "change", "click")
+class Gallery(IOComponent, Changeable, Clickable):
     """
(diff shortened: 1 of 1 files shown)


abidlabs:
@pngwn I'll close this PR for now?

temirov:
The PR seems to be closed without merging, but it would have fixed many reported issues. Has it not worked as expected? (I see no tests in the PR)

abidlabs:
We were unable to get a fix that truly worked. However, I agree that this is a pretty important issue. We'll look into this in the near future

temirov:
Is it on the roadmap, if I may ask? E.g. do we have a vague feeling of the ETA? :-)

abidlabs:
It is on the roadmap, but not something we can attend to in the next month at least

Issue #2292, opened 2022-09-19
Cant mount gradio as a sub-application if the queue is enabled


freddyaboulton:
### Describe the bug

From @radames 

> I was happy to learn about mounting Gradio with Fast Api!! https://gradio.app/sharing_your_app/#mounting-within-another-fastapi-app
I’m trying to enable queue on a Block Interface with FastAPI serving static files, two things are happening:
first the websocket address has an extra slash via UI ws://127.0.0.1:8000/gradio//queue/join  ( I guess it’s the frontend bug, due to the mounting path different than /)
I’m not sure it this is the right way to enable Queue before mounting FastAPI app? while I can connect and send and receive the estimation response I don’t get any final responses

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

First, you won't be able to launch the following app because the websocket endpoint will be incorrect. If you manually modify the frontend to fix the ws endpoint, you won't be to run a prediction on this app. That is because the thread that runs the queue is never started.

```python
import gradio as gr
from fastapi import FastAPI
from fastapi.staticfiles import StaticFiles


app = FastAPI()

block = gr.Blocks()
with block:
    prompt = gr.Text()
    gallery = gr.Gallery().style(grid=[2], height="auto")
    contains_nfsw = gr.JSON(visible=False)
    btn = gr.Button("Generate")
    btn.click(dream, inputs=prompt, outputs=[gallery, contains_nfsw])
    
block.queue()
block.config['enable_queue'] = True
block.config['dev_mode'] = False

gradio_app = gr.routes.App.create_app(block)

app.mount("/gradio", gradio_app)
app.mount("/", StaticFiles(directory="static", html=True), name="static")
```

### Logs

```shell
-
```

### System Info

```shell
3.3
```

### Severity

serious, but I can work around it

radames:
hi @freddyaboulton I've tested #2319 thanks for the fast fix! when are you releasing it on beta?

freddyaboulton:
@radames I think we're doing a full release today!

abidlabs:
Yes full release today or tomorrow, but in the meantime, I believe it is out on `gradio==3.4b3`

PR #2319, opened 2022-09-21
Start queue when gradio is a sub application

freddyaboulton:
# Description

Closes: #2292

There were two problems as noted on the original issue:

1. The ws socket path was incorrect when running as a sub application
2. The thread to run the queue was not started because that's started by the `startup-events` route that's called during `Blocks.launch`.

I think I figured out the ws path issue although I'd like some feedback from @aliabid94 .

The queue thread not starting automatically is an interesting problem. I think the cleanest thing would be to use fastapi startup events to start the queue however there are two issues with that:

- Startup events are not started for sub applications ([docs](https://fastapi.tiangolo.com/advanced/events/)) so it wouldn't solve the problem of the queue not starting when gradio is mounted
- The queue needs to know about the URL where the gradio app is running in order to get the predictions. Since startup events happen before the app is running, the URL is not defined when the event runs.

I opted for adding a helper method called `mount_gradio_app` that will add a startup event to the parent application that runs the queue. This is not as nice as using the built-in FastAPI `app.mount(gradio_app, ...)` syntax but it's the best solution I see now. 

### Changes to the guide

image

# Checklist:

### DIFF
--- a/demo/custom_path/run.py
+++ b/demo/custom_path/run.py
@@ -5,13 +5,17 @@ CUSTOM_PATH = "/gradio"
 
 app = FastAPI()
 
+
 @app.get("/")
 def read_main():
     return {"message": "This is your main app"}
 
+
 io = gr.Interface(lambda x: "Hello, " + x + "!", "textbox", "textbox")
 gradio_app = gr.routes.App.create_app(io)
+app = gr.mount_gradio_app(app, gradio_app, server_name="localhost", port="8000", path="/gradio")
 
-app.mount(CUSTOM_PATH, gradio_app)
 
 # Run this from the terminal as you would normally start a FastAPI app: `uvicorn run:app` and navigate to http://localhost:8000/gradio in your browser.
+# Note that if you run uvicorn by setting the --host and --port flags, the values you set must match the values passed to
+# mount_gradio_app
--- a/demo/custom_path/run.py
+++ b/demo/custom_path/run.py
@@ -12,7 +12,7 @@ def read_main():
 
 
 io = gr.Interface(lambda x: "Hello, " + x + "!", "textbox", "textbox")
-app = gr.mount_gradio_app(app, io, path="/gradio")
+app = gr.mount_gradio_app(app, io, path=CUSTOM_PATH)
 
 
 # Run this from the terminal as you would normally start a FastAPI app: `uvicorn run:app`
--- a/demo/custom_path/run.py
+++ b/demo/custom_path/run.py
@@ -16,6 +16,7 @@ gradio_app = gr.routes.App.create_app(io)
 app = gr.mount_gradio_app(app, gradio_app, server_name="localhost", port="8000", path="/gradio")
 
 
-# Run this from the terminal as you would normally start a FastAPI app: `uvicorn run:app` and navigate to http://localhost:8000/gradio in your browser.
-# Note that if you run uvicorn by setting the --host and --port flags, the values you set must match the values passed to
-# mount_gradio_app
+# Run this from the terminal as you would normally start a FastAPI app: `uvicorn run:app`
+# and navigate to http://localhost:8000/gradio in your browser.
+# Note that if you run uvicorn by setting the --host and --port flags, the values you set
+# must match the values passed to mount_gradio_app
--- a/demo/custom_path/run.py
+++ b/demo/custom_path/run.py
@@ -12,11 +12,8 @@ def read_main():
 
 
 io = gr.Interface(lambda x: "Hello, " + x + "!", "textbox", "textbox")
-gradio_app = gr.routes.App.create_app(io)
-app = gr.mount_gradio_app(app, gradio_app, server_name="localhost", port="8000", path="/gradio")
+app = gr.mount_gradio_app(app, io, path="/gradio")
 
 
 # Run this from the terminal as you would normally start a FastAPI app: `uvicorn run:app`
 # and navigate to http://localhost:8000/gradio in your browser.
-# Note that if you run uvicorn by setting the --host and --port flags, the values you set
-# must match the values passed to mount_gradio_app
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -912,6 +912,7 @@ class Blocks(BlockContext):
             update_intervals=status_update_rate if status_update_rate != "auto" else 1,
             max_size=max_size,
         )
+        self.config = self.get_config_file()
         return self
 
     def launch(
@@ -1257,3 +1258,8 @@ class Blocks(BlockContext):
                     no_target=True,
                     queue=False,
                 )
+
+    def startup_events(self):
+        if self.enable_queue:
+            utils.run_coro_in_background(self._queue.start)
+        utils.run_coro_in_background(self.create_limiter)
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -1260,6 +1260,7 @@ class Blocks(BlockContext):
                 )
 
     def startup_events(self):
+        """Events that should be run when the app containing this block starts up."""
         if self.enable_queue:
             utils.run_coro_in_background(self._queue.start)
         utils.run_coro_in_background(self.create_limiter)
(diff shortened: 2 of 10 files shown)


ui/packages/app/src/api.ts
```diff
@@ -116,16 +116,18 @@ export const fn =
 			var ws_protocol = ws_endpoint.startsWith("https") ? "wss:" : "ws:";
 			if (is_space) {
 				const SPACE_REGEX = /embed\/(.*)\/\+/g;
-				var ws_path = Array.from(ws_endpoint.matchAll(SPACE_REGEX))[0][1];
+				var ws_path = Array.from(
```
freddyaboulton, line 119: My idea was just to always make sure the ws_path always ends in `/` but there could be other (better) solutions

gradio/blocks.py
```diff
@@ -912,6 +912,7 @@ def queue(
             update_intervals=status_update_rate if status_update_rate != "auto" else 1,
             max_size=max_size,
         )
+        self.config = self.get_config_file()
```
freddyaboulton, line 915: So that `queue_enabled` shows up in the config if you call `queue` outside of a context manager and without calling `launch`.

abidlabs:
Nice PR @freddyaboulton! I had a question about supplying the `server_name` and `port` to the `mount_gradio_app()` function. It would be great if the user didn't have to supply that manually and ensure its consistency with what's written on the terminal. 

Do we really need to know the absolute path for the local server? It seems like that is used inside the `Queue.call_prediction()` function, but it's unclear to me why we need the absolute path here.

freddyaboulton:
@abidlabs I agree it's not ideal to have to specify the server and port in two places. I think `mount_gradio_app` needs the full url so that it can call `gradio_app.blocks._queue.set_url`. Otherwise the queue won't know which url to send the post requests to in `call_prediction`. 

Do you have other solutions in mind?

aliabid94:
websocket part LGTM.

abidlabs:
> @abidlabs I agree it's not ideal to have to specify the server and port in two places. I think mount_gradio_app needs the full url so that it can call gradio_app.blocks._queue.set_url. Otherwise the queue won't know which url to send the post requests to in call_prediction.

I might be missing something but can't the queue just use the relative URL like all of the other route? Basically here:

https://github.com/gradio-app/gradio/blob/581fbabe07383fe50231c9053d3a40ba9c395b83/gradio/queue.py#L217

Why can't just be:

```py
url="/api/predict",
```

freddyaboulton:
@abidlabs The queue is running on a separate thread than the main gradio app so it needs the full url path to the predict endpoint to access the gradio api. 

Just tried out your suggestion and I got an error with the `blocks_essay` demo with queue enabled:

image

abidlabs:
Gotcha, testing the PR now

gradio/blocks.py
abidlabs: To confirm, is this just needed for the new test in `test/test_blocks`?

gradio/utils.py
```diff
@@ -680,3 +680,28 @@ def validate_url(possible_url: str) -> bool:
 
 def is_update(val):
     return type(val) is dict and "update" in val.get("__type__", "")
+
+
+def mount_gradio_app(
+    app: fastapi.FastAPI, gradio_app: App, server_name: str, port: str, path: str
```
abidlabs, line 686: Can we make `port` an `int` to make it consistent the type of the server port in other parts of the codebase?

gradio/utils.py
```diff
@@ -680,3 +680,28 @@ def validate_url(possible_url: str) -> bool:
 
 def is_update(val):
     return type(val) is dict and "update" in val.get("__type__", "")
+
+
+def mount_gradio_app(
```
abidlabs, line 685: Nit/suggestion: it feels more natural to me for this method to belong in `routes.py` with all of the other fastapi-related stuff

demo/custom_path/run.py
```diff
@@ -5,13 +5,18 @@
...
 
+
 @app.get("/")
 def read_main():
     return {"message": "This is your main app"}
 
+
 io = gr.Interface(lambda x: "Hello, " + x + "!", "textbox", "textbox")
 gradio_app = gr.routes.App.create_app(io)
+app = gr.mount_gradio_app(app, gradio_app, server_name="localhost", port="8000", path="/gradio")
```
abidlabs, line 16: A suggestion for this function: can we have it take a Gradio Blocks object instead of a FastAPI app as its second parameter? And then internally it calls it `gr.routes.App.create_app(io)`. This saves our users a line of code

abidlabs:
Tested and this PR works great, and allows a FastAPI app to mount a Gradio app with or without queue enabled.

My main suggestion would be to try to simplify the usage for our users. I noticed that there is actually a way to get the server URL without needing the user to specify it. In any FastAPI route, if you add the `request: Request` parameter, you can access the server's URL with `request.url._url` (or `str(request.url)` which may be better since it doesn't require accessing a private variable). Similarly, for a websocket route, you can do `websocket.url._url`.

So this suggests an idea:
1. When the FastAPI server first launches, set `_queue.server_path=None` (this is already the case)
2. In the `@app.websocket("/queue/join")` route, check the queue's `server_path`. If it is currently `None`, then set it to a stripped down version of `websocket.url._url`

This should eliminate the `server_name` and `port` parameters from `mount_gradio_app()`

demo/custom_path/run.py
freddyaboulton: Sure thing! Good point.

gradio/utils.py
freddyaboulton: Done!

gradio/blocks.py
freddyaboulton: It's so that the queue is enabled in the config when you do `blocks.queue()` but don't call `.launch()`

freddyaboulton:
Thanks for the awesome suggestion @abidlabs ! I agree that it's much better to not have users specify the same information twice. I pushed up the suggestion you made - we can now set the gradio api endpoint dynamically from the queue join route if it's not specified.

This works! I deployed a fastapi app with gradio as a sub application here: https://gradio-sub-app.onrender.com/

If you go to https://gradio-sub-app.onrender.com/gradio/ you will see a gradio app!

image

You can also access the lion image here: https://gradio-sub-app.onrender.com/lion.jpg (I mounted a static file app on the root path).

The problem with this approach is that it won't work on spaces because spaces only expose websockets on a specific endpoint that's different from the one gradio is running on, e.g. `wss://spaces.huggingface.tech/space-name/queue/join`.

What if we keep an optional parameter called `gradio_api_endpoint` to `mount_gradio_app` that let's users manually set the gradio endpoint if they desire.

With that approach, I was able to get gradio sub-applications to work on spaces here: https://huggingface.co/spaces/freddyaboulton/gradio-subapp

abidlabs:
Thank you so much for trying out the suggestions @freddyaboulton!

> What if we keep an optional parameter called gradio_api_endpoint to mount_gradio_app that let's users manually set the gradio endpoint if they desire.

Sounds good! Let's add and document this parameter :)

freddyaboulton:
@abidlabs Just pushed up the doc changes!

Documentation here:

image

Updated guide here:

image

Should be good for a final look now!

gradio/routes.py
```diff
@@ -382,3 +385,42 @@ def get_types(cls_set: List[Type]):
...
+
+set_documentation_group("routes")
+
+
+@document()
+def mount_gradio_app(
+    app: fastapi.FastAPI,
+    blocks: gradio.Blocks,
+    path: str,
+    gradio_api_url: Optional[str],
```
abidlabs, line 405:
```suggestion
    gradio_api_url: Optional[str] = None,
```

abidlabs:
Works great @freddyaboulton. I'm going to push some small tweaks to the documentation (easier than commenting in all of the places) if you don't mind! Feel free to take or leave

abidlabs (approved):

abidlabs:
Awesome @freddyaboulton! LGTM

freddyaboulton:

Issue #2311, opened 2022-09-21
`copy to clipboard` is not working


secsilm:
### Describe the bug

`copy to clipboard` is not working when `output="json"`. Click it and no response.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```python
import gradio as gr

demo = gr.Interface(lambda name: {"output": f"Hello {name}!"}, inputs="text", outputs='json')

if __name__ == "__main__":
    demo.launch(server_name='0.0.0.0', debug=True)
```

### Screenshot

![image](https://user-images.githubusercontent.com/8894643/191458063-3b786c4f-cace-4324-bb92-6b782145d79e.png)

### Logs

```shell
Running on local URL:  http://0.0.0.0:7860

To create a public link, set `share=True` in `launch()`.
```

### System Info

- Gradio version: 3.3.1
- OS: Ubuntu 18.04
- Browser: Chrome 105.0.5195.102

### Severity

annoying

abidlabs:
Thanks @secsilm can confirm the issue! @dawoodkhan82 would you be able to take a look at this?

freddyaboulton:
If you remove `server_name` it works 🤔

freddyaboulton:
Chrome automatically blocks access to the clipboard since the site is not secure:

![image](https://user-images.githubusercontent.com/41651716/191551867-c83e9797-588f-4254-b8eb-20340a704d7a.png)

abidlabs:
Ah interesting, it also doesn't work inside iframes (e.g. inside of a jupyter notebook). 

It seems like we should be able to fix that: https://stackoverflow.com/questions/61401384/can-text-within-an-iframe-be-copied-to-clipboard

ihmily:
> If you remove `server_name` it works 🤔

It works! Thank you very much.

PR #2321, opened 2022-09-22
Clipboard fix for iframes

abidlabs:
Previously, the copy button in the JSON output would not work when the Gradio demo was embedded in a jupyter/colab notebook. 

This grants the <iframe> permission to copy. 

Fixes: #2311 (well not the original issue, since that's beyond our control, but the version of it for iframes)


freddyaboulton (approved):
Thanks @abidlabs! Tested and works great

aliabid94 (approved):
LGTM

Issue #248, opened 2021-07-26
Opening Gradio inside jupyter notebook which his hosted in a docker container


Mcprasad:
I have a Jupyter notebook hosted in a docker container - which is used to run my data science workloads. I would like to use gradio locally inside that jupyter notebook.. 

To expose gradio locally inside the container - I am setting the server_address as 0.0.0.0 according to the documentation.  I am able to view gradio UI in the local host but it does not open up inside jupyter.. 

any suggestions ?

![image](https://user-images.githubusercontent.com/19703489/127023478-a4114628-fc1c-4f7c-b4a6-f7569c9e8b88.png)

but if i explicitly try top open it in a browser by overriding 0.0.0.0 to local host it works.. i would like to open it right inside jupyter inside the docker container..

Mcprasad:
![image](https://user-images.githubusercontent.com/19703489/127023433-19700313-6f38-483b-8649-aa9da54684ff.png)

abidlabs:
Hmm. Can you try setting `server_address` to be `localhost` and see what happens?

Mcprasad:
@abidlabs - Yes, i tried that too.. same error.. any suggestions ?

abidlabs:
Unfortunately, I can't seem to reproduce this on my end. When I test it on a hosted jupyter notebook, I observe the following behavior:

- If port forwarding is set up correctly in docker container, then the interface appears in the notebook
- If port forwarding is not set up, then the interface doesn't appear in line or in a new tab. 

In the 2nd case, you can still see the interface inline by setting `share=True`. Perhaps try that with a simple function to see if that solves your issue? If your model is too sensitive to set public sharing, you can set up a username/password by using the `auth` parameter in the `Interface` class. Or you can try [Gradio Private](https://gradio.app/introducing-private), in which case, we set up a secure on-premise version of Gradio for you.

aliabid94:
Closing due to lack of response. Feel free to reopen @Mcprasad

PR #2324, opened 2022-09-22
Adds the option to run gradio from the remote Jupyter notebook accessible at the non-localhost IP address.

adamryczkowski:
# Description

## Motivation

I want to run gradio on the Jupyter notebook on the server accessible via the LAN/VPN

## Summary of the change 

Adds a new launch parameter, `network_interface_address`, that denotes the IP address of the network interface to be used to serve the gradio content. Defaults to the local interface (127.0.0.1) if `server_name` is None, or all local interfaces (0.0.0.0) if `server_name` is specified. 

Changes the meaning of the `server_name` parameter. Now it is used to specify the address of the gradio server as seen from the browser perspective. Defaults to the server's fully qualified domain name if `network_interface_address` is provided, otherwise `localhost`. 

My change modifies the number of the elements returned by the `networking.py:start_server` function. 

I did not write unit tests for this change.

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -1067,7 +1067,7 @@ class Blocks(BlockContext):
         # If running in a colab or not able to access localhost,
         # a shareable link must be created.
         is_colab = utils.colab_check()
-        if is_colab or (_frontend and not networking.url_ok(self.local_url)):
+        if is_colab or (_frontend and network_interface_address == "127.0.0.1" and not networking.url_ok(self.local_url)):
             if not self.share:
                 raise ValueError(
                     "When running in Google Colab or when localhost is not accessible, a shareable link must be created. Please set share=True."
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -1034,8 +1034,8 @@ class Blocks(BlockContext):
         else:
             server_name, network_interface_address, server_port, local_url, app, server = networking.start_server(
                 self,
-                server_name,
                 network_interface_address,
+                server_name,
                 server_port,
                 ssl_keyfile,
                 ssl_certfile,
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -930,6 +930,7 @@ class Blocks(BlockContext):
         prevent_thread_lock: bool = False,
         show_error: bool = False,
         server_name: Optional[str] = None,
+        network_interface_address: Optional[str] = None,
         server_port: Optional[int] = None,
         show_tips: bool = False,
         height: int = 500,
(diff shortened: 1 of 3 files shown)


abidlabs:
Hi @adamryczkowski thanks for opening this PR. Can you help me understand why the `server_name` parameter is not sufficient for this use case?

adamryczkowski:
As of now, the `server_name` does not work, because it will be replaced with `localhost` anyway. 
It could be a bug, and in fact my first attempt to this PR was to fix it, so the `server_name` would be enough. Then I realized that this may brake some of your other applications, because there are two completely different concepts: one is where the `gradio` should be served from the server (backend) perspective, and the other, how the gradio should be referred to from the back-end perspective. 

That's why I ended up adding additional parameter.

abidlabs:
Hi @adamryczkowski would you be able to provide more details with what's going wrong with the original implementation? I'm afraid I don't quite follow the motivation. It would be great if you can file an issue with code / reproduction instructions that this PR closes.

abidlabs:
Hi @adamryczkowski, will close this PR as part of some general repo housekeeping. Still would like to help if you're running into any issues, but please create a dedicated issue and then we can work from there. Thanks!

adamryczkowski:
Does it mean you are abandoning my PR without giving an explanation why?

Sep 29, 2022 22:18:10 Abubakar Abid ***@***.***>:

abidlabs:
Hi @adamryczkowski like I mentioned in my earlier comments, we need more details to understand the issue that you are facing and the motivation for this PR. Please open an issue providing a reproducible code example so that we can understand the problem. Then, we'd be happy to review a PR to solve that problem. Currently, it is not clear what this PR is solving.

abidlabs:
Thanks @adamryczkowski I've created the issue here: https://github.com/gradio-app/gradio/issues/2382 and we'll take a look it. Feel free to add more details

wormcode:
I have come across the same issue on amazon SageMaker studio lab platform, the following will give some test results:
1.code
`import gradio as gr

def greet(name):
    return "Hello " + name + "!"

demo.launch() #share=True)  
`

gradio_jupyternotebook_error

the corresponding iframe html is



<iframe src="http://127.0.0.1:7866/" allow="autoplay; camera; microphone; clipboard-read; clipboard-write;" allowfullscreen="" width="100%" height="500" frameborder="0"></iframe>


2. change server_name to remote domain name :
`import gradio as gr

def greet(name):
    return "Hello " + name + "!"

demo.launch(server_name="h3wvtdkldcldp1uvqp.studio.us-east-2.sagemaker.aws") #share=True) `

results:
ERROR:    [Errno 99] error while attempting to bind on address ('18.190.140.180', 7869): cannot assign requested address
ERROR:    [Errno 99] error while attempting to bind on address ('18.190.96.147', 7870): cannot assign requested address
ERROR:    [Errno 99] error while attempting to bind on address ('18.190.66.74', 7871): cannot assign requested address
ERROR:    [Errno 99] error while attempting to bind on address ('18.190.140.180', 7872): cannot assign requested address
....
seems endless attempting

@abidlabs 
if you would like to reproduce this issue, you could apply a sagemaker studio lab account to try for yourself 

I think the following bugs share the same symptom 
[https://github.com/gradio-app/gradio/issues/2382](url)
[https://github.com/gradio-app/gradio/issues/248](url)
[https://github.com/gradio-app/gradio/pull/2324](url)

Issue #2325, opened 2022-09-22
Microphone cannot be stopped after starting without permissions


osanseviero:
### Describe the bug

([internal discussion](https://huggingface.slack.com/archives/C02QZLG8GMN/p1663834347523179))

Using the microphone component in any Gradio app (such as https://huggingface.co/spaces/ysharma/Talk_to_Multilingual_AI_WhisperBloomCoqui), in Chrome, **without** microphone permissions, allows me to click "Record from microphone". Afterward, I cannot click "Stop recording".

The UX should not have allowed me to click "Record", as I didn't give permission. Being stuck in recording was very confusing

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Any gradio app, e.g. https://huggingface.co/spaces/ysharma/Talk_to_Multilingual_AI_WhisperBloomCoqui

### Logs

```shell
-
```

### System Info

```shell
Any version, Chrome
```

### Severity

annoying

Nachimak28:
Have been facing this issue since yesterday so can confirm.
Even after providing the correct permissions in site settings of chrome, the issue persists.
My gradio version is 3.3.1.

This is what the console displays after pressing the record and stop buttons.
![image](https://user-images.githubusercontent.com/23210132/191964337-e4687b93-77fc-4411-81fa-54dfef810746.png)

PR #2330, opened 2022-09-24
Catch the permission exception on the audio component

Ian-GL:
# Description

Currently the Audio component fails silently if the microphone permissions were not granted and the UI seems as if it was recording. This fix catches the permission exception and raises an alert that tells the user the permissions are needed. It also prevents the UI from showing the stop button as if it had started recording.

Screen Shot 2022-09-24 at 9 51 50

Closes: #2325 

# Checklist:


abidlabs:
Thanks @Ian-GL for the fix! This looks pretty good, but I wonder if it would be possible to use Gradio's Error instead of alert() for a better UI? cc @aliabid94

Ian-GL:
> Thanks @Ian-GL for the fix! This looks pretty good, but I wonder if it would be possible to use Gradio's Error instead of alert() for a better UI? cc @aliabid94

Hey @abidlabs , I wasn't aware of the error. Yes, if you have an example of another component that uses it I could use that as a guide to add it here.

ui/packages/audio/src/Audio.svelte
```diff
@@ -57,7 +57,26 @@
...
+			.catch((err) => {
+				if (err instanceof DOMException && err.name == "NotAllowedError") {
+					alert(
+						"Please allow the use of microphone if you wish to record audio."
+					);
+					return null;
+				} else {
+					throw err;
+				}
+			});
```
pngwn, line 74: We shouldn't be mixing await and `then` chaining. Lets wrap this in a `try` `catch` block instead.

Additionally should think of something less intrusive than an `alert` which blocks all interaction with the page? cc @gary149 @abidlabs

pngwn (changes requested):

aliabid94:
I'll make a PR against this fork to launch a gradio.Error from within a component - we don't have any examples of doing this before so I'll take care of that.

aliabid94:
Just made the changes to use gradio error messaging, @pngwn please rereview

pngwn:
Pushed a change to convert the chained promises to `try` `catch` style, so we aren't mixing await + promise chaining, and fixed and issue with the type.

Looks good, thanks @Ian-GL and @aliabid94!

pngwn (approved):

pngwn:
Will merge when CI passes with everything merged in.

Ian-GL:
Thanks for taking care of the changes, @pngwn and @aliabid94 !

Issue #2337, opened 2022-09-26
Analytics-Python update


celestinoxp:
### Describe the bug

analytics-python is in version 1.4.0 and does not allow updating its dependencies, thus creating conflicts with other packages that require these updated dependencies.

Dependencies: backoff, monotonic, python-dateutil, requests, six

Turns out analytics-python has been updated to new versions but under another name, changed from "analytics-python" to "segment-analytics-python" which is now at version 2.2.1.

When upgrading from analytics-python to segment-analytics-python, dependencies can be updated and thus avoiding dependency conflicts in other packages.

The GitHub is the same: https://github.com/segmentio/analytics-python

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

.

### Logs

```shell
.
```

### System Info

```shell
latest
```

### Severity

serious, but I can work around it

PR #2347, opened 2022-09-27
Removes `analytics` dependency

abidlabs:
It turns out we never needed `analytics` given that our analytics are processed through our own backend server hehe

Closes: #2337


freddyaboulton (commented):
Thanks @abidlabs !

aliabd (approved):
LGTM

PR #2351, opened 2022-09-28
Stream works now

### LINKED ISSUE #1710: Streaming Audio is choppy
### Describe the bug

When streaming audio, the reconstructed signal of the streamed chunks sounds choppy.

### Reproduction

To reproduce, run:

```py
import gradio as gr
import numpy as np


def run(audio, state):
    sr, data = audio
    if state is None:
        state = data
    else:
        state = np.concatenate([state, data])
    audio = sr, state
    return audio, state


gr.Interface(
    fn=run,
    inputs=[
        gr.Audio(source="microphone", type="numpy", streaming=True),
        "state"
    ],
    outputs=[
        "audio",
        "state"
    ],
    live=True,
).launch()
```

On your local machine and listen to the recorded audio.
The problem also persists for the audio streaming demo:
https://github.com/gradio-app/gradio/tree/main/demo/stream_audio

### Screenshot

_No response_

### Logs

```shell
-
```

### System Info

```shell
Gradio 3.0.24
Ubuntu 20.04.4 LTS
Firefox 101.0.1 (64-bit)
```

### Severity

aliabid94:
Streaming audio fixed now, thanks to @yannickfunk!  Basically originally streamed audio was choppy, because we were constantly stopping and starting the MediaRecorder to get chunks of audio to stream. Now, we use npm libraries "extendable-media-recorder" and extendable-media-recorder-wav-encoder" to let us encode the data as in WAV/pcm format, which allows us to send continuous chunks audio perfectly spliced together. Full discussion of this methof in #1710. 

Fixes #1710.


ui/packages/audio/src/Audio.svelte
```diff
@@ -57,35 +65,47 @@
...
-		recorder = new MediaRecorder(stream);
-
-		recorder.addEventListener("dataavailable", (event) => {
-			audio_chunks.push(event.data);
-		});
-
-		recorder.addEventListener("stop", async () => {
-			if (!streaming) {
-				recording = false;
+		recorder = new MediaRecorder(stream, { mimeType: "audio/wav" });
```
pngwn, line 70: These new dependencies are pretty big, do we know how big the treeshaken module is (as we are only importing parts of it)? I'm guessing it is still quite sizable, given what it is doing.

If it is still quite big, it would be good if we only load them when streaming is enabled (via dynamic imports). Happy to help add this functionality if we decide to go this route.

pngwn (changes requested):
Looks good other than the issues of the dependency size when users aren't streaming. If we can clarify that it won't have an impact or take care of it, then I think it is good to go.

ui/packages/audio/src/Audio.svelte
yannickfunk: I can imagine it would be also sufficient to just use the necessary implementations from this dependency

pngwn: Vendoring in the code would put us on the hook for maintenance which probably isn't ideal as this is not anyone on the team's area of expertise as far as I'm aware.

pngwn:
Have confirmed there is a substantial size increase. Unminified we go from ~100KB -> ~500KB. Minified we go from ~30KB -> ~135KB.

Will resolve conflicts now + sync with @aliabid94 later to address this.

ui/packages/audio/src/Audio.svelte
yannickfunk: Thats a good point

ui/packages/audio/src/Audio.svelte
```diff
@@ -73,38 +96,72 @@
...
-		if (stream === null) {
-			return;
-		}
+		if (stream == null) return;
 
-		recorder = new MediaRecorder(stream);
+		if (streaming) {
+			const [{ MediaRecorder, register }, { connect }] = await Promise.all(
+				module_promises
+			);
```
pngwn, line 104: We started the module fetching earlier on to avoid unnecessary delays but when someone presses record we need to make sure the modules have finished importing. We can do this by awaiting the promises we created earlier. If they are still loading them will will wait for them, then assign the exports we care about to these variables (we are deconstructing the arrays + objects).

If the promise has already resolved (imports loaded) then it will simply unpack the value and assign the parts we care about these values immediately.

ui/packages/audio/src/Audio.svelte
```diff
@@ -29,13 +31,33 @@
...
 	let crop_values = [0, 100];
+	const STREAM_TIMESLICE = 500;
+	const NUM_HEADER_BYTES = 44;
+	let audio_chunks: Array<Blob> = [];
+	let audio_blob;
+	let module_promises:
+		| [
+				Promise<typeof import("extendable-media-recorder")>,
+				Promise<typeof import("extendable-media-recorder-wav-encoder")>
+		  ];
```
pngwn, line 49: We needed the types of the things we are actually importing, we can do this by doing `typeof <thing we want the type of>` in the case of `typeof import(...)` it gives us the type of the dynamically imported module (not of the return value of `import(module)`). Since `import()` always returns a promise we pass this to the `Promise<T>` generic, so TS knows it is a promise containing that module.

ui/packages/audio/src/Audio.svelte
```diff
@@ -29,13 +31,33 @@
...
+	function get_modules() {
+		module_promises = [
+			import("extendable-media-recorder"),
+			import("extendable-media-recorder-wav-encoder")
+		];
+	}
+
+	if (streaming) {
+		get_modules();
+	}
```
pngwn, line 60: We only get the modules if we are streaming. The vanilla stuff is fine for recording.

pngwn (commented):
@aliabid94 I've commented the code to explain what is going on this the lazy loading.

Unfortunately I've had to add back the previous implementation because we don't want to use the new modules when we aren't streaming due to the 5x increase in bundle size for this component. So we use the new implementation when streaming and the old when we aren't. There is a little reuse but it isn't ideal. I don't know how much more can shared.

One big issue I am seeing with streaming though is that the final chunk gets cut off depending when you stop the stream. If you stop it just as a chunk is sent off (basically when a `dataavailable` event is fired) then it is fine but if you don't then the end will be cut off.

I did have a quick play around to see why it might be happening but I didn't have much luck. It could be because we only call `dispatch` when we are in a certain state. Another weird thing is that a `dataavailable` event is raised _after_ a `stop` event is raised which isn't what MDN says should happen. 

Either way, if you could take a look that would be great. Other than this, it looks good!

aliabid94:
Fixed the "last chunk missing" error on streaming, basically what was happening was that if a user stopeed the recording, and the backend function was still pending, the last few chunks in the time between the last backend call and stopping the recorder never got called.

So now, it waits until the backend call is complete, and submits the remaining chunks to the backend.

Let me know if it looks good @pngwn, feel free to merge if there are no more issues :)

abidlabs:
@aliabid94 I tested `demo/stream_audio` and this works great. No audio gaps or clipping!

However, I tried using it with a real speech-to-text pipeline the way our users might (see the demo `streaming_wav2vec`, which I just pushed), and the problem is that the output progress bar keeps going over the text, which makes it impossible to see the output until you are done. Am I doing something wrong because I thought we solved this issue already? For streaming events we should disable the progress indicator by default? 

image

abidlabs (approved):

pngwn (approved):
Looks great! Awesome work.

aliabid94:
> @aliabid94 I tested `demo/stream_audio` and this works great. No audio gaps or clipping!
> 
> However, I tried using it with a real speech-to-text pipeline the way our users might (see the demo `streaming_wav2vec`, which I just pushed), and the problem is that the output progress bar keeps going over the text, which makes it impossible to see the output until you are done. Am I doing something wrong because I thought we this was an old issue... For streaming events we should disable the progress indicator by default?
> 
> image

Fixed @abidlabs

abidlabs (approved):
Tested both:
* stream_audio
* streaming_wav2vec

Work great, thanks for the fix @aliabid94!

Issue #2320, opened 2022-09-21
Components/Blocks have duplicate IDs if rendered more than once


abidlabs:
### Describe the bug

If a component or Interface is rendered, but has already been rendered previously, it will add blocks with duplicate IDs to the config. Here's an example:

```py
io3b = gr.Interface(
    lambda x: [x["mask"], x["image"]],
    gr.ImageMask(),
    [gr.Image(), gr.Image()],
)

with gr.Blocks():
  io3b.render()
  io3b.render()
```

We assume uniqueness of component IDs so this can cause all sorts of problems. The safest solution would be to raise an `Exception` to prevent a component/blocks from being rendered multiple times.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

See above

### Logs

```shell
N/A
```

### System Info

```shell
3.3.1
```

### Severity

serious, but I can work around it

PR #2359, opened 2022-09-28
Raise error if Blocks has duplicate component with same IDs

abidlabs:
Fixes: #2320 

Also avoids making a call to figure out the IP address if analytics is disabled. The reason for that change is that I was working with airplane Wifi and noticed it was *extremely* slow to launch a Gradio Interface because it was taking a long time to get the IP address. The real solution is to use async requests, but this is a temporary bandaid


gradio/exceptions.py
```diff
@@ -1,3 +1,7 @@
+class DuplicateBlockError(KeyError):
```
freddyaboulton, line 1: I think KeyError is usually reserved for when a key in a dictionary is not present? Minor point but I think a `ValueError` or `LookupError` would be more appropriate here.

freddyaboulton (approved):
@abidlabs Thanks! The changes related to the DuplicateBlockError look good to me but some of the tests are failing due to the refactoring of the analytics payload. What's the intended purpose of that change?

gradio/exceptions.py
abidlabs: Sure, will change to `ValueError`

abidlabs:
Thanks for the review @freddyaboulton! I added the reason for the analytics refactor above:

> Also avoids making a call to figure out the IP address if analytics is disabled. The reason for that change is that I was working with airplane Wifi and noticed it was extremely slow to launch a Gradio Interface because it was taking a long time to get the IP address. The real solution is to use async requests, but this is a temporary bandaid

abidlabs:
Should be all set to merge now (cc @freddyaboulton).

gradio/interface.py
```diff
@@ -486,7 +485,8 @@ def render_flag_btns(flagging_options):
 
                     with Column(variant="panel"):
                         for component in self.output_components:
-                            component.render()
+                            if not (isinstance(component, State)):
```
freddyaboulton, line 488: Out of curiosity - this was a pre-existing bug?

abidlabs: Originally we would unnecessarily create a second `State` component with the identical properties as the first. It didn't break anything, but better to avoid it imo.

PR #2363, opened 2022-09-28
Change version.txt to 3.4 instead of 3.4.0

aliabd:
Spaces seems to have stricter version checking than pypi, and doesn't accept 3.4.0 for 3.4 

We read from version.txt and when that's updated, we automatically update all the spaces in the demos tab in the website. Currently all are down because they can't use 3.4.0


abidlabs:
Yikes sorry! We should probably update the reading logic so that it properly formats the version #. But this is good for a hotfix

PR #2365, opened 2022-09-29
Fix image_classifier_interface_load demo

freddyaboulton:
# Description

Fixes image_classifier_interface_load so that path to images is correct


aliabid94 (approved):
LGTM

Issue #2040, opened 2022-08-18
gr.Shortcuts are missing from docs


FarukOzderim:
- [X] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
We don't have components inside Templates in docs with their class names. I think we do miss that? See [here](https://github.com/gradio-app/gradio/blob/main/gradio/__init__.py#L58-L70)

**Describe the solution you'd like**  
We can add them to shortcuts as a new column.
Alternatively drop listing string shortcuts and only list gr.Shortcut

Issue #134, opened 2021-03-09
Image mode not supporting alpha channel


mortium91:
We noticed for image inputs only rgb and black/white is supported and the alpha channel gets removed. You guys plan to support transparency anytime soon? We use .png's with transparency a lot!

Besides that, Gradio is awesome btw!

aliabid94:
You can actually pass image_mode="RGBA". Thanks for pointing that this is missing from the documentation! We use Pillow for image processing, so any image_mode supported by PIL is supported by gradio

Skiddoh:
Hi, sorry to bother you, but this took me forever to figure out, because inputting a image with alpha channel results in a dummy picture being loaded.
I searched the docs and havent found your hint regarding RGBA as image_mode. It only says ""RGB" if color, or "L" if black and white." at https://gradio.app/docs/#image

PR #2372, opened 2022-09-29
Small fixes to docs for `Image` component

abidlabs:
I noticed that the string shortcuts for some of the new variants of `gr.Image` like `gr.Paint()` were not showing up correctly because of a missing colon:

image

This PR fixes that and also fixes two other outstanding issues with the docs, fixes: #2040, fixes: #134. Now:

image


aliabd (approved):
Nice fix @abidlabs, I must have missed this. Also love the new format with the template class name. 

LGTM

abidlabs:
Thanks for the fast review @aliabd!

PR #2374, opened 2022-09-29
Misc queue changes test

freddyaboulton:
# Description

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

Closes: # (issue)

# Checklist:


abidlabs:
Thanks for making the fix @freddyaboulton. IIRC the reason `gather_data_for_first_ranks()` existed was to reduce the amount of time needed to send the input data between the browser and server. Instead of waiting until it was a user's turn, the queue would request data from users while they were waiting on the queue. I don't think we should remove it as the latency for communicating e.g. image data could be quite high

freddyaboulton:
@abidlabs Maybe the right next step is to performance test this branch with main with image, audio, and text inputs? 

This is a hard issue to repro so it's hard to know what the best fix is. The SD space is patched for now so we have some time to think of the best fix.

Regarding the latency though, why not just send the data when the client first connects to the queue? I feel like we originally did not do that because the queue size was unbounded but I think most queues implement a max size anyways.

aliabid94:
I don't think we should collect data from all users as soon as they join the queue, then we have no upper bound on the amount of data in memory used for the queue.

FarukOzderim:
Could you share the details of the race condition @freddyaboulton, from the design perspective it feels like it should not go into a race condition, or it should not affect anything since both write the same data.

1. We could sleep for sleep_when_free after starting background jobs to prevent this softly
2. Or await gather_data_and_broadcast instead of running it in the background.
3. Or use a state variable inside events, so that we know there is an ongoing communication on that event, and wait until that state is unlocked.

abidlabs:
It seems that there are a couple of reasons to use `gather_data_for_first_ranks()`: reduce memory for large queue sizes and reduce latency in general. I think the right thing to do would be to understand why the race condition is happening and eliminate that logic or use a lock if we need to. WDYT @freddyaboulton?

FarukOzderim:
Theoretically this design should not have any race conditions, since the async code is not parallel but concurrent. Hence I don't understand why this kind of situation could appear, there might be something else as well. Let me leave some suggestions below:

**Short term**: Disable early data loading to resolve stuck queue problem (This PR)
**Mid Term**: Analyze the situation at dead-lock and see which parts of the code goes into conflict.
**Long term**: Solve the situation in mid-term or, design a communication that will change the data field only in a single part of the code.
an overview:
```
a. if the rank is <data_gathering_threshold when it joines the queue, directly ask for data in route
b. remove data gathering in process_event, and sleep for some time if the data is not available yet.
If the data is still not available either drop it or put it at the end of the queue.
c. have a variable for data-communication called data_communication_active. Use this in gather_data_and_broadcast_estimations and a. as a lock.
If it is not available at the start of gather_data_and_broadcast_estimations, directly return the function. Because this means communication did not finish yet, and we can skip communication on that turn.
```

let me know if anything is unclear or you don't agree with

aliabid94:
I think this makes sense as a fix for now. Another quickfix possibility is getting user data right when the user joins the queue, rather than when it is their turn for their data to be processed  - should save some time. Will dive into a more perfect fix but I think this can be merged for now.

abidlabs:
Before we merge I think we need to test to make sure this doesn't significantly increase inference latency? Particularly with heavy inputs such as images and videos? My worry is that it might increase the inference latency by ~1s roundtrip on Spaces

freddyaboulton:
I can test tomorrow with a smaller batch size so that the current queue implementation doesnt get stuck like it does with a batch size of 50 and we can compare the latency between the two.

FarukOzderim:
@aliabid94 I have a question, does frontend send data as many times server asks for it?

I think I found the cause behind the problem. It seems to me that frontend responds to send_data only once, and this causes problems when there are several `send_data` calls.

aliabid94:
Actually the frontend is capable of sending the `send_data` event multiple times, I just tested this. So our assumption (or maybe just my assumption) that calling `send_data` twice causes the error might be wrong, having `send_data` called twice might be the side effect of the actual issue. But the frontend is capable of handling `send_data` multiple times.

freddyaboulton:
@aliabid94 @abidlabs I just rewrote the benchmark script slightly so that the whole script doesn't hang when one of the requests gets stuck due to the race condition. That lets us get data to compare between the two implementations. 

I'm also using a smaller batch size of 30. That should let us get a good sense of the performance penalty of this PR. Since the benchmark space has `client_position_to_load_data` of 30 and `concurrency_count` of 20, requests 21 - 30 should get their data before it's their turn which should be a speedup as compared to this PR.

Here are the numbers. There is a penalty of about 0.1 seconds for audio, image, text. I think that's something we can improve in the future but not something that should hold up this PR.

# Current implementation
`python benchmark_script_gather.py wss://spaces.huggingface.tech/gradio/queue-benchmark/queue/join`
| Input type | Average time to complete prediction over 100 requests |
| ------------- | ------------- |
| Audio  |1.0815931215818348 |
| Image  |0.9865580356538528 |
| Text |0.9900706960954739|
| Video |1.7919933594042254|

# This PR
`python benchmark_script_gather.py wss://spaces.huggingface.tech/gradio/queue-benchmark-proposed/queue/join`

| Input type | Average time to complete prediction over 100 requests |
| ------------- | ------------- |
| Audio  |1.1740408795220512|
| Image  |1.1396675520929798 |
| Text |1.1253375146903244|
| Video |1.4886653151938585|


This is the script:

```python
import asyncio
from datetime import datetime
import json
import random
import time
import websockets
import gradio as gr
import pandas as pd
import argparse

FN_INDEX_TO_DATA = {
}

results = []

        start = time.time()
        try:
            while True:
                msg = json.loads(raw_msg)
                msg = msg["msg"]
                if msg == "queue_full":
                    break
                elif msg == "send_data":
                elif msg == "estimation":
                    pass
                elif msg == "process_starts":
                    pass
                    end = time.time()
                    duration = end - start
                    async with asyncio.Lock():
                        results.append({"fn_index": name, "duration": duration})
        except:
            pass


    running_tasks = set([])

    for _ in range(50):
        await asyncio.sleep(0.01)
        task = asyncio.create_task(get_prediction(host))
        running_tasks.add(task)
        task.add_done_callback(running_tasks.discard)

    while True:
        await asyncio.sleep(0.01)
        if len(results) >= n_results:
            break
        if len(running_tasks) < 50:
            for _ in range(len(running_tasks), 50):
                task = asyncio.create_task(get_prediction(host))
                running_tasks.add(task)
                task.add_done_callback(running_tasks.discard)

    data = pd.DataFrame(results[:n_results]).groupby("fn_index").agg({"mean"})
    data = data.reset_index()
    return data


if __name__ == "__main__":
    args = parser.parse_args()
    print(out)
```

aliabid94:
Seems like negligible overhead, thanks for the benchmarking work! Let's get this in - I think there was another space that got stuck today.

abidlabs:
Agreed, thanks for benchmarking this @freddyaboulton! LGTM and we can close the alternative PRs (#2399 and #2420)

freddyaboulton:
Thank you for the reviews!

Issue #2371, opened 2022-09-29
New typeable slider doesn't respect the `minimum` and `maximum` values


apolinario:
### Describe the bug

Typable sliders are amazing! However they are currently not respecting the minimum and maximum values of the slider. It works for clicking the arrow, but if the user types, they can type an arbitrary number

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

![image](https://user-images.githubusercontent.com/788417/193131281-33b8d7fb-ba92-475d-bc4b-042895dbb82a.png)

Any slider with `maximum` or `minimum` values can have that abused by typing a value outside of the intended range. This is the [example](https://huggingface.co/spaces/lambdalabs/text-to-pokemon) where the range should be 1-4. 5 would make it OOM

### Logs

```shell
No logs
```

### System Info

```shell
Gradio==3.4
```

### Severity

serious, but I can work around it

abidlabs:
Good catch @apolinario! @dawoodkhan82 would you be able to take a look?

PR #2329, opened 2022-09-24
Typeable value on gradio.Slider

dawoodkhan82:
# Description
Allows slider value to be typeable.

https://user-images.githubusercontent.com/12725292/192082923-40c7f4dc-2e67-4c7c-aa0d-10a6cfa757d2.mov

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 
* any additional dependencies that are required for this change.

Closes: #936

# Checklist:

- [ ] I have performed a self-review of my own code
- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #2380, opened 2022-10-03
Clamps values for typeable slider input when out of range

dawoodkhan82:
# Description
Shows warning popup on slider textbox when value is out of range

https://user-images.githubusercontent.com/12725292/193510091-e8fa212e-91a9-40ae-a37d-4b12a63bd963.mov

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

Closes: #2371 

# Checklist:


dawoodkhan82:
TODO: Stop from submitting values out of range.

pngwn:
I think it would be better to just clamp the value as it requires less effort from the user, which would also stop the form from submitting the wrong value. 
- if > max set to max
- if < min set to min

Something like this will do it:

```js
const clamp = (num, min, max) => Math.min(Math.max(num, min), max);
```

I think setting the min/max on the input (as this PR does) is a good idea too.

abidlabs:
@dawoodkhan82 and I chatted on Friday about clamping and our initial concern was that someone might be in the process of typing out a larger value and the "intermediate value" could get clamped if it is lower than the slider minimum. But if the clamping only happens if a user unfocuses from the typing box, then it should be fine

abidlabs:
Side note: given https://github.com/gradio-app/gradio/pull/2329#issuecomment-1265434011, we may want to give users a way to disable the typeable box.

apolinario:
I also agree with clamping - if we can find a good UX for it - the clamping after focus only sounds good to me) - as just a warning may not be enough if someone is deliberately interested in going over the application limits

dawoodkhan82:
Updated so that on blur (input box is unfocused), the value is set to the min or max when typeable value is out of range 


https://user-images.githubusercontent.com/12725292/193939053-2115fb55-c6dc-4e2b-9ea3-2fad0724f36b.mov

pngwn (approved):
looks great!

abidlabs (approved):
Good stuff! Let's merge it in?

PR #2396, opened 2022-10-05
FRP Poc

XciD:
FRP Proof of concept

Currently working with `gradio.proxy.huggingface.tech`
I'm running it in a EC2 with docker.

Implementation was not easy, frp use their own protocol to communicate.
I've implemented the basics.

I had to remove the encryption/decryption (we need to check how to implement it with python) ( see https://github.com/huggingface/frp/pull/1)

Things to do / check:

- ctrl-c is not working correctly, I think that threading is the issue.
- we need to handle reconnection, if for some reason the server reboot, we should be able to reconnect and keep the same id ?
- we may want to disable some feature on server side to constraint client

Nota bene: sorry for code quality, python is not my language of choice ;D


julien-c:
On the Python/Gradio side, maybe a good idea for @wauplin to take a look (maybe pairing with @freddyaboulton) at the Python side?

But @abidlabs will know better =)

abidlabs:
Taking a look now, but would definitely appreciate a review by @Wauplin as I'm not super confident around all the tunneling and socket stuff

Wauplin:
Hi @XciD @abidlabs, as said [on slack](https://huggingface.slack.com/archives/C045FUG7Z9U/p1665396056804599) (private conv), it overall looks good to me apart from some python styling stuff.

The main problem seems to be with threads and how exceptions are caught (*ctrl-c is not working correctly, I think that threading is the issue.* => I'm pretty sure it is as well). In Python exceptions in a thread are not propagated to the main thread so if you don't pay attention, they are silently ignored. The child thread stops but not the main one nor the other children. 

One way to solve that is to keep pointers to your active threads, check when one fails and cancel the others on exception (especially ctrl+c). In a [totally unrelated issue](https://github.com/huggingface/huggingface_hub/pull/888#discussion_r892029574) in `huggingface_hub` I described a workaround for that.

Otherwise I think it's worth to try the asyncio API that has a quite high-level api to work with streams/sockets. Would be interesting to look into [asyncio.open_connection](https://docs.python.org/3/library/asyncio-stream.html#asyncio.open_connection). I can give an help here if needed/interested. Main advantages I see is in the case of opening a lot of tunnels and to handle exception nicely. Otherwise it can be superficial compared to having an already working solution. 

In any case, please let me know how/where I can help best here :) (@abidlabs you mentioned to present me gradio a bit more in-depth, that could be a good start).

(side note: I don't have access to https://github.com/huggingface/frp/pull/1 mentioned above)

julien-c:
> (side note: I don't have access to [huggingface/frp#1](https://github.com/huggingface/frp/pull/1) mentioned above)

Should be good now (@XciD, added the usual team teamBG to that repo settings)

Wauplin:
> Should be good now
It is, thanks :)

Wauplin:
I made some tests and create a PR from my fork: https://github.com/XciD/gradio/pull/1. I'll let you merge it so that you can see the changes. I mainly changes the exception handling.

Wauplin:
Side note: when a shared link is not active anymore, we receive a default message from the FRP server. I think we could change it to something more Gradio-related like "this is an expired link, gradio app is not running anymore. Please do this or that." Example: https://e7305353-6018-4811-aa60-a30bc0adc120.gradio.proxy.huggingface.tech/

frp

Wauplin:
@XciD @abidlabs Actually I tried again and now get a weird issue. I am using the basic "hello world" example from Gradio

```py
import gradio as gr

def greet(name):
    return "Hello " + name + "!"

demo.launch(share=True)
```

When I run it on `main`, it works fine. When I run it on `XciD:frp-poc` or `Wauplin:wauplin-frp-poc-exception-handler-v2` (my fork) the shared link is generated, I can access it but the output block doesn't show the "hello Lucain" message. However if I open the developer settings I can see that the message is well returned by the server so it doesn't seem to be a tunnel issue... (see screenshot below)

[FRP_issue.webm](https://user-images.githubusercontent.com/11801849/196212305-266186ce-412e-48a4-8930-6c6179956362.webm)

XciD:
> 

Thanks lucain. Feel free to merge you PR on mine ;)

abidlabs:
Going to release this as `gradio==3.6b10` so we can easily test on Colab, etc.

abidlabs:
I'm going to merge this into a branch called `frp` (instead of a fork) so that we can run specific CI actions on it (that aren't possible with forks)

PR #2399, opened 2022-10-06
Split queues to fix race condition

aliabid94:
This is an approach to fix the race condition in our event queue by splitting the event_queue into 3 queues:

1) event_queue_no_data (these events have no data yet)
2) event_queue_gathering_data (these events are gathering data)
3) event_queue_with_data (these events have gathered their data)

Events move across the queues in the above listed order. 
- `start_processing` finds how many events to move from 1 -> 2 based on `data_gathering _start`. 
- Then `gather_data_for_first_ranks` moves this many events from 1 -> 2, requests data, then moves from 2 -> 3.  
- `start_processing` pops off events only from 3 into `active_jobs` if there is space in `active_jobs`.

### DIFF
--- a/demo/calculator/run.py
+++ b/demo/calculator/run.py
@@ -1,6 +1,8 @@
 import gradio as gr
+import time
 
 def calculator(num1, operation, num2):
+    time.sleep(10)
     if operation == "add":
         return num1 + num2
     elif operation == "subtract":
@@ -29,5 +31,6 @@ demo = gr.Interface(
     title="Toy Calculator",
     description="Here's a sample toy calculator. Enjoy!",
 )
+demo.queue(client_position_to_load_data=3)
 if __name__ == "__main__":
     demo.launch()
--- a/gradio/queue.py
+++ b/gradio/queue.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import json
+import queue
 import time
 from typing import Dict, List, Optional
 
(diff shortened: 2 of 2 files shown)


freddyaboulton:
@aliabid94 The benchmark script is also getting stuck on this PR

```python
import asyncio
from datetime import datetime
import json
import random
import time
from typing import List
import websockets
import gradio as gr
import pandas as pd
import argparse

FN_INDEX_TO_DATA = {
}


        start = time.time()
        while True:
            msg = json.loads(raw_msg)
            msg = msg["msg"]
            if msg == "queue_full":
                return
            elif msg == "send_data":
            elif msg == "estimation":
                pass
            elif msg == "process_starts":
                pass
                end = time.time()
                duration = end - start
                return {"fn_index": name, "duration": duration}


    results = []
    while len(results) < n_results:
        print(f"N results {len(results)}")
        batch_results = await asyncio.gather(*[get_prediction(host) for _ in range(20)])
        for result in batch_results:
            if result:
                results.append(result)

    data = pd.DataFrame(results).groupby("fn_index").agg({"mean"})
    data = data.reset_index()
    return data


if __name__ == "__main__":
    args = parser.parse_args()
    print(out)
```

Run with

`python benchmark_script_gather.py wss://spaces.huggingface.tech/gradio/queue-benchmark-three-queues/queue/join`

Been stuck like this for a couple of minutes:

image

PR #2405, opened 2022-10-07
Automatically restart spaces if they're down

aliabd:
This PR adds a script that will check if any of the spaces in the /demos tab are down, and if they are it will restart them. This is important because spaces have gone down recently, especially when we programmatically update many of them at the same time. And often they can be down for a while before we notice. 

Once it's merged I will create a cronjob in the server to run it every 15mins. 

Also, stopped the website build from failing if the AUTH_TOKEN env variable is missing, so that we can build it locally, like @freddy suggested. 

### DIFF
--- a/website/homepage/src/demos/restart_demos.py
+++ b/website/homepage/src/demos/restart_demos.py
@@ -1,0 +1,10 @@
+from upload_demos import demos_by_category, upload_demo_to_space, AUTH_TOKEN, gradio_version
+from gradio.networking import url_ok
+
+
+for category in demos_by_category:
+    for demo in category["demos"]:
+        space_id = "gradio/" + demo["dir"]
+        if not url_ok(f"https://hf.space/embed/{space_id}/+"):
+            print(f"{space_id} was down, restarting")
+            upload_demo_to_space(demo_name=demo["dir"], space_id=space_id, hf_token=AUTH_TOKEN, gradio_version=gradio_version)
--- a/website/homepage/src/demos/upload_demos.py
+++ b/website/homepage/src/demos/upload_demos.py
@@ -68,8 +68,9 @@ def upload_demo_to_space(
     return f"https://huggingface.co/spaces/{space_id}"
 
 if __name__ == "__main__":
-    for category in demos_by_category:
-        for demo in category["demos"]:
-            space_id = "gradio/" + demo["dir"]
-            upload_demo_to_space(demo_name=demo["dir"], space_id=space_id, hf_token=AUTH_TOKEN, gradio_version=gradio_version)
+    if AUTH_TOKEN is not None:
+        for category in demos_by_category:
+            for demo in category["demos"]:
+                space_id = "gradio/" + demo["dir"]
+                upload_demo_to_space(demo_name=demo["dir"], space_id=space_id, hf_token=AUTH_TOKEN, gradio_version=gradio_version)
 
--- a/website/homepage/src/demos/upload_demos.py
+++ b/website/homepage/src/demos/upload_demos.py
@@ -67,8 +67,9 @@ def upload_demo_to_space(
         )
     return f"https://huggingface.co/spaces/{space_id}"
 
-for category in demos_by_category:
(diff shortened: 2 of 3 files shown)


website/homepage/src/demos/restart_demos.py
```diff
@@ -0,0 +1,10 @@
+from upload_demos import demos_by_category, upload_demo_to_space, AUTH_TOKEN, gradio_version
+from gradio.networking import url_ok
+
+
+for category in demos_by_category:
+    for demo in category["demos"]:
+        space_id = "gradio/" + demo["dir"]
+        if not url_ok(f"https://hf.space/embed/{space_id}/+"):
+            print(f"{space_id} was down, restarting")
+            upload_demo_to_space(demo_name=demo["dir"], space_id=space_id, hf_token=AUTH_TOKEN, gradio_version=gradio_version)
```
abidlabs, line 10: Will this trigger a restart if this files haven't changed?

website/homepage/src/demos/upload_demos.py
```diff
@@ -67,8 +67,10 @@ def upload_demo_to_space(
         )
     return f"https://huggingface.co/spaces/{space_id}"
 
-for category in demos_by_category:
-    for demo in category["demos"]:
-        space_id = "gradio/" + demo["dir"]
-        upload_demo_to_space(demo_name=demo["dir"], space_id=space_id, hf_token=AUTH_TOKEN, gradio_version=gradio_version)
+if __name__ == "__main__":
+    if AUTH_TOKEN is not None:
+        for category in demos_by_category:
```
abidlabs, line 72: Looks like this includes all of the demos in `recipe_demos.json` but what about demos in other parts of the website like the docs and homepage?

abidlabs:
Script looks good -- couple of clarification questions @aliabd

website/homepage/src/demos/restart_demos.py
aliabd: Yes, which is intended. We don't only restart demos when files have changed. We also restart them when the version is changed. This just restarts them if they're down.

abidlabs: Agreed. My question is, will this be *sufficient* to restart the Space? In other words, if you push to a Space and the files haven't changed, will it actually restart the Space?

abidlabs: Agreed. My question is, will this be *sufficient* to restart the Space? In other words, if you push to a Space and the files haven't changed, will it actually restart the Space?

website/homepage/src/demos/restart_demos.py
```diff
@@ -0,0 +1,10 @@
+from upload_demos import demos_by_category, upload_demo_to_space, AUTH_TOKEN, gradio_version
+from gradio.networking import url_ok
+
+
+for category in demos_by_category:
+    for demo in category["demos"]:
+        space_id = "gradio/" + demo["dir"]
+        if not url_ok(f"https://hf.space/embed/{space_id}/+"):
```
julien-c, line 8: note that those URLs will change soon-ish :) 

(but when they change, they will be then considered "public API" ie. they won't change again after that)

website/homepage/src/demos/restart_demos.py
aliabd: Oh got it @abidlabs, yes it will.

website/homepage/src/demos/restart_demos.py
aliabd: Thanks for clarifying @julien-c, will make sure to change this when the urls change. 

Do you know if there's a way in the api to programmatically check if the space has a build error? Instead of the way I'm doing it now which is checking if the url returns 200

abidlabs (approved):

website/homepage/src/demos/restart_demos.py
julien-c: yes there's a way 

Stage should be exposed in `SpaceInfo` in `huggingface_hub`, for instance

PR #2420, opened 2022-10-08
Await gather_data to fix queue getting stuck

FarukOzderim:
Tries to fix #2374

### DIFF
--- a/gradio/queue.py
+++ b/gradio/queue.py
@@ -23,12 +23,12 @@ class Estimation(BaseModel):
 
 class Queue:
     def __init__(
-      self,
-      live_updates: bool,
-      concurrency_count: int,
-      data_gathering_start: int,
-      update_intervals: int,
-      max_size: Optional[int],
+        self,
+        live_updates: bool,
+        concurrency_count: int,
+        data_gathering_start: int,
+        update_intervals: int,
+        max_size: Optional[int],
     ):
         self.event_queue = []
         self.events_pending_reconnection = []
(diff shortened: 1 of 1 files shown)


FarukOzderim:
@freddyaboulton could you test the queue with the latest changes, whether if it gets stuck or not?

https://huggingface.co/spaces/gradio-pr-deploys/pr-2420-all-demos

freddyaboulton:
@FarukOzderim For sure! Will test tomorrow. Thank you for the help 😊 

One question though - shouldn’t we await gather_data before starting process_event in the background? If we don’t, I think the same problem  can happen where the background function sends the ‘send_data’ message when gather_data coroutine is running.

FarukOzderim:
I don't think that will happen.

Because if event.data is None that means gather_data was not awaited for that event, ie. It came into an empty queue.

If gather_data is awaited for that event it means event.data is not None, so no network call be realized.

Advantage of this is, process and gather_data runs concurrently.

freddyaboulton:
@FarukOzderim I think this is still getting stuck. I have tried twice and the script is not able to complete. This is what I'm running.

```python
import asyncio
from datetime import datetime
import json
import random
import time
from typing import List
import websockets
import gradio as gr
import pandas as pd
import argparse

FN_INDEX_TO_DATA = {
}


        start = time.time()
        while True:
            msg = json.loads(raw_msg)
            msg = msg["msg"]
            if msg == "queue_full":
                return
            elif msg == "send_data":
            elif msg == "estimation":
                pass
            elif msg == "process_starts":
                pass
                end = time.time()
                duration = end - start
                return {"fn_index": name, "duration": duration}


    results = []
    while len(results) < n_results:
        print(f"N results {len(results)}")
        batch_results = await asyncio.gather(*[get_prediction(host) for _ in range(20)])
        for result in batch_results:
            if result:
                results.append(result)

    data = pd.DataFrame(results).groupby("fn_index").agg({"mean"})
    data = data.reset_index()
    return data


if __name__ == "__main__":
    args = parser.parse_args()
    print(out)
```

freddyaboulton:
Thank you @FarukOzderim ! Closing for now but we might revisit again in the future.

PR #2433, opened 2022-10-11
Cancel events from other events

### LINKED ISSUE #2220: Pressing "clear" should interrupt and clear an iterative output
Currently, when you have an iterative output like this:

```py
import gradio as gr
import numpy as np
import time


def fake_diffusion(steps):
    for _ in range(steps):
        time.sleep(1)
        image = np.random.random((600, 600, 3))
        yield image
    
    image = "https://i.picsum.photos/id/867/600/600.jpg?hmac=qE7QFJwLmlE_WKI7zMH6SgH5iY5fx8ec6ZJQBwKRT44" 
    yield image


demo = gr.Interface(fake_diffusion, 
                    inputs=gr.Slider(1, 10, 3), 
                    outputs="image")
demo.queue()

if __name__ == "__main__":
    demo.launch()
```

If you press the "Clear" button while the output is iterating, it momentarily clears, but then the output continues iterating. This is not the expected behavior -- instead the output should remain clear and allow the user to resubmit a new input.

freddyaboulton:
# Description
Fixes #2220 

This PR allows you to cancel other running events when another event is triggered. The functionality for interface has not been implemented yet. Opening this up as a draft so others can test this out before proceeding.

### Changes to user-facing API
Adds a `cancels` parameter to all event handlers. Users can specify a list of events to cancel.
```python
buton.click(prediction, inputs=[...], outputs=[...], cancels=[other_prediction])
```

### Limitations
1. The queue must be enabled for this to work.

### Example you can use to test this
Build the front-end first.
```python
import time
import gradio as gr

def fake_diffusion(steps):
    for i in range(steps):
        print(f"Current step: {i}")
        time.sleep(1)
        yield str(i)

def long_prediction(*args, **kwargs):
    time.sleep(10)
    return 42


with gr.Blocks() as demo:
    with gr.Row():
        with gr.Column():
            n = gr.Slider(1, 10, value=9, step=1, label="Number Steps")
            run = gr.Button()
            output = gr.Textbox(label="Iterative Output")
            stop = gr.Button(value="Stop Iterating")
        with gr.Column():
            textbox = gr.Textbox(label="Prompt")
            prediction = gr.Number(label="Expensive Calculation")
            run_pred = gr.Button(value="Run Expensive Calculation")
        with gr.Column():
            cancel_on_change = gr.Textbox(label="Cancel Iteration and Expensive Calculation on Change")
            cancel_on_submit = gr.Textbox(label="Cancel Iteration and Expensive Calculation on Submit")
            echo = gr.Textbox(label="Echo")
    with gr.Row():
        with gr.Column():
            image = gr.Image(source="webcam", tool="editor", label="Cancel on edit", interactive=True)
        with gr.Column():
            video = gr.Video(source="webcam", label="Cancel on play", interactive=True)

    click_event = run.click(fake_diffusion, n, output)
    stop.click(fn=None, inputs=None, outputs=None, cancels=[click_event])
    pred_event = run_pred.click(fn=long_prediction, inputs=[textbox], outputs=prediction)

    cancel_on_change.change(None, None, None, cancels=[click_event, pred_event])
    cancel_on_submit.submit(lambda s: s, cancel_on_submit, echo, cancels=[click_event, pred_event])
    image.edit(None, None, None, cancels=[click_event, pred_event])
    video.play(None, None, None, cancels=[click_event, pred_event])


demo.queue(concurrency_count=20, max_size=20).launch()
```

### Cancel on click

cancel_on_button

### Cancel on change
cancel_on_change

### Cancel on submit
cancel_on_submit

### Cancel on edit
cancel_on_edit

### Cancel on play
cancel_on_play

# Checklist:

# A note about the CHANGELOG


dawoodkhan82:
@freddyaboulton maybe we can consolidate the api with #2436

freddyaboulton:
> @freddyaboulton maybe we can consolidate the api with #2436

@dawoodkhan82 Do you have something in mind?

gradio/blocks.py
```diff
@@ -119,7 +119,8 @@ def set_event_trigger(
         js: Optional[str] = None,
         no_target: bool = False,
         queue: Optional[bool] = None,
-    ) -> None:
+        cancels: List[int] | None = None,
```
abidlabs, line 122: Add `cancels` to docstring below

gradio/events.py
```diff
@@ -22,6 +76,7 @@ def change(
         queue: Optional[bool] = None,
         preprocess: bool = True,
         postprocess: bool = True,
+        cancels: List[Dict[str, Any]] | None = None,
```
abidlabs, line 79: Add `cancels` to docstring for all events

gradio/routes.py
```diff
@@ -254,6 +259,16 @@ def file(path: str):
         def file_deprecated(path: str):
             return file(path)
 
+        @app.post("/reset/")
+        @app.post("/reset")
+        async def reset_iterator(body: ResetBody):
+            if body.session_hash not in app.iterators:
+                return {"success": False}
+            async with asyncio.Lock():
```
abidlabs, line 267: I don't think `async with asyncio.Lock():` will have any effect here since each request is instantiating a new `asyncio.Lock()` object. A single `asyncio.Lock()` object should be instantiated outside of the this function and that object should be acquired here. https://docs.python.org/3/library/asyncio-sync.html

gradio/events.py
```diff
@@ -1,14 +1,68 @@
...
+
+
+def set_cancel_events(block: Block, event_name: str, cancels: List[Dict[str, Any]]):
+    if cancels:
+        cancel_fn, output, fn_indices_to_cancel = get_cancel_function(cancels)
+        block.set_event_trigger(
+            event_name,
+            cancel_fn,
+            inputs=None,
+            outputs=output,
```
abidlabs, line 59: Out of curiosity, do we actually need to supply the outputs?

abidlabs:
Hey @freddyaboulton this is a really really cool implementation! 

I was testing with the following script and everything seems to work well:

```py
import gradio as gr
import time

def run(x):
    time.sleep(3)
    return x

def run2(x):
    for i in range(3):
        time.sleep(1)
        yield i

with gr.Blocks() as demo:
    x = gr.Textbox()
    y = gr.Textbox()
    b = gr.Button()
    b2 = gr.Button()
    c = gr.Button("Cancel")
    e = b.click(run, x, y)
    e2 = b2.click(run2, x, y)
    c.click(None, None, None, cancels=[e, e2])
    
    demo.queue()
    demo.launch()
```

However, one piece of feedback is that currently, when you "cancel" an event, the output component that was being populated *clears* (i.e. its value gets reset). I don't think this should happen -- rather, the final value should remain there. For example, if someone is generating a sequence of images, they may want to "stop" on a particular image. WDYT?

abidlabs:
This might close https://github.com/gradio-app/gradio/issues/1323 as well?

pngwn:
@freddyaboulton looks good to me! I like the extra `variant` (will be helpful for theming).

Could we add `"cancels"` to the union in the button component defined here: https://github.com/gradio-app/gradio/blob/main/ui/packages/button/src/Button.svelte#L8

abidlabs:
Fantastic PR @freddyaboulton! 

One tiny thing I noticed was that when running in a colab / jupyter notebook, because of the iframe's width, the stop button is pushed to the next row:

image

We could either (1) increase the iframe width, but it seems like you need an iframe width of at least 1280 for it to render correctly, which itself can cause the Interface to be cut off:

image

Or (2) we could move it next to the "Flag" button: 

image

freddyaboulton:
@abidlabs Thanks for the catch about rendering in the notebook! I'm worried about putting it under the output component because if it's really long (like an image in diffusion or an RL env frame) the button might be pushed outside the screen.

For example, look at the atari agent space:

image


I like the idea of keeping the stop button in the same group as submit and don't mind it's on the next row. Maybe we keep as is for now? Happy to jump on a huddle to discuss further.

abidlabs:
Ok I think that’s fine. LGTM thank you!

abidlabs (approved):

freddyaboulton:

dawoodkhan82:
@freddyaboulton Good stuff! Just tested it, and works great.

Seedmanc:
How to make an event cancel itself?
When I'm trying to do run = button.click(..., cancels=[run]), it's not yet defined at that moment.

taoari:
if an event can be cancelled, is it possible to recover it? For example, in #1323, after we stop the streaming, how can we resume it?

abidlabs:
@taoari it is not possible to resume an event after it has been canceled. If you have a job that is running, perhaps you could store the intermediate results in a `gr.State()` and then your event trigger can read the value of the `gr.State()` and resume from there?

taoari:
@abidlabs Thank you for your suggestion! But currently, I would like to enable webcam streaming again for new results, so it can not store intermediate results in a `gr.State()`.

Issue #1954, opened 2022-08-05
When multiple events are triggered consecutively, the final event should always be sent


freddyaboulton:
### Describe the bug

If you have a live interface, the predict function sometimes is not run for every update of the component - especially if the predict function has some delay and the updates happen faster than the runtime of the predict function.

You can repro with this interface

```python
import gradio as gr
import time

def print_with_delay(x):
    time.sleep(0.2)
    return x

gr.Interface(print_with_delay, "text", "text", live=True).launch()
```

Note that I typed "freddy" but the text box only shows "fred".

![live_interface_delay](https://user-images.githubusercontent.com/41651716/183110511-253137cb-37f3-4727-b8d9-ad9d30fe0222.gif)

A more realistic example is https://huggingface.co/spaces/adirik/OWL-ViT , where running with `live=True` will make the object detection query not contain all the key words.

https://user-images.githubusercontent.com/41651716/183110769-c907cc8b-6c83-48ac-b472-5b6b6e241cfa.mp4

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

-

### Logs

```shell
-
```

### System Info

```shell
3.1.3
```

### Severity

serious, but I can work around it

abidlabs:
Yup! The problem is that is a new prediction isn't sent until the previous one is complete, which can create a very "non-live" experience, especially on Spaces and Colab.

I know we considered adding a debounce to the Textbox and Slider components for this reason. Should we add it here @pngwn @aliabid94?

pngwn:
We did have a debounce but we removed it because it caused some bugs. https://github.com/gradio-app/gradio/pull/932

I think the correct solution here is to send all requests and ensure the most recent request/ action takes priority over earlier ones. This will require some changes to how we are processing request both on the frontend and the backend.

In general with live updates we need to reduce latency as much as possible, using websockets for everything would provide the technical foundation for this kinda of responsiveness. Something I mentioned recently on slack.

pngwn:
Copied from response in slack:

Live updates need to be looked at holistically anyway imo, its a really important usecase but Gradio doesn't really do a great job right now because of how we process requests, and how we handle loading states. Queueing also introduces some additional challenges here, at least from a design PoV.

Loading states in general are pretty rudimentary right now and only really support the classic `action -> run prediction -> wait -> display` flow. They aren't suited to these kinds of real time updates or for streamed updates. None of these issues are impossible to resolve, they just require some thought and a little engineering effort.

[...]

Issue #1453, opened 2022-06-02
Blocks Slider change() outputs incorrect value on drag


mindwrapped:
### Describe the bug

If you click and drag on the slider then it will pass an incorrect value to change(). 

If you only click on the slider then it will pass the correct value to change().

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

https://huggingface.co/spaces/mindwrapped/blocks-slider-test 

Click and drag -> incorrect value (old value, not the current one displayed on the slider)
Click -> correct value (same value as on the slider)

```python
import gradio as gr

def update_value(val):
  return f'Value is set to {val}'

demo = gr.Blocks()

with demo:
  inp = gr.Slider(0, 100, label='Value')
  md = gr.Markdown('Select a value')

  inp.change(fn=update_value, inputs=inp, outputs=md)

demo.launch()
```

### Screenshot

![Screen Shot 2022-06-02 at 1 11 00 PM](https://user-images.githubusercontent.com/61097415/171719313-c4325208-8482-45ec-9983-e565a3e036cf.png)

### Logs

```shell
From HF Spaces:

Running on local URL:  http://localhost:7860/

To create a public link, set `share=True` in `launch()`.
```

### System Info

```shell
Huggingface Spaces
Gradio: 3.0.10
```

### Severity

blocking upgrade to latest gradio version

cyyeh:
I found this issue happened when it is deployed using Spaces, there's no issue if I run this application locally.

aliabid94:
Will fix, thanks for raising this issue!

NimaBoscarino:
I'm running into this as well on Gradio 3.3. Here's a notebook to reproduce: https://colab.research.google.com/drive/1JoDcK5R8lcA7Ty4-zlXnQNBwKXdfgRpb?usp=sharing

Same as mentioned above, works fine locally but on Colab + Spaces it's broken.

abidlabs:
This should be solved by #2436, so moving to the 4.0 milestone

abidlabs:
Special case of #1954, so closing this issue

axkoenig:
Hi there, 
I think the issue still persists when there is some lag introduced inside the `update_value` function mentioned above. 
Consider the below example:
```python
import gradio as gr

def update_value(val):
  import time
  time.sleep(2) # simulate ML model inference
  return f'Value is set to {val}'

demo = gr.Blocks()

with demo:
[...]

PR #2436, opened 2022-10-12
Change behavior of pending event triggers

dawoodkhan82:
# Description

Change default behavior of event triggers to "wait" for pending requests to complete. Fixes issues related textbox change events not registering latest change. ex: #1954 

TODO: make this work for queue

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

Closes: #1954 

# Checklist:

# A note about the CHANGELOG


abidlabs:
Interesting, some question I have:
* What does "existing pending request" mean? Does it mean when there is (A) any request that the backend is processing, (B) any request from the same user that the backend is processing, or (C) any request from the same user and same type that the backend is processing
* What will be the default `behavior_when_pending`? I think most users will expect the default to be `"wait"`. I'm not sure when you'd want the other two -- can you explain the use cases for those options?

pngwn:
I don't think we want both behaviours, and I don't really like the idea of adding more API surface area to solve this. I think we should just honour the latest request at any given moment (letting them all through), rather than rejecting subsequent request until we have a response.

- req 1 
- req 2 
- req 3 
- res 1 (ignore)
- res 2 (ignore)
- res 3 (handle)

Now it would also be nice if we could 'abort' those processes on the backend where possible too but depending on the stage they are at that might not be possible. If the queue is active we could remove them from the queue though.

You can't _really_ abort requests in the browser but we could signal to the server that it should ignore them if they are not already being processed. I think we'll need to assign ids or hashes to requests though because, especially for the queue otherwise we won't know what response we are dealing with.

aliabid94:
I think there are clear cases where we want each of these behaviors:
- `block`: this is the current default behavior. In an interface with a heavy model, you usually don't want the user to be able to click Submit while the previous request is pending.
- `run`: Users may wish to allow multiple requests to be triggered. You may wish to spin off multiple backend processes without waiting for the previous process to finish.
- `wait`: this is very useful for "change" event listeners (the original reason for this PR). If a textbox has a change event listener, you don't want to fire on every single letter if the last pending request has not completed (as would happen in `run` mode). But `block` mode is not appropriate either: if a user types letter "A" (which fires a request), then types two more letters "BC" while the request is pending, those letters will not trigger a request. Later after the request completes, the user types "D", then this will trigger a change with the value "ABCD". However, we never triggered on "ABC" when there was no request pending. With `wait`, after the first request completes, it sees the value has changed since the request was first fired would have fired again with "ABC".

I definitely can see a user wanting to be able to use each of these three behaviors.

abidlabs:
I also think we avoid increasing the API. Instead of having our users think through these options, what about if we were to simply change the default behavior to "wait"? Wouldn't that solve all of the issues?

* If a user has a heavy model, then if they click submit a bunch of times, only the first request and final request will go through (assuming that the model is heavy so it will be running during the intermediate requests)
* Clearly, it will solve the original issue #1954 
* I don't think the use case for `run` is particularly common and it could clog up the queue so we shouldn't support it

pngwn:
We could just batch requests per client. 

- send request (gets processed)
- request
- request 
- first request completes and is returned
- subsequent requests are handled as a single step (either they're collected and batched or one is chosen and processed)

It works for heavy models: you'll never run another while one is running. And if there's is a large queue of requests you still only run one. This assumes they are idempotent which they might not be due to state, they could be batched. 

For live text inputs, same story really, we discard the intermediate requests. 

I'm still not sure what the run usecase is really, since the output would be a single component they would just overwrite each other, with multiple threads are there even any guarantees of order? I guess if you were triggering a bunch of side effects or logging stuff the then you would be happy to just launch them whenever, some of these kinds of requests could probably be batched though. Would need to understand the conference usecases for this.

abidlabs:
@pngwn by batching, are you referring to collecting a list of samples and passing them into the prediction function ([as in this PR that I've been working on](https://github.com/gradio-app/gradio/pull/2218)). This requires the function to accept batched inputs, so it wouldn't always work.

But basically as long as we:

> we discard the intermediate requests.

I think we should be fine

pngwn:
Not really. I guess just discarding intermediate requests is fine but this won't work when we have state. If those intermediate make use of `gr.State` then the results will be wrong if we don't batch those state updates somehow. But that might be difficult.

aliabid94:
I actually feel that all three modes will have very valid use-cases as people build more complex interfaces, and I don't think that this is significantly expanding the API surface. It's also possible to disallow "run" mode if any of the inputs or outputs are state.
I'm okay to move to "wait" behavior only for all event listeners for now, but will not be surprised if a user later asks to be able to launch multiple processes by clicking a button multiple times.

abidlabs:
@dawoodkhan82 this might close https://github.com/gradio-app/gradio/issues/1453 as well?

abidlabs:
I synced with @dawoodkhan82 on this a bit, and here's what we're thinking:

* Let's make the default behavior to be `wait` and leave everything else to avoid unnecessary complexity for now. If there are strong use cases for other behaviors then we can consider adding them, but for now they are too "low-level".

cc @freddyaboulton @pngwn @aliabid94

pngwn:
This is technically a breaking change. Would we consider this a bug fix instead? Have we tested to ensure this doesn't break any other existing use-cases or cause issues because of the change in behaviour?

abidlabs:
Good point @pngwn. Here's a demo that would "no longer work" as expected:

```py
import gradio as gr
import time

def slow_identity(x):
    time.sleep(1)
    return x

with gr.Blocks() as demo:
    
    with gr.Row():
        i1 = gr.Textbox()
        o1 = gr.Textbox()
    with gr.Row():
        i2 = gr.Textbox()
        o2 = gr.Textbox()
    with gr.Row():
        i3 = gr.Textbox()
        o3 = gr.Textbox()        
    i1.change(slow_identity, i1, o1)
    i2.change(slow_identity, i2, o2)
    i3.change(slow_identity, i3, o3)  
    
demo.launch()
```

I still think that having the "wait" behavior is overall better, but perhaps we should make this change in 4.0 to avoid any breaking changes we don't foresee. WDYT about kicking this PR and associated issues to 4.0 milestone?

pngwn:
Closing for now.

PR #2448, opened 2022-10-12
Add blur event to textbox and number

dawoodkhan82:
# Description
Add blur event. Triggered when user clicks out of textbox.

https://user-images.githubusercontent.com/12725292/195451431-d652365b-ded9-46ec-a6ec-454835fd0679.mov

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

Closes: #2449

# Checklist:

# A note about the CHANGELOG

### DIFF
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -47,7 +47,7 @@ from gradio.events import (
     Playable,
     Streamable,
     Submittable,
-    Blurable,
+    Blurrable,
 )
 from gradio.layouts import Column, Form, Row
 from gradio.serializing import (
@@ -255,7 +255,7 @@ class FormComponent:
 
 
 @document("change", "submit", "blur", "style")
-class Textbox(Changeable, Submittable, Blurable, IOComponent, SimpleSerializable, FormComponent):
+class Textbox(Changeable, Submittable, Blurrable, IOComponent, SimpleSerializable, FormComponent):
     """
     Creates a textarea for user to enter string input or display string output.
     Preprocessing: passes textarea value as a {str} into the function.
@@ -421,7 +421,7 @@ class Textbox(Changeable, Submittable, Blurable, IOComponent, SimpleSerializable
 
 
 @document("change", "submit", "style")
-class Number(Changeable, Submittable, Blurable, IOComponent, SimpleSerializable, FormComponent):
+class Number(Changeable, Submittable, Blurrable, IOComponent, SimpleSerializable, FormComponent):
     """
     Creates a numeric field for user to enter numbers as input or display numeric output.
     Preprocessing: passes field value as a {float} or {int} into the function, depending on `precision`.
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -47,6 +47,7 @@ from gradio.events import (
     Playable,
     Streamable,
     Submittable,
+    Blurable,
 )
 from gradio.layouts import Column, Form, Row
 from gradio.serializing import (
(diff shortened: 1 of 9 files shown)


gradio/events.py
```diff
@@ -465,3 +465,52 @@ def stream(
...
+            outputs: List of outputs
+            api_name: Defining this parameter exposes the endpoint in the api docs
+            scroll_to_output: If True, will scroll to output component on completion
+            show_progress: If True, will show progress animation while pending
+            queue: If True, will place the request on the queue, if the queue exists
+            preprocess: If False, will not run preprocessing of component data before running 'fn' (e.g. leaving it as a base64 string if this method is called with the `Image` component).
+            postprocess: If False, will not run postprocessing of component data before returning 'fn' output to the browser.
+        """
+        # _js: Optional frontend js method to run before running 'fn'. Input arguments for js method are values of 'inputs' and 'outputs', return should be a list of values for output components.
+        if status_tracker:
```
aliabid94, line 499: remove deprecated arguments in new Components

gradio/components.py
```diff
@@ -47,6 +47,7 @@ class DataframeData(TypedDict):
     Playable,
     Streamable,
     Submittable,
+    Blurable,
```
aliabid94, line 50: according to wiktionary, should be "Blurrable" (two r's)

aliabid94:
lgtm

pngwn:
Is this really a thing people want to do?

abidlabs:
> Is this really a thing people want to do?

It could be a useful way to get a .change()-like effect that doesn't fire with every character that is typed.

* Tested and works well with `gr.Textbox()`, however, it doesn't seem to work for `gr.Number()` for me.

```py
import gradio as gr
import time


with gr.Blocks() as demo:
    a = gr.Number()
    c = gr.Number()
    
    a.blur(lambda x:x, a, c)
    
demo.launch()
```

aliabid94:
> could be a useful way to get a .change()-like effect that doesn't fire with every character that is typed

But the user **has** to click out of the box to trigger it, and they may not be aware of it. Personally no strong feeling whether or not to merge this.

abidlabs:
The use case that I envision here is a form with multiple textboxes / numbers, and you want the output to update as a user updates the values of the form components. It's nice for the user to be able flow between the form components without having to click a submit button.

Imo it doesn't the surface area too much or in any significantly new directions, so would be in favor of merging once it's fully working.

dawoodkhan82:
@abidlabs should work for number now.

abidlabs (approved):
Works great thanks @dawoodkhan82! LGTM

pngwn:
This would be better as a `change` event, rather than a `blur` to cover the usecase that @abidlabs mentioned. The described behaviour is almost exactly what the native `change` event does. `blur` is really about focusing and unfocusing.

Note: Although _our_ event is called `change` the actually event we listen for is `input` which occurs on every keystroke.

dawoodkhan82:
> 

Right, I was considering re-implementing the change event by using `blur` but I believe the plan is to fix the current change event to work for every keystroke, like this PR would (https://github.com/gradio-app/gradio/pull/2436)

abidlabs:
Yeah I think we shouldn't change the `change()` event dramatically as this might break expectations around existing demos. However, changing the behavior of it to "wait" would resolve the problems people are having.

pngwn:
Sorry, I meant that this blur functionality is better covered by the native change DOM event, not that we should alter our change event. 

One benefit of the native change event is that it doesn't trigger when the value has not changed even if the user focuses in and out of an input.

PR #2453, opened 2022-10-13
Port reuse block fix

aliabid94:
@pngwn fixed ports with the `reuse_ports` argument in `get_first_available_port`, but when it was called from networking.start_server, it was never set to True. I just removed the argument and made the reuse logic always happen. Hopefully will never have to deal with this annoying problem again!

Btw the port would only get clogged if you make an actual request to the server, e.g. open it in a tab. If you just started an app and killed it right after, it wouldn't. FYI in case you're testing this.


pngwn (approved):
This is awesome thanks! That pngwn character sounds like an idiot.

PR #2455, opened 2022-10-13
Allow gradio embeds in colab without share

aliabid94:
Embeds Gradio apps directly into using colabs `output.serve_kernel_port_as_iframe`. Had to change default api endpoint name to `/run` (though `/api` is still supported for backwards compatibility.).

Note: websockets with queueing are not supported, so will still use sharing in that case.

Example colab: https://colab.research.google.com/drive/1oOyMGV0cxziK_ViSmh0y2c2FxPOHPFMs?usp=sharing


aliabid94:
Still a draft because the colab.html code I copied from your colab @pngwn but I assumed that might not be the final version of that code. feel free to edit this PR directly

gradio/routes.py
```diff
@@ -302,7 +304,7 @@ async def predict(
                 if body.fn_index is None:
                     return JSONResponse(
                         content={
-                            "error": f"This app has no endpoint /api/{api_name}/."
+                            "error": f"This app has no endpoint /run/{api_name}/."
```
abidlabs, line 307: It looks like we are migrating from "/api" to "/run" as the official API endpoint. Let's make sure to update the "view API" page as well.

Also this is a bit of an edge case but what happens if a user-supplied `api_name` includes the string "api". Perhaps we can print a warning to users letting them know that their demos will not work inside Colab

gradio/blocks.py
```diff
@@ -1110,26 +1098,38 @@ def reverse(text):
...
-        is_colab = utils.colab_check()
-        if is_colab or (_frontend and not networking.url_ok(self.local_url)):
+
+        if self.is_colab and not quiet:
+            if debug:
+                print(strings.en["COLAB_DEBUG_TRUE"])
+            else:
+                print(strings.en["COLAB_DEBUG_FALSE"])
+
+        if (self.is_colab and not self.share) or (
```
abidlabs, line 1119: I believe this should be:
```suggestion
        if (self.is_colab and and self.enable_queue and not self.share) or (
```

The conditions in this line are hard to parse. Consider refactoring. Also, can you add tests to make sure that all of the combinations of share/queue/colab are correctly handled?

abidlabs:
This is really cool stuff @aliabid94 and @pngwn!

I released a new version `gradio==3.0.1b123` to test with the latest changes. However, this version is not working for me. Anytime I try to launch an Interface, I get "ValueError: When using queueing in Colab or when localhost is not accessible, a shareable link must be created. Please set share=True." I think I know the culprit (it's the if condition that I commented on above). 

In the meantime, I'm testing with `gradio==3.0.1b121` which seems to work

gradio/templates/colab.html
```diff
@@ -0,0 +1,26 @@
...
+    function rehydrate() {
+        for (const script of document.querySelectorAll("script")) {
+            const newScript = document.createElement('script')
+            newScript.type = script.type;
+            newScript.src = script.src;
+            newScript.textContent = script.textContent;
+            const head = document.getElementsByTagName('head')[0];
+            if (script.textContent.includes("window.gradio_config =")) {
+                window.gradio_config = JSON.parse(script.textContent.replace("window.gradio_config =", "").replace(/;$/, "").trim());
+            }
```
pngwn, line 13:
```suggestion
            if (script.textContent.includes("window.gradio_config =")) {
                const config_re = /^\s*window.gradio_config\s*=\s*({[^]+})\s*;\s*$/;
                window.gradio_config = JSON.parse(config_re.exec(script.textContent)[1]);            }
```

I think this regex is a bit more robust (slightly). 

I'm not entirely sure why the config isn't getting set correctly but it could be to do with different window scopes? or maybe there is something blocking it. I don't know if there is another solution but this does fix it.

gradio/templates/colab.html
```diff
@@ -0,0 +1,26 @@
...
+            head.appendChild(newScript);
+            newScript.remove();
+        }
+
+        console.log(window)
```
pngwn, line 18: This can be removed.

abidlabs:
Tested a bunch of different Gradio demos with `gradio==3.0.1b121`and generally everything seems to be working! Some feedback:

* When you launch a demo on colab, some text is printed (the URL for localhost, etc.) before the demo overrides the text. The experience is a little disorienting, we should try to leave the text there or hide the text altogether 

image


* If you launch a demo, and have a few minutes of inactivity, the following error happens:

image

I thought this error only happens if you have queuing enabled, but I experienced with a very simple UI without the demo enabled. We should investigate what's going on here.

* queueing() does not work. You may have fixed it after `gradio==3.0.1b121` was released, but this needs to be tested

gradio/templates/colab.html
```diff
@@ -0,0 +1,26 @@
...
+    }
+    fetch(".")
+        .then((x) => x.text())
+        .then((html) => {
+            document.body.innerHTML = html;
+        })
+        .then(() => rehydrate());
```
pngwn, line 25: maybe we could add a `.catch` here and have a small error just in case something goes wrong.

```js
.catch(() => {
   document.body.innerHTML =`<h1 color="red">There was a problem running Gradio in colab, please rerun the block and file an issue on GitHub if it doesn't not work.</h1>`;
})
```

pngwn (commented):
This is great. I think most of the frontend code is fine, left a few comments but I can't think of anything else we really need to cover.

gradio/templates/colab.html
aliabid94: instead of parsing the js code, we code also get the config via GET request at `/config`? it requires an extra request but its more robust

pngwn: Yeah, we can do. I just figured it would be quite a bit slower. I'm happy with either, I only did this because it was already in the document.

gradio/routes.py
aliabid94: Does colab block any call with a substring "api"? Anyways, even if they create a custom named API, the frontend does not use the names API's, so gradio UI will still work fine.

gradio/templates/colab.html
```diff
@@ -0,0 +1,27 @@
...
+            newScript.remove();
+        }
+    }
+    fetch(".")
+        .then((x) => x.text())
+        .then((html) => {
+            document.body.innerHTML = html;
+        })
+        .then(() => rehydrate()).catch(() => {
+            document.body.innerHTML = `<h1 color="red">There was a problem running Gradio in colab, please rerun the block with gradio==3.5 and file an issue on GitHub if it doesn't not work.</h1>`;
```
abidlabs, line 25:
```suggestion
            document.body.innerHTML = `<h1 color="red">There was a problem running Gradio in colab, please rerun the block with gradio>=3.6 and file an issue on GitHub if it doesn't not work.</h1>`;
```

abidlabs: Not how this message could ever appear in an older version of `gradio` anyways?

aliabid94:
Updated and ready for rereview. Try everything to break it pls

abidlabs:
This is *really* cool! I tested a variety of demos which work great (and in fact run faster than they did before!). I also see that enabling the queue automatically generates the share link. A few remaining points:

* **Microphone/webcam access** 

Is it possible to grant microphone/webcam access to the iframe, just as we currently do? Right now, if you try to run the `stream_audio` demo, you'll notice that it currently errors out with no way to actually enable microphone access:

image

* **Respecting the `height` parameter in `launch()`**

Launching a very tall demo is frustrating because you have to scroll to see the full demo. It used to be that we could set the `height` parameter to adjust the height of the embedded but this is no longer respected. See the `blocks_layout` demo for example:

image


* **Removing the localhost link**

The printed message could be improved -- in particular, there is no point printing the full localhost link since clicking on it does not take the user to the actual Gradio demo. Here's what I suggest should be printed instead. Notice that we do not print the localhost link and we add a few other instrucitons:

```
Colab notebook detected. To show errors in colab notebook, set `debug=True` in `launch()`

Using Embedded Colab Mode, a beta feature.  Running app on port XXXX.
If you have issues, please revert to gradio<=3.8.2 and file a Github issue. 

Note: opening the browser inspector may crash Embedded Colab Mode.

To change the height of this embedded demo, set a `height` (in pixels) in `launch()`
To create a public link or a link that can be opened in a new tab, set `share=True` in `launch()`.
```

gradio/blocks.py
```diff
@@ -1299,26 +1292,40 @@ def reverse(text):
                     )
         utils.launch_counter()
 
+        self.is_colab = utils.colab_check()
+        self.share = (
+            share
+            if share is not None
+            else True
+            if self.is_colab and self.has_any_queue
```
abidlabs, line 1300: Why do you need this additional `self.has_any_queue` parameter? Isn't `self.queue_enabled` enough?

gradio/routes.py
```diff
@@ -304,6 +304,8 @@ async def run_predict(
                 output["data"] = output["data"][0]
             return output
 
+        @app.post("/run/{api_name}", dependencies=[Depends(login_check)])
```
abidlabs, line 307: Consider adding a comment here for future readers of the codebase explaining why we have a second route

pngwn:
In the colab you linked, I'm getting a 403 when the app tries to run.

gradio/strings.py
```diff
@@ -19,6 +19,9 @@
     "COLAB_DEBUG_TRUE": "Colab notebook detected. This cell will run indefinitely so that you can see errors and logs. "
     "To turn off, set debug=False in launch().",
     "COLAB_DEBUG_FALSE": "Colab notebook detected. To show errors in colab notebook, set debug=True in launch()",
+    "COLAB_BETA": "\nNOTE: Using Embedded Colab Mode, a beta feature."
+    "\nIf you have issues, please revert to gradio==3.8.2 and file a Github issue. "
```
freddyaboulton, line 23: If they encounter an issue, using a share link should work right? I think that'd be better than asking them to downgrade.

freddyaboulton:
Thanks @aliabid94 ! I tried a couple of demos and they all seem to work well!

Here is the colab I've been using,.
https://colab.research.google.com/drive/10xwmQKKOgc8fpk6bzjCOjUzCN0vMsyK2#scrollTo=PmW1BSEenAtl

One thing I noticed is that the Gallery component on colab won't let you click on individual images?

Here is the behavior I'm seeing running it locally:
fake_ga

Here is the behavior when I run it with share=True in the colab:
fake_gan_2

aliabid94:
> In the colab you linked, I'm getting a 403 when the first gradio app tries to run. The others work fine.

Yikes, what? that makes no sense.. can you make sure the server is running via 
```python
import requests
requests.post("http://localhost:7860/api/predict", json={"fn_index": 0, "data": ["test"], "session_hash": "asdf"})
```

aliabid94:
@freddyaboulton thanks for the in depth testing!

1. I believe this occurs because at certain window sizes, clicking on gallery images does not expand them (right @pngwn?). Currently the iframe sizes of direct embedding vs share links are different, will fix.  If you change the window size, you'll see that clicking will be responsive in galleries.
2. Hmm I think I need to enable permissions on the iframe for file downloads, will look into how to do this via the colab API.

gradio/strings.py
aliabid94: Done

gradio/routes.py
aliabid94: Done

gradio/blocks.py
aliabid94: You're right, fixed

aliabid94:
Okay everything should be fixed!
- Copied logic of internal google.colab.output.serve_kernel_port_as_iframe and modified it to allow iframe permissions
- Window sizes now respect height and width
- File downloads do not respect the port proxying, so implemented something that detects colab, and if so, opens file downloads in a new tab (which works)
- Updated language and comments as suggested

I think it's good to go!

abidlabs:
LGTM, I can confirm that the issues have been fixed 🚀

abidlabs (approved):

freddyaboulton:
What's the right version to test with? 3.0.1b121?

Issue #2419, opened 2022-10-08
Identify the source of the change event (automatic, or manual upload) for video/audio components


altryne:
- [x ] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
Would like to be able to diffrentiate between a user initiated manual upload to a Video and an automatic value update from somewhere else 

**Describe the solution you'd like**  
Add a .upload event, that causes the .change event to have metadata for the upload. 

**Additional context**  
This is helpful as right now, I have to use 2 separate elements for uploading and for showing a result, I would like that to be 1 element

abidlabs:
Thanks for creating the issue @altryne. Just to confirm, when you say 

> automatic value update from somewhere else

you mean if the `Video` component is being set as the output of some function, correct?

pngwn:
Yeah. The change is fired when the user uploads some media but also when the component data is set via a predict function. There is currently no way to differentiate. 

I actually think having a general change event is quite nice because many people don’t care how the data is set, they just want to know when it is but it would be good to have some way to differentiate.

altryne:
> I actually think having a general change event is quite nice

Completely agree, I think change should work all the time. 

But maybe an .upload event could be fired before so that I could save a reference somehow? Not sure if I overstep with implementation suggestions.

Issue #1431, opened 2022-05-31
Distinguish between user-triggered events and gradio-triggered events


FarukOzderim:
### Describe the bug

Created a hacky blocks party space for fun, but did not work as intended, creates infinite recursion with events triggering each other. 
Question: Is there a way we can avoid this? @aliabid94, @pngwn 

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

https://huggingface.co/spaces/Gradio-Blocks/Hacky-Blocks-Party

### Screenshot

![image](https://user-images.githubusercontent.com/44067394/171199697-b09b0125-0549-4b4f-98c8-08b2d53de329.png)

### Logs

```shell
No logs
```

### System Info

```shell
spaces
```

### Severity

annoying

pngwn:
What would you expect to happen here? It looks like you've explicitly coded the recursion in.

FarukOzderim:
I actually didn't expect that the events were going to trigger each other. Thought changes other than the user made in the UI were not triggering the events.

abidlabs:
I think we can close this issue now. We now support this for many components using `.select()` event trigger (which only fires if a user manually selects an option). We can also only fire an event if there's a programmatic change to a value using `.then()`. 

Haven't seen any real-world use cases of the kind of recursion described in the issue above. But even that can be achieved (albeit in a somewhat hacky way) using `cancels`: 

```py
import gradio as gr

with gr.Blocks() as demo:
    a = gr.Textbox()
    b = gr.Textbox()
    
    ev1 = a.change(lambda x:x, a, b)
    ev2 = b.change(lambda x:x, b, a)
    a.change(None, None, None, cancels=ev2)
    b.change(None, None, None, cancels=ev1)
    
demo.queue().launch()
```

aliabid94:
This is not resolved, will fix now

PR #2456, opened 2022-10-13
Load Event Trigger

dawoodkhan82:
# Description
Add event trigger for when a user uploads to video, audio, file, and image components.

https://user-images.githubusercontent.com/12725292/195712881-7f1d2ef6-5f9a-44e8-a756-543e766c7cf1.mov

```
import gradio as gr

def test(video):
    return video

with gr.Blocks() as demo:
    video1 = gr.Video()
    video2 = gr.Video()
    video1.load(test, video1, video2)

demo.launch()
```

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

Closes: #2419 

# Checklist:

# A note about the CHANGELOG


CHANGELOG.md
```diff
@@ -96,8 +100,7 @@ No changes to highlight.
 * Lets users provide a `gr.update()` dictionary even if post-processing is diabled [@abidlabs](https://github.com/abidlabs) in [PR 2385](https://github.com/gradio-app/gradio/pull/2385)
 * Fix bug where errors would cause apps run in reload mode to hang forever by [@freddyaboulton](https://github.com/freddyaboulton) in [PR 2394](https://github.com/gradio-app/gradio/pull/2394)
 * Fix bug where new typeable slider doesn't respect the minimum and maximum values [@dawoodkhan82](https://github.com/dawoodkhan82) in [PR 2380](https://github.com/gradio-app/gradio/pull/2380)
-* Add guide on creating a map demo using the `gr.Plot()` component [@dawoodkhan82](https://github.com/dawoodkhan82) in [PR 2402](https://github.com/gradio-app/gradio/pull/2402)
-* Add blur event for `Textbox` and `Number` components [@dawoodkhan82](https://github.com/dawoodkhan82) in [PR 2448](https://github.com/gradio-app/gradio/pull/2448)
+
```
dawoodkhan82, line 103: Moving these to the right section

abidlabs:
Very cool @dawoodkhan82! Tested and works great for all four of the components. I do have one suggestion, which is to rename the event from `.load()` to `.upload()` for two reasons:

* We already use `.load()` as an event trigger for `Blocks()` where it has a different meaning: it triggers the function call when the Blocks page _loads_. Whereas here we are using it differently, so this might be confusing to users.

* This new event trigger has no effect when the `source="webcam"` or `"canvas"`, or anything besides `source="upload"` (tested and confirmed). So calling it `.upload()` will be consistent with this fact

I also think that *every* component should have a version of this event trigger. For example, `Textbox()` should have a `.type()` trigger which only fires when someone types in the textbox, but not if the change happens due to a function outputting into the textbox. This would allow us to close #1431 for example. But we can save this for a later PR

pngwn:
I think this is a good addition until we change the API but I'm not sure this _really_ fixes #2419 because `change` will still be firing every time the value changes (which it should), in order to separate the two events requires some pretty unreliable user code.

An app author would need to track when both events are triggering in order to distinguish them, but I don't think that is reliable unless we send a session ID with each request and I'm not sure if we document the order in which events are fired or maintain that as a part of the API, so the required user code to make this work would be even more complex.

So this fixes #2419 in one direction but not in the other.

CHANGELOG.md
dawoodkhan82: @freddyaboulton Seems like changelog check is failing. Is it because I'm moving these to the right section?

CHANGELOG.md
```diff
@@ -48,6 +48,8 @@ In the example above, 16 requests could be processed in parallel (for a total in
 time of 5 seconds), instead of each request being processed separately (for a total
 inference time of 80 seconds).
 
+* Add load event for `Video`, `Audio`, `Image`, and `File` components [@dawoodkhan82](https://github.com/dawoodkhan82) in [PR 2448](https://github.com/gradio-app/gradio/pull/2456)
```
abidlabs, line 51:
```suggestion
* Add upload event for `Video`, `Audio`, `Image`, and `File` components [@dawoodkhan82](https://github.com/dawoodkhan82) in [PR 2448](https://github.com/gradio-app/gradio/pull/2456)
```

gradio/events.py
```diff
@@ -653,3 +652,51 @@ def blur(
...
+        inputs: List[Component],
+        outputs: List[Component],
+        api_name: AnyStr = None,
+        scroll_to_output: bool = False,
+        show_progress: bool = True,
+        queue: Optional[bool] = None,
+        preprocess: bool = True,
+        postprocess: bool = True,
+        cancels: List[Dict[str, Any]] | None = None,
+        _js: Optional[str] = None,
```
abidlabs, line 670: Should accept `batch` parameters like the other event triggers do now

abidlabs (approved):
Left a few minor comments, but otherwise looks good!

PR #2469, opened 2022-10-16
Small fix for version check before reuploading demos

aliabd:
Demos were unnecessarily reuploading even when the version hasn't changed. This fixes that by correcting the comparison.


abidlabs (approved):

abidlabs:
Makes sense thanks for the fix @aliabd

Issue #2173, opened 2022-09-03
Optional Audio/Sound alert when a job is complete


nelsonjchen:
- [X] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
I frequently go browse other tabs or applications when I am running a long running batch of jobs. Sometimes I fall into a hole and take a bit longer than expected to get back to the tab. If I stay in the research hole a bit too long, my hardware is idle when it could be working on jobs batches. I want to maximize utilization of my hardware, iterate more, and the limiter is me.

**Describe the solution you'd like**  
I would like some sort of optional audio/sound notification to be played when generation is complete so I know to return to the UI, evaluate the results, tinker, and kickoff the next batch. This should help keep my hardware working and help increase output and utilization.

I don't know Gradio's API that well to be honest and this may very well be implementable downstream as a Gradio consumer without upstream support. I do know that if the functionality is implemented upstream and is turnkey that downstream Gradio users are more likely to implement and turn on a pre-existing turnkey audio/sound alert solution. I do see a lot more Gradio in my future.

**Additional context**  
I have a more specific version of this request here for the hlky/stable-diffusion's gradio implementation: https://github.com/hlky/stable-diffusion-webui/discussions/232 

I figure it might be a good idea to get this idea upstream so that maybe it can flow back downstream someday. I believe it to be a pretty generic idea that isn't specific to my use case and is applicable to other uses as well.

I do have a suggestion for a default sound which I don't think should kick off anyone's tinnitus: https://freesound.org/people/collierhs_colinlib/sounds/588718/

It reminds me of the buckle your seatbelt sign/turbulence for passengers warning on an airplane for some reason.

abidlabs:
I really like this suggestion!

nelsonjchen:
https://github.com/AUTOMATIC1111/stable-diffusion-webui/pull/1126

This was added to sthe AUTOMATIC1111 SD UI.

The code doesn't look heavy. Still would love more proliferation. Maybe throw something similar into some samples/documentation?

freddyaboulton:
Very cool @nelsonjchen !

abidlabs:
Great @nelsonjchen! Yes we can add this as a demo people can refer to!

PR #2478, opened 2022-10-18
Adds a demo to show how a sound alert can be played when a prediction is complete

abidlabs:
Adds a demo (`sound_alert.py`) to show how a sound alert can be played when a prediction is complete.

CLoses: #2173


demo/gif_maker/run.py
```diff
@@ -1,10 +1,8 @@
 import cv2
 import gradio as gr
-import tempfile
```
abidlabs, line 3: Just removes some unused imports I noticed

aliabd:
Nice @abidlabs 

I'm seeing a console error on the demo, even though it runs as expected

```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'forEach')
    at $ (index.b57b8f62.js:55:2384)
```

Also, consider adding the demo to the docstrings where applicable (maybe at blocks?) or to the demos page. Otherwise LGTM

abidlabs:
Thanks for the review @aliabd! Will address these

freddyaboulton (approved):
Looks good to me @abidlabs!

dawoodkhan82:
nice demo!

dawoodkhan82 (approved):

abidlabs:
Thanks for the review folks! @aliabd I believe the JS error is happening because this event trigger doesn't have any inputs or outputs. Gonna leave that for a future PR but I've gone ahead and added this demo the `Blocks` class.

Issue #2468, opened 2022-10-15
File paths not properly escaped in /file route, causing RuntimeError (file does not exist)


cpupower:
### Describe the bug

When files are served in the UI via temporary files, the path passed via the `path`/`file` parameter of the `/file` [route](https://github.com/gradio-app/gradio/blob/831ae1405f0f2cddf8d508775731eef39c5a09df/gradio/routes.py#L230) is not properly escaped for use in URLs.

This is for example an issue if a hash (`#`) is part of the path, which will cause the remaining string to be interpreted as the fragment part of an URL, which is not sent to the server. It can be reproduced with the `Gallery` component, but possibly other components are affected as well.

A (likely incomplete) workaround is to escape the path in the client-side `normalise_file` function in [utils.ts](https://github.com/gradio-app/gradio/blob/831ae1405f0f2cddf8d508775731eef39c5a09df/ui/packages/upload/src/utils.ts#L3), however I am not sure whether this is the appropriate location.

Possibly related issues:
- https://github.com/gradio-app/gradio/issues/2129
- https://github.com/gradio-app/gradio/issues/2335

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```py
import gradio as gr
from PIL import Image

def poc():
    with open("./red#.png", "wb") as f:
        Image.new("RGBA", (1,1), (255,0,0,255)).save(f)
    return ["./red#.png"]

demo = gr.Interface(
    fn=poc,
    inputs=[],
    outputs=[gr.Gallery()]
)
demo.launch()
```

In this example, the `red#.png` will not be correctly retrieved to to the hash `#` character, and therefore it will not be displayed in the gallery. The UI would make a request to e.g. `http://127.0.0.1:7860/file=/tmp/tmpfolder/red#random.png`, however while the file `/tmp/tmpfolder/red#random.png` exists, the server only gets to see the path `/tmp/tmpfolder/red`, and therefore likely throws a `RuntimeError: File at path /tmp/tmpfolder/red does not exist.`.

### Logs

```shell
Running on local URL:  http://127.0.0.1:7860

To create a public link, set `share=True` in `launch()`.
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/home/USER/.local/share/virtualenvs/reproduce-bug/lib/python3.8/site-packages/starlette/responses.py", line 331, in __call__
    stat_result = await anyio.to_thread.run_sync(os.stat, self.path)
[...]

PR #2480, opened 2022-10-18
Removes special characters from temporary filenames

abidlabs:
Removes special characters from temporary filenames so that the files can be served by components. 

Closes: #2468

### DIFF
--- a/gradio/processing_utils.py
+++ b/gradio/processing_utils.py
@@ -9,7 +9,6 @@ import shutil
 import subprocess
 import tempfile
 import warnings
-from ctypes import util
 from io import BytesIO
 from typing import Dict
 
--- a/gradio/processing_utils.py
+++ b/gradio/processing_utils.py
@@ -9,6 +9,7 @@ import shutil
 import subprocess
 import tempfile
 import warnings
+from ctypes import util
 from io import BytesIO
 from typing import Dict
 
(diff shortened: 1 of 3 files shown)


gradio/processing_utils.py
```diff
@@ -9,6 +9,7 @@
 import subprocess
 import tempfile
 import warnings
+from ctypes import util
```
freddyaboulton, line 12: This is the wrong import

gradio/processing_utils.py
```diff
@@ -329,6 +330,7 @@ def create_tmp_copy_of_file(file_path, dir=None):
     if "." in file_name:
         prefix = file_name[0 : file_name.index(".")]
         extension = file_name[file_name.index(".") + 1 :]
+    prefix = utils.strip_invalid_filename_characters(prefix)
```
freddyaboulton, line 333: Would be good to add a unit test for this!

freddyaboulton (approved):
Looks good to me @abidlabs ! Thanks for the fix.

abidlabs:
Thanks for the review @freddyaboulton! Used this excuse to organize the `test\test_processing_utils.py` a bit

ColtonBehannon:
I am curious if there is any way to still get the special characters out of the file names with fileobj.name? If I want my application to be able to upload that same file to other resources such Azure Storage Explorer I need to be able to access the original file name but the current workflow appears to remove special characters such as the "()" in test_file(1).txt.

abidlabs:
Not possible at the moment @ColtonBehannon since these are stripped. Is it possible for you to strip the same special characters on your end?

ColtonBehannon:
@abidlabs Thanks for the quick response! I think that could work though I'm not sure what all characters are considered special in this context and would need to be stripped. For example, are underscores included as well or not. Any suggestions here other than testing them all?

abidlabs:
Here's the logic: 

https://github.com/gradio-app/gradio/blob/52f7831751b432411e109bd41add4ab286023a8e/client/python/gradio_client/utils.py#L421

so yes underscores are valid

PR #2502, opened 2022-10-19
Fix share error message when files are built locally

abidlabs:
As discussed with @aliabid94, if a user builds `gradio` locally, we do not want them to be using `share=True` because local changes to the frontend will not be used. Instead, the frontend files from AWS S3 will be fetched. This could lead to some confusion, so instead we raise an Exception.


aliabid94:
Lgtm

PR #2507, opened 2022-10-20
client-side hotfix for port issue

aliabid94:
Basically waits until app is up, then checks if the app config matches the share link config. If not, tries to create another share link.
Possible issues:
- Currently tries 10 times to connect to share link, 1 per second. Some apps may take longer than 10 seconds to come up. (could wait forever also?)
- Checking to see if the config is up every second might add load to the server.


abidlabs:
I think we can close this PR. See: https://huggingface.slack.com/archives/C02QNREU24S/p1666288866211929

abidlabs:
Closing after discussion with @aliabid94

Issue #2260, opened 2022-09-14
Outputs that are larger than 2 MB do not work with `share=True`


MrNeon:
### Describe the bug

When using the shared tunneled link I get this error when receiving outputs that are around 2MB or larger.

Error in Chrome: net::ERR_HTTP2_PROTOCOL_ERROR 200
![Screenshot_20220914_140154](https://user-images.githubusercontent.com/7870061/190165237-fb1bd7eb-5f09-44a3-8230-27e2458560b3.png)

Firefox errors too but without any code, just an uncaught exception.

 When using localhost or using ngrok to tunnel the connection the issue never happens.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```python
import numpy as np
import gradio as gr
from PIL import Image

im = Image.open("test_image_over_2MB.png")

def get_image():
    return im

with gr.Blocks() as demo:
    image_output = gr.Image()
    image_button = gr.Button("Get")
    image_button.click(get_image, inputs=None, outputs=image_output)

demo.launch(share=True, debug=True)
```

### Logs

```shell
No logs
```

### System Info

```shell
Gradio version: 3.3
Firefox 104.0
Chrome 105.0.5195.52
```

### Severity

annoying

abidlabs:
Thanks @MrNeon for the interesting issue. This might be a limitation based on how the sharing servers have been configured. Will take a look and get back to you

dylancl:
Is there any update on this?

abidlabs:
I've been able to reproduce the issue, looking into a possible fix

guumaster:
Not sure if this helps, but if you use ngrok inside Google Colab and use it to point to localhost:7860 and then access through ngrok url instead of gradio.app, it never hangs. Here is the code:

```
import os
from flask import Flask
from pyngrok import ngrok

ngrok.set_auth_token(ngrok_token) #PUT TOKEN HERE
public_url = ngrok.connect(7860, bind_tls=True).public_url

os.environ["FLASK_ENV"] = "development"
app = Flask(__name__)
app.config["BASE_URL"] = public_url
app.run()
```

Maybe it helps to see what ngrok is doing differently, unless is a server setting 🤷

R4ZZ3:
Any estimate when this fix will be live?

abidlabs:
[...]

Issue #2180, opened 2022-09-05
Increase # of demos that can be shared simultaneously using `share=True`


abidlabs:
Given the increasing usage of Gradio, we should assess how many demos we can share simultaneously with `share=True` (@dawoodkhan82 can you look into this?) and then increase that capability by creating replicas of the Gradio API server.

PR #2509, opened 2022-10-21
Switch from SSH tunneling to FRP

abidlabs:
Continuing from: https://github.com/gradio-app/gradio/pull/2396

Remaining TODOs:
- [x] Add back md5 encryption (@Wauplin would you be able to take a look at this?)
- [x] Add https for the share links (@abidlabs)
- [x] Fix the connection persistency issue that @freddyaboulton brought up below (@XciD)
- [x] Add a better page when Gradio Interface is no longer active (@abidlabs)
- [x] Add tests (@abidlabs)
- [x] Expire links in 72 hours (@XciD would you be able to take a look at this?)
- [x] Do some load testing to figure out if we need to set up horizontal scaling (@aliabid94)
- [x] Don't hardcode IP address and port (@abidlabs)

Also:
* closes: #2260
* closes: #2180


freddyaboulton:
This works pretty well on the demos I've tried! 

The one thing I noticed that's weird is that the frp client seems to drop the connection to the gradio demo after a couple of minutes of inactivity.

I have this demo running on a jupyter notebook:

image


It worked great the first times I tried it locally and with the share link. However, I left it alone for a couple of minutes and then when I went back to the share link url I got a "Connection Error" and then the "not found" page even though the demo was still running on my machine. 

frp_timeout

Refreshing the page seems to fix it but might be confusing to users. I'm wondering what would happen if the prediction takes a couple of minutes to run?

abidlabs:
@Wauplin do you have any ideas of what could be causing the behavior @freddyaboulton is describing?

abidlabs:
Thanks to @XciD, this is now working with http://*.testing.gradiodash.com/!

abidlabs:
Added this page to appear if a link expires or is invalid:

image

Wauplin:
@abidlabs I've made a small PR to generate the privilege key dynamically based on timestamp: https://github.com/gradio-app/gradio/pull/2519 (including also some cosmetic changes). It generates exactly the same privilege key for the example that was previously hard-coded. I've tested it locally and the tunnel works fine.

But when I was talking about encryption I was referring encrypting the json payloads. @XciD haven't you talked about something like that in your presentation ? Something about the  fact that you created a special docker image where encryption is skipped just to do the tests but that we should set it back once we have a working version ? Or have I hallucinated this ? :smile:

Wauplin:
> The one thing I noticed that's weird is that the frp client seems to drop the connection to the gradio demo after a couple of minutes of inactivity.

> @Wauplin do you have any ideas of what could be causing the behavior @freddyaboulton is describing?

I sorry, I haven't been able to reproduce this error. I left a tunnel opened (from script in terminal) with a pending google chrome tab connected to it. Tried it again 1 hour later and haven't got any connection issue :confused:

XciD:
> But when I was talking about encryption I was referring encrypting the json payloads. @XciD haven't you talked about something like that in your presentation ? Something about the fact that you created a special docker image where encryption is skipped just to do the tests but that we should set it back once we have a working version ? Or have I hallucinated this ? 😄 ([related internal slack thread](https://huggingface.slack.com/archives/C045FUG7Z9U/p1666371051716369))

Nan you did not hallucinated: 
I've commented this code:
https://github.com/huggingface/frp/pull/1/files#diff-6e71c8e7a9485928fab7bf90204fd6404bccf67194bc471d3aa3bcd51f794facR302

Coming from: https://github.com/fatedier/golib/tree/dev/crypto

aliabid94:
I created a few colab notebooks to do some load testing. This was the setup:
- 3 identical colab notebooks for INTERFACE GENERATION ([1](https://colab.research.google.com/drive/1AMNHQrlcMmuN9mE--FMZ3Anf5mFU-gqh?usp=sharing), [2](https://colab.research.google.com/drive/1SeP8rNYB8g6w3fc5zXKRkfCrgFPW5qe8?usp=sharing), [3](https://colab.research.google.com/drive/1HikILoQQwe1mTnwyHkn7q3xswD3fpLhq?usp=sharing)) that create 100 interfaces each with `share=True` for a total of 300 interfaces using share simultaneously. The interface predictions take on average ~15 seconds to run (random duration between 1 and 30 seconds).
- 2 identical test notebooks for LOAD TEST ([1](https://colab.research.google.com/drive/1tW7Q8hQC0P0XNpQkc5GPfUrqhutz5OFt?usp=sharing), [2](https://colab.research.google.com/drive/1WmeOtY5MluF_A1Dx8tMeogAzUfyo1qTy?usp=sharing)) that send 5 kb of data at regular intervals in parallel via threads to the list of interfaces generated by the previous notebooks.

I varied the sleep duration between sending requests in the LOAD TEST notebooks. The success rate indicates how often the POST requests came back successfully.

When I ran both LOAD TEST notebooks simultaneously with the following sleep durations:
- 0.5 second (avg 60 concurrent requests): 100% success
- 0.25 second (avg 120 concurrent requests): 99.5% success
- 0.1 second (avg 240 concurrent requests): 91% success

I believe the drop in success rate is not due to the infrastructure, but because the INTERFACE GENERATION colab notebooks cannot handle the load. The interfaces that failed cannot accept POST requests anymore, even at slower rates.

I'm not sure if this setup is the ideal way to load test, open to suggestions.

abidlabs:
Great thanks @aliabid94! This is very helpful and I think generally looks promising. Would it be possible to investigate two additional things:

* **What is our capacity for total shared connections?** I.e. how many Interfaces with `share=True` can we support at the same time? This will tell us if we need to add horizontal scaling. Given that we currently have about ~2k concurrent requests, we should make sure that our capacity is well above that. Note that it might be faster to use `networking.create_tunnel()` rather than to create individual Interfaces (maybe?)
*  **Is there any increase in prediction latency as the number of connections increases?** Since the Gradio server instance is very small (t2.micro I believe), I was thinking maybe CPU can't handle switching between different connections very efficiently. If so, this might be reflected in an increase in latency as the number of connections increases. We could fix the prediction time and see if this increases as the number of concurrent connections increases?

aliabid94:
- We max out at 2k concurrent connections, probably a lot less than 2k concurrent requests. The best way to test this would be to launch more INTERFACE GENERATION notebooks, however google maxes out my colab sessions. If we can sync (maybe with one more person) and each run the colabs from our account, we can probably get close to 2k sessions running together
- I can test this. Btw isn't the gradio server a t2.2xlarge?

Wauplin:
That load testing is very interesting. I think we should also compare the numbers we obtain with the RFC tunnel against what we have with the existing SSH tunnel.

Also if the google colab are not powerful enough we can always spawn aws instances (I would also try on a personal computer but I'd be a bit more worried about the latency).

abidlabs:
+1 to @Wauplin's suggestion. @aliabid94 can we compare these benchmarks to ssh tunneling? I believe you can test ssh tunneling with gradio-staging.com

abidlabs:
@freddyaboulton I have not been able to reproduce the "Connection Error" problem either. Is it possible that your network connection was interrupted? I just ran a prediction that lasted ~10 min and it ran fine

abidlabs:
So I discussed with @XciD and it's rather tricky to enforce the 72 hour limit on the server side. For now, I have set the limit in the python library. This isn't tamper-proof but it should suffice for now, while [we look into a server-side solution](https://github.com/huggingface/frp/issues/2).

This is **ready for review**! (Note that in order to make it easy to test on colab, etc. I have released the latest version of the library as `gradio==3.6.0b7`)

gradio/blocks.py
```diff
@@ -1265,8 +1266,14 @@ def reverse(text):
                 raise RuntimeError("Share is not supported when you are in Spaces")
             try:
                 if self.share_url is None:
-                    share_url = networking.setup_tunnel(self.server_port, None)
-                    self.share_url = share_url
+                    print(
+                        "\nSetting up a public link... we've recently upgraded the "
+                        "way public links are generated. But if you encounter any"
+                        "issues, please try with gradio <= 3.7.\n"
```
Wauplin, line 1272: Maybe we could add a `"and please report the issue on https://github.com/gradio-app/gradio"` ?

test/test_networking.py
```diff
@@ -75,62 +74,17 @@ def test_start_server(self):
...
-#         aiohttp.ClientSession.post.assert_called()
-#         self.assertEqual(response.status_code, 200)
-#         io.close()
+class TestURLs(unittest.TestCase):
+    def test_url_ok(self):
+        assert networking.url_ok("https://www.gradio.app")
 
 
-class TestURLs(unittest.TestCase):
+class TestSharing:
```
Wauplin, line 82: Make it inherit from `unittest.TestCase` just in case ?

Wauplin (approved):
Looks good to me ! 
(for the record, I haven't tested it again myself)

freddyaboulton (approved):
Tested and works great! Thank you @Wauplin @XciD and others 👏

gradio/blocks.py
abidlabs: Agreed, made the change.

abidlabs:
Noticed an issue while doing a round of testing -- it seems like the `kitchen_sink` demo does not work with FRP approach. I get the following error:

```
Exception in thread Thread-50-handle_req_work_conn:
  File "c:\users\islam\dev\gradio-repos\gradio\gradio\tunneling.py", line 28, in _inner_target
    target(*args)
  File "c:\users\islam\dev\gradio-repos\gradio\gradio\tunneling.py", line 75, in handle_req_work_conn
    socket_worker.send(data)
ConnectionResetError: [WinError 10054] An existing connection was forcibly closed by the remote host
```

I think this happens when a lot of files are being loaded over the TCP connection. Here are the docker logs:

image

which are suggesting some files not found (?) -- going to look into this some more, as this is blocking merging this branch.

To reproduce:
* Go to `/demo/kitchen_sink/run.py` and set `share=True` in the `launch()` method
* Run `python demo/kitchen_sink/run.py`
* When the UI loads, click "Submit" 

If anyone has any ideas why this might be happening, I'm all ears! @aliabid94 @Wauplin @XciD

Wauplin:
Very dumb question but has anyone tested from a windows machine before ? (I only tested from Linux laptop)

abidlabs:
Hmmm I've been doing a lot of testing from a Windows machine. Generally everything is working fine, but this is the first demo  that doesn't seem to work. 

Can someone with a linux machine try the `kitchen_sink` demo to see if it works for them?

Wauplin:
Ok good, was just to double check. I'll check on Monday.

Wauplin:
@abidlabs @XciD I have no clue what is the issue for kitchen_sink demo.

It seems that a lot of connections are opened (~90 `handle_req_work_conn` launched). Those are started when FRP "asks for it" (code `114`). Any idea why on a very complex page the number of sockets have to be that high ?  For the basic hello world example, we have ~13 threads. Maybe that's why at some point a socket is closed server-side which causes the BrokenPipe issue.

Anyway, I created a branch and a PR to share my debugging (see https://github.com/gradio-app/gradio/pull/2615). In particular, I have added logs for when threads are launched/closed and when sockets are written/closed. I think it has to be debugged alongside the FRP server itself.

abidlabs:
Thanks so much @Wauplin for looking into this. I have no idea why some connections are created, but it's good to know what might be the likely culprit. Will look into seeing if we can possibly reduce the # of connections

XciD:
Just wondering if it's not related to TCPMux, 
I've not implemented it, I think the purpose is to re-use the same connection for multiple calls.

Python implementation: https://github.com/xtqxk/pyamux

abidlabs:
We checked the logs on the server. The only thing that we could see that corresponded to the disconnect was this line:

```
read error: read tcp 172.31.14.105:7000->69.107.139.81:54835: read: connection reset by peer
```

This indicates to me that the server is saying that the *client* is resetting the connection. However, on the client, we see this error: 

```
ConnectionResetError: [WinError 10054] An existing connection was forcibly closed by the remote host
```

which to me seems like its blaming the server. We couldn't make much progress unfortunately

abidlabs:
Hi everyone! Thanks to some incredible work by @XciD and @Wauplin, we have now revamped sharing Gradio demos (using `share=True`) to use a much more secure mechanism: FRP. Because this represents a major shift into how we do tunneling AND because we do somethings differently depending on what platform you are running on, it would be GREAT if everyone can test this out and share any issues they find below! 

👇👇👇👇👇👇👇👇👇👇👇👇👇👇👇👇👇👇👇👇👇👇👇👇

freddyaboulton (commented):
Tested the following demos and they all work great! Thanks @Wauplin @XciD 

Will test more tomorrow

```
Titanic_survival
stream_audio
video_identity
fake_gan
image_mod_default_image
fake_diffusion_with_gif
cancel_events
zip_files
csv_upload
```

Wauplin:
Following @abidlabs's [message on slack](https://huggingface.slack.com/archives/C02SPHC1KD1/p1670572997824979?thread_ts=1670334435.819239&cid=C02SPHC1KD1) (internal link) about issues to run FRP server in a notebook due to asyncio, I investigated it and decided to completely remove the async stuff as it was more painful that anything.

Pushed the fix in https://github.com/gradio-app/gradio/pull/2509/commits/c0b6801936840387a87711ceb9a7d8efcf74b38e. Changes are:
1. Normal execute of subprocess (no asyncio)
2. No need to have a pending thread that runs indefinitely (since to loop to run anymore)
3. Changed the `CURRENT_TUNNEL` singleton to a `CURRENT_TUNNELS` list in case someone runs several demos in the same notebook.
4. Registered to kill the subprocess (see [`atexit.register`](https://docs.python.org/fr/3/library/atexit.html)) when script exits. It's not bullet-proof since if Python is killed too hard the cleanup code is never called. In general, it will happen that some users get pending processes running on their machine and I don't think we can avoid that :confused:

abidlabs:
> Normal execute of subprocess (no asyncio)
> No need to have a pending thread that runs indefinitely (since to loop to run anymore)
> Changed the CURRENT_TUNNEL singleton to a CURRENT_TUNNELS list in case someone runs several demos in the same notebook.
> Registered to kill the subprocess (see [atexit.register](https://docs.python.org/fr/3/library/atexit.html)) when script exits. It's not bullet-proof since if Python is killed too hard the cleanup code is never called. In general, it will happen that some users get pending processes running on their machine and I don't think we can avoid that 😕

Amazing @Wauplin! Testing right now

abidlabs:
Thank you so much @Wauplin, tested and looks awesome! 

I just made a beta release `gradio==3.12.0b7`. Let's do some more testing early next week and plan to release mid-next week if everything looks good.

abidlabs:
Testing has gone quite well! @aliabid94 is going to do some load-testing -- assuming that goes well, we should be good to merge tomorrow.

abidlabs:
Thank you everyone, particularly @XciD and @Wauplin for putting this together. Excited to get this out to all of our users :)

abidlabs:
As a security precaution, we haven't released the full configuration of our FPRS server. We may consider doing so in the future

speaknowpotato:
> As a security precaution, we haven't released the full configuration of our FPRS server. We may consider doing so in the future

hi @abidlabs , is it possible use my own FRPS server for the shared link? 
if yes, would you mind sharing some sample code for the FRPS server configuration. thanks!

Issue #2050, opened 2022-08-19
Make it easier to work with streaming data


freddyaboulton:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
A data scientist might be interested in building a dashboard that pulls live data periodically. 

For example, every five minutes visualize the locations of the most popular tweets talking about Gradio. 
Every hour, refresh a plot showing our sales trend over the last quarter. 

**Describe the solution you'd like**  
Add an event trigger that sets a time delta when the page should be refreshed.

```python
with gr.Blocks() as demo:
    ....
    demo.refresh(fn=pull_data, inputs=[...], outputs=[...], on=gr.TimeDelta.Minutes * 5)
```

Open question: Build this for interfaces?

**Additional context**  
Might lead to interesting demos but not an immediate priority for Data Science milestone

abidlabs:
I wonder if it would make sense to use the `streaming` parameter that we have for Image/Audio applied to other components, or to the demo as a whole?

PR #2512, opened 2022-10-21
Run events every given number of seconds

freddyaboulton:
# Description
Adds the `every` key word argument to the events so that they run `every` number of seconds.

Fixes: #2050

Demo
```python
import math
import gradio as gr
import datetime
import plotly.express as px
import numpy as np


def get_time():
    return datetime.datetime.now()


plot_end = 2 * math.pi


def get_plot(period=1):
    global plot_end
    x = np.arange(plot_end - 2 * math.pi, plot_end, 0.02)
    y = np.sin(2*math.pi*period * x)
    fig = px.line(x=x, y=y)
    plot_end += 2 * math.pi
    return fig


with gr.Blocks() as demo:
    with gr.Row():
        with gr.Column():
            c_time2 = gr.Textbox(label="Current Time refreshed every second")
            gr.Markdown("Change the value of the slider to automatically update the plot")
            period = gr.Slider(label="Period of plot", value=1, minimum=0, maximum=10, step=1)
            plot = gr.Plot(label="Plot (updates every half second)")
        with gr.Column():
            name = gr.Textbox(label="Enter your name")
            greeting = gr.Textbox(label="Greeting")
            button = gr.Button(value="Greet")
            button.click(lambda s: f"Hello {s}", name, greeting)

    demo.load(lambda: datetime.datetime.now(), None, c_time2, every=1)
    dep = demo.load(get_plot, None, plot, every=0.5)
    period.change(get_plot, period, plot, every=0.5, cancels=[dep])

if __name__ == "__main__":
    demo.queue().launch()
```
live_demo

# Checklist:

# A note about the CHANGELOG


abidlabs:
So cool! Also `every` and `cancels` are a nice combo!

gradio/utils.py
```diff
@@ -689,6 +690,39 @@ def is_update(val):
...
+            if called_before:
+                time.sleep(every)
+            else:
+                called_before = True
+            yield output
+
+    return continuous_fn
+
+
+async def cancel_tasks(task_ids: List[str]):
```
freddyaboulton, line 707: Refactoring some of the code from the cancels PR so that it's easier to use in other parts of the codebase

gradio/blocks.py
```diff
@@ -983,8 +1002,9 @@ def load(
         api_key: Optional[str] = None,
         alias: Optional[str] = None,
         _js: Optional[str] = None,
+        every: None | int = None,
```
abidlabs, line 1005: Unrelated to your PR, this event trigger is missing many arguments, such as batch. It would be good to add if possible (or we can create a separate issue for it)

gradio/blocks.py
```diff
@@ -134,6 +139,7 @@ def set_event_trigger(
         batch: bool = False,
         max_batch_size: int = 4,
         cancels: List[int] | None = None,
+        every: int | None = None,
```
abidlabs, line 142:
```suggestion
        every: float | None = None,
```

gradio/utils.py
```diff
@@ -689,6 +690,39 @@ def is_update(val):
...
+        called_before = False
+        while True:
+            output = fn(*args)
+            if called_before:
+                time.sleep(every)
+            else:
+                called_before = True
+            yield output
+
+    return continuous_fn
```
abidlabs, line 704:
```suggestion
def get_continuous_fn(fn, every):
    def continuous_fn(*args):
        while True:
            output = fn(*args)
            yield output
            time.sleep(every)
    return continuous_fn
```

Also, would it be better to use `asyncio.sleep` here instead of `time.sleep` so that the queue worker is free to do something else in the meantime?

abidlabs: Ah I see so you later avoid putting such events in the queue altogether because they don't get completed. So I'm not sure how much of a difference it makes to use `asyncio.sleep` but I do think it's probably better than having a background thread just sit there doing nothing

abidlabs:
This looks great @freddyaboulton! Tested it and works well, and plays nicely with `cancels`. 

One suggestion would be to add some more documentation about it (maybe [in this Guide](https://gradio.app/blocks_and_event_listeners/)?) since this is a pretty big feature.  I would also mention somewhere that if you define an event that runs `every` seconds, this does not take into account the runtime of the function itself -- so a 2 second prediction function running with `every=5` actually runs every 7 seconds.

abidlabs (approved):

pngwn (approved):
lgtm! Awesome work @freddyaboulton!

gradio/blocks.py
freddyaboulton: I'll do that in a separate PR! Thank you for pointing this out!

gradio/utils.py
freddyaboulton: Lol love this refactor! Thank you.

Yea originally I had this as an async function but the problem with that is that you only want to sleep after the first function call. Tricky to implement without a generator and we don't currently support async generators. Will think about it though (may add async generators to #2552 )

PR #2528, opened 2022-10-25
Remove special ws path for spaces

abidlabs:
As [discussed internally](https://huggingface.slack.com/archives/C02Q9UFS8AW/p1666267765528299), it is no longer necessary to have a different URL for websocket connections on Spaces. 

Quick PR to fix that.


freddyaboulton:
@abidlabs The deployed space for this PR seems broken - does not respect the event triggers. See below:

image

abidlabs:
Merged in @pngwn's fix and now everything seems to be looking fine. @freddyaboulton if you can give this another quick look

freddyaboulton (approved):
Looks great @abidlabs ! Thanks for the fix

PR #2544, opened 2022-10-26
Dropdown for seeing docs as latest or main 

aliabd:
Allows users to see the docs for both main and latest release. Live on [staging](https://www.gradio-staging.com/docs/). 

Workflow: 
* Cronjob on the instance pulls from github every minute to see if there are any changes and if there are it builds the website using the Dockerfile 
* Dockerfile creates a gradio wheel and uploads it to s3, also uploads all the demos to spaces (with the suffix `_main`) and adds the wheel to their requirements. It also uploads all the demos on spaces to use the latest gradio version if a new one exists. 
* Dockerfile builds the website and the docs. The `/docs` page points to a static file that is already generated and saved in the repo named `v{X.X.X}_template.html `. The `/docs/main` page is built on every commit. There's a dropdown on each page for readers to switch between them. 
* When a new version is released (`version.txt` is changed in main), a github action opens a PR to add the new versioned docs (new `v{X.X.X}_template.html ` file).  The action installs the new version and builds the docs, so there should not be any conflicts, it also uses the versioned gradio.js instead of the built one in `/docs/main`, so it basically behaves entirely like a snapshot of the docs when the version was released. 

I tested the github action on a fork of the repo. You can see an example of it working [here](https://github.com/aliabd/gradio/pull/11) or failing [here](https://github.com/aliabd/gradio/actions/runs/3392712879) because the version was a prerelease/not latest (we don't want to show docs for prereleases)

Closes: #1650


website/homepage/upload_demos.py
```diff
@@ -29,21 +31,25 @@ def upload_demo_to_space(
...
+                            sdk_version: {gradio_version}
+                            app_file: run.py
+                            pinned: false
+                            ---
+                            """
+        readme.open("w").write(textwrap.dedent(readme_content))
+        
+        if gradio_wheel_url:
+            with open(os.path.join(tmpdir, "requirements.txt"), "a+") as r:
+                r.write(gradio_wheel_url)
```
freddyaboulton, line 51: I think we need a newline here:

https://huggingface.co/spaces/gradio/fraud_detector_main/blob/main/requirements.txt

aliabd: good catch thanks! fixed: https://huggingface.co/spaces/gradio/fraud_detector_main/blob/main/requirements.txt

.github/workflows/build-version-docs.yml
```diff
@@ -0,0 +1,44 @@
...
+      run: cd website && python check_version.py
+  build-docs:
+    needs: check-version
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install requirements
+        run: python -m pip install -r website/homepage/requirements.txt 
+      - name: Install gradio
+        run: python -m pip install gradio -U
```
freddyaboulton, line 35: We install from pip here right? And the reason is so that we do not use the local front-end code?

website/check_version.py
```diff
@@ -9,30 +9,15 @@
...
-            r = requests.head(f"https://pypi.org/project/gradio/{version}/", timeout=3, verify=False)
-        if r.status_code == 200:
-            return True
-    except (ConnectionError, requests.exceptions.ConnectionError):
-        return False
-
 def wait_for_version(version: str):
     for _ in range(10):
-        if is_version_up(version):
+        latest_gradio_stable = requests.get("https://pypi.org/pypi/gradio/json").json()["info"]["version"]
```
freddyaboulton, line 14: nit: Maybe it's worth putting this in its own helper as I see this logic is in separate files?

website/homepage/src/docs/__init__.py
```diff
@@ -101,12 +101,30 @@ def find_cls(target_cls):
...
+    output = template.render()
+    version_docs_file = os.path.join(output_folder, "index.html")
+    with open(version_docs_file, "w") as index_html:
+        index_html.write(output)
+
+def build_pip_template(version, jinja_env):
+    docs_files = os.listdir("src/docs")
+    for file in docs_files:
+        if file.endswith("_template.html") and file.startswith("v"):
+            os.remove(os.path.join("src/docs", file))
```
freddyaboulton, line 125: We don't need to remove the old version builds right? Might be good to keep them for just-in-case even though there are not displayed on the website?

CHANGELOG.md
```diff
@@ -22,7 +22,9 @@ The `api_name` parameter will take precendence over the `fn_index` parameter.
 * `gr.Blocks.load()` now correctly loads example files from Spaces [@abidlabs](https://github.com/abidlabs) in [PR 2594](https://github.com/gradio-app/gradio/pull/2594)
 
 ## Documentation Changes:
-No changes to highlight.
+* See docs for the latest gradio commit to main as well the latest pip release:
```
abidlabs, line 25: Let's move this up to 3.9 section

freddyaboulton (approved):
@aliabd Thank you for this useful PR! I reviewed the upload demo and buiild logic and it looks good to me. Left some questions but nothing blocking merge.

abidlabs:
LGTM @aliabd! One thing that is happening for me is that the select dropdown to toggle between `v3.9` and `main` sometimes takes a while to load and in the meantime I see this empty dropdown:

image

freddyaboulton:
Also these two spaces are not building because they specify incompatible gradio versions in the requirements.txt

Not sure if they are needed or we can delete.
image

abidlabs:
I don't see any reason why we've pinned the Gradio versions in these demos. We should be able to unpin and they'll probably work fine

CHANGELOG.md
aliabd: done

.github/workflows/build-version-docs.yml
aliabd: Not sure what you mean by local front-end code. 

We install from pip the latest released version because that's where we want to generate these docs from. Part of the docs comes from gradio as a module and not form the `/website` dir. For example `gradio.generate_documentation()`. 

But on what will show up in `/docs/main` we install gradio from the repo and not from pip (this happens inside the Dockerfile) 

Lmk if this isn't what you mean

website/check_version.py
aliabd: good idea, done

website/homepage/src/docs/__init__.py
aliabd: yup you're right, might be useful! done

aliabd:
@abidlabs I don't see this specific dropdown issue. Do you see that @freddyaboulton? 


https://user-images.githubusercontent.com/9021060/200069518-b96b8225-6898-4fe9-9a52-3fffb0702d7e.mov

That being said though, there is some really bad sluggishness on the website now. For example go https://gradio.app/docs and scroll really fast. you'll see it takes a while for things to render. I guess the page is just really heavy now and we'll have to look into that from the embedding side. 

No idea why those two demos pinned a gradio version to their requirements, fixed though and should auto update the spaces once this is merged.

freddyaboulton:
@aliabd The dev server behaves the same for me. The dropdown shows both options without lag. Switching versions takes about ~1 second during which the page hangs a bit. We can probably tune that with the other website performance improvements?

abidlabs:
Ok here's what it looks like for me:

Recording 2022-11-04 at 16 42 45

But I'm okay with getting this big one in and working on general website tune-ups in a separate PR.

abidlabs (approved):

aliabid94 (approved):
Just tested, LGTM!

Issue #2522, opened 2022-10-25
Avoid needing to cast audio file to `int16` for `Audio` component


abidlabs:
It seems that some audio files cannot be passed natively into the `Audio` component -- rather they need to be cast into `int16` before they can be played. E.g. this does not work:

```py
import gradio as gr
from datasets import load_dataset

speech_commands = load_dataset("speech_commands", "v0.02", split="test")

def generate_audio():
    example = speech_commands.shuffle()[0]
    audio = example["audio"]
    return audio["sampling_rate"], audio["array"]
    

demo = gr.Interface(fn=generate_audio, inputs=None, outputs="audio")
demo.launch()
```

whereas this does:

```py
import IPython.display as ipd

ipd.Audio(audio["array"], rate=audio["sampling_rate"])
```

Relevant internal discussion: https://huggingface.slack.com/archives/C02QZLG8GMN/p1666702652619089

abidlabs:
Thanks for the catch @freddyaboulton!

eervin123:
I'm not sure but it looks like this same thing is happening with `int32` formats. I'm getting the following warnings.

UserWarning: Trying to convert audio automatically from int32 to 16-bit int format.
  warnings.warn(warning.format(data.dtype))
  
  This is working, it is currently just a warning.

Ram-Pasupula:
I am getting the below error for int8 format.

Audio data cannot be converted automatically from int8 to 16-bit int format.

PR #2545, opened 2022-10-26
Adds support for audio samples in `float64`, `float16`, or `uint16` formats

abidlabs:
Previously, we would not be able to handle audio samples if they were provided as `numpy` arrays in `float64`, `float16`, or `uint16` formats. This adds support for that and closes: #2522 

thanks @nateraw @lewtun for flagging this


freddyaboulton (approved):
Looks good to me @abidlabs ! Would be good to add a unit test for this

abidlabs:
Thanks for the quick reviews @freddyaboulton! I added a unit test specifically for `float64` as that was the feature request we heard. For most of the others, it's basically just inverting the arithmetic so not sure how necessary the unit tests are.

PR #2558, opened 2022-10-28
A Guide on Configuring a Gradio Queue for High-Volume Traffic

abidlabs:
* Adds a Guide on Setting Up a Gradio Demo for Maximum Performance (closes: #2016)
* Renames the Guides to remove the parentheses


guides/06_other_tutorials/setting_up_a_gradio_demo_for_maximum_performance.md
```diff
@@ -0,0 +1,101 @@
...
+To address these limitations, any Gradio app can be converted to use **websockets** instead, simply by adding `.queue()` before launching an Interface or a Blocks. Here's an example:
+
+```py
+app = gr.Interface(lambda x:x, "image", "image")
+app.queue()  # <-- Sets up a queue with default parameters
+app.launch()
+```
+
+In the demo `app` above, predictions will now be sent over a websocket instead.
+Unlike POST requests, websockets do not timeout and they allow bidirectional traffic. On the Gradio server, a **queue** is set up, which adds each request that comes to a list. When a worker is free, the first available request is passed into the worker for inference. When the inference is complete, the queue sends the prediction back through the websocket tothe particular Gradio user who called that prediction.
```
freddyaboulton, line 29: Typo: No space between to and the. `the websocket tothe particular Gradio ... `

guides/06_other_tutorials/setting_up_a_gradio_demo_for_maximum_performance.md
```diff
@@ -0,0 +1,101 @@
...
+
+### The `max_batch_size` parameter
+
+Another way to increase the parallelism of your Gradio demo is to write your function so that it can accept **batches** of inputs. Most deep learning models can process batches of samples more efficiently than processing individual samples. 
+
+If you write your function to process a batch of samples, Gradio will automatically batch incoming requests together and pass them into your function as a batch of samples. You need to set `batch` to `True` (by default it is `False`) and set a `max_batch_size` (by default it is `4`) based on the maximum number of samples your function is able to handle. These two parameters can be passed into  `gr.Interface()` or to an event in Blocks such as `.click()`. 
+
+While setting a batch is conceptually similar to having workers process requests in parallel, it is often *faster* than setting the `concurrency_count` for deep learning models. The downside is that you might need to adapt your function a little bit to accept batches of samples instead of individual samples. 
+
+Here's an example of a saimple function that does *not* accept a batch of inputs -- it processes a single input at a time:
```
freddyaboulton, line 61: Typo: `saimple`

guides/06_other_tutorials/setting_up_a_gradio_demo_for_maximum_performance.md
```diff
@@ -0,0 +1,101 @@
...
+
+### The `max_batch_size` parameter
+
+Another way to increase the parallelism of your Gradio demo is to write your function so that it can accept **batches** of inputs. Most deep learning models can process batches of samples more efficiently than processing individual samples. 
+
+If you write your function to process a batch of samples, Gradio will automatically batch incoming requests together and pass them into your function as a batch of samples. You need to set `batch` to `True` (by default it is `False`) and set a `max_batch_size` (by default it is `4`) based on the maximum number of samples your function is able to handle. These two parameters can be passed into  `gr.Interface()` or to an event in Blocks such as `.click()`. 
+
+While setting a batch is conceptually similar to having workers process requests in parallel, it is often *faster* than setting the `concurrency_count` for deep learning models. The downside is that you might need to adapt your function a little bit to accept batches of samples instead of individual samples.
```
freddyaboulton, line 59: Do we have an answer as to whether `concurrency_count` plays well on gpus? 

I wonder if we should mention that `concurrency_count` is better suited to IO-bound demos and `batch_size` is better suited to CPU/GPU bound demos. You kind of hint at that below where you say that high batch size likely means that concurrency count should be set to 1.

guides/06_other_tutorials/setting_up_a_gradio_demo_for_maximum_performance.md
```diff
@@ -0,0 +1,101 @@
+# Setting Up a Gradio Demo for Maximum Performance
+
+Let's say that your Gradio demo goes *viral* on social media -- you have lots of users trying it out simultaneously, and you want to provide your users with the best possible experience or, in other words, minimize the amount of time that each user has to wait in the queue to see their prediction.
+
+How can you configure your Gradio demo to handle the most traffic? In this Guide, we dive into some of the parameters of Gradio's `.queue()` method as well as some other related configuations, and discuss how to set these parameters in a way that allows you to serve lots of users simultaneously withminimal latency.
```
freddyaboulton, line 5: typo: `configuations`

freddyaboulton (approved):
Awesome guide @abidlabs ! I love the structure and concrete recommendations. Noticed a couple of typos but that's about it.

Is it worth mentioning `api_open` in the sense that "closing" the api may help with scaling by making sure the queue isn't skipped?

abidlabs:
Thanks for the review @freddyaboulton! Will fix the typos and add a section on `api_open` as well as we've seen at least one concrete example of that being an issue

guides/06_other_tutorials/setting_up_a_gradio_demo_for_maximum_performance.md
abidlabs: Checking this right now

abidlabs:
@aliabd I'm getting a weird behavior when I try to add my Guide to the `Interface` class and build the website locally: only this Guide is showing up underneath the docs for the `Interface` class:

image


----

What's weird is that if I remove my additional guide from the `Interface` docstring (in other words, revert all changes to the `interface.py` file), not a single Guide shows up:

image


Even though the `Interface` class lists 6 guides here:

```py
    Guides: quickstart, key_features, sharing_your_app, interface_state, reactive_interfaces, advanced_interface_features
```

Do you know what might be going on?

guides/06_other_tutorials/setting_up_a_gradio_demo_for_maximum_performance.md
abidlabs: I assigned a GPU to this Space and it seems to work just fine: https://huggingface.co/spaces/abidlabs/image-classifier. Specifically, I was able to process requests in parallel and cut the latency in half on average by using `concurrency_count=2`!

A user just needs to keep in mind that their GPU memory might be different than their CPU memory so they need to ensure that multiple workers will not OOM their GPU. I'll add a note in the hardware section

guides/06_other_tutorials/setting_up_a_gradio_demo_for_maximum_performance.md
```diff
@@ -0,0 +1,112 @@
...
+
+There are several parameters that can be used to configure the queue and help reduce latency. Let's go through them one-by-one.
+
+### The `concurrency_count` parameter
+
+The first parameter we will explore is the `concurrency_count` parameter of `queue()`. This parameter is used to set the number of worker threads in the Gradio server that will be processing your requests in parallel. By default, this parameter is set to `1` but increasing this can **linearly multiply the capacity of your server to handle requests**.
+
+So why not set this parameter much higher? Keep in mind that since requests are processed in parallel, each request will consume memory to store the data and weights for processing. This means that you might get out-of-memory errors if you increase the the `concurrency_count` too high.
+
+**Recommendation**: Increase the `concurrency_count` parameter as high as you can until you hit memory limits on your machine. You can [read about Hugging Face Spaces machine specs here](https://huggingface.co/docs/hub/spaces-overview).
```
FarukOzderim, line 43: Hello @abidlabs, a beautiful guide as always ❤️ 

IMO this recommendation is not a good one, because increasing concurrency does not directly translate into performance due to various reasons as costs associated with context switching and GIL limitations. So I would suggest to change it to smt like, 

'Increase concurrency as long as you see improvement in the throughput.'

abidlabs: Ah I see, thanks for the suggestion @FarukOzderim! I'll update the Guide to reflect that

PR #2576, opened 2022-10-31
Theme base

pngwn:
# Description

The beginnings of a CSS variable driven theming system.

This is unfinished as i still have a bunch of grunt work to do but you can get an idea of what is happening.

We havea theme defined in a JS object to populate the initial values, some of those values come directly from tailwind. Some of the definition are very general `colors.background.primary` and others are more specific `chatbot.bot.background.color.latest`. Most of the more specific variables are fed from the more general tokens. 

This allows quite a large amount of flexibility when creating a theme. If a user just wants to change those base tokens, they will cascade across the app and apply the theme everywhere that is appropriate. Likewise if they want to theme certain things very differently from other parts of the app then they can do that too. There is also the middle option (probably the most common), redefining the base tokens and then adding a few more specific tweaks here and there.

Currently this isn't integrated into our tooling at and is a simple script. There is a function that walks through the objects and turns them into a css variable based on the object path. Anything there is a normal value then it uses that literal value, where it sees dot notation, that is an indication that the variable should reference _another_ variable. That why the weird dot notation is scattered around.

I've commited the generated CSS file for now so everyone can see what that will look like but that will probably be ignored when this is actually ready.

The tests give a good idea of how the objects gets mapped to css variables.

lmk if you have any questions.

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

Closes: # (issue)

# Checklist:

# A note about the CHANGELOG


pngwn:
**Context**

The purpose of this work is two fold. 

- Implement a comprehensive theming system
- Refactor/ improve styling so that it is maintainable, resilient, and easily understandable.

**Why are our styles not maintainable?**

- Specificity issues due to use of global component styles.
- Difficult to understand components due to global styles being defined outside of the components themselves. More than just being difficult to reason about this actually breaks the component model to a certain degree as we have little confidence and no guarantees that component style changes will not impact other parts of the application (which is the point of the component model).
- Difficult to reason about dark/ light mode. The way dark mode is implemented is by overwriting the values of the light mode  (default) colors (i.e. `.dark .bg-gray-100 { @apply bg-gray-800 }`). This is incredibly confusing. Additionally we _sometimes_ use the dark tailwind modifier as well. It is almost impossible to tell what a component is supposed to look like in dark mode when looking at a component.
- Overuse of `@apply`. This is broadly considered a bad practice for a variety of reasons (further info from the official docs: https://tailwindcss.com/docs/reusing-styles#avoiding-premature-abstraction)

Due to the above reasons we have a huge amount of duplicated styles (because we don't know what already exists and it isn't easy to discover), we have a confusing mix of different styling approaches using tailwind, our CSS is starting to fall into the 'append only' trap that was common before style scoping was a thing (because we don't know if style modifications will break other things, we often duplicate things that already exist), we actually aren't getting much out of tailwind (as discussed in the above tailwind link) because we are using a significant amount of global component styles.

**Theming**

In order to implement theming we need a few things, tailwind doesn't make any of them particularly easy. 

In order to make the tailwind work properly with our CSS variables we need to actually integrates these CSS variables into tailwind, otherwise many of the utility classes won't actually work. Aside from this we still need to _initialise_ these variables, create some custom classes for more specific variables values (in order to be consistent with tailwind's class-based approach) and then 'apply' them everywhere they are needed. This means theming would be very confusing as we would typically have the following flow if trying to understand a css class:

- See component style in components (e.g. `.gr-input`)
- Find `.gr-input` in global style files, see applied utility class(es) or base tailwind class(es). 
- Go to tailwind config file in order to find the css variable used for that particular class.
- Go _back_ to the global style file to see what _value_ was assigned to that specific css variable.

This is unnecessarily complicated and unreasonably difficult to understand (especially for OSS contributors). We could reduce some of these layers by refactoring some things but there is always going to be a degree of indirection. It also makes the theme packages pretty tightly coupled to tailwind as it would need to include configuration that is specific to tailwind and would not work with other approaches, not a huge deal but slightly annoying.

**Future**

As we look to publishing individual components, either on npm as svelte components or web components or on a cdn as web components (or both/ all three) the question of how do we ship our styles becomes more important. Tailwind as a monolithic css library works well for apps but it gets pretty complicated for users. If we don't ship tailwind then we are requiring users to add tailwind themselves (and then they have to match version etc), if we _do_ ship tailwind then they may end up with a duplicate tailwind if they _are_ using it, or just a large CSS bundle if they aren't (in addition to their own CSS solution). Utility css frameworks tend be heavier than some of the alternatives.

Publishing isn't on the immediate roadmap but I don't want to be blocked on this down the road as we have spoken extensively about wanting to publish individual part ins various ways. Factoring this into our decision making today will make our lives easier in the future.

**Is this actually tailwinds fault?**

On the theming side? Absolutely. That is just the nature of how tailwind's configuration works and the realty of how much control we need in our theming system. Just doesn't seem to be a particularly good fit.

On the maintenance side? Not particularly. But there is no doubt in my mind that the use of tailwind has lead us down this path to a degree. It is also true that we have no real safeguards against some of the issues and have to rely entirely on developer discipline, something that I am firmly against, especially given the pace at which we need to move and the fact that this is an open source library and we want to encourage contributions from people with a variety or experience levels. We can't _ban_ global styles because that is how tailwind works and we would still need _some_ utility styles.

We need something that not only supports but _guides us_ to the following:

- closely fits the component model
- offers guarantees that changing styles won't break anything
- theming can be implemented in a sensible straight forward way.

This isn't a criticism of tailwind particularly; tailwind is a tool and like any tool it is not inherently 'good' or 'bad'. I don't think tailwind is a good fit for our usecase and I don't think it is a good fit for our team. Tailwind can work very well in variety of different scenarios but it hasn't been working well for us, as the styling started bad and has progressively got worse. I especially think it is important that we think more carefully about safety and stability as usage grows, so scoped styles + embracing the component model (including encapsulation) are important.

**Next steps**

Unless someone can give me concrete solutions to the above challenges, I am going to start implementing theming with a CSS variable first approach. This is an internal change and will have no impact on users, everything will be identical.

Issue #2476, opened 2022-10-17
Using the "x" button opens file load dialog automatically.


mezotaken:
### Describe the bug

User clicks "x" button on loaded image -> Element is cleared, then load dialog pops up, the usual one for this ui element when clicked while it is empty. 
If this dialog is cancelled then the element can be used as usual, else it becomes impossible to load any image until after page reload. Drag&Drop use fixes it though. Obviously this dialog shouldnt open in the first place.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

It may be reproduced right on your own demo page under section "An Image example": https://gradio.app/getting_started/#an-image-example
But it originated here: [stable-diffusion-webui/2963](https://github.com/AUTOMATIC1111/stable-diffusion-webui/issues/2963)
It definitely happened somewhere in transition from gradio 3.4.1 to gradio 3.5.

### Screenshot

![GIF 17 10 2022 21-16-34](https://user-images.githubusercontent.com/32306715/196253349-b2194efd-28e8-4246-a509-3a61d6e41fb5.gif)

### Logs

Sadly, there were no errors logged when stable-diffusion-webui mentioned in reproduction was launched by me with --gradio-debug argument. This argument is used as a value for a wrapper call of gradio.Interface.launch().

### System Info

```shell
Gradio 3.5, Windows 10, Google Chrome
Gradio 3.5, Windows 11, Firefox
```

### Severity

serious, but I can work around it

abidlabs:
Can confirm the problem! @dawoodkhan82 would you be able to take a look?

PR #2577, opened 2022-10-31
Fixed upload dialog being showed after pressing "x" button

mezotaken:
# Description

The issue caused by commit https://github.com/gradio-app/gradio/commit/996cd8b814e7198cc50cf2f6356237828bce7350 as it seems. Variable `disable_click` being set to false every time when "x" button is used because at the moment it happens there exists no image.
~~`click` variable serves no purpose, not imported anywhere else in the project (correct me if i'm wrong).~~
To prevent this behavior stopPropagation property was applied to click event on the "x" buttons. Should be done for every similar button, but my main concern is for the images.

Also user cannot upload same image after clear, because upload is triggered with `on:change` event for `hidden_upload` element, so that's fixed too. 

Closes: #2476 

# Checklist:


abidlabs:
Thank you very much for the contribution @mezotaken. I've tested the code and confirms this solves the underlying issue.

@pngwn would you be able to review the code changes and if it looks good, we can merge this in?

mezotaken:
@abidlabs as i can see from the failed static check, the `click` variable is used after all. Albeit only once in the table.svelte and it does exactly what disable_click does? am i correct on this?
So changing 
image
to 
image
should be fine? I need confirmation on this

pngwn:
I won't have time to look over this for a week or so unfortunately.

dawoodkhan82:
> @abidlabs as i can see from the failed static check, the `click` variable is used after all. Albeit only once in the `table.svelte` and it does exactly what disable_click does? am i correct on this? So changing image to image should be fine? I need confirmation.

Yeah Looks like it was an unnecessary variable.

dawoodkhan82:
tested and it fixes the issue; also code looks good. LGTM!

abidlabs:
Thanks so much @mezotaken and @dawoodkhan82!

PR #2579, opened 2022-10-31
Refactoring `external.py`

abidlabs:
While working with `external.py`, I noticed that the module had gotten quite bloated and hard-to-read, so I refactored the file. Major changes:

* Moved many helper functions from `external.py` to `external_utils.py`
* Added some typing and changed `test_external.py` to use pytest-style tests
* Moved the functions to create an `Interface` from a `transformers.Pipeline` object to a separate `pipelines.py` and its test to a `test_pipelines.py`

As part of this refactoring, I noticed that we never tested `gr.Interface.from_pipeline()` method (the original test was a dud!). So I've added a real test for the function -- no other code logic was changed.


gradio/external.py
```diff
@@ -393,7 +310,7 @@ def query_huggingface_api(*params):
     return interface
 
 
-def get_spaces(space_name: str, api_key: str | None, alias: str, **kwargs) -> Blocks:
+def from_spaces(space_name: str, api_key: str | None, alias: str, **kwargs) -> Blocks:
```
freddyaboulton, line 313: Unrelated to this change but if `iframe_url` is none we should raise an exception

gradio/dataclasses.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
freddyaboulton, line 1: How come you added this?

gradio/external.py
abidlabs: Ah yes we can do that

abidlabs:
Thanks for the review @freddyaboulton! Will address the points and then merge

Issue #2062, opened 2022-08-23
Button for uploading images


abidlabs:
In some cases, the Image component takes up too much space, and it would be nice to have a more visually compact way to upload images (and other file types). In this case, a button for uploading images like this would be nice, something that looks like:

![image](https://user-images.githubusercontent.com/1778297/186172409-b419c787-c928-42cb-bda8-4c6e90c8703d.png)

Related Slack discussion: https://huggingface.slack.com/archives/C03HBN1C8CW/p1661180466071549

FYI we've seen requests to have a "Download file" button as well (#2029)

PR #2591, opened 2022-11-02
Add Upload Button

dawoodkhan82:
# Description
Add an upload button component. Like file upload but in button form. Main use case is for the new improved chatbot coming soon.

Not sure about the style of the button.

https://user-images.githubusercontent.com/12725292/199436347-e3ddff14-2ee0-4344-a477-a1a138b99c8e.mov

```
import gradio as gr

def test(name):
    return name.name

with gr.Blocks() as demo:
    image = gr.Image(interactive=False)
    upload_button = gr.UploadButton()
    upload_button.upload(fn=test, inputs=upload_button, outputs=image)

demo.launch()
```

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

Closes: #2062 

# Checklist:

# A note about the CHANGELOG


abidlabs:
Very cool @dawoodkhan82! I think this can be very useful in general (and also closes #2062 by the way).

A high-level question: have you considered whether `UploadButton` be its own component or should it be something that is generated by passing in a parameter to the regular `Button` class? On one hand, I feel like there is going to be a fair bit of duplicated code, but on the other hand, we might want to add parameters like the ability to restrict upload file types that are specific to uploading.

dawoodkhan82:
> 

Yeah I thought about this. It didn't make sense to create a parameter on the `Button` class since, we wanted to specify file types + maybe add more customization to `UploadButton` in the future.

abidlabs:
@pngwn what is `Upload`?

dawoodkhan82:
`Upload` is only a frontend component that's used by `File`, `Model3D`, `Video`, etc. as a generic upload component

dawoodkhan82:
@pete do you mean `File` component and `UploadButton` are too similar?

pngwn:
Oh yeah, I meant the various uploads we have i guess. In this case it seems `File` is the most similar.

abidlabs:
I'm on the fence about combining `File` and `UploadButton` -- there could certainly be shared parameters / internal logic, such as that for restricting file types, but on the other hand, there are lots of parameters that differ -- for example, "label" is something that applies to a `File`, but not an `UploadButton`, and "variant" is something that applies to `UploadButton` and not to `File`. But we could have both sets of parameters and only apply them depending on which version we are using.

One other suggestion though -- in the same way that a `File` can be used to upload and download, so this should this component in principle. So if we end up deciding to create a separate component, perhaps we call it a `FileButton`?

pngwn:
I don't really think they should be merged, I don't think that have overlapping APIs and that could be confusing.

The only reason I raise it is that we have rejected component requests in the past for being too niche, I don't know which Upload (Button or File) is the most useful but having both seems overkill. And I don't want us to get into a situation where we overlook this discrepancy when we need something (for whatever reason) but outside contributors have their proposals rejected. Typically we've told people to wait for custom components when this has happened in the past.

I'm not against the idea of an UploadButton particularly.

dawoodkhan82:
Hmm could it be possible, to add more style parameters to the `File` component to make it more button-ish. Maybe that could be an alternative to making a new component.

abidlabs:
I synced with @dawoodkhan82 on this a bit, and here's what we're thinking:

*  Let's make this component a variant of the `File` component given the numerous similarities. We can add a parameter (what's a good name? "layout"?) that allows a user to select between the two variants (the "button" vs "wide layout")

pngwn:
Maybe `variant="panel"` and `variant="button"`?

dawoodkhan82:
@pngwn as discussed, kept the upload button as a separate component.

CHANGELOG.md
```diff
@@ -50,10 +69,15 @@ No changes to highlight.
 No changes to highlight.
 
 ## Full Changelog:
+<<<<<<< HEAD
```
abidlabs, line 72: Looks like this merge conflict still needs to be resolved

ui/packages/app/src/components/File/File.svelte
```diff
@@ -52,6 +52,6 @@
 			upload_text={$_("interface.click_to_upload")}
 		/>
 	{:else}
-		<File value={_value} {label} {show_label} {file_count} />
```
abidlabs, line 55: Why did we remove `file_count`? I thought it was needed to be able to upload multiple files or a directory

gradio/components.py
```diff
@@ -2761,6 +2761,138 @@ def style(self, *, full_width: Optional[bool] = None, **kwargs):
...
+
+    def __init__(
+        self,
+        value: Optional[str | List[str] | Callable] = None,
+        *,
+        visible: bool = True,
+        elem_id: Optional[str] = None,
+        type: str = "file",
+        file_type: str = "file",
+        label: str = None,
```
abidlabs, line 2782: Could we have the `UploadButton` support multiple files similar to the `File` component through a `file_count` parameter?

And could we add support for `file_type` in the `File` component? 

Would be good to have a unified interface as much as possible IMO.

gradio/components.py
```diff
@@ -2761,6 +2761,138 @@ def style(self, *, full_width: Optional[bool] = None, **kwargs):
...
+        **kwargs,
+    ):
+        """
+        Parameters:
+            value: Default file to display, given as str file path. If callable, the function will be called whenever the app loads to set the initial value of the component.
```
abidlabs, line 2787: I think the `value` parameter doesn't do anything right now, correct? It would be better if the `value` parameter is the text that is displayed on the button, similar to the `Button` component.

gradio/components.py
```diff
@@ -2761,6 +2761,138 @@ def style(self, *, full_width: Optional[bool] = None, **kwargs):
...
+            type: Type of value to be returned by component. "file" returns a temporary file object whose path can be retrieved by file_obj.name and original filename can be retrieved with file_obj.orig_name, "binary" returns an bytes object.
+            file_type: Type of file to be uploaded. "file" allows any file to be uploaded, "image" allows only image files to be uploaded, "audio" allows only audio files to be uploaded, "video" allows only video files to be uploaded, "text" allows only text files to be uploaded.
+            label: Text to display on the button. Defaults to "Upload a File".
```
abidlabs, line 2790: When I tested the `label` parameter, it seems like it wasn't having an effect:
image

gradio/components.py
```diff
@@ -2761,6 +2761,138 @@ def style(self, *, full_width: Optional[bool] = None, **kwargs):
...
+        type: str = "file",
+        file_type: str = "file",
+        label: str = None,
+        **kwargs,
+    ):
+        """
+        Parameters:
+            value: Default file to display, given as str file path. If callable, the function will be called whenever the app loads to set the initial value of the component.
+            type: Type of value to be returned by component. "file" returns a temporary file object whose path can be retrieved by file_obj.name and original filename can be retrieved with file_obj.orig_name, "binary" returns an bytes object.
+            file_type: Type of file to be uploaded. "file" allows any file to be uploaded, "image" allows only image files to be uploaded, "audio" allows only audio files to be uploaded, "video" allows only video files to be uploaded, "text" allows only text files to be uploaded.
```
abidlabs, line 2789: Suggestion: maybe allow users to provide a list of extensions as well?

abidlabs:
Very cool @dawoodkhan82! I tested and the overall it works fantastic. Was able to use it to input a variety of file types. I have a few suggestions / questions above^

gradio/components.py
dawoodkhan82: Oh right, meant to remove it. I added a label param instead, but I guess that isn't consistent with `gr.Button`. Will change it.

dawoodkhan82: Actually using `value` won't work because the `postprocess` is called on the initial value and in this case the upload button postprocesses files. Am I right here?

abidlabs: Do we even need a `postprocess()` function as `UploadButton` component can only be used as an input?

dawoodkhan82: Right, we actually don't need postprocess. Removed it

dawoodkhan82: And made it so that `value` is the button label

CHANGELOG.md
```diff
@@ -1,6 +1,25 @@
 # Upcoming Release 
 
 ## New Features:
+
+### Upload Button
+There is now a new component called the `UploadButton` which is a file upload component but in button form! You can also specify what file types it should accept (ex: `image`, `video`, `audio`, `text`, or generic `file`).
```
abidlabs, line 6:
```suggestion
There is now a new component called the `UploadButton` which is a file upload component but in button form! You can also specify what file types it should accept (ex: `image`, `video`, `audio`, `text`, or generic `file`). Added by [@dawoodkhan82](https://github.com/dawoodkhan82) in [PR 2591](https://github.com/gradio-app/gradio/pull/2591)
```

gradio/components.py
```diff
@@ -2761,6 +2765,118 @@ def style(self, *, full_width: Optional[bool] = None, **kwargs):
         return IOComponent.style(self, **kwargs)
 
 
+@document("click", "style")
```
abidlabs, line 2768:
```suggestion
@document("click", "upload", "style")
```

abidlabs:
Thanks for the fix @dawoodkhan82 -- I can confirm that `value` parameter works as expected for me. Noticed a couple of other things I added to the review, but besides that, the backend code looks good. Would be great to get a review from @pngwn or @aliabid94 as well for the svelte

gradio/components.py
dawoodkhan82: done

abidlabs:
Hmm not a fan of `button_label` as a parameter name. I suppose we could just call it `label` now, right?

The other thing is that I think this parameter (button_label or label) should be the first argument, so that you could instantiate an `UploadButton` more easily. Something like:

`gr.UploadButton("Click to Upload")`

rather than having to do:

`gr.UploadButton(label="Click to Upload")`

aliabid94:
Just seeing this now - I agree with @pngwn's original opinion that this should just be a kwarg to the File component. Especially the variant solution, that seems very clean. I think the `variant="button"` could be applied to all upload media types, e.g. Audio, Video, File, etc. And then the `label=` kwarg just refers to the button text.

ui/packages/app/src/components/UploadButton/UploadButton.svelte
```diff
@@ -0,0 +1,41 @@
...
+
+	async function handle_upload({ detail }: CustomEvent<FileData>) {
+		value = detail;
+		await tick();
+		dispatch("change", value);
+		dispatch("upload", detail);
+	}
+
+	const dispatch = createEventDispatcher<{
+		change: FileData | null;
```
aliabid94, line 24: This component is only capable of change and upload, right?

ui/packages/app/src/components/UploadButton/types.ts
```diff
@@ -0,0 +1,7 @@
+export interface FileData {
```
aliabid94, line 1: You've defined this twice, can you reuse and import from the other? Ideally, we should merge all the definitions of interface FileData into `ui/packages/app/src/components/types.ts` instead of reimplenting over and over, if you can do that that'd be great.

aliabid94 (approved):
LGTM besides small comments

gradio/components.py
dawoodkhan82: done

gradio/components.py
```diff
@@ -2761,6 +2765,139 @@ def style(self, *, full_width: Optional[bool] = None, **kwargs):
...
+    def __init__(
+        self,
+        label: str = "Upload a File",
+        value: Optional[str | List[str] | Callable] = None,
+        *,
+        visible: bool = True,
+        elem_id: Optional[str] = None,
+        type: str = "file",
+        file_count: str = "single",
+        file_type: List[str] = ["file"],
```
abidlabs, line 2787: Tested and this works great! I like how you've made this parameter a list, but I think it's better to avoid using [lists as default arguments](https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/). I would set it to `None`, which allows any file to be uploaded. 

Secondly, I would rename this parameter to be `file_types` which in my mind makes it clearer that it expects a list.

abidlabs: Thirdly, when I tried passing in a regular string instead of a list, the frontend crashed. It would better to raise an exception

gradio/components.py
```diff
@@ -2761,6 +2765,139 @@ def style(self, *, full_width: Optional[bool] = None, **kwargs):
...
+        **kwargs,
+    ):
+        """
+        Parameters:
+            value: Default text for the button to display.
+            type: Type of value to be returned by component. "file" returns a temporary file object whose path can be retrieved by file_obj.name and original filename can be retrieved with file_obj.orig_name, "binary" returns an bytes object.
+            file_count: if single, allows user to upload one file. If "multiple", user uploads multiple files. If "directory", user uploads all files in selected directory. Return type will be list for each file in case of "multiple" or "directory".
+            file_type: Type of files to be uploaded. "file" allows any file to be uploaded, "image" allows only image files to be uploaded, "audio" allows only audio files to be uploaded, "video" allows only video files to be uploaded, "text" allows only text files to be uploaded.
```
abidlabs, line 2795: This description needs to be updated to reflect the fact that the parameter now takes in a list

gradio/components.py
```diff
@@ -2086,6 +2086,7 @@ def __init__(
         value: Optional[str | List[str] | Callable] = None,
         *,
         file_count: str = "single",
+        file_type: List[str] = ["file"],
```
abidlabs, line 2089: See my comments on below on the `file_type` parameter of the `UploadButton` component

gradio/components.py
```diff
@@ -2761,6 +2765,139 @@ def style(self, *, full_width: Optional[bool] = None, **kwargs):
...
+        type: str = "file",
+        file_count: str = "single",
+        file_type: List[str] = ["file"],
+        **kwargs,
+    ):
+        """
+        Parameters:
+            value: Default text for the button to display.
+            type: Type of value to be returned by component. "file" returns a temporary file object whose path can be retrieved by file_obj.name and original filename can be retrieved with file_obj.orig_name, "binary" returns an bytes object.
+            file_count: if single, allows user to upload one file. If "multiple", user uploads multiple files. If "directory", user uploads all files in selected directory. Return type will be list for each file in case of "multiple" or "directory".
```
abidlabs, line 2794: file_count="directory" does not seem to work for `UploadButton`. It works fine for `File` component

abidlabs:
Pointed out a few remaining nits, but otherwise looks great. Very cool @dawoodkhan82!

abidlabs (approved):

ui/packages/app/src/components/File/File.svelte
dawoodkhan82: Only actually used in the `FileUpload` component, to determine what can be uploaded.

gradio/components.py
```diff
@@ -2761,6 +2765,139 @@ def style(self, *, full_width: Optional[bool] = None, **kwargs):
...
 
 
+@document("click", "upload", "style")
+class UploadButton(Clickable, Uploadable, IOComponent, SimpleSerializable):
+    """
+    Used to create an upload button, when cicked allows a user to upload files that satisfy the specified file type or generic files (if file_type not set).
+    Preprocessing: passes the uploaded file as a {file-object} or {List[file-object]} depending on `file_count` (or a {bytes}/{List{bytes}} depending on `type`)
+    Postprocessing: expects function to return a {str} path to a file, or {List[str]} consisting of paths to files.
+    Examples-format: a {str} path to a local file that populates the component.
+    Demos:
```
abidlabs, line 2775:
```suggestion
    Demos: upload_button
```

abidlabs:
Did another round of testing and looks good @dawoodkhan82! Awesome PR

PR #2593, opened 2022-11-02
Call fn by api_name when using Blocks.load

### LINKED ISSUE #2523: Use the api_name when calling an app like a function

Properly using a space like an api requires knowing the `fn_index`. If you're a downstream consumer of an app, you may not know the `fn_index` and the fn_index can change if the upstream app is refactored.

Would be nice to be able to do something like 

```python
demo = gr.Blocks.load("path-to-demo")
output = demo(input, api_name="prediction-route")
```

Where "prediction-route" is the api_name given to the event in the upstream space.

**Additional context**

freddyaboulton:
# Description
Fixes #2523

Added unit tests. Can also test with some of the spaces we have in the gradio org.

# Checklist:

# A note about the CHANGELOG

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -45,7 +45,7 @@ from gradio.documentation import (
     document_component_api,
     set_documentation_group,
 )
-from gradio.exceptions import DuplicateBlockError
+from gradio.exceptions import DuplicateBlockError, InvalidApiName
 from gradio.utils import (
     check_function_inputs_match,
     component_or_layout_class,
@@ -734,7 +734,7 @@ class Blocks(BlockContext):
 
         return True
 
-    def __call__(self, *inputs, fn_index: int = 0):
+    def __call__(self, *inputs, fn_index: int = 0, api_name: str = None):
         """
         Allows Blocks objects to be called as functions. Supply the parameters to the
         function as positional arguments. To choose which function to call, use the
@@ -743,7 +743,19 @@ class Blocks(BlockContext):
         Parameters:
         *inputs: the parameters to pass to the function
         fn_index: the index of the function to call (defaults to 0, which for Interfaces, is the default prediction function)
+        api_name: The api_name of the dependency to call. Will take precedence over fn_index.
         """
+        if api_name is not None:
+            fn_index = next(
+                (
+                    i
+                    for i, d in enumerate(self.dependencies)
+                    if d.get("api_name") == api_name
+                ),
+                None,
+            )
+            if fn_index is None:
+                raise InvalidApiName(f"Cannot find a function with api_name {api_name}")
         if not (self.is_callable(fn_index)):
             raise ValueError(
                 "This function is not callable because it is either stateful or is a generator. Please use the .launch() method instead to create an interactive user interface."
--- a/test/test_external.py
+++ b/test/test_external.py
@@ -11,6 +11,7 @@ import pytest
 import gradio
 import gradio as gr
 from gradio import utils
+from gradio.exceptions import InvalidApiName
 from gradio.external import (
     TooManyRequestsError,
     cols_to_rows,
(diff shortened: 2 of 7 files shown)


gradio/blocks.py
```diff
@@ -743,7 +743,19 @@ def __call__(self, *inputs, fn_index: int = 0):
...
         """
+        if api_name is not None:
+            fn_index = next(
+                (
+                    i
+                    for i, d in enumerate(self.dependencies)
+                    if d.get("api_name") == api_name
+                ),
+                None,
+            )
```
abidlabs, line 756: Just out of curiosity, what is the advantage of this next(generator) approach over doing list.index()? I find the latter more readable

test/test_external.py
```diff
@@ -417,5 +418,20 @@ def test_respect_queue_when_load_from_config():
...
+def test_raise_value_error_when_api_name_invalid():
+    with pytest.raises(InvalidApiName):
+        demo = gr.Blocks.load(name="spaces/gradio/hello_world")
+        demo("freddy", api_name="route does not exist")
+
+
+def test_use_api_name_in_call_method():
+    # Interface
+    demo = gr.Blocks.load(name="spaces/gradio/hello_world")
+    assert isinstance(demo("freddy", api_name="predict"), str)
```
abidlabs, line 430: Suggestion -- instead of just checking the type, can we assert the correct value for a more stringent test?

test/test_external.py
```diff
@@ -417,5 +418,20 @@ def test_respect_queue_when_load_from_config():
...
+
+    app = gr.Blocks.load(name="spaces/gradio/queue-benchmark")
+    assert isinstance(app("freddy", api_name="text"), str)
```
abidlabs, line 433: Same as above. Instead of just checking the type, can we assert the correct value for a more stringent test? 

Also, I think it would be good to test with a different Space that (A) does not have a time.sleep (B) and has multiple functions that do different things to ensure that the correct function is being called

gradio/blocks.py
freddyaboulton: I like it cause in the case of the api_name not being there, it will return None instead of raising an exception.

Are you suggesting?

```python
try:
    fn_index = [d.get('api_name') for d in self.dependecies].index(api_name)
except ValueError:
    raise (....)    
```    

Happy to switch implementation!

abidlabs:
Thanks for adding this @freddyaboulton! Please see a few suggestions above

test/test_external.py
freddyaboulton: For sure!

test/test_external.py
freddyaboulton: This space has multiple functions. I'll make a copy without the sleep!

gradio/blocks.py
abidlabs: Yes that's what I was suggesting, but either way is fine! It probably speaks more to my personal comfort that I prefer the second over the first, I don't think there is one that's actually more pythonic

test/test_external.py
abidlabs: Yup great but it would also be good if the functions were doing different things to ensure that the correct function is being called

gradio/blocks.py
freddyaboulton: Yea I agree both are equivalently pythonic (or non pythonic 😆 ) 

I personally prefer if the failure cause of a function is None (or some value like that) as opposed to raising an exception 🤷

freddyaboulton:
Thanks for the review @abidlabs ! Updated the test to use this space which has two separate functions that do separate computations

https://huggingface.co/spaces/gradio/multiple-api-name-test

abidlabs (approved):
Amazing, thanks for addressing the suggestions @freddyaboulton!

Issue #2554, opened 2022-10-27
Load examples when Interface/Block is loaded unless examples are passed


merveenoyan:
When we load a gradio Interface from huggingface, the examples aren't loaded. I know this is intended behavior but maybe on top of it retrieving the example files when loading the Interface or Block would be good when they're loaded from a Space, _unless_ the user passes the examples parameter explicitly. WDYT? Also pinging @sayakpaul & @osanseviero

sayakpaul:
Even after passing examples, they did not show up in this Space (which uses multiple Spaces):

https://huggingface.co/spaces/sayakpaul/maxim-spaces

freddyaboulton:
Lol so it looks like we manually make the examples dataset invisible when loading from spaces 😬 

```python
    def streamline_config(config: dict) -> dict:
        """Streamlines the blocks config dictionary to fix components that don't render correctly."""
        # TODO(abidlabs): Need a better way to fix relative paths in dataset component
        for c, component in enumerate(config["components"]):
            if component["type"] == "dataset":
                config["components"][c]["props"]["visible"] = False
        return config
```

I think we should first fix the problem that @sayakpaul brings up about `Interface.load` not respecting the `examples` argument even when the examples are local. Later we can try to figure out a way to load examples from the upstream space.

sayakpaul:
Thanks!

abidlabs:
Haha yea that's my bad. Left that as a future TODO but didn't get a chance to come back to it. @freddyaboulton I can take this if you haven't started yet

sayakpaul:
@abidlabs has this been prodded yet? I still don't see the examples here: https://huggingface.co/spaces/sayakpaul/maxim-spaces.

abidlabs:
Hey @sayakpaul can you update the sdk_version in the readme on your Space to 3.9.1?

sayakpaul:
Just did! I didn't update the other spaces `maxim-spaces` shelters, though. 

Works perfectly.

PR #2594, opened 2022-11-02
Load example from external Spaces

abidlabs:
`gr.Blocks.load()` or `gr.Interface.load()` now correctly loads examples from Spaces. You can test by running this:

```py
import gradio as gr

gr.Interface.load("spaces/sayakpaul/raindrop-deraining-maxim").launch()
```

or a more complex example:

```py
import gradio as gr

gr.Interface.load("spaces/gradio/test-loading-examples").launch()
```

Loading more than 1 Space in one Blocks:

```py
import gradio as gr

with gr.Blocks() as demo:
    with gr.Tab("Demo 1"):
        gr.Interface.load("spaces/sayakpaul/raindrop-deraining-maxim")
    with gr.Tab("Demo 2"):
        gr.Interface.load("spaces/gradio/test-loading-examples")

demo.launch()
```

It is unfortunately not going to be possible to override examples when you load a gradio Interface or Blocks running 3.x. It is not possible to override *any* of the Interface properties (e.g. title) because we do not capture any reference to these Interface-specific properties. We instead load a generic Blocks that "happens" to look like an Interface. It might be possible to parse  Interface-specific properties but it will be quite complex and I do not think the added complexity is worth it.

Fixes: #2554


gradio/blocks.py
```diff
@@ -586,6 +605,8 @@ def get_block_instance(id: int) -> Block:
             block_config["props"].pop("type", None)
             block_config["props"].pop("name", None)
             style = block_config["props"].pop("style", None)
+            if block_config["props"].get("root_url") is None:
```
abidlabs, line 608: If the Space is already loading from another Space, it should use the root_url of the parent Space. This allows loading a chain of Spaces

gradio/blocks.py
```diff
@@ -603,6 +624,9 @@ def iterate_over_children(children_list):
                         iterate_over_children(children)
 
         with Blocks(theme=config["theme"], css=config["theme"]) as blocks:
+            # ID 0 should be the root Blocks component
+            original_mapping[0] = Context.root_block or blocks
```
abidlabs, line 628: Fixing a different bug that was preventing `gr.Interface.load()` from working if a dataset had example caching on

gradio/blocks.py
```diff
@@ -66,11 +66,20 @@
...
 
 class Block:
-    def __init__(self, *, render=True, elem_id=None, visible=True, **kwargs):
+    def __init__(
+        self,
+        *,
+        render: bool = True,
+        elem_id: str | None = None,
+        visible: bool = True,
+        root_url: str | None = None,  # URL that is prepended to all file paths
```
freddyaboulton, line 75: Should this be at the Blocks level instead? All individual Block(s) are within the same Blocks right?

gradio/external.py
```diff
@@ -398,7 +396,7 @@ def fn(*data):
             fns.append(fn)
         else:
             fns.append(None)
-    return gradio.Blocks.from_config(config, fns)
+    return gradio.Blocks.from_config(config, fns, iframe_url)
```
freddyaboulton, line 399: When loading a private space, the inputs are shown correctly but the outputs are.

image

Testing with this private space: https://huggingface.co/spaces/gradio/test-loading-examples-private

freddyaboulton: The boat image shows up on the left hand side but not in the dataset component. 

image

freddyaboulton: Curious what behavior you see, I ran it again on the private space and now everything but the model3d shows up 🤔 

image

gradio/blocks.py
```diff
@@ -618,13 +642,13 @@ def iterate_over_children(children_list):
...
                 ]
                 dependency.pop("status_tracker", None)
-                dependency["_js"] = dependency.pop("js", None)
                 dependency["preprocess"] = False
                 dependency["postprocess"] = False
 
                 for target in targets:
-                    event_method = getattr(original_mapping[target], trigger)
-                    event_method(fn=fn, **dependency)
+                    original_mapping[target].set_event_trigger(
```
freddyaboulton, line 649: How come we need to make this change?

freddyaboulton (commented):
Thanks for the fix @abidlabs ! Looks good.

Noticed some weird behavior when trying to load a private space. Posting it here so you can review.

gradio/external.py
abidlabs: Lol that's interesting. I get the same behavior as you the 2nd time -- everything but `Model3D` shows up. Nice catch, let me investigate

gradio/blocks.py
abidlabs: I put it at the `Block` level so that we can handle the case of loading from different Spaces within a single `Blocks`. E.g. 

```py
import gradio as gr

with gr.Blocks() as demo:
    with gr.Tab("Demo 1"):
        gr.Interface.load("spaces/sayakpaul/raindrop-deraining-maxim")
    with gr.Tab("Demo 2"):
        gr.Interface.load("spaces/gradio/test-loading-examples")

demo.launch()
```

gradio/external.py
abidlabs: This seems like a weird CORS issue and tbh I'm not sure what to do here. I think since this is a relatively niche use case (loading private Spaces, with (only?) the Model3D component), we can save for a future PR. I know @aliabid94 will be looking to CORS soon as well so flagging for your attention @aliabid94

abidlabs:
See my comments above @freddyaboulton, thanks for the review!

gradio/external.py
freddyaboulton: Sounds good with me. What makes you think it's a CORS issue?

gradio/blocks.py
freddyaboulton: Makes sense! Thank you for explaining.

gradio/external.py
freddyaboulton: Will also not display the dataframe properly in the dataset if it's a private space:

image

Works fine for public spaces though. Confused as to why it would not impact images,audio, etc. Agree we can leave it for another issue but let's file it so we don't forget!

freddyaboulton (approved):
Thanks @abidlabs ! Awesome PR

Issue #2598, opened 2022-11-03
`gr.Markdown()` adds significant extra whitespace


abidlabs:
### Describe the bug

When using `gr.Markdown()`, it seems to create significant whitespace above and below it. I'm pretty sure this is a regression. Example: 

![image](https://user-images.githubusercontent.com/1778297/199765877-eddfe1e4-8aa8-4759-9e6c-2289f0097a3a.png). 

h/t @osanseviero 

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

import gradio as gr

gr.Interface(lambda x:x, "textbox", "textbox", title="Hi").launch()

### Screenshot

<img width="620" alt="image" src="https://user-images.githubusercontent.com/1778297/199766119-90b08534-ee5a-4c56-8771-b83ba4888c13.png">

### Logs

```shell
N/A
```

### System Info

```shell
3.8
```

### Severity

annoying

abidlabs:
The reason we have this is because we introduced a minimum height to the Markdown and HTML components in this PR (#2474), which adds too much padding. I started removing it, but then I realized that the reason that the height was added was to ensure that the status tracker for these elements always showed up correctly. @pngwn do you have any suggestions on how to fix this?

abidlabs:
Fixed by #2623, thanks @pngwn!

PR #2599, opened 2022-11-03
Attempt to Fix Markdown/HTML spacing

abidlabs:
I'm attempting to fix this issue here: #2598 

The reason we have this is because we introduced a minimum height to the `Markdown` and `HTML` components, which adds too much padding. I've removed it in this PR, but then I realized that the reason that the height was added was to ensure that the status tracker for these elements always showed up correctly. @pngwn do you have any suggestions on how to fix this?

Here's a minimal example showing the problem:

```py
import gradio as gr
import time

def sl(x):
    time.sleep(1)
    return x

with gr.Blocks() as demo:
    t = gr.Textbox("hi")
    h = gr.Markdown("hi")
    gr.Textbox("hi")
    t.change(sl, t, h)

demo.launch()
```

### DIFF
--- a/ui/packages/html/src/HTML.svelte
+++ b/ui/packages/html/src/HTML.svelte
@@ -10,6 +10,6 @@
 	$: value, dispatch("change");
 </script>
 
-<div class="output-html min-h-[4rem]" id={elem_id} class:!hidden={!visible}>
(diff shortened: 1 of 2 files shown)


abidlabs:
Gonna close this PR since there's not much in the way of changes and instead continue the conversation in the issue.

Issue #2595, opened 2022-11-02
`gr.Sketchpad` does not take arguments


noobmldude:
### Describe the bug

When using the `gr.Sketchpad` component with any of the parameters in the docstring, it fails.
`gr.Sketchpad` only works without any parameter.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

To reproduce this, you can run this snippet:
```python
import gradio as gr

def show(img):
    return img

demo = gr.Interface(show, 
                    inputs=gr.Sketchpad(invert_colors=False), 
                    outputs='image')
demo.launch()
```

### Screenshot

On Investigation, it looks like the SketchPad component has a hardcoded set of keyword arguments passed to the super class. Not sure about the reason for it. 

```python
class Sketchpad(components.Image):
    """
    Sets: image_mode="L", source="canvas", shape=(28, 28), invert_colors=True
    """

    is_template = True

    def __init__(self, **kwargs):
        super().__init__(
            image_mode="L",
            source="canvas",
            shape=(28, 28),
            invert_colors=True,
            interactive=True,
            **kwargs
        )
```

This is probably the reason for the error
`TypeError: gradio.components.Image.__init__() got multiple values for keyword argument 'invert_colors'`

### Logs

```shell
In the above example, when passing the keyword arguments like `invert_colors` or `shape` to `gr.Sketchpad()` it throws the error:


      5 
      6 demo = gr.Interface(show, 
----> 7                     inputs=gr.Sketchpad(invert_colors=False),
      8                     outputs='image')
      9 demo.launch()

/usr/local/lib/python3.9/site-packages/gradio/templates.py in __init__(self, **kwargs)
     43 
     44     def __init__(self, **kwargs):
---> 45         super().__init__(
     46             image_mode="L",
     47             source="canvas",

TypeError: gradio.components.Image.__init__() got multiple values for keyword argument 'invert_colors'
```

### System Info

```shell
Gradio version: '3.8.2'
OS: MacOS
Browser: Chrome
```

### Severity

blocking all usage of gradio

PR #2600, opened 2022-11-03
Allow `gr.Templates` to take in arguments

abidlabs:
A fix to `gr.Templates` so that they can take in arguments. Fixes: #2595


gradio/templates.py
```diff
@@ -97,7 +106,9 @@ class Pil(components.Image):
     is_template = True
 
     def __init__(self, **kwargs):
```
freddyaboulton, line 108: Wouldn't it be better if we specified the kwargs the class supports as opposed to accepting arbitrary kwargs?

Right now you could do `gr.Pil(type="numpy")` which is confusing/defeats the purpose of the template. Also means devs won't get hints from their editors and makes it easy to make a typo and only be aware at runtime.

abidlabs: Ah yeah you're right. I think it's okay to override the template params but you're right it would be good to specify them explicitly

abidlabs:
Fixed now!

freddyaboulton (approved):

Issue #2278, opened 2022-09-15
Add "Run in Colab" Links to demos on the website Demos Page


freddyaboulton:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
It would help transition gradio users into gradio developers if they could easily run the demos hosted on our website on colab.

**Describe the solution you'd like**  
Add `Run in Colab` links to all the demos on the website.

aliabd:
Great suggestion @freddyaboulton, I'll also add these for the demos on Docs.

PR #2608, opened 2022-11-05
Add open in colab buttons to demos in docs and /demos

aliabd:
Adds buttons to open a demo in colab: 

https://user-images.githubusercontent.com/9021060/200097976-7089cf0e-6d27-4280-880d-a46726af4fee.mov

If a contributor wants to change a demo in the /demo dir, they just have to run /demo/generate_notebooks.py to regenerate all the .ipynb notebooks. I could create a github action for this but seems like overkill. 

Note that not all the demos will work on colab, the ones that reference another file in the dir won't work (for [example](https://colab.research.google.com/github/gradio-app/gradio/blob/aliabd/colab-buttons/demo/main_note/run.ipynb)). I could detect this and not show a colab button for those, but I feel like it would still be useful for a reader to be able to play with the code (so don't know if it's the right idea to skip them), and usually you can easily replace the file (I could link to the dir in github so they can move the files they need?) 

Closes: #2278 

# Checklist:


abidlabs:
Wow this is great @aliabd! LGTM with a few suggestions on the points you raised:

> If a contributor wants to change a demo in the /demo dir, they just have to run /demo/generate_notebooks.py to regenerate all the .ipynb notebooks. I could create a github action for this but seems like overkill.

I took a look at the `/demo` directory and we do edit it relatively frequently so I actually think it *would* be good to have a GitHub action to run to make sure no demos and notebooks go out of sync. 

> Note that not all the demos will work on colab, the ones that reference another file in the dir won't work 

I think that's fine. I think it's pretty clear to users when this happens and how to fix it. As a general note, perhaps we should switch to using URLs in our demos instead as much as possible to avoid

abidlabs (approved):

aliabid94:
This is great.

I think there should be a github action that opens a commit if a demo changes and the notebook hasn't.
As for demos with files in the folder, add a cell block before the demo that curls and downloads the files into the colab environment, so the paths resolve the same way.

aliabd:
Ok I updated this: 

* There's now a github check that will run on every PR that changes something in the  `/demo` path. If the demo run.py has changed but not the generated notebook, the check fails. (check works by running `python generate_notebooks.py` and asserting that there's nothing to commit). You can see an example of it failing in this fake commit [here](https://github.com/gradio-app/gradio/actions/runs/3505924074/jobs/5872586674), and an example of it passing once I regenerated the notebooks [here](https://github.com/gradio-app/gradio/actions/runs/3505926943/jobs/5872591250).
* When the notebooks are generated, we automatically add a cell to wget the necessary files from the demo folder (and put them in the correct structure) so every demo should run immediately.

.github/workflows/check-demo-notebooks.yml
```diff
@@ -0,0 +1,27 @@
+# This workflow will check if the run.py files in every demo match the run.ipynb notebooks.
+
+name: Check Demos Match Notebooks
+
+on:
+  pull_request:
+    types: [opened, synchronize, labeled, unlabeled]
```
abidlabs, line 7: I don't think we need labeled/unlabeled, but we might want reopened
```suggestion
    types: [opened, synchronize, reopened]
```

demo/generate_notebooks.py
```diff
@@ -0,0 +1,75 @@
+import nbformat as nbf
+import os 
+import json
+import random
+
+GRADIO_DEMO_DIR = os.getcwd()
+DEMOS_TO_SKIP = {"all_demos", "reset_components", "custom_path", "kitchen_sink_random"}
```
abidlabs, line 7: Why are we skipping "reset_components", "custom_path", "kitchen_sink_random"

abidlabs:
Awesome PR @aliabd! Left a couple of nits above but otherwise LGTM.

 One suggestion -- if the CI check fails, would it be possible to print the command that needs to be run to regenerate the notebooks?

abidlabs (approved):

demo/generate_notebooks.py
aliabd: `reset_components` and `kitchen_sink_random` import from other demos so can't be used on colab alone. We also skip them when uploading to spaces because they don't work there either. They're only used for the PR demos. (along with `all_demos`)

`custom_path` is a demo of  launching gradio using uvicorn from the terminal, so not really useful on colab either (also skipped on spaces)

.github/workflows/check-demo-notebooks.yml
aliabd: you're right we don't need labeled/unlabeled but do need reopened. committed

aliabd:
@abidlabs thanks for the suggestion. The github action will now comment on the PR if the check fails with the command that needs to be run. The last two commits show an example of this.

.github/workflows/check-demo-notebooks.yml
```diff
@@ -0,0 +1,38 @@
...
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+      - name: Generate Notebooks
+        run: |
+          pip install nbformat && cd demo && python generate_notebooks.py
+      - name: Print Git Status
+        run: echo $(git status) && echo $(git diff)
+      - name: Assert Notebooks Match 
+        id: assertNotebooksMatch
+        run: git status | grep "nothing to commit, working tree clean"
+      - name: Comment On PR
```
freddyaboulton, line 29: Does this run on forked repos?

demo/generate_notebooks.py
```diff
@@ -0,0 +1,75 @@
...
+    with open(output_notebook, 'w') as f:
+        nbf.write(nb, f)
+
+    with open(output_notebook, "r") as f:
+        content = f.read()
+    
+    content = json.loads(content)
+    for i, cell in enumerate(content["cells"]):
+        random.seed(i)
+        cell["id"] = random.getrandbits(128)
```
freddyaboulton, line 72: Just wondering why this is neccessary?

freddyaboulton (approved):
Super cool @aliabd !

.github/workflows/check-demo-notebooks.yml
aliabd: Good question, I just checked on this PR https://github.com/gradio-app/gradio/pull/2691 

The check fails as expected, but it [can't comment](https://github.com/gradio-app/gradio/actions/runs/3518591252/jobs/5897671236). Do you know how to fix that? Apparently it's an [open issue](https://github.com/actions/first-interaction/issues/10).

demo/generate_notebooks.py
aliabd: Good question... 

The way the github check works is by running `python generate_notebooks.py` and seeing if there's anything to commit. If there is, it must mean that run.py files and run.ipynb notebooks don't match. 

The problem is that jupyter notebooks create a random id for every cell that's different each time, and creates a huge mess on git diffs even when the notebook is functionally the same. So everytime `python generate_notebooks.py` is run it will show that every notebook has changed. There's no way to set the id using nbformat so instead I just manually set them in the file from python and define a seed to make sure they're always the same.

.github/workflows/check-demo-notebooks.yml
freddyaboulton: We need to follow this approach I think: https://securitylab.github.com/research/github-actions-preventing-pwn-requests/#:~:text=The%20main%20differences%20between%20the,but%20not%20from%20external%20forks. I think it may also be possible to comment on the pr if the action is defined in a separate repo? I think that's what doc-builder does.

@pngwn May now some best practices for commenting on PRs from forks.

demo/generate_notebooks.py
freddyaboulton: Very cool - thanks for explaining!

.github/workflows/check-demo-notebooks.yml
aliabd: Ah nice

aliabd:
@freddyaboulton I updated this to use `pull_request_target` and the action now works as expected for both branches and forks. Thanks for the suggestion! Tested here in a [branch](https://github.com/gradio-app/gradio/pull/2701) and a [fork](https://github.com/gradio-app/gradio/pull/2702).

aliabid94 (approved):
looks great!

Issue #1784, opened 2022-07-13
Refactor tests to use pytest fixtures and plain asserts


freddyaboulton:
We should consider rewriting our tests to the pytest style (functions prefixed with `test_` and plain asserts) as opposed to using unittest.Testcase classes. Switching over to just plain functions and assets makes it really easy to use pytest's built in fixtures and makes it easy to integrate with community-built fixtures. Pytest fixtures can do everything we can do with `setUp` and `tearDown`.

Context:

I think this is a great reason to consider switching over to pytest fully. Pytest comes with a built in fixture called `capsys` that does this for you. 

This is how we can rewrite one of the tests here in this style:

```python
def test_integration_wandb(capsys):

    wandb.log = mock.MagicMock()
    wandb.Html = mock.MagicMock()
    demo = gr.Blocks()
    with demo:
        gr.Textbox("Hi there!")
    demo.integrate(wandb=wandb)

    text = capsys.readouterr()
    assert text.out.strip() == "The WandB integration requires you to `launch(share=True)` first."
    demo.share_url = "tmp"
    demo.integrate(wandb=wandb)
    wandb.log.assert_called_once()
```

Running `pytest test/test_blocks.py::test_integration_wandb` will work. I'll create an issue so that we don't forget and can continue discussing there.

_Originally posted by @freddyaboulton in https://github.com/gradio-app/gradio/pull/1776#discussion_r920252997_

freddyaboulton:
@FarukOzderim Agreed that it's not high priority.

With this kind of issue, it may be best to use a gradual approach rather than all-or-nothing.

Rather than someone re-writting the entire test suite in a couple of days, it may be best to start writing new tests in the pytest style and encourage PR authors to do the same.

FarukOzderim:
> @FarukOzderim Agreed that it's not high priority.
> 
> With this kind of issue, it may be best to use a gradual approach rather than all-or-nothing.
> 
> Rather than someone re-writting the entire test suite in a couple of days, it may be best to start writing new tests in the pytest style and encourage PR authors to do the same.

Totally agree, I actually have started to do it without mentioning it 😿, lets ping this out in slack as well. We can unittests to pytest style whenever we refactor a test as well.

PR #2610, opened 2022-11-06
Refactoring to replace unittests with pytests

abidlabs:
I started going through our tests and converting them from unittest to pytest, when I noticed something wrong -- some of our tests were passing even though they should not have been. It seems that pytest was letting async tests automatically pass if they were written as part of a unittest class. 

For example, if you change the `TestBlocksMethods::test_dict_inputs_in_config` test in `test_blocks.py` to anything else, you'll find that it still passes:

```py
    async def test_dict_inputs_in_config(self):
        1/0
```

This is not the case if we change the tests to be pure pytests. So I went through and converted all of our tests to pytest. I noticed some other tests that should not have been passing, either due to the tests being incorrect or there being bugs in our codebase. Nothing too major, but here are the changes I made:

* Many tests in `test_components.py` passed in a *list* to `Interface.__call__` instead of the input parameters as separate arguments. This is not correct and the tests have been fixed. 
* In `test_components.py`, several tests used `.serialize()` in a way that is not supported or used in our library anymore. They've been rewritten or removed.
* Many of the components were not raising a `ValueError()` when passed an invalid argument for `type`. They now correctly do this. 

All of the above issues are now fixed in the library, and all of our tests have been converted to pytest format. 

Closes: #1784


freddyaboulton:
@abidlabs I noticed on Friday that pytest emits some warnings along the lines of “coroutine test_….. was never awaited”. Which means those tests are never really checked 😬 

Confusingly, these tests have “@pytest.mark.asyncio”. 

I wonder if this PR will get rid of those will get rid of those warnings.

abidlabs:
Exactly @freddyaboulton it seems that some async tests are not running at all! l elaborate in a follow up message

abidlabs:
I've updated the original message with more details, and this is now ready for review

test/test_components.py
```diff
@@ -1,14 +1,20 @@
+"""
+Tests for all of the componets defined on components.lpy. They are divided into two types of tests:
```
freddyaboulton, line 2: nit typo: `components.lpy`

freddyaboulton (approved):
Thanks for the fix @abidlabs ! I think the fact some components now raise ValueError about invalid `type` parameter values is worthy of a mention in the changelog though.

test/test_components.py
abidlabs: Ah thanks

abidlabs:
That's fair, thanks @freddyaboulton for the review! Will add a changelog and merge in to avoid conflicts

PR #2428, opened 2022-10-11
Update queue with using deque & Update requirements

GLGDLY:
# Description

The performance of queue instance might be improved with the use of double ended queue and `.popleft()`, rather than the list used currently.

Also, for the requirements part, websockets with version<=9.1 would raise `TypeError: WebSocketCommonProtocol.__init__() got an unexpected keyword argument 'logger'`, so it would be better to ensure its version to be >=10.0.

# Checklist:

- [X] I have performed a self-review of my own code
- [X] I have added a short summary of my change to the CHANGELOG.md
- [X] My code follows the style guidelines of this project
- [X] I have commented my code in hard-to-understand areas
- [X] I have made corresponding changes to the documentation
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] New and existing unit tests pass locally with my changes

Issue #293, opened 2021-10-09
Cannot queue with encryption or authentication enabled.


johnyquest7:
Unable to queue when authentication is enabled.

abidlabs:
How significant is this use case @johnyquest7? Queueing is designed for public demos with high traffic (e.g. on Space), while I envision authentication essentially for private demos with lower traffic.

johnyquest7:
Not a priority now. Thanks for looking into this.

xxWeiDG:
> Thanks for letting us know @johnyquest7!

Hi Does queue support auth now?

freddyaboulton:
Yes fixed in #2611

PR #2611, opened 2022-11-06
Allow auth with using queue

GLGDLY:
# Description

Allow auth with using queue, initiated by #2215 , which if queue is on by default, authentication then cannot be used by default. A partial fix to #293 as well.

Also fix wrong position in changlog from the last pr #2428 

# Checklist:


aliabid94:
Very cool, thanks @GLGDLY! Testing now

aliabid94:
Works great, thanks a lot @GLGDLY!

test/test_blocks.py
```diff
@@ -911,5 +911,76 @@ def test_queue_enabled_for_fn():
...
+                                "session_hash": "enwpitpex2q",
+                            }
+                        )
+                    )
+                    msg = json.loads(await ws.recv())
+                    assert msg["msg"] == "process_starts"
+                if msg["msg"] == "process_completed":
+                    assert msg["success"]
+                    assert msg["output"]["data"] == ["Hello 123!"]
+                    assert _loop.time() > _time
```
freddyaboulton, line 974: This is not passing on our CI. I don't think we need to check the time. I think it would be sufficient to test that the output is correct. Might make sense to rewrite this loop so that the "name" is different for each event in the queue.

test/test_blocks.py
```diff
@@ -911,5 +911,76 @@ def test_queue_enabled_for_fn():
     assert demo.queue_enabled_for_fn(1)
 
 
+@pytest.mark.asyncio
+async def test_queue_when_using_auth():
+    sleep_time = 5
+
+    async def say_hello(name):
+        await asyncio.sleep(sleep_time)
```
freddyaboulton, line 919: I don't think we need to sleep for so long. The test takes quite a while to run right now.

freddyaboulton (commented):
Thank you for the very nice PR @GLGDLY ! Left some comments and questions. Once those are addressed these should be good to merge 😄

GLGDLY:
Thank you @aliabid94 @freddyaboulton , I will take a look at the problems later today🙏🏻

gradio/routes.py
```diff
@@ -391,13 +403,13 @@ async def join_queue(websocket: WebSocket):
...
 
-        @app.get(
-            "/startup-events",
-            dependencies=[Depends(login_check)],
-        )
+        @app.get("/startup-events")
         async def startup_events():
-            app.blocks.startup_events()
-            return True
+            if not app.startup_events_triggered:
```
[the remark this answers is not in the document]
GLGDLY: the comments are removed [here](https://github.com/gradio-app/gradio/pull/2611/commits/1bb1579f15902dbb61c59b1542a68295c8aeb54a), it is changed to allow the initialization to run the queue background coros

test/test_blocks.py
GLGDLY: "name" is different for each event in the queue is added now, time is checked as I think we need to ensure the queue is actually working

test/test_blocks.py
GLGDLY: it is 1 sec and 3 times now

freddyaboulton:
Thank you @GLGDLY !

GLGDLY:
> Thank you @GLGDLY !

ok thanks!

PR #2618, opened 2022-11-08
Reload website on every commit

aliabd:
Small change to the reload website script that makes it track main again.


abidlabs (approved):

abidlabs:
Thanks for the fix @aliabd

PR #2639, opened 2022-11-10
Fix bug with gr.update and interactive=True

### LINKED ISSUE #2636: `gr.Update` doesn't work for the `interactive` property
### Describe the bug

I would like to programatically swap whether a field is interactive or not depending on other things that happen in my application, however the `interactive` property seem not to be working with the `gr.Update` function

### Reproduction

By taking the demo in the docs page and just adding an `interactive=False` to the first choice:
```py
import gradio as gr


def change_textbox(choice):
    if choice == "short":
        return gr.Textbox.update(lines=2, visible=True, interactive=False)
    elif choice == "long":
        return gr.Textbox.update(lines=8, visible=True)
    else:
        return gr.Textbox.update(visible=False)


with gr.Blocks() as demo:
    radio = gr.Radio(
        ["short", "long", "none"], label="What kind of essay would you like to write?"
    )
    text = gr.Textbox(lines=2, interactive=True)

    radio.change(fn=change_textbox, inputs=radio, outputs=text)

if __name__ == "__main__":
    demo.launch()
```

I get the following error
```
raceback (most recent call last):
  File "/home/poli/miniconda3/envs/multimodal/lib/python3.9/site-packages/gradio/routes.py", line 289, in run_predict
    output = await app.blocks.process_api(
  File "/home/poli/miniconda3/envs/multimodal/lib/python3.9/site-packages/gradio/blocks.py", line 983, in process_api
    data = self.postprocess_data(fn_index, result["prediction"], state)
  File "/home/poli/miniconda3/envs/multimodal/lib/python3.9/site-packages/gradio/blocks.py", line 924, in postprocess_data
    prediction_value = postprocess_update_dict(
  File "/home/poli/miniconda3/envs/multimodal/lib/python3.9/site-packages/gradio/blocks.py", line 424, in postprocess_update_dict
    prediction_value = block.get_specific_update(update_dict)
  File "/home/poli/miniconda3/envs/multimodal/lib/python3.9/site-packages/gradio/blocks.py", line 265, in get_specific_update
    generic_update = cls.update(**generic_update)
TypeError: update() got an unexpected keyword argument 'mode'
```

### Screenshot

_No response_

### Logs

```shell
-
```

### System Info

```shell
Gradio == 3.9.1
```

### Severity

freddyaboulton:
# Description

Fixes #2636 

The original bug is fixed by modifying `Block.get_specific_update` to not pass "mode" as a kwarg to `update` as that's not a user facing argument.

In addition, I made it so that the default value of `interactive` in `update` methods is True. This is because if an update method sets `interactive=False` the user has to explicitly set `interactive=True` again in a separate update call if they want to make the component interactive.

For example, in the original repro, if the default value of `interactive` in the textbox update method is None, after clicking on `short` the textbox will never be interactive again. I figured this might be too confusing for users. Wondering what others thoughts are. See the example below:

### Cant type in "long" after clicking on "short" if the default value of interactive is None
interactivity

### Can type in "long" a
interactivity_2
fter clicking on "short" if the default value of interactive is None

Repro code:
```python
```python

def change_textbox(choice):
    if choice == "short":
        return gr.Textbox.update(lines=2, visible=True, interactive=False)
    elif choice == "long":
        return gr.Textbox.update(lines=8, visible=True)
    else:
        return gr.Textbox.update(visible=False)


with gr.Blocks() as demo:
    radio = gr.Radio(
        ["short", "long", "none"], label="What kind of essay would you like to write?"
    )
    text = gr.Textbox(lines=2, interactive=True)
    gr.Textbox(value="Foo")

    radio.change(fn=change_textbox, inputs=radio, outputs=text)

demo.launch()
```

# Checklist:

# A note about the CHANGELOG

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -262,7 +262,11 @@ class Block:
     @classmethod
     def get_specific_update(cls, generic_update):
         del generic_update["__type__"]
+        mode = generic_update.get("mode", None)
+        generic_update.pop("mode", None)
         generic_update = cls.update(**generic_update)
+        if mode:
+            generic_update["mode"] = mode
         return generic_update
 
 
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -262,11 +262,7 @@ class Block:
     @classmethod
     def get_specific_update(cls, generic_update):
         del generic_update["__type__"]
-        mode = generic_update.get("mode", None)
-        generic_update.pop("mode", None)
         generic_update = cls.update(**generic_update)
-        if mode:
-            generic_update["mode"] = mode
         return generic_update
 
 
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -425,10 +425,11 @@ def postprocess_update_dict(block: Block, update_dict: Dict, postprocess: bool =
         update_dict: The original update dictionary
         postprocess: Whether to postprocess the "value" key of the update dictionary.
     """
-    prediction_value = block.get_specific_update(update_dict)
-    if prediction_value.get("value") is components._Keywords.NO_VALUE:
-        prediction_value.pop("value")
-    prediction_value = delete_none(prediction_value, skip_value=True)
+    if update_dict.get("__type__", "") == "generic_update":
+        update_dict = block.get_specific_update(update_dict)
+    if update_dict.get("value") is components._Keywords.NO_VALUE:
+        update_dict.pop("value")
+    prediction_value = delete_none(update_dict, skip_value=True)
     if "value" in prediction_value and postprocess:
         prediction_value["value"] = block.postprocess(prediction_value["value"])
     return prediction_value
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ No changes to highlight.
 ## Bug Fixes:
 * Updated the minimum FastApi used in tests to version 0.87 [@freddyaboulton](https://github.com/freddyaboulton) in [PR 2647](https://github.com/gradio-app/gradio/pull/2647)
 * Fixed bug where interfaces with examples could not be loaded with `gr.Interface.load` by [@freddyaboulton](https://github.com/freddyaboulton) [PR 2640](https://github.com/gradio-app/gradio/pull/2640)
+* Fixed bug where the `interactive` property of a component could not be updated by [@freddyaboulton](https://github.com/freddyaboulton) in [PR 2639](https://github.com/gradio-app/gradio/pull/2639)
 
 ## Documentation Changes:
 No changes to highlight.
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -330,7 +330,7 @@ class Textbox(
         label: Optional[str] = None,
         show_label: Optional[bool] = None,
         visible: Optional[bool] = None,
-        interactive: Optional[bool] = True,
+        interactive: Optional[bool] = None,
     ):
         updated_config = {
             "lines": lines,
@@ -507,7 +507,7 @@ class Number(
         value: Optional[float] = _Keywords.NO_VALUE,
         label: Optional[str] = None,
         show_label: Optional[bool] = None,
-        interactive: Optional[bool] = True,
+        interactive: Optional[bool] = None,
         visible: Optional[bool] = None,
     ):
         updated_config = {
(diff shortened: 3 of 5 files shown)


abidlabs:
Thanks for the fix @freddyaboulton! 

I'm trying to understand this part:

> In addition, I made it so that the default value of interactive in update methods is True. This is because if an update method sets interactive=False the user has to explicitly set interactive=True again in a separate update call if they want to make the component interactive.

> For example, in the original repro, if the default value of interactive in the textbox update method is None, after clicking on short the textbox will never be interactive again. I figured this might be too confusing for users. Wondering what others thoughts are. See the example below:


Why are passing in a non-None value of `interactive` by default? Can't we leave the interactivity of the component to be whatever it was before the gr.update? That's how we do it for every other platform? If I understand what you're saying, it would be very confusing to a user who is changing one property to notice that the interactivity of the component has also changed

freddyaboulton:
> Why are passing in a non-None value of interactive by default? Can't we leave the interactivity of the component to be whatever it was before the gr.update? 

I thought it was maybe a bit confusing that in the first gif, the "long" version of the textbox is not interactive after you click on the "short" button.

 If we keep the default value of `interactive` to None, then the demo author will have to set the "long" textbox to be interactive manually:

```python
import gradio as gr


def change_textbox(choice):
    if choice == "short":
        return gr.Textbox.update(lines=2, visible=True, interactive=False)
    elif choice == "long":
        return gr.Textbox.update(lines=8, visible=True, interactive=True)
    else:
        return gr.Textbox.update(visible=False)


with gr.Blocks() as demo:
    radio = gr.Radio(
        ["short", "long", "none"], label="What kind of essay would you like to write?"
    )
    text = gr.Textbox(lines=2, interactive=True)
    gr.Textbox(value="Foo")

    radio.change(fn=change_textbox, inputs=radio, outputs=text)

demo.launch()
```

Which I think is ok but I'm worried it will confuse users who are used to thinking of output components as being interactive by default.

gradio/blocks.py
```diff
@@ -262,7 +262,11 @@ def get_config(self):
     @classmethod
     def get_specific_update(cls, generic_update):
         del generic_update["__type__"]
+        mode = generic_update.get("mode", None)
+        generic_update.pop("mode", None)
```
abidlabs, line 266:
```suggestion
        mode = generic_update.pop("mode", None)
```

abidlabs:
Thanks for the fix @freddyaboulton! I tested and it works. However, I think there is another way to do it which avoids an unnecessary method call. 

Right now, we have the following chain of methods that are being called:
* User calls in `gr.Textbox.update(...)`, which returns a dict with key `__type__` set to `update` and replaces the `interactive` key with `mode`
* In `postprocess_update_dict()` (in `blocks.py`), we call `get_specific_update()` which is meant to convert a "generic update" (i.e. `gr.update()`) to a component-specific `update()` dict
* However, the component-specific `update()` cannot handle the `mode` key anymore (since it is expecting the `interactive` parameter), so we see an error

Although this fixes the problem by extracting the `mode` key before calling the component-specific `update()`, the component-specific `update()` is still being called twice.

It would be sufficient to add an if condition to `postprocess_update_dict()` as follows to only call the component-specific `update()` if indeed we have a generic update, like this:

```py
def postprocess_update_dict(block: Block, update_dict: Dict, postprocess: bool = True):
    """
    Converts a dictionary of updates into a format that can be sent to the frontend.
    E.g. {"__type__": "generic_update", "value": "2", "interactive": False}
    Into -> {"__type__": "update", "value": 2.0, "mode": "static"}

    Parameters:
        block: The Block that is being updated with this update dictionary.
        update_dict: The original update dictionary
        postprocess: Whether to postprocess the "value" key of the update dictionary.
    """
    if update_dict.get("__type__", "") == "generic_update":
        update_dict = block.get_specific_update(update_dict)
    if update_dict.get("value") is components._Keywords.NO_VALUE:
        update_dict.pop("value")
    update_dict = delete_none(update_dict, skip_value=True)
    if "value" in update_dict and postprocess:
        update_dict["value"] = block.postprocess(update_dict["value"])
    return update_dict
```

freddyaboulton:
@abidlabs Thank you for the very thoughtful review! I followed your suggestion. Should be good for another review now. 

Was there a reason we were calling the component specific update twice? Was it a way to create a copy of the dict before deleting keys? I wonder if I should also copy the dict in `postprocess_update_dict` to match the previous behavior as closely as possible.

abidlabs:
> Was there a reason we were calling the component specific update twice? 

No I'm pretty sure I know what happened, basically it was a mistake I made when I was refactoring the code

freddyaboulton:
@abidlabs Yes, you are right! Reverted the changes to `get_specific_update` and updated the unit tests!

abidlabs:
Thank you @freddyaboulton for fixing this, looks great!

abidlabs (approved):

PR #2641, opened 2022-11-12
Adding ability to access in Fastapi request object into your function

abidlabs:
Adds ability to access the underlying FastAPI request from within your Python function. Simple test function:

```py
import gradio as gr

def test(name, request: gr.Request):
    print(request.headers)
    return name

io = gr.Interface(test, "textbox", "textbox").launch()
```

Fixes: #2379


freddyaboulton:
This is really cool @abidlabs ! 

One thing I noticed is that the Request object is different depending on whether or not the function is being run on the queue. That is because when the queue is active, requests are made from the server and not the client. If the purpose of this feature is to allow users to extract info from requests from the client, should we update the implementation to work better with the queue?

You can see what I'm talking about here: 

https://huggingface.co/spaces/freddyaboulton/fastapi-request

abidlabs:
Really good catch @freddyaboulton! Yes let me fix that

abidlabs:
Update: @freddyaboulton suggested a wrapper `gr.Request` over both websockets and regular requests, and I like it!

gradio/queue.py
```diff
@@ -157,18 +160,6 @@ async def broadcast_live_estimations(self) -> None:
         if self.live_updates:
             await self.broadcast_estimations()
 
-    async def gather_data_for_first_ranks(self) -> None:
```
abidlabs, line 160: Unused method

gradio/utils.py
```diff
@@ -395,19 +396,19 @@ def async_iteration(iterator):
         raise StopAsyncIteration()
 
 
-class Request:
+class AsyncRequest:
```
abidlabs, line 399: Renamed this class to avoid confusion with `gr.Request`

abidlabs:
Thanks for the suggestion @freddyaboulton. I've implemented it so that queuing passes the parameters from the original request, instead of the request from the server. This utilizes our own `gr.Request` class, which is also added to the docs.

freddyaboulton:
@abidlabs I lost this notification in my inbox! Will take a look today

gradio/routes.py
```diff
@@ -462,6 +478,59 @@ def get_server_url_from_ws_url(ws_url: str):
...
+
+    def dict_to_obj(self, d):
+        if isinstance(d, dict):
+            return json.loads(json.dumps(d), object_hook=Obj)
+        else:
+            return d
+
+    def __getattr__(self, name):
+        if self.request:
+            return self.dict_to_obj(getattr(self.request, name))
```
freddyaboulton, line 529: For consistency the request and websocket cases should both raise an AttributeError if the attribute doesn't exist or return None. Right now only the request case raises an AttributeError

gradio/blocks.py
```diff
@@ -456,6 +458,23 @@ def convert_component_dict_to_list(outputs_ids: List[int], predictions: Dict) ->
...
+    """
+    Adds the FastAPI Request object to the inputs of a function if the type of the parameter is FastAPI.Request.
+    """
+    param_names = inspect.getfullargspec(fn)[0]
+    try:
+        parameter_types = typing.get_type_hints(fn)
+        for idx, param_name in enumerate(param_names):
+            if parameter_types.get(param_name, "") == routes.Request:
+                inputs.insert(idx, request)
+    except TypeError:  # A TypeError is raised if the function is a partial or other rare cases.
```
freddyaboulton, line 473: Should we just raise an error here? If we can't add a request, user will get a "TypeError: fn missing 1 required positional argument: 'request'" or something like that down the line. I wonder if that error would be more actionable if we controlled the exception message, e.g. `Cannot add request to function signature. This typically happens if the function is a partial. Please refactor your function` ?

gradio/examples.py
```diff
@@ -290,7 +290,7 @@ async def cache(self) -> None:
                 if self.batch:
                     processed_input = [[value] for value in processed_input]
                 prediction = await Context.root_block.process_api(
-                    fn_index, processed_input
+                    fn_index=fn_index, inputs=processed_input, request=None
```
freddyaboulton, line 293: Might be worth mentioning that example caching may raise an exception if the function takes in a request.

guides/01_getting_started/03_sharing_your_app.md
```diff
@@ -106,6 +106,24 @@ def same_auth(username, password):
 demo.launch(auth=same_auth)
 ```
 
+## Accessing the Network Request Directly
+
+When a user makes a prediction to your app, you may need the underlying network request, in order to get the request headers (e.g. for advanced authentication), log the client's IP address, or for other reasons. Gradio supports this in a similar manner to FastAPI: simply add a function parameter whose type hint is `gr.Request` and Gradio will pass in the network request as that parameter. Here is an example:
+
+```python
+import gradio as gr
+import fastapi
```
freddyaboulton, line 115: fastapi not used here

gradio/blocks.py
```diff
@@ -456,6 +458,23 @@ def convert_component_dict_to_list(outputs_ids: List[int], predictions: Dict) ->
     return predictions
 
 
+def add_request_to_inputs(
```
freddyaboulton, line 461: I would unit test this

freddyaboulton (approved):
Awesome PR @abidlabs !  Left mostly nits. Tested and works well!

gradio/routes.py
abidlabs: Nice agreed

guides/01_getting_started/03_sharing_your_app.md
abidlabs: Thanks will remove

gradio/blocks.py
abidlabs: 👍

gradio/examples.py
abidlabs: Hmm let me test this out

abidlabs:
Thanks for the great review @freddyaboulton!

gradio/blocks.py
abidlabs: If we were to raise an error here, then it would break support for partial functions altogether since `add_request_to_inputs()` is called in all cases, regardless of whether the user actually supplies a parameter with type hint `gr.Request`. The reason I put the `TypeError` was because `typing.get_type_hints()` raises a `TypeError` when it inspects a partial function, so this way, we simply don't add the request object when working with partials. 

I think that partials lose all typing hint information so it might not be possible to use this approach with partials. I don't think it's that big of a deal, but open to suggestions.

gradio/blocks.py
abidlabs: Added

gradio/examples.py
abidlabs: Yeah ok so the value of `request` will be `None`. I'll add to the Guide that you should handle the case when `request` is None.

abidlabs:
All righty, this PR should be ready to go. I'll merge later tonight

abidlabs:
Thanks again for the detailed review @freddyaboulton. Merging!

abidlabs:
Unfortunately, I don't think any 3.x works without internet connection

PR #2652, opened 2022-11-15
UI update: API documentation page

gary149:
New footer (we can grayscale icons color if too visible, and keep the colors on hover):

image

New doc page:

image

Dark mode:

image


ui/packages/app/index.html
```diff
@@ -44,7 +44,7 @@
 			crossorigin="anonymous"
 		/>
 		<link
-			href="https://fonts.googleapis.com/css?family=Source Sans Pro"
+			href="https://fonts.googleapis.com/css2?family=Source+Sans+Pro:wght@400;600&display=swap"
```
gary149, line 47: @pngwn font-semibold was never imported? on purpose?

ui/packages/app/src/Blocks.svelte
```diff
@@ -457,31 +474,52 @@
...
-					api_docs_visible = !api_docs_visible;
+					set_api_docs_visible(!api_docs_visible);
 				}}
+				class="flex items-center hover:text-gray-500"
 			>
-				{#if api_docs_visible}hide{:else}view{/if} api
-			</div>
-			&nbsp; &bull; &nbsp;
+				View <img src={api_logo} alt="" class="w-2.5 md:w-3 mx-1" />API
+				documentation
```
gary149, line 487: Probably going to have to rework a bit on this (not sure it's very clear)
image

ui/packages/app/src/ApiDocs.svelte
```diff
@@ -12,130 +32,282 @@
...
+								> component
+							</span>
+							<br />
+						{/each}
+						&nbsp;&nbsp;]<br />
+						&#125;
+					</div>
+					<button
+						on:click={run.bind(null, dependency_index)}
+						class="gr-button gr-button-lg gr-button-primary w-full mt-4"
```
gary149, line 161: We could add a clear/reset button.
image

ui/packages/app/src/ApiDocs.svelte
```diff
@@ -12,130 +32,282 @@
...
-			There are no named API Routes for
-			<span class="italic text-amber-500">
+					<h4 class="font-bold mt-8 mb-3 flex items-center">
+						<svg width="1em" height="1em" viewBox="0 0 24 24" class="mr-1.5"
+							><path
+								fill="currentColor"
+								d="m8 18l-6-6l6-6l1.425 1.425l-4.6 4.6L9.4 16.6Zm8 0l-1.425-1.425l4.6-4.6L14.6 7.4L16 6l6 6Z"
+							/></svg
+						>
+						Code examples
```
gary149, line 225:
```suggestion
						Code snippets
```

abidlabs:
Absolutely love the new look @gary149!

ui/packages/app/src/Blocks.svelte
abidlabs: Agree, I think this should be tweaked so that the icons are not in the middle of the text. Perhaps "Use as API" instead of "View API documentation"?

julien-c: or maybe "Use this app API" to focus that it's *this* app's API

julien-c: (not sure)

ui/packages/app/index.html
gary149: Most text is rendered with Source Sans Pro, IBM Plex Mono is only used for the mono font. Also we use [semibold weight](https://github.com/search?q=repo%3Agradio-app%2Fgradio%20font-semibold&type=code) in the codebase.

ui/packages/app/src/ApiDocs.svelte
aliabid94: haha I feel like at this point we're just recreating gradio in the api page :P

ui/packages/app/index.html
aliabid94: yeah looks like our semibolds weren't rendering, good catch

ui/packages/app/src/Blocks.svelte
aliabid94: went with "Use via API", open to suggestions

aliabid94:
Should work with every component now, ready for review

abidlabs:
Love the new design of the footer and API page! @aliabid94 I just built the frontend to test locally and I'm running into this weird behavior where the modal pops up when the Gradio app is running inside of an iframe (e.g. on Spaces or in a jupyter notebook), but not if it is opened in a new browser tab. See here:

7df17382-9158-4d6e-95fd-5e459d70bef4


Are you able to reproduce?

abidlabs:
Thanks for the fix @aliabid94, this is working now for me. Generally works great across many different demos.

I noticed a few small UI improvements/nits:

* When a demo has multiple API endpoints, it's easy to miss that fact because the documentation for each endpoint takes so much room. I would add at the top after the URL the # of API endpoints

image

* I don't think it's super clear what the "run" button does, for someone who is seeing this page for the first time. I think the old name of "Try it now" or something is clearer

image

abidlabs (approved):

PR #2659, opened 2022-11-16
Quick fix to `gr.utils.validate_url`

abidlabs:
**Context**: we use `gr.utils.validate_url` to determine if a string is a URL or a filepath. The logic is quite simple -- we make a GET request to the URL and see if we get a result. 

However, some common media hosting sites (e.g. WikiCommons) [require a User Agent to be set](https://meta.wikimedia.org/wiki/User-Agent_policy#Python) when making a request -- otherwise, the request is immediately rejected. This fact causes `gr.utils.validate_url` to fail for these URLs. In addition, since we were checking for a strict 200 response, URLs that redirected to other URLs would fail.

This PR fixes those issues and add tests. 

Code to reproduce:

```py
import gradio as gr

with gr.Blocks() as demo:
    gr.Audio("https://upload.wikimedia.org/wikipedia/commons/e/ef/Yellowstone_sound_library_-_Grizzly_Bears_Roar_-_001.mp3")
    
demo.launch()
```

### DIFF
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -488,16 +488,20 @@ class TestSanitizeForCSV:
         ) == [["'=abc", "def", "'gh,+ij"], ["abc", "'=def", "'+ghij"]]
         assert sanitize_list_for_csv([1, ["ab", "=de"]]) == [1, ["ab", "'=de"]]
 
+
 class TestValidateURL:
     def test_valid_urls(self):
         assert validate_url("https://www.gradio.app")
         assert validate_url("www.gradio.dev")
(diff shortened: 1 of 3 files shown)


test/test_utils.py
```diff
@@ -488,6 +489,20 @@ def test_list(self):
         assert sanitize_list_for_csv([1, ["ab", "=de"]]) == [1, ["ab", "'=de"]]
 
 
+class TestValidateURL:
```
freddyaboulton, line 492: Should probably mark as flaky if for whatever reason these urls are unavailable

freddyaboulton (approved):
Thanks @abidlabs !

test/test_utils.py
abidlabs: Good catch, thanks @freddyaboulton!

Issue #2665, opened 2022-11-17
 **kwargs are not being passed down for gr.Interface.load


AK391:
### Describe the bug

It looks like the **kwargs are not being passed down as seen in this demo https://huggingface.co/spaces/akhaliq/Nitro-Diffusion/blob/main/app.py

description is passed but not showing up 

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

https://huggingface.co/spaces/akhaliq/Nitro-Diffusion/blob/main/app.py

### Logs

```shell
spaces
```

### System Info

```shell
spaces
```

### Severity

annoying

PR #2669, opened 2022-11-18
pass in kwargs to `Blocks.load()`

abidlabs:
It seems like we accidentally stopped passing in `**kwarg` arguments from `gr.Interface.load()` to `gr.Blocks.load()`. This brings that back and adds a test to prevent regressions in the future.

Fixes: #2665

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 No changes to highlight.
 
 ## Bug Fixes:
-No changes to highlight.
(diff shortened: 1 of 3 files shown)


freddyaboulton (approved):
On mobile right now so i cant verify for myself but this looks good to me! Thanks @abidlabs

abidlabs:
Thanks for the quick review @freddyaboulton!

PR #2674, opened 2022-11-18
Fix website launch script

aliabd:
Adds the url arg to the website launch script. Note that this arg is only needed to display to the wheel url in the instructions on building from main in https://gradio.app/docs/main Because running the script locally won't actually build and upload a gradio wheel to aws, that url won't always work, but the website will otherwise build fine.


abidlabs (approved):
Thanks @aliabd this fixed it for me!

Issue #2667, opened 2022-11-17
Truncate dataframe cell contents


freddyaboulton:
- [ ] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
The dataframe component will show all the cell contents but this can make you dataframe really wide if it has really long text values (common in NLP). 

![dataframe_wide](https://user-images.githubusercontent.com/41651716/202560057-721a353f-538f-4a14-8985-d03fa6054e3b.gif)

**Describe the solution you'd like**  
Dataframe cell contents are truncated after 20 or so characters. You can double click on the cell to expand it and double click to collapse it again.

**Additional context**

Issue #2666, opened 2022-11-17
Scroll bars on dataframes


freddyaboulton:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
If your gradio demo displays a dataframe with many rows or columns, all of the dataframe will be displayed at once and users have to scroll all the way down the page to see their data.

![dataframe_long](https://user-images.githubusercontent.com/41651716/202559043-f02fbd8f-a7a2-48a5-a5de-0cfca4042f1a.gif)

**Describe the solution you'd like**  
Keep the dataframe a fixed size but add a scroll bar to the right hand side and bottom of the dataframe. That way the scroll bar doesn't increase the size of the entire page.

**Additional context**

abidlabs:
I suppose pagination for dataframes (as suggested here: #1245) would be the alternative

freddyaboulton:
I think we can support both? To me, pagination means you have to click on a separate button to take you to the next "page" of the dataframe, which may not be desired for some cases and it's not mutually exclusive with scroll bars?

liuheng2cqupt:
Does the latest version (3.28) already support pagination or scroll?  why mine doesn't work?

abidlabs:
Hi @liuheng2cqupt no we haven’t implemented this yet, it’s still an open issue

mazzma12:
An horizontal scrollbar would make sense as well, since it's hard to visualize dataframe > 10 columns

pseudotensor:
Yes, it's an issue, the page gets very long, so I only do head.  But columns have no scroll, so a hacky way is to drag the mouse over the page to get it to move right and left.

![image](https://github.com/gradio-app/gradio/assets/2249614/1204819c-7c6f-4c75-91f1-43684a388c5f)

gvlachos-au:
FYI as a temporary workaround you can try the following CSS solution (works in Firefox AFAIK) - also make sure to update `3-38-0` to the version of gradio you are using:

.gradio-container-3-38-0 .table-wrap {
  overflow-x: auto !important;
  overflow-y: auto !important;
  scrollbar-width: thin !important;
}

pseudotensor:
Hmm, I tried that but actually opposite happened, I lost the vertical scroll bar and still don't have horizontal scroll bar.

![image](https://github.com/gradio-app/gradio/assets/2249614/383d114c-7b72-41af-b020-7716bde75f4f)

PR #2677, opened 2022-11-18
[WIP] Add DatasetViewer class

freddyaboulton:
# Description

A very rough WIP of what it would look like to add a dataset viewer "widget" to gradio. 

### To demo this

```python
import gradio as gr
import datasets

apes = gr.DatasetViewer("huggingnft/boredapeyachtclub")
imdb = gr.DatasetViewer("imdb")
speech = gr.DatasetViewer(datasets.load_dataset("asapp/slue", "voxpopuli", split="train"))
with gr.Blocks() as demo:
    with gr.Tabs():
        with gr.Tab("Image Dataset"):
            apes.render()
        with gr.Tab("NLP Dataset"):
            imdb.render()
        with gr.Tab("Audio Dataset"):
            speech.render()

demo.launch()
```

You can also just launch an individual DatasetViewer

```python
imdb = gr.DatasetViewer("imdb").launch()
```

You will have to wait for the datasets library to cache all the data. Can take a while!

Some improvements that come to mind:
- Display audio files in Dataset component
- Support streaming datasets to prevent downloading the entire dataset
- Pass in a filepath

We also need to better display large amounts of text in a text box and add scroll bars to the dataset. Similar to #2666 and #2667 

I initially wanted to do the sql query component of it as well but I was having trouble getting [datasets_sql](https://github.com/mariosasko/datasets_sql) to work. Will need to catch up with @mariosasko to figure out why.

If you run into`MemoryError: Cannot allocate write+execute memory for ffi.callback()` you may need to upgrade to 3.10 [[Apple M1] MemoryError: Cannot allocate write+execute memory for ffi.callback()](https://github.com/huggingface/datasets/issues/4965#top)

FYI @abidlabs 

# Checklist:

# A note about the CHANGELOG


julien-c:
> You will have to wait for the datasets library to cache all the data. Can take a while!

the idea is to use https://github.com/huggingface/datasets-server instead of datasets in the future, right?

freddyaboulton:
> the idea is to use https://github.com/huggingface/datasets-server instead of datasets in the future, right?

Yes I think the datasets server would simplify things! Right now, the main limitation is that it only supports 100 rows and it can truncate the response. Very excited for https://github.com/huggingface/datasets-server/pull/631, so that we can do sql + stream the results over http.

abidlabs:
Very exciting @freddyaboulton! 

Is the idea that the `DatasetViewer` would also include some ability to explore the dataset as well or just display the dataset? If it's just to display the dataset, how would it differ from `gr.Dataset()`? 

>* Display audio files in Dataset component
>* Support streaming datasets to prevent downloading the entire dataset
>* Pass in a filepath

All of these are great ideas -- if we are using `datasets` library for now, then I [streaming is already supported](https://huggingface.co/docs/datasets/stream) out-of-the-box for all datasets, correct?

abidlabs:
cc @yvrjsharma, would love your thoughts, particularly how we could make this abstraction useful for the PlaygroundAI dataset visualization you shared last week

freddyaboulton:
@abidlabs I started out with adding a text box to DatasetViewer that would allow users to write sql queries but I couldn’t get the datasets_sql library to work. I’ll work with Mario to figure it out. 

I’m not sure if streaming is supported for all datasets but we’ll need https://github.com/huggingface/datasets-server/pull/631 to support sql for all datasets over http. Needless to say, I’ll work with the datasets team to make sure we can create a roadmap towards arbitrary data analysis with sql over http.

julien-c:
datasets streaming is not supported for all datasets, which is why we wish for this integration to use `datasets-server`

Strategically, it will be a great showcase of what that new API enables. And it's going to be ready (in an initial version) in the next couple of weeks, so I'd rather not duplicate efforts and for this to use `datasets-server` from the get go

WDYT?

cc @severo @lhoestq

julien-c:
(Something like `datasets_sql` will also be more future-proof robust with the new API too)

severo:
+1. The idea is that every dataset will be available in parquet format over HTTP, thus the viewer should not depend on the "datasets" library.

Besides this, the datasets server also has an infrastructure to preprocess things (stats on the columns, NLP metrics, etc.) so instead of computing them in Gradio, it would be better to create a new "worker" in the datasets server and access the result through the HTTP API too. This way, it's computed once and for all by the datasets server, and everybody can access it.

freddyaboulton:
@julien-c @severo I agree the "data on the hub" case is best handled by the datasets server. Definitely plan on using it when it's ready! But for this feature to be most useful, we should let users visualize/explore their data wherever it lives - database, local file, s3 etc - provided it's compatible with the datasets library. For those cases we'd still need the datasets library right? I don't think we'd be duplicating efforts.

Also, what do you mean by `(Something like datasets_sql will also be more future-proof robust with the new API too)`?

julien-c:
> For those cases we'd still need the datasets library right? I don't think we'd be duplicating efforts.

For local only Gradio, yes. But for most use cases we would push the datasets-server, that's where many upcoming features will live (and the code is open source too)

abidlabs:
Some more thoughts around the `DatasetViewer` class

A) Demoing a model is a very well-defined use-case, and as such we've been able to build the `gr.Interface()` abstraction in a very general manner: if you have pretty much any machine learning model you'd like to demo, you can use `gr.Interface` to build the input/output UIs and then add other flourishes, like a title, description, examples, etc. However, viewing a dataset is less defined: do we have a good sense of what people want to do when they explore a dataset? Currently, the `DatasetViewer` dataset is **only designed to let users filter** datasets. I think it's fine to start here, but I imagine there are other things we should support (e.g. find duplicate entries, see the embeddings of the data etc.) How support these use cases down the line with the `DatasetViewer` class? (@freddyaboulton perhaps your work with the data measurement tool has given you some ideas)

B) A UI works for pretty much anyone. But querying via SQL is not always the most user-friendly approach to explore a dataset. Should we allow users to query a dataset using other languages (e.g. pandas syntax, or thinking even more out-of-the-box: natural language). 

These are things we can add later, but I want to bring them up now so that we design the `DatasetViewer` class in the most general way possible. cc @aliabid94 @pngwn @osanseviero @merveenoyan for your thoughts!

freddyaboulton:
Going to close for now and circle back later!

PR #2679, opened 2022-11-18
Correctly building frontend in gradio wheel (colab bug)

aliabd:
The Dockerfile wasn't correctly moving  the frontend files in the uploaded gradio wheel. 

This led to an issue on colab where the interface hangs forever because it's directing to `/api/predict/` instead of `/run/predict/`. Which is the old way of doing things. You can see this issue by installing the wheel for the latest commit to main: https://colab.research.google.com/drive/1QD_YrLrOgqgNtDLAadAomKZm4mn_GsAs?usp=sharing 

And you can see this fixes it by installing the wheel for the latest commit to this branch: https://colab.research.google.com/drive/1w97YXgNZ5riGlQy0poKTPrFg2nCoERWJ?usp=sharing 

I also tested it locally and on spaces.


abidlabs (approved):
Great catch @aliabd! Tested and the second colab works great

PR #2684, opened 2022-11-21
Make login page UI consistent

aliabid94:
Previously our login page had its own CSS, had no dark mode, and had an ugly json message on the wrong credentials.
Made the page more aesthetically consistent, added dark mode support, and a nicer error message.

Before:

After


freddyaboulton (approved):
New look of the login page looks great to me @aliabid94 !

abidlabs:
LGTM @aliabid94. cc @gary149 

One nit: replace "login" with "Login" as we don't use lowercase font anywhere in the UI

abidlabs (approved):

Issue #2681, opened 2022-11-20
OSError: [Errno 36] File name too long


SkyTNT:
### Describe the bug

`Audio.preprocess` will throw error If the file name of input audio is too long.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Name an audio file `水本ゆかりです｡私､こんなかわいい服は初めて着ました…｡なんだかうれしくって､楽しいです｡歌いたくなる気分って､初めてです｡これがｱｲﾄﾞﾙってことなのかもしれませんね!7046dya9`  and put it into 
any Audio component in [https://huggingface.co/spaces/skytnt/moe-tts](https://huggingface.co/spaces/skytnt/moe-tts)

### Logs

```shell
Traceback (most recent call last):
  File "/home/user/.local/lib/python3.8/site-packages/gradio/routes.py", line 284, in run_predict
    output = await app.blocks.process_api(
  File "/home/user/.local/lib/python3.8/site-packages/gradio/blocks.py", line 980, in process_api
    inputs = self.preprocess_data(fn_index, inputs, state)
  File "/home/user/.local/lib/python3.8/site-packages/gradio/blocks.py", line 891, in preprocess_data
    processed_input.append(block.preprocess(inputs[i]))
  File "/home/user/.local/lib/python3.8/site-packages/gradio/components.py", line 1842, in preprocess
    file_obj = processing_utils.decode_base64_to_file(
  File "/home/user/.local/lib/python3.8/site-packages/gradio/processing_utils.py", line 277, in decode_base64_to_file
    file_obj = tempfile.NamedTemporaryFile(
  File "/usr/local/lib/python3.8/tempfile.py", line 540, in NamedTemporaryFile
    (fd, name) = _mkstemp_inner(dir, prefix, suffix, flags, output_type)
  File "/usr/local/lib/python3.8/tempfile.py", line 250, in _mkstemp_inner
    fd = _os.open(file, flags, 0o600)
OSError: [Errno 36] File name too long: '/tmp/水本ゆかりです｡私､こんなかわいい服は初めて着ました…｡なんだかうれしくって､楽しいです｡歌いたくなる気分って､初めてです｡これがｱｲﾄﾞﾙってことなのかもしれませんね!7046dya9.wav'
```

### System Info

```shell
gradio3.9, space [https://huggingface.co/spaces/skytnt/moe-tts](https://huggingface.co/spaces/skytnt/moe-tts)
```

### Severity

annoying

[...]

PR #2685, opened 2022-11-21
Fix filename too long in `processing_utils.decode_base64_to_file` and `processing_utils.create_tmp_copy_of_file`

SkyTNT:
Shorten the filename if it is too long in `processing_utils.decode_base64_to_file` and `processing_utils.create_tmp_copy_of_file`

Closes: #2681


abidlabs:
Thanks for creating this PR @SkyTNT!

I looked into the code a little bit as I was reviewing this PR and it seems we already trim the prefix in `utils.strip_invalid_filename_characters()` to 200 characters. We call this method in `processing_utils.create_tmp_copy_of_file()` but not in `processing_utils.decode_base64_to_file()`. I think we add a call to this method in `processing_utils.decode_base64_to_file()`, that should be sufficient, right?

SkyTNT:
> Thanks for creating this PR @SkyTNT!
> 
> I looked into the code a little bit as I was reviewing this PR and it seems we already trim the prefix in `utils.strip_invalid_filename_characters()` to 200 characters.
> 
> We call this method in `processing_utils.create_tmp_copy_of_file()` but not in `processing_utils.decode_base64_to_file()`. I think we add a call to this method in `processing_utils.decode_base64_to_file()`, that should be sufficient, right?

I don't think so. `strip_invalid_filename_characters` just simply trim the str but not the bytes. The bytes length of prefix may still be too long.

abidlabs:
I see thanks for clarifying. Let's modify the `utils.strip_invalid_filename_characters()` function to trim the filename based on bytes instead of characters, then. That's the right place for this logic to live.

Would you be able to do that @SkyTNT?

SkyTNT:
Done!

abidlabs:
Thanks @SkyTNT for the fix. I think we'll want to add a test to confirm and then add an entry in the changelog to give you credit and then we should be good to merge in!

abidlabs:
Sure happy to next week!

abidlabs:
Thank you so much for the nice PR @SkyTNT. Tested & merged in

PR #2692, opened 2022-11-21
Embeds: support `<gradio-app host="julien-c-coqui.hf.space"></gradio-app>`

julien-c:
to support the 'embed this Space' modal that will be shipped soon by [internal PR](https://github.com/huggingface/moon-landing/pull/4546/)


abidlabs:
Thanks @julien-c. I think some of these iframes can be replaced with web components altogether -- let me update them and then we can merge this in.

abidlabs (approved):
Thank you for these updates @julien-c!

Issue #2693, opened 2022-11-21
More actionable error message if loading a model into an interface fails


freddyaboulton:
- [ ] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
When you try to load a model into an interface, we check if the model exists by sending a get request to the inference api for that model. 

```python
response = requests.request("GET", api_url, headers=headers)
assert response.status_code == 200, "Invalid model name or src"
```

However, this request can also fail if the model exists but you have not accepted the conditions to use the model from your HF account.

**Describe the solution you'd like**  
We should give more actionable error messages that explain why loading a model failed and what the can do about it. 

**Additional context**  
This came up in this forum discussion: https://discuss.huggingface.co/t/gr-interface-load-hugging-model-repo-assertionerror-invalid-model-name-or-src/26471

Issue #2671, opened 2022-11-18
Overriding interface loaded from a space does not work


osanseviero:
### Describe the bug

In https://huggingface.co/spaces/osanseviero/AnimeGANv2-webcam, I'm overriding the inputs and description of the interface, but when launched, it shows the exact same interface to the original.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Check out https://huggingface.co/spaces/osanseviero/AnimeGANv2-webcam/blob/main/app.py or run in a Colab

```
!pip install gradio==3.1.3
import gradio as gr

description = "Gradio Demo for AnimeGanv2 using your webcam! Find original repo at https://huggingface.co/spaces/akhaliq/AnimeGANv2."

gr.Interface.load(
  "spaces/akhaliq/AnimeGANv2", 
  inputs=[
    gr.inputs.Image(label="Input Image", source="webcam"),
    gr.inputs.Radio(['version 1 (🔺 stylization, 🔻 robustness)','version 2 (🔺 robustness,🔻 stylization)'], type="value", default='version 2 (🔺 robustness,🔻 stylization)', label='version')
  ],
  article="",
  description=description,
).launch(debug=True)
```

### Logs

```shell
/usr/local/lib/python3.7/dist-packages/gradio/inputs.py:257: UserWarning: Usage of gradio.inputs is deprecated, and will not be supported in the future, please import your component from gradio.components
  "Usage of gradio.inputs is deprecated, and will not be supported in the future, please import your component from gradio.components",
/usr/local/lib/python3.7/dist-packages/gradio/deprecation.py:40: UserWarning: `optional` parameter is deprecated, and it has no effect
  warnings.warn(value)
/usr/local/lib/python3.7/dist-packages/gradio/inputs.py:183: UserWarning: Usage of gradio.inputs is deprecated, and will not be supported in the future, please import your component from gradio.components
  "Usage of gradio.inputs is deprecated, and will not be supported in the future, please import your component from gradio.components",
[...]

ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/dist-packages/starlette/responses.py", line 331, in __call__
    stat_result = await anyio.to_thread.run_sync(os.stat, self.path)
  File "/usr/local/lib/python3.7/dist-packages/anyio/to_thread.py", line 32, in run_sync
    func, *args, cancellable=cancellable, limiter=limiter
[...]

PR #2694, opened 2022-11-22
Adds better warnings / error messages for `gr.Interface.load()`

abidlabs:
This PR:

* Fixes: #2693 
* Fixes: #2671
* Adds a test for the latter
* Tidies up `test/test_external.py` to use `gr.Audio` instead of `gr.component.Audio`, and so on


aliabid94 (approved):
LGTM

Issue #2603, opened 2022-11-04
allow_flagging="auto" not working on hugging face


jbpolle:
### Describe the bug

Hello,

First of all thanks for this great library, really easy to use, well documented and useful.

Regarding the bug, it seems that the allow_flagging="auto" feature doesn't work. I am using it with flagging_callback type HuggingFaceDatasetSaver.
If I change to "manual" mode, the data are correctly saved in the dataset. My understanding of the doc is that in auto mode data would be saved automatically after each inference? 
Doc says: "auto: users will not see a button to flag, but every sample will be flagged automatically."

I am using gradio 3.9

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Created a space here:
https://huggingface.co/spaces/Jean-Baptiste/test_gradio

As expected there is no flagging buttons, but after inference no data are saved in the HF dataset

### Logs

```shell
/home/user/.local/lib/python3.8/site-packages/gradio/interface.py:330: UserWarning: Currently, only the 'default' theme is supported.
  warnings.warn("Currently, only the 'default' theme is supported.")
/home/user/.local/lib/python3.8/site-packages/huggingface_hub/hf_api.py:101: FutureWarning: `name` and `organization` input arguments are deprecated and will be removed in v0.10. Pass `repo_id` instead.
  warnings.warn(
Cloning https://huggingface.co/datasets/Jean-Baptiste/crowdsourced-calculator-demo into local empty directory.
Running on local URL:  http://0.0.0.0:7860

To create a public link, set `share=True` in `launch()`.
```

### System Info

```shell
Gradio 3.9
```

### Severity

annoying

freddyaboulton:
Sorry about this @jbpolle ! Yes this is a bug we'll look into.

PR #2695, opened 2022-11-22
Fixes flagging when `allow_flagging` is set to "auto"

abidlabs:
Turns out we didn't implement `allow_flagging` set to "auto" when we re-implemented `Interface` to be constructed from `Blocks`. This restores that functionality. 

Test with a script like this:

```py
import os
import gradio as gr

    if operation == "add":
        return num1 + num2
    elif operation == "subtract":
        return num1 - num2
    elif operation == "multiply":
        return num1 * num2
    elif operation == "divide":
        return num1 / num2

iface = gr.Interface(
    calculator,
    ["number", gr.Radio(["add", "subtract", "multiply", "divide"]), "number"],
    "number",
    allow_flagging="auto",
)

iface.launch()
```

Fixes: #2603 

After discussion with @aliabid94, modified the functionality of "auto" to only log the input data. This simplifies the implementation considerably because there's no easy way to flag the data "once the prediction has completed running", and we think approach is reasonable for this relatively rare use case. Updated the documentation to reflect this functionality


gradio/flagging.py
```diff
@@ -548,3 +548,17 @@ def get_unique_name(self):
...
+
+class FlagMethod:
+    """
+    Helper class that contains the flagging button option and callback
+    """
+
+    def __init__(self, flagging_callback, flag_option=None):
+        self.flagging_callback = flagging_callback
+        self.flag_option = flag_option
+        self.__name__ = "Flag"
```
aliabid94, line 561: why set `__name__`?

test/test_flagging.py
```diff
@@ -117,3 +118,19 @@ def test_flagging_no_permission_error_with_flagging_disabled(self):
                 self.fail("launch() raised a PermissionError unexpectedly")
 
         io.close()
+
+
+class TestInterfaceConstructsFlagMethod:
```
aliabid94, line 123: feels like this tests internal implementation rather than functionality, idk if needed

aliabid94 (approved):
LGTM

gradio/flagging.py
abidlabs: This existed in the `FlagMethod` class already, I didn't add it. However, I think the reason for it is because in Blocks, the event listener needs a name for the function

abidlabs:

Issue #538, opened 2022-02-03
Add LaTeX support for Markdown in `description` and `article`


abidlabs:
**Is your feature request related to a problem? Please describe.**
Right now, we can't easily add mathematical notation to the description and article. 

**Describe the solution you'd like**
Ideally, if any math is passed between $ signs, it should get rendered into LaTeX. Several libraries exist to do this, particularly KaTeX: https://katex.org/

lvwerra:
This would be a great feature - especially for metrics we would like to display the formula's to calculate them.

mishig25:
for reference, we are already using katex [here](https://github.com/huggingface/doc-builder/blob/b648e065ff4f0d391d6a063e6a91f27e541ef4fb/kit/preprocess.js#L5) in doc-buidler for rendering latex in docs.

please let me know if you have any questions regarding parsing or katex :)

PR #2696, opened 2022-11-22
Latex support

kashif:
# Description

Started to add math support for markdown

Closes: #538 

# Checklist:

# A note about the CHANGELOG


gradio/components.py
```diff
@@ -2500,7 +2501,7 @@ def __process_markdown(cls, data: List[List[Any]], datatype: List[str]):
             return data
 
         if cls.markdown_parser is None:
-            cls.markdown_parser = MarkdownIt().enable("table")
+            cls.markdown_parser = MarkdownIt().use(dollarmath_plugin).enable("table")
```
pngwn, line 2504: I'm not sure we want to support math syntax in tables, we certainly want to test this thoroughly as `display mode` will probably cause problems. Inline mode is probably fine.

ui/packages/markdown/src/Markdown.svelte
```diff
@@ -1,14 +1,28 @@
...
+		let latex_elements = target.querySelectorAll(".math");
+		if (latex_elements.length === 0) {
+			return;
+		}
+
+		if (katex === undefined) {
+			katex = await import("./katex");
+		}
+		katex.render(latex_elements);
+	});
```
pngwn, line 24: This only needs to happen when the `value` changes rather than when _any_ props change.

Pull the body of this into a function and use a reactive statement:

```ts
aysnc function handle_katex() {
  let latex_elements = target.querySelectorAll(".math");
  if (latex_elements.length === 0) {
    return;
  }

  if (katex === undefined) {
    katex = await import("./katex");
  }
  katex.render(latex_elements);
}
```

And:

```ts
$: value && handle_katex();
```

I'd also test this with some more complex blocks. You may need to `await tick()` at the very beginning, to ensure changes have been flushed to the DOM.

ui/packages/markdown/src/katex/package-lock.json
```diff
@@ -0,0 +1,54 @@
+{
```
pngwn, line 1: This needs to be removed, we don't use `npm` to manage packages, we use `pnpm`.

pngwn (dismissed):
We don't want to vendor in the css and fonts here, that will be difficult to maintain. We can load them directly from the node package via `katex/path/to/assets`. Vite can handle css and fonts.

ui/packages/markdown/src/Markdown.svelte
aliabid94: now rendering is server-side, but good to know!

aliabid94:
I've implemented support for rendering mathtext serverside, a subset of latex that ships with matplotlib. Screenshot from demo/latex/run.py

Screen Shot 2022-11-28 at 12 54 48 PM

kashif:
nice!

gradio/components.py
aliabid94: I don't believe mathtext supports display mode, only inline latex.

gradio/components.py
```diff
@@ -27,6 +27,7 @@
 import PIL.ImageOps
 from ffmpy import FFmpeg
 from markdown_it import MarkdownIt
+from mdit_py_plugins.dollarmath import dollarmath_plugin
```
abidlabs, line 30: Do we need to add anything to `requirements.txt`?

abidlabs: Never mind

aliabid94: pretty sure it comes with `markdown_it` automatically

abidlabs:
Very nice. I added some LaTeX to `demo/blocks_kinematics` and it looks great!

image

Two nits:
* Let's add documentation for this (maybe in the docstring for `Markdown` component) as well as tests to prevent regressions
* I get this warning when I include LaTeX:

```
savefig() got unexpected keyword argument "frameon" which is no longer supported as of 3.3 and will become an error two minor releases later
```

You might have to set `facecolor='white', transparent=False` instead according to this answer: https://stackoverflow.com/questions/59204749/matplotlib-savefig-background-always-transparent

abidlabs (approved):
LGTM besides a couple of nits (see above)

kashif:
@abidlabs and how does it look in "dark mode"?

abidlabs:
@kashif good question -- it looks good to me: 

image

CHANGELOG.md
```diff
@@ -1,7 +1,10 @@
 # Upcoming Release
 
 ## New Features:
-No changes to highlight.
+
+### Latex support
+Added mathtext (a subset of latex) support to gr.Markdown. Added by [@kashif](https://github.com/kashif) in [PR 2696](https://github.com/gradio-app/gradio/pull/2696).
```
abidlabs, line 6:
```suggestion
Added mathtext (a subset of latex) support to gr.Markdown. Added by [@kashif](https://github.com/kashif) and [@aliabid94](https://github.com/aliabid94) in [PR 2696](https://github.com/gradio-app/gradio/pull/2696).
```

abidlabs: I would really add an example here since this is a huge addition!

aliabid94:
also there is a hidden formula span, so if you try to copy text with latex, it will render the formula.

aliabid94:
@abidlabs I am not able to reproduce your warning - can you check if your suggestion works?

abidlabs:
> @abidlabs I am not able to reproduce your warning - can you check if your suggestion works?

I upgraded to the latest version of matplotlib and it is indeed throwing an error! Sure I'll see if I can fix it on my end

abidlabs:
Ok so I just removed the `frameon` parameter and it looks like it's fine now @aliabid94

gradio/components.py
kashif: should i then add the `utils.tex2svg` renderer here?

aliabid94: oh good catch, I'll do that

kashif:
awesome! thanks!

PR #2705, opened 2022-11-22
Make try examples button more prominent

aliabd:
We've gotten feedback from several users that the Try Examples website in the docs is easy to miss. So I've made it more prominent. 

* If the object has an 'Example Usage' section, there will be a button under it named 'More Examples' 
* If the object doesn't have an 'Example Usage' section, there will be a greyed out div of the first demo's code with the button overlayed on top. 

Feedback on styling and wording welcome. To see this just checkout the branch and run `sh scripts/launch_website.sh`

https://user-images.githubusercontent.com/9021060/203439981-bb385de0-e063-4f0c-9014-b5589b39a45c.mov


lewtun:
This looks wonderful @aliabd! 

Not related to this PR, but I think having an "Example Usage" section for most components would go a long way towards making the docs more intuitive / accessible (kind of how every model in `transformers` has a little example you can copy-paste and play with)

aliabd:
Agreed @lewtun, we support that already on the website side, now just need to add it to the docstrings for the components that don't have it. Will get that in

abidlabs:
Very nice @aliabd!

PR #2706, opened 2022-11-23
Audio waveform

dawoodkhan82:
# Description
Add a utility function that creates a video when supplied with audio and image files. 

https://user-images.githubusercontent.com/12725292/203449101-39a603be-fe8f-4ca5-9b24-1d06ad2db812.mov

Test:

```python
def audio_waveform(audio, image):
    print(audio.name)
    return utils.audio_to_video(audio.name, image.name)


gr.Interface(audio_waveform, inputs= [gr.Audio(type="file"), gr.Image(type="file")], outputs=gr.Video()).launch()
```

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

Closes: # (issue)

# Checklist:

# A note about the CHANGELOG


dawoodkhan82:
thoughts on adding this as a flag to the audio or video component?

dawoodkhan82:
Also tried to make this look prettier, but I'm limited with what can be done with ffmpeg. (https://trac.ffmpeg.org/wiki/Waveform)

abidlabs:
Awesome @dawoodkhan82! A couple of thoughts:

* What about adding a `image` parameter to `Audio` component, which if supplied, makes the Audio component appear like this?
* The waveform looks a little jagged / sharp to me. Is it possible to make it a little more discrete by putting little bars instead? Here are a couple of examples of what imo looks a little bit nicer:

image

image

abidlabs:
Oh just saw your comment about ffmpeg's limitation, that's okay then imo! cc @aliabid94 @pngwn @gary149 for your thoughts as well!

dawoodkhan82:
> Awesome @dawoodkhan82! A couple of thoughts:
> 
> * What about adding a `image` parameter to `Audio` component, which if supplied, makes the Audio component appear like this?
> * The waveform looks a little jagged / sharp to me. Is it possible to make it a little more discrete by putting little bars instead? Here are a couple of examples of what imo looks a little bit nicer:
> 
> image
> 
> image

That would require our frontend `audio` component to also be a `video` component. Which is fine imo.

pngwn:
> That would require our frontend audio component to also be a video component. Which is fine imo.

I disagree, I think we just change the backend to generate an `video` component in the config when those options are passed. I don't want the audio component being a video component as well, that doesn't make any sense and means we need to ship more unnecessary code (at least until we have more granular on demand loading).

abidlabs:
We shouldn’t change the config — among other things, it would break Interface.load(), which recreates an Interface based on its config. But would it be possible in this case for the Audio component to “import” the code for the Video component and pass along the video in the frontend  to avoid duplicating code?

pngwn:
Duplicate code isn't the issue, unused code is. The video would be shipped for anyone using the Audio, even if they weren't using the 'to video' functionality. We can fix this in the future when we support more granular on demand loading.

aliabid94:
So I've been a lotta time looking into this, and this is where I'm at.

If we use the FFMPEG library, then we get the waveform @dawoodkhan82 posted above, it takes a somewhat reasonable time to run (maybe 5 seconds for 30 seconds of audio) but it doesn't look very good. I think it actually looks bad enough to hurt virality. I played around with the FFMPEG arguments and couldn't get it looking much better, though I think if I can at least make the wave form symmetric across the horizontal axis through some clever FFMPEG manipulation, it might look a little better. 

An alternative is to write some custom python code to generate a custom waveform. I looked into this, and was able to a very nice looking animated waveform. However this is much slower, about as slow as requiring half a second per second of audio to render. So this would take about 18 seconds of extra processing to render a waveform video for a 30 second audio clip. This is obviously unacceptable. The reason it's so slow is because we have to save an image for every single frame of the video, and then we use FFMPEG to compile all the images into a single video. 

A third solution might be to have a static image with a static waveform representing the whole audio clip combined into a video. Nothing would animate over the course of the audio playing. A benefit would be rendering would be extremely fast. It would be a shareable video. However it's not as aesthetic as an animation obviously.

Let me know what you guys think.

abidlabs:
I am inclined towards option 3, as I think it's best to keep things simple and not add too much inference time. If we could have a single audio waveform representing the entire audio file overlayed on top of a static image, and it looks nice, let's go for it. Can you share an example of what that would look like?

Maybe we could add some flair on top of that, like as the video plays, it colors the part of the waveform that you are at. But again, I would only do that if it's fairly simple and if it improves aesthetics/shareability

freddyaboulton:
I also think option 3 is the way to go but seeing examples of all three would help evaluate the trade off !

aliabid94:
Okay, here it is.

For a simple waveform:
```python
import gradio as gr

def audio_waveform(audio):
    return gr.Waveform(audio)

gr.Interface(
    audio_waveform, gr.Audio(type="filepath"), gr.Audio()
).launch()
```
which creates (click play on the recordings below):

Recording 2022-12-06 at 17 02 28

You can customize the waveform visual in many ways, such as adding background image, number and styling of bars, etc.

Example:
```python
import gradio as gr

def audio_waveform(audio, image):
    return gr.Waveform(audio, bg_image=image)

gr.Interface(
    audio_waveform,
    inputs=[gr.Audio(type="filepath"), gr.Image(type="filepath")],
    outputs=gr.Audio(),
).launch()
```
Recording 2022-12-06 at 17 03 21

It is very fast, ~1s to render the video. This is because the video is simply an image, (static waveform ) and audio, so it is light and quick to render. The playback animation seen above of the gray bar crossing the video is actually rendered via JS, so it will not exist as part of any copied / shared video.

We needed to create a separate frontend Waveform,svelte component to render the playback animation.

abidlabs:
I'm a little confused about this Python API:

```
import gradio as gr

def audio_waveform(audio, image):
    return gr.Waveform(audio, bg_image=image)

gr.Interface(
    audio_waveform,
    inputs=[gr.Audio(type="filepath"), gr.Image(type="filepath")],
    outputs=gr.Audio(),
).launch()
```

Why are you returning a `gr.Waveform` from the function? Shouldn't the function return `gr.update(audio, bg_image=image)`

aliabid94:
Okay made animation part of ffmpeg video generation. The default way of configuring a waveform is by setting it up in the Audio component constructor. e.g.
```python
gr.Audio(waveform=True)
```
or
```python
gr.Audio(waveform=gr.Waveform(bar_color="green", bar_count="200", bg_image="cat.jpg"))
```

however if you wish to update the waveform in the function, you cannot use `gr.update()` because we currently do not have anywhere we store updated config values for a session in the backend, only the frontend. So to return an updated waveform (e.g. to change the bg_image), you must return a new waveform

or
```python
return gr.Waveform(audio="audio.wav,  bg_image="cat.jpg")
```

demo/waveform/run.py has examples of all of these. See below
Recording 2022-12-09 at 00 37 37

abidlabs:
Looks beautiful @aliabid94! Taking a closer look, but I wanted to clarify what you meant by this comment:

> however if you wish to update the waveform in the function, you cannot use gr.update() because we currently do not have anywhere we store updated config values for a session in the backend, only the frontend

Why do you need to update anything on the backend?

gradio/utils.py
```diff
@@ -805,6 +811,142 @@ def __str__(self):
...
+        plt.bar(
+            np.arange(0, self.bar_count),
+            samples * 2,
+            bottom=(-1 * samples),
+            width=self.bar_width,
+            color=color,
+        )
+        plt.axis("off")
+        plt.margins(x=0)
+        tmp_img = tempfile.NamedTemporaryFile(suffix=".png")
```
abidlabs, line 900:
```suggestion
        tmp_img = tempfile.NamedTemporaryFile(suffix=".png", delete=False)
```

abidlabs:
Sharing my thoughts on the Python API, I think it's a little bit confusing and overcomplicated for the most common use cases. Here's my suggestion for how to make it easier, while still retaining the cool functionality that you've included here:


**Use case 1: someone wants to make a quick and dirty waveform to display a static audio file with a background image.**

Suggested API:

```py
with gr.Blocks() as demo:
    gr.Audio("test.wav", image="stars.png")
demo.launch()
``` 

In other words, if an `image` is supplied, then a waveform is automatically created with default parameters

**Use case 2: you want to show a generated audio file with a fixed image** 

```py
gr.Interface(fn, "text", gr.Audio(image="stars.png")).launch()
``` 

So `image` is a parameter like any other parameter and gets saved to the config

**Use case 3: you want to show a generated audio file with randomly selected images** 

```py

def fn(text):
  ....
  return gr.update("audio.wav", image=random_image)

gr.Interface(fn, "text", "audio").launch()
``` 

So you can update `image` like any other parameter in the config

**Use case 4: you want to customize the waveform** 

This is an advanced use case so it's a little more complex:

```py
gr.Interface(fn, "text", gr.Audio(image="stars.png", waveform=gr.Waveform(fg_alpha=0.5))).launch()
```

But this is the only case you need to know about the `gr.Waveform` class.

aliabid94:
Okay update:

you can set wavefoem to True, a background image, or a waveform object. Added ability to save configuration updates in the backend, so we now support `gr.update` with waveform.

Example code showing all possible features:
```python
def audio_waveform(audio, image):
    return (
        audio,
        audio,
        audio,
        audio,
        gr.Audio.update(
            value=audio,
            waveform=gr.Waveform(bg_image=image, bars_color=random.choice(COLORS)),
        ),
    )


gr.Interface(
    audio_waveform,
    inputs=[gr.Audio(type="filepath"), gr.Image(type="filepath")],
    outputs=[
        gr.Audio(), # basic output
        gr.Audio(waveform=True), # autogenerated waveform
        gr.Audio(waveform="drake.jpg"), # waveform with set background image
        gr.Audio(
            waveform=gr.Waveform( # custom waveform
                bars_color=("#00ff00", "#0011ff"), bar_count=100, bg_color="#000000"
            )
        ),
        gr.Audio(waveform=True), # waveform is updated with gr.Audio.update
    ],
).launch()
```

freddyaboulton:
@aliabid94 This is working great!

I'm wondering if we can avoid modifying the postprocess signature of every component in order to ship it.

If users want to update their waveform via gr.Audio.update, can we make something like the following work?

def update_waveform(audio):
    new_audio = gr.WaveForm(image="drake.jpg").make_video(audio)
    return gr.Audio.update(value=new_audio)

It is more code that users have to write but I think it's fine to expect users to do the processing in their event handlers and that way we don't have to modify all the component postprocess functions. 

It's confusing that the front-end audio component is now serving video as well. I think the original approach where we just provide a helper class to convert audio to waveform (but it's on the user to add a gr.Video component) is a lot simpler architecturally. How come we moved away from that?

abidlabs:
> I'm wondering if we can avoid modifying the postprocess signature of every component in order to ship it.

+1. We should keep `.postprocess()` method signature as straightforward as possible especially as people will be writing their own for custom components. 
 
> It's confusing that the front-end audio component is now serving video as well. I think the original approach where we just provide a helper class to convert audio to waveform (but it's on the user to add a gr.Video component) is a lot simpler architecturally. How come we moved away from that?

This will lead to some limitations -- for example, it won't be possible to extract the audio and use it as the input for a different component. But overall, this might be the saner solution (and we might not even need a `gr.Waveform` class -- we could just generate a waveform `gr.make_waveform(**params)` which generates a Video)

aliabid94:
But we DO want preprocess and postprocess to be able to take into account updated configuration, right? for example with the Radio element and updated choices. To do that, we'd need to update these methods.

gradio/components.py
```diff
@@ -2042,18 +2077,30 @@ def postprocess(self, y: Tuple[int, np.array] | str | None) -> str | None:
...
+                file = processing_utils.create_tmp_copy_of_file(y, dir=self.temp_dir)
         elif isinstance(y, tuple):
             sample_rate, data = y
             file = tempfile.NamedTemporaryFile(
                 suffix=".wav", dir=self.temp_dir, delete=False
             )
             processing_utils.audio_to_file(sample_rate, data, file.name)
         else:
-            file = processing_utils.create_tmp_copy_of_file(y, dir=self.temp_dir)
+            print(">", y.__class__)
```
abidlabs, line 2100:
```suggestion
```

abidlabs:
> But we DO want preprocess and postprocess to be able to take into account updated configuration, right? for example with the Radio element and updated choices. To do that, we'd need to update these methods.

We definitely need to update the methods, but is the best approach changing the method signatures? A related issue that I see here is that we only save certain fields in the state dictionary, but I feel like this can lead to bugs where users forget to add certain configuration parameters to the state dictionary. 

What about the following approach instead: what if every component stores *all* of its parameters in a state dictionary, where the key is the session hash and the value is a dict of config parameters that have been updated in that session. In `preprocess()` and `postprocess()`, we don't access the class attributes directly, instead we call `get_config()` to get the component parameters for that particular state?

aliabid94:
we could do something along those lines, but I feel like it'd make more sense to take a second argument called configuration, which takes all the component configurations., rather than a `get_config()` method e.g.
```python
class Radio:
  def postprocess(y, config):
    if config.type == "index":
      return config.choices.index(y):
    else:
      return y
```

why don't we want to modify the method signature exactly?

freddyaboulton:
Yea I think storing the latest state in the front-end is a great feature we'll add in the future but I'm not sure if it's the best way to ship the waveform component. I think the simplest thing to do to close this out is the `gr.make_waveform(**params)` approach @abidlabs commented above. We could hash out the design details/implementation of storing state in the backend in a separate issue/PR.

abidlabs:
> we could do something along those lines, but I feel like it'd make more sense to take a second argument called configuration, which takes all the component configurations., rather than a get_config() method e.g.

Actually I like this more. I didn't want a second parameter that users would have to reason about as they were designing their component, but config is pretty intuitive IMO.

aliabid94:
So that's what I had initially, I had a `return gr.Waveform` type solution and @abidlabs didn't like that, because so far we've always had return types by standard data types, and things related to display configuration are set in the component constructor. That's why I changed to a `gr.update` solution. 
I agree this is a structural change that requires discussion, but let's just have that discussion now?  Especially since this is needed to solve other bugs like the `gr.Radio` and choices update issue we mentioned earlier.

abidlabs:
> So that's what I had initially, I had a return gr.Waveform type solution and @abidlabs didn't like that (and I agreed), because so far we've always had return types be standard data types, and things related to display configuration are set in the component constructor. That's why I changed to a gr.update solution.

Actually what @freddyaboulton is suggesting is different -- he's not saying return the `gr.Waveform` but instead create a utility function (say `gr.make_waveform()`) that returns a video, which we *pass into the Video component*. Are you opposed to this solution?

aliabid94:
So the output component is now Video, rather than audio?

abidlabs:
Exactly

aliabid94:
ok sounds good, I think that works for now

aliabid94:
ok implemented @freddyaboulton's suggestion, now a much simpler PR. we can figure out backend configuration later.

```python
def audio_waveform(audio, image):
    return (
        audio,
        gr.make_waveform(audio),
        gr.make_waveform(audio, bg_image=image, bars_color=random.choice(COLORS)),
    )


gr.Interface(
    audio_waveform,
    inputs=[gr.Audio(), gr.Image(type="filepath")],
    outputs=[
        gr.Audio(),
        gr.Video(),
        gr.Video(),
    ],
).launch()

```

aliabid94:
ready for rereview

CHANGELOG.md
```diff
@@ -137,6 +137,9 @@ Add Brazilian Portuguese translation (pt-BR.json) by [@pstwh](http://github.com/
 
 <img width="951" alt="image" src="https://user-images.githubusercontent.com/1778297/206615305-4c52031e-3f7d-4df2-8805-a79894206911.png">
 
+### Add Waveform Visual Support to Audio
+Add Waveform visual support to audio component, for output audio only [@dawoodkhan82](http://github.com/dawoodkhan82) in [PR 2706](https://github.com/gradio-app/gradio/pull/2706. Helpful for making audio outputs much more shareable.
```
abidlabs, line 141:
```suggestion
Add Waveform visual support to audio component, for output audio only by [@dawoodkhan82](http://github.com/dawoodkhan82) and [@aliabid94](http://github.com/aliabid94) in [PR 2706](https://github.com/gradio-app/gradio/pull/2706. Helpful for making audio outputs much more shareable.
```

gradio/utils.py
```diff
@@ -805,6 +811,121 @@ def __str__(self):
...
+    :param fg_alpha: Opacity of foreground waveform
+    :param bars_color: Color of waveform bars. Can be a single color or a tuple of (start_color, end_color) of gradient
+    :param bar_count: Number of bars in waveform
+    :param bar_width: Width of bars in waveform. 1 represents full width, 0.5 represents half width, etc.
+    """
+    if isinstance(audio, str):
+        audio_file = audio
+        audio = processing_utils.audio_from_file(audio)
+    else:
+        tmp_wav = tempfile.NamedTemporaryFile(suffix=".wav")
```
abidlabs, line 837: Needed for Windows
```suggestion
        tmp_wav = tempfile.NamedTemporaryFile(suffix=".wav", delete=False)
```

gradio/utils.py
```diff
@@ -805,6 +811,121 @@ def __str__(self):
...
+    plt.bar(
+        np.arange(0, bar_count),
+        samples * 2,
+        bottom=(-1 * samples),
+        width=bar_width,
+        color=color,
+    )
+    plt.axis("off")
+    plt.margins(x=0)
+    tmp_img = tempfile.NamedTemporaryFile(suffix=".png")
```
abidlabs, line 884:
```suggestion
    tmp_img = tempfile.NamedTemporaryFile(suffix=".png", delete=False)
```

gradio/utils.py
```diff
@@ -805,6 +811,121 @@ def __str__(self):
...
+    bar_width: float = 0.6,
+):
+    """
+    :param audio: Audio file path or tuple of (sample_rate, audio_data)
+    :param bg_color: Background color of waveform (ignored if bg_image is provided)
+    :param bg_image: Background image of waveform
+    :param fg_alpha: Opacity of foreground waveform
+    :param bars_color: Color of waveform bars. Can be a single color or a tuple of (start_color, end_color) of gradient
+    :param bar_count: Number of bars in waveform
+    :param bar_width: Width of bars in waveform. 1 represents full width, 0.5 represents half width, etc.
```
abidlabs, line 831: Docstring style is different from everywhere else in the codebase

abidlabs: I would also add a high-level comment explaining what this function is for, as we expose it directly to users. 

Can we add it to the docs as well? Perhaps under the https://gradio.app/docs/#component-helpers section

abidlabs:
@aliabid94 works and looks beautiful! I added some comments above. We should make sure this is documented well in the docs page as it's a user-facing function. In addition, I think it would be good to write a unit test for it so we don't break it in the future.

gradio/utils.py
aliabid94: disagree, why should intermediate files not have delete=True? When this function ends, the files will be cleaned up by garbage collection which is what we want,

gradio/utils.py
aliabid94: disagree for same reason

gradio/utils.py
```diff
@@ -805,6 +811,125 @@ def __str__(self):
...
+    Parameters:
+        audio: Audio file path or tuple of (sample_rate, audio_data)
+        bg_color: Background color of waveform (ignored if bg_image is provided)
+        bg_image: Background image of waveform
+        fg_alpha: Opacity of foreground waveform
+        bars_color: Color of waveform bars. Can be a single color or a tuple of (start_color, end_color) of gradient
+        bar_count: Number of bars in waveform
+        bar_width: Width of bars in waveform. 1 represents full width, 0.5 represents half width, etc.
+    Returns:
+        A filepath to the output video.
```
abidlabs, line 835: Nit
```suggestion
        A filepath to the output video that can be passed into a `gr.Video` component
```

gradio/utils.py
```diff
@@ -805,6 +811,125 @@ def __str__(self):
...
+    *,
+    bg_color: str = "#f3f4f6",
+    bg_image: str = None,
+    fg_alpha: float = 0.75,
+    bars_color: str | Tuple[str, str] = ("#fbbf24", "#ea580c"),
+    bar_count: int = 50,
+    bar_width: float = 0.6,
+):
+    """
+    Generates a waveform video from an audio file. Useful for creating an easy to share audio visualization.
```
abidlabs, line 825:
```suggestion
    Generates a waveform video from an audio file. Useful for creating an easy to share audio visualization. The output should be passed into a `gr.Video` component
```

abidlabs:
By the way, we can remove the `Waveform.svelte` now, correct? And

aliabid94:
Added docs and removed frontend files @abidlabs

abidlabs:
Why did we switch to camelcase instead of the original [more pythonic](https://stackoverflow.com/a/8908798/5209347) `gr.make_waveform`?

CHANGELOG.md
```diff
@@ -175,6 +175,9 @@ Add Brazilian Portuguese translation (pt-BR.json) by [@pstwh](http://github.com/
 
 <img width="951" alt="image" src="https://user-images.githubusercontent.com/1778297/206615305-4c52031e-3f7d-4df2-8805-a79894206911.png">
 
+### Add Waveform Visual Support to Audio
```
freddyaboulton, line 178: Add this to the latest release section please

CHANGELOG.md
```diff
@@ -175,6 +175,9 @@ Add Brazilian Portuguese translation (pt-BR.json) by [@pstwh](http://github.com/
...
+Adds a `gr.makeWaveform()` function that creates a waveform video by combining an audio and an optional background image by [@dawoodkhan82](http://github.com/dawoodkhan82) and [@aliabid94](http://github.com/aliabid94) in [PR 2706](https://github.com/gradio-app/gradio/pull/2706. Helpful for making audio outputs much more shareable.
```
freddyaboulton, line 179: nit: Would be really cool to embed a gif of the waveform here. If you edit from the github ui directly, you can drag-and-drop a gif.

freddyaboulton (approved):
Awesome PR @aliabid94 ! Thanks for making the changes. Agree with @abidlabs that snake case is more common for a helper function.

gradio/components.py
```diff
@@ -3269,10 +3269,6 @@ def postprocess(
                     running_category = category
                 elif category == running_category:
                     running_text += self.adjacent_separator + text
-                elif not text:
```
freddyaboulton, line 3272: This got added by mistake no?

abidlabs: Yeah it looks like it. That's why the test is failing!

CHANGELOG.md
aliabid94: done

CHANGELOG.md
aliabid94: done.

abidlabs (approved):

abidlabs:
LGTM, awesome @aliabid94!

Issue #2700, opened 2022-11-22
Using `gr.File` as an input fails with large files when the queue is activated


apolinario:
### Describe the bug

If a `gr.File` working as an input, gets either one large file or small files that sum to large-ish values (>10MB) uploaded to it, Gradio will not process the upload and will show a `Connection errored out` error. This only happens if the queue is active.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

This demo fails if you upload multiple 5MB images (3+) or single big files:
```py
import gradio as gr

def process(files):
    print(files)
    return "It worked!"
    
with gr.Blocks() as demo:
    files = gr.File(file_count="multiple")
    button = gr.Button()
    output = gr.Textbox()
    button.click(fn=process,inputs=files,outputs=output)
    
demo.queue(default_enabled=True).launch(debug=True)
```
Removing the queue works. 

### Screenshot

![image](https://user-images.githubusercontent.com/788417/203387191-3299f79b-e803-409b-97b6-2f6aa34bd714.png)

### Logs

```shell
No logs
```

### System Info

```shell
Gradio==3.10.1
```

### Severity

blocking upgrade to latest gradio version

PR #2709, opened 2022-11-23
Increase websocket size to 1 GB

abidlabs:
Thanks @aliabid94 for the help chasing this down, ended up being a pretty simple fix :)

Fixes: #2700


abidlabs:
You can test by going to the "zip_files" demo in the PR deploy and zipping files that are > 12 MB in size (which would previously fail)

aliabid94 (approved):
LGTM. tested with a 150 MB file on spaces (though it did seem to put strain on the ui)

Issue #2704, opened 2022-11-22
Recording Audio and Video from webcam simultaneously


MandarGogate:
- [X] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
I want to build a demo for machine learning model that processes Audio-Visual data from webcam. The current Video component records from webcam without audio. 

**Describe the solution you'd like**  
Since the Video component uses MediaRecorder API it can record audio and video simultaneously.

MandarGogate:
I managed to update the Webcam.svelte to fix this. You can have a look at [https://github.com/MandarGogate/gradio/commit/2d291e7f311192899e01b5f82133905ce1dace67](https://github.com/MandarGogate/gradio/commit/2d291e7f311192899e01b5f82133905ce1dace67). I tested this on Safari, Chrome and Firefox.

abidlabs:
Great, thanks so much @MandarGogate! Would you be able to open a PR and happy to review it?

PR #2721, opened 2022-11-25
enhancement: record audio and video from webcam simultaneously

jeeanribeiro:
# Description

From @MandarGogate issue:
> **Is your feature request related to a problem? Please describe.**
I want to build a demo for machine learning model that processes Audio-Visual data from webcam. The current Video component records from webcam without audio.
> **Describe the solution you'd like**
Since the Video component uses MediaRecorder API it can record audio and video simultaneously.

Changes:
```
- Enables audio in Webcam.svelte inside `access_webcam` function
```

Closes: #2704

# Checklist:

- [x] I have performed a review of the code
- [x] The code follows the style guidelines of this project


jeeanribeiro:
@MandarGogate
I think that for this PR to be mergeable, it just needs the summary of the changes added in CHANGELOG.md, you can check the instructions about it here: https://github.com/gradio-app/gradio/blob/main/.github/PULL_REQUEST_TEMPLATE.md
As I'm trying to merge your fork, I don't have the permission to do this change

@abidlabs 
Waiting for your review :smile:

pngwn:
We should probably make this an option to ensure the author has choice and that users aren't bombarded with permission requests when they run apps. Many models don't need or can't use both audio and video inputs.

abidlabs:
Agreed with @pngwn‘s suggestion — ideally we have a Boolean parameter e.g. “include_audio” which can be True by default and applies to the webcam scenario but also to the regular Video component. In other words, if it is False then the audio from an uploaded Video would be removed as well

MandarGogate:
I have added `include_audio` parameter. You can try it yourself using [https://gist.github.com/MandarGogate/d49d1fd871b85c9840d5df7656c10c41](https://gist.github.com/MandarGogate/d49d1fd871b85c9840d5df7656c10c41)

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ No changes to highlight.
 * Make try examples button more prominent by [@aliabd](https://github.com/aliabd) in [PR 2705](https://github.com/gradio-app/gradio/pull/2705)
 * Fix id clashes in docs by [@aliabd](https://github.com/aliabd) in [PR 2713](https://github.com/gradio-app/gradio/pull/2713)
 * Fix typos in guide docs by [@andridns](https://github.com/andridns) in [PR 2722](https://github.com/gradio-app/gradio/pull/2722)
+* Add include_audio option in Video component by [@mandargogate](https://github.com/MandarGogate) in [PR 2721](https://github.com/gradio-app/gradio/pull/2721)
```
pngwn, line 22:
```suggestion
* Add option to `include_audio` in Video component. When `True`, for `source="webcam"` this will record audio and video, for `source="upload"` this will retain  the audio in an uploaded video by [@mandargogate](https://github.com/MandarGogate) in [PR 2721](https://github.com/gradio-app/gradio/pull/2721)
```

gradio/components.py
```diff
@@ -1612,7 +1613,8 @@ def __init__(
             interactive: if True, will allow users to upload a video; if False, can only be used to display videos. If not provided, this is inferred based on whether the component is used as an input or output.
             visible: If False, component will be hidden.
             elem_id: An optional string that is assigned as the id of this component in the HTML DOM. Can be used for targeting CSS styles.
-            mirror_webcam: If True webcma will be mirrored. Default is True.
+            mirror_webcam: If True webcam will be mirrored. Default is True.
+            include_audio: If False, component will only record/keep muted video.
```
pngwn, line 1617:
```suggestion
            include_audio: If False, component will not record or retain the audio track for a video.
```

pngwn (commented):
Looks good to me. Left a few suggestions for the wording of the change log and docstring.

pngwn:
Is defaulting to `True` a breaking change? Are there any times when the presence of an audio track will cause the predict function/ model that is processing the video to choke + fail?

abidlabs:
> Is defaulting to True a breaking change? Are there any times when the presence of an audio track will cause the predict function/ model that is processing the video to choke + fail?

Good point, let's test with a few Spaces that currently take in a webcam video. I believe @nateraw has built some...

abidlabs:
Sorry for the late response @jeeanribeiro. I tested this out and it turns out setting `include_audio` to be `True` by default will be a breaking change because some users have webcam -> video demos where they do not expect the webcam to be recording the video. My suggestion would be to have `include_audio` be `None`, which gets set to `True` by default if `source` is `upload` but `False` if `source` is `webcam`. Would you be able to make that change and merge in the latest changes from `main`. Happy to re-review

cc @pngwn @aliabid94

MandarGogate:
@abidlabs I have updated the default behaviour of include_audio and merged latest changes from `main`

abidlabs:
Tested and works great. Couple of tests need to be fixed but I can go ahead and do that -- thanks for this addition to the gradio library @jeeanribeiro!

Issue #2717, opened 2022-11-24
Docs are unusable on mobile 


aliabd:
### Describe the bug

Docs don't even render fully on mobile. Likely all the embedded demos are too heavy on the page. 

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Just go to www.gradio.app/docs on mobile

### Screenshot

https://user-images.githubusercontent.com/9021060/203853895-e18580a8-96e6-4e89-af7f-b1c9c5828602.mp4

### Logs

![Screen Shot 2022-11-25 at 5 10 56 PM](https://user-images.githubusercontent.com/9021060/204067120-9660fcb5-bc68-42c7-9f35-12c9dca8f43c.png)

![Screen Shot 2022-11-27 at 12 28 13 PM](https://user-images.githubusercontent.com/9021060/204162091-79f6eda1-0769-4ba8-9c1e-d76c8eddb8e0.png)

### System Info

```shell
Safari and Chrome on iPhone
```

### Severity

serious, but I can work around it

PR #2730, opened 2022-11-28
Performance improvements to docs 

aliabd:
I made some performance improvements to the docs, especially on mobile. I know we've talked about moving the website to a better framework, and we should still do that, but we have to make fixes now as they are currently unusable. 

* Enabled text compression (on nginx) 
* For mobile on load gradio js when example window is clicked. (this means no embedded components on mobile, but much better performance) 
* Figured out a memory leak that was causing the page to climb to insane amounts of memory (>9GB). Was due to some terrible js I wrote for hiding footers from embedded components. Fixed now and page stays at ~350MB.
* Fixed some annoying layout shifts 
* Made the page mobile responsive 

Before changes: 

Screen Shot 2022-11-25 at 5 10 56 PM

After: 

Screen Shot 2022-11-27 at 10 05 09 PM

Try comparing between the [website](https://gradio.app/docs/main) and [dev server](https://www.gradio-staging.com/docs/main) on mobile. Note that most of these changes will only apply to /docs/main/ until we release a new pip version and /docs gets updated. 

Closes: #2717

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ No changes to highlight.
 * Fixed bug to ensure that filenames are less than 200 characters even for non-English languages by [@SkyTNT](https://github.com/SkyTNT) in [PR 2685](https://github.com/gradio-app/gradio/pull/2685) 
 
 ## Documentation Changes:
-No changes to highlight.
+* Performance improvements to docs on mobile by  [@aliabd](https://github.com/aliabd) in [PR 2730](https://github.com/gradio-app/gradio/pull/2730)
 
 ## Testing and Infrastructure Changes:
 No changes to highlight.
--- a/website/homepage/src/docs/obj_doc_template.html
+++ b/website/homepage/src/docs/obj_doc_template.html
@@ -4,6 +4,13 @@
 <div class="obj" id="{{ obj['name'].lower() }}">
 {% endif %}
 
+{% if version == "main" %}
+{% set gradio_js = '/assets/index.js' %}
+{% else %}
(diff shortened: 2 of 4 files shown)


abidlabs:
1 -> 91 🔥🔥🔥

pngwn (approved):
I'm not seeing the 91, i get ~60 after a few tries but ~1 -> ~60 is huge. There is also a noticeable improvement in performance even without using automated testing. The old version of the docs is frequently unusable, even on desktop for me and closing it takes many seconds + locks up my browser. On mobile the current docs are completely unusable.

Regardless of whether it is 60 or 90 we should ship this asap, huge improvement. Awesome work @aliabd!

aliabd:
Hmm that's weird @pngwn, are you sure that's for /docs/main on www.gradio-staging.com and testing for mobile? I just got 93. I did see high 80s a few times but never 60. Maybe it's a latency thing! 

But yes, the number doesn't really matter.

aliabd:
Btw @pngwn do you know how we'd enable text compression on the gradio.js served from s3? like https://gradio.s3-us-west-2.amazonaws.com/3.11.0/gradio.js 

I think we'll have to zip them before uploading to s3. but not sure if there's a better way?

abidlabs:
LGTM @aliabd! The staging website loads significantly faster for me on desktop as well. 

The CircleCI test that is failing is because the Hub is down. Rerunning that should cause it to pass (although we should mark it flaky for future reference)

abidlabs (approved):

aliabd:
Will try to figure out gzipping the gradio.js s3 module on a separate PR

Issue #2498, opened 2022-10-19
Add support for images in chatbot component


dawoodkhan82:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
To support the multimodal project, the demo requires support for something like the image below.

<img width="606" alt="Capture d’écran 2022-10-19 à 11 05 07 AM" src="https://user-images.githubusercontent.com/12725292/196763541-aa6a06a5-ec50-41d3-9505-6ca9f80f71bd.png">

**Describe the solution you'd like**  
Add support for markdown in chatbot

**Additional context**

abidlabs:
Cool! I suggest we support Markdown more generally?

abidlabs:
Cool, yeah probably makes sense that people want to input an image more easily then typing out the Markdown for it

cmccomb:
Markdown support for the chatbot component would be AWESOME!

PR #2731, opened 2022-11-28
Add Markdown support in chatbots

abidlabs:
Adds support for Markdown in the chatbot component. Closes: #2498  

Demo code 1 (supply your own image instead of `lion.jpg`):

```py
import gradio as gr

with gr.Blocks(css="#chatbot .overflow-y-auto{height:700px!important}") as demo:
    gr.Chatbot([("hi", "hello **abubakar**"), ("![](/file=lion.jpg)", "cool pic")], elem_id="chatbot")
    
demo.launch()
```

Looks like this:

image

Demo code 2 (this has been added in the `demo/` directory):

```py
import gradio as gr

def add_text(state, text):
    state = state + [text] + [text + "?"]
    return state, state

def add_image(state, image):
    state = state + [f"!()[/file={image}]"] + ["Cool pic!"]
    return state, state


with gr.Blocks() as demo:
    chatbot = gr.Chatbot()
    state = gr.State([])
    
    with gr.Row():
        with gr.Column(scale=0.85):
            btn = gr.UploadButton("🖼️", file_types=["image"])
            
    txt.submit(add_text, [state, txt], [state, chatbot])
    btn.upload(add_image, [state, btn], [state, chatbot])
            
demo.launch()
```


abidlabs:
> The demo notebooks don't match the run.py files. Please run this command from the root of the repo and then commit the changes:

Thanks @aliabd!

gradio/components.py
```diff
@@ -2857,11 +2857,13 @@ def process_single_file(f):
             )
             if self.type == "file":
                 if is_file:
-                    file = processing_utils.create_tmp_copy_of_file(file_name)
+                    file = processing_utils.create_tmp_copy_of_file(
```
abidlabs, line 2860: Fixed the `UploadButton` component to save files in the right location

aliabid94 (approved):
LGTM

abidlabs:
> Why markdown support over just supporting certain media discreetly? Does supporting all markdown cause any issues?

Because we'd need to add support for each of those individually and come up with an appropriate way to include them whereas Markdown already does that. I haven't seen any issues when I tried using Markdown, but are there specific things you're concerned about?

pngwn:
> A little bit yes, but everything is well-documented which is a big plus over having to figure out gradio-specific documentation

A path and a file extension could determine how we handle the media, markdown has to be learned too and isn't as ubiquitous and some people say.

> I haven't seen any issues when I tried using Markdown, but are there specific things you're concerned about?

Markdown supports pretty much all HTML which could easily break the rendering. Additionally, certain block level constructs could cause issues.

Additionally, we should also round the corners of images/ videos to ensure they match the style of the chatbox.

abidlabs:
> A path and a file extension could determine how we handle the media, markdown has to be learned too and isn't as ubiquitous and some people say.

But keep in mind that the `Chatbot` component takes in a set of strings, so within the string, we still way of designating that this particular string denotes a file as opposed to just a string that happens to include a path -- we need some sort of syntax, and Markdown offers an existing standard as opposed to us coming up with our own that we have to maintain and document.

> Markdown supports pretty much all HTML which could easily break the rendering. Additionally, certain block level constructs could cause issues.

Let me investigate this more

> Additionally, we should also round the corners of images/ videos to ensure they match the style of the chatbox.

Sure, I'll try to do this but may need some help on the css side!

abidlabs:
Discussed with @pngwn and we'll go with Markdown, but clearly document that only a subset of Markdown is supported. Also fixed the images to have a rounded border radius

abidlabs:
Here's how the rounded image looks: image

pngwn (approved):

abidlabs:
Thanks for the reviews @aliabid94 and @pngwn!

johnPertoft:
Cool change! But it broke my app unfortunately. Is it on the user now to parse the html/markdown output if we want to access the chat history or is there some option to do this?

abidlabs:
How did it break your app @johnPertoft? Can you open a new issue with more details and a code example so that we can reproduce?

johnPertoft:
It's really not a big deal but

In `gradio < 3.12` the history passed in looks like
```
[('hello!', " don't know")]
```

In `gradio >=3.12` the history passed in looks like
```
[('<p>hello!</p>\n', "<p>I don't know</p>\n")]
```

So the app "broke" in the sense that it now requires the user to remove the html/markdown added by this change if depending on the chat history. It's a very small change though so I'm not sure it warrants any action taken.

pngwn:
This feels like a bug to me. The history should contain the source you passed in, not the post-processed data. In this case it might be simple enough to work with the markup but some markdown transforms are very complex and will be impossible to do anything with other than appending to the history.

falcondai:
An ergonomic/QoL issue with the multimedia messages: unlike an `Image` or `Video` component, the preprocess method does not copy the inputed file into a private temp directory. I am using `TempFileManager` to copy the media files before passing them into a `Chatbot` component. Feels awkward. 

Happy to contribute a patch to preprocess in a manner similar to that of `Image`.

abidlabs:
Hi @falcondai good catch. Please do open a PR and I'm happy to review

Simon1V:
empty

With a jpg file called lion.jpg in the same directory, I only get this^^ 
Using the latest version of gradio

abidlabs:
What's your full code @Simon1V?

Simon1V:
> What's your full code @Simon1V?

The original one, just copy & pasted 
`import gradio as gr

with gr.Blocks(css="#chatbot .overflow-y-auto{height:700px!important}") as demo:
    gr.Chatbot([("hi", "hello **abubakar**"), ("", "cool pic")], elem_id="chatbot")
    
demo.launch()`

I just tried with another 350x350 jpg, also named lion.jpg to no avail
I got 
- main.py with that code 
- lion.jpg

in the same folder

abidlabs:
Now you should just be able to do this:

```py
import gradio as gr

with gr.Blocks(css="#chatbot .overflow-y-auto{height:700px!important}") as demo:
    gr.Chatbot([("hi", "hello **abubakar**"), ("('lion.jpg', )", "cool pic")], elem_id="chatbot")
    
demo.launch()
```

Simon1V:
Thank you, this worked : )

keranrong:
> Thank you, this worked : ) (Is there also a way to send images and text simultaneously? For multimodal input)

+1 And is there a way of make the image inline with the text please? (such as we send multiple images in one line)

semirke:
Hi there,

  is it possible to insert an image between the lines? (So like when the query engine responds with text containing markdown.)
For eg.: "As you can see on the image , all fields has to be filled."

 I couldn't make it work within the messages trying to link files, only with URL.

Thanks,
Semirke

Issue #2725, opened 2022-11-26
Returning `{}` from a function to update nothing raises an `Error`


CoffeeVampir3:
Hello, this request arose from the following example. I've currently got a Gradio UI which I'm loading configuration values of the last run back into:

```python
app_inputs = [prompt_textbox, negative_prompt_textbox, seed, in_parallel_slider, generation_runs_slider, width_slider, height_slider, num_steps_slider, cfg_slider]
app.load(ui_config.load_ui_config, inputs=None, outputs=app_inputs)
```
```python
def load_ui_config():
    dir_path = Path("./configs")
    file_name = "last_run.yaml"
    dest = (dir_path/file_name)

    if not dest.exists():
        return ("", "", 1, 1, 1, 768, 768, 20, 7)

    config = OmegaConf.load(dest)
    return list(config.values())
```
You can see, I'm forced to return some dubious default values. On the converse side, my proposal would be something like this:

```python
def load_ui_config():
   #...
    if not dest.exists():
        return None
 ```
 
The problem with the existing implementation is that in order to return valid defaults, you have to know your UI setup ahead of time. This is actually not necessary with the simple addition here, now instead of having to manually add or change new defaults every time we change components, we can simply do nothing and let the components handle their own defaults. I think this is probably a fairly common usage of load and would allow cleaner and more modular code.

Cheers,
Z.

pngwn:
Another issue with returning hard coded defaults is that it ought not be the desired behaviour. If a user has changed the values of certain components, either manually or through dispatching a previous event, those values will no longer be the defaults and the author may not want those values reset to default but rather left however they are. 

In terms of implementation, return `None` might be problematic right now as that could be a valid value for some components (need to validate @aliabid94 @abidlabs @freddyaboulton @aliabd). In v4 I think returning `None` should do nothing and all components should accept a dictionary instead. In the meantime we could add a `gr.None` helper that could be returned similar to `gr.Update()` if returning `None` isn't possible. 

We could even make the behaviour configurable from 'reset to defaults' to 'do nothing' (I don't think this is currently possible but could be mistaken).

abidlabs:
Hi @CoffeeVampir3 you can always return a dictionary of components in order to only update a subset of the output components. For example, if your function were to return `{prompt_textbox: "hi"}`, it would only update the `prompt_textbox` component. In this case, you can just return an empty dictionary in order to not update any of those components:

```py
def load_ui_config():
   #...
    if not dest.exists():
        return {}
[...]

PR #2735, opened 2022-11-28
Fix returning `{}` from a function to update nothing raises an Exception

abidlabs:
Quickly fixes: #2725.

Context: returning a `{}` from a function would previously raise an Exception. For some reason @aliabid94 you wrote an explicit check in the case that a function returned a dictionary update to check to see if the length of the dictionary was greater than 1 before converting it to a regular list update. I don't see why this check is needed, and removing it solves the bug. 

@aliabid94 can you review this to make sure I'm not missing something?


abidlabs:
Ahh actually I see why this check is included -- for some components, such as the `JSON` component, an empty dictionary has an existing meaning (namely, a null output)

abidlabs:
Gonna close this and suggest something else to the OP

PR #2736, opened 2022-11-28
Update Label color from backend

### LINKED ISSUE #2629: Improvements to Label for Dashboarding

As discussed today in our workshop, the gr.Number component does not work great for dashboarding uses where you want to display a number.

image

@abidlabs suggested using the Label component instead of a number and it works much better!

image

However there some improvements we can make (detailed below)

- Be able to set a color for the Label's Box. That way developers can use it to create "Alert" boxes for metrics.
Something like 

```python
return gr.Label.update(value=100, color="#FF0000")
```

**Additional context**

freddyaboulton:
# Description

Demo usecase

```python
import gradio as gr
import random

def update_color(value):
    if value == "bad":
        # This is bad so use red
        return "brown"
    elif value == "so-so":
        # Ok but pay attention (use orange)
        return "#ff9966"
    else:
        # Nothing to worry about
        return None


def update_value():
    choice = random.choice(['good', 'bad', 'so-so'])
    color = update_color(choice)
    return gr.Label.update(value=choice, color=color)

with gr.Blocks() as demo:
    label = gr.Label(value=-10)
    demo.load(lambda: update_value(), inputs=None, outputs=[label], every=1)
demo.queue().launch()
```

label_bg_color_update

Question about the api: Went with the suggestion to use a function to determine the label color. The one thing I don't like about it is that it couples the styling of the component with its value. We keep those separate for the other components. Curious what others think before going further. @abidlabs @pngwn @aliabid94 @aliabd 

Fixes #2629

# Checklist:

# A note about the CHANGELOG

### DIFF
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -3111,9 +3111,14 @@ class Label(Changeable, IOComponent, JSONSerializable):
         visible: Optional[bool] = None,
         color: Optional[str] = _Keywords.NO_VALUE,
     ):
-        # If color is None
+        # If color is not specified (NO_VALUE) map it to None so that
+        # it gets filtered out in postprocess. This will mean the color
+        # will not be updated in the front-end
         if color is _Keywords.NO_VALUE:
             color = None
+        # If the color was specified by the developer as None
+        # Map is so that the color is updated to be transparent,
+        # e.g. no background default state.
         elif color is None:
             color = "transparent"
         updated_config = {
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -3038,7 +3038,7 @@ class Label(Changeable, IOComponent, JSONSerializable):
         show_label: bool = True,
         visible: bool = True,
         elem_id: Optional[str] = None,
-        color: Optional[Callable[[Any], str]] = None,
+        color: Optional[Callable[[float | str], str]] = None,
         **kwargs,
     ):
         """
@@ -3049,6 +3049,7 @@ class Label(Changeable, IOComponent, JSONSerializable):
             show_label: if True, will display label.
             visible: If False, component will be hidden.
             elem_id: An optional string that is assigned as the id of this component in the HTML DOM. Can be used for targeting CSS styles.
+            color: A function that takes a value of the label (either a float or string) and maps it to a color (either a valid css color name or hexadecimal string). Whenever the value of the Label is updated, this function will be applied to the value to determine the color. This does not apply if the value of the `Label` includes confidences.
         """
         self.num_top_classes = num_top_classes
         self.color_fn = color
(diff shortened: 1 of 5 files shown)


abidlabs:
I find it a little simpler than the alternative (reproduced below), but I don't have a strong opinion and am fine with this too:

```py
import gradio as gr
import random

def update_color(value):
    if value < 0:
        # This is bad so use red
        return "#FF0000"
    elif 0 <= value <= 20:
        # Ok but pay attention (use orange)
        return "#ff9966"
    else:
        # Nothing to worry about
        return None

def get_update():
    val = random.randrange(-20, 60)
    return gr.update(value=val, color=update_color(val))

with gr.Blocks() as demo:
    label = gr.Label(value=-10)
    demo.load(get_update, inputs=None, outputs=[label], every=1)
demo.queue().launch()
```

ui/packages/label/src/Label.svelte
```diff
@@ -5,16 +5,28 @@
...
 
 <div class="output-label">
-	<div
-		class:sr-only={!show_label}
-		class="output-class font-bold text-2xl py-6 px-4 flex-grow flex items-center justify-center dark:text-slate-200"
-		class:no-confidence={!("confidences" in value)}
-	>
-		{value.label}
-	</div>
+	{#if color !== null}
```
freddyaboulton, line 12: What's the best practice to reduce this code dupe? The only difference is `style:background-color` but you can't place an {#if} around just the style

pngwn: You can create a conditional expression in the curly braces, falling back tot he 'default' which is probably transparent in this case (swap transparent for something else if that isn't the case).

```svelte
<div style:background-color={color || "transparent"}>
  ...
</div>

Comment (freddyaboulton) [reply ui/packages/label/src/Label.svelte]: Thank you for the tip @pngwn ! Works and makes it much cleaner!

**Comment by abidlabs** [comment]
Awesome PR @freddyaboulton! Specifying the color via a function which works well, and it's nice that `None` maps to the default background color.

One thing I'm having an issue with is passing a color function to `update()`. Here's what I tried, which doesn't seem to set the background color to red:

```py
def get_update():
    val = random.randrange(-20, 60)
    return gr.update(value=val, color=lambda x:"red")

with gr.Blocks() as demo:
    label = gr.Label(value=-10)
    demo.load(get_update, inputs=None, outputs=[label])

demo.queue().launch()
```

freddyaboulton:
Chatted with @abidlabs - will refactor so that `color` accepts a string as opposed to a function

gradio/components.py
```diff
@@ -3105,12 +3109,24 @@ def update(
...
         visible: Optional[bool] = None,
+        color: Optional[str] = _Keywords.NO_VALUE,
     ):
+        # If color is not specified (NO_VALUE) map it to None so that
+        # it gets filtered out in postprocess. This will mean the color
+        # will not be updated in the front-end
+        if color is _Keywords.NO_VALUE:
+            color = None
+        # If the color was specified by the developer as None
+        # Map is so that the color is updated to be transparent,
```
abidlabs, line 3120: Is there a typo here or am I misunderstanding?
```suggestion
        # Make it so that the color is updated to be transparent,
```

abidlabs (approved):
LGTM, thanks for getting it to work with `update()`. Works great now!

Issue #2335, opened 2022-09-26
Gradio fails to render thumbnails when image directory path is specified


neerajprad:
### Describe the bug

When I instantiate the Interface class with `examples={directory path}` (this is one of the options given in the documentation for the [Interface](https://gradio.app/docs/#interface-header) class) for an image classification model, the app fails to generate the thumbnail images for the examples. (See screenshot)

Clicking on the blank thumbnail still correctly loads the example. I checked that replacing this with a list to the absolute path for each of the example images works as expected. 

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Relevant commit: [99c4e70901c27da87e3313902128fb629e31278d](https://huggingface.co/spaces/neerajprad/leaf-classifier-demo/commit/99c4e70901c27da87e3313902128fb629e31278d)

### Screenshot

<img width="832" alt="Screen Shot 2022-09-25 at 4 41 35 PM" src="https://user-images.githubusercontent.com/1762463/192187911-14d1b40d-56af-4497-b5c3-a9b414e38dbd.png">

### Logs

```shell
N/A
```

### System Info

```shell
App running on Huggingface Spaces
Browser: Chrome
```

### Severity

annoying

freddyaboulton:
Hi @neerajprad ! Thanks for filing this. I think I know what's happening - the examples component is not loading the full path to the file. This is similar to https://github.com/gradio-app/gradio/issues/2129 but I it's an actually a different case.

neerajprad:
> This is similar to https://github.com/gradio-app/gradio/issues/2129 but I it's an actually a different case.

Ahh..I missed that. Feel free to close/merge the issues if you think the fix is the same for both. Thanks for taking a look.

evanlesmez:
I still see this with gradio 3.33.1

evanlesmez:
The problem seems to stem from using images from a path with a different parent directory

wywywywy:
> The problem seems to stem from using images from a path with a different path of parents

Sorry can you explain what you mean by this please? @evanlesmez

evanlesmez:
> > The problem seems to stem from using images from a path with a different path of parents
> 
> Sorry can you explain what you mean by this please? @evanlesmez

Sorry I should have been more clear and included more information at the time. I do remember seeing blank thumbnails. Not sure if it was because of relative path or not but I can see now that in a project with sample images I provided an absolute path for examples.

xiexh20:
[...]

Issue #2129, opened 2022-08-30
When specifying examples with a filepath + csv file, the front-end can't display the example images


freddyaboulton:
### Describe the bug

When specifying examples with a filepath + csv file, the front-end can't display the example images.

The reason is that in the csv file, the paths are relative to the `image_log` directory. 

You could specify a full path in the csv file but the implementation of `Examples` [assumes](https://github.com/gradio-app/gradio/blob/main/gradio/examples.py#L171) the paths are relative. Some discussion about this [here](https://github.com/gradio-app/gradio/pull/1952#issuecomment-1207092844)

Basically, it seems the log directory must be in the same directory as the demo, which is confusing and not documented as far as I can tell.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Make the following demos to the `image_mod` demo

```python
import gradio as gr
import os


def image_mod(image):
    return image.rotate(45)


demo = gr.Interface(image_mod, gr.Image(type="pil"), "image",
                    flagging_options=["blurry", "incorrect", "other"],
                    examples=os.path.join(os.path.dirname(__file__), "../../test/test_files/images_log"))

if __name__ == "__main__":
    demo.launch()
```

### Screenshot

![image](https://user-images.githubusercontent.com/41651716/187537665-4adecd30-7dcb-4a40-a54c-a01ad8245b82.png)

### Logs

```shell
raise RuntimeError(f"File at path {self.path} does not exist.")
RuntimeError: File at path /Users/freddy/sources/gradio/im/bus_copy.png does not exist.
```

### System Info

```shell
main
```

### Severity

serious, but I can work around it

sachit-menon:
Hi, any update on this? It is doable but a bit of a pain to have to list every item in the directory individually every time.

abidlabs:
This issue got buried @sachit-menon, but I'll take a look this week! Thanks for the reminder

sachit-menon:
Thanks!

PR #2745, opened 2022-11-30
Show image thumbnails when example directory is provided

abidlabs:
Previously image thumbnails would not show up if a directory of images was provided. Now, they should work just fine: 

* Fixes: #2129
* Fixes: #2335


freddyaboulton (approved):
Looks good to me @abidlabs !

abidlabs:
Appreciate the review, thanks @freddyaboulton!

PR #2748, opened 2022-11-30
Resize images for chatbot

abidlabs:
@dawoodkhan82 noticed that images are not resized in the chatbot, which can lead to a degraded experience for very large images. A quick css fix for that.

Before:

image

After:

image

Thanks for testing @dawoodkhan82!


freddyaboulton (approved):
Thanks for the fix @abidlabs @dawoodkhan82 !

Issue #340, opened 2021-11-08
Progress bar component


abidlabs:
@aliabid94 I think this is a great suggestion! It would be nice if people can provide their own status updates at different parts of the code. Perhaps this could be possible as you revamp the queuing mechanism?

### Discussed in https://github.com/gradio-app/gradio/discussions/339

<div type='discussions-op-text'>

<sup>Originally posted by **abnerLing** November  8, 2021</sup>
I have an app that takes a video input then extracts audio to run speech-to-text inference. I was wondering if there is a way to display the progress of the video upload? For example, using Hugging Face ([https://huggingface.co/spaces/abnerh/video-to-subs](https://huggingface.co/spaces/abnerh/video-to-subs)) or a shared link will just show a timer but I want the user to see how long the upload time takes.

Thanks</div>

xiaol:
Hi,  i use a image as a inference input, this would be very smooth for user experience.how can i help this to be done?

apolinario:
+1 for this. Now that we have iterative outputs, should be great to have a progress bar component, if it were integrated with tqdm automagically even better as so many implementations already use it to give progress bar estimation.

I think this could come together with this issue for doing the same thing but with the native progress bar embeded into the output components: https://github.com/gradio-app/gradio/issues/2683

fcakyon:
Any updates on the component? It would make things mush smoother :)

abidlabs:
We have a WIP in #2750, let us know if you have any thoughts on it! Aiming to merge in next week

Issue #2683, opened 2022-11-21
Allow for a tqdm and custom ETA as opposed to the native one in Gradio


apolinario:
- [x] I have searched to see if a similar issue already exists.
(I don't think it is the same as https://github.com/gradio-app/gradio/issues/1596, but if it is feel free to close)

**Is your feature request related to a problem? Please describe.**  
Gradio automatic calculation of progress and estimation is a great approximation based on previous runs. However, it doesn't work if: 
- It is the first run
- Different runs have different times (applications with user-defined settings)

`tqdm` already solves this problem by providing a per-iteration estimation and is widely used

**Describe the solution you'd like**  
I would like for users to have a possibility that, in output fields, the user could choose *not* to use the native estimated time, but rather choose to use the tqdm estimation from within whatever loop in their processing function they choose
![image](https://user-images.githubusercontent.com/788417/204126602-c2dee5e2-c090-4e7a-b1b8-88bc777074dd.png)

**Additional context**  
If there's more than one task inside of a function (multiple loops), there could be counters like: 
`task 1/3: iteration 2/10 [00:01<00:07,  1.12it/s]`

abidlabs:
This is possible with the new progress bar, thanks @aliabid94!

PR #2750, opened 2022-12-01
Add Progress Bar component

aliabid94:
Add support for progressbar component. Feel free to leave feedback on API or visual design.

See example in demo/progress/run.py

Snippet for tqdm-like wrapper
```python
def load_set(text, progress=gr.Progress()):
        imgs = [None] * 24
        for img in progress.tqdm(imgs, message="Loading from list"):
            time.sleep(0.1)
        return "done"
    load_set_btn.click(load_set, text, text2)
```

Snippet for explicit status:
```python
def clean_imgs(text, progress=gr.Progress()):
        progress(0.2, message="Collecting Images")
        time.sleep(1)
        progress(0.5, message="Cleaning Images")
        time.sleep(1.5)
        progress(0.8, message="Sending Images")
        time.sleep(1.5)
        return "done"
    clean_imgs_btn.click(clean_imgs, text, text2)
```

Recording 2022-11-30 at 18 01 51

Closes: #340


gradio/components.py
```diff
@@ -2915,6 +2915,51 @@ def style(self, *, full_width: Optional[bool] = None, **kwargs):
...
+        self,
+        value: float | None = 0,
+        *,
+        visible: bool = True,
+        elem_id: Optional[str] = None,
+        **kwargs,
+    ):
+        """
+        Parameters:
+            value: Default text for the button to display. If callable, the function will be called whenever the app loads to set the initial value of the component.
```
abidlabs, line 2938:
```suggestion
            value: value of the progress bar -- should be a `float` between 0 or 1. 
```

abidlabs:
Discussed further with @aliabid94 and we're rethinking this approach. Rather than `ProgressBar` being a separate component, it would make more sense to utilize the ETA status tracker that we have already implemented. The alternative approach would allow functions to `yield` a special type of "progress update" (syntax TBD) that would serve as a manual update of the ETA in the status bar. This would also be reflected in the progress bar that is built-in to each component. 

That being said, we've released `gradio==3.12.0b1` in case you want to use this stopgap progress bar @apolinario. (Also @apolinario would be interested in hearing your thoughts on these two options)

apolinario:
Some thoughts:
- I _quite_ like the idea of an independent progress bar. From a user perspective I think if both approaches existed could be cool. This is something the AUTOMATIC1111 would use as well as they already have a hacked version for e.g. - however I think it would be nice if it could also fit more info such as ETA, it/s

- Regardless if it is a progress bar or the output component (or both), I think it would be great to have `tqdm` automatic integration. So alternatively to the user doing their own yields. This is just a default way to do progress bar that is so simple and it contains so much information (it/s, progress %, ETA) that recreating it with yields would be a bit laborious

aliabid94:
Oh I really like the idea of tqdm integration, will look into that!

abidlabs:
+1 on the tqdm integration

Not sure about having two different ways of showing progress bars though. I prefer having a single way unless there is a strong reason to have a separate progress bar component. I know AUTOMATIC1111 GUI has a progress bar, but wouldn't it be better if the progress bar is incorporated into the output component itself, as is our current status tracker?

pngwn:
I'm a bit confused here. The `StatusTracker` _is_ a progress bar just the styling is different. If we want to change the styling then that is a separate issue.

I thought the requests we have had a few times is for a separate `ProgressBar`, like in the SD WebGUI, but I do think that it should be bound to an event and receive the same data as the built in status tracker, `currenttime / eta` with `queue position / queue size` all included, otherwise I don't really see the usefulness of this. Since everything happens on the server what would be timed if not the duration of an event? And if it is a bar rather than an infinite spinner, how would a user be able to provide an accurate estimated 100% value for the bar to 'complete'?

pngwn:
> I know AUTOMATIC1111 GUI has a progress bar, but wouldn't it be better if the progress bar is incorporated into the output component itself, as is our current status tracker?

This has really been my position all along but as above, the `StatusTracker` _is_ a progress bar.

abidlabs:
Yes so to summarize my takeaways:

* We don't create a separate `ProgressBar` component
* We allow our current StatusTracker to be able to set manually (or via tqdm) -- which would be valuable for the reasons @apolinario described in #2683
* When we support custom components, it wouldn't be hard for users to create a `ProgressBar` component, but we do not include this in core

How does that sound @aliabid94?

pngwn:
Thing is with tqdm is that this isn't really true:

> This is just a default way to do progress bar that is so simple and it contains so much information (it/s, progress %, ETA) that recreating it with yields would be a bit laborious

We would have to pass that information down to the component and render it in some way, so we'd need to support a vast array of inputs in the UI. We aren't just taking the default tqdm output and rendering it directly into the webpage somehow, we need to take the data is spits out and _do_ something with it. 

I'm strongly against dumping the stdout text into the page, i think it looks really bad and goes against the whole point of gradio. These UIs are for anyone, not just people familiar with typical CLI outputs. This needs designing properly before being included. cc @gary149 

If we are going to add this can we get some API + design proposals as well as breakdown of the various possible bits of UI that we would render based on the tqdm options the user passes?

abidlabs:
Not sure I follow @pngwn. We wouldn't display the raw progress bar that tqdm prints. We would extract the information and plug that into our existing status tracker. So the output would look pretty much the same.

pngwn:
But even dumping the text isn't great either, what works in a CLI tool isn't a great experience for end users. And that means we need to match the `tqdm` API if we want to support it fully.

apolinario:
@pngwn there's a TQDM Notebook class that afaik exposes some of the outputs for the programs to use, I don't really know how it works under the hood but could potentially be used for this: https://tqdm.github.io/docs/notebook/

Regarding the progress bar. I still see how it could be interesting for users to choose whether their output is in a progress bar or on the output field. I think there's room for both, but I agree this may make sense as a custom/user component. But just FYI if I had to choose between a custom progress bar or a native one on the output field for this particular demo I would choose a custom progress bar

pngwn:
The big issue with a custom component ProgressBar is that we currently don't have a mechanism to pass anything but the output components the current progress of a request/ queue request. We'll need to build in more flexibility to be able to do that.

aliabid94:
To summarize:
- We already have an automatic status tracker that uses average of past ETAs to predict ETAs. 
- we want users to able to give more accurate progress updates through custom logic. This should also be able to take the form of % done rather than purely eta
- we should make the custom progress bar visually consistent with the current progress bar
- would be nice to have tqdm-like wrappers around loops

Here's a possible implementation:
```python
import gradio as gr
import img_model

with gr.Blocks() as demo:
     prompt = gr.Textbox(label="Enter text")
     img = gr.Image()

     def load_image(data):
          total_steps = 50
          for step in img_model.load():
               # ...
               yield gr.Status(step / total_steps, text="Loading model")
          return img_model(data[prompt])

     prompt.submit(load_image, {prompt}, img)

demo.queue().launch()
```

which would look something like:

Screen Shot 2022-12-05 at 2 04 48 PM

For a tqdm-like wrapper, we could have:

```python
import gradio as gr
import img_model

with gr.Blocks() as demo:
     prompt = gr.Textbox(label="Enter text")
     img = gr.Image()

     def load_image(data):
          total_steps = 50
          for _ in gr.Track(img_model.load()):
               # ...
               yield gr.Status(text="Loading model")
          return img_model(data[prompt])

     prompt.submit(load_image, {prompt}, img)

demo.queue().launch()
```

where the `gr.Track` can automatically keep track of length and progress, and yielding a gr.Status() will automatically get that information.

abidlabs:
For the tqdm integration, it would be better if users can somehow return or yield the tqdm object itself so that people don’t have to learn a new syntax. Users are already using tqdm in their code so it would be better if we integrated with tqdm instead of offering a replacement

aliabid94:
Yeah I thought about that but in the syntax 
```python
for obj in tqdm(objects):
   yield ...
```
there isn't really access to the tqdm object within the loop, so not sure how to yield it

aliabid94:
the other reason not to use tqdm is that we would then have the users printing to the console as a result of using tqdm. If multiple users are being processed at once, the output will get ugly

abidlabs:
> there isn't really access to the tqdm object within the loop, so not sure how to yield it

So `tqdm(iterable)` returns a new iterable, and this contains all of the info you'd need. Here's a complete example:

```py
from tqdm import tqdm

# Create a list of numbers to iterate over
numbers = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]

# Create a tqdm object to track the progress of the iteration
pbar = tqdm(numbers)

# Iterate over the list of numbers
for number in pbar:
    time.sleep(0.5)
    # Print the current status of the tqdm object
    print("current_iteration", pbar.n)
    print("out of", pbar.total)
```

abidlabs:
Question about the `gr.Status` implementation, would this replace `gr.ProgressBar`?

apolinario:
> The main thing is that people are already using tqdm and so forcing them to switch to something else so that their function works with Gradio is not ideal (cc @apolinario for your thoughts too)

I agree. Also there's quite a bit of people that re-use the same code for Colab and Gradio demos, so if tqdm worked for both that would be great. 

>  the other reason not to use tqdm is that we would then have the users printing to the console as a result of using tqdm. If multiple users are being processed at once, the output will get ugly

FYI this already happens on most Spaces that have generative media that I come across/build. However most of them have queues (otherwise they OOM) so the logs don't get that messy/overlappy and it is usually good to be able to track usage

apolinario:
Another point that makes me thing how integrating with tqdm would be great: 
I would like to integrate the `snapshot_download` function from `huggingface_hub` with a Gradio progressbar. Without native integration, I would need to fork huggingface_hub just to inject Gradio integration (or capture the terminal output and throw some yields), but if Gradio can bind to already existing tqdm from other functions that would make things easier

abidlabs:
Very cool @aliabid94. Going through this PR bit by bit and leaving my comments here: 

* First, on the user-facing API, I prefer the original syntax you proposed in the PR: `progress.tqdm()` over `progress.track()`. People are already familiar with `tqdm.tqdm()` syntax and will expect it to return the iterable just like `tqdm.tqdm()`

guides/01_getting_started/02_key_features.md
```diff
@@ -182,6 +182,15 @@ Note that we've added a `time.sleep(1)` in the iterator to create an artificial
 
 Supplying a generator into Gradio **requires** you to enable queuing in the underlying Interface or Blocks (see the queuing section above).
 
+## Progress Updates
```
abidlabs, line 185:
```suggestion
## Progress Bars
```

guides/01_getting_started/02_key_features.md
```diff
@@ -182,6 +182,15 @@ Note that we've added a `time.sleep(1)` in the iterator to create an artificial
...
+
+Ypu can explicitly update the progress of a function by adding an argument to your method that has a default value of `gradio.Progress()`. Then you can update the progress levels by calling this argument directly, or using the `track()` method of `Progress` to track progress over an iterable. Currently, queueing must be enabled for progress updates.
```
abidlabs, line 187:
```suggestion
Gradio supports the ability to create a custom Progress Bars to have customizability and control over the progress update that you show to the user. In order to enable this, simply add an argument to your method that has a default value of a `gradio.Progress` instance. Then you can update the progress levels by calling this argument directly with a float between 0 and 1, or using the `tqdm()` method of the `Progress` instance to track progress over an iterable. Queueing must be enabled for progress updates.
```

guides/01_getting_started/02_key_features.md
```diff
@@ -182,6 +182,15 @@ Note that we've added a `time.sleep(1)` in the iterator to create an artificial
...
+
+$code_progress_simple
+$demo_progress_simple
+
+You can even make automatic progress updates from any `tqdm.tqdm` by setting the default argument as  `gr.Progress(track_tqdm=True)`.
```
abidlabs, line 192:
```suggestion
If you use the `tqdm` library, you can even report progress updates automatically from any `tqdm.tqdm` that already exists within your function by setting the default argument as  `gr.Progress(track_tqdm=True)`!
```

demo/progress_simple/run.py
```diff
@@ -0,0 +1,16 @@
+import gradio as gr
+import time
+
+def reverse(word, progress=gr.Progress()):
```
abidlabs, line 4:
```suggestion
def slowly_reverse(word, progress=gr.Progress()):
```

demo/progress_simple/run.py
```diff
@@ -0,0 +1,16 @@
...
+    progress(0, desc="Starting")
+    time.sleep(1)
+    new_string = ""
+    for letter in progress.track(word, desc="Reversing"):
+        time.sleep(0.25)
+        new_string = letter + new_string
+    return new_string
+
+demo = gr.Interface(reverse, gr.Text(), gr.Text())
```
abidlabs, line 13:
```suggestion
demo = gr.Interface(slowly_reverse, gr.Text(), gr.Text())
```

demo/progress_simple/run.py
```diff
@@ -0,0 +1,16 @@
+import gradio as gr
+import time
+
+def reverse(word, progress=gr.Progress()):
+    progress(0, desc="Starting")
+    time.sleep(1)
+    new_string = ""
+    for letter in progress.track(word, desc="Reversing"):
```
abidlabs, line 8:
```suggestion
    for letter in progress.tqdm(word, desc="Reversing"):
```

CHANGELOG.md
```diff
@@ -1,7 +1,23 @@
...
+Send custom progress updates by adding a `gr.Progress` argument after the input arguments to any function. Example:
+
+```python
+def clean_imgs(text, progress=gr.Progress()):
+    progress(0.2, message="Collecting Images")
+    time.sleep(1.5)
+    progress(0.8, message="Sending Images")
+    time.sleep(1.5)
+    return "done"
+clean_imgs_btn.click(clean_imgs, text, text2)
```
abidlabs, line 16: I'd give a full example with `gr.Interface` here

website/homepage/src/docs/template.html
```diff
@@ -10,268 +10,295 @@
...
-        <section class="pt-2">
-          {% for component in docs["component-helpers"] %}
+        <section id="helpers" class="pt-2 flex flex-col gap-10 mb-8">
+          <div>
+            <h2 id="helpers-header"
+                class="text-4xl font-light mb-2 pt-2 text-orange-500">
+              Helpers
+            </h2>
+            <p class="mt-8 text-lg">
+              Gradio includes helper classes that abstract over existing components. The goal of these classes is to help you
```
abidlabs, line 273:
```suggestion
              Gradio includes helper classes and methods that interact with existing components. The goal of these classes and methods is to help you
```

website/homepage/src/docs/template.html
```diff
@@ -10,268 +10,295 @@
...
+              add common functionality to your app without having to repeatedly create the same components and event listeners.
```
abidlabs, line 274:
```suggestion
              add common functionality to your app without having to rewrite common functions.
```

guides/01_getting_started/02_key_features.md
```diff
@@ -182,6 +182,15 @@ Note that we've added a `time.sleep(1)` in the iterator to create an artificial
 
 Supplying a generator into Gradio **requires** you to enable queuing in the underlying Interface or Blocks (see the queuing section above).
 
+## Progress Updates
+
+Ypu can explicitly update the progress of a function by adding an argument to your method that has a default value of `gradio.Progress()`. Then you can update the progress levels by calling this argument directly, or using the `track()` method of `Progress` to track progress over an iterable. Currently, queueing must be enabled for progress updates.
+
+$code_progress_simple
```
abidlabs, line 189: This example is great! I would include another example that shows how to use with directly with a float as well, just because I think that's a common use case that people might not really understand

abidlabs:
When run without queuing enabled, I get the following:

```
AttributeError: 'Interface' object has no attribute '_queue'
``

It's pretty self-evident, but may want catch this and tell users to enable queueing

**Comment by abidlabs** [comment]
When I run this:

```py
import gradio as gr
import time

progress = gr.Progress()

def test(x, progress=progress):
    for i in range(10):
        time.sleep(0.3)
        progress(i/10)
    return x

gr.Interface(test, "number", "number").queue().launch()
```

it works great, but what is the "0 steps" that it shows on the top right?

image

gradio/helpers.py
```diff
@@ -0,0 +1,780 @@
...
+
+def create_tracker(block_parent, event_id, fn, track_tqdm):
+
+    progress = Progress(
+        _active=True, _callback=block_parent._queue.set_progress, _event_id=event_id
+    )
+    if not track_tqdm:
+        return progress, fn
+
+    _tqdm = __import__("tqdm")
```
abidlabs, line 513: What if `tqdm` is not installed in the system?

demo/progress/run.py
```diff
@@ -0,0 +1,97 @@
+import gradio as gr
+import random
+import time
+import tqdm
```
abidlabs, line 4: This demo needs a requirements file for `tqdm` and `dataset`

gradio/blocks.py
```diff
@@ -831,6 +762,7 @@ async def call_function(
         processed_input: List[Any],
         iterator: Iterator[Any] | None = None,
         request: routes.Request | List[routes.Request] | None = None,
+        event_id: str | None = None,
     ):
         """Calls and times function with given index and preprocessed input."""
```
abidlabs, line 767: Even though this is an internal function, it would be super helpful if we could add docstring to explain the parameters. Would you be able to add?

gradio/helpers.py
```diff
@@ -0,0 +1,780 @@
+"""
```
abidlabs, line 1: Nice to have all of the component helpers in one place.

Do you think we should add `gr.Request` and `mount_gradio_app()` here as well? I don't think a separate "Routes" section makes sense

gradio/helpers.py
```diff
@@ -0,0 +1,780 @@
...
+Defines helper methods useful for loading and caching Interface examples.
```
abidlabs, line 2: This comment isn't correct

gradio/documentation.py
```diff
@@ -58,8 +58,8 @@ def document_fn(fn: Callable) -> Tuple[str, List[Dict], Dict, Optional[str]]:
             if mode == "description":
                 description.append(line if line.strip() else "<br>")
                 continue
-            assert line.startswith(
-                "    "
+            assert (
+                line.startswith("    ") or line.strip() == ""
```
abidlabs, line 62: Can you explain this change?

gradio/helpers.py
```diff
@@ -0,0 +1,780 @@
...
+        self.unit = unit
+        self._tqdm = _tqdm
+        self.progress = progress
+
+
+@document("__call__", "track")
+class Progress(Iterable):
+    """
+    The Progress class provides a custom progress tracker that is used in a function signature.
+    To attach a Progress tracker to a function, simply add a parameter right after the input parameters that has a default value set to `gradio.Progress()`.
```
abidlabs, line 358:
```suggestion
    To attach a Progress tracker to a function, simply add a parameter right after the input parameters that has a default value set to a `gradio.Progress()` instance.
```

gradio/queue.py
```diff
@@ -24,18 +26,34 @@ class Estimation(BaseModel):
...
+    index: Optional[int]
+    length: Optional[int]
+    unit: Optional[str]
+    progress: Optional[float]
+    desc: Optional[str]
+
+
+class Progress(BaseModel):
+    msg: str = "progress"
+    progress_data: List[ProgressUnit] = []
```
abidlabs, line 39: Can we move these to `dataclasses.py` file?

gradio/queue.py
```diff
@@ -24,18 +26,34 @@ class Estimation(BaseModel):
     queue_eta: int
 
 
+class ProgressUnit(BaseModel):
+    index: Optional[int]
+    length: Optional[int]
+    unit: Optional[str]
+    progress: Optional[float]
```
abidlabs, line 33: Should allow `Tuple[int, int]` as well I believe. This throws an error:
```py
import gradio as gr
import time

def reverse(word, progress=gr.Progress()):
    for letter in word:
        progress((1, 3))
    return word

demo = gr.Interface(reverse, gr.Text(), gr.Text())
demo.queue(concurrency_count=4).launch()
```

test/test_blocks.py
```diff
@@ -1043,6 +1043,90 @@ def generation():
         assert "At step 9" not in captured.out
 
 
+class TestProgressBar:
+    @pytest.mark.asyncio
+    async def test_progress_bar(self):
+        from tqdm import tqdm
```
abidlabs, line 1049: Let's add `tqdm` to `requirements.in`

abidlabs: Nice test. I would also add a unit test to make sure the regular `gr.Progress()` with `track_tqdm=False` also works

gradio/routes.py
```diff
@@ -312,6 +312,10 @@ async def run_predict(
             else:
                 session_state = {}
                 iterators = {}
+            if hasattr(body, "event_id"):
+                event_id = body.event_id
+            else:
+                event_id = None
```
abidlabs, line 318:
```suggestion
             event_id = getattr(body, "event_id", None)
```

gradio/helpers.py
```diff
@@ -0,0 +1,780 @@
...
+        desc: str | None = None,
+        total: float | None = None,
+        unit: str = "steps",
+        _tqdm=None,
+    ):
+        """
+        Updates progress tracker with progress and message text.
+        Parameters:
+            progress: If float, should be between 0 and 1 representing completion. If Tuple, first number represents steps completed, and second value represents total steps or None if unknown. If None, hides progress bar.
+            desc: description to display.
```
abidlabs, line 433: Add docstring for `total` and `unit` as well

gradio/helpers.py
```diff
@@ -0,0 +1,780 @@
...
+                    break
+            self._callback(
+                event_id=self._event_id,
+                iterables=self.iterables,
+            )
+        else:
+            return
+
+
+def create_tracker(block_parent, event_id, fn, track_tqdm):
```
abidlabs, line 505: Consider renaming `block_parent` as `root_blocks`. `block_parent` sounds like the immediate parent and we use `root_blocks` in other parts of the codebase

gradio/helpers.py
```diff
@@ -0,0 +1,780 @@
...
+        unit: str = "steps",
+        _tqdm=None,
+        *args,
+        **kwargs,
+    ):
+        """
+        Attaches progress tracker to iterable.
+        Parameters:
+            iterable: iterable to attach progress tracker to.
+            desc: description to display.
```
abidlabs, line 458: Add docstring for other parameters

gradio/helpers.py
```diff
@@ -0,0 +1,780 @@
...
+
+        """
+        if iterable is None:
+            new_iterable = TrackedIterable(None, 0, total, desc, unit, _tqdm)
+            self.iterables.append(new_iterable)
+            self._callback(event_id=self._event_id, iterables=self.iterables)
+            return
+        length = len(iterable) if hasattr(iterable, "__len__") else None
```
abidlabs, line 466:
```suggestion
        length = getattr(iterable, "__len__", None)
```

gradio/queue.py
```diff
@@ -65,12 +83,15 @@ def __init__(
         self.queue_duration = 1
         self.live_updates = live_updates
         self.sleep_when_free = 0.05
+        self.progress_update_sleep_when_free = 0.2
```
abidlabs, line 86: This sets the floor on how frequently the progress bar can be udated, right? Maybe 0.1 seconds is good?

gradio/helpers.py
```diff
@@ -0,0 +1,780 @@
...
+
+    def close_tqdm(self):
+        if self._progress is not None:
+            self._progress.close(self)
+        return self.__close__orig__()
+
+    def exit_tqdm(self, exc_type, exc_value, traceback):
+        if self._progress is not None:
+            self._progress.close(self)
+        return self.__exit__orig__(exc_type, exc_value, traceback)
```
abidlabs, line 546: This is brilliant! `tqdm.tqdm` continues to work even afterwards so that's really nice

abidlabs:
Awesome PR @aliabid94! I left a lot of feedback since it's massive PR, but they're mostly pretty minor things. Overall, works great.

gradio/blocks.py
```diff
@@ -843,18 +775,30 @@ async def call_function(
...
         )
 
         start = time.time()
 
         if iterator is None:  # If not a generator function that has already run
-            if inspect.iscoroutinefunction(block_fn.fn):
-                prediction = await block_fn.fn(*processed_input)
+            if progress_tracker is not None:
+                progress_tracker, fn = create_tracker(
+                    self, event_id, block_fn.fn, progress_tracker.track_tqdm
```
freddyaboulton, line 791: I think it will be cleaner if the `event_id` is not passed to the Blocks instance and instead the progress tracker is created in the route, similar to how `Requests` are created and then passed to the function

freddyaboulton (commented):
Awesome PR @aliabid94 ! I really like the demo you created + the documentation. 

Left some comments here. I agree with @abidlabs 's comments as well.

website/homepage/src/docs/obj_doc_template.html
```diff
@@ -110,7 +110,7 @@ <h4 class="mt-4 p-3 font-semibold">Example Usage</h4>
 
     {% elif obj["demos"] %}
 
-    <div class="container relative">
+    {# <div class="container relative">
```
aliabd, line 113: Why did you comment this out?

website/homepage/src/docs/template.html
```diff
@@ -10,268 +10,295 @@
   <body>
     <div class="container mx-auto flex flex-wrap justify-between flex-row relative items-center px-4 py-5 gap-6 text-lg z-50">
       <div class="flex flex-row items-center">
-      <a href="/">
-        <img src="/assets/gradio.svg">
-      </a>
-
-      <div class="pl-2 pr-1 py-1 w-fit text-gray-600 font-semibold">
-        <select id="toggle" onchange="this.options[this.selectedIndex].value && (window.location = this.options[this.selectedIndex].value);" class="form-select appearance-none block px-3 text-base font-normal text-gray-700 bg-white bg-clip-padding bg-no-repeat border border-solid border-gray-300 rounded-full transition ease-in-out m-0 focus:text-gray-700 focus:bg-white focus:border-blue-600 focus:outline-none w-max" style="width: 135%; padding-top: 0.125rem; padding-bottom: 0.125rem">
+        <a href="/">
```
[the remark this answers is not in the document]
aliabd: Are you sure you're viewing main on the docs? you'll need to switch the toggle from 3.15.0 to see the changes. Nothing in this file was changed, just formatting. (@aliabid94 correct me if i'm wrong)

aliabd (changes requested):

demo/hello_world/run.py
```diff
@@ -6,4 +6,4 @@ def greet(name):
 demo = gr.Interface(fn=greet, inputs="text", outputs="text")
     
 if __name__ == "__main__":
-    demo.launch()   
\ No newline at end of file
+    demo.queue().launch()
```
aliabd, line 9: fyi @aliabid94 this is why the notebook diff action is failing

gradio/helpers.py
aliabid94: now caught in try catch

gradio/helpers.py
aliabid94: PR is already huge, can do cleanup in another pr

gradio/documentation.py
aliabid94: ^ correct, to allow empty lines in docstring, used in Example section where example code may have empty line

gradio/queue.py
aliabid94: set to 0.1 for now.

gradio/blocks.py
```diff
@@ -1647,8 +1596,12 @@ def attach_load_events(self):
 
     def startup_events(self):
         """Events that should be run when the app containing this block starts up."""
+
         if self.enable_queue:
-            utils.run_coro_in_background(self._queue.start)
+            progress_tracking = any(
```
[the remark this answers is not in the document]
aliabid94: fixed, good catch

gradio/helpers.py
```diff
@@ -0,0 +1,780 @@
...
+        if self._active:
+            self._callback(
+                event_id=self._event_id,
+                iterables=self.iterables
+                + [TrackedIterable(None, 0, total, desc, unit, _tqdm, progress)],
+            )
+        else:
+            return progress
+
+    def track(
```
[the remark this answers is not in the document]
aliabid94: zip returns an iterable with no fixed length. Therefore we only print steps so far in the top right, no progress bar since we don't know full length

gradio/queue.py
```diff
@@ -132,6 +153,51 @@ async def start_processing(self) -> None:
...
+    async def start_progress_tracking(self) -> None:
+        while not self.stopped:
+            if not any(self.active_jobs):
+                await asyncio.sleep(self.progress_update_sleep_when_free)
+                continue
+
+            for job in self.active_jobs:
+                if job is None:
+                    continue
+                for event in job:
```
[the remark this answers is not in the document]
aliabid94: worked for me? have made code changes since, so might have resolved itself.

website/homepage/src/docs/template.html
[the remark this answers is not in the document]
aliabid94: ^ yes, change docs to main at the top

website/homepage/src/docs/obj_doc_template.html
```diff
@@ -63,9 +63,9 @@
   {% if obj['override_signature'] %}
   <div class="codeblock"><pre><code class="lang-python">{{ obj['override_signature'] }}</code></pre></div>
   {% else %}
-  <div class="codeblock"><pre><code class="lang-python">{{ parent }}.<span>{{ obj["name"] }}(</span><!--
+  <div class="codeblock"><pre><code class="lang-python">{{ parent }}{% if obj["name"] != "__call__" %}.<span>{{ obj["name"] }}{% endif %}(</span><!--
 -->{% for param in obj["parameters"] %}<!--
-  -->{% if "kwargs" not in param and "default" not in param and param["name"] != "self" %}<!--
+  -->{% if "positional" in param["kind"] and param["name"] != "self" %}<!--
```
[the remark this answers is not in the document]
aliabid94: reverted

website/homepage/src/docs/obj_doc_template.html
aliabid94: oh I just find it a waste of space. will uncomment, we can redesign later

aliabid94:
I think I addressed every concern @abidlabs @freddyaboulton @aliabd, thanks for the feedback so far. if everything looks good to you guys, I'll merge in tomorrow

aliabd (approved):

Issue #2714, opened 2022-11-23
Add and document `queue` parameter in `blocks.load()`


apolinario:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
If a job is running, the gr.Load() adds its tasks to the queue, kind of ruining the goal of running code at startup.

**Describe the solution you'd like**  
I would like to be able to do a `queue=False` for `blocks.load()` while still leave the queue active globally for the app, in the same way I can do it for `.click()` or `.update()` functions

pngwn:
You can set `queue=False` on a per event basis:

```py
blocks.load(fn, inputs, outputs, queue=False)
```

pngwn:
I dont understand why this isn't documented though. https://www.gradio.app/docs/#blocks cc @aliabd

aliabd:
The queue parameter in `blocks.load()` isn't documented because it's not defined as a parameter, but looks like it works through the kwargs. The docs pull from the docstrings and [it isn't there](https://github.com/gradio-app/gradio/blob/78796f6efe6eafe77d93eaf0dd473829d9a0804f/gradio/blocks.py#L1091).

The `blocks.queue()` method isn't documented because it the `@document` decorator  above `Blocks` [only has `load` in it](https://github.com/gradio-app/gradio/blob/78796f6efe6eafe77d93eaf0dd473829d9a0804f/gradio/blocks.py#L464). As opposed to `Interface` [which has](https://github.com/gradio-app/gradio/blob/78796f6efe6eafe77d93eaf0dd473829d9a0804f/gradio/interface.py#LL44-L44C2): `launch`, `load`, `from_pipeline`, `integrate`, and `queue`. Not really sure why we made that call. 

Would be easy fix to document the `queue` method (and also the others?), but for the parameter in load we'd probably have to define it outside the kwargs.

apolinario:
I don't think `queue=False` in the `load` component is working. I set it to False and it still shows me queue

pngwn:
Do you have a simple repro that we can look at?

Issue #2754, opened 2022-12-01
Refactor `gr.Blocks.load()` in 4.0


abidlabs:
`gr.Blocks.load()` is both a class method and an instance method, which is confusing. We should refactor this in 4.0. 

Here's a proposal:
* Deprecate `gr.Blocks.load()` and `gr.Interface.load()`
* Instead of the class method, define a `gr.load()` that can be used to load any Space whether it be Blocks/Interface, or a model
* For the event trigger, define an instance method `Blocks.page_load()`

aliabd:
+1 to this, I was confused myself.

freddyaboulton:
This is already done on `v4` but `Blocks.load` is now an instance method and only defines load events. No `page_load` was added.

PR #2755, opened 2022-12-01
Adds missing parameters to `Blocks.load()` 

abidlabs:
Adds `queue` and other missing parameters to `Blocks.load()`, as well as to the documentation.

Fixes: #2714 

To test, run

```py
import gradio as gr
import time

def identity():
    time.sleep(5)
    return "hello"

with gr.Blocks() as demo:
    textbox = gr.Textbox()
    demo.load(identity, None, textbox, queue=False)
    
demo.queue().launch()
```

and open on multiple tabs. Since `queue` is disabled, you should not see a queue on the subsequent tabs and instead they should all run concurrently.

As a side note, we should refactor `Blocks.load()` altogether, as I suggest in #2754


gradio/blocks.py
```diff
@@ -1093,13 +1093,21 @@ def load(
...
+        postprocess: bool = True,
+        every: float | None = None,
+        _js: Optional[str] = None,
         *,
         name: Optional[str] = None,
         src: Optional[str] = None,
         api_key: Optional[str] = None,
         alias: Optional[str] = None,
-        _js: Optional[str] = None,
-        every: None | int = None,
```
abidlabs, line 1102: No problem bringing a parameter after "*" to before "*" (the other way around would cause backwards incompatibility)

aliabd: Looks like gh markdown messed up what you're trying to say? I don't understand this

abidlabs: Whoops I meant to say that it's safe to make the `_js` and `every` positional parameters instead of just named parameters. Wouldn't have been safe the other way around

aliabd (approved):
LGTM, but maybe we should make same changes to `Interface.load()`?

abidlabs:
`Interface.load()` is only the class method, it doesn't have the event trigger associated with it (although Interface().load() inherits from Blocks, so you can use it like that)

Issue #2732, opened 2022-11-28
`gr.File` transfers the same file to a different temp folder every time it is executed


apolinario:
### Describe the bug

(Not sure if a bug, feel free to re-tag this is a feature request) 

I have a `gr.File` component that should display a 5GB file, this component is constantly mounted/unmounted as the application is loaded or refreshed 

Everytime the component is mounted the file is transferred to a new temporary folder/file like `/tmp/tmpyesjr1q0/diffusers_model4x4bg8ew.tar`

So not rarely I have this issue
```
OSError: [Errno 28] No space left on device: 'diffusers_model.tar' -> '/tmp/tmpyesjr1q0/diffusers_model4x4bg8ew.tar'
```
And I have to manually go and clean up the `/tmp/` folder

I'm not sure how best to solve this issue, however I would suggest a workaround that actually I wish it existed:
- Have a prop to serve the file from where it currently lives (without a need transfer it to `tmp`). Specifically for Spaces that have a file structure of `https://<space-url>/file/the-file-I-want.tar` if I could just show the real path on the `gr.File` component instead of `tmp` shenanigans

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

No applicable 

### Logs

```shell
No logs
```

### System Info

```shell
Gradio == 3.11
```

### Severity

serious, but I can work around it

abidlabs:
Hey @apolinario thanks for creating the issue. Will take a look this week!

abidlabs:
@apolinario are you using the `gr.File()` to display the output of a function, or to only display a fixed file? It would be great if you can share more code so that I can understand how you are using it

apolinario:
I'm using it to display a fixed file that is created by a function. Sorry the code is a bit convoluted but it is the Dreambooth Training Space: https://huggingface.co/spaces/multimodalart/dreambooth-training

I have a `gr.File()` that showcases the generated tar file of the model & the converted ckpt. They can be outputs of the function but not always. And they regardless live in the folder structure anyway, so copying it to a temp folder sounds a bit of an extra- step

Issue #2660, opened 2022-11-17
Disable saving files in /tmp


MrNocTV:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
Currently we've just migrated Gradio from 3.1.1 to 3.4 to utilize Websocket feature.
We deploy Gradio on Kubernetes.
Each application is deployed on multiple Kubernetes pods.

We test the Websocket and it works just fine.
This is the content of `app.py`
We have problems with examples due to new way of fetching files (using /tmp/...)
```python
import gradio as gr
import os
import mediapy

import cv2
# import mmcv
import numpy as np
import torch

# from mmedit.apis import init_model, restoration_video_inference
# from mmedit.core import tensor2img
# from utils.app_utils import cpu_model_config
from utils.app_utils import preproces_data

import base64
import json
import requests

import time
import subprocess

VIDEO_EXTENSIONS = ('.mp4', '.mov')
FFMPEG_PATH = subprocess.check_output(["which","ffmpeg"], universal_newlines=True)[:-1]

mediapy.set_ffmpeg(FFMPEG_PATH)

api_url = 'http:///-ml-model-dev/vsr-gpu-101-test/vsr'
cpu_mode = False

def video_identity(video):
    print(video)
    return video

def video_sr(video, strong, max_video_size, max_process_length):
    print(video)
    file_extension = os.path.splitext(video)[1]    
    if file_extension not in VIDEO_EXTENSIONS:  # save as video
        raise ValueError("The file type should be one of {} but it is {}".format(VIDEO_EXTENSIONS, file_extension))

    start_infer_time = time.time()
    print(time.ctime())    
    if cpu_mode:
        """
        config = cpu_model_config(video, max_video_size, max_process_length)
        # generate RealBasicVSR
        output, fps = restoration_video_inference(model, video,
                                                  config["window_size"], config["start_idx"],
                                                  config["filename_tmpl"], config["max_seq_len"], 
                                                  config["rescale_ratio"], 
                                                  (config["too_long"], config["max_process_length"]), strong)

        print("infer_time_time : {}".format(time.time() - start_infer_time))
        img_array_realvsr = []
        img_array_output = []
        img_array_bicubic = []

        if strong:
            for i in range(0, output.size(1)):
                img = tensor2img(output[:, i, :, :, :])
                img = cv2.cvtColor(img, cv2.COLOR_BGR2RGB)
                img = img.astype(np.uint8)
[...]

PR #2758, opened 2022-12-02
Create fewer temp files and make them consistently-named

abidlabs:
Fixes: #2732 
Fixes: #2660

Recently, we've had a number of issues around the creation of temporary files:

* In #2732, @apolinario reported that a large number of identical temporary video files are being created, eating up memory
* In #2660, @MrNocTV pointed out that because we generate temporary video files for all example videos with different paths when the Gradio app is launched, this means such gradio apps won't be compatible with Kubernetes where the temporary file might be created on one replica but the user is accessing the app from another replica

It's a waste of resources (time, disk space) to create so many temporary files. I've gone back and forth between different ideas on how to fix this, and here's what I think we should do: **add a flag to remove the creation of temporary files in .postprocess()**

Let me elaborate:

The reason we create temporary files in the first place is because (A) we want to avoid serialization/deserialization of files, which is slow (B) we don't want to modify a user-supplied file directly, because we don't want to change any files on their file system. This means that we *must* create temporary files in `.preprocess()` before the prediction function runs, in case the prediction function modifies the input file. 

However, we may not *need* to create temporary copies of the output file in `.postprocess()` -- we can simply return it directly. Now, it turns out `.postprocess()` is actually called from 3 different places:

1. It's called when a user simply displays a static file, like this: `gr.Video("test.mp4")` 
2. It's called when a user supplies an example file, `gr.Interface(..., examples=["test.mp4"])`
3. It's called when a user returns a file from their prediction function

Now, for (1) and (2), we currently always create a temporary copy of the file. This is sometimes necessary, for example if the user has configured the `Video` component to have a particular format, such as `.mp4`, but the video file supplied is of a different format. But in other cases, it is not necessary and returning the original file has no downsides at all. 

Now for (3), there is actually a benefit to making a copy of the output file before returning it -- which is that if your function returns a fixed video filepath, and multiple people are simultaneously using your Gradio app, you might run into issues. For example, 

* User A runs function 1, which returns a video with a fixed filepath. 
* Before User A does anything else, User B runs function 1, which returns a video with a fixed filepath. 
* Then, User A tries to run function 2, which takes as input the filepath returned by function 1

Instead of using User A's video, it will actually User B's video, which could be bad. 

----

So I think the only solution (which admittedly complicates the signature of `.postprocess()`) is to add a flag to the `.postprocess()` method which disables the creation of temporary files for cases (1) and (2) while leaving it for (3). Does that make sense? @freddyaboulton @aliabid94 @pngwn @apolinario


abidlabs:
I don't think this solves your problem fully @apolinario because we still need to create a copy of a file if it is returned by a function to avoid the race condition that I described above

abidlabs:
Some more ideas after internal discussion:

* For use case (3), save the temporary files based on the hash names to avoid creating multiple copies of the same file
* Whitelist based on file name instead of file directory so that we can avoid having non-deterministic directory names which will be different across different replicas

freddyaboulton:
Will whitelisting individual files greatly increase the size of this set? https://github.com/gradio-app/gradio/blob/main/gradio/blocks.py#L560

Right now, there is one temp dir per block when the demo is created but thousands of individual files can be created across all client sessions so memory should increase by at least an order of magnitude, right?

abidlabs:
That’s true but it’s just a list of strings so even if it’s thousands of
files, it’s not a problem?

abidlabs:
This PR is finally ready for review -- the main thing that this PR does is that creates temporary files with *deterministically-generated names* by hashing the file contents using SHA1 and using that as the suffix for the temporary files. This means if you are running Gradio across different replicas (as in #2732), it should no longer be an issue since the file should be named the same across different replicas. 

It doesn't exactly solve #2660 since temp files are still created, but for the case mentioned in this issue, temp files will only be created a single time when the app first launches. After that, since the hash is deterministic, it will simply return the temp file that was created the first time that the function runs. The reason I decided not to skip creation of temporary files for scenarios (2) and (3) is because it would complicate the signature of the `.postprocess()` function. There's actually a better way, which is to use `serialize()` for (2) and (3) instead of `postprocess()`, but that might cause some breaking changes and so I'll leave that for 4.0, and will explain my thinking in a separate issue.

Some additional things in this PR:
* The addition of a `TempFileManager` class to handle all of this cleanly
* Deprecated the `type` `"file"` for `Image` and `Audio` components as they had been deprecated for a while and required unnecessary maintenance
*  Eliminated the concepts of `temp_dirs` for each component and instead each component now tracks `temp_files` 
* Serve URLs directly in various components (e.g. Gallery) instead of downloading the URLs as files and then serving them
* Added tests to ensure that unnecessary temp files are not being created

This PR primarily affects the following components: `Video, Audio, File, UploadButton, Gallery, Model3D`, so if you could please test out these components, and confirm that they are working as expected, that would be great!

abidlabs:
Btw I know that a bunch of tests are not passing, I'm a little confused as to why since many seem to be unrelated to my changes. 

But I'll be OOO for the next 2 days, so I'd appreciate @freddyaboulton and @aliabid94 if you guys can still take a look at this PR and confirm that demos are still working (I've tested the kitchen sink demo and it works fine, but it would be great to test the 6 components affected in this PR more thoroughly). If you see errors, I'll fix them when I get back!

gradio/components.py
```diff
@@ -2184,7 +2207,9 @@ def process_single_file(f):
             )
             if self.type == "file":
                 if is_file:
-                    file = processing_utils.create_tmp_copy_of_file(file_name)
+                    temp_file_path = self.make_temp_copy_if_needed(file_name)
```
freddyaboulton, line 2210: The front-end can't fetch the input files of the zip_files demo

zip_file_download

freddyaboulton: When you don't cache the examples, you can't fetch the output or the input. When you do cache examples you can download the output. 

Here is the behavior without example caching

zip_file_output_download

gradio/components.py
```diff
@@ -1886,25 +1905,22 @@ def preprocess(self, x: Dict[str, str] | None) -> Tuple[int, np.array] | str | N
         )
         crop_min, crop_max = x.get("crop_min", 0), x.get("crop_max", 100)
         if is_file:
-            file_obj = processing_utils.create_tmp_copy_of_file(file_name)
+            temp_file_path = self.make_temp_copy_if_needed(file_name)
```
freddyaboulton, line 1908: The streaming audio demo doesn't work now. The front end can't fetch the audio files. 

stream_audio_tmp_file

freddyaboulton: I don't think Audio is playable now

image

gradio/components.py
```diff
@@ -1682,15 +1685,15 @@ def preprocess(self, x: Dict[str, str] | None) -> str | None:
             x.get("is_file", False),
         )
         if is_file:
-            file = processing_utils.create_tmp_copy_of_file(file_name)
+            file = self.make_temp_copy_if_needed(file_name)
```
freddyaboulton, line 1688: I'm getting an error when running a video identity demo

```python
import gradio as gr

def video_identity(video):
    return video


demo = gr.Interface(
    video_identity,
    gr.Video(source="webcam"),
    "playable_video",
)

if __name__ == "__main__":
    demo.launch()
```

video_bug

freddyaboulton (commented):
Great PR @abidlabs ! I think the new temp file creation logic is causing some problems for the front end since they it can no longer fetch files for Video, Audio, Model3d, etc. I left some comments + repros for those below. 

The test failures in `test_components` are pretty confusing. Hopefully fixing the demos will eliminate those failures!

abidlabs:
Ahh this is so helpful. Thanks @freddyaboulton for reviewing the PR while it still needed some work. I think I know what the issue is with the tests now, will fix when I'm back

MrNocTV:
How is things going guys? I cant wait for this to be released 🙏

test/test_components.py
```diff
@@ -12,7 +12,7 @@
 import tempfile
```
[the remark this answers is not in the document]
abidlabs: Resolved now

gradio/components.py
```diff
@@ -1726,10 +1731,23 @@ def postprocess(self, y: str | None) -> Dict[str, str] | None:
         if y is None:
             return None
 
-        if utils.validate_url(y):
-            y = processing_utils.download_to_file(y, dir=self.temp_dir).name
-
         returned_format = y.split(".")[-1].lower()
+
+        if self.format is None or returned_format == self.format:
```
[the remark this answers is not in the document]
abidlabs: This should have only applied to URLs. Fixed now

gradio/processing_utils.py
```diff
@@ -323,10 +310,55 @@ def file_to_json(file_path):
...
+    It should be instantiated in the __init__ method of the component.
+    """
+
+    def __init__(self) -> None:
+        # Set stores all the temporary files created by this component.
+        self.temp_files = set()
+
+    def hash_file(self, file_path: str, chunk_num_blocks: int = 128) -> str:
+        sha1 = hashlib.sha1()
+        with open(file_path, "rb") as f:
```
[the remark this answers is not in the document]
abidlabs: Fixed issue with `reverse_audio` demo

gradio/components.py
abidlabs: Fixed now

gradio/components.py
abidlabs: Both of these work for me now

gradio/processing_utils.py
```diff
@@ -323,10 +310,55 @@ def file_to_json(file_path):
     return json.load(open(file_path))
 
 
+class TempFileManager:
```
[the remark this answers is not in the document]
abidlabs: Working now

gradio/components.py
```diff
@@ -3815,13 +3848,14 @@ def preprocess(self, x: Dict[str, str] | None) -> str | None:
             x.get("is_file", False),
         )
         if is_file:
```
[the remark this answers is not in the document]
abidlabs: Working now with and without caching

abidlabs:
All righty, I believe I've sorted out the issues from your review @freddyaboulton (thanks again for the detailed review) and have tested with a variety of demos. Should be ready for another pass!

@MrNocTV thanks again for creating the original issue and following up. While we're reviewing, I've created a beta release that you can try out immediately, simply install `pip install gradio==3.13.0b1`. If you find any issues, don't hesitate to let us know!

test/test_components.py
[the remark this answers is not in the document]
freddyaboulton: What was the problem? Just curious

gradio/processing_utils.py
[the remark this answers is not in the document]
freddyaboulton: I think the behavior of this demo is different when compared to main.

When you click submit again, the same files as the previous generation show up at first but then they get replaced with the new generation.

gallery_tmp_files


On main, you get new images immediately

gallery_tmp_files_main

gradio/components.py
freddyaboulton: Got a black screen just now:

image

gradio/components.py
freddyaboulton: It's working on spaces but not for me locally. I'm getting the following for audio debugger

```ValueError: File cannot be fetched: /var/folders/t1/j7cmtcgd0mx43jh9nj_r9mmw0000gn/T/cantina92aff6ffe140da201a4a94cb3c3b9e1ff0b2a25a.wav. All files must contained within the Gradio python app working directory, or be a temp file created by the Gradio python app.
```

I'm wondering if it has anything to do with previous runs of this demo on other branches on my machine

abidlabs: Can't reproduce this. Is this with caching enabled?

gradio/processing_utils.py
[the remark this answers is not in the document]
abidlabs: Nice catch!!

gradio/components.py
freddyaboulton: Oh that could be it! Will check when I’m back home. Disregard for now 😄

gradio/processing_utils.py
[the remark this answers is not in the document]
abidlabs: Ok so the reason this is happening is because we no longer download images from a URL if we don't need to. So previously, we would download the image from the unsplash to a temporary file, and then serve the image from the temporary file in the Gallery. This is not good because (1) it unnecessarily downloads files from a URL (2) it slows down the inference process. Now, for outputs, we simply pass through the URL and the Gallery component displays the URLs. However, this means that it spends a little bit of time fetching the image from the URL. Coupled with lazy loading of images and the fact that the Gallery doesn't seem to "destroy" its children images, we get the behavior shown here. 

I think this is still preferable to what we had before. Perhaps the Gallery component should "destroy" the images instead of just replacing the URLs, but that requires some frontend shenanigans. I think this is a relatively rare case so we don't need to worry about it unless we hear otherwise from users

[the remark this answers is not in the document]
freddyaboulton: Ok I think that's ok since it only happens with urls and urls + gallery are pretty rare probably! Thanks for looking into it!

gradio/processing_utils.py
```diff
@@ -325,10 +304,98 @@ def file_to_json(file_path):
...
+            total_read += chunk_num_blocks * sha1.block_size
+            if not data or total_read > max_file_size:
+                break
+            sha1.update(data)
+        return sha1.hexdigest()
+
+    def get_temp_file_path(self, file_path: str) -> str:
+        file_name = os.path.basename(file_path)
+        prefix, extension = file_name, None
+        if "." in file_name:
```
freddyaboulton, line 340: For the bug bash, maybe we can refactor this to make use of pathlib `stem`, `root`, `suffix` properties of paths to not have to create this logic ourselves in multiple methods.

freddyaboulton (approved):
Thanks for the great PR and addressing all comments @abidlabs ! Left some none blocking comments here.

gradio/processing_utils.py
abidlabs: Yes let me create an issue for this

gradio/version.txt
```diff
@@ -1 +1 @@
-3.13.2
```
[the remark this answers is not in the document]
abidlabs: Lol a blank line ending I believe

abidlabs:
Should be all set -- will merge after the tests pass

abidlabs:
Thank you for the very thorough review @freddyaboulton!!

PR #2764, opened 2022-12-05
[WIP] gr.ScatterPlot component

freddyaboulton:
# Description

Still WIP - need to add tests and figure out responsive sizing  but wanted to open PR before weekly synch.

Implements a `gr.ScatterPlot` component. It also makes it so that you can do `gr.Plot(...).style(container=False)` for any kind of plot. Sometimes plots look nicer without the container. 

My plan is to have a separate component for each type of plot but they all inherit for `gr.Plot`. They should only differ from `Plot` in their init signatures and postprocess methods. This saves us from creating a different ui component as well. 

Right now, developers are expected to configure how their plot is constructed via the init arguments. For example, the title, labels, color field etc. These are stored in the scatter plot component instance. However, this means that users cannot update these via `gr.ScatterPlot.update`. They can only update the data that is plotted `value` and other ui related properties, e.g. `visible`, `label`. 

To test:

```python
import gradio as gr
from vega_datasets import data


cars = data.cars()

io = gr.Interface(fn=lambda:cars,
                     inputs=None,
                     outputs=[gr.ScatterPlot(x="Horsepower", y="Miles_per_Gallon", color="Cylinders")])

with gr.Blocks() as demo:
    with gr.Row():
        with gr.Column():
            gr.ScatterPlot(value=cars, x="Horsepower", y="Miles_per_Gallon",
                    color="Origin", tooltip="Name", title="Car Data", y_title="Miles per Gallon")
        with gr.Column():
            gr.ScatterPlot(show_label=False,
                        value=cars,
                        x="Horsepower",
                        y="Miles_per_Gallon",
                        color="Origin",
                        tooltip="Name",
                        title="Car Data",
                        y_title="Miles per Gallon",
                        legend_title="Origin of Car").style(container=False)
        with gr.Column():
            gr.ScatterPlot(value=cars, x="Horsepower", y="Miles_per_Gallon",
                           tooltip="Name", title="Car Data").style(container=False)
    with gr.Row():
        io.render()

demo.launch()
```

image

Closes: # (issue)

# Checklist:

# A note about the CHANGELOG


ui/packages/app/src/main.ts
```diff
@@ -192,6 +192,7 @@ function create_custom_element() {
 		root: ShadowRoot;
 		wrapper: HTMLDivElement;
 		_id: number;
+		theme: string;
```
freddyaboulton, line 195: All the changes here are so that the `Plot` component has access to the current theme

ui/packages/plot/src/Plot.svelte
```diff
@@ -1,20 +1,55 @@
...
+	import {create_config} from "./utils"
 
 	export let value;
 	export let target;
 	let spec = null;
+	export let colors: Array<string> = [];
+	export let theme: string;
+	export let caption: string;
+	
+	function get_color(index: number) {
```
freddyaboulton, line 18: I yanked this from `Chart.svelte`. Ideally this would be consolidated but it relies on the global `colors` prop so I wasn't sure what the best way to refactor was.

freddyaboulton: This function should produce the same colors for qualitative scales as the `Chart.svelte` (e.g. `TimeSeries`) component which is why I used it here

ui/packages/plot/src/Plot.svelte
```diff
@@ -1,20 +1,55 @@
...
+	}
+
+	$: darkmode = theme == "dark"
 
 	$: if(value && value['type'] == "altair") {
 		spec = JSON.parse(value['plot'])
+		const config = create_config(darkmode);
+		spec['config'] = config;
+		switch (value['chart'] || '') {
+			case "scatter":
```
freddyaboulton, line 39: My guess is that we'll have a separate case for each plot type. I'd expect this code to change in the future though as we add more types.

ui/packages/app/src/main.ts
```diff
@@ -260,6 +262,7 @@ function create_custom_element() {
 			const _autoscroll = autoscroll === "true" ? true : false;
 
 			this.wrapper.style.minHeight = initial_height || "300px";
+			console.log(this.theme);
```
pngwn, line 265: remove.

ui/packages/plot/src/Plot.svelte
```diff
@@ -1,20 +1,55 @@
...
+			return current_color;
+		}
+	}
+
+	$: darkmode = theme == "dark"
 
 	$: if(value && value['type'] == "altair") {
 		spec = JSON.parse(value['plot'])
+		const config = create_config(darkmode);
+		spec['config'] = config;
```
pngwn, line 37:
```suggestion
		spec.config = config;
```

This is more idiomatic.

ui/packages/plot/src/Plot.svelte
```diff
@@ -1,20 +1,55 @@
...
+			return color_palette[get_next_color(index) as keyof typeof color_palette]
+				.primary;
+		} else {
+			return current_color;
+		}
+	}
+
+	$: darkmode = theme == "dark"
 
 	$: if(value && value['type'] == "altair") {
```
pngwn, line 34:
```suggestion
	$: if(value && value.type == "altair") {
```

ui/packages/plot/src/Plot.svelte
```diff
@@ -1,20 +1,55 @@
...
 		spec = JSON.parse(value['plot'])
+		const config = create_config(darkmode);
+		spec['config'] = config;
+		switch (value['chart'] || '') {
```
pngwn, line 38:
```suggestion
		switch (value.chart || '') {
```

ui/packages/plot/src/Plot.svelte
```diff
@@ -1,20 +1,55 @@
...
+			case "scatter":
+				if (spec['encoding']['color'] && spec['encoding']['color']['type'] == 'nominal') {
+					spec['encoding']['color']['scale']['range'] = spec['encoding']['color']['scale']['range'].map((e, i) => get_color(i));
+				}
+				else if (spec['encoding']['color'] && spec['encoding']['color']['type'] == 'quantitative') {
+					spec['encoding']['color']['scale']['range'] = ['#eff6ff', '#1e3a8a'];
+					spec['encoding']['color']['scale']['interpolate'] = "hsl";
+				}
```
pngwn, line 46: I won't suggest but use dot notation for property access:  `object.prop1.prop2.prop3` etc.

ui/packages/plot/src/Plot.svelte
```diff
@@ -1,20 +1,55 @@
-<script>
+<script lang='ts'>
 	//@ts-nocheck
 	import Plotly from "plotly.js-dist-min";
 	import { Plot as PlotIcon } from "@gradio/icons";
+	import { colors as color_palette, ordered_colors } from "@gradio/theme";
+	import { get_next_color } from "@gradio/utils";
 	import { Vega } from "svelte-vega";
```
pngwn, line 7: We'll probably want to dynamically load the SDK for the specific plot in the future but this is fine for now. Just mentioning it here. No changes required.

ui/packages/plot/src/Plot.svelte
pngwn: We'll probably want to update the color schemes for visualisations in the future with a few pre-baked options for users. Data-viz colour schemes have some specific requirements/ demands that fall outside of typical 'design' issues. We can address that (and the duplication) later.

pngwn (changes requested):

ui/packages/plot/src/Plot.svelte
freddyaboulton: Completely agree! Filed #2780

pngwn (approved):
Code looks good. Not sure about the pending design questions but I think we can solve them in a follow-up, if needed.

CHANGELOG.md
```diff
@@ -2,6 +2,41 @@
 
 ## New Features:
 
+### Scatter plot component
+
+It is now possible to create a scatter plot without knowledge of a plotting library!
```
abidlabs, line 7: Just a suggestion, feel free to ignore
```suggestion
It is now possible to create a scatter plot natively in Gradio!
```

gradio/components.py
```diff
@@ -3937,8 +3939,296 @@ def postprocess(self, y: str | None) -> Dict[str, str] | None:
...
+        tooltip: Optional[List[str] | str] = None,
+        x_title: Optional[str] = None,
+        y_title: Optional[str] = None,
+        color_legend_title: Optional[str] = None,
+        size_legend_title: Optional[str] = None,
+        shape_legend_title: Optional[str] = None,
+        height: Optional[int] = None,
+        width: Optional[int] = None,
+        caption: Optional[str] = None,
+        interactive: Optional[bool] = True,
```
abidlabs, line 3978: Should we use a different parameter than `interactive`? For all of the other components, `interactive` means whether or not the component is used as an input or an output -- if we were to extend plots to be input components like we've considered, it might be helpful to have a separate parameter

gradio/components.py
```diff
@@ -3965,8 +3967,296 @@ def postprocess(self, y: str | None) -> Dict[str, str] | None:
...
+    Postprocessing: expects a pandas dataframe with the data to plot.
+
+    Demos: native_plots
+    """
+
+    def __init__(
+        self,
+        x: Optional[str] = None,
+        y: Optional[str] = None,
+        value: Optional[pd.DataFrame] = None,
```
abidlabs, line 3992: I would have expected the `value` be the first parameter (that's what's in the proposed API in the PR) but also it's similar to how other plotting libraries do it. Suggest reordering? 

Also what happens if `x` and `y` are not provided? Should we throw an error or is there an automatic way to infer them? 

```suggestion
        value: Optional[pd.DataFrame] = None,
        x: Optional[str] = None,
        y: Optional[str] = None,
        *
```

gradio/components.py
abidlabs: Never mind, I think `interactive` might actually be appropriate here. We don't need to overthink it

test/test_blocks.py
```diff
@@ -294,7 +294,9 @@ def test_slider_random_value_config(self):
         assert not any([dep["queue"] for dep in demo.config["dependencies"]])
 
     def test_io_components_attach_load_events_when_value_is_fn(self, io_components):
-        io_components = [comp for comp in io_components if not (comp == gr.State)]
+        io_components = [
+            comp for comp in io_components if comp not in [gr.State, gr.ScatterPlot]
```
abidlabs, line 298: `gr.ScatterPlot` should also be able to take in a function for the `value`, no?

abidlabs (approved):
Powerhouse of a PR @freddyaboulton! I only had a suggestion for the order of the parameters in the class, but otherwise LGTM. Tested and works great

freddyaboulton:
Thank you all so much for the help and reviews!

PR #2767, opened 2022-12-07
Merge highlighted text when neighbors

payoto:
In the previous implementation dictionary inputs were not correctly
processed meaning: a None token was added on the empty string between
each entity. That None entry is removed letting neighbouring entities
be merged.

# Description

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

Closes: # (issue)

# Checklist:

# A note about the CHANGELOG


test/test_components.py
```diff
@@ -1449,6 +1449,16 @@ def test_postprocess(self):
         result_ = component.postprocess({"text": text, "entities": entities})
         assert result == result_
 
+        # Test split entity is merged
+        text = "Wolfgang lives in Berlin"
```
freddyaboulton, line 1453: Thanks for the PR @payoto !

This test is no longer passing with your change. Can you fix?

gradio/components.py
```diff
@@ -3257,7 +3257,8 @@ def postprocess(
                 index = 0
                 entities = sorted(entities, key=lambda x: x["start"])
                 for entity in entities:
-                    list_format.append((text[index : entity["start"]], None))
+                    if index != entity["start"]:
```
freddyaboulton, line 3260: Just to be clear - users would still need to set `combine_adjacent=True` right?

test/test_components.py
payoto: Yes I can, FYI the fix I can see would be to edit the reference result which includes some of the `('', None)`, is that an ok fix?

Entries are still only merged when the user passes the specific argument.

This change does include a change in the behaviour of the postprocess method. An alternative way to fix this bug which would not change the default behavior would be to ignore entries which are `('', None)` when merging neighbors. Which solution do you prefer?

Personally I think the first one is the way to go but you certainly have more context as to how these features might be used than I do.

gradio/components.py
payoto: Yes, that's correct

test/test_components.py
payoto: Ok in the end I went with the change that creates least changes of behaviour, so I do the merge by not considering the entries which say `('', None)` in the merge strategy

abidlabs:
Thanks so much for the PR @payoto! So that we can test it out, can you provide an example of an example that was being processed incorrectly previously?

payoto:
Merging on dictionary inputs would simply not work:

```python
import gradio as gr

nlp = pipeline("ner", model="dslim/bert-base-NER")
example = "My name is Sylvain and I live in Berlin"

ner_results = nlp(example)
print(ner_results)

gr.Interface(
    lambda p: dict(text=p, entities=nlp(p)),
    inputs=gr.Textbox(),
    outputs=gr.HighlightedText(combine_adjacent=True,),
    examples=[example],
).launch()

```
Raw input
```
[{'entity': 'B-PER', 'score': 0.9987564, 'index': 4, 'word': 'S', 'start': 11, 'end': 12}, {'entity': 'B-PER', 'score': 0.93149716, 'index': 5, 'word': '##yl', 'start': 12, 'end': 14}, {'entity': 'B-PER', 'score': 0.79625636, 'index': 6, 'word': '##va', 'start': 14, 'end': 16}, {'entity': 'B-PER', 'score': 0.8565875, 'index': 7, 'word': '##in', 'start': 16, 'end': 18}, {'entity': 'B-LOC', 'score': 0.9995517, 'index': 12, 'word': 'Berlin', 'start': 33, 'end': 39}]
```

Before the patch:

image

After the patch:

image

test/test_components.py
```diff
@@ -1449,6 +1449,27 @@ def test_postprocess(self):
...
 
+        # Test split entity is merged when combine adjacent is set
+        text = "Wolfgang lives in Berlin"
+        entities = [
+            {"entity": "PER", "start": 0, "end": 4},
+            {"entity": "PER", "start": 4, "end": 8},
+            {"entity": "LOC", "start": 18, "end": 24},
+        ]
+        # After a merge empty entries are stripped except the leading one
+        result_after_merge = [("", None)] + [
```
abidlabs, line 1460: Nit: I think it would be clearer for readers of the codebase to just write this out:
```suggestion
        result = [
            ("", None),
            ("Wolfgang", "PER"),
            (" lives in ", None),
            ("Berlin", "LOC"),
        ]
```

test/test_components.py
```diff
@@ -1449,6 +1449,27 @@ def test_postprocess(self):
...
+        ]
+        default_adjacent = component.combine_adjacent
+        component.combine_adjacent = False
+        result_ = component.postprocess({"text": text, "entities": entities})
+        assert result != result_
+        assert result_after_merge != result_
+        component.combine_adjacent = True
+        result_ = component.postprocess({"text": text, "entities": entities})
+        assert result_after_merge == result_
+        component.combine_adjacent = default_adjacent
```
abidlabs, line 1471: Instead of changing the attribute of the component, simply create a new component when necessary
```suggestion
        result_ = component.postprocess({"text": text, "entities": entities})
        assert result != result_
        assert result_after_merge != result_
        
        component = gr.HighlightedText(combine_adjacent=True)
        result_ = component.postprocess({"text": text, "entities": entities})
        assert result_after_merge == result_
        
        component = gr.HighlightedText(combine_adjacent=True)
```

abidlabs:
Thanks for making the PR @payoto! Just pushed Was thinking through this PR and while this does fix the bug, it treats the `None` label differently than the labels for the other entities. It would be better if we could treat all of the entities the same, and I think we can do that by changing the condition from `elif text == "" and category is None:` to `elif text == ""` (or the more Pythonic elif not text).

Let me know what you think of the suggestion. I also added some suggestions for the test. 

If this looks good, please go ahead and make the suggestions, and happy to review again. cc @freddyaboulton

freddyaboulton (commented):
This looks good to me @payoto ! Thank you for addressing the changes and I can confirm it now properly merges adjacent entries. Please fix the linting `bash scripts/format_backend.sh` and I will merge 🚀 

### This branch
image

### Main
image

abidlabs:
Merging, thank you so much @payoto for this fix.

PR #2772, opened 2022-12-08
Deindex share URLs from search

aliabd:
Adds `/robots.txt` to every gradio app (that disallows all) so that it doesn't get indexed by google and show up in searches. [Example of indexed links](https://www.google.com/search?q=site:gradio.app&sxsrf=ALiCzsYLU0TuFIIwuxxbrPfOPqR6bjApKQ:1670479584635&ei=4H6RY8CxJqeawbkP4ru6sA4&start=40&sa=N&ved=2ahUKEwjAofyxren7AhUnTTABHeKdDuY4HhDy0wN6BAggEAs&biw=1973&bih=1084&dpr=0.9).


julien-c:
this should only apply to the apps hosted on `*.gradio.app`, right? On Spaces (and presumably when people self-host) we want the apps to be indexed...

aliabd:
Turns out it's non trivial to do this on the api server side so will keep it here. But I changed it to only disallow indexing when `share=True` which means it shouldn't affect self hosting or spaces.

abidlabs (approved):

freddyaboulton (approved):

PR #2774, opened 2022-12-08
Redirect old links and fix broken ones

aliabd:
I did a crawl of the website to see all the broken links and noticed a lot of issues, including links with some equity that no longer exist (now have different URLs). I fixed the broken ones and redirected old ones to the new URLs in the website nginx.


guides/06_other_tutorials/create_your_own_friends_with_a_gan.md
```diff
@@ -162,7 +162,7 @@ gr.Interface(
 
 The `examples` parameter takes a list of lists, where each item in the sublists is ordered in the same order that we've listed the `inputs`. So in our case, `[seed, num_punks]`. Give it a try!
 
-You can also try adding a `title`, `description`, and `article` to the `gr.Interface`. Each of those parameters accepts a string, so try it out and see what happens 👀 `article` will also accept HTML, as [explored in a previous guide](./adding_rich_descriptions_to_your_demo)!
+You can also try adding a `title`, `description`, and `article` to the `gr.Interface`. Each of those parameters accepts a string, so try it out and see what happens 👀 `article` will also accept HTML, as [explored in a previous guide](https://gradio.app/key_features/#descriptive-content)!
```
abidlabs, line 165:
```suggestion
You can also try adding a `title`, `description`, and `article` to the `gr.Interface`. Each of those parameters accepts a string, so try it out and see what happens 👀 `article` will also accept HTML, as [explored in a previous guide](./key_features/#descriptive-content)!
```

website/homepage/nginx.conf
```diff
@@ -3,44 +3,38 @@ server {
...
     location / {
         root   /usr/share/nginx/html;
         try_files $uri $uri/ $uri/index.html =404;
         index  index.html index.htm;
         autoindex on;
     }
 
-    error_page  404              /404.html;
+    location /introduction_to_blocks {
+        return 301 /quickstart;
```
abidlabs, line 14: Is it possible to include the anchor link specifically to this section? May not be possible via nginx I'm not sure

website/homepage/nginx.conf
```diff
@@ -3,44 +3,38 @@ server {
...
+    }
+
+    location /adding_examples_to_your_app {
```
abidlabs, line 17: Same as above

website/homepage/nginx.conf
```diff
@@ -3,44 +3,38 @@ server {
...
+    }
 
-    # redirect server error pages to the static page /50x.html
-    #
+    location /guides.html {
+        return 301 /guides;
+    }
+
+    
+    error_page  404              /404.html;
```
abidlabs, line 34: Where is this 404.html file? I don't see it. Nor do I see the /50x.html file below.

Navigating manually to `http://localhost:8000/404.html` doesn't work for me either

website/homepage/src/guides/__init__.py
```diff
@@ -153,18 +153,6 @@ def build_guides(output_dir, jinja_env):
...
-            output_folder = os.path.join(output_dir, "getting_started")
-            os.makedirs(output_folder)
-            index_file = os.path.join(output_folder, "index.html")
-            with open(index_file, "w") as index_html:
-                index_html.write(output)
-            output_folder = os.path.join(output_dir, "quickstart")
-            os.makedirs(output_folder)
-            index_file = os.path.join(output_folder, "index.html")
-            with open(index_file, "w") as index_html:
-                index_html.write(output)
```
abidlabs, line 166: What were these files for before?

website/homepage/src/guides/__init__.py
```diff
@@ -174,10 +162,6 @@ def build_gallery(output_dir, jinja_env):
     output_file = os.path.join(output_folder, "index.html")
     with open(output_file, "w") as index_html:
         index_html.write(output)
-    output_file = os.path.join(output_dir, "guides.html")
-    with open(output_file, "w") as index_html:
-        index_html.write(output)
```
abidlabs, line 179: Same, what was this for?

abidlabs:
Thanks for fixing these links @aliabd! Some suggestions / questions above

website/homepage/nginx.conf
aliabd: will try

website/homepage/nginx.conf
aliabd: This line isn't doing anything (already existed before), but if we want to create a 404/503 page we can. I can make a simple one in a separate PR to look better than the standard nginx pages.

website/homepage/src/guides/__init__.py
aliabd: when we changed `/getting_started` to be `/quickstart` I wrote this so that the old links to /getting_started would still work. not necessary now because I redirect them on nginx. which is much better for seo because its an actual redirect to a single page instead of this which was creating too separate pages with the same content.

website/homepage/src/guides/__init__.py
aliabd: Same reason as above, we had launched guides with a link to `/guides.html` but then switched to `/guides` so I had created both to be supported. but now I redirect them in nginx.

website/homepage/nginx.conf
abidlabs: Got it thanks -- agreed we can do in a separate PR, not very important

website/homepage/nginx.conf
aliabd: This isn't possible with nginx, basically the # tag is a browser thing and the server ignores it. https://stackoverflow.com/a/9794542

abidlabs: Ok good to know!

aliabd: actually I was wrong, it is. Will fix those

PR #2785, opened 2022-12-10
Add component HostFile that allows users to browse and select files from host machine

aliabid94:
Requested by @camenduru and @apolinario 

Recording 2022-12-09 at 21 22 36


abidlabs:
@aliabid94 given how different this is from a typical input/output component, I think we should hold off on this for now, close this PR, and resurrect it as a custom component when we add support for that in early 2023.

Issue #2658, opened 2022-11-16
Allow gr.DataFrame() to take all pandas dataframe types


aliabid94:
E.g. timestamps currently don't work

PR #2804, opened 2022-12-13
Add support for numpy array and other types to `gr.Dataframe()` initial value

abidlabs:
As I was writing the BigQuery guide, I noticed that we would run into an error if we passed in a `pd.DataFrame` that contained numpy arrays or any non-builtin python types as the `value` of a `gr.DataFrame`. In other words, this would error out:

```py
import pandas as pd
import numpy as np
import gradio as gr

df = pd.DataFrame(
    {
        "date_1": pd.date_range("2021-01-01", periods=2),
        "date_2": pd.date_range("2022-02-15", periods=2).strftime("%B %d, %Y, %r"),
        "number": np.array([0.2233, 0.57281]),
        "number_2": np.array([84, 23]).astype(np.int64),
        "bool": [True, False],
        "markdown": ["# Hello", "# Goodbye"],
    }
)

with gr.Blocks() as demo:
    gr.DataFrame(df)
demo.launch()
```

Interestingly, it would not error out if we returned such as pandas array from a function. The reason is that we were using different ways to serialize data before sending it to the frontend depending on if it was part of the config or not. This fixes that by using the same way (`orjson.dumps`) to serialize both. It should make the `pd.Dataframe` a lot more robust to different types. Also added a test and fixed some other tests which were incorrectly marked as async.

Closes: #2658


freddyaboulton (approved):
Looks good @abidlabs ! Thanks for the fix. This fixes #2658 no? I would add pd.DataFrame to the value typehint for dataframe since that's missing right now.


I wonder if we can simplify the datatype argument for 4.0. It supports "datetime" which you'd think would take care of dates but it doesn't do anything except do markdown conversion if the datatype arg is markdown. Just a random thought while reviewing.

abidlabs:
Thanks for the review @freddyaboulton! I had missed that issue, so will add that in. I'll change the base branch to `main` so that we can merge that in.

PR #2805, opened 2022-12-13
Typo in the parameter `visible` in `templates.py`

abidlabs:


abidlabs:
Going through the components, I realized the `requires_permissions` parameters is no longer used, so I removed that as well

gradio/components.py
```diff
@@ -104,14 +104,12 @@ def __init__(
         show_label: bool = True,
         interactive: Optional[bool] = None,
         visible: bool = True,
-        requires_permissions: bool = False,
```
freddyaboulton, line 107: Why is this deleted?

gradio/templates.py
```diff
@@ -261,7 +261,7 @@ def __init__(
             label=label,
             show_label=show_label,
             interactive=interactive,
-            visibile=visible,
+            visible=visible,
```
freddyaboulton, line 264: 🙈

freddyaboulton (approved):
Thanks @abidlabs !

gradio/components.py
abidlabs: We used to use `requires_permissions` to figure out if a component needed to request permissions in colab and print a message accordingly. We don't do need to do that anymore. (So unrelated to the original reason I made this PR, but when I saw it, I figured I'd just remove it)

abidlabs:

PR #2810, opened 2022-12-14
Patching `test_get_ip` attempt 2

abidlabs:
Previously, we were using `https://api.ipify.org` to get the IP address. But this website has started to occasionally return a "Bad Gateway" response. So I switched to `https://checkip.amazonaws.com/`, which should be more reliable and also marked the test as flaky as it still relies on an external service


abidlabs:
Thanks @aliabid94

PR #2819, opened 2022-12-15
Remove `jinja2>=3.0` dependencies and release `gradio==3.13.2`

abidlabs:
It turns out that `jinja2.utils.htmlsafe_json_dumps` was introduced in 3.0, which means that on environments (like Colab) where it is not the default, it will break. Rather than pinning the jinja version, this manually implements the `htmlsafe_json_dumps` method.

Note that I've included `markupsafe` in the `requirements.txt` file but this is not really a new dependency as it is already a dependency of `jinja2`.


abidlabs:
Both local and public Gradio apps seem to work on Colab with the latest changes to requirements.txt file: https://colab.research.google.com/drive/1wlc8vncpPLFOTSzU1tbO8pzmCb90yvri?usp=sharing

Should be good to release `gradio==3.13.2` now!

gradio/routes.py
```diff
@@ -73,7 +73,13 @@ def _render_str(content: Any) -> str:
 
 
 def toorjson(value):
-    return htmlsafe_json_dumps(value, dumps=ORJSONResponse._render_str)
+    return markupsafe.Markup(
```
freddyaboulton, line 76: Is this the implementation of html_safe_json_dumps?

freddyaboulton (approved):
Thanks @abidlabs 👏

gradio/routes.py
abidlabs: Yup basically

Issue #2843, opened 2022-12-18
Gradio seems to break Pillow / PIL.Image format support


jokkebk:
### Describe the bug

Encountered this on latest Automatic1111 Stable Diffusion Webui, suddenly PIL on my Win10 machine had lost all but PNG support:

1. Importing PIL.Image and printing `registered_extensions` gives me about 20 file formats
2. BUT if I `import gradio` first and then PIL.Image, I only get `{'.png': 'PNG', '.apng': 'PNG'}`

I am not well versed enough with Gradio to understand what gets broken, my hunch would be Gradio tinkering with Pillow (as it is one requirements) upon initialization, or Gradio introducing some global variable/function that "pollutes" PIL.Image initialization.

Tried with WSL2 on Ubuntu (Gradio 3.14) and same happens, so it is not "some Windows issue" 😇 

Workaround is to `import PIL.Image` *before* gradio, but this should not happen.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Normal PIL Image support:

```
from PIL import Image
print(Image.registered_extensions())
```

[...]

PR #2846, opened 2022-12-19
Add workaround to prevent importing gradio breaking PIL.Image.registered_extensions()

aliencaocao:
# Description
This is a Pillow issue but the folks at Pillow suggested a workaround in https://github.com/python-pillow/Pillow/issues/6809#issuecomment-1356864228
Until Pillow merges https://github.com/python-pillow/Pillow/pull/6811 and publishes a new release (in Jan 2023 according to https://github.com/python-pillow/Pillow/issues/6750), this would be the only way to fix it.

The issue is importing matplotlib.figure will in turn import PIL.PngInfoPlugin, which causes existing extensions to not be initialized fully but only initialized with PNG formats. Thus, Image.init() have to be called before this import.

Closes: #2843 

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ No changes to highlight.
 No changes to highlight.
 
 ## Full Changelog:
-No changes to highlight.
+* Fixed importing gradio can cause PIL.Image.registered_extensions() to break by `[@aliencaocao](https://github.com/aliencaocao)` in `[PR 2846](https://github.com/gradio-app/gradio/pull/2846)`
 
 ## Contributors Shoutout:
 No changes to highlight.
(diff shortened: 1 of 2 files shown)


aliencaocao:
All the failing tests are due to the use of Image.init() before importing matplotlib and others, however this is necessary for the fix

abidlabs:
Thanks for the PR @aliencaocao! I tested it and it works for me. However, it also works for me if you move include the line `PIL.Image.init()` after all of the import statements (e.g. to line 67, right above `set_documentation_group("component")`)

Can you try this and see if it works for you as well? That way, we can satisfy the linter as well

abidlabs:
Looks good! Thanks again @aliencaocao for the fix

radarhere:
Pillow 9.4.0 has now been released

aliencaocao:
gradio 3.15 is also pinned to this repo, that version already includes a fix

Issue #1964, opened 2022-08-06
Error when loading min-dalle interface


nagolinc:
### Describe the bug

When importing https://huggingface.co/spaces/kuprel/min-dalle, gradio fails

```
dalle = gr.Interface.load("spaces/kuprel/min-dalle")

Fetching interface from: https://huggingface.co/spaces/kuprel/min-dalle

---------------------------------------------------------------------------

InvalidSchema                             Traceback (most recent call last)

[/usr/local/lib/python3.7/dist-packages/gradio/processing_utils.py](https://localhost:8080/#) in encode_url_or_file_to_base64(path)
     30     try:
---> 31         requests.get(path)
     32         return encode_url_to_base64(path)

19 frames

[...]

PR #2855, opened 2022-12-20
Fixed loading Spaces that have components with default values

abidlabs:
Previously, we were unable to load many Spaces if they included components with default values, i.e. had a `value` parameter specified in the config. The reason for that is that the config includes the *post-processed* value of the `value` parameter, but during the loading process, we would pass that value through the `.postprocess()` function of the component again. 

So this PR is a fairly straightforward and not-so-elegant solution to this problem. It skips the postprocessing of the `value` parameter for components in a gradio demo that is loaded using `gr.Interface.load()`. Open to feedback on the implementation side. 

I created two Spaces to replicate the problem and test the solution:
* `spaces/abidlabs/min-dalle` -- running `gradio==3.3.1` like in the original issue
* `spaces/abidlabs/min-dalle-later` -- running `gradio==3.14.0` to confirm that the issue is not specific to the base64 encoding that the Gallery component used to use
 
Fixes: #1964
Fixes: #2859


gradio/blocks.py
```diff
@@ -657,13 +660,14 @@ def iterate_over_children(children_list):
...
 
             # Allows some use of Interface-specific methods with loaded Spaces
-            blocks.predict = [fns[0]]
-            blocks.input_components = [
-                Context.root_block.blocks[i] for i in first_dependency["inputs"]
-            ]
-            blocks.output_components = [
-                Context.root_block.blocks[o] for o in first_dependency["outputs"]
-            ]
+            if len(fns):
```
abidlabs, line 663: Separate issue to be able to load Spaces without any defined event triggers

freddyaboulton: Nice

test/test_external.py
```diff
@@ -384,6 +384,11 @@ def check_dataset(config, readme_examples):
         ]
 
 
+def test_load_blocks_with_default_values():
```
freddyaboulton, line 387: I would check at least some of the values in the config. Simply omitting the values from the new config would also make the test pass but that's not what we want

freddyaboulton (approved):
Thanks @abidlabs !

test/test_external.py
abidlabs: Yup good point, will do

abidlabs:
Thanks for the review @freddyaboulton! Made a small tweak so that it fixes #2859 as well

abidlabs:
Will merge in once the tests pass

Issue #2823, opened 2022-12-15
Update README.md with correct instructions on how to build website


aliabd:
### Describe the bug

npm build command now needs --url arg, need to document it somehow to avoid the blocking error to building website locally. 

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

`cd website/homepage && npm run build`

### Screenshot

![Screen Shot 2022-12-13 at 4 48 50 PM](https://user-images.githubusercontent.com/9021060/207880914-cdbf03f0-3367-4954-9655-5f99839db519.png)

### Logs

```shell
> build
> npm run build:html && npm run build:css


> build:html
> python build.py --url $npm_config_url

usage: build.py [-h] [--url URL]
build.py: error: argument --url: expected one argument
```

### System Info

```shell
v1.13.0
```

### Severity

annoying

PR #2858, opened 2022-12-20
Update website README.md with correct instructions on how to build website 

aliabd:
Closes: #2823


freddyaboulton (approved):

website/README.md
```diff
@@ -21,16 +20,14 @@ Alternatively, for development, read the `homepage` section below:
 
 ## The `homepage` docker
 
-The homepage folder builds the static content of the website into standalone files, served by nginx in docker. For development purposes, instead of running docker to test changes, follow these steps in the `gradio/website/homepage` folder.
+The homepage folder builds the static content of the website into standalone files, served by nginx in docker. For development purposes, instead of running docker to test changes, just run `sh scripts/launch_website.sh` from the root of the repo. The steps in that script are the following: 
 
 - `npm install`
-- `npm run build` (or `npm run build-mac` on Mac OSX)
+- `npm run build --url=https://gradio-main-build.s3.amazonaws.com/XXX/`
```
abidlabs, line 26: Just remove this? I don't think we need to explicitly list these out since users can just read the script. Better to have a single source of truth

abidlabs (approved):

website/README.md
aliabd: Good point

Issue #2827, opened 2022-12-15
Correctly remove footers from embedded components in the docs


aliabd:
This problem has been on and off for a while. JS isn't always behaving as expected and sometimes the footers and extra space show up under the embedded components, which makes for really bad page design.  Should be fixed at the space level in css.

PR #2860, opened 2022-12-20
Remove footers and min-height the correct way 

aliabd:
The footers and min-height on embedded components in the docs create annoying space and ruin the flow. We've been removing them with js on the page but it hasn't been reliable and often doesn't work because the components load late. This removes them using css inside the demos themselves. Live on [staging](https://www.gradio-staging.com/docs/main)

Closes: #2827


aliabd:
You can see my struggle above with the demo notebooks action, but it turned out there was actually a problem. Locally I had a `gradio_cached_examples` folder in  `fake_gan` (for some reason), so when generating the notebook it thought that was a file the demo needed to run, and added a wget for it. But `gradio_cached_examples` is in gitignore, so when the commands are run from the action there's no wget and it causes the fail. 

Will add `gradio_cached_examples` to list of files to ignore when notebooks are generated (in a separate PR)

abidlabs:
Beautiful, looks great @aliabd! Can you just verify one thing:

The embedded demos in "try examples" are not working for me on staging.

For example: https://www.gradio-staging.com/docs/main/#radio, when I click try examples, I get this: 
image


They _are_ working if I set the toggle to 3.14.0 and on the main website. But I can't see what it might have to do with your change. If they *are* working for you, then I assume it's something weird on my device, but if they aren't, then we should investigate

aliabd:
@abidlabs there's an issue with the spaces themselves, and not all of them. Just some of the `_main` spaces (the versioned ones don't get updated except on new releases). For some reason many of the `_main` spaces returned a build error. Will try again and if there's still a problem will check with the spaces team. Unrelated to embedding or this PR though so I will merge

aliabd:
@abidlabs that is to say, those demos are also broken on gradio.app/main

PR #2861, opened 2022-12-20
Re-enable Example pagination on the front-end

### LINKED ISSUE #2186: Pagination for Examples
As a gradio user, I would like to provide many examples for my app without displaying all of them in the page when the app loads.

As a possible solution, we should expose an `overflow_row_behavior` for `Dataset`/`Examples` like we do for `Dataframe`.

This was originally asked in the Discussion sections:

> Most of the demo that use gr.Examples add sample statically. Is it currently possible to load/update samples dynamically in the case one have a large dataset so potentially need pagination ?
> 
> _Originally posted by @jrabary in https://github.com/gradio-app/gradio/discussions/2163_

freddyaboulton:
# Description

Fixes #2186

The Dataset component already supported pagination! (We should publicize this!!)

So the only thing left to do was to make sure you could set `examples_per_page` from the Examples component. The parameter already existed, it just wasn't used.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,7 +23,7 @@ gr.LinePlot(stocks,
 By [@freddyaboulton](https://github.com/freddyaboulton) in [PR 2807](https://github.com/gradio-app/gradio/pull/2807) 
 
 ## Bug Fixes:
-No changes to highlight.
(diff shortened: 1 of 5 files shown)


abidlabs (approved):
LGTM @freddyaboulton great fix!

freddyaboulton:
Thanks @abidlabs !

PR #2870, opened 2022-12-21
Removes the share warning msg and upgrades to `gradio==3.15.0`

abidlabs:
Now that FRP has been working quite well for ~2 weeks, this PR removes the extra message that we display for share links. Creating a share link will now show the standard message, same as before:

image

In parallel, I'll be sending a message to everyone using an older version of Gradio to upgrade to `gradio>=3.14.0` so it would have been a little circular to keep this.

Also removes some unused strings. 

Releases `gradio==3.15.0`


abidlabs:
And if you use an older version of gradio, you will see a message like this:

image

aliabd (approved):

freddyaboulton (approved):
Thanks @abidlabs !

Issue #1076, opened 2022-04-25
Enable multi-select on gradio.Dropdown


apolinario:
**Is your feature request related to a problem? Please describe.**
Sometimes I want to have plenty of multi-select options and I also don't want to clutter the UI with a lot of checkboxes using `gr.inputs.CheckboxGroup` 

**Describe the solution you'd like**
Enabling a multi-select option on the `gradio.inputs.Dropdown` component can be an elegant way to keep multi-option selection concise but yet powerful and functional
<img width="727" alt="image" src="https://user-images.githubusercontent.com/788417/165058185-652d2d85-6c54-4dc9-884b-74962728531a.png">
Example implementation of such a feature

abidlabs:
Great suggestion, thanks @apolinario!

colelandolt33:
Hi @abidlabs , has there been any progress on implementing this feature?

abidlabs:
Not yet! @dawoodkhan82 is this something you can take on? 

@colelandolt33 can you tell us a little more about your use case so that we can make sure that it is supported by our implementation of this feature?

colelandolt33:
> @colelandolt33 can you tell us a little more about your use case so that we can make sure that it is supported by our implementation of this feature?

I'm working on an application where the user inputs a dataset and selects columns to be used in an analysis. Currently the only way to allow the user to select multiple columns is by using `gr.CheckboxGroup`, but when there are hundreds of columns, this isn't very user friendly. I'm looking for a dropdown menu where the user can make multiple selections, very similar to this [Streamlit widget](https://doc-multiselect.streamlitapp.com/). Ideally, the user could type to search for options.

AmbiakaTT:
Any updates on the feature yet? It would be really really helpful for me. By the way, for a list of 100+ options, would using checkboxgroup be helpful

abidlabs:
We haven't had a chance to look into this yet @AmbiakaTT. We'll update here when we have more to add

wjizhong:
Any updates on the feature yet? It would be really really helpful for me. By the way, for a list of 100+ options, would using checkboxgroup be helpful

abidlabs:
Hi @wjizhong not yet unfortunately. @aliabid94 or @dawoodkhan82 if you have the bandwidth to take a look at this next week, we're getting a lot of interest in this component.

Ahanmr:
Hi, is this released in the latest version as such? I'd love to know since I'm interested in using the multi-input in dropdown function. Thank you!

PR #2871, opened 2022-12-21
Enable multi-select on gradio.Dropdown

dawoodkhan82:
# Description
Add option to select multiple options on `gr.Dropdown`

Will be updating the colors to match other form components.

https://user-images.githubusercontent.com/12725292/210434026-737809ed-5e6c-4b10-97ff-e489bbfe72e4.mov

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

Closes: #1076

# Checklist:

# A note about the CHANGELOG


gradio/components.py
```diff
@@ -1204,9 +1204,9 @@ def style(
 
 
 @document("change", "style")
-class Dropdown(Radio):
+class Dropdown(Changeable, IOComponent, SimpleSerializable, FormComponent):
     """
-    Creates a dropdown of which only one entry can be selected.
+    Creates a dropdown of which only entries can be selected.
```
abidlabs, line 1209:
```suggestion
    Creates a dropdown of choices from which entries can be selected.
```

abidlabs: Can we add a multiselect demo and link to it in the Demos line?

abidlabs:
Awesome, love the fact that the dropdown is typeable as well. Will be useful for very large number of choices. 

Some suggestions on the UI / UX side:

1. The appearance of `Dropdown(multiselect=False)` and `Dropdown(multiselect=True)` are very different. We should make them more consistent, with the only difference being that you can only choose a single element if `multiselect=False`. 

image

image

Among other things, the container with the `Dropdown` label is now completely gone. We should bring that back. But also, the entire design of the list of choices is different, the textbox part is different, etc.

2. You can't seem to currently use `Dropdown(multiselect=False)` as an output component the way you can with `Dropdown(multiselect=True)`.

Try running:

```py
import gradio as gr

with gr.Blocks() as demo:
    gr.Dropdown(["angola", "pakistan", "canada"], multiselect=True, value=["angola"])
    
demo.launch()
```

vs.

```py
import gradio as gr

with gr.Blocks() as demo:
    gr.Dropdown(["angola", "pakistan", "canada"], multiselect=False, value="angola")
    
demo.launch()
```

3. Typing doesn't filter choices. As I start typing "Pa", I would expect the choices to filter to just Pakistan, and that I could use the down arrow key and enter to go down and select it. However, this does not happen:
4. 
image

gradio/components.py
```diff
@@ -1230,28 +1231,115 @@ def __init__(
...
+            "value": self.value,
+            "multiselect": self.multiselect,
+            **IOComponent.get_config(self),
+        }
+
+    @staticmethod
+    def update(
+        value: Optional[Any] = _Keywords.NO_VALUE,
+        choices: Optional[str | List[str]] = None,
+        multiselect: bool = False,
```
freddyaboulton, line 1280: Better to make `multiselect=None` otherwise any update to a multi-select dropdown will switch off the multiselect

ui/packages/form/src/Dropdown.svelte
```diff
@@ -1,27 +1,139 @@
...
+    export let placeholder = '';
 	export let label: string;
-	export let value: string | undefined = undefined;
+	export let value: string | Array<string> | undefined = undefined;
+	export let multiselect: boolean = false;
 	export let choices: Array<string>;
 	export let disabled: boolean = false;
 	export let show_label: boolean;
+  
+    let input,
```
freddyaboulton, line 14: There is a weird bug where hitting enter on the typeable field makes the whole app crash

multiselect_enter

ui/packages/form/src/Dropdown.svelte
```diff
@@ -1,27 +1,139 @@
...
-	import { createEventDispatcher } from "svelte";
+    import { fly } from 'svelte/transition';
 	import { BlockTitle } from "@gradio/atoms";
-
+    export let id = '';
+    export let readonly = false;
+    export let placeholder = '';
 	export let label: string;
-	export let value: string | undefined = undefined;
+	export let value: string | Array<string> | undefined = undefined;
```
freddyaboulton, line 8: Can't seem to update the value on a multi select drop down

multiselect_update


This is the code I ran

```python

import gradio as gr
import random

with gr.Blocks() as demo:
    dd = gr.Dropdown(["angola", "pakistan", "canada"], multiselect=True, value=["angola"], interactive=True)
    update = gr.Button(value="Update")
    update.click(lambda: gr.Dropdown.update(
                                            value=[random.choice(["angola", "pakistan", "canada"])],
                                            multiselect=True),
                 inputs=None,
                 outputs=[dd])

demo.launch()
```

freddyaboulton (commented):
@dawoodkhan82 This is so cool! It's great that developers can update whether or not multi-select is enabled via `gr.Dropdown.update` as well.

Left comments on some bugs I think I found.

In addition, I agree with @abidlabs comment about filtering with what's been typed and the look being too different from the existing dropdown.

ui/packages/atoms/src/Block.svelte
```diff
@@ -45,7 +45,7 @@
 	data-testid={test_id}
 	id={elem_id}
 	class:!hidden={visible === false}
-	class="gr-block gr-box relative w-full overflow-hidden {styles[
```
freddyaboulton, line 48: How come you made this change?

abidlabs:
Was just testing it out @dawoodkhan82, UI looks nice! Just wanted to point out that if you click on the scrollbar, an entry named "undefined" shows up in the dropdown:

image

Also maybe it's just me, but I think having just 2 entries visible at a time in the Dropdown select is too little. It should be at least 5-6 imo

dawoodkhan82:
> Was just testing it out @dawoodkhan82, UI looks nice! Just wanted to point out that if you click on the scrollbar, an entry named "undefined" shows up in the dropdown:
> 
fixed. good catch, weird edge case...
> image
> 
> Also maybe it's just me, but I think having just 2 entries visible at a time in the Dropdown select is too little. It should be at least 5-6 imo
> 
updated max height
> Is the typeability gone?

Typeability is something I had enabled by accident. I can look into making it work. Is it something we definitely want to add?

abidlabs:
> Typeability is something I had enabled by accident. I can look into making it work. Is it something we definitely want to add?

Yeah I'm imagining the case with 100s of choices (this has been asked about frequently). It's going to be hard to choose the right choices without typeability I believe

abidlabs:
Super nice @dawoodkhan82!! I tested it out and multiselect dropdown works really well. Just a couple of minor points I noticed:

(1) The blue colors used as the  background for the selected choices feels a little "different" than our usual palette. How about some faint orange colros?

image

(2) In dark mode, the background color of the dropdown choices is still white: 

image

We should make it the same color as what shows if `multiselect=False`

image

(3) The `multiselect` mode doesn't respect `interactive=False` in the way that I expect

If `multiselect=False`, the Dropdown choice is displayed, and the dropdown itself cannot be clicked on:

image

If `multiselect=True`, the Dropdown choice are displayed, but the dropdown can still be clicked on:

image

We should make it un-changeable so it is truly a non-interactive component

Other than these, LGTM! Would be good to get a review on the frontend code, perhaps by @aliabid94?

abidlabs (approved):

dawoodkhan82:
@abidlabs fixed all 3

aliabid94:
great work. styling is still a bit off, I'll just make changes directly to branch.

aliabid94:
Just looking at this, I feel like should be a completely different element than Dropdown, e.g. a `gr.Select`. It has a different type (List of strings vs string), different purpose, and different visual layout than Dropdown. There's so little in common that all the logic branches whether or not it is a multiselect. @abidlabs @dawoodkhan82

abidlabs:
> Just looking at this, I feel like should be a completely different element than Dropdown, e.g. a gr.Select. It has a different type (List of strings vs string), different purpose, and different visual layout than Dropdown. There's so little in common that all the logic branches whether or not it is a multiselect. @abidlabs @dawoodkhan82

From a user's perspective, it feels "natural" to me to have a multiselect variant of the dropdown alongside a regular single-select dropdown. But I don't have a strong opinion, I'm fine with a separate component. If we do end up making it a different component, I think `gr.Multiselect` is more of an appropriate name

dawoodkhan82:
Synced with @aliabid94 , there's changes he suggested which makes sense and I'll implement:
- [x] change the styling of the selected option bubbles to match the radio button
- [ ] separate the front end code into a separate "Multiselect.svelte" component
- [ ] re-implement the multiselect dropdown to use `datalist`
- [ ] Maybe separate the component completely (frontend + backend)? Not sure about this, I think for the same reasons as @abidlabs mentioned, it makes sense from a developer's pov to have it be the same component.

abidlabs:
This is going to be a really slick component!

dawoodkhan82:
synced with @aliabid94 and @pngwn. Decided to move forward with the custom drop-down. Will look into combining the custom dropdown with datalist to improve accessibility, along with the other forms.

gradio/components.py
aliabid94: We do not support `update` changing backend configuration yet, such as `multiselect`. `update` only allows changing frontend configuration. So please remove `multiselect` as an option to update

ui/packages/app/src/components/Column/Column.svelte
```diff
@@ -12,7 +12,7 @@
 
 <div
 	id={elem_id}
-	class="overflow-hidden flex flex-col relative col {create_classes(style)}"
+	class="flex flex-col relative col {create_classes(style)}"
```
aliabid94, line 15: I can't think of any harm that's done by removing `overflow-hidden` here, but can you confirm @pngwn?

aliabid94: This is done to allow the dropdown to extend past the parent.

ui/packages/form/src/MultiSelect.svelte
```diff
@@ -0,0 +1,195 @@
...
+		}
+	}
+
+	function handleBlur(e: any) {
+		optionsVisibility(false);
+	}
+
+	function handleKeyup(e: any) {
+		// enter key
+		if (e.keyCode === 13) {
```
aliabid94, line 64: keycode is deprecated in favor of `event.key === 'Enter'`

ui/packages/form/src/MultiSelect.svelte
```diff
@@ -0,0 +1,195 @@
...
+			if (Array.isArray(value) && activeOption != undefined) {
+				value.includes(activeOption) ? remove(activeOption) : add(activeOption);
+				inputValue = "";
+			}
+		}
+		if ([38, 40].includes(e.keyCode)) {
```
aliabid94, line 70: same here

CHANGELOG.md
```diff
@@ -20,6 +20,14 @@ demo = gr.Interface(reverse, gr.Text(), gr.Text())
 Progress indicator bar by [@aliabid94](https://github.com/aliabid94) in [PR 2750](https://github.com/gradio-app/gradio/pull/2750).
 
 * Added `title` argument to `TabbedInterface` by @MohamedAliRashad in [#2888](https://github.com/gradio-app/gradio/pull/2888)
+* Added `multiselect` option to `Dropdown` by @dawoodkhan82 in [#2871](https://github.com/gradio-app/gradio/pull/2871)
+
+### With `multiselect` set to `true` a user can now select multiple options from the dropdown menu.
```
abidlabs, line 25:
```suggestion
### With `multiselect` set to `true` a user can now select multiple options from the `gr.Dropdown` component.
```

abidlabs (approved):
Looks beautiful! Nit: can we add backend tests to ensure that preprocessing/postprocess multiselect works as we expect it to?

ui/packages/app/src/components/Column/Column.svelte
pngwn: We just need to make sure that this wasn't responsible for enforcing rounded borders (which I don't think it is/was, Block does that). Otherwise safe to remove.

Issue #2832, opened 2022-12-15
Fix typing issues in the Python codebase


abidlabs:
We use typehints across the Python codebase, but they are not checked because we would have a very large number of type errors if we were to try to enforce them in any way. 

In an iterative way, we should start fixing the typing issues so that they can be checked and expose any underlying bugs that we may have missed.

Issue #2831, opened 2022-12-15
Refactor `os.path`, `pathlib`, and manual filepath processing


abidlabs:
In various parts of the codebase, we use `os.path`, `pathlib`, and manually process filepaths (see https://github.com/gradio-app/gradio/pull/2758#discussion_r1050046009). Would be good to refactor all of these to use a single library (there have already been issues because of inconsistencies between the implementations of `os.path.abspath` and `pathlib().resolve()` for example) -- preferably `pathlib`

PR #2887, opened 2022-12-26
Enforcing typing in `blocks.py` and `context.py `

abidlabs:
Fixing typing issues (#2832) and path-related refactoring (#2831) for the following files:

* `blocks.py`
* `context.py`

Adding typing has already discovered some bugs, such as [a function that would](https://github.com/gradio-app/gradio/blob/88c920615991522453ebb8aac992b3b733aa311c/gradio/blocks.py#L767) `raise False` in place of `return False`!

Adds `pyright` to the CircleCI workflow to enforce typing in these two files. I used `pyright` because it was working more reliably than `mypy` for me in WSL and because it is already integrated into VSCode using the Pylance extension

Before (`blocks.py`): 

image

After:

image

I had to touch other files in some cases to fix typing as it related to one of the two files above, but those other files are not fully correctly-typed yet. Also renamed `queue.py` to `queuing.py` and `dataclasses.py` to `datacls.py` because `queue` and `dataclasses` are built-in modules in Python, which was confusing the type checker


gradio/utils.py
```diff
@@ -292,7 +295,7 @@ def format_ner_list(input_string: str, ner_groups: Dict[str : str | int]):
     return output
 
 
-def delete_none(_dict, skip_value=False):
+def delete_none(_dict: T, skip_value: bool = False) -> T:
```
freddyaboulton, line 298: Isn't this type too permissive?

gradio/components.py
```diff
@@ -273,12 +273,13 @@ def as_example(self, input_data):
...
 class FormComponent:
-    expected_parent = Form
+    def get_expected_parent(self) -> Type[Form]:
+        return Form
 
 
 @document("change", "submit", "blur", "style")
 class Textbox(
-    Changeable, Submittable, Blurrable, IOComponent, SimpleSerializable, FormComponent
+    FormComponent, Changeable, Submittable, Blurrable, IOComponent, SimpleSerializable
```
freddyaboulton, line 282: Why did we need to change this order?

scripts/type_check_backend.sh
```diff
@@ -0,0 +1,8 @@
+cd "$(dirname ${0})/.."
+source scripts/helpers.sh
+
+pip_required
+
+pip install --upgrade pip
+pip install pyright
+pyright gradio/context.py  gradio/blocks.py
```
freddyaboulton, line 8: Nice! Let's file issues for adding typing to the other modules?

gradio/routes.py
```diff
@@ -37,10 +37,10 @@
 
 import gradio
 from gradio import encryptor, utils
-from gradio.dataclasses import PredictBody, ResetBody
+from gradio.datacls import PredictBody, ResetBody
```
freddyaboulton, line 40: What about renaming to `data_classes`? I feel like the abbreviation is a bit awkward

gradio/blocks.py
```diff
@@ -250,6 +246,10 @@ def get_config(self):
             "root_url": self.root_url,
         }
 
+    @staticmethod
```
freddyaboulton, line 249: If this is the case, then the following would be clearer, no?

```python
@staticmethod
@abstractmethod
```

I'm thinking about custom components. Making it an abstract method would make it clear to devs that the method has to be implemented.

freddyaboulton (approved):
Awesome PR @abidlabs !! Will be very helpful as more contributors contribute to gradio as well.

I went through the deployed demos as well and everything looks good from the file refactor. If that's the case after merging in latest `main`, feel free to merge!

abidlabs:
Thank you for the review!

gradio/routes.py
abidlabs: Yea agreed

scripts/type_check_backend.sh
abidlabs: I'll just go through the rest in the next few days ;)

gradio/components.py
abidlabs: Ah so I made `get_expected_parent()` a method that is defined both in the class `Block` and in the class `FormComponent`. For the one in `FormComponent` to override the one in `Block`, it has to come before `IOComponent`.

abidlabs: I don't love this btw because it is kind of error-prone but I'll look for a way to improve when I work on `components.py`

gradio/utils.py
abidlabs: Yeah it is technically but I need a way of saying that the output type is the exact same as the input type (otherwise, type checker complains elsewhere). I couldn't find a way of saying that *and* that the type can be one of List, Dict, etc. If you have other suggestions, lmk. Otherwise, will revisit when I come to `utils.py`

gradio/blocks.py
abidlabs: Sounds good!

abidlabs:
All the demos look good to me!

PR #2889, opened 2022-12-27
Fixes error message display when raising errors from a generator function

JaySmithWpg:
# Description

When an error is raised in a function that returns iterative output to Gradio, the error is not correctly displayed in the user interface. The message is never delivered to the browser.

This function correctly displays the modal error message in the browser:
```python
def foo():
    raise gr.Error("Bar") # This will appear in the browser
```
But if the function is a generator that yields any results before the error is raised, the error message never appears in the browser:
```python
def foo():
    yield {status_text: "It's about to fail"}
    raise gr.Error("Bar") # This will not appear in the browser
```

Console output, thankfully, is not impacted by this bug.

This happens because Queue.process_events does not check the status of the last message received. It gets the status of the second last message received.

This PR changes Queue's behaviour to check the final message status and use that status to determine which message should be used to communicate the end of the generator to the browser.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 
 ## Bug Fixes:
 * Fixed bug where setting `default_enabled=False` made it so that the entire queue did not start by [@freddyaboulton](https://github.com/freddyaboulton) in [PR 2876](https://github.com/gradio-app/gradio/pull/2876)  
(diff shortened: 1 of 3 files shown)


abidlabs:
Thanks for this contribution @JaySmithWpg, including the test! Testing it out now

abidlabs:
LGTM, thanks @JaySmithWpg for fixing this issue in the library 🧡

PR #2895, opened 2022-12-27
Deploy PRs from forks to spaces

freddyaboulton:
# Description

Deploy PRs from forks to spaces

## Motivation
One of our goals next year is to increase the number of outside contributors to gradio.

In order to do that sustainably, we should deploy pull requests from forks to spaces to make sure regressions/bugs are not introduced.

## Implementation
Previously we didn't deploy prs from forks because running the deploy action required using repo secrets. This PR fixes it by doing the deployment in two separate actions:

1. First action builds a wheel from the PR and uploads it to github as an artifact. This is done on `pull-request` target, which means no secrets are copied over to the fork.
2. Second action uploads the wheel. This action will run on the `main` branch of the `gradio-app/gradio` repo and not on the developer's fork. This will prevents developers from accessing our secrets.

## gradio-pr-bot
I created a new github account called @gradio-pr-bot so that comments on PRs come from that account instead of the generic "github-actions". Much nicer as the gradio logo will show up on the comments.

I created the `COMMENT_TOKEN` secret so that comments on prs show up from @gradio-pr-bot

As a follow up, I will rework the notebook check action to use the `COMMENT_TOKEN` so that comments come from @gradio-pr-bot there too.

## Testing

I have been testing this on my fork. The `Deploy PR to Spaces` action in this PR must be on the `main` branch for it to run so I would not have been able to test it on this repo without spamming commits directly to main 😂 

Here is a pr from a branch of my fork showing this workflow working: https://github.com/freddyaboulton/gradio/pull/22

Here is a pr from a fork of my fork showing that the spaces are deployed correctly for forks: https://github.com/freddyaboulton/gradio/pull/23


freddyaboulton:
All the demos for this PR have been deployed at https://huggingface.co/spaces/gradio-pr-deploys/pr-2895-all-demos

.github/workflows/build-pr.yml
```diff
@@ -31,35 +30,33 @@ jobs:
...
-      run: pip install huggingface-hub
+    - name: Upload wheel
+      uses: actions/upload-artifact@v3
+      with:
+        name: gradio-${{ env.GRADIO_VERSION }}-py3-none-any.whl
+        path: dist/gradio-${{ env.GRADIO_VERSION }}-py3-none-any.whl
     - name: Set up Demos
       run: python scripts/copy_demos.py https://gradio-builds.s3.amazonaws.com/${{ github.sha }}/gradio-${{ env.GRADIO_VERSION }}-py3-none-any.whl
-    - name: Upload kitchen sink to spaces
+    - name: Upload all_demos
```
freddyaboulton, line 50: Uploading all_demos as well so that we deploy the code from the branch/fork

scripts/download_artifacts.py
```diff
@@ -0,0 +1,56 @@
+import argparse
```
freddyaboulton, line 1: Using the github rest api to download the artifacts created in the "build pr artifacts" workflow.

There is [technically](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#using-data-from-the-triggering-workflow) a way to do it directly on the github actions but that requires using js and it's harder to test/debug in my opinion. I couldn't get it to work.

abidlabs: 👍

.github/workflows/deploy-pr-to-spaces.yml
```diff
@@ -0,0 +1,65 @@
...
+    steps:
+    - uses: actions/checkout@v3
+    - name: Install Python
+      uses: actions/setup-python@v3
+      with:
+        python-version: '3.9'
+    - name: Install pip
+      run: python -m pip install build requests
+    - name: Download metadata
+      run: python scripts/download_artifacts.py ${{github.event.workflow_run.id }} metadata.json ${{ secrets.COMMENT_TOKEN }} --owner freddyaboulton
```
abidlabs, line 24: For all of these calls to `scripts/download_artifacts.py`, why do you set `--owner freddyaboulton` instead of using the default `gradio-app` account?

.github/workflows/build-pr.yml
abidlabs: Nice

abidlabs:
Very cool idea @freddyaboulton! I had a question -- you mention that the second action runs on the `main` branch. But I don't see anything in `.github/workflows/deploy-pr-to-spaces.yml` that tells it to run on the `main` branch -- how is this accomplished?

.github/workflows/deploy-pr-to-spaces.yml
freddyaboulton: Need to change that, good catch!

freddyaboulton:
@abidlabs The workflow trigger is ‘workflow_run’, which only runs on the default branch. In our case, that’s main. 

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run

.github/workflows/deploy-pr-to-spaces.yml
freddyaboulton: Just pushed up a change to remove this. Should be good for another review!

abidlabs (approved):
Looked through it and everything looks good to me. This is really cool!

PR #2896, opened 2022-12-27
Added typing to more files

abidlabs:
Typing is now enforced on:

* `components.py` 
* `data_classes.py` 
* `deprecation.py` 
* `documentation.py`
* `encryptor.py`
* `events.py`

Also migrated from `os.path` to `pathlib` in these files


gradio/components.py
```diff
@@ -2758,8 +2766,11 @@ def postprocess(self, y: str | pd.DataFrame) -> Dict:
         if y is None:
             return None
         if isinstance(y, str):
-            y = pd.read_csv(y)
-            return {"headers": y.columns.values.tolist(), "data": y.values.tolist()}
+            dataframe = pd.read_csv(y)
```
freddyaboulton, line 2769: Interesting that the type checker does not recognize that `y` is now a dataframe in this case.

freddyaboulton (approved):
Thank you for making all these changes @abidlabs ! Looks good to me

gradio/components.py
```diff
@@ -2278,12 +2283,12 @@ def process_single_file(f):
                     temp_file_path = self.make_temp_copy_if_needed(file_name)
                     file = tempfile.NamedTemporaryFile(delete=False)
                     file.name = temp_file_path
-                    file.orig_name = file_name
+                    file.orig_name = file_name  # type: ignore
```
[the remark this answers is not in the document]
abidlabs: The type checker complains when an attribute is added to an object after it is constructed. I think in order to fix this, we would have to subclass `tempfile.NamedTemporaryFile` to return a class that includes an `orig_name` attribute. It seems unnecessary at this point so I just ignored the type hint

gradio/components.py
abidlabs: Yes these type checkers is not very good when the same variable is reassigned to a variable of a different type. Anyways, I think this change makes the code a little more readable so I just made the change

abidlabs:
Really appreciate the review @freddyaboulton!

PR #2898, opened 2022-12-28
Fix "audio used before assigned" bug

FelixDombek:
`audio_from_file` did not correctly handle the case where the file cannot be opened. This PR fixes e.g. the crash mentioned in [TheAlly's Riffusion tutorial](https://theally.notion.site/theally/Quick-n-Dirty-Riffusion-txt2audio-Tutorial-18e57df9ef214c3280efc5998bbf774d).

# Description

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

Closes: # (issue)

# Checklist:

# A note about the CHANGELOG

### DIFF
--- a/gradio/processing_utils.py
+++ b/gradio/processing_utils.py
@@ -189,8 +189,8 @@ def audio_from_file(filename, crop_min=0, crop_max=100):
     except FileNotFoundError as e:
         isfile = os.path.isfile(filename)
         msg = (f"Cannot load audio from file: `{'ffprobe' if isfile else filename}` not found."
-               + "Please install `ffmpeg` in your system to use non-WAV audio file formats "
-                 "and make sure `ffprobe` is in your PATH." if isfile else "")
+               + " Please install `ffmpeg` in your system to use non-WAV audio file formats"
+                 " and make sure `ffprobe` is in your PATH." if isfile else "")
         raise RuntimeError(msg) from e
     if crop_min != 0 or crop_max != 100:
         audio_start = len(audio) * crop_min / 100
--- a/gradio/processing_utils.py
+++ b/gradio/processing_utils.py
@@ -187,12 +187,14 @@ def audio_from_file(filename, crop_min=0, crop_max=100):
     try:
         audio = AudioSegment.from_file(filename)
     except FileNotFoundError as e:
-        error_message = str(e)
-        if "ffprobe" in error_message:
+        isfile = os.path.isfile(filename)
+        print(f"Cannot load audio from file: `{'ffprobe' if isfile else filename}` not found.")
+        if isfile:
             print(
-                "Please install `ffmpeg` in your system to use non-WAV audio file formats."
+                "Please install `ffmpeg` in your system to use non-WAV audio file formats "
+                "and make sure `ffprobe` is in your PATH."
             )
-        return 0, np.empty(0)
+        raise
     if crop_min != 0 or crop_max != 100:
         audio_start = len(audio) * crop_min / 100
         audio_end = len(audio) * crop_max / 100
--- a/gradio/processing_utils.py
+++ b/gradio/processing_utils.py
@@ -188,9 +188,9 @@ def audio_from_file(filename, crop_min=0, crop_max=100):
         audio = AudioSegment.from_file(filename)
     except FileNotFoundError as e:
         isfile = os.path.isfile(filename)
-        msg = (f"Cannot load audio from file: `{'ffprobe' if isfile else filename}` not found." +
-                "Please install `ffmpeg` in your system to use non-WAV audio file formats "
-                "and make sure `ffprobe` is in your PATH." if isfile else "")
+        msg = (f"Cannot load audio from file: `{'ffprobe' if isfile else filename}` not found."
+               + "Please install `ffmpeg` in your system to use non-WAV audio file formats "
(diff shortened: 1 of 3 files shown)


gradio/processing_utils.py
```diff
@@ -192,6 +192,7 @@ def audio_from_file(filename, crop_min=0, crop_max=100):
             print(
                 "Please install `ffmpeg` in your system to use non-WAV audio file formats."
             )
+        return 0, np.empty(0)
```
freddyaboulton, line 195: Thanks @FelixDombek ! Wouldn't it be better to raise the exception in this case? The user would get an empty audio and not know why.

FelixDombek: You're probably right. If I replace the line with a simple `raise` , I get:
```
C:\ai\stable-diffusion-webui\venv\lib\site-packages\pydub\utils.py:198: RuntimeWarning: Couldn't find ffprobe or avprobe - defaulting to ffprobe, but may not work
  warn("Couldn't find ffprobe or avprobe - defaulting to ffprobe, but may not work", RuntimeWarning)
  File "C:\ai\stable-diffusion-webui\venv\lib\site-packages\gradio\routes.py", line 284, in run_predict
    output = await app.blocks.process_api(
  File "C:\ai\stable-diffusion-webui\venv\lib\site-packages\gradio\blocks.py", line 980, in process_api
  File "C:\ai\stable-diffusion-webui\venv\lib\site-packages\gradio\blocks.py", line 891, in preprocess_data
  File "C:\ai\stable-diffusion-webui\venv\lib\site-packages\gradio\components.py", line 1845, in preprocess
    sample_rate, data = processing_utils.audio_from_file(
  File "C:\ai\stable-diffusion-webui\venv\lib\site-packages\gradio\processing_utils.py", line 193, in audio_from_file
    audio = AudioSegment.from_file(filename)
  File "C:\ai\stable-diffusion-webui\venv\lib\site-packages\pydub\audio_segment.py", line 728, in from_file
    info = mediainfo_json(orig_file, read_ahead_limit=read_ahead_limit)
  File "C:\ai\stable-diffusion-webui\venv\lib\site-packages\pydub\utils.py", line 274, in mediainfo_json
    res = Popen(command, stdin=stdin_parameter, stdout=PIPE, stderr=PIPE)
  File "C:\Program Files\Python310\lib\subprocess.py", line 969, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "C:\Program Files\Python310\lib\subprocess.py", line 1438, in _execute_child
    hp, ht, pid, tid = _winapi.CreateProcess(executable, args,
FileNotFoundError: [WinError 2] The system cannot find the file specified.
```
The file that isn't found is actually ffprobe. The first line also tells us it's missing. Just putting this into PATH makes the crash go away. (I wonder what is going on here, since Riffusion works just fine without this huge dependency except for this one (circumventable) crash. Maybe it can be fixed to avoid this call path altogether, at least I don't understand it's purpose.) 

It looks as if the code here was written to look for this condition, but the error message is different on Windows and doesn't contain `"ffprobe"`. So Gradio doesn't output the useful hint (line 193). 

**Should we just make the output unconditional? Something like "Make sure ffprobe is in your PATH". WDYT?**

FelixDombek: I made a commit to that effect, let me know what you think. :)

gradio/processing_utils.py
```diff
@@ -187,11 +187,14 @@ def audio_from_file(filename, crop_min=0, crop_max=100):
     try:
         audio = AudioSegment.from_file(filename)
     except FileNotFoundError as e:
-        error_message = str(e)
-        if "ffprobe" in error_message:
+        isfile = os.path.isfile(filename)
+        print(f"Cannot load audio from file: `{'ffprobe' if isfile else filename}` not found, {e}.")
```
freddyaboulton, line 191: Thanks for the change @FelixDombek !

Why not `raise ValueError(<text you added in this pr>)` ? 

Happy to merge once this is resolved.

gradio/processing_utils.py
freddyaboulton: Thank you for the change! Maybe it's better if the exception you raise contains the helpful text you added as opposed to the print. I'm worried that text will be lost in the console when the original exception is raised. Let me know what you think.

gradio/processing_utils.py
freddyaboulton: The added benefit of adding this text to the exception is that if the demo is launched with `show_error=True`, this helpful error message will show up in the client. Making it much more actionable.

image

FelixDombek: The `ValueError` doc says it's used when "an argument has the right type but the wrong value", but ffprobe isn't an argument. I think it would be nicer to pass on the FileNotFoundError with the additional info, let me try that ...

freddyaboulton: Sounds good!

FelixDombek: The Python docs raise a RuntimeError in this case (https://docs.python.org/3/tutorial/errors.html#exception-chaining), so that's what I went with now.

freddyaboulton (approved):
Thank you @FelixDombek ! Will merge once CI passes

abidlabs:
Thanks for the contribution to the library @FelixDombek!

PR #2900, opened 2022-12-28
Added typing to even more files 

abidlabs:
* `examples.py`
* `exceptions.py`
* `external.py`
* `external_utils.py`
* `serializing.py`
* `layouts.py`
* `flagging.py`


gradio/components.py
```diff
@@ -2353,6 +2353,7 @@ def serialize(
         self, x: str, load_dir: str = "", encryption_key: bytes | None = None
     ) -> Dict:
         serialized = FileSerializable.serialize(self, x, load_dir, encryption_key)
+        assert serialized is not None
```
freddyaboulton, line 2356: Is this to make the type checker happy? Can we make a change to `serialize` so that we don't have to add these asserts?

abidlabs: Hmmmm I think so. Let me check

abidlabs: Actually `serialized` _can_ be `None` if the input `x` is `None`. Handled that case now

gradio/external_utils.py
```diff
@@ -41,7 +46,7 @@ def get_tabular_examples(model_name: str) -> Dict[str, List[float]]:
     # replace nan with string NaN for inference API
     for data in example_data.values():
         for i, val in enumerate(data):
-            if isinstance(val, numbers.Number) and math.isnan(val):
+            if isinstance(val, float) and math.isnan(val):
```
aliabid94, line 49: not possible for this to be int?

gradio/flagging.py
```diff
@@ -51,9 +52,9 @@ def _get_dataset_features_info(is_new, components):
                 headers.append(component.label + " file")
                 for _component, _type in file_preview_types.items():
                     if isinstance(component, _component):
-                        infos["flagged"]["features"][component.label + " file"] = {
-                            "_type": _type
-                        }
+                        infos["flagged"]["features"][
+                            component.label or "" + " file"
```
aliabid94, line 56: `(component.label or "") + " file"` 
not sure if order of operations does that already but better to be explicit

aliabid94 (approved):
left minor comments, lgtm otherwise

gradio/flagging.py
abidlabs: Sg thanks

gradio/external_utils.py
abidlabs: It is but an int would never be nan so the second part of the condition would fail

Issue #2776, opened 2022-12-08
Upload filetypes of specified extensions.


davidbernat:
I would like my application to use a proprietary text file format using a non-standard extension. The text file is still a simple text file, and the non-standard extension is used internally by my software to specify an expected formatting of the text. To my knowledge, `UploadFile` and `File` do not allow specifying what extensions to accept, only to specific from an internally represented list of text file extensions using the `file_types['text']` parameter. Is this accurate? If so, I request the feature. Thank you.

abidlabs:
Thanks for the suggestion @davidbernat, agreed that this would be good to have.

PR #2901, opened 2022-12-28
Upload filetypes of specified extensions

dawoodkhan82:
# Description
Add support for specifying file extensions for `gr.File` and `gr.UploadButton`, using `file_types` parameter (e.g  `gr.File(file_count="multiple", file_types=["text", ".json", ".csv"])`

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

Closes: #2776 

# Checklist:

# A note about the CHANGELOG


ui/packages/file/src/FileUpload.svelte
```diff
@@ -44,7 +44,13 @@
...
-		file_types.forEach((type) => (accept_file_types += type + "/*, "));
+		for (const type of file_types) {
+			if (type.includes(".")) {
+				accept_file_types += type + ", ";
+			} else {
+				accept_file_types += type + "/*, ";
+			}
+		}
 	} catch (err) {
 		if (err instanceof TypeError) {
```
aliabid94, line 55: this try/catch is weird. The backend allows the user to send either a list of strings or None (which is the default value of file_types). In the frontend, we should allow file_types to be type `file_types: string[] | null = null`. We should explicitly handle the case where file_types is null, not wrap it in this try/catch. 
If `file_types` == null, then `accept_file_types` should also equal null (not empty string), because later we set `accept={accept_file_types}` and we want to remove the `accept=` attribute entirely if we're not filtering for file types. In svelte, this is done by passing null to an attribute.

ui/packages/upload-button/src/UploadButton.svelte
```diff
@@ -18,7 +18,13 @@
 	const dispatch = createEventDispatcher();
 	let accept_file_types = "";
 	try {
```
aliabid94, line 20: Again same here.

ui/packages/upload-button/src/UploadButton.svelte
```diff
@@ -18,7 +18,13 @@
...
-		file_types.forEach((type) => (accept_file_types += type + "/*, "));
+		for (const type of file_types) {
+			if (type.includes(".")) {
```
aliabid94, line 22: more accurate to say `type.startswith(".")`?

ui/packages/file/src/FileUpload.svelte
```diff
@@ -44,7 +44,13 @@
 
 	let accept_file_types = "";
 	try {
-		file_types.forEach((type) => (accept_file_types += type + "/*, "));
+		for (const type of file_types) {
+			if (type.includes(".")) {
+				accept_file_types += type + ", ";
```
aliabid94, line 49: there will be a trailing comma at the end if you're adding a comma after every type, I don't know how that is supported with every browser. Would be cleaner to create a list and use `list.join(", ")` to only insert commas in between.

dawoodkhan82:
@aliabid94 fixed all comments. Ready for re-review

Issue #2851, opened 2022-12-20
Split the `Interface` class into separate classes for the different kinds of Interfaces


abidlabs:
Right now, the `Interface` class is very convoluted with many if conditions, etc. because we create slightly different layouts depending on if the Interface is:

* standard interface
* input-only interface
* output-only interface
* a "united" interface where the input is the output

We should refactor this code to have separate classes for each of these types to make the code easier to maintain and understand.

freddyaboulton:
This is an internal refactoring right? 

There will still only be one `Interface` class but internally it will delegate to `OutputOnlyInterface`, `StandardInterface`, etc?

abidlabs:
Yes exactly, this won't be a user-facing change

PR #2902, opened 2022-12-29
Refactored `interface.py`

abidlabs:
Refactored `interface.py` to add typing and clean up the rendering logic. I don't think there is any need for #2851 so we can close it. Closes: #2851


gradio/blocks.py
```diff
@@ -519,8 +519,8 @@ def __init__(
         self,
         theme: str = "default",
         analytics_enabled: bool | None = None,
-        mode: str = "blocks",
-        title: str = "Gradio",
+        mode: str | None = "blocks",
```
aliabid94, line 522: why allow `blocks` and `mode` to be null? Not clear what null even means in this case.

aliabid94 (approved):

gradio/blocks.py
abidlabs: Sounds good, will change

PR #2906, opened 2022-12-29
More typing!

abidlabs:
Done:
* `mix.py` 
* `networking.py`
* `pipelines.py`
* `processing_utils.py`
* `reload.py`
* `strings.py`
* `tunneling.py`
* `queueing.py`
* `serializing.py`
* `routes.py`


gradio/mix.py
```diff
@@ -32,7 +32,7 @@ def __init__(self, *interfaces: gradio.Interface, **options):
         Returns:
             an Interface object comparing the given models
         """
-        outputs: List[IOComponent] = []
+        outputs: List[Component] = []
```
aliabid94, line 35: It does have to be an IOComponent, right?

gradio/mix.py
```diff
@@ -44,7 +44,7 @@ def __init__(self, *interfaces: gradio.Interface, **options):
 
         async def parallel_fn(*args):
             return_values_with_durations = await asyncio.gather(
-                *[interface.call_function(0, args) for interface in interfaces]
+                *[interface.call_function(0, list(args)) for interface in interfaces]
```
aliabid94, line 47: why recast as list, isn't it already a list? calling list is making a shallow copy for no reason?

gradio/routes.py
```diff
@@ -379,61 +375,60 @@ async def predict(
 
             # If this fn_index cancels jobs, then the only input we need is the
             # current session hash
-            if app.blocks.dependencies[body.fn_index]["cancels"]:
+            if app.get_blocks().dependencies[body.fn_index]["cancels"]:
                 body.data = [body.session_hash]
             if body.request:
                 if body.batched:
-                    request = [Request(**req) for req in body.request]
+                    request_ = [Request(**req) for req in body.request]
```
aliabid94, line 382: I've always seen the underscore come first when avoiding another namespace, e.g. `_id` instead of `id`.

aliabid94 (approved):
lgtm

gradio/mix.py
abidlabs: Yeah I'm actually going to just remove this type hint

gradio/mix.py
abidlabs: It's actually a tuple not a list :)

gradio/routes.py
abidlabs: Sure thanks

abidlabs:
Thanks for all of the reviews @aliabid94!

Issue #2872, opened 2022-12-21
Refactor interpretation so that no interpretation-specific code lives inside the `Component` class


abidlabs:
Thinking ahead to custom components and the 4.0 launch, we will want to simplify the components classes, and one way we can do that is to remove interpretation-specific functions (like `get_interpretation_neighbors`, etc.) from all of the component classes that currently have them. This is just a suggestion but I think it will be helpful in keeping the code for the components as simple as possible.

freddyaboulton:
This is an interesting point when it comes to custom components. 

If we keep `get_interpretation_neighbors` as a method on the component class, then the interpretation module doesn't need to know anything about the component, it can just call that method. That will make it easier to integrate custom components with interpretation, imo.

The other approach would be to "centralize" all the logic in the interpretation module and call the right `get_neighbors` method depending on the component class. That will make it harder to support custom components. 

The same thing will happen with `Dataset` and `Examples`. The Dataset class behaves differently depending on the components in the dataset, which will make it harder to support custom components.

I'm not opposed to this refactor, just pointing out that we need to think hard about the component interface (and stick to it) for 4.0 and beyond.

abidlabs:
Yes exactly, there's a tradeoff when it comes to centralization vs. keeping the code within each component class: on one hand the code for the components can be far simpler by centralizing. On the other hand, then the interpretation needs to know about each component separately.

My 2 cents (but def open to changing my mind) is that for examples, it's better to have the code within each component class because:
(1) Each component really should implement `def as_example()` or user the default implementation because examples are used very frequently.
(2) The implementation of `def as_example()` is usually quite straightforward

However, interpretation is not widely used and its implementation can be quite complex so I would suggest keeping it out of the main component classes.

PR #2911, opened 2022-12-31
Final typing: `interpretation.py` and `helpers.py`

abidlabs:
This adds typing to `interpretation.py` and moves some of the interpretation-related helper methods from components to `interpretation.py`. This is the final typing-related PR so:

Closes: #2832
Closes: #2831
Closes: #2872 (not really needed)


abidlabs:
Jk need to add typing to `helpers.py` too, will do that later today

aliabid94 (approved):
LGTM

abidlabs:
Ok just put the final typing touches to `helpers.py`. If you wanna take another look @aliabid94, that would be great

gradio/helpers.py
```diff
@@ -433,7 +438,8 @@ def __call__(
             total: estimated total number of steps.
             unit: unit of iterations.
         """
-        if self._active:
+        if self._callback:
+            assert self._callback is not None, "Callback not set."
```
aliabid94, line 442: this assertion can never be reached?

aliabid94 (approved):
lgtm

gradio/helpers.py
abidlabs: Oops meant to remove that, thanks for the catch

Issue #578, opened 2022-02-09
Optimise images on the website for loading in browsers


pngwn:
What it says on the tin. Check all images, make sure they are optimised. This should preferably be automated.

abidlabs:
@aliabd and I are pretty sure this is not an issue anymore since website performance is quite good now. Correct us if we are wrong @pngwn

Issue #577, opened 2022-02-09
Convert large GIFs into images


pngwn:
There is a least one _very_ large GIF. This should be a video which is  more appropriate way to present this content.

PR #2922, opened 2023-01-03
Optimize images and gifs 

aliabd:
Converts gifs to html video, and compresses some images. @pngwn I'm not 100% sure what you meant here https://github.com/gradio-app/gradio/issues/578, almost all the images in the website are pretty small. Or does optimizing mean something else? Please take a look and let me know if this is what you meant. 

Closes https://github.com/gradio-app/gradio/issues/577, closes https://github.com/gradio-app/gradio/issues/578


pngwn (approved):
I don't think this addresses #578 because we can ship different sized images for different screen sizes and optimise those automatically during build time. I think this would be nice to do since we don't need to worry about adding images to guides, we can just do it as we please and the build tooling will handle the optimisations for us.

That said I don't think it is a huge concern right now and this PR looks great to me!

pngwn:
It would be good to keep #578 open for visibility.

aliabd:
Got it, thanks @pngwn

PR #2950, opened 2023-01-08
Hotfix audio output, HighlightedText skip

aliabid94:
Two quick hotfixes - audio output wasn't working on mac bc of path.resolve, and highlightedtext was not working with gr.skip or gr.update


abidlabs (commented):
LGTM thanks for the fix @aliabid94

PR #2957, opened 2023-01-08
Fix validate url

freddyaboulton:
# Description

`ImgSerializable.serialize` would fail on an image url because `Path(load_dir) / x` would strip out some of the slashes in the url which would cause the url to be treated as a file.

Doing `Path(load_dir) / x ` basically assumes `x` is a filepath instead of a url but that has not been determined at the time that line is reached so I refactored the logic a bit so that `Path(load_dir) / x ` is only called if `x` is not a url.

This comes at the cost of two get requests to the external url but the benefit is that we don't have to modify any of the public apis of our util functions.

# Checklist:

# A note about the CHANGELOG


abidlabs:
This works great @freddyaboulton, nice catch!

Regarding the speed issue, we could change the `validate_url` method to use `requests.head()` instead of `requests.get()` so that it does not download the whole file when checking if the file exists:

```py
def validate_url(possible_url: str) -> bool:
    headers = {"User-Agent": "gradio (https://gradio.app/; team@gradio.app)"}
    try:
        return requests.head(possible_url, headers=headers).ok
    except Exception:
        return False
```

Thoughts? A quick benchmark shows that the time to run `validate_url` on a 5 MB file drops from 1s to 0.3s if we use `head()`:

```py
import requests
import time

t = time.time()
headers = {"User-Agent": "gradio (https://gradio.app/; team@gradio.app)"}
requests.get("https://edmullen.net/test/rc.jpg", headers=headers).ok
print(time.time() - t)

t = time.time()
requests.head("https://edmullen.net/test/rc.jpg", headers=headers).ok
print(time.time() - t)
```

freddyaboulton:
Thanks for the suggestion @abidlabs ! 

The one snag is that the gradio file route does not support head requests. I made that addition to the routes.py file but that means that we will first try `head` and if that's not supported we'll try with `get`.

Let me know what you think before I merge!

abidlabs:
Great catch @freddyaboulton, LGTM

abidlabs (approved):

freddyaboulton:
Thank you for the re-review!

Issue #2646, opened 2022-11-14
Audio component does not record for users on iOS 


abidlabs:
### Describe the bug

We've been getting reports that the Microphone does not seem to work for users on iOS, even once they provide permissions. Here are some of the reports:

* https://twitter.com/ManuelFaysse/status/1592125388452691968
* https://huggingface.co/spaces/fffiloni/whisper-to-stable-diffusion/discussions/4

I've also confirmed I see the problem with a simple repro:

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Try running this:

```py
import gradio as gr

gr.Interface(lambda x:x, "mic", "audio").launch(share=True)
```

And open the share link on an iOS device

### Logs

```shell
Hard to see the logs since it's on iOS
```

### System Info

```shell
gradio==3.9.1
```

### Severity

serious, but I can work around it

fffiloni:
Recording works fine i guess, as we can listen to the recorded audio from the html5 player, but it may be a format/codec issue from iOS 🤗

MaxBrodeurUrbas:
+1 for this issue. Audio recording fails consistently on iOS and a user of our gradio demo also reported the same issue when trying the demo on a macbook + chrome.

KamyMouss:
+1, doesn't work on my end either. Would appreciate a fix asap!

aliabid94:
Taking a look!

leobaumgardt:
+1.

Also, in my case, the recorded audio is silent. That is to say, if I record for 5 seconds, I will get a player with 5 seconds of recorded audio. But when I play it, I will hear nothing.

This is using iOS Safari via the "share" URL.

I tested record/playback on another site (https://addpipe.com/simple-web-audio-recorder-demo/), and there it works.

Perhaps it has something to do with iOS using "audio/webm"?

MaxBrodeurUrbas:
I've also had users report audio recording completely fails on older versions of MacOS (Catalina 10.15). Failing occurs regardless of browser and selecting example audio files from the gradio.example list is also broken. Not sure if this is relevant to the current bug.

abidlabs:
This should be solved soon with #2961, thanks @aliabid94!

PR #2961, opened 2023-01-09
Support media output, "/file=" endpoint on iOS

aliabid94:
Currently, audio/video examples and output were broken on iOS. This is because the `/file=` endpoint is not supported on iOS for audio / video files, because iOS safari sends "ranged requests" for audio/video files, which specify a range of bytes in the request header. FastAPI and starlette do not support ranged requests, despite many issues, see [this for context, code was also taken from PR discussion](https://github.com/encode/starlette/pull/1090).

Fixed by adding ranged request monkeypatch provided in link above.

Test this by running any demo with audio/video output and examples (e.g. reverse_audio, video_identity). They do not work before this PR.

Closes: #2646


gradio/ranged_response.py
```diff
@@ -0,0 +1,185 @@
+# Taken from https://gist.github.com/kevinastone/a6a62db57577b3f24e8a6865ed311463
+# Context: https://github.com/encode/starlette/pull/1090
+
+import os
+import re
+import stat
+import typing as t
```
abidlabs, line 7: To be consistent with other modules, I would just do `from typing import NamedTuple`, etc. 

We also don't use `typing.Union` and instead use `|`

gradio/ranged_response.py
```diff
@@ -0,0 +1,185 @@
...
+from starlette.datastructures import Headers
+from starlette.exceptions import HTTPException
+from starlette.responses import Response, guess_type
+from starlette.staticfiles import StaticFiles
+from starlette.types import Receive, Scope, Send
+
+RANGE_REGEX = re.compile(r"^bytes=(?P<start>\d+)-(?P<end>\d*)$")
+
+
+PathLike = t.Union[str, "os.PathLike[str]"]
```
abidlabs, line 21: You don't need quotes around types anymore I believe

abidlabs:
Very nice @aliabid94, this looks good to me. Given that this touches all of the files that we serve through the `/file=`, I would definitely add some tests for this. There's some typing issues that's causing the ci to fail, but otherwise LGTM

abidlabs:
Hmm actually I'm getting an error when running `python demo/kitchen_sink_random/run.py`: 

```
  File "C:\Users\islam\AppData\Local\Programs\Python\Python39\lib\site-packages\fastapi\routing.py", line 159, in run_endpoint_function
    return await dependant.call(**values)
  File "c:\users\islam\dev\gradio-repos\gradio\gradio\routes.py", line 269, in file
    range_val = request.headers["Range"].strip()
  File "C:\Users\islam\AppData\Local\Programs\Python\Python39\lib\site-packages\starlette\datastructures.py", line 542, in __getitem__
    raise KeyError(key)
KeyError: 'Range'
```

And the Gallery component isn't rendering correctly:

image

gradio/ranged_response.py
aliabid94: I didn't edit this file because it's copy-pasted from the github link at the top of the file - it's almost like a "third-party" library. Would think it's better not to touch third party code?

aliabid94: have moved to folder called "third_party", which is ignored by style and codecov tests.

gradio/ranged_response.py
aliabid94: see above, this is third-party code

aliabid94:
fixed mistake, added tests, ready for rereview @abidlabs

gradio/ranged_response.py
pngwn: It isn't third-party if we bring it into the repo manually, so it meeds meet the same code standards as the rest of the code-base, since we need to maintain it.

aliabid94: hm okay if that's standard practice, that makes sense

aliabid94: fixed

gradio/ranged_response.py
aliabid94: nvm above comment, fixed

freddyaboulton:
Thanks @aliabid94 !

Maybe something is wrong with Files + examples?

In the deployed demo I clicked on the example for `zip_files` and the whole app crashed

image

```
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'length')
    at j (utils.ts:17:10)
    at H (FileUpload.svelte:91:8)
    at je (FileUpload.svelte:88:6)
    at Array.ze (FileUpload.svelte:87:22)
    at Object.p (FileUpload.svelte:63:54)
    at ql (index.f82c080a.js:4:1891)
    at bt (index.f82c080a.js:4:1560)
```

freddyaboulton:
Oops my bad - looks like my comment above is caused by the example loading logic should be fixed in @2955!

MaxBrodeurUrbas:
Hello! Thank you for addressing this issue so quickly!

I've tried the "reverse audio" demo for this PR on an iPhone Xs (iOS 16.1.1) and an iPhone 14 (iOS 16.2) on both Safari and Chrome. Audio recording still seemed to be broken. The audio recorded didn't play back and the reversed audio has a 0 length. 

Perhaps I was testing it incorrectly but  wanted to give a heads up in case!

abidlabs:
@MaxBrodeurUrbas, that's interesting. The "reverse audio" tab here: https://huggingface.co/spaces/gradio-pr-deploys/pr-2961-all-demos is working for me on an iPhone 12 with iOS v.14. 

Can you confirm that you are recording an audio clip (and not selecting an example -- that doesn't seem to be working, but for a different reason).

abidlabs:
I'll go ahead and merge this in. @MaxBrodeurUrbas if you continue to see this issue after we release, please feel free to open another issue

MaxBrodeurUrbas:
> I'll go ahead and merge this in. @MaxBrodeurUrbas if you continue to see this issue after we release, please feel free to open another issue

Sorry for the delayed reply! I am still experiencing the issue and can link a screen recording reproducing the error in a few minutes..

abidlabs:
Thanks for the repro @MaxBrodeurUrbas. I'm going to create a separate issue with your bug to investigate further. I'll merge in this PR since it does fix the bug at least on some iOS devices

Issue #1756, opened 2022-07-11
Document the four different kinds of interfaces


freddyaboulton:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
As a gradio user, I wish the four different kinds of interfaces were covered in the documentation and/or guides.

**Describe the solution you'd like**  
I think it makes sense to briefly mention this concept in the Docs but then link to a guide. This should go in the advanced interface features guide or a new guide. May be best to include in it's own guide.

**Additional context**  
https://twitter.com/abidlabs/status/1527738833537683456

PR #3003, opened 2023-01-17
Guide on 4 kinds of interfaces

abidlabs:
Writing a Guide on the 4 kinds of Interfaces with @yvrjsharma 

Closes: #1756


abidlabs:
@yvrjsharma the updates to the Guide look good to me. One thing though -- since we're showing the 4 kinds of *interfaces*, it would be better to rewrite the `unified_demo_text_generation` to be a `gr.Interface` instead of a `gr.Blocks`.  Something along these lines should work:

```py
import gradio as gr

def generate_text(text_prompt):
    return "..."

textbox = gr.Textbox()

demo = gr.Interface(generate_text, textbox, textbox).launch()

if __name__ == "__main__":
    demo.launch()
```

guides/02_building-interfaces/05_four-kinds-of-interfaces.md
```diff
@@ -0,0 +1,47 @@
...
+$code_sepia_filter
+$demo_sepia_filter
+
+
+## Output-only demos
+
+What about demos that only contain outputs? In order to build such a demo, you simply set the value of the `inputs` parameter in `Interface()` to `None`. Here's an example demo of a mock image generation model:
+
+$code_fake_gan_no_input
+$demo_fake_gan_no_input
```
aliabd, line 33: This demo is producing really huge images (taking up insane amount of the page) I looked into it and super easy to fix. Just modify the width param on the urls it's picking from (unsplash urls)

demo/unified_demo_text_generation/run.py
```diff
@@ -0,0 +1,18 @@
+from transformers import pipeline
```
aliabd, line 1: This demo needs a requirements.txt file with transformers and torch in it. Generally a good idea to try uploading the demo to spaces to see if it will work on the website.

demo/save_file_no_output/run.py
```diff
@@ -0,0 +1,16 @@
...
+    image.save(random_string)
+    print(f"Saved image to {random_string}!")
+
+demo = gr.Interface(
+    fn=save_image_random_name, 
+    inputs=gr.Image(type="pil"), 
+    outputs=None,
+)
+if __name__ == "__main__":
+    demo.launch()
```
aliabd, line 16: There's no feedback at all to the user in this demo. Maybe that's the point, but it can get a little confusing. I think the image should at least disappear. Could be done with blocks and a clear event I suppose, maybe not worth it so will leave it up to you. But try it out here if you haven't to see what I mean, just feels like button isn't doing anything: https://huggingface.co/spaces/gradio/save_file_no_output

guides/02_building-interfaces/05_four-kinds-of-interfaces.md
```diff
@@ -0,0 +1,47 @@
...
+2. **Output-only demos**: which don't take any input but produce on output (e.g. an unconditional image generation model)
+3. **Input-only demos**: which don't produce any output but do take in some sort of input (e.g. a demo that saves images that you upload to a persistent external database)
+4. **Unified demos**: which have both input and output components, but the input and output components *are the same*. This means that the output produced overrides the input (e.g. a text autocomplete model)
+
+Depending on the kind of demo, the user interface (UI) looks slightly different:
+
+![](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/gradio-guides/interfaces4.png)
+
+
+Let's see to build each kind of demo using the `Interface` class, along with examples:
```
aliabd, line 17: typo 

> Let's see to build

Let's see *how to build

guides/02_building-interfaces/05_four-kinds-of-interfaces.md
```diff
@@ -0,0 +1,47 @@
...
+
+
+## Standard demos
+
+To create a demo that has both the input and the output components, you simply need to set the values of the `inputs` and `outputs` parameter in `Interface()`. Here's an example demo of a mock image classification model:
```
aliabd, line 22: Below demo is an image-to-image demo, so not really image classification at all, would just call it something else.

demo/unified_demo_text_generation/run.py
yvrjsharma: Thanks Ali for all the comments, I have added the requirement.txt file.

demo/save_file_no_output/run.py
abidlabs: That's the point though. If we rewrote it with Blocks, it would no longer be about the kinds of Interfaces

guides/02_building-interfaces/05_four-kinds-of-interfaces.md
abidlabs: Fixed

guides/02_building-interfaces/05_four-kinds-of-interfaces.md
abidlabs: Good catch

guides/02_building-interfaces/05_four-kinds-of-interfaces.md
abidlabs: Fixed

abidlabs:
Thanks for the awesome review @aliabd and for working together with me on this @yvrjsharma. Merging

Issue #2920, opened 2023-01-03
`gr.Interface.load()` not working with image as the output


apolinario:
### Describe the bug

Trying to load the following Space https://huggingface.co/spaces/kakaobrain/karlo
```py
karlo = gr.Interface.load("kakaobrain/karlo", src="spaces")
prompt="a mecha robot in a favela"
images_result = karlo(prompt)
```

It seems the function actually runs remotely in the requested Space, but something wrong happens when processing the output, after it returns the result - some missing temp file:

```
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/tmp336s5u4z.png'
```

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Full traceback: 
```
Traceback (most recent call last):
  File "/home/poli/miniconda3/envs/multimodal/lib/python3.9/site-packages/gradio/routes.py", line 325, in run_predict
    output = await app.blocks.process_api(
  File "/home/poli/miniconda3/envs/multimodal/lib/python3.9/site-packages/gradio/blocks.py", line 983, in process_api
    result = await self.call_function(
  File "/home/poli/miniconda3/envs/multimodal/lib/python3.9/site-packages/gradio/blocks.py", line 821, in call_function
    prediction = await anyio.to_thread.run_sync(
  File "/home/poli/miniconda3/envs/multimodal/lib/python3.9/site-packages/anyio/to_thread.py", line 28, in run_sync
    return await get_asynclib().run_sync_in_worker_thread(func, *args, cancellable=cancellable,
  File "/home/poli/miniconda3/envs/multimodal/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 818, in run_sync_in_worker_thread
    return await future
  File "/home/poli/miniconda3/envs/multimodal/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 754, in run
    result = context.run(func, *args)
  File "/mnt/c/Users/poli/Documents/GitHub/karlo-upscaler/app.py", line 6, in run
    results = karlo(prompt)
  File "/home/poli/miniconda3/envs/multimodal/lib/python3.9/site-packages/gradio/blocks.py", line 776, in __call__
    processed_outputs = self.deserialize_data(fn_index, outputs)
  File "/home/poli/miniconda3/envs/multimodal/lib/python3.9/site-packages/gradio/blocks.py", line 872, in deserialize_data
    deserialized = block.deserialize(outputs[o])
[...]

PR #3004, opened 2023-01-17
Fixing deserialize when loading external Spaces

abidlabs:
There was a bug (reported in #2920) where users would not be able to use external Spaces as functions if the Spaces returned any sort of file. This fixes that and adds a unit test. You can test by running: 

```py
from pathlib import Path
import gradio as gr

demo = gr.Interface.load("spaces/gradio/simple_gallery")
```

Closes: #2920


abidlabs:
Thanks for the review @aliabid94. You're referring to the captions.json file inside the folder, correct? It contains a dictionary mapping every image path to the caption. What format did you expect?

PR #3005, opened 2023-01-17
Correctly add uploaded files to temp set

abidlabs:
In some cases, files uploaded through the `File` and `UploadButton` components were not being added to the respective temp sets, which broke the `chatbot_multimodal` demo. This is a quick fix

cc @dawoodkhan82


dawoodkhan82 (approved):
Nice! fixes the image issue

PR #3008, opened 2023-01-18
Sharing Infra for custom themes

freddyaboulton:
# Description

A proposal for how users can share custom themes via the hub.

## Developer Experience

1. Create a new theme by modifying either the default or solid built-in themes. If we add a method to go from python theme class to json, then we can let users build off of existing themes too!
2. Call the `upload_to_hub` method on `Theme`
image
3. Then pass the name of the theme to `gr.Blocks()`!

```python
import gradio as gr

# This is the xray demo

with gr.Blocks(theme="freddyaboulton/my-solid-theme") as demo:
    gr.Markdown(
        """
# Detect Disease From Scan
With this model you can lorem ipsum
- ipsum 1
- ipsum 2
"""
...
```
image

## Implementation
* The `upload_to_hub` method will create a dataset repo to store the theme files 
* Currently we only store the css but if we add a `Theme.to_json` method to `Theme` class we can also store the json file. That will enable users to instantiate a theme's python class from the hub and will make it much easier to create new themes by tweaking an existing theme!

# Checklist:

# A note about the CHANGELOG


freddyaboulton:
Thanks for the questions @pngwn !

> Wouldn't this need to a theme.load(user/theme-name)

Yea completely agree. The reason why that's not implemented in this PR is that there is no way to serialize an existing theme to disk in a python-friendly way in #2931 . Once we have `Theme.to_json()` or `Theme.save()`, we can implement `Theme.load`.

> Would it not be simpler to use if the upload_to_hub feature was a CLI command rather than a method

Yes, once we have a way to serialize a Theme on disk we can switch to a cli command. Would be more general and consistent with how we'd (probably) share custom components as well.

> How does versioning work? Would the theme schema contain that? 

Not 100% sure how versioning would work. My initial thought was we'd save a semver tag with the theme file. Your `@` syntax for specifying a theme version looks good to me!

Issue #3009, opened 2023-01-18
`gr.HTML` with `<img` throws error when loading a local file on Spaces


apolinario:
### Describe the bug

I’ve updated to the newest version of Gradio and I’m getting a `ERROR: Exception in ASGI application` if I do a 
`gr.HTML('<img src="file/cat-toy.png">')` on a Space (that evidently contains such file) 

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Here’s a Space where you can see it reproduced: https://huggingface.co/spaces/multimodalart/repro_image_error 

### Logs

```shell
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/home/user/.local/lib/python3.8/site-packages/fastapi/encoders.py", line 152, in jsonable_encoder
    data = dict(obj)
TypeError: 'coroutine' object is not iterable

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/user/.local/lib/python3.8/site-packages/fastapi/encoders.py", line 157, in jsonable_encoder
    data = vars(obj)
TypeError: vars() argument must have __dict__ attribute

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/user/.local/lib/python3.8/site-packages/uvicorn/protocols/http/h11_impl.py", line 407, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
  File "/home/user/.local/lib/python3.8/site-packages/uvicorn/middleware/proxy_headers.py", line 78, in __call__
    return await self.app(scope, receive, send)
  File "/home/user/.local/lib/python3.8/site-packages/fastapi/applications.py", line 270, in __call__
    await super().__call__(scope, receive, send)
  File "/home/user/.local/lib/python3.8/site-packages/starlette/applications.py", line 124, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/home/user/.local/lib/python3.8/site-packages/starlette/middleware/errors.py", line 184, in __call__
    raise exc
  File "/home/user/.local/lib/python3.8/site-packages/starlette/middleware/errors.py", line 162, in __call__
    await self.app(scope, receive, _send)
  File "/home/user/.local/lib/python3.8/site-packages/starlette/middleware/cors.py", line 84, in __call__
    await self.app(scope, receive, send)
  File "/home/user/.local/lib/python3.8/site-packages/starlette/middleware/exceptions.py", line 79, in __call__
    raise exc
[...]

PR #3010, opened 2023-01-18
Fixed old files route

abidlabs:
Fixes the `/file/` route that was broken in a recent change by making it async and adds a test as well

Fixes: #3009


freddyaboulton (approved):

abidlabs:
Thanks for the review @freddyaboulton

Issue #1372, opened 2022-05-24
Prevent big files from being merged to the master with CI/CD


FarukOzderim:
- [X] I have searched to see if a similar issue already exists.

- [ ] Find or write a CI tool to warn or report about introduced size difference by PRs.

FarukOzderim:
Added the CI for this and testing it in #1376, however there are some problems with the sizewatcher, might use smt different if it won't be solved.

Current policy of the check:
10MB+ diff: error
1MB+ diff: warning

FarukOzderim:
I think we should also clear the big files from history while we're at it. Clearing the history is easy with bfg like in #813, however we should back it up with CI or we get big files back.

I say we remove unnecessary files with >1MB from history, and have ban against them. If there is a gif or image, let them get uploaded to cloud in the future, as a policy. Don't want to go overkill about this, let me hear your thoughts as well. @aliabd, @aliabid94 , @abidlabs.
```

➜  gradio git:(main) ✗ git rev-list --objects --all |
  git cat-file --batch-check='%(objecttype) %(objectname) %(objectsize) %(rest)' |
  sed -n 's/^blob //p' |
  sort --numeric-sort --key=2 |
  cut -c 1-12,41- |
  $(command -v gnumfmt || echo numfmt) --field=2 --to=iec-i --suffix=B --padding=7 --round=nearest | tail -n 100
410e6cff9a29  957KiB demo/screenshots/image_mod/1.png
41e4b7abd736  958KiB gradio/frontend/static/bundle.js
6f747d0ec84f  958KiB gradio/frontend/static/bundle.js
ff7abf8a0c05  958KiB gradio/frontend/static/bundle.js
7e45d3e6b424  975KiB gradio/test_data.py
3657cb4601e5  975KiB gradio/test_data.py
f1dc8c5d95e0  975KiB gradio/test_data.py
69dc98958dbd  988KiB gradio/media_data.py
43d3f34dab16  988KiB gradio/media_data.py
1f91c6e92f67  988KiB gradio/media_data.py
97b4b5af5fa9  988KiB gradio/media_data.py
2b38ac091c4a  1.1MiB demo/blocks_flipper/screenshot.gif
c8af75396128  1.1MiB demo/image_mod/screenshot.png
1e6ea1d63285  1.1MiB demo/hello_world_2/screenshot.gif
6df68af06d56  1.1MiB demo/screenshots/image_mod/cheetah2.png
[...]

PR #3013, opened 2023-01-18
Adds a GitHub action to detect if any files > 5 MB are present in the PR

abidlabs:
Moving everything to GitHub actions reminded me of this old issue that we had: https://github.com/gradio-app/gradio/issues/1372

This PR adds the action and causes the check to fail if any file > 5 MB is present. 

You can see it in action here:
* Successful run (no large files): https://github.com/abidlabs/gradio/pull/1
* Unsuccessful run (large file present): https://github.com/abidlabs/gradio/pull/2

Closes: #1372

### DIFF
--- a/.github/workflows/large-files.yml
+++ b/.github/workflows/large-files.yml
@@ -11,14 +11,8 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - name: Checkout code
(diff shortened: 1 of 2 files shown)


freddyaboulton (approved):
Thanks @abidlabs !

PR #3016, opened 2023-01-19
Switch all pages on the website to use latest stable gradio

aliabd:
Switches all the pages on the website except for `/docs/main` to use the latest gradio version instead of building from main. Stops embedding from breaking when there's an unintentional change to main. 

Will fix current embedded interfaces in `index`, `demos`, and `guides`. This will always use latest gradio (version isn't hardcoded) the same way it does for `/docs`. This has the caveat that if we publish a guide on something new we'd need a version bump to show the embedded interfaces in it the correct way, but we can move guides to be from main as well if we choose later.


aliabd:
Working as expected on staging https://www.gradio-staging.com/

PR #3026, opened 2023-01-19
Frontend fixes

aliabid94:
Fixes the following:
- Restores gallery mode for single components in Examples / Dataset
- Fixes loader size on initial load
- Fixes loading animation on pending request
- Shrinks size of shadow on input highlight in dark mode


abidlabs:
Hmm when I click on: https://huggingface.co/spaces/gradio-pr-deploys/pr-3026-all-demos, I get the CSS all broken: 

image

aliabid94:
@abidlabs it works when I load the internal frame directly? Wonder why that might be, looking into it

aliabid94:
ok fixed @abidlabs

abidlabs:
Thanks @aliabid94 for these fixes. However, there are still two differences I noticed between the original styles and the new styles: 

1. with the new styles, there is a shadow that appears when a Gradio app first loads covering the entire screen. This wasn't there in the old style and it looks weird on Spaces where the background of the page is white. It also makes the word "loading" hard to read:

**New style**:
image


**Old style**:

image

2. I _think_ the style of the status tracker is reversed. In the old style, the background of a component would be white and the status tracker would appear as a shadow that progressed from left to right. In the new style, the whole component becomes shadowed and the tracker appears from right to left, which is very unintuitive. The colors of the shadows are slightly different as well

**New style**

image

**Old style**

image

PR deploy for new style: https://huggingface.co/spaces/gradio-pr-deploys/pr-3026-all-demos
PR deploy for old style: https://huggingface.co/spaces/gradio-pr-deploys/pr-3002-all-demos

aliabid94:
All fixed @abidlabs

abidlabs (approved):
Beautiful thanks @aliabid94!

PR #3027, opened 2023-01-19
Adding missing embedded components on docs

aliabd:
This adds missing embedded components to the docs, and fixes some broken ones. 
* Adds embedded component demo for `Dataset`, `Examples`, `Progress`, `UploadButton`,  and `Interpretation`
* Fixes broken image on `Gallery` embedded component
* Removes the footer on `LinePlot`

Live on staging: https://www.gradio-staging.com/docs/main though obviously with the current css issues on main


demo/examples_component/run.py
```diff
@@ -0,0 +1,29 @@
...
+        with gr.Column(elem_id="col_4"):
+            txt_3 = gr.Textbox(value="", label="Output", elem_id="output_text")
+    btn.click(combine, inputs=[txt, txt_2], outputs=[txt_3])
+
+    gr.Examples(
+        [["Hello", "World"], ["Machine", "Learning"], ["Gradio", "Docs"]],
+        [txt, txt_2],
+        txt_3,
+        combine,
+        cache_examples=True,
```
abidlabs, line 26: Why cache examples here?

demo/examples_component/run.py
```diff
@@ -0,0 +1,29 @@
...
+    with gr.Row():
+        with gr.Column(elem_id="col_1"):
+            txt = gr.Textbox(label="Input 1", elem_id="input_1_text")
+        with gr.Column(elem_id="col_2"):
+            txt_2 = gr.Textbox(label="Input 2", elem_id="input_2_text")
+        with gr.Column(elem_id="col_3"):
+            btn = gr.Button(value="Combine Inputs", elem_id="combine_button")
+        with gr.Column(elem_id="col_4"):
+            txt_3 = gr.Textbox(value="", label="Output", elem_id="output_text")
+    btn.click(combine, inputs=[txt, txt_2], outputs=[txt_3])
```
abidlabs, line 19: Do you want to show all of this? I thought the whole point of the components demos was only to show the component itself. 

Also my suggestion would be to include an images or something other than text as one of the inputs since we already showcase text in the Dataset component as well.

abidlabs: Also the "combine inputs" button looks kind of weird by itself in a column. What if you put it under the first two inputs?

demo/gallery_component/run.py
```diff
@@ -9,7 +9,6 @@
         "https://img.etimg.com/thumb/msid-50159822,width-650,imgsize-129520,,resizemode-4,quality-100/.jpg",
         "https://nationalzoo.si.edu/sites/default/files/animals/cheetah-002.jpg",
         "https://images.theconversation.com/files/375893/original/file-20201218-13-a8h8uq.jpg?ixlib=rb-1.1.0&rect=16%2C407%2C5515%2C2924&q=45&auto=format&w=496&fit=clip",
-        "https://encrypted-tbn0.gstatic.com/images?q=tbn:ANd9GcQeSdQE5kHykTdB970YGSW3AsF6MHHZzY4QiQ&usqp=CAU",
         "https://www.lifegate.com/app/uploads/ghepardo-primo-piano.jpg",
         "https://i.natgeofe.com/n/60004bcc-cd85-4401-8bfa-6f96551557db/cheetah-extinction-3_3x4.jpg",
```
abidlabs, line 13: Not sure why but the `natgeofe` image isn't being recognized as a valid URL for me. I get this: 
```
OSError: [WinError 123] The filename, directory name, or volume label syntax is incorrect: 'https:\\i.natgeofe.com\\n\\60004bcc-cd85-4401-8bfa-6f96551557db\\cheetah-extinction-3_3x4.jpg'
```

abidlabs:
Went through the demos locally and they look good @aliabd! Left a few nits above but otherwise LGTM

abidlabs (approved):

demo/examples_component/run.py
aliabd: Was just trying to reduce vertical space and focus the demo more on the examples component, but don't think it matters.

Issue #3024, opened 2023-01-19
Path.resolve break symlink and result in error


KohakuBlueleaf:
### Describe the bug

I use mklink to make symlink to my NAS' folder (connect with smb), and when it return the img with its path, the path become original smb path (\\\\server\\...\\..) and just broken all things

(I'm using A1111's sd-webui but this issue is more like gradio's bug)

For now, I just replace all the Path.resolve to Path.absolute and it works great now.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

use a1111's sd-webui, set output dir to an symlink to SMB folder (on windows)

and will get something like this:
```
ValueError: File cannot be fetched: /192.168.88.239/MainNas/sd-webui/main-outputs/txt2img-images/01335-2023-01-20_d7ff2a1d60_KBlueLeaf_KBlueLeaf_3273134466-512x512.png. 
All files must contained within the Gradio python app working directory, or be a temp file created by the Gradio python app.
```

### Logs

```shell
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "C:\Users\apoll\Desktop\application\stable-diffusion-webui\venv\lib\site-packages\uvicorn\protocols\http\h11_impl.py", line 407, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
  File "C:\Users\apoll\Desktop\application\stable-diffusion-webui\venv\lib\site-packages\uvicorn\middleware\proxy_headers.py", line 78, in __call__
    return await self.app(scope, receive, send)
  File "C:\Users\apoll\Desktop\application\stable-diffusion-webui\venv\lib\site-packages\fastapi\applications.py", line 270, in __call__
    await super().__call__(scope, receive, send)
  File "C:\Users\apoll\Desktop\application\stable-diffusion-webui\venv\lib\site-packages\starlette\applications.py", line 124, in __call__
    await self.middleware_stack(scope, receive, send)
  File "C:\Users\apoll\Desktop\application\stable-diffusion-webui\venv\lib\site-packages\starlette\middleware\errors.py", line 184, in __call__
    raise exc
  File "C:\Users\apoll\Desktop\application\stable-diffusion-webui\venv\lib\site-packages\starlette\middleware\errors.py", line 162, in __call__
    await self.app(scope, receive, _send)
  File "C:\Users\apoll\Desktop\application\stable-diffusion-webui\venv\lib\site-packages\starlette\middleware\gzip.py", line 24, in __call__
[...]

PR #3037, opened 2023-01-20
Added symlink support

abidlabs:
When we switched from `os.path.abspath()` to `pathlib.Path.resolve()`, we ended up resolving symlinks which can be a problem in certain cases (see #3024, as well as the discussion here: https://discuss.python.org/t/pathlib-absolute-vs-resolve/2573/3). I've changed the way that we resolve absolute paths to avoid resolving symlinks. 

It would be great if someone on a Mac could test this (by running `kitchen_sink` as well).

Fixes: #3024


gradio/utils.py
```diff
@@ -850,3 +850,11 @@ def tex2svg(formula, *args):
     svg_code = re.sub(r"<metadata>.*<\/metadata>", "", svg_code, flags=re.DOTALL)
     copy_code = f"<span style='font-size: 0px'>{formula}</span>"
     return f"{copy_code}{svg_code}"
+
+
+def abspath(path: str | Path) -> Path:
+    """Returns absolute path of a str or Path path, but does not resolve symlinks."""
+    if Path(path).is_symlink():
+        return Path.cwd() / path
```
abidlabs, line 858: Inspired by this: https://stackoverflow.com/questions/42513056/how-to-get-absolute-path-of-a-pathlib-path-object/44569249#44569249

aliabid94 (approved):
lgtm, kitchen_sink works on mac

abidlabs:

Issue #2798, opened 2022-12-12
Remove backticks for inline code in `Markdown`


abidlabs:
### Describe the bug

When rendering inline code within `gr.Markdown()`, backticks surround the code, instead of being removed like other Markdown processors. Here's an example:

![image](https://user-images.githubusercontent.com/1778297/207112943-57ea32f2-7f9e-4e71-a81e-c1062ecf880a.png)

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

https://huggingface.co/spaces/society-ethics/disaggregators

### Logs

```shell
N/A
```

### System Info

```shell
gradio==3.12
```

### Severity

annoying

dawoodkhan82:
@abidlabs  Seems like this is only on spaces. Tested locally, and the backticks are removed.

abidlabs:
Oh really, very interesting... so some css on Spaces affecting the markdown code blocks?

dawoodkhan82:
That's seems likely. Although if you can double check by running a markdown demo, that'd be great.

abidlabs:
Works beautifully for me locally:

<img width="579" alt="image" src="https://user-images.githubusercontent.com/1778297/214435705-ab179f7a-0468-4f8a-9762-b8f5922a909a.png">

<img width="660" alt="image" src="https://user-images.githubusercontent.com/1778297/214435729-08d7c66d-4c78-4b83-be1c-e0e92f2bb761.png">

Testing on Spaces now

abidlabs:
Actually it's working nicely for me on Spaces too now: https://huggingface.co/spaces/gradio-pr-deploys/pr-3051-all-demos! 

<img width="1005" alt="image" src="https://user-images.githubusercontent.com/1778297/214437031-d92883a5-2ead-415f-925a-e75911f0550b.png">

Maybe @pngwn revamped code styling as part of the theming PR? In any case it looks like we can close this @dawoodkhan82

pngwn:
Yeah sorry forgot to mention, this was addressed when I did the new typography as part of the theming PR.

PR #3051, opened 2023-01-24
Updated series demo

abidlabs:
Tiny update to the `interface_series_load` demo as part of testing #2798 on Spaces


freddyaboulton (approved):
Thanks @abidlabs !

abidlabs:

Issue #3057, opened 2023-01-25
the 'package' argument is required to perform a relative import


xianbaoqian:
### Describe the bug

Failed to run `gradio app.py` due the following error.

Looks that there are some weirdness with uvicorn. Could I be installed the wrong version?

Thx

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

- python -m venv env
- export PATH=env/bin:$PATH
- pip install gradio
- gradio app.py

(env) wangtz@Tiezhens-MBP hf-puller % gradio app.py

Launching in *reload mode* on: http://127.0.0.1:7860 (Press CTRL+C to quit)

Watching: '/Users/wangtz/projects/tools/hf-puller/env/lib/python3.9/site-packages/gradio', '.'

Process SpawnProcess-1:
Traceback (most recent call last):
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/Users/wangtz/projects/tools/hf-puller/env/lib/python3.9/site-packages/uvicorn/_subprocess.py", line 76, in subprocess_started
    target(sockets=sockets)
  File "/Users/wangtz/projects/tools/hf-puller/env/lib/python3.9/site-packages/uvicorn/server.py", line 60, in run
    return asyncio.run(self.serve(sockets=sockets))
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/asyncio/base_events.py", line 642, in run_until_complete
    return future.result()
  File "/Users/wangtz/projects/tools/hf-puller/env/lib/python3.9/site-packages/uvicorn/server.py", line 67, in serve
    config.load()
  File "/Users/wangtz/projects/tools/hf-puller/env/lib/python3.9/site-packages/uvicorn/config.py", line 477, in load
    self.loaded_app = import_from_string(self.app)
[...]

PR #3059, opened 2023-01-25
fix css scoping

### LINKED ISSUE #2763: Inconsistent behavior with Javascript when running app from an iframe
### Describe the bug

If someone wants to select things on a Gradio app using Javascript, the behavior changes if they are in an iframe or outside of an iframe. Outside of an iframe, a `shadowRoot` is required to select the element. But when inside an iframe, having a `shadowRoot` doesn't select the element and breaks the app.

### Reproduction

[An example](https://huggingface.co/spaces/multimodalart/shadow-root/blob/main/app.py#L19) using `shadowRoot`: 
- Works when it is direct access/localhost, like here: https://multimodalart-shadow-root.hf.space/
- Breaks when it is an iframe/Space, like here:
https://huggingface.co/spaces/multimodalart/shadow-root

[An example](https://huggingface.co/spaces/multimodalart/shadow-root-reverse/blob/main/app.py#L19) *not* using `shadowRoot`
- Works when it is an iframe/Space, like here: https://huggingface.co/spaces/multimodalart/shadow-root-reverse
- Breaks when it is direct access/localhost, like here: https://multimodalart-shadow-root-reverse.hf.space/

This makes the AUTOMATIC1111 WebUI incompatible with running on Hugging Face Spaces (or any other iframe) for e.g., as they manipulate the app with Javascript with a shadowRoot: https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/master/script.js#L2 

### System Info

```shell
Gradio == 3.12.0
```

### Severity

Except if there's a universal way to select elements we are missing, it is blocking all usage of gradio inside a Space if Javascript is manipulated

### LINKED ISSUE #3036: Option to disable shadow dom
- [ ✔] I have searched to see if a similar issue already exists.

Feels like many developers are just struggling to add custom features to gradio components because the shadow dom prevent javascript from reaching the components easily. I've searched a lot in the issues but none of them allows accessing shadow dom. It took me a whole day trying to figure out how to add a click event to the gallery with external javascript (added to the closing of body) but ended up deciding to switch to another framework...

Just simply add an option to disable the shadow dom and make all html elements visible to external javascript will be much helpful. 

**Additional context**  
If it's not possible to disable shadow dom, I would also appreciate that if we can have an interface to load js into the whole app (instead of for individual components) or allow directly editing the html of the coponents.

pngwn:
# Description

Ignore the branch name. This PR fixes the weird unstyled Gradio we have been seeing on the Docs.

The actual reason for this is kinda weird but essentially it only happened when there was more than one instance of gradio on the page, embedded via the web component.

What this PR does is removes the use of the shadow DOM and mounts the CSS + JS into the head of the parent document, sharing the styles across all instances of `gradio-app` rather than each `gradio-app` needing their own set of styles. We can do this now because svelte is handling most of the scoping for us. We _do_ still need to scope css resets and the few global styles we have. I have done this by prefixing all non hashed (svelte) css selectors with `gradio-container-<VERSION>` (except`.` has been converted to `-`), so something like `gradio-container-3-16-1 h1 { ... }`. This means you could potentially have multiple different versions of gradio on the same page and they wouldn't conflict.

This also simplifies the code somewhat, we needed multiple build time modifications in order to make things work with custom elements and the shadow DOM. Since they only worked during build and not dev we needed to mount the app differently in different modes. These distinctions are now gone.

tldr: we're scoping differently.

closes #3036. probably closes #2763.

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

Closes: # (issue)

# Checklist:

# A note about the CHANGELOG


aliabd (approved):
Tested on [staging](https://www.gradio-staging.com/docs/main) and everything works as expected! 

Small note on min-height css. You'll see the embedded component demos have css that overrides min-height and and removes the footers so they don't take unnecessary space. Here's an [example](https://huggingface.co/spaces/gradio/textbox_component/blob/main/run.py#L3). Should I change that selector to `gradio-container-3-16-2`? That's what i'm seeing now. 

Also all that type of css will need to be much more specific now without the shadow DOM. For example all footers on every demo are gone. The docs page footer is gone too! Not a big deal though, I can clean this up after the PR is merged.

abidlabs:
Does this mean that if a user updates their version of gradio, they would have to change any custom `css` that they have passed into the Blocks/Interface? That would be a pretty big inconvenience imo

pngwn:
Good point. I'll change this so that there is both the version specific class name (for our own scoping) in addition to a normal `gradio-container` (which is easier for users).

pngwn:
This PR would mean that there is an increased chance of collisions, it is technically possible to keep the shadow DOM and fix this issue but quite difficult due to shortcomings in the web component API. The API that facilitates this is actually supported in chrome but not safari. 

In order to keep things fully scoped (with the shadow DOM) but fix the issues we are facing it would take a bit of time because some of our tooling would need to be rewritten. I can take that approach if we are uncomfortable with the risk this change would introduce.

abidlabs:
I actually think removing the shadow DOM is a good idea for our users as well (it's caused lots of confusion / questions in the past). My main concern was users having to change the CSS between versions -- if that's not going to the case, this sounds great!

pngwn:
@aliabd Can you test this again now. I've added back the `gradio-container` class so that should styling a bit easier, although as the shadow DOM is gone anyone embedding will need to be careful not to  write styles that impact the embedded gradio app. Gradio itself will never affect the containing page though.

aliabd:
Just tested on staging @pngwn everything looks good: https://www.gradio-staging.com/docs/main

abidlabs:
Nice fix @pngwn! One nit: can we update the documentation on embedding with web components to include this caveat when using your own `css`. 

Relevant file: https://github.com/gradio-app/gradio/blob/main/guides/01_getting-started/03_sharing-your-app.md#embedding-with-web-components

pngwn:
@abidlabs I have added a note to that guide, explaining the caveat and which kinds of selectors are more likely to cause issues. If there is a better way to phrase or a better place to put it then let me know!

abidlabs:
LGTM @pngwn thanks for the awesome PR!

abidlabs (approved):

SoulflareRC:
Is this fix added in 3.17? I'm very excited with the removal of shadow DOM! With 3.16.2 I'm still having trouble running the reload mode as #3057 describes. I can only develop with 3.15...

abidlabs:
Hi @SoulflareRC yes it's added in 3.17 which was just released. Let us know if you are still having any issues.

SoulflareRC:
Is it possible to add custome js and css after this update? I was trying to do so by modifying the template response but neither jquery or js seem to be able to access any gradio component.

pngwn:
Nothing should changed regarding custom CSS or JS. Could you open a new issue or discussion or post on discord with some code?

PR #3075, opened 2023-01-28
add argument to allow serving files from additional directories

maxaudron:
# Description

Add the `file_directories` argument to allow serving files from additional directories.
I was working on separating user data from source code for AUTOMATIC1111/stable-diffusion-webui to a directory external to the source and ran into trouble with various files the webui tries to serve from these locations.

# Checklist:

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -513,6 +513,8 @@ class Blocks(BlockContext):
         self.api_mode = None
         self.progress_tracking = None
 
+        self.file_directories = []
+
         if self.analytics_enabled:
             data = {
                 "mode": self.mode,
@@ -1259,6 +1261,7 @@ class Blocks(BlockContext):
         ssl_keyfile_password: str | None = None,
         quiet: bool = False,
         show_api: bool = True,
+        file_directories: List[str] | None = None,
         _frontend: bool = True,
     ) -> Tuple[FastAPI, str, str]:
         """
@@ -1288,6 +1291,7 @@ class Blocks(BlockContext):
             ssl_keyfile_password: If a password is provided, will use this with the ssl certificate for https.
             quiet: If True, suppresses most print statements.
             show_api: If True, shows the api docs in the footer of the app. Default True. If the queue is enabled, then api_open parameter of .queue() will determine if the api docs are shown, independent of the value of show_api.
+            file_directories: List of directories that gradio is allowed to serve files from (in addition to the directory containing the gradio script). Must be absolute paths. Warning: these files are exposed to all users of your app.
         Returns:
             app: FastAPI app object that is running the demo
             local_url: Locally accessible link to the demo
@@ -1330,6 +1334,8 @@ class Blocks(BlockContext):
         if self.enable_queue and not hasattr(self, "_queue"):
             self.queue()
         self.show_api = self.api_open if self.enable_queue else show_api
+
+        self.file_directories = file_directories if file_directories is not None else []
 
         if not self.enable_queue and self.progress_tracking:
             raise ValueError("Progress tracking requires queuing to be enabled.")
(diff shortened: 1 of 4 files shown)


abidlabs:
Hi @maxaudron can you elaborate on the issue this PR is fixing a little more?

maxaudron:
@abidlabs I'm using nix to install stable-diffusion-webui as a proper python package, this is stored on a read-only filesystem, but really the same applies for any installed python packages as they should not be directly written to.

stable-diffusion-webui can for example use extensions that provide additional css or javascript files to serve. These are now stored outside the python app source, again, because read-only filesystem of the python app source, and just general good manners.

The source directory is also not necessarily the working directory, causing another potential detachment, I had worked around that by setting the working directory manually to the python app source.

It is a good security consideration to not allow serving from arbitrary directories, but serving files from outside the application's source has plenty of potential uses, and having an escape hatch to allow that is useful.

This way the application can provide a number of directories that are valid to serve files from, like it's own root based on the current file location, or any number of directories provided by configuration.

abidlabs:
Got it, thanks for the clarification. In the future, a better workflow would be to open an issue with the bug / feature request, so that the community and maintaining team can discuss the issue. But as a one off, I'll review this PR.

gradio/blocks.py
```diff
@@ -1258,6 +1260,7 @@ def launch(
         ssl_keyfile_password: str | None = None,
         quiet: bool = False,
         show_api: bool = True,
+        file_directories: List[str] = [],
```
abidlabs, line 1263: Better to [avoid using a mutable as a default](https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/). Can you set this to be `None` and assign to it a list inside of the function?

gradio/blocks.py
```diff
@@ -1287,6 +1290,7 @@ def launch(
             ssl_keyfile_password: If a password is provided, will use this with the ssl certificate for https.
             quiet: If True, suppresses most print statements.
             show_api: If True, shows the api docs in the footer of the app. Default True. If the queue is enabled, then api_open parameter of .queue() will determine if the api docs are shown, independent of the value of show_api.
+            file_directories: List of directories that gradio is allowed to serve files from. Must be absolute paths. Default []
```
abidlabs, line 1293:
```suggestion
            file_directories: List of directories that gradio is allowed to serve files from in addition to the directory containing the gradio script. Must be absolute paths.
```

gradio/routes.py
```diff
@@ -265,7 +265,12 @@ async def file(path: str, request: fastapi.Request):
             created_by_app = str(utils.abspath(path)) in set().union(
                 *blocks.temp_file_sets
             )
-            if in_app_dir or created_by_app:
+            in_file_dir = any(
+                dir in (str(p) for p in utils.abspath(path).parents)
+                for dir in blocks.file_directories
+            )
```
abidlabs, line 271: Can we do it this way to take advantage of pathlib's built-in `in` functionality
```suggestion
            in_file_dir = any((utils.abspath(dir) in utils.abspath(path).parents for dir in blocks.file_directories))
```

abidlabs:
Hi @maxaudron thanks again for the PR. I made some suggested changes to the PR. In addition, would you please be able to add a unit test that tests this functionality? You can see here for some similar tests: https://github.com/gradio-app/gradio/blob/4aa7d8d74b2d98f7c314e43a902983861d3d4dc9/test/test_routes.py#L183

abidlabs (changes requested):
See above

abidlabs:
@maxaudron is this ready to be re-reviewed?

maxaudron:
@abidlabs yes, I've applied your suggestions and added a test

abidlabs (approved):

abidlabs:
Made a tweak to the test code, will merge in after it passes, thanks for the nice PR @maxaudron!

PR #3079, opened 2023-01-29
Chatbot loader

aliabid94:
Changed the statustracker animation to a chatbot specific animation, see below. Also fixed bug where chat would not stick to the bottom when images were involved.

Recording 2023-01-28 at 20 30 55


pngwn:
Shouldn't the loading animation be on the response that is being generated rather than the message written by the user?

pngwn (approved):
lgtm!

freddyaboulton (approved):
Nice change @aliabid94 ! 

I wonder if we can add some artificial latency to the "bot" response as well to give the impression that the response for the bot is still being calculated. This looks great to me as is though!

abidlabs:
To follow up with @pngwn's point, the animation currently shows up on the user message, right? Which makes sense since we send the user response to the backend to get it preprocessed. Just wanted to make sure I'm not missing anything between @pngwn's comment and his approval of the PR

image

pngwn:
Approved because i think it is already better, implementation looks good, and i don't want to block.

I _do_ think the loader should be on the bot message tho, since that is what is actually pending. 

It should actually be like this:

- user types message into input box + submits
- users message show up immediately (different to current behaviour)
- pending dots show for bot message
- when bot response comes back from the server it replaces the dots.

This type of loader is basically the 'currently typing' status used by every chat app out there, they all work as i described above rather than how they work in this PR.

I _think_ it should be possible to immediately show the users message (which is necessary for the bot loader to make any sense) but I haven't actually checked.

aliabid94:
Completely agree with points above. I will merge this in for now and open issue for other points, since they are more involved than it may appear.

usuyama:
Thank you for this improvement!

@aliabid94 do you already have a new issue opened for this?

> Completely agree with points above. I will merge this in for now and open issue for other points, since they are more involved than it may appear.

PR #3089, opened 2023-01-30
Update docs to explain how reload mode can be used with queue/auth

### LINKED ISSUE #1883: Warn enable_queue in reload mode
Warn the user when enable_queue is true in the reload mode(gradio app.py)

**Additional context**  
Since we launch the app with uvicorn in the reload mode, only fastapi app is run. We could make workarounds like triggering an endpoint to the start queueing, however I don't think it is necessary.

### LINKED ISSUE #2158: Auth not set when using reload mode
### Describe the bug

When using reload mode, any of the features that you configure in the `launch` method will not be respected, including auth, queue, show_error.

That's expected because reload mode doesn't call the launch method but this may be confusing to users and maybe calls for a redesign of how reload mode works. 

I don't think this is a high priority issue though.

### Reproduction

```python
import gradio as gr
import os


def identity(x):
    return x[::-1]


with gr.Blocks() as demo:
    output = gr.Textbox(label="Output", lines=2)
    txt.change(identity, txt, output)

if __name__ == "__main__":
    demo.launch(auth=[("admin", "admin")])
```

### Screenshot

reload_mode_auth

### Logs

```shell
-
```

### System Info

```shell
3.2
```

### Severity

annoying

freddyaboulton:
# Description

Fixes #2158
Fixes #1883

I noticed that if you call  `launch(....)` outside of the `if __name__ == "__main__"` block, then the `gradio` reload command will properly pick up whether or not the queue should be enabled as well as `auth` and the other `launch`-related parameters.

The downside is that the page will not reload automatically, but if you reload it yourself, it will reflect the latest state of `app.py`.

I don't think this is too annoying and it beats having to refactor the `launch` method to properly work with reload mode so I'm proposing we just update the docs to explain this behavior and call it a day.

Let me know if you agree!

The script I used to test

```python
import gradio as gr


    if operation == "add":
        return num1 + num2
    elif operation == "subtract":
        return num1 - num2
    elif operation == "multiply":
        return num1 * num2
    elif operation == "divide":
        return num1 / num2


with gr.Blocks() as demo:
    with gr.Row():
        with gr.Column():
            num_1 = gr.Number(value=6)
            operation = gr.Radio(["add", "subtract", "multiply", "divide"])
            num_2 = gr.Number(value=0)
            submit_btn = gr.Button(value="Calculate")
        with gr.Column():
            result = gr.Number()

    submit_btn.click(calculator, inputs=[num_1, operation, num_2], outputs=[result])
    examples = gr.Examples(examples=[[5, "add", 3],
                                     [4, "divide", 2],
                                     [-4, "multiply", 2.5],
                                     [0, "subtract", 1.2]],
                           inputs=[num_1, operation, num_2])

demo.queue().launch(auth=("foo", "foo"), auth_message="Please log in.")
```

### Working with auth
reload_with_auth_and_queue

### Reflect latest app.py after manual refresh

reload_working_queue

### Docs changes
image

# Checklist:

# A note about the CHANGELOG

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -1232,6 +1232,7 @@ class Blocks(BlockContext):
             blocks_dependencies=self.dependencies,
         )
         self.config = self.get_config_file()
+        self.app = routes.App.create_app(self)
         return self
 
     def launch(
--- a/gradio/routes.py
+++ b/gradio/routes.py
@@ -479,6 +479,7 @@ class App(FastAPI):
         async def get_queue_status():
             return app.get_blocks()._queue.get_estimation()
 
+        @app.on_event("startup")
         @app.get("/startup-events")
         async def startup_events():
             if not app.startup_events_triggered:
--- a/guides/06_other-tutorials/developing-faster-with-reload-mode.md
+++ b/guides/06_other-tutorials/developing-faster-with-reload-mode.md
@@ -54,10 +54,7 @@ The important part here is the line that says `Watching...` What's happening her
 
 As a small aside, this auto-reloading happens if you change your `app.py` source code or the Gradio source code. Meaning that this can be useful if you decide to [contribute to Gradio itself](https://github.com/gradio-app/gradio/blob/main/CONTRIBUTING.md) ✅
 
-⚠️ It is customary to configure the behavior of your gradio app via the `launch()` or `queue()` methods. For example, setting `auth`, or `show_error` in `launch()`. If this is the case, make sure that these methods are
-called **outside** of the `if __name__ == "__main__":` block, otherwise the `gradio` command will not know the 
-value of these parameters. You may have to manually reload the web browser page yourself, but it will reflect
-the latest changes of your `app.py` file!
+⚠️ The `gradio` command will not detect the parameters passed to the `launch()` methods. For example, setting `auth`, or `show_error` in `launch()` will not be reflected in the app.
 
 ## Jupyter Notebook Magic 🔮
 
--- a/guides/06_other-tutorials/developing-faster-with-reload-mode.md
+++ b/guides/06_other-tutorials/developing-faster-with-reload-mode.md
@@ -54,6 +54,10 @@ The important part here is the line that says `Watching...` What's happening her
 
 As a small aside, this auto-reloading happens if you change your `app.py` source code or the Gradio source code. Meaning that this can be useful if you decide to [contribute to Gradio itself](https://github.com/gradio-app/gradio/blob/main/CONTRIBUTING.md) ✅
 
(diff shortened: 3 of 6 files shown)


abidlabs:
Hmm running the script you provided (or any script in which `demo.launch()` is not enclosed in the  `if __name__ == "__main__"` block) produces the following error message:

image

And no webpage appears:

image

Looking more deeply to understand why

guides/06_other-tutorials/developing-faster-with-reload-mode.md
```diff
@@ -54,6 +54,11 @@ The important part here is the line that says `Watching...` What's happening her
 
 As a small aside, this auto-reloading happens if you change your `app.py` source code or the Gradio source code. Meaning that this can be useful if you decide to [contribute to Gradio itself](https://github.com/gradio-app/gradio/blob/main/CONTRIBUTING.md) ✅
 
+⚠️ It is customary to configure the behavior of your gradio app via the `launch()` or `queue()` methods. For example, setting `auth`, or `show_error` in `launch()`. If this is the case, make sure that these methods are
```
abidlabs, line 57: This only applies to `launch()`, right? You could have `queue()` outside of the block and everything should work fine I believe

freddyaboulton:
Going to think about this some more since it doesn't work for windows ⚰️

guides/06_other-tutorials/developing-faster-with-reload-mode.md
abidlabs: Actually no because the queue doesn't get configured

freddyaboulton:
@abidlabs Thanks for bouncing some ideas off with me yesterday!

Here's what I did:
1. reworked `Blocks` a tiny amount so that `queue` works with reload mode
2. reworded the docs so that they mention that reload mode does NOT work with parameters set in `launch`

Here's a gif of reload mode + queue:

queue_reload

Should be good for another look!

gradio/routes.py
```diff
@@ -479,6 +479,7 @@ async def join_queue(
         async def get_queue_status():
             return app.get_blocks()._queue.get_estimation()
 
+        @app.on_event("startup")
```
freddyaboulton, line 482: So that the queue can start with calling launch. I checked and there's no problem in calling this route twice on `startup` and during `launch`

gradio/blocks.py
```diff
@@ -1232,6 +1232,7 @@ def queue(
             blocks_dependencies=self.dependencies,
         )
         self.config = self.get_config_file()
+        self.app = routes.App.create_app(self)
```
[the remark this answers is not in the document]
abidlabs: Nice

abidlabs (approved):
Beautiful! Tested and works great. May want to add a unit test, but will leave that up to you. LGTM!

freddyaboulton:

Just added a unit test.

Issue #3056, opened 2023-01-25
Infinite loop when on cascade update if JSON type component in cascade


Norbiox:
### Describe the bug

Hi everyone,
I've tried to achieve cascade of updating component  using `change` event. However if one of components in cascade is of type JSON then this field is constantly updated with no sign of ever stopping. I think it's a bug because if I change this field to Text type everything works as expected. 

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```python
#!/usr/bin/env python3
import gradio as gr

def returns_dict(value):
    return {"key": value}

with gr.Blocks() as demo:
    with gr.Row():
        field1 = gr.Text(interactive=True)
        field2 = gr.JSON()
        field3 = gr.Text()

    field1.change(fn=returns_dict, inputs=field1, outputs=field2)
    field2.change(fn=returns_dict, inputs=field2, outputs=field3)

if __name__ == "__main__":
    demo.launch()
```

### Logs

```shell
Running on local URL:  http://127.0.0.1:7860

To create a public link, set `share=True` in `launch()`.

(I set `debug=True` in `launch` method but that's all I got)
```

### System Info

```shell
Gradio: 3.16.2
Python: 3.10.9
System: ArcoLinux with kernel 6.1.5-arch2-1
```

### Severity

serious, but I can work around it

aliabid94:
Looking into this, thanks for reporting!

PR #3095, opened 2023-01-31
Fix change event listener

aliabid94:
Change event listener was broken on a couple components that simply check for a change in value if the value type is an object. Basically this line:
```javascript
$: value: dispatch("change")
```
will lead to an infinite loop if the value type is an object. See #3056 for example code that causes this bug. 

The reason for this infinite loop is because in a change event listener loop:
1. the value changes, which 
2. fires the event listener, which 
3. triggers a re-render of the whole tree, which 
4. will create a new value object for all components. Values for primitives are still === to their previous values, but objects will not be considered equivalent to their previous values (they have the same internal values to their previous values, but not are not ===). 
5. This causes another dispatch("change"), which repeats the cycle.

Fixes: #3056

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 No changes to highlight.
 
 ## Bug Fixes:
-No changes to highlight.
+- Fix change event listener for JSON, HighlightedText, Chatbot by [@aliabid94](https://github.com/aliabid94) in [PR 3095](https://github.com/gradio-app/gradio/pull/3095)
 
 ## Documentation Changes:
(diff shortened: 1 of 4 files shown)


freddyaboulton (approved):
Tested with all three components changed here and can confirm the fix! Thanks @aliabid94 !

pngwn (approved):
I'm hopeful this can be fixed more elegantly when I refactor that state management, but this looks good to me. tested everything listed and it works as expected vs main.

pngwn:
Need to work off this for another PR, so merging.

Issue #3105, opened 2023-02-01
In v3.17.0, Gallery can't show a grid of images!


myhz0606:
### Describe the bug

Gallery can not show a grid of Images, though set ` gr.Gallery(label="Generated images", show_label=False, elem_id="my_gallery").style(grid=6)` .
Version of 3.16.0 didn't have this problem.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

import os
import random
import requests
from PIL import Image 
 
import gradio as gr
 
 
def fake_gan(*args):
    images = [
        (random.choice(
            [
                "https://images.unsplash.com/photo-1507003211169-0a1dd7228f2d?ixlib=rb-1.2.1&ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&auto=format&fit=crop&w=387&q=80",
                "https://images.unsplash.com/photo-1554151228-14d9def656e4?ixlib=rb-1.2.1&ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&auto=format&fit=crop&w=386&q=80",
                "https://images.unsplash.com/photo-1542909168-82c3e7fdca5c?ixlib=rb-1.2.1&ixid=MnwxMjA3fDB8MHxzZWFyY2h8MXx8aHVtYW4lMjBmYWNlfGVufDB8fDB8fA%3D%3D&w=1000&q=80",
                "https://images.unsplash.com/photo-1546456073-92b9f0a8d413?ixlib=rb-1.2.1&ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&auto=format&fit=crop&w=387&q=80",
                "https://images.unsplash.com/photo-1601412436009-d964bd02edbc?ixlib=rb-1.2.1&ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&auto=format&fit=crop&w=464&q=80",
            ]
        ), f"label {i}")
        for i in range(10)
    ]
    return images
 
 
with gr.Blocks() as demo:
    with gr.Column():
        with gr.Row():
            text = gr.Textbox(
                label="Enter your prompt",
                show_label=False,
                max_lines=1,
                placeholder="Enter your prompt",
            ).style(
                container=False,
            )
            btn = gr.Button("Generate image").style()
[...]

PR #3107, opened 2023-02-01
Fix gallery grid

aliabid94:
Style for gallery grid broke in tailwind -> pollen migration, fixed: #3105


abidlabs:
Updated the notebook, and tested. LGTM!

abidlabs (approved):

ui/packages/gallery/src/Gallery.svelte
```diff
@@ -293,8 +293,34 @@
...
+	@media (min-width: 1280px) {
+		.grid-container {
+			grid-template-columns: var(--xl-grid-cols);
+		}
+	}
+	@media (min-width: 1536px) {
+		.grid-container {
+			grid-template-columns: var(--2xl-grid-cols);
+		}
+	}
```
pngwn, line 323: Use the media query variables here instead of hardcoded values:

```
--screen-sm: (min-width: 640px)
--screen-md: (min-width: 768px)
--screen-lg: (min-width: 1024px)
--screen-xl: (min-width: 1280px)
--screen-xxl: (min-width: 1536px)
```

You can just do this:

```css
@media (--screen-md) {
 ...
}
```

etc.

pngwn (changes requested):
Just one small thing to keep things consistent. Looks good otherwise.

ui/packages/gallery/src/Gallery.svelte
aliabid94: Oh neat, done.

pngwn (approved):
Awesome. Looks good!

abidlabs:
Will go ahead and merge this in so that we can get a release out with these changes

PR #3108, opened 2023-02-01
Fixes URL resolution on Windows

abidlabs:
Previously, we would run `utils.abspath()` on a path before we knew whether it was a URL or a string path. On Windows, this would lead to an error if the path provided was a URL. This fixes that and makes the code a little clearer to read. 

Test code (run this on a Windows machine):

```py
import gradio as gr
 
with gr.Blocks() as demo:
    gr.Gallery(["https://images.unsplash.com/photo-1507003211169-0a1dd7228f2d?ixlib=rb-1.2.1&ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&auto=format&fit=crop&w=387&q=80"])

demo.launch()
```


abidlabs:
Very strange, it isn't working for me on Windows @freddyaboulton.  Can you try this simpler repro: 

```py
from pathlib import Path

url = "https://images.unsplash.com/photo-1507003211169-0a1dd7228f2d?ixlib=rb-1.2.1&ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&auto=format&fit=crop&w=387&q=80"

Path(url).resolve()
```

freddyaboulton:
Yea calling `.resolve()` on a url returns a faulty path on windows but also ubuntu!

### Windows
image


### Ubuntu
image

It's strange it's broken for you since I just looked at the `file` route code again and we call `validate_url` on `path` not `abspath` so it should correctly detect it's a url no? 

```python
        @app.head("/file={path:path}", dependencies=[Depends(login_check)])
        @app.get("/file={path:path}", dependencies=[Depends(login_check)])
        async def file(path: str, request: fastapi.Request):
            abs_path = str(utils.abspath(path))
            blocks = app.get_blocks()
            if utils.validate_url(path):
                return RedirectResponse(url=path, status_code=status.HTTP_302_FOUND)
```

I wonder if something else is at play 🤔

abidlabs:
Ahh it looks like it's a Python 3.9 vs. 3.10 thing, specifically on Windows 🤯 

When I run the same code I asked you to run on Python 3.9, I get this error raised:

image

Apparently, pathlib fixed it at some point in Python 3.10: https://github.com/python/cpython/commit/4696f1285d83dd7b69f459c63e14080b1f87f797

abidlabs:
You're right, we do call `validate_url`, but the problem is that we call it *after* we call `utils.abspath(path)`. 

This PR changes the order so that we only call `abspath` after we've confirmed that it is not a URL. Let me know if that makes sense

freddyaboulton (approved):
Thanks for the fix + debugging with me @abidlabs !

abidlabs:
Thank you @freddyaboulton!

Issue #3122, opened 2023-02-03
Cache examples don't work with components w/o label attr


radames:
### Describe the bug

In some instances, a user may want to make whole blocks visible or invisible by using Row, Column, Box or Group. Currently, this works correctly with `cache_examples=False`, but this impedes the caching feature.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Please uncomment different to test with different components. 

```python
import gradio as gr

blocks = gr.Blocks()


def predict(text):
  return text, gr.update(visible=True)

with blocks:
  with gr.Row():
    in_text = gr.Textbox(label="Input")
    out_text = gr.Textbox(label="Output")
  # with gr.Accordion(label="Test", visible=False) as box_el:
  # with gr.Group(visible=False) as box_el:
  # with gr.Box(visible=False) as box_el:
  # with gr.Row(visible=False) as box_el:
  with gr.Column(visible=False) as box_el:
    gr.Textbox(value="HELLO")
  btn = gr.Button("Run")
  btn.click(fn=predict, inputs=[in_text], outputs=[out_text, box_el])
  gr.Examples([
      ["example A"],
      ["example B"]
  ],
  fn=predict,
  inputs=[in_text],
  outputs=[out_text, box_el],
  cache_examples=True)

blocks.launch(debug=True, inline=True)

```

### Logs

```shell
AttributeError: 'Group' object has no attribute 'label'
```
```
AttributeError: 'Column' object has no attribute 'label'
```
```
AttributeError: 'Box' object has no attribute 'label'
```
```
AttributeError: 'Row' object has no attribute 'label'
```

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-7-e30545eddc2f> in <module>
     18   btn = gr.Button("Run")
     19   btn.click(fn=predict, inputs=[in_text], outputs=[out_text, box_el])
---> 20   gr.Examples([
     21       ["example A"],
     22       ["example B"]

7 frames
/usr/local/lib/python3.8/dist-packages/gradio/flagging.py in <listcomp>(.0)
    202         is_new = not Path(log_filepath).exists()
    203         headers = [
--> 204             component.label or f"component {idx}"
    205             for idx, component in enumerate(self.components)
    206         ] + [

AttributeError: 'Column' object has no attribute 'label'
```
```


### System Info

```shell
Gradio 3.17.1
```

### Severity

annoying

abidlabs:
Working on it!

PR #3123, opened 2023-02-03
Example caching now works with components without a label attribute (e.g. Column)

abidlabs:
Fixes: #3122 and adds a unit test


freddyaboulton (approved):

Looks good to me. Thanks for adding the unit test.

radames:
very cool! I've tested on Colab using the CI built. Thanks for the quick fix!

radames (approved):

abidlabs:
Thanks for the reviews folks!

Issue #3104, opened 2023-02-01
Add documentation on how to create a variable number of outputs


abidlabs:
We frequently get the question about how to create a variable number of outputs in Gradio. Given that it's a little complex, I think it's worth adding some explicit documentation about it. 

Spurred by this discussion: https://github.com/gradio-app/gradio/discussions/3076

PR #3127, opened 2023-02-04
Adds a demo / information about how to create demos with variable #s of outputs

abidlabs:
Closes: #3104


demo/variable_outputs/run.py
```diff
@@ -0,0 +1,19 @@
...
+
+def variable_outputs(k):
+    k = int(k)
+    return [gr.Textbox.update(visible=True)]*k + [gr.Textbox.update(visible=False)]*(max_textboxes-k)
+
+with gr.Blocks() as demo:
+    s = gr.Slider(1, max_textboxes, step=1, label="How many textboxes to show:")
+    textboxes = []
+    for i in range(max_textboxes):
+        t = gr.Textbox(f"Textbox {i}")
```
freddyaboulton, line 13: I think it would be better if only one textbox is showing on app load since that's the default value of the slider

freddyaboulton (approved):
Thanks @abidlabs ! Looks good to me. I have a minor comment on the actual demo code though.

demo/variable_outputs/run.py
abidlabs: Sounds good, for simplicity of the demo, I'll just set the default value of the slider to be 10

abidlabs:
Thanks for the review @freddyaboulton, will merge after the tests pass

Issue #399, opened 2021-12-10
Gradio doesn't automatically create a shareable link in sagemaker


abidlabs:
The `url_ok()` method needs some work since there are a bunch of open issues about it and it doesn't create a shareable link in Sagemaker notebooks:

![image](https://user-images.githubusercontent.com/1778297/145524558-ea3d4389-9ec5-4772-bec2-833ed01377df.png)

cri5Castro:
Hello I replicated this error today in gradio version 3.1.2
inline doesn't work and also the shareable link 
<img width="983" alt="Captura de Pantalla 2022-08-01 a la(s) 22 16 47" src="https://user-images.githubusercontent.com/11454650/182284468-59f8f80d-32b1-439d-971d-12c83311cfe6.png">
<img width="1007" alt="Captura de Pantalla 2022-08-01 a la(s) 22 17 19" src="https://user-images.githubusercontent.com/11454650/182284473-0ae5d9bc-11df-4387-b483-beebd3bca604.png">

hamkasatria:
hello. I get same error today. gradio cant share links in gradio version 3.15.0

![image](https://user-images.githubusercontent.com/48576063/210221393-65dab1ff-158d-42dc-8c36-6499df1f7194.png)

PR #3137, opened 2023-02-07
Sagemaker check

abidlabs:
Similar to the Kaggle check, this PR adds a check to determine if the script is running on a Sagemaker instance and if so, creates a share link automatically. Closes: #399

### DIFF
--- a/gradio/utils.py
+++ b/gradio/utils.py
@@ -195,6 +195,16 @@ def kaggle_check() -> bool:
     )
 
 
+def sagemaker_check() -> bool:
+    try:
+        import boto3
+        client = boto3.client("sts")
+        response = client.get_caller_identity()        
+        return "sagemaker" in response["Arn"].lower()
+    except:
+        return False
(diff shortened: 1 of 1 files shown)


gradio/utils.py
```diff
@@ -195,6 +195,17 @@ def kaggle_check() -> bool:
     )
 
 
+def sagemaker_check() -> bool:
+    try:
+        import boto3  # type: ignore
```
freddyaboulton, line 200: Is boto3 installed automatically on sagemaker? Wondering if we should tell users that they may need to install boto3 for us to detect if they are running in sagemaker

freddyaboulton (approved):
This LGTM once CI passes! Thanks @abidlabs !

gradio/utils.py
abidlabs: Yes it's installed automatically

abidlabs:

Issue #3060, opened 2023-01-25
No interface is running right now


tushar-31093:
### Describe the bug

![image](https://user-images.githubusercontent.com/40886260/214615974-7c1809a3-c2d1-4cc7-9a86-c96510fd37b4.png)

Just host anything on EC2 and using the --share parameter the url crashes after a few hours despite the code running on the backend. The problem is consistent with all sorts of implementations.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

![image](https://user-images.githubusercontent.com/40886260/214615974-7c1809a3-c2d1-4cc7-9a86-c96510fd37b4.png)

Just host anything on EC2 and using the --share parameter the url crashes after a few hours despite the code running on the backend. The problem is consistent with all sorts of implementations.

### Screenshot

![image](https://user-images.githubusercontent.com/40886260/214615974-7c1809a3-c2d1-4cc7-9a86-c96510fd37b4.png)

### Logs

```shell
NA.. running it on a non-colab machine. I will run it again with this enabled to update it here.
```

### System Info

```shell
g5dn.4xlarge instance on AWS EC2
```

### Severity

blocking all usage of gradio

freddyaboulton:
Hi @tushar-31093 !

To make sure I understand, the problem is that the link expires after a couple of hours (can you say how many)?

tushar-31093:
I ran 2 instances yesterday around 10 p.m. 

I checked the links today morning and they are not running.

MoonGotArts:
Been running into the same issue with the AUTOMATIC1111 UI for Stable Diffusion, links used to last the claimed 72 hours, but now they go bad at random times, sometimes as little as 1 hour.

tushar-31093:
> Been running into the same issue with the AUTOMATIC1111 UI for Stable Diffusion, links used to last the claimed 72 hours, but now they go bad at random times, sometimes as little as 1 hour.

@freddyaboulton 

Yeah that too. I have Stable Diffusion, GPT-J and Neo running along with a few other prototypes, all use gradio and all are crashing on AWS EC2 instances.

Simultaneously, if i want to use the FastAPI version of loading it and then using it with NGNIX, it doesn't work then either, some config issue for sure. Cant tell.

treksis:
same issue. using paperspace

abidlabs:
Thanks all, we'll look into this next week.

sgtdingo:
Same issue here running Automatic1111

[...]

PR #3149, opened 2023-02-08
Persistent FRP 

abidlabs:
Starting to work on a way to have persistent FRP connections. See [internal Slack discussion](https://huggingface.slack.com/archives/C02SPHC1KD1/p1675807052117969) for context

cc @XciD @Wauplin @aliabid94 

To use this version of Gradio for local testing:
* Check out this branch
* Run `cp gradio/templates/frontend/index.html gradio/templates/frontend/share.html`
* Create a Gradio interface with sharing enabled i.e.:

```py
import gradio as gr

gr.Interface(lambda x:x, "textbox", "textbox").launch(share=True)
```

It will hopefully fix the #3060 issue but I'd prefer to keep the issue open until the community has tested this fix


abidlabs:
Released version `3.18.1b2` for easier testing. To test, do the following steps:

1. `pip install gradio==3.18.1b2`
2. Create a simple gradio demo like above with `share=True`
3. Leave it running for many hours and ensure that that the demo continues to run

gradio/tunneling.py
```diff
@@ -85,7 +86,7 @@ def _start_tunnel(self, binary: str) -> str:
             f"{self.remote_host}:{self.remote_port}",
             "--disable_log_color",
         ]
-
+        print("Command:", " ".join(command))
```
freddyaboulton, line 89: Do we want to keep this?

freddyaboulton (approved):
Tested on colab and my demo stayed up until colab disconnected me for inactivity (about 4 hours)! And the link generated in the colab is still valid.

Also verified that closing and launching and interface will preserve the same URL. 

Awesome!

abidlabs:
> And the link generated in the colab is still valid.

Yeah same for me... dunno how :D

gradio/tunneling.py
abidlabs: Will remove, thanks!

abidlabs:
Has been working well for me too, so I'll go ahead and merge. Let's release a new version soon and have the community test this out. Thanks @Wauplin and @XciD for your help on this PR!

PR #3151, opened 2023-02-08
Fix demos page css and add close demos button

aliabd:
* Adds a close demos button on embedded demos on docs, as its otherwise impossible to exit the window on mobile. 
* Fixes the css leak on the demos page by replacing the stable diffusion demo with fake_gan

Live on dev: https://www.gradio-staging.com/docs/main and https://www.gradio-staging.com/demos


freddyaboulton (approved):
I confirmed the demos page looks as expected and verified that I can close out of `Try Examples` on my phone. Thanks @aliabd !

The one thing I think could be improved for the mobile experience is adding a scrollbar for `Try Examples` cause I can't see all the available demos for some components or navigate to them easily, e.g. the `Try Examples` for `Blocks`. That can happen in another PR though just mentioning it here so I don't forget.

aliabd:
Good catch @freddyaboulton! just fixed in this pr and updated on dev

abidlabs (approved):
LGTM @aliabd!

Issue #3139, opened 2023-02-07
Alphabetic sorting of components in docs


cakiki:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
Not sure there's a thematic ordering that i'm missing, but I always struggle when looking up components in the documentation sidebar as these are not sorted alphabetically.

**Describe the solution you'd like**  
Alphabetical sorting of the components.

**Additional context**  
N/A

PR #3152, opened 2023-02-08
Sort components in docs by alphabetic order

aliabd:
Components in docs are currently ordered as they appear in `components.py` but there doesn't seem to be any specific reason for this order. It's easier for navigation if they're ordered alphabetically. 

Here are the current order and new one side by side: 
  
current-order new-order

I left the other objects in the docs the same order they were in as those seem to be ordered by complexity (ie Interface -> Blocks etc). 

This was a very simple (one-line) change.

Closes: #3139


freddyaboulton (approved):
Thanks @aliabd !

Issue #3092, opened 2023-01-31
Make user's message in a `gr.Chatbot` appear immediately


aliabid94:
- the user's message should appear immediately
- in addition, loading bubbles should appear on the bot side

abidlabs:
I don't know if this is actually possible given that the user message could contain markdown and so does need to be processed in the backend. Perhaps as discussed @aliabid94, we could rewrite the chatbot function itself to yield the user message first, and then return the full response

abidlabs:
@usuyama brought up another good suggestion here: https://github.com/gradio-app/gradio/discussions/3129: the ability add a single message from the bot/user side. I would have expected something like this to work:

`gr.Chatbot([("Hello", None)])`

or 

gr.Chatbot([(None, "Hello")])

But neither of them do. cc @dawoodkhan82

pngwn:
#3156 didn't close this.

abidlabs:
Just to clarify, it didn't close the first bullet point:

> the user's message should appear immediately

But the other two things mentioned here are closed, right?

Yelrose:
> @usuyama brought up another good suggestion here: #3129: the ability add a single message from the bot/user side. I would have expected something like this to work:
> 
> `gr.Chatbot([("Hello", None)])`
> 
> or
> 
> `gr.Chatbot([(None, "Hello")])`
> 
> But neither of them do. cc @dawoodkhan82

Is there any quick way to do this? I have tried to inheritance Chatbot and rewrite the post-processing function to support these. But it doesn't work. 😅

abidlabs:
It should now work as of the latest version of gradio (`gradio==3.19.1`) @Yelrose

abidlabs:
I think making a user's message appear immediately should not be a feature of the chatbot, as much as how a user sets up the chatbot component. Specifically, a user should use the `_js` parameter of the event trigger to pass the user message into the chatbot to have it immediately render. Then, the function should run to get the bot's response and that should then appear underneath the user's response. Note that this will mean that the user's response cannot contain markdown (as their message does not get rendered by the backend) but that should almost always be fine.

I tried to put together a simple example but ran into issues with my knowledge of JS. @dawoodkhan82 do you wanna see if this is doable with Gradio? If so, perhaps we can update our demo / guide with these instructions. If it's not doable, then we can work on making it possible so that we have a really nice UX with the chatbot.

dawoodkhan82:
[...]

PR #3165, opened 2023-02-09
Add ability add a single message from the bot/user side

dawoodkhan82:
# Description
Add ability add a single message from the bot/user side. Plus add ability to specify if the chat starts with bot or user.
`gr.Chatbot([("Hi, I'm DialoGPT. Try asking me a question.", None)], starts_with="bot")`

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

Closes: #3092

# Checklist:

# A note about the CHANGELOG


gradio/components.py
```diff
@@ -3871,6 +3872,7 @@ def __init__(
         """
         Parameters:
             value: Default value to show in chatbot. If callable, the function will be called whenever the app loads to set the initial value of the component.
+            starts_with: Determines whether the chatbot starts with a user message or a bot message. Must be either "user" or "bot". Default is "user".
```
abidlabs, line 3875:
```suggestion
            starts_with: Determines whether the chatbot starts with a user message (appearing on the right) or a bot message (appearing on the left). Must be either "user" or "bot". Default is "user".
```

abidlabs:
Nice @dawoodkhan82! I noticed a few issues when I was testing with `demo\chatbot_multimodal`:

1) If you set `starts_with="bot"`, the top of the first message is cut off by the label:

image

1b) Also it's interesting the *color* of the messages has also switched. In other words, user messages are now gray and bot messages are now orange. I don't think the colors should switch?

2) If you have an initial message (using `None` as one of the entries of the tuple), then it disappears after the next messages show up

Recording 2023-02-09 at 17 07 08

3) If you return a tuple like `("message", None)` from your function, then interestingly you see `null`:

image

3b) For some reason, the markdown / image postprocessing isn't working here^

abidlabs:
(3) makes me think it might not be possible to use `None` as the symbol to indicate that a chatbot message doesn't show up. Perhaps `""` or any falsey string should be what users use to indicate that a chatbot message shouldn't appear. I can't think of any use case where you actually want to show an empty message like that. In which case, we wouldn't even need the `starts_with` parameter. WDYT?

aliabid94:
Hmm I've never really liked the data structure of chatbot because it was so restrictive - simply assuming one chat between a user and bot back and forth, starting with a user. I think we should be considering making the underlying data structure much more flexible, rather than adding these args. Would love a data structure such as:
```json
[
  ["user", "what color is the sky"],
  ["bot", "the sky is blue"],
]
```
 This would make many things possible - setting which message comes first, allowing multiple messages from one party in a row, or even allowing multiple parties in a chat.

Could be set via `gr.Chatbot(type="sender_message")` vs `gr.Chatbot(type="response_pairs")` (can't think of better names for now).

If we don't want to change the data structure for now, I'm not sure we need a "starts_with". Why can't we just have the first message be [None, "I am a bot"]? None seems like the natural way to skip a message, disagree with @abidlabs that we should use an empty string instead.

I think we can just remove the BlockLabel for Chatbot, or make it non-floating, like we do with DataFrame so it doesn't block the chat view (@pngwn might have thoughts on that).

gradio/components.py
dawoodkhan82: yup that's right.

dawoodkhan82:
@aliabid94 I considered changing the underlying data structure completely. It would be a breaking change to all chatbot demos tho (which there is a lot of now). I like the structure of a list of tuples, where the first element is "user" or "bot", etc. and second element is the message. 

When passing `[None, "I am a bot"]` the frontend received `["I am a bot"]` a single element tuple. Since I believe we remove None types.

aliabid94:
Shouldn’t be a breaking change if we use types to keep the formats separate and support both. 
Where are the None’s removed from the output? Just remove that logic.

abidlabs:
@aliabid94 the problem is that we always remove any `None`'s from the config dictionary or any update before we pass it into the frontend -- see the `gr.utils.delete_none()` method. I believe we do this because `None` / `null` behaves weird in JS (@pngwn might remember more details about this). I don't think we should carve out an exception for `Chatbot` specifically. 

_However_, I have an idea @dawoodkhan82 -- which is that in the postprocessing method of the Chatbot, we can programmatically replace any entry of `None` with a special enum such as `gr.components._Keywords.NO_VALUE` and then in the frontend, we treat this special value as null. 

That should allow us to handle these kinds of cases without having to change the data structure. We could even handle the multiple messages from one user in a row case that @aliabid94 is talking about by having successive messages in which the "bot entry" is `None`.

* Standard message: `[("hi", "hi"), ("how are you?", "I'm good")]`
* Bot starts message: `[(None, "welcome to the bot!"), ("how are you?", "I'm good")]`
* Multiple messages from the bot: `[(None, "welcome to the bot!"), (None, "please say something to get started")]`

aliabid94:
I don't understand this utils.delete_none. None gets converted to null in javascript. We need to preserve the Nones. So right now, if a user returns this dict to a gr.JSON component
```python
{"a": 1, "b": None}
```
the "b" key will be removed? This is incorrect behaviour, the null is part of the JSON output.

I thought the original intention of utils.delete_none was when an update() was called, a value with None meant no update. But this recursive removal of None on all output doesn't make sense to me.

pngwn:
We can just add some padding to the chatbot to get rid of the label obstruction, that is what we do with other components. Check `HighlightedText` or `JSON` for an example.

I don't like the `starts_with="bot"` kwarg really. I agree with @aliabid94 that the data structure should be enough to control that. I actually don't love the idea of adding kwargs just so we can change the data structure. I think that the data structure should be:

```
[
  {speaker: bot, message: "blah"},
  {speaker: user, message: "blah"},
  {speaker: user, message: "blah"},
  {speaker: bot, message: "blah"},
]
```

But I don't think it is essential we change it now. If `None` is being removed from _inside_ of a data structure then that is a bug. Only top level 'props' need their None properties removing, that should never happen to the data inside a structure held in a prop, that is potentially very problematic. for example:

```
{
  "props": {
    "prop1": None -> should be removed, not converted to null
    "prop2": [["hi", None], [None, "hi"]] -> [["hi", null], [null, "hi"]]
  }
}
```

dawoodkhan82:
I think for now I'll implement @abidlabs approach. Since it won't require changing the data structure. We can look into  make the underlying data structure into a list of dicts if it makes sense in the future. The only limitation of the current data structure is, it doesn't allow for multiple "users". Which we haven't heard the need for _yet_. I didn't realize the removing of `None` in this case was a bug. Will remove the unnecessary `starts_with` kwarg.

pngwn:
If we just fix the bug, doesn't mean we can simply pass `None` and change the frontend?

abidlabs:
I think so @pngwn. Let me get a fix in for #3183

dawoodkhan82:
> If you have an initial message (using None as one of the entries of the tuple), then it disappears after the next messages show up

@abidlabs This is because the initial message is never saved into the history `state`. Would it be fine to expect the developer to pass in the same default value into the `state` object? Or do you have a better suggestion.

dawoodkhan82:
> 3b) For some reason, the markdown / image postprocessing isn't working here^ but it's also broken on main -- do you experience this as well with demo/chatbot_multimodal?

works for me. When running on the built version of the frontend.

abidlabs:
Let me retest

abidlabs:
Hmm @dawoodkhan82 it's not working for me locally or on the deployed PR Space from this branch: 

image

dawoodkhan82:
> Hmm @dawoodkhan82 it's not working for me locally or on the deployed PR Space from this branch:
> 
> image

Ok I'll take a look at this as a separate issue. Seems like the chatbot just can't locate the tmp file. But also happening on main right?

abidlabs:
> But also happening on main right?

Yep

ui/packages/chatbot/src/ChatBot.svelte
```diff
@@ -94,7 +96,7 @@
 
 <style>
 	.wrap {
-		height: var(--size-80);
```
dawoodkhan82, line 97: This should fix the height issue @yvrjsharma was facing with his demo

gradio/components.py
```diff
@@ -3863,12 +3863,12 @@ class Chatbot(Changeable, IOComponent, JSONSerializable):
     Preprocessing: this component does *not* accept input.
     Postprocessing: expects a {List[Tuple[str, str]]}, a list of tuples with user inputs and responses as strings of HTML.
```
abidlabs, line 3864:
```suggestion
    Postprocessing: expects function to return a {List[Tuple[str | None, str | None]]}, a list of tuples with user inputs and responses as strings of HTML or Nones. Messages that are `None` are not displayed.  
```

abidlabs:
Looks really good @dawoodkhan82! One small issue I noticed was that if a message is `None`, then no message bubble appears but there is a blank space where the bubble would have appeared. 

See this:

image

Or this:

image

I don't think any blank space should appear so that the effect is that of getting multiple messages one after the other

dawoodkhan82:
Oh I see, yeah good catch. Ill fix this

abidlabs:
Side note: the image issue seems to have been fixed!

dawoodkhan82:
Huh how? Was it fixed by another pr

dawoodkhan82:
@abidlabs Fixed the display issue

abidlabs (approved):
LGTM tested & works very nicely @dawoodkhan82!

PR #3167, opened 2023-02-10
Support binary file api

aliabid94:
Very large files uploaded to the File Component were freezing the browser, because of the step the convert the binary data into string dataurls, which is very expensive. So instead, data for files in the File Component is kept in File Blob format, a low level binary format. 
However, we cannot use the existing POST request format for sending File Blobs to the backend, since we send text-encoded JSON objects which do not support binary (which is part of the reason we converted things to dataurls in the first place). So I had to create another backend endpoint for support requests that have binary files in them.

This is basically the current flow of data if an event trigger involves a File Component input:
- FRONTEND:
  - When a file is uploaded, the `value` object of the component stores the file(s) in binary blob format in a property called `blob` 
  - When the event is triggered and values are being gathered from every input component, we check each input component for a `extract_binary` method
  - The presence of this method indicates the Component value has a binary component to it, and the method splits the value into a text and binary component (currently only present in the File component)
  - If any input Component produces binary content (as the File component does here), we create a binary encoded FormData object
  - This FormData has to bundle all the binary files into one array, so we keep track of which binary file corresponds to which input component in another list called input_id_per_file
  - We send this binary `multipart/form-data` object to `/run/binary/predict/`. This is in contrast to the string `text/json` object that would be sent to `run/predict` if our payload had no binary content,
- BACKEND W/ POST REQUEST
  - Once we hit `/run/binary/predict/`, we receive three things:
    1. The non-binary JSON part of the data
    2. The list of binary files sent in the request
    3. A list tracking which binary file corresponds to which input component
  - We need to merge all this into one payload to pass to the standard predict endpoint.
  - We do this by calling the `incorporate_binary_data` method of any component that accepts binary data (currently only File), which takes the binary and non-binary parts of the data and merges them into one object.
    - For the File component, this is done by saving the binary content to the disk, and then editing the filename in the data to point to the on disk file instead.
  - Now we can call the standard `/predict` endpoint internally with this data
- BACKEND W/ WEBSOCKET
  - Slightly more complicated when websockets are involved
  - Cannot combine non-binary and binary data into formdata and send via websocket
  - Instead, first send non-binary data, which includes info on how many binary files to expect
  - Then sends binary files one by one, saving them to disk
  - Then runs the `incorporate_binary_data` as before


pngwn:
If this is going to require additional endpoints, does it have implications for users using gradio as an API? 

I'll need some time to review this properly but do we definitely want to add this before 4.0 considering it is such a substantial behaviour change? I think we should probably handle most file types in the same way and send to the server immediately, especially if it will create API discrepancies.

aliabid94:
This has backwards compatibility such that the backend can still take in the File object as a string via the regular URL, so API does not break.

pngwn:
But do users need to use a new API in order to get the binary version?

aliabid94:
There is a new API if you want to use the binary endpoint (which the frontend will do in the PR), but the non-binary endpoint is still supported for files, where you send the file as a data url. So an older frontend / client will be able to query a newer backend.

pngwn:
Cool. Will review properly next week.

gradio/routes.py
```diff
@@ -448,12 +458,42 @@ async def predict(
...
                 request=gr_request,
             )
             return result
 
+        @app.post("/run/binary/{api_name}", dependencies=[Depends(login_check)])
+        async def predict_binary(
+            api_name: str,
+            request: fastapi.Request,
+            payload: str = Form(...),
+            binary_files: List[UploadFile] = File(...),
```
freddyaboulton, line 472: Something to discuss at the synch: Did you try adding `binary_files` argument to the standard `run/predict/` route? Think would be simpler for api users and us as maintainers if there was just one route.

Issue #2415, opened 2022-10-07
Split Docs into multiple pages so they can be indexed by google


aliabd:
A common workflow for users is to search the query on google, but then the results only lead to the docs page where they have to search the query again. It might be better to split the docs into multiple pages (which is not uncommon practice, for example: [tailwind](https://tailwindcss.com/docs), [huggingface](https://huggingface.co/docs/tokenizers)) so google can index them.

pngwn:
We need a full search solution before we can do this really because it breaks ctrl + f. I’m also not 100% certain it is necessary, I think most people will go to the Gradio site to search for Gradio solutions. 

I think a full site search that searches docs and guides would be more valuable but is a prerequisite for this anyway really.

PR #3179, opened 2023-02-11
Splitting Docs into multiple pages

aliabd:
Splitting the docs into multiple pages so that: 

* Users can land in the correct page from a google search (See #2415) 
* Pages load much faster 
* We can simplify the original docs page to make navigation on them easier.

Still a WIP. Need to figure out how to handle main vs pip on all the individual pages, and need to iron out some issues and refactor a bit. 

Please try stuff out and give me feedback 🙏 

https://www.gradio-staging.com/docs/main/
https://www.gradio-staging.com/docs/textbox/

I'm struggling in particular with figuring out what should go in the individual pages and what should be on the full docs page, and how users should move between them. 

Also thoughts on removing the full docs page entirely? I personally think we should just switch to individual pages, and will feel the same to the user navigating through the sidebar. (for example [tailwind](https://tailwindcss.com/docs/margin) does it this way). Only downside of that is ctrl+f search won't work, so we would need to add some really good search (which I'm planning to do)


freddyaboulton (commented):
This is pretty nice @aliabd ! Looks good on mobile too.

I have some nits about how each component subpage looks, for your consideration. I think this is stuff that predates this PR so we can do them in a separate PR as well.
 
1. We should add a section title above each separate part of the page to make it easier to navigate. In the screenshot below, I added "Behavior" and "Initialization" headers which I think explains better what each section is about. 
2. The Try examples button looks a bit out of place in the screenshot below. 
3. Each of the sections should be linkable too.

image

4. I think `Class` and `Methods` are too small font, especially since the method name, which is a subsection, is much larger.
5. Maybe we can find another place for the info in the "Class" section? Maybe with the section that has the `As Input` and `As Output` parts.

image


6. Once you're at the examples, would be nice to jump back up to the top of the page.

PR #3186, opened 2023-02-13
Automatically cache Examples in Spaces

osanseviero:
# Description

To be consistent with `gr.Interface`, this PR will automatically cache the examples also with `gr.Examples` when working in Spaces. This could potentially break some Spaces if upgrading to newer versions as the current requirement for caching is that `gr.Examples` provides a `fn` and `outputs`, so not sure if we want this

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@ No changes to highlight.
 No changes to highlight.
 
 ## Breaking Changes:
+No changes to highlight.
 
 ## Full Changelog:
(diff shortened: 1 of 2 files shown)


abidlabs:
We should definitely not cache examples if `fn` and `outputs` are not provided, and thinking about this some more, I do feel like too much is happening under the hood if we only cache if `fn` and `outputs` are provided, and the app is running on Spaces. Perhaps we just accept the discrepancy and close this PR.

cc @freddyaboulton

abidlabs:
Discussed with @freddyaboulton @aliabid94 and we don't really see a path forward in which this can happen without too much magic. Closing, sorry @osanseviero for not clarifying earlier

Issue #3183, opened 2023-02-12
Make `gr.utils.delete_none` only delete props that are `None`, not values that contain `None` 


abidlabs:
### Describe the bug

Right now, the implementation of `gr.utils.delete_none` deletes any `None` found anywhere inside of a config dictionary, even if it is nested inside of a tuple / dict / etc. This implementation is buggy, as discussed here: [internal Slack](https://huggingface.slack.com/archives/C02SPHC1KD1/p1676067662339889)

We should rewrite this function

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

N/A

### Logs

```shell
N/A
```

### System Info

```shell
All
```

### Severity

serious, but I can work around it

PR #3188, opened 2023-02-13
Fixes `gr.utils.delete_none`

abidlabs:
Fixes: #3183 

A few tests needed to be updated, but the demos all seem to work fine with this update!

You can see the effect of this PR by running:

```py
import gradio as gr

with gr.Blocks() as demo:
    gr.JSON({'a': 'hello', 'b': None})
    
demo.launch()
```

on `main`:

image

on this branch:

image


test/test_external.py
```diff
@@ -329,7 +329,7 @@ def test_loading_files_via_proxy_works(self):
         app, _, _ = io.launch(prevent_thread_lock=True)
         test_client = TestClient(app)
         r = test_client.get(
-            "/proxy=https://gradio-tests-test-loading-examples-private.hf.space/file=/tmp/tmprahzj703/Bunnyual53t2x.obj"
+            "/proxy=https://gradio-tests-test-loading-examples-private.hf.space/file=/tmp/tmpdd31pyri/Bunnyp0c5zw__.obj"
```
abidlabs, line 332: Unrelated test that needed to be fixed because the Space went to sleep. I've upgraded the Space so that it should stay running and this test should pass in the future.

freddyaboulton (approved):

Checked the deployed demos and ran some other demos locally and it all works great 👍

dawoodkhan82 (approved):
Nice, thanks for this. I merged this to my chatbot branch to check, and I do indeed get `null` back to the frontend.

abidlabs:
Great thanks for reviewing guys! I'll merge this so @dawoodkhan82 you can use it for the chatbot PR

Issue #2421, opened 2022-10-10
Large input files freeze/crash the interface


ValentinKoch:
### Describe the bug

Hi, In my usecase a user uploads a rather large file (300-500mb) and it freezes the app for several seconds which makes for a very bad user experience. It happens after selecting a large file to a gradio File block. It freezes again when pressing a button that processes this data. Is there a way to circumvent this behaviour? It is running locally, as far as a I know are those sizes not supported anyway non-locally right? Currently am hosting my app on AWS. Would be cool if you could fix this behaviour, very much appreciated :)

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction
```
def test():
    return

with gr.Blocks() as demo:
    inp = gr.File(label="Input")
    vis_out=gr.File(label="output")
    with gr.Row():
        btn = gr.Button("Run")
    btn.click(fn=test, inputs=[inp], outputs=[vis_out])
demo.launch() 
```
Then just add a large file as the input.

### Logs

```shell
no logs
```

### System Info

```shell
gradio v 3.4.0, occurs with different os on which the app is hosted,e.g. linux and mac 
also across browsers, chrome takes a long time, mozilla even crashes
```

### Severity

serious, but I can work around it

pngwn:
Could you give some more info about what you'd like to do with these large files when you have them? How we handle the files that are loaded by users dictates what you can do with it. 

I haven't looked into this too much but I'm guessing that this is happening because we load the file into memory, in full. If the file were bigger 1GB+ it would probably consistently crash the browser.

ValentinKoch:
Also it takes very long until this file is uploaded, not sure if this is related. The demo i am doing will be used in a clinical study and speed really matters. Would be awesome if you could help me out there :) would be also happy to promote gradio in some way!

pngwn:
In this case we can (probably) chunk/ stream the data from the client without needing to store the whole thing in memory at any point. This would resolve the browser issues. I think it would still take a while to upload but maybe something else is going on that is slowing that down too. 

The challenge is figuring out when to hold things in memory and when not to. There are only circumstance in which _not_ holding things in memory makes sense, from gradio's point of view.

ValentinKoch:
any feedback on this @pngwn ?

ValentinKoch:
or @abidlabs ? would be very much appreciated

ValentinKoch:
[...]

Issue #2718, opened 2022-11-25
Form validation


apolinario:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
There's no form validation on Gradio

**Describe the solution you'd like**  
Implement basic form validation for form fields. Specifically, a suggestion would be to add a `mandatory` parameter to all input fields, if `true` and one tries to submit without filling that in, they get a front-end error without actually submitting

**Additional context**  
Implementing form validation increases complexity on apps, I feel that could be sorted on Gradio side

julien-c:
yes i've had same feedback before (`required` prop on inputs)

pngwn:
We used to have something similar and then it was removed in 3.0 and then i was going to reimplement recently and then we decided not to. But i can't find the issues where this happened or remember why. @abidlabs Can you remember why we decided not to?

I agree though, I think it is a useful feature. Do we want more complex input validation here or just 'required' fields? 

We could easily accept a regex as a string, for validation purposes, a function would be more difficult because then the validation wouldn't happen on the frontend. Realistically, we would _also_ want to validate on the backend because bypassing the frontend checks would probably be pretty easy. But i do think we should _also_ validate on the frontend if possible just to keep things fast.

`required` could take a bool or a string: `required=True` `required="[a-zA-Z0-9]+"`

abidlabs:
Yes, definitely doable. The only tricky thing is whether or not a component is required may depend on what event trigger is being called. I.e. you could have 2 events, and in one of them, a component is required, but in the other, the component is not required. How we fit this into the Python API I'm not sure cc @aliabid94 @freddyaboulton

apolinario:
Great point @abidlabs, then maybe the required props don't live on each component - but rather on the event trigger (`button.click` for example), could have a list of which components are mandatory

> Realistically, we would also want to validate on the backend because bypassing the frontend checks would probably be pretty easy. But i do think we should also validate on the frontend if possible just to keep things fast.

@pngwn IMO would be fine to split the two tasks and start with front-end validation only. I am aware it is easy to meddle with front-end validation (also for existing limitations e.g.: sliders https://github.com/gradio-app/gradio/issues/1071), but I think just front-end may be simpler and solve a big pain-point

freddyaboulton:
[...]

PR #3191, opened 2023-02-14
Support large files try 2

aliabid94:
Rewrote large file issue fixer with @freddyaboulton's approach. When a file is uploaded to FileUpload, it is automatically sent to the backend, and then the frontend receives a reference to the name of this new file. Much cleaner, thanks @freddyaboulton!

Fixes: #2421


aliabid94:
ready for review btw

gradio/routes.py
```diff
@@ -509,6 +511,27 @@ async def join_queue(
...
+            files: List[UploadFile] = File(...),
+        ):
+            output_files = []
+            for input_file in files:
+                if "." in input_file.filename:
+                    prefix, suffix = input_file.filename.rsplit(".", 1)
+                else:
+                    prefix = input_file.filename
+                    suffix = ""
+                output_file_obj = tempfile.NamedTemporaryFile(delete=False, suffix=f".{suffix}", prefix=f"{prefix}_")
```
abidlabs, line 525: Nit: if `suffix` is `""`, this will add a `.` to the end of the temporary filename

Note, there might also be issues with the filename containing invalid characters, etc. I would reuse the implementation of `TempFileManager.get_prefix_and_extension()` (maybe make that a classmethod) where we have already handled this logic

test/test_routes.py
```diff
@@ -54,6 +54,15 @@ def test_get_config_route(self, test_client):
...
         assert response.status_code == 200
 
+    def test_upload_route(self, test_client):
+        response = test_client.post(
+            "/upload", files={"files": open("test/test_files/alphabet.txt", "r")}
+        )
+        assert response.status_code == 200
+        files = response.json()
+        with open(files[0]) as saved_file:
+            assert saved_file.read() == "abcdefghijklmnopqrstuvwxyz"
```
abidlabs, line 64: You may want to add a test to ensure that the temporary file name is what you expect

I also I would add a test containing a larger file. I guess it doesn't have to be massive but maybe more than 1024 bytes so that we know that the chunking is accurate.

abidlabs:
Really cool PR @aliabid94! I tested it locally and on Spaces 

**Locally, I ran the following demo:**

```py
import gradio as gr

def test(x):
    return x.name

with gr.Blocks() as demo:
    inp = gr.File(label="Input")
    vis_out=gr.File(label="output")
    with gr.Row():
        btn = gr.Button("Run")
    btn.click(fn=test, inputs=[inp], outputs=[vis_out])

demo.launch() 
```

**Feedback:**

* The UI looks good:
image
* The demo works well for small file sizes, except that the download link on the input file component is broken:

image

Also I see this error in the js console but it doesn't seem to have a detrimental effect:

image

Details of the error:
```
{"detail":[{"loc":["body","files",0],"msg":"Expected UploadFile, received: <class 'str'>","type":"value_error"}]}
```

* However, if one clicks on the submit button while the `File` component is still uploading the file (which for me takes about 10 seconds for a 5 MB file by the way), then an empty file is sent to the backend, resulting in this:

image

* As we discussed briefly in our sync, I think the `File` component would need to "block" the submit button from happening until the file has been fully uploaded. When someone tries to submit, I suppose we would need to "signal" that the file component is still uploading and that's why the event cannot start. This could be similar to whatever UI affordance we use to designate that a field is required (#2718) 
* As a side note, I don't find the orange border indicator as too visually obvious that the file is uploading. Perhaps the "Download" link could be replaced with "Uploading..." in italics until the file finishes uploading?
* For larger file sizes, it just takes *forever* to upload the files locally. I'd say it took about 5-10 minutes for a 200 MB file, which seems kind of weird since everything is local

**on Spaces, using the deployed PR link, I looked at the `zip_files` demo**:

* Generally the same feedback as above, although in the case of the `zip_files` demo, neither the input file nor the output file download link worked for me

**Other notes**

* Finally, I wonder if we should add a `max_size` parameter to govern the maximum file size that someone can submit and set it to a reasonably small default like 100 MB. Otherwise, given the relatively small disk space on Spaces (50GB), it could easily get filled up

----

The critical questions imo are:
* why is the upload speed so slow?
* why are the download links broken?
* how do we prevent users from triggering events before the file has finished uploading?
 
Happy to give this another look!

freddyaboulton:
> I think the File component would need to "block" the submit button from happening until the file has been fully uploaded.

Without this PR, users can still hit submit with missing components and get errors, so I don't think we should increase the scope of this PR to include #2718.

I agree that we should make the "uploading" state more obvious though! Perhaps we can grey out the component and add "Uploading ..." with a spinner?

This is a really awesome PR @aliabid94 ! I agree with all of @abidlabs other comments. The one thing I noticed is that with cached examples, the file names change when you hit `Submit` 

upload_cache_file_change_name

abidlabs:
> Without this PR, users can still hit submit with missing components and get errors, so I don't think we should increase the scope of this PR to include https://github.com/gradio-app/gradio/issues/2718.

Makes sense, agreed. Then it's just important to make sure the component makes it very clear visually when it's in the uploading state. 

> The one thing I noticed is that with cached examples, the file names change when you hit Submit

This is actually an older issue I believe, not introduced in this PR. I see it here as well: https://huggingface.co/spaces/gradio-pr-deploys/pr-3149-all-demos

aliabid94:
Ok made the following changes:
- should upload much faster! (~200MB/s locally for me) problem was I was reading the stream 1 kb at a time, now reading 100 MB at a time. 
- fixed console error issue
- uploaded files are downloadable
  - added an uploaded_file dir that keeps uploaded files
- use TempFileManager in uploads
- Shows "Uploading..." while uploading. Since it's so fast now, not too urgent to make it look better.

will set max size and block submits on upload in another PR.

gradio/routes.py
```diff
@@ -292,17 +295,16 @@ async def file(path_or_url: str, request: fastapi.Request):
...
-                *blocks.temp_file_sets
-            )
+            created_by_app = abs_path in set().union(*blocks.temp_file_sets)
             in_file_dir = any(
                 (
                     utils.abspath(dir) in utils.abspath(path_or_url).parents
                     for dir in blocks.file_directories
                 )
             )
+            was_uploaded = abs_path.startswith(app.uploaded_file_dir)
```
abidlabs, line 305: Is `startswith()` robust? I think using the `in` operator with two `Path`s is more robust

gradio/routes.py
```diff
@@ -292,17 +295,16 @@ async def file(path_or_url: str, request: fastapi.Request):
                 )
             abs_path = str(utils.abspath(path_or_url))
             in_app_dir = utils.abspath(app.cwd) in utils.abspath(path_or_url).parents
-            created_by_app = str(utils.abspath(path_or_url)) in set().union(
-                *blocks.temp_file_sets
-            )
+            created_by_app = abs_path in set().union(*blocks.temp_file_sets)
```
abidlabs, line 298: Thanks for fixing

abidlabs:
Beautiful, LGTM @aliabid94! I left a couple of nits on the backend code and I'd suggest @pngwn or @dawoodkhan82 do another pass on the svelte, but works beautifully for me!

abidlabs (approved):

gradio/routes.py
aliabid94: No, it's the opposite - the `in` keyword is not robust. For example, if the upload dir is "a/b", and a user tries to access file "c/d/a/b", the `in` keyword will allow access to this. This is not secure. We should modify the other lines to use `startswith` as well - I can do this is in this PR

abidlabs: Not `in` with two strings, but `in` with two `pathlib.Path` objects. See here: https://stackoverflow.com/a/34236245/5209347

pngwn (approved):
Frontend changes look good to me.

gradio/routes.py
aliabid94: oh yep you're right

Issue #3082, opened 2023-01-30
Improve error messages when returning incorrect number of values from an event handler


space-nuko:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
I have a lot of event handlers with a large number of values returned. If something goes wrong, like if I send the wrong number of return values, I get this error

```
Traceback (most recent call last):
  File "C:\build\stable-diffusion-webui\venv\lib\site-packages\gradio\routes.py", line 337, in run_predict
    output = await app.get_blocks().process_api(
  File "C:\build\stable-diffusion-webui\venv\lib\site-packages\gradio\blocks.py", line 1018, in process_api
    data = self.postprocess_data(fn_index, result["prediction"], state)
  File "C:\build\stable-diffusion-webui\venv\lib\site-packages\gradio\blocks.py", line 935, in postprocess_data
    if predictions[i] is components._Keywords.FINISHED_ITERATING:
IndexError: tuple index out of range
```

Now from trial and error I know the issue stems from a code mistake in an event handler, but I have no idea where it is because the traceback isn't placed within my code. There isn't any context or other things to help me locate the error, just the (also unclear) implication that there's a faulty event handler somewhere

**Describe the solution you'd like**  
It would at least be helpful to know the function name and the number of arguments it expects. Maybe printing out what the arguments are with `pprint` would be useful too.

**Additional context**  
N/A

Issue #3077, opened 2023-01-28
Filter out files based on file_types when choosing the `directory` option


Dsantra92:
- [x] I have searched to see if a similar issue already exists.

Currently `gr.File(file_count="multiple", file_types=["image"])` checks if the file is an image before uploading. I was hoping for a similar behavior for `gr.File(file_count="directory", file_types=["image"])` where using the `file_types` option with `directory` should selective upload files with matching types. 

With the current implementation, mentioning `file_types` with `file_count=directory` doesn't have effect.

freddyaboulton:
To be sure, `gr.File(file_count="multiple", file_types=["image"])` works as expected though, right? It only let's you upload images.

Dsantra92:
> To be sure, `gr.File(file_count="multiple", file_types=["image"])` works as expected though, right? It only let's you upload images.

Yes, it works as expected.

abidlabs:
It feels quite niche to me to support filtering file types when a directory is uploaded. Perhaps we should just have a warning displayed if someone uses `file_types` with file_count="directory". @dawoodkhan82 @freddyaboulton thoughts?

ROBERT-MCDOWELL:
@abidlabs 
ok so does it mean that if directory is selected the user can upload whatever files?
also if I set the component dynamicall from "single" to "directory", I have not way to remove the warning saying "file_types" is ignoredd... 
if multiple files can be filtered, the directory option should too to avoid upload abuse. 
make a condition that only the files allowed weill be in the list.....

abidlabs:
cc @dawoodkhan82

ROBERT-MCDOWELL:
would you like I open a new issue on this subject?
to see "E:\Z______MYDATA\DEV\PYTHON\ebook2audiobook\python_env\Lib\site-packages\gradio\components\file.py:113: UserWarning: The `file_types` parameter is ignored when `file_count` is 'directory'.
  warnings.warn(" everytime we select 'directory' as file_count is frustrating... 
to not filter like multiple files and not limit the total size of uploads is going to create a massive security breach through gradio application IMHO....

ROBERT-MCDOWELL:
kk

PR #3194, opened 2023-02-14
Miscellaneous warnings and errors 

abidlabs:
Just a few improvements to our warnings and errors:

* Raises a warning in the `File` component or `UploadButton` component if someone passes in a `file_types` parameter along with `file_count=="dir"` (closes: #3077) 
* Raises a `gr.Error` instead of a regular Python error when you use `gr.Interface.load()` to load a model and there's an error querying the HF API to make it more easily readable (closes: #3194), looks like this:

image

* Improves error messages when returning incorrect number of values from an event handler (closes: #3082)


CHANGELOG.md
```diff
@@ -53,6 +53,8 @@ No changes to highlight.
 
 ## Full Changelog:
 * Fix demos page css and add close demos button by [@aliabd](https://github.com/aliabd) in [PR 3151](https://github.com/gradio-app/gradio/pull/3151)
+* Better warnings and error messages (e.g. when there is a mismatch between the number
```
freddyaboulton, line 56: Nit: Mention the other issues we're closing as well

test/test_blocks.py
```diff
@@ -531,6 +533,15 @@ def generic_update():
             }
             assert output["data"][1] == {"__type__": "update", "mode": "dynamic"}
 
+    def test_error_raised_if_num_outputs_mismatch(self):
+        with gr.Blocks() as demo:
+            textbox1 = gr.Textbox()
+            textbox2 = gr.Textbox()
+            button = gr.Button()
+            button.click(lambda x: x, textbox1, [textbox1, textbox2])
+        with pytest.raises(ValueError):
```
freddyaboulton, line 542: nit: I would `match=` against the expected error message here but maybe I'm just too paranoid 😂

freddyaboulton (approved):
Looks great to me @abidlabs !! Thanks

abidlabs:
Thanks for the review @freddyaboulton! Agree with the suggestions and have made the updates

Issue #3189, opened 2023-02-13
Cache Data on the frontend side


Anwarvic:
- [x] I have searched to see if a similar issue already exists.

## Problem
Right now, I'm building an application where the user records a video and pass the video to my model. My model first needs to preprocess the recorded video and then use the processed version for inference. The pre-processing step takes long time and it would be better to know if the video has been recorded before or it's a new record.

## Suggested Solution
My suggested solution for this case is to return the same video temp path as long as it's the same recorded video. In this way, I will know that this video was recorded earlier (using the temp path) and i can use the preprocessed version directly without preprocessing the whole video again.

## Example
The following is a simple interface to explain more what I mean:
```
def dummy(model_type, webcam_video):
    print(webcam_video)
    return webcam_video


iface = gr.Interface(
    fn=dummy,
    inputs=[
        components.Dropdown(
            label="Model", choices=["model1", "model2"]
        ),
        components.Video(
            source="webcam", include_audio=True, mirror_webcam=False
        ),
    ],
    outputs=[
        components.Video(
            label="Lip Movement", source="upload", include_audio=True
        ),
)
iface.launch(debug=True)
```

Now, if the user decides to record a video and presses "submit", the `webcam_video` variable will hold the video's temp path (e.g: `/tmp/sample0b7j8h5q.webm`). When the user presses "submit" again without recording a new video, a new temp path will be assigned (e.g `/tmp/sampleascvxtdq.webm`) even though it's the same video as before (i.e `/tmp/sample0b7j8h5q.webm`).

IMO, that is not a desired behavior. It would be better to return the same path since it is the same video. Also, this can be applied to other components, i.e `Audio`.

PR #3197, opened 2023-02-15
Cache temp files created from base64 data

abidlabs:
This PR slightly changes the way that we generate temp files from base64 input data by giving them a deterministic path based on the contents of data. In other words, if a user submits the same input twice, it will be produce the same temp file. If the temp file already exists, then the conversion from base64 to binary (which is very time consuming) can be skipped and the file can be returned directly. 

Fixes: #3189

Probably the easiest way to test this would be to run a demo like this:

```py
import gradio as gr

gr.Interface(lambda x:x, gr.Video(source="webcam"), gr.Video()).launch()
```

* Record a video that is at least a few seconds long
* Click the submit button (note the inference time)
* Click the submit button again with the same video (note the inference time, should be much faster on this branch)


test/test_processing_utils.py
```diff
@@ -194,6 +194,30 @@ def test_make_temp_copy_if_needed(self, mock_copy):
         )
         assert len(temp_file_manager.temp_files) == 2
 
+    def test_base64_to_temp_file_if_needed(self):
```
freddyaboulton, line 197: I would test the case of the same audio file but with different min/max values. That's an interesting case I would not have thought of!

freddyaboulton (approved):

gradio/components.py
```diff
@@ -2041,20 +2037,26 @@ def preprocess(
...
 
         sample_rate, data = processing_utils.audio_from_file(
             temp_file_path, crop_min=crop_min, crop_max=crop_max
         )
 
+        # Need a unique name for the file to avoid re-using the same audio file if
+        # a user submits the same audio file twice, but with different crop min/max.
+        temp_file_path = Path(temp_file_path)
+        output_file_name = str(
+            temp_file_path.with_name(
```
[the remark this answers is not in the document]
abidlabs: I had originally used `with_stem()` but turns out it's not compatible with Python 3.7-3.8: https://stackoverflow.com/a/67612849/5209347

test/test_processing_utils.py
abidlabs: Good idea thanks!

aliabid94:
As I understand it, on the frontend side, we still have to upload any file. The only step we save is on the backend, where we don't have to convert the base64 stream back to a binary file.

While this PR does improve things, a better approach would be to do caching on the frontend side. Because right now, let's say you have a live interface with a Video and Slider input. If you attach a video, as you move the slider, the video gets sent every single time you move the slider. If the frontend can keep track of previously uploaded files and the file names of those in the backend (which should be very feasible with PR #3191), then we only even send the file to the backend if its a new file, which saves a lot more processing and network time.

abidlabs:
Agreed @aliabid94 that would be the optimal solution. However, which components are we planning on doing this large file upload caching with? I thought we were only applying it to the `File` and `UploadButton` components -- is it reasonable to apply the same approach to all of the other components affected by this PR: `Video`, `Audio`, `Model3D`?

test/test_processing_utils.py
abidlabs: Added a test in `test/test_components`

abidlabs:
Synced with @aliabid94 and we'll merge this in while @aliabid94 works on a proper fix via the upload logic in #3191

abidlabs:
Thanks for the review and feedback guys!

Issue #3083, opened 2023-01-30
Enter Chinese in the DataFrame component，..


WangJ970316:
### Describe the bug

https://user-images.githubusercontent.com/48046091/215408060-c4699e01-d9db-4b24-8d59-0bab5baf25eb.mov

When I enter Chinese in the DataFrame component, each keyboard entry will be retained in the text box, but I haven't completed the entry yet.
When I enter the contents of headers, there is no such problem

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

As shown in the video

### Logs

```shell
As shown in the video
```

### System Info

```shell
As shown in the video
```

### Severity

annoying

xianbaoqian:
I think this might be related to the handling of this event. https://developer.mozilla.org/en-US/docs/Web/API/Element/compositionstart_event

WangJ970316:
> I think this might be related to the handling of this event. https://developer.mozilla.org/en-US/docs/Web/API/Element/compositionstart_event

I read the content of the problem carefully, but I still don't know how to solve the problem in the video

aliabid94:
Fixed! Should be in the next release of gradio.

xianbaoqian:
Thanks @aliabid94 !

PR #3206, opened 2023-02-15
Fix Table entry for Pinyin

aliabid94:
Right now, entering Pinyin in a Dataframe is not possible - see #3083. The issue is because when a cell is being edited in EditableCell.svelte, its value is bound to the array from the parent Table.svelte. So every edit causes the table to rerender, resetting the value of the input field with pinyin. This treats the next letter as a completely new entry, breaking the pinyin chain of letters. Now, the value is only updated when the input blurs, meaning when a user has completed input.

To test, install the Chinese pinyin keyboard, and enter the keys in order: `"w", "o", "1"`. Originally, this would not get converted into 我, but it should now.

I also believe this should improve performance as the table does not rerender on every edit, only when the cell finishes editing.

Fixes: #3083


pngwn (approved):
lgtm!

PR #3212, opened 2023-02-16
Fix bokeh Plots

### LINKED ISSUE #1942: Bokeh plots do not appear
### Describe the bug

When using the `Plot` component, bokeh plots do not appear at all.

This is documented here: #1632 

> Pictures cannot be generated in Bokeh mode.

And in this PR: #1609 

> The Bokeh plots are currently broken. The reason for this is that bokehJS internally uses getElementById to get the container of the plot and render it. Since Gradio UI is using the shadow DOM to render, this step fails.

> I have tried here a workaround where I added a new hidden div in the index.html file to use as a helper to render the plot, once it is rendered, then the content is appended to the actual div that should have the plot. This part is all working, unfortunately although I can see the div with the expected content, the plot is still not showing.

The PR was closed because it was an attempt to fix the problem using an older version of bokeh. We still need to find a solution using the latest version of bokeh.

### Reproduction

https://github.com/gradio-app/gradio/blob/main/demo/outbreak_forecast/run.py

### Screenshot

_No response_

### Logs

```shell
N/A
```

### System Info

```shell
3.1.3
```

### Severity

freddyaboulton:
# Description

Fixes #1942

### Introduces the following changes:
* Fixes serialization of bokeh plots on the backend
* Removed hardcoded bokeh versions from the front-end so we can support both bokeh 2.x and 3.x. I tested with the latest bokeh 2.x version (2.4.3) and 3.x version (3.0.3).
* Fixed bug where bokeh plot was showing up more than once.

You can test with `bokeh_plots` demo! The variable names in the altair_plot and bokeh_plots demo conflict so I will fix in a separate PR and deploy to spaces.

### Examples

Bokeh
bokeh_works

Matplotlib
matplotlib_works

Altair

altair_works

Plotly

plotly_works

# Checklist:

# A note about the CHANGELOG


ui/packages/plot/src/Plot.svelte
```diff
@@ -32,8 +33,29 @@
 
 	$: darkmode = theme == "dark";
 
-	$: if (value && value.type == "altair") {
-		spec = JSON.parse(value["plot"]);
+	$: plot = value?.plot;
```
freddyaboulton, line 36: When I was fixing bokeh, I noticed that the DOM was updating more than once for each event trigger. This was actually causing multiple copies of the plot to show up. 

I figured it was because all the reactivity was based on `value` and `value` is an object so I thought there was some unintended mutation of that object. Unpacking `value` into it's component parts fixed the problem and there is only one DOM update every time the value changes from the backend as far as I can tell.

gradio/components.py
```diff
@@ -4127,7 +4127,17 @@ def __init__(
         )
 
     def get_config(self):
-        return {"value": self.value, **IOComponent.get_config(self)}
+        try:
+            import bokeh
+
+            bokeh_version = bokeh.__version__
+        except ImportError:
+            bokeh_version = "3.0.3"
```
freddyaboulton, line 4135: It doesn't matter which version is the "fallback" version since there can't be a bokeh plot if bokeh is not installed so it's not like the front-end and backend bokeh versions would be out of synch.

abidlabs: Why do we need a fallback version anyways as opposed to being `None`?

freddyaboulton: I should test this but my understanding was that the front end always loads the bokeh assets from the cdn, in which case we None wouldn’t work

abidlabs: Ah I see, but maybe we don't have to import the bokeh assets at all if bokeh is not installed?

abidlabs:
Works great @freddyaboulton! We should have someone else take a look at the svelte changes, but overall behavior LGTM. 

Couple o' nits about the demo:
* Initially I tested `bokeh_plots` demo with `bokeh==2.4.2` and the Whisper plot produces a `ValueError` (the other two plots work fine). Same is true for `2.4.3`, so I think we may need to pin the requirements to install `bokeh>=3.0` for all the plots to render correctly
* When the demo loads, no plot shows up. It would be a nicer UI to have the `scatter` plot show up by default

demo/bokeh_plot/run.py
```diff
@@ -0,0 +1,93 @@
+import gradio as gr
+import xyzservices.providers as xyz
+from bokeh.plotting import figure
+from bokeh.tile_providers import get_provider
```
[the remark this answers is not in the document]
freddyaboulton: yea I'm using the demo from the issue which was written in version 2. I think it's fine to keep as is for now.

freddyaboulton:

I those improvements to the demo and figured out how to skip loading bokeh on the frontend if it's not installed in the backend.

Will wait for review on the svelte code before merging!

pngwn (approved):
lgtm!

freddyaboulton:
Thank you so much for the review @pngwn !

PR #3220, opened 2023-02-17
Added Bounding-Box tool for Image Component

CtrlAltDeplete:
# Description

This change adds a new tool to the Image component: boxes. This tool allows users to draw boxes on an image as part of the input, which are passed along as an Array of Arrays of number (top, left, bottom, right).

This is useful in applications where one image has multiple points of interested, and needs to be broken down into multiple inputs, whereas the existing crop would require users to upload the same image for each input. Additionally, this could be used in a similar way as the mask capability but gives users a box-shaped tool, so they don't have to draw shaky lines.

# Checklist:


abidlabs:
Thanks for opening this PR @CtrlAltDeplete! We will take a look at this and get back on the next steps

pngwn:
Hey @CtrlAltDeplete, thanks for opening this PR. We appreciate the effort you haver invested in this but we have decided not to add this feature at the minute while we decide what the feature set of the Image component should be.

In future it would be a good idea to open an issue and make a proposal before starting work.

Thanks for understanding.

PR #3221, opened 2023-02-17
FRP version fix

abidlabs:
When we fixed persistence of FRP links in PR #3149, we upgraded the version of the FRP client binary file (from v0.1 to v0.2) on the Hugging Face CDN. This is the file that is downloaded when a user creates a share link. However, it is not downloaded if it is already present on the user's file system. Meaning that when if a user upgrades their version of Gradio, Gradio will not use the latest version of the FRP client.

That is why the beta release (`gradio==3.18.1b2`) fixed the FRP problem when testing in a fresh environment like Colab, but did not seem to solve the problem for users testing it as part of the discussion in #3060. 

This fixes that problem by attaching the version # to the downloaded file. 

In order to test this, it would be great if you could *locally* do the following:
* Don't use a fresh environment but instead **in your current gradio environment, upgrade** to `gradio==3.18.1b7` (built off of this branch, so you don't have to check out this branch) 
* Create a Gradio demo locally with a share link
* Confirm that a share link is successfully created!


freddyaboulton (approved):
Thanks @abidlabs !

abidlabs:
Thanks for testing @freddyaboulton!

Issue #2833, opened 2022-12-15
Add an example of using batching with diffusers/transformers


freddyaboulton:
- [ ] I have searched to see if a similar issue already exists.

We should write a guide/create an example demo on using diffusers and or transformers with batching! This came up in this conversation: https://huggingface.slack.com/archives/C047GPHCG6A/p1671134140869209

Should be easy straightforward to just add batching to an existing demo:

https://colab.research.google.com/drive/12_3jGdXj9AaUSGiXKSxL1KgJSR45Z7YK?usp=sharing

**Describe the solution you'd like**  

**Additional context**

PR #3224, opened 2023-02-17
Added diffusers demo with batching & some misc improvements to doc

abidlabs:
Made a few tweaks to the docs, most significantly adding a demo that shows how easy it is to use Gradio's batching mode with diffusers pipelines, which closes: #2833


guides/03_building-with-blocks/02_controlling-layout.md
```diff
@@ -57,8 +57,10 @@ Both Components and Layout elements have a `visible` argument that can set initi
 $code_blocks_form
 $demo_blocks_form
 
+## Variable Number of Outputs
```
abidlabs, line 60: I received 3 separate questions about how to do this today, suggesting that this might not be very visible in the docs or searchable by Google. Therefore separating this out to be its own section which follows logically from the previous section

guides/01_getting-started/02_key-features.md
```diff
@@ -258,6 +258,8 @@ demo.launch()
 
 In the example above, 16 requests could be processed in parallel (for a total inference
 time of 5 seconds), instead of each request being processed separately (for a total
-inference time of 80 seconds).
+inference time of 80 seconds). Many Hugging Face `transformers` and `diffusers` models
+work very naturally with Gradio's batch mode: here's [an example demo using diffusers to
+generate images in batches](https://github.com/gradio-app/gradio/blob/main/demo/diffusers_with_batching/run.py)
```
abidlabs, line 263: Decided to add this as a link instead of a code block as this section would get too long imo

aliabd, line 263: Why aren't we embedding this demo anywhere?

guides/03_building-with-blocks/02_controlling-layout.md
aliabd: I wonder if variable outputs should also be a type of interface in the 4 (i guess 5 now) types of interfaces guide. Something to think about

guides/01_getting-started/02_key-features.md
aliabd: I guess looking at the demo the usage itself doesn't show anything? So up to you

demo/diffusers_with_batching/run.py
```diff
@@ -0,0 +1,16 @@
...
+def generate(prompts):
+  images = generator(list(prompts)).images
+  return [images]
+
+gr.Interface(generate, 
+             "textbox", 
+             "image", 
+             batch=True, 
+             max_batch_size=4  # Set the batch size based on your CPU/GPU memory
+).queue().launch()
```
aliabd, line 16: demo builds  fine on spaces https://huggingface.co/spaces/gradio/diffusers_with_batching but processes an input indefinitely. Is it working correctly?

aliabd (approved):

demo/diffusers_with_batching/run.py
abidlabs: I think it's way too slow without a GPU. Let me test

guides/01_getting-started/02_key-features.md
abidlabs: Mostly because I felt it was taking up too much space already

abidlabs: The demo itself doesn't change, mostly just the code usage

demo/diffusers_with_batching/run.py
abidlabs: Working with the GPU, so we should be good to go 
image

abidlabs:
Thanks for the review @aliabd!

PR #3242, opened 2023-02-19
Fix Gradio 3.19 UI breaks

aliabid94:
Two things broke in gradio 3.19:
Screen Shot 2023-02-19 at 3 36 26 PM

Footer clings to top,
and
Screen Shot 2023-02-19 at 3 36 30 PM
Input fields are themed incorrectly in API docs in dark mode.

Quick fixes and a release for 3.19.1.


abidlabs:
LGTM @aliabid94 can confirm those two issues are fixed. It looks like the response text field is still themed incorrectly? Screenshot: 

image

aliabid94:
oh you're right, lemme fix that too

aliabid94:
fixed

Issue #3243, opened 2023-02-20
Adjust rendered LaTeX size according to font size


abidlabs:
Currently, any LaTeX inside a `gr.Markdown` is rendered as an image, but the size of that image is always the same, regardless of the font size of adjacent font. This becomes especially noticeable when the LaTeX is next to regular size font in a <p> without any heading. See this demo for example:

```py
import gradio as gr

with gr.Blocks() as demo:
    gr.Markdown("""
# Try $2+2=4$
## Try $2+2=4$
### Try $2+2=4$
#### Try $2+2=4$
##### Try $2+2=4$
Try $2+2=4$
    """)
    
demo.launch()
```

<img width="613" alt="image" src="https://user-images.githubusercontent.com/1778297/219985269-33dd847f-2432-41db-a909-843f3faac045.png">

If possible, it would be great if the height of the LaTeX matched the height of the adjacent font.

PR #3258, opened 2023-02-21
LaTeX height fix

abidlabs:
Fixes LaTeX heights so that they are set based on the height of surrounding text instead of being hardcoded by the rendering engine. Fixes: #3243 

Before:

image

After:

image

### DIFF
--- a/ui/packages/markdown/src/Markdown.svelte
+++ b/ui/packages/markdown/src/Markdown.svelte
@@ -34,8 +34,8 @@
 
 	div :global(.math.inline svg) {
 		display: inline;
-		height: 1em;
 		margin-bottom: 0.15em;
+		height: 1em;
 	}
 
 	div {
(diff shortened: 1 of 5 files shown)


abidlabs:
Here's another example: 

image

ui/packages/markdown/src/Markdown.svelte
```diff
@@ -34,6 +34,8 @@
 
 	div :global(.math.inline svg) {
 		display: inline;
+		margin-bottom: 0.15em;
```
aliabid94, line 37: margin-bottom: 0.25em looks better to me, seems like it hangs under the baseline at 0.15em

aliabid94 (approved):
just the margin comment, lgtm otherchange

abidlabs:
Yeah I agree, thanks for the review and suggestion @aliabid94! Here's how it looks with 0.25em;

image

pngwn:
Isn't the paragraph level text basically unreadable now?

I think that will be one of the more common contexts for this syntax too.

abidlabs:
@aliabid94 actually just said the same thing. When the LaTeX is multi-level (fractions, integral), it becomes unreadable. Let me fix this

Issue #3084, opened 2023-01-30
No `interactive` option for `gr.Button()`


space-nuko:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
I want to disable a button, but `gr.Button` has no `interactive` property. I have to hide it instead

**Describe the solution you'd like**  
Add `interactive` to `gr.Button` like there is for `gr.Dropdown`, `gr.Textbox`, `gr.Slider`, `gr.Checkbox`...

**Additional context**  
N/A

PR #3266, opened 2023-02-21
Add `interactive=False` mode to `gr.Button`

abidlabs:
Like other components, `gr.Button` should support a disabled mode if `interactive=False` is set. Note that unlike other components, `gr.Button()`'s interactive state should not be inferred based on whether it is an input component or not, so it needs to be explicitly set to `interactive=False` in order to be disabled. 

The current way of designating the disabled state is to show a "can't-click" cursor and show no color change when a user hovers over the button. Open to feedback on this, though I wanted to keep it pretty simple so as to not need new css variables, etc. 

Fixes: #3084

Recording 2023-02-21 at 10 56 39 (1)

### DIFF
--- a/ui/packages/app/src/components/Button/Button.svelte
+++ b/ui/packages/app/src/components/Button/Button.svelte
@@ -11,6 +11,13 @@
 	export let mode: "static" | "dynamic";
 </script>
 
-<Button {variant} {elem_id} {style} {visible} disabled={mode === "static"} on:click>
+<Button
+	{variant}
+	{elem_id}
+	{style}
+	{visible}
+	disabled={mode === "static"}
+	on:click
+>
 	{$_(value)}
 </Button>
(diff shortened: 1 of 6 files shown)


pngwn (approved):
lgtm!

ui/packages/button/src/Button.svelte
```diff
@@ -7,8 +7,11 @@
 	export let visible: boolean = true;
 	export let variant: "primary" | "secondary" | "stop" | "plain" = "secondary";
 	export let size: "sm" | "lg" = "lg";
+	export let disabled: boolean = false;
 
 	$: ({ styles } = get_styles(style, ["full_width"]));
+	console.log("size", size);
+	console.log("disabled", disabled);
```
pngwn, line 14: remove these,

ui/packages/app/src/components/Button/Button.svelte
```diff
@@ -8,8 +8,16 @@
 	export let visible: boolean = true;
 	export let value: string;
 	export let variant: "primary" | "secondary" = "primary";
+	export let mode: "static" | "dynamic";
```
pngwn, line 11: If we add this, the TypeScript error should go away:

```suggestion
	export let mode: "static" | "dynamic" = "dynamic";
```

abidlabs:
Thanks @pngwn for the review and suggesting the fix!

Issue #2422, opened 2022-10-10
Programmatically run performance tests on queue


freddyaboulton:
As part of debugging why the queue is getting stuck (e.g. #2374 #2399 #2420 ) we've been running a benchmark script that tests how long the queue takes to process 1000 requests in a space that has image, audio, video, and text components.

We should enable devs to run these tests programmatically from github actions. Might be good to get in the habit of running these tests whenever we test a new queue implementation or when we do a new release.

abidlabs:
Thinking about adding this as a non-required test (since it can be long-running)

freddyaboulton:
Agreed about it not being required!

I think reporting the average latency (with some other summary statistics maybe) for the branch compared to main would suffice. And we can decide whether or not it passes.

Something like this maybe:

Component | Main | PR | 
--------------|-------|-----|
Video| 10.2 | 10.3 |
Audio | 9.8 | 9.9|
........|......|....|

PR #3272, opened 2023-02-21
Adding a script to benchmark the queue

abidlabs:
Adds a script to benchmark the performance of the queue and adds some instructions on how to use it. Adapted from @freddyaboulton's script here: https://github.com/gradio-app/gradio/pull/2420

Closes: https://github.com/gradio-app/gradio/pull/2422

### DIFF
--- a/scripts/benchmark_queue.py
+++ b/scripts/benchmark_queue.py
@@ -1,0 +1,116 @@
+'''
+A script that benchmarks the queue performance, can be used to compare the performance 
+of the queue on a given branch vs the main branch. By default, runs 100 jobs in batches
+of 20 and prints the average time per job. The inference time for each job (without the
+network overhead of sending/receiving the data) is 0.5 seconds. Each job sends one of:
+a text, image, audio, or video input and the output is the same as the input.
+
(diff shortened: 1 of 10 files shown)


scripts/benchmark_queue.py
```diff
@@ -0,0 +1,116 @@
...
+
+Navigate to the scripts/ directory and run:
+>> python benchmark_queue.py
+
+You can specify the number of jobs to run and the batch size with the -n parameter:
+>> python benchmark_queue.py -n 1000
+
+The results are printed to the console, but you can specify a path to save the results 
+to with the -o parameter:
+>> python benchmark_queue.py -n 1000 -o results.json
```
freddyaboulton, line 16: nit: I had to run this from the top level of the repo, otherwise it complained about not being able to find the image files.

freddyaboulton (approved):
Thanks @abidlabs ! Looks good to me!

scripts/benchmark_queue.py
abidlabs: Ah thanks, will fix so it's consistent with the instructions

PR #3276, opened 2023-02-21
LaTeX fix attempt #2

abidlabs:
Based on perceptive feedback from @aliabid94 and @pngwn, this fixes LaTeX font size rendering so that the height adjusts doesn't just match the surrounding text, but also adjusts based on the contents of the LaTeX formula itself. This allows multiline math (fractions, integrals) to be visible, particularly in regular paragraph tags. 

**Before (at 100% zoom)**

image

**After (at 100% zoom)**

image


pngwn:
The stacked fractions are still almost impossible to read. The issue is partly the height but also partly the weight of the 'font' at the reduced size.

`demo/latex`:

Screenshot 2023-02-22 at 11 00 39

```
    Start with $\frac{\frac{x+1}{x+2}}{x+3}$ $\frac{x+1}{x+2}$then we get $ 2+x $ and $3$.
```
Another for reference:
Screenshot 2023-02-22 at 11 11 27


View the images at 1200px wide to see them at their correct size.

abidlabs:
Agree with the font weight, but as far as I know, we can't do much about that (without significant effort), since we are using matplotlib's LaTeX rendering to create the svgs.

aliabid94 (approved):
LGTM

pngwn (approved):
lgtm!

abidlabs:
Thanks guys for your reviews!

PR #3288, opened 2023-02-22
Some fixes to disabled buttons

abidlabs:
Was testing setting `interactive=False` for `gr.Button()` when I noticed two issues:

1. Setting `interactive=False` via a `gr.update()` wasn't working

2. It wasn't visually clear that a button was disabled unless you actually hovered over the button

This PR fixes both of those issues.

To test, run this script:

```py
import gradio as gr

with gr.Blocks() as demo:
    a = gr.Button()
    b = gr.Button(interactive=True)
    
    a.click(lambda :gr.Button.update("hi", interactive=False), None, b)

    gr.Button(variant="primary")
    gr.Button(variant="primary", interactive=False)
    
    gr.Button(variant="stop")
    gr.Button(variant="stop", interactive=False)

demo.launch()
```

Which looks like this:

image

### DIFF
--- a/ui/packages/button/src/Button.svelte
+++ b/ui/packages/button/src/Button.svelte
@@ -39,9 +39,9 @@
 	}
 
 	button[disabled] {
-		cursor: not-allowed;
 		opacity: 0.5;
-  		filter: grayscale(30%);
+		filter: grayscale(30%);
+		cursor: not-allowed;
 	}
 
 	.hide {
--- a/ui/packages/button/src/Button.svelte
+++ b/ui/packages/button/src/Button.svelte
@@ -40,6 +40,8 @@
 
 	button[disabled] {
 		cursor: not-allowed;
+		opacity: 0.5; /* lower the opacity to 50% */
+  		filter: grayscale(30%); /* apply 100% desaturation */	
 	}
 
 	.hide {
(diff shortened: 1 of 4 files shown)


ui/packages/button/src/Button.svelte
```diff
@@ -39,6 +39,9 @@
 	}
 
 	button[disabled] {
+		opacity: 0.5; lowed;
+		opacity: 0.5; /* lowe
+		filter: grayscale(30%); ilter: grayscale(30%); /* app
 		cursor: not-allowed;
```
pngwn, line 45: Something we're has happened here. The css auto-ordering vscode extension doesn't play too nicely if there are comments and sometimes garbles the output like this.

pngwn (changes requested):

ui/packages/button/src/Button.svelte
abidlabs: Ah interesting, the comments are unnecessary anyways. Will remove and rerun formatter

abidlabs:
Should be ready for another pass @pngwn. Also fixed the test script above

freddyaboulton (approved):
Looks great @abidlabs ! Really cool

pngwn (approved):
lgtm!

abidlabs:
Thanks for the reviews folks!

Issue #3201, opened 2023-02-15
Add user feedback when flagging is done 


merveenoyan:
I think with loading Interfaces with flagging options allows people to set labelling for active learning loops and is very useful.
It would be nice to give user feedback saying "Your flag is logged!" after user clicks flag by default, as of now, nothing happens and people feel like their inputs aren't logged.
See here: https://huggingface.co/spaces/deprem-ml/ner-active-learning

freddyaboulton:
How/where should the notification be displayed? I'm thinking it looks like the `gr.Error` but it's green and has a friendly message, like this:

![image](https://user-images.githubusercontent.com/41651716/219106243-fd7a4503-51a9-4a78-bd7b-ecb8cd3473b7.png)
 
Maybe this can be a general component that can be leveraged with gr.Blocks as well.

merveenoyan:
@freddyaboulton It could either change color of the button or can be a progress bar on the button or output if it works.

dvsrepo:
Hi, I'm Dani from Argilla, congratulations for such a great tool. This enhancement is highly relevant for us too. I was thinking also about not allowing users to click twice on the same button, but I understand the logic will get more involved, and we could maybe get around by keeping some state inside the ArgillaCallback not to log multiple times the same feedback 

We're going to be using more and more Gradio Apps for collecting human feedback with Argilla. It's a great integration for both public crowd-sourcing efforts like @merveenoyan's amazing project and less open use cases where one can use Gradio+Argilla to iterate on data and models.

Let us know if we can help!

abidlabs:
Hi all, and welcome @dvsrepo! 

What about the Flag Button becoming disabled while the function is running and then becoming enabled again when the flagging is complete? We could make that the default behavior in a `gr.Interface`, as that's the UI I think most users would expect?

dvsrepo:
Sounds great to me! Would the user be notified that the flag info was successfully submitted/saved too, as discussed above?

abidlabs:
I wonder if that is necessary? It feels like a more complex design whereas I think the button becoming disabled and then enabled might be sufficient to provide the user with visual feedback that the flagging is complete

dvsrepo:
thanks @abidlabs ! Yes, I agree that it might be sufficient visual feedback with the button becoming disabled. Thanks so much! Let me know if I can help to test once you plan to include this! 🙌

abidlabs:
[...]

Issue #1764, opened 2022-07-12
Allow changing Flag button text


osanseviero:
- [x] I have searched to see if a similar issue already exists.

Right now, the flagging options only allow to add "Flag as X", but I would like to change the "Flag" name to something entirely different. Is this a use case that could potentially be supported?

freddyaboulton:
Seems reasonable to me @osanseviero ! 

Technically this should already be possible with the blocks api - once #1675 is done, it will be clearer how to do that - but I agree that we should make this easier for the interface api as well.

PR #3289, opened 2023-02-22
Some improvements to Flag

abidlabs:
This PR makes two improvements to the Flag button in `gr.Interface`:

1. Lifts the restriction that flagging options must be in the form ("flag as X"). To do so in a backwards-compatible manner, `flagging_options` can now accept a list of tuples in the form (`label`, `value`), where the label is the text that appears on the button and `value` is the text that is saved into the csv log. (Closes: #1764)

Example:

```py
import gradio as gr

gr.Interface(
    lambda x:x,
    inputs="textbox",
    outputs="textbox",
    flagging_options=[("👍", "positive"), ("👎", "negative")]
).launch()
```

Recording 2023-02-23 at 13 18 28

2. Adds visual feedback that the Flag button has been clicked. (Closes: #3201) 

Recording 2023-02-23 at 12 24 50

Also does some general cleanup related to typing & adds tests. Together, these updates collectively close: #1496 as well.


freddyaboulton:
Really cool way to build off of previous work! 

Yea I I think the switch right now is too fast to notice. What if we made it invisible while the callback is running and replace it with a non-interactive "Flagging in progress.." button ? Would that be more obvious?

I think a closeable pop-up message like `gr.Error` would be the most obvious, though.

dawoodkhan82:
As a user, I would expect some type of confirmation that I successfully flagged. I think a popup would be best.

abidlabs:
I really feel like popups/modals are very intrusive, but maybe that's just me :D. Happy to close this PR if that's the case. Does anyone have any other quick thoughts @aliabid94 @pngwn?

pngwn:
I think we should do a little tick on the button itself that fades after a second or so, i don't think we want a modal that needs dismissing.

abidlabs:
👍 I like @pngwn’s suggestion, let me see if it’s doable with the existing feature set

abidlabs:
Thought a lot about how to implement @pngwn's suggestion from the backend, but I don't think it's possible since we can't trigger two events in a row (make a checkmark appear, and then disappear) from a single button click (without queuing, which is typically set to false for flagging). Ideally, we wouldn't have to write custom js to get this to work. Does anyone have any suggestions for a solution that can be implemented entirely from the backend?

freddyaboulton:
@abidlabs Yea I think what @pngwn is proposing would be tough to do strictly from the backend without css or js.

just building off what you proposed at first, what if we add some artificial delay to the flagging callback replace the `Flag` text with `Saving...` and then switch it back to `Flag` after the saving is done?

I think that makes it clear that work happened and is now finished. If for whatever reason flagging fails, we can raise a gr.Error in that case.

Demo:

replace_button


```python
import gradio as gr
import time

def flagging():
    time.sleep(1)
    return [gr.Button.update(visible=True), gr.Button.update(visible=False)]

with gr.Blocks() as demo:

    with gr.Row():
        flag_btn = gr.Button(value="Flag")
        spinner = gr.Button(value='Saving...', interactive=False, visible=False)

    flag_btn.click(flagging, inputs=None, outputs=[flag_btn, spinner])
    flag_btn.click(lambda: [gr.Button.update(visible=False), gr.Button.update(visible=True)],
                   inputs=None, outputs=[flag_btn, spinner])

demo.launch()
```

abidlabs:
Was thinking exactly along these lines @freddyaboulton!

pngwn:
That's nice, maybe we can add loading state to the button at some point for this kind of usecase.

aliabid94:
I think we should implement it so we always have some button animation when buttons have no outputs. This shouldn't be a flag specific thing imo. I'd imagine it's a common usecase for a button to trigger some job that has no outputs and we should solve that from within the gradio library.

Spinner / check seems like a good approach to me

aliabid94:
My only design issue with spinner and check is that I don't want the text to move from the center. Because for very short requests, it will be jarring to see the text shift right for 0.2s to make space forthe spinner, and then jerk back. Maybe we can reuse the animated white progress background we use for all our outputs

pngwn:
We should just do something like this but with a tick at the end: https://codepen.io/dcode-software/pen/abZPmRb

We can avoid any flashing by just ensuring it does a full run at minimum even if it completes pretty much instantly.

abidlabs:
Adding something to the button itself when there is no output is a cool idea but it feels like too much effort to me given that this is a pretty niche use case outside of flagging. It also raises UX questions (what if showing a tick mark isn't really appropriate for the button based on the action that it's triggering, should it be configurable a parameter?). I think we have a decent solution in what @freddyaboulton proposed above, let's go with that, and we can revisit based on user feedback and real use cases that we see.

merveenoyan:
@abidlabs I think the solution you offered is quite nice it's just can we increase the time duration and not let it flash for a millisecond but a bit longer?

pngwn:
@abidlabs Agree, don't think it needs to happen in this PR, was a suggestion for the future. We can discuss it some more before decided. I think the current proposal is a huge improvement already.

abidlabs:
Sounds good! Yes @merveenoyan will increase the duration for sure

abidlabs:
Ready for review! Tagged on a few related issues to this PR, please see the top comment for an updated description.

gradio/interface.py
```diff
@@ -733,26 +738,43 @@ def attach_interpretation_events(
...
                     flag_components = self.input_components
                 else:
                     flag_components = self.input_components + self.output_components
-                for flag_btn, flag_option in flag_btns:
-                    flag_method = FlagMethod(self.flagging_callback, flag_option)
+
+                for flag_btn, (label, value) in zip(flag_btns, self.flagging_options):
+                    assert isinstance(value, str)
+                    flag_method = FlagMethod(self.flagging_callback, label, value)
+                    flag_btn.click(
```
freddyaboulton, line 769: I don't think this should be queued

gradio/flagging.py
```diff
@@ -575,13 +576,29 @@ def dump_json(self, thing: dict, file_path: str | Path) -> None:
...
     def __call__(self, *flag_data):
-        self.flagging_callback.flag(list(flag_data), flag_option=self.flag_option)
+        try:
+            self.flagging_callback.flag(list(flag_data), flag_option=self.value)
+        except Exception as e:
+            raise gr.Error("Error while flagging: {}".format(e))
+        if not self.visual_feedback:
+            return
+        time.sleep(0.8)  # to provide enough time for the user to observe button change
+        return gr.Button.update(value=self.label, interactive=True)
```
freddyaboulton, line 604: The only problem I see here is that if there is an exception, the button won't change back.

I really like the idea of raising gr.Error here though. So I see two solutions:

1. Be able to `return` a gr.Error instead of just raising it to give visual feedback about the error and return the other updates along side it.
2. Add a  `Press Clear to reset Flag button message to the exception` and modify the Clear button to reset the sates of the flag buttons.

What do you think?

freddyaboulton (commented):
Looking great @abidlabs !  Main question I have is about how best to turn the saving button back to the original state.

gradio/flagging.py
abidlabs: Ah good catch @freddyaboulton. (1) would be cool but I think it would make return values from functions more complicated. I suppose the way to do it would be that you can return a `gr.Error` along with the other return values and it is skipped over when assigning return values to output components. We could consider that but it's definitely outside the scope of this PR. I'll implement (2)

gradio/interface.py
abidlabs: You're right, thanks @freddyaboulton!

gradio/flagging.py
abidlabs: So this is actually  trickier than I thought. The reason is that the `Button` component does not have a `StatusTracker`, so it cannot actually show the `gr.Error` modal unlike the other components. I tried a few different things:

(1) adding a "hidden" output component which could be the target of the `gr.Error` --> however, hidden components don't show the `StatusTracker` either
(2) making Interface's output components the target of the `gr.Error` --> however, this means that the output components show a loading indicator when the flag button is clicked, which is not ideal
(3) adding a `StatusTracker` to the `Button` component --> however, this messed up the layout of the `Button` (doable but would require a bit of frontend work)

In the end, I've just removed the `gr.Error` for now --> instead, if an error is raised, the `Button` just displays the string "Error!", the error is printed to the console / logs, and you can press the clear button to reset flagging. Given that errors during flagging should be pretty rare, I think this should be good enough for now

abidlabs: To test, run a simple Interface like this:

```py
import gradio as gr

gr.Interface(lambda x:x, "textbox", "textbox", flagging_options=["a", "b"]).launch()
```

After the interface launches, delete the "flagged" directory that has been created in the same directory as the Gradio app script. Then, enter something into the input textbox and try flagging it. You'll see the error message appear on the button like this:

image

abidlabs:
Thanks for the review @freddyaboulton! Addressed the issues you raised best I could

gradio/flagging.py
pngwn: > So this is actually trickier than I thought. The reason is that the Button component does not have a StatusTracker, so it cannot actually show the gr.Error modal unlike the other components.

I actually think this is a design error, the global error model should not require a status tracker. Will file an issue.

pngwn: #3347

abidlabs: Thanks @pngwn! Happy to wait if it’s a quick thing to allow any function to raise an error, but otherwise I think we should get this in and potentially update the behavior in the future

freddyaboulton: Thanks for looking into this @abidlabs  and for filing @pngwn !

pngwn: Yeh I would just get this in for now, we can tweak later when we fix that.

gradio/components.py
```diff
@@ -1241,7 +1241,9 @@ def __init__(
             if isinstance(value, str):
                 value = [value]
         if not multiselect and max_choices is not None:
-            warnings.warn("The `max_choices` parameter is ignored when `multiselect` is False.")
+            warnings.warn(
```
freddyaboulton, line 1244: Why is black complaining about this now? 🤔 

Did you update versions?

freddyaboulton (approved):
Thanks for the fixes @abidlabs !

gradio/components.py
abidlabs: Hmm where is it complaining? The GitHub actions are running fine (except for that one flaky test which I need to fix)

freddyaboulton: I'm wondering why it started modifying this line out of a sudden if it was being linted before.

abidlabs: Oh interesting, this change is not part of this PR anymore, so the same change must have happened in another PR that got merged in.

abidlabs: It turns out that I did actually upgrade black versions. Have downgraded down to `black==22.6.0` to match what we have on CI. No complaints from black as far as I can tell, so I think we're good for now.

abidlabs:
Thank you everyone for your feedback and reviews! Will go ahead and merge this in once the CI passes

oostopitre:
Thanks for this feature. It is great to have visual feedback that clicking 'flag' has done something.

ankur-bohra:
For anyone else implementing flagging inside Blocks using FlagMethods, the implementation is mostly in gradio.Interface:
https://github.com/gradio-app/gradio/blob/d9700676be86985773e4235a80af66003d2f681f/gradio/interface.py#L799-L812
FlagMethod's `visual_feedback` argument only returns the button to its original state. The initial state change must be triggered separately.
<details>
<summary>Demo</summary>

https://github.com/gradio-app/gradio/assets/56445395/1851fa15-cfc5-42e7-9c04-4dbd618d9db7

```py
import gradio as gr


def submit(textbox):
    return textbox


with gr.Blocks() as demo:
    with gr.Row():
        with gr.Column():
            textbox = gr.Textbox(label="Input")
            submit_btn = gr.Button("Submit")
        with gr.Column():
            label = gr.Label(label="Output")
            flag_btn = gr.Button("Flag as incorrect")

        submit_btn.click(submit, textbox, label)

        # Add flagging callback
        csv_logger = gr.CSVLogger()
        csv_logger.setup([textbox], "flagged")
        flag_method = gr.flagging.FlagMethod(
            csv_logger,
            "Flag as incorrect",  # The label that the button will return to
            "Incorrect",
        )
        flag_btn.click(
            flag_method,
            inputs=[textbox],
            outputs=[flag_btn],  # Required for the button to return to original state
            preprocess=False,
            queue=False,
        )

        # Change button state when clicked
        flag_btn.click(
            lambda: gr.Button.update(value="Saving...", interactive=False),
            None,
            flag_btn,
        )


demo.launch()

```

</details>

Issue #3259, opened 2023-02-21
Be able to access current username from application


freddyaboulton:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
When using gradio's authentication, it would be awesome if I could know which user is logged in to my app to tailor my app's behavior to that user.

Examples:
- Displaying greeting messages
- Creating an admin panel for a gradio app

**Describe the solution you'd like**  
Maybe this could be exposed in `gr.Request` ?

**Additional context**  

Discord discussion here: https://discord.com/channels/879548962464493619/1077391349617262713/1077391349617262713

abidlabs:
Do the fastapi request headers contain the logged in username? Unless I'm mistaken, they don't have this info @lvelvee

KataiKi:
It doesn't. It keeps the access-token, but not the username used during authentication.

abidlabs:
I agree with @freddyaboulton we can add it as a separate field to the `gr.Request` headers

KataiKi:
If the username is in Headers, would that still be accessible while loading the UI immediately after logging in?  If the username is made to be part of UI, for example, it wouldn't be useful to wait until the user hits Submit before we can read headers.

abidlabs:
Interesting, good point @KataiKi. Can you elaborate on your use case a bit more -- how are you planning on using the username if it's before the submit button?

KataiKi:
The simplest case would just be to have the name displayed somewhere on the UI, like on the top bar somewhere. If I want to load a profile based on the user (profile picture, email, etc), that would be important as well.  If I wanted to display that user's previous settings, I would need to know which user I'm talking to when the user logs in but before they make any submissions.

freddyaboulton:
> If the username is in Headers, would that still be accessible while loading the UI immediately after logging in?

@KataiKi Yes, I think so, you can just pass the request to the `load` event.

```python
import gradio as gr

def create_greeting(request: gr.Request):
    return gr.Markdown.update(value=f"Thanks for connecting from: {request.client}")

with gr.Blocks() as demo:
    user_ip = gr.Markdown(value="Not logged in")

    demo.load(create_greeting, inputs=None, outputs=user_ip)

demo.launch(auth=("admin", "admin"))
```

inewlife:
How to get user info in js function

freddyaboulton:
> It does not use Gradio's authentication so not sure how to do that
[...]

PR #3296, opened 2023-02-23
Access username from `gr.Request` class

abidlabs:
Allows access to the username of a logged-in user from the `gr.Request()` object using the `.username` attribute. Test code:

```py
import gradio as gr

def create_greeting(request: gr.Request):
    return gr.Markdown.update(value=f"Thanks for logging in, {request.username}")

with gr.Blocks() as demo:
    user_ip = gr.Markdown(value="Not logged in")

    demo.load(create_greeting, inputs=None, outputs=user_ip)

demo.launch(auth=("admin", "admin"))
``` 

* Fixes: #3259


freddyaboulton (approved):
Beautiful @abidlabs !

dwipper:
Hi @abidlabs @freddyaboulton. I upgraded my spaces to Gradio 4.0.2. Unfortunately, the login styling is broken again, so it looks like the fix didn't get pushed to the new build.

tpastore-01123:
Hi sorry but I dont understand where is that Request parameter coming from if the inputs is = None :S Im trying to get the username at the breakpoint() of this code:

freddyaboulton:
The request is passed when the `PROMO_ASSISTANT.answer` function is run by the web server, not when the demo is being created.

tpastore-01123:
oh thanks for your pormpt response, so I should be able to access a request parameter as a keyword argument inside PROMO_ASSISTANT.answer ?

dwipper:
@tpastore-01123 There's a pretty good example here of how it works: https://www.gradio.app/docs/request . My understanding is that when I function is called on a Gradio component, gr.Request is available as a incoming param on the function, even though it isn't passed as an input= param.

tpastore-01123:
Nice I see now username from inside the .answer method. Though my use case is that I want to manage different sessions for different usernames having independent ChatInterfaces for each, so I dont know if this is possible since "chatbot" is outside that scope

pjharel:
@tpastore-01123 Not to be considered prod-ready, but this code allows multi-user sessions through a "sessions" dictionary that keeps track of session_ids and usernames. I only keep the 5 most recent session_ids per user, in case one user has multiple tabs and different information need to be stored within each session. The log_history just prints to the console for additional checks.

```python
import gradio as gr
import uuid
import time

class SessionManager:
    def __init__(self):
        self.sessions = {}
        self.current_session_id = None

    def get_current_user(self, req: gr.Request):
        for user, session_list in self.sessions.items():
            if user == req.username:
                return user
        return "No user logged in with this username"

    def get_user(self, req: gr.Request):
        session_id = str(uuid.uuid4())
        timestamp = time.time()
        if req.username not in self.sessions:
            self.sessions[req.username] = [(session_id, timestamp)]
        else:
            self.sessions[req.username].append((session_id, timestamp))
            self.sessions[req.username].sort(key=lambda x: x[1]) # sort by timestamp
            if len(self.sessions[req.username]) > 5:
                self.sessions[req.username].pop(0) # remove the oldest sessions
        current_user = self.get_current_user(req)
        return {user : gr.Markdown(value=f"Thanks for logging in, {req.username}, with session ID {session_id}. Current user: {current_user}")}

manager = SessionManager()

def display_user(req: gr.Request):
    current_user = manager.get_current_user(req)
    log_history(req)
    return {"user_display": current_user}

def log_history(req: gr.Request):
    current_user = manager.get_current_user(req)
    print(current_user)

with gr.Blocks() as demo:
    user = gr.Markdown(value="User not logged in")

    user_display = gr.Textbox(label="Display User")
    display_btn = gr.Button("Display User")

    demo.load(manager.get_user, None, user)
    display_btn.click(fn=display_user, inputs=None, outputs=user_display)

if __name__ == "__main__":
    demo.launch(
        server_name="127.0.0.1",
        show_api=False,
        server_port=8000,
        debug=True,
        auth=[('user1','123'), ('user2','456'), ('user3','789')])
```

tpastore-01123:
Hi again! Related to this...  I've been trying to add

peeter2:
Instead of username does anyone have a clue how to get the email?

Issue #2516, opened 2022-10-24
Add a Python client to make it easy to connect to hosted Gradio apps & respect the queue


freddyaboulton:
- [x] I have searched to see if a similar issue already exists.

As we make the queue the primary way of connecting with gradio apps programmatically, we should offer developers a friendly client that lets them connect to `queue/join` and get the websocket messages.

This will make it easy for developers to treat gradio/spaces as a backend in their apps without worrying about the low-level details of our queue implementation. This will also make it easy for us to change our queue implementation without breaking our user's apps. 

**Describe the solution you'd like**  
Add a python and javascript client for connecting with `queue/join` and getting messages.
The `View Api` page should change to show how to use the client.

FYI @pngwn

abidlabs:
On the Python side, we already have `gr.Interface.load()`, which respects the queue. However, it only supports Spaces right now. Would be good to add support for arbitrary URLs

abidlabs:
Revisiting this -- I think we should develop a separate Python library / client for connecting to hosted Gradio Spaces. The main advantage would be that if someone wants to use a Space as an API, they would not need to have `gradio` (along with its many dependencies) as a dependency, instead they could have `gradio-client`, which would be just a wrapper over network calls, as a dependency. 

Here are the functions that the v1 Python client should support:

```python
import gradio_client as grc

app = grc.load("space/name")  # by default, loads from a Space
app = grc.load("space/name", access_token="...":)  # for private Spaces

app.payload() # provides the expected API format in a programmatic way
print(app)  # prints the expected API format

job = app.predict(arg1, arg2)  # submits the job to the queue
job.status  # PROCESSING or COMPLETED or ERROR

job.result()  # gets the result of the job if completed, otherwise returns `None`
```

Similarly, we can put together a gradio client in JS for JS apps

cc @freddyaboulton @apolinario @osanseviero @yvrjsharma @pngwn for your thoughts

aliabid94:
makes sense to me

freddyaboulton:
Agreed that implementing a client in a separate library is a good idea. I would like it if gradio_client was a dependency of gradio (and that we used it in .load) to make sure it doesn't get stale with the main codebase.

If we're going to make `predict` non-blocking, I think `job` should implement the concurrent.futures interface https://docs.python.org/3/library/concurrent.futures.html#future-objects to make it easy to run them asynchronously with `as_completed`.

[...]

PR #3300, opened 2023-02-24
Sets up the Python `gradio` client

abidlabs:
Creates a lightweight Python library called `gradio_client`

Documentation:

-----------

# `gradio_client`: Use any Gradio app as an API -- in 3 lines of Python

## Installation

The lightweight `gradio_client` package can be installed from `pip` (or `pip3`) and works with Python versions 3.7 or higher:

```bash
$ pip install gradio_client
```

### Connecting to a Space or a Gradio app

Start by connecting instantiating a `Client` object and connecting it to a Gradio app 
that is running on Spaces (or anywhere else)!

**Connecting to a Space**

```python
import gradio_client as grc

client = grc.Client(space="abidlabs/en2fr")
```

**Connecting a general Gradio app**

If your app is running somewhere else, provide the full URL instead to the `src` argument. Here's an example of making predictions to a Gradio app that is running on a share URL:

```python
import gradio_client as grc

client = grc.Client(src="btd372-js72hd.gradio.app")
```

### Making a prediction

The simplest way to make a prediction is simply to call the `.predict()` function with the appropriate arguments and then immediately calling `.result()`, like this:

```python
import gradio_client as grc

client = grc.Client(space="abidlabs/en2fr")

client.predict("Hello").result()

>> Bonjour
```

**Running jobs asyncronously**

Oe should note that `.result()` is a *blocking* operation as it waits for the operation to complete before returning the prediction. 

In many cases, you may be better off letting the job run asynchronously and waiting to call `.result()` when you need the results of the prediction. For example:

```python
import gradio_client as grc

client = grc.Client(space="abidlabs/en2fr")

job = client.predict("Hello")

# Do something else

job.result()

>> Bonjour
```

**Adding callbacks**

Alternatively, one can add callbacks to perform actions after the job has completed running, like this:

```python
import gradio_client as grc

def print_result(x):
    print(x"The translated result is: {x}")

client = grc.Client(space="abidlabs/en2fr")

job = client.predict("Hello", result_callbacks=[print_result])
```

Closes: #2516


pngwn:
I've created #3310 and renamed #2516 so we can close them separately.

freddyaboulton:
Very nice PR @abidlabs ! 🔥 

> By high-level, I mean referring to the organization of files and the usage syntax.

### Regarding organization of files
I think the addition of `client` directory with sub-directories for each language makes sense. Nothing much to say here from my end!

### Regarding usage
I like the way clients are instantiated and how the `predict` method is used. I think that's easy to use and intuitive. And the API introspection is really cool! Questions/Notes I had:

1. How does `predict` behave in the case of generator functions?
2. On a similar note, I think we should allow callbacks for each of the queue messages, not just when the job is complete. Some use cases I can see for this: providing ETA updates from the upstream queue, stitching together the intermediate outputs of a diffusion model into a gif. 
3. I see the api introspection only documents named routes. I think that's reasonable but I have a sense that most routes are actually unnamed. I wonder if we can still provide something for the un-named routes? We have all the input component info in the config even if the route is unnamed so it should still be possible to document them.

 Will think about this some more over the weekend!

abidlabs:
Thanks for the feedback @freddyaboulton! 

> 1. How does predict behave in the case of generator functions?

Still need to handle this case. I'm planning on doing something similar to @pngwn's js client, where predict() gets you the first output of a generator

> 2. On a similar note, I think we should allow callbacks for each of the queue messages, not just when the job is complete. Some use cases I can see for this: providing ETA updates from the upstream queue, stitching together the intermediate outputs of a diffusion model into a gif.

Sounds good. If it's straightforward enough, I'll add to this PR, otherwise will save for a future PR

> 3. I see the api introspection only documents named routes. I think that's reasonable but I have a sense that most routes are actually unnamed. I wonder if we can still provide something for the un-named routes? We have all the input component info in the config even if the route is unnamed so it should still be possible to document them.

I was thinking exactly the same thing. For Blocks, routes are unnamed by default so many of the popular Spaces don't show any named routes right now. I don't have a good solution for this because, although we can certainly provide the information for unnamed routes, this would include *every* single event listener in the app which could be a lot. Suggestions welcome! (cc @aliabid94)

client/python/gradio_client/client.py
```diff
@@ -0,0 +1,320 @@
...
+        self,
+        space: str | None = None,
+        src: str | None = None,
+        access_token: str | None = None,
+        max_workers: int = 40,
+    ):
+        self.access_token = access_token
+        self.headers = (
+            {"Authorization": "Bearer {access_token}"} if access_token else {}
+        )
```
Wauplin, line 28: Would be a could use case for `huggingface_hub` here:

```py
from huggingface_hub.utils import build_hf_headers

headers = build_hf_headers(token=access_token)
# or even
headers = build_hf_headers(token=access_token, library_name="gradio_client", library_version"0.xx")
```
Especially helpful if the user has used `huggingface-cli login` or `huggingface_hub.login` (+it adds a user-agent if you are interested in telemetry)

client/python/gradio_client/client.py
```diff
@@ -0,0 +1,320 @@
...
+
+        if space is None and src is None:
+            raise ValueError("Either `space` or `src` must be provided")
+        elif space and src:
+            raise ValueError("Only one of `space` or `src` must be provided")
+        self.space = space
```
Wauplin, line 34: It seems `self.space` is never used except in `__init__.py` to compute the src url.
I would rather not set it and pass an arg to `_space_name_to_src`.

client/python/gradio_client/client.py
```diff
@@ -0,0 +1,320 @@
...
+            (
+                i
+                for i, d in enumerate(self.config["dependencies"])
+                if d.get("api_name") == api_name
+            ),
+            None,
+        )
+        if inferred_fn_index is None:
+            raise ValueError(f"Cannot find a function with api_name: {api_name}")
+        return inferred_fn_index
```
Wauplin, line 116: (nit) it feels more readable to me to condense it like this:

```suggestion
        for i, d in enumerate(self.config["dependencies"]):
            if d.get("api_name") == api_name:
                return i
        raise ValueError(f"Cannot find a function with api_name: {api_name}")
```

client/python/gradio_client/utils.py
```diff
@@ -0,0 +1,245 @@
...
+# Data processing utils
+########################
+
+
+def download_tmp_copy_of_file(
+    url_path: str, access_token: str | None = None, dir: str | None = None
+) -> tempfile._TemporaryFileWrapper:
+    if dir is not None:
+        os.makedirs(dir, exist_ok=True)
+    headers = {"Authorization": "Bearer " + access_token} if access_token else {}
```
Wauplin, line 76: Could use `huggingface_hub.build_hf_headers` here as well

client/python/test/test_utils.py
```diff
@@ -0,0 +1,71 @@
...
+import tempfile
+
+import pytest
+from gradio import media_data
+
+from gradio_client import utils
+
+data = {
+    
+}
```
Wauplin, line 11: unused?
```suggestion
```

Wauplin (commented):
Nice python client @abidlabs! I started to read it and ended up with a few very low-level comments -you can discard them if you want-. My main feedback is that I think we can use `huggingface_hub.build_hf_headers` to handle authentication. It would be more Hub-focused rather than generic to Gradio but it would ease things for users that are already used to login via the cli/in a notebook.

abidlabs:
Thanks @Wauplin! Appreciate the suggestions, will utilize `huggingface_hub` more

abidlabs:
Question for you: is there a way to replace: 

```py
    def _space_name_to_src(self) -> str | None:
        return (
            requests.get(
                f"https://huggingface.co/api/spaces/{self.space}/host",
                headers=self.headers,
            )
            .json()
            .get("host")
        )
```

with a function from `huggingface_hub`? I think when we discussed it last, there wasn't a way but I wanted to double check

Wauplin:
> Question for you Wauplin: is there a way to replace `_space_name_to_src` with a function from huggingface_hub?

This is very new from the Hub side but yes it is possible:

```py
from huggingface_hub import space_info

def _space_name_to_src(self) -> str | None:
    info = space_info("microsoft/visual_chatgpt")
    return f"https://{info.subdomain}.hf.space"
```

It seems that `host` is not returned but `subdomain` is. I am not sure API-side is you are guaranteed to have a subdomain so you might need to to `subdomain = getattr(info, "subdomain", None); if subdomain is not None: ...`.

(cc @coyotte508 @julien-c since I know returning "host" has been discussed _at some point_ but couldn't retrieve the convo)

coyotte508:
You are guaranteed to get a `subdomain`.

Regarding host, it'd be nice to have I agree

julien-c:
FWIW there's also a `GET /api/:repoType(spaces)/:namespace/:repo/host` endpoint which is already used from the gradio codebase

client/python/gradio_client/client.py
abidlabs: Done

abidlabs:
Hi everyone! This PR is ready for review. It's not quite ready for launch yet (and I think we should hold off from launching until after theming), so I've removed it from the changelog for now. But the core pieces are in place and I did want to merge it in before it becomes stale, since a lot of files are affected. Specifically, this PR:

* Creates a `client/python/gradio_client` directory for the new `gradio_client` library
* Adds various scripts / tests / packaging-related files so that we can easily test the `gradio_client` and release it to pypi as needed
* Moves various pieces of the `gradio` library to `gradio_client`, most importantly `serializing.py`
* Of course, creates the core `gradio_client.Client` class itself. 
* Ensures all old tests continue to pass while adding a few new ones

Here are the things that are still left to do (I'll create a tracking issue afterwards):

* add support for iterators -- right now iterators throw return a meaningless update as a result
* add status callbacks, not just result callbacks
* add ability to inspect and quick get an idea of how to use each of the API endpoints for a given gradio app
* add this PR back to the changelog 

Please let me know if you find any other gaps or bugs as you test it out. Instructions to test are at the top of the PR.

abidlabs:
From our sync today, adding a few more TODOs:

- Update this PR to resolve issues related to temp files
- Document gr.load only
- rename acces_token to hf_token
- clarify that callbacks are executed sequentially
- Files should be uploaded via /upload
- Session management / reset()
- Analytics 
- HuggingFace Hub as a dependency

Some of these will be tackled in future PRs

abidlabs:
Made some changes to this PR:

* Reconciled with `main` 
* Documented `gr.load()`. I included the documentation for `gr.Blocks.load()` and `gr.Interface.load()` but marked them as methods that will be deprecated:

image


image

* Added telemetry using the `huggingface_hub` library

client/python/README.md
```diff
@@ -0,0 +1,105 @@
+# `gradio_client`: Use any Gradio app as an API -- in 3 lines of Python
+
+This directory contains the source code for `gradio_client`, a lightweight Python library that makes it very easy to use any Gradio app as an API. This library is CURRENTLY IN ALPHA. APIS MAY CHANGE. WE DO NOT RECOMMEND USING IT EXCEPT IN EXPERIMENTAL PROJECTS.
```
aliabid94, line 3: can we remove this comment? I think it discourages users unnecessarily - if this is trusted enough to be used for the core gradio library, then its good enough for our users, they'll open issues if things break

client/python/README.md
```diff
@@ -0,0 +1,105 @@
...
+Here's the entire code to do it:
+
+```python
+import gradio_client as grc
+
+client = grc.Client(space="stability-ai/stable-diffusion")
+job = client.predict("a hyperrealistic portrait of a cat wearing cyberpunk armor")
+job.result()
+
+>> URL
```
aliabid94, line 16: unclear `>> URL` was supposed to be the output from `job.result()`

client/python/gradio_client/client.py
```diff
@@ -0,0 +1,300 @@
...
+        self.executor = concurrent.futures.ThreadPoolExecutor(max_workers=max_workers)
+
+        # Disable telemetry by setting the env variable HF_HUB_DISABLE_TELEMETRY=1
+        threading.Thread(target=self._telemetry_thread).start()
+
+    def predict(
+        self,
+        *args,
+        api_name: str | None = None,
+        fn_index: int = 0,
```
aliabid94, line 66: allow result_callbacks to be a singleton, and wrap as list if so?

client/python/gradio_client/client.py
```diff
@@ -0,0 +1,300 @@
...
+                        callback(future.result())
+
+                return fn
+
+            for callback in result_callbacks:
+                job.add_done_callback(create_fn(callback))
+
+        return job
+
+    ##################################
```
aliabid94, line 92: dont like this big comment block, private methods are self evident via underscore prefix

aliabid94 (approved):
lgtm!

client/python/README.md
abidlabs: I'll soften the message, but the library is not meant for external release. We can always change the names of variables, etc. for internal use but if people build on top of it, then it may cause their code to break.

client/python/README.md
abidlabs: Ok I'll just remove this

client/python/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+```python
+import gradio_client as grc
+
+def print_result(x):
+    print(x"The translated result is: {x}")
+
+client = grc.Client(space="abidlabs/en2fr")
+
+job = client.predict("Hello", callbacks=[print_result])
```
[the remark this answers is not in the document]
abidlabs: Ok I prefer `callbacks` too but we are also planning on adding `status_callbacks`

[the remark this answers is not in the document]
abidlabs: We can change this later if needed (since library is in alpha)

client/python/gradio_client/client.py
abidlabs: Sure sg

client/python/gradio_client/client.py
abidlabs: Sg

client/python/gradio_client/client.py
```diff
@@ -0,0 +1,298 @@
...
+    def predict(self, *data) -> Tuple:
+        data = json.dumps({"data": data, "fn_index": self.fn_index})
+        hash_data = json.dumps(
+            {"fn_index": self.fn_index, "session_hash": str(uuid.uuid4())}
+        )
+        if self.use_ws:
+            result = utils.synchronize_async(self._ws_fn, data, hash_data)
+            output = result["data"]
+        else:
+            response = requests.post(self.api_url, headers=self.headers, data=data)
```
freddyaboulton, line 184: Right now there is no way to use a space with auth via api right? I feel like that is something we should support? Curious on your thoughts

freddyaboulton: Maybe we should set is_valid to false for endpoints in auth enabled apps?

client/python/gradio_client/client.py
```diff
@@ -0,0 +1,298 @@
...
+                output = result["data"]
+            except KeyError:
+                if "error" in result and "429" in result["error"]:
+                    raise utils.TooManyRequestsError(
+                        "Too many requests to the Hugging Face API"
+                    )
+                raise KeyError(
+                    f"Could not find 'data' key in response. Response received: {result}"
+                )
+        return tuple(output)
```
freddyaboulton, line 196: Curious how come we're returning tuples everywhere given that "data" is already a list?

gradio/external.py
```diff
@@ -33,6 +31,45 @@
     from gradio.interface import Interface
 
 
+set_documentation_group("helpers")
```
freddyaboulton, line 34: nit: I think the way `set_documentation_droup` works, it doesn't really allow calling it twice because it resets the classes to document for that group every time it's called. 

Just tested locally and this is not showing up. 

Let's fix before next release!

scripts/format_backend.sh
```diff
@@ -3,6 +3,6 @@
 cd "$(dirname ${0})/.."
 
 echo "Formatting the backend... Our style follows the Black code style."
-python -m black gradio test
-python -m isort --profile=black gradio test
-python -m flake8 --ignore=E731,E501,E722,W503,E126,E203,F403 gradio test --exclude gradio/__init__.py
+python -m black gradio test client/python/gradio_client
+python -m isort --profile=black gradio test client/python/gradio_client
+python -m flake8 --ignore=E731,E501,E722,W503,E126,E203,F403 gradio test client/python/gradio_client --exclude gradio/__init__.py client/python/gradio_client/__init__.py
```
freddyaboulton, line 8: I think we need an `--exclude` for each file?

This script isn't working for me

image

test/test_blocks.py
```diff
@@ -126,6 +126,7 @@ def fake_func():
             demo.load(fake_func, [], [textbox])
 
         config = demo.get_config_file()
+        print("config\n", config)
```
freddyaboulton, line 129: It's in a test file but let's still get rid of it

freddyaboulton (commented):
Super awesome PR @abidlabs !

Left some comments that can be handled in future PRs

abidlabs:
Thanks for the review @freddyaboulton and @aliabid94! Will address these comments

client/python/scripts/ci.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+cd "$(dirname ${0})/.."
+
+echo "Linting..."
+python -m black --check test gradio_client
+python -m isort --profile=black --check-only test gradio_client
+python -m  flake8 --ignore=E731,E501,E722,W503,E126,E203,F403,F541 test gradio_client --exclude gradio_client/__init__.py
```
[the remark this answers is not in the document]
abidlabs: Fixed thanks

client/python/gradio_client/client.py
```diff
@@ -0,0 +1,298 @@
...
+            if d.get("api_name") == api_name:
+                return i
+        raise ValueError(f"Cannot find a function with api_name: {api_name}")
+
+    def __del__(self):
+        if hasattr(self, "executor"):
+            self.executor.shutdown(wait=True)
+
+    def _space_name_to_src(self, space) -> str | None:
+        return (
```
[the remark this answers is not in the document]
abidlabs: Done

client/python/gradio_client/client.py
abidlabs: Let me add it as a list of things to support in the tracking issue

client/python/gradio_client/client.py
abidlabs: It helped with the typing I believe

abidlabs:
Ok I think that should be everything! Will merge in after CI passes

gradio/external.py
freddyaboulton: Not the first time that has tripped me up 🤦 Disregard!

scripts/format_backend.sh
abidlabs: Fixed. Turns out you can use comma (no space) to separate multiple filenames

Issue #3309, opened 2023-02-24
Change, Stream, Change (again) section in Image docs.


pngwn:
### Describe the bug

The docs for`Image` has the `change` method listed twice.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

https://gradio.app/docs/#image-change-header

`change` is listed here, scroll past `stream` and yuou will see change again.

### Logs

```shell
-
```

### System Info

```shell
-
```

### Severity

annoying

PR #3318, opened 2023-02-24
Fix change event listed twice in image docs

aliabd:
Very simple fix to bug in image docs.

Closes: #3309


freddyaboulton (approved):
Thanks @aliabd !

Issue #3326, opened 2023-02-27
A status page for `gradio.live`


abidlabs:
We should set up a status page for gradio.live and link to it from the gradio library if a share link is not successfully created. That way the community can know that it's a technical issue on our end that we're working on (as opposed to a usage issue or some other issue on their end).

PR #3331, opened 2023-02-27
Gradio Status Page 

aliabd:
Created a status page to monitor a few different things: https://status.gradio.app

* Sharing (https://gradio.live) 
* API (https://api.gradio.app/)
* Website (https://gradio.app) 

Will probably change some things around in the page so please give feedback :) 

Also added it to the `COULD_NOT_GET_SHARE_LINK` string message. Which I discovered doesn't actually get printed when there's no internet because we were only catching RuntimeErrors but [this line](https://github.com/gradio-app/gradio/blob/53e4733f2a0cbb9a696a28330dc3927c7997c165/gradio/networking.py#L160) causes a ConnectionError

Closes: #3326

### DIFF
--- a/gradio/strings.py
+++ b/gradio/strings.py
@@ -10,7 +10,7 @@ en = {
     "RUNNING_LOCALLY": "Running on local URL:  {}",
     "RUNNING_LOCALLY_SEPARATED": "Running on local URL:  {}://{}:{}",
     "SHARE_LINK_DISPLAY": "Running on public URL: {}",
-    "COULD_NOT_GET_SHARE_LINK": "\nCould not create share link, please check your internet connection.",
+    "COULD_NOT_GET_SHARE_LINK": "\nCould not create share link. Please check your internet connection or our status page: https://status.gradio.app",
     "COLAB_NO_LOCAL": "Cannot display local interface on google colab, public link created.",
     "PUBLIC_SHARE_TRUE": "\nTo create a public link, set `share=True` in `launch()`.",
     "MODEL_PUBLICLY_AVAILABLE_URL": "Model available publicly at: {} (may take up to a minute for link to be usable)",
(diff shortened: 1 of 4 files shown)


abidlabs:
Beautiful @aliabd! As commented internally, it would be great if we could somehow combine API & Share links since for end users if either go down, share links go down and it would be a little clearer. Otherwise, LGTM.

abidlabs (approved):

abidlabs:
Several Spaces went down, causing the flaky tests to fail. Restarting them now

gradio/strings.py
```diff
@@ -10,7 +10,7 @@
     "RUNNING_LOCALLY": "Running on local URL:  {}",
     "RUNNING_LOCALLY_SEPARATED": "Running on local URL:  {}://{}:{}",
     "SHARE_LINK_DISPLAY": "Running on public URL: {}",
-    "COULD_NOT_GET_SHARE_LINK": "\nCould not create share link, please check your internet connection.",
+    "COULD_NOT_GET_SHARE_LINK": "\nCould not create share link. Please check your internet connection or our status page: https://status.gradio.app",
```
freddyaboulton, line 13: I wonder if we should link to this somewhere on the website, too?

freddyaboulton (approved):
Super cool @aliabd ! Super fast turnaround 🏃‍♂️

gradio/strings.py
abidlabs: Good point, we could add a link in the footer next to the logo: 

image

abidlabs: image

aliabd: Added it at the bottom 

Screen Shot 2023-02-27 at 9 24 23 AM

abidlabs: Hmm not sure but maybe it should be separated a bit more? #notagraphicdesigner but
image

abidlabs: Or maybe even having it aligned on the right next to twitter / github makes would look nicer color-theme-wise

abidlabs: LGTM!

PR #3338, opened 2023-02-27
fix

### LINKED ISSUE #3319: 3.16.0 introduced breaking change for Dropdown
### Describe the bug

The addition of the multiselect option to `Dropdown` introduced a breaking change for the Dropdown components output signature. The default case when there is no value (for example, an optional dropdown that a user does not interact with before submitting) is now an array (list on the python side) instead of an empty string. This means checks like `dropdown == ""` etc will now be False rather than True.

I think it is fine if users select an option and _then_ select nothing.

This is the offending line: https://github.com/gradio-app/gradio/pull/2871/files#diff-f0b03055e354ad5f6f0c4c006adba9a9fe60a07ca1790b75e20cd0b69737ede8R11

### Reproduction

Above.

### Screenshot

-

### Logs

```shell
-
```

### System Info

```shell
-
```

### Severity

pngwn:
# Description

closes #3319

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

Closes: # (issue)

# Checklist:

# A note about the CHANGELOG

### DIFF
--- a/ui/packages/app/src/components/Dropdown/Dropdown.svelte
+++ b/ui/packages/app/src/components/Dropdown/Dropdown.svelte
@@ -9,7 +9,7 @@
 	export let info: string | undefined = undefined;
 	export let elem_id: string = "";
 	export let visible: boolean = true;
-	export let value: string | Array<string> = [];
+	export let value: string | Array<string>;
 	export let multiselect: boolean = false;
 	export let max_choices: number;
 	export let choices: Array<string>;
@@ -18,6 +18,12 @@
 	export let loading_status: LoadingStatus;
 
 	export let mode: "static" | "dynamic";
+
+	if (multiselect && !value) {
+		value = [];
+	} else {
+		value = "";
+	}
 </script>
 
 <Block


ui/packages/app/src/components/Dropdown/Dropdown.svelte
```diff
@@ -18,6 +18,12 @@
 	export let loading_status: LoadingStatus;
 
 	export let mode: "static" | "dynamic";
+
+	if (multiselect && !value) {
+		value = [];
+	} else {
```
abidlabs, line 24: Should this be
```suggestion
	} else if (!value) {
```
to avoid overwriting a provided value to dropdown?

pngwn: Yeah defo. Its broken in more ways than just that atm tho, just making some changes.

pngwn:
I'm pretty sure the multiselect is broken on main, i don't think it will allow you select anything if you don't pass a `max_choice` option which should be optional. Fixed that anyway.

This should work now.

When `Multiselect=True` the initial empty value should be an empty list, when `Multiselect=False` it should be a string. This should also be true when clearing the value with the `clear` button in `Interface`.

gradio/components.py
```diff
@@ -1257,7 +1259,10 @@ def __init__(
             value=value,
             **kwargs,
         )
-        self.cleared_value = self.value
+
+        self.cleared_value = (
+            self.value if self.value is not None else [] if multiselect is True else ""
```
abidlabs, line 1264: Might be helpful for readability
```suggestion
        self.cleared_value = (
            self.value if self.value is not None else ([] if multiselect is True else "")
```

gradio/components.py
```diff
@@ -1257,7 +1259,10 @@ def __init__(
...
+        )
```
abidlabs, line 1265:
```suggestion
        self.cleared_value = self.value or ([] if multiselect else "")
```

abidlabs:
As I was testing this, I noticed one issue, which is that when the "x" button is clicked on the dropdown, the value of the dropdown is set to `[]` even if `multiselect=False`. Code to reproduce:

```py
import gradio as gr

with gr.Blocks() as demo:
    a = gr.Dropdown(multiselect=True, choices=["abc", "def"])
    b = gr.Textbox()
    
    demo.load(lambda x:x, a, b)
    a.change(lambda x:x, a, b)
    
demo.launch()
```

You'll notice that when the page loads, the textbox is initially set to "", which is correct. But if you click on the "x" button on the dropdown, the value in the textbox will change to [], which is wrong.

ui/packages/app/src/components/Dropdown/Dropdown.svelte
```diff
@@ -9,7 +9,7 @@
 	export let info: string | undefined = undefined;
 	export let elem_id: string = "";
 	export let visible: boolean = true;
-	export let value: string | Array<string> = [];
+	export let value: string | Array<string>;
```
freddyaboulton, line 12: Shouldn't this also include 'null'?

freddyaboulton (approved):
Thanks for the fix @pngwn !

abidlabs:
LGTM @pngwn! Works exactly as expected. 

Ignore the flaky test for now, I'll fix that on a separate PR

abidlabs (approved):

PR #3343, opened 2023-02-27
Apply height in Gallery.style

### LINKED ISSUE #3320: Gallery.style does not take effect
### Describe the bug

I was trying to increase the height of the gallery with `Gallery.style(height="800px")`.
I found confirmation someone else couldn't do it too: https://github.com/gradio-app/gradio/issues/654#issuecomment-1346641312

I looked through the CSS and have some questions.

Here's my process of trying to make it work:

1. First I found the top level gallery component and added the height 800px property. It added a chunk of black below when i expected the images to be filled there.
image

2. Went 1 level down the DOM into first child with `class="wrap default svelte-h2bzqy hide"`. Mouseover shows it's height got changed to 800px so don't think this is the problem

3. 2nd child is for the Label on top left, feels irrelevant
4. 3rd child with `class="grid-wrap svelte-ywf5u7 fixed-height"` mouseover shows the original small box, something is wrong here. I deleted fixed-height class from the element (`    --size-80: 320px;`) and got the desired longer gallery. I could also freely edit `height` on this node with effects seen. However without JS i couldn't delete `fixed-height` using `with gr.Blocks(css=custom_css`), so I tried to overwrite the `fixed-height` properties with specificity/precedence
```
.fixed-height.svelte-ywf5u7.svelte-ywf5u7 {
    min-height: var(--size-80);
    max-height: 55vh;
}
```
image

5. So i added `height:800px` to 3rd child with `class="grid-wrap svelte-ywf5u7 fixed-height"` and no visual changes happened. Then i tried `min-height:800px` just to match what fixed-height was using and it worked to overwrite fix-height effects while leaving it there. 

6. To keep things clean i deleted `height 800px` from the top level gallery component and it still works

7. I fixed my issue eventually using  (didn't add `!important` unlike answer linked in top of this issue)
```
div#gallery_search > div:nth-child(3) {
                                     min-height: 1000px;
                            }
```

Questions

1. Why is gallery.style(height:"800px") not responding? Is it an issue with `fixed-height` class?
2. Within 3rd child, why does height:800px not work but min-height:800px does? I thought these 2 properties are not mutually exclusive and min-height doesn't overwrite height?
3. Why does removing height:800px from top level gallery component still work?
4. Is there any way to hack my way through this using `height: 800px !important` in any of the nodes so i can avoid considering difference between height vs min-height? 
5. What's the purpose of the 1st child `class="wrap default svelte-h2bzqy hide"`?
6. How do users know what styles can be provided using the any_component.style() api?

### Reproduction

```
import gradio as gr


file = "https://stpaulpet.com/wp-content/uploads/dog-facts-cat-facts.jpg"

with gr.Blocks() as demo:
    gallery = gr.Gallery([file]*50).style(grid=[4], height="1000px")

if __name__ == "__main__":
    demo.launch()
```

### Screenshot

image

### Logs

```shell
NA
```

### System Info

```shell
NA
```

### Severity

annoying

freddyaboulton:
# Description

Fixes #3320

Example on `gallery_component` demo

### 700 px
image

### 300 px

image

### Default

image

# Checklist:

# A note about the CHANGELOG

### DIFF
--- a/ui/packages/gallery/src/Gallery.svelte
+++ b/ui/packages/gallery/src/Gallery.svelte
@@ -17,9 +17,6 @@
 	export let style: Styles = { grid: [2], height: "auto" };
 	let styles: string;
 
-	$: console.log(style);
-	$: console.log(style.height !== "auto");
-
 	$: _value =
 		value === null
 			? null
@@ -98,8 +95,6 @@
 	}
 
 	$: styles = add_height_to_styles(style);
-
-	$: console.log(styles);
 
 	let height = 0;
 	let window_height = 0;
--- a/ui/packages/gallery/src/Gallery.svelte
+++ b/ui/packages/gallery/src/Gallery.svelte
@@ -87,7 +87,7 @@
 
 	$: can_zoom = window_height >= height;
 
-	function add_height_to_styles(style: Styles) {
+	function add_height_to_styles(style: Styles): string {
 		styles = get_styles(style, ["grid"]).styles;
 
 		return styles + ` height: ${style.height}`;
--- a/ui/packages/gallery/src/Gallery.svelte
+++ b/ui/packages/gallery/src/Gallery.svelte
@@ -15,6 +15,10 @@
 	export let root_url: null | string = null;
 	export let value: Array<string> | Array<FileData> | null = null;
 	export let style: Styles = { grid: [2], height: "auto" };
+	let styles: string;
+
+	$: console.log(style);
+	$: console.log(style.height !== "auto");
 
 	$: _value =
 		value === null
@@ -87,7 +91,15 @@
 
 	$: can_zoom = window_height >= height;
 
-	$: ({ styles } = get_styles(style, ["grid"]));
+	function add_height_to_styles(style: Styles) {
+		styles = get_styles(style, ["grid"]).styles;
+
+		return styles + ` height: ${style.height}`;
+	}
+
+	$: styles = add_height_to_styles(style);
+
+	$: console.log(styles);
 
 	let height = 0;
 	let window_height = 0;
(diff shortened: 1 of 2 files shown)


ui/packages/gallery/src/Gallery.svelte
```diff
@@ -151,7 +163,7 @@
 	<div
 		bind:clientHeight={height}
 		class="grid-wrap"
-		class:fixed-height={style.height !== "auto"}
+		class:fixed-height={!style.height || style.height == "auto"}
```
freddyaboulton, line 166: Before, fixed-height would always be set if the user specified a height value in `style`. This was restricting the size of the div containing the gallery.

Now it's only set if `height` is not set, or if it's auto.

ui/packages/gallery/src/Gallery.svelte
```diff
@@ -87,7 +91,15 @@
 
 	$: can_zoom = window_height >= height;
 
-	$: ({ styles } = get_styles(style, ["grid"]));
+	function add_height_to_styles(style: Styles) {
+		styles = get_styles(style, ["grid"]).styles;
```
freddyaboulton, line 95: Adding `height` to the array passed to `get_styles` only adds height to the css if height=auto which is not what we want in this case.

ui/packages/gallery/src/Gallery.svelte
```diff
@@ -15,6 +15,7 @@
 	export let root_url: null | string = null;
 	export let value: Array<string> | Array<FileData> | null = null;
 	export let style: Styles = { grid: [2], height: "auto" };
+	let styles: string;
```
pngwn, line 18: We can remove this i think, we don't need to define the variable upfront when using `$: x = y`. The type should be inferred.

pngwn (approved):
Looks good! Thanks!

freddyaboulton:
Thank you for the review @pngwn !

PR #3349, opened 2023-03-01
Google Analytics fix

abidlabs:
It turns out that in the front end, `analytics_enabled` was always hardcoded to be `False`, which meant that Google analytics were never called. This fixes that so that we have more reliable usage analytics.

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -585,7 +585,7 @@ class Blocks(BlockContext):
 
         derived_fields = ["types"]
 
-        with Blocks(theme=config["theme"], css=config["theme"]) as blocks:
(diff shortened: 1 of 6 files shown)


gradio/blocks.py
```diff
@@ -585,7 +585,7 @@ def iterate_over_children(children_list):
 
         derived_fields = ["types"]
 
-        with Blocks(theme=config["theme"], css=config["theme"]) as blocks:
```
abidlabs, line 588: Unrelated, but noticed that this issue as well

freddyaboulton (approved):
Thanks @abidlabs !

abidlabs:

Issue #2520, opened 2022-10-25
Buttons do not render in html 


ValentinKoch:
### Describe the bug

Buttons dont render in HTML. Are clickable but do not look as intended, also when providing a dedicated css.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```
with gr.Blocks() as demo:
    subject = gr.HTML(value='<button>Click Me!</button>')
demo.launch()
```

### Screenshot

<img width="66" alt="Bildschirmfoto 2022-10-25 um 16 35 11" src="https://user-images.githubusercontent.com/55893735/197803160-6c55815f-842d-423b-89c0-a28cf72f2972.png">

### Logs

```shell
no logs
```

### System Info

```shell
gradio 3.4, Chrome
```

### Severity

annoying

freddyaboulton:
I can't reproduce so I will close this out! Feel free to re-open if I'm misunderstanding @ValentinKoch !

```python
import gradio as gr

with gr.Blocks() as demo:
    subject = gr.HTML(value="""<button type="button" onclick="alert('Hello world!')">Click Me!</button>""")
demo.launch()
```

![button_click](https://user-images.githubusercontent.com/41651716/208181688-ff6bc1a8-9229-46f0-99b7-fd1865375b16.gif)

ValentinKoch:
Hi @freddyaboulton , the Button is clickable but the design is not rendering. A Button looks usually like this: https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_button_test
so I'd kindly ask you to reopen the issue. It makes for a weird design if buttons are just text that you can click on.
Best
Valentin

freddyaboulton:
Reopened! Thanks @ValentinKoch

davidbernat:
@freddyaboulton: this may be out of scope, but may I ask these two questions as this ticket is nearest in scope?
1. Is it possible to shut off all default style within a `gr.HTML()` so that it is, effectively, an `iframe`.
2. Your demo here use `onclick()` to launch `alert` but could you demonstrate a correct design pattern for calling a linked snippet of JavaScript which may or may not be embedded in the element, and how you would embed it in the element?

pngwn:
This is happening because we 'reset' styles for easier styling later. I don't of a great solution to this unless we move all of our resets inline. We can fix the button but there will be other similar issues with different elements.

ValentinKoch:
[...]

PR #3352, opened 2023-03-01
Do not `inherit` `text-decoration` on `<a>`

### LINKED ISSUE #837: HTML Output Hyperlink Styling
HTML outputs should underscore and style like hyperlinks.

image

We can style the hyperlinks in the description as well:
image

Hyperlink example:
image

sambostock:
# Description

It seems reasonable that by default links should be decorated (underlined), and that undecorated links are the special case, not the other way around.

Closes #837 (tentatively

# Checklist:

- ~I have commented my code in hard-to-understand areas~

---

cc. @pngwn & @aliabid94, who originally added this in d6fbc32ed1fa9c55b291e901d0ba1a68c458bf8e


pngwn:
I'll try to get to this before the end of the week. I don't have any issue with the code but this could break the UI anywhere we are using an anchor tag for anything, so will need thorough testing.

pngwn:
As an aside, are sure #837 is still present? The HTML component gets the `prose` class added (https://github.com/gradio-app/gradio/blob/main/ui/packages/html/src/HTML.svelte#L14) and `.prose a` has text decoration (https://github.com/gradio-app/gradio/blob/main/ui/packages/theme/src/typography.css#L170-L175)

abidlabs:
Tested, you're right @pngwn links show up just fine:

```py
import gradio as gr

with gr.Blocks() as demo:
    gr.HTML(
"""
Visit <a href="www.google.com">www.google.com</a> and click <button>here</button>
""")
    
demo.launch()
```

Looks like we can close this PR and the underlying issue (#837). Also tested a related issue (#2520), which I can confirm is still open on the latest version of gradio.

pngwn:
Awesome.

Issue #2899, opened 2022-12-28
Output file is put at /tmp file and it doesn't work on cluster environment


MrNocTV:
### Describe the bug

I'm using version 3.14, and there is an issue with `/tmp` file.
After `/join` processing , Gradio returns the path to the output file.

<img width="549" alt="Screen Shot 2022-12-28 at 13 58 25" src="https://user-images.githubusercontent.com/11014873/209771615-2f77219a-1023-43cd-b7da-5998f664220e.png">

Then it tries to fetch that file, but unfortunately it failed.
In a cluster environment like multiple pods on K8s, requests are routed to different pod every single times.

In this case, the request was routed to a different pod, and it logs out this error.

```
ValueError: File cannot be fetched: /tmp/output_vsre9e5bfed6ae3cdf251047d5d5ddb1bc5e6d4b1d0.mp4. All files must contained within the Gradio python app working directory
```

I think the only way to fix this problem is to return the output file in websocket message. 

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Run Gradio 3.14 on kubernetes with at least 2 pods

### Logs

```shell
2022-12-28T05:25:55.012+0000 raw_response = await run_endpoint_function(                                                                                                                              
2022-12-28T05:25:55.013+0000 File "/usr/local/lib/python3.9/site-packages/fastapi/routing.py", line 163, in run_endpoint_function                                                                     
2022-12-28T05:25:55.015+0000 return await run_in_threadpool(dependant.call, **values)                                                                                                                 
2022-12-28T05:25:55.016+0000 File "/usr/local/lib/python3.9/site-packages/starlette/concurrency.py", line 41, in run_in_threadpool                                                                    
2022-12-28T05:25:55.020+0000 return await anyio.to_thread.run_sync(func, *args)                                                                                                                       
2022-12-28T05:25:55.021+0000 File "/usr/local/lib/python3.9/site-packages/anyio/to_thread.py", line 31, in run_sync                                                                                   
2022-12-28T05:25:55.023+0000 return await get_asynclib().run_sync_in_worker_thread(                                                                                                                   
2022-12-28T05:25:55.025+0000 File "/usr/local/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 937, in run_sync_in_worker_thread                                                        
2022-12-28T05:25:55.027+0000 return await future                                                                                                                                                      
[...]

PR #3355, opened 2023-03-02
allow set file_directories via environment variable and force GA valu…

MrNocTV:
- Allow owner of the environment (where Gradio is deployed) to override and force user to turn of Google Analytics. 
- Allow owner of the environment (where Gradio is deployed) to override and force user to use value of `file_directories`

# Description
I believe that it would be beneficial to have the feature that I'm proposing. This feature would allow users to have greater flexibility when working with the code, as it wouldn't completely restrict their actions. In certain environments, such as within an enterprise, sending user data to Google Analytics may be prohibited due to privacy or security concerns. Additionally, the infrastructure owner may want to specify a directory from which users can freely access files without any security concerns. This is entirely up to the infrastructure team, and in my opinion, it should be implemented as an optional feature to give users more control and flexibility.

Closes: # (https://github.com/gradio-app/gradio/issues/2899)

Please help to review the PR if we agree upon the features above.

# Checklist:

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -482,11 +482,10 @@ class Blocks(BlockContext):
 
         # For analytics_enabled and allow_flagging: (1) first check for
         # parameter, (2) check for env variable, (3) default to True/"manual"
-        self.analytics_enabled = (
-            analytics_enabled
-            if analytics_enabled is not None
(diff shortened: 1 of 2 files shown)


MrNocTV:
@abidlabs 
Thank you for your response, the only solution left for us is to clone Gradio and continue using our custom version.

PR #3370, opened 2023-03-03
Updated chat ui

aliabid94:
Wanted to update the chat ui before the new theming stuff comes in so we can release it asap.

Screen Shot 2023-03-02 at 9 59 22 PM
Screen Shot 2023-03-02 at 10 00 05 PM


dawoodkhan82:
Nice work. I definitely think this is more readable. Although i don't know how I feel about orange text.

dawoodkhan82:
Chat bubble with images needs fixing.

Screenshot 2023-03-03 at 1 36 28 AM

dawoodkhan82:
Also the chatbot doesn't seem to be  responsive?
Screenshot 2023-03-03 at 1 41 18 AM

ui/packages/app/src/components/Chatbot/Chatbot.svelte
```diff
@@ -30,7 +30,6 @@
 		/>
 	{/if}
 	<ChatBot
-		{style}
```
dawoodkhan82, line 33: we shouldn't get rid of this. I actually have a pr using style so devs can set the height of the chatbot (https://github.com/gradio-app/gradio/pull/3369)

abidlabs:
Thoughts on also updating the docs to use this format @aliabid94:

```py
def respond(chat_history, message):
  response = ...
  return chat_history + [[message, response]]

with gr.Blocks():
  chatbot = gr.Chatbot()
  msg = gr.Textbox()
  clear = gr.Button("Clear")

  msg.submit(respond, [chatbot, msg], chatbot)
  clear.click(lambda: None, None, chatbot)
```

abidlabs:
Speaking of the Guide, I would replace: 

```py
txt.submit(lambda :"", None, txt)
```
with
```py
txt.submit(None, None, txt, _js="() => {''}")  
```

as clearing it with javascript means that the textbox will clear instantaneously instead of the function joining the queue

freddyaboulton:
@abidlabs If I new nothing about gradio and wanted to get a chatbot going, I would prefer seeing 

```python
txt.submit(lambda :"", None, txt, queue=False)
```

Than

```python
txt.submit(None, None, txt, _js="() => {''}")  
```

I think the `_js` parameter would be confusing to newcomers (leading underscore in param name + most(?) may not know javascript syntax). 

Just my two cents. Don't want to hijack the PR 😂 Proposed styling looks good to me pending @dawoodkhan82 's comments @aliabid94 ! Thanks 🔥

abidlabs:
That makes sense, but perhaps we can clarify what's happening with a comment. The problem is that using the existing logic, if you post your chatbot to Spaces, you get the following UI (the progress bar appears in the textbox, which is not pretty):

image

aliabid94:
Agree with @freddyaboulton we shouldn't be encouraging js based clear, we need to add support for a native clear button.

abidlabs:
Actually just setting queue=False on the clear event should do the trick @aliabid94

aliabid94:
> Actually just setting queue=False on the clear event should do the trick @aliabid94
True!

demo/chatbot_multimodal/run.py
```diff
@@ -1,27 +1,27 @@
...
     
     with gr.Row():
         with gr.Column(scale=0.85):
             txt = gr.Textbox(show_label=False, placeholder="Enter text and press enter, or upload an image").style(container=False)
         with gr.Column(scale=0.15, min_width=0):
             btn = gr.UploadButton("🖼️", file_types=["image"])
             
-    txt.submit(add_text, [state, txt], [state, chatbot])
+    txt.submit(add_text, [chatbot, txt], [chatbot])
     txt.submit(lambda :"", None, txt)
```
abidlabs, line 23:
```suggestion
    txt.submit(lambda :"", None, txt, queue=False)
```

aliabd:
I'm not sure about the orange text either. I think it would look better without it (with just the orange background and border) 

Also I think it should be the other way around: the output message should be highlighted with the orange, not the input message. My first impression from the screenshot was that the model was predicting on the right.

abidlabs:
Found a few small things:

1. Links don't appear visibly as links:

```py
import gradio as gr

with gr.Blocks() as demo:
    c = gr.Chatbot([("visit [https://www.google.com](https://www.google.com) for more info", "def \n sadf")])
    
demo.queue().launch()
```

image

2. If an image is included in the chatbot (like several multimodal models do), there should be a little more padding above and below the image:

```py
import gradio as gr

with gr.Blocks() as demo:
    c = gr.Chatbot([("visit ![](file=lion.jpg) for more info", "def \n sadf")])
    
demo.queue().launch()
```

3. Newline (\n) should be respected -- the ChatGPT API returns \n characters, and currently, they don't appear at all in our chatbot, leading to a degraded experience. 

```py
import gradio as gr

with gr.Blocks() as demo:
    c = gr.Chatbot([("a \n b \n c", "def \n sadf")])
    
demo.queue().launch()
```

image

Otherwise, LGTM! Nice upgrade

demo/chatbot_simple_demo/run.py
```diff
@@ -0,0 +1,17 @@
...
+  response = random.choice(["Yes", "No"])
+  return chat_history + [[message, response]]
+
+with gr.Blocks() as demo:
+    chatbot = gr.Chatbot()
+    msg = gr.Textbox()
+    clear = gr.Button("Clear")
+
+    msg.submit(respond, [chatbot, msg], chatbot)
+    clear.click(lambda: None, None, chatbot)
```
abidlabs, line 14:
```suggestion
    clear.click(lambda: None, None, chatbot, queue=False)
```

aliabid94:
fixed

aliabd:
Just noticed something, the chatbot doesn't auto scroll down on new input. Makes it feel like nothing happened. 


https://user-images.githubusercontent.com/9021060/222864467-27cd4a4c-172f-4e2c-b152-3b95c15081af.mov

aliabd:
Actually, I tested this locally and scrolls as expected. I think it's only an issue in the pr-deploys space

abidlabs (approved):

abidlabs:
Hi @bent-verbiage thanks for raising this point. We're working on a completely new way to provide styling information, which will be released soon, but apologies for this breaking change. In the meantime, you should be able to achieve the same effect with a little bit of custom css like this:


```py
import gradio as gr

css = """
.message.user{
border-color: green !important;
background: #5eb27d !important;
}
.message.bot{
border-color: red !important;
background: #b25e73 !important;
}
"""

    gr.Chatbot([("abc def", "ghi jkl"), ("abc def", "ghi jkl")])
    
demo.launch()
```

image

abidlabs:
Yes I believe this should continue working (though the implementation is still WIP so can’t fully guarantee it).

PR #3388, opened 2023-03-06
JS client take 2

pngwn:
# Description

Closes https://github.com/gradio-app/gradio/issues/3310. Closes https://github.com/gradio-app/gradio/issues/3262. Closes https://github.com/gradio-app/gradio/issues/3263. 

New API:

**Initialise the client**

```js
const app = await client('some/space', space_status_callback);
```

Supports the following types for tyhe 'some/space' argument:
- `owner/name`
- `sub-dom-ain.hf.space`
- `https://sub-dom-ain.hf.space`
- `https://my-own-host.com`

**Make a prediction**

```js
const result = await app.predict('/predict', payload);
```

This will return the result for simple endpoints of the first generated value for generator endpoints, regardless of whether there are a finite number of iterations or require cancellation.

Returning the first result in this manner means that generator functions will never block for longer than a simple prediction function, even if they are infinite.

**Get data updates other than the initial result**

For generator functions, users may be interested in data other than the initial value.

```js
app.on('data', event => {
  event.type /// === 'data'
  event.endpoint // === '/predict' or whichever endpoint the data corresponds to
  event.data // the payload/ data
  event.fn_index // === the fn_index
})
```

**Get status updates**

Users may be interested in what their prediction is up to:

```js
app.on('status', event => {
  event.type /// === 'status'
  event.endpoint // === '/predict' or whichever endpoint the data corresponds to
  event.status // the status for this prediction? need to double check
  event.fn_index // === the fn_index
})
```

To cancel a generator function:

```js
app.cancel('/predict', optional_fn_index);
```

This does nothing if the function is not a generator.


julien-c (commented):
i haven't dived enough at all, but what would you think of this living (or minimally, being duplicated/exposed) in https://github.com/huggingface/huggingface.js, maybe as a `@huggingface/spaces` package ?

Would be quite consistent with Inference APi and Endpoints.

cc @coyotte508 @Wauplin and others

pngwn:
Reposting:

The client will work with any app regardless of how it is hosted, so i don't think it makes sense to require users to install a hugging face specific lib if they're only using the library and not hugging face. 

I also think it makes sense to keep them separate so we can change things more easily with less red tape, gradio functions have some features that require specific APIs. 

We also use the client library in gradio itself, so we need to be able to tweak + develop locally, it is core to gradio so having it somewhere else doesn't make sense.

The hub library could include it as a dependency and wrap it but i wouldn't want to do it the other way around.

coyotte508:
We can link it in a "see also" section in the hugginface.js's main README for example?

pngwn:
This library isn't being published yet anyway so there is no huge rush on this, we'd like to dogfood it a little before releasing to npm.

pngwn:
Will probably release a 0.x at some point just to get the publish tooling sorted + tested. We'll be releasing other stuff under the gradio namespace too soonish™️.

abidlabs:
I did one round of testing and the demos look good @pngwn. Can you go ahead and resolve the merge conflict and I'll do some more testing?

pngwn:
Should be okay now @abidlabs.

pngwn:
The frontend of the chatbot replaces the link sources for a reasons that aren't entirely clear to me. 

This PR changes how we handle basepaths a little.

Will fix.

abidlabs (approved):
Tested a variety of demos and everything is working great! LGTM

Issue #3362, opened 2023-03-02
Remove encryption-related functionality from the library


abidlabs:
We currently have legacy code related to encryption in the `gradio` library, i.e. see:

https://github.com/gradio-app/gradio/blob/5c01a029cef77157c28c224946afe45fda421104/gradio/blocks.py#L1276

as well as `encryptor.py`. This code has several problems:

* It makes various other parts of the codebase more complex (see `serializing.lpy` and `processing_utils.py`) 
* It adds a problematic dependency to our codebase: `pycryptodome` (see #2447)
* It doesn't work! Somehow it must have broken in the migration from 2.x to 3.x. This can easily tested by setting `encrypt=True` in `launch()` and then flagging an image-to-image Interface. If `encrypt=True`, the images should be encrypted before being saved to disk but they are not.

Given this, I think we should just remove all of this related code from the codebase.

Issue #2447, opened 2022-10-12
Can't run gradio on M1


gabrielhuang:
### Describe the bug

I can't import gradio. Installed Python 3.9.12 using Miniconda on Mac M1.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```bash
(base) ➜  ~ python
Python 3.9.12 (main, Apr  5 2022, 01:52:34)
[Clang 12.0.0 ] :: Anaconda, Inc. on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import gradio
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/gabriel.huang1/miniconda3/lib/python3.9/site-packages/gradio/__init__.py", line 3, in <module>
    import gradio.components as components
  File "/Users/gabriel.huang1/miniconda3/lib/python3.9/site-packages/gradio/components.py", line 39, in <module>
    from gradio import media_data, processing_utils, utils
  File "/Users/gabriel.huang1/miniconda3/lib/python3.9/site-packages/gradio/processing_utils.py", line 17, in <module>
    from gradio import encryptor, utils
  File "/Users/gabriel.huang1/miniconda3/lib/python3.9/site-packages/gradio/encryptor.py", line 2, in <module>
    from Crypto.Cipher import AES
  File "/Users/gabriel.huang1/miniconda3/lib/python3.9/site-packages/Crypto/Cipher/__init__.py", line 27, in <module>
    from Crypto.Cipher._mode_ecb import _create_ecb_cipher
  File "/Users/gabriel.huang1/miniconda3/lib/python3.9/site-packages/Crypto/Cipher/_mode_ecb.py", line 35, in <module>
    raw_ecb_lib = load_pycryptodome_raw_lib("Crypto.Cipher._raw_ecb", """
  File "/Users/gabriel.huang1/miniconda3/lib/python3.9/site-packages/Crypto/Util/_raw_api.py", line 309, in load_pycryptodome_raw_lib
    raise OSError("Cannot load native module '%s': %s" % (name, ", ".join(attempts)))
[...]

PR #3396, opened 2023-03-06
Remove everything related to encryption

abidlabs:
As I was refactoring some of the code related to serialization, I tested the encryption functionality that's built into gradio (setting `encrypt=True` in `launch()` is supposed to encrypt the flagged data on disk) and discovered that it was broken: **it had no effect on the flagged data.** 

It's easy enough to verify this:

```py
import gradio as gr

gr.Interface(lambda x,y:(x,y), ["image", "textbox"], ["image", "textbox"]).launch(encrypt=True)
```

Run this, and go to the flagged directory --> the images and text will be saved in an unencrypted format, viewable to anyone.

Going back, it looks like `encrypt=True` broke when we migrated to 3.0.
* [A Colab showing that encryption works on 2.9.4](https://colab.research.google.com/drive/13yhu4XIlZWTBZf6m6EmUjS0vltBfY9ap#scrollTo=uasakojv3EyT)
* [Colab showing that encryption does *not* works on 3.0](https://colab.research.google.com/drive/10JTaxWf5yK84eblCKnGtRrHR0ww846t7#scrollTo=uasakojv3EyT)

Given that no one has complained about this bug for ~1 year, it's safe to say that we can remove this. This has the benefit, besides simplifying the codebase in various places, of also removing a dependency, `pycryptodome`, which has [caused issues for users](https://github.com/gradio-app/gradio/issues/2447) installing Gradio in the past.

Closes: #3362


aliabid94 (approved):
LGTM!

abidlabs:
Thanks for the speedy review!

PR #3399, opened 2023-03-06
Adding the .select event listener

aliabid94:
Adds the `.select()` event listener to various components. This event has `EventData`, data that is passed along the API to provide additional information about the event trigger. Specifically,

* `Gallery`: The `EventData` for the Gallery component is the index of the gallery item that was clicked.
* `Chatbot`: users can select a message
* `CheckboxGroup`: refers to user selected option
* `Dataframe`: users can select a specific cell
* `Dropdown`: refers to user selected option
* `File`: users can select a file from a list of files
* `Gallery`: users can select an image from a gallery
* `HighlightedText`: users can selected a highlighted span
* `Label`: users can select a category
* `Radio`: refers to user selected option
* `TabItem`: refers to own selected tab name
* `Tab`: refers to user selected tab name
* `Textbox`: Users can select a subset of text

Fixes: #1976


abidlabs:
A suggestion on naming: how about `select()` instead of `focus()`? For two reasons:

* For js users, I'm pretty sure that `focus()` is the opposite of `blur()` and so `.focus()` would naturally mean an event that gets fired when a component (like a `Textbox`) is focused on, the opposite of our `.blur()` 
* In my mind, `.select()` is more intuitive for this particular action because you are "selecting" one of the images. Similarly, we might "select" one of the Chatbot messages or select one of the choices in a Dropdown or select one of named entities in a `HighlightedText`

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 
 
 ## New Features:
-No changes to highlight.
+- Added the `.focus()` event listener, applicable right now only to the Gallery component. This event has EventData, data that is passed along the API to provide additional information about the event trigger. The EventData for the Gallery component is the index of the gallery item that was clicked. By [@aliabid94](https://github.com/aliabid94) in [PR 3399](https://github.com/gradio-app/gradio/pull/3399)
```
abidlabs, line 5: Would be great to include a code snippet showing usage in the changelog since this is a pretty big new feature

abidlabs:
Very cool @aliabid94! Tested and works well. Should get the frontend reviewed as well though

My little demo:

```py
import gradio as gr
import numpy as np

def display(images, event_data: gr.EventData):
    return images[event_data.data["index"]].name

with gr.Blocks() as demo:
    with gr.Row():
        f = gr.Files(file_types=["image"], file_count="multiple")
        g = gr.Gallery().style(preview=True)
        i = gr.Image()
    with gr.Row():
        s = gr.Button("Start")
        
    s.click(lambda files: [i.name for i in files], f, g)
    g.focus(display, f, i)


if __name__ == "__main__":
    demo.launch()
```

abidlabs (approved):

aliabid94:
> A suggestion on naming: how about select() instead of focus()? For two reasons:

The reason I didn't want to choose select is because it seems to be very applicable to Radio, Dropdown, etc., and they already have a `change` method. We could add `select` to all of those too I suppose.

> I opened PR https://github.com/gradio-app/gradio/pull/3403 to this PR (lol) to show how we can add type hints to the Gallery Event data. I think that will be better for developer experience.

Amazing, thanks @freddyaboulton!

abidlabs:
I think we should actually add a `.select()` to those components @aliabid94. There's a difference between `change()` and `.select()`, which is that `.change()` is triggered even when the value of the Radio, etc. is changed programmatically whereas `select()` shouldn't. 

See this issue: https://github.com/gradio-app/gradio/issues/1431

pngwn:
@aliabid94 Needs rebasing on main before i can review.

pngwn:
I'm also happy to take care of it if you have other stuff to get on with.

pngwn:
Is the event data the last parameter to the predict function if it is available, or do you have to pull it off with some magic?

aliabid94:
Have revamped and expanded this PR a lot - now most components have a `.select` method that carries Event data signifying the user selecting some part of the Component.

For each event trigger, we carry a value and an index (see gr.SelectEvent)

Chatbot: users can select a message
CheckboxGroup: refers to user selected option
Dataframe: users can select a specific cell
Dropdown: refers to user selected option
users can select a file from a list of files
Gallery: users can select an image from a gallery
HighlightedText: users can selected a highlighted span
Label: users can select a category
Radio: refers to user selected option
TabItem: refers to own selected tab name
Tab: refers to user selected tab name
Textbox: Users can select a subset of text

See specific demos of select at gallery_select and tictactoe, and test all components in blocks_kitchen_sink

abidlabs:
Wow this is a beast of a PR! I've updated the parent comment to reflect its scope of what's happening. 

Gonna start taking a look at it now -- this PR certainly unlocks a lot of new possibilities but let's make sure to test this thoroughly.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - New code component allows you to enter, edit and display code with full syntax highlighting by [@pngwn](https://github.com/pngwn) in [PR 3421](https://github.com/gradio-app/gradio/pull/3421)
 
 ![](https://user-images.githubusercontent.com/12937446/224116643-5cfb94b3-93ce-43ee-bb7b-c25c3b66e0a1.png)
+- Added the `.select()` event listener to the Gallery component. This event listener has EventData, data that is passed along the API to provide additional information about the event trigger. The EventData for the select event listener is the index of the item that was clicked. By [@aliabid94](https://github.com/aliabid94) in [PR 3399](https://github.com/gradio-app/gradio/pull/3399)
```
abidlabs, line 9: Needs to be updated (perhaps replaced with the contents of the parent comment in this PR) & let's also include a code example

test/test_events.py
```diff
@@ -17,6 +18,26 @@ def fn_img_cleared():
 
         assert demo.config["dependencies"][0]["trigger"] == "clear"
 
+    def test_event_data(self):
```
abidlabs, line 21: I would test the other Selectable components too, maybe not in this way (since launching Interfaces is slow) but at least make sure the events are added to the config successfully

gradio/events.py
```diff
@@ -264,3 +265,34 @@ def __init__(self):
...
+            self, "select", callback=lambda: setattr(self, "selectable", True)
+        )
+        """
+        This event is triggered when the user selects from within the Component.
+        This event has EventData of type gradio.SelectEvent that carries information, accessible through SelectEvent.index and SelectEvent.value.
+        See EventData documentation on how to use this event data.
+        """
+
+
+class SelectEvent(EventData):
```
abidlabs, line 284: I would find it more intuitive to call this class `SelectData` instead of `SelectEvent`, as it just holds the data related to the selection

abidlabs:
Notes as I've been testing the various components:

* `Gallery` ⤵️

works great. However, it seems like we broke the `preview` parameter of `.style()` at some point, which means that I can't start a Gallery in preview mode, which is probably what I want to do if there's a `.select()` event trigger. 

* `Chatbot`, `Dataframe`, `Textbox` ✅

works great, able to get the index as a list of indices, enables very cool functionalities

* `CheckboxGroup`, `Dropdown`, `Radio` ⤵️

works, but there is a small issue. When the frontend figures out whether a component should be interactive or not, it doesn't take into account whether the component has a `.select()` trigger. Let me give a concrete example:

```py
import gradio as gr

def get_index(select_data: gr.SelectEvent):
    return select_data.index

with gr.Blocks() as demo:
    text = gr.Textbox()
    dropdown = gr.Dropdown(["a", "b"])
    textbox = gr.Textbox()
    text.submit(lambda x: x, text, dropdown)
    dropdown.select(get_index, None, textbox)
    
demo.launch()
```

Here', the dropdown should be interactive, but it's not. I wonder if we should fix the logic in the frontend to account for this? The complication is that for other components, like `Textbox`, it's actually okay that it's not interactive since `.select()` applies even to static Textboxes. Perhaps it's fine to leave it as is and just ask users to set `interactive=True`. Thoughts?

* `File`, `HighlightedText`, `Label` ✅

Works great, able to get index as a scalar! Blurring the line between input and output components!

* `Tab` ❌

Not working for me. Also I'm confused why you make a distinction between `Tab` and `TabItem` -- I thought they were aliases. Here's the code I tried:

```py
import gradio as gr

def get_index(select_data: gr.SelectEvent):
    return select_data.index

with gr.Blocks() as demo:
    with gr.Tab("abc") as tab1:
        gr.Image()
    with gr.Tab("def") as tab2:
        text = gr.Textbox()
    
    tab2.select(get_index, None, textbox)
    
demo.launch()
```

**Overall comments**

This is a really nice PR and will enable a lot of new use cases with Gradio (as the tic-tac-toe demo attests). The main problem I see is that the details of the `.select()` method for each component are different and they are not documented anywhere. If I go to the website, all I see is:

image

image

This is fine for the other event listeners because they are very simple and don't carry any payloads. But since `select()` does include a payload, which differs between different components, we definitely need to document this somewhere. I'm thinking ideally in the docs for the `.select()` component of each component  -- would that be possible?

aliabid94:
> Here, the dropdown should be interactive, but it's not. I wonder if we should fix the logic in the frontend to account for this? 

I don't think component triggers should define interactivity, e.g. imagine adding event data to Click, and clicking on part of a static image. I also don't think dropdown, checkbox, and radio should ever be non-interactive - no one uses them for outputs.

gradio/events.py
aliabid94: agreed.

test/test_events.py
aliabid94: it's not a very useful to add more tests without the frontend, as the backend does nothing different per component.

aliabid94:
> Tab: Not working for me. Also I'm confused why you make a distinction between Tab and TabItem -- I thought they were aliases. Here's the code I tried:

Ok fixed. Initially I had only attached the `.select` to the parent tabs (see blocks_kitchen_sink), but you can now attach it directly to a child Tab as well

> I'm thinking ideally in the docs for the .select() component of each component

Done.

gradio/helpers.py
```diff
@@ -797,3 +806,24 @@ def get_color_gradient(c1, c2, n):
...
+
+    Example:
+        gallery = gr.Gallery()
+        focused = gr.Textbox()
+
+        def on_focus(evt: gr.SelectData):  # SelectData is a subclass of EventData
+            return evt.index
+
+        gallery.select(on_focus, None, focused)
+    """
```
abidlabs, line 825: This example used the old "focus" terminology: 
```suggestion
    Example:
        gallery = gr.Gallery()
        index = gr.Textbox()

        def on_select(evt: gr.SelectData):  # SelectData is a subclass of EventData
            return evt.index

        gallery.select(on_select, None, index)
    """
```

gradio/helpers.py
```diff
@@ -797,3 +806,24 @@ def get_color_gradient(c1, c2, n):
...
+
```
abidlabs, line 826: Can we add the 2 demos here so that users have more examples?

gradio/helpers.py
```diff
@@ -797,3 +806,24 @@ def get_color_gradient(c1, c2, n):
...
+    def __init__(self, target: Block | None, data: Any):
+        self.target = target
```
freddyaboulton, line 829: Is target used anywhere yet? We should add an example and explain what it can be used for!

guides/03_building-with-blocks/01_blocks-and-event-listeners.md
```diff
@@ -145,3 +145,14 @@ Here is an example of a sine curve that updates every second!
 
 $code_sine_curve
 $demo_sine_curve
+
+## Gathering Event Data
```
freddyaboulton, line 149: 🎉

gradio/helpers.py
```diff
@@ -797,3 +806,24 @@ def get_color_gradient(c1, c2, n):
 
     subprocess.call(ffmpeg_cmd, shell=True)
     return output_mp4.name
+
+
+@document()
+class EventData:
```
abidlabs, line 812: This part of the docs is still confusing for me: 

image

It says that the attributes of `EventData` are `target` and `data` but the example shows the subclass `gr.SelectData` having attributes `index`.

abidlabs:
LGTM @aliabid94! Overall docs look great, just left a few nits above for your consideration

abidlabs (approved):

gradio/helpers.py
freddyaboulton: We should also document the target and data parameters!

aliabid94: well EventData only has target, but the subclass carries extra attributes. Will try to clarify but feel free to edit.

abidlabs: Can we document the subclasses too then? Nowhere do we document that `SelectData` has `index` for example

gradio/helpers.py
aliabid94: added to example code snippet

aliabid94:
> We should also document the target and data parameters

documented target, actually hid data for now because I want users to use the subclasses attributes directly. Data is just an ugly json they shouldn't worry about.

abidlabs:
LGTM @aliabid94!

abidlabs:
I think we're ready to go!! 🚀

Issue #3409, opened 2023-03-07
Test fails locally on Windows/Mac but passes on CI


abidlabs:
### Describe the bug

As reported by @gitgithan (https://github.com/gradio-app/gradio/pull/3405), one of our tests is failing locally:

`test/test_components.py::TestModel3D::test_in_interface`

Running `pytest test/test_components.py::TestModel3D::test_in_interface` on Windows/Mac seems to result in a test failure. However, this test works fine on CI. 

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

pytest test/test_components.py::TestModel3D::test_in_interface

### Screenshot

N/A

### Logs

```shell
E       AssertionError: assert False
E        +  where False = <built-in method endswith of str object at 0x000001FF8FEA50F0>('.gltf')
E        +    where <built-in method endswith of str object at 0x000001FF8FEA50F0> = 'C:\\Users\\islam\\AppData\\Local\\Temp\\Boxabdb9e60cc6dda6a2f81d4764a34e24fd47220509a788219e16779754072681e88277aba8baa417er_j5cdfr.txt'.endswith
```

### System Info

```shell
main
```

### Severity

annoying

PR #3411, opened 2023-03-07
Fix local tests

abidlabs:
Some tests were failing locally, leading to confusion among contributors. This fixes those tests and closes: #3409 

To reproduce, just run `pytest` from the gradio directory on a Mac or Windows and all tests should pass.


gradio/processing_utils.py
```diff
@@ -376,9 +376,12 @@ def get_temp_url_path(self, url: str) -> str:
...
         return prefix + file_hash + extension
 
-    def get_temp_base64_path(self, base64_encoding: str, prefix: str) -> str:
-        extension = get_extension(base64_encoding)
-        extension = "." + extension if extension else ""
+    def get_temp_base64_path(
+        self, base64_encoding: str, prefix: str, extension: str
+    ) -> str:
+        guess_extension = get_extension(base64_encoding)
+        if not extension and guess_extension:
```
abidlabs, line 383: The Model3D test was failing because the extension was not being correctly guessed from the base64. This fixes that. I have no idea why the test was passing on CI

aliabid94 (approved):
lgtm

abidlabs:
Thanks for reviewing @aliabid94!

Issue #943, opened 2022-04-07
Access filename of uploaded image file


shubham-goel-zefr:
When an image file object is uploaded, you cannot retrieve it's original filename. You can only get the temporary filename associated with it. I'm using the flagging feature to log mistakes in my model, and then use the generated csv to further retrain my model. It'd be helpful if instead of Image/0.png, Image/1.png, and so on, I get the ability to use the original filename in the longs.

**Describe the solution you'd like**
1. Modify the flagging option to use the original filename instead of a iterator starting from 0. 
2. Have access to the original filename, when the type of input image is set to 'pil'.

abidlabs:
Thanks @shubham-goel-zefr, this makes sense. Let us look into seeing if we can implement this.

stunlocked1:
can this be added to gradio.components.Audio?

joshwilson-dev:
Is this included in Gradio 4.0 yet? If so, how do I access the filename from the uploaded image object?

marcelloexp386:
I have a similar question. Is it possible to keep the datetime of the original picture shot in the metainformation of the uploaded image? In the case I am working on, I am more interested in knowing the time when an uploaded image was shot than the original name. In any case, if the name of the uploaded image could be preserved would represent a great help because with (Samsung) Android the name of the jpeg file of the picture taken by your smartphone is the timestamp of the shot.

Issue #360, opened 2021-11-12
Input Components File returns a temporary file object with a random ending


DavidZha1994:
After upgrading from gradio-2.3.6 to gradio-2.3.6, random ending of file object is not beginning with "_" anymore,This makes it hard for me to separate it from the original name of the file.

abidlabs:
@aliabid94 it's possible to get the original name in a different way, right?

DavidZha1994:
any updates?

piegu:
Hi, I've still the same question. 

How to get the name of the input file and not the one of the temporary file created by Gradio?

The following code from a pdf file named "example.pdf" gives for example "/tmp/example958q5pem.pdf", not "example.pdf".

```
import gradio as gr

def get_filename(file_obj):
    return file_obj.name

gr.Interface(get_filename, "file", "text").launch()
```

aliabid94:
Hmm sorry for the late response, currently we automatically append the hash. Will reopen this issue to fix this.

DoctorXK:
> Hmm sorry for the late response, currently we automatically append the hash. Will reopen this issue to fix this.

Hi there, 

I think having access to the original filename is a very signficant property which will be frequently used. Hope you can fix this soon :)

abidlabs:
Sorry for the long wait, but this will be fixed by #2073. In Gradio 3.2, you will be able to do `file_obj.orig_name` to get the original uploaded file name.

Seedmanc:
After the breaking change to the Files component involving uploading, the orig_name now returns path instead of just name. How to get the name again?

abidlabs:
Ah thanks for notifying us @Seedmanc, will reopen this issue to get this fixed.

Redcof:
> Sorry for the long wait, but this will be fixed by #2073. In Gradio 3.2, you will be able to do `file_obj.orig_name` to get the original uploaded file name.

Thanks @abidlabs

PR #3412, opened 2023-03-07
Preserve uploaded file names

abidlabs:
Currently, names of files that are uploaded through Gradio components are not preserved. Building off of a nice suggestion by @aliabid94, this PR persists file names of files uploaded through any Gradio component.

The basic change is that instead of files being saved in `/tmp/{orig_filename}_{hash}.{suffix}` path, they are now saved in the `/tmp/{hash}/{orig_filename}.{suffix}` path. This applies to both output files and uploaded files.

Removes the need for `file.orig_name` as `file.name` will return the original filename.

Fixes: #360 
Fixes: #943


freddyaboulton:
@abidlabs I'm having trouble running this locally on mac.

Getting this when running `model3D` demo

```bash
ValueError: File cannot be fetched: /var/folders/t1/j7cmtcgd0mx43jh9nj_r9mmw0000gn/T/78170efd55f69ea61bda4ab184ba0776c78c5b5c/Duck.glb. All files must contained within the Gradio python app working directory, or be a temp file created by the Gradio python app.
```

And the model3d files do not show up

image

Can someone else on mac make sure I'm not going crazy?

freddyaboulton:
If I remember correctly, last time this happened is because on Mac, `/var/` is `/private/var`, I think we had to call `.resolve()` on the path?

abidlabs:
Thanks for testing, let me fix that!

abidlabs:
Can you test it now @freddyaboulton? I think it should work

aliabid94:
filenames preserved in mac

aliabid94 (approved):
LGTM

abidlabs:
Thanks @aliabid94 and @freddyaboulton for reviewing!

Issue #3339, opened 2023-02-27
Chatbot rendering image broken when file name has space in it


dawoodkhan82:
### Describe the bug

When running `chatbot_multimodal` demo, and uploading an image with spaces in the file name, the image does not render in the chatbot.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

`chatbot_multimodal` demo

### Logs

```shell
N/A
```

### System Info

```shell
3.19.1
```

### Severity

annoying

dawoodkhan82:
The markdown parser `markdown-it` doesn't support links with spaces (https://github.com/markdown-it/markdown-it/issues/147). It doesn't recognize it as a link.

abidlabs:
Very intriguing. Does URL encoding the space (converting it to "%20") work?

dawoodkhan82:
@abidlabs the dev can add `image.name = image.name.replace(" ", "%20")`. I guess we can add this in the postprocessing for the chatbot. But I'm wondering how I can determine the message being passed in the `postprocess` of chatbot is an image url vs text.

abidlabs:
Cool that’s good to know, yeah I think we’d have to do some regex to figure out that it’s an image file path in Markdown’s image format. ChatGPT is pretty good at regex btw!

dawoodkhan82:
Seems like this change fixed this issue (https://github.com/gradio-app/gradio/pull/3370/files#diff-82459fb7cccafabf1195b5a189b3bdef828616963ca2ff3b8dfafc5e64f8a487R14-R22)

abidlabs:
I can't reproduce this issue anymore on `main` either @dawoodkhan82, can you?

dawoodkhan82:
It's actually because we're replacing spaces in the demo code `history = history + [(f"![](/file={quote(image.name)})", "Cool pic!")]`. I'll push a fix for this issue in the postprocessing.

Issue #2768, opened 2022-12-07
support Audio IO in Chatbot


sijunhe:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
Given that Chatbot now supports images after v3.12.0, I think we should double down on the multi-modality and add Audio IO to the chatbot interface. It will be a great interface for Voice Assistant demos.

**Describe the solution you'd like**  
Gradio already all the building blocks for this feature and just need to put it together. I think something like the WeChat interface will be nice. Inside the chatbot interface, we replace each text with a bar, which is the Audio output and plays the audio when clicked. Then we can put the audio input button somewhere prominent for easy input. 
![image](https://user-images.githubusercontent.com/11987277/206097158-e86652ce-3923-4825-b8f8-2c092234f742.png)

paulocoutinhox:
+1

abidlabs:
cc @dawoodkhan82

rishikeshF:
+1

abidlabs:
Hi @rishikeshF you can already add audio files to your `gr.Chatbot` as long as you are using the latest version of gradio. Upgrade to the latest version (`pip install --upgrade gradio`) and then take a look at this section to add media files: https://gradio.app/creating-a-chatbot/#adding-markdown-images-audio-or-videos

jpmcarrilho:
Hello, I'm building a full ai assistant bot on top of gradio (wwd, s2t, information retrial), I have to control many interfaces and components rendering or given outputs. There's no good reactive and fast response on any component, but the chatbot element, that's why I love it. It's very powerful and straightforward usage. But this issue is still not finished I think, we need the capability of recording audio the same way we write a text and press enter to interact now: record with audio, stop recording, press enter, deal with the possible post-process (wav2vec2, whisper), use LLMs and neural search. I think this is the correct implementation. And the field tendency is to add more and more multimodality, write, record, and send an image, as of today's 'XXX GPTs'. This implementation would certainly be a nice addition to the library.

abidlabs:
Hi @jpmcarrilho just to confirm, you'd like to be able to record audio, do some preprocessing and display it in a Chatbot? This is already possible by using the `gr.Audio(source="microphone")` component

jpmcarrilho:
[...]

PR #3413, opened 2023-03-07
Making the chatbot even more multimodal

dawoodkhan82:
# Description
- Adding support for audio + video in chatbot
- Now images can be uploaded to chatbot without markdown format
- Also makes issue [#3339](https://github.com/gradio-app/gradio/issues/3339) obsolete

https://user-images.githubusercontent.com/12725292/223579746-740752fd-5ee6-461b-ba7c-f97f149c313c.mov

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

Closes: #2768, Closes: #3339

# Checklist:

# A note about the CHANGELOG


abidlabs:
Whoa this looks incredible!

abidlabs:
Technically, this could be a problem for users if they type in a string message which happens to be the path of an image/audio/file, correct?

dawoodkhan82:
> if I understand correctly, this could be a problem for users if they type in a string message which happens to be the path of an image/audio/file, right?

Right, so if the message is a valid file path, then it will show that image/audio/video. If someone really wants to write out a filepath they can wrap it in some markdown syntax or include any text as part of the message.

> Also is there any way to include images/video/audio along with text in a single chat message as we've seen in some multimodal models chatbot demos?

You can't include video/audio in the same message as text. But we also now have support for `None`, so you can send multiple messages from either the user/bot. Which I think is better.

> One alternative could be to extend the meaning of  from referring just to images, to media resources of any kind. So you could embed an audio or a video the same way as you currently do an image: . Apparently, some other [markdown parsers have considered the same idea](https://talk.commonmark.org/t/embedded-audio-and-video/441).

I could look into it. But I don't really like the idea of the dev _having_ to write markdown to get the chatbot to work with other gradio input components. I feel like our input components should easily plug into the chatbot output, which is what I was trying to achieve with this PR.

pngwn:
I think we have to remember that the users of these chatbots are hopefully not technical users, so they won't know thee first thing about markdown. And the image syntax for markdown is one its ugliest and least intuitive features. 

I don't mind us supporting markdown _as well_, like whats app + discord do, but we also need normal uploads.

I'd like it to be more like discord, slack, etc. Where you can attach asset to _accompany_ a message (which is text by default but doesn't _need_ to contain any).

pngwn:
I think it's far messier for us to use markdown because if we want to tweak anything then we need to do it via markdown it plugins (which are not fun) rather than just dealing with data.

abidlabs:
Ok I'm convinced that the breakage will be very rare and that it's more intuitive not to use Markdown. Will go ahead and test out this PR

demo/chatbot_multimodal/run.py
```diff
@@ -5,22 +5,21 @@ def add_text(history, text):
...
-    chatbot = gr.Chatbot(elem_id="chatbot").style(height=500)
+    chatbot = gr.Chatbot(elem_id="chatbot").style(height=750)
     
     with gr.Row():
         with gr.Column(scale=0.85):
             txt = gr.Textbox(show_label=False, placeholder="Enter text and press enter, or upload an image").style(container=False)
         with gr.Column(scale=0.15, min_width=0):
-            btn = gr.UploadButton("🖼️", file_types=["image"])
-            
+            btn = gr.UploadButton("🖼️", file_types=["image", "video", "audio"])
```
abidlabs, line 19:
```suggestion
            btn = gr.UploadButton("📁", file_types=["image", "video", "audio"])
```
since we now accept multiple files!

gradio/components.py
```diff
@@ -3993,9 +3990,24 @@ def update(
...
+            return self.md.renderInline(chat_message)
+
     def postprocess(
-        self, y: List[Tuple[str | None, str | None]]
-    ) -> List[Tuple[str | None, str | None]]:
+        self, y: List[Tuple[str | None | Dict, str | None | Dict]]
+    ) -> List[Tuple[str | Dict | None, str | Dict | None]]:
         """
         Parameters:
             y: List of tuples representing the message and response pairs. Each message and response should be a string, which may be in Markdown format.
```
abidlabs, line 4013:
```suggestion
            y: List of tuples representing the message and response pairs. Each message and response should be a string, which may be in Markdown format. It can also be a string filepath or URL to an image/video/audio in which case the file is displayed. It can also be None, in which case that message is displayed.
```

gradio/components.py
```diff
@@ -3993,9 +3990,24 @@ def update(
...
+            }
+        elif isinstance(chat_message, dict) or chat_message is None:
+            return chat_message
+        else:
+            return self.md.renderInline(chat_message)
+
     def postprocess(
-        self, y: List[Tuple[str | None, str | None]]
-    ) -> List[Tuple[str | None, str | None]]:
+        self, y: List[Tuple[str | None | Dict, str | None | Dict]]
```
abidlabs, line 4009:
```suggestion
        self, y: List[Tuple[str | None, str | None]]
```

abidlabs: I don't think the input type has changed, only the output, right?

gradio/components.py
abidlabs, line 4013:
```suggestion
            y: List of tuples representing the message and response pairs. Each message and response should be a string, which may be in Markdown format.  It can also be a string filepath or URL to an image/video/audio in which case the file is displayed. It can also be None, in which case that message is displayed.
```

gradio/components.py
```diff
@@ -3993,9 +3990,24 @@ def update(
         }
         return updated_config
 
+    def _process_chat_messages(self, chat_message: str | Dict | None):
+        if isinstance(chat_message, str) and os.path.exists(chat_message):
```
abidlabs, line 3994: I'm confused how `chat_message` could ever be a `Dict`:
```suggestion
    def _process_chat_messages(self, chat_message: str | None):
        if chat_message and os.path.exists(chat_message):
```

abidlabs:
Backend LGTM @dawoodkhan82! Left a few suggestions above. 

I also think we should also check for URLs but that's not so critical and can be a separate PR if users request it.

abidlabs (approved):

ui/packages/app/src/components/Chatbot/Chatbot.svelte
```diff
@@ -4,20 +4,33 @@
...
 
+	$: _value = value
+		? value.map(([user_msg, bot_msg]) => [
+				typeof user_msg === "string"
+					? redirect_src_url(user_msg)
+					: normalise_file(user_msg, root, root_url),
+				typeof bot_msg === "string"
+					? redirect_src_url(bot_msg)
+					: normalise_file(bot_msg, root, root_url)
+		  ])
```
pngwn, line 32: Could we break this callback to map into its own function, just for ease of readability.

pngwn (approved):
Just a small comment but everything else looks good to me.

abidlabs:
Thanks for the review @pngwn! I added an entry in the changelog so the only things left are:
* addressing your comment
* fixing the typecheck

pngwn:
I think the type error is just a test that needs updating.

aliabid94:
Overall, very cool! What I worry about is this automatic check if a response is a file. There's a risk that the bot will coincidentally return a path that happens to be a file name, and that will cause issues.

What I suggest: if returning a media type, return a tuple with the file path and associated caption (None if there is no caption). That way, we can be certain about the intent to return a file.

E.g.

```python 
chat_history = [
  ["who are you", "I am an artist chatbot"],
  ["Draw me an image of a cat", ("cat,jpeg", "A cat illustration")],
]
```

abidlabs:
Although I had basically the same concern @aliabid94, I think it's an okay risk to take. In fact, GitHub-flavored markdown basically does the same thing for displaying videos (although with URLs and not filepaths). The nested tuple structure seems too complicated for me

aliabid94:
It's different in the case of GitHub, you can see and adjust your markdown if it does not render correctly. The bot will fail in realtime without awareness from the demo creator.

aliabid94:
If you don't want to provide a caption, you can do this:
```python
messages = ["I want a cat pic", ("cat.jpg",)]
```

abidlabs:
Synced with @aliabid94 and I'm convinced that the tuple structure (no caption, just a tuple) is better than just putting in the filename b/c of the risk of accidentally exposing files that you don't want to users. 

Furthermore, we realized that there's actually a big security risk with the current implementation. Right now, **any filepath that exists on the host comptuer can be accessed via the Chatbot.** That's because when the user puts in a message in the Chatbot, the Chatbot checks if the file exists, and if it does, it adds it to the list of allowed files, even if that file is outside of the gradio app directory. Proof of concept:

image


If there are no objections, we'll modify this PR to fix these issues. cc @dawoodkhan82 @pngwn @freddyaboulton

aliabid94:
What I had suggested was that you could return a single element tuple `("cat.jpg",)`, or a tuple with caption `("cat.jpg", "A cat pic")`.

freddyaboulton:
Good catch! I definitely agree we should make sure that only files in the allowed directories are served.

I'm ok with the proposed tuple structure to differentiate files vs regular strings but wouldn't returning a pathlib Path be more explicit that the message is supposed to be a file?

abidlabs:
One thing we may want to add in the near future is support for URLs so something more general than pathlib.Path would be nice

pngwn:
> any filepath that exists on the host comptuer can be accessed via the Chatbot.

This is arguably true for a few components, depending how the author allows them to be used. File component is probably the biggest offender. I always thought our position was always that it is down to the app author to protect against this?

I do think it is more likely with the chatbot though.

abidlabs:
> This is arguably true for a few components, depending how the author allows them to be used. File component is probably the biggest offender. I always thought our position was always that it is down to the app author to protect against this?

> I do think it is more likely with the chatbot though.

To recap the discussion from Slack, although several different components are able to serve files, the developer should always *knowingly* allow this. For example, for the `File` component, if the function returns a fixed string path, only that path would be exposed to users. It's only when the function returns a string that is dynamically generated would there be a risk that arbitrary files can be accessed. The problem with the current implementation of the `Chatbot` is that for pretty much any function, the user can simply type in a file path and the `Chatbot` will try to find that path and return the media object. I'll make the changes discussed above to this PR to mitigate that.

abidlabs:
Ok I've made the changes. As a side benefit, we now support URLs as well. The only thing left is:

* Some type checks are still failing
* Addressing @pngwn's comment above

and then we should be good to merge!

abidlabs:
Added a test, LGTM!

abidlabs (approved):

aliabid94:
@abidlabs can you update guides/06_other-tutorials/creating-a-chatbot.md as well

abidlabs:
Cleaned up the Guide and demos. Perhaps @dawoodkhan82 you can make a final pass tomorrow and then we can merge it in for release?

Issue #2560, opened 2022-10-28
Events that are triggered when another event is complete - needed for several bugs


vzakharov:
The `_js` parameter is very useful if you want to do something before running the python code (i.e. making the API call), but sometimes you want to do the opposite.

For example, I’m using [WaveSurfer](http://wavesurfer-js.org/) to display/play audio in my app, and I need to run `wavesurfer.load('/file=preview.wav ')` once the audio is generated.

So far, I’m still struggling to find an optimal solution for this, choosing between:
- Creating an (artificial) checkbox that would fire once the audio is loaded and whose firing will on turn run the script (via `_js`)
- Creating a script HTML as the output of the audio loading function that would do the same

Both approaches look very makeshift, while having a `_js_after_python` would make this really easy and elegant.

Hope this helps!

abidlabs:
@aliabid94 @pngwn as we think about making `js` an officially supported parameter, perhaps we should have both a `js_before` and a `js_after`?

Another idea that might be more useful actually would be to have a `event.then()` function which allows you to trigger events when other events complete. I've needed this use case often and usually have had to come up with a hack

abidlabs:
Actually another more general idea would be to have a `event.then()` function which allows you to trigger events when other events complete. I've needed this use case often and usually have had to come up with a hack

pngwn:
@abidlabs Do you have an example usecase for `event.then()`, I'd like to understand more.

abidlabs:
Sure, @vzakharov's original post is one example, where you could define a regular backend function and then use `.then()` to define a js function that runs after the backend function is complete. [Here's a similar example](https://github.com/gradio-app/gradio/blob/main/demo/sound_alert/run.py), where a sound rings after a prediction is complete. 

There are lots of use cases where people want to do something _after_ a prediction is finished running. I remember seeing a user ask about having a prediction which produces an image and saves it to disk, and then running a 2nd function with that image afterwards. Can't find the issue right now, but it was similar to this one: https://github.com/gradio-app/gradio/discussions/2491

vzakharov:
Oh, `then` would be even better!

Note that this would likely also mean allowing events to only have local
(js) processing without making an API call. (Currently I have to do weird
hacks like intentionally raising a JavaScript error to avoid making a call.)

abidlabs:
@vzakharov can't you just set `fn=None` in your event trigger to avoid making an API call?

vzakharov:
[...]

PR #3414, opened 2023-03-07
Clean up event listener logic, add support for consecutive events with `.then`

aliabid94:
Cleaned up the repeated logic across all event listeners. Created a Callable `EventListenerMethod` and appended it as an attribute in the constructor of the EventListeners.

Fixes: #2560

Also added the ability to chain event listeners, using a `.then()` method. For example:
```python
num1 = gr.Number()
num2 = gr.Number()
num3= gr.Number()
num4 = gr.Number()
btn = gr.Button()

btn.click(add, [num1, num2], num3).then(square, num3, num4)
```

See a more useful example in demo/chatbot_simple_demo.


abidlabs:
Niceeee

gradio/events.py
```diff
@@ -78,703 +96,123 @@ def change(
...
-        queue=None,
-        batch: bool = False,
-        max_batch_size: int = 4,
-        preprocess: bool = True,
-        postprocess: bool = True,
-        cancels: Dict[str, Any] | List[Dict[str, Any]] | None = None,
-        every: float | None = None,
-        _js: str | None = None,
-    ):
+    def __init__(self):
```
freddyaboulton, line 132: The typehints are preserved which is awesome but we're losing the docstring.

### Main
image


### This branch
image


I think a good workaround would be manually adding the docstring of `EventListenerMethod` to the docstring you already have here. For example,

```python

    def __init__(self):
        self.click = EventListenerMethod(self, "click")
        f"""
        This event is triggered when the component (e.g. a button) is clicked.
        This method can be used when this component is in a Gradio Blocks.
        
        {self.click.__doc__}
        """
```

gradio/events.py
```diff
@@ -32,19 +32,40 @@ def set_cancel_events(
...
-    pass
+    def __init__(self):
+        for event_listener_class in EventListener.__subclasses__():
+            if isinstance(self, event_listener_class):
+                event_listener_class.__init__(self)
 
 
-class Changeable(EventListener):
-    def change(
+class Dependency(dict):
```
aliabid94, line 41: inherits from dict so event listener return type is backwards compatible

freddyaboulton (commented):
I love the clean up @aliabid94 ! Only one comment on that, rest looks good to me.

I'm not 100% sure about `then` though. It works great but I think last time we talked about this, there were some concerns about implementing it. I'll try to dig that up.

Also not sure about having the canonical chatbot "hello world" example use `then`.

gradio/events.py
aliabid94: clever! will do

aliabid94: hmm I still didn't get the type hint - does it work for you @freddyaboulton?

aliabid94:
happy to rekindle that conversation! (some of it is in the linked issue).

The chatbot was actually what pushed me to implement `.then`, I didn't like that there was a wait and then both the user and bot respond together.

gradio/events.py
aliabid94: btw the behaviour in this branch rn for me is: I get the docstring while typing `click`, and argument type hints when I open the parentheses.

freddyaboulton: Dang yea thought this was working for me. Let me think about this some more, I think the problem is that the `click` is determined at runtime. Should be fine to proceed without this though?

abidlabs: I get the same thing @aliabid94. Docstring while typing in `.click()`: 

image

Type hints after opening the parentheses:

image

Very cool!

pngwn:
Can you add more context to the PR description please. There was a lot of discussion about .then with a number of concerns, I'm not sure if these have been addressed. 

I'm not personally convinced that this is the right approach, I don't think it is a good API. 

Feels like we're just ramming in a bunch of new API atm because it enables something useful for one usecase without actually considering the API that much.

pngwn:
I still don't think this feature is an idealw solution for the problem stated because the demo added doesn't really give normal chatbot functionality. You actually need to set the textbox as an output of the first function too:

```py
def add_message(chat_history, message):
    return [chat_history + [[message, None]], ""]

def respond(chat_history):
    time.sleep(1)
    response = random.choice(["Yes", "No"])
    chat_history[-1][1] = response
    return chat_history

with gr.Blocks() as demo:
    chatbot = gr.Chatbot()
    msg = gr.Textbox()
    clear = gr.Button("Clear")

    msg.submit(add_message, [chatbot, msg], [chatbot, msg]).then(
        respond, chatbot, chatbot
    )
    clear.click(lambda: None, None, chatbot)
```

This still feels like a lot of work to meet the needs of the standard usecase for the chatbot.

abidlabs:
Let's split up the PR, will make it easier to discuss the implications of adding `.then()` to the API. Just to clarify, though, there are definitely lots of concrete use cases for `.then()`. I mention 2 in the original issue, having a proper chatbot is good reason, and there are others that have come up recently I can describe.

aliabid94:
ok will split this up, but let's continue the `.then` discussion in the issue

PR #3415, opened 2023-03-08
Remove codecov from the repo 

aliabd:
Codecov has been acting erratically and failing for no reason sometimes. This PR removes it from the repo. 

* removes the action from `backend.yml` (but keeps the coverage command in case we want to do something else for coverage...)
* removes the `codecov.yml` file
* removes it from the READMEs


freddyaboulton (approved):
Looks good to me @aliabd !

Thanks for taking care of this

abidlabs:
LGTM @aliabd! This will also make the experience for contributors a bit less confusing, thanks!

PR #3417, opened 2023-03-08
Fix link

eltociear:
# Description

Fixed link in Chinese README.


abidlabs:
Thanks @eltociear for the fix! Can you add an entry in `CHANGELOG.md` so that we can credit you for the change?

abidlabs:
Added an entry to the changelog so that we can merge this in. Thanks again for the fix @eltociear

eltociear:
@abidlabs 
Sorry it took me so long to read your message.
Thanks for adding me!

PR #3420, opened 2023-03-08
Event listener cleanup

aliabid94:
Clean up event listener logic.


abidlabs:
@aliabid94 are all of these changes just for the event listener cleanup? So many frontend files are changed

aliabid94:
Okay finally this PR should be ready for review!

To avoid code reuse, I changed the event listeners from being class-bound methods to instance attributes that are objects of EventListener method type, which has a `__call__` method. The call method uses the same arguments and parameter docstrings, so we save a lot on code reuse there.

However we want the methods to have different method docstrings while sharing parameter docstrings. We do this by attaching attribute documentation via a string placed right after the attribute. VSCode respects these attribute docstrings.

However! There is no way to cleanly extract these attribute docstrings when we generate documentation for the website - Python ignores attribute docstrings so we can't use `.__doc__` to get them like we usually would for class-bound methods. So instead, we have to use `inspect.getsource` to read the source code of the constuctor method itself and manually extract the attribute documentation.

In the process, I also changed the way documentation works, so the documentation of these event listeners is automatically inherited by their subclass Components.

gradio/documentation.py
```diff
@@ -16,18 +17,41 @@ def set_documentation_group(m):
...
+def document(*fns, inherit=False):
     """
     Defines the @document decorator which adds classes or functions to the Gradio
     documentation at www.gradio.app/docs.
 
     Usage examples:
     - Put @document() above a class to document the class and its constructor.
-    - Put @document(fn1, fn2) above a class to also document the class methods fn1 and fn2.
+    - Put @document("fn1", "fn2") above a class to also document methods fn1 and fn2.
+    - Put @document("*fn3") with an asterisk above a class to document the instance attribute methods fn1 and fn2.
```
abidlabs, line 48: I assume this should be
```suggestion
    - Put @document("*fn3") with an asterisk above a class to document the instance attribute methods f3.
```

abidlabs: Could we make the syntax be `classname().fn_name` to be more explicit and understandable to anyone who reads through the codebase? Still very easy to parse (just look for a `.` in the function name)

gradio/documentation.py
```diff
@@ -16,18 +17,41 @@ def set_documentation_group(m):
...
+
+def document(*fns, inherit=False):
     """
     Defines the @document decorator which adds classes or functions to the Gradio
     documentation at www.gradio.app/docs.
 
     Usage examples:
     - Put @document() above a class to document the class and its constructor.
-    - Put @document(fn1, fn2) above a class to also document the class methods fn1 and fn2.
+    - Put @document("fn1", "fn2") above a class to also document methods fn1 and fn2.
```
abidlabs, line 47: Why did you change from `class methods` to `methods`?

aliabid94: its not a class method, (which is something else), its a class bound instance method, the default behaviour for a method.

gradio/documentation.py
```diff
@@ -16,18 +17,41 @@ def set_documentation_group(m):
...
+    - Put @document("*fn3") with an asterisk above a class to document the instance attribute methods f3.
```
abidlabs, line 48: Could we change the syntax to `Class().fn_name` as it's more obvious to someone reading the codebase about what's happening? Still easy to parse (just look for a `.`)

gradio/documentation.py
```diff
@@ -16,18 +17,41 @@ def set_documentation_group(m):
...
+    lines = [line.strip() for line in code.split("\n")]
+    found_attr = False
+    i = 0
+    for i, line in enumerate(lines):
+        if line.startswith("self." + attr):
+            found_attr = True
+            break
+    assert found_attr, f"Could not find {attr} in {cls.__name__}"
+    start_line = lines.index('"""', i)
+    assert start_line != -1, f"Could not find docstring for {attr} in {cls.__name__}"
```
abidlabs, line 31: How could `start_line` or `end_line` ever be -1?

gradio/documentation.py
aliabid94: what do you mean? Can you give an example

gradio/documentation.py
abidlabs: If `lines.index('"""', i)` fails, a `ValueError` is thrown, so I don't think you need the assertion errors

gradio/documentation.py
abidlabs: Instead of 

```py
@document("*change")
```

which seems kind of magical, do this instead:

```py
@document("Textbox().change")
```

gradio/documentation.py
aliabid94: -1 meaning it could not find the docstring start quotes

abidlabs: That's not how `.index()` works though

gradio/documentation.py
aliabid94: That does not distinguish between instance bound or class bound methods, as they both can be called in that manner. It's also weird to have to rewrite the class name twice like this, whats the point of the decorator. if you have to write the class name again

```python
@document("Editable.edit")
class Editable
  ...
```

abidlabs: But not a big deal, ignore this

abidlabs (approved):
LGTM, demos work great & documentation shows up when I build the website locally!

gradio/documentation.py
aliabid94: Oh you're right, I'm thinking of js

PR #3421, opened 2023-03-08
add code component

### LINKED ISSUE #1014: Feature request: `Code` component
An idea for a new `Code` Component, which supports:
* editable code (for inputs) or displayed code (for outputs)
* simple syntax highlighting for various common languages

Inspired by:
* https://huggingface.co/spaces/facebook/incoder-demo

image

### LINKED ISSUE #3423: Weird behaviour when two options have same substring in Dropdown
### Describe the bug

When the dropdown component has two options sharing the same substring, for example, 'cat' and 'wild cat', the selection would include both options when I click 'wild cat'. But it works fine when I click 'cat'.

The problem only has an impact on the UI; the value still aligns with the latest selection. For example, when I click 'wild cat', although it's showing both 'cat' and 'wild cat' are selected, the value passes into function is 'wild cat'.

### Reproduction

def test(animal):
    return animal

demo = gr.Interface(
    test,
    [gr.Dropdown(["cat", "wild cat"], label="Cats"), multiselect=False],
    "text",
)

if __name__ == "__main__":
    demo.launch()

### Screenshot

_No response_

### Logs

```shell
NA
```

### System Info

```shell
gradio==3.20.1
```

### Severity

annoying

pngwn:
Adds code component.

Closes #1014. Closes #3423.

Still have a bunch of cleanup to do but thought I'd pop this here for testing + feedback.

Check out `demo/code`.


pngwn:
fixed the file output thing in 3.8

abidlabs:
Spent some time thinking about how the `Code` component should handle preprocessing/postprocessing because one of the nice things about Gradio components when used as an input is that they pass data into user functions in very intuitive ways: 

* `gr.Textbox()` sends as a string
* `gr.Image()` (in default conditions) sends as a numpy array
* `gr.Dataframe` (by default) sends as a pandas dataframe. 

Similarly, your function can return data in the most convenient format for you and the output component will be able to infer how to display it. The user doesn't need to worry about putting the data in the correct dictionary format, as is the case right now with the `Code` component, for both the input and output case. 

So here's what I propose for the preprocessing and postprocessing function for `gr.Code` component instead:

**`.preprocess(x): -> str`**

* The `Code` component should _only_ send the code string to the user function. This is going to be the most intuitive format for most users. If we want to keep the same payload from the frontend to the backend, that is fine, but in the `.preprocess()` function, we should only send the string back.
* `Code()` should accept a parameter called `lang` that can be used to set the language. It can be updated using `Code.update()` analogous to the other components parameters. 

**`.postprocess(x: str)`**
* The `Code` component's postprocess should accept just a regular string. Again, for most users, this is going to be the most intuitive format. The `Code` component should check if this string is an existing filepath and if so, treat it as such. Otherwise, it should assume it's a code string. We use this logic in a few places inside the codebase ([such as here](https://github.com/gradio-app/gradio/blob/0036b2e0b05dca734200ed90d81c92e916d408de/gradio/blocks.py#L475)) and hasn't caused any issues.
* But just to be on the safe side, we can add an optional parameter `output_format`, which can be `filepath` or `code`. If provided, it takes precedence instead of the postprocess() method inferring the format.

If you'd like @pngwn I can make a PR into this branch with these changes. (cc @aliabid94 for your thoughts as well)

pngwn:
How does the user function know what language the code is if we only send the code and not the lang?

abidlabs:
Well the language has to be set somewhere in the backend, right? In `demo/code`, you would read it from the `Dropdown`. In other cases, the user would set the language using the `lang` parameter in `Code` so they would already know what the language is.

pngwn:
Likewise how does would an output code component know what language to display the provided code as (as code string not filepath), if that information isn't there.

pngwn:
I guess, feels clunky to map everything together like that when we already have all information in one component. I don't think sending the language to a code viewer / editor is unintuitive at all, quite the opposite. Users would expect to be able to pass code and the lang directly as that is data not app state imo.

abidlabs:
In the output side, if you are working with a single language, you would instantiate the code component with that language, `gr.Code(lang=lang)` and your function is very simple, it just returns a string.

If you are working with multiple languages, you would return `gr.Code.update(code_string, lang=lang)`.

abidlabs:
Dictionaries are annoying to work with in Python -- they don't have any typing info so a user couldn't easily know what keys a dictionaries are expected to have. They'd have to look up the docs to figure out the expected keys, etc. Even `gr.Code.update()` is better since at least you have the type hints.

abidlabs:
Cool, do you want me to PR this?

pngwn:
Sorry was sleeping when you asked, I attempted it myself. It works as far as I can tell but I dunno if i missed anything.

I don't think we need a custom preprocess if we're just delivering the content as is (a string) but maybe I'm wrong.

I just made the frontend + backend match because the frontend doesn't care how it receives the code/ lang, nor does it need to receive the lang every time we pass the code. So i just made it accept the code alone matching postprocess.

I added a static field `languages` to `gr.Code` which lists the supported languages. You can see it used in `demo/code`. I think it is helpful to programatically get a list of supported languages.

 I do not know how to document static attributes with docstrings, according to google it isn't possible cos guido didn't like the proposal but maybe there is a way. 

I made some UI tweaks, made the download + copy buttons work. If there is a language selected they should use it as the file extension. If it is an unknown language then it just creates ` .txt`

pngwn:
i do not know how to write tests for components or if we normally do that, so let me know what i should do.

ui/packages/form/src/DropdownOptions.svelte
```diff
@@ -28,7 +29,12 @@
 				data-value={choice}
 				aria-label={choice}
 			>
-				<span class:hide={!value?.includes(choice)} class="inner-item pr-1">
+				<span
+					class:hide={!(Array.isArray(value) ? value : [value])?.includes(
+						choice
+					)}
+					class="inner-item pr-1"
+				>
```
pngwn, line 37: This was a bug in single mode. `py` was also matching `python` because `python` conatin `py`

ui/packages/form/src/Dropdown.svelte
```diff
@@ -77,6 +77,7 @@
 			if (!multiselect) {
 				value = option;
 				inputValue = "";
+				showOptions = false;
```
pngwn, line 80: This just closes the dropdown after you have selected something in single mode, which is the expected behaviour (and matches the previous).

abidlabs: Thanks yeah I noticed this issue as well

abidlabs:
> i do not know how to write tests for components or if we normally do that, so let me know what i should do.

We do, we basically just ensure that `.preprocess()`, `.postprocess()`, `.serialize()`, `.deserialize()` are working as expected. I'm happy to write them if you'd like, or you can look at the other tests in `test/test_components.py` as a template

ui/packages/form/src/DropdownOptions.svelte
abidlabs: Is this the same bug as https://github.com/gradio-app/gradio/issues/3423?

pngwn:
Do we test with real files where needed?

abidlabs:
Yes we do. We store these files in `test/test_files`

gradio/components.py
```diff
@@ -5330,6 +5331,94 @@ def as_example(self, input_data: str | None) -> str:
...
+        "yaml",
+        "dockerfile",
+        "shell",
+        "r",
+    ]
+
+    def __init__(
+        self,
+        value: str | None = None,
+        language: str | None = None,
```
abidlabs, line 5359: How does the code editor behave when language is `None`? Should we set a popular default like `python`?

gradio/components.py
```diff
@@ -5330,6 +5331,94 @@ def as_example(self, input_data: str | None) -> str:
         return postprocessed if postprocessed else ""
 
 
+@document("change")
+class Code(Changeable, IOComponent, SimpleSerializable):
+    """
+    Creates a Code editor for entering, editing or viewing code.
+    Preprocessing: passes a {str} of code.
```
abidlabs, line 5338:
```suggestion
    Creates a Code editor for entering, editing or viewing code.
    Preprocessing: passes a {str} of code into the function.
```

gradio/components.py
```diff
@@ -5330,6 +5331,94 @@ def as_example(self, input_data: str | None) -> str:
...
+        *,
+        label: str | None = None,
+        show_label: bool = True,
+        visible: bool = True,
+        elem_id: str | None = None,
+        **kwargs,
+    ):
+        """
+        Parameters:
+            value: Default value. If callable, the function will be called whenever the app loads to set the initial value of the component.
```
abidlabs, line 5369:
```suggestion
            value: Default value to show in the code editor. If callable, the function will be called whenever the app loads to set the initial value of the component.
```

gradio/components.py
pngwn: It just display unhighlighted code. I don't think we should set a default because it will be confusing for users who are either using a language we don't support or are using it for plain text. And 'wrong' highlighting is arguably more confusing than no highlighting but i don't feel strongly either way.

gradio/components.py
```diff
@@ -5330,6 +5331,94 @@ def as_example(self, input_data: str | None) -> str:
...
+            language: The language to display the code as. Supported languages listed on `gr.Code.languages`.
```
abidlabs, line 5370:
```suggestion
            language: The language to display the code as. Supported languages listed in `gr.Code.languages`.
```

gradio/components.py
```diff
@@ -5330,6 +5331,94 @@ def as_example(self, input_data: str | None) -> str:
...
+    def __init__(
+        self,
+        value: str | None = None,
+        language: str | None = None,
+        *,
+        label: str | None = None,
+        show_label: bool = True,
+        visible: bool = True,
+        elem_id: str | None = None,
+        **kwargs,
```
abidlabs, line 5365: Does `Code` support an `interactive` parameter like the other components that can be used both as an input and an output?

gradio/components.py
```diff
@@ -5330,6 +5331,94 @@ def as_example(self, input_data: str | None) -> str:
...
+
+    def get_config(self):
+        return {
+            "value": self.value,
+            "language": self.language,
+            **IOComponent.get_config(self),
+        }
+
+    def postprocess(self, y):
+        if y is not None and os.path.exists(y):
```
abidlabs, line 5395: I'd suggest using 
```suggestion
        if y is not None and os.path.isfile(y):
```
to eliminate directories

gradio/components.py
```diff
@@ -5330,6 +5331,94 @@ def as_example(self, input_data: str | None) -> str:
...
+                return file_data.read()
+        return y
+
+    @staticmethod
+    def update(
+        value: str | None | Literal[_Keywords.NO_VALUE] = _Keywords.NO_VALUE,
+        label: str | None = None,
+        show_label: bool | None = None,
+        visible: bool | None = None,
+        language: str | None = None,
```
abidlabs, line 5406: If `Code` does support `interactive`, then let's add it here as well

CHANGELOG.md
```diff
@@ -2,7 +2,8 @@
 
 ## New Features:
 
-No changes to highlight.
+- New code component allows you to enter, edit and display code with full syntax highlighting by [@pngwn](https://github.com/pngwn) in [PR 3421](https://github.com/gradio-app/gradio/pull/3421)
```
abidlabs, line 5: Could we add a screenshot here? I think lots of people would be interested in knowing how it looks!

gradio/components.py
abidlabs: Sg

abidlabs:
Beautiful component @pngwn! Everything works as expected and the copy/download are great touches as well

I've put some relatively minor suggestions on the backend above, but otherwise LGTM!

demo/code/run.py
```diff
@@ -0,0 +1,41 @@
...
+    with gr.Row():
+        code_in = gr.Code(language="python", label="Input")
+        code_out = gr.Code(label="Ouput")
+    btn = gr.Button("Run")
+    btn_two = gr.Button("Load File")
+
+    lang.change(set_lang, inputs=lang, outputs=code_in)
+    btn.click(code, inputs=[lang, code_in], outputs=code_out)
+    btn_two.click(set_lang_from_path, inputs=None, outputs=code_out)
+    gr.Code.languages
```
abidlabs, line 37: Accidental? 
```suggestion
```

pngwn:
@abidlabs thanks! Will write some tests and address the comments.

ui/packages/form/src/DropdownOptions.svelte
pngwn: yus! will add to body!

pngwn:
Screenshot 2023-03-09 at 18 06 36

abidlabs:
Beautiful! I shoulda tweeted out that screenshot

pngwn:
software ppl love darkmode for some reason

pngwn:
@abidlabs I have added the interactive kwarg and written some tests. 

The component is pretty simple on the backend so they are quite basic to make sure it doesn't do something stupid but lmk if anything else is needed.

gradio/components.py
```diff
@@ -5330,6 +5331,102 @@ def as_example(self, input_data: str | None) -> str:
...
+        updated_config = {
+            "label": label,
+            "show_label": show_label,
+            "visible": visible,
+            "value": value,
+            "language": language,
+            "interactive": interactive,
+            "__type__": "update",
+        }
+        return updated_config
```
abidlabs, line 5424:
```suggestion
        updated_config = {
            "label": label,
            "show_label": show_label,
            "visible": visible,
            "value": value,
            "language": language,
            "__type__": "update",
        }
        return IOComponent.add_interactive_to_config(updated_config, interactive)
```

abidlabs:
LGTM this is awesome @pngwn! Thanks for handling all of the backend stuff

abidlabs (approved):

ui/packages/code/Index.svelte
```diff
@@ -0,0 +1,60 @@
...
+	export let elem_id: string = "";
+	export let visible: boolean = true;
+	export let mode: "static" | "dynamic";
+	export let label: string = "Code";
+	export let loading_status: LoadingStatus;
+
+	let dark_mode = target.classList.contains("dark");
+
+	$: dispatch("change", value);
+	$: console.log(mode);
```
aliabid94, line 30: remove

gradio/components.py
```diff
@@ -5330,6 +5331,101 @@ def as_example(self, input_data: str | None) -> str:
         return postprocessed if postprocessed else ""
 
 
+@document("change")
```
aliabid94, line 5334: as of #3420, you can remove the "change", and just have `@document()`, event listener documentation is automatically added

gradio/components.py
```diff
@@ -5330,6 +5331,101 @@ def as_example(self, input_data: str | None) -> str:
...
+
+    def get_config(self):
+        return {
+            "value": self.value,
+            "language": self.language,
+            **IOComponent.get_config(self),
+        }
+
+    def postprocess(self, y):
+        if y is not None and os.path.isfile(y):
```
aliabid94, line 5398: not secure, we have TempFileManager to handle sending files back. (I can take care of this if you'd like.)

ui/packages/code/interactive/Code.svelte
```diff
@@ -0,0 +1,194 @@
...
+<div class="wrap">
+	<div class="codemirror-wrapper {classNames}" bind:this={element} />
+</div>
+
+<style>
+	.codemirror-wrapper {
+		padding-top: 25px;
+		min-height: 250px;
+		max-height: 480px;
+		overflow: scroll;
```
aliabid94, line 183: can we use `overflow: auto`? otherwise I get the scrollbars visible even when there isn't overflow content

ui/packages/code/interactive/Copy.svelte
```diff
@@ -0,0 +1,59 @@
...
+		width: 22px;
+		height: 22px;
+	}
+
+	.check {
+		position: absolute;
+		top: 0;
+		right: 0;
+		z-index: var(--layer-top);
+		background: var(--color-green-500);
```
aliabid94, line 53: not a big fan of using green to signify "complete" - would rather use `--color-accent-primary`. It will flow with themes better.

ui/packages/code/interactive/Download.svelte
```diff
@@ -0,0 +1,91 @@
...
+	import { fade } from "svelte/transition";
+	import { Download, Check } from "@gradio/icons";
+
+	export let value: string;
+	export let language: string;
+
+	$: ext = get_ext_for_type(language);
+
+	function get_ext_for_type(type: string) {
+		console;
```
aliabid94, line 12: typo? what does this do

ui/packages/code/interactive/Download.svelte
```diff
@@ -0,0 +1,91 @@
...
+		position: relative;
+		cursor: pointer;
+		padding: 5px;
+
+		width: 22px;
+		height: 22px;
+	}
+
+	.copied {
+		color: var(--color-green-500);
```
aliabid94, line 77: same comment as above

aliabid94 (approved):
Overall looks great!

ui/packages/code/interactive/Code.svelte
```diff
@@ -0,0 +1,194 @@
...
+	});
+</script>
+
+<div class="wrap">
+	<div class="codemirror-wrapper {classNames}" bind:this={element} />
+</div>
+
+<style>
+	.codemirror-wrapper {
+		padding-top: 25px;
```
aliabid94, line 180: In the theme PR, I changed how we handle the space for the BlockLabel. Instead of having to provide padding-top/margin-top in the nested component, just set the `<BlockLabel float={false}>` This makes `position: static` for BlockLabel, so the nested component is automatically positioned underneath. Otherwise, if the user changes the font or padding of BlockLabel, this specific padding will not be sufficient

ui/packages/code/interactive/Download.svelte
pngwn: typo

ui/packages/code/interactive/Copy.svelte
pngwn: Green is pretty universally a 'success' or 'completion' colour, not really sure why we'd deviate from that but I'll make the change.

ui/packages/code/interactive/Code.svelte
pngwn: This is applied to the code editor itself. I don't want to float the block label, I want the code to scroll underneath the label because it looks far cleaner. So it needs to be absolutely positioned. I'll change it to rems, so it scales with font size but I don't know how to account for padding applied elsewhere.

gradio/components.py
abidlabs: Good catch

pngwn: could you please elaborate.

pngwn:
Addressed most comments.

The padding of the code-mirror instance now computes its value from the style values of the block-label. Updating those variables will now update the padding for the code container.

Obviously these styles are absurd but just to show it working.

Screenshot 2023-03-10 at 12 12 01

Screenshot 2023-03-10 at 12 12 15

I tried the primary colour for the copy/download feedback but orange just looks like a warning. I've just use the default background colours instead.

Still waiting for elaboration on the file issue, so I can address it.

ui/packages/code/interactive/Code.svelte
pngwn: fixed this now.

ui/packages/code/interactive/Download.svelte
pngwn: swapped to neutral 'background' colour. orange was too 'warningy'.

ui/packages/code/interactive/Copy.svelte
pngwn: swapped to neutral 'background' colour. orange was too 'warningy'.

gradio/components.py
abidlabs: After thinking about this some more, we don't actually have to use `TempFileManager` here, for reasons discussed in this internal thread: https://huggingface.slack.com/archives/C02SPHC1KD1/p1678445758606429

pngwn:
small documentation fix (thanks @freddyaboulton !). one last ci run and ill merge.

PR #3428, opened 2023-03-09
Sharing themes

freddyaboulton:
# Description

Adds a `push_to_hub` and `from_hub` class methods to the `ThemeClass` that allows uploading and downloading themes from the hub.  Also added a `upload-theme` cli command.

### Uploading API
image

image

### Space that's created
https://huggingface.co/spaces/freddyaboulton/dracula_revamped

image

### Downloading API

You can specify a theme with a string with the following pattern `<author>/<repo-name>@<semver-expression>`

Not specifying a version will match the latest version. For example of the semver syntax, see the unit tests in `test_theme_sharing.py`

```python
# match latest
with gr.Blocks(theme="freddyaboulton/dracula_revamped") as demo:
    ....

# match a version range
with gr.Blocks(theme="freddyaboulton/dracula_revamped@>=0.1.0,<0.2") as demo:
    ...
```

You can also instantiate a theme with `from_hub` and then pass that to blocks:

```python

dracula = gr.themes.Base.from_hub("freddyaboulton/dracula_revamped@>=0.1.0,<0.2")

with gr.Blocks(theme=dracula) as demo:
    ...
```

# Checklist:

# A note about the CHANGELOG


abidlabs:
Just a general questionas I haven't looked too closely yet. Isn't a theme at the end of the day just a css file? If so, why are we not saving as a simple css file as opposed to a json file?

Perhaps it's better to save it as a json because we might rename css variables, etc. and we have more control over the Theme class and can ensure its backwards compatibility. Is that the reason?

freddyaboulton:
Good question @abidlabs !

Yea the main reason is because the underlying css variables may change but the python theme api should be stable. 

Also, it's hard to go from a css file to a python theme instance since the python class does some "magic" to go from its instance attributes to their css representation. 

If we just wanted to be able to load upstream themes into demos, aka `gr.Blocks(theme="....")`, then just the css would suffice but I think it's important to be able to instantiate themes from the hub since that lets users easily "remix" each other's themes. For example, 

```python
dracula_remix = ThemeClass.from_hub('freddyaboulton/dracula').set(<change-some-button-colors>)
```

abidlabs:
Makes sense!

freddyaboulton:
Still writing out docstrings and other cosmetic changes but this should be good to review now!

gradio/themes/base.py
```diff
@@ -32,6 +42,146 @@ def repl_func(match):
...
+        _ = semver.Version(version)
+
+        from gradio import __version__
+        from gradio.themes import app
+
+        api = huggingface_hub.HfApi()
+
+        author = api.whoami(hf_token)["name"]
+
+        huggingface_hub.create_repo(
```
freddyaboulton, line 119: This is the structure of the uploaded space

https://huggingface.co/spaces/freddyaboulton/dracula_revamped/tree/main

image

All of the theme versions are in the `themes` directory.

image


We had talked about using different branches for the different versions but I think it will be clearer and more discoverable if everything is on the main branch

gradio/themes/app.py
```diff
@@ -0,0 +1,139 @@
...
+    toggle_dark.click(
+        None,
+        _js="""
+        () => {
+            document.body.classList.toggle('dark');
+            document.querySelector('gradio-app').style.backgroundColor = 'var(--color-background-primary)'
+        }
+        """,
+    )
+
```
freddyaboulton, line 26: Think it would be really cool to add a dropdown to let users switch between the different versions of a theme at runtime. Will think about how to do that.

requirements.txt
```diff
@@ -22,4 +22,6 @@ httpx
 pydantic
 websockets>=10.0
 typing_extensions
-aiofiles
\ No newline at end of file
+aiofiles
+huggingface_hub
```
freddyaboulton, line 26: I thought huggingface_hub was already added for the new analytics?

The new semantic_version dependency makes finding a right semver match a one liner.

CHANGELOG.md
```diff
@@ -2,7 +2,68 @@
 
 ## New Features:
 
-No changes to highlight.
+### Theme Sharing 🎨 🤝
```
abidlabs, line 5: Given that we've never actually added a changelog entry for themes, I think we should have one big entry in the changelog for themes which includes both creation and sharing. But @aliabid94 can do that in his PR with the theme documentation

CHANGELOG.md
```diff
@@ -2,7 +2,68 @@
...
+You can now share your gradio themes with the world!
+
+After creating a theme, you can upload it to the HuggingFace Hub to let others view it, use it, and build off of it!
+
+### Uploading
+There are two ways to upload a theme, via the theme class instance or the command line.
+
+1. Via the class instance
+```python
+my_theme.to_hub(repo_name="my_theme",
```
abidlabs, line 16: Should we rename to `push_to_hub()` to make it consistent with [other HF libraries](https://huggingface.co/docs/transformers/v4.26.1/en/main_classes/trainer#transformers.Trainer.push_to_hub)?

CHANGELOG.md
```diff
@@ -2,7 +2,68 @@
...
+                version="0.2.0",
+                hf_token="...")
+```
+
+2. Via the command line
+
+First save the theme to disk
+```python
+my_theme.dump(path="./", filename="my_theme.json")
```
abidlabs, line 25: Is there a reason you separate `path` and `filename` instead of just accepting a filepath e.g. like `json.dump`?

CHANGELOG.md
freddyaboulton: Agreed

gradio/themes/app.py
```diff
@@ -0,0 +1,139 @@
+import time
+
+import gradio as gr
+
+with gr.Blocks(theme=gr.themes.Default()) as demo:
+    gr.Markdown(
+        """
+    # {THEME} Preview
+
+    To use this theme, set `theme='{AUTHOR}/{SPACE_NAME}'` in `gr.Blocks()` or `gr.Interface`.
```
pngwn, line 10:
```suggestion
    To use this theme, set `theme='{AUTHOR}/{SPACE_NAME}'` in `gr.Blocks()` or `gr.Interface()`.
````

for consistency?

Comment (freddyaboulton) [reply gradio/themes/app.py:10]: Good catch! Pushing up now

File: CHANGELOG.md (hunk 6 above)
Comment (freddyaboulton) [reply CHANGELOG.md:25]: None 🙈 Making the change + pushing up now.

File: gradio/themes/base.py [hunk 8]
```diff
@@ -32,7 +46,179 @@ def repl_func(match):
...
+            repo_type="space",
+            filename=f"themes/theme_schema@{matching_version.version}.json",
+        )
+        download = requests.get(url)
+        if download.ok:
+            return cls.from_dict(download.json())
+        else:
+            raise ValueError(
+                f"Error downloading vile themes/theme_schema@{matching_version}.json from {repo_name}!"
+            )
```
abidlabs, line 122: Perhaps we should fallback to `DefaultTheme()` if the theme cannot be found and raise a warning instead of an exception? I'm thinking about the case where someone is using someone else's theme and the theme gets deleted. It would be good if the Space is still runnable

abidlabs:
It looks like Markdown text with the `dracula_revamped` theme doesn't show up that clearly in light mode:

image

I think this would be a great theme to show off so it would be great if the text is working. Not sure if this is an issue that needs be fixed in the Base theme or in Dracula

gradio/themes/app.py
```diff
@@ -0,0 +1,139 @@
...
+                (
+                    "https://gradio-static-files.s3.us-west-2.amazonaws.com/logo.png",
+                    "logo",
+                ),
+                (
+                    "https://gradio-static-files.s3.us-west-2.amazonaws.com/tower.jpg",
+                    "tower",
+                ),
+            ]
+        )
```
abidlabs, line 111: Currently, the Gallery is taking up too much vertical space
```suggestion
        ).style(height="200px", grid=2)
```

gradio/themes/base.py
```diff
@@ -1,8 +1,22 @@
...
+from huggingface_hub import CommitOperationAdd
+
+from gradio.documentation import document, set_documentation_group
+from gradio.themes.utils import colors, get_matching_version, get_theme_assets, sizes
+from gradio.themes.utils.readme_content import README_CONTENT
+
+set_documentation_group("themes")
 
 
 class ThemeClass:
```
abidlabs, line 22: Consider renaming to `class Theme`

gradio/themes/base.py
```diff
@@ -32,7 +46,179 @@ def repl_func(match):
         dark_css += "}"
         return css + "\n" + dark_css
 
+    def to_dict(self):
```
abidlabs, line 49: I think we should add some unit tests for these new methods?

CHANGELOG.md
```diff
@@ -2,7 +2,69 @@
...
+
+This creates a space on the huggingface hub to host the theme files and show potential users a preview of your theme.
+
+An example theme space is here: https://huggingface.co/spaces/freddyaboulton/dracula_revamped 
+
+### Downloading
+To use a theme from the hub, use the `from_hub` method on the `ThemeClass` and pass it to your app:
+
+```python
+my_theme = gr.Themes.Base.from_hub("freddyaboulton/my_theme")
```
abidlabs, line 49:
```suggestion
my_theme = gr.themes.ThemeClass.from_hub("freddyaboulton/my_theme")
```

abidlabs:
Fantastic PR @freddyaboulton! Everything works really well

I had some small suggestions in order to make the process of uploading / downloading themes easier:

* When uploading, would it be possible to avoid having to provide a version? Like perhaps it can automatically just increment the patch version from the latest existing version, starting from 0.0.1? Otherwise, it forces people to think about semver which is just a mental obstacle imo. The fewer required parameters the better!
* When uploading, we should raise an error if someone tries to override a theme with an existing semver (of course, they can still do it manually, but we shouldn't let them do it easily through the class) 
* The required syntax for downloading a theme is fairly long: `gr.themes.ThemeClass.from_hub("theme_name")`. Perhaps we can import `ThemeClass` as `Theme` in `__init__.py` so that we can just to `gr.Theme.from_hub()`

Otherwise, LGTM!

CHANGELOG.md
freddyaboulton: I think it's better to use `Base` as that has the `set` method defined on it.

gradio/themes/base.py
freddyaboulton: I think that would be more confusing for users using `ThemeClass.from_hub` actually. I see your point though.

How about this, `from_hub` should raise an error if the theme can't be found but when passing a string-valued theme to `Blocks` or `Interface`, this error should be caught and the `DefaultTheme` should be used instead?

freddyaboulton: I think that's a happy medium. App authors who just want to load a theme are likely to use the `Blocks(theme="abidlabs/....")` syntax. Theme creators who are loading a theme and modifying it will use `from_hub` method instead.

abidlabs: Yeah sounds good!

gradio/themes/app.py
```diff
@@ -0,0 +1,139 @@
+import time
+
+import gradio as gr
+
+with gr.Blocks(theme=gr.themes.Default()) as demo:
+    gr.Markdown(
+        """
+    # {THEME} Preview
```
abidlabs, line 8:
```suggestion
    # Preview: `{THEME}`
```

gradio/themes/base.py
freddyaboulton: Done!

abidlabs:
I think `version` is still a required parameter in `push_to_hub()` and `upload_theme()`. It should be optional now, right?

abidlabs:
Yes, to confirm, if the parameter `version` is set to `None`, then I don't get an error. But if it is not provided, I get the following error:

```
TypeError: push_to_hub() missing 1 required positional argument: 'version'
```

A few other small DX pain points I noticed. These are all pretty small and can be addressed later on, but noting them here:

(1) When I tried pushing to the Hub, I naturally used the name of the Space until I realized that I needed to remove my username:

```py
theme.push_to_hub("abidlabs/pakistan",...) # not valid
theme.push_to_hub("pakistan",...) # valid``` 
```
I wonder if we can accept the username if it is provided

(2) Do we need to keep providing the HF_TOKEN? I wonder if we can support logging in via hf cli, similar to this: https://huggingface.co/docs/transformers/model_sharing#setup

(3) The theme preview Space does not show the version number, which makes it not obvious to know what version to pin if I want the displayed theme. I think we include the version at the top next to the theme name 

image

freddyaboulton:
@abidlabs Thanks for the comments! 

Will fix the issue about the version not being optional!

Good point about the token. For logging in via cli, what I think we can do is make the token optional as well, and if the user is already logged in via cli, we will not ask users for the token. If the user is not logged in and no token was provided, we raise error asking them to either log in or provide a token.

I can display the version number in the theme preview space but my idea was to have users not worry about the version number and instead use the latest via `abidlabs/pakistan` as opposed to `abidlabs/pakistan@<version>`. In the future it would be cool to toggle between the different versions in the theme preview space so that it's clearer what each of the different versions do. I'll see if I can do that real quick.

abidlabs:
@freddyaboulton thanks for addressing the suggestions. Now that we've released `gradio==3.22.0`, I'm going to go ahead and merge this in so that we can all easily build and share themes while working from `main`.

PR #3430, opened 2023-03-10
consecutive triggers with `.then`

aliabid94:
Bringing back `.then`! 

Syntax:
```python
btn.click(fn, inputs, outputs).then(fn2, inputs2, outputs2) # executed regardless of success of fn
btn.click(fn, inputs, outputs).success(fn2, inputs2, outputs2) # executed only if fn is successful
```

Closes: #2560


pngwn:
Code looks good to me!

abidlabs:
Clean syntax. Is this ready to be reviewed?

guides/03_building-with-blocks/01_blocks-and-event-listeners.md
```diff
@@ -135,3 +123,23 @@ $demo_blocks_essay_update
...
 
+## Running Events Consecutively
+
+You can also run events consecutively by using the `then` method of an event listener. This will run an event after the previous event has finished running. This is useful for running events that update components in multiple steps. 
+
+For example, in the chatbot example below, we first update the chatbot with the user message immediately, and then update the chatbot with the computer response after a simulated delay.
+
+$code_chatbot_simple_demo
+$demo_chatbot_simple_demo
+
```
abidlabs, line 134:
```suggestion
The `.then()` method of an event listener executes the subsequent event regardless of whether the previous event raised any errors. If you'd like to  only run subsequent events if the previous event executed successfully, use the `.success()` method, which takes the same arguments as `.then()`.
```

gradio/blocks.py
```diff
@@ -158,7 +160,10 @@ def set_event_trigger(
             batch: whether this function takes in a batch of inputs
             max_batch_size: the maximum batch size to send to the function
             cancels: a list of other events to cancel when this event is triggered. For example, setting cancels=[click_event] will cancel the click_event, where click_event is the return value of another components .click method.
-        Returns: None
+            every: Run this event 'every' number of seconds while the client connection is open. Interpreted in seconds. Queue must be enabled.
+            triggered_after: if set, this event will be triggered after 'triggered_after' function index
+            trigger_only_on_success: if True, this event will only be triggered if the previous event was successful
```
abidlabs, line 165:
```suggestion
            trigger_only_on_success: if True, this event will only be triggered if the previous event was successful (only applies if `triggered_after` is set)
```

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## New Features:
 
+- Allow consecutive function triggers with `.then` and `.success` by [@aliabid94](https://github.com/aliabid94) in [PR 3430](https://github.com/gradio-app/gradio/pull/3430)
```
abidlabs, line 5: Would be good to include a code example in the changelog since this is a pretty big feature

abidlabs:
Very nice PR @aliabid94! I tested `.then()` in different setups and it's very fun to play around with:    

* **Chatbot example** ✅

Works beautifully! Now let's bring back bubbles on one side so that we get the full messaging experience (just kidding)

* **Examples from the issue** ✅

```py
import time
import gradio as gr

def fast_event(x):
    if not x:
        raise gr.Error("A Protein must be selected")
    return x

def slow_event(x):
    time.sleep(2)
    return (x + " ")*100

with gr.Blocks() as demo:
    dropdown = gr.Dropdown(["Protein A", "Protein B", "Protein C"])
    textbox = gr.Textbox()
    sequence = gr.Textbox()
    dropdown.change(fast_event, dropdown, textbox).success(slow_event, textbox, sequence)
    textbox.submit(slow_event, textbox, sequence)

demo.launch()
```

* **Chained events having different kinds of queue** ✅

* Tried same demo as before but with different settings for `queue` parameter. Works as far as I can tell! 

* **With canceling**: see below

A -> B -> C

Canceling A while A is running does cancel A, B, and C as expected, nice!

_However_, canceling B while A is running does not cancel B and C. 

```py
import time
import gradio as gr


def slow_event(x):
    time.sleep(2)
    return x

with gr.Blocks() as demo:
    textbox1 = gr.Textbox()
    textbox2 = gr.Textbox()
    textbox3 = gr.Textbox()
    textbox4 = gr.Textbox()
    btn = gr.Button("Cancel", variant="stop")
    
    ev1 = textbox1.submit(slow_event, textbox1, textbox2)
    ev2 = ev1.then(slow_event, textbox2, textbox3)
    ev3 = ev2.then(slow_event, textbox3, textbox4)
    btn.click(None, None, None, cancels=ev2)
    

demo.queue().launch()
```

Not sure if this is what users will expect. But I think it's okay because a user can always replace `cancels=ev2` with `cancels=[ev1, ev2]

* **With iterators**: see below

A (iterator) -> B (regular function)

`.then()` only executes after the iterator is completely finished. That's fine, but perhaps we document this behavior somewhere? 

Works great with `cancels`

A (regular function) -> B (iterator)

Works great!

* **With Blocks inside another Blocks** ❌

Doesn't work if there's a `Blocks` that's nested inside another Blocks. Here's an example:

```py
import time
import gradio as gr

def iterator(x):
    for i in range(len(x)):
        time.sleep(1)
        yield x[:i+1]

with gr.Blocks() as demo:
    textbox1 = gr.Textbox()
    textbox2 = gr.Textbox()
    textbox3 = gr.Textbox()
    btn = gr.Button("Cancel", variant="stop")
    
    ev1 = textbox1.submit(iterator, textbox1, textbox2)
    ev2 = ev1.then(lambda x:x, textbox2, textbox3)

with gr.Blocks() as parent:
    with gr.Tab("A"):
        gr.Interface(lambda x:x, "textbox", "textbox")    
    with gr.Tab("B"):
        demo.render()
        
#demo.queue().launch()
parent.queue().launch()
```

Launching `demo` by itself works, but launching `parent` does not. The reason is that the dependency indices change when a Blocks is loaded inside of another Blocks, so this has to be handled. We do the same thing for the `cancel` parameter: https://github.com/gradio-app/gradio/blob/11bb732f5d12284aad1a139fd3891bd01ff1167d/gradio/blocks.py#L686

**Other notes**

Ideally, we test this with an end-to-end test, but at least we should write some backend tests to make sure that the dependencies are being set up correctly.

demo/chatbot_simple_demo/run.py
```diff
@@ -1,17 +1,25 @@
...
+    def user_message(message, history):
+        return history + [[message, None]]
+
+    def bot_message(history):
+        response = random.choice(["Yes", "No"])
+        history[-1][1] = response
+        time.sleep(1)
+        return history
+
+    msg.submit(user_message, [msg, chatbot], chatbot).then(
```
abidlabs, line 19: Love it! Let's update the other chatbot demos to use this syntax as well?

abidlabs, line 19:
```suggestion
    msg.submit(user_message, [msg, chatbot], chatbot, queue=False).then(
```

demo/chatbot_simple_demo/run.py
```diff
@@ -1,17 +1,25 @@
...
+import time
 
 with gr.Blocks() as demo:
     chatbot = gr.Chatbot()
     msg = gr.Textbox()
     clear = gr.Button("Clear")
 
-    msg.submit(respond, [chatbot, msg], chatbot)
+    def user_message(message, history):
+        return history + [[message, None]]
```
abidlabs, line 11: Let's make the textbox clear on submit as well?
```suggestion
    def user_message(message, history):
        return "", history + [[message, None]]
```

demo/chatbot_simple_demo/run.py
aliabid94: Done.

aliabid94: updated demo/chatbot_multimodal, demo/chatbot was too complicated

aliabid94:
> With Blocks inside another Blocks ❌

Fixed.

> at least we should write some backend tests to make sure that the dependencies are being set up correctly, including in the nested Blocks case.

Done.

abidlabs:
LGTM @aliabid94! Great PR

abidlabs (approved):

PR #3431, opened 2023-03-10
Site slugs for new website

pngwn:
@aliabd 

I did this work in a PR to your branch just so you could easily see what I'm doing here. Feel free to adapt the code in here, copy the changes across however you like. You can close this PR whenever you want to.

Don't know if everyone is familiar with the term 'slug' but it just means then end of the url, after the domain. It needs to be a valid url. 

So the first case is that we have some data, rather than markdown. In this case we can simply pass the most appropriate string to the `slugifier`, i grabbed one from npm. For `fn` names i used a combination of the class name + the function name so we get slugs like `#interface-load`. The slugifier also handles collisions, so if we ever have two links with the same name on the same page then it will generate a different url (it just adds a number). I generate these slugs and store them as additional properties on the docs json object. under `.slug`. You can see this better in the code.

For markdown it is moire complicated but the propcess is essentially the same. The only difference is I had to write it as a plugin. We generate the slugs and modify the generated html to add the necessary info + elements. We _also_ store a list of the headings + links for each section so we can build a navigation. The plugin code is what it is, we can't control that plugin API but it is relatively easy to understand and is short at least.

lmk if you have any questions

### DIFF
--- a/ui/packages/_website/package.json
+++ b/ui/packages/_website/package.json
@@ -25,6 +25,7 @@
 	"type": "module",
 	"dependencies": {
 		"@sindresorhus/slugify": "^2.2.0",
+		"hast-util-to-string": "^2.0.0",
 		"mdsvex": "^0.10.6",
 		"postcss": ">=8.3.3 <9.0.0"
 	}
--- a/ui/packages/_website/package.json
+++ b/ui/packages/_website/package.json
@@ -24,6 +24,7 @@
 	},
 	"type": "module",
 	"dependencies": {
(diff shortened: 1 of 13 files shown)


aliabd:
Thanks @pngwn!! This is exactly what I needed for guides and docs. I think you missed flagging in docs so I will add that. And I will also add the slugs to the docs navigation (as subheaders) as well as add anchor tags for them etc. Basically just like gudies. 

But this is awesome!

PR #3450, opened 2023-03-13
Adds `accepts` parameter to `Code` component to control filepath/string

abidlabs:
Fixes: #3436 and adds a test

Not adding an entry to changelog since we haven't released the `Code` component yet


gradio/components.py
```diff
@@ -5428,7 +5430,7 @@ def get_config(self):
         }
 
     def postprocess(self, y):
-        if y is not None and os.path.isfile(y):
+        if y is not None and self.accepts == "filepath":
```
freddyaboulton, line 5433: The one done side of attaching this to `self` is that it can't be updated at runtime.

What if we adopt a similar approach to the chatbot where the return type dictates whether it should be treated as a file?

In this case, something like `pathlib` file could work or we could introduce our own light wrapper type for maximum flexibility (to support urls for example).

abidlabs: Good point @freddyaboulton. A few options:
* We do something similar to the `Chatbot` component where if you pass in a tuple `(filepath, )`, then it is treated as a filepath. In the case of media files, the second element is the alt-text, here it could be the language
* We add our own wrapper, but it adds significant complexity just to serve a file / URL
* We leave it as is for now, but down the line, we have the ability to `gr.update()` backend processing parameters as well. We need this for various other issues anyways so it would not be adding additional work

I'm inclined between (1) and (3) myself, what do you think?

freddyaboulton: Thanks for the breakdown!

Over 1 and 3, I prefer 1. 

For 4.0, I think it may be worthwhile to introduce our own lightweight classes (even if just for internal use) using dataclasses or pydantic models for the messaging between frontend/backend/user-defined-functions. The built-in classes like tuple and dict are easy to type but a bit harder to reason about and use due to lack of typehints. Definitely a separate issue from this PR though!

abidlabs: Sg will make the change

aliabid94: I don't know if it makes sense for the second argument in the tuple to be the language, because you might also want to be able to specify the language for the text return as well (actually makes more sense there, since you can infer language from a filepath but not a chunk of text). Besides that, I don't have a strong opinion between 1 + 3

abidlabs: Hmm good point. You can always update the language with `gr.update(value=value, language=language)` so I agree, not much value in having language as the second parameter

abidlabs:
Have implemented the changes that we discussed @freddyaboulton @aliabid94, should be ready for a quick re-review!

freddyaboulton (approved):
Nice svelte @abidlabs !

This looks good to me! Tested and is working well

Issue #3445, opened 2023-03-12
Json component doesn't respect `show_label`


braisedpork95:
### Describe the bug

The label covers part of the output content. Neither `show_label` nor `container` flag works.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```python
import gradio as gr

gr.Interface(
    lambda name: {'name': name},
    'text',
    gr.Json(label="json", show_label=False).style(container=True),
    allow_flagging="never",
).launch()
```

### Screenshot

<img width="1252" alt="image" src="https://user-images.githubusercontent.com/46232992/224528045-2cc5ea36-795e-4048-9359-60062f0781aa.png">

### Logs

```shell
No response
```

### System Info

```shell
gradio==3.20.1
```

### Severity

annoying

abidlabs:
@braisedpork95 we've fixed this issue in the latest version of `gradio` ([instructions for using here](https://gradio.app/docs/main/)), and we'll likely release to pypi early next week. Here's how it looks:

<img width="585" alt="image" src="https://user-images.githubusercontent.com/1778297/224563320-576eec39-5c94-41a5-89d5-fe317dbe5783.png">

However, it seems like it still does not respect the `show_label` parameter so I've renamed the issue to reflect that

PR #3451, opened 2023-03-13
Makes a few fixes to the JSON component

abidlabs:
Noticed a few small issues with the JSON component, this PR:

* Respects the `show_label` parameter for JSON component
* Removes the `interactive` parameter as that is not supported by the JSON component
* Uses the copy and checkmark icons instead of the words "copy to clipboard" and "copied"
 
Fixes: #3445

### DIFF
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -3589,7 +3589,7 @@ class JSON(Changeable, IOComponent, JSONSerializable):
     """
     Used to display arbitrary JSON output prettily.
     Preprocessing: this component does *not* accept input.
-    Postprocessing: expects a valid JSON {str} -- or a {list} or {dict} that is JSON serializable.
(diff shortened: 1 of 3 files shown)


gradio/components.py
```diff
@@ -3637,7 +3637,6 @@ def update(
         label: str | None = None,
         show_label: bool | None = None,
         visible: bool | None = None,
-        interactive: bool | None = None,
```
freddyaboulton, line 3640: I guess this never worked so it's not a breaking change 😂 ?

freddyaboulton (approved):
Tested and works great @abidlabs !

gradio/components.py
abidlabs: I suppose there is a *tiny* possibility someone wrote code with update(interactive=something) but I think that is outweighed by the benefit of having accurate type hints

abidlabs:
Thanks for the quick review @freddyaboulton, much appreciated!

PR #3452, opened 2023-03-13
Adds copy button to textbox via `.style()`

### LINKED ISSUE #2585: Add copy to clipboard button to other input types

The only input type that shows the copy to clipboard button is JSON. 

It would be great if users can optionally show this button for other input types such as gr.TextBox as well.

abidlabs:
Wanted to do some more touching up of different components, so I tackled long-standing feature request: adding a copy button to the `gr.Textbox` component. 

I think we can close #2585 because although we considered adding this button to other component types, the only one that has been requested by the community (and it's been requested often) is adding it to the `gr.Textbox` component.

To test: 

```py
import gradio as gr

io = gr.Interface(lambda x:x, ["text", "image"], [gr.Textbox(), gr.Textbox().style(show_copy_button=True)])
io.launch()
```


abidlabs:
Keeping it as a draft for now. Will add a parameter to `.style()` to control whether or not the clipboard is displayed since presumably we don't want this changing the UI of every existing interface out there with a `gr.Textbox`

freddyaboulton (approved):
Thanks @abidlabs ! LGTM

abidlabs:

g-h-chen:
> 🎉 The demo notebooks match the run.py files! 🎉

@gradio-pr-bot I think the copy button does not function as it should in my case.
Here is my code:
`
hint_textbox = gr.Textbox(
                label="label",
                visible=True,
                show_label=True,
                placeholder='placeholder',
                interactive=True,
                show_copy_button=True,
            ).style(show_copy_button=True, container=False)
`

LeopoldACC:
> > 🎉 The demo notebooks match the run.py files! 🎉
> 
> @gradio-pr-bot I think the copy button does not function as it should in my case. Here is my code: `hint_textbox = gr.Textbox( label="label", visible=True, show_label=True, placeholder='placeholder', interactive=True, show_copy_button=True, ).style(show_copy_button=True, container=False)`
> 
> I tried both 3.23.0 and 3.35. The copy button shows up, but the content is not copied after I click on the copy button.

hi，do you fix the bug, I meet the same error as you. After I click the copy button, the content in textbox do not copy into clipboard

sonnguyen-TS:
Hi, I have the same issue. I had set `show_copy_button=True, show_label=True` for the Textbox and the copy button work well in my local laptop, but when I set **server_name** `demo.queue().launch(share=False, server_name=<my-server-ip>)` then the copy button doesn't work anymore. The copy button is still there but there is nothing happens when it's clicked.

abidlabs:
Interesting, could you open a new issue @sonnguyen-TS with a full repro?

sonnguyen-TS:
@abidlabs Thanks for your reply, I open a new issue [#6695](https://github.com/gradio-app/gradio/issues/6695) for easy to track.

Issue #3454, opened 2023-03-13
The `gr.Request` object doesn't handle dictionaries if nested keys cannot be converted to variable names


abidlabs:
### Describe the bug

As discovered by @radames, the `gr.Request` object doesn't correctly handle the `headers` dict and other dictionary attributes if they contain keys whose names include hyphens or other characters that are not allowed in variable names. Easiest to illustrate with an example:

```py
import gradio as gr

def predict(text, request: gr.Request):
    ip = request.headers
    print(ip)
    ip = request.headers['user-agent']
    return ip
    
gr.Interface(predict, "textbox", "textbox").queue().launch()
```

As can be seen in the print statement, `request.headers` definitely includes a key named `user-agent` but because it contains a `-`, it cannot be accessed correctly. Only happens with queuing enabled.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

See above

### Logs

```shell
Traceback (most recent call last):
  File "c:\users\islam\dev\gradio-repos\gradio\gradio\routes.py", line 385, in run_predict
    output = await app.get_blocks().process_api(
  File "c:\users\islam\dev\gradio-repos\gradio\gradio\blocks.py", line 1037, in process_api
    result = await self.call_function(
  File "c:\users\islam\dev\gradio-repos\gradio\gradio\blocks.py", line 849, in call_function
    prediction = await anyio.to_thread.run_sync(
  File "c:\users\islam\appdata\local\programs\python\python39\lib\site-packages\anyio\to_thread.py", line 28, in run_sync
    return await get_asynclib().run_sync_in_worker_thread(func, *args, cancellable=cancellable,
  File "c:\users\islam\appdata\local\programs\python\python39\lib\site-packages\anyio\_backends\_asyncio.py", line 818, in run_sync_in_worker_thread
    return await future
  File "c:\users\islam\appdata\local\programs\python\python39\lib\site-packages\anyio\_backends\_asyncio.py", line 754, in run
    result = context.run(func, *args)
  File "<ipython-input-13-cbe80223059c>", line 6, in predict
    ip = request.headers['user-agent']
TypeError: 'Obj' object is not subscriptable
```

### System Info

```shell
3.20.1
```

### Severity

serious, but I can work around it

abidlabs:
Closed via #3459

PR #3459, opened 2023-03-14
implement missing methods handle nested dict

radames:
Quick fix for #3454 

with this extra methods on `class Obj` we can now do

```python

request.headers["user-agent"]

"user-agent" in request.headers

dict(request.headers)

for key, value in request.headers:
    print(key, value)

```
still missing `.keys()`, `.values()` `.items()`


freddyaboulton (commented):
Looks good to me @radames ! Thank you for writing the tests as well.

Supporting the missing attributes you mention should be straightforward, no?

```python
    def keys(self):
        return self.__dict__.keys()

    def values(self):
        return self.__dict__.values()

    def items(self):
        return self.__dict__.items()
```

Should we do that in this PR?

abidlabs:
LGTM @radames! Tested and confirmed that accessing via attributes continues to work, while adding support for accessing via keys. As @freddyaboulton suggested, let's add the missing methods (and credit you in the changelog) and we should be good to merge

radames:
thanks @abidlabs and @freddyaboulton , update with the missing attributes and add extra asserts

abidlabs:
Thank you so much @radames!

Issue #3328, opened 2023-02-27
File.upload returns an empty file in 3.19.1


oobabooga:
### Describe the bug

If I create a `gr.File` object with `type='binary'` and attach a `upload` event to it, the returned file is empty if I upload a .zip file.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```python
import gradio as gr
import time

def f(file):
    print(file)

with gr.Blocks() as interface:
    x = gr.File(type="binary")
    x.upload(f, x, None)

interface.queue()
interface.launch(debug=True)
```

Launch this interface and try to upload a .zip file. The following will be printed:

```
b''
```

Meaning that the temporary file created by gradio is empty.

### Logs

```shell
No errors.
```

### System Info

```shell
gradio==3.19.1
Linux
```

### Severity

blocking upgrade to latest gradio version

abidlabs:
@aliabid94 maybe this broke as part of the large file issue?

fkhan0520:
Hey @oobabooga I wasn't able to reproduce with the provided code on gradio==3.19.1. Are you able to share the zip file this fails with?

oobabooga:
Sure, here is an example of .zip file that generates `b''` for the sample code above:

https://files.catbox.moe/as41my.zip

abidlabs:
Was able to reproduce, thanks @oobabooga -- working on a fix

PR #3462, opened 2023-03-14
Fix `File.upload()` event

abidlabs:
Fixes the `File.upload()` event trigger which broke as part of the change in how we uploaded files. Thanks @aliabid94 for the help with svelte

Test with code similar to what was posted in the original issue:

```py
import gradio as gr
import time

def f(file):
    print(file[:100])

with gr.Blocks() as interface:
    x = gr.File(type="binary")
    x.upload(f, x, None)

interface.queue()
interface.launch()
```

or with this code:

```py
import gradio as gr

with gr.Blocks() as interface:
    file1 = gr.File()
    file2 = gr.File()
    file1.upload(lambda x:x.name, file1, file2)

interface.launch()
```

The `UploadButton` component was already working:

```py
import gradio as gr

with gr.Blocks() as interface:
    file1 = gr.UploadButton()
    file2 = gr.File()
    file1.upload(lambda x:x.name, file1, file2)

interface.launch()
```

Fixes: #3328


freddyaboulton (approved):
Tested and works! Thanks @abidlabs !

Want to add a svelte unit test while you're at it? 😄

abidlabs:
Thanks @freddyaboulton! Tbh I've had enough Svelte for now, back to Python. But if anyone would like to add a test, feel free to -- I'll keep this PR open for a few hours

Issue #2962, opened 2023-01-09
Add ability to append additional classes to html element when creating a component


Gerschel:
- [✔️  ] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  

Styling a component, or targeting with javascript. Element ID's should be unique, but I can target groups
that have the same (html) class. I don't have access to append custom classes when creating the component.  
This means, that I have to write a rule for each element id, instead of a single rule for those that match by class.

**Describe the solution you'd like**  
Similar to `elem_id`, but appends to class list that's autogenerated
```python
gr.Component(elem_class = "my-class")
```
Or maybe
```python
gr.Component(elem_class = ["my-first-class", "my-second-class"])
```

aliabid94:
I'll take this

PR #3466, opened 2023-03-14
Add elem classes

aliabid94:
Add classes to components with `elem_classes` just like `elem_id`
Fixes: #2962


abidlabs:
One thing I noticed is that when you are applying css to classes, you have to use `!important` to override the default css, whereas you don't have to do this for ids. Here's example:

```py
import gradio as gr

css = """
.abc {
    background-color: green;
}

#hi {
    background-color: red;
}
"""

gr.Interface(
    lambda x:x, 
    [
        gr.Textbox(elem_classes="abc"),
        gr.Textbox(elem_id="hi"),
        gr.Textbox(),
    ],        
    gr.Image(elem_classes=["abc"]),
    css=css,
).launch()
```

Produces:

image


Whereas if you add the `!important`, it works:

```py
import gradio as gr

css = """
.abc {
    background-color: green !important;
}

#hi {
    background-color: red;
}
"""

gr.Interface(
    lambda x:x, 
    [
        gr.Textbox(elem_classes="abc"),
        gr.Textbox(elem_id="hi"),
        gr.Textbox(),
    ],        
    gr.Image(elem_classes=["abc"]),
    css=css,
).launch()
```

image

Is this expected?

abidlabs:
The three Plot components are missing the docstring for `elem_classes` although they include the parameter: 
* BarPlot
* LinePlot
* ScatterPlot

abidlabs:
Other than those two points, LGTM @aliabid94!

abidlabs (approved):

aliabid94:
ok I updated the docstrings.

Unfortunately I cannot think of a good way to have class rules overwrite the default css without `!important`. The reason is because the default CSS rules are attached by combing the css rules in svelte with a randomly generated class identifier at runtime, to keep the svelte css scoped. See below, the random string class attached to each element and ruleset are generated at compile time.
Screen Shot 2023-03-14 at 8 01 27 PM

because these default rules are selected by their default specifier AND a random class name, they overrule the CSS rules with just a class specifier - in CSS, the more specific rule always wins.

If you have ideas lmk, otherwise we can document that class rules should have `!important`. Using `!important` is not good practice, but neither is using `css=` lol

abidlabs:
Thanks for the explanation @aliabid94. I think that's okay, it won't be necessary for every css property, just the ones that conflict with the built-ins. Let's definitely document this in the Guide & I'll comm to the Automatic1111 folks as well.

abidlabs:
Not to block merging, but let's tag @pngwn for his thoughts on this when he's back.

PR #3465, opened 2023-03-14
Document gr.Error 

aliabd:
Adds docstrings for gr.Error and document it under helpers in the website. 

Screen Shot 2023-03-14 at 1 35 09 PM

Also fixes a broken link in the guides.

PR #3468, opened 2023-03-15
Fix primary bg button color

aliabid94:
Brighten button primary color in dark mode

Screen Shot 2023-03-14 at 8 51 43 PM

Fixes: #3465


abidlabs (approved):
Looks very nice in the deployed PR Space! Thanks @aliabid94

Issue #3438, opened 2023-03-10
Refactor `components.py` to simplify the creation of new components


abidlabs:
Ahead of custom components, it would be good to simplify the process of creating new component classes. Many opportunities for simplification:

- [x]  Remove `IOComponent.add_interactive_to_config` from `.update()`
- [x]  On that note, can we remove `.update()` altogether since its almost a duplicate of `get_config()`?
- [x]  On that note, can we remove `get_config()` since it basically just returns all of the constructor parameters. Should be a simpler way to do this
- [x]  Can we get rid of `generate_sample()`
- [x]  Should we remove interpretation (!!!)? It seems like almost no one uses it and causes significant complexity in the code base
- [x]  Make it easier for users to figure out which serialization class to mix in
- [x]  Should we break up each component into separate python file? So it would mirror the structure in the frontend
- [x]  Why do we have `Component` and `IOComponent` as separate classes -- does this distinction still make sense?
- [ ]  Clarify what it means to be used as an input component vs. output component. Technically any component, like the Chatbot, can also be used as an input component
- [x]  Having to explicitly call the parent methods `IOComponent.` is not ideal
- [x]  Can we get rid of `.style()`, and make every parameter part of the constructor? That way we don't have to worry about updating style parameters separately too
- [x] Simplify tempfile creation

Definitely a 4.0 thing

space-nuko:
> Why do we have Component and IOComponent as separate classes -- does this distinction still make sense?

I have actually run into a case where this distinction is important, I needed to pass a dynamic list of components into an input field, but I wasn't able to tell which ones were valid for use with inputs. Without being able to do `isinstance(component, IOComponent)` I would have gotten the error "Component is not a valid input component"

If these classes are merged then there should still be a way to programmatically tell if a component can be used for input/output

abidlabs:
Thanks @space-nuko, that's a good point. Would you be able to share a code snippet or link to the Space? It would be helpful to have some more context as we think about this point.

space-nuko:
Actual code where I needed to this was

https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/a441b6fb1d5376cc946f4c25f5229284a577cd0c/modules/generation_parameters_copypaste.py#L142

freddyaboulton:
Most have been checked off! Closing!!

PR #3476, opened 2023-03-15
Remove `IOComponent.add_interactive_to_config`

space-nuko:
Related: #3438


abidlabs:
Hi @space-nuko, thanks for making this contribution. I think the best way to make this fix is actually to treat `interactive as any other parameter in `.update()` -- so instead of:

```py
"mode": "dynamic" if interactive else "static",
```

we change it to:

```py
"interactive": interactive
```

And then in the `postprocess_update_dict()` in `blocks.py`, we convert the "interactive" key to the "mode" key. That way, we avoid a lot of the code repetition. 

Would you be able to make that change?

Also, thanks for updating the Guide. Are you interested in contributing a component? We're working on revamping that whole process to make it much better.

abidlabs:
Thanks for making the update @space-nuko! I believe the tests are failing because we don't leave the key `interactive` in the config dictionary. I've suggested a fix above. Once this is fixed, and you've added an entry to `CHANGELOG.md` so that we can credit you for the fix, we should be good to merge.

abidlabs:
Thanks for getting the tests to work out @space-nuko, LGTM! We’ll merge in later today or tomorrow before we release the next  version of Gradio to Pypi

abidlabs (approved):

space-nuko:
Yeah np, sorry about that it's difficult for me to run the tests on my box right now with the dependency issue

abidlabs:
There was just one issue with the line endings. Apparently `test/test_blocks.py` had been committed with windows-style endings, which meant that the GitHub diff showed every line as changed. I've converted that back to Linux style line endings (you can configure that locally for yourself by running `git config --global core.autocrlf true` [as described here](https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings))

abidlabs:
Merged. Thanks again @space-nuko for the PR!

PR #3487, opened 2023-03-16
Update test requirements and script

### LINKED ISSUE #3484: `mlflow` dependency prevents test requirements from installing on Windows
### Describe the bug

I want to run the `pytest` tests on Windows, but I can't install the requirements:

```
ERROR: Could not find a version that satisfies the requirement pywin32==227 (from docker==5.0.3->-r test/requirements.txt (line 65)) (from versions: 302, 303, 304, 305)
```

Long story short, this is caused by an old version of the `docker` package pinning a version of `pywin32` that isn't compatible with Python 3.10 (and has security vulnerabilities!): https://github.com/docker/docker-py/issues/2902

...and this version of docker is added as a transitive dependency by `mlflow`:

https://github.com/gradio-app/gradio/blob/19eaafb2adf1e9de49d172cc1c4695f977b691b0/test/requirements.txt#L65-L66

...and `mlflow` later updated their dependencies to relax the `docker` version requirement but `mlflow` hasn't been updated in gradio yet: https://github.com/mlflow/mlflow/issues/6562

So I think `mlflow` in the test requirements should be updated.

### Reproduction

`pip install -r test/requirements.txt` on Windows

### Screenshot

_No response_

### Logs

```shell
N/A
```

### System Info

```shell
19eaafb2, Windows Server 2022, Chromium
```

### Severity

blocking all usage of gradio

space-nuko:
Closes #3484


space-nuko:
```
ERROR: Ignored the following versions that require a different python version: 1.0.7 Requires-Python >=3.8
ERROR: Could not find a version that satisfies the requirement contourpy==1.0.7 (from versions: 0.0.1, 0.0.2, 0.0.3, 0.0.4, 0.0.5, 1.0.0, 1.0.1, 1.0.2, 1.0.3, 1.0.4, 1.0.5, 1.0.6)
```

Should we pin dependencies in `test/requirements.in` for 3.7 tests?

freddyaboulton:
Hi @space-nuko ! Thanks for your patience on this. 

I agree we should deprecate support for 3.7 but not in this PR. I think what we can do is have separate requirement files for 3.7 and 3.9 until we deprecate 3.7.

freddyaboulton:
Actually, the other thing we can do is just remove `mlflow` as a dependency. If you remove `mlflow` from `requirements.in` are you able to generate all the dependencies properly?

space-nuko:
The errors are in CI, not on my box, since I use Python 3.10

And correct me if I'm wrong but is `contourpy` a dependency of `matplotlib`? I don't have errors related to `mlflow` anymore with the regenerated deps

freddyaboulton:
Thanks for the patience @space-nuko ! 

If you pull in latest main, you will see we have separate dependencies for 3.7 and 3.9. You should now be able to update the 3.9 dependencies without breaking the python 3.7 build.

abidlabs:
Hi @space-nuko would you like to update the 3.9 dependencies? (And use that environment for testing locally)

space-nuko:
Just getting around to it actually, thanks for reminding me

space-nuko:
Now fails because of https://github.com/scikit-image/scikit-image/pull/6567

abidlabs:
Honestly, I'd be fine with getting rid of the test that tests the `mlflow` integration. It's heavily mocked up so it's not testing much anyways.

Let me open a PR

space-nuko:
I don't think it's just `mlflow` in this case, it's also a `scikit-image` integration that enables interpretation for `gr.Image`

abidlabs:
Hmm interesting.. okay taking another look at this

abidlabs:
Hi @space-nuko, I did some cleanup in a separate PR. Closing this one, I'd appreciate if you can take a look at #3608 and confirms that it works for you

PR #3489, opened 2023-03-17
Release v3.22.0

abidlabs:


gradio/themes/base.py
```diff
@@ -452,7 +452,7 @@ def set(
         button_small_text_size=None,
         button_small_text_weight=None,
         button_transition=None,
-    ):
+    ) -> Base:
```
freddyaboulton, line 455: Added this typehint for better IDE completion on `push_to_hub`, `dump` methods after `set`

abidlabs: Thanks @freddyaboulton!

PR #3497, opened 2023-03-17
More theme updates

aliabid94:
Added doc strings, renamed some vars, removed unused ones.


gradio/themes/base.py
```diff
@@ -224,65 +224,65 @@ def expand_shortcut(shortcut, mode="color", prefix=None):
     def set(
```
freddyaboulton, line 224: We should add `font` and `font_mono` as arguments to `set` so that people can load a font from the hub and then be able to change the font

gradio/blocks.py
```diff
@@ -1105,7 +1105,6 @@ def get_config_file(self):
             "show_api": self.show_api,
             "is_colab": utils.colab_check(),
             "stylesheets": self.stylesheets,
-            "root": self.root,
```
freddyaboulton, line 1108: Did you mean to get rid of this?

CHANGELOG.md
```diff
@@ -5,7 +5,7 @@
 
 ### Official Theme release
 
-Gradio now supports a new theme system, which allows you to customize the look and feel of your app. You can now use the `theme=` kwarg to pass in a prebuilt theme, or customize your own! See guides/06_other-tutorials/theming-guide for more details. By [@aliabid94](https://github.com/aliabid94) in [PR 3494](https://github.com/gradio-app/gradio/pull/3494) 
+Gradio now supports a new theme system, which allows you to customize the look and feel of your app. You can now use the `theme=` kwarg to pass in a prebuilt theme, or customize your own! See guides/06_other-tutorials/theming-guide for more details. By [@aliabid94](https://github.com/aliabid94) in [PR 3470](https://github.com/gradio-app/gradio/pull/3470) and [PR 3497](https://github.com/gradio-app/gradio/pull/3497)
```
freddyaboulton, line 8: I think `guides/06_other-tutorials/theming-guide` won't actually link to the guide:

image

gradio/blocks.py
aliabid94: oh whoops, did not resolve merge conflict correctly

CHANGELOG.md
aliabid94: I actually hadn't meant for it to be a link, just a reference within the repo where the guide is. I could link to http://gradio.app/theming-guide instead?

gradio/themes/base.py
```diff
@@ -224,65 +224,65 @@ def expand_shortcut(shortcut, mode="color", prefix=None):
...
         self,
         *,
+        # Body
```
freddyaboulton, line 227: BTW I think the refactoring broke `blocks_kitchen_sink` and my dracula theme. Get the same error in both.
```bash
❯ python demo/blocks_kitchen_sink/run.py
  File "demo/blocks_kitchen_sink/run.py", line 7, in <module>
    base_theme = gr.themes.Base()
  File "/Users/freddy/sources/gradio/gradio/themes/base.py", line 222, in __init__
    self.set()
  File "/Users/freddy/sources/gradio/gradio/themes/base.py", line 714, in set
    self.background_accent = background_accent or getattr(
NameError: name 'background_accent' is not defined
```

CHANGELOG.md
freddyaboulton: I think that would make sense!

gradio/themes/base.py
aliabid94: whoops not sure when that broke, fixed now. renamed background_accent to color_accent since it was being used for text color as well. you may need to fix dracula if you changed that var.

aliabid94:
Okay everything should be working now as far as I can tell!

CHANGELOG.md
abidlabs: Agree

abidlabs:
LGTM @aliabid94! Let me go ahead and merge this in so we can review the final release PR

PR #3503, opened 2023-03-18
Use soft theme instead of glass theme in demo

aliabid94:
minor theme guide change


abidlabs:
LGTM @aliabid94 thanks for making the fixes!

PR #3519, opened 2023-03-19
Improve error messages for when argument lists mismatch

space-nuko:
# Description

When an event handler throws an error, I usually have no idea which event handler it was. This PR attempts to rectify that by introducing the function name and received/required arguments lists in the exception. It also improves the output message for when a JavaScript function modifies the input arguments, but those arguments don't match the list of inputs to the event handler.

Related: #3082

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -1006,6 +1006,7 @@ class Blocks(BlockContext):
             wanted = ", ".join(wanted_args)
             received = ", ".join(received_args)
 
+            # JS func didn't pass enough arguments
             raise ValueError(f"""An event handler{name} didn't receive enough input values (needed: {len(dep_inputs)}, got: {len(inputs)}).
 Check if the event handler calls a Javascript function, and make sure its return value is correct.
 Wanted inputs:
@@ -1058,7 +1059,6 @@ Received inputs:
             wanted = ", ".join(wanted_args)
             received = ", ".join(received_args)
 
-            # JS func didn't pass enough arguments
             raise ValueError(f"""An event handler{name} didn't receive enough output values (needed: {len(dep_outputs)}, received: {len(predictions)}).
 Wanted outputs:
     [{wanted}]
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -989,7 +989,11 @@ class Blocks(BlockContext):
         # Only check not enough args case, ignore extra arguments (for now)
         # TODO: make this stricter?
         if len(inputs) < len(dep_inputs):
-            name = f" ({block_fn.name})" if block_fn.name and block_fn.name != "<lambda>" else ""
+            name = (
+                f" ({block_fn.name})"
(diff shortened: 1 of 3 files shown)


freddyaboulton (commented):
Nice PR @space-nuko ! Thanks for the contribution. Can you merge in the latest and I can take a closer look?

gradio/blocks.py
```diff
@@ -997,6 +1034,36 @@ def preprocess_data(self, fn_index: int, inputs: List[Any], state: Dict[int, Any
             processed_input = inputs
         return processed_input
 
+    def validate_outputs(self, fn_index: int, predictions: List[Any]):
+        block_fn = self.fns[fn_index]
+        dependency = self.dependencies[fn_index]
+
+        dep_outputs = dependency["outputs"]
+
+        if len(predictions) < len(dep_outputs):
```
freddyaboulton, line 1043: In the case of a single output, predictions will not be a list and `len` would fail, e.g. for a `Number`

gradio/blocks.py
```diff
@@ -978,10 +978,47 @@ def deserialize_data(self, fn_index: int, outputs: List[Any]) -> List[Any]:
...
+                    v = f'"{inp}"'
+                else:
+                    v = str(inp)
+                received_args.append(v)
+
+            wanted = ", ".join(wanted_args)
+            received = ", ".join(received_args)
+
+            # JS func didn't pass enough arguments
+            raise ValueError(f"""An event handler{name} didn't receive enough input values (needed: {len(dep_inputs)}, got: {len(inputs)}).
```
freddyaboulton, line 1009: Can you add a test for this case (js function not passing enough args)? Thank you!

space-nuko: Can we run JS in tests? I'm fine with mocking it otherwise

gradio/blocks.py
freddyaboulton: I'm talking about this case

```python
import gradio as gr

with gr.Blocks() as demo:

    num1 = gr.Number()
    num2 = gr.Number()
    
    btn = gr.Button(value='1')
    btn_2 = gr.Button(value='2')

    btn.click(lambda x: x, num1, num2,)
    btn_2.click(lambda a: a, num1, [num1, num2])

demo.launch()
```

I get `TypeError: object of type 'float' has no len()` when clicking button 2

gradio/blocks.py
freddyaboulton: Actually it's probably fine to leave as is in this case!

gradio/blocks.py
abidlabs: Did you get a chance to look into this @space-nuko? Happy to help if needed and get this PR across the finish line

space-nuko: Sorry got busy, should be resolved now

freddyaboulton (approved):
Thanks @space-nuko ! Will merge once CI passes

PR #3529, opened 2023-03-20
fix login

aliabid94:
login broke, fixed now


freddyaboulton (commented):
Confirmed it works locally! Going to build cdn assets to test embed.

freddyaboulton:
@aliabid94 Like I said before, this works great locally!

However, I noticed that embedding with auth doesn't work as of this PR of 3.22.1. You can test with 

```html

<script
	type="module"
	src="https://gradio.s3-us-west-2.amazonaws.com/3.22.1/gradio.js"
></script>


<div>
    <gradio-app space="gradio/hello_login"></gradio-app>
</div>
```

It's trying to fetch the config, but it can't since it's not authenticated, hence the error.

image

Fine with me if we decide to tackle that in a separate PR but we should fix prior to next release cc @pngwn

pngwn:
@freddyaboulton We should take care of it in a separate PR. I noticed this earlier when I was testing locally.

freddyaboulton (approved):
Sounds good! Filed #3546

Issue #3552, opened 2023-03-20
Strip extra indentation / blank lines in `gr.Code.postprocess()`


abidlabs:
When people pass in a Python block string into the value of `gr.Code`, it often includes extra indentation or blank lines. We should strip this, similar to how we do for `Markdown`, etc. 

Came up while reviewing #3550

> The code is showing up with extra indentation, as well as an empty line at the top: 

<img width="386" alt="image" src="https://user-images.githubusercontent.com/1778297/226476219-5c5e6c9c-216c-47ef-b5cc-1b2244c93325.png">

abidlabs:
Yup @freddyaboulton I created an issue for that here: #3467

freddyaboulton:
Thanks for linking that issue - completely missed it! Glad I didn't file a new issue 😄

abidlabs:
Haha yea no problem!

PR #3556, opened 2023-03-21
Removes leading spaces from all lines of code uniformly in the `gr.Code()` component

abidlabs:
Fixes: #3552

Test code:

```py
import gradio as gr

with gr.Blocks() as demo:
    gr.Code("""
    def fn(a):
        return a
    """, language="python")
    
demo.launch()
```


aliabid94 (approved):
LGTM!

abidlabs:

PR #3562, opened 2023-03-21
Mobile responsive iframes in themes guide

aliabd:
There's a css issue with embedding the demos in the theming guide using web components. The other themes basically clash and the last one ends up applying to all the embedded demos. We had to resort to iframes but the current set up is really not mobile responsive at all. Very annoying for anyone reading the guide on a phone. 

Replacing with this. Please try it first though https://www.gradio-staging.com/theming-guide/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,8 @@ No changes to highlight.
 
 ## Full Changelog:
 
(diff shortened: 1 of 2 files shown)


abidlabs:
LGTM @aliabd! Much nicer on mobile

PR #3563, opened 2023-03-21
Remove extra $demo from theme guide

aliabd:
Missed a $demo from guide that needs to be removed.


freddyaboulton (approved):
Thanks @aliabd !

PR #3572, opened 2023-03-22
fix wc error in dev mode

pngwn:
This fixes an error we have been getting with the web component in dev mode. I do not know if this was causing any issues, nor do I know if this will fix the recent issues some people have had with dev mode not working correctly.


freddyaboulton:
How do I test this? dev mode reloads fine for me on latest main

pngwn:
If there are no errors in the console when running dev mode then it is working basically. Not sure if it had any impact really but thought it was worth cleaning up.

freddyaboulton (approved):
Cool, thanks @pngwn !

freddyaboulton:
Merging - thanks again @pngwn !

PR #3573, opened 2023-03-22
ensure css loads before mounting app

### LINKED ISSUE #3467: `Code` component label overlays on text after refreshing the page
### Describe the bug

Very strange bug:

When the Gradio app page containing the `Code` component is refreshed (but NOT hard-refreshed), then the label in the `Code` component overlays on the code string. Here's how it looks:

image

### Reproduction

Visit: https://huggingface.co/spaces/loubnabnl/santacoder-demo

Do a regular refresh (not a hard refresh) to see the issue 

### Screenshot

_No response_

### Logs

```shell
N/A
```

### System Info

```shell
Gradio 3.21
```

### Severity

annoying

pngwn:
Closes #3467.

We had a bit of a race condition because we weren't waiting for theme CSS to finish loading before mounting the app.

I've changed the CSS loading so that we ensure it has loaded before we try to mount the app. This also reduces some of the flashing we were seeing on first load.


pngwn:
This is very tricky to reproduce as it is a race condition but if you reload the `code` demo enough times (not in dev mode, not with the dev tools open) it is possible to reproduce on main.

freddyaboulton (approved):
Tested and works well with `code_component` @pngwn !

Thank you. is there a way we can unit test this?

pngwn:
Almost impossible to unit test because it was a race condition so a failing case would be almost impossible to create. We could mock stuff but then we would really be testing anything.

aliabid94 (approved):
ah I had noticed this, thanks for fixing! lgtm

Issue #2211, opened 2022-09-08
Cannot set theme in <gradio-app> meta tag


greendra8:
### Describe the bug

When using iframe to call a gradio element, +?__theme=dark can be appended to set the theme. When using <gradio-app> this does not work. 

This line in the gradio js:

let source = space ? `https://hf.space/embed/${space}/+/` : this.getAttribute("src");

should allow for a theme query string to be appended on the end eg https://hf.space/embed/${space}/+?__theme=${theme}

which would allow users to specify a theme tag in html. 

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Cannot set theme in <gradio-app> meta tag

### Logs

```shell
.
```

### System Info

```shell
Gradio 3.2
```

### Severity

annoying

PR #3590, opened 2023-03-23
UI fixes for themes

aliabid94:
Was playing around building some more themes, and came across a few theme variables with issues, now fixes:
- `bg_background_fill` did not cover the entire background, and dark mode used a hardcoded dark value to set to the overall background. I've fixed this so that the background_fill will affect the entire body if not in embedded mode.
  - while I was at it, simplified the dark mode detection logic, and also allow users to specify light or dark when embedding, e.g. `<gradio-app space="..." theme="light" />` so that it is consistent with the look of the host site
- the radius of a block form element (textbox, checkbox, etc) should be tied to `block_radius`, it was tied to `radius_lg`. Fixed
- a block label had the same shadow as a block, which looked weird at heavier shadows and there was no reason for them to be tied to each other. created a new css variable called `block_label_shadow`, and enabled backwards compatibility when loading a theme by filling any attributes in Base that were not present in the loaded theme with the value from the base theme.

can separate this PR into 2 PRs if that would help parse this.

Fixes: #2211


abidlabs:
It looks like this should close https://github.com/gradio-app/gradio/issues/2211 as well. (They asked for an iframe fix, but we can point them to web components as the preferred way of embedding)

freddyaboulton:
BTW @pngwn showed me that you can test the embed locally. 

```bash
> pnpm build:cdn-local
> pnpm preview:cdn-local
```

`localhost:3001` will display this [page](https://github.com/gradio-app/gradio/blob/main/ui/packages/_cdn-test/index.html) which you can modify to embed different demos.

abidlabs:
Nice!

ui/packages/app/src/Embed.svelte
```diff
@@ -48,7 +48,7 @@
 		display: flex;
 		position: relative;
 		flex-direction: column;
-		background: var(--body-background-fill);
+		/* background: var(--body-background-fill); */
```
freddyaboulton, line 51: Should we just delete instead?

freddyaboulton (commented):
Theme changes look good to me! Left a small comment about making the same modification to `themes/app.py`

ui/packages/app/src/main.ts
```diff
@@ -85,7 +85,7 @@ function create_custom_element() {
 					eager: eager === "true" ? true : false,
 					// gradio meta info
 					version: GRADIO_VERSION,
-					theme: this.theme,
+					theme: theme,
```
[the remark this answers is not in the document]
abidlabs: Agree. Maybe “mode”? We haven’t released yet so still  time to rename

ui/packages/app/src/Index.svelte
```diff
@@ -122,21 +122,15 @@
...
-			theme = darkmode(target);
+		let url_color_mode: "light" | "dark" | "system" | null =
+			url.searchParams.get("__theme") as "light" | "dark" | "system" | null;
+		url_color_mode =
+			url_color_mode ||
+			(url.searchParams.get("__dark-theme") === "true" ? "dark" : null);
+		theme = url_color_mode || theme;
+
+		if (theme === "dark" || theme === "light") {
+			darkmode(target, theme);
```
pngwn, line 133: Can this be refactored to remove the shorthand's and ternaries? This is far more difficult to reason about than the previous block of if statements.

pngwn (changes requested):

ui/packages/app/src/main.ts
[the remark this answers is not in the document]
pngwn: `mode` is too vague. `theme_variant`, `theme_mode`?

ui/packages/app/src/Index.svelte
pngwn: The user provided theme value should take precedence here as well.user mode > query param > system. I think we can get rid of them `__dark-theme` thing now but will update when @julien-c or @gary149 get back to me.

Also we could create a type for `type ThemeVariant =  "light" | "dark" | "system"` for ease of maintenance / readability.

So if we leave `user_theme_variant` undefined if it isn't passed, i think this can all be simplified to:

```ts
let theme_variant_query_param = url.searchParams.get("__theme") as ThemeVariant || undefined
active_theme_variant: ThemeVariant || undefined = user_theme_variant || theme_variant_query_param

if (theme_variant) {
  darkmode(target, theme_variant);
} else {
  active_theme_variant = use_system_theme(target);
}
```

ui/packages/app/src/Index.svelte
```diff
@@ -70,7 +70,7 @@
 	export let initial_height: string;
 	export let app_mode: boolean;
 	export let is_embed: boolean;
-	export let theme: "light" | "dark" = "light";
+	export let theme: "light" | "dark" | "system" = "system";
```
pngwn, line 73: Could we rename this to `user_theme_variant` or `user_theme_mode` (depending what we decide on for the attribute name) and create a new variable for the actual 'current' theme variant. `active_theme_variant` or `current_theme_variant` or something. I think it is clearer.

pngwn (changes requested):
Made some more noted after looking more closely. I think we can remove `__dark-mode` now. which simplifies things a lot.

julien-c:
> we can get rid of them __dark-theme thing

yep i think so

abidlabs:
@aliabid94 is this PR still relevant?

aliabid94:
Sorry for the delay on this one, I believe everything has been addressed

pngwn (approved):
Looks good! Thanks @aliabid94!

Issue #3588, opened 2023-03-22
demo.launch() within Blocks scope causes state error


aliabid94:
### Describe the bug

If demo.launch() is called within the Blocks scope, any state variable set with a Callable is shared across sessions. See example code - if you unindent demo.launch(), problem goes away

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```python
import gradio as gr
import uuid

def welcome(name):
    return f"Welcome to Gradio, {name}!"

with gr.Blocks() as demo:
    session = gr.State(uuid.uuid4)
    btn = gr.Button()
    out = gr.Textbox()
    btn.click(welcome, session, out)

    demo.launch()
```

### Logs

```shell
None
```

### System Info

```shell
3.22
```

### Severity

annoying

freddyaboulton:
Dang.

I wonder if it's because the load events aren't configured until you exit the blocks context:

```python
    def __exit__(self, *args):
        super().fill_expected_parents()
        Context.block = self.parent
        # Configure the load events before root_block is reset
        self.attach_load_events()
        if self.parent is None:
            Context.root_block = None
        else:
            self.parent.children.extend(self.children)
        self.config = self.get_config_file()
        self.app = routes.App.create_app(self)
        self.progress_tracking = any(block_fn.tracks_progress for block_fn in self.fns)
```

PR #3600, opened 2023-03-23
Fixes Blocks exit issue

abidlabs:
Fixes an issue where if the Blocks scope was not exited, then State could be shared across sessions. 

Fixes: #3588


freddyaboulton (approved):

abidlabs:
Likewise for the quick review @freddyaboulton!

PR #3605, opened 2023-03-24
move files

pngwn:
This PR moves the s js client into the `client` directory but in order to do so, I need to move some thing up a level so that the repo root is the root for the JS stuff too.
 
 I have no fully checked whether or not this will have an impact on the python builds, so could do with some reviews there.


pngwn:
Also need to test ci properly.

abidlabs:
Thanks @pngwn. Whenever tests are passing, I can make a beta release to pypi to ensure that python packaging still works

pngwn:
I think CI should pass now. I've always tweaked CI so the NPM publish thing should only run when JS related files have changed.

abidlabs:
Releasing `3.23.1b1` to test python build process

abidlabs:
The demos in the deployed PR look good. 

However, when I [triggered the deploy to pypi, it ran into an error](https://github.com/gradio-app/gradio/actions/runs/4533753152/jobs/7986992373): 

> The user-provided path gradio/templates/cdn does not exist.

It looks like this step is failing:

> aws s3 cp gradio/templates/cdn "s3://gradio/${new_version}/" --recursive

in [this script](https://github.com/gradio-app/gradio/blob/move-js-client/scripts/gh_action_pypi.sh). Does that make sense to you @pngwn?

pngwn:
I think i missed a `cd`. Will push a tweak.

pngwn:
We can do the JS stuff from root now, the pypi script was leaving the gradio directory where the necessary files do not exist. I think it should work now.

abidlabs:
Cool I'll rerun the action

abidlabs:
Still the same issue it looks like: https://github.com/gradio-app/gradio/actions/runs/4533753152/jobs/7987567548

abidlabs (approved):

pngwn:
Merged main, just gonna check the embeds work with that beta then I'll merge.

abidlabs:
Since this PR touches the entire frontend, let's merge in to avoid a bunch of annoying conflicts? Let me check the embeds

abidlabs:
Merging once the tests pass!

PR #3608, opened 2023-03-24
Simplify tests

abidlabs:


abidlabs:
Glad to hear it's working. 

> In the future, might be worth setting up Windows CI too

Yup that's something we can look into in the future

abidlabs:
Thanks guys for reviewing!

Issue #3481, opened 2023-03-16
Blocks.load should respect upstream theme


freddyaboulton:
- [ ] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
In the upcoming theme + theme sharing release, it would be desireable if loading a blocks via `load` and then `launching` it would respect the upstream theme.

**Describe the solution you'd like**  
```
gr.Blocks.load("freddyaboulton/dracula_revamped").launch()
```

respects the dracula_revamped theme.

**Additional context**

abidlabs:
It should probably respect upstream css as well

PR #3641, opened 2023-03-27
Load upstream theme

abidlabs:
Fixes: #3481 so that the theme from an upstream Space is loaded when you do `gr.load()`


gradio/blocks.py
```diff
@@ -506,11 +517,14 @@ def __init__(
         if theme is None:
             theme = DefaultTheme()
         elif isinstance(theme, str):
-            try:
-                theme = Theme.from_hub(theme)
-            except Exception as e:
-                warnings.warn(f"Cannot load {theme}. Caught Exception: {str(e)}")
-                theme = DefaultTheme()
+            if theme.lower() in BUILT_IN_THEMES:
```
freddyaboulton, line 520: Not related to your pr but let's add theme to the docstring (didn't realize it was missing) and explain it can be a string (built-in theme or upstream theme) or theme instance

freddyaboulton: Nit, but being able to pass a str for a built in theme should go on the changelog

freddyaboulton (approved):
Originally, I thought we would fetch the css for the theme directly from the `/css` route but it's better to only apply this to new releases.

Thanks @abidlabs ! I wonder if `theme` should be added to the Config class in the front end ?

abidlabs:
Thanks for the review @freddyaboulton! As far as:

> I wonder if theme should be added to the Config class in the front end ?

It already is:

https://github.com/gradio-app/gradio/blob/a46b1806bb029924c5368edb9d58065e4b642de3/globals.d.ts#L23

gradio/blocks.py
abidlabs: Sounds good! Will make these changes

abidlabs:
Merging now, thanks again @freddyaboulton!

Issue #3564, opened 2023-03-21
More than one `gr.Dropdown` item can receive `.selected` class


space-nuko:
### Describe the bug

This isn't a user-facing issue unless custom CSS is used, but I ran across it when trying to add item highlighting to dropdowns

It seems like more than one element in the dropdown can be given a `.selected` class if another element's text is contained within the currently selected element's text

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

1. Create a `gr.Dropdown` with a bunch of items, all with shared prefixes
2. Select the longest element
3. Inspect the CSS classes given to each `li` in the dropdown

### Screenshot

<img width="303" alt="2023-03-21 18_17_32-Stable Diffusion - Chromium" src="https://user-images.githubusercontent.com/24979496/226754565-da5909ef-b445-42b4-b0ba-29b769d2fac2.png">

### Logs

```shell
N/A
```

### System Info

```shell
3.22.1, Windows, Chrome
```

### Severity

annoying

PR #3644, opened 2023-03-27
Only give same values .selected class in gr.Dropdown

space-nuko:
# Description

Closes: #3564

Still gives elements with the same value in the same dropdown the selected checkmark, unsure if this is desirable

2023-03-27 16_32_01-Gradio - Chromium

# Checklist:


js/form/src/DropdownOptions.svelte
```diff
@@ -49,7 +49,7 @@
 			<li
 				class="item"
 				role="button"
-				class:selected={value?.includes(choice)}
+				class:selected={value === choice}
```
aliabid94, line 52:
```suggestion
				class:selected={Array.isArray(value) ? value.includes(choice) : value === choice}
```
So value can either be a single string, or an array of strings in multiselect mode. I believe this is why the original logic used "includes" to handle multiselect mode, but this suggestion should handle both cases.

pngwn (approved):
When we've fixed the changelog and run `pnpm format:write` to fix the formatting, i think this is good to go.

js/form/src/DropdownOptions.svelte
```diff
@@ -32,6 +32,11 @@
 	}
 
 	const dispatch = createEventDispatcher();
+
+	function isSelected(choice) {
```
pngwn, line 36:
```suggestion
	function isSelected(choice: string) {
```

pngwn (changes requested):
This should fix the TS error.

pngwn (approved):

Issue #3565, opened 2023-03-21
Remaining blockers for Python client


abidlabs:
From #3300, here are the remaining TODOs for the Gradio Python client:

**Necessary**

- [x] add a changelog for the client library -- @abidlabs 
- [x] add to view API page -- @aliabd 
- [x] Files should be uploaded via /upload -- @freddyaboulton 
- [x] Add a Guide for using the python client -- @abidlabs & @aliabd 
- [ ] rate limit public Spaces, and display a meaningful error when rate limited (ideally saying something like you can duplicate this Space and make it private to avoid being rate limited)
- [x] add documentation for this on the website (@aliabd)
- [x] add support for iterators -- right now iterators return a meaningless update as a result @freddyaboulton 
- [x] add a way to get the status of a job @freddyaboulton 
- [x] add ability to inspect and quick get an idea of how to use each of the API endpoints for a given gradio app, including unnamed endpoints
- [x] add more tests @freddyaboulton 
- [x] add to the changelog
- [x] clarify that callbacks are executed sequentially
- [x] Session management / a method to reset session
- [x] add new version release GH action
- [x] add analytics
- [x] Document that the HF token is pulled if you are logged in via HF CLI
- [x] `__repr__` returns error
- ~Support passing in images as numpy arrays / pil images (note: I don't think we should do this in the client)~ 
- ~Paths to files / images are confusing. Can we display them instead?~ 
- [x] "/predict" instead of "predict"
- [x] we shouldn't assume fn_index=0

abidlabs:
Thank you so much @freddyaboulton!

freddyaboulton:
We should add a CHANGELOG for the clients, no?

abidlabs:
👍 sounds good, added to the list

pngwn:
JS client already has one, fully automated: https://github.com/gradio-app/gradio/blob/8eec0541b281c1adb968b8100c0d7d16d7eab1a2/client/js/CHANGELOG.md

freddyaboulton:
@abidlabs Is support for spaces with auth on the roadmap?

abidlabs:
Yes let's create a separate issue for that and the other nice-to-haves because they are not blockers to launch (I can create the issues)

abidlabs:
Everything except rate limiting is complete, congrats everyone 🥳

Will close since rate limiting is a moon-landing issue, not a gradio issue

justinzyw:
[...]

PR #3645, opened 2023-03-27
Add status for Python Client Jobs

freddyaboulton:
# Description

From #3565 

Adds a method called `status()` to the `Job` class that returns the latest status from the backend. This status update has the following interface:

```python
@dataclass
class StatusUpdate:
    """Update message sent from the worker thread to the Job on the main thread."""

    code: Status
    rank: int | None
    queue_size: int | None
    eta: float | None
    success: bool | None
    time: datetime | None
```

and the `status` is an enum with the following values

```python
class Status(Enum):
    """Status codes presented to client users."""

    STARTING = "STARTING"
    JOINING_QUEUE = "JOINING_QUEUE"
    QUEUE_FULL = "QUEUE_FULL"
    IN_QUEUE = "IN_QUEUE"
    SENDING_DATA = "SENDING_DATA"
    START_ITERATION = "START_ITERATION"
    ITERATING = "ITERATING"
    FINISHED = "FINISHED"
```

For apps that are not queued, the `status` is manually set to be either `ITERATING` or `FINISHED` depending on the value of `job.done()`. 

Example running on stable diffusion space (outdated as it doesn't have the eta field):

queue_updates
 
### Additional Modifications

- The code snippet in the README for the python client was not working so I changed it
- Modified the CI to install the local version of gradio_client
- Modified format_backend and lint_backend to check the python client test directory

# Checklist:

# A note about the CHANGELOG


client/python/gradio_client/utils.py
```diff
@@ -38,6 +42,55 @@ class InvalidAPIEndpointError(Exception):
     pass
 
 
+class Status(Enum):
+    STARTING = "STARTING"
+    JOINING_QUEUE = "JOINING_QUEUE"
+    QUEUE_FULL = "QUEUE_FULL"
+    IN_QUEUE = "IN_QUEUE"
+    SENDING_DATA = "SENDING_DATA"
+    START_ITERATION = "START_ITERATION"
```
freddyaboulton, line 51: Using the terms iteration and iterating to be consistent with generator functions. But maybe we can rename `START_ITERATION`  to `PROCESSING` instead?

client/python/gradio_client/client.py
```diff
@@ -156,42 +163,53 @@ def __init__(self, client: Client, fn_index: int, dependency: Dict):
...
-    def end_to_end_fn(self, *data):
-        if not self.is_valid:
-            raise utils.InvalidAPIEndpointError()
-        inputs = self.serialize(*data)
-        predictions = self.predict(*inputs)
-        outputs = self.deserialize(*predictions)
-        if len(self.dependency["outputs"]) == 1:
-            return outputs[0]
-        return outputs
+    def make_end_to_end_fn(self, queue: LifoQueue | None = None):
```
freddyaboulton, line 166: What I did here was make these functions into factory functions so that we can thread the queue all the way down

freddyaboulton:
Still need to write docstrings and make some other minor changes but this is pretty stable now and would love to get comments on the api :)

scripts/install_gradio.sh
```diff
@@ -7,3 +7,4 @@ pip_required
 
 echo "Installing Gradio..."
 pip install -e .
+pip install -e client/python
```
abidlabs, line 10: I actually think the client should *not* be editable installed. Rather, it should pull the latest from pypi as this is consistent with how users will be using the gradio library. 

Suppose the following happens:
* We make some changes in both the client library and gradio library (e.g. add a util to the client library, and add some feature in the gradio library which uses this util) 
* If the client is editable installed, then all tests will continue to pass, and we won't realize that we've made a breaking change. We could release the next version of the gradio library thinking everything is working fine because all tests are passing. But, in reality, gradio won't work properly
* If the client is installed from pypi, then the gradio tests will fail because the tests will correctly catch that we have not made a gradio client release to pypi yet.

client/python/README.md
```diff
@@ -56,13 +56,12 @@ client = grc.Client(src="btd372-js72hd.gradio.app")
...
 The simplest way to make a prediction is simply to call the `.predict()` function with the appropriate arguments and then immediately calling `.result()`, like this:
 
-
 ```python
 import gradio_client as grc
 
 client = grc.Client(space="abidlabs/en2fr")
 
-client.predict("Hello").result()
+client.make_predict("Hello").result()
```
abidlabs, line 64: Could we keep the name of the function `.predict()` for a nicer user experience?

freddyaboulton: The name is still predict - this is the auto refactoring feature gone wrong. Will fix!

abidlabs:
@freddyaboulton and I synced on the internal implementation. 

In terms of the user facing API, this looks good @freddyaboulton. The only additional piece -- can we get an ETA back well? This could be useful if people are building something with the client and want to show an estimated ETA

client/python/gradio_client/utils.py
```diff
@@ -38,6 +42,92 @@ class InvalidAPIEndpointError(Exception):
...
+
+@dataclass
+class JobStatus:
+    """The job status.
+
+    Keeps strack of the latest status update and intermediate outputs (not yet implements).
+    """
+
+    latest_status: StatusUpdate = field(default_factory=create_initial_status_update)
+    outputs: List[Any] = field(default_factory=list)
```
freddyaboulton, line 120: The outputs field is not used yet but I’m adding it here because we will need it for storing intermediate outputs of generators

freddyaboulton:
@abidlabs simplified the implementation and added the eta to the StatusUpdate!

client/python/gradio_client/utils.py
abidlabs: I prefer PROCESSING as well

abidlabs:
LGTM @freddyaboulton this is awesome! The `status()` function works great in my tests, including when a client runs multiple jobs. You might want to add a test for that: having a client run multiple jobs in parallel and confirm that the statuses for each job appear correctly.

abidlabs (approved):

freddyaboulton:
Thanks for the review @abidlabs ! Just added a test for the two concurrent jobs case. Will merge once CI passes.

abidlabs:
@freddyaboulton do you know why this happens sometimes:

image

It says your tests have passed but it didn't even run the backend tests. Perhaps this related to tests queueing up?

freddyaboulton:
Yea I think it's the queue build up @abidlabs !

Issue #3526, opened 2023-03-20
`Remove All` button on `gr.Dropdown` should be hideable/configurable


space-nuko:
### Describe the bug

Dropdowns with a single element might not want to support removal of the single element. There should be a property on `gr.Dropdown` that configures if the button is hidden or not.

Also, the user might want to configure the `Remove All` behavior. Say there's a `None` element in the dropdown they'd want to switch to if the button is clicked, it should be able to be passed as a property, and the button should be hidden if that item is selected

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

1. Create `gr.Dropdown()` with `multiselect=false`
2. Populate dropdown with items
3. `Remove All` button appears

### Logs

```shell
N/A
```

### System Info

```shell
3.22.1, Windows, Chromium
```

### Severity

annoying

space-nuko:
Actually this is a duplicate of #3435

pngwn:
Dupe of #3435 as mentioned. Closing.

Issue #3435, opened 2023-03-10
Add option to disable `Clear All` button in Dropdown or add a min_choices parameter


freddyaboulton:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
All dropdowns now come with an `x` button that allows the user to clear all selected choices.

As discussed in this [discord thread](https://discord.com/channels/879548962464493619/1083290221828907108/1083290221828907108), this button can cause problems if you have a `change` event on `Dropdown` and you don't expect the value of the drop down to be empty.

**Describe the solution you'd like**  
Three possible solutions I can think of:
1. Add `clear_all=bool` to `style` method to disable the clear_all button
2. Add `min_choices` parameter. `Change` event won't fire unless this is met. Kinda clunky now that I think about it lol.
3. Do nothing. It's up for the app author to guard against this case.

I'm between 1 and 3. Curious on everyone's thoughts.

**Additional context**

abidlabs:
I think this is a broader limitation right now of Gradio components -- that we don't have any way to indicate *required* inputs (issue: #2718). I think we should design a general solution around validation / indicating that some inputs are required (and preventing predictions from going through if those fields are empty) rather than specifically handling this for Dropdown component.

msampson3d:
Agree that an easy to use validation framework would be great, which allowed users to flag components as required and do a validation on a specific trigger (button click etc)

msampson3d:
One additional improvement on Dropdown: if the dropdown isn't multiselect, the dropdown should close when you click an entry. Right now you have to click out of it.

msampson3d:
I'm an advocate for option 1, personally

g-wellsa:
for example  The dropdown how to set default option

g-wellsa:
your project is so bad

msampson3d:
> for example The dropdown how to set default option

Its right in the docs... https://gradio.app/docs/#dropdown  Set the 'value' attribute.

abidlabs:
But until we have a proper validation framework for all components, I think (1) is a pretty reasonable addition to the API.

pngwn:
The issue here is that the Dropdown doesn't match standard dropdown semantics, not arguing against the need for validation but I don't think that is the issue here. I don't think this needs an option either, just better defaults. The single dropdown shouldn't have a clear button and if there is no `None` option and a `value` (initial) is passed then it won't be possible to submit `None`.

PR #3648, opened 2023-03-27
Make Remove All button in `gr.Dropdown` configurable

space-nuko:
# Description

- Makes Remove All button in `gr.Dropdown` hidable with `show_remove_all=False`
- Provide `default_value` to switch to if button is clicked in single-select mode
- Remove All button is hidden in single-select if default value is selected
- Remove All button is hidden in multiselect if no values are selected

Closes: #3526, closes: #3435

# Checklist:


abidlabs:
I find the name `default_value` confusing, as it sounds like the value that should be shown in the `Dropdown` component when it is first loaded (what `value` currently refers to). A better name imo would be `cleared_value` -- and, I would set the `.cleared_value` attribute of the `Dropdown` class to be equal to this `cleared_value` parameter. The effect of this would be that in `gr.Interface`, when the "Clear" button is clicked, then the `Dropdown` would reset to this value. (and I would similarly rename `show_remove_all` to be `show_clear_button` -- to refer to this parameter and be parallel to `show_copy_button`).

freddyaboulton:
I'm wary of adding `cleared_value` to the API since an app author can implement that themselves in their event handler (`dropdown_value = cleared_value if not value else value`) and it's very different from how the other components behave. What's the use case?

space-nuko:
The `cleared_value` can be used even if there are no event handlers on the `gr.Dropdown`, it affects the behavior of the button instead of anything caused by an event

pngwn:
Why not just add a clear event instead?

I don't understand why we would want to remove the clear button for multiselect dropdowns either. Single select should not have a clear button at all.

space-nuko:
So instead of the additions there should be a `clear()` event for dropdowns?

> `default_value` seems unnecessary when you can choose a default in the predict function that the dropdown is an input for or select an initial value.

What does this mean? I was imagining having a choices list like `["None"] + models_list` autogenerated and being able to easily default to `"None"` if the button is clicked, otherwise it can be hidden

pngwn:
> What does this mean? I was imagining having a choices list like ["None"] + models_list autogenerated and being able to easily default to "None" if the button is clicked, otherwise it can be hidden

Ah, I misunderstood. `default_value` is intended as `cleared_value` rather than a default. I think this could be handle with events without the need for additional constructor kwargs.

See what @freddyaboulton and @abidlabs think.

abidlabs:
Reading over this PR again, I agree with the hesitation about adding another parameter, `cleared_value`, to the `gr.Dropdown`, for several reasons:

* It breaks user expectations -- when a user clicks on an "X" in a Dropdown, they expect the Dropdown to clear, not to be set to some customizable value
* The `value` parameter already exists and `cleared_value` forces developers to think about two different kinds of defaults: one when the page loads and one when the dropdown is cleared

I suggest we adopt @pngwn's suggestion here and simply do this:

* remove the clear button for single select dropdowns (this would solve #3435)
* keep the clear button for multiselect dropdowns as is

This wouldn't solve the second part of #3526, but like I mentioned above, imo that's not a good UX anyways.

space-nuko:
Okay thanks for the feedback, I will redo my changes in accordance with @pngwn's ideas

Issue #3534, opened 2023-03-20
Inpainting canvas isn't full size


space-nuko:
### Describe the bug

I'm noticing I have to add `height: 100%` to the canvas wrapper (the thing with `data-testid="image"`) so that the image takes up the whole size of the element, otherwise this happens:

<img width="572" alt="2023-03-20 02_16_13-Stable Diffusion - Chromium" src="https://user-images.githubusercontent.com/24979496/226266029-06482609-0ad1-4749-9506-67eed1e9d242.png">

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Create `gr.Image` with `tool="sketch"`

### Logs

```shell
N/A
```

### System Info

```shell
Gradio 3.22.1, Windows, Chrome
```

### Severity

annoying

Issue #466, opened 2022-01-19
`Image` component requests


pngwn:
**Is your feature request related to a problem? Please describe.**

We are getting a lot of feature requests for the different interactive `Image` variants, historically we had numerous different tools to handle the different kinds of Image editing functionality, this has improved somewhat but the Image is less feature rich than it used to be. Sadly over time the new (kinda) `Image` component has become difficult to maintain and extend and need substantial refactoring to realise its full potential.

We are also finding the current signature of pre and post-process limiting, the challenge here is that the sensible thing would be to issue a breaking change to make a clean break with the past but we don't want to introduce too much churn for users.

This issue will collate all feedback we have received so far (assuming I can find it all) and act as a single place to discuss features and design of a new unified image component.

**overview**

Broadly speaking the `Image` component (as an interactive input) has two key parts: the source of the image and the editing capabilities. The rewrite that will stem from this issue will preserve the different inputs (and maybe there are more people would like to see) but unify the editing tools into a single, simple (hopefully) GUI. The high level thinking is that the Gradio developer will be able to toggle and constrain the various features one by one if necessary (with defaults and templates making this even simpler for users who do not need such granular control).

References to "Gradio API" refer to controlling the feature via the Gradio Python API when creating the app.
References to "GUI" refer to end users controlling the feature in the browser when interacting with the tool.

**inputs/source**

- **none** - blank background on which to sketch (other uses?)
  - Gradio API - Can be constrained to a specific background colour (removing the config from the GUI), or a set of valid colours, or left open.
  - GUI - Can be changed by users if left open of limited to a set of users. No config if a single background colour constraint is set in Gradio API.
- **webcam** - Webcam image snapshot can be taken to be used as a background image.
- **upload** - Background image can be uploaded from the users computer (either by browsing or drag+drop)

One thing that has crossed my mind is allowing multiple inputs, this would allow Gradio app authors to be very flexible with what kinds of `Image` sources are set which will work well for some more general models. Could be controllable via the GUI (defaulting to blank background but with buttons/ toggles to enable different source modes).

Note: `source=canvas` would be deprecated as everything can be a canvas in the new world.

Are there other possible inputs we should consider here?

**editor tools**

[...]

PR #3649, opened 2023-03-27
Fix sketch tool `gr.Image` not filling up the entire component size

space-nuko:
# Description

Closes: #3534

2023-03-27 18_47_04-Gradio - Chromium

I'm not sure what the `.fixed-height` class was meant to achieve, can @pngwn or someone else clarify?

# Checklist:


abidlabs:
Thanks @space-nuko! As you may have noticed, we just [moved all of the frontend files](https://github.com/gradio-app/gradio/pull/3605/files). Maybe not the best timing given all of your PRs, but if you could kindly update your PRs to reflect the changes in the new files, we'll do our best to review the PRs quickly.

pngwn:
Pretty sure we don't want to remove this entirely. Will take a closer look at some point but might take me a while to get to.

Image is a low priority atm because it will be rewritten shortly and generally works ok. I'd like to minimise the amount of time I spend on this until we have changed some things.

space-nuko:
Okay thanks for the update @pngwn. Is there a tracking issue for such rewrite yet?

pngwn:
It's here: https://github.com/gradio-app/gradio/issues/466

I've started work on this in the background so this is very much on the radar and coming when a few other things are taken care off. As a result we want to minimise effort spent on the Image component, outside of simple fixes because we don't want to waste effort on something that will be removed shortly.

js/atoms/src/Block.svelte
```diff
@@ -32,6 +32,7 @@
 		"" +
 		(typeof style.height === "number" ? `height: ${style.height}px; ` : "") +
 		(typeof style.width === "number" ? `width: ${style.width}px;` : "");
+	console.log(style);
```
abidlabs, line 35: Remove

abidlabs, line 35:
```suggestion
```

abidlabs:
@aliabid94 can confirm the original issue is fixed. However, the cropper (which I think you were also trying to fix in this PR) is behaving very strangely. Here's a GIF:

```py
import gradio as gr

gr.Interface(
    lambda x: x,
    gr.Image(source='image', tool='editor'),
    [gr.Image()],
).launch()
```

Recording 2023-04-10 at 14 03 29

abidlabs (approved):

abidlabs:
Thanks @space-nuko and @aliabid94 for the fix!

Issue #3551, opened 2023-03-20
Can't focus `gr.Code` if there's extra space after empty code lines


space-nuko:
### Describe the bug

I can't click on the extra space not covered by the code lines with `gr.Code` to focus it, I have to click within the area delineated by the blue rectangle instead

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```python
#!/usr/bin/env python

import gradio as gr


with gr.Blocks() as demo:
    with gr.Row():
        gr.Textbox("Focus me", interactive=True)
    with gr.Row():
        gr.Code("a = 42", language="python", interactive=True)
        gr.HTML("test<br>" * 40)


demo.queue().launch(server_port=9876)
```

### Screenshot

<img width="864" alt="2023-03-20 17_15_15-Stable Diffusion — Firefox Developer Edition" src="https://user-images.githubusercontent.com/24979496/226468155-91faa640-bf9d-42b4-b7ce-d2c9f08cda3a.png">

### Logs

```shell
N/A
```

### System Info

```shell
4.22.1, Windows, Firefox Nightly
```

### Severity

annoying

abidlabs:
Sorry having a little trouble understanding this issue @space-nuko. Can you attach a code snippet or a GIF to help understand what the issue is? For what it's worth, I tried running this:

```py
import gradio as gr

with gr.Blocks() as demo:
    gr.Code("""
    def print(x):
      pass
      
      
      
    """, interactive=True)
    
demo.launch()
```

And I'm not having any issue focusing on the `gr.Code` component: 

<img width="609" alt="image" src="https://user-images.githubusercontent.com/1778297/226484269-77fca5f4-dc6a-4772-a1b7-2d817c3e02f1.png">

space-nuko:
Added a demo, the problem happens when you have two columns with a very tall height, `gr.Code` has a lot of blank space

abidlabs:
Oh I see the issue now. For future reference:

![image](https://user-images.githubusercontent.com/1778297/227567112-592fdd9b-1fbd-475d-a368-300f11a2c095.png)

Issue #3528, opened 2023-03-20
`gr.Code` should support `lines`


space-nuko:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
I want to use `gr.Code` like a `gr.Textbox`, however the UI has some finicky width/height requirements. But `gr.Code` has no `lines` property, so I can't manipulate the size in an obvious way

**Describe the solution you'd like**  
`lines` for `gr.Code` with the equivalent functionality of `gr.Textbox`'s version

**Additional Context**
https://discuss.codemirror.net/t/making-the-editor-n-lines-high-by-default/3198/6

PR #3651, opened 2023-03-28
Make `gr.Code` support `lines`, expand to fill entire element

space-nuko:
# Description

- `gr.Code` now supports `lines` like `gr.Textbox`
- `gr.Code` expands to fill the entire height of its wrapper

Closes: #3551, closes: #3528

Testcase:

```python
#!/usr/bin/env python

import gradio as gr

with gr.Blocks() as demo:
    with gr.Row():
        gr.Textbox("Focus me", interactive=True)
    with gr.Row():
        gr.Code("a = 42", language="python", interactive=True)
        gr.HTML("test<br>" * 40)
    with gr.Row():
        gr.Code("a = 42", language="python", interactive=True, lines=12)


demo.queue().launch(server_port=9876)
```

# Checklist:


aliabd:
Great work @space-nuko! I tested this and it mostly works as expected, but i'm still seeing the behavior you describe in #3551 when using Interface (vs Blocks). Not sure why.. maybe some style elements are clashing.. Let's have @pngwn take a look as well.

aliabd:
Unrelated, I noticed an issue where using `gr.Code.update()` sets `interactive=True` which it shouldn't, so pushed a commit that fixes that (for `Code` and also `BarPlot`)

space-nuko:
> i'm still seeing the behavior you describe in https://github.com/gradio-app/gradio/issues/3551 when using Interface (vs Blocks)

Would you have a repro for that?

space-nuko:
It's because of the `.unequal-height` class, which `gr.Interface` seems to add internally

Don't know if that should be made configurable?

abidlabs:
@aliabid94 or @pngwn do you have any suggestions here? Or should we just merge this in for now and then address the Interface case later?

pngwn:
Yeah. I think we can fix that issue in another PR. Will take one last look over before merging.

PR #3663, opened 2023-03-28
Documentation-related fixes to the python client

### LINKED ISSUE #2364: Feature request: API Introspection
Making an external frontend integrating with a gradio interface using the API can be really difficult if the gradio API is constantly changing.

Please provide a way to detect the inputs and outputs of a gradio API (it's too bad the interface is not made using GraphQL or we would be able to use [introspection](https://graphql.org/learn/introspection/))

Context: I'm currently writing a [GUI](https://github.com/leszekhanusz/diffusion-ui) for a gradio interface and the gradio interface is constantly changing, breaking the API in the process.
I would like to programmatically send a request to the API to detect the different functions, fn_id, inputs and outputs of the interface, the default input values, min, max, type of input, etc...

abidlabs:
This PR:

* Adds docstrings for the `Client` class and methods, clarifies that callbacks are executed sequentially
* Adds usage information for a given API. For example, for this Space (https://abidlabs-stt-or-tts.hf.space/), getting the usage info looks like this:

```py
client = grc.Client(space="abidlabs/stt_or_tts")
client.view_api()
```

which produces:

```
Client.predict() Usage Info
---------------------------
Named endpoints: 2
 - predict(input, api_name="predict") -> audio
    Parameters:
     - [Textbox] input: str (value)
    Returns:
     - [Audio] audio: str (filepath or URL)
 - predict(input, api_name="predict_1") -> output
    Parameters:
     - [Audio] input: str (filepath or URL)
    Returns:
     - [Textbox] output: str (value)

Unnamed endpoints: 1
 - predict(parameter_0, fn_index=2) -> input,audio
    Parameters:
     - [Dataset] parameter_0: str (value)
    Returns:
     - [Textbox] input: str (value)
     - [Audio] audio: str (filepath or URL)

```

You can also get the API in a machine-friendly way by simply doing `client.view_api(return_info=True)`, so this closes #2364 

In order to get this info programmatically, I expanded the set of Serializers. 

Related: #3565


abidlabs:
Although still need to add tests and fix old tests, I'd appreciate a review for feedback on this documentation format and approach for getting the documentation.

cc @aliabd -- I was able to get all of the documentation entirely from the backend. I think we can remove the documentation from the frontend and use this instead.

abidlabs:
Synced with @freddyaboulton, I'll make some changes to this PR before opening it for reivew

abidlabs:
Opening this up for review

(Note that GitHub actions is currently down, but ci.sh can be run locally)

client/python/gradio_client/client.py
```diff
@@ -27,6 +27,13 @@ def __init__(
         hf_token: str | None = None,
         max_workers: int = 40,
     ):
+        """
+        Parameters:
+            space: The name of the Space to load, e.g. "abidlabs/pictionary". If it is a private Space, you must provide an hf_token. app. Either `space` or `src` must be provided.
```
freddyaboulton, line 32: Not related to your PR but just thought about this - what if we collapse `space` and `src` into the same parameter? Basically we automatically detect if the value is a space name or a url.

client/python/gradio_client/client.py
```diff
@@ -93,6 +109,67 @@ def fn(future):
 
         return job
 
+    def usage(self, all_endpoints=True, print_usage=True) -> str | None:
```
freddyaboulton, line 112: What if we call the method `view_api`? I like that it matches what a user would see when opening the UI.

client/python/gradio_client/serializing.py
```diff
@@ -12,33 +12,22 @@
 
 class Serializable(ABC):
     @abstractmethod
-    def serialize(self, x: Any, load_dir: str | Path = ""):
+    def get_input_type() -> str:
```
freddyaboulton, line 15: nit but these should take `self` as well

client/python/gradio_client/client.py
freddyaboulton, line 112: I think setting `all_endpoints=True` as the default will generate a lot of noise for developers. The `Dataset.click` event will be shown for every interface, for example.

We should start encouring app authors to document the methods they intend to be used via api via api_name (or some other mechanism?)

client/python/gradio_client/client.py
```diff
@@ -153,10 +231,39 @@ def __init__(self, client: Client, fn_index: int, dependency: Dict):
...
+                        )
+                    parameters[label] = info
+        returns = {}
+        for o, output in enumerate(self.dependency["outputs"]):
+            for component in self.config["components"]:
+                if component["id"] == output:
+                    label = component["props"].get("label", f"parameter_{o}").lower()
+                    if "info" in component:
+                        info = component["info"]["output"]
+                    info = (
```
freddyaboulton, line 259: There should be an `else` here, no?

client/python/gradio_client/client.py
```diff
@@ -92,6 +108,67 @@ def fn(future):
 
         return job
 
+    def usage(self, all_endpoints=True, print_usage=True) -> str | None:
```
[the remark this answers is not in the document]
abidlabs: Yup sounds good!

client/python/gradio_client/client.py
abidlabs: Sure, I don't think that shouldn't lead to any conflicts so let's do it

client/python/gradio_client/client.py
abidlabs: Yup but the reason is that most developers who are working with Blocks don't add api names. As a result, people don't even know the Space *could* be used as an API, which I think is a bigger problem.

One thing we could do is by default set `all_endpoints=False`, but then show at the bottom: "3 unnamed endpoints. To show usage for unnamed endpoints, set `all_endpoints=True`. WDYT?

freddyaboulton:
This looks so slick for the stable diffusion space BTW! 🔥 

image

client/python/gradio_client/client.py
abidlabs: Or another idea is that if `all_endpoints` is not explicitly provided, we show all endpoints if there are no named endpoints.

freddyaboulton (commented):
@abidlabs Thanks for making the changes! This looks great. I guess we need to release to PyPi to fix the CI?

freddyaboulton (approved):

abidlabs:
Thanks @freddyaboulton for the great suggestions! Yup still doing some more tweaks to get the machine readable part working, but I'll release to pypi once everything is working

abidlabs:
All righty, did some more cleanup, now we have a machine-parsable info, everything is passing, and docs looks pretty nice imo.  

Will release a new version of the client to pypi

abidlabs:
Lots of little annoying fixes, but we should be there soon (cc @aliabd)

abidlabs:
Thanks for the review @freddyaboulton, finally in!

PR #3664, opened 2023-03-28
Theme builder

aliabid94:
Created automatic theme builder, launchable via 
```python
import gradio as gr
gr.themes.builder()
```
Screen Shot 2023-03-28 at 3 07 12 AM

Decided to make it a local app instead of on our site or on spaces, because the app itself is heavy and would otherwise be slow.

In the process of building this, fixed / changed some UI elements:
- Added support for dropdowns to allow custom text values
- Added blur event listener to colorpicker


CHANGELOG.md
```diff
@@ -4,7 +4,14 @@
...
 
-- Trigger the release event when Slider number input is released or unfocused  by [@freddyaboulton](https://github.com/freddyaboulton) in [PR 3589](https://github.com/gradio-app/gradio/pull/3589)   
+- Trigger the release event when Slider number input is released or unfocused  by [@freddyaboulton](https://github.com/freddyaboulton) in [PR 3589](https://github.com/gradio-app/gradio/pull/3589)
+- Created Theme Builder, which allows users to create themes without writing any code, by [@aliabid94](https://github.com/aliabid94) in [PR 3664](https://github.com/gradio-app/gradio/pull/3664). Launch by:
+    
+    ```python
+    import gradio as gr
+    gr.themes.builder()
+    ```
+
```
abidlabs, line 14:
```suggestion
- Created Theme Builder, which allows users to create themes without writing any code, by [@aliabid94](https://github.com/aliabid94) in [PR 3664](https://github.com/gradio-app/gradio/pull/3664). Launch by:
    
    ```python
    import gradio as gr
    gr.themes.builder()
    ```

- The `Dropdown` component now has a `allow_custom_value` parameter that lets users type in custom values not in the original list of choices.
- The `Colorpicker` component now has a `.blur()` event
```

demo/blocks_kitchen_sink/run.py
```diff
@@ -123,7 +123,7 @@
 
                 def go(*args):
                     time.sleep(3)
-                    return "https://gradio.app/assets/img/header-image.jpg"
+                    return "https://i.ibb.co/6BgKdSj/groot.jpg"
```
abidlabs, line 126: Why this change?

gradio/themes/builder.py
```diff
@@ -0,0 +1,945 @@
...
+                theme_repo_name,
+                theme_version,
+                theme_description,
+            },
+            theme_upload_status,
+        )
+
+
+if __name__ == "__main__":
+    demo.launch()
```
abidlabs, line 945: Suggestion:
```suggestion
    demo.launch(inline=False, inbrowser=True)
```

gradio/themes/builder.py
```diff
@@ -0,0 +1,945 @@
...
+}
+
+#controls::-webkit-scrollbar-thumb {
+  border-radius: 4px;
+  background-color: rgba(0, 0, 0, .5);
+  box-shadow: 0 0 1px rgba(255, 255, 255, .5);
+}
+"""
+
+with gr.Blocks(theme=gr.themes.Base(), css=css) as demo:
```
abidlabs, line 78: Suggestion:
```suggestion
with gr.Blocks(theme=gr.themes.Base(), css=css, title="Gradio Theme Builder") as demo:
```

abidlabs:
We've now removed the "X" symbol from the single-select `Dropdown`, so it's very hard to know that the Dropdown is typeable. I would suggest we add the blinking line that indicates that you can type somewhere when you make a selection from the Dropdown. Here's how it currently looks like for me:

Recording 2023-03-29 at 09 56 57 (1)

CHANGELOG.md
freddyaboulton: The image from the PR description would go nicely under the codeblock

abidlabs:
It feels like there are too many fields here, which makes it unnecessarily difficult for people to upload their themes:

image

* Why is theme name different from repo name? Let's just use the repo name for both, like we do in `.push_to_hub()`. I would also convert any spaces to underscores, etc. to ensure that the repo name is valid
* Imo we should remove description. Requires someone to come up with a description on the spot, which again, is an obstacle
* It would be cool if the theme builder detected that you were logged in via the HF CLI and then didn't ask you for your HF token
* Nit: the default version should be 0.0.1 to be consistent with the `.push_to_hub()` method

abidlabs:
@aliabid94 the Theme Builder is amazing! I have some suggestions above, but they are mostly small. Let's launch this!

gradio/themes/builder.py
aliabid94: Well this actually isn't where it's launched really, it's launched from  `themes/__init__.py` as:
```python
def builder(*args, **kwargs):
    from gradio.themes.builder import demo

    demo.launch(*args, **kwargs)
```
So users can pass any launch args

aliabid94:
Fixed all above requested changes, please rereview!

abidlabs (approved):
Just tried it out and the experience pushing the Hub was so smooth. Love it!! LGTM

freddyaboulton:
Awesome PR @aliabid94 ! Thanks for making the changes.

I accidentally didn't include my HF token when pushing to the hub because I thought it was optional if you were already signed in via cli. The error didn't get reported to the UI and the "Uploading...." text got displayed forever and didn't change back. I think we should catch any exceptions that could happen, report them to the user, and then change the button back in that case!

freddyaboulton (approved):
Thanks for addressing all the feedback @aliabid94 !

This is very useful 🔥

abidlabs:
Hi @tacediak currently, you cannot load an existing theme into the theme builder -- you can only create new ones from the existing themes. Agreed that this would be a handy feature but probably not one we'll release in the near future. We welcome any PRs to this effect.

PR #3669, opened 2023-03-28
Nit in ValueError

osanseviero:
# Description

I think there was a small typo in the wording here. Is it required to change the CHANGELOG.md for such small changes?


pngwn:
Yeah, anything affecting the library (and requiring a new release) gets a changelog entry but we can add that later!

Issue #2945, opened 2023-01-06
Subtitle support for gr.Video component


fcakyon:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
I am currently using a Gradio-based web application for video+subtitle labeling but I have to use Video+Textbox components to show the video and subtitle separately and cannot synchronize the subtitle as the video flows. It would be much better if subtitles could be rendered inside gr.Video component.

**Describe the solution you'd like**  
Adding subtitle argument (file path or URL to an `.srt` formatted subtitle file) to gr.Video component and subtitle text will be shown over the video.

```python
import gradio as gr

video = gr.Video(value='video.mp4', subtitle='subtitle_file.srt')
```

```python
import gradio as gr

video_path_or_url = Textbox()
video_subtitle_path_or_url = Textbox()
video_with_subtitle = gr.Video()

def set_video(file_path, subtitle_path):
     return {
        video_with_subtitle: gr.update(value=file_path, subtitle=subtitle_path)
    }

demo = gr.Interface(
    fn=set_video,
    inputs=[video_path_or_url, video_subtitle_path_or_url],
    outputs=video_with_subtitle
)

demo.launch() 
```

**Additional context**  
Desired output would look like this:
<img width="500" alt="demo" src="https://user-images.githubusercontent.com/34196005/211105468-52fad6d3-48c6-45d3-9424-72a57f4a7f3a.png">

abidlabs:
Great feature suggestion @fcakyon! 👍

pngwn:
Related discussion: https://github.com/gradio-app/gradio/discussions/2968

tomchang25:
Maybe you would like to check this [demo](https://github.com/tomchang25/gradio/blob/main/demo/video_with_caption/run.py)
```
import gradio as gr

css = (
    "footer {display: none !important;} .gradio-container {min-height: 0px !important;}"
)

with gr.Blocks(css=css) as demo:
    gr.Video(
        label="Video File Test",
        show_label=True,
        interactive=True,
        value="mp4/en.mp4",
        caption="tmp/en.vtt",
    )

demo.launch()

```
![image](https://user-images.githubusercontent.com/31534748/211816981-851cd802-8b11-4fbb-aedf-61be651e79ed.png)

But I just learned how to code svelte since last month, so there are lots of "dirty" code.
Any suggestion will be appreciate.

fcakyon:
[...]

PR #3673, opened 2023-03-28
Gradio subtitle

tomchang25:
# Description

Add subtitle feature to video, remain in draft until enough discussion.

Closes: #2945

# Checklist:

# A note about the CHANGELOG


demo/video_component/run.py
```diff
@@ -6,15 +6,25 @@
...
     outputs=gr.Video(),
     examples=[
-        [a],
-        [b],
-        [c],
+        [a, s1],
+        [b, s2],
+        [c, None],
     ],
 )
```
abidlabs, line 29: I wouldn't change this particular demo as it's used to illustrate the basic `Video` component in the docs. However, we could create a new `video_subtitles` demo

gradio/components.py
```diff
@@ -1977,6 +1979,8 @@ def __init__(
             raise ValueError(
                 f"Invalid value for parameter `source`: {source}. Please choose from one of: {valid_sources}"
             )
+
+        self.subtitle = self.format_subtitle(subtitle)
```
abidlabs, line 1983: Should be added to the docstring

gradio/components.py
```diff
@@ -2133,6 +2141,32 @@ def style(self, *, height: int | None = None, width: int | None = None, **kwargs
         )
         return self
 
+    @staticmethod
+    def format_subtitle(subtitle):
+        if subtitle is not None:
+            valid_subtitles = ("srt", "vtt")
+
+            if not subtitle.lower().endswith(valid_subtitles):
```
abidlabs, line 2149: More robust
```suggestion
            valid_extensions = (".srt", ".vtt")
            
            if not Path(subtitle).suffix in valid_extensions:
```

gradio/components.py
```diff
@@ -2133,6 +2141,32 @@ def style(self, *, height: int | None = None, width: int | None = None, **kwargs
...
+                raise ValueError(
+                    f"Invalid value for parameter `subtitle`: {subtitle}. Please choose from one of: {valid_subtitles}"
```
abidlabs, line 2151:
```suggestion
                    f"Invalid value for parameter `subtitle`: {subtitle}. Please choose a file with one of these extensions: {valid_subtitles}"
```

abidlabs:
Thanks @tomchang25 for this PR! I think adding support for subtitles for the `Video` component is a good idea. It's been requested by the community frequently and it's good for accessibility. 

I've reviewed the backend, generally looks good, left a few minor comments. My main concern is adding `pysub2` as a dependency. We'd like to limit additional dependencies in the gradio package. Is there a way to support `.srt` files without adding another dependency? If not, I think we should stick to whichever formats the browser supports. 

cc @aliabid94 @pngwn for your thoughts, and could you guys take a look at the Svelte code?

gradio/components.py
tomchang25: It's actually a private function and should not be called from outside. I will add docstring and change it to private.

demo/video_component/run.py
tomchang25: Done, should I clean the subtitle files from video_componment from github history or it's ok to remain this. 
(About 600 Bytes)

tomchang25:
Fix 
- revert video_componment demo and add another video_subtitles demo 
- format_subtitle docstring 
- clean valid_subtitles, ValueError code
- _subtitle declaration removed

> My main concern is adding pysub2 as a dependency. We'd like to limit additional dependencies in the gradio package.

I used pysub2 so I can easily support other format in the future.
I think I can rewrite it in a simple function because VTT and SRT converstion is pretty easy.
As for other format like xml may need much time to deal with, but since these two format cover about 90% application. 
It's may not a big problem.

In addition, I find a bug that subtitles will not reset if changing the video and subtitles but has not changed the type of screen mode.



I will check this issue in the next few days.

abidlabs:
Thanks @tomchang25 for addressing the issues above! Will take another look

aliabid94:
Thanks for the PR @tomchang25, I think the community will really appreciate this one.
My concern is with the API, in the the subtitle track is considered part of the component rather than part of the value itself. This is problematic because if a method returns a new value for the video without setting the subtitle track, it will actually use the previous subtitle track. The subtitle track is couple to an individual video, not the video component. 
I believe the best solution would to allow the Video component to accept a Tuple, in which case the second value would be the subtitle track. (@abidlabs any thoughts on this?)

Happy to take over and make these changes to get this PR across the finish line.

abidlabs:
Good catch @aliabid94. Agree that the proposed format is better, and it's kind of consistent to (images, alt_text) in the `gr.Chatbot` component, for example.

tomchang25:
Hello, @abidlabs

The Tuple version is now done. I believe most of the issues have been addressed. However, I am unable to fix the bug where the subtitles are not reset if the video is changed before resetting the subtitles. The subtitles will use the previous subtitle track in this case. For some unknown reason, if the video source is changed without setting the subtitle to None, it will cause an issue with the subtitle cannot be edited by the subtitle track component.

I'm not sure if this bug is caused by the browser or Svelte. Nonetheless, since this bug is not impacting the functionality significantly, it can be easily fixed by rescaling the video.

Perhaps we can create an issue post and merge it into the main branch first?

Thank you.

demo/video_component/run.py
abidlabs: Nw about the github history, 600 bytes is fine

abidlabs:
Hi @tomchang25 I tested and the demo did not work for me. There were some, I believe accidental errors like the `block` argument to `launch()`. Even after fixing those, when I run the demo, I get a blank white screen: 

image

Does the demo work for you locally? Maybe there were some local changes that you haven't pushed yet?

tomchang25:
Hi, @abidlabs

Thank you for your feedback. I have reviewed all the files and confirmed that there are no missing files. Additionally, I have cloned the pure branch, reinstalled it, and rebuilt the front-end, and it appears to be working fine on my end. However, I understand that the issue may be specific to my local environment, which is running Windows 10.

In any case, I have modified the demo to use gr.Interface. I hope this resolves the issue.

abidlabs:
Hmm interesting I'm testing on Windows as well. Can you try running this example and let me know if it works for you? 

```py
import gradio as gr

gr.Interface(lambda x:x, "video", "video").launch()
```

When I submit a video, I get the following error:

```

  File "c:\users\islam\dev\gradio-repos\gradio\gradio\routes.py", line 395, in run_predict
  File "c:\users\islam\dev\gradio-repos\gradio\gradio\blocks.py", line 1191, in process_api
  File "c:\users\islam\dev\gradio-repos\gradio\gradio\blocks.py", line 1039, in preprocess_data
  File "c:\users\islam\dev\gradio-repos\gradio\gradio\components.py", line 2038, in preprocess
    video = x[0]
KeyError: 0
```

tomchang25:
I have checked another PC that has never had Gradio installed before, and the components, subtitles, and Gr.Interface demos are all working fine.

Have you remembered to update the front-end? I made a significant change to the front-end value format from FileData | null to Array(FileData) | null. It seems like the front-end is still returning FileData.

abidlabs:
Hi @tomchang25 you're completely right, my bad! I was rebuilding the frontend, but the job was silently failing and didn't realize it! The original video demos as well as the demo that you added work well.

There is still an issue around serialization with the new format (you can try this by flagging a video with subtitles). I'll fix that and then I think we should be good to merge.

tomchang25:
Great, @abidlabs . I really appreciate you taking the time to troubleshoot and fix the issue. By the way, once you've finished merging everything, I'll create an issue post about the subtitle not resetting when changing the video source. Thanks again for your help!

tomchang25:
Thank you for keeping the work going for a few more hours. I've reviewed the component section and it looks much better now. Nice job! I'm glad to hear that the PR has been completed.

client/python/gradio_client/client.py
```diff
@@ -84,6 +84,7 @@ def __init__(
         if self.verbose:
             print(f"Loaded as API: {self.src} ✔")
 
+        print("self.src", self.src)
```
aliabid94, line 87: remove

gradio/components.py
```diff
@@ -2082,49 +2098,152 @@ def preprocess(self, x: Dict[str, str] | None) -> str | None:
...
+                - 'name': None
+                - 'data': Base64 encode the processed subtitle data.
+                - 'is_file': False
+            - If subtitle is None, returns (video, None).
+            - If both video and subtitle are None, returns None.
+        """
+
+        if y is None or y == [None, None] or y == (None, None):
+            return None
+        if isinstance(y, str):  # type: string
```
aliabid94, line 2125:
```suggestion
        if isinstance(y, str):
```
(unnecessary comment)

gradio/components.py
```diff
@@ -2082,49 +2098,152 @@ def preprocess(self, x: Dict[str, str] | None) -> str | None:
...
+            - 'name': None
+            - 'data': base64-encoded subtitle data.
+            - 'is_file': False
+        """
+
+        def srt_to_vtt(srt_file_path, vtt_file_path):
+            """Convert an SRT subtitle file to a VTT subtitle file"""
+            with open(srt_file_path, "r", encoding="utf-8") as srt_file, open(
+                vtt_file_path, "w", encoding="utf-8"
+            ) as vtt_file:
```
aliabid94, line 2210: several unnecessary / self-evident comments below

js/app/src/components/Video/Video.svelte
```diff
@@ -14,8 +14,8 @@
 	export let elem_id: string = "";
 	export let elem_classes: Array<string> = [];
 	export let visible: boolean = true;
-	export let value: FileData | null | string = null;
-	let old_value: FileData | null | string = null;
+	export let value: Array<FileData> | null = null;
```
aliabid94, line 17: value is specifically a [FileData, FileData] I believe, more accurate than just an Array of arbitrary size. Also applies to type hints below

aliabid94 (commented):
Great work @tomchang25! Let me look into this subtitle not changing issue.

aliabid94:
Also is the multiline supposed to render the subtitles in different lines? I see visible `\n` in the captions

Screen Shot 2023-04-11 at 1 58 41 PM

abidlabs:
> Also is the multiline supposed to render the subtitles in different lines? I see visible \n in the captions

Oh yeah I noticed that as well. Forgot to mention

aliabid94:
> ? I feel like new lines could clutter the UI

If a user intends to make something multiline, it should be. Looking into this.

aliabid94:
Actually its as simple as just adding an actual new line in the subtitle file, instead of using `\n`
Screen Shot 2023-04-11 at 2 24 38 PM

aliabid94:
Believe I fixed the subtitles issue, I think it was an issue with chrome if you changed video srcs for an element, the subtitles of the previous video would freeze onto the new video. Came up with a (hacky?) solution to clear the element between rerenders, @pngwn if you can give feedback when you get back on [this commit](https://github.com/gradio-app/gradio/pull/3673/commits/1a23747397d1f7a90aa5eba32b78059bcc7d2da0#diff-c3fca79de2c5ae38db6ce20bb95921436034ec5574fbc282795eb1f0f7640b3bR29), that'd be great.

Other than a few comments I left above, this PR lgtm!

aliabid94:
I'll go ahead and make the changes I requested.

aliabid94:
Merged! Thanks @tomchang25

abidlabs:
Thanks @tomchang25 for initiating this and thanks @aliabid94 for fixing the remaining issues. Solid PR!

PR #3692, opened 2023-03-29
node support for js client

### LINKED ISSUE #3495: Node js support for the JS client

I would like to be able to leverage the gradio client in my nodejs server to run inference on a private gradio api.

* Add [isomorphic-fetch](https://www.npmjs.com/package/isomorphic-fetch)
* Add [isomorphic-ws](https://www.npmjs.com/package/isomorphic-ws)
* Allow the client to pass a bearer token to be used in the headers of subsequent fetch requests and websocket connections

**Additional context**  
I've made these changes to the gradio client locally and it works like a charm. If ya'll are interested, I can pose a PR.

### LINKED ISSUE #3583: better build artifacts for the published js client
Right now we are just shipping the tyepscript as is, we need to transpile this to JS + include the dts type defs. We are already using vite in the project so we should probably just use that, although esbuild is faster and would be fine here.

pngwn:
Closes #3495. Closes #3583.

This PR adds node JS support for the client by dynamically importing a polyfill for web sockets in node. Creates and ESM bundle of the library to publish to NPM and generates typescript declaration files so users still get nice tooling support/ autocomplete.

Will need a publish to test fully but node scripts that use the client should work locally .

To test create a `test.js` file in `client/js` and add the following:

```js
import { client } from "./dist/index.js";

async function run() {
	const app = await client("gradio/hello_world_3", (d) => console.log(d));
	app.on("status", console.log).on("data", console.log);
	const p = await app.predict("/predict", { data: ["Peter", false, 0] });

	console.log(p);
}

run();
```

Then run it `node test.js`. It should work as expected.

This _does_ require node 18 which is the minimum node version I'd like us to support. node 18 makes some things _much_ nicer and reduces maintenance bruden on us. It is also the current active version of node, so I think it is a good mimimim to set.


pngwn:
@logesh2496 You can just import it normally. 

```js
```

The client itself will detect that it is in a node environment and run the correct code.

pngwn:
The client requires node 18+

Issue #3683, opened 2023-03-29
Some example Spaces in docs are broken


space-nuko:
### Describe the bug

Here are two I found so far:

https://huggingface.co/spaces/gradio/sentence_builder_main
https://huggingface.co/spaces/gradio/titanic_survival_main

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

1. Go to https://www.gradio.app/docs/main/#dropdown
2. Try the examples

### Logs

```shell
Runtime error
Space failed to start. Exit code: 1
Container logs:
```

### System Info

```shell
N/A
```

### Severity

annoying

abidlabs:
cc @aliabd

PR #3698, opened 2023-03-29
Fix broken spaces in docs

aliabd:
Fixes the two broken spaces in docs: [titanic_survival](https://huggingface.co/spaces/gradio/titanic_survival) and [sentence_builder](https://huggingface.co/spaces/gradio/sentence_builder). I already tested them on spaces and they work as expected. 

Closes: #3683


freddyaboulton (approved):
Thanks for the quick fix @aliabd 💪

Issue #2479, opened 2022-10-18
Expose `js` parameter and add it to documentation


abidlabs:
Currently, all `gradio.Blocks` event triggers include a special `_js` parameter that is used within the library for things like the Clear button in a `gradio.Interface`. We don't really have any docs around it since it is not intended for public use. However, there is growing interest in using it (for example, e.g. it allows #2173), so I think we should rename it to `js` and expose it as a regular parameter. 

I remember we had punted some discussion like:
* Do we really want users writing javascript functions as strings in Python or can we provide a better DX?
* What should the method signature of the javascript function be? @pngwn had some thoughts here

freddyaboulton:
This got done by @dawoodkhan82 in the v4 branch!

PR #3709, opened 2023-03-30
Docstring argument fixes

akx:
# Description

* Some of the docstrings for arguments were out of date c.f. the actual arguments.
  * For arguments I couldn't find descriptions for, I used `TODO: undocumented`; that's easy enough for someone to grep and fix.


abidlabs:
Thanks @akx we can fix these docstring issues and merge this. Just FYI, `_js` is intentionally undocumented because it is intended for internal use at this point, and it's API may change, until we close #2479

akx:
> Thanks @akx we can fix these docstring issues and merge this. Just FYI, `_js` is intentionally undocumented because it is intended for internal use at this point, and it's API may change, until we close #2479

The parameter is being used in the wild, which was exactly why I was confused about the lack of docstring...

Would you like me to remove that change from the PR, or maybe add "experimental, may change" or something to the doc?

abidlabs:
> The parameter is being used in the wild, which was exactly why I was confused about the lack of docstring...

Makes sense, I'll add some documentation which explicitly states that this is an experimental parameter. 

Btw can you allow maintainers to push to this PR? I'll fix the remaining docstrings

freddyaboulton:
Hi @akx ! This looks good to me. Can you re-open the PR with permission for outside commits so that @abidlabs can fix the remaining docstrings? Thank you.

akx:
> Btw @akx can you allow maintainers to push to this PR? I'll fix the remaining docstrings

Done. Sorry for the delay

abidlabs:
Thanks @akx! Fixing the docstrings now

abidlabs:
I'm running into some issues trying to push changes to this branch. Started a new PR with these changes (#3740)

PR #3710, opened 2023-03-30
Switch linting to Ruff

akx:
# Description

This PR switches the linting + import formatting for Gradio from flake8 + isort to [Ruff](https://github.com/charliermarsh/ruff).

Ruff is a lot faster than flake8 and isort, and also catches more errors; see some of the fixes done here, and I didn't even configure any [additional lint rules](https://beta.ruff.rs/docs/rules/).

Please see each individual commit and commit message for further detail.


abidlabs:
Thanks @akx for initiating this PR. Let me discuss with the other maintainers of this repo if we'd like to make this change. Just FYI we prefer if contributors raise major changes as issue so we can discuss there before opening a PR directly.

pyproject.toml
```diff
@@ -60,3 +60,27 @@ include = [
...
+ignore = [
+  "E501", # from scripts/lint_backend.sh
+  "E722", # from scripts/lint_backend.sh
+  "E731", # from scripts/lint_backend.sh
+  "F403", # from scripts/lint_backend.sh
+  "F541", # from scripts/lint_backend.sh
+]
+
+[tool.ruff.per-file-ignores]
+"demo/*" = [
```
freddyaboulton, line 78: In my opinion, it's fine if we don't lint the demo files

website/check_version.py
```diff
@@ -1,8 +1,7 @@
-import time
```
freddyaboulton, line 1: Also fine if we don't lint the website

abidlabs:
Agree with @freddyaboulton's suggestion about not linting the demos. Because the demos are code that we embed in our documentation and guides, there are times when we don't want to be following pep8 standards, e.g. because it might add too much whitespace. For the website, I don't know, maybe we _should_ actually be linting it -- it might help us catch bugs. WDYT @freddyaboulton @aliabd?

akx:
>  there are times when we don't want to be following pep8 standards, e.g. because it might add too much whitespace. 

`ruff` doesn't do formatting (yet), just linting. If you look at the changes enacted for the demos, they're pretty much just removals of unused imports and variables.

scripts/lint_backend.sh
```diff
@@ -1,7 +1,5 @@
 #!/bin/bash
 
 cd "$(dirname ${0})/.."
-
-python -m black --check gradio test client/python/gradio_client
-python -m isort --profile=black --check-only gradio test client/python/gradio_client
-python -m  flake8 --ignore=E731,E501,E722,W503,E126,E203,F403,F541 gradio test client/python/gradio_client --exclude gradio/__init__.py,client/python/gradio_client/__init__.py
\ No newline at end of file
+ruff gradio test client
+black --check gradio test client
```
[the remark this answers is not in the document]
freddyaboulton: Ah got it!

gradio/__init__.py
```diff
@@ -16,8 +16,8 @@
     Carousel,
     Chatbot,
     Checkbox,
-    Checkboxgroup,
     CheckboxGroup,
+    Checkboxgroup,
```
freddyaboulton, line 20: Let's get rid of these duplicate imports?

gradio/components.py
```diff
@@ -3073,7 +3063,6 @@ def update(
             "interactive": interactive,
             "visible": visible,
             "value": value,
-            "interactive": interactive,
```
freddyaboulton, line 3076: Lol nice catch by ruff

freddyaboulton (commented):
Thanks @akx ! This looks good to me - just left some minor comments. @abidlabs Any further thoughts before merging?

gradio/__init__.py
akx: I think `Checkboxgroup` is a legacy compatibility alias for `CheckboxGroup`, so getting rid of it would probably break some code out there.

pyproject.toml
```diff
@@ -60,3 +60,27 @@ include = [
...
+  "E402", # Demos may have imports not at the top
+  "E741", # Demos may have ambiguous variable names
+  "F405", # Demos may use star imports
+  "I", # Don't care about import order
+]
+"gradio/__init__.py" = [
```
freddyaboulton, line 84: We can get rid of this since you already added `__all__` right?

gradio/__init__.py
freddyaboulton: You're right!

pyproject.toml
akx: Nope - that was a different module (`client/python/gradio_client/__init__.py`). 

I felt like adding about 80 `__all__` entries would've been a bigger change that can be done separately. 😁

freddyaboulton (approved):
Thanks for the contribution @akx ! This looks good to me. Will leave open in case @abidlabs wants to make any further comments. Should be good to merge shortly

test/test_examples.py
```diff
@@ -258,9 +258,7 @@ def many_missing(a, b, c):
     @pytest.mark.asyncio
     async def test_caching_with_batch(self):
         def trim_words(words, lens):
-            trimmed_words = []
-            for w, l in zip(words, lens):
-                trimmed_words.append(w[:l])
```
abidlabs, line 263: Was this caught by `ruff` or did you change manually @akx?

abidlabs:
LGTM as well. It looks like `ruff` is catching some additional things that are good to fix. I'm curious if there are things that `ruff` is missing that would be caught by `isort`/`flake`? I don't think anything major and it looks like the other HF libraries are moving towards `ruff` as well, so I'm in favor of making this change

In terms of whether we should `ruff`:
* the demos: I was initially concerned about whitespace but this doesn't seem to have been affected, while other unnecessary imports seem to have removed so it looks like a positive change
* the website: Let's just make sure @aliabd, @aliabid94, and anyone else who works on the website is aware that now they'll have to run `format_backend.sh` to ensure that the website python files are linted

abidlabs:
Merging now, thanks for this nice PR @akx!

akx:
> * they'll have to run `format_backend.sh` to ensure that the website python files are linted

I could follow up with a PR that switches the formatting/linting from those scripts to [`pre-commit`](https://pre-commit.com/), if that helps :)

test/test_examples.py
akx: ruff was complaining about `l` being an ambiguous name, so while renaming, I fixed these up to the equivalent listcomp.

PR #3713, opened 2023-03-30
fix dropdowns, release 3.24

aliabid94:
Release 3.24. Fix Dropdowns I broke in theme builder. 

Basically textbox input wasnt working in multiselect. Now, run:
```python
import gradio as gr

def greet(a, b, c):
    return a, b, c

with gr.Blocks() as demo:
    drop1 = gr.Dropdown(["World", "Gradio", "You"], label="simple")
    drop2 = gr.Dropdown(["World", "Gradio", "You"], label="multi", multiselect=True)
    drop3 = gr.Dropdown(["World", "Gradio", "You"], label="custom", allow_custom_value=True)
    btn = gr.Button("Click me!")
    val1 = gr.Textbox(label="simple")
    val2 = gr.Textbox(label="multi")
    val3 = gr.Textbox(label="custom")
    btn.click(greet, [drop1, drop2, drop3], [val1, val2, val3])

demo.launch()
```

Note you can enter text in all dropdowns, but only the custom box will collect non-choice values entered.


freddyaboulton (approved):
Thanks @aliabid94 !

gradio/version.txt
```diff
@@ -1 +1 @@
-3.23.1b3
\ No newline at end of file
+3.24
```
freddyaboulton, line 1: Let's make this 3.24.0

abidlabs:
Everything looks good to me @aliabid94! Let's release

Issue #3684, opened 2023-03-29
Chatbot automatically add empty line


XiPotatonium:
### Describe the bug

Each time chatbot is updated, a new line break are automatically added to old multi-line messages.
It seems that a new `<br>` is added before `\n`

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

It can be reproduced in gradio demo.
https://gradio.app/creating-a-chatbot/
I also encounter this bug with my installed gradio 3.23.0

### Screenshot

![QQ截图20230329141020](https://user-images.githubusercontent.com/8765462/228442436-32bebd89-8d09-4b0c-bea2-579698ac37f3.png)

### Logs

```shell
N/A
```

### System Info

```shell
gradio==3.23.0
```

### Severity

annoying

abidlabs:
Can reproduce, we'll take a look! cc @dawoodkhan82

enbacoo:
I have the same problem~

HaoHoo:
Hoops I just want to add an issue for this... Same problem when I try to chatbot for GPT

abidlabs:
Hi folks! Sorry about that, we've fixed this issue in #3717 and a made a new release (gradio==3.24.1) which should be out soon with the fixes.

PR #3717, opened 2023-03-31
Fix chatbot newline issue

aliabid94:
Basically when the chatbot is being passed as a whole back and forth between the frontend and backend, everytime it goes from backend to frontend, the postprocess runs the markdown->html method on the messages. Part of that step is to convert `\n` into  `<br>` but it actually converts `\n` to `<br>\n`. As html, this is fine, but when it goes through the postprocess loop again, the `\n` keeps becoming another `<br>\n`. Manually converted `\n` to `<br>` to fix.

Fixes: #3684

Also releases a new patch version with the fix


abidlabs:
Thanks for the fix @aliabid94! Added a test and fixed up the changelog. Will merge after CI passes

PR #3736, opened 2023-04-03
Add root_url to serializers in gradio_client

freddyaboulton:
# Description

The root_url is not being set for the file-based serializers so the client can't download output files or videos.

Closes: # (issue)

# Checklist:

# A note about the CHANGELOG

### DIFF
--- a/client/python/gradio_client/serializing.py
+++ b/client/python/gradio_client/serializing.py
@@ -11,9 +11,6 @@ from gradio_client import utils
 
 
 class Serializable(ABC):
-    def __init__(self, root_url: str | None = None):
-        self.root_url = root_url
-
     @abstractmethod
     def input_api_info(self) -> Tuple[str, str]:
         """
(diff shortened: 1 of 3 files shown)


client/python/gradio_client/serializing.py
```diff
@@ -164,7 +167,10 @@ def serialize(
         """
         if x is None or x == "":
             return None
-        filename = str(Path(load_dir) / x)
+        if utils.is_valid_url(x):
```
freddyaboulton, line 170: There was a bug where urls were not being downloaded during pre-processing

client/python/gradio_client/serializing.py
```diff
@@ -11,6 +11,9 @@
 
 
 class Serializable(ABC):
+    def __init__(self, root_url: str | None = None):
```
freddyaboulton, line 14: I think the code is simpler if this is passed in as a root_url and it makes sense since each serializable will only have one root_url during it's lifetime. It does make it a bit awkward since `root_url` is also a parameter of the method.

abidlabs:
Thanks @freddyaboulton! I think it would be better if we stuck with the existing approach of passing in `root_url` into `deserialize()` as opposed to the constructor in order to avoid having two parameters that have the same approach. The approach in this PR could lead to some confusion / complicated debugging down the line. As far as I understand, it shouldn't be much more complicated -- instead of passing in `root_url` in line 521, you'd just do it in line 471.

freddyaboulton:
@abidlabs Good point - I made that switch!

client/python/gradio_client/serializing.py
abidlabs: Nice catch

abidlabs:
Awesome, thanks for the update @freddyaboulton! LGTM

abidlabs (approved):

PR #3740, opened 2023-04-03
Fix missing docstrings (new PR)

abidlabs:
Just redoing #3709 as the original PR wasn't allowing me to make changes. Mostly just adds the docstrings that @akx noted were missing. Also slightly changes how streaming() works internally to avoid code/docstring duplication, cc @aliabid94


abidlabs:
Actually, I removed the docstrings for `_js` because our docs generator explicitly ignores parameters beginning with underscores. We'll add the docstrings as part of #2479

abidlabs:
Otherwise, LGTM. Should be good to merge after CI passes

abidlabs:
Thanks for suggesting these fixes @akx!

PR #3743, opened 2023-04-04
Adds a pypi release action for the gradio python client

abidlabs:
Similar to the Gradio library, when `client/python/gradio_client/version.txt` is updated with a new version library, and the PR is merged to `main`, we will automatically release the updated library to pypi.


.github/workflows/deploy-client-pypi.yml
```diff
@@ -0,0 +1,40 @@
...
+      uses: actions/setup-python@v3
+      with:
+        python-version: '3.x'
+    - name: Check new pypi version
+      run: python client/python/scripts/check_pypi.py
+    - name: Build pypi package
+      run: |
+        client/python/scripts/build_pypi.sh
+    - name: Publish package
+      uses: pypa/gh-action-pypi-publish@27b31702a0e7fc50959f5ad993c78deac1bdfc29
```
freddyaboulton, line 36: What if we try this:

https://github.com/pypa/gh-action-pypi-publish#-master-branch-sunset-

abidlabs: Isn't that what we are already doing? We're providing the exact hash: 27b31702a0e7fc50959f5ad993c78deac1bdfc29

abidlabs:
Ok so I'm going to mark this open for review. Initially, I was testing and testing this on a fork, but the pypi package just would not be published. After a lot of headscratching, I'm realizing that that is probably because secrets don't work from forks. 

You can see that the GH action works successfully up until the final step of publishing on my fork: https://github.com/abidlabs/gradio/actions/runs/4610860974/jobs/8149851765

I'm pretty sure (or at least hope dearly) it will work when this gets merged into `main`.

.github/workflows/deploy-client-pypi.yml
```diff
@@ -0,0 +1,40 @@
...
+      with:
+        user: aabid93
+        password: ${{ secrets.PYPI_GRADIO_CLIENT_TOKEN }}
+        packages_dir: client/python/dist
```
freddyaboulton, line 40: I think you need `packages-dir` instead of `packages_dir`?

https://github.com/pypa/gh-action-pypi-publish#customizing-target-package-dists-directory

abidlabs: I actually thought the same but that parameter must have gotten renamed. The correct parameter (at least for this version of the GH action) is `packages_dir`. You can see the error when I try using `packages-dir` here: https://github.com/abidlabs/gradio/actions/runs/4603068339

client/python/scripts/build_pypi.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -e
+
+cd "$(dirname ${0})/.."
+
+python3 -m pip install build twine
```
freddyaboulton, line 6: We don't need twine anymore right?

freddyaboulton (approved):
Thanks @abidlabs ! LGTM

client/python/scripts/build_pypi.sh
abidlabs: Will remove!

abidlabs:
Thanks for the review and suggestions @freddyaboulton! Will merge once the CI passes and hopefully we will see a new release of the `gradio_client` library

PR #3754, opened 2023-04-04
When auth is enabled, if the HTTPS request is not used when accessing…

nyl1001:
… the UI remotely, the client will not pass the access-token parameter for the websocket request, which will cause the interface to report a 403 error. Therefore, in order to avoid such problems, a new https control parameter is added to expose to the foreground business, so that auth can be used normally in non-https scenarios.

# Description

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

Closes: # (issue)

# Checklist:

# A note about the CHANGELOG


nyl1001:
When auth is enabled, if the HTTPS request is not used when accessing the UI remotely, the client will not pass the access-token parameter for the websocket request, which will cause the interface to report a 403 error. Therefore, in order to avoid such problems, a new https control parameter is added to expose to the foreground business, so that auth can be used normally in non-https scenarios.

abidlabs:
Thanks for the PR @nyl1001. I believe @freddyaboulton has *just* merged in a PR for this issue here: https://github.com/gradio-app/gradio/pull/3735

Although the PR is a little different, I think it accomplishes the same thing, can you confirm? You can try it out by installing gradio from main as described here: https://gradio.app/docs/main

nyl1001:
> Thanks for the PR @nyl1001. I believe @freddyaboulton has _just_ merged in a PR for this issue here: #3735
> 
> Although the PR is a little different, I think it accomplishes the same thing, can you confirm? You can try it out by installing gradio from main as described here: https://gradio.app/docs/main

I think you should reject the solution #3735

abidlabs:
Can you elaborate on why the other PR will make https websocket cookies insecure?

nyl1001:
> Can you elaborate on why the other PR will make https websocket cookies insecure?

Yes. that solution means the cookie access-token is not necessary, the client just need pass access-token-insecure.

nyl1001:
> Can you elaborate on why the other PR will make https websocket cookies insecure?


I think it makes sense that we should leave the choice to the applicable party

nyl1001:
> Can you elaborate on why the other PR will make https websocket cookies insecure?


I think it makes sense that we should leave the choice to the applicable party

nyl1001:
> Can you elaborate on why the other PR will make https websocket cookies insecure?

Yes. that solution means the cookie access-token is not necessary, the client just need pass access-token-insecure. And the access-token-insecure will always be showed in HTTP or https request.

nyl1001:
> Can you elaborate on why the other PR will make https websocket cookies insecure?

Yes. that solution means the cookie access-token is not necessary, the client just need pass access-token-insecure. And the access-token-insecure will always be showed in HTTP or https websocket request.

nyl1001:
> Can you elaborate on why the other PR will make https websocket cookies insecure?

Yes. that solution means the cookie access-token is not necessary, the client just need pass access-token-insecure. And the access-token-insecure will always be showed to users in HTTP or https websocket request.

nyl1001:
> Can you elaborate on why the other PR will make https websocket cookies insecure?

Yes. that solution means the cookie access-token is not necessary, the client just need pass access-token-insecure instead. And the access-token-insecure will always be showed to all client users in HTTP or https websocket request.

abidlabs:
I see, much appreciated @nyl1001! We'll look into this soon.

nyl1001:
> I see, much appreciated @nyl1001! We'll look into this soon.

thanks～

abidlabs:
Hi @nyl1001! @freddyaboulton and I discussed this issue further, and we don't follow this statement:

> Yes. that solution means the cookie access-token is not necessary, the client just need pass access-token-insecure instead. And the access-token-insecure will always be showed to all client users in HTTP or https websocket request.

It seems to us that this PR and the earlier PR #3735 accomplish the same thing, the only difference is the name of the cookie: `access-token` vs. `access-token-unsecure`. 

You'll notice that in this PR:

```py
                    response.set_cookie(
                        key="access-token",
                        value=token,
                        httponly=True,
                    )
```
vs. 
```py
                response.set_cookie(
                    key="access-token-unsecure", value=token, httponly=True
                )
```
The only difference is the name of the cookie -- otherwise, the token is in both of them, and both of them are `httponly=True`. 

Furthermore, you introduce a parameter called `enable_https` but this is set to `True` in the constructor and never set otherwise, this doesn't seem to correct either.

abidlabs:
Hi @nyl1001 I'll close this PR for lack of follow up. If you still feel like #3735 is the wrong approach, I'd be interested in hearing your reasoning. Thanks!

PR #3757, opened 2023-04-04
Updates serializers for `Label` and `Chatbot`

abidlabs:
During testing, it was brought up that the client often returns string filepaths, which can be hard to deal with. In the case of certain components, like `Image`, there's no real alternative to files. However, for the `Label` and `Chatbot`, which are two components people are likely to use via API, it would be better if we didn't serialize into files. This leaves `Label` and `Chatbot` as dictionaries and lists respectively. 

This is a breaking change, but I think the breaking changes are unlikely to affect many users because of when `.serialize()` is actually used in practice:

* `serialize()` and `deserialize()` are used when caching examples or flagging data. However, this is an internal use case and someone upgrading to a later version of Gradio will not notice any difference
* `serialize()` and `deserialize()` are also used when a Gradio Space is loaded and used as an API. However, unless someone actually updates the Gradio version on a Space, they won't experience any breaking change. 

Also changes the order that the serializer for a given component isdetermined:
* Now we first look to see if there is a serializer defined in `COMPONENT_MAPPING`
* Then, we look to see if there is a serializer defined in the gradio app config

This has the advantage of allowing us to change serializers in a backward-compatible manner (i.e. the change will apply if you try to load a Space running an older version of Gradio)


freddyaboulton:
Won't this break flagging for Label/Chatbot (the latter being a rare case I guess)?

abidlabs:
Thanks @freddyaboulton I think it should work, but I could be wrong!

freddyaboulton:
@abidlabs I should have tested this a bit before commenting so I apologize about that!

Yes, you're right, that flagging should still work. My main concern was hitting the limit of the csv writer but I checked and it's 120k something characters which should be enough for all practical cases.

While testing, I found some bugs that I think we can take care of in this PR:
- Flagging doesn't work if the output is a Label or Chatbot - I get `AttributeError: 'dict' object has no attribute 'startswith'` from `sanitize_value_for_csv`
- After fixing that, I think the loading from cache when `cache_examples=True` is breaking for label output. The output is not showing up for me
label_flagging
- For some reason, each flagged entry is being recorded twice? This is my `log.csv` currently
image


BTW, I got this with the following demo

```python
import gradio as gr
import random

def big_label(num):
    return {f"class_{i}": random.random() for i in range(5)}

gr.Interface(big_label, gr.Number(), gr.Label(),
             examples=[0, 1, 2], cache_examples=True).launch()
```

On a separate note, I wonder if the better solution is to configure this at the client level? Like `deserialize=False` in `predict`?

abidlabs:
> On a separate note, I wonder if the better solution is to configure this at the client level? Like deserialize=False in predict?

I think this is a great idea. It applies to other data types as well. There are cases where users have requested images to be returned as base64, for example. And it would be 100% backwards compatible. The only tricky thing is updating the `view_api` documentation appropriately. Let me hold off on this PR until we see more demand for this and then we can add it.

PR #3760, opened 2023-04-04
Hot fix token in github action to publish py client to pypi

abidlabs:
The python client didn't successfully push to pypi. But that might be because I forgot to revert something I had changed during testing. Here's a quick reversion


freddyaboulton (approved):

abidlabs:
It worked @freddyaboulton! Thanks!! https://pypi.org/project/gradio-client/0.0.6b10/

PR #3761, opened 2023-04-04
Adds a changelog for the python client

abidlabs:
Now that we're getting ready for the release of the Python client, let's add a changelog.

This PR adds a changelog in a similar format to the main `gradio` library. Also modifies the changelog check GH action to check this changelog too. We don't have to merge this in soon but I just wanted to get the PR out

cc @aliabd let's get this on the website as well


abidlabs:
Converting to a draft because I think the best time to add this is actually when we are ready to release the client library

PR #3764, opened 2023-04-05
Document python client in gradio.app/docs

abidlabs:
TODOs:

- [x] Add remaining docstrings to python client -- @abidlabs 
- [x] Move documentation.py to client -- @abidlabs 
- [x] Rename .predict() to .submit() -- @abidlabs 
- [x] Add sections on the gradio.app/docs website for the python client -- @aliabd


abidlabs:
Also will rename `.predict()` to `.submit()` in this PR, pending the internal discussion here: https://huggingface.slack.com/archives/C02SPHC1KD1/p1680687086785429

aliabd:
@abidlabs does doing an editable install of gradio not install `gradio_client`? I have to also do an editable install of `gradio_client` to get this line to work:

`from gradio_client.documentation import document_cls, generate_documentation`

Will have to edit the website build docker file if this is the case

aliabd:
Added the client to the docs, you can see it here: https://www.gradio-staging.com/docs/main/#clients Lmk if you want me to make changes, and please check if anything is missing. 

The init parameters for Job aren't documented, and one of them is required (?), kind of confusing. I guess Job is only meant to be used as returned from submit? so maybe we don't document its parameters?

abidlabs:
This looks really nice @aliabd! 

 > The init parameters for Job aren't documented, and one of them is required (?), kind of confusing. I guess Job is only meant to be used as returned from submit? so maybe we don't document its parameters?

I think it's worth still documenting the parameters in the docstring to help anyone who is reading the codebase. I'll do that! But in the meantime, opening this up for review

freddyaboulton (approved):
This looks very nice @aliabd @abidlabs ! 

My only comments are:

- If we are going to document the init parameters of `Job`, let's also mention that users are not expected to instantiate the class themselves as that's handled by the client
- Would be helpful to give an example output of a status update for `Job.status()`
- Would be helpful to mention that you can iterate over a `job`'s outputs with `for o in job` to get the results as soon as they are ready!

Other than that, it looks great!!

abidlabs:
Thanks @freddyaboulton! Agree with all of those points, will make the changes.

PR #3765, opened 2023-04-05
Update view api page to use Python client

abidlabs:
TODOs:

- [x] Expose a route (`/info/`) that returns the view_info as a json -- @abidlabs 
- [x] Based on the json, build a nice View API page featuring the python client -- @aliabd 
- [x] Use Space API fetcher to get api info for older versions of Gradio -- @freddyaboulton


abidlabs:
@aliabd and @freddyaboulton, I added the `/info` route, and also added the ability to run the client without serialization. This should now be ready for your parts. Once we have all of the pieces in place, we should write some final tests, etc as well.

gradio/routes.py
```diff
@@ -251,6 +251,17 @@ def main(request: fastapi.Request, user: str = Depends(get_current_user)):
                         "the frontend by running /scripts/build_frontend.sh"
                     )
 
+        @app.get("/info/", dependencies=[Depends(login_check)])
```
freddyaboulton, line 254: Nit but I think having a query param for `raw: bool` would be more appropriate than two separate routes

freddyaboulton: Also `serialize: bool` may be a better name than raw for the concept of whether we should serialize/deserialize 😂

gradio/blocks.py
```diff
@@ -1173,14 +1173,80 @@ def getLayout(block):
...
-                block_config["info"] = {
-                    "input": list(block.input_api_info()),  # type: ignore
-                    "output": list(block.output_api_info()),  # type: ignore
-                }
+                block_config["info"] = block.api_info()
             config["components"].append(block_config)
         config["dependencies"] = self.dependencies
         return config
 
+    def get_api_info(self, raw=False):
```
freddyaboulton, line 1181: @abidlabs If we want the api fetcher to call `gr.load`, then we need `gr.load` to work for apps not hosted in spaces. Is that what we want to do? There was some hesitation about that before: https://huggingface.slack.com/archives/C02SPHC1KD1/p1667487172025779?thread_ts=1667401251.893519&cid=C02SPHC1KD1

gradio/routes.py
abidlabs: Agreed

abidlabs: Fixed

abidlabs:
The route part should be ready to build off of now @aliabd

gradio/blocks.py
freddyaboulton: @abidlabs Would it be possible to refactor the existing logic so that the api info can be fetched entirely from the config without having to instantiate a client or call `gr.external.from_spaces_blocks()`.

Cause I just realized that, as things stand now, the api fetcher would not work for apps running locally, but a workaround could be to send the config over to the server. That would work for local, public, private apps.

abidlabs: Maybe I'm misunderstanding, but even if I refactor the existing logic so that there's a way to get the api info without instantiating a client, that would only apply to new versions of Gradio. In fact, such a method does exist -- `Blocks.get_api_info()`. But the API fetcher needs to be able to go from an older version of Gradio -> API Info, and the only way that would be possible is to load the Space

abidlabs: Oh nvm I understand what you mean now

abidlabs: Okay yeah I think I can quicklyrewrite `get_api_info` to go from config to API info

freddyaboulton: On main, my understanding was that the logic for `Client.view_api` works for all versions of gradio? 

I'm suggesting we refactor that so that it's a standalone function/static method.

Let me know if I'm crazy lol

freddyaboulton: Lol sweet thanks @abidlabs

abidlabs: Okay fixed. So now `get_api_info` is a standalone method in `gradio.blocks` that goes from config to api info. Lmk if that works!

abidlabs: For testing purposes, you could even just copy paste the `gradio.blocks.get_api_info ()` function into your Space (rather than having to do a release of the library to use in the Space)

freddyaboulton: It’s all good - I’m uploading the wheel to s3 so we’re covered

freddyaboulton: I believe this checks for  `if not component.get("serializer"):` in the function make it so that the function will not return anything for older spaces

image

abidlabs: Ah good point, will fix tomorrow

abidlabs: Should be working now: 

image

gradio/routes.py
freddyaboulton: Nice!

gradio/blocks.py
freddyaboulton: Awesome thank you! Just pushed up the api fetching logic

abidlabs: Sweet! Will test it out

client/python/gradio_client/media_data.py
freddyaboulton, line 1: This is technically a breaking change - I wonder if we can keep for stability?

client/python/gradio_client/client.py
```diff
@@ -59,17 +59,20 @@ def __init__(
...
         max_workers: int = 40,
+        serialize: bool = True,
         verbose: bool = True,
     ):
         """
         Parameters:
             src: Either the name of the Hugging Face Space to load, (e.g. "abidlabs/whisper-large-v2") or the full URL (including "http" or "https") of the hosted Gradio app to load (e.g. "http://mydomain.com/app" or "https://bec81a83-5b5c-471e.gradio.live/").
             hf_token: The Hugging Face token to use to access private Spaces. Automatically fetched if you are logged in via the Hugging Face Hub CLI. Obtain from: https://huggingface.co/settings/token
             max_workers: The maximum number of thread workers that can be used to make requests to the remote Gradio app simultaneously.
+            serialize: Whether the client should serialize the inputs and deserialize the outputs of the remote API. If set to False, the client will pass the inputs and outputs as-is, without serializing/deserializing them. E.g. you if you set this to False, you'd submit an image in base64 format instead of a filepath, and you'd get back an image in base64 format from the remote API instead of a filepath.
```
freddyaboulton, line 70: Tested with images and `serialize=False` is working

image

requirements.txt
```diff
@@ -3,7 +3,7 @@ aiohttp
 altair>=4.2.0
 fastapi
 ffmpy
-gradio_client>=0.0.10
+gradio_client==0.1.2b1
```
freddyaboulton, line 6: Triple check this is the right version 😂

test/test_blocks.py
```diff
@@ -126,6 +127,7 @@ def fake_func():
             demo.load(fake_func, [], [textbox])
 
         config = demo.get_config_file()
+        print(config)
```
freddyaboulton, line 130: I think we can remove this print

client/python/gradio_client/client.py
```diff
@@ -492,19 +536,24 @@ def _space_name_to_src(self, space) -> str | None:
...
-            )
-        return config
+        r = requests.get(
+            urllib.parse.urljoin(self.src, utils.CONFIG_URL), headers=self.headers
+        )
+        if r.ok:
+            return r.json()
+        else:  # to support older versions of Gradio
+            r = requests.get(self.src, headers=self.headers)
+            # some basic regex to extract the config
```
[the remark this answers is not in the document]
freddyaboulton: I think we'll need to fetch config from the index file to support spaces with auth in the future

freddyaboulton (approved):
Thanks @abidlabs ! This looks good to me

requirements.txt
abidlabs: Good catch!

aliabid94 (approved):

abidlabs:
Thanks for the reviews @aliabid94 @freddyaboulton! Have addressed the feedback. Will release the `gradio_client` library first, and then merge in this PR when everything passes

Issue #3479, opened 2023-03-16
Whisper OpenAI - Microphone Audio Input


MichaelMares:
### Describe the bug

Hey, I noticed this from a couple of days ago. The microphone input (as audio) is now throwing an error with the OpenAI Whisper model. What used to work fine now requires an additional step. What needs to be done now is to convert the input into a .wav format. Not sure wether it is a Gradio bug or OpenAI changing the API but I thought I'd share my findings

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Original code that no longer works:

#Function that takes in the audio as an input and returns the transcript
def transcribe(audio):
    global messages
     
    audio_file = open(audio, "rb")
    transcript = openai.Audio.transcribe("whisper-1", audio_file)
    print(transcript)
...
ui = gr.Interface(
    fn=transcribe,
    inputs=gr.Audio(source="microphone", type="filepath", label = "What's on your mind?"),...

New function that works:

def transcribe_audio(audio_file):
    # Load the audio file and convert it to .wav format
    audio_segment = AudioSegment.from_file(audio_file).export("converted_audio.wav", format="wav")
    
    with open("converted_audio.wav", "rb") as audio:
        response = openai.Audio.transcribe('whisper-1', audio)
        
    return response["text"]

### Logs

```shell
Colab notebook detected. This cell will run indefinitely so that you can see errors and logs. To turn off, set debug=False in launch().
Note: opening Chrome Inspector may crash demo inside Colab notebooks.

To create a public link, set `share=True` in `launch()`.
/usr/local/lib/python3.9/dist-packages/gradio/processing_utils.py:239: UserWarning: Trying to convert audio automatically from int32 to 16-bit int format.
  warnings.warn(warning.format(data.dtype))
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/dist-packages/gradio/routes.py", line 393, in run_predict
    output = await app.get_blocks().process_api(
  File "/usr/local/lib/python3.9/dist-packages/gradio/blocks.py", line 1059, in process_api
    result = await self.call_function(
  File "/usr/local/lib/python3.9/dist-packages/gradio/blocks.py", line 868, in call_function
    prediction = await anyio.to_thread.run_sync(
  File "/usr/local/lib/python3.9/dist-packages/anyio/to_thread.py", line 31, in run_sync
    return await get_asynclib().run_sync_in_worker_thread(
[...]

PR #3770, opened 2023-04-05
Fix for default file name of recorded audio files

abidlabs:
When audio files were being recorded, the filepath would be "audio" instead of "audio.wav", which would lead to problems like: #3479 and https://github.com/gradio-app/gradio/issues/2768#issuecomment-1497976532

This is a quick fix for this. Thanks @aliabid94 for helping me debug

Test by running:

```py
import gradio as gr

gr.Interface(lambda x:x, gr.Audio(source="microphone", type="filepath"), "textbox").launch()
```

It should now include the `.wav` suffix.

Fixes: #3479


aliabid94 (approved):
lgtm

freddyaboulton (approved):
Tested and works! Thanks @abidlabs @aliabid94

abidlabs:
Thanks guys for the quick review! Will merge after CI passes

abidlabs:
Some Spaces flakiness is happening. I'm going to set the client tests to not be required for now (they shouldn't be required since they are all pretty flkay)

Issue #3729, opened 2023-04-02
How do I create upload button for doc/pdf/ csv in the blocks. we have documentation around input text but not uploading single file


pearl04:
- [ ] I have searched to see if a similar issue already exists.
Yes, already done that , cant find anything

**Is your feature request related to a problem? Please describe.**  
A clear and concise description of what the problem is.
I have been trying to create upload button for doc/pdf/ csv inside the gr.blocks , but getting lots of errors and there is no documentation around that. 

**Describe the solution you'd like**  
A documentation/code of how we can add a doc/pdf/csv upload button and launch it using gr.blocks

**Additional context**  
[Testing on docx data - Jupyter Notebook.pdf](https://github.com/gradio-app/gradio/files/11130975/Testing.on.docx.data.-.Jupyter.Notebook.pdf)

Yes, I have my current code, attaching that for reference.

abidlabs:
Hi @pearl04 have you tried:

```py
gr.UploadButton(file_types=[".pdf",".csv",".doc"])
```

pearl04:
Yeah I tried that :( 

Nothing is getting picked up . Could you please guide me whats wrong in my code please . So just reiterating codes I am using below (context is create a chatbot interface , that picks up say doc file and you can ask questions from it )

```
from pathlib import Path
from llama_index import download_loader
DocxReader = download_loader("DocxReader")
loader = DocxReader()
```

```
#build chatbot function to create index
def build_the_bot(input_text):
    documents = loader.load_data(file=Path(input_text))
    global index
    service_context = ServiceContext.from_defaults(llm_predictor=llm_predictor)
    index = GPTSimpleVectorIndex.from_documents(documents, service_context=service_context)
    return('Index saved successful!!!')

```

```
#preserve chat history and create response
def chat(chat_history, user_input):
    bot_response = index.query(user_input)
      #print(bot_response)
    response = ""
    for letter in ''.join(bot_response.response): #[bot_response[i:i+1] for i in range(0, len(bot_response), 1)]:
        response += letter + ""
        yield chat_history + [(user_input, response)]

```
        

```
#created block, added inout uploader button 
def upload_file(files):
    file_paths = [file.name for file in files]
    return file_paths
with gr.Blocks() as demo:
    gr.Markdown('# Q&A Bot with OpenAI Models')
    with gr.Tab("Input Text Document"):
        file_output = gr.File()
        upload_button=gr.UploadButton(file_types=[".pdf",".csv",".docx"])
[...]

PR #3771, opened 2023-04-05
Python Client: Upload files via /upload for File components 

freddyaboulton:
# Description

- Similar to the front-end client, filepaths for `File` components are first uploaded via `/upload` route
- Modified the `FileSerializable.serialize` method to not convert to base64 if the file was already uploaded. Determining  if the file was already uploaded because the data will be a dict.
- Noticed a bug where passing a url ending with `/` to `Client(src=...` would cause the routes to get messed up so using `urllib.parse.urljoin` to combine the routes in a more robust way

# Checklist:

# A note about the CHANGELOG


abidlabs:
Btw we should check if we need to rename `/upload` (see https://github.com/gradio-app/gradio/issues/3729#issuecomment-1497926221), cc @aliabid94

freddyaboulton:
Tests are erroring-out because my space can't build: https://huggingface.co/spaces/freddyaboulton/sentiment-classification

But this should be good for review!

freddyaboulton:
@abidlabs yea we should test if renaming the route will fix the colab issues but I think that should happen in a separate PR?

client/python/gradio_client/serializing.py
```diff
@@ -312,6 +314,7 @@ def deserialize(
 
 SERIALIZER_MAPPING = {cls.__name__: cls for cls in Serializable.__subclasses__()}
 SERIALIZER_MAPPING["Serializable"] = SimpleSerializable
+SERIALIZER_MAPPING["File"] = FileSerializable
```
abidlabs, line 317: Do we need this?

freddyaboulton: Yes I believe so. The config is returning `File` as the serializer name for File components since File components implement their own custom serialize method.

abidlabs: Oh interesting, I did not know they implemented their own. Looks like the UploadButton also implements its own.

abidlabs: For simplicity, can we just include the file size in `FileSerializable.serialize` as well, and then use `FileSerializable` in `File`? We'll still have to include this line for backwards compatibility but going forward it should be a little simpler

abidlabs: Actually, we'll _need_ to do this. Otherwise, if you use the `File` component via API, it won't include the 'size' parameter

freddyaboulton: Yea agreed - was just starting to think the same thing!

abidlabs:
> @abidlabs yea we should test if renaming the route will fix the colab issues but I think that should happen in a separate PR?

Sure sounds good. I was initially think we should do it before we launch the client publicly in case we need to rename the upload route, but actually we could just keep both routes.

freddyaboulton:
We'd have to keep both routes for backwards compatibility, no?

abidlabs:
One concern I have @freddyaboulton is that in several places, we check to see if the component_type is a "file". We also need include "upload_button", but also thinking ahead for community component, it might be a good idea to create a mixin called `LargeFileUploader` or something similar. If a component is of this type, then the `/upload` route is used. Of course, the client doesn't know about components, so we'd have to include this as a flag in the config as well for the client.

freddyaboulton:
That's a great point @abidlabs ! I wonder if long term if wouldn't be simpler (esp for custom components) and better for all components that inherit from FileSerializable to use the upload route? I'm just considering File in this PR cause that's what the front-end client does (although I completely forgot about UploadButton!)

abidlabs:
Yeah good point. I don't see any reason why other file based components don't use /upload, except maybe from a UI perspective (what to show while the file is uploading) -- cc @aliabid94. Perhaps this is a change we make as part of 4.0, in which case it'd be fine to just include `UploadButton` for now and file an issue for the 4.0 milestone

freddyaboulton:
@abidlabs I modified the code to upload files for UploadButton, added a test for the UploadButton case, and  filed #3779 !

abidlabs:
Thanks @freddyaboulton! We still need to do this, right? 

> For simplicity, can we just include the file size in FileSerializable.serialize as well, and then use FileSerializable in File? We'll still have to include this line for backwards compatibility but going forward it should be a little simpler

Otherwise, if you use the File component via the client, it won't include the 'size' key

freddyaboulton:
@abidlabs Thanks for the reminder! File and UploadButton now use FileSerializable and FileSerializable now includes the size. Also handled the multiple files case in FileSerializable

abidlabs:
Awesome thanks so much @freddyaboulton! Will review later tonight

freddyaboulton:
No rush, just remembered I need to update deserialize as well 😂

client/python/gradio_client/client.py
```diff
@@ -526,16 +531,51 @@ def _predict_resolve(self, *data) -> Any:
             return outputs[0]
         return outputs
 
+    def upload(self, file_paths: List[str]) -> List[str] | List[Dict[str, Any]]:
```
abidlabs, line 534: Consider calling `_upload` to designate this as a non-user-facing method

abidlabs:
LGTM @freddyaboulton besides the changes to deserialize and a nit above. Cool PR!

freddyaboulton:
Should finally be ready now @abidlabs ! Thanks for the all the reviews 🙌

abidlabs:
Hi @freddyaboulton I see an issue when `file_count="multiple"`. Basically this line in the `_upload()` method:

```py
        files = [("files", (Path(f).name, open(f, "rb"))) for f in file_paths]
```

is only designed to handle the case where a File/UploadButton component has a single file. However, if `file_count="multiple"`, then `f` here could actually be a list of files. 

You can see the issue when trying to do something like this:

```py
c = Client("gradio/zip_files")
c.predict(["lion.jpg"], fn_index=0)
```

freddyaboulton:
Good catch @abidlabs ! That should be fixed now!

abidlabs:
Thanks for the fix @freddyaboulton! Will review again soon

abidlabs:
LGTM, thanks for this great PR @freddyaboulton!

abidlabs (approved):

PR #3772, opened 2023-04-05
Dynamically get the API endpoint's parameter names

abidlabs:
This PR adds and documents a `.set_endpoint()` method to the `Client` class. This class can be used to set the API name once without having to repeatedly pass it into `.predict()`, which should avoid some code duplication.

The additional thing about `.set_endpoint()` is that it's also able get the parameter names corresponding to the API endpoint and make that the function signature of `client.predict()` and `client.submit()`. Which means that you get the ability to pass in keyword arguments like this:

```py
client = Client("gradio-tests/titanic-survival")
client.set_endpoint('/predict')
client.predict(age=100, sex="male", fare_british_pounds=100)
```

In interactive environments, such as jupyter notebooks, you'll even get IDE parameter name hinting and autocompletion

image

However, this last part does come with some limitations, so would appreciate feedback

**Limitations / considerations**
*  You don't get parameter name hinting and autocompletion if you write your entire script at once e.g. in a python file
*  You don't get typing (I looked into this but it's pretty much impossible with complex types such as `List[str]`)
*  You can no longer use `api_name` or `fn_index` in `predict()`/`submit()` after you've called `set_endpoint()`
*  I have to introduce `**kwargs` to `predict()` and `submit()` which are only used if `set_endpoint()`

### DIFF
--- a/client/python/gradio_client/client.py
+++ b/client/python/gradio_client/client.py
@@ -10,12 +10,13 @@ import uuid
 from concurrent.futures import Future, TimeoutError
 from datetime import datetime
 from threading import Lock
-from typing import Any, Callable, Dict, List, Tuple
+from typing import Any, Callable, Dict, List, Tuple, get_type_hints
 
 import huggingface_hub
 import requests
 import websockets
 from huggingface_hub.utils import build_hf_headers, send_telemetry
+from makefun import with_signature
 from packaging import version
 from typing_extensions import Literal
 
@@ -75,6 +76,7 @@ class Client:
         self.ws_url = utils.WS_URL.format(self.src).replace("http", "ws", 1)
         self.config = self._get_config()
         self.session_hash = str(uuid.uuid4())
+        self.default_fn_index = None
 
         self.endpoints = [
             Endpoint(self, fn_index, dependency)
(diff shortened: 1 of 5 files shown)


freddyaboulton:
Very interesting PR @abidlabs ! I don't like that we're overriding the default implementation of `predict` and `submit` and that we added **kwargs to `predict` as I think that will degrade the experience of users not using the `set_endpoint` api.

What if instead of overriding those methods, we dynamically create new methods, whose name is equal to the `api_name` given or `fn_{fn_index}`  if the route is not named. To avoid collisions with the `predict` method of the client, we can add to a different namespace.

So something like: 

```python
client = Client("freddyaboulton/sentiment-classification")
client.routes.classify(input_text="I love the gradio client")
client.routes.sleep()
```

But `client.predict("I love the gradio_client", api_name="/classify")` works as expected. 

Also we should make sure that whatever API additions we make are general enough to be ported to other languages. I'm not sure what you or I proposed can be implemented in typescript for example and I wouldn't want the APIs to be so different across languages.

abidlabs:
Very valid points @freddyaboulton. I like what you proposed a lot, but let's hold off on this experiment until we get some more traction around the gradio client and can validate the need for the routes. Closing for now

PR #3781, opened 2023-04-06
Py client verbose

abidlabs:
Small PR to add a verbose flag to the client and also print a message if a user checks the status of a job and there's a long queue

### DIFF
--- a/client/python/gradio_client/client.py
+++ b/client/python/gradio_client/client.py
@@ -649,7 +649,7 @@ class Job(Future):
         """
         Parameters:
             future: The future object that represents the prediction call, created by the Client.submit() method
-            communicator: The c ommunicator object that is used to communicate between the client and the background thread running the job
+            communicator: The communicator object that is used to communicate between the client and the background thread running the job
             verbose: Whether to print any status-related messages to the console
             space_id: The space ID corresponding to the Client object that created this Job object
         """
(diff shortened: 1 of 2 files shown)


client/python/gradio_client/client.py
```diff
@@ -635,15 +643,21 @@ def __init__(
...
         future: Future,
         communicator: Communicator | None = None,
+        verbose: bool = True,
+        space_id: str | None = None,
     ):
         """
         Parameters:
             future: The future object that represents the prediction call, created by the Client.submit() method
-            communicator: The communicator object that is used to communicate between the client and the background thread running the job
+            communicator: The c ommunicator object that is used to communicate between the client and the background thread running the job
```
freddyaboulton, line 652: Weird space between `c` and `ommunicator`

client/python/gradio_client/client.py
```diff
@@ -772,6 +786,12 @@ def status(self) -> StatusUpdate:
                 )
             else:
                 with self.communicator.lock:
+                    eta = self.communicator.job.latest_status.eta
+                    if self.verbose and self.space_id and eta and eta > 0:
```
freddyaboulton, line 790: I think `eta > 0` is too loose of a condition, no?

freddyaboulton (approved):
Looks good @abidlabs !

client/python/gradio_client/client.py
abidlabs: Oh sorry that was supposed to >30 lol

abidlabs: Changed it during test, will revert

abidlabs:
Thanks for the quick review @freddyaboulton! Fixed these two mistakes

PR #3791, opened 2023-04-08
Dicts as inputs with specified keys

jrhe:
# Description

Adds the ability to use dictionary inputs with specified keys. 

Current dict input support sets the keys to be the components of the inputs. This isn't great as it means that event handlers have to bring the input components into their scope to key into the dictionary or iterate over the dictionary to find the component they want, tightly coupling input components and event handlers. With this change event handlers can be independent of Gradio.

*WIP Branch - needs refactoring before merging.*

# Checklist:

Documentation update to follow once the code is in decent shape.

Issues:
- unclear if `input_to_key` should sit on `dependency` or the `BlockFunction`
- Two failing tests due to the extra value `input_to_key` on `dependency`
- General messiness in `Block` `set_event_trigger` and `Blocks` `preprocess_data` which might be improved by substituting `inputs_as_dict` for an input type variable which holds the type of the inputs passed in.

I'm not at all familiar with the codebase so wanted to stop here to get some input.

@abidlabs @aliabid94 @freddyaboulton

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 
 ## New Features:
 
-No changes to highlight.
+* Allow event listeners to accept a single dictionary as its argument, where they keys are user specified, and the values are the component values. This is set by passing passing input components a dictionary with specific key names instead of a list or set. [@jrhe](https://github.com/jrhe) in [PRX](https://github.com/gradio-app/gradio/pull/X)
 
 ## Bug Fixes:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 
 ## New Features:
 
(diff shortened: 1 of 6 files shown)


abidlabs:
Hi @jrhe thanks for opening up PR. The approach that you've suggested here --> using a dictionary to map parameter names to components seems reasonable to me. @freddyaboulton @aliabid94 do you see any issues with this approach?

jrhe:
@abidlabs That sounds great, I'll try and do it over the next few days. I have had a few more ideas surrounding this so I think it makes a lot of sense to discuss an approach first. 

I think the dicts are a specific case. In general, I think making the parameters able to take any inputs a function would take is beneficial, to allow more expressive callback functions. How far we can get to this I'm not sure.

A particular use case of this came up recently in an app I'm building. The app has two sets of input text boxes of variable length. Currently, in my callback inputs, I have to pass in the number of inputs currently active in each set, the max number of inputs in each set, and then each set of input text boxes. In the callback function, I then have to manually split the positional args into the two sets and then split out those that are currently active. Ideally, I'd just pass in the two arrays of inputs. Will give an example in the new issue.

abidlabs:
Very interesting, thanks so much @jrhe! Looking forward to the discussion, I will close this PR in the meantime

PR #3809, opened 2023-04-10
Allow client to be duplicated from an existing Space

abidlabs:
Adds and documents a `Client.duplicate_from()` factory method

Not sure if there's a good way to test this without revealing write-access keys which is not a good idea, Lmk if you have suggestions!

Closes: #3804


aliabid94:
I see, makes sense to have a separate constructor since you'll also want to be able to send Space settings (GPU, public/private, sleep inactivity duration, etc.).

abidlabs:
Hmm looks like some tests are failing, will check when I’m back at the computer

abidlabs:
Will add some tests using mocks, but otherwise, this should be ready to review @aliabid94 @freddyaboulton!

client/python/gradio_client/client.py
```diff
@@ -81,14 +85,20 @@ def __init__(
                 )
             self.space_id = src
         self.src = _src
-        if self.verbose:
-            print(f"Loaded as API: {self.src} ✔")
         state = self._get_space_state()
+        if state == utils.BUILDING_RUNTIME:
```
aliabid94, line 89: maybe add a link to logs? (or stream them in verbose mode?)

client/python/README.md
```diff
@@ -53,6 +53,26 @@ from gradio_client import Client
...
+
+client = Client.duplicate_from("abidlabs/whisper") 
+client.predict("audio_sample.wav")  
+
+>> "This is a test of the whisper speech recognition model."
+```
+
+If you have previously duplicated a Space, re-running `duplicate_from` will *not* create a new Space. Instead, the Client will attach to the previously-created Space. So it is safe to re-run the `Client.duplicate_from` method multiple times. 
+
+**Note:** if the original Space uses GPUs, your private Space will as well, and your Hugging Face account will get billed based on the price of the GPU. To minimize charges, your Space will automatically go to sleep after 1 hour of inactivity. You can also set the hardware using the `hardware` parameter of `duplicate()`.
```
aliabid94, line 74: parameter of `duplicate_from()` (though I think calling the method `duplicate` would be fine too)

client/python/gradio_client/client.py
```diff
@@ -110,6 +120,90 @@ def __init__(
...
+            to_id: The name of the new Hugging Face Space to create, e.g. "whisper-duplicate". Do not add your Hugging Face username. If not provided, the new Space will be named "{you_hf_username}/{from_id}".
+            hf_token: The Hugging Face token to use to access private Spaces. Automatically fetched if you are logged in via the Hugging Face Hub CLI. Obtain from: https://huggingface.co/settings/token
+            private: Whether the new Space should be private (True) or public (False). Defaults to True.
+            hardware: The hardware tier to use for the new Space. Defaults to the same hardware tier as the original Space. Options include "cpu-basic", "cpu-upgrade", "t4-small", "t4-medium", "a10g-small", "a10g-large", "a100-large", subject to availability.
+            secrets: A dictionary of (secret key, secret value) to pass to the new Space. Defaults to None.
+            max_workers: The maximum number of thread workers that can be used to make requests to the remote Gradio app simultaneously.
+            verbose: Whether the client should print statements to the console.
+        """
+        # Duplicate a Hugging Face Space
```
aliabid94, line 157:
```suggestion
```

client/python/gradio_client/client.py
```diff
@@ -110,6 +120,90 @@ def __init__(
...
+            )
+        if to_id:
+            space_id = huggingface_hub.get_full_repo_name(to_id, token=hf_token)
+        else:
+            space_id = huggingface_hub.get_full_repo_name(
+                from_id.split("/")[1], token=hf_token
+            )
+        try:
+            huggingface_hub.get_space_runtime(space_id, token=hf_token)
+            if verbose:
```
aliabid94, line 172: btw instead of having `if verbose` everywhere, you can use the builtin logger which respects the set verbosity

aliabid94 (approved):
lgtm

abidlabs:
Thanks for the review and suggestions @aliabid94! Will fix and merge in soon

client/python/README.md
abidlabs: Yeah actually `duplicate` is shorter and more grammatically correct. Will change everywhere

client/python/gradio_client/client.py
```diff
@@ -110,6 +120,90 @@ def __init__(
...
+        Face Hub CLI.
+
+        The new Space will be private by default and use the same hardware as the original
+        Space. This can be changed by using the `private` and `hardware` parameters. For
+        hardware upgrades (beyond the basic CPU tier), you may be required to provide
+        billing information on Hugging Face: https://huggingface.co/settings/billing
+
+        Parameters:
+            from_id: The name of the Hugging Face Space to duplicate, e.g. "abidlabs/whisper".
+            to_id: The name of the new Hugging Face Space to create, e.g. "whisper-duplicate". Do not add your Hugging Face username. If not provided, the new Space will be named "{you_hf_username}/{from_id}".
```
freddyaboulton, line 149: I got confused by the docstring here. Are you able to duplicate into to an org?

abidlabs: Yes, if you provide the HF_TOKEN corresponding to the org. What could make this clearer?

freddyaboulton: It says

> The name of the new Hugging Face Space to create, e.g. "whisper-duplicate". Do not add your Hugging Face username.

Which made me think `gradio/whisper-duplicate` was not a valid value as it seems it only expects the name after the `/`

abidlabs:
Just added a default timeout of 300 seconds to any Space that is duplicated btw

client/python/gradio_client/client.py
abidlabs: Yes, that's correct. You should provide `whisper-duplicate`, not `gradio/whisper-duplicate`

abidlabs: The username is automatically inferred from the HF_TOKEN

abidlabs: I can change it so that you can provide one but it is ignored

freddyaboulton: I think right now you are not able to clone to an org you are a member of?
Because the token will correspond to my personal namespace and not the one for the org I want.

Using the token for the org won't work because you need a personal token to create new repos

image

abidlabs: Oh interesting, was not aware of that. In that case, let me change this

abidlabs: Pushed a fix

abidlabs:
Should be ready for merge if there are no objections

abidlabs:
Thanks for the review and feedback @aliabid94 and @freddyaboulton!

client/python/gradio_client/client.py
```diff
@@ -110,11 +120,101 @@ def __init__(
...
+                f"Could not find Space: {from_id}. If it is a private Space, please provide an `hf_token`."
+            )
+        if to_id:
+            if "/" in to_id:
+                to_id = to_id.split("/")[1]
+            space_id = huggingface_hub.get_full_repo_name(to_id, token=hf_token)
+        else:
+            space_id = huggingface_hub.get_full_repo_name(
+                from_id.split("/")[1], token=hf_token
+            )
```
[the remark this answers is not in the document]
Wauplin: Here is an example of how I would do it to avoid multiple HTTP calls before duplicating the Space (hence speeding up the process):

```py
import huggingface_hub
from huggingface_hub.utils import RepositoryNotFoundError, HfHubHTTPError
from typing import Dict


@classmethod
def duplicate(
    cls,
    from_id: str,
    to_id: str | None = None,
    hf_token: str | None = None,
    private: bool = True,
    hardware: str | None = None,
    secrets: Dict[str, str] | None = None,
    sleep_timeout: int = 5,
    max_workers: int = 40,
    verbose: bool = True,
) -> "Client":
    # Duplicate space
    try:
        space_url = huggingface_hub.duplicate_space(from_id=from_id, to_id=to_id, token=hf_token, private=private)
        if verbose:
            print(f"Created a duplicate of {from_id} for your own use: {space_url} 🤗")
        is_new = True
    except RepositoryNotFoundError:
        raise ValueError(f"Could not find Space: {from_id}. If it is a private Space, please provide an `hf_token`.")
    except HfHubHTTPError as e:
        if e.response.status_code == 409:
            if verbose:
                print(f"Using your existing Space: {space_url} 🤗")
            is_new = False
        elif e.response.status_code == 403:
            raise ValueError(f"Cannot duplicate to {to_id}: you don't have permissions for this namespace.") from e
        elif e.response.status_code == 401:
            raise ValueError(f"Please make sure you provided a valid write-token.") from e
        else:
            raise

    space_id = space_url.repo_id

    # Set timeout
    if is_new:
        utils.set_space_timeout(space_id, hf_token=hf_token, timeout_in_seconds=sleep_timeout * 60)
        if verbose:
            print(f"Sleep timeout has been set to {sleep_timeout}min 🤗")

    # Set hardware
    current_info = huggingface_hub.get_space_runtime(space_id, token=hf_token)
    current_hardware = (
        current_info.hardware or current_info.requested_hardware or huggingface_hub.SpaceHardware.CPU_BASIC
    )
    if hardware is not None and current_hardware != hardware:
        huggingface_hub.request_space_hardware(space_id, hardware)
        print(
            f"-------\nNOTE: this Space uses upgraded hardware: {hardware}... see billing info at"
            " https://huggingface.co/settings/billing\n-------"
        )

    # Set secrets
    if secrets is not None:
        for key, value in secrets.items():
            huggingface_hub.add_space_secret(space_id, key, value, token=hf_token)
        if verbose:
            print(f"Secrets have been set for {', '.join(secrets.keys())} 🤗")
    if verbose:
        print()

    return cls(space_id, hf_token=hf_token, max_workers=max_workers, verbose=verbose)
```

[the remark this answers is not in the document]
Wauplin: Some notes about :arrow_up: 
- When checking the space hardware I think you should check both "current_hardware" and "requested_hardware". If the Space is re-building on a new hardware but it's not yet completed, there no need to re-request the hardware.
- You can use `huggingface_hub.SpaceHardware.CPU_BASIC` enum instead of `"cpu-basic"`
- Adding secrets to a Space will trigger a re-build. I _think_ it will be the case no matter if the secret changed or not. So here it's a bit annoying since it means that if secrets are used the Space will be re-built each time you instantiate the client. I don't have a proper solution for that since it's not possible to retrieve the existing secret values from the API.
    - maybe a solution is to set the secrets only on new spaces. If the space already exists, print a message to the user that secrets must be set manually
    - also possible to have a `reset_secrets`  boolean parameter but I'm less of a fan
- to please mypy, return type "Client" would be nice :)

Wauplin (commented):
Hey @abidlabs , I'm sorry I did not review on time this PR - I just realized it existed when you pinged me on it :grimacing: 

I know the PR is already merged but I think the implementation can be improved, especially to:
- reduce # of HTTP calls to the Hub (+cleaner logic)
- raise an error if sleeptime is not set correctly

client/python/gradio_client/utils.py
```diff
@@ -18,23 +18,24 @@
...
+import huggingface_hub
 import requests
 from websockets.legacy.protocol import WebSocketCommonProtocol
 
 API_URL = "/api/predict/"
 WS_URL = "/queue/join"
 UPLOAD_URL = "/upload"
 RESET_URL = "/reset"
-DUPLICATE_URL = "https://huggingface.co/spaces/{}?duplicate=true"
+SPACE_URL = "https://hf.space/{}"
```
Wauplin, line 29: not needed if you use `RepoUrl` (see above)

abidlabs:
❤️ thanks for all of the great feedback @Wauplin! I'll create a separate PR to include these fixes

client/python/gradio_client/client.py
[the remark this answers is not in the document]
abidlabs: In the code example you provided @Wauplin, `space_url` would not be defined if the Space already exists in the user's namespace. If the Space already exists, is it possible to get the name of the Space in the user's namespace without using `get_full_repo_name()`?

client/python/gradio_client/client.py
```diff
@@ -110,11 +120,101 @@ def __init__(
...
+            utils.set_space_timeout(
+                space_id, hf_token=hf_token, timeout_in_seconds=sleep_timeout * 60
+            )
+            if verbose:
+                print(f"Created new Space: {utils.SPACE_URL.format(space_id)}")
+        current_info = huggingface_hub.get_space_runtime(space_id, token=hf_token)
+        current_hardware = current_info.hardware or "cpu-basic"
+        if hardware is None:
+            hardware = info.hardware
```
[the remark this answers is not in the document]
abidlabs: This is referring to the original Space's info from line 160. Renaming to make clearer

PR #3836, opened 2023-04-13
HighlightedImage component

aliabid94:
Created a new component called `HighlightedImage`. Allows user to overlay masks on top of an image to highlight regions. Masks can be rectangles (object detection) or 0-1 pixel masks (image segmentation). See demo below for an example:

Recording 2023-04-14 at 12 54 37

Also made some small visual fixes to HighlightedText


CHANGELOG.md
```diff
@@ -43,6 +43,26 @@ with gr.Blocks() as demo:
     gr.Video(("video.mp4", "captions.srt"))
 ```
 
+### HighlightedImage Component
```
abidlabs, line 46: Suggestion that I also put in Slack: how about `AnnotatedImage` component?

CHANGELOG.md
```diff
@@ -43,6 +43,26 @@ with gr.Blocks() as demo:
...
+
+```python
+with gr.Blocks() as demo:
+    img = gr.Image()
+    img_section = gr.HighlightedImage()
+    def mask(img):
+        top_left_corner = [0, 0, 20, 20]
+        random_mask = np.random.randint(0, 2, img.shape[:2])
+        return (img, [(top_left_corner, "left corner"), (random_mask, "random")])
+    img.change(highlight_handler, img, img_section)
```
freddyaboulton: nit: highlight_handler not defined - I think you mean `mask`

js/app/src/components/HighlightedImage/HighlightedImage.svelte
```diff
@@ -0,0 +1,174 @@
...
+					/>
+				{/each}
+			</div>
+			{#if show_label}
+				<div class="legend">
+					{#each _value ? _value[1] : [] as [file, label], i}
+						<!-- svelte-ignore a11y-click-events-have-key-events -->
+						<div
+							class="legend-item"
+							style="background-color: {_value
```
freddyaboulton: This is very neat. I think users will want to customize their colors (match branding/theme). I think we should add `color_map`? Or maybe the third element in the tuple is the desired color?

abidlabs:
Really cool tweak to `HighlightedText` @aliabid94!

Recording 2023-04-14 at 11 21 00

aliabid94:
> Really cool tweak to HighlightedText

That's @pngwn's work I believe! I took the inspiration to this component. The fixes to HighlightedText were just some spacing stuff.

CHANGELOG.md
aliabid94: Done

gradio/components.py
```diff
@@ -3930,6 +3930,172 @@ def style(
...
+
+    def __init__(
+        self,
+        value: Tuple[
+            np.ndarray | _Image.Image | str,
+            List[Tuple[np.ndarray | Tuple[int, int, int, int], str]],
+        ]
+        | None = None,
+        *,
+        show_legend: bool = True,
```
abidlabs, line 3951: I don't think this parameter has any effect. Setting it to `False` still displayed the legend

aliabid94: fixed

gradio/components.py
```diff
@@ -3930,6 +3930,172 @@ def style(
         return self
 
 
+@document("style")
+class AnnotatedImage(Selectable, IOComponent, JSONSerializable):
```
abidlabs, line 3934: Considering adding "Annotatedimage" as an alias for "AnnotatedImage" like we do for "Highlightedtext = HighlightedText" at the bottom of this file

aliabid94: done

CHANGELOG.md
```diff
@@ -3,7 +3,25 @@
...
+
+```python
+with gr.Blocks() as demo:
+    img = gr.Image()
+    img_section = gr.AnnotatedImage()
+    def mask(img):
+        top_left_corner = [0, 0, 20, 20]
+        random_mask = np.random.randint(0, 2, img.shape[:2])
+        return (img, [(top_left_corner, "left corner"), (random_mask, "random")])
+    img.change(highlight_handler, img, img_section)
```
abidlabs, line 20:
```suggestion
    img.change(mask, img, img_section)
```

freddyaboulton (commented):
This looks very good to me @aliabid94 🔥 

Left a comment about letting users control the colors for the classes.

Also, we need to modify `Dataset` so that this component can be used in an example, no?

abidlabs (approved):
Really nice component @aliabid94! Tested in Blocks and Interface, as well serialization / deserialization as well and it works too.

cc @hysts this could make for some really cool demos

abidlabs:
> Also, we need to modify Dataset so that this component can be used in an example, no?

I don't think we need to since this is strictly an output component, so it wouldn't ever appear in a `Dataset` (similar to `Label`, for example)

freddyaboulton:
Sounds good - let's not add to dataset unless there is user need! I don't think the distinction between output and input component makes sense though, you can pass any component as input and output to an event handler.

freddyaboulton:
Also, something to keep in mind for 4.0 is giving developers access to pydantic-style data classes for the tuple-typed components, e.g. Code, AnnotatedImage. Dealing with a class is so much nicer than a complex tuple and it makes it easier to add stuff to the interface.

Between these two functions, I think the second is easier to read.  (`img` is supposed to be an AnnotatedImage.)

```python
    def get_selected_label(img, evt: gr.SelectData):
        return [tup[1] for tup in img[1]][evt.index]
    
    def get_selected_label(img: AnnotatedImage, evt: gr.SelectData):
        return [select.label for select in img.selections][evt.index]
```

aliabid94:
Added support for color_map parameter via `.style`! Will merge this in soon if there's not more comments

js/app/src/components/HighlightedImage/HighlightedImage.svelte
aliabid94: Done

freddyaboulton (approved):
You're the bomb @aliabid94 !

Issue #3795, opened 2023-04-09
Event callbacks throw warning when callback is a class method


AscendantAI-Robbie:
### Describe the bug

When the Gradio components are wrapped withing a class and the callback method is part of the class the check_function_inputs_match function raises a warning stating that the inputs don't match. This is likely because "self" is being counted as an argument.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

import gradio as gr
from PIL import Image

class Test:
    def __init__(self):
        image = Image.new(mode='RGB', size=(1024, 1024))
        self.gallery = gr.Gallery(value=[image] * 6)
        self.selected = gr.Number(show_label=False)

        self.gallery.select(self.get_select_index, None, self.selected)

    def get_select_index(self, evt: gr.SelectData):
        return evt.index

with gr.Blocks() as demo:
    test = Test()
demo.launch()

### Screenshot

venv\lib\site-packages\gradio\utils.py:874: UserWarning: Expected 1 arguments for function <bound method UIArtistView.select_image of <ui.art_ui.UIArtistView object at 0x000002488FC119C0>>, received 0.
  warnings.warn(

venv\lib\site-packages\gradio\utils.py:878: UserWarning: Expected at least 1 arguments for function <bound method UIArtistView.select_image of <ui.art_ui.UIArtistView object at 0x000002488FC119C0>>, received 0.
  warnings.warn(

### Logs

```shell
N/A
```

### System Info

```shell
gradio==3.24.1
Windows10
Localhost
firefox
python3.10.6
```

### Severity

annoying

freddyaboulton:
Would you be interested in contributing a PR to fix this @AscendantAI-Robbie ?

yiyuezhuo:
I had taken a look and found it's due to `SelectData` is not properly ignored:

```python
    def is_special_typed_parameter(name):
        from gradio.helpers import EventData
        from gradio.routes import Request

        """Checks if parameter has a type hint designating it as a gr.Request or gr.EventData"""
        is_request = parameter_types.get(name, "") == Request
        # use int in the fall-back as that will always be false
        is_event_data = issubclass(parameter_types.get(name, int), EventData)  # * False but it's expected to be True
        return is_request or is_event_data
```

And it's due to `parameter_type` is set to `{}`

```python
    parameter_types = typing.get_type_hints(fn) if inspect.isfunction(fn) else {}  
```

[...]

PR #3837, opened 2023-04-13
Fix false positive warning in `check_function_inputs_match`

yiyuezhuo:
# Description

* When non-function callable objects (e.x. methods and objects with `__call__` defined) is used with `EventData`, false positive warnings are raised. See #3795. It can be fixed by relaxing the function checking.
* A new `get_type_hints` is added with some tests.
* #3795 is fixed.
* No additional dependencies 

Closes: #3795 

# Checklist:


test/test_utils.py
```diff
@@ -587,3 +590,51 @@ def test_abspath_no_symlink(self):
...
+        class C:
+            def f(self, s: str, evt: EventData):
+                return s
+
+        def f(s: str, evt: EventData):
+            return s
+
+        test_objs = [F(), C().f, f]
+
+        with warnings.catch_warnings():
```
freddyaboulton, line 636: This test isn't testing anything I thin. Is the intention to check that no warnings are being emitted?

freddyaboulton (commented):
Looks good @yiyuezhuo ! Thanks for the contribution. Left a comment about the test. Please add a line to the `CHANGELOG` so we can properly credit you with this change.

test/test_utils.py
yiyuezhuo: Yes, this test will fail if the previous code is used.

freddyaboulton: Should be good to go once the CHANGELOG is fixed

freddyaboulton (approved):
Thanks @yiyuezhuo ! Will merge after CI passes

Issue #3856, opened 2023-04-14
Updating `DataFrame` will invoke an unexpected `select` event


yiyuezhuo:
### Describe the bug

Once DataFrame's `select` is invoked once, updating to the `DataFrame` will invoke an unexpected `select` event.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```python
import pandas as pd
import gradio as gr

df = pd.DataFrame([{"a":1, "b":2}, {"a":2, "b":3}])

with gr.Blocks() as demo:
    gr_df = gr.DataFrame(df)
    
    button = gr.Button("Update DataFrame")
    
    def select(evt:gr.EventData):
        print("select")
    def click():
        df.iloc[0]["a"] += 1
        return df
        
    gr_df.select(select, [], [])
    button.click(click, [], [gr_df])

demo.launch()
```

- Select a cell in the DataFrame and a `select` is shown.
- Click the button, but `select` is also shown unexpectedly.

PS: If DataFrame has never been selected, the updating will not invoke the `select` event, as expectedly. If the updating just return the identical value, the `select` is not invoked as well.

### Screenshot

https://user-images.githubusercontent.com/12798270/232085043-0ff4fe69-0c94-4887-b51d-ac42ddf5e3db.mp4

### Logs

```shell
No Error but the behavior is not expected.
```

### System Info

```shell
3.25.1b1
Windows
Edge 112.0.1722.39
```

### Severity

serious, but I can work around it

yiyuezhuo:
Looks like that it's due to variable `selected` is not reset to "unselected state":

```javascript
	$: {
		if (selected !== false) {
			const loc = selected.split("-");
			const row = parseInt(loc[0]);
			const col = parseInt(loc[1]);
			dispatch("select", { index: [row, col], value: data[row][col].value });
		}
	}
```

abidlabs:
Thanks @yiyuezhuo for creating this issue and finding the problem. Would you be able to create a PR to fix this? We're happy to review

PR #3861, opened 2023-04-14
Fixes data updating invoke the `select` event

yiyuezhuo:
# Description

Now it's strange that `data` updating will invoke a `select` event once the DataFrame is selected:

```javascript
	$: {
		if (selected !== false) {
			selected_invoked = true;
			const loc = selected.split("-");
			const row = parseInt(loc[0]);
			const col = parseInt(loc[1]);
			dispatch("select", { index: [row, col], value: data[row][col].value });
		}
	}
```

This should be fixed by adding a `selected_invoked` variable, which permits the dispatching of the `select` one time once a new cell is selected:

```javascript
	let selected_invoked: boolean = false;

	$: {
		if (selected !== false) {
			selected_invoked = false; // Prevent data updating invoke the `select` event.
		}
	}
	$: {
		if (selected !== false && !selected_invoked) {
			selected_invoked = true;
			const loc = selected.split("-");
			const row = parseInt(loc[0]);
			const col = parseInt(loc[1]);
			dispatch("select", { index: [row, col], value: data[row][col].value });
		}
	}
```

Is it required to write unit tests for the fix? The frontend test seems very sparse (if the `npm test` actually covers all the tests), so I don't know where to start.

Further works:

- Variable `selected` should be set to `false` at some places (e.x. when the pointed cell doesn't exist in the table anymore due to updating). But it's complicated, since we may still want the previous pointed location for some reason. So it's desired to be able to update the `selected` state from Python.
- Should clicking the same cell invoke multiple times invoke the `select` event multiple times? (Now it will only invoke once.)

Closes: #3856

No additional dependencies.

# Checklist:

- [X] I have added a short summary of my change to the CHANGELOG.md


abidlabs:
Thanks @yiyuezhuo for the fix! @aliabid94 can you review this PR?

aliabid94:
Thanks for the PR @yiyuezhuo! I modified it slightly so that the data variable is not inside the reactive statement, so it should not be dispatching select in response to data changes.

aliabid94:
@pngwn feel free to give my edit a quick review, otherwise will merge later today

qtli:
Hi, thanks for your all valuable commitment. I use the newest Gradio version (3.27.0). However, the bug mentioned in #3861 is still there. 

I don't know whether I miss some updated version.

yiyuezhuo:
@qtli 3.27.0 is released 2 weeks ago so I guess you need to clone and build from the latest main branch. For reference: https://github.com/gradio-app/gradio/blob/main/CONTRIBUTING.md

abidlabs:
Yes that's right, thanks @yiyuezhuo. @qtli you can install gradio from main by following the instructions here: https://gradio.app/docs/main

Issue #563, opened 2022-02-08
Supporting HTTPS when running locally


abidlabs:
Hi Adi,
Thanks for the quick response.
So explain in detail, I have made a gradio app and it by default ran on  a http connection when I provided my vm ip address in the server_name parameter for the launch function.http://12.13.65.70:3031 (my VM ip address).

But due to company requirements, I have to run this on https. Like https://12.13.65.70:3031.
We have a cert and key available but I am not sure how can I pass them to launch function.

I tried wrapping the gradio interface in  a flask app and tried to run it by additionally providing ssl_context. But unfortunately I did not work.
Can you suggest any idea here?

_Originally posted by @vijay2405 in https://github.com/gradio-app/gradio/discussions/370#discussioncomment-2135364_

abidlabs:
Hi @vijay2405, this is now supported in the latest version of `gradio` on `master` which takes two arguments directly: `ssl_keyfile` and `ssl_certfile` in the `launch()` method. You can test this by generating a certificate for localhost (e.g. following this guide: https://dev.to/rajshirolkar/fastapi-over-https-for-development-on-windows-2p7d) and confirm this works for you.

We will release a new version to PyPi by the end of this week . If you need this fix before then, please take a look at the [contributing guidelines](https://github.com/gradio-app/gradio/blob/master/CONTRIBUTING.md) about how to install from the `master` branch.

Escovilla:
I encountered an issue with using a self-signed certificate and encountered an SSL error with the message "certificate verify failed: unable to get local issuer certificate". To resolve the issue, I followed a solution provided by @abidlabs https://github.com/gradio-app/gradio/issues/563#issuecomment-1033365497 , but it still returned error `certificate verify failed: unable to get local issuer 
certificate .` However, after checking another resource, I  made a change in the blocks.py file of Gradio by setting verify=False in the requests.get function, and that solved the issue.
` requests.get(f"{self.local_url}startup-events", verify=False)`

abidlabs:
Or actually would like to open a PR with the fix? Would be much appreciated!

abidlabs:
Or actually @Escovilla would you like to open a PR with the fix? Would be much appreciated!

Escovilla:
[...]

PR #3873, opened 2023-04-15
Add "ssl_verify" param to allow use of self-signed certs

garrettsutula:
# Description

Implements changes mentioned in #563 by making it possible to pass `ssl_verify=False` when calling `blocks.launch()`.

Notes on things that aren't ✅ below:
- No docs exist for using params `ssl_certfile` and others so I didn't add/update docs.
- No test coverage for tls certs, self-signed or otherwise so I didn't add/update tests.

Closes: #563

# Checklist:


garrettsutula:
I am going to evaluate refactoring to move the instantiation of http clients up into the "main" modules like `blocks.py` - lines like `self.queue_client = httpx.AsyncClient()` need to be passed ssl context or `verify=False` I think.

papuSpartan:
@garrettsutula I think it might be best to just set the circular call made by requests to `verify=False`. There isn't really any reason from a security standpoint to try and prevent MITM attacks on loopback.

garrettsutula:
> @garrettsutula I think it might be best to just set the circular call made by requests to `verify=False`. There isn't really any reason from a security standpoint to try and prevent networking attacks on loopback.

Yeah, that would definitely work and it is somewhat difficult to refactor to solve in another way becuse in `blocks.py` the `queue()` method can and is invoked separately from `launch()` - the tls params would have to be optional & passed to initialize an http client correctly for use by the queue.

@abidlabs do you have an opinion on this? This would work, other approaches feel "more correct" but have much broader impact...

https://github.com/garrettsutula/gradio/commit/417004f9465aa5182b74093d0f245b4f76159864#diff-cab37b7b6ce45220156832fd4ea1597bedcb26f5cfc43ab5c678fc8a586a0232R72

abidlabs:
Thanks for this PR @garrettsutula! I looked into it and I think in some cases, the websocket endpoints of your deployed app can be on a different network location than the gradio app itself, and we do want to support these use cases. See https://github.com/gradio-app/gradio/blob/d8bfa63a38b2623f79e38d2291c8a6cb33e8e6e5/gradio/routes.py#L745

So the approach I'd suggest is that we add an `ssl_verify` parameter to `Queue.start()`, which is passed into `httpx.AsyncClient()`. The `Queue.start()` method is only called once the app has launched, in `Blocks.startup_events()`, so you can pass in the `ssl_verify` attribute.

Would you be able to make the suggested change? Happy to review and merge in after that

garrettsutula:
> Thanks for this PR @garrettsutula! I looked into it and I think in some cases, the websocket endpoints of your deployed app can be on a different network location than the gradio app itself, and we do want to support these use cases. See
> 
> https://github.com/gradio-app/gradio/blob/d8bfa63a38b2623f79e38d2291c8a6cb33e8e6e5/gradio/routes.py#L745
> 
> So the approach I'd suggest is that we add an `ssl_verify` parameter to `Queue.start()`, which is passed into `httpx.AsyncClient()`. The `Queue.start()` method is only called once the app has launched, in `Blocks.startup_events()`, so you can pass in the `ssl_verify` attribute there.
> 
> Would you be able to make the suggested change? Happy to review and merge in after that if everything looks good.

Sure thing! Latest commit adds the changes you suggested. Thank you for the guidance!

abidlabs:
Thanks for making the changes! Tested and works for me. Will merge once CI is green

Thanks again @garrettsutula!

abidlabs:
Merging, thanks again @garrettsutula!

clarriu97:
I'm just needing this feature, is there any idea when a new version will be released? Thank you so much!

abidlabs:
Hi @clarriu97 you can install Gradio from `main` to use this feature until we issue a new release. You can install gradio from main by following the instructions here: https://gradio.app/docs/main

clarriu97:
> Hi @clarriu97 you can install Gradio from `main` to use this feature until we issue a new release. You can install gradio from main by following the instructions here: https://gradio.app/docs/main

Thanks for the advise @abidlabs but I need the release to make a clean install. I look forward to it! 😃

shanky100:
Is the release released, if yes then can you please tell the specific version of it.

I am continously getting
requests.exceptions.SSLError: HTTPSConnectionPool(host='localhost', port=7000): Max retries exceeded with url: /startup-events (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate (_ssl.c:1131)')))

I have tried using latest gradio version i.e 3.39

PR #3895, opened 2023-04-17
Fix HTTP code for file request (static, assets, file)

micky2be:
# Description

- requesting for a non existing file, mostly through `file=path/filename.ext` would trigger a 500 error
- this update allows for a more smooth HTTP response like 404 or even 403
- simplified the `safe_join` for the use within gradio

# Checklist:

### DIFF
--- a/gradio/routes.py
+++ b/gradio/routes.py
@@ -318,6 +318,9 @@ class App(FastAPI):
             if in_app_dir or created_by_app or in_file_dir or was_uploaded:
                 if not abs_path.exists():
                     raise HTTPException(404, "File not found")
+                if abs_path.is_dir():
+                    raise HTTPException(403)
+
                 range_val = request.headers.get("Range", "").strip()
                 if range_val.startswith("bytes=") and "-" in range_val:
                     range_val = range_val[6:]
--- a/gradio/routes.py
+++ b/gradio/routes.py
@@ -8,7 +8,6 @@ import inspect
 import json
 import mimetypes
 import os
-import posixpath
 import secrets
 import tempfile
 import traceback
@@ -265,16 +264,12 @@ class App(FastAPI):
         @app.get("/static/{path:path}")
         def static_resource(path: str):
             static_file = safe_join(STATIC_PATH_LIB, path)
-            if static_file is not None:
-                return FileResponse(static_file)
-            raise HTTPException(status_code=404, detail="Static file not found")
+            return FileResponse(static_file)
 
         @app.get("/assets/{path:path}")
         def build_resource(path: str):
             build_file = safe_join(BUILD_PATH_LIB, path)
-            if build_file is not None:
-                return FileResponse(build_file)
-            raise HTTPException(status_code=404, detail="Build file not found")
+            return FileResponse(build_file)
 
         @app.get("/favicon.ico")
         async def favicon():
@@ -309,21 +304,20 @@ class App(FastAPI):
                 return RedirectResponse(
                     url=path_or_url, status_code=status.HTTP_302_FOUND
                 )
-            abs_path = str(utils.abspath(path_or_url))
-            in_app_dir = utils.abspath(app.cwd) in utils.abspath(path_or_url).parents
-            created_by_app = abs_path in set().union(*blocks.temp_file_sets)
+            abs_path = utils.abspath(path_or_url)
+            in_app_dir = utils.abspath(app.cwd) in abs_path.parents
+            created_by_app = str(abs_path) in set().union(*blocks.temp_file_sets)
             in_file_dir = any(
                 (
-                    utils.abspath(dir) in utils.abspath(path_or_url).parents
+                    utils.abspath(dir) in abs_path.parents
                     for dir in blocks.file_directories
                 )
             )
-            was_uploaded = (
-                utils.abspath(app.uploaded_file_dir)
-                in utils.abspath(path_or_url).parents
-            )
+            was_uploaded = utils.abspath(app.uploaded_file_dir) in abs_path.parents
 
             if in_app_dir or created_by_app or in_file_dir or was_uploaded:
+                if not abs_path.exists():
+                    raise HTTPException(404, "File not found")
                 range_val = request.headers.get("Range", "").strip()
                 if range_val.startswith("bytes=") and "-" in range_val:
                     range_val = range_val[6:]
@@ -341,8 +335,9 @@ class App(FastAPI):
                 return FileResponse(abs_path, headers={"Accept-Ranges": "bytes"})
 
             else:
-                raise ValueError(
-                    f"File cannot be fetched: {path_or_url}. All files must contained within the Gradio python app working directory, or be a temp file created by the Gradio python app."
+                raise HTTPException(
+                    403,
+                    f"File cannot be fetched: {path_or_url}. All files must contained within the Gradio python app working directory, or be a temp file created by the Gradio python app.",
                 )
 
         @app.get("/file/{path:path}", dependencies=[Depends(login_check)])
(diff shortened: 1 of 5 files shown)


abidlabs:
Thanks @micky2be for making this fix! Generally looks good -- however, we should leave `gr.utils.abspath()` as is. The reason for that is because the `pathlib.Path.absolute()` path hasn't been well-tested in Python 3.7 and so we have avoided using it. You can [read (a lot) more about it here if you are interested](https://discuss.python.org/t/pathlib-absolute-vs-resolve/2573/13).

If you can revert that change, happy to review th rest.

micky2be:
@abidlabs I see. Reverting as recommended.

gradio/routes.py
```diff
@@ -588,29 +588,26 @@ def robots_txt():
...
-
-
-def safe_join(directory: str, path: str) -> str:
-    if path == "":
-        raise HTTPException(400)
-
-    filename = Path(path)
-    root = Path(directory)
-    full_path = root.joinpath(filename).resolve()
+def safe_join(directory: str, path: str) -> str | None:
```
micky2be: Not sure why you rolled back that change. It will break other part of the code which now rely on this.

abidlabs:
Made some more tweaks since the `is_relative_to` method was introduced in Python 3.9, but `gradio` currently supports 3.7

micky2be:
You didn't have to roll back. Your tweak is breaking the code, and will not trigger a proper 404.
I'm not using `Path.is_relative_to` in the code

abidlabs:
You were using it here: 

>     if filename.is_absolute() or not is_relative_to(full_path.parent, directory):
>         raise HTTPException(403)

micky2be:
`is_absolute` exist in 3.7

gradio/routes.py
```diff
@@ -592,26 +588,29 @@ def robots_txt():
 ########
 
 
-def safe_join(directory: str, path: str) -> str | None:
-    """Safely path to a base directory to avoid escaping the base directory.
-    Borrowed from: werkzeug.security.safe_join"""
-    _os_alt_seps: List[str] = list(
-        sep for sep in [os.path.sep, os.path.altsep] if sep is not None and sep != "/"
-    )
+def is_relative_to(self, other):
```
micky2be, line 591: @abidlabs this is what you are looking for

abidlabs: But `is_relative_to` relies on the pathlib's Path's `relative_to` method internally. So it will not work on Python 3.7, right?

micky2be: Yes, `relative_to ` was introduced a while ago

abidlabs: Ah my apologies for the confusion

abidlabs:
I'm re-reviewing the code. Regarding rewriting the `safe_join` method -- it's a critical method as it governs file access throughout all of gradio, so unless there is a strong reason to do it, I'm still not in favor of it. I especially do not want to use `pathlib.Path.absolute()` because, as I mentioned earlier, it is severely undertested in Python 3.7.

Would you be willing to leave the logic of `safe_join` as is and make the other changes outside of the method to raise the 403/404s as I had started to do?

micky2be:
I'm not using `pathlib.Path.absolute()` at all.
Previous logic was a copy/paste of a library that does slightly different things (allow multiple arguments as path)
Here the code is much simpler and make sure `path` is not accessing files outside of `directory`.
That's why `test_route.py::TestRoutes::test_static_files_served_safely` is still passing

`safe_join` is used by the `/static/{path:path}` and `/assets/{path:path}` routes

micky2be:
But if changing `safe_join` is such a deal breaker for you (_despite not doing much beside looking for absolute path and access outside of `directory`_), I'll update the code tomorrow

gradio/routes.py
```diff
@@ -592,26 +588,29 @@ def robots_txt():
...
-    return posixpath.join(directory, filename)
+def safe_join(directory: str, path: str) -> str:
+    if path == "":
+        raise HTTPException(400)
+
+    filename = Path(path)
+    root = Path(directory)
+    full_path = root.joinpath(filename).resolve()
+
+    if filename.is_absolute() or not is_relative_to(full_path.parent, directory):
```
micky2be, line 608: This cover Lines 609 to 611 from previous code.
- check if `path` is an absolute path
- avoid `path` accessing files outside of `directory`

Basically forbidden access, hence the 403

gradio/routes.py
```diff
@@ -592,26 +588,29 @@ def robots_txt():
...
-        any(sep in filename for sep in _os_alt_seps)
-        or os.path.isabs(filename)
-        or filename == ".."
-        or filename.startswith("../")
-    ):
-        return None
-    return posixpath.join(directory, filename)
+def safe_join(directory: str, path: str) -> str:
+    if path == "":
+        raise HTTPException(400)
```
micky2be, line 602: This will return an error instead of a 200 when accessing the root path of `directory`

gradio/routes.py
```diff
@@ -592,26 +588,29 @@ def robots_txt():
...
+
+    filename = Path(path)
+    root = Path(directory)
+    full_path = root.joinpath(filename).resolve()
+
+    if filename.is_absolute() or not is_relative_to(full_path.parent, directory):
+        raise HTTPException(403)
+
+    if not full_path.exists():
```
micky2be, line 611: This check if we can actually return the requested file

micky2be (commented):
Making sure the code is understood

gradio/routes.py
micky2be: Something to add and doable is to check if `path` is requesting a folder and instead of a file

abidlabs:
> But if changing safe_join is such a deal breaker for you (despite not doing much beside looking for absolute path and access outside of directory), I'll update the code tomorrow

Thanks for your patience @micky2be! I would strongly prefer that we do not change the logic of the `safe_join` method. There are small differences in how different OS-es handle path resolution and we've spent a lot of time in the past debugging these. I would rather not go through all of that testing again. 

If you can please refactor this so that the logic of the `safe_join` method is the same as before, I'd be happy to merge this in. If you would like to raise the HTTP errors inside the `safe_join()` method to reduce code duplication, that is fine. But we should use the same `os.path` module instead of `pathlib` and keep the overall logic the same.

micky2be:
Alright! Hope that fits the need this time 😄
@abidlabs let me know if it looks good to you.
Not sure how to trigger the workflows.

micky2be:
Interesting! It seems my new tests exposed an issue related to `pathlib.Path.resolve()` (fixed in Python 3.10) when trying to resolve the path of a non existing file on Windows.

Working on a fix for it right now.

micky2be:
Fixed the issue with unresolved path (for non existing path) in `utils.abspath`
Also noticed `is_symlink` is not recursive so it would miss symlinks and would resolve it.
There was also a potential issue if an absolute path would be passed and be a symlink.

abidlabs:
Cool thanks for making these fixes @micky2be! OOO but will review when I'm back in the office (cc @freddyaboulton or @aliabid94 if you guys wanna take a look in the meantime)

gradio/routes.py
```diff
@@ -309,21 +304,20 @@ async def file(path_or_url: str, request: fastapi.Request):
...
                 )
             )
-            was_uploaded = (
-                utils.abspath(app.uploaded_file_dir)
-                in utils.abspath(path_or_url).parents
-            )
+            was_uploaded = utils.abspath(app.uploaded_file_dir) in abs_path.parents
 
             if in_app_dir or created_by_app or in_file_dir or was_uploaded:
+                if not abs_path.exists():
```
freddyaboulton, line 319: We should raise a 403 if abs_path is a directory

freddyaboulton (approved):
Thanks for the improvement @micky2be and for collaborating with @abidlabs ! 

Tested on deployed spaces and locally and it's working well. Left a small comment for improvement. Should be good to go after that

gradio/routes.py
micky2be: That's really subjective. Could be 400, 403, or 404.
But we are not checking for directory here anyway.
Is it something you want me to add?

micky2be: Good idea!

freddyaboulton:
Thanks again @micky2be !

micky2be:
Hey guys (@freddyaboulton @abidlabs)
As I was expending the battery of test of the routes for the sake of it (and for fun) I realized two things I'm not sure about:
- The upload path is absolute and create files outside of the application folder (default for `tempfile.mkdtem()`)
  - Windows: `C:\Users\<user>\AppData\Local\Temp\<hash>/`
  - Linux: `/tmp/<hash>/`
- Users have full access of any files of the application (in_app_dir)
  - Potential security risk for some

abidlabs:
Hi @micky2be both of those are expected behavior, and the latter is mentioned here: https://gradio.app/sharing-your-app/#security-and-file-access

We may explicitly allow blacklisting files as  mentioned here: https://github.com/gradio-app/gradio/issues/3939

Thanks so much again for the contribution @micky2be and for reviewing @freddyaboulton!

PR #3903, opened 2023-04-18
Fix some documentation-related issues in Guides

abidlabs:
Makes some fixes to the Guides/documentation:

Now the docs for Chatbot
* first show a very simple Chatbot UI
* then add streaming + .then()
* conclude with multimodal chatbot

Also fixes the issue with the missing HF_TOKEN that @pngwn pointed out yesterday, as well as adds `duplicate()` to the Client docs.

Finally fixes missing Guides linked from the docs. This broke because we renamed the guides to have hyphens instead of underscores

@aliabd I'm going to be OOO so feel free to push directly to this branch and merge when ready


freddyaboulton (approved):
Thanks for fixing the guides @abidlabs !

ningzichun:
Hello, it seems that the demo at [#running-events-consecutively](https://gradio.app/blocks-and-event-listeners/#running-events-consecutively) is still the old one that prints "Yes" or "No"

Issue #3890, opened 2023-04-17
Image is resized to a fixed height in newest gradio version


jingedawang:
### Describe the bug

In gradio 3.21.0, the image as an output always have 100% width.
But in gradio 3.27.0, the image as an output always have a fixed height. This is a worse experience because user is happy to scroll the page up and down, but not in a horizontal way. When the height of the image is bigger than the width, it will be resized to very small.

Is this a bug or is this by design?
Is there any option I can set to resume the original appearance?

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Here is the code:
```Python
import gradio as gr

# Draw an UI with a image output
with gr.Interface(
    fn=lambda x: x,
    inputs=gr.inputs.Image(type='pil'),
    outputs=gr.outputs.Image(label="Image", type='pil'),
) as app:
    app.launch()
```

With `gradio~=3.21.0` and `gradio~=3.27.0`, we will got different UI.

### Screenshot

With `gradio~=3.21.0`, we got
![image](https://user-images.githubusercontent.com/7996256/232483862-8e7bb775-78d3-4d0a-8485-9a3abde741ff.png)

With `gradio~=3.27.0`, we got
![image](https://user-images.githubusercontent.com/7996256/232484553-47a3131e-0bd9-47ff-a600-261141536432.png)

This is the picture I was using for the test: 
![WeChat Image_20230417202653](https://user-images.githubusercontent.com/7996256/232484754-f549d2ad-9afd-4a8a-8679-1c15702f4504.jpg)

### Logs

```shell
None
```

### System Info

```shell
Gradle 3.21.0 and 3.27.0
Windows, Android
Edge, Wechat Web preview
```

### Severity

blocking upgrade to latest gradio version

PR #3905, opened 2023-04-18
Fix output image heights

aliabid94:
Image output heights automatically expand now, unless explicitly defined.

Fixes: #3890


freddyaboulton (approved):
Tested and works well! Thanks @aliabid94

PR #3927, opened 2023-04-21
Add JS to use via api 

aliabd:
Adds the JS client to the use-via-api page. 

Code snippets need testing.


js/app/src/api_docs/ApiDocs.svelte
```diff
@@ -62,16 +64,25 @@
 		let data = await response.json();
 		return data;
 	}
+	async function get_js_info() {
+		let js_api_info = await app.view_api();
+		console.log(js_api_info);
```
pngwn, line 69: Remove log

Issue #3949, opened 2023-04-24
Rendering image URL as a response in chatbot


sunilkumardash9:
### Describe the bug

Not able to render image URLs as a response in the chatbot.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```
`def add_text_for_images(history, text):
    history = history + [(text,'')]
    return history

def generate_image(history, text):
    img_url = 'https://oaidalleapiprodscus.blob.core.windows.net/private/org-T6WvejtuZhCqGecac5Yz862J/user-8LOO4C5fnrZYzYKwpILUEVUF/img-W3B70rebpSI4Gqqb2PJpPlsP.png?st=2023-04-24T02%3A11%3A04Z&se=2023-04-24T04%3A11%3A04Z&sp=r&sv=2021-08-06&sr=b&rscd=inline&rsct=image/png&skoid=6aaadede-4fb3-4698-a8f6-684d7786b067&sktid=a48cca56-e6da-484e-a814-9c849652bcb3&skt=2023-04-24T03%3A04%3A50Z&ske=2023-04-25T03%3A04%3A50Z&sks=b&skv=2021-08-06&sig=yn/05MJTXiYSf5ECQmDgZKkoo5fvTu2jlDbcKzeFSbk%3D'
    history[-1][1] = (img_url, None)
    return history`

with gr.Blocks() as demo:

    chatbot = gr.Chatbot(value=[], elem_id="chatbot").style(height=650)
    with gr.Row():
        with gr.Column(scale=0.70):
            txt = gr.Textbox(
                show_label=False,
                placeholder="Enter text and press enter, or upload an image",
            ).style(container=False) 

    txt.submit(add_text_for_images, [chatbot, txt], chatbot).then(
                generate_image, [chatbot,txt], chatbot )

if __name__ == "__main__":
    demo.launch()  
```

### Logs

```shell
Traceback (most recent call last):
  File "/home/sunil/.cache/pypoetry/virtualenvs/textgpt-3e1gj0HQ-py3.10/lib/python3.10/site-packages/gradio/routes.py", line 401, in run_predict
    output = await app.get_blocks().process_api(
  File "/home/sunil/.cache/pypoetry/virtualenvs/textgpt-3e1gj0HQ-py3.10/lib/python3.10/site-packages/gradio/blocks.py", line 1305, in process_api
    data = self.postprocess_data(fn_index, result["prediction"], state)
[...]

PR #3953, opened 2023-04-24
Allow image url as ChatBot response

tye-singwa:
# Description
- added check if it is valid url, then pass to ui as is, else try to make temporary copy

Closes: #3949

# Checklist:


gradio/components.py
```diff
@@ -4581,9 +4581,13 @@ def _postprocess_chat_messages(
         if chat_message is None:
             return None
         elif isinstance(chat_message, (tuple, list)):
-            filepath = chat_message[0]
+            image_uri = chat_message[0]
+            if utils.validate_url(image_uri):
+                filepath = image_uri
+            else:
+                filepath = self.make_temp_copy_if_needed(image_uri)
```
abidlabs, line 4588:
```suggestion
            file_uri = chat_message[0]
            if utils.validate_url(file_uri):
                filepath = file_uri
            else:
                filepath = self.make_temp_copy_if_needed(file_uri)
```

abidlabs: Just a suggestion to rename the parameter is it does not necessarily have to be an image -- can be a video or audio as well

tye-singwa: right, thanks!

abidlabs:
LGTM thank you very much for the contribution @tye-singwa!

PR #3956, opened 2023-04-24
Move Google Tag Manager code to analytics_enabled block in Blocks.svelte

akx:
It's not really well documented that using Gradio may make your application communicate with Google Tag Manager, which allows whoever owns the property `UA-156449732-1` (hopefully someone benign) to execute basically **any code**; that should probably be made clearer.

# Description

* relevant motivation: I don't think Google Analytics code (even if it does nothing unless GTM is enabled) should exist in the base HTML template when analytics are not enabled.
* a summary of the change: move the initialization code to the Svelte template.

# Checklist:


abidlabs:
Thanks for the PR @akx! Seems reasonable to me. Taking a look now

abidlabs:
LGTM @akx! Thanks again. I just ran `scripts/format_frontend.sh` so that the static checks for the frontend pass. Will merge once the CI is green

Issue #3544, opened 2023-03-20
Theme-related requests


pngwn:
Is there something you would like to style via the theming API but isn't currently supported?

Please add a reply to this issue detailing exactly what you would like to style and how you would like to style it! Screenshots are very welcome!

space-nuko:
The [section on CSS theme variables](https://gradio.app/theming-guide/#css-variable-naming-conventions) makes it sound like there's an autogenerated CSS class for all major components/layouts, but actually this isn't the case, the available CSS vars are only whatever's defined by the devs in the `.svelte` files, and not all components support them

For example, I want to get rid of all padding on elements, but not on `gr.Accordion`s. Turns out the block that sets the `padding` class wraps the inner `Accordion` element is what defines the padding. In this case it seems like the specificity I need is not covered by the theme API

In fact, I couldn't even write a good custom CSS rule for this case, I needed to manually add a `.gr-accordion` class to every created `gr.Accordion` to something besides an autogenerated class to target (what I was mentioning in #3545). The important thing is `elem_classes` doesn't help since I want to target *all* `gr.Accordion` elements, including ones added by extensions (where I might not be able to control what classes they have)

And it looks like the only CSS vars exposed for accordions are some `--section-header-text-*` ones that target other elements also, there is nothing that contains an `--accordion` prefix

https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/34bbbbc28ea59d45de4b9faa80820edafacc3871/modules/scripts.py#L554

anapnoe:
469 --vars  already most of them are generic 

this is a complicated issue and there are a lot of different design patterns 
most frameworks had spend years to define and find balance between flexibility and bulkiness
if you want to do it right at least study some other frameworks 

1.first of all you need to prefix your vars so there are no conflict with other libraries this is important gr is nice and small
2.seperate the components by feature some of them may need more attributes to be defined keep consistency in names
3.write the mixin for each variant

this is an example code of how bootstrap5 does it 
```
@mixin table-variant($state, $background) {
  .table-#{$state} {
    $color: color-contrast(opaque($body-bg, $background));
    $hover-bg: mix($color, $background, percentage($table-hover-bg-factor));
    $striped-bg: mix($color, $background, percentage($table-striped-bg-factor));
    $active-bg: mix($color, $background, percentage($table-active-bg-factor));
    $table-border-color: mix($color, $background, percentage($table-border-factor));

    --#{$prefix}table-color: #{$color};
[...]

PR #3957, opened 2023-04-24
Allow setting default theme

akx:
# Description

* relevant motivation: https://github.com/gradio-app/gradio/issues/3544#issuecomment-1520695468
* a summary of the change: instead of referring to `Default`, look at `Context.default_theme_class` (which can be set via envvar too).

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@
 
 ## New Features:
 
-No changes to highlight.
+- You can now set the default theme for all `gr.Blocks()` that don't specify a theme with `gradio.context.Context.default_theme`.
+  Implemented by [@akx](https://github.com/akx) in [PR 3957](https://github.com/gradio-app/gradio/pull/3957)
 
 ## Bug Fixes:
 - Fix duplicate play commands in full-screen mode of 'video'. by [@tomchang25](https://github.com/tomchang25) in [PR 3968](https://github.com/gradio-app/gradio/pull/3968).
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -29,15 +29,14 @@ from gradio import (
     queueing,
     routes,
     strings,
-    themes,
     utils,
 )
 from gradio.context import Context
 from gradio.deprecation import check_deprecated_parameters
 from gradio.exceptions import DuplicateBlockError, InvalidApiName
 from gradio.helpers import EventData, create_tracker, skip, special_args
-from gradio.themes import Default as DefaultTheme
 from gradio.themes import ThemeClass as Theme
+from gradio.themes import is_custom_theme, resolve_theme
 from gradio.tunneling import CURRENT_TUNNELS
 from gradio.utils import (
     GRADIO_VERSION,
@@ -55,17 +54,6 @@ if TYPE_CHECKING:  # Only import for type checking (is False at runtime).
     from fastapi.applications import FastAPI
 
     from gradio.components import Component
-
-BUILT_IN_THEMES: Dict[str, Theme] = {
-    t.name: t
-    for t in [
-        themes.Base(),
-        themes.Default(),
-        themes.Monochrome(),
-        themes.Soft(),
-        themes.Glass(),
-    ]
-}
 
 
 class Block:
@@ -626,23 +614,9 @@ class Blocks(BlockContext):
         # Cleanup shared parameters with Interface #TODO: is this part still necessary after Interface with Blocks?
         self.limiter = None
         self.save_to = None
-        if theme is None:
-            theme = DefaultTheme()
-        elif isinstance(theme, str):
-            if theme.lower() in BUILT_IN_THEMES:
-                theme = BUILT_IN_THEMES[theme.lower()]
-            else:
-                try:
-                    theme = Theme.from_hub(theme)
-                except Exception as e:
-                    warnings.warn(f"Cannot load {theme}. Caught Exception: {str(e)}")
-                    theme = DefaultTheme()
-        if not isinstance(theme, Theme):
-            warnings.warn("Theme should be a class loaded from gradio.themes")
-            theme = DefaultTheme()
-        self.theme: Theme = theme
-        self.theme_css = theme._get_theme_css()
-        self.stylesheets = theme._stylesheets
+        self.theme: Theme = resolve_theme(theme)
+        self.theme_css = self.theme._get_theme_css()
+        self.stylesheets = self.theme._stylesheets
         self.encrypt = False
         self.share = False
         self.enable_queue = None
@@ -697,15 +671,11 @@ class Blocks(BlockContext):
         self.file_directories = []
 
         if self.analytics_enabled:
-            is_custom_theme = not any(
-                self.theme.to_dict() == built_in_theme.to_dict()
-                for built_in_theme in BUILT_IN_THEMES.values()
-            )
             data = {
                 "mode": self.mode,
                 "custom_css": self.css is not None,
                 "theme": self.theme.name,
-                "is_custom_theme": is_custom_theme,
+                "is_custom_theme": is_custom_theme(self.theme),
                 "version": GRADIO_VERSION,
             }
             utils.initiated_analytics(data)
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -605,7 +605,7 @@ class Blocks(BlockContext):
     ):
         """
         Parameters:
-            theme: a Theme object or a string representing a theme. If a string, will look for a built-in theme with that name (e.g. "soft" or "default"), or will attempt to load a theme from the HF Hub (e.g. "gradio/monochrome"). If None, will use the Default theme.
+            theme: a Theme object or a string representing a theme. If a string, will look for a built-in theme with that name (e.g. "soft" or "default"), or will attempt to load a theme from the HF Hub (e.g. "gradio/monochrome"). If None, will use the default theme class configured with `gradio.themes.set_default_theme_class()` (which defaults to `Default`).
             analytics_enabled: whether to allow basic telemetry. If None, will use GRADIO_ANALYTICS_ENABLED environment variable or default to True.
             mode: a human-friendly name for the kind of Blocks or Interface being created.
             title: The tab title to display when this is opened in a browser window.
--- a/gradio/themes/__init__.py
+++ b/gradio/themes/__init__.py
@@ -1,3 +1,8 @@
+from __future__ import annotations
+
+import warnings
+from typing import Dict
+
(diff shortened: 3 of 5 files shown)


abidlabs:
Thanks @akx! Instead of using a global variable, it would be better to use an environmental variable to be consistent with other features in the gradio library (such as setting `analytics_enabled`). Would you be able to make the change & also add a test? Otherwise, the refactoring is nice

abidlabs:
Ah makes sense, in that case no worries about the environmental variable. It's better if there aren't too many ways to acheive the same thing. We actually have a dedicated `Context` class in `gradio.context.py` designed to hold all global variables. Can you add a field to this class called `default` theme class which is set by the `set_default_theme_class()` you defined? Would still be good to add a test

akx:
@abidlabs I reworked this a bit. Some client tests seem to be flaky though 😩 

* The theme-related internals from `gradio.blocks` now live (unsurprisingly) in `gradio.themes`.
  * I also optimized `is_custom_theme` a tiny bit: it doesn't `.to_dict()` the theme over and over for each built-in theme... :)
* In the interest of There's Only One Way To Do It, the way to get and set the default theme class is to just plonk it in `Context`; there's no more separate setter and getter function.
* I added the `GRADIO_DEFAULT_THEME` envvar anyway since it was pretty trivial to do so.

CHANGELOG.md
```diff
@@ -3,7 +3,9 @@
 
 ## New Features:
 
-No changes to highlight.
+- You can now set the default theme class for all `gr.Blocks()` that don't specify a theme with `gradio.themes.set_default_theme_class()`; this must be called before instantiating a `gr.Blocks`.
```
abidlabs, line 6: Needs to be updated to reflect the new usage with `Context`

CHANGELOG.md
```diff
@@ -3,7 +3,9 @@
...
+  You can also set the default theme class from the built-in themes with the environment variable `GRADIO_DEFAULT_THEME` (e.g. `GRADIO_DEFAULT_THEME=soft`).
```
abidlabs, line 7: I would remove this for the reasons mentioned previously. While I agree with you that implementing this is straightforward, this really is unnecessary when we have the `Context` approach already, and potentially makes debugging the behavior of themes more difficult

gradio/blocks.py
```diff
@@ -617,7 +605,7 @@ def __init__(
     ):
         """
         Parameters:
-            theme: a Theme object or a string representing a theme. If a string, will look for a built-in theme with that name (e.g. "soft" or "default"), or will attempt to load a theme from the HF Hub (e.g. "gradio/monochrome"). If None, will use the Default theme.
+            theme: a Theme object or a string representing a theme. If a string, will look for a built-in theme with that name (e.g. "soft" or "default"), or will attempt to load a theme from the HF Hub (e.g. "gradio/monochrome"). If None, will use the default theme class configured with `gradio.themes.set_default_theme_class()` (which defaults to `Default`).
```
abidlabs, line 608: Needs to be updated

gradio/themes/__init__.py
```diff
@@ -14,3 +21,65 @@ def builder(*args, **kwargs):
...
+        Soft(),
+        Glass(),
+    ]
+}
+
+# You can set the default theme to one of the built-in themes by setting the environment variable GRADIO_DEFAULT_THEME.
+# If the environment variable is not set or does not refer to a built-in theme, the default theme is Default.
+Context.default_theme_class = _built_in_themes.get(
+    os.environ.get("GRADIO_DEFAULT_THEME")
+)
```
abidlabs, line 41: Please remove as suggested above
```suggestion
```

gradio/blocks.py
```diff
@@ -626,23 +614,9 @@ def __init__(
...
-                    theme = Theme.from_hub(theme)
-                except Exception as e:
-                    warnings.warn(f"Cannot load {theme}. Caught Exception: {str(e)}")
-                    theme = DefaultTheme()
-        if not isinstance(theme, Theme):
-            warnings.warn("Theme should be a class loaded from gradio.themes")
-            theme = DefaultTheme()
-        self.theme: Theme = theme
-        self.theme_css = theme._get_theme_css()
-        self.stylesheets = theme._stylesheets
```
abidlabs, line 645: Nice refactoring here

gradio/themes/__init__.py
```diff
@@ -14,3 +21,65 @@ def builder(*args, **kwargs):
...
+        Default(),
+        Monochrome(),
+        Soft(),
+        Glass(),
+    ]
+}
+
+# You can set the default theme to one of the built-in themes by setting the environment variable GRADIO_DEFAULT_THEME.
+# If the environment variable is not set or does not refer to a built-in theme, the default theme is Default.
+Context.default_theme_class = _built_in_themes.get(
```
abidlabs, line 39: I think this would be better if it was the theme object rather than the class. That way, you could load a theme from the Hub and assign it as a global theme as well

abidlabs: In which case, I'd rename `Context.default_theme_class` to `Context.default_theme`

abidlabs:
> In the interest of There's Only One Way To Do It, the way to get and set the default theme class is to just plonk it in Context; there's no more separate setter and getter function.

I think this is fine, but we should just make sure to document `Context.default_theme` somewhere. Perhaps we can mention it in the docstring for the `Blocks` and `Interface` classes for the `theme` parameter

pngwn:
I don't love this API, it might be more verbose but passing the theme into each `Block` is less error prone and doesn't rely on shared mutable state.

akx:
@pngwn Thanks for the feedback! I'd call this configuration instead of shared mutable state. 

The issue is an application can easily have a bunch of `gr.Blocks()`es (my case in point being https://github.com/AUTOMATIC1111/stable-diffusion-webui, with 12 invocations if you don't count all of the extensions one could have loaded), so either the application itself would need to have a way to set the default theme, which sounds like a framework concern more than an application concern.

abidlabs:
I'm thinking about this some more and trying to figure out if this is really needed. Regarding:

> @pngwn Thanks for the feedback! I'd call this configuration instead of shared mutable state.
> 
> The issue is an application can easily have a bunch of `gr.Blocks()`es (my case in point being https://github.com/AUTOMATIC1111/stable-diffusion-webui, with 12 invocations if you don't count all of the extensions one could have loaded).
> 
> There would need to have a way to set the default theme without having to touch all that code, which sounds like a framework concern more than an application concern.
> 
> Also, before this PR if you're not passing in a theme, you're implicitly passing in `Default` – this just makes that implicit default changeable.

Isn't it the case that the 12 `gr.Blocks()` are enclosed in one parent `gr.Blocks()`? One could simply set the theme in the parent `gr.Blocks()` and it would apply to all of the children Blocks?

akx:
> Isn't it the case that the 12 `gr.Blocks()` are enclosed in one parent `gr.Blocks()`? One could simply set the theme in the parent `gr.Blocks()` and it would apply to all of the children Blocks?

As far as I can tell, no. Sorry for a screenshot-of-code, it was the easiest way to illustrate this since each of those folded blocks contains 200+ lines of stuff.

Screen Shot 2023-04-27 at 8 23 42

abidlabs:
But if you look here (specifically line 1577): https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/22bcc7be428c94e9408f589966c2040187245d81/modules/ui.py#L1568

All of these sub Blocks are included in a parent `gr.Blocks()`

Given that this is the case (you can always include sub-Blocks inside a larger Blocks and just apply the theme to the parent Blocks), I'm inclined towards closing this PR for now.

akx:
@abidlabs Good catch! 

I'd still think this is a valuable addition without much added complexity (most of the diff is documentation or comments!), because it allows setting the theme without having to touch application code.

That said, what _is_ the mechanism in the code that makes sub-blocks use their parents' theme?

abidlabs:
> I'd still think this is a valuable addition without much added complexity (most of the diff is documentation or comments!), because it allows setting the theme without having to touch application code.

That's true, however, the API is messy (I don't like exposing the `Context` object to users) and the use of a global variable can lead to confusion or problems debugging when a user reruns the application code and sees a different UI created. To put it another way, I think that setting the theme *should* require touching application code. If someone copies the code for their Gradio demo and puts it on Spaces, or runs it in a Colab notebook, they should get the same exact UI created, which might not happen if a global variable can be used to set the theme. 

> That said, what is the mechanism in the code that makes sub-blocks use their parents' theme?

Turns out that if you create or .render() a Blocks inside a parent Blocks, then all of the children components and event triggers are re-created inside the parent Blocks. The code for that is here: https://github.com/gradio-app/gradio/blob/847247e1dd8e4ede9e7031ff15f207067b11cd6b/gradio/blocks.py#L841

Appreciate you creating this PR @akx but for the reason mentioned above and the lack of a compelling use case, I'll go ahead and close it.

akx:
@abidlabs 

* We could come up with an API that doesn't require touching `Context`, for sure. (Renaming `Context.default_theme` to `Context._default_theme` and giving it a setter function sounds like a solid plan.)
* You'd still need to touch the _program_ code to call the default theme setter, you just wouldn't need to hunt around for the basest `gr.Blocks` or `gr.Interface` call to add a theme there.
* The refactorings from this PR could still be useful, I think?

Also, given how the theme mechanism now works, the code

```python
import gradio as gr

with gr.Blocks(theme="soft") as demo:
    gr.Button("soft button")
    with gr.Blocks(theme="glass"):
        gr.Button("glass button")

if __name__ == "__main__":
    demo.launch()
```

yields

Screen Shot 2023-04-28 at 9 41 58

– i.e. two soft buttons – which is pretty unintuitive. Maybe `theme` shouldn't be on `gr.Blocks` or `gr.Interface` _at all_ since it only works on the topmost level?

abidlabs:
I don't think the added complexity is worth it. For example, what happens if you call the setter function in the middle of a Blocks? 

> – i.e. two soft buttons – which is pretty unintuitive. Maybe theme shouldn't be on gr.Blocks or gr.Interface at all since it only works on the topmost level?

I do agree with this point -- perhaps we should display a warning if a Blocks that is not top-level is created with a theme.

akx:
> I don't think the added complexity is worth it. For example, what happens if you call the setter function in the middle of a Blocks?

The same thing that happens in the above example: nothing. Whatever had been the default theme at the time the topmost `gr.Blocks` was configured is the theme for the app.

Since the theme is, de facto, currently app-wide, I think it shouldn't be an option on `gr.Blocks` and `gr.Interface` at all, and there should be an API to set it.

PR #3958, opened 2023-04-24
gradio.themes: get rid of useless re-exports + noqas

akx:
# Description

## Summary

This PR:

* removes the unhygienic and unused star re-exports from `gradio.themes.utils`; yes, this means you can't do e.g. `gradio.themes.utils.stone` anymore. Those exports seemingly weren't used by anything in `gradio` itself.
* gets rid of the `noqa: F401`s in `gradio.themes` by simply specifying the things we want to export (or we could just ignore F401s for _all_ `__init__.py`s (see discussion in https://github.com/gradio-app/gradio/pull/3710/files#r1156310154).
  * In general, if you find yourself adding `noqa`s in new code, you may want to reconsider :)
* fixes the import for `create_theme_dropdown` in `gradio.themes.app` to not rely on `sys.path` shenanigans anymore.

## Relevant motivation

Down with star exports that make static analysis hard! Down with peppering `noqa` around the code!

# Checklist:


abidlabs:
Thanks again @akx! cc @aliabid94 can you take a quick look to make sure this doesn't break anything

gradio/themes/app.py
```diff
@@ -1,8 +1,7 @@
 import time
 
-from theme_dropdown import create_theme_dropdown  # noqa: F401
-
 import gradio as gr
+from gradio.themes.utils.theme_dropdown import create_theme_dropdown
```
freddyaboulton, line 4: Let's delete the `theme_dropdown.py` from the space that's pushed up (in themes/base.py)

```
            CommitOperationAdd(
                path_in_repo="theme_dropdown.py",
                path_or_fileobj=str(
                    Path(__file__).parent / "utils" / "theme_dropdown.py"
                ),
            ),
```

akx: Done.

akx:
gradio-backend / client-test CI is being flaky 😞

freddyaboulton:
Yep, planning a rewrite this week to improve that.

abidlabs:
> Thanks for the PR @akx! I agree this is the correct approach and the changes look good to me. I don't think any user should have been poking around with gr.themes.utils.blue, so I think this is okay as an internal API change rather than a breaking change that would require a version bump (wdyt @abidlabs?)

I think it's okay too. A [full text search on Spaces](https://huggingface.co/search/full-text?q=gr.themes.utils) didn't find any such usage

abidlabs:
Thanks @akx for addressing the feedback! Rerunning the failed tests and will merge once CI passes

gradio/themes/app.py
abidlabs, line 4: Actually this change breaks theme uploading. See here: https://huggingface.co/spaces/abidlabs/test-yellow

abidlabs (changes requested):
Working on a fix for the theme upload issue

abidlabs:
Fixed the issue above, as can be verified in this Space which is running Gradio off of this branch: https://huggingface.co/spaces/abidlabs/test-yellow3

abidlabs (approved):

abidlabs:
I believe it should be good to merge now once the CI passes, @freddyaboulton could you take one look at the change I made as well? https://github.com/gradio-app/gradio/pull/3958/commits/1e590254891b73a0c5c43a7513b63e1759e8b4f4

gradio/themes/utils/theme_dropdown.py
```diff
@@ -7,7 +7,7 @@
 def create_theme_dropdown():
     import gradio as gr
 
-    asset_path = pathlib.Path(__file__).parent / "themes"
+    asset_path = pathlib.Path() / "themes"
```
freddyaboulton, line 10: Yep this makes sense. Before this file was placed alongside `app.py` in the space but now it's imported from the lib so `__file__` was wrong.

Thanks @abidlabs !

abidlabs: Thank you @freddyaboulton for the quick reply! Will merge then

abidlabs:
Everything passes except for a flaky test on Windows, so will go ahead and merge in. Thanks @akx for cleaning up these files.

Issue #3675, opened 2023-03-28
`demo.close()` doesn't release the port if it is not 7860


oobabooga:
### Describe the bug

If I initialize a a `gr.Blocks()` interface with a custom `server_port`, the port is not released after `demo.close()`.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

1. Launch the script below.
2. Access the web UI in `http://127.0.0.1:9441/?__theme=dark`
3. Click on "Apply and restart the interface"
4. An error about port 9441 already being in use will appear in the terminal

```python
import time

import gradio as gr

title ='Text generation web UI'
need_restart = False

def set_interface_arguments():
    global need_restart
    need_restart = True
    return

def create_interface():
    with gr.Blocks(analytics_enabled=False) as interface:
        reset_interface = gr.Button("Apply and restart the interface", type="primary")

        reset_interface.click(set_interface_arguments, None, None)
        reset_interface.click(lambda : None, None, None, _js='() => {document.body.innerHTML=\'<h1 style="font-family:monospace;margin-top:20%;color:lightgray;text-align:center;">Reloading...</h1>\'; setTimeout(function(){location.reload()},2500); return []}')

    # Launch the interface
    interface.queue()
    interface.launch(prevent_thread_lock=True, server_name='0.0.0.0', server_port=9441)

    return interface

interface = create_interface()

while True:
    time.sleep(0.5)
    if need_restart:
        need_restart = False
        interface.close()
        interface = create_interface()
```

### Logs

```shell
Closing server running on port: 9441
/home/me/miniconda3/lib/python3.10/site-packages/gradio/deprecation.py:40: UserWarning: The 'type' parameter has been deprecated. Use the Number component instead.
  warnings.warn(value)
Traceback (most recent call last):
  File "/home/me/miniconda3/lib/python3.10/site-packages/gradio/networking.py", line 119, in start_server
    s.bind((LOCALHOST_NAME, server_port))
OSError: [Errno 98] Address already in use

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/tmp/test.py", line 33, in <module>
    interface = create_interface()
  File "/tmp/test.py", line 22, in create_interface
    interface.launch(prevent_thread_lock=True, server_name='0.0.0.0', server_port=9441)
  File "/home/me/miniconda3/lib/python3.10/site-packages/gradio/blocks.py", line 1454, in launch
    server_name, server_port, local_url, app, server = networking.start_server(
[...]

PR #3959, opened 2023-04-24
After closing demo, reuse port if its specified

freddyaboulton:
# Description

Closes: #3675

Issue was that the socket was not setup to reuse a port.

# Checklist:

# A note about the CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ No changes to highlight.
 - Fixes type in client `Status` enum by [@10zinten](https://github.com/10zinten) in [PR 3931](https://github.com/gradio-app/gradio/pull/3931)
 - Fix `gr.ChatBot` to handle image url [tye-singwa](https://github.com/tye-signwa) in [PR 3953](https://github.com/gradio-app/gradio/pull/3953)
 - Move Google Tag Manager related initialization code to analytics-enabled block by [@akx](https://github.com/akx) in [PR 3956](https://github.com/gradio-app/gradio/pull/3956)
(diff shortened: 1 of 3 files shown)


abidlabs:
I couldn't reproduce the original issue on Windows (is this a rare example where Windows is working better than Linux?) so I can't verify if this fixes it. However, this seems to be the right thing to do given this [StackOverflow answer](https://stackoverflow.com/questions/14388706/how-do-so-reuseaddr-and-so-reuseport-differ):

> The question is, how does the system treat a socket in state TIME_WAIT? If SO_REUSEADDR is not set, a socket in state TIME_WAIT is considered to still be bound to the source address and port and any attempt to bind a new socket to the same address and port will fail until the socket has really been closed. So don't expect that you can rebind the source address of a socket immediately after closing it. In most cases this will fail. However, if SO_REUSEADDR is set for the socket you are trying to bind, another socket bound to the same address and port in state TIME_WAIT is simply ignored, after all its already "half dead", and your socket can bind to exactly the same address without any problem. In that case it plays no role that the other socket may have exactly the same address and port. Note that binding a socket to exactly the same address and port as a dying socket in TIME_WAIT state can have unexpected, and usually undesired, side effects in case the other socket is still "at work", but that is beyond the scope of this answer and fortunately those side effects are rather rare in practice.

Furthermore, this doesn't seem to break anything on Windows so I think this is fine to merge in if this fixes the problem for you @freddyaboulton

abidlabs (approved):

diegomontoya:
Without this fix, linux can take upwards of 30+ seconds to release the tcp port for next launch. Verified this fixed issue on Ubuntu 22.04 with latest stable 6.2 kernel.

freddyaboulton:
Thank you for the review @abidlabs and @diegomontoya !

PR #3960, opened 2023-04-24
CI: only run backend lints once, not in all test environments

akx:
# Description

No need to spend cycles running lints multiple times; just run it once in a separate step.

# Checklist:

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -15,6 +15,7 @@ env:
 
 jobs:
   client-test:
+    needs: [client-lint]
     strategy:
       matrix:
         os: ["ubuntu-latest", "windows-latest"]
@@ -79,6 +80,7 @@ jobs:
     - name: Lint
       run: bash scripts/lint.sh
   test:
+    needs: [lint]
     strategy:
       matrix:
         os: ["ubuntu-latest", "windows-latest"]
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -56,7 +56,28 @@ jobs:
     - name: Tests
       shell: bash
       run: |
-        bash scripts/ci.sh
+        bash scripts/test.sh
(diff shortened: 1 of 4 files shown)


abidlabs:
Hi @akx thanks yet again! This change looks fine to me, but let's run linting before we run the tests. Linting is much faster than running the tests, so better for it to fail quickly if the code is not linted rather than having to wait for all of the tests to run

.github/workflows/backend.yml
```diff
@@ -56,7 +56,28 @@ jobs:
     - name: Tests
       shell: bash
       run: |
-        bash scripts/ci.sh
+        bash scripts/test.sh
+  client-lint:
```
freddyaboulton, line 60: Thanks for the PR @akx . I think `needs` is the right approach here: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds

That way the tests don't run if lint fails

akx:
@abidlabs @freddyaboulton Done deal. `client-test` needs `client-lint`, `test` needs `lint` now.

freddyaboulton (approved):
Thanks @akx !

freddyaboulton:
Can't update the branch and merge @akx  - please merge in main and let me know so we can merge this in.

akx:
@freddyaboulton Rebased, tests were flaky so I re-pushed, we'll see...

freddyaboulton:
Hi @akx - can you reopen the PR and allow edits from maintainers? That way I can merge in main and get this merged! Otherwise might be tricky to align on timing

akx:
@freddyaboulton Sorry, enabled "allow edits" now and rebased.

freddyaboulton:
Thanks again @akx !

abidlabs:
Thanks @akx for addressing the feedback!

Issue #3950, opened 2023-04-24
after prediction, the input and output files remain in the temporary directory.


falibabaei:
Hello, I am using gradio to create a UI for my API. The problem is that after prediction, the input and output files remain in the temporary directory. This causes two problems
1. user privacy
2. I will run out of memory in the future.
Is it possible to remove this automatically?

lewisushindi:
@falibabaei Yes, it's possible to remove the input and output files from the temporary directory automatically. You can use the `tempfile` module in Python to create temporary files and directories, and set them to delete automatically when the program ends.

falibabaei:
Yes, I know I can program it to remove the files after the request is finished, but what I mean is that Gradio removes the files themselves. I  have swagger UI and there I do not need any code to remove the input and output files. The output file is just a stream and after the user is done with the prediction everything is deleted without me having to enter any code.

abidlabs:
Hi @falibabaei the basic reason for this is that Gradio doesn't "know" when a user has stopped using the application and it is safe to clear the temporary files. I'll open this issue up for brainstorming if anyone has any suggestions

tomchang25:
There is an ugly solution in the Gradio interface.
I haven't started tracking the code blocks yet, so I'm not sure if there is a similar solution.

```python
import gradio as gr
import os


def ret(name):
    return name

demo = gr.Interface(
    ret,
    "video",
    "video",
)

if __name__ == "__main__":
    try:
        demo.launch()
    finally:
        print(len(demo.input_components[0].temp_files))
        print(len(demo.input_components))
        print(len(demo.output_components[0].temp_files))
        print(len(demo.output_components))
        for x in demo.input_components[0].temp_files:
            try:
                os.remove(x)
            except OSError as e:
                print(f"Error deleting file: {e}")
            finally:
                print(x)

        for x in demo.output_components[0].temp_files:
            try:
                os.remove(x)
            except OSError as e:
                print(f"Error deleting file: {e}")
            finally:
                print(x)
```

But there are two issues with it: 

1. It requires the user to write additional code
2. The image will not be deleted cleanly when using gr.Image(type='filepath') because there are two temporary files.
![image](https://user-images.githubusercontent.com/31534748/234060691-fee270dd-26f3-41c0-bd67-d503014c8036.png)
![image](https://user-images.githubusercontent.com/31534748/234060778-eba8c764-6d98-4159-99bc-929df63db827.png)

falibabaei:
[...]

PR #3967, opened 2023-04-25
One of the solution to clean tempfile

tomchang25:
# Description

This is a draft solution for addressing issue #3950. Any suggestions or feedback are welcome.
> The input and output files remain in the temporary directory.
### Environment
- [x] Interface
- [ ] Blocks
- [ ] gradio app

- [x] Image
- [ ] File
  - [x] make_temp_copy_if_needed
  - [x] save_uploaded_file
  - [ ] upload_file(Upload from js and `file_manager = gradio.File()`)
  - [ ] process_single_file*
- [ ] Video
- [ ] Audio
- [ ] Any others?

```python
def process_single_file(f) -> bytes | tempfile._TemporaryFileWrapper:
    file_name, data, is_file = (
        f["name"],
        f["data"],
        f.get("is_file", False),
    )
    if self.type == "file":
        if is_file:
            temp_file_path = self.make_temp_copy_if_needed(file_name)
            file = tempfile.NamedTemporaryFile(delete=False)
            file.name = temp_file_path
            file.orig_name = file_name  # type: ignore
        else:
            file = client_utils.decode_base64_to_file(data, file_path=file_name)
            file.orig_name = file_name  # type: ignore
            self.temp_files.add(str(utils.abspath(file.name)))
        return file
```
*The file created by tempfile.NamedTemporaryFile(delete=False) is redundant because it gets immediately overwritten(temp_file_path). Additionally, I think there is another issue discuss about change file returned?

# Checklist:

# A note about the CHANGELOG


abidlabs:
Hi @tomchang25, thanks for opening this PR. However, as far as I understand, this doesn't actually address the issue in #3950. The problem is that the approach in this PR requires a user to call `demo.deconstruct()` after they are finished running the demo. However, in the original issue, I believe @falibabaei had a Gradio app that was continuously running and they wanted the temporary files to be cleared after each prediction was complete. Like I mentioned in the issue, this is hard to do in a generic manner since we don't want to clear the temporary files if the user is still using the demo and we don't have a way of knowing when a user has stopped using the demo.

I'll let @falibabaei chime in case I misunderstood

falibabaei:
@abidlab you are right. I have a container where the app runs, and users can make predictions.

PR #3816, opened 2023-04-11
Fixes chatbot does not display code blocks correctly

dawoodkhan82:
# Description
Fixes issue where the markdown parser renders code blocks in line, instead of retaining new line info.

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 
* any additional dependencies that are required for this change.

Closes: #3531

# Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have added a short summary of my change to the CHANGELOG.md
- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

# A note about the CHANGELOG

Hello 👋 and thank you for contributing to Gradio!

All pull requests must update the change log located in CHANGELOG.md, unless the pull request is labeled with the "no-changelog-update" label.

Please add a brief summary of the change to the Upcoming Release > Full Changelog section of the CHANGELOG.md file and include
a link to the PR (formatted in markdown) and a link to your github profile (if you like). For example, "* Added a cool new feature by `[@myusername](link-to-your-github-profile)` in `[PR 11111](https://github.com/gradio-app/gradio/pull/11111)`".

If you would like to elaborate on your change further, feel free to include a longer explanation in the other sections.
If you would like an image/gif/video showcasing your feature, it may be best to edit the CHANGELOG file using the 
GitHub web UI since that lets you upload files directly via drag-and-drop.

PR #3981, opened 2023-04-26
code block support in chatbot

dawoodkhan82:
# Description
Adds support for code component in the chatbot
Screenshot 2023-04-26 at 2 55 51 PM

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

Closes: # (issue)

# Checklist:

# A note about the CHANGELOG

### DIFF
--- a/js/chatbot/src/ChatBot.svelte
+++ b/js/chatbot/src/ChatBot.svelte
@@ -74,6 +74,15 @@
 									{/each}
 								</div>
 							{/if}
+						{:else if message !== null && message["code_block"] !== undefined}
+							{#await import("@gradio/code") then Code}
+								<Code.default
+									value={message.code_block}
+									language={message.language}
+									mode={"static"}
+									target={div}
+								/>
+							{/await}
 						{:else if message !== null && message.mime_type?.includes("audio")}
 							<audio
 								controls
--- a/js/app/src/components/Chatbot/Chatbot.svelte
+++ b/js/app/src/components/Chatbot/Chatbot.svelte
@@ -20,18 +20,34 @@
 	export let root: string;
 	export let root_url: null | string;
 	export let selectable: boolean = false;
+	const parser = new DOMParser();
 
 	const redirect_src_url = (src: string) =>
 		src.replace('src="/file', `src="${root}file`);
 
+	function process_message(message) {
+		if (typeof message === "string") {
+			const parseHtml = parser.parseFromString(message, "text/xml");
+			const codeElement = parseHtml.getElementsByTagName("code");
+			if (codeElement.item(0)) {
+				return {
+					code_block: codeElement.item(0).innerHTML,
+					language: codeElement
+						.item(0)
+						.classList.item(0)
+						.replace("language-", "")
+				};
+			} else {
+				return redirect_src_url(message);
+			}
+		} else {
+			return normalise_file(message, root, root_url);
+		}
+	}
 	$: _value = value
 		? value.map(([user_msg, bot_msg]) => [
-				typeof user_msg === "string"
-					? redirect_src_url(user_msg)
-					: normalise_file(user_msg, root, root_url),
-				typeof bot_msg === "string"
-					? redirect_src_url(bot_msg)
-					: normalise_file(bot_msg, root, root_url)
+				process_message(user_msg),
+				process_message(bot_msg)
 		  ])
 		: [];
 	export let loading_status: LoadingStatus | undefined = undefined;
--- a/js/chatbot/package.json
+++ b/js/chatbot/package.json
@@ -10,6 +10,7 @@
 	"dependencies": {
 		"@gradio/theme": "workspace:^0.0.1",
 		"@gradio/utils": "workspace:^0.0.1",
-		"@gradio/upload": "workspace:^0.0.1"
+		"@gradio/upload": "workspace:^0.0.1",
+		"@gradio/code": "workspace:^0.0.1"
 	}
 }


abidlabs:
Hmm I wasn't able to get this to work. I tried different variations:

(1) Example with code and text in the same messsage (what we usually get from Chatbot responses)

````py
import gradio as gr

response = """
Hello, here's an example of a program in python:
```py
import gradio as gr

def p(x):
  return x
  
assert p(5) == 5
```
And here's a *simple* program in javascript:
```js
alert('hello')
```
"""

with gr.Blocks() as demo:
    gr.Chatbot([
        ("abc", response)
    ])
demo.launch()
````

Here's how it appears:

image

(2) Simple example with just code:

````py
import gradio as gr

response = """
```python
import gradio as gr
    
def p(x):
  return x
  
assert p(5) == 5
```
"""

with gr.Blocks() as demo:
    gr.Chatbot([
        ("Write some code", response)
    ])
demo.launch()
````

Now nothing appears:

image


Am I missing something?

**Comment by dawoodkhan82** [comment]
@abidlabs weird, I must've broken it with a recent change. Will take a look now.

File: js/chatbot/src/ChatBot.svelte [hunk 1]
```diff
@@ -74,6 +74,15 @@
 									{/each}
 								</div>
 							{/if}
+						{:else if message !== null && message["code_block"] !== undefined}
+							{#await import("@gradio/code") then Code}
+								<Code.default
+									value={message.code_block}
```
aliabid94, line 80: can you add `show_label={false}`

dawoodkhan82:
working with this markdown parser is annoying.

dawoodkhan82:
@abidlabs your two examples should work and look like this now. For the first one with code + text. Do we want to split the message up into multiple messages where the code appears in a code component? I think that makes the most sense.

Screenshot 2023-04-28 at 8 14 04 PM
Screenshot 2023-04-28 at 8 13 48 PM

abidlabs:
@dawoodkhan82 if a message contains both string and code, the code block should be in the same message. That's the standard way it's displayed, e.g.: 

image

abidlabs:
Also I think the code component looks a little weird if `show_label=False` -- it has an empty space in the top left corner @aliabid94

dawoodkhan82:
@abidlabs updated!
Screenshot 2023-04-29 at 6 55 29 PM

I think I can change the min height of the code component. Too much empty space. But otherwise everything should work now. Didn't fully test everything out yet tho

abidlabs:
Looks very nice @dawoodkhan82! However, I'm unable to build this locally to test it out, I get this error when I try to build the frontend:

image

It seems like CI is also having some trouble building it. Any suggestions?

dawoodkhan82:
There's one issue regarding rending the html that I can't seem to figure out. Will sync with @pngwn about it when he gets back.

pngwn:
I commented here: https://github.com/gradio-app/gradio/pull/3816#issuecomment-1504973102 about not using the Code component. It is _far_ too heavy for this usecase. All this needs to be a blob of HTML, it doesn't need to be a fully functioning IDE, which is basically what codemirror is. 

They are entirely different usecases, the usecase here is akin to highlighting some code in a blog post. The Code component is designed with authoring code in mind. 

@abidlabs you said this:
> Sadly, markdown-it-py does not seem to have support for server-side syntax highlighting as far as I could tell.

But I'm not sure that is true (partly because no-one would ever use this lib if it was). 

I checked and the documentation is very bad but you can pass in a highlight option to the `options` Dict: https://github.com/executablebooks/markdown-it-py/blob/d1852a5f8e2cd33602d670cf602593e664c7a2a7/markdown_it/renderer.py#L263-L268 

This should be a function that is passed the code and returns a highlighted string. We can use https://pygments.org/ for the actual highlighting.

dawoodkhan82:
@pngwn I thought since I'm dynamically loading the code component that would help. But I guess I didn't test for having 20+ code components embedded in a single chatbot...

pngwn:
Even if we just had one code block, loading hundreds of kbs for a bit of styling is overkill. The markdown approach is also much simpler from an implementation pov and will also benefit the markdown component.

dawoodkhan82:
@pngwn Yeah that makes sense. I'll leave this branch open, in case we ever revisit it. But I'll look into the markdown approach.

medram:
I was having the same issue the whole day, and a found a fix for it, here is what I did.

```bash
pip install markdown2 pygments
```
Generate a CSS file for styling (using terminal).
```bash
pygmentize -S material -f html -a .codehilite > assets/css/material.css
```

Add the styling to gradio.
```python
with gr.Blocks(css=os.path.abspath("assets/css/material.css")) as demo:
    # normal code here.
```


```python
from markdown2 import Markdown
from pygments import highlight
from pygments.lexers import get_lexer_by_name
from pygments.formatters import HtmlFormatter


def highlight_code(code, lang):
    lexer = get_lexer_by_name(lang, stripall=True)
    formatter = HtmlFormatter()
    return highlight(code, lexer, formatter)

m = Markdown(extras=["fenced-code-blocks", "tables"])
m.fenced_code_block_class = "highlight"
m.code_block_class = "highlight"
m.fenced_code_block_processor = highlight_code

# this tool returns just markdown codes (custom tool).
llm_tool_response = gpt_3_5_llm(self._prompt.format(subject=subject, language=language))

# Clean code to display (markdown to html to be highlighted)
m.convert(llm_tool_response)
```

And with a little of css modification in "material.css" here is the result:
demo
demo2



I hope this helps someone :D

PR #3982, opened 2023-04-27
CI: Simplify Python CI

akx:
# Description

This PR simplifies the backend CI flow, hot on the heels of #3960.


abidlabs:
LGTM, thanks so much @akx!

abidlabs:
Will merge in after (the updated) CI runs

PR #3983, opened 2023-04-27
Deduplicate code in gradio/templates.py by using __init_subclass__

akx:
# Description

* Relevant motivation: I noticed there was a lot of hard-to-maintain copypaste initialization code in gradio/templates.py.
  * In fact, some of the inits' signatures differed from the superclasses' already.
* Summary: Instead of copy-pasting the `__init__`s of the superclasses, use [`__init_subclass__`](https://docs.python.org/3.11/reference/datamodel.html#object.__init_subclass__) in a small mixin class to override the template classes' ctors to pass those kwargs.

# Checklist:

  - Nothing changes for end-users. 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ No changes to highlight.
 ## Testing and Infrastructure Changes:
 
 - CI: Python backend lint is only run once, by [@akx](https://github.com/akx) in [PR 3960](https://github.com/gradio-app/gradio/pull/3960)
+- Deduplicated code in `gradio/templates.py`, by [@akx](https://github.com/akx) in [PR 3983](https://github.com/gradio-app/gradio/pull/3983)
 
 ## Breaking Changes:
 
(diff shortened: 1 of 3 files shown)


abidlabs:
Hmm I don't think this will work @akx because the type hints for the template subclasses are no longer accurate. 

Here is the type hint for `gr.Sketchpad` on main: 

image

And on this branch:

image

akx:
The argument defaults are inaccurate, sure. If that is a problem, then this won't do (but someone should probably do the grunt work to make sure the argument types are otherwise accurate between the subclasses and superclasses, and that any and all arguments added to the superclasses are always reflected in the subclasses...).

abidlabs:
> The argument defaults are inaccurate, sure. If that is a problem, then this won't do 

Yes this is important, I would not be in favor of a solution in which the type hints are inaccurate, even if it reduces code duplication. Will go ahead and close this PR. 

> (but someone should probably do the grunt work to make sure the argument types are otherwise accurate between the subclasses and superclasses, and that any and all arguments added to the superclasses are always reflected in the subclasses...).

I believe we've done this, but if you see any examples where this is not the case, please feel free to point out.

PR #3979, opened 2023-04-26
Ruff update + strictening

akx:
# Description

This PR removes some of the ignores left in the ruff config in #3710, making the configuration stricter (thus catching more bugs).

It fixes those issues where possible, and adds a couple of noqas where it was hard to tell what would be the right fix.

Each commit is quite atomic and can be reviewed separately.

# Checklist:

- [x] I have performed a self-review of my own code
- [x] I have added a short summary of my change to the CHANGELOG.md
- [x] My code follows the style guidelines of this project
- [ ] I have commented my code in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #3984, opened 2023-04-27
Use f-strings where possible

akx:
# Description

* Relevant motivation: f-strings are cleaner to read and generally more performant than concatenation or `.format`.
* Summary of the change: wherever possible, use of `.format` etc. is replaced with f-strings. Initial automated conversion using ikamensh/flynt followed by manual fixes.

# Checklist:

### DIFF
--- a/gradio/interface.py
+++ b/gradio/interface.py
@@ -385,7 +385,7 @@ class Interface(Blocks):
                 if len(self.output_components) == 1:
                     component.label = "output"
                 else:
-                    component.label = "output " + str(i)
+                    component.label = f"output {str(i)}"
 
         if self.allow_flagging != "never":
             if (
@@ -461,9 +461,7 @@ class Interface(Blocks):
     def render_title_description(self) -> None:
         if self.title:
             Markdown(
-                "<h1 style='text-align: center; margin-bottom: 1rem'>"
-                + self.title
-                + "</h1>"
+                f"<h1 style='text-align: center; margin-bottom: 1rem'>{self.title}</h1>"
             )
         if self.description:
             Markdown(self.description)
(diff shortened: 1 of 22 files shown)


gradio/interface.py
```diff
@@ -385,7 +385,7 @@ def clean_html(raw_html):
                 if len(self.output_components) == 1:
                     component.label = "output"
                 else:
-                    component.label = "output " + str(i)
+                    component.label = f"output {str(i)}"
```
abidlabs, line 388:
```suggestion
                    component.label = f"output {i}"
```

gradio/utils.py
```diff
@@ -907,7 +904,7 @@ def tex2svg(formula, *args):
     DPI = 300
     plt.rc("mathtext", fontset="cm")
     fig = plt.figure(figsize=(0.01, 0.01))
-    fig.text(0, 0, r"${}$".format(formula), fontsize=FONTSIZE)
+    fig.text(0, 0, f"${formula}$", fontsize=FONTSIZE)
```
abidlabs, line 907:
```suggestion
    fig.text(0, 0, fr"${formula}$", fontsize=FONTSIZE)
```

abidlabs:
Thanks so much for fixing these @akx! Just noticed a couple of issues, which I'll commit and then merge in this branch

gradio/utils.py
akx: It doesn't matter, since $ isn't an escape sequence...

akx:
> Thanks so much for fixing these @akx! Just noticed a couple of issues, which I'll commit and then merge in this branch

You're welcome! Should probably enable Ruff UP032 to catch these in the future, but maybe that's after #3979 is merged (the full Ruff UP suite will also change type annotations to new-style ones, which should be fine since most of the codebase is `from __future__ import annotations` anyway... anyway, that's a separate future PR :) )

gradio/utils.py
abidlabs: I think you're right lol

Issue #934, opened 2022-04-06
Implement minimum and maximum to the number input


apolinario:
**Is your feature request related to a problem? Please describe.**
It is not always I need a number slider, however I am forced to use it because I want to constraint the user to a certain range of values

**Describe the solution you'd like**
Maximum and minimum values could take many forms:
- If the user types a lower or higher value, the value could "snap" to the highest or lowest allowed value
- If the user went lower or higher, it could show that this is an invalid field and gray out the submit button

Edit: Step and Randomize functions could be added as well

abidlabs:
This is a good suggestion! We would want to communicate the range clearly to the user as well. Do you have thoughts on how that might be communicated?

apolinario:
A few ideas: 
- Have the allowed value range on the placeholder
- Have the allowed value range with some other color next or below the label
- Have the allowed value range below the input field itself

pngwn:
A common pattern is to use labelled 'pips'. Little lines underneath the range input. Sometimes they are used to show 'steps' but they are commonly used to show min + max values.

This kind of thing.

![pips](https://user-images.githubusercontent.com/12937446/162259126-fe908367-2540-49ba-a0cf-6c77e94cd0c5.png)

apolinario:
Hi @pngwn! I love your suggestion and I think it could make a lot of sense to the Slider input

But just a a clarification, this feature request actually referred to the possibility of enabling maximum and minimum values to the `gradio.inputs.Number` field

apolinario:
Also for the way to inform the user @abidlabs , I don't love this one only happens after they inputed something wrong, but could be a food for thought.

(PS: This example contains a way to add the "Steps" option to the `inputs.Number`, I didn't open an issue for it but I would love this as a feature too) 

https://user-images.githubusercontent.com/788417/162432035-f9e1671c-7e51-49f4-a35f-54f0e9f0d592.mov

Issue #3878, opened 2023-04-16
Prevent `gradio_client` from sending out of range value for `gr.Slider`


hysts:
- [x] I have searched to see if a similar issue already exists.

When using `gr.Slider`, users can restrict the possible input values with `minimum, maximum, step`. For example, we can limit the value to multiples of 5 from 0 to 100 by doing `gr.Slider(minimum=0, maximum=100, step=5, value=0)`, but currently `Client` can send arbitrary values, like `1000000` or `1.23`, ignoring this restriction.
This is problematic, for example, when the batch size is set with `gr.Slider`. `Client` can send a value like 100, which would probably cause GPU OOM and crash the Space. Once GPU OOM occurs, a Space will stop working properly until manually restarted.
I guess Space developers can add some logic in their functions to reject unacceptable values, but it will increase their work. So, it would be nice if it was taken care of on gradio side.

missionfloyd:
Clamp it.
```python
def clamp(n, smallest, largest):
    return max(smallest, min(n, largest))

batch_size = clamp(batch_size, 0, 4)
```

freddyaboulton:
I don't think the best fix is at the client level. Even if you set `steps` in the app, the front-end won't prevent you from typing a value for the slider that's out of range or incorrect precision.

<img width="1346" alt="image" src="https://github.com/gradio-app/gradio/assets/41651716/64949298-1d16-4d25-b22a-5ab4246cc600">

I think the better approach is to tackle #2718 and then if the client sends a request with an invalid value it will automatically fail.

PR #3991, opened 2023-04-27
gr.Number() max/min limit

dawoodkhan82:
# Description
Building off of [artegoser](https://github.com/artegoser) PR (which is merged into this branch). Adding frontend support for min/max of `gr.Number()` component.

As of now, the border color for the input turns red when a number out of range is typed in.

Screenshot 2023-04-27 at 3 57 34 PM

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

Closes: #934

# Checklist:

# A note about the CHANGELOG


abidlabs:
Very cool @dawoodkhan82! This is the first time we're doing any sort of validation in the frontend, so cc @pngwn @aliabid94 @gary149 on the design piece.

gradio/components.py
```diff
@@ -667,15 +674,41 @@ def _round_to_precision(num: float | int, precision: int | None) -> float | int:
...
+            "max": self.max,
+            "min": self.min,
             **IOComponent.get_config(self),
         }
 
     @staticmethod
     def update(
         value: float | Literal[_Keywords.NO_VALUE] | None = _Keywords.NO_VALUE,
+        max: float | int = float("-inf"),
+        min: float | int = float("-inf"),
```
abidlabs, line 711: I believe these should be `None`. Otherwise, calling `.update()` will reset the min/max:
 ```suggestion
        max: float | int | None = None,
        min: float | int | None = None,
```

**Comment by abidlabs** [comment]
I tested this code: 

```py
import gradio as gr

gr.Interface(lambda x:x, gr.Number(min=10, max=20), gr.Number(min=15, max=25)).launch()
```

and on the default theme, it looks and works great:

image

However, on all of the other themes, I don't see any validation error bar:

```py
import gradio as gr

gr.Interface(lambda x:x, gr.Number(min=10, max=20), gr.Number(min=15, max=25), theme="monochrome").launch()
```

image

image

artegoser:
Okay, I changed my mind.

aliabid94:
So if a user submits a number out of range, it gets clamped in the backend? That might be a bit "magical", a user might think the function ran on what they entered rather than the clamped value.

Proposal that may be out of the scope of this: I think we need to allow blocking an event trigger if an input component has an invalid input. Another use-case might be an `optional` kwarg, that when set to False, behaves as such:
- for Image/Audio/File/etc component , something must be uploaded for that event to trigger 
- for Dropdown/Radio/CheckboxGroup with `optional=False`, something must be selected for that event to trigger
- Textboxes/Numbers cannot be empty

If an event tries to trigger with an empty `optional=False` component, the StatusTracker for that component could become a red outline, similar to how in "generating" mode the StatusTracker has a glowing orange outline.

abidlabs:
> So if a user submits a number out of range, it gets clamped in the backend? That might be a bit "magical", a user might think the function ran on what they entered rather than the clamped value.

Ohh good point @aliabid94. Maybe we raise a `gr.Error()` instead of clamping the values?

> Proposal that may be out of the scope of this: I think we need to allow blocking an event trigger if an input component has an invalid input. Another use-case might be an optional kwarg, that when set to False, behaves as such:

Yup this would be handy also to block the Submit button while a file is uploading, but out of scope of this PR for sure

artegoser:
Maybe then it would be better to limit the number in the frontend, too, so that the user can't add a number that is out of bounds. When the number changes, check it and limit it if necessary.

abidlabs:
> Maybe we raise a gr.Error() instead of clamping the values?

Can we do the same for `Slider`? Would close #3878

abidlabs:
Let's get this merged in @dawoodkhan82. For now, the only change that needs to be made is that we should raise a `gr.Error()` if a user submits a value that is outside of the range of the `gr.Number()`, instead of clamping it. 

Later on, we can implement @aliabid94's suggestion, which is to block the submit button altogether

cc @hannahblair if you have any suggestions on validation of the `gr.Number()`

dawoodkhan82:
@abidlabs ready for re-review

abidlabs:
I think this breaks if min/max are not set. I tried this:

```py
import gradio as gr

gr.Interface(lambda x:x, gr.Number(), gr.Number()).launch()
```

And I got:

image

Furthermore, looking at the code, I think it's better if we raise the `gr.Error()` in the backend (in the `Number.preprocess()` function) as opposed to the frontend. That way, it also applies if someone tries to use the app via the Python or JS clients.

abidlabs:
Also, do the `min`/`max` have any effect if the the component is used as an output. I haven't tested but if they do not have any effect, it would be good to explicitly mention in the docstring that the `min`/`max` only apply if `Number` is used as an input.

dawoodkhan82:
> I think this breaks if min/max are not set. I tried this:
> 
> ```python
> import gradio as gr
> 
> gr.Interface(lambda x:x, gr.Number(), gr.Number()).launch()
> ```
> 
> And I got:
> 
> image
> Furthermore, looking at the code, it would be better if we raise the `gr.Error()` in the backend (in the `Number.preprocess()` function) as opposed to the frontend. That way, it also applies if someone tries to use the app via the Python or JS clients.

Moved raising `gr.Error` to the backend. The error is raised in the console, but the modal doesn't seem to popup... is this a bug? @abidlabs

abidlabs:
Interesting @dawoodkhan82 it's actually showing up fine for me!

image

This is running on Chrome in MacOS after I built the frontend. Lmk if you wanna sync!

gradio/components.py
```diff
@@ -670,8 +673,13 @@ def __init__(
             elem_id: An optional string that is assigned as the id of this component in the HTML DOM. Can be used for targeting CSS styles.
             elem_classes: An optional list of strings that are assigned as the classes of this component in the HTML DOM. Can be used for targeting CSS styles.
             precision: Precision to round input/output to. If set to 0, will round to nearest integer and convert type to int. If None, no rounding happens.
+            min: Minimum value. Only applied when component is used as an input.
```
abidlabs, line 676:
```suggestion
            min: Minimum value. Only applied when component is used as an input. If a user provides a smaller value, a gr.Error exception is raised by the backend.
```

gradio/components.py
```diff
@@ -670,8 +673,13 @@ def __init__(
...
+            max: Maximum value. Only applied when component is used as an input.
```
abidlabs, line 677:
```suggestion
            max: Maximum value. Only applied when component is used as an input. If a user provides a larger value, a gr.Error exception is raised by the backend.
```

abidlabs:
Some nits above & tests need to be fixed, but otherwise LGTM!

abidlabs (approved):

pngwn (approved):

js/app/src/components/Number/Number.svelte
```diff
@@ -29,6 +31,8 @@
 
 	<Number
 		bind:value
+		bind:max
```
aliabid94, line 34: no need to bind max and min, can just use `{max} {min}` instead, as values will only be passed down and cannot be passed back up

aliabid94:
In Slider, we've called the arguments `minimum` and `maximum`. I feel like we should use the same arg names (could change slider to use min and max, and then pull minimum and maximum out of kwargs with a warning for backwards compatibility)

aliabid94:
The error border theme variable that you've changed also affects the error messages. See below the pink borders. 
Screen Shot 2023-06-07 at 2 30 58 PM
Screen Shot 2023-06-07 at 2 32 18 PM

akx:
How are there 177 commits in this PR..?

Issue #3994, opened 2023-04-27
Dropdown default not displaying in 3.28.0


jhj0517:
### Describe the bug

Thank you for your work.
I noticed that Gradio was recently updated to version 3.28.0 a few hours ago, 
but it seems that the Dropdown default value is not being displayed.

I confirmed that the default value is indeed "selected" within the Dropdown, but it doesn't "appear" to be shown as expected.
Probably this issue is related with HTML rendering?

I came across a similar bug that occurred when using  `gr.update` in gradio 3.2, but this issue is not related to the `gr.update`.

- similar bug
   #2153 

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```
def fn(value):
    print(f"{value}")

class App:
    def __init__(self):
        self.app = gr.Blocks()

    def launch(self):
        with self.app:
            with gr.Row():
                dd_models = gr.Dropdown(label="Model", value="model1", choices=["model1", "model2", "model3"])
                btn_print = gr.Button("PRINT VALUE", variant="primary")
            btn_print.click(fn=fn, inputs=[dd_models], outputs=None)
        self.app.queue().launch()


if __name__ == "__main__":
    app = App()
    app.launch()
```

I implemented the same code in a Huggingface Space, but still can't see the default value.
 
[huggingface space for reproduce](https://huggingface.co/spaces/jhj0517/DropDown_Error_Reproduce)  
    

### Screenshot

![image](https://user-images.githubusercontent.com/97279763/235003195-f8e7bd28-46f2-4cd4-8d3e-ae69576156ae.png)

### Logs

```shell
N/A
```

### System Info

```shell
Gradio 3.28.0, Win10, Edge/Chrom
```

### Severity

annoying

abidlabs:
Thanks @jhj0517 for the catch! Issuing a fix now

PR #3996, opened 2023-04-27
Fix dropdown default issue

aliabid94:
Fixes: https://github.com/gradio-app/gradio/issues/3994

Try with: 
```python
import gradio as gr

choices = ["World", "Gradio", "World2", "abc "*50, "You"]
with gr.Blocks() as demo:
    drop1 = gr.Dropdown(choices, label="simple", value="World")
    drop2 = gr.Dropdown(choices, label="multi", multiselect=True)
    drop3 = gr.Dropdown(choices, label="custom", allow_custom_value=True)
    drop4 = gr.Dropdown(choices, label="multi-max3", multiselect=True, max_choices=3)
    btn = gr.Button("Click me!")
    with gr.Row():
        with gr.Column():
            gr.Markdown("On Click")
            val1 = gr.Textbox(label="simple")
            val2 = gr.Textbox(label="multi")
            val3 = gr.Textbox(label="custom")
            val4 = gr.Textbox(label="multi-max3")
        with gr.Column():
            gr.Markdown("On Change")
            ins1 = gr.Textbox(label="simple")
            ins2 = gr.Textbox(label="multi")
            ins3 = gr.Textbox(label="custom")
            ins4 = gr.Textbox(label="multi-max3")
    btn.click(lambda *args: args, [drop1, drop2, drop3, drop4], [val1, val2, val3, ins4])
    drop1.change(lambda x:x, drop1, ins1)
    drop2.change(lambda x:x, drop2, ins2)
    drop3.change(lambda x:x, drop3, ins3)
    drop4.change(lambda x:x, drop4, ins4)
        
demo.launch()
```


abidlabs:
LGTM, thanks for the quick PR @aliabid94!

abidlabs (approved):

abidlabs:
Actually hold on, the deployed Spaces are not loading. Seeing this error in the console:

image

abidlabs (changes requested):
See console error above. Seems related to this change

js/form/src/Dropdown.svelte
```diff
@@ -25,6 +25,10 @@
 		showOptions = false,
 		filterInput: HTMLElement;
 
+	if (typeof value === "string") {
```
freddyaboulton, line 28: If the problem is that non-string values can't be displayed in the drop-down, then I think it would be better to do the coercion in the postprocess method so that even if someone accidentally updates the value to be a non-string after the app launches then we will be covered too.

And I think at that point, we can get rid of this `if` on the front-end since we're guaranteeing that the value will be string so we can always set inputValue=value on init? 


Also think it would be ok to raise an error on the backend would be ok in this case.

aliabid94: Actually the reason for this is to exclude the cases of `multiselect=True` (where value would be a list of strings) or if it's null. The inputValue is only set in the case where there is a value and it is a single string, and this is the simplest way to do it to also pass type checks. This isn't a guard against improper types being passed by a user.

abidlabs:
LGTM @aliabid94!

abidlabs (approved):

js/form/src/Dropdown.svelte
freddyaboulton: Gotcha!

freddyaboulton (approved):

abidlabs:
Merging, thanks guys!

PR #4009, opened 2023-04-28
Version 3.28.1

abidlabs:


freddyaboulton:
In the deployed space in the `blocks_mask` demo, taking a photo with `5c` makes the image component disappear.

blocks_mask_5c_bug

abidlabs:
@freddyaboulton I can confirm the issue, but it looks like it was a pre-existing condition: https://huggingface.co/spaces/gradio-pr-deploys/pr-3996-all-demos

abidlabs:
I suggest we make that an issue and release this version if everything else looks ok

freddyaboulton:
Sounds good @abidlabs ! I can file an issue for that. 

The other thing I noticed was that in `blocks_update` the animal drop down is not updated when the species is updated. This may be due to the recent dropdown changes.

dropdown_update_bug

abidlabs:
Ah you're right here's a simple repro:

```py
import gradio as gr

choices = ["apple", "banana"]

gr.Interface(lambda x:x, gr.Dropdown(choices=choices), gr.Dropdown(choices=choices)).launch()
```

This is a pretty bad bug. @aliabid94 could you take a look at this?

freddyaboulton:
Rest of the deployed demos look good to me!

aliabid94:
Ok I think I fixed it in this PR

abidlabs:
LGTM, thanks @aliabid94! The failing tests are because of Spaces-related issues as far as I can tell, so I will go ahead and merge!

Issue #4018, opened 2023-04-30
User warning Matplotlib GUI while streaming to chatbot


bent-verbiage:
### Describe the bug

When streaming from an LLM to the chatbot, once every so often the error below is shown. I'm not sure if it has issues with  a specific token, or that there is a delay in the LLM response which causes this issue. 

It doesn't lead to crashes, so not a stopping issue at the moment.

```
site-packages/gradio/utils.py:906: UserWarning: Starting a Matplotlib GUI outside of the main thread will likely fail.
  fig = plt.figure(figsize=(0.01, 0.01))
2023-04-30 13:36:32.519 python[70501:1579381] WARNING: NSWindow drag regions should only be invalidated on the Main Thread! This will throw an exception in the future.
```

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```
from queue import Queue
from typing import Any

import gradio as gr
from anyio.from_thread import start_blocking_portal
from langchain.callbacks.base import AsyncCallbackHandler, AsyncCallbackManager
from langchain.chat_models import ChatOpenAI
from langchain.schema import HumanMessage


class StreamingLLMCallbackHandler(AsyncCallbackHandler):
    """Callback handler for streaming LLM responses to a queue."""

    def __init__(self, q):
        self.q = q

    def on_llm_new_token(self, token: str, **kwargs: Any) -> None:
        self.q.put(token)


def bot(chatbot):
    user_msg = chatbot[-1][0]

    prompt = user_msg
    q = Queue()
    job_done = object()

    async def task(prompt):
        llm = ChatOpenAI(
            verbose=True,
            temperature=0,
            streaming=True,
            callback_manager=AsyncCallbackManager([StreamingLLMCallbackHandler(q)]),
        )

        ret = await llm.agenerate([[HumanMessage(content=prompt)]])
        q.put(job_done)
        return ret

    with start_blocking_portal() as portal:
        portal.start_task_soon(task, prompt)

        content = ""
        while True:
            next_token = q.get(True, timeout=1)
            if next_token is job_done:
                break
            content += next_token
            chatbot[-1][1] = content

            yield chatbot


with gr.Blocks() as demo:
    chatbot = gr.Chatbot()
    msg = gr.Textbox(label="Prompt")

    def chat(user_message, history):
        return "", history + [[user_message, None]]

    msg.submit(chat, [msg, chatbot], [msg, chatbot], queue=False).then(
        bot, chatbot, chatbot
    )

[...]

Issue #3943, opened 2023-04-23
Streaming Chatbot component crashes gradio process when the text includes two $-signs


rueckstiess:
### Describe the bug

I'm using the example code for a Chatbot component with streaming from here: https://gradio.app/creating-a-chatbot/#add-streaming-to-your-chatbot. I'm running `gradio app.py` locally on my Macbook M1 Pro. 

When my bot response includes two `$` signs, the `gradio` crashes with a `NSInternalInconsistencyException` and a stack trace. It looks like it's trying to open a Matplotlib window outside of the main thread, which causes the exception. Perhaps this is related to rendering LaTeX equations with the `$` as delimiters? If this is the cause, is there a way to disable the behavior, as I don't need any LaTeX rendering in my chat UI?

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Save the following code as `app.py` and run it with `gradio app.py` locally. Then open the URL, type anything into the text box and hit enter. 

(Note the two `$` signs in the hard-coded bot response.)

```
import gradio as gr
import random
import time

with gr.Blocks() as demo:
    chatbot = gr.Chatbot()
    msg = gr.Textbox()
    clear = gr.Button("Clear")

    def user(user_message, history):
        return "", history + [[user_message, None]]

    def bot(history):
        bot_message = "I want to spend $1000 on travelling and use the remaining $4000 to buy a car."
        history[-1][1] = ""
        for character in bot_message:
            history[-1][1] += character
            time.sleep(0.05)
            yield history

    msg.submit(user, [msg, chatbot], [msg, chatbot], queue=False).then(
        bot, chatbot, chatbot
    )
    clear.click(lambda: None, None, chatbot, queue=False)

demo.queue()
demo.launch()
```

### Screenshot

<img width="800" alt="Screenshot 2023-04-23 at 16 22 29" src="https://user-images.githubusercontent.com/99221/233823549-54211451-91e9-4e0f-af5a-7609dbf1bfb8.png">

### Logs

```shell
❯ gradio app.py

Launching in *reload mode* on: http://127.0.0.1:7860 (Press CTRL+C to quit)

Watching: '/Users/tr/code/python/ml-demo-apps/chatbot-docs/.venv/lib/python3.10/site-packages/gradio', '/Users/tr/code/python/ml-demo-apps/chatbot-docs'

Running on local URL:  http://127.0.0.1:7861

To create a public link, set `share=True` in `launch()`.
[...]

PR #4029, opened 2023-05-01
Sets matplotlib backend to agg before rendering math

abidlabs:
This PR:

* sets the `matplotlib` backend to `Agg` before trying to render LaTeX (this is required for LaTeX to render properly)
* warns a user if `matplotlib` backend is not `Agg` when using the `gr.Plot` component with a `matplotlib` plot 

Closes: #4018, Closes: #3943 (hopefully, but needs to be tested on a Mac)


gradio/utils.py
```diff
@@ -900,34 +899,45 @@ def __str__(self):
...
+        height_match = re.search(r'height="([\d.]+)pt"', svg_code)
+        if height_match:
+            height = float(height_match.group(1))
+            new_height = height / FONTSIZE  # conversion from pt to em
+            svg_code = re.sub(
+                r'height="[\d.]+pt"', f'height="{new_height}em"', svg_code
+            )
+        copy_code = f"<span style='font-size: 0px'>{formula}</span>"
+        return f"{copy_code}{svg_code}"
+    finally:
```
abidlabs, line 939: Note: `finally` blocks get executed before the `return` above even if the `try` block runs successfully

gradio/components.py
```diff
@@ -4856,6 +4856,10 @@ def postprocess(self, y) -> Dict[str, str] | None:
         if y is None:
             return None
         if isinstance(y, (ModuleType, matplotlib.figure.Figure)):  # type: ignore
+            if not matplotlib.get_backend().lower() == "agg":
+                warnings.warn(
+                    "Matplotlib backend is not set to 'Agg'. Plot may not display correctly. To fix this, add the following to the top of your script: import matplotlib; matplotlib.use('Agg')"
```
freddyaboulton, line 4861: I wonder if we can raise this somewhere up the chain (plot component init?). May be some false warnings but I worry the app would have already crashed before this is printed.

freddyaboulton (approved):
Can confirm this fixes the original problem! Code looks good - only one comment. Thanks @abidlabs !

gradio/components.py
abidlabs: Good point, let me think about this some more as we don't want too many false warnings.

abidlabs: What if we were to try/catch the function execution, specifically looking for "main thread is not in main loop"

freddyaboulton: Yea I think that will work

gradio/utils.py
```diff
@@ -899,34 +899,48 @@ def __str__(self):
         return ""
 
 
+class MatplotlibBackendMananger:
```
freddyaboulton, line 902: Nice!

abidlabs:
Refactored the matplotlib logic into a context manager so now we can use it everywhere. This has the advantage of users not needing to put `matplotlib.use("Agg")` at the top of their code. Most of the additional changes are just removing these extraneous lines from our demos.

@freddyaboulton would you be able to give this another pass?

freddyaboulton (approved):
There's a call to `process_api` in the Blocks `__call__` method but maybe that's ok since it's running in the parent thread.

Anyways, this LGTM! Thanks @abidlabs !

abidlabs:
> There's a call to process_api in the Blocks __call__ method but maybe that's ok since it's running in the parent thread.

Yes, I think that one should work as is!

Thanks for reviewing! Will take a look at these tests and then merge in.

abidlabs:
Strange, all of these tests are passing for me locally on Windows...

abidlabs:
Tests are passing, will merge. Thanks once again @freddyaboulton for the review!

Issue #4020, opened 2023-04-30
Error when using State component in a TabbedInterface


beskrovnykh:
### Describe the bug

I was attempting to use a State component in a TabbedInterface to share state between different tabs. However, when running the application, it crashed with the error "DuplicateBlockError: At least one block in this Blocks has already been rendered." Upon further investigation, I discovered that the State component was being treated as a rendered block, even though it was not supposed to be. I believe that this is a bug in the Gradio library. A workaround is to use low-level blocks instead of interfaces, but this requires a significant rewrite of the code.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

1. Create a new gradio interface with two tabs using TabbedInterface.
2. Inside each tab, add an interface component that uses a shared state variable as shown below
3. Try to run the app.
4. The error DuplicateBlockError: At least one block in this Blocks has already been rendered. should appear in the console.

```
import gradio as gr

state = gr.State("")
gr.TabbedInterface([
    gr.Interface(lambda _, x: (x, f"I don't know"), inputs=[state, gr.Textbox()], outputs=[state, gr.Textbox()]),
    gr.Interface(lambda s: (s, f"User question: {s}"), inputs=[state], outputs=[state, gr.Textbox(interactive=False)])
], ["Ask question", "Show question"], title="Example of usage state variable in tabbed environment.").launch()
```

### Logs

```shell
Traceback (most recent call last):
  File "/Users/beskrovnykh/Work/ws/projects/danielsearch-embeddings/tabbed_app_with_state.py", line 4, in <module>
    gr.TabbedInterface([
  File "/Users/beskrovnykh/Work/ws/projects/danielsearch-embeddings/venv/lib/python3.8/site-packages/gradio/interface.py", line 864, in __init__
    interface.render()
  File "/Users/beskrovnykh/Work/ws/projects/danielsearch-embeddings/venv/lib/python3.8/site-packages/gradio/blocks.py", line 848, in render
    raise DuplicateBlockError(
gradio.exceptions.DuplicateBlockError: At least one block in this Blocks has already been rendered.
```

### System Info

```shell
3.27.0
```

### Severity

blocking all usage of gradio

beskrovnykh:
I looked in the debugger for what is the reason, the check below does not pass
```
 if not set(Context.root_block.blocks).isdisjoint(self.blocks):
      raise DuplicateBlockError(
          "At least one block in this Blocks has already been rendered."
      )
```

The problem is that

[...]

PR #4030, opened 2023-05-01
Allow state to be rendered multiple times

abidlabs:
Fixes: #4020 

Generally, we don't allow components to be re-used between Blocks. However, we should make an exception for `gr.State`, which actually can be legitimately reused between different Blocks as the original issue shows. This carves out an exception for `gr.State` and adds a test. See original issue for code to test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@
 
 ## Bug Fixes:
 
-No changes to highlight.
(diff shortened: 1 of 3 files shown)


freddyaboulton (approved):
Looks good to me! Thanks @abidlabs

abidlabs:
Thanks @freddyaboulton!

PR #4033, opened 2023-05-02
Push to spaces

aliabid94:
Added ability to push a gradio app directly to spaces, from the CLI: `gradio deploy`

Upload folder with python script from CLI:
upload-script


gradio/upload.py
```diff
@@ -0,0 +1,173 @@
...
+import huggingface_hub
+import json
+
+
+def get_existing_configuration(readme_file):
+    with open(readme_file, "r") as f:
+        readme = f.read()
+    if not readme.strip().startswith("---"):
+        return None
+    readme = readme.strip()
```
akx, line 12:
```suggestion
    with open(readme_file, "r") as f:
        readme = f.read().strip()
    if not readme.startswith("---"):
        return None
```

gradio/upload.py
```diff
@@ -0,0 +1,173 @@
...
+    if "---" not in readme[3:]:
+        return None
+    readme = readme.split("---")[1].strip()
+    config = {}
+    for line in readme.split("\n"):
+        if ":" not in line:
+            continue
+        key, value = line.split(":", 1)
+        config[key.strip()] = value.strip()
```
akx, line 21: `str.partition` is faster:
```suggestion
        key, sep, value = line.partition(":")
        if sep:
            config[key.strip()] = value.strip()
```

gradio/upload.py
```diff
@@ -0,0 +1,173 @@
...
+---
+"""
+    with open(readme_file, "r") as f:
+        readme = f.read()
+    with open(readme_file, "w") as f:
+        f.write(readme_header + readme)
+
+
+def notebook_to_script(notebook_path, script_path):
+    print(f"Converting {notebook_path} to {script_path}.")
```
akx, line 80: You might want to look at [`nbconvert`](https://nbconvert.readthedocs.io/en/latest/usage.html#convert-script) instead of rolling your own.

gradio/upload.py
```diff
@@ -0,0 +1,173 @@
+import os
+import random
+import huggingface_hub
+import json
+
+
+def get_existing_configuration(readme_file):
```
Wauplin, line 7: Can be replaced to use `huggingface_hub`'s parsing method `metadata_load`:

```py
>>> from huggingface_hub import metadata_load
>>> metadata_load("/home/wauplin/projects/git-vs-http-benchmark/README.md")
{'title': 'Git Vs Http Benchmark', 'emoji': '🐢', 'colorFrom': 'indigo', 'colorTo': 'blue', 'sdk': 'gradio', 'sdk_version': '3.17.0', 'app_file': 'app.py', 'pinned': False}
```

You also have `metadata_save` to save metadata back to a README file while keeping the existing "raw" content. Both methods should be robust to special cases (like \r\n line break on Windows).

gradio/upload.py
```diff
@@ -0,0 +1,173 @@
...
+colorFrom: {color}
+colorTo: {color}
+sdk: gradio
+{f'notebook_file: {notebook_file}' if notebook_file else ''}
+---
+"""
+    with open(readme_file, "r") as f:
+        readme = f.read()
+    with open(readme_file, "w") as f:
+        f.write(readme_header + readme)
```
Wauplin, line 76: As mentioned above, `huggingface_hub.metadata_save` can do this part for you.

gradio/upload.py
```diff
@@ -0,0 +1,173 @@
...
+        not os.path.exists(requirements_file)
+        and input("Create requirements.txt file? [y/n]: ").lower() == "y"
+    ):
+        with open(requirements_file, "w") as f:
+            f.write("\n".join(dependencies))
+        input(
+            "Created requirements.txt file. Please add any dependencies to this file. Press enter to continue. "
+        )
+    name = whoami["name"]
+    space_id = f"{name}/{configuration['title']}"
```
Wauplin, line 160: This line will break if title is not a valid ID.

For example the [Dolly V2](https://huggingface.co/spaces/RamAnanth1/Dolly-v2/blob/main/README.md) Space has title `Dolly V2` but space id `Dolly-v2`. In general the rule is _"Only regular alphanumeric characters, "-", ".", "`_`" supported."_ (also no consecutive double "-", not starting with ".").

gradio/cli.py
```diff
@@ -0,0 +1,12 @@
+import sys
+import gradio.reload
+import gradio.upload
+
+def cli():
+    args = sys.argv[1:]
+    if len(args) == 0:
+        raise ValueError("No file specified.")
+    if args[0] == "--upload-to-spaces":
```
Wauplin, line 9: Really a nit but I'd rather name it "--deploy". It's more concise (quicker to type) and implicitly make "spaces" the default way of deploying a Gradio app. But this is personal opinion so feel free to ignore.

Wauplin (approved):
Really nice feature! It'll be very handy to iterate over the implementation of a Space. I left a few comments about stuff that could be delegated to `huggingface_hub`, otherwise the use of `upload_folder` is great :)

If possible, I would also add the possibility in the CLI prompts to set a custom hardware and secrets. Hardware can be requested at repo creation with [`space_hardware`](https://huggingface.co/docs/huggingface_hub/package_reference/hf_api#huggingface_hub.HfApi.create_repo) parameter (requires hfh>=0.14). For secrets it's not possible at repo creation but you can use [`add_space_secret`](https://huggingface.co/docs/huggingface_hub/v0.14.1/en/package_reference/hf_api#huggingface_hub.HfApi.add_space_secret).

gradio/upload.py
```diff
@@ -0,0 +1,173 @@
...
+
+    huggingface_hub.create_repo(
```
Wauplin, line 162: `create_repo` returns a RepoUrl object with a repo_id attribute. Instead of doing "whoami" to get the namespace, it can be implicitly inferred by the server and returned.

So instead of:
```py
space_id = f"{name}/{configuration['title']}"
huggingface_hub.create_repo(space_id, ...)
```

you should do
```py

space_id = huggingface_hub.create_repo(configuration["xxx"], ...).repo_id
```

That way, user can be prompted to chose the repo_id with or without namespace (e.g. either `dolly-v2` => implicit `username/dolly-vy` or `my-organization/dolly-v2` => explicit org)

Wauplin (commented):
added more comments about `whoami` and implicit/explicit repo ids

gradio/cli.py
aliabid94: Agreed!

gradio/upload.py
```diff
@@ -0,0 +1,173 @@
...
+    try:
+        whoami = hf_api.whoami()
+        if whoami["auth"]["accessToken"]["role"] != "write":
+            login = True
+    except:
+        login = True
+    if login:
+        print("Need 'write' access token to create a Spaces repo.")
+        huggingface_hub.login(add_to_git_credential=False)
+        whoami = hf_api.whoami()
```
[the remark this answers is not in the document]
julien-c: Thanks for the great review @Wauplin and agree with your suggestion: **it's important IMO to reuse as much as possible the logic from `huggingface_hub`** otherwise we would be reinventing the wheel and increase brittleness

gradio/upload.py
aliabid94: reason why I didn't:
- didn't want to add a dependency
- want to use custom logic to convert !shell commands to subprocess calls

gradio/upload.py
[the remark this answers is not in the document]
aliabid94: > it's important IMO to reuse as much as possible the logic from huggingface_hub otherwise we would be reinventing the wheel and increase brittleness

Just to be clear, there is no existing logic for this in hf_hub library, correct? and that @Wauplin was suggesting an api within hf_hub that does not yet exist?

gradio/upload.py
aliabid94: Created custom function to format the title, see `format_title`

aliabid94:
Think I addressed everything!

gradio/upload.py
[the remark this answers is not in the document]
Wauplin: Yes exact. Would you mind opening a PR to integrate your logic in `huggingface_hub`? :pray: 

For what I see, `overwrite: bool=True` and  `write_permission: bool=False` should be added to `login`, `notebook_login`, `interpreter_login` and `_login` [in this file](https://github.com/huggingface/huggingface_hub/blob/main/src/huggingface_hub/_login.py#L39).

gradio/deploy_space.py
```diff
@@ -0,0 +1,183 @@
...
+
+    color_set = ["red", "yellow", "green", "blue", "indigo", "purple", "pink", "gray"]
+    default_color = random.choice(color_set)
+    color = input(f"Enter Spaces Card color [{default_color}]: ") or default_color
+    configuration["colorFrom"] = color
+    configuration["colorTo"] = color
+    configuration["sdk"] = "gradio"
+    huggingface_hub.metadata_save(readme_file, configuration)
+
+    configuration["hardware"] = (
```
freddyaboulton, line 60: I selected `cpu-upgrade` but my space did not get built with that hardware

image

https://huggingface.co/spaces/freddyaboulton/calculator-deploy-test

freddyaboulton: Also, I added a requirements.txt file to this same space via cli but the space linked above does not have a requirements.txt file. 

image

Also, I think it'd be better UX if each dependency got a separate line in the cli as opposed to having to type them all in at once. Similar to how the secrets work.

gradio/cli.py
```diff
@@ -0,0 +1,14 @@
...
+
+import gradio.deploy_space
+import gradio.reload
+
+
+def cli():
+    args = sys.argv[1:]
+    if len(args) == 0:
+        raise ValueError("No file specified.")
+    if args[0] == "--deploy":
```
freddyaboulton, line 11: Can we call it `deploy` as opposed to `--deploy`? I think `--` is usually used for a flag modifier on a command

gradio/deploy_space.py
```diff
@@ -0,0 +1,183 @@
...
+    emoji_set = "🤯🤖🧠🐶👑💥🐮🎁🐙🦋"
+    default_emoji = random.choice(emoji_set)
+    configuration["emoji"] = (
+        input(f"Enter Spaces Card emoji [{default_emoji}]: ") or default_emoji
+    )
+
+    color_set = ["red", "yellow", "green", "blue", "indigo", "purple", "pink", "gray"]
+    default_color = random.choice(color_set)
+    color = input(f"Enter Spaces Card color [{default_color}]: ") or default_color
+    configuration["colorFrom"] = color
```
freddyaboulton, line 55: Would be nice to add sdk_version and python_version as well.

gradio/deploy_space.py
```diff
@@ -0,0 +1,183 @@
...
+def format_title(title: str):
+    title = title.replace(" ", "_")
+    title = re.sub(r"[^a-zA-Z0-9\-._]", "", title)
+    title = re.sub("-+", "-", title)
+    while title.startswith("."):
+        title = title[1:]
+    return title
+
+
+def notebook_to_script(notebook_path, script_path):
```
freddyaboulton, line 97: This conversion script messed up for me: https://huggingface.co/spaces/freddyaboulton/calculator-notebook-deploy/tree/main

I wonder if we can use an existing tool for this (optional dependency) or leave notebook deploy out of scope for now.

freddyaboulton (commented):
Great PR @aliabid94 ! Some things were not working as expected for me so I left comments. But I think this will be a very useful feature.

freddyaboulton (commented):
Great PR @aliabid94 ! Some things were not working as expected for me so I left comments. But I think this will be a very useful feature.

gradio/deploy_space.py
aliabid94: I think you're right. Would be nice to support a "push from colab" but I think it's out of scope for now.

gradio/cli.py
aliabid94: hmm I did want to clearly distinguish between using the existing gradio command for launch gradio in "auto relaunch mode", e.g. `gradio run.py`. I suppose it's unlikely that there's a python file named "deploy". Or we could choose another command line name like "gradio-deploy" though that feels ugly.

gradio/deploy_space.py
aliabid94: I guess this interaction wasn't clear - I had intended the user to go edit the requirements file and then come back and press Enter. You're right, it would be more consistent to have the user enter them through the CLI

gradio/cli.py
freddyaboulton: good point about conflicting with filenames. Ok let's keep as is with `--deploy`!

gradio/cli.py
```diff
@@ -0,0 +1,14 @@
...
+        gradio.deploy_space.deploy()
+    else:
+        gradio.reload.run_in_reload_mode(*args)
```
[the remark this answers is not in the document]
aliabid94: I cannot reproduce this, can you expand on this? By the way if you are in editable mode (`pip install -e .`) you may have to rerun it since I don't think CLI bindings automatically update.

gradio/cli.py
abidlabs: Agree with @freddyaboulton. Using `--deploy` to indicate an action instead of a flag is very unconventional. I would prefer to use `gradio deploy` (chance of a collision is so low)

gradio/upload.py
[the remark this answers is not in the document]
aliabid94: @Wauplin https://github.com/huggingface/huggingface_hub/pull/1476/

gradio/deploy_space.py
aliabid94: Seems like the hardware issue is a huggingface_hub library problem. Will remove from here for now.

gradio/cli.py
aliabid94: I don't love `gradio deploy` but ok

gradio/deploy_space.py
aliabid94: Done.

gradio/deploy_space.py
```diff
@@ -0,0 +1,144 @@
...
+    default_color = random.choice(color_set)
+    color = input(f"Enter Spaces Card color [{default_color}]: ") or default_color
+    configuration["colorFrom"] = color
+    configuration["colorTo"] = color
+    configuration["sdk"] = "gradio"
+    configuration[
+        "python_version"
+    ] = f"{sys.version_info.major}.{sys.version_info.minor}.{sys.version_info.micro}"
+    configuration["sdk_version"] = gr.__version__
+    huggingface_hub.metadata_save(readme_file, configuration)
```
abidlabs, line 59: Throws an error on Windows: `UnicodeEncodeError: 'charmap' codec can't encode character '\U0001f916' in position 46: character maps to <undefined>`

Thi

gradio/deploy_space.py
abidlabs: So just to confirm, we don't support automagically pushing from a notebook? It'd be a really slick workflow but we can leave it as a future PR.

abidlabs: And if we don't support conversion of notebooks, I don't think having a `gr.deploy()` makes much sense. You'd always do `gradio deploy` from the terminal

abidlabs:
Very cool @aliabid94! My main concern is around the interactive way in which parameters are supplied to `gradio deploy`. It would be much better if these parameters were simply command line flags (`--name`, `--app_file`, etc.) for two reasons:

(1) You can re-run the command in your terminal much easier and more consistently
(2) You can skip having to deal with optional parameters like emoji, card color, which most people wouldn't care about setting

You can still keep the sensible defaults like using the directory name as the name of the Space, etc.

gradio/deploy_space.py
aliabid94: The huggingface_hub.metadata_save line? Can you file an issue on https://github.com/huggingface/huggingface_hub/

aliabid94:
Ok added argparse flags for `gradio deploy` as well, e.g. `gradio deploy --title=Test --app-file=run.py`

gradio/deploy_space.py
abidlabs: PR here: https://github.com/huggingface/huggingface_hub/pull/1484/files

cc @Wauplin

Wauplin: Thanks :pray: Merging it :)

aliabid94:
Added the ability to automatically attach a Github Action to track changes and upload to spaces, as @pngwn suggested. To test this out:
1. Check out this gradio branch 
2. you will have to edit `deploy_space_action.yaml`  to change `python -m pip install gradio` to `python -m pip install https://gradio-builds.s3.amazonaws.com/c0138fa3cf227b01cdc468c7817298d5f7849c9e/gradio-3.32.0-py3-none-any.whl` for testing.
3. run `pip install -e .`
4. Create a git repo with a gradio runfile
5. Run `gradio deploy` and follow instructions.

guides/01_getting-started/03_sharing-your-app.md
```diff
@@ -32,7 +32,10 @@ Share links expire after 72 hours.
 
 If you'd like to have a permanent link to your Gradio demo on the internet, use Hugging Face Spaces. [Hugging Face Spaces](http://huggingface.co/spaces/) provides the infrastructure to permanently host your machine learning model for free! 
 
-You can either drag and drop a folder containing your Gradio model and all related files, or you can point Spaces to your Git repository and Spaces will pull the Gradio app from there. See [this guide how to host on Hugging Face Spaces](https://huggingface.co/blog/gradio-spaces) for more information. 
+After you have [created a free Hugging Face account](https://huggingface.co/join), you have three methods to deploy your Gradio app to Hugging Face Spaces:
+1. From terminal: run `gradio deploy` in your app directory. The CLI will gather some basic metadata and then launch your app. Whenever you update your app, just run this command again.
```
abidlabs, line 36: Should we add a note here on the GitHub Action?

CHANGELOG.md
```diff
@@ -52,6 +52,7 @@ No changes to highlight.
 - Added `format` argument to `Audio` component by [@freddyaboulton](https://github.com/freddyaboulton) in [PR 4178](https://github.com/gradio-app/gradio/pull/4178)
 - Add JS client code snippets to use via api page by [@aliabd](https://github.com/aliabd) in [PR 3927](https://github.com/gradio-app/gradio/pull/3927).
 - Update to the JS client by [@pngwn](https://github.com/pngwn) in [PR 4202](https://github.com/gradio-app/gradio/pull/4202)
+- Introduced `gradio deploy` to launch a Gradio app to Spaces directly from your terminal. By [@aliabid94](https://github.com/aliabid94) in [PR 4033](https://github.com/gradio-app/gradio/pull/4033).
```
abidlabs, line 55: Let's move this to the top of the changelog and add some more details on usage

gradio/deploy_space.py
```diff
@@ -0,0 +1,167 @@
...
+        )
+        os.makedirs(os.path.dirname(github_action_file), exist_ok=True)
+        with open(github_action_template) as f:
+            github_action_content = f.read()
+        github_action_content = github_action_content.replace("$branch", track_branch)
+        with open(github_action_file, "w") as f:
+            f.write(github_action_content)
+
+        print(
+            "Github action created. To your Github repository, add your Hugging Face write token as a Secret named 'hf_token'. Github Secrets can be set in the Github Settings page."
```
abidlabs, line 96:
```suggestion
            "Github Action created. Add your Hugging Face write token (from https://huggingface.co/settings/tokens) as an Actions Secret named 'hf_token' to your GitHub repository. This can be set in your repository's settings page."
```

abidlabs (approved):
LGTM @aliabid94! Let's document this a bit better (see comments above) and release!

pngwn:
I already created an actual GH action to deploy to spaces here https://github.com/gradio-app/github. We could publish this in the marketplace too. 

I think this approach solves that specific problem better as you do not need to install an additional dependency so it is faster, you don't needs python, and it is only one step rather than 2 or 3. It is also easier to pin to a version in the action.

aliabid94:
> I already created an actual GH action to deploy to spaces here https://github.com/gradio-app/github. We could publish this in the marketplace too.

Sounds good, how about I merge this in for now and you can make a PR to replace the logic with this repo instead?

Issue #3939, opened 2023-04-22
Best way to protect private files from being accessed by `http://localhost:port/file=/path/to/file.txt` ?


binary-husky:
If there are sensitive scripts/data under gradio project that I want to forbid access from browser (and any other possible method), 
e.g. a txt file under the project folder, what is the best way prevent such access?

![image](https://user-images.githubusercontent.com/96192199/233781427-5107b6e3-1cc3-4993-b109-4586c79bf262.png)

I have tried to search issues but have not find anything related (maybe I not using the right words to search them)

abidlabs:
Hi @binary-husky currently it is not possible blacklist files, but this is certainly something we can consider adding.

Issue #3597, opened 2023-03-23
Ability to customize Gradio's output folder for `gr.File` uploads


space-nuko:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
The ability to specify where files served by `gr.File` are uploaded would be very useful, right now it's always set to a temporary folder named after the hash of the file and it is not configurable

https://github.com/gradio-app/gradio/blob/7856382de623bfd157b8961fcf626a9b0c35ad0f/gradio/components.py#L240-L241

That would allow cleaning up the temporary files on each startup, currently there's no centralized place in the temporary folder where Gradio stashes its temporary files, they just accumulate over time

It should allow specifying a path outside the system temporary directory also. Of course the path specified should be added to `temp_file_sets` so it's treated as an allowed path

And filename collisions would need to be dealt with somehow. Maybe a way for the user to specify an output subdirectory/filepath to the value sent to a `gr.File` output. Or automatically renaming duplicate files. But the current solution of putting each file in a separate hashed folder isn't very useful, they should all be placed in one folder ideally.

One more thing I noticed, this hashing scheme also leaves behind a 0-sized temporary file in the created folder, I think it's unintentional - https://github.com/AUTOMATIC1111/stable-diffusion-webui/pull/8752#issuecomment-1481020332

**Describe the solution you'd like**  
Ability to set `Interface.upload_path` to the root folder where `IOComponent.make_temp_copy_if_needed` should dump files

**Additional context**
Specific usecase is better articulated by https://github.com/AUTOMATIC1111/stable-diffusion-webui/pull/8752#issuecomment-1481731062

abidlabs:
> And filename collisions would need to be dealt with somehow. Maybe a way for the user to specify an output subdirectory/filepath to the value sent to a gr.File output. Or automatically renaming duplicate files. But the current solution of putting each file in a separate hashed folder isn't very useful, they should all be placed in one folder ideally.

The reason we do it like this is:
* The hash ensures that the file paths are deterministic (if the same file is processed multiple times, it gets mapped to the same path avoiding the creation of unnecessary files)
* The fact that the hash is part of the directory as opposed to the file name allows us to preserve file names of uploaded files

I don't think we'd change the above behavior, but we could certainly allow users to set a parent directory where all temporary files would be stored so that;

> That would allow cleaning up the temporary files on each startup, currently there's no centralized place in the temporary folder where Gradio stashes its temporary files, they just accumulate over time

[...]

Issue #3592, opened 2023-03-23
Gradio creates a lot of empty temporary folders on each run


space-nuko:
### Describe the bug

Described in https://github.com/AUTOMATIC1111/stable-diffusion-webui/pull/8752#issuecomment-1481020332, every time a Gradio app is launched it creates multiple temporary folders, and a lot of them remain empty. This is somewhat important because a lot of webui users want to inspect intermediate outputs and it's hard to sift through dozens of temporary folders.

Even if there needs to be an autogenerated temp folder each startup, would be better to cut down on the number of temporary folders introduced

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Just run a Gradio app with a large number of components and watch how many temporary folders are added

### Logs

```shell
N/A
```

### System Info

```shell
3.23.0, Windows, Chrome
```

### Severity

annoying

abidlabs:
Interesting. Empty folders should not be created -- I'll take a look at this

freddyaboulton:
Hi @space-nuko !

Is this still an issue in the latest gradio release? All the files created by gradio should now be located in `tempfile.gettempdir() / "gradio"` by default and any subdirectories are created when postprocessing of a component is run, e.g. Gallery.postprocess, Video.postprocess. I can't see the empty directories created on launch!

If there is a repro we can run outside the sd web ui, that would be much appreciated!

PR #4047, opened 2023-05-03
Add ability to blocklist filepaths, ability to specify where gradio temp files are created

abidlabs:
Security around filepaths that Gradio exposes has been something that's come up frequently. This PR adds the ability to blocklist filepaths (and also improves the allowlist mechanism). 

Closes: #3939

Edit: I just merged in #4053, which further adds the ability to specify a location where all temporary files created by Gradio are saved using an environmental variable. I considered also making this a parameter in `launch()` but the problem is that Gradio components are the ones that save temporary files and they are not aware of the `Blocks` instance. Furthermore, the Gradio temporary file location should actually *not* be a parameter in code because it could lead to code breaking if it is uploaded from e.g. a Windows local system to a Linux system like Spaces

Closes: #3597

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ## New Features:
 
 - Add support for `visual-question-answering`, `document-question-answering`, and `image-to-text` using `gr.Interface.load("models/...")` and `gr.Interface.from_pipeline` by [@osanseviero](https://github.com/osanseviero) in [PR 3887](https://github.com/gradio-app/gradio/pull/3887)
+- Adds the ability to blocklist filepaths (and also improves the allowlist mechanism by [@abidlabs](https://github.com/abidlabs) in [PR 4047](https://github.com/gradio-app/gradio/pull/4047).
 
 ## Bug Fixes:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,8 @@
 ## New Features:
 
 - Add support for `visual-question-answering`, `document-question-answering`, and `image-to-text` using `gr.Interface.load("models/...")` and `gr.Interface.from_pipeline` by [@osanseviero](https://github.com/osanseviero) in [PR 3887](https://github.com/gradio-app/gradio/pull/3887)
-- Adds the ability to blocklist filepaths (and also improves the allowlist mechanism by [@abidlabs](https://github.com/abidlabs) in [PR 4047](https://github.com/gradio-app/gradio/pull/4047).
+- Adds the ability to blocklist filepaths (and also improves the allowlist mechanism) by [@abidlabs](https://github.com/abidlabs) in [PR 4047](https://github.com/gradio-app/gradio/pull/4047).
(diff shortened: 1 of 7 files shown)


gradio/utils.py
```diff
@@ -962,6 +962,16 @@ def abspath(path: str | Path) -> Path:
         return path.resolve()
 
 
+def is_in(path_1: str | Path, path_2: str | Path):
```
freddyaboulton, line 965: I wonder if this should return true if the paths are equal - can help clean up some of the code

i.e

```
utils.is_in(abs_path, allowed_path) or abs_path == utils.abspath(allowed_path)
``

File: test/test_routes.py [hunk 2]
```diff
@@ -207,14 +228,74 @@ def test_get_file_allowed_by_file_directories(self):
...
-            file_directories=[os.path.dirname(allowed_file.name)],
+            allowed_paths=[os.path.dirname(allowed_file.name)],
         )
         client = TestClient(app)
 
         file_response = client.get(f"/file={allowed_file.name}")
         assert file_response.status_code == 200
         assert len(file_response.text) == len(media_data.BASE64_IMAGE)
 
+        app, _, _ = gr.Interface(lambda s: s.name, gr.File(), gr.File()).launch(
```
freddyaboulton, line 239: nit: I would close the app after you're done with and launch the next one

freddyaboulton (approved):
Tested and works well! Thanks for adding the tests @abidlabs

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 ## New Features:
 
 - Add support for `visual-question-answering`, `document-question-answering`, and `image-to-text` using `gr.Interface.load("models/...")` and `gr.Interface.from_pipeline` by [@osanseviero](https://github.com/osanseviero) in [PR 3887](https://github.com/gradio-app/gradio/pull/3887)
+- Adds the ability to blocklist filepaths (and also improves the allowlist mechanism by [@abidlabs](https://github.com/abidlabs) in [PR 4047](https://github.com/gradio-app/gradio/pull/4047).
```
abidlabs, line 10: I should add a changelog entry for the abiliy to specify where temp files are created

abidlabs:
Thanks for the review @freddyaboulton! Will address your suggestions and merge

abidlabs:
Oh I think we'll need to release a new version of the gradio client library (why are the tests passing? That's weird)

freddyaboulton:
> Oh I think we'll need to release a new version of the gradio client library (why are the tests passing? That's weird)

How come?

abidlabs:
Oh awkward.. I have changes that I haven't pushed to this branch

The reason is that there are a bunch of other places where we create temporary files (search for `tempfile.NamedTemporaryFile` in the codebase) and I refactored all of them to use the temporary directory. For that, I created a helper function in the client library.

But for now, let me not push my changes and just remove #3592 from the list of closed issues. Will just move that to a new PR

Issue #4034, opened 2023-05-02
Safari buttons were not fixed


cowanAI:
### Describe the bug

output text was fixed but not the button that look "off"

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Same issue discussed as before

### Screenshot

<img width="1323" alt="image" src="https://user-images.githubusercontent.com/128092052/235597474-09d2f52c-3e40-460c-9cab-dedf7813836f.png">

### Logs

```shell
no logs
```

### System Info

```shell
3.28.1, safari
```

### Severity

annoying

abidlabs:
Hmm @aliabid94 can you investigate? @cowanAI can you provide a full code example we can use to reproduce?

cowanAI:
for sure: 

"ui = gr.Interface(
    fn=transcribe, 
    theme='gradio/soft',
    inputs=[
        gr.Audio(source="microphone", type="filepath", label="Upload audio"),
        gr.Textbox(label="Or chat here:"),
        
    ],
    outputs=gr.Textbox(interactive=True,label="Alicja AI:").style(show_copy_button=True),
    #allow_flagging=('never'),
    examples=[    ["Medical_Transcriber.mp3", "Medical transcriber: Alicja, please summarize the patient visit and auto-fill the “X” form."],
    ["Medical_Writing.mp3", "Medical writing: Hey Alicja, can you help me organize my thoughts for this research paper?"],
    ["Clinical_Trials.mp3", "Clinical trials: Alicja, what is the recommended sample size for this clinical trial and can you help me analyze the results?"],
    ["Language_Translation.mp3", "Language translation: Alicja, can you help me translate what this patient is saying into English?"],
    ["Medical_Research.mp3", "Medical research: Hey Alicja, can you help me find some relevant literature for this research question?"],
    ["Patient_Education.mp3", "Patient education: Alicja, can you explain this medical condition to my patient in simple terms?"],
],
    css="footer {visibility: hidden}",
    allow_flagging=('never'),
)



ui.queue(concurrency_count=2)
app = gr.mount_gradio_app(app, ui, path=CUSTOM_PATH)"

I think maybe it's because the bug was fixed for Safari mobile but not Safari Desktop maybe?

cowanAI:
ui = gr.Interface(
    fn=transcribe, 
    theme='gradio/soft',
    inputs=[
        gr.Audio(source="microphone", type="filepath", label="Upload audio"),
[...]

PR #4070, opened 2023-05-04
Soft theme fix

aliabid94:
Quick fix to dark title text color

Fixes: #4034


abidlabs:
LGTM @aliabid94. Confirmed the fix with both `"gradio/soft"` and `themes.Soft()`

image
image

Issue #4021, opened 2023-05-01
Support audio uploads on iOS? 


drscotthawley:
- [ x] I have searched to see if a similar issue already exists.

Following up on this Discussion and request to add Issue by @abidlabs : 

https://github.com/gradio-app/gradio/discussions/3998

iOS is a little funny and while the way Gradio configures Audio input elements should be ok, somehow it's not.

Pasting from the above-mentioned Discussion:

> Hi, I have a new audio-in, audio-out Gradio demo I'm working on (sorry, can't share any details yet).
> 
> But I notice that if I'm on my iOS devices, when I press the audio upload box ("Drop Audio here -or- click to Upload") and pick "Choose File", and go through the list of files to upload into the audio box, all the (.wav, .mp3, .aiff) audio files are "greyed out", i.e. I'm not allowed to select any to upload. Doesn't matter if I'm selecting files in iCloud Drive, or on my Device, or in Dropbox: All audio files are non-selectable.
> 
> I'm not sure if this is an iOS limitation or a Gradio issue....? Can anyone help?
> 
> This happens when using Safari or Chrome on my iPhone or iPad, pretty recent iOS versions. Brave browser on latest iPad version behaves similarly.
> 
> (No problems running from desktop!)

---

> Through experimentation*, I find that this seems to be a bug(/feature) in iOS whereby HTML input field with accept="audio/*" does not actually allow audio to be selected on iOS.
> 
> If Gradio wanted support audio uploads on iOS, one workaround on your end end could be to either change the accept string generated for the input field , either to accept="*/*" or to list each allowed file extension e.g accept="audio/mp3, audio/wav,..." etc. @abidlabs @yvrjsharma
> 
> (You shouldn't need to do this but iOS is a part of the market, and this change would still be compatible with other devices.)
> 
> (*Source. Trying it manually, on this page: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file )

abidlabs:
Thanks for creating the issue @drscotthawley will look into it!

abidlabs:
Btw would you like to open a PR since it looks like you’ve already started with the solution? Otherwise we can do it on our end

drscotthawley:
@abidlabs Re. the PR, I'd be happy to try but I'm not sure where in your code you're supplying that (MIME?) type information.  

[...]

PR #4071, opened 2023-05-04
Upload audio ios

aliabid94:
iOS is not able to filter for audio files correctly. Seems like a longstanding bug, and prevents users form uploading any audio files. So if on iOS, allows user to upload any file for the audio component.

Fixes: #4021


abidlabs:
Cool @aliabid94! Based on the discussion in the original issue, I think a simpler solution might be just to list the mimetypes out explicitly: list each allowed audio type explicitly, `accept="audio/mp3, audio/wav,..."`, and then you can include "audio/*" at the end to be safe. (Haven't tested this, but cc @drscotthawley)

drscotthawley:
This works on my iPhone. 👍

aliabid94:
> list each allowed audio type explicitly, accept="audio/mp3, audio/wav,...", and then you can include "audio/*" 

This works, changed to this

abidlabs:
Nice, it even allows you to upload a video from your phone and it automatically extracts the audio. Very cool LGTM @aliabid94!

abidlabs (approved):

Issue #3738, opened 2023-04-03
`.change()` event dependencies not triggered if target element is contained inside closed `gr.Accordion`


space-nuko:
### Describe the bug

If an element that is the target of a `.change()` event has the event triggered while inside a `gr.Accordion` that is closed, any dependent events fired by that element will not be reflected when the accordion is opened

Common use case is an "enabled" checkbox that toggles the visibility of some elements inside the accordion, and an external event that controls the value of the enabled checkbox

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

1. Launch demo, open accordion
2. Click "Set Enabled" to toggle, then close accordion and click "Set Enabled"
3. Open accordion again, note element visibility
4. Click "Set Enabled" several times to see updates

```python
#!/usr/bin/env python

import gradio as gr

with gr.Blocks() as demo:
    is_enabled = False
    with gr.Accordion("Accordion", open=False):
        with gr.Row():
            checkbox2 = gr.Checkbox(value=is_enabled)
            with gr.Column():
                checkbox = gr.Checkbox(value=False, visible=is_enabled)
                slider = gr.Slider(value=0, minimum=0, maximum=100, visible=is_enabled)
                dropdown = gr.Dropdown(["a", "b", "c"], value="a", visible=is_enabled)

    checkbox.change(lambda i: print(f"checkbox changed {i}"), inputs=[checkbox], outputs=[])
    slider.change(lambda i: print(f"slider changed {i}"), inputs=[slider])
    dropdown.change(lambda i: print(f"dropdown changed {i}"), inputs=[dropdown], outputs=[])

    changeall = gr.Button(value="Change All")
    changeall.click(lambda: [True, 42, "c"], inputs=[], outputs=[checkbox, slider, dropdown])

    checkbox2.change(lambda x: [gr.update(visible=x)] * 3, inputs=[checkbox2], outputs=[checkbox, slider, dropdown])
    changeall2 = gr.Button(value="Set Enabled", variant="primary")
    changeall2.click(lambda x: not x, inputs=[checkbox2], outputs=[checkbox2])

if __name__ == "__main__":
    demo.queue().launch()
```

### Logs

```shell
N/A
```

### System Info

```shell
3.23.0, Windows, Chrome
```

### Severity

annoying

muerrilla:
[...]

Issue #3657, opened 2023-03-28
Custom JavaScript breaks when accordion is closed


pkuliyi2015:
### Describe the bug

When I use change() to change the component values:
 1) The slider will call its js twice.
 2) When close the accordion, all inside js will be broken.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Simply run this code:

```
import gradio as gr

with gr.Blocks() as demo:
	checkbox = gr.Checkbox(label='Will change the label inside', interactive=True)
	with gr.Accordion(label='Test Accordion'):
		checkbox_js = gr.Checkbox(label='Custom Js for Checkbox', interactive=True)
		slider_js = gr.Slider(label='Custom Js Slider', minimum=0, maximum=1, step=0.01, value = 0.66, interactive=True)
		checkbox_js.change(fn=lambda x:x, inputs=checkbox_js, outputs=checkbox_js, _js='(value)=> alert("Checkbox Js is working") === null ? value : value')
		slider_js.change(fn=lambda x:x, inputs=slider_js, outputs=slider_js, _js='(value)=>  (alert("Slider Js is working") === null && value == 0) ? 0.2 : 0.8')
	checkbox.change(fn=lambda x: (x, 0 if x else 1), inputs=checkbox, outputs=[checkbox_js, slider_js])
demo.launch()
```

When you check the outside checkbox, the checkbox js will run once but the slider js will run twice.
When you check the inside checkbox its js run normally, but the slider js will still run twice.
When you close the accordion, none of the js will work.

### Logs

```shell
No errors, but it breaks all custom javascripts inside the accordion, hence all calculation breaks. 

I suspected it is because you removed the DOM when accordion close and custom js becomes no longer available.

I can't find anyway to bypass the bug and continue, so have to downgrade to an older version.
```

### System Info

```shell
Gradio==3.23.0, Python = 3.10.9. Clean environment with no other packages installed.

Note: on gradio==3.16.0 the accordion won't break the js, but checkbox js won't work.
```

### Severity

blocking upgrade to latest gradio version

space-nuko:
It's because the lack of custom components means state has to be tracked through frontend kludges, the only way for it to work without keeping the elements on the page would be #1432 being implemented, so the state can be saved to the backend

It's surely possible to make the DOM removing behavior configurable but it would be a workaround for an implementation detail that's subject to change at any time

pkuliyi2015:
[...]

Issue #3461, opened 2023-03-14
Block load event not triggerred if put components in gr.Accordion(label='Accordion', open=False) as input to load


shawnbow:
### Describe the bug

I made a demo to tell the issue:
https://huggingface.co/spaces/shawnbow/report_gradio_issue

```
import gradio as gr
def _on_login(state):
    print('Loaded and Display')
    return 'Loaded and Display', 'Loaded and Display', state

with gr.Blocks() as demo:
    state = gr.State({})

    with gr.Accordion(label='Accordion', open=False):

        markdown = gr.Markdown(label='Markdown', value='')
        text_box = gr.Textbox(label="Text", interactive=True, placeholder="Load Failed.....")

    demo.load(
        fn=_on_login,
        inputs=[state],
        outputs=[text_box, markdown, state],
    )
demo.queue()
demo.launch()
```
The _on_login not triggered when loading, untill you click Accordion to open.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

https://huggingface.co/spaces/shawnbow/report_gradio_issue

### Logs

```shell
No logs, because the load event not triggerred, all you can see is the log not printed.
```

### System Info

```shell
3.20.1, python 3.9
```

### Severity

serious, but I can work around it

Issue #3198, opened 2023-02-15
Weird interaction between Accordion and Tab


wfng92:
### Describe the bug

Given a `gr.Accordion` component with multiple `gr.Tab` components, the content of the contained `gr.Tab` components is not showing when you close and open back the `gr.Accordion` component (only applicable after clicking on any Tab component other than the first one). As a result, users need to manually click the `gr.Tab` component again to see the underlying content. This affects user experience.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Given the following code:

```python
import gradio as gr

with gr.Blocks() as demo:
    with gr.Accordion('Accordion', open=False):
        with gr.Tab('Tab1'):
            text1 = gr.Textbox('text1')
        with gr.Tab('Tab2'):
            text2 = gr.Textbox('text2')
        with gr.Tab('Tab3'):
            text3 = gr.Textbox('text3')

demo.launch()
```

Reproduction steps:
1. Click on the second or third Tab
2. Close the accordion
3. Open the accordion (notice that there are no content showing)

Desired interaction:
1. Click on the third tab
2. Close the accordion
3. Open the accordion (UI will show the content of the third tab)

### Screenshot

![image](https://user-images.githubusercontent.com/43742196/218932661-8d435990-4710-427a-b1db-9f842e9732c3.png)

No content is showing on opening of a closed `gr.Accordion`. By right, should show the content of the last selected `gr.Tab`.

### Logs

```shell
N/A
```

### System Info

```shell
Gradio version: '3.18.0'
Browser: Firefox and Chrome
```

### Severity

annoying

abidlabs:
Can confirm the weird behavior, we'll look into it @wfng92!

space-nuko:
Still an issue as of 3.23.0, some users of webui have experienced it in for example https://github.com/Mikubill/sd-webui-controlnet

anapnoe:
also if I have inside tab a list of some items and I scroll to the middle if re-trigger the tab it will start from top not from the last position ideally  `scrollTop` property should also be restored

PR #4073, opened 2023-05-04
Keep hidden Accordion content in DOM

aliabid94:
Fixes: https://github.com/gradio-app/gradio/issues/3738
Fixes: #3657 
Fixes: #3461
Fixes: #3198


abidlabs:
I think this would close #3657 and #3461 as well? Maybe even #3198? 

cc @pngwn

abidlabs:
Yes, confirmed that all three of the other issues have been fixed as well, added to parent-level comment. I don't see any issue but let's get @pngwn's thoughts in case there are any downsides to this

pngwn (approved):
Looks good!

abidlabs:
Thanks for the review @pngwn! Will merge in once CI passes @aliabid94

Issue #4079, opened 2023-05-05
AttributeError: 'Base' object has no attribute 'primary_xpand' on 3.28.3


reilnuud:
### Describe the bug

Nearly all xhr requests in theme builder result in a 500 error. Looking in the server log, you can see the error `AttributeError: 'Base' object has no attribute 'primary_xpand'` on nearly every request.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

1. Run pip install gradio.
2. Create `builder.py` with the following code per the theming tutorial:
```python
import gradio as gr

gr.themes.builder()
```
3. Run `builder.py`.
4. Attempt to edit theme or change base theme. Nothing happens.

### Logs

```shell
Traceback (most recent call last):
  File "/Users/neilruud/.pyenv/versions/3.10.6/lib/python3.10/site-packages/gradio/routes.py", line 412, in run_predict
    output = await app.get_blocks().process_api(
  File "/Users/neilruud/.pyenv/versions/3.10.6/lib/python3.10/site-packages/gradio/blocks.py", line 1299, in process_api
    result = await self.call_function(
  File "/Users/neilruud/.pyenv/versions/3.10.6/lib/python3.10/site-packages/gradio/blocks.py", line 1021, in call_function
    prediction = await anyio.to_thread.run_sync(
  File "/Users/neilruud/.pyenv/versions/3.10.6/lib/python3.10/site-packages/anyio/to_thread.py", line 31, in run_sync
    return await get_asynclib().run_sync_in_worker_thread(
  File "/Users/neilruud/.pyenv/versions/3.10.6/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 937, in run_sync_in_worker_thread
    return await future
  File "/Users/neilruud/.pyenv/versions/3.10.6/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 867, in run
    result = context.run(func, *args)
  File "/Users/neilruud/.pyenv/versions/3.10.6/lib/python3.10/site-packages/gradio/themes/builder.py", line 812, in render_variables
    generate_theme_code(
  File "/Users/neilruud/.pyenv/versions/3.10.6/lib/python3.10/site-packages/gradio/themes/builder.py", line 616, in generate_theme_code
    final_attr_values[final_theme_attr] = getattr(
AttributeError: 'Base' object has no attribute 'primary_xpand'
```
```


### System Info

```shell
Mac OS 13.0.1
Python 3.10.6
[...]

PR #4080, opened 2023-05-05
Theme builder fixes

abidlabs:
Two fixes for the theme builder:

1. Renames the file from `gradio.themes.builder.py` to `gradio.themes.builder_app.py` because there was an issue if you would re-reun `gradio.themes.builder()` twice (could easily happen in a notebook setting), it would think that you are trying to call the module instead of the function. The code has been refactored so that `gradio.themes.builder()` still works.

2. Manually removed `expand` from the list of attributes (fixes: #4079), but I can't figure out what broke this in the first place. @aliabid94 do you know? 

I also noticed a different error: the code that has been generated for the theme looks wrong: 

* It always includes the `font_mono` parameter even if I didn't change it
* The single quotes are repeated twice around the font names
* The syntax highlighting is broken

image

@aliabid94 could you give this a look?


aliabid94:
Fixed. Seems like https://github.com/gradio-app/gradio/commit/12a97746ff4fcb5aebbd66a4738137999f538791 added the `expand` method to the colors for linting purposes, so we need to filter that out.

gradio/themes/builder_app.py
```diff
@@ -610,6 +610,7 @@ def generate_theme_code(
                 final_attr_values = {}
                 diff = False
                 for attr in dir(source_obj):
+                    print(attr)
```
abidlabs, line 613:
```suggestion
```

abidlabs:
LGTM @aliabid94! Will merge once CI passes

PR #4090, opened 2023-05-06
Allow textbox / number submits to trigger Interface submit

aliabid94:
Allows users to hit Enter on a textbox / number in interface to trigger a submit.

was annoying me.


abidlabs:
Nice! UX is much better, agreed. 

One problem though is that the submit trigger does not convert the "submit" button to the "stop" button when working with generators. For example, try:

```py
import gradio as gr
import time

def test(x):
    time.sleep(3)
    yield x

gr.Interface(test, "text", "text").queue().launch()
```

Put something in the textbox and press the enter key. The submit button _won't_ be changed to the stop button.

I think the relevant section of the code is here: https://github.com/gradio-app/gradio/blob/bebfb72b353a4280155cf7070441fc476ac10172/gradio/interface.py#L639

aliabid94:
> One problem though is that the submit trigger does not convert the "submit" button to the "stop"
Fixed.

abidlabs:
LGTM @aliabid94 thanks for making the fix!

abidlabs (approved):

PR #4095, opened 2023-05-07
Better range number input handling

space-nuko:
# Description

Tries to fix two issues

- Changing the slider's number input by typing into it didn't trigger the `released` event
- Possible to delete the entire number in the input, leaving an empty space. Now the value is clamped to the slider's min/max if it can't be parsed as a float

# Checklist:


space-nuko:
Also a thing I noticed, with this implementation you can input leading zeroes into the number field. Don't know if it's desirable

abidlabs:
LGTM thanks for the fix @space-nuko

abidlabs:
Sorry responded to wrong PR

abidlabs:
Thanks @space-nuko for opening this PR. My feedback:

> Changing the slider's number input by typing into it didn't trigger the released event

I don't think the `release()` event should be triggered in this case. "Release" signifies that there's some continuous action (such as dragging the slider) which _stops_. That's not what typing into the trigger means to me. The `.change()` event, which already fires whenever the numerical value is changed, is more appropriate here. 

If you would like to add a `.blur()` event which gets fired after someone finishes typing and then clicks outside of the number field, then I would be supportive of that.

> Possible to delete the entire number in the input, leaving an empty space. Now the value is clamped to the slider's min/max if it can't be parsed as a float

This is problematic because someone could be in the process of typing a number and then the value that they are typing in gets clamped. For example, try running:

```py
import gradio as gr

with gr.Blocks() as demo:
    s1 = gr.Slider(minimum=20, interactive=True)
    
demo.launch()
```

and typing in "33". It's basically impossible to do so. The current behavior, which only clamps when the number field is blurred, is better for usability. Forthose reasons, I'll go ahead and close this PR

PR #4106, opened 2023-05-08
Client symlink [WIP]

abidlabs:
Working on a way to improve the linking between the `gradio` and `gradio_client` libraries to avoid compatibility issues


abidlabs:
Sadly my idea does not seem to work

PR #4112, opened 2023-05-08
Fixes to view API page

abidlabs:
Makes two fixes to the view API page:

1. The logo next to "view API" was not showing correctly. There was a 403 error on the console log.
2. The "view API" page was showing "undefined" for some properties because we changed the format of the `/info` route

Both are now fixed


abidlabs:
Thanks for taking this over, @aliabd! Merging based on our discussion here: https://huggingface.slack.com/archives/C02SPHC1KD1/p1683603264539489?thread_ts=1683599572.486589&cid=C02SPHC1KD1

Issue #4113, opened 2023-05-08
then() won't run after a failure


freddyaboulton:
### Describe the bug

If you have a `then` event, it won't run if the triggering event failed.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

In this example, I would expect the text to display in the textbox but the function never runs.

```python
import gradio as gr

def fail():
    raise ValueError("Failed!")


with gr.Blocks() as demo:
    num = gr.Number()
    textbox = gr.Textbox()
    button = gr.Button()

    button.click(fail, inputs=None, outputs=num).then(
        lambda: "After failure",
        inputs=None,
        outputs=textbox
    )

demo.launch()
```

### Screenshot

<img width="1512" alt="image" src="https://user-images.githubusercontent.com/41651716/236941157-0901f686-5de3-4fcd-8101-881d3c197716.png">

### Logs

```shell
-
```

### System Info

```shell
3.28.3 / main
```

### Severity

serious, but I can work around it

PR #3821, opened 2023-04-11
Add support for async generators

freddyaboulton:
# Description

Closes: #3618 . @deep-diver would appreciate if you could test this to see if it fixes your issue. Click on the link on the gradio-pr-bot message, Go to Files in the top right, requirements.txt will have a url hosting a wheel built for this PR.

### Blocks case
```python
import asyncio
import gradio as gr
import time


async def async_iteration(count: int):
    for i in range(count):
        yield i
        await asyncio.sleep(1)


def iteration(count: int):
    for i in range(count):
        yield i
        time.sleep(1)


with gr.Blocks() as demo:
    with gr.Row():
        with gr.Column():
            num1 = gr.Number(value=4, precision=0)
            o1 = gr.Number()
            async_iterate = gr.Button(value="Async Iteration")
            async_iterate.click(async_iteration, num1, o1)
        with gr.Column():
            num2 = gr.Number(value=4, precision=0)
            o2 = gr.Number()
            iterate = gr.Button(value="Iterate")
            iterate.click(iteration, num2, o2)

demo.queue(concurrency_count=20).launch()
```

### Interface case
```python
app1 = gr.Interface(async_iteration, gr.Number(precision=0, value=4), gr.Number())
app2 = gr.Interface(iteration, gr.Number(precision=0, value=4), gr.Number())

with gr.Blocks() as demo:
    with gr.Row():
        app1.render()
    with gr.Row():
        app2.render()

demo.queue(concurrency_count=20).launch()
```

# Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have added a short summary of my change to the CHANGELOG.md
- [ ] My code follows the style guidelines of this project
- [ ] I have commented my code in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

# A note about the CHANGELOG

Hello 👋 and thank you for contributing to Gradio!

All pull requests must update the change log located in CHANGELOG.md, unless the pull request is labeled with the "no-changelog-update" label.

Please add a brief summary of the change to the Upcoming Release > Full Changelog section of the CHANGELOG.md file and include
a link to the PR (formatted in markdown) and a link to your github profile (if you like). For example, "* Added a cool new feature by `[@myusername](link-to-your-github-profile)` in `[PR 11111](https://github.com/gradio-app/gradio/pull/11111)`".

[...]

PR #4115, opened 2023-05-09
Fix then logic to work after failures

aliabid94:
Not sure if this was always broken or broke as a result of changes in client behaviour. Now should handle `.then` if first event fails. 

Fixes: #4113


abidlabs:
Thanks @aliabid94! Checked the 2x2 combination of [`'.then()'`, `'.success()'`] with [successful event, failed event], and all four cases work. Awesome!

abidlabs:
Let's merge it in after CI passes so that we can get #3821 in fully

Issue #3726, opened 2023-04-01
Can't use argparse in reload mode


LZY-the-boys:
### Describe the bug

When I use argparse in reload mode, it cause errors, for example:
```
import gradio as gr
import argparse
with gr.Blocks() as demo:
    gr.Markdown("# Greetings from Gradio!")
    inp = gr.Textbox(placeholder="What is your name?")
    out = gr.Textbox()

    inp.change(fn=lambda x: f"Welcome, {x}!", 
               inputs=inp, 
               outputs=out)

if __name__ == "__main__":
    parser = argparse.ArgumentParser()
    parser.add_argument("--a", type=str, default="decapoda-research/llama-7b-hf")
    parser.add_argument("--b", type=str, default="./lora-Vicuna/checkpoint-3000")
    parser.add_argument("--c", type=int, default=1)
    parser.add_argument("--e", type=int, default=1)
    parser.add_argument("--d", type=int, default=1)
    args, unknown = parser.parse_known_args()
    demo.launch()
```

When I launch it using `gradio tmp.py --a 1` ， it throws 
```
ERROR:    Error loading ASGI app. Attribute "--a.app" not found in module "tmp".
```
It seems that it interpet `--a 1` to "--a.app" . I don't find any description in documents

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

above code 

### Logs

```shell
ERROR:    Error loading ASGI app. Attribute "--a.app" not found in module "tmp".
```

### System Info

```shell
linux
```

### Severity

annoying

freddyaboulton:
Hi @LZY-the-boys ! Thanks for filing.

I don't think we'll be able to support this use case. What `gradio` command does is delegate to uvicorn's run-in-reload mode, which imports the `app` object from your file, so the part in the `if __name__ == '__main__'` clause will not be run. Also, uvicorn's command line does not support arbitrary extra arguments ([docs](https://www.uvicorn.org/#usage)).

What I suggest you do is load those variables via an environment variable or from a config file!

LZY-the-boys:
OK, thanks for help

PR #4119, opened 2023-05-09
Programmatically launch reload to allow factories and arguments to be passed to the app

micky2be:
# Description

Some application have their own optional arguments passed through CLI, others due to their size and use are not directly creating a Gradio app available publicly. 
By using Uvicorn API instead of the CLI, this changes will allow both above issues to be compatible with the Gradio reloading development tool.
_Note that instead of using `uvicorn.run` the logic has been decoupled between setting it up and running the server to allow proper testing_

Closes: #3726 

# Checklist:

# Further plausible update
- dedicated path of an env file to be loaded to allow additional configuration such as host, port, or SSL
- allowing the Gradio app to be self launch (easy to implement, need proper testing)


abidlabs:
Thanks @micky2be for opening this PR! We have a lot going on at the moment, but we'll take a look at this PR soon

gradio/reload.py
```diff
@@ -3,22 +3,28 @@
 Contains the functions that run when `gradio` is called from the command line. Specifically, allows
 
 $ gradio app.py, to run app.py in reload mode where any changes in the app.py file or Gradio library reloads the demo.
-$ gradio app.py my_demo, to use variable names other than "demo"
+$ gradio app.py my_demo.app, to use variable names other than "demo"
```
abidlabs, line 6: Can we make this change backwards compatible by keeping the demo name without the ".app" suffix. Instead, we can add the suffix internally in `_setup_config()`

guides/07_other-tutorials/developing-faster-with-reload-mode.md
```diff
@@ -50,7 +50,9 @@ WARNING:  The --reload flag should not be used in production on Windows.
 
 The important part here is the line that says `Watching...` What's happening here is that Gradio will be observing the directory where `app.py` file lives, and if the file changes, it will automatically rerun the file for you. So you can focus on writing your code, and your Gradio demo will refresh automatically 🥳
 
-⚠️ Now, there is one important thing to keep in mind when using the reload mode: Gradio specifically looks for a Gradio Blocks/Interface demo called `demo` in your code. If you have named your demo something else, you can pass that as the 2nd parameter in your code, like this: `gradio app.py my_demo`
+⚠️ Now, there is one important thing to keep in mind when using the reload mode: Gradio specifically looks for a Gradio Blocks/Interface demo called `demo` in your code. If you have named your demo something else, you can pass that as the 2nd parameter in your code, like this: `gradio app.py my_demo.app`
+
+⚠️ If your application accepts cli arguments using a function/factory you can now run `gradio app.py create_app --option1 --option2`, assuming `create_app` will return the `app` object of your Gradio app
```
abidlabs, line 55: Would be good to add a small self-contained example of passing in command line arguments and using it in your Gradio demo

abidlabs:
Thanks @micky2be for the nice PR! Tested and functionality looks good. The only real change that's needed is to preserve backwards compatibility -- see my first comment above. If you could also add a small example in the Guide, that'd be appreciate and we should be good to merge after that.

gradio/reload.py
micky2be: I don't see a nice way to do it. Uvicorn is not triggering any Exepction I could hatch on to try one or the other.
Since this is a cli primary used for development, I'm not sure such breaking change is too much of an issue.

guides/07_other-tutorials/developing-faster-with-reload-mode.md
micky2be: Would you rather an example written here, or a real life demo?

gradio/reload.py
abidlabs: I don't quite understand sorry. What do you mean "The problem with adding the .app is for factories. It won't work in these cases."

guides/07_other-tutorials/developing-faster-with-reload-mode.md
abidlabs: Let me see if I can add something real quick. Something like what you have but maybe a bit simpler

gradio/reload.py
abidlabs: Ok yeah it doesn't seem like Uvicorn throws an exception until the web app is actually running. So for now, I've added a warning if someone passes in a string that does not include "." for the app name, letting them know that the syntax has deprecated

abidlabs:
LGTM. @freddyaboulton would you be able to take a quick look as well?

freddyaboulton (approved):
Thanks @micky2be ! Tested that the queue works in reload mode as well.

freddyaboulton (approved):
Thanks @micky2be ! Tested that the queue works in reload mode as well.

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## New Features:
 
+- The reloader command (`gradio app.py`) can now accept argumnets to be passed to your application by [@micky2be](https://github.com/micky2be) in [PR 4119](https://github.com/gradio-app/gradio/pull/4119)
```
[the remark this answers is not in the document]
abidlabs: Fixed

guides/07_other-tutorials/developing-faster-with-reload-mode.md
```diff
@@ -48,13 +48,56 @@ Watching...
...
+args, unknown = parser.parse_known_args()
+
+with gr.Blocks() as demo:
+    gr.Markdown(f"# Greetings {args.name}!")
+    inp = gr.Textbox()
+    out = gr.Textbox()
+
+    inp.change(fn=lambda x: x, inputs=inp, outputs=out)
+
+app = demo.app
```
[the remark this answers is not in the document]
abidlabs: Indeed we do not

gradio/reload.py
```diff
@@ -36,21 +44,46 @@ def run_in_reload_mode():
...
+        port=port,
+        log_level="warning",
+        reload_dirs=watching_dirs,
+    )
+
+
+def main():
+    # default execution pattern to start the server and watch changes
+    config = _setup_config()
+    server = Server(config)
```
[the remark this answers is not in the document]
abidlabs: Ah yeah let me make the change and confirm everything still works

abidlabs:
Thanks for the PR @micky2be and for reviewing @freddyaboulton. Will merge in once CI is green

test/test_reload.py
```diff
@@ -56,7 +55,3 @@ def test_config_load_default(self, config):
     def test_config_load_factory(self, config):
         config.load()
         assert config.loaded is True
-
-    def test_reload_run_default(self, reloader):
```
micky2be: @abidlabs you shouldn't remove this test.
That was the only way to actively test if the passed config is actually correct and the server running

abidlabs: This was breaking other tests. I think b/c it wasn't closed. Let me see if I can bring it back, thanks

abidlabs: Brought back test but with threading (a bit more consistent with how we use `networking.Server` in the library)

test/test_reload.py
```diff
@@ -5,7 +5,8 @@
 
 import gradio
 import gradio as gr
-from gradio.reload import Server, _setup_config
+from gradio.networking import Server
```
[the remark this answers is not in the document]
abidlabs: When we switched to using networking.Server, this broke the original import. We could refactor but I don't think it's really necessary.

PR #4130, opened 2023-05-09
Fix create version docs action

aliabd:
Create Version Docs gh action has been creating docs with the wrong version pretty much every time. My guess is it takes a bit of time for the version that's on the pypi url (https://pypi.org/pypi/gradio/json) to update and install correctly. Made a few changes to the file including a 5 minute sleep that should fix things. Don't have a nice way to test so we'll see if it keeps happening..


abidlabs:
Makes sense, I think that's the issue.

abidlabs (approved):

abidlabs (approved):

Issue #3472, opened 2023-03-15
UI not displaying correctly


areeb1245:
### Describe the bug

The UI displays correctly on the local system. But when I upload it to the server, the UI CSS is not present. 

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

     NA

### Screenshot

<img width="1440" alt="Screenshot 2023-03-15 at 12 23 45 AM" src="https://user-images.githubusercontent.com/71621292/225127688-2ce7189c-70f5-44c4-bce7-ee96040967a6.png">

### Logs

```shell
No logs. Only a 404 error.
```

### System Info

```shell
This is hosted on railway.app.
```

### Severity

annoying

areeb1245:
Here's the URL:
https://fastapi-production-72d5.up.railway.app/answer-ui/

abidlabs:
I'm not super familiar with Railway but it seems like the problem is that Gradio needs access to `/theme.css` to load the CSS, and either this file is missing from the Railway app, or Railway is blocking this.

areeb1245:
I tried loading the thing on a different account and it seems to be working. But it's only this account for some reason. Do you know how I can get this sorted?

areeb1245:
> I'm not super familiar with Railway but it seems like the problem is that Gradio needs access to `/theme.css` to load the CSS, and either this file is missing from the Railway app, or Railway is blocking this file

Is there any way I can index the theme directly in the code?

abidlabs:
> I tried loading the thing on a different account and it seems to be working. But it's only this account for some reason. Do you know how I can get this sorted?

It works on a different Railway app? Then it doesn't sound like a Gradio issue, it sounds like an issue with Railway, no?

> Is there any way I can index the theme directly in the code?

I don't think so. cc @aliabid94 for your thoughts

areeb1245:
Yes, you're right. It's probably an issue with them. But is there any CDN that I can use to index the theme.css file?
@aliabid94

abidlabs:
No the theme.css file is creating specifically for your app and it might differ from app to app, so it wouldn't be available on a CDN. I'm going to close this issue as it doesn't seem related to Gradio, but rather Railway. If you could reach out to them and see how to serve a static css file, that should solve the issue you're seeing.

hinaloe:
It looks gradios's bug.

I ran into the same problem while trying the following case:

```python
app = gr.mount_gradio_app(app, io, path="/gradio")
```

[...]

PR #4133, opened 2023-05-09
Fix isssue with running with nginx

abidlabs:
Fixes: #3472 by adding a `root_path` parameter to `launch()` which should be set to the same as the subpath of the nginx reverse proxy. I looked pretty hard to see if there was a way that we didn't have to manually specify this parameter but it seems that FastAPI apps (or any web apps) have no way of knowing what the path of the proxy is so this needs to be manually specified: https://fastapi.tiangolo.com/advanced/behind-a-proxy/#setting-the-root_path-in-the-fastapi-app

Hard to write a test for this, but I replicate the nginx conf & test code from this comment: https://github.com/gradio-app/gradio/issues/3472#issuecomment-1537979058

And you can see it running successfully here: http://54.191.13.1/app/gradio/


gradio/routes.py
```diff
@@ -222,15 +223,17 @@ def login(form_data: OAuth2PasswordRequestForm = Depends()):
         def main(request: fastapi.Request, user: str = Depends(get_current_user)):
             mimetypes.add_type("application/javascript", ".js")
             blocks = app.get_blocks()
+            root_path = request.scope.get("root_path")
```
freddyaboulton, line 226: @abidlabs Thanks for the fix! If we're reading the root_path from the request, why do we need an argument in blocks?

abidlabs: Good question. The `root_path` from the `request` will include the subpath if you mount Gradio within a FastAPI app. However, it will *not* include any subpath that comes from the nginx reverse proxy since the FastAPI app does not know about that. So you need to put both

gradio/routes.py
```diff
@@ -778,7 +784,6 @@ def read_main():
         # Then run `uvicorn run:app` from the terminal and navigate to http://localhost:8000/gradio.
     """
     blocks.dev_mode = False
-    blocks.root = path[:-1] if path.endswith("/") else path
```
freddyaboulton, line 781: This is essentially deprecating the `path` argument of `mount_gradio_app`. I'm on board since it is superfluous. Let's update the docstring and maybe print a warning.

abidlabs: Well we still need the `path` argument to specify what route to mount the `gradio` app within the larger FastAPI app (see line 793), so we shouldn't deprecate. But we no longer need to use the `path` parameter to set the `root` here.

gradio/routes.py
aliabid94, line 226:
```suggestion
            root_path = request.scope.get("root_path", "")
```
I don't believe root_path is guaranteed anywhere to be present, and we need to fall back on an empty string

gradio/routes.py
```diff
@@ -261,8 +264,11 @@ def api_info(serialize: bool = True):
 
         @app.get("/config/", dependencies=[Depends(login_check)])
         @app.get("/config", dependencies=[Depends(login_check)])
-        def get_config():
-            return app.get_blocks().config
+        def get_config(request: fastapi.Request):
+            root_path = request.scope.get("root_path")
```
aliabid94, line 268:
```suggestion
            root_path = request.scope.get("root_path", "")
```

gradio/routes.py
freddyaboulton: Makes sense!

gradio/routes.py
aliabid94: where is the Blocks.root_path used? I only see the root_path from the request being used.

abidlabs: In `configure_app`. The FastAPI's app's `root_path` is set to  Blocks's `root_path`

abidlabs: When you do `request.scope.get("root_path")`, it combines both the FastAPI app's explicit root path (the same one coming from the nginx reverse proxy), along with any subpath that your app may have because it's mounted within a larger FastAPI

abidlabs: Take a look at this user's 2 examples to see why you need both: https://github.com/gradio-app/gradio/issues/3472#issuecomment-1541446133

aliabid94: I see it being set to `self.root_path` but never being used, is that something the FastAPI app automatically recognizes?

abidlabs: Yup the `root_path` is used internally by `request.scope.get("root_path")`. This is kinda mentioned here: https://fastapi.tiangolo.com/advanced/behind-a-proxy/#setting-the-root_path-in-the-fastapi-app

aliabid94:
Looking at where we use root in the frontend:
### File requests, in `normalise_file`
```ts
export function normalise_file(
	file: Array<FileData> | FileData | string | null,
	root: string,
	root_url: string | null
): Array<FileData> | FileData | null {
  // ...     
	if (file.is_file) {
		if (root_url == null) {
			file.data = root + "/file=" + file.name;
		} else {
			file.data = "/proxy=" + root_url + "/file=" + file.name;
		}
	}
	return file;
}
```
  a) root will never end in slash, correct?
  b) can we get rid of `root_url`? If not, can you explain where that's used real quick
### Loading theme / stylesheets
```ts
		await mount_css(config.root + "/theme.css", document.head);
```
Looks fine.
### API Docs
```ts
	if (root === "") {
		root = location.protocol + "//" + location.host + location.pathname;
	}
	if (!root.endsWith("/")) {
		root += "/";
	}
        
// ...
	let response = await fetch(root + "info");

// ...
		let [response, status_code] = await post_data(
			`${root}run/${dependency.api_name}`,
			{
				data: inputs
			}
		);
```
I think we can get rid of the root formatting at the top. Everything else looks fine

abidlabs:
> a) root will never end in slash, correct?

It shouldn't as long as the user sets it up correctly. Here's what I say in the docstring for `root_path`: 

 The root path (or "mount point") of the application, if it's not served from the root ("/") of the domain. Often used when the application is behind a reverse proxy that forwards requests to the application. For example, if the application is served at "https://example.com/myapp", the `root_path` should be set to "/myapp".

We could check for a trailing slash and remove it just in case

> b) can we get rid of root_url? If not, can you explain where that's used real quick

Nope we need `root_url` because that's used in place of `root` when a Gradio app is loaded from an external source like Spaces using `gr.load()`. All files, etc. should use the `root_url` of the Space (since that's where they'd be generated and stored), not the `root_url` of the current Space.

We could rename `root_url` to be `remote_url` to be more clear

aliabid94 (approved):
ok LGTM

abidlabs:
> a) root will never end in slash, correct?

I just checked and even if a user does put the additional trailing slash, FastAPI removes it, so we should be good. 

It looks like all questions have been answered so I'll go ahead and merge! Thanks @aliabid94 @freddyaboulton for the detailed review.

meijunfeng2019:
location / {
            proxy_pass http://xx.xx.xx.xx:7860;
            proxy_set_header  Host  $host;
        }
       location /theme.css{
           proxy_pass http://xx.xx.xx.xx:7860/app/theme.css;
           proxy_set_header Host $host;
           proxy_set_header X-Real-Ip $remote_addr;
           proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
           proxy_set_header X-Forwarded-Proto $scheme;
       }
加上下面的 location /theme.css可以解决这个问题

Issue #4134, opened 2023-05-09
Username field in the flag report is blank


gsachdev-adeptia:
### Describe the bug

The flag log file created from Gradio is having username column but value is empty. 

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

#login function
def same_auth(username, password):
    return username == password

#setup gradio interface
iface = gr.Interface(fn=chatbot,
                    inputs=gr.components.Textbox(lines=7, label="Enter your text"),
                    outputs="text",
                   title="testing", allow_flagging="manual",flagging_options=["correct","wrong","ambiguous"])

#launch gradio
iface.launch(share=True ,auth=same_auth)

### Screenshot

![image](https://github.com/gradio-app/gradio/assets/12921808/c2888a2b-0985-43f1-9979-61eee5b6d009)

### Logs

```shell
NA
```

### System Info

```shell
Gradio Version 3.29.0,Operation System - Windows 10 Pro, Browser - Chrome  112.0.5615.138
```

### Severity

annoying

abidlabs:
Thanks @gsachdev-adeptia for reporting! This is indeed broken, though we have a fix here: #4135

abidlabs:
The fix is on `main` if you'd like to test it out.

gsachdev-adeptia:
I uninstall the gradio and installed it from the main branch as you indicated. Still not seeing any value filed in for the username in the flagged file. Do I need to do anything additional to have that written in the flagged file?

abidlabs:
I tested with this code and it's working for me on `main`:

```py
import gradio as gr
#login function
def same_auth(username, password):
    return username == password

#setup gradio interface
iface = gr.Interface(fn=lambda x:x,
    inputs=gr.components.Textbox(lines=7, label="Enter your text"),
    outputs="text",
    title="testing", allow_flagging="manual",flagging_options=["correct","wrong","ambiguous"])

#launch gradio
iface.launch(auth=same_auth)
```

Can you try it and see if it works for you?

gsachdev-adeptia:
I am still getting blank user name. Is the below pip to get from main branch correct
pip install https://gradio-main-build.s3.amazonaws.com/2cf13a1c698afed7f4418bc705902fa7868c6890/gradio-3.29.0-py3-none-any.whl

I suggest the above link is not correct.

abidlabs:
Please try:
```
[...]

PR #4135, opened 2023-05-10
Record username when flagging

abidlabs:
Fixes: #4134

Learned something very interesting when I was working on this. We were using `inspect.Parameter.annotation` in several places. This is a method that returns the type hint of a parameter. In Python 3.x, this returns a _class_ if the type hint is a class. [However, in the future, Python (4.x?), this will be replaced with a string](https://peps.python.org/pep-0563/). More importantly, if a user has:

```
```

in their code while running Python 3.x, this will also return a string (the future behavior). This was causing a subtle bug that took some time to figure out. Anyways, this is now fixed because we are using `typing.get_type_hints()` instead, which is the recommendation in the PEP.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,21 +2,11 @@
 
 ## New Features:
 
-- Support for asynchronous iterators by [@freddyaboulton](https://github.com/freddyaboulton) in [PR 3821](https://github.com/gradio-app/gradio/pull/3821)
-- Returning language agnostic types in the `/info` route by [@freddyaboulton](https://github.com/freddyaboulton) in [PR 4039](https://github.com/gradio-app/gradio/pull/4039)
+No changes to highlight.
 
 ## Bug Fixes:
 
-- Allow users to upload audio files in Audio component on iOS by by [@aliabid94](https://github.com/aliabid94) in [PR 4071](https://github.com/gradio-app/gradio/pull/4071). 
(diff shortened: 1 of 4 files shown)


gradio/helpers.py
```diff
@@ -619,19 +620,18 @@ def special_args(
     progress_index = None
     event_data_index = None
     for i, param in enumerate(positional_args):
+        type_hint = type_hints.get(param.name)
```
abidlabs, line 623: See note in PR on why this was changed

gradio/helpers.py
```diff
@@ -611,6 +611,7 @@ def special_args(
         updated inputs, progress index, event data index.
     """
     signature = inspect.signature(fn)
+    type_hints = utils.get_type_hints(fn)
```
abidlabs, line 614: See note in PR on why this was changed

CHANGELOG.md
```diff
@@ -2,21 +2,11 @@
 
 ## New Features:
 
-- Support for asynchronous iterators by [@freddyaboulton](https://github.com/freddyaboulton) in [PR 3821](https://github.com/gradio-app/gradio/pull/3821)
```
freddyaboulton, line 5: This line about the async generator got deleted by mistake

freddyaboulton (commented):

This breaks the `HuggingFaceDatasetSaver` because it's flag method does not accept a username. Once that's in this should be good to merge.

abidlabs:
Thanks @freddyaboulton for the review! I have made the fix so that usernames also work with the HF Savers. Thanks for catching that. Will merge once CI cooperates

Issue #4092, opened 2023-05-07
Extremely slow generation in chatbot in some cases


pseudotensor:
### Describe the bug

https://github.com/h2oai/h2ogpt/issues/54

For some inputs, when gradio needs to process the chat input and output, the time it takes to go through that processing is exceeding long.  E.g. can be 1 character every few seconds!

This is for two reasons:

1) The processing is very slow single core operation and on long inputs takes too long
2) The chat interface requires pushing through to chatbot the full content every update when doing streaming, leading to an O(N^2) problem.  This is fine in general, that streaming is dynamic and so any processing is dynamic, but for long inputs it reveals that the first problem (1) becomes too expensive.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

https://github.com/h2oai/h2ogpt/issues/54 shows examples.  E.g. when doing URLs, math, markdown leads to major slowdown if input into chatbot is large.

Key to reproduction is to copy-paste long and non-trivial text, so gradio thinks it needs to go through the markdown and other post-processing.

I identified the exact place is always in _postprocess_chat_messages, e.g.: https://github.com/h2oai/h2ogpt/issues/54#issuecomment-1515516527

I hacked a work-around: https://github.com/h2oai/h2ogpt/issues/54#issuecomment-1515520308

But it ends up now always showing even basic new lines so had to avoid the work-around.

So I have no work-around.

### Logs

```shell
NA
```

### System Info

```shell
gradio==3.27.0
gradio_client==0.1.3


Severity levels shown don't apply. It does not block all usage, but when it happens it makes for a horrible UX.
```

### Severity

annoying

abidlabs:
Thanks @pseudotensor for flagging this! cc @aliabid94 who has been looking into this

Issue #3680, opened 2023-03-29
Additional `Chatbot` Markdown Support


abetlen:
Hi there,

First I gotta say thank you for the awesome library, definitely speeds up my ml prototyping process a lot!

I was wondering if it's possible to extend the limited markdown support for the `Chatbot` block to include some additional expected features like headings, lists and tables. I'm trying to display structured data (list of query results for example) in the `Chatbot` body but currently this is quite difficult to do without support for tables and proper headings to seperate sections of the interface.

jrowen:
It looks like `Chatbot` should support html tables (https://github.com/gradio-app/gradio/blob/main/gradio/utils.py#L1010), but the two lines below prevent the table from rendering correctly (https://github.com/gradio-app/gradio/blob/main/gradio/components.py#L4597-L4598):

    chat_message = chat_message.replace("\n", "<br>")
    return self.md.renderInline(chat_message) 

If I replace them with the following, the table renders as expected:

    return self.md.render(chat_message)

abidlabs:
Hi @jrowen we have a WIP PR that, among other things, does this: https://github.com/gradio-app/gradio/pull/4048/files

Would you be able to see if this PR works for you (in adding support for tables, etc.)?

jrowen:
I'd be happy to test. What are the steps I need to follow to install via PR?  The following doesn't seem to be sufficient (seeing a 500 error):

    pip install git+https://github.com/gradio-app/gradio.git@chatbot-code-highlight

abidlabs:
@jrowen can you try `pip install https://gradio-builds.s3.amazonaws.com/7fddac232b2f6f652890dd5689d5d7313da27267/gradio-3.28.1-py3-none-any.whl`?

jrowen:
Looks good so far.  

It would be useful to include more plugins or allow users to specify which plugins to use.  For example, I'd like to use `emoji_plugin` and need to override the class to load:

```
from gradio import Chatbot as grChatbot
from mdit_py_emoji import emoji_plugin

class Chatbot(grChatbot):

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.md = self.md.use(emoji_plugin)
```

abidlabs:
Thanks @jrowen we've released the PR in `3.28.3` so it should be easier to use. Adding support for additional plugins is not a cool idea, we'll consider it!

abidlabs:
[...]

PR #4150, opened 2023-05-10
Using marked for chatbot markdown parsing

dawoodkhan82:
# Description
Move chatbot markdown parsing to frontend using `marked`, and `highlight.js`.

https://github.com/gradio-app/gradio/assets/12725292/57c542d0-4cab-47b4-989f-d1edd0d002aa

todo:
- fix css of code blocks
- cleanup code

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

Closes: #4146 

# Checklist:

# A note about the CHANGELOG

### DIFF
--- a/js/chatbot/src/ChatBot.svelte
+++ b/js/chatbot/src/ChatBot.svelte
@@ -1,4 +1,8 @@
 <script lang="ts">
+	import { marked } from 'marked';
+	import {markedHighlight} from "marked-highlight";
+	import hljs from 'highlight.js';
+	import 'highlight.js/styles/github.css';
 	import { beforeUpdate, afterUpdate, createEventDispatcher } from "svelte";
 	import type { Styles, SelectData } from "@gradio/utils";
 	import type { FileData } from "@gradio/upload";
@@ -14,6 +18,26 @@
 	export let feedback: Array<string> | null = null;
 	export let style: Styles = {};
 	export let selectable: boolean = false;
+
+	marked.setOptions({
+	renderer: new marked.Renderer(),
+	gfm: true,
+	tables: true,
+	breaks: true,
+	pedantic: false,
+	sanitize: true,
+	smartLists: true,
+	smartypants: false
+	});
+
+
+	marked.use(markedHighlight({
+	langPrefix: 'hljs language-',
+	highlight(code: string, lang: string) {
+		const language = hljs.getLanguage(lang) ? lang : 'plaintext';
+		return hljs.highlight(code, { language }).value;
+	}
+	}));
 
 	let div: HTMLDivElement;
 	let autoscroll: Boolean;
@@ -96,7 +120,7 @@
 							})}
 					>
 						{#if typeof message === "string"}
-							{@html message}
+							{@html marked.parse(message)}
 							{#if feedback && j == 1}
 								<div class="feedback">
 									{#each feedback as f}
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -4599,9 +4599,6 @@ class Chatbot(Changeable, Selectable, IOComponent, JSONSerializable):
             }
         elif isinstance(chat_message, str):
             chat_message = inspect.cleandoc(chat_message)
-            chat_message = cast(str, self.md.render(chat_message))
-            if chat_message.startswith("<p>") and chat_message.endswith("</p>\n"):
-                chat_message = chat_message[3:-5]
             return chat_message
         else:
             raise ValueError(f"Invalid message for Chatbot component: {chat_message}")
--- a/js/chatbot/package.json
+++ b/js/chatbot/package.json
@@ -10,6 +10,9 @@
 	"dependencies": {
 		"@gradio/theme": "workspace:^0.0.1",
 		"@gradio/utils": "workspace:^0.0.1",
-		"@gradio/upload": "workspace:^0.0.1"
+		"@gradio/upload": "workspace:^0.0.1",
+		"marked": "^5.0.1",
+		"marked-highlight":"^1.0.1",
+		"highlight.js":"^11.8.0"
 	}
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -117,7 +117,7 @@ importers:
         version: 4.7.4
       vite:
         specifier: ^4.2.1
-        version: 4.2.1(@types/node@17.0.14)
+        version: 4.2.1
       vitest:
         specifier: ^0.29.8
         version: 0.29.8(happy-dom@2.49.0)(playwright@1.27.1)
@@ -324,6 +324,15 @@ importers:
       '@gradio/utils':
         specifier: workspace:^0.0.1
         version: link:../utils
+      highlight.js:
+        specifier: ^11.8.0
+        version: 11.8.0
+      marked:
+        specifier: ^5.0.1
+        version: 5.0.1
+      marked-highlight:
+        specifier: ^1.0.1
+        version: 1.0.1(marked@5.0.1)
 
   js/code:
     dependencies:
@@ -1623,7 +1632,7 @@ packages:
       magic-string: 0.26.1
       svelte: 3.57.0
       svelte-hmr: 0.14.11(svelte@3.57.0)
-      vite: 4.2.1(@types/node@17.0.14)
+      vite: 4.2.1
     transitivePeerDependencies:
       - supports-color
     dev: false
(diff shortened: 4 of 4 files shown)


pngwn:
looks so cool.

Maybe if we add a min height to the chatbubbles that is roughly equal to one line, it will prevent that initial jumpy feeling?

pngwn:
Maybe consider prism over highlightjs for syntax highlighting. I think it is a little more complete and slightly better maintained. We're also using it elsewhere in the repo.

abidlabs:
@dawoodkhan82 this is still a draft or ready to be reviewed?

dawoodkhan82:
@abidlabs will be ready for review later today. have to make some fixes first

js/chatbot/src/ChatBot.svelte
```diff
@@ -1,8 +1,16 @@
 <script lang="ts">
+	import { marked } from "marked";
+	import Prism from "prismjs";
+	import "prismjs/components/prism-python";
```
dawoodkhan82, line 4: @pngwn I switched to `prism`, but I can't find a way to easily load additional languages without having to import each individually. Seems like `loadlanguages()` doesn't work.

aliabid94:
Great work @dawoodkhan82. Can we make the codeboxes more clearly separate from the rest of the content?
Screen Shot 2023-05-12 at 12 34 32 PM
Screen Shot 2023-05-12 at 12 34 43 PM
Screen Shot 2023-05-12 at 12 37 07 PM


There is more of a visible difference on main:
Screen Shot 2023-05-12 at 12 37 01 PM

abidlabs:
Haven't taken a deep look at this PR yet (having some issues building with pnpm) but wondering if this also solves (or can easily be adapted to solve) some of the other Markdown-related issues in the Chatbot, like rendering [headings, lists and tables](https://github.com/gradio-app/gradio/issues/3680) 

Also, does this work nicely with LaTeX rendering in the backend?

dawoodkhan82:
See updated screenshot for fixes. Ready for re-review @abidlabs @pngwn

abidlabs:
Very cool! Taking a look now

abidlabs:
This is fantastic @dawoodkhan82! Works quite well in my tests. This also closes #3680 I believe so I've added that to the parent comment. 

I did notice some design nits:

**Code**

1. In Dark mode, code with a specified language has a weird gold border and gold background color:

image

Code to reproduce:

````py
import gradio as gr

code = """
This is some code in Python:

```python
def test(x):
  return x
```
"""

value = [
    ("Hello", code),
    (code, "Nice!")
]

with gr.Blocks() as demo:
    gr.Chatbot(value)
    
demo.launch()
````

In light mode, it looks great. It would be good to check the other built-in gradio themes to confirm that they work appropriately

2. Code with no language specified has reduced margin

There's slightly less vertical margin around the code if no language is specified. I think it should be a little more:

image

Code to reproduce:

````py
import gradio as gr

code = """
This is some code in Python:

```
def test(x):
  return x
```
"""

value = [
    ("Hello", code),
    (code, "Nice!")
]

with gr.Blocks() as demo:
    gr.Chatbot(value)
    
demo.launch()
````

**Comment by abidlabs** [comment]
**Tables and Lists**

3. It would be good to add a tiny bit of vertical margin above and below the tables, just like there is with the code:

image

4. In Dark mode, there is no visible table border when the table is on the right side of the chat:

image

It would be good to test with the other built-in themes as well

5. The lists look good when they are on the left side of the chat. When they are on the right side, there isn't enough margin on the left side:

image

Code to reproduce the above:

````py
import gradio as gr

code = """
And here is a table:
| Syntax      | Description |
| ----------- | ----------- |
| Header      | Title       |
| Paragraph   | Text        |

And here is a list:
1. ad
2. ad
3. ad

And here are bullets
* sdf
* adslkfjasldkf
* dasf
"""

value = [
    ("Hello", code),
    (code, "Nice!")
]

with gr.Blocks() as demo:
    gr.Chatbot(value)
    
demo.launch()
````

**Comment by dawoodkhan82** [comment]
Fixed the styling issues @abidlabs

**Comment by abidlabs** [comment]
Looks lovely!

**Review by abidlabs — APPROVED** [review]

**Comment by freddyaboulton** [comment]
This also fixes #4143 

image

**Comment by abidlabs** [comment]
Oh actually this doesn't work with LaTeX anymore, right? As in LaTeX isn't rendered at all because we are no longer processing the markdown in the backend? Just tested with 

```py
import gradio as gr
import time

code = r"""
And here is an equation: $1+1=2$ and 
"""

value = [
    ("Hello", code),
    (code, "Nice!")
]

with gr.Blocks() as demo:
    c = gr.Chatbot(value)
    
demo.queue().launch()
```

and we get: 

image

dawoodkhan82:
@abidlabs Sure, I can look into latex support in a separate pr

abidlabs:
Based on @aliabid94’s benchmarking, this should close #4092 as well

Keldos-Li:
**little suggestion:**

How should you fix list styling:

```css
message-wrap :global(ol),
message-wrap :global(ul) {
    padding-inline-start: 2em;
}
```

Keldos-Li:
In [our own project](https://github.com/GaiZhenbiao/ChuanhuChatGPT) build with gradio, we rendered latex using mathjax:

https://github.com/gradio-app/gradio/assets/23137268/cbec6aa0-0439-48f3-88c1-d45129714a81

BUT PLEASE gives users or developers a choice of whether or not to render in-line formulas! Because the $ symbol can really cause a lot of problems! The $ in the table or the $ in the code block displaying shell code is likely to cause the entire display to be confused! Do not add formula rendering directly by default please. As the developer who actually uses Gradio for user-oriented projects, I believe we have encountered more issue with $ than you think...

Keldos-Li:
In fact, I felt really upset when I saw this PR 🤧. Maybe you can _let developers decide_ what to do with chatbot messages?😢 The origin method using python to post_process messages allows us to achieve some cool effects such as:
iShot_2023-05-18_上午12 57 55


And for us, at least we believe and implement the principle that all messages _sent by users_ should remain in their original format without being rendered by markdown or something. Though I know there may be different opinions, but at least, the origin method makes it possible for us to override and take control of the styles of the output... I'm now very nervous that we may lose our features with this gradio update.

*I would also say this should be written in **breaking change** rather than other changes*😕

or maybe something cool may happen with svelte? I'm not quite sure with that.

abidlabs:
Hi @Keldos-Li I think it should still be possible to achieve what you're showing in the screenshot. Basically the difference between the two is that in one case, you escape the Markdown, while in the other, you do not, correct? If you escape the Markdown and send it to the frontend, it will (should?) remain in its raw form even if the rendering is moved the frontend. 

Are there any other concerns that you have? There are a variety of reasons that we moved the rendering to the frontend (you can take a look at the linked issues to see, among them being: (1) more expected results when using Gradio Chatbots via APIs and (2) speed

abidlabs:
> **little suggestion:**
> 
> How should you fix list styling:
> 
> ```css
> message-wrap :global(ol),
> message-wrap :global(ul) {
>     padding-inline-start: 2em;
> }
> ```

Adding a little bit of padding on the left of the lists is not a bad idea @dawoodkhan82

Keldos-Li:
> If you escape the Markdown and send it to the frontend, it will (should?) remain in its raw form even if the rendering is moved the frontend. 

That makes sense！We'll try to handle this. Thanks for the reply! 

Now I understand why it's important for you to move rendering to the front-end. You are right.

Yeah... I think I'll just have no more concerns for now if we still get a chance to take control of the message output. Thanks.

Gradio is such a great app, we really appreciate your contribution.

pseudotensor:
Awesome!

abidlabs:
> I was wondering if it's possible to set the sanitize option in marked.setOptions to false? 

We are working on exactly this! PR in progress: https://github.com/gradio-app/gradio/pull/4360

Apologies for the breaking change, but we are working on restoring functionality soon & hopefully with the move the frontend, processing should be much faster.

Keldos-Li:
> We are working on doing something like this in this PR in progress: https://github.com/gradio-app/gradio/pull/4360 (feel free to chime in if that does not fix it for you)

Such great news! Thank you!!!

jefffohl:
Thanks for this @dawoodkhan82 - this is great. However, I noticed that there seems to be a problem with dark/light mode handling and the Prism code highlighting. When the operating system has selected "light" or "dark" themes, the Chatbot component works fine. However, when using the dark/light mode theming pattern established elsewhere in Gradio, which uses the class name `dark` applied to the `<body>` element, we run into problems. When the OS is indicating the preferred theme to be "dark", but then we toggle the them to "light" using the classname method, the Prism code highlighting is not also toggled. 

Example: 
Screenshot 2023-06-13 at 9 27 59 AM

then, toggling the "Dark Mode" by clicking the button which applies the class `dark` to the `<body>` element of the app, which enables the dark mode of the theme (as applied in the `base` Gradio theme):

Screenshot 2023-06-13 at 9 28 09 AM

We see that the logic for applying the Prism code highlighting is separate from the CSS pattern described above, which results in what is seen in the second example shown above.

Am I simply implementing this incorrectly, or is there more work to do here to sync up the code highlighting dark mode with the Gradio theme for dark mode handling? If the latter, I would be happy to try and help.

pseudotensor:
Thanks for giving details @jefffohl !

dawoodkhan82:
@jefffohl seems like its a bug, where the prism highlighting doesn't react to the dark mode change. Can you open a new issue with this context and assign it to me? I'll take a look. Thanks!

jefffohl:
Will do - thanks @dawoodkhan82 !

jefffohl:
@dawoodkhan82 issue created ^^ - I don't seem to have permission to assign it to anyone though. Thanks for your help!

PR #4153, opened 2023-05-10
Fix api info for File component

freddyaboulton:
# Description

There were two issues before:
- The type agnostic info for `File(file_count='multiple')` showed that it could be either an array or a single item. Now it only shows it can be an array
- Example inputs for `File(file_count='multiple')` will now show a list. This will only apply for new demos going forward.

image

In order for CI to pass we need to release a new version of the python client

# Checklist:

# A note about the CHANGELOG


client/python/gradio_client/utils.py
```diff
@@ -29,7 +29,7 @@
 CONFIG_URL = "/config"
 API_INFO_URL = "/info"
 RAW_API_INFO_URL = "/info?serialize=False"
-SPACE_FETCHER_URL = "https://gradio-space-api-fetcher-v2.hf.space/api"
+SPACE_FETCHER_URL = "https://freddyaboulton-space-api-fetcher-dev.hf.space/api"
```
freddyaboulton, line 32: Will install this PR to the v2 api fetcher space once it's approved

gradio/blocks.py
```diff
@@ -551,8 +551,13 @@ def get_api_info(config: dict, serialize: bool = True):
                 continue
             label = component["props"].get("label", f"value_{o}")
             serializer = serializing.COMPONENT_MAPPING[type]()
-            assert isinstance(serializer, serializing.Serializable)
-            info = serializer.api_info()
+            if component.get("api_info") and after_new_format:
```
freddyaboulton, line 554: This is the same logic applied to inputs also applied to outputs. I think it was just an oversight it wasn't done before.

pngwn (approved):
Thanks @freddyaboulton! Looks good to me but I'll let someone else take a look as well. I'll also need to switch the version check in the js client like you have here.

abidlabs:
I'm confused why we have this special case:

```py
                if (
                    isinstance(serializer, serializing.FileSerializable)
                    and component["props"].get("file_count", "single") != "single"
                ):
                    python_info = serializer._multiple_file_serialized_info()
```

in `get_api_info()` in `blocks.py`. Shouldn't it pull from `FileSerializable` since the same logic exists there?

Nit:
* The `FileSerializable` class is starting to have a lot of one-line methods, whereas it would be a lot simpler just to have if/else statements. For example, we can replace:

```py
    def example_inputs(self) -> dict[str, Any]:
        return self._single_file_example_inputs()

    def _single_file_example_inputs(self) -> dict[str, Any]:
        return {
            "raw": {"is_file": False, "data": media_data.BASE64_FILE},
            "serialized": "https://github.com/gradio-app/gradio/raw/main/test/test_files/sample_file.pdf",
        }

    def _multiple_file_example_inputs(self) -> dict[str, Any]:
        return {
            "raw": [{"is_file": False, "data": media_data.BASE64_FILE}],
            "serialized": [
                "https://github.com/gradio-app/gradio/raw/main/test/test_files/sample_file.pdf"
            ],
        }
```

with 

```py
    def example_inputs(self, file_count="single") -> dict[str, Any]:
        if file_count=="single":
          return {
              "raw": {"is_file": False, "data": media_data.BASE64_FILE},
              "serialized": "https://github.com/gradio-app/gradio/raw/main/test/test_files/sample_file.pdf",
          }
        else:
          return {
              "raw": [{"is_file": False, "data": media_data.BASE64_FILE}],
              "serialized": [
                  "https://github.com/gradio-app/gradio/raw/main/test/test_files/sample_file.pdf"
              ],
          }
```
which is more condensed

abidlabs:
The actual funcitonality looks good though, LGTM

abidlabs (approved):

abidlabs (approved):

freddyaboulton:

We pull the serialized data format by instantiating the serializer but that serializer doesn't know about single vs multiple since that's from the component. So we have to check the component config to see if it's a multiple file component and then get the multiple file info from the serializer.

Yea I get what you mean about the single line methods but I don't think it's good to add a parameter to example_inputs as the api for example_inputs should be the same across serializers.

abidlabs:
I see, thanks for the explanation. The info generation is starting to get a bit messy and hard to maintain IMO. Let’s tackle that in a future PR. This LGTM

freddyaboulton:
Yes! I think a lot of the complexity comes from having to return the serialized types in the server. In an ideal world, the serialized types are an implementation detail of the client and the server wouldn't know about that.

PR #4128, opened 2023-05-09
Dispatch dropdown change on value change for any reason

aliabid94:
fixes: #4076

PR #4157, opened 2023-05-10
Input listener

aliabid94:
Created new `.input` listener, which specifically listens to component value changes as a result of user input (vs `.change` which responds to component value changes for *any* reason, including being updated as part of an output of an event listener.) 

Fixes: #1431


abidlabs (approved):

abidlabs:
Actually there does seem to be a problem with Dropdown with multiselect=True

```py
import gradio as gr

with gr.Blocks() as demo:
    txt = gr.Textbox()
    value = gr.Dropdown(choices=['A', 'B', 'C'], multiselect=True)
    btn = gr.Button()

    value.input(fn=lambda x: str(x), inputs=value, outputs=txt)
    btn.click(fn=lambda: ['C'], outputs=value)

demo.launch()
```

I noticed two issues:
* Changing the dropdown manually crashes the app (similar to my comment on #4128)
* Clicking on the button shouldn't affect the Textbox at all (since I'm using input), but interestingly, it sets the value of the Textbox to [], which means the event is firing but without the payload? See:

Recording 2023-05-11 at 08 36 34

aliabid94:
Added dataframe `.input` listener, also fixed issue with dropdown. Also fixed a pretty major bug where adding a new column to dataframe broke everything. 
@pngwn, would be great if you could take a quick look to approve the overall approach.

pngwn:
Yeah sorry been busy finalising the client today. Will review it first thing tomorrow!

abidlabs:
LGTM everything looks great from my tests! Thanks @aliabid94 for the nice PR

abidlabs (approved):

pngwn (approved):
Eventually i think it would be good to move this logic out of the components and up a layer but I think that will require some refactoring, we can do that when we refactor for 4.0.

lgtm!

abidlabs:
Thanks @aliabid94, I'll go ahead and merge this in so that we can get it out in the release

PR #4163, opened 2023-05-11
Simplifies format of Changelog

abidlabs:
Reduces the number of categories in the changelog to 4:

* new features
* bug fixes
* other changes
* breaking changes

This should make it easier for contributors to add changes in the right section, as well as clarify that documentation or website-related changes do not need to be included in the changelog.


pngwn (commented):
Looks good to me! Do we need to update the action? I think it generates new sections when we clear the changelog?

pngwn (approved):
Awesome.

freddyaboulton:
This deploy-pypi gh action is running this script: https://github.com/gradio-app/gradio/blob/main/scripts/format_release_notes.py which has the old section names.

abidlabs:
Wait what, we have an action for this? I've always done it manually lol

Issue #3868, opened 2023-04-15
/info route breaks for entire app if certain components are present (interpretation, annotatedimage)


aliabd:
### Describe the bug

Some of the demos don't have the /info route, which is breaking the use via api page. example:

https://colab.research.google.com/github/gradio-app/gradio/blob/main/demo/sentiment_analysis/run.ipynb

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Click use via api button on this colab's demo: 

https://colab.research.google.com/github/gradio-app/gradio/blob/main/demo/sentiment_analysis/run.ipynb

### Logs

```shell
Failed to load resource: the server responded with a status of 500 ()       /info:1
```

### System Info

```shell
gradio v3.27.0 
gradio-client v0.1.3
```

### Severity

annoying

freddyaboulton:
For the sentiment classification demo, the problem is caused by the interpretation component cc @abidlabs 


![image](https://user-images.githubusercontent.com/41651716/232520725-f2b486d3-8eca-45e9-a199-40d291942c2f.png)

abidlabs:
Ahhh interpretation strikes again. Starting to agree with @aliabid94 that we should nuke it in 4.0

But yeah this should be an easy fix -- can take of it

segments-tobias:
Does the new `AnnotatedImage` component also break the API page?
It's not working for my demo: [https://huggingface.co/spaces/segments/panoptic-segment-anything](https://huggingface.co/spaces/segments/panoptic-segment-anything)

abidlabs:
Thanks @segments-tobias! No need to create a new issue, I'll tackle both issues in the same PR

PR #4167, opened 2023-05-11
Adds missing components to gradio client serializing's component mapping

abidlabs:
Fixes: #3868

Test by running e.g. `demo\sentiment_analysis\run.py` and clicking on "view API"

Also tees up a new release of the Client


test/test_components.py
```diff
@@ -1474,10 +1474,9 @@ def test_component_functions(self):
 
 class TestNames:
     # This test ensures that `components.get_component_instance()` works correctly when instantiating from components
-    def test_no_duplicate_uncased_names(self):
-        subclasses = gr.components.Component.__subclasses__()
-        unique_subclasses_uncased = {s.__name__.lower() for s in subclasses}
-        assert len(subclasses) == len(unique_subclasses_uncased)
+    def test_no_duplicate_uncased_names(self, io_components):
```
abidlabs, line 1477: Unrelated, but I realized that this test was set up incorrectly since it only looked at the 4 direct subclasses of `gr.Component` instead of all of the descendants. Refactored to use the fixture.

freddyaboulton (approved):
Looks good to me @abidlabs !

abidlabs:

PR #4173, opened 2023-05-11
hotfix for test

abidlabs:
Quick fix for one test in python client library


client/python/test/test_client.py
```diff
@@ -747,7 +747,8 @@ def test_file_io(self, file_io_demo):
                 "description": "List of filepath(s) or URL(s) to files",
             }
             # Will change to list when we do the next gradio release
-            assert isinstance(inputs[0]["example_input"], str)
+            assert isinstance(inputs[0]["example_input"], list)
```
freddyaboulton, line 750: I'm so good at predicting the future 😂

freddyaboulton (approved):
Thanks @abidlabs ! I guess we can get rid of the comment now.

PR #4194, opened 2023-05-12
Do not do any analytics requests if analytics are disabled

akx:
# Description

* relevant motivation
  * https://github.com/AUTOMATIC1111/stable-diffusion-webui/pull/10324
  * https://github.com/oobabooga/text-generation-webui/blob/c746a5bd00c39776bec0091a570f3348aa785caa/server.py#L12-L16
* a summary of the change 
  * Gradio should not call home if the user asks to disable analytics. That should be pretty self-evident.

# Checklist:


abidlabs:
Thanks for creating the PR @akx. I don't follow the original issue. In the `Blocks` class, we already check if `self.analytics_enabled` before making any call to the analytics methods: `utils.error_analytics()`, `utils.launched_telemetry()`, `utils.initiated_analytics(), etc.. In other words, this PR would not have any effect. Am I missing something?

akx:
Please see the linked issues. People are seeing Gradio do network calls at initial import and working around that.

akx:
@abidlabs To elaborate:

* In particular, `Interface.__init__()` seems to unconditionally call `version_check()`, which calls home to `https://api.gradio.app/pkg-version`.
* All of the analytics functions also use Amazon's https://checkip.amazonaws.com/ service, another third party. (Furthermore, I don't think it's quite OK that supposedly anonymous analytics data also includes a public IP address.)

akx:
Done, @abidlabs

abidlabs:
Thanks for the important fix @akx and the cleanup. LGTM 

I'd like to get a review by @dawoodkhan82 and then we should be good to merge in.

abidlabs (approved):

gradio/analytics.py
```diff
@@ -0,0 +1,187 @@
...
+    threading.Thread(
+        target=_do_analytics_request,
+        kwargs={
+            "url": f"{ANALYTICS_URL}gradio-initiated-analytics/",
+            "data": {**data, "ip_address": get_local_ip_address()},
+        },
+    ).start()
+
+
+def launched_telemetry(blocks: gradio.Blocks, data: dict[str, Any]) -> None:
```
dawoodkhan82, line 97: Can we rename this to `launched_analytics`

gradio/interface.py
```diff
@@ -362,7 +362,7 @@ def __init__(
         self.local_url = None
 
         self.favicon_path = None
-        utils.version_check()
+        analytics.version_check()
```
dawoodkhan82, line 365: This check won't work now if analytics is disabled. @abidlabs

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 ## Bug Fixes:
 
+- Fix "TypeError: issubclass() arg 1 must be a class" When use Optional[Types] by [@lingfengchencn](https://github.com/lingfengchencn) in [PR 4200](https://github.com/gradio-app/gradio/pull/4200). 
+- Gradio will no longer send any analytics if analytics are disabled with the GRADIO_ANALYTICS_ENABLED environment variable. By [@akx](https://github.com/akx) in [PR 4194](https://github.com/gradio-app/gradio/pull/4194)
```
dawoodkhan82, line 12: Were analytics being sent when the env variable was set to false before this PR?

dawoodkhan82 (changes requested):

CHANGELOG.md
abidlabs: What we call analytics weren't being sent, but network requests were being made to the Gradio server, see https://github.com/gradio-app/gradio/pull/4194#issuecomment-1546937585

gradio/interface.py
abidlabs: Yes I think that's an acceptable sacrifice to ensure no network requests to gradio servers

gradio/analytics.py
abidlabs: Agreed, I'll make the change

abidlabs:
Thanks @dawoodkhan82 for giving this a look. Renamed the function and we should be good for a merge once CI passes

PR #4204, opened 2023-05-13
Fix Explanation Regarding NumPy Image Component

der3318:
# Description

In docs and README, gradio has the following statement:

> An "Image" component with type "numpy" expects a numpy array with shape (**width**, **height**, 3).

But the app actually behaves in the opposite way:

```python
import gradio as gr # gradio==3.29.0
import numpy as np  # numpy==1.24.3

demo = gr.Interface(
    fn=lambda nparray, pil: (nparray.shape, pil.size, np.array(pil).shape),
    inputs=[
        gr.Image(type="numpy", label="as numpy"),
        gr.Image(type="pil", label="as PIL"),
    ],
    outputs=[
        gr.Textbox(label="nparray.shape"),
        gr.Textbox(label="pil.size"),
        gr.Textbox(label="np.array(pil).shape"),
    ],
)
demo.launch()
```

Imgur

In fact numpy uses row-major schema for data storing, having shape (**height**, **width**, 3) by design. This also applies when using `np.array(im)` (e.g., in [_format_image](https://github.com/gradio-app/gradio/blob/main/gradio/components.py#L1746)) to convert a PIL object into ndarray, even though PIL itself is indeed column-major. See PIL's implmentaion of [numpy array interface support](https://github.com/python-pillow/Pillow/blob/main/src/PIL/Image.py#L650) and [_conv_type_shape](https://github.com/python-pillow/Pillow/blob/main/src/PIL/Image.py#L228) for more context.

I believe updating docs would be a better choice, instead of changing the underlying impl.

# Checklist:

# ~~A note about the CHANGELOG~~

Tagging "no-changelog-update" needs permission. I assume changelog is not required for the patch. Please let me know if it's not.


abidlabs:
You are indeed correct @der3318! Thank you for the fix the docs and the demo. Added a line in the changelog crediting you with the fix, and will merge this in when CI passes.

Issue #4216, opened 2023-05-15
On Python 3.9, `typing.get_type_hints()` throws an error if "|" is used in the type hint


abidlabs:
### Describe the bug

The following code works with Python 3.10, but not with Python 3.9, as reported by @hysts 

`app.py`

```py
#!/usr/bin/env python

import gradio as gr

from utils import fn

with gr.Blocks() as demo:
    text = gr.Text()
    text2 = gr.Text()
    text.submit(fn=fn, inputs=text, outputs=text2)

demo.queue(api_open=False).launch()
```

`utils.py`

```py
from __future__ import annotations

def fn(a: str | int) -> str:
    return f'{a}'
```

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

See above

### Logs

```shell
Traceback (most recent call last):
  File "/home/hysts/.pyenv/versions/3.9.16/lib/python3.9/multiprocessing/process.py", line 315, in _bootstrap
    self.run()
  File "/home/hysts/.pyenv/versions/3.9.16/lib/python3.9/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/home/hysts/.virtualenvs/py3916/lib/python3.9/site-packages/uvicorn/_subprocess.py", line 76, in subprocess_started
    target(sockets=sockets)
  File "/home/hysts/.virtualenvs/py3916/lib/python3.9/site-packages/uvicorn/server.py", line 60, in run
    return asyncio.run(self.serve(sockets=sockets))
  File "/home/hysts/.pyenv/versions/3.9.16/lib/python3.9/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/home/hysts/.pyenv/versions/3.9.16/lib/python3.9/asyncio/base_events.py", line 647, in run_until_complete
    return future.result()
  File "/home/hysts/.virtualenvs/py3916/lib/python3.9/site-packages/uvicorn/server.py", line 67, in serve
    config.load()
  File "/home/hysts/.virtualenvs/py3916/lib/python3.9/site-packages/uvicorn/config.py", line 477, in load
    self.loaded_app = import_from_string(self.app)
  File "/home/hysts/.virtualenvs/py3916/lib/python3.9/site-packages/uvicorn/importer.py", line 21, in import_from_string
    module = importlib.import_module(module_str)
[...]

PR #4228, opened 2023-05-16
Fix for `typing.get_type_hints()` on Python 3.9 or lower

abidlabs:
A fix for the issue pointed out by @hysts which was breaking Gradio demos running `Python<=3.9`

It's not a perfect fix, as it parses type annotations as strings, so for example for `gr.Request`, the type annotation has to be `gr.Request` or `gradio.Request` or `Request` for it to be recognized correctly. But I don't believe there is any other way to do it, and this should be strictly better than what we have right now since this alternative logic is only used if an exception would be have been thrown otherwise. 

Fixes: #4216

### DIFF
--- a/gradio/utils.py
+++ b/gradio/utils.py
@@ -707,16 +707,16 @@ def get_type_hints(fn):
                 continue
             if "|" in str(param.annotation):
                 continue
-                # To convert the string annotation to a class, we use the
-                # internal typing._eval_type function. This is not ideal, but
-                # it's the only way to do it without eval-ing the string.
-                # Since the API is internal, it may change in the future.
-                try:
-                    type_hints[name] = typing._eval_type(  # type: ignore
-                        typing.ForwardRef(param.annotation), globals(), locals()
-                    )
-                except (NameError, TypeError):
-                    pass
+            # To convert the string annotation to a class, we use the
+            # internal typing._eval_type function. This is not ideal, but
+            # it's the only way to do it without eval-ing the string.
+            # Since the API is internal, it may change in the future.
+            try:
+                type_hints[name] = typing._eval_type(  # type: ignore
+                    typing.ForwardRef(param.annotation), globals(), locals()
+                )
+            except (NameError, TypeError):
+                pass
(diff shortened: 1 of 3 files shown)


hysts (approved):

gradio/utils.py
```diff
@@ -683,11 +683,40 @@ async def cancel(session_hash: str) -> None:
...
+    except TypeError:
+        # On Python 3.9 or earlier, get_type_hints throws a TypeError if the function
+        # has a type annotation that include "|". We resort to parsing the signature
+        # manually using inspect.signature.
+        type_hints = {}
+        sig = inspect.signature(fn)
+        for name, param in sig.parameters.items():
+            if param.annotation is not inspect.Parameter.empty and "|" not in str(
+                param.annotation
+            ):
```
akx, line 708: This could be cleaner if it was separated to two negated `if`s:

```suggestion
            if param.annotation is inspect.Parameter.empty:
                continue
            if "|" in str(param.annotation):
                continue
```

abidlabs: Sure thanks for the suggestion @akx

gradio/utils.py
```diff
@@ -683,11 +683,41 @@ async def cancel(session_hash: str) -> None:
...
+        for name, param in sig.parameters.items():
+            if param.annotation is inspect.Parameter.empty:
+                continue
+            if "|" in str(param.annotation):
+                continue
+                # To convert the string annotation to a class, we use the
+                # internal typing._eval_type function. This is not ideal, but
+                # it's the only way to do it without eval-ing the string.
+                # Since the API is internal, it may change in the future.
+                try:
```
freddyaboulton, line 714: This would never run because it's after the continue. Am I missing something?

abidlabs: Whoops indentation mistake. Let me fix

freddyaboulton (approved):
Verified - thanks @abidlabs for the pr and correction!

abidlabs:
Thanks @freddyaboulton for the catch!

Issue #4221, opened 2023-05-15
Python client strips underscores from parameter names in view api method


freddyaboulton:
### Describe the bug

When using `view_api` in the python client, any underscores are stripped from parameter names and that can make them harder to read.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```python
import gradio_client as grc

client = grc.Client("curt-park/segment-anything-with-clip")
client.view_api()
```

### Screenshot
The actual labels are `predicted_iou_threshold`, etc

![image](https://github.com/gradio-app/gradio/assets/41651716/e8029498-838e-4245-9ab8-7a3b81354ccb)

### Logs

```shell
-
```

### System Info

```shell
latest main
```

### Severity

annoying

Issue #4081, opened 2023-05-06
chatbot in gradio server operates through json files, so gradio client cannot actually use remotely


pseudotensor:
### Describe the bug

Related:

https://github.com/h2oai/h2ogpt/issues/115#issuecomment-1536936221

This was my quick test to get chatbot through client working, but because it passes back files via API, it's useless remotely.  So in later versions of the test code I just removed that code.  But here it is in old state:

https://github.com/h2oai/h2ogpt/blob/05c742a1a283bf307bc745e8f276b0678f0ed9a9/client_test.py#L81-L107

Work-around is to just not support chatbot for client, which is disappointing.

Did I miss something?

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

https://github.com/h2oai/h2ogpt/blob/05c742a1a283bf307bc745e8f276b0678f0ed9a9/client_test.py#L81-L107

### Logs

```shell
Not needed
```

### System Info

```shell
gradio==3.27.0
gradio_client==0.1.3
```

### Severity

serious, but I can work around it

abidlabs:
Could you try setting `serialize=False` when you instantiate the Client? I think it should do what you need it to do:

https://github.com/gradio-app/gradio/blob/b66ecff6715be03030b3547c30d26dc3902f9094/client/python/gradio_client/client.py#LL66C15-L66C31

Just a head's up, I think this is still a little beta so signatures may change. Maybe we also need a better name for this parameter if you have any suggestions

pseudotensor:
Ok, sorry if I missed that, will try it and report back.

pseudotensor:
Hi @abidlabs .  The problem with serialize=False is that I use a gr.State() as an input on server side.  If I serialize=True, even my normal non-chatbot client calls work fine even though I'm only passing in all objects except the state.

However, if I use serialize=False, then even that basic test fails saying I'm missing an input, which is the state.

So forget about chatbot for now, just normal click behavior changes with serialize=True or False.

Here's the gradio server part:

https://github.com/h2oai/h2ogpt/blob/main/gradio_runner.py#L743-L746

See it has [model_state] as first entry.

Then see client code:

https://github.com/h2oai/h2ogpt/blob/main/client_test.py#L101-L106

See I'm using that API name, but not passing state.

In general, I cannot literally pass a state object (in general is a giant LLM) to the server from the client.  The state should be something dynamic on the server yet no need to pass.

Thanks for any advice!

[...]

PR #4230, opened 2023-05-16
Some tweaks to the Client

abidlabs:
This PR:

* Fixes a small typo in `Client.view_api(all_endpoints=True)`
* Brings back underscores in parameter names when displaying API info (Closes: #4221)
* Moves the state-removal logic outside of serialization (Closes: #4081) and adds a test
* Fixes a separate bug, which is that state-removal logic wouldn't be applied to generators
* Releases a new version of the Client, since we likely (hopefully) won't work on it for the rest of the 3.x milestone


client/python/test/test_client.py
```diff
@@ -308,6 +308,18 @@ def test_upload_file_upload_route_does_not_exist(self):
                 client.submit(1, "foo", f.name, fn_index=0).result()
                 serialize.assert_called_once_with(1, "foo", f.name)
 
+    def test_state_without_serialize(self, stateful_chatbot):
```
freddyaboulton, line 311: We can add a `serialize` argument to `connect` to use it in this test.

freddyaboulton (approved):
Thanks for the fix @abidlabs ! Confirmed it fixes the original issue

client/python/test/test_client.py
abidlabs: Agreed, that would be better!

abidlabs: Fixed

abidlabs:
🧡

Issue #3794, opened 2023-04-09
queue sometimes hangs forever with share=True


oobabooga:
### Describe the bug

In `gradio==3.24.1`, queue events sometimes hang and never complete when executed through a gradio share link. This severely impacts Google Colab usage.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```python
import gradio as gr
import time

with gr.Blocks() as interface:
    txt1 = gr.Textbox(lines=1)
    txt2 = gr.Textbox(lines=1)
    txt3 = gr.Textbox(lines=1)
    txt4 = gr.Textbox(lines=1)
    txt5 = gr.Textbox(lines=1)
    btn = gr.Button("Submit")

    btn.click(lambda: "Hi1", None, txt1).then(
        lambda: "Hi2", None, txt2).then(
        lambda: "Hi3", None, txt3).then(
        lambda: "Hi4", None, txt4).then(
        lambda: "Hi5", None, txt5)

interface.queue()
interface.launch(share=True)
```

Launch the interface above, enter the share link, and click on the Submit button. If it doesn't hang in the first try, click on Submit again. It usually doesn't take more than 5 attempts to get an infinite loading bar.

### Screenshot

![hanged](https://user-images.githubusercontent.com/112222186/230753907-ba75c924-aa6c-437d-82de-ad3220bf01b6.png)

In this example, the interface hanged for 78 seconds once it reached the third event in the queue.

### Logs

```shell
No errors are printed
```

### System Info

```shell
gradio==3.24.1
Linux
```

### Severity

blocking upgrade to latest gradio version

abidlabs:
Hmmm that's strange @oobabooga we'll look into it!

freddyaboulton:
Definitely the share link - running locally is very snappy

![chain_ws_share](https://user-images.githubusercontent.com/41651716/231198884-04618f8c-e0d0-4458-bd51-1eddcead0efa.gif)

oobabooga:
@freddyaboulton I have tested `gradio==3.24.1` and `gradio==3.25.0` and the problem persists.

![3 24 1](https://user-images.githubusercontent.com/112222186/231216066-a243acd3-a4b7-4c92-9385-d949f034f7ec.png)

It is worth mentioning that this kind of freezing doesn't seem to happen in `gradio==3.18.0`, so I believe that it is due to a recent change.

oobabooga:
This issue prevents me from creating a colab notebook for my web ui :(

[...]

PR #4235, opened 2023-05-16
Increase websocket messaging timeout

abidlabs:
Small fix that increases the timeout of the websocket messaging (how long the websocket tries to send the message before reporting a timeout). Initially, this was set to 1 second, which was causing timeouts when users would be on slow networks with sharing enabled. Have increased this to 5 seconds. 

To test, first run copy-paste the share template (`cp .\gradio\templates\frontend\index.html .\gradio\templates\frontend\share.html`) so that you do sharing with local builds of gradio. Then run this:

```py
import gradio as gr

with gr.Blocks() as interface:
    txt1 = gr.Textbox(lines=1)
    txt2 = gr.Textbox(lines=1)
    txt3 = gr.Textbox(lines=1)
    txt4 = gr.Textbox(lines=1)
    txt5 = gr.Textbox(lines=1)
    btn = gr.Button("Submit")

    btn.click(lambda: "Hi1", None, txt1).then(
        lambda: "Hi2", None, txt2).then(
        lambda: "Hi3", None, txt3).then(
        lambda: "Hi4", None, txt4).then(
        lambda: "Hi5", None, txt5)

interface.queue()
interface.launch(share=True)
```

Then open the share link, and in your chrome inspector network tab, set throttling to "Slow 3G." Then run the demo, and it should work without having any infinitely long loading bars. 

Closes: #3794

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@
 
 - Update the js client by [@pngwn](https://github.com/pngwn) in [PR 3899](https://github.com/gradio-app/gradio/pull/3899)
 - Fix documentation for the shape of the numpy array produced by the `Image` component by [@der3318](https://github.com/der3318) in [PR 4204](https://github.com/gradio-app/gradio/pull/4204).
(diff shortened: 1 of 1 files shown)


abidlabs:
Thanks @freddyaboulton

Issue #4237, opened 2023-05-16
JSONDecodeError when importing gradio behind firewall


davidai:
### Describe the bug

When simplejson package is installed and trying to import gradio on a machine without an external internet connection, JSONDecodeError occurs.

I plan to submit a PR for this issue.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

`import gradio`

### Logs

```shell
Exception in thread Thread-1:
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/requests/models.py", line 971, in json
    return complexjson.loads(self.text, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/simplejson/__init__.py", line 514, in loads
    return _default_decoder.decode(s)
  File "/usr/local/lib/python3.8/dist-packages/simplejson/decoder.py", line 386, in decode
    obj, end = self.raw_decode(s)
  File "/usr/local/lib/python3.8/dist-packages/simplejson/decoder.py", line 416, in raw_decode
    return self.scan_once(s, idx=_w(s, idx).end())
simplejson.errors.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
 
During handling of the above exception, another exception occurred:
 
Traceback (most recent call last):
  File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.8/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/local/lib/python3.8/dist-packages/gradio/strings.py", line 38, in get_updated_messaging
    updated_messaging = requests.get(MESSAGING_API_ENDPOINT, timeout=3).json()
  File "/usr/local/lib/python3.8/dist-packages/requests/models.py", line 975, in json
    raise RequestsJSONDecodeError(e.msg, e.doc, e.pos)
```

### System Info

```shell
gradio 3.29.0, Python 3.8.10, Ubuntu 20.04.4 LTS
```

### Severity

blocking all usage of gradio

freddyaboulton:
Do you want to put up a PR to catch this error @davidai ?

davidai:
I have the [changes](https://github.com/gradio-app/gradio/compare/main...davidai:gradio:fix_jsondecodeerror) ready but having difficulty following the steps in the `CONTRIBUTING.md` file. In particular, `bash scripts/build_frontend.sh` is giving the error

```
[...]

PR #4241, opened 2023-05-16
Fix JSONDecodeError

davidai:
# Description

When using gradio behind a company firewall, `import gradio` will cause requests to generate a JSONDecodeError. The current implementation only tests for `json.decoder.JSONDecodeError`. But if simplejson is installed the requests library will use `simplejson.errors.JSONDecoderError` instead.

Modified except block to catch exception used from requests package.

Closes: #4237

# Checklist:


abidlabs:
Thanks for the PR @davidai! Honestly this actually such a big issue (`gradio` not being able to be imported) that I would be in favor of catching all exceptions and letting them pass

davidai:
@abidlabs Made the requested change. Code now catches `Exception`. Should I also resolve the conflict with `CHANGELOG.md`?

abidlabs:
Thanks @davidai I can take care of that

Issue #3980, opened 2023-04-26
run_on_click flag of gr.Examples is not working well on the first try


MrSyee:
### Describe the bug

I use gr.Blocks and gr.Examples like:
```
def mirror(x):
    print(x.shape())
    return x

with gr.Blocks() as demo:
    with gr.Row():
        input_img = gr.Image(label="Input image").style(height=600)
        output_img = gr.Image(label="Output image").style(height=600)

    gr.Markdown("## Image Examples")
    gr.Examples(
        examples=[
            [os.path.join(os.path.dirname(__file__), "examples/dog.jpg")]
        ],
        inputs=[input_img],
        outputs=output_img,
        fn=mirror,
        run_on_click=True,
    )
```

In UI, I bumped into error when I clicked example.
<img width="1366" alt="image" src="https://user-images.githubusercontent.com/17582508/234595067-eb7acc07-a207-4562-bd68-7b3b7f1ca48a.png">

But on the second click, It is working well.

I think the example doesn't save well to the variable on the first click.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```
def mirror(x):
    print(x.shape())
    return x

with gr.Blocks() as demo:
    with gr.Row():
        input_img = gr.Image(label="Input image").style(height=600)
        output_img = gr.Image(label="Output image").style(height=600)

    gr.Markdown("## Image Examples")
    gr.Examples(
        examples=[
            [os.path.join(os.path.dirname(__file__), "examples/dog.jpg")]
        ],
        inputs=[input_img],
        outputs=output_img,
        fn=mirror,
        run_on_click=True,
    )
```

### Screenshot

<img width="1366" alt="image" src="https://user-images.githubusercontent.com/17582508/234595067-eb7acc07-a207-4562-bd68-7b3b7f1ca48a.png">

### Logs

```shell
Traceback (most recent call last):
  File "/home/hpds/.conda/envs/dl_apps/lib/python3.9/site-packages/gradio/routes.py", line 401, in run_predict
    output = await app.get_blocks().process_api(
  File "/home/hpds/.conda/envs/dl_apps/lib/python3.9/site-packages/gradio/blocks.py", line 1302, in process_api
    result = await self.call_function(
  File "/home/hpds/.conda/envs/dl_apps/lib/python3.9/site-packages/gradio/blocks.py", line 1025, in call_function
    prediction = await anyio.to_thread.run_sync(
[...]

PR #4258, opened 2023-05-17
Fix run on click

abidlabs:
Quick fix for issue #3980 using the handy `.then()`

Couldn't think of a way to test behavior not implementation, so deferring that until we can do proper e2e tests. For now, you can test with a script like this:

```py
import gradio as gr

def mirror(x):
    return x

with gr.Blocks() as demo:
    with gr.Row():
        input_img = gr.Image(label="Input image").style(height=600)
        output_img = gr.Image(label="Output image").style(height=600)

    gr.Examples(
        examples=["lion.jpg"],
        inputs=[input_img],
        outputs=output_img,
        fn=mirror,
        run_on_click=True,
    )
    
demo.launch()
```

Fixes: #3980


abidlabs:
Thanks for the quick review @aliabid94!

PR #4262, opened 2023-05-18
Dropdown improvement

dongweiming:
# Description

Enter a comma-separated string, and the options will be automatically recognized when press Enter

# Checklist:


abidlabs:
Nice! Thanks for the PR @dongweiming, we'll take a look

abidlabs:
Hi @dongweiming sorry for the late review. I don't think this PR is working correctly. Try running:

```py
import gradio as gr

with gr.Blocks() as demo:
    gr.Dropdown(["abc", "bcd", "cde", "def"], multiselect=True)
demo.launch()
```

It doesn't work for me:

Recording 2023-06-08 at 16 08 49

Furthermore, will this work if the choices themselves contain commas?

As a general rule, we only allow PRs that target existing issues that have been greenlit by the maintainers of the repo (except for small issues). Especially right now, as part of our preparation for releasing Gradio 4.0 (which will involve significant changes to the codebase), I think it's best if we close this PR for now. Feel free to open an issue and we can indicate when is a better time to tackle this. Thanks again!

Issue #4266, opened 2023-05-18
Latex Support in Chatbot


dawoodkhan82:
### Describe the bug

https://github.com/gradio-app/gradio/pull/4150#issuecomment-1550214131

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

https://github.com/gradio-app/gradio/pull/4150#issuecomment-1550214131

### Logs

```shell
N/A
```

### System Info

```shell
N/A
```

### Severity

annoying

abidlabs:
Related: #4244, #4075

Issue #4244, opened 2023-05-17
The chatbot interface does not like certain text that includes MathJax/LaTeX.


taesiri:
### Describe the bug

Gradio's Chatbot interface crashes when *some* MathJax/LaTeX commands, such as `$\\texttt{sometext}$`, are included in the bot's messages. This is not the bug with Gradio itself but with the underlying libraries to parse Markdown.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```python
import gradio as gr
import random
import time

def main():
    with gr.Blocks() as demo:
        chatbot = gr.Chatbot()
        msg = gr.Textbox()
        clear = gr.Button("Clear")

        def respond(message, chat_history):
            bot_message = random.choice(["$\\texttt{text}$"])  # LaTeX command that causes the issue
            chat_history.append((message, bot_message))
            time.sleep(1)
            return "", chat_history

        msg.submit(respond, [msg, chatbot], [msg, chatbot])
        clear.click(lambda: None, None, chatbot, queue=False)

    demo.launch()

if __name__ == "__main__":
    main()

```

### Screenshot

<img width="1360" alt="Screenshot 2023-05-16 at 11 09 10 PM" src="https://github.com/gradio-app/gradio/assets/588431/4b062423-5cbc-4a80-892d-c2fc064a2a00">

### Logs

```shell
Traceback (most recent call last):
  File "/home/taesiri/miniconda3/envs/torch2/lib/python3.10/site-packages/gradio/routes.py", line 414, in run_predict
    output = await app.get_blocks().process_api(
  File "/home/taesiri/miniconda3/envs/torch2/lib/python3.10/site-packages/gradio/blocks.py", line 1323, in process_api
    data = self.postprocess_data(fn_index, result["prediction"], state)
  File "/home/taesiri/miniconda3/envs/torch2/lib/python3.10/site-packages/gradio/blocks.py", line 1257, in postprocess_data
    prediction_value = block.postprocess(prediction_value)
  File "/home/taesiri/miniconda3/envs/torch2/lib/python3.10/site-packages/gradio/components.py", line 4638, in postprocess
    self._postprocess_chat_messages(message_pair[1]),
  File "/home/taesiri/miniconda3/envs/torch2/lib/python3.10/site-packages/gradio/components.py", line 4608, in _postprocess_chat_messages
    chat_message = cast(str, self.md.render(chat_message))
[...]

Issue #4075, opened 2023-05-05
Chatbot latex invisible in dark theme, and add two '\n' automatically


LZY-the-boys:
### Describe the bug

Though in #3229 has metioned this problem, I still suffer from it when generating in chatbot 

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```
import sys
import gradio as gr
import argparse
import warnings
import time
import os
import json
def save():
    pass

def evaluate(
    inputs,
    extra,
    history,
    temperature=0.1,
    top_p=0.75,
    top_k=40,
    num_beams=4,
    max_new_tokens=128,
    min_new_tokens=1,
    repetition_penalty=2.0,
    max_memory=1024,
    do_sample=False,
    prompt_type='0',
    **kwargs,
):
    print(gr.__version__)
    history = [] if history is None else history
    return_text = [(item['input'], item['output']) for item in history]
    show_text = ''
    for i in range(10):
        show_text += "$" + str(i) +'$'
        yield return_text +[(inputs, show_text)], history
    print('[EOS]', end='\n')
    output = show_text
    history.append({
        'input': inputs,
        'output': output,
    })           
    return_text += [(inputs, show_text)]
    yield return_text, history


with gr.Blocks() as demo:
    fn = evaluate
    title = gr.Markdown("Start with $\frac{\frac{x+1}{x+2}}{x+3}$ $\frac{x+1}{x+2}$then we get $ 2+x $ and $3$")
    history = gr.components.State()
    with gr.Row().style(equal_height=False):
        with gr.Column(variant="panel"):
            input_component_column = gr.Column()
            with input_component_column:
                input = gr.components.Textbox(
                    lines=2, label="Input", placeholder="请输入问题."
                )
                extra = gr.components.Textbox(
                    lines=2, label="Extra", placeholder="context or persona"
                )
                temperature = gr.components.Slider(minimum=0, maximum=1, value=1.0, label="Temperature")
                topp = gr.components.Slider(minimum=0, maximum=1, value=0.9, label="Top p")
                topk = gr.components.Slider(minimum=0, maximum=100, step=1, value=60, label="Top k")
                beam_number = gr.components.Slider(minimum=1, maximum=10, step=1, value=4, label="Beams Number")
                max_new_token = gr.components.Slider(
                    minimum=1, maximum=2000, step=1, value=256, label="Max New Tokens"
                )
                min_new_token = gr.components.Slider(
[...]

PR #4285, opened 2023-05-19
Katex support in chatbot

dawoodkhan82:
# Description
Addes L
Screenshot 2023-05-19 at 12 26 52 PM
aTex support using `Katex` for `gr.Chatbot` in frontend.

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

Closes: #4266 
Closes: #4244
Closes: #4075

# Checklist:

# A note about the CHANGELOG


js/chatbot/src/ChatBot.svelte
```diff
@@ -91,6 +95,16 @@
 			node.style.position = "relative";
 			node.appendChild(copy_div);
 		});
+
+		renderMathInElement(document.body, {
```
abidlabs, line 99: I don't think we want to render Math everywhere in `document.body`, just in the Chatbot element itself, correct? (Could be misunderstanding what this element does)

js/chatbot/src/ChatBot.svelte
```diff
@@ -91,6 +95,16 @@
...
+			delimiters: [
+				{ left: "$$", right: "$$", display: true },
+				{ left: "$", right: "$", display: false },
+				{ left: "\\(", right: "\\)", display: false },
+				{ left: "\\[", right: "\\]", display: true }
```
abidlabs, line 104: I think we should remove these two ways of defining LaTeX to avoid ambiguity. In practice, the $ sign approach is the most common way to define LaTeX and we should just stick to that
```suggestion
```

abidlabs:
Works great @dawoodkhan82 on both streaming and the regular use case! This should also close #4244 and #4075 so I've added that to the parent comment, and pointed out a few nits. 

The only other nit is that the size of the LaTeX is quite a bit bigger than the neighboring font. Maybe we can reduce it just a little bit?

image

js/chatbot/src/ChatBot.svelte
pngwn: Yeah. We could limit this to the container of the chat bot. Any parent element in this file should be fine in place of `document.body`

js/chatbot/src/ChatBot.svelte
pngwn: Sounds good. Will probably prevent sneaky bugs/ syntax collisions down the road. We just need to document it.

js/chatbot/src/ChatBot.svelte
```diff
@@ -2,6 +2,9 @@
 	import { marked } from "marked";
 	import Prism from "prismjs";
 	import "prismjs/components/prism-python";
+	import "prismjs/components/prism-latex";
+	import "katex/dist/katex.min.css";
+	import renderMathInElement from "katex/dist/contrib/auto-render.js";
```
pngwn, line 7: Can we change this to snake case.

pngwn (commented):
Looks good, just left a small comment and I agree with @abidlabs suggestions.

abidlabs:
Beautiful! 

image

LGTM @dawoodkhan82

abidlabs (approved):

abidlabs:
I'll go ahead and merge it in so that we can do a release

PR #4295, opened 2023-05-21
Queue monitor UI

aliabid94:
I was working on fixing the queue for not dequeuing dropped connections, and needed a UI to watch the queue (which is something folks have requests too). Built this queue watcher first. Got a little carried away with features, would be good to get some feedback.

Recording 2023-05-23 at 06 01 31


gradio/activity_log.py
```diff
@@ -0,0 +1,120 @@
...
+    request_breakdown: Union[RequestBreakdown, None]
+    requests_per_fn: list[FunctionStats] = []
+    event_count_per_stage: list[int] = [0, 0, 0, 0]
+    queue_preview: Optional[list[list[Task]]] = None
+    active_workers: int = 0
+
+class ActivityLog:
+    def __init__(self, dependencies: list[dict], queue: Queue | None, **kwargs):
+        self.activity = Activity(active_workers=queue.max_thread_count if queue else 0)
+        self.activity.request_breakdown = {}
```
akx, line 45:
```suggestion
        self.activity.request_breakdown = collections.Counter()
```

gradio/activity_log.py
```diff
@@ -0,0 +1,120 @@
...
+        self.activity.queue_preview = [[], [], [], []]
+        self.queue = queue
+        self.completed_tasks = deque(maxlen=PREVIEW_SIZE)
+        self.lock = threading.Lock()
+
+    def _request_update(self, fn_index: int, status: StatusType, diff: int):
+        with self.lock:
+            self.activity.request_breakdown[status] = (
+                self.activity.request_breakdown.get(status, 0) + diff
+            )
```
akx, line 61: With a `Counter`:
```suggestion
            self.activity.request_breakdown[status] += diff
```

gradio/activity_log.py
```diff
@@ -0,0 +1,120 @@
...
+    status: Optional[StatusType]
+
+
+RequestBreakdown = Dict[StatusType, int]
+
+
+class FunctionStats(BaseModel):
+    fn: Union[str, int]
+    duration: float
+    request_breakdown: RequestBreakdown
```
akx, line 31:
```suggestion
    request_breakdown: collections.Counter
```

gradio/activity_log.py
```diff
@@ -0,0 +1,120 @@
...
+    queue_preview: Optional[list[list[Task]]] = None
+    active_workers: int = 0
+
+class ActivityLog:
+    def __init__(self, dependencies: list[dict], queue: Queue | None, **kwargs):
+        self.activity = Activity(active_workers=queue.max_thread_count if queue else 0)
+        self.activity.request_breakdown = {}
+        self.fn_names = [dep.get("api_name") or i for i, dep in enumerate(dependencies)]
+        self.activity.requests_per_fn = [
+            FunctionStats(fn=fn_name, duration=0, request_breakdown={})
```
akx, line 48:
```suggestion
            FunctionStats(fn=fn_name, duration=0, request_breakdown=collections.Counter())
```

gradio/activity_log.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def _request_update(self, fn_index: int, status: StatusType, diff: int):
+        with self.lock:
+            self.activity.request_breakdown[status] = (
+                self.activity.request_breakdown.get(status, 0) + diff
+            )
+            self.activity.requests_per_fn[fn_index].request_breakdown[status] = (
+                self.activity.requests_per_fn[fn_index].request_breakdown.get(status, 0)
+                + diff
+            )
```
akx, line 65: and
```suggestion
            self.activity.requests_per_fn[fn_index].request_breakdown[status] += diff
```

gradio/activity_log.py
```diff
@@ -0,0 +1,120 @@
...
+        with self.lock:
+            self.activity.sessions += 1
+
+    def update_request(
+        self,
+        fn_index: int = 0,
+        old_state: str | None = None,
+        new_state: str | None = None,
+        avg_duration: float | None = None,
+    ):
```
akx, line 77:
```suggestion
    ) -> None:
```

gradio/activity_log.py
```diff
@@ -0,0 +1,120 @@
...
+                queued.append(task)
+            self.activity.queue_preview[0] = queued
+            pending = []
+            self.activity.event_count_per_stage[1] = sum(
+                0 if job is None else 1 for job in self.queue.active_jobs
+            )
+            for evts in self.queue.active_jobs:
+                if len(pending) >= PREVIEW_SIZE:
+                    break
+                if evts is None or len(evts) == 0:
```
akx, line 100:
```suggestion
                if not evts:
```

gradio/queueing.py
```diff
@@ -32,6 +35,8 @@ def __init__(
         self.token: str | None = None
         self.progress: Progress | None = None
         self.progress_pending: bool = False
+        self.idx = Event.idx
+        Event.idx += 1
```
akx, line 39: This doesn't look thread-safe.

js/app/src/Blocks.svelte
```diff
@@ -77,17 +80,27 @@
 	});
 
 	let params = new URLSearchParams(window.location.search);
-	let api_docs_visible = params.get("view") === "api";
-	const set_api_docs_visible = (visible: boolean) => {
-		api_docs_visible = visible;
+	let visible_side_panel: null | "api" | "traffic" = params.get("view") as
```
akx, line 83: The repeated `null | "api" | "traffic"` could be lifted into a `type SidePanelType = null | "api" | "traffic"`.

akx, line 83: That's an unsafe `as` cast since nothing is actually _validating_ `params.get("view")` and it's user input.

js/app/src/Blocks.svelte
```diff
@@ -77,17 +80,27 @@
...
+		| "traffic"
+		| null;
+	const set_side_panel = (panel: null | "api" | "traffic") => {
+		visible_side_panel = panel;
 		let params = new URLSearchParams(window.location.search);
-		if (visible) {
+		if (panel === "api") {
 			params.set("view", "api");
+		} else if (panel === "traffic") {
+			params.set("view", "traffic");
```
akx, line 93: How about
```suggestion
		if (panel === "api" || panel === "traffic") {
			params.set("view", panel);
```

js/app/src/components/Row/Row.svelte
```diff
@@ -2,7 +2,7 @@
 	import type { Styles } from "@gradio/utils";
 
 	export let style: Styles = {};
-	export let elem_id: string;
+	export let elem_id: string | null = null;
```
akx, line 5: Why is the option `null` here when it's `undefined` in other files?

js/app/src/traffic_stats/Queue.svelte
```diff
@@ -0,0 +1,116 @@
...
+	let flattened_queue: Array<[Task, number, number]>;
+	$: {
+		flattened_queue = [];
+		for (let i = 0; i < queue_preview.length; i++) {
+			for (let j = 0; j < queue_preview[i].length; j++) {
+				flattened_queue.push([queue_preview[i][j], i, j]);
+			}
+		}
+		flattened_queue = flattened_queue;
+		console.log(flattened_queue);
```
akx, line 21: Stray `console.log`?

js/app/src/traffic_stats/Queue.svelte
```diff
@@ -0,0 +1,116 @@
...
+					? task.status
+					: i == 0
+					? "queued"
+					: i == 1
+					? "pending"
+					: i == 2
+					? "success"
+					: i == 3
+					? "waiting_connection"
+					: undefined}
```
akx, line 54: Feels like this should be a mapping lookup instead of a ternary.

js/app/src/traffic_stats/Traffic.svelte
```diff
@@ -0,0 +1,51 @@
...
+			return;
+		}
+		pending_request = true;
+		fetch(root + "traffic")
+			.then((response) => response.json())
+			.then((data) => {
+				activity = data;
+				pending_request = false;
+			});
+	};
```
akx, line 29: Or
```suggestion
	async function get_traffic() {
		if (pending_request) {
			return;
		}
		try {
			pending_request = true;
			const resp = await fetch(root + "traffic");
			activity = await resp.json();
		} finally {
			pending_request = false;
		}
	};
```

akx (commented):
This is quite the hefty PR 😅

gradio/activity_log.py
```diff
@@ -0,0 +1,116 @@
...
+                    break
+                if not evts:
+                    continue
+                task = Task(
+                    fn=self.fn_names[evts[0].fn_index],
+                    idx=(
+                        evts[0].idx
+                        if len(evts) == 1
+                        else "|".join(str(evt.idx) for evt in evts)
+                    ),
```
akx, line 104: This simplifies to
```suggestion
                    idx="|".join(str(evt.idx) for evt in evts),
```

gradio/queueing.py
```diff
@@ -32,6 +37,9 @@ def __init__(
         self.token: str | None = None
         self.progress: Progress | None = None
         self.progress_pending: bool = False
+        with new_event_lock:
+            self.idx = Event.idx
+            Event.idx += 1
```
akx, line 42: This is still not thread-safe, it may be async-safe but not thread-safe.

I would consider a lockless system, e.g. `self.idx = uuid.uuid4()` or [an ULID](https://github.com/valohai/ulid2) if you need chronological ordering.

js/app/src/traffic_stats/Dashboard.svelte
```diff
@@ -0,0 +1,101 @@
...
+
+	const total_requests_fn = (breakdown: RequestBreakdown) => {
+		return Object.values(breakdown).reduce((a, b) => a + b, 0);
+	};
+	export let sessions: ActivityLog["sessions"];
+	export let request_breakdown: ActivityLog["request_breakdown"];
+	export let requests_per_fn: ActivityLog["requests_per_fn"];
+	export let event_count_per_stage: ActivityLog["event_count_per_stage"];
+	export let queue_preview: ActivityLog["queue_preview"] | undefined =
+		undefined;
```
akx, line 20: Undefined is implicit.
```suggestion
	export let queue_preview: ActivityLog["queue_preview"] | undefined;
```

js/app/src/traffic_stats/Dashboard.svelte
```diff
@@ -0,0 +1,101 @@
...
+				{#each requests_per_fn as { fn, duration, request_breakdown }}
+					{#if total_requests_fn(request_breakdown) > 0}
+						<div class="fn-breakdown">
+							<div class="fn">
+								<span>{fn}</span>
+								<span class="details"
+									>{total_requests_fn(request_breakdown)} requests &bull; {duration.toFixed(
+										1
+									)}s of processing per success</span
+								>
```
akx, line 58: Something like
```suggestion
								<span class="details">
									{total_requests_fn(request_breakdown)} requests
									&bull;
									{duration.toFixed(1)}s of processing per success
								</span>
```
would be more readable.

js/app/src/traffic_stats/Queue.svelte
```diff
@@ -0,0 +1,111 @@
...
+    let active_workers_array = Array.from({ length: active_workers });
+    let flattened_queue: Array<[Task, number, number]>;
+    $: {
+        flattened_queue = [];
+        for (let i = 0; i < queue_preview.length; i++) {
+            for (let j = 0; j < queue_preview[i].length; j++) {
+                flattened_queue.push([queue_preview[i][j], i, j]);
+            }
+        }
+        flattened_queue = flattened_queue;
```
akx, line 20: 😑

js/app/src/traffic_stats/Queue.svelte
```diff
@@ -0,0 +1,111 @@
...
+    {#each flattened_queue as [task, i, j] (task.idx)}
+        <div
+            class="task-holder"
+            style:top="{i * SPACING}px"
+            style:left="{j * SPACING}px"
+            animate:flip
+        >
+            <TaskCard
+                {task}
+                status={task.status ? task.status : STATUS_PER_STAGE[i]}
```
akx, line 49: Use the nullish coalescing operator, Luke
```suggestion
                status={task.status ?? STATUS_PER_STAGE[i]}
```

aliabid94:
Thanks for the great review @akx! I incorporated all your suggestions except for `collections.Counter`, because pydantic doesn't support Counters, and also it's not possible to validate the keys of the object are of type RequestStatus.

js/app/src/traffic_stats/Queue.svelte
aliabid94: actually needed to cause Svelte to re-render!

gradio/queueing.py
aliabid94: Why isn't it thread safe? (I'm using a threading lock now.)
I did want the idx to be a counter so in the traffic stats sidepanel, every task shows which event number they are. I think this would be helpful for watching the queue status.

pngwn:
Do we really want this in Gradio directly? Seems pretty Niche to include in the main app.

abidlabs:
I agree with @pngwn here that this is quite niche and increase surface area for maintenance. Let's keep it as something useful internally to solve queue-related issues, but not release

PR #4303, opened 2023-05-22
Don't allow dotfiles for /file= route

akx:
# Description

* relevant motivation: in general, filenames starting with `.` are considered hidden; I don't think Gradio ought to expose them. The same stands for files in hidden-esque directories.

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,7 @@
 
 ## Breaking Changes:
 
-No changes to highlight.
+- The `/file=` route no longer allows accessing dotfiles or files in "dot directories" by [@akx](https://github.com/akx) in [PR 4303](https://github.com/gradio-app/gradio/pull/4303)
 
 # 3.32.0
 
--- a/gradio/routes.py
+++ b/gradio/routes.py
@@ -326,7 +326,7 @@ class App(FastAPI):
                 utils.is_in_or_equal(abs_path, blocked_path)
                 for blocked_path in blocks.blocked_paths
             )
-            if in_blocklist:
+            if in_blocklist or any(part.startswith(".") for part in abs_path.parts):
                 raise HTTPException(403, f"File not allowed: {path_or_url}.")
 
             in_app_dir = utils.abspath(app.cwd) in abs_path.parents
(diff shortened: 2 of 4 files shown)


abidlabs:
Thanks @akx. This seems like a good change to me, but two things:
* Can we mark this as a breaking change in the changelog?
* This is a good change to the default behavior, but it should be override-able with the `allowed_paths` parameter. Right now, this takes precedence over the `allowed_lists` parameter, can we fix this?

akx:
> * Can we mark this as a breaking change in the changelog?

Sure.

> * This is a good change to the default behavior, but it should be override-able with the `allowed_paths` parameter. Right now, this takes precedence over the `allowed_paths` parameter, can we fix this?

I'm not sure that's a good idea, because there's no way of then _not_ exposing dotfiles in the `allowed_paths`. Can you think of an usecase where you'd need to serve a dotfile?

abidlabs:
I looked into this some more @akx as part of reviewing our security procedures around the `/file` route. We're going to be fixing some access issues, but changing our policies around dotfiles should not be one of them. 

Dotfiles "are not a security mechanism because access is not restricted – usually the intent is simply to not "clutter" the display of the contents of a directory listing with files the user did not directly create." [[1](https://dbpedia.org/page/Hidden_file_and_hidden_directory)] so the `/file` route should not behave differently with regards to dotfiles. Thanks for creating this PR @akx but I'll go ahead and close this.

abidlabs:
Reopened after discussion with @akx

abidlabs:
LGTM @akx. Will just update the security section in the guides to reflect this change

abidlabs (approved):

torridgristle:
This has broken everything that I run from Google Drive with shortcuts shared between drives since they appear as "/content/drive/.shortcut-targets-by-id/". I've tried placing this and various other paths before and after it in the allowed paths. Still getting 403s and "file not allowed".

I will simply not have any secret files in a public link, I don't need this "feature".

cdjm:
Can we please make the dot security feature configurable so we can disable it?

This breaks guided installers of Stable Diffusion WebUI 1.6.0.
For example Stability Matrix is installed as AppImage package and saves its content in ~/.config directory.

I hope that you can consider a review of this change.

abidlabs:
Hi @cdjm yes I think we'll make it possible to override this using the existing `allowed_paths` parameter (e.g. if you pass paths with dotfiles into the `allowed_paths` parameter, then it will allow you to serve files from those directories) . We'll try to get to this soon -- in the meantime, happy to review any PRs.

abidlabs:
Makes sense to me @wfjsw.

Issue #4332, opened 2023-05-24
Gradio remains stuck in "Loading" phase in LAN configuration (not connected to the Internet)


musiscient:
### Describe the bug

Gradio remains stuck in "Loading" phase in LAN configuration (not connected to the Internet) if I clean all images and cached files from my different test browsers (Chrome, Firefox, Opera, Brave), even with localhost:7860,

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

# I'm running the basic script of the "(Getting Started) - Quick Start" of Gradio :
#
import gradio as gr
print(gr.__version__)

def greet(name):
    return "Hello " + name + "!"

demo = gr.Interface(
    fn=greet,
    inputs=gr.Textbox(lines=2, placeholder="Name Here..."),
    outputs="text",
)
demo.launch(debug=True) 

-----
- It works perfectly as long as the computer is connected to the internet.

- I disconnect from the internet to connect to a LAN, or disconnect completely.

Gradio is then still accessible by Chrome or other browsers.

But : 
If I clean up the cached images and files of my different browsers, 
and, as long as I am not connected to the internet, either because I am connected to a LAN not connected to the internet, or because the computer is completely disconnected from all external links, (RJ45 and Wifi disconnected) so, even in localhost:7860, the Gradio web page remains blocked in the loading phase. Impossible to reach the content of the Gradio web page.

I tested this, because I installed Automatic1111 on Ubuntu Server 22.04 to run inside a LAN outside the internet, (with argument --listen) and the same phenomenon occurs. Once the images and cached files of the browsers have been cleaned up following an internet connection to connect in LAN configuration, access to Gradio pages remains blocked in the "loading" situation. Impossible to access the Gradio interface pages.

Thank you for any indications to allow the use and access of Gradio via a LAN not connected to the Internet.

Cheers

-----

### Screenshot

![Capture d’écran du 2023-05-24 18-22-50](https://github.com/gradio-app/gradio/assets/17192687/2533d942-83bf-4ee2-8580-aa8d7a31ba0e)

### Logs

```shell
# from terminal :
/home/ia/anaconda3/envs/Gradio_studies/bin/python /home/ia/anaconda3/envs/Gradio_studies/ia_python_code/greet.py 
3.32.0
Running on local URL:  http://127.0.0.1:7860
```

### System Info

```shell
Gradio = 3.32.0
OS (es) :
- Ubuntu 20.04
[...]

PR #4335, opened 2023-05-25
Wait for up to 1.5 seconds for external CSS to load

akx:
# Description

* relevant motivation: #4332 – if Google Fonts doesn't respond (e.g. in a LAN), the UI hangs.
* a summary of the change: wait for up to 1.5 seconds for external CSS to load before continuing anyway.
* which issue is ~fixed~ touched upon: #4332

# Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,8 @@ No changes to highlight.
 
 ## Bug Fixes:
 
-No changes to highlight.
+- Gradio will no longer hang forever when waiting for external CSS to load by [@akx](https://github.com/akx) in [PR 4335](https://github.com/gradio-app/gradio/pull/4335)
+
 
 ## Other Changes:
 
--- a/js/app/src/Index.svelte
+++ b/js/app/src/Index.svelte
@@ -100,7 +100,7 @@
 			style.innerHTML = css_string;
 			target.appendChild(style);
 		}
-		await mount_css(config.root + "/theme.css", document.head);
+		await mount_css(config.root + "/theme.css");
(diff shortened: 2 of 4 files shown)


pngwn:
This isn't the right solution for offline loading and creates other possible issues on slow connections. We'll solve this properly in the near future.

pngwn:
Unfortunately it would also create problems that don't currently exist. So it isn't a viable approach.

akx:
@pngwn I'm curious – can you elaborate? What external CSS is critical enough to warrant waiting for? 🤔

pngwn:
If we don't load the CSS then the UI is basically unusable.While fonts aren't critical, other CSS files are, and this function is used to load many different kinds of CSS, not just fonts.

This change would potentially give users on slow connections a broken UI which isn't acceptable since Gradio is primarily a UI tool. While this does solve the offline usecase, that is a very niche usecase, whereas slow connections happen for a variety of reasons from poor infrastructure to poor mobile signal (and many reason in between). Creating a worse experience for those users in order to enable offline usage of gradio (which has been requested a few times but only by a handful of users), is not a good trade-off for the library.

We will fix this soon but not at the expense of others users' experience.

akx:
> If we don't load the CSS then the UI is basically unusable.While fonts aren't critical, other CSS files are, and this function is used to load many different kinds of CSS, not just fonts.

Right, which is why this PR only adds the timeout to CSS that is loaded from an absolute source -- though that could also be changed to something like `url.includes('googlefonts') ? 1500 : undefined` :)

PR #4336, opened 2023-05-25
Load iframe resizer asynchronously

akx:
# Description

Please include: 
* relevant motivation: when there's no internet connection, it'll take a long while for a browser to figure out that it can't load iframeResizer.contentWindow.min.js from Cloudflare. 
* a summary of the change: make the script `async`. This ought to be safe, as there's a specific check for that. https://github.com/davidjbradshaw/iframe-resizer/blob/1ab689163f9e2505779b5f200b4f28adbddfc165/src/iframeResizer.contentWindow.js#L1292-L1302
* which issue is ~fixed~ touched upon: #4332

# Checklist:


pngwn (approved):
Thanks @akx!

pngwn:
This is causing the browser tests to fail and I'm not entirely sure why. Gradio itself works fine. Will take a look later today.

pngwn:
We've changed how we do browser tests around a bit which seems to have resolved everything here.

Sorry for the delay and thanks again @akx!

PR #4349, opened 2023-05-28
Remove statustracker flicker on quick requests

aliabid94:
Added a small opacity transition to remove the flicker on very quick requests. 

Test by running any of the quick demos, e.g. `demo/calculator`


abidlabs:
Creates potential for confusion imo -- if the output doesn't change for a fast demo, you have no visual indication that the request actually went through

pngwn:
The fact that it is a request is an implementation detail that the end user won't necessarily know about. For certain types of actions, i think there is the expectation that the UI will update in 'realtime' and no loading spinner is necessary. The fact that it requires a network request is really specific to how gradio works.

abidlabs:
Ok I don’t feel too strongly about this (I also think it’s very rare that a Gradio endpoint runs in <0.1s) so if you think it’s a good change, it’s fine for me

pngwn:
Also don't feel strongly, but I think it probably makes things nicer in a small handful of cases. Especially when working locally.

pngwn (approved):
If we are okay with adding this then lgtm.

Issue #4183, opened 2023-05-12
Interactive value not respected for interface output component


freddyaboulton:
### Describe the bug

Setting an output textbox as interactive does not make it interactive

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```
import gradio as gr

def greet(name):
    return "Hello " + name + "!"

demo = gr.Interface(fn=greet, inputs="text", outputs=gr.Textbox(interactive=True))
demo.launch()
```

You can work around this by setting `interactive=True` from a `gr.update`

### Screenshot

![interactive_textbox](https://github.com/gradio-app/gradio/assets/41651716/00307ab0-3e95-4270-9069-a75d5753084a)

### Logs

```shell
-
```

### System Info

```shell
3.29.0
```

### Severity

serious, but I can work around it

abidlabs:
This is expected behavior as we force all outputs to be non-interactive. Are there use cases where this is not desired?

freddyaboulton:
Someone asked on discord about this. 

I think it’s fine if gr.Textbox() is none interactive but feels weird that gr.Textbox(interactive=True) is ignored.

there’s a workaround so this is low priority for me.

yvrjsharma:
I'd feel the same. After setting interactive=True, one would expect an element to act that way.

pngwn:
Yeah this should be the behaviour. non-inputs are `interactive=False` by default but we should always be respecting explicit kwargs.

abidlabs:
Sounds good, we have a nice use case for this here: https://github.com/gradio-app/gradio/issues/4327

PR open: https://github.com/gradio-app/gradio/pull/4356

PR #4356, opened 2023-05-29
Respect `interactive=True` in output of `gr.Interface`

abidlabs:
You can now do:

```py
import gradio as gr

output_textbox = gr.Textbox(interactive=True)
gr.Interface(lambda x: x, "textbox", output_textbox).launch()
```

and the output Textbox will be interactive. Closes: #4183


freddyaboulton (approved):
Thanks for the fix and test @abidlabs !

abidlabs:
Thanks for reviewing @freddyaboulton

gradio/interface.py
```diff
@@ -264,8 +264,8 @@ def __init__(
         ]:
             for o in self.output_components:
                 assert isinstance(o, IOComponent)
-                o.interactive = False  # Force output components to be non-interactive
-
+                if o.interactive is None:
+                    o.interactive = False
```
akx, line 268: I'd say the comment was valuable.
```suggestion
                if o.interactive is None:
                    # Unless explicitly otherwise specified, 
                    # force output components to be non-interactive
                    o.interactive = False
```

Issue #4316, opened 2023-05-22
Sanitize feature of Marked breaking HTML additions to markdown on gr.Chatbot()


decent-engineer-decent-datascientist:
### Describe the bug

The new improved markdown support for gr.Chatbot() is wonderful, but enabling the "sanitize" option on Marked has broken existing implementations that leveraged HTML to enhance chatbot capabilities.

For example, we have a deployed QA+Search chatbot that returned answers and relevant sources like so (using gradio 3.29.0):
![image](https://github.com/gradio-app/gradio/assets/77806775/28be2016-a8c2-447b-93be-bff3b9ddcdc6)

However, using the latest version, this is what is produced. 
<img width="993" alt="Screenshot 2023-05-22 at 12 45 40 PM" src="https://github.com/gradio-app/gradio/assets/77806775/d40de9dd-c487-4980-99fa-9c33bfa058a0">

We're assuming the "sanitize" option in marked (this line https://github.com/gradio-app/gradio/blob/38607c9967ba6421a9b53c1fd9c94c29e4338365/js/chatbot/src/ChatBot.svelte#L42) is causing this "issue".

Could we get the ability to disable this sanitization? I understand wanting it on by default.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Can write one up if needed, but seem straightforward enough.

### Logs

```shell
N/A
```

### System Info

```shell
Gradio 3.32.0
```

### Severity

annoying

abidlabs:
For now, I actually think we should just set `sanitize=False` to prevent any breaking changes. WDYT @dawoodkhan82?

decent-engineer-decent-datascientist:
Personally would very much like to see this disabled, if only so we could catch up to the latest gradio in our deployment. Another thing to note is that Marked contributors themselves discourage the use/existence of sanitize:
https://github.com/markedjs/marked/discussions/1232

cowanAI:
still it looks like a mess 
<img width="328" alt="Screenshot 2023-05-28 at 9 45 34 AM" src="https://github.com/gradio-app/gradio/assets/128092052/977fadb6-1410-4ba5-8199-e4748af6326a">

abidlabs:
This is pretty high-priority, let's add to 3.x milestone @dawoodkhan82

dawoodkhan82:
@decent-engineer-decent-datascientist can you test #4360 to see if it fixes your issue.

lsqingfeng:
[...]

PR #4360, opened 2023-05-30
Use DOMPurify to sanitize html in chatbot

dawoodkhan82:
# Description
Replaces default `markedjs` html sanitizer with custom DOMPurify sanitizer for `gr.Chatbot()`

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

Closes: #4316

# Checklist:

# A note about the CHANGELOG


js/chatbot/src/ChatBot.svelte
```diff
@@ -39,7 +40,7 @@
 		gfm: true,
 		breaks: true,
 		pedantic: false,
-		sanitize: true,
+		sanitize: false,
 		smartLists: true,
 		smartypants: false
```
pngwn, line 45: We are getting some warnings about deprecations (which is slightly annoying because they are defaults and not being set by us) but lets explicitly set them to false here:

```suggestion
		smartypants: false,
		headerIds: false,
		mangle: false
```

pngwn: We are also getting a warning about the highlight method, so lets fix that here too. Instead of using `marked.setOptions({ highlight: ...})`. We need to add `marked-highlight` as a dependency of chabot and then do something like the following:

```js
	import { markedHighlight } from "marked-highlight";
  ...
  
  marked.use(
		markedHighlight({
			highlight(code, lang) {
				return Prism.highlight(code, Prism.languages[lang], lang);
			}
		})
	);
```

I tested this and it seems to work fine.

pngwn (changes requested):
The failing test is only happening because the unit tests are run in node and DOMPurifier is DOM only. We can fix this if we upgrade `happy-dom`.

From root: `pnpm i happy-dom@latest -w`

This and the other comments should be everything. Works perfectly though!

abidlabs:
In case anyone would like to try this out, you can install the wheel with 

`pip install https://gradio-builds.s3.amazonaws.com/db234a0f26a2720b1764be97e3cc4bfa70795275/gradio-3.32.0-py3-none-any.whl`

js/chatbot/src/ChatBot.svelte
pngwn: Actually I need to make some changes to the other chatbot PR. I'll do this there and we can merge them in here.

Keldos-Li:
This seems to work well in our repository. Great work!

pngwn (approved):

pngwn:
This will go out in the next release, sometime today.

decent-engineer-decent-datascientist:
@pngwn This is working for me too! Great work; any idea when this'll be on pypi?

pngwn:
Should be today sometime.

abidlabs:
`gradio==3.33.0` is out @decent-engineer-decent-datascientist :)

PR #4362, opened 2023-05-30
Remove old versioned docs and fix repo classification 

aliabd:
As discussed it would be better for visibility if gradio's repo was classified as Python instead of HTML (would show up in more relevant places in trending, recommendations etc). To do this I removed the old versioned docs and kept only the most recent one. When new docs are created on a new release, the old versioned docs will automatically get removed in the same PR. But since the docs file is so large, it still showed up as majority HTML with just one versioned docs (most recent version), so I created a gitattributes file that tells [linguist](https://github.com/github-linguist/linguist) to ignore /website in its language classification. (I still opted to remove the old versioned docs since it doesn't seem like we'll ever use them) 

Could take a while for the language calc to update (its a low priority action that runs in the background) but you can see it worked on this [fork](https://github.com/aliabd/gradio).


freddyaboulton (commented):
🤯 
image

nice @aliabd ! What if we store the versions in an s3 bucket or HF dataset as opposed to hosting the latest version in the repo?

abidlabs:
LGTM @aliabd, thanks for making this change!

I think saving to HF dataset like @freddyaboulton suggested would be great to reduce the size of the git repo. Can also be done in a separate PR if we want to get this out

aliabd:
Thanks for the suggestion @freddyaboulton! I went ahead and did this. All versioned docs are now stored here: https://huggingface.co/datasets/gradio/docs and none are stored in the repo. When a new version is released, the gh action will upload to that dataset, and when the website is launched the versioned docs file is downloaded if it doesn't already exist. Also no more "New Version Docs" PRs

Because we're not storing any versioned docs in the repo, we actually no longer need to tell linguist to ignore /website. So I removed that, so it better represents the actual code in the repo.

freddyaboulton (approved):
Very nice @aliabd ! Thanks for the fix

abidlabs:
Merging so that we can use this before we do more releases! Thanks for the nice PR @aliabd

Issue #4342, opened 2023-05-27
Submit button not disabled when streaming (chatbot)


bent-verbiage:
### Describe the bug

Following the chatbot streaming method that is showcased in the chatbot_streaming_main on HuggingFace, the submit (and the textarea field) are not disabled while the chatbot is streaming. For longer bot responses this is causing issues. 

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

- Go to [https://huggingface.co/spaces/gradio/chatbot_streaming_main](https://huggingface.co/spaces/gradio/chatbot_streaming_main)
- Type a message, press enter
- Immediately type another message and press enter before the bot is done completing it's message

The input box (and submit event) are not blocked while the bot is streaming and gives unexpected behaviour. This is likely because the "user" function has returned and then releases the disabled fields before it moves on to the "bot" function.

`msg.submit(user, [msg, chatbot], [msg, chatbot], queue=False).then( bot, chatbot, chatbot)`

One solution would be to keep the disabled status during the "then" function, or otherwise make it possible to manually disable and enable fields in the gradio code.

### Logs

```shell
No errors are shown, it is resulting in unexpected UI behaviour.
```

### System Info

```shell
Gradio==3.32.0
```

### Severity

serious, but I can work around it

PR #4363, opened 2023-05-30
Make chatbot input fields non-interactive while bot is streaming

freddyaboulton:
# Description

Closes: #4342 

- Changed the demo to make the textbox not interactive while the chatbot is streaming. That way devs who build off our demo code will start off with the best practice.
- Updated the guide as well

# A note about the CHANGELOG


pngwn:
> .then events weren't triggered if they were already running. Don't think that's the correct behavior because the event would get lost and never trigger. Curious what other's thoughts on this are. Happy to back out if we like the current behavior.

I don't think we should change this now, the current implementation will ensure _any_ event can run if an event is still running which is something we decided not to do until 4.0. Isn't this basically some variation on #2436?

freddyaboulton:
Good point - will back out tat change until 4.0 @pngwn

abidlabs:
Agreed with @pngwn's suggestion

Also, if we update these demos, we should update the corresponding Guide in which they are described: https://gradio.app/creating-a-chatbot/, to say that we are now chaining 3 events

pngwn (approved):
lgtm, other than `queue queue` typo!

abidlabs:
@freddyaboulton I'm seeing some weird behavior with the `chatbot_streaming` demo: 

Recording 2023-05-31 at 12 40 31

I'm also seeing it in `main` so it's not related to this PR, but are you able to reproduce?

abidlabs:
I'll go ahead and merge in this PR but we should solve the problem above before releasing^

freddyaboulton:
I can't reproduce @abidlabs !

chatbot_streaming

PR #4368, opened 2023-05-30
Do not send HF token to other domains via `/proxy`

abidlabs:
As pointed out by @akx it's not secure to send the HF token to other domains via the `/proxy` route. This PR makes a change so that the HF token is only sent to HF domains

### DIFF
--- a/gradio/routes.py
+++ b/gradio/routes.py
@@ -145,6 +145,20 @@ class App(FastAPI):
         return self.blocks
 
     @staticmethod
+    def build_proxy_request(url_path):
+        url = httpx.URL(url_path)
+        is_hf_url = (
+            url.host.endswith(".huggingface.co")
+            or url.host.endswith(".hf.space")
+            or url.host.endswith(".hf.co")
+        )
+        headers = {}
+        if Context.hf_token is not None and is_hf_url:
+            headers["Authorization"] = f"Bearer {Context.hf_token}"
+        rp_req = client.build_request("GET", url, headers=headers)
+        return rp_req
+
+    @staticmethod
     def create_app(
         blocks: gradio.Blocks, app_kwargs: Dict[str, Any] | None = None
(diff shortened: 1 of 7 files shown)


gradio/routes.py
```diff
@@ -144,6 +144,20 @@ def get_blocks(self) -> gradio.Blocks:
             raise ValueError("No Blocks has been configured for this app.")
         return self.blocks
 
+    @staticmethod
```
akx, line 147: This could just as well not be a `@staticmethod` so it's easy to override in a subclass..?

akx: To have an application be able to customize a forwarded request without having to copy-paste as much code.

abidlabs: Hmm I'm having a hard time seeing this in practice -- nor have I ever seen the `App` class subclassed in any gradio app. If you have a concrete use case, feel free to share, otherwise I'll keep as is

gradio/routes.py
```diff
@@ -144,6 +144,20 @@ def get_blocks(self) -> gradio.Blocks:
...
+    def build_proxy_request(url_path):
+        url = httpx.URL(url_path)
+        is_hf_url = (
+            url.host.endswith(".huggingface.co")
+            or url.host.endswith(".hf.space")
```
freddyaboulton, line 152: Since we just need this to fetch files from upstream spaces, we just need `.hf.space`, right?

freddyaboulton (approved):

gradio/routes.py
abidlabs: I think it should be fine to limit to just `.hf.space` yes, I'll make the change

abidlabs:
Addressed the suggestion and added a warning to the docstring about providing HF tokens only when loading trusted Spaces. Will merge after CI is green

Issue #4274, opened 2023-05-19
`/favicon.ico` route returns `403`


catboxanon:
### Describe the bug

Gradio seems to currently return `403` for the [`/favicon.ico` route](https://github.com/gradio-app/gradio/blob/7145327058a8152045c7f14bda64832e2bc5913b/gradio/routes.py#L283-L289). I think it may have something to do with the usage of [`safe_join`](https://github.com/gradio-app/gradio/blob/7145327058a8152045c7f14bda64832e2bc5913b/gradio/routes.py#L274-L275). This also seems to happen regardless of Python venv location, or even when attempting to manually add the file location via `allowed_paths`.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

1.
```python
import gradio as gr

with gr.Blocks() as demo:
    pass

demo.launch()
```

2. Visit http://127.0.0.1:7860/favicon.ico
3. Page returns `403`, `{"detail":"Forbidden"}`

### Logs

```shell
N/A
```

### System Info

```shell
Gradio v3.31.0, Windows 11, Firefox
```

### Severity

annoying

freddyaboulton:
I cannot reproduce @catboxanon 

<img width="558" alt="image" src="https://github.com/gradio-app/gradio/assets/41651716/38a04aaf-8c62-4f63-84c1-7c341d4496b5">

abidlabs:
Can confirm the issue on Windows. I think the fix should look a bit different, since the favicon is obtained through a different route:

```py
        @app.get("/favicon.ico")
        async def favicon():
```

Anyways, we'll look into it, thanks for reporting

PR #4369, opened 2023-05-30
Favicon fix for Windows

abidlabs:
Quick fix so that the favicon path works on windows instead of returning a 404

Closes: #4274


gradio/routes.py
```diff
@@ -292,7 +292,7 @@ def build_resource(path: str):
         async def favicon():
             blocks = app.get_blocks()
             if blocks.favicon_path is None:
-                return static_resource("img/logo.svg")
+                return static_resource("favicon.svg")
```
freddyaboulton, line 295: Rather than adding a new favicon file - can we use pathlib here to construct the path correctly?

abidlabs: Yes let's do that, you're right

abidlabs:
Now properly fixed. The issue before was that we were converting all paths to be Windows-style, which meant that any "/" would get converted to "\\" which was seen as insecure by the `safe_join` method. Now, we limit ourselves to the Linux-style separator, as is the case with the original `safe_join` method:

https://github.com/pallets/werkzeug/blob/76b049dd45fd072fb62a54bccc0e8d513b03f4d8/src/werkzeug/security.py#L142

freddyaboulton (approved):
Thanks @abidlabs !

PR #4371, opened 2023-05-31
Prevent label change event from triggering itself

### LINKED ISSUE #4309: Label.change event re-triggers itself
### Describe the bug

The `Label.change` event fires when its value did not change

In fact the completion of handling the `change` event seems to re-trigger the `change` event

(I encountered this while trying to use a hidden Label as a workaround for `State` objects have no change event https://github.com/gradio-app/gradio/issues/2107)

Issue may be specific to the `Label` component, if I substitute a hidden `Textbox` the issue does not happen

### Reproduction

```python
import gradio as gr

def set_label(val):
    print("set_label")
    return f"> {val}"

def on_change(val):
    print(f"on_change: {val}")

with gr.Blocks() as demo:
    inp = gr.Textbox(label="Set label")
    label = gr.Label("", visible=False)

    inp.submit(set_label, [inp], [label])

    label.change(on_change, [label], None)

demo.launch(debug=True)
```

### Screenshot

The code above gets stuck in a loop

Screenshot 2023-05-22 at 16 56 26

it seems like the change event re-triggers itself

### Logs

```shell
n/a
```

### System Info

```shell
gradio 3.32.0
```

### Severity

annoying

freddyaboulton:
# Description

Fixes #4309

To verify fix run repro from issue. Only one message gets printed

```
❯ python label_change.py
Running on local URL:  http://127.0.0.1:7861

set_label
on_change: {'label': '> s'}
```

# Checklist:

# A note about the CHANGELOG


pngwn:
I'm not a fan of this cheap deep equal approach. Gonna see if I can find another way.

freddyaboulton:
Ok sounds good @pngwn - I used this approach because that's what I've seen done in other components (JSON, video).

pngwn:
@freddyaboulton Yeah, that's why I'd like to figure something cleaner and faster out really. It is spreading. Essentially anytime the 'value' is an object.

pngwn:
I pushed an alternative approach that I think is cleaner (an is definitely faster as it lets Svelte to the comparisons). I tested against the origina repro but could you double check it works as expected?

Essentially the issue is that new objects are only ever equal to themselves, so we keep triggering the dispatch. Svelte does compare values and bail if nothing has change but since objects in JS are only ever equal to themselves, if we create a new object with the same values svelte thinks the value has 'changed'.

What I've done here is deconstruct the object, pulling off the properties. We then use these values as a requirement for triggering change events. This gives svelte an easier time as it is able to compare the label string directly.

pngwn:
Good to merge if this still work properly + CI is green.

pngwn:
If this approach works fine, we should also update the other cases to use this approach (in a follow up PR).

freddyaboulton:
Thanks @pngwn !

Issue #2395, opened 2022-10-04
Allow `gr.update()` to accept a component's `.style()` parameters? 


fffiloni:
In a Blocks example, I need to update the gallery grid size when a slider value changes.

The grid size can be set like this : `gradio.Gallery().style(grid=2)`.

How can i update a value inside .style() with gr.update() ?

```python
nb_img = gr.Slider(1, 4, value = 1, step = 1, label = 'Number of Images')
gallery = gr.Gallery().style(grid=1)
nb_img.change(lambda n: gr.update( ?????????????? ), inputs=nb_img, outputs=gallery)
```

Does it work if i write:
```python
nb_img.change(lambda n: gr.update( ).style(grid=n), nb_img, gallery)
```

abidlabs:
Unfortunately this is not currently possible, but I agree this would be really nice to support. cc @pngwn @aliabid94  

From the backend, it would be pretty straightforward to allow the `.update()` parameter to accep the `style()` kwargs, but I don't know if the frontend would be able to parse it?

fffiloni:
the css rules involved here are :

```css
.grid-cols-2 {
    grid-template-columns: repeat(2, minmax(0, 1fr));
}
@media (min-width: 1024px)
.lg\:grid-cols-2 {
    grid-template-columns: repeat(2, minmax(0, 1fr));
}
@media (min-width: 768px)
.md\:grid-cols-2 {
    grid-template-columns: repeat(2, minmax(0, 1fr));
}
@media (min-width: 640px)
.sm\:grid-cols-2 {
    grid-template-columns: repeat(2, minmax(0, 1fr));
}
```

pngwn:
We'll probably leave this one for now as we are reworking styling + theming but this is an important feature.

I don't think we want update to accept style kwargs because there is no guarantee that they will always be unique across both methods but we can figure out some API for it.

fffiloni:
What we could do instead, is provide a slider to the gallery component UI to readjust the grid size dynamically.
Maybe this would be simpler ?

PR #4374, opened 2023-05-31
Remove style parameter

aliabid94:
Wanted to get this cleanup out of the way before starting 4.0 stuff.

- Fully deprecated `.style` paramater, moved arguments to constructor
- `.update` supports all arguments formerly in style
- `scale` and `min_width` can be applied directly to any component, rather than requiring a Column nesting that would break aligned widths and other issues.

Closes: #2395


pngwn:
We will need to merge this into the 4.0 branch as it is a breaking change.

pngwn:
Ah okay.

pngwn (approved):
I tested as much as I could and lgtm! But we should wait until @dawoodkhan82 has tested as well.

demo/blocks_layout/run.py
```diff
@@ -5,12 +5,20 @@
...
+        gr.Image(interactive=True, scale=2)
         gr.Image()
     with gr.Row():
         gr.Textbox(label="Text")
-        gr.Number(label="Count")
+        gr.Number(label="Count", scale=2)
         gr.Radio(choices=["One", "Two"])
+    with gr.Row():
+        gr.Button("500", scale=0, min_width=500)
+        gr.Button("A").style(full_width=False)
```
neggles, line 16: should this `.style()` still be here? 🤔

gradio/blocks.py
```diff
@@ -757,6 +759,11 @@ def from_config(
         """
         config = copy.deepcopy(config)
         components_config = config["components"]
+        for component_config in components_config:
+            # for backwards compatibility, extract style into props
+            if "style" in component_config["props"]:
+                component_config["props"].update(component_config["props"]["style"])
+                del component_config["props"]["style"]
```
abidlabs, line 766: Nice good catch

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@
 ## Other Changes:
 
 - Performance optimization in the frontend's Blocks code by [@akx](https://github.com/akx) in [PR 4334](https://github.com/gradio-app/gradio/pull/4334)
+- Deprecated `.style` parameter and moved arguments to constructor. Added support for update to all arguments initially in style. Added `scale` and `min_width` support to every Component. By [@aliabid94](https://github.com/aliabid94) in [PR 4374](https://github.com/gradio-app/gradio/pull/4374)
```
abidlabs, line 22:
```suggestion
- Deprecated `.style` parameter and moved arguments to constructor. Added support for `.update()` to all arguments initially in style. Added `scale` and `min_width` support to every Component. By [@aliabid94](https://github.com/aliabid94) in [PR 4374](https://github.com/gradio-app/gradio/pull/4374)
```

guides/03_building-with-blocks/02_controlling-layout.md
```diff
@@ -22,6 +22,20 @@ with gr.Blocks() as demo:
         btn2 = gr.Button("Button 2")
 ```
 
+The widths of elements in a Row can be controlled via `scale` and `min_width` arguments that are present in every Component.
```
abidlabs, line 25:
```suggestion
The widths of elements in a Row can be controlled via a combination of `scale` and `min_width` arguments that are present in every Component.
```

abidlabs:
Some issues I noticed:

* `show_copy_button` in `Textbox` not working:

```py
import gradio as gr

def test(x):
    return x

with gr.Blocks() as demo:
    a = gr.Textbox(show_copy_button=True)
    b = gr.Textbox(show_copy_button=True)
    a.change(test, a, b)
    
demo.launch()
```

* Setting `container=False` looks very ugly, is this expected?

```
import gradio as gr

def test(x):
    return x

with gr.Blocks() as demo:
    a = gr.Textbox(show_copy_button=True)
    b = gr.Textbox(container=False)
    a.change(test, a, b)
    
demo.launch()
```

image

* `height` in `Chatbot` not having any effect:

```py
import gradio as gr

def test(x):
    return x

with gr.Blocks() as demo:
    a = gr.Textbox(show_copy_button=True)
    b = gr.Chatbot(height=10)
    a.change(test, a, b)
    
demo.launch()
```

* `height` and `width` in `Image` not having any effect:

```py
import gradio as gr

def test(x):
    return "lion.jpg"

with gr.Blocks() as demo:
    a = gr.Textbox()
    b = gr.Image(height=10, width=200)
    a.change(test, a, b)
    
demo.launch()
```

pngwn:
Regarding #2395, do they actually work if you update them?

guides/03_building-with-blocks/02_controlling-layout.md
```diff
@@ -22,6 +22,20 @@ with gr.Blocks() as demo:
...
+
+```python
+with gr.Blocks() as demo:
+    with gr.Row():
+        btn0 = gr.Button("Button 0", scale=0)
+        btn1 = gr.Button("Button 1", scale=1)
+        btn2 = gr.Button("Button 2", scale=2)
+```
+
+- `min_width` will set the minimum width the element will take. The Row will wrap if there isn't sufficient space to satisfy all `min_width` values.
```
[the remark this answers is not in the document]
aliabid94: Hmm it's not the same - most Components have 160px min, columns still have 320px (not 340, they've always been 320), buttons have no minimum width.

aliabid94:
Fixed everything @abidlabs, can test with this:

```python
import gradio as gr

with gr.Blocks() as demo:
    a = gr.Textbox(show_copy_button=True)
    b = gr.Textbox(show_copy_button=True)
    img = gr.Image(height=200, width=200)
    img_size = 200
    grow_btn = gr.Button("Grow Image")
    def grow_image():
        global img_size
        img_size += 100
        return gr.Image.update(width=img_size, height=img_size)
    grow_btn.click(grow_image, None, img)
    chat = gr.Chatbot([["hello", "hi"], ["hi", "hello"]], height=240)
    a.change(test, a, b)
    
demo.launch()
```
> Regarding https://github.com/gradio-app/gradio/issues/2395, do they actually work if you update them?

Yes, can test using the demo above as well

abidlabs:
Tested this pretty extensively, both the deprecated & new versions work quite well! A couple of small issues:

* `gr.Textbox()` doesn't seem to support updating the style properties:

```py
import gradio as gr

with gr.Blocks() as demo:
    c = gr.Textbox("cheetah.jpg", show_copy_button=True)
    demo.load(lambda :gr.update(show_copy_button=False), None, c)
    
demo.launch()
```

* I am confused constructor parameters in `gr.Gallery()`. For example, the style parameters `rows` and `columns` have been renamed to `grid_rows` and `grid_columns`. Why? Better to keep the names the same in my opinion, they're simpler as they are. More importantly, the parameters seem to have no effect. Neither does `preview` nor `update()`. Am I missing something?

```py
import gradio as gr

with gr.Blocks() as demo:
    c = gr.Gallery(["cheetah.jpg"]*6, preview=False, grid_cols=2)
    demo.load(lambda :gr.update(height=100), None, c)
    
demo.launch()
```

Good to go once these are fixed!

aliabid94:
> gr.Textbox() doesn't seem to support updating the style properties:

Fixed

> am confused constructor parameters in gr.Gallery(). For example, the style parameters rows and columns have been renamed to grid_rows and grid_columns. Why? Better to keep the names the same in my opinion, they're simpler as they are. More importantly, the parameters seem to have no effect. Neither does preview nor update(). Am I missing something?

Fixed and changed names.

Issue #4347, opened 2023-05-27
Share= True fails to generate public share link


Gothos:
### Describe the bug

Attempting to run the "Hello, World!" example given in the gradio docs, with share=True set, on a linux machine (Ubuntu 22.04, gradio version =3.32.0) simply results in a localhost link. No public sharing link is generated despite waiting for 4+ minutes.
I have looked at the previous issues involving failures in the link creation, but those explicitly return an error message.
The localhost link generated works fine.

Downgrading to gradio version = 3.13.0 successfully creates a public share link. Versions 3.15,3.16 and so on up to 3.21 do not generate links. I have not verified for versions from 3.21 to 3.32.
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```
import gradio as gr

def greet(name):
    return "Hello " + name + "!"

demo = gr.Interface(fn=greet, inputs="text", outputs="text")

demo.launch(share=True) 
```

### Logs

```shell
No errors are shown, and the localhost link works just fine.
```

### System Info

```shell
Gradio version:3.32.0, OS: Ubuntu:22.04, Browser: Firefox
```

### Severity

serious, but I can work around it

YanJiaHuan:
same @ericbellet  any progress?

ericbellet:
> same @ericbellet any progress?

No, same error. I discovered If I run the code inside a Docker container, the URL is well generated. Still don't understand why in my local machine the url is not generated

abidlabs:
Hi @ericbellet sometimes this can happen if your antivirus software blocks the download of this file. If so, you can install manually by following the steps that are printed in your console. Do you see those steps (you should run the latest version of `gradio`?

SkyWay003:
Turn off the VPN and LAN, and then try again.

PR #4380, opened 2023-06-01
Update message when share link binary is blocked

abidlabs:
Aggressive antivirus software sometimes blocks the share link binary file. Here, we make the error message more explicit so that users can find the binary file themselves and whitelist it from their antivirus software settings.

Closes: #4347


gradio/blocks.py
```diff
@@ -1845,7 +1845,7 @@ def reverse(text):
                     analytics.error_analytics("Not able to set up tunnel")
                 self.share_url = None
                 self.share = False
-                print(strings.en["COULD_NOT_GET_SHARE_LINK"])
+                print(strings.en["COULD_NOT_GET_SHARE_LINK"].format(BINARY_FILENAME))
```
freddyaboulton, line 1848: I think it would be better if we displayed BINARY_PATH so that users don't have to search for the full path themselves. They may not know where the site-packages directory is on their machine.

freddyaboulton (approved):
Minor nit but otherwise looks good to me @abidlabs !

abidlabs:
Updated. The error message looks like this:

image

PR #4383, opened 2023-06-01
Add pypi classifiers to pyproject.toml

WilliamRoyNelson:
# Description

Adds [PyPI project classifiers](https://pypi.org/classifiers/)
Classifiers are used to add information to the classifiers panel in PyPI. Example from [NumPy](https://pypi.org/project/numpy/): 
image

The reason that it is very important for this information to be present is that in an enterprise environment, security tools like [Sonatype Nexus IQ](https://help.sonatype.com/iqserver) are used to manage open source software risk. Nexus IQ specifically can be configured to [classify packages according to their license](https://help.sonatype.com/iqserver/managing/policy-management/license-threat-groups). This prevents developers from inadvertently using licenses like [GNU General Public License v2.0](https://www.tldrlegal.com/license/gnu-general-public-license-v2) without realizing that they may be legally obligated to make their entire project open source. 

My understanding is that Nexus IQ uses the classifiers panel to determine a project's license.
Because Gradio does not currently have a classifiers panel, Nexus cannot determine the license and treats Gradio as a high-risk package.
image

By adding classifiers, Gradio will have increased availability within enterprise environments.

This change does not affect the library. I noted a [comment](https://github.com/gradio-app/gradio/pull/4337#issuecomment-1567467950) in another PR that a changelog entry may not be required if the library is not affected.

# Testing/Risks
Because this change is to the build process, it's not straightforward to test this change. (e.g., you can't write a unit test)
I used the [Python Packaging Projects Tutorial](https://packaging.python.org/en/latest/tutorials/packaging-projects/#configuring-metadata) as reference when writing the classifiers.
I can't/don't know how perform a test upload of Gradio to PyPI to ensure that this works properly, so I would appreciate careful review and validation.
In addition to the License classifiers, I added a small collection of what I think are reasonable classifiers: Development Status, Programming Language, and Topics. There may be other appropriate classifiers that should be added.
If future updates to Gradio lead to Python 3.7 or other versions no longer being supported, these classifiers should be updated.

# Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,6 +22,7 @@ keywords = ["machine learning", "reproducibility", "visualization"]
 classifiers = [
   'Development Status :: 5 - Production/Stable',
   'License :: OSI Approved :: Apache Software License',
+  'Operating System :: OS Independent',
   'Programming Language :: Python :: 3',
   'Programming Language :: Python :: 3 :: Only',
   'Programming Language :: Python :: 3.7',
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,6 +19,21 @@ authors = [
 ]
 keywords = ["machine learning", "reproducibility", "visualization"]
 
+classifiers = [
+  'Development Status :: 5 - Production/Stable',
+  'License :: OSI Approved :: Apache Software License',
+  'Programming Language :: Python :: 3',
+  'Programming Language :: Python :: 3 :: Only',
+  'Programming Language :: Python :: 3.7',
+  'Programming Language :: Python :: 3.8',
+  'Programming Language :: Python :: 3.9',
+  'Programming Language :: Python :: 3.10',
+  'Programming Language :: Python :: 3.11',
+  'Topic :: Scientific/Engineering',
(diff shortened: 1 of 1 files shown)


freddyaboulton:
Thanks for the contribution @WilliamRoyNelson ! The classifiers you've added are reasonable. Let's also include `Operating System :: OS Independent` - see it [here](https://pypi.org/classifiers/)

I did an upload to test pypi and you can see the classifiers here: https://test.pypi.org/project/gradio-test/3.32.0/

WilliamRoyNelson:
> Thanks for the contribution @WilliamRoyNelson ! The classifiers you've added are reasonable. Let's also include `Operating System :: OS Independent` - see it [here](https://pypi.org/classifiers/)
> 
> I did an upload to test pypi and you can see the classifiers here: https://test.pypi.org/project/gradio-test/3.32.0/

Thank you for testing, I was really worried about that! 
I've added OS Independent as you suggested.

freddyaboulton (approved):
Thanks again @WilliamRoyNelson !!

PR #4388, opened 2023-06-01
Update utils.py for validate_url

alvindaiyan:
# Description
AWS s3 presigned url cannot work for both HEAD and GET. So when HEAD the presigned url, it will return 403(Forbidden)

Please include: 

AWS s3 presigned url can only support ONE http method. When HEAD the url, it will return 403.
To avoid expensive GET request, I added another condition for check if HEAD fails for 403, then check the GET.

Closes: # (issue)

# Checklist:


abidlabs:
Hi @alvindaiyan this looks good to me, let's use this opportunity to add a comment explaining this check

gradio/utils.py
```diff
@@ -616,7 +616,7 @@ def validate_url(possible_url: str) -> bool:
     headers = {"User-Agent": "gradio (https://gradio.app/; team@gradio.app)"}
     try:
         head_request = requests.head(possible_url, headers=headers)
-        if head_request.status_code == 405:
+        if head_request.status_code == 405 or head_request.status_code == 403:
```
abidlabs, line 619:
```suggestion
        # some URLs, such as AWS S3 presigned URLs, return a 405 or a 403 for HEAD requests
        if head_request.status_code == 405 or head_request.status_code == 403:
```

abidlabs:
Added a quick changelog, should be good to merge after CI runs

Issue #4353, opened 2023-05-28
doc: gr.Base should be gr.themes.Base


thiswillbeyourgithub:
### Describe the bug

Hi,

Reading [the documentation](https://gradio.app/docs/#themes) I saw 
![image](https://github.com/gradio-app/gradio/assets/26625900/5afde05f-d3d1-41e3-b054-b0c4fe716dd3)

Which confused me because `gr.Base` failed (no attribute) whereas `gr.themes.Base` worked perfectly.

Couldn't find where to change this in the code so I figured I'd open an issue.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

* import gradio
* gradio.themes.Base()
> <gradio.themes.base.Base object at 0x7f5823be8160>
* gradio.Base()
> Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'gradio' has no attribute 'Base'

### Logs

```shell
Python 3.9.16 (main, Dec  7 2022, 01:12:08)
[GCC 11.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import gradio
>>> gradio.themes.Base()
<gradio.themes.base.Base object at 0x7f5823be8160>
>>> gradio.Base()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'gradio' has no attribute 'Base'
```
```


### System Info

```shell
gradio==3.32.0
gradio_client==0.2.5

Python 3.9.16
```

### Severity

annoying

Issue #4338, opened 2023-05-25
`gr.Group` missing in the docs


abidlabs:
As reported [internally](https://huggingface.slack.com/archives/C02QZLG8GMN/p1685003897991719?thread_ts=1684926311.502659&cid=C02QZLG8GMN), we do not currently document `gr.Group` in the docs.

PR #4391, opened 2023-06-01
Fix Base parent in Docs, add Group

aliabd:
- Fixes Base's parent in Docs. Currently shows gradio.Base instead of gradio.themes.Base in the signature. 
- Adds gr.Group to docs. 

Closes: #4353
Closes: #4338


abidlabs:
LGTM, thank you @aliabd!

Issue #3914, opened 2023-04-19
Chatbot component does not show status tracker


freddyaboulton:
### Describe the bug

When running a chatbot demo, I noticed that the chatbot component does not display the queue status or progress information for a pending event.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Add a long sleep to the `chatbot_multimodal` demo

```python
import gradio as gr
import time

def add_text(history, text):
    history = history + [(text, None)]
    return history, ""

def add_file(history, file):
    history = history + [((file.name,), None)]
    return history

def bot(history):
    time.sleep(10)
    response = "**That's cool!**"
    history[-1][1] = response
    return history

with gr.Blocks() as demo:
    chatbot = gr.Chatbot([], elem_id="chatbot").style(height=750)
    
    with gr.Row():
        with gr.Column(scale=0.85):
            txt = gr.Textbox(
                show_label=False,
                placeholder="Enter text and press enter, or upload an image",
            ).style(container=False)
        with gr.Column(scale=0.15, min_width=0):
            btn = gr.UploadButton("📁", file_types=["image", "video", "audio"])
            
    txt.submit(add_text, [chatbot, txt], [chatbot, txt]).then(
        bot, chatbot, chatbot
    )
    btn.upload(add_file, [chatbot, btn], [chatbot]).then(
        bot, chatbot, chatbot
    )

if __name__ == "__main__":
    demo.launch()
```

Then launch with queueing with `SYSTEM=spaces python demo/chatbot_multimodal/run.py`

Open up two tabs and send a message at around the same time. 

The textbox component shows the queue information:

<img width="1136" alt="Screen Shot 2023-04-19 at 3 07 07 PM" src="https://user-images.githubusercontent.com/41651716/233175794-54467aa2-4299-4609-9163-9259cc5a6262.png">

But the chatbot does not:
<img width="1406" alt="image" src="https://user-images.githubusercontent.com/41651716/233175761-338ebd44-031a-410c-abec-832ce493becb.png">

This is a problem because if the event handler did not return the textbox, the queue information would not be displayed to the user at all!

### Logs

```shell
-
```

### System Info

```shell
3.27.0
```

### Severity

serious, but I can work around it

aliabid94:
[...]

PR #4396, opened 2023-06-01
Corner status tracker

aliabid94:
Adds support for a status tracker mode that stays in the top right corner and does not cover the output component, which can be set in an event listener with `show_progress="corner"`. This is the mode always used for chatbots, so the messages do not get covered.

Fixes: #3914 

Recording 2023-06-01 at 13 44 04


pngwn:
Could we call it `transparent` or something rather than `corner`? Or maybe `minimal` might work?

aliabid94:
> Could we call it transparent or something rather than corner? Or maybe minimal might work?

sure, how about "full", "minimal" and "hidden" then?

pngwn:
Yeah. I think that works. Just in case we move it or redesign it or whatever.

abidlabs:
Isn't this a breaking change? We should still support `show_progress`=True/False for backwards-compatibility. We can just map `True` to "full" and `False` to "hidden".

aliabid94:
> Chatbot also supports show_progress="hidden" (in case someone does not want to show the eta)
okay will do
> If someone does set show_progress="full" for a Chatbot, we should at least display a warning saying that this is not supported for Chatbot
This is kinda tricky, because the default value of the param is "full", so there's no way to tell if the user set it or not. Also, show_progress is tied to event listeners, not components. I think it's fine for this component to not support "full" mode

abidlabs:
> This is kinda tricky, because the default value of the param is "full", so there's no way to tell if the user set it or not. Also, show_progress is tied to event listeners, not components. I think it's fine for this component to not support "full" mode

Ok that's fine then

PR #4398, opened 2023-06-01
Restore offline support

aliabid94:
Allow gradio to work offline

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ No changes to highlight.
 
 ## Bug Fixes:
 
-No changes to highlight.
+- Allow gradio to work offline, by [@aliabid94](https://github.com/aliabid94) in [PR 4398](https://github.com/gradio-app/gradio/pull/4398).  
 
 ## Other Changes:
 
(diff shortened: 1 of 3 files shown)


pngwn (approved):
Looks good to me. Although this doesn't actually close the issue. There are some other cases where an internet connection is required (certain Plots) and I think gradio plus fonts should be working before we close that issue. 

I think this is a good quick fix though.

aliabid94:
cool, removed the fix claim, this should be good for most use cases

akx:
According to the comment in https://github.com/gradio-app/gradio/pull/4335#issuecomment-1567643746 ignoring CSS load errors would make the UI unusable..? 😅

pngwn:
Changed my mind. The snark is really endearing tho.

Saad-Aljebreen:
Still not able to run ChatInterface offline

kmulka-bloomberg:
@Saad-Aljebreen The ChatInterface example from the docs is working for me offline in version 4.9.0. Could you let us know which version isn't working, and a minimal reproduction of the problem?

marouahamdi:
hello everyone i m stuck with this project where in the company we work fullyyy without internet. any help plz?

Saad-Aljebreen:
Hi @marouahamdi,
I solved this through this [,](https://github.com/gradio-app/gradio/issues/4332) it helped me as in the offline mode Gradio is trying to access google fonts API.

Issue #3667, opened 2023-03-28
Change target of <a> tag in gr.HTML()


bradley-pearson6597:
### Describe the bug

I'm trying to change the href of an <a> tag so that it stays on the same page. For my use case I want to send someone back to the top of the page rather than them needing to scroll. However, when I explicitly label `target="_self"` it is overwritten to `target="_blank"`.

Is this deliberate and if so, is there a workaround?

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```python
back_to_top_btn_html = '''
<a href="#top" target="_self">Top</a></button>
'''

with gr.Blocks() as app:
    gr.HTML(back_to_top_btn_html)

app.launch(debug = True,
           share = False)
```

### Screenshot

![image](https://user-images.githubusercontent.com/44512815/228264561-e28e4ddc-9153-471c-9e5b-5fd1e054cbd8.png)

### Logs

```shell
No errors
```

### System Info

```shell
gradio == '3.19.1'
Chrome
```

### Severity

serious, but I can work around it

abidlabs:
It looks like a bug. We open links in new pages because that is better for embedded environments like Spaces but it should not apply for anchor links. Thanks for raising this issue @bradley-pearson6597!

racinmat:
I stumbled upon it too, it's problematic e.g. in Windows Kiosk mode, which is a setup with only single tab open where people do not have any way of switching tabs.

pngwn:
We could add a flag to control this behaviour.

abidlabs:
Agreed let's fix anchor links.

I don't think having a flag for non-anchor links is a good idea because it could lead to degraded experience on Spaces or other embedded environments.

racinmat:
But opening links in new tab results in degraded experience in the windows kiosk mode with browser goes to invalid state if you visit different domain than you stated when setting up the kiosk mode.
Having flag which you could turn off or on seems like viable option for me, because then people can toggle it based on their usecase.

pngwn:
> I don't think having a flag for non-anchor links is a good idea because it could lead to degraded experience on Spaces or other embedded environments.

It also makes using gradio for the whole page a bit tedious if you prefer not to open links in new tabs (which is a common use case, opening everything in a new tab is slightly obnoxious). The default should be to open links in a new tab. It I dint see any harm in allowing the author to override. 

This wouldn't affect all embedding, only iframe embedding (so basically just spaces).

PR #4405, opened 2023-06-02
Remove anchor tag target override

hannahblair:
# Description

Removes the `target` override for anchor tags with a target value that wasn't equal to `_blank`. The logic has been amended to set the `target` attribute to  `_blank` by default if no `target` value has been defined **and** the href link is external. 

Closes: #3667 

# Checklist:

# A note about the CHANGELOG

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -91,10 +91,26 @@ bash scripts/format_backend.sh
 bash scripts/format_frontend.sh
 ```
 
-You can run the circleci checks locally as well.
+## CI checks
+
+Currently the following checks are run in CI:
+
+### Gradio library (`gradio` package)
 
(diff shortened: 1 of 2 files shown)


pngwn:
We don't currently want to respect user provided target attributes because of how same tab links behave in iframes. We should hold off on making this change until we decide on an approach for that.

The issue isn't particularly clear  but what we specifically need to fix here is to skip adding the target if the href is a hash link (starts with `#`).

pngwn (approved):
Looks good to me! 

I think hash links can technically include the current page like `page#hash` and a few other variants but I don't think we need to cover all cases. We can document this clearly somewhere (maybe in a guide) as 'the way' to use anchor/ hash links.

PR #4406, opened 2023-06-02
Restrict proxy URLs

abidlabs:
Restricts proxy URLs to those that a user has loaded through `gr.load()`.

Solves two issues:

(1) Prevents users from carrying out SSRF (using other people's Spaces to make requests through the /proxy route)
(2) Prevents harvesting of a Space's HF tokens

cc @akx

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -774,7 +774,7 @@ class Blocks(BlockContext):
         components_config = config["components"]
         theme = config.get("theme", "default")
         original_mapping: dict[int, Block] = {}
-        root_urls = set()
+        root_urls = {root_url}
 
         def get_block_instance(id: int) -> Block:
             for block_config in components_config:
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -865,7 +865,6 @@ class Blocks(BlockContext):
                 blocks.__name__ = "Interface"
                 blocks.api_mode = True
 
-        root_urls.add(root_url)
         blocks.root_urls = root_urls
         return blocks
 
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -787,7 +787,7 @@ class Blocks(BlockContext):
             block_config["props"].pop("name", None)
             style = block_config["props"].pop("style", None)
             # If a Gradio app B is loaded into a Gradio app A, and B itself loads a
-            # Gradio app C, then the root_urls of the components in A needs to be the
+            # Gradio app C, then the root_urls of the components in A need to be the
             # URL of C, not B. The else clause below handles this case.
             if block_config["props"].get("root_url") is None:
                 block_config["props"]["root_url"] = f"{root_url}/"
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -762,7 +762,8 @@ class Blocks(BlockContext):
         root_url: str,
     ) -> Blocks:
         """
-        Factory method that creates a Blocks from a config and list of functions.
+        Factory method that creates a Blocks from a config and list of functions. Used
+        internally by the gradio.external.load() method.
 
         Parameters:
         config: a dictionary containing the configuration of the Blocks.
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -51,6 +51,11 @@ from gradio.utils import (
     get_continuous_fn,
 )
 
+try:
+    import spaces  # type: ignore
+except Exception:
+    spaces = None
+
 set_documentation_group("blocks")
 
 if TYPE_CHECKING:  # Only import for type checking (is False at runtime).
@@ -388,6 +393,14 @@ class BlockFunction:
         self.total_runs = 0
         self.inputs_as_dict = inputs_as_dict
         self.name = getattr(fn, "__name__", "fn") if fn is not None else None
+        self.spaces_auto_wrap()
+
+    def spaces_auto_wrap(self):
+        if spaces is None:
+            return
+        if not utils.is_space():
+            return
+        self.fn = spaces.gradio_auto_wrap(self.fn)
 
     def __str__(self):
         return str(
@@ -704,7 +717,7 @@ class Blocks(BlockContext):
         self.height = None
         self.api_open = True
 
-        self.is_space = os.getenv("SYSTEM") == "spaces"
+        self.is_space = utils.is_space()
         self.favicon_path = None
         self.auth = None
         self.dev_mode = True
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -786,6 +786,9 @@ class Blocks(BlockContext):
             block_config["props"].pop("type", None)
             block_config["props"].pop("name", None)
             style = block_config["props"].pop("style", None)
+            # If a Gradio app B is loaded into a Gradio app A, and B itself loads a
+            # Gradio app C, then the root_urls of the components in A needs to be the
+            # URL of C, not B. The else clause below handles this case.
             if block_config["props"].get("root_url") is None:
                 block_config["props"]["root_url"] = f"{root_url}/"
             else:
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -725,6 +725,7 @@ class Blocks(BlockContext):
         self.allowed_paths = []
         self.blocked_paths = []
         self.root_path = ""
+        self.root_urls = set()
 
         if self.analytics_enabled:
             is_custom_theme = not any(
@@ -745,7 +746,7 @@ class Blocks(BlockContext):
         cls,
         config: dict,
         fns: list[Callable],
-        root_url: str | None = None,
+        root_url: str,
     ) -> Blocks:
         """
         Factory method that creates a Blocks from a config and list of functions.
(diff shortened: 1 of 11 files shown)


gradio/blocks.py
```diff
@@ -758,20 +759,21 @@ def from_config(
         cls,
         config: dict,
         fns: list[Callable],
-        root_url: str | None = None,
+        root_url: str,
```
akx, line 762: Wouldn't this be a breaking change for downstream users using `from_config`? What should they put here? (Could the docstring be improved? I wouldn't understand the implications of "an external url to use as a root URL when serving files for components in the Blocks.")

gradio/routes.py
```diff
@@ -145,12 +145,16 @@ def get_blocks(self) -> gradio.Blocks:
...
-        is_hf_url = url.host.endswith(".hf.space")
+        assert self.blocks
+        # Don't proxy a URL unless it's a URL specifically loaded by the user using
+        # gr.load() to prevent SSRF or harvesting of HF tokens by malicious Spaces.
+        is_safe_url = any(url.host.endswith(root) for root in self.blocks.root_urls)
+        if not is_safe_url:
+            raise ValueError("This URL cannot be proxied.")
         headers = {}
-        if Context.hf_token is not None and is_hf_url:
+        if Context.hf_token is not None:
```
akx, line 157: I think `is_hf_url` is still relevant, because someone could `gr.load()` from a non-hf-spaces URL.

gradio/routes.py
```diff
@@ -145,12 +145,16 @@ def get_blocks(self) -> gradio.Blocks:
...
-    def build_proxy_request(url_path):
+    def build_proxy_request(self, url_path):
         url = httpx.URL(url_path)
-        is_hf_url = url.host.endswith(".hf.space")
+        assert self.blocks
+        # Don't proxy a URL unless it's a URL specifically loaded by the user using
+        # gr.load() to prevent SSRF or harvesting of HF tokens by malicious Spaces.
+        is_safe_url = any(url.host.endswith(root) for root in self.blocks.root_urls)
+        if not is_safe_url:
+            raise ValueError("This URL cannot be proxied.")
```
akx, line 155: I think ValueError is a bit too generic, how about
```suggestion
            raise PermissionError("This URL cannot be proxied.")
```

gradio/routes.py
```diff
@@ -311,7 +315,10 @@ async def favicon():
         @app.get("/proxy={url_path:path}", dependencies=[Depends(login_check)])
         async def reverse_proxy(url_path: str):
             # Adapted from: https://github.com/tiangolo/fastapi/issues/1788
-            rp_req = app.build_proxy_request(url_path)
+            try:
+                rp_req = app.build_proxy_request(url_path)
+            except ValueError as err:
```
akx, line 320:
```suggestion
            except PermissionError as err:
```

gradio/blocks.py
abidlabs: No downstream users should be using `from_config`. It's meant to be an internal method used by `gr.load()`

gradio/routes.py
abidlabs: That's actually why I made this change. The idea is that only URLs that a developer trusts (because they have explicitly `gr.load()`-ed them) will be sent requests through `/proxy`

abidlabs: I.e.
* In a Gradio app, a developer may call `gr.load()` to load external Gradio apps from Spaces or elsewhere. The URLs of these apps will be whitelisted
* Users of that Gradio app will only be able to call `/proxy` using the whitelisted URLs

akx: But why would a non-`.hf.space` origin need a `hf_token` authorization header?

abidlabs: You're right, it wouldn't. I suppose we could still leave that there

gradio/routes.py
```diff
@@ -145,12 +145,16 @@ def get_blocks(self) -> gradio.Blocks:
...
 
-    @staticmethod
-    def build_proxy_request(url_path):
+    def build_proxy_request(self, url_path):
         url = httpx.URL(url_path)
-        is_hf_url = url.host.endswith(".hf.space")
+        assert self.blocks
+        # Don't proxy a URL unless it's a URL specifically loaded by the user using
+        # gr.load() to prevent SSRF or harvesting of HF tokens by malicious Spaces.
+        is_safe_url = any(url.host.endswith(root) for root in self.blocks.root_urls)
```
akx, line 153: If `self.blocks.root_urls` is `['foo.hf.space']` this will consider `'barfoo.hf.space'` safe. I don't think that's what you meant here?

abidlabs: Ah right, will fix

abidlabs: Fixed now

abidlabs:
Suggestions should be addressed now @akx, please feel free to re-review!

test/test_routes.py
```diff
@@ -400,12 +400,15 @@ def test_mount_gradio_app_raises_error_if_event_queued_but_queue_disabled(self):
 
     def test_proxy_does_not_leak_hf_token_externally(self):
         gr.context.Context.hf_token = "abcdef"
-        r = routes.App.build_proxy_request(
-            "https://gradio-tests-test-loading-examples-private.hf.space/file=Bunny.obj"
-        )
-        assert "authorization" in dict(r.headers)
-        r = routes.App.build_proxy_request("https://google.com")
-        assert "authorization" not in dict(r.headers)
```
akx, line 408: Did you mean to delete this test? Auth headers are orthogonal to URL restriction.

test/test_routes.py
```diff
@@ -400,12 +400,15 @@ def test_mount_gradio_app_raises_error_if_event_queued_but_queue_disabled(self):
...
+        app = routes.App()
+        interface = gr.Interface(lambda x: x, "text", "text")
+        app.configure_app(interface)
+        with pytest.raises(ValueError):
+            app.build_proxy_request(
+                "https://gradio-tests-test-loading-examples-private.hf.space/file=Bunny.obj"
+            )
+        with pytest.raises(ValueError):
+            app.build_proxy_request("https://google.com")
```
akx, line 411: These are probably checking for the wrong exception now.

I think it would also be worth it to have a test that a proxy request _can_ be built when `root_urls` would allow it?

gradio/blocks.py
```diff
@@ -758,20 +759,22 @@ def from_config(
...
         config: a dictionary containing the configuration of the Blocks.
         fns: a list of functions that are used in the Blocks. Must be in the same order as the dependencies in the config.
-        root_url: an optional root url to use for the components in the Blocks. Allows serving files from an external URL.
+        root_url: an external url to use as a root URL when serving files for components in the Blocks.
         """
         config = copy.deepcopy(config)
         components_config = config["components"]
         theme = config.get("theme", "default")
         original_mapping: dict[int, Block] = {}
+        root_urls = set()
```
akx, line 777:
```suggestion
        root_urls = {root_url}
```

gradio/blocks.py
```diff
@@ -860,6 +865,8 @@ def iterate_over_children(children_list):
                 blocks.__name__ = "Interface"
                 blocks.api_mode = True
 
+        root_urls.add(root_url)
```
akx, line 868: Won't be necessary if `root_urls` is initialized to contain it:
```suggestion
```

gradio/blocks.py
```diff
@@ -783,8 +786,10 @@ def get_block_instance(id: int) -> Block:
             block_config["props"].pop("type", None)
             block_config["props"].pop("name", None)
             style = block_config["props"].pop("style", None)
-            if block_config["props"].get("root_url") is None and root_url:
+            if block_config["props"].get("root_url") is None:
                 block_config["props"]["root_url"] = f"{root_url}/"
+            else:
+                root_urls.add(block_config["props"]["root_url"])
```
akx, line 792: So this will implicitly add all of the root_urls from the (remote?) block config to the set of trusted URLs for `/proxy=`? Is that okay?

test/test_routes.py
abidlabs: > These are probably checking for the wrong exception now.

Thanks, will fix!

> I think it would also be worth it to have a test that a proxy request can be built when root_urls would allow it?

There's a different test that tests this: `test/test_external.py::TestLoadInterface::test_loading_files_via_proxy_works`

gradio/blocks.py
abidlabs: Yes, this is necessary. Consider a Gradio app B that `gr.loads()` a Gradio app A from Spaces. Then, suppose Gradio app B is published to Spaces. If a Gradio app C loads the Gradio app B, the files that are generated will actually live on the machine that is running Gradio app A, so app C needs to be able to proxy A.

akx: Yeah, figured it'd be something like that. That would probably be good to document in a comment above this line.

gradio/routes.py
freddyaboulton: You can't gr.load a non-space though?

akx: @freddyaboulton the docs for `gr.load`, emphasis mine.

> Method that constructs a Blocks from a Hugging Face repo. Can **accept model repos (if src is "models") or** Space repos (if src is "spaces"). The input and output components are automatically loaded from the repo.

abidlabs: @freddyaboulton is correct in that `gr.load()` can only be used to load Gradio apps from Spaces. You can use it with HF _model_ repos as well but the behavior is quite different (it creates a Gradio app which calls the HF Inference API under the hood). That being said @freddyaboulton I don't think this check hurts and may be useful in the future if we extend `gr.load()` to be able to load Gradio apps from arbitrary URLs similar to the gradio client.

abidlabs:
Tests have been updated -- @freddyaboulton @dawoodkhan82 or @aliabid94 would you be able to give this a pass before I merge in?

freddyaboulton (approved):
Thanks @abidlabs and @akx ! 

The changes look good to me! It looks like we're constructing the URLs incorrectly on `main` so we can fix that in another PR.

image

abidlabs:
> The changes look good to me! It looks like we're constructing the URLs incorrectly on main so we can fix that in another PR.

Hmm interesting, let me see what's happening there

abidlabs:
Ok let me merge this in and then I'll look at the file path issue in a separate PR. Thanks again for the reviews @akx and @freddyaboulton!

Issue #2209, opened 2022-09-08
max_rows/pagination does not seem to work


ankrgyl:
### Describe the bug

If you create a `DataFrame` with a certain number of `max_rows`, it still shows all of the original rows. I found https://github.com/gradio-app/gradio/issues/1201, which says they're not implemented, but both the code and documentation seem to suggest they are.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```
import gradio as gr

with gr.Blocks() as demo:
    a = gr.DataFrame(headers=['a'], value=[[1]]*20, max_rows=1)

if __name__ == "__main__":
    demo.launch()
```

### Screenshot

![Screen Shot 2022-09-07 at 5 07 18 PM](https://user-images.githubusercontent.com/565363/189005893-f0dfe52d-2192-4316-bcb5-c035987dbbfe.png)

### Logs

```shell
N/A
```

### System Info

```shell
tried with 3.1.7 and 3.2
```

### Severity

serious, but I can work around it

EmilyAlsentzer:
+1 I'm also experiencing this issue

freddyaboulton:
@dawoodkhan82 Let's work together to fix this! 

@EmilyAlsentzer @ankrgyl Thanks for reporting - we'll try to get a fix out soon 😄

arielgamino:
Experiencing the same. Hope this gets implemented as max_rows and overflow_row_behaviour don't seem to be working as indicated in docs.

Here are my examples https://colab.research.google.com/drive/1gQDcFd-nUx4iZrhew_HGFGq6_cV32OnJ?usp=sharing

freddyaboulton:
Thanks for letting us know @arielgamino !

GuillemGSubies:
Any updates here?

HennethAnnun:
Would be great to have this fixed. Best wishes!

saikrishna-punna:
is this not fixed yet? Is there a workaround?

ltm920716:
hi @freddyaboulton,

Is there any plan to fix this problem for 'max_rows'?  Or add a 'overflow_col_behaviour' parameter？

Thanks！

shon-cf:
is there any workaround meanwhile?

tommasodelorenzo:
Interested as well. Any update?

ronsamuel84629:
How can developers put something in documentation that is not working as stated. So much disappointing. Time to switch to streamlit or javascript framework

[...]

PR #4408, opened 2023-06-02
Support max_rows, max_cols in gr.Dataframe

1lint:
# Description
Add support for `max_rows `and `max_cols `parameters in `gr.Dataframe` by slicing input dataframes in the `postprocess` function. 

These parameters didn't seem to do anything when I tried to use them, then I saw it was an open issue

Here's a quick UI test to see that it works
```
import gradio as gr
import pandas as pd
import numpy as np

with gr.Blocks() as demo:
    df_display = gr.Dataframe(max_rows = 2, max_cols = 4)

    examples_rows = 5
    example_cols = 10

    with gr.Row():
        gr.Examples(
            inputs = df_display,
            examples=[
                np.zeros((examples_rows, example_cols)), # numpy array
                [[0]*example_cols]*examples_rows, # python array
                pd.DataFrame([[0]*example_cols]*examples_rows), # pandas array
                {
                    "headers" : list(range(example_cols)), # dict array
                    "data" : [[0]*example_cols]*examples_rows
                }
            ],
            examples_per_page=5,
        )

demo.launch()
```

Closes: # (issue)
https://github.com/gradio-app/gradio/issues/2209
# Checklist:

- [x ] I have performed a self-review of my own code
- [x ] I have added a short summary of my change to the CHANGELOG.md
- [x ] My code follows the style guidelines of this project
- [x ] I have commented my code in hard-to-understand areas
- [x ] I have made corresponding changes to the documentation
- [x ] I have added tests that prove my fix is effective or that my feature works
- [x ] New and existing unit tests pass locally with my changes

# A note about the CHANGELOG

### DIFF
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -3058,7 +3058,7 @@ class Dataframe(Changeable, Inputable, Selectable, IOComponent, JSONSerializable
             return y
         if isinstance(y, str):
             dataframe = pd.read_csv(y, nrows=self.max_rows)
-            dataframe = dataframe.loc[:, dataframe.columns[: self.max_cols]]
+            dataframe = dataframe.iloc[:, : self.max_cols]
             return {
                 "headers": list(dataframe.columns),
                 "data": Dataframe.__process_markdown(
@@ -3066,7 +3066,7 @@ class Dataframe(Changeable, Inputable, Selectable, IOComponent, JSONSerializable
                 ),
             }
         if isinstance(y, pd.DataFrame):
-            y = y.loc[: self.max_rows - 1, y.columns[: self.max_cols]]
+            y = y.iloc[: self.max_rows, : self.max_cols]
             return {
                 "headers": list(y.columns),  # type: ignore
                 "data": Dataframe.__process_markdown(
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -3057,8 +3057,8 @@ class Dataframe(Changeable, Inputable, Selectable, IOComponent, JSONSerializable
             )
             return y
         if isinstance(y, str):
-            dataframe = pd.read_csv(y)
-            dataframe = dataframe.loc[: self.max_rows - 1, y.columns[: self.max_cols]]
+            dataframe = pd.read_csv(y, nrows=self.max_rows)
+            dataframe = dataframe.loc[:, dataframe.columns[: self.max_cols]]
             return {
                 "headers": list(dataframe.columns),
                 "data": Dataframe.__process_markdown(
--- a/gradio/components.py
+++ b/gradio/components.py
@@ -3051,9 +3051,14 @@ class Dataframe(Changeable, Inputable, Selectable, IOComponent, JSONSerializable
         if y is None:
             return self.postprocess(self.empty_input)
         if isinstance(y, dict):
+            y["headers"] = y["headers"][: self.max_cols]
+            y["data"] = list(
+                map(lambda row: row[: self.max_cols], y["data"][: self.max_rows])
+            )
             return y
         if isinstance(y, str):
             dataframe = pd.read_csv(y)
+            dataframe = dataframe.loc[: self.max_rows - 1, y.columns[: self.max_cols]]
             return {
                 "headers": list(dataframe.columns),
                 "data": Dataframe.__process_markdown(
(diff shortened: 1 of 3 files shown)


pngwn:
I think we should probably hold off on adding features with 4.0 around the corner. We may want to revisit some of the Dataframe features and kwargs. 

Cc @abidlabs

abidlabs:
@pngwn while it's true that we shouldn't add new features to the `gr.Dataframe`, this is technically a bug fix since these parameters are already in the constructor of `gr.Dataframe`. 

Imo we should review this PR and merge in if the fix looks good because if @1lint hadn't opened this PR, @hannahblair or I would have taken this issue on as part of 4.0 work anyways to ensure that the `gr.Dataframe` is in a good state before the 4.0 release.

pngwn:
Sounds good.

1lint:
Great glad I could help! I just pushed an updated test to cover string inputs, and a related bug fix for my 1st commit

gradio/components.py
```diff
@@ -3051,16 +3051,22 @@ def postprocess(
...
+            dataframe = pd.read_csv(y, nrows=self.max_rows)
+            dataframe = dataframe.loc[:, dataframe.columns[: self.max_cols]]
             return {
                 "headers": list(dataframe.columns),
                 "data": Dataframe.__process_markdown(
                     dataframe.to_dict(orient="split")["data"], self.datatype
                 ),
             }
         if isinstance(y, pd.DataFrame):
+            y = y.loc[: self.max_rows - 1, y.columns[: self.max_cols]]
```
freddyaboulton, line 3069: We should be using .iloc here for the rows as .loc is label-based and will not work if the dataframe has a custom index.

image

gradio/components.py
```diff
@@ -3051,16 +3051,22 @@ def postprocess(
...
         if isinstance(y, dict):
+            y["headers"] = y["headers"][: self.max_cols]
+            y["data"] = list(
+                map(lambda row: row[: self.max_cols], y["data"][: self.max_rows])
+            )
             return y
         if isinstance(y, str):
-            dataframe = pd.read_csv(y)
+            dataframe = pd.read_csv(y, nrows=self.max_rows)
+            dataframe = dataframe.loc[:, dataframe.columns[: self.max_cols]]
```
freddyaboulton, line 3061: .loc is ok here since we're only indexing the columns by name

freddyaboulton (commented):
Thank you for the contribution @1lint ! And for writing the test.

Broader question for @pngwn and @abidlabs - shouldn't this be a front-end change though? I think it would be best if these parameters worked in conjunction with pagination to control the data that's initially displayed. Doing it in the backend means the data outside of `max_rows` and `max_cols` is "lost".

gradio/components.py
1lint: Good catch! I will update the slicing and test accordingly

gradio/components.py
1lint: Thanks for pointing this out, will update it to iloc as well for consistency, should be more efficient too

1lint:
@freddyaboulton thanks for the great catch on `loc` vs `iloc` usage in the dataframe!

That makes sense that the user may expect the `max_rows/max_cols` parameters to only affect the dataframe display, but preserve the rest of the data when using the dataframe as an input component. 
I would be worried that the user may be using these parameters because they have a large dataframe, in which case the truncated data may take up alot of bandwidth during the round trip to/from the frontend. 

It may be better to keep the truncated data in some sort of internal state component, and only send to the frontend the data that the user can see/modify. Then the (potentially) modified data read from the frontend can be recombined with the full data saved in the state component

This is essentially what I'm doing for an app to apply SQL queries to HF datasets - the user can select what dataset to load, but only the first 40 rows of the dataset are shown on the frontend, while the entire dataset is saved in gr.State. The user can quickly try out sql queries on this first 40 rows to see how it affects the output inside gr.Dataframe. Finally when they are happy with the result, they can apply the query to the whole dataset in gr.State and download the result.

pngwn:
This is why I wanted to leave this for now. It really depends on how people will use it but because we have sortable data frames the expectation would be that the first n rows would be returned of the complete sorted dataset. Likewise, if someone is aggregating the data in other output components then those aggregations should operate on the entire dataset, not just a subset of it.

pngwn:
I think some of these cases will require significant changes to work correctly and there may need to be breaking changes too. Right now whatever the user function returns is what gets displayed, whatever gets passed back is the data you can operate on. Internal state would break this expectation.  

Adding in some concept of internal state, would probably cause issues for the API as well.

abidlabs:
I agree that `max_rows` and `max_cols` should filter on the frontend. If only a subset of data should be returned from the backend, that can be a part of the prediction function.

@1lint let's go ahead and close this PR since we'd need to approach quite differently. And @freddyaboulton let's move #2209 to the 4.0 milestone?

1lint:
Sounds good @abidlabs, good point this approach should be handled in the prediction function

PR #4409, opened 2023-06-03
Add pypi classifiers to client pyproject.toml

WilliamRoyNelson:
# Description

Adds classifiers to [gradio-client](https://pypi.org/project/gradio-client/) as a continuation of #4383 
which added the following classifiers to the PyPI metadata for gradio:
image

The only difference with this is that it's for [gradio-client](https://pypi.org/project/gradio-client/), and the Scientific/Engineering :: Visualization tag has been replaced with Software Development :: User Interfaces

Previous change, #4383, was successful and can be seen at https://pypi.org/project/gradio/

No Changelog is necessary as it does not change the functionality of the library

- [X] I have added a short summary of my change to the CHANGELOG.md
directly via drag-and-drop.

### DIFF
--- a/client/python/pyproject.toml
+++ b/client/python/pyproject.toml
@@ -19,6 +19,22 @@ authors = [
 ]
 keywords = ["machine learning", "client", "API"]
 
+classifiers = [
+  'Development Status :: 4 - Beta',
+  'License :: OSI Approved :: Apache Software License',
+  'Operating System :: OS Independent',
+  'Programming Language :: Python :: 3',
+  'Programming Language :: Python :: 3 :: Only',
+  'Programming Language :: Python :: 3.7',
+  'Programming Language :: Python :: 3.8',
+  'Programming Language :: Python :: 3.9',
+  'Programming Language :: Python :: 3.10',
+  'Programming Language :: Python :: 3.11',
+  'Topic :: Scientific/Engineering',
+  'Topic :: Scientific/Engineering :: Artificial Intelligence',
+  'Topic :: Software Development :: User Interfaces',
+]
+
 [project.urls]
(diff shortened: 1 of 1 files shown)


client/python/pyproject.toml
```diff
@@ -19,6 +19,22 @@ authors = [
 ]
 keywords = ["machine learning", "client", "API"]
 
+classifiers = [
+  'Development Status :: 5 - Production/Stable',
```
freddyaboulton, line 23: Only nit here @WilliamRoyNelson is that we should probable use `Development Status :: 4 - Beta` as the client is still pre 1.0 version and some apis may change. Do you mind making the change?

Otherwise this looks good to me! Thanks again for the help.

WilliamRoyNelson: Thank you. I've made the requested change... and accidentally closed the PR while trying to rebase, but I think it's good and reopened now.

abidlabs: Thanks @WilliamRoyNelson!

abidlabs:
Will go ahead and merge this in, thanks again @WilliamRoyNelson and @freddyaboulton

WilliamRoyNelson:
@abidlabs @freddyaboulton 
I was wondering if this could get pushed to PyPI to update the metadata.

Also, I've noticed that the version on PyPI is listed as 0.2.5, but the [changelog](https://github.com/gradio-app/gradio/blob/main/client/python/CHANGELOG.md) is still at 0.2.4
Do you want me to put in a PR to move the documentation forward to reflect PyPI's current state?
https://github.com/WilliamRoyNelson/gradio/commit/e41f8b66723dc5d469c780182deaf2364d9df159

PR #4422, opened 2023-06-05
add new audio + video events

### LINKED ISSUE #4026: Stop event handler not triggered for audio
### Describe the bug

The `stop` event handle is not triggered for audio

### Reproduction

```
import gradio as gr

def stop_streaming():
    print("Streaming has stopped")

if __name__ == "__main__":

    with gr.Blocks() as demo:

        stream_input = gr.Audio(source="microphone")
        stream_input.stop(stop_streaming)

    demo.queue().launch(share=False,  debug=False)
```

### Screenshot

"Streaming has stopped" is not printed to the console.

image

### Logs

```shell
-
```

### System Info

```shell
3.28.1
```

### Severity

pngwn:
# Description

Closes #4026.

This PR is part fix, part proposal. 

Firstly I have fixed the `stop` event, that was never actually working. Secondly I have added two new events, `start_recording` and `stop_recording`. I wanted to add these to the `Streamable` class but I couldn't because Video isn't streamable but `Image` is, so I have added a new `Recordable` Class that we mix into `Video` and `Audio`.

For clarity the 'media' type events are as follows:

- `play` - when a media track starts playing
- `pause` - when a media track stops playing
- `stop` - when a media track has ended
- `start_recording` - when a user starts to record a user input (mic or webcam input)
- `stop_recording` - when a user stops recording a user input (mic or webcam)

You can test Video with the following:

```py
import gradio as gr

def start_recording():
    print("Recording has started")

def stop_recording():
    print("Recording has stopped")

def start_playing():
    print("Playing has started")

def stop_playing():
    print("Playing has stopped")

def pause_playing():
    print("Playing has paused")


with gr.Blocks() as demo:
    stream_input = gr.Video(source="webcam")

    stream_input.play(start_playing)
    stream_input.pause(pause_playing)
    stream_input.stop(stop_playing)
    stream_input.start_recording(start_recording)
    stream_input.stop_recording(stop_recording)
        
demo.launch()
````

You can use the same test for Audio, you'll just need to change the component to this:

```py
stream_input = gr.Audio(source="microphone")
```

We can change the names but i think the semantics make sense. lmk what you think.


pngwn:
Maybe we need tests for this?

aliabid94:
I don't think we have any existing tests for different types of event listeners. Probably something to consider when we add end to end tests

js/video/src/Video.svelte
```diff
@@ -18,14 +18,16 @@
...
+		change: any;
 		clear: undefined;
 		play: undefined;
 		pause: undefined;
 		ended: undefined;
 		drag: boolean;
 		error: string;
 		upload: FileData;
+		start_recoring: undefined;
+		stop_recoring: undefined;
```
aliabid94, line 30:
```suggestion
		start_recording: undefined;
		stop_recording: undefined;
```

aliabid94 (approved):
lgtm. I do think recordable is something separate from streamable so it makes sense to have another class to me. Was trying to think of cleaner words for `start_recording`/`stop_recording` but was tricky (maybe `record`/`cut`?). Also wonder if `stop` should be `finish` or something else. Though this naming is not something I feel strongly about.

pngwn:
Yeah stop is technically wrong but I thought we could fix that in 4.0 rather than supporting both (to avoid breaking changes) since stop has never actually worked. `ended` is the DOM event for this and would probably be a good name. I can add ended as well though if needed. Then we can deprecate `stop` and it is a simple removal in 4.0. 

I don't feel strongly about the names. `record` makes sense but `cut` makes me think of cutting it trimming a track.

pngwn:
I added `ended` and deprecated `stop` (we can remove in 4.0). 

Let me know if there are final comments about this API before I merge. Cc @abidlabs

gradio/events.py
```diff
@@ -264,6 +275,25 @@ def check_streamable(self):
...
+        """
+
+        self.stop_recording = EventListenerMethod(self, "stop_recording")
+        """
+        This listener is triggered when the user stops recording with the component (e.g. audio or video).
+        This method can be used when this component is in a Gradio Blocks.
+        """
+
+    def check_streamable(self):
+        pass
```
abidlabs, line 294: Why do we have this function?

abidlabs:
Why don't we just use `.start()` and `.stop()` as the names of the events that happen when you start recording or stop recording? Those conform with our existing events, which are all one syllable present-tense verbs. I think the usage is clear enough:

```py
mic = gr.Audio(source="microphone")
mic.start(...)
mic.stop(...)
```

Looking back at the discussion issue, #4026, I'm pretty sure `stop()` was expected to fire when the user stopped recording, not when a tracked finished playing, so we don't need to deprecate anything either.

abidlabs:
In other words, I'm suggesting the following:

* In the `Recordable` class, rename `.start_recording()` to `.start()` and `.stop_recording()` to `.stop()`
* In the `Playable` class, rename `.stop()` to `.finish()`

pngwn:
How is a user supposed to know what they refer to, that is just confusing imo. Start, play, stop, finish. They could refer to anything. I think the events should be easy to figure out without always referring to the docs, I don't think that is the case with these suggestions. Stop is especially problematic because it generally (in common usage) refers to stopping the media not stopping a recording. 

I also don't think it is very forward thinking, for example we have briefly discussed improving the audio component I think pausing recording would be one of those improvements. How would we add a pause event for pausing recording without it being inconsistent?

I understand the desire for short event names but I think this change would prioritise terseness over intuitiveness.

pngwn:
> Looking back at the discussion issue, #4026, I'm pretty sure stop() was expected to fire when the user stopped recording, not when a media file finished playing, so we don't need to deprecate anything either.

I disagree with this. That was @freddyaboulton's interpretation but stop was always intended to mirror the 'stop' of a media track (I remember because I write the original video component). I would always expect stop to mean stopping the media.

abidlabs:
It’s not just about terseness, but about what API a user would expect. Given that our other events are named “change”, “blur”, etc, an event named “start_recording” feels very unexpected to me. I personally think “start” (or “record”) and “stop” are more expected. I do see your point about them being less clear (but so are some of our other event names), so  I don’t feel too strongly about these two. I do think we should rename “ended()” to “end()” or “finish()” though

pngwn:
Yeah, I agree about `ended` that is the wrong tense.

PR #4430, opened 2023-06-05
Fix loading private Spaces

abidlabs:


freddyaboulton (commented):
Thanks @abidlabs ! I tested and it fixes the issue but I wonder if we can make the front-end more robust to not include double `//` in the future so that this doesn't break again with a future refactor.

In python there's `urljoin` which is built in and takes care to not add double slashes:

image

I wonder if we can leverage something similar in js cc @pngwn @hannahblair

abidlabs:
Yes, I was thinking the same. Looks like there are some 3rd party libraries but nothing built into JS for this. Correct me if I'm wrong @pngwn @hannahblair

abidlabs:
Nice find! Will update thanks for the suggestion @freddyaboulton @hannahblair

abidlabs:
Sounds good to me @hannahblair! Thanks for looking into it and thanks @freddyaboulton for the review.

Will merge in after CI passes

pngwn:
This stuff happens on the client, everything we do in JS is client only, so node builtins won't work. We can create some kind of abstraction though.

Issue #4364, opened 2023-05-30
UploadButton cannot be made non-interactive


freddyaboulton:
### Describe the bug

The UploadButton can be made non-interactive but I can imagine a use case where this is desired: While a function is running prevent the user from uploading another file.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

`gr.UploadButton(interactive=False)` will still let you upload a file

### Logs

```shell
-
```

### System Info

```shell
Latest main
```

### Severity

annoying

abidlabs:
As discussed via Slack DM, I'd like to give this a shot @pngwn

PR #4436, opened 2023-06-06
Adds `variant` and `interactive` props to `gr.UploadButton`

abidlabs:
Closes: #4364 

Test code:

```py
import gradio as gr

with gr.Blocks() as demo:
    gr.Button()
    gr.Button(interactive=False)
    gr.Button(variant="primary")
    gr.Button(variant="stop", interactive=False)
    gr.UploadButton()
    gr.UploadButton(interactive=False)
    gr.UploadButton(variant="primary")
    gr.UploadButton(variant="stop", interactive=False)
```

image

### DIFF
--- a/js/app/src/components/UploadButton/UploadButton.svelte
+++ b/js/app/src/components/UploadButton/UploadButton.svelte
@@ -18,7 +18,6 @@
 	export let root: string;
 	export let mode: "static" | "dynamic" = "dynamic";
 	export let variant: "primary" | "secondary" | "stop" = "secondary";
-	console.log("mode", mode);
 
 	async function handle_upload({ detail }: CustomEvent<FileData>) {
 		value = detail;
--- a/js/app/src/components/UploadButton/UploadButton.svelte
+++ b/js/app/src/components/UploadButton/UploadButton.svelte
@@ -16,6 +16,9 @@
 	export let file_count: string;
 	export let file_types: Array<string> = ["file"];
 	export let root: string;
+	export let mode: "static" | "dynamic" = "dynamic";
+	export let variant: "primary" | "secondary" | "stop" = "secondary";
+	console.log("mode", mode)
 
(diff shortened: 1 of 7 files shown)


js/app/src/components/UploadButton/UploadButton.svelte
```diff
@@ -16,6 +16,9 @@
 	export let file_count: string;
 	export let file_types: Array<string> = ["file"];
 	export let root: string;
+	export let mode: "static" | "dynamic" = "dynamic";
+	export let variant: "primary" | "secondary" | "stop" = "secondary";
+	console.log("mode", mode);
```
dawoodkhan82, line 21: remove console.log

dawoodkhan82 (changes requested):

dawoodkhan82:
Seems Like the UploadButton is still interactive even when set to false.

dawoodkhan82:
Also what is the purpose of adding variants?

js/app/src/components/UploadButton/UploadButton.svelte
abidlabs: Whoops fixed

abidlabs:
> Also what is the purpose of adding variants?

Just for further customization. I imagine someone might want primary/secondary at least. 

> Seems Like the UploadButton is still interactive even when set to false.

Really? Is this after the frontend has been rebuilt? Here's what I get testing the code above:

image

dawoodkhan82:
works for me, didn't have the latest changes pulled

dawoodkhan82:
> Just for further customization. I imagine someone might want primary/secondary at least.
Should we remove `stop` as a variant? Seems confusing for this component

Otherwise lgtm!

abidlabs:
> Should we remove stop as a variant? Seems confusing for this component

I think it's better to make the API between `Button` and `UploadButton` as consistent as possible. It's fine if the "stop" variant is not really used, but it might turn out to be useful in styling. 

Issue #3402, opened 2023-03-07
Support the `paused` status of spaces in embeds.


pngwn:
- [x] I have searched to see if a similar issue already exists.

We don't currently support the `PAUSED` status of spaces which occurs when the space author has explicitly stopped the space.

It would be nice to support this.

PR #4438, opened 2023-06-06
Support PAUSED Space status when embedding

abidlabs:
Made a small PR to add support for PAUSED status in Spaces when embedding. However, I then realized that I don't know a good way to test this locally, so any guidance on testing this would be appreciated @pngwn! 

Closes: #3402


client/js/src/client.ts
```diff
@@ -1129,9 +1129,16 @@ async function check_space_status(
...
-			}, 1000);
-			break;
-		// poll for status
+			}, 1000);  // poll for status
+			break; 
+		case "PAUSED":
+			status_callback({
+				status: "paused",
+				load_status: "pending",
+				message: "This space has been paused.",
```
pngwn, line 1138: Maybe we could prompt the viewer to duplicate the space in this case (since the author has probably paused it for cost reason).

```suggestion
				message: "This space has been paused by the author. If you would like to try this demo consider duplicating the space",
```

Maybe we should add a link here but we don't have a great way of including links atm.

abidlabs:
Ran the formatter. Anything else I can do here to confirm the fix works @pngwn?

pngwn:
If you run `pnpm build:cdn-local` and then run `pnpm preview:cdn-local`. A preview of the CDN build will start, follow the link in your terminal to preview it. In `ui/_cdn-test`, you can edit the `index.html` to preview an embedded spaces app. Pause the spaces app and reload the preview page to check the changes.

Will make this easier at some point.

abidlabs:
Thanks @pngwn for the guidance! I've added a paused Space to `index.html` to confirm that the PAUSED status shows up correctly:

image

client/js/src/client.ts
```diff
@@ -1117,6 +1117,8 @@ async function check_space_status(
 		id: space_name
 	} = response;
 
+	console.log(id, "stage", stage);
```
pngwn, line 1120: Remove comment

client/js/src/client.ts
```diff
@@ -1129,9 +1131,18 @@ async function check_space_status(
...
 				check_space_status(id, type, status_callback);
-			}, 1000);
+			}, 1000); // poll for status
+			break;
+		case "PAUSED":
+			status_callback({
+				status: "paused",
+				load_status: "error",
+				message:
+					"This space has been paused by the author. If you would like to try this demo, consider duplicating the space",
```
pngwn, line 1141: nit:

```suggestion
					"This space has been paused by the author. If you would like to try this demo, consider duplicating the space.",
```

pngwn (commented):
@abidlabs In your screenshot it says 'contact the author of the page' but you have discussions enabled on your space, so it should be outputting a link to create a discussion instead (with the message that your wrote in `Index.svelte` line 256). Any idea what might be going on?

abidlabs:
> In your screenshot it says 'contact the author of the page' but you have discussions enabled on your space, so it should be outputting a link to create a discussion instead (with the message that your wrote in Index.svelte line 256). Any idea what might be going on?

Interesting, I'll take a look!

abidlabs:
Thanks @pngwn for catching that. Fixed now:

image

pngwn (approved):
Awesome, I checked at everything seems to be working. Thanks @abidlabs!

abidlabs:
Thanks for the review @pngwn!

PR #4443, opened 2023-06-07
Fix Httpx request UI timeout issue, time-consuming task waiting for 30 seconds

Neokmi:
Fix Httpx request UI timeout issue, time consuming task waiting for 30 seconds
#Description
During the use of multiple SD-WebUI plugins, strange errors are often thrown by the front-end UI after clicking the submit button After checking and debugging later, it was found that the error was caused by the last response result issued by "ws://localhost: 7860/queue/join" The code was located because the default timeout for httpx requests was too short, resulting in a response timeout But in reality, there are many backend tasks that are time-consuming (such as loading the model before processing the operation). Usually, it takes more than ten seconds, so the timeout must be set higher
I have completed multiple plugin tests on SD WebUI locally And they run well without any further errors, and even the 15 second processing task returns correct results without any errors
Please include:
*Relevant motivation
 Fix bugs and optimize user experience
*A summary of the change
 I'm not sure if it will cause other issues, such as long user waiting times
*Which issue is fixed
 Yes
*Any additional dependencies that are required for this change
 No need
Closes: # (issue)
#Checklist:
-[Y] I have performed a self review of my own code
-[Y] I have added a short summary of my change to the CHANGELOG.md
-[Y] My code follows the style guidelines of this project
-[Y] I have commented my code in hard to understand areas
-[Y] I have made corresponding changes to the documentation
-[Y] I have added tests that protect my fix is effective or that my feature works
-[Y] New and existing unit tests pass locally with my changes
#A note about the CHANGELOG
Hello 👋  And thank you for contributing to Graph!
All pull requests must update the change log located in CHANGELOG.md, less the pull request is labeled with the "no changelog update" label
Please add a brief summary of the change to the Upcoming Release>Full Changelog section of the CHANGELOG.md file and include
A link to the PR (formatted in markup) and a link to your github profile (if you like) For example, "* Added a cool new feature by ` [@ myusername] (link to your github profile) ` in ` [PR 11111]（ https://github.com/gradio-app/gradio/pull/11111 ）`
If you would like to elaborate on your change further, feel free to include a longer explanation in the other sections
GitHub web UI since that lets you upload files directly via drag and drop. cuowu


gradio/utils.py
```diff
@@ -432,7 +432,8 @@ async def __run(self) -> AsyncRequest:
         """
         try:
             # Send the request and get the response.
-            self._response: httpx.Response = await self.client_.send(self._request)
+            # Fix Httpx request UI timeout issue, time-consuming task waiting for 30 seconds
```
abidlabs, line 435: No need for this comment
```suggestion
```

CHANGELOG.md
```diff
@@ -10,7 +10,7 @@
 - Fixed bug where `gr.File(file_count='multiple')` could not be cached as output by [@freddyaboulton](https://github.com/freddyaboulton) in [PR 4421](https://github.com/gradio-app/gradio/pull/4421)
 - Restricts the domains that can be proxied via `/proxy` route  by [@abidlabs](https://github.com/abidlabs) in [PR 4406](https://github.com/gradio-app/gradio/pull/4406). 
 - Fix video rendering in Safari by [@aliabid94](https://github.com/aliabid94) in [PR 4433](https://github.com/gradio-app/gradio/pull/4433).
-
+- Fix Httpx request UI timeout issue, time-consuming task waiting for 30 seconds
```
abidlabs, line 13:
```suggestion
- Fix Httpx request timeout issue by waiting for 30 seconds by @Neokmi](https://github.com/neokmi) in [PR 4443](https://github.com/gradio-app/gradio/pull/4443).
```

abidlabs:
Hi @Neokmi when I test this locally, I get the following error:

image

I'm running `httpx` version 0.24.1 on MacOS. I see the same problem on Spaces here: https://huggingface.co/spaces/gradio-pr-deploys/pr-4443-all-demos

abidlabs (changes requested):
See above

Neokmi:
Okay, mine is the Win11 system. I'll check it again

Neokmi:
> 见上文

My httpx version is 0.13.3, and the code on my end is supported.
It may be a compatibility issue with the httpx version However, in my httpx source code, the AsyncClient.send method found the timeout parameter. I need to debug the httpx code for this
Snipaste_2023-06-08_08-18-31
Snipaste_2023-06-08_08-13-20

According to official documentation, the default timeout for the client is 5 seconds, which is still too short

Snipaste_2023-06-08_08-16-32

This can cause communication errors in backend remote calls :
Snipaste_2023-06-07_14-33-13



Perhaps setting a global timeout when initializing the AsyncClient client would be better

Snipaste_2023-06-08_08-10-41

akx:
According to [the changelog](https://github.com/encode/httpx/blob/master/CHANGELOG.md#0200-13th-october-2021), `client.send` does not accept `timeout=` since httpx 0.20.0, which was released in October 2021 (nearly 1 year 8 months ago).

The version you're using is over 3 years old.

I would maybe recommend changing the `httpx` dependency for `gradio` to `>=0.20.0`, and then changing this code to just always use `build_request` as recommended in the changelog there.

Neokmi:
Indeed, some frameworks mandate the use of older versions of httpx, but the new version of httpx.send does not support timeout settings Unless changed to another method setting Well, keep this suggestion for now because it is indeed a problem

akx:
@Neokmi What frameworks are those? You should talk with their maintainers to allow a newer version than something 3 years old – who knows how many security issues have been fixed since!

Issue #3418, opened 2023-03-08
Design of the error labels have changed and they don't look good


pngwn:
### Describe the bug

Dark mode change isn't awful, although it is less noticeable. The light mode one doesn't look good at all though.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Before:

<img width="117" alt="Screenshot 2023-03-08 at 11 46 16" src="https://user-images.githubusercontent.com/12937446/223706328-9805cb65-0960-4477-b943-8a7ca38e7720.png">
<img width="117" alt="Screenshot 2023-03-08 at 11 47 11" src="https://user-images.githubusercontent.com/12937446/223706331-cf495742-ace2-451d-8c37-bb1c2ebf92b2.png">

After: 
<img width="117" alt="Screenshot 2023-03-08 at 11 41 04" src="https://user-images.githubusercontent.com/12937446/223706369-ffe64523-6cb3-4551-9457-57c99213e9dc.png">
<img width="117" alt="Screenshot 2023-03-08 at 11 41 20" src="https://user-images.githubusercontent.com/12937446/223706373-8f450832-b1ee-44a5-9c45-c04711efd46d.png">

### Logs

```shell
-
```

### System Info

```shell
`main`
```

### Severity

annoying

pngwn:
cc @aliabid94

abidlabs:
Kinda feeling the opposite lol. New light mode looks nice but dark mode looks basic to me. Fine with reverting to original though

pngwn:
Light mode looks curved because of the gradient now, which none of the rest of UI does. Dark mode looks clean but isn't as obvious as it used to be.

I didn't really like the old ones though, I'd be happy with some thing else.

abidlabs:
> Light mode looks curved because of the gradient now, which none of the rest of UI does. Dark mode looks clean but isn't as obvious as it used to be.

I see, good point

aliabid94:
Hmm ok so I was consolidating CSS variables and I consolidated variables between this and the error modal, so they have the same background. I'd like to keep them the same - would be happy to remove the gradient from both! wdyt?

![Screen Shot 2023-03-14 at 8 49 52 PM](https://user-images.githubusercontent.com/7870876/225201608-21346732-78af-4cdd-ad61-af3cec895d77.png)

abidlabs:
Seems reasonable, I agree both the Error message and the Error modal should have the same background. Can we see what the error modal looks like without the gradient?

[...]

PR #4454, opened 2023-06-08
Amend error toast and pill styling

hannahblair:
# Description

This PR removes the gradient of the error toast and pill components. The background is now a solid red in light mode and the toast has a dark red border to the toast as seen in the screenshots. 

Screenshot 2023-06-08 at 14 31 01
Screenshot 2023-06-08 at 14 30 48

Closes: #3418 

# Checklist:

# A note about the CHANGELOG

### DIFF
--- a/test/test_theme_sharing.py
+++ b/test/test_theme_sharing.py
@@ -145,6 +145,12 @@ dracula = gr.themes.Base(
 ).set(
     body_background_fill=dracula_gray.c500,
     color_accent_soft=dracula_gray.c100,
+    error_border_color="#fecaca",
+    error_background_fill="#fee2e2",
+    error_icon_color="#b91c1c",
+    error_icon_color_dark="#ef4444",
+    error_text_color="#ef4444",
+    error_text_color_dark="#ef4444",
     background_fill_primary=dracula_gray.c500,
     background_fill_secondary=dracula_gray.c500,
     block_background_fill=dracula_gray.c300,


pngwn:
I like this better but i still think the pink background is weird (i'm no designer).

The only thing I am genuinely worried about with the background colour is red on pink, for accessibility reasons. I'm pretty sure the contrast wouldn't pass WCAG guidelines, although I haven't checked.

abidlabs:
Tested on a variety of themes/demos and really like the sharp errors! Let's check the accessibility concerns @pngwn pointed out (perhaps we can make the red font darker if needed), but otherwise LGTM

dawoodkhan82:
Looks clean. In order to affect the other themes, should we add these changes to `base.py` instead?

dawoodkhan82:
Also I never really liked the big "X" button on the toast. I remember not even realizing it was a button at first, just thought it was an icon.

gradio/themes/default.py
```diff
@@ -47,6 +47,8 @@ def __init__(
         super().set(
             # Colors
             input_background_fill_dark="*neutral_800",
+            error_background_fill=colors.red.c200,
+            error_border_color=colors.red.c600,
```
dawoodkhan82, line 51: In order to affect the other themes, should we add these changes to `base.py` instead?

pngwn:
Regarding @dawoodkhan82's comment, what if the `X` icon was just an error icon and we had a standard close `x` button on the right (centred or top-right)? I think that would be clearer. It isn't obvious at first glance that the `X` icon is actually clickable. This would also work better if we added `info` `warn` variants in the future as we discussed recently.

Another alternative to the pinkish background is to make it more consistent with the dark mode one, a flat light background with red title text + icon. This would resolve the a11y concerns too. Just ideas though, we can also merge this in an iterate more in the future.

Some examples showing what I mean (although these don't match the style of gradio and some of the colours are a bit washed out):

---

I don't love the alignment on this but it is very readable. I also don't think this stands out enough but it is a decent start.

Screenshot 2023-06-09 at 13 28 21

---

I think these are actually quite good, although the line is kinda weird.

Screenshot 2023-06-09 at 13 28 41

---

This demonstrates how much cleaner the plain background is although the left border just looks out of place and these are slightly more compact than our error toast.

Screenshot 2023-06-09 at 13 29 50

---

These don't stand out nearly enough. The colours aren't saturates enough but the principle is nice.

Screenshot 2023-06-09 at 13 31 04

---

I quite like this one, especially if it stood out a little more (maybe a larger icon and something more dramatic than a sad face). That red bar at the bottom is actually a progress bare for when the toast will be removed, which is a nice touch imo. Especially if we added 'persistent' and 'non-persistent' toasts. A solid, unmoving bar (or the absence of a bar) could act a a subtle queue that the toast will remain. The presence of a moving bar suggests it will vanish.

Screenshot 2023-06-09 at 13 31 24

---

This is actually a really nice example of what I mean, although the error colour is not saturate enough.

Screenshot 2023-06-09 at 13 32 01

---

Just some ideas, like I say we could revisit this if we felt it was needed.

aliabid94:
Great work! It does throw me off a little is that there are two crosses. 
> Also I never really liked the big "X" button on the toast. I remember not even realizing it was a button at first, just thought it was an icon.

I had also felt this way, so I think it's definitely better to have an explicit close button, but maybe that can be the word 'close' like in pete's screenshots so there aren't two X's.

pngwn:
Love this! Could we figure out a way to make dark and light more more consistent? One is a lot more red than the other. I feel like the light mode on is more 'errory' than the dark mode one.

dawoodkhan82:
> Great work! It does throw me off a little is that there are two crosses.

I think changing the big "X" icon to an exclamation point icon could work

abidlabs (approved):
LGTM from my end, thanks @hannahblair!

aliabid94:
Looks great! 

I worry a little about how many theme variables will expand as a result of these, because we now have 10 theme variables for error, so we'll add 10 more for Warning and Info too? It's not a huge concern but we can't remove theme variables (for backwards compatibility) so it's something to keep in mind, cc @pngwn.

pngwn:
If they are necessary then there isn't much we can do if we want to keep things themeable, the shorthands and theme builder also simplify it somewhat but maybe we can figure out a way to reduce some of them or combine them in 4,0 (when we can remove stuff).

PR #4455, opened 2023-06-08
UI hotfixes for Examples and Dataframe

aliabid94:
Examples had picked up an ugly container from the styles deprecation PR:
Screen Shot 2023-06-08 at 2 12 53 PM
Dataframe had a margin top that made alignment weird - we use Row/Column padding and flex gaps for spacing between components everywhere else.


abidlabs:
Can confirm both of these issues were present and are now fixed. No side effects observed. LGTM

abidlabs (approved):

Issue #1719, opened 2022-07-06
Expose Reset/Clear button as a component


freddyaboulton:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
As a gradio user, I would like an easy way to add a clear button to a demo built with the blocks api. 

Right now, it's not clear how to reset components to their original states with blocks (see #1672). The right approach involves setting the `_js` parameter in a button update and that parameter is not documented anywhere. 

**Describe the solution you'd like**  
Add a `gr.Reset` component that is a short-hand for resetting a list of components to their original states. I think this demo adapted from #1672 showcases a possible api

```python
import gradio as gr

demo = gr.Blocks()

with demo:
    text = gr.Textbox()
    explanation = gr.HighlightedText()
    with gr.Row():
        submit_btn = gr.Button(value="Submit")
        reset = gr.Reset(components=[text, explanation])
    samples = gr.Dataset(
        components =[text],
        samples=[["Example of a text. This should reset highlighted text but does not."]]
    )
    submit_btn.click(
      fn = lambda text: gr.HighlightedText.update(value=[("Fake output", 1)]), inputs=[text], outputs=[explanation]
    )
    samples.click(lambda sample: gr.Textbox.update(value=sample[0]),
                  inputs=[samples],
                  outputs=[text])

demo.launch()
```

**Additional context**

fffiloni:
Subscribed to be notified when gr.Reset() is released :)

freddyaboulton:
I can pick this up after the bug bash sprint. Should be straight forward to do it on the backend.

PR #4456, opened 2023-06-08
Add a `gr.ClearButton`

abidlabs:
Adds a `gr.ClearButton` which can be used to clear the value of components in `gr.Blocks()`. It is also used internally in `gr.Interface`.

Example:

```py
import gradio as gr

with gr.Blocks() as demo:
    chatbot = gr.Chatbot([("Hello", "How are you?")])
    with gr.Row():
        textbox = gr.Textbox(scale=3, interactive=True)
        gr.ClearButton([textbox, chatbot], scale=1)
    
demo.launch()
```

Also removes the `cleared_value` attribute from components, which will be better for community components

**Discussion**: There was a discrepancy between the behavior of the Clear button for different components. For 4 of our components (`Slider`, `CheckboxGroup`, `Radio`, `Dropdown`), the Clear button set the value to the original default value set by the app developer. For all of our other components, such as `Textbox`, `Checkbox`, `Image`, it explicitly cleared the value (e.g. setting it to "" for `Textbox`, unchecked for `Checkbox`, or no uploaded image for `Image` even if there was originally a default). We should have a single behavior for all of our components -- I've set it so it always clears the value as this is what I think most people expect when they hit Clear.

Fixes: #1719
Fixes: #4473


gradio/components.py
```diff
@@ -4047,7 +4050,7 @@ def postprocess(self, y: dict[str, float] | str | float | None) -> dict | None:
             Object with key 'label' representing primary label, and key 'confidences' representing a list of label-confidence pairs
         """
         if y is None or y == {}:
-            return None
+            return {}
```
abidlabs, line 4053: Fixes: #4473

gradio/buttons.py
```diff
@@ -0,0 +1,61 @@
+""" Predefined buttons with bound events that can be included in a gr.Blocks for convenience. """
+
+from typing_extensions import Literal
+
+import json
+from gradio_client.documentation import document, set_documentation_group
+
+from gradio.components import Button, Component
+
+set_documentation_group("component")
```
abidlabs, line 10: Added to the component section for now. Once we have more buttons, we can have a dedicated section for Buttons in the docs page.

abidlabs:
Will fix the tests tomorrow, but everything is working locally so opening up for review!

js/form/src/Dropdown.svelte
```diff
@@ -23,12 +23,12 @@
 		blur: undefined;
 	}>();
 
-	let inputValue: string,
+	let inputValue: string | undefined,
```
abidlabs, line 26: This is needed so that the `gr.Dropdown` can handle `null` values properly. Previously, sending `null` to a `Dropdown` would cause the value to be subdued:

image

js/app/src/Blocks.svelte
```diff
@@ -259,15 +259,13 @@
 		};
 
 		if (dep.frontend_fn) {
-			console.log("running frontend fn");
```
abidlabs, line 262: Added accidentally in #4486

gradio/helpers.py
```diff
@@ -263,6 +263,7 @@ async def load_example(example_id):
                 show_progress="hidden",
                 postprocess=False,
                 queue=False,
+                api_name="load_example",
```
abidlabs, line 266: I added this in the process of fixing some tests. Makes testing more robust and can be useful for developers as well.

.vscode/settings.json
```diff
@@ -1,22 +0,0 @@
-{
```
freddyaboulton, line 1: how come this is being deleted?

gradio/buttons.py
```diff
@@ -0,0 +1,70 @@
...
+
+from gradio_client.documentation import document, set_documentation_group
+from typing_extensions import Literal
+
+from gradio.components import Button, Component
+
+set_documentation_group("component")
+
+
+@document("add")
```
aliabid94, line 15: why are we documenting `add`? Seems like an internal method.

freddyaboulton:
When filing #1719, I had this functionality in mind so it's ok to close it out with this PR.

gradio/buttons.py
```diff
@@ -0,0 +1,70 @@
...
+        """
+        if not components:
+            # This needs to be here because when the ClearButton is created in an gr.Interface, we don't
+            # want to create dependencies for it before we have created the dependencies for the submit function.
+            # We generally assume that the submit function dependency is the first thing created in an gr.Interface.
+            return self
+
+        if isinstance(components, Component):
+            components = [components]
+        clear_values = json.dumps(
```
aliabid94, line 66: clever use of postprocess

aliabid94 (approved):
LGTM

gradio/buttons.py
aliabid94: oh I guess if the cleared components come after huh

freddyaboulton: Yea but I wonder how common that is outside of the `interface.py` code

guides/07_other-tutorials/creating-a-chatbot.md
```diff
@@ -14,7 +14,7 @@ This tutorial will show how to make several kinds of chatbot UIs with Gradio: fi
 $demo_chatbot_streaming
 
 **Prerequisite**: We'll be using the `gradio.Blocks` class to build our Chatbot demo.
-You can [read the Guide to Blocks first](https://gradio.app/quickstart/#blocks-more-flexibility-and-control) if you are not already familiar with it.
+You can [read the Guide to Blocks first](https://gradio.app/quickstart/#blocks-more-flexibility-and-control) if you are not already familiar with it. Also please make sure you are using the **latest version** version of Gradio: `pip install --upgrade gradio`.
```
freddyaboulton, line 17: nit: This sentence in the guide is outdated now that the clear internal is abstracted away: 

```
Finally, when the `Clear` button is clicked, we assign `None` to the value of the `Chatbot` to clear its entire history.
```

freddyaboulton (approved):
Code for `ClearButton` looks great @abidlabs !

guides/07_other-tutorials/creating-a-chatbot.md
abidlabs: Ooh good catch

gradio/buttons.py
abidlabs: I don't have a strong opinion, but it might come in handy for complex demos so I've added a docstring and to the docs page.

gradio/buttons.py
abidlabs: Thx

abidlabs:
> @abidlabs I agree it's best if the Clear button behaves the same for all components but it's behaved this way for so long I wonder if users have come to expect the current behavior? Maybe it's safest to use the current clear button implementation in interface unitl 4.0?

I think it's okay because if they are looking for that functionality, they can always just refresh the page manually.

abidlabs:
in light of the above, I’ll go ahead and merge. Thanks @freddyaboulton @aliabid94 for the detailed reviews!

eshaanagarwal:
Hey has this been released as part of the new release of Gradio ?

abidlabs:
Not yet, soon!

.vscode/settings.json
pngwn: This was a mistake. It should not be gitignored or all of the extensions settings will be broken and will not work.

We've also lost the recommended extensions somewhere. I'm not sure why the `.vscode` was added to the gitignore in an unrelated PR months ago.

I'll add this back to get things working again.

Issue #2749, opened 2022-11-30
"Duplicate Space" & "Share to Community" buttons


gary149:
Implement the button directly in Gradio API.
`gr.ShareButton` by @merveenoyan could work well 👍 

![image](https://user-images.githubusercontent.com/3841370/204886429-01bfb364-52bf-4207-9aaf-f31f4473f62a.png)

abidlabs:
Thanks for the suggestion @gary149! Similar to #2269, we need to figure out the right way to share Gradio outputs in a general way.

merveenoyan:
I was wondering about whether to enable an option to share the output when creating the Interface or Blocks. If we know the app is launched from Spaces (which I don't know if it's the case) we could enable sharing the output button by default.

julien-c:
i would also support integrating it into Gradio

gary149:
I've been thinking about it a bit more and I think it makes sense to expand the issue to "Expose some Spaces options inside Gradio", I think it will really improve DX and UX.

<img width="1412" alt="image" src="https://user-images.githubusercontent.com/3841370/214537151-f53ed99e-3e08-4990-a9bb-0a3395882e02.png">

abidlabs:
cc @dawoodkhan82 @aliabid94 something to think about after the bug bashing sprint!

yvrjsharma:
This feature implementation can help us drive a good amount of growth cc: @AK391 @abidlabs

PR #4458, opened 2023-06-09
Duplicate spaces integration into gradio

aliabid94:
EDIT: Scroll below, PR revamped

This was the best integration of the Duplication workflow I could think of:

1. Add the Duplicate spaces link in the footer
Screen Shot 2023-06-08 at 5 01 52 PM

2. For any process with a queue > 3 and ETA > 10s shows a link in the status tracker to duplicate the space. 
Screen Shot 2023-06-08 at 5 03 15 PM

I think this is the most natural integration of duplication into Gradio, and requires nothing from our users.


aliabid94:
Oh wait this doesn't work - I thought gradio apps had access to the space name via the attribute. Fixing.

abidlabs:
Nice! Quick comments:

(1) I really like the "duplicate Space" in the footer
(2) The "duplicate Space to skip queue" text feels too clunky. Especially if there are multiple output components it would clutter the UI quite a bit. What if we just had the icon? If the same icon also appears at the bottom of the page, then I think people will quickly build an association that this means to duplicate the Space. You could maybe have the icon appear on the same line as the queue
(3) Speaking of the icon, it would be really nice if it matched the shape of the duplicate icon in Spaces. It's very close, but if we could add the additional circle, then the association would be even stronger:

image


cc @julien-c @gary149 @pngwn

js/app/src/Blocks.svelte
```diff
@@ -448,15 +449,25 @@
...
 						set_api_docs_visible(!api_docs_visible);
 					}}
-					class="show-api"
+					class="footer-link"
 				>
 					Use via API <img src={api_logo} alt="" />
 				</button>
 				<div>·</div>
 			{/if}
+			{#if true}
```
freddyaboulton, line 458: I think there should be a way to not display this by the app author. Some spaces are not meant to be duplicated (they use a custom backend like stable diffusion or whisper-jax which doesn't transfer on udplication) or app author just doesn't want to include it.

aliabid94: Added parameter to remove.

aliabid94:
> The "duplicate Space to skip queue" text feels too clunky

Simplified.

> Speaking of the icon, it would be really nice if it matched the shape of the duplicate icon in Spaces

Done.

Screen Shot 2023-06-09 at 4 45 20 PM

freddyaboulton (approved):
Tested and works and looks as expected in both light and dark mode. Nice @aliabid94 !

image

gradio/utils.py
```diff
@@ -107,8 +107,10 @@ def ipython_check() -> bool:
     return is_ipython
 
 
-def is_space() -> bool:
-    return os.getenv("SYSTEM") == "spaces"
+def get_space() -> str | None:
```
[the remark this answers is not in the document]
aliabid94: this will first check that the env variable "SYSTEM" is set to "spaces", which is the "official" indicator of whether we are in a space. I don't think it really matters but its slightly safer to first check the "system" variable first imo.

abidlabs:
Still think putting the word "Duplicate" and the icon underneath the Gradio loading logo looks crowded, particularly if the component has a small height or there are many output components. @aliabid94 what do you think about having just the duplication symbol appear next to the queue indicator, maybe right before the word "queue"? cc @gary149 

I think it's very important that we are not too intrusive as we are significantly changing the UI on Gradio developers -- in a way that they might not expect as it only appears if they are on Spaces & have traffic. In fact, we should almost certainly have a way for developers to disable this feature if they find it too intrusive. Can we add a `suggest_duplication` flag in the `launch()` method? 

Was discussing with @julien-c earlier about this approach vs. a having a `gr.DuplicateButton()` that a developer can manually put in the UI wherever they prefer. Overall, I think the approach in this PR is better as (1) the UI more cleanly integrates into the Gradio app (2) is *opt-out* vs. *opt-out* so uptake of the feature will be much higher

julien-c:
for me it's better if it's opt-in by the developer (and thus, located in a more prominent place)

Here in the footer it feels redundant with the Duplicate this Space modal in the hf.co "chrome" which is what we wanted to prevent.

Very much in favor of doing a `gr.DuplicateButton()` and a `gr.ShareButton`

abidlabs:
We discussed this with @aliabid94 and the team, and we'll make the following changes:

* We remove `Duplicate Space` from the footer and from the status tracker
* If there is a long queue, we instead show an "info" modal that suggests that a user can duplicate the queue (this is an opt-out feature)
* We also release a `gr.DuplicateButton` which a user can position wherever they want in a `gr.Blocks()` (that is an opt-in feature)
* To allow the same functionality for `gr.Interface`, we'll add a keyword arg that a user can use to add this "Duplicate Button" in their Interface

aliabid94:
This PR is ready for a re-review (cc: @abidlabs  @freddyaboulton). Now:
- There is a `gr.Duplicate` button, which will be invisible unless on spaces
- Interfaces automatically have a Duplicate Button
- On Spaces, if there is a queue >= 3 and ETA > 15s, a modal pops up to suggest Duplication to skip the queue. Currently this modal looks like an Error, so this PR should be merged after gr.Info and gr.Warning work is complete on the frontend (cc: @hannahblair)

This also fixes an issue where a standalone button with `scale=0` in a Column was taking full width. Fixed that so duplication button looks like this:

In Blocks:
Screen Shot 2023-06-19 at 10 47 56 PM

In Interface:
Screen Shot 2023-06-19 at 10 50 23 PM

abidlabs:
Very cool @aliabid94! Have some suggestions on the UI/UX here:

1. Would it be possible to make the `gr.DuplicateButton()` a little smaller and a little nicer looking? Something closer to what @gary149 mocked up in https://github.com/gradio-app/gradio/issues/2749 -- one way we could implement it would be to allow gr.Button() to take in paths to icons and using `size="sm"`? 
2. Instead of making the `gr.DuplicateButton()` invisible unless on Spaces, I suggest making it not do anything (or raise an  a modal saying that it only has an effect if on Spaces). That way, there is less of a UI gap between local and Spaces
3. I really *don't* think we should have it show up by default on `gr.Interface`. We're already showing the duplication modal on Spaces, so I think this will be too much of a push to duplication all at once

aliabid94:
> 1. Would it be possible to make the `gr.DuplicateButton()` a little smaller and a little nicer looking? Something closer to what @gary149 mocked up in https://github.com/gradio-app/gradio/issues/2749 -- one way we could implement it would be to allow gr.Button() to take in paths to icons and using `size="sm"`? 

Tricky because now we need to consider how a new button variant works with themes and also currently do not support images in buttons. Something for a future PR.

> 2. Instead of making the `gr.DuplicateButton()` invisible unless on Spaces, I suggest making it not do anything (or raise an  a modal saying that it only has an effect if on Spaces). That way, there is less of a UI gap between local and Spaces

Done

> 3. I really *don't* think we should have it show up by default on `gr.Interface`. We're already showing the duplication modal on Spaces, so I think this will be too much of a push to duplication all at once

Done

Issue #2687, opened 2022-11-21
Make gr.Error `position: Fixed` in an iframe


apolinario:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
Currently gr.Error is on the top-right. However if your application scrolls down, the user can never read what error it was unless they scroll back up
![image](https://user-images.githubusercontent.com/788417/203115423-55e97884-fb17-4e5b-86ff-0f0a28a9f721.png)

**Describe the solution you'd like**  
Have `gr.Error` to have a `position: Fixed` or analogous to show on the top-right regardless of scroll

abidlabs:
Makes sense, thanks for the nice suggestion @apolinario!

abidlabs:
Hmm interesting I just tested this out and for me, locally the `gr.Error()` appears in a fixed position. However, when embedded in Spaces, it does not appear in fixed place.

This seems like a limitation of iframes, according to https://stackoverflow.com/questions/9578364/absolute-position-elements-within-iframe
> An element within an iFrame cannot be positioned relative to an element in the parent document. iFrames and their parent windows don't share resources such as CSS (and therefore also don't share CSS rules). There is no pure CSS way to accomplish this.

cc-ing @pngwn, do you know if this is doable?

Here's my code to reproduce:

```py
import gradio as gr
import time

def func(x):
    time.sleep(2)
    raise gr.Error("abc")
    return x
    
md = "a<br>"*40

with gr.Blocks() as demo:
    a = gr.Markdown(md)
    b = gr.Textbox()
    demo.load(func, a, b)

demo.launch()
```

pngwn:
Sorry, missed this. This isn't possible because of how the iFrame is resized/ scrolled as far a I'm aware but we may be able to improve it somewhat by scrolling to the correct location if there is an error.

pngwn:
We can discuss this today but this is really problematic and the fixes are not satisfactory. I'll be testing a few ideas but testing them might be tedious.

PR #4459, opened 2023-06-09
improve error UX

### LINKED ISSUE #3347: Let any function raise a UI error

Current in order to show a UI Error modal one of the output components needs to have a status tracker, which we use internally to show loading progress. This make sense to some degree but introduces limitations.

Decouple the error modal from the status tracker and allow any component to show it via `gr.Error`

**Additional context**  
https://github.com/gradio-app/gradio/pull/3289#discussion_r1119580630

pngwn:
# Description

Closes #3347. 

This PR changes the error modal behaviour. 

- Any component can generate an error, regardless of whether or not it has a status tracker. Even if a user disables a status tracker (or if a custom component doesn't have one in the future), the error modal can still be triggered by raising a `gr.Error`
- More than one error can be displayed at a time (and potentially other notifications in the future). The Error modal is now part of a toast display, the errors queue up in the order they are triggered. There can never be more than one error for a single function index.
- Error modals will be dismissed after a time limit, currently set to 10 seconds. In order to communicate this to the user a timer bar has been added (to the bottom). We can iterate on this but i think it works well + looks decent for now.
- On mobile the toasts pop in from the bottom, rather than the top (the order is reversed too). This makes them slightly less likely to cover the whole screen (although this is unavoidable to a degree) and makes them more convenient to dismiss.

You can test this behaviour with the following code:

```py
import gradio as gr
import time


def one(text):
    raise gr.Error("This is the first error")


def two():
    time.sleep(1)


def three():
    raise gr.Error("This is the second error")


def four():
    time.sleep(1)


def five():
    raise gr.Error(
        "This is the third error, and it is much longer than the others. In fact it is much, much longer."
    )


def six():
    time.sleep(1)


def seven():
    raise gr.Error("This is the fourth error, and it is a conventional length.")


with gr.Blocks() as demo:
    input = gr.TextArea()
    output = gr.TextArea()
    gr.Button().click(fn=one, inputs=input, outputs=output).then(two).then(three).then(
        four
    ).then(five).then(six).then(seven)

demo.launch()
```

Some screencaps:

**Desktop**

https://github.com/gradio-app/gradio/assets/12937446/8f0f1a8d-c4b9-4dae-9326-e43289bbec17

**Mobile (ish)**

https://github.com/gradio-app/gradio/assets/12937446/b5e4ceb1-1648-4674-8ef2-301ac208fa0b


pngwn:
@hannahblair Sorry about the overlap, I didn't realise you were also working on error modals. We can figure out the conflicts. We should probably get one or the other merged in soonish to make life easier.

js/app/src/Blocks.svelte
```diff
@@ -304,6 +313,21 @@
 					}
 
 					if (status.stage === "error") {
+						if (status.message) {
+							console.log();
```
pngwn, line 317: Remove.

demo/audio_component/run.py
```diff
@@ -1,6 +1,8 @@
 import gradio as gr
 
-css = "footer {display: none !important;} .gradio-container {min-height: 0px !important;}"
+css = (
+    "footer {display: none !important;} .gradio-container {min-height: 0px !important;}"
+)
```
pngwn, line 5: Revert.

js/app/src/Blocks.svelte
```diff
@@ -304,6 +313,21 @@
...
+							messages = [
+								{
+									type: "error",
+									message: status.message.substring(
```
freddyaboulton, line 321: This is causing the first and last characters of the error message (not quotes) to be stripped if the user doesn't raise a `gr.Error`. I made the following modifications to your demo (load event error + show_error=True):

```python
def load_error():
    raise ValueError("This is an error!")

with gr.Blocks() as demo:
    input = gr.TextArea()
    output = gr.TextArea()
    gr.Button().click(fn=one, inputs=input, outputs=output).then(two).then(three).then(
        four
    ).then(five).then(six).then(seven)
    demo.load(load_error, None, None)

demo.launch(show_error=True)
````

Screenshot 2023-06-09 at 1 36 22 PM

Other than that this LGTM @pngwn !

pngwn: I've adjusted this to use regex-based replacement. It will now only trim if the message begins and ends with a single quote.

pngwn:
Thanks for the reviews @hannahblair and @freddyaboulton. Implemented all suggestions and will merge when CI is green (and I've done one last test).

pngwn:
This doesn't close #2687. I'm going to look at that separately. Scrolling the page didn't work, it put the page into a really bizarre state.

Issue #4359, opened 2023-05-30
Gradio 4.0 Tracking Issue


abidlabs:
**Tracking Issue for Gradio 4.0** (see [milestone](https://github.com/gradio-app/gradio/milestone/14))

**Goals**
* Launch 4.0 with support for community components
* Launch quickly (speed & experimentation, not perfection)
* Minimize breaking changes in 4.0 as much as possible so users can easily upgrade
* Have no need for breaking changes from 4.0 to 4.x

**TODOs**

_Build End-to-End Test Suite_
- [x] Get the kitchen sink test suite working without mocks (@pngwn)
- [x] Brainstorm various paths (queue vs. no queue, mount vs. no mount, then(), cancels, every, embedded demos) and write tests for them (@dawoodkhan82)
- [x] Reduce CI time to <10 minutes (@freddyaboulton)

_Strengthening Core Components_  (@hannahblair for frontend, @abidlabs for backend)

- [x] Go through all 33 components one-by-one and fix bugs so that the components serve as good templates for community components. Be good to take an accessibility pass while we’re doing this, as well as organize code for readability
- [x] Known dataframe improvements: #2209, #2666, #3284, #3286
- [x] Consolidate Dataframe and Dataset components - mentioned in slack and noting here so we don't forget. Can remove if we disagree
- [x] Improve names of parameters - make breaking changes where necessary. For example in #4470, we noted that the `preview` and `enable_preview` parameters (both `bool`) of Gallery are confusing and kind of overlap.

_Refactor components.py to simplify creation of new components_ (@aliabid94 @freddyaboulton)

- [x] simplify components.py to have just a couple of functions
- [x] Make it so that examples can accept any value in preprocess
- [x] Remove as_example(), style(), interpetation, `IOComponent`
- [x] Simplify `update()` so it doesn't need to be explicitly written for every component
- [x] See #3438 

_Standardize the data types between frontend/backend_ (@freddyaboulton)

- [x]  Use dataclasses instead of dictionaries
- [x] Consistency between components
- [x] Fix API for Image component
- [x] Fix API for File component
- [x] Fix API for all JSONSerializable components
- [x] Use the same dataclass/schema for different components where possible

_Creating community components locally_ (@freddyaboulton @pngwn)

- [x] Define workflow for creating components
- [x] Be able to load a folder containing py/js (register custom components)
- [x] Frontend decides how to display component examples

_Dev environment for local components_ (@freddyaboulton @pngwn)

[...]

PR #4477, opened 2023-06-11
Knob component

2-5:
WIP.

I think gradio could use a `Knob` component for dense UIs, where sliders use too much space. I implemented one, and would like your feedback.

It's based on the `Slider` component, the Python code is copy/paste, This raises the question if `Knob` should be just a `Slider` variant, something like `Slider(style="knob")`.

You might find it weird that the knob needs to be dragged up/down to change value, instead of circularly. This is not a mistake, this is the correct movement for such a component. All audio applications default to this style (some allow a circular movement option).

Movement is handled with pointer events, which have a number of issues. In particular the mouse cursor moves as you drag it, and it might bump into a screen edge where further movement is blocked. All issues can be fixed by using the [Pointer Lock API](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_Lock_API), which allows you to hide the mouse cursor and keep it in place while the knob is grabbed. But this API has a downside - it might prompt a browser permission request. Maybe it could be an option - `Knob(pointer_capture=True)`

It looks a bit weird to see the value two times, once in the center of the knob, and then below it in the number input control. The numerical input control is probably required for accessibility. So maybe there could be an option to hide the center value display.

Demo video:

https://github.com/gradio-app/gradio/assets/3753822/00c99ef4-66af-4d9f-baa7-9f5a51752d53

# Checklist:


abidlabs:
Hi @2-5 thanks for the cool PR! We are actually in the process of refactoring the library for [4.0 release](https://github.com/gradio-app/gradio/issues/4359) so we are not doing major additions to the library. So I'll go ahead and close this PR at the moment. That being said, one of the goals of the 4.0 release is the ability to easily support components created by the community. So you should be able to take this code and hopefully without too many changes, create a community component from it once we release 4.0.

PR #4481, opened 2023-06-12
Fix docstrings in HTML Component and missing @document 

aliabd:
@aliabid94 this PR added `scale` and `min_width` to the docstrings for the `HTML` `__init__`, which doesn't have those params and it broke docs on the website: https://github.com/gradio-app/gradio/pull/4374


aliabd:
also @aliabid94 you removed the `@document()` from many of the components when i think you meant to just remove 'style' from inside it. This will make so many components disappear from the docs, will add them back in this PR

Issue #4485, opened 2023-06-12
`_js` does not update outputs for certain return values


abidlabs:
### Describe the bug

If the return value from your function does not contain any dictionary updates, then `_js` will not actually update the output components. Here are a couple of examples:

```py
import gradio as gr

with gr.Blocks() as demo:
    a = gr.Textbox()
    b = gr.Textbox()
    a.change(lambda x:x, a, b)
    c = gr.Button()
    c.click(None, [], [a, b], _js='() => ["", ""]')

demo.launch()
```

```py
import gradio as gr


with gr.Blocks() as demo:
    a = gr.Textbox()
    c = gr.Button()
    c.click(None, None, a, _js="() => {console.log('abc'); return 'abc';}")

demo.launch()
```

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

See above 

### Logs

```shell
Nothing in the console
```

### System Info

```shell
gradio==3.34
```

### Severity

blocking all usage of gradio

PR #4486, opened 2023-06-12
Fix JS call

aliabid94:
the `_js` parameter would not show updates to components in the cases where both:
- there was no accompanying backend python function
- all updated values were values only, and not `gr.update` JSONs

The reason for this is the line to cause svelte to re-render (`root_node = root_node`) was for some reason only in the code when `gr.update` were loaded, or there was an update to a status-tracker. All backend python functions make status-tracker updates, so if there's any python function then it always updates. Simply moving the re-render line to run after any update should fix this. Test with

```python
import gradio as gr

with gr.Blocks() as demo:
    a = gr.Textbox()
    c = gr.Button()
    c.click(None, None, a, _js="() => {console.log('abc'); return ['abc'];}")

demo.launch()
``` 

Closes: #4485


pngwn:
I probably broke this.

pngwn (approved):
Code looks good. Can't test right now.

freddyaboulton:
I was able to test and it works so if @pngwn says the code looks good then this can be merged once CI passes @aliabid94 !

abidlabs:
I'll merge this in so that I can use it for #4456. Quick turnaround!

Issue #4495, opened 2023-06-13
Allow specifying the download directory for `gradio_client.Client`


hysts:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  

Related to https://github.com/gradio-app/gradio/issues/4494

Files under `/tmp` will be deleted, so we need to move them if we want to keep them. It would be useful if it was possible to specify download directory.

freddyaboulton:
Makes sense @hysts !

Issue #4494, opened 2023-06-13
Output images are downloaded to the current directory when using `gradio_client` for `gr.Gallery`


hysts:
### Describe the bug

Not sure if this is a bug or as intended, but `gradio_client.Client` saves output images of `gr.Gallery` in a directory under the current directory, while files are downloaded to a directory under `/tmp` for `gr.Audio`, `gr.Model3D`, and `gr.Video`, and `gr.Image` is downloaded to `/tmp`.

I don't like this behavior because it clutters my working directory and because it's not easy to delete generated directories as they have random names generated by `uuid.uuid4()`.

It would be nice if the outputs of `gr.Gallery` are also downloaded to a directory under `/tmp`.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```python
from gradio_client import Client

client = Client('stabilityai/stable-diffusion')
out = client.predict('an astronaut riding a horse', '', 9, fn_index=1)
```

### Logs

```shell
n/a
```

### System Info

```shell
gradio==3.34.0
gradio_client==0.2.6
```

### Severity

annoying

abidlabs:
Looks like a bug indeed!

abidlabs:
This is why the working directory gets cluttered when running the pytests locally as well, which has been annoying me for a while. Will fix!

PR #4501, opened 2023-06-13
Fix output directory of files in client & when calling Blocks as function

abidlabs:
Previously, different serializers saved output files in inconsistent places. This meant that if you used the client to do a prediction, it could save output files in various places -- including the working directory for the `Gallery` component (see #4494). This fixes that -- all of the serializers save output files to a folder created within the `GRADIO_TEMP_DIR` directory.

Test with:

```py
import gradio as gr

demo = gr.Interface(
    lambda x:["cheetah.jpg"],
    inputs=[gr.Slider(value=3, minimum=1, maximum=3, step=1)],
    outputs=[gr.Gallery(columns=2, preview=True)],
)

demo(4)
```

```py
demo = gr.Interface(
    lambda x:"cheetah.jpg",
    inputs=[gr.Slider(value=3, minimum=1, maximum=3, step=1)],
    outputs=[gr.Image()],
)

demo(4)
```

and so on.

Furthermore, if you are using the python `Client`, you can specify the output directory with a new `output_dir` parameter. 

Closes: #4494 
Closes: #4495


test/test_blocks.py
```diff
@@ -148,7 +148,7 @@ def update(name):
 
             inp.submit(fn=update, inputs=inp, outputs=out, api_name="greet")
 
-            gr.Image().style(height=54, width=240)
+            gr.Image(height=54, width=240)
```
abidlabs, line 151: Just updates the syntax to remove the deprecated `.style()` function

freddyaboulton:
@abidlabs is this ready? Got tagged for review but it's still a draft. Quick turnaround!

abidlabs:
Just adding tests @freddyaboulton! Will open up soon

client/python/gradio_client/utils.py
```diff
@@ -383,8 +383,8 @@ def decode_base64_to_file(
     dir: str | Path | None = None,
     prefix: str | None = None,
 ):
-    if dir is not None:
-        os.makedirs(dir, exist_ok=True)
+    directory = Path(dir or tempfile.gettempdir()) / secrets.token_hex(20)
```
freddyaboulton, line 386: I'm not sure we need the `secrets.token_hex(20)` here since the use of tempfile should already create a unique name.

freddyaboulton (approved):
Looks good @abidlabs !

client/python/gradio_client/utils.py
abidlabs: You're right, but I did it this way just to have consistent behavior for all of the components.

abidlabs:
Thanks @freddyaboulton for the review! Will address the suggestions tomorrow

abidlabs:
I'll need to release a new version of `gradio_client` for the regular tests to pass

gradio/blocks.py
```diff
@@ -1135,7 +1141,10 @@ def deserialize_data(self, fn_index: int, outputs: list[Any]) -> list[Any]:
                 block, components.IOComponent
             ), f"{block.__class__} Component with id {output_id} not a valid output component."
             deserialized = block.deserialize(
-                outputs[o], root_url=block.root_url, hf_token=Context.hf_token
+                outputs[o],
+                save_dir=DEFAULT_TEMP_DIR,
```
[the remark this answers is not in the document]
abidlabs: Nice

PR #4510, opened 2023-06-14
Fix demos on the landing page

aliabd:
Demos on the landing page were broken for many different reasons, this PR fixes them. Tested already on spaces so you can see them working fine now.


freddyaboulton (approved):
Thank you @aliabd !

Issue #4428, opened 2023-06-05
Dollar signs automatically render latex - can happen unintentionally


aliabid94:
Should figure out a better way to explicitly render latex in Chatbot/Markdown, currently can easily happen accidentally

freddyaboulton:
Didn’t this get fixed by https://github.com/gradio-app/gradio/issues/4143

pseudotensor:
How do I disable latex processing for now?  It's very intrusive, and my main use case never involves latex.

pseudotensor:
Found: https://github.com/gradio-app/gradio/issues/4143#issuecomment-1542315064

abidlabs:
A simpler solution would be to just have a flag that enables/disables LaTeX. Not as fully featured, but would probably be a simpler DX. cc @dawoodkhan82

PR #4516, opened 2023-06-14
Adds Latex Delimiters Parameter

dawoodkhan82:
# Description
Adds `enable_latex` and `latex_delimiters` parameters.

Please include: 
* relevant motivation
* a summary of the change 
* which issue is fixed. 

Closes: #4428

# Checklist:

# A note about the CHANGELOG


dawoodkhan82:
@abidlabs lmk if you think adding both params is unnecessary. Maybe we could use passing `None` into `latex_delimiters` as a way to turn off latex rendering.

abidlabs:
Nice! Yeah, I think both is unnecessary. You could just pass in an empty list to `latex_delimiters` to indicate no LaTeX should be rendered. (We should add this to the docstring for `latex_delimiters`)

Also, we should mark this as a breaking change in the changelog as now by default we are only rendering LaTeX if enclosed with double $ signs.

gradio/components/chatbot.py
```diff
@@ -38,6 +38,8 @@ def __init__(
         | Callable
         | None = None,
         color_map: dict[str, str] | None = None,  # Parameter moved to Chatbot.style()
+        enable_latex: bool = True,
+        latex_delimiters: list[dict[str, str | bool]] | None = [{"left": "$$", "right": "$$", "display": True}],
```
abidlabs, line 42: Just wondering, do we need the `display` parameter? Confused by what it does

abidlabs: I would personally prefer just a list of strings if possible. Like: `["$", "$$"]` to mean that either single dollar signs of double dollar signs get converted to latex

abidlabs: I know some delimeters are not symmetric, but in that case, we could just accept the "open" delimeters and take care of everything else ourself. Would be better DX than having to deal with this nested dict

gradio/components/chatbot.py
```diff
@@ -54,6 +56,8 @@ def __init__(
         """
         Parameters:
             value: Default value to show in chatbot. If callable, the function will be called whenever the app loads to set the initial value of the component.
+            enable_latex: If True, will render LaTeX expressions enclosed in $$ delimiters or custom delimiters passed in `latex_delimiters` parameter. True by default.
+            latex_delimiters: A list of dicts of the form {open_delimiter, close_delimiter, display} that will be used to render LaTeX expressions. By default this is set to `[{ left: "$$", right: "$$", display: true }]`, so LaTeX expressions enclosed in $$ delimiters will be rendered.
```
abidlabs, line 60: I think it would be explain this needs to be a subset of KaTeX delimeters and maybe link to their docs

dawoodkhan82:
Display false:
Screenshot 2023-06-14 at 3 14 01 PM

Display True:
Screenshot 2023-06-14 at 3 13 44 PM

Basically inline or not

dawoodkhan82:
> You could just pass in an empty list to latex_delimiters to indicate no LaTeX should be rendered. (We should add this to the docstring for latex_delimiters)

Should it also accept `None` so either empty list or `None` can disable latex.

abidlabs:
@dawoodkhan82 this looks good. However, it's strongly recommended[ not to use lists (or other mutables) as argument defaults.](https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments) 

So I'm going to change this to set the default value of `latex_delimiters` to `None`, which internally will be set to `[{"display": True, "left": "$$", "right": "$$"}]`. This means that we'll have to reverse this change:

> Should it also accept None so either empty list or None can disable latex.

So now if you want to disable LaTeX altogether, you need to pass in an empty list

gradio/components/chatbot.py
```diff
@@ -64,6 +65,7 @@ def __init__(
             elem_id: An optional string that is assigned as the id of this component in the HTML DOM. Can be used for targeting CSS styles.
             elem_classes: An optional list of strings that are assigned as the classes of this component in the HTML DOM. Can be used for targeting CSS styles.
             height: height of the component in pixels.
+            latex_delimiters: A list of dicts of the form {open_delimiter, close_delimiter, display} that will be used to render LaTeX expressions. Display set to `False` means LaTeX will be rendered inline. If not provided, `latex_delimiters` is set to `[{ left: "$$", right: "$$", display: true }]`, so only LaTeX expressions enclosed in $$ delimiters will be rendered. Pass in an empty list to disable LaTeX rendering. For more information, see the [KaTeX documentation](https://katex.org/docs/autorender.html).
```
abidlabs, line 68:
```suggestion
            latex_delimiters: A list of dicts of the form {"left": open_delimiter, "right": close_delimiter, "display": display_in_newline} that will be used to render LaTeX expressions. Setting "display" to `False` means LaTeX will be rendered inline instead of in a newline. If not provided, `latex_delimiters` is set to `[{ "left": "$$", "right": "$$", "display": True }]`, so only LaTeX expressions enclosed in $$ delimiters will be rendered, and in a new line. Pass in an empty list to disable LaTeX rendering. For more information, see the [KaTeX documentation](https://katex.org/docs/autorender.html).
```

gradio/components/chatbot.py
```diff
@@ -64,6 +65,7 @@ def __init__(
             elem_id: An optional string that is assigned as the id of this component in the HTML DOM. Can be used for targeting CSS styles.
             elem_classes: An optional list of strings that are assigned as the classes of this component in the HTML DOM. Can be used for targeting CSS styles.
             height: height of the component in pixels.
+            latex_delimiters: A list of dicts of the form {"left": open_delimiter, "right": close_delimiter, "display": display_in_newline} that will be used to render LaTeX expressions. Setting "display" to `False` means LaTeX will be rendered inline instead of in a newline. If not provided, `latex_delimiters` is set to `[{ "left": "$$", "right": "$$", "display": True }]`, so only LaTeX expressions enclosed in $$ delimiters will be rendered, and in a new line. Pass in an empty list to disable LaTeX rendering. For more information, see the [KaTeX documentation](https://katex.org/docs/autorender.html).
```
abidlabs, line 68:
```suggestion
            latex_delimiters: A list of dicts of the form {"left": open delimiter (str), "right": close delimiter (str), "display": whether to display in newline (bool)} that will be used to render LaTeX expressions. Setting "display" to `False` means LaTeX will be rendered inline instead of in a newline. If not provided, `latex_delimiters` is set to `[{ "left": "$$", "right": "$$", "display": True }]`, so only LaTeX expressions enclosed in $$ delimiters will be rendered, and in a new line. Pass in an empty list to disable LaTeX rendering. For more information, see the [KaTeX documentation](https://katex.org/docs/autorender.html).
```

abidlabs (approved):

js/chatbot/src/ChatBot.svelte
```diff
@@ -19,6 +19,11 @@
 	let old_value: Array<
 		[string | FileData | null, string | FileData | null]
 	> | null = null;
+	export let latex_delimiters: Array<{
+		left: string;
+		right: string;
+		display: boolean;
+	}> = [{ left: "$$", right: "$$", display: true }];
```
pngwn, line 26: Ditto for here.

pngwn (approved):
Few nits, but lgtm!

abidlabs:
Tested with:

```py
import gradio as gr

latex = """
I want to eat $3.4 worth of bananas and $2 worth of apples. The equation for the sum is $$3.4+2=5.4$$ thanks!
"""

message = [(
    "write a python program",
    latex
), (
    "write a python program",
    latex
)]

with gr.Blocks() as demo:
    gr.Chatbot(
        message
    )
    
demo.launch()
```

And looks good to me!

pngwn:
Awesome, thanks @abidlabs!

PR #4515, opened 2023-06-14
Conditionally run backend ci

freddyaboulton:
# Description

Only run the python client tests if the client files have been modified. Only run the gradio lib tests if the backend or python client have been modified. 

Should help reduce time-to-merge for PRs that do not modify the backend. 

Did a minor refactor of `backend.yml`, just moved the `lint` job to be on top of `test` (same for client) since the tests depend on the lint that makes more sense to me.

Downside is that we may have to remove tests from the branch protection rules.

# A note about the CHANGELOG

Hello 👋 and thank you for contributing to Gradio!

All pull requests must update the change log located in CHANGELOG.md, unless the pull request is labeled with the "no-changelog-update" label.

Please add a brief summary of the change to the Upcoming Release > Full Changelog section of the CHANGELOG.md file and include
a link to the PR (formatted in markdown) and a link to your github profile (if you like). For example, "* Added a cool new feature by `[@myusername](link-to-your-github-profile)` in `[PR 11111](https://github.com/gradio-app/gradio/pull/11111)`".

If you would like to elaborate on your change further, feel free to include a longer explanation in the other sections.
If you would like an image/gif/video showcasing your feature, it may be best to edit the CHANGELOG file using the 
GitHub web UI since that lets you upload files directly via drag-and-drop.

PR #4517, opened 2023-06-14
Only run install steps in backend ci if not previously cached

freddyaboulton:
# Description

- Sets up a cache of the python dependencies and front-end build so we can skip having to reinstall all dependencies for python tests
- Uses virtualenv and then caches that env with all the dependencies. Since the command for activating a virtual env is different between windows and linux, we need to add separate job steps for windows vs linux 😓 
- Removed the lint job and just run it before the linux tests. Too much overhead of installing python + deps in a separate job. Lint runs pretty fast on linux. Not linting windows because it should match linux result and linting on windows GH runners is really slow.
- Windows machines on github actions are just really slow 😭 

### With cache
image

### Without cache
image

About a 2.x speedup! 

# Checklist:

# A note about the CHANGELOG


abidlabs:
Just to confirm, this is not WIP anymore right @freddyaboulton and is ready for review?

freddyaboulton:
@abidlabs - please review! I think it's finally working in a stable manner. Might keep open for a couple of more days just to be extra sure.

.github/workflows/backend.yml
```diff
@@ -86,37 +62,56 @@ jobs:
...
+        python -m virtualenv venv
+    - uses: actions/cache@master
+      id: cache
+      with:
+        path: |
+          client/python/venv/*
+        key: python-client-${{ runner.os }}-pip-${{ hashFiles('client/python/requirements.txt') }}-${{ hashFiles('client/python/test/requirements.txt') }}
+        restore-keys: |
+          python-client-${{ runner.os }}-pip-
+          python-client-${{ runner.os }}-
```
abidlabs, line 77: Do we we want to include `restore-keys`? My understanding is that GH will look for caches with these keys if if the primary key is not found. But we don't want that, right?

.github/workflows/backend.yml
```diff
@@ -86,37 +62,56 @@ jobs:
...
+    - name: Install Test Requirements (Linux)
+      if: steps.cache.outputs.cache-hit != 'true' && runner.os == 'Linux'
+      run: |
+        . venv/bin/activate
+        python -m pip install -r test/requirements.txt
+    - name: Install Client Library (Linux)
+      if: runner.os == 'Linux'
```
abidlabs, line 84: This does not need to be run if the cache is found, right? 
```suggestion
      if: steps.cache.outputs.cache-hit != 'true' and runner.os == 'Linux'
```

.github/workflows/backend.yml
```diff
@@ -86,37 +62,56 @@ jobs:
...
+        python -m pytest -m "${{ matrix.test-type }}"
+    - name: Install Test Requirements (Windows)
+      if: steps.cache.outputs.cache-hit != 'true' && runner.os == 'Windows'
+      run: |
+        venv\Scripts\activate
         pip install -r test/requirements.txt
-    - name: Tests
-      shell: bash
+    - name: Install Client Library (Windows)
+      if: runner.os == 'Windows'
```
abidlabs, line 104: Same as above I think
```suggestion
      if: steps.cache.outputs.cache-hit != 'true' and runner.os == 'Windows'
```

.github/workflows/backend.yml
```diff
@@ -135,34 +130,82 @@ jobs:
...
+        python -m virtualenv venv
+    - uses: actions/cache@v3
+      id: cache
+      with:
+        path: |
+          venv/*
+        key: gradio-lib-${{ runner.os }}-pip-${{ hashFiles('client/python/requirements.txt') }}-${{ hashFiles('requirements.txt') }}-${{ hashFiles('test/requirements.txt') }}
+        restore-keys: |
+          gradio-lib-${{ runner.os }}-pip-
+          gradio-lib-${{ runner.os }}-
```
abidlabs, line 145: Same as above. Do we want these restore keys?

.github/workflows/backend.yml
```diff
@@ -135,34 +130,82 @@ jobs:
...
-      run: |
-        pip install -r test/requirements.txt
     - name: Build frontend
-      shell: bash
+      if: steps.frontend-cache.outputs.cache-hit != 'true'
       run: |
         pnpm i --frozen-lockfile
         pnpm build
+    - name: Install Gradio (Linux)
+      if: runner.os == 'Linux'
```
abidlabs, line 169: Same as above
```suggestion
      if: steps.cache.outputs.cache-hit != 'true' and runner.os == 'Linux'
```

.github/workflows/backend.yml
```diff
@@ -135,34 +130,82 @@ jobs:
...
+        . venv/bin/activate
+        bash scripts/type_check_backend.sh
+    - name: Run tests (Linux)
+      if: runner.os == 'Linux'
+      run: |
+        . venv/bin/activate
+        python -m coverage run -m pytest -m "${{ matrix.test-type }}" --ignore=client
+        python -m coverage xml
+    - name: Install Gradio (Windows)
+      if: runner.os == 'Windows'
```
abidlabs, line 197: Same as above
```suggestion
      if: steps.cache.outputs.cache-hit != 'true' && runner.os == 'Windows'
```

.github/workflows/backend.yml
```diff
@@ -39,32 +39,8 @@ jobs:
             - '.github/**'
           scripts:
             - 'scripts/**'
-  client-lint:
-    needs: changes
-    if: needs.changes.outputs.python-client == 'true' ||  needs.changes.outputs.workflows == 'true'
```
abidlabs, line 44: Why did we get rid of the "if"s from all of here and below? Won't that undo all the gains we get from #4515

.github/workflows/backend.yml
```diff
@@ -86,37 +62,56 @@ jobs:
           client/python/requirements.txt
```
abidlabs, line 62: Do we still need lines 60-64?

.github/workflows/backend.yml
```diff
@@ -86,37 +62,56 @@ jobs:
...
+          client/python/venv/*
+        key: python-client-${{ runner.os }}-pip-${{ hashFiles('client/python/requirements.txt') }}-${{ hashFiles('client/python/test/requirements.txt') }}
+        restore-keys: |
+          python-client-${{ runner.os }}-pip-
+          python-client-${{ runner.os }}-
+    - name: Install Test Requirements (Linux)
+      if: steps.cache.outputs.cache-hit != 'true' && runner.os == 'Linux'
+      run: |
+        . venv/bin/activate
+        python -m pip install -r test/requirements.txt
```
abidlabs, line 82: One idea that might offer some speedup is to also install `python -m pip install -r requirements.txt` in this step. That way, the `python -m pip install -e .` step (which can't be cached) would still run faster. 

Will probably make a bigger difference for the core `gradio` library, not so much for the client.

abidlabs (approved):
LGTM @freddyaboulton! Very nice. Left some clarifications/suggestions above (particularly around reversing the logic in #4515), but otherwise the logic looks good to me overall. I also inspected the jobs that were run by the CI and they all look good.

I think we can merge this in, and then all of us inspect the CI jobs that run in the next few days to make sure that changes that should bust the cache actually do bust the cache

.github/workflows/backend.yml
freddyaboulton: I think so. From what I read here: https://github.com/actions/setup-python#caching-packages-dependencies, this caches the package download (not the install), so it can still be useful when we have to rebuild the env when there's a cache miss.

.github/workflows/backend.yml
freddyaboulton: Yep you're right! Let me add back

.github/workflows/backend.yml
freddyaboulton: I'm not sure it will make a difference. Here's my thought process:

All dependencies are installed in the virtualenv and since we're caching the entire virtualenv, it doesn't matter in which step they're installed. Doing `python -m pip install -e.` will always try to install the deps in `requirements.txt` anyways but it will be fast(er) if there's a cache hit as they are already installed. 

You can see in this screenshot that the gradio install step goes by pretty fast (9 seconds) since the deps are already in the cache:

image

freddyaboulton:
Thanks for the review @abidlabs ! Will push up the changes to make sure we still keep the gains from #4515 (good catch!) will merge afterwards!!

.github/workflows/backend.yml
freddyaboulton: Yea we should remove! Will do that now.

.github/workflows/backend.yml
abidlabs: Nice

abidlabs:
Thanks @freddyaboulton for addressing the points!

PR #4518, opened 2023-06-14
Create Warning, Info modals

aliabid94:
Introduce `gr.Warning` and `gr.Info`, which can be written anywhere in a function and will cause a Modal to pop up instantly. Requires websockets. 

Example code:
```python
import gradio as gr
import time

def greet(name):
    gr.Info("Starting...")
    time.sleep(2)
    if len(name) < 3:
        gr.Warning("Name is short")
    time.sleep(2)
    raise gr.Error("Error!")

demo.queue().launch() 
```

Did not try to capture every `warnings.warn` that was raised, as it would be quite tricky (having to hijack the warnings library just as we did with tqdm) and I don't think there would be many use cases.

The modal looks the same for Error, Info, and Warning - @hannahblair would you like to take on styling those as well?


gradio/queueing.py
```diff
@@ -436,7 +474,8 @@ async def send_message(self, event, data: dict, timeout: float | int = 1) -> boo
                 event.websocket.send_json(data=data), timeout=timeout
             )
             return True
-        except Exception:
+        except Exception as e:
+            print(e)
```
abidlabs, line 478: Remove
```suggestion
```

gradio/blocks.py
```diff
@@ -1071,11 +1071,16 @@ async def call_function(
...
+                queueing.thread_data.blocks = self
+                queueing.thread_data.event_id = event_id
+                return fn(*args, **kwargs)
+
             if inspect.iscoroutinefunction(fn):
                 prediction = await fn(*processed_input)
             else:
                 prediction = await anyio.to_thread.run_sync(
-                    fn, *processed_input, limiter=self.limiter
+                    fn_wrap, *processed_input, limiter=self.limiter
```
abidlabs, line 1083: This means that this will only work for regular functions, not async functions or generators (?)

gradio/exceptions.py
```diff
@@ -40,3 +43,44 @@ def __init__(self, message: str):
 
     def __str__(self):
         return repr(self.message)
+
+
+def log_message(message: str, level: Literal["info", "warning"] = "info"):
+    from gradio import queueing
+
+    if not hasattr(queueing.thread_data, "blocks"):  # Function called outside of Gradio
```
abidlabs, line 51: If the function is called outside of Gradio (or even inside Gradio), then I think it should still raise a regular `warnings.warn()` (in the case of `gr.Warning`) and maybe log a print statement (in the case of `gr.Info`)

abidlabs: We do the same for `gr.Error()` -- also raise a regular Exception in Python

gradio/blocks.py
abidlabs: Why do we need to do this? Is it so that the queue knows which blocks corresponds to the function it is currently processing? I don't think this is necessary -- we don't need it to handle generators for example

abidlabs: I think you can handle `gr.Info` and `gr.Warning` in the same way generators are handled -- see https://github.com/gradio-app/gradio/blob/d65512cb3a91e2c5b6abf70b52f56a51195c47e7/gradio/queueing.py#L367

aliabid94: This is needed because there needs to be some communication between the gr.Warning() and the Gradio server. Since the gr.Warning is not returned, the server does not know if the gr.Warning is issued. Therefore, the gr.Warning needs to have a reference to the blocks server and the event id that issued it. I do this by storing references to both within the thread as thread locals. `anyio.to_thread` spins off a new thread, so I need to put this in another function called `fn_wrap`.

This works fine with generators and async. Try here:
```python
import gradio as gr
import time

def greet(name):
    gr.Info("Starting...")
    time.sleep(2)
    for i, letter in enumerate(name):
        if letter == " ":
            gr.Warning("Space detected!")
        yield "Hello " + name[:i+1] + "!"
        time.sleep(0.5)
    time.sleep(2)
    raise gr.Error("Error!")

demo.queue().launch() 
```

gradio/exceptions.py
aliabid94: done. `raise gr.Error` always raised a regular error

gradio/blocks.py
freddyaboulton: Will take a look today/tomorrow @aliabid94 ! I can confirm generators work in your use case, but warnings are not displayed with `every`

```python
import gradio as gr


def warn_1():
    gr.Warning("Warning 1!")
    return "foo"

def warn_2():
    gr.Warning("Warning 2!")
    return "bar"

with gr.Blocks() as demo:
    t1 = gr.Textbox()
    t2 = gr.Textbox()

    button = gr.Button("Warning 1")
    button.click(warn_1, None, t1, every=1)

    button2 = gr.Button("Warning 2")
    button2.click(warn_2, None, t2)
    

demo.queue().launch()
```

abidlabs:
Nice @aliabid94! I'm still doing some testing, but I noticed that if there's a `gr.Warning` raised shortly before the function ends, it does not appear. Here's an example:

```py
import gradio as gr
import time

def test(x):
    gr.Warning("step 1")
    yield "step 1"
    time.sleep(1)
    gr.Warning("step 2")
    yield "step 2"
    time.sleep(1)
    gr.Warning("done")
    # time.sleep(1) # If you uncomment this line, it works 
    yield "done"

gr.Interface(test, "textbox", "textbox").queue().launch()
```

gradio/queueing.py
```diff
@@ -2,18 +2,28 @@
...
+    Estimation,
+    LogMessage,
+    PredictBody,
+    Progress,
+    ProgressUnit,
+)
 from gradio.helpers import TrackedIterable
 from gradio.utils import AsyncRequest, run_coro_in_background, set_task_name
 
+thread_data = threading.local()
```
freddyaboulton, line 25: I think we can store this in `Context` more consistent with how we share global variables and it feels more standard than storing a variable at the module level.

gradio/exceptions.py
```diff
@@ -40,3 +43,51 @@ def __init__(self, message: str):
...
+            f"Queueing must be enabled to issue {level.capitalize()}: '{message}'."
+        )
+        return
+    print("here")
+    queueing.thread_data.blocks._queue.log_message(
+        event_id=queueing.thread_data.event_id, log=message, level=level
+    )
+
+
+@document()
```
freddyaboulton, line 68: nit but these aren't exceptions so I think it'd be better if they were in a different module

gradio/exceptions.py
```diff
@@ -1,4 +1,7 @@
+import warnings
+
 from gradio_client.documentation import document, set_documentation_group
+from typing_extensions import Literal
```
freddyaboulton, line 4: We no longer support 3.7 so you can import Literal from typing now

gradio/blocks.py
```diff
@@ -1071,6 +1071,8 @@ async def call_function(
             else:
                 fn = block_fn.fn
 
+            fn = utils.get_function_with_locals(fn, self, event_id)
```
freddyaboulton, line 1074: You should call the output variable something else. The thing you return from this is not a coroutine so `iscoroutine` will always be false.

I can't get async functions to work (repro below) and I think it's because of that:

```python
import gradio as gr
import asyncio


async def warn_1():
    for i in range(10):
        gr.Warning(f"Warning {i}!")
        await asyncio.sleep(0.5)
    return "foo"

def warn_2():
    gr.Warning("Warning 2!")
    return "bar"

with gr.Blocks() as demo:
    t1 = gr.Textbox()
    t2 = gr.Textbox()

    button = gr.Button("Warning 1")
    button.click(warn_1, None, t1)

    button2 = gr.Button("Warning 2")
    button2.click(warn_2, None, t2)
    

demo.queue(concurrency_count=2).launch()
```

image

gradio/exceptions.py
```diff
@@ -40,3 +43,51 @@ def __init__(self, message: str):
...
+            print(message)
+        elif level == "warning":
+            warnings.warn(message)
+        return
+    if not queueing.thread_data.blocks.enable_queue:
+        warnings.warn(
+            f"Queueing must be enabled to issue {level.capitalize()}: '{message}'."
+        )
+        return
+    print("here")
```
freddyaboulton, line 62: FYI extra print

js/app/src/components/StatusTracker/Toast.svelte
```diff
@@ -9,9 +9,9 @@
 <div class="toast-wrap">
 	{#each messages as { type, message, id } (id)}
 		<div animate:flip={{ duration: 300 }} style:width="100%">
-			{#if type === "error"}
-				<Error {message} on:close {id} />
-			{/if}
+			<!-- {#if type === "error"} -->
```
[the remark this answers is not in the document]
pngwn: We will add the if block back when we add `Info` and `Warning`, which I'm guessing is why they were only commented out.

client/js/src/client.ts
```diff
@@ -1180,7 +1188,7 @@ function handle_message(
 	data: any,
 	last_status: Status["stage"]
 ): {
-	type: "hash" | "data" | "update" | "complete" | "generating" | "none";
+	type: "hash" | "data" | "update" | "complete" | "generating" | "log" | "none";
```
pngwn, line 1191: I was thinking about this for a while, `error` being a separate type from `log` but I think it is correct even though we display them in a consistent manner. `gr.Error()` is relating to the status of the function itself, while `gr.Info()` and `gr.Warning()` are completely unrelated.

js/app/src/Blocks.svelte
```diff
@@ -234,6 +234,21 @@
...
+		messages = [
+			{
+				message,
+				fn_index,
+				type,
+				id: ++_error_id
+			},
+			...messages
+		];
+	};
```
pngwn, line 251: Can we refactor this to be pure (not modify outside state and just return a value), I'd like to start doing this more in the future so we may as well start. It will allow us to pull more stuff outside of the svelte files + make them easier to unit tests.

Something like :
```js
const create_message = (
  message: string,
  fn_index: number,
  type: ToastMessage["type"]
) => ({
    message,
    fn_index,
    type,
    id: ++_error_id
});
```

Then we could use it like:

```js
messages = [ create_message(_message, fn_index, "error"), ...messages];
```

Or we can pass in the original `message` array as well and actually concatenate them in the fn:

```js
messages = add_message(messages, _message, fn_index, "error");
```

pngwn: I don't mind which.

pngwn (commented):
Apart from one small comment, I'm happy with the frontend code so far. When we're happy with the core implementation, I guess @hannahblair will take over to add the new UI.

missionfloyd:
While you're at it, could you make it work with `show_progress=False`? It (`gr.Error()`) doesn't currently.

gradio/blocks.py
aliabid94: fixed

gradio/queueing.py
aliabid94: done

aliabid94:
> While you're at it, could you make it work with show_progress=False? It (gr.Error()) doesn't currently

This seems fixed in main.

aliabid94:
Ok this should now work with every type of function: Can test with this demo:
```python
import gradio as gr
import time
import asyncio

def normal(x):
    gr.Info("step 1")
    time.sleep(1)
    gr.Warning("step 2")
    time.sleep(1)
    gr.Warning("done")
    return "step done"

def generator(x):
    gr.Info("step 1")
    yield "step 1"
    time.sleep(1)
    gr.Warning("step 2")
    yield "step 2"
    time.sleep(1)
    gr.Warning("done")

async def asyncfn(x):
    gr.Info("step 1")
    await asyncio.sleep(3)
    gr.Warning("step 2")
    await asyncio.sleep(2)
    gr.Warning("done")
    return "step done"

async def asyncgen(x):
    gr.Info("step 1")
    yield "step 1"
    await asyncio.sleep(3)
    gr.Warning("step 2")
    yield "step 2"
    await asyncio.sleep(2)
    gr.Warning("done")


with gr.Blocks() as demo:
    text = gr.Textbox()
    text2 = gr.Textbox()
    with gr.Row():
        normal_btn = gr.Button("normal")
        generator_btn = gr.Button("generator")
        async_btn = gr.Button("async")
        asyncgen_btn = gr.Button("asyncgen")

    normal_btn.click(normal, text, text2)
    generator_btn.click(generator, text, text2)
    async_btn.click(asyncfn, text, text2)
    asyncgen_btn.click(asyncgen, text, text2)

demo.queue().launch()
```

aliabid94:
ready for re-review

CHANGELOG.md
```diff
@@ -111,8 +111,9 @@ demo.launch()
...
 - Add `autoplay` kwarg to `Video` and `Audio` components by [@pngwn](https://github.com/pngwn) in [PR 4453](https://github.com/gradio-app/gradio/pull/4453)
 - Add `allow_preview` parameter to `Gallery` to control whether a detailed preview is displayed on click by
-  [@freddyaboulton](https://github.com/freddyaboulton) in [PR 4470](https://github.com/gradio-app/gradio/pull/4470)
+[@freddyaboulton](https://github.com/freddyaboulton) in [PR 4470](https://github.com/gradio-app/gradio/pull/4470)
 - Add `latex_delimiters` parameter to `Chatbot` to control the delimiters used for LaTeX and to disable LaTeX in the `Chatbot` by [@dawoodkhan82](https://github.com/dawoodkhan82) in [PR 4516](https://github.com/gradio-app/gradio/pull/4516)
+- Can now issue `gr.Warning` and `gr.Info` modals. Simply put the code `gr.Warning("Your warning message")` or `gr.Info("Your info message")` as a standalone line in your function. By [@aliabid94](https://github.com/aliabid94) in [PR 4518](https://github.com/gradio-app/gradio/pull/4518).
```
abidlabs, line 116: Would be great to include a screenshot of the modal here

guides/01_getting-started/02_key-features.md
```diff
@@ -25,9 +25,11 @@ You can load a large dataset into the examples to browse and interact with the d
 
 Continue learning about examples in the [More On Examples](https://gradio.app/more-on-examples) guide.
 
-## Errors
+## Alerts
 
-You wish to pass custom error messages to the user. To do so, raise a `gr.Error("custom message")` to display an error message. If you try to divide by zero in the calculator demo above, a popup modal will display the custom error message. Learn more about Error in the [docs](https://gradio.app/docs#error).
+You wish to pass custom error messages to the user. To do so, raise a `gr.Error("custom message")` to display an error message. If you try to divide by zero in the calculator demo above, a popup modal will display the custom error message. Learn more about Error in the [docs](https://gradio.app/docs#error). 
+
+You can also issue `gr.Warning("message")` and `gr.Info("message")` by having them as standalone lines in your function, which will immediately display modals while continuing the execution of your function. Queueing needs to be enabled for this to work.
```
abidlabs, line 32: Include an example here (standalone lines maybe not be totally clear)

CHANGELOG.md
abidlabs, line 116: Would be good to include some sample code here

gradio/helpers.py
```diff
@@ -893,3 +896,50 @@ def __init__(self, target: Block | None, _data: Any):
...
+        return
+    context.thread_data.blocks._queue.log_message(
+        event_id=context.thread_data.event_id, log=message, level=level
+    )
+
+
+@document()
+class Warning:
+    """
+    This class allows you to pass custom warning messages to the user. You can do so simply with `gr.Warning('message here')`, and when that line is executed the custom message will appear in a modal on the demo.
```
abidlabs, line 923:
```suggestion
    This class allows you to pass custom warning messages to the user. You can do so simply by instantiating `gr.Warning('message here')`, and when that line is executed the custom message will appear in a modal on the demo.
```

pngwn (approved):
Reviewed the frontend code a few times now, and looks good to me! Thanks @hannahblair and @aliabid94!

gradio/helpers.py
```diff
@@ -893,3 +896,50 @@ def __init__(self, target: Block | None, _data: Any):
...
+        Parameters:
+            message: The warning message to be displayed to the user.
+        """
+        log_message(message, level="warning")
+
+
+@document()
+class Info:
+    """
+    This class allows you to pass custom info messages to the user. You can do so simply with `gr.Info('message here')`, and when that line is executed the custom message will appear in a modal on the demo.
```
abidlabs, line 937:
```suggestion
    This class allows you to pass custom info messages to the user. You can do so simply by instantiating `gr.Info('message here')`, and when that line is executed the custom message will appear in a modal on the demo.
```

gradio/helpers.py
aliabid94: actually changing this to a function, no need to be a class.

gradio/helpers.py
aliabid94: actually changing this to a function, no need to be a class.

Issue #4524, opened 2023-06-15
Audio Output - regression from 3.29 -> 3.34


sanchit-gandhi:
### Describe the bug

A Gradio demo that outputs audio previously worked with v3.29. It now does not work with v3.34 (or v3.33). Wonder if there's a workaround or if this is an issue on main?

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Dummy example of the demo being used as part of the HF Audio course. This code snippet will run for Gradio v3.29, but not v3.34.

Select v3.29:
```
pip install gradio==3.29
```
Then:
```python
from datasets import load_dataset
import gradio as gr

minds = load_dataset("hf-internal-testing/librispeech_asr_dummy", "clean", split="validation")
minds = minds.add_column("intent_class", [0] * len(minds))

id2label = lambda x : "positive"

def generate_audio():
    example = minds.shuffle()[0]
    audio = example["audio"]
    return (
        audio["sampling_rate"],
        audio["array"],
    ), id2label(example["intent_class"])


with gr.Blocks() as demo:
    with gr.Column():
        for _ in range(4):
            audio, label = generate_audio()
            output = gr.Audio(audio, label=label)

demo.launch(debug=True)
```

Select v3.34:
```
pip install gradio=3.34
```
(or equivalently `pip install -U gradio`)

Then re-run the code snippet to hit the error log pasted below.

### Logs

```shell
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[2], line 16
     15             audio, label = generate_audio()
---> 16             output = gr.Audio(audio, label=label)
     18 demo.launch(debug=True)

File ~/venv/lib/python3.9/site-packages/gradio/components.py:2390, in Audio.__init__(self, value, source, type, label, every, show_label, interactive, visible, streaming, elem_id, elem_classes, format, **kwargs)
   2387     raise ValueError(
   2388         "Audio streaming only available if source is 'microphone'."
   2389     )
-> 2390 IOComponent.__init__(
   2391     self,
   2392     label=label,
   2393     every=every,
   2394     show_label=show_label,
   2395     interactive=interactive,
   2396     visible=visible,
   2397     elem_id=elem_id,
   2398     elem_classes=elem_classes,
   2399     value=value,
   2400     **kwargs,
   2401 )
   2402 TokenInterpretable.__init__(self)

[...]

PR #4525, opened 2023-06-15
Fix regression in `gr.Audio`

abidlabs:
Restores the ability to pass in a tuple `(sample rate, audio array)` to `gr.Audio()` 

Closes: #4524


freddyaboulton (approved):

abidlabs:
Thanks @freddyaboulton!

Issue #4534, opened 2023-06-15
Gradio 3.35.0 breaks gr.Chatbot streaming


decent-engineer-decent-datascientist:
### Describe the bug

Gradio 3.35.0 breaks gr.Chatbot streaming. 

previous behavior updated the chat history at every yield.
Latest gradio version only displays initial yield result.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Working on it

### Logs

```shell
N/A
```

### System Info

```shell
gradio==3.35.0, OS X 13.4 and Ubuntu 22.04, Safari, chrome
```

### Severity

blocking upgrade to latest gradio version

PR #4537, opened 2023-06-15
Chatbot streaming hotfix

aliabid94:
Chatbot streaming broke as a result of latex changes, because the rendered Markdown only rendered onMount, and did not react to message changes. Changed to `afterUpdate`

Run demo/chatbot_streaming to see fix.

Fixes: #4534


freddyaboulton (approved):
Verified this fix the streaming case and to my knowledge it doesn't break any of the other chatbot demos.

pngwn:
Ah, thanks for catching that!

redrocket8:
If the string has latex code then the latex code is not rendered

PR #4538, opened 2023-06-15
fix modal position

pngwn:
# Description

Fixes the error modal position. I've also bumped the text size on the error modal because at 12px it is borderline unreadable.


freddyaboulton (approved):
Looks good @pngwn ! Thank you 

error_modal

PR #4540, opened 2023-06-15
Fix chatbot height

aliabid94:
On main, chatbot is broken in that it doesn't scroll properly to the last message, and the bottom of the chat is cut off. Fixed here.


js/chatbot/src/MarkdownCode.svelte
```diff
@@ -19,6 +20,7 @@
 			requestAnimationFrame(() => {
 				el.innerHTML = DOMPurify.sanitize(marked.parse(message));
 				mounted = true;
+				dispatch("load");
```
pngwn, line 23: nice

pngwn (commented):
Hmm, scrolling down is fixed but I don't seem able to scroll up now.

pngwn (approved):
weird. lgtm!

PR #4550, opened 2023-06-16
Remove mocks

pngwn:
# Description

This PR removes all of the mocks from out browser based tests and runs the actual python apps instead. I have automated this as much as possible, all a developer needs to do is create a test file with the same name as a demo and before the test is run that demo will be launched and the correct page navigated to.

Devs do not need to manually navigate to the correct page any more, that is taken care of automatically, so inside tests you can simply start writing tests! I have re-enabled parallelism for test files but all tests within a file will run in serial. This has meant that applications cannot all start on the same port, the port is `787` + the worker index which is safe as the worker is tied to the test file. We do not start a new instance of the gradio app for every tests but we share them as long as we are in the same test file. I chose a slightly different gradio port to default to avoid collisions.

Due to the way I needed to implement this, this can result in hanging python processes at the end (because it is actually subsequent tests run that kill prior processes). As part of the clean up I am currently killing all current python processes, which isn't ideal. I have access to the PIDs but I don't have a great way of passing them back tot he parent process via the test framework right now. When I figure that out I will kill the relevant processes in a more granular fashion.


pngwn:
and obviously it doesn't work in ci, why would it.

js/tootils/src/index.ts
```diff
@@ -6,4 +10,103 @@ export function wait(n: number) {
...
+				...process.env,
+				GRADIO_SERVER_PORT: `787${workerIndex}`,
+				PYTHONUNBUFFERED: "true"
+			}
+		});
+		_process.stdout.setEncoding("utf8");
+
+		_process.stdout.on("data", (data) => {
+			const _data = data.toString();
+			const PORT_RE = new RegExp(`:787${workerIndex}`);
```
abidlabs, line 44: Just FYI this will not work if there are more than 10 tests.

js/tootils/src/index.ts
```diff
@@ -6,4 +10,103 @@ export function wait(n: number) {
...
+						process: _process,
+						file: file
+					});
+					update_env(workerIndex, _process.pid!);
+				} catch (e) {
+					throw e;
+				}
+			}
+
+			await page.goto(`localhost:787${workerIndex}`);
```
abidlabs, line 103: Same this will not work if there are more than 10 tests.

js/tootils/src/index.ts
freddyaboulton: What if we create one app that has all the gradio apps as subpaths (via `mount_gradio_app`)? That way we can have more than 10 apps and we can access them as `locahost:7860/app_{i}` for i in [0, k)

abidlabs: 👍 It would be nicer as we wouldn't have to hardcode the port either or worry about closing it

freddyaboulton: Also a good way to test `mount_gradio_app`. Sometimes that has broken on release unexpectedly.

pngwn: Workers get reused as it is a worker pool, so I guess a better way of phrasing is that tests are tied to workers rather than workers being tied to tests. So it is unlikely to cause issues unless you have more than 20 physical cores. 

That said it is a good point as some contributors may have 24+ core machines. 

I like the sub path idea, I'll try that.

demo/blocks_xray/run.py
```diff
@@ -1,14 +1,13 @@
 import gradio as gr
-import random
 import time
 
+disease_values = [0.25, 0.5, 0.75]
+
 def xray_model(diseases, img):
-    time.sleep(4)
```
pngwn, line 6: We don't want sleeps in anything we're using to test because it just slows them down for no reason.

demo/blocks_xray/run.py
```diff
@@ -1,14 +1,13 @@
...
-    return [{disease: random.random() for disease in diseases}]
```
pngwn, line 7: We also don't want any randomness either.

js/app/test/kitchen_sink.spec.ts
```diff
@@ -262,5 +95,10 @@ test("test outputs", async ({ page }) => {
 
 	const audio = await page.locator("audio").nth(0);
 	const audio_data = await audio.getAttribute("src");
-	await expect(audio_data).toEqual(BASE64_AUDIO);
+
+	await expect(
+		audio_data?.endsWith(
+			"gradio/92aff6ffe140da201a4a94cb3c3b9e1ff0b2a25a/cantina.wav"
+		)
+	).toBeTruthy();
```
pngwn, line 103: The paths generated by gradio are different on different machines but I think the end of the path should be consistent.

js/app/test/outbreak_forecast.spec.ts
```diff
@@ -21,24 +15,12 @@ test("matplotlib", async ({ page }) => {
 
 	const matplotlib_img = await page.locator("img").nth(0);
 	const matplotlib_img_data = await matplotlib_img.getAttribute("src");
-	await expect(matplotlib_img_data).toEqual(BASE64_PLOT_IMG);
+	await expect(matplotlib_img_data).toBeTruthy();
```
pngwn, line 18: I'm really not sure what is going on here. I'd expect the image data to be identical everywhere. The original test passed fine on my machine but fails in CI. This isn't a super important test because it is mostly testing the user function rather than anything else but the inconsistency is annoying. A visual test would be better here.

js/app/test/outbreak_forecast.spec.ts
```diff
@@ -21,24 +15,12 @@ test("matplotlib", async ({ page }) => {
...
-			}
-		]
-	]);
-	await mock_theme(page);
-	await page.goto("http://localhost:9876");
-
 	await page.getByLabel("Plot Type").click();
-	await page.getByRole("button", { name: "Matplotlib" }).click();
+	await page.getByRole("button", { name: "Plotly" }).click();
```
pngwn, line 23: This was always wrong since we were mocking it really didn't matter what we selected.

js/app/test/slider_release.spec.ts
```diff
@@ -47,6 +43,6 @@ test("slider release", async ({ page }) => {
 
 	const val = await slider.inputValue();
 	expect(parseInt(val)).toBeCloseTo(70);
-	expect(value).toHaveValue("70");
+	expect(parseInt(await value.inputValue())).toBeCloseTo(70);
```
pngwn, line 46: Now we are testing the actual UI there is less accuracy here but this works nicely.

package.json
```diff
@@ -19,8 +19,9 @@
 		"test": "pnpm --filter @gradio/client build && vitest dev --config .config/vitest.config.ts",
 		"test:run": "pnpm --filter @gradio/client build && vitest run --config .config/vitest.config.ts",
 		"test:node": "TEST_MODE=node pnpm  vitest run --config .config/vitest.config.ts",
-		"test:browser": "pnpm --filter @gradio/app test:browser:full",
+		"test:browser": "pnpm --filter @gradio/app test:browser",
 		"test:browser:full": "run-s build test:browser",
+		"test:browser:verbose": "GRADIO_TEST_VERBOSE=true pnpm test:browser",
```
pngwn, line 24: I silence the python child process's stderr and stdout by default because it is incredibly noisy and interferes with the tests but if something goes wrong we really need to be able to see all of the logs that the child process is generating. This script enable 'verbose mode' which will print anything that is logged.

pngwn:
Okay I have rejigged this to use `mount_gradio_app` and it seems to work much more nicely. Only requires one child process with less stopping and starting, and the code is a bit simpler too.

I needed to update some tests because they were wrong/not working for various reasons. I have commented the code were necessary to explain why the demos + tests were modified.

The tests actually now seem to run slightly faster, even though were are spinning up an actual app rather than mocking. Probably because the 'generate config' step we have to go through previously is almost as expensive as running an app.

Please give this a run on your machines to make sure everything works as expected.

pngwn:
we green now

js/app/test/kitchen_sink.spec.ts
abidlabs: Yes, they should. In this path:

* `gradio/`comes from the global temp directory set by the GRADIO_TEMP_DIR env variable (defaults to `/tmp/gradio`)
* `92aff6ffe140da201a4a94cb3c3b9e1ff0b2a25a` is based on the hash of the file
* `cantina.wav` is the name of the file

.config/playwright-setup.js
```diff
@@ -1,21 +1,98 @@
...
+import uvicorn
+from fastapi import FastAPI
+import gradio as gr
+${demos.map((d) => `from demo.${d}.run import demo as ${d}`).join("\n")}
+print("hi")
+app = FastAPI()
+
+${demos
+	.map((d) => `app = gr.mount_gradio_app(app, ${d}, path="/${d}")`)
+	.join("\n")}
```
abidlabs, line 93: Very nice

demo/blocks_xray/run.py
abidlabs: Agreed

abidlabs:
LGTM @pngwn! Very fast & all tests passing locally.

The only issue I faced was when port 7879 was already occupied. In which case the tests would hang indefinitely without printing any error messages.

abidlabs (approved):

pngwn:
Merging this now, everything looks good. Will see how we can do more to share CI code in a future PR.

abidlabs:
Very nice @pngwn! Unlocks a lot of 4.0 work

cc @dawoodkhan82

Issue #4543, opened 2023-06-16
`allow_preview=False` doesn't seem to work properly


hysts:
### Describe the bug

This is very weird and I have no idea what is happening, but the following code doesn't work properly on my local environment with `gradio==3.35.2`, while it works on HF Spaces.

```python
import pathlib

import gradio as gr

paths = [
    path.as_posix() for path in sorted(pathlib.Path('images').rglob('*.jpg'))
]


def fn(data: gr.SelectData) -> int:
    return data.index


with gr.Blocks() as demo:
    gallery = gr.Gallery(value=paths,
                         allow_preview=False,
                         columns=2,
                         rows=2,
                         height='600px',
                         object_fit='scale-down')
    selected_index = gr.Number(label='selected index', precision=0, value=-1)
    gallery.select(fn=fn, outputs=selected_index)
demo.queue().launch()
```

On HF Spaces, it works fine (though there's another issue https://github.com/gradio-app/gradio/issues/4542), but when I run this code on my local environment, somehow `.select` doesn't work when I launched the app. But surprisingly, it works when I change the display scale.

https://github.com/gradio-app/gradio/assets/25161192/adbb899a-1e68-43a8-ae51-c5494f1fd666

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

https://huggingface.co/spaces/hysts-debug/gallery-allow-preview-false

### Logs

```shell
n/a
```

### System Info

```shell
gradio==3.35.2
```

### Severity

blocking upgrade to latest gradio version

PR #4551, opened 2023-06-16
Fix gallery select event with custom height and allow_preview=False

freddyaboulton:
# Description

Closes: #4543

When setting a height on the gallery, the select event is not being triggered when `allow_preview=False` because `can_zoom` is being calculated as `False`. 

Not too familiar with the gallery code but I assumed the `can_zoom` logic is correct. Since the gallery won't zoom with `allow_preview=False`, my fix is to set `can_zoom=true` in that case to trigger the event. But please check me @pngwn @hannahblair @aliabid94  @dawoodkhan82 

# Checklist:

# A note about the CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ No changes to highlight.
 
 ## Bug Fixes:
 
-No changes to highlight.
+- Fix bug where `select` event was not triggered in `gr.Gallery` if `height` was set to be large with `allow_preview=False` by [@freddyaboulton](https://github.com/freddyaboulton) in [PR 4551](https://github.com/gradio-app/gradio/pull/4551)
 
 ## Other Changes:
 
--- a/js/gallery/src/Gallery.svelte
+++ b/js/gallery/src/Gallery.svelte
@@ -130,7 +130,8 @@
 		});
 	}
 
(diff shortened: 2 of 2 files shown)


pngwn (approved):
Hmm. Interesting. When `allow_preview=False` technically it isn't possible to 'select' an image (going by the semantics of 'selected' in this component). You are just clicking it, but maybe that distinction isn't important. I do think we should probably refactor this code to handle selection + full size previewing separately to avoid confusion. And also because i think we should implement `multi-select` for the gallery too at some point (and you obviously can't preview multiple fullsize images at once).

If we're happy with the select semantics in this PR them I'm happy for it to be merged (code looks fine).

cc @abidlabs @aliabd @dawoodkhan82 @hannahblair

abidlabs:
The semantics are good for me. We use `.select()` in other components to mean that you've selected on some *portion* of the value as opposed to clicking on the entire thing, and we provide the event data to go along with that. E.g. in `gr.Image` clicking on a particular pixel triggers the `.select()` event handler even though there is no change in the UI. `So in this case, I think `.select()` still works.

pngwn:
Sounds good. Happy to merge then!

abidlabs:
Sounds good let’s merge it in

Issue #4547, opened 2023-06-16
audio.stop_recording, how can the audio itself be used as input to the function in this event?


Nancis1130:
I need to establish an event that occurs when recording is stopped, and the input for the event is the recorded audio.
When I do this, the function does not get input, the input is None.

```
def test(ori):
    print(ori)
    return 'hi'
    
with gr.Blocks() as demo:
    with gr.Row():
        audio = gr.Audio(source='microphone', label='请输入语音', scale=1)
        text = gr.Textbox(scale=2.5)
        sub = gr.Button('submit', scale=0.5)
    audio.stop_recording(test, inputs = audio, outputs = text)

if __name__ == "__main__":
    demo.launch()

```

Nancis1130:
My version is 3.25.0.
What I want to achieve is that when I click the "stop Recording" button, the recorded audio can automatically serve as input to a certain function. 
If the method "audio.stop_recording" is incorrect, please point it out! I really need this, thank you!

freddyaboulton:
Hi @Nancis1130 ! You can use `.change` event in the meantime but I believe the `stop_recording` event should work as you describe so will look into it. cc @pngwn

pngwn:
This seems odd to me because, if I'm understanding, the function is executed (so the event is working) but the input is incorrect. 

But all of our events work pretty much the same as far as I understand. 

Will try to repro.

Nancis1130:
> Hi @Nancis1130 ! You can use `.change` event in the meantime but I believe the `stop_recording` event should work as you describe so will look into it. cc @pngwn

This method worked, I really appreciate it!

Nancis1130:
> This seems odd to me because, if I'm understanding, the function is executed (so the event is working) but the input is incorrect.
> 
> But all of our events work pretty much the same as far as I understand.
> 
> Will try to repro.

Thank you for your reply. I started the HTTP service using a remote server IP and port number. I'm not sure if this is the cause of the problem.
Additionally, when using the chatbot component, I referred to the [chatbot_streaming](https://huggingface.co/spaces/gradio/chatbot_streaming). However, I found that when I tested locally, the chatbot could stream full output, while when I started the service on a remote server, it could only output one or a few characters.
I have been using Gradio for a short period of time, so I'm sorry if I need to open a new issue to raise this issue.
When the service is local:
[...]

PR #4554, opened 2023-06-16
Send component data on stop_recording event

freddyaboulton:
# Description

Closes: #4547 

The stop_recording event does not send the microphone data when the event is triggered. This PR fixes that. I think the same thing is happening on video, will investigate.

To test, run this demo and observe that the output audio and video components are populated correctly.

```python
import gradio as gr

def test(ori):
    ori
    
with gr.Blocks() as demo:
    with gr.Row():
        audio = gr.Audio(source='microphone')
        video = gr.Video(source='webcam')
        audio_output = gr.Audio()
        video_output = gr.Video()
    audio.stop_recording(lambda x: x, inputs=audio, outputs=audio_output)
    video.stop_recording(lambda x: x, inputs=video, outputs =video_output)


if __name__ == "__main__":
    demo.launch()
```
This repro from the original PR that introduced this PR still works as expected

```python
import gradio as gr

def start_recording():
    print("Recording has started")

def stop_recording():
    print("Recording has stopped")

def start_playing():
    print("Playing has started")

def stop_playing():
    print("Playing has stopped")

def pause_playing():
    print("Playing has paused")


with gr.Blocks() as demo:
    stream_input = gr.Video(source="webcam")

    stream_input.play(start_playing)
    stream_input.pause(pause_playing)
    stream_input.stop(stop_playing)
    stream_input.start_recording(start_recording)
    stream_input.stop_recording(stop_recording)
        
demo.launch()
```

# Checklist:

# A note about the CHANGELOG

### DIFF
--- a/js/audio/src/Audio.svelte
+++ b/js/audio/src/Audio.svelte
@@ -86,7 +86,7 @@
 
 	const dispatch_blob = async (
 		blobs: Array<Uint8Array> | Blob[],
-		event: "stream" | "change"
+		event: "stream" | "change" | "stop_recording"
 	) => {
 		let audio_blob = new Blob(blobs, { type: "audio/wav" });
 		value = {
@@ -150,6 +150,7 @@
 			recorder.addEventListener("stop", async () => {
 				recording = false;
 				await dispatch_blob(audio_chunks, "change");
+				await dispatch_blob(audio_chunks, "stop_recording");
 				audio_chunks = [];
 			});
 		}
@@ -185,7 +186,6 @@
 	});
 
 	const stop = async () => {
-		dispatch("stop_recording");
 		recorder.stop();
 		if (streaming) {
 			recording = false;
--- a/js/image/src/Webcam.svelte
+++ b/js/image/src/Webcam.svelte
@@ -23,7 +23,11 @@
 			| string;
 		error: string;
 		start_recording: undefined;
-		stop_recording: undefined;
+		stop_recording: {
+			data: FileReader["result"];
+			name: string;
+			is_example?: boolean;
+		};
 	}>();
 
 	onMount(() => (canvas = document.createElement("canvas")));
@@ -74,13 +78,17 @@
 
 	function take_recording() {
 		if (recording) {
-			dispatch("stop_recording");
 			media_recorder.stop();
 			let video_blob = new Blob(recorded_blobs, { type: mimeType });
 			let ReaderObj = new FileReader();
 			ReaderObj.onload = function (e) {
 				if (e.target) {
 					dispatch("capture", {
+						data: e.target.result,
+						name: "sample." + mimeType.substring(6),
+						is_example: false
+					});
+					dispatch("stop_recording", {
 						data: e.target.result,
 						name: "sample." + mimeType.substring(6),
 						is_example: false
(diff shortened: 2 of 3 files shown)


js/image/src/Webcam.svelte
```diff
@@ -85,6 +88,11 @@
 						name: "sample." + mimeType.substring(6),
 						is_example: false
 					});
+					dispatch("stop_recording", {
```
freddyaboulton, line 91: What is the `capture` event above @pngwn @aliabid94 ? It doesn't exist in the python api afaik

pngwn: So , I think there is some confusion here. This fix works but not because we are sending the audio data back with the event. That only necessary if we wish to populate the `EventData` argument which event related metadata. However that isn't the case here, we just need the value to be updated _before_ the `stop_recording` event is dispatched. The issues here is actually a race condition/ bug in the original code due to _when_ we are dispatching the event rather than what that event contains.

`capture` is an internal event (not public) used by the Video component to set the data when it changes. That is why I was confused about this issue, the event and the setting of data are independent in almost all components (and particularly here), `capture` handles the setting of data regardless of events.

The reason this fix works is because you moved the `stop_recording` dispatch to the `onload` of the `FileReader` instance and placed it _after_ the `capture` event, which is correct and an oversight from the original PR.

We can remove the event data as it isn't needed (and would populate the `EventData` for functions that use) which would duplicate the data and is inconsistent with other events/ components.

js/image/src/Webcam.svelte
```diff
@@ -85,6 +88,11 @@
...
+						data: e.target.result,
+						name: "sample." + mimeType.substring(6),
+						is_example: false
+					});
```
pngwn, line 95: This is the tweak. If you run the demo you posted, you'll notice is still works as expected.

```suggestion
					dispatch("stop_recording");
```

js/image/src/Webcam.svelte
```diff
@@ -23,7 +23,11 @@
 			| string;
 		error: string;
 		start_recording: undefined;
-		stop_recording: undefined;
+		stop_recording: {
+			data: FileReader["result"];
+			name: string;
+			is_example?: boolean;
+		};
```
pngwn, line 30: We can update this type (see later comment for context).

```suggestion
		stop_recording: undefined;
```

js/audio/src/Audio.svelte
```diff
@@ -150,6 +150,7 @@
 			recorder.addEventListener("stop", async () => {
 				recording = false;
 				await dispatch_blob(audio_chunks, "change");
+				await dispatch_blob(audio_chunks, "stop_recording");
```
pngwn, line 153: This however is correct, the original implementation was just buggy, this is essentially a similar race condition in that the `stop_recording` event is sent synchronously when the button is clicked while the data being processed is asynchronously and is only complete _after_ the API request is sent.

pngwn (changes requested):
Ok, I was a little confused by this at first but I understand it now. The audio fix is correct but the video fix includes some unnecessary additions. I've left comments in the code to explain more.

freddyaboulton:
Thank you for the detailed explanation @pngwn ! I've addressed your comments.

pngwn (approved):
Awesome, thanks @freddyaboulton!

Issue #4564, opened 2023-06-19
Gradio 3.35.2 breaks compatibility due to missing components


vladmandic:
### Describe the bug

Gradio 3.35.2 does not define some components that were previsouly available  
This is most likely due to changes in #4487  

Currently using Gradio 3.32.0 where such components are available  
This is blocking any possible upgrades  

Impact is to app ecosystem with 20k+ user installs

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```python
import gradio as gr
Form = gr.components.Form
FormComponent = gr.components.FormComponent
```

### Logs

```shell
AttributeError: module 'gradio.components' has no attribute 'Form'
AttributeError: module 'gradio.components' has no attribute 'FormComponent'
```

### System Info

```shell
Gradio 3.35.2
Chrome 114
Windows 11
```

### Severity

blocking upgrade to latest gradio version

abidlabs:
Hi @vladmandic thanks for the catch. We'll restore these for backwards compatibility, even though these classes are meant to be used internally. To understand better, what are you using these classes for?

vladmandic:
To create custom gradio components that work well with forms.
Example: <https://github.com/vladmandic/automatic/blob/master/modules/ui_components.py>

Even if code is not what it should be, its used by tons of 3rd party extensions and change would have to be done on massive scale - basically locking app down forever to old version of gradio.

abidlabs:
Fair enough

PR #4566, opened 2023-06-19
Restore some missing imports in `__init__`

abidlabs:
Closes: #4564

### DIFF
--- a/gradio/components/__init__.py
+++ b/gradio/components/__init__.py
@@ -37,7 +37,6 @@ from gradio.components.model3d import Model3D
 from gradio.components.number import Number
 from gradio.components.plot import Plot
 from gradio.components.radio import Radio
-from gradio.components.scatter_plot import ScatterPlot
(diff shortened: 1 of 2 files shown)


gradio/components/__init__.py
```diff
@@ -2,8 +2,12 @@
 from gradio.components.audio import Audio
 from gradio.components.bar_plot import BarPlot
 from gradio.components.base import (
+    Column,
     Component,
+    Form,
+    FormComponent,
     IOComponent,
+    Row,
```
abidlabs, line 10: It would be better for users to import `Column`, `Form`, and `Row` from `gr.layouts` instead of `gr.components`, but for the sake of backwards compatibility, I have included them here

aliabid94 (approved):
LGTM, we can deprecate this in 4.0.

abidlabs:
Thanks @aliabid94!

Issue #4505, opened 2023-06-14
Visible attribute has no effect on gradio.Group()


lcrh:
### Describe the bug

This is the same issue as in https://github.com/gradio-app/gradio/issues/2246, which was fixed and seems to have regressed as of 3.34.0.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

See the repro in https://github.com/gradio-app/gradio/issues/2246

### Screenshot

![image](https://github.com/gradio-app/gradio/assets/14020730/8aba39e1-9550-4bfb-970f-ff36fe660fd2)

### Logs

```shell
Running on local URL:  http://127.0.0.1:7861
Running on local URL:  http://127.0.0.1:7861
Running on public URL: https://4b374dd45cea09ab90.gradio.live

This share link expires in 72 hours. For free permanent hosting and GPU upgrades, run `gradio deploy` from Terminal to deploy to Spaces (https://huggingface.co/spaces)
```

### System Info

```shell
3.34.0, Windows, Chrome Version 114.0.5735.110 (Official Build) (64-bit)
```

### Severity

serious, but I can work around it

apolinario:
This issue seems to have regressed on `3.36.1`

PR #4567, opened 2023-06-19
setting `visible=False` to `gr.Group` hides it

abidlabs:
Closes: #4505

In order to make sure this regression does not happen again, I wrote a TS test. However, I'm not happy with this test, as it tests implementation (the presence of a class name) instead of behavior (checking to see if the CSS display is actually set to "none"). 

I tried adding these lines of JS to the test but it wouldn't work for me:

at the beginning of the test:

```js
// Wait for any potential reactivity updates in the component
await waitFor(() => {});
```

and at the end of the test:

```js
const computedStyles = window.getComputedStyle(groupElement);
assert(computedStyles.display === 'none', "Group element is not hidden.");
```

If we can get this to work, that'd be much better!

### DIFF
--- a/js/app/src/components/Group/Group.svelte
+++ b/js/app/src/components/Group/Group.svelte
@@ -4,7 +4,7 @@
 	export let visible: boolean = true;
 </script>
 
-<div id={elem_id} class={elem_classes.join(" ")} class:hidden={!visible}>
+<div id={elem_id} class={elem_classes.join(" ")} class:hide={!visible}>
 	<slot />
 </div>
 
@@ -25,5 +25,9 @@
 
 	div > :global(* + *:not(.absolute)) {
 		border-top: none !important;
+	}
+
+	.hide {
+		display: none;
 	}
 </style>
--- a/js/app/src/components/Group/Group.test.ts
+++ b/js/app/src/components/Group/Group.test.ts
@@ -1,0 +1,26 @@
+import { test, describe, assert, afterEach, vi } from "vitest";
+import { cleanup, render } from "@gradio/tootils";
+
+import Group from "./Group.svelte";
+
+describe("Group", () => {
+	afterEach(() => {
+		cleanup();
+		vi.restoreAllMocks();
+	});
+
+	test("setting visible to false hides the Group", async () => {
+		render(Group, {
+			elem_id: "group",
+			visible: false
+		});
+
+		const groupElement = document.getElementById("group");
+
+		assert(groupElement !== null, "Group element not found.");
+		assert(
+			groupElement.classList.contains("hide"),
+			"Group element is not hidden."
+		);
+	});
+});
--- a/js/app/src/components/Group/Group.test.ts
+++ b/js/app/src/components/Group/Group.test.ts
@@ -6,7 +6,6 @@ import Group from "./Group.svelte";
 describe("Group", () => {
 	afterEach(() => {
 		cleanup();
-		vi.restoreAllMocks();
 	});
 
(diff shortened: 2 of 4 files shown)


js/app/src/components/Group/Group.svelte
```diff
@@ -4,7 +4,7 @@
 	export let visible: boolean = true;
 </script>
 
-<div id={elem_id} class={elem_classes.join(" ")} class:hidden={!visible}>
+<div id={elem_id} class={elem_classes.join(" ")} class:hide={!visible}>
```
abidlabs, line 7: Renaming to be consistent with `gr.Column`, `gr.Row`, etc.

pngwn:
I'd expect the code you posted to work without any need to await reactive changes because it should all happen synchronously. Will take a look.

abidlabs:
Thanks @pngwn that's what I had thought as well!

pngwn:
Okay, I understand it now. These virtual DOM environments that we use to test in node (`JSDom` `happy-dom`) don't fully implement `window.getComputedStyles()`. I'm not sure about `JSDom` but `happy-dom` seems to do some half-way thing (that I think it confusing). If the style is set directly with the `style` property `<element style="display: none;" />` then it will show up on `window.getComputedStyles()` but if it is attached via CSS, it doesn't show up.

I guess because it isn't actually rendering anything, it is pretty challenging to do. Ideally we would actually not render anything if `visible=False` and then we could just check the element doesn't exist but we can't do that currently for various reasons.

I think this is probably the best we can do until we make some architectural changes and/ or add some visual testing.

js/app/src/components/Group/Group.test.ts
```diff
@@ -0,0 +1,26 @@
+import { test, describe, assert, afterEach, vi } from "vitest";
+import { cleanup, render } from "@gradio/tootils";
+
+import Group from "./Group.svelte";
+
+describe("Group", () => {
+	afterEach(() => {
+		cleanup();
+		vi.restoreAllMocks();
```
pngwn, line 9: We can remove this since we haven't mocked anything in this test.

```suggestion
```

pngwn (approved):
See the post comments for more info, I think this is okay for now. It will prevent regression and we can improve our testability at some point as well as our testing!

aliabid94 (approved):

abidlabs:
Ah I see, that’s good to know ahead of the test hacking. Thanks for the review!

aliabid94 (approved):
Works. I hate gr.Group and gr.Box, they render so poorly, we should fix or deprecate in 4.0.

abidlabs:
Hmm really? I never really use them myself but feel free to open an issue @aliabid94 and we can assign to 4.0

taesiri:
@abidlabs 

I want to test this right away for my app. If I install Gradio from sources, should this work? I tested Gradio version `3.35.2` (installed from GitHub 3 minutes ago), but I can't hide Groups by setting `visible=False`.

@abidlabs 

> Works. I hate gr.Group and gr.Box, they render so poorly, we should fix or deprecate in 4.0.

Please don't 😢!  Groups are actually useful for my tasks.

pngwn:
@taesiri Yes. If you install gradio from the `main` branch (and build the frontend), you should be able to test this change.

abidlabs:
Yes @taesiri you can install from `main` by following the instructions here: https://gradio.app/docs/main/

taesiri:
@pngwn  @abidlabs Thank you both. I just use the latest commit and it worked!

apolinario:
This issue seems to be back on `3.36.1`

PR #4569, opened 2023-06-19
Update pull request template

abidlabs:


pngwn (approved):
Looks good to me! Nice change, I'd started just deleting the whole thing recently because the size of it got a little annoying.

pngwn (approved):

PR #4570, opened 2023-06-20
Improve make_waveform

akx:
# Description

This PR fixes `make_waveform` to

1. check that the file exists (e.g. if an user passes in a path, it's an actual existing file)
2. not use an insecure method to craft the ffmpeg command. As a side effect, this also makes `make_waveform` work when the input filename contains a space.

The latter doesn't seem to generally be an issue for gradio component users, as machinery exists to clean up the more evil characters from uploaded `filepath`s, but can be if one uses `make_waveform` directly. 

# Checklist:


abidlabs:
Thanks @akx, taking a look at this now. Just a reminder that we appreciate if you can create an issue first with a repro instead of opening the PR directly.

abidlabs:
Pushed a test as well. Should be good to merge once CI passes

abidlabs (approved):

PR #4572, opened 2023-06-20
Aliabd/website sveltekit test

pngwn:
testy testy


pngwn:
@aliabd Can you try out this branch please.

I've merged main in to this branch, resolved conflicts, and made some tweaks. Unfortunately, I'm still seeing the issue you described, where the `space` attribute isn't getting updated correctly, even though the value is indeed changing. Could be a bug in our code somewhere deeper but I haven't been able to figure it out just yet. It could also be a svelte/ svelte-kit bug specific to custom elements.

I have fixed it for now, by adding a `key` when we use the `gradio-app` custom element, this doesn't have any negative impact really, it does pretty much exactly what changing the `space` name directly _should_ do.

After you have tested this, feels free to merge it into your branch then we can get down to reviewing, cleaning up + finalising. the PR. It would be good to get this in ASAP so we can iterate + test docs changes quickly while we are developing 4.0. But we have a bunch of cleanup to do before we can do that.

PR #4578, opened 2023-06-20
make_waveform: raise exception if ffmpeg fails

akx:
## Description

I realized this was missing out of #4570 -- I think it should raise instead of returning the name to an empty file.

I suppose this can be marked `no-changelog-update`?


abidlabs:
Thanks @akx that makes sense. It would be better to adapt this line:

> Fix make_waveform to work with paths that contain spaces [@akx](https://github.com/akx) in https://github.com/gradio-app/gradio/pull/4570

to 

> Fix make_waveform to work with paths that contain spaces [@akx](https://github.com/akx) in https://github.com/gradio-app/gradio/pull/4570 and https://github.com/gradio-app/gradio/pull/4578

just so we can trace back changes in the future. 

Also, it would be great if we can add a quick test for this in the new `/test/test_helpers.py` file.

akx:
> It would be better to adapt this line [...]

Sure, done.

> just so we can trace back changes in the future.

Regarding tracing back changes (which I've had to do a lot in both OSS and $dayjob work), the `git blame` (or `annotate`) tool is IMO absolutely invaluable. However it'd require somewhat better commit hygiene, for lack of a better term, to work.

> Also, it would be great if we can add a quick test for this in the new `/test/test_helpers.py` file.

Done.

abidlabs:
LGTM thanks @akx!

Issue #4582, opened 2023-06-20
Possible regression: `gr.Dropdown` takes up too much vertical Space


abidlabs:
### Describe the bug

Putting a `gr.Dropdown()` in a `gr.Column()` next to another, taller `gr.Column()` leads to a problem: the `gr.Dropdown()` takes up the full vertical Space. 

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```py
import gradio as gr

with gr.Blocks() as demo:
    with gr.Row():
        with gr.Column():
            with gr.Tabs():
                with gr.Tab("a"):
                    gr.Image()
                with gr.Tab("b"):
                    pass
        with gr.Column():
            gr.Dropdown(["a", "b"]). # this Dropdown is too tall
#             gr.Textbox(["a", "b"]). # uncomment this, and you'll see that gr.Textbox() works as expected
        
demo.launch()
```

### Screenshot

<img width="799" alt="image" src="https://github.com/gradio-app/gradio/assets/1778297/31cbab63-79ed-4072-a67d-5e37c1111d0a">

### Logs

```shell
None
```

### System Info

```shell
gradio==3.35.2
```

### Severity

I can work around it

PR #4584, opened 2023-06-20
Fix scale issues within Dropdown

aliabid94:
## Description

Issue was happening for two reasons:
- Dropdown for some reason were not being treated as FormComponents because of the order of superclasses. Rearranged superclasses to match other form components. Now Dropdown will merge with other FormComponents
- The default scale value for Blocks should be null, so that the Rows can target them instead here: https://github.com/gradio-app/gradio/blob/c6c545cfe0e78aef67375f157b8f8eccf5bc4441/js/app/src/components/Row/Row.svelte#L51 which sets the scale to 1 if is a child of Row and 0 if child of Column

Closes: #4582


abidlabs (approved):

abidlabs:
Looks good @aliabid94! Can confirm that the original issue has been fixed and Dropdowns stick together now.

Moving the `FormComponent` around seems a little fragile though -- is there a better way? At the very least, we should have a test for this to prevent this regression again in the future.

PR #4586, opened 2023-06-21
Add .git-blame-ignore-revs

akx:
## Description

This PR adds an initial `.git-blame-ignore-revs` file, to [make GitHub's Blame view](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view) as well as various other tools that support it (so long as you've run `git config blame.ignoreRevsFile .git-blame-ignore-revs` in the repo directory) "skip over" e.g. large reformatting/refactoring commits, to make it easier to see the original commit changing a given line.

It currently includes the commit that merges in https://github.com/gradio-app/gradio/pull/4487, which split out `components.py` to many separate files.

## 🎯 PRs Should Target Issues

There is no existing issue for "it's hard to use Git blame on the Gradio repository", sorry. 😁


abidlabs:
Nice thanks @akx!

PR #4593, opened 2023-06-21
Eslint

pngwn:
## Description

This PR adds eslint to the repo with some simple rules. The repo doesn't not pass linting right now.

Linting is enabled in CI but the step is allowed to fail.

I've also added back the `.vscode extensions` and `.vscode recommendations` as they are very helpful for new users.


.config/.eslintrc.cjs
```diff
@@ -0,0 +1,72 @@
...
+	{}
+);
+
+const js_rules_disabled = Object.keys(js.configs.all.rules).reduce(
+	(acc, rule) => ({
+		...acc,
+		[`${rule}`]: "off"
+	}),
+	{}
+);
```
akx, line 18: Might be easier to read as
```suggestion
const ts_rules_disabled = Object.fromEntries(Object.keys(ts.rules).map(rule => [`@typescript-eslint/${rule}`, "off"]));
const js_rules_disabled = Object.fromEntries(Object.keys(js.configs.all.rules).map(rule => [rule, 'off']));
```

.config/.eslintrc.cjs
```diff
@@ -0,0 +1,72 @@
...
+		// add more generic rule sets here, such as:
+		"eslint:recommended",
+		"plugin:@typescript-eslint/recommended",
+		"plugin:svelte/recommended"
+	],
+	rules: {
+		...ts_rules_disabled,
+		...js_rules_disabled,
+		"no-console": [2, { allow: ["warn", "error"] }],
+		"no-constant-condition": 2,
```
akx, line 50: Maybe "error" instead of the seemingly magic number 2? :D

akx (commented):
Nice to see this happening in an official way :)

.config/.eslintrc.cjs
pngwn: I guess that would make more sense. :D

.config/.eslintrc.cjs
pngwn: Nice, thx!

.vscode/settings.json
```diff
@@ -0,0 +1,34 @@
...
+	"eslint.validate": ["javascript", "typescript", "html", "markdown", "svelte"],
+	"eslint.lintTask.options": ".",
+	"eslint.options": {
+		"overrideConfigFile": "./.config/.eslintrc.cjs",
+		"ignorePath": "./.config/.eslintignore"
+	},
+	"[python]": {
+		"editor.defaultFormatter": "ms-python.black-formatter"
+	}
+}
```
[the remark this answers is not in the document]
pngwn: These options are either there to make extensions work without contributors needing to loads of manual work or disable things to prevent conflicts. These aren't things anyone would want to switch off really, they are welcome to not use extensions but if they _do_ want to use them, not using these settings will cause them to fail. SO I don't think it is particularly restrictive.

I have reduced them a little by switching config formats but most of these are needed.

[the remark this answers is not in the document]
pngwn: (Most of these settings are just being reinstated from a while ago when they were incorrectly deleted).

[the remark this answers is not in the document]
pngwn: Just for context, it has taken me hours to get the correct combination of eslint + eslint extension settings to make it work, thats why i've added them. I don't think we want our contributors to go through that process.

pngwn:
Awesome. Thanks for the review @hannahblair!

pngwn:
For anyone who finds this PR post-merge, we are not planning to fix all of these errors at once and PRs doing so will be closed. We will be addressing the issues incrementally, component by component or package by package to reduce disruption. Most of this work will happen in conjunction with other improvements to the codebase and is not urgent.

However if you are working on something else, feel free to fix a few errors if they do not require significant refactoring. 

This PR is part of a general effort to increase static + dynamic checks and improve code quality.

PR #4603, opened 2023-06-21
Styling for error, warning and info toasts

hannahblair:
## Description

Implements the styling for the new Error, Warning and Info toasts. 

Screenshot 2023-06-22 at 16 26 34
Screenshot 2023-06-22 at 16 26 05

Notes:
- These toasts are currently fixed with these colours and they're not theme-able as of yet. This is something we can revisit in the future. 

- The error pill has been amended to align with the toasts (border has changed) but it is still using theme variables. 

- We no longer need the themed variables for errors. When should these be removed?

Closes: #(issue)

## 🎯 PRs Should Target Issues

## Tests & Changelog


js/app/src/components/StatusTracker/Toast.svelte
```diff
@@ -1,17 +1,47 @@
...
 </script>
 
 <div class="toast-wrap">
 	{#each messages as { type, message, id } (id)}
 		<div animate:flip={{ duration: 300 }} style:width="100%">
-			<!-- {#if type === "error"} -->
-			<Error {message} on:close {id} />
-			<!-- {/if} -->
+			{#if type === "warning"}
+				<ToastContent
```
aliabid94, line 15: let's just have one ToastContent with different theme_color and description instead of three different ToastContents

js/app/src/components/StatusTracker/ToastContent.svelte
```diff
@@ -0,0 +1,173 @@
...
+
+	afterUpdate(updateCssVarStyles);
+
+	onMount(() => {
+		setTimeout(() => {
+			close_message();
+		}, 10000);
+	});
+
+	$: updateCssVarStyles();
```
aliabid94, line 52: This won't react to changes in props since there's no variable prop in this line. If you instead remove the function and directly set cssVarStyles above as
```
$: cssVarStyles = Object.entries(...
```
then it should react to changes in theme_mode

js/app/src/components/StatusTracker/ToastContent.svelte
```diff
@@ -0,0 +1,173 @@
+<script lang="ts">
+	import { Alert, Information, Warning } from "@gradio/icons";
+	import { createEventDispatcher, onMount, afterUpdate } from "svelte";
+	import { fade } from "svelte/transition";
+	import type { ThemeMode } from "../types";
+
+	export let message: string = "";
+	export let description: string = "";
```
aliabid94, line 8: title seems like a better name than description?

js/app/src/components/StatusTracker/ToastContent.svelte
```diff
@@ -0,0 +1,173 @@
...
+		setTimeout(() => {
+			close_message();
+		}, 10000);
+	});
+
+	$: updateCssVarStyles();
+</script>
+
+<div
+	style={cssVarStyles}
```
aliabid94, line 56: I think it might be more readable if, instead of merging everything into one style string, we keep each property separate, e.g.
```svelte
<div
    style:background-color={theme["toast-bg-color"]}
    style:color={theme["toast-text-color"]}
    ...
>
```
And the theme variable will just be either lightStyles or darkStyles depending on theme_mode

js/app/src/components/StatusTracker/ToastContent.svelte
```diff
@@ -0,0 +1,173 @@
+<script lang="ts">
+	import { Alert, Information, Warning } from "@gradio/icons";
+	import { createEventDispatcher, onMount, afterUpdate } from "svelte";
+	import { fade } from "svelte/transition";
+	import type { ThemeMode } from "../types";
+
+	export let message: string = "";
+	export let description: string = "";
+	export let type: string = "";
```
aliabid94, line 9:
```suggestion
export let type: ToastMessage["type"];
```
can't be any arbitrary string

js/app/src/components/StatusTracker/ToastContent.svelte
```diff
@@ -0,0 +1,173 @@
...
+	export let theme_mode: ThemeMode;
+	export let id: number;
+
+	const dispatch = createEventDispatcher();
+
+	function close_message() {
+		dispatch("close", id);
+	}
+
+	let lightStyles = {
```
aliabid94, line 20: Actually, I think the better approach might be to keep our CSS in the CSS instead of putting it in javascript. So what I would suggest is, instead of have theme_color as a prop as well as theme_mode, we only take in the `type` and add that as a class and everything in the CSS responds to that. So the div below would become:
```
<div class="toast-body {type}">
```
and then the CSS at the bottom becomes
```
<style>
  .toast-body.error {
    background-color: var(--color-red-50);
    color: var(--color-red-700);
    ...
  }
  :global(.dark) .toast-body.error {
    background-color: var(--color-red-950);
    ...
  }

Comment (aliabid94) [reply js/app/src/components/StatusTracker/ToastContent.svelte:20]: Since we only have 3 different types, I think this is the most manageable approach.

File: js/app/src/components/StatusTracker/ToastContent.svelte [hunk 7]
```diff
@@ -0,0 +1,275 @@
...
+	const dispatch = createEventDispatcher();
+
+	function close_message() {
+		dispatch("close", id);
+	}
+
+	onMount(() => {
+		setTimeout(() => {
+			close_message();
+		}, 10000);
```
pngwn, line 20: Maybe we should make this configurable via a prop (also makes for easier testing).

js/app/src/components/StatusTracker/ToastContent.svelte
```diff
@@ -0,0 +1,275 @@
...
+	in:fade={{ duration: 200, delay: 100 }}
+	out:fade={{ duration: 200 }}
+>
+	<div class="toast-icon {type}">
+		{#if type === "warning"}
+			<Warning />
+		{:else if type === "info"}
+			<Information />
+		{:else if type === "error"}
+			<Alert />
```
pngwn, line 38: Can we rename this to `Info` to match the python class?

js/app/src/components/StatusTracker/ToastContent.svelte
```diff
@@ -0,0 +1,275 @@
...
+		position: relative;
+		top: var(--size-8);
+		right: 0;
+		left: 0;
+		align-items: center;
+		margin: var(--size-6) var(--size-4);
+		margin: auto;
+		border-radius: var(--container-radius);
+		max-width: 1200px;
+		max-width: 610px;
```
pngwn, line 73: Not sure what is going on with the duplication, but do we need this? The width is handled in the parent (or should be).

js/icons/src/Warning.svelte
```diff
@@ -0,0 +1,18 @@
+<svg
+	fill="none"
+	stroke="currentColor"
+	stroke-width="2"
+	viewBox="0 0 24 24"
+	width="33"
+	height="33"
```
pngwn, line 7: Can we make this consistent with other icons? Other icons have their width + height set too 100% and the size is controlled by the parent.

Same for the other two icons.

pngwn (approved):
Few small things but looks good otherwise.

pngwn (approved):
Awesome looks great! Thanks for this @hannahblair!

gradio/themes/base.py
```diff
@@ -1420,7 +1420,7 @@ def set(
             self, "error_background_fill_dark", "*background_fill_primary"
         )
         self.error_border_color = error_border_color or getattr(
-            self, "error_border_color", colors.red.c100
```
aliabid94, line 1423: the modal is not using this variable, correct? is this an unrelated fix?

aliabid94 (approved):
looks great!

PR #4624, opened 2023-06-21
Getting the CI to pass consistently on Windows (hopefully)

abidlabs:
We had an issue where the [backend CI tests on Windows would sometimes time out](https://huggingface.slack.com/archives/C02SPHC1KD1/p1687360760619149), maybe 40% of the time. Upon investigation, the issue was that one of the tests would try to run a Gradio demo on a port that was not available:

image

This shouldn't have been possible, because we first check to find an available port `get_first_available_port` method. *However*, it was possible that a port could become occupied between the time this method is run and the Gradio server launches. In that case, the server launch would silently fail and the script would hang forever.

What this PR does is:
* Decrease the steps between the method call and server launch
* More importantly, it also adds a server timeout -- if the server is not successfully launched within 5 seconds, then next port is tried
* Also does some minor cleanups of the code

This PR is a little tricky to test besides re-running the CI, which I've done several times and it seems like everything passes consistently now.

-- also feel free to merge to this branch directory, or merge this PR in if ready, as I'll be OOO. Thanks!


gradio/networking.py
```diff
@@ -35,8 +36,13 @@ def install_signal_handlers(self):
     def run_in_thread(self):
         self.thread = threading.Thread(target=self.run, daemon=True)
         self.thread.start()
+        start = time.time()
         while not self.started:
             time.sleep(1e-3)
+            if time.time() - start > 5:
```
abidlabs, line 42: It takes <10 ms on my computer to start the server. So 5 seconds is a very generous allotment for slow computers.

gradio/networking.py
```diff
@@ -148,20 +130,51 @@ def start_server(
...
+        [server_port]
+        if server_port is not None
+        else range(INITIAL_PORT_VALUE, INITIAL_PORT_VALUE + TRY_NUM_PORTS)
     )
-    server = Server(config=config)
-    server.run_in_thread()
+
+    for port in server_ports:
+        try:
+            # The fastest way to check if a port is available is to try to bind to it with socket.
```
abidlabs, line 141: Sorry for the copious comments, but I wanted to make sure the new logic is clear to someone who is reading the code for the first time.

gradio/networking.py
```diff
@@ -108,36 +114,12 @@ def start_server(
         app: the FastAPI app object
         server: the server object that is a subclass of uvicorn.Server (used to close the server)
     """
-    server_name = server_name or LOCALHOST_NAME
-    # if port is not specified, search for first available port
-    if server_port is None:
-        port = get_first_available_port(
```
abidlabs, line 114: I haven't removed this method `get_first_available_port` from the code since we use it in reload mode. I do think we need to overhaul reload mode at some point, and then we could remove this method too

pngwn (approved):
Will leave the code for others to review properly but looks good to me and seems to work.

gradio/networking.py
freddyaboulton: nit: Might be good to add "to start within 5 seconds." for more context. Will push this up since you're away

gradio/networking.py
freddyaboulton: I appreciate it!

gradio/networking.py
freddyaboulton: Actually this message is not displayed to the user so won't make the change

freddyaboulton (approved):
Thanks @abidlabs !

Tested reload mode, server_name, and the repro from pr https://github.com/gradio-app/gradio/issues/3675 and everything is working correctly

PR #4651, opened 2023-06-22
Share button

aliabid94:
Creates shareable button attached to Image, Gallery, Chatbot, Audio, Video.
Automatically set to True on Spaces. See example code below.

To test locally:
1) replace `https://huggingface.co/uploads` with `https://radames-temp-proxy.hf.space/https://huggingface.co/uploads` in utils.ts (to avoid CORS error)
2) no dev mode, need to `pnpm run build` to avoid CORS error
3) replace kitchen_sink/run.py with code below (to use example files)

Should also test on spaces using built package for gradio version (can be found in requirements.txt of https://huggingface.co/spaces/gradio-pr-deploys/pr-4651-all-demos)

Screen Shot 2023-06-26 at 11 22 49 PM

```python
import gradio as gr
import os
os.environ['SYSTEM'] = "spaces"
os.environ['SPACE_ID'] = "aliabid94/audiotest100"


with gr.Blocks() as demo:
    with gr.Row():
        with gr.Column():
            text = gr.Textbox()
            num = gr.Slider()
            radio = gr.Radio(["a", "b", "c"], value="a")
            file = gr.File()
            btn = gr.Button("Run")
        with gr.Column():
            img = gr.Image()
            audio = gr.Audio()
            video = gr.Video()
            gallery = gr.Gallery()
            chatbot = gr.Chatbot()

    btn.click(
        lambda *args: [
            "files/lion.jpg",
            "files/cantina.wav",
            "files/world.mp4",
            ["files/lion.jpg", "files/tower.jpg"],
            [["Hey", "I'm bot"], ["I'm human", "Ok"]],
        ],
        [text, num, radio, file],
        [img, audio, video, gallery, chatbot],
    )

if __name__ == "__main__":
    demo.launch()
```  

Closes: #2749


gradio/components/audio.py
```diff
@@ -86,6 +87,7 @@ def __init__(
             elem_id: An optional string that is assigned as the id of this component in the HTML DOM. Can be used for targeting CSS styles.
             elem_classes: An optional list of strings that are assigned as the classes of this component in the HTML DOM. Can be used for targeting CSS styles.
             format: The file format to save audio files. Either 'wav' or 'mp3'. wav files are lossless but will tend to be larger files. mp3 files tend to be smaller. Default is wav. Applies both when this component is used as an input (when `type` is "format") and when this component is used as an output.
+            shareable: If True, will allow user to share generation on Hugging Face Spaces Discussions. Can also provide a list of strings and Components that will be concatenated into the title post.
```
abidlabs, line 90: This docstring should include the behavior when `shareable` is `None` (default). Also, it's not clear to me what the behavior is when `shareable` is set to `True` (what is the title in that case?). Also, how does this concatenation happen?

I feel like this parameter is doing too much. What do you think about separating it into `show_share_button` (which mirrors the `show_copy_button` that other components have) and a separate `share_post_title` parameter?

```suggestion
            shareable: If True, will allow user to share generation on Hugging Face Spaces Discussions. Can also provide a list of strings and Components that will be concatenated into the title post.
```

js/app/src/Blocks.svelte
```diff
@@ -420,6 +437,53 @@
...
 				});
+
+			// share events
+			const FORM_INPUTS = [
+				"checkbox",
+				"dropdown",
+				"number",
+				"textbox",
+				"radio",
+				"slider"
```
abidlabs, line 448: This is not super maintainable especially with community components down the road. This is basically why we were designing the mixin to get the serialized values from the components

abidlabs: I'm wondering whether we should even support custom titles at this point? Perhaps for the v1, we simply use the filename as the custom title and then allow customization down the road depending on usage of the feature, WDYT?

abidlabs:
Btw I noticed that when images or audio are shared, they preview nicely in the discussion post. However, video does not because the `<video>` html tags are not included

image

abidlabs:
It would be good to add a note in this section https://gradio.app/sharing-your-app/#hosting-on-hf-spaces about changes that automatically apply when apps are deployed to Spaces -- so that if users are wondering how to enable/disable the button, they can find it in the docs easily

pngwn (commented):
I'm not a fan of this feature in its current form for a few reasons:

1. I think it clutters the UI. If we to add this then I think we should use the standard share arrow and lose the text. We don't have text labels for any other icons the smileys everywhere look very pushy. I also think we should position it somewhere else, not where out normal UI icons go but in another corner.
2. I don't think this is really very maintainable, every component has its own share icon. It is similar to the StatusTracker issue, updating it is a nightmare. This also impacts custom components, is every custom component expected to implement a share icon in order to get that functionality? I think this needs rethinking.
3. I don't think that sharing individual outputs makes sense in a lot of cases, a lot of the time those outputs are bound together and it makes sense to share them together. We could expand the API for this but then having the button on a specific component doesn't make much sense.
4. No ability for other kinds of sharing and no API to implement a custom share function (to share on twitter for example)

If we are ok with 3 & 4 then thats fine, it isn't a hill i'll die on. I do feel more strongly about 1 & 2 though.

aliabid94:
@pngwn @freddyaboulton @radames @gary149 @apolinario @abidlabs 
To consolidate the feedback across Slack (thread [1](https://huggingface.slack.com/archives/C02SPHC1KD1/p1686782633136469) and [2](https://huggingface.slack.com/archives/C02SPHC1KD1/p1687903366933109)) and here, here's I think a compromise of the remaining requests:
1. Some folks have been asking to make the button more prominent and others less prominent, so I will remove the label "Share" from the share button, but otherwise keep the colored icon that matches the Community icon.
2. Will make the shared post only create the body of the post from the shared component. No automatically generated title.
3. In a future PR, will add a `.share` listener that can create a completely customizable share action. When this is implemented, then in non-Space environments, the share button will use the standard share arrow. Will probably look like this in the future:
```python
with gr.Blocks() as demo:
  input_img = gr.Image()
  edit_command = gr.Textbox()
  output_image = gr.Image()

  def share(img, cmd, out):
    return f"Edit with {cmd}", f"""
      Input image: {img}, 
      Command: {cmd}
      Output image: {out} 
    """
  output_image.share(share, [input_img, edit_command, output_image])
```

aliabid94:
Implemented the above. I do think the share icon looks weird now, its unclear what it does and the color pops out a lot more. Maybe the text label could appear on hover? (I did add a hover tooltip as a `title=` but it takes time to appear) 

Screen Shot 2023-06-28 at 4 36 31 PM

@gary149 @pngwn lemme know what you guys think would make this look best. Otherwise this PR is ready.

aliabid94:
> Btw I noticed that when images or audio are shared, they preview nicely in the discussion post. However, video does not because the `<video>` html tags are not included

HF discussion bug, HTML video elements render weird, have to use URL.

pngwn:
> Implemented the above. I do think the share icon looks weird now, its unclear what it does and the color pops out a lot more. Maybe the text label could appear on hover? (I did add a hover tooltip as a title= but it takes time to appear)

This is why I think a classic share icon might be best. We can add a hover but users on mobile won't see it.

abidlabs:
> This is why I think a classic share icon might be best. We can add a hover but users on mobile won't see it.

I agree the classic share icon in grayscale is better, so that we can add sharing to Twitter, etc. in the future

Overall, I like the approach in this PR of adding to individual components for now, we can add the ability to handle more advanced sharing later on. Is it ready for a re-review @aliabid94?

aliabid94:
Changed to this:
Screen Shot 2023-06-29 at 5 16 59 PM

If we feel this is too subtle, we can ask Space developers to add "Don't forget to share your generations with the community via the share icon" text on their apps manually at first.

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 - Spaces Duplication built into Gradio, by [@aliabid94](https://github.com/aliabid94) in [PR 4458](https://github.com/gradio-app/gradio/pull/4458)
 - The `api_name` parameter now accepts `False` as a value, which means it does not show up in named or unnamed endpoints. By [@abidlabs](https://github.com/aliabid94) in [PR 4683](https://github.com/gradio-app/gradio/pull/4683)
 - Added support for `pathlib.Path` in `gr.Video`, `gr.Gallery`, and `gr.Chatbot` by [sunilkumardash9](https://github.com/sunilkumardash9) in [PR 4581](https://github.com/gradio-app/gradio/pull/4581).
+- Spaces sharing built into Gradio, by [@aliabid94](https://github.com/aliabid94) in [PR 4651](https://github.com/gradio-app/gradio/pull/4651)
```
abidlabs, line 9:
```suggestion
- The `gr.Video`, `gr.Audio`, `gr.Image`, and `gr.Gallery` components now include a share icon when deployed on Spaces. This behavior can be modified by setting the `shareable` parameter in the component classes. by [@aliabid94](https://github.com/aliabid94) in [PR 4651](https://github.com/gradio-app/gradio/pull/4651)
```

abidlabs: Can you include a screenshot here? This is a major UI change

gradio/components/audio.py
```diff
@@ -86,6 +87,7 @@ def __init__(
             elem_id: An optional string that is assigned as the id of this component in the HTML DOM. Can be used for targeting CSS styles.
             elem_classes: An optional list of strings that are assigned as the classes of this component in the HTML DOM. Can be used for targeting CSS styles.
             format: The file format to save audio files. Either 'wav' or 'mp3'. wav files are lossless but will tend to be larger files. mp3 files tend to be smaller. Default is wav. Applies both when this component is used as an input (when `type` is "format") and when this component is used as an output.
+            shareable: If True, will allow user to share generation on Hugging Face Spaces Discussions.
```
abidlabs, line 90:
```suggestion
            shareable: If True, will show a share icon in the corner of the component that allows user to share outputs to Hugging Face Spaces Discussions. If False, icon does not appear. If set to None (default behavior), then the icon appears if this Gradio app is launched on Spaces, but not otherwise.
```

abidlabs: I think we should rename this to `show_share_button` because it is more consistent with the `show_copy_button` parameter which behaves similarly

guides/01_getting-started/03_sharing-your-app.md
```diff
@@ -33,14 +33,19 @@ Share links expire after 72 hours.
...
+
 2. From your browser: Drag and drop a folder containing your Gradio model and all related files [here](https://huggingface.co/new-space).
+
 3. Connect Spaces with your Git repository and Spaces will pull the Gradio app from there. See [this guide how to host on Hugging Face Spaces](https://huggingface.co/blog/gradio-spaces) for more information. 
 
 <video autoplay muted loop>
   <source src="/assets/guides/hf_demo.mp4" type="video/mp4" />
 </video>
 
+Note: Some components, like `gr.Image`, will display a "Share" button only on Spaces, so that users can share generated output easily. You can disable this via `gr.Image(shareable=False)`
```
abidlabs, line 47: Include screenshot here as well

abidlabs, line 47:
```suggestion
Note: Some components, like `gr.Image`, will display a "Share" button only on Spaces, so that users can share the generated output to the Discussions page of the Space easily. You can disable this via `gr.Image(shareable=False)`
```

abidlabs:
@aliabid94 this is really cool! Some minor UI feedback:

* When sharing video files, it took a few seconds before anything happened. That could be confusing for users. Could we add some visual feedback (e.g. icon faded out) while the file is uploading?
* Sometimes, an incompatible file format is generated by the component. For example, in the `stt_or_tts` demo, a .flac audio file is shown on the output. When you try to share this file, the UI silently fails. It would be better if either an error modal appeared saying that "Only PNG, JPG, JPEG, GIF, MP4, MOV, QT, WEBM, MP3, MPGA, WAV files are supported" or (even better imo) the share icon did not appear if the file format was incompatible
* If the Gallery component has scroll bars, the positioning of the share icon feels a bit off:

Otherwise, looks and works great!

abidlabs (approved):

pngwn (approved):
Frontend looks good to me! I'd like us to have a better way to pass this kind of thing in in the future but it was wait for a followup/ refactor.

js/video/src/StaticVideo.svelte
```diff
@@ -59,21 +63,37 @@
...
 		>
 			<IconButton Icon={Download} label="Download" />
 		</a>
+		{#if shareable}
+			<IconButton
+				Icon={Community}
+				label="Share"
+				on:click={async () => {
+					if (!value) return;
+					let url = await uploadToHuggingFace(value.data, "url");
```
freddyaboulton, line 80: I'm confused how this works 😅 

If you load an example video from the cache it will be base64 but the upload still seems to work even though we treat it as url?

From the video_component demo: 
image

js/chatbot/src/ChatBot.svelte
```diff
@@ -82,6 +86,31 @@
...
+			Icon={Community}
+			label="Share"
+			on:click={async () => {
+				if (value === null) return;
+				const formatted_chat = await format_chat_for_sharing(value);
+				let title = "";
+				if (shareable === true) {
+					for (let msg of value[0]) {
+						if (msg !== null && typeof msg === "string") {
+							title = msg;
```
freddyaboulton, line 101: For consistency with other components we should not be setting the title for the discussion post

freddyaboulton (commented):
Fire PR @aliabid94 ! Thanks for consolidating all the feedback from all the different sources!

I left some comments (mainly around gallery). The one thing missing is a global way to disable sharing.

js/gallery/src/Gallery.svelte
```diff
@@ -248,6 +254,31 @@
 			{/each}
 		</div>
 	</div>
+	{#if shareable}
```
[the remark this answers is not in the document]
aliabid94: Believe this was because the shared content did not have flex wrap, so it got cut off. Fixed.

gradio/components/audio.py
aliabid94: Done above, except:

> shareable=True does not work if you are not on spaces

if a user is going out of their way by setting show_share_button=True outside of a Space, it is presumably to just see what the button will look like on Spaces. I don't think it's necessary to handle what to do on click yet, now just has an Error modal, in the future when we add custom share listeners we can handle better.

js/video/src/StaticVideo.svelte
aliabid94: Weird! Seems like you can `fetch()` a base64 string just like a url and it just returns the data haha! works on safari firefox and chrome so its standard behaviour. We shouldn't return base64 at all though for cached content, that's a lot of video content to convert to base64

aliabid94:
Should've addressed everything, except a global `show_share_button=False`. We don't have a concept of accessing Blocks level configuration from within a component yet - if users complain about this we can consider it then.,

Can I get a re-review @abidlabs

abidlabs:
All of my comments seem to have been addressed, thanks @aliabid94!

Just noticed a few other small issues when testing it out:

1. When sharing the audio from the `audio_debugger` tab in the deployed PR, nothing happens when I click the share button

2. When sharing the gallery from the `fake_gan` demo, I get this error even though it was working before:

image

3. Video works now!

4. When trying sharing locally, I get a pretty cryptic error message. It would be better to say that sharing is only supported if you host your demo on Spaces (and link to https://hf.space if possible from the modal)

image

abidlabs (approved):
LGTM @aliabid94! Thanks for addressing all of the feedback. One tiny thing is that the Gallery sometimes takes a while to upload for me, but that might just because it has to upload all of the images in series. Something we can try to optimize later on, but this PR is great

aliabid94:
>  One tiny thing is that the Gallery sometimes takes a while to upload for me, but that might just because it has to upload all of the images in series

Can't really parallelize download / uploads, network bandwidth is a fixed bottleneck. The reason the Gallery uploads take a while in the fake_gan scenario is that these are URLs for some large images, and we need to download each file, convert to a blob, and re-upload the file to hf.co/uploads. hf.co/uploads can't take a direct URL so we need to do this extra step.

freddyaboulton:
It's probably the same problem as https://github.com/gradio-app/gradio/issues/4841

pngwn:
@catboxanon Yeah this introduced some nasty performance issues for large apps. I'll be fixing this this week.

Issue #4385, opened 2023-06-01
Delay between file upload and change file handler with no animation


aliabid94:
### Describe the bug

If there's a File component with a change listener, there's a period of time between upload and firing the change event with no pending animation. Can be very long when uploading large files, especially on Spaces. 

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```python
import os
from zipfile import ZipFile
import gradio as gr


def zip_files(files):
    with ZipFile("tmp.zip", "w") as zipObj:
        for idx, file in enumerate(files):
            zipObj.write(file.name, file.name.split("/")[-1])
    return "tmp.zip"

with gr.Blocks() as demo:
    file = gr.File(file_count="multiple")
    zipped = gr.File()
    file.change(zip_files, file, zipped)


if __name__ == "__main__":
    demo.launch()
```

### Logs

```shell
N/A
```

### System Info

```shell
gradio==3.32.0
mac os
```

### Severity

serious, but I can work around it

PR #4661, opened 2023-06-23
fix upload file delay

aliabid94:
There was a delay between an uploaded file and then trigger for `change` and `upload` listeners. This was caused by the fact that files were actually getting uploaded *twice*, once in the svelte File component, and once in the client library when submitting data. This us because the client looks to see if there's any blob in the component data and uploads it. I removed the blob in the component so that the client does not resubmit - ideally we could let the client handle all uploading but the upload animation and dispatching of "change" and "submit" within the component won't work with that, so it's handled only in the component now. 

Fixes: #4385 

To test, run any demo that has a change or upload trigger on file, example below. Previously, you would see a lag between the finishing of the upload animation and triggering of the backend. You would also two network requests made "/upload" made sequentially. These should be fixed.

```python
import os
from zipfile import ZipFile
import gradio as gr


def zip_files(files):
    with ZipFile("tmp.zip", "w") as zipObj:
        for idx, file in enumerate(files):
            zipObj.write(file.name, file.name.split("/")[-1])
    return "tmp.zip"

with gr.Blocks() as demo:
    file = gr.File(label="a", file_count="multiple")
    zipped = gr.File(label="b")
    file.change(zip_files, file, zipped)

demo.queue().launch()
```

### DIFF
--- a/js/app/src/components/File/File.svelte
+++ b/js/app/src/components/File/File.svelte
@@ -71,7 +71,7 @@
 						(Array.isArray(_value) ? _value : [_value]).forEach(
 							async (file_data, i) => {
 								file_data.data = await blobToBase64(file_data.blob!);
-								file_data.blob = null;								
+								file_data.blob = undefined;
 							}
 						);
 					} else {
@@ -81,7 +81,7 @@
 									file_data.orig_name = file_data.name;
 									file_data.name = response.files[i];
 									file_data.is_file = true;
-									file_data.blob = null;
+									file_data.blob = undefined;
 								}
 							}
 						);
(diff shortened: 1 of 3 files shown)


pngwn:
Ah, my bad. Nice catch.

js/app/src/components/File/File.svelte
```diff
@@ -80,10 +81,11 @@
 									file_data.orig_name = file_data.name;
 									file_data.name = response.files[i];
 									file_data.is_file = true;
+									file_data.blob = null;
```
pngwn: Think we just need to use `undefined` for all of these empty blob values and those errors.

pngwn: beat me to it.

pngwn (approved):
Tested and working. Looks good to me! Thanks @aliabid94!

abidlabs:
Thanks guys! Will go ahead and merge in so that we can work on reducing the number of open PRs!

PR #4678, opened 2023-06-26
[WIP] Functional tests for `yield` and `cancel`

abidlabs:


js/app/test/cancel_events.spec.ts
```diff
@@ -0,0 +1,40 @@
+import { test, expect } from "@gradio/tootils";
+
+test("yielding", async ({ page }) => {
+	let first_iteration;
+	let fourth_iteration;
+	let last_iteration;
```
pngwn, line 6: We store these references in the correct scope so we can access them later.

js/app/test/cancel_events.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+
+	page.on("websocket", (ws) => {
```
pngwn, line 8: We listen for the websocket event, so we know it exists and can start listeneing for specific messages as soon as possible, we don't want the websocket connection to progress before we had a change to listen to it.

js/app/test/cancel_events.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+			predicate: (event) => {
+				return JSON.parse(event.payload as string)?.output?.data?.[0] === "3";
+			}
+		});
+
+		last_iteration = ws.waitForEvent("framereceived", {
+			predicate: (event) => {
+				return JSON.parse(event.payload as string).msg === "process_completed";
+			}
+		});
```
pngwn, line 25: Here we are using `waitForEvent` with a predicate function. The predicate dictates whether or not we have received the event we are waiting for. We are checking against the actual content of the message here to make sure it is the correct one.

`waitForEvent` returns a promise, we don't actual await it here because we don't want to block anything just yet. We want to interweave our 'blocks' with assertions later on. Promises in javascript will start/continue in the background if you do not await them directly. So instead of awaiting here, we assign these promises the references we defined earlier so we can wait for them later.

js/app/test/cancel_events.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+	});
+
+	const start_button = await page.locator("button", {
+		hasText: /Start Iterating/
+	});
```
pngwn, line 30: Now we've set everything up we can start kicking things off.

js/app/test/cancel_events.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+	const textbox = await page.getByLabel("Iterative Output");
+	await start_button.click();
+
+	await first_iteration;
+	await expect(textbox).toHaveValue("0");
+	await fourth_iteration;
+	await expect(textbox).toHaveValue("3");
+	await last_iteration;
+	await expect(textbox).toHaveValue("8");
```
pngwn, line 39: This is what I mean about interweaving awaits and assertions, we need to assert after we have processed a certain websocket message, then we can check that the UI correctly represents the contents of that message. The first iteration checks that we've received the first response (process_generating), and the UI correctly reflects that value.

We do the same for the others. It doesn't matter what the specifics of the predicate are as long as it waits for the correct message (roughly).

.config/playwright-setup.js
```diff
@@ -86,33 +85,14 @@ function spawn_gradio_app(app, port, verbose) {
 				res(_process);
 
 				if (!verbose) {
-					_process.stdout.destroy();
-					_process.stderr.destroy();
+					_process.stdout.off("data", std_out);
+					_process.stderr.off("data", std_out);
 				}
```
pngwn, line 90: I didn't realise that `destroy` actually destroyed the stdio functionality of the child process, I thought it just removed listeners. This was causing any python process that tried to print to the terminal to throw an exception. I've changed this to just unlisten rather than destroy.

.config/playwright-setup.js
```diff
@@ -86,33 +85,14 @@ function spawn_gradio_app(app, port, verbose) {
...
-				process.stderr.write(kl.bgBlue("\n =========== END =========== "));
-				res(_process);
+		}
 
-				if (!verbose) {
-					_process.stdout.destroy();
-					_process.stderr.destroy();
-				}
-			}
-		});
```
pngwn, line 115: I've also removed this duplication, the two callback functions are identical.

demo/cancel_events/run.py
```diff
@@ -5,7 +5,7 @@
 def fake_diffusion(steps):
     for i in range(steps):
         print(f"Current step: {i}")
-        time.sleep(1)
+        time.sleep(0.2)
```
pngwn, line 8: I reduced the time here just to speed up the test.

pngwn (commented):
I have pushed up some changes and left detailed descriptions in code comments. Let me know if you'd like to sync on this and go through it.

I also merged in main because there have been a bunch of improvements to the test setup.

abidlabs:
Thanks so much @pngwn for the detailed notes. Makes sense to me, I'll work on finishing off this test

js/app/test/cancel_events.spec.ts
```diff
@@ -0,0 +1,68 @@
+import { test, expect } from "@gradio/tootils";
+
+test("yielding", async ({ page }) => {
```
pngwn, line 3:
```suggestion
test("when using an iterative function the UI should update over time as iteration results are received", async ({ page }) => {
```

js/app/test/cancel_events.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+
+	await first_iteration;
+	await expect(textbox).toHaveValue("0");
+	await fourth_iteration;
+	await expect(textbox).toHaveValue("3");
+	await last_iteration;
+	await expect(textbox).toHaveValue("8");
+});
+
+test("cancel", async ({ page }) => {
```
pngwn, line 43:
```suggestion
test("when using an iterative function it should be possible to cancel the function, after which the UI should stop updating", async ({ page }) => {
```

pngwn (approved):
Few small comments about test description, otherwise look good! Thanks @abidlabs! Good to have tests for these more complex usecases!

abidlabs:
Merged in the latest changes, thanks @pngwn for all of the guidance

Issue #4565, opened 2023-06-19
Theme builder stopped working


cowanAI:
### Describe the bug

It wont load any theme, the buttons also stopped working as shown below:

<img width="1343" alt="image" src="https://github.com/gradio-app/gradio/assets/128092052/417e3adf-410d-475f-833d-1e138b3404c8">

https://huggingface.co/spaces/gradio/theme_builder

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

https://huggingface.co/spaces/gradio/theme_builder

### Screenshot

https://huggingface.co/spaces/gradio/theme_builder

### Logs

```shell
https://huggingface.co/spaces/gradio/theme_builder
```

### System Info

```shell
https://huggingface.co/spaces/gradio/theme_builder
```

### Severity

annoying

pngwn:
This is specific to the theme builder. The config is so large that the `/info` endpoint payload is over 500MB. Other _very_ complex apps would have a similar issue. I don't think many apps are affected by this right now (although the theme builder is an 'official' which makes this annoying).

I think we need to rethink the `info` and configs in general, if we want to avoid this issue.

abidlabs:
I have an idea for a fix here, let me try something

PR #4683, opened 2023-06-26
Updates to the API endpoints / docs, and fixing the Theme builder

abidlabs:
This WIP PR makes several changes to the API endpoints / docs:

* Allows one to pass in `False` to `api_name` to remove the endpoint from appearing in `/info` altogether
* Sets `api_name=False` to all of the event handlers in the theme builder so that the theme builder is able to launch
* Removes unnecessary boilerplate code from the JS code snippet and some other cleanup of the API docs page
* `gr.Examples` now takes in a parameter that can be used to set the api endpoint name, and this is set to `False` by default. This means that the the explicit named endpoint "load_examples" is no longer created. This was introduced in #4456 but I think it's a bad idea since it creates a second named endpoint in `gr.Interface`. This is technically a breaking change, however since it was just introduced, I think we can get away with it. 
* Makes some small formatting fixes to the client -- in part so that the client tests are triggered to make sure they all pass too 

Note that when testing the theme builder, you'll have to use your arrow keys to select the theme because the dropdown is not visible. That's being addressed in a separate PR: #4691 

Closes: #4565


client/python/gradio_client/client.py
```diff
@@ -235,7 +235,8 @@ def duplicate(
         current_hardware = (
             current_info.hardware or huggingface_hub.SpaceHardware.CPU_BASIC
         )
-        hardware = hardware or original_info.hardware
+        if hardware is None and original_info.hardware is not None:
```
abidlabs, line 238: Fixed a typing error here

client/python/gradio_client/client.py
```diff
@@ -354,7 +355,7 @@ def view_api(
         Prints the usage info for the API. If the Gradio app has multiple API endpoints, the usage info for each endpoint will be printed separately. If return_format="dict" the info is returned in dictionary format, as shown in the example below.
 
         Parameters:
-            all_endpoints: If True, prints information for both named and unnamed endpoints in the Gradio app. If False, will only print info about named endpoints. If None (default), will only print info about unnamed endpoints if there are no named endpoints.
+            all_endpoints: If True, prints information for both named and unnamed endpoints in the Gradio app. If False, will only print info about named endpoints. If None (default), will print info about named endpoints, unless there aren't any -- in which it will print info about unnamed endpoints.
```
abidlabs, line 358: Clearer language

freddyaboulton:
Is this ready for review @abidlabs ?

abidlabs:
Yes it is, thanks @freddyaboulton

gradio/blocks.py
```diff
@@ -910,7 +914,7 @@ def render(self):
             dependency_offset = len(Context.root_block.dependencies)
             for i, dependency in enumerate(self.dependencies):
                 api_name = dependency["api_name"]
-                if api_name is not None:
+                if api_name is not None and not False:
```
freddyaboulton, line 917: Is the intention to not run this code block if `api_name` is False? If so, I don't think this is working correctly. `not False` always evaluates to `True`:

image

client/python/gradio_client/client.py
freddyaboulton: What's the type error? I thought these are equivalent.

image

gradio/blocks.py
abidlabs: Lol that should be: `if api_name is not None and api_name is not False`

guides/01_getting-started/03_sharing-your-app.md
```diff
@@ -120,21 +120,21 @@ Note: if you use IFrames, you'll probably want to add a fixed `height` attribute
...
 ## API Page
 
-$demo_hello_world
+You can use almost any Gradio app as an API! In the footer of a Gradio app [like this one](https://huggingface.co/spaces/gradio/hello_world), you'll see a "Use via API" link.  
 
-If you click and open the space above, you'll see a "Use via API" link in the footer of the app. 
+![Use via API](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/gradio-guides/api3.gif)
 
-![Use via API](/assets/guides/use_via_api.png)
+This is a page that lists the endpoints that can be used to query the Gradio app, via our supported clients: either [the Python client](), or [the JavaScript client](). For each endpoint, Gradio automatically generates the parameters and their types, as well as example inputs.
```
freddyaboulton, line 127: nit - you should link to the client source or pypi/npm pages?

js/app/src/api_docs/CodeSnippets.svelte
```diff
@@ -90,18 +90,14 @@ print(result)</pre>
...
 					<pre>import &lbrace; client &rbrace; from "@gradio/client";
-<!--
--->
-
-async function run() &lbrace;
 {#each blob_examples as { label, type, python_type, component, example_input, serializer }, i}<!--
 -->
-	const response_{i} = await fetch("{example_input}");
-	const example{component} = await response_{i}.blob();
+const response_{i} = await fetch("{example_input}");
```
freddyaboulton, line 95: This looks much nicer!

js/app/src/api_docs/CodeSnippets.svelte
```diff
@@ -137,10 +133,7 @@ async function run() &lbrace;
 						{/each}
 	]);
 
-	console.log(result?.data);
-&rbrace;
-
-run();
+console.log(result.data);
```
freddyaboulton, line 136: Don't we need to keep `result?.data`

freddyaboulton (commented):
Looks good @abidlabs !

I think there is an issue where the python client will not be able to load a demo if any of the routes have `api_name=False`. We should probably add a test for this. Other than that, I think this looks about ready to merge.


image

js/app/src/api_docs/CodeSnippets.svelte
abidlabs: I don't think it's necessary but @pngwn could you confirm?

abidlabs:
> I think there is an issue where the python client will not be able to load a demo if any of the routes have api_name=False. We should probably add a test for this. Other than that, I think this looks about ready to merge.

Thanks @freddyaboulton I'll take a look at what's happening here.

> What's the type error? I thought these are equivalent.

Let me double check I may have made a mistake here


--

Thanks for the detailed review @freddyaboulton!

js/app/src/api_docs/CodeSnippets.svelte
abidlabs: Will revert unless I hear otherwise from @pngwn

abidlabs:
> I think there is an issue where the python client will not be able to load a demo if any of the routes have api_name=False. We should probably add a test for this. Other than that, I think this looks about ready to merge.

Fixed this and added a test

client/python/gradio_client/client.py
abidlabs: Ok so it was the fact that pyright didn't like `hardware` getting reassigned to have a different type. Originally `hardware` could only be a string or null, but then after this step, it could also be `SpaceHardware`. Now, it can be `SpaceHardware` from the beginning, which might be useful.

abidlabs:
Should be ready for a re-review @freddyaboulton!

client/python/test/test_client.py
```diff
@@ -791,6 +782,12 @@ def test_unnamed_endpoints_use_fn_index(self, count_generator_demo):
             assert "fn_index=0" in info
             assert "api_name" not in info
 
+    def test_api_false_endpoints_do_not_appear(self, count_generator_demo):
+        # This demo has no api_name
+        with connect(count_generator_demo) as client:
+            info = client.view_api(return_format="dict")
+            assert len(info["named_endpoints"]) + len(info["unnamed_endpoints"]) == 2
```
freddyaboulton, line 789: Wouldn't `assert len(info["unnamed_endpoints"]) == 2 and len(info["named_endpoints"]) == 0 ` be better?

freddyaboulton (approved):
Thanks for making the changes @abidlabs !

client/python/test/test_client.py
abidlabs: Yea

abidlabs: Don't know why I wrote it this way

abidlabs:
Thanks @freddyaboulton for the great review! Will merge in after CI

Issue #4687, opened 2023-06-26
.then .success events are triggered correctly


freddyaboulton:
We should have an e2e test that verifies that the `.then` event is triggered correctly. The `.then` event should be covered by the chatbot 2e2 tests.

PR #4688, opened 2023-06-26
Add success e2e test

freddyaboulton:
## Description

Closes: #4687

## 🎯 PRs Should Target Issues

## Tests & Changelog


abidlabs:
Nice!!

abidlabs (approved):

abidlabs:
Maybe good to add a test to ensure that `.success()` successfully chains more than 2 events. I.e. .success().success() works as expected when intermediate events pass/fail

pngwn:
I think we could also add a test to check the error modal shows with the correct settings set (and maybe test `gr.Error()` here as well.

freddyaboulton:
Both good suggestions - will modify the test here and ping you again

freddyaboulton:
@pngwn @abidlabs Made the suggested changes to the test!

js/app/test/blocks_chained_events.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+
+	expect(textbox).toHaveValue("Success event triggered");
+});
+
+test("Consecutive success event", async ({ page }) => {
+	const textbox = page.getByLabel("Consecutive Event");
+
+	await page.click("text=Trigger Consecutive Success");
+	// need to wait for all the events to fire
+	await page.waitForTimeout(1000);
```
abidlabs, line 27: Not sure myself, but is there a better way than to have timeouts? I wonder if we might end up with flaky tests as a result of using timeouts

demo/blocks_chained_events/run.py
```diff
@@ -0,0 +1,31 @@
...
+    with gr.Row():
+        result = gr.Textbox(label="Result")
+        result_2 = gr.Textbox(label="Consecutive Event")
+    with gr.Row():
+        success_btn = gr.Button(value="Trigger Success")
+        success_btn_2 = gr.Button(value="Trigger Consecutive Success")
+        failure_btn = gr.Button(value="Trigger Failure")
+        failure_exception = gr.Button(value="Trigger Failure With ValueError")
+
+        success_btn_2.click(success, None, None).then(success, None, None).success(lambda: "Consecutive Event Triggered", None, result_2)
```
abidlabs, line 25: Nit: this is great, as it tests `.then().success()`, but would it be also possible to test `.success().success()`?

js/app/test/blocks_chained_events.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+		page.waitForResponse("**/run/predict")
+	]);
+
+	const toast = page.getByTestId("error-toast");
+	expect(toast).toContainText("Something went wrong");
+	const close = page.getByTestId("error-close");
+	await close.click();
+	// need to let the toast fade out
+	await page.waitForTimeout(2000);
+	expect(await page.getByTestId("error-toast").count()).toEqual(0);
```
pngwn, line 44: Instead of a timeout which could fail if the animation takes a little longer, we can wait for the element to vanish and await the assertion:

```suggestion
	await expect(page.getByTestId("error-toast")).toHaveCount(0);

```

js/app/test/blocks_chained_events.spec.ts
pngwn, line 27: We can `waitForResponse` with a predicate here:

```suggestion
	let count = 0;
	await page.waitForResponse((url) => {
		if (url.url().endsWith("/run/predict")) {
			return count++ == 2;
		}

		return false;
	});
```

js/app/test/blocks_chained_events.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+		page.waitForResponse("**/run/predict")
+	]);
+
+	const toast = page.getByTestId("error-toast");
+
+	expect(toast).toContainText("Something went wrong");
+	const close = page.getByTestId("error-close");
+	await close.click();
+	await page.waitForTimeout(2000);
+	expect(await page.getByTestId("error-toast").count()).toEqual(0);
```
pngwn, line 60: Same thing here:

```suggestion
	await expect(page.getByTestId("error-toast")).toHaveCount(0);

```

js/app/test/blocks_chained_events.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+	expect(toast).toContainText("Something went wrong");
+	const close = page.getByTestId("error-close");
+	await close.click();
+	// need to let the toast fade out
+	await page.waitForTimeout(2000);
+	expect(await page.getByTestId("error-toast").count()).toEqual(0);
+});
+
+test("ValueError Triggers modal", async ({ page }) => {
+	// The demo has show_error=True so ValueError should show the modal
```
pngwn, line 48: We shouldn;t need to add comments to explain the test. Can you update the description to include this information and remove the comments? It doesn't matter if they are a bit longer, the descriptions are for humans.

js/app/test/blocks_chained_events.spec.ts
pngwn, line 27: we can `waitForResponse` with a predicate here:

```suggestion
	let count = 0;
	await page.waitForResponse((url) => {
		if (url.url().endsWith("/run/predict")) {
			return count++ == 2;
		}

		return false;
	});
```

pngwn (commented):
Few small things but looks good otherwise.

js/app/test/blocks_chained_events.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+
+test("Success event", async ({ page }) => {
+	const textbox = page.getByLabel("Result");
+	await expect(textbox).toHaveValue("");
+
+	await Promise.all([
+		page.click("text=Trigger Failure"),
+		page.waitForResponse("**/run/predict")
+	]);
+	// Since the event is not triggered, the value should not change
```
pngwn, line 11: See my comment below about comments in tests.

js/app/test/blocks_chained_events.spec.ts
freddyaboulton: Yes!

js/app/test/blocks_chained_events.spec.ts
freddyaboulton: Nice!!

js/app/test/blocks_chained_events.spec.ts
freddyaboulton: Great suggestion! But confused how this works. How does `await expect(page.getByTestId("error-toast")).toHaveCount(0);` mean that we'll wait for the element to disappear?

pngwn: Playwright has this philosophy of trying to avoid flake at all costs. The reason the 'locators' return promises (and need to be awaited) is because they will keep trying to find an element. 

In this case instead of awaiting the locator we pass the 'attempt at location' into expect. When you do that you await the expectation will will retry until it is true. 

These attempts at location will not hang indefinitely, they use the global timeout (30sec by default I think) after which the test will fail. You can set a custom timeout by doing `toHaveCount(0, { timeout: 5000 })`.

js/app/test/blocks_chained_events.spec.ts
freddyaboulton: Done!

js/app/test/blocks_chained_events.spec.ts
freddyaboulton: Very nice. Thank you for explaining!

freddyaboulton:
@pngwn Thank you for the suggestions! No more timeouts and the tests have more descriptive names cc @abidlabs

pngwn (approved):
Looks good to me! Thanks @freddyaboulton!

demo/blocks_chained_events/run.py
```diff
@@ -0,0 +1,31 @@
...
+        failure_btn = gr.Button(value="Trigger Failure")
+        failure_exception = gr.Button(value="Trigger Failure With ValueError")
+
+        success_btn_2.click(success, None, None).success(lambda: "First Event Trigered", None, result).success(lambda: "Consecutive Event Triggered", None, result_2)
+        success_btn.click(success, None, None).success(lambda: "Success event triggered", inputs=None, outputs=result)
+        failure_btn.click(failure, None, None).success(lambda: "Should not be triggered", inputs=None, outputs=result)
+        failure_exception.click(exception, None, None)
+
+if __name__ == "__main__":
+    demo.launch(show_error=True)
```
abidlabs, line 31: I don't think `show_error=True` will work because the demos are not launched according to this script, but instead the `demo` object is imported and they are mounted within the fastapi app

js/app/test/blocks_chained_events.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+	const first = page.getByLabel("Result");
+
+	await page.click("text=Trigger Consecutive Success");
+	let count = 0;
+	await page.waitForResponse((url) => {
+		if (url.url().endsWith("run/predict")) {
+			count += 1;
+			return count == 3;
+		}
+		return false;
```
abidlabs, line 36: Sorry I don't quite follow what this count is doing, could you explain?

abidlabs: It looks like it's also tripping up the CI

demo/blocks_chained_events/run.py
abidlabs: Ah but `show_error` is set to `True` when instantiating a `Blocks` so I guess it will pass.

abidlabs:
Just one question, otherwise reviewed and looks good, thanks @freddyaboulton!

abidlabs (approved):

js/app/test/blocks_chained_events.spec.ts
pngwn: We are making sure that the predict endpoint is called a specific number of time before continuing. I'm trying to figure out the flake, it passes locally.

abidlabs: Ah got it

pngwn: I think this could be a race condition since the events could be processed at varying times, although they are likely to happen one after the other there could be a chance that they dont. Hard to say without knowing more about how playwright schedules work. I am trying an alternative.

demo/blocks_chained_events/run.py
freddyaboulton: Good point regardless @abidlabs

js/app/test/blocks_chained_events.spec.ts
pngwn: I was wrong but changed it anyway, new way is cleaner.

freddyaboulton:
Thank you @pngwn and @abidlabs for the reviews and @pngwn for helping figure out the flake!!

abidlabs:
Really great test to have, thanks @freddyaboulton!

Issue #4690, opened 2023-06-26
Dropdown is cut off at the bottom of Form


abidlabs:
### Describe the bug

If you create a `gr.Dropdown()`, it will be cut off at the bottom when its parent form ends. Here's how it looks in a mild case:

<img width="617" alt="image" src="https://github.com/gradio-app/gradio/assets/1778297/8ab0dbd6-3479-414b-a8a1-ec140ad9afe2">

An even more egregious example:

<img width="610" alt="image" src="https://github.com/gradio-app/gradio/assets/1778297/ef6d2c2d-2459-439a-83c1-1e86e54cc29f">

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

Any dropdown:

```py
import gradio as gr

gr.Interface(lambda x:x, gr.Dropdown(["abc", "def"]), gr.Dropdown(["abc", "def"])).launch()
```

### System Info

```shell
gradio on main
```

### Severity

I can work around it

PR #4691, opened 2023-06-26
Dropdown cutoff fix

abidlabs:
Closes: #4690 as can be verified by kitchen sink demo:

image

Thanks @aliabid94 for the suggestion on how to debug this


js/app/src/components/Form/Form.svelte
```diff
@@ -23,7 +23,7 @@
 		border: var(--block-border-width) solid var(--border-color-primary);
 		border-radius: var(--block-radius);
 		background: var(--border-color-primary);
-		overflow: hidden;
```
freddyaboulton, line 26: Does `overflow-y` also work? Seems like it would be a more targeted fix if so

freddyaboulton:
@abidlabs The corners of the forms in the `blocks_layout` demo are cut off in the deployed demos:

### This PR

image

### Main
image

aliabid94:
> @abidlabs The corners of the forms in the blocks_layout demo are cut off in the deployed demos:

This is a good catch. I'm trying to think of the best way to fix this.

aliabid94:
Ok I came up with a solution, somewhat hacky. We create 4 "corner elements", absolutely positioned, that are just the rounded corner itself at each of the corners of the form element. It is just the rounded border and nothing else.

Let me know if you have better ideas @pngwn @hannahblair @abidlabs, happy to implement this if there's no better idea.

abidlabs:
I think that'd work @aliabid94! Keeping in mind that some themes like monochrome don't have the rounded corners at all

abidlabs:
In the meantime I'll address the other suggestions, thanks @freddyaboulton!

js/app/src/components/Form/Form.svelte
abidlabs: Made the change

aliabid94:
actually this approach feels quite hacky because I have to figure out the background color behinf the corners too - I'll wait to see if @pngwn has a better approach

aliabid94:
@pngwn what if dropdown options had position: fixed, with position calculated at open time? That should make it break out of the overflow parent, right?

aliabid94:
actually pushed a fix using `position: fixed`, not sure if there might be unintended consequences of it. Changes here: [cb8dd45](https://github.com/gradio-app/gradio/pull/4691/commits/cb8dd456b5f83209e32d26af7b346d3cff443182)

pngwn:
I'm not fan of making things `position:fixed`, i feel like that can cause lots of issues.

This CSS seems to work:

```css
	.form > :global(*) {
		box-shadow: none;
		border-width: 0px;
		border-radius: 0px;
	}

	.form > :global(*):first-child {
		border-top-left-radius: var(--block-radius);
		border-top-right-radius: var(--block-radius);
	}

	.form > :global(*):last-child {
		border-bottom-left-radius: var(--block-radius);
		border-bottom-right-radius: var(--block-radius);
	}
```

aliabid94:
I don't think this is guaranteed to work. If the form is in Row layout, then it would be different. And if the Row wraps, then it could be 4 different elements at each corner.

pngwn:
Ah yeah, you're right, even if we accounted for row/col the wrapping would cause issues. 

Let me try one last idea.

pngwn:
Lets go with the position fixed thing for now. I don't love it but it definitely works and I havent seen any issues just yet.

I think there may be some other options but we can revisit.

pngwn (changes requested):
The current implementation isn't correct, we need to recalculate the fixed position whenever we scroll or the dropdown will not move with the rest of the page. It works on spaces because the way the iframe resizer break position fixed but locally this does not work correctly:


https://github.com/gradio-app/gradio/assets/12937446/72dd4433-3d22-4c6c-9424-1e2a24fbc398

pngwn:
Yeah, you'll need to listen to the windows scroll event but you'll need to throttle it or it will kill the performance of the page.

You'll also need to double check spaces when you implement it as that is a weird environment.

aliabid94:
Made the changes https://github.com/gradio-app/gradio/pull/4691/commits/37b77275288efe4bc3b81f76b673f0b62bcb9bcd - I used `window` directly, should I use `svelte:window` instead?

pngwn:
>Made the changes https://github.com/gradio-app/gradio/commit/37b77275288efe4bc3b81f76b673f0b62bcb9bcd - I used window directly, should I use svelte:window instead?

Yeah, preferably, it is a neater abstraction with no extra code.

js/form/src/DropdownOptions.svelte
```diff
@@ -10,9 +10,30 @@
...
+	let input_width: number;
 	let refElement: HTMLDivElement;
 	let listElement: HTMLUListElement;
 	let top: string | null, bottom: string | null, max_height: number;
+	let innerHeight: number;
+
+	const calculate_window_distance = () => {
+		distance_from_top = refElement.getBoundingClientRect().top;
+		distance_from_bottom =
+			innerHeight - refElement.getBoundingClientRect().bottom;
```
akx, line 22: This could be optimized to only get the bounding rect once, like so:

```suggestion
    const {top, bottom} = refElement.getBoundingClientRect(); 
		distance_from_top = innerHeight - top;
		distance_from_bottom = innerHeight - bottom;
```

js/form/src/DropdownOptions.svelte
```diff
@@ -23,18 +44,18 @@
...
 				}
 			}
-			distance_from_top = refElement.getBoundingClientRect().top;
-			distance_from_bottom =
-				window.innerHeight - refElement.getBoundingClientRect().bottom;
+			calculate_window_distance();
 			input_height =
 				refElement.parentElement?.getBoundingClientRect().height || 0;
+			input_width =
+				refElement.parentElement?.getBoundingClientRect().width || 0;
```
akx, line 51: Same here:

```suggestion
      const rect = refElement.parentElement?.getBoundingClientRect();
			input_height = rect?.height || 0;
			input_width = rect?.width || 0;
```

js/form/src/DropdownOptions.svelte
```diff
@@ -10,9 +10,30 @@
...
 	let refElement: HTMLDivElement;
 	let listElement: HTMLUListElement;
 	let top: string | null, bottom: string | null, max_height: number;
+	let innerHeight: number;
+
+	const calculate_window_distance = () => {
+    const {top, bottom} = refElement.getBoundingClientRect(); 
+		distance_from_top = innerHeight - top;
+		distance_from_bottom = innerHeight - bottom;
+	};
```
pngwn, line 23: This logic isn't quite right. What we care about here is how close we are to the edge of the viewport. These values are always equal to one another because the height of the element is 0, so their `top` and `bottom` is always the same. `getBoundingClientRect` returns values relative to the top of viewport. We need to do this:

```js
distance_from_top = top;
distance_from_bottom = innerHeight - bottom;
```

js/form/src/DropdownOptions.svelte
```diff
@@ -10,9 +10,30 @@
...
+	let scroll_timeout: NodeJS.Timeout | null = null;
+	const scroll_listener = () => {
+		if (scroll_timeout !== null) {
+			clearTimeout(scroll_timeout);
+		}
+
+		scroll_timeout = setTimeout(() => {
+			calculate_window_distance();
+			scroll_timeout = null;
+		}, 200);
```
pngwn, line 34: I would change this to 5/ 10 ms or something. This is mainly to ensure we aren't firing thousand of callbacks off per second, we still want to do it quickly to keep things looking good.

aliabid94:
thanks for the suggestions @pngwn, should all be addressed.

pngwn (approved):
Awesome, looks great. Thanks for taking care of this @aliabid94!

PR #4692, opened 2023-06-27
fix(Image): Image mod is mask or color-sketch bug;

AmyOrz:
click undo -> the shadow line not real time change

## Description

ok, I will provide more information for you.
in the gradio/demo/blocks_mask -> io3a, io3b, io5a demo has the same bug:
1. first time I draw 3 line in canvas, get this result:
image
2.then i click the undo icon and draw other line, click submit button get this result:
image
3.click undo icon, two line is disappear. click submit button get error result; 
image

### in the gradio/demo/blocks_mask -> io3a and io3b demo. 
Operate the above three steps, get the error result: the black background image;
image


pngwn:
Have confirmed the bug and the fix.

pngwn (approved):
Looks great! Thanks for the detailed reproduction steps and the fix @AmyOrz!

PR #4693, opened 2023-06-27
Form min_width guard

akx:
## Description

This PR ensures no weird algebra is being attempted between integers and Nones for `max_width`, as was already done for `scale`, in `Form.add_child` (ref https://github.com/gradio-app/gradio/pull/4374).

On a similar note, it ensures an user attempting to do `Form(min_width=None)` will get an error immediately rather than at a later stage.

I don't have a great standalone repro for this, but the symptom is 

```
  File "modules/ui.py", line 311, in create_toprow
    with gr.Row(elem_id=f"{id_part}_tools"):
  File "gradio/blocks.py", line 366, in __exit__
    self.fill_expected_parents()
  File "gradio/blocks.py", line 358, in fill_expected_parents
    pseudo_parent.add_child(child)
  File "gradio/layouts.py", line 334, in add_child
    self.min_width += getattr(child, "min_width", 0)
TypeError: unsupported operand type(s) for +=: 'int' and 'NoneType'
```

### DIFF
--- a/gradio/layouts.py
+++ b/gradio/layouts.py
@@ -328,8 +328,9 @@ class Form(BlockContext):
     def add_child(self, child: Block):
         if isinstance(self.parent, Row):
             scale = getattr(child, "scale", None)
+            min_width = getattr(child, "min_width", None)
             self.scale += 1 if scale is None else scale
-            self.min_width += getattr(child, "min_width", 0) or 0
+            self.min_width += 0 if min_width is None else min_width
         super().add_child(child)
 
     def get_config(self):
--- a/gradio/layouts.py
+++ b/gradio/layouts.py
@@ -321,8 +321,8 @@ class Form(BlockContext):
             scale: relative width compared to adjacent Columns. For example, if Column A has scale=2, and Column B has scale=1, A will be twice as wide as B.
             min_width: minimum pixel width of Column, will wrap if not sufficient screen space to satisfy this value. If a certain scale value results in a column narrower than min_width, the min_width parameter will be respected first.
         """
-        self.scale = scale
-        self.min_width = min_width
+        self.scale = int(scale)
+        self.min_width = int(min_width)
         super().__init__(**kwargs)
 
     def add_child(self, child: Block):
--- a/test/test_layouts.py
+++ b/test/test_layouts.py
@@ -1,0 +1,14 @@
+from gradio.components import Button
+from gradio.layouts import Form, Row
+
+
+def test_form_min_width_guard():
+    """
+    Test that `Form.add_child` does not crash with a component whose `min_width` is `None`.
+    """
+    r = Row()
+    f = Form()
+    f.parent = r
+    b = Button()
+    f.add_child(b)
+    assert f.min_width == 0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 - Fix double upload bug that caused lag in file uploads by [@aliabid94](https://github.com/aliabid94) in [PR 4661](https://github.com/gradio-app/gradio/pull/4661)
 - `Progress` component now appears even when no `iterable` is specified in `tqdm` constructor by [@itrushkin](https://github.com/itrushkin) in [PR 4475](https://github.com/gradio-app/gradio/pull/4475)
 - Deprecation warnings now point at the user code using those deprecated features, instead of Gradio internals, by (https://github.com/akx) in [PR 4694](https://github.com/gradio-app/gradio/pull/4694)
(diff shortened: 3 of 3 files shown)


abidlabs:
LGTM, thanks @akx.

@aliabid94 could you also take a quick look? I believe you wrote the related code

abidlabs (approved):

gradio/layouts.py
```diff
@@ -323,15 +323,16 @@ def __init__(self, *, scale: int = 0, min_width: int = 0, **kwargs):
             scale: relative width compared to adjacent Columns. For example, if Column A has scale=2, and Column B has scale=1, A will be twice as wide as B.
             min_width: minimum pixel width of Column, will wrap if not sufficient screen space to satisfy this value. If a certain scale value results in a column narrower than min_width, the min_width parameter will be respected first.
         """
-        self.scale = scale
-        self.min_width = min_width
+        self.scale = int(scale)
```
aliabid94, line 326: why is this casting necessary? the constructor expects an int

akx: The type annotation doesn't prevent anyone from passing in something else.

This makes e.g. an inadvertent `max_width=None` (which would be valid for some other components) fail at construction time instead of when something is attempting to do math with it.

test/test_layouts.py
```diff
@@ -0,0 +1,14 @@
+from gradio.components import Button
+from gradio.layouts import Form, Row
+
+
+def test_form_min_width_guard():
+    """
+    Test that `Form.add_child` does not crash with a component whose `min_width` is `None`.
+    """
+    r = Row()
+    f = Form()
```
aliabid94, line 10: this seems to test the internal implementation, using `.parent` and `.add_child`. How about this instead, which is how Blocks are naturally built?

```
with Row() as r:
  with Form() as f:
    b = Button() 
assert f.min_width == 0 
```

gradio/layouts.py
pngwn: I don't think we should guard against people passing the wrong type in, do we do that anywhere else?

aliabid94: I see, `assert isinstance(min_width, int)` would be clearer in intent

test/test_layouts.py
akx: @aliabid94 This doesn't trigger the original bug (for whichever reason) – I did mention I didn't have a great standalone repro :(

gradio/layouts.py
akx: @aliabid94 `assert`s could be compiled out in `-O` mode. See https://beta.ruff.rs/docs/rules/assert/

akx: @pngwn Why not? I think it makes sense here since the overwhelmingly more common types for `scale` and `min_width` are

```
scale: int | None = None,
min_width: int | None = None,
```

but this component only accepts `int`s and will break otherwise.

abidlabs: Yeah I think it's a good idea to check here, even if we don't do it generally.

test/test_layouts.py
abidlabs: That's really strange, can we figure out why the bug is not being triggered?

akx: I think it has to do with the `pseudo_parent` logic in `fill_expected_parents`, and truth be told my eyes glazed over trying to figure out what happens there. 😅

aliabid94: I don't think this is a reasonable test as it uses an internal unsupported API. We have to reproduce the original problem with our supported API. Was the original bug discovered when using the API in this way as well?

akx: No, the original bug simply appears when trying to upgrade the `stable-diffusion-webui` repo to the latest version of Gradio. It doesn't use the internal APIs (but it does have some [special subclassed components](https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/master/modules/ui_components.py)).

abidlabs: @akx as a general principle, we can't be making changes to the internals of the `gradio` library based on nonstandard usage. The fix or modification should be downstream library using `gradio`. If you can find an example where using `gradio` as intended produces this bug, then we're happy to fix, but in the meantime I think we should close this PR

pngwn:
Closing until we have a better repro: https://github.com/gradio-app/gradio/pull/4693#discussion_r1244619493

test/test_layouts.py
akx: @abidlabs What constitutes "nonstandard usage"? 😁 As I see it, no non-public APIs are being used in the webui.

pngwn: We need a simple repro so we can test against regressions without testing implementation details.

PR #4203, opened 2023-05-13
Fix deprecation stacklevels

akx:
# Description

## relevant motivation

Before:
```
gradio\deprecation.py:43: UserWarning: You have unused kwarg parameters in Row, please remove them: {'equal_height': True}
  warnings.warn(
gradio\deprecation.py:43: UserWarning: You have unused kwarg parameters in Tab, please remove them: {'open': False}
  warnings.warn(
```
After:
```
a1111-sd\extensions\sd-webui-controlnet\scripts\controlnet.py:241: UserWarning: You have unused kwarg parameters in Row, please remove them: {'equal_height': True}
  with gr.Row(equal_height=True):
a1111-sd\extensions\sd-webui-controlnet\scripts\movie2movie.py:79: UserWarning: You have unused kwarg parameters in Tab, please remove them: {'open': False}
  with gr.Tab(f"ControlNet-{i}", open=False):
```

# Checklist:

- [x] I have performed a self-review of my own code
- [x] I have added a short summary of my change to the CHANGELOG.md
- [x] My code follows the style guidelines of this project
- [ ] I have commented my code in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

PR #880, opened 2022-03-25
[BIG PR] Gradio blocks & redesigned components

abidlabs:
Opening this PR as a draft so that we can start thinking about merging it into `main`. 

Of course, before we can merge it in, we need to:

- [x] Rewrite the core `Interface` class in terms of Blocks
- [x] Restore the full functionality of the `Interface` class 
- [x] Ensure that all old python unit tests pass (for backwards compatibility)
- [x] Redesign all of the components (at least at a basic level -- a full redesign can come later)

PR #4694, opened 2023-06-27
More proper deprecation warnings

akx:
## Description

This PR improves and deduplicates a whole bunch of deprecation warnings around Gradio, following up on #4203, by:

* adding some plumbing:
  * a new `GradioDeprecationWarning` warning class (since `DeprecationWarning`s aren't visible by default)
  * a `find_user_stack_level` helper inspired by Pandas' similar helper that allows an user of `warnings.warn()` to find the first user code stack level (so warnings point at user code, not at Gradio internals)
* refactoring: 
  * adding and using a single `warn_deprecation` helper that uses the above utilities
  * using a single `warn_style_method_deprecation` for the repeated `.style()` warning message from #4374
  * using a single `warn_output_deprecation` for the repeated output warning message from #880
  * using a single `warn_input_deprecation` for the repeated input warning message from #880

The end result is that the user gets pinpointed warnings, e.g.

```
modules/ui.py:1058: GradioDeprecationWarning: The `style` method is deprecated. Please set these arguments in the constructor instead.
  with gr.Row().style(equal_height=False):
modules/ui.py:1090: GradioDeprecationWarning: The `style` method is deprecated. Please set these arguments in the constructor instead.
  with gr.Row().style(equal_height=False):
```

instead of having to hunt around for these around their codebase since a single warning is only shown once by default.

Also, there's less code than before, in total.

## 🎯 PRs Should Target Issues

This PR does not target an issue, since this was yet another small annoyance I came across while attempting to upgrade Gradio in the stable-diffusion-webui project.

I can create an issue to say that "warnings aren't helpful", if that... helps.


abidlabs (approved):
Thanks @akx! LGTM + appreciate the more helpful warnings

PR #4695, opened 2023-06-27
URL validation/access fixes

akx:
## Description

I was profiling things and noticed a whole lot of time was spent in `is_valid_url` and was pretty surprised that that function actually does HTTP requests if it can.

This PR:

* makes that function share a `requests` session if it needs to do both HEAD + GET
* renames that function to `probe_url` because that's more what it does, with a deprecation trampoline for possible external users
* changes uses where `is_valid_url` was only used to check whether another function that does a `requests` request should be called to the new `is_http_url_like` function.

While working on that I realized there were small bugs in `encode_url_to_base64` and `download_tmp_copy_of_file` where they didn't check for the response status, and would thus happily end up writing e.g. an error page out.

These changes are in the `gradio_client` library, but these functions are gratuitously used by `gradio` itself.

## 🎯 PRs Should Target Issues

This PR does not target an issue right now, sorry; this was just another small papercut sort of problem I could easily fix.

### DIFF
--- a/client/python/gradio_client/utils.py
+++ b/client/python/gradio_client/utils.py
@@ -196,10 +196,11 @@ class Communicator:
 def is_valid_url(possible_url: str) -> bool:
     headers = {"User-Agent": "gradio (https://gradio.app/; team@gradio.app)"}
     try:
-        head_request = requests.head(possible_url, headers=headers)
-        if head_request.status_code == 405:
-            return requests.get(possible_url, headers=headers).ok
-        return head_request.ok
+        with requests.session() as sess:
+            head_request = sess.head(possible_url, headers=headers)
+            if head_request.status_code == 405:
+                return sess.get(possible_url, headers=headers).ok
+            return head_request.ok
     except Exception:
         return False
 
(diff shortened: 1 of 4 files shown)


client/python/gradio_client/utils.py
```diff
@@ -193,17 +194,38 @@ class Communicator:
...
     except Exception:
         return False
 
 
+def is_valid_url(possible_url: str) -> bool:
+    """
+    Check if the given string is a valid URL.
+    """
+    warnings.warn(
+        "is_valid_url is deprecated. Use is_http_url_like and probe_url, as suitable, instead.",
```
abidlabs, line 224:
```suggestion
        "is_valid_url will be deprecated. Use is_http_url_like and probe_url, as suitable, instead.",
```

abidlabs:
Nice improvements @akx!

Could we just add some tests for this:

> While working on that I realized there were small bugs in encode_url_to_base64 and download_tmp_copy_of_file where they didn't check for the response status, and would thus happily end up writing e.g. an error page out.

Should be good to merge then

abidlabs (approved):

abidlabs:
Thank you very much @akx!

PR #4705, opened 2023-06-28
4549 autoplay

### LINKED ISSUE #4549: autoplay not working when providing new files inside function
### Describe the bug

Autoplay doesn't work when multiple calls are made, even though the underlying audio file has changed. 

In the example below, when the button is pressed for the first time the first file will autoplay. However, autoplay does not work for the second file when the button is pressed the second time.

I want to create a streaming audio output so that, when new audio is received, it is automatically played. Without this being fixed I can't see a way around it.

### Reproduction

```
import gradio as gr

audio_files = [path1, path2]
idx = 0

def output():
    global idx
    file = audio_files[idx]
    idx += 1
    return file

with gr.Blocks() as demo:
    audio = gr.Audio(type="filepath", label="Audio", visible=False, autoplay=True)
    button = gr.Button(value="play")
    button.click(output, outputs=[audio])
    
demo.launch()
```

### Screenshot

_No response_

### Logs

```shell
There is no error log - the audio still plays when you press it manually.

Looking through the code, I suspect this function is the culprit: https://github.com/gradio-app/gradio/blob/ec8af6c0937a3223532779d5f1dd50312674a307/js/audio/src/Audio.svelte#L264
```

### System Info

```shell
3.35.2
```

### Severity

blocking all usage of gradio

### LINKED ISSUE #4707: Video Unit tests
@pngwn taking care of this in #4705

pngwn:
## Description
 
Closes #4549. Closes #4707

Updating the src now triggers autoplay as expected. Tests added to prevent regression.

I also experienced some flake so I have adjusted one of our test utilities a little to prevent that in the future.


js/app/src/components/Audio/Audio.test.ts
```diff
@@ -18,10 +20,14 @@ const loading_status = {
 };
 
 describe("Audio", () => {
+	beforeAll(() => {
+		window.HTMLMediaElement.prototype.play = vi.fn();
+		window.HTMLMediaElement.prototype.pause = vi.fn();
+	});
```
pngwn, line 26: I tired to mock this in a slightly more clever way but it didn't work at all, this does work but I'd like to figure out a batter way to do it in the future.

js/app/src/components/Audio/Audio.test.ts
```diff
@@ -18,10 +20,14 @@ const loading_status = {
...
 	afterEach(() => cleanup());
 
 	test("renders provided value and label", async () => {
-		const { getByTestId, queryAllByText } = render(Audio, {
+		const { getByTestId, queryAllByText } = await render(Audio, {
```
pngwn, line 30: `render` is now async because it ensure the DOM is updated before resolving. We need to await it to ensure this has all happened.

js/audio/src/Audio.svelte
```diff
@@ -216,10 +217,21 @@
...
+
+		node.addEventListener("loadeddata", handle_playback);
 		node.addEventListener("timeupdate", clamp_playback);
 
 		return {
-			destroy: () => node.removeEventListener("timeupdate", clamp_playback)
+			destroy(): void {
+				node.removeEventListener("loadeddata", handle_playback);
+				node.removeEventListener("timeupdate", clamp_playback);
+			}
```
pngwn, line 234: This is basically the fix.

js/audio/src/StaticAudio.svelte
```diff
@@ -33,20 +34,38 @@
...
+				await node.play();
+			}
+		}
+
+		node.addEventListener("loadeddata", handle_playback);
+
+		return {
+			destroy: () => node.removeEventListener("loadeddata", handle_playback)
+		};
+	}
```
pngwn, line 66: Actually I'll deupe all of this.

js/gallery/src/Gallery.svelte
```diff
@@ -161,6 +161,8 @@
 			)
 			.join(" ");
 	}
+
+	$: console.log(container_element?.tagName);
```
pngwn, line 165: remove

pngwn:
Not happy now I've reviewed, will change to wip.

pngwn:
This can be tested with the following snippet:

```py
import gradio as gr

TEST_AUDIO_A = "a.mp3"
TEST_AUDIO_B = "b.wav"
audio_files = [TEST_AUDIO_A, TEST_AUDIO_B]
a_idx = 0

TEST_VIDEO_A = "a.mp4"
TEST_VIDEO_B = "b.mp4"
video_files = [TEST_VIDEO_A, TEST_VIDEO_B]
v_idx = 0

def output_audio():
    global a_idx
    file = audio_files[a_idx]
    a_idx = 1 if a_idx == 0 else 0
    return file

def output_video():
    global v_idx
    file = video_files[v_idx]
    v_idx = 1 if v_idx == 0 else 0
    return file

with gr.Blocks() as demo:
    audio = gr.Audio(type="filepath", label="Audio", autoplay=True)
    button = gr.Button(value="play audio")
    button.click(output_audio, outputs=[audio])
    video = gr.Video(type="filepath", label="Video", autoplay=True)
    button_two = gr.Button(value="play video")
    button_two.click(output_video, outputs=[video])
    
demo.launch()
```

Use some audio files from the demos folder. When swapping between them they should autoplay

js/app/src/components/Audio/Audio.test.ts
freddyaboulton: Interesting - the same function can be be simultaneously async and not async? 🤯

js/app/src/components/Video/Video.test.ts
```diff
@@ -0,0 +1,249 @@
...
+	describe,
+	assert,
+	afterEach,
+	vi,
+	beforeAll,
+	beforeEach,
+	expect
+} from "vitest";
+import { spyOn } from "tinyspy";
+import { cleanup, render, wait_for_event, wait } from "@gradio/tootils";
```
freddyaboulton, line 12: Unused `wait_for_event` , `wait`, and `event`

js/app/src/components/Video/Video.test.ts
```diff
@@ -0,0 +1,249 @@
...
+			root_url: null,
+			streaming: false,
+			pending: false,
+			source: "microphone",
+			autoplay: true
+		});
+		const startButton = getByTestId<HTMLAudioElement>("test-player");
+		const fn = spyOn(startButton, "play");
+		startButton.dispatchEvent(new Event("loadeddata"));
+		component.$set({
```
freddyaboulton, line 183: $set is simulating the change event?

js/video/src/utils.ts
```diff
@@ -15,3 +17,22 @@ export const playable = (): boolean => {
...
 	return true; // FIX BEFORE COMMIT - mime import causing issues
 };
+
+export function loaded(
+	node: HTMLVideoElement,
+	{ autoplay }: { autoplay: boolean }
+): ActionReturn {
+	async function handle_playback(): Promise<void> {
+		if (!autoplay) return;
+		// node.pause();
```
freddyaboulton, line 27: nit: stray comment

js/app/src/components/Audio/Audio.test.ts
```diff
@@ -184,4 +191,130 @@ describe("Audio", () => {
...
+		});
+
+		const startButton = getByTestId<HTMLAudioElement>("static-audio");
+		const fn = spyOn(startButton, "play");
+		startButton.dispatchEvent(new Event("loadeddata"));
+
+		assert.equal(fn.callCount, 1);
+	});
+
+	test("when autoplay is true `media.play` should be called in static mode", async () => {
```
freddyaboulton, line 221: I think you mean dynamic mode instead of static mode in the test name here

freddyaboulton (approved):
Thank you for the fix @pngwn ! Verified it works. Left some nits - mostly questions for me.

js/app/src/components/Video/Video.test.ts
pngwn: `$set()` is the programmatic API for changing a components data.

PR #1090, opened 2022-04-26
clean-deprecated-parameters

FarukOzderim:
Closes #920 

- clean deprecated parameters
- move possible deprecation notices under deprecation.py

PR #4709, opened 2023-06-28
Fix Wrong Deprecation Warning message for type/plot parameter

freddyaboulton:
## Description
While testing #4705 , I noticed that the Video component will emit a deprecation warning telling the user to use the Number component instead which makes no sense:

image

going back to https://github.com/gradio-app/gradio/pull/1090/files, this warning should only be emitted if `type` is used for the Textbox component. However, the `Textbox` component already checks if a faulty `type` arg is passed [in](https://github.com/gradio-app/gradio/blob/main/gradio/components/textbox.py#L94) so I think we can just get rid of that warning string.

Similarly the `plot` parameter originally applied to `Image` so we should only emit it for the image component.

Added a `GradioUnusedKwargWarning` class as well since an unused kwarg is not a deprecation warning

## 🎯 PRs Should Target Issues

## Tests & Changelog


freddyaboulton:
Thank you @pngwn ! Will merge once CI passes!

PR #4714, opened 2023-06-28
scroll to top when api clicked in spaces

### LINKED ISSUE #4682: UI: scroll to top when clicking on "Use via API"
That would be quite useful (as often the content is hidden):

image

pngwn:
Closes #4682.

Click view api on a space and it should scroll to the top.


pngwn:
the PR preview is a bit messed up because there are so many endpoints on the page but it seem to work.

pngwn:
This is very similar to the error modal fix in that it is spaces only and currently cannot be tested.

abidlabs:
Nice! Taking a look now

pngwn:
@hannahblair 
> Tested and looks like the fix works! Is there a way we can get scroll-behavior: smooth work here? Not a blocker, but it would make for nicer UX. I've never tested whether that works in an iframe.

I'm not actually sure, this API is provided by the iFrameResizer library that we use and the documentation is a little thin. I'll see if I can get the behaviour argument working.

pngwn:
I checked and we can't do this from our side just yet but we can do it if we make a tweak on spaces. I'll figure out how to setup moonlanding and see if they'd be open to a PR.

Issue #4671, opened 2023-06-25
Error Updating info in components


jebarpg:
### Describe the bug

When running the code:

`textbox.update(info="new text") `

I get the following error:

`TypeError: Textbox.update() got an unexpected keyword argument 'info'`

How do I update the info text?

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

def show_info(show:bool):
    return (
        gr.Textbox.update(info="my info text here" if show else "")
	)		
		
with gr.Blocks(css="./theme.css") as demo:
    show_info_chkbox = gr.Checkbox(
        value=True, 
		label="Show Info?", 
		visible=True, 
		interactive=True
    )
	
	my_textbox = gr.Textbox(
		info="my info text here",
		placeholder="Enter some text",
		elem_id="prompt_txt",
		lines=3,
		scale=3,
		show_label=False,
	)
	
	show_info_chkbox.input(
        fn=show_info, 
		inputs=[
			show_info_chkbox
		], 
        outputs=[, 
            my_textbox,
		])
```

### System Info

```shell
Gradio version 3.35.2, windows 10, opera gx
```

### Severity

Blocking usage of gradio

jebarpg:
I discovered that info was not added to the update function, so I manually added them in my version of gradio, but I believe the should be added to gradio universally in the main repo.

abidlabs:
Thanks @jebarpg! I agree -- would you be able to open a pull request to make that contribution?

jebarpg:
> Thanks @jebarpg! I agree -- would you be able to open a pull request to make that contribution? 

Made a PR.

PR #4715, opened 2023-06-29
Updated components with 'info' attribute to update when update() is called on them.

jebarpg:
## Description

Updated components with 'info' attribute to update when update() is called on them.

Closes: #4671


pngwn (approved):
This looks great. Thanks for this @jebarpg!

I fixed the tests and tweaked the changelog.

jebarpg:
@pngwn thanks, I didn't know how to get the tests to pass, so cheers 🥂  for that!

Issue #4719, opened 2023-06-29
Telemetry appears to break the app


rsxdalv:
### Describe the bug

---> 20 demo2.launch(
     21     enable_queue=True,
     22 )

File [c:\Users\admin\Desktop\one-click-installers-tts-main\installer_files\env\lib\site-packages\gradio\blocks.py:1964](file:///C:/Users/admin/Desktop/one-click-installers-tts-main/installer_files/env/lib/site-packages/gradio/blocks.py:1964), in Blocks.launch(self, inline, inbrowser, share, debug, enable_queue, max_threads, auth, auth_message, prevent_thread_lock, show_error, server_name, server_port, show_tips, height, width, encrypt, favicon_path, ssl_keyfile, ssl_certfile, ssl_keyfile_password, ssl_verify, quiet, show_api, file_directories, allowed_paths, blocked_paths, root_path, _frontend, app_kwargs)
   1951 if getattr(self, "analytics_enabled", False):
   1952     data = {
   1953         "launch_method": "browser" if inbrowser else "inline",
   1954         "is_google_colab": self.is_colab,
   (...)
   1962         "mode": self.mode,
   1963     }
-> 1964     analytics.launched_analytics(self, data)
   1966 utils.show_tip(self)
   1968 # Block main thread if debug==True

File [c:\Users\admin\Desktop\one-click-installers-tts-main\installer_files\env\lib\site-packages\gradio\analytics.py:122](file:///C:/Users/admin/Desktop/one-click-installers-tts-main/installer_files/env/lib/site-packages/gradio/analytics.py:122), in launched_analytics(blocks, data)
    115 for x in blocks.dependencies:
    116     targets_telemetry = targets_telemetry + [
    117         # Sometimes the target can be the Blocks object itself, so we need to check if its in blocks.blocks
    118         str(blocks.blocks[y])
    119         for y in x["targets"]
    120         if y in blocks.blocks
    121     ]
--> 122     inputs_telemetry = inputs_telemetry + [
    123         str(blocks.blocks[y]) for y in x["inputs"]
    124     ]
    125     outputs_telemetry = outputs_telemetry + [
    126         str(blocks.blocks[y]) for y in x["outputs"]
    127     ]
    128 additional_data = {
    129     "version": GRADIO_VERSION,
    130     "is_kaggle": blocks.is_kaggle,
   (...)
    145     "events": [str(x["trigger"]) for x in blocks.dependencies],
    146 }

[...]

PR #4738, opened 2023-06-29
Fix for analytics error

abidlabs:
So this is a weird issue. It turns out that if you define two Blocks and reference components in one Blocks from the events in the other Blocks, our analytics errors out. 

Here's an example (note that this is invalid Gradio):

```py
import gradio as gr

with gr.Blocks() as demo1:
    t1 = gr.Textbox()

with gr.Blocks() as demo2:
    t2 = gr.Textbox()
    t2.change(lambda x:x, t2, t1)

demo2.launch()
```

Although this is invalid set of Blocks, this might be a WIP state. Anyways, the analytics error is not illuminating at all and prevents the app from launching, so I added a new `InvalidBlockError` with some diagnostic info that only gets triggered if someone tries to use the invalid event trigger. It also suggests at the fix, which is to wrap both Blocks inside a parent Blocks -- this is valid Gradio:

```py
import gradio as gr

with gr.Blocks() as demo:
    with gr.Blocks() as demo1:
        t1 = gr.Textbox()

    with gr.Blocks() as demo2:
        t2 = gr.Textbox()
        t2.change(lambda x:x, t2, t1)

demo.launch()
```

Closes: #4719
Closes: #4642


gradio/exceptions.py
```diff
@@ -23,6 +23,12 @@ class ServerFailedToStartError(Exception):
     pass
 
 
+class InvalidBlockError(ValueError):
+    """Raised when an en event in a Blocks contains a reference to a  Block that is not in the original Blocks"""
```
akx, line 27:
```suggestion
    """Raised when an event in a Blocks contains a reference to a Block that is not in the original Blocks"""
```

test/test_blocks.py
```diff
@@ -216,6 +216,17 @@ def test_initiated_analytics(self, mock_post, monkeypatch):
             pass
         mock_post.assert_called_once()
 
+    def test_launch_analytics_does_not_error_with_invalid_blocks(self, monkeypatch):
+        monkeypatch.setenv("GRADIO_ANALYTICS_ENABLED", "True")
```
akx, line 220: You may want to monkeypatch the `_do_analytics_request` function to be a no-op though so this doesn't actually send spurious analytics from every CI run.

gradio/blocks.py
```diff
@@ -1269,7 +1288,14 @@ def postprocess_data(
                     "Number of output components does not match number "
                     f"of values returned from from function {block_fn.name}"
                 ) from err
-            block = self.blocks[output_id]
+
+            try:
+                block = self.blocks[output_id]
+            except KeyError as e:
+                raise InvalidBlockError(
+                    f"Output component with id {output_id} used in {dependency['trigger']}() event not found in this gr.Blocks context. You are allowed to nest gr.Blocks contexts, but there must be a gr.Blocks context that contains all components and events."
```
akx, line 1296: For symmetry with the other message:

```suggestion
                    f"Output component with id {output_id} used in {dependency['trigger']} event not found in this gr.Blocks context. You are allowed to nest gr.Blocks contexts, but there must be a gr.Blocks context that contains all components and events."
```

akx: You might also consider using `{...!r}` for `repr` formatting.

gradio/analytics.py
```diff
@@ -120,10 +120,10 @@ def launched_analytics(blocks: gradio.Blocks, data: dict[str, Any]) -> None:
             if y in blocks.blocks
         ]
         inputs_telemetry = inputs_telemetry + [
-            str(blocks.blocks[y]) for y in x["inputs"]
+            str(blocks.blocks[y]) for y in x["inputs"] if y in blocks.blocks
```
freddyaboulton, line 123: If `y` if not in `blocks.blocks` shouldn't we also raise an InvalidBlocksError here?

abidlabs: I don't believe so. That would prevent the blocks from launching, which is what we should allow IMO because someone could be iteratively developing -- we should let the blocks launch but only raise an error if they try to use the event that is invalid

freddyaboulton: Should we not launch telemetry then? I'm worried the data would be nonsensical if the block is invalid

freddyaboulton (commented):
Thanks for the fix @abidlabs ! Left a non-blocking comment. Otherwise looks good!

gradio/analytics.py
abidlabs: Agreed, that's why we don't include the block in the list of blocks if it's invalid (i.e. we only include `blocks.blocks[y]` if `y in blocks.blocks`)

abidlabs:
Thanks @freddyaboulton and @akx for the review. Everything should be addressed, will merge shortly

Issue #2075, opened 2022-08-24
Multi-tasking on mobile (iOS/Android) seems to silently remove user from the queue


apolinario:
### Describe the bug

If I am running a Gradio demo on Android (Galaxy Fold 3) and I do multi-tasking (leave the browser app and go to another app), when I go back to the browser the processing counter is still counting - however the queue never moves, which makes me think that I got de-registered from the queue but was not informed.

--

At least I think it should be shown to user they got de-registered (the execution was stopped) - though In an the ideal world a multi-tasking phone user would not get de-registered (I think?) - but I don't know if there's anything to differentiate a user that closed the tab - which we want to dequeue, from a user that is multi-tasking and may have the tab suspended

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Run it on your Android phone

### Screenshot

![image](https://user-images.githubusercontent.com/788417/186379693-7313cf02-ab35-40a8-9c26-71158647bf55.png)

### Logs

```shell
No logs
```

### System Info

```shell
Stable Diffusion Spaces demo running on Galaxy Fold 3
```

### Severity

serious, but I can work around it

abidlabs:
Now that we show an error if the connection is disconnected, we can close this issue, right @aliabid94 @apolinario?

apolinario:
Imo ideally we would try to find ways to not disconnect the user in case they are multi-tasking on mobile. I wonder if the Page Visibility API or a "reconnect if they are back after x amount of time" would be avenues worth exploring. [Some more context for it](https://huggingface.slack.com/archives/C02QZLG8GMN/p1661757861712289?thread_ts=1661446477.396999&cid=C02QZLG8GMN) (internal)

abidlabs:
Makes sense. I believe @aliabid94 looked into it, WDYT of the feasibility of this @aliabid94?

aliabid94:
Yes it's possible and definitely needs to be done. Will require restructuring our queue system a bit. Will work on this very soon.

abidlabs:
That makes it easier to test, thanks @apolinario!

PR #4742, opened 2023-06-30
Mobile queue fixes

aliabid94:
Closes: #2075


gradio/helpers.py
```diff
@@ -893,3 +896,50 @@ def __init__(self, target: Block | None, _data: Any):
...
+    """
+    This class allows you to pass custom info messages to the user. You can do so simply with `gr.Info('message here')`, and when that line is executed the custom message will appear in a modal on the demo.
+    """
+
+    def __init__(self, message: str = "Info issued."):
+        """
+        Parameters:
+            message: The info message to be displayed to the user.
+        """
+        log_message(message, level="info")
```
akx, line 945: This API seems pretty weird – why are these classes? (Also, does it make sense for them to have a default message?)

gradio/queueing.py
```diff
@@ -415,6 +453,8 @@ async def process_events(self, events: list[Event], batch: bool) -> None:
             end_time = time.time()
             if response.status == 200:
                 self.update_estimation(end_time - begin_time)
+        except Exception as e:
+            print(e)
```
akx, line 457: 😨

gradio/helpers.py
aliabid94: I had set the wrong target branch - you probably want to comment here: https://github.com/gradio-app/gradio/pull/4518

abidlabs:
Working on testing this

js/app/src/Blocks.svelte
```diff
@@ -240,12 +240,30 @@
...
 	const MESSAGE_QUOTE_RE = /^'([^]+)'$/;
 
 	const DUPLICATE_MESSAGE =
 		"There is a long queue of requests pending. Duplicate this Space to skip.";
+	const MOBILE_QUEUE_WARNING =
+		"On mobile, the connection can break if this tab is unfocused or the device sleeps, losing your position in queue.";
+	const MOBILE_RECONNECT_MESSAGE =
+		"Lost connection due to leaving page. Rejoining queue...";
 	const SHOW_DUPLICATE_MESSAGE_ON_ETA = 15;
+	const SHOW_MOBILE_QUEUE_WARNING_ON_ETA = 10;
```
abidlabs, line 260: Perhaps also set this to 15 seconds so that we have a consistent definition of a "long waiting time" that we can communicate to users

abidlabs:
Very nice @aliabid94! Can confirm that I see this warning modal when I initiate the job:

image

(though only with queuing enabled, which I think is fine)

And I see this when I switch tabs. (Btw on my iPhone, I only see this if I set it to low power mode, otherwise the connection doesn't break):

image

abidlabs (approved):
LGTM!

freddyaboulton:
Maybe this is part of the scroll_to_output problem on spaces but on the deployed demo the warning gets automatically closed 


https://github.com/gradio-app/gradio/assets/41651716/d74a87da-bc9f-4d04-8b72-c014b80d2114

abidlabs:
Hmm @freddyaboulton it's working fine for me on iPhone (tested on Chrome and Safari)

image

Are you on iPhone?

freddyaboulton:
yes - iphone and on safari 🤔 I have an ancient iphone 10 though. let me try again.

js/app/src/Blocks.svelte
freddyaboulton: Makes sense

freddyaboulton (commented):
Modal stopped disappearing for me @aliabid94 ! Thanks for the fix

js/app/src/Blocks.svelte
aliabid94: I chose 10 seconds specifically because that's how long it seems iPhones break a websocket connection on low power mode. I don't think there's any need for these to match

akx: That should probably be documented in a comment near the constant :)

PR #4743, opened 2023-06-30
kill process properly in test teardown

pngwn:
It seem the port was being freed up but the python process was running in the background. It seem it wasn't responding to SIGTERM. I've changed this to SIGKILL and it seems to clean things up as expected.


freddyaboulton (approved):
Awesome thanks @pngwn !

Issue #4748, opened 2023-06-30
Clear button unit tests


freddyaboulton:
We should test that the clear button can correctly clear each of the component values on the page.

PR #4749, opened 2023-06-30
ClearButton browser test

freddyaboulton:
## Description

Closes: #4748

## 🎯 PRs Should Target Issues

## Tests & Changelog


js/app/test/clear_components.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+	await page.click("text=Clear");
+	await Promise.all([
+		page.waitForResponse("**/run/predict"),
+		page.click("text=Get Values")
+	]);
+	await expect(page.getByLabel("Are all cleared?")).toHaveValue("True");
+});
+
+test("Components can be hidden and removed from page", async ({ page }) => {
+	[...Array(5).keys()].forEach(async (val) => {
```
freddyaboulton, line 18: Having some difficulty testing all components. I think because not all of them fit in the page, some are not visible when they should be. But I think this is still good enough.

pngwn:
I think we should avoid end to end tests that only test one thing because of how expensive they are. I'd consider adding a clear test to one of the existing tests instead.

freddyaboulton:
I see your point @pngwn - we can end up with a really slow CI if we add lots of e2e tests willy nilly but I think it's ok to add a standalone demo for the clear button because:

1. It tests that every component can be cleared. The implementation relies on every component being able to accept a `None` from the backend side and it's possible we refactor the backend in a way that breaks that. Would be hard to work that into an existing demo.
2. It tests more than just the clear button. Every component's value initial value is being set by a python callable. That's a part of our api that's not tested anywhere else that I know of. It also tests that components can be hidden and un-hidden via the python api via `update`.

abidlabs:
I was just going to suggest that we add more to this test, particularly this:

> Every component's value initial value is being set by a python callable

As far as I can tell, this isn't currently being tested, right @freddyaboulton?

freddyaboulton:
@abidlabs if it wasn't possible the demo itself would not start but I can add specific test checking that some of the components are set properly!

abidlabs:
> @abidlabs if it wasn't possible the demo itself would not start but I can add specific test checking that some of the components are set properly!

I guess it depends on the mode of failure. For example, one failure mode could be that setting the value of a callable could simply set the value to `None`

freddyaboulton:
We check that all of the values are not-None to begin with [here](https://github.com/gradio-app/gradio/pull/4749/files#diff-c54e3d1da28d91bb9c207fe124c3af9548de362c20d779b84b010613849d6d81R21) but I also added a test case so that it's more explicit!

PR #4764, opened 2023-07-03
decode_base64_to_image: don't crash if EXIF data is broken

akx:
## Description

This adds a simple guard for `decode_base64_to_image` to not just throw when transposing the image fails.

I can't share the file that causes this issue, but the traceback is

```
  File "gradio/routes.py", line 439, in run_predict
  File "gradio/blocks.py", line 1351, in process_api
  File "gradio/blocks.py", line 1201, in preprocess_data
  File "gradio/components/image.py", line 247, in preprocess
    im = processing_utils.decode_base64_to_image(x)
  File "gradio/processing_utils.py", line 58, in decode_base64_to_image
    exif = img.getexif()
  File "python3.10/site-packages/PIL/PngImagePlugin.py", line 1028, in getexif
    return super().getexif()
  File "python3.10/site-packages/PIL/Image.py", line 1455, in getexif
    self._exif.load(exif_info)
  File "python3.10/site-packages/PIL/Image.py", line 3719, in load
    self._info = TiffImagePlugin.ImageFileDirectory_v2(self.head)
  File "python3.10/site-packages/PIL/TiffImagePlugin.py", line 507, in __init__
    raise SyntaxError(msg)
SyntaxError: not a TIFF file (header b"b'Exif\\x" not valid)
```

## 🎯 PRs Should Target Issues

This PR doesn't target an issue right now. If you like, I can open an issue that says the above.


abidlabs:
Hi @akx we definitely need an example image or some other way to reproduce this, and a test to ensure that we don't have a regression in the future

akx:
@abidlabs Guess I'll close this then 👍

akx:
@abidlabs Reopened, since I figured out a reliable repro 😁

gradio/processing_utils.py
```diff
@@ -55,10 +57,15 @@ def extract_base64_data(x: str) -> str:
 def decode_base64_to_image(encoding: str) -> Image.Image:
     image_encoded = extract_base64_data(encoding)
     img = Image.open(BytesIO(base64.b64decode(image_encoded)))
-    exif = img.getexif()
-    # 274 is the code for image rotation and 1 means "correct orientation"
-    if exif.get(274, 1) != 1 and hasattr(ImageOps, "exif_transpose"):
```
abidlabs, line 60: Why did we remove this check `exif.get(274, 1) != 1`

akx: As the commit message implies, `exif_transpose` does it by itself.

https://github.com/python-pillow/Pillow/blob/3ffa8dca446ba22f3f143331e86cfef0e9caf8f2/src/PIL/ImageOps.py#L593-L602

abidlabs:
Thanks @akx LGTM. Will merge in after CI runs

PR #4786, opened 2023-07-04
Fix: Dispatched errors from inside components bubble to top

aliabid94:
`dispatch("error")` stopped working during an error modal refactor, since we now handle errors at the top. Fixed.


js/app/src/Blocks.svelte
```diff
@@ -417,6 +418,19 @@
 					handled_dependencies[i].push(id);
 				});
 		});
+		if (!attached_error_listeners) {
```
aliabid94, line 421: Apparently `handle_mount` runs multiple times, so I had to use attached_error_listeners to make sure I only attach error listeners once. Could not figure out why handle mount runs multiple times.

pngwn: That's by design, because `onMount` from child components will run at different times with different elements being ready at different points. So we need to allow this onMount callback to rerun as many times as is necessary to ensure that we handle the mounting of all child components, that said I wanna reword this when I refactor this code shortly.

js/app/src/Blocks.svelte
```diff
@@ -417,6 +418,19 @@
...
+				if (c.instance) {
+					c.instance.$on("error", (event_data: any) => {
+						messages = [
+							new_message(event_data.detail, -1, "error"),
+							...messages
+						];
+					});
+				}
+			});
+		}
```
pngwn, line 433: Couldn't this potentially only attach a single listener when there could be multiple?

js/app/src/Blocks.svelte
aliabid94: So onMount runs every time a child is added?

js/app/src/Blocks.svelte
aliabid94: So instead of only running this once, every time we should track which components have an error listener attached, and add to the ones that don't, correct?

js/app/src/Blocks.svelte
aliabid94: ok, did something similar

aliabid94: attached_error_listeners is an array of ids now

js/app/src/Blocks.svelte
pngwn: Yes, sorry (missed this earlier).

js/app/src/Blocks.svelte
pngwn: Will re-review.

pngwn (approved):
Looks good to me! Thanks for fixing this @aliabid94!

Issue #4589, opened 2023-06-21
Image.change event is triggered twice since 3.31.0


nqdeng:
### Describe the bug

Before 3.31.0, Image.change event is only triggered once by a function update, But since 3.31.0, the event is triggered twice.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```
import gradio as gr

def onchange(image):
    print(image)

block = gr.Blocks().queue()

with block:
    temp = gr.Image(source='upload', type='pil')
    image = gr.Image(source='upload', type='pil')
    temp.change(onchange, inputs=[temp])
    image.upload(lambda x:x, inputs=[image], outputs=[temp])

block.launch(server_name='0.0.0.0', server_port=5000)
```

After the service is started, you could click the second Image control to upload an image, and will see the following logs:

```
Running on local URL:  http://0.0.0.0:5000

To create a public link, set `share=True` in `launch()`.
<PIL.Image.Image image mode=RGB size=512x512 at 0x1628DF070>
<PIL.Image.Image image mode=RGB size=512x512 at 0x162B1EF80>
```

This issue only happens when gradio version >= 3.31.0, 3.30.0 and below are fine to me.

### System Info

```shell
gradio>=3.31.0
MacOS 13.3.1
Chrome 113.0.5672.126
```

### Severity

Blocking usage of gradio

picobyte:
I've come across this issue as well. in my branch of [stable-diffusion-webui-wd14-tagger](https://github.com/toriato/stable-diffusion-webui-wd14-tagger/pull/93) I solved it with this:
```python
def on_interrogate_image_change(*args):
    # FIXME for some reason an image change is triggered twice.
    # this is a dirty hack to prevent summation/flushing the output.
    print("interrogator: "+args[1])
    It.image_counter += 1
    if It.image_counter & 1 == 0:
        return It.results()
    return on_interrogate_image(*args)


def on_interrogate_image(image: Image, interrogator: str, unload_after: bool):
    if image is None:
        return [None, None, None, 'No image']

    if interrogator not in utils.interrogators:
        return [None, None, None, f"'{interrogator}': invalid interrogator"]

    interrogator: It = utils.interrogators[interrogator]
    return interrogator.interrogate_image(image, unload_after)
```

freddyaboulton:
Thanks for filing @nqdeng and for posting the workaround @picobyte . Going to look into this.

picobyte:
[...]

Issue #4378, opened 2023-05-31
Changing Audio value programmatically doesn't trigger `.change()` event 


abidlabs:
### Describe the bug

If the value of the `Audio` component is changed programmatically, it doesn't trigger the `.change()` event. Example:

```py
import gradio as gr

with gr.Blocks() as demo:
    a = gr.Audio()
    b = gr.Audio()
    c = gr.Audio()
    
    a.change(lambda x:x, a, b)
    b.change(lambda x:x, b, c)
    
demo.launch(debug=True)
```

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

See above

### Logs

```shell
N/A
```

### System Info

```shell
3.32
```

### Severity

serious, but I can work around it

PR #4793, opened 2023-07-05
Fixes `.change()` in `Video`, `Audio`, `Image`

abidlabs:
We had multiple separate issues:

* The `change` event would get triggered twice when a file would be uploaded to `Video` or `Image`

Test with:

```py
import gradio as gr

def test(x):
    print("change called")
    return x

with gr.Blocks() as demo:
    a = gr.Video()
    b = gr.Video()
    
    a.change(test, a, b)
    
demo.launch()
```

And:

```py
import gradio as gr

def test(x):
    print("change called")
    return x

with gr.Blocks() as demo:
    a = gr.Image()
    b = gr.Image()
    
    a.change(test, a, b)
    
demo.launch()
```

or

```py
import gradio as gr

def testa():
    print("a change called")
    return 

def testb():
    print("b change called")
    return 


with gr.Blocks() as demo:
    a = gr.Image(source="canvas", tool="color-sketch")  # try different tools here
    b = gr.Image()
    c = gr.Image()
    
    a.change(lambda x:x, a, b)
    a.change(testa, None, None)
    b.change(lambda x:x, b, c)
    b.change(testb, None, None)
    
demo.launch()
```

* The `change` event would not get triggered for `Audio` if its value was changed programmatically

Try:

```py
import gradio as gr

with gr.Blocks() as demo:
    a = gr.Audio()
    b = gr.Audio()
    c = gr.Audio()
    
    a.change(lambda x:x, a, b)
    a.change(lambda x:print("a change_called"), a, None)
    b.change(lambda x:x, b, c)
    b.change(lambda x:print("b change_called"), b, None)
    
demo.launch()
```

This PR fixes these issues above

Closes: #4378, Closes: #4589

Adds unit tests and updates a demo (`change_vs_input`) that can be used to test the changes as well

### DIFF
--- a/js/app/test/blocks_chained_events.spec.ts
+++ b/js/app/test/blocks_chained_events.spec.ts
@@ -1,9 +1,7 @@
 import { test, expect } from "@gradio/tootils";
 import type { Response } from "@playwright/test";
 
-test(".success event runs after function successfully completes. .success should not run if function fails", async ({
-	page
-}) => {
+test(".success should not run if function fails", async ({ page }) => {
 	let last_iteration;
 	const textbox = page.getByLabel("Result");
 	await expect(textbox).toHaveValue("");
@@ -19,10 +17,14 @@ test(".success event runs after function successfully completes. .success should
 	await page.click("text=Trigger Failure");
 	await last_iteration;
 	expect(textbox).toHaveValue("");
+});
 
+test(".success event runs after function successfully completes", async ({
+	page
+}) => {
+	const textbox = page.getByLabel("Result");
 	await page.click("text=Trigger Success");
-	await last_iteration;
-	expect(textbox).toHaveValue("Success event triggered");
+	await expect(textbox).toHaveValue("Success event triggered");
 });
 
 test("Consecutive .success event is triggered successfully", async ({
@@ -79,18 +81,7 @@ test("ValueError makes the toast show up when show_error=True", async ({
 });
 
 test("gr.Info makes the toast show up", async ({ page }) => {
-	let complete;
-	page.on("websocket", (ws) => {
-		complete = ws.waitForEvent("framereceived", {
-			predicate: (event) => {
-				return JSON.parse(event.payload as string).msg === "process_completed";
-			}
-		});
-	});
-
 	await page.click("text=Trigger Info");
-	await complete;
-
 	const toast = page.getByTestId("toast-body");
 
 	expect(toast).toContainText("This is some info");
@@ -100,17 +91,7 @@ test("gr.Info makes the toast show up", async ({ page }) => {
 });
 
 test("gr.Warning makes the toast show up", async ({ page }) => {
-	let complete;
-	page.on("websocket", (ws) => {
-		complete = ws.waitForEvent("framereceived", {
-			predicate: (event) => {
-				return JSON.parse(event.payload as string).msg === "process_completed";
-			}
-		});
-	});
-
 	page.click("text=Trigger Warning");
-	await complete;
 
 	const toast = page.getByTestId("toast-body");
 	expect(toast).toContainText("This is a warning!");


js/app/src/components/Video/Video.svelte
```diff
@@ -57,8 +57,6 @@
 		} else {
 			value = null;
 		}
-
-		dispatch("change");
```
abidlabs, line 61: Not needed because we already dispatch change if the value changes in the subsequent lines

CHANGELOG.md
```diff
@@ -6,7 +6,7 @@ No changes to highlight.
 
 ## Bug Fixes:
 
-No changes to highlight.
+* The `.change()` event is fixed in `Video`, `Image`, `Audio` by [@abidlabs](https://github.com/abidlabs) in [PR 4793](https://github.com/gradio-app/gradio/pull/4793)
```
pngwn, line 10: Can we add a more descriptive changelog entry here explaining exactly what was fixed?

js/app/test/change_vs_input.spec.ts
pngwn, line 1: I think these should be unit tests as events can easily be tested without spinning up a browser.

abidlabs: Makes sense, I'll look into it!

abidlabs:
Thanks @pngwn for the suggestion. I've replaced the e2e test with unit tests for each of the affected components, following the approach in #4811. 

Ready for a re-review!

abidlabs:
Hmmm CI is complaining about this error: 

```
⎯⎯⎯⎯ Unhandled Rejection ⎯⎯⎯⎯⎯
ReferenceError: CustomEvent is not defined
 ❯ Module.custom_event node_modules/.pnpm/svelte@4.0.0/node_modules/svelte/src/runtime/internal/dom.js:1072:2
    1070|   * @returns {void}
    1071|   */
    1072|  c(html) {
       |  ^
    1073|   this.h(html);
    1074|  }
 ❯ node_modules/.pnpm/svelte@4.0.0/node_modules/svelte/src/runtime/internal/lifecycle.js:111:40
 ❯ dispatch_blob js/audio/src/Audio.svelte:919:3
 ❯ js/audio/src/Audio.svelte:972:5
 ❯ Object.<anonymous> js/app/src/components/Audio/Audio.test.ts:135:11

This error originated in "js/app/src/components/Audio/Audio.test.ts" test file. It doesn't mean the error was thrown inside the file itself, but while it was running.
This error was caught after test environment was torn down. Make sure to cancel any running tasks before test finishes:
```

But I don't get this error when running locally. I wonder why?

js/app/src/components/Audio/Audio.svelte
```diff
@@ -41,6 +44,13 @@
 	let _value: null | FileData;
 	$: _value = normalise_file(value, root, root_url);
 
+	$: {
```
freddyaboulton, line 47: I think we're trying to move away from JSON.stringify. I think the `Label` component has the "preferred" way but gonna defer to @pngwn

js/app/src/components/Video/Video.test.ts
```diff
@@ -125,7 +125,7 @@ describe("Video", () => {
 			root_url: null,
 			streaming: false,
 			pending: false,
-			source: "microphone",
```
freddyaboulton, line 128: Why do we need to change the source?

freddyaboulton (commented):
This looks good to me @abidlabs ! Gonna let @hannahblair @pngwn @dawoodkhan82 give the final approval.

pngwn (approved):
Looks good to me! Thanks for fixing this @abidlabs!

js/app/src/components/Audio/Audio.svelte
abidlabs: Ah good to know! Will take a look later

js/app/src/components/Video/Video.test.ts
freddyaboulton: lol

PR #4815, opened 2023-07-06
Release v3.36.0

abidlabs:
Here's hoping the next one of these release PRs is automatic ;)


abidlabs:
Went through all of the demos and they look good to me ✅

freddyaboulton:
@aliabd heads up for some reason the deployed docs get stuck when clicking around the components

deploy_website

freddyaboulton:
@aliabid94 I think `gr.Info` and `gr.Warning` are not working with generators:

This is not triggering any info or warning on main for me

```python
import gradio as gr
import time

def greet(name):
    gr.Info("Starting...")
    time.sleep(2)
    for i, letter in enumerate(name):
        if letter == " ":
            gr.Warning(f"Space detected {i}!")
        yield "Hello " + name[:i+1] + "!"
        time.sleep(0.5)
    time.sleep(2)

demo.queue(concurrency_count=10).launch()
```

PR #4816, opened 2023-07-06
Install gradio from source for chromatic

abidlabs:
Noticed that Chromatic [job was failing here](https://github.com/gradio-app/gradio/pull/4815) because it was looking for a non-existent version of `gradio`. We should be installing `gradio` from source to test the latest changes on the PR instead of the version of `gradio` that is already released to pypi

### DIFF
--- a/.github/workflows/ui.yml
+++ b/.github/workflows/ui.yml
@@ -65,7 +65,7 @@ jobs:
         with:
             fetch-depth: 0
       - name: install gradio
-        run: python -m pip install gradio==$(cat gradio/version.txt)
+        run: python -m pip install -e .
       - name: generate theme.css
         run: python scripts/generate_theme.py --outfile js/storybook/theme.css
       - name: install dependencies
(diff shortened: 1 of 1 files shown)


.github/workflows/ui.yml
```diff
@@ -65,7 +65,7 @@ jobs:
         with:
             fetch-depth: 0
       - name: install gradio
-        run: python -m pip install gradio==$(cat gradio/version.txt)
+        run: python -m pip install -e .
       - name: generate theme.css
         run: python scripts/generate_theme.py --outfile js/storybook/theme.css
       - name: install dependencies
```
freddyaboulton, line 71: Doesn't install dependencies also install gradio from source? Maybe we can just switch the order of install step and generate theme.css step

abidlabs: Nice catch. Let me try that

abidlabs:
Should be fixed now @freddyaboulton @hannahblair

freddyaboulton (approved):
Sweet thanks @abidlabs !

PR #4822, opened 2023-07-06
Prevent broken scroll_to_output in Spaces

aliabid94:
Previously, a simple demo like below was broken on mobile on spaces, because Interface would force a scroll to output, and some glitch in iframeResizer would block scrolling back. Removed scroll_to_output on spaces

```python
import gradio as gr

def do(a):
    import time
    time.sleep(5)
    return a

gr.Interface(do, "text", "text").launch()
```


abidlabs:
Can't reproduce the issue on `gradio==3.35.2`: https://huggingface.co/spaces/abidlabs/test-scroll

Seeing if something broke on `main`

gradio/blocks.py
```diff
@@ -283,7 +283,7 @@ def set_event_trigger(
             "js": js,
             "queue": False if fn is None else queue,
             "api_name": api_name,
-            "scroll_to_output": scroll_to_output,
+            "scroll_to_output": False if utils.get_space() else scroll_to_output,
```
abidlabs, line 286: You shouldn't change it here -- otherwise even if a developer explicitly sets `scroll_to_output=True` in the event, it won't take effect in Spaces. I would change the the default value in `set_event_trigger()`

freddyaboulton:
https://github.com/gradio-app/gradio/assets/41651716/532b88e8-b1e7-4dc3-968c-172e18ceeba0

I can recreate on main (see above) and it’s fixed in this or (see below):


https://github.com/gradio-app/gradio/assets/41651716/c6f1e9b3-bbdb-47e8-91aa-97445400e592

So this looks good to me from that standpoint!

freddyaboulton:
Really confused why we’re seeing different behavior though. Maybe I have some ancient cookie in my safari browser? Or maybe my safari version is really old?

gradio/blocks.py
aliabid94: Disagree, it will break the app on Spaces, so even if it's explicitly allowed we stop it on Spaces. If a user sets it locally and then uploads it to spaces, it shouldn't break the app on spaces.

aliabid94:
Thanks for testing!

gradio/blocks.py
abidlabs: 👍

Issue #4821, opened 2023-07-06
Navigation bar on changelog page not working


freddyaboulton:
### Describe the bug

The navigation side bar on the changelog page is not working

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

https://www.gradio.app/changelog

![changelog_navbar](https://github.com/gradio-app/gradio/assets/41651716/b199ceb4-7228-4293-9597-560bb8f2088c)

### System Info

```shell
-
```

### Severity

I can work around it

PR #4828, opened 2023-07-07
Fix Changelog Navbar in Website

aliabd:
The [navbar on the changelog](https://www.gradio.app/changelog) doesn't do anything because I never created anchor links for the headers and added hrefs. This fixes that. 

Closes: #4821


freddyaboulton (approved):
Works like a charm @aliabd !

PR #4856, opened 2023-07-10
Update chromatic job with build details

hannahblair:
## Description

This PR moves the chromatic build job to on push, as [according to their docs](https://www.chromatic.com/docs/github-actions#recommended-configuration-for-build-events), running it on a `pull_request` event can be problematic. To get the PR for the push event that triggered the workflow, I've used a new action: [8BitJonny/gh-get-current-pr@2.2.0](https://github.com/8BitJonny/gh-get-current-pr).  This checks whether a PR associated with the push commit exists, and if so then proceeds with the PR comment job. 

Also: 

- Details about the build (changes and failed tests) have been added to the PR comment. We have the option to include the build exit code but not a text version of that without creating a .json file of the [exit codes and the corresponding descriptions](https://www.chromatic.com/docs/cli#exit-codes). All the output options from the chromatic CLI are [here](https://github.com/chromaui/chromatic-cli/blob/main/action.yml#L100). If you think any others would be valuable in the PR comment lmk. 
- Changes the PR comment when a new build is running

Closes: #(issue)


pngwn:
Looks good! Let get it merged to test it properly. 

Thanks for refining this workflow @hannahblair!

pngwn (approved):

abidlabs:
Small nit, but otherwise LGTM! This is very helpful: 

image

abidlabs (approved):

Issue #4104, opened 2023-05-08
Component by Events Table Not Showing Correct Data


freddyaboulton:
### Describe the bug

The table that displays the events that each component supports does not show the correct data. Every component has an `x` for every event.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

<img width="1084" alt="image" src="https://user-images.githubusercontent.com/41651716/236828725-0488d38e-5d78-40fa-99cb-c9a3fdbf0a18.png">

### Logs

```shell
-
```

### System Info

```shell
Docs on latest stable: https://gradio.app/docs/#components
```

### Severity

annoying

Issue #4804, opened 2023-07-06
Issues with website after migration 


aliabd:
List of various issues with the website after the migration to the sveltekit: 

- [ ] Guides need to be redirected from /[guide] to /guides/[guide]
- [ ] Trailing slash causing 404s 
- [ ] Undesigned 404 page
- [ ] Advanced Interface Features guide has horizontal scrolling 
- [ ] Guides headers on navbar not getting updated 
- [ ] AnnotatedImage doesn't have an embedded component
- [ ] Missing meta tags on Twitter

abidlabs:
It looks like we lost the docs for `gr.Group` in the migration. (This was previously reported as an issue: https://github.com/gradio-app/gradio/issues/4338)

abidlabs:
Nit: the font size for the code in all of the guides is much larger than the surrounding text:

<img width="931" alt="image" src="https://github.com/gradio-app/gradio/assets/1778297/af0fcf93-5597-4b83-9434-616570e43f32">

abidlabs:
Also it looks like the code blocks don't have copy buttons anymore?

PR #4863, opened 2023-07-10
Fix component events matrix in docs

aliabd:
Components events matrix in the docs is broken, this PR fixes it and adds the missing events too. 

Closes one issue in #4804:
- [x] Components events matrix not showing any event

Closes: #4104


freddyaboulton:
Also closes #4104

freddyaboulton:
Nice @aliabd ! How do I test this? It's not showing up correctly for me in the preview

image

aliabd:
@freddyaboulton yup missed that issue! 

I'm not sure why it isn't showing up correctly either, debugging now.

aliabd:
@freddyaboulton just fixed, please take another look :)

js/_website/generate_jsons/src/docs/__init__.py
```diff
@@ -54,35 +54,38 @@ def add_demos():
...
-    "Play()",
-    "Pause()",
-    "Stream()",
-    "Blur()",
-    "Upload()",
-]
-
-
-def add_supported_events():
+events = [
```
freddyaboulton, line 57: How can we keep this up to date in the future?

Maybe we can programmatically import all `EventListeners` and programmatically document all the properties that are of instance `EventListenerMethod`. Or we have `EventListenerMethod` update a global dict with the event name?

aliabd:
thanks @freddyaboulton was thinking the same as i was making the PR. I'll go through and refactor a lot of this soon.

js/_website/generate_jsons/src/docs/__init__.py
aliabd: so I actually got stuck on this earlier: when I list the attributes of one of the EventListeners classes, it doesn't show any of the EventListenerMethods.. not sure why.

```
<class 'gradio.events.Changeable'>
{'__module__': 'gradio.events', '__init__': <function Changeable.__init__ at 0x7fcf3004f1f0>, '__doc__': None}
```

aliabd: just fixed it @freddyaboulton, not hardcoded anymore. can you take another look? 

Btw what is 'end'? not showing up as a method for any component

js/_website/generate_jsons/src/docs/__init__.py
```diff
@@ -54,35 +55,26 @@ def add_demos():
...
-]
-
-
-def add_supported_events():
+def create_events_matrix():
+    events = []
+    for c in EventListener.__subclasses__():
+        methods = c().__dict__
+        for m in methods: 
+            if m[:1] != '_' and isinstance(methods[m], EventListenerMethod) and m[:1] not in events:
```
freddyaboulton, line 63: I think you mean `m not in events` here?

freddyaboulton: Otherwise looks good!

aliabd: lol

js/_website/generate_jsons/src/docs/__init__.py
freddyaboulton: The end event should be triggered when the video/audio stops playing. We should add it to document decorator for Playable class. I think that was an oversight.

aliabd: got it thanks! added it, and showing up correctly with video and audio :) will merge. thanks for the help!

Issue #3510, opened 2023-03-19
Simpler Chatbot Component / API


abidlabs:
Building a Chatbot is a very common use case that is currently fairly complex to do with Gradio. Here's all the code that's needed for the simple Chatbot: 

```py
import gradio as gr
import random
import time

with gr.Blocks() as demo:
    chatbot = gr.Chatbot()
    msg = gr.Textbox()
    clear = gr.Button("Clear")

    def user(user_message, history):
        return "", history + [[user_message, None]]

    def bot(history):
        bot_message = random.choice(["Yes", "No"])
        history[-1][1] = bot_message
        time.sleep(1)
        return history

    msg.submit(user, [msg, chatbot], [msg, chatbot], queue=False).then(
        bot, chatbot, chatbot
    )
    clear.click(lambda: None, None, chatbot, queue=False)

demo.launch()
``` 

We should consider having a custom component / higher-level abstraction for the Chatbot.

pGit1:
I agree. :( @abidlabs. Your posts are useful but this is really hard unless yo spend alot of time figuring it out. 

For instance how do I get a "regenerate" tab to work (to get a new chat generation for the same input in case it was poor)? Secondly how do I get this chatbot to produce outputs as it is typing like chatgpt or even this demo does: https://chat.lmsys.org/?

Its really hard and I find myself unable to understand what is going on in Gradio code alot of times as it takes away from ml development. 

If any thing it would be nice to just drop models into a predefined template chatbot or some thing with several toggleable features.

abidlabs:
Thanks @pGit1 for your questions and feedback. We'll think about this as we build towards 4.0.

In the meantime, you can get streaming responses by using `yield` instead of `return`, as described here; https://gradio.app/creating-a-chatbot/#add-streaming-to-your-chatbot

For regenerate, you would simply rerun the bot() method when the regenerate button is pressed.

pGit1:
@abidlabs the streaming output makes sense I think, and I will update my code accordingly. 

Thanks for the event listeners section. I briefly looked over it but dont seem to find the answer for regenerate button related to the answer you provided.

But Concerning the "regenerate" button would my code look something like: 

```
def return_bot():
    return bot

regenerate.click(return_bot, inputs=None, outputs=None)
# or 
regenerate.click(lambda:None, None, return_bot,  queue=False)
```

pGit1:
@abidlabs I suspected this would not work for me:

```
  def bot(history):
[...]

PR #4869, opened 2023-07-10
A high-level `ChatInterface` abstraction

abidlabs:
Simple examples of usage:

```py
import gradio as gr

def echo(message, history):
   return message

gr.ChatInterface(echo).launch()
```

```py
from gradio import ChatInterface
import time

def double(message, history):
    time.sleep(1)
    return message + message

ChatInterface(double, examples=["bbbb", "abcdef"], cache_examples=True).queue().launch()
```

```py
from gradio import ChatInterface
import time

def echo_stream(message, history):
    for i in range(len(message)):
        time.sleep(0.3)
        yield message[:i]

ChatInterface(echo_stream).queue().launch()
```

The UI:

image

We'll address in future PRs if there's demand:
* Handle async fns and async generators
* Convert submit button to stop button while streaming
* Add other helpful API routes

Closes: #3510


osanseviero (commented):
Cool stuff!

gradio/chat_interface.py
```diff
@@ -0,0 +1,193 @@
...
+        delete_last_btn: str | None | Button = None,
+        clear_btn: str | None | Button = "🗑️  Clear History",
+    ):
+        """
+        Parameters:
+            fn: the function to wrap the chat interface around. Should accept two parameters: a string input message and list of two-element lists of the form [[user_message, bot_message], ...] representing the chat history, and return a string response. See the Chatbot documentation for more information on the chat history format.
+            chatbot: an instance of the gr.Chatbot component to use for the chat interface. If not provided, a default gr.Chatbot component will be created.
+            textbox: an instance of the gr.Textbox component to use for the chat interface. If not provided, a default gr.Textbox component will be created.
+            examples: sample inputs for the function; if provided, appear below the chatbot and can be clicked to populate the chatbot input.
+            cache_examples: If True, caches examples in the server for fast runtime in examples. The default option in HuggingFace Spaces is True. The default option elsewhere is False.
```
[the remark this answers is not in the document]
abidlabs: Thanks @osanseviero I'll take a look

gradio/chat_interface.py
```diff
@@ -0,0 +1,268 @@
...
+
+    def _api_submit_fn(self, message: str, history: list[list[str]]):
+        history = self._display_input(message, history)
+        history = self._submit_fn(message, history)
+        return history
+        
+    def _api_stream_fn(self, message: str, history: list[list[str]]):
+        history = self._display_input(message, history)
+        for response in self._stream_fn(message, history):
+            return response
```
abidlabs, line 258:
```suggestion
            yield response
```

gradio/chat_interface.py
```diff
@@ -0,0 +1,268 @@
...
+        if inspect.isgeneratorfunction(self.fn):
+            api_fn = self._api_stream_fn
+        else:
+            api_fn = self._api_submit_fn
+
+        # Use a gr.State() instead of self.chatbot so that the API doesn't require passing forth
+        # a chat history, instead it is just stored internally in the state.
+        history = State([])
+
+        self.fake_api_btn.click(
```
freddyaboulton, line 226: Something's not right here - the view api page and client both show that this route takes and returns nothing

image

image

This is with this demo

```python
from gradio import ChatInterface
import time

def double(message, history):
    time.sleep(1)
    return message + message

ChatInterface(double, examples=["bbbb", "abcdef"], cache_examples=True).queue().launch()
```

freddyaboulton:
Thanks @abidlabs ! Will take a closer look tomorrow

gradio/chat_interface.py
```diff
@@ -0,0 +1,268 @@
...
+            mode="chat_interface",
+            css=css,
+            title=title or "Gradio",
+            theme=theme,
+        )
+        self.fn = fn
+        self.examples = examples
+        self.cache_examples = cache_examples
+        self.buttons: list[Button] = []
+        self.history = []
```
freddyaboulton, line 76: nit: this isn't used anywhere

abidlabs: Removed

gradio/chat_interface.py
abidlabs: Should be fixed now, thanks @freddyaboulton!

gradio/chat_interface.py
```diff
@@ -0,0 +1,269 @@
...
+
+    def _api_submit_fn(self, message: str, history: list[list[str]]):
+        response = self.fn(message, history)
+        history.append([message, response])
+        return response, history
+        
+    def _api_stream_fn(self, message: str, history: list[list[str]]):
+        history = self._display_input(message, history)
+        for response in self.fn(message, history):
+            history[-1][1] = response
```
freddyaboulton, line 258: history can actually be a `List[Tuple[str]]` which will error out when trying to do the item assignment. This is caused by the client since it converts everything to tuples (I think we should remove in 4.0). 


Repro


image

```bash
  File "/home/user/.local/lib/python3.10/site-packages/gradio/routes.py", line 439, in run_predict
  File "/home/user/.local/lib/python3.10/site-packages/gradio/blocks.py", line 1384, in process_api
    result = await self.call_function(
  File "/home/user/.local/lib/python3.10/site-packages/gradio/blocks.py", line 1103, in call_function
    prediction = await utils.async_iteration(iterator)
  File "/home/user/.local/lib/python3.10/site-packages/gradio/utils.py", line 343, in async_iteration
    return await iterator.__anext__()
  File "/home/user/.local/lib/python3.10/site-packages/gradio/utils.py", line 336, in __anext__
    return await anyio.to_thread.run_sync(
  File "/home/user/.local/lib/python3.10/site-packages/anyio/to_thread.py", line 33, in run_sync
    return await get_asynclib().run_sync_in_worker_thread(
  File "/home/user/.local/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 877, in run_sync_in_worker_thread
    return await future
  File "/home/user/.local/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 807, in run
  File "/home/user/.local/lib/python3.10/site-packages/gradio/utils.py", line 319, in run_sync_iterator_async
    return next(iterator)
  File "/home/user/.local/lib/python3.10/site-packages/gradio/utils.py", line 688, in gen_wrapper
    yield from f(*args, **kwargs)
  File "/home/user/.local/lib/python3.10/site-packages/gradio/chat_interface.py", line 258, in _api_stream_fn
    history[-1][1] = response
TypeError: 'tuple' object does not support item assignment
```

gradio/chat_interface.py
```diff
@@ -0,0 +1,269 @@
...
+                    if isinstance(submit_btn, Button):
+                        submit_btn.render()
+                    elif isinstance(submit_btn, str):
+                        submit_btn = Button(submit_btn, variant="primary")
+                    else:
+                        raise ValueError(f"The submit_btn parameter must be a gr.Button, string, or None, not {type(submit_btn)}")
+                    self.buttons.append(submit_btn)
+                self.stop_btn = Button("Stop", variant="stop", visible=False)
+                
+                for btn in [retry_btn, delete_last_btn, clear_btn]:
```
freddyaboulton, line 106: nit: why not use a gr.ClearButton here?

gradio/chat_interface.py
```diff
@@ -0,0 +1,269 @@
...
+                    self.chatbot = chatbot.render()
+                else:
+                    self.chatbot = Chatbot(label="Input")
+                if textbox:
+                    self.textbox = textbox.render() 
+                else: 
+                    self.textbox = Textbox(show_label=False, placeholder="Type a message...")
+            
+            with Row():
+                if submit_btn:
```
freddyaboulton, line 96: I think the UI would look a lot sleeker if the submit button was next to the textbox

Screenshot 2023-07-13 at 10 57 11 AM

Similar to ChatGPT, HuggingChat, and iOS

gradio/chat_interface.py
```diff
@@ -0,0 +1,269 @@
...
+        )
+
+    def _clear_and_save_textbox(self, message):
+        return "", message
+
+    def _display_input(self, message: str, history: list[list[str]]):
+        history.append((message, None))
+        return history
+
+    def _submit_fn(self, message: str, history: list[list[str]]):
```
freddyaboulton, line 240: So the history contains the user message (since we call _display_history)? I think that's confusing. I would expect the history to only contain the history up to but not including the current message.

gradio/chat_interface.py
abidlabs: Nice catch

gradio/chat_interface.py
abidlabs: Sure, will update

gradio/chat_interface.py
abidlabs: Ok let me play around with that and see how it looks on different screen sizes

gradio/chat_interface.py
abidlabs: Yeah its confusing since In `_display_input`, `history` contains the history not including the current message, but by the time that `_submit_fn` is called, `history` includes the current message. These are all internal methods so I don't think it matters too much, but I'll rename `history` to be `history_with_input` or something

abidlabs:
Thanks for the initial review. I'm going to be working on addressing these comments & finishing off the remaining items

gradio/chat_interface.py
freddyaboulton: I'm worried this will introduce bugs into demos built with ChatInterface. I think it would be natural for a developer to do something like the following (pseudocode):

```python
def chat(message, history):
    chain.history = history
    return chain.predict(message)
```
And then pass duplicate entries for `message` to the chain. BTW I think there is an inconsistency between `_api_stream_fn` and `_api_fn` the latter does not include the latest message in the chain but the former does.

Whatever we pick we should be consistent but I think it's more natural for the history to not include the latest message.

abidlabs:
Feedback from @yvrjsharma along with my comments:

> I wondered why we need to explicitly mention the delete button name to include it in the UI. While it's not the case with Retry and clear buttons.

Yeah I think I'll make the delete button appear by default, especially if we bring the submit button to the same line as the Textbox

> I think the docstring for ChatInterface says that we can include a Chatbot component as a parameter, or one would be included by default. I was trying to understand in which cases would we need to explicitly mention the chatbot param?

If you want to customize the Chatbot in any way (e.g. the height or the latex_delimiters). I'll explain this in the docstring

> I also found it difficult to understand how to pass parameters like temperature, max_tokens, top_p, etc., while calling my inference (predict) function.

Yeah this isn't currently possible, and I agree its a pretty big limitation. I'm thinking we can have an additional `inputs` parameter that takes inputs beyond the textbox and passes them into your functions. My main question is where should these questions show up in the UI?

> One strange behavior I noticed was that on Colab, the initial value of chatbot comes in as a blank list [], while with the exact same code on Spaces, the initial value captured is chatbot = [[first_user_input, None]]  I had to handle this when I moved my code from Colabs to a Space.

That seems weird, can you share a minimal code example?

> Till now, we used to add the bot response to the chatbot list (a list of sub-lists of user and corresponding bot messages) and use to return that to the output component (chatbot). However, now we just need to return the bot_response, and the interface takes care of everything else. I really liked this part as it reduced the number of lines of code on my end, reducing the chances of error.

🔥

gradio/chat_interface.py
abidlabs: Oh wait I made a mistake in `_submit_fn`. The history should absolutely NOT include the last message. I misunderstood what you were saying earlier. Sorry, I'll fix and test this out

gradio/chat_interface.py
abidlabs: Hmm it actually complicates things a bit because we have to separately handle the case if users supply a button (in which case we still have to add the event handlers manually). I'll just leave as is

abidlabs:
@freddyaboulton the issues you pointed out should be fixed. Here's example code:

```py
from gradio import ChatInterface
import random

def test(message, history):
    return f"{len(history)+1} messages so far. Here's a random number: {random.random()}"

ChatInterface(test).launch()
```

Tested with the Client as well:

```py

client = Client(url)

client.reset_session()
client.predict("test")
```


Testing streaming right now

abidlabs:
Streaming is working too. Test code:

```py
from gradio import ChatInterface
import random
import time

def test(message, history):
    message = f"{len(history)+1} messages so far. Here's a random number: {random.random()}"
    for i in range(len(message)):
        time.sleep(0.01)
        yield message[:i]

_, url, _ = ChatInterface(test).queue().launch()


# from gradio_client import Client

# client = Client(url)

# for x in client.submit("test"):
#     print(x)
```

abidlabs:
@yvrjsharma your first two points have been addressed as well.

Working on tests now

freddyaboulton:
> My main question is where should these questions show up in the UI?

I like the parameters accordion in falcon chat demo: https://huggingface.co/spaces/HuggingFaceH4/falcon-chat

Something we can think about for the future - how to make our high-level abstractions more modular? If you want to use ChatInterface for convenience but your use case isn't 100% covered you have to use gr.Blocks() from scratch.  Maybe we can let the function be None and let users configure it themselves if they know what they're doing. This example introduces an `add_fn` method (not advocating for that api just pointing out what I think would be cool to do)

```python
chat = gr.ChatInterface(fn=None, delete_last_btn="del").queue()

def predict_with_params(msg, history, temperature):
    # do cool stuff
    pass

def handle_upload(file, msg, history):
    # As a question about the uploaded file
    pass

with gr.Blocks() as demo:
    chat.render()
    with gr.Row():
        with gr.Column():
            with gr.Accordion(label="Parameters"):
                temp = gr.Slider()
        with gr.Column():
            upload = gr.UploadButton()
    chat.add_fn(predict_with_params, inputs=[temp])
    chat.add_fn(handle_upload, inputs=[upload])
```

Something to think about for later.

yvrjsharma:
> My main question is where should these questions show up in the UI?
I agree with @freddyaboulton's suggestion on this. An accordion below the buttons would make a good resting place.

abidlabs:
@yvrjsharma can't reproduce this locally (with and without queuing), testing on Spaces rn

yvrjsharma:
Thanks for checking this, @abidlabs. I now understand this was because of the different wheels between us.

abidlabs:
After merging in the the changes from the `fix_group` branch, the UI looks like this:

image

freddyaboulton (commented):
This is looking really good! Noticed a typo in the guide lol

abidlabs:
Thanks @freddyaboulton! Still a fair bit of to do work to do on the docs and the functional test, but feel free to review everything else

abidlabs:
Ok all right, tests are in place. I'm going to make one more pass through the docs and then this should be ready

abidlabs:
> I also found it difficult to understand how to pass parameters like temperature, max_tokens, top_p, etc., while calling my inference (predict) function.

I have a simple idea to address this. We could add an optional `additional_inputs` parameter to `gr.ChatInterface()` which accepts a list of components. The values of these components would be passed into your chat function after `message` and `history`. You could create the components anywhere within a larger `gr.Blocks()` context

Example usage:

```py
def chat(message, history, system_prompt):
   ...

with gr.Blocks() as demo:
     prompt = gr.Textbox()
     gr.ChatInterface(chat, additional_inputs=prompt)
demo.launch()
```

You could still do the accordion, but you'd just create it yourself (which I think is better than forcing this particular UI design)

taoari:
@abidlabs This is nice work, it makes Chatbot much easier to use!

I would suggest adding an upload button on the left of the input box, and with some callback like:

```py
def _upload_file(msg, filepath):
    # history = history + [((file.name,), None)]
    mtype = mimetypes.guess_type(filepath.name)[0]
    if mtype.startswith('image'):
        msg += f'<img src="\\file={filepath.name}" alt="{os.path.basename(filepath.name)}"/>'
    elif mtype.startswith('audio'):
        msg += f'<audio controls><source src="\\file={filepath.name}">{os.path.basename(filepath.name)}</audio>'
    elif mtype.startswith('video'):
        msg += f'<video controls><source src="\\file={filepath.name}">{os.path.basename(filepath.name)}</video>'
    else:
        msg += f'<a href="\\file={filepath.name}">📁 {os.path.basename(filepath.name)}</a>'
    return msg

upload = gr.UploadButton("📁", file_types=["image", "video", "audio", "file"])
upload.upload(_upload_file, [msg, upload], [msg], queue=False)
```

for basic multimodal input support.

taoari:
@abidlabs This is nice work, it makes Chatbot much easier to use!

I would suggest adding an upload button on the left of the input box, and with some callback like:

```py
def _upload_file(msg, filepath):
    # history = history + [((file.name,), None)]
    mtype = mimetypes.guess_type(filepath.name)[0]
    if mtype.startswith('image'):
        msg += f'<img src="\\file={filepath.name}" alt="{os.path.basename(filepath.name)}"/>'
    elif mtype.startswith('audio'):
        msg += f'<audio controls><source src="\\file={filepath.name}">{os.path.basename(filepath.name)}</audio>'
    elif mtype.startswith('video'):
        msg += f'<video controls><source src="\\file={filepath.name}">{os.path.basename(filepath.name)}</video>'
    else:
        msg += f'<a href="\\file={filepath.name}">📁 {os.path.basename(filepath.name)}</a>'
    return msg

upload = gr.UploadButton("📁", file_types=["image", "video", "audio", "file"])
upload.upload(_upload_file, [msg, upload], [msg], queue=False)
```

for basic multimodal input support.

taoari:
Another thing is that I found the current implementation is not streaming friendly. For example, you have to use the following code to stream the chat message:

```py
  def echo(message, history):
      for i, c in enumerate(message):
          yield message[:i+1]
```

It would be better if the following callback function would work:

```py
def echo(message, history):
    for c in message:
        yield c
```

Finally, I would suggest "Delete Last" could be changed to "Undo". When undo, the last message should also come back to the textbox. In this case, "Retry" may be removed, as it is simply an "Undo"+"Submit". "Clear History" may also be renamed to "Clear" for simplicity. The UI may be like the following to save space:

image

guides/04_chatbots/01_creating-a-chatbot-fast.md
```diff
@@ -0,0 +1,254 @@
...
+            yield partial_message 
+
+gr.ChatInterface(predict).queue().launch() 
+```
+
+## Examples using open-source LLMs with Hugging Face!
+
+Of course, why use a closed-source model when you can use an open-source one instead? Here's the equivalent example using Together's RedePajama model.
+
+```py
```
freddyaboulton, line 202: Should we deploy as a space so that we can embed it in the docs and have users try it out? WDYT @yvrjsharma ?

gradio/chat_interface.py
```diff
@@ -0,0 +1,326 @@
...
+                    f"<h1 style='text-align: center; margin-bottom: 1rem'>{self.title}</h1>"
+                )
+            if description:
+                Markdown(description)
+
+            with Group():
+                if chatbot:
+                    self.chatbot = chatbot.render()
+                else:
+                    self.chatbot = Chatbot(label="Input")
```
freddyaboulton, line 117: Shouldn't the default label for the chatbot be "chat" or "chatbot" as opposed to "Input"?

abidlabs:
Thanks for the suggestions @taoari! Here are my thoughts:

> I would suggest adding an upload button on the left of the input box, and with some callback like:

We'll add multimodal support in a future PR (probably something like what you proposed)

> Finally, I would suggest "Delete Last" could be changed to "Undo". When undo, the last message should also come back to the textbox. In this case, "Retry" may be removed, as it is simply an "Undo"+"Submit". "Clear History" may also be renamed to "Clear" for simplicity. The UI may be like the following to save space and fit into single row:

I like "Undo"! Will update the UI. Not sure if putting all of the buttons in the same line will look good, but will play around with this

> Another thing is that I found the current implementation is not streaming friendly. For example, you have to use the following code to stream the chat message:

There are two reasons for this approach:

1. Its consistent with how generator functions work in `gr.Interface()` and `gr.Blocks()`
2. Its more general than just sending the individual chunks back because e.g. it allows the chatbot streaming function to "edit" previous responses. 

I do see the point in having a simpler interface like you've suggested, perhaps that could be toggled via a parameter at some point, but not the default.

aliabd:
@abidlabs i made some fixes to the website stuff, please take a look at the preview and lmk what you think.

abidlabs:
Thanks @aliabd perfect!

abidlabs:
I've addressed the feedback from @freddyaboulton and switched from "delete previous" to "undo" and updated the tests accordingly. Here's how the UI currently looks:

image


Will discuss internally on whether to move the buttons to a single line or not.

taoari:
In streaming mode, one has to add `yield ""` in front of the bot function, otherwise, an error will occur if the user presses ENTER with empty input. Anyway to fix this?

```py
from gradio import ChatInterface
import time

def echo_stream(message, history):
    yield ""
    for i in range(len(message)):
        time.sleep(0.3)
        yield message[:i+1]

ChatInterface(echo_stream).queue().launch()
```

abidlabs:
Thanks @taoari for catching that. I've handled this edge case so that if a generator does not `yield` anything, it is automatically assumed to `yield None` and the chatbot does not error out (in the UI, no bot response apperas). That's analogous to a regular function not having a return statement which is assumed to `return None` in Python.

PR #4870, opened 2023-07-10
Quick fixes to video test

abidlabs:
When running the tests locally, I noticed that a file would be created as a test artifact: `test/test_files/playable_but_bad_container.mp4`. 

Upon a closer look at the tests, realized they weren't actually testing anything meaningful, so I refactored a bit and consolidated the tests together


test/test_components.py
```diff
@@ -1441,10 +1441,8 @@ def test_video_postprocess_converts_to_playable_format(self):
             bad_vid = str(test_file_dir / "bad_video_sample.mp4")
             assert not processing_utils.video_is_playable(bad_vid)
             shutil.copy(bad_vid, tmp_not_playable_vid.name)
-            _ = gr.Video().postprocess(tmp_not_playable_vid.name)
-            # The original video gets converted to .mp4 format
-            full_path_to_output = Path(tmp_not_playable_vid.name).with_suffix(".mp4")
-            assert processing_utils.video_is_playable(str(full_path_to_output))
```
abidlabs, line 1447: This test wasn't actually testing the postprocess function at all! We were ignoring the postprocessed file and manually replacing the extension with ".mp4"

test/test_components.py
```diff
@@ -1453,18 +1451,8 @@ def test_video_postprocess_converts_to_playable_format(self):
             bad_vid = str(test_file_dir / "playable_but_bad_container.mkv")
             assert not processing_utils.video_is_playable(bad_vid)
             shutil.copy(bad_vid, tmp_not_playable_vid.name)
-            _ = gr.Video().postprocess(tmp_not_playable_vid.name)
-            full_path_to_output = Path(tmp_not_playable_vid.name).with_suffix(".mp4")
-            assert processing_utils.video_is_playable(str(full_path_to_output))
```
abidlabs, line 1458: Same as above

test/test_components.py
```diff
@@ -1453,18 +1451,8 @@ def test_video_postprocess_converts_to_playable_format(self):
...
-
-    def test_convert_video_to_playable_format(self, monkeypatch, tmp_path):
-        test_file_dir = Path(Path(__file__).parent, "test_files")
-        monkeypatch.setenv("GRADIO_TEMP_DIR", str(tmp_path))
-        video = gr.Video(format="mp4")
-        output = video.postprocess(
-            str(test_file_dir / "playable_but_bad_container.mkv")
-        )
-        assert Path(output[0]["name"]).suffix == ".mp4"
```
abidlabs, line 1467: When the tests above is fixed, this test is redundant

abidlabs:
Looks like the flaky test is back @freddyaboulton 😅

freddyaboulton:
Darn. Will take a look at this PR when I’m back home. Diff is hard to read on mobile. In the meantime, I think it’s ok to comment out the flaky Warning modal text while we figure out what’s happening.

test/test_components.py
```diff
@@ -1441,10 +1441,8 @@ def test_video_postprocess_converts_to_playable_format(self):
             bad_vid = str(test_file_dir / "bad_video_sample.mp4")
             assert not processing_utils.video_is_playable(bad_vid)
             shutil.copy(bad_vid, tmp_not_playable_vid.name)
-            _ = gr.Video().postprocess(tmp_not_playable_vid.name)
-            # The original video gets converted to .mp4 format
-            full_path_to_output = Path(tmp_not_playable_vid.name).with_suffix(".mp4")
-            assert processing_utils.video_is_playable(str(full_path_to_output))
+            output = gr.Video().postprocess(tmp_not_playable_vid.name)
```
freddyaboulton, line 1444: Can you add `monkeypatch.setenv("GRADIO_TEMP_DIR", str(tmp_path))` here (from the deleted test)? That way the local temp dir won't get filled up with lots of temp files. Will figure out a way to do that automatically for all tests going forward in a separate PR

freddyaboulton (approved):
Thanks @abidlabs !

test/test_components.py
abidlabs: Thanks, sounds good! I've done it now, but yeah ideally we set that env variable for all of the pytests at once

freddyaboulton:
Thanks @abidlabs !

Issue #4836, opened 2023-07-07
Allow accessing endpoints that include some non-serializable components via the Client


abidlabs:
### Describe the bug

If an endpoint has any input or output components that are non-serializable, then the Client will not see it as a valid endpoint. This should be [relaxed as discussed internally](https://huggingface.slack.com/archives/C02QZLG8GMN/p1688752887573919?thread_ts=1688550619.459869&cid=C02QZLG8GMN).

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```py
import gradio as gr
from gradio_client import Client

client_base = Client("fffiloni/zeroscope")
result = client_base.view_api()
```

produces no valid endpoint

### System Info

```shell
gradio==3.36
```

### Severity

I can work around it

PR #4871, opened 2023-07-11
Client: Support endpoints that return layout components

freddyaboulton:
## Description

Closes: #4836

If an endpoint returns a layout component, it will not be returned by the api but all other IOComponents will (excluding State). These components are excluded from the view_api page as well. But the endpoint is still considered valid for the client.

## 🎯 PRs Should Target Issues

## Tests & Changelog


freddyaboulton:
UI and gradio unit tests should pass when new version of the client is released

abidlabs:
Instead of making "row", etc. all serializable components, wouldn't a simpler approach be to modify `skip_components` here:

https://github.com/gradio-app/gradio/blob/5dc445b08fd76cacec1354b05a2d44f23d127a30/gradio/blocks.py#L505

to include "row", etc. Then, we could just remove this logic altogether:

https://github.com/gradio-app/gradio/blob/5dc445b08fd76cacec1354b05a2d44f23d127a30/gradio/blocks.py#L520

I think that's more appropriate since the layouts are not actually serializable and we shouldn't treat them as such.

client/python/gradio_client/client.py
```diff
@@ -427,7 +427,7 @@ def view_api(
 
         # Versions of Gradio older than 3.29.0 returned format of the API info
         # from the /info endpoint
-        if version.parse(self.config.get("version", "2.0")) > version.Version("3.29.0"):
+        if version.parse(self.config.get("version", "2.0")) > version.Version("3.36.1"):
```
akx, line 430: Is this `>` on purpose (c.f. `>=`)?

client/python/gradio_client/utils.py
```diff
@@ -31,10 +31,22 @@
 CONFIG_URL = "config"
 API_INFO_URL = "info"
 RAW_API_INFO_URL = "info?serialize=False"
-SPACE_FETCHER_URL = "https://gradio-space-api-fetcher-v2.hf.space/api"
+SPACE_FETCHER_URL = "https://freddyaboulton-space-api-fetcher.hf.space/api"
 RESET_URL = "reset"
 SPACE_URL = "https://hf.space/{}"
 
+SKIP_COMPONENTS = [
```
akx, line 38: This should probably be a set for faster lookup.

freddyaboulton:
@abidlabs The skip_components in `blocks.py` are already updated to include state and all the layout components. The reason I included them in the serializer map is to preserve the principle that each output component has a serializer. That way we treat the layout components the same way we do state. I think that it's conceptually sound (they are both returned by the server but we don't want to display to users) and it makes the footprint of this PR pretty small. Although the layout components are in the serializer map, they will not be displayed in the api info or returned to users and the serializing logic is a no-op.

abidlabs:
> The skip_components in blocks.py are already updated to include state and all the layout components

Nice, I missed this. 

But if we are already skipping the layout components, why do we need to add serializers for them? Imo that doesn't make sense because the serializers should only exist for components that can hold values since they define how this value is serialized between different formats. `gr.State` can hold a value, but the layout components don't hold any values. 

Happy to sync if that's easier

freddyaboulton:
> We should instead "skip" the serialization of these components. (Skipping could just mean the identity function)

In `COMPONENT_MAPPING`, I map the layout components to the `SimpleSerializable`. That's the same as the identify function no?

I guess I I could remove the layout components from the `COMPONENT_MAPPING` and instead modify the following code to not trigger the `assert` for the layout components and manually add a `SimpleSerializer` for the layout components. That's equivalent to what the current code does though and feels harder to maintain.

```python
                        assert (
                            component_name in serializing.COMPONENT_MAPPING
                        ), f"Unknown component: {component_name}, you may need to update your gradio_client version."
                        serializer = serializing.COMPONENT_MAPPING[component_name]
                    serializers.append(serializer())  # type: ignore
```

abidlabs:
I think we should replace:

```py
                    else:
                        assert (
                            component_name in serializing.COMPONENT_MAPPING
                        ), f"Unknown component: {component_name}, you may need to update your gradio_client version."
                        deserializer = serializing.COMPONENT_MAPPING[component_name]

```

with
```py
                    elif component_name in utils.SKIP_COMPONENTS:
                         deserializer = serializing.SimpleSerializable
                    else:
                        assert (
                            component_name in serializing.COMPONENT_MAPPING
                        ), f"Unknown component: {component_name}, you may need to update your gradio_client version."
                        deserializer = serializing.COMPONENT_MAPPING[component_name]
```

And then just remove these from serializing.COMPONENT_MAPPING:

```py
    "row": SimpleSerializable,
    "column": SimpleSerializable,
    "tabs": SimpleSerializable,
    "tab": SimpleSerializable,
    "box": SimpleSerializable,
    "form": SimpleSerializable,
    "accordion": SimpleSerializable,
    "group": SimpleSerializable,
```

That reduces the footprint of the PR and should make it even easier to maintain since we only need to add components in one place: SKIP_COMPONENTS.

freddyaboulton:
Followed your suggestion @abidlabs !

client/python/gradio_client/utils.py
```diff
@@ -31,10 +31,22 @@
 CONFIG_URL = "config"
 API_INFO_URL = "info"
 RAW_API_INFO_URL = "info?serialize=False"
-SPACE_FETCHER_URL = "https://gradio-space-api-fetcher-v2.hf.space/api"
+SPACE_FETCHER_URL = "https://freddyaboulton-space-api-fetcher.hf.space/api"
```
abidlabs, line 34: We just need to remember to update this

client/python/test/test_serializing.py
```diff
@@ -22,7 +22,9 @@ def test_duplicate(serializer_class):
 def test_check_component_fallback_serializers():
     for component_name, class_type in COMPONENT_MAPPING.items():
         # skip components that cannot be instantiated without parameters
-        if component_name in ["dataset", "interpretation"]:
+        if component_name in ["dataset", "interpretation"] + list(
+            SKIP_COMPONENTS - {"state"}
```
abidlabs, line 26: Should we just add `["dataset", "interpretation"]` to SKIP_COMPONENTS and remove them from COMPONENT_MAPPING?

abidlabs:
Small nit above. Code looks good + tested. LGTM @freddyaboulton

abidlabs (approved):

freddyaboulton:
Thanks for the review @abidlabs !!

PR #4882, opened 2023-07-11
Use tempdir for all python unit tests tests

freddyaboulton:
## Description

As discussed in #4870 

Also automatically setting GRADIO_ENABLED_ANALYTICS to false for all test files (client and library)

## 🎯 PRs Should Target Issues

## Tests & Changelog

### DIFF
--- a/client/python/test/conftest.py
+++ b/client/python/test/conftest.py
@@ -228,3 +228,15 @@ def all_components():
             subclasses.append(subclass)
 
     return subclasses
+
+
+@pytest.fixture(autouse=True)
+def disable_telemetry(monkeypatch):
+    """Automatically disable analytics."""
+    monkeypatch.setenv("HF_HUB_DISABLE_TELEMETRY", "1")
+
+
+@pytest.fixture(autouse=True)
+def disable_analytics(monkeypatch):
+    """Automatically disable analytics."""
+    monkeypatch.setenv("GRADIO_ANALYTICS_ENABLED", "False")
--- a/test/conftest.py
+++ b/test/conftest.py
@@ -55,3 +55,18 @@ def connect():
             demo.server.thread.join(timeout=1)
 
     return _connect
+
+
+@pytest.fixture(autouse=True)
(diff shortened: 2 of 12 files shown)


client/python/test/conftest.py
```diff
@@ -228,3 +228,15 @@ def all_components():
             subclasses.append(subclass)
 
     return subclasses
+
+
+@pytest.fixture(autouse=True)
+def disable_telemetry(monkeypatch):
```
abidlabs, line 234: Nit: Instead of an autouse fixture, which means that this would be run before every single test, what about setting up these environment variables in `pytest_configure` so that they run a single time before all of the tests?

freddyaboulton: Agreed - done!

abidlabs: Thanks @freddyaboulton! So now we don't need these two fixtures, right?

test/conftest.py
```diff
@@ -55,3 +55,12 @@ def _connect(demo: gr.Blocks, serialize=True):
...
     return _connect
+
+
+@pytest.fixture(autouse=True)
+def gradio_temp_dir(monkeypatch, tmp_path):
+    """tmp_path is unique to each test function.
+    It will be cleared automatically according to pytest docs: https://docs.pytest.org/en/6.2.x/reference.html#tmp-path
+    """
+    monkeypatch.setenv("GRADIO_TEMP_DIR", str(tmp_path))
+    yield tmp_path
```
abidlabs, line 66: Nit, but I think `return` would be simpler here: 
```suggestion
    return tmp_path
```

abidlabs, line 66: Nit: do we need `yield` here? The monkeypatch should automatically revert the environmental variable after the test is complete
```suggestion
    return tmp_path
```

abidlabs:
Just a few nits, otherwise LGTM @freddyaboulton

freddyaboulton:
Thanks @abidlabs ! Took care of the last nits. Will merge to clear the PR queue.

Issue #4523, opened 2023-06-15
Render Markdown and LaTeX in `gr.Markdown` in the frontend, like `gr.Chatbot`


abidlabs:
Recently, we've made some changes to `gr.Chatbot()` to render markdown and LaTeX in the frontend instead of the backend. We should do the same in the `gr.Markdown()` component. 

This will allow us to get rid of:

```
markdown-it-py[linkify]>=2.0.0
mdit-py-plugins<=0.3.3
```

from our requirements

abidlabs:
See also: https://github.com/gradio-app/gradio/issues/4563

Jack7510:
when will this issue be fixed?

PR #4885, opened 2023-07-12
Pinning dependencies in requirements.txt

abidlabs:
For `gradio`, here is what I did with each external dependency:
* [aiofiles](https://pypi.org/project/aiofiles/#history): the first post-zero release was 22.x. I confirmed gradio's `/file` route works with audio files for this version of aiofiles so I allow 22.x, 23.x
* [aiohttp](https://pypi.org/project/aiohttp/#history): the current major (3.x) is >5 years old, so I've just pinned to 3.x
* [altair](https://pypi.org/project/altair/#history): just released 5.0 in May. Gradio works with both 4.x and 5.x so I've pinned to 4.x or 5.x
* [fastapi](https://pypi.org/project/fastapi/#history): still a pre-release so I haven't pinned
* [ffmpy](https://pypi.org/project/ffmpy/#history): still a pre-release so I haven't pinned
* [httpx](https://pypi.org/project/httpx/#history): still a pre-release so I haven't pinned
* [Jinja2](https://pypi.org/project/Jinja2/#history): we've had issues before between jinja2 and starlette compatibilties so I didn't want to pin too aggressively. The current version is 3.1.2 (released more than a year ago), I just pinned <4.0. This shouldn't have much of an effect
* [markdown-it-py](https://pypi.org/project/markdown-it-py/#history) and  [mdit-py-plugins](https://pypi.org/project/mdit-py-plugins/#history) -- didn't pin these because I'm hoping we can remove these dependencies altogether as a result of #4523 
* [pygments](https://pypi.org/project/pygments/#history) -- unused dependency! removed
* [matplotlib](https://pypi.org/project/matplotlib/#history) -- the current major (3.x) is >15 years old, so I've just pinned to 3.x
* [numpy](https://pypi.org/project/numpy/#history) -- the current major (1.x) is >15 years old, so I've just pinned to 1.x
* [orjson](https://pypi.org/project/orjson/#history) -- the current major (3.x) is >3 years old, so I've just pinned to 3.x
* [pandas](https://pypi.org/project/pandas/#history) -- I confirmed gradio works with 1.x, 2.x so I pin those
* [PIL](https://pypi.org/project/pillow/#history) -- I confirmed gradio works from 8.0 (>3 years old) onwards so I pin anything between 8.x to 10.x. PIL is actively released so this is one that we'd want to keep track of and expand the pinned range if they release 11.0
* [pydantic](https://pypi.org/project/pydantic/#history) -- the source of our troubles. @freddyaboulton has confirmed that pydnatic 1.x and 2.x are supported in the latest version of gradio so I've pinned those
* [python-multipart](https://pypi.org/project/python-multipart/#history): still a pre-release so I haven't pinned
* [pydub](https://pypi.org/project/pydub/#history): still a pre-release so I haven't pinned
* [pyyamml](https://pypi.org/project/pyyamml/#history) -- I confirmed `gr.load()` works with tabular examples with 5.x, 6.x so I pin those
* [requests](https://pypi.org/project/requests/#history) -- the current major (2.x) is >9 years old, so I've just pinned to 2.x
* [semantic_version](https://pypi.org/project/semantic_version/#history) -- the current major (2.x) is >10 years old, so I've just pinned to 2.x
* [uvicorn](https://pypi.org/project/uvicorn/#history): still a pre-release so I haven't pinned (beyond the existing pin)
* [websockets](https://pypi.org/project/websockets/#history): already pinned to >=10.0. Added a pin <12.0

For `gradio_client`:
* requests -- made same change as for `gradio`
* websockets -- made same change as for `gradio`
* [packaging](https://pypi.org/project/packaging/#history)  - I didn't pin this library because we use it in a very basic way and they release new majors every year
* [fsspec](https://pypi.org/project/fsspec/#history)  - pretty much same situation as `packaging`
* [typing_extensions](https://pypi.org/project/typing_extensions/#history) -- pined to 4.x
* [httpx](https://pypi.org/project/httpx/#history): still a pre-release so I haven't pinned

Separately, I also added:
* typing_extensions 
* packaging

as requirements to `gradio` since we use them explicitly in the `gradio` source code. Safer to include them rather than depend on them being available from `gradio_client` since the client source code may change.


requirements.txt
```diff
@@ -1,26 +1,26 @@
...
 mdit-py-plugins<=0.3.3
-markupsafe
-matplotlib
-numpy
-orjson
-pandas
-pillow
-pydantic
-python-multipart
+markupsafe~=2.0
```
freddyaboulton, line 12: Version 2 came out two years ago so this makes sense

requirements.txt
```diff
@@ -1,26 +1,27 @@
...
+matplotlib~=3.0
+numpy~=1.0
+orjson~=3.0
+packaging
+pandas>=1.0,<3.0
+pillow>=8.0,<11.0
+pydantic>=1.0,<3.0
+python-multipart  # required for fastapi forms
```
freddyaboulton, line 20: Was gonna say 😄

requirements.txt
```diff
@@ -1,26 +1,27 @@
...
-pydantic
-python-multipart
+markupsafe~=2.0
+matplotlib~=3.0
+numpy~=1.0
+orjson~=3.0
+packaging
+pandas>=1.0,<3.0
+pillow>=8.0,<11.0
+pydantic>=1.0,<3.0
```
freddyaboulton, line 19: Nit: Let's use the same [pin](https://github.com/tiangolo/fastapi/blob/master/pyproject.toml#L45) as fastapi. Version 1.0 is actually not allowed.

requirements.txt
```diff
@@ -1,26 +1,27 @@
...
+altair>=4.2.0,<6.0
 fastapi
 ffmpy
 gradio_client>=0.2.7
 httpx
 huggingface_hub>=0.14.0
-Jinja2
+Jinja2<4.0
 markdown-it-py[linkify]>=2.0.0
-pygments>=2.12.0
```
freddyaboulton, line 11: Makes sense. Should have been deleted as part of https://github.com/gradio-app/gradio/pull/4150

freddyaboulton (approved):
Thanks for going through this abidlabs. Since we're only pinning to majors that were released at least a couple of years ago I think this will not cause unintended upgrade/downgrades of dependencies in developer's environments. Would be good to communicate to developers that we're making this change though?

requirements.txt
abidlabs: Interesting. Do we have to do this? I don't think we need to use the same versions of fastapi (that's a bit redundant anyways since fastapi is one our dependencies). We just need to make sure that our library's use of `pydantic` is valid according to the version of `pydantic` installed on user's machines. 

That being said, I haven't tested different versions of `pydantic` so maybe we actually do need to be more strict

abidlabs:
> Would be good to communicate to developers that we're making this change though?

Yeah good point, I'll add a note in the breaking changes, and let's post on Twitter + Discord when we do the next release.

akx:
Was there an issue or something tracking why this is needed for so many packages? Overly tight pins _will_ cause issues in downstream projects utilizing Gradio – and furthermore, not all Python packages do Semver at all, so pinning to a major version may not mean what you expect.

abidlabs:
Updated pins and improved changelog message @freddyaboulton -- I think this PR should be good to go

freddyaboulton (approved):
Thanks for the nice PR @abidlabs !!

Issue #3441, opened 2023-03-11
When "allow_flagging" = auto, allow saving output as well


lakshman111:
- [yes] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
When "allow_flagging" = "auto", every input the user submits is automatically flagged (https://www.gradio.app/docs/#dataset).  Outputs are not flagged.  However, I need outputs to be flagged as well, similar to how inputs and outputs are flagged when "allow_flagging" = "manual".

**Describe the solution you'd like**  
When "allow_flagging" = "auto", allow outputs to be flagged as well so that they can also be auto saved along with the inputs to a Hugging Face dataset (similar to the behavior when "allow_flagging" = "manual").

**Additional context**  
My usecase is that I'd like to be able to see all input and output images, so that I can make sure the model is outputting quality results.

Reekomer:
Is there anyone working on this issue? I also would like to be able to save inputs/outputs when flagging auto.

lakshman111:
@Reekomer I ended up just adding an API call as a part of my function to upload.io to store both the input and the output.  May switch that storage to Supabase though!

abidlabs:
We should fix this, this should now be possible by re-writing the flagging function to first wait until the prediction is complete and then use `.then()` to do the flagging. Will mark this open for contributors!

davidberenstein1957:
I will work on this.

davidberenstein1957:
@abidlabs I believe this should be an easy fix. I also missed the docs w.r.t. creating custom components so I added this BaseFlaggingCallback to the docs too.

aperikleous:
@abidlabs Is there still a plan to add this feature? My use case requires constant flagging of both input and output, so this functionality would be really helpful.

abidlabs:
Yes its on the roadmap to get this fixed, but we haven't had a chance yet. It may take us some time -- if anyone has bandwidth to fix this, we are happy to review a PR

PR #4889, opened 2023-07-12
include output in auto mode

zhlim-ml:
Even if it's auto flagging mode, we should still include output component.

## Description
Essentially the auto mode should only determine whether the flagging mode is auto or not. 
If there's a need to remove output component, should introduce a new config for that.
https://www.gradio.app/guides/using-flagging
the documentation never states that as well 
`auto: users will not see a button to flag, but every sample will be flagged automatically.` 

Closes: #3441

## Tests & Changelog


abidlabs:
Thanks @zhlim-ml but this doesn't fix the issue at hand actually because when the flagging action happens, the output component is not populated so its an empty value that gets flagged. I'll fix this issue another way

Issue #3360, opened 2023-03-02
clean up component folder structure + move components out of the `app` package


pngwn:
Get the components out of the `app`. This is the really disruptive part, so I want to keep it simple conceptually.

- No more weird wrappers, everything will be in the core component in its entirety. 
    - There was a reason for doing things the way we are currently but it was wrong decision. I thought we should separate the app logic from the component logic but they are inherently linked anyway and all it did was serve some possible future usecase without really addressing our needs.
- create a 1-2-1 relationship between gradio component s and component packages.
  - The current situation is confusing when you are working across the python and javascript
- create a new folder structure inside each package that will contain different component variants
    - `/interactive/` this will contain the interactive variant of a component - **optional** - static used when not present
    - `/static/` this will contain the static variant of a component - **mandatory**
    - `/example/` this will contain the preview variant of a component - **optional** - fallback used when not present
    - We will also need a `@gradio/fallback` package which we defer to in certain cases.
    - This paves the way for custom components but also brings some order to what is currently quite confusing. Every sub folder will have an `index.ts` that exports the correct component. The components can be called anything. We won't explicitly use these subfolder for now (we will have an `index.svelte` in the root) but will be cleaned up in a future refactor. `/preview/` will be empty for a short period of time.
- Certain things will just be passed down as props for now (like the i18n helper), this may change in the future but don't want to overthink.

This will get closed when all of this work is done. tracking issue for the first part:

- [x] #5006

abidlabs:
Nice @pngwn! 

I wonder if we should make `static` the required one and `interactive` the optional one that uses `static` as a fallback. As far as I understand it, some components can be used as outputs but not inputs (e.g. `JSON` or `Markdown` or `Label`) -- so they would support a static mode but not an interactive one?

pngwn:
Ah that's a good point, yeah maybe that makes more sense!

abidlabs:
Oh and this might be a minor point, but this is probably best time to bring it up -- can we rename the component modes "input", "output", and "example" instead of "interactive", "static", and "preview"? Naming doesn't matter if we're using these internally but since we're planning on letting users contribute custom components, this naming would be more intuitive (since they're already familiar with inputs/outputs) and be one less thing to remember

pngwn:
Well, the only reason that might not make sense if you can use interactive as output and static as input if you want to. Since blocks, those distinctions aren't as clear any more. 

[...]

PR #4900, opened 2023-07-12
move all of the chatbot component in to js/chatbot

pngwn:
## Description

This is the file move for the chatbot component, to get out of @dawoodkhan82's way. Apologies for the delay!

## 🎯 PRs Should Target Issues

## Tests & Changelog


abidlabs:
Thanks @pngwn! Just a couple of questions:

1. Do we need this file: `js/app/src/components/Chatbot/index.ts`? 

* We no longer use this information in the API docs since we get it from the `/info` route:

```js
	type: {
		payload: "Array<[string, string]>"
	},
	description: {
		payload: "list of message pairs of"
	},
	example_data: config.value?.length
		? config.value
		: [
				["Hi", "Hello"],
				["1 + 1", "2"]
		  ]
});
```

So the file doesn't seem to have much purpose. Thinking ahead to community components, it would be really nice if all of the files related to a component were entirely in a single directory (e.g.: `js/chatbot/`)

pngwn:
> 1. Do we need this file: `js/app/src/components/Chatbot/index.ts`?
> 
> * We no longer use this information in the API docs since we get it from the `/info` route:
> 
> ```js
> 	type: {
> 		payload: "Array<[string, string]>"
> 	},
> 	description: {
> 		payload: "list of message pairs of"
> 	},
> 	example_data: config.value?.length
> 		? config.value
> 		: [
> 				["Hi", "Hello"],
> 				["1 + 1", "2"]
> 		  ]
> });
> ```
> 
> So the file doesn't seem to have much purpose. Thinking ahead to community components, it would be really nice if all of the files related to a component were entirely in a single directory (e.g.: `js/chatbot/`)

I didn't realise that, in that case we can remove all of these snippets and I'll make sure to do that when I make a big PR for everything else (hopefully tomorrow/fri).

Regarding the removal of the file, that is the ultimate goal yes but that will be 'stage 2' listed [here](https://github.com/gradio-app/gradio/issues/3360#issuecomment-1609753735). Eventually we will just load directly from `/interactive/` or `/static/` but things will break if i do that now and I don't want to introduce  a special case for one components. So the idea is just move the files with as few changes to the core app as possible and make all of those changes at once to `js/app` when the components have been moved.

This PR specifically was mainly to just get the movage out of the way so that I don't end up with conflicts as @dawoodkhan82 starts to work on the chatbot. There will probably be one or two minor changes but they shouldn't touch the core chatbot code which is the main issue.

> 2. The file above includes this line:
> 
> ```
> export const modes = ["static"];
> ```
> 
> But we have the files for the `Chatbot` source an the `interactive/` folder. Is there a reason for this?

So I'm a bit confused about chatbot really, is it static or interactive? It _feels_ like an interactive component but i guess all it actually does is renders content we pass to it? I guess it is less meaningful when there is only one version of a component (unlike say `Image` which has two distinct variants). Maybe I should make this a `static` folder rather than interactive, since that is the default we decided on. Or maybe it is both but the `static` version disables things like upvoting comments etc?

abidlabs:
> I didn't realise that, in that case we can remove all of these snippets and I'll make sure to do that when I make a big PR for everything else (hopefully tomorrow/fri).

I believe that's the case, @aliabd can you confirm that these snippets are no longer used? 

Everything else looks good to me @pngwn!

PR #4905, opened 2023-07-12
Speed up gradio.launch, especially when there's no internet

aliabid94:
There are a lot of blocking network calls that are made before the local URL gets printed - these slow down the display of the local URL. Without internet this is especially bad, it takes 60 seconds between running hello_world/run.py and getting the local URL. This is because the is_sagemaker_check has a 60 second timeout that cannot be configured.

I moved the is_sagemaker_check to after printing the local URL, and also made the analytics network calls happen in a separate thread so they are non blocking.

Time from running `python hello_world/run.py` and printing the local URL:
With internet:

- Before: 1.36s
- After: 1.22s

Without internet:
- Before: 60.5s
- After: 1.29s

Would be good to find a better way to check if the environment is sagemaker without making a network call. Wanted to take a look into speeding up launch even more at some point, 1.1s seconds are spent on just "import gradio".


gradio/analytics.py
```diff
@@ -27,15 +27,14 @@ def analytics_enabled() -> bool:
...
 def _do_analytics_request(url: str, data: dict[str, Any]) -> None:
+    data["ip_address"] = get_local_ip_address()
     try:
         requests.post(url, data=data, timeout=5)
     except (requests.ConnectionError, requests.exceptions.ReadTimeout):
         pass  # do not push analytics if no network
 
 
 def version_check():
-    if not analytics_enabled():
```
freddyaboulton, line 37: This API is broken - we're returning version 3.14 as the latest version lol

gradio/analytics.py
```diff
@@ -27,15 +27,14 @@ def analytics_enabled() -> bool:
 
 
 def _do_analytics_request(url: str, data: dict[str, Any]) -> None:
+    data["ip_address"] = get_local_ip_address()
```
freddyaboulton, line 30: Makes sense so that launching the thread does not have to wait for this to complete 👍

freddyaboulton (approved):
Thoughtful fix. Thanks @aliabid94

gradio/analytics.py
abidlabs: We haven't updated this since it prints an annoying message to anyone using an older version. Specifically chose 3.14 as anyone using an older version uses an insecure way to generate share links

But if needed this, can be updated here; https://github.com/gradio-app/gradio-api-server/blob/1b62edbd86eed452d901f4f0b76bb7824ca45827/server.py#L31

freddyaboulton: Got it. Yea would be annoying to receive every time you run an app. If it’s annoying maybe we can remove this function entirely?

abidlabs: I think its still valuable in rare cases we have a security situation and we want to push a message out to all users telling them to upgrade

freddyaboulton: Sound good

abidlabs:
I'll fix the tests so that we can merge this in

abidlabs (approved):
LGTM @aliabid94

PR #4916, opened 2023-07-13
Fix gr.Group, container=False

aliabid94:
The rendering for gr.Group and container=False had been wack for a while. Fixed in this PR. Only Textbox, Number, and Dropdown can set container=False.

In another PR, we can remove Forms entirely and use Groups to bind Form elements, since they render the same. Can do in 4.0 to deprecate Forms.

Recording 2023-07-13 at 13 40 04

Example code:
```python
import gradio as gr

def greet(name):
    return "Hello " + name + "!"

with gr.Blocks() as demo:
    gr.Markdown("### This is a couple of elements without any gr.Group. Form elements naturally group together anyway.")
    gr.Textbox("A")
    gr.Number(3)
    gr.Button()
    gr.Image()
    gr.Slider()

    gr.Markdown("### This is the same set put in a gr.Group.")
    with gr.Group():
        gr.Textbox("A")
        gr.Number(3)
        gr.Button()
        gr.Image()
        gr.Slider()

    gr.Markdown("### Now in a Row, no group.")
    with gr.Row():
        gr.Textbox("A")
        gr.Number(3)
        gr.Button()
        gr.Image()
        gr.Slider()

    gr.Markdown("### Now in a Row in a group.")
    with gr.Group():
        with gr.Row():
            gr.Textbox("A")
            gr.Number(3)
            gr.Button()
            gr.Image()
            gr.Slider()

    gr.Markdown("### Several rows grouped together.")
    with gr.Group():
        with gr.Row():
            gr.Textbox("A")
            gr.Number(3)
            gr.Button()
        with gr.Row():
            gr.Image()
            gr.Audio()

    gr.Markdown("### Several columns grouped together. If columns are uneven, there is a gray group background.")
    with gr.Group():
        with gr.Row():
            with gr.Column():
                btn = gr.Button("Hello")
                gr.Dropdown(["a", "b", "c"], interactive=True)
                gr.Number()
                gr.Textbox()
            with gr.Column():
                gr.Image()
                gr.Dropdown(["a", "b", "c"], interactive=True)
                with gr.Row():
                    gr.Number(scale=2)
                    gr.Textbox()

    gr.Markdown("### container=False removes label, padding, and block border, placing elements 'directly' on background.")
    gr.Radio([1,2,3], container=False)
    gr.Textbox(container=False)
    gr.Image("https://picsum.photos/id/237/200/300", container=False, height=200)

    gr.Markdown("### Textbox, Dropdown, and Number input boxes takes up full space when within a group without a container.")


    with gr.Group():
        output = gr.Textbox(show_label=False, container=False)
        with gr.Row():
            gr.Dropdown(["a", "b", "c"], interactive=True, container=False)
            gr.Textbox(container=False)
            gr.Number(container=False)
            gr.Image(height=100)


    gr.Markdown("### More examples")

    with gr.Group():
        gr.Chatbot()
        with gr.Row():
            name = gr.Textbox(label="Prompot", container=False)
            go = gr.Button("go", scale=0)

    with gr.Column():
        gr.Radio([1,2,3], container=False)
        gr.Slider(0, 20, container=False)

    with gr.Group():
        with gr.Row():
            gr.Dropdown(["a", "b", "c"], interactive=True, container=False, elem_id="here2")
            gr.Number(container=False)
            gr.Textbox(container=False)

    with gr.Row():
        with gr.Column():
            gr.Dropdown(["a", "b", "c"], interactive=True, container=False, elem_id="here2")
        with gr.Column():
           gr.Number(container=False)
        with gr.Column():
            gr.Textbox(container=False)


   

if __name__ == "__main__":
    demo.launch()
```


js/atoms/src/Block.svelte
```diff
@@ -35,6 +35,7 @@
 	style:overflow={allow_overflow ? "visible" : "hidden"}
 	style:flex-grow={scale}
 	style:min-width={`calc(min(${min_width}px, 100%))`}
+	style:border-width="var(--block-border-width)"
```
aliabid94, line 38: no idea why I needed to do this but was not picking up the css variable otherwise

gradio/components/dropdown.py
```diff
@@ -101,6 +101,13 @@ def __init__(
             raise ValueError(
                 "Custom values are not supported when `multiselect` is True."
             )
+        if not container:
+            if show_label:
+                warnings.warn("show_label has no effect when container is False.")
```
freddyaboulton, line 106: We should use `utils.warn_deprecation` here!

js/app/src/components/TimeSeries/TimeSeries.svelte
```diff
@@ -42,7 +42,6 @@
 	export let label: string;
 	export let show_label: boolean;
 	export let colors: Array<string>;
-	export let container: boolean = false;
```
freddyaboulton, line 45: In some of these files we delete the `container` prop (like in this one) but in others we keep the prop but don't pass it to the component (like video). I think we should remove the prop entirely from all the front-end components (except the form ones you laid out).

BTW, why only Dropdown, Number, Textbox and not all the form components? Just wondering

gradio/components/video.py
```diff
@@ -36,7 +36,6 @@ class Video(
 ):
     """
     Creates a video component that can be used to upload/record videos (as an input) or display videos (as an output).
-    For the video to be playable in the browser it must have a compatible container and codec combination. Allowed
```
freddyaboulton, line 39: Why was this deleted? A sed command gone wrong? 😂

freddyaboulton (commented):
Made some minor comments but otherwise this looks good to me! I would add the demo you made in the PR description to all_demos so we can keep track of this behavior before release (in lieu of app-level visual testing). Might be good to add some visual stories for the form components with and without containers too.

abidlabs:
@aliabid94 I was testing this branch in conjunction with the `gr.ChatInterface()` and I noticed a small issue. If you add `gr.Column()` within a `gr.Group()`, it produces a double border. Here's an example:

```py
with gr.Blocks() as demo:
    with gr.Group():
        gr.Chatbot()
        with gr.Row():
            with gr.Column(scale=10):
                gr.Textbox(container=False)
            with gr.Column(scale=1, min_width=0):
                gr.Button()

demo.launch()
```

Which produces:

image

Notice the double border on the bottom

This problem goes away if I remove the `gr.Column()`:

```py
with gr.Blocks() as demo:
    with gr.Group():
        gr.Chatbot()
        with gr.Row():
            gr.Textbox(container=False, scale=10)
            gr.Button(scale=1, min_width=0)

demo.launch()
```

but I need the `gr.Column()` because a user can supply their own submit button and I need to wrap it in a `gr.Column()` so that it scales appropriately

abidlabs:
Also even without `gr.Column()`, if you look at the border on top of the Textbox, its thicker than the border on the bottom:

image

I think its because the bottom border of the chatbot and the top border of the textbox are overlapping. This is a nit, but I think it is noticeable in the `gr.ChatInterface`. Would be great to fix if possible!

pngwn (commented):
The implementation looks good to me in general but I do have a question. 

What is the rationale in removing `container` on some of these component? I understand for components that fill their container completely things (like `Image`). But components like `Label`, `Checkbox`, `Plot` do not fill their group and user might want a simple way to remove both the container _and_ the padding that doesn't require some pretty specific CSS.

In this PR, it seem like `container=False` has only been considered in so much as it serve `Group` but `container=False` has broader application than that.

The other thing is that this is a massive breaking change. We need to add the `container` kwarg back for everything that had them previously with similar functionality until 4.0.

abidlabs:
> The other thing is that this is a massive breaking change. We need to add the container kwarg back for everything that had them previously with similar functionality until 4.0.

Yes we'll need to do that, and put a deprecation warning like @freddyaboulton suggested above

gradio/components/bar_plot.py
```diff
@@ -59,7 +59,6 @@ def __init__(
         interactive: bool | None = True,
         label: str | None = None,
         show_label: bool = True,
-        container: bool = True,
```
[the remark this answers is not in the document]
aliabid94: It shouldn't "break" in the sense of causing an error, because we pick up ununsed kwargs and throw a warning for them.

gradio/components/video.py
aliabid94: hahah yep! any line with "container" had to go

gradio/components/bar_plot.py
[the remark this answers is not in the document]
aliabid94: Actually restored container for everything, ignore this

js/app/src/components/TimeSeries/TimeSeries.svelte
aliabid94: Restored container everywhere.
Textbox, Dropdown, and Number have slightly container behavour in that instead on the block border disappearing, the input border element disappears. Makes it easier to play with gr.Group.

aliabid94:
I actually restored container kwarg for all components. Updated demo code at top that should showcase everything that has changed.

abidlabs:
Did a little bit of testing, and it looks `gr.Group()` seems apply a `gr.Column()` automatically.

For example, if you run:

```py
import gradio as gr

with gr.Blocks() as demo:
    with gr.Row():
        with gr.Group():
            gr.Button(scale=1)
            gr.Button(scale=1)
            gr.Button(scale=1)
        
demo.launch()
```

produces this:

image

If you reverse the order of the `gr.Group()` and `gr.Row()`, it renders properly:

```py
import gradio as gr

with gr.Blocks() as demo:
    with gr.Group():
        with gr.Row():
            gr.Button(scale=1)
            gr.Button(scale=1)
            gr.Button(scale=1)
        
demo.launch()
```

it renders properly:

image

aliabid94:
> and it looks gr.Group() seems to apply a gr.Column() automatically

Yeah this one's quite tricky, I struggled with it for a while. Would you consider it a blocker? fwiw before this PR, Groups also imposed a Column layout, they did not inherit flex-direction.

This will have a follow up PR later where I replace Form entirely with Group in 4.0, was thinking of handling this case then too.

abidlabs:
> Yeah this one's quite tricky, I struggled with it for a while. Would you consider it a blocker? fwiw before this PR, Groups also imposed a Column layout, they did not inherit flex-direction.

Nope not a blocker. Just something I noticed in testing. In fact, if it was the previous behavior, we should document this and keep as is to prevent apps from breaking when they upgrade `gradio`

I'll do some more testing of this PR to see if I notice any breaking changes!

gradio/components/base.py
```diff
@@ -133,7 +133,7 @@ def __init__(
         value: Any = None,
         label: str | None = None,
         info: str | None = None,
-        show_label: bool = True,
+        show_label: bool | None = None,
```
abidlabs, line 136: Would be good to update the docstring to reflect that `show_label=None` falls back to the `container` param

abidlabs:
Tested a bunch of demos, no issues from my end. `gr.ChatInterface` looks good with this PR as well. I went through the visual changes on Chromatic, they were all very minor so I approved them.

Would be good to get another pair of eyes on this, @pngwn @freddyaboulton

abidlabs (approved):

freddyaboulton (approved):
Thanks @aliabid94 ! Went through the demos again and did not notice any unintended regressions.

PR #4924, opened 2023-07-14
Fix API page code snippets in safari

aliabid94:
API page used to render copy buttons weird on safari. Fixed. Also quick fix to spacing in code snippet.


pngwn (approved):
Looks great. Reproed bug on main + confirmed fix in this PR. Thanks for fixing this so quickly @aliabid94!

Issue #3570, opened 2023-03-22
Support generator functions as `fn` in `gr.Example`


hysts:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  

When using `gr.Examples` with `cache_examples=True`, it will run the function specified with `fn=`, but it doesn't support generator functions.
I know there's a pretty simple workaround. We can just prepare a function that returns the last element of the generator function we want to apply and pass it to `fn` instead of the original generator function.
```python
def process_example(args):
    for x in generator_fn(args):
        pass
    return x
```

But the gradio feature to support generator functions as usual run callback is mainly for slow processes that can return intermediate results, and people want to cache examples for slow processes. So, I think it makes a lot sense to support it on gradio side.

hysts:
@abidlabs Thanks for the reply. I'm not sure if I follow what you mean.
I think it's just a bug on users' side to set an infinite generator for `fn` in `gr.Examples` with `cache_examples=True`. If users want to use infinite generator with `gr.Examples` for some reasons, they simply shouldn't set `cache_examples=True` because pre-computing examples and caching their results doesn't make sense for infinite generator.
To make it easier for users to notice it, maybe gradio can just show log messages that tell the start and end of each pre-computation of examples. With such logs, users can notice that something went wrong while processing the first example and that their function didn't terminate.
Also, I think users set `fn` and `cache_examples=True` at the very last stage of their development and probably they'll use the same function they used for the normal run. So, if they accidentally set infinite generator, they would have noticed it beforehand when they tested their function for the normal run.

abidlabs:
Got it, that makes sense. I think we can provide a reasonable UI around this. Will take a look next week

PR #4927, opened 2023-07-14
Allow caching generators and async generators

abidlabs:
Previously, we couldn't cache examples if the function was a generator or an async generator. This fixes that by getting the last yielded value and using that as the output. I also add a print statement to print out the status of caching. If there's a generator that takes a long time to finish iterating, users can notice that something went wrong and terminate (as suggested by @hysts)

Example:

```py
import gradio as gr
import time

def test_generator(x):
    for y in range(len(x)):
        time.sleep(0.3)
        yield x[: y + 1]

gr.Interface(test_generator, "textbox", "textbox", examples=[["abcdef"]], cache_examples=True).queue().launch()
```

This will be useful in `gr.ChatInterface` so that users can cache examples for streaming chatbots.

Closes: #3570


freddyaboulton (approved):
Looks good to me @abidlabs - nice we can finally support this

abidlabs:
Thanks for the awesome fast review @freddyaboulton

Issue #4325, opened 2023-05-23
RTL support in `gr.Textbox`


MohamedAliRashad:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
I want to build Arabic demo websites.

**Describe the solution you'd like**  
I want a toggle to switch the web app from `LTR` to `RTL`

abidlabs:
Hi @MohamedAliRashad can you elaborate more on what the desired behavior is? Is it just that the `Textbox` component should support text written from RTL, or something more?

MohamedAliRashad:
I was talking more about the whole AI. An option that i can flip from `False` to `True` that transforms the whole web page from `LTR` to `RTL`.

zaidalyafeai:
It seems the current `Textbox` doesn't support rtl languages like Arabic, correct me if I am wrong?

MohamedAliRashad:
@zaidalyafeai No, Nothing support RTL out of the box.

(BTW, Thanks for the `whisper-ar` model (and space) on huggingface.)

abidlabs:
In what sense are they not supported? I tested right now and the text is written from RTL as expected, although it is not right-aligned. Is that the concern? 

<img width="771" alt="image" src="https://github.com/gradio-app/gradio/assets/1778297/1deb9dd6-ca8c-4890-b1f1-dcbf8e9a4300">

MohamedAliRashad:
@abidlabs The Arabic text itself is supported. I am talking about the right-alignment using a flag in the gradio interface.
```python
gr.interface(rtl=True)
# or
gr.Blocks(rtl=True)
```
(It's possible to force something like this using global css i think).

abidlabs:
Yes understood @MohamedAliRashad. I was referring to @zaidalyafeai's comment about `gr.Textbox` specifically

zaidalyafeai:
Yeah, I meant the alighnment, something like html dir this will be cool 
```html
<input dir="rtl"/>
```

zaidalyafeai:
@abidlabs I am releasing a demo soon, is there any way we can add `dir='rtl'` to `textarea`? I have modified it using `inspect` and it works.

zaidalyafeai:
Also, allowing more freedom in passing styles will be awesome!
```python
gr.inputs.Textbox().style({'text-align': 'right'})
```

abidlabs:
[...]

PR #4933, opened 2023-07-16
Add RTL support to `Textbox`, `Markdown`, `Chatbot`

abidlabs:
Adds support for right-to-left languages, such as Arabic or Hebrew, to `gr.Textbox`, `gr.Markdown`, `gr.Chatbot`

I realized halfway through that this is a bit tricky, because there are two different aspects of an RTL language:
* The text should be right-aligned
* The direction of the typing should be on the left (the cursor, and new characters should appear on the left)

The way I've implemented this is as follows:
* The `Textbox` component now takes a boolean parameter `rtl` which sets direction to RTL and right-aligns the text
* In case someone wants to align the text but not change the directionality, there's a separate `text_align` parameter that only aligns the text
* `Markdown` and `Chatbot` also take the `rtl` parameter. Since they don't accept text, they don't have a separate `text_align` parameter.

Closes: #4325 

Test code:

```py
import gradio as gr

with gr.Blocks() as demo:
    gr.Textbox("شلاؤ abc", interactive=True)
    gr.Textbox("شلاؤ abc", interactive=True, text_align="right")
    gr.Textbox("شلاؤ abc", interactive=True, rtl=True)
    gr.Markdown("شلاؤ" + " abc")
    gr.Markdown("شلاؤ" + " abc", rtl=True)
    gr.Chatbot([[
        "شلاؤ" + " abc",
        "شلاؤ" + " abc"
    ]])
    gr.Chatbot([[
        "شلاؤ" + " abc",
        "شلاؤ" + " abc"
    ]], rtl=True)
    
demo.launch()
```


abidlabs:
I also added a few stories for visual testing of `Textbox`, `Markdown`, and `Chatbot` while I was at it. I didn't worry about tabulating the full parameter list (since they will eventually be populated programmatically)

Overall, it was very pleasant to add the visual tests!

pngwn (changes requested):
This isn't the right way to handle rtl, it isn't just about alignment. We should be using `dir=rtl` in all of these cases. It is supported by pretty much every element and handles alignment by default.

We will also want to support rtl for _all_ components i think, because everything has some text or label. Components like HighlightedText should also support rtl.

We also want to consider if we should support rtl at the document level, to make the whole gradio app rtl.

pngwn:
Would we not also want to flip the orientation of our layouts for global rtl content?

abidlabs:
> This isn't the right way to handle rtl, it isn't just about alignment. We should be using dir=rtl in all of these cases. It is supported by pretty much every element and handles alignment by default.

For the `gr.Textbox()`, which can be interactive, I did use the CSS property `direction` which is equivalent to `dir="rtl"`. I can change to use the HTML attribute, if that's preferable

I also added `text_align` as a separate property because there may be instances when users want to change the alignment, but _not_ the direction

Since `gr.Markdown()` and `gr.Chatbot()` do not accept user input, RTL is equivalent to setting the text to be right-aligned. At least as far as I can tell. Is there a difference?

> We will also want to support rtl for all components i think, because everything has some text or label. Components like HighlightedText should also support rtl, for which alignment won't work.

> We also want to consider if we should support rtl at the document level, to make the whole gradio app rtl.

We can certainly add RTL support to other components as well such `gr.HighlightedText()` as well as to the whole Blocks() but we can add them in a future PR. I prioritized these components because of recent interest in chatbots and upcoming announcement of `gr.ChatInterface`

pngwn:
> I also added text_align as a separate property because there may be instances when users want to change the alignment, but not the direction

Can't they do that with a theme?

abidlabs:
Disagree a bit @pngwn @hannahblair -- applying RTL to components vs. the entire Blocks are two separate issues. 

Applying RTL to the whole app is a lot more work since we have to modify entire layouts and is out of scope of this PR. 

But there are cases where only specific component (e.g. a `gr.Textbox()`) is expected to take e.g. Arabic and should be rendered RTL whereas the other parameters are in English. Here's kind-of an example of this: https://huggingface.co/spaces/bkhmsi/Font-To-Sketch

pngwn:
> Since gr.Markdown() and gr.Chatbot() do not accept user input, RTL is equivalent to setting the text to be right-aligned. At least as far as I can tell. Is there a difference?

As @hannahblair said, rtl handles more than just alignment. It will also lay out adjacent elements right to left, which alignment won't. This is a little more reliable, especially if we were to expand any of these components in the future to have more complex markup. With `dir=rtl`, it would 'just work' whereas alignment won't in many cases.

abidlabs:
> Can't they do that with a theme?

No a theme would apply to the entire app -- whereas this property applies on a component level

pngwn:
I've mentioned before that we could pass theme properties into components to have them only apply to that component. I'd argue that text alignment is stylistic (so a theme should handle it) but rtl is structural.

In the example space you linked, what would we do about the table elements/ examples?

The example does help clarify though, this is less about localisation and more about supporting multilingual apps with bits of rtl text.

abidlabs:
> The example does help clarify though, this is less about localisation and more about supporting multilingual apps with bits of rtl text.

Yes exactly! We could support full RTL too, but that's for a future PR :)

> In the example space you linked, what would we do about the table elements/ examples?

The examples actually look great -- since the only thing that needs to be rendered RTL is the text in the  "text" column, which is already the case

abidlabs:
Ok so maybe just to summarize the changes that are needed on this PR (as opposed to things we could include in a future PR):

* We should use the HTML `dir` attribute instead of the CSS `direction` 
* We should use that in the `gr.Chatbot()` and `gr.Markdown()` as well -- rename the Python parameter from `text_align` to `rtl` and implement it using the the `dir` attribute

Is that

pngwn:
@hannahblair The only issue with applying it to the parent is that it would `rtl` the labels and explanation text etc. Which i don't think is the intention of this PR. In order to _only_ rtl granularly, we can just add `dir=rtl` to things that currently have the text_align thing.

abidlabs:
Ok great, thanks @pngwn @hannahblair for the feedback! I'll make the changes requested

pngwn (approved):
Looks good! Thanks for fixing this @abidlabs!

abidlabs:
Thanks folks!

PR #4956, opened 2023-07-18
Filter-out special args when setting Interface component labels

cbensimon:


abidlabs:
Change looks good @cbensimon! Could we just add a test similar to this one to prevent regressions? https://github.com/gradio-app/gradio/blob/9787d0284ab7b504badfd2c99677657d447cf280/test/test_interfaces.py#L60

cbensimon:
> Change looks good @cbensimon! Could we just add a test similar to this one to prevent regressions?
> 
> https://github.com/gradio-app/gradio/blob/9787d0284ab7b504badfd2c99677657d447cf280/test/test_interfaces.py#L60

Done ! I confirm the new test fails without the fix and succeeds with the fix

```
=========================== short test summary info ============================
FAILED test/test_interfaces.py::TestInterface::test_input_labels_extracted_from_method
===== 1 failed, 504 passed, 54 deselected, 49 warnings in 81.97s (0:01:21) =====
```

abidlabs:
Perfect, thanks for the fix @cbensimon! Merging

Issue #4962, opened 2023-07-18
`gr.ChatInterface` - convert Submit button to Stop button while streaming


abidlabs:
While streaming, the Submit button in `gr.ChatInterface` should change to a Stop button that cancels the streaming

PR #4969, opened 2023-07-19
Fixes `cancels` for generators

abidlabs:
There was a bug in `cancels` where if a generator function was canceled, then subsequent runs of that function would continue iterating from where the previous iterator left off. Did some related cleanup of the code

I spent a lot of time trying to write an e2e test for this, but wasn't able to get it to work (see below). If anyone knows how to get it to work, I'm open to suggestions or feel free to push to this PR. But in the meantime, you can test by running the `cancel_events` demo or the test code below:

```py
import gradio as gr
import time

def test_streaming(message):
    for i in range(len(message)):
        time.sleep(1)
        yield message[: i+1]

with gr.Blocks() as demo:
    a = gr.Textbox()
    b = gr.Textbox()
    btn = gr.Button("Stop")
    ev = a.submit(test_streaming, a, b)
    btn.click(None, None, None, cancels=ev)
    
demo.queue().launch()
```

Steps to reproduce:
1. Run demo code
2. Start generator
3. Before generator completes, cancel it
4. Re-run generator (notice where it starts)

(Observed this while working on #4962)

### DIFF
--- a/gradio/routes.py
+++ b/gradio/routes.py
@@ -110,6 +110,7 @@ class App(FastAPI):
         self.blocks: gradio.Blocks | None = None
         self.state_holder = {}
         self.iterators = defaultdict(dict)
+        self.iterators_to_reset = defaultdict(set)
         self.lock = asyncio.Lock()
         self.queue_token = secrets.token_urlsafe(32)
         self.startup_events_triggered = False
@@ -389,14 +390,11 @@ class App(FastAPI):
         @app.post("/reset/")
         @app.post("/reset")
         async def reset_iterator(body: ResetBody):
-            print("iterators", app.iterators)
-            print("body", body)
             if body.session_hash not in app.iterators:
                 return {"success": False}
             async with app.lock:
                 app.iterators[body.session_hash][body.fn_index] = None
-                app.iterators[body.session_hash]["should_reset"].add(body.fn_index)
-            print("iterators", app.iterators)
+                app.iterators_to_reset[body.session_hash].add(body.fn_index)
             return {"success": True}
 
         async def run_predict(
@@ -404,6 +402,7 @@ class App(FastAPI):
             request: Request | List[Request],
             fn_index_inferred: int,
         ):
+            fn_index = body.fn_index
             if hasattr(body, "session_hash"):
                 if body.session_hash not in app.state_holder:
                     app.state_holder[body.session_hash] = {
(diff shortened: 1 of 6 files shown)


gradio/routes.py
```diff
@@ -409,21 +411,22 @@ async def run_predict(
...
                 # The should_reset set keeps track of the fn_indices
                 # that have been cancelled. When a job is cancelled,
                 # the /reset route will mark the jobs as having been reset.
                 # That way if the cancel job finishes BEFORE the job being cancelled
                 # the job being cancelled will not overwrite the state of the iterator.
-                # In all cases, should_reset will be the empty set the next time
-                # the fn_index is run.
-                app.iterators[body.session_hash]["should_reset"] = set()
+                if fn_index in app.iterators_to_reset[body.session_hash]:
+                    iterators = {}
```
abidlabs, line 420: The basic problem was that we would reset `app.iterators[body.session_hash]["should_reset"]` to an empty set here, meaning that (the old) line 451 would never take effect. That meant we would continue using the previous iterator instead of setting it to `None`, even if the function was reset

test/test_blocks.py
```diff
@@ -1324,24 +1336,6 @@ def iteration(a):
                 cancel.click(None, None, None, cancels=[click])
             demo.queue().launch(prevent_thread_lock=True)
 
-    @pytest.mark.asyncio
```
abidlabs, line 1327: This test wasn't really testing anything

abidlabs, line 1327: This test wasn't really testing anything new

abidlabs:
Here was the e2e test I attempted to write. However, apparently stopping and starting a websocket through an error in playright:

```js
 test("test that canceling and restarting a generator function starts from the beginning of the iteration", async ({
    page
}) => {
    let first_iteration;
    let fourth_iteration;

    const start_button = await page.locator("button", {
        hasText: /Start Iterating/
    });
    const stop_button = await page.locator("button", {
        hasText: /Stop Iterating/
    });
    const textbox = await page.getByLabel("Iterative Output");

    page.on("websocket", (ws) => {
        first_iteration = ws.waitForEvent("framereceived", {
            predicate: (event) => {
                return JSON.parse(event.payload as string).msg === "process_generating";
            }
        });

        fourth_iteration = ws.waitForEvent("framereceived", {
            predicate: (event) => {
                return JSON.parse(event.payload as string)?.output?.data?.[0] === "3";
            }
        });
    });

    await start_button.click();

    await fourth_iteration;
    await expect(textbox).toHaveValue("3");
    await stop_button.click();
    await start_button.click();
    await first_iteration;
    await expect(textbox).toHaveValue("0");
});


```

freddyaboulton:
@abidlabs I can't repro the original problem. In both the demos you mention in the PR description, the generator restarts from the beginning

cancel_demo_1

cancel_demo_2

abidlabs:
So strange! This is the behavior I see on `main` (after building the front end): 

Recording 2023-07-20 at 06 08 45

abidlabs:
Ok something interesting I noticed. For me, when I click the "stop iterating" button _twice_, the behavior is correct. I only see this bug if the "stop iterating" button is only clicked once. 

That's consistent with my understanding of the bug since it happens only when the the `cancels` event is called *while* the original event is still happening

abidlabs:
We can close this as these changes are also in #4971

pseudotensor:
@abidlabs This definitely happens for me, just not always.  I have users reporting the same thing: https://github.com/h2oai/h2ogpt/issues/538

Sometimes generators cancel, sometimes they don't.  It's not just about hitting it once or twice.

pseudotensor:
Also, why was it listed in bug fixes when the PR was closed?

https://github.com/gradio-app/gradio/blob/main/CHANGELOG.md#bug-fixes
```
Bug Fixes:
Fixes cancels for generators so that if a generator is canceled before it is complete, subsequent runs of the event do not continue from the previous iteration, but rather start from the beginning. By [@abidlabs](https://github.com/abidlabs) in https://github.com/gradio-app/gradio/pull/4969.
```

pseudotensor:
@abidlabs It's definitely a bad bug.  Sometimes it continues using the old iterator.

freddyaboulton:
This got merged in #4971. Does the bug still happen for you?

pseudotensor:
Sorry I didn't respond quickly.  Yes, that works well now without flakiness.  Still the generator, if using a thread like for streaming in transformers, can still go and I'd need to handle that separately.

PR #4971, opened 2023-07-19
Converts "Submit" button to "Stop" button in ChatInterface while streaming

abidlabs:
While the `gr.ChatInterface` is streaming, the Submit button gets converted to a Stop button. This PR includes the changes from the `cancelfix` branch (PR #4969) to make it easier to test out.  

You can test any chatbot with streaming, such as this:

```py
import gradio as gr
import time

def test_streaming(message, history):
    for i in range(len(message)):
        time.sleep(0.5)
        yield message[: i+1]

gr.ChatInterface(test_streaming).queue().launch()
```

Closes: #4962


gradio/chat_interface.py
```diff
@@ -123,24 +127,37 @@ def __init__(
...
-                                submit_btn, variant="primary", scale=1, min_width=0
+                                submit_btn, variant="primary", scale=1, min_width=150
                             )
                         else:
                             raise ValueError(
                                 f"The submit_btn parameter must be a gr.Button, string, or None, not {type(submit_btn)}"
                             )
-                    self.buttons.append(submit_btn)
+                    if stop_btn:
+                        if isinstance(stop_btn, Button):
```
freddyaboulton, line 144: We can check if the button `visible` prop is true here and raise an error if not

gradio/chat_interface.py
```diff
@@ -77,6 +79,7 @@ def __init__(
             css: custom css or path to custom css file to use with interface.
             analytics_enabled: Whether to allow basic telemetry. If None, will use GRADIO_ANALYTICS_ENABLED environment variable if defined, or default to True.
             submit_btn: Text to display on the submit button. If None, no button will be displayed. If a Button object, that button will be used.
+            stop_btn: Text to display on the stop button, which replaces the submit_btn when the submit_btn or retry_btn is clicked and response is streaming. Clicking on the stop_btn will halt the chatbot response. If set to None, stop button functionality does not appear in the chatbot. If a Button object, that button will be used as the stop button (the Button should set visible=False).
```
freddyaboulton, line 82: I think `submit_btn=None` is broken now

```bash
  File "/Users/freddy/sources/scratch/chatinterface_test.py", line 52, in <module>
    gr.ChatInterface(predict, submit_btn=None).queue().launch()
  File "/Users/freddy/sources/gradio/gradio/chat_interface.py", line 199, in __init__
    self._setup_events()
  File "/Users/freddy/sources/gradio/gradio/chat_interface.py", line 227, in _setup_events
    self._setup_stop_events(self.textbox.submit, submit_event)
  File "/Users/freddy/sources/gradio/gradio/chat_interface.py", line 330, in _setup_stop_events
    event_to_cancel.then(
  File "/Users/freddy/sources/gradio/gradio/events.py", line 139, in __call__
    dep, dep_index = self.trigger.set_event_trigger(
  File "/Users/freddy/sources/gradio/gradio/blocks.py", line 282, in set_event_trigger
  File "/Users/freddy/sources/gradio/gradio/blocks.py", line 282, in <listcomp>
AttributeError: 'NoneType' object has no attribute '_id'
```

With this


```python
gr.ChatInterface(predict, submit_btn=None).queue().launch()
```

gradio/chat_interface.py
```diff
@@ -270,11 +300,50 @@ def _setup_events(self):
...
+                )
+            else:
+                event_trigger(
+                    lambda: Button.update(visible=True),
+                    None,
+                    [self.stop_btn],
+                    api_name=False,
+                    queue=False,
+                )
+                event_to_cancel.then(
```
freddyaboulton, line 330: I don't think this is right. We're making `self.submit_btn` visible but in this branch submit_btn was not defined. I think we mean to hide the stop_btn in this case

freddyaboulton (commented):
Overall looks great and works great with all the default arguments! Left some comments about what happens when some of the buttons are None

gradio/chat_interface.py
abidlabs: Updated to just set the prop to False

gradio/chat_interface.py
abidlabs: Fixed

gradio/chat_interface.py
abidlabs: Nice catch, fixed

abidlabs:
Thanks @freddyaboulton everything should be fixed now!

gradio/chat_interface.py
abidlabs: Let me do this for a few other components as well actually

abidlabs:
Fixed the things that broke in merge, tests will hopefully pass now

gradio/routes.py
```diff
@@ -409,21 +411,22 @@ async def run_predict(
...
-                iterators = app.iterators[body.session_hash]
                 # The should_reset set keeps track of the fn_indices
                 # that have been cancelled. When a job is cancelled,
                 # the /reset route will mark the jobs as having been reset.
                 # That way if the cancel job finishes BEFORE the job being cancelled
                 # the job being cancelled will not overwrite the state of the iterator.
-                # In all cases, should_reset will be the empty set the next time
-                # the fn_index is run.
-                app.iterators[body.session_hash]["should_reset"] = set()
+                if fn_index in app.iterators_to_reset[body.session_hash]:
```
freddyaboulton, line 419: Don't quite understand why you were having issues with cancelling! Regardless, event cancelling is working well for me in this branch as well so I think we're ok to merge this change in.

freddyaboulton (approved):
LGTM @abidlabs ! Thanks for making the fixes.

freddyaboulton:
Gonna merge in! Thanks @abidlabs !

abidlabs:
Thanks @freddyaboulton appreciate the review and the merge!

PR #4979, opened 2023-07-20
Allow setting sketch color default

aliabid94:
Set brush_color via `gr.Image(brush_color="#FF0000")`.

Looks like this:

Recording 2023-07-24 at 05 54 26 (1)


gradio/components/image.py
```diff
@@ -109,9 +110,11 @@ def __init__(
             elem_classes: An optional list of strings that are assigned as the classes of this component in the HTML DOM. Can be used for targeting CSS styles.
             mirror_webcam: If True webcam will be mirrored. Default is True.
             brush_radius: Size of the brush for Sketch. Default is None which chooses a sensible default
+            brush_color: Default color of the brush for Sketch, should be hex code such as "#FF0000".
```
freddyaboulton, line 113: I think we should add this to the template classes too? Paint/SketchCam?

gradio/components/image.py
```diff
@@ -276,6 +282,10 @@ def preprocess(
 
         if self.tool == "sketch" and self.source in ["upload", "webcam"]:
             mask_im = processing_utils.decode_base64_to_image(mask)
+
+            if mask_im.mode == "RGBA":  # whiten any opaque pixels in the mask
```
freddyaboulton, line 286: I don't follow this. Can you explain what's happening?

gradio/components/image.py
dawoodkhan82, line 113: nit, maybe change to:
"Color of the brush for Sketch. Default is #000000"

dawoodkhan82:
Tested and looks good! Should we eventually also make the alpha of the brush settable?

dawoodkhan82 (approved):

gradio/components/image.py
aliabid94: yep, so previously in the frontend, whenever a user drew a mask, we were drawing it twice - once on the visible layer directly on the image, and then again on a hidden layer called "mask", which instead drew on a white brush on a black background. This mask layer was what was getting sent to the backend on submit.
I simplified the frontend so that there is a single mask layer, now visible, that hovers over the drawing. A user is no longer drawing directly on the image, but instead on this visible mask layer. On submit, we send this mask layer. However, this mask layer is no longer the white on black image of before - it has opacity and may be whatever color that brush_color has been set to. So now we convert this mask layer into the B+W mask of before in the backend with these lines of code.
We support "backwards compatibility" in case someone uses a client, with `if mask_im.mode == "RGBA": `, because previously masks were only RGB.
cc @pngwn for context to the frontend changes

aliabid94:
> Tested and looks good! Should we eventually also make the alpha of the brush settable?

I don't think there's a need for this. We can see if we get any requests.

gradio/components/image.py
pngwn: Are we certain that this conversion is always accurate (going from colours to b+w, especially since the sketches are antialiased)? Masks are typically alpha layers and lack any colours, I'm curious about why we would ever want a mask layer with colour.

abidlabs:
> Should we eventually also make the alpha of the brush settable?

Sorry for the late reply, but I agree with @dawoodkhan82 on this. We're making a breaking UI change with no way to for users to restore the original behavior. I would add a `brush_opacity` parameter (default can be `0.7` as in this PR)

aliabid94:
> Sorry for the late reply, but I agree with @dawoodkhan82 on this. We're making a breaking UI change with no way to for users to restore the original behavior

Disagree - there's slight opacity to a mask, which makes it easier to use. We are just going to have argument bloat if we add a configurable argument just for every minor UI change we've made.

abidlabs:
I agree that for most use cases that come to mind, setting the mask opacity to 0.7 is a good thing. However, with hundreds of thousands of gradio apps, there's a good chance that someone is using the mask in a different way than we imagine and wants the mask to be fully opaque. Upgrading gradio would break their demo without offering them a way to fix it. We've seen this problem several times already (changing the chatbot bubbles, making "enter" submit an Interface, etc.). I really think providing arguments to ensure backwards compatibility outweighs the concern of argument bloat.

gradio/components/image.py
aliabid94: Confirmed that antialiasing still works - there are values between 0 and 255 at the edges of the mask

aliabid94:
ok done, added `mask_opacity`

PR #4985, opened 2023-07-20
Adds `additional_inputs` to `gr.ChatInterface`

abidlabs:
Adds `additional_inputs` to `gr.ChatInterface`. 

Basic usage example:

```py
import gradio as gr
import time

def echo(message, history, system_prompt, tokens):
    response = f"System prompt: {system_prompt}\n Message: {message}."
    for i in range(min(len(response), int(tokens))):
        time.sleep(0.05)
        yield response[: i+1]

demo = gr.ChatInterface(echo, 
                        additional_inputs=[
                            gr.Textbox("You are helpful AI.", label="System Prompt"), 
                            gr.Slider(10, 100)
                        ]
                       )

if __name__ == "__main__":
    demo.queue().launch()
```

which produces this:

image

The nice thing is that you don't *have* to render the `additional_inputs` in this way. By putting the `gr.ChatInterface` in a `gr.Blocks`, you can have the `additional_inputs` show up wherever you want. For example, to have the system prompt show up on the top:

```py
import gradio as gr
import time

def echo(message, history, system_prompt, tokens):
    response = f"System prompt: {system_prompt}\n Message: {message}."
    for i in range(min(len(response), int(tokens))):
        time.sleep(0.05)
        yield response[: i+1]

with gr.Blocks() as demo:
    system_prompt = gr.Textbox("You are helpful AI.", label="System Prompt")
    slider = gr.Slider(10, 100, render=False)

    gr.ChatInterface(
        echo, additional_inputs=[system_prompt, slider]
    )

demo.queue().launch()
```

I've added tests and updated the documentation. I'll be OOO so feel free to push directly to this branch if needed or merge it in.


gradio/chat_interface.py
```diff
@@ -199,12 +212,25 @@ def __init__(
 
                 self.examples_handler = Examples(
                     examples=examples,
-                    inputs=self.textbox,
+                    inputs=[self.textbox] + self.additional_inputs,
                     outputs=self.chatbot,
                     fn=examples_fn,
-                    cache_examples=self.cache_examples,
+                    # cache_examples=cache_examples
```
freddyaboulton, line 218: I'm gonna delete this

gradio/chat_interface.py
```diff
@@ -130,9 +142,10 @@ def __init__(
                             self.textbox = Textbox(
                                 container=False,
                                 show_label=False,
+                                label="Message",
```
freddyaboulton, line 145: Why are we adding a label here if `show_label` is false?

freddyaboulton (approved):
Looks good to me @abidlabs ! Thanks for the tests and feature. Will merge by the end of the weekend unless there are other comments.

gradio/chat_interface.py
abidlabs: Thanks!

gradio/chat_interface.py
abidlabs: Because the label is used as the header of the table when examples are provided with additional inputs

abidlabs:
Appreciate the review @freddyaboulton! I’d keep the label but remove the comment, thanks

PR #4986, opened 2023-07-20
unified release process

### LINKED ISSUE #123: The ability to not have all those extra image editing tools, only cropping
One of our users wants a cleaner image editing interface with only cropping. Is that possible?

### LINKED ISSUE #4851: unify release process

We currently have two ways to release changes to packages in the gradio repo causing confusion and generating friction. 

Python packages (`gradio` and `gradio_client`) have a manual release process with manual changelog updates. This requires contributors to create a PR before they can update the changelog and causes lots of conflicts due to the same file being modified in most PRs. We also duplicate the workflow, each package has their own.

The JS packages (`@gradio/client` and `@gradio/lite`) use `changesets` to manage release which is a little simpler, reduces conflicts, and has no duplication.

Use changesets for everything.

pngwn:
closes #4851.

This PR overhauls our versioning and release process, and unifies it across the JS and Python files/libraries.

---

 Before I explain this PR I just want to point out that change will increase the frequency with which you have to pull changes from a PR branch because the bot will be pushing to the PR. There should never be conflicts because it will only create `changeset` files (so `.changeset/something.md`) but you may get errors if you try to push and you have pulled when the bot has added a commit.
 
 In order to work around this, I recommend setting up some global git config to make pulling easier:

```sh
git config --global push.autosetupremote true # not required but is nice
git config --global push.default upstream
git config --global branch.autosetupmerge simple # i think this is the default
git config --global pull.rebase true # rebase pull is cleaner than merge
```

These settings will automatically setup up the tracking branch for you (as long as you are working off this repo). If you need to pull you'll just be able to `git pull` without tracking remote explicitly and you'll also be able to just `git push` instead of `git push origin whatever`. You don't _have_ to do this but it'll make life way easier. Being able to just `push` / `pull` is super nice.

---

There are quite a few moving parts in this PR (honestly it was a much more enormous amount of work than i expected), so I'll will explain what happens and then i will explain how.

**What does it do**

This PR introduces two distinct things: automatic versioning/ changelog generation/ release and automatic changeset generation. This will make sense in a sec.

Whenever we make a PR that requires a release (so any code change) instead of updating the changelog manually, we just add a `changeset` file that describe the change. The only information needed is:

- the package(s) that have changed
- the type of change (`patch`, `minor` or `major`)
- a description of the change (for the changelog, so friendly descriptions)
- the 'type' of change (`fix`, `feat`, `highlight`). This is to allow us to format our changelogs nicely for each releases. We can add more `types` if we thing they are necessary.

A changeset file looks like this:

```md
---
<package-name>: <patch|minor|major>
<package-name>: <patch|minor|major>
...
---
<fix|feat>: description
```

A changeset with a `highlight` type ha slightly different requirements:

```md
---
<package-name>: <patch|minor|major>
<package-name>: <patch|minor|major>
...
---
highlight: 

#### a title

some content
```

`highlight` types _must_ have a level 4 markdown heading on a separate line from the `highlight:` text. This looks nicer and is validated in CI.

When this file is added to a PR and that PR is merged, a robot will create a new 'release' PR this PR takes care of figuring our the correct version bump for each package based on the combined changesets file, it also generates a changelog from those changesets. It link the PR, the merge commit and thanks the user (with a link) in the change logs. Normal entries look like this:

```md
- [#1234](https//link.to.pr) [`abcd123`](https://link.to.ref) - The description of the change. Thanks [@user](https://link.to.user)!
```

Rendered:

> - [#1234](https://link.to.pr) [`abcd123`](https://link.to.ref) - The description of the change. Thanks [@user](https://link.to.user)!

highlight entries look like this:

```md
#### The description of the change / title ([#1234](https//link.to.pr) [`abcd123`](https://link.to.ref))

any content

Thanks [@user](https://link.to.user)!
```

Rendered:

> **The description of the change / title ([#1234](https://link.to.pr) [`abcd123`](https://link.to.ref))**
>
> any content
>
> Thanks [@user](https://link.to.user)!

This generated PR also bumps all of the version files (package.json + version.txt) and bumps the version of any local dependencies (so when `gradio_client` gets bumped, the `requirements.txt` of `gradio` gets bumped to match this version).

When we merged this PR, the second part of teh same CI workflow will see that the changesets have been removed + the version updated. It will then check that version does not exist, run any build scripts and release the package. Python packages are expected to have a `build_pypi.sh` script in their directory (we can move this to `scripts/` if needed).

`changesets` will also generate tags + github releases but I need to double check that is enabled.

The second part os the automatic changeset generation. Generating the changesets can be a bit tedious so we have a bot that does it for you. The bot does this:

- By default, it will check the changed packages, check any linked issues (closes #123) and generate the correct changeset based on that. If there are no release it will set the bump as `minor` and the type as `feat`.
- We can change the version and type with labels. `v: (patch|major|minor)` for the bump type and `t: (fix|feat|highlight)` for the 'type'.
- We can select the packages that should be affected if the bot is wrong. There is a PR comment that we can interact with to enter "manual package selection mode". Then we will get a checklist of all packages that we can choose from
- We can enter full manual mode byt editing the generated changeset file. The bot will stop doing stuff (other than updating the comment) if the changeset file has been edited manually.
- We can delete the changeset file and remove all labels to start again.

**How does it work**

**Changeset generation**

- When a PR is opened, reopened, synchronised, edited or labelled, or the issue generated by the job is edited a CI job runs. This job does nothing and only exists to trigger the action that does the work (`pull_request` event or `issues_edited` is the event used here).
- Another action (with the actual logic in it) will run when the above action has finished (`workflow_run` event). This is for security reasons. We need access to secrets to do anything but we don't want PRs from forks to have access to those secrets. `pull_request` evens do not expose secrets. This 'other' action runs in the context of main, so a PR would need to be merged any code within it could have any impact.
- This action figures out the PR via a reusbale action which can be found [here](https://github.com/gradio-app/github/tree/main/packages/get-pr-branch). This action uses the source repo + branch name and cross-references that with any open PRs to find the PR number.
- When it has the PR number, this is passed into the changeset generation / update action which is also reusable and can be found [here](https://github.com/gradio-app/github/tree/main/packages/generate-changeset). I can go through how this works if needed but there is a bit of complexity there.
- One difference between vanilla changesets and our version is how we handle dependencies. Pretty much any change we make should end up in the `gradio` changelog in order to be useful to users. This isn't typically how it works in JS. The default behaviour is for individual packages to contain their changelog (which is good) but for packages that depend on those packages to simple contain an 'dependencies updated' changelog entry (which is not very useful). So I have introduced a `main_changeset` field to the package.json of any package whose changelog should be added to the 'main' package, in this case `gradio`. What this mean is that if you change the `Plot` component's JS/Svelte code with a title of "ensure the plot uses the correct styling in dark mode" that description will be added to the `@gradio/plot` package (which is nice for us to keep track) but it will _also_ be added to the `gradio` package (which is nice for our users).

**Versioning**

- We build on top of changesets to do this, this is why i have added `package.json` files to the gradio libs, to kind of treat them as 'packages' (changesets only really works with npm packages).
- We follow the normal changesets flow with a few customisation. The changeset files we generate match what changesets expects, so we don't need to do anything clever there. All we really need to do is add a little metadata to introduce the concept of `type` (`fix`, `feat`, `highlight`). So when making PRs and merging them into `main` the flow is pretty consistent with the changesets documentation. 
- After we have merged into main we do a bunch of more custom stuff. There are two scripts that handle this. `./changeset/changeset.cjs` and `./changeset/fix_changelogs.cjs`. 
- changesets doesn't have an official API to create a custom changelog format, so we need to hack around it a bit.
- `./changeset/changeset.cjs` has a function or two that run for every 'release line', i.e. every changelog entry for each package. We do some github api lookups to get some details about a PR, parse the changeset file to see if it is a `fix`, `feat`, or `highlight` and then write those details to a JSON file. These function also return a string that becomes the new changelog for each package but we don't care about that because we overwrite them in the next step. 
- The above command runs as part of the `changeset version` command, which also bumps the JS packages. As we are storing the version for python packages in `package.json` files they get bumped correctly, however they are not in the correct place (`version.txt`), so we need to copy those new version into a few places to play nice with the python conventions. This is also when see if any python packages depend on another python package locally. In this case `gradio` depends on `gradio_client`. If `gradio_client` has been bumped, then we will update the `requirements.txt` of `gradio` to match this new version. This is because that is how we tests in CI, latest with latest, so that is what we pin to.
- This is _also_ where we reformat the changelogs to respect our custom format (which is with `Highlights`, "Features", and "Fixes" sections). We use the JSON file we saved earlier to know what has changed for each package.
- The `changesets` action  creates a descriptive PR message from the changelogs themselves, so this pretty much works without any intervention.

**Publishing**

This is also part of the 'changesets' CI job.

- The JS stuff 'just works' by default because that is what changesets does.
- The python stuff does not. We have some custom stuff to make that go.
- The logic lives in [this reusable action](https://github.com/gradio-app/github/tree/main/packages/publish-pypi). This is basically the same as the pypi action we used to use except faster (and not quite as feature rich). 
 - It checks that the version have not already been published. If they have it bails.
 - It runs the `build_pypi.sh` script that ever package is expected to have
 - It uploads the new version with `twine`
 - (I can go through this action in more detail if needed as well)

---

checklist for me as i go through.

- [x] change `pngwn/gradio` reference to ``gradio-app/gradio`
- [x] check all python modules have a build script in the correct place
- [x] remove the current changesets ci thing
- [ ] reformat existing changelogs (i'll explain this)
- [x] check all secrets (i think i named them the same but need to double check)
- [x] check all package names + package.json are correct
- [x] explain wtf this does and where everything is in this PR body


gradio/package.json
```diff
@@ -0,0 +1,6 @@
+{
+	"name": "gradio",
+	"version": "3.37.0",
```
abidlabs, line 3:
```suggestion
	"version": "3.38.0",
```

pnpm-workspace.yaml
```diff
@@ -1,3 +1,5 @@
 packages:
   - 'js/*'
-  - 'client/js'
+  - "gradio"
+  - "client/*"
+  - "client/*/*"
```
abidlabs, line 5: This isn't supposed to be
```suggestion
  - "client/**/*"
```
is it?

abidlabs:
Very cool @pngwn! Thanks for putting this together. I looked through the changes, I couldn't follow everything, but down to  test it out with the next release

pnpm-workspace.yaml
pngwn: I might actually just make this client/python/*. The reason I didn't use the more general glob is because I don't want it to make stuff inside node modules which I think the looser glob would. Will check.

.github/workflows/generate-changeset.yml
```diff
@@ -0,0 +1,38 @@
...
+        run: echo "PR number is ${{ steps.get-pr.outputs.pr_number }}"
+      - uses: actions/checkout@v3
+        with:
+          repository: ${{ github.event.workflow_run.head_repository.full_name }}
+          ref: ${{ github.event.workflow_run.head_branch }}
+          fetch-depth: 0
+      - name: generate changeset
+        uses: "gradio-app/github/actions/generate-changeset@main"
+        with: 
+          github_token: ${{ secrets.PAT }}
```
freddyaboulton, line 36: I just checked and @gradio-pr-bot is in the org so I think we can use `secrets.COMMENT_TOKEN` here.

.github/workflows/publish-npm.yml
```diff
@@ -42,9 +42,11 @@ jobs:
...
-          PR_NUMBER: ${{ steps.changesets.outputs.pullRequestNumber }}
-          GITHUB_TOKEN: ${{ secrets.GRADIO_PAT }} # Not sure if this one is required
\ No newline at end of file
+      - name: publish to pypi
+        if: steps.changesets.outputs.hasChangesets != 'true'
+        uses: "gradio-app/github/actions/publish-pypi@main"
+        with:
+          user: __token__
+          passwords: |
+            gradio-test-pypi:${{ secrets.PYPI_API_TOKEN }}
```
freddyaboulton, line 51: These tokens look good to me.

freddyaboulton (approved):
@pngwn the CI master !!! Excited to test this out!!

js/button/README.md
```diff
@@ -1,4 +1,4 @@
-# `@gradio/button`
+# `@gradio/boop`
```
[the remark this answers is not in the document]
pngwn: ah, yes.

.github/workflows/generate-changeset.yml
```diff
@@ -0,0 +1,38 @@
...
+
+jobs:
+  version:
+    permissions: write-all
+    name: static checks
+    runs-on: ubuntu-22.04
+    if: (github.event.workflow_run.head_repository.full_name == 'gradio-app/gradio' && github.event.workflow_run.head_branch != 'main') || github.event.workflow_run.head_repository.full_name != 'gradio-app/gradio'
+    steps:
+      - id: 'get-pr'
+        uses: "gradio-app/github/actions/get-pr-branch@main"
```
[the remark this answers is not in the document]
pngwn: It's similar but not quite the same. I'm not 100% certain if that action would work for us as ours is tailored for workflow runs (and doesn't require any input), we use the branch name + repo owner to check the PR rather than a commit sha which is what i think that action is doing. It could work tho, not certain.

.github/workflows/generate-changeset.yml
pngwn, line 36:
```suggestion
          github_token: ${{ secrets.COMMENT_TOKEN }}
```

js/button/README.md
pngwn, line 1:
```suggestion
# `@gradio/button`
```

PR #4991, opened 2023-07-20
Release 3.38

abidlabs:
3.38.0 Release. 

Added the `chatinterface_streaming_echo` to `all_demos` so that we can preview `gr.ChatInterface` with every PR.

Removed a test that was flaking because our API server is acting weird. The test is pretty pointless so I've removed it


freddyaboulton (approved):
Pushed up the theme variable rename. The deployed demos look good to me!!

abidlabs:
Fixed one small thing (size of the submit button and stop buttons weren't the same on small screens), but otherwise, lgtm as well. Merging after CI passes!

PR #4998, opened 2023-07-22
Fix changesets action

pngwn:
I forgot to filter an array.


pngwn:
This doesn't touch anything, so I'm gonna merge in to test it.

Issue #5002, opened 2023-07-23
Documentation issue: wrong example for the ".then()" functionality


fabiencro:
### Describe the bug

Hi,

I noticed that the example code for the section [Running Events Consecutively](https://www.gradio.app/guides/blocks-and-event-listeners#running-events-consecutively) in the guide "Blocks and event Listener" do not match the section and its description. Indeed, there is no use of `.then()`, and the code will not do what the description says (i.e. updating the user message first, and later generate the bot message).

A possible fix would be this example code:
```python
import gradio as gr
import random
import time

with gr.Blocks() as demo:
    chatbot = gr.Chatbot()
    msg = gr.Textbox()
    # clear = gr.ClearButton([msg, chatbot])
    

    def respond(message, chat_history):
        bot_message = random.choice(["How are you?", "I love you", "I'm very hungry"])
        chat_history[-1][1]= bot_message
        time.sleep(2)
        return "", chat_history

    def update_message(message, chat_history):
        chat_history.append([message, None])
        return "", chat_history
    
    msg.submit(update_message, [msg, chatbot], [msg, chatbot]).then(respond, [msg, chatbot], [msg, chatbot])

demo.launch()
```

If I have time, I will submit a small PR for this.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

Go to https://www.gradio.app/guides/blocks-and-event-listeners#running-events-consecutively and check that the code under the section "Running Events Consecutively" do not match the description.

### System Info

```shell
This is from the current version of the official website
```

### Severity

I can work around it

PR #5003, opened 2023-07-23
Documentation: made correct example for the .then() event listener method

fabiencro:
## Description
Small fix to the documentation. The example code for the section [Running Events Consecutively](https://www.gradio.app/guides/blocks-and-event-listeners#running-events-consecutively) in the guide "Blocks and event Listener" was not matching the section and its description:
1. no use of .then()
2. code do not do what the description says (i.e. updating the user message first, and later generate the bot message).

I made a demo code that works (mostly by slightly modifying the existing code).

However, I am not sure I totally understand the way the doc/website is generated. Someone should check that this was done the correct way.

Closes: #5002


freddyaboulton (approved):
Thanks @fabiencro !

Issue #4307, opened 2023-05-22
Add a download button for Gallery


innat-asj:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
In gradio-based apps that give visual output(s) should have a download option. Currently the `gr.Image()` supports this feature but `gr.Gallery` does not yet. 

**Describe the solution you'd like**  
As said, the `gr.Image()` has this option. It would be great to add support for `gr.Gallery`.

PR. https://github.com/gradio-app/gradio/issues/3155

**Additional context**  

1. Galley may or may not have 1 or multiple visual outputs. Each should contain a download button separately which downloads the image as PIL or np array. And the output widget should also have a global download button which will first zip the output(s) and download as a zip file. 
2. Respecting the input file_name. extension with `output` as a postfix would be a great addition. https://discuss.huggingface.co/t/how-to-download-files-with-name-and-extension/19907
3. For `gr.Image` or `gr.Gallery`, IMO, there should be a positional argument for the visual widget. For example `gr.Image(download=True)` or `gr.Gallyer(download=False)` etc.

## Reproducible Code (RC)

```python
import numpy as np
import gradio as gr # v. 3.32.0

def flip_text(x):
    return x[::-1]

def flip_image(x):
    return [np.fliplr(x)] + [x]
```
```python
with gr.Blocks() as demo:
    with gr.Row():
        image_input = gr.Image()
        # image_output = gr.Image() # download [ok]
        image_output = gr.Gallery(
                label='Output',
                show_label=False,
                elem_id='gallery'
            ).style(
                grid=2,height='auto'
            )
    image_button = gr.Button("Flip")
    image_button.click(
        flip_image, 
        inputs=image_input, 
        outputs=image_output
    )
demo.launch(server_port=8088)
```

Issue #3158, opened 2023-02-09
Do not serialize images into base64 format before they are sent to the frontend


abidlabs:
For all of the file-based components, except for the `Image` component, we no longer serialize data into base64 format. Rather, we return the path to a copy of the file. This has the advantage of being much faster, particularly for large files. If possible, we should do the same for images as that would avoid issues like #2635.

However, it might be the case that our `Image` component requires the image format to be base64 format so that users can edit the image (this used to be the case with our previous complex image editor, @pngwn do you know if this is still the case?). In which case, we could potentially provide advanced users with a parameter that lets them choose whether the image is serialized to base64 or not.

If we do this, this should happen in 4.0 as it would change the REST API format of Spaces using the Image component.

pngwn:
This wouldn't solve #2635 because those files are causing a problem before the backend event knows about them. For images, video, and audio we need some kind of valid URL. On the frontend we always convert to a base64, so we can use that as a source for the relevant media element (which we need for both previews and editing). To change this we need to immediately send the image to the backend to create a temp file + return the filepath when they are uploaded (not when the prediction is run).

@aliabid94 Is looking at #2635 atm so he can discuss what he is looking at currently to solve that issue but essentially the solution is to just send the blob rather than converting to a string (which is the slow part on the frontend), this is only currently viable for `File`s and nothing else, we need more significant changes for other media types.

MilkClouds:
In my opinion, simply not using base64 would not resolve https://github.com/gradio-app/gradio/issues/2635, since it is not because of base64; it is because of image encoding logic of pillow. If gradio use pillow to save image to file and upload the url of the file, same issue would occur because there exists a "save image" logic.

pngwn:
Fixed by #6169.

PR #5025, opened 2023-07-26
Add download button to selected images in `Gallery`

hannahblair:
## Description

Adds a download button icon next to the clear button when selecting image in a `Gallery`.

**_Note - edge case for cross-origin data_**

We currently don't convert the image values passed into the `Gallery` component to base64 like we do with `Image`. This means that if a cross-origin `href` is used (e.g. an external image link on unsplash) in the anchor tag, it doesn't directly download to the user's file system, but rather opens up a new tab with the image link (due to browser security policy). We can either leave this be or implement base 64 conversion to `Gallery` images. Open to thoughts!

Demo 

```
import gradio as gr
import pathlib

current_dir = pathlib.Path(__file__).parent

images = [str(current_dir / "../../test/test_files/bus.png"), str(current_dir / "../../test/test_files/cheetah1.jpg")]

with gr.Blocks() as demo:
    with gr.Row():  
        gal2 = gr.Gallery(
            value=images,
            label="Gallery local files",
            show_download_button=True,
        )

    with gr.Row():
        gal1 = gr.Gallery(
            label="Gallery external imgs",
            value=[
                "https://source.unsplash.com/random/?Dog&1",
                "https://source.unsplash.com/random/?Dog&2",
                "https://source.unsplash.com/random/?Dog&3",
            ],
            show_download_button=True,
        )

demo.launch()
```

Closes: #4307 

## 🎯 PRs Should Target Issues

## Tests


aliabid94:
Thanks Hannah, looks great but I couldn't get it to work right when running demo/gallery_selections/run.py and setting `show_download_button=True`. See screenshot below:
Screen Shot 2023-07-26 at 12 12 31 PM

The download button links to "127.0.0.1:7860/[Object object]".

abidlabs:
> it doesn't directly download to the user's file system, but rather opens up a new tab with the image link (due to browser security policy). We can either leave this be or implement base 64 conversion to Gallery images. Open to thoughts

Let's leave as is. Conversion to base64 can be very slow for large images. In fact, there's an existing issue to _not_ convert `Image` component images to base64 (#3158), part of the 4.0 milestone

dawoodkhan82:
Can we have the download button show by default. Do we need a param for this at all? Maybe I'm missing something.

dawoodkhan82 (changes requested):
@hannahblair I'm also seeing the same error @aliabid94 was seeing. The download button still links to `http://localhost:9876/[object%20Object]` for me.
Screenshot 2023-07-28 at 8 13 46 PM

dawoodkhan82 (approved):

Issue #5029, opened 2023-07-28
Gradio Highlittext method not working


rajeunoia:
### Describe the bug

I am using, 
gradio==3.39.0
gradio_client==0.3.0
i get following error , 
gradio with ner events fails with this error, highlighted_text.py", line 162, in postprocess (text[entity[“start”] : entity[“end”]], entity[“entity”]) keyerror: ‘entity’

I think the issue is with the key "entity" where output of the model is giving "entity_group"

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

Code that i have tried, which calls highlight in output
```
demo = gr.Interface(fn=ner,
                    inputs=[gr.Textbox(label="Text to find entities", lines=2)],
                    outputs=["highlight"],
                    title="NER with dslim/bert-base-NER",
                    description="Find entities using the `dslim/bert-base-NER` model under the hood!",
                    allow_flagging="never",
                    #Here we introduce a new tag, examples, easy to use examples for your application
                    examples=["My name is Andrew and I live in California", "My name is Poli and work at HuggingFace"])
demo.launch(share=True, server_port=7862)
```

Code in 162 line of highlighted_text.py
```
list_format.append(
                        (text[entity["start"] : entity["end"]], entity["entity"])
                    )
```

### Logs

```shell
gradio with ner events fails with this error, highlighted_text.py", line 162, in postprocess (text[entity[“start”] : entity[“end”]], entity[“entity”]) keyerror: ‘entity’
not the complete error log
```

### System Info

```shell
MacBook Pro , M1 chip
```

### Severity

I can work around it

rajeunoia:
My code is from the latest short course on Deeplearning.ai

def get_completion(inputs,
parameters=None,ENDPOINT_URL=HF_API_SUMMARY_BASE):
    headers = {
      "Authorization": f"Bearer {hf_api_key}",
      "Content-Type": "application/json"
    }
    data = { "inputs": inputs }
    if parameters is not None:
        data.update({"parameters": parameters})
    response = requests.request("POST",
                                ENDPOINT_URL, headers=headers,
                                data=json.dumps(data)
                               )
    return json.loads(response.content.decode("utf-8"))

import gradio as gr
API_URL = "https://api-inference.huggingface.co/models/dslim/bert-base-NER"
text = "My name is Andrew, I'm building DeepLearningAI and I live in
California"
get_completion(text, parameters=None, ENDPOINT_URL= API_URL)

Output:

[...]

PR #5030, opened 2023-07-28
highlightedtext throws an error basing on model

rajeunoia:
## Description

It closes an issue , highlightext fails with an error reported in Issue 5029 , as the model output entity key name is different , fixed to accept both entity and entity_group, not sure if there are any other possibilities. 

Closes: #5029 

## 🎯 PRs Should Target Issues

## Tests


gradio/components/highlighted_text.py
```diff
@@ -158,8 +158,9 @@ def postprocess(
                 entities = sorted(entities, key=lambda x: x["start"])
                 for entity in entities:
                     list_format.append((text[index : entity["start"]], None))
+                    entity_group = "entity_group" if "entity_group" in entity else "entity"
                     list_format.append(
-                        (text[entity["start"] : entity["end"]], entity["entity"])
+                        (text[entity["start"] : entity["end"]], entity_group)
```
abidlabs, line 163: This should be 
```suggestion
                        (text[entity["start"] : entity["end"]], entity[entity_group])
```
right?

abidlabs:
Hi @rajeunoia thanks for issuing this fix. Fix looks good, I'll just add a quick test and then we can merge it in.

abidlabs:
Should be good to go now. Can be tested with this code:

```py
import requests, json, gradio as gr

API_URL = "https://api-inference.huggingface.co/models/dslim/bert-base-NER"

def get_completion(inputs):
    data = { "inputs": inputs }
    response = requests.request("POST",
                                API_URL, 
                                data=json.dumps(data)
                               )
    return {"text": inputs, "entities": json.loads(response.content.decode("utf-8"))}

gr.Interface(get_completion, "text", "highlightedtext").launch()
```

abidlabs:
Merging, thanks @rajeunoia!

Issue #4668, opened 2023-06-25
Multimodal Input TextBox


taoari:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  

Multimodal LLMs become popular nowadays. However, for multimodal input, the current gradio app has to use separate widgets for images, videos, audio, and files (attachments). The UI is super non-intuitive, it would be good to have a multimodal input textbox.

**Describe the solution you'd like**  

Any modern chat app has a multimodal input textbox, e.g. Slack, Teams, etc. The screenshot would be the Slack input box, it would be nice to has something similar.

![image](https://github.com/gradio-app/gradio/assets/3815006/0b95aba4-92f4-444b-b2a7-12132cfa68ca)

**Additional context**  

It would also be great that the gr.Chatbot can be updated accordingly that can show text, images, videos and attachments in a single message. The current version of gr.Chatbot only shows a single modality (text or image but not both in one message). PR467 https://github.com/gradio-app/gradio/issues/4667 is a bug that does show the file. It would also be great if the Chatbot can show 3D models, as there is a gr.Model3D component.

pngwn:
I don't know if we'd want a full rich text input like slack but something like discord or WhatsApp  might be nice. The ability to upload text along with various media (images, audio, video). 

Cc @dawoodkhan82

dawoodkhan82:
@pngwn @taoari I've actually thought about this, and I think it's a good idea. Especially as more multimodal projects become popular, it would be good to have a component that supports them. I think making the chatbot a single component (input + output) also makes a lot of sense plus easier for devs to use. We can explore this in 4.0

taoari:
@pngwn @dawoodkhan82 It's great to see this is on the wish list, looking forward to it.

abidlabs:
The rich textbox should be a separate component, particularly if we want to support files, as that would involve changing the API (we could do a similar tuple format to support files).

dawoodkhan82:
@abidlabs we can allow file upload and the text styling features to be turned off for the rich textbox, in case a dev wants only one feature and not the other.

abidlabs:
Hey! We've now made it possible for Gradio users to create their own custom components -- meaning that you can write some Python and JavaScript (Svelte), and publish it as a Gradio component. You can use it in your own Gradio apps, or share it so that anyone can use it in their Gradio apps. Here are some examples of custom Gradio components:

[...]

PR #5032, opened 2023-07-28
RichTextbox component

dawoodkhan82:
## Description

Adds a new component called `Richtextbox`. The `Richtextbox` component allows users to submit text and file uploads all from one component. Makes creating `Chatbot` demos a lot easier. The component allows for `single` or `multiple` file upload, and can be specified by the `file_count` parameter.

https://github.com/gradio-app/gradio/assets/12725292/cda08dac-50cd-4ff1-89c1-bbe2d3bb2017

Closes: #4668

## 🎯 PRs Should Target Issues

## Tests


abidlabs:
Exciting! When trying to build the frontend using `bash scripts/build_frontend.sh`, I get this error:

```
[vite]: Rollup failed to resolve import "@gradio/multimodaltextbox" from "/Users/abidlabs/dev/gradio-repos/gradio/js/app/src/components/MultimodalTextbox/index.ts".
```

Do I need to do something else first?

dawoodkhan82:
> Do I need to do something else first?

@abidlabs  I think you might have to run `pnpm i` to install the packages first.

abidlabs:
Tried that, I got:

```
 ERR_PNPM_NO_MATCHING_VERSION_INSIDE_WORKSPACE  In js/app: No matching version found for @gradio/model3D@workspace:^0.0.1 inside the workspace
```

abidlabs:
Still seeing the same errors @dawoodkhan82 :(

dawoodkhan82:
@abidlabs I realized I didn't properly push the changes. try now

abidlabs:
Cool so `pnpm i` is working. However after running that, if I try to build the frontend using `scripts/build_frontend.sh`, it still errors out with the error:

```
```

dawoodkhan82:
@abidlabs fixed (hopefully).

dawoodkhan82:
Realized I forgot to add a kwarg to disable the upload button. Will do that

abidlabs:
Nice, it builds now. Taking a look now @dawoodkhan82

demo/hello_blocks/run.py
abidlabs, line 1: I wouldn't change the `hello_blocks` demo -- we use it in a lot of introductory places and adding a `MultimodalTextbox` will complicate the examples.

abidlabs: Instead we can add a new demo if needed.

gradio/components/multimodaltextbox.py
```diff
@@ -0,0 +1,267 @@
...
+            elem_id: An optional string that is assigned as the id of this component in the HTML DOM. Can be used for targeting CSS styles.
+            elem_classes: An optional list of strings that are assigned as the classes of this component in the HTML DOM. Can be used for targeting CSS styles.
+            type: The type of textbox. One of: 'text', 'password', 'email', Default is 'text'.
+            text_align: How to align the text in the textbox, can be: "left", "right", or None (default). If None, the alignment is left if `rtl` is False, or right if `rtl` is True. Can only be changed if `type` is "text".
+            rtl: If True and `type` is "text", sets the direction of the text to right-to-left (cursor appears on the left of the text). Default is False, which renders cursor on the right.
+            show_copy_button: If True, includes a copy button to copy the text in the textbox. Only applies if show_label is True.
+            file_count: if single, allows user to upload one file. If "multiple", user uploads multiple files. If "directory", user uploads all files in selected directory. Return type will be list for each file in case of "multiple" or "directory".
```
abidlabs, line 101:
```suggestion
            file_count: if "single", allows user to upload one file. If "multiple", user uploads multiple files. If "directory", user uploads all files in selected directory. Return type will be list for each file in case of "multiple" or "directory".
```

gradio/components/multimodaltextbox.py
```diff
@@ -0,0 +1,267 @@
...
+    Submittable,
+    Clickable,
+    Uploadable,
+)
+
+set_documentation_group("component")
+
+
+@document()
+class MultimodalTextbox(
```
abidlabs, line 35: I think the name `MultimodalTextbox` is a little unclear. What about `RichTextbox` as that seems to be used in other places as well: https://www.google.com/search?q=richtextbox&oq=richtextbox&aqs=chrome..69i57j0i67i650j0i20i263i512l2j0i512l3j69i60.1407j0j7&sourceid=chrome&ie=UTF-8

client/python/gradio_client/serializing.py
```diff
@@ -515,6 +515,7 @@ def deserialize(
 
 COMPONENT_MAPPING: dict[str, type] = {
     "textbox": StringSerializable,
+    "multimodaltextbox": StringSerializable,
```
abidlabs, line 518: I don't think its `StringSerialable` is correct. We'll need to create a new class that can take in a dict with string text and dict file, and serialize it properly.

gradio/components/multimodaltextbox.py
```diff
@@ -0,0 +1,267 @@
...
+        text_align: Literal["left", "right"] | None = None,
+        rtl: bool = False,
+        show_copy_button: bool = False,
+        file_count: Literal["single", "multiple", "directory"] = "single",
+        file_types: list[str] | None = None,
+        **kwargs,
+    ):
+        """
+        Parameters:
+            value: default text to provide in textarea. If callable, the function will be called whenever the app loads to set the initial value of the component.
```
abidlabs, line 81: Needs to be updated since value is no longer just text.

abidlabs:
Overall works very nicely and passes files as expected into the function! Some miscellaneous feedback:

1. UI is a bit off compared to the regular `gr.Textbox`. You can see it here: 

image

* Should have rounded borders
* Should have a more defined border
* Should not be "resizable" by dragging the bottom right corner
* The UI when focused is different too, run it yourself with:

```py
import gradio as gr

with gr.Blocks() as demo:
    gr.Textbox()
    gr.MultimodalTextbox()
    gr.Textbox(show_label=False)
    gr.MultimodalTextbox(show_label=False)
    
demo.launch()
``` 

2. Tested most of the parameters, and they all worked with the exception of setting `interactive=False`. This should disable the textbox and hide the upload plus button. 

3. Speaking of the plus button, its alignment is not centered. Its a bit too much to the right. Maybe it should also have a very slight border, similar to the copy button? 

image

abidlabs:
4. I tried a relatively simple demo:

```py
import gradio as gr

with gr.Blocks() as demo:
    m1 = gr.MultimodalTextbox()
    m2 = gr.MultimodalTextbox()
    m1.change(lambda x:x, m1, m2)
    
demo.launch()
```

It worked when I typed the first character, but after that, stopped responding. 

5. Uploading a file also caused the demo to crash, and it looked like this:

image

6. Similarly with a demo like this:

```py
import gradio as gr

def test(x):
    print(x)

with gr.Blocks() as demo:
    m1 = gr.MultimodalTextbox()
    m2 = gr.MultimodalTextbox()
    m1.change(test, m1, None)
    
demo.launch()
```

I kept getting "Connection errored out"

demo/hello_blocks/run.py
dawoodkhan82: Oh yeah commited this by accident.

dawoodkhan82:
@abidlabs Thanks for the review. Will address these bugs now

pngwn:
> I think clicking the file icon should download the files. Right now, there is no way to get the files at all. There should be a separate button to clear the files, but it shouldn't be the file icon itself. I also think it would be nice if the file icon showed the number of files.

Why would you want to download files that you're just uploaded?

I also think the design should be a little better integrated into the textbox. I think the `+` should be inside the textbox (either left or right), the text starts next to it, like discord. We could do it underneath like slack but slack has more options than just upload file, so alongside seems cleaner.

I think the file icons should appear either inside or above the textbox,  I think there should be an icon per file with different icons for each file type (img, music, video, text/doc, everything else) + an `x` or `delete` icon on each one to remove it. Maybe up to a limit of 5-7 files after which there should be an icon with `+7` on it if there are 7 additional attachments.

abidlabs:
> Why would you want to download files that you're just uploaded?

Not if you've just uploaded them, but if this component is used in output/static mode

abidlabs:
Will take another pass when I can. Small nit, but I think it would be great to include a highlight section in the changelog for this PR, similar to https://github.com/gradio-app/gradio/pull/5057

dawoodkhan82:
I think examples may be broken. Will fix

aliabid94:
Don't need two separate files for Static and Interactive if there's just one line difference between them. Also, is the upload button disabled on static mode?

aliabid94:
can you please merge interactive/InteractiveRichTextbox.svelte, shared/RichTextbox.svelte, and static/StaticRichTextbox into a single file @dawoodkhan82? No need for them to be separate and causing lots of code duplication. Will be easier to review once that's all merged.

dawoodkhan82:
> can you please merge interactive/InteractiveRichTextbox.svelte, shared/RichTextbox.svelte, and static/StaticRichTextbox into a single file @dawoodkhan82? No need for them to be separate and causing lots of code duplication. Will be easier to review once that's all merged.

I thought we wanted separate files. As per @pngwn new file structure PR. I followed the same file format as textbox for example.

aliabid94:
oh hmm I missed this refactor, my bad. Lemme take a look at the other files

client/python/gradio_client/serializing.py
```diff
@@ -353,6 +353,113 @@ def deserialize(
             )
 
 
+class RichtextboxSerializable(Serializable):
```
aliabid94, line 356: do we need a new serializeable type? Is JSON serializable not sufficient (its enough for Chatbot, HighlightedText, etc.)

demo/kitchen_sink/run.py
```diff
@@ -98,9 +98,9 @@ def fn(
         gr.Radio(label="Radio", choices=CHOICES, value=CHOICES[2]),
         gr.Dropdown(label="Dropdown", choices=CHOICES),
         gr.Dropdown(label="Multiselect Dropdown (Max choice: 2)", choices=CHOICES, multiselect=True, max_choices=2),
-        gr.Image(label="Image"),
+        gr.Image(label="Image", tool="sketch", brush_radius=20),
```
aliabid94, line 101: ?

gradio/components/rich_textbox.py
```diff
@@ -0,0 +1,295 @@
...
+
+from typing import Callable, Literal, Any
+import warnings
+import tempfile
+from pathlib import Path
+
+from gradio import utils
+from gradio_client import utils as client_utils
+from gradio_client.documentation import document, set_documentation_group
+from gradio_client.serializing import StringSerializable
```
aliabid94, line 13: Two different types of Serializable? Why?

aliabid94 (commented):
Running chatbot_multimodal/run.py I get `TypeError: 'NoneType' object is not subscriptable` when I upload a file.

aliabid94:
> Running chatbot_multimodal/run.py I get TypeError: 'NoneType' object is not subscriptable when I upload a file.

No longer getting this on git pull

gradio/components/rich_textbox.py
```diff
@@ -0,0 +1,295 @@
...
+            "__type__": "update",
+        }
+
+    def preprocess(
+        self, x: dict[str, str | list[dict[str, Any]]] | None
+    ) -> (
+        dict[
+            str,
+            str
+            | bytes
```
freddyaboulton, line 213: Why not just always return a list even if it has one element? Would make the types nicer and there's no need to modify the event handler function in your demo if you switch between single and multiple upload

client/python/gradio_client/serializing.py
abidlabs: Yeah we do, `JSONSerializable` can't handle files which have this ugly dictionary structure. This will get removed in 4.0 so I wouldn't worry too much about it

abidlabs: That being said, tiny nit here regarding capitalization:

`RichtextboxSerializable` should be `RichTextboxSerializable`

@dawoodkhan82

js/richtextbox/shared/RichTextbox.svelte
```diff
@@ -0,0 +1,388 @@
...
+	}
+
+	$: value, el && lines !== max_lines && resize({ target: el });
+
+	const dispatch = createEventDispatcher<{
+		change: undefined;
+		submit: undefined;
+		blur: undefined;
+		select: SelectData;
+		input: undefined;
```
pngwn, line 54: These should also be typed as `never`:
```suggestion
		change: never;
		submit: never;
		blur: never;
		select: SelectData;
		input: never;
```

pngwn (commented):
Looks good, few small things. Seems we have some linting issues there too, you should be able to autofix most in vscode. `shift` + `cmd` + `p` -> `Eslint: Fix all autofixable problems`

dawoodkhan82:
@abidlabs any idea why I'm getting this import error `ImportError: cannot import name 'RichtextboxSerializable' from 'gradio_client.serializing' (/home/runner/work/gradio/gradio/venv/lib/python3.8/site-packages/gradio_client/serializing.py)`

abidlabs:
Hmm are you getting it locally or on CI?

dawoodkhan82:
> Hmm are you getting it locally or on CI?
Not seeing it locally

abidlabs:
Ok it could be that in the CI we are installing `gradio_client` from `pypi` instead of installing from local source files. Let me check

abidlabs:
Ok I think what's happening is that the GitHub action is not re-installing the client library because its hitting a cache in this step:

https://github.com/gradio-app/gradio/actions/runs/5813116844/workflow?pr=5032#L173

I'll modify the workflow so that it installs the client library regardless of the cache. Give me a minute and I'll try pushing to this branch

abidlabs:
Now its installing the `gradio_client` locally @dawoodkhan82!

abidlabs:
@dawoodkhan82 the reason the client test is failing is because its using the version of `gradio` from pypi. I'll change it to use the version from this branch.

Now that we release the client and main libraries together (thanks to the changesets PR @pngwn put together), this should be safe to do. However, one thing we should add to the changesets PR is pin a specific version of `gradio_client` in `gradio`'s dependencies (let's discuss this separately)

abidlabs:
Sadly this will slow down the client tests a bit because installing `gradio` locally includes building the frontend as well, but I think its a necessary step to ensure that the client and main library are always in sync and to avoid cases like this^

cc @freddyaboulton

abidlabs:
Seeing some small issues left in the `RichTextbox`:

1. There's a subtle shadow visible in light mode (compare with regular textbox) for example:

image


Code to reproduce:

```py
import gradio as gr

with gr.Blocks() as demo:
    r = gr.RichTextbox()
    t = gr.Textbox()
        
demo.launch()
```

Also
* By default, the `RichTextbox` is non-interactive -- not sure why this is the case, it should be interactive like our other components
* You can also see that it's taller than our regular `Textbox` -- can we make them the same size?

2. The preprocessing/postprocessing isn't working like I expected it to. Try either of these demos:

```py
import gradio as gr

with gr.Blocks() as demo:
    r = gr.RichTextbox(interactive=True)
    r2 = gr.RichTextbox()
    t = gr.Textbox()
    
    r.change(lambda x: x, r, r2)
    
demo.launch()
```

or 

```py
import gradio as gr

with gr.Blocks() as demo:
    r = gr.RichTextbox(interactive=True)
    t = gr.Textbox()
    
    r.change(lambda x: x["text"], r, t)
    
demo.launch()
```

4. The `.change()` trigger seems to trigger indefinitely (you'll see infinite errors happening if you run the code above)

abidlabs:
> The CI should be good now I believe.

spoke too soon 😅

abidlabs:
@dawoodkhan82 CI ran successfully but it looks like the functional test based on the multimodal chatbot needs to be updated since we no longer use the the `UploadButton`

.github/workflows/backend.yml
```diff
@@ -73,41 +70,65 @@ jobs:
         path: |
           client/python/venv/*
         key: python-client-${{ runner.os }}-pip-${{ hashFiles('client/python/requirements.txt') }}-${{ hashFiles('client/python/test/requirements.txt') }}
+    - uses: actions/cache@v3
```
freddyaboulton, line 73: How do the backend tests depend on the front-end being built?

abidlabs: If you install gradio from source and you try to run a simple gradio demo, you will get errors if the frontend is not built

abidlabs:
Reversed the order of installs, thanks @pngwn for catching that

freddyaboulton:
> Sadly this will slow down the client tests a bit because installing gradio locally includes building the frontend as well, but I think its a necessary step to ensure that the client and main library are always in sync and to avoid cases like this^

You should get a cache hit for a PR that doesn't modify the front-end files. This PR modifies all points of the repo which is why every run has to build everything (if I understand correctly).

@abidlabs Should we pull out the CI changes into a separate PR given that we're gonna make significant changes to this PR?

abidlabs:
> @abidlabs Should we pull out the CI changes into a separate PR given that we're gonna make significant changes to this PR?

Yes sorry should have mentioned here, but I pulled out the CI changes into the the changesets release PR: https://github.com/gradio-app/gradio/pull/5038

I'll remove them from this PR

dwipper:
Hi @abidlabs. Is it possible to get a build with this component loaded in our environment? We have a key business case where we need to prototype its use. Thanks!

freddyaboulton:
Hi @dwipper - I'm going to try to make this into a custom component that you can use when gradio 4.0 launches next week!

dwipper:
Hi @freddyaboulton , Thanks! That would awesome!

dwipper:
@freddyaboulton I saw that @abidlabs published this demo of the component: https://huggingface.co/spaces/abidlabs/gradio_rich_textbox . Is there a list somewhere of what other formatting is supported other than [b] and [u]? i.e. is there a strike-through? How to do colors? Since there isn't a toolbar, I'm wondering how a user would actually enter formatted text into the field?

freddyaboulton:
There are two separate things -  the original PR implements a textbox that comes with an upload button. That's implemented as a custom component here: https://huggingface.co/spaces/freddyaboulton/gradio_textwithattachments

@abidlabs 's component, lets you style text using bbcode syntax: https://en.wikipedia.org/wiki/BBCode. For colors you can do [color=green]Hi![/color]. Strikethrough is [s]Word[/s]

richtext_box

PR #5035, opened 2023-07-28
JS Client: Fixes cannot read properties of null (reading 'is_file')

raymondtri:
## Description

In transform_output, typeof d can be an object even if the property is_file doesn't exist on d. Therefore attempting to access d.is_file will error out. This commit fixes that by failing quietly if the is_file property doesn't exist on d.

Closes: (no issue, the fix is very small)

## 🎯 PRs Should Target Issues

> Before your create a PR, please check to see if there is [an existing issue](https://github.com/gradio-app/gradio/issues) for this change. If not, please create an issue before you create this PR, unless the fix is very small. 

The fix is very small.

### DIFF
--- a/client/js/src/client.ts
+++ b/client/js/src/client.ts
@@ -395,7 +395,8 @@ export function api_factory(fetch_implementation: typeof fetch) {
 						});
 
 						post_data(
-							`${http_protocol}//${host + config.path}/run${_endpoint.startsWith("/") ? _endpoint : `/${_endpoint}`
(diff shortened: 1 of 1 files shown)


abidlabs:
Hi @raymondtri thanks for the contribution. Can you give a quick code example of where you're seeing this issue?

abidlabs:
Hi @raymondtri just to follow up, can you share an example of what issue this PR is fixing?

Issue #4555, opened 2023-06-16
Dropdown is cleared on focus


missionfloyd:
### Describe the bug

When clicking or otherwise focusing a dropdown with `allow_custom_value=True`, the current value is cleared and is not restored when focus is lost.

Only seems to affect the frontend, the value is not updated unless another is selected or entered. Leaving it blank causes it to return the previous value.

Ideally, the value shouldn't be cleared at all (with `allow_custom_value=True` anyway). We're unable to edit the value without having to enter it again.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```python
import gradio as gr

demo = gr.Interface(lambda x: x, inputs=gr.Dropdown(choices=["A", "B", "C"], allow_custom_value=True), outputs="text")
demo.launch()
```

### Logs

```shell
n/a
```

### System Info

```shell
Gradio 3.35.2
Python 3.10.11
Windows 10
Edge 114.0.1823.51
```

### Severity

annoying

Muhtasham:
Also Dropdown with long strings breaks down and doesn't let choose other values

abidlabs:
Hi @Muhtasham can you create a new issue if your issue is not already covered by #4503 or #4292

Issue #4503, opened 2023-06-13
[`gr.Dropdown`] Options with whitespace in name leads to invalid query selector 


xenova:
### Describe the bug

When testing out this HF space: https://huggingface.co/spaces/facebook/MMS, I selected a language from the dropdown, and when attempting to select another language, it throws an error: `Uncaught (in promise) DOMException: Failed to execute 'querySelector' on 'Document': 'li[data-value="abi: Abidji "]' is not a valid selector. (DropdownOptions.svelte:19)`

Similar to https://github.com/gradio-app/gradio/issues/4292

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

1. Go to https://huggingface.co/spaces/facebook/MMS
2. Select a language from the dropdown
3. Select another language from the dropdown. This will throw the error.

### Screenshot

The dropdown becomes empty:
![image](https://github.com/gradio-app/gradio/assets/26504141/567660a3-4855-4d3a-b92a-3367de130ea9)

And this error is logged in the console:
![image](https://github.com/gradio-app/gradio/assets/26504141/f57fab91-75dc-4825-b669-e46f85973798)

### Logs

```shell
Uncaught (in promise) DOMException: Failed to execute 'querySelector' on 'Document': 'li[data-value="abi: Abidji "]' is not a valid selector.
    at Je.t.$$.update (https://facebook-mms.hf.space/assets/index-315bc3fa.js:1:3366)
    at Dr (https://facebook-mms.hf.space/assets/index-607392ea.js:4:1833)
    at lr (https://facebook-mms.hf.space/assets/index-607392ea.js:4:1548)
```

### System Info

```shell
Latest version of chrome, and whichever version of gradio that is shipped with HF spaces.
```

### Severity

serious, but I can work around it

lcrh:
Similar to https://github.com/gradio-app/gradio/issues/4292

pngwn:
The issue here is the whitsepace i think. @radames had a similar issue. This issue is basically the same as #4292 in that there are invalid characters.

We could escape those character but we shouldn't be using `document. querySelector` in a svelte component at all. The fix is to refactor the component not to use that.

cc @hannahblair @dawoodkhan82

Issue #4292, opened 2023-05-20
Doublequotes in Dropdown values break Dropdown menu.


lcrh:
### Describe the bug

I loaded a gradio app I wrote a month ago and noticed that the dropdown menu was non-functional. (It used to work as of mid-April).

In particular, after clicking the down arrow, the UI would become unresponsive and get stuck with the dropdown menu open and unclosable:

![image](https://github.com/gradio-app/gradio/assets/14020730/f654ea96-11ed-4b27-b0cc-298cf6bce5bd)

Looking at the javascript console, I see an error (pasted under "Logs" below).

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Run from a notebook cell.
```
import gradio as gr

with gr.Blocks() as blocks:
  dropdown = gr.Dropdown(
    choices=['', '"test"'],
    value='')
blocks.launch(share=True)
```

To reproduce:
1. Click dropdown menu to reveal choices
2. Select "test"
3. Click dropdown menu again to reveal choices

This manifests slightly different than the issue I described above (the menu closes), but shows the same JavaScript error:
```
Uncaught (in promise) DOMException: Failed to execute 'querySelector' on 'Document': 'li[data-value=""test""]' is not a valid selector.
```

### Screenshot

![image](https://github.com/gradio-app/gradio/assets/14020730/f654ea96-11ed-4b27-b0cc-298cf6bce5bd)

### Logs

```shell
DropdownOptions.svelte:19 Uncaught (in promise) DOMException: Failed to execute 'querySelector' on 'Document': 'li[data-value="22da73fd-0f88-51bf-b5f6-462c0fc0782b / public / "dog""]' is not a valid selector.
    at Je.t.$$.update (https://gradio.s3-us-west-2.amazonaws.com/3.31.0/assets/index-10df479e.js:1:3366)
    at Dr (https://gradio.s3-us-west-2.amazonaws.com/3.31.0/assets/index-20f57901.js:4:1833)
    at lr (https://gradio.s3-us-west-2.amazonaws.com/3.31.0/assets/index-20f57901.js:4:1548)
Je.t.$$.update	@	DropdownOptions.svelte:19
Dr	@	index.mjs:1269
lr	@	index.mjs:1233
Promise.then (async)		
sr	@	index.mjs:1184
Wr	@	index.mjs:2066
(anonymous)	@	index.mjs:2104
ve	@	Dropdown.svelte:203
```
```


### System Info

```shell
[...]

PR #5039, opened 2023-07-28
`gr.Dropdown()` now supports values with arbitrary characters and doesn't clear value when re-focused

abidlabs:
Fixes two separate issues in `gr.Dropdown()`:

**1. The value of the `gr.Dropdown()` was getting cleared when clicked.** 

The reason that we were doing this was because we wanted users to be able to easily see the full list of choices when they focused on a `gr.Dropdown` and had already previously selected a choice. Ordinarily, the list of choices would get filtered by the value of the `gr.Dropdown()` so you needed to clear the value to see the full list. However, it was a little weird to see the value of the `gr.Dropdown()` clearing on focus and was especially problematic if `allow_custom_value` was set to `True` as it made the value not really editable after you had left the focus on the `gr.Dropdown()`. 

Instead, I tweaked the behavior of the `gr.Dropdown()` so that the full list of choices is always shown immediately on focus without needing to clear the value of the `gr.Dropdown()`. Once you start typing, the list of choices gets filtered again.

Test with:

```py
import gradio as gr

with gr.Blocks() as demo:
    gr.Dropdown(choices=["America", "Azerbaijan", "Bahrain"], allow_custom_value=True)    
    gr.Dropdown(choices=["America", "Azerbaijan", "Bahrain"], allow_custom_value=False)
    
demo.launch()
```

 
Closes: #4555. 

**2. The `gr.Dropdown()` was not working properly if values with whitespaces and other special characters were used.**

Because we were using 

```js
document.querySelector(`li[data-value="${value}"]`
```

we couldn't handle certain characters. Now we avoid using `document.querySelector()`. Test with:

```py
import gradio as gr

with gr.Blocks() as demo:
    gr.Dropdown(choices=["I love you", "Can't believe it", "Here's a \" for you", "Normal"])
    
demo.launch()
```

Closes: #4503, Closes: #4292

Other small changes:
* I reordered the parameters to make `allow_custom_value` more discoverable in the [docs as discussed internally](https://huggingface.slack.com/archives/C02QZLG8GMN/p1690426170569499?thread_ts=1690369398.426209&cid=C02QZLG8GMN)
* Fixed some small linting issues

Kudos to the function tests which caught a subtle bug in my original implementation!

cc @abhishekkrthakur @yvrjsharma


tsconfig.json
```diff
@@ -2,7 +2,7 @@
 	"compilerOptions": {
 		"moduleResolution": "node",
 		"module": "es2020",
-		"lib": ["es2020", "DOM"],
+		"lib": ["es2020", "DOM", "dom.iterable"],
```
abidlabs, line 5: This is needed so that TS doesn't complain about this line:
```
				for (const element of elements) {
```

abidlabs:
Thank you for the review @hannahblair!

abhishekkrthakur:
will this be released soon?

abidlabs:
Probably later this week. But in the meantime, you can install using the link above: pip install https://gradio-builds.s3.amazonaws.com/2b70de43cc1731a843392fbbbddff97ef5e10d18/gradio-3.39.0-py3-none-any.whl

Issue #4737, opened 2023-06-29
Add `step` to `gr.Number`


apolinario:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
`gr.Number()` doesn't have a step param 

**Describe the solution you'd like**  
I would like a 'step' param to `gr.Number` given that it has the HTML arrows to change the step
<img width="113" alt="image" src="https://github.com/gradio-app/gradio/assets/788417/f3304a4c-4a1c-4c4d-9fec-653f94980f36">

PR #5047, opened 2023-07-31
Add `step` param to `Number`

hannahblair:
## Description

Adds a `step` param to the `Number` component so we can define the interval at which the input value increases or decreases with the HTML up/down arrows.

Demo: 

```
import gradio as gr

with gr.Blocks() as demo:
    gr.Number(maximum=50, label="Number", steps=10)

demo.launch()
```

Closes: #4737 

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/curly-nights-yawn.md
+++ b/.changeset/curly-nights-yawn.md
@@ -1,0 +1,7 @@
+---
+"@gradio/app": minor
+"@gradio/gallery": minor
+"gradio": minor
+---
+
+feat:Add download button to selected images in `Gallery`
--- a/.changeset/legal-beds-beam.md
+++ b/.changeset/legal-beds-beam.md
@@ -1,0 +1,7 @@
+---
+"@gradio/app": minor
+"@gradio/highlighted-text": minor
+"gradio": minor
+---
+
+feat:Allow new lines in `HighlightedText` with `/n` and preserve whitespace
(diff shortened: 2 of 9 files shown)


gradio/components/number.py
```diff
@@ -79,10 +80,12 @@ def __init__(
             precision: Precision to round input/output to. If set to 0, will round to nearest integer and convert type to int. If None, no rounding happens.
             minimum: Minimum value. Only applied when component is used as an input. If a user provides a smaller value, a gr.Error exception is raised by the backend.
             maximum: Maximum value. Only applied when component is used as an input. If a user provides a larger value, a gr.Error exception is raised by the backend.
+            step: The interval between numbers in the input field. If None, a default of 1 is set.
```
abidlabs, line 83: Nit: updating the wording a little bit based on the change above.
```suggestion
            step: The interval between allowed numbers in the component. Can be used along with `minimum` and `maximum` to create a range of legal values.
```

gradio/components/number.py
```diff
@@ -127,6 +130,7 @@ def get_config(self):
             "value": self.value,
             "minimum": self.minimum,
             "maximum": self.maximum,
+            "step": self.step,
```
abidlabs, line 133: We should add this parameter to the `Number.update()` method as well.

abidlabs (approved):

PR #5048, opened 2023-07-31
Use `importlib` in favor of deprecated `pkg_resources`

jayceslesar:
## Description

Bad practice to be using just `import pkg_resources` instead of `import importlib.resources as pkg_resources`, as this led to my org having trouble installing and explicitly needing to add setuptools as a dependency which doesnt quite make sense. This is a drop in replacement and has existed since python 3.7. 

The issue arose from having `gradio` installed in an application my org is developing, and running it resulted in 

```
gradio/flagging.py", line 17, in <module>
    import pkg_resources
ModuleNotFoundError: No module named 'pkg_resources'
```

This is a drop in replacement.

Closes: #(issue)
N/A

## 🎯 PRs Should Target Issues

## Tests


jayceslesar:
@abidlabs going to be a little more involved, I will patch the rest of `pkg_resources` stuff

jayceslesar:
@abidlabs should be good now!

abidlabs:
Thanks @jayceslesar! It turns out that we didn't even need that version check in `flagging.py` so I've removed it and made a few other tweaks. Will merge in once CI passes. Thanks again

jayceslesar:
> 

Absolutely! Ill see what I can do about getting types in a good spot to get CI to pass

gradio/routes.py
```diff
@@ -4,6 +4,11 @@
 from __future__ import annotations
 
 import asyncio
+import sys
+if sys.version_info >= (3, 9):
+    from importlib.resources import files
+else:
+    from importlib_resources import files
```
abidlabs, line 11: Hmm we'll need to add `importlib_resources` as a dependency if its not available in python 3.8

jayceslesar: good catch

jayceslesar:
@abidlabs thanks for the cleanup!

abidlabs:
No worries! I'll fix the final typing issues and then merge once CI passes

abidlabs:
Ok I think we should be good to go. I've just asked @yvrjsharma to test on a Windows machine as well to make sure all the paths resolve correctly

yvrjsharma:
I have just run couple of demos locally on my Windows machine, and all work fine.

abidlabs:
Thanks @jayceslesar and @yvrjsharma will merge in after CI passes

PR #5057, opened 2023-08-01
Return final output for generators in Client.predict

### LINKED ISSUE #4887: Provide a better code snippet for generators in the "view API" page
Say you have a Gradio app like this:

```py
import gradio as gr

def test(x):
    for i in range(len(x)):
        yield x[:i]

gr.Interface(test, "textbox", "textbox").queue().launch()
```

The view API page is still going to suggest that you use the `.predict()` function to call this method via an API:

image

However, this only gives you the first iteration, which is not usually what you want. Instead, it would be better to show the `.submit()` usage here. 

Inspired by this question: https://discuss.huggingface.co/t/gradio-api-streaming/45395

### LINKED ISSUE #4888: Install `gradio_client` locally in CI
I think we should install `gradio_client` in editable mode in CI, so the local version of `gradio_client` is used by `gradio` for Ci checks.

Currently `gradio` relies on the version of `gradio_client` available on pypi, which is correct. However, if we make changes to `gradio_client` that impact `gradio` CI checks can fail giving as those changes won't be used by the gradio version used in CI.

Currently this is annoying but understandable because the release processes for `gradio` and `gradio_client` are independent, we can release changes to one without the other even if said changes exist on main. When we move to a unified release process for the whole repository (#4851) any changes in gradio + gradio client will be released together (along with any other changes on main). This will mean that they can't get out of sync and an editable install of all python packages is safe (and desired because tests should reflect what will happen when the libraries are published as closely as possible).

The way the release process will work is that we look for dependency changes, if `gradio_client` is bumped and `gradio` depends on it, then the version of `gradio_client` in the requirements file will be bumped (if the current version doesn't satisfy the semver range of the dependency in the requirements file). This is also how JS dependencies will work in the future (if we have libraries that depend on libraries, which we will shortly).

Would love to get people's thoughts on this, and any context for the original decision to _not_ install the client in editable mode.

cc @freddyaboulton @abidlabs

freddyaboulton:
## Description

Closes #4887
Closes #4888

Adds a `finish` method to the client jobs to make it easier to wait for job completion.

Also installs the client locally for CI as discussed in #4888 

### Python

image

### JS
image

## 🎯 PRs Should Target Issues

## Tests


.github/workflows/backend.yml
```diff
@@ -170,7 +170,7 @@ jobs:
       if: steps.cache.outputs.cache-hit != 'true' && runner.os == 'Linux'
       run: |
         . venv/bin/activate
-        python -m pip install -e . -r test/requirements.txt
```
freddyaboulton, line 173: For #4888

abidlabs: Nice

client/js/src/client.ts
```diff
@@ -589,7 +590,23 @@ export function api_factory(fetch_implementation: typeof fetch) {
 					narrowed_listener_map[eventType] = listeners;
 					listeners?.push(listener);
 
-					return { on, off, cancel, destroy };
+					return { on, off, cancel, destroy, finish };
+				}
+
+				async function finish() {
```
abidlabs, line 596: Nice

abidlabs:
Looks good @freddyaboulton! I tested the Python client and works great. Trying to figure out how to test the JS client right now.

A thought in the meantime -- I know we discussed this before and I was initially against it -- but should we just have the `.result()` method actually return the final result from `.outputs()` instead of the first result? 🙈 

Pros:

1. People only need to remember one method and we don't need to have different code snippets
2. Its much simpler syntax than `finish()` and indexing the final result from `outputs()`
3. Its what people expect -- we've had a number of people asking why they are only getting the first response from generative outputs

Cons:

1. It would be a breaking change (but we are pre-release anyways)
2. We would have to handle `every` separately (but these endpoints are pretty rare anyways). Maybe we don't even expose endpoints that are `every` via API

freddyaboulton:
I agree it's what people expect! Ok let's do that.

freddyaboulton:
> Shouldn't it return all of the results instead of just the last one? I think that is more likely to be what people expect.

I think people using the clients don't know about whether the endpoint streams results and they just want to get the final result from the chatbot/diffusion process. I've seen this come up in discord/internal threads numerous times. If a user wants all the outputs, they have ways to do that with `submit` already.

> However, i feel very strongly that it should throw an exception if if it is an every/ continuous function.

Yea agreed. What I plan on doing in the python client is saying that `every` functions are "invalid" so they will not show up in `view_api`/error if user tries to call them. Will do something similar for js.

pngwn:
Oh, they should show up. We have the right APIs to support them. We just shouldn't allow a user to call predict on them. They should also show up in view_api imo.

freddyaboulton:
Ok I made the changes so that `predict` always returns the final output from a generator and `predict` will raise an error if you try to call it on a continuous function.

The only downside is that the code snippet in the View API page will show `predict` for `generators` which will error immediately. Given that this is probably a very small use case and the error is informative, I think that will be ok.

freddyaboulton:
Should be good for another look though!

.changeset/two-spies-shine.md
```diff
@@ -0,0 +1,28 @@
+---
+"@gradio/app": minor
+"gradio": minor
+"gradio_client": minor
+---
+
+highlight:
+
+#### Better Code Snippet for Streaming API Routes
```
abidlabs, line 9: Needs to be updated

client/js/src/client.ts
```diff
@@ -53,6 +53,7 @@ type SubmitReturn = {
 	off: event;
 	cancel: () => Promise<void>;
 	destroy: () => void;
+	finish: () => Promise<unknown>;
```
abidlabs, line 56: Do we still need to add this `finish()` method?

abidlabs:
I was able to run the Python client and confirm that it works. However, I was getting an error when trying to use the JS client. Not sure if I'm testing correctly. Here's what I did:

0. Start `python demo/chatinterface_streaming_echo/run.py` running on port 7860
1. Navigate to an empty directory
2. Run `npm init`  and accept the defaults
3. Run `npm i -D @gradio/client` (probably not necessary)
4. Then copy paste into `index.mjs`:

```js
import { client } from "../../dev/gradio-repos/gradio/client/js/dist/index.js";

const app = await client("http://127.0.0.1:7860/");
const result = await app.predict("/chat", [		
				"Howdy!", // string  in 'Message' Textbox component
	]);

console.log(result.data);
```

Then, I ran the file with `node index.mjs`. It ran but threw this error:

```js
node:internal/process/esm_loader:97
    internalBinding('errors').triggerUncaughtException(
                              ^
{
  type: 'status',
  endpoint: '/chat',
  fn_index: 24,
  time: 2023-08-07T12:38:21.122Z,
  queue: true,
  message: null,
  stage: 'error',
  code: undefined,
  success: false
}
```


cc @pngwn @hannahblair @dawoodkhan82 maybe y'all can try out the js client?

freddyaboulton:
@abidlabs - I think the js client is not handling state for the user like the python client does.

This is what I get from the server

```bash
ValueError: An event handler (_api_stream_fn) didn't receive enough input values (needed: 2, got: 1).
Check if the event handler calls a Javascript function, and make sure its return value is correct.
Wanted inputs:
    [textbox, state]
Received inputs:
    ["Hello"]
```

Can you try with another demo that doesn't use state, e.g. `count_generator`. I can look into the state issue in another PR.

abidlabs:
Thanks @freddyaboulton! I still get the same error, but I'm also seeing this on `main` so clearly there's something wrong with my setup. I'll look into what I'm doing wrong, but perhaps we can have someone else review the js client to unblock.

client/js/src/client.ts
```diff
@@ -341,24 +341,39 @@ export function api_factory(fetch_implementation: typeof fetch): Client {
...
 						})
 						.on("status", (status) => {
 							if (status.stage === "error") rej(status);
-							if (status.stage === "complete" && data_returned) {
-								app.destroy();
-							}
 							if (status.stage === "complete") {
 								status_complete = true;
+								app.destroy();
+								res(result);
```
pngwn, line 376: Whats this stuff?

pngwn: oh of course, my bad.

The only problem is it was implemented this way because i was having an issue receiving message out of order, which isn't technically supposed to be possible but it still happened. So the original logic allowed either message to resolve the promise whereas this only allows the status function to finish it, which if it is processed _before_ the final data message means that the return value of predict would be the second to last value.

Maybe we could use the `data.completed` property and add the dual handling back?

freddyaboulton: Ok I added the dual handling back

pngwn (approved):
Looks good to me! Thanks @freddyaboulton!

freddyaboulton:
Thank you @pngwn !!

PR #5058, opened 2023-08-01
Dont format on save for vscode

freddyaboulton:
## Description
Whenever I update a `.svelte`, `.json` or `.ts` file, vs code will do some autoformatting that introduces a lot of additional changes irrelevant to my PR and doesn't conform to our standard lint commands anyways.

So I'm deleting it so that global user preferences takeover.

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -6,7 +6,6 @@
 	"cssvar.extensions": ["js", "css", "html", "jsx", "tsx", "svelte"],
 	"python.analysis.extraPaths": ["./gradio/themes/utils"],
 	"prettier.useTabs": true,
-	"editor.formatOnSave": false,
 	"svelte.plugin.svelte.format.enable": true,
 	"svelte.plugin.svelte.diagnostics.enable": false,
 	"prettier.configPath": ".config/.prettierrc.json",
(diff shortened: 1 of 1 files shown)


freddyaboulton:
Stuff like this - for example, editing a single line in a github workflow file will cause all lines to indent which then causes the prettier check to fail.

every_line_indent

image

pngwn:
Ah, maybe we should just enable this for specific languages:

```
"[javascript][typescript][svelte][json][markdown]: {
    "editor.formatOnSave": true
}
```

pngwn:
Sorry, change my mind. We should leave this decision up to the developer, they can set this themselves in their editor settings.

.vscode/settings.json
```diff
@@ -6,7 +6,6 @@
 	"cssvar.extensions": ["js", "css", "html", "jsx", "tsx", "svelte"],
 	"python.analysis.extraPaths": ["./gradio/themes/utils"],
 	"prettier.useTabs": true,
```
pngwn, line 8: Could you remove this as well? The prettier config should be the source of truth.

pngwn (approved):
Looks good, there is one other thing that should be removed as well i think. Thanks for fixing this @freddyaboulton!

PR #5061, opened 2023-08-01
Ensure `gradio_client` is backwards compatible with `gradio==3.24.1`

abidlabs:
[As discussed internally](https://huggingface.slack.com/archives/C02QZLG8GMN/p1690856034533989), some of the changes to the client were not fully backwards compatible, which broke older versions of `gradio` if they were using later versions of the client. This fixes that, at least for `gradio==3.24.1` (I have not tested other versions of `gradio`). 

To test:

1. Install `gradio==3.24.1`
2. Install the latest version of `gradio_client` from pypi
3. Run any Gradio app (it will not work)
4. Install `gradio_client` from this branch
5. It should work

### DIFF
--- a/.changeset/eight-humans-sniff.md
+++ b/.changeset/eight-humans-sniff.md
@@ -1,0 +1,6 @@
+---
+"@gradio/form": patch
+"gradio": patch
+---
+
+fix:`gr.Dropdown` now has correct behavior in static mode as well as when an option is selected
--- a/.changeset/silver-signs-speak.md
+++ b/.changeset/silver-signs-speak.md
@@ -1,0 +1,7 @@
+---
+"@gradio/app": minor
+"@gradio/form": minor
+"gradio": minor
+---
+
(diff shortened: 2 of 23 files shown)


freddyaboulton:
I think we're going to do a hard pin of gradio_client in gradio requirements instead

abidlabs:
We should do both @freddyaboulton -- make this fix to address older versions of `gradio` while also pinning `gradio_client` going forward (I'll add the pin to this PR)

freddyaboulton (approved):
Thanks @abidlabs !

abidlabs:
Appreciate the review @freddyaboulton

pngwn:
The hard pin needs to be updated in CI because the requirements get updated every time we make a release. i'll find that and take care of it.

freddyaboulton:
Thanks @pngwn !

pngwn:
https://github.com/gradio-app/gradio/pull/5222

Issue #4280, opened 2023-05-19
Impossible to add values to dropdown inside gr.block when initialized


tr31zh:
### Describe the bug

Cannot add values to Dropdown when following conditions:
- Dropdown is in side gr.Block
- Dropdown is multiselect
- Dropdown has been created with selected value(s)

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

import gradio as gr

with gr.Blocks() as gr_training_zoo:
    tst = gr.Dropdown(
        ["ran", "swam", "ate", "slept"],
        value=[],
        multiselect=True,
        label="Not initialized",
        info="Changes allowed",
    )
    gr.Dropdown(
        ["ran", "swam", "ate", "slept"],
        value="swam",
        multiselect=True,
        label="Initialized",
        info="Only value removal allowed, no adding",
    )

gr_training_zoo.launch()

### Logs

```shell
No logs
```

### System Info

```shell
gradio-3.31.0 gradio-client-0.2.5
chrome Version 113.0.5672.92
python 3.10.11
mac os 13.0 (22A380)
```

### Severity

serious, but I can work around it

abidlabs:
Hi @tr31zh interesting... you can fix this by adding `interactive=True` to the `gr.Dropdown()`, but I'm not sure why the error happens in the first place. Will investigate.

Here's the workaround for now:

```py
import gradio as gr

with gr.Blocks() as gr_training_zoo:
    tst = gr.Dropdown(
        ["ran", "swam", "ate", "slept"],
        value=[],
        multiselect=True,
        label="Not initialized",
        info="Changes allowed",
        interactive=True
    )
    gr.Dropdown(
        ["ran", "swam", "ate", "slept"],
        value="swam",
        multiselect=True,
        label="Initialized",
        info="Only value removal allowed, no adding",
        interactive=True
    )

gr_training_zoo.launch()
```

Issue #4269, opened 2023-05-18
Dropdown pops up if focus is lost during change.


yownas:
### Describe the bug

If you switch focus to another window while gradio.dropdown.change() is executing, when you return focus to the browser again, the dropdown will  pop up for a split second and sometimes trick you into making another selection.

Keeping the focus on the browser while it is working/loading doesn't seem to trigger the issue.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

I wrote this reproducer. Make a selection and then change focus to another window while it is "working", if you then click the textbox (or any area that would be taken up by the dropdown), the dropdown pops up and tricks you into making another selection.

```
import gradio
import time

the_items = [1,2,3,4,5,6,7,8,9,10]

def make_selection(selection):
    time.sleep(5)
    return(selection)

with gradio.Blocks(analytics_enabled=False) as ui:
    with gradio.Row():
        dropdown = gradio.Dropdown(label="Make selection and change focus to another window before it is done", choices=the_items)
    with gradio.Row():
        textbox = gradio.Textbox(label="Click in textbox after selection has been made")

    dropdown.change(fn=make_selection, inputs=dropdown, outputs=textbox)

ui.launch(debug=True, server_name="0.0.0.0")
```

### Logs

```shell
No output.
```

### System Info

```shell
Gradio versions tested: 3.21.0, 3.27.0, 3.29.0, 3.31.0
Ubuntu 22.04
Browser: Firefox and Brave tested
```

### Severity

annoying

PR #5062, opened 2023-08-01
`gr.Dropdown` now has correct behavior in static mode as well as when an option is selected

abidlabs:
2 more fixes to `gr.Dropdown()`:

1. In static mode, the `gr.Dropdown(multiselect=True)` still had the X icons, and clicking on an option would remove it. This behavior has been disabled in static mode. Test with:

```py
import gradio as gr

with gr.Blocks() as gr_training_zoo:
    d2 = gr.Dropdown(
        ["ran", "swam", "ate", "slept"],
        value=["swam"],
        multiselect=True,
        label="Initialized",
        info="Only value removal allowed, no adding",
    )

gr_training_zoo.launch()
``` 

Also we had never really documented how Gradio determines the interactivity of a component if `interactive` was not specified and the component was neither an input nor an output. I've added that to the relevant Guide. Together, this closes: #4280. 

2. Clicking on a dropdown option `gr.Dropdown(multiselect=False)` would select it, yet it would keep the focus on the Dropdown and in particular would keep the cursor typing. This led to an interesting bug described in #4269. I've changed the behavior so that clicking on a choice actually unfocuses from the dropdown. This is a more satisfying UX, more similar to the native select element's UX, and closes: #4269.

Also added a few `gr.Dropdown` stories to prevent regression of the former issue.

### DIFF
--- a/js/form/src/Dropdown.svelte
+++ b/js/form/src/Dropdown.svelte
@@ -72,8 +72,11 @@
 	}
 
 	function remove(option: string): void {
-		value = value as string[];
-		value = value.filter((v: string) => v !== option);
+		if (!disabled)
+		{
+			value = value as string[];
+			value = value.filter((v: string) => v !== option);
+		}
 		dispatch("select", {
 			index: choices.indexOf(option),
 			value: option,
@@ -87,7 +90,7 @@
 		e.preventDefault();
 	}
 
-	function handle_blur(e: FocusEvent) {
+	function handle_blur(e: FocusEvent): void {
 		if (multiselect) {
 			inputValue = "";
 		} else if (!allow_custom_value) {
@@ -104,14 +107,10 @@
 		dispatch("blur");
 	}
 
-	function handle_focus(e: FocusEvent){
+	function handle_focus(e: FocusEvent): void{
 		dispatch("focus");
-		showOptions = !showOptions;
-		if (showOptions) {
-			filtered = choices;
-		} else {
-			filterInput.blur();
-		}
+		showOptions = true;
+		filtered = choices;
 	}
 
 	function handleOptionMousedown(e: any): void {
@@ -137,6 +136,7 @@
 					value: option,
 					selected: true
 				});
+				filterInput.blur();
 			}
 		}
 	}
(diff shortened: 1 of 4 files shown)


js/form/src/Dropdown.svelte
```diff
@@ -104,14 +107,10 @@
...
 
-	function handle_focus(e: FocusEvent){
+	function handle_focus(e: FocusEvent): void{
 		dispatch("focus");
-		showOptions = !showOptions;
-		if (showOptions) {
-			filtered = choices;
-		} else {
-			filterInput.blur();
-		}
```
abidlabs, line 114: Unless I'm missing something, these lines can never be reached  @dawoodkhan82?

dawoodkhan82: I think your cleanup of this makes sense. But I think these lines should've been hit. How else was the input being blurred? But I also didn't go back and test on main, so not sure.

dawoodkhan82 (approved):
Looks good. I also tested `kitchen-sink` and the dropdown works fine. One nit for the `static` dropdown would be to change the cursor from pointer to default when hovering over the token.

abidlabs:

> One nit for the static dropdown would be to change the cursor from pointer to default when hovering over the token.

I'll fix that!

> But I think these lines should've been hit. How else was the input being blurred? 

What do you mean? Why should the input be blurred in the focus method? Maybe I'm missing something

This was never hit because `showOptions` would always be false before you focused. In the focus method, we invert it, so it would become true, so this logic was never hit.

dawoodkhan82:
> This was never hit because showOptions would always be false before you focused. In the focus method, we invert it, so it would become true, so this logic was never hit.

Ah yes I see what you mean. Yes we're good to remove it then!

abidlabs:
Sounds good, thanks for the confirmation.

> One nit for the static dropdown would be to change the cursor from pointer to default when hovering over the token.

I was rethinking this, and I think its better to leave as is because clicking while disabled still triggers the `.select()` event trigger (similar to the other components that support `.select()`)

Issue #4774, opened 2023-07-03
Design: Configs per session, and removing duplicated functions in a Component


aliabid94:
Two issues to be resolved with the same approach:

1) Right now, a Gradio Blocks app has a single config. We need there to be one config *per user session*, so that we can support:
  a. Backend parameter updates (e.g. type, Dropdown choices)
  b. Different components for different sessions (needed for 4.0 rendering)
2) For a component, the functions `__init__`, `get_config` and `update` have essentially the same parameters. This leads to lots of code duplication, and will be annoying for custom components in 4.0.

Proposal:
1. Instead of storing component properties as attributes of the object, we store them in a new Config object

CHANGE:
```diff
class Textbox(...):
    def __init__(lines=..., placeholder=....)
-        self.lines = lines
-        self.placeholder = placeholder
+        cfg = Config()
+        cfg.lines = lines
+        cfg.placeholder = placeholder
```
2. All Component backend methods take a Config, and use that for logic. This Config is stored per user session, similar to States.

CHANGES:
```diff
class Number(...):
-    def preprocess(self, x: float | None) -> float | None:
-        if self.minimum is not None and x < self.minimum:
+    def preprocess(self, x: float | None, cfg: Config) -> float | None:
+        if cfg.minimum is not None and x < cfg.minimum:
            ...
```

3. Remove the `get_config` function entirely. Wherever we need it, we simply get the `Config` object, whose initial state is stored in the Component object, and whose updated states are tracked per user session.

4. Remove `Component.update` entirely. Instead the user returns a Component object with the values to change. To do this, we will have to:
  a. Set all `__init__` default values to None, and overwrite None with actual default values in the constructor body. This allows us to know when a user has overriden values when the Component is being used as an update
  b. Use thread locals to know if a Component is initialized inside an event listener, and to treat it as an update if so. 

CHANGES:
```diff
def update_dropdown():
-    return gr.Dropdown.update(choices=["A", "B", "C", value="A"])
+    return gr.Dropdown(choices=["A", "B", "C", value="A"])
```

abidlabs:
I really like it. Just some quick thoughts:

1. Could we make this backwards-compatible and merge it into `main`? Two big advantages to that: (1) this will affect the backend code a lot (in a good way), which will be good to have in `main` to avoid merge conflicts that happen because I'm cleaning up the backend code for all of the components (2) The more code we can merge into `main` the better, so that we can test it out thoroughly before 4.0 launch

[...]

Issue #4590, opened 2023-06-21
update supports updating the elem_id or elem_classes of a component


Kilig947:
- [ 1] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
css code 
```
@keyframes highlight {
  0%, 100% {
    border: 2px solid transparent;
  }
  50% {
    border-color: yellow;
  }
}

#highlight_update {
  animation-name: highlight;
  animation-duration: 0.75s;
  animation-iteration-count: 3;
}
```
python code 
```
with gr.Blocks(css=theme.custom_css) as demo:
        with column:
            with gr.Column(elem_id='examples_col') as examples_column:
                gr.Markdown('# Get Started Quickly')
                with gr.Row():
                    hide_components = gr.Textbox(visible=False)
                    guidance_example = gr.Examples(examples=[[Noo.update(visible=False), gr.State('412211'), '123123124124', submit3.update(elem_id='highlight_update')]],
                                                   inputs=[Noo, i_say, hide_components, submit3],
                                                   label='基础对话'
                                                   )
                    guidance_plugins = gr.Examples(examples=[[Noo.update(visible=False), gr.State('41241241'), '123123124124'], []],
                                                   inputs=[Noo, i_say, hide_components],
                                                   label='插件功能',
                                                   )
                    guidance_settings = gr.Examples(examples=[[Noo.update(visible=False), '123123', '123123124124'], []],
                                                   inputs=[Noo, i_say, hide_components],
                                                   label='更多设置',
                                                   )
```

I would like to implement that when I click on the case corresponding to Examples, some buttons will be highlighted to indicate that the user should click here, please ask why update(elem_id='xxxxx') is not supported.

Kilig947:
![image](https://github.com/gradio-app/gradio/assets/46100050/0b7613e8-1409-4121-9c05-d919cc5b566c)

After trying to change the source code, I was able to achieve what I wanted

Hope to develop this soon

abidlabs:
Thanks @Kilig947, we'll probably tackle this when we improve `.update()` as part of #4359. I'll leave this open to make sure we fix this.

pngwn:
I don't think we want to 'fix' this. User shouldn't be dynamically updating IDs, they should persist for the lifetime of the app. Changing IDs at runtime is likely to cause all kinds of unexpected behaviour, especially with CSS.

That said, we should allow updating the class names that are applied to the element.

abidlabs:
[...]

PR #5066, opened 2023-08-02
Remove update(), get_config() from every component

### LINKED ISSUE #5068: Not able to update num_top_classes for Label component
### Describe the bug

Is there any reason for not allowing the `num_top_classes` to be updated using the `update` method for `Label` component?

I would like to dynamically choose this value based on user input. Kindly advise, if there are any other ways to achieve this. Thanks.

### Reproduction

gr.Label.update(top_num_classes=x)

### Screenshot

_No response_

### Logs

```shell
TypeError: update() got an unexpected keyword argument 'top_num_classes'
```

### System Info

```shell
Mac
```

### Severity

Blocking usage of gradio

aliabid94:
No more update() and get_config()! Everything is done through the constructor now, see blocks_essay/run.py below. Also now keep a copy of components per session.

Should make our code a lot easier to manage. Two main changes as a result:
1. No default values in the constructor, use default() method in body instead. This allows us to set the value when rendering, but not for updates which need to be None so we know to skip those values.
2. Any argument that is in the constructor that becomes a component property will be considered part of the config and sent to the backend (open to other logic here)

Still in draft for cleanup reasons (and some components are in a broken state with update mode) but open to feedback. uses design from #4774 

```python
import gradio as gr


def change_textbox(choice):
    if choice == "short":
        return gr.Textbox(lines=2, visible=True)
    elif choice == "long":
        return gr.Textbox(lines=8, visible=True, value="Lorem ipsum dolor sit amet")
    else:
        return gr.Textbox(visible=False)


with gr.Blocks() as demo:
    radio = gr.Radio(
        ["short", "long", "none"], label="What kind of essay would you like to write?"
    )
    text = gr.Textbox(lines=2, interactive=True, show_copy_button=True)

    radio.change(fn=change_textbox, inputs=radio, outputs=text)


if __name__ == "__main__":
    demo.launch()
```

Fixes #5068


abidlabs:
Nice @aliabid94! High-level feedback:

We should make `.update()` a method of `Component` which is an alias for the constructor to preserve backwards compatibility. 

> No default values in the constructor, use default() method in body instead. This allows us to set the value when rendering, but not for updates which need to be None so we know to skip those values.

A side effect of this will be that we no longer will have the default values documented in the docs. Is there a way for `default()` to let the docs know what the default values are?

> Any argument that is in the constructor that becomes a component property will be considered part of the config and sent to the backend (open to other logic here)

This is great -- sets us up nicely for configs-per-session. But we should implement configs-per-session soon because otherwise people will try to update these properties and report that they don't work.

freddyaboulton:
Why haven't any of the checks run in this PR?

freddyaboulton:
Was looking at the fastapi source code and what if we borrow [their approach](https://github.com/tiangolo/fastapi/blob/master/fastapi/routing.py#L199) to determining default values, where they wrap the default value, by a class called `Default`.

The benefit is that we don't have to modify each constructor arg to be None (I think that's odd for custom components) and the documentation will still work (with minor changes). I also think we can get rid of accessing the thread local since we can change `delete_none` to `delete_default` and omit any values that are an instance of `Default`.

A sketch of what what would look like for textbox. 

```python
class Default:
    """
    Do not use this class directly.

    It's used internally to recognize when a default value has been changed
    """

    def __init__(self, value: Any):
        self.value = value


@document()
class Textbox:

    def __init__(
        self,
        value: str | Callable | None = Default(""),
        *,
        lines: int = Default(1),
        max_lines: int = Default(20),
        placeholder: str | None = Default(None),
        label: str | None = Default(None),
        info: str | None = Default(None),
        every: float | None = None,
        show_label: bool | None = Default(None),
        container: bool = Default(True),
        scale: int | None = Default(None),
        min_width: int = Default(160),
        interactive: bool | None = Default(None),
        visible: bool = Default(True),
        elem_id: str | None = None,
        autofocus: bool = Default(False),
        type: Literal["text", "password", "email"] = Default("text"),
        text_align: Literal["left", "right"] | None = Default(None),
        rtl: bool = Default(False),
        show_copy_button: bool = Default(False),
        **kwargs,
    ):
        pass
```

Both this approach and the one proposed have some boilerplate but I think I prefer this. I can fork off this PR and implement this so we can compare side by side?

Regarding things like:
> Any argument that is in the constructor that becomes a component property will be considered part of the config and sent to the backend (open to other logic here)

Are props like `elem_classes` and `elem_id` meant to be updated from the backend?

aliabid94:
Can you expand your code to show how the value held by the Default class would get extracted, depending on whether it is a new class being created or just an "update"?

aliabid94:
> We should make .update() a method of Component which is an alias for the constructor to preserve backwards compatibility.

Already added as part of Block class, is that not sufficient

> No default values in the constructor, use default() method in body instead. This allows us to set the value when rendering, but not for updates which need to be None so we know to skip those values.

> A side effect of this will be that we no longer will have the default values documented in the docs. Is there a way for default() to let the docs know what the default values are?

Hmm hadn't thought of that. If @freddyaboulton's approach works, then we could preserve this more easily

> This is great -- sets us up nicely for configs-per-session. But we should implement configs-per-session soon because otherwise people will try to update these properties and report that they don't work.

Actually decided to go ahead and add it to this PR, just fleshing it out

abidlabs:
> Are props like elem_classes and elem_id meant to be updated from the backend?

In https://github.com/gradio-app/gradio/issues/4590, @pngwn suggested allowing `elem_classes` but *not* `elem_id` to be updated. We can handle `elem_id` as a special case @aliabid94

aliabid94:
Ok now configs are stored per user. Try the updated blocks_essay/run.py. The minimum and maximum sliders affect the minimum and maximum of the gr.Number, which if submitted with an out of range number, will throw an error in the backend for that specific session. If you open a second session, you will see that that session will have it's own minimum and maximum unaffected by the other session.

aliabid94:
> In https://github.com/gradio-app/gradio/issues/4590, @pngwn suggested allowing elem_classes but not elem_id to be updated. We can handle elem_id as a special case

This is gonna make it messy making exceptions, I'm not sure I get why users can't update an elem_id if they'd like. Seems very reasonable to do so.

aliabid94:
I believe all that's left is considering @freddyaboulton's suggested API change, and cleanup/tests.

abidlabs:
Very cool! Will give this PR a spin tomorrow

abidlabs:
I don't think that the defaults are being preserved correctly. Started trying this out -- ran into an error:

```py
import gradio as gr

def update_type():
    return gr.Image(type="filepath")

with gr.Blocks() as demo:
    im = gr.Image()
    btn = gr.Button()
    btn.click(update_type, None, im)
    
demo.launch()
```

Clicking on the button gives me this error: 

```
ValueError: Invalid value for parameter `source`: None. Please choose from one of: ['upload', 'webcam', 'canvas']
```

aliabid94:
Went ahead and implemented @freddyaboulton's design, I agree it's much cleaner:
```python
class Textbox(
...):
    def __init__(
        self,
        value: str | Callable | Default = Default(""),
        *,
        lines: int | Default = Default(1),
        max_lines: int | Default = Default(20),
        placeholder: str | None | Default = Default(None),
        label: str | None | Default = Default(None),
        info: str | None | Default = Default(None),
        ...
    ):
        """
        self.lines = get(lines)
        self.max_lines = get(max_lines)
        self.autofocus = get(autofocus)
        self.type = get(type)
```

The get method resolves a Default object appropriately depending on whether the code is within an update.

abidlabs:
> The get method resolves a Default object appropriately depending on whether the code is within an update.

Nice! Does this also allow us to keep the default value in the docs? 

I'll open this up for review and give it another pass

abidlabs:
Documenting issues I noticed as I was testing this:

1. Creating an `Interface` raises a warning:

```py
import gradio as gr

gr.Interface(lambda x:x, "text", "text").launch()
```

2. `gr.update()` and `gr.Component.update()` do not seem to be working properly. Try with:

```py
import gradio as gr

def test(x):
    return gr.update(visible=False)
    # or try
    # return gr.Textbox.update(visible=False)

with gr.Blocks() as demo:
    t1 = gr.Textbox()
    t2 = gr.Textbox()
    t1.change(test, t1, t2)
    
demo.launch()
```

abidlabs:
3. I'm getting some issues using this new approach with components like `gr.Audio` and `gr.Image`. The error is of this form:

```
ValueError: Invalid value for parameter `source`: <class 'gradio.blocks.NoOverride'>. Please choose from one of: ['upload', 'microphone']
```

Here's the example demo:

```py
import gradio as gr

def test(x):
    return gr.Audio(x, visible=False)

def test2(x):
    return gr.Audio(x, visible=True)

with gr.Blocks() as demo:
    t1 = gr.Audio()
    t2 = gr.Audio()
    t3 = gr.Audio()
    t1.change(test, t1, t2)
    t3.change(test2, t1, t2)
    
demo.launch()
```

abidlabs:
4. Maybe we should raise a warning if people return components of incompatible types? E.g.

```py
import gradio as gr

def test(x):
    return gr.Button(visible=False)

with gr.Blocks() as demo:
    t1 = gr.Audio()
    t2 = gr.Audio()
    t1.change(test, t1, t2)
    
demo.launch()
```

pngwn:
@aliabid94 
> This is gonna make it messy making exceptions, I'm not sure I get why users can't update an elem_id if they'd like. Seems very reasonable to do so.

Because people tend to use ids for styling and if they change them stuff will break and we don't have a way to reference a component by a fixed id (this was kinda the fixed id). 

I don't feel very strongly about it though, so I'm fine either way if it is messy to change.

aliabid94:
@abidlabs I believed I have fixed all your issues, can you give this another pass.

> Because people tend to use ids for styling and if they change them stuff will break and we don't have a way to reference a component by a fixed id 

We don't use elem_ids internally to reference anything, right? It's only if the users were tracking things by id, and it changed, which I think users should be allowed to do

pngwn:
i heard that this PR might include changes. Could just be a vicious rumour tho.

abidlabs:
Can confirm that all the original issues have been fixed! I'll give this another complete pass when I can.

It'd be good to get some more eyes on this since this touches a lot of fiels -- cc @dawoodkhan82 can you take a look as well?

.changeset/beige-ants-drop.md
```diff
@@ -0,0 +1,5 @@
+---
+"gradio": patch
```
abidlabs, line 2: Should be a minor since we're updating the API. 

Can you also add a highlight, [like done in this PR](https://github.com/gradio-app/gradio/pull/5057), showing how to use the new API to update the components?

demo/hello_blocks/run.py
```diff
@@ -8,6 +8,8 @@ def greet(name):
     output = gr.Textbox(label="Output Box")
     greet_btn = gr.Button("Greet")
     greet_btn.click(fn=greet, inputs=name, outputs=output, api_name="greet")
+
+    print(greet_btn.get_config())
```
abidlabs, line 12: Remove

abidlabs, line 12: Remove
```suggestion
```

demo/blocks_kitchen_sink/run.py
```diff
@@ -112,7 +112,7 @@
             check = gr.Checkbox(label="Go")
         with gr.Column(variant="panel", scale=2):
             img = gr.Image(
-                "https://gradio.app/assets/img/header-image.jpg", label="Image"
+                # "https://gradio.app/assets/img/header-image.jpg", label="Image"
```
abidlabs, line 115: Did you mean to comment this otu?

abidlabs:
Added some nits above. We should also update this Guide: https://www.gradio.app/guides/blocks-and-event-listeners#updating-component-configurations

and the description and the Interface example in the docs here: https://www.gradio.app/docs/update

gradio/routes.py
```diff
@@ -409,9 +409,8 @@ async def run_predict(
             if hasattr(body, "session_hash"):
                 if body.session_hash not in app.state_holder:
                     app.state_holder[body.session_hash] = {
-                        _id: deepcopy(getattr(block, "value", None))
+                        _id: deepcopy(block)
```
freddyaboulton, line 412: We should only deep copy the blocks that are inputs to the event, no? Doing a copy of every block will probably be really slow for big apps

gradio/blocks.py
```diff
@@ -304,39 +329,49 @@ def set_event_trigger(
...
+            for k in signature.parameters:
+                if hasattr(self, k):
+                    v = getattr(self, k)
+                    config[k] = v
+
+        return config
 
     @staticmethod
-    @abstractmethod
     def update(**kwargs) -> dict:
```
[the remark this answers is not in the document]
aliabid94: ahh dammit looks like we can't get rid of update from every component yet then, since we need to respect the order we set the arguments in the update method of Component :/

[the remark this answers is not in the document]
freddyaboulton: Oh dang yea that's right the arg order in init vs update is sometimes different

PR #5070, opened 2023-08-02
Remove explicit return types rule

hannahblair:
## Description

I'm proposing we remove eslint's [explicit function return types rule](https://typescript-eslint.io/rules/explicit-function-return-type/). We have quite a few of these errors around our repo (e.g `Dataset.svelte`) but can't fix these if the function is inline because Svelte can't parse non-standard JS expressions in the markup of the file without adding an external transpiler into the mix, it can only do so in the `<script />`. The only way of resolving the error is moving the func to the `<script />` but this isn't ideal, and I think we should just remove this rule to tidy up our errors.  [This is a known limitation.](https://github.com/sveltejs/svelte-preprocess/blob/main/docs/preprocessing.md#typescript---limitations) [Also a discussion around it here.](https://github.com/sveltejs/svelte/issues/4701)

Open to discussion around this!

Screenshot 2023-08-02 at 15 18 12


pngwn:
We don't need to remove the rule, just play with the options I've fixed this in my refactor PR by tweaking this rule: https://github.com/gradio-app/gradio/blob/de24ec62bba9307645c3c3722a9673afa9828726/.config/eslint.config.js#L38-L41

PR #5077, opened 2023-08-03
Live audio streaming output

aliabid94:
This PR allows users to stream audio out. See demo/streaming_audio_out for an example that streams out pieces of an audio file second by second.

Fixes: #5110


abidlabs:
Tested this out and it works great @aliabid94 even with multiple outputs! However, I'm concerned about the fact that it uses a completely different mechanism for streaming, as compared to regular generator function including a separate `/stream` route. 

Is this actually necessary? I don't have a concrete alternative right now, but among other things, this breaks the client for any route with streaming. You can try by running:

```py
import gradio as gr
import numpy as np
from pydub import AudioSegment
import time

def stream_audio(lag):
    audio_file = 'test.mp3'  # Your audio file path
    audio = AudioSegment.from_mp3(audio_file)
    chunk_length = 1000
    chunks = []
    while len(audio) > chunk_length:
        chunks.append(audio[:chunk_length])
        audio = audio[chunk_length:]
    if len(audio):  # Ensure we don't end up with an empty chunk
        chunks.append(audio)

    def iter_chunks():  
        for chunk in chunks:
            file_like_object = chunk.export(format="mp3")
            data = file_like_object.read()
            time.sleep(lag)
            yield data

    return iter_chunks(), "fixed response"

demo = gr.Interface(
    stream_audio,
    gr.Slider(0, 3, 0, label="lag", info="Duration before generating next second of audio. >1s to cause lag."),
    [gr.Audio(autoplay=True), gr.Textbox()]
)

if __name__ == "__main__":
    _, url, _ = demo.launch()
```

and then:

```py

client = Client(url)
result = client.predict(
				0,	# int | float (numeric value between 0 and 3) in 'lag' Slider component
				api_name="/predict"
)
print(result)
```

abidlabs:
Ok so here's an idea that doesn't fix everything above but I think would allow you to use the above developer API.

Steps:
* Developer writes a regular generator function, something like this:

```py
def stream_audio(lag):
   ...
        for chunk in chunks:
            file_like_object = chunk.export(format="mp3")
            data = file_like_object.read()
            time.sleep(lag)
            yield data, "fixed response"
```

and sets `streaming=True` in the `Audio()` output component like this:

```py
demo = gr.Interface(
    stream_audio,
    gr.Slider(0, 3, 0, label="lag", info="Duration before generating next second of audio. >1s to cause lag."),
    [gr.Audio(autoplay=True, streaming=True), gr.Textbox()]
)
```

* Gradio sees if *any* of the outputs have set `streaming=True` and if so, doesn't evaluate the generator function, but instead pass it into a FastAPI StreamingResponse
* In the Client, we can check to see if an endpoint has any outputs that stream, and if so, we make them invalid endpoints so that they don't show up in the view API page

aliabid94:
> Gradio sees if any of the outputs have set streaming=True and if so, doesn't evaluate the generator function, but instead pass it into a FastAPI StreamingResponse

StreamingResponse requires a generator that only yields bytes. We could "wrap" the generator with another generator that tosses out all other outputs. However this will obviously ignore the intended user behaviour of setting the other outputs. There's no way we would be able to get access to the other outputs because we don't have access to the outputs as they are being yielded - only FastAPI does, so we can't send updates or anything with those outputs.

abidlabs:
Suppose you you wanted to create a demo that streamed music and also generated lyrics in realtime for the streaming music. That would not be possible with this API, correct?

abidlabs:
I think we need to do something like this:

* When a user passes in a regular generator function, and one of the output components has `streaming=True`, then a `pending_streams` is created for that component. Think of the `pending_streams` a list.
* On every iteration of the generator function, we take the output corresponding corresponding to that component and append it to `pending_streams`. Once the generator function is complete, we append a special `StopIteration` token to `pending_streams`

In the `/stream` route, we define a generator that looks like this:

```py
def stream_until_complete():
   chunks = pending_streams[id]
   chunk = None
   index = 0
   while not chunk == StopIteration:
      yield chunk
      if index >= len(chunks):
         yield None
      else:
          chunk = chunks[index]
          index += 1
```

(code may need to be tweaked but this is the general idea)

Then you pass in `stream_until_complete` into FastAPI's `StreamingResponse`

aliabid94:
Ok now I accept direct yielding from the function, see demo/stream_audio_out/ for an example. Ready for re-review @abidlabs

demo/stream_audio_out/run.py
```diff
@@ -0,0 +1,40 @@
...
+        chunks.append(audio[:chunk_length])
+        audio = audio[chunk_length:]
+    if len(audio):  # Ensure we don't end up with an empty chunk
+        chunks.append(audio)
+
+    for i, chunk in enumerate(chunks):
+        file = f"/tmp/{i}s.mp3"
+        chunk.export(file, format="mp3")
+        time.sleep(lag)
+        yield file, i
```
abidlabs, line 21: So you create the list of chunks and then yield then one at a time. To simplify the demo (and maybe make it more realistic), why not just yield each chunk directly as you get it?

guides/02_building-interfaces/02_reactive-interfaces.md
```diff
@@ -19,4 +19,6 @@ The difference between `gr.Audio(source='microphone')` and `gr.Audio(source='mic
 
 Here is example code of streaming images from the webcam.
 
-$code_stream_frames
\ No newline at end of file
+$code_stream_frames
+
+Streaming can also be done in an output component. A `gr.Audio(streaming=True)` output component can take a stream of audio data yielded piece-wise by a generator function and combines them into a single audio file.
```
abidlabs, line 24: Let's put the `stream_audio_out` example demo here (ideally after simplifying it a bit)

.changeset/famous-rice-taste.md
```diff
@@ -0,0 +1,6 @@
+---
+"@gradio/upload": patch
+"gradio": patch
+---
+
+fix:Live audio streaming output
```
abidlabs, line 6: Let's include a highlight here, see https://github.com/gradio-app/gradio/pull/5057 for an example

gradio/blocks.py
```diff
@@ -1331,13 +1333,37 @@ def postprocess_data(
 
         return output
 
+    def handle_streaming_outputs(
+        self, fn_index: int, data: list, session_hash: str | None, run: int | None
```
abidlabs, line 1337: Just wondering, why do you need a separate `run` index? Isn't `session_hash` + `fn_index` enough?

abidlabs:
Here's a simplified demo you can use @aliabid94:

```py
import gradio as gr
from pydub import AudioSegment
import time


def stream_audio(audio_file, lag):
    audio = AudioSegment.from_mp3(audio_file)
    chunk_length = 1000
    chunks = []
    while len(audio) > chunk_length:
        chunks.append(audio[:chunk_length])
        audio = audio[chunk_length:]
    if len(audio):  # Ensure we don't end up with an empty chunk
        chunks.append(audio)

    for i, chunk in enumerate(chunks):
        file = f"/tmp/{i}s.mp3"
        chunk.export(file, format="mp3")
        time.sleep(lag)
        yield file, i


demo = gr.Interface(
    stream_audio,
    [
        gr.Audio(type="filepath", label="Audio File"),
        gr.Slider(
            0,
            3,
            0,
            label="lag",
            info="Duration before generating next second of audio. >1s to cause lag.",
        ),
    ],
    [gr.Audio(autoplay=True, streaming=True), "textbox"],
)

if __name__ == "__main__":
    demo.queue().launch()
```

abidlabs:
Noticing some small issues:

1. When trying to stream two output audio files at the same time, this doesn't work (raises a mysterious `KeyError`):

Here's an adaption of the code above:

```py
import gradio as gr
from pydub import AudioSegment
import time


def stream_audio(audio_file, lag):
    audio = AudioSegment.from_mp3(audio_file)
    i = 0
    chunk_size = 1000
    
    while chunk_size*i < len(audio):
        chunk = audio[chunk_size*i:chunk_size*(i+1)]
        i += 1
        if chunk:
            file = f"/tmp/{i}.mp3"
            chunk.export(file, format="mp3")            
            yield file, file
        
demo = gr.Interface(
    fn=stream_audio,
    inputs=[
        gr.Audio(type="filepath", label="Audio file to stream"),
        gr.Slider(0, 3, 0,
            label="lag",
            info="Duration before generating next second of audio. Set >1s to cause lag.",
        ),
    ],
    outputs=[
        gr.Audio(
            autoplay=True, 
            streaming=True), # needed to stream output audio
        gr.Audio(
            autoplay=True, 
            streaming=True), # needed to stream output audio
    ],
)

if __name__ == "__main__":
    demo.queue().launch()
```

2. There's a very brief but slightly jarring discontinuity in between the chunks when livestreaming the output audio. Not clear to me if its an issue with the chunking logic or with the streaming logic.

aliabid94:
> When trying to stream two output audio files at the same time, this doesn't work (raises a mysterious KeyError):

Fixed.

> There's a very brief but slightly jarring discontinuity in between the chunks when livestreaming the output audio

I think it's because we were streaming 1 second chunks, which was too frequent. Increased to 3 second chunks in the demo and the breaks are much better imo.

abidlabs:
> I think it's because we were streaming 1 second chunks, which was too frequent. Increased to 3 second chunks in the demo and the breaks are much better imo.

I think I still hear them in the 3 second, but its very minor so not a blocker imo.

abidlabs (approved):
Awesome PR @aliabid94!

PR #5092, opened 2023-08-04
generate docs json in ci, reimplement main vs release

pngwn:
## Description

Alternative to #4953

This PR generates the docs in CI and brings back the main vs release docs.

Our current website/ implementation is kinda weird because we generate docs manually, so the chance for things to get out of date (especially the docs) is very high. This generates the content in CI to prevent that issue. It also brings back the separate main + release docs.

On the website side, I have changed things around a little. Navigating to `/<version>/ docs` or `/<version>/guides/` or `/<version>/guides/<guide>` will now show the docs for that release version (or `main` if the version is `main`). Navigating to `/docs` or `/guides/` or `/guides/<guide>` will now show the docs for the latest release (as defined by the local package.json, which is always correct). This will technically allow us to add as many version as we want to, tot he site but I haven't added that option in the navigation just yet.

The way it works is that `/main` will always use the locally generated json files, everything else will fetch the correct docs from an AWS bucket that I have set up. We push to this AWS bucket in the changeset release PR.  This is safe because the package.json has only been bumped in that PR and not on main. So even though we are writing the files to the aws bucket early, the website won't show those docs because the package.json on main will still show the currently published version. This also has the added benefit of allowing us to publish the docs that _will_ be published in that PR. So this kind of completes the package for the release automation, coupling our docs to releases.

I do have ti publish the site semi-manually via CI (using the Vercel CLI) rather than the GitHub integration which is slightly annoying but pretty straightforward.

Few small things to take care of but the code stuff is mostly done. Will mark as ready when I'm finished.

## 🎯 PRs Should Target Issues

## Tests


pngwn:
just need to sort CI

abidlabs:
Amazing @pngwn!

aliabd:
this is great @pngwn thanks for handling this! will review when the PR is ready but this will make so many problems go away.

PR #5080, opened 2023-08-03
Add icon and link params to `gr.Button`

hannahblair:
## Description

This PR allows us to add a link and icon to the `gr.Button` component. The default icon placement is central alongside the text:

Screenshot 2023-08-03 at 16 03 37

If a different icon placement is desired, we can use custom css. 

Screenshot 2023-08-03 at 16 15 10

--- 
Icon Demo

```
import gradio as gr

with gr.Blocks(css = """button {justify-content: left !important;}""") as demo:
    gr.Button(value="Login with HF", icon='https://huggingface.co/front/assets/huggingface_logo-noborder.svg')
    
demo.launch()
```

--- 
Link Demo

```
import gradio as gr

with gr.Blocks() as demo:
    gr.Button(value="Login with HF", link="https://huggingface.co/login")
    
demo.launch()
```

Closes: #5060 

## 🎯 PRs Should Target Issues

Before your create a PR, please check to see if there is [an existing issue](https://github.com/gradio-app/gradio/issues) for this change. If not, please create an issue before you create this PR, unless the fix is very small. 

Not adhering to this guideline will result in the PR being closed. 

## Tests

1. PRs will only be merged if tests pass on CI. To run the tests locally, please set up [your Gradio environment locally](https://github.com/gradio-app/gradio/blob/main/CONTRIBUTING.md) and run the tests: `bash scripts/run_all_tests.sh`

2. You may need to run the linters: `bash scripts/format_backend.sh` and `bash scripts/format_frontend.sh`

PR #5111, opened 2023-08-07
Remove icon and link from DuplicateButton docstring

aliabd:
PR #5080 added icon and link params to `gr.Button` and `gr.ClearButton` (which inherits from `Button`). But it didn't add them to `gr.DuplicateButton` and `gr.UploadButton` which also inherit from `Button`. That's fine for `UploadButton` because it defines its own docstrings, but `DuplicateButton` doesn't and inherits its docstrings from `Button`. This broke website docs generation because it was looking for the icon/link params that exist in the docstrings and not in the `__init__
`. 

Simple fix was adding docstrings to `DuplicateButton` that don't include icon and link.


abidlabs:
Thanks for catching this @aliabd. I agree with having `DuplicateButton` having its own docstring, but could we also add the `icon` and `link` parameters to these buttons? 

I also think it would be nice if we made the Spaces duplicate logo (screenshotted below) the default `icon` for the `gr.DuplicateButton`

image

Could you make these changes?

aliabd:
@abidlabs I added icon/link as params to `DuplicateButton`, but it seems to ignore them when I test locally. Not sure why? @hannahblair can you take a look? 

Also, I didn't add them to `UploadButton` because I realized that actually doesn't inherit from `Button`

abidlabs:
LGTM @aliabd!

abidlabs (approved):

abidlabs:
I'll go ahead and merge this in since @aliabd is ooo for a couple hours and this may unblock #5092

Issue #5103, opened 2023-08-04
Clearing a textbox will set its value to None, not ""


aliabid94:
This can cause issues if a user clicks clear, then submits. for example, run demo/hello_world, click clear, and then click submit. Getting None causes an error in the backend, which expects a string.

abidlabs:
Hmm interesting... this is true of all components... the Clear button always sets to None. In the frontend for Textbox, maybe we need to make None correspond to "".

PR #5114, opened 2023-08-07
Reset textbox value to empty string when value is None

hannahblair:
## Description

When the textbox value is cleared, it's set to `None` on the backend. When the value is set to None in the backend, this PR adds a func set it to `""` in the frontend. 

Closes: #5103

## 🎯 PRs Should Target Issues

## Tests


freddyaboulton (approved):
Thanks for the quick fix @hannahblair !

Issue #5115, opened 2023-08-07
Add support for async fn to `ChatInterface` 


ZeusFSX:
### Describe the bug

I tried to mount the async function to ChatInterface and got the error. Please take a look at the logs included below.
If you have suggestions on how I can implement the Chat Interface with the async function, I'd like to read them.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

I used the FastApi interface to mount the application
```
async def gradio_app_for_bi(message: str, history: list[list[str]]) -> str:
    ....

chatbot = gr.ChatInterface(fn=gradio_app_for_bi, title="Business Demo")
app = gr.mount_gradio_app(app, chatbot, path="/gradio")
```

### Logs

```shell
ValueError: Invalid message for Chatbot component: <coroutine object gradio_app_for_bi at 0x16ed689e0>
```

### System Info

```shell
Gradio Environment Information:
Operating System:  Darwin


gradio version:  3.39.0

gradio Dependencies:
  aiofiles: 23.1.0
  aiohttp: 3.8.4
  altair: 5.0.1
  fastapi: 0.100.1
  ffmpy: 0.3.0
  gradio-client: 0.3.0
  httpx: 0.24.1
  huggingface-hub: 0.16.4
  jinja2: 3.1.2
  markdown-it-py: 2.2.0
  markupsafe: 2.1.3
  matplotlib: 3.7.2
  mdit-py-plugins: 0.3.3
  numpy: 1.25.1
  orjson: 3.9.2
  packaging: 23.1
  pandas: 2.0.3
  pillow: 10.0.0
  pydantic: 1.10.11
  pydub: 0.25.1
  python-multipart: 0.0.6
  pyyaml: 6.0
  requests: 2.31.0
  semantic-version: 2.10.0
  typing-extensions: 4.5.0
  uvicorn: 0.22.0
  websockets: 11.0.3


gradio_client version:  0.3.0

gradio_client Dependencies:
  fsspec: 2023.6.0
  httpx: 0.24.1
  huggingface-hub: 0.16.4
  packaging: 23.1
  requests: 2.31.0
  typing-extensions: 4.5.0
  websockets: 11.0.3
```

### Severity

Blocking usage of gradio

abidlabs:
Hi @ZeusFSX not at the moment, that's on my list of todos. Will take a look at this soon.

Leumastai:
Checking in if this has been implemeted?

[...]

PR #5116, opened 2023-08-07
Add support for async functions and async generators to `gr.ChatInterface`

abidlabs:
Added async support for all of the routes: the submit function, the example function, and the API functions. Test with any async function or generator, e.g.: 

```py
async def slow_echo(message, history):
    for i in range(len(message)):
        time.sleep(0.05)
        yield "You typed: " + message[: i+1]

demo = gr.ChatInterface(slow_echo, examples=["abcdef"], cache_examples=True).queue()

demo.launch()
```

Closes: #5115


gradio/chat_interface.py
```diff
@@ -397,68 +403,100 @@ def _display_input(
...
-    ) -> Generator[tuple[list[list[str | None]], list[list[str | None]]], None, None]:
+    ) -> AsyncGenerator:
         history = history_with_input[:-1]
         generator = self.fn(message, history, *args, **kwargs)
         try:
-            first_response = next(generator)
+            if self.is_async:
+                first_response = await async_iteration(generator)
+            else:
+                first_response = next(generator)
```
freddyaboulton, line 434: Since we're switching everything to coroutines, any possible blocking code that's not async needs to be run in a separate thread otherwise we'll block the event loop. So for example use `anyio.to_thread.run_sync` to run the non-async code. 

Might be better to `utils.SyncToAsyncIterator` for the generators to match what we do in `blocks`. 

Otherwise this looks good though!

abidlabs: Ah nice I was going to ask why we do `utils.SyncToAsyncIterator` in `blocks.py`

abidlabs:
Ok I've updated to use `SyncToAsyncIterator` @freddyaboulton. Let me know if you see anything else or if this is good to merge.

gradio/chat_interface.py
```diff
@@ -397,67 +403,86 @@ def _display_input(
...
         history_with_input: list[list[str | None]],
         *args,
         **kwargs,
     ) -> tuple[list[list[str | None]], list[list[str | None]]]:
         history = history_with_input[:-1]
-        response = self.fn(message, history, *args, **kwargs)
+        if self.is_async:
+            response = await self.fn(message, history, *args, **kwargs)
+        else:
+            response = self.fn(message, history, *args, **kwargs)
```
freddyaboulton, line 417: We need to run the non-coroutine in a separate thread otherwise we'll block the event loop. This is because the _submit_fn is now async.

If you run this code on your branch, two events can't run concurrently anymore

```python
import random
import gradio as gr
import time

def random_response(message, history):
    time.sleep(5)
    return random.choice(["Yes", "No"])

demo = gr.ChatInterface(random_response)

if __name__ == "__main__":
    demo.launch()
```

Note how the textbox submit event has to wait:
chatbot_concurrency

compare to main:
chatbot_concurrency_main

abidlabs: Ah nice thanks @freddyaboulton

abidlabs:
Now should be fixed. Tested with regular functions as well as with iterators (the latter requires enabling queuing and setting concurrency count > 1), and regular functions do not seem to block.

freddyaboulton (approved):
Thanks for the enhancement @abidlabs !!

abidlabs:
Appreciate the guidance around non-blocking thanks @freddyaboulton!

PR #5125, opened 2023-08-08
chatbot conversation nodes can contain a copy button

fazpu:
## Description
The PR introduces a new parameter `show_chat_copy_button` to the `gr.Chatbot`. When `True`, it exposes the `Copy` button for each converstation node.

image


pseudotensor:
Cool!

gradio/components/chatbot.py
```diff
@@ -53,6 +53,7 @@ def __init__(
         latex_delimiters: list[dict[str, str | bool]] | None = None,
         rtl: bool = False,
         show_share_button: bool | None = None,
+        show_chat_copy_button: bool | None = None,
```
abidlabs, line 56: Can just set to False by default
```suggestion
        show_chat_copy_button: bool = False,
```

demo/chatbot_simple/run.py
```diff
@@ -3,7 +3,7 @@
 import time
 
 with gr.Blocks() as demo:
-    chatbot = gr.Chatbot()
+    chatbot = gr.Chatbot(show_chat_copy_button=True)
```
abidlabs, line 6: I'd revert this change, this demo is intentionally designed to be as simple as possible. Furthermore, this parameter won't actually have any effect since there's no way to populate this chatbot with messages.
```suggestion
    chatbot = gr.Chatbot()
```

gradio/components/chatbot.py
abidlabs: Can we name this argument just `show_copy_button` to make a little bit shorter.

abidlabs:
I tested the copy button and I think its fine to include. I still would disable it by default but providing developers with the option to include it if they want for their Chatbots seems like a good idea. 

@hannahblair and I added some comments above. One more thing -- clicking on the copy button does not provide any sort of visual feedback that the button has been clicked. @fazpu could you do something similar to what we do for the `JSON` component where the copy button changes to a checkmark?

You can see that by running something like this:

```py
with gr.Blocks() as demo:
    gr.JSON({"ab": "cd"})
    
demo.launch()
```

Once the changes have been made, I'm fine with merging, wdyt @freddyaboulton @hannahblair?

gradio/components/chatbot.py
freddyaboulton: Agreed

freddyaboulton:
Sounds good to me @abidlabs @fazpu !

gradio/components/chatbot.py
fazpu: changed

gradio/components/chatbot.py
```diff
@@ -72,6 +73,7 @@ def __init__(
             latex_delimiters: A list of dicts of the form {"left": open delimiter (str), "right": close delimiter (str), "display": whether to display in newline (bool)} that will be used to render LaTeX expressions. If not provided, `latex_delimiters` is set to `[{ "left": "$$", "right": "$$", "display": True }]`, so only expressions enclosed in $$ delimiters will be rendered as LaTeX, and in a new line. Pass in an empty list to disable LaTeX rendering. For more information, see the [KaTeX documentation](https://katex.org/docs/autorender.html).
             rtl: If True, sets the direction of the rendered text to right-to-left. Default is False, which renders text left-to-right.
             show_share_button: If True, will show a share icon in the corner of the component that allows user to share outputs to Hugging Face Spaces Discussions. If False, icon does not appear. If set to None (default behavior), then the icon appears if this Gradio app is launched on Spaces, but not otherwise.
+            show_chat_copy_button: If True, will show a copy button for each chatbot answer.
```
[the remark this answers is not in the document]
fazpu: done

js/chatbot/static/ChatBot.svelte
```diff
@@ -122,6 +125,12 @@
 									{/each}
 								</div>
 							{/if}
+
+							{#if show_chat_copy_button && message}
+								<div class="icon-button">
+									<Copy on:error value={message} />
```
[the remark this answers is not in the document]
fazpu: thanks, removed

js/chatbot/static/Copy.svelte
```diff
@@ -1,11 +1,10 @@
 <script lang="ts">
 	import { onDestroy } from "svelte";
-	import { fade } from "svelte/transition";
 	import { Copy, Check } from "@gradio/icons";
 
 	let copied = false;
 	export let value: string;
-	let timer: NodeJS.Timeout;
```
fazpu, line 8: this was not formally correct as we are dealing with client-side code. IDE protested as well :)

js/chatbot/static/Copy.svelte
```diff
@@ -46,30 +45,25 @@
...
 	{/if}
 </button>
 
 <style>
 	button {
 		position: relative;
-		cursor: pointer;
-		padding: 5px;
+		top: 0;
+		right: 0;
```
fazpu, line 60: added the positioning to improve readability and avoid unexpected behaviour

js/chatbot/static/Copy.svelte
```diff
@@ -46,30 +45,25 @@
 </script>
 
 <button on:click={handle_copy} title="copy">
-	<span class="copy-text" class:copied><Copy /> </span>
```
fazpu, line 49: this logic seemed to be unmaintained - it was not working. I simplified and removed all the outdated and non-necessary parts.

fazpu:
> I tested the copy button and I think its fine to include. I still would disable it by default but providing developers with the option to include it if they want for their Chatbots seems like a good idea.
> 
> @hannahblair and I added some comments above. One more thing -- clicking on the copy button does not provide any sort of visual feedback that the button has been clicked. @fazpu could you do something similar to what we do for the `JSON` component where the copy button changes to a checkmark?
> 
> You can see that by running something like this:
> 
> ```python
> with gr.Blocks() as demo:
>     gr.JSON({"ab": "cd"})
>     
> demo.launch()
> ```
> 
> Once the changes have been made, I'm fine with merging, wdyt @freddyaboulton @hannahblair?

@abidlabs Thank you for the feedback! The logic was already there in the Copy component. It just needed a small refresh. Please have a look.

abidlabs:
Tested, looks good. Thanks @fazpu for the great contribution 👍

abidlabs (approved):

js/chatbot/static/Copy.svelte
abidlabs: Getting a type error in our CI, which is probably due to this change:

```
/home/runner/work/gradio/gradio/js/chatbot/static/Copy.svelte:12:3
Error: Type 'Timeout' is not assignable to type 'number'. (ts)
		if (timer) clearTimeout(timer);
		timer = setTimeout(() => {
			copied = false;
```

fazpu: yes, fixed, thanks

abidlabs:
Thanks @fazpu for this PR! Merging

luoweb:
> I tested the copy button and I think its fine to include. I still would disable it by default but providing developers with the option to include it if they want for their Chatbots seems like a good idea.
> 
> @hannahblair and I added some comments above. One more thing -- clicking on the copy button does not provide any sort of visual feedback that the button has been clicked. @fazpu could you do something similar to what we do for the `JSON` component where the copy button changes to a checkmark?
> 
> You can see that by running something like this:
> 
> ```python
> with gr.Blocks() as demo:
>     gr.JSON({"ab": "cd"})
>     
> demo.launch()
> ```
> 
> Once the changes have been made, I'm fine with merging, wdyt @freddyaboulton @hannahblair?

no properties with gradio v 3.39.0, how to configure?  

GradioUnusedKwargWarning: You have unused kwarg parameters in Chatbot, please remove them: {'show_copy_button': True}

abidlabs:
@luoweb you have to upgrade to the latest version of `gradio`:

```bash
pip install --upgrade gradio
```

luoweb:
> @luoweb you have to upgrade to the latest version of `gradio`:
> 
> ```shell
> pip install --upgrade gradio
> ```

it's OK now, Thanks

huni0103:
Hi there.

I would like to make an additional suggestion regarding Chatbot Copy.
In the Chatbot, I am replacing "<" , ">" with "\>", "\<", etc. to display "<" , ">".
However, when I copy the Chatbot text, the "\" is also being copied.
I only want "<" and ">" to be copied as they appear in the Chatbot.

Furthermore, I'm representing newlines as "<br/>", but when I copy the Chatbot text, it's copying them as "<br/>" even though they are displayed as "\n" in the Chatbot.

Is there any way to address the issues mentioned above?"

PR #5135, opened 2023-08-08
Fix dataset features and dataset preview for HuggingFaceDatasetSaver

freddyaboulton:
## Description

Fixes:
- If a media file (Image/Audio) was None, flagging would fail
- Jsonl dataset format was not compatible with datasets library
- Paths in json files now include the full path relative to the root of the dataset dir. This helps loading to HuggingFace datasets library.

Internal thread where this came up: https://huggingface.slack.com/archives/C02QZLG8GMN/p1691444822924499

To test:

Set up a demo with flagging
```python
import gradio as gr

def echo(msg, history, img):
    return msg

with gr.Blocks() as demo:
    image = gr.Image(label='Image')
    chat = gr.ChatInterface(echo, additional_inputs=[image])
    flagger = gr.Button("Flag")

    flagging = gr.HuggingFaceDatasetSaver(
        dataset_name="chatinterface_with_image_csv",
        hf_token="<your token>",
        separate_dirs=True)
    flagging.setup([chat.chatbot, image], "chatinterface_with_image_csv_2")
    flagger.click(lambda *args: flagging.flag(flag_data=args), inputs=[chat.chatbot , image], outputs=None, preprocess=False)

demo.launch()
```

Flag a couple with and without the image as None, then check out the dataset:
https://huggingface.co/datasets/freddyaboulton/chatinterface_with_image_json

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/gradio/flagging.py
+++ b/gradio/flagging.py
@@ -261,6 +261,7 @@ class HuggingFaceDatasetSaver(FlaggingCallback):
                 ]
             },
             overwrite=True,
+            token=self.hf_token
         )
 
         # Setup flagging dir
--- a/gradio/flagging.py
+++ b/gradio/flagging.py
@@ -433,7 +433,8 @@ class HuggingFaceDatasetSaver(FlaggingCallback):
                 assert Path(deserialized).exists()
                 row.append(str(Path(deserialized).relative_to(self.dataset_dir)))
             except (AssertionError, TypeError, ValueError):
-                row.append(str(deserialized))
+                deserialized = "" if deserialized is None else str(deserialized)
+                row.append(deserialized)
 
             # If component is eligible for a preview, add the URL of the file
             # Be mindful that images and audio can be None
--- a/gradio/flagging.py
+++ b/gradio/flagging.py
@@ -248,6 +248,20 @@ class HuggingFaceDatasetSaver(FlaggingCallback):
             repo_type="dataset",
             exist_ok=True,
         ).repo_id
+        path_glob = "**/*.jsonl" if self.separate_dirs else "data.csv"
+        huggingface_hub.metadata_update(
+            repo_id=self.dataset_id,
+            repo_type="dataset",
+            metadata={
+                "configs": [
+                    {
+                        "config_name": "default",
+                        "data_files": [{"split": "train", "path": path_glob}],
+                    }
+                ]
+            },
+            overwrite=True,
+        )
 
         # Setup flagging dir
         self.components = components
@@ -284,7 +298,7 @@ class HuggingFaceDatasetSaver(FlaggingCallback):
         if self.separate_dirs:
             # JSONL files to support dataset preview on the Hub
             unique_id = str(uuid.uuid4())
-            components_dir = self.dataset_dir / str(uuid.uuid4())
+            components_dir = self.dataset_dir / unique_id
             data_file = components_dir / "metadata.jsonl"
             path_in_repo = unique_id  # upload in sub folder (safer for concurrency)
         else:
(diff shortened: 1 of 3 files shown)


gradio/flagging.py
```diff
@@ -248,6 +248,20 @@ def setup(self, components: list[IOComponent], flagging_dir: str):
             repo_type="dataset",
             exist_ok=True,
         ).repo_id
+        path_glob = "**/*.jsonl" if self.separate_dirs else "data.csv"
```
freddyaboulton, line 251: By default, dataset preview and datasets library will load all json files. This is a problem because we serialize some components to json, e.g. ChatBot and Gallery. This is so that only the jsonl files are used to compute the features. Thanks @severo and @ albertvillanova for the help.

gradio/flagging.py
```diff
@@ -416,28 +430,32 @@ def _deserialize_components(
...
-                    Path(deserialized).relative_to(self.dataset_dir)
-                ).replace(
-                    "\\", "/"
-                )
-                row.append(
-                    huggingface_hub.hf_hub_url(
-                        repo_id=self.dataset_id,
-                        filename=path_in_repo,
-                        repo_type="dataset",
+                if deserialized:
```
freddyaboulton, line 444: This was causing errors if the image was None

gradio/flagging.py
```diff
@@ -284,7 +298,7 @@ def flag(
         if self.separate_dirs:
             # JSONL files to support dataset preview on the Hub
             unique_id = str(uuid.uuid4())
-            components_dir = self.dataset_dir / str(uuid.uuid4())
+            components_dir = self.dataset_dir / unique_id
```
freddyaboulton, line 301: Before the json files included the path in the local directory, not the path in the directory of the uploaded dataset. This makes it so that they are both the same.

abidlabs:
Taking a look!

gradio/flagging.py
abidlabs: Nice

abidlabs:
Thanks for the fix @freddyaboulton! I can confirm that the flagging does not error if the Image is `None`. Generally the changes look good. 

Tested with some old demos and they continue to work. I ran your demo as well. Two things I noticed in the flagged dataset: https://huggingface.co/datasets/abidlabs/chatinterface_with_image_csv

1. When there is no image, the image string is `"None"` -- just `""` would be better
2. I don't get to see any image previews in the dataset viewer. Were you able to see image previews?

severo:
> I don't get to see any image previews in the dataset viewer. Were you able to see image previews?

to have images in the dataset viewer, the column has to use the Image type. More details here: https://huggingface.co/docs/datasets/image_dataset.

freddyaboulton:
@abidlabs - I'm not sure if our flagging format has been compatible with the expected format by datasets/datasets-server. I spent some time looking into it this morning but I think it will take a bigger effort than is worth it now given our other priorities. I will file an issue to track my thoughts.

gradio/flagging.py
```diff
@@ -416,28 +431,33 @@ def _deserialize_components(
             features[label] = {"dtype": "string", "_type": "Value"}
             try:
                 assert Path(deserialized).exists()
-                row.append(Path(deserialized).name)
+                row.append(str(Path(deserialized).relative_to(self.dataset_dir)))
             except (AssertionError, TypeError, ValueError):
-                row.append(str(deserialized))
+                deserialized = "" if deserialized is None else str(deserialized)
```
freddyaboulton, line 436: None is now "" @abidlabs

abidlabs: Cool tested and confirmed

abidlabs:
> @abidlabs - I'm not sure if our flagging format has been compatible with the expected format by datasets/datasets-server. I spent some time looking into it this morning but I think it will take a bigger effort than is worth it now given our other priorities. I will file an issue to track my thoughts.

I agree, thanks @freddyaboulton -- LGTM

abidlabs (approved):

PR #5137, opened 2023-08-08
Use font size `--text-md` for `<code>` in Chatbot messages

jaywonchung:
## Description

Some language models generate code with just triple backticks (<code>```</code>), whereas other models are capable of using triple backticks *with* the name of the language (<code>\`\`\`python</code>).

In the latter case Gradio does syntax highlighting via a different code path and sets `font-size: var(--text-md)`. However, `<code>` without highlighting is by default `--text-lg`, creating visual inconsistency when we're comparing these two cases side by side:

Screen Shot 2023-08-08 at 3 11 46 PM

After setting `font-size: var(--text-md)` for normal `<code>` elements inside `.message-wrap` it looks like this:

Screen Shot 2023-08-08 at 3 12 27 PM

PS. I believe that this is a very small change that doesn't need an issue before a PR. Please let me know otherwise.


jaywonchung:
Hmm, I'm afraid I don't know how to fix the CSS lint error 😢  Web frontend is way outside my expertise.

abidlabs:
Cool I'll go ahead and merge this in so we can get it in for the release

PR #5151, opened 2023-08-09
Concurrency count rephrase

yvrjsharma:
## Description

Rephrased messaging for gr.queue() for both, Interface and Blocks. Also, changes the concurrency_count to 1 (original value =3).

## 🎯 PRs Should Target Issues

## Tests


demo/unified_demo_text_generation/run.ipynb
```diff
@@ -1 +1,64 @@
-{"cells": [{"cell_type": "markdown", "id": 302934307671667531413257853548643485645, "metadata": {}, "source": ["# Gradio Demo: unified_demo_text_generation"]}, {"cell_type": "code", "execution_count": null, "id": 272996653310673477252411125948039410165, "metadata": {}, "outputs": [], "source": ["!pip install -q gradio torch transformers"]}, {"cell_type": "code", "execution_count": null, "id": 288918539441861185822528903084949547379, "metadata": {}, "outputs": [], "source": ["import gradio as gr\n", "from transformers import pipeline\n", "\n", "generator = pipeline('text-generation', model = 'gpt2')\n", "\n", "def generate_text(text_prompt):\n", "  response = generator(text_prompt, max_length = 30, num_return_sequences=5)\n", "  return response[0]['generated_text']\n", "\n", "textbox = gr.Textbox()\n", "\n", "demo = gr.Interface(generate_text, textbox, textbox)\n", "\n", "if __name__ == \"__main__\":\n", "    demo.launch()\n"]}], "metadata": {}, "nbformat": 4, "nbformat_minor": 5}
\ No newline at end of file
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
yvrjsharma: yes, sorry about that, there was a mistake. I've deleted that file from my PR now. It is a small change to just one file (gradio/blocks.py).Thanks for the review!

gradio/blocks.py
```diff
@@ -1637,11 +1637,11 @@ def queue(
             with gr.Blocks() as demo:
                 button = gr.Button(label="Generate Image")
                 button.click(fn=image_generator, inputs=gr.Textbox(), outputs=gr.Image())
-            demo.queue(concurrency_count=3)
+            demo.queue(concurrency_count=1)
```
abidlabs, line 1640: Since `concurrency_count=1` by default, it seems redundant to include it as a parameter. How about this instead?
```suggestion
            demo.queue(max_size=10)
```

gradio/blocks.py
```diff
@@ -1637,11 +1637,11 @@ def queue(
...
             demo.launch()
         Example: (Interface)
             demo = gr.Interface(image_generator, gr.Textbox(), gr.Image())
-            demo.queue(concurrency_count=3)
+            demo.queue(concurrency_count=1)
```
abidlabs, line 1644: Similarly, you can remove here. 
```suggestion
            demo.queue(max_size=20)
```

gradio/blocks.py
yvrjsharma: yes, I was not sure about not mentioning concurrency_count at all. This param can still be cool if there is no GPU usage, for example, a chatbot that is calling an API.

freddyaboulton: I think including it is ok to let users know that it exists. But don't feel super strongly since it's in the docstring.

abidlabs: Yup sounds good either way

abidlabs (approved):

PR #5156, opened 2023-08-09
fix changelogs

pngwn:
## Description

Closes: #(issue)

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/cyan-dolls-visit.md
+++ b/.changeset/cyan-dolls-visit.md
@@ -1,4 +1,0 @@
----
----
-
-feat:Removed out-dated comment lines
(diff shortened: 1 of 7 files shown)


pngwn:
We shouldn't need to do anything, its just because those changeset files existed before i refactored the packages, they were correct at first but became incorrect when i renamed some things. The changeset bot only uses the existing package names, so unless we make a manual mistake this won't happen again.

There is nothing wrong with `js/form` but `js/form` _used_ to be the package containing the various 'form' components. Now it is the package containing the `Form` component that we use to wrap adjacent components to create those 'form groupings'. I just changed it so the changeset appears in the correct place.

abidlabs:
Ah got it, thanks for explaining. LGTM

abidlabs (approved):

Issue #4182, opened 2023-05-12
Detect when the <gradio-app> web component it is loading/loaded on the client side.


alexcumplido:
- [ x] I have searched to see if a similar issue already exists.

When embedding the `<gradio-app>` web component isn't possible to know when the embedding it's loading or already loaded.

Use case: If I want to implement a spinner awaiting for the component to be loaded there is no event indicating such state. Would be nice be able to handle from the client side some sort of "onload" event.

The embedding includes an internal spinner but obviously it can be accessed since come from the gradio app.

In this case the web component it is embedded into a Svelte component that receives as props the space url.

`<script>
    export let src;
</script>
<gradio-app src={src}></gradio-app>`

JIBSIL:
For what it's worth I got around it with this:

``` <gradio-app id="hf-onload" src="https://your-gradio.app"></gradio-app>```

``` 
function delay(ms) { return new Promise(resolve => setTimeout(resolve, ms)) }

    async function fire() {
        await delay(100);
        const hf = document.getElementById("hf-onload");

        while (!(hf.querySelector("style") && hf.querySelector("#component-1"))) {
            await delay(200)
        }

        // you can reference components like this
        let component1 = hf.querySelector("#component-1")
    }
    document.addEventListener("DOMContentLoaded", async function () { await fire() }
)```

It just sleeps until it detects that the gradio app has been loaded and allows you to reference its DOM

PR #5158, opened 2023-08-09
Add `render` function to `<gradio-app>`

hannahblair:
## Description

Adds a `render` event to the `<gradio-app>` web component. 

To test I got the website running and added an `on:render` event to the <gradio-app> line in `Demos.svelte`

```
{#key name}
	<gradio-app
		on:render={() => console.log("loaded" + name)}
		space={"gradio/" + name}
	/>
{/key}
```

Closes: #4182 

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/stale-houses-grin.md
+++ b/.changeset/stale-houses-grin.md
@@ -7,7 +7,7 @@ highlight:
 
 #### Add `render` function to `<gradio-app>`
 
-We now have an event `render` on the <gradio-app> web component, which is triggered once the embedded space has finished loading.
+We now have an event `render` on the <gradio-app> web component, which is triggered once the embedded space has finished rendering.
 
 ```html
 <script>


pngwn:
Could we use an event for this, rather than an attribute? Svelte has some special handling for attributes vs properties but in vanilla JS users could run into issue because attributes are always strings, so the function passed down may need to be evaled in order to run correctly in certain scenarios.

Events won't have this issue. I think we have one example of the web component emitting an event in the codebase. Happy to sync if necessary!

js/app/src/main.ts
```diff
@@ -61,20 +61,26 @@ function create_custom_element(): void {
...
+				bubbles: true,
+				cancelable: false,
+				composed: true
+			});
+
+
 			const observer = new MutationObserver((mutations) => {
-				this.dispatchEvent(event);
+				this.dispatchEvent(domChangeEvent);
+				this.dispatchEvent(readyEvent);
```
pngwn, line 79: We don't want to put this here, this is specific to the DOM changing. We instead want to listen to the `Index` component' `ready` event and dispatch the custom event when it is fired.

Something like 

```js
this.app.$on('ready', () => {
  this.dispatchEvent(...)
)}
  ```

**Comment by aliabid94** [comment]
Nice. How is this used outside of svelte, when gradio-app is embedded as a regular web component? Can we add documentation about this to the sharing_your_app guide, in the Embedding with Web Components section?

File: .changeset/stale-houses-grin.md [hunk 2]
```diff
@@ -0,0 +1,6 @@
+---
+"@gradio/app": minor
+"gradio": minor
+---
+
+feat:Add `loadcomplete` function to `<gradio-app>`
```
freddyaboulton, line 6: I would make this a highlight and add the code snippet from the guide. Looks really cool!

js/app/src/Blocks.svelte
pngwn, line 1: Change `handle_mount` on line 500 to accept the id we are already passing it:

```js
async function handle_mount()
```

becomes

```js
async function handle_mount({ detail })
```

`detail` should contain the id we care about now.

js/app/src/Blocks.svelte
```diff
@@ -45,6 +45,9 @@
 
 	let loading_status = create_loading_status_store();
 
+	const rendered_node_ids = new Set();
```
pngwn, line 48: Maybe we can change this to `walked_node_ids` or something. `render` and `mount` mean the same thing to me.

js/app/src/Blocks.svelte
```diff
@@ -552,6 +561,8 @@
 		});
 
 		components.forEach((c) => {
+			mounted_node_ids.add(c.id);
```
pngwn, line 564: I think this got missed when we changed the position.

guides/01_getting-started/03_sharing-your-app.md
```diff
@@ -119,6 +120,19 @@ Here's an example of how to use these attributes to create a Gradio app that doe
 ></gradio-app>
 ```
 
+Here's another example of how to use the `render` event. An event listener is used to capture the `render` event and will call the `handleLoadComplete()` function once rendering is complete. 
+
+```html
+<script>
+	function handleLoadComplete() {
+		console.log("Embedded space has finished loading");
```
pngwn, line 128:
```suggestion
		console.log("Embedded space has finished rendering");
```

.changeset/stale-houses-grin.md
```diff
@@ -0,0 +1,20 @@
...
+highlight:
+
+#### Add `render` function to `<gradio-app>`
+
+We now have an event `render` on the <gradio-app> web component, which is triggered once the embedded space has finished loading.
+
+```html
+<script>
+	function handleLoadComplete() {
+		console.log("Embedded space has finished loading");
```
pngwn, line 15:
```suggestion
		console.log("Embedded space has finished rendering");
```

pngwn (approved):
Small tweaks but looks good otherwise and is definitely working! Thanks for adding this @hannahblair and sorry for all of the back and forth!

Issue #5147, opened 2023-08-09
Inline images in a chatbot don't get shared while using the share_button


yvrjsharma:
### Describe the bug

When embedding inline images in a chatbot using the following methods, the images don't get shared or displayed as expected while using the share_button functionality.

1. Base64 encoded string and Markdown, and  
2. Temporary filepath in the markdown

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

**Steps to Reproduce:**

- For 1 :

```
import gradio as gr
import base64

def image_to_base64(filepath):
    with open(filepath, 'rb') as image_file:
        encoded_string = base64.b64encode(image_file.read())
    return encoded_string.decode('utf-8')

def generate(x):
    #print(x)
    b64 = image_to_base64(x)
    data_url = f"data:image/jpeg;base64,{b64}"
    return [
        (f"abc ![]({data_url})", "def"),
        ("ghi", "jkl")]


with gr.Blocks() as demo:
    i = gr.Image(type="filepath")
    c = gr.Chatbot()
    i.change(generate, i, c)

demo.launch(debug=True)
```

- For 2:

```
import gradio as gr
import base64

def generate(x):
    #print(x)
    return [
        (f"abc ![]({data_url})", "def"),
        ("ghi", "jkl")]


with gr.Blocks() as demo:
    i = gr.Image(type="filepath")
    c = gr.Chatbot()
    i.change(generate, i, c)

demo.launch(debug=True)
```

- Use the share button to share the chatbot conversation.

### Screenshot

- For temporary files - 

https://github.com/gradio-app/gradio/assets/48665385/8d5dee19-f63f-455f-b41a-a0af87bc39cb

- For base64 

https://github.com/gradio-app/gradio/assets/48665385/46cc4988-4ef4-4150-b2a1-97d72a26219b

### System Info

```shell
Huggingface Spaces; Gradio 3.39.0
```

### Severity

Blocking usage of gradio

PR #5159, opened 2023-08-09
Share all media content in chat

aliabid94:
Use regex to parse through chat messages on share, find any inline file content, upload to HF, and replace the URL. Fixes: #5147 

Test this on Space with:
```python
import gradio as gr

def generate(x):
    return [
        (f"abc ![](/file={x})", "def"),
        ("ghi", "jkl")]


with gr.Blocks() as demo:
    i = gr.Image(type="filepath")
    c = gr.Chatbot()
    i.change(generate, i, c)

demo.launch(debug=True)
```


abidlabs:
Thanks @aliabid94 for jumping on this so fast! Maybe we can get this out in the release as well

abidlabs:
Beautiful @aliabid94! Works great in my test Space: https://huggingface.co/spaces/abidlabs/test-chat-share

abidlabs (approved):

abidlabs:
cc @yvrjsharma

PR #5173, opened 2023-08-10
Ensure gradio client works as expected for functions that return nothing.

raymondtri:
## Description

I ran into this issue with h2ogpt, but it seems like the problem is with the gradio client app. This could be an issue with my lack of python proficiency, but this fix does get it working.

Closes: #641

## 🎯 PRs Should Target Issues

I figured that creating an issue in both repositories would be redundant, please correct me if I am wrong


pngwn (approved):
Looks great, thanks for taking care of this @raymondtri.

I really didn't think the array was optional, i thought it was an empty array if it was empty but this is a tidy fix!

PR #5175, opened 2023-08-10
fix toml

abidlabs:
To fix the error here: https://github.com/gradio-app/gradio/actions/runs/5826030686/job/15799036715


freddyaboulton (approved):
Thanks @abidlabs !!

PR #5177, opened 2023-08-10
chore: update versions

pngwn:
This PR was opened by the [Changesets release](https://github.com/changesets/action) GitHub action. When you're ready to do a release, you can merge this and the packages will be published to npm automatically. If you're not ready to do a release yet, that's fine, whenever you add more changesets to main, this PR will be updated.

# Releases

> The changelog information of each package has been omitted from this message, as the content exceeds the size limit.


freddyaboulton:
Everything looks good to me except the following two points!

I'm having a hard time using the dropdowns. Looks like it collapses on itself.
dropdown_selection

The bottom corners of the output file component in zip_demos is cut off. Not sure if that's an old thing.

image

abidlabs:
> I'm having a hard time using the dropdowns. Looks like it collapses on itself.

Yeah I just noticed this as well. The issue is that if a value is already in the dropdown, the dropdown filters by that value. You can backspace to get the full list of options. Was this an intentional change?

pngwn:
@abidlabs Not at all, no.

Issue #5168, opened 2023-08-10
Audio streaming still not working properly with latest fixes


snarb:
### Describe the bug

I have tested Audio streaming with latest changes  Live audio streaming output https://github.com/gradio-app/gradio/pull/5077. I have make a build from the repository.  demo\stream_audio_out is working. But still there are couple of issues:
1) 
```
import gradio as gr
from pydub import AudioSegment

def stream_audio(audio_file):
    for i in range(3):
        print(i)
        yield r'Recording.mp3'

demo = gr.Interface(
    fn=stream_audio,
    inputs=gr.Audio(type="filepath", label="Audio file to stream"),
    outputs=gr.Audio(autoplay=True, streaming=True),
)

if __name__ == "__main__":
    demo.queue().launch()

```

It will play/stream the Recording.mp3 2 times instead of 3.

2) 
```
import gradio as gr
from pydub import AudioSegment

def stream_audio(audio_file):
    for i in range(3):
        print(i)
        yield r'Recording.wav'

demo = gr.Interface(
    fn=stream_audio,
    inputs=gr.Audio(type="filepath", label="Audio file to stream"),
    outputs=gr.Audio(autoplay=True, streaming=True),
)

if __name__ == "__main__":
    demo.queue().launch()

```
Wav files are played only single time. Looks like streaming is broken for them.

3) audio file that is passed inside the stream audio is a cropped version of the file we chose with 100 ms step. Don't know is it by design or not. More looks like a bug for me because _streaming_ is not set to _True_ for the inputs.

4) For a list of mp3 files looks like it skips the first one.
@aliabid94

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

run the code

### System Info

```shell
windows 10
```

### Severity

Can find workaround

freddyaboulton:
Seems related to #5160

aliabid94:
looking into this

Issue #5160, opened 2023-08-09
Streaming Audio Output latency / gap issues


freddyaboulton:
### Describe the bug

I'm running `stream_audio_output` and when I run with a file shorter than 3 seconds, no audio gets streamed to the output component. If I use a file with more than 3 seconds, the output skips the first three seconds.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr
from pydub import AudioSegment

def stream_audio(audio_file):
    audio = AudioSegment.from_mp3(audio_file)
    i = 0
    chunk_size = 3000
    
    while chunk_size*i < len(audio):
        chunk = audio[chunk_size*i:chunk_size*(i+1)]
        i += 1
        if chunk:
            file = f"/tmp/{i}.mp3"
            chunk.export(file, format="mp3")            
            yield file
        
demo = gr.Interface(
    fn=stream_audio,
    inputs=gr.Audio(type="filepath", label="Audio file to stream"),
    outputs=gr.Audio(autoplay=True, streaming=True),
)

if __name__ == "__main__":
    demo.queue().launch()
```

with this file, 8 out of the first three seconds play. The first three seconds are missing.
https://github.com/gradio-app/gradio/assets/41651716/48415057-f79c-4557-bea8-11779b9f2f03

<img width="1731" alt="image" src="https://github.com/gradio-app/gradio/assets/41651716/dfe4e492-bb73-4b0c-a170-bfb595a1968c">

With cantina.wav, no audio is streamed
<img width="1615" alt="image" src="https://github.com/gradio-app/gradio/assets/41651716/4c8bbecf-5a1c-4fc0-b5ab-a92f364caa4c">

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Darwin
gradio version: 3.39.0
gradio_client version: 0.3.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
aiohttp: 3.8.5
altair: 4.2.0
fastapi: 0.87.0
ffmpy: 0.3.1
gradio-client==0.3.0 is not installed.
httpx: 0.23.0
huggingface-hub: 0.14.0
importlib-resources: 6.0.1
jinja2: 3.1.2
markdown-it-py: 2.2.0
markupsafe: 2.1.1
matplotlib: 3.7.2
mdit-py-plugins: 0.3.3
numpy: 1.21.6
orjson: 3.9.4
packaging: 22.0
[...]

PR #5179, opened 2023-08-10
Fixes audio streaming issues

aliabid94:
Two issues fixed (linked):
1. Was originally skipping first chunk yielded
2. Streaming really only works with mp3 types so convert wav files to mp3 if streaming.

Fixes: #5168


gradio/components/audio.py
```diff
@@ -343,11 +343,20 @@ def postprocess(
...
+                sample_rate,
+                format="mp3" if self.streaming else self.format,
             )
             self.temp_files.add(file_path)
         else:
-            file_path = self.make_temp_copy_if_needed(y)
+            if isinstance(y, Path):
+                y = str(y)
+            if self.streaming and not y.endswith(".mp3"):
+                sample_rate, data = processing_utils.audio_from_file(y)
```
freddyaboulton, line 355: What's the issue with wav? Do you have an example where it fails? cantina.wav works fine for me without this.

gradio/blocks.py
```diff
@@ -1354,7 +1359,10 @@ def handle_streaming_outputs(
         for i, output_id in enumerate(self.dependencies[fn_index]["outputs"]):
             block = self.blocks[output_id]
             if isinstance(block, StreamableOutput) and block.streaming:
-                stream = block.stream_output(data[i])
+                if end:
+                    stream = None
+                else:
+                    stream = block.stream_output(data[i])
```
freddyaboulton, line 1365: Why is this needed, isn't the end result of the iterator going to be None anyways?

freddyaboulton (approved):
Looks good @aliabid94 ! Thanks for the fix. Left some questions but not blocking merge.

gradio/components/audio.py
aliabid94: really? cantina.wav specifically does not work for me on chrome. It may be differ browser to browser - the problem with streaming wav's is that wav files have a header in each chunk that can confuse the browser when they are naively chained to gether as a stream - mp3's do not have this problem. As explained to me by chatgpt here: https://chat.openai.com/share/dc4b8dd7-faf3-43f6-98df-ea8d7da23fcc

gradio/blocks.py
aliabid94: you're right! will remove

abidlabs:
I'll bump the version and do the release manually -- I won't merge this PR because we should sort out the changeset issue first

snarb:
I don't know gradio very well,  but want to share a general cocern, that reading and writing stream chunks to files microphone/local wav/100 Ms chunks/wav in temp folder/mp3/memory looks problematic for real time streaming to me . Disk Io operations  especially under heavy load are not very predictable and stable from the timing perspective. Something like webrtc looks more promising and easy to maintain as for me

freddyaboulton:
> that reading and writing stream chunks to files

Maybe we can add an option to return bytes directly from your function

abidlabs:
Let's go ahead and merge this in since we've already released this. But I'll leave #5160 open since there's some outstanding issues related to latency

PR #5183, opened 2023-08-11
Fix Add row_count and col_count to Dataframe.update

### LINKED ISSUE #5182: Gradio.Dataframe: Expose an option to hide the "Add Row" button

I want to fill a Gradio dataframe with the output of my machine learning model. For that purpose, it does not make sense for users to extend the table by themselves, so I would like to get rid of the "Add Row" button. I can make the "Add Column" button disappear by specifying a `header` and limiting the columns (`col_count=(5, 'fixed')`). However, as my model creates a varying number of rows, I cannot hide the "Add Row" button that way. If I try limiting the number of rows, the table is truncated in cases where my models produces more entries.

I would like to hide the "Add Row" button explicitly via an instantiation parameter in `Gradio.Dataframe`. For example `hide_add_row` could be set to `True` in order to hide the button.

**Additional context**  

My current workaround is to hide the button using CSS like so:

```css
div > div > div.controls-wrap > span > button {
    display: none !important;
}
```

That feels sketchy since there are few CSS classes and thus, this might match other elements, too.

freddyaboulton:
## Description

Fixes #5182

As discussed in #5066, we can't get rid of the update method yet for backwards compatibility so we still need to add the correct parameters to the function signature.

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/gradio/components/dataframe.py
+++ b/gradio/components/dataframe.py
@@ -11,6 +11,7 @@ from gradio_client.serializing import JSONSerializable
 
 from gradio import utils
 from gradio.components.base import IOComponent, _Keywords
+from gradio.deprecation import warn_deprecation
 from gradio.events import (
     Changeable,
     EventListenerMethod,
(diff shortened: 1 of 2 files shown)


freddyaboulton:
We can't get rid of update yet in a backwards compatible way because you can call update without kwargs. See this comment: https://github.com/gradio-app/gradio/pull/5066#discussion_r1286008103

abidlabs:
> We can't get rid of update yet in a backwards compatible way because you can call update without kwargs. See this comment: https://github.com/gradio-app/gradio/pull/5066#discussion_r1286008103

Yup I know I'm not suggesting we remove `.update()` but I'm saying that instead of adding new parameters to `.update()` going forward, we should instead tell users to use the `return Dataframe` syntax which will already support all of the parameters

freddyaboulton:
Misunderstood your comment! My initial thought was that even though it was deprecated it should still work but this doesn't work at all right now. Let's close. Will be fixed in the next release hopefully.

PR #5186, opened 2023-08-11
homepage demo update

freddyaboulton:
## Description

Last night I noticed that the demos on the landing page don't reflect the latest "gen ai boom" so I added the llama2 ChatInterface and a Stable Diffusion demo.

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/js/_website/src/lib/components/DemosLanding.svelte
+++ b/js/_website/src/lib/components/DemosLanding.svelte
@@ -3,20 +3,18 @@
 
 	let tabs = [
 		{
-			title: "Chat with Llama 2",
-			code: chat,
-			demo: "ysharma/Explore_llamav2_with_TGI"
+			title: "Sketch Recognition",
+		  	code: sketch,
+			demo: "gradio/pictionary" },
+		{
+			title: "Image Segmentation",
+			code: img,
+			demo: "gradio/Echocardiogram-Segmentation"
 		},
 		{
 			title: "Stable Diffusion",
 			code: stable_diffusion,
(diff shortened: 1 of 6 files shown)


aliabid94:
Awesome! I do think the order of these should introduce Interface, then Blocks, then ChatInterface. 

I also think we can drop or swap out the echocardiogram one too possibly. Also the Stable Diffusion one's layout looks weird, lemme see if I can fix hysts space

freddyaboulton:
> I do think the order of these should introduce Interface, then Blocks, then ChatInterface.

Completely agreed!

aliabid94:
@hysts https://huggingface.co/spaces/hysts/SD-XL/discussions/15

js/_website/src/lib/components/DemosLanding.svelte
```diff
@@ -1,18 +1,21 @@
 <script lang="ts">
-	import { q_a, img, sketch } from "../assets/demo_code";
+	import { stable_diffusion, img, sketch, chat } from "../assets/demo_code";
```
freddyaboulton, line 2: Much easier to update this than the raw html. Nice work @aliabd @pngwn

aliabd (approved):
Code looks good! I agree with @aliabid94 that we should probably kill Echocardiogram

pngwn (approved):
Looks good! Thanks for taking care of this @freddyaboulton!

freddyaboulton:
Thank you all for the input!!

PR #5188, opened 2023-08-11
Fix the images in the theme builder to use permanent URI

abidlabs:
The header image in the theme builder was referencing a file from our GitHub repo which was moved due to recent refactoring. I've uploaded the image to the S3 bucket and updated the references.


freddyaboulton (approved):
Thanks @abidlabs !

abidlabs:
Thanks @freddyaboulton!

PR #5215, opened 2023-08-14
Lazy load interactive or static variants of a component individually, rather than loading both variants regardless. This change will improve performance for many applications.

pngwn:
## Description

Lazy load only the interactive or static variant of a component as required rather than loading both. This needs a follow up or two but works well.

Closes: #3360

## 🎯 PRs Should Target Issues

## Tests


pngwn:
@abidlabs READY FOR REVIEW!!!

dawoodkhan82:
@pngwn Don't think updating `interactive=False` is working properly. I tried modifying `hello_blocks` for example: 

```
import gradio as gr

def greet(name):
    return "Hello " + name + "!", gr.Textbox.update("sorry", interactive=False)

with gr.Blocks() as demo:
    greet_btn.click(fn=greet, inputs=[name], outputs=[output, name], api_name="greet")

if __name__ == "__main__":
    demo.launch()
```

aliabid94:
> Don't think updating interactive=False is working properly

I think this is because now we're loading the appropriate interactive/static version once at load time, right? 

One issue I noticed is that even as an output component, the dataframe component still allows for editing the component. Seems like this issue was introduced before this PR though. 

Otherwise looks great!

pngwn:
> I think this is because now we're loading the appropriate interactive/static version once at load time, right?

I added some logic to load new components (if necessary when `interactive` changes), it just wasn't working all of the time, cos i messed up the order.

@dawoodkhan82 it should work now, if you can retest.

@aliabid94 I think i broke the dataframe in my refactor. Fixed that here as well.

dawoodkhan82:
@pngwn tested and it works great now!

dawoodkhan82 (approved):

PR #5222, opened 2023-08-14
pin local dependencies when auto-bumping

pngwn:
related to #5061.


abidlabs (approved):
Sweet!

freddyaboulton (approved):
Thank you @pngwn !

Issue #5224, opened 2023-08-15
Interface.from_pipeline() does not work with Blocks()


Welkin-Y:
### Describe the bug

Using Interface.from_pipeline() with Blocks() will trigger DuplicateBlockError: A block with id: 2 has already been rendered in the current Blocks.

However, code below will work:
```
import gradio as gr

with gr.Blocks() as demo:
    with gr.Tab("Image Inference"):
        gr.Interface(lambda x:x, "image", "image")
demo.launch()
```

After checking the Block ids I found that blocks created by Interface.from_pipeline() method are NOT counted in the Blocks() class.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```
import gradio as gr
from transformers import pipeline
classifier = pipeline(task="image-classification")

with gr.Blocks() as demo:
    with gr.Tab("Image Inference"):
        gr.Interface.from_pipeline(classifier)
demo.launch()
```

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Linux
gradio version: 3.40.1
gradio_client version: 0.4.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
aiohttp: 3.8.4
altair: 5.0.1
fastapi: 0.101.0
ffmpy: 0.3.1
gradio-client: 0.4.0
httpx: 0.24.1
huggingface-hub: 0.14.1
importlib-resources: 6.0.1
jinja2: 3.1.2
markdown-it-py: 2.2.0
markupsafe: 2.1.2
matplotlib: 3.7.1
mdit-py-plugins: 0.3.3
numpy: 1.24.3
orjson: 3.9.4
packaging: 23.1
pandas: 1.5.2
pillow: 9.5.0
pydantic: 2.1.1
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0
requests: 2.30.0
semantic-version: 2.10.0
typing-extensions: 4.7.1
uvicorn: 0.23.2
websockets: 11.0.3
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2023.5.0
httpx: 0.24.1
huggingface-hub: 0.14.1
packaging: 23.1
requests: 2.30.0
typing-extensions: 4.7.1
websockets: 11.0.3
```

### Severity

I can work around it

abidlabs:
[...]

PR #5231, opened 2023-08-15
Allow `gr.Interface.from_pipeline()` and `gr.load()` to work within `gr.Blocks()`

abidlabs:
We were prematurely rendering components when using `Interface.from_pipeline()` as well as using `gr.load()` which meant that you could not use those functions inside of a `Blocks()`. This fixes those issues and adds a couple of unit tests.
 
Fixes: #5224

Repros:

```py
import gradio as gr
classifier = pipeline(task="image-classification")

with gr.Blocks() as demo:
    with gr.Tab("Image Inference"):
        gr.Interface.from_pipeline(classifier)
demo.launch()
```

```py
import gradio as gr

with gr.Blocks() as demo:
    gr.load("models/google/vit-base-patch16-224")
demo.launch()
```


aliabid94 (approved):
LGTM

test/test_pipelines.py
```diff
@@ -5,9 +5,20 @@
...
+@pytest.mark.flaky
+def test_interface_in_blocks():
+    pipe1 = transformers.pipeline(model="sshleifer/bart-tiny-random")
+    pipe2 = transformers.pipeline(model="sshleifer/bart-tiny-random")
+    with gr.Blocks() as demo:
+        with gr.Tab("Image Inference"):
+            gr.Interface.from_pipeline(pipe1)
+        with gr.Tab("Image Inference"):
+            gr.Interface.from_pipeline(pipe2)
+    demo.launch(prevent_thread_lock=True)
```
freddyaboulton, line 24: Let's close the demo when we're done with the test

freddyaboulton (approved):
Thanks @abidlabs !

PR #5232, opened 2023-08-15
`gr.Radio` and `gr.CheckboxGroup` can now accept different names and values

### LINKED ISSUE #4754: `gr.Radio` and `gr.CheckboxGroup` choices display text.

Usually for the `gr.Radio` and `gr.CheckboxGroup`, the choices are  different for the program to process and UI display to the user. For example, in a program, it might be easier to have choices named like ["openai_completion", "openai_chat"], but it would be great to have the UI display as ["OpenAI (Completion)", "OpenAI (Chat)"].

 It would be great if there could be an additional argument 
 e.g. `choices_display` for  `gr.Radio` and `gr.CheckboxGroup` that can be used specifically for UI display.

**Additional context**

abidlabs:
The `choices` parameter in `gr.Radio` and `gr.CheckboxGroup` can now accept tuples in the form `(name, value)` where the `name` is the string that is displayed in the frontend while `value` is the string or numeric that is passed into the backend. This answers a long-standing request, closes #4754 

Adds stories for both components. Also fixes some unrelated tests that were failing for me locally on `main` as well (see comments on the PR). 

Test with:

```py
import gradio as gr

with gr.Blocks() as demo:
    c = gr.CheckboxGroup(["abc", ("def", "long_name")])
    t = gr.Textbox()
    t.change(lambda x:x, t, c)
    
demo.launch()
```

There were many ways this could have been implemented -- I chose this way to be backwards compatible when loading from configs, which means that we can still `gr.load` Spaces that use radio/checkboxgroup, e.g.

```py
import gradio as gr

gr.load("spaces/abidlabs/titanic-survival").launch()
```

Next up, doing the same for `gr.Dropdown()` and `gr.Examples`!


test/test_components.py
```diff
@@ -2370,13 +2369,6 @@ def test_no_color(self):
         assert config["encoding"]["x"]["field"] == "Horsepower"
         assert config["encoding"]["x"]["title"] == "Horse"
         assert config["encoding"]["y"]["field"] == "Miles_per_Gallon"
-        assert config["selection"] == {
```
abidlabs, line 2373: This part of the config does not always appear, I think depending on what version of altair you are using. I think its okay to leave out.

test/test_components.py
```diff
@@ -697,7 +697,6 @@ def test_component_functions(self):
         with pytest.raises(ValueError):
             gr.Image(type="unknown")
         image_input.shape = (30, 10)
-        assert image_input._segment_by_slic(img) is not None
```
abidlabs, line 700: This was failing for me locally. I spent some time investigating but couldn't figure out the reason (it relates to image modes). But since we are planning on deprecating interpretation anyways, I think its okay to remove.

gradio/components/checkboxgroup.py
```diff
@@ -109,8 +113,8 @@ def get_config(self):
 
     def example_inputs(self) -> dict[str, Any]:
         return {
-            "raw": self.choices[0] if self.choices else None,
-            "serialized": self.choices[0] if self.choices else None,
+            "raw": self.choices[0][1] if self.choices else None,
```
aliabd, line 116: this is actually incorrect and should be 
`[self.choices[0][1]] if self.choices else None` for both raw and serialized (its also incorrect in main). Should be this way, similar to Dropdown multiselect, otherwise the code snippets in api docs don't work as expected.

gradio/components/radio.py
```diff
@@ -55,22 +55,26 @@ def __init__(
     ):
         """
         Parameters:
-            choices: list of options to select from.
-            value: the button selected by default. If None, no button is selected by default. If callable, the function will be called whenever the app loads to set the initial value of the component.
+            choices: A list of string or numeric options to select from. An option also be a tuple of the form (name, value), where name is the displayed name of the radio button and value is the value to be passed to the function, or returned by the function.
```
aliabd, line 58:
```suggestion
            choices: A list of string or numeric options to select from. An option can also be a tuple of the form (name, value), where name is the displayed name of the radio button and value is the value to be passed to the function, or returned by the function.
```

gradio/components/checkboxgroup.py
```diff
@@ -35,7 +35,7 @@ class CheckboxGroup(
 
     def __init__(
         self,
-        choices: list[str | float | int] | None = None,
+        choices: list[str | int | float | tuple[str, str | int | float]] | None = None,
         *,
         value: list[str | float | int] | str | float | int | Callable | None = None,
```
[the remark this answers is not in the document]
freddyaboulton: I think it should just be a str | number, which is what the backend logic currently expects.

js/checkboxgroup/index.svelte
```diff
@@ -9,7 +9,7 @@
 	export let visible = true;
 	export let value: string[] = [];
 	export let value_is_output = false;
-	export let choices: string[];
+	export let choices: string[][];
```
freddyaboulton, line 12: I think this would be a better type to indicate it can only be tuples:  `[string, string | number][]`

freddyaboulton (commented):
Looks good @abidlabs! Noticed one comment about the examples.

abidlabs:
Thanks guys! Will address the reviews tomorrow

gradio/components/checkboxgroup.py
abidlabs: Yeah you're right

js/checkboxgroup/index.svelte
abidlabs: Nice thanks

gradio/components/radio.py
```diff
@@ -150,22 +154,22 @@ def preprocess(self, x: str | int | float | None) -> str | int | float | None:
...
+            value of the selected choice as string or index within choice list
         """
         if self.type == "value":
             return x
         elif self.type == "index":
             if x is None:
                 return None
             else:
-                return self.choices.index(x)
+                return [value for _, value in self.choices].index(x)
```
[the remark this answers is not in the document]
abidlabs: Makes sense!

abidlabs:
Addressed all of the feedback, thanks again @freddyaboulton and @aliabd. Will merge in a bit if there are no other suggestions.

PR #5238, opened 2023-08-15
Improve audio streaming

aliabid94:
Fix a lot of audio issues with streaming:
- Wav file streaming should work flawlessly (though wav files are huge and not ideal for streaming)
- Mp3 file streaming still has clicks at the seams between chunks - very difficult to remove this due to  how mp3s are encoded
- Byte streaming now supported, pretty flawless

Fixes: #5160 

Updated streaming_audio_out to test all options


pngwn:
>Fix a lot of audio issues with streaming:
>
> - Wav file streaming should work flawlessly (though wav files are huge and not ideal for streaming)
> - Mp3 file streaming still has clicks at the seams between chunks - very difficult to remove this due to how mp3s are encoded
> - Byte streaming now supported, pretty flawless

Could we add this description to the changeset, instead of just "improve audio streaming"? It will make for a nicer changelog. It might be easier to edit the changeset manually for this one.

gradio/components/audio.py
```diff
@@ -328,8 +328,8 @@ def get_masked_inputs(self, tokens, binary_mask_matrix):
         return masked_inputs
 
     def postprocess(
-        self, y: tuple[int, np.ndarray] | str | Path | None
-    ) -> str | dict | None:
+        self, y: tuple[int, np.ndarray] | str | Path | bytes | None
+    ) -> str | dict | bytes | None:
```
abidlabs, line 332: Nice! We should add to the docstring of `Audio` that the returned audio can be in `bytes` format. (Perhaps we even mention that this is the recommended format when streaming). Also, let's add `stream_audio_out` to the list of demos

abidlabs:
Tested with with different file formats and everything lgtm. The bytes format is especially smooth 🤌

abidlabs (approved):

pngwn:
Should we just make this PR a `highlight`? i think the changelog rendering will be weird as it spans many lines.

abidlabs:
@freddyaboulton brought up a good point here that we should address to get a good experience on Spaces: https://huggingface.slack.com/archives/C02SPHC1KD1/p1692210607417629

PR #5240, opened 2023-08-15
Removing .update and get_config, attempt 2

aliabid94:
Trying a different approach (much simpler!!!) to removing get_config and update.

get_config is removed, the config used is simply any attribute that is in the Block that shares a name with one of the constructor paramaters.

update is not removed for backwards compatibility, but deprecated. Instead return the component itself. Created a `updateable` decorator that simply checks to see if we're in an update, and if so, skips the constructor and wraps the args and kwargs in an update dictionary. easy peasy.


freddyaboulton (commented):
The new update logic is awesome. Definitely much simpler and I think will work well with custom components.

I see a shortcoming with the proposed get_config change in that it won't get the props from a parent class. For example, Chatbot currently doesn't list the selectable prop in the config but before it did. Also there are some component props that are not listed in the init constructor like bokeh_version for the plots. Will think about this today.

gradio/blocks.py
```diff
@@ -2241,3 +2246,29 @@ def queue_enabled_for_fn(self, fn_index: int):
...
         return self.dependencies[fn_index]["queue"]
+
+
+def is_update():
+    from gradio import context
+
+    return hasattr(context.thread_data, "blocks")
+
+
+def updateable(fn):
```
freddyaboulton, line 2257: Maybe we can automatically wrap the init with updateable in a metaclass so that custom component authors don't even have to worry about this when creating their own components.

abidlabs:
> I see a shortcoming with the proposed get_config change in that it won't get the props from a parent class. For example, Chatbot currently doesn't list the selectable prop in the config but before it did. Also there are some component props that are not listed in the init constructor like bokeh_version for the plots. Will think about this today.

Good point. Just to add -- this is actually a potential bug currently in `gradio` because we use the properties in the config to recreate components when using `gr.load()` with Spaces. So right now, if you were to `gr.load()` a Space that has a `gr.Plot`, it would raise an unused kwarg warning because it would try to pass in `bokeh_version`. We're planning on turning these warnings into exceptions in 4.0, so it could be problematic independent of this PR.

abidlabs:
Agree with @freddyaboulton's suggestions. Tested `.update()` out and just wanted to confirm a few things:

1.

 > update is not removed for backwards compatibility, but deprecated

I don't think its been deprecated yet just fyi. I don't see any warnings when using the Component.update() methods

2.

In your previous PR, you had implemented:

> Any argument that is in the constructor that becomes a component property will be considered part of the config and sent to the backend (open to other logic here)

Sending configs to the backend so that they affect the preprocessing/postprocessing is not part of this PR, right? E.g. 

```py
import gradio as gr

def update_image_type():
    print("updated")
    return gr.Image(type="filepath")

with gr.Blocks() as demo:
    i = gr.Image()
    t = gr.Textbox()
    btn = gr.Button()
    btn.click(update_image_type, None, i)
    i.change(lambda x:x, i, t)
    
demo.launch()
```

does not work

gradio/blocks.py
aliabid94: was trying to think how to do this. Not familiar with metaclasses, will take a look!

aliabid94:
> I see a shortcoming with the proposed get_config change in that it won't get the props from a parent class. For example, Chatbot currently doesn't list the selectable prop in the config but before it did. Also there are some component props that are not listed in the init constructor like bokeh_version for the plots. Will think about this today.

We could still have components implement their own get_configs, which return the extra parameters plus a call to super()? Something like

```python
def get_config(self):
  return {
    "selectable": self.selectable,
    ...super().get_config()
  }

> I don't think its been deprecated yet just fyi.

Yes sorry not planning on making the deprecation as part of this PR because of the next comment

> Just to confirm, sending configs to the backend so that they affect the preprocessing/postprocessing is not part of this PR, right? 

Yep, I'll do this in a subsequent PR, trying to keep things smaller and more testable. So for that reason, not deprecating or even adding this to the guide yet.

**Comment by freddyaboulton** [comment]
> We could still have components implement their own get_configs, which return the extra parameters plus a call to super()?

Yea I think that makes sense! Although in the specific case of selectable, I think it should be added to the config automatically if the component is selectable without users having to worry about it.

File: .changeset/many-tips-create.md
Comment (pngwn) [inline .changeset/many-tips-create.md:1]: I need to make the changeset thing support multiline comments for fix + feat. I'll do that tomorrow, should be quick.

**Comment by aliabid94** [comment]
> For example, Chatbot currently doesn't list the selectable prop in the config but before it did. Also there are some component props that are not listed in the init constructor like bokeh_version for the plots

Overrode get_config in these two cases to restore these config variables. Scanning through the rest of the components I couldn't find any other cases where the config had additional fields.

**Comment by aliabid94** [comment]
@abidlabs @freddyaboulton this is ready for re-review! Added support for:
- updating backend variables, which are stored in a new StateHolder object. So we now support backend updates.
- allow controlling memory used to store state/backend updates by setting `state_session_capacity`, which will drop "least recently used" sessions. Defaults to 10000 sessions in memory.

**Comment by abidlabs** [comment]
Very nice @aliabid94! I tested a lot . Let me post issues here as I notice them.

Nits:

1. gr.update(), gr.Textbox.update(), etc. don't raise deprecation warnings. Also if you return a `gr.Textbox()` to e.g. an Audio component, no error or warning is raised. Or if you pass in non-existent parameters to `gr.Textbox()` as you return it, no errors are raised, which can make it hard to debug the issue. 

2. My understanding is that when there are too many sessions, the oldest session will be popped off, and if that user tries to update the session state, a new session will be created using the original values of the Blocks. Perhaps a `gr.Warning` should be raised letting the user know that this has happened? 

3. Can we add unit tests for `StateHolder` and `SessionState`? It would be great to add an e2e test using `blocks_essay` as well

4. Can we improve the changelog entry (make it a highlight)

Bugs:

Haven't found any yet! Nice job.

I'll keep looking :)

As discussed, let's merge this in after we've done a release

**Comment by aliabd** [comment]
@aliabid94 

- since we're deprecating `.update` can you change the demos that use it? It's still in many of them. 
- can you change the docstring on [update](https://github.com/gradio-app/gradio/blob/0159898edacd5500d539586967e3ce857bc71ae4/gradio/helpers.py#L788), we pick that up on the docs [here](https://gradio-pfh3ov6yb-hugging-face.vercel.app/docs/update).

**Comment by abidlabs** [comment]
Did some more testing, and finally found an issue 🙃 

So basically, when we call `.update()`, we do assign the new parameters to the config. *However*, we don't re-run all of the logic in the `__init__` method. This can lead to some potential issues. 

Example 1: parameter choices are not validated (*minor issue*)

```py
import gradio as gr

with gr.Blocks() as demo:
    r = gr.Audio(source="upload", interactive=True)
    b1 = gr.Button("microphone")
    b2 = gr.Button("abc")
    b1.click(lambda: gr.Audio(source="microphone"), None, r)
    b2.click(lambda: gr.Audio(source="abc"), None, r)
    
demo.launch()
```

Clicking on the first button will correctly change the source to microphone. Clicking on the second issue will collapse the audio component instead of raising an error that its an invalid source. 

Example 2: sometimes parameters need to be preprocessed before they are sent to the config (**bigger issue**)

```py
import gradio as gr

with gr.Blocks() as demo:
    r = gr.Radio([("abc", "a"), "d"])
    b1 = gr.Button("print")
    b2 = gr.Button("update")
    t = gr.Textbox()
    b1.click(lambda x:x, r, t)
    b2.click(lambda :gr.Radio([("efg", "e"), "h"]), None, r)
    
demo.launch()
```

Clicking on the "update" button, choosing the "h" option, and then clicking "print" will not print anything in the textbox. "h" should be printed, but its not. The reason is that `self.choices` is not the `choices` parameter. Its actually:

```py
        self.choices = (
            [c if isinstance(c, tuple) else (str(c), c) for c in choices]
            if choices
            else []
        )
```

but this never gets called in `.update()`

abidlabs:
I can think of two ideas:

1. Add a `post_init()` type method that does validation / modifies parameters as needed, and is called after both `init()` and `update()`

2. Move any logic that modifies the constructor parameters to the frontend (e.g. the choices code), but this may not always be possible 

I suggest we go with approach 1. We can formally do by this converting every component to a `dataclass` and adding a `__post_init__()` method where needed

aliabid94:
> So basically, when we call .update(), we do assign the new parameters to the config. However, we don't re-run all of the logic in the __init__ method. This can lead to some potential issues.

Okay fixed in this PR. Went with suggestion 2:

> Move any logic that modifies the constructor parameters to the frontend (e.g. the choices code), but this may not always be possible

So all constructor arguments to a Component must be set directly as the components property. Other than 'value' which gets postprocessed, these properties will get sent straight to the frontend in the same format.

The problem with:

> Add a post_init() type

The point of this PR was to remove the update method, which had all the same arguments as the constructor in the first place. If we introduce post_init, we bring back another method that has all the same arguments as init, add more complexity to a custom components, and also may create weird logic where a user might update some properties and not others, which may conflict if post_init has custom logic that involves multiple argument types.

abidlabs:
Ok that makes sense -- I actually think we should keep these changes and *still* convert all of the classes to dataclasses. The reason being that users who create custom components may not realize that they should not modify the constructor arguments before assigning them to the class attributes, whereas using dataclasses would enforce that behavior (and be much cleaner code). 

For the validation logic that is currently in `__init__()`, we can simply move that to a `__post_init__()` method. Not every class will need a `__post_init__()`, and this `__post_init__()` doesn't need to be called after `.update()` if you think it could lead to conflicts, but at least it will allow us to keep the validation logic that is currently in the constructor. WDYT?

js/checkboxgroup/interactive/InteractiveCheckboxgroup.svelte
```diff
@@ -11,7 +11,17 @@
 	export let visible = true;
 	export let value: (string | number)[] = [];
 	export let value_is_output = false;
-	export let choices: [string, number][];
+	export let choices: [string, string | number][] | string[] | null;
```
abidlabs, line 14: Can we move this logic to a `gradio/utils` instead of repeating multiple times?

aliabid94:
I changed my mind, forbidding users to have any custom logic on initialization will be limiting, as will only allowing dataclasses (which would block custom init methods). Here's what I'm going to do instead: 
- Allow any custom logic on init, as was always the case
- Automatically save all the arguments passed to a Component on construction
- on update, combine the new updated arguments and the saved previous arguments to create a new instance of that component, passing this combined set of arguments to the constructor

This allows us to do any custom logic, validation, etc. in init. Working on this now.

abidlabs:
Sounds good 👍 -- we should just make sure there are no side effects of creating a new component (e.g. rendering)

aliabid94:
@abidlabs ready for re-review

abidlabs:
Taking a look!

gradio/utils.py
```diff
@@ -646,11 +650,14 @@ def wrapper(*args, **kwargs):
...
 
-def get_function_with_locals(fn: Callable, blocks: Blocks, event_id: str | None):
+def get_function_with_locals(
+    fn: Callable, blocks: Blocks, event_id: str | None, in_event_listener: bool
+):
     def before_fn(blocks, event_id):
         from gradio.context import thread_data
 
         thread_data.blocks = blocks
+        thread_data.in_event_listener = in_event_listener
```
abidlabs, line 660: I'm trying to understand -- when is `thread_data.in_event_listener` set to `True`?

gradio/blocks.py
```diff
@@ -1784,6 +1857,7 @@ def launch(
             blocked_paths: List of complete filepaths or parent directories that gradio is not allowed to serve (i.e. users of your app are not allowed to access). Must be absolute paths. Warning: takes precedence over `allowed_paths` and all other directories exposed by Gradio by default.
             root_path: The root path (or "mount point") of the application, if it's not served from the root ("/") of the domain. Often used when the application is behind a reverse proxy that forwards requests to the application. For example, if the application is served at "https://example.com/myapp", the `root_path` should be set to "/myapp". Can be set by environment variable GRADIO_ROOT_PATH. Defaults to "".
             app_kwargs: Additional keyword arguments to pass to the underlying FastAPI app as a dictionary of parameter keys and argument values. For example, `{"docs_url": "/docs"}`
+            state_session_capacity: The maximum number of sessions that can be stored in the state. If the number of sessions exceeds this number, the oldest sessions will be removed. This prevents memory leaks when using stateful components or updates. Defaults to 10000.
```
abidlabs, line 1860:
```suggestion
            state_session_capacity: The maximum number of sessions whose information to store in memory. If the number of sessions exceeds this number, the oldest sessions will be removed. Reduce capacity to reduce memory usage when using gr.State or returning updated components from functions. Defaults to 10000.
```

abidlabs:
This needs to be updated: `/guides/03_building-with-blocks/01_blocks-and-event-listeners.md` to remove mention of `.update()`.

abidlabs:
Some of the components, such as the `Scatterplot`, `Barplot`, and `Lineplot` components, don't support kwargs in their init, which means they can't take the `render` kwarg. This breaks the `native_plots` demo as well as the `blocks_multiple_event_triggers` demo. We should add these arguments to the init. 

That's the only bug I could find! Otherwise lgtm. Very nice @aliabid94!

abidlabs (approved):

abidlabs:
I'm still seeing an error with `gr.LinePlot` -- you can see it erroring in `demo/blocks_multiple_event_triggers/run.py`.

Once the plot components are fixed and tested -- should be good to merge in!

abidlabs:
Still seeing one issue @aliabid94 -- will Slack you

PR #5248, opened 2023-08-16
Enable streaming audio in python client

freddyaboulton:
## Description

Update the FileSerializable so that it can handle streaming responses introduced in #5077

```python
import gradio_client as grc
client = grc.Client("<local-url-of-streaming-audio-demo>")
job = client.submit("/Users/freddy/Pictures/bark_demo.mp4", fn_index=1)

# The entire audio stream
job.result()

# Each individual chunk
job.outputs()
```

## 🎯 PRs Should Target Issues

## Tests


gradio/blocks.py
```diff
@@ -1362,10 +1362,10 @@ def handle_streaming_outputs(
                 if run not in self.pending_streams[session_hash]:
                     self.pending_streams[session_hash][run] = defaultdict(list)
                 self.pending_streams[session_hash][run][output_id].append(stream)
-                data[i] = {
-                    "name": f"{session_hash}/{run}/{output_id}",
-                    "is_stream": True,
-                }
+                if data[i]:
+                    data[i]["is_file"] = False
```
freddyaboulton, line 1366: Preserve the `orig_name` so that clients have an easier time downloading the byte stream

gradio/blocks.py
```diff
@@ -1362,10 +1362,10 @@ def handle_streaming_outputs(
...
+                    data[i]["name"] = f"{session_hash}/{run}/{output_id}"
```
freddyaboulton, line 1367: @aliabid94 Thoughts on including the `stream` endpoint name in the route so that it doesn't have to be hardcoded in the client?

client/python/gradio_client/serializing.py
```diff
@@ -291,6 +300,24 @@ def _deserialize_single(
...
+                    self.stream = self._setup_stream(
+                        root_url + "stream/" + x["name"], hf_token=hf_token
+                    )
+                try:
+                    chunk = next(self.stream)
+                except StopIteration:
+                    self.stream = self._setup_stream(
+                        root_url + "stream/" + x["name"], hf_token=hf_token
+                    )
+                    chunk = next(self.stream)
```
abidlabs, line 315: Sorry can you explain why these two lines are repeated from above?

gradio/blocks.py
abidlabs: But isn't ["name"] assigned to the same value in the next line?

client/python/gradio_client/utils.py
```diff
@@ -387,6 +387,16 @@ def encode_url_or_file_to_base64(path: str | Path):
...
 
 
+def download_byte_stream(url: str, hf_token=None):
+    arr = bytearray()
+    headers = {"Authorization": "Bearer " + hf_token} if hf_token else {}
+    with httpx.stream("GET", url, headers=headers) as r:
+        for data in r.iter_bytes():
+            arr += data
+            yield data
+    yield arr
```
abidlabs, line 397: Am I understanding correctly that what's happening is that we are iterating chunk by chunk and then finally yielding the full array so that `.predict()` returns the full data array? 2 questions:

(1) How come outputs() doesn't include this final full data array?
(2) Will `r.iter_bytes()` work if the streaming response returns binary data (as in #5238) instead of file data?

abidlabs:
Really nice @freddyaboulton! Just a few questions to understand the implementation -- will test more thoroughly tomorrow.

client/python/gradio_client/utils.py
freddyaboulton: (1) outputs only includes the values returned from the server during the `process_generating` event
(2) yea the `stream` enpoint only serves bytes so it should work when #5238 is merged

client/python/gradio_client/serializing.py
freddyaboulton: It was to determine when the stream changed but there's a better/clearer way. Will push up now!

gradio/blocks.py
freddyaboulton: Yea we're using "name" to mean the url from which to download the stream so I'm keeping that the same. Just preserving `orig_name` (if it's there) so that we can give a meaningful name when we go to download the byte stream to a file.

abidlabs:
I'll go ahead and merge this in so that we can consolidate things before the release

PR #5249, opened 2023-08-17
Fix for `gr.File()` "stuck" uploading

dawoodkhan82:
## Description
Checks for file blob, and updates download url for file.

Closes: #(issue)

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/js/file/shared/FilePreview.svelte
+++ b/js/file/shared/FilePreview.svelte
@@ -33,9 +33,9 @@
 					</td>
 
 					<td class="download">
-						{#if file.data}
+						{#if file.data || file.blob}
 							<a
-								href={file.data}
+								href={file.data ? file.data : URL.createObjectURL(file.blob)}
 								target="_blank"
 								download={window.__is_colab__
 									? null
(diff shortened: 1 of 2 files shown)


abidlabs:
@dawoodkhan82 very nice, this does fix the issue for me! However, I'm pretty sure that the data URLs have changed. When you used to click on them, you'd get a path to the temporary file on the server, now you get a more mysterious URL like: `blob:http://127.0.0.1:7864/6caa7c02-484d-4a0f-b2db-5fc6bd462df4`

abidlabs:
Also is a complete demo working for you? here's what I tried (nothing seems to happen):

```py
import gradio as gr

with gr.Blocks() as demo:
    f = gr.File()
    f2 = gr.File()
    f.upload(lambda x:x.name, f, f2)
    
demo.launch()
```

dawoodkhan82:
Oh I see, will take a look

js/file/shared/FilePreview.svelte
```diff
@@ -33,9 +33,9 @@
 					</td>
 
 					<td class="download">
-						{#if file.data}
+						{#if file.data || file.blob}
```
freddyaboulton, line 36: Why do we have blobs now? All files are uploaded/stored on the server. file.data should not be null 🤔

Issue #5162, opened 2023-08-10
Empty `gr.Dataframe` is not empty


hysts:
### Describe the bug

When passing an empty dataframe, say `pd.DataFrame([], columns=['a', 'b'])`, to `gr.Dataframe`, the value becomes non-empty dataframe `pd.DataFrame(['', ''], columns=['a', 'b'])`.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr
import pandas as pd

df_orig = pd.DataFrame([], columns=['a', 'b'])

with gr.Blocks() as demo:
    df = gr.Dataframe(value=df_orig, interactive=False)
    btn = gr.Button()
    text = gr.Textbox()
    btn.click(fn=lambda x: x.shape, inputs=df, outputs=text)
demo.queue().launch()
```
https://huggingface.co/spaces/hysts-debug/empty-dataframe

An empty row is shown in the table view, and it says the shape is `(1, 2)` when clicking the button, but the table shouldn't show the empty row and the shape should be `(0, 2)`.

### Screenshot

![](https://github.com/gradio-app/gradio/assets/25161192/386696b6-c0c0-4433-b893-d91408bf1b5a)

### System Info

```shell
gradio==3.39.0
```

### Severity

I can work around it

abidlabs:
When trying to reproduce this, I noticed that I can't run your example fully. Clicking on the button gives me:

> ValueError: 3 columns passed, passed data had 2 columns

Can you retry your example on the latest version of `gradio`?

hysts:
Thanks. Yeah, looks like the code raises different error when using `gradio==3.40.1`. https://huggingface.co/spaces/hysts-debug/empty-dataframe-gradio-3-40-1
(FYI, the reproduction with the old `gradio==3.39.0` is here: https://huggingface.co/spaces/hysts-debug/empty-dataframe)

The error I get with `gradio==3.40.1` is
```
Running on local URL:  http://0.0.0.0:7860

To create a public link, set `share=True` in `launch()`.
Traceback (most recent call last):
  File "/home/user/.local/lib/python3.10/site-packages/pandas/core/internals/construction.py", line 934, in _finalize_columns_and_data
    columns = _validate_or_indexify_columns(contents, columns)
  File "/home/user/.local/lib/python3.10/site-packages/pandas/core/internals/construction.py", line 981, in _validate_or_indexify_columns
    raise AssertionError(
[...]

Issue #5236, opened 2023-08-15
new DataFrame are interactive even if shouldn't be


pseudotensor:
### Describe the bug

Ever since upgrading to gradio==3.40.1 from gradio==3.39.0, dataframes shows add row/column and one can change the dataframe, even though interactive=False

Related issues:
https://github.com/gradio-app/gradio/issues/5182
https://github.com/gradio-app/gradio/pull/5183

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```
import gradio as gr


def filter_records(records, gender):
    return records[records["gender"] == gender]


demo = gr.Interface(
    filter_records,
    [
        gr.Dataframe(
            headers=["name", "age", "gender"],
            datatype=["str", "number", "str"],
            row_count=5,
            col_count=3,
            interactive=False,
        ),
        gr.Dropdown(["M", "F", "O"]),
    ],
    "dataframe",
    description="Enter gender as 'M', 'F', or 'O' for other.",
)

if __name__ == "__main__":
    demo.launch()

```

### Screenshot

![image](https://github.com/gradio-app/gradio/assets/2249614/c2ba8a56-3d6c-4545-b4c8-426a52a5fa6d)

In https://github.com/gradio-app/gradio/issues/5237 you can see how in 3.39.0 the add row-column thing is not present, there is no way to change the frame in the UI as is correct when interactive=False

### System Info

```shell
gradio 3.40.1

I work around by downgrading gradio back to 3.39.0
```

### Severity

I can work around it

pseudotensor:
Also, it used to just listen to the column names, not have to pass separate "headers" which is hassle.

abidlabs:
Interesting, will take a look.

abidlabs:
Unfortunate regression, but can confirm that this is now fixed on `main`: 

<img width="763" alt="image" src="https://github.com/gradio-app/gradio/assets/1778297/836150dc-b14b-4ff8-9d53-c8ebd8a33602">

I'll add some visual tests in a separate PR

PR #5256, opened 2023-08-17
Better handling of empty dataframe in `gr.DataFrame`

abidlabs:
We were handling empty dataframes incorrectly in the `gr.Dataframe` --> we were turning them into single-row arrays with each cell being "".

This fixes that -- would appreciate if folks could test this out to make sure I didn't miss anything as the dataframe frontend code is quite complex.

Fixes: #5162

Test code:

```py
import gradio as gr
import pandas as pd

df_orig = pd.DataFrame([], columns=['a', 'b', 'c'])

df_fill = pd.DataFrame([[1, 2, 3]], columns=['a', 'b', 'c'])


with gr.Blocks() as demo:
    df = gr.Dataframe(value=df_orig, interactive=True, row_count=(0, "dynamic"))
    df2 = gr.Dataframe(value=df_fill, interactive=True, row_count=(0, "dynamic"))
    btn = gr.Button()
    text = gr.Textbox()
    btn.click(fn=lambda x: x.shape, inputs=df, outputs=text)
demo.queue().launch()
```

Also adds some stories for `gr.DataFrame` to prevent regressions like #5236 and some stories (unrelatedly) for `gr.Dropdown()`

### DIFF
--- a/.changeset/khaki-otters-wear.md
+++ b/.changeset/khaki-otters-wear.md
@@ -1,0 +1,6 @@
+---
+"@gradio/dataframe": patch
+"gradio": patch
+---
+
+fix:Better handling of empty dataframe in `gr.DataFrame`
--- a/js/dataframe/Dataframe.stories.svelte
+++ b/js/dataframe/Dataframe.stories.svelte
@@ -1,0 +1,63 @@
+<script>
+	import { Meta, Template, Story } from "@storybook/addon-svelte-csf";
+	import Table from "./shared/Table.svelte";
+</script>
+
+<Meta
+	title="Components/DataFrame"
+	component={Table}
+	argTypes={{
+		editable: {
+			control: [true, false],
+			description: "Whether the DataFrame is editable",
+			name: "interactive",
+			value: true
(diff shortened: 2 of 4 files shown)


pngwn:
This is interesting. So I'm not convinced this issue is actually a bug, it was by design. The reason some empty rows are rendered is so a user can click into the cells and start typing, tabbing, etc. With no rows present this isn't possible.

abidlabs:
> The reason some empty rows are rendered is so a user can click into the cells and start typing, tabbing, etc. With no rows present this isn't possible.

We shouldn't be mutating the data to achieve this. As it stands, if you pass in a dataframe with 0 rows into a `gr.DataFrame` and try to get it back, it will have a single row, which can cause problems for downstream processing. 

I don't think the UX issue is really there either, since we include a prominent "Add Row" button underneath the dataframe (as long as `interactive=True`).

pngwn:
Yeah, that makes sense.

abidlabs:
Thanks @pngwn!

Issue #3769, opened 2023-04-05
Assign names and avatars in `gr.Chatbot`


markwsac:
It would be great if we can assign names to characters in Chatbot. For e.g. "Human" before each prompt. "AI" before each response. I have read documentation and couldn't find any way to do this.

abidlabs:
+1, and perhaps small avatars as well?

markwsac:
that would be great

90bird90:
How to adjust the initial width of the textbox of the chatbot Although it is of course self-adaptive, the initial size is really large

CynthiaChuang:
It seems to only support avatar display without the ability to specify a name？

efaraneda:
@markwsac @abidlabs Not the ideal solution, but works for me! You can try playing around adding some HTML right before every answer like this:

<img width="494" alt="image" src="https://github.com/gradio-app/gradio/assets/71103961/ec287ad2-6e67-4fe5-9617-7140bb46c41e">

And then, you can set a dynamic variable to capture the user's name as an input!

chigkim:
Hi, Is there any progress on assigning custom names for user and bot?

abidlabs:
Not yet, we are working on other things at the moment, but are open to reviewing PRs around this

PR #5258, opened 2023-08-18
Chatbot Avatar Images

dawoodkhan82:
## Description
Ability to add avatar images for the "user" and "bot" in `gr.Chatbot()`. Simply provide image(s) `gr.Chatbot(avatar_images=["user_image", "bot_image"])`.

Screenshot 2023-08-18 at 12 03 33 PM

Closes: #3769

## 🎯 PRs Should Target Issues

## Tests


abidlabs:
Sorry @dawoodkhan82 pushed to this branch by accident -- will revert

abidlabs:
Tested, changes look good! My only suggestion would be to use `self.make_temp_copy_if_needed()` and pass the file path to the frontend instead of passing in the base64. For large files, serialization can be slow, and we don't take advantage of the browser caching the file. 

Ref: https://stackoverflow.com/questions/1574961/how-much-faster-is-it-to-use-inline-base64-images-for-a-web-site-than-just-linki

yvrjsharma:
Thanks, @dawoodkhan82 results look amazing. I passed none for user_image and idefics logo for the bot in this attached example. 
image

I was wondering if you plan to add avatar images to the _shared discussions_ as well. I can see that right now, it is not supported.
image

Sample Demo here - https://huggingface.co/spaces/ysharma/dummy_avatar

dawoodkhan82:
@abidlabs Do you know how the shared discussions are generated?

dawoodkhan82:
@abidlabs ready to merge after these tests pass.

dawoodkhan82:
Want to confirm that `avatar_images` being an array and having an option to pass `None` is the best implementation right?

abidlabs:
I would have probably picked tuple of 2 strings (tuples since its a fixed length). Allowing `None` is great!

yvrjsharma:
Allowing None was genius tbh!

js/chatbot/static/ChatBot.svelte
```diff
@@ -42,6 +44,9 @@
 
 	let div: HTMLDivElement;
 	let autoscroll: boolean;
+	let images_dir: string = root_url
```
dawoodkhan82, line 47: @abidlabs Isn't this what `normalise_files` is meant to do?

abidlabs:
LGTM @dawoodkhan82! Among other things, I confirmed:

(1) That this works with `.update()`

```py
import gradio as gr

with gr.Blocks() as demo:
    c = gr.Chatbot([("abc", "def")], avatar_images=["cat.jpg", None])
    btn = gr.Button()
    btn.click(lambda : gr.update(avatar_images=(None, "cheetah.jpg")), None, c)
    
demo.launch()
```

2. as well as with gr.ChatInterface:

```py
import gradio as gr

ch = gr.Chatbot(avatar_images=["cat.jpg", None])

gr.ChatInterface(lambda x,y:x, chatbot=ch).launch()
```

abidlabs (approved):

js/chatbot/static/ChatBot.svelte
abidlabs: Good point. I just took a look though and it looks like `normalise_files` doesn't handle the string case fully. In particular, it doesn't attach the `/file=` or `/proxy` prefixes if the input is a string.

I'm not sure about the full implications of patching `normalise_files` so I just added the logic there myself.

PR #5260, opened 2023-08-18
Stop passing inputs and preprocessing on iterators

aliabid94:
There's no need to be passing inputs and preprocessing them on every generator call - this is extremely inefficient and causes slowdowns if the preprocess step is expensive. Was causing significant issues raised internally here: https://huggingface.slack.com/archives/C02QZLG8GMN/p1692350120362899?thread_ts=1692278870.837829&cid=C02QZLG8GMN

Fixed now, preprocessing only happens on the first step.

### DIFF
--- a/.changeset/hip-queens-grin.md
+++ b/.changeset/hip-queens-grin.md
@@ -2,4 +2,4 @@
 "gradio": minor
 ---
 
-feat:ignore this, testing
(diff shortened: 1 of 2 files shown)


freddyaboulton (approved):
Awesome fix thanks @aliabid94 !

PR #5265, opened 2023-08-18
Removes scrollbar from File preview when not needed

abidlabs:
The scrollbar would always show in the `gr.File` preview. This quick fix removes it when not needed.

Test code:

```py
import gradio as gr

with gr.Blocks() as demo:
    f = gr.File(["cheetah.jpg"]*10)
    
demo.launch()
```


pngwn (approved):
Looks good! Thanks @abidlabs!

abidlabs:
Thanks for the guidance @pngwn!

abidlabs:
Thanks guys! Merging

Issue #5167, opened 2023-08-10
Model3D: allow setting initial camera position


mbahri:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
Some 3D assets are not oriented in a convenient way for the default camera settings in Babylon.js, resulting in the need to rotate them manually for visualization.

**Describe the solution you'd like**  
it would be convenient to be able to set the initial camera rotation (or full position) when the viewer loads.
```
# Example API
gr.Model3D(alpha=<longitudinal rotation>, beta=< latitudinal rotation>, radius=None)
```

**Additional context**  
https://doc.babylonjs.com/features/featuresDeepDive/scene/fastBuildWorld#create-default-camera-or-light
https://doc.babylonjs.com/features/featuresDeepDive/cameras/camera_introduction#arc-rotate-camera

mbahri:
I can try but I'll need approval from my employer before contributing the change here.

A quick attempt at hardcoding the angles in Svelte worked for setting the viewing angle for Model3D components used as outputs, but not for the preview of assets uploaded as inputs.

I also need to fix the passing of parameters from the Python component to the actual JS component. I'll have another look tomorrow but it wasn't immediately clear to me how the wrappers mentioned in https://www.gradio.app/guides/creating-a-new-component (`js/app/src/components/Model3D/index.ts`) are forwarding arguments from the Python constructor.

I tried adding an `alpha` param. in Python and adding `export let alpha: number = 180;` to `js/model3D/static/Model3D.svelte` but that didn't work.

mbahri:
PR submitted: https://github.com/gradio-app/gradio/pull/5266

A-pril:
Hello~I wanna know if  I can get the real-time camera pose since I use it as a input to my model. Thanks a lot!

fabienbaradel:
@marcnotenboom did you find a way to solve this ? I am facing the same issue

marcnotenboom:
> @marcnotenboom did you find a way to solve this ? I am facing the same issue

I moved away from it and am using Three.js now

PR #5266, opened 2023-08-18
Makes it possible to set the initial camera position for the `Model3D` component as a tuple of (alpha, beta, radius)

mbahri:
## Description

This change addresses improvement request https://github.com/gradio-app/gradio/issues/5167 and makes it possible to set the initial ArcCamera position for Model3D (input and output).

Closes: #5167


abidlabs:
Thanks @mbahri for the contribution! Instead of adding 3 separate parameters, could we add a single parameter `initial_position` which takes in a `tuple` of `(alpha, beta, radius)`?

mbahri:
> Thanks @mbahri for the contribution! Instead of adding 3 separate parameters, could we add a single parameter `initial_position` which takes in a `tuple` of `(alpha, beta, radius)`?

I'll have a look :)

abidlabs:
Did you get a chance to look into this @mbahri?

mbahri:
Hi @abidlabs, yes I've made the change in the code, just haven't had time to test it yet. I'll update the PR today

abidlabs:
Sounds good thanks @mbahri

abidlabs:
Thanks @mbahri for pushing the changes. Some tests may need to be fixed, I can take care of that.

abidlabs:
whoops I think we might have conflicted right there

abidlabs:
Are you still making changes to the PR or should I take over?

mbahri:
Hey @abidlabs, sorry, I didn't expect you to be working on this now. I've made some minor formatting changes and added a docstring for the new parameter.

I'm not too familiar with your coding style so please don't hesitate to rename variables or the type alias `NumberOrNone = int | float | None` I added for brevity to the model3d.py file (L.22).

abidlabs:
Made a few tweaks -- for example the conversion factor from radians to degrees should be pi/180, not pi/360, correct?

mbahri:
Correct, I mixed it up because I needed to set the angle to pi/2 to have my test models visually do a "180" on screen

I'd like to change the docstring for formatting and specifying which unit the angles are in but don't want to interfere now:
```
initial_camera_position: Tuple `(alpha, beta, radius)`. Sets the starting longitudinal rotation (`alpha`), latitudinal rotation (`beta`) and distance to the 3D model being displayed (`radius`) for the camera. Set any value to None to use the default value. `alpha` and `beta` should be provided in degrees.
```

abidlabs:
Does this work? 

>  camera_position: initial camera position of scene, provided as a tuple of (alpha, beta, radius). Each value is optional. If provided, alpha and beta should be in degrees from -180 to 180 reflecting the angular position along the longitudinal and latitudinal axes, respectively. Radius corresponds to the distance from the center of the object to the camera.

mbahri:
Sure, I just changed my commit to add backticks for the doc

js/model3D/interactive/Model3DUpload.svelte
```diff
@@ -101,6 +108,19 @@
 			scene,
 			() => {
 				scene.createDefaultCamera(true, true, true);
+				var helperCamera = scene.activeCamera;
```
dawoodkhan82, line 111: Seems like we need a null check on `helperCamera` for the typescript errors.

mbahri:
> from -180 to 180

Do we need to specify this range? I'm checking a couple things rn

js/model3D/interactive/Model3DUpload.svelte
mbahri: The current code assumes `scene.createDefaultCamera` succeeds, do you want to add error handling (and in that case, what needs to happen?) or just assume there's an active camera?

I'm pushing a change that keeps assuming the call worked - as a starting point

abidlabs:
> Do we need to specify this range? I'm checking a couple things rn

ah maybe not, I thought the range in babylon was -pi to pi radians, but I haven't tested to see if the values wrap around.

abidlabs:
Used this opportunity to do some further refactoring & reduce the duplicated code in the component (not strictly related to your changes @mbahri)

js/model3D/interactive/Model3DUpload.svelte
dawoodkhan82: Actually looks like the type error is gone, so should be fine as is.

dawoodkhan82:
Code looks good to me!

dawoodkhan82 (approved):

abidlabs:
Cool thanks! And thanks @mbahri for contributing this feature & bearing with us as we reviewed it!

PR #5267, opened 2023-08-18
Faster reload mode

freddyaboulton:
## Description

Skip restarting the server when in reload mode. This is faster in my local experience.

This works by reloading the module to get the latest instance of the demo. This comes with some caveats.

- ~If you don't have `'if __name__ == "__main__"` you're going to block the thread when we reimport your module. So we'll never be able to refresh the page. I suspect this is part of the reason people have a bad experience with reload mode today. We know raise an error before launching the demo in reload mode in this case.~ If you don't have `if__name__ ...`, we will allow reload mode. 
- If you have really expensive code in your module, e.g. loading a big ML model, we will have to rerun that every time. That's how module reloading works in python. You can see this for yourself by adding an arbitrary time.sleep(10) to the top of any demo file and running reload mode. This is a limitation of the current uvicorn reloader as well since they also reload the entire module.
- I got it to work in jupyter notebook too! Should work in colab as well 😄 In jupyter notebooks, we only pay attention to changes in the cell that's called with `%%blocks`. You'll notice that in the jupyter demo, the UI gets a page refresh as opposed to being modified in place like when you edit from source files. That's because of how jupyter cells work. We _have_ to return a view of the UI from the cell magic otherwise the view will go away entirely. This looks like a page refresh.

new_reload_mode

add_an_event

jupyter_reload

## 🎯 PRs Should Target Issues

## Tests


abidlabs:
So much faster, the video didn't do it justice 🔥🔥

pngwn:
@freddyaboulton @abidlabs I've pushed some changes so the frontend doesn't need to be reloaded, and we don't even need to rerender all blocks if nothing has changed, so it should be pretty snappy if you run `gradio run.py` and `pnpm dev` together.

I agree with @abidlabs regarding the errors, it is pretty annoying to restart with dev mode because you also need to reload the frontend (as the websocket connections closes). Is there a way catch any exceptions before they get to the 'reload' mechanism (so it doesn't error everything out) and continue to reload on subsequent changes.

I guess even if we do this the WS might still get blown away. It might be worth standing up separate server to handle the dev mode WS, so it doesn't break if the gradio app breaks. 

Regarding watch files, can't we use something like [pydeps](https://github.com/thebjorn/pydeps) to get a dependency graph and only watch files included in the graph (filtering out anything non-local, and limiting ourselves to a certain directory for security reasons)? I would limit things to cwd by default but allow users to pass in a directory to overwrite this, would be helpful for the frontend too eventually.

pngwn:
I also resolved the merge conflicts.

pngwn:
@freddyaboulton I think this should be working now. 

I've fixed a bug where components were always ending up as static when reloading if `interactive` wasn't explicitly set to true. And adding or changing events should work as expected too.

I made some changes to the dev websocket ping event as well, so that it wasn't spamming constantly, it fires every 1 second now. I needed to refactor this a little to make sure it didn't block.

freddyaboulton:
This is finally ready for review! I got it to work for jupyter notebooks too! I updated the PR description to reflect this. Thanks @pngwn for the front-end help.

To answer your questions @abidlabs and @pngwn:

> (1) it seems that the reload mode is triggered if certain gradio files (such as the component/*.py files) are edited, but not others (e.g. routes.py) are. Where is this defined and what's the logic for this? (Note that the console does print "Changes detected in: /Users/abidlabs/dev/gradio-repos/gradio/gradio/routes.py" but the page does not refresh)

Since the server is not being restarted, code chages to the server logic won't be reflected. Basically only things from `gr.Blocks` and below will be reflected.

> (2) Is environment variables the best way to pass the watch directories? I think it might be an issue if people are developing multiple gradio apps at the same time (though not very likely)

I think we're actually ok because the subprocess created by `subprocess.Popen` will get it's own env.

> (3) In the case of an exception, the reload mode stops working. That's fine and the behavior of some other hot reload modes I've used, but I wonder (since we're rolling our own) whether we can have it reload once the exception has been fixed? Sometimes you can have an exception if you save your file in the middle of a change.

I changed it so that if there's an exception reloading the module, the stacktrace shows up with a message saying that reloading failed but the demo keeps running.

> I would limit things to cwd by default but allow users to pass in a directory to overwrite this, would be helpful for the frontend too eventually.

Yea I can add an additional watch_dirs argument to the command

pngwn:
I think we also added it so that we could generate the configs by importing those demo modules, without starting the app. We don't do that anymore, so that part doesn't matter

gradio/reload.py
```diff
@@ -3,34 +3,59 @@
...
-        demo_name = "demo.app"
+        demo_name = "demo"
     else:
         demo_name = args[1]
-        if "." not in demo_name:
+        if "." in demo_name:
+            demo_name = demo_name.split(".")[0]
             print(
-                "\nWARNING: As of Gradio 3.31, the parameter after the file path must be the name of the FastAPI app, not the Gradio demo. In most cases, this just means you should add '.app' after the name of your demo, e.g. 'demo' -> 'demo.app'."
+                "\nWARNING: As of Gradio 3.41.0, the parameter after the file path must be the name of the Gradio demo, not the FastAPI app. In most cases, this just means you should remove '.app' after the name of your demo, e.g. 'demo.app' -> 'demo'."
```
abidlabs, line 30:
```suggestion
                "\nWARNING: As of Gradio 3.42.0, the parameter after the file path must be the name of the Gradio demo, not the FastAPI app. In most cases, this just means you should remove '.app' after the name of your demo, e.g. 'demo.app' -> 'demo'."
```

abidlabs: We should also update this guide here: https://www.gradio.app/guides/developing-faster-with-reload-mode

abidlabs:
> I changed it so that if there's an exception reloading the module, the stacktrace shows up with a message saying that reloading failed but the demo keeps running.

If there's e.g. a SyntaxError, I do see the error in the stacktrace, however reload mode doesn't resume for me. I can share a repro if needed, but it seems to happen for any type of exception. It may be related to the comment I left above.

abidlabs:
Looks great @freddyaboulton! I left a couple of small notes but working blazing fast. I agree with @aliabid94 and @pngwn that if its possible to remove the limitation of launching your gradio demo from within the `if __name__ == "__main__"`, that would make it even better. 

cc @julien-c who asked for a better reload mode waay back when Gradio was acquired!

gradio/utils.py
```diff
@@ -57,6 +63,142 @@
...
+                if sourcefile and is_in_or_equal(sourcefile, dir_):
+                    del sys.modules[k]
+            try:
+                module = importlib.import_module(reloader.watch_file)
+                module = importlib.reload(module)
+            except Exception as e:
+                print(
+                    f"Reloading {reloader.watch_file} failed with the following exception: "
+                )
+                traceback.print_exception(etype=None, value=e, tb=None)
```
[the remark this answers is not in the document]
freddyaboulton: Seems like the api changed between 3.8 and 3.11 😱 Will look into it

abidlabs:
> Is this on 3.11?

Yup on 3.11

freddyaboulton:
@abidlabs I addressed all of your comments, updated the guide, and made it so that `if __name__` is not required anymore! Should be good for another look.

abidlabs:
@freddyaboulton when I test this without the `if __name__ == "__main__"`, the program terminates prematurely and the Gradio app is never launched. Here's my file:

```py
import gradio as gr

with gr.Blocks() as demo:
    gr.Textbox()
```

When I run `gradio test3.py`, I get this:

```
Watching: '/Users/abidlabs/dev/gradio-repos/gradio/gradio', '/Users/abidlabs/dev/gradio-repos/gradio'
```

and the program then terminates without the server launching on any port. Whereas if I do include the clause, I get this:

```
Watching: '/Users/abidlabs/dev/gradio-repos/gradio/gradio', '/Users/abidlabs/dev/gradio-repos/gradio'

Running on local URL:  http://127.0.0.1:7863
```

Let me know if you are able to reproduce this behavior!

abidlabs:
Synced with @freddyaboulton to resolve the issue. It seems like when I manually restarted the script, Chrome was being finicky and required a manual reload of the page. 

Everything else looks great -- sweet PR @freddyaboulton!

abidlabs (approved):

freddyaboulton:
Thanks everyone for the reviews!!

PR #5268, opened 2023-08-18
Move markdown & latex processing to the frontend for the gr.Markdown and gr.DataFrame components

abidlabs:
Moves markdown and latex processing to the frontend for the `gr.Markdown` and `gr.DataFrame` components. This has several advantages:

1. Removes 2 dependencies from the `gradio` python package 🥳 
2. Katex.js renders LaTeX much more nicely and without the alignment issue that we were seeing with `markdown_it` (see screenshots below)
3. Moves some of the work client side making gradio apps faster (we've seen this before with `gr.Chatbot`)

It should be noted that this increases the javascript bundle size for these two components. 

Screenshots (before):

image

image

Screenshots (after):

image

image

Closes: #4523 

Also added:
* the `latex_delimiters` parameter to `gr.Markdown()` and `gr.DataFrame` for control over whether to parse LaTeX
* stories for these two components as a visual test.

Thanks @dawoodkhan82 for the pointers!


abidlabs:
Fighting an issue with the frontend build step, which was working until just recently

dawoodkhan82:
We use `$$` as the default `latex_delimiters` for chatbot. We should be consistent across components.

dawoodkhan82:
Tested both `markdown` and `dataframe`. Both look good to me!

abidlabs:
> We use $$ as the default latex_delimiters for chatbot. We should be consistent across components.

That is true -- however, the previous behavior of `gr.Markdown` and `gr.Dataframe` was to render LaTeX enclosed in `$` so I kept it that way for backwards compatibility. Maybe we make the change in 4.0?

abidlabs:
Fixed up the backend tests -- also I've recently had to make a lot of changes manually to the XRAY_CONFIG e.g. every time I add a new parameter to a component. I wrote a script `scripts/overwrite_xray_config.py` that overwrites the test XRAY_CONFIG. Of course, it should be used with caution since we should only override the test config if we're sure that there aren't any accidental changes. But it can help avoid tedious manual work particularly for the component cleanup. cc @dawoodkhan82 @hannahblair for visibility

abidlabs:
Thanks for the review @dawoodkhan82! Merging

Issue #5243, opened 2023-08-16
Allow dynamic update of `gr.Button` icon value


hannahblair:
- [x] I have searched to see if a similar issue already exists.

Update gr.Button to allow the dynamic update of the icon

PR #5285, opened 2023-08-21
Tweaks to `icon` parameter in `gr.Button()`

abidlabs:
When I was working with @dawoodkhan82 on adding avatar images, we realized that we should slightly change the way that we handle icons in `gr.Button()` as well. I've implemented that logic for `gr.Button()` here. 

The main advantage of this approach is that we no longer do any preprocessing in the backend of the `icon` parameter in the `__init__` of `gr.Button()`. This is important because @aliabid94 is working on unifying the `__init__` and `update()` of each component so we should make both as simple as possible. As a bonus, this fixes: #5243

Made some tweaks to `gr.Chatbot` and `gr.Dataset` as well.

cc @hannahblair @dawoodkhan82 

Feel free to wreck my frontend code.

Test code:

```py
import gradio as gr

path = "cheetah.jpg"
url = "https://fastly.picsum.photos/id/322/200/300.jpg?hmac=q6h4jr1n6SrXrRCeqCblcexGQCfYmSXhr8Oo5EGoHIU"

with gr.Blocks() as demo:
    b1 = gr.Button(icon=path)
    b2 = gr.Button(icon=url)
    
    b2.click(lambda :gr.Button.update(icon=url, value="hello"), None, b1)
    b1.click(lambda :gr.update(icon=path, value="hi"), None, b2)    
    
demo.launch()
```


js/upload/src/utils.ts
```diff
@@ -57,6 +57,25 @@ export function normalise_file(
...
+function is_url(str: string): boolean {
+	try {
+	  const url = new URL(str);
+	  return url.protocol === "http:" || url.protocol === "https:";
+	} catch {
+	  return false;
+	}
+  }
+  
+export function get_fetchable_url_or_file(path: string | null, root: string, root_url: string | null): string {
```
freddyaboulton, line 69: We should call this in `normalise_file` so that both are in synch if we ever need to make changes. Otherwise looks good to me! Agree we should be setting the url in the front end.

abidlabs:
Thanks @freddyaboulton @hannahblair I've addressed your comments. Let me know if you have other feedback!

freddyaboulton (approved):
Thanks for the refactor and fix @abidlabs !

PR #5286, opened 2023-08-22
Removes unnecessary assignment in `interface.py`

JalinWang:
## Description

Remove a useless assignment

`self.fn` is already used in this function (L680)

image


abidlabs:
Thanks @JalinWang -- I'll go ahead and merge this in as the relevant tests are passing.

Issue #4520, opened 2023-06-15
devcontainer postCreateCommand fails


charlesfrye:
### Describe the bug

Thanks for offering a devcontainer! Made it much easier to do my first PR.

But the postCreateCommand failed for me, which made my onboarding rougher than it needed to be.

The troublesome code is [here](https://github.com/gradio-app/gradio/blob/d65512cb3a91e2c5b6abf70b52f56a51195c47e7/.devcontainer/devcontainer.json#L34).

One issue was that the second step in the command is just `scripts/install_test_requirements.sh` instead of `bash scripts/install_test_requirements.sh` -- unlike the other two steps, which are prefixed with `bash`. So you get a perms error.

Additionally, the smallest Codespace on GitHub fails with an OOM error when running the final step. I believe the workaround is to set some configuration in GitHub Codespaces to make sure you developers always get the right instance type, but I'm not sure.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

Create a new Codespace and wait for the postCreateCommand to run. It errors out with a permissions failure when installing test requirements. Run `bash scripts/install_test_requirements.sh` to resolve.

Then try running `bash scripts/build_frontend.sh` and observe an OOM failure if you are in the smallest Codespace instance (2-core, 4GB RAM). Specifically, its a system exit code 143, which a cursory Google search suggests is called by containers when they hit the memory limits imposed by the host OS.

### Logs

```shell
Sorry, forgot to screenshot or copy. Should be easy to repro with Codespaces though!
```

### System Info

```shell
devcontainer on GitHub CodeSpaces
```

### Severity

serious, but I can work around it

freddyaboulton:
Thanks for reporting @charlesfrye ! Will take a look soon

PR #5289, opened 2023-08-22
Fix devcontainer postCreateCommand

Faiga91:
## Description

- Update the `postCreateCommand` in the `devcontainer` , to use `chmod` commands to grant execute permissions to each script before running them.

- Increase the memory limit by adding `NODE_OPTIONS` environment variable in the `postCreateCommand`.

- Sat minimum requirements for the Codespace machines.

🎯 Closes: #4520


abidlabs:
Thanks so much @Faiga91 for the PR. 

#4520 also mentions another issue:

> Additionally, the smallest Codespace on GitHub fails with an OOM error when running the final step. I believe the workaround is to set some configuration in GitHub Codespaces to make sure you developers always get the right instance type, but I'm not sure.

Would you be able add this fix as well to this PR?

Faiga91:
@abidlabs I increased the memory limit in the `postCreateCommand`, and sat a [minimum specifications requirements for Codespace machine](https://docs.github.com/en/codespaces/setting-up-your-project-for-codespaces/configuring-dev-containers/setting-a-minimum-specification-for-codespace-machines). This solves the OOM problem at my end.

abidlabs:
Thank you so much @Faiga91 lgtm!

abidlabs (approved):

PR #5295, opened 2023-08-22
Allow caching examples with streamed output

aliabid94:
This PR allows setting cache_examples=True when a function generates streamed content. It works by capturing the yielded content, and for any StreamingOutput components, concatening the output binary content and saving it as a file.

In the process of doing this, I realized that when loading from cache, we use the serialize method, which unnecessarily converts the content into base64 in this case. I added an `allow_links` argument which will let the cached examples return links, making them faster. (This was also done because the the audio playback was giving me issues when trying to set the src of the audio element to a base64 without specifying the mimetype).


gradio/helpers.py
```diff
@@ -337,6 +342,38 @@ async def get_final_item(*args):
...
+                            dir=tempdir, delete=False
+                        )
+                        with open(temp_file.name, "wb") as f:
+                            f.write(binary_data)
+
+                        output[output_index] = {
+                            "name": temp_file.name,
+                            "is_file": True,
+                            "data": None,
+                        }
```
abidlabs, line 375: Can pull this out as a separate method and unit test it?

abidlabs:
> In the process of doing this, I realized that when loading from cache, we use the serialize method, which unnecessarily converts the content into base64 in this case. I added an allow_links argument which will let the cached examples return links, making them faster. (This was also done because the the audio playback was giving me issues when trying to set the src of the audio element to a base64 without specifying the mimetype).

Nice catch @aliabid94. I looked into this and I'm pretty sure that instead of adding another parameter to all of the serialization methods, we can actually make this the *default behavior* of `FileSerializable.deserialize()`. The reason is that all of the `FileSerializable` components (video, audio, file, uploadbutton, model3d) support data being in the format of files instead of base64. So if the data is already in the form of a file, we should just return that filepath instead of returning a base64 string. This is not only cleaner but also lets us benefit from this when using the client to call any FileSerializable space (no conversion to base64 is necessary).

abidlabs:
Generally looks good @aliabid94, made some suggestions above!

aliabid94:
> lgtm @aliabid94, if we could add a unit test for the chunk concatenation, that'd be great.

Done

Issue #5274, opened 2023-08-21
Languages method of gr.Code broken on docs


aliabd:
There's a languages method on the code component but it's broken on the docs. 

![Screen Shot 2023-08-20 at 6 26 00 PM](https://github.com/gradio-app/gradio/assets/9021060/99d9909c-698f-4de6-9e1d-c156eb410fae)

Issue #5208, opened 2023-08-14
Reduce space by initialization argument table of event listeners


aliabd:
All the event listeners take the same parameters so it doesn't make sense to keep listing the (super long) initialization table in the docs. For example, look at https://www.gradio.app/docs/audio which has 10 event listeners, the page is unnecessarily long because each method is repeating the same info. Should be reworked so that the table shows up once and only the name, signature and descriptions are listed for each listener.

PR #5298, opened 2023-08-23
Create event listener table for components on docs

aliabd:
All the methods for components are event listeners, and all the event listeners take in the same arguments. so there is needless repetitive content on the docs. This PR reworks things so that the arguments table for the event listeners shows up once and a new table shows supported listeners and their descriptions. 

Closes: #5208 

Also removes `gr.Code.languages` from documentation (already documented in the langauge param) 

Closes: #5274 

Also fixes a broken image path in a guide.


freddyaboulton:
I love this! Two nits:

* `gr.Code.languages` is showing up as an event listener for code
image

* Maybe call it `Event Arguments` as opposed to `Arguments` in case someone is skimming the page? 





Also, when I click around the components the page eventually gets stuck. Happens pretty consistently

website_code_component

aliabd:
@freddyaboulton 

- Languages was removed from docs in this pr, but will show up if you're not on main. can you use the toggle to switch from 3.40.0 to main?
- Good idea, switched to 'Event Arguments'
- Yeah i'm aware of this issue and working on it in a separate PR. Though fyi the error only appears when clicking around really fast, so I don't think a lot of users should be seeing it. still though, will fix soon.

freddyaboulton (approved):
Looks good to me! Thanks @aliabd ! Totally missed the new version toggle. Super cool

js/_website/src/lib/components/EventListeners.svelte
```diff
@@ -0,0 +1,96 @@
+<script lang="ts">
+	export let fns: any[];
+</script>
+
+
+<div id="event-listeners-description">
+    <h4 class="mt-8 text-xl text-orange-500 font-light group">
+        Description
+    </h4>
+    <p class="mb-2 text-lg text-gray-600">Event listeners are functions that are called when an event is triggered. Every event listener takes the same arguments, shown in the arguments table below.</p>
```
abidlabs, line 10:
```suggestion
    <p class="mb-2 text-lg text-gray-600">Event listeners allow you to capture and respond to user interactions with the UI components you've defined in a Gradio Blocks app. When a user interacts with an element, such as changing a slider value or uploading an image, a function is called.</p>
```

js/_website/src/lib/components/EventListeners.svelte
```diff
@@ -0,0 +1,96 @@
...
+    <p class="mb-2 text-lg text-gray-600"> The table below shows the event listeners supported by {fns[0].parent.replace("gradio.","")}.</p>
```
abidlabs, line 11: Nit: could we make the component name monospace. Something like this might work:
```suggestion
    <p class="mb-2 text-lg text-gray-600"> The table below shows the event listeners supported by the `{fns[0].parent.replace("gradio.","")}` component.</p>
```

abidlabs:
Also a couple of notes regarding the event listener docstrings themselves (not affected by your PR but would be great if we could fix):

1. Can we remove "This method can be used when this component is in a Gradio Blocks." from each docstring?
2. Many of the docstrings in the "Event arguments" section are copied from `Interface` are actually subtly incorrect. For example:

> the function to wrap an interface around. Often a machine learning model's prediction function. Each parameter of the function corresponds to one input component, and the function should return a single value or a tuple of values, with each element in the tuple corresponding to one output component.

should be:

> the function to call when this event is triggered. Often a machine learning model's prediction function. Each parameter of the function corresponds to one input component, and the function should return a single value or a tuple of values, with each element in the tuple corresponding to one output component.


 

<br class="Apple-interchange-newline">

aliabd:
@abidlabs thanks for the review! made all those changes, can you take a quick look?

abidlabs (approved):
Looks great!

PR #5313, opened 2023-08-23
Restores missing part of bottom border on file component

abidlabs:
Fixes the issue @freddyaboulton found. Test with:

```py
import gradio as gr

path = "cheetah.jpg"

with gr.Blocks() as demo:
    gr.File([path]*5, interactive=False)
    gr.File([path]*5, interactive=True)
    
demo.launch()
```


pngwn (approved):
lgtm! Thanks @abidlabs!

abidlabs:
Thank you @pngwn!

PR #5315, opened 2023-08-23
Support deleting columns/rows in `gr.DataFrame`

abidlabs:


js/dataframe/shared/Table.svelte
```diff
@@ -639,46 +639,26 @@
...
 		<div class="controls-wrap">
 			{#if row_count[1] === "dynamic"}
-				<span class="button-wrap">
-					<BaseButton variant="secondary" size="sm" on:click={() => add_row()}>
-						<svg
-							xmlns="http://www.w3.org/2000/svg"
-							xmlns:xlink="http://www.w3.org/1999/xlink"
-							aria-hidden="true"
-							role="img"
-							width="1em"
```
abidlabs, line 649: I replaced SVGs with HTML symbols because it was easier for me to work with for the new symbols. But if there are some advantages to using SVGs, let me know

pngwn: Yeah we don't really want these as ascii symbols because they could appear differently for people. SVGs give us consistency everywhere.

abidlabs: ah I see nice

abidlabs:
Will restart this later

PR #5316, opened 2023-08-23
Add ignore scripts to pnpm i in pypi action

aliabd:
It appears `pnpm i` runs the `js/_website` `pnpm prepare` command (`svelte-kit sync`), and that fails if jsons haven't been generated because it needs version.json. This was causing the deploy pypi action to fail.

Fix is to add `--frozen-lockfile --ignore-scripts` in `/build_pypi.sh`. 

Also removed `scripts/gh_action_pypi.sh` and the `deploy-pypi` and `deploy-client-pypi` workflows which are apparently not used.


abidlabs:
will try running the pypi release from this branch!

pngwn (changes requested):
We don't actually use this script any more, we can probably remove it. The offending code is in `/build_pypi.sh`. 

I would add two flags to this install command as well `--frozen-lockfile --ignore-scripts`.

pngwn (approved):
Looks good. Thanks for taking care of this @aliabd!

PR #5319, opened 2023-08-24
Fix: wrap avatar-image in a div to clip its shape

### LINKED ISSUE #5318: The avatars in chatbot don't appear round or square
### Describe the bug

The `border-radius` didn't actually take effect.

Even if it did take effect, the shape would be ugly with a non-1 aspect ratio image.

Moreover, I believe it's reasonable to allow passing data URLs to the `avatar_images` parameter for SVG usage.

I have already completed the coding, and I will submit a pull request for this.

### Reproduction

e.x. 
img url: https://www.apple.com.cn/leadership/images/overview/ceo_image.jpg

### Screenshot

radius setting failed:
image

when make radius work
image

### Logs

_No response_

### System Info

```shell
gradio 3.41.0
```

### Severity

I can work around it

Keldos-Li:
Closes #5318

## Description

image

The upper image is what it looks like before. If I set the image to one with a non-1 aspect ratio, although the maximum edge is limited to 35px, it looks strange and the rounded corners appear odd. Now the image has been stretched to cover the 35x35 parent container, constraining the avatar within a square.

Furthermore, the original code's `border-radius` didn't take effect, as it was overridden by another line of CSS. I added a `not` condition to address this.

Moreover, I want the avatar to be set as an SVG using a Data URL, for example:

image
image

So, I made some modifications to the `get_fetchable_url_or_file` function, but I'm not sure if this can be correctly referenced.

## 🎯 PRs Should Target Issues

- #5318

## Tests

I haven't tested it yet.


Keldos-Li:
Umm, would it be possible for someone to review this? I'd greatly appreciate it.🥺

abidlabs:
Setting via a base64 is not the best idea in this case since the browser won't be caching the repeated images. I think we should leave that out of this PR. 

Regarding the rounded borders, I'll let @dawoodkhan82 comment on that.

dawoodkhan82:
@Keldos-Li the avatar image changes look good to me. Thanks for making that fix

Keldos-Li:
Thank you! 

@abidlabs Umm, actually, I think Data URL and Base64 would be slightly different. Data URL is a more general concept that encompasses various types of data, not just Base64-encoded data. When I mean svg in Data URL, it actually takes a form similar to:
```html
<img src="data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg' width='100' height='100'><circle cx='50' cy='50' r='40' stroke='black' stroke-width='2' fill='red' /></svg>" alt="SVG Image" />
```
(while some characters should be percent-encoded as %xx.)
as opposed to:
```html
<img src="data:image/svg+xml;base64,PHN2ZyB4bWxucz0ia..." alt="SVG Image" />
```

`data:image/svg+xml`, not `data:image/svg+xml;base64`.

So, when I include SVG this way (without Base64), I think it's just like placing the raw svg code  into the HTML document for the browser, no need for base64 decoding, and caching might not be necessary.

However, I would agree that images shouldn't be added in base64 (which can be cumbersome and less efficient). And you are right, even if data url were acceptable , it might shouldn't be in this pull request anyway.

I think I'll push a commit to revert what I did for the `get_fetchable_url_or_file` then?

abidlabs:
Thanks you're right @Keldos-Li, but it would be better to keep this separate, as its adding a new functionality that we should document as well.

> I think I'll push a commit to revert what I did for the get_fetchable_url_or_file then?

That would be great!

Rest of the PR looks good I believe so we should be able to merge in

Keldos-Li:
> I think I'll push a commit to revert what I did for the get_fetchable_url_or_file then?
Done that~ 

（Hope you could consider data url when possible (´・ω・`) （It seems quite complicated to think about it like this.））

PR #5323, opened 2023-08-24
ensure dropdown stays open when identical data is passed in

### LINKED ISSUE #5314: Dropdown refilters/closes after every event listener completes
### Describe the bug

As [reported here](https://github.com/gradio-app/gradio/issues/2362#issuecomment-1690135315) as well as [internally](https://huggingface.slack.com/archives/C02QNREU24S/p1692826590897779), the `gr.Dropdown()` component refilters its choices (or closes altogether depending on your version of Gradio) when other, unrelated events complete.

This makes it impossible to use dropdowns if you have a very frequent events.

### Reproduction

```
import gradio as gr


with gr.Blocks() as demo:
    drop = gr.Dropdown(["Apple", "Banana", "Orange"])
    num = gr.Number(value=0)

    demo.load(lambda x: x, num, num, every=1)
      
demo.queue().launch()
```

Select an option, then click the dropdown box again.
### Screenshot

_No response_

### Logs

_No response_

### System Info

```shell
3.40.1
```

### Severity

I can work around it

pngwn:
## Description

Closes #5314.

This was happening because we pass data into component quite frequently, the issue is that Svelte has no way of knowing if the contents of an array have changed (without doing deep equality checks which are expensive).

I've added a simple check around the `filter` logic to make sure the `choices` have changed before we do anything.

There is a better solution to this which is to compile with `immutable: true`. This would require non primitive values to create new structures if the data changes. We actually do this anyway when passing data down but we don't do it inside the component, which means updating a component from outside works fine but Svelte thinks nothing has changed _inside_ of the components, so we never get fresh state.

If we tweak how we gather data to send it back to the server then we could use this `immutable` option which will prevent issues like this across all components, and propbably give us some small performance boosts too.

This fix is fine for now.

I've added a few tests to the Dropdown including one testing against this regression. I wrote the test before the fix and it definitely failed + passed after.

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/vast-wolves-kick.md
+++ b/.changeset/vast-wolves-kick.md
@@ -3,4 +3,4 @@
 "gradio": patch
 ---
 
-fix:ensure dropdown stays open when iodentical data is passed in
+fix:ensure dropdown stays open when identical data is passed in
--- a/.changeset/vast-wolves-kick.md
+++ b/.changeset/vast-wolves-kick.md
@@ -1,0 +1,6 @@
+---
+"@gradio/dropdown": patch
+"gradio": patch
+---
+
+fix:ensure dropdown stays open when iodentical data is passed in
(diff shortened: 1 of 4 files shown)


dawoodkhan82:
Tested and fixes the original issue. Thanks for the quick fix + tests! @pngwn

dawoodkhan82 (approved):

Issue #4420, opened 2023-06-05
Chatbot: limit width of chat bubbles to be the size of the messages


lucasjinreal:
Is there a way to add max width constraints to bubbles containing the messages?

![image](https://github.com/gradio-app/gradio/assets/21303438/8af56935-acf5-4a76-94fa-0e295a8292c5)

abidlabs:
Hi @lucasjinreal what do you mean by bubbles? Can you please provide example code to repro and explain what you are getting vs. what you are expecting?

lucasjinreal:
@abidlabs  Hello, its about ChatBot UI component.

1. As you can see, the bubbles in both incoming and outcoming are have a large free space：

![image](https://github.com/gradio-app/gradio/assets/21303438/4d1b1d1b-12e6-4ece-82b5-63ffdb295909)

Would it better to make it fit automatically to content just like iMessages?

2. My bottom have a TextBox and a Button, how can I make it align vertically?

abidlabs:
Thanks for the clarification, @lucasjinreal I've renamed the issue to reflect #1, it's something we'll think about!

lucasjinreal:
@abidlabs thank u.

I just find a repo using gradio made it, do u know what code they possible use? (nevertheless you are the creator of gradio!)

<img width="1162" alt="261efb275be94aec0368d4af46ed364" src="https://github.com/gradio-app/gradio/assets/21303438/0f610d48-a947-44d3-ba4d-05958283bfd2">

as you can see, their incoming bubble of mine, is limited width

abidlabs:
You can see their code here: https://huggingface.co/spaces/project-baize/chat-with-baize/blob/main/app.py

1ittlesnow:
`
with gr.Blocks(
        title="Chat with Open Large Language Models",
        theme=gr.themes.Base(),
        css=block_css,
    ) as demo:
`
When you create Blocks, there is a block_css. Ctrl click it. 
Change 'block_css'. 
Add [data-testid = "bot"] and [data-testid = "user"]. 
The property 'max_width' is its limit width of chat bubbles.


`block_css = (
    code_highlight_css
    + """
pre {
    white-space: pre-wrap;       /* Since CSS 2.1 */
    white-space: -moz-pre-wrap;  /* Mozilla, since 1999 */
    white-space: -pre-wrap;      /* Opera 4-6 */
    white-space: -o-pre-wrap;    /* Opera 7 */
[...]

PR #5334, opened 2023-08-24
Add chat bubble width param

dawoodkhan82:
## Description
Adds the highly requested param to change the chatbot message bubble width to fit message content.

Screenshot 2023-08-24 at 3 37 09 PM

Closes: #4420

## 🎯 PRs Should Target Issues

## Tests


gradio/components/chatbot.py
```diff
@@ -55,6 +55,7 @@ def __init__(
         show_share_button: bool | None = None,
         show_copy_button: bool = False,
         avatar_images: tuple[str | Path | None, str | Path | None] | None = None,
+        chat_bubble_fit_content: bool = False,
```
abidlabs, line 58: I'm wondering if we can shorten the name of the parameter? Maybe `bubble_full_width` (which is default `True`)

abidlabs:
LGTM @dawoodkhan82! Could we add a story for this as well?

austinmw:
Does this work for `gr.ChatInterface`?

abidlabs:
Not necessary, you can pass in a custom `Chatbot` instance as the `chatbot` parameter to `gr.ChatInterface` to get this behavior. 

See this Space by @yvrjsharma for example: https://huggingface.co/spaces/ysharma/falcon-180b-demo/blob/92739960e894d417f240b2ec6f55e92ab6b0c788/app.py#L124

lucasjinreal:
@abidlabs  thanks for this useful feature.

Is the possible add further a max_width here when full_width is false?

Issue #4907, opened 2023-07-13
ImageMask can not be updated


taoari:
### Describe the bug

ImageMask can not be updated. ImageMask returns a dict of 'image' and 'mask', however, this dictionary can not update the ImageMask component itself. See the following code, if the Image is not empty, an error will occur.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python

import gradio as gr

def bot(image, msg):
    return image, msg

with gr.Blocks() as demo:
    image = gr.ImageMask(type='pil', interactive=True)
    msg = gr.Textbox(show_label=False,
        placeholder="Enter text and press ENTER", container=False)
    msg.submit(bot, [image, msg], [image, msg])
    
demo.launch()
```

### System Info

```shell
(1) gradio 3.36.1 (2) Windows 11
```

### Severity

Blocking usage of gradio

Ra1Nik:
Are there any updates on this?

JalinWang:
I found where the bug comes from but have no idea about how to fix it.

![image](https://github.com/gradio-app/gradio/assets/20621784/d8fb7582-4e33-41b7-b66d-d6f8ffbe2be0)

During `postprocess`, the dict returned by `ImageMask` is processed by the function of `Image` which takes a PIL.Image

PR #5337, opened 2023-08-25
Support updating ImageMask by dict(image, mask)

JalinWang:
## Description

As reported in https://github.com/gradio-app/gradio/issues/4907, ImageMask can't be updated by a dict of 'image' and 'mask'. This PR tries to fix it.
I'm not sure if the function naming & docstring are ok. Please give me feedback if changes are needed :)

Closes: https://github.com/gradio-app/gradio/issues/4907

## 🎯 PRs Should Target Issues

## Tests


abidlabs:
Hi @JalinWang really appreciate the contribution! This looks good to me. The only thing I'd suggest is adding a unit test to make sure we don't undergo a regression. You can add a unit test in the same style as:

https://github.com/gradio-app/gradio/blob/a74605572dd0d6bb41df6b38b120d656370dd67d/test/test_components.py#L715

pngwn:
Thanks for the PR @JalinWang but unfortunately this doesn't work, it would need quite a few frontend change in order to make it work. The reason it _looks_ like it works using the repro from the original issue is because the mask lines are not removed when the value is changed.

Try this demo instead:

```py
import gradio as gr

def bot(image, msg):
    return image, msg

with gr.Blocks() as demo:
    image = gr.ImageMask(type='pil', interactive=True)
    msg = gr.Textbox(show_label=False,
        placeholder="Enter text and press ENTER", container=False)
    image2 = gr.ImageMask(type='pil', interactive=True)
    msg2 = gr.Textbox(show_label=False,
        placeholder="Enter text and press ENTER", container=False)

    msg.submit(bot, [image, msg], [image2, msg2])
    
demo.launch()
```

The background image is set but the mask strokes are not. There is currently no way this could possibly work until the frontend is changed. Partly because we process mask images anyway, so recreating the original image from the data sent to the server is impossible.

I think we should close this and take care of it when we tackle the Image component.

abidlabs:
Nice catch @pngwn -- yep let's close this in favor of fixing the underlying issue in the redesigned Image component.

PR #5342, opened 2023-08-25
virtualise the dataframe

pngwn:
## Description

Closes: #(issue)

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/js/dataframe/shared/Table.svelte
+++ b/js/dataframe/shared/Table.svelte
@@ -8,6 +8,7 @@
 	import EditableCell from "./EditableCell.svelte";
 	import type { SelectData } from "@gradio/utils";
 	import { _ } from "svelte-i18n";
+	import VirtualTable from "./VirtualTable.svelte";
 
 	type Datatype = "str" | "markdown" | "html" | "number" | "bool" | "date";
 
@@ -163,7 +164,7 @@
 
 	function get_sort_status(
 		name: string,
-		_sort: number,
+		_sort?: number,
 		direction?: SortDirection
 	): "none" | "ascending" | "descending" {
 		if (!_sort) return "none";
@@ -338,16 +339,8 @@
 	$: set_focus(selected, "select");
 
 	type SortDirection = "asc" | "des";
-	let sort_direction: SortDirection;
-	let sort_by: number;
-
-	function sort(col: number, dir: SortDirection): void {
-		if (dir === "asc") {
-			data = data.sort((a, b) => (a[col].value < b[col].value ? -1 : 1));
-		} else if (dir === "des") {
-			data = data.sort((a, b) => (a[col].value > b[col].value ? -1 : 1));
-		}
-	}
+	let sort_direction: SortDirection | undefined;
+	let sort_by: number | undefined;
 
 	function handle_sort(col: number): void {
 		if (typeof sort_by !== "number" || sort_by !== col) {
@@ -361,7 +354,7 @@
 			}
 		}
 
-		sort(col, sort_direction);
+		// sort(col, sort_direction);
 	}
 
 	let header_edit: string | false;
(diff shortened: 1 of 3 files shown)


js/dataframe/shared/Table.svelte
```diff
@@ -361,7 +354,7 @@
 			}
 		}
 
-		sort(col, sort_direction);
+		// sort(col, sort_direction);
```
abidlabs, line 357: Remove comment
```suggestion
```

js/dataframe/shared/Table.svelte
```diff
@@ -722,19 +808,22 @@
...
+		/* overflow: hidden; */
 		color: var(--body-text-color);
 		font-size: var(--input-text-size);
 		line-height: var(--line-md);
 		font-family: var(--font-mono);
+		border-spacing: 0;
 	}
 
-	table.dragging {
+	/* table.dragging {
```
abidlabs, line 824: Remove comment

abidlabs:
The vertical scrollbar is only visible if you've scrolled all the way to the right. I think it would be better if you didn't have to scroll all the way to the right, at least while scrolling. 

image

image

pngwn:
Something about the loading is wrong. I think we're supposed to have borders around the df, even when loading.

abidlabs:
Took another look, all of the bugs seem to have been resolved! The only thing that I could identify was that when you click on the "new row" button, the animation to load the new row is a bit too sloooow:

Test with:

```py
import gradio as gr

demo = gr.Interface(lambda x:x, gr.Dataframe(), gr.Dataframe())

demo.launch()
```

Recording 2023-08-30 at 13 11 25

Otherwise changes are 🔥

pngwn:
weird, will check that one out

abidlabs:
Stress tested the `gr.Dataframe`. Works great! Just a couple of small issues:

1. After a cell has been edited once, if you try editing it again, then as soon as you double click on it (or press enter), the content clear. This is annoying since you lose the content of the cell entirely and can't get it back. Also, if you edit a cell and then click outside of the cell, the contents of the cell clear and the cell does not lose focus.

Recording 2023-09-06 at 20 35 19

2. If you have selected a cell, and you sort a column *ascending downwards*, then you are taken to that cell after the sorting happens. However, if you sort a column *descending downwards*, then you are not taken to that cell. Easier to explain via gif:

Recording 2023-09-06 at 20 38 26

3. It seems that the calculation of the width/height is slightly off so that vertical and horizontal scrollbars *always* appear, even when they shouldn't be necessary. It's probably easiest to see this with the `demo/matrix_transpose/run.py` demo

image


Other than those, everything LGTM! Nice @pngwn

pngwn:
@abidlabs Could you try again now? I think those issues should be resolved now.

abidlabs:
Nice @pngwn! Can confirm all of those issues are fixed. Noticed one last thing -- if a DataFrame is inside of a Tab, and the Tab is not the initially active Tab, then the DataFrame's initial value does not render. 

* You can see this in the `matrix_transpose` demo in the Spaces Preview: https://huggingface.co/spaces/gradio-pr-deploys/pr-5342-all-demos

* Here's a local repro (note that the dataframe in the first tab appears fine, but the one in the second tab doesn't have any initial value):

```py
import numpy as np

import gradio as gr


def transpose(matrix):
    return matrix.T


demo1 = gr.Interface(
    transpose,
    gr.Dataframe(type="numpy", datatype="number", row_count=5, col_count=3),
    "numpy",
    examples=[
        [np.zeros((3, 3)).tolist()],
        [np.ones((2, 2)).tolist()],
        [np.random.randint(0, 10, (3, 10)).tolist()],
        [np.random.randint(0, 10, (10, 3)).tolist()],
        [np.random.randint(0, 10, (10, 10)).tolist()],
    ],
    cache_examples=False
)

demo2 = gr.Interface(
    transpose,
    gr.Dataframe(type="numpy", datatype="number", row_count=5, col_count=3),
    "numpy",
    examples=[
        [np.zeros((3, 3)).tolist()],
        [np.ones((2, 2)).tolist()],
        [np.random.randint(0, 10, (3, 10)).tolist()],
        [np.random.randint(0, 10, (10, 3)).tolist()],
        [np.random.randint(0, 10, (10, 10)).tolist()],
    ],
    cache_examples=False
)

demo = gr.TabbedInterface([demo1, demo2])

if __name__ == "__main__":
    demo.launch()

```

abidlabs:
Amazing @pngwn, thanks for addressing all of the issues! LGTM

abidlabs (approved):

pngwn:
Thanks for the detailed testing & review @abidlabs!

PR #5352, opened 2023-08-26
fix: error byte range

Codle:
## Description

When I run the stable diffusion webui, I found it will be slower and slower and there are many threads never stop. After doing further research, I found that there will be an infinite loop at the following code：

```python
# byte_range is (0, 1000), and len(byte_range) will be 1001
remaining_bytes = len(byte_range)
if not byte_range:
    await send(
        {"type": "http.response.body", "body": b"", "more_body": False}
    )
    return

while remaining_bytes > 0:

    # assume chunksize is 1000
    chunk_size = min(self.chunk_size, remaining_bytes)
    
    # for the first time, len(chunk) will read 1000, and never can read other data from file.
    # So, it will be 0 at the second time.
    chunk = await file.read(chunk_size)

    # remaining_bytes = 1001 - 1000 = 1 and will never change again.
    remaining_bytes -= len(chunk)
    await send(
        {
            "type": "http.response.body",
            "body": chunk,
            "more_body": remaining_bytes > 0,
        }
    )
```

After reading the following information, I think we should set the range.end to the total_length - 1:
b047686465edf99549bb5315c564610e

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/gradio/ranged_response.py
+++ b/gradio/ranged_response.py
@@ -102,7 +102,7 @@ class RangedFileResponse(Response):
                 if not stat.S_ISREG(mode):
                     raise RuntimeError(f"File at path {self.path} is not a file.")
 
-        byte_range = self.range.clamp(0, self.stat_result.st_size)
+        byte_range = self.range.clamp(0, self.stat_result.st_size - 1)
         self.set_range_headers(byte_range)
 
         async with aiofiles.open(self.path, mode="rb") as file:


Codle:
@aliabid94 can you help review this pr?

aliabid94:
Taking a look. FWIW I copied this code from https://gist.github.com/kevinastone/a6a62db57577b3f24e8a6865ed311463

aliabid94:
We implemented Ranged Responses specifically because iOS only supports Range Requests to load audio files. What browser and what request within the sd webui is sending ranged requests to the gradio backend? @Codle

abidlabs:
Yes as @aliabid94 said we’ll need a self-contained example with the issue for us to confirm that this fixes the issue @Codle

abidlabs:
Going to close this PR for now. @Codle please create an issue with the self-contained example so that we can examine (and potentially reopen) this PR.

Issue #5348, opened 2023-08-25
Dropdown Filter Cancel


deckar01:
### Describe the bug

The `Dropdown` component allows the options to be filtered through text input. If the input is changed then blurred without selecting an option, the input is cleared, a `null` value is sent to the backend, and the backend triggers `change`, `input`, and `blur` events with a `None` value. The `select` event is not fired despite the frontend changing the selected value.

The documentation for `allow_custom_value` establishes an expectation that `value` will always be a member of `choices`. This is likely to produce key errors in naive listeners.

When a user does not find the option they are searching for, the input should revert to the selected value and not produce an event, since no choice was made.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```py
import gradio as gr


with gr.Blocks() as demo:
    dropdown = gr.Dropdown(['a', 'b'], value='a')
    preview = gr.Label('a')

    dropdown.change(
        lambda choice: choice,
        inputs=dropdown,
        outputs=preview,
    )

demo.launch(debug=True)

```

### Screenshot

`change`, `input`, and `blur`

![image](https://github.com/gradio-app/gradio/assets/3108007/9c3f692e-cc8b-41cb-8656-ef0dd800e6bc)

`select`

![image](https://github.com/gradio-app/gradio/assets/3108007/1958c2b4-36d0-4bb9-a7ab-3eabe62d020a)

### Logs

```shell
{"data":["a"],"event_data":null,"fn_index":0,"session_hash":"..."}
{"data":[null],"event_data":null,"fn_index":0,"session_hash":"..."}
```

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Windows
gradio version: 3.40.1
gradio_client version: 0.4.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
aiohttp: 3.8.5
altair: 5.0.1
fastapi: 0.102.0
ffmpy: 0.3.1
gradio-client: 0.4.0
httpx: 0.24.1
huggingface-hub: 0.16.4
importlib-resources: 6.0.1
jinja2: 3.1.2
markdown-it-py: 2.2.0
markupsafe: 2.1.3
matplotlib: 3.7.2
mdit-py-plugins: 0.3.3
numpy: 1.25.2
orjson: 3.9.5
packaging: 23.1
pandas: 2.0.3
pillow: 10.0.0
pydantic: 1.10.12
pydub: 0.25.1
[...]

PR #5360, opened 2023-08-28
Cancel Dropdown Filter

deckar01:
## Description

When a `Dropdown(multiselect=False, allow_custom_value=False)` is filtered then blurred, the input value now unconditionally reverts to its selected value and no `change` or `select` events occur.

Closes: #5348


deckar01:
@abidlabs Thank you for your feedback on #5348. This should be ready for review.

Issue #4330, opened 2023-05-24
model3d style function is empty,  and can't control the size of model3d manually


shinetzh:
can not manual control the size of model3d

shinetzh:
can not manual control the size of model3d

abidlabs:
Hi @shinetzh can you please provide example code we can use to reproduce and explain what you expect vs. what the code is producing?

andreped:
If @shinetzh does not have a simple use case, I recently made a post at the forum regarding this (see [here](https://discuss.huggingface.co/t/change-size-and-alignment-of-annotatedimage-and-model3d-widgets/41305)).

More information can be found from the link (see [here](https://huggingface.co/spaces/andreped/livermask) to see the space directly), as well as a link to the source code and deployed app on spaces.

In summary I am attempting to place a 2D and 3D viewer side-by-side, and I do not see a simple way when using `gradio.Blocks` to place these side by side such that they are center aligned within respective boxes and scaled identally in size, which would make the demo app look much nicer. I believe this is the same issue as what @shinetzh has.

<img width="1541" alt="Screenshot 2023-05-29 at 13 23 21" src="https://github.com/gradio-app/gradio/assets/29090665/40e1bab9-67d7-4923-a230-0da84f5d60ec">

radames:
simple bug snippet
```python
css="""
#model-out {
  height: 512px;
}
#model-in {
  height: 512px;
}
"""
with gr.Blocks(css=css) as demo:
  with gr.Row():
    with gr.Column():
      inp = gr.Model3D(elem_id="model-in")
      btn = gr.Button("Hello")
    with gr.Column():
      out = gr.Model3D(elem_id="model-out")
```
expected 

```python
with gr.Blocks(css=css) as demo:
  with gr.Row():
    with gr.Column():
      inp = gr.Model3D().style(height=512)
      btn = gr.Button("Hllo")
    with gr.Column():
      out = gr.Model3D().style(height=512)
```
here is a notebook demo
https://colab.research.google.com/drive/1_IVwRLB5G5hu5Zdh3XzR2n1gzuYSxB9E

andreped:
Worked wonders, @radames! Cheers.

To address alignment, I just needed to add `margin: auto` to the css.
In my case I was using an `AnnotatedImage` as 2D viewer and `Model3D` as 3D viewer. Here is the final result:

[...]

Issue #989, opened 2022-04-13
`Model3D` component does not work well on mobile


dawoodkhan82:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Additional context**

pngwn:
It would be good to tweak the scroll-zoom a little if possible too, it zooms quite aggressively.

pngwn:
cc @gary149 for design guidance.

radames:
Possibly related to this issue. The Model3D viewer currently doesn't work well on mobile specially the pinch multi-touch gestures. Here is a playground example I've found that seems to work well on mobile https://playground.babylonjs.com/#4U6TVQ#3

gary149:
> Possibly related to this issue. The Model3D viewer currently doesn't work well on mobile specially the pinch multi-touch gestures. Here is a playground example I've found that seems to work well on mobile https://playground.babylonjs.com/#4U6TVQ#3

Nice find, just tried it on an iPhone and it works very well. Maybe let's try to get the exact same behavior for our 3d viewer @radames?

abidlabs:
I think the most important thing here is to make sure the Model3D component works well on mobile -- everything else can be part of a custom component (or if heavily requested, can be added to our Model3D component) after 4.0.

I'll rename the issue to focus on the mobile issue.

PR #5373, opened 2023-08-30
Adds `height` and `zoom_speed` parameters to `Model3D` component, as well as a button to reset the camera position

abidlabs:
Follow-up PR for `gr.Model3D` that:

* adds a `height` parameter (also makes the height of InteractiveModel3D the same as StaticModel3D -- they were previously different)
* disables panning (better for mobile-responsiveness)
* reduces the zoom speed (but allows this be configured as the best zoom speed may be different for different use cases)
* Based on review feedback, a button to reset the camera position 

I wrestled with adding stories for `Model3D` but unfortunately I don't think the Babylon JS viewer plays well within storybook.

Closes: #989 
Closes: #4330


dawoodkhan82 (approved):
LGTM, tested and zoom/height kwargs work great

dawoodkhan82:
We might have an issue for this already. But I think we should add functionality for a button/double clicking to reset the zoom/rotate of the object back to it's origin.

dawoodkhan82:
@abidlabs While testing, I noticed the `model3D-demo` has no demo, can we delete that in this pr

abidlabs:
Thanks @dawoodkhan82 for the review! I agree with the "reset camera" suggestion -- let me add it to this PR. 

What do you mean the "model3D-demo has no demo"? Are you referring to this file? https://github.com/gradio-app/gradio/blob/3d-fixes/demo/model3D/run.py

dawoodkhan82:
@abidlabs Actually nvm, I had an empty demo dir called `model3D_demo` but that just existed for me locally, not sure why

abidlabs:
@dawoodkhan82 added a "reset"/"undo" button to both the static and interactive -- lmk if you think the icon is intuitive or a different one would be better:

image

dawoodkhan82:
@abidlabs Tested the undo button on the `model3D` demo. Works great! Quick question: It seems like we're disposing the scene and adding the object back as a way of resetting the camera. Is this the best way? Will it be too expensive/slow when its a large 3d model file with meshes and textures?

abidlabs:
Good point! There's probably a better way, let me see

abidlabs:
Fixed so that only the camera is reset, but otherwise the canvas stays the same. Noticeably faster to reset now & also did some other cleanups. Thanks for the suggestion @dawoodkhan82!

dawoodkhan82 (approved):
lgtm

abidlabs:
Thanks for all of the reviews @dawoodkhan82 appreciate it!

longshiine:
Could you also add the option to set camera panning (true/false)? To use the ability to move objects from the application, we have built and used a edited source code...

abidlabs:
Hi @longshiine would you be interested in opening a PR for this? I think we could probably support this in the core component but it would be helpful to see the changes that you've made

longshiine:
Ah, it's not a big change at the moment.
Inside the create_camera of shared/utils.ts

// helperCamera.panningSensibility = 0;
helperCamera.attachControl(true, true, 2);

I just changed it to and built it.
So, I was wondering if you have any plans to add options like allow_panning to gr.model3D() in the future.

abidlabs:
Hi @longshiine the reason we disabled panning was that it ruined the experience of the 3D viewer on mobile. We could potentially add a flag, but there's probably a better way. We'll revisit this after we release 4.0 -- feel free to create a separate issue for this.

Issue #5247, opened 2023-08-16
Add a gradient that goes from translucent to transparent at the edge to indicate overflow content


abidlabs:
For several of our components, such as `Gallery`, `Chatbot`, and `File`, when a fixed height is set, we don't have a good way to indicate that there's more content that can be scrolled and viewed. 

Proposal: add a gradient that looks something like this:

![image](https://github.com/gradio-app/gradio/assets/1778297/9caa8d06-f7e3-4e70-b2b6-b36c8a29df25)

abidlabs:
Also as pointed by out @dawoodkhan82 in #5221,

> Setting the height to something really small (50px) kinda breaks the UI. Maybe we have a min-height?

We should handle the case of small heights more gracefully

abidlabs:
I'm rethinking whether this is actually necessary. As it turns out, [there is a setting on Macs](https://www.macrumors.com/how-to/make-scroll-bars-always-visible/) to hide scrollbars unless you are actively scrolling a div that has overflowed. If you disable this, our components show scrollbars that make it very clear that there is overflowing content:

<img width="781" alt="image" src="https://github.com/gradio-app/gradio/assets/1778297/3c042fa8-f02d-475a-a5b0-1c009a92bc4f">

If you enable this setting, then the scrollbars go away, potentially hiding the fact that there is overflowing content:

<img width="772" alt="image" src="https://github.com/gradio-app/gradio/assets/1778297/c68e3176-d621-46b1-a074-653ca9c4c01e">

But this the case for any website with overflowing divs. Here are some screenshots from other websites:

<img width="950" alt="image" src="https://github.com/gradio-app/gradio/assets/1778297/c8da1267-aaf4-4224-9bf5-ca87ff427087">

<img width="1016" alt="image" src="https://github.com/gradio-app/gradio/assets/1778297/bc3809b7-0598-404a-84ac-739aff142004">

Practically speaking, this has never posed a problem for me when using any other website so I'm wondering whether this problem may be inflated in our heads. An alternative option would be to display the total number of files for the `gr.File` component or the number of images in the `gr.Gallery` somewhere at the top, which might be generally useful as well as indicate to users that there are more files/images that they are not seeing.

Curious for your thoughts @pngwn @dawoodkhan82 @hannahblair @aliabid94

dawoodkhan82:
@abidlabs Good point. I don't think it's necessary then.

pngwn:
[...]

PR #5383, opened 2023-08-30
WIP: Adding overflow gradients

abidlabs:
#5247


abidlabs:
Closing given that we will not be pursuing #5247

PR #5386, opened 2023-08-31
Playground v1

aliabd:
Playground for gradio lite


abidlabs:
This should work in the [Vercel preview](https://gradio-pskckrzfe-hugging-face.vercel.app/demos) right? I wanted to play around with it, but right now I'm seeing an error:

image

aliabd:
Weird @abidlabs, can you try refreshing the page to get it to work? Not sure what this error is but will look into it.

abidlabs:
Sorry for the late response @aliabd -- I just tried it now, and it worked. *However* it seems that hard-refreshing was the problem, because if I try hard refreshing then it shows the same error as above. So there's definitely something flaky happening. But when it was working, I tested modifying the code for the Hello World example and it worked (!)

Like we discussed before, it could be nice to have a placeholder while gradio lite is loading (maybe the embedded demo) and then once it loads, the code becomes interactive. Happy to try this out some more when ready!

abidlabs (approved):
Gave [some internal feedback](https://huggingface.slack.com/archives/C02QNREU24S/p1697732724876949), but otherwise LGTM! Beautiful stuff @aliabd

PR #5390, opened 2023-08-31
Fix dataset demo

freddyaboulton:
## Description

Fixes the dataset demo that was breaking: https://huggingface.co/spaces/gradio/dataset_main

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/test/test_components.py
+++ b/test/test_components.py
@@ -522,8 +522,8 @@ class TestCheckboxGroup:
         cbox = gr.CheckboxGroup(choices=["a", "b"], value="c")
         assert cbox.get_config()["value"] == ["c"]
         assert cbox.postprocess("a") == ["a"]
-        with pytest.raises(BaseException):
-            gr.CheckboxGroup().as_example("a") == "a"
+        with pytest.raises(ValueError):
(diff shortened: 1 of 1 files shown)


abidlabs:
Isn't this the expected behavior? We should error out if the examples provided are not valid choices in the `gr.CheckboxGroup`

freddyaboulton:
Maybe checkbox group with empty choices is the culprit here. It looks really bad 

image

When would that be expected? As output component maybe?

abidlabs:
> Maybe checkbox group with empty choices is the culprit here. It looks really bad

> When would that be expected? As output component maybe?

It really never is expected. Even when a `CheckboxGroup` is an output component, it should be instantiated with the possible choices.

abidlabs:
The demo doesn't make sense -- it creates a `gr.CheckboxGroup` without any choices, yet the examples imply that there should be 3 options to select from

freddyaboulton:
Fixed the demo by passing choices for Checkboxgroup @abidlabs

abidlabs:
Nice! Thanks @freddyaboulton

Issue #5250, opened 2023-08-17
Add liking/voting functionality to `gr.Chatbot` messages


nguyenquocduongqnu:
I want to custom add like and dislike buttons to the chatbot's reply content as described in the image. Can you give me the solution. Thanks

![Untitled](https://github.com/gradio-app/gradio/assets/63329565/33f6fb9e-e865-46cf-99b9-9204bf393632)

PR #5391, opened 2023-08-31
Like/Dislike Button for Chatbot

dawoodkhan82:
## Description
Adds like/dislike events trigger and buttons for chatbot messages

Screenshot 2023-09-01 at 4 38 45 PM

Here's an example of how to use the "Like" event trigger:

```    
def on_like(evt: gr.LikeData):
    if evt.liked:
        print(f"The chat message {evt.value} was liked at {evt.index} from {evt.target}")
    if evt.liked:
        print(f"The chat message {evt.value} was disliked at {evt.index} from {evt.target}")

with gr.Blocks() as demo:
    chatbot = gr.Chatbot([])
    chatbot.like(on_like, <input>, <output>)
```
Closes: #5250

## 🎯 PRs Should Target Issues

## Tests


abidlabs:
Works very nicely @dawoodkhan82! 

In terms of the UI, are you thinking of putting the icons in the top right, similar to the copy button? 

image

dawoodkhan82:
> In terms of the UI, are you thinking of putting the icons in the top right, similar to the copy button?

Experimenting with a couple things to see what looks best. Probably just going to end up going with chatgpt UI
Screenshot 2023-08-31 at 6 04 12 PM

Shelliemayinforest:
hi,I  pull this pr and build with build_pypi.sh，but I skip the “aws”step because I can not install aws in my developing env.
image

then install with`  pip instal gradio-3.41.2.1-py3-none-any.whl`
it is successfully installed.
image

but when I run with my old programs, it seems to be blank page in the browser.
image

then I tryed the "hello world"demo, it has the same result.

can you please tell me does the "aws" step will the reason? and how can I deploy with this pr to try the like button.

thanks very much.

Shelliemayinforest:
hi,  is this a bug or is it my problem？
image

dawoodkhan82:
@Shelliemayinforest Currently fixing the UI.

dawoodkhan82:
@abidlabs ready for re-review

abidlabs:
Did something happen to the height of the chatbot bubbles @dawoodkhan82? I'm seeing this for a regular Chatbot:

image

And this is how the likable chatbot looks:

image

Test code:

```py
import gradio as gr

def test(x: gr.LikeData):

with gr.Blocks() as demo:
    c = gr.Chatbot([("abc", "def"), ("egh", "ijk")])
    c.like(test, None, None)
    
demo.launch()
```

abidlabs:
Super nit: could we put the thumbs icon before the copy button, and align all of the copy buttons to be at the same x-coordinate? So that the copy buttons for the user message and bot message are horizontally aligned?

image

abidlabs:
Finally, it would be great to add an example of how to use the `like()` method in the docs. 

Maybe we could add a section right after streaming in this Guide? https://www.gradio.app/guides/creating-a-custom-chatbot-with-blocks

dawoodkhan82:
Ah yeah I did mess up the height. Was testing with the copy button enabled. Will address in a bit

dawoodkhan82:
@abidlabs Ok the UI should be good now

dawoodkhan82:
@abidlabs Will update the guide in a separate PR

abidlabs:
LGTM @dawoodkhan82! I added a small section in our Chatbot guide for now -- though feel free to edit or expand it.

abidlabs (approved):

jeffchy:
How about adding a labeling field which accept comments similar to the like and dislike button?

abidlabs:
Hi @jeffchy you could add a `gr.Textbox` underneath the `gr.Chatbot` -- and use that as input in the `chatbot.like` event.

PrashantSaikia:
How do I store the like value in a json or yaml file? As of now, I think the like/unlike can only be clicked and it doesn't do anything else, am I right? What if I want to record the user preferences somewhere?

abidlabs:
Hi @PrashantSaikia, you can run arbitrary code (such as logging to a csv) within the function that is attached to the `.like()` event, e.g. `test()` in the example below:

```py
import gradio as gr

def test(x: gr.LikeData):

with gr.Blocks() as demo:
    c = gr.Chatbot([("abc", "def"), ("egh", "ijk")])
    c.like(test, None, None)
    
demo.launch()
```

PrashantSaikia:
Hi @abidlabs,

The problem with the `.like()` method is that it can only see the liked response, it does not see the query for whom the response is being given. I've written the following code to save the upvoting history:

```
import gradio as gr
import json, pandas, datetime

def append_row(df, row):
    return pandas.concat([
                df, 
                pandas.DataFrame([row], columns=row.index)]
           ).reset_index(drop=True)

def vote(data: gr.LikeData):
    df = pandas.DataFrame(columns=['Answer', 'Response', 'Time'])
    if data.liked:
        answer = data.value
        response = 1
        time = datetime.datetime.now().strftime("%I:%M%p on %B %d, %Y")

        new_row = pandas.Series({'Answer': answer, 'Response': response, 'Time': time})
        df = append_row(df, new_row)

        df.to_csv('temp.csv', index=False)
    else:
        answer = data.value
        response = -1
        time = datetime.datetime.now().strftime("%I:%M%p on %B %d, %Y")

        new_row = pandas.Series({'Answer': answer, 'Response': response, 'Time': time})
        df = append_row(df, new_row)

        df.to_csv('temp.csv', index=False)

    with open('temp.csv', 'r') as f1:
        temp = f1.read()

    with open('vote_history.csv', 'a') as f2:
        f2.write(temp)

with gr.Blocks() as ui:
    with gr.Row():
        with gr.Column():
            msg = gr.Textbox(label="", placeholder="Enter a question")

            chatbot = gr.Chatbot(label=" ")
            chatbot.like(vote, None, None)  # Adding this line causes the like/dislike icons to appear in your chatbot

            bot_message = retrieval_qa.run(message)

            with open('chat_history.json', 'w') as file:
                    json.dump(chat_history, file)

            return "", chat_history

ui.launch(server_name="0.0.0.0", server_port=8080)
```

As you can see, I'm having to save the chat hostory and the voting history separately, as the `like()` function can't see outside of the response that's being upvoted or downvoted. Ideally, I'd like to have the upvote history as a column in the chat history dump. 

Also, and perhaps this is outside the scope of my question strictly speaking, but as you can see my implementation of dumping the chat and upvote histories is real time. Ideally, I'd prefer to have a button that says "Download chat history", and upon clicking it, it would download the chat history including the upvoting history. How to do that?

abidlabs:
> Ideally, I'd like to have the upvote history as a column in the chat history dump.

If I understand correctly, you should be able to achieve this by also passing the chatbot as an input to your `vote` function. The chatbot input will pass in the history of messages:

```py
import gradio as gr

def test(x: gr.LikeData, history):
    print(history)

with gr.Blocks() as demo:
    c = gr.Chatbot([("abc", "def"), ("egh", "ijk")])
    c.like(test, c, None)
    
demo.launch()
```

PrashantSaikia:
> ```python
> import gradio as gr
> 
> def test(x: gr.LikeData, history):
>     print(x.index, x.value, x.liked)
>     print(history)
> 
> with gr.Blocks() as demo:
>     c = gr.Chatbot([("abc", "def"), ("egh", "ijk")])
>     c.like(test, c, None)
>     
> demo.launch()
> ```

With this, there is no way to know what was the query whose response was being upvoted or downvoted, right? That's what I want to be added, I guess some kind of handle would need to be there which shows which user query precedes the chatbot response. Am I correct or is it already there?

abidlabs:
Hi @PrashantSaikia you can get the query by looking at the index -- x.index will be something like [3, 1], which means the bot response (1) to the 4th message (since 0-indexed). If you look at the `history[3][0]`, that will be the user query

PrashantSaikia:
Ah, ok, got it now. So its these then:

```
query = history[x.index[0]][x.index[1]-1]
response = x.value
upvote = x.liked # true or false
```

abidlabs:
Yes that looks right

Issue #5355, opened 2023-08-26
problem with streaming a string which has latex code


redrocket8:
### Describe the bug

When I try to run [gradio/chatbot_multimodal](https://huggingface.co/spaces/gradio/chatbot_multimodal) then when a string has latex code then the latex rendered equation is not displayed, instead the code is streamed as the rest of the characters.

example string = **The** square root of 4 is 2 $\\sqrt{4}=2 $

I have set 

```
chatbot = gr.Chatbot(
        latex_delimiters=[{ "left": "$", "right": "$", "display": False }],
        elem_id="chatbot",
        avatar_images=(None, (os.path.join(os.path.abspath(''), "avatar.png"))),
    )
```

I have also tried example string = **The** square root of 4 is 2 $\\\\sqrt{4}=2 $ with no results

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
!pip install -q gradio 
import os
!wget -q https://github.com/gradio-app/gradio/raw/main/demo/chatbot_multimodal/avatar.png

import gradio as gr
import os
import time

# Chatbot demo with multimodal input (text, markdown, LaTeX, code blocks, image, audio, & video). Plus shows support for streaming text.


def add_text(history, text):
    history = history + [(text, None)]
    return history, gr.update(value="", interactive=False)


def add_file(history, file):
    history = history + [((file.name,), None)]
    return history


def bot(history):
    response = "**The** square root of 4 is 2 $\\\\sqrt{4}=2$"
    history[-1][1] = ""
    for character in response:
        history[-1][1] += character
        time.sleep(0.05)
        yield history
    history[-1][1]=""
    return history


with gr.Blocks() as demo:
    chatbot = gr.Chatbot(
        latex_delimiters=[{ "left": "$", "right": "$", "display": False }],
        elem_id="chatbot",
        avatar_images=(None, (os.path.join(os.path.abspath(''), "avatar.png"))),
    )

    with gr.Row():
        txt = gr.Textbox(
            scale=4,
            show_label=False,
            placeholder="Enter text and press enter, or upload an image",
            container=False,
        )
        btn = gr.UploadButton("📁", file_types=["image", "video", "audio"])

    txt_msg = txt.submit(add_text, [chatbot, txt], [chatbot, txt], queue=False).then(
        bot, chatbot, chatbot
    )
    txt_msg.then(lambda: gr.update(interactive=True), None, [txt], queue=False)
[...]

PR #5393, opened 2023-08-31
Renders LaTeX that is added to the page in `gr.Markdown`, `gr.Chatbot`, and `gr.DataFrame`

abidlabs:
Fixes: #5355. Doesn't break anything as far as I could test.

Try any of the repros in the original issue

Thanks @pngwn for the suggestion


aliabid94 (approved):
nice, lgtm

abidlabs:
Thanks @aliabid94!

PR #5395, opened 2023-08-31
Introduce decorator syntax for event listeners

aliabid94:
Added the ability to attach event listeners via decorators, e.g.

```python
with gr.Blocks() as demo:

    @greet_btn.click(inputs=name, outputs=output)
    def greet(name):
        return "Hello " + name + "!"
```

I think this syntax will be much more intuitive and maintainable, especially because the arguments of the function will more clearly correspond to their respective components since they are one line away from each other.


abidlabs:
Nice @aliabid94! Should we merge into `main` or keep on a separate `dev-4.0` branch? I'm inclined towards merging into `main` but not "popularizing" this too much until we release 4.0

aliabid94:
I actually think we can just merge to main and also publicize, it's not a big change just some syntactic sugar

demo/blocks_flipper/run.py
abidlabs, line 1: We shouldn't edit the `blocks_flipper` demo as it is used in the quick start (before the decorator syntax is introduced)

.changeset/fuzzy-numbers-repair.md
```diff
@@ -0,0 +1,5 @@
+---
+"gradio": minor
+---
+
+feat:Introduce decorator syntax for event listeners
```
abidlabs, line 5: Can we make this a highlight?

gradio/blocks.py
```diff
@@ -1626,7 +1626,7 @@ def get_time():
                 every=every,
                 no_target=True,
             )
-            return Dependency(self, dep, dep_index)
```
abidlabs, line 1629: Need to handle the the decorator case for `Blocks.load()` as well, right?

gradio/events.py
```diff
@@ -90,7 +95,7 @@ def __init__(
 
     def __call__(
         self,
-        fn: Callable | None,
+        fn: Callable | None | Literal["wrapper"] = "wrapper",
```
abidlabs, line 98: Just wondering, where is this "wrapper" coming from? I don't see it anywhere else in the code.

abidlabs (approved):

gradio/blocks.py
aliabid94: the Blocks.load method is such a nasty mess right now, we can add this later once we fix the load method in 4.0.

gradio/events.py
aliabid94: it's just the default value - if you don't set it, it acts like a decorator. There's no way for a function to know if it's being used as a decorator, so if you don't set fn, we assume it's a decorator. Previously fn had no default value so this is not a breaking change.

demo/blocks_flipper/run.py
aliabid94: ok

gradio/events.py
aliabid94: also renamed to "decorator" to be clearer

aliabid94:
@abidlabs addressed everything

abidlabs:
The syntax for the `highlight` changelog isn't quite correct. You have to use `####` for the heading

abidlabs:
Like this: https://github.com/gradio-app/gradio/pull/5279/files

Otherwise looks good!

abidlabs:
cool should be good to merge now

FarukOzderim:
Cool and convenient API and DX @aliabid94, like it!

Issue #2444, opened 2022-10-12
Input in HighlightedText


shahbuland:
- [*] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
As it stands highlighted text can only be used for output (i.e. in case of demonstrating a NER model). There could be a lot of use to adding interactivity to it. i.e. allowing users to highlight text and add labels/comments to it (similar to inline comments in a google doc). 

**Describe the solution you'd like**  
The ability to make HighlightedText interactive, such that a user can highlight text and click a button to write a label/comment for their highlighted text.

**Additional context**  
I'm working on CHEESE, a labeling/evaluation/HITL platform that runs primarily off gradio. Having this kind of feature would allow us to add NER labelling tasks or generally any task that could make use of inline labels/comments.

abidlabs:
Hi @shahbuland thanks for creating the issue! We've talked a little bit about this internally but haven't had a chance to implement. I do think this would be a good addition to the library to work on in the next few weeks. @dawoodkhan82 thoughts on this?

gsarti:
Any update on this? I'd also be interested in allowing users to highlight text

abidlabs:
For anyone who is interested, this is now easily possible! Here's a code snippet:

```py
import gradio as gr

with gr.Blocks() as demo:
    h1 = gr.HighlightedText([("Hello my name is ", None), ("Abubakar", "PER"), (" and I live in ", None), ("Palo Alto", "LOC")], interactive=True)
    h2 = gr.HighlightedText()
    h1.change(lambda x:x, h1, h2)
    
demo.launch()
```

abidlabs:
Hi @gsarti not exactly -- the purpose of the `gr.HighlightedText` component is for named-entity recognition, where you usually have multiple entities. However, you could get something like what you're describing by setting `show_legend=True` in which case the label appears in a legend and not next to each word:

```py
import gradio as gr

with gr.Blocks() as demo:
	h1 = gr.HighlightedText(
		[("Hello my name is ", None), ("Abubakar", "label"), (" and I live in ", None), ("Palo Alto", "label")],
		interactive=True,
		show_legend=True
	)

demo.launch()
```

which produces:

<img width="760" alt="image" src="https://github.com/gradio-app/gradio/assets/1778297/e118eb50-0345-473a-9dd3-886889696ba0">


[...]

PR #5400, opened 2023-09-01
Allow interactive input in `gr.HighlightedText`

hannahblair:
## Description

This PR allows gr.HighlightedText to be used as an input. 

- Supports categories mode and scores mode
- Double click on a word or highlight any selection of words within the value passed in to the component
    - A label will appear, with the text `label` by default (if in `categories` mode; will be 1 in `scores` mode). 
    - Clicking away from this label or pressing enter will set it to the value of the input, and clicking back onto the label will allow you to edit it again
    - You can't highlight text which already has a highlight; we don't support overlapping entities
    - If you select text across multiple labels, (e.g non-highlighted text before a highlight and text after a highlight) this intentionally won't do anything (to prevent any overlapping entity weirdness)
- Hovering on a label will display the remove button, and the selected highlight's corresponding label will be removed, and the value of that label will be `null`. If there are adjacent elements also with a label value of`null`, they will be merged
- Everything is accessible via the keyboard :tada:
    - You can try navigating solely via the keyboard with caret browsing mode in your browser (F7 in Chrome). You can tab to the text you want to highlight, press enter to highlight it, and tab to the remove label button to clear it. 
- The logic to merge adjacent values has been moved from `highlight_text.py` into the frontend, to a function called `merge_elements`. It takes a param `mergeMode` of type `"empty" | "equal"`. `mergeElements` is run with `"equal"` every time the value changes if `combine_adjacent` is `True`. 

Categories 
Kapture 2023-09-06 at 17 12 07

Scores
Kapture 2023-09-06 at 17 14 02

Legend
Kapture 2023-09-06 at 17 15 09

Keyboard navigation 
Kapture 2023-09-06 at 17 16 23

**Notes**

- Is clicking to edit the label and pressing enter to finish editing the label clear enough for users and/or keyboard users? There isn't any explicit instruction to do this, but to me it feels like natural UX. 
- Theres a bug right now which means we're not detecting whether the component is an input or output, so you have to use `Interactive=True`. I could use some help figuring out why this is 🐛 
- ~Issue with substring selection - [see comment](https://github.com/gradio-app/gradio/pull/5400#issuecomment-1708307656). Thoughts on this plz~

Demo: 

```
import gradio as gr 

def output_text(text):
    return text

with gr.Blocks() as demo:
    hl1 = gr.HighlightedText(value= [("hello", "greeting"), ('goodbye', "farewell"),("howdy", "greeting")], show_legend=True, interactive=True)
    hl2 = gr.HighlightedText(interactive=True)
    btn = gr.Button(value="submit")
    btn.click(output_text, hl1, hl2)

    gr.HighlightedText(value=[('dog', 1), ('cat', 2), ('mouse', 3)], show_legend=True, interactive=True)
    gr.HighlightedText(value=[('michael', "1"), ('jim', "2"), ('dwight', "3")], interactive=True)

if __name__ == "__main__":
    demo.launch()
```

to do:

- [x] general clean up
- [x] make this accessible via the keyboard + a11y compliant
- [x] clean up adjacent list items with blank category names which are creating spacing between elements
- [x] tweak scores mode to be editable as well
- [x] remove label button
- [x] make this work for score mode

Closes: #2444

## 🎯 PRs Should Target Issues

## Tests


abidlabs:
Very sleek @hannahblair! I played around with a little bit and love the UI/UX around removing and editing labels 😍 

Recording 2023-09-05 at 08 33 16

Noticed a couple of small things as I was testing:
* Editing a label currently does not trigger the `.change()` event (creating a label does)
* If I label a new entity with the same label as an old entity, it doesn't necessarily share the same color
* It seems that if I set `show_legend=True`, then I don't have the ability to edit new labels that I create
* The ability to select multiple words and label them with one entity works well most of the time. However, sometimes the new label doesn't pop up. In the example gif above I had a hard time reselecting and labeling "Palo Alto"

Code to repro:

```py
import gradio as gr

with gr.Blocks() as demo:
    h1 = gr.HighlightedText([("Hello my name is", None), ("Abubakar", "Name"), ("I live in ", None), ("Palo Alto", "City")], interactive=True, show_legend=True)
    h2 = gr.HighlightedText(show_legend=True)
    h1.change(lambda x:x, h1, h2)
    
demo.launch()
```

abidlabs:
> An outstanding challenge of this selection implementation is the breaking up of words, and joining them back together.
When I select Hi in["Hi there", null] , it'll turn to ["Hi", "label"], [there", null]. When the Hi label is removed again, it'll change to ["Hi there", null], which is fine - the merge logic adds a space between adjacent elements with equal labels.

Discussed with @hannahblair -- we have a fix for this. I'll comment in the code as well

js/highlightedtext/utils.ts
```diff
@@ -0,0 +1,68 @@
...
+
+export function merge_adjacent_empty_elements(
+	arr: HighlightValueType[]
+): HighlightValueType[] {
+	let result: HighlightValueType[] = [];
+	let temp: string | null = null;
+
+	for (const [str, val] of arr) {
+		if (val === null) {
+			temp = temp ? temp + " " + str : str;
```
abidlabs, line 53: I believe the fix being that we don't add a space here

abidlabs:
looks great @hannahblair! Tested a bunch of demos everything worked well for the case where labels are strings. 

For the case where labels are numbers, I couldn't get the interactive behavior to work as expected. As shown in the GIF, clicking on a span of text only removed the label, it didn't allow me to edit the label. I was also not able to add new labels:

Recording 2023-09-06 at 12 05 16


Code to reproduce:

```py
import gradio as gr

with gr.Blocks() as demo:
    h1 = gr.HighlightedText([("Hello my name is", None), ("Abubakar", 1), ("I live in ", 0.5), ("Palo Alto", 0.5)], interactive=True)
    h2 = gr.HighlightedText()
    h1.change(lambda x:x, h1, h2)
    
demo.launch()
```

gradio/components/highlighted_text.py
```diff
@@ -89,6 +88,8 @@ def __init__(
             elem_id=elem_id,
             elem_classes=elem_classes,
             value=value,
+            interactive=interactive,
+            combine_adjacent=combine_adjacent,
```
abidlabs, line 92: Don't need this here. Is causing a warning to be raised.
```suggestion
```

abidlabs:
Ah cool it's working now thanks @hannahblair!

A few nits I noticed:
* Can we reduce the "step size" of the numerical textbox from 1 to 0.1 (since the expected range is always -1 to 1)
* Pressing "enter" while focused on the numerical textbox should set the value (just like it does with the strings)

abidlabs:
Hmm editing string labels isn't working anymore. See:

Recording 2023-09-06 at 13 14 28

Code to reproduce:

```py
import gradio as gr

with gr.Blocks() as demo:
    h1 = gr.HighlightedText([("Hello my name is ", "abc"), ("Abubakar", "def"), (" and I live in ", "ghi"), ("Palo Alto", "abc")], interactive=True, show_legend=True)
    h2 = gr.HighlightedText()
    h1.change(lambda x:x, h1, h2)
    
demo.launch()
```

abidlabs:
Hi @hannahblair still seeing the error on my side (after pulling and rebuilding the frontend). I see an error in the console right after I edit the label:

```
Highlightedtext.svelte:198 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'secondary')
    at Object.p (Highlightedtext.svelte:198:33)
    at Object.p (Highlightedtext.svelte:189:27)
    at Object.p (Highlightedtext.svelte:188:11)
    at Object.p (Highlightedtext.svelte:187:10)
    at Object.p (Highlightedtext.svelte:348:6)
    at Wi (scheduler.js:119:30)
    at To (scheduler.js:79:5)
```

GIF:

Recording 2023-09-07 at 07 39 30

gradio/components/highlighted_text.py
```diff
@@ -49,6 +45,7 @@ def __init__(
         visible: bool = True,
         elem_id: str | None = None,
         elem_classes: list[str] | str | None = None,
+        interactive: bool = False,
```
freddyaboulton, line 48: > Theres a bug right now which means we're not detecting whether the component is an input or output, so you have to use Interactive=True

Set the default value to None and Blocks.svelte will set it to be interactive if it's used as input to an event!

freddyaboulton:
Looks good @hannahblair ! I pulled and cloned the latest commit and I'm seeing weird (but separate) behavior IN @abidlabs example.

Removing the label from "Abubakar" fuses it with "is" and there's no way to separate the two words.

highlight_text_bug

> The logic to merge adjacent values has been moved from highlight_text.py into the frontend,

I see the logic is still in the backend. TBH I think that makes sense otherwise API usage of the highlight text component would break.

abidlabs:
Awesome @hannahblair! It looks like all of the issues have been fixed. I'll go ahead and merge so that we can get it out in a release

abidlabs (approved):

Issue #5403, opened 2023-09-02
Spaces API Fetcher 500


deckar01:
### Describe the bug

The login page sends app configurations to the [Spaces API Fetcher](https://huggingface.co/spaces/gradio/space-api-fetcher-v2/tree/main) even when `SYSTEM` is not `spaces`.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```py
import gradio as gr

with gr.Blocks() as demo:
    pass

demo.launch(auth=('a', '1'))
```

### Logs

```shell
Request URL: https://gradio-space-api-fetcher-v2.hf.space/api
Request Method: POST
Request Body: {"serialize":false,"config":"{"auth_required":true,"auth_message":null,"space_id":null,"root":"http://localhost:7860","path":""}"}

Status Code: 500
Response Body: {"detail":"'dependencies'"}
```

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Windows
gradio version: 3.42.0
gradio_client version: 0.5.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.0.1
fastapi: 0.99.1
ffmpy: 0.3.1
gradio-client==0.5.0 is not installed.
httpx: 0.24.1
huggingface-hub: 0.16.4
importlib-resources: 6.0.1
jinja2: 3.1.2
markupsafe: 2.1.3
matplotlib: 3.7.2
numpy: 1.25.2
orjson: 3.9.5
packaging: 23.1
pandas: 2.0.3
pillow: 10.0.0
pydantic: 1.10.12
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0.1
requests: 2.31.0
semantic-version: 2.10.0
typing-extensions: 4.7.1
uvicorn: 0.23.2
websockets: 11.0.3
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2023.6.0
httpx: 0.24.1
huggingface-hub: 0.16.4
packaging: 23.1
requests: 2.31.0
typing-extensions: 4.7.1
websockets: 11.0.3
```

### Severity

I can work around it

freddyaboulton:
This is done to fetch the contents of the View Api page

deckar01:
[...]

PR #5412, opened 2023-09-05
Skip view_api request in js client when auth enabled

freddyaboulton:
## Description

Closes: #5403

## 🎯 PRs Should Target Issues

## Tests


pngwn:
The only issue with this is that we use the view api payload to correctly upload binary data where necessary. 

I don't think the client works for auth enabled apps anyway though, so it might not make a difference. This won't affect gradio itself (yet) because gradio components upload binary data early, rather than sending it to the client.

freddyaboulton:
@pngwn I think we're ok. `auth_required` is only set in the config when the user has not authenticated yet. Upon successful authentication, the page is refreshed and a new client is created. Since `auth_required` is not set in the config, the api will be fetched.

Relevant backend code in routes.py

```python
            if app.auth is None or user is not None:
                config = app.get_blocks().config
                config["root"] = root_path
            else:
                config = {
                    "auth_required": True,
                    "auth_message": blocks.auth_message,
                    "space_id": app.get_blocks().space_id,
                    "root": root_path,
                }
```

pngwn (approved):
Looks great! Thanks @freddyaboulton!

freddyaboulton:
Thanks for the review @pngwn !

Issue #4862, opened 2023-07-10
Auto scroll to bottom in components 


yvrjsharma:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
Textbox doesn't scroll to the bottom or the latest line if we use it to populate real-time data. This behavior is consistent with Markdown and HTML components too.

**Describe the solution you'd like**  
If I am populating my logs in a textbox (/markdown/html), having the component automatically scroll to the latest line will be useful. This is similar to how it happens in a Chatbot. The Chatbot component scrolls to the newest message.

**Additional context**  
https://github.com/gradio-app/gradio/assets/48665385/dd11b36c-489b-46ff-9a0e-4167b0f00cc3

abidlabs:
Would probably solve this as well: https://github.com/gradio-app/gradio/issues/4044

SpenserCai:
I think you can use Interval to do that
```python
gr.Textbox(lines=25, readonly=True, elem_id="idxx")
start_jscode = """
                function() {
                    if (window.iidxx) {
                        window.clearInterval(window.iidxx);
                    }
                    var dividxx = document.getElementById('idxx');
                    var textarea = dividxx.querySelector('textarea');
                    window.iidxx = window.setInterval(function() {
                        textarea.scrollTop = textarea.scrollHeight;
                    }, 500);
                    
                }
"""
start_button = gr.Button("Start")
start_button.click(inputs=[....],outputs=[....],fn=start_bot,_js=start_jscode)
```

Node-imapsalient11:
https://transvestite72.github.io/shooter/



See this answer for details

Issue #4044, opened 2023-05-02
Scroll not working when stream text generation


miguelwon:
### Describe the bug

Hi,
I'm working in a summariser demo, where I'm streaming  OpenAI generation of text to a TextBox.

The generated text will be quite large and at some point the I want to be able to scroll down the TextBox do be able to check the live stream. 

The problem is that the scroll only works when the stream stops, i.e., when full text has been generated. 

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```
import gradio as gr
import time

TEXT = """What's the Hugging Face Hub?
We are helping the community work together towards the goal of advancing Machine Learning .

The Hugging Face Hub is a platform with over 120k models, 20k datasets, and 50k demos in which people can easily collaborate in their ML workflows. The Hub works as a central place where anyone can share, explore, discover, and experiment with open-source Machine Learning.

No single company, including the Tech Titans, will be able to “solve AI” by themselves – the only way we’ll achieve this is by sharing knowledge and resources in a community-centric approach. We are building the largest open-source collection of models, datasets, demos and metrics on the Hugging Face Hub to democratize and advance ML for everyone .

We encourage you to read the Code of Conduct and the Content Guidelines to familiarize yourself with the values that we expect our community members to uphold .

What can you find on the Hub?
The Hugging Face Hub hosts Git-based repositories, which are version-controlled places that can contain all your files. 

On it, you’ll be able to upload and discover…

Models, hosting the latest state-of-the-art models for NLP, vision, and audio tasks
Datasets, featuring a wide variety of data for different domains and modalities..
Spaces, interactive apps for demonstrating ML models directly in your browser.
The Hub offers versioning, commit history, diffs, branches, and over a dozen library integrations! You can learn more about the features that all repositories share in the Repositories documentation.

Models
You can discover and use dozens of thousands of open-source ML models shared by the community. To promote responsible model usage and development, model repos are equipped with Model Cards to inform users of each model’s limitations and biases. Additional metadata about info such as their tasks, languages, and metrics can be included, with training metrics charts even added if the repository contains TensorBoard traces. It’s also easy to add an inference widget to your model, allowing anyone to play with the model directly in the browser! For production settings, an API is provided to instantly serve your model.

To upload models to the Hub, or download models and integrate them into your work, explore the Models documentation. You can also choose from over a dozen libraries such as Transformers, Asteroid, and ESPnet that support the Hub.

Datasets
[...]

PR #5417, opened 2023-09-05
Auto scroll to bottom of textbox

dawoodkhan82:
## Description
Auto scroll to bottom of textbox.

Closes: #4862 
Closes: #4044

## 🎯 PRs Should Target Issues

## Tests


abidlabs:
Nice @dawoodkhan82! Works great.

One comment -- we create the `autoscroll` prop in the frontend but we don't allow users to set it in the backend. How about we add the `autoscroll` attribute in the `gr.Textbox()` class as well (can be `True` by default)?

As an aside, I believe this closes: https://github.com/gradio-app/gradio/issues/4044 as well. The user wanted to scroll down to the bottom of the textbox to ensure that their textbox was still streaming and this change handles the underlying concern.

dawoodkhan82:
@abidlabs Any reason why a user would want to disable auto scroll? I feel like our attributes are getting too granular.

abidlabs:
No reason comes to mind, but that's what I thought you had intended since you defined a prop (`autoscroll`) in the frontend and check if it's true/false.

dawoodkhan82:
@abidlabs Oh I see, I can rename. But that param is keeping track to see if the textbox _needs_ to be scrolled.

abidlabs:
oh sorry I misunderstood -- my bad we're good, nvm!

abidlabs (approved):

w-e-w:
> No reason comes to mind, but that's what I thought you had intended since you defined a prop (autoscroll) in the frontend and check if it's true/false.

I can think of some reasons why one wants to be able to disable this or otherwise customize the behavior

basically "any time when the last row of text is not the point of interest"

I know of a [project](https://github.com/AUTOMATIC1111/stable-diffusion-webui-extensions/pull/179) of a frend of my that uses Textbox as constantly updating log of activities
If I understand correctly this code would make it not possible to scroll up and look at what's above when thers an update to the contacts
in his use case he has implemented a auto scroll `start/stop` https://github.com/SpenserCai/sd-webui-discord-ex/blob/82f247df77aaa8f13cf4b5fd6f6d6e5f5293351a/scripts/ui_bot_mgr.py#L73-L99

or imagine if the logs is in reverse order, the newest line appears at the top
this would force one to look at the most bottom line which is undesirable

ideally not only one would want the ability to disable the, better if it's able to customise the direction
better if it behaviour like a terminal (only auto scroll when the pointer is at the bottom and disable auto scroll when the user scroll up look at other parts of text)

abidlabs:
I agree with @w-e-w here. I actually think that autoscrolling on the Textbox may not always be a good idea since the Textbox can be used to output text all at once, where the user probably wants to start reading from the beginning. e.g. an example like this:

```py
import gradio as gr

value = "Lorem Ipsum is simply dummy text of the printing and typesetting industry. Lorem Ipsum has been the industry's standard dummy text ever since the 1500s, when an unknown printer took a galley of type and scrambled it to make a type specimen book. It has survived not only five centuries, but also the leap into electronic typesetting, remaining essentially unchanged. It was popularised in the 1960s with the release of Letraset sheets containing Lorem Ipsum passages, and more recently with desktop publishing software like Aldus PageMaker including versions of Lorem Ipsum."

gr.Interface(lambda x:value, "textbox", gr.Textbox(max_lines=4)).launch()
```

I think we should add an `autoscroll` parameter to `gr.Textbox()` to control this behavior @dawoodkhan82

dawoodkhan82:
@w-e-w @abidlabs Oh yeah makes sense. Will add!

yummy527:
How to scroll a embedded gr.HTML nested in gr.Column?

Issue #5421, opened 2023-09-05
Add new optional placeholder arg to ChatInterface


bcattaneo:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
No.

**Describe the solution you'd like**  
Add an optional arg _placeholder_ to `ChatInterface`, which should default to "Type a message...". Use such arg on the default `TextBox` (when user defined textbox is undefined).

**Additional context**  
This is a nice feature because it allows to fully customize the chat interface. We're able to modify pretty much any other component within a chat interface but this one. I've a PR ready and will attach to this issue now.

abidlabs:
Hi @bcattaneo thanks for creating this issue! We actually don't need a separate parameter because you can pass your own instance of a `gr.Textbox` inside `gr.ChatInterface` which means you have full customizability over the appearance.

Here's a quick example:

```py
import gradio as gr

textbox = gr.Textbox(placeholder="type your funky message here", scale=7)

gr.ChatInterface(lambda x,y:x, textbox=textbox).launch()
```

PR #5422, opened 2023-09-05
Add new optional placeholder arg to ChatInterface

bcattaneo:
## Description

Adds an optional arg _placeholder_ to `ChatInterface`, which defaults to "Type a message...". Uses such arg on the default `TextBox` (when user defined textbox is undefined).

Closes: #5421


abidlabs:
Thanks @bcattaneo!

PR #5423, opened 2023-09-06
Remove stable diffusion demo from landing page

aliabd:
Remove stable diffusion demo from landing page due to issues with embedding it.


abidlabs:
thanks @aliabd! we can bring it back once we support embedding Spaces w/ replicas

abidlabs (approved):

Issue #5434, opened 2023-09-06
`show_api` doesn't work when queue is enabled


abidlabs:
### Describe the bug

Setting `show_api=False` doesn't have any effect if queueing is enabled 

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```py
import gradio as gr

with gr.Blocks() as demo:
    gr.Dropdown(["name 1", "name 3"])
    
demo.queue().launch(show_api=False)
```

### System Info

```shell
gradio 3.42.0
```

### Severity

I can work around it

PR #5436, opened 2023-09-06
api_open does not take precedence over show_api 

freddyaboulton:
## Description

The two parameters now operate independently. If api_open=False, then the api_page will still be shown. This is a slight change in behavior but users still have the ability to hide them if they wish. If we automatically hide the api docs, user's don't have a way to override that. 

Closes: #5434 

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -1778,7 +1778,7 @@ Received outputs:
             ssl_keyfile_password: If a password is provided, will use this with the ssl certificate for https.
             ssl_verify: If False, skips certificate validation which allows self-signed certificates to be used.
             quiet: If True, suppresses most print statements.
-            show_api: If True, shows the api docs in the footer of the app. Default True. If the queue is enabled, then api_open parameter of .queue() will determine if the api docs are shown, independent of the value of show_api.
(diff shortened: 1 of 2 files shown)


gradio/blocks.py
```diff
@@ -1778,7 +1778,7 @@ def launch(
             ssl_keyfile_password: If a password is provided, will use this with the ssl certificate for https.
             ssl_verify: If False, skips certificate validation which allows self-signed certificates to be used.
             quiet: If True, suppresses most print statements.
-            show_api: If True, shows the api docs in the footer of the app. Default True. If the queue is enabled, then api_open parameter of .queue() will determine if the api docs are shown, independent of the value of show_api.
+            show_api: If True, shows the api docs in the footer of the app. Default True.             show_api: If True, shows the api docs in the footer of the app. Default True. If the queue is enabled and api_open=True, then api docs are not shown, independent of the value of show_api.
```
abidlabs, line 1781: Docstring repeated here

abidlabs (approved):
Just the nit about docstring, otherwise lgtm!

freddyaboulton:
Thanks for the review @abidlabs!

PR #5480, opened 2023-09-11
Prevent race conditions when deploying website demos

aliabd:
There are currently two issues with deploying website demos, both occur when a new version is released: 

- The demos are uploaded before the version is published to pypi. I fixed this by adding a check in the upload script to see if version is live. It will check every second for 10 minutes, and then fail. 
- There's a mismatch between the uploaded gradio wheel to aws, and the one referenced in the demo requirements, causing spaces to fail because the url doesn't exist. This is because the uploaded wheel uses the old version in the url (it gets it from pypi not version.txt), while the requirements file uses the current version. I assume the wheel uses the pypi json so it can work on PRs that change the version.txt. So to fix this I changed the build pr action metadata to read from version.txt only when the workflow is triggered on main. And I read from the same argument in the website upload script so there's no mismatch. 

Tested both of these on this [fork](https://github.com/aliabd/gradio) and they worked as expected.


freddyaboulton:
I think you can test by changing the project name field in pyproject.toml to something like `gradio-test-upload` ?

scripts/upload_website_demos.py
```diff
@@ -95,15 +93,28 @@ def upload_demo_to_space(
...
 demos = [demo for demo in demos if demo not in DEMOS_TO_SKIP and os.path.isdir(os.path.join(GRADIO_DEMO_DIR, demo)) and  os.path.exists(os.path.join(GRADIO_DEMO_DIR, demo, "run.py"))]
 
+def version_exists(version: str):
+    """Checks if the given version of gradio has been uploaded to pypi"""
+    for _ in range(600):
+        if requests.get("https://pypi.org/pypi/gradio/json").json()["info"]["version"] == version:
+            return True
+        time.sleep(1)
+    return False
+
```
pngwn, line 103: I don't think we should do this, as it will hog the runner until the pypi publish has completed, possibly forever if the publish fails/ doesn't trigger.

pngwn (commented):
Left a comment about why I'm not sure this is the best approach. What I think we should do instead is the following:

- add a `workflow_dispatch` trigger to the deploy-pr-to-spaces workflow.
- skip deploy-pr-to-spaces if the current branch is main.
- Trigger the deploy-pr-to-spaces workflow programatically (via the github API) at the end of the run. This will handle both `main` deploys and latest release deploys.

It will make our main deploys a little slower but is much safer i think.

Wat do you think @freddyaboulton?

aliabd:
@pngwn this makes sense to me

scripts/upload_website_demos.py
aliabd: we set the timeout so it won't go forever but you're right it will hog the runner. workflow dispatch makes more sense

pngwn:
`workflow_dispatch` can also accept inputs, so if you need to pass anything around you have that option.

aliabd:
@pngwn did what you suggested. Here's the new flow: 

- build-pr is triggered either by a pull_request or workflow_dispatch. publish-npm programmatically calls build-pr. deploy-pr-to-spaces runs when build-pr completes. 

Tested on a fork (though simulated the publishing by just letting it sleep) and everything works as expected: 

- these [build-pr](https://github.com/aliabd/gradio/actions/runs/6218593456) and [deploy-pr-to-spaces](https://github.com/aliabd/gradio/actions/runs/6218613519/job/16875353265) ran correctly on a [pull request](https://github.com/aliabd/gradio/pull/11). 
- these [build-pr](https://github.com/aliabd/gradio/actions/runs/6218664292) and [deploy-pr-to-spaces](https://github.com/aliabd/gradio/actions/runs/6218677928/job/16875477201) were triggered by this [publish-npm ](https://github.com/aliabd/gradio/actions/runs/6218634886/job/16875365523) and ran correctly after the pr was merged.

aliabd:
Also the pr that removed version.txt removed [this line](https://github.com/gradio-app/gradio/pull/5514/files#diff-15b4de36ac797b24d56f24c4403e979f20eabd06c3d49f3ffdceb0ebec3c0fb2) from build-pr, but this is necessary so that demos are uploaded correctly on a PR that changes versions. I added it back (using package.json) but @freddyaboulton lmk if this makes sense to you. also tested this on my fork

.github/workflows/build-pr.yml
```diff
@@ -38,15 +36,17 @@ jobs:
...
+          echo "GRADIO_VERSION=$(python -c 'import requests;print(requests.get("https://pypi.org/pypi/gradio/json").json()["info"]["version"])')" >> $GITHUB_OUTPUT
           python -c "import os;print(os.environ['GITHUB_REF'].split('/')[2])" > pr_number.txt
           echo "PR_NUMBER=$(cat pr_number.txt)" >> $GITHUB_OUTPUT
         else 
+          echo "GRADIO_VERSION=$(python -c 'import json; print(json.load(open("gradio/package.json"))["version"])')" >> $GITHUB_OUTPUT
           echo "PR_NUMBER='main'" >> $GITHUB_OUTPUT
         fi
     - name: Build pr package
       run: |
+        python -c 'import json; j = json.load(open("gradio/package.json")); j["version"] = "${{ steps.get_pr_number.outputs.GRADIO_VERSION }}"; json.dump(j, open("gradio/package.json", "w"))'
```
freddyaboulton, line 49: This makes sense @aliabd ! Thanks for the fix

pngwn (approved):
Looks good! Thanks for figuring this out @aliabd!

PR #5482, opened 2023-09-11
Update links to no longer give error 404

sebastianbadea197:
## Description

Previously the link towards "building with blocks" from the quickstart page was giving a 404 error. Now it correctly links to the appropriate tutorial. 

Closes: #(issue)
The fix is very small and therefore no issue was created for it.

## Tests
Because the fix is very small and part of the documentation, no tests were run.


sebastianbadea197:
> Thanks for catching and fixing this @sebastianbadea197!
> 
> We also have instances of the broken link in `/gradio/guides/cn/01_getting-started/01_quickstart.md` and `/gradio/readme_files/zh-cn/README.md` as well. Would you mind updating those too? 🙂
> 
> Edit: Ah, we also have redirects to `/building_with_blocks` in place in `vercel.json` and `redirects.js`

Yes! Will do

sebastianbadea197:
Hey @hannahblair, would it be best to only modify vercel.json and redirects.js so that the links from "building_with_blocks" redirect automatically to "blocks-and-event-listeners" instead of modifying the .MD directly?

aliabd (approved):
@sebastianbadea197 Thanks for the contribution! This is fine as is, will merge.

Issue #5486, opened 2023-09-11
Specifying same-type component as input to interface within block scope breaks


osanseviero:
### Describe the bug

Easier to explain with code below :)

When defining an input as `gr.Slider()` as input to an `Interface`, I get a DuplicateBlockError

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

This code works
```
with gr.Blocks() as demo:
  iface = gr.Interface(fn=greet, inputs=["slider", "slider"], outputs="text")
demo.launch()
```

This code works

```
iface = gr.Interface(fn=greet, inputs=[gr.Slider(), gr.Slider()], outputs="text")
with gr.Blocks() as demo:
  iface.render()
demo.launch()
```

This code breaks with `DuplicateBlockError`
```
with gr.Blocks() as demo:
  iface = gr.Interface(fn=greet, inputs=[gr.Slider(), gr.Slider()], outputs="text")
demo.launch()
```

```
DuplicateBlockError: A block with id: 268 has already been rendered in the current Blocks.
```

### System Info

```shell
!gradio environment
```

### Severity

I can work around it

abidlabs:
Thanks @osanseviero! The issue is that when you create a component (`gr.Slider()`) that is inside the `gr.Interface()` as well as inside the `gr.Blocks()` -- both the Interface and Blocks try to render it, leading to the error. 

The problem is the `gr.Blocks()` "knows about it first" which means that you'd have a Block with two sliders before the Interface object is even created.

One idea here could be to give priority to the `gr.Interface` --> if a Gradio component is rendered, and it appears again inside of a `gr.Blocks()`, then it should be unrendered from the Blocks and re-rendered inside the Interface. I *think* that's pretty much always the desired behavior.

PR #5496, opened 2023-09-12
Allow interface with components to be run inside blocks

abidlabs:
Small change that allows us to create an `Interface` with components inside of a `Blocks`.

```py
with gr.Blocks() as demo:
  iface = gr.Interface(fn=lambda x,y: x, inputs=[gr.Slider(), gr.Slider()], outputs="text")
demo.launch()
```

Note that if a component is created before an Interface class is defined, then it is unrendered and re-rendered inside of the gr.Interface. E.g. 

```py
import gradio as gr

with gr.Blocks() as demo:
    s = gr.Slider()
    iface = gr.Interface(fn=lambda x, y:x, inputs=[s, gr.Slider()], outputs="text")

demo.launch()
```

(previously both of these would raise an error).

Fixes: #5486


aliabid94 (approved):
this makes sense, tested and works. lgtm

abidlabs:
Thanks @aliabid94!

PR #5508, opened 2023-09-12
Adds a `filterable` parameter to `gr.Dropdown` that controls whether user can type to filter choices

abidlabs:


dawoodkhan82:
Looks good. Maybe we should return a warning when a user tries to set `allow_custom_value=True` and  `filterable=False`. Also right now when `allow_custom_value=True` and  `filterable=False` is set, typing is disabled but I think filterable should be ignored and typing should be enabled.

abidlabs:
Makes sense @dawoodkhan82, I'll make the changes

abidlabs:
Made the changes & fixed tests -- merging

pseudotensor:
Thanks!

Issue #5489, opened 2023-09-11
Remove version.txt from gradio python packages


freddyaboulton:
- [x] I have searched to see if a similar issue already exists.

Right now, we have a `version.txt` for each python package in the repo. This is used for hatch, our build tool, to know the right version for the build. We also have a package.json file that stores the same version number as `version.txt`. I believe that's used by changesets.

We should just use the package.json. Should be possible to configure hatch to read from the json file via a regex pattern (https://hatch.pypa.io/latest/version/#configuration).

This is a good first issue, but we need to make sure with @pngwn that this doesn't break the changesets release (which I believe modifies the `version.txt` file).

**Describe the solution you'd like**  

**Additional context**

pngwn:
We will need to remove the logic that creates/ updates the txt file but it only writes it for the python build process. So shouldn't be an issue.

DarhkVoyd:
I'm a newcomer to the project and I'd like to work on this issue.
I'm eager to help with this, so please let me know how I can get started. Thanks!

pngwn:
Hi @DarhkVoyd! Thanks for showing interest.

This issue requires a couple of steps:

- Remove the `version.txt` files in `/gradio` and `/client/python/gradio_client`
- Update the `tool.hatch.version` section of the `pyproject.toml` in the root of the repo and in `/client/python` to point the path to the `package.json` files instead. The `gradio `package.json is in `gradio/package.json` and the `gradio_client` one is in `client/python/gradio_client/package.json`.
- Update the `pattern` regex in each `pyproject.toml` to match the  version in the `package.json` (let us know if you need some help with this). The version in the package.json is stored like `"version": "1.1.1"` although the number could be anything and it isn't the only text in that file.
- The versioning logic is located [here](https://github.com/gradio-app/gradio/blob/main/pyproject.toml#L44-L46) for `gradio` and [here](https://github.com/gradio-app/gradio/blob/main/client/python/pyproject.toml#L40-L42) fro `gradio_client`. The `package.json` for `gradio` is [here](https://github.com/gradio-app/gradio/blob/main/gradio/package.json) and for `gradio_client` is [here](https://github.com/gradio-app/gradio/blob/main/client/python/gradio_client/package.json).
- Remove the logic that writes the version in `package.json` to `version.txt`. This logic is here: https://github.com/gradio-app/gradio/blob/main/.changeset/fix_changelogs.cjs#L84-L87

[...]

PR #5514, opened 2023-09-13
refactor: Use package.json for version management

DarhkVoyd:
## Description
- This PR updates the project configuration to use the package json file for version management. The change involves configuring reading the ```hatch version``` number from package.json using a regex pattern. 
- Additionally, it removes the logic that creates or updates the version.txt file, which is now redundant.
- Many files used ```pkgutil``` to extract the version, which lead to multiple similar errors after making the changes and deleting the version.txt files. Replaced it by loading version through package.json

Closes: #5489


abidlabs:
Thanks for opening this PR @DarhkVoyd!

There's a few issues that we'll need to fix

1. There's a lot of code duplication across files -- we should consolidate into a single function (ideally located in `gradio_client/utils.py` since that can be imported from `gradio` as well)

2. The `package.json` won't be accessible directly using `open("package.json")` since its located inside the directory where `gradio` is installed. You'll need to do something like: `pkgutil.get_data(__name__, "package.json")`

3. If for some reason, the version cannot be fetched, we should *not* raise an error since this will prevent anyone from being able to use Gradio. Instead it should silently set the value of version to an empty string as in the original logic:

```py
# __version__ = (pkgutil.get_data(__name__, "version.txt") or b"").decode("ascii").strip()
```

DarhkVoyd:
@abidlabs I have made changes and removed the code duplication by creating a function in utils as you asked.

abidlabs:
Looks good thanks @DarhkVoyd! Let me fix the conflict and just confirm everything works locally

gradio/__init__.py
```diff
@@ -104,7 +106,4 @@
 )
 from gradio.themes import Base as Theme
 
-current_pkg_version = (
-    (pkgutil.get_data(__name__, "version.txt") or b"").decode("ascii").strip()
-)
-__version__ = current_pkg_version
```
abidlabs, line 110: We should keep `__version__` as doing `gradio.__version__` is the canonical way to get the version of a package

gradio/__init__.py
```diff
@@ -104,7 +106,4 @@
...
+current_pkg_version = get_package_version()
```
abidlabs, line 109: This gets the gradio_client version even more the main gradio package

abidlabs:
Noticed a couple of small things -- I'll push the fixes so that we can get this merged in.

abidlabs:
There were some other places that we used `version.txt` -- I've updated the other references. This should be good to go, I'll merge later today & do some more testing

abidlabs:
Merged, thanks for the contribution @DarhkVoyd!

Issue #5515, opened 2023-09-13
Custom value in `gr.Dropdown` is not reflected without pressing enter


hysts:
### Describe the bug

Not sure if this is a bug, but it would be nice if it was reflected without pressing the enter key.
It's kind of unexpected behavior that if you didn't press enter, the dropdown menu shows the text you typed but it's actually not used.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

https://huggingface.co/spaces/hysts-debug/dropdown-custom-value
```python
import gradio as gr

with gr.Blocks() as demo:
    with gr.Row():
        with gr.Column():
            dropdown = gr.Dropdown(
                choices=["abc", "def", "ghi"],
                value="abc",
                allow_custom_value=True,
            )
            button = gr.Button()
        with gr.Column():
            out = gr.Textbox()

    button.click(fn=lambda x: x, inputs=dropdown, outputs=out)

if __name__ == "__main__":
    demo.queue().launch()
```

### Screenshot

https://github.com/gradio-app/gradio/assets/25161192/dae1b8b8-280e-46d0-b6e7-ad2ab33a9608

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Linux
gradio version: 3.44.0
gradio_client version: 0.5.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.1.1
fastapi: 0.103.1
ffmpy: 0.3.1
gradio-client==0.5.0 is not installed.
httpx: 0.25.0
huggingface-hub: 0.16.4
importlib-resources: 6.0.1
jinja2: 3.1.2
markupsafe: 2.1.3
matplotlib: 3.7.3
numpy: 1.25.2
orjson: 3.9.7
packaging: 23.1
pandas: 2.0.3
pillow: 10.0.0
pydantic: 1.10.12
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0.1
requests: 2.31.0
semantic-version: 2.10.0
typing-extensions: 4.7.1
uvicorn: 0.23.2
websockets: 11.0.3
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2023.6.0
httpx: 0.25.0
huggingface-hub: 0.16.4
packaging: 23.1
requests: 2.31.0
typing-extensions: 4.7.1
[...]

PR #5525, opened 2023-09-13
Ensure input value saves on dropdown blur

hannahblair:
## Description

We weren't saving the current input value of the dropdown on blur - this PR fixes that 🐛 

Closes: #5515 

## 🎯 PRs Should Target Issues

## Tests


abidlabs:
Thanks @hannahblair -- this change looks good. Could we also make the same change when `multiselect=True`?

If possible, would be great to add unit tests for this to to prevent any regressions. I'll try to increase the coverage of the unit tests for Dropdown/Multiselect in a separate PR.

abidlabs:
Thanks for the quick fix @hannahblair! Taking a look

abidlabs:
I used this code to test:

```py
import gradio as gr

def printy(x):
    print(x)
    return x

with gr.Blocks() as demo:
    gr.Markdown("Inputs")
    d1 = gr.Dropdown(["choice 1", ("choice 2 [with value]", "value 2"), "choice 3"], label="single select", filterable=True,
                    multiselect=False)
    d2 = gr.Dropdown(["choice 1", ("choice 2 [with value]", "value 2"), "choice 3"], label="multi select", filterable=True,
                    multiselect=True, max_choices=2)
    d3 = gr.Dropdown(["choice 1", ("choice 2 [with value]", "value 2"), "choice 3"], label="single select allow_custom_value",
                    multiselect=False, allow_custom_value=True)
    d4 = gr.Dropdown(["choice 1", ("choice 2 [with value]", "value 2"), "choice 3"], label="multi select allow_custom_value",
                    multiselect=True, max_choices=2, allow_custom_value=True)
    
    gr.Markdown("Outputs")
    t = gr.Textbox()
    
    d1.change(printy, d1, t)
    d2.change(printy, d2, t)
    d3.change(printy, d3, t)
    d4.change(printy, d4, t)
    
demo.launch()
```

1. If `allow_custom_value=False` *and* `multiselect=False` (dropdown 1), this change is preventing the user from selecting any choices

3. If `allow_custom_value=True` *and* `multiselect=True` (dropdown 4), the custom value does get saved -- however, it overrides the previously selected choice instead of being _added_ to the list of selected choices. In the 4th dropdown, try selecting a choice first, and then typing in a custom value, and then blurring (versus pressing enter -- the behaviors are different).

js/dropdown/shared/Multiselect.svelte
```diff
@@ -93,6 +93,16 @@
 		if (!allow_custom_value) {
 			input_text = "";
 		}
+
+		if (active_index !== null) {
+			add_or_remove_index(active_index);
+		} else {
```
abidlabs, line 99: I don't think we should add this. If I hover over a choice, and then blur out, I would not expect that choice to be added to the list of selected choices.

abidlabs:
Beautiful @hannahblair! Just one nit above, otherwise looks good to merge.

abidlabs (approved):

abidlabs:
Merging so we can do the release. Thanks @hannahblair!

PR #5526, opened 2023-09-13
Cc fe 2

pngwn:
## Description

Few extra thigns for custom components:

- examples should now work for custom components.
- I added a `@gradio/fallback` package with static, interactive and example entries. We are using the example fall back in the case where a component doesn't have one but I think we can use the interactive and static in place of the `NotTemplateComponent.svelte` @freddyaboulton.
- I tweaked the `gradio_component create` command to _not_ replace the version if the cwd is `js/gradio-preview/test` this allows us to test locally without needing to publish.
- I tweaked the create script to only replace dep versions if they start with `@gradio/` so that we don't try mess around with random 3rd party deps.

## 🎯 PRs Should Target Issues

## Tests


gradio/cli/commands/create.py
```diff
@@ -125,7 +130,9 @@ def _create_backend(name: str, template: str, directory: pathlib.Path):
 from {name.lower()} import {name}
 
 with gr.Blocks() as demo:
-    {name}()
+    {name}(interactive=True)
```
freddyaboulton, line 133: Something is up when cloning the image component FYI

```console
UploadText.svelte:17 Uncaught (in promise) TypeError: ctx[1] is not a function
    at create_fragment (UploadText.svelte:17:7)
    at init (svelte-internal.js:3237:34)
    at new UploadText (UploadText.svelte:19:37)
    at createProxiedComponent (svelte-hooks.js:341:9)
    at new ProxyComponent (proxy.js:242:7)
    at new Proxy<UploadText> (proxy.js:349:11)
    at Array.create_default_slot_1 (InteractiveImage.svelte:37:15)
    at create_slot (svelte-internal.js:165:23)
    at Array.create_if_block_1 (Image.svelte:219:52)
    at Array.create_default_slot (Image.svelte:238:95)
```


```python

import gradio as gr
from newimage import NewImage

with gr.Blocks() as demo:
    inp = NewImage(interactive=True)
    out = NewImage(interactive=False)
    # gr.Examples(["https://gradio-builds.s3.amazonaws.com/diffusion_image/cute_dog.jpg"],
    #             [inp], [out], fn=lambda s: s, cache_examples=True)
    

demo.launch()
```

freddyaboulton (approved):
Examples works really well for textbox @pngwn !

custom_components_examples

Let's merge this in. Noticed an issue with image that we can fix in a separate PR.

gradio/cli/commands/create.py
freddyaboulton: Actually nvm this was my bad

Issue #2236, opened 2022-09-12
Lazy load images with `gr.Gallery` or `gr.Image`


apolinario:
**Is your feature request related to a problem? Please describe.**  
Currently if you use a lot of `gr.Gallery` or `gr.Image` components with the `value` filled in, the app pre-loads all images before actually loading the Gradio demo or application

![image](https://user-images.githubusercontent.com/788417/189625756-ec934c5c-0da6-4cbe-af49-66bdb21f6a4e.png)

**Describe the solution you'd like**  
Add a setting to allow for lazy loading the images when possible, e.g.:
`gr.Gallery(value=["https://a.co/1.jpg","https://a.co/2.jpg"], lazy=True)`. Maybe a way is using `loading="lazy"` to the HTML `<img>` tag or any other lazy loading method that make more sense for Gradio of course

aliabid94:
Yesss, we need to do this for sure. Will take a look!

TashaSkyUp:
add lazy loading to tabs as well?

PR #5528, opened 2023-09-13
Lazy load all images

aliabid94:
Lazy loading of images in Image and Gallery

Fixes: #2236


pngwn (approved):
Looks great, was worried the background image with sketch might be affected but all looks consistent with main! Thanks @aliabid94!

Issue #5519, opened 2023-09-13
gr.Dropdown.update unable to render updated choices


RajarshiBhadra:
### Describe the bug

The dropdown update is returning corrupted values if triggered via change/click/submit

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```!pip install gradio
import gradio as gr

def update_dropdown():
  new_choices = gr.Dropdown.update(choices = ["Hyundai","Suzuki","Mclaren"],label = "Test Dropdown")
  return new_choices

def update_text(text):
  return "You have chosen: " + text

with gr.Blocks() as demo:
  with gr.Row():
    with gr.Column():
      test = gr.Dropdown(choices = ["Honda","BMW","Mercedes"], value = "Honda",label = "Test Dropdown")
      output = gr.Textbox(lines = 5, label="Details")
    with gr.Column():
      test_button = gr.Button(value = "Test",variant = "primary" )
    test.change(update_text,[test],output)
    test_button.click(update_dropdown,[] ,[test])
    
demo.launch(debug = True)
```

### Screenshot

![Gradio Issue](https://github.com/gradio-app/gradio/assets/8817352/a7726525-2a88-4ca6-ab7d-123bfad583a5)

### Logs

```shell
No error
```

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Linux
gradio version: 3.44.0
gradio_client version: 0.5.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 4.2.2
fastapi: 0.103.1
ffmpy: 0.3.1
gradio-client==0.5.0 is not installed.
httpx: 0.25.0
huggingface-hub: 0.17.1
importlib-resources: 6.0.1
jinja2: 3.1.2
markupsafe: 2.1.3
matplotlib: 3.7.1
numpy: 1.23.5
orjson: 3.9.7
packaging: 23.1
pandas: 1.5.3
pillow: 9.4.0
pydantic: 1.10.12
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0.1
requests: 2.31.0
semantic-version: 2.10.0
typing-extensions: 4.5.0
uvicorn: 0.23.2
websockets: 11.0.3
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2023.6.0
httpx: 0.25.0
[...]

PR #5529, opened 2023-09-13
Fix `.update()` method in `gr.Dropdown()` to handle `choices`

abidlabs:
Fixes: #5519 

Test with original repro in issue:

```py
import gradio as gr

def update_dropdown():
  new_choices = gr.Dropdown.update(choices = ["Hyundai","Suzuki","Mclaren"],label = "Test Dropdown")
  return new_choices

def update_text(text):
  return "You have chosen: " + text

with gr.Blocks() as demo:
  with gr.Row():
    with gr.Column():
      test = gr.Dropdown(choices = ["Honda","BMW","Mercedes"], value = "Honda",label = "Test Dropdown")
      output = gr.Textbox(lines = 5, label="Details")
    with gr.Column():
      test_button = gr.Button(value = "Test",variant = "primary" )
    test.change(update_text,[test],output)
    test_button.click(update_dropdown,[] ,[test])
    
demo.launch()
```

### DIFF
--- a/test/test_components.py
+++ b/test/test_components.py
@@ -650,11 +650,20 @@ class TestDropdown:
         """
         Interface, process
         """
-        checkboxes_input = gr.CheckboxGroup(["a", "b", "c"])
-        iface = gr.Interface(lambda x: "|".join(x), checkboxes_input, "textbox")
+        dropdown_input = gr.Dropdown(["a", "b", "c"])
+        iface = gr.Interface(lambda x: "|".join(x), dropdown_input, "textbox")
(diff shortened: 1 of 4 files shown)


test/test_components.py
```diff
@@ -650,11 +650,20 @@ def test_in_interface(self):
         """
         Interface, process
         """
-        checkboxes_input = gr.CheckboxGroup(["a", "b", "c"])
```
aliabid94, line 653: lol whoops

aliabid94 (approved):
tested, works

abidlabs:
Thanks!

PR #5531, opened 2023-09-13
Fix typo in Gallery.stories.svelte

eltociear:
## Description
colums -> columns


abidlabs:
Thank you! We need grammarly inside vscode

PR #5533, opened 2023-09-13
Use Fallback component for the no template case

freddyaboulton:
## Description

- Use Fallback component when a template component is not specified. FYI @pngwn looks like jsonVIEW can't display a basic number/string. Might be good to fix in the frontend but otherwise we can just modify the NoTemplateComponent postprocess.
- Use `pnpm i --ignore-scripts` if installing in test dir as opposed to npm. Otherwise it's tedious to always type it out yourself.

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/gradio/cli/commands/create.py
+++ b/gradio/cli/commands/create.py
@@ -1,6 +1,5 @@
 """Create a custom component template"""
 
-import copy
 import inspect
 import json
 import pathlib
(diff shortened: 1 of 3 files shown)


freddyaboulton:
Gonna merge but opening PR for visibility @pngwn

Issue #4255, opened 2023-05-17
Sliders should look consistent in all kinds of web browsers


Keldos-Li:
- [x] I have searched to see if a similar issue already exists.

## Description
I'm a Mac user (and developer). Sliders in safari look really ugly and annoying... What's worse, the variable `color_accent` in theme builder makes no change to sliders in safari.
<img width="551" alt="image" src="https://github.com/gradio-app/gradio/assets/23137268/477c7fc5-650b-40f7-bdb9-6a9f86bd37e6">

In fact, the slider is such an odd html element that varies from browser to browser:
![](https://user-images.githubusercontent.com/23137268/236451494-cf5ab484-4ffb-4922-826a-91b23e28dacd.png)

**It _should_ and _could_ be improved to make a theme prettier and consistent.**

## Solution
In my attempt a few weeks ago, I made sliders look from this
![](https://user-images.githubusercontent.com/23137268/236452538-e4189eff-7c2d-4cf2-b64b-181f0b646054.png)
to this
![](https://user-images.githubusercontent.com/23137268/236452453-37fbd6cd-845f-487f-8176-3146e264f79e.png)
(Firefox, Chrome and Safari in turn)

You can see my original PR and changes [here](https://github.com/GaiZhenbiao/ChuanhuChatGPT/pull/734). 
(Maybe you can also check the live preview in [this hugging face space](https://huggingface.co/spaces/JohnSmith9982/ChuanhuChatGPT) a few minutes later. This space is outdated for now... I've just asked my teammate to update it).

The main core change in the code is
```css
input[type="range"] {
     -webkit-appearance: none;
     height: 4px;
     background: var(--input-background-fill);
     border-radius: 5px;
     background-image: linear-gradient(var(--primary-500),var(--primary-500));
     background-size: 0% 100%;
     background-repeat: no-repeat;
 }
 input[type="range"]::-webkit-slider-thumb {
     -webkit-appearance: none;
     height: 20px;
     width: 20px;
     border-radius: 50%;
     border: solid 0.5px #ddd;
     background-color: white;
     cursor: ew-resize;
     box-shadow: var(--input-shadow);
     transition: background-color .1s ease;
 }
 input[type="range"]::-webkit-slider-thumb:hover {
     background: var(--neutral-50);
 }
 input[type=range]::-webkit-slider-runnable-track {
     -webkit-appearance: none;
     box-shadow: none;
     border: none;
     background: transparent;
 }
```
```js
[...]

PR #5535, opened 2023-09-13
Makes sliders consistent across all browsers

dawoodkhan82:
## Description
Solution suggested by @Keldos-Li. Also now `--slider-color` from `gr.Themes` should work for all browsers. 

Closes: #4255

## 🎯 PRs Should Target Issues

## Tests


dawoodkhan82:
Just realized i forgot to update the default slider color.

abidlabs:
Nice! We should test with https://www.browserstack.com

Keldos-Li:
image
It will be nicer if you add me as co-author in a commit or something

js/slider/shared/Range.svelte
```diff
@@ -31,6 +31,7 @@
 	}
```
captain-hoity-toity, line 31: There are a few things that could be improved in this code:

1. **Variable Declaration**: All variables should be declared at the top of the script or function to improve readability and maintainability.

```typescript
// Before
var rangeInputs = null;
var numberInputs = null;

// After
let rangeInputs = null;
let numberInputs = null;
```

2. **Type Annotations**: Use TypeScript's type annotations to ensure type safety. For example, the `handle_release` function's parameter `e` should have a type annotation.

```typescript
// Before
function handle_release(e: MouseEvent): void {...}

// After
function handle_release(e: Event): void {...}
```

3. **Event Listeners**: It's a good practice to remove event listeners when they are no longer needed to avoid memory leaks. Consider using Svelte's lifecycle functions to add and remove event listeners.

4. **CSS Best Practices**: Use CSS variables for repeated values and for better maintainability. Also, use a CSS preprocessor like SASS or LESS to make your CSS more readable and easier to maintain.

5. **Accessibility**: Ensure your HTML elements have appropriate ARIA roles and labels for better accessibility. For example, the range input field could have an ARIA role of `slider` and an appropriate label.

6. **Code Duplication**: The code has some repetition that could be avoided. For example, the `setSliderRange` function is called in both the `setSlider` function and the event listeners. Consider refactoring this to avoid repetition.

```typescript
// Before
rangeInputs.forEach((rangeInput) => {
	rangeInput.addEventListener("input", setSliderRange);
});

// After
rangeInputs.forEach((rangeInput) => {
	rangeInput.addEventListener("input", () => setSliderRange(rangeInput));
});
```

7. **Function Naming**: Function names should be descriptive and follow a consistent naming convention. For example, `handle_change` could be renamed to `handleChange`.

8. **Use of `let` and `const`**: Use `let` only when you plan to reassign a variable. Otherwise, use `const`. 

```typescript
// Before
let _id = 0;

// After
const _id = 0;
```

9. **Code Organization**: Organize related code into functions or modules to improve readability and maintainability.

js/slider/shared/Range.svelte
```diff
@@ -41,13 +42,40 @@
 		dispatch("release", value);
 		value = Math.min(Math.max(value, minimum), maximum);
 	}
+
+	var rangeInputs = null;
+	var numberInputs = null;
```
Keldos-Li, line 47: I think we don't need to define global variables now. I wrote this before because at that time gradio-app didn't have event listeners like "render", and we needed to detect if sliders are added to the DOM tree.But now it does, maybe just define & assignment in the internal function would be better.

dawoodkhan82:
> It will be nicer if you add me as co-author in a commit or something

Did so in the latest commit (I think it worked), let me know if it didn't!

abidlabs:
Btw @dawoodkhan82 maybe we can add stories for the `Slider` component as part of this PR?

Keldos-Li:
> Did so in the latest commit (I think it worked), let me know if it didn't!

Emm, I think it didn't work. The correct syntax should also add `< >` before and after the email address. 
```

Co-authored-by: Keldos <hi@keldos.me>
```

But anyway, it's okay not to add me. Really love your work.

dawoodkhan82:
@Keldos-Li Added you! Thanks for your help with this

js/slider/Slider.stories.svelte
```diff
@@ -0,0 +1,48 @@
...
+	name="Slider with step of 10"
+	args={{
+		step: 10
+	}}
+/>
+
+<Story
+	name="Slider in disabled state"
+	args={{
+		mode: "static"
```
abidlabs, line 30: Nit: this one isn't static actually. I think its because you're importing Slider from "./interactive";

js/slider/Slider.stories.svelte
```diff
@@ -0,0 +1,48 @@
...
+/>
+
+<Story
+	name="Slider with min 0 and max 100 and default value of 50"
+	args={{
+		value: 50,
+		minimum: 0,
+		maximum: 100
+	}}
+/>
```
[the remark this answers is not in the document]
dawoodkhan82: Got it, will replace with `./interactive`. Thanks!

abidlabs:
I'll go ahead and merge this in, thanks @dawoodkhan82 @hannahblair

Issue #5540, opened 2023-09-14
`gr.Dropdown` stops working when I type in a custom value and press enter if `allow_custom_value=False`


hysts:
### Describe the bug

`gr.Dropdown` stops working when I type in a custom value and press enter if `allow_custom_value=False`.
When this happens, the dropdown menu will not respond to anything.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

https://huggingface.co/spaces/hysts-debug/dropdown-incorrect-value

### Screenshot

https://github.com/gradio-app/gradio/assets/25161192/75cc9354-b4a5-42f1-9dd6-4fac3bac312c

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Linux
gradio version: 3.44.1
gradio_client version: 0.5.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.1.1
fastapi: 0.103.1
ffmpy: 0.3.1
gradio-client==0.5.0 is not installed.
httpx: 0.25.0
huggingface-hub: 0.16.4
importlib-resources: 6.0.1
jinja2: 3.1.2
markupsafe: 2.1.3
matplotlib: 3.7.3
numpy: 1.25.2
orjson: 3.9.7
packaging: 23.1
pandas: 2.0.3
pillow: 10.0.0
pydantic: 1.10.12
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0.1
requests: 2.31.0
semantic-version: 2.10.0
typing-extensions: 4.7.1
uvicorn: 0.23.2
websockets: 11.0.3
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2023.6.0
httpx: 0.25.0
huggingface-hub: 0.16.4
packaging: 23.1
requests: 2.31.0
typing-extensions: 4.7.1
websockets: 11.0.3
```

### Severity

I can work around it

abidlabs:
Yikes thanks for catching this regression @hysts!

PR #5544, opened 2023-09-14
Fixes dropdown breaking if a user types in invalid value and presses enter

abidlabs:
Fixes: #5540

Repro: 

```py
import gradio as gr

with gr.Blocks() as demo:
    with gr.Row():
        with gr.Column():
            dropdown = gr.Dropdown(
                choices=["abc", "def", "ghi"],
                value="abc",
                allow_custom_value=False,
            )
            button = gr.Button()
        with gr.Column():
            out = gr.Textbox()

    button.click(fn=lambda x: x, inputs=dropdown, outputs=out)

if __name__ == "__main__":
    demo.queue().launch()
```

Notably, the Multiselect was not experiencing this bug. 

Adds a couple of unit tests which don't actually test this exact bug (I couldn't figure out simulating the enter key -- tried keyboard("{Enter}") but it wasn't working for me), but they do test related behavior.

### DIFF
--- a/.changeset/sour-falcons-marry.md
+++ b/.changeset/sour-falcons-marry.md
@@ -1,0 +1,6 @@
+---
+"@gradio/dropdown": patch
+"gradio": patch
+---
+
+fix:Fixes dropdown breaking if a user types in invalid value and presses enter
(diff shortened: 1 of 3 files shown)


hysts (approved):
Thanks for the quick fix! I tested it and it works fine for me.

PR #5550, opened 2023-09-14
Adding basque language

EkhiAzur:
## Adding basque language

Added some basque translations.


abidlabs:
Thanks @EkhiAzur! Interesting that the language code for Basque is `eu`.

Issue #5561, opened 2023-09-14
Add missing main/pip toggle to guide pages


aliabd:
We only support switching from pip to main on the guides gallery page. should support it on the guides themselves

Issue #5558, opened 2023-09-14
Fix routing on website for deprecated paths


aliabd:
Routing breaks on the website when something is deprecated because routing only checks main jsons and not pip, so ends up with pip version of docs looking for a path that you can't route to.

PR #5565, opened 2023-09-14
Route docs and guide urls correctly 

aliabd:
There's a bug with routing on the website that causes the build to fail when a page is deprecated on the docs or a guide is deleted. This fixes routing for both docs and guides. Removed param matching and just handle it on load

Also adds a main/pip toggle to guide pages. 

Closes: #5558 
Closes: #5561


pngwn (approved):
Looks good! Thanks @aliabd!

Issue #5121, opened 2023-08-07
Round numbers in a dataframe


clefourrier:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
On the Open LLM Leaderboard, I'd like to displayed scores rounded to one decimal point, to make it readable by users. However, if I truncate/round the inputs, then in some cases (1.23 vs 1.21, both rounded to 1.2), rows sorted by value will not be sorted correctly. 

**Describe the solution you'd like**  
In pandas dataframe, you can use `pd.set_option('display.precision', 1)` to display/round all numbers to one decimal point. I'd be very happy if a similar system existed in gradio.

**Additional context**  
Related [user issue](https://huggingface.co/spaces/HuggingFaceH4/open_llm_leaderboard/discussions/171)

PR #5569, opened 2023-09-14
Added support for pandas `Styler` object to `gr.DataFrame` (initially just sets the `display_value`)

abidlabs:
This PR allows users to pass in a pandas `Styler` object to `gr.DataFrame`. Right now, the main benefit is that it allows us to set a "display_value" (e.g. the precision of numbers) that is different than the underlying value used to sort the columns.

The "display_value" has other uses as well, as described here: https://pandas.pydata.org/docs/user_guide/style.html

Example usage:

```py
import pandas as pd
import gradio as gr

df = pd.DataFrame({
    "name": ["Adam", "Mike"],
    "gpa": [1.1, 1.12],
    "sat": [800, 800]
})
s = df.style.format(precision=1, decimal=",")


with gr.Blocks() as demo:
    gr.DataFrame(s)
    
demo.launch()
```

In this example, clicking on the "gpa" column will allow you to change the sort order because even though the displayed values are the same, the underlying values are different. On the other hand, clicking on the "sat" column will have no effect since the underlying values are exactly the same. 

Did some other cleanup and added a visual test. This is also setting the stage for adding colors using the same `Styler` object. 

Fixes: #5121


requirements.txt
```diff
@@ -12,7 +12,7 @@ matplotlib~=3.0
 numpy~=1.0
 orjson~=3.0
 packaging
-pandas>=1.0,<3.0
+pandas>=1.3.0,<3.0
```
abidlabs, line 15: The `df._compute()._translate(None, None)` method signature was adopted in `pandas 1.3.0`. Should be fine to constrain the requirements a bit given that pandas 1.3.0 is over 2 years old.

abidlabs:
There's some types that need to be fixed -- I'll do that tomorrow, but otherwise the PR should be ready for review.

js/dataframe/static/StaticDataframe.svelte
```diff
@@ -85,11 +82,8 @@
 		{label}
 		{row_count}
 		{col_count}
-		values={value}
+		{value}
 		{headers}
-		on:change={({ detail }) => {
```
abidlabs, line 90: I had to remove this `on:change` handler -- otherwise I'd get an infinite loop. I didn't see any repercussions but I don't think this is the right thing to do

js/dataframe/shared/Table.svelte
```diff
@@ -152,7 +160,8 @@
 	$: _headers &&
 		dispatch("change", {
 			data: data.map((r) => r.map(({ value }) => value)),
-			headers: _headers.map((h) => h.value)
+			headers: _headers.map((h) => h.value),
+			metadata: editable ? null : { display_value: display_value }
```
abidlabs, line 164: Originally just had:

> metadata: {display_value: display_value }

but I'd get an infinite loop in the InteractiveDataframe. I think I'm doing something wrong here

js/dataframe/static/StaticDataframe.svelte
pngwn: The static dataframe never updates from inside so this won't cause any issues (and I couldn't find any from testing). This is probably a leftover from a refactor or something.

js/dataframe/shared/Table.svelte
pngwn: I'm entirely sure why this would be happening. I know that the dataframe does update more than it should and can be very sensitive to changes triggering new updates but I don't really understand why this would stop the update from happening because this branch trigger an update to the value regardless. I'll have a poke around and see if I can figure out what might be causing it.

I'll probably take a look at the way we handle state in the dataframe shortly and do sonmething more explicit than what we currently have. For now I don't want to block this. I can see any actuall issues, so if my poking and prodding doesn't turn up anything I think it is fine to merge until we refine how we handle state and updates in general for the DF.

abidlabs:
Thanks for reviewing @pngwn! I tested and I don't see any issues so I'll go ahead and get this in (once I fix the backend tests).

Issue #5577, opened 2023-09-15
Problem with Gradio auth in applications on different ports on the same server


vivianerenizia:
### Problem description:

I have Gradio applications on the same server, but on different ports, and I am facing the following problem:

- When I open application A, it works normally.
- Then I open application B, which also works normally.
- However, when I go back to application A, application A breaks and displays an error.
- Furthermore, if I open application C, application B also stops working.

### Have you searched existing issues?  🔎

- [x] I have searched and found no existing issues

### Codes to test:

Here is sample code that demonstrates the problem:

- Code A:

```
import gradio as gr
user_db = {"123": "456"}
def greet(name):
    return "Hello " + name + "!!"
demo = gr.Interface(fn=greet, inputs="text", outputs="text")
demo.launch(auth=lambda u, p: user_db.get(u) == p)
```

- Code B:

```
import gradio as gr
user_db = {"foo": "bar"}
def greet(name):
    return "Hello " + name + "!!"
demo = gr.Interface(fn=greet, inputs="text", outputs="text")
demo.launch(auth=lambda u, p: user_db.get(u) == p)
```

### To reproduce the issue, follow these steps:

1. Run code A.
2. Test application A and keep it open.
3. Run code B.
4. Test application B.
5. Go back to application A and test again. The error will most likely occur in application A.

### Screenshot

1. Run code A.
2. Test application A and keep it open.
![image](https://github.com/gradio-app/gradio/assets/31120743/81a08650-3538-494e-ad04-8a2c1020329c)

3. Run code B.
4. Test application B.
![image](https://github.com/gradio-app/gradio/assets/31120743/a0c5990a-e0b1-4644-981c-c25f97c0c75b)

5. Go back to application A and test again. The error will most likely occur in application A.
![image](https://github.com/gradio-app/gradio/assets/31120743/4b149b22-5d81-47f5-b06f-569b64875421)

### System Info

```shell
- Gradio: 3.44.0
```

### Severity

I can work around it

abidlabs:
So it seems to happen because a cookie is generated every time you log into a gradio app. When you log into the 2nd gradio app, the cookie is deleted for the first gradio app. Will see if we can make the cookie persist

vivianerenizia:
Thank you a lot, @abidlabs!

PR #5588, opened 2023-09-18
Allow multiple instances of Gradio with authentication to run on different ports

abidlabs:
Turns out that on localhost, cookies [are not port-specific](https://stackoverflow.com/questions/1612177/are-http-cookies-port-specific). This meant that if you were running Gradio w/ auth on one port, and then you logged into a second Gradio instance, the second cookie would override the first once -- since all cookies shared the same name. 

Now, we use a unique name per cookie so this does not happen. 

Closes: #5577

### DIFF
--- a/gradio/analytics.py
+++ b/gradio/analytics.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 import asyncio
 import json
 import os
-import pkgutil
 import threading
 import urllib.parse
 import warnings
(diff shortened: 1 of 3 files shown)


gradio/analytics.py
```diff
@@ -4,7 +4,6 @@
 import asyncio
 import json
 import os
-import pkgutil
```
abidlabs, line 7: unrelated but not sure why we're just catching this now!

freddyaboulton: I think its https://github.com/gradio-app/gradio/pull/5514/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711

freddyaboulton (approved):
This looks good to me @abidlabs ! Thanks

abidlabs:
Awesome thanks @freddyaboulton

Issue #5585, opened 2023-09-18
Add id, class, and or attribute to a Tab's button


CurtisDS:
**Is your feature request related to a problem? Please describe.**  
I have Javascript that is added to the page which will activate a tab during specific instances. I do this by scanning the tab buttons in the tab bar and checking the label of each one to know which one to activate, however, an issue has popped up where someone is using a translation extension that swaps out the labels on the buttons and that breaks the functionality I was using to find the tab button.

**Describe the solution you'd like**  
If we were able to add an id, class, or attribute to the tab's button I could then use that to find the correct button instead of a string compare on the label.

Example: 
```python
with gr.Tab(
    my_tab_label,
    elem_id=f"my_tab_id",
    elem_classes="my_tab_class",
    btn_id=f"my_tabs_btn_id",
    btn_classes="my_tabs_btn_class"
):
```

Would result in:

```html
<div class="tabs gradio-tabs" id="tabs">
    <div class="tab-nav">
        <button id="my_tabs_btn_id" class="my_tabs_btn_class">my_tab_label</button>
    </div>
    <div id="my_tab_id" class="tabitem gradio-tabitem my_tab_class">
        <!-- tab content -->
    </div>
</div>
```

abidlabs:
I'm hesitant to add two separate parameters. However, what if we were to automatically set the btn id to be "{elem_id}-btn", allowing you to target it directly?

CurtisDS:
that would be perfect

PR #5590, opened 2023-09-18
Attach `elem_classes` selectors to layout elements, and an id to the Tab button (for targeting via CSS/JS)

abidlabs:
This PR does two things:

1. It passes the `elem_classes` attribute to the Blocks layouts (e.g. Row, Accordion, Tab, etc.) We had these attributes set up correctly in the frontend, but we were never passing them in from the frontend

2. For the `TabItem`, the `elem_id` would get attached to the div holding the tab contents. We now also attach an `{elem_id}-button` to the button that you click on to select the tab. This closes: #5585 

Test code:

```py
import gradio as gr

css = """
#blue-button{
    background-color:aqua
}
#green-button {
    background-color:green
}
#red{
    background-color:red
}
"""

    with gr.Tab(
            "Blue",
            elem_id="blue",
        ):
            gr.Textbox()
    with gr.Tab(
            "Green",
            elem_id="green",
        ):
            with gr.Accordion("Red background", elem_id="red"):
                gr.Textbox()
                gr.Textbox()
        
    demo.launch()
```


js/tabs/static/Tabs.svelte
```diff
@@ -58,11 +59,15 @@
 	<div class="tab-nav scroll-hide">
 		{#each tabs as t, i (t.id)}
 			{#if t.id === $selected_tab}
-				<button class="selected">
+				<button
+					class="selected"
+					{...elem_id ? { id: t.elem_id + "-button" } : {}}
```
pngwn, line 64: In Svelte if you set an attribute to null it gets removed. That might be a slightly cleaner way of expressing this.

abidlabs: Ah nice. Figured there'd a nicer way. Will update

dawoodkhan82:
Nice! just tested and everything looks good. 

Small nit: wondering if we should name the tab button `{elem_id}-tab-button` to be more specific.

dawoodkhan82 (approved):

abidlabs:
Thanks @dawoodkhan82 for reviewing! 

I don't think we need to be more specific, users can select by `tab-nav>{elem-id}-button` if there's a different element with the `{elem-id}-button` ID somewhere on the page.

Issue #5584, opened 2023-09-18
Stuck with `HighlightedText` when using `combine_adjacent` without a `value`


volltin:
### Describe the bug

There is a bug in the `gr.HighlightedText` component when using the `combine_adjacent=True` parameter without providing a `value`. When these parameters are used, the application becomes stuck on "Loading...".

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

Stuck on "Loading"

```python
import gradio as gr

with gr.Blocks() as demo:
    # stuck on "Loading"
    ht = gr.HighlightedText(
        combine_adjacent=True,
    )

demo.launch()
```

While other cases work well:

```python
import gradio as gr

with gr.Blocks() as demo:
    # Case 1: None, OK
    ht1 = gr.HighlightedText(
    )
    # Case 2: Empty string, OK
    ht2 = gr.HighlightedText(
        value="",
    )
    # Case 3: Empty string and combine_adjacent, OK
    ht3 = gr.HighlightedText(
        value="",
        combine_adjacent=True,
    )

demo.launch()
```

### Logs

```shell
No logs from python.

Logs from the browser console:


Uncaught (in promise) TypeError: s is not iterable
    at C (utils.ts:53:27)
    at Bl.e.$$.update (InteractiveHighlightedText.svelte:37:11)
    at kt (Component.js:132:5)
    at new Hl (InteractiveHighlightedText.svelte:37:40)
    at Fc (dom.js:1197:9)
    at Ql (Render.svelte:72:37)
    at kt (Component.js:136:34)
    at new Gt (Render.svelte:63:25)
    at Pt (Render.svelte:83:9)
    at Tt (Render.svelte:83:9)
```
```


### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Darwin
gradio version: 3.44.3
gradio_client version: 0.5.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.1.1
fastapi: 0.103.1
ffmpy: 0.3.1
gradio-client==0.5.0 is not installed.
httpx: 0.25.0
huggingface-hub: 0.17.1
importlib-resources: 6.0.1
jinja2: 3.1.2
markupsafe: 2.1.3
matplotlib: 3.8.0
numpy: 1.25.2
orjson: 3.9.7
packaging: 23.1
pandas: 2.1.0
pillow: 10.0.1
pydantic: 2.3.0
[...]

PR #5602, opened 2023-09-19
Ensure `HighlightedText` with `merge_elements` loads without a value 

hannahblair:
## Description

`merge_elements()` was trying to run on initial load even when `value` was empty, resulting in an error on load. I've added a flag to check that `value` isn't empty before we run that func. 

demo
```
import gradio as gr

with gr.Blocks() as demo:
    ht = gr.HighlightedText(
        combine_adjacent=True,
    )

demo.launch()
```

Closes: #5584 

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/chubby-hounds-itch.md
+++ b/.changeset/chubby-hounds-itch.md
@@ -3,4 +3,4 @@
 "gradio": patch
 ---
 
-fix:Ensure HighlightedText with `merge_elements` loads without a value 
+fix:Ensure `HighlightedText` with `merge_elements` loads without a value 


js/highlightedtext/utils.ts
```diff
@@ -45,29 +45,32 @@ export function correct_color_map(
 export function merge_elements(
 	value: HighlightValueType[],
 	mergeMode: "empty" | "equal"
-): HighlightValueType[] {
-	let result: HighlightValueType[] = [];
-	let tempStr: string | null = null;
-	let tempVal: string | number | null = null;
+): HighlightValueType[] | null {
+	if (value) {
```
pngwn, line 49: Could we flip this around to do an early return  (`if (!value) return null`)?

pngwn (approved):
Looks great! Thanks for taking care of this @hannahblair !

Issue #5513, opened 2023-09-13
Copy-Paste Formatting Issues with Chatbot Text


huni0103:
### Describe the bug

Hi there.

I would like to make an additional suggestion regarding Chatbot Copy.
In the Chatbot, I am replacing "<" , ">" with ">", "<", etc. to display "<" , ">".
However, when I copy the Chatbot text, the "" is also being copied.
I only want "<" and ">" to be copied as they appear in the Chatbot.

Furthermore, I'm representing newlines as "
", but when I copy the Chatbot text, it's copying them as "
" even though they are displayed as "\n" in the Chatbot.

Is there any way to address the issues mentioned above?

We might not be able to use the Copy & Paste feature in the Chatbot until the issues mentioned above are resolved. 😔

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

The result of clicking the Copy button for the screenshot below.
![image](https://github.com/gradio-app/gradio/assets/120710184/2518b841-af19-4edf-b613-3a6ba607876b)

### Screenshot

![image](https://github.com/gradio-app/gradio/assets/120710184/fb34cfa4-30a1-4c6c-b7a2-2868b2af5380)

### System Info

```shell
gradio == 3.42.0
```

### Severity

I can work around it

abidlabs:
Hi @huni0103 just to make sure we understand the issue, can you provide a code snippet we can use to reproduce? It seems that some information was lost in the GitHub markdown processing? You an provide a code snippet like this:

```py
with gr.Blocks() as demo:
   gr.Chatbot(...)
```

huni0103:
Hi @abidlabs.
I apologize in advance for my shortcomings English proficiency.

I would like to express my gratitude for your quick response to the issue. Due to internal security concerns, we cannot share our entire codebase. However, we can provide you with reproducible code and the tested text values.

You've added backslashes(\) before the "<" and ">" symbols in the input text to ensure they are displayed correctly in the ChatBot. Without the backslashes, they seem to be interpreted as HTML tags and do not display correctly in the ChatBot.

Additionally, to ensure proper line breaks in the ChatBot output, you've replaced newline characters (\n) with <br/>.

I would like the Copy button to produce output in the same way it appears to the user in the ChatBot. Specifically, I want "<br/>" to be copied as line breaks, and I prefer not to include the backslash before "<" and ">" symbols. 
It would be preferable for users to be able to copy the text exactly as it appears in the ChatBot.

[...]

PR #5604, opened 2023-09-19
Add `render_markdown` parameter to chatbot

dawoodkhan82:
## Description
Allow users to `disable_markdown` for chatbot messages.

Closes: #5513

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/gradio/components/chatbot.py
+++ b/gradio/components/chatbot.py
@@ -53,12 +53,12 @@ class Chatbot(Changeable, Selectable, Likeable, IOComponent, JSONSerializable):
         elem_classes: list[str] | str | None = None,
         height: int | None = None,
         latex_delimiters: list[dict[str, str | bool]] | None = None,
-        disable_markdown: bool = False,
         rtl: bool = False,
         show_share_button: bool | None = None,
         show_copy_button: bool = False,
         avatar_images: tuple[str | Path | None, str | Path | None] | None = None,
         sanitize_html: bool = True,
+        render_markdown: bool = True,
         bubble_full_width: bool = True,
         **kwargs,
     ):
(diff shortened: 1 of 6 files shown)


gradio/components/chatbot.py
```diff
@@ -76,6 +77,7 @@ def __init__(
             elem_classes: An optional list of strings that are assigned as the classes of this component in the HTML DOM. Can be used for targeting CSS styles.
             height: height of the component in pixels.
             latex_delimiters: A list of dicts of the form {"left": open delimiter (str), "right": close delimiter (str), "display": whether to display in newline (bool)} that will be used to render LaTeX expressions. If not provided, `latex_delimiters` is set to `[{ "left": "$$", "right": "$$", "display": True }]`, so only expressions enclosed in $$ delimiters will be rendered as LaTeX, and in a new line. Pass in an empty list to disable LaTeX rendering. For more information, see the [KaTeX documentation](https://katex.org/docs/autorender.html).
+            disable_markdown: If True, will disable Markdown rendering for chatbot messages.
```
abidlabs, line 80: tiniest nit: you could consider calling it `render_markdown` and setting it to True by default to be more parallel with `sanitize_html`

js/markdown/static/MarkdownCode.svelte
```diff
@@ -24,10 +25,19 @@
...
+
+	function process_message(value: string): string {
+		if (sanitize_html && !disable_markdown) {
+			return DOMPurify.sanitize(marked.parse(value));
+		} else if (sanitize_html && disable_markdown) {
+			return DOMPurify.sanitize(value);
+		} else if (!sanitize_html && !disable_markdown) {
+			return marked.parse(value);
+		}
+		return value;
```
abidlabs, line 37: You could simplify this by doing something like:
```
if (!disable_markdown): {
   value = marked.parse(value)
}
if (sanitizer_html) {
   value = DOMPurify.sanitize(value)
}
return value;
```

abidlabs:
@dawoodkhan82 I think there's still one change that needs to be made -- if Markdown is disabled, then we should be rendering the string the user as a regular string, instead of as html.

I tested this example:

```py
import gradio as gr

long_msg = """
Hi **Test**

one1 <Color2>

<Color3> two2
<Color4> three3
"""

with gr.Blocks() as demo:
    gr.Chatbot(
        [[long_msg, long_msg]],
        height=550,
        disable_markdown=True,
        sanitize_html=False,
    )
demo.launch()
```

You'll notice that the `<Color>` tags don't appear:

image

I believe that's because we're still rendering the content as html instead of as a regular string:

https://github.com/gradio-app/gradio/blob/1bc8471abc31dd4a7323b7275971dbcaf4f8fdd2/js/markdown/static/MarkdownCode.svelte#L56

dawoodkhan82:
Thanks for the review, and catching the html bug. I made all the changes

abidlabs:
Looks good @dawoodkhan82 -- one last thing I noticed is that if `render_markdown` is False, then new lines are not preserved. So the message ends up looking like this:

image

We should be able to fix this by adding `white-space: pre-line;` to the style if `render_markdown` is False.

dawoodkhan82:
@abidlabs fixed the new line issue

abidlabs:
LGTM thanks for addressing everything @dawoodkhan82! I'll merge this in to reduce the number of open PRs :)

PR #5605, opened 2023-09-19
Add gr.Request in ThreadData context

cbensimon:
## PR content
- Allows accessing the `gr.Request` object with `[gradio].context.thread_data.request`
- Type the `threading.local()` class (inspired from https://github.com/python/mypy/issues/2388#issuecomment-257968459)

This can be used as an alternative to the `gr.Request` dependency injection

### DIFF
--- a/.changeset/whole-rabbits-walk.md
+++ b/.changeset/whole-rabbits-walk.md
@@ -1,0 +1,5 @@
+---
+"gradio": minor
+---
+
+feat:Add gr.Request in ThreadData context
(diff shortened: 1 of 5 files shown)


abidlabs:
Thanks @cbensimon! Would you be able to review @aliabid94?

aliabid94:
Thanks @cbensimon! Can you explain the reasoning for this? Right now, we don't expose gr.context.thread_data as a public API, it's used for internal logic only. What is the benefit of supporting this way of passing the request in addition to the standard function argument method?

cbensimon:
Closing for https://github.com/gradio-app/gradio/pull/5625

Issue #5592, opened 2023-09-18
regression in showing pandas dataframes -- chopped and force wrapped


pseudotensor:
### Describe the bug

Regression in showing pandas dataframes -- chopped and force wrapped

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```
import gradio as gr

demo = gr.Blocks()

with demo:
    import pandas as pd
    df0 = pd.read_csv('df0.csv')
    langchain_mode_path_text = gr.Dataframe(value=df0,
                                            visible=True,
                                            label='LangChain Mode-Path',
                                            show_label=False,
                                            interactive=False)
demo.queue()
demo.launch()
```

[df0.csv](https://github.com/gradio-app/gradio/files/12654055/df0.csv)

Only seen after going from gradio==3.41.0 to gradio==3.44.3

Also, there seems to be some unnecessary animation of the dataframe, showing it shrink into place, while before it just shows up nicely at once.

### Screenshot

gradio==3.44.3

![image](https://github.com/gradio-app/gradio/assets/2249614/e5458a4f-346c-4ee7-9e65-a4908b58bc3b)

gradio==3.41.0:

![image](https://github.com/gradio-app/gradio/assets/2249614/7a0ce523-6376-4d51-8e06-9dcd16483e8c)

### System Info

```shell
gradio==3.44.3
gradio_client==0.5.0
```

### Severity

Not quite Blocking usage of gradio, just can't upgrade to this version until fixed since too noticeable visual bug.

abidlabs:
Thanks @pseudotensor -- yes we're looking into this here: https://github.com/gradio-app/gradio/issues/5458#issuecomment-1712313516, but let's keep this issue open for visibility. We'll get this fixed soon!

PR #5616, opened 2023-09-19
Fix width and height issues that would cut off content in `gr.DataFrame`

abidlabs:
This is an attempt to fix: https://github.com/gradio-app/gradio/issues/5592. Previously, if you would run a demo like `matrix_transpose` or code like this:

```py
import gradio as gr

df = [[f"col: {i} -- row:{j}" for j in range(1, 3)] for i in range(1, 5)]

with gr.Blocks() as demo:
    gr.DataFrame(df, headers=list(range(1, 11)))
    gr.DataFrame(df, headers=list(range(1, 11)), interactive=True)

if __name__ == "__main__":
    demo.launch()
```

you'd end up with a compressed dataframe whose content was cut off and scrollbars would always be present:

image

This made small dataframes frustrating to use. I think the issue was that the scrollbar width/height was not taken into account. Now, both are taken into account. The scrollbar holders are always present, but the scroll bars themselves only appear when needed:

image

Would appreciate a review @pngwn to ensure that this doesn't break anything else and this is the correct approach.

### DIFF
--- a/js/dataframe/shared/Table.svelte
+++ b/js/dataframe/shared/Table.svelte
@@ -529,7 +529,7 @@
 
 	function set_cell_widths(): void {
 		const widths = cells.map((el, i) => {
-			return el?.clientWidth || 0;
+			return el?.offsetHeight || 0;
 		});
 
 		if (widths.length === 0) return;
--- a/js/dataframe/shared/Table.svelte
+++ b/js/dataframe/shared/Table.svelte
@@ -28,7 +28,7 @@
 
 	export let editable = true;
 	export let wrap = false;
-	export let height: number | undefined;
+	export let height: number;
 	let selected: false | [number, number] = false;
 
 	$: {
@@ -541,7 +541,7 @@
 		}
 	}
 
-	let table_height: number = height || 500;
+	let table_height: number = height;
 	let scrollbar_width = 0;
 
 	function sort_data(
@@ -683,7 +683,7 @@
 			<VirtualTable
 				bind:items={data}
 				table_width={t_width}
-				max_height={height || 500}
+				max_height={height}
 				bind:actual_height={table_height}
 				bind:table_scrollbar_width={scrollbar_width}
 				selected={selected_index}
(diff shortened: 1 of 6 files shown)


js/dataframe/shared/Table.svelte
```diff
@@ -529,7 +529,7 @@
 
 	function set_cell_widths(): void {
 		const widths = cells.map((el, i) => {
-			return el?.clientWidth || 0;
+			return el?.offsetHeight || 0;
```
pngwn, line 532: Why are we measuring the height to get the width of the cell?

js/dataframe/shared/VirtualTable.svelte
```diff
@@ -56,14 +56,18 @@
 		end = i;
 		const remaining = _items.length - end;
 
+		const scrollbarHeight = viewport.offsetHeight - viewport.clientHeight;
+		if (scrollbarHeight > 0) {
+			content_height += scrollbarHeight;
+		}
+
```
pngwn, line 63: The only issue with this is that a scrollbars are often dynamic, hiding and showing in response to interaction.

js/dataframe/shared/Table.svelte
abidlabs: Whoops my bad, will fix

js/dataframe/shared/VirtualTable.svelte
abidlabs: From my tests, this didn’t appear to be an issue. Does this not work for you?

pngwn: I've only reviewed code so far, will test properly in the morning.

abidlabs: sg thanks

abidlabs:
I also updated the python docstrings. To avoid confusion, I set the default `height` to `500` directly in the Python class constructor.

pngwn (approved):
Tested this now and the code looks good, nice solution! Thanks for taking care of this @abidlabs!

abidlabs:
Thanks @pngwn for looking it over, much appreciated.

Issue #5632, opened 2023-09-20
Radio button broken when using gr.load() to load Huggingface space


utatane-kio:
### Describe the bug

I have a gradio app hosted on a private HuggingFace space that has a radio button.
Depending on the choice of the radio button, a `gr.Textbox` will switch between `interactive=True` and `interactive=False` .
When I try to use `gr.load()` to load that space in a colab notebook or a different HuggingFace space, the radio button does not function properly and does not seem to be passing the correct values to the update function.
Works just fine on the original space.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

Source gradio app hosted on huggingface space
```
import gradio as gr
def change_textbox(choice):
    if choice in {"A", "B"}:
        return gr.update(interactive=False, value="not interactive")
    elif choice == "C":
        return gr.update(interactive=True)
  
with gr.Blocks() as demo:
    radio_button = gr.Radio(["A", "B", "C"], type="value")
    text_box = gr.Textbox()
    radio_button.change(fn=change_textbox, inputs=radio_button, outputs=text_box)
 
demo.queue().launch()
```
code to load from huggingface space
```
import gradio as gr
demo = gr.load(name="organization_name/space_name", src="spaces")
demo.queue(api_open=False)
demo.launch(debug=True)
```

### Screenshot

<img width="1122" alt="スクリーンショット 2023-09-20 23 39 10" src="https://github.com/gradio-app/gradio/assets/115803278/23624b26-8d0c-440e-8e12-5810cc1d3182">

### Logs

```shell
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/gradio/queueing.py", line 388, in call_prediction
    output = await route_utils.call_process_api(
  File "/usr/local/lib/python3.10/dist-packages/gradio/route_utils.py", line 219, in call_process_api
    output = await app.get_blocks().process_api(
  File "/usr/local/lib/python3.10/dist-packages/gradio/blocks.py", line 1437, in process_api
    result = await self.call_function(
  File "/usr/local/lib/python3.10/dist-packages/gradio/blocks.py", line 1109, in call_function
    prediction = await anyio.to_thread.run_sync(
  File "/usr/local/lib/python3.10/dist-packages/anyio/to_thread.py", line 33, in run_sync
    return await get_asynclib().run_sync_in_worker_thread(
[...]

PR #5633, opened 2023-09-20
Allow Gradio apps containing `gr.Radio()`, `gr.Checkboxgroup()`, or `gr.Dropdown()` to be loaded with `gr.load()`

abidlabs:
We allow these components to take in tuples of choices, but we when tuples are converted to javascript, they are converted to lists, so we need to handle that as well.

Test with:

```py
import gradio as gr
demo = gr.load(name="abidlabs/test-radio", src="spaces")
demo.queue(api_open=False)
demo.launch()
```

Fixes: #5632


freddyaboulton (approved):
Nice! Looks good to me @abidlabs

abidlabs:
Thanks @freddyaboulton!

Issue #5634, opened 2023-09-20
Cannot display outputs when reusing cached examples


aaronng91:
### Describe the bug

When using Examples and setting cache_examples=True, the inputs/outputs load fine when you start app for the first time. However, if you reload the app which reads from the same cache, the outputs no longer load. No examples or methods have changed.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```
import gradio as gr

def combine(a, b):
    return a + " " + b

with gr.Blocks() as demo:
    txt = gr.Textbox(label="Input", lines=2)
    txt_2 = gr.Textbox(label="Input 2")
    txt_3 = gr.Textbox(value="", label="Output")
    btn = gr.Button(value="Submit")
    btn.click(combine, inputs=[txt, txt_2], outputs=[txt_3])
    
    gr.Markdown("## Text Examples")
    gr.Examples(
        [["hi", "Adam"], ["hello", "Eve"]],
        [txt, txt_2],
        txt_3,
        combine,
        cache_examples=True,
    )

if __name__ == "__main__":
    demo.launch()
```

### Screenshot

Output loads fine when initializing cache for the first time
![Screenshot 2023-09-20 at 17 18 18](https://github.com/gradio-app/gradio/assets/8050514/91414b6e-f97d-41c5-be8a-14774bd2dcb3)

Output no longer loads when reusing the same cache
![Screenshot 2023-09-20 at 17 17 57](https://github.com/gradio-app/gradio/assets/8050514/3c337792-a2ed-4262-951e-30e5c97dd36f)

### Logs

Logs when initializing cache for the first time
```
Caching examples at: '/home/aaronn/git/test/gradio_cached_examples/6'
Caching example 1/2
Caching example 2/2
Caching complete

Running on local URL:  http://0.0.0.0:7860

To create a public link, set `share=True` in `launch()`.
```

Logs when reusing cache
```
Using cache from '/home/aaronn/git/test/gradio_cached_examples/6' directory. If method or examples have changed since last caching, delete this folder to clear cache.

Running on local URL:  http://0.0.0.0:7860

To create a public link, set `share=True` in `launch()`.
```

### System Info

```shell
gradio==3.41.2
```

### Severity

Blocking usage of gradio

freddyaboulton:
Thanks @aaronng91 - I can reproduce.

Issue #5521, opened 2023-09-13
Caching with AnnotatedImage as output does not work on app reload


Blubbaa:
### Describe the bug

I am using the `gr.AnnotatedImage` component as an output in a simple interface. When I enable caching I noticed that the cached examples are only loaded and displayed correctly in the first startup (that also creates the cache). Any subsequent app starts will not display any output but also no error message. I have condensed my code to a minimal working example below (using some fake annotation from the docs).

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import random

import gradio as gr
import numpy as np

MAX_WIDTH = 640
MAX_HEIGHT = 480

SECTION_LABELS = [
    "apple",
    "banana",
    "carrot",
    "donut",
    "eggplant",
    "fish",
    "grapes",
    "hamburger",
    "ice cream",
    "juice",
]


def section(img, num_boxes, num_segments):
    sections = []
    for a in range(num_boxes):
        x = random.randint(0, img.shape[1])
        y = random.randint(0, img.shape[0])
        w = random.randint(0, img.shape[1] - x)
        h = random.randint(0, img.shape[0] - y)
        sections.append(((x, y, x + w, y + h), SECTION_LABELS[a]))
    for b in range(num_segments):
        x = random.randint(0, img.shape[1])
        y = random.randint(0, img.shape[0])
        r = random.randint(0, min(x, y, img.shape[1] - x, img.shape[0] - y))
        mask = np.zeros(img.shape[:2])
        for i in range(img.shape[0]):
            for j in range(img.shape[1]):
                dist_square = (i - y) ** 2 + (j - x) ** 2
                if dist_square < r**2:
                    mask[i, j] = round((r**2 - dist_square) / r**2 * 4) / 4
        sections.append((mask, SECTION_LABELS[b + num_boxes]))
    return (img, sections)


# Create interface function
def create_annotation(input_image, other_args=1):
    # Generate random annotations
    output = section(input_image, num_boxes=3, num_segments=2)

    # return output tuple
    return output


if __name__ == "__main__":
    # some example images
    examples = [
        "image/0000155DA.jpg",
        "image/0000155DF.jpg",
    ]

    # Create a simple Gradio interface
    simple_demo = gr.Interface(
        fn=create_annotation,
        inputs=gr.Image(width=MAX_WIDTH, height=MAX_HEIGHT),
[...]

PR #5636, opened 2023-09-20
Fix bug in example cache loading event

freddyaboulton:
## Description

Closes: #5634

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/test/test_helpers.py
+++ b/test/test_helpers.py
@@ -132,6 +132,39 @@ class TestProcessExamples:
         prediction = await io.examples_handler.load_from_cache(1)
         assert prediction[0] == "Hello Dunya"
 
+    def test_example_caching_relaunch(self, connect):
+        def combine(a, b):
+            return a + " " + b
+
+        with gr.Blocks() as demo:
+            txt = gr.Textbox(label="Input")
+            txt_2 = gr.Textbox(label="Input 2")
+            txt_3 = gr.Textbox(value="", label="Output")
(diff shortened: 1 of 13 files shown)


test/test_helpers.py
```diff
@@ -118,6 +118,43 @@ def test_some_headers(self):
...
+        assert client.predict(1, api_name="/examples") == (
+            "hello",
+            "Eve",
+            "hello Eve",
+        )
+
+    # Let the server shut down
+    time.sleep(1)
+
+    with connect(demo, server_port=7859) as client:
```
abidlabs, line 150: Just to confirm, you launch on the same port to ensure that the server truly has shutdown? I don't think you need to do this.

Even if `demo.launch()` were to launch a FastAPI app on a new port, it would still read from the same cache since the ids of the components hasn't changed.

So you can remove the hardcoding of the server ports (which will be better in case this port is occupied)

abidlabs:
Bad whitespace! Nice fix @freddyaboulton 

This actually also fixes: https://github.com/gradio-app/gradio/issues/5521

Left a nit about the tests. Otherwise lgtm!

abidlabs (approved):

test/test_helpers.py
freddyaboulton: The time.sleep is to ensure shutdown. Hardcoding the same port is for paranoia 😂 Let me fix.

Issue #5072, opened 2023-08-02
Add .then, .success to component docs


freddyaboulton:
- [ ] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
The `.then`,  and `.success` events are only listed in the guides but not in the docs. We should list them in the docs.

**Describe the solution you'd like**  
I think we could just add them to the `@document` decorator so that they show up along all other the events for the components.

**Additional context**

freddyaboulton:
If you go to any event on the docs page, there will be a table with all the supported events.

https://www.gradio.app/docs/audio

However, `then` and `success` are not listed even though there are valid events for all components. Not sure the right technical approach but maybe @aliabd can explain.

sabir-sayyed:
This table..? if yes please provide the link where i can get more info on .then and .success to add that into the table

<img width="613" alt="image" src="https://github.com/gradio-app/gradio/assets/137613597/f721a4c4-8da7-4b0d-851c-73ecec2bc1da">

abidlabs:
It would be good to include the forthcoming `.on()` in the docs as well. Quoting from #5639:

> By the way, it would be good to include .on() in the docs, perhaps along with .then() and .success() (we can tackle this in https://github.com/gradio-app/gradio/issues/5072). In particular, the second usage example where the trigger is the change event of the input components could be not obvious (though I do think its handy).

PR #5639, opened 2023-09-20
Add `gr.on` listener method

aliabid94:
Created `gr.on` event listener to bind multiple triggers to a single event. This will
1. reduce code repetition when multiple events trigger the same event
2. allow these triggers to behave as a single event, particularly in that they:
  - create a single API endpoint
  - block each other
 
Two main places I used to see this come up:
1. With a form, having textbox.submit and button.click trigger the same workflow:
```python
with gr.Blocks() as demo:

    def greet(name):
        return "Hello " + name + "!"

    gr.on(
        triggers=[name.submit, greet_btn.click],
        fn=greet,
        inputs=name,
        outputs=output,
    )
```

2.  Simulating live events without having to bind .change to every input
```python
with gr.Blocks() as demo:
    with gr.Row():
        num1 = gr.Slider(1, 10)
        num2 = gr.Slider(1, 10)
        num3 = gr.Slider(1, 10)
    output = gr.Number(label="Sum")

    @gr.on(inputs=[num1, num2, num3], outputs=output)
    def sum(a, b, c):
        return a + b + c
```

`.then` and decorator syntax supported just like regular event listeners. Note in the second example, when triggers are not specified, the inputs are automatically used as live inputs by binding to their `.change`.


demo/hello_blocks/run.py
abidlabs, line 1: Let's not change this demo, its used in the quickstart. Can we find another demo to showcase `gr.on`?

gradio/components/interpretation.py
```diff
@@ -41,7 +41,7 @@ def __init__(
         Component.__init__(
             self, visible=visible, elem_id=elem_id, elem_classes=elem_classes, **kwargs
         )
-        self.component = component
```
abidlabs, line 44: Why did you make this change? It broke interpretation, test e.g. with: `demo/gender_sentence_default_interpretation/run.py`

abidlabs (approved):
Very nice PR @aliabid94!

I noticed a couple of small issues above -- in addition, it would be great to improve the changeset to include a highlight.

Other than that -- should be good to go!

abidlabs:
By the way, it would be good to include `.on()` in the docs, perhaps along with `.then()` and `.success()` (we can tackle this in #5072). In particular, the second usage example where the trigger is the change event of the input components could be not obvious (though I do think its handy).

gradio/chat_interface.py
```diff
@@ -245,8 +245,14 @@ def __init__(
...
     def _setup_events(self) -> None:
         submit_fn = self._stream_fn if self.is_generator else self._submit_fn
+        submit_triggers = (
+            [self.textbox.submit, self.submit_btn.click]
+            if self.submit_btn
+            else [self.textbox.submit]
+        )
         submit_event = (
-            self.textbox.submit(
+            on(
```
freddyaboulton, line 254: I think we're triggering an event twice here. 

Run `chat_interface_streaming_echo` on main and on 3.44.2. Clear the network panel and then click the submit button.

Dependency with id 8 is being run twice.

### This branch
image


### 3.44.2


image


In the case of the chatinterface this is not a noticeable difference but depending on the root cause it may cause expensive events to be re-triggered.

freddyaboulton (commented):
Looks good @aliabid94! The demo code looks a lot cleaner.

Have a question about making `on` be opt-in (remove from Interface) as well as some events maybe being triggered twice.

gradio/interface.py
```diff
@@ -654,65 +651,61 @@ def attach_submit_events(self, submit_btn: Button | None, stop_btn: Button | Non
...
-                            self.output_components,
-                            api_name=self.api_name if i == 0 else None,
-                            scroll_to_output=True,
-                            preprocess=not (self.api_mode),
-                            postprocess=not (self.api_mode),
-                            batch=self.batch,
-                            max_batch_size=self.max_batch_size,
-                        )
-                    )
+                on(
```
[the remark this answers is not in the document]
abidlabs: But the original behavior is a bug right? We always block identical events from triggering each other in 3.x (we plan to change this a bit in 4.0 to ensure that the final event is sent through) -- so this would actually be making it more consistent with the way events are handled.

gradio/components/interpretation.py
aliabid94: oh hmm well interpretation is broken in main, try titanic_survival demo. I made this fix for titanic_survival but I guess I broke gender_sentence_default_interpretation instead. Will revert this change, and interpretation can be properly fixed in another PR.

gradio/interface.py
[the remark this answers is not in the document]
aliabid94: yeah it's actually a bug that you can trigger another submit through a different trigger. I see this as a bugfix.

gradio/chat_interface.py
aliabid94: good catch! was an issue specific to chat interface implementation, fixed.

aliabid94:
Think I addressed everything, should be good to merge

abidlabs:
Very cool @aliabid94! Merging

PR #5642, opened 2023-09-21
Improve plot rendering

aliabid94:
The graphs rendered poorly before - they did not fit in the parent block and were not responsive. Was bothering me so made a quick fix.

Before:
old

After:
new


js/plot/static/Plot.svelte
```diff
@@ -192,7 +192,7 @@
 	<div data-testid={"bokeh"} id={divId} class="gradio-bokeh" />
 {:else if type == "altair"}
 	<div data-testid={"altair"} class="altair layout">
-		<Vega {spec} />
+		<Vega {spec} options={{ actions: false }} />
```
freddyaboulton, line 195: If we're going to remove this, then I suggest we add a param/prop for it because some users may be relying on the action button being present.

js/plot/static/StaticPlot.svelte
```diff
@@ -37,6 +37,7 @@
 	{container}
 	{scale}
 	{min_width}
+	allow_overflow={false}
```
freddyaboulton, line 40: I get the same behavior without this line?

freddyaboulton (commented):
Awesome fix @aliabid94 ! Don't think we should remove the action button without a way to preserve it but other than that this is great.

js/plot/static/StaticPlot.svelte
aliabid94: See bokeh_plot demo, the plot leaks out of the box

js/plot/static/Plot.svelte
aliabid94: Added `show_actions_button` kwarg to configure this.

freddyaboulton (approved):
Thanks @aliabid94 !

freddyaboulton:
Gonna merge in to clear up the PR backlog. Thanks again @aliabid94 !

Issue #5485, opened 2023-09-11
Make it easier to get URL query parameters in a Gradio app


abidlabs:
Right now, the way to get URL query parameters from within a Gradio app is quite hacky -- e.g. see https://huggingface.co/spaces/radames/gradio-url-params?test=%22dsds%22&value=23231

It would be nice if we could get the URL query parameters from the `gr.Request` object

DarhkVoyd:
I'm eager to help with this, so please let me know how I can get started. Thanks!

DarhkVoyd:
@abidlabs can I help with this?

abidlabs:
Hi @DarhkVoyd yes definitely -- let's get your other PR over the finish line first and then I can give some pointers.

DarhkVoyd:
@abidlabs can I help with this now?

abidlabs:
Yes, a good place to start would be to look at the implementation of the `gr.Request` class and see what would be needed to add the URL query parameters to it. Let me know if you have specific questions!

DarhkVoyd:
@abidlabs I have created a PR, kindly review.

abidlabs:
The way I'd suggest implementing this:

1. Currently we make requests from the frontend to the backend in these two places:

https://github.com/gradio-app/gradio/blob/26fef8c7f85a006c7e25cdbed1792df19c512d02/client/js/src/client.ts#L442

(if queuing is disabled)

and 

https://github.com/gradio-app/gradio/blob/26fef8c7f85a006c7e25cdbed1792df19c512d02/client/js/src/client.ts#L510

(if queuing is enabled)

- As you can see, no query parameters are included in these requests. We should copy the query parameters from the page's URL and into both of these requests

Then, the `gr.Request` object (which is just a wrapper over the FastAPI Request object), should automatically contain the `query_params`

DarhkVoyd:
@abidlabs hey, I have made necessary changes and it works locally. Should I create a PR?

abidlabs:
Yes please @DarhkVoyd!

PR #5647, opened 2023-09-21
get URL query parameters from the gr.Request object

DarhkVoyd:
## Description

- Currently, the way to extract URL query parameters from within a Gradio app is using _js: `URLSeachParams(window.location.search)`.
- This PR adds the `query_params` property to the `gr.Request` object. This makes it easier to get URL query params without the need for custom js function.

Closes: #5485

### DIFF
--- a/.changeset/fuzzy-terms-trade.md
+++ b/.changeset/fuzzy-terms-trade.md
@@ -1,0 +1,5 @@
+---
+"gradio": minor
+---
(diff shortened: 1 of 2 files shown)


abidlabs:
Hi @DarhkVoyd this isn't the correct approach. I'll close this PR and suggest an approach in the original issue

Issue #5518, opened 2023-09-13
Stop/resume auto-scrolling of `gr.Textbox`


hysts:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  

It would be awesome if the autoscroll behavior of `gr.Textbox` was like, when a user scrolls up during auto-scrolling, auto-scrolling stops and the user can freely scroll, and auto-scrolling resumes when the user scrolls back down to the bottom.

I'm thinking of something like this behavior of `tmux`:

https://github.com/gradio-app/gradio/assets/25161192/18982f2c-60d1-4466-9136-d9709e056705

PR #5652, opened 2023-09-21
Pause autoscrolling if a user scrolls up in a `gr.Textbox` and resume autoscrolling if they go all the way down

abidlabs:
QoL improvement for autoscrolling in `gr.Textbox`

Closes: #5518 

Test code:

```py
import gradio as gr
import time

s = "abcd deaf asdfhasd asdfla;"*1000

def test(x):
    for i in range(len(s)):
        time.sleep(0.001)
        yield s[: i+1]

gr.Interface(test, "textbox", gr.Textbox(max_lines=4, autoscroll=True)).queue().launch()
```


dawoodkhan82:
@abidlabs When the textbox first starts filling up, I have to scroll down a bit to "activate" the autoscroll.

abidlabs:
@dawoodkhan82 I can't repro that behavior. What code are you using to test?

Can you test on the latest version? Happy to sync on this if needed.

dawoodkhan82 (approved):
Cannot reproduce the issue I was seeing before. So should be good to merge

abidlabs:
Sounds good thanks @dawoodkhan82 and @hysts. Let me know if the issue comes back.

yvrjsharma:
I tried it [here](https://huggingface.co/spaces/ysharma/testing_gradio_wheels) successfully. However, the scrollbar doesn't appear in the textbox. Do you know why that is happening?

PR #5653, opened 2023-09-21
Prevent Clients from accessing API endpoints that set `api_name=False`

abidlabs:
Clients should not be able to access API endpoints that have been explicitly set to `False`. This enforces that. 

See internal discussion: https://huggingface.slack.com/archives/C05KSMD8EH2/p1695345160848209?thread_ts=1695237191.303499&cid=C05KSMD8EH2

Adds a test for the python client. The js client can be manually tested:

```js

// const app = await client("gradio/hello_world");
const app = await client("abidlabs/test-client-false");
const transcription = await app.predict(0, ["Abubakar"]);
console.log(transcription);

```

### DIFF
--- a/.changeset/hip-chefs-share.md
+++ b/.changeset/hip-chefs-share.md
@@ -1,4 +1,5 @@
 ---
+"@gradio/client": minor
 "gradio": minor
 "gradio_client": minor
 ---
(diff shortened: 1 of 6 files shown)


freddyaboulton (approved):
Nice fix @abidlabs !

abidlabs:
Thanks for the review @freddyaboulton! Updated the label to be a patch -- will merge in after CI runs.

Issue #5263, opened 2023-08-18
Bug downloading files from clients with replicas with clients


freddyaboulton:
### Describe the bug

As noted [here](https://huggingface.slack.com/archives/C02QZLG8GMN/p1692345753750099), when connecting to a space that has replicas under the hood, the client may fail to download temp files associated with a prediction. This is because the file is located on a specific replica and the load balancer may not put you in the correct replica.

@XciD can help us get the url of the replica the file is in.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

-

### Screenshot

-

### Logs

```shell
-
```

### System Info

```shell
-
```

### Severity

I can work around it

XciD:
We have two solutions:
1. Use two env variables `HOSTNAME` and `SPACE_HOST`
```
HOSTNAME: s-xcid-test-adrien-4-9007e-5b6bcfbd89-r2z6g
SPACE_HOST: https://xcid-test-adrien-4.hf.space/
```
Do some magic to obtain:
```
https://xcid-test-adrien-4--r2z6g.hf.space/
```


2. Add an header in the proxy with the direct url `X-Direct-Url` (for example), and re-use it

we can’t compute the env var before spawning the pod as the name at the end is random and only known at runtime
second solution is 1 line of code in proxy code, so quite easy

abidlabs:
Side note: we should confirm that embedding demos with multiple replicas works after this fix.

abidlabs:
I'll take this issue if you haven't already started on it @freddyaboulton.

freddyaboulton:
I did start. Let me push up so you can work off it.

abidlabs:
Closed by #5668

PR #5654, opened 2023-09-21
5263 x direct url

freddyaboulton:
## Description

Fixes: #5263 

cc @abidlabs 

## 🎯 PRs Should Target Issues

## Tests


abidlabs:
@freddyaboulton wouldn't a simpler solution be to just modify

https://github.com/gradio-app/gradio/blob/8a99fc97e0109cae224072059261090e6f618c28/client/python/gradio_client/client.py#L587

to point to the URL in the `X-Direct-Url` header? i.e. use that URL as the `src` for everything?

freddyaboulton:
@abidlabs Each prediction request can be routed to a different replica, no?

abidlabs:
Perhaps @XciD can confirm, but I think if we connect to underlying replica URL (`https://xcid-test-adrien-4--r2z6g.hf.space/
`) directly -- every request will go to that machine and any files that are produced as a result are also stored on that machine

XciD:
> Perhaps @XciD can confirm, but I think if we connect to underlying replica URL (`https://xcid-test-adrien-4--r2z6g.hf.space/ `) directly -- every request will go to that machine and any files that are produced as a result are also stored on that machine

Thats right, but I think you should only connect directly for assets and use the base host for predict (for load balancing purpose)

abidlabs:
@freddyaboulton I've opened an alternative PR #5668, will go ahead and close this one.

PR #5660, opened 2023-09-22
Fix secondary hue bug in gr.themes.builder()

hellofreckles:
## Description

When building a custom theme locally using `gr.themes.builder()`, I encountered an error while attempting to modify the `Secondary Hue` under the `CoreColor` tab. Subsequently, I discovered a spelling mistake in the `builder_app.py` file and made the necessary correction.


freddyaboulton (approved):
Thank you for the fix @hellofreckles !

PR #5668, opened 2023-09-24
Fully resolve generated filepaths when running on Hugging Face Spaces with multiple replicas

abidlabs:
This PR makes a change to how we serve filepaths when running on Hugging Face Spaces with multiple replicas. 

When a Gradio app is running on a Space with $N$ replicas (e.g. https://huggingface.co/spaces/abidlabs/test-client-replica), then the "real URL" of the Gradio app can be one of $N$ different URL strings. A user is directed to a particular URL string based on a hash of their IP address. 

This means that when a file is generated on that Space (e.g. `https://huggingface.co/spaces/abidlabs/test-client-replica/file=cheetah.jpg`, it is actually only generated on one specific machine. This is fine for that user, because they will still be directed to that machine. However, if they try sharing this file with another user, the other user may get a 404 since the Spaces machine they are redirected to may be different than the original user's, and so the file may not actually exist for them. 

This PR leverages our existing `root_url` and `/proxy` routes to fully resolve any generated files. Meaning that instead of generated files having this filepath:

```
https://huggingface.co/spaces/abidlabs/test-client-replica/file=cheetah.jpg
```

they'll have:

```
https://abidlabs-test-client-replica.hf.space/proxy=https://abidlabs-test-client-replica--fttzk.hf.space/file=/tmp/gradio/4884a6ee81331aa870d691096911b279c8808bc9/cheetah.jpg
```

Notice the hash at the end of this: `https://abidlabs-test-client-replica--fttzk.hf.space`

This means that this URL should always work for all users. 

I've installed this PR on this Space: https://huggingface.co/spaces/abidlabs/test-client-replica for testing purposes. 

cc @XciD @freddyaboulton


XciD (approved):

gradio/route_utils.py
```diff
@@ -244,3 +245,21 @@ async def call_process_api(
...
+    we pass in the direct URL to the replica so that we have the fully resolved path to any files
+    on that machine. This direct URL can be shared with other users and the path will still work.
+    """
+    parsed_url = httpx.URL(replica_url)
+    stripped_url = parsed_url.copy_with(query=None)
+    stripped_url = str(stripped_url)
+    if not stripped_url.endswith("/"):
+        stripped_url += "/"
+
+    for component in config["components"]:
```
freddyaboulton, line 262: Nice approach @abidlabs ! Very clean.

Two comments:
1. Should we be creating a new copy of the config here? We're modifying the config in place and we won't add the `root_url` if it's already present. So I think we're always going to get the first root_url that gets added
2. We still need to fix the client right? It's still configured to download from `/file=` and we don't look in the config to see if a root_url is set

abidlabs: > Should we be creating a new copy of the config here? We're modifying the config in place and we won't add the root_url if it's already present. So I think we're always going to get the first root_url that gets added

Good point, I'll update in place instead

> We still need to fix the client right? It's still configured to download from /file= and we don't look in the config to see if a root_url is set

Nope the client doesn't need to know about this change. Since the Client's IP address is presumably fixed, it'll always get the correct asset. 

You can confirm with 

```py

client = Client("abidlabs/test-client-replica")
client.predict("abc")
```

abidlabs: Updated the method to fix the dict in place and installed on the Space here: https://huggingface.co/spaces/abidlabs/test-client-replica

abidlabs: + added unit tests

abidlabs: Actually fixed now @freddyaboulton

freddyaboulton (approved):
Thanks @abidlabs !

abidlabs:
Thanks @freddyaboulton @XciD!

Issue #3681, opened 2023-03-29
Select a folder and then just get the filenames from that folder as a list


FurkanGozukara:
We have a script that allow users upload files

But we are requesting a feature that we can give path and it can read all files from given folder path

`  reference_imgs = gr.UploadButton(label="Upload Guide Frames", file_types = ['.png','.jpg','.jpeg'], live=True, file_count = "directory") `

So this above code will be modified to a textbox input that takes folder path to read all files from that path

In this case images

Thank you

joytsay:
I made something similiar with tkinter:
[  #2515#issuecomment-1676824284](https://github.com/gradio-app/gradio/issues/2515#issuecomment-1676824284)

henryruhs:
@FurkanGozukara I suggest we break down File to FileUpload and FileDialog and DirectoryDialog. FileDialog and DirectoryDialog do not create temporary files and can be use as a Entrypoint for Image, Video or Textbox but without coupling.

FurkanGozukara:
> @FurkanGozukara I suggest we break down `File` to `FileUpload` and `FileDialog` and `DirectoryDialog`. `FileDialog` and `DirectoryDialog` do not create temporary files and can be use as a Entrypoint for Image, Video or Textbox but without coupling.

i am not gradio expert but thanks for suggestions 

gradio really need to fine tune an LLM and release it for questions

C0untFloyd:
@abidlabs 
Is this still being worked on? I don't think a new component is even needed, just a flag to avoid copying the temporary files around and instead use the files from a local folder. This could be even bound to the global share option on server launch.

henryruhs:
Flags or modes telling a component to behave different are a **bad design pattern** to avoid at all cost.

I rather wait 2 -3 more months for a clean solution. One component, one task is the better approach on long term.

Take a look into `gradio.Image`. It does literally everything: being a WebcamStream, ImageEditor, ImageUpload and simple Image display component. Without even know the code, I know this needs refactoring.

C0untFloyd:
I generally agree and fully when it comes to the Image Component, which tries hard to be everything. But for the File Component this isn't another extension or different usage, it's just about how the input data is prepared/used. It's a boolean choice only, act locally or not, there's almost no possibility for bad design. Just my 2 cents, let's leave it to the gradio devs, they might know best.

PS: of course splitting and separating different user dialogs into separate components would be better design, I just don't think it's really necessary for the temp file processing.

[...]

PR #5672, opened 2023-09-24
Add component <-> server direct communication support, as well as a "file explorer" component

aliabid94:
Closes: #3681 

Added ability for components to directly communicate with the backend. This PR includes a basic example component HostFile that utilizes this to get the directory structure from the backend. In `host_file.py` we have this function:

```python
    @server
    def ls(self, y: list[str] | None) -> tuple[list[str], list[str]] | None:
        ...
        files = []
        folders = []
        for f in os.listdir(absolute_path):
            if os.path.isfile(os.path.join(absolute_path, f)):
                files.append(f)
            else:
                folders.append(f)
        return folders, files
```

And then we can use this directly in our frontend component as such:

```ts
	export let server: {
		ls: (arg0: string[]) => Promise<[string[], string[]]>;
	};
        contents = await server.ls(path);
```

~~The HostFile implemented here works and was added to demonstrate how the syntax of this works. It could be extracted out as a custom component. It could also be cleaned up and properly implemented as a core component, for the purpose of having a core component that can be used as a template for a component with a server function.~~

---


abidlabs:
cc @osanseviero @apolinario who was asking for a "file explorer" type component

osanseviero:
cc @Wauplin for vis as well

julien-c:
nice! cc @gary149 too

pngwn:
I picked this up and cleaned up the `HostFile`. It has now been renamed to `FileExplorer` and focuses only on selecting files from the FS.

API looks like this:

```py
gr.FileExplorer(
    ["themes/utils"],
    glob="**/{components,themes}/**/*.py",
    root=absolute_path,
    ignore_glob="**/__init__.py",
    file_count="multiple"
)
```

Focused on globs as they are powerful for inclusion. Value is a path to select, passing a folder selects all files in that folder. When used as an input it _always_ passes a list of absolute paths, so should be pretty easy to use. `file_count` can be `"multiple" or `"single"`, same as `gr.File()`.

vidjjer:

https://github.com/gradio-app/gradio/assets/12937446/3711f70a-1505-450c-8f23-20dd4481d0e3

Run `demo/file_explorer` to test locally.

pngwn:
This needs a highlight.

demo/file_explorer/screenshot.gif
abidlabs, line 1: This screenshot.gif should be removed

demo/hello_world/config.py
abidlabs, line 1: Why is this file even in the gradio repo? It should be removed (I think it is a remnant from when we used to hardcode configs for our e2e tests)

demo/file_explorer/config.py
abidlabs, line 1: Do we need this file?

gradio/components/file_explorer.py
```diff
@@ -0,0 +1,215 @@
...
+)
+
+set_documentation_group("component")
+
+
+@document()
+class FileExplorer(Changeable, IOComponent, JSONSerializable):
+    """
+    Creates a file component that allows uploading generic file (when used as an input) and or displaying generic files (output).
+    Preprocessing: passes the uploaded file as a {tempfile._TemporaryFileWrapper} or {List[tempfile._TemporaryFileWrapper]} depending on `file_count` (or a {bytes}/{List{bytes}} depending on `type`)
```
abidlabs, line 28:
```suggestion
    Preprocessing: passes the selected file as a {str} absolute path or {list[str}} depending on `file_count`
```

gradio/components/file_explorer.py
```diff
@@ -0,0 +1,215 @@
...
+    Postprocessing: expects function to return a {str} path to a file, or {List[str]} consisting of paths to files.
+    Examples-format: a {str} path to a local file that populates the component.
+    Demos: zip_to_json, zip_files
+    """
+
+    def __init__(
+        self,
```
abidlabs, line 35: I actually think we should reorder the parameters in this component to put `glob` first, as its the most common one for users to tweak, and we should make it possible for users to do something like `gr.FileExplorer('.')`. (Similar to e.g. the Dropdown component having `choices` as the first parameter)

abidlabs:
Tried to test this @pngwn but I'm not able to run or build the frontend. I did: 

1. `pnpm i --ignore-scripts`
2. `bash scripts/build_frontend.sh`

But I'm seeing this error:

```
Failed to resolve import "dequal" from "../fileexplorer/shared/utils.ts". Does the file exist?
12:05:11 PM [vite] Internal server error: Failed to resolve import "dequal" from "../fileexplorer/shared/utils.ts". Does the file exist?
```

Same when trying to just run the frontend: 

image

Do you have any suggestions for how to fix? I was able to go through the code changes, they generally look good, although I've suggested a few tweaks for the backend. I think we'll also need to modify the `gr.load()` function so that it can handle server functions. I'm happy to make the backend changes, once I can get the frontend to work properly

As an aside, a lot of untouched files have are included in this PR because they've been reformatted by prettier. Can we remove those files from this PR to make it easier to review?

gradio/components/file_explorer.py
pngwn: This can actually be either a file or directory and the path is relative to root. Although it should (and might already) accept an absolute path as well. I haven't tested that.

I changed to 

```
    Preprocessing: passes the selected file or directory as a {str} path (relative to root) or {list[str}} depending on `file_count`
```

pngwn: Supporting `.` will probably need some frontend changes but very simple to implement.

gradio/components/file_explorer.py
```diff
@@ -0,0 +1,215 @@
...
+        visible: bool = True,
+        elem_id: str | None = None,
+        elem_classes: list[str] | str | None = None,
+        **kwargs,
+    ):
+        """
+        Parameters:
+            root: Path to root directory to select files from. If not provided, defaults to current working directory.
+            value: Default path to select, given as str file path. If callable, the function will be called whenever the app loads to set the initial value of the component.
+            type: Type of value to be returned by component. "file" returns a temporary file object with the same base name as the uploaded file, whose full path can be retrieved by file_obj.name, "binary" returns an bytes object.
```
abidlabs, line 59:
```suggestion
```

abidlabs:
Looks great @pngwn! Just noticed a couple of things. 

(1) We should handle the case better if there are no matching files. Right now, it just collapses to a very narrow height:

```py
import gradio as gr

with gr.Blocks() as demo:
    gr.FileExplorer("**/*.abc")
    
demo.launch()
```

image

(2) Right now, the glob `**/*` doesn't seem to work, although `**/*.*` does. Looking into why

(3) When you create a Blocks/Interface, for some reason, every single component has a prop in the config called `server_fns` whose value is `['ls']`. I assume that this should not be the case -- only the `gr.FileExplorer` should have that. 

Try with an example like:

```py
import gradio as gr

# with gr.Blocks() as demo:
demo = gr.Interface(lambda x:x,
                    gr.FileExplorer("**/*.*", root="flagged", file_count="single"),
                    gr.FileExplorer("**/*.*", root="flagged", file_count="single")
                   )
    
demo.launch()
```

abidlabs (approved):

pngwn:
Will fix that empty thing and merge it in.

abidlabs:
Oh we also need to implement an `Example` version of this component @pngwn

abidlabs:
Thanks for adding the example component @pngwn! One thing I noticed is that if a single file is provided as the example, then the filename is spread out over multiple lines:

```py
import gradio as gr

demo = gr.Interface(lambda x:x,
                    gr.FileExplorer("**/*.*", root="flagged"),
                    gr.FileExplorer("**/*.*", root="flagged"),
                    examples=["log.csv"]
                   )
    
demo.launch()
```

image

pngwn:
Ah, I'll fix that in the morning. I forgot that a single value can be passed.

abidlabs (approved):
Did some more testing and that was the only issue I could find. This is really nice!

freddyaboulton:
hi @AnyaCoder ! Please create a new issue so we can take a look. Please be sure to include the gradio version you are using

AnyaCoder:
> hi @AnyaCoder ! Please create a new issue so we can take a look. Please be sure to include the gradio version you are using

thanks, my gradio version is 3.45, the problem is "It is not a directory tree", these files are listed above directly

PR #5681, opened 2023-09-25
add query parameters to the `gr.Request` object through the `query_params` attribute

DarhkVoyd:
## Description

- Currently, the way to extract URL query parameters from within a Gradio app is using custom _js: `URLSeachParams(window.location.search)`.
- This PR extracts the query parameters from the page's URL and into both of requests from the frontend to the backend so that the `gr.Request` object automatically contains the `query_params`. This makes it easier to get URL query params without the need for custom js function.

Closes: #5485


abidlabs (approved):
Nice @DarhkVoyd! Works great. I'll just update the documentation a bit but otherwise lgtm

abidlabs:
The relevant tests are passing, thanks @DarhkVoyd for contributing this.

PR #5687, opened 2023-09-26
Fix functional tests, part 2

abidlabs:
Ok so after seeing https://github.com/gradio-app/gradio/pull/5682#discussion_r1336653033, I changed the defaults from `false` to `'false'` so that `inputs` and `github.events.inputs` are treated the same (both as strings). 

But actually the main issue was that the frontend cache would not get busted if the files in `js/client` were changed, even though they could change the behavior of the frontend. This is now fixed.

I made a commit to break the frontend, and then the functional tests failed as they ought to: https://github.com/gradio-app/gradio/actions/runs/6308688390/job/17128030668?pr=5687

Then I reverted the change, and the functional tests passed: https://github.com/gradio-app/gradio/actions/runs/6309397573/job/17129233002?pr=5687

### DIFF
--- a/.changeset/deep-pumas-add.md
+++ b/.changeset/deep-pumas-add.md
@@ -1,0 +1,6 @@
+---
+"@gradio/client": minor
+"gradio": minor
+---
+
+feat:Fix functional tests, part 2
--- a/.changeset/deep-pumas-add.md
+++ b/.changeset/deep-pumas-add.md
@@ -1,6 +1,0 @@
----
-"@gradio/client": minor
-"gradio": minor
----
-
-feat:Fix functional tests, part 2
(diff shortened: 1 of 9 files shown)


abidlabs:
Ok so after seeing https://github.com/gradio-app/gradio/pull/5682#discussion_r1336653033, I changed the defaults from `false` to `'false'` so that `inputs` and `github.events.inputs` are treated the same (both as strings). I also cleared the cached action files.

This seems to have fixed things. I made a commit to break the frontend, and now the functional tests are failing as they ought to: https://github.com/gradio-app/gradio/actions/runs/6308688390/job/17128030668?pr=5687

I'll revert the break and hopefully we'll see the functional tests pass again

freddyaboulton (approved):
Sweet, thanks @abidlabs !

pngwn:
`package.json` changes could too. Might be worth reviewing the skip logic at some point but I think this should serve us well for now.

Issue #5666, opened 2023-09-23
apps with `gr.Examples`  unexpected behaviour with `gr.load` 


utatane-kio:
### Describe the bug

When loading apps with `gr.load` that has `gr.Examples` component, the loaded app has unexpected components (same as the examples). 
Also, examples which are tied to a radio button is shown as null in the table of Examples. 

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

original app
```
import gradio as gr

with gr.Blocks() as demo:
    txt1 = gr.Textbox(label="input1")
    txt2 = gr.Textbox(label="input2")
    radio1 = gr.Radio(choices=["1", "2"], label="radio1")
    example = gr.Examples([["a", "b", "1"]], [txt1, txt2, radio1])

demo.launch(debug=True)
```

### Screenshot

![スクリーンショット 2023-09-23 17 37 51](https://github.com/gradio-app/gradio/assets/115803278/0001fb26-8acf-4997-a515-a0a0b2591e6c)

### System Info

```shell
gradio==3.44.4
```

### Severity

I can work around it

abidlabs:
Can reproduce! Thanks @utatane-kio we'll take a look at this

PR #5690, opened 2023-09-26
Fix incorrect behavior of `gr.load()` with `gr.Examples`

abidlabs:
Fixes: #5666

There were two subtle bugs that led to this issue. 

(1) Any components specified in a `gr.Dataset` would get rendered twice if they had not already been rendered when the `gr.Dataset` was originally created. This is not normally a problem because you almost always use `gr.Dataset`/`gr.Examples` after you have created a component. However, this was causing a problem when using `gr.load()` with `gr.Examples()` because when you call `gr.load()`, you create an examples/datasets with specific components that have not been rendered yet (they are only present in the externally-loaded config file).

We didn't have a way to tell `gr.Dataset` -- **don't render this component again**. This rendering logic was supposed to be taken care of by the `get_component_instance()` method. However, this method couldn't handle all the cases that we needed it to. So I added separate parameters `render` and `unrender`. 

(2) The second problem, which was much worse, is that the config generated by `gr.Dataset` did not have a reference to specific component objects, only the component names. This meant that you couldn't really use `gr.load()` with `gr.Dataset` if the component's attributes mattered in displaying the examples. For example, you couldn't use it with a `gr.Radio()` because now the `.choices` attribute of `gr.Radio()` affect the displayed name. I now pass in a `component_ids` in the config for a `gr.Dataset` which allow us to have a reference to the input components that a particular `gr.Dataset`. 

Test with: 

```py
import gradio as gr

gr.load("spaces/abidlabs/test-examples").launch()
```

https://huggingface.co/spaces/abidlabs/test-examples is a duplicate of the Space in the original issue but with this PR installed.


abidlabs:
Still have to add tests / fix tests but should be ready for an initial review

gradio/interface.py
```diff
@@ -250,10 +250,10 @@ def __init__(
             self.cache_examples = False
 
         self.input_components = [
-            get_component_instance(i, render=False) for i in inputs  # type: ignore
```
freddyaboulton, line 253: When I use the original code the repro still loads fine for me (no extra component). In which case is this change needed?

freddyaboulton (approved):
Tested and works well! I have a question about `unrender` though.

gradio/interface.py
abidlabs: The original behavior of setting `render=False` is the same as the current behavior of setting `unrender=True`, so yes both should have worked here.

The problem happens in `gr.Dataset` where we didn't really have a way to say -- don't render the components, but also don't unrender them. 

I also think having two explicit boolean parameters is cleaning than having a single boolean parameter that can also be `None`.

abidlabs:
Thanks for the review @freddyaboulton! I've added unit tests for both of the issues described, so this should be good to merge after CI runs.

PR #5693, opened 2023-09-26
Context-based Progress tracker

cbensimon:
## PR content
Use `gradio.context.LocalContext` in `gr.Progress` so that the injected "special arg" instance is serializable
It also has the benefit of unifying the logic with `gr.Info` and `gr.Warning`

## Unrelated
Type `gradio.utils.py.function_wrapper` args


abidlabs:
Thanks @cbensimon for this PR! Generally looks good. One issue I found is that it doesn't seem to detect nested tqdms. For example, if you run:

```py
import tqdm 
import time
import gradio as gr

with gr.Blocks() as demo:
    track_tqdm_btn = gr.Button("Bind TQDM")
    text2 = gr.Textbox()

    def track_tqdm(progress=gr.Progress(track_tqdm=True)):
        for i in tqdm.tqdm(range(5), desc="outer"):
            for j in tqdm.tqdm(range(4), desc="inner"):
                time.sleep(1)
        return "done"
    
    track_tqdm_btn.click(track_tqdm, None, text2)
    
demo.queue().launch()
```

On `main`, you'll see that both of the tqdms have been detected:

image


Whereas on this branch, only the inner tqdm is detected:

image

abidlabs (changes requested):
Would you be able to fix this? Happy to take another look afterwards

cbensimon:
Thanks for having detected this bug. I'm working on it

cbensimon:
Works great now!

Even tested with `concurrency_count=2` + `async` handler (which I think would have produced a bug prior to this PR) :

Screenshot 2023-09-27 at 11 09 09

abidlabs:
Beautiful! Works great and changes look good. Thanks @cbensimon

abidlabs (approved):

PR #5696, opened 2023-09-26
setting share=True on Spaces or in wasm should warn instead of raising error

abidlabs:
Quick fix for: [internal link]

Running on Space here: https://huggingface.co/spaces/abidlabs/share-test


dawoodkhan82 (approved):
Tested on spaces. Looks good

abidlabs:
Thanks @dawoodkhan82!

Issue #5700, opened 2023-09-27
Unable to Update Empty gr.Markdown() Initialization in Version 3.45.0


chaofanqw:
### Describe the bug

When initializing the interface without an initial value for the gr.Markdown() method, it is not possible to subsequently update its value. The issue persists regardless of using the gr.Markdown(value="New Value") method or the gr.update(value="New Value") method. **Only appears at 3.45.0.**

Steps to Reproduce:
1. Initialize gr.Markdown() without an initial value.
2. Attempt to update its value using gr.Markdown(value="New Value").
3. Alternatively, try updating using gr.update(value="New Value").

Expected Behavior:
The value of the Markdown should update to "New Value".

Actual Behavior:
The Markdown value remains unchanged.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```
import gradio as gr


def update(name):
    return f"Welcome to Gradio, {name}!"


with gr.Blocks() as demo:
    gr.Markdown("Start typing below and then click **Run** to see the output.")
    with gr.Row():
        inp = gr.Textbox(placeholder="What is your name?")
        out = gr.Markdown()
    btn = gr.Button("Run")
    btn.click(fn=update, inputs=inp, outputs=out)

demo.launch()
```

### Screenshot

![example](https://github.com/gradio-app/gradio/assets/44720017/4bdaeb6b-cb84-4027-98e3-095dc7fb8315)

### System Info

```shell
python==3.10
gradio==3.45.0
```

### Severity

I can work around it

abidlabs:
Thanks @chaofanqw -- I can reproduce the issue and have a fix in #5701.

PR #5701, opened 2023-09-27
Fix for regression in rendering empty Markdown

abidlabs:
We had a regression -- rendering an empty string / null value in Markdown broke the component. This fixes: #5700


hysts (approved):
tested it. LGTM!

abidlabs:
Thanks @hysts!

PR #5709, opened 2023-09-27
ensure client is copied along with components

pngwn:
## Description

We weren't copying the client to the`/ _frontend_code` folder, so the CLI was erroring when trying to replace the version. Only a few components use the js client, so this hasn't come up until now.

cc @freddyaboulton 

## 🎯 PRs Should Target Issues

## Tests


freddyaboulton (commented):
Thanks @pngwn !

PR #5625, opened 2023-09-20
Use ContextVar instead of threading.local()

cbensimon:
Closes https://github.com/gradio-app/gradio/issues/5623

Also add `gr.Request` in the new `LocalContext`

PR #5711, opened 2023-09-27
fix: prevent internal log_message error from `/api/predict`

cbensimon:
Since https://github.com/gradio-app/gradio/pull/5625, calling `gr.Info` or `gr.Warning` on a queue-enabled Gradio app with a request originating from `/api/predict` (instead of `/queue/join`) results in an `AssertionError`

This PR fixes this by simply logging or warning, just as if those helpers were called from outside of Gradio
(before https://github.com/gradio-app/gradio/pull/5625, calling `gr.Info` or `gr.Warning` from `/api/predict` was simply a no-op)

### DIFF
--- a/.changeset/sour-pillows-hug.md
+++ b/.changeset/sour-pillows-hug.md
@@ -1,5 +1,5 @@
 ---
-"gradio": minor
+"gradio": patch
 ---
 
 feat:fix: prevent internal log_message error from `/api/predict`
--- a/.changeset/sour-pillows-hug.md
+++ b/.changeset/sour-pillows-hug.md
@@ -2,4 +2,4 @@
 "gradio": minor
 ---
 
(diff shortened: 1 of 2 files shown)


freddyaboulton (approved):
Looks good to me @cbensimon ! Thanks for the fix.

abidlabs:
Tested with:

```py
import gradio as gr
import time

def test(x):
    time.sleep(1)
    gr.Info("abc")
    time.sleep(1)
    gr.Warning("def")
    return x

with gr.Blocks() as demo:
    a = gr.Textbox()
    b = gr.Textbox()
    a.change(test, a, b, queue=False)
    
demo.queue().launch()
```

works great, thanks for catching this @cbensimon! I'll just update the docstrings of Info and Warning to document this behavior and then we should be good to merge.

Issue #5706, opened 2023-09-27
gr.select() Method Issues with Dataframe Cells (since Version 3.43.0)


chaofanqw:
### Describe the bug

**Description:**

While interacting with a Dataframe using the `gr.select()` method in version 3.43.0, two distinct bugs were observed:

1. **Inability to Interact with Previously Selected Cell**
   - **Description:** After selecting a cell, attempting to re-select the same cell results in no interaction, and the callback function does not get called. This inhibits the ability to interact with the same cells consecutively.
   - **Expected Outcome:** The interface should remain responsive and the callback should be invoked, regardless of whether the same cell is clicked consecutively.
   - **Attachments:** Screenshot 1

2. **Interface Freeze on Last Cell Selection**
   - **Description:** Clicking on the last cell of the dataframe causes the callback function to be called; however, the interface becomes unresponsive.
   - **Expected Outcome:** The interface should remain responsive when the last cell is selected, and the callback should execute smoothly.
   - **Attachments:** Screenshot 2

**Steps to Reproduce:**

1. Initialize a Dataframe and integrate it with the `gr.select()` method.
2. Click on any cell. Then, click on the same cell again to observe the first issue.
3. Click on the last cell of the dataframe to reproduce the second issue.

**Environment:**
- Version: 3.43.0

**Additional Notes:**
These issues appear to be specific to version 3.43.0, as they were not experienced in earlier versions. Would appreciate prompt attention, as this affects the usability of the application.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```
import gradio as gr
import pandas


def update(selected_index: gr.SelectData, dataframe_origin, dataframe_target):
    val = dataframe_origin.iloc[selected_index.index[0]]
    dataframe_target = pandas.concat([dataframe_target, pandas.DataFrame([val], columns=dataframe_target.columns)],
                                     ignore_index=True)
    dataframe_origin = dataframe_origin.drop(selected_index.index[0])

    print(selected_index.index[0], dataframe_origin, dataframe_target)

    return dataframe_origin, dataframe_target


range_val = 4
dataframe_1_value = pandas.DataFrame([val for val in range(range_val)])
dataframe_2_value = pandas.DataFrame([val for val in range(range_val, range_val * 2)])

with gr.Blocks() as demo:
    with gr.Row():
        dataframe_1 = gr.Dataframe(dataframe_1_value, interactive=False)
[...]

PR #5713, opened 2023-09-27
Fixes gr.select() Method Issues with Dataframe Cells

dawoodkhan82:
## Description
Fixes gr.select() Method Issues with Dataframe Cells

Closes: #5706

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/js/dataframe/shared/Table.svelte
+++ b/js/dataframe/shared/Table.svelte
@@ -548,9 +548,11 @@
 		col?: number,
 		dir?: SortDirection
 	): void {
-		const id = selected && selected[0] in data && selected[1] in data[selected[0]]
-		? data[selected[0]][selected[1]].id
-		: null;
+		const id = null
+		//Checks if the selected cell is still in the data
+		if (selected && selected[0] in data && selected[1] in data[selected[0]]) {
+			const id = data[selected[0]][selected[1]].id
+		}
 		if (typeof col !== "number" || !dir) {
 			return;
 		}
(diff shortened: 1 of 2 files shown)


js/dataframe/shared/Table.svelte
```diff
@@ -325,7 +325,6 @@
 
 	async function handle_cell_click(i: number, j: number): Promise<void> {
 		if (dequal(editing, [i, j])) return;
-		if (dequal(selected, [i, j])) return;
```
dawoodkhan82, line 328: Not sure why we needed this line, but it was causing the "double select" bug. cc @pngwn

pngwn: Hmm, okay I guess that makes sense. Could introduce other bugs thought because we refocus at the bottom. Might be okay though.

js/dataframe/shared/Table.svelte
```diff
@@ -549,7 +548,9 @@
 		col?: number,
 		dir?: SortDirection
 	): void {
-		const id = selected ? data[selected[0]][selected[1]]?.id : null;
+		const id = selected && selected[0] in data && selected[1] in data[selected[0]]
+		? data[selected[0]][selected[1]].id
+		: null;
```
pngwn, line 553: Can we rewrite this without the short circuiting + ternaries? I can't really make sense of this line.

abidlabs (approved):
Works great! Thanks @dawoodkhan82

abidlabs:
Merging in for the patch release

Issue #5277, opened 2023-08-21
Missing docstring for `HighlightedText.color_map`


abidlabs:
### Describe the bug

Missing a docstring here: https://www.gradio.app/docs/highlightedtext

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

https://www.gradio.app/docs/highlightedtext

### System Info

```shell
N/A
```

### Severity

I can work around it

aliabd:
@abidlabs is this still an issue? Can't tell

abidlabs:
Yes, there's no docstring for `color_map` here:

<img width="903" alt="image" src="https://github.com/gradio-app/gradio/assets/1778297/8b0f98aa-b85c-4d42-84c3-c8d036db3e98">

Issue #5210, opened 2023-08-14
Add back missing copy buttons on docs 


aliabd:
The copy buttons are missing from some parts of the docs (only show up in demos but not example usage), and the guides. Should be added back.

Issue #4917, opened 2023-07-13
Better messaging to indicate that the client returns filepaths on your local computer


hoalarious:
### Describe the bug

API returns wrong filepath for audio file

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

In a Kaggle notebook:
`!pip install gradio==3.36.1`
```
import gradio as gr

def something_function(inp):
    return "/kaggle/input/dataset/output.mp3"

demo = gr.Interface(fn=something_function, inputs="audio", outputs="audio")
demo.launch(debug=True, share=True)
```

In a Colab notebook
`!pip install gradio_client==0.2.9`

```
import gradio as gr
from gradio_client import Client

def resolve_gradio_filepath(share_url, filepath):
    return f'{share_url}/file={filepath}'

share_url = 'https://b61c53#######85d48.gradio.live'

client = Client(share_url)
result = client.predict(
        'input.mp3',	# str representing input in 'inp' Audio component
        api_name="/predict"
)
print(result)
print(resolve_gradio_filepath(share_url, result))
```

### Screenshot

API returns:
/tmp/gradio/2ec306c7b672def5ea91249481954dce0e87e551/output.mp3

Going to:
https://b61c53ca2d08285d48.gradio.live/file=/tmp/gradio/2ec306c7b672def5ea91249481954dce0e87e551/output.mp3

Leads to response of file not found.

In the Kaggle notebook we can see that none of the tmp folders match the hash from the filepath.

One of these folders contain the uploaded audio file and the other is the output audio file. 

![image](https://github.com/gradio-app/gradio/assets/16618954/394f6952-57ea-4d1a-9326-5589e88aa402)

The hash in the filepath returned by the API is incorrect.

### System Info

```shell
Kaggle notebook running:
gradio==3.36.1

Google colab running:
gradio_client==0.2.9
```

### Severity

I can work around it

freddyaboulton:
Hi @hoalarious - that /tmp/ file is created in the machine you made the API request from, not the machine running the demo. So it should be in the colab machine.

hoalarious:
Hi @freddyaboulton, thanks for the quick response. I did indeed check the /tmp/file on the server and it is showing a different hash.

The Kaggle notebook is the API server.

Hosting the API on my own machine also gives a different filepath but slightly different.

It returns:
[...]

PR #5721, opened 2023-09-27
Adds copy buttons to website, and better descriptions to API Docs

aliabd:
- Adds copy buttons to all codeblocks on the website (docs, guides, landing page, demos, etc) Closes: #5210 
- Adds missing docstring to HighlightedText Closes: #5277 
- Adds description to return types in Api Docs, and uses better language for filepaths Closes:  #4917

Before & after for api docs 

Screen Shot 2023-09-27 at 4 31 39 PM

Screen Shot 2023-09-27 at 4 22 56 PM


client/python/gradio_client/serializing.py
```diff
@@ -152,7 +152,10 @@ class ImgSerializable(Serializable):
     """Expects a base64 string as input/output which is serialized to a filepath."""
 
     def serialized_info(self):
-        return {"type": "string", "description": "filepath or URL to image"}
+        return {
+            "type": "string",
+            "description": "filepath on your computer, or URL to image",
```
abidlabs, line 157: Is this clearer? 
```suggestion
            "description": "filepath on your computer (or URL) of image",
```

abidlabs: Similarly for the others

gradio/components/highlighted_text.py
```diff
@@ -52,6 +52,7 @@ def __init__(
         """
         Parameters:
             value: Default value to show. If callable, the function will be called whenever the app loads to set the initial value of the component.
+            color_map: A dictionary mapping labels to colors. The colors must be specified as hex codes.
```
abidlabs, line 55: I don't think they must be specified as hex codes actually. Would be good to maybe give an example, e.g. `{"person": "red", "location": "#FFEE22"}

abidlabs:
@aliabd there's a small issue with the API docs, which is that part of the comment overflows to the next line. Specifically the:

`in 'x' Audio component`

When you copy the code, it turns out to be invalid because it copies the comment as code. This is what gets copied:


```

client = Client("http://127.0.0.1:7868/")
result = client.predict(
				"https://github.com/gradio-app/gradio/raw/main/test/test_files/audio_sample.wav",	# str (filepath on your computer, or URL to file)
								in 'x' Audio component
				api_name="/predict"
)
print(result)
```

Could we fix this?

abidlabs:
Otherwise, looks good to me. Feels very nice to have copy buttons back again!

aliabd:
Thanks for reviewing @abidlabs. Fixed up everything, please take another look.

js/app/src/api_docs/CodeSnippets.svelte
```diff
@@ -64,19 +64,16 @@ result = client.predict(<!--
 				-->{/if}<!--
 			--><span class="desc"
 								><!--
-			-->	# {python_type.type} {#if python_type.description}({python_type.description})
-								{/if}<!--
-			-->in '{label}' <!--
+			-->	# {python_type.type} {#if python_type.description}({python_type.description}){/if}<!---->in '{label}' <!--
```
abidlabs, line 67: There should be a space before the "in":

image

abidlabs (approved):

abidlabs:
Tiny nit, otherwise lgtm! Thanks @aliabd for the nice fixes

js/app/src/api_docs/CodeSnippets.svelte
aliabd: oops! missed that

PR #5722, opened 2023-09-28
Fix for deepcopy errors when running the replica-related logic on Spaces

abidlabs:
My PR #5668 broke the deployed Space because we tried to deepcopy a config dictionary which apparently can include some values that are not pickle-able. 

I'm proposing an alternative solution here -- we instead modify the config dictionary in place but keep track of the other replica URLs, which should be overridden with a machine's own replica URL. 

I also add an async.lock to prevent two different users from editing the config dictionary at the same time. I don't think this should have any noticeable affect on performance since editing a dictionary in Python should be super fast -- but we should keep an eye to see if load times are affected on Spaces.

### DIFF
--- a/gradio/routes.py
+++ b/gradio/routes.py
@@ -53,7 +53,7 @@ from gradio.deprecation import warn_deprecation
 from gradio.exceptions import Error
 from gradio.oauth import attach_oauth
 from gradio.queueing import Estimation, Event
-from gradio.route_utils import Request  # noqa: F401
+from gradio.route_utils import Request, set_replica_url_in_config  # noqa: F401
 from gradio.state_holder import StateHolder
 from gradio.utils import (
     cancel_tasks,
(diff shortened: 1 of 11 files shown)


abidlabs:
I've installed this branch here: https://huggingface.co/spaces/abidlabs/test-client-replica 

The Spaces preview from this branch is here: https://huggingface.co/spaces/gradio-pr-deploys/pr-5722-all-demos

both seem to be working as expected!

gradio/routes.py
```diff
@@ -303,7 +303,7 @@ def login(form_data: OAuth2PasswordRequestForm = Depends()):
 
         @app.head("/", response_class=HTMLResponse)
         @app.get("/", response_class=HTMLResponse)
-        def main(request: fastapi.Request, user: str = Depends(get_current_user)):
+        async def main(request: fastapi.Request, user: str = Depends(get_current_user)):
```
freddyaboulton, line 306: Should be good to change to async since TemplateResponse doesn't block

freddyaboulton (approved):
Thanks @abidlabs !

gradio/routes.py
abidlabs: Thanks for confirming

PR #5724, opened 2023-09-28
Add host to dev mode for vite

freddyaboulton:
## Description

Not sure if this is the best way @pngwn 

## 🎯 PRs Should Target Issues

## Tests


gradio/cli/commands/components/dev.py
```diff
@@ -25,6 +25,12 @@ def _dev(
             help="The directory with the custom component source code. By default, uses the current directory."
         ),
     ] = Path("."),
+    host: Annotated[
+        str,
+        typer.Option(
+            help="The host to run the front end server on. Defaults to localhost.",
+        ),
+    ] = "localost"
```
pngwn, line 33:
```suggestion
    ] = "localhost"
```

js/preview/src/dev_server.ts
```diff
@@ -27,19 +27,21 @@ interface ServerOptions {
...
-	backend_port
+	backend_port,
+	host
 }: ServerOptions): Promise<void> {
 	process.env.gradio_mode = "dev";
 	const imports = generate_imports(component_dir);
 
 	const NODE_DIR = join(root_dir, "..", "..", "node", "dev");
-
+	const host_ = host == "localhost" ?  undefined : host;
```
pngwn, line 44: I think we can skip this and just pass `host` in as it is. We are setting a default value in the CIL that is equal to vite's default value, so passing it through shouldn't cause any issues.

pngwn (approved):
Small comment and a typo, but looks good otherwise. Thanks for fixing this @freddyaboulton!

Issue #3677, opened 2023-03-28
Share link issues


Ju1-js:
### Describe the bug

There was an issue for this before, but even the test code that worked for them doesn't work for me.
The code works for me on Gradio 1.18.0, but the moment I upgrade past it ex. 1.19.x+. The share link breaks.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```python
import gradio as gr

def greet(name):
    return "Hello " + name + "!"

demo = gr.Interface(fn=greet, inputs="text", outputs="text")
    
demo.launch(share=True, debug=True)
```

### Logs

```shell
py --version
Python 3.11.1
```
```
py -m pip list
Package            Version
------------------ ---------
...
gradio             3.23.0
...
```

### System Info

```shell
Gradio version: 3.23.0
Os: Windows 10 21H2 (OS Build 19044.2604)
Browser: Brave - Chromium
```

### Severity

Blocking upgrade to latest gradio version

Chris-pap:
I have the same problem with VSCode.

<img width="730" alt="ddd" src="https://user-images.githubusercontent.com/72304041/228462923-fb912e02-b755-42af-8a01-07b24b554326.png">

I have the latest version of vscode (1.76.2) and from Gradio (3.23.0)
Python (3.10.7)

abidlabs:
Taking a look! @Chris-pap what OS are you running?

abidlabs:
**I'm also going to pin this issue. If anyone else is experiencing share link issues, please chime into this issue**

Chris-pap:
> Taking a look! @Chris-pap what OS are you running?
Windows 11 Home , 22H2, 22621.1413

abidlabs:
Do you guys have an antivirus that's running? It looks like Windows Defender sometimes blocks the share link binary file from running. See https://github.com/gradio-app/gradio/issues/3230

Chris-pap:
> Do you guys have an antivirus that's running? It looks like Windows Defender sometimes blocks the share link binary file from running. See #3230


I have windows defender and Malwarebytes, i deactivate both of them but it didn't fix the problem.

Ju1-js:
![image](https://user-images.githubusercontent.com/40339350/228890485-f34d4459-4a5d-4737-bc28-b76a9bb9e0ed.png)
That was the issue for me! Thanks!
[...]

PR #5731, opened 2023-09-28
Added timeout and error handling for frpc tunnel

cansik:
## Description

As discussed in #3677, the current implementation for sharing gradio services does not handle [frpc](https://github.com/fatedier/frp) errors well. The error message is not displayed and the while loop to check for success runs forever.

### Technical fix
This PR adds the ability to set a timeout (in seconds) for tunneling. This allows all the logs from frpc to be collected and printed in the error stream when an error occurs. It also fixes the possibility of an infinite loop. If possible, the error or success is handled before the timeout by parsing the current logs (as in the actual implementation).

### Usability
Since creating a tunnel can take a long time (default: `30 seconds`), I would suggest showing a message that the shared link is being created. Otherwise it is not clear to the user that gradio is still doing something. Printing `Running on local URL` may give the user the impression that gradio is finished and only running locally. This PR prints the message `Creating shared link...` as user feedback.

### Example
Here is an example when the firewall blocks the request:

```
Running on local URL:  http://127.0.0.1:7860
Creating shared link...
2023/09/28 10:56:04 [W] [service.go:132] login to server failed: dial tcp 44.237.78.176:7000: i/o timeout

Could not create share link. Please check your internet connection or our status page: https://status.gradio.app.
```

Closes: #3677

### DIFF
--- a/.changeset/honest-phones-like.md
+++ b/.changeset/honest-phones-like.md
@@ -1,0 +1,5 @@
+---
+"gradio": patch
+---
+
+fix:Added timeout and error handling for frpc tunnel
--- a/gradio/tunneling.py
+++ b/gradio/tunneling.py
@@ -4,6 +4,8 @@ import platform
 import re
 import stat
 import subprocess
+import sys
+import time
 from pathlib import Path
 from typing import List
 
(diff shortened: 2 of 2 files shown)


abidlabs:
made some small tweaks -- stripping the log lines was breaking the regex to identify the url for example -- but otherwise tested and looks good to me. Thanks @cansik!

abidlabs (approved):

cansik:
Ok, that sounds great, I may have overlooked this mistake. But I guess it would make sense to add the `strip` for the messages that are added to the log, otherwise we have unnecessary empty lines in there.

The `line` usually contained a `\n`, which is basically treated as no output. Have you tried it with a blocked firewall as well? Otherwise I would do that just to see if it prints correctly now.

abidlabs:
> Ok, that sounds great, I may have overlooked this mistake. But I guess it would make sense to add the strip for the messages that are added to the log, otherwise we have unnecessary empty lines in there.

Yup makes sense, will add that back in

meihao5631:
I have perfectly solved the problem where "share=true" does not work. You need to add two lines of code to this location in the "tunneling.py" file.
the location of "tunneling.py" file: /home/[your anaconda virtual environment
]/lib/python3.11/site-packages/gradio/tunneling.py
recLj6mg4CfaMGEooD2P57tunRn7MyuTqaa

Issue #5727, opened 2023-09-28
Gallery image download button


breenglespic:
### Describe the bug

The download button in the Gallery component has different behavior depending on the image caption presence -- without a caption, it downloads a `.png` file otherwise it is some `.html` file.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```import gradio as gr
from PIL import Image


def process():
    images_wo_caption = [Image.new("RGB", (64, 64)) for _ in range(3)]
    images_with_caption = [(Image.new("RGB", (64, 64)), f"{idx}") for idx in range(3)]
    return images_wo_caption, images_with_caption


demo = gr.Interface(
    process,
    inputs=[],
    outputs=[gr.Gallery(label="without caption", columns=3), gr.Gallery(label="with caption", columns=3)],
)
demo.queue().launch()
```

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Darwin
gradio version: 3.45.1
gradio_client version: 0.5.2

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.1.1
fastapi: 0.103.1
ffmpy: 0.3.1
gradio-client==0.5.2 is not installed.
httpx: 0.25.0
huggingface-hub: 0.17.3
importlib-resources: 6.1.0
jinja2: 3.1.2
markupsafe: 2.1.3
matplotlib: 3.8.0
numpy: 1.26.0
orjson: 3.9.7
packaging: 23.1
pandas: 2.1.1
pillow: 10.0.1
pydantic: 2.4.1
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0.1
requests: 2.31.0
semantic-version: 2.10.0
typing-extensions: 4.8.0
uvicorn: 0.23.2
websockets: 11.0.3
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2023.9.2
httpx: 0.25.0
huggingface-hub: 0.17.3
packaging: 23.1
requests: 2.31.0
typing-extensions: 4.8.0
websockets: 11.0.3
```

### Severity

I can work around it

PR #5735, opened 2023-09-28
Ensure images with no caption download in gallery

hannahblair:
## Description

When getting the href value of an image url, we weren't taking into account images without a caption that have a different shape. This fixes that 🐛 

demo:

```
from PIL import Image
import gradio as gr

def process():
    images_wo_caption = [Image.new("RGB", (64, 64)) for _ in range(3)]

    images_with_caption = [(Image.new("RGB", (64, 64)), f"{idx}") for idx in range(3)]
    
    return images_wo_caption, images_with_caption

demo = gr.Interface(
    process,
    inputs=[],
    outputs=[gr.Gallery(label="without caption", columns=3), gr.Gallery(label="with caption", columns=3)],
)

if __name__ == "__main__":
    demo.launch()
```

Closes: #5727 

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/true-candles-pay.md
+++ b/.changeset/true-candles-pay.md
@@ -1,0 +1,5 @@
+---
+"gradio": patch
+---
+
+feat:Fix for deepcopy errors when running the replica-related logic on Spaces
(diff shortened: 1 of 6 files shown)


dawoodkhan82 (approved):
lgtm!

Issue #4509, opened 2023-06-14
`gr.Dataset` does not render Markdown


Rogerspy:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  

Example code here:
```python
import gradio as gr


def add_to_dataset(input_text, samples):
    samples.append([input_text, input_text, input_text])
    print(samples)
    return samples, samples

a = [['a', 'b', '1'], ['b', 'c', '[aaa](https://www.google.com)']]


with gr.Blocks() as demo:
    samples = gr.State(a)

    with gr.Row():
        with gr.Column():
            with gr.Row():
                input_text = gr.Textbox(label="text from exemple")
            with gr.Row():
                btn = gr.Button(value="Add to dataset")
            with gr.Row():
                ds = gr.Dataset(components=['text', 'text', 'markdown'], type='index', headers=['n1', 'n2', 'n3'], samples=a)

    # print('tset samples', dataset_text_list.samples)
    btn.click(fn=add_to_dataset, inputs=[input_text, samples], outputs=[ds, samples])
    demo.launch()
```

![](https://github.com/gradio-app/gradio/assets/26625102/cc4ec80a-b9ac-4fb6-ae1d-36eb1bb80e3f)

This result is OK! But, when I insert data by click the button 'Add to dataset',

![图片](https://github.com/gradio-app/gradio/assets/26625102/88c0c604-4b4c-4402-94d3-b14520c45732)

`gr.Dataset` can not render markdown text

abidlabs:
Fixed in the `v4` branch, thank you @Rogerspy!

PR #5741, opened 2023-09-28
render markdown examples with markdown

abidlabs:
Small fix to actually render the Markdown examples using Markdown

Fixes: #4509 

Test with repro in the original example. 

One slightly annoying thing is that we don't have access to the Markdown component's attributes in the example frontend component, like we do in the static/interactive versions of the component. So what I did instead was to use some sensible defaults -- sanitize html, but no latex rendering.


pngwn:
> One slightly annoying thing is that we don't have access to the Markdown component's attributes in the example frontend component, like we do in the static/interactive versions of the component.

This would be relatively easy to do as long as know which instance of each component each example belongs to (preferably via the id).

abidlabs:
> This would be relatively easy to do as long as know which instance of each component each example belongs to (preferably via the id).

Oh interesting, I actually _just_ added support for this in the backend to solve a different bug. The `Dataset` component now has a `component_ids` value that it exposes in the config.

pngwn:
I need to check where the example changes are. Not sure if they are in main or v4, to decide where to implement this.

pngwn:
I'm going to refactor component loading next week so maybe I can do it at the same time.

abidlabs:
That'd be great thanks @pngwn. Probably some custom components could also benefit from having component props in the examples

In the meantime, I'll revert this to a draft PR

abidlabs:
I'm going to do this again, but targeting `v4` now

Issue #5762, opened 2023-10-02
Gradio apps not working in Google Colab without share links


abidlabs:
### Describe the bug

Getting 403s trying to run any Gradio app in Colab without share links:

<img width="1245" alt="image" src="https://github.com/gradio-app/gradio/assets/1778297/1cd4b24e-d4a7-4294-909e-1d4310f2531d">

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```py
import gradio as gr

with gr.Blocks() as demo:
  gr.Textbox()

demo.launch()
```

### System Info

```shell
3.45.2
```

### Severity

Blocking usage of gradio

PR #5767, opened 2023-10-02
Set share=True for all Gradio apps in Colab by default

abidlabs:
Gradio apps running without share links in Colab are acting strange. While we investigate the issue, this is workaround that sets `share=True` for all Gradio apps in Colab. 

Closes: #5762


yvrjsharma (approved):
LGTM!

abidlabs:
Thanks @yvrjsharma for testing!

PR #5775, opened 2023-10-03
fix pending chatbot message styling and ensure messages with value `None` don't render

### LINKED ISSUE #5792: gr.Chatbot generates an empty dialogue bubble when provided with a 'None' value input
### Describe the bug

When interacting with the gr.Chatbot, providing a 'None' value input results in an empty dialogue bubble appearing in the chat interface. Previously, when 'None' was input, the system would skip the value and not generate any dialogue bubbles.

The bug is new since gradio==3.46.0

### Reproduction

```python
import gradio as gr

with gr.Blocks() as demo:
    gr.Chatbot([('Hello', None), ('How are you?', None)])

demo.launch()
```

### Screenshot

Screenshot 2023-10-04 at 14 56 30

### Logs

_No response_

### System Info

```shell
python=3.10
gradio=3.46.0
```

### Severity

Blocking usage of gradio

hannahblair:
## Description

PR #5699 caused the styling for the pending message to look janky. It wasn't immediately clear when testing locally but the kitchen_sink demo is a little bit slow, so you're able to see the pending message component more clearly 😬 

i've also added the hide css styling back in because that was accidentally removed too 

closes #5792.

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/rude-dots-grab.md
+++ b/.changeset/rude-dots-grab.md
@@ -3,4 +3,4 @@
 "gradio": patch
 ---
 
-fix:fix pending chatbot message styling and ensure `None` messages don't render
+fix:fix pending chatbot message styling and ensure messages with value `None` don't render


dawoodkhan82:
Good catch on this bug! One nit tho is that the three dots position would look better in the pending message bubble and centered vertically for the panel layout. I don't remember what it looked like before but would be nice to fix either in this PR or a new one.

Screenshot 2023-10-03 at 2 16 25 PM

Screenshot 2023-10-03 at 2 15 57 PM

MaximeLas:
@hannahblair It seems to me that your recent addition of the Chatbot layout introduces some 'breaking' changes for displaying empty messages (set to None).
It used to simply not display a message if set to None, but now it does. So, for example, my chatbot has the ability to output multiple messages one after another, without any user messages in between. The expected behavior would simply be to display those chatbot messages, however, v3.46.0 seems to introduce empty user messages in between. This is the case even when the new layout parameter isn't set.

It would be great if you could fix this issue as well in this PR. 🙏 
Essentially, the chatbot shouldn't display a (empty) message that was set to None.

Here is an example of empty bubbles which weren't there before. (the behavior is similar when **layout='panel'**)
The first user bubble shouldn't be shown, nor should the two middle ones.
Screenshot 2023-10-04 at 1 23 57

js/chatbot/Chatbot.test.ts
pngwn, line 1: nice

pngwn (approved):
Changes look good! Thanks for taking acre of this @hannahblair!

MaximeLas:
Thanks @hannahblair !

MaximeLas:
@dawoodkhan82 Would it be possible to release a new minor version now that those fixes have been merged? Can't wait to make use of the recent features... 😄

dawoodkhan82:
@MaximeLas Yes we'll have a new minor version out today/tomm to get this fix out!

dawoodkhan82:
@MaximeLas actually @hannahblair Just released a new version [3.46.1](https://pypi.org/project/gradio/3.46.1/)

MaximeLas:
@hannahblair @dawoodkhan82 Hey guys, thanks again for the fix and the quick release last week! :)

Just wanted to follow up here because there is still an issue related to all of this... While the empty bubbles/panels are now generally gone, they do reappear during the rendering of the message (transitions).
So I suppose your latest fix wasn't properly applied to the rendering scenario?

Here is a screen recording where you can see what I mean. During both the 1st and 2nd transitions we have empty panels showing up for my None messages:

https://github.com/gradio-app/gradio/assets/26530285/3ca5cee8-dcd4-4a48-a786-5c92ccb54837

MaximeLas:
@hannahblair hey! Apologies for the late reply to this.

I am still experiencing the same issues (also after updating to the latest gradio version 4.1.1).
 I'm not sure why you haven't managed to reproduce it... But as you asked, I've put together the following basic example for you to run, which will hopefully clarify things.

Hopefully, you're able to fix this issue soon as it looks sort of pretty bad from a UX point of view, which has forced me to downgrade to an old gradio version for now for my demos.


```
import gradio as gr

def handle_btn_clicked(
    btn_label: str,
    chat_history: list[list],
    count: int
):
    chat_history += [[f'**{btn_label}**', f'That was click #{count} ⬆️']]
    chat_history += [[None, 'Please click once again to continue with this truly revolutionary chatbot 🚀']]
    chat_history += [[None, f'Click on Yes or No to register your click #{count+1}']]

    return chat_history, count + 1


with gr.Blocks() as demo:
    with gr.Row():
        chatbot = gr.Chatbot(
            value=[[None, "Welcome! 👋\n" +
                "I'm your amazing chatbot.\n" +
                "Are you ready to start doing cool stuff with me?"]],
            label='Chatbot',
            bubble_full_width=False,
            height=900,
            layout='panel'
        )

    with gr.Row():
        yes_btn = gr.Button(
            value='Yes',
            variant='primary'
        )
        no_btn = gr.Button(
            value='No',
            variant='stop'
        )
    
    count = gr.State(value=1)
    
    for btn in [yes_btn, no_btn]:
        btn.click(
            fn=handle_btn_clicked,
            inputs=[btn, chatbot, count],
            outputs=[chatbot, count]
        )

if __name__ == '__main__':
    demo.launch()
```


Let me also just put here a screen recording of when I'm running this.

https://github.com/gradio-app/gradio/assets/26530285/cf2a6ddd-a83f-4043-989a-f2ffc7ce9309

Thanks!

MaximeLas:
@hannahblair @dawoodkhan82 @pngwn  Hi guys, could you please look into this? 🙏
Many thanks!

dawoodkhan82:
@MaximeLas Taking a look! Thanks for the updated code snippet

dawoodkhan82:
@MaximeLas I have a PR open to fix this issue (https://github.com/gradio-app/gradio/pull/6386)

MaximeLas:
Great, thanks so much ! Looking forward to the next minor version then 😉

Issue #5127, opened 2023-08-08
Gallery front-end cannot display a url from web


kaleidoscopical:
### Describe the bug

I found the `Gallery` front-end cannot display a url from web. It seems a strange prefix url is added to the front. 

example:
```html
<src="https://zb-dsw-dsw32719-80.pcs-svr.alibaba-inc.com/dsw32719/proxy/7862/file=https://images.unsplash.com/photo-1554151228-14d9def656e4?ixlib=rb-1.2.1&ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&auto=format&fit=crop&w=386&q=80">
```

The image I want to reach is from the url `https://images.unsplash.com/photo-1554151228-14d9def656e4?ixlib=rb-1.2.1&ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&auto=format&fit=crop&w=386&q=80`. But the prefix added prevents the browser to render the image correctly.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import random
import gradio as gr


with gr.Blocks() as demo:
    gallery = gr.Gallery(value=["https://images.unsplash.com/photo-1554151228-14d9def656e4?ixlib=rb-1.2.1&ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&auto=format&fit=crop&w=386&q=80"], preview=True)

demo.launch(root_path='')
```

### System Info

```shell
any
```

### Severity

Blocking usage of gradio

freddyaboulton:
Works for me @kaleidoscopical !

<img width="1794" alt="image" src="https://github.com/gradio-app/gradio/assets/41651716/1d67cf0b-7b00-4e7c-b1de-e98af56657e2">

kaleidoscopical:
hi @freddyaboulton ! Thanks for quick reply : )

It also works in my local environment with url `http://127.0.0.1:7860/file=https://images.unsplash.com/photo-1554151228-14d9def656e4?ixlib=rb-1.2.1&ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&auto=format&fit=crop&w=386&q=80`, but a prefix is also added, e.g. `http://127.0.0.1:7860/file=`.

[...]

PR #5776, opened 2023-10-03
Revert replica proxy logic and instead implement using the `root` variable

freddyaboulton:
This reverts commit d626c21e91df026b04fdb3ee5c7dba74a261cfd3.

This is related to #5127 

Fixes: #5127 

Using "fixes" to link the PR and issue. This PR should not be merged as is. Hoping someone can investigate while I go to the conference.

## Description

Closes: #(issue)

## 🎯 PRs Should Target Issues

## Tests


abidlabs:
Ok so this should be ready for review now -- we've reverted the changes to `root_path` that we had before, and now we instead use the `root` property to designate replica URLs. This avoids having to "proxy" every request through the fastapi server (which in hindsight is very expensive) and instead just modifies the file path url, which is much cleaner. 

This *does* make changes to the js client, which I'm a little nervous about, so it'd be great if we could test this thoroghly.

I've installed this PR in this Space: https://huggingface.co/spaces/abidlabs/test-client-replica where it's working properly.

pngwn (approved):
I couldn't find any issues with this and the client code looks good to me! Thanks for fixing this @abidlabs!

PR #5782, opened 2023-10-04
Name Endpoints if api_name is None

freddyaboulton:
## Description

As discussed, the api_name will be set to the name of the function. If the fn is None, set to "unnamed". If it is a lambda, it will be set to "lambda". Multiple lambdas will have an index appended to their name in the order that they're added (this logic already existed).

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/gradio/events.py
+++ b/gradio/events.py
@@ -3,6 +3,7 @@ of the on-page-load event, which is defined in gr.Blocks().load()."""
 
 from __future__ import annotations
 
+import string
 from functools import partial, wraps
 from typing import TYPE_CHECKING, Any, Callable, Literal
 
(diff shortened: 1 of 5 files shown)


gradio/events.py
```diff
@@ -224,6 +225,18 @@ def inner(*args, **kwargs):
             if isinstance(show_progress, bool):
                 show_progress = "full" if show_progress else "hidden"
 
+            if api_name is None:
+                if fn is not None:
```
abidlabs, line 229: There are a few edge cases where a user could supply a function, but it doesn't have a `__name__` (e.g. if a user supplies a `functools.partial`). We should also set the name to "unnamed" in that case

gradio/events.py
```diff
@@ -224,6 +225,18 @@ def inner(*args, **kwargs):
...
+                    api_name = "".join(
+                        [
+                            s
+                            for s in fn.__name__
+                            if s not in set(string.punctuation) - {"-", "_"}
```
abidlabs, line 234: We could potentially use `utils.strip_invalid_filename_characters` for this

abidlabs:
See small nit above, otherwise looks good! Thanks for taking care of this @freddyaboulton

gradio/events.py
freddyaboulton: Yea or a class with a `__call__` method (reason why CI is failing)

freddyaboulton:
Thanks for the comments @abidlabs ! lambda functions show up as "lambda", partial functions show up as "partial" and callable classes show up as the name of the class.

I'm setting `api_name=False` for some of the events that get auto-added to Interfaces beyond the predict function, e.g. hiding stop button, submit events on inputs that are submittable, etc

Issue #5737, opened 2023-09-28
Raise an error if an object that cannot be pickled is passed into `gr.State` 


arian81:
### Describe the bug

This is quite a weird bug specially since there's no errors. But when I return a vertexai object to be stored in state, gradio web server doesn't start. 

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```py
import gradio as gr
import vertexai
from vertexai.preview.language_models import ChatModel
from typing import List, Optional


vertexai.init(project="PROJECT NAME", location="us-central1")


def palm_init(
    persona_type: str,
    history: Optional[List[List[str]]] = None,
    custom_prompt: Optional[str] = None,
) -> ChatModel:
    chat_model = ChatModel.from_pretrained(model_name="chat-bison@001")

    chat = chat_model.start_chat()
    return chat


with gr.Blocks() as demo:
    state = gr.State(palm_init("angry"))
    butt = gr.Button()

demo.launch()
```

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Windows
gradio version: 3.44.4
gradio_client version: 0.5.1

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.1.1
fastapi: 0.103.1
ffmpy: 0.3.1
gradio-client==0.5.1 is not installed.
httpx: 0.25.0
huggingface-hub: 0.17.3
importlib-resources: 6.1.0
jinja2: 3.1.2
markupsafe: 2.1.3
matplotlib: 3.8.0
numpy: 1.26.0
orjson: 3.9.7
packaging: 23.1
pandas: 2.1.1
pillow: 10.0.1
pydantic: 2.4.1
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0.1
requests: 2.31.0
semantic-version: 2.10.0
typing-extensions: 4.8.0
uvicorn: 0.23.2
websockets: 11.0.3
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2023.9.2
httpx: 0.25.0
huggingface-hub: 0.17.3
packaging: 23.1
requests: 2.31.0
typing-extensions: 4.8.0
websockets: 11.0.3
```

### Severity

Blocking usage of gradio

abidlabs:
[...]

PR #5790, opened 2023-10-04
added try except block in `state.py`

SrijanSahaySrivastava:
added try except block in `state.py` which will raise a "ValueError"

## Description

added try except block in `state.py` which will raise a "ValueError" if the object cannot be deep copied.

Fix: #5737


abidlabs:
Thanks for the PR @SrijanSahaySrivastava! Left some comments. 

It would be great to add a test in `test/test_components.py` to test this change.

SrijanSahaySrivastava:
> Thanks for the PR @SrijanSahaySrivastava! Left some comments. 
> 
> It would be great to add a test in `test/test_components.py` to test this change. 

For some reason i can't see any of your comment.

I'll create some tests for it till then.

abidlabs:
Ah sorry I think they were in a pending state. You should be able to see them now

SrijanSahaySrivastava:
@abidlabs please check the changes now.

abidlabs:
This looks much better! The only thing is that a lambda function *can* in fact be deep-copied, so I've changed the test to try to instantiate with the gradio module itself, which is not deep-copyable. 

Thanks for contributing this PR @SrijanSahaySrivastava! I'll merge it in after the tests run

SrijanSahaySrivastava:
@hannahblair 😄  I sent it!!!

SrijanSahaySrivastava:
@hannahblair I've sent the mail, but haven't received any follow up. Do i have to follow a specific structure?

Issue #5449, opened 2023-09-07
`gr.Plot()` injected multiple Bokeh js scripts, breaking `Bokeh.Models`


relubwu:
### Describe the bug

When `gr.Plot()` is called more than once, `load_bokeh()` in `Plot.svelte` would inject Bokeh dependencies multiple times to the frontend document. This causes some conflict on the Bokeh side and subsequently causes 66 models missing from `Bokeh.Models._known_models` 

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

1. Create a simple gradio app: 
```python
def cb():
    return DataTable()

with gr.Blocks() as app: 
    btn = gr.Button()
    plt = gr.Plot()
    btn.click(cb, []. plt)

app.launch()
```
2. This snippet works fine. If you check `Bokeh.Models._known_models` on the frontend, you'll see 366 models loaded. 
3. Now, add another `gr.Plot()`: 
```python
with gr.Blocks() as app: 
    btn = gr.Button()
    plt = gr.Plot()
    gr.Plot()
    btn.click(cb, []. plt)

app.launch()
```
The first plot component would collapse due to `Error: could not resolve type 'DataTable', which could be due to a widget or a custom model not being registered before first usage`. Now `Bokeh.Models._known_models` only shows 300 models loaded, missing 66. 

4. Upon closer inspection, you would find bokeh js scripts tags appear multiple times in the frontend document: 
```html
<script src="https://cdn.bokeh.org/bokeh/release/bokeh-3.2.2.min.js"></script>
<script src="https://cdn.bokeh.org/bokeh/release/bokeh-3.2.2.min.js"></script>
<script src="https://cdn.bokeh.org/bokeh/release/bokeh-3.2.2.min.js"></script>
<script src="https://cdn.bokeh.org/bokeh/release/bokeh-widgets-3.2.2.min.js"></script>
<script src="https://cdn.bokeh.org/bokeh/release/bokeh-tables-3.2.2.min.js"></script>
<script src="https://cdn.bokeh.org/bokeh/release/bokeh-gl-3.2.2.min.js"></script>
<script src="https://cdn.bokeh.org/bokeh/release/bokeh-api-3.2.2.min.js"></script>
<script src="https://cdn.bokeh.org/bokeh/release/bokeh-widgets-3.2.2.min.js"></script>
<script src="https://cdn.bokeh.org/bokeh/release/bokeh-tables-3.2.2.min.js"></script>
[...]

PR #5795, opened 2023-10-04
Prevent bokeh from injecting bokeh js multiple times 

abidlabs:
Fixes: #5449 and does some general cleanup of `Plot.svelte`


demo/bokeh_plot/run.py
abidlabs, line 1: Removes deprecated bokeh syntax

pngwn (approved):
Looks good to me! Thanks for taking care of this @abidlabs!

abidlabs:
Awesome, thanks!

Issue #5758, opened 2023-10-01
target None in SelectData


C0untFloyd:
### Describe the bug

A variant of the old issue #5327 is back again!
When accessing target in SelectData in a select handler for any component (tested checkbox & dropdown), target is always None.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

Add a select handler to a checkbox or dropdown, SelectData.target will be None in the handler function.

### System Info

```shell
New issue in Gradio v3.45.2 wasn't present in v3.44.2
```

### Severity

I can work around it

PR #5798, opened 2023-10-04
Fix `gr.SelectData` so that the target attribute is correctly attached, and the filedata is included in the data attribute with `gr.Gallery`

abidlabs:
Fixes: #5758

Test with:

```py
import gradio as gr

with gr.Blocks() as demo:
    table = gr.Dataframe([[1, 2, 3], [4, 5, 6]])
    textbox = gr.Textbox("Hello World!")

    statement = gr.Textbox()

    def on_select(evt: gr.SelectData):  # SelectData is a subclass of EventData
        return f"You selected {evt.value} at {evt.index} from {evt.target}"

    table.select(on_select, None, statement)
    textbox.select(on_select, None, statement)

demo.launch()
```


gradio/route_utils.py
```diff
@@ -186,22 +187,30 @@ def restore_session_state(app: App, body: PredictBody):
     return session_state, iterators
 
 
+def prepare_event_data(
```
abidlabs, line 190: Extracted this as a separate method to make it easier to test

js/gallery/static/Gallery.svelte
```diff
@@ -131,7 +131,7 @@
 			if (selected_image !== null) {
 				dispatch("select", {
 					index: selected_image,
-					value: _value?.[selected_image][1]
+					value: [_value?.[selected_image][0].data, _value?.[selected_image][1]]
```
abidlabs, line 134: Before, the value of the selection was just the caption, which was usually `None`. This changes it to be [image_data, caption], where the image_data is usually a filepath/url.

abidlabs:
Let me fix the conflicts real quick

dawoodkhan82 (approved):
Looks good, also tested with chatbot. Works great

abidlabs:
Thanks @dawoodkhan82! I'll fix the conflicts and merge in

Issue #5297, opened 2023-08-22
Add support for `gr.Request` and other special typed parameters to `gr.ChatInterface`


abidlabs:
Currently, doing something like:

```py
import gradio as gr

def example(x, y, request: gr.Request):
    print(request)
    return x

gr.ChatInterface(example).launch()
```

does not work and in fact raises an error. We should add support, as requested here: https://github.com/gradio-app/gradio/issues/4410#issuecomment-1688927278

DarhkVoyd:
@abidlabs Kindly assign this issue to me, I would like to help with this. Any suggestion or pointers to where I should start?

abidlabs:
Hi @DarhkVoyd we don't formally assign issues to contributors, feel free to open PRs for any issues that don't already have PRs. 

In terms of where to start, I'd recommend looking at the error when you run the example above. You'll see it points to this part of the codebase:

https://github.com/gradio-app/gradio/blob/6b8c8afd981fea984da568e9a0bd8bfc2a9c06c4/gradio/chat_interface.py#L403

The problem is that the `_submit_fn` doesn't get the `gr.Request`. We should add a parameter to `_submit_fn` that is typed as `gr.Request` and then pass it into `self.fn` if it has such a typed parameter.

davies-w:
I need this ASAP please!

PR #5819, opened 2023-10-06
Add support for gr.Request to gr.ChatInterface

DarhkVoyd:
## Description

- Currently, `gr.Request` does not work with gr.ChatInterface and adding gr.Request parameter to the callable causes error.
- This PR adds support by adding a parameter to _submit_fn that is typed as gr.Request and then passes it into self.fn if it has a similarly-typed parameter.

Closes: #5297


DarhkVoyd:
@abidlabs Kindly review and suggest any changes.

abidlabs:
Hi @DarhkVoyd this is the right approach, thanks! There are two issues:

1. We can loosen this check a little bit:

```
if (
            "request" in fn_signature.parameters
            and fn_signature.parameters["request"].annotation == gr.Request
        ):
```

The name of the parameter doesn't have to be "request" necessarily. It could be anything, as long as the type is `gr.Request`. 

2. We need to make sure we insert the request object in the right place. It may not necessarily come at the end, after `history`. 

You can take a look at this method and perhaps reuse it:

https://github.com/gradio-app/gradio/blob/6495334673fb0fcf83c07caeebbcae3022417a0d/gradio/helpers.py#L708

Finally, would you be able to add some tests to make sure the Request object is being passed in correctly? 

Something along these lines:

https://github.com/gradio-app/gradio/blob/afcf3c48e82712067d6d00a0caedb1562eb986f8/test/test_routes.py#L518

DarhkVoyd:
@abidlabs Hey, I made some changes to loosen the check. Kindly review so that I can write the required tests. Also, I don't have much experience writing tests so some pointers would be great.

DarhkVoyd:
So, I tried writing some tests after studying present tests and I came up with this. I think maybe this matches our new changes, is this a valid test?

bcattaneo:
> So, I tried writing some tests after studying present tests and I came up with this. I think maybe this matches our new changes, is this a valid test?

I think those are fine. Just in case, [here](https://github.com/gradio-app/gradio/pull/2641/files#diff-753696ffcde1f3bbff08ba3f511b0f1cdbb087ab1d06508958800b76794d727d) are some tests that were included on the implementation for `Interface`

gradio/chat_interface.py
```diff
@@ -394,15 +396,34 @@ async def _submit_fn(
...
+        request_param = next(
+            (
+                param_name
+                for param_name, param in fn_signature.parameters.items()
+                if param.annotation == gr.Request
+            ),
+            None,
+        )
+
+        if request_param:
```
abidlabs, line 417: Is this check even necessary, I think `special_args` will take care of it either way

test/test_routes.py
```diff
@@ -586,6 +586,31 @@ def identity(name, request: gr.Request):
         output = dict(response.json())
         assert output["data"] == ["test"]
 
+    def test_request_argument_passing(self):
+        def identity(message, req: gr.Request, history):
+            return message
```
abidlabs, line 591: I think this method needs to include something like this:

https://github.com/gradio-app/gradio/blob/afcf3c48e82712067d6d00a0caedb1562eb986f8/test/test_routes.py#L521

to make sure req is actually passed in

test/test_routes.py
```diff
@@ -586,6 +586,31 @@ def identity(name, request: gr.Request):
...
+            json={"data": ["test", None], "fn_index": 0, "session_hash": "_"},
+        )
+        client.post(
+            "/api/predict/",
+            json={"data": [None, []], "fn_index": 1, "session_hash": "_"},
+        )
+        response = client.post(
+            "/api/predict/",
+            json={"data": [None, None], "fn_index": 2, "session_hash": "_"},
+        )
```
abidlabs, line 609: I don't think we need multiple requests. However, it would be good to test multiple functions, with the special `gr.Request` in different positions to confirm that the request is always passed in correctly.

abidlabs:
Thanks @DarhkVoyd this looks much better! Gave a couple points of feedback above

gradio/chat_interface.py
DarhkVoyd: You are right, removed the check. My mistake.

test/test_routes.py
DarhkVoyd: Hey, I am stuck here. Initially, I couldn't simulate the client's behavior, so I manually inspected the frontend to identify the requests being sent. Each time, I noticed three similar requests like the one below, so I adopted this approach, which worked. If it's possible to consolidate these into a single request? Regarding multiple tests, I think that won't be necessary as here its already in an arbitrary place with unrelated name so this should be fine.

DarhkVoyd:
@abidlabs I have made some changes, kindly review.

test/test_routes.py
abidlabs: Ok let me see if I can push the changes based on my suggestion

abidlabs:
Thanks @DarhkVoyd for your contribution! I made the rest of the changes that were necessary & added a couple of tests. This should be good to go now. Feel free to take a look at the [final changes here](https://github.com/gradio-app/gradio/pull/5819/files) for future reference, and thanks again!

DarhkVoyd:
@hannahblair Yeah sure, I have sent an email to the provided address. Thank you so much for this. I am very eager and excited to do more for the community.

rmuhawieh:
> ## Description
> * Currently, `gr.Request` does not work with gr.ChatInterface and adding gr.Request parameter to the callable causes error.
> * This PR adds support by adding a parameter to _submit_fn that is typed as gr.Request and then passes it into self.fn if it has a similarly-typed parameter.
> 
> Closes: #5297

Can someone please explain the implementation of how to get the gr.Request to work for the gr.ChatInterface. I keep getting the error Exception: PrivateGptUi._chat() missing 1 required keyword-only argument: 'request'

AbhilashPal:
The following works fine for me: 

```
import gradio as gr
import os

# Disable analytics for this example
os.environ["GRADIO_ANALYTICS_ENABLED"] = "false"

# Define the function to handle chat inputs
def echo_chat(text, request: gr.Request, *args):
    if request:
        print("Request headers dictionary:", request.headers)
        print("IP address:", request.client.host)
        print("Query parameters:", dict(request.query_params))
        print("Session hash:", request.session_hash)
    return text

# Set up the ChatInterface
chat_interface = gr.ChatInterface(
    fn=echo_chat,  # The function handling the input
    textbox=gr.Textbox(lines=2, placeholder="Enter your message here"),  # Input textbox configuration
    examples=["Hello", "How are you?", "What is the time?"],  # Example inputs
    cache_examples=False,
    submit_btn="Submit",
    stop_btn="Stop",
    clear_btn="Clear",
    title="Chat Interface Example",
    description="This is a chat interface demonstrating request inspection."
)

# Launch the chat interface
chat_interface.launch(
    share=False,
    server_name="localhost",
    server_port=3004
)

```

However, is it possible to pass the request in `additional_inputs` when chat interface is defined ?

Issue #5582, opened 2023-09-18
dropdown becomes inconsistent with change effects


pseudotensor:
### Describe the bug

When over internet connection (not local) there is a 0.5s-1s delay in the change operation, and if user changes the dropdown on a shorter timescale, then the effects are somehow not queued in sequence properly and the UI state becomes inconsistent with the cumulative changes.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

* Need to run over slower internet, like server in India and UI in USA, or go over slow connection to mimic.
* Run script
* Do what video shows, by quickly deselecting some, see how not consistent in end.
* Worse when there is queue of events, so lag is even longer

```
import random
import time

import gradio as gr


def sentence_builder(quantity, animal, countries, place, activity_list, morning):
    return f"""The {quantity} {animal}s from {" and ".join(countries)} went to the {place} where they {" and ".join(activity_list)} until the {"morning" if morning else "night"}"""


demo = gr.Blocks()

with demo:
    num = 10
    choices = [str(x) for x in list(range(0, num))]
    choices0 = [str(x) for x in list(range(0, num - 1))]
    vis_list = gr.Dropdown(
        choices,
        value=choices0,
        multiselect=True,
        label="Model",
    )

    msg = gr.Textbox()

    def respond(*args):
        message = args[0]
        chat_historys = args[1:]
        bot_message = random.choice(["How are you?", "I love you", "I'm very hungry"])
        for chat_history in chat_historys:
            chat_history.append((message, bot_message))
        time.sleep(2)
        return tuple(chat_historys)

    chatbots = [gr.Chatbot(visible=str(x) in choices0, label=str(x)) for x in range(0, num)]
    msg.submit(respond, [msg] + chatbots, chatbots)

    def vis_change(vis_list1):
        vis_list1 = [int(x) for x in vis_list1]
        ret_list = [gr.update(visible=xi in vis_list1) for xi in list(range(0, 10))]
        return tuple(ret_list)
    vis_list.change(vis_change, inputs=vis_list, outputs=chatbots)
demo.queue()
demo.launch(server_name='0.0.0.0')

```

### Screenshot

https://github.com/gradio-app/gradio/assets/2249614/bd3ddb45-0f66-4c01-aec2-faf48fec020c

### Logs

```shell
Not quite blocking all usage, but leads to poor UX.
```

### System Info

```shell
gradio==3.41.0
gradio_client==0.3.0
```

### Severity

[...]

Issue #1592, opened 2022-06-17
Queue bug with examples


osanseviero:
### Describe the bug

When I go to https://huggingface.co/spaces/keras-io/structured-data-classification, and I click the selected example at bottom, it shows a queue of 11/11 (it's super fast). Is it possible each element of the inputs is being updated at a time, and each of those is making a new inference call?

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

See above

### Logs

```shell
-
```

### System Info

```shell
Spaces, 3.0.11
```

### Severity

annoying

FarukOzderim:
Looks like a visual bug to me because it jumps to 1 from 11 really fast.

freddyaboulton:
The interface is live so when you click on an example, it will update every input component, and each one of those changes trigger a prediction. 

Maybe we can modify the front-end to not respect `live` for examples but sounds low priority. What do you think @aliabid94 @abidlabs ?

freddyaboulton:
This issue is very stale and as described in the previous comment, this is actually the intended behavior. So will close!

PR #5826, opened 2023-10-06
Pending events behavior

dawoodkhan82:
## Description
Change default behavior of event triggers to "wait" for pending requests to complete. Fixes issues related textbox change events not registering latest change.

Closes: #1954
Closes: #5582

## 🎯 PRs Should Target Issues

## Tests


abidlabs:
@dawoodkhan82 I think we need to merge this into `v4` not `main`, which might explain all the conflicts and file changes

dawoodkhan82:
@abidlabs So this simple change fixes issue #1954. Is there anything else needed to be dont to handle "pending" requests?

abidlabs:
I'll take a look!

abidlabs:
Note that this should also close https://github.com/gradio-app/gradio/issues/5582

abidlabs:
And it should close this as well: https://github.com/gradio-app/gradio/issues/1592 -- @dawoodkhan82 I would test and make sure this is fixed as well

abidlabs:
@dawoodkhan82 I tested this locally and it seems to work! The final event always goes through. The only thing that's a bit confusing to me is that the loading indicator does not show up for events that were triggered while another event was running. Its easiest to see this when you increase the delay significantly, like this:

```py
import gradio as gr
import time

def print_with_delay(x):
    time.sleep(5)
    return x

with gr.Blocks() as demo:
    with gr.Row():
        t1 = gr.Textbox()
        t2 = gr.Textbox()
    with gr.Row():
        t3 = gr.Textbox()
        t4 = gr.Textbox()      
        
    t1.change(print_with_delay, t1, t2)
    t3.change(lambda x:x, t3, t4)
        
demo.launch()
```

You'll see that the loading indicator appears for the first time the top change event is called (for the letter "b") for the letters "c" and "d", there is no loading indicator.

abidlabs:
Also, I think the behavior with queueing is not ideal -- all of the intermediate events add to the queue, whereas I think it would be better if they were skipped -- only the final event should be added:

```py
import gradio as gr
import time

def print_with_delay(x):
    time.sleep(5)
    return x

with gr.Blocks() as demo:
    with gr.Row():
        t1 = gr.Textbox()
        t2 = gr.Textbox()
    with gr.Row():
        t3 = gr.Textbox()
        t4 = gr.Textbox()      
        
    t1.change(print_with_delay, t1, t2)
    t3.change(lambda x:x, t3, t4)
        
demo.queue().launch()
```

dawoodkhan82:
> Also, I think the behavior with queueing is not ideal -- all of the intermediate events add to the queue, whereas I think it would be better if they were skipped -- only the final event should be added:

Ah yeah, makes sense.  Will fix that

aliabid94:
So this will also cause other events, e.g. button clicks, to fire after an event completion as well (which may be fine!). A while back when we were discussing this issue, I remember proposing 3 different event behaviours, e.g. `btn.click(on_pending="...")`:

1. `on_pending="block"`

This would not allow any submissions while an event is pending

2. `on_pending="wait"`

This would would let a second submission to go through after the pending event is complete

3.  `on_pending="submit"`

This would allow unlimited submissions while pending.

By default, `c.change()` has `on_pending="wait"`, and the other event listeners have `on_pending="block"`. Thoughts on implementing as such?

There are cases where I believe `on_pending="submit"` would be useful, e.g. starting jobs that are not blocking.

abidlabs:
> So this will also cause other events, e.g. button clicks, to fire after an event completion as well (which may be fine!). 

Ah good point. 

So if there's a job that has a long inference time, and you happen to click on the submit button twice, the job would restart as soon as it would complete, which would be quite annoying for users.

abidlabs:
I like the suggestion @aliabid94 proposed, although the name of the parameter and the values that they can take are not super intuitive for me

Maybe `event_fire_mode`, which can be `once` (default for most events), `multiple`, or `always_last` (default for `.change()`)

gradio/routes.py
```diff
@@ -575,6 +574,10 @@ async def join_queue(
                 )
                 set_task_name(task, event.session_hash, event.fn_index, batch=False)
             else:
+                if blocks._queue.event_queue.__len__() > 0:
```
dawoodkhan82, line 577: @abidlabs Not sure if this is the right way to go about "ignoring" events from the queue

abidlabs: Taking a look

gradio/blocks.py
```diff
@@ -267,6 +269,9 @@ def set_event_trigger(
         elif every:
             raise ValueError("Cannot set a value for `every` without a `fn`.")
 
+        if event_name == "change":
+            trigger_mode = "multiple"
```
abidlabs, line 273: This will override the `trigger_mode` even if a user wants to pass in a different one for `.change()`. I'd suggest setting the default value of `trigger_mode` to `None` and then if its still `None` here, setting it to "always_last" (not "multiple") for `change` and "once" for everything else

js/app/src/Blocks.svelte
```diff
@@ -502,7 +502,16 @@
...
-				make_prediction();
+				if (dep.trigger_mode === "once") {
+					if (!dep.pending_request) make_prediction();
+				} else if (dep.trigger_mode === "multiple") {
+					do {
+						dep.pending_request = false;
+						make_prediction();
+					} while (dep.pending_request);
+				} else if (dep.trigger_mode === "always_last") {
+					make_prediction();
```
abidlabs, line 513: I'm a little confused here

It seems like the original behavior (which was to just call `make_prediction()`) mapes to the case of `dep.trigger_mode === "always_last"`.

But that should be the case, it should map to `dep.trigger_mode === "once"`

gradio/blocks.py
```diff
@@ -226,6 +227,7 @@ def set_event_trigger(
             collects_event_data: whether to collect event data for this event
             trigger_after: if set, this event will be triggered after 'trigger_after' function index
             trigger_only_on_success: if True, this event will only be triggered if the previous event was successful (only applies if `trigger_after` is set)
+            trigger_mode: If "once" (default for all events except `.change()`) would not allow any submissions while an event is pending. If set to "multiple", unlimited submissions are allowed while pending, and "always_last" (default for `.change()` event) would allow a second submission after the pending event is complete.
```
abidlabs, line 230:
```suggestion
            trigger_mode: If "once" (default for all events except `.change()`) would not allow any submissions of the same event by the same if they have an an existing event is pending. If set to "multiple", unlimited submissions are allowed while pending, and "always_last" (default for the `.change()` event) always executes the final event if a chain of the same events are triggered by the same user.
```

js/app/src/Blocks.svelte
abidlabs, line 513: I'm a little confused. Are "always_last" and "multiple" switched? Can you explain what's happening here: 

```js
					do {
						dep.pending_request = false;
						make_prediction();
					} while (dep.pending_request);
```

gradio/routes.py
abidlabs: I left some comments on the frontend parts. I need to understand those better before I can comment here. But my feeling is that you shoudn't even need to make any changes to the backend if the frontend changes are correct.

If `trigger_mode` is `always_last`, then in the frontend, if an event is pending, you'd just stack the incoming requests, and only send to the backend the final one (so that'd be the only one added to the queue)

gradio/blocks.py
dawoodkhan82: Oh yeah, I changed it to "multiple" for testing, forgot to change it back

abidlabs:
@dawoodkhan82 looks like this PR is stale now because of recent changes to the `v4` branch -- let me see if I can fix it as I test it out. Otherwise might need you to rebase it

aliabid94:
> Although not sure once, multiple, always_last is any more intuitive

agree.

abidlabs:
Ok with renaming the parameter values. @dawoodkhan82 please ping when this is ready for re-review

gradio/blocks.py
```diff
@@ -184,6 +184,148 @@ def get_block_name(self) -> str:
     def get_expected_parent(self) -> type[BlockContext] | None:
         return None
 
+    def set_event_trigger(
```
abidlabs, line 187: Looks like there are two `set_event_trigger` methods in `blocks.py`?

abidlabs:
I'm seeing an issue where if I submit an event with `trigger_mode="once"` multiple times quickly, the Interface errors out. 

Example code:

```py
import gradio as gr
import time

def print_with_delay(x):
    time.sleep(2)
    return x

gr.Interface(print_with_delay, "text", "text").launch()
```

Run this and click submit multiple times, and you'll see this:

image

abidlabs:
Still isn't fully fixed @dawoodkhan82. Now no error appears, but if you press the submit button multiple times in a row, you can't re-submit after the event has finished executing. Here's what I'm seeing:
Recording 2023-10-18 at 14 06 46

dawoodkhan82:
Just pushed a fix. It seems to work fine now, but didn't get to fully test everything out.

abidlabs:
@dawoodkhan82 thanks for making the fixes! I tested with and without queuing, and the behavior of the default event listeners is great!

However, it doesn't seem possible to change the default trigger mode. For example, this gives me a `TypeError`:

```py
import gradio as gr
import time

def test(x):
    print(x)
    time.sleep(2)
    return x

with gr.Blocks() as demo:
    t1 = gr.Textbox()
    t2 = gr.Textbox()
    t1.change(test, t1, t2, trigger_mode="once")
    
demo.queue().launch()
```

It would be great to support specifying the trigger_mode in the event listeners

abidlabs:
By the way, this PR should also close https://github.com/gradio-app/gradio/issues/1592 and https://github.com/gradio-app/gradio/issues/5582, so I've added them to the parent comment. But we should confirm that is indeed the case

dawoodkhan82:
> It would be great to support specifying the trigger_mode in the event listeners

Fixed.

Also tests are passing, except deploying to vercel, which i don't think is expected to pass right?

abidlabs:
> Also tests are passing, except deploying to vercel, which i don't think is expected to pass right?

Correct. I believe @freddyaboulton is looking into this

> Fixed.

Nice, reviewing now.

freddyaboulton:
Yep on it here #6082

gradio/blocks.py
```diff
@@ -819,6 +821,11 @@ def set_event_trigger(
         elif every:
             raise ValueError("Cannot set a value for `every` without a `fn`.")
 
+        if _targets[0][1] == "change" and trigger_mode == None:
+            trigger_mode = "always_last"
+        elif trigger_mode == None:
+            trigger_mode = "once"
```
abidlabs, line 827: Nit: we could also add a check here to see if someone passes in a trigger_mode that is not "once", "multiple", or "always_last" and if so, raise a `ValueError` (right now, it silently fails)

abidlabs (approved):
Looks good @dawoodkhan82! Thanks for addressing all of the feedback

Issue #5484, opened 2023-09-11
Mention that audio is normalized when converting to wav in docs


freddyaboulton:
Hi there! I actually had this same issue today. I think it might be worth mentioning in the docs that the audio is normalized to avoid confusion.

_Originally posted by @osanseviero in https://github.com/gradio-app/gradio/issues/2789#issuecomment-1712868841_

Ifeanyi55:
I think that is a great idea!

abidlabs:
Feel free to open a PR if you are interested @Ifeanyi55!

Ifeanyi55:
I will do so

PR #5835, opened 2023-10-07
Mention that audio is normalized when converting to wav in docs

aileenvl:
## Description

When converting audio data from a float format to WAV (an audio file format), the audio is being normalized by its peak value (this means the volume of the audio is being adjusted). If algorithms modify the loudness of the audio, the user might not hear these changes because the normalization alters the volume. This could be problematic when the algorithm or model is specifically manipulating loudness and the user expects to hear these changes. 

Mention this in the documentation

Closes: #5484

### DIFF
--- a/.changeset/five-showers-bet.md
+++ b/.changeset/five-showers-bet.md
@@ -1,0 +1,5 @@
+---
+"gradio": minor
+---
+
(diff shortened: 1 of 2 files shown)


freddyaboulton (approved):
Thanks @aileenvl ! 🤗

Issue #5759, opened 2023-10-02
Dropdown Block crashed when dragging the slider


luzy99:
### Describe the bug

When I drag the slider of a Dropdown Block, the browser throws an uncatched exception, and the page is unable to interact with until refresh the page.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

Add a long choice list in Dropdown block and drag the slider

### Screenshot

![image](https://github.com/gradio-app/gradio/assets/43703357/7f50eb55-f600-462c-ad4a-34bb2dbd7cd9)
![image](https://github.com/gradio-app/gradio/assets/43703357/bd6ec532-073a-49bd-9d36-f88040f7d637)
![image](https://github.com/gradio-app/gradio/assets/43703357/d66a4d6b-602b-43d5-8830-4009f2b4ef54)

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Linux
gradio version: 3.45.2
gradio_client version: 0.5.3

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.1.0
altair: 4.2.2
fastapi: 0.88.0
ffmpy: 0.3.0
gradio-client==0.5.3 is not installed.
httpx: 0.23.3
huggingface-hub: 0.16.4
importlib-resources: 5.12.0
jinja2: 3.1.2
markupsafe: 2.1.2
matplotlib: 3.7.1
numpy: 1.23.5
orjson: 3.8.7
packaging: 23.0
pandas: 1.5.3
pillow: 8.4.0
pydantic: 1.10.7
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 5.4.1
requests: 2.28.1
semantic-version: 2.10.0
typing-extensions: 4.5.0
uvicorn: 0.21.1
websockets: 10.4
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2022.2.0
httpx: 0.23.3
huggingface-hub: 0.16.4
packaging: 23.0
requests: 2.28.1
typing-extensions: 4.5.0
websockets: 10.4
```

### Severity

Blocking usage of gradio

abidlabs:
I can reproduce with a simple example!

```py
import gradio as gr

with gr.Blocks() as demo:
    gr.Dropdown(["abc"]*100)
    
demo.launch()
```

anandshah98:
[...]

PR #5839, opened 2023-10-08
Fix error when scrolling dropdown with scrollbar

Kit-p:
## Description

As the target index is `undefined` in the on change event caused by dragging the scrollbar inside a dropdown, it should not be treated as something is selected.

Closes: #5759


abidlabs:
LGTM thanks for this contribution @Kit-p! I added a comment and then this should be good to merge in

abidlabs (approved):

Kit-p:
@abidlabs Thanks for the follow-up!

Issue #5583, opened 2023-09-18
TypeError: Cannot set property WebSocket of #<Object> which has only a getter / Gradio App not working in NextJS


TommeZ:
### Describe the bug

I can't find anything that resolves my specific issue.

I'm trying to make a fetch request to my HuggingFace space via the Gradio app on my NextJS project, but I'm getting the following error: 

> TypeError: Cannot set property WebSocket of #<Object> which has only a getter
    at eval (webpack-internal:///(rsc)/./node_modules/@gradio/client/dist/index.js:282:34)
- error node_modules\@gradio\client\dist\index.js (271:24) @ eval
- error unhandledRejection: Error [TypeError]: Cannot set property WebSocket of #<Object> which has only a getter
    at eval (webpack-internal:///(rsc)/./node_modules/@gradio/client/dist/index.js:282:34) {
  digest: undefined
}

Does anyone know if there's a way to resolve this, or know what's going on here that I can fix?

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

For private reasons, I have removed the names of the actual response picture and Hugging Face space.

`import { client } from "@gradio/client";

const HF_ACCESS_TOKEN = process.env.HF_ACCESS_TOKEN

export default async function getDetections() {

    const response = await fetch("jpg");
    const exampleImage = await response.blob();

    const app = await client("HF_Name/MLDetector", { hf_token: HF_ACCESS_TOKEN });

    const result = await app.predict("/predict", [
        exampleImage, 	// blob in 'img' Image component
    ]);

}`

### Screenshot

![Capture](https://github.com/gradio-app/gradio/assets/29313293/56ff3e33-189a-4c3e-9bfb-fc6b8e6b163f)

### System Info

```shell
Gradio Environment Information:
Operating System:  Windows


gradio version:  3.38.0   

gradio Dependencies:      
  aiofiles: 23.1.0        
  aiohttp: 3.8.5
  altair: 5.0.1
  fastapi: 0.100.0        
  ffmpy: 0.3.1
  gradio-client: 0.2.10
  httpx: 0.24.1
  huggingface-hub: 0.16.4
  jinja2: 3.0.2
  markdown-it-py: 2.2.0
  markupsafe: 2.0.1
  matplotlib: 3.7.1
  mdit-py-plugins: 0.3.3
  numpy: 1.24.3
  orjson: 3.9.2
  packaging: 23.1
  pandas: 1.3.4
[...]

PR #5840, opened 2023-10-08
Ensure websocket polyfill doesn't load if there is already a `global.Webocket` property set

Jay2theWhy:
## Description
A quick conditional check for global.WebSocket. Similar issue to this: https://github.com/vercel/next.js/issues/41866 which was resolved with the following PR: https://github.com/spacebudz/lucid/pull/108/commits/b550ebc8b0bcb9399b6df0eefff172ecad81a99e

Closes: #5583 

## Tests
All tests passed:
image


client/js/src/client.ts
```diff
@@ -259,7 +259,7 @@ export function api_factory(
 				const ws = await import("ws");
 				NodeBlob = (await import("node:buffer")).Blob;
 				//@ts-ignore
-				global.WebSocket = ws.WebSocket;
+				if (!global.WebSocket) global.WebSocket = ws.WebSocket;
```
pngwn, line 262: Could we live this check to line 258 instead? If we don't need the poly fill then we shouldn't bother loading it in the first place.

pngwn (commented):
Left a small comment. What I am curious about is why websockets exist in a non-browser environment, is the situation here that a mock or polyfill has already been loaded?

client/js/src/client.ts
Jay2theWhy: Good call. Changed 👍

pngwn (approved):
This is great. Thanks for taking care of this @Jay2theWhy!

Issue #5217, opened 2023-08-14
create a really big + complex app as a demo


pngwn:
This would be really helpful internally to understand the UX and impact of our changes on large + complex gradio apps.

We probably won't want to deploy this or tests against it due to how expensive it would be to run (in terms of time) but it would be useful as a demo.

This demo should have many components (thousands) and inference functions that do _something_ but nothing complex. It should also have no dependencies or at least no _new_ dependencies. This way it mimics a real app but is easy to run and quick to test against.

shafiqihtsham:
@abidlabs @pngwn Hello, hope you are doing well. I am very interested in working on this issue. I have read the previous comments left by @abidlabs and just wanted to clarify the complexity of the components we would be adding. Would creating something that uses an external dataset like https://huggingface.co/spaces/gradio/map_airbnb this be too much? Also what about using python libraries such as numpy, pandas etc? Just want to understand what is covered by no dependencies or no new dependencies. Thank you!

pngwn:
Hello @shafiqihtsham I am very well thank you. I hope you are also doing well. 

Using an external Dataset is not an issue as long as it doesn't make starting the app very slow and doesn't require authentication. We will use this app in various testing scenarios so it is important to that it doesn't take too long to start and that it can easily be used locally (as well as in spaces). 

Regarding dependencies, we mainly want to avoid any new, complex or heavy dependencies. We are already have quite a few dependencies and list of these are fine to use. Things like numpy, pandas, pillow are all fine. I'd just avoid things like transformers, diffusers, etc. 

The main purpose of this app isn't the logic specifically but the components being in lots of different layouts, tabs, lots of components acting as inputs to lots of other components. Different tabs with more of the same etc. 

We want it to emulate a complex app (like Stable Diffusion Web UI, etc) but without any of the heavy processing. That is obviously a little vague but outside of that we don't have any strong opinions.

xiehongxin:
您好，我已收到您的邮件，查收后会尽快给您答复，辛苦您了，祝好！

teja-787:
Hi @pngwn! I'd love to work on this. 
I'm planning to build a complex multi-tab medical/data science 
demo app with hundreds of components across multiple tabs, 
using only numpy, pandas and pillow. No heavy dependencies.
Would this approach work?

freddyaboulton:
[...]

PR #5842, opened 2023-10-09
Create app.py

sambhavnoobcoder:
created complex demo demonstrating capabilities of gradio to handle large workloads , fixes issue  #5217

## Description

issue required a demo demonstrating the capability  of gradio to handle complex apps with thousands of components and inferences . this pr closes the issue .

Closes: #( #5217)

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/demo/complex_demo/app.py
+++ b/demo/complex_demo/app.py
@@ -1,0 +1,34 @@
+import gradio as gr
+
+# Define custom input components
+def create_custom_components():
+    components = []
+    for i in range(1000):
+        components.append(gr.Textbox(label=f"Text Box {i}"))
+        components.append(gr.Slider(label=f"Slider {i}", minimum=0, maximum=100))
+        components.append(gr.Number(label=f"Number {i}"))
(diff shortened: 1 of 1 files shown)


abidlabs:
Hi @sambhavnoobcoder! Thanks for the contribution, but this app does not capture several sources of complexity that are present in complex Gradio apps, such as [the Automatic1111 Web UI](https://github.com/AUTOMATIC1111/stable-diffusion-webui).

1. It only has single event --> the complex app should have hundreds of events (they can all be mock events but there should many different
2. It should use `gr.Blocks` not `gr.Interface` since `gr.Blocks` offers more level of customization, etc.
3. There should be many different tabs --> tabs hide and display content on a page, which is something we can measure the performance of

Thinking about this some more, I wonder if the Gradio Theme Builder could serve the purposes of this complex app, wdyt @pngwn @aliabid94?

abidlabs:
I'll go ahead and close this given the points I raised above

PR #5849, opened 2023-10-09
Add HighlightedTextbox Component

gsarti:
TBD


pngwn:
@gsarti Just an FYI, we aren't accepting PRs for new components at the minute and it would be best to open an issue to discuss before working on a PR.

PR #5862, opened 2023-10-11
Remove deprecated `.update()` usage from Interface internals

abidlabs:
We used the `.update()` logic various places inside `Interface`, which meant that a warning was raised e.g. any time you clicked on the Clear button or passed in an iterator. This replaces with modern syntax.

Test code:

```py
import gradio as gr
import time

def identity(x):
    for i in range(len(x)):
        time.sleep(0.5)
        yield x[:i+1]
        
gr.Interface(identity, "textbox", "textbox").queue().launch()
```

### DIFF
--- a/.changeset/spicy-cycles-send.md
+++ b/.changeset/spicy-cycles-send.md
@@ -1,5 +1,5 @@
 ---
-"gradio": minor
+"gradio": patch
 ---
 
 feat:Remove deprecated `.update()` usage from Interface internals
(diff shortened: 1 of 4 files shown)


freddyaboulton (approved):
There's a usage of `update` in the theme builder but other than that this looks good to me @abidlabs 

https://github.com/gradio-app/gradio/blob/4e0d87e9c471fe90a344a3036d0faed9188ef6f3/gradio/themes/builder_app.py#L969

abidlabs:
Thanks for the review @freddyaboulton and nice catch. I'll fix the theme builder too and get this in.

Issue #5850, opened 2023-10-09
gradio 3.45 breaks dropdown component when multiselect is enabled


vladmandic:
### Describe the bug

as title says, when multiselect is enabled in `gr.Dropdown`, result is client-side js error (no errors on python side).

it seems that issue is introduced via #5384 
and svelte Multiselect component has its arguments shifted to the left by one
(so it receives type as label and choices become null)

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

N/A

### Screenshot

![image](https://github.com/gradio-app/gradio/assets/57876960/864dbd71-cef9-4319-a9b1-4b551129a7b3)

### Logs

```shell
attached screenshot shows chrome session with log point set (you can see where in the code)
and resulting log shows two ok dropdowns (which are not multiselect) and third which shows behavior i've described above.
```

and actual error is:
![image](https://github.com/gradio-app/gradio/assets/57876960/6a692e87-c3e1-484e-bdb8-9a8ef5ac485e)

### System Info

```shell
gradio==3.47.1
```

### Severity

Blocking usage of gradio

aliabid94:
Yikes. Fixing now

aliabid94:
Is this a gradio==3.45 or a gradio==3.47 issue? in the title you've put 3.45 but in the version you've set 3.47. I've reproduced in 3.45 but not 3.47.

vladmandic:
3.44.4 ok
3.45.0 not ok
3.46.0 not ok
3.47.0 not ok
3.47.1 not ok

aliabid94:
Can you provide a reproducible code snippet? The following works for me on 3.47.1

```
import gradio as gr


def sentence_builder(activity_list):
    return f"""We {" and ".join(activity_list)}"""


demo = gr.Interface(
    sentence_builder,
    [
        gr.Dropdown(
            ["ran", "swam", "ate", "slept"],
            multiselect=True,
        ),
    ],
    "text",
)

if __name__ == "__main__":
    demo.launch()
```

vladmandic:
this is example of a typical code through entire app and it works fine in all gradio versions below 3.45.0:

```python
import gradio as gr

class FormComponent:
    def get_expected_parent(self):
        return gr.components.Form

class DropdownMulti(FormComponent, gr.Dropdown):
    def __init__(self, **kwargs):
        super().__init__(multiselect=True, **kwargs)

    def get_block_name(self):
        return "dropdown"

[...]

PR #5874, opened 2023-10-11
Allow **kwargs to be passed as props to frontend as well

aliabid94:
Fixes: #5850


abidlabs:
Will close as we don't want any PRs targeting `main` atm

Issue #2688, opened 2022-11-21
Display Cell Colors in a Dataframe


freddyaboulton:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
In pandas, you can use the [style](https://pandas.pydata.org/pandas-docs/stable/user_guide/style.html) property to apply formatting to your data (cell background colors, rounding decimals) and then get an HTML representation of the table.

Gradio developers can display this data with the HTML component but at that point you loose the ability to add/remove/edit cells and sort by cell contents. 

It would be better if the dataframe component could take in the style defined on the backend.

**Describe the solution you'd like**  

The gr.DataFrame component takes the output of the pandas.style method. 

Open to other suggestions for how to support the colored cell case.

**Additional context**  
Came up during the Data Measurements project (https://github.com/huggingface/data-measurements-tool/pull/45)

freddyaboulton:
Need to work on this soon as more people have asked about it (@sashavor)

pngwn:
I don't think this is the right approach because it won't work with themes.

Does pandas have a way of semantically labelling cells? Stuff like the formatting of contents is fine but not things like colors. I understand the need to 'highlight' certain fields or rows/ cols but i think it  should be framed in that way. If we need our own API for this then I think that is okay too, if it is possible.

freddyaboulton:
The use case I've seen for this is creating heatmaps for your data. This is how the pandas syntax works right now, you basically pass a `colormap` that gets applied to a subset of columns that you choose.

This is the pandas syntax:

```python
import pandas as pd
import numpy as np

rs = np.random.RandomState(0)
df = pd.DataFrame(rs.rand(10, 10))
corr = df.corr()
corr.style.background_gradient(cmap='coolwarm')
```

<img width="686" alt="image" src="https://user-images.githubusercontent.com/41651716/220666267-8f5136aa-e8b4-4149-ab0c-c625d0ccd912.png">

Are you suggesting the color map be included in the theme in the theme api?

pngwn:
That is a very good question. It would certainly be cool if 'official' themes actually supported a set of visualisation pallettes. Which has now reminded me of that issue that I forgot about: #2780

freddyaboulton:
Just got another request for this here: https://discord.com/channels/879548962464493619/1139122387921666048

sedrick-keh-tri:
+1 on this. Would be very helpful to have this!

abidlabs:
[...]

PR #5877, opened 2023-10-11
Add styling (e.g. font colors and background colors) support to `gr.DataFrame` through the `pd.Styler` object

abidlabs:
This PR adds support for font colors, background colors, and other forms of styling to the `gr.DataFrame`. Building on top of #5569, it uses the `pandas.Styler` object to get this information. 

So the workflow is that the the user (1) styles their `pandas.DataFrame` using `pandas.Styler`, and then (2) passes in the `pandas.Styler` to the `gr.DataFrame`. Internally, the `gr.Dataframe` object extracts the styling information in `.postprocess` and then creates a 2D matrix of style strings that it passes to the frontend. I originally implemented this differently (with the entire css string being passed to the frontend) but this way is cleaner. Note that it *is* technically possible to pass in styling information without the use of `pandas.Styler` by manually constructing the metadata dictionary but will leave this out of documentation since its more cumbersome.

Here are a couple of examples:

```py
import pandas as pd 
import gradio as gr

df = pd.DataFrame({"A" : [14, 4, 5, 4, 1], 
				"B" : [5, 2, 54, 3, 2], 
				"C" : [20, 20, 7, 3, 8], 
				"D" : [14, 3, 6, 2, 6], 
				"E" : [23, 45, 64, 32, 23]}) 

t = df.style.highlight_max(color = 'lightgreen', axis = 0)

with gr.Blocks() as demo:
    gr.Dataframe(t)
    
demo.launch()
```

image

And

```py
import pandas as pd 
import gradio as gr

df = pd.DataFrame({"A" : [14, 4, 5, 4, 1], 
				"B" : [5, 2, 54, 3, 2], 
				"C" : [20, 20, 7, 3, 8], 
				"D" : [14, 3, 6, 2, 6], 
				"E" : [23, 45, 64, 32, 23]}) 

def highlight_cols(x): 
	df = x.copy() 
	df.loc[:, :] = 'color: purple'
	df[['B', 'C', 'E']] = 'color: green'
	return df 

s = df.style.apply(highlight_cols, axis = None)

with gr.Blocks() as demo:
    gr.Dataframe(s)
    
demo.launch()
```

 
image

Finally, adds a Story and fixes the other stories for Dataframe, which were broken.

Closes: #2688


js/dataframe/shared/EditableCell.svelte
```diff
@@ -93,6 +97,10 @@
 		background: transparent;
 	}
 
+	span.focused {
```
abidlabs, line 100: Moved the box-shadow from the `<td>` to the `<span>` so that it is visible even when there is a background-color

aliabid94:
Do you think it make more sense to make the styling a function that is part of the dataframe component? E.g.

```python
import pandas as pd 
import gradio as gr

df = pd.DataFrame({"A" : [14, 4, 5, 4, 1], 
				"B" : [5, 2, 54, 3, 2], 
				"C" : [20, 20, 7, 3, 8], 
				"D" : [14, 3, 6, 2, 6], 
				"E" : [23, 45, 64, 32, 23]}) 


def highlight_max(df):
  return df.style.highlight_max(color = 'lightgreen', axis = 0)

with gr.Blocks() as demo:
    gr.Dataframe(df, styler=highlight_max)
    
demo.launch()
```

And

```python
import pandas as pd 
import gradio as gr

df = pd.DataFrame({"A" : [14, 4, 5, 4, 1], 
				"B" : [5, 2, 54, 3, 2], 
				"C" : [20, 20, 7, 3, 8], 
				"D" : [14, 3, 6, 2, 6], 
				"E" : [23, 45, 64, 32, 23]}) 

def highlight_cols(x): 
	df = x.copy() 
	df.loc[:, :] = 'color: purple'
	df[['B', 'C', 'E']] = 'color: green'
	return df 

with gr.Blocks() as demo:
    gr.Dataframe(df, styler=highlight_cols)
```

This will be cleaner in cases where dataframe is an output of a function, so the function code doesn't get polluted with styling code, and decouples content from styling. E.g.

Currently:

```python
import pandas as pd
import numpy as np
import gradio as gr

with gr.Blocks() as demo:
  df = pd.Dataframe()
  random = gr.Button("Randomize")
  
  def randomize():
    new_df = pd.DataFrame(np.random.rand(5, 5))
    df.loc[:, :] = 'color: purple'
    df[['B', 'C', 'E']] = 'color: green'
    return df

  random.click(randomize, None, df)
```

versus decoupled:

```python
import pandas as pd
import numpy as np
import gradio as gr

with gr.Blocks() as demo:
  def style(df):
    df.loc[:, :] = 'color: purple'
    df[['B', 'C', 'E']] = 'color: green'
    return df

  df = pd.Dataframe(styler=style)
  random = gr.Button("Randomize")
  
  def randomize():
    return np.random.rand(5, 5)

  random.click(randomize, None, df)
```

So the styler would be run automatically in postprocessing

js/dataframe/shared/Table.svelte
```diff
@@ -773,7 +781,8 @@
 							on:touchstart={() => start_edit(index, j)}
 							on:click={() => handle_cell_click(index, j)}
 							on:dblclick={() => start_edit(index, j)}
-							style="width: var(--cell-width-{j});"
+							style={"width: var(--cell-width-{j});" + styling?.[index]?.[j] ||
+								""}
```
aliabid94, line 785:
```suggestion
							style:width="var(--cell-width-{j})"
							style={styling?.[index]?.[j] || ""}

aliabid94 (approved):
Tested, works well

abidlabs:
> Do you think it make more sense to make the styling a function that is part of the dataframe component? 

I actually really like this. It would allow us to handle the case of styling non-df types (e.g. regular Python arrays or numpy arrays) easily as well, cc @freddyaboulton @pngwn 

The only problem is that we've already added direct support for `pd.Styler` in #5569. So in order to preserve backwards-compatibility, we'd have to add this as second method (e.g. you can also pass in a separate `styler` function that gets applied to your value even if you update it).

 I'll file an issue for this but I think we should leave for a separate PR.

js/dataframe/shared/Table.svelte
abidlabs: Did not know you could do that, cool

Issue #5802, opened 2023-10-05
Markdown list is not rendered properly in `gr.Dataframe`


hysts:
### Describe the bug

When using markdown in `gr.Dataframe`, the bullet points of list is rendered in the next column.

Might be related  to #5523 

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr
import pandas as pd

df = pd.DataFrame(
    data={
        "a": ["a", "b"],
        "b": [
            """
- Item 1
- Item 2
- Item 3
""",
            "c",
        ],
    }
)

with gr.Blocks() as demo:
    gr.Dataframe(value=df, datatype="markdown", line_breaks=False)

if __name__ == "__main__":
    demo.queue().launch()
```

https://huggingface.co/spaces/hysts-debug/dataframe-markdown-list

### Screenshot

![](https://github.com/gradio-app/gradio/assets/25161192/f721223d-aa57-4d06-bbb7-8710213aa5a0)

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Linux
gradio version: 3.46.1
gradio_client version: 0.5.3

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.1.2
fastapi: 0.103.2
ffmpy: 0.3.1
gradio-client==0.5.3 is not installed.
httpx: 0.25.0
huggingface-hub: 0.17.3
importlib-resources: 6.1.0
jinja2: 3.1.2
markupsafe: 2.1.3
matplotlib: 3.8.0
numpy: 1.26.0
orjson: 3.9.7
packaging: 23.2
pandas: 2.1.1
pillow: 10.0.1
pydantic: 1.10.13
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0.1
requests: 2.31.0
semantic-version: 2.10.0
typing-extensions: 4.8.0
uvicorn: 0.23.2
websockets: 11.0.3
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2023.6.0
httpx: 0.25.0
huggingface-hub: 0.17.3
packaging: 23.2
requests: 2.31.0
typing-extensions: 4.8.0
websockets: 11.0.3
```

### Severity

I can work around it

SrijanSahaySrivastava:
[...]

PR #5878, opened 2023-10-12
Keep Markdown rendered lists within dataframe cells

abidlabs:
Fixes: #5802 for `<ul>` and `<ol>`:

Before:

image

After:

image

Note that chatbot markdown actually looks better without this change, so I excluded chatbot markdown from the css selector.


hysts (approved):
Thanks! LGTM!

abidlabs:
Thanks @hysts! Looking into your other issues related to the Dataframe as well

Issue #5523, opened 2023-09-13
Overflow of `gr.Dataframe` table columns


hysts:
### Describe the bug

Somehow, some of the links in the table in [this Space](https://huggingface.co/spaces/hysts/list-of-demos) are overflowing. ([internal slack](https://huggingface.slack.com/archives/C02QZLG8GMN/p1694596884767699))

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

https://huggingface.co/spaces/hysts/list-of-demos

### Screenshot

![00](https://github.com/gradio-app/gradio/assets/25161192/fb13e67b-f65f-4af2-9da0-358322a69d12)

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Linux
gradio version: 3.44.0
gradio_client version: 0.5.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.1.1
fastapi: 0.103.1
ffmpy: 0.3.1
gradio-client==0.5.0 is not installed.
httpx: 0.25.0
huggingface-hub: 0.16.4
importlib-resources: 6.0.1
jinja2: 3.1.2
markupsafe: 2.1.3
matplotlib: 3.7.3
numpy: 1.25.2
orjson: 3.9.7
packaging: 23.1
pandas: 2.0.3
pillow: 10.0.0
pydantic: 1.10.12
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0.1
requests: 2.31.0
semantic-version: 2.10.0
typing-extensions: 4.7.1
uvicorn: 0.23.2
websockets: 11.0.3
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2023.6.0
httpx: 0.25.0
huggingface-hub: 0.16.4
packaging: 23.1
requests: 2.31.0
typing-extensions: 4.7.1
websockets: 11.0.3
```

### Severity

I can work around it

rabroldan:
hey there is this issue still up for grabs?

abidlabs:
Yes it is!

abidlabs:
Ok so this issue is actually more complicated than I originally thought. The problem is that the column widths are set based on the length of the content of the visible rows. That means if there's a particularly wide column later on, its content will get cut off. There's no real way of knowing the maximum column width without losing all of the benefits of using a virtual table.

[...]

PR #5881, opened 2023-10-12
dataframe wrap links

abidlabs:
Fixes: #5523

Nvm: this is not a good solution since wrapping the links adjusts the column widths. 

I think the issue is that the column widths are set based on the length of the content of the visible rows. That means if there's a particularly wide column later on, its content will get cut off. Will see if there's a way to fix.


abidlabs:
Nvm this approach doesn't work

Issue #5606, opened 2023-09-19
unix socket support


cyberb:
Is it possible to support unix sockets?
Uvicorn seems to support it.
I need it for self hosting platform (https://github.com/syncloud/platform) where multiple apps run behind a single nginx and it is easier to control backend apps based on files versus ports.

abidlabs:
Hi @cyberb can you elaborate a bit more -- what would supporting unix sockets entail?

Just fyi we have a guide on setting up gradio apps behind nginx here: https://www.gradio.app/guides/running-gradio-on-your-web-server-with-nginx

cyberb:
I mean when running nginx (or any other reverse proxy) in front of an unknown combination of web apps (installed by user) it is easier to proxy backends by unix socket file paths like this:
```
proxy_pass      http://unix:/var/snap/$app/common/web.socket: ;
```

Instead of assigning ports per app which may result in a conflict.

I understand it is a specific use case but I think having such an option is a useful thing.

I am actually doing this for https://github.com/AUTOMATIC1111/stable-diffusion-webui/tree/master so I can add it to Syncloud app store.

I can actually try to prepare PR if you want.

abidlabs:
Thanks for explaining and for offering to make the PR @cyberb! Before you do the work, what changes would be needed to implement this in `gradio`? It sounds like a parameter in `.launch()` analogous to `server_port` or `server_name` would be needed. What internal changes would need to be done?

cyberb:
form what I can see uvicorn.Config
https://github.com/gradio-app/gradio/blob/88d43bd124792d216da445adef932a2b02f5f416/gradio/networking.py#L182
supports additional parameter called uds
https://github.com/encode/uvicorn/blob/8a8b7adf275e1edfc1cc6fa7486a6159f13c3c17/uvicorn/config.py#L201

so yes, something like uds or server_unix_socket if provided should override server_host and server_port.
Also thins like port availability checks should also become conditional.
https://github.com/gradio-app/gradio/blob/88d43bd124792d216da445adef932a2b02f5f416/gradio/networking.py#L167

I was planning to test this on stable-diffusion app first, just wanted to make sure it does not exist already.

abidlabs:
On further thought, I think this is quite niche, and we wouldn't want to expose every uvicorn argument as parameters in gradio. Instead, a different approach might be to mount Gradio within a larger FastAPI instance which can be launched on a unix socket. Would this work? 

[...]

PR #5883, opened 2023-10-12
Unix socket support

cyberb:
Clients can pass `server_name="/path/to/socket"`
Tested this on https://github.com/AUTOMATIC1111/stable-diffusion-webui

Closes: #5606


abidlabs:
Thanks @cyberb we'll take a look!

abidlabs:
Ah looking at this PR, I noticed that it requires adding `requests-unixsocket==0.3.0` to requirements.txt. That's a pretty big downside, as we're working hard to limit dependencies in `gradio` for a variety of reasons. Is this strictly necessary?

cyberb:
It is actually only because we are calling local_url for some reason here:

https://github.com/gradio-app/gradio/pull/5883/files#diff-d181109e25b5f0829054b7774d79faf8fdff3eba38069c63fcf0651a9b84d0aeL2083

abidlabs:
> It is actually only because we are calling local_url for some reason here:

The reason is in case the Gradio app is launched from another FastAPI app, we need to ensure that the startup events are correctly called. 

If you would like to duplicate the code and include it in your PR to avoid the external dependency, we could consider this PR. Otherwise, its a no-go from my perspective because adding more dependencies is problematic from a maintenance and future-proofing perspective.

cyberb:
just pushed, is something like this better?
let me test it one more time, I will let you know tomorrow.

abidlabs:
We could consider this yes, but if we do add this functionality as something we officially support in gradio, we'll want to add tests to ensure that we don't break this in future versions.

cyberb:
added a test for test_start_server in unix socket mode.

I could copy requests_unixsocket lib tests as well but they are based on waitress lib looks like, is it ok?
https://github.com/msabramo/requests-unixsocket/blob/master/requests_unixsocket/tests/test_requests_unixsocket.py

cyberb:
did full test everything seems to work, but I am testing this change applied to v3.41.2 as this is where sd web is at (https://github.com/AUTOMATIC1111/stable-diffusion-webui)

Ta180m:
> did full test everything seems to work, but I am testing this change applied to v3.41.2 as this is where sd web is at (https://github.com/AUTOMATIC1111/stable-diffusion-webui)

@cyberb Are you sure that this change works for the SD web UI? I was getting Websockets-related errors so I had to use the flag `--no-gradio-queue` to force the SD web UI to use HTTP requests instead of Websockets.

(Also, you might want to consider setting the Unix socket's permissions to 660 by default for security reasons or make the permissions configurable somehow.)

cyberb:
In the end I had to use gradio 3.42.2 and it worked with the following parameters:
```
export COMMANDLINE_ARGS=--skip-torch-cuda-test
export GRADIO_SERVER_NAME=/var/snap/stable-diffusion/common/web.socket
export PATH=$PATH:${DIR}/python/bin
/bin/rm -f $GRADIO_SERVER_NAME
cd $DIR/python/webui
exec ${DIR}/python/bin/python -u ${DIR}/python/webui/launch.py \
  --precision full --no-half \
  --skip-prepare-environment \
  --data-dir '/var/snap/stable-diffusion/current' \
  --ckpt "/var/snap/stable-diffusion/current/v1-5-pruned-emaonly.safetensors" \
  --ldap-uri 'ldap://localhost:389' \
  --ldap-bind-dn 'cn={username},ou=users,dc=syncloud,dc=org' \
  --outdir-samples '/data/stable-diffusion/output'
```
not sure why `--no-gradio-queue` was not required for me, it definitely worked. I use this nginx config in front of it maybe that is why it worked for me:

https://github.com/syncloud/platform/blob/master/config/nginx/public.conf
```
  # apps proxy
  server {
  
    listen 443 ssl http2;
    listen [::]:443 ssl http2;
    
    server_name ~^(?P<app>.*)\.{{ domain }}$;
    
    ssl_certificate     /var/snap/platform/current/syncloud.crt;
    ssl_certificate_key /var/snap/platform/current/syncloud.key;
    ssl_protocols       TLSv1.2 TLSv1.3;
    # ssl_ciphers         ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES256-GCM-SHA384:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-CHACHA20-POLY1305:ECDHE-RSA-CHACHA20-POLY1305:DHE-RSA-AES128-GCM-SHA256:DHE-RSA-AES256-GCM-SHA384;
    ssl_ciphers ECDHE-RSA-AES128-GCM-SHA256:ECDHE-ECDSA-AES128-GCM-SHA256:ECDHE-RSA-AES256-GCM-SHA384:ECDHE-ECDSA-AES256-GCM-SHA384:DHE-RSA-AES128-GCM-SHA256:DHE-DSS-AES128-GCM-SHA256:kEDH+AESGCM:ECDHE-RSA-AES128-SHA256:ECDHE-ECDSA-AES128-SHA256:ECDHE-RSA-AES128-SHA:ECDHE-ECDSA-AES128-SHA:ECDHE-RSA-AES256-SHA384:ECDHE-ECDSA-AES256-SHA384:ECDHE-RSA-AES256-SHA:ECDHE-ECDSA-AES256-SHA:DHE-RSA-AES128-SHA256:DHE-RSA-AES128-SHA:DHE-DSS-AES128-SHA256:DHE-RSA-AES256-SHA256:DHE-DSS-AES256-SHA:DHE-RSA-AES256-SHA:!aNULL:!eNULL:!EXPORT:!DES:!RC4:!3DES:!MD5:!PSK;
    ssl_prefer_server_ciphers on;

    add_header Strict-Transport-Security "max-age=31536000; includeSubdomains";

    location / {
        proxy_set_header X-Forwarded-Proto $scheme;
        proxy_set_header X-Forwarded-Host $http_host;
        proxy_set_header X-Real-IP $remote_addr;
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_http_version 1.1;
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection "upgrade";
        proxy_set_header Host $http_host;
        proxy_pass      http://unix:/var/snap/$app/common/web.socket: ;
        proxy_redirect  http://unix:/var/snap/$app/common/web.socket: $scheme://$http_host ;
    }
  }
```

> (Also, you might want to consider setting the Unix socket's permissions to 660 by default for security reasons or make the permissions configurable somehow.)

Looks like uvicorn sets the permission anyway:
https://github.com/encode/uvicorn/blob/master/uvicorn/config.py#L522
```
  def bind_socket(self) -> socket.socket:
        logger_args: List[Union[str, int]]
        if self.uds:  # pragma: py-win32
            path = self.uds
            sock = socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
            try:
                sock.bind(path)
                uds_perms = 0o666
                os.chmod(self.uds, uds_perms)
            except OSError as exc:
                logger.error(exc)
                sys.exit(1)
```

Ta180m:
> In the end I had to use gradio 3.42.2 and it worked with the following parameters:

Thanks, I compiled your [3.43.2](https://github.com/cyberb/gradio/tree/3.43.2) branch and that fixed the error for me. I also had to edit the `requirements_versions.txt` file to prevent the SD web UI from uninstalling the patched Gradio and installing Gradio 3.41.2 from the PyPI, but the flag `--skip-prepare-environment ` that you used also solves this.

cyberb:
> I also had to edit the requirements_versions.txt file 

Right that was part of my PR to them which also adds ldap auth support:
https://github.com/AUTOMATIC1111/stable-diffusion-webui/pull/13805

Issue #5863, opened 2023-10-11
Please do not remove gr.update()


oobabooga:
After updating from gradio 3.33 to 3.47, I noticed that the `gr.update` method has been deprecated.

I would like to request that this parameter remains part of gradio. Here are some of its use cases:

1) Hiding an showing elements using events:

```
lambda: gr.update(visible=True), None, my_element
```

2) Making elements interactive or not:

```
lambda: gr.update(interactive=False), None, my_element
```

3) Updating the list of choices for a dropdown and the selected value:

```
lambda x: gr.update(choices=get_available_characters(), value=x), character_name, character_menu
```

For complex applications with many components, redeclaring the component each time one of its attributes needs to be updated becomes a major inconvenience.

abidlabs:
Just to confirm, you'd like for us to keep `gr.update()` but are fine with deprecating the specific component.update methods (since they're easy to replace with just the component themselves)?

Seems reasonable to me, cc @aliabid94 wdyt?

oobabooga:
Yes, that is correct. I think I have never used a `component.update()` method, but I use `gr.update()` very extensively. For some examples, see:

https://github.com/oobabooga/text-generation-webui/blob/f63361568c476ff4b2414a06027bc22e6123941f/modules/ui_chat.py#L238

henryruhs:
I disagree, it was the right decision to get rid of any update() method as it just returns a dict and is not typesafe.

aliabid94:
> For complex applications with many components, redeclaring the component each time one of its attributes needs to be updated becomes a major inconvenience

@oobabooga the way that updating works now is that you only set the values of the arguments that are going to be updated in the returned component, you don't have to redeclare the other properties of the component. So now, `return gr.Dropdown(visible=False)` and `return gr.update(visible=False)` have **exactly** the same behaviour - the former though also provides type hints. Isn't this strictly better?

You would have to change `gr.update` to the name of the component class that is being updated, that's all

oobabooga:
Thanks for the clarification @aliabid94. That's better, but imo there is still a loss in removing `gr.update()`. Take it for instance here:

https://github.com/oobabooga/text-generation-webui/blob/main/modules/ui.py#L180

This function is used to update all elements in the interface, and `gr.update()` is key for that.

The syntax is also very readable - `gr.update(visible=False)` is more readable than `gr.Dropdown(visible=False)`.

[...]

Issue #5746, opened 2023-09-29
Deprecated Warning for gr.update() Despite Using New Update Format


chaofanqw:
### Describe the bug

When using the recommended update format for Gradio, such as gr.Markdown() or gr.Dataframe(), from a callback function, a warning message is mistakenly displayed. The warning indicates that the usage of gr.update() is deprecated, even though the deprecated method was not used.

**Steps to Reproduce:**
Create a callback function that returns using the new format, for instance, return gr.Markdown("Updated Content").
Observe the warning message in the console or terminal indicating the use of the deprecated gr.update() method.

**Expected Behavior:**
No warning should be displayed when using the new update format as recommended.

**Actual Behavior:**
A warning message regarding the deprecated gr.update() method is displayed, even though the new format was used.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```
import gradio as gr
import logging


def update():
    return {dataframe: gr.Dataframe([[4, 5, 6]], headers=["a", "b", "c"])}


logging.basicConfig(level=logging.INFO, format='%(asctime)s - %(levelname)s - %(message)s')

with gr.Blocks() as demo:
    markdown = gr.Markdown("## Hello world!")
    dataframe = gr.Dataframe([[1, 2, 3], [4, 5, 6]], headers=["a", "b", "c"])

    button = gr.Button(value="Click me!")
    button.click(update, outputs=[dataframe])

demo.launch()
```

### Logs

```shell
lib/python3.10/site-packages/gradio/helpers.py:818: UserWarning: Using the update method is deprecated. Simply return a new object instead, e.g. `return gr.Textbox(...)` instead of `return gr.update(...)
  warnings.warn(
```

### System Info

```shell
gradio==3.45.2
python=3.10
```

### Severity

I can work around it

abidlabs:
Can reproduce! We'll take a look.

cc @aliabid94

dalalvarun:
Warning is not shown when you return gr.Dataframe as follows:

`import gradio as gr
import logging


def update():
    return gr.Dataframe([[4, 5, 6]], headers=["a", "b", "c"])


logging.basicConfig(level=logging.INFO, format='%(asctime)s - %(levelname)s - %(message)s')

with gr.Blocks() as demo:
    markdown = gr.Markdown("## Hello world!")
    dataframe = gr.Dataframe([[1, 2, 3], [4, 5, 6]], headers=["a", "b", "c"])

[...]

PR #5890, opened 2023-10-12
Remove deprecation warning from `gr.update` and clean up associated code

abidlabs:
As suggested in #5863, this removes the deprecation warning from `gr.update()` as it can be handy in some situations, such as when you need to update the same property but across different output components. 

Also cleans up some old code that was causing warnings to be emitted incorrectly. I'm pretty sure its safe to delete, but would appreciate another pair of eyes.

Closes: #5863
Closes: #5746


sumitsahoo:
Waiting to get this merged. Been receiving the warnings when using `llama-index`

```
UserWarning: Using the update method is deprecated. Simply return a new object instead, e.g. `return gr.Button(...)` instead of `return gr.Button.update(...)`.
```

gradio/blocks.py
```diff
@@ -225,14 +225,6 @@ def get_config(self):
     def update(**kwargs) -> dict:
         return {}
 
-    @classmethod
```
freddyaboulton, line 228: Looks like this was added here: https://github.com/gradio-app/gradio/pull/2309

The repros in the issue still work so I think we're ok to delete.

freddyaboulton (approved):
Looks good to me @abidlabs !

aliabid94 (approved):
LGTM

abidlabs:
Thanks folks for reviewing!

Issue #5520, opened 2023-09-13
Allow table column widths to be specified


hysts:
- [x] I have searched to see if a similar issue already exists.

(I think I raised this internally a few month ago, but it looks like I didn't open an issue.)

**Is your feature request related to a problem? Please describe.**  

Table columns sometimes become too wide and it degrades UX. For example, in [this Space](https://huggingface.co/spaces/hysts/daily-papers), the title column is too wide and it makes some columns out of the view. (There's the upvotes column on the right.)

![](https://github.com/gradio-app/gradio/assets/25161192/1aa56b9e-4dcf-41b4-9e5e-335c4b74bb6a)

Also, in the case of [this Space](https://huggingface.co/spaces/ICCV2023/ICCV2023-papers), I needed to use `gr.HTML` instead to avoid the UX degradation.

![](https://github.com/gradio-app/gradio/assets/25161192/06c6461a-a780-4308-88b1-6ec570cd8bd3)

Currently, `gr.Dataframe` has some nice features, like sorting column and scrolling, so I feel like it would be better to use `gr.Dataframe` to show tables in general, but this UX degradation is sometimes so bad that I have to choose `gr.HTML`.

It would be nice to be able to specify the column width of `gr.Dataframe`.

abidlabs:
Maybe we could provide width scales that serve as ratios for the columns -- we'll consider this, thanks @hysts

PR #5894, opened 2023-10-12
Adds `column_widths` to `gr.Dataframe` and hide overflowing text when `wrap=False`

abidlabs:
This PR adds the ability to set column widths to `gr.Dataframe` using the `column_widths` parameter, which takes in a list of strings that can be either pixel values or percentages.

Separately, it also tweaks the behavior so that if there's text that overflows a column, it is hidden, rather than overflowing into the next column. I think this looks better (if you need to see the full text, you can set `wrap=True`). 

Here's how it looks:

```py
import gradio as gr
import pandas as pd

a = []
b = []
for i in range(20):
    a.append("abcde"*i*4)
    b.append("abcde")

df = pd.DataFrame(
    data={
        "a": a,
        "b": b,
        "c": b,
        "d": b,
    },
)

with gr.Blocks() as demo:
    gr.Dataframe(value=df)
    gr.Dataframe(value=df, wrap=True)
    gr.Dataframe(value=df, column_widths=["25%", "25%", "10%", "40%"])
    gr.Dataframe(value=df, column_widths=["25%", "25%", "10%", "40%"], wrap=True)

if __name__ == "__main__":
    demo.queue().launch()
```

**No wrapping, no column widths**: If the first column is too wide, you can't see the second column unless you scroll:

image

**Wrapping, no column widths**: It tries to fit everything in the page width and shows everything, but column widths may not be what you want:

image

**No wrapping, but column widths**: Column widths look much better but some text is cut off:

image

**Wrapping and column widths**: Column widths look much better and all text is visible ✅ 

image

Fixes: #5520
Fixes: #5523


js/dataframe/shared/Table.svelte
```diff
@@ -663,6 +669,7 @@
 						<th
 							class:editing={header_edit === i}
 							aria-sort={get_sort_status(value, sort_by, sort_direction)}
+							style="width: {column_widths[i]};"
```
aliabid94, line 672:
```suggestion
style:width={column_widths.length ? column_widths[i] : undefined}
```
So no rule is applied if there's no column_widths[i]

aliabid94:
Users should be allowed to pass numbers, and the "px" suffix is automatically appended. We support widths in other contexts (e.g. min_width in other components) and there we expect numbers, not strings. 
Another approach could be numbers < 1 get appended with "%" with a 100x multiplier, and numbers >= 1 get "px" suffix.

aliabid94:
When I ran it and in your screenshots, I see a horizontal scrollbar even if there's no horizontal scroll content. (May not have been introduced in this PR.) Probably need to swap out `overflow-x: visible` with `overflow-x: auto` somewhere.

aliabid94 (approved):
LGTM beyond comments above

abidlabs:
> Users should be allowed to pass numbers, and the "px" suffix is automatically appended. We support widths in other contexts (e.g. min_width in other components) and there we expect numbers, not strings.

hmm okay will add for consistency

abidlabs:
> When I ran it and in your screenshots, I see a horizontal scrollbar even if there's no horizontal scroll content. (May not have been introduced in this PR.) Probably need to swap out overflow-x: visible with overflow-x: auto somewhere.

Just to confirm, you see the scrollbar container, but not the scrollbar itself, right?

abidlabs:
Thanks @aliabid94! I added support for ints as well. Can be tested with:

```py
import gradio as gr
import pandas as pd

a = []
b = []
for i in range(20):
    a.append("abcde"*i*4)
    b.append("abcde")

df = pd.DataFrame(
    data={
        "a": a,
        "b": b,
        "c": b,
    },
)

with gr.Blocks() as demo:
    gr.Dataframe(value=df, column_widths=["100px", 200, "50%"])

if __name__ == "__main__":
    demo.queue().launch()
```

KansaiTraining:
Excellent!
Just to report (no biggie though) that in the third option I actually didnt got the text cut but superimposed
Screenshot from 2024-06-24 13-45-27
But the fourth one worked great!

PR #5897, opened 2023-10-13
Fix Dataframe `line_breaks`

dawoodkhan82:
## Description
`line_breaks` param actually _was_ being passed down properly. But when adding two line breaks, markdown actually creates a new `<p>` tag for the following text. In the chatbot we added padding between `<p>` tags. I added the same here for dataframe to be consistent.
Screenshot 2023-10-13 at 1 08 09 AM

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/thirty-planets-smash.md
+++ b/.changeset/thirty-planets-smash.md
@@ -1,0 +1,6 @@
+---
+"@gradio/markdown": patch
+"gradio": patch
+---
+
+fix:Fix Dataframe `line_breaks`
--- a/js/markdown/static/MarkdownCode.svelte
+++ b/js/markdown/static/MarkdownCode.svelte
@@ -119,4 +119,7 @@
 	span:not(.chatbot) :global(ol) {
 		list-style-position: inside;
 	}
+	span :global(p:not(:first-child)) {
+		margin-top: var(--spacing-xxl);
+	}
 </style>


abidlabs:
Hmm agree that there should be some padding between the `<p>`. But this doesn't seem to fix the original issue because setting `line_breaks=False` should actually cause no new line to be created when there is a single new lines. 

When I run @hysts's original repro:

```py
import gradio as gr
import pandas as pd

df = pd.DataFrame(
    data={
        "col": [
            """
Single newline:
Message continues here
""",
            """
Double newline:

Message continues here
""",
            "Single newline:\nMessage continues here",
            "Double newline:\n\nMessage continues here",
        ],
    }
)

with gr.Blocks() as demo:
    with gr.Row():
        with gr.Column():
            gr.Dataframe(value=df, datatype="markdown", line_breaks=False)
        with gr.Column():
            gr.Dataframe(value=df, datatype="markdown", line_breaks=True)

if __name__ == "__main__":
    demo.queue().launch()
```

Both of the dataframes appear the same to me (just as in your screenshot) whereas the expected behavior is that the in the dataframe on the left, the "single new line: message continues here" should be all on one line.

abidlabs:
I'll merge this PR in but keep the original issue open

PR #5910, opened 2023-10-13
Use tags to identify custom component dirs and ignore uninstalled components

freddyaboulton:
## Description

During the build and dev steps, we read the TOML file and see if it's tagged as a `"gradio custom component"`. We also ignore uninstalled packages to avoid errors in parsing the results of `examine.py`

## 🎯 PRs Should Target Issues

## Tests


js/preview/src/examine.py
```diff
@@ -9,58 +9,67 @@
...
-            if get_relative_path(Path("..") / x)
-            == get_relative_path(file_location / value.TEMPLATE_DIR)
-        ]
-        if len(found) == 0:
-            artifacts.append(
-                os.path.abspath(file_location / value.TEMPLATE_DIR)
-                .replace(os.path.abspath(Path("..")), "")
-                .lstrip("/")
-            )
+if __name__ == "__main__":
```
freddyaboulton, line 12: Best practice is to but entrypoints under __name__

js/preview/src/dev.ts
```diff
@@ -89,8 +90,11 @@ function find_frontend_folders(start_path: string): string[] {
 	dir.forEach((dir) => {
 		const filepath = join(start_path, dir);
 		if (fs.existsSync(filepath)) {
-			if (fs.existsSync(join(filepath, "pyproject.toml")))
-				results.push(filepath);
+			if (fs.existsSync(join(filepath, "pyproject.toml"))) {
+				if (parse(fs.readFileSync(join(filepath, "pyproject.toml"))).project.keywords.indexOf("gradio custom component") > -1) {
```
freddyaboulton, line 94: @pngwn Let me know if this is bad lol

pngwn: It's fine but we are doing all of the parsing in python. I removed any JS parsing so we didn't parse the file twice.

js/preview/src/examine.py
```diff
@@ -9,58 +9,67 @@
...
+        pyproject_source = f.read()
+
+    pyproject_toml = parse(pyproject_source)
+    module_name = pyproject_toml["project"]["name"]
+
+    try:
+        module = importlib.import_module(module_name)
+    except:
+        # The importer will know not to attempt to load
+        # this component if this string is returned
```
pngwn, line 28: I don't think it needs to be installed. I tested this with my custom component explicitly uninstalled and not present in `pip list`. Since we are in `/backend` when this script is run the module name will act as a relative import right? Like `import package` will import whatever is in `backend/package`?

js/preview/src/index.ts
```diff
@@ -156,14 +156,15 @@ export function examine_module(
...
-				line.split("~|~|~|~");
-
-			return {
-				name: name.trim(),
-				template_dir: template_dir.trim(),
-				frontend_dir: frontend_dir.trim(),
-				component_class_id: component_class_id.trim()
-			};
-		});
+			if (line !== "~|~|~|~NOT INSTALLED~|~|~|~") {
```
pngwn, line 159: We might want to throw an error for this case. But it might actually be safer to check that the next 4 argument are always present instead of this because if there is some kind of exception thrown from the python file we won't catch it with either of these branches.

js/preview/src/dev.ts
pngwn: I think we can handle this in examine.py. We can skip any packages that don't match the tag.

js/preview/src/examine.py
freddyaboulton: This was my situation:

```
- component_dir
  - comp1
    - backend
    - frontend
    - pyproject.toml
    - demo
      - app.py
  - comp2
    - backend
    - frontend
    - pyproject.toml
  - comp3
    - backend
    - frontend
    - pyproject.toml
 ```

js/preview/src/index.ts
freddyaboulton: I don't think we should throw an error. Because maybe a component that's not used in the demo you're running is not installed. Don't think we should stop the dev mode launch in that case.

But let me rework to check for all four variables being present!

js/preview/src/examine.py
pngwn: Ah right. I'd actually like this to work without needing the packages installed. Will take a look at this case.

pngwn: Especially the build, I think it is reasonable to expect the packages to be installed in dev mode.

Issue #5916, opened 2023-10-14
Invalid Notebook error in most of the run.ipynb files in gradio/demo/<any_folder> in GitHub


Archit-Kohli:
### Describe the bug

None of the notebooks named run.ipynb in any of the demos are visible in GitHub, this is because of an error in the cell id's that are not strings. There is a simple fix to make all the cell id's a string and that makes them visible.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

Open any of the run.ipynb files in any folder in gradio/demo/ from GitHub, it will show an Invalid notebook error

### Screenshot

![image](https://github.com/gradio-app/gradio/assets/118905854/b907a084-69a6-4ee8-a69e-c2b4b162ee6f)

### System Info

```shell
I am just running it in the GitHub website on
Chrome Browser Version 117.0.5938.152 (Official Build) (64-bit)
```

### Severity

I can work around it

THEGAMECHANGER416:
I would like to work on this issue

Archit-Kohli:
It is an easy fix, I would also like to work on it with @THEGAMECHANGER416. @pngwn please assign this issue to us both if possible

pngwn:
Feel free to take it. We don't need to assign.

Archit-Kohli:
on it 🫡

PR #5917, opened 2023-10-15
Fixed visibility issue for all notebooks on GitHub

Archit-Kohli:
## Description

This pull request just changes the inner JSON format of all run.ipynb files in gradio/demo/<any-project> and makes the cell ids strings, thus making it a valid notebook which can now be seen in GitHub itself

Closes: #5916 

Please merge my pull request and if possible add the hacktoberfest label


pngwn:
@Archit-Kohli Thanks for taking care of this!

We also need to update the python file that generates the notebooks to make sure we don't reintroduce this problem. It is in `demo/generate_notebooks.py`. Could you update that as well?

Archit-Kohli:
Just did that 👍

Archit-Kohli:
> @Archit-Kohli Thanks for taking care of this!
> 
> We also need to update the python file that generates the notebooks to make sure we don't reintroduce this problem. It is in `demo/generate_notebooks.py`. Could you update that as well?

Just did that 👍

abidlabs:
Thank you @Archit-Kohli this is very nice!

Archit-Kohli:
Sure

Issue #5692, opened 2023-09-26
Tabbed Interface cannot manage two different flagging directories


EgonFerri:
### Describe the bug

When using the `gr.TabbedInterface` to create multiple Gradio interfaces and flagging the results from both demos, all flagged results are saved to the directory specified for the second demo (`flagged_2`). This behavior is unexpected, as flagged results from each demo should be saved in their respective directories.

**Steps to Reproduce:**
1. Run the provided Python script to launch the Gradio app with two demos (`demo_1` and `demo_2`) within a tabbed interface.
2. Enter a name in each of the demos and click the submit button to trigger the `greet` function.
3. Flag the results from both demos.
4. Check the contents of the `flagged_1` and `flagged_2` directories.

**Expected Result:**
Flagged results from `demo_1` should be saved in the `flagged_1` directory, and flagged results from `demo_2` should be saved in the `flagged_2` directory.

**Actual Result:**
All flagged results are saved to the `flagged_2` directory, regardless of which demo they originated from. This behavior is not in line with the expected behavior, as it mixes the flagged results from both demos in the `flagged_2` directory.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```
import gradio as gr


def greet(name):
    return "Hello " + name + "!"


demo_1 = gr.Interface(fn=greet, inputs="text", outputs="text", flagging_dir='flagged_1')
demo_2 = gr.Interface(fn=greet, inputs="text", outputs="text", flagging_dir='flagged_2')

demo_tabbed = gr.TabbedInterface(
    [demo_1, demo_2],
    ["demo1", "demo2"],
    title="Tabbed demo")

if __name__ == "__main__":
    demo_tabbed.launch()
```

If you open the app, try both demos, and flag both results, everything is saved to flagged_2

### System Info

```shell
**Environment:**
- Operating System: mac os
- Python Version: 3.9.4
- Gradio Version: 3.44.4
```

### Severity

Blocking usage of gradio

ahmedxomar101:
I am also facing the same issue. For all tabs, flagging is saved for the last passed flagging_dir.
I am using gradio==3.44.4 version.

harry-urek:
hey I was trying to solve this issue 
Can someone verify my findings I'm very new to Open Source 😔

**This is what I learned -**

- It's not a problem related to flagging_dir pointing towards a global variable
- In `Interface.py` `flagging_callback : FlaggingCallback = CSVLogger()` 
[...]

PR #5928, opened 2023-10-16
fix: allow multiple flagging directories

EgonFerri:
## Description

In order to solve the issue grant the possibility to link different flagging directories to different interfaces in the case of multiple interfaces like gr.TabbedInterface, I tried moving flagging_dir also as a parameter of flag(), and so as a parameter of __call__ in FlagMethod (all in flagging.py).
Subsequently, I adapted some code in interface.py, helpers.py, and tests to make it all work. 

This is my first public PR, I tried my best to do everything right but if not please tell me what could be done better. 

## 🎯 Target Issue

Closes: #5692

## Tests

1. I followed the guidelines to contribute (https://github.com/gradio-app/gradio/blob/main/CONTRIBUTING.md) and ran the tests: `bash scripts/run_all_tests.sh`. All test passed except sum test linked to hugging face that were not related with the addressed issue, and broke also before my changes (for unloadable spaces or hugginface API limit reach).

### DIFF
--- a/.changeset/plain-berries-tap.md
+++ b/.changeset/plain-berries-tap.md
@@ -1,0 +1,5 @@
+---
+"gradio": patch
+---
+
+fix:fix: allow multiple flagging directories
--- a/gradio/flagging.py
+++ b/gradio/flagging.py
@@ -46,6 +46,7 @@ class FlaggingCallback(ABC):
     def flag(
         self,
         flag_data: list[Any],
+        flagging_dir: str,
         flag_option: str = "",
         username: str | None = None,
     ) -> int:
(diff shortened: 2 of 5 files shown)


EgonFerri:
Could you review this, please? @abidlabs (:

abidlabs:
Thanks so much @EgonFerri will take a look when I have a chance!

abidlabs:
Hi @EgonFerri this is a nice PR, but all of these changes should not be necessary as far as I understand. Each Interface has its own FlaggingCallback, and each FlaggingCallback has its own `self.flagging_dir` attribute, which is where the data should be saved. 

I can reproduce the bug, but I think this approach (changing the `.flag()` method signature) to solve it is incorrect, since each FlaggingCallback instance already has the information it needs in order to save the data.

I don't have bandwidth to look into this more, but perhaps the information above might guide you towards a solution.

abidlabs:
Ah good catch @EgonFerri. 

How about we simply fix this by changing the default value of `flagging_callback` to `None`, and then in the constructor, doing

```py
        if flagging_callback is None:
```

EgonFerri:
@abidlabs your suggested approach was indeed the best way to solve the issue, I think. I opened a new [pull request](https://github.com/gradio-app/gradio/pull/6011) that addresses the issue in a more concise and straightforward way.

PR #5934, opened 2023-10-16
Fix styling issues with Audio, Image and Video components

aliabd:
Fixes some styling issues introduced by #5627


abidlabs:
Nice @aliabd! Changes lgtm

abidlabs:
I'll go ahead and merge in so that we can get the release ready

PR #5944, opened 2023-10-17
Show empty JSON icon when `value` is `null`

hannahblair:
## Description

We previously showed the Empty state even if we weren't actually rendering the JSON node. 

Now, if the value is `undefined` or an empty string, the JSON component will not render. 
If the JSON value is `null`, the Empty state will show. (Previously, if it was `null`, it would show the text "null"). 

Kapture 2023-10-17 at 11 38 38

Closes: #5845 

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/twenty-dryers-wave.md
+++ b/.changeset/twenty-dryers-wave.md
@@ -1,5 +1,6 @@
 ---
 "@gradio/atoms": patch
+"@gradio/json": patch
 "gradio": patch
 ---
 


pngwn (dismissed):
Looks good! Thanks for taking care of this @hannahblair!

pngwn (approved):
Looks good! Thanks @hannahblair!

PR #5955, opened 2023-10-17
Fix dev mode model3D

freddyaboulton:
## Description

Two issues with dev mode with a fresh Model3D template.

- Urls are not supported with Model3D so using `example_inputs` was causing a blank model3D component to be rendered.
- dev mode was failing because `BABYLON_LOADERS.OBJFileLoader` is not defined when the app launches. I think it's some sort of race condition that doesn't happen in main gradio.

## 🎯 PRs Should Target Issues

## Tests


js/model3D/interactive/Model3DUpload.svelte
```diff
@@ -88,7 +88,14 @@
...
 
-	BABYLON_LOADERS.OBJFileLoader.IMPORT_VERTEX_COLORS = true;
+	$: {
+		if (
+			BABYLON_LOADERS.OBJFileLoader != undefined &&
+			!BABYLON_LOADERS.OBJFileLoader.IMPORT_VERTEX_COLORS
+		) {
+			BABYLON_LOADERS.OBJFileLoader.IMPORT_VERTEX_COLORS = true;
+		}
+	}
```
pngwn, line 98: Hmm, I'm not sure this will ever rerun. I also don't actually know what this does.

@hannahblair @dawoodkhan82 any ideas?

dawoodkhan82: This imports the colors that some `.obj` files have. I think `BABYLON_LOADERS.OBJFileLoader.IMPORT_VERTEX_COLORS = true;` could go in `on_mount()`.

dawoodkhan82:
@freddyaboulton Also, unrelated to this PR but I noticed the `model3D` demo creates two upload components, instead of an output static component. Probably caused by the recent frontend file refactor.

js/model3D/interactive/Model3DUpload.svelte
freddyaboulton: I tried adding it to `onMount` but I still get the same error 😅

```js
Model3DUpload.svelte:96 Uncaught (in promise) TypeError: Cannot set properties of undefined (setting 'IMPORT_VERTEX_COLORS')
    at instance (Model3DUpload.svelte:96:64)
    at init (svelte.js:3237:5)
    at new Model3DUpload (Model3DUpload.svelte:116:28)
    at createProxiedComponent (svelte-hooks.js:341:9)
    at new ProxyComponent (proxy.js:242:7)
    at new Proxy<Model3DUpload> (proxy.js:349:11)
    at Array.create_default_slot_2 (Index.svelte:98:22)
    at create_slot (svelte.js:165:23)
    at create_dynamic_element (Block.svelte:35:23)
    at create_fragment (Block.svelte:22:11)
```

dawoodkhan82: I can't seem to reproduce this issue. Are you just running the `model3D` demo on frontend dev mode?

freddyaboulton: Running on custom components dev mode. So the following was not working for me:

```bash
gradio cc create MyModel3D --template Model3D --install
cd mymodel3d
gradio cc dev
```

dawoodkhan82 (approved):
Pushed changes to fix the `interactive` bug I noticed. Also your changes seem to work, so fine with merging!

PR #5960, opened 2023-10-17
rererefactor frontend files

pngwn:
## Description

Refactors the frontend files (yet again) to make things simpler.

Component packages have a maximum of two entry points:

- `Index.svelte` - this is the component (not optional)
- `Example.svelte` - this is the example component (optional)

By convention any other source files are store in a `/shared` folder. I have also removed all the re-exporting `index.ts` files because things were just getting confusing.

This has simplified many packages and many components.

- The layout components like `Group`, `Form` are most just a single file now.
- The form input components now have a much simplified interactive/static switching (just a prop)

This should make things much less overwhelming for custom component authors using our packages as templates.

`Image` is still an absolute disaster obviously.

Closes: #(issue)

## 🎯 PRs Should Target Issues

## Tests


js/fileexplorer/Index.svelte
```diff
@@ -0,0 +1,101 @@
...
+			{...loading_status}
+			autoscroll={gradio.autoscroll}
+			i18n={gradio.i18n}
+		/>
+
+		<DirectoryExplorer
+			bind:value
+			{file_count}
+			{server}
+			mode="interactive"
```
freddyaboulton, line 97: Do we need the if/else clause here. We can have one branch with {mode}, no?

pngwn: Yes! I missed this one.

js/uploadbutton/Index.svelte
```diff
@@ -62,20 +63,39 @@
...
+	<UploadButton
+		{elem_id}
+		{elem_classes}
+		{visible}
+		{file_count}
+		{file_types}
+		{size}
+		{scale}
+		{min_width}
+		disabled
```
freddyaboulton, line 76: I think we can condense the code with disabled={mode === "static"}?

freddyaboulton (commented):
This is great @pngwn !

The deployed demo is failing. Maybe it can't import state component?

image

Also, I noticed some of the unit test files still import from `static`, e.g. Audio. Would be good to fix for the person who will get CI green for the UI.

Other than that, should be good to merge!

pngwn:
>The deployed demo is failing. Maybe it can't import state component?

Yeah thats just a casing issue, should be fixed in the latest commit (when it deploys).

js/uploadbutton/Index.svelte
pngwn: Will check!

pngwn: fixd

pngwn:
Should be ok now @freddyaboulton

freddyaboulton (approved):

Issue #5720, opened 2023-09-27
is there any way to change the microphone component to allow clients to pause and then resume recordings?


cowanAI:
- [ ] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  

**Describe the solution you'd like**  

**Additional context**

abidlabs:
Not currently, but we are redesigning the `gr.Audio` component and I think this would be nice to include if not too much work. wdyt @hannahblair?

cowanAI:
Happy to send $300 usd right now to Hannah or to the gradio project if this
is possible, I know it’s not a lot but anything I can do to help let me
know.

abidlabs:
Well thanks @cowanAI for the generous offer, but maintaining `gradio` is part of our jobs :) 

We'll see what @hannahblair thinks when she's online. 

By the way we're working on making it possible for Gradio users to create their own custom components -- meaning that you'll be able to take an existing Gradio component and clone it, modify the backend or the frontend, and use it in your Gradio apps. This might help with these kinds of feature requests in the future!

cowanAI:
Awesome @abidlabs , extremely grateful at the gradio support, I know gradio is designed for prototype uses but my last question, do you know if in the roadmap you guys have planned to support user login and session/output storage, i.e. the user logins and has access to it's own outputs? if you had any documentation or if this is not possible it would be great to know, and any way we can support the gradio team just let us know, we are happy for your help!!

abidlabs:
Are you running on Hugging Face Spaces? We actually just released experimental support for users on HF: https://huggingface.co/docs/hub/spaces-oauth

cc @Wauplin

Wauplin:
@cowanAI definitely a use case meant to be supported yes! As @abidlabs shared, "Sign in with Hugging Face" button is already available when running on Spaces. Meaning you can handle user sessions & cie. Here is a user-history feature I have being working on: https://huggingface.co/spaces/Wauplin/gradio-user-history/blob/main/README.md

cowanAI:
@abidlabs @Wauplin wow that user history looks amazing! is there any way to instead of utilizing hugging face sign in I could use cognito or my own authentication with your user history plugin @Wauplin ? so far this is an incredible progress of the gradio interface not being a simple prototyping tool anymore!

Wauplin:
[...]

Issue #5321, opened 2023-08-24
Proposal: enhancements and redesign of `Audio`


hannahblair:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
The limitations of Audio have been a frequent point of discussion, and with the increased usage as well as 4.0 coming up, now would be a good time to address the bugs and feature requests to improve the developer and user experience around it. The key issues with it are around the user experience; the editing functionality is quite ambiguous, the slider is quite finicky to use, the design needs an update, and there are a few trimming/device bugs to tackle, but there is some additional functionality which would be useful to implement. 

**Describe the solution you'd like**  
A new Audio component, which has a new UI and additional functionality. This issue serves as a discussion around its updated implementation. 

**Functionality** 

- Trimming/Cropping
    - Selecting the edit/pencil/scissors icon will enable the trimming mode
    - The trimming mode allows you to select a segment of audio
    - If a particular segment of audio is currently selected, the play icon will play the selected segment
    - The cropped selection of audio is sent to the prediction function on submit

- Input/Output
    - Selecting a device icon should show the available I/O devices to allow users to select devices other than their system defaults. 
    - It should be clear if either the microphone or the speaker permissions are not enabled. This would help combat confusion around whether the Gradio component is actually working or not. 

- Enforce limits
    - A new param `max_length` should be implemented in `Audio.py` to allow developers to limit the length of Audio that can be processed. This is pretty valuable as developers often want to limit queue load times. 
    - We could also have a `min_length` so that developers can ensure meaningful predictions are made in their apps.
    - This could be one param passed in like `[5, 20]`. We should allow not setting one or the other like `[None, 20]`. **Question on that** - what should that be named? We currently just use `minimum` and `maximum` for `gr.Slider` but it should be clear that this is the min + max of what a user of their app can process. 
- Label regions
    - This is probably a feature for either a future iteration or a custom component. An issue has been raised (#2023) around labelling certain sections of idea which is quite an interesting one. The suggestion is that you can pass a list of tuples like `[(start_time_seconds, end_time_seconds, region_label)]`. It could be like YouTube's region labelling: 
[...]

Issue #5320, opened 2023-08-24
Allow previewing selected audio snippet in gr.Audio 


hannahblair:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  

@hysts raised the issue that there is no way of previewing the region that has been selected in the Audio component before clipping. This would be a valuable addition in the updated Audio component. 

**Describe the solution you'd like**  

**Additional context**

Issue #5299, opened 2023-08-23
Editing `gr.Audio` causes error if the value was set from the beginning


hysts:
### Describe the bug

If an audio file was given as the initial value of `gr.Audio`, editing doesn't work and an error is raised. The same error occurs when using `gr.Examples`.
The error doesn't occur if the file was drag and dropped.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

https://huggingface.co/spaces/hysts-debug/audio-edit

```python
import gradio as gr

with gr.Blocks() as demo:
    audio = gr.Audio(value="sample.mp3")
    btn = gr.Button()
    audio2 = gr.Audio()
    btn.click(fn=lambda x: x, inputs=audio, outputs=audio2)
demo.queue().launch()
```

https://huggingface.co/spaces/hysts-debug/audio-edit-2

```python
import gradio as gr

with gr.Blocks() as demo:
    audio = gr.Audio()
    btn = gr.Button()
    audio2 = gr.Audio()
    gr.Examples(examples=["sample.mp3"], inputs=audio)
    btn.click(fn=lambda x: x, inputs=audio, outputs=audio2)
demo.queue().launch()
```

### Logs

```shell
To create a public link, set `share=True` in `launch()`.
Traceback (most recent call last):
  File "/home/user/.local/lib/python3.10/site-packages/gradio/routes.py", line 488, in run_predict
    output = await app.get_blocks().process_api(
  File "/home/user/.local/lib/python3.10/site-packages/gradio/blocks.py", line 1428, in process_api
    inputs = self.preprocess_data(fn_index, inputs, state)
  File "/home/user/.local/lib/python3.10/site-packages/gradio/blocks.py", line 1245, in preprocess_data
    processed_input.append(block.preprocess(inputs[i]))
  File "/home/user/.local/lib/python3.10/site-packages/gradio/components/audio.py", line 220, in preprocess
    sample_rate, data = processing_utils.audio_from_file(
  File "/home/user/.local/lib/python3.10/site-packages/gradio/processing_utils.py", line 145, in audio_from_file
    audio = AudioSegment.from_file(filename)
  File "/home/user/.local/lib/python3.10/site-packages/pydub/audio_segment.py", line 773, in from_file
    raise CouldntDecodeError(
pydub.exceptions.CouldntDecodeError: Decoding failed. ffmpeg returned error code: 1

Output from ffmpeg/avlib:

[...]

PR #5966, opened 2023-10-17
Improve Audio Component

hannahblair:
## Description

This PR updates our Audio component. 

to do:

- [x] Visual waveform (Recordings & Uploads)
- [x] Play and pause controls
- [x] Show timestamps
- [x] Playback Speed
- [x] Download file
- [x] Select snippets
- [x] Clear audio
- [x] Recording
- [x] Timestamps while recording
- [x] Pause and resume recording
- [x] Audio device enumeration
- [x] Trim recordings
- [x] Trim uploaded audio
- [x] Streaming
- [x] WASM changes
- [x] Dispatch events  
    - [x] Events.stream
    - [x] Events.change
    - [x] Events.clear
    - [x] Events.play
    - [x] Events.stop
    - [x] Events.pause
    - [x] Events.start_recording
    - [x] Events.stop_recording
    - [x] Events.pause_recording (new - add this to backend)
    - [x] Events.upload
- [x] Tests
- [x] Update backend with new params
- [x] Update documentation
    - [x] New params (waveform settings)
    - [x] Update demos 
- [ ] Allow passing captions
- [ ] Allow passing video

Closes: #(issue)

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/test/test_components.py
+++ b/test/test_components.py
@@ -804,7 +804,8 @@ class TestAudio:
             "streaming": False,
             "show_label": True,
             "label": None,
-            "source": "upload",
+            "max_length": None,
+            "min_length": None,
             "container": True,
             "min_width": 160,
             "scale": None,
@@ -817,6 +818,8 @@ class TestAudio:
             "type": "filepath",
             "format": "wav",
             "streamable": False,
+            "sources": ["microphone", "upload"],
+            "waveform_options": None,
         }
 
         output1 = audio_output.postprocess(y_audio.name)


abidlabs:
Really nice UI/UX @hannahblair! 🔥🔥

Here's what I tested and a few issues I found:

1. Testing displaying an audio file: both interactive and static mode work great

```py
with gr.Blocks() as demo:
    gr.Audio("cantina.wav")
    gr.Audio("cantina.wav", interactive=True)
    
demo.launch()
```

Closes: #5299

2. Tested input and output `Interface`: 

```py
import gradio as gr

gr.Interface(lambda x:x, "audio", "audio").launch()
```

Works fantastic. I had one question: it looks like we duplicate the icons in the interactive Audio component: we have two icons to trim the audio, and two to reset. Is this intentional? If not, my suggestion would be to:
* remove the pencil icon altogether
* change the behavior of the "circling back" icon to reset the trim. That's what I would have expected it to do, but it actually reset the entire audio component

3. Tested trimming: the UI is great, and it processing the value well. My main question was what does the number in the orange represent? I thought it was the length of the trimmed audio, but it always was always stuck at 0:01 even when I adjusted the length of the trimmed section

image

Closes: #5320

3. Tested microphone:

```
import gradio as gr

gr.Interface(lambda x:x, "mic", "audio").launch()
```
 
* I wasn't able to get the microphone value to be sent to the backend. You can see in the example above, even if you record, and then click submit, it always sends an empty value
* Its awesome that you now have a dropdown to select the microphone sources 🚀 -- but I noticed that before you grant microphone permissions, you see an empty dropdown which looks unusual:

image

* Love that you can pause and resume recordings. Nit: should we add an icon before resume here? Nit: should we add an icon before the audio sources? 

image

Closes: #4987, Closes: #5720

4. Tested setting min and max length:

```
import gradio as gr

gr.Interface(lambda x:x, gr.Audio(max_length=1, source="upload"), "audio").launch()
```

* Works great -- enforces it on the backend, which is ideal
* I wonder if we should also enforce it on the frontend? Like if someone uploads an audio file that is too long, perhaps they should know immediately so that they can trim it to get it into the right range. 

Closes: #428

Really awesome redesign @hannahblair!

Closes: #467, Closes: #5321

abidlabs:
As discussed in sync, it would be really nice if we could allow source to be a list `["microphone", "upload"]`, which would allow a user to choose either a microphone or upload a file (and close a long-standing issue: #3099). On the other hand, I love how clean this UI is and maybe that would clutter things up too much. Let's see how @pngwn implements it for the Image component and if it makes sense to apply here as well

gradio/components/audio.py
```diff
@@ -102,13 +105,11 @@ def __init__(
...
+            max_length: The maximum length of audio (in seconds) that the user can pass into the prediction function. If None, there is no maximum length.
+            waveform_options: A dictionary of options for the waveform display. Options include: waveform_color (str), waveform_progress_color (str), show_controls (bool). Default is None, which uses the default values for these options.
         """
-        valid_sources = ["upload", "microphone"]
-        source = source if source else ("microphone" if streaming else "upload")
-        if source not in valid_sources:
-            raise ValueError(
-                f"Invalid value for parameter `source`: {source}. Please choose from one of: {valid_sources}"
-            )
+        source = list(set(source)) if source is not None else (["microphone"] if streaming else ["microphone", "upload"])
```
abidlabs, line 112: I like that you convert strings to lists, but just wondering, do you need the `set` here? Or could you just do:
```suggestion
        source = list(source) if source is not None else (["microphone"] if streaming else ["microphone", "upload"])
```

abidlabs: Ideally if a user accidentally passes in a string (e.g. "upload"), it should treat it as the same as ["upload"], which `list(source)` does, but `list(set(source))` does not

abidlabs:
Gave this another spin, love the joint upload/mic functionality. It seems that most of my original issues have been addressed, thank you!

> It is indeed the length of the trimmed audio - your audio there is fairly short at 0:03, so your selection duration is rounded to the nearest second and we haven't got any more granularity in that value

I tried with a larger audio clip, and I'm still seeing this behavior. The second duration remains fixed, even if I change the width of the trimming window. You can see here, the trimming window is covering most of the width of the 11 second clip, but its only showing 3 seconds as the duration in orange:

image

The color of the "selected" source could be improved in light mode. Right now it becomes white, which is very hard to see on the white background: 

image

(it looks great in dark mode btw).

Very excited for this 🚀🚀

pngwn:
Everything looks great from my limited testing. The only thing I did notice was that certain things didn't work as expected on mobile. Could you just test the buttons + trimming behaviour on mobile to make sure everything works as expected? We might need to scatter in some touch events.

gradio/components/audio.py
```diff
@@ -57,7 +58,7 @@ def __init__(
         self,
         value: str | Path | tuple[int, np.ndarray] | Callable | None = None,
         *,
-        source: Literal["upload", "microphone"] | None = None,
+        source: list[Literal["upload", "microphone"]] | None = None,
```
pngwn, line 61: Could we rename this to `sources` to match the image component?

abidlabs: Let's keep note of this as a deprecation

gradio/components/audio.py
```diff
@@ -102,24 +108,26 @@ def __init__(
...
-            raise ValueError(
-                f"Invalid value for parameter `source`: {source}. Please choose from one of: {valid_sources}"
-            )
-        self.source = source
+        sources = (
+            list(sources)
+            if sources is not None
+            else (["microphone"] if streaming else ["microphone", "upload"])
+        )
+        self.sources = sources
```
abidlabs, line 120: nit: slightly more concise
```suggestion
        self.sources = (
            list(sources)
            if sources is not None
            else (["microphone"] if streaming else ["microphone", "upload"])
        )
```

gradio/events.py
```diff
@@ -464,6 +464,10 @@ class Events:
         "start_recording",
         doc="This listener is triggered when the user starts recording with the {{ component }}.",
     )
+    pause_recording = EventListener(
```
abidlabs, line 467: Cool addition

gradio/components/audio.py
```diff
@@ -102,24 +108,26 @@ def __init__(
...
-        if source not in valid_sources:
-            raise ValueError(
-                f"Invalid value for parameter `source`: {source}. Please choose from one of: {valid_sources}"
-            )
-        self.source = source
+        sources = (
+            list(sources)
+            if sources is not None
+            else (["microphone"] if streaming else ["microphone", "upload"])
+        )
```
abidlabs, line 119: nit: handle the str case as well for better DX
```suggestion
        if sources is None:
            sources = ["microphone"] if streaming else ["microphone", "upload"]
        if isinstance(sources, str):
            sources = [sources]
        elif isinstance(sources, list):
            pass
        else:
            raise ValueError('`sources` must a list consisting of elements in ["microphone", "upload"]')
```

abidlabs (approved):
LGTM @hannahblair so awesome!

gradio/templates.py
```diff
@@ -385,7 +385,7 @@ def __init__(
         self,
         value: str | tuple[int, np.ndarray] | Callable | None = None,
         *,
-        source: Literal["microphone"] = "microphone",
+        sources: list[Literal["upload", "microphone"]] | None = None,
```
abidlabs, line 388: T
```suggestion
        sources: list[Literal["upload", "microphone"]] | None = ["microphone"],
```

Issue #5830, opened 2023-10-06
Clean up the code related to `root` and `root_url`


abidlabs:
All the code to resolve URL paths correctly with `root` and `root_url` has gotten quite messy. 

It would be good to clean this up. In particular, now that replicas [use subdirectories instead of subdomains](https://huggingface.slack.com/archives/C048K60MPNF/p1696569642166729), we can revert most of the changes we made in PRs #5776 and instead just use the same logic we use for the regular fastapi root path.

For clarity, we should also rename `root_url` to `remote_url` or something.

HeetVekariya:
I would like to work on this issue.
As i am new to this repo, can you please locate the files that is to be refactored ?

abidlabs:
Hi @HeetVekariya this isn't an issue I'd suggest if you are new to the repo, as there is some internal discussion and context that is important. Instead, I'd suggest picking one of the issues from this list: https://github.com/gradio-app/gradio/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22

abidlabs:
Let's rename `root_url` to `_remote_url` to designate it as an internal parameter in the component classes

abidlabs:
Also see [this internal discussion](https://huggingface.slack.com/archives/C02SPHC1KD1/p1697823622462199?thread_ts=1697724198.460679&cid=C02SPHC1KD1)

abidlabs:
Closed on `v4`!

PR #5968, opened 2023-10-17
Removes deprecated arguments and parameters from v4

abidlabs:
Removes deprecated arguments and parameters from v4 and small cleanups

For the purposes of documentation, these are the breaking changes related to components:

* Removes `**kwarg` from every component, meaning that components cannot accept arbitrary (unused) parameters. Previously, warnings would be thrown.
* Removes deprecated parameters. For example, `plain` is no longer an alias for `secondary` for the `variant` argument in the `gr.Button` class
* Removes the deprecated `Carousel` class and `StatusTracker` class and `Box` layout class
* Removes the deprecated `Variable` alias for `State`
* Removes the deprecated `.style()` methods from component classes
* Removes the deprecated `.update()` method from component classes
* Removes `get_interpretation_neighbors()` and `get_interpretation_scores()` from component classes
* Removes `deprecation.py` -- this was designed for internal usage so unlikely to break gradio apps

These are the breaking changes related to other parts of Gradio

* Removes the deprecated `status_tracker` parameter from events
* Removes the deprecated `HuggingFaceDatasetJSONSaver` class
* Now `Blocks.load()` can only be use an is instance method to attach an event that runs when the page loads. To use the class method, use `gr.load()` instead
* Similarly, `Interface.load()` has been removed
* Removes deprecated parameters, such as `enable_queue` from `launch()`

Non-breaking changes:
* Some warnings have been changed from `GradioDeprecationWarning` to `UserWarning` (they were previously incorrectly-labeled)
* Tests relating to `.update()` and other deprecated methods have been removed


freddyaboulton:
Nice! Will take a closer look tomorrow!

abidlabs:
> Nice! Will take a closer look tomorrow!

Thanks @freddyaboulton. Should be finally ready for a look

gradio/components/annotated_image.py
```diff
@@ -64,7 +63,9 @@ def __init__(
         visible: bool = True,
         elem_id: str | None = None,
         elem_classes: list[str] | str | None = None,
-        **kwargs,
+        render: bool = True,
+        root_url: str | None = None,
```
freddyaboulton, line 67: So someone can update the `root_url` from an event? 😅

freddyaboulton: I guess this can happen already?

abidlabs: Haha yea I can't think of any reason anyone would want to but yes this is now possible. We need to include this in the constructor since we no longer accept arbitrary *kwarg params, and this is set in `Blocks.from_config`. If we want, we could rename this to be an internal variable `_root_url` (or even better `_remote_url`) but that can be a separate PR I think

freddyaboulton (approved):
Love the clean up @abidlabs !!! 🧹

gradio/components/annotated_image.py
freddyaboulton: I think treating it like _skip_init_processing makes sense (also related to spaces)

abidlabs: Sounds good -- I've added as a todo here: https://github.com/gradio-app/gradio/issues/5830

gradio/components/markdown.py
```diff
@@ -31,23 +30,33 @@ def __init__(
...
+        root_url: str | None = None,
+        _skip_init_processing: bool = False,
         sanitize_html: bool = True,
         line_breaks: bool = False,
-        **kwargs,
     ):
         """
         Parameters:
             value: Value to show in Markdown component. If callable, the function will be called whenever the app loads to set the initial value of the component.
+            label: The name of component.
```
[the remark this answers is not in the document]
abidlabs: `label` is also used internally for a few things, such as if you the header of the tables when you have examples. So let's keep it, but I'll improve the docstring

abidlabs:
Thanks for the review @freddyaboulton! I'll fix the docstrings a bit and then merge

PR #5985, opened 2023-10-18
Relax pin on typing_extensions

freddyaboulton:
## Description

CI was breaking because new FastAPI release needs typing_extensions>4.8.0 but we had version 4.7.1. We should not be pinning in this case since we deliberately have a relaxed pin for typing_extensions in our requirements.txt to avoid this kind of situation.

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/test/requirements.txt
+++ b/test/requirements.txt
@@ -238,13 +238,6 @@ traitlets==5.3.0
     #   matplotlib-inline
 transformers==4.20.1
     # via -r requirements.in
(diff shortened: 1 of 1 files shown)


freddyaboulton:
I'm making the same change in #5970 for v4

abidlabs:
Agreed.

abidlabs (approved):

Issue #5564, opened 2023-09-14
4.0 beta tracking + feedback


freddyaboulton:
Place to track all of the bugs, nits, and comments found in the 4.0 branch (`v4`)

# Pre launch
- [x] Get the CI green ✅ 
- [x] python template will replace from PIL import Image with from PIL import <custom-component-name>
- [x] js server launching commands should use full path to executables, e.g. gradio, python, node
- [x] FileExplorer not template_able
- [x] Unpin version before release for docs
- [x] Create simple templates for form components
- [x] Before Release - Set gradio>=4.0,<5.0 in pyproject.toml file for custom components
- [x] https://github.com/gradio-app/gradio/issues/5952
- [x] Add a tag to pyprojecttoml to identify a package as a custom component package
- [x] Front-end dev server will use a port that's used by another app
- [x] Use the data_model in the preprocess
- [x] Model3D component can't be used in dev mode
- [x] Overwrite should clear out the previous directory
- [x] Windows build in CI does not work (frontend)
- [x] Add a server to get backwards compatible type info
- [x] Fix docstrings for events
- [x] Fix website build in v4
- [x] #6223
- [x] ImageMask not working
- [x] ImagePaint not working
- [x] gr.Image(source='upload', tool='color-sketch') not working 
- [x] Tidy up temp file creation in pre/postprocess (particularly Image).
- [x] Empty JSON component looks weird in kitchen_sink_random when cleared ![image](https://github.com/gradio-app/gradio/assets/41651716/2171b46b-faeb-474e-917d-565c7c22b0e4)
- [x] In the external.py module, base64 is being returned as opposed to a temp file for audio/image etc. UI works but caching examples doesn't.
- [x] Interface/Blocks `__call__` method is still calling serialize so it doesn't work
- [x] Have the Fallback component display plain text/number values
- [x] Reduce size of node folder before shipping
- [x] Fix Literal type hint in models
- [x] `gradio_component build` builds the front-end
- [x] Make sure all the dependencies are absolute and not relative in front-end modules
- [x] Layout components can't be templated

# Nice to have
- [x] Create simple templates for Media components
- [x] #6221
- [x] #6220
- [x] #6224
- [x] Emit warning when template directories are overwritten in `gradio cc build` (post launch)
- [x] Write (more) unit tests for gradio_component cli
- [x] Example Css for custom components: https://github.com/gradio-app/gradio/pull/5990/files#r1364407102

# Definitely post launch
- [x] #6222

[...]

PR #5990, opened 2023-10-18
V4: Simple textbox

freddyaboulton:
## Description

Introduces SimpleTextbox. More to come.

Don't think we should introduce an extra CLI parameter to distinguish between the "simple" and default component templates.

I think 

```bash
gradio cc create MyTextbox --template SimpleTextbox --install
```

Is better than

```bash
gradio cc create MyTextbox --template Textbox --simple --install
```

## 🎯 PRs Should Target Issues

## Tests


js/simpletextbox/Example.svelte
```diff
@@ -0,0 +1,46 @@
...
+	let el: HTMLDivElement;
+
+	function set_styles(element: HTMLElement, el_width: number): void {
+		if (!element || !el_width) return;
+		el.style.setProperty(
+			"--local-text-width",
+			`${el_width < 150 ? el_width : 200}px`
+		);
+		el.style.whiteSpace = "unset";
+	}
```
pngwn, line 18: Hmm, forgot about this. Maybe we can figure out a way to move it into the Dataset.

gradio/_simple_templates/simpletextbox.py
```diff
@@ -0,0 +1,100 @@
...
+    """
+
+    EVENTS = [
+        Events.change,
+        Events.input,
+        Events.select,
+        Events.submit,
+        Events.focus,
+        Events.blur,
+    ]
```
pngwn, line 26: Could we slim this down? We could maybe get rid of a few events and just have 2/3 submit/ change related ones?

gradio/_simple_templates/simpletextbox.py
```diff
@@ -0,0 +1,100 @@
...
+        label: str | None = None,
+        every: float | None = None,
+        show_label: bool | None = None,
+        scale: int | None = None,
+        min_width: int = 160,
+        interactive: bool | None = None,
+        visible: bool = True,
+        rtl: bool = False,
+        elem_id: str | None = None,
+        elem_classes: list[str] | str | None = None,
```
pngwn, line 42: I guess these are all pretty standard for all components (or essential)

js/simpletextbox/Index.svelte
```diff
@@ -0,0 +1,167 @@
...
+		const target: HTMLTextAreaElement | HTMLInputElement = event.target as
+			| HTMLTextAreaElement
+			| HTMLInputElement;
+		const text = target.value;
+		const index: [number, number] = [
+			target.selectionStart as number,
+			target.selectionEnd as number,
+		];
+		gradio.dispatch("select", { value: text.substring(...index), index: index });
+	}
```
pngwn, line 54: i think we could definitely get rid of the select event. I don't think it used very much and this adds a lot of noise to the file.

js/simpletextbox/Index.svelte
```diff
@@ -0,0 +1,167 @@
...
+			class="scroll-hide"
+			bind:value
+			bind:this={el}
+			{placeholder}
+			disabled={mode === "static"}
+			dir={rtl ? "rtl" : "ltr"}
+			on:keypress={handle_keypress}
+			on:blur={() => gradio.dispatch("blur")}
+			on:select={handle_select}
+			on:focus={() => gradio.dispatch("focus")}
```
pngwn, line 103: As above i think we could get rid of some of these and it would help keep things streamlined.

pngwn (commented):
Left a few comments, to see if we could strip things down further but looks good in general. Much easier to follow. I wonder if there is a generic way we could get rid of some of the noise of all of those props, at least in the svelte components.

freddyaboulton:
Thanks @pngwn. You de best 🙏 Will slim down the events. Agreed we don't need them all. Not sure what to do about the example comment. Will make a note in #5564 so we don't forget.

gradio/_simple_templates/simpletextbox.py
freddyaboulton: Yea the only ones that are "textbox-specific" here are `rtl` and `placeholder`. All the other ones are pretty standard for it to look like the other gradio components.

gradio/_simple_templates/simpletextbox.py
freddyaboulton: Word. Keeping change, input, and submit!

abidlabs:
> Don't think we should introduce an extra CLI parameter to distinguish between the "simple" and default component templates. I think

> gradio cc create MyTextbox --template SimpleTextbox --install

> is better than

> gradio cc create MyTextbox --template Textbox --simple --install

Agreed, I find it more intuitive

pngwn:
Agree with the comment about the CLI argument. 

Regarding the comment on the example, feel free to disregard for this PR. More of a note for myself / thinking out loud. 

I'd actually like to remove the gallery/ table distinction from examples completely. It is a lot of repetition and I think it can be handled in dataset.

js/simpletextbox/Index.svelte
```diff
@@ -0,0 +1,129 @@
...
+	export let value_is_output = false;
+	export let mode: "static" | "interactive";
+	export let rtl = false;
+
+
+	let el: HTMLTextAreaElement | HTMLInputElement;
+	const container = true;
+
+	function handle_change(): void {
+		gradio.dispatch("change", value);
```
abidlabs, line 35: Why do we dispatch the value for "change" but not for the other events? Can we do:

```
gradio.dispatch("change");
```

gradio/_simple_templates/simpletextbox.py
```diff
@@ -0,0 +1,95 @@
...
+
+from typing import Any, Callable
+
+from gradio.components.base import FormComponent
+from gradio.events import Events
+
+
+class SimpleTextbox(FormComponent):
+    """
+    Creates a textarea for user to enter string input or display string output.
```
abidlabs, line 11: Since we don't support multiple lines anymore, it would be good to replace "textrea" with "textbox" throughout the docstring

js/simpletextbox/Index.svelte
```diff
@@ -0,0 +1,129 @@
...
+		await tick();
+		if (e.key === "Enter") {
+			e.preventDefault();
+			gradio.dispatch("submit");
+		}
+	}
+
+	$: if (value === null) value = "";
+
+	$: value, handle_change();
```
abidlabs, line 51: I always found this syntax a bit confusing. I don't know if there's a way to rewrite more clearly in svelte. Maybe its worth an explanatory comment.

freddyaboulton: We want `handle_change` to run when the value changes so I guess we can do

```ts
$:   handle_change(value)
```

Do you have thoughts on this @pngwn ?

js/simpletextbox/Index.svelte
pngwn: Yeah. I don't think we use this value at all.

js/simpletextbox/Index.svelte
pngwn: Yep. That's a better way to do it if it makes sense for the function we are calling.

freddyaboulton: Ok will push up. Verified it works!

abidlabs: But if we make the change above, handle_change() doesn't utilize the value at all

abidlabs: This change: 

image

freddyaboulton: I think that's ok?

abidlabs: It won't cause any issues but it doesn't seem like good practice to pass a value into the function that it won't use. I personally think adding a comment for clarity to the original syntax, but its okay either way.

freddyaboulton: let's use a comment!

freddyaboulton:
Thanks for the reviews! Stay tuned for `SimpleDropdown` 😎

js/simpletextbox/Index.svelte
pngwn: It would also be a type or linting error, so comment works well.

Issue #5461, opened 2023-09-08
`@gradio/client` options parameter is not optional


pngwn:
### Describe the bug

> When you use gradio client in ts, const gradioApp = await client("https://mishig-embeddings-similarity.hf.space/")  (which is what’s suggested in the docs), gives you an error An argument for 'options' was not provided. So I had to provide empty options const gradioApp = await client("https://mishig-embeddings-similarity.hf.space/", {})

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

When you use gradio client in ts, const gradioApp = await client("https://mishig-embeddings-similarity.hf.space/")  (which is what’s suggested in the docs), gives you an error An argument for 'options' was not provided. So I had to provide empty options const gradioApp = await client("https://mishig-embeddings-similarity.hf.space/", {})

Need to make this optional.

### Screenshot

-

### Logs

```shell
-
```

### System Info

```shell
-
```

### Severity

I can work around it

shwethabm:
Can I work on this issue?

abidlabs:
Go for it @shwethabm! Looking forward to reviewing your PR

PR #5997, opened 2023-10-19
make options parameter optional in client

aileenvl:
## Description

Make options parameter optional in client 

Closes: #5461

### DIFF
--- a/.changeset/loud-birds-crash.md
+++ b/.changeset/loud-birds-crash.md
@@ -1,0 +1,6 @@
+---
+"@gradio/client": patch
+"gradio": patch
+---
+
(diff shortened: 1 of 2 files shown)


abidlabs:
Hi @aileenvl thanks for contributing this PR. We are actually going to merge the v4 branch into `main` soon which has considerable changes from main. Which means that all PRs that are currently based off of `main` will have significant merge conflicts, and it would be much easier to actually just create a new PR based off the of `v4` branch instead. 

I'll go ahead and close this PR, but if you can create a new one setting v4 as the base branch, that would be much better. Thank you!

PR #6011, opened 2023-10-19
fix: allow multiple flagging directories

EgonFerri:
## Description

Solves the issue of a bug impeding linking different flagging directories to different interfaces in the case of multiple interfaces like gr.TabbedInterface, as discussed with @abidlabs and @harry-urek in pull request [5928](https://github.com/gradio-app/gradio/pull/5928), and in issue #5692.

The cause of the bug was residing in the fact that even if each interface, in theory, should have its proper FlaggingCallback, which in theory should have its own `self.flagging_dir`, both FlaggingCallaback pointed to the same `CSVLogger()` because the argument flagging_callback of `Interface` class in `interface.py` created a FlaggingCallback that pointed to the `CSVLogger()` class and not, as it should, provide to create instance of this class.

 I fixed the problem by changing the default value of flagging_callback to None, and moving the flagging_callback build  in the constructor of Interface like:

```
    if flagging_callback is None:
```

## 🎯 PRs Should Target Issues

Closes: #5692

## Tests

1. I followed the guidelines to contribute (https://github.com/gradio-app/gradio/blob/main/CONTRIBUTING.md) and ran the tests: `bash scripts/run_all_tests.sh`. All tests passed except for some tests linked to hugging face that were not related to the addressed issue and were already broken before my changes (for unloadable spaces or hugginface API limit reach).

### DIFF
--- a/gradio/interface.py
+++ b/gradio/interface.py
@@ -142,7 +142,7 @@ class Interface(Blocks):
         allow_flagging: str | None = None,
         flagging_options: list[str] | list[tuple[str, str]] | None = None,
         flagging_dir: str = "flagged",
-        flagging_callback: FlaggingCallback = CSVLogger(),
+        flagging_callback: FlaggingCallback | None = None,
         analytics_enabled: bool | None = None,
         batch: bool = False,
         max_batch_size: int = 4,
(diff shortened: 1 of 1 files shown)


abidlabs:
LGTM thanks so much for fixing this @EgonFerri!

abidlabs:
Just wanted to thank you again for taking the time to fix this issue! We're also sharing some Hugging Face swag with top contributors during Hacktoberfest and I'd like to include you!

Can you send an email to [team@gradio.app](mailto:team@gradio.app) and we'll go from there? cc @hannahblair

EgonFerri:
I'm delighted to see we merged :)
In the end, it was an easy fix, but I'm glad I had the opportunity to look inside a library that I've been using for years and drastically helped me on multiple occasions.

@abidlabs Super thanks for everything, including the lightning-fast support and the nudges in the right direction! 
I'll write the e-mail (: have a nice day!

PR #6018, opened 2023-10-19
Add a cli command to list available templates

freddyaboulton:
## Description

Add a cli command to list all the available templates with some nice metadata about the kind of component.

gradio_cc_show

## 🎯 PRs Should Target Issues

## Tests


abidlabs:
Love it @freddyaboulton, particularly the "Beginner Friendly" tags. 

The only thing is that when all of the components are printed, you miss the simple components since they go all the way to the top. Is it possible to print only as many items that can fit on the terminal screen and then a user has to press the enter key to see more? If not, then maybe we put the "Beginner Friendly" ones at the bottom?

freddyaboulton:
Great idea @abidlabs ! It was really easy to add pagination actually. 

gradio_cc_show_paginate

There's a note in the rich docs about how some shells may not support colors with pagination, so I added a parameter `--color` or `--no-color` to let people disable them if something goes wrong.

abidlabs:
Thanks @freddyaboulton! Pagination looks great. I'm seeing this on the VSCode terminal, which I guess doesn't show colors: 

image

If I do `--no-colors`, it looks much better:

image

But the default `gradio cc show` command output on the top looks broken. Is there any way to detect whether the terminal can support colors and just show the appropriate output?

freddyaboulton:
No easy way from what I can tell. Let's remove the colors and keep pagination.

freddyaboulton:
Added some emojis to make them "pop" a bit

abidlabs:
Beautiful! Thanks for updating the PR @freddyaboulton

abidlabs (approved):

freddyaboulton:
Thank you for the helpful feedback!!

PR #6026, opened 2023-10-20
V4: Single-file implementation of form components

freddyaboulton:
## Description

I think the js code for our form components is straightforward but after discussing with @pngwn we can make it even easier for people by having the entire implementation in the Index.svelte file. Using a single file will mean that developers only have to add/remove props in one file as opposed to two. 

For checkboxgroup, I rewrote the check that triggers the change event to not use the JSON.stringify. Thought it was weird.

## 🎯 PRs Should Target Issues

## Tests


gradio/components/video.py
```diff
@@ -52,6 +52,7 @@ class Video(Component):
         Events.play,
         Events.pause,
         Events.end,
+        Events.upload,
```
freddyaboulton, line 55: Tiny oversight on my part

freddyaboulton:
@pngwn I refactored Checkbox and Radio into the `BaseCheckbox` and `BaseRadio` pattern. This should be good for a review.

pngwn (approved):
Looks great! Thanks for taking care of these changes @freddyaboulton!

freddyaboulton:
Thank you @pngwn ! 🙏

PR #6027, opened 2023-10-20
V4: Fix component update bug 

freddyaboulton:
## Description

`elem_classes` was being stored as `None` in the `constructor_args` property of the `Block`. When you returned a new component in an event, `elem_classes=None` was being passed to the frontend app, which caused it to break since it was expecting a list.

You can see the bug by running `fake_diffusion_with_gif` on `v4`. After you click the submit button, it will crash after the first update.

## 🎯 PRs Should Target Issues

## Tests


abidlabs:
Tested and works! However, I would have expected this to cause an issue if an output component initially has some `elem_classes` but then it is updated by returning a new component. I would have expected the custom `elem_classes` to be gone. However, this does not appear to be the case, do you know why?

abidlabs (approved):

freddyaboulton:
@abidlabs We update the props that are sent to the front-end based on the props set during the event. So if elem_classes is not updated, `[]` gets sent to the front-end. If it does, the updated value gets sent. This is the same behavior on main, haven't changed that in v4 branch.

freddyaboulton:
Thanks for the review!!

abidlabs:
Ah I see thanks @freddyaboulton!

Issue #5932, opened 2023-10-16
gr.Launch(auth = ) UI not correct in code deployed on huggingface.com URL


dwipper:
### Describe the bug

I implemented the gr.Launch(auth = ) with the following code:

iface.launch(auth=login_auth, auth_message= "Enter your username and password that you received from CIMS.AI. To request a login, please email 'info@cims.ai'",share=True)

While the login form displays correct on the local browser, when deployed on a huggingface.com URL, the styling is missing and the login button doesn't work.

Please see the attached screen shots.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

https://huggingface.co/spaces/dwipper/NILI

### Screenshot

![image](https://github.com/gradio-app/gradio/assets/140648761/fcaf4012-8135-4e99-a852-30c7d891223d)

![image](https://github.com/gradio-app/gradio/assets/140648761/7efd4d33-baab-4cca-a75c-9a327fa49f34)

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Darwin
gradio version: 3.47.1
gradio_client version: 0.6.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.1.1
fastapi: 0.103.2
ffmpy: 0.3.1
gradio-client==0.6.0 is not installed.
httpx: 0.25.0
huggingface-hub: 0.16.4
importlib-resources: 6.1.0
jinja2: 3.1.2
markupsafe: 2.1.3
matplotlib: 3.7.2
numpy: 1.25.2
orjson: 3.9.7
packaging: 23.1
pandas: 2.0.3
pillow: 10.0.0
pydantic: 1.10.13
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0.1
requests: 2.31.0
semantic-version: 2.10.0
typing-extensions: 4.5.0
uvicorn: 0.23.2
websockets: 11.0.3
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2023.9.2
httpx: 0.25.0
huggingface-hub: 0.16.4
packaging: 23.1
requests: 2.31.0
typing-extensions: 4.5.0
websockets: 11.0.3
```

### Severity

Blocking usage of gradio

abidlabs:
Super strange! We'll look into it thanks @dwipper

[...]

PR #6042, opened 2023-10-20
Fix `root` when user is unauthenticated so that login page appears correctly

abidlabs:
Sorry for targeting `main` but this is a pretty bad bug -- the login page css is [broken on Spaces](https://huggingface.co/spaces/abidlabs/test-auth):

image

This PR fixes it. Verified by installing this PR on a different Space: https://huggingface.co/spaces/abidlabs/test-auth2

Fixes: #5932

### DIFF
--- a/.changeset/puny-teeth-notice.md
+++ b/.changeset/puny-teeth-notice.md
@@ -1,0 +1,5 @@
+---
+"gradio": patch
+---
+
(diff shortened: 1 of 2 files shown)


freddyaboulton (approved):

abidlabs:
Thanks @freddyaboulton!

PR #6058, opened 2023-10-23
Use str instead of repr on Error messages

normster:
## Description

I believe it the `__str__` method of `gradio.Error` should call `str` instead of `repr`, otherwise newlines and whitespace does not display correctly in the error modal in browser.

### DIFF
--- a/.changeset/floppy-taxes-rule.md
+++ b/.changeset/floppy-taxes-rule.md
@@ -1,0 +1,5 @@
+---
+"gradio": minor
+---
+
+feat:Use str instead of repr on Error messages
--- a/gradio/exceptions.py
+++ b/gradio/exceptions.py
@@ -80,4 +80,4 @@ class Error(Exception):
         super().__init__(self.message)
 
     def __str__(self):
-        return repr(self.message)
+        return str(self.message)


abidlabs:
Hi @normster thanks for contributing? Can you provide a self-contained example showing the issue with using `repr`?

normster:
```
import gradio as gr


def raise_error():
    raise gr.Error("This should be line 1\nAnd this should be line 2\nAnd finally line 3 which goes on and on and on and on and on and on and on and on and on and on and on and on and on and on and on and on")


def main():
    with gr.Blocks() as block:
        button = gr.Button("Raise error")
        button.click(raise_error)
    block.launch(share=True)


if __name__ == "__main__":
    main()

```

Which results in this malformed error message: image

abidlabs:
Running the same code on this branch, I don't see the problem resolved:

image

normster:
That's strange, it's wrong in a different way. "And this should be line 2" is displayed on a separate line.

I was using `3.42.0` and just updated to `3.50.2`. Both version yield the same result, shown in my screenshot.

abidlabs:
I don't see any screenshot @normster

normster:
Ah sorry, I was referring to the current behavior in gradio. Yes, you're right that building from this PR doesn't seem to work with the demo. Previously, I was testing with another gradio app where it does seem to fix the issue.

Do you have any suggestions on how to handle the newlines correctly? In any case, I think `str` is still more correct than `repr`.

freddyaboulton:
The error response is JSON-encoded so `\n` gets turned to `\\n`. My guess is the javascript is not parsing that correctly.

abidlabs:
Hi @normster I'd suggest we close this PR and instead open an issue which describes this clearly. Then, if you'd like, you can open a PR to fix this issue altogether. Also, I should note that it would be better to start off working on the `v4` branch as we will merge that into `main` soon, and it has considerable changes from `main`.

Issue #3661, opened 2023-03-28
More intuitive preprocessing for `File` component


abidlabs:
Right now, the `gr.File` component preprocessing produces a `tempfile._TemporaryFileWrapper`. This is unintuitive for several reasons:

* Users are unfamiliar with this class (it is a private class after all)
* Users expect `gr.File` to return a filepath like the other components, as well as this is what the `gr.File` expects as output

See #3658 or [this tweet](https://twitter.com/ducha_aiki/status/1642951961233477632) for an example of confusion caused by this. Users always just do `x.name` to get the filename anyways. So in 4.0, I suggest we change the preprocessing of `gr.File` to return a filepath instead of a `tempfile._TemporaryFileWrapper` object. It would be really cool if we could override the string class so that `x.name` also works for backwards compatibility

abidlabs:
Closed in `v4` branch

PR #6060, opened 2023-10-23
Clean up backend of `File` and `UploadButton` and change the return type of `preprocess()` from TemporaryFIle to string filepath

abidlabs:
We've discussed previously about changing the type of the value returned by #3661 from NamedTemporaryFile to string filepath as its more intuitive for users. Made that change here in a somewhat backwards-compatible way. 

Test with:

```py
import gradio as gr

gr.Interface(lambda x:x.name, "file", "file").launch()
```

and 

```py
import gradio as gr

gr.Interface(lambda x:x, "file", "file").launch()
```

(both should work)

same changes for `gr.UploadButton`, so:

```py
import gradio as gr

gr.Interface(lambda x:x.name, "uploadbutton", "file").launch()
```

and

```py
import gradio as gr

gr.Interface(lambda x:x, "uploadbutton", "file").launch()
```

should work. Also did some other cleanup. I refactored `_preprocess_single_file` into an instance method instead of a static method because the code reads cleaner and you don't have to jump between two components (helpful esp. if you're created custom components), though it does lead to a slight increase in code duplication. 

Also I changed the acceptable values of the of the `type` of both components from `file` to `filepath` to be consistent with other components. 

Closes: https://github.com/gradio-app/gradio/issues/3661


gradio/utils.py
```diff
@@ -959,3 +959,19 @@ def recover_kwargs(config: dict, additional_keys_to_ignore: list[str] | None = N
...
+    a value from the `.preprocess()` methods of the File and UploadButton components. Before Gradio 4.0, these methods returned a file
+    object which was then converted to a string filepath using the `.name` attribute. In Gradio 4.0, these methods now return a str
+    filepath directly, but to maintain backwards compatibility, we use this class instead of a regular str.
+    """
+
+    def __new__(cls, *args):
+        instance = str.__new__(cls, *args)
+        return instance
+
+    def __init__(self, *args):
```
freddyaboulton, line 976: I think we can just add the `name` parameter explicitly and attach it to self. Probably a good idea to call `str`'s init too?

```python
    def __init__(self, name: str | None = None):
        super().__init__()
        self.name = name if name else ""
```

freddyaboulton (approved):
Nice @abidlabs !

gradio/components/upload_button.py
```diff
@@ -115,14 +124,31 @@ def example_inputs(self) -> Any:
...
             ]
 
+    def _process_single_file(self, f: dict[str, Any]) -> bytes | str:
+        file_name, data, is_file = (
+            f["name"],
+            f["data"],
+            f.get("is_file", False),
+        )
+        if self.type == "filepath":
+            file = tempfile.NamedTemporaryFile(delete=False, dir=self.GRADIO_CACHE)
```
freddyaboulton, line 134: Actually this isn't really backwards compatible because you can't really access the file anymore. Like what if a demo is using `file.read()` or something? Maybe we should just get rid of the file concept altogether and return the string?

gradio/utils.py
abidlabs: Ok this works! Much simpler, thanks

abidlabs:
Made the change you suggested to `NamedString` @freddyaboulton, thanks. Will go ahead and merge once tests pass

freddyaboulton:
Nice - thanks @abidlabs !

Issue #5504, opened 2023-09-12
Can't use `str` builtin function as `fn` for `gradio.Interface`


vadimkantorov:
### Describe the bug

```shell
# want to check what shape the Audio returns for mono and stereo inputs
python3 -c 'import gradio; gradio.Interface(str, gradio.Audio(), gradio.Textbox()).launch()'
# File "/home/vadimkantorov/.local/lib/python3.8/site-packages/gradio/interface.py", line 376, in __init__
#    param_names = inspect.getfullargspec(self.fn)[0]
#  File "/usr/lib/python3.8/inspect.py", line 1144, in getfullargspec
#    raise TypeError('unsupported callable') from ex
#TypeError: unsupported callable
```

```python
# equivalent, but works
import gradio
def f(audio):
    return str(audio)
gradio.Interface(f, gradio.Audio(), gradio.Textbox()).launch()
```

`gradio.Interface` [docs](https://www.gradio.app/docs/interface) say that the `fn`'s type is `Callable`, so the error is quite ironic :) 
<img width="629" alt="image" src="https://github.com/gradio-app/gradio/assets/1041752/57a6c1a4-3e3f-4b06-bad4-c1258455bdfc">

Overall, this is minor and currently boils down to python's inspect.getfullargspec strangely not supporting many builtin functions: https://github.com/python/cpython/issues/76533

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

python3 -c 'import gradio; gradio.Interface(str, gradio.Audio(), gradio.Textbox()).launch()'

### System Info

```shell
`gradio.__version__`: '3.39.0'
```

### Severity

I can work around it

vadimkantorov:
Btw, it would be good if https://www.gradio.app/docs/audio directly specified what are shapes for mono and stereo. Currently it's (T,) for mono and (T, 2) for stereo (i.e. ndim is not equal). For many applications, it's a bit easier to have more homogeneously (T, 1) and (T, 2). And also being able to specify channel constraints on gradio.Audio (e.g. `.mono = True`, `.stereo = True`, sth like that). If only mono is supported, it might return `(T,)` if wished. But I understand it's more a question of taste

DarhkVoyd:
@abidlabs Can I help with this? please let me know how I can get started.

PR #6062, opened 2023-10-23
check for built in functions #5504

SrijanSahaySrivastava:
Try-Except block to check for unsupported built in function for issue #5504

## Description

Try-Except block to check for unsupported built in function 

Closes: #5504


SrijanSahaySrivastava:
@abidlabs Please look into these changes, i tried to add a try block for the issue.

abidlabs:
Hi @SrijanSahaySrivastava thanks for contributing this PR. Actually, we should not prevent users from using built-ins, but we can give them generic names for their inputs, e.g. `input_1, `input_2`, etc.

Secondly, it would be better to start off working on the `v4` branch as we will merge that into main soon, and it has considerable changes from main. 

I'll go ahead and close this PR, but if you can create a new one setting `v4` as the base branch, that would be much better. Thank you!

Issue #3805, opened 2023-04-10
Let Clients connect to Spaces with Gradio auth


abidlabs:
Right now, the Python Client can't connect to Spaces or other hosted apps with auth enabled. Let's enable that

justinzyw:
looking forward to the release of this function

tomsoderlund:
Following this, also interested

freddyaboulton:
This got done in gradio 4.0 by @aliabid94. Closing!

wujohns:
so...how to use it

ggoonnzzaallo:
Hey @aliabid94 , would you mind sharing your implementation?

PawKanarek:
@ggoonnzzaallo do you know how to implement client with authentication? CC @aliabid94

PR #6069, opened 2023-10-23
Swap websockets for SSE

aliabid94:
Replace websockets with server side events. Everything should work.

Fixes: #3805


aliabid94:
Everything should be ready to review pretty much. I scraped most of test_queueing.py because it was a mess that just used way too much mocking of internals. May add more tests there.

client/python/gradio_client/client.py
```diff
@@ -73,6 +74,7 @@ def __init__(
         serialize: bool = True,
         output_dir: str | Path = DEFAULT_TEMP_DIR,
         verbose: bool = True,
+        auth: tuple[str, str] | None = None,
```
freddyaboulton, line 77: Looks like this fixes #3805  ?

client/python/gradio_client/client.py
```diff
@@ -781,7 +809,7 @@ def __init__(self, client: Client, fn_index: int, dependency: dict):
         self.api_name: str | Literal[False] | None = (
             "/" + api_name if isinstance(api_name, str) else api_name
         )
-        self.use_ws = self._use_websocket(self.dependency)
+        self.protocol = self._get_protocol(self.dependency)
```
freddyaboulton, line 812: Just merged a PR that causes some conflicts. But all the prototcol related stuff for backwards compatibility with websockets/http can be added to `EndpointV3Compatibility` class

client/python/gradio_client/client.py
```diff
@@ -850,23 +878,26 @@ def _inner(*data):
...
+            }
+
+            if self.protocol == "ws":
                 result = utils.synchronize_async(self._ws_fn, data, hash_data, helper)
                 if "error" in result:
+                    raise ValueError(result["error"])
+            elif self.protocol == "sse":
+                result = utils.synchronize_async(self._sse_fn, data, hash_data, helper)
+                if "error" in result:
+                    print(">>>", result)
```
freddyaboulton, line 899: Remember to delete

client/python/gradio_client/utils.py
```diff
@@ -297,13 +298,119 @@ async def get_pred_from_ws(
...
+                            except Exception as e:
+                                result = [e]
+                            helper.job.outputs.append(result)
+                        helper.job.latest_status = status_update
+
+                    if resp["msg"] == "queue_full":
+                        raise QueueError("Queue is full! Please try again.")
+                    elif resp["msg"] == "send_data":
+                        event_id = resp["event_id"]
+                        req = requests.post(
```
freddyaboulton, line 396: requests.post is blocking - we should be using the async client.post method

gradio/routes.py
```diff
@@ -628,17 +606,49 @@ async def join_queue(
...
+            )
+
+        @app.post("/queue/data", dependencies=[Depends(login_check)])
+        async def queue_data(
+            body: PredictBody,
+            request: fastapi.Request,
+            username: str = Depends(get_current_user),
+        ):
+            blocks = app.get_blocks()
+            blocks._queue.attach_data(body)
```
freddyaboulton, line 651: This request always succeeds even if we don't attach the data. I think we should return an appropriate status code and let the clients handle it.

freddyaboulton (commented):
Really good stuff @aliabid94 ! I noticed something off about event cancelling and left some other comments.

gradio/queueing.py
```diff
@@ -5,9 +5,9 @@
 import json
 import time
 import traceback
-from asyncio import TimeoutError as AsyncTimeOutError
+import uuid
```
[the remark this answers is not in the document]
aliabid94: Fixed I believe!

gradio/routes.py
aliabid94: done

gradio/blocks.py
```diff
@@ -543,6 +543,8 @@ def __init__(
             }
             analytics.initiated_analytics(data)
 
+        self.queue()
```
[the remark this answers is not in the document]
aliabid94: Queue is always on, with 40 workers it should have the same behaviour as the current "queue=False" mode. Is there any reason to support no queue? API requests can still hit the /run/ endpoint directly if they choose to.

client/python/gradio_client/client.py
```diff
@@ -327,7 +333,7 @@ def submit(
         inferred_fn_index = self._infer_fn_index(api_name, fn_index)
 
         helper = None
-        if self.endpoints[inferred_fn_index].use_ws:
+        if self.endpoints[inferred_fn_index].protocol in ("ws", "sse"):
```
freddyaboulton, line 336: minor: let's modify and `use_ws` to check if the protocol is "ws" or "sse" since we're using it in different places

client/python/gradio_client/utils.py
```diff
@@ -297,13 +299,119 @@ async def get_pred_from_ws(
...
+                                result = [e]
+                            helper.job.outputs.append(result)
+                        helper.job.latest_status = status_update
+
+                    if resp["msg"] == "queue_full":
+                        raise QueueError("Queue is full! Please try again.")
+                    elif resp["msg"] == "send_data":
+                        event_id = resp["event_id"]
+                        helper.event_id = event_id
+                        async with httpx.AsyncClient() as http:
```
freddyaboulton, line 399: Nice. Can we use the same client as the one that's streaming or do we need to create a new one?

gradio/queueing.py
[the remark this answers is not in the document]
freddyaboulton: Nice what was the fix?

client/python/gradio_client/client.py
```diff
@@ -602,14 +610,33 @@ def __del__(self):
...
         r = requests.get(
-            urllib.parse.urljoin(self.src, utils.CONFIG_URL), headers=self.headers
+            urllib.parse.urljoin(self.src, utils.CONFIG_URL),
+            headers=self.headers,
+            cookies=self.cookies,
         )
         if r.ok:
             return r.json()
+        elif r.status_code == 401:
+            raise ValueError("Could not load {self.src}. Please login.")
```
freddyaboulton, line 635: nit: Need an fstring here

gradio/blocks.py
```diff
@@ -1708,7 +1697,6 @@ def launch(
         inbrowser: bool = False,
         share: bool | None = None,
         debug: bool = False,
-        max_threads: int = 40,
```
freddyaboulton, line 1711: I think we should still keep this parameter. It's not the same as the concurrency_count of the queue. You may want to increase it to match the concurrency_count for example.

gradio/blocks.py
```diff
@@ -1380,7 +1380,7 @@ async def process_api(
         inputs: list[Any],
         state: SessionState | None = None,
         request: routes.Request | list[routes.Request] | None = None,
-        iterators: dict[int, Any] | None = None,
```
freddyaboulton, line 1383: We never even use this parameter? Lol

freddyaboulton (approved):
This looks great @aliabid94 !

client/python/test/test_client.py
```diff
@@ -282,14 +282,18 @@ def test_cancel_subsequent_jobs_state_reset(self, yield_demo):
             time.sleep(3)
             job1.cancel()
 
+            assert len(job1.outputs()) > 0
             assert len(job1.outputs()) < len("abcdefefadsadfs")
             assert job1.status().code == Status.CANCELLED
+            print(">>1 ", job1.outputs())
```
abidlabs, line 288: remove

client/python/test/test_client.py
```diff
@@ -282,14 +282,18 @@ def test_cancel_subsequent_jobs_state_reset(self, yield_demo):
...
 
             job2 = client.submit("abcd", api_name="/predict")
+            assert len(job2.outputs()) == 0
             while not job2.done():
                 time.sleep(0.1)
             # Ran all iterations from scratch
             assert job2.status().code == Status.FINISHED
+            print(">>>", job2.outputs())
```
abidlabs, line 296: remove

abidlabs:
Nice @aliabid94! Looks like @freddyaboulton already reviewed this extensively, so I just tried out a few demos which work well. Note that a bunch of demos still have `.queue()` -- we can remove those. 

And we should implement the one-worker-per-event in the queue so that setting the concurrency_count to 40 doesn't break a bunch of demos.

client/python/gradio_client/client.py
aliabid94: sorry don't understand what you're suggesting

gradio/blocks.py
aliabid94: I'll restore it in this PR, but do we need it really?  Seems like controlling the number of workers is the main thing that a user would care about cc @abidlabs

client/python/gradio_client/client.py
freddyaboulton: Sorry, instead of `self.endpoints[inferred_fn_index].protocol in ("ws", "sse")` in multiple places, we do `self.endpoints[inferred_fn_index].use_http` or something like that.

gradio/blocks.py
freddyaboulton: I agree the concurrency count of the queue is the main parameter users will want to control. But the way I understand it, starlette will use the same threadpool to run all the functions (unless they're coroutines). That's what `anyio.to_thread.run_sync` is doing. So if we remove the param, only 40 functions can only run concurrently.

Maybe we can set max_threads to always match the concurrency_count but I still don't think that's the best solution. Since events can still hit the predict route directly.

abidlabs:
Synced with @aliabid94 on this. Instead of having two separate requests to send the data and join the queue, we should just do it in a single step, now that we use file uploads for everything. Should make it easier if someone is trying to query via curl.

Speaking of which, it'd be great to add curl to the API docs @aliabid94 👀

abidlabs:
Oh a big thing -- it'd be good to confirm that this allows Gradio apps to run with the queue on Colab!

gradio/routes.py
```diff
@@ -20,10 +21,9 @@
 import threading
 import time
 import traceback
-from asyncio import TimeoutError as AsyncTimeOutError
-from collections import defaultdict
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Type
+from queue import Empty as EmptyQueue
+from typing import TYPE_CHECKING, Any, AsyncIterator, Dict, List, Optional, Type
```
freddyaboulton, line 26: reload mode is connecting via websocket.
I think we can refactor that to use SSE as well?
Would allow us to get rid of websockets dependency in the requirements file.
We can also get rid of `ws_login_check` right now though.

PR #6071, opened 2023-10-23
Fixes markdown rendering in examples

abidlabs:
Ensures that the Example component of Markdown renders Markdown correctly. Thanks to @pngwn's recent changes, I'm even able to pull in the props from the original component to ensure that the Example component has those same attributes. 

Here's an example:

```py
import gradio as gr

a = [['$1+1=2$', '$1+1=2$'], ['[aaa](https://www.google.com)', '[aaa](https://www.google.com)']]


with gr.Blocks() as demo:
    m = gr.Markdown("$x+y=z$", latex_delimiters=[])
    m2 = gr.Markdown("$x+y=z$", latex_delimiters=[{"left": "$", "right": "$", "display": False}])
    ds = gr.Examples(a, [m, m2])
    demo.launch()
```

Notice that in the examples, both columns render the link correctly, but only the 2nd column renders the LaTeX because only the 2nd Markdown component uses the default LaTeX delimiters.

image

Closes: #4509

Finally, makes the `latex_delimiters` consistent across all of our components. This is a breaking change, but probably welcome from users since dollar signs were being converted to LaTeX inadvertently. Its easy enough to get the original behavior by passing in custom `latex_delimiters`. This was something we had discussed with @dawoodkhan82 earlier.


abidlabs:
Pretty minor PR, so I'll go ahead and merge this in just fyi @dawoodkhan82 @freddyaboulton

Issue #6072, opened 2023-10-24
Cannot load an interface with a load event attached as of gradio 3.50.2


freddyaboulton:
### Describe the bug

If you try to load an interface with a load event using `gr.load`, it will error.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
df = gr.load("gradio-tests/dataframe_load", src="spaces")
```

### Logs

```shell
KeyError                                  Traceback (most recent call last)
Cell In[2], line 1
----> 1 df = gr.load("gradio-tests/dataframe_load", src="spaces")

File ~/miniconda3/envs/blank/lib/python3.10/site-packages/gradio/external.py:70, in load(name, src, api_key, hf_token, alias, **kwargs)
     65     warn_deprecation(
     66         "The `api_key` parameter will be deprecated. "
     67         "Please use the `hf_token` parameter going forward."
     68     )
     69     hf_token = api_key
---> 70 return load_blocks_from_repo(
     71     name=name, src=src, hf_token=hf_token, alias=alias, **kwargs
     72 )

File ~/miniconda3/envs/blank/lib/python3.10/site-packages/gradio/external.py:109, in load_blocks_from_repo(name, src, hf_token, alias, **kwargs)
    104         warnings.warn(
    105             """You are loading a model/Space with a different access token than the one you used to load a previous model/Space. This is not recommended, as it may cause unexpected behavior."""
    106         )
    107     Context.hf_token = hf_token
--> 109 blocks: gradio.Blocks = factory_methods[src](name, hf_token, alias, **kwargs)
    110 return blocks

File ~/miniconda3/envs/blank/lib/python3.10/site-packages/gradio/external.py:505, in from_spaces(space_name, hf_token, alias, **kwargs)
    498 if kwargs:
    499     warnings.warn(
    500         "You cannot override parameters for this Space by passing in kwargs. "
    501         "Instead, please load the Space as a function and use it to create a "
    502         "Blocks or Interface locally. You may find this Guide helpful: "
    503         "https://gradio.app/using_blocks_like_functions/"
    504     )
--> 505 return from_spaces_blocks(space=space_name, hf_token=hf_token)

File ~/miniconda3/envs/blank/lib/python3.10/site-packages/gradio/external.py:511, in from_spaces_blocks(space, hf_token)
    509 client = Client(space, hf_token=hf_token)
    510 predict_fns = [endpoint._predict_resolve for endpoint in client.endpoints]
[...]

PR #6073, opened 2023-10-24
Fix remaining xfail tests in backend

freddyaboulton:
## Description

* Make `gradio_client` backwards compatible with v3 spaces
* Raise an error if someone tries to load a v3 space with gradio v4 via `gr.load`
* Fix the remaining tests that are failing in the backend. All of them related to `external.py` or `gr.load` 😓 . This uncovered some gaps/bugs in v4.

~There's still an outstanding issue - since we removed **kwargs from component init, we can't call `gr.load` on a space that has examples because the dataset component has `component_props` saved in the config but `component_props` is not a parameter of init. @abidlabs let's synch tomorrow on the best pattern to deal with this.~

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/client/python/gradio_client/client.py
+++ b/client/python/gradio_client/client.py
@@ -30,8 +30,9 @@ from huggingface_hub.utils import (
 )
 from packaging import version
 
-from gradio_client import utils
+from gradio_client import serializing, utils
 from gradio_client.documentation import document, set_documentation_group
+from gradio_client.exceptions import SerializationSetupError
 from gradio_client.utils import (
     Communicator,
     JobStatus,
@@ -128,11 +129,15 @@ class Client:
         self.upload_url = urllib.parse.urljoin(self.src, utils.UPLOAD_URL)
         self.reset_url = urllib.parse.urljoin(self.src, utils.RESET_URL)
         self.config = self._get_config()
+        self.app_version = version.parse(self.config.get("version", "2.0"))
         self._info = self._get_api_info()
         self.session_hash = str(uuid.uuid4())
 
+        endpoint_class = Endpoint
+        if self.app_version < version.Version("4.0.0"):
+            endpoint_class = EndpointV3Compatibility
         self.endpoints = [
-            Endpoint(self, fn_index, dependency)
+            endpoint_class(self, fn_index, dependency)
             for fn_index, dependency in enumerate(self.config["dependencies"])
         ]
 
@@ -360,9 +365,7 @@ class Client:
         else:
             api_info_url = urllib.parse.urljoin(self.src, utils.RAW_API_INFO_URL)
 
-        # Versions of Gradio older than 3.29.0 returned format of the API info
-        # from the /info endpoint
-        if version.parse(self.config.get("version", "2.0")) > version.Version("3.36.1"):
+        if self.app_version > version.Version("3.36.1"):
             r = requests.get(api_info_url, headers=self.headers)
             if r.ok:
                 info = r.json()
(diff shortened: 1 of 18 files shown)


abidlabs:
Hmm I thought backend tests were passing?

abidlabs:
> dataset component has component_props saved in the config but component_props is not a parameter of init. @abidlabs let's synch tomorrow on the best pattern to deal with this.

We could modify the `recover_kwargs()` to additionally remove `component_props` and `component_ids` if we are working with a `Dataset` here:

https://github.com/gradio-app/gradio/blob/447dfe06bf19324d88696eb646fd1c5f1c4e86ed/gradio/utils.py#L955

test/test_external.py
```diff
@@ -441,14 +452,15 @@ def check_dataset(config, readme_examples):
         assert dataset["props"]["samples"] == [[cols_to_rows(readme_examples)[1]]]
 
 
+@pytest.mark.xfail
```
freddyaboulton, line 455: This is #6072

client/python/test/test_client.py
```diff
@@ -64,12 +67,24 @@ def test_numerical_to_label_space(self):
         ):
             client.predict("male", 77, 10, api_name="predict")
 
+    @pytest.mark.flaky
+    def test_numerical_to_label_space_v4(self):
```
freddyaboulton, line 71: I ported over the spaces to use v4 gradio so we can have coverage before release

gradio/components/dataset.py
```diff
@@ -61,16 +63,19 @@ def __init__(
             container: If True, will place the component in a container - providing some extra padding around the border.
             scale: relative width compared to adjacent Components in a Row. For example, if Component A has scale=2, and Component B has scale=1, A will be twice as wide as B. Should be an integer.
             min_width: minimum pixel width, will wrap if not sufficient screen space to satisfy this value. If a certain scale value results in this Component being narrower than min_width, the min_width parameter will be respected first.
+            component_props: This is created when the dataset is initialized. Not intended to be set manually. But we add it to init so that Datasets can be loaded from spaces with gr.load.
+            component_ids: This is created when the dataset is initialized. Not intended to be set manually. But we add it to init so that Datasets can be loaded from spaces with gr.load.
         """
-        super().__init__(visible=visible, elem_id=elem_id, elem_classes=elem_classes)
+        super().__init__(visible=visible, elem_id=elem_id, elem_classes=elem_classes, root_url=root_url,
```
freddyaboulton, line 69: root_url and _skip_init_processing was not being set for dataset

gradio/components/image.py
```diff
@@ -249,7 +249,7 @@ def postprocess(
         elif isinstance(y, _Image.Image):
             path = processing_utils.save_pil_to_cache(y, cache_dir=self.GRADIO_CACHE)
         elif isinstance(y, (str, Path)):
-            path = y if isinstance(y, str) else y.name
+            path = y if isinstance(y, str) else str(y)
```
freddyaboulton, line 252: We should have always returned the full path. Bug on my part.

abidlabs:
```suggestion
            path = str(y)
```

client/python/gradio_client/client.py
```diff
@@ -128,11 +129,15 @@ def __init__(
         self.upload_url = urllib.parse.urljoin(self.src, utils.UPLOAD_URL)
         self.reset_url = urllib.parse.urljoin(self.src, utils.RESET_URL)
         self.config = self._get_config()
+        self.app_version = version.parse(self.config.get("version", "2.0"))
         self._info = self._get_api_info()
         self.session_hash = str(uuid.uuid4())
 
+        endpoint_class = Endpoint
+        if self.app_version < version.Version("4.0.0"):
+            endpoint_class = EndpointV3Compatibility
```
abidlabs, line 138: Nice

gradio/components/dataset.py
```diff
@@ -61,16 +63,19 @@ def __init__(
...
+                        _skip_init_processing=_skip_init_processing)
         self.container = container
         self.scale = scale
         self.min_width = min_width
         self._components = [get_component_instance(c) for c in components]
         self.component_props = [
             utils.recover_kwargs(
                 component.get_config(),
-                ["value"],
+                ["value", "component_props", "component_ids"],
```
abidlabs, line 83: I don't think this addition is necessary. These are attributes of the `Dataset` class, not of its components
```suggestion
                ["value"],
```

gradio/components/dataset.py
```diff
@@ -61,16 +63,19 @@ def __init__(
             container: If True, will place the component in a container - providing some extra padding around the border.
             scale: relative width compared to adjacent Components in a Row. For example, if Component A has scale=2, and Component B has scale=1, A will be twice as wide as B. Should be an integer.
             min_width: minimum pixel width, will wrap if not sufficient screen space to satisfy this value. If a certain scale value results in this Component being narrower than min_width, the min_width parameter will be respected first.
+            component_props: This is created when the dataset is initialized. Not intended to be set manually. But we add it to init so that Datasets can be loaded from spaces with gr.load.
+            component_ids: This is created when the dataset is initialized. Not intended to be set manually. But we add it to init so that Datasets can be loaded from spaces with gr.load.
```
abidlabs, line 67: Instead of adding these parameters, which are never used, it would be better to update this logic:

https://github.com/gradio-app/gradio/blob/797e026c87983bc1b51cb3fe05e1d7b19d15635c/gradio/blocks.py#L612

to remove `component_props`

gradio/components/dataset.py
abidlabs: Let's add `render` here as well, its not being passed up right now

abidlabs:
Thanks for taking care of this @freddyaboulton! Very nice to have client compatibility with v3. Changes look good & tested it out. Just a few nits above.

gradio/external.py
```diff
@@ -498,7 +520,13 @@ def from_spaces(
 
 def from_spaces_blocks(space: str, hf_token: str | None) -> Blocks:
     client = Client(space, hf_token=hf_token)
-    predict_fns = [endpoint._predict_resolve for endpoint in client.endpoints]
+    if client.app_version < version.Version("4.0.0"):
+        raise GradioVersionIncompatibleError(
+            f"Gradio version 4.x cannot load spaces with versions less than 4.x ({client.app_version})."
```
abidlabs, line 525: consider
```suggestion
            f"Gradio version 4.x cannot load Spaces with versions less than 4.x (in this case, {client.app_version})."
```

gradio/components/dataset.py
freddyaboulton: I think we need a general pattern to exclude props that are in the config but not in the init signature. Otherwise, a custom component that sets a property in the init method that's not in the signature won't be able to be loaded. I'll think about this and push up.

gradio/components/dataset.py
freddyaboulton: Yep you're right!

abidlabs (approved):
Nice! Changes look good to me @freddyaboulton

freddyaboulton:

Issue #6043, opened 2023-10-20
Some Example components are broken on v4


abidlabs:
### Describe the bug

The Markdown Example component is broken on `v4`, as can be confirmed with the repro here: https://github.com/gradio-app/gradio/issues/4509

The Multiselect Dropdown component also seems to be broken (try running `python demo/kitchen_sink/run.py`) 

Looking into it

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

see above

### System Info

```shell
v4 branch
```

### Severity

I can work around it

abidlabs:
I think `File` is also broken, at least when multiple files are selected. See the `zip_files` demo

PR #6075, opened 2023-10-24
Fix Dropdown Example component and example processing logic

abidlabs:
Fixes: #6043 

Check by running `demo/kitchen_sink/run.py` or

```py
import gradio as gr

example_1 = [["a", "b"]]
example_2 = [["a"]]

io = gr.Interface(lambda x:x, 
                  gr.Dropdown(["a", "b"], 
                              multiselect=True), "dropdown", examples=[example_1, example_2])
io.launch()
```

### DIFF
--- a/gradio/helpers.py
+++ b/gradio/helpers.py
@@ -259,7 +259,6 @@ class Examples:
 
         async def load_example(example_id):
             processed_example = self.non_none_processed_examples[example_id]
-            # examples = utils.resolve_singleton(processed_example)
             if len(self.inputs_with_examples) == 1:
                 return update(
                     value=processed_example[0], **self.dataset.component_props[0]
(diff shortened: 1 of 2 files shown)


gradio/helpers.py
```diff
@@ -259,16 +259,14 @@ def create(self) -> None:
...
-                    for i, ex in enumerate(examples)
-                ]
-            )
+            if len(self.inputs_with_examples) == 1:
+                return update(
+                    value=processed_example[0], **self.dataset.component_props[0]
+                )
+            return [
+                update(value=processed_example[i], **self.dataset.component_props[i])
+                for i in range(len(self.inputs_with_examples))
```
abidlabs, line 268: Should be `self.inputs_with_examples` not `examples`

pngwn: Ah cool. I have to admit all of this code confused me a great deal. I still don't know what all of the references to different examples/ inputs are.

abidlabs: Yeah the code for examples is messy, will clean up at some point. Your PR using gr.update is nice though, captures an edge case we didn’t think about originally!

PR #6077, opened 2023-10-24
Proposal: sample demo for custom components should be a `gr.Interface`

abidlabs:
Should we switch the sample demo to be an `Interface` instead? It allows the user to immediately try out the component in a functioning Gradio app and can support examples very easily as well. I've commented out the example functionaly by default since not every Gradio component has an "example version". 

This PR is a suggested implementation


pngwn (approved):
Looks good to me! Thanks @abidlabs!

freddyaboulton (commented):
Thanks @abidlabs ! 

One issue is that not every component has an interactive version (AnnotatedImage, Gallery, JSON, maybe others?) so using as input in an interface is not great with the default interface code. User's won't be able to click on the gallery to upload one for example.

Maybe we can just special demos for these components in the OVERRIDES: https://github.com/gradio-app/gradio/blob/6b82a9760f7c10fc76b357d7472864bcc97a4025/gradio/cli/commands/components/_create_utils.py#L41

abidlabs:
Yeah makes sense. So for the static-only components, I'll use this demo:

```py
example = {name}().example_inputs()

with gr.Blocks() as demo:
    {name}(),  # blank component
    {name}(value=example),  # populated component
)
```

For layouts, how about this:

```py
with gr.Blocks() as demo:
    with {name}():
        gr.Textbox()
        gr.Textbox()
)
```

freddyaboulton:
Yep agree with the proposal @abidlabs !

PR #6079, opened 2023-10-24
fix cc build

pngwn:
## Description

Pathing issue withe frontend build should be resolved now. Also tweaked the name we use in the package.json to match the gradio package name.

Removed the old test components because they don't reflect recent changes. Added a new one in that can be used to test the build.

Closes: #(issue)

## 🎯 PRs Should Target Issues

## Tests


js/preview/src/build.ts
```diff
@@ -29,7 +29,7 @@ export async function make_build({
 
 			const exports: string[][] = [
 				["component", pkg.exports["."] as string],
-				["exmaple", pkg.exports["./example"] as string]
+				["example", pkg.exports["./example"] as string]
```
pngwn, line 32: horrendous typo

js/preview/src/build.ts
```diff
@@ -43,23 +43,25 @@ export async function make_build({
 							emptyOutDir: true,
 							outDir: join(template_dir, entry),
 							lib: {
-								entry: join(source_dir, entry, "index.ts"),
+								entry: join(source_dir, path),
```
pngwn, line 46: typo didn't matter however because we were look for `Index.svelte/index.ts` and `Example.svelte/index.ts` which is nonsensical.

js/preview/src/build.ts
```diff
@@ -43,23 +43,25 @@ export async function make_build({
...
 								fileName: "index.js",
 								formats: ["es"]
 							},
+							minify: true,
 							rollupOptions: {
 								output: {
-									entryFileNames: "[name].js"
+									entryFileNames: (chunkInfo: PreRenderedChunk) =>
+										`${chunkInfo.name.toLocaleLowerCase()}.js`
```
pngwn, line 54: Changed so we get `index.js` not `Index.js`

js/preview/src/build.ts
```diff
@@ -43,23 +43,25 @@ export async function make_build({
...
 					});
 				} catch (e) {
-					console.error(e);
+					throw e;
 				}
 			}
 		}
 	} catch (e) {
-		console.error(e);
+		throw e;
```
pngwn, line 65: We were squashing errors in the try/catch which is why this didn't error out as it should have. This will now fail noisily if the build doesn't succeed.

js/preview/src/build.ts
```diff
@@ -43,23 +43,29 @@ export async function make_build({
...
+							minify: true,
 							rollupOptions: {
 								output: {
-									entryFileNames: "[name].js"
+									entryFileNames: (chunkInfo: PreRenderedChunk) => {
+										if (chunkInfo.isEntry) {
+											return "index.js";
+										}
+										return `${chunkInfo.name.toLocaleLowerCase()}.js`;
+									}
```
pngwn, line 58: Okay I've changed this so we always get an `index.js` for the main entry point (which is what we look for when serving the custom component). Nice thing is this will even work if users change their entrypoints in the package.json.

gradio/cli/commands/components/_create_utils.py
```diff
@@ -217,7 +219,7 @@ def ignore(s, names):
         ignore=ignore,
     )
     source_package_json = json.loads(Path(frontend / "package.json").read_text())
-    source_package_json["name"] = name.lower()
+    source_package_json["name"] = package_name
```
pngwn, line 222: Tweaked this so the name in the `package.json` is the name in `pyproject.toml`

js/preview/src/build.ts
pngwn: outdated.

freddyaboulton (approved):
Thanks @pngwn ! I added a unit test too. Should be good to merge whenever.

pngwn:
Thanks @freddyaboulton. A unit tests seem to be failing now.

freddyaboulton:
Lol 😆 I guess the cache was not properly restored?

.github/workflows/backend.yml
```diff
@@ -168,6 +168,7 @@ jobs:
       with:
         path: |
           gradio/templates/*
+          gradio/node/*
```
freddyaboulton, line 171: 👍

freddyaboulton:
Client tests are fixed in another PR #6073 . Gonna merge just so that we have a working build. Will look into why the test is not passing in CI.

Issue #5866, opened 2023-10-11
open-source FRP share server & allow users to point Gradio to their own installation


abidlabs:
We see a lot of issues around users unable to connect to the share server, or the share links terminating after a small amount of times. It might be a good idea to open-source the FRP share server and let users run it on their own machines so that they can run it for however long they want.

abidlabs:
Done in v4 branch: https://github.com/gradio-app/gradio/pull/6091

PR #6091, opened 2023-10-25
Open source FRP server and allow `gradio` to connect to custom share servers

abidlabs:
Open-sourced: https://github.com/huggingface/frp and wrote up a detailed guide on the readme. 

I set up a custom share server myself at `my-gpt-wrapper.com` so that we can test this PR:

```py
import gradio as gr

io = gr.Interface(lambda x: x.shape, "image", "text").launch(
   share=True, 
   share_server_address="my-gpt-wrapper.com:7000"
)
```

Closes: #5866 

cc @XciD


XciD (approved):

gradio/blocks.py
```diff
@@ -1727,10 +1728,13 @@ def launch(
         show_api: bool = True,
         allowed_paths: list[str] | None = None,
         blocked_paths: list[str] | None = None,
+        *,
```
aliabid94, line 1731: what about making all of the arguments keyword-only in launch()? I would support doing this throughout the codebase for 4.0

aliabid94:
To make this work for me locally, I had to replace 
`template = "frontend/share.html" if blocks.share else "frontend/index.html"`
with
`template = "frontend/index.html"`
because share links try to load from CDN in `share.html` and therefore can only run from pip, which has the CDM uploads. This may be something to document (or even better, automatically use index.html if not running off of pip)

aliabid94 (approved):
Looks great

gradio/blocks.py
abidlabs: I didn't want to be too aggressive here since it'll just cause unnecessary breaking of parameters. But let me move it a few positions back

abidlabs:
> To make this work for me locally, I had to replace...

Yes, but this shouldn't affect anyone who's installed from `pip` even if they are setting up their own share server

abidlabs:
Thanks for the review!

PR #6092, opened 2023-10-25
Add a stand-alone install command and tidy-up the fallback template

freddyaboulton:
## Description

Add a stand-alone install command. This will be useful when cloning a custom component hosted on any git repo. Just clone as you normally would and then run `gradio cc install`. Should make it easier to "remix" custom components.

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/mighty-coats-do.md
+++ b/.changeset/mighty-coats-do.md
@@ -1,0 +1,6 @@
+---
+"@gradio/fallback": minor
+"gradio": minor
+---
+
+feat:Add a stand-alone install command and tidy-up the fallback template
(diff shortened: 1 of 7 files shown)


abidlabs:
Nice @freddyaboulton this looks good. 

Just two asides about the CLI:
* I keep typing in `gradio dev` instead of `gradio cc dev` and `gradio build` instead of `gradio cc build` -- perhaps we can print, "did you mean gradio cc dev" etc. to avoid confusion for users
* Can we make the `--install` flag default for `gradio cc create`? Its probably the desired behavior for the vast majority of users executing the custom components workflow

abidlabs (approved):

freddyaboulton:
Thanks @abidlabs ! Implemented your suggestions. Merging

PR #6094, opened 2023-10-25
Image v4

pngwn:
## Description

Closes: #(issue)

## 🎯 PRs Should Target Issues

## Tests


client/python/gradio_client/client.py
```diff
@@ -871,6 +871,7 @@ def _predict(*data) -> tuple:
                 )
                 result = json.loads(response.content.decode("utf-8"))
             try:
+                print(result)
```
freddyaboulton, line 874: Replace `FILE_DATA` in the client utils with `FILE_DATA = "'Dict(path: str, url: str | None, size: int | None, orig_name: str | None, mime_type: str | None)'"`

Will think of a better way to do this going forward

gradio/imagetools.py
```diff
@@ -0,0 +1,50 @@
+from typing import Literal
```
freddyaboulton, line 1: add `from __future__ import annotations` or else we can't import in 3.8 because of `|`

gradio/data_classes.py
```diff
@@ -119,10 +119,9 @@ def from_json(cls, x) -> GradioRootModel:
 
 
 class FileData(GradioModel):
-    name: Optional[str] = None
-    data: Optional[str] = None  # base64 encoded data
+    path: str  # server filepath
```
freddyaboulton, line 122: Also change `serialize_data` in blocks to be 

```python
        def format_file(s):
            return FileData(path=s).model_dump()
```

client/js/package.json
```diff
@@ -13,6 +13,7 @@
 		"./package.json": "./package.json"
 	},
 	"dependencies": {
+		"@gradio/upload": "workspace:0.3.0-beta.4",
```
pngwn, line 16: change to looser pin

client/js/src/client.ts
```diff
@@ -843,28 +844,28 @@ export function api_factory(
...
-						// orig_name: "file.csv"
-					};
-					update_object(data, o, path);
+					// const o = {
+					// 	is_file: true,
+					// 	name: `${file_url}`,
+					// 	data: null
+					// 	// orig_name: "file.csv"
+					// };
+
```
pngwn, line 866: remove

client/python/gradio_client/client.py
```diff
@@ -929,20 +930,16 @@ def _upload(
...
                 else:
                     o = next(o for ix, o in enumerate(result) if indices[ix] == i)
                     res = {
-                        "is_file": True,
-                        "name": o,
+                        "path": o,
                         "orig_name": Path(fs).name,
-                        "data": None,
                     }
                 uploaded.append(res)
```
pngwn, line 944: Do we need to check for the version here and do both? I don't think this is backwards compatible.

client/python/gradio_client/client.py
```diff
@@ -1015,18 +1012,19 @@ def _download_file(
...
+                assert data is not None, f"The 'data' field is missing in {x}"
+                file_name = utils.decode_base64_to_file(data, dir=save_dir).name
+            if x.get("is_file", True):
+                filepath = x.get("path")
+                assert filepath is not None, f"The 'path' field is missing in {x}"
                 file_name = utils.download_file(
                     root_url + "file=" + filepath,
                     hf_token=hf_token,
                     dir=save_dir,
                 )
```
pngwn, line 1026: I swapped this around because we don't have `is_file` any more, `is_file` is implicit. For previous version i think the `data` check should work, if it is present it should be base64 data. New versions that do not have a data key will skip this branch.

client/python/gradio_client/data_classes.py
```diff
@@ -6,7 +6,7 @@
 
 
 class FileData(TypedDict):
-    name: str | None  # filename
+    path: str | None  # filename
     data: str | None  # base64 encoded data
```
pngwn, line 10: I left this on the client `FileData` model for the above backwards compat reasons.

client/python/gradio_client/utils.py
```diff
@@ -660,6 +660,7 @@ def traverse(json_obj: Any, func: Callable, is_root: Callable) -> Any:
 
 def value_is_file(api_info: dict) -> bool:
     info = _json_schema_to_python_type(api_info, api_info.get("$defs"))
+    print("INFO", info)
```
pngwn, line 663: remove

gradio/__init__.py
```diff
@@ -82,19 +82,18 @@
...
     Mic,
     Microphone,
     Numpy,
-    Paint,
-    Pil,
+    # Paint,
+    # Pil,
     PlayableVideo,
-    Sketchpad,
+    # Sketchpad,
```
pngwn, line 95: So we don't actually have the ImageEditor in this PR, so these templates don't make much sense. I will bring them back but they might not make a huge amount of sense in v4. We can see.

I don't think the `Webcam` templates really makes any sense now so I did remove that one. I also think `Pil` is a little excessive.

gradio/blocks.py
```diff
@@ -1108,6 +1108,7 @@ def format_file(s):
                 )
             api_info = block.api_info()
             if client_utils.value_is_file(api_info):
+                print("TRUE FOR VALUE IS FILE")
```
pngwn, line 1111: remove

gradio/cli/commands/components/build.py
```diff
@@ -55,6 +55,7 @@ def _build(
                 return
             else:
                 live.update(":white_check_mark: Build succeeded!")
+                live.update(pipe.stdout)
```
pngwn, line 58: remove, wrong branch

gradio/components/image.py
```diff
@@ -47,15 +45,16 @@ def __init__(
...
             "1", "L", "P", "RGB", "RGBA", "CMYK", "YCbCr", "LAB", "HSV", "I", "F"
         ] = "RGB",
-        invert_colors: bool = False,
-        source: Literal["upload", "webcam", "canvas"] = "upload",
-        tool: Literal["editor", "select", "sketch", "color-sketch"] | None = None,
+        sources: list[Literal["upload", "webcam", "clipboard"]] = [
+            "upload",
+            "webcam",
+            "clipboard",
+        ],  # ruff: noqa: B006
```
pngwn, line 57: ruff doesn't like lists as a default argument

gradio/components/image.py
```diff
@@ -104,40 +97,34 @@ def __init__(
...
+        self.sources = sources
+
         self.streaming = streaming
         self.show_download_button = show_download_button
-        if streaming and source != "webcam":
-            raise ValueError("Image streaming only available if source is 'webcam'.")
+        if streaming and len(sources) != 1 and sources[0] != "webcam":
+            raise ValueError(
+                "Image streaming only available if sources is ['webcam']. Streaming not supported with multiple sources."
+            )
```
pngwn, line 127: new streaming check.

gradio/components/image.py
```diff
@@ -160,76 +147,26 @@ def __init__(
...
-                mask_im = _Image.merge("RGB", [alpha_data, alpha_data, alpha_data])
-            return {
-                "image": self._format_image(im),
-                "mask": self._format_image(mask_im),
-            }
-
-        return self._format_image(im)
+        return imagetools.format_image(
+            im, cast(Literal["numpy", "pil", "filepath"], self.type), self.GRADIO_CACHE
+        )
```
pngwn, line 169: I move this into a utils file because i'll need it for `ImageEditor` too and it cleans up the component a bit. Had to cast cos the typing sucks

gradio/components/image.py
```diff
@@ -242,21 +179,14 @@ def postprocess(
...
-            )
-        elif isinstance(y, _Image.Image):
-            path = processing_utils.save_pil_to_cache(y, cache_dir=self.GRADIO_CACHE)
-        elif isinstance(y, (str, Path)):
-            path = y if isinstance(y, str) else y.name
-        else:
-            raise ValueError("Cannot process this value as an Image")
-        return FileData(name=path, data=None, is_file=True)
+
+        return FileData(path=imagetools.save_image(y, self.GRADIO_CACHE))
```
pngwn, line 183: Same as above with saving images, made it a util for reuse.

gradio/components/upload_button.py
```diff
@@ -124,21 +123,15 @@ def example_inputs(self) -> Any:
                 "https://github.com/gradio-app/gradio/raw/main/test/test_files/sample_file.pdf"
             ]
 
-    def _process_single_file(self, f: dict[str, Any]) -> bytes | str:
-        file_name, data, is_file = (
-            f["name"],
-            f["data"],
-            f.get("is_file", False),
-        )
+    def _process_single_file(self, f: dict[str, Any]) -> bytes | NamedString:
```
pngwn, line 126: type was wrong here

gradio/components/upload_button.py
```diff
@@ -147,8 +140,8 @@ def _process_single_file(self, f: dict[str, Any]) -> bytes | str:
             )
 
     def preprocess(
-        self, x: list[dict[str, Any]] | None
-    ) -> bytes | str | list[bytes | str] | None:
+        self, x: list[dict[str, Any]] | dict[str, Any] | None
+    ) -> bytes | NamedString | list[bytes | NamedString] | None:
```
pngwn, line 144: same

gradio/components/video.py
```diff
@@ -153,9 +153,10 @@ def preprocess(self, x: dict | VideoData) -> str | None:
         """
         if x is None:
             return None
+        print(x)
```
pngwn, line 156: remove

gradio/processing_utils.py
```diff
@@ -259,29 +259,20 @@ def move_files_to_cache(data: Any, block: Component):
...
-                    cache_dir=block.GRADIO_CACHE,
+                temp_file_path = save_file_to_cache(
+                    payload.path, cache_dir=block.GRADIO_CACHE
                 )
-                payload.is_file = True
+
             block.temp_files.add(temp_file_path)
-            payload.name = temp_file_path
+            payload.path = temp_file_path
         return payload.model_dump()
```
pngwn, line 276: I think this is correct but not certain

gradio/templates.py
pngwn, line 1: I've mostly just commented these out for now.

js/atoms/src/IconButton.svelte
```diff
@@ -1,13 +1,18 @@
 <script lang="ts">
-	export let Icon: any;
+	import { type ComponentType } from "svelte";
+	export let Icon: ComponentType;
 	export let label = "";
 	export let show_label = false;
 	export let pending = false;
+	export let size: "small" | "large" = "small";
+	export let padded = false;
```
pngwn, line 8: Change to `bordered`

js/atoms/src/UploadText.svelte
```diff
@@ -13,24 +15,39 @@
...
+		border-color: var(--block-border-color);
+		border-radius: var(--block-radius);
+		padding: var(--size-6) var(--size-8);
+		border-style: dashed;
 		display: flex;
-		flex-direction: column;
+
 		justify-content: center;
-		min-height: var(--size-60);
+		/* min-height: var(--size-60); */
```
pngwn, line 33: remove

js/upload/src/utils.ts
```diff
@@ -1,57 +1,57 @@
...
-	root: string,
-	root_url: string | null
-): FileData[] | FileData | null;
-
-export function normalise_file(
-	file: FileData[] | FileData | string | null,
-	root: string,
-	root_url: string | null
+	server_url: string, // root: string,
+	proxy_url: string | null // root_url: string | null
```
pngwn, line 18: Renamed these args for clarity.

js/upload/src/utils.ts
```diff
@@ -65,69 +65,99 @@ function is_url(str: string): boolean {
...
+	}) {
+		this.path = path;
+		this.url = url;
+		this.orig_name = orig_name;
+		this.size = size;
+		this.blob = blob;
+		this.is_stream = is_stream;
+		this.mime_type = mime_type;
+		this.alt_text = alt_text;
+	}
```
pngwn, line 162: `FileData` is a class now, felt a little simpler and better parity with the backend.

test/test_components.py
```diff
@@ -595,56 +583,28 @@ def test_component_functions(self, gradio_temp_dir):
...
-        assert PIL.Image.open(output).size == (10, 30)
+        image_output = gr.Image(type="pil")
+        processed_image = image_output.postprocess(PIL.Image.open(img["path"]))
+        assert processed_image is not None
+        if processed_image is not None:
+            processed = client_utils.encode_url_or_file_to_base64(
+                cast(dict, processed_image).get("path", "")
+            )
+            source = client_utils.encode_url_or_file_to_base64(img["path"])
+            assert processed == source
```
pngwn, line 607: I didn't think we needed all of these base64 tests but could be wrong.

For some reason this test is not passing even thought it seems to be the same. WOuld love some help debugging.

test/test_routes.py
```diff
@@ -298,27 +298,27 @@ def test_get_blocked_paths(self):
...
             "/api/predict/",
             json={
                 "data": [
                     {
-                        "data": media_data.BASE64_IMAGE,
-                        "name": "bus.png",
-                        "is_file": False,
-                        "size": len(media_data.BASE64_IMAGE),
+                        "path": file_response.json()[0],
+                        "size": os.path.getsize("test/test_files/alphabet.txt"),
```
pngwn, line 314: Changes this to upload a file instead, since the base64 thing gives validation errors now.

client/python/gradio_client/client.py
freddyaboulton: It's ok - backwards compatibility is handled by another endpoint class entirely

client/python/gradio_client/client.py
pngwn: @freddyaboulton does that mean I can get rid of this too?

gradio/components/image.py
freddyaboulton: Mutable default arguments are actually not good practice because they get shared across all instances of the class. 

I think we can:
* Set default to None and set sources to ["upload", "webcam", "clipboard"] if it's None
* Set the type to `Iterable[Literal` and use a tuple as the default argument.

Second might be nicer.

gradio/components/image.py
freddyaboulton: Shouldn't it be `not "webcam" in sources` since it could be in any location in the sources array?

gradio/components/image.py
freddyaboulton: Nice

gradio/processing_utils.py
freddyaboulton: Looks good to me tbh

js/upload/src/Upload.svelte
```diff
@@ -82,7 +109,7 @@
 	<input
 		type="file"
 		bind:this={hidden_upload}
-		on:change={loadFilesFromUpload}
+		on:change={load_files_from_upload}
```
freddyaboulton, line 112: The 🐍 -nification of javascript lol

js/upload/src/utils.ts
```diff
@@ -65,69 +65,99 @@ function is_url(str: string): boolean {
...
 export function get_fetchable_url_or_file(
 	path: string | null,
-	root: string,
-	root_url: string | null
+	server_url: string,
+	proxy_url: string | null
 ): string {
 	if (path == null) {
-		return root_url ? `/proxy=${root_url}file=` : `${root}/file=`;
+		return proxy_url ? `/proxy=${proxy_url}file=` : `${server_url}/file=`;
```
freddyaboulton, line 72: I think this is wrong - path is never null since we set it in `upload` so we don't append the `/file=` part

freddyaboulton (commented):
Backend changes look good to me @pngwn !! 

Since you said the front-end is still wip, I didn't take too close a look. The redesign is awesome though!! I love how the new image upload interface.

js/image/shared/ImageUploader.svelte
```diff
@@ -0,0 +1,214 @@
...
+		on:error
+		disable_click={!!value}
+		{root}
+	>
+		{#if _value === null || streaming}
+			<slot />
+		{/if}
+	</Upload>
+	{#if _value !== null}
+		<img src={_value.path} alt="" />
```
freddyaboulton, line 186: This should be url. Actually I think my comment about `normallise_file` above is wrong then.

gradio/components/image.py
pngwn: cool

gradio/components/image.py
pngwn: No, so the idea (for now at least) is that iit must be exactly 'webcam' for streaming to take effect because i doubt people want to handle streams of image snapshot as well as discreet images. Streaming also won't work with other sources types i think (maybe).

So if the sources == `['webcam', 'upload']` then this error should be raised.

js/image/shared/ImageUploader.svelte
pngwn: Yah, I've fixed this now and also got rid of the `_value` thing, we don't need it now.

js/upload/src/utils.ts
pngwn: I still think it might be wrong, i thought that when i refactored but I didn't wanna change anything. I don't really understand when this would ever be useful, even if it is supported. Unless we use it to set some kind of base path for `sample_dir` in examples or something?

freddyaboulton: I think this might be ok after all. I think images were not being loaded because we were using `url` instead of `path` as the img soruce

gradio/imagetools.py
abidlabs, line 1: Consider calling this file `image_utils.py` for consistency

client/python/gradio_client/data_classes.py
abidlabs: Let's comment to explain this is only here for backwards compatibility. 

Regarding this, why is `data` needed for backwards compatibility, but renaming `name` -> `path` doesn't break backwards compatibility?

client/python/gradio_client/client.py
```diff
@@ -874,6 +874,7 @@ def _predict(*data) -> tuple:
...
                 result = json.loads(response.content.decode("utf-8"))
             try:
+                print(result)
```
abidlabs, line 877: Remove

gradio/components/image.py
abidlabs: We use the first approach in most other places, so maybe we should go with that. But don't feel too strongly.

gradio/components/image.py
```diff
@@ -47,15 +45,16 @@ def __init__(
...
         width: int | None = None,
         image_mode: Literal[
             "1", "L", "P", "RGB", "RGBA", "CMYK", "YCbCr", "LAB", "HSV", "I", "F"
         ] = "RGB",
-        invert_colors: bool = False,
-        source: Literal["upload", "webcam", "canvas"] = "upload",
-        tool: Literal["editor", "select", "sketch", "color-sketch"] | None = None,
+        sources: list[Literal["upload", "webcam", "clipboard"]] = [
+            "upload",
+            "webcam",
```
abidlabs, line 55: Once we create the `ImageEditor` component, it'd be nice to add the `tool` argument back to `Image`, and if a person tries to create an `Image` with a tool, we display a warning and just create the `ImageEditor` component for them. I think it'll help with the migration

abidlabs:
Awesome cleanup of the backend @pngwn! Tested it a few times, love the simplified component. Regarding the UI:

image

* For the upload icon, can we use the same one as in the Audio component for consistency?

image

* I also like the soft horizontal rule on top of the icons in the `Audio` component. For consistency, should we add it for the Image component? Not sure if it'll look good when there's an image uploaded so we should try it and see how it looks

* I like the paste icon a lot, but the webcam icon looks more like video than webcam. Could we use an icon like this instead?

image

* The "X" icon was handy to clear an uploaded image -- it'd be great to have that again for interactive mode.

client/python/gradio_client/client.py
freddyaboulton: Sorry I missed this - what do you mean?

gradio/components/image.py
freddyaboulton: Ah ok. I think it would be better to rewrite as `if streaming and sources != ['webcam']` then!

client/python/gradio_client/client.py
pngwn: This is for backwards compat too, was referring to the above comment as backwards compat being handled by another class.

client/python/gradio_client/data_classes.py
pngwn: @freddyaboulton Can i remove this data field as per the above comment or do I need to add `path` back?

pngwn:
Will address these comments and get this merged to unblock stuff, we can continue to tweak after that if needed.

gradio/components/image.py
pngwn: Sounds good.

client/python/gradio_client/client.py
```diff
@@ -1026,18 +1023,19 @@ def _download_file(
         if isinstance(x, str):
             file_name = utils.decode_base64_to_file(x, dir=save_dir).name
         elif isinstance(x, dict):
-            if x.get("is_file"):
-                filepath = x.get("name")
-                assert filepath is not None, f"The 'name' field is missing in {x}"
+            data = x.get("data", None)
```
freddyaboulton, line 1026: This function is only called on `V4` spaces so it's ok to delete the data check since data will no longer be present in V4 after this is merged.

Don't worry if client tests fail as long as they're connecting to a space on spaces. It's cause they're running an older version of the client.

client/python/gradio_client/data_classes.py
freddyaboulton: This is only used for type hints in the `serializing.py` class, which is only needed for backwards compatibility with v3.

So let's change `path` back to `name` and make it clear this is only used in V3 spaces.

pngwn:
@abidlabs I've added back the clear icon and changed the Upload and Webcam icons. I added a soft border above the icons too. It looked weird with an image so it goes away when an image is uploaded but is there when the image is empty.

Will sort out conflicts and tests and get this merged. Feel free to rereview before or after it is merged and we can address any issues.

abidlabs:
Small UI bug: if the length of the webcam is too long, it appears cut off like this:

image

Otherwise, UI looks super nice! Doing some more testing with complete apps

gradio/components/image.py
```diff
@@ -104,40 +97,34 @@ def __init__(
...
-            )
-        self.source = source
-        if tool is None:
-            self.tool = "sketch" if source == "canvas" else "editor"
-        else:
-            self.tool = tool
-        self.invert_colors = invert_colors
+        valid_sources = ["upload", "webcam", "clipboard"]
+
+        for source in sources:
```
abidlabs, line 115:
```suggestion
        if isinstance(sources, str) and sources in valid_sources:
            sources = [sources]
        for source in sources:
```

gradio/components/image.py
```diff
@@ -104,40 +97,34 @@ def __init__(
...
-                f"Invalid value for parameter `source`: {source}. Please choose from one of: {valid_sources}"
-            )
-        self.source = source
-        if tool is None:
-            self.tool = "sketch" if source == "canvas" else "editor"
-        else:
-            self.tool = tool
-        self.invert_colors = invert_colors
+        valid_sources = ["upload", "webcam", "clipboard"]
+
```
abidlabs, line 114:
```suggestion
        if isinstance(sources, str):
            sources = [sources]
```

gradio/components/image.py
```diff
@@ -104,40 +97,34 @@ def __init__(
...
+        for source in sources:
+            if source not in valid_sources:
+                raise ValueError(
+                    f"Invalid value for parameter `source`: {source}. Please choose any of: {valid_sources}"
```
abidlabs, line 118:
```suggestion
                    f"`sources` must a list consisting of elements in {valid_sources}"
```

abidlabs:
Tested a bunch of demos, and generally everything is looking good. We need to update `source` -> `sources` in some of the demos, such as `kitchen_sink` but that can be done in a separate PR. The only thing that I noticed that wasn't working was examples. 

E.g. 

```py
import gradio as gr

gr.Interface(lambda x:x, gr.Image(), gr.Image(), examples=["cheetah.jpg"]).launch()
```

Clicking on the example image produced a collapsed input image:

image

abidlabs (approved):
Awesome redesign @pngwn! Approving as the few things I noticed can be done in a separate PR

pngwn:
Thanks for the pointers and reviews @abidlabs & @freddyaboulton! Much appreciated.

I've fixed the examples which is the biggest issue, i'll leave the dropdown for another PR and merge this in when CI is good.

pngwn:
I have had to comment out a few tests because we don't have the ImageEditor atm. I'll add those back and fix the demos when ImageEditor has returned.

abidlabs:
Sounds great @pngwn thanks for addressing the review!

js/audio/static/StaticAudio.svelte
```diff
@@ -43,9 +38,9 @@
 	<div class="icon-buttons">
 		{#if show_download_button}
 			<a
-				href={value.data}
```
freddyaboulton, line 46: @pngwn here we're using `url` but in `VideoPreview` we're using `data`. Which is intended?

PR #6098, opened 2023-10-26
Gradio custom component publish

freddyaboulton:
## Description

Still some things I want to iron out but wanted to get feedback as soon as possible.

`gradio cc publish` will guide the user through publishing their custom component. The only pre-req is for the wheel to have been built.

There are two steps - uploading to pypi and uploading a demo to spaces. Neither is required. We use a docker space instead of a gradio space so that we can install the wheel locally in the space (so that pypi upload is not required).

Users can skip the interactive walk-through by providing all the arguments to the command.

Example space here: https://huggingface.co/spaces/freddyaboulton/gradio_datepicker
Pypi here: https://pypi.org/project/gradio-datepicker/

https://github.com/gradio-app/gradio/assets/41651716/11faf416-d4c5-4208-b52a-096f25c44aac

Some things to iron out:

- [x] Let users opt out of pypi publish or hf publish if in interactive mode
- [x] Add the component template to the space README tag

## 🎯 PRs Should Target Issues

## Tests


gradio/cli/commands/components/publish.py
```diff
@@ -0,0 +1,167 @@
...
+    if upload_demo and not demo_dir:
+        panel = Panel(
+            "You have opted to upload to [blue]https://huggingface.co/spaces[/] but no demo directory has been provided.\n"
+            "Please provide the path to the [magenta]demo directory[/] for your custom component.\n"
+            "This directory should contain [magenta]all the files[/] it needs to run successfully.\n"
+            "Please make sure the gradio app is in an [magenta]app.py[/] file.\n"
+            "If you need additional python requirements, add a [magenta]requirements.txt[/] file to this directory."
+        )
+        print(panel)
+        demo_dir = Path(Prompt.ask(":hugging_face: Please enter demo directory"))
```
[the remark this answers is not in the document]
freddyaboulton: if hf_token is None, the hub library will automatically use the one you set during the cli login. If it can't find it, it will raise the appropriate error about not being logged in. So I don't think we need to check ourselves if they are logged in.

freddyaboulton:
Ok I think this is looking good now!

gradio/cli/commands/components/files/README
```diff
@@ -0,0 +1,11 @@
+---
+tags: [gradio-custom-component, gradio-custom-component-template-{{ template }}]
+title: {theme_name}
```
abidlabs, line 3: typo

gradio/cli/commands/components/files/README
```diff
@@ -0,0 +1,11 @@
...
+colorFrom: orange
+colorTo: purple
+sdk: gradio
```
abidlabs, line 6: Confused about this -- why is the sdk gradio if this is a Docker Space? I'm probably missing something

abidlabs: Is this file not the readme in the uploaded Space?

gradio/cli/commands/components/publish.py
```diff
@@ -0,0 +1,232 @@
...
+
+from gradio.cli.commands.components._create_utils import PATTERN_RE
+
+colors = ["red", "yellow", "green", "blue", "indigo", "purple", "pink", "gray"]
+
+PYPI_REGISTER_URL = "https://pypi.org/account/register/"
+
+README_CONTENTS = """
+---
+tags: [gradio-custom-component{template}]
```
abidlabs, line 24: as in the readme above:
```suggestion
tags: [gradio-custom-component, gradio-custom-component-{template}]
```

abidlabs:
Beautiful @freddyaboulton! Just a few suggestions:

-------

If I'm not mistaken, for the other commands (`gradio cc dev` and `gradio cc build`), we assume that we are running from the root of the custom component directory. I think we should assume the same and provide default values for the different options that are presented to the user. 

i.e.
* `gradio cc publish` should automatically look for the wheel in `dist/*.whl` instead of the user having to copy-paste this file. If this file exists, it should confirm with the user, otherwise it should prompt for the path. 
* Similarly, for the path to the demo, it should present the default `demo/` (if it is exists)
* And similarly for the path to the source, it should present `.`

-------

It would be amazing the pypi project description included a short snippet showing how to use the custom component:

image

Maybe this could be the same as the default template?

-------

Nit: my terminal is including the `!` in the highlighted URL:

image

-------

[The created Space](https://huggingface.co/spaces/abidlabs/gradio_richtextbox) doesn't load for me. I get: 

image

The reason is that my demo has examples, which are not being cached successfully. I looked into this, tried updating the [dockerfile to add chmod commands](https://huggingface.co/spaces/abidlabs/gradio_richtextbox2/blob/main/Dockerfile), but it doesn't work. I think the problem is that while we can change the permissions of `gradio_cached_examples`, we can't change the permissions of its subdirectories at launch time. We'd have to change the code in gradio ([probably here](https://github.com/gradio-app/gradio/blob/6b0bb5e6a252ce8c4ef38455a9f56f1dcda56ab0/gradio/flagging.py#L142)) to fix this. But we *should* make this change so that any demos with examples work.

-------

Cool stuff!!

abidlabs:
Also, we should double check that the pypi package can be successfully imported. I'm getting this error on Colab, but it may be because I created the richtextbox using an older beta release:

image

gradio/cli/commands/components/files/README
freddyaboulton: Yea sorry I need to delete this file! It's not used right now.

gradio/cli/commands/components/publish.py
freddyaboulton: If for whatever reason we can't identify the template (because the developer removed the tag from the pyproject.toml) I only want to include the `gradio-custom-component` tag so that's why it's like this. The `{template}` gets replaced with `, gradio-custom-component-TextBox` when we go upload the space.

freddyaboulton:
Thanks for the feedback @abidlabs !

> gradio cc publish should automatically look for the wheel in 

What I did is that rather than specifying a path to a wheel file, the dev will specify the path to the `dist` directory. We will then upload all the `.wheel` and `.tar.gz` files in that directory. The default value is `./dist` so `gradio cc publish` will do the right thing if you call it from within your component directory.

> It would be amazing the pypi project description included a short snippet showing how to import use the custom component

I modified the `gradio cc create` command to add a project README.md file. By default, it has the code to import the package but devs can modify it to include whatever data they want. During publish, this README will be set to the pypi project description.

> Nit: my terminal is including the ! in the highlighted URL:

Fixed!

> [The created Space](https://huggingface.co/spaces/abidlabs/gradio_richtextbox) doesn't load for me

I fixed the Dockerfile! I got richtextbox and an image clone (both using examples) to work. See here: https://huggingface.co/spaces/freddyaboulton/gradio_richtextbox2 and https://huggingface.co/spaces/freddyaboulton/gradio_coolimage

> we should double check that the pypi package can be successfully imported

The issue on Colab is that the custom component package depends on gradio 4.0 which isn't on pypi yet so colab installs the latest 3.0 version of gradio. So it doesn't work. Once gradio 4.0 is out, that won't happen. I installed the development versions of gradio 4.0 in colab and I can load richtextbox: https://colab.research.google.com/drive/1asGFESN6t06ZEdmPbeGCGZh6v1tBgP7i?usp=sharing

abidlabs:
You're amazing @freddyaboulton! Just checking one quick thing

abidlabs:
LGTM @freddyaboulton, thanks for addressing everything! Just two quick things:

> I modified the gradio cc create command to add a project README.md file.

Could we similarly add a blank `requirements.txt` file and reference it in the pyproject.toml file. I think it'll be easier for developers who are creating custom components if there's already a blank requirements.txt file they just need to update

> I'm starting to see this error again with `gradio cc build`:

image

My guess is that its not related to your changes but something else is going on. We can fix it in another PR but just flagging here (do you see the issue on your end?)

abidlabs (approved):

gradio/cli/commands/components/publish.py
```diff
@@ -0,0 +1,247 @@
...
+        if upload_demo:
+            panel = Panel(
+                "Please provide the path to the [magenta]demo directory[/] for your custom component.\n\n"
+                "This directory should contain [magenta]all the files[/] it needs to run successfully.\n\n"
+                "Please make sure the gradio app is in an [magenta]app.py[/] file.\n\n"
+                "If you need additional python requirements, add a [magenta]requirements.txt[/] file to this directory."
+            )
+            print(panel)
+            demo_dir_ = Prompt.ask(
+                f":roller_coaster: Please enter the path to the demo directory. Leave blank to use {(Path('.') / 'demo')} directory"
```
abidlabs, line 154:
```suggestion
                f":roller_coaster: Please enter the path to the demo directory. Leave blank to use {(Path('.') / 'demo')}/ directory"
```

abidlabs, line 154: Wasn't clear that this was a path:

image

Perhaps:

```suggestion
                f":roller_coaster: Please enter the path to the demo directory. Leave blank to use: {(Path('.') / 'demo')}"
```

abidlabs:
please ignore the error screenshot, can't repro anymore

freddyaboulton:
@abidlabs and I agreed we're going to keep the dependencies in the toml file but we'll call it out in the docs!

Thanks @abidlabs @pngwn!

PR #6112, opened 2023-10-26
fix build

pngwn:
## Description

Closes: #(issue)

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/js/preview/src/build.ts
+++ b/js/preview/src/build.ts
@@ -35,6 +35,8 @@ export async function make_build({
 			for (const [entry, path] of exports) {
 				try {
 					const x = await build({
+						root: source_dir,
+						configFile: false,
 						plugins: [
 							...plugins,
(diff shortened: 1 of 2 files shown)


js/preview/src/build.ts
```diff
@@ -35,6 +35,8 @@ export async function make_build({
 			for (const [entry, path] of exports) {
 				try {
 					const x = await build({
+						root: source_dir,
```
pngwn, line 38: So weird thing but we don't actually use `root` at all because we are building in library mode, but some plugins _do_ use `root` as a reference point. Because we weren't setting it, it defaults to the current working directory which is one level higher than we need to be to find a package.json, since we run everything from the component_dir root.

freddyaboulton (approved):
Nice @pngwn 🔥 Confirmed the fix!!

PR #6118, opened 2023-10-27
Improve Video Component

hannahblair:
## Description

Closes: #(issue)

## 🎯 PRs Should Target Issues

## Tests


abidlabs:
For the `Video` component, I'd suggest the default source that is shown when all sources are selected is actually the file upload instead of the webcam, as people might be started to see their own webcam feed

image

abidlabs:
By the way @hannahblair just to confirm, the issue with the output file not showing up only happens after you've trimmed the video, right? I don't see any issue with untrimmed video files. Also, I do see an issue where the video after being trimmed shows up all black on the input side:

image


My test code is just:

```py
import gradio as gr

gr.Interface(lambda x:x, "video", "video").launch()
```

abidlabs:
Nice! So the output video *is* showing up now, however both the input and output videos are completely black (the audio is there though):

image

abidlabs:
I tried this with a few different .mp4 files and they are all giving the same result

abidlabs:
Looks good @hannahblair! Two final things I noticed:

* As mentioned on Slack, it takes quite a while to trim videos for me (it might be my network speed), and its not clear that the video is still being trimmed, perhaps we can show visual indicator while video is being trimmed?

* The webcam recording isn't working for me. Its passing in a blank value as far as I can tell. 

e.g. when I launch:

```py
import gradio as gr

gr.Interface(lambda x:x, "video", "video").launch()
```

And record something with the webcam, the output video is blank

gradio/components/video.py
```diff
@@ -159,8 +173,17 @@ def preprocess(self, x: dict | VideoData) -> str | None:
         file_name = Path(data.video.path)
         uploaded_format = file_name.suffix.replace(".", "")
         needs_formatting = self.format is not None and uploaded_format != self.format
-        flip = self.source == "webcam" and self.mirror_webcam
+        flip = self.sources == ["webcam"] and self.mirror_webcam
+        duration = processing_utils.get_video_length(file_name)
 
+        if self.min_length is not None and duration < self.min_length:
+            raise ValueError(
```
abidlabs, line 180: We should raise a `gr.Error()` instead of a `ValueError` for these so that they are shown to users trying out the app. Otherwise, they'd be hidden away in the terminal. Same for below, and I would do the same in `audio.py`

abidlabs: Let's double check it works 😄

abidlabs:
LGTM looks great @hannahblair! 

The only thing left is the webcam issue, but since that's broken on `v4` aswell, we could merge this PR in, and fix that in a separate PR. (To reduce conflicts as we have some big juicy refactor PRs coming)

Issue #6119, opened 2023-10-27
Model3D panning and UX improvements


dylanebert:
Changes for improving Model3D
- Enable panning support
- Adapt zoom and pan sensibility to the camera radius, creating a more consistent experience across various model sizes
- Set lower radius to prevent zoom inversion (zooming below 0)

PR #6120, opened 2023-10-27
model3D panning, improved UX

dylanebert:
## Description

Improvements to the Model3D components
- Enabled panning
- Adapt pan and zoom speed to scene radius, creating a consistent experience across various model sizes
- Set lower radius limit to prevent zoom inversion (zooming in below 0)

Closes: #6119

### DIFF
--- a/js/model3D/shared/utils.ts
+++ b/js/model3D/shared/utils.ts
@@ -20,7 +20,7 @@ const create_camera = (
 	}
 	helperCamera.lowerRadiusLimit = 0.1;
 	const updateCameraSensibility = (): void => {
-		helperCamera.wheelPrecision = (150 * zoom_speed) / helperCamera.radius;
+		helperCamera.wheelPrecision = 250 / (helperCamera.radius * zoom_speed);
 		helperCamera.panningSensibility = (10000 * pan_speed) / helperCamera.radius;
 	};
 	updateCameraSensibility();
--- a/js/model3D/shared/utils.ts
+++ b/js/model3D/shared/utils.ts
@@ -4,7 +4,8 @@ import * as BABYLON from "babylonjs";
 const create_camera = (
 	scene: BABYLON.Scene,
 	camera_position: [number | null, number | null, number | null],
-	zoom_speed: number
+	zoom_speed: number,
+	pan_speed: number
 ): void => {
 	scene.createDefaultCamera(true, true, true);
 	var helperCamera = scene.activeCamera! as BABYLON.ArcRotateCamera;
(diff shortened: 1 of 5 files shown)


abidlabs:
Thank you so much @dylanebert! Two quick things:

* Would you be able to target the `v4` branch instead of `main`? We're about to merge in `v4` which will lead to quite a bit of conflicts with this PR.
* We used to have panning, but it degraded the experience on mobile. Have you tested this out on mobile by any chance?

dylanebert:
Sure!

I have not tested panning in this PR but I am using the exact same library/camera in https://huggingface.co/spaces/dylanebert/igf
and it works well on mobile. I believe the issues may have come from varying model size and the speed not scaling with radius.

js/model3D/shared/utils.ts
```diff
@@ -17,11 +18,14 @@ const create_camera = (
...
 		helperCamera.radius = camera_position[2];
 	}
-	// Disable panning. Adapted from: https://playground.babylonjs.com/#4U6TVQ#3
-	helperCamera.panningSensibility = 0;
-	helperCamera.attachControl(false, false, -1);
-	helperCamera.pinchToPanMaxDistance = 0;
-	helperCamera.wheelPrecision = 2500 / zoom_speed;
+	helperCamera.lowerRadiusLimit = 0.1;
+	const updateCameraSensibility = (): void => {
+		helperCamera.wheelPrecision = (150 * zoom_speed) / helperCamera.radius;
```
abidlabs, line 23: I believe increasing the wheel precision actually decreases the zoom speed: https://forum.babylonjs.com/t/zoom-sensitivity/10338/3, which is why I had it in the denominator before. What's the reason for the change?

abidlabs:
The panning behavior looks good @dylanebert but I have a question about the changes to the zooming behavior above.

js/model3D/shared/utils.ts
abidlabs: When I test these changes using the `demo/model3D/run.py` demo, I'm finding it bit too sensitive for me

dylanebert: Thank you for your review! I agree with you concerns and adjusted the zoom computation.

Zoom speed is now correctly in the denominator, and I've set the speed to 250, which seems to approximately reflect that of [sketchfab](https://sketchfab.com/3d-models/deadly-melodic-d60f5ca4c0bf4794a5dee82a040ce99f) on my machine.

abidlabs:
In principle, these changes look good @dylanebert, but since we merged `v4` into `main`, the git histories have messed up. Would you be able to open a new PR targeting `main` with these changes?

Sorry for the back-and-forth, but we should be able to get this merged in pretty quickly after its targeting `main`.

dylanebert:
Sure, no problem.

Issue #5401, opened 2023-09-02
Support decorator event listener listener syntax for Blocks.load


aliabid94:
This should be done alongside deprecating the class method use of Blocks.load

abidlabs:
Also we should support `gr.on` with `Blocks.load()` as pointed out by @hysts

abidlabs:
Fixed in `v4`

PR #6126, opened 2023-10-28
Refactor `Blocks.load()` so that it is in the same style as the other listeners

abidlabs:
Decorator-based events was actually broken on `v4`.

This fixes that and fixes: #5401, which means `demo.load` works with decorator-based syntax:

Test with:

```py
import gradio as gr

with gr.Blocks() as demo:
    
    @demo.load(outputs=name)
    def test():
        return "Abubakar"

demo.launch()
```

It also works with `gr.on`:

```py
import gradio as gr

with gr.Blocks() as demo:
    name = gr.Textbox(value="Abubakar")
        
    gr.on([demo.load, name.submit], lambda x:x, name, output)

demo.launch()
```

### DIFF
--- a/gradio/blocks_events.py
+++ b/gradio/blocks_events.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from gradio.component_meta import create_or_modify_pyi
 from gradio.events import EventListener, Events
 
--- a/gradio/blocks_events.py
+++ b/gradio/blocks_events.py
@@ -8,14 +8,12 @@ BLOCKS_EVENTS: list[EventListener | str] = [Events.load]
 
 class BlocksMeta(type):
     def __new__(cls, name, bases, attrs):
-        new_events = []
         for event in BLOCKS_EVENTS:
             trigger = (
                 event
                 if isinstance(event, EventListener)
                 else EventListener(event_name=event)
             ).copy()
-            new_events.append(trigger)
             trigger.set_doc(component=name)
             attrs[event] = trigger.listener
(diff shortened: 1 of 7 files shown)


abidlabs:
So getting typechecking to work was quite tricky. I ended up going with the same "*.pyi" magic that @freddyaboulton implemented for the components, but instead of stubbing the `Blocks` class, which was leading to lots of issues, I added a parent class of `Blocks` called `BlocksEvents` and stubbed that instead. Its a bit more verbose but it gets the job done, and now all of our events follow the same syntax

gradio/blocks_events.py
```diff
@@ -0,0 +1,25 @@
...
+                else EventListener(event_name=event)
+            ).copy()
+            trigger.set_doc(component=name)
+            attrs[event] = trigger.listener
+        component_class = super().__new__(cls, name, bases, attrs)
+        create_or_modify_pyi(BlocksEvents, "BlocksEvents", BLOCKS_EVENTS)
+        return component_class
+
+
+class BlocksEvents:
```
freddyaboulton, line 24: I would add a docstring here explaining why we have this even though it's empty

freddyaboulton (approved):
Looks good to me @abidlabs ! Thanks for the clean up.

Don't know why chromatic is failing. I can run the failing command fine locally

image

abidlabs:
Thanks @freddyaboulton for the review. I don't see the chromatic issue anymore -- the check isn't running because presumably there are no frontend changes to any components. I'll go ahead and merge this in.

PR #6127, opened 2023-10-28
Use data models in `preprocess()`

abidlabs:
As [discussed internally](https://huggingface.slack.com/archives/C02QNREU24S/p1698449902074779?thread_ts=1698448727.887339&cid=C02QNREU24S) with @freddyaboulton, this uses data models in the `.preprocess()`, not just the `.postprocess()`

This PR makes the following changes:
* Removes the `serializes` function: https://github.com/gradio-app/gradio/blob/v4/gradio/component_meta.py#L59 as FastAPI automatically converts Pydantic models into JSON
* Pre-Preprocesses incoming payload from the frontend to convert to Pydantic data models before sending to `.preprocess()`
* Cleans up the code as a result of the above changes, adds typing to missing places, etc. 
* Rename `x` to `payload` in preprocess() to signify that its the payload from the frontend
* Rename `y` to `value` in postprocess() to signify its the same type as the `value` of the component
* Removes docstrings from preprocess() and postprocess() because many were incorrect / outdated / rudimentary. The typing is better than the docstrings.

I think it would be better to merge this after @pngwn's #6094 PR to avoid introducing many backend conflicts to that PR.

Closes: #1798


freddyaboulton:
Nice 🎉

abidlabs:
Working on this, but so many tests need to be updated. Might take some time ❗

abidlabs:
Rebased to main in #6176

Issue #4609, opened 2023-06-21
The directory containing the gradio python file should not be served by default.


aeciorc:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
Currently, the default behaviour for the app is to serve any files in the directory of the Gradio script. This is in the docs and is not considered a bug or a vulnerability, but I think it's problematic for a few reasons:

1. Users want to share their app, but not necessarily the code. Gradio is used not only for open-source projects, but also to share and get feedback on MVPs, research projects, work projects, etc. Even for open-source models, people might want to  protect their checkpoints and datasets. Giving access to the project directory by default might catch them by surprise, which brings me to my next point.

2. Is the default behaviour expected by most users? I'm not sure it is. Academic or non-technical users might not know or think about this at all, while developers expect a backend to serve the files they explicitly asked it to serve (definitely not the backend code itself.)

3. It's not crazy to assume that some users, especially beginners, will put gradio files not in a project folder, but in a higher directory in their system, like ~/Documents, ~/Downloads, or even worse, the home directory. With the current behaviour, this could cause them a lot of problems.

**Describe the solution you'd like**  
The solution is simple: don't serve the script directory files by default. Raw outputs are always fine, and Gradio already handles filepath outputs with tempfiles. This is good enough for 99% of use cases.

If a request is made for a file that's not accessible, an error log can link to the section of the documentation about allowed_paths.

**Additional context**

abidlabs:
Hi @aeciorc thanks for creating the detailed issue. Agreed I've been thinking we should do something about this, perhaps 4.0 is a good time to make this change, we'll look into it!

In the meantime, you can use the `blocked_paths` parameter in `launch()` to block specific directories or files.

aeciorc:
Sounds good @abidlabs. I can give this a shot if you need a hand.

abidlabs:
Thanks @aeciorc it needs some thinking through I think. The reason is that we allow serving files from the working directory because it allows users to conveniently serve example files or include default values. For example:

```py
with gr.Blocks():
  gr.Image("lion.jpg")
```

or 

```py
gr.Interface(lambda x:x, "image", "image", examples=["lion.jpg"])
```

We'd need to make sure these use cases are not disrupted. I think it is possible by copying over the referenced files to the temporary directory and serving from there. We may already do this for some components but we'd need to make sure all of the components handle this correctly.

aeciorc:
[...]

PR #6128, opened 2023-10-28
Don't serve files in working directory by default

### LINKED ISSUE #5407: Let `allowed_paths` allow access to dotfiles
### Describe the bug

Using Automatic's Stable Diffusion WebUI, Gradio looks broken and does not function properly. After a helpful comment -- https://github.com/AUTOMATIC1111/stable-diffusion-webui/issues/13027#issuecomment-1704377186 -- in an issue I opened yesterday, it looks like having a dot in the file path was causing this.

So for example: if Gradio is located in `/home/john/.local/` (notice the dot) Gradio will not function properly. Moving the project to `/home/john/` made the issue go away. I am just assuming this is a Gradio related issue and not something else.

I did manually upgrade to version 3.42.0. Same issue.

### Reproduction

- Have a dot in the file path
- Gradio does not function properly

### Screenshot

image

### Logs

_No response_

### System Info

```shell
https://github.com/AUTOMATIC1111/stable-diffusion-webui/files/12505964/sysinfo-2023-09-03.txt
```

### Severity

I can work around it

abidlabs:
Fixes: #4609

I think this is a good change to make. Files returned by a user function work fine (because they are copied to the cache), but examples with media files currently don't work. I think we should just copy them over to the cache too, or perhaps allow access to any example files.

Also let's fix #5407 while we're at it


abidlabs:
It turns out I need to make some changes to the way files are copied to cache to make this work, which depends on #6094, so I'll pick this back up after that PR is in

PR #6135, opened 2023-10-29
Fix selectable prop in the backend

freddyaboulton:
## Description

The selectable prop was not properly being set in the config in the backend.

This fixes that. We need to add `selectable` to the init for `gr.load` to work. This is a bug on main actually. If you `gr.load("spaces/gradio/tictactoe")`, the dataframe's selectable prop will be `False`

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/hungry-spiders-cross.md
+++ b/.changeset/hungry-spiders-cross.md
@@ -1,4 +1,9 @@
 ---
+"@gradio/chatbot": minor
+"@gradio/file": minor
+"@gradio/highlightedtext": minor
+"@gradio/image": minor
+"@gradio/label": minor
 "gradio": minor
 ---
 
--- a/.changeset/hungry-spiders-cross.md
+++ b/.changeset/hungry-spiders-cross.md
@@ -1,0 +1,5 @@
+---
+"gradio": minor
+---
+
+feat:Fix selectable prop in the backend
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -206,7 +206,7 @@ class Block:
         for e in self.events:
             to_add = e.config_data()
             if to_add:
-                config = {**config, **to_add}
+                config = {**to_add, **config}
         config.pop("_skip_init_processing", None)
         config.pop("render", None)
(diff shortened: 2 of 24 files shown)


pngwn (approved):
Looks good! Thanks @freddyaboulton!

abidlabs:
Thanks @freddyaboulton! Thoughts on renaming this to be `_selectable` to indicate that it should not be set by the user? I was planning on doing the same thing with `root_url` -> `_remote_url`

pngwn:
>  I was planning on doing the same thing with root_url -> _remote_url

In the frontend i changed `root_url` to `proxy_url`. Maybe that naming can be consistent, whatever we decide.

abidlabs:
`_proxy_url` works for the backend!

freddyaboulton:
Yep can rename to _selectable!

freddyaboulton:
@abidlabs I think the name of the python property has to match the name used in the frontend. Are you also planning on renaming `root` to `_proxy_url` in the frontend?

abidlabs:
I thought @pngwn had already renamed it to `proxy_url`. If needed, I can rename to `_proxy_url` yeah

freddyaboulton:
Ok will rename to _selectable in the front end as well!

pngwn:
I don't think we need to do the underscore thing in the frontend. It's a bit weird and doesn't really serve a purpose. But if it's easier to pass things down, I'm not against it.

freddyaboulton:
Agree it's unconventional. But it's easier to pass down. Right now, the name of the parameter in the init, the name of the property it's stored as in the class, and the name used store it in the config have to match. Would be nice to not have a layer somewhere doing the renaming between front-end name and backend name.

abidlabs (approved):
LGTM @freddyaboulton!

abidlabs:
Let me fix the test real quick

abidlabs:
Actually sorry I think this needs to be updated (or removed altogether):

https://github.com/gradio-app/gradio/blob/f82c8840326b13936481250a204ab067225c3fcf/gradio/events.py#L490

Right now, the config of a component has both the `selectable` and `_selectable` attributes

PR #6140, opened 2023-10-30
Fix video

abidlabs:
Basically we should be using `video.url`, not `video.path`

I'll go ahead and merge it in since I need it for #6128, but cc-ing @pngwn just in case I misunderstood anything 

Test code:

```py
import gradio as gr

with gr.Blocks() as demo:
    gr.Video("test.mp4")
    gr.Interface(lambda x:x, "video", "video", examples=["test.mp4"])

demo.launch()
```

or run `python demo/kitchen_sink/run.py`


js/video/shared/VideoPreview.svelte
```diff
@@ -65,9 +65,9 @@
 	{/key}
 	<div class="icon-buttons" data-testid="download-div">
 		<a
-			href={value.path}
+			href={value.url}
 			target={window.__is_colab__ ? "_blank" : null}
-			download={value.orig_name || value.path}
+			download={value.orig_name || value.url}
```
freddyaboulton, line 70: I think this should just be value.url like it is for `StaticAudio`?

freddyaboulton (approved):
Noticed this too. Looks good to me!

js/video/shared/VideoPreview.svelte
abidlabs: Looked this up because I wasn't sure. From https://www.w3schools.com/tags/att_a_download.asp:

> value of the download attribute will be the new name of the file after it is downloaded

So I actually think `value.orig` is the appropriate one. I'm not sure if the latter parameter is ever used, but if anything, it should probably stay `value.path`. I'll let @hannahblair change this if needed.

abidlabs:
Thanks @freddyaboulton for the quick review!

pngwn:
Yep this is all correct, missed this when doing the images changes. Not sure why ts:check didn't fail though.

js/video/shared/VideoPreview.svelte
freddyaboulton: Yea read into it and it's the name the file will be given when downloaded. So `orig_name` or `path` is the way to go.

PR #6155, opened 2023-10-30
Moves `gradio_cached_folder` inside the gradio temp direcotry

abidlabs:
This should fix the issue where the cached example files were not showing up, and also doesn't pollute the user's working directory with cached examples, which are really an internal gradio resource. 

cc @hannahblair @freddyaboulton 

Will test in @hannahblair's PR to confirm

### DIFF
--- a/gradio/helpers.py
+++ b/gradio/helpers.py
@@ -20,6 +20,7 @@ import numpy as np
 import PIL
 import PIL.Image
 from gradio_client import utils as client_utils
+from gradio_client.client import DEFAULT_TEMP_DIR
 from gradio_client.documentation import document, set_documentation_group
 from matplotlib import animation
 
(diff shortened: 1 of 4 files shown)


gradio/helpers.py
```diff
@@ -33,7 +34,7 @@
 if TYPE_CHECKING:  # Only import for type checking (to avoid circular imports).
     from gradio.components import Component
 
-CACHED_FOLDER = "gradio_cached_examples"
```
freddyaboulton, line 36: Yep - makes sense to me 👍

abidlabs:
Actually I was thinking about this, one downside of this, that @radames and others have brought up is that they actually like to upload the `gradio_cached_examples` directory along with their app to Spaces to speed up launch times. I'll revert but just expose the `gradio_cached_examples` directory

abidlabs:
Confirmed the fix on the `video-improvements` branch, so I'll merge this in.

Issue #4733, opened 2023-06-29
Make output components not editable if they are being updated


pseudotensor:
The show_progress functionality is nice.  However, for radio, text boxes, etc. the options don't seem to be best.

* "full" totally hides the underlying content, leading to UI/UX confusion when the elements disappear entirely during progress.
* "minimal" still allows one to click on the element, which is also confusing UX-wise since we are updating that element's values perhaps, and we don't want the user to yet again try to switch to something
* "hidden" leaves the original text in the textbox visible until there is completion, which is confusing UX-wise since it's editable while it's being updated.

One could add more .then() steps like to make a radio or text box interactive=False and then use "minimal", but this ends up being some ugly code.  I wonder if it's easy for gradio team to implement this directly?  So new option would be:

* "non_interactive": Element would be clearly visible, so no odd context switches, and it would  not be changable, so no UX confusions.  It would be like when interactive=False, a bit dim and not changable/clickable/etc.

I think this would make the most sense in many situations.  Thanks!

abidlabs:
Good point, why are output components even editable if they are being updated? I think we should just make them non-interactive while the function is running. We can make this change as part of 4.0 along with the work we're doing in #1592  and #1954

cc @dawoodkhan82

pseudotensor:
I'm not sure you can just always make them non-interactive.  The element can queue a next action, which is allowable even if I don't use things that way.

I think it has to be an option.

abidlabs:
It is allowable, but should it be? I think it would usually be better to have two separate components if one needs a component to be both an input and output at the same time

I'll try to think of some concrete use cases to see if this would be too much of a breaking change

pseudotensor:
@abidlabs Do you know of a temporary hack to work-around this so every output becomes non-interactive?  Just don't want to write so much code to do the switch-a-roo myself :)  Thanks!

abidlabs:
Good question @pseudotensor, I don't think there's a simple hack unfortunately. You'd have to do add `.update(interactive=False)` to your function return and do a `.then()` to set the interactivity to True. We have done this within the Gradio code base itself in a couple of places so afaik there's no workaround

abidlabs:
Fixed in `v4`!

abidlabs:
Btw @pseudotensor we are planning on reverting this behavior as it introduced a lot of undesired behavior. For example, a dropdown whose choices were being updated would lose focus when it became static, and so you would not be able to dynamically update a dropdown's choices 

[...]

PR #6157, opened 2023-10-30
Make output components not editable if they are being updated

dawoodkhan82:
## Description
Make output components not editable if they are being updated

Closes: #4733

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/js/app/src/Blocks.svelte
+++ b/js/app/src/Blocks.svelte
@@ -429,6 +429,10 @@
 				})
 				.on("status", ({ fn_index, ...status }) => {
 					tick().then(() => {
+						const outputs = dependencies[fn_index].outputs;
+						outputs.forEach((id) => {
+							instance_map[id].props.interactive = status.stage === "pending" ? false : true;
+						});
 						//@ts-ignore
 						loading_status.update({
 							...status,
--- a/.changeset/shy-bugs-sneeze.md
+++ b/.changeset/shy-bugs-sneeze.md
@@ -1,0 +1,6 @@
+---
+"@gradio/app": minor
+"gradio": minor
+---
+
+feat:Make output components not editable if they are being updated


js/app/src/Blocks.svelte
```diff
@@ -429,6 +429,10 @@
 				})
 				.on("status", ({ fn_index, ...status }) => {
 					tick().then(() => {
+						const outputs = dependencies[fn_index].outputs;
+						outputs.forEach((id) => {
+							instance_map[id].props.interactive = status.stage === "pending" ? false : true;
```
freddyaboulton, line 434: Not every component has an interactive prop (gallery). Is this going to work?

abidlabs:
Works great in my tests! 

Tested with:

```py
import gradio as gr
import time

def test(x):
    time.sleep(3)
    return x

with gr.Blocks() as demo:
    with gr.Row():
        d1 = gr.Dropdown(choices=["a", "b"])
        d2 = gr.Dropdown(choices=["a", "b"])
        d3 = gr.Dropdown(choices=["a", "b"])
    
    d1.change(test, d1, d2, show_progress="minimal")
    d2.change(test, d2, d3, show_progress="minimal")
    
demo.launch()
```

and

```py
import gradio as gr
import time

def test(x):
    time.sleep(3)
    return x

with gr.Blocks() as demo:
    with gr.Row():
        d1 = gr.Dropdown(choices=["cat.jpg", "cheetah.jpg"], multiselect=True)
        d2 = gr.Gallery()
    
    d1.change(test, d1, d2, show_progress="minimal")
    
demo.launch()
```

I guess if the prop doesn't exist, there's no harm in setting it.

abidlabs (approved):

abidlabs:
@dawoodkhan82 I'll merge this in so that we have more time to try it out as we work on the `v4` branch

freddyaboulton:
@dawoodkhan82 This makes all output components interactive after the prediction finishes!

I think what we need to do is set it to `false` as the prediction is running but set it back to the original state after it finishes

dawoodkhan82:
@freddyaboulton oh yikes, you're right. Will fix!

PR #6158, opened 2023-10-30
Queue concurrency count

aliabid94:
Allows setting concurrency count to limit the number of how many of a single event can be running at once in the queue. Also add concurrency_id to set limits for groups. For example:

```python
        with gr.Blocks() as demo:
            a = gr.Number()
            b = gr.Number()
            output = gr.Number()

            add_btn = gr.Button("Add")

            @add_btn.click(inputs=[a, b], outputs=output, concurrency_limit=2)
            def add(x, y):
                time.sleep(4)
                return x + y


            mul_btn = gr.Button("Multiply")

            @mul_btn.click(
                inputs=[a, b],
                outputs=output,
                concurrency_limit=2,
                concurrency_id="muldiv",
            )
            def mul(x, y):
                time.sleep(4)
                return x * y

            div_btn = gr.Button("Divide")

            @div_btn.click(
                inputs=[a, b],
                outputs=output,
                concurrency_limit=2,
                concurrency_id="muldiv",
            )
            def div(x, y):
                time.sleep(4)
                return x / y
```

Estimation calculation is now a lot harder - will tackle this in another PR.


gradio/blocks.py
```diff
@@ -340,6 +344,10 @@ def __init__(
         self.preprocess = preprocess
         self.postprocess = postprocess
         self.tracks_progress = tracks_progress
+        self.concurrency_limit = concurrency_limit
+        self.concurrency_id = concurrency_id or id(fn)
```
abidlabs, line 348: Nice -- so all events that call the same function already share the same concurrency_id

gradio/blocks.py
```diff
@@ -852,15 +864,21 @@ def set_event_trigger(
         _, progress_index, event_data_index = (
             special_args(fn) if fn else (None, None, None)
         )
+        if concurrency_limit is None and utils.get_space() is not None:
+            concurrency_limit = 1
```
freddyaboulton, line 868: I don't think we should do this. It means only one event can be pulled off the queue in spaces even if concurrency_count is set by the space author to be greater than 1?

gradio/blocks.py
```diff
@@ -797,6 +807,8 @@ def set_event_trigger(
             trigger_after: if set, this event will be triggered after 'trigger_after' function index
             trigger_only_on_success: if True, this event will only be triggered if the previous event was successful (only applies if `trigger_after` is set)
             trigger_mode: If "once" (default for all events except `.change()`) would not allow any submissions while an event is pending. If set to "multiple", unlimited submissions are allowed while pending, and "always_last" (default for `.change()` event) would allow a second submission after the pending event is complete.
+            concurrency_limit: If set, this this is the maximum number of this event that can be running simultaneously. Extra events triggered by this listener will be queued. On Spaces, this is set to 1 by default.
```
abidlabs, line 810: Can't remember, why are we setting this to be 1 only on Spaces? I think it makes sense to set it to 1 locally (since you could be using a GPU-based function locally too) and better to have a unified experience locally and on Spaces

gradio/blocks.py
abidlabs: My suggestion (related to my previous comment about Spaces) would be to set `concurrency_limit` to be `1` by default everywhere -- if you set it to `None`, it becomes unlimited (or the same as `concurrency_count`

abidlabs:
LGTM @aliabid94 this works really nicely! I just have a note about setting the default `concurrency_limit` to 1 everywhere

abidlabs (approved):

gradio/blocks.py
aliabid94: Agree with having a unified experience b/w local and Spaces, I'm just thinking there may be unintended consequences of setting concurrency limit to 1 by default. wdyt @freddyaboulton ?

abidlabs: The main one that comes to my mind is if that there's someone calling an external API, they are now limited by `concurrency_limit` even if they set a high value of `concurrency_count`. But this is in my view a bigger deal when hosting on Spaces. When you're running things locally, having a low `concurrency_count` shouldn't be a huge problem. And anyways, they can always set `concurrency_count` to `None`

freddyaboulton: I think we're introducing a really big change here if we set `concurrency_limit` to be 1 everywhere by default.

`gr.Interface(....).queue(concurrency_count=20)` would only run one event at a time. That seems confusing to me.

I think we should make `concurrency_limit` opt in?

abidlabs: Let's add the `concurrency_limit` parameter to `Interface, which applies it to the main submit function. It would have the equivalent effect of setting `.queue(concurrency_count=...)` but now, we don't really use `queue()` anymore anyways. 

> I think we should make concurrency_limit opt in?

The problem with making it opt in is that most machine learning functions consume a lot of GPU RAM, so having `concurrency_count > 1` by default will lead to lots of OOMs

aliabid94: > The problem with making it opt in is that most machine learning functions consume a lot of GPU RAM, so having concurrency_count > 1 by default
Essentially concurrency_count was always > 1, and we never had issues with it.

aliabid94: I think this is still confusing

abidlabs: > Essentially concurrency_count was always > 1 before, and we never had issues with it.

Its true that `concurrency_count > 1` locally, and we never had an issue but that's because locally, you don't have to worry about multiple users running the demo simultaneously. It only becomes a problem when you host the demo on Spaces or elsewhere. We used to have `concurrency_count > 1` on Spaces as well, but @hysts pointed out that several of his demos were not working, and we set `concurrency_count = 1.`

freddyaboulton: I think setting `concurrency_limit` to 1 for everything is still too aggressive. I hear you about heavy events causing OOM but most events are "cheap" and can be run concurrently. On spaces, you'll have to remember to set `concurrency_count` > 1 for all other events which doesn't sound great.

Maybe we can set it to 1 by default for interfaces. Also, we should support `-1` or something to signify that the concurrency should not be limited (or to limit it to `concurrency_count`)

aliabid94: I think I agree with @freddyaboulton that things are getting confusing now, especially with:

> gr.Interface(....).queue(concurrency_count=20) would only run one event at a time. That seems confusing to me

How about we also deprecate concurrency_count entirely. If a user tries to set it, then we raise an error and say, set the concurrency_limit directly on the event listener. We then use `max_threads` from launch to determine the max number of workers for  the queue.

As for what the default concurrency_limit for an event should be, I can see it both ways.  

> I hear you about heavy events causing OOM but most events are "cheap" and can be run concurrently.

But if they are cheap (and fast), then they won't block each other even if the concurrency_limit is 1. 

Thoughts?

abidlabs: I'm fine with this!

aliabid94:
This PR got in a weird state for me after v4 merged into main - deleting and opening a new PR

PR #6159, opened 2023-10-31
Clean root url

abidlabs:
Makes a few cleanups related to the `root_url` parameter. 

* Removes it from every component. 
* Does the same change to `_skip_init_processing` and `_selectable`
* Where present, it renames it to `proxy_url` (since it is no longer present in every component, we don't have to use the underscore notation to designate internal use) -- I just kept it the same as the frontend for simplicity

Closes: #5830

The main goal here is to simplify the creation of custom components


gradio/blocks.py
```diff
@@ -196,6 +196,13 @@ def get_block_name(self) -> str:
     def get_expected_parent(self) -> type[BlockContext] | None:
         return None
 
+    def set_extra_params(self, **kwargs):
```
freddyaboulton, line 199: Is this meant to be used anywhere?

gradio/blocks.py
```diff
@@ -207,9 +214,11 @@ def get_config(self):
             to_add = e.config_data()
             if to_add:
                 config = {**to_add, **config}
-        config.pop("_skip_init_processing", None)
         config.pop("render", None)
-        return {**config, "root_url": self.root_url, "name": self.get_block_name()}
+        config = {**config, "proxy_url": self.proxy_url, "name": self.get_block_name()}
+        if (_selectable := getattr(self, "_selectable", None)) is not None:
```
freddyaboulton, line 219: The := is not valid until 3.10 I think.

I think this looks good, I don't like treating some props as special but I guess we can make an exception here since they're pretty opaque parameters.

I think we need to change "selectable" to "_selectable" in the event callback so that it gets set

```python
    select = EventListener(
        "select",
        callback=lambda block: setattr(block, "_selectable", True),
        doc="Event listener for when the user selects or deselects the {{ component }}. Uses event data gradio.SelectData to carry `value` referring to the label of the {{ component }}, and `selected` to refer to state of the {{ component }}. See EventData documentation on how to use this event data",
    )
```

abidlabs: https://www.geeksforgeeks.org/walrus-operator-in-python-3-8/ looks like Python 3.8 should support walrus

abidlabs: Ah interesting, I thought it was set but I'll confirm

gradio/blocks.py
abidlabs: Whoops extraneous

gradio/blocks.py
freddyaboulton: > looks like Python 3.8 should support walrus

nice

abidlabs: Done

abidlabs:
@freddyaboulton does this look good to merge in?

freddyaboulton:
Taking a look!

abidlabs:
Thank you so much for reviewing @freddyaboulton!

PR #6165, opened 2023-10-31
Fixes remaining backend tests

abidlabs:
Fix all remaining backend tests.


gradio/blocks.py
```diff
@@ -618,24 +618,24 @@ def get_block_instance(id: int) -> Block:
...
-
             # If a Gradio app B is loaded into a Gradio app A, and B itself loads a
             # Gradio app C, then the proxy_urls of the components in A need to be the
             # URL of C, not B. The else clause below handles this case.
             if block_config["props"].get("proxy_url") is None:
-                block_proxy_url = f"{proxy_url}/"
-            else:
-                block_proxy_url = block_config["props"]["proxy_url"]
-                proxy_urls.add(block_proxy_url)
+                block_config["props"]["proxy_url"] = f"{proxy_url}/"
```
abidlabs, line 625: Slightly refactored to this code so that `proxy_url` would get passed into `gr.Dataset`

gradio/components/dataset.py
```diff
@@ -57,6 +58,7 @@ def __init__(
             container: If True, will place the component in a container - providing some extra padding around the border.
             scale: relative width compared to adjacent Components in a Row. For example, if Component A has scale=2, and Component B has scale=1, A will be twice as wide as B. Should be an integer.
             min_width: minimum pixel width, will wrap if not sufficient screen space to satisfy this value. If a certain scale value results in this Component being narrower than min_width, the min_width parameter will be respected first.
+            proxy_url: The URL of the external Space used to load this component. Set automatically when using `gr.load()`. This should not be set manually.
```
abidlabs, line 61: We actually do need `proxy_url` in `gr.Dataset` because the behavior is different if we are loading examples from an external Space (in which case the examples are already processed)

aliabid94 (approved):
LGTM

test/test_components.py
```diff
@@ -1501,19 +1501,6 @@ def test_video_preprocessing_flips_video_for_webcam(self, mock_ffmpeg):
         assert ".avi" in list(output_params.keys())[0]
         assert ".avi" in output_file
 
-    @pytest.mark.flaky
```
abidlabs, line 1504: Pretty pointless test -- it was behaving really weird, passing locally but not on CI. Its actually unclear to me whether we would ever have this situation, where a URL is passed into `gr.Video().preprocess()` so deleting for now.

PR #6172, opened 2023-10-31
Queue concurrency count

aliabid94:
Allows setting concurrency count to limit the number of how many of a single event can be running at once in the queue. Also add concurrency_id to set limits for groups. For example:

```python
        with gr.Blocks() as demo:
            a = gr.Number()
            b = gr.Number()
            output = gr.Number()

            add_btn = gr.Button("Add")

            @add_btn.click(inputs=[a, b], outputs=output, concurrency_limit=2)
            def add(x, y):
                time.sleep(4)
                return x + y


            mul_btn = gr.Button("Multiply")

            @mul_btn.click(
                inputs=[a, b],
                outputs=output,
                concurrency_limit=2,
                concurrency_id="muldiv",
            )
            def mul(x, y):
                time.sleep(4)
                return x * y

            div_btn = gr.Button("Divide")

            @div_btn.click(
                inputs=[a, b],
                outputs=output,
                concurrency_limit=2,
                concurrency_id="muldiv",
            )
            def div(x, y):
                time.sleep(4)
                return x / y
```

Estimation calculation is now a lot harder - will tackle this in another PR.


client/python/test/conftest.py
```diff
@@ -258,7 +258,7 @@ def respond(message, st, chat_history):
 
         msg.submit(respond, [msg, st, chatbot], [msg, chatbot], api_name="submit")
         clear.click(lambda: None, None, chatbot, queue=False)
-        demo.queue()
+        demo
```
abidlabs, line 261:
```suggestion
```

gradio/blocks.py
```diff
@@ -1611,19 +1626,22 @@ def queue(
             demo.queue(max_size=20)
             demo.launch()
         """
-        if concurrency_count is None:
-            concurrency_count = 1 if utils.get_space() else 40
+        if "concurrency_count" in kwargs:
+            raise DeprecationWarning(
+                "concurrency_count has been deprecated. Set the concurrency_limit directly on event listeners e.g. btn.click(fn, ..., concurrency_limit=10) or gr.Interface(concurrency_limit=10). If necessary, the total number of workers can be configured via `max_threads` in launch()."
+            )
+        if len(kwargs):
```
abidlabs, line 1633: Won't this raise both a warning AND a ValueError if "concurrency_count" in kwargs?

gradio/events.py
```diff
@@ -341,6 +349,8 @@ def on(
     postprocess: bool = True,
     cancels: dict[str, Any] | list[dict[str, Any]] | None = None,
     every: float | None = None,
+    concurrency_limit: int | None = None,
```
abidlabs, line 352: We should set this to 1, right? Like we do in Interface

gradio/events.py
```diff
@@ -206,6 +206,8 @@ def event_trigger(
             cancels: dict[str, Any] | list[dict[str, Any]] | None = None,
             every: float | None = None,
             trigger_mode: Literal["once", "multiple", "always_last"] | None = None,
+            concurrency_limit: int | None = None,
```
abidlabs, line 209: We should set this to 1, right? Like we do in Interface

gradio/blocks.py
```diff
@@ -856,15 +867,21 @@ def set_event_trigger(
         _, progress_index, event_data_index = (
             special_args(fn) if fn else (None, None, None)
         )
+        if concurrency_limit is None and utils.get_space() is not None:
```
abidlabs, line 870: I thought we decided to unify

abidlabs:
Nice! Thanks for updating the guides as well. Gave some feedback on the default values, otherwise looks good

gradio/blocks.py
aliabid94: Nope, DeprecationWarning is actually an exception, not a warning

abidlabs: But if you put `raise` before it, it will interrupt the program

abidlabs: e.g. test with:

```py
a = "5"
raise DeprecationWarning
print(a)
```

I can fix it in my PR

abidlabs: Oh never mind I misunderstood

PR #6177, opened 2023-10-31
Part I: Remove serializes

abidlabs:
Part I of #6176

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -38,7 +38,7 @@ from gradio import (
 )
 from gradio.blocks_events import BlocksEvents, BlocksMeta
 from gradio.context import Context
-from gradio.data_classes import FileData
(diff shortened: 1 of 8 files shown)


gradio/blocks.py
```diff
@@ -392,6 +392,8 @@ def postprocess_update_dict(
             )
         if postprocess:
             attr_dict["value"] = block.postprocess(update_dict["value"])
+            if isinstance(attr_dict["value"], (GradioModel, GradioRootModel)):
```
freddyaboulton, line 395: Oh, so pydantic won't serialize it automatically?

abidlabs: Yea it didn't work without me manually serializing the value when it was in an update dictionary. If it was a regular prediction, it would be fine.

PR #6137, opened 2023-10-29
JS Param

dawoodkhan82:
## Description
Adds `js` and `head` params to Blocks init.
Changes `_js` -> `js` for events

Closes: #(issue)

## 🎯 PRs Should Target Issues

Before your create a PR, please check to see if there is [an existing issue](https://github.com/gradio-app/gradio/issues) for this change. If not, please create an issue before you create this PR, unless the fix is very small. 

Not adhering to this guideline will result in the PR being closed. 

## Tests

1. PRs will only be merged if tests pass on CI. To run the tests locally, please set up [your Gradio environment locally](https://github.com/gradio-app/gradio/blob/main/CONTRIBUTING.md) and run the tests: `bash scripts/run_all_tests.sh`

2. You may need to run the linters: `bash scripts/format_backend.sh` and `bash scripts/format_frontend.sh`

PR #6181, opened 2023-10-31
modify preprocess to use pydantic models

abidlabs:
Part II of #6176


gradio/blocks.py
```diff
@@ -1260,6 +1260,14 @@ def preprocess_data(
                     inputs_cached = processing_utils.move_files_to_cache(
                         inputs[i], block
                     )
+                    if getattr(block, "data_model", None) and inputs_cached is not None:
+                        if issubclass(block.data_model, GradioModel):  # type: ignore
+                            print("block.data_model", block.data_model, block)
```
freddyaboulton, line 1265: Can we get rid of the print statements?

abidlabs: Oops I'll do it on @dawoodkhan82's branch, thanks for catching

abidlabs: Done here: https://github.com/gradio-app/gradio/pull/6137

abidlabs: Let me know if you find anything else! I merged this just now so that we could have everything in place and do another release

PR #6183, opened 2023-10-31
Cleanup theme vars

aliabid94:
Wasn't much to cleanup, just chatbot_code_background_color was too component specific, and  not accurate either because all markdown used this variable. Renamed to code_background_fill, (change color -> fill for background, because we always use fill for background variable names, not color, because these can also be images/gradients/etc)


abidlabs (approved):
lgtm!

abidlabs:
Let's merge this in so that we can consolidate the PRs

PR #6211, opened 2023-11-01
fix`FileExplorer` preprocess

pngwn:
## Description

The datamodel for FileExplorer changed with a recent pr (i think #6181) but the preprocess was not modified to reflect this. The Signature for preprocess also wasn't updated which is probably what allowed this to creep in.

run `demo/file_explorer` to test.

Closes: #(issue)

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/tame-singers-join.md
+++ b/.changeset/tame-singers-join.md
@@ -1,0 +1,5 @@
+---
+"gradio": patch
+---
+
+fix:fix`FileExplorer` preprocess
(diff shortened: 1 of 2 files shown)


pngwn:
Cc @abidlabs

abidlabs:
I missed this, thanks @pngwn! Would be great to add some tests for `gr.FileExplorer`

pngwn:
Yeah @abidlabs, I agree! It was something I wanted to do at the time but I was running out time with v4 stuff. I'll revisit that shortly.

Issue #6230, opened 2023-11-01
Chatbot like/dislike not displaying even when like event is set


freddyaboulton:
The like/dislike event won't show up

```python
import gradio as gr
import random

def vote(tmp, index_state, data: gr.LikeData):
    value_new = data.value
    index_new = data.index
    if len(index_state) == 0 :
        index_state.append(index_new)
    else:
        if index_new in index_state:
            return "Your feedback is already saved", index_state
        else:
            index_state.append(index_new)
    return str(data.value) + ";" + str(data.index)+";"+ str(data.liked)+";"+str(index_state), index_state

with gr.Blocks() as demo:
    tmp = gr.Textbox(visible=True, value="")
    chatbot = gr.Chatbot(layout='panel')
    msg = gr.Textbox()
    clear = gr.ClearButton([msg, chatbot])
    index_state = gr.State(value=[])
    
    def respond(message, chat_history):
        bot_message = random.choice(["How are you?", "I love you", "I'm very hungry"])
        chat_history.append((message, bot_message))
        return "", chat_history

    msg.submit(respond, [msg, chatbot], [msg, chatbot])
    chatbot.like(vote, [tmp, index_state], [tmp, index_state])

demo.launch()
```

<img width="1277" alt="image" src="https://github.com/gradio-app/gradio/assets/121886414/f6e80293-87ae-4ace-9ec4-1ba10f417efb">

Would be great to have this built in the ChatInterface

_Originally posted by @surajspass in https://github.com/gradio-app/gradio/issues/5871#issuecomment-1789159931_

sajjadmosaheb:
how can I add a like/dislike function to Chatinterface?
this is possible with blocks but I can't add to a Chatinterface.

import gradio as gr

def greet(history, input):
    return history + [(input, "Hello, " + input)]

def vote(data: gr.LikeData):
    if data.liked:
        print("You upvoted this response: " + data.value)
    else:
        print("You downvoted this response: " + data.value)
    

with gr.Blocks() as demo:
    chatbot = gr.Chatbot()
    textbox = gr.Textbox()
    textbox.submit(greet, [chatbot, textbox], [chatbot])
    chatbot.like(vote, None, None)  # Adding this line causes the like/dislike icons to appear in your chatbot
    
demo.launch()

PR #6231, opened 2023-11-01
Add likeable to config for Chatbot

freddyaboulton:
## Description

Repro in issue.

We added some under-the-hood logic to Blocks to set `_selectable` in the config without it being present in the init signature. I don't like treating some props as privileged though. Doesn't scale for custom component authors who want to set event-based props in their custom component. So I'm adding likeable to the config.

Closes: #6230 

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/test/test_components.py
+++ b/test/test_components.py
@@ -1895,6 +1895,16 @@ class TestJSON:
             "_selectable": False,
         }
 
+    def test_chatbot_selectable_in_config(self):
+        with gr.Blocks() as demo:
+            cb = gr.Chatbot(label="Chatbot")
+            cb.like(lambda: print("foo"))
+        for component in demo.config["components"]:
+            if component["props"]["label"] == "Chatbot":
+                assert component["props"]["likeable"]
+                return
+        raise AssertionError()
+
     @pytest.mark.asyncio
     async def test_in_interface(self):
         """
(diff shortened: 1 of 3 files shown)


test/test_components.py
```diff
@@ -1895,6 +1895,16 @@ def test_component_functions(self):
             "_selectable": False,
         }
 
+    def test_chatbot_selectable_in_config(self):
```
abidlabs, line 1898:
```suggestion
    def test_chatbot_likeable_in_config(self):
```

abidlabs: Maybe we can also test to ensure that likable is _not_ present if no `.like` event has been set

abidlabs:
> I don't like treating some props as privileged though. Doesn't scale for custom component authors who want to set event-based props in their custom component. So I'm adding likeable to the config.

Makes sense. Let's also do the same for `_selectable` then for consistency across components. We'll add it back as a param to all of the components and do it in a backwards compatible way to support `gr.load()` components with the `_selectable` key. I can take care of this in a separate PR.

abidlabs:
Oh tested, works well. LGTM!

test/test_components.py
freddyaboulton: Yea good idea!

freddyaboulton:
Thanks @abidlabs !

PR #6232, opened 2023-11-01
Remove **kwargs from queue, raise warning instead of error concurrency_count

freddyaboulton:
## Description

Don't think removing kwargs is a breaking change because we were already not allowing arbitrary kwargs. Removing them from the signature has the benefit of having your editor warn you before you run your code and reducing code for us. Also treat concurrency_count as a warning instead of an error.

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -1624,7 +1624,7 @@ Received outputs:
         status_update_rate: float | Literal["auto"] = "auto",
         api_open: bool | None = None,
         max_size: int | None = None,
-        **kwargs,
+        concurrency_count: int | None = None,
     ):
         """
         By enabling the queue you can control when users know their position in the queue, and set a limit on maximum number of events allowed.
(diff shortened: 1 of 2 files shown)


aliabid94:
Why not raise an exception if we no longer support concurrency count? We deprecated it moving into 4.0. I think it's a good way to force old apps to think where they want to set their concurrency limits instead.

freddyaboulton:
You want to keep in signature but raise an error instead of warning? Happy to but `main` branch is using a deprecation warning so I thought the intention is to warn.

aliabid94:
> You want to keep in signature but raise an error instead of warning? 

I removed it from the signature because it was being deprecated, but I added **kwargs so that if an 3.x user tried to use it, they got a specific error telling them it was deprecated. (DeprecationWarning is an exception type, not a warning, despite its name).

I feel like that's a proper deprecation compared to keeping it in the signature and it being ignored.

freddyaboulton:
All warnings are exception classes but the convention is to call via `warnings.warn` in my opinion. I don't feel strongly about it though I can add that part back.

I don't want us to add back **kwargs though. I think its nicer if you get visual feedback that a parameter is not allowed as opposed to having to run your code and then get an error.

image

pngwn:
It isn't deprecated if it is gone. A deprecation indicates that the thing still works but that it isn't recommended/ optimal/ may be removed in the future. 

I think this would be a `ValueError` or something.

aliabid94:
I don't have a strong opinion either, but

> I think its nicer if you get visual feedback that a parameter is not allowed before you run the code as opposed to having to run your code and then get an error

This is the logic I used for getting *rid* of concurrency_count from the arguments. I didn't want users to see this as a valid parameter, but I still wanted to pass a specific message for users who tried using it.

abidlabs:
Good with this change. I agree that we should avoid `**kwargs` as much as possible. They make it very easy to have typos, and we lose the benefit of static typechecking of arguments. I think a DeprecationWarning is reasonable here since we are likely to remove this argument in the near future.

abidlabs (approved):

freddyaboulton:
Thanks for the discussion all! Keeping the warning raise but removing kwargs.

PR #6236, opened 2023-11-01
Ensure `gr.CheckboxGroup` updates as expected.

### LINKED ISSUE #6208: [Gradio 4] updating value of `gr.CheckboxGroup` doesn't work properly
### Describe the bug

It seems updating `gr.CheckboxGroup` is now broken with `gradio==4.0.2`.

### Reproduction

```py
import gradio as gr

CHOICES = ["(NONE)", "a", "b", "c"]


def fn(choices):
    print(choices)
    if "(NONE)" in choices:
        return []
    else:
        return choices


with gr.Blocks() as demo:
    checkbox = gr.CheckboxGroup(choices=CHOICES, value=["a", "b", "c"])
    checkbox.change(fn=fn, inputs=checkbox, outputs=checkbox)

if __name__ == "__main__":
    demo.queue().launch()
```

Gradio 4.0.2
repro Space: https://huggingface.co/spaces/hysts-debug/gradio-4-checkboxgroup-update

https://github.com/gradio-app/gradio/assets/25161192/f7a0cd6c-a73c-47c5-a7ae-5d66291df4f2

Gradio 3.50.2
repro Space: https://huggingface.co/spaces/hysts-debug/gradio-3-50-checkboxgroup-update

https://github.com/gradio-app/gradio/assets/25161192/1e8c88d9-d554-453b-bc0b-f986cf734f11

This code works as intended with `gradio==3.50.2`, but doesn't work with `gradio==4.0.2`.

### Screenshot

_No response_

### Logs

_No response_

### System Info

```shell
gradio==4.0.2
```

### Severity

I can work around it

pngwn:
## Description

Fixes #6208.

I was a bit confused about the dispatch behaviour here. We weren't really using the old value in any meaningful way, so I refactored things a little and i think it is simpler now.

I added `immutable={true}` which means that mutations won't update cause a rerender, but i needed to tweak thigns a little to ensure we always created a new array when values changed. Externals changes _always_ give us a new structure so this is cool.

Closes: #(issue)

## 🎯 PRs Should Target Issues

## Tests


abidlabs:
LGTM @pngwn, tested and confirmed the fix. Also code is easier to parse 👍👍

pngwn:
good shout

pngwn:
Actually i think this is wrong, it won't dispatch `change` when we pass a value from outside.

We definitely need those tests!

js/tootils/src/render.ts
```diff
@@ -69,6 +83,7 @@ export async function render<
 		target,
 		//@ts-ignore
 		props: {
+			loading_status,
```
pngwn, line 86: I've updated the test utils so we don't need to pass `loading_status` to every component in tests. We _can_ pass one in if we want to test it for any reason but a default is passed now.

js/checkboxgroup/Index.svelte
```diff
@@ -73,33 +58,33 @@
 	<BlockTitle {show_label} {info}>{label}</BlockTitle>
 
 	<div class="wrap" data-testid="checkbox-group">
-		{#each choices as choice, i}
-			<label class:disabled class:selected={value.includes(choice[1])}>
+		{#each choices as [display_value, internal_value], i}
```
pngwn, line 61: Deconstructed this so is easier to read.

js/checkboxgroup/Index.svelte
```diff
@@ -14,46 +15,30 @@
...
 		}
-		old_value = [...old_value];
-	}
 
-	function handle_change(): void {
-		gradio.dispatch("change");
-		if (!value_is_output) {
-			gradio.dispatch("input");
-		}
+		gradio.dispatch("input");
```
pngwn, line 36: Its pretty easy to just dispatch `input` explicitly here, we don't need the confusing `value_is_output` logic.

js/checkboxgroup/Index.svelte
```diff
@@ -14,46 +15,30 @@
...
-	export let value_is_output = false;
-	export let choices: [string, number][];
+	export let choices: [string, string | number][];
 	export let container = true;
 	export let scale: number | null = null;
 	export let min_width: number | undefined = undefined;
 	export let label = gradio.i18n("checkbox.checkbox_group");
 	export let info: string | undefined = undefined;
-	export let show_label: boolean;
+	export let show_label = true;
```
pngwn, line 24: Added default for easier testing.

js/checkboxgroup/Index.svelte
```diff
@@ -14,46 +15,30 @@
...
 
 	export let loading_status: LoadingStatus;
-	export let interactive: boolean;
-
-	let old_value: (string | number)[] = value.slice();
+	export let interactive = true;
```
pngwn, line 27: Added default for easier testing.

js/app/vite.config.ts
```diff
@@ -181,7 +194,10 @@ export default defineConfig(({ mode }) => {
 					? ["**/*.node-test.{js,mjs,cjs,ts,mts,cts,jsx,tsx}"]
 					: ["**/*.test.{js,mjs,cjs,ts,mts,cts,jsx,tsx}"],
 			exclude: ["**/node_modules/**", "**/gradio/gradio/**"],
-			globals: true
+			globals: true,
+			onConsoleLog(log, type) {
+				if (log.includes("was created with unknown prop")) return false;
+			}
```
pngwn, line 200: Got rid of those annoying warnings in tests that render stdio useless.

js/checkboxgroup/checkboxgroup.test.ts
pngwn, line 1: So these are my units tests. The real value of unit tests is two fold:

- they are cheap so you test lots of variants.
- They test only one thing at a time so when one fails you know exactly what has failed.

js/checkboxgroup/checkboxgroup.test.ts
```diff
@@ -0,0 +1,345 @@
...
+	show_progress: "full"
+};
+
+beforeEach(() => {
+	setupi18n();
+});
+
+afterEach(cleanup);
+
+describe("Values", () => {
```
pngwn, line 26: I use describe blocks to make it clear what a series of tests are testing.

There is no value in a suite call `checkboxgroup.test.ts` having a describe block of `CheckboxGroup`.

In this suite we are focused on the value being updated correctly both in the UI and the actual `value` of the component. This is very easy to test multiple variants. Here i tested checking things of worked + updated the value/ checked status of the input in various forms.

I also checked that setting + updating the value from outside works as expected. They are different code paths so it is important to test them separately.

js/checkboxgroup/checkboxgroup.test.ts
```diff
@@ -0,0 +1,345 @@
...
+		const item_one = getByLabelText("Choice One") as HTMLInputElement;
+		const item_two = getByLabelText("Choice Two") as HTMLInputElement;
+		const item_three = getByLabelText("Choice Three") as HTMLInputElement;
+		assert.equal(item_one.checked, true);
+		assert.equal(item_two.checked, false);
+		assert.equal(item_three.checked, true);
+
+		expect(item_one).toBeChecked();
+		expect(item_two).not.toBeChecked();
+		expect(item_three).toBeChecked();
```
pngwn, line 64: Common wisdom says 'one unit test, one assertion' but this is unnecessarily restrictive. Multiple assertions are fine if they are all test a discreet thing. If multiple things are being tested however, they should be broken into new tests.

Overlap in unit tests is fine, they are cheap.

js/checkboxgroup/checkboxgroup.test.ts
```diff
@@ -0,0 +1,345 @@
...
+
+		component.value = ["choice_one"];
+
+		expect(item_one).not.toBeChecked();
+		expect(item_two).not.toBeChecked();
+		expect(item_three).not.toBeChecked();
+	});
+});
+
+describe("Events", () => {
```
pngwn, line 179: Events are the other important thing in this case. We want to test a few different cases again. We have three events to think about here and their behaviour is slightly different. We want to test that the fire when they should but we also want to ensure they do not fire when they shouldn't too. We also need to check payloads in certain cases (`select`).

We have a `listen` utility to help with this (and it is fully typechecked so you should get autocomplete on these event names).

js/checkboxgroup/checkboxgroup.test.ts
```diff
@@ -0,0 +1,345 @@
...
+		const item_one = getByLabelText("Choice One") as HTMLInputElement;
+
+		const mock = listen("input");
+		await (component.value = [1]);
+
+		expect(mock.callCount).toBe(0);
+	});
+});
+
+describe("interactive vs static", () => {
```
pngwn, line 273: And then we have interactive and static. We need to test these in each component because exactly what interactive and static means is different for each component and checking the functionality is also specific.

We need more of these test, we have had quite a few issues with this behaviour just as a result of human error when refactoring.

js/app/vite.config.ts
```diff
@@ -147,7 +155,12 @@ export default defineConfig(({ mode }) => {
 				inspector: true,
 				compilerOptions: {
 					dev: true,
-					discloseVersion: false
+					discloseVersion: false,
+					accessors: mode === "test"
```
pngwn, line 159: I've turned on `accessors` for all components during tests, so we can just do `component.value` to get the value and `component.value = ...` to set the value. Easier than messing around with component methods. It adds extra code so we don't want to enable it all of the time.

pngwn (commented):
Left some comments on the PR and specifically the tests to show my approach here. Also made some tweaks to make testing a touch easier.

All in all I'm pretty confident that this suite would have caught both the issue this PR is fixing as well as a host of other issues that have come up at different times.

The big takeaway here is just the number of tests I was able to write with no real degradation in CI. I think there are about 20. The other thing is that the tests are very specific. If a tests fails it should be incredibly obvious exactly what went wrong, that is the main value of unit tests over integration of functional tests (and something we should think about when writing python unit tests).

Hope this is helpful!

js/tootils/src/render.ts
abidlabs: Nice!

js/checkboxgroup/Index.svelte
abidlabs: Woudln't this also dispatch `input` if the value changes programmatically instead of via user input?

abidlabs: Well I tested it and it doesn't seem to, though I'm not totally sure why

abidlabs:
Awesome PR @pngwn! Tested and both `.input()` and `.change()` work as expected.

Test code:

```py
import gradio as gr

CHOICES = ["(NONE)", "a", "b", "c"]


def fn(choices):
    print(choices)
    if "(NONE)" in choices:
        return []
    else:
        return choices

def test_change():
    print("Changed")

def test_input():
    print("Inputted")
    
with gr.Blocks() as demo:
    checkbox1 = gr.CheckboxGroup(choices=CHOICES, value=["a", "b", "c"])
    checkbox2 = gr.CheckboxGroup(choices=CHOICES, value=["a", "b", "c"], interactive=True)
    checkbox1.change(fn=fn, inputs=checkbox1, outputs=checkbox2)
    checkbox2.change(fn=test_change, inputs=None, outputs=None)
    checkbox2.input(fn=test_input, inputs=None, outputs=None)

if __name__ == "__main__":
    demo.queue().launch()
```

js/checkboxgroup/Index.svelte
pngwn: `toggle_choice` is only called when the `input` elements `change` event is fired. This doesn't fire when the value is set programmatically, only when a user changes it in the UI.

PR #6240, opened 2023-11-01
Model3D panning, improved UX

dylanebert:
Improvements to the Model3D component

- Enabled panning
- Adapt pan and zoom speed to scene radius, creating a consistent experience across various model sizes
- Set lower radius limit to prevent zoom inversion (zooming in below 0)

Closes: #6119


abidlabs:
LGTM tested on mobile as well. This:

> Set lower radius limit to prevent zoom inversion (zooming in below 0)

is really nice, thanks @dylanebert!

Issue #6239, opened 2023-11-01
Browser closing does not drop user from Queue


aliabid94:
Apparently SSE does not detect browser closing as a disconnection, so it keeps these sessions in queue.

PR #6241, opened 2023-11-01
Remove session if browser closed on mobile

aliabid94:
Fixes: #6239 

Previously if a user closed the browser, their session was not removed from queue. This is now handled.

Test this via:
```python
import gradio as gr

def greet(name):
    import time
    time.sleep(20)
    return "Hello " + name + "!"

if __name__ == "__main__":
    demo.launch()   
```
Launch a couple sessions across a few tabs and then close some tabs. The positions in the other tabs should update.

### DIFF
--- a/gradio/routes.py
+++ b/gradio/routes.py
@@ -615,28 +615,32 @@ class App(FastAPI):
                     await blocks._queue.send_estimation(event, estimation, rank)
 
             async def sse_stream(request: fastapi.Request):
-                last_heartbeat = time.perf_counter()
-                while True:
-                    if await request.is_disconnected():
-                        await blocks._queue.clean_event(event)
-                    if not event.alive:
-                        return
-
-                    heartbeat_rate = 15
-                    check_rate = 0.05
-                    message = None
-                    try:
-                        message = event.message_queue.get_nowait()
-                        if message is None:  # end of stream marker
+                try:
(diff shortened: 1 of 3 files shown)


gradio/routes.py
```diff
@@ -615,28 +615,32 @@ async def queue_join(
...
+                                return
+                        except EmptyQueue:
+                            await asyncio.sleep(check_rate)
+                            if time.perf_counter() - last_heartbeat > heartbeat_rate:
+                                message = {"msg": "heartbeat"}
+                                last_heartbeat = time.time()
+
+                        if message:
+                            yield f"data: {json.dumps(message)}\n\n"
+                except asyncio.CancelledError as e:
```
abidlabs, line 641: What line is it that actually throws the asyncio.CancelledError? Could it be a lot of different lines in the `try` block? If not, the code would be easier to maintain if we could narrow the `try` block

abidlabs:
Nice @aliabid94 can confirm the fix!

abidlabs (approved):

gradio/routes.py
aliabid94: It could happen anywhere in the block, this is the standard convention I've seen. It's like a KeyboardInterrupt error

abidlabs: Ok sg

PR #6243, opened 2023-11-02
Some tweaks to the Custom Components Guide

abidlabs:
As I was building a custom component, noticed some small areas of improvement for the custom components guides:

* Removed the old "Create a New Component" Guide since that was the first thing that showed up when I googled Gradio Custom Component, and made the old URL redirect to the new Guide
* Various nits


guides/05_custom-components/02_key-component-concepts.md
```diff
@@ -39,16 +39,15 @@ with gr.Blocks() as demo:
 demo.launch()
 ```
 
-The interactive version of the component is much more complex -- you can upload and edit images, draw sketches, etc. -- while the static version does not do any of that.
+The interactive version of the component is much more complex -- you can upload images or snap a picture from your webcam -- while the static version can only be used to display images.
 
-Not every component has an interactive version.
-For example, the `gr.AnnotatedImage` only appears as a static version since there's no way to interactively change the value of the annotations or the image.
+Not every component has a distinct interactive version. For example, the `gr.AnnotatedImage` only appears as a static version since there's no way to interactively change the value of the annotations or the image.
```
freddyaboulton, line 44: This and the current source are equivalent.

js/_website/src/routes/redirects.js
```diff
@@ -1,4 +1,5 @@
 export const redirects = {
+	"/guides/creating-a-new-component": "/guides/five-minute-guide",
```
freddyaboulton, line 2: 👍

freddyaboulton (approved):
Thanks @abidlabs !

abidlabs:
Cool, working on the BlurHashGallery custom component just fyi

freddyaboulton:
The what? 👀

PR #6244, opened 2023-11-02
Fix stories

abidlabs:
Fixes some Storybook stories that had broken when we refactored `FileData`


pngwn (approved):
Awesome. Thanks @abidlabs!

pngwn:
Gonna merge this in so i can get my Ci to pass properly.

Issue #6196, opened 2023-10-31
When flagging with CSVLogger(), directories are made also for components that should not trigger it (Ex. textual components)


EgonFerri:
### Describe the bug

The 4.x update caused an unexpected behavior (I suspect due to some changes in Components internals).

When flagging with CSVLogger(), repositories are generated for each component (also for those that don't need it).
The repositories should be generated only for the components that require additional files (ex. images, video, audio). For those components, the repositories are correctly generated.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```
import gradio as gr


def greet(name):
    return "Hello " + name + "!"


demo = gr.Interface(fn=greet, inputs="text", outputs="text", flagging_dir='flagged')

if __name__ == "__main__":
    demo.launch()

```

When flagging, directories `name` and `output` are generated. 
In the CSV, by the way, the values are correctly saved.

### Screenshot

![image](https://github.com/gradio-app/gradio/assets/43636749/c37acad2-e9bd-4fac-b36c-323630fa700b)

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Darwin
gradio version: 4.0.2
gradio_client version: 0.7.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.1.1
fastapi: 0.103.1
ffmpy: 0.3.1
gradio-client==0.7.0 is not installed.
httpx: 0.25.0
huggingface-hub: 0.17.2
importlib-resources: 6.0.1
jinja2: 3.1.2
markupsafe: 2.1.3
matplotlib: 3.8.0
numpy: 1.26.1
orjson: 3.9.7
packaging: 23.1
pandas: 2.1.0
pillow: 10.1.0
pydantic: 2.3.0
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0.1
requests: 2.31.0
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.9.0
typing-extensions: 4.8.0
uvicorn: 0.23.2
websockets: 11.0.3
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2023.9.1
httpx: 0.25.0
huggingface-hub: 0.17.2
packaging: 23.1
requests: 2.31.0
typing-extensions: 4.8.0
websockets: 11.0.3
```

### Severity

[...]

PR #6245, opened 2023-11-02
fix: avoid unnecessary folders creation when flagging

EgonFerri:
## Description

As discussed with @abidlabs in #6196, fixes the issue by moving the creation of a folder for flagged components inside the `.flag()` method in `components/base.py`, making it run only when data_model is not `none` before the payload is copied inside of it.

## 🎯 PRs Should Target Issues

Closes: #6196

## Tests

1. I followed the guidelines to contribute (https://github.com/gradio-app/gradio/blob/main/CONTRIBUTING.md) and ran the tests: `bash scripts/run_all_tests.sh`. All tests passed.

### DIFF
--- a/gradio/components/base.py
+++ b/gradio/components/base.py
@@ -273,6 +273,7 @@ class Component(ComponentBase, Block):
         """
         if self.data_model:
             payload = self.data_model.from_json(payload)
+            flag_dir.mkdir(exist_ok=True)
             return payload.copy_to_dir(flag_dir).model_dump_json()
         return payload
 
--- a/gradio/components/base.py
+++ b/gradio/components/base.py
@@ -273,7 +273,7 @@ class Component(ComponentBase, Block):
         """
         if self.data_model:
             payload = self.data_model.from_json(payload)
-            flag_dir.mkdir(exist_ok=True)
+            Path(flag_dir).mkdir(exist_ok=True)
             return payload.copy_to_dir(flag_dir).model_dump_json()
         return payload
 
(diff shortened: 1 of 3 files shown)


abidlabs:
LGTM thanks @EgonFerri for the contribution and nice fix.

abidlabs:
I'm just going to add a test here to prevent regressions and then will go ahead and merge this in

abidlabs:
Could you [allow maintainers to push to this branch please](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork)?

abidlabs:
Or feel free to make the following changes -- I added this test:

```py
    def test_flagging_does_not_create_unnecessary_directories(self):
        with tempfile.TemporaryDirectory() as tmpdirname:
            io = gr.Interface(lambda x: x, "text", "text", flagging_dir=tmpdirname)
            io.launch(prevent_thread_lock=True)
            io.flagging_callback.flag(["test", "test"])
            assert os.listdir(tmpdirname) == ["log.csv"]
```

to `test/test_flagging.py`

and I updated `flag_dir.mkdir(exist_ok=True)` to be `Path(flag_dir).mkdir(exist_ok=True)` in case `flag_dir` is a `str`

EgonFerri:
Since the fork is hosted by my company, I think I am not able to allow you to push easily, btw I incorporated your suggestions.

A test to prevent regressions seems the right choice. Thanks for the help (: we should be ready to merge

abidlabs:
Perfect, thank you again @EgonFerri!

PR #6249, opened 2023-11-02
ensure radios have different names

### LINKED ISSUE #6198: Radio bug
### Describe the bug

Reported on discord. 

Is there a new bug in radio choices after the update? I had my demo code in Colab, which included some radio “yes/no” choices. However, today when I started the runtime and ran the app again in Colab, I couldn’t choose “all yes” or “all no”! When I tried to do so, the previous answer disappeared. Now, I wrote a sample code to test if it remains or not, and I see the same results. Here is the sample code:

```py
import gradio as gr

with gr.Blocks() as demo:
  with gr.Row():
      with gr.Column():
        Drug = gr.Radio(["Yes","No"])
        Family = gr.Radio(["Yes","No"])
  if __name__ == "__main__":
      demo.queue()
      demo.launch()
```
When I click “yes” for the first box and then do the same for the second radio, the first answer gets removed! Would you please help me with this problem?

### Reproduction

```py
import gradio as gr

with gr.Blocks() as demo:
  with gr.Row():
      with gr.Column():
        Drug = gr.Radio(["Yes","No"])
        Family = gr.Radio(["Yes","No"])
  if __name__ == "__main__":
      demo.queue()
      demo.launch()
```

### Screenshot

_No response_

### Logs

_No response_

### System Info

```shell
4.0
```

### Severity

Blocking usage of gradio

pngwn:
## Description

Closes #6198 

Funny issue. The `name` attribute on a Radio button has to be unique across the page or there will be conflicts. We were using the internal value (in the PR example 'Yes' and 'No'). SO when we had duplicates with the same value only one 'Yes' or one 'No' could be 'checked' at the same time (because radio inputs are xor).

Added a few defaults again to make testing easier and added a unit test for this case.

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/js/radio/Radio.test.ts
+++ b/js/radio/Radio.test.ts
@@ -23,16 +23,13 @@ describe("Radio", () => {
 		["dog", "dog"],
 		["cat", "cat"],
 		["turtle", "turtle"]
-	];
+	] as [string, string][];
 
 	test("renders provided value", async () => {
 		const { getAllByRole, getByTestId } = await render(Radio, {
-			show_label: true,
-			loading_status,
 			choices: choices,
 			value: "cat",
-			label: "Radio",
-			interactive: true
+			label: "Radio"
 		});
 
 		assert.equal(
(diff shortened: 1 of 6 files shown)


abidlabs:
Nice @pngwn. Confirms this fixes the original issue, thanks for adding the tests!

abidlabs (approved):

js/radio/Radio.test.ts
```diff
@@ -79,4 +73,29 @@ describe("Radio", () => {
 			true
 		);
 	});
+
+	test("when multiple radios are on the screen, they should not conflict", async () => {
```
freddyaboulton, line 77: Will learn from this 🙏

freddyaboulton (approved):

Issue #6205, opened 2023-11-01
[Gradio 4] Use different colors for waveform and trimming box in `gr.Audio`


hysts:
- [x] I have searched to see if a similar issue already exists.

The audio trimming feature is awesome, but maybe it would be nice if the colors of the waveform and the box are different.

![](https://github.com/gradio-app/gradio/assets/25161192/a3d8f136-874f-453e-9fca-1444cf30d4c8)

Issue #6201, opened 2023-11-01
[Gradio 4] Add audio volume control to `gr.Audio` and `gr.Video`


hysts:
- [x] I have searched to see if a similar issue already exists.

There was a volume control button for `gr.Audio` in `gradio==3.50.2`, but now it's gone in `gradio==4.0.2`. Also, it would be nice if `gr.Video` had a volume control button as well.

Issue #6195, opened 2023-10-31
v4: Audio stop_recording event doesn't send filepath as output anymore


jackielii:
### Describe the bug

Just upgraded to version 4.0.2. The Audio component's stop_recording event doesn't send filepath as output anymore

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

sample code:

```python
import gradio as gr

with gr.Blocks() as demo:
    audio = gr.Audio(sources=["microphone", "upload"], type="filepath", label="Audio Input")
    output = gr.Textbox()
    audio.stop_recording(lambda x: x, audio, output)

demo.queue().launch(debug=True)
```

output always gets None

### System Info

```shell
❯ gradio environment
Gradio Environment Information:
------------------------------
Operating System: Darwin
gradio version: 4.0.2
gradio_client version: 0.7.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.1.2
fastapi: 0.104.1
ffmpy: 0.3.1
gradio-client==0.7.0 is not installed.
httpx: 0.25.0
huggingface-hub: 0.18.0
importlib-resources: 6.1.0
jinja2: 3.1.2
markupsafe: 2.1.3
matplotlib: 3.8.0
numpy: 1.26.1
orjson: 3.9.10
packaging: 23.2
pandas: 2.1.2
pillow: 10.1.0
pydantic: 2.4.2
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0.1
requests: 2.31.0
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.9.0
typing-extensions: 4.8.0
uvicorn: 0.23.2
websockets: 11.0.3
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2023.10.0
httpx: 0.25.0
huggingface-hub: 0.18.0
packaging: 23.2
requests: 2.31.0
typing-extensions: 4.8.0
websockets: 11.0.3
```

### Severity

Blocking usage of gradio

aliabid94:
Taking a look!

PR #6254, opened 2023-11-02
Add volume control to Audio

### LINKED ISSUE #6207: [Gradio 4] Adjust waveforms depending on the sound volume range in `gr.Audio`

For loud audio input, the waveform appears like this, and is almost meaningless. So, it would be nice if the display scale of the waveform could be changed according to the sound volume of the input audio.

hannahblair:
## Description

This PR

- Adds ability to adjust volume of waveform
- Adds audio normalisation
- Allows setting trim region colour
- Ensures we send the file detail to `stop_recording`
- Adds a few tests
- Improves typing in some files

Closes: #6205 
Closes #6207 
Closes: #6201
Closes: #6195

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/js/audio/recorder/AudioRecorder.svelte
+++ b/js/audio/recorder/AudioRecorder.svelte
@@ -183,12 +183,12 @@
 </script>
 
 <div class="component-wrapper">
-	<div bind:this={microphoneContainer} data-testid="microphone-waveform" />
 	<div
 		class="microphone"
-		bind:this={recordingContainer}
-		data-testid="recording-waveform"
+		bind:this={microphoneContainer}
+		data-testid="microphone-waveform"
 	/>
+	<div bind:this={recordingContainer} data-testid="recording-waveform" />
 
 	{#if timing || recordedAudio}
 		<div class="timestamps">


aliabid94:
I tried running the reverse_audio demo and it threw an error, seems like it was sending null as the value after recording an audio clip.

abidlabs:
I'm seeing the same error @aliabid94 described:

Recording 2023-11-13 at 14 03 05

The microphone waveform seems to also be moving very fast.

I tested `audio_debugger` -- seing the same issue with the input microphone file being null. Note that if I upload a file or click on an example file, there's no issue.

js/audio/shared/VolumeControl.svelte
```diff
@@ -0,0 +1,74 @@
...
+	export let currentVolume = 1;
+	export let show_volume_slider = false;
+	export let waveform: WaveSurfer;
+
+	onMount(() => {
+		adjustSlider();
+	});
+
+	const adjustSlider = (): void => {
+		let slider = document.getElementById("volume");
```
aliabid94, line 14: won't this approach only allow one audio element at a time? we should be binding elements and using references to them: https://svelte.dev/repl/76a65cdbd14f40a5818ab327c7b3105b?version=3.29.4

aliabid94:
played around with it and everything worked well! there was a glitch when I opened two sliders at the same time (the coloring messed up) I believe this is for the reason I commented above. Other than that, works great! Code looked good to me from what I could make of it. Great work @hannahblair

aliabid94 (approved):

PR #6261, opened 2023-11-02
Improve Embed and CDN handling and fix a couple of related bugs 

pngwn:
## Description

Closes: #(issue)

## 🎯 PRs Should Target Issues

## Tests


abidlabs:
@aliabd you wanna review this?

build_pypi.sh
```diff
@@ -9,12 +9,9 @@ new_version=$(python -c "import json; f = open('$FILE', 'r'); data = json.load(f
 GRADIO_VERSION=$new_version
 
 rm -rf gradio/templates/frontend
-rm -rf gradio/templates/cdn
 pnpm i --frozen-lockfile --ignore-scripts
 GRADIO_VERSION=$new_version pnpm build
-GRADIO_VERSION=$new_version pnpm build:cdn
-aws s3 cp gradio/templates/cdn "s3://gradio/${new_version}/" --recursive --region us-west-2
-cp gradio/templates/cdn/index.html gradio/templates/frontend/share.html
+aws s3 cp gradio/templates/frontend "s3://gradio/${new_version}/" --recursive --region us-west-2
```
pngwn, line 14: Since we use the same build for everything now, we only need to build once and we can upload the files to the server from the `templates/frontend` dir. It contains everything we need.

.config/basevite.config.ts
```diff
@@ -27,10 +27,7 @@ const version = JSON.parse(readFileSync(version_path, { encoding: "utf-8" }))
 
 //@ts-ignore
 export default defineConfig(({ mode }) => {
-	const production =
-		mode === "production:cdn" ||
-		mode === "production:local" ||
-		mode === "production:website";
+	const production = mode === "production";
```
pngwn, line 30: Dont need all of these modes anymore.

js/app/build_plugins.ts
```diff
@@ -17,70 +17,47 @@ export function inject_ejs(): Plugin {
...
+			const transformed_html =
+				(bundle["index.html"].source as string).substring(0, script?.range[0]) +
+				`<script type="module" crossorigin src="${cdn_base}/${version}/gradio.js"></script>` +
+				(bundle["index.html"].source as string).substring(
+					script?.range[1],
+					source.length
+				);
+
+			const share_html_location = join(config.dir, "share.html");
+			writeFileSync(share_html_location, transformed_html);
```
pngwn, line 60: This is for the `share.html` rather than copying it over like we used to, we created it directly from the `index.html`. What we are doing here is prepending the `cdn_base` to the scripts src. This is the only time we need to do this.

js/app/build_plugins.ts
```diff
@@ -109,18 +87,7 @@ export function generate_dev_entry({ enable }: { enable: boolean }): Plugin {
...
-    script.type = 'module';
-    script.setAttribute("crossorigin", "");
-    script.src = src;
-    document.head.appendChild(script);
-}`;
-
-	return `
-${make_script}
-make_script("${script}");
+	return `import("${script}");
```
pngwn, line 90: I dunno why i chose the hard route previsouly.

The problem we always had was that anything that is requested from an html tag will always treat its own location as the base url from which to resolve files. With `import` it will use the files own location.

This `gradio.js` file will be request like: `https://aws-etc.com/4.0.0/gradio.js` so when we request`./assets/index-asdasdasd.js` it will be considered relative to that `gradio.js` url. 

When we add a script to the page the 'file' in that case is the HTML file, which isn't where the cdn assets live.

js/app/build_plugins.ts
```diff
@@ -192,7 +151,19 @@ export function handle_ce_css(): Plugin {
...
+			const node = Array.from(
+				share_tree.querySelectorAll("link[rel=stylesheet]")
+			).find((node) => /.*\/index(.*?)\.css/.test(node.attributes.href));
+
+			if (!node) return;
+			const transformed_html =
+				share_html.substring(0, node.range[0]) +
+				share_html.substring(node.range[1], share_html.length);
+
+			writeFileSync(share_html_location, transformed_html);
```
pngwn, line 166: We are just stripping out the css file for the `share.html`, to prevent network errors. The CSS is loaded by the JS later anyway, so its no big deal.

js/app/src/css.ts
pngwn, line 1: This file was actually the real problem. The CSS assets can only be loaded via a style tag, so we have the issues i mentioned earlier where the base URL was always the location of the HTML file. I fixed this by constructing a new URL based on the location of the file. This seems to work both for remote links (cdn) and for local (normal gradio).

js/app/src/main.ts
```diff
@@ -21,6 +21,15 @@ let FONTS: string | [];
...
 let IndexComponent: typeof Index;
+let _res: (value?: unknown) => void;
+let pending = new Promise((res) => {
+	_res = res;
+});
+async function get_index(): Promise<void> {
+	IndexComponent = (await import("./Index.svelte")).default;
+	_res();
+}
+
```
pngwn, line 32: This is unrelated to the build changed but fixes a race condition I noticed.

pngwn:
This _can_ be tested locally.

You need to build gradio:

```bash
pnpm build
```

Start a local server to serve the gradio files:

```bash
pnpm preview:cdn-server
```

Then change `js/_website/src/routes/+layout.svelte` `Line 77` to `"http://localhost:4321/gradio.js"`

build_pypi.sh
freddyaboulton: Oh nice

aliabd (approved):
thanks @pngwn! tested locally and everything works great!

Issue #6197, opened 2023-10-31
gradio4: updating interactivity of buttons not working?


jerpint:
### Describe the bug

I've been having issues migrating my gradio3 app to gradio4 when it comes to updating the interactivity of buttons.

To try and debug, I ran the simplest possible setup with updating components, which i loosely adapted from [gradio docs](https://www.gradio.app/guides/blocks-and-event-listeners#updating-component-configurations):

inside `app.py`:

```python
import gradio as gr


def update_button(choice):
    print("Stepped into this function...")
    if choice == "short":
        print("short")
        return gr.Button(interactive=True)
    elif choice == "long":
        print("long")
        return gr.Button(interactive=True)
    else:
        print("other")
        return gr.Button(interactive=True)


with gr.Blocks() as demo:
    radio = gr.Radio(
        ["short", "long", "none"], label="Click any radio to activate the button"
    )
    button = gr.Button(value="Ask", interactive=False)
    radio.change(fn=update_button, inputs=radio, outputs=button)


demo.launch()
```

I then run `gradio app.py`. 

In the console, I can see the print statements appear, but the button remains inactive...

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

inside `app.py`:

```python
import gradio as gr


def update_button(choice):
    print("Stepped into this function...")
    if choice == "short":
        print("short")
        return gr.Button(interactive=True)
    elif choice == "long":
        print("long")
        return gr.Button(interactive=True)
    else:
        print("other")
        return gr.Button(interactive=True)


with gr.Blocks() as demo:
    radio = gr.Radio(
        ["short", "long", "none"], label="Click any radio to activate the button"
    )
    button = gr.Button(value="Ask", interactive=False)
    radio.change(fn=update_button, inputs=radio, outputs=button)


demo.launch()
```

I then run `gradio app.py`. 

### Logs

```shell
No visible errors from the console...


Running on local URL:  http://127.0.0.1:7860

To create a public link, set `share=True` in `launch()`.
Stepped into this function...
long
Stepped into this function...
other
Stepped into this function...
short
```
```


### System Info

```shell
System info:

python=3.10.10
gradio==4.0.0
gradio_client==0.7.0
```
```


### Severity

Blocking usage of gradio

PR #6266, opened 2023-11-02
Fix updating interactive prop

abidlabs:
There were two issues that prevented `.update(interactive=...)` from working:

1. We changed `interactive` to be a bool prop in the frontend, but the backend `.update()` logic had not been updated to reflect this 
2. We automatically set all output components to be interactive after the request was complete, this was overriding an explicit `.update(interactive-...)`. 

Working on a functional test for this, but otherwise is good for a review. Closes: #6197


dawoodkhan82 (approved):
Tested and fixes the issue! good catch on the setting all outputs to interactive bug.

abidlabs:
Thanks for the quick review @dawoodkhan82! Added a functional test as well, will merge in once CI is green

PR #6267, opened 2023-11-02
API to improve chatbot performance

aliabid94:
Our chatbot streaming doesn't have the best performance because it requires the entire chat history to be sent back and forth per response. This was an issue here (https://huggingface.slack.com/archives/C03JMRM52C9/p1698681324041859?thread_ts=1698679327.131709&channel=C03JMRM52C9&message_ts=1698681324.041859) and may be causing us to lose adoption for chat interfaces.
Was trying to come up with a better API that wouldn't require changing any reaking changes or changes to the client and here's what I came up with: a chatbot function, in addition to returning/yielding the entire chat history, has the option to just yield/return the next token. If the output is a string instead of the entire chat history, we assume this is the next token to append to the chat.

```python
    chatbot = gr.Chatbot()
    msg = gr.Textbox()
    clear = gr.Button("Clear")

    def user(user_message, history):
        return "", history + [[user_message, None]]

    def bot(history):
        for character in bot_message:
            time.sleep(0.05)
            yield character

    msg.submit(user, [msg, chatbot], [msg, chatbot], queue=False).then(
        bot, chatbot, chatbot
    )
```
Implemented simply in this PR, will flesh out the PR once we agree on this API. Try out chatbot_streaming to test current implementation.


abidlabs:
> Was trying to come up with a better API that wouldn't require changing any reaking changes or changes to the client and here's what I came up with: a chatbot function, in addition to returning/yielding the entire chat history, has the option to just yield/return the next token. If the output is a string instead of the entire chat history, we assume this is the next token to append to the chat.

You mention not having to make changes to the Clients, but we actually *should* make changes to the Client so that they can also support this. The Clients should show the concatenated messages in the same way that if we stream audio, the Client knows to combine it together

abidlabs:
Thinking about this some more, this resembles in many ways the behavior when `streaming=True` for an `Audio` output component. Would it make sense to use the same parameter here?

freddyaboulton:
I agree with @abidlabs that we should support "doing the right thing" in the client. Right now it's just returning a list of characters and the last character.

image

We need to do it in a way that's agnostic to the underlying component so that it also works for custom components. 
I like the idea of leveraging the existing streaming infrastructure. Maybe we can add a `StreamingOutput` data model. When clients receive it, they know they should concatenate all the results at the end of the stream. The end is maybe denoted by `stopped=True` flag on the data model. The reason we should introduce a new model is that we shouldn't use the FileData for something that's not a file.

That being said, it still wouldn't preserve the current behavior of returning the entire chat history since the client has no knowledge of past predictions. Is that something we want to enable? I guess we should for consistency but we're starting to add a lot of complexity to the client. How would that tie into curl usage for the API?

freddyaboulton:
This reminds me I don't think the new python client currently supports audio streaming 😅

aliabid94:
Ok I'm gonna close this issue for now bc this is going to be trickier than expected, and the chatbot complaints so far seem to be related to re-renders slowing down other components, or wanting richer features.

PR #6268, opened 2023-11-02
Fix various issues with demos on website

aliabd:
Fixes some issues with deprecated code on demos in the website. 

Also makes code interactive on playground


abidlabs:
Looks good @aliabd! Thanks for taking care of all of this. My only suggestions:

* Could we fix or remove the `stream_frames` demo so that the Spaces PR deploy has demos that are fully working? 
* Also would be to choose different demos for the landing page, ideally demos that are more connected to the zeitgeist in machine learning. Here are some potential options:

- `map_airbnb`
- `chatinterface_streaming_echo`
- A simplified version of `image_segmentation`
- A version of `fake_gan` called `diffusion_faces` with a nicer selection of images

Ok to do this in a separate PR, but I think it helps to have nice demos on the landing page

abidlabs (approved):

aliabd:
Thanks @abidlabs for the suggestions!  Made all the changes you requested. But will continue to clean up demos  in a separate pr.

Issue #6271, opened 2023-11-02
Image streaming with webcam is broken


aliabd:
### Describe the bug

Image streaming using webcam as a source is broken. It returns 

`ValueError: Image streaming only available if sources is ['webcam']. Streaming not supported with multiple sources.`

even though source is only webcam:

` gr.Image(sources=["webcam"], streaming=True)`

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

import gradio as gr
import numpy as np

def flip(im):
    return np.flipud(im)

demo = gr.Interface(
    flip, 
    gr.Image(sources=["webcam"], streaming=True), 
    "image",
    live=True
)
if __name__ == "__main__":
    demo.launch()
    

### Logs

```shell
File "/Users/aliabdalla/Desktop/dev/gradio/gradio/components/image.py", line 122, in __init__
    raise ValueError(
ValueError: Image streaming only available if sources is ['webcam']. Streaming not supported with multiple sources.
```

### System Info

```shell
gradio 4.0.2
```

### Severity

I can work around it

abidlabs:
Thanks @aliabd let me push a quick fix

abidlabs:
Closed via #6272

PR #6272, opened 2023-11-03
Fixes input `Image` component with `streaming=True`

abidlabs:
Quick fix for #6271


aliabd (approved):
thanks for the super quick fix @abidlabs! works great

PR #6285, opened 2023-11-03
Add migration tips to 4.0.0 changelog

abidlabs:
Fleshes out the changelog a bit more.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -138,7 +138,7 @@ To summarize migration:
 
 * For events that execute quickly or don't use much CPU or GPU resources, you should set `concurrency_limit=None` in Gradio 4.0. (Previously you would set `queue=False`.)
 * For events that take significant resources (like the prediction function of your machine learning model), and you only want 1 execution of this function at a time, you don't have to set any parameters.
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -15,25 +15,49 @@
 
 ## 4.0.0
 
-**v4.0 is a big release, so here are the main highlights:** (full list of changes below)
+### Highlights
```
freddyaboulton, line 18: Looks good! Should we also highlight changes to the client (no JSON files returned by label, chatbot)?

I think we should also put a link to this in the guides at least temporarily. It will increase visibility.

freddyaboulton (approved):

CHANGELOG.md
abidlabs: Thanks @freddyaboulton! I'll add the changes to the client.

Which guides are you thinking?

abidlabs:
I've added a section with the Client changes @freddyaboulton. Not certain where to link from in the Guides, perhaps we can do that in another PR as this is good to get out.

freddyaboulton:
Yep agreed

freddyaboulton:
cc @aliabd if you have thoughts

PR #6294, opened 2023-11-04
 fix regarding callable function error

SrijanSahaySrivastava:
## Description(Reopened)

catch if param_names = inspect.getfullargspec(self.fn)[0] doesn't get arguments.

raise error only when uncallable functions are called.

Closes: #5504


SrijanSahaySrivastava:
@abidlabs Please look into this, i've reopened this with base branch as new main.

abidlabs:
Thanks @SrijanSahaySrivastava, will take a look soon!

abidlabs:
Thanks @SrijanSahaySrivastava I tested the approach in the original PR, but it didn't work for me with `str`. I did a slightly more general fix, and added a test. I'll get this merged in once CI passes!

PR #6307, opened 2023-11-06
Provide status updates on file uploads

### LINKED ISSUE #6218: Improve /upload experience while file is still loading (?)
From the 4.0 tracking issue.

freddyaboulton:
## Description

Closes #6218

* Adds an optional upload_id query parameter to the `/upload` route. When specified, you can listen to progress updates on the `upload_progress` route. I originally wanted the `/upload` route to return the `upload_id` you can use to listen to updates but we can't send intermediate information from the `/upload` route without closing the request and the `/upload` route needs the request to stay open to stream the files.
* Adds a `UploadProgress.svelte` class in `@gradio/upload` that streams and displays progress updates from the server.
* The `upload_id` parameter is added to the `@gradio/client` but not the python client.

Try it here: https://huggingface.co/spaces/freddyaboulton/file_component

file_upload_preview

Questions:
* There's a bit of flicker if the uploaded file is really small. Should we only show progress for large files? My thought is no because it depends on the network speed.
* 

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/tired-berries-tease.md
+++ b/.changeset/tired-berries-tease.md
@@ -1,0 +1,7 @@
+---
+"@gradio/client": minor
+"@gradio/upload": minor
+"gradio": minor
+---
+
+feat:Provide status updates on file uploads
--- a/client/js/src/client.ts
+++ b/client/js/src/client.ts
@@ -156,7 +156,8 @@ interface Client {
 	upload_files: (
 		root: string,
 		files: File[],
-		token?: `hf_${string}`
+		token?: `hf_${string}`,
+		upload_id?: string
 	) => Promise<UploadResponse>;
 	client: (
 		app_reference: string,
@@ -208,7 +209,8 @@ export function api_factory(
 	async function upload_files(
 		root: string,
 		files: (Blob | File)[],
-		token?: `hf_${string}`
+		token?: `hf_${string}`,
+		upload_id?: string
 	): Promise<UploadResponse> {
 		const headers: {
 			Authorization?: string;
@@ -225,7 +227,10 @@ export function api_factory(
 				formData.append("files", file);
 			});
 			try {
-				var response = await fetch_implementation(`${root}/upload`, {
+				const upload_url = upload_id
+					? `${root}/upload?upload_id=${upload_id}`
+					: `${root}/upload`;
+				var response = await fetch_implementation(upload_url, {
 					method: "POST",
 					body: formData,
 					headers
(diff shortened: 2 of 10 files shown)


freddyaboulton:
I think this is in good enough shape to get some thoughts from people!

abidlabs:
This is awesome @freddyaboulton!

abidlabs:
I do notice the flicker as well when I test it out -- my thinking is that this means we should change the UI a bit. One idea would be instead of a flicker that covers the whole component, a % that shows up in place of the file size while the file is being uploaded

aliabid94:
> There's a bit of flicker if the uploaded file is really small. Should we only show progress for large files? My thought is no because it depends on the network speed.

We had this problem with StatusTracker where quick functions had a flicker of showing the progress bar and then dissapearing. I solved this by adding a transparency animation of 0.1s, so the progress bar quickly comes into view and leaves the view, but for very quick functions it basically never has time to transition into view at all. See here:

https://github.com/gradio-app/gradio/blob/5a745ed0842df19c9a93224f3df025765b51f79d/js/statustracker/static/index.svelte#L285

pngwn:
Another approach is to set a threshold, or 0.2s or so. If something takes less than the threshold, do nothing, otherwise show the animation.

aliabid94:
> Another approach is to set a threshold, or 0.2s or so. If something takes less than the threshold, do nothing, otherwise show the animation.

if it ends up taking 0.21 seconds, you'd still get a flicker though right?

pngwn:
Yah you would (although you could also animate it). But the same is kinda true with an animation, you can still get flashes if the timing is wrong, they are just a little gentler. I don't think there is a great way to solve it though.

There are APIs (in chrome) to check for networks speed / datasaving mode, but they aren't super informative.

aliabid94 (approved):
Code looks good to me! 
I was thinking about the design of the progress update requests being called directly from UploadProgress.svelte. In my mind, it would have been cleaner if the client handled tracking the upload progress (bc it's the client's responsibility to handle communication with the backend) and the client.upload method can take a callback function that it calls with upload progress updates. Don't feel too strongly about it though.

freddyaboulton:
Thanks for the reviews all! Implemented the opacity transition to reduce flicker. Going to merge but @hannahblair is going to open a PR to improve the progress bar styling later this week!

abidlabs:
Perhaps we should have avoided merging this in until the UI was ready? As it blocks releasing from main. Not a huge deal

pngwn:
Yeah. We should always keep main releasable (it also muddies the change log).

Issue #6304, opened 2023-11-06
Dropdown choices only returning first letter of each from list


UnreleasedDev:
### Describe the bug
Gradio version: 4.1.1

When using a dropdown to dynamically update another dropdown's choices based on the selected value, the gr.update function is only returning the first letter of each string in the list of choices. The expected behavior is to display it in full ie "JAN" not "J".

The issue does not occur in version 4.0.2 of Gradio for me. Downgrading to this version resolves the problem and the dropdown works as expected.

### Have you searched existing issues?  🔎

- [x] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

SITES_SUPPORTED = ['Bedford', 'Luton']
DATES_SUPPORTED = {'Bedford': ['JAN', 'FEB', 'MAR'], 'Luton': ['SEP', 'OCT', 'NOV']}

def update_dropdown_choices(site_selected):
    selected_dates = DATES_SUPPORTED[site_selected]
    print(selected_dates)
    return gr.update(choices=selected_dates, value=selected_dates[0], visible=True)

with gr.Blocks() as demo:
    sites_dropdown = gr.Dropdown(choices=SITES_SUPPORTED, label="Sites", interactive=True)
    dates_availability = gr.Dropdown(choices=[], label="Dates", interactive=True)
    sites_dropdown.select(update_dropdown_choices, [sites_dropdown], [dates_availability])

demo.launch()
```

### Screenshot

![](https://i.imgur.com/17DUe8E.png)

### System Info

```
Gradio Environment Information:
------------------------------
Operating System: Windows
gradio version: 4.1.1
gradio_client version: 0.7.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.1.2
fastapi: 0.104.1
ffmpy: 0.3.1
gradio-client==0.7.0 is not installed.
httpx: 0.25.1
huggingface-hub: 0.18.0
importlib-resources: 6.1.0
jinja2: 3.1.2
markupsafe: 2.1.3
matplotlib: 3.8.1
numpy: 1.26.1
orjson: 3.9.10
packaging: 23.2
pandas: 2.1.2
pillow: 10.1.0
pydantic: 2.4.2
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0.1
requests: 2.31.0
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.9.0
typing-extensions: 4.8.0
[...]

PR #6309, opened 2023-11-06
Fix updating choices in `gr.Dropdown` and updates related to other components

abidlabs:
Instead of sending the update dictionary directly to the frontend, we need to update it with the final values of the attributes from the Blocks instance. This approach should work generally, even for custom components.

Test with:

```py
import gradio as gr

SITES_SUPPORTED = ['Bedford', 'Luton']
DATES_SUPPORTED = {'Bedford': ['JAN', 'FEB', 'MAR'], 'Luton': ['SEP', 'OCT', 'NOV']}

def update_dropdown_choices(site_selected):
    selected_dates = DATES_SUPPORTED[site_selected]
    print(selected_dates)
    return gr.Dropdown(choices=selected_dates, value=selected_dates[0], visible=True)

with gr.Blocks() as demo:
    sites_dropdown = gr.Dropdown(choices=SITES_SUPPORTED, label="Sites", interactive=True)
    dates_availability = gr.Dropdown(choices=[], label="Dates", interactive=True)
    sites_dropdown.select(update_dropdown_choices, [sites_dropdown], [dates_availability])

demo.launch()
```

Fixes: #6304 

Will add tests but otherwise ready for review:
* I'll add some python unit tests for this method in `blocks.py`
* Will add a functional test because this behavior has broken several times in the past in different ways


abidlabs:
Added unit tests and a functional test -- should be good for review.

abidlabs:
The failing functional test is also happening on `main`. Started an [internal discussion](https://huggingface.slack.com/archives/C02QNREU24S/p1699309796664009)

gradio/blocks.py
```diff
@@ -382,13 +382,16 @@ def postprocess_update_dict(
...
     """
-    update_dict = {k: update_dict[k] for k in update_dict if hasattr(block, k)}
-    if update_dict.get("value") is components._Keywords.NO_VALUE:
-        update_dict.pop("value")
-    elif "value" in update_dict and postprocess:
-        update_dict["value"] = block.postprocess(update_dict["value"])
-        if isinstance(update_dict["value"], (GradioModel, GradioRootModel)):
-            update_dict["value"] = update_dict["value"].model_dump()
+    value = update_dict.pop("value", components._Keywords.NO_VALUE)
+    print("value", value)
```
dawoodkhan82, line 386: print statement

dawoodkhan82 (approved):
lgtm!

gradio/blocks.py
abidlabs, line 386:
```suggestion
```

abidlabs:

Issue #6297, opened 2023-11-04
pydantic error occurs when passing `gr.Dataframe`


hysts:
### Describe the bug

I get the following error when I call a function that takes `gr.Dataframe` as input. The error doesn't occur when using `gradio==3.50.2`.

```
Running on local URL:  http://0.0.0.0:7860

To create a public link, set `share=True` in `launch()`.
Traceback (most recent call last):
  File "/home/user/.local/lib/python3.10/site-packages/gradio/queueing.py", line 427, in call_prediction
    output = await route_utils.call_process_api(
  File "/home/user/.local/lib/python3.10/site-packages/gradio/route_utils.py", line 232, in call_process_api
    output = await app.get_blocks().process_api(
  File "/home/user/.local/lib/python3.10/site-packages/gradio/blocks.py", line 1482, in process_api
    inputs = self.preprocess_data(fn_index, inputs, state)
  File "/home/user/.local/lib/python3.10/site-packages/gradio/blocks.py", line 1262, in preprocess_data
    inputs_cached = block.data_model(**inputs_cached)  # type: ignore
  File "/home/user/.local/lib/python3.10/site-packages/pydantic/main.py", line 164, in __init__
    __pydantic_self__.__pydantic_validator__.validate_python(data, self_instance=__pydantic_self__)
pydantic_core._pydantic_core.ValidationError: 2 validation errors for DataframeData
metadata.display_value
  Input should be a valid list [type=list_type, input_value=None, input_type=NoneType]
    For further information visit https://errors.pydantic.dev/2.4/v/list_type
metadata.styling
  Input should be a valid list [type=list_type, input_value=None, input_type=NoneType]
    For further information visit https://errors.pydantic.dev/2.4/v/list_type
Traceback (most recent call last):
  File "/home/user/.local/lib/python3.10/site-packages/gradio/queueing.py", line 427, in call_prediction
    output = await route_utils.call_process_api(
  File "/home/user/.local/lib/python3.10/site-packages/gradio/route_utils.py", line 232, in call_process_api
    output = await app.get_blocks().process_api(
  File "/home/user/.local/lib/python3.10/site-packages/gradio/blocks.py", line 1482, in process_api
    inputs = self.preprocess_data(fn_index, inputs, state)
  File "/home/user/.local/lib/python3.10/site-packages/gradio/blocks.py", line 1262, in preprocess_data
    inputs_cached = block.data_model(**inputs_cached)  # type: ignore
[...]

PR #6310, opened 2023-11-06
Fix data model for `gr.DataFrame`

abidlabs:
Closes: #6297

Test with:

```py
import gradio as gr
import pandas as pd


def fn(df: pd.DataFrame) -> str:
    return df


df_orig = pd.DataFrame(data={"a": [1, 2, 3]})

with gr.Blocks() as demo:
    df = gr.Dataframe(value=df_orig, type="pandas")
    df2 = gr.Dataframe(value=df_orig, type="pandas")
    btn = gr.Button()

    btn.click(fn=fn, inputs=df, outputs=df2)

demo.launch()
```

or repro in original issue

### DIFF
--- a/gradio/components/dataframe.py
+++ b/gradio/components/dataframe.py
@@ -17,11 +17,9 @@ from gradio.events import Events
 
 
 class DataframeData(GradioModel):
-    headers: List[str]
+    headers: Optional[List[str]] = None
     data: List[List[Any]]
-    metadata: Optional[
-        Dict[str, List[Any]]
-    ] = None  # Optional[Dict[str, List[Any]]] = None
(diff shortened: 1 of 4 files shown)


gradio/components/dataframe.py
```diff
@@ -17,11 +17,9 @@
 
 
 class DataframeData(GradioModel):
-    headers: List[str]
+    headers: Optional[List[str]] = None
```
freddyaboulton, line 20: oo Interesting when did we make the headers optional?

abidlabs: actually you're right they're not, will revert this part.

hysts (approved):
Thanks! LGTM!

Issue #6298, opened 2023-11-05
gr.Theme.from_hub error


shpower:
### Describe the bug

 my_theme=gr.Theme.from_hub('xiaobaiyuan/theme_brief@>=1.0.0,<2.0.0')
   
When I run this line I got this error: raise ValueError(f"{space_info.id} is not a valid gradio-theme space!")
gradio version: 4.1.1

Very urgent! 

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

import gradio as gr

my_theme=gr.Theme.from_hub('xiaobaiyuan/theme_brief@>=1.0.0,<2.0.0')

### System Info

```shell
gradio version: 4.1.1
```

### Severity

Blocking usage of gradio

abidlabs:
Hmm looks like this might be an issue with the `huggingface_hub` library not parsing tags correctly, I'll push a temporary fix and then create an issue there. cc @Wauplin

PR #6311, opened 2023-11-06
Temporary fix to be able to load themes from Hub

abidlabs:
Commented out a check to make sure that the loaded theme has the correct tags, as it seems that the `huggingface_hub` library is not parsing tags from Spaces

Closes: #6298

### DIFF
--- a/.changeset/neat-cups-jump.md
+++ b/.changeset/neat-cups-jump.md
@@ -1,0 +1,5 @@
+---
+"gradio": patch
+---
+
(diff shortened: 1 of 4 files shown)


freddyaboulton (approved):
Looks good to me.

cc @Wauplin . Looks like something is up with getting tags from space readmes?

abidlabs:
Thanks @freddyaboulton for the review. Created an issue here: https://github.com/huggingface/huggingface_hub/issues/1806

PR #6312, opened 2023-11-06
WIP: Look into functional test flakes

freddyaboulton:
## Description

Closes: #(issue)

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/js/app/test/gallery_component_events.spec.ts
+++ b/js/app/test/gallery_component_events.spec.ts
@@ -6,13 +6,13 @@ test("Gallery preview mode displays all images correctly.", async ({
 	await page.getByRole("button", { name: "Run" }).click();
 	await page.getByLabel("Thumbnail 2 of 3").click();
 
-	await expect(await page.getByTestId("detailed-image").getAttribute("src")).toEqual(
-		"https://gradio-builds.s3.amazonaws.com/assets/lite-logo.png"
-	);
+	await expect(
+		await page.getByTestId("detailed-image").getAttribute("src")
+	).toEqual("https://gradio-builds.s3.amazonaws.com/assets/lite-logo.png");
 
-	await expect(await page.getByTestId("thumbnail 1").getAttribute("src")).toEqual(
-		"https://gradio-builds.s3.amazonaws.com/assets/cheetah-003.jpg"
-	);
+	await expect(
+		await page.getByTestId("thumbnail 1").getAttribute("src")
+	).toEqual("https://gradio-builds.s3.amazonaws.com/assets/cheetah-003.jpg");
 });
 
 test("Gallery select event returns the right value", async ({ page }) => {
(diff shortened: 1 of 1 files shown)


js/app/test/gallery_component_events.spec.ts
```diff
@@ -18,6 +18,7 @@ test("Gallery preview mode displays all images correctly.", async ({
 test("Gallery select event returns the right value", async ({ page }) => {
 	await page.getByRole("button", { name: "Run" }).click();
 	await page.getByLabel("Thumbnail 2 of 3").click();
+	await page.waitForTimeout(200);
```
freddyaboulton, line 21: So for some reason, playwright immediately says the value is `""`. Any thoughts why @pngwn ?

Adding a brief pause, or setting `queue=False` in the share event seems to make the problem go away.

abidlabs:
Very curious. Will merge this into the other open PRs so that they are unblocked

js/app/test/gallery_component_events.spec.ts
freddyaboulton: Also weird cause I swear this wasn't flaking when we first wrote this

js/app/test/gallery_component_events.spec.ts
```diff
@@ -18,6 +18,7 @@ test("Gallery preview mode displays all images correctly.", async ({
...
 	expect(await page.getByLabel("Select Data")).toHaveValue(
 		"https://gradio-builds.s3.amazonaws.com/assets/lite-logo.png"
 	);
```
pngwn: Could you try removing the timeout and doing this instead. See if the results are any different:

```suggestion
	await expect(page.getByLabel("Select Data")).toHaveValue(
		"https://gradio-builds.s3.amazonaws.com/assets/lite-logo.png"
	);
```

pngwn: Actually I've just checked the other tests and I'm almost certain this is the issue.

The way retries in Playwright work is that you pass in a `Locator` (with is what this `page.getByLabel(...)` call returns) and it can use that structure to requery if the value isn't ready yet. In order for those retries to take affect, we need to await the expectation itself rather than the locator.

Assuming this works, could you update the other tests in this file to match?

pngwn: That wasn't very clear I'll try again.

The `Locator` is a reference to some part of the dom. Its the `toHaveValue(...)` that retries. In order for it to not immediately return we need to await it, to give a chance to actually retry.

freddyaboulton: Ohhh ok. I'm pretty sure you have explained this to me before. Will fix. Thanks so much!

pngwn: Looking at the other test files, it seems the approach of "await everything just incase" seems to work well enough because there are definitely some unnecessary awaits in those but they don't seem too fléké.

Although i do sometime have issue with the streaming chatbot test locally (never in ci).

freddyaboulton:
Thanks for the review and explanation @pngwn ! Added the `await` outside. Going to merge now.

Issue #6284, opened 2023-11-03
Change default source in `Audio` and allow customizing the default source from a list of `sources` in `Image`, `Audio`, `Video`


abidlabs:
### Describe the bug

In our `Audio` component, the default source is microphone, but users might miss the fact that they can also upload files because the upload files icon is not as well-recognized ([internal discussion](https://huggingface.slack.com/archives/C02QZLG8GMN/p1699029301081169)). It would be better to make the default source `upload` because users are more likely to recognize the mic icon as allowing them to upload audio. 

At the same time, it would be good to allow users to change the default source if they'd like. We could add a separate parameter, or use a convention like the first source in the list of sources is the default source (personally prefer the latter over adding a parameter).

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

https://huggingface.co/spaces/distil-whisper/whisper-vs-distil-whisper

### System Info

```shell
gradio==4.0.2
```

### Severity

I can work around it

freddyaboulton:
> At the same time, it would be good to allow users to change the default source if they'd like. We could add a separate parameter, or use a convention like the first source in the list of sources is the default source (personally prefer the latter over adding a parameter).

Interesting, I thought that was already the behavior

abidlabs:
Oh it is indeed! I had no idea, very cool @hannahblair 

In that case, it would just be a matter of reversing the default list `["microphone", "upload"]`

pngwn:
Oh, that's clever. We should add this to image.

hysts:
The documentation says that the default order is `["upload", "microphone"]`
![](https://github.com/gradio-app/gradio/assets/25161192/4498e6a7-cd60-4dec-adc7-f82d5ae84f76)
but the actual order for now is not: 
https://github.com/gradio-app/gradio/blob/1d986217f6f4fc1829e528d2afe365635788204f/gradio/components/audio.py#L110

It would be nice if it's documented that the order matters.

hysts:
BTW, it would be nice if the small icon at the bottom was in the selected state by default. IMO, it would make it visually clearer that one can change the input source with these icons.

![](https://github.com/gradio-app/gradio/assets/25161192/fd4620da-171d-42f7-8b14-911782af57d2)

![](https://github.com/gradio-app/gradio/assets/25161192/bee040a0-5ff4-4b39-971f-008ca0cadd6b)

[...]

PR #6314, opened 2023-11-07
Improve default source behaviour in Audio

hannahblair:
## Description

- Notes that the default input source is the first element in the `sources` list
- Highlights the current active state in the source selection
- Sets `upload` as the default source if sources is `None`

Closes: #6284 

Screenshot 2023-11-07 at 13 51 18

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/polite-views-wear.md
+++ b/.changeset/polite-views-wear.md
@@ -3,4 +3,4 @@
 "gradio": patch
 ---
 
-fix:Improve default audio behaviour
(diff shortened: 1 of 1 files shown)


abidlabs (approved):
Sweet! lgtm 

Is the same behavior true for Video and Image as well? Would be good to note in their docstrings as well. (can be a separate pr)

Issue #6315, opened 2023-11-07
Apply formatting to a Dataframe's subset of columns hangs the page


clefourrier:
### Describe the bug

When trying to apply formatting **to a subset of columns** in a gradio.Dataframe component using the styling from pandas.Dataframe, the gradio page hangs, then crashes.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

1. Create a pandas dataframe with mixed column types
2. Apply `df = df.style.format("{:.2%}", subset=numeric_cols)` on your numeric columns
3. Load `df` in a Gradio dataframe component
4. Launch the gradio app.

If you remove the styling part, the dataframe should display without problem

### Screenshot

Page gets stuck
![image](https://github.com/gradio-app/gradio/assets/22726840/f97341c4-2322-435f-b321-0496c9b31860)

When finally loading (20 sec for 5 rows), Dataframe is empty
![image](https://github.com/gradio-app/gradio/assets/22726840/20e80b16-4cf8-472d-bd17-42cb11964e16)

Then the page crashes.

### Logs
Nothing in the log after 

```shell
Running on local URL:  http://127.0.0.1:7860

To create a public link, set `share=True` in `launch()`.
```
even when launching with debug=True

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Linux
gradio version: 4.1.1
gradio_client version: 0.7.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.1.0
altair: 4.2.2
fastapi: 0.104.1
ffmpy: 0.3.0
gradio-client==0.7.0 is not installed.
httpx: 0.24.0
huggingface-hub: 0.18.0
importlib-resources: 6.0.1
jinja2: 3.1.2
markupsafe: 2.1.2
matplotlib: 3.7.1
numpy: 1.24.2
orjson: 3.8.10
packaging: 23.1
pandas: 2.0.0
pillow: 9.5.0
pydantic: 2.4.2
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0
requests: 2.28.2
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.9.0
typing-extensions: 4.8.0
uvicorn: 0.21.1
websockets: 11.0.1
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2023.6.0
[...]

PR #6318, opened 2023-11-07
Fix for stylized DataFrame

abidlabs:
Fixes a regression: https://github.com/gradio-app/gradio/issues/6315

Sample code:

```py
import pandas as pd 
import gradio as gr

# Creating a sample dataframe
df = pd.DataFrame({
    "A" : [14, 4, 5, 4, 1], 
    "B" : [5, 2, 54, 3, 2], 
    "C" : [20, 20, 7, 3, 8], 
    "D" : [14, 3, 6, 2, 6], 
    "E" : [23, 45, 64, 32, 23]
}) 

# Applying style to highlight the maximum value in each row
styler = df.style.highlight_max(color = 'lightgreen', axis = 0)

# Displaying the styled dataframe in Gradio
with gr.Blocks() as demo:
    gr.DataFrame(styler)
    
demo.launch()
```

Explanation: the introduction of

```js
		on:change={(e) => (value = e.detail)}
```

was causing an infinite loop when working with metadata. This fixes it, but as I'm not sure if this is the best fix, feel free to suggest an alternative. But basically here, we check to see if a Dataframe is interactive before updating the value. I've tested with `matrix_transpose`, which continues to work. 

I also added a functional test since neither the backend nor the frontend tests alone caught this.

### DIFF
--- a/.changeset/floppy-crews-do.md
+++ b/.changeset/floppy-crews-do.md
@@ -1,0 +1,6 @@
+---
+"@gradio/dataframe": minor
+"gradio": minor
+---
+
+feat:Fix for stylized DataFrame
--- a/js/app/test/dataframe_colorful.spec.ts
+++ b/js/app/test/dataframe_colorful.spec.ts
@@ -1,0 +1,11 @@
+import { test, expect } from "@gradio/tootils";
+
+test("first couple of cells in table are highlighted correctly", async ({
+	page
+}) => {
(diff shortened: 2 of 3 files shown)


abidlabs:
Thanks @hannahblair I'll merge this in so that we can get a release out with the fix. 

@pngwn if you get a chance to take a look at this, would appreciate it! I think there must be a better way to do this.

Issue #6303, opened 2023-11-06
textbox cannot reforcus after submit


Da-Capo:
### Describe the bug

https://www.gradio.app/docs/chatbot#demos
this textbox cannot reforcus after submit at gradio 4
gradio3 can reforcus to textbox

### Have you searched existing issues?  🔎

- [ ] I have searched and found no existing issues

### Reproduction

1

### System Info

```shell
1
```

### Severity

I can work around it

abidlabs:
Hi @Da-Capo can you please provide a repro? I don't see the issue you are describing with this demo:

```py
import gradio as gr

def echo(message, history):
    return message

demo = gr.ChatInterface(fn=echo, examples=["hello", "hola", "merhaba"], title="Echo Bot")
demo.launch()
```

on 4.0

Da-Capo:
https://github.com/gradio-app/gradio/assets/10624004/57bcfde8-522a-49fc-a994-22f5b0ccfcf5

In this video, left is 3.42  right is 4.1.1

in 3.42 i can continue edit the textbox after submit,
but in 4.1.1 i need another click to focus it.

It means that each chat requires an additional click.

congchan:
I faced the same issue, any progress on it?

Issue #6233, opened 2023-11-01
focused & unfocused/blurred event listeners for Code components


rsr23:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
the `gradio.Textbox.focus(fn, ···)` and `gradio.Textbox.blur(fn, ···)` event listeners are great for textbox Components.
But similar event listeners are not available for Code components.

**Describe the solution you'd like**  
Event listeners like `gradio.Code.focus(fn, ···)` and `gradio.Code.blur(fn, ···)`

**Additional context**  
None

abidlabs:
Thanks @rsr23! Would you like to open a PR for this?

abidlabs:
Closed via https://github.com/gradio-app/gradio/pull/6323

PR #6323, opened 2023-11-07
Textbox and Code Component Blur/Focus Fixes

dawoodkhan82:
## Description
- Re-focus textbox after submit
- Add focus and blur events to code component

Closes: #6303 & #6233

## 🎯 PRs Should Target Issues

## Tests


js/textbox/shared/Textbox.svelte
```diff
@@ -68,6 +68,7 @@
 			scroll();
 		}
 		value_is_output = false;
+		el.focus();
```
abidlabs, line 71: Should this be at line 114 instead? Right now, it causes any change in the textbox value to focus on the textbox. Which means, e.g. if you launch 

```py
import gradio as gr

gr.Interface(lambda x:x, "textbox", "textbox").launch()
```

it focuses on the input textbox, which is not desired

abidlabs:
I'm not able to get the `.blur()` and `.focus()` events to fire. Here's my code:

```py
import gradio as gr

def blur():
    print("blur")
    
def focus():
    print("focus")
    
def change():
    print("change")
    
with gr.Blocks() as demo:
    c = gr.Code()
    c.blur(blur, None, None)
    c.focus(focus, None, None)
    c.change(change, None, None)
    
demo.launch()
```

Note that `.change()` fires fine -- am I missing something?

abidlabs:
Will take a look! Maybe we add a unit test for the blurring behavior?

abidlabs:
Tested with the same demo as above, couple of issues:

* `blur()` event fires when the demo is first loaded
* the `focus()` event fires every time I type a character into the `Code` component
* In my tests, #6303 is not fixed.

I just tested:

```py
import gradio as gr

def echo(message, history):
    return message

demo = gr.ChatInterface(fn=echo, examples=["hello", "hola", "merhaba"], title="Echo Bot")
demo.launch()
```

after submitting, the textbox does not get re-focused

dawoodkhan82:
Not sure what caused the regression for the textbox re-focus issue, but this pr should fix it. Also the code component blur and focus should work as expected now, although sometimes it misfires for me when testing. Let me know if you notice the same @abidlabs

abidlabs:
LGTM working well for me @dawoodkhan82!

VigneshHexo:
> LGTM working well for me @dawoodkhan82!

Could you please let me know, how can I accomplish the same re-focus feature while building Custom Chatbot with Gradio blocks? For example, here is a sample code snippet

```python
import gradio as gr
import random
import time

with gr.Blocks() as demo:
    chatbot = gr.Chatbot()
    msg = gr.Textbox()

        time.sleep(2)
        return "", chat_history
    
if __name__ == "__main__":
    demo.queue()
    demo.launch(show_api=False, share=False, inline=False)
```

PR #6326, opened 2023-11-07
Fix Model3D template

freddyaboulton:
## Description

Can't repro but some people on WSL reported trouble templating Model3D. This has been verified to fix it.

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/vast-gifts-win.md
+++ b/.changeset/vast-gifts-win.md
@@ -1,0 +1,5 @@
+---
+"gradio": minor
+---
+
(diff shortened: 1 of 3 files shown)


abidlabs:
Looks good @freddyaboulton! Nit: for the demo code, I think we should use Interface, instead of Blocks as `model3D` can be used for input as well.

abidlabs (approved):

abidlabs:
Will go ahead and merge so that we can do the release

Issue #6283, opened 2023-11-03
URL query parameters with queue is broken


radames:
### Describe the bug

Since this PR we were able to get the query params via `gr.request`, however some new bug was introduce and on version `4.0.2`. If you disable the queue the request comes with expected query data

https://github.com/gradio-app/gradio/issues/5485

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

The workaround is to disable the queue

```python
import gradio as gr

def echo(params):
    print(params)
    return params

def get_params(request: gr.Request):
  params = request.query_params
  ip = request.client.host
  return {"params": params, 
          "ip": ip}

  
with gr.Blocks() as demo:
  url_params = gr.State()
  text_in = gr.Textbox()
  text_out = gr.JSON()
  btn = gr.Button()
  btn.click(echo, inputs=[url_params], outputs=[text_out])
  demo.load(get_params, None, url_params, queue=True)
  #demo.load(get_params, None, url_params, queue=False)

demo.launch()
```

### System Info

```shell
Gradio 4.0.2
```

### Severity

I can work around it

abidlabs:
Looks like there was a regression, will take a look!

PR #6327, opened 2023-11-07
Restore query parameters in request

aliabid94:
We lost query parameters in the request in upgrading to 4.0, because we were sending the queue information (session_hash, fn_index) via query params - now we send both URL params and queue information via request.url_params. 
This does mean that when a user searches the request for url_params, there will be additional url_parameters they will find now in 4.0. I wanted to strip these but the fastapi.Request object is immutable and would be very messy / hacky to create a new object without these URL parameters. I think this is okay because users are usually looking for a specific key when searching url parameters.

Fixes: #6283 

Try with

```python
import gradio as gr

def echo(params):
    print(params)
    return params

def get_params(request: gr.Request):
  params = request.query_params
  ip = request.client.host
  return {"params": params, 
          "ip": ip}

  
with gr.Blocks() as demo:
  url_params = gr.State()
  text_in = gr.Textbox()
  text_out = gr.JSON()
  btn = gr.Button()
  btn.click(echo, inputs=[url_params], outputs=[text_out])
  demo.load(get_params, None, url_params, queue=True)
  #demo.load(get_params, None, url_params, queue=False)

demo.launch()
```

### DIFF
--- a/client/js/src/client.ts
+++ b/client/js/src/client.ts
@@ -460,8 +460,7 @@ export function api_factory(
 						});
 
 						post_data(
-							`${http_protocol}//${resolve_root(host, config.path, true)}/run${
-								_endpoint.startsWith("/") ? _endpoint : `/${_endpoint}`
+							`${http_protocol}//${resolve_root(host, config.path, true)}/run${_endpoint.startsWith("/") ? _endpoint : `/${_endpoint}`
 							}${url_params ? "?" + url_params : ""}`,
 							{
 								...payload,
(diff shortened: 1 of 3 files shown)


abidlabs:
LGTM @aliabid94

abidlabs (approved):

client/js/src/client.ts
```diff
@@ -657,7 +657,7 @@ export function api_factory(
 								host,
 								config.path,
 								true
-							)}/queue/join?${params}`
+							)}/queue/join?${params}${url_params ? "&" + url_params : ""}`
```
abidlabs, line 660: Should we put `url_params` before `params`. Right now, if `url_params` includes e.g. `fn_index`, it'll override the `param` and cause the request to fail

aliabid94: done

liquidcarbon:
this is currently not working:
`TypeError: Type is not JSON serializable: QueryParams`

Issue #6220, opened 2023-11-01
The windows tests are only on release


freddyaboulton:
Coming from #5564 .

@pngwn and I talked about running the windows tests only on the release branch by changesets to speed up CI and reduce noise.

abidlabs:
Good idea. Also, the [CI job for flaky Windows tests always passes, even when the tests themselves fail](https://huggingface.slack.com/archives/C02QNREU24S/p1698754876557979).

PR #6330, opened 2023-11-07
Fix backend tests on CI

abidlabs:
* Ensures that CI on Windows fails if any of the pytests fail
* Ensures that both python client tests and core gradio tests run when either are changed


abidlabs:
Removing `coverage` seems to fix the issues for Windows CIs. Now, they "successfully fail", as you can see here:

* https://github.com/gradio-app/gradio/actions/runs/6805412176/job/18504878095?pr=6330
* https://github.com/gradio-app/gradio/actions/runs/6805412176/job/18504878322?pr=6330

abidlabs:
Fixing this revealed several tests that are broken on Windows: https://github.com/gradio-app/gradio/actions/runs/6805506600/job/18505174184?pr=6330 -- let me fix those

abidlabs:
Also, as pointed out by @aliabid94 recently, we should run the python client tests when the core library changes and vice versa, since they can affect each other

abidlabs:
>  Also, as pointed out by @aliabid94 recently, we should run the python client tests when the core library changes and vice versa, since they can affect each other (now that we release both together)

Note that this will actually slow down the CI a little bit but we can get a significant speed up once we follow this PR with one addressing https://github.com/gradio-app/gradio/issues/6220

pngwn (approved):
Looks good to me!  Thanks for figuring this one out @abidlabs!

abidlabs:
Thanks @pngwn appreciate the review!

freddyaboulton (commented):
Thanks @abidlabs !

PR #6343, opened 2023-11-08
Fix audio streaming output issues in 4.0

aliabid94:
Streaming audio out broke in 4.0, as can be seen in the `stream_audio_out` demo. The waveform generator does not support a streaming audio source, so did a fallback onto a regular audio component. Test this via `stream_audio_out` demo.

There was another issue that whenever the DOM re-rendered for any reason, the waveform would have to regenerate, causing it to fetch the audio source on every DOM re-render. This is because normalize_files caused a new JSON to be passed to the AudioPlayer, so any DOM change caused a re-render. Now only re-renders if the URL changes. Test this via demo below while watching the Network Chrome inspector:

```python
import gradio as gr


def greet(audio):
    for i in range(10):
        import time
        time.sleep(1)
        yield i


with gr.Blocks() as demo:
    name = gr.Audio(label="Name")

if __name__ == "__main__":
    demo.launch()
```

### DIFF
--- a/.changeset/red-ties-turn.md
+++ b/.changeset/red-ties-turn.md
@@ -1,0 +1,6 @@
+---
+"@gradio/audio": minor
+"gradio": minor
+---
+
+feat:Fix audio streaming output issues in 4.0
--- a/gradio/processing_utils.py
+++ b/gradio/processing_utils.py
@@ -327,7 +327,6 @@ def resize_and_crop(img, size, crop_type="center"):
 
 def audio_from_file(filename, crop_min=0, crop_max=100):
     try:
-        print(">>>", filename)
         audio = AudioSegment.from_file(filename)
     except FileNotFoundError as e:
         isfile = Path(filename).is_file()
--- a/gradio/processing_utils.py
+++ b/gradio/processing_utils.py
@@ -327,6 +327,7 @@ def resize_and_crop(img, size, crop_type="center"):
 
 def audio_from_file(filename, crop_min=0, crop_max=100):
     try:
+        print(">>>", filename)
         audio = AudioSegment.from_file(filename)
     except FileNotFoundError as e:
         isfile = Path(filename).is_file()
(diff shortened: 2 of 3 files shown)


abidlabs:
Will take a look! But could we add some tests to prevent regressions in the future? I think an e2e playwright test is warranted here

abidlabs:
> There was another issue that whenever the DOM re-rendered for any reason, the waveform would have to regenerate, causing it to fetch the audio source on every DOM re-render. This is because normalize_files caused a new JSON to be passed to the AudioPlayer, so any DOM change caused a re-render. Now only re-renders if the URL changes

Ah nice, I had noticed this issue in the Spaces deploy before

abidlabs:
> Test this via stream_audio_out demo

The output streaming isn't quite working for me. I get the behavior that some other users have described, which is that the audio file doesn't play until the full audio has been streamed. I've tried a few different things:

* Trying audio files of different lengths
* Hitting the play button manually vs. waiting for it to autoplay
* Decreasing the chunk size in the demo

I'll upload a video with the behavior I'm seeing. This is the 60 second audio file I'm trying: https://www2.cs.uic.edu/~i101/SoundFiles/CantinaBand60.wav

abidlabs:
Synced with @aliabid94, I was testing wrong. LGTM once we can fix the static checks and potentially add tests

abidlabs (approved):

abidlabs:
Don't love the fallback UI, but we should get this in so that streaming works again. Perhaps @hannahblair you could take a look at the UI in a separate PR?

aliabid94:
Adding playwright tests to this is going to be quite tricky because streaming is handled by the browser, not any javascript, so I'm not sure what we can specifically test is being done correctly in streaming. As far as the JS is aware, the HTML is set to a single source the same way a non-streaming audio file is.

aliabid94:
@hannahblair if you have ideas to make the streaming audio player prettier than just using the default player, might be something you can take a look at!

abidlabs:
Sounds good, let's merge this in for now

PR #6347, opened 2023-11-08
Fix `colorFrom` in theme space readmes

abidlabs:
As pointed out by @coyotte508, we use an invalid  `colorFrom` in the theme space readmes, which is preventing theme spaces from returning the tags correctly. 

This changes `orange` to `red`

cc @freddyaboulton @Wauplin


freddyaboulton (approved):

abidlabs:
Thanks @freddyaboulton

PR #6356, opened 2023-11-09
Redesign file upload

hannahblair:
## Description

Bar in the background total progress of all files, and a circular progress bar next to the file name shows progress of file currently uploading and changes to the next one once that's done. 

Screenshot 2023-11-09 at 12 42 58

Screenshot 2023-11-09 at 12 43 07

Kapture 2023-11-09 at 11 51 27

## 🎯 PRs Should Target Issues

## Tests


abidlabs:
Looks super nice @hannahblair! We noticed a couple of issues when we were testing it in sync:

* In the `Image` component (and maybe the other components as well), there's a shaded progress bar that appears in the background of the component, it looks a little weird and starts about halfway in the screen. Its not necessary since we have the spinner, would be great if we could remove it: 

image

* After the image is uploaded, sometimes the Image component collapses before showing the image. Would be great to remove that

* It looks like not all of the file names are shown when uploading multiple files. It seems that at least the first file name is skipped. Not sure if this is a frontend issue or some info missing from the `/upload` route. cc @freddyaboulton

freddyaboulton:
I removed the background, fixed the "skip the first filename" issue, and removed the flicker caused by transitioning from UploadProgress -> Upload -> Original component.

@hannahblair this looks good to me! Feel free to merge if you're happy with the changes 

https://github.com/gradio-app/gradio/assets/41651716/f8a88236-9c69-4302-b770-42103d96f988

freddyaboulton (approved):

freddyaboulton:
Ah ok, maybe what happened was that since the first file uploaded super quickly, the shaded part started at the half of the screen and we got confused.

I think the spinner is also showing the total progress and not the progress for the current file. But I think the main takeaway I got from the meeting was that we preferred the spinner over the background shade as a measure of total upload progress.

freddyaboulton:
Do you have thoughts on the above @pngwn @aliabid94 @dawoodkhan82 (cc @gary149 )?

In summary - which of the following should we go with?

1) the shaded background show total progress and the wheel show individual file progress
2) The wheel shows total progress (what it is now)
3) The wheel shows individual file progress

pngwn:
I like the spinner for total progress with the file name changing as it progresses with no shaded background.

PR #6363, opened 2023-11-09
Fix image upload

freddyaboulton:
## Description

Fix image upload

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/js/image/Index.svelte
+++ b/js/image/Index.svelte
@@ -122,7 +122,7 @@
 
 		<ImageUploader
 			bind:active_tool
-			bind:value={_value}
+			bind:value
 			selectable={_selectable}
 			{root}
 			{sources}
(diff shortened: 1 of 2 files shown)


js/image/Index.svelte
```diff
@@ -122,7 +122,7 @@
 
 		<ImageUploader
 			bind:active_tool
-			bind:value={_value}
+			bind:value
```
freddyaboulton, line 125: What was happening is that we were binding to `_value` so the `value` prop (which is what gets passed to events) was never updated during the upload.

I think this is the right thing to do. By binding to value, _value will automatically be updated. Also the ImageUploaded calls `normalise_file` so value will have the right url

freddyaboulton: Both of the test cases from #6322 work with this change as far as I can tell

image

image

abidlabs:
LGTM thanks @freddyaboulton for the nice fix!

abidlabs (approved):

freddyaboulton:
Thanks for the reivew @abidlabs !!

PR #6368, opened 2023-11-10
Fix component update bug

freddyaboulton:
## Description

Fixes: #6342

The problem was that if the component was initialized for the first time in an event listener, __init__ was not being called. This was a bug that was in version 3 too. 

The reason the bug doesn't "surface" for a normal generator is that "in_event_listener" returns false - probably because it's running in a separate thread (to avoid blocking the event loop) - so the __init__ is always called.

The fix is to call init if it hasn't been called before in an event listener

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/early-geese-relate.md
+++ b/.changeset/early-geese-relate.md
@@ -1,0 +1,5 @@
+---
+"gradio": patch
+---
+
+feat:Fix component update bug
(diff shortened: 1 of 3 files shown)


abidlabs:
Niceee fix @freddyaboulton! LGTM

abidlabs (approved):

Issue #6371, opened 2023-11-11
sources parameter error


ShamHyper:
### Describe the bug

When I add the sources="upload" parameter to gradio.Image(), the Gradio tab cannot load

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```py
import gradio as gr
with gr.Blocks() as example:
    with gr.Tab("example"):
        with gr.Row():
            image_input = gr.Image(sources="upload", width=200, height=400)
example.launch()
```

### Screenshot

![image](https://github.com/gradio-app/gradio/assets/104412752/9e1cbcfc-5453-4a4f-8558-34651ac4bc1b)
Error in DevTools console

### Logs

```shell
debug=True did not provide me with information :)
```

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Windows
gradio version: 3.42.0
gradio_client version: 0.5.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.1.1
fastapi: 0.103.1
ffmpy: 0.3.1
gradio-client==0.5.0 is not installed.
httpx: 0.25.0
huggingface-hub: 0.17.2
importlib-resources: 6.0.1
jinja2: 3.1.2
markupsafe: 2.1.3
matplotlib: 3.8.0
numpy: 1.24.3
orjson: 3.9.7
packaging: 23.1
pandas: 2.1.0
pillow: 10.0.0
pydantic: 1.10.13
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0.1
requests: 2.31.0
semantic-version: 2.10.0
typing-extensions: 4.5.0
uvicorn: 0.23.2
websockets: 11.0.3
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2023.9.1
httpx: 0.25.0
huggingface-hub: 0.17.2
packaging: 23.1
requests: 2.31.0
typing-extensions: 4.5.0
websockets: 11.0.3
```

### Severity

Blocking usage of gradio

ShamHyper:
Oh, this code doesn't reproduce the problem. Maybe the problem is elsewhere, but the bottom line is that sources="upload" is really causing the problem.

I can show the full code to reproduce the problem... just add to any gr.Image() source and it will break...

ShamHyper:
[...]

Issue #6337, opened 2023-11-08
Page hangs when `sources` for `gr.Image` is provided as a string instead of a list


quillbit:
### Describe the bug

When I launch the program, I try to open the page using my browser, but the page keeps loading. I opened the browser developer tools and found the console output:
```
Uncaught (in promise) TypeError: sources.sort is not a function
    at $$self.$$.update (ImageUploader.svelte:68:30)
    at init (svelte.js:3246:5)
    at new ImageUploader (ImageUploader.svelte:101:26)
    at Array.create_default_slot_1 (Index.svelte:149:21)
    at create_slot (svelte.js:165:23)
    at create_dynamic_element (index.js:73:1)
    at create_fragment$1 (Block.svelte:22:11)
    at init (svelte.js:3250:34)
    at new Block (Block.svelte:18:52)
    at Array.create_else_block (Index-0f548159.js:50:10)
```

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### System Info

```shell
gradio >= 4.1.1
```

### Severity

I can work around it

abidlabs:
Hi @quillbit can you provide a more detailed reproduction? What page are you opening?

quillbit:
> 嗨，你能提供更详细的复制品吗？您正在打开哪个页面？

![1](https://github.com/gradio-app/gradio/assets/130726845/45221c2d-a0cc-432a-9806-c10bf9ca39a6)

abidlabs:
What is the code for this Gradio app? We need a minimal code example that we can use to reproduce the issue above. See: https://stackoverflow.com/help/minimal-reproducible-example

quillbit:
> 这个 Gradio 应用程序的代码是什么？我们需要一个最小的代码示例，我们可以使用它来重现上述问题。请参见：https://stackoverflow.com/help/minimal-reproducible-example

ok this is my sample code：
```python
import gradio as gr

def test(image_file: str):
    return image_file

with gr.Blocks() as demo:
    with gr.Row():
        with gr.Column():
            with gr.Row():
                inputs_image = gr.Image(image_mode="RGB", sources="upload", type="filepath")
            with gr.Row():
                btn = gr.Button(variant="primary")

        with gr.Column():
            with gr.Row():
                outputs_image = gr.Image(type="pil")
        
    btn.click(fn=test, inputs=[inputs_image], outputs=[outputs_image])


demo.queue().launch(inbrowser=True, server_name="0.0.0.0")
```

[...]

PR #6378, opened 2023-11-11
Allows `sources` to be a string for `gr.Image`

abidlabs:
A typo was causing the frontend to crash if `sources` was specified as a string

Closes: #6371 
Closes: #6337

### DIFF
--- a/.changeset/eleven-windows-relate.md
+++ b/.changeset/eleven-windows-relate.md
@@ -1,5 +1,5 @@
 ---
-"gradio": minor
+"gradio": patch
 ---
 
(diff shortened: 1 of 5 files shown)


pngwn (approved):
Looks good! We should change this release to a minor as this is a feature addition rather than a bug fix. (Neither of the issues linked are bugs despite being labelled so).

Thanks @abidlabs!

aliabid94 (approved):
LGTM

abidlabs:
Thanks guys! Will update the label and merge in

Issue #6329, opened 2023-11-07
Chatbot avatar image not loading in chatbot_multimodal demo


freddyaboulton:
### Describe the bug

The chatbot avatar image is not loading

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

send a message to the bot in the `chatbot_multimodal` demo and observe the avatar image.

### Screenshot

<img width="1557" alt="image" src="https://github.com/gradio-app/gradio/assets/41651716/61292b12-11f3-4b53-9366-2d9551ce71ca">

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Darwin
gradio version: 4.1.1
gradio_client version: 0.7.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.1.0
altair: 5.0.1
fastapi: 0.100.0
ffmpy: 0.3.0
gradio-client==0.7.0 is not installed.
httpx: 0.24.1
huggingface-hub: 0.16.4
importlib-resources: 6.1.1
jinja2: 3.1.2
markupsafe: 2.1.3
matplotlib: 3.7.2
numpy: 1.25.1
orjson: 3.9.2
packaging: 23.1
pandas: 2.0.3
pillow: 10.0.0
pydantic: 2.4.2
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0
requests: 2.31.0
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.9.0
typing-extensions: 4.7.1
uvicorn: 0.22.0
websockets: 11.0.3
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2023.6.0
httpx: 0.24.1
huggingface-hub: 0.16.4
packaging: 23.1
requests: 2.31.0
typing-extensions: 4.7.1
websockets: 11.0.3
```

### Severity

I can work around it

abidlabs:
Ah this is because of the new file policy. We'll need to move the avatar images to the cache as well

freddyaboulton:
But the demo is setting allowed_paths?

abidlabs:
Because `allowed_paths` is relative to the working directory, you have to run the demo from the `demo/chatbot_multimodal/` folder for it to work:

[...]

PR #6379, opened 2023-11-11
Processes `avatar_images` for `gr.Chatbot` and `icon` for `gr.Button` correctly, so that respective files are moved to cache

abidlabs:
Closes: https://github.com/gradio-app/gradio/issues/6329

Test with `chatbot_multimodal` demo


aliabid94 (approved):
LGTM. Seems like it would be cleaner to move `move_resource_to_block_cache` to be a `Block` method, instead of having to pass `self` everywhere it's called.

abidlabs:
Thanks @aliabid94. That's a good idea, I'll make the change

abidlabs:
I'll refactor it so that the `Block` class has its own `move_resource_to_block_cache()` instance method, but I'll also leave the original `move_resource_to_block_cache` method in `postprocessing_utils.py` for backwards compatibility for any custom components created in Gradio 4.2.0 or earlier.

aliabid94:
>> I'll also leave the original move_resource_to_block_cache method in postprocessing_utils.py for backwards compatibility

Then have postprocessing_utils.move_resource_to_block_cache just call the block.move_resource_to_block_cache?

abidlabs:
Made the refactor, and then also included a test using the `gradio_pdf` custom component that @freddyaboulton created to ensure backwards compatibility. Will merge after CI is green. Thanks for the review @aliabid94!

PR #6380, opened 2023-11-11
Makes the Windows tests only run on the release PR

abidlabs:
We now have a `windows-tests` labels that can be used to control whether the backend Windows tests should be run or not. By default, this label is not applied to PRs, which is why the Windows tests have not run on this PR. 

Adding the label will cause the Windows tests to run as well. On the release PR, this label gets added automatically so all tests run. Fixes: https://github.com/gradio-app/gradio/issues/6220

Overall speeds up the CI significantly (took 6 minutes for all tests to run).

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -51,6 +51,8 @@ jobs:
         os: ["ubuntu-latest", "windows-latest"]
         test-type: ["not flaky", "flaky"]
         python-version: ["3.8"]
+        exclude:
+          - os: ${{ github.event_name == 'pull_request' && contains( github.event.pull_request.labels.*.name, 'windows-tests') && 'dummy' || 'windows-latest' }}
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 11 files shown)


.github/workflows/backend.yml
```diff
@@ -51,6 +52,8 @@ jobs:
         os: ["ubuntu-latest", "windows-latest"]
         test-type: ["not flaky", "flaky"]
         python-version: ["3.8"]
+        exclude:
+          - os: ${{ github.event_name == 'pull_request' && contains( github.event.pull_request.labels.*.name, 'windows-tests') && 'dummy' || 'windows-latest' }}
```
abidlabs, line 56: This is a cool way to conditionally exclude cases from the matrix. See https://stackoverflow.com/a/73822998

pngwn (approved):
Looks good! Thanks @abidlabs!

abidlabs:
Thanks @pngwn appreciate the quick review!

Issue #6367, opened 2023-11-10
evt.target incorrectly assigned with gr.on listener method


srednors:
### Describe the bug

When you add the event data class (evt) as a type hint to an argument in the event listener function ([Gathering Event Data](https://www.gradio.app/guides/blocks-and-event-listeners#gathering-event-data)), evt.target always gets assigned to the first component in the triggers list if using the gr.on listener method ([Binding Multiple Triggers to a Function](https://www.gradio.app/guides/blocks-and-event-listeners#binding-multiple-triggers-to-a-function)), regardless of the actual component from that list that is triggering the event.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```
import gradio as gr

with gr.Blocks() as demo:
    text1 = gr.Textbox(label="Textbox 1")
    text2 = gr.Textbox(label="Textbox 2")
    text = gr.Textbox(label="Output")

    def test(evt: gr.SelectData):
        return 'You selected some text in ' + evt.target.label
 
        # Output textbox displays 'You selected some text in Textbox 1' 
        # regardless of which textbox the selection is actually occuring in.

    gr.on(
        triggers=[text1.select, text2.select],
        fn=test,
        outputs=text
    )
    
demo.launch()
```

### Screenshot

<img width="648" alt="Screenshot 2023-11-10 at 15 24 16" src="https://github.com/gradio-app/gradio/assets/16206207/7864c06f-6849-4cfb-8629-4b64f68b6773">

### System Info

```shell
Operating System: Darwin
gradio version: 4.1.2
gradio_client version: 0.7.0
```

### Severity

I can work around it

PR #6383, opened 2023-11-12
Fix event target

aliabid94:
With a gr.on() event listener, we were not setting the correct trigger target in a SelectEvent / EventData argument. See issue. Fixes: #6367, Fixes: #6407

Example Code:
```python
import gradio as gr

with gr.Blocks() as demo:
    text1 = gr.Textbox(label="Textbox 1")
    text2 = gr.Textbox(label="Textbox 2")
    text = gr.Textbox(label="Output")

    def test(evt: gr.SelectData):
        return 'You selected some text in ' + evt.target.label
 
        # Previously, output textbox displays 'You selected some text in Textbox 1' 
        # regardless of which textbox the selection is actually occuring in.

    gr.on(
        triggers=[text1.select, text2.select],
        fn=test,
        outputs=text
    )
    
demo.launch()
```


test/test_route_utils.py
```diff
@@ -1,19 +0,0 @@
-import gradio as gr
```
abidlabs, line 1: Fine to remove this as this was most testing implementation, but could we add some test to prevent this from regressing?

abidlabs: Some end to end tests around `.select()` would be good -- we don't really test anywhere and we've had regressions pop up before

abidlabs:
LGTM tested and works well @aliabid94!

aliabid94:
Edited to fix: #6407

abidlabs:
Awesome, can we add functional tests for both?

PR #6399, opened 2023-11-13
Improve CSS token documentation in Storybook

hannahblair:
## Description

This PR adds a few more stories with the css variables we use. It also removes hardcoded widths and heights for any icons and moves that sizing to their respective component. This improves the icons story a bit more.

## 🎯 PRs Should Target Issues

## Tests

3. You may need to run the linters: `bash scripts/format_backend.sh` and `bash scripts/format_frontend.sh`

### DIFF
--- a/.changeset/slow-buttons-cough.md
+++ b/.changeset/slow-buttons-cough.md
@@ -2,6 +2,7 @@
 "@gradio/chatbot": patch
 "@gradio/dropdown": patch
 "@gradio/icons": patch
+"@gradio/image": patch
 "@gradio/row": patch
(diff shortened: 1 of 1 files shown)


abidlabs:
@hannahblair can you explain the addition of the `ShadowRow` and `TokenRow` -- what do they achieve?

A few things I was unsure about:

1) In the Shadows section, there's a box with "Gradio" written in it. Clicking on it copies the token, but that's not super intuitive to me:

image

2) In the Row Layout, I think there should be two images side by side, but none show up:

image

Issue #6391, opened 2023-11-13
`api_name` is set in `gr.Examples` if not specified


hysts:
### Describe the bug

![](https://github.com/gradio-app/gradio/assets/25161192/e8958c85-89b0-495a-a28c-2db5203a1754)

The documentation of `gradio==4.2.0` says that its default is `False` and the code indeed seems so.
https://github.com/gradio-app/gradio/blob/9d6d72f44370c45d9e213421a5586c25c5789278/gradio/helpers.py#L106

But if one doesn't specify `api_name`, the endpoint is exposed with a name `/load_example`.

Weird thing is that the endpoint is not exposed if one explicitly set `api_name=False` though its default is `False`.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```py
import gradio as gr


with gr.Blocks() as demo:
    text = gr.Textbox()

    gr.Examples(examples=["a", "b"], inputs=text)
    gr.Examples(examples=["c", 'd'], inputs=text)


if __name__ == "__main__":
    demo.queue().launch()
```
When running this code, the log and the API documentation show like this:
```
/home/ubuntu/.virtualenvs/py31011/lib/python3.10/site-packages/gradio/blocks.py:890: UserWarning: api_name load_example already exists, using load_example_1
  warnings.warn(f"api_name {api_name} already exists, using {api_name_}")
Running on local URL:  http://127.0.0.1:7860

To create a public link, set `share=True` in `launch()`
```

![](https://github.com/gradio-app/gradio/assets/25161192/820e0f1b-ce4b-40fe-85db-cda793be0707)

On the other hand, if I run the following, it doesn't.
```py
import gradio as gr


with gr.Blocks() as demo:
    text = gr.Textbox()

    gr.Examples(examples=["a", "b"], inputs=text, api_name=False)
    gr.Examples(examples=["c", 'd'], inputs=text, api_name=False)


if __name__ == "__main__":
    demo.queue().launch()
```

### System Info

```shell
gradio==4.2.0
```

### Severity

I can work around it

abidlabs:
As an aside, we should document that if `api_name` is `False`, then downstream users cannot use that endpoint including users who may `gr.load()` that Space

PR #6405, opened 2023-11-13
Fix docstrings and default value for `api_name`

abidlabs:
The docstrings for `api_name` were out-of-date, and in the case of `gr.Examples`, an inappropriate default value was being used. 

Fixes: #6391


gradio/helpers.py
```diff
@@ -103,7 +103,7 @@ def __init__(
         run_on_click: bool = False,
         preprocess: bool = True,
         postprocess: bool = True,
-        api_name: str | None | Literal[False] = False,
+        api_name: str | Literal[False] = "load_example",
```
abidlabs, line 106: There's no reason to support `api_name` here since `None` will default to "load_example" in all cases

freddyaboulton (approved):
Thanks @abidlabs !!

abidlabs:
Thanks @freddyaboulton for the review!

Issue #6394, opened 2023-11-13
Timeout


eta1232002:
### Describe the bug

I used the queue() , but I still get timeout after 70 seconds .. With enabled debugging, the output appears in the colab but does not appear in gradio output.. pls resolve this issue urgently

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

gr.Interface(
    title = 'Speech Recognition Gradio Web UI',
    fn=speech2txt,
    inputs="text", outputs="text",
     live=True).queue().launch(share=True,debug=True)

### System Info

```shell
yes

! pip install gradio -q
```

### Severity

Blocking usage of gradio

abidlabs:
What version of gradio are you using? Please upgrade to the latest version of Gradio and also share the audio that you are trying as well.

eta1232002:
I am using the last version of gradio.

This is the video I try to transcribe and it takes around 90 secs
https://www.youtube.com/watch?v=LrkFYGRVJYE

eta1232002:
I am waiting your reply urgently

gibiee:
I tried to reproduce this problem in Colab.
```
!pip install gradio -q

# Python 3.10.12
# gradio==4.3.0
# gradio_client==0.7.0
```

I ran the code below in Colab.
```py
import gradio as gr

def wait(input_text) :
    print(f"INPUT : {input_text}")

    total = 0
    for i in range(1000000000) :
        total  += i
    
    output = f"{total} with {input_text}"
    print(f"OUTPUT : {output}")
    return output

gr.Interface(title='TEST', fn=wait, inputs="text", outputs="text", live=True).queue().launch(share=True, debug=True)
```
(Why use `live=True`? Even if I enter "abc", only "a" is entered.)

If I run `def wait()` function alone, it takes about 145~150 seconds.
But, when I run the code, there is a bug that the same input is queued repeatedly. The process in demo never ends.
Here's a screenshot.
![repeat](https://github.com/gradio-app/gradio/assets/37574274/41595b0c-fcc6-4624-ab92-3ada753b85cf)

gibiee:
@eta1232002 
For now, you can solve the problem by `!pip install -q gradio==3.50.2`

![solved](https://github.com/gradio-app/gradio/assets/37574274/82a6cca1-f936-404a-9259-cb7ea8b5e349)

eta1232002:
@gibiee Thanks for your effort, it works on gradio==3.50.2

PR #6418, opened 2023-11-14
Send more than one heartbeat message

freddyaboulton:
## Description

Only one heartbeat message was being sent so the EventStream thought the request failed so it retried. I think there's like a 30-60 second timeout where it will retry if no new messages are sent. So the queue was growing infinitely.

Hard to write a unit test for this that doesn't test implementation but you can verify the fix with the following demo. The event should never finish on main but it will in this pr. I think you need to access it via the share link to repro.

```python
import gradio as gr
import time

def wait(input_text) :
    print(f"INPUT : {input_text}")

    total = 0
    time.sleep(150)
    
    output = f"{total} with {input_text}"
    print(f"OUTPUT : {output}")
    return output

gr.Interface(title='TEST', fn=wait, inputs="text", outputs="text", live=True).queue().launch(share=True, debug=True)
```
image

Closes: #6394

## 🎯 PRs Should Target Issues

## Tests


abidlabs (approved):
Amazing, works great! LGTM @freddyaboulton

freddyaboulton:
Whoops - sorry I missed this @abidlabs ! Thanks for the review. Will merge.

PR #6424, opened 2023-11-15
Do not show warnings when renaming api_names

abidlabs:
In Gradio 4.0, all API endpoints are named by default, which means that warnings about duplicated api names pop up too easily and pollute the stdout. This PR removes those warnings, which are not too necessary.


hysts (approved):
Thanks! LGTM!

abidlabs:
Thanks @hysts

Issue #6408, opened 2023-11-13
Migration issue 3.x.x to 4: Resetting gr.Dropdown doesn't work


sbarman-mi9:
### Describe the bug

In 3.42.0, if we set `value=None` or `value=''`, dropdown component would reset/clear and we can select it's value(s) again. In 4, it retains the selected value even after clearing the component.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```
import gradio as gr

def do_something():
    return gr.update(value=None) # doesn't work with value=''

with gr.Blocks() as demo:
    drop_dwn = gr.Dropdown(["a", "b", "c", "d", "e"], value='b', label="Select", interactive=True)
    reset_btn = gr.Button("Reset")

    reset_btn.click(do_something, None, [drop_dwn])

demo.launch(server_name="0.0.0.0", server_port=7860)
```

### Screenshot

https://github.com/gradio-app/gradio/assets/14152255/3b9c91db-5835-4ea0-b811-38b6c6546ce8

### System Info

```shell
------------------------------
Operating System: Linux
gradio version: 4.2.0
gradio_client version: 0.7.0

------------------------------------------------

python==3.10
```

### Severity

Blocking usage of gradio

PR #6425, opened 2023-11-15
Update the selected indices in `Dropdown` when value changes programmatically

abidlabs:
Due to a regression, we weren't updating the "selected indices" (which dropdown options had a tick next to them) when the value was programmatically changed. 

This PR fixes that and adds a unit test.

Closes: #6408

### DIFF
--- a/.changeset/whole-eels-poke.md
+++ b/.changeset/whole-eels-poke.md
@@ -1,0 +1,6 @@
+---
+"@gradio/dropdown": patch
+"gradio": patch
+---
+
+fix:Update the selected indices in `Dropdown` when value changes programmatically
--- a/js/dropdown/dropdown.test.ts
+++ b/js/dropdown/dropdown.test.ts
@@ -327,7 +327,7 @@ describe("Dropdown", () => {
 		expect(item.value).toBe("applepie");
 	});
 
-	test("setting a value should update the displayed value and selected indicies", async () => {
+	test("setting a value should update the displayed value and selected indices", async () => {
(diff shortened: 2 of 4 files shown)


abidlabs:
Thanks @hannahblair for reviewing! Will fix the typo and merge in

sbarman-mi9:
There's still an issue. This is using 4.4.0. Previously selected value isn't re-selected after reset. 

<u>**Steps**</u>:
1. select option 'd'
2. reset
3. try selecting the _previously_ selected value  'd'

It doesn't render on the component while showing a selected check mark when you open the list of options.
Just so we're not hallucinating, we can also test the values outside of the wrapper (JS console). 

Please let us know if we are missing something that isn't available in the docs.

```
import gradio as gr

def do_something():
    return gr.update(value=None)

with gr.Blocks() as demo:
    drop_dwn = gr.Dropdown(["a", "b", "c", "d", "e"], 
                           value='b', 
                           label="Select", 
                           elem_id="drp_dwn",
                           interactive=True
                           )
    reset_btn = gr.Button("Reset")

    reset_btn.click(do_something, None, [drop_dwn])

demo.queue(default_concurrency_limit=13).launch(server_name="0.0.0.0", server_port=7860)
```


https://github.com/gradio-app/gradio/assets/14152255/159fb969-2de4-47e0-92fe-84f31228482a

abidlabs:
Ah sorry you're right, thanks for catching @sbarman-mi9. I'll get that fixed

PR #6427, opened 2023-11-15
Allow google analytics to work on Spaces (and other iframe situations)

abidlabs:
Following https://stackoverflow.com/questions/74494818/google-analytics-not-working-for-web-site-embeded-in-iframe (thanks for sharing @aliabd), modified our google analytics tag to work on Spaces. 

Not sure if there's an easy way to test that this is working, but will take a look at our GA to see if anything shows up due to the deployed Space PR.

### DIFF
--- a/.changeset/slimy-oranges-dress.md
+++ b/.changeset/slimy-oranges-dress.md
@@ -1,0 +1,7 @@
+---
+"@gradio/app": minor
+"gradio": minor
+"website": minor
+---
+
+feat:Allow google analytics to work on Spaces (and other iframe situations)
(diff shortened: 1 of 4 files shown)


aliabd:
Yes we should be able to use the deployed space to check

aliabd:
actually @abidlabs i think something more complicated is happening. I think spaces is removing the gtag code inside the iframe... it doesn't show up even on the iframed domain: https://gradio-pr-deploys-pr-6427-all-demos.hf.space/ try looking for it in inspect element 

whereas shows up fine locally, on colab etc

aliabd:
nvm like you pointed out it's cause we disable analytics on pr spaces. ignore what i said

aliabd (approved):

abidlabs:
Thanks @aliabd for debugging this with me!

PR #6428, opened 2023-11-15
Extract video filenames correctly from URLs

112292454:
## Description

before cannot process url like "https:/example.com/avatar/xxxx.mp4?se=2023-11-16T06:51:23Z&sp=r" correctly, will get format as "mp4?se=2023-11-16T06:51:23Z&sp=r", and go to unnecessary format convert


abidlabs:
Thanks @112292454! This looks good. I wonder if there's a way to get the base filepath in a URL without having to do manual parsing of the string?

Would you be able to add a test to `test_components.py`?

112292454:
> Thanks @112292454! This looks good. I wonder if there's a way to get the base filepath in a URL without having to do manual parsing of the string?
> 
> Would you be able to add a test to `test_components.py`?

now use urllib and os.path
I don’t often write py ,  may be some specification issues
The urls may be can used for other tests

abidlabs:
Thanks @112292454 I refactored this out to be a separate util function and added a test. Merging!

112292454:
@abidlabs 😭 i'm sorry, i has did some change when last comment, but i commit it incorrectly(merge from my another branch. but didn't put that code into this pr)
i check it an commit again?

abidlabs:
Its okay, I think my changes should have covered everything, right? Unless I missed something?

112292454:
you're right, just different in test case, and if it end with /
maybe no need to change it
[that commit](https://github.com/112292454/gradio/commit/8a23f7775ac61d76fab2c3072f7ba8c5a52ef25e)

Issue #6354, opened 2023-11-09
Custom Components CLI improvements


duerrsimon:
- [x] I have searched to see if a similar issue already exists.

**Following features are still missing**  

- [x] `gradio cc publish` deployment to pypi doesn't support accounts with 2FA
- [ ] `gradio cc create` should not actually create the folder if the template isn't found
- [ ]  `pyproject.toml` could also set a few more inputs based on CLI. In oarticular the `[project]` part

```
[project]
--
  | name = "gradio_molecule3d"
  | version = "0.0.1"  # set each time cc publish is called
  | description = "Molecule3D custom component to visualize pdb or sdf files" # gradio cc create
  | readme = "README.md"
  | license = "MIT" # gradio cc create
  | requires-python = ">=3.8" # gradio cc create
  | authors = [{ name = "Simon Dürr", email = "dev@simonduerr.eu" }] # gradio cc create
  | keywords = [
  | "machine learning",
  | "reproducibility",
  | "visualization",
  | "gradio",
  | "gradio custom component",
  | "gradio-template-File",
  | "protein"
  | ] # gradio cc create
```

freddyaboulton:
Thanks @duerrsimon ! I like the idea of optionally setting more config via the cli.

BTW `publish` already supports accounts with 2fa. Use the username `__token__`  and the password is your API key. Will make this clearer.

Issue #6349, opened 2023-11-08
Improve the custom component deploy dockerfile


freddyaboulton:
- [ ] I have searched to see if a similar issue already exists.

We should add the following lines to the `Dockerfile` used in custom component demos so that developers can download transformers models without getting an error that the cache is not set.

```
RUN mkdir -p /tmp/cache/
RUN chmod a+rwx -R /tmp/cache/
ENV TRANSFORMERS_CACHE=/tmp/cache/ 
```

The reason we're using a Dockerfile is so that users can install their component in a space without deploying to pypi but maybe we can use the gradio sdk space if they did deploy to pypi.

**Is your feature request related to a problem? Please describe.**  

**Describe the solution you'd like**  

**Additional context**

PR #6436, opened 2023-11-15
Custom Component CLI Improvements

freddyaboulton:
## Description

Closes: #6354 
Closes: #6349 

Summary of changes:
* The `gradio cc create` command now asks users to specify some project metadata as suggested in #6354. There's a cli flag to disable this and answering the questions is optional. Adding the keywords will help with custom component discoverability!
* The `gradio cc build` clearly mentions which version will be used for the wheel file. Added a `--bump-version` flag to automatically do a patch bump. Defaults to false because people can build multiple times.
* The `gradio cc publish` now explains how to publish if your pypi account has 2fa. Also we can read credentials automatically from a `~/.pypirc` file (the community standard). 
* Improve the default Docker image as described in #6349
* Automatically adds a title and description to the Dockerfile of the published space as well.
* Rename the five minute guide for SEO based on @aliabid94 's comment

## 🎯 PRs Should Target Issues

## Tests


freddyaboulton:
I think there's an issue with the redirect. @aliabd Can you help me? Otherwise, I think this is good to review.

js/_website/src/routes/redirects.js
```diff
@@ -1,5 +1,7 @@
 export const redirects = {
-	"/guides/creating-a-new-component": "/guides/five-minute-guide",
+	"/guides/creating-a-new-component":
```
abidlabs, line 2: Nice

abidlabs:
Generally looks good -- I haven't gone through the whole flow just yet -- will do that later today to review fully

abidlabs:
Tested, love the new workflow.

One tiny nit is that when you do `gradio cc publish` and you have multiple versions' wheels, then all of them get uploaded to pypi, which is unnecessary (and could slow things down if there are lots and lots of versions). Only the latest version needs to be uploaded:

image


Otherwise LGTM @freddyaboulton everything looks stellar!!

abidlabs (approved):

js/_website/src/routes/redirects.js
```diff
@@ -1,5 +1,7 @@
...
+		"/guides/custom-components-in-five-minutes",
+	"/guides/five-minute-guide": "/guides/custom-components-in-five-minutes",
```
aliabd, line 4: this (and the redirect before it) are breaking the website currently because they try to redirect to "/guides/custom-components-in-five-minutes" which doesn't exist yet (it's still /main/guides/custom-components-in-five-minutes until the guide is included in a released version) I would just remove them until we release or redirect to main.

freddyaboulton: Thanks @aliabd !

freddyaboulton:
Thanks for the review @abidlabs ! Doing a semver sort to only upload the latest release but pypi actually uses a different pre-release format so in that case we'll upload all files. Should be minor inconvenience.

PR #6437, opened 2023-11-15
chore: rename api_key to hf_token

NickCrews:
The param name changed in the past but these comments and error messages are stale.

Closes: NA


abidlabs:
LGTM, thanks @NickCrews for the contribution! I also updated a few of our demo notebooks and added here.

abidlabs (approved):

NickCrews:
Good catch, thank you!

Issue #6435, opened 2023-11-15
Be able to set a "default" `concurrency_limit` for all events to something other than 1


abidlabs:
Based on issues brought up by several internal discussions ([here](https://huggingface.slack.com/archives/C04ABS7C13K/p1699981121268479), [here](https://huggingface.slack.com/archives/C02QZLG8GMN/p1700037321624929),  [here](https://huggingface.slack.com/archives/C02QZLG8GMN/p1698867714281819?thread_ts=1698857187.185849&cid=C02QZLG8GMN)), I think we should let users set a `default_concurrency_limit`  in `.launch()` that sets the concurrency_limit for all events to a higher number than 1 or to `None`.

To support the first use case, we should also allow this to be set with an environment variable, `GRADIO_DEFAULT_CONCURRENCY_LIMIT`

cc @cbensimon @apolinario @clefourrier

PR #6439, opened 2023-11-15
Allow setting a `default_concurrency_limit` other than 1

abidlabs:
Adds a `default_concurrency_limit` to `launch()` that sets the `concurrency_limit` of all events in the `Blocks` that don't have an explicit `concurrency_limit`. 

Instead of `default_concurrency_limit`, you can set the "GRADIO_DEFAULT_CONCURRENCY_LIMIT" env variable. 

Closes: #6435 

cc @cbensimon @apolinario @clefourrier


gradio/blocks.py
```diff
@@ -1731,6 +1732,7 @@ def launch(
         state_session_capacity: int = 10000,
         share_server_address: str | None = None,
         share_server_protocol: Literal["http", "https"] | None = None,
+        default_concurrency_limit: int | None = None,
```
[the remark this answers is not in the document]
abidlabs: No because it also looks for the environment variable

gradio/blocks.py
```diff
@@ -1827,6 +1830,12 @@ def reverse(text):
         self.config = self.get_config_file()
         self.max_threads = max_threads
         self._queue.max_thread_count = max_threads
+        self._queue.default_concurrency_limit = int(
```
aliabid94, line 1833: can we call self._queue.set_concurrency_limit from here instead? Feels like this logic should be handled by the queue

abidlabs: Oh whoops that's what I meant to do, will fix

freddyaboulton (commented):
Looks good to me @abidlabs ! 

But I think it makes more sense to have this parameter in queue as opposed to launch?

abidlabs:
Have made the change -- will go ahead and merge this in. Thanks all!

Issue #6433, opened 2023-11-15
upload() isnt working on gradio 4.3.0


zenafey:
### Describe the bug

Found that .upload() method of gr.Image() isnt working in my app, i thout something changed since 3.x gradio but docs saying everything should be fine, i created simple example that shows this issue

### Have you searched existing issues?  🔎

- [x] I have searched and found no existing issues

### Reproduction

1. Install gradio 4.3.0
2. Create simple app that uses gr.Image() and .upload()
```py
import gradio as gr

with gr.Blocks() as demo:
    image_input = gr.Image(sources=['clipboard'])

    def foo(image):
        print("TEST STARTED =>")
        print(image)
        print("TEST ENDED")

    image_input.upload(foo, inputs=[image_input])

demo.launch()
```
3. see that nothing happen when you upload image

### Screenshot

https://github.com/gradio-app/gradio/assets/118455214/c650be14-7c4c-491a-9feb-d4317bf7b156

### Logs

```shell
Running on local URL:  http://127.0.0.1:7861

To create a public link, set `share=True` in `launch()`.
```

### System Info

```shell
OS: Windows 11
Gradio: 4.3.0
Python: 3.10.6
```

### Severity

Blocking usage of gradio

PR #6441, opened 2023-11-15
Image Fixes

freddyaboulton:
## Description

Closes: #6433

* The upload event was not triggering at all
* The paste-from-clipboard was not triggering an upload event
* The clear button was not triggering a change event
* The change event was triggering infinitely
* Uploaded images were not preserving their original names

## 🎯 PRs Should Target Issues

## Tests


.config/playwright.config.js
```diff
@@ -1,7 +1,12 @@
 export default {
 	use: {
 		screenshot: "only-on-failure",
-		trace: "retain-on-failure"
+		trace: "retain-on-failure",
+		permissions: ["clipboard-read", "clipboard-write"],
+		bypassCSP: true,
+		launchOptions: {
```
freddyaboulton, line 7: Needed otherwise you get CORS errors

freddyaboulton:
I have to afk but I figured I would open for review since the functional tests passed. Will fix the backend lint when I'm back.

abidlabs:
Awesome PR @freddyaboulton! 

The only issue I noticed is that pasting an image from the clipboard triggers the `.upload()` event twice. Besides that, LGTM!

Test code:

```py
import gradio as gr

with gr.Blocks() as demo:
    image_input = gr.Image()
    textbox = gr.Textbox()

    def upload(image):
        print("upload")
        return "upload"

    def change(image):
        print("change")
        return "change"
        
    image_input.upload(upload, inputs=[image_input], outputs=textbox)
    image_input.change(change, inputs=[image_input], outputs=textbox)

demo.launch()
```

Btw, separately from this PR, I noticed that if you have an image already uploaded, and you upload another image, the progress bar appears *above* the old image instead of clearing the old image first.

js/app/test/image_component_events.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+			await fetch(
+				`https://gradio-builds.s3.amazonaws.com/assets/PDFDisplay.png`
+			)
+		).blob();
+		navigator.clipboard.write([new ClipboardItem({ [blob.type]: blob })]);
+	});
+
+	// Not sure if there's a better way to click the clipboard button
+	await page.pause();
+	await page.getByRole("button").nth(2).click();
```
pngwn, line 58: Give the button a data test id.

freddyaboulton: Hard to give the button a test id without adding an extra prop to IconButton. Using the label instead and I think it's working! Thanks

freddyaboulton:
Thanks for the review @abidlabs ! The two upload event issue is fixed! Also fixed the issue with the image not being cleared out if you upload another one.

abidlabs (approved):
Good stuff!

freddyaboulton:

Issue #6434, opened 2023-11-15
Audio Event Handlers not working on Output Audio Component


srednors:
### Describe the bug

The Play, Pause, Stop and Change Event Handlers work fine for Input Audio Components but don't seem to fire for Output Audio Components.

Related to #2518. Although in that report the reproduction code only contained an input Audio Component which is maybe why the problem didn't surface when tested.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```
import gradio as gr

def onPause():
    return "Audio Paused"

def onPlay():
    return "Audio Played"

def onStop():
    return "Audio Stopped / Ended"

def onChange():
    return "Audio Changed"

with gr.Blocks() as demo:

    outputAudio = gr.Audio(
        "https://samplelib.com/lib/preview/mp3/sample-3s.mp3", label="Output Audio"
    )
    textout = gr.Textbox(label="Output Audio")

    outputAudio.stop(onStop, outputs=textout, api_name=False)
    outputAudio.pause(onPause, outputs=textout, api_name=False)
    outputAudio.play(onPlay, outputs=textout, api_name=False)
    outputAudio.change(onChange, outputs=textout, api_name=False)

    inputAudio = gr.Audio(sources="microphone", label="Input Audio")
    textin = gr.Textbox(label="Input Audio")

    inputAudio.stop(onStop, outputs=textin, api_name=False)
    inputAudio.pause(onPause, outputs=textin, api_name=False)
    inputAudio.play(onPlay, outputs=textin, api_name=False)
    inputAudio.change(onChange, outputs=textin, api_name=False)

demo.launch()
```

### Screenshot

![audio](https://github.com/gradio-app/gradio/assets/16206207/fac2fddd-cf11-40e5-b441-79ba5f8f528b)

### System Info

```shell
Operating System: Darwin
gradio version: 4.3.0
gradio_client version: 0.7.0
FFmpeg: 6.0
```

### Severity

Blocking usage of gradio

PR #6454, opened 2023-11-16
Ensure Audio ouput events are dispatched

hannahblair:
## Description

We weren't dispatching our events properly for the static audio component. This PR fixes that, and adds output event e2e tests. 

Closes: #6434 

## 🎯 PRs Should Target Issues

## Tests


js/audio/player/AudioPlayer.svelte
```diff
@@ -78,7 +86,6 @@
 	$: waveform?.on("finish", () => {
 		playing = false;
 		dispatch("stop");
-		dispatch("end");
```
freddyaboulton, line 81: Is "end" just for the end of the recording session?

freddyaboulton: I don't think we're ever triggering the end event anymore. Is that intended?

freddyaboulton (commented):
Thanks @hannahblair ! Looks good to me. Just have a question about the end event. Nit but can we also test the input stop event like we do for output?

freddyaboulton (approved):

PR #6456, opened 2023-11-16
Preserve original image extension in backend processing

freddyaboulton:
## Description

Jpegs were being converted to png always.

e2e test reflects this change.

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/gradio/image_utils.py
+++ b/gradio/image_utils.py
@@ -16,6 +16,7 @@ def format_image(
     type: Literal["numpy", "pil", "filepath"],
     cache_dir: str,
     name: str = "image",
+    format: str = "png",
 ) -> np.ndarray | _Image.Image | str | None:
     """Helper method to format an image based on self.type"""
     if im is None:
@@ -27,7 +28,7 @@ def format_image(
         return np.array(im)
     elif type == "filepath":
         path = processing_utils.save_pil_to_cache(
-            im, cache_dir=cache_dir, name=name, format=fmt or "png"  # type: ignore
+            im, cache_dir=cache_dir, name=name, format=fmt or format  # type: ignore
         )
         return path
     else:
@@ -39,10 +40,12 @@ def format_image(
 
 
 def save_image(y: np.ndarray | _Image.Image | str | Path, cache_dir: str):
+    # numpy and PIL get saved to png as default format
     if isinstance(y, np.ndarray):
         path = processing_utils.save_img_array_to_cache(y, cache_dir=cache_dir)
     elif isinstance(y, _Image.Image):
         path = processing_utils.save_pil_to_cache(y, cache_dir=cache_dir)
+    # We preserve the format if the developer passes a filepath
     elif isinstance(y, Path):
         path = str(y)
     elif isinstance(y, str):
--- a/gradio/image_utils.py
+++ b/gradio/image_utils.py
@@ -27,9 +27,15 @@ def format_image(
     elif type == "numpy":
         return np.array(im)
     elif type == "filepath":
-        path = processing_utils.save_pil_to_cache(
-            im, cache_dir=cache_dir, name=name, format=fmt or format  # type: ignore
-        )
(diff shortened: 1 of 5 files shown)


gradio/image_utils.py
```diff
@@ -27,7 +28,7 @@ def format_image(
         return np.array(im)
     elif type == "filepath":
         path = processing_utils.save_pil_to_cache(
-            im, cache_dir=cache_dir, name=name, format=fmt or "png"  # type: ignore
+            im, cache_dir=cache_dir, name=name, format=fmt or format  # type: ignore
```
abidlabs, line 31: The only risk (admittedly small) I can see here is that there are some image formats that PIL can read but not save as: https://pillow.readthedocs.io/en/stable/handbook/image-file-formats.html#read-only-formats

They all look very rare to me, but perhaps a more robust approach would be to try / catch this, and save as "png" as a fallback

js/app/test/image_component_events.spec.ts
```diff
@@ -21,7 +21,7 @@ test("Image click-to-upload uploads image successfuly. Clear button dispatches e
 	await page.getByLabel("Download").click();
 	const download = await downloadPromise;
 	// Automatically convert to png in the backend since PIL is very picky
-	await expect(download.suggestedFilename()).toBe("cheetah1.png");
+	await expect(download.suggestedFilename()).toBe("cheetah1.jpeg");
```
abidlabs, line 24: nice

abidlabs:
Awesome thanks for taking care of this @freddyaboulton. Left a small note, otherwise lgtm

gradio/image_utils.py
freddyaboulton: Yea that's a good point

freddyaboulton:
Just pushed up suggested changes @pngwn and @abidlabs !

gradio/image_utils.py
```diff
@@ -39,10 +46,21 @@ def format_image(
...
         path = processing_utils.save_img_array_to_cache(y, cache_dir=cache_dir)
     elif isinstance(y, _Image.Image):
-        path = processing_utils.save_pil_to_cache(y, cache_dir=cache_dir)
+        fmt = y.format
+        try:
+            path = processing_utils.save_pil_to_cache(
+                y, cache_dir=cache_dir, format=fmt  # type: ignore
+            )
+        # Catch error if format is not supported by PIL
+        except Exception:
```
abidlabs, line 60: looks like its a KeyError that PIL raises for invalid formats
```suggestion
        except KeyError:
```

gradio/image_utils.py
```diff
@@ -26,9 +27,15 @@ def format_image(
...
     elif type == "filepath":
-        path = processing_utils.save_pil_to_cache(
-            im, cache_dir=cache_dir, name=name, format=fmt or "png"  # type: ignore
-        )
+        try:
+            path = processing_utils.save_pil_to_cache(
+                im, cache_dir=cache_dir, name=name, format=fmt or format  # type: ignore
+            )
+        # Catch error if format is not supported by PIL
+        except Exception:
```
abidlabs, line 35: looks like its a KeyError that PIL raises for invalid formats
```suggestion
        except KeyError:
```

abidlabs:
Nice thanks @freddyaboulton!

abidlabs (approved):

gradio/image_utils.py
freddyaboulton: Is that the only one?

abidlabs: For invalid formats, yes. Tested with:

```py
from PIL import Image

Image.open("cheetah.jpg").save("c2.jpg", format="cur")  # cur is a read only format
```

freddyaboulton: Let's see what happens!

freddyaboulton:

PR #6478, opened 2023-11-18
Add back templates based on `gr.ImageEditor`

abidlabs:
Adds back template classes:
* ImageMask
* Sketchpad
* Paint

Deletes:
* Pil (kinda pointless)
* ImagePaint (the default ImageEditor)

### DIFF
--- a/gradio/components/image_editor.py
+++ b/gradio/components/image_editor.py
@@ -116,8 +116,8 @@ class ImageEditor(Component):
         data_mode: Literal["image", "pathline"] = "image",
         crop_size: tuple[int | float, int | float] | str | None = None,
         transforms: Iterable[Literal["crop", "rotate"]] = ("crop", "rotate"),
-        eraser: Eraser | None = Eraser(),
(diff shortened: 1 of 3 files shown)


gradio/components/image_editor.py
```diff
@@ -116,8 +116,8 @@ def __init__(
         data_mode: Literal["image", "pathline"] = "image",
         crop_size: tuple[int | float, int | float] | str | None = None,
         transforms: Iterable[Literal["crop", "rotate"]] = ("crop", "rotate"),
-        eraser: Eraser | None = Eraser(),
-        brush: Brush | None = Brush(),
```
abidlabs, line 120: This is not recommended as the objects will only be instantiated once per Python session, even if you have multiple classes

pngwn: Maybe we can use `False` for that case.

pngwn (approved):
I forgot about these! Thanks for taking care of this @abidlabs!

PR #6502, opened 2023-11-20
Ensure image editor crop and draw cursor works as expected when the scroll position changes

pngwn:
## Description
 _I am not here. I did not make this PR_

There was a bug (pretty bad one imo), where the crop and draw/erase cursor would get funky if the user scrolled the page after uploading an image. This is because they rely on DOM measurements but those measurements are not updated on scroll.  This PR fixes that.

Also corrected the typos in the docstrings.

Closes: #(issue)

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/whole-buckets-add.md
+++ b/.changeset/whole-buckets-add.md
@@ -1,0 +1,6 @@
+---
+"@gradio/imageeditor": patch
+"gradio": patch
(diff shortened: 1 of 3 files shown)


pngwn:
cc @hysts

PR #6512, opened 2023-11-21
Update zh-CN.json

cibimo:
Update zh-CN.json


xianbaoqian:
Looks great! Thanks a lot @cibimo !

js/app/src/lang/zh-CN.json
```diff
@@ -1,16 +1,115 @@
...
+        "select_brush_color": "选择画笔颜色",
+        "start_drawing": "开始绘画",
+        "use_brush": "使用画笔"
+    },
+    "label": {
+        "label": "标签"
+    },
+    "login": {
+        "enable_cookies": "如果您正在使用隐身模式访问 HuggingFace 空间，您必须启用第三方 cookie。",
+        "incorrect_credentials": "凭据不正确",
```
xianbaoqian, line 92: 凭据 sounds a bit weird.  Do you consider 密码不正确？ or  验证失败？

xianbaoqian (approved):

abidlabs:
Hi @cibimo would you be willing to adjust the translation as suggested by @xianbaoqian in your PR?

abidlabs:
Will go ahead and merge as is and we can patch later as needed.

@hannahblair do you think we could update all of the existing i18n files to make it easier for contributors to know what the missing values are? E.g. in all of the existing i18n files, could we do something like: `"allow_recording_access": null` for all of the fields with a missing translation --> it would make it very easy for a contributor to come in an add translations

PR #6517, opened 2023-11-21
Allow reselecting the original option in `gr.Dropdown` after value has changed programmatically

abidlabs:
Fixes the issue identified by [sbarman-mi9](https://github.com/sbarman-mi9) in https://github.com/gradio-app/gradio/pull/6425#issuecomment-1817426966

Test with:

```py
import gradio as gr

def do_something():
    return gr.update(value=None)

with gr.Blocks() as demo:
    drop_dwn = gr.Dropdown(["a", "b", "c", "d", "e"], 
                           value='b', 
                           label="Select", 
                           elem_id="drp_dwn",
                           interactive=True
                           )
    reset_btn = gr.Button("Reset")

    reset_btn.click(do_something, None, [drop_dwn])

demo.queue(default_concurrency_limit=13).launch()
```

### DIFF
--- a/.changeset/nine-bats-sin.md
+++ b/.changeset/nine-bats-sin.md
@@ -1,0 +1,6 @@
+---
+"@gradio/dropdown": minor
+"gradio": minor
+---
+
+feat:Allow reselecting the original option in `gr.Dropdown` after value has changed programmatically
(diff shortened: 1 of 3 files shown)


abidlabs:
Thanks @hannahblair!

Issue #6493, opened 2023-11-19
Suggestion : Add option to open by default the accordion with additional inputs in ChatInterface


Jeronymous:
Would it be possible to add an option to `ChatInterface` so that the window with additional inputs is opened by default.

Something like
```
ChatInterface(..., additional_inputs_accordion_open=False, ...) # by default
```

where `additional_inputs_accordion_open` would be the new option used in this line (instead of `False`):
https://github.com/gradio-app/gradio/blob/2780d067f9f801016c0254de679b56794859abed/gradio/chat_interface.py#L235

I am interested with that because I am building a Chat that proposes several non usual input fields, and I'd like them not be hidden to the user by default.

And thanks for all the great work in gradio :+1:

Jeronymous:
I also wonder if it is possible to have an accordion for the "Examples" section, that can be closed by default
(so that it can contain many Examples, among which non typical ones that I would like to be seen only by user who knows what they are doing)

PR #6518, opened 2023-11-21
Allows setting parameters of `gr.ChatInterface`'s `Accordion`

abidlabs:
Deprecates the overly restrictive `additional_inputs_accordion_name` argument and instead replaces it with a `additional_inputs_accordion` argument which can take either a string or a whole `gr.Accordion` object. 

This is more inline with the other arguments in `gr.ChatInterface`, such as `submit_btn`

Test with:

```py
import gradio as gr

def echo_system_prompt_plus_message(message, history, system_prompt, tokens):
    response = f"{system_prompt} {message}"
    for i in range(min(len(response), int(tokens))):
        yield response[: i + 1]

chatbot = gr.ChatInterface(
    echo_system_prompt_plus_message,
    additional_inputs=["textbox", "slider"],
    additional_inputs_accordion=gr.Accordion(open=True, label="MOAR"),
)
```

Closes: https://github.com/gradio-app/gradio/issues/6493

### DIFF
--- a/.changeset/warm-otters-rest.md
+++ b/.changeset/warm-otters-rest.md
@@ -1,0 +1,5 @@
+---
+"gradio": minor
+---
+
+feat:Allows setting parameters of `gr.ChatInterface`'s `Accordion`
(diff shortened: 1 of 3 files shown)


hysts (approved):
LGTM!

freddyaboulton (approved):
Thanks @abidlabs !

abidlabs:
Thanks guys

PR #6524, opened 2023-11-21
Fix gradio_client duplicating upgraded Spaces

osanseviero:
## Description

[internal convo](https://huggingface.slack.com/archives/C02QZLG8GMN/p1700475888741849). When passing the token explicitly, e.g.

```python
client = Client.duplicate("abidlabs/whisper", hf_token="token")
```
the repo is duplicated with a CPU but then the client fails with a 

```
Repository Not Found for url: https://huggingface.co/api/spaces/osanseviero/whisper/hardware.
Please make sure you specified the correct `repo_id` and `repo_type`.
If you are trying to access a private or gated repo, make sure you are authenticated.
Invalid username or password.
``` 

error. This is due to the token not being passed when requesting a hardware.

### DIFF
--- a/client/python/gradio_client/client.py
+++ b/client/python/gradio_client/client.py
@@ -253,7 +253,7 @@ class Client:
         )
         hardware = hardware or original_info.hardware
         if current_hardware != hardware:
-            huggingface_hub.request_space_hardware(space_id, hardware)  # type: ignore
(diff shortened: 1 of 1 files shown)


freddyaboulton (approved):
Thanks @osanseviero !

PR #6525, opened 2023-11-21
Fixes Drag and Drop for Upload

dawoodkhan82:
## Description
Fixes the issue around the mime-type checks in the Upload component, which was causing errors for drag and dropping files.

Closes: #6490

## 🎯 PRs Should Target Issues

## Tests


freddyaboulton (commented):
Can you modify the `model3d_component_events.spec.ts` and maybe write some 2e2 tests for `gr.File`? We have some for UploadButton but not File. There's an example of DnD in `video_component_events.spec.ts`.

abidlabs:
Looks like this might have broken a couple of functional tests? I'm rerunning them now

dawoodkhan82:
Hmm, so the failing functional tests, are actually passing for me locally. Also by re-running these tests i notice some of them are pretty flaky.

dawoodkhan82:
Tests are passing now! Ready for re-review @abidlabs @freddyaboulton

abidlabs:
Nice @dawoodkhan82! The issue is fixed for me for the `gr.File()` component, but not for the `gr.Model3D` component. Here's the code I'm trying, and I'm attaching the file (its a zipped .obj file) as well. Are you able to repro?

```py
import gradio as gr

def noop(*args):
    return None

demo = gr.Interface(
    fn=noop,
    inputs=[gr.File(), gr.Model3D(), gr.Image()],
    outputs=None,
)

demo.launch()
```

[Bunny.obj.zip](https://github.com/gradio-app/gradio/files/13665067/Bunny.obj.zip)

abidlabs:
@dawoodkhan82 and I looked into it, it seems like it only affects `.obj` files

dawoodkhan82:
@abidlabs Fixed, without (hopefully) breaking anything else!

abidlabs:
Awesome @dawoodkhan82 works for me! Thanks for finding the solution.

abidlabs (approved):

abidlabs:
I'll go ahead and merge this in so that we can do a patch release with #6693 and this

PR #6529, opened 2023-11-21
Fix broken link in gradio_client README

zainhaq-h2o:
## Description

This PR fixes a broken link in the gradio_client README.


abidlabs:
Thanks @zainhaq-h2o for the fix!

PR #6534, opened 2023-11-21
Add anyio to requirements

fakerybakery:
It's used by Gradio.


requirements.txt
```diff
@@ -1,3 +1,4 @@
+anyio
```
freddyaboulton, line 1: I don't think we need to do this. anyio is a dependency of starlette, which is installed by fastapi.

fakerybakery: OK, thanks.

Issue #6531, opened 2023-11-21
Flaky issue when using `gr.themes.builder()`


abidlabs:
I'm unable to run the theme builder locally. I get this error in the console:

> Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'forEach')
    at handle_update (Blocks.svelte:316:8)
    at Blocks.svelte:451:7

This error happens consistently for me, but not for @freddyaboulton so it could be related to some race condition.

**Repro**

1. Run `gr.themes.builder()`
2. Try to change the base theme -- the UI freezes up

abidlabs:
This fixes it for me, thanks @hannahblair! Any idea why `data` is undefined in the first place?

PR #6536, opened 2023-11-22
Fix undefined `data` TypeError in Blocks 

hannahblair:
## Description

`data` is undefined on load in `Blocks.svelte` which causes a TypeError when `handle_update` runs. This PR adds a check to make sure the `forEach` func runs on `data` if it's truthy. 

To test:

- Run `theme_builder` demo
- Change the base theme

Closes: #6531

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/itchy-rocks-bow.md
+++ b/.changeset/itchy-rocks-bow.md
@@ -3,4 +3,4 @@
 "gradio": patch
 ---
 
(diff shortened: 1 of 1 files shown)


abidlabs:
This fixes it for me, thanks @hannahblair! Any idea why data is undefined in the first place?

abidlabs (approved):

PR #6538, opened 2023-11-22
Some tweaks to `ImageEditor`

abidlabs:
Some small tweaks to documentation, preprocess, and templates


aliabd:
@abidlabs I'm a bit confused why you added the imageeditor stuff to the 4.0 changelog?

hysts (approved):
LGTM!

gradio/components/image_editor.py
```diff
@@ -234,7 +235,7 @@ def preprocess(self, x: EditorData | None) -> EditorValue | None:
         if x is None:
             return x
 
-        bg = self.convert_and_format_image(x.background, crop_size=self.crop_size)
+        bg = self.convert_and_format_image(x.background)
```
pngwn, line 238: Ah nice catch. Didn't realise i screwed this up, explains the weird behaviour.

aliabd:
hmm @abidlabs not sure that makes too much sense, maybe we should create a separate migration to gradio 4 document, but idk if it makes sense adding it as a changelog entry to a version that doesn't support it. Maybe you should add a qualifier ("supported in 4.5") or something? Probably not a big deal

abidlabs:
@aliabd I do mention "as of Gradio 4.5.0" -- do you have something else in mind?

aliabd (approved):

freddyaboulton (approved):
I think this should be a patch no? Otherwise, looks good to me! Thanks @abidlabs !

abidlabs:
Thanks everyone for the reviews! I'll go ahead and merge it in after CI runs

Issue #6365, opened 2023-11-10
Documentation for older versions


BramVanroy:
It seems that documentation is only available for the most recent stable version as well as `main` but historical documentation is not there on https://www.gradio.app/docs/interface. Would be great if people who are stuck on older versions could also find the docs somewhere.

abidlabs:
Agreed, @aliabd is looking into this

aliabd:
Thanks for opening the issue @BramVanroy, will add support for 3.x this week.

PR #6549, opened 2023-11-22
Add 3.x docs to the website! 

aliabd:
Long overdue but this PR finally adds support for 3.x docs to the website. ~Right now it shows `3.50.1`, because of an issue during launch where the `3.50.2` jsons (latest `3.x`) were overwritten. This can be fixed and should be before we merge this PR (I can switch `3.50.1` with `3.50.2` on this PR once issue is fixed). Internal discussion: https://huggingface.slack.com/archives/C02SPHC1KD1/p1700014390607209~ Updated to 3.50.2

* New option in toggle for `3.50.2 `
* On `3.50.1` toggle (or `/3.50.2/docs/` urls) website uses the `3.50.2` embed gradio.js, and loads `3.50.2` demos from spaces. So all the demos are correct and reflect 3.x 

~There's one issue with guides, the embedded demos in the guide md just reference the name of the demo, which is a problem now that there are two versions. We will need to retroactively change those references to append a "_3-x" at the end of the name so that it loads the correct demo from spaces. I've done that here in this branch: https://github.com/gradio-app/gradio/tree/3-50-2-jsons This is the branch we should use to manually upload the `3.50.2` jsons. Other than that, the guides load the right gradio.js they should correctly reflect 3.x in their content and embedded demos once this is fixed~ Guides load correct demos with correct js :) 

closes: #6365


abidlabs:
Amazing @aliabd! Looks very clean

abidlabs (approved):

Issue #6483, opened 2023-11-19
Can't disable waveform for microphone


pseudotensor:
### Describe the bug

Upgraded from 3.50.2 to 4.4.1 to try out new features in h2oGPT, but hitting some issues.  One is that it seems like now the microphone always has to show a waveform, which for me is not desired.  This isn't listed in the breaking changes that waveform is required now, so I assume it's a bug or I'm missing something.

I tried `waveform_options=dict(show_controls=False)` passing to gr.Audio() but it has no effect.

Kinda related to this I guess: https://github.com/gradio-app/gradio/issues/6338

Related, it seems very graphics intensive, so definitely don't want to just make it invisible somehow if it's going to continue generating such frames.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```
audio = gr.Audio(sources=['microphone'], streaming=True, visible=True,
                                                         waveform_options=dict(show_controls=False),
                                                         )
```

### Screenshot

![image](https://github.com/gradio-app/gradio/assets/2249614/9d68b8a3-7d12-4968-bb12-015808f2a927)

In looking at demos at: https://www.gradio.app/main/docs/audio

I see this in demo, but I don't see those per-device controls ever:

![image](https://github.com/gradio-app/gradio/assets/2249614/06246296-dc91-4c95-8a44-2adad1291ef3)

### System Info

```shell
-gradio==3.50.2
+gradio==4.4.1

Blocking usage of new gradio.
```

### Severity

Blocking usage of gradio

pseudotensor:
thanks!

PR #6551, opened 2023-11-22
Add `show_recording_waveform` to Audio

hannahblair:
## Description

This change allows us to disable the rendering of the microphone during recording. I'm planning on doing the same for the playback waveform and replacing it with the native audio player, but I'd rather do that in a separate PR as there will be a few moving pieces. 

Closes: #6483

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/heavy-planes-dig.md
+++ b/.changeset/heavy-planes-dig.md
@@ -1,0 +1,5 @@
+---
+"gradio": minor
+---
+
+feat:Fix encoding issue #6364 of reload mode
--- a/.changeset/little-dodos-turn.md
+++ b/.changeset/little-dodos-turn.md
@@ -1,0 +1,7 @@
+---
+"@gradio/image": patch
+"@gradio/textbox": patch
+"gradio": patch
+---
+
+fix:Quick Image + Text Component Fixes
--- a/.changeset/purple-deer-double.md
+++ b/.changeset/purple-deer-double.md
@@ -1,0 +1,6 @@
+---
+"@gradio/video": patch
+"gradio": patch
+---
+
+fix:Improve video trimming and error handling
--- a/gradio/cli/commands/reload.py
+++ b/gradio/cli/commands/reload.py
@@ -29,9 +29,10 @@ def _setup_config(
     demo_path: Path,
     demo_name: str = "demo",
     additional_watch_dirs: list[str] | None = None,
+    encoding: str = "utf-8",
 ):
     original_path = Path(demo_path)
-    app_text = original_path.read_text()
+    app_text = original_path.read_text(encoding=encoding)
 
     patterns = [
         f"with gr\\.Blocks\\(\\) as {demo_name}",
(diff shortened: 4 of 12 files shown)


gradio/components/audio.py
```diff
@@ -17,13 +18,13 @@
...
 set_documentation_group("component")
 
-
-class WaveformOptions(TypedDict, total=False):
-    waveform_color: str
-    waveform_progress_color: str
-    show_controls: bool
-    skip_length: int
-
+@dataclasses.dataclass
```
abidlabs, line 21: Thanks for fixing this @hannahblair. For backwards compatibility, it'd be good if we still accepted a `dict` of waveform options (and internally converted to this dataclass)

abidlabs: Sorry I wasn't clear -- the dataclass for `WaveformOptions` you had before was great. But if someone has already written code where they pass `waveform_options` as a dictionary, we should continue to support that (by simply converting the dict to an object of type WaveformOptions`

e.g. something like:

```py
if isinstance(waveform_options, dict):
   waveform_options = WaveformOptions(**waveform_options)
```

gradio/components/audio.py
```diff
@@ -142,7 +145,12 @@ def __init__(
             if show_share_button is None
             else show_share_button
         )
-        self.waveform_options = waveform_options
+
+        self.waveform_options = (
```
abidlabs, line 149: looks good!

gradio/components/audio.py
```diff
@@ -18,11 +19,13 @@
...
-    waveform_progress_color: str
-    show_controls: bool
-    skip_length: int
+@dataclasses.dataclass
+class WaveformOptions:
+    waveform_color: Optional[str] = None
+    waveform_progress_color: Optional[str] = None
+    show_controls: Optional[bool] = False
+    skip_length: Optional[int] = None
+    show_recording_waveform: Optional[bool] = True
```
abidlabs, line 28: Nit: we usually adopt this style in the codebaser:
```suggestion
    waveform_color: str | None = None
    waveform_progress_color: str | None = None
    show_controls: bool | None = False
    skip_length: int | None = None
    show_recording_waveform: bool | None = True
```

abidlabs: Followup: should the two `bool` parameters actually be optional?

gradio/components/audio.py
abidlabs: Actually we may also need to handle the case when `waveform_options=None`. Right now, if I don't pass in any `waveform_options`, no waveform is shown while recording, instead of the default `WaveformOptions()`

abidlabs:
Looks good @hannahblair! Left a few nits on the backend (let me know if anything is unclear)

Also pushed a change so that we can directly access the `WaveformOptions` class as `gr.WaveformOptions`.

abidlabs:
It'd be good to use the `gr.WaveformOptions` class in one of our demos, maybe `reverse_audio`?

gradio/components/audio.py
abidlabs: Ah sorry just to clarify, I mean that we should do something like this:

```py
    waveform_color: str | None = None
    waveform_progress_color: str | None = None
    show_controls: bool = False
    skip_length: int | None = None
    show_recording_waveform: bool = True
```

So they are indeed optional and if they are not provided, they take the default values, but they do not take `None` as an option

Issue #6366, opened 2023-11-10
ValueError("Event not found", event_id)


SRbone:
### Describe the bug

```
Running on local URL:  http://127.0.0.1:7860

To create a public link, set `share=True` in `launch()`.
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/usr/local/anaconda3/envs/lib/python3.8/site-packages/uvicorn/protocols/http/h11_impl.py", line 408, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
  File "/usr/local/anaconda3/envs/lib/python3.8/site-packages/uvicorn/middleware/proxy_headers.py", line 84, in __call__
    return await self.app(scope, receive, send)
  File "/usr/local/anaconda3/envs/lib/python3.8/site-packages/fastapi/applications.py", line 1106, in __call__
    await super().__call__(scope, receive, send)
  File "/usr/local/anaconda3/envs/lib/python3.8/site-packages/starlette/applications.py", line 122, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/usr/local/anaconda3/envs/lib/python3.8/site-packages/starlette/middleware/errors.py", line 184, in __call__
    raise exc
  File "/usr/local/anaconda3/envs/lib/python3.8/site-packages/starlette/middleware/errors.py", line 162, in __call__
    await self.app(scope, receive, _send)
  File "/usr/local/anaconda3/envs/lib/python3.8/site-packages/starlette/middleware/cors.py", line 91, in __call__
    await self.simple_response(scope, receive, send, request_headers=headers)
  File "/usr/local/anaconda3/envs/lib/python3.8/site-packages/starlette/middleware/cors.py", line 146, in simple_response
    await self.app(scope, receive, send)
  File "/usr/local/anaconda3/envs/lib/python3.8/site-packages/starlette/middleware/exceptions.py", line 79, in __call__
    raise exc
  File "/usr/local/anaconda3/envs/lib/python3.8/site-packages/starlette/middleware/exceptions.py", line 68, in __call__
    await self.app(scope, receive, sender)
  File "/usr/local/anaconda3/envs/python3.8/site-packages/fastapi/middleware/asyncexitstack.py", line 20, in __call__
    raise e
  File "/usr/local/anaconda3/envs/lib/python3.8/site-packages/fastapi/middleware/asyncexitstack.py", line 17, in __call__
    await self.app(scope, receive, send)
[...]

Issue #6274, opened 2023-11-03
`ValueError: ('Event not found')`


KeremTurgutlu:
### Describe the bug

I've a gradio web UI which I've deployed in a google cloud VM. Everything works fine expect when I try to use the demo from my custom domain url `mydomain.com` it fails and hangs forever once an event is triggered.

When I trigger an event UI just hangs indefinitely, from the looks there seems to be a new event id which can't be found and it keeps trying with a new one. This just keeps on going with a new event id everytime and doesn't stop unless the browser is closed.

In Google Cloud I've opened HTTP, HTTPS and a custom TCP Port 7860.

After launching the demo I can successfully use the demo if I access it from https://External-ID:7860 however when I try to use it from my custom domain URL, it fails whenever I trigger an action such as clicking the submit button, changing slider values etc..This can also be seen in the logs.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

**Demo UI** 
```python
import gradio as gr
from gradio.components import Textbox, Radio
import json
from datetime import datetime
import pandas as pd
import os

# TODO: Add functionality to jump to any task by index, 
# also show if a task is already rated or not.
# Show a green checkmark if rated, red cross if not rated using Gradio.
# Workers should be able to overwrite their ratings if they want to.
class RateTexts:
    def __init__(self, ratings_filename=None, resume_from_filename=None, debug=False):
        
        if resume_from_filename is None:
            self.ratings = {}
        else:
            with open(resume_from_filename, "r") as f:
                self.ratings = json.load(f)

        self.current_index = 0

        # Sample list of tuples
        self.texts = [
            ("Main Text 1", "Option A1", "Option B1", "Option C1"),
            ("Main Text 2", "Option A2", "Option B2", "Option C2"),
            ("Main Text 3", "Option A3", "Option B3", "Option C3"),
            ("Main Text 4", "Option A4", "Option B4", "Option C4"),
        ]
        
        # load texts from file if None raise erro
        if not debug:
            if ratings_filename is None:
                raise ValueError("Please provide a ratings csv filename with 4 columns, 1 main text and 3 options text.")
            
            if os.path.exists(ratings_filename):    
                self.texts = [tuple(o) for o in pd.read_csv(ratings_filename).values]
            else:
                raise ValueError(f"Ratings filename {ratings_filename} doesn't exist.")
        
    def rate_texts(self, rating):
        # Save the rating
        if self.current_index < 0:
[...]

PR #6556, opened 2023-11-23
Fix api event drops

aliabid94:
I believe this should fix the issue of random "Event ID not found". 

Previously, we had two http requests that get fired for each prediction - first the SSE connection that gets establishes a continuous connection for the duration of the request, and then later when this session is popped off the queue in the backend, the SSE requests data from the frontend, so the frontend fires off a POST request to upload the data. However, the backend only waits 5 seconds for the data to get uploaded, after that it assumes the connection has dropped. So if the frontend's ability to send that POST request gets blocked and it takes longer than 5 seconds to upload data, we see the error we are seeing (event id not found when uploading data, because the event was killed). 

I'm not sure if 100% of these issues can be explained by this, but what I think was happening was the the follow-up POST request was being blocked because of the limitation of 6 pending HTTP connections limit of browsers to the same domain. If there were multiple long SSE requests pending, then the follow POST request would get queued up and sent past the 5 second timeout.   

My solution for now is to send the POST request with the data first, and then immediately after establish an SSE connection. This way, we will never have the problem of the backend requesting data when the queue is free but no data is available. This means the backend will hold on to data for sessions before the session gets popped off the queue. This also means that for connections that get broken between the POST request and the SSE request (even though these requests are right after each other), the backend holds on to the data indefinitely. To prevent this from leading to an ever-expanding data structure in memory, I store the data from the POST request in an LRUCache with a maximum size of 1000 items. I think this will be more than plenty.

Because the API has changed, I updated the clients to support this new API. 

Fixes: #6366 #6274 

To test this:

Run this simple demo:
```python
import gradio as gr


def greet(name):
    import time
    time.sleep(20)
    return "Hello " + name + "!"


with gr.Blocks() as demo:

if __name__ == "__main__":
    demo.launch()
```

Then run this demo across 6+ tabs and submit simultaneously. Old gradio would have `event_id not found` error, now resolved.


client/python/gradio_client/utils.py
```diff
@@ -413,6 +421,75 @@ async def stream_sse(
...
+            sse_data_url,
+            json={**data, **hash_data},
+            cookies=cookies,
+        )
+        req.raise_for_status()
+        resp = req.json()
+        event_id = resp["event_id"]
+
+        async with client.stream(
+            "GET", sse_url, params={"event_id": event_id}, cookies=cookies
```
abidlabs, line 444: Needs to include `headers` as done here: https://github.com/gradio-app/gradio/pull/6602

abidlabs:
Started testing this and it does seem like the issue goes away. But there might actually be a much simpler way to solve this issue by using the “max Keep-Alive requests” header as described here: https://www.geeksforgeeks.org/what-are-max-parallel-http-connections-in-a-browser/

> For example, if the “Keep-Alive” header specifies a “max Keep-Alive requests” value of 100, it means that the browser can initiate up to 100 requests over the persistent connection before it has to re-establish a new connection.

Can we try this approach first so that we don't end potentially with other issues that may arise due to this refactor?

aliabid94:
> Can we try this approach first so that we don't end potentially with other issues that may arise due to this refactor?

This PR is overall a better approach, as the queue will never be paused because it is waiting for data from a client. It is also needed to move in the direction of one-event-stream-per-browser-session instead of one-event-stream-per-event-trigger.

abidlabs (approved):
Tested this with a lot of different demos. Looks good to me, and can confirm that the issue in the repro you provided is fixed. 

We should add some tests that ensure that this works nicely with Spaces. Could you create 2 spaces -- one private and one public that have this PR installed? And make sure that the `gr.Client` is able to work successfully with both Spaces?

aliabid94:
Actually I'm gonna rework this PR to instead use a single event source for a whole browser session. Should scale better.

freddyaboulton:
Nice work! Have not reviewed all the code yet but was thinking about what you said about the server needing to close all running tasks. I wonder if the cleanest way to do that is on `app.on_event("shutdown")` as opposed to during the stream request itself?

aliabid94:
> Nice work! Have not reviewed all the code yet but was thinking about what you said about the server needing to close all running tasks. I wonder if the cleanest way to do that is on app.on_event("shutdown") as opposed to during the stream request itself?

This was one of my approaches. Issues I faced:
- shutdown was not always called
- there is no record of connections to kill. So I had to track all connections as generators, and forcibly close them on shutdown. 
- if an backend is killed in this manner,the frontend has no idea the backend is killed, so it keeps trying to reconnect. This was killing our tests, which never ended because the client thread kept running.

client/python/gradio_client/client.py
```diff
@@ -1068,24 +1158,32 @@ def process_predictions(self, *predictions):
...
                 hash_data,
                 helper,
-                sse_url=self.client.sse_url,
-                sse_data_url=self.client.sse_data_url,
-                headers=self.client.headers,
-                cookies=self.client.cookies,
+                self.client.sse_url,
+                self.client.sse_data_url,
+                self.client.headers,
+                self.client.cookies,
```
abidlabs, line 1171: jw why did we remove the kwargs?

client/python/gradio_client/utils.py
```diff
@@ -307,29 +323,61 @@ async def get_pred_from_ws(
...
+    event_id: str,
+) -> dict[str, Any] | None:
+    done, pending = await asyncio.wait(
+        [
+            asyncio.create_task(check_for_cancel(helper, cookies)),
+            asyncio.create_task(
+                stream_sse_v1(
+                    helper,
+                    pending_messages_per_event,
+                    event_id,
```
abidlabs, line 380: also here I believe you'll need to send headers

gradio/routes.py
```diff
@@ -664,8 +655,14 @@ async def queue_data(
             request: fastapi.Request,
             username: str = Depends(get_current_user),
         ):
-            blocks = app.get_blocks()
-            blocks._queue.attach_data(body)
+            from datetime import datetime
+
+            print("recieved data @", datetime.now().strftime("%H:%M:%S.%f")[:-3])
```
abidlabs, line 660: remove print statement

gradio/routes.py
```diff
@@ -583,61 +580,36 @@ async def predict(
 
         @app.get("/queue/join", dependencies=[Depends(login_check)])
```
freddyaboulton, line 581: The route names are a bit confusing now. `queue/data` joins the queue and `queue/join` gives you the data for the prediction lol. Can we switch?

freddyaboulton (commented):
Nice @aliabid94 ! Confirmed it fixes the issue and the deployed demos still look good. Left some comments.

client/python/gradio_client/client.py
```diff
@@ -152,6 +155,78 @@ def __init__(
...
+        resp = req.json()
+        event_id = resp["event_id"]
+
+        if not self.stream_open:
+            self.stream_open = True
+
+            def open_stream():
+                return utils.synchronize_async(self.stream_messages)
+
+            if self.streaming_future is None or self.streaming_future.done():
```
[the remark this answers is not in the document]
aliabid94: The backend currently only supports a stream listening to a session_id. This makes it impossible to have a separate stream per event from the client as you suggest, unless every client call had a separate session_id. We probably don't want that behaviour because of session state.
I like the idea of a stream per event, but that would mean creating a separate endpoint just for that in the backend. It would be complex to support both endpoints at the same time, something we could consider in another PR.

client/python/gradio_client/client.py
```diff
@@ -1068,24 +1158,32 @@ def process_predictions(self, *predictions):
...
+                self.client.headers,
+                self.client.cookies,
             )
 
+    async def _sse_fn_v1(self, helper: Communicator, event_id: str):
+        return await utils.get_pred_from_sse_v1(
+            helper,
+            self.client.cookies,
+            self.client.pending_messages_per_event,
+            event_id,
```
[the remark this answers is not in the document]
aliabid94: this function doesn't make any backend calls, other than check_for_cancel, which doesn't use headers previously either. Will add to check_for_cancel though

gradio/routes.py
```diff
@@ -664,8 +655,14 @@ async def queue_data(
...
         ):
-            blocks = app.get_blocks()
-            blocks._queue.attach_data(body)
+            from datetime import datetime
+
+            print("recieved data @", datetime.now().strftime("%H:%M:%S.%f")[:-3])
+            if blocks._queue.server_app is None:
+                blocks._queue.set_server_app(app)
+
+            event_id = await blocks._queue.push(body, request, username)
```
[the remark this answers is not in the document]
aliabid94: wdym by "connect after a job finishes"? Specifically if you try to connect to the SSE stream when there are no pending events, it will throw an error. But the client doesn't let the user arbitrarily connect to the stream endpoint.

client/python/gradio_client/utils.py
```diff
@@ -100,6 +113,7 @@ class Status(Enum):
     PROGRESS = "PROGRESS"
     FINISHED = "FINISHED"
     CANCELLED = "CANCELLED"
+    LOG = "LOG"
```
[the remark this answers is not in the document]
aliabid94: yes that's what log messages are. previously we ignored all log messages in the python client and I wanted them stored somewhere. could seperate log messages in a separate data structure in a separate PR?

client/python/gradio_client/client.py
```diff
@@ -152,6 +155,77 @@ def __init__(
...
+
+        if not self.stream_open:
+            self.stream_open = True
+
+            def open_stream():
+                return utils.synchronize_async(self.stream_messages)
+
+            if self.streaming_future is None or self.streaming_future.done():
+                self.streaming_future = self.executor.submit(open_stream)
+                self.streaming_future.add_done_callback(
```
freddyaboulton, line 223: Let's remove this print statement

client/python/gradio_client/client.py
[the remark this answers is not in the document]
freddyaboulton: Got it! We talked about exploring "api-only" usage next year. Would be good to revisit this then because the api is getting complex (for valid reasons) but it limits the ability to use gradio from another language

freddyaboulton (approved):
Nice work @aliabid94 !

PR #6565, opened 2023-11-23
Fix uploaded file wasn't moved to custom temp dir at different disks

dodysw:
## Description

When gradio temporary directory is overridden to a different disk path, uploads were not accessible via UI (e.g. audio can't be played). This is due to a ?code typo? below.  

Closes: #(issue)

## 🎯 PRs Should Target Issues

## Tests


freddyaboulton:
Thanks for the fix! Can you please fix the type error? Should just need to call `str(dest)`.

Issue #4966, opened 2023-07-18
Chatbot height should be optionally 100% or not set at all


fazpu:
Chatbot component accepts `height` as input in pixels. If not defined, the default value is set to 400px.

However, there should be a third option - it should be possible to either set it to a percent value (e.g. `100%`) or **not set it at all**.

The issue is that the `height` is set in the `style` attribute. Consequently, you can't override it via css of the `id` or `class` of given `div`, because `style` value always takes precedence.

<img width="2542" alt="image" src="https://github.com/gradio-app/gradio/assets/18228330/061f28ce-db86-41e2-8702-e58c30e2a5ad">

you can also check out the branch:
https://github.com/h2oai/h2ogpt/compare/jp/main/full-height?expand=1

abidlabs:
I remember @dawoodkhan82 @aliabid94 you discussed this before. Thoughts on allowing a percentage height as well?

Solomin0:
+1 for this. more control over the chatbot display would be great. 
Dragging a corner would be ideal from my POV, providing some congruence with textboxes and letting the users easily adjust from the same screen instead of having them modify a height slider somewhere.

PR #6569, opened 2023-11-24
Allow passing height and width as string in `Blocks.svelte`

hannahblair:
## Description

Allows height and width to be passed as a string into `Blocks.svelte`, thus allowing height to be set as a percentage (or any unit available for height). This will apply to any components that accept `height` 👀 

Closes: #4966

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/guides/03_building-with-blocks/02_controlling-layout.md
+++ b/guides/03_building-with-blocks/02_controlling-layout.md
@@ -55,7 +55,7 @@ You can control the height and width of various components, where the parameters
 
 Below is an example illustrating the use of viewport width (vw):
 
-```
+```python
 import gradio as gr
 
 with gr.Blocks() as demo:
@@ -69,7 +69,7 @@ demo.launch()
 When using percentage values for dimensions, you may want to define a parent component with an absolute unit (e.g. `px` or `vw`). This approach ensures that child components with relative dimensions are sized appropriately:
 
 
-```
+```python
 import gradio as gr
 
 css = """


js/atoms/src/Block.svelte
```diff
@@ -27,7 +38,7 @@
 	class:padded={padding}
 	class:border_focus={border_mode === "focus"}
 	class:hide-container={!explicit_call && !container}
-	style:height={typeof height === "number" ? height + "px" : undefined}
+	style:height={getHeight()}
```
aliabid94, line 41: I don't believe this is reactive, e.g. if height changes this won't update. `height` should be an argument to this function to make it reactive

aliabid94:
Didn't work for me:

```python
with gr.Blocks() as demo:
    name = gr.Chatbot(value=[["1", "2"]], height="100%")
    greet_btn = gr.Button("Test")
```

shows
Screen Shot 2023-11-24 at 11 01 02 AM

Also I think it's somewhat unclear what the percentage would be relative to.

abidlabs:
> This should actually be documented for clarity. Maybe I'll update the Python docstring maybe? Open to thoughts

Agreed about documentation. It might be a lot to squeeze into the docstrings. What about adding a section in this Guide: https://www.gradio.app/guides/controlling-layout

gradio/components/chatbot.py
dawoodkhan82, line 1: We should update ` height: int | None = None,` to include `string` type. Also should we update the other components that have a `height` param?

dawoodkhan82 (changes requested):

pngwn (approved):
I would echo @dawoodkhan82's comment about the docstrings but other than that looks great. Thanks @hannahblair!

aliabid94 (approved):
LGTM!

pngwn (approved):
Looks good, thanks for taking care of the improvements @hannahblair.

Feel free to dismiss the stale review and merge.

pngwn (commented):
tweaks for formatting

Issue #5871, opened 2023-10-11
Upvote/Downvote Functionality Issue


yvrjsharma:
### Describe the bug

- Expected Behavior:

When a user upvotes or downvotes a chat message, the functionality should reflect the selected vote, visually indicating that it has been clicked, and subsequently disable further clicks on the same button. Additionally, the other vote button should disappear to prevent multiple votes on the same message.

- Current Behavior:

Currently, you can click on the upvote or downvote button multiple times. While the color of the button changes to indicate that it has been clicked, it remains clickable, and the other vote button doesn't disappear as expected.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

1. Run the code provided below.
2. Send a chat message and await the bot's reply.
3. Click the upvote or downvote button repeatedly.

```
import gradio as gr
import random

def vote(tmp, index_state, data: gr.LikeData):
    value_new = data.value
    index_new = data.index
    if len(index_state) == 0 :
        index_state.append(index_new)
    else:
        if index_new in index_state:
            return "Your feedback is already saved", index_state
        else:
            index_state.append(index_new)
    return str(data.value) + ";" + str(data.index)+";"+ str(data.liked)+";"+str(index_state), index_state

with gr.Blocks() as demo:
    tmp = gr.Textbox(visible=True, value="")
    chatbot = gr.Chatbot(layout='panel')
    msg = gr.Textbox()
    clear = gr.ClearButton([msg, chatbot])
    index_state = gr.State(value=[])
    
    def respond(message, chat_history):
        bot_message = random.choice(["How are you?", "I love you", "I'm very hungry"])
        chat_history.append((message, bot_message))
        return "", chat_history

    msg.submit(respond, [msg, chatbot], [msg, chatbot])
    chatbot.like(vote, [tmp, index_state], [tmp, index_state])

demo.launch()

```

### Screenshot

The below video explains what is currently happening - 
https://github.com/gradio-app/gradio/assets/48665385/edcfd67f-848f-46a2-a866-4afda61a138b

The below video explains how ChatGpt handles such a situation - 
https://github.com/gradio-app/gradio/assets/48665385/fa5faa7e-dcc1-4e1b-8513-24e007f7179c

### System Info

```shell
Gradio Version: 3.47.1
```

### Severity

I can work around it

yvrjsharma:
Also, it would be ideal if only the chosen option remained visible throughout the chat session. Thanks @hannahblair.

[...]

PR #6572, opened 2023-11-24
Improve like/dislike functionality

hannahblair:
## Description

Ensures like and dislike are mutually exclusive events - i.e. a message can either be liked or disliked. 

I've added the like/dislike logic to the `chatbot_component` demo since we didn't have an example of it yet and added an e2e test using this demo. 

Closes: #5871 

## 🎯 PRs Should Target Issues

## Tests


js/app/test/chatbot_component.spec.ts
```diff
@@ -0,0 +1,18 @@
+import { test, expect } from "@gradio/tootils";
+
+test("chatbot like and dislike functionality", async ({ page }) => {
+	const textbox = await page.getByTestId("textbox");
```
abidlabs, line 4: What is this textbox? I don't see one in the `chatbot_component` demo?

abidlabs:
Thanks so much @hannahblair for fixing this. Tested and LGTM except just confused about the test (see comment).

abidlabs (approved):

js/app/test/chatbot_component.spec.ts
abidlabs: Ah makes sense then cool!

Issue #6430, opened 2023-11-15
Gradio APIs breaks calling private endpoints on HF


fffiloni:
### Describe the bug

I think i found the reason for bugs while calling gradio APIs:

Here you can try several examples : https://huggingface.co/spaces/fffiloni/calling-gradio-api-tests
The demo is calling endpoints from this one : https://huggingface.co/spaces/fffiloni/gradio-api-tests [which is PRIVATE]

Everything works fine if the endpoints space is PUBLIC, otherwise everything breaks if the endpoints space is PRIVATE
GRADIO VERSION: 4.3.0

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

Here you can try several examples : https://huggingface.co/spaces/fffiloni/calling-gradio-api-tests

### Logs

```shell
Traceback (most recent call last):
  File "/home/user/.local/lib/python3.10/site-packages/gradio/queueing.py", line 427, in call_prediction
    output = await route_utils.call_process_api(
  File "/home/user/.local/lib/python3.10/site-packages/gradio/route_utils.py", line 232, in call_process_api
    output = await app.get_blocks().process_api(
  File "/home/user/.local/lib/python3.10/site-packages/gradio/blocks.py", line 1525, in process_api
    result = await self.call_function(
  File "/home/user/.local/lib/python3.10/site-packages/gradio/blocks.py", line 1147, in call_function
    prediction = await anyio.to_thread.run_sync(
  File "/home/user/.local/lib/python3.10/site-packages/anyio/to_thread.py", line 33, in run_sync
    return await get_asynclib().run_sync_in_worker_thread(
  File "/home/user/.local/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 877, in run_sync_in_worker_thread
    return await future
  File "/home/user/.local/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 807, in run
    result = context.run(func, *args)
  File "/home/user/.local/lib/python3.10/site-packages/gradio/utils.py", line 672, in wrapper
    response = f(*args, **kwargs)
  File "/home/user/app/app.py", line 17, in api_t2i
    result = client.predict(
  File "/home/user/.local/lib/python3.10/site-packages/gradio_client/client.py", line 305, in predict
    return self.submit(*args, api_name=api_name, fn_index=fn_index).result()
  File "/home/user/.local/lib/python3.10/site-packages/gradio_client/client.py", line 1456, in result
[...]

Issue #5927, opened 2023-10-16
gr.load() failing with Error


amitagh:
### Describe the bug

I am trying load a pvt space from a public space.
below is code for public space:
demo = gr.load(name="spaces/amitagh/TestChatIntf", hf_token = MY_HF_TOKEN_KEY)
demo.launch()

-----------------------
pvt space amitagh/TestChatIntf has just following simple code:
import gradio as gr

def echo(message, history):
    return message

demo = gr.ChatInterface(fn=echo, examples=["hello", "hola", "merhaba"], title="Echo Bot")
demo.launch()

Getting below error:
Fetching Space from: https://huggingface.co/spaces/amitagh/TestChatIntf
Loaded as API: https://amitagh-testchatintf.hf.space ✔
/home/user/app/app.py:9: GradioUnusedKwargWarning: You have unused kwarg parameters in Chatbot, please remove them: {'likeable': False}
  demo = gr.load(name="spaces/amitagh/TestChatIntf", hf_token = hf_key, api_key = hf_key)
Traceback (most recent call last):
  File "/home/user/app/app.py", line 9, in <module>
    demo = gr.load(name="spaces/amitagh/TestChatIntf", hf_token = hf_key, api_key = hf_key)
  File "/home/user/.local/lib/python3.10/site-packages/gradio/external.py", line 70, in load
    return load_blocks_from_repo(
  File "/home/user/.local/lib/python3.10/site-packages/gradio/external.py", line 109, in load_blocks_from_repo
    blocks: gradio.Blocks = factory_methods[src](name, hf_token, alias, **kwargs)
  File "/home/user/.local/lib/python3.10/site-packages/gradio/external.py", line 504, in from_spaces
    return from_spaces_blocks(space=space_name, hf_token=hf_token)
  File "/home/user/.local/lib/python3.10/site-packages/gradio/external.py", line 510, in from_spaces_blocks
    return gradio.Blocks.from_config(client.config, predict_fns, client.src)
  File "/home/user/.local/lib/python3.10/site-packages/gradio/blocks.py", line 801, in from_config
    targets = [
  File "/home/user/.local/lib/python3.10/site-packages/gradio/blocks.py", line 803, in <listcomp>
    original_mapping[
KeyError: None

abidlabs:
Hi @amitagh can you confirm both Spaces are running the latest version of Gradio? The Gradio version is set by the the `sdk_version` in the README file

amitagh:
[...]

PR #6602, opened 2023-11-29
Fix: Gradio Client work with private Spaces

abidlabs:
In theory, this bug should have been caught by `test_private_space_v4` test, but the `https://huggingface.co/spaces/gradio-tests/not-actually-private-spacev4-sse` Space was accidentally set to public! Updated the test to ensure that the Spaces are private.

Fixes: #6430
Fixes: #6388
Fixes: #5927


aliabid94 (approved):
LGTM

abidlabs:
Thanks @aliabid94

freddyaboulton:
Nice thanks @abidlabs !!

Issue #6606, opened 2023-11-29
Passing a FileExplorer without selection to a gradio.Button.click gives an error


v-chabaux:
### Describe the bug

I pass a FileExplorer with nothing select to a Button.click() and i have an error. So i have no way to check if its empty or not.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```
def get_values(a):
  pass

output_folder = gr.FileExplorer(root="/content/drive/",glob="**/", file_count="single", value = ["/content/drive/MyDrive/"])
scrap_button.click(get_values, inputs=output_folder, outputs=None)

```

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/gradio/queueing.py", line 456, in call_prediction
    output = await route_utils.call_process_api(
  File "/usr/local/lib/python3.10/dist-packages/gradio/route_utils.py", line 232, in call_process_api
    output = await app.get_blocks().process_api(
  File "/usr/local/lib/python3.10/dist-packages/gradio/blocks.py", line 1520, in process_api
    inputs = self.preprocess_data(fn_index, inputs, state)
  File "/usr/local/lib/python3.10/dist-packages/gradio/blocks.py", line 1303, in preprocess_data
    processed_input.append(block.preprocess(inputs_cached))
  File "/usr/local/lib/python3.10/dist-packages/gradio/components/file_explorer.py", line 116, in preprocess
    return self._safe_join(payload.root[0])
IndexError: list index out of range
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/gradio/queueing.py", line 456, in call_prediction
    output = await route_utils.call_process_api(
  File "/usr/local/lib/python3.10/dist-packages/gradio/route_utils.py", line 232, in call_process_api
    output = await app.get_blocks().process_api(
  File "/usr/local/lib/python3.10/dist-packages/gradio/blocks.py", line 1520, in process_api
    inputs = self.preprocess_data(fn_index, inputs, state)
  File "/usr/local/lib/python3.10/dist-packages/gradio/blocks.py", line 1303, in preprocess_data
    processed_input.append(block.preprocess(inputs_cached))
  File "/usr/local/lib/python3.10/dist-packages/gradio/components/file_explorer.py", line 116, in preprocess
    return self._safe_join(payload.root[0])
IndexError: list index out of range

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
[...]

PR #6607, opened 2023-11-30
Update file_explorer.py - Fixing error if nothing selected in file_count=single mode (return None rather)

v-chabaux:
Fixing error if nothing selected in file_count=single mode (return None rather)

Closes: #6606


abidlabs:
LGTM. Will add a couple of tests and then this should be good to go

abidlabs:
Merging -- thanks again @v-chabaux for this bug report and quick fix!

Issue #6364, opened 2023-11-10
Encoding issues in newer Gradio versions


BramVanroy:
### Describe the bug

In 3.33.1, my gradio app could be launched perfectly fine on both Windows and Linux. However, when I upgraded to the most recent version, I was getting encoding issues left and right. I traced it back to the following line:

https://github.com/gradio-app/gradio/blob/4d3aad33a0b66639dbbb2928f305a79fb7789b2d/gradio/cli/commands/reload.py#L34

If I change that to the following everything works as intended.

```python
app_text = Path(original_path).read_text(encoding="utf-8")
```

I understand that people may DO want to use cp1252 on Windows, or any other encoding scheme, but it'd be great to allow for more control here. Either by an environment variable or even a launch argument where a user can specify the encoding to use.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

I cloned this repo and tried to run it locally: https://huggingface.co/spaces/uonlp/open_multilingual_llm_leaderboard

### Logs

```shell
[notice] A new release of pip available: 22.3.1 -> 23.3.1
[notice] To update, run: python.exe -m pip install --upgrade pip
(leaderboard) PS F:\python\open_multilingual_llm_leaderboard> gradio .\app.py
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ F:\python\open_multilingual_llm_leaderboard\.venv\lib\site-packages\gradio\cli\commands\reload.p │
│ y:97 in main                                                                                     │
│                                                                                                  │
│    94 │   demo_path: Path, demo_name: str = "demo", watch_dirs: Optional[List[str]] = None       │
│    95 ):                                                                                         │
│    96 │   # default execution pattern to start the server and watch changes                      │
│ ❱  97 │   filename, path, watch_dirs, demo_name = _setup_config(                                 │
│    98 │   │   demo_path, demo_name, watch_dirs                                                   │
│    99 │   )                                                                                      │
│   100 │   # extra_args = args[1:] if len(args) == 1 or args[1].startswith("--") else args[2:]    │
│                                                                                                  │
│ ╭────────────── locals ──────────────╮                                                           │
│ │  demo_name = 'demo'                │                                                           │
│ │  demo_path = WindowsPath('app.py') │                                                           │
│ │ watch_dirs = []                    │                                                           │
│ ╰────────────────────────────────────╯                                                           │
│                                                                                                  │
[...]

PR #6622, opened 2023-12-02
Fix encoding issue #6364 of reload mode

curiousRay:
## Description

This fix is target on Windows platform.

In reload mode, the dev-created Python script shall be crunched via cmd, however the script's encoding format is not informed for it. This PR fix the problem.

Meanwhile, if I'd like to use GBK encoding (or anything but UTF-8) for my script, such error occurs:

```
UnicodeDecodeError: 'gbk' codec can't decode byte 0xac in position 2517: illegal mutibyte sequence
```

Therefore I added a `encoding` parameter into cli, allowing devs free to specify any encoding format they want.

Doc changes are included in the PR.

Closes: [#6364]

### DIFF
--- a/.changeset/heavy-planes-dig.md
+++ b/.changeset/heavy-planes-dig.md
@@ -1,0 +1,5 @@
+---
+"gradio": minor
+---
+
+feat:Fix encoding issue #6364 of reload mode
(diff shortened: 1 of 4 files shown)


freddyaboulton (approved):
Thanks @curiousRay !

Issue #6426, opened 2023-11-15
gr.Blocks head argument not working as intended


jonathanasdf:
Looking at the html source for the gradio app and testing this a bit more, it seems like only the first script tag from the head is being applied

```
"""
head = """
<script>console.log('hello')</script>
<script>console.log('world')</script>
"""
with gr.Blocks(head=head):
```

![image](https://github.com/gradio-app/gradio/assets/511073/f19f7b8a-a1fc-4161-ac8c-27039cdc8d71)

Additionally, in this case it seems like the script tag is added to the head section after the page already loads, so it never gets executed? The console output is empty.

_Originally posted by @jonathanasdf in https://github.com/gradio-app/gradio/issues/5954#issuecomment-1811749402_

abidlabs:
cc @dawoodkhan82

PR #6639, opened 2023-12-03
Fix issue with `head` param when adding more than one script tag

dawoodkhan82:
## Description
Fixes issue with `head` param when adding more than one script tag.

Closes: #6426

## 🎯 PRs Should Target Issues

## Tests


dawoodkhan82:
@abidlabs Trying to debug, not sure what changed.

abidlabs:
@dawoodkhan82 just fyi I think this was the case before as well -- the script would not get executed

abidlabs:
Based on this part of the original issue:

> Additionally, in this case it seems like the script tag is added to the head section after the page already loads, so it never gets executed? The console output is empty.

pngwn:
The script being added later shouldn't stop it from being executed.

dawoodkhan82:
Fixed, realized I was pushing to the wrong branch...

dawoodkhan82:
Also, now it works when trying to call an external `js` file:

```
import gradio as gr
import os

path = os.path.join(os.path.dirname(__file__), "test.js")
head = """
<script type="text/javascript" src={}></script>
<script>console.log('hello')</script>
""".format(path)

with gr.Blocks(head=head) as demo:
    pass

demo.launch()
```

freddyaboulton:
I took a pass and I wasn't able to get the external script loading to work. @dawoodkhan82 is looking into it! 🙏

abidlabs:
Beautiful! Works for me. 

The original code I shared works as well as loading js from an external PR. 

Here's what I tried:

```py
import gradio as gr
import os

head = """
<script type="text/javascript" src="/file=test.js"></script>
<script>console.log('hello')</script>
""".format(path)

with gr.Blocks(head=head) as demo:
    pass

demo.launch(allowed_paths=["test.js"])
```

Here's `test.js`:

```
console.log('world');
```

abidlabs (approved):
Awesome stuff @dawoodkhan82

dawoodkhan82:
@abidlabs Thanks for the review!
@freddyaboulton does adding `demo.launch(allowed_paths=["test.js"])` to the demo fix the issue you were seeing?

abidlabs:
merging ahead of the release

Issue #6646, opened 2023-12-04
Clicking on the docs sidebar doesn't navigate you exactly to the relevant section in the docs


abidlabs:
When I click on the sidebar in docs (e.g. the "Authentication" section in "Sharing Your App"), it doesn't take me _exactly_ to the correct section, you can see its slightly off:

<img width="1246" alt="image" src="https://github.com/gradio-app/gradio/assets/1778297/d894c04d-ffa4-430b-a0a0-165e2f8d7441">

However, if I click on the section header, then it snaps into place and also updates the URL with the anchor link:

<img width="1339" alt="image" src="https://github.com/gradio-app/gradio/assets/1778297/c7accc63-e537-4e41-a48a-73cc200f09b2">

The latter behavior is what I expect would happen when I click on the sidebar as well.

PR #6650, opened 2023-12-04
Removes smooth scrolling from website

aliabd:
Fixes the issue with the navbar not leading to correct position on the page, and adds the target elem's id to the url to make sharing easier. Keeps the nice smooth scrolling effect (just via css now) 

there's one small issue still which is that smooth scrolling finds the element and its offset from the top and scrolls to it (nice effect so user sees a glimpse of the guide/element's position) but while this is happening if the page is still loading the target element's position could shift down. so the scroll will go to above it, but that's only if the page hasn't loaded fully so i don't think it's a big deal (try testing the preview). If it's still a problem i can remove smooth scrolling

EDIT: removes smooth scrolling altogether 

closes: #6646


abidlabs:
I lean towards it being more important to scroll to the right place rather than the smooth scroll effect. I did notice it scrolling to an incorrect position first while the apps are still loading. Not a huge deal.

Also it seems that if you drop the direct link in the browser, it doesn't seem to scroll to the right section. For example if I visit this link: https://gradio-jju9hexib-hugging-face.vercel.app/guides/quickstart#chatbots, it doesn't go to the chatbots section. Does it go for you?

aliabd:
@abidlabs actually i agree going to kill smooth scrolling

abidlabs:
Working great now, thanks @aliabd!

PR #6658, opened 2023-12-04
Hotfix related to stories

abidlabs:
Some quick fixes to get the storybook CI to work.

### DIFF
--- a/js/video/Video.stories.svelte
+++ b/js/video/Video.stories.svelte
@@ -76,23 +76,7 @@
 	}}
 	play={async ({ canvasElement }) => {
 		const canvas = within(canvasElement);
-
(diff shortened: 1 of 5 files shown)


js/video/Video.stories.svelte
```diff
@@ -76,23 +76,7 @@
 	}}
 	play={async ({ canvasElement }) => {
 		const canvas = within(canvasElement);
-
 		const trimButton = canvas.getByLabelText("Trim video to selection");
-
 		userEvent.click(trimButton);
-
-		await new Promise((r) => setTimeout(r, 4000));
```
abidlabs, line 84: This timeout was not sufficiently long enough for the element to appear, causing the storybook CI action to crash. @hannahblair I think we'll need to use a different approach.

PR #6666, opened 2023-12-05
Set gradio api server from env

aisensiy:
## Description

Set share link server from env `GRADIO_API_SERVER` so that I can make the gradio automatically use another gradio share server without changing the source code.

## 🎯 PRs Should Target Issues

This fix is very small, so I do not create an issue for it.

## Tests

### DIFF
--- a/.changeset/odd-garlics-teach.md
+++ b/.changeset/odd-garlics-teach.md
@@ -1,0 +1,5 @@
+---
+"gradio": minor
+---
+
+feat:Set gradio api server from env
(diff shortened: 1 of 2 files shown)


freddyaboulton:
Do we need the extra environment variable here? You can use `share_server_address` in `launch`. Also see here about using your own share server: https://github.com/huggingface/frp

abidlabs:
Hi @aisensiy we could add this environment variable (we do have a similar environment variables for the arguments in `.launch()` such as GRADIO_SERVER_PORT), but it should take the place of the `share_server_address` argument in `launch()`.

So we would need to make two changes.

1. Let's call the environment variable GRADIO_SHARE_SERVER_ADDRESS
2. It should set the value of `share_server_address` unless an explicit parameter for `share_server_address` has been provided

Would you be able to make these two changes?

abidlabs:
LGTM, thanks @aisensiy!

PR #6669, opened 2023-12-05
Update HF token used in CI tests

abidlabs:
The `gradio-tests` org token that we were previously using as part of our tests is now invalidated. So now, I've added the Gradio team to the `gradio-tests` to the org. Also created an `team@gradio.app` account on HF, and added its token to the GitHub CI. 

So now our individual HF accounts are being used in the tests instead of the org token.

### DIFF
--- a/.changeset/big-camels-refuse.md
+++ b/.changeset/big-camels-refuse.md
@@ -1,5 +1,0 @@
----
-"gradio": minor
----
-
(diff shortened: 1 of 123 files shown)


abidlabs:
Env variable isn't being read. I'm going to close this PR and re-open in case that helps.

Issue #6510, opened 2023-11-20
Add an easy way to maintain session state within `gr.ChatInterface` other than just appending history of messages


andrewrreed:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
Seems like [a common use case](https://discuss.huggingface.co/t/session-state-in-the-new-chatinterface/51374/1) that you'd want to use some type of agent based memory (e.g. ConversationalBufferMemory from LangChain) that handles state for you.

**Describe the solution you'd like**  
I have an example that's using gr.State as an additional_component. It works for maintaining state across sessions, however it doesn't respect the Clear button. Is there an easy way to do this or is it best to just build separately with gr.Blocks?

Here's an example app, where get_agent_executor loads an initialized agent from LangChain:

```python
from agent import get_agent_executor


import gradio as gr
from dotenv import load_dotenv

load_dotenv(override=True)


def predict(message, history, agent_executor):
    res = agent_executor.invoke({"input": message})

    return res["output"]


gr.ChatInterface(
    predict,
    additional_inputs=[gr.State(get_agent_executor)],
    chatbot=gr.Chatbot(show_label=False),
    textbox=gr.Textbox(),
    undo_btn=None,
).launch(debug=True, show_api=False)
```

I think there should be optional behavior so that clicking the Clear button clears the additional inputs.

andrewrreed:
Related to #6465

Issue #6465, opened 2023-11-17
ClearButton cannot reset State()


nuzar:
### Describe the bug

Click ClearButton() cannot reset State().
Maybe the same problem as https://github.com/gradio-app/gradio/issues/885 .

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr


def fn(state: int):
    state += 1
    return str(state), state


with gr.Blocks() as demo:
    state = gr.State(value=0)
    textbox = gr.Textbox(
        interactive=True,
        label="press enter to submit",
    )
    textbox.submit(
        fn,
        [state],
        [textbox, state],
    )
    clear_button = gr.ClearButton([textbox, state])

demo.launch()
```

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Linux
gradio version: 4.4.0
gradio_client version: 0.7.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.1.2
fastapi: 0.103.2
ffmpy: 0.3.1
gradio-client==0.7.0 is not installed.
httpx: 0.25.0
huggingface-hub: 0.18.0
importlib-resources: 6.1.0
jinja2: 3.1.2
markupsafe: 2.1.3
matplotlib: 3.8.0
numpy: 1.25.2
orjson: 3.9.9
packaging: 23.2
pandas: 2.1.0
pillow: 10.0.1
pydantic: 2.4.2
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0.1
requests: 2.31.0
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.9.0
typing-extensions: 4.8.0
uvicorn: 0.23.2
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2023.9.2
httpx: 0.25.0
huggingface-hub: 0.18.0
packaging: 23.2
requests: 2.31.0
typing-extensions: 4.8.0
websockets: 11.0.3
```

### Severity

I can work around it

nuzar:
We can register an event listener on clear button, is this the recommended way?
```python
    clear_button.click(
        lambda _: 0,
        [state],
        [state],
    )
```

PR #6680, opened 2023-12-06
Cause `gr.ClearButton` to reset the value of `gr.State` 

abidlabs:
Simple repro:

```py
import gradio as gr

gr.Interface(lambda x,y:(x+y, x+y), [gr.Textbox(), gr.State("")], [gr.Textbox(), gr.State("")]).launch()
```

Fixes: #6465
Fixes: #6510


freddyaboulton (approved):
Nice fix @abidlabs !

Issue #6686, opened 2023-12-06
FileExplorer can't recursively list directories only


vchabaux:
### Describe the bug

Since update #6550 in fileExplorer I can no longer recursively list only directories with parameter glob='**/'. This worked in version 4.7.1. Now it only displays directories in the root directory.
I'm terrible with glob, maybe there's a better approach (that works) for my case, however I affirm that this worked before the fix #6550

thx

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr
from google.colab import drive
import os

if not os.path.exists("/content/drive/MyDrive/") :
    drive.mount('/content/drive/')

with gr.Blocks() as demo:
  with gr.Tab("Collecter des documents"):      
    output_folder = gr.FileExplorer(glob="**/", file_count="single")

demo.launch(debug=True)

```

### Screenshot

![image](https://github.com/gradio-app/gradio/assets/149407738/14635063-70d4-4c4b-b333-e1e53cfa79bc)

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Linux
gradio version: 4.8.0
gradio_client version: 0.7.1

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 4.2.2
fastapi: 0.104.1
ffmpy: 0.3.1
gradio-client==0.7.1 is not installed.
httpx: 0.25.2
huggingface-hub: 0.19.4
importlib-resources: 6.1.1
jinja2: 3.1.2
markupsafe: 2.1.3
matplotlib: 3.7.1
numpy: 1.23.5
orjson: 3.9.10
packaging: 23.2
pandas: 1.5.3
pillow: 9.4.0
pydantic: 2.5.2
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0.1
requests: 2.31.0
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.9.0
typing-extensions: 4.8.0
uvicorn: 0.24.0.post1
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2023.6.0
httpx: 0.25.2
huggingface-hub: 0.19.4
packaging: 23.2
requests: 2.31.0
typing-extensions: 4.8.0
websockets: 11.0.3
```

### Severity

Blocking usage of gradio

PR #6689, opened 2023-12-06
Fix directory-only glob for FileExplorer

freddyaboulton:
## Description

Closes: #6686

Added a unit test too. The problem was that the previous glob-based implementation assumed directory paths always ended with `/`. When we switched to pathlib so that it could run on 3.8, we did not take that into account.

## 🎯 PRs Should Target Issues

## Tests


abidlabs:
Otherwise, lgtm

abidlabs:
Thanks @freddyaboulton! It does fix the empty file, but there's a different issue, which is that that there's no way to select a directory. Per @vchabaux's comment, previously you'd select the "empty file" to select the directory, but now there's no such file. 

I think the way it should work is that you should do something like this:

```py
import gradio as gr


demo = gr.Interface(
    lambda x:x,
    gr.FileExplorer(glob="**/", file_count="directory"),
    "textbox"
)

demo.launch()
```

and then you should be able to select a directory to pass into your prediction function. However, that doesn't seem to work -- nothing gets passed into the function even if you select a directory.

abidlabs:
Taking another look

abidlabs:
@freddyaboulton I'm not able to select folders. What's the code for your demo? Here's what I tried:

```py
import gradio as gr


demo = gr.Interface(
    lambda x:x,
    gr.FileExplorer(glob="**/", file_count="single"),
    "textbox"
)

demo.launch()
```

abidlabs:
Hmm that behavior doesn't make sense to me -- so `single` doesn't allow you to select directories, but `multiple` does?

vchabaux:
Initially my usage was to select a directory on the machine where python runs (and only one) as the directory for my output. I think this usage is quite common, and it would be sad not to be able to use FileExplorer for this.

freddyaboulton:
@pngwn @abidlabs i think this is finally working as intended. If file_count=`single`, clicking on a directory won't automatically select all its children.

abidlabs:
@freddyaboulton here's my proposal to address the point (2) above: we add an `include_dirs` parameter in `FileExplorer`, which is `False` by default. Essentially it allows users to select directories in `FileExplorer`, but it has two effects: one effect in the UI, and one effect in the data that is passed into the user function

We can consider the cases `file_count=="single"` and `file_count=="multiple"` separately:

* If `file_count=="single"` and `include_dirs is True`, then one is able to select a directory, and it only selects that particular directory. Naturally, it passes the selected directory into the user function. Note that if `include_dirs is False` and `file_count=="single"`, we keep the current behavior, which disables directory selection altogether. 

 
* If `file_count="multiple"` and `include_dirs is True`, then I'd suggest that selecting a directory *does not* automatically select its children. Multiple directories and/or files are selected, and they are all passed into the user function individually. If `include_dirs is False` we do not pass directories into the user function. Selecting a directory has the effect of selecting of its all children files and only those files are passed into the user function.

tldr: setting `include_dirs` to `True` allows users to select directories individually in the `FileExplorer` and those are sent to the user function. Keeping `include_dirs` as `False` guarantees that only files are sent to the user function, no directories.

freddyaboulton:
@abidlabs @pngwn I implemented what we discussed earlier - including the directory in the file tree with a new directory icon and the value ".".

The default glob pattern excludes directories (that's behavior on `main`). You can glob for everything with `glob="**/*"` and you can ignore directories by using the default glob or with `ignore_glob="**/"`. You can include only directories with `glob="**/"`.

Here is a gif showing what it looks like now:

file_count_multiple

js/fileexplorer/icons/light-folder.svg
abidlabs, line 1: Nice where's this svg from?

abidlabs:
Fantastic solution @freddyaboulton! Tested it out and working well for me

abidlabs:
I think @vchabaux will be a fan too

abidlabs (approved):

vchabaux:
Awesome solution, thanks a lot @freddyaboulton  !

Issue #6601, opened 2023-11-28
gradio 4 critical bug -- all return messages truncated to 65k


pseudotensor:
### Describe the bug

In gradio 3 there were never any issues with returning large amounts of text or data.  However, in gradio 4 this is totally broken.

This is super critical bug given how broadly it applies to all API calls of any return type (text, audio, etc.).

Related: https://github.com/gradio-app/gradio/issues/6319

that is, even once heartbeat bug is fixed, long output still hits json error mentioned there.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

server:
```
import gradio as gr
import random
import time

with gr.Blocks() as demo:
    chatbot = gr.Chatbot()
    msg = gr.Textbox()
    clear = gr.Button("Clear")

    def user(user_message, history):
        return "", history + [[user_message, None]]

    def bot(history):
        bot_message = ' '.join(['a'] * 100000)
        history[-1][1] = bot_message
        yield history

    msg.submit(user, [msg, chatbot], [msg, chatbot], queue=False).then(
        fn=bot, inputs=chatbot, outputs=chatbot, api_name='bot',
    )
    clear.click(lambda: None, None, chatbot, api_name='clear')

demo.queue()
demo.launch()
```

client:
```
import time
from gradio_client import Client

client = Client('http://localhost:7860', serialize=False)

args = [[['Who are you?', None]]]
res = client.predict(*tuple(args), api_name='/bot')
print(res)
```

Error:

```
(/data/conda/h2ogpt) jon@pseudotensor:~/h2ogpt$ python testchat_nostream_client.py 
Loaded as API: http://localhost:7860/ ✔
Traceback (most recent call last):
  File "/home/jon/h2ogpt/testchat_nostream_client.py", line 7, in <module>
    res = client.predict(*tuple(args), api_name='/bot')
  File "/data/conda/h2ogpt/lib/python3.10/site-packages/gradio_client/client.py", line 305, in predict
    return self.submit(*args, api_name=api_name, fn_index=fn_index).result()
  File "/data/conda/h2ogpt/lib/python3.10/site-packages/gradio_client/client.py", line 1456, in result
    return super().result(timeout=timeout)
  File "/data/conda/h2ogpt/lib/python3.10/concurrent/futures/_base.py", line 445, in result
    return self.__get_result()
  File "/data/conda/h2ogpt/lib/python3.10/concurrent/futures/_base.py", line 390, in __get_result
[...]

Issue #6541, opened 2023-11-22
gradio_client fails when full queue


osanseviero:
### Describe the bug

I'm trying to use a Space as an API via `gradio_client`, but I get an error due to full queue (I think) with a `KeyError` message

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```
from gradio_client import Client

client = Client("https://multimodalart-stable-video-diffusion.hf.space/--replicas/z2trf/")
result = client.predict(
		"https://raw.githubusercontent.com/gradio-app/gradio/main/test/test_files/bus.png",	# filepath  in 'Upload your image' Image component
		0,	# float (numeric value between 0 and 9223372036854775807) in 'Seed' Slider component
		True,	# bool  in 'Randomize seed' Checkbox component
		1,	# float (numeric value between 1 and 255) in 'Motion bucket id' Slider component
		5,	# float (numeric value between 5 and 30) in 'Frames per second' Slider component
							api_name="/video"
)
print(result)
```

### Logs

```shell
Loaded as API: https://multimodalart-stable-video-diffusion.hf.space/--replicas/z2trf/ ✔
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-4-ec496743e655> in <cell line: 4>()
      2 
      3 client = Client("https://multimodalart-stable-video-diffusion.hf.space/--replicas/z2trf/")
----> 4 result = client.predict(
      5                 "https://raw.githubusercontent.com/gradio-app/gradio/main/test/test_files/bus.png",     # filepath  in 'Upload your image' Image component
      6                 0,      # float (numeric value between 0 and 9223372036854775807) in 'Seed' Slider component

13 frames
/usr/local/lib/python3.10/dist-packages/gradio_client/client.py in predict(self, api_name, fn_index, *args)
    303                 "Cannot call predict on this function as it may run forever. Use submit instead."
    304             )
--> 305         return self.submit(*args, api_name=api_name, fn_index=fn_index).result()
    306 
    307     def new_helper(self, fn_index: int) -> Communicator:

/usr/local/lib/python3.10/dist-packages/gradio_client/client.py in result(self, timeout)
   1454             >> 9
   1455         """
-> 1456         return super().result(timeout=timeout)
   1457 
   1458     def outputs(self) -> list[tuple | Any]:

/usr/lib/python3.10/concurrent/futures/_base.py in result(self, timeout)
    456                     raise CancelledError()
[...]

Issue #6494, opened 2023-11-19
gradio_client fails most of the time with JSONDecodeError (on a Chat interface that provides results in streaming)


Jeronymous:
### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Describe the bug

I have an app that is using `gradio.ChatInterface` on a function that is a generator (`Iterator[str]`).
It is hosted on HuggingFace and very similar to  [Llama-2-7b-chat](https://huggingface.co/spaces/huggingface-projects/llama-2-7b-chat/blob/main/app.py#L44)

When I want to request this API with `gradio_client` using the code  given by HuggingFace to do so, I have the following failure (with my Chat, not with Llama Chat).

A difference between Llama Chat and my Chat is the version of `gradio` in the docker container:
* Llama2 is using `gradio==3.48.0`
* I am using more recent `gradio==4.4.1`

Another difference who may explains why it fail with my chat and not Llama2 is that the last token might be generated super quickly (without inference step in a neural network, just a last normalization step).

Error message:
```
Traceback (most recent call last):
  File "/home/jlouradour/.local/lib/python3.10/site-packages/gradio_client/client.py", line 305, in predict
    return self.submit(*args, api_name=api_name, fn_index=fn_index).result()
  File "/home/jlouradour/.local/lib/python3.10/site-packages/gradio_client/client.py", line 1456, in result
    return super().result(timeout=timeout)
  File "/usr/lib/python3.10/concurrent/futures/_base.py", line 458, in result
    return self.__get_result()
  File "/usr/lib/python3.10/concurrent/futures/_base.py", line 403, in __get_result
    raise self._exception
  File "/usr/lib/python3.10/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/jlouradour/.local/lib/python3.10/site-packages/gradio_client/client.py", line 869, in _inner
    predictions = _predict(*data)
  File "/home/jlouradour/.local/lib/python3.10/site-packages/gradio_client/client.py", line 894, in _predict
    result = utils.synchronize_async(self._sse_fn, data, hash_data, helper)
  File "/home/jlouradour/.local/lib/python3.10/site-packages/gradio_client/utils.py", line 666, in synchronize_async
    return fsspec.asyn.sync(fsspec.asyn.get_loop(), func, *args, **kwargs)  # type: ignore
[...]

Issue #6319, opened 2023-11-07
KeyError: 'heartbeat' when python client request took more than approximately 16 seconds


LeoAtlanto:
### Describe the bug

**Bug Info**
I'm working on a project where several hosted gradio apps are connected by a python gradio client. One of the hosted gradio apps is a bit complicated and it would take about 30 seconds to finish the request. Under such circumstances, the error _"KeyError: 'heartbeat'"_ occurs and the request fails.

**Solution**
```/usr/local/lib/python3.10/dist-packages/gradio_client/utils.py```
Add a key-value pair in function msg_to_status below line 136, or something like this:
```
    @staticmethod
    def msg_to_status(msg: str) -> Status:
        """Map the raw message from the backend to the status code presented to users."""
        return {
            "send_hash": Status.JOINING_QUEUE,
            "queue_full": Status.QUEUE_FULL,
            "estimation": Status.IN_QUEUE,
            "send_data": Status.SENDING_DATA,
            "process_starts": Status.PROCESSING,
            "process_generating": Status.ITERATING,
            "process_completed": Status.FINISHED,
            "progress": Status.PROGRESS,
            "heartbeat": Status.PROGRESS,      # add this line
        }[msg]

```

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

1.Create two scripts.

(1) host.py

```
import time
import gradio as gr

def greet(name):
    time.sleep(20)
    return "Hello " + name + "!"

demo = gr.Interface(fn=greet, inputs="text", outputs="text")
    
if __name__ == "__main__":
    demo.launch(show_api=False, server_name='0.0.0.0', server_port=12345)
```
(2) client.py
```
import gradio as gr
from gradio_client import Client

def greet(name):
    client = Client(src="http://xxx.xxx.xxx.xxx:12345")
    job = client.submit(name, fn_index=0)
    return job.result()

demo = gr.Interface(fn=greet, inputs="text", outputs="text")
    
if __name__ == "__main__":
    demo.launch(show_api=False, server_name='0.0.0.0', server_port=12346)
```
2.Launch two apps.
```
python host.py
python client.py
```
3.Use the client app to connect the host app.

### Logs

```shell
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/gradio/queueing.py", line 427, in call_prediction
    output = await route_utils.call_process_api(
[...]

PR #6693, opened 2023-12-06
Python client properly handles hearbeat and log messages. Also handles responses longer than 65k

freddyaboulton:
## Description

Closes: #6319 
Closes: #6601 
Closes: #6541 
Closes: #6494
Closes: #6776

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/client/python/gradio_client/utils.py
+++ b/client/python/gradio_client/utils.py
@@ -162,9 +162,7 @@ class Status(Enum):
             ServerMessage.process_generating: Status.ITERATING,
             ServerMessage.process_completed: Status.FINISHED,
             ServerMessage.progress: Status.PROGRESS,
-        }[
-            msg
-        ]  # type: ignore
+        }[msg]  # type: ignore
 
 
 @dataclass
--- a/client/python/gradio_client/utils.py
+++ b/client/python/gradio_client/utils.py
@@ -162,7 +162,9 @@ class Status(Enum):
             ServerMessage.process_generating: Status.ITERATING,
             ServerMessage.process_completed: Status.FINISHED,
             ServerMessage.progress: Status.PROGRESS,
-        }[msg]
+        }[
+            msg
+        ]  # type: ignore
 
 
 @dataclass
--- a/client/python/gradio_client/utils.py
+++ b/client/python/gradio_client/utils.py
@@ -113,6 +113,7 @@ class ServerMessage(str, Enum):
     log = "log"
     progress = "progress"
     heartbeat = "heartbeat"
+    server_stopped = "server_stopped"
 
 
 class Status(Enum):
@@ -162,6 +163,8 @@ class Status(Enum):
             ServerMessage.process_generating: Status.ITERATING,
             ServerMessage.process_completed: Status.FINISHED,
             ServerMessage.progress: Status.PROGRESS,
+            ServerMessage.log: Status.LOG,
+            ServerMessage.server_stopped: Status.FINISHED,
         }[msg]  # type: ignore
 
 
@@ -519,7 +522,7 @@ async def stream_sse_v1(
 
             with helper.lock:
                 log_message = None
-                if msg["msg"] == "log":
+                if msg["msg"] == ServerMessage.log:
                     log = msg.get("log")
                     level = msg.get("level")
                     if log and level:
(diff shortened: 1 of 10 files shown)


freddyaboulton:
I believe test is a flake. Will look into it but this should be good to review.

abidlabs:
Awesome! Taking a look

gradio/routes.py
```diff
@@ -659,7 +663,14 @@ async def queue_join(
             if blocks._queue.server_app is None:
                 blocks._queue.set_server_app(app)
 
-            event_id = await blocks._queue.push(body, request, username)
+            succes, event_id = await blocks._queue.push(body, request, username)
+            if not succes:
+                status_code = (
+                    status.HTTP_503_SERVICE_UNAVAILABLE
```
abidlabs, line 669: Nice

abidlabs:
Just to understand, what was the root cause of https://github.com/gradio-app/gradio/issues/6319?

freddyaboulton:
> Just to understand, what was the root cause of https://github.com/gradio-app/gradio/issues/6319?

I think there were two problems. 1) the "heartbeat" message was not being handled by the client 2) `aiter_text` was yielding an incomplete message (can't be parsed to json) if it was very long

gradio/routes.py
```diff
@@ -659,7 +663,14 @@ async def queue_join(
             if blocks._queue.server_app is None:
                 blocks._queue.set_server_app(app)
 
-            event_id = await blocks._queue.push(body, request, username)
+            succes, event_id = await blocks._queue.push(body, request, username)
+            if not succes:
```
abidlabs, line 667:
```suggestion
            success, event_id = await blocks._queue.push(body, request, username)
            if not success:
```

client/python/gradio_client/utils.py
```diff
@@ -436,9 +451,14 @@ async def stream_sse_v0(
...
         ) as response:
-            async for line in response.aiter_text():
+            async for line in response.aiter_lines():
+                line = line.rstrip("\n")
+                if len(line) == 0:
+                    continue
                 if line.startswith("data:"):
                     resp = json.loads(line[5:])
+                    if resp["msg"] in [ServerMessage.log, ServerMessage.heartbeat]:
+                        continue
```
abidlabs, line 461: Note that these changes were made in `stream_sse_v0`. Do you also want to make similar changes (particularly the heartbeat one) in `stream_sse_v1`?

gradio/queueing.py
```diff
@@ -144,15 +145,16 @@ def _resolve_concurrency_limit(self, default_concurrency_limit):
 
     async def push(
         self, body: PredictBody, request: fastapi.Request, username: str | None
-    ):
+    ) -> tuple[bool, str]:
```
abidlabs, line 148: This might have broken `max_size`. 

I ran this demo and submitted jobs on 3 separate tabs:

```py
import time
import gradio as gr

def greet(name):
    time.sleep(20)
    return "Hello " + name + "!"

demo = gr.Interface(fn=greet, inputs="text", outputs="text").queue(max_size=1)
    
if __name__ == "__main__":
    demo.launch(show_api=False)
```

On the final tab, I see this error modal (instead of one telling me that the queue is full):

image

abidlabs:
I took a first pass on this PR, and I can confirm https://github.com/gradio-app/gradio/issues/6319
and https://github.com/gradio-app/gradio/issues/6601 have been fixed. 

I left some small comments above @freddyaboulton if you can take a look. Happy to take another pass on this PR soon!

gradio/queueing.py
freddyaboulton: @abidlabs Did you build the frontend before running?

client/python/gradio_client/utils.py
freddyaboulton: Whoops - yea I think it should be both!

freddyaboulton: Actually the stream_sse_v1 case is handled in `Client.stream_messages`! Stream_sse_v1 doesn't directly connect to the SSE endpoint like v0 does. It just pulls messages off a message queue.

gradio/queueing.py
abidlabs: I did not in fact :) but after rebuilding it, I still don't see any modal saying that the queue is full. Instead, on the 3rd tab, I just see a job that never enters the queue:

image

abidlabs:
Thanks @freddyaboulton I took another look and this is great! 

Just wanted to note a few things:

* #6319 is fixed
* #6601 is fixed
* #6541 is fixed in the sense that now we get `QueueError: Queue is full! Please try again.` in the Client. 
* #6494 - the original Space is down so I can't repro but it looks like you've captured the gist of the repro in a unit test
* #6776 is not fixed for me, even after rebuilding the frontend

abidlabs (approved):
Fixed for me now, thank you! Great PR @freddyaboulton 

Let's do a patch release with this PR and #6525

freddyaboulton:
Thanks @abidlabs !

Issue #6596, opened 2023-11-28
With the Dropdown in Gradio 4.7.1 and tuple values, it's possible to clear / blank the selection


cpbotha:
### Describe the bug

With the dropdown in gradio 4.7.1 fed with a list of tuples, it's possible to click outside of the component and then have it clear its selection completely.

If I supply that list of tuples, and a default value, my expectation is that it will always have a value selection thereafter. See minimal reproducer below.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

Run the following example. For the top dropdown, click it, then click outside, and notice it's now blank (value: None)

With the second dropdown, this is not possible.

The only difference is that the first one uses list of tuples value, and the second is just list of values.

```python
import gradio as gr

with gr.Blocks(
    title="dropdown tuples bug reproducer",
) as demo:
    # click outside of the dropdown, at which point it will go blank (NO selection)
    gr.Dropdown(
        label="Dropdown with tuples",
        choices=[("Model 1", "model1"), ("Model 2", "model2")],
        value="model1",
        interactive=True,
    )
    # this dropdown maintains its selection
    gr.Dropdown(
        label="Dropdown with values",
        choices=["model1", "model2"],
        value="model1",
        interactive=True,
    )

demo.queue()
demo.launch()
```

### Screenshot

<img width="425" alt="image" src="https://github.com/gradio-app/gradio/assets/937871/19cfdc64-851a-4bb2-85cd-6e007210171f">

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Darwin
gradio version: 4.7.1
gradio_client version: 0.7.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.1.2
fastapi: 0.104.0
ffmpy: 0.3.1
gradio-client==0.7.0 is not installed.
httpx: 0.25.0
huggingface-hub: 0.18.0
importlib-resources: 6.1.0
jinja2: 3.1.2
markupsafe: 2.1.3
matplotlib: 3.8.0
numpy: 1.26.1
orjson: 3.9.9
packaging: 23.2
pandas: 2.1.1
pillow: 10.1.0
pydantic: 2.4.2
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0.1
requests: 2.31.0
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.9.0
[...]

PR #6694, opened 2023-12-06
Fix dropdown blur bug when values are provided as tuples

abidlabs:
Fixes: #6596 and adds a test

Test with:

```py
import gradio as gr

with gr.Blocks(
    title="dropdown tuples bug reproducer",
) as demo:
    # click outside of the dropdown, at which point it will go blank (NO selection)
    gr.Dropdown(
        label="Dropdown with tuples",
        choices=[("Model 1", "model1"), ("Model 2", "model2")],
        value="model1",
        interactive=True,
    )
    # this dropdown maintains its selection
    gr.Dropdown(
        label="Dropdown with values",
        choices=["model1", "model2"],
        value="model1",
        interactive=True,
    )

demo.launch()
```

### DIFF
--- a/.changeset/small-crabs-take.md
+++ b/.changeset/small-crabs-take.md
@@ -1,0 +1,6 @@
+---
+"@gradio/dropdown": patch
+"gradio": patch
+---
+
+fix:Fix dropdown blur bug when values are provided as tuples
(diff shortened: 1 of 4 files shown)


abidlabs:
Thanks a bunch @dawoodkhan82!

PR #6709, opened 2023-12-08
Remove progress animation on streaming

aliabid94:
Fixes: #1494 

Also improves UI when streaming from webcam. To test:

```python
import gradio as gr

def iden(img):
    return img

with gr.Blocks() as demo:
    with gr.Row():
        input_img = gr.Image(sources=["webcam"])
        output_img = gr.Image()
        input_img.stream(lambda x:x, input_img, output_img)

if __name__ == "__main__":
    demo.launch(show_api=False) 
```

```python
import gradio as gr

def iden(img):
    return img

with gr.Blocks() as demo:
    with gr.Row():
        input_img = gr.Audio(sources=["microphone"])
        output_img = gr.Audio()
        input_img.stream(lambda x:x, input_img, output_img)

if __name__ == "__main__":
    demo.launch(show_api=False) 
```


js/image/shared/ImageUploader.svelte
```diff
@@ -153,7 +154,8 @@
 	{/if}
 	<div class="upload-container">
 		<Upload
-			hidden={value !== null || active_tool === "webcam"}
+			hidden={value !== null ||
```
freddyaboulton, line 157: Should we disable upload progress from the webcam in general? Still seeing it when you take a picture from webcam without streaming.

webcam_upload_progress

Other than that (and the type error) this looks good to me! Thanks for the fix

abidlabs:
@nateraw will be happy we finally got around to this issue!

js/image/shared/ImageUploader.svelte
pngwn: I really think we should just use the local binary data for the preview image (same for all media) with a subtle uploading animation + block the submit button/ submit target until it is done.

For another PR though. Will open an issue.

Issue #3851, opened 2023-04-14
setting css not effective


17818587795:
### Describe the bug

Unable to add background color to Button

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

```python
import gradio as gr

with gr.Blocks(css="#warning {background-color: red}  .feedback {font-size: 24px}") as demo:
    box1 = gr.Textbox(value="Good Job", elem_classes="feedback")
    box2 = gr.Button(value="Failure", elem_id="warning", elem_classes="feedback")

if __name__ == "__main__":
    demo.launch(debug=True)
```

### Logs

```shell
no logs
```

### System Info

```shell
3.23.0
```

### Severity

annoying

gtm-nayan:
There are two problems here,
1. There is a background-image property, that's drawn over the background-color. So, even if your style for background is actually applying to the element, it's going to covered by the gradient that's set as the background-image.

2. Your font-size from feedback won't apply to the element, because `.lg.svelte-xyz` will have more specificity than just `.feedback`.

The workarounds specific to this example would be to set background-image to none in your css and repeat the feedback class to be `.feedback.feedback`, but I doubt the workarounds will be of much use in general.

abidlabs:
Hi @17818587795 for the custom css, you have to look at the inspector console to figure out how to target your elements correctly. And, if you use custom classes, you may have to use "!important" for specificity as @gtm-nayan points out. If you do go down this route, you can achieve what you want like this:

```py
import gradio as gr

css = """
#warning {background: red;} 

.feedback {font-size: 24px !important;}
.feedback textarea {font-size: 24px !important;}
"""

with gr.Blocks(css=css) as demo:
    box1 = gr.Textbox(value="Good Job", elem_classes="feedback")
    box2 = gr.Button(value="Failure", elem_id="warning", elem_classes="feedback")
    
demo.launch()
```

which produces this:

<img width="576" alt="image" src="https://user-images.githubusercontent.com/1778297/232513804-5cadd217-91ae-4bce-bf0a-7f1f02363f3b.png">

However, custom css isn't a recommended approach in Gradio and it may not work consistently across Gradio versions. What we recommend instead is to use themes: https://gradio.app/theming-guide/

PR #6738, opened 2023-12-11
reload on css changes + fix css specificity

### LINKED ISSUE #6735: Dev mode does not reload custom css

When running dev mode either normally or via the cc dev command, custom csss is not reloaded. You have to restart the server manually.

### LINKED ISSUE #6736: Writing custom css is frustrating

Writing custom css is incredibly frustrating, to the point where I'm not sure how people have the patience to even bother. 

We need to improve this somehow. Have a few ideas that I'll add shortly.

pngwn:
## Description

This has been driving me up the wall all weekend, so here is a fix.

- Ensure the app reloads + css is correctly applied when CSS files in a watch dir are edited. cc @freddyaboulton 
- Ensure CSS is correctly applied when CSS (as a string) changed.
- Make working with custom CSS much simpler (genuine fix for #3851) cc @aliabid94 @abidlabs 

This should work fine now with no need for `!important` modifiers or other CSS hacks:

```python
import gradio as gr

css = """
#custom-md h1 {
  color: hotpink;
}

.custom-md h2 {
  color: orange;
}
"""

  gr.Markdown("# Heading one", elem_id="custom-md")
  gr.Markdown("## Heading two", elem_classes=["custom-md"])

demo.launch()
```

We do this by modifying the selectors passed by a user, prepending them with a series of addition selectors to increase their specificity enough that they override the svelte hashed classes. It's hacky but should be relatively robust.

Closes #6735. Closes #6736.

### DIFF
--- a/js/app/src/css.ts
+++ b/js/app/src/css.ts
@@ -18,3 +18,39 @@ export function mount_css(url: string, target: HTMLElement): Promise<void> {
 		target.appendChild(link);
 	});
 }
+
+export function prefix_css(
+	string: string,
+	version: string,
+	style_element = document.createElement("style")
+): HTMLStyleElement {
+	//  create style element
+	style_element.remove();
+
+	const stylesheet = new CSSStyleSheet();
+	stylesheet.replaceSync(string);
+
+	const rules = stylesheet.cssRules;
+
+	let css_string = "";
+
+	for (let i = 0; i < rules.length; i++) {
+		const rule = rules[i];
+
+		if (rule instanceof CSSStyleRule) {
+			const selector = rule.selectorText;
+			if (selector) {
+				const new_selector = selector
+					.split(",")
+					.map((s) => `.gradio-container-${version} .contain ${s.trim()}`)
+					.join(",");
+
+				css_string += rule.cssText.replace(selector, new_selector);
+			}
+		}
+	}
+	style_element.textContent = css_string;
+
+	document.head.appendChild(style_element);
+	return style_element;
+}
--- a/js/app/src/css.ts
+++ b/js/app/src/css.ts
@@ -24,7 +24,6 @@ export function prefix_css(
 	version: string,
 	style_element = document.createElement("style")
 ): HTMLStyleElement {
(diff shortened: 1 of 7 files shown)


js/app/src/css.ts
```diff
@@ -18,3 +18,39 @@ export function mount_css(url: string, target: HTMLElement): Promise<void> {
...
+		if (rule instanceof CSSStyleRule) {
+			const selector = rule.selectorText;
+			if (selector) {
+				const new_selector = selector
+					.split(",")
+					.map((s) => `.gradio-container-${version} .contain ${s.trim()}`)
+					.join(",");
+
+				css_string += rule.cssText.replace(selector, new_selector);
+			}
```
pngwn, line 49: Lets leave the existing styles in as well, as they are, in case they are targetting something outside of the gradio app + to avoid breaking changes.

pngwn:
`gradio cc dev` was completely broken on main because of a dependency issues, I've fixed that in this branch too.

pngwn:
Need to merge to check some other issues. Feel free to review and I'll address anything in another PR

js/app/src/css.ts
```diff
@@ -18,3 +18,42 @@ export function mount_css(url: string, target: HTMLElement): Promise<void> {
...
+				const new_selector = selector
+					.split(",")
+					.map(
+						(s) =>
+							`gradio-app .gradio-container.gradio-container-${version} .contain ${s.trim()}`
+					)
+					.join(",");
+
+				css_string += rule.cssText;
+				css_string += rule.cssText.replace(selector, new_selector);
```
abidlabs, line 51: Nice. I don't think repeating a css rule should have any (non-performance-related) downsides so this should be good

abidlabs:
Cool, looks great to me @pngwn! Thanks for dogfooding this workflow :)

js/app/src/css.ts
pngwn: The number of rules should generally be pretty modest, so I think the performance should be ok.

Issue #5128, opened 2023-08-08
Expose option to allow/forbid users to edit audio


Bengt:
- [x] I have searched to see if a similar issue already exists.

**Related problem**

I want to parse information from the audio file name, as that is common practice in the domain I am writing my app for. When users edit an audio file added in gradio's audio element, parsing breaks because the file now has a temporary file name like so:

```
/tmp/gradio/e0e507a69b71e1e5cc9d7b83d9b16435abb4caf7/audio-1-100.wav
```

**Solution I'd like** 
 
Expose an option to disable the file editing feature of gradio's audio element. E.g.

```python
audio: gradio.Audio = gradio.Audio(
    editable=False,
)
```

Currently, `interactive=True` means both that users can upload and that users can edit, while I want to have my users upload, but not edit. I would suppose to distinguish between these two:

![image](https://github.com/gradio-app/gradio/assets/11575/69d85d23-5cf2-45b1-8b11-41154e12f41b)

Source: https://www.gradio.app/docs/audio

**Additional context**  

I am parsing the date and time of a recording audio from its file name. So, another solution that would work for me would be to name the temporary files similarly to the uploaded file names, e.g.:

```
/tmp/gradio/e0e507a69b71e1e5cc9d7b83d9b16435abb4caf7/Prefix_2023-08-08-08T15-27-57-1-100.wav
```

Bengt:
For the time being, this CSS hides the slider element so that users can no longer edit any audio files, which is fine in my case:

```css
div.rangeSlider.range.hoverable {
    display: none;
}
```

Bengt:
Hello @hannahblair! Thanks for taking this on. I am glad there is an easy fix for this, but I would not have known how to implement it in this project.

Bengt:
Thanks for the quick turnaround! This change is already live in my automatically deployed BirdNET application on Hugging Face:

<https://huggingface.co/spaces/XPRIZE-Brazilian-Team/Bengt-BirdNET-Annotator>

thiswillbeyourgithub:
Unfortunately this feature was removed in version 4.0+ I think. Too bad.

thiswillbeyourgithub:
Thanks @hannahblair a

For anyone else wondering, here's my js workaround to hide some stuff:
(be careful of #6729)
```
import gradio as gr
hide_some_components = """
() => {
const hideElements = (selector) => {
  const elements = document.querySelectorAll(selector);
  elements.forEach(el => {
[...]

PR #6745, opened 2023-12-11
Add `editable` parameter to Audio

hannahblair:
## Description

We removed the `show_edit_button` because editing was no longer done via an edit button in the component (currently done via a trim button), and we therefore allowed all interactive elements to be editable. That said, preventing users from editing audio still has its use cases. This PR adds the functionality back in, via an `editable` param. In theory, this just means users can't use the trimming functionality, but naming this `editable` creates space for any further editing functionality we may add in future. 

Closes: #5128 (again 😅)

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/js/audio/shared/WaveformControls.svelte
+++ b/js/audio/shared/WaveformControls.svelte
@@ -238,21 +238,21 @@
 	</div>
 
 	<div class="settings-wrapper">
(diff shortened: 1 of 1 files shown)


freddyaboulton (approved):
Awesome @hannahblair !

PR #6781, opened 2023-12-13
Fix backend tests on Windows

abidlabs:
2 non-flaky tests were failing on Windows. Turns out one was a problem with the test, while the other was actually an issue in the code. Both should be fixed now.

### DIFF
--- a/.changeset/vast-kings-deny.md
+++ b/.changeset/vast-kings-deny.md
@@ -1,0 +1,5 @@
+---
+"gradio": minor
+---
+
+feat:Fix backend tests on Windows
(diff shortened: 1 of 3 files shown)


abidlabs:
Ok the `FileExplorer` test that is failing is kind of weird. It looks like there are actually significant differences in the postprocessed value of `gr.FileExplorer(...)` on Windows. I'm going to look into this later today on my Windows laptop to see what the issue is.

freddyaboulton (approved):
thanks @abidlabs !

abidlabs:
Thanks @freddyaboulton

PR #6787, opened 2023-12-14
allow custom component authors to provide custom vite plugins and svelte preprocessors

### LINKED ISSUE #7006: [Custom Components] Provide some configuration for building frontend

Some javascript libraries need additional processing to be used correctly, could gradio provide some configuration to do this?

**Additional context**  
I found this PR before: https://github.com/gradio-app/gradio/pull/6787, will there be more updates?

### LINKED ISSUE #7927: Support tailwind in custom components

Developers really want to use tailwind in their custom components. We should make it easy to integrate it into the custom component build process and add the necessary guides.

**Additional context**

### LINKED ISSUE #8044: Using SCSS for Gradio Custom Components

Currently, it seems like only pure CSS is available for creating Gradio custom components; CSS preprocessors are not supported.

I hope to be able to use CSS preprocessors like SCSS/SASS for creating Gradio custom components.

pngwn:
## Description

Closes #8044. Closes #7927. Closes #7006.

## 🎯 PRs Should Target Issues

## Tests


js/preview/src/build.ts
```diff
@@ -26,6 +26,22 @@ export async function make_build({
...
+				svelte: {
+					preprocess: []
+				}
+			};
+
+			if (
+				comp.frontend_dir &&
+				fs.existsSync(join(comp.frontend_dir, "gradio.config.js"))
+			) {
+				const m = await import(join(comp.frontend_dir, "gradio.config.js"));
```
pngwn, line 40: I went for `gradio.config.js` because that is a pretty solid convention in the JS world.

It has to be a `js` file rather than yaml/toml/json because these configs require you to run/ import code inside them. It also needs to be in the frontend dir because in the default case that is the dir with access to node modules.

This is a little complicated though because you could multiple frontend dirs. We can technically support a config per component for the build as each component build is a discreet process.

js/preview/src/dev.ts
```diff
@@ -109,13 +109,47 @@ function to_posix(_path: string): string {
...
 	const component_entries = components.flatMap((component) => {
 		return examine_module(component, root, "dev");
 	});
 
+	let component_config = {
+		plugins: [],
+		svelte: {
+			preprocess: []
+		}
+	};
```
pngwn, line 134: However for dev mode, this is one process for all components, so we can only use one of these config files.

Ideally we'd put this in the root but that isn't where we install the packages, so this won't really work (no node_modules).

pngwn: In hindsight, it might have been better to put the package.json in the root and install the node_modules there but this would be a breaking change now.

pngwn: What we could do is check for a config file in the frontend dirs, then recursivelky walk up the tree. And check each dir above. This is how most config files in the JS world work and how node_modules works.

This would give us something that works in the default case but allow people with more complex setups to add a single config file to a shared root. It would be on them to add a package.json in the root but this might be nice for more complex setups.

js/preview/src/build.ts
freddyaboulton: Should we create an empty `gradio.config.js` file during `gradio cc create`?

pngwn: Yeah I was thinking about that. Not sure. It's helpful but also adds noise.

pngwn:
We need to do something else.

pngwn:
To test this:

run `pnpm build` from the root.

Copy paste this code in to the `js/preview/test/test/frontend/gradio.config.js`:

```js
import tailwindcss from "@tailwindcss/vite";

export default {
	plugins: [tailwindcss()]
};
```

Add this file in the frontend dir:

`main.css`:

```css
@import "tailwindcss";
```

Add this code to `Index.svelte`:

```js
import './main.css';
```

Run this command in the `frontend` dir:

```bash
pnpm add -D tailwindcss@next @tailwindcss/vite@next
```



cd back into `test/test` and run:

 ```bash
 gradio cc install --npm-install="pnpm i"
 gradio cc dev
 ```

pngwn:
tbh i'll just push these changes

pngwn:
New instructions:

```
pnpm i
ppm build
cd js/preview/frontend/test/test
gradio cc install --npm-install="pnpm i"
gradio cc dev
```

freddyaboulton (approved):
Nice @pngwn ! Let's please write some docs for this in the frontend guide: https://www.gradio.app/guides/frontend

I still think we should provide an empty `gradio.config.js` file in the create command so that users know about it without having to use the guides. Maybe we include some commented out code like "to add tailwind, add the following"

pngwn:
Good idea. Will update.

pngwn:
@freddyaboulton I have done the following:

- we now generate a `gradio.config.js` as part of the `create` command. I haven't 'filled it in' atm. Its kinda noisy with all the comments so not sure if that is best, but don't feel strongly.
- Added a section of custom configuration to the frontend guide: https://gradio-q7zldhffx-hugging-face.vercel.app/main/guides/frontend#custom-configuration

guides/05_custom-components/05_frontend.md
```diff
@@ -256,6 +256,32 @@ For those interested in design customization, we provide the CSS variables consi
...
+
+Vite options:
+- `plugins`: A list of vite plugins to use.
+
+Svelte options:
+- `preprocess`: A list of svelte preprocessors to use.
+
+The `gradio.config.js` file should be placed in the root of your component's `frontend` directory. A default config file is created for you when you create a new component. But you can also create your own config file and use it to customize your component's build process.
+
+```typescript
```
freddyaboulton, line 273: Need to close this codeblock

image

freddyaboulton (commented):
Docs changes look good to me @pngwn !

PR #6794, opened 2023-12-14
Fix SSRF vulnerability on `/file=` route

abidlabs:
There's a potential risk of people using server-side request forgery using the `/file` route since it used to perform a GET/HEAD request to determine if a filepath was a possible URL.

This should mitigate that.


freddyaboulton (approved):
This looks good to me!

abidlabs:
Thanks!!

Issue #6713, opened 2023-12-08
Investigate whether Exceptions are being handled correctly by the new queue


abidlabs:
Expected behavior: when an exception happens in a user function, the worker thread should return the exception and then immediately terminate. A new request should be picked up from the queue and get processed. 

However, we've received a number of reports that suggest that this might not be happening -- exceptions might be clogging up the queue:

* [Internal: "It seems now multiple thread is running at the same time, messing around with their (shared?) variables"](https://huggingface.slack.com/archives/C068FP3MBDJ/p1701992804324409?thread_ts=1701990184.706129&cid=C068FP3MBDJ)
* [Internal: "GPU utilisation is... 0%, but the queue is max’d out at 10](https://huggingface.slack.com/archives/C05N2SUPJP5/p1702034997040039)
* [queue stuck reports](https://huggingface.co/spaces/haodongli/LucidDreamer/discussions/2)

PR #6798, opened 2023-12-14
Improve how server/js client handle unexpected errors

freddyaboulton:
## Description

Noticed two things when I was debugging flaky tests:
- If there's an exception in the `queue/data/` route the client will hang forever. the server now catches general exceptions
- If the message didn't have an `event_id`, the client would error out, e.g. heartbeat, server_stopped. This wouldn't cause any noticeable errors to users (except console exceptions) but wanted to fix.My guess is that the playwright tests are mysteriously erroring server side so printing the exception message in the UI can help us debug.
- `gr.Info` and `gr.Warning` would also raise exceptions in the client. Would not be obvious to users but not desirable anyways.
- Think I found the issue for the weird HTTPExceptions in logs

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/client/js/src/client.ts
+++ b/client/js/src/client.ts
@@ -1051,7 +1051,7 @@ export function api_factory(
 				event_stream = new EventSource(url);
 				event_stream.onmessage = async function (event) {
 					let _data = JSON.parse(event.data);
-					if (_data.session_hash) {
+					if (!("event_id" in _data)) {
 						await Promise.all(
 							Object.keys(event_callbacks).map((event_id) =>
 								event_callbacks[event_id](_data)
--- a/client/js/src/client.ts
+++ b/client/js/src/client.ts
@@ -931,7 +931,6 @@ export function api_factory(
 											time: new Date()
 										});
 										close_stream();
-										console.trace()
 									}
 								};
 								event_ids.push(event_id);
--- a/client/js/src/client.ts
+++ b/client/js/src/client.ts
@@ -289,7 +289,6 @@ export function api_factory(
 			let stream_open = false;
 			let event_stream: EventSource | null = null;
 			const event_callbacks: Record<string, () => Promise<void>> = {};
-			const event_ids = [];
 			let config: Config;
 			let api_map: Record<string, number> = {};
 
@@ -933,7 +932,6 @@ export function api_factory(
 										close_stream();
 									}
 								};
-								event_ids.push(event_id);
 								event_callbacks[event_id] = callback;
 								if (!stream_open) {
 									open_stream();
@@ -1055,7 +1053,9 @@ export function api_factory(
 					let _data = JSON.parse(event.data);
 					if (_data.session_hash) {
 						await Promise.all(
-							event_ids.map((event_id) => event_callbacks[event_id](_data))
+							Object.keys(event_callbacks).map((event_id) =>
+								event_callbacks[event_id](_data)
+							)
 						);
 						return;
 					}
--- a/client/js/src/client.ts
+++ b/client/js/src/client.ts
@@ -837,14 +837,6 @@ export function api_factory(
 											return;
 										}
 
-										// TODO: Find out how to print this information
-										// only during testing
-										// console.info("data", type, status, data);
-
-										if (type == "heartbeat") {
-											return;
-										}
-
 										if (type === "update" && status && !complete) {
 											// call 'status' listeners
 											fire_event({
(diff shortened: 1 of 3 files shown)


client/js/src/client.ts
```diff
@@ -821,84 +822,129 @@ export function api_factory(
...
+										) {
+											if (event_callbacks[event_id]) {
+												delete event_callbacks[event_id];
+												if (Object.keys(event_callbacks).length === 0) {
+													close_stream();
+												}
 											}
 										}
+									} catch (e) {
+										console.error("Unexpected client exception", e);
```
freddyaboulton, line 930: This should fix the `HTTPException` errors we keep seeing in the logs.

### Explanation
https://www.loom.com/share/5f7abfdb2f2143c58982f11dab2538b2?sid=d176fd1c-b5f5-43f5-be5f-f72af9b3abea

### Fix
https://www.loom.com/share/d6b9d87d05b64f69bfb8158ac565593f?sid=ca833ba1-4579-4536-8e14-50396ba3261c

freddyaboulton:
Not sure how to test this actually. Should be good for review though!

abidlabs:
Is this expected to resolve https://github.com/gradio-app/gradio/issues/6713?

freddyaboulton:
@abidlabs Hard to know for sure since we don't know how to repro those issues. However, I did just verify that exceptions in user functions don't break the queue/app

https://www.loom.com/share/0b7d99654b444b6a8ff6d0c320c878ed?sid=bb08ee79-eb06-48f0-822d-c89d0819a1db

gradio/routes.py
```diff
@@ -644,9 +650,17 @@ async def sse_stream(request: fastapi.Request):
                                     )
                                 ):
                                     return
-                except asyncio.CancelledError as e:
-                    del blocks._queue.pending_messages_per_session[session_hash]
-                    await blocks._queue.clean_events(session_hash=session_hash)
+                except Exception as e:
```
aliabid94, line 653: Has to be BaseException to catch everything - asyncio.CancelledError is not an instance of Exception

client/js/src/client.ts
```diff
@@ -821,84 +822,122 @@ export function api_factory(
...
+											endpoint: _endpoint,
+											fn_index,
+											time: new Date()
+										});
+										close_stream();
 									}
 								};
+								event_ids.push(event_id);
 								event_callbacks[event_id] = callback;
+								if (!stream_open) {
```
aliabid94, line 938: why did you move this?

client/js/src/client.ts
```diff
@@ -1014,6 +1053,12 @@ export function api_factory(
 				event_stream = new EventSource(url);
 				event_stream.onmessage = async function (event) {
 					let _data = JSON.parse(event.data);
+					if (_data.session_hash) {
+						await Promise.all(
+							event_ids.map((event_id) => event_callbacks[event_id](_data))
```
aliabid94, line 1058: If this is the only reason we're keeping track of event_ids, we already have that list in the keys of event_callbacks. No need to create a second list right?

client/js/src/client.ts
freddyaboulton: Figured the stream should be opened after the callback got added

aliabid94:
> If there's an exception in the queue/data/ route the client will hang forever. the server now catches general exceptions

What exceptions were happening?

client/js/src/client.ts
aliabid94: ok sure, but if a second request is sent while a first one is pending, the stream will be open from the first request, so we shouldn't be expecting it to be the case that the stream is receiving messages for an event only after post request is complete. 
Therefore, it can technically happen that the event stream starts sending messages for an event before the POST request to upload data finishes, and therefore the event callback to handle the message is not ready, which is a race condition that I do not know if we are hitting. I think I do resolve this in a PR I'm working on though

freddyaboulton:
> What exceptions were happening?

We don't know because the client would hang forever. Although I think it's the reason the playwright tests were just hanging forever in the cases they failed.

client/js/src/client.ts
aliabid94: also would prefer if the logic for targetting every event listener was more explicit than if the session_hash key is present. We don't use the session_hash value anyway. I think it'd make more sense that if an event_id is not provided, we trigger every event.

freddyaboulton: Done!

freddyaboulton: Also made the change about checking the absence of event_id to trigger all listeners!

aliabid94 (approved):
LGTM!

freddyaboulton:
Thanks @aliabid94 !

Issue #6784, opened 2023-12-14
ImageEditor(value=..) does not work


stefanengelmann:
### Describe the bug

The value parameter of the ImageEditor component does not work, but for the Image component it works fine. I need the value parameter such that I can upload file formats, that the ImageEditor component does not support such as TIFF, convert them to NumPy arrays, and pass them as a value to the ImageEditor component

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr
from PIL import Image
import numpy as np
import wget

# Example image
wget.download('https://people.sc.fsu.edu/~jburkardt/data/png/dragon.png')
im_frame = Image.open('dragon.png')
im = np.array(im_frame)

with gr.Blocks() as demo:
    gr.Image(value=im) # Does not work
    gr.ImageEditor(value=im) # Works

demo.launch()

```

### Logs

```shell
---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
Cell In[14], line 12
     10 with gr.Blocks() as demo:
     11     gr.Image(value=im)
---> 12     gr.ImageEditor(value=im)
     14 demo.launch()

File ~/.conda/envs/qim/lib/python3.10/site-packages/gradio/component_meta.py:155, in updateable..wrapper(*args, **kwargs)
    153     return None
    154 else:
--> 155     return fn(self, **kwargs)

File ~/.conda/envs/qim/lib/python3.10/site-packages/gradio/components/image_editor.py:180, in ImageEditor.__init__(self, value, height, width, image_mode, sources, type, label, every, show_label, show_download_button, container, scale, min_width, interactive, visible, elem_id, elem_classes, render, mirror_webcam, show_share_button, _selectable, crop_size, transforms, eraser, brush)
    177 self.eraser = eraser if eraser is not None and eraser is not True else Eraser()
    178 self.brush = brush if brush is not None and brush is not True else Brush()
--> 180 super().__init__(
    181     label=label,
    182     every=every,
    183     show_label=show_label,
    184     container=container,
    185     scale=scale,
    186     min_width=min_width,
    187     interactive=interactive,
    188     visible=visible,
    189     elem_id=elem_id,
    190     elem_classes=elem_classes,
    191     render=render,
    192     value=value,
    193 )

File ~/.conda/envs/qim/lib/python3.10/site-packages/gradio/component_meta.py:155, in updateable..wrapper(*args, **kwargs)
    153     return None
    154 else:
[...]

Issue #6634, opened 2023-12-03
Document helper classes for ImageEditor and Audio 


aliabd:
May need to add more to this list but these helper classes need to be documented: 
- WaveformOptions in Audio 
- Brush and Erasor in ImageEditor

They don't have docstrings. 

This can be converted to a tracking issue if that's better. 

Related: https://github.com/gradio-app/gradio/issues/6633

Issue #6338, opened 2023-11-08
Describe the documentation of `waveform_options` of `gr.Audio` in more detail


hysts:
- [x] I have searched to see if a similar issue already exists.

It'd be nice if `waveform_options` of `gr.Audio` is documented in more detail.

`waveform_color` and `waveform_progress_color` are obvious, so it's OK, but I feel the documentation of `show_controls` and `skip_length` is insufficient.

I tried setting `show_controls`, but I'm not sure what this is for. Apparently, the default value is `False`, but it appears a bit weird to me if set `True`. I wonder what's the expected use case of this option.

![](https://github.com/gradio-app/gradio/assets/25161192/59264c72-563f-4e6f-9bbd-49711918613e)

Also, I don't know what `skip_length` does. I tried some values but couldn't figure out what was changed.

NewEpoch2020:
I have the same confusion

thiswillbeyourgithub:
I have the same confusion and would like to add that I really miss the feature to disable edit in audio. In version prio to 4.0 we could use `show_edit_button=False` per #5149

All those buttons are cluttering my screen (I use multiple audio components)

PR #6799, opened 2023-12-14
Adds docstrings for `gr.WaveformOptions`, `gr.Brush`, and `gr.Eraser`, fixes examples for `ImageEditor`, and allows individual images to be used as the initial `value` for `ImageEditor`

abidlabs:
Other minor refactoring as well.

Closes: https://github.com/gradio-app/gradio/issues/6338
Closes: https://github.com/gradio-app/gradio/issues/6634

Note that this doesn't actually surface the documentation (that's a separate issue: #6633)

Edit: thanks to @pngwn's suggestion, now:

Closes: #6784
Closes: https://github.com/gradio-app/gradio/issues/6752

### DIFF
--- a/demo/image_editor/run.py
+++ b/demo/image_editor/run.py
@@ -11,6 +11,7 @@ with gr.Blocks() as demo:
     im = gr.ImageEditor(
         type="pil",
         crop_size="1:1",
+        eraser=gr.Eraser()
     )
 
     with gr.Group():
--- a/demo/image_editor/run.py
+++ b/demo/image_editor/run.py
@@ -11,7 +11,6 @@ with gr.Blocks() as demo:
     im = gr.ImageEditor(
         type="pil",
         crop_size="1:1",
-        eraser=gr.Eraser()
     )
 
     with gr.Group():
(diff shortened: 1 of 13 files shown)


pngwn (approved):
Looks good to me!

The value type is currently incorrect though (#6784). It should be set to the types duct defined above (or None). Could we fixed that here too?

abidlabs:
Fixed the `value` and also fixed examples for `ImageEditor`, so now you can pass in a dict to use as the examples. It will render as just a regular image. It will render the composite image if available, and if not, will fall back to use the background image.

abidlabs:
@hannahblair just fyi I did notice this issue while testing out the waveform options: https://github.com/gradio-app/gradio/issues/6801

pngwn:
I don't know that I agree with the post process change. The main reason for splitting Image and InageEditor was the consistency of the signatures and this introduces some inconsistency. 

That said preprocessor is still consistent, so it might not be that bad. And it is is consistent regardless of type which was the main issue with the previous Image.

pngwn (approved):
I'm convinced! Thansk for taking care of this @abidlabs!

abidlabs:
Thanks @hannahblair and @pngwn for reviewing!

Issue #6341, opened 2023-11-08
Running Gradio using Nginx HTTPS in a Web Server


MarkIvanDev:
### Describe the bug

We already have a guide for serving gradio apps on a web serving using nginx here: https://www.gradio.app/guides/running-gradio-on-your-web-server-with-nginx. But converting this to use https leads to problems. The gradio app (in my case, chat) does not stop getting of the queueing stage and when I inspect the developer tools, it seems that when making a post to the /queue/data path, it results in an Internal Server Error.

I am using this nginx configuration to enable https:

```
server {
    server_name <SERVER_NAME>;

    location / {
        proxy_pass              http://localhost:7860;
        proxy_http_version      1.1;
        proxy_redirect          off;
        proxy_set_header        Upgrade $http_upgrade;
        proxy_set_header        Connection "upgrade";
        proxy_set_header        Host $host;
    }

    listen 443 ssl; # managed by Certbot
    ssl_certificate /etc/letsencrypt/live/<domain>/fullchain.pem; # managed by Certbot
    ssl_certificate_key /etc/letsencrypt/live/<domain>/privkey.pem; # managed by Certbot
    include /etc/letsencrypt/options-ssl-nginx.conf; # managed by Certbot
    ssl_dhparam /etc/letsencrypt/ssl-dhparams.pem; # managed by Certbot

}

server {
    if ($host = <DOMAIN>) {
        return 301 https://$host$request_uri;
    } # managed by Certbot

    server_name <DOMAIN>;
    listen      80;
    return 404; # managed by Certbot
}
```

And I am using this nginx configuration for http. The gradio app works perfectly when using this, but I can only use an https so I need to convert this into https if possible:

```
server {
    server_name <SERVER_NAME>;
    listen 80;

    location / {
        proxy_pass              http://localhost:7860;
        proxy_http_version      1.1;
        proxy_redirect          off;
        proxy_set_header        Upgrade $http_upgrade;
        proxy_set_header        Connection "upgrade";
        proxy_set_header        Host $host;
    }
}
```

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

Use the nginx configuration as above and restart the nginx service and use the https version of the gradio app site.

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Linux
gradio version: 4.1.1
gradio_client version: 0.7.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
[...]

PR #6804, opened 2023-12-15
Update running-gradio-on-your-web-server-with-nginx.md

lfuelling:
## Description

This PR adds the `proxy_buffering off;` parameter to the example NGINX config.

Closes: #6341

## Tests

I've only changed a guide, no tests should be necessary for this.

### DIFF
--- a/guides/09_other-tutorials/running-gradio-on-your-web-server-with-nginx.md
+++ b/guides/09_other-tutorials/running-gradio-on-your-web-server-with-nginx.md
@@ -36,6 +36,7 @@ server {
 
     location /gradio-demo/ {  # Change this if you'd like to server your Gradio app on a different path
         proxy_pass http://127.0.0.1:7860/; # Change this if your Gradio app will be running on a different port
+        proxy_buffering off;
         proxy_redirect off;
         proxy_http_version 1.1;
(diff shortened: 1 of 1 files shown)


abidlabs:
Thanks @lfuelling we can update this. Could you please just confirm that you've tested this in the latest version of Gradio (4.9.1) and this configuration works for you?

abidlabs:
Just wanted to check in @lfuelling on the comment above?

lfuelling:
@abidlabs sorry for the slow response.

The `proxy_buffering off;` directive was required to get it to work at all in my case, previously requests would fail in a weird way.

abidlabs:
Thanks makes sense, which version of Gradio were you running?

lfuelling:
I am using Gradio 4.4.1

abidlabs:
Ok great, I'll go ahead and merge this in then, as Gradio 4.4.1 uses the new SSE approach in favor of websockets. Thanks for the fix!

lfuelling:
Ah that makes sense, I read something about websockes and was massively confused when I only saw "regular" HTTP :)

abidlabs:
Ah that's probably outdated documentation, working on fixing that!

Issue #6772, opened 2023-12-13
Customize `gr.LoginButton()` text


apolinario:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
Currently you can't customize the text in `gr.LoginButton()`

**Describe the solution you'd like**  
I'd like to customize the text in `gr.LoginButton()` for specific call to actions. Example use-case
<img width="765" alt="image" src="https://github.com/gradio-app/gradio/assets/788417/5cd0c44c-8515-4879-915c-f830377a88c8">

**Alternatives**  
[There's a way to do it that is hacky](https://huggingface.co/spaces/multimodalart/civitai-to-hf/blob/main/app.py#L366) with CSS, by leveraging `#login:after` but would be great to not need such hacky approaches

abidlabs:
Thanks @apolinario this is a bug as the `value` parameter is supposed to be the text that is displayed. However, it doesn't seem to have any effect at the moment. 

I assume you also want to be able to customize the "signed in" text as well? We could add a `signed_in_value` parameter?

apolinario:
Yes, I also have to hack to customize the signed in button text

PR #6808, opened 2023-12-15
Ensure LoginButton `value` text is displayed

hannahblair:
## Description

When initialising a `LoginButton`, we were hardcoding the button with the `value` "Sign in with Hugging Face". This changes that, so that `self.value` is used instead, which defaults to `Sign in with Hugging Face` if not provided. 

Closes: #6772

Test:

```
import gradio as gr 

with gr.Blocks() as demo:
    gr.LoginButton(value="boop")

demo.launch()
```

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/full-gifts-press.md
+++ b/.changeset/full-gifts-press.md
@@ -1,0 +1,7 @@
+---
+"@gradio/audio": patch
+"@gradio/imageeditor": patch
+"gradio": patch
+---
+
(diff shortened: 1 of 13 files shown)


abidlabs:
This looks good @hannahblair -- could we also add a `signed_in_value` parameter as mentioned in: https://github.com/gradio-app/gradio/issues/6772#issuecomment-1854480825

abidlabs:
LGTM thanks @hannahblair for taking care of this!

abidlabs (approved):

PR #6810, opened 2023-12-15
Fix `gr.load()` so that it works with the SSE v1 protocol

abidlabs:
Stupidly simple fix 🥲 

Fixes: https://github.com/gradio-app/gradio/issues/5927

### DIFF
--- a/gradio/external.py
+++ b/gradio/external.py
@@ -538,9 +538,7 @@ def from_spaces_blocks(space: str, hf_token: str | None) -> Blocks:
     predict_fns = []
     for fn_index, endpoint in enumerate(client.endpoints):
         assert isinstance(endpoint, Endpoint)
-        helper = None
-        if endpoint.protocol in ("ws", "sse", "sse_v1"):
-            helper = client.new_helper(fn_index)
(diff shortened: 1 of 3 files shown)


gradio/external.py
```diff
@@ -538,9 +538,7 @@ def from_spaces_blocks(space: str, hf_token: str | None) -> Blocks:
     predict_fns = []
     for fn_index, endpoint in enumerate(client.endpoints):
         assert isinstance(endpoint, Endpoint)
-        helper = None
-        if endpoint.protocol in ("ws", "sse"):
-            helper = client.new_helper(fn_index)
+        helper = client.new_helper(fn_index)
```
abidlabs, line 541: Initially I had added "sse_v1" to the list but actually I don't think there's any reason to have this check because `gr.load()` only works with `gradio==4.x` which includes the protocol anymore

freddyaboulton: I think this makes sense

freddyaboulton (approved):
Thanks @abidlabs !

abidlabs:
Thanks @freddyaboulton for the quick review!

Issue #6651, opened 2023-12-04
Queue Issues: Estimations, Freezing, Losing position


aliabid94:
Issue for several bugs related to the queue, including:

1. Estimations are incorrect
2. Some queues are freezing in place
3. Some sessions are losing their position or going backward in the queue

PR #6814, opened 2023-12-16
Refactor queue so that there are separate queues for each concurrency id

aliabid94:
I made some sweeping changes how queues are handled. Previously, we had a single queue that kept track of all events as they came in. However, with the introduction of concurrency counts, this no longer made sense. Imagine a queue with events of type A and B, each with concurrency limit = 1, queued as such:

Queue: A1, A2, B1, B2, A3, A4, B3

Previously, we would say that the B3 at the end was at position 7. However, there are not 6 events that must be complete before the final B can be processed - there are only 2 B's that must be processed. B3 is actually at queue position 3. After this refactor, we have one queue per concurrency id, as such:

Queue A: A1, A2, A3, A4
Queue B: B1, B2, B3

This intuitively tracks how events are actually processed, and makes it much to see if an event needs to be popped off the queue, where an event is in its queue, and how to calculate the estimated time for an event in queue.

This also fixes estimation calculations, fixes: #6651 

Made other fixes to the queue as well, such as:
 
Resolving race conditions that may occur as a result of event messages in the frontend being received the front end before the POST request to upload the data is complete. Imagine we fire two events, first A, then B. This is the race condition that could happen:

1. Event A submits the POST request with data
2. Event A is added to the queue in the backend
3. POST request for A completes, event_id for A is returned to client
4. Callback listener set up for event_id A in the frontend
5. Stream is opened on client for SSE updates that will trigger the relevant callback
6. Event B submits the POST request
7. Event B is added to the queue in the backend
8. Event B sends event message (an estimation, or even process completion if it is very fast)
9. Existing stream on frontend receives event message for Event B
10. POST request for B completes, event_id for B is returned to client
11. Callback listener set up for event_id B in the frontend

The race condition is that 9 happens before 10 and 11. Since the stream has already been open from event A (step 5), it was able to receive updates about event B (atep 9) before the POST request for B even complete and the callback listener was set up for it (step 11)


abidlabs:
Nice @aliabid94!! Is there a risk that events with concurrency_limit=None could hog all of the worker threads? Or how do you ensure that workers go to all of the different queues?

gradio/blocks.py
```diff
@@ -376,7 +376,7 @@ def __init__(
         self.preprocess = preprocess
         self.postprocess = postprocess
         self.tracks_progress = tracks_progress
-        self.concurrency_limit = concurrency_limit
+        self.concurrency_limit: int | None | Literal["default"] = concurrency_limit
```
abidlabs, line 379: just wondering why you added the type hint here? its already specified in the method signature

client/js/src/client.ts
```diff
@@ -932,6 +933,12 @@ export function api_factory(
 										close_stream();
 									}
 								};
+								if (event_id in pending_stream_messages) {
```
abidlabs, line 936: What will happen if `event_id` is not in `pending_stream_messages`? Will the client send another SSE message?

gradio/queueing.py
```diff
@@ -345,81 +368,70 @@ async def clean_events(
...
-        estimation.rank = rank
-
-        if self.avg_concurrent_process_time is not None:
-            estimation.rank_eta = (
-                estimation.rank * self.avg_concurrent_process_time
-                + self.avg_process_time
+    def broadcast_estimations(
+        self, concurrency_id: str, after: int | None = None
+    ) -> None:
+        wait_so_far = 0
```
abidlabs, line 396: Can you explain what `wait_so_far` is measuring and why it can be `None`? Why is it added to `process_time_for_fn`?

gradio/queueing.py
```diff
@@ -345,81 +368,70 @@ async def clean_events(
...
+                if process_time_for_fn is not None and wait_so_far is not None
+                else None
+            )
+            if (
+                rank_eta is not None
+                and event_queue.current_concurrency == event_queue.concurrency_limit
+                and process_time_for_fn is not None
+                and event_queue.concurrency_limit is not None
+            ):
+                rank_eta += process_time_for_fn / event_queue.concurrency_limit / 2
```
abidlabs, line 415: What is this doing?

abidlabs:
Confused about a few of the changes, please see qs above. In addition, I had some concerns about two somewhat related issues:

(1) How do we ensure that the workers take tasks from each queue, rather than just taking off tasks from a single queue? Imagine you have two tasks: A (with `concurrency_limit = None`) and B (with `concurrency_limit_1`). If there are a large number of tasks in each queue (e.g. a viral app), what would prevent the workers from always picking tasks from the frist queue? This would lead to a frustrating experience for users who are waiting on task B.

Suggestion: when a worker is free, it should take a task from any of the queues that hasn't hit its concurrency limit

(2) Are the estimations correct in the case where the number of different concurrency ids exceeds the number of workers (e.g. an app like automatic1111 or other apps with lots of functions). In that case, you need to account for the fact that first, you need to have an available worker which can then pick up a task from the current queue. In other words, you can't just treat all the queues as completely independent.

Suggestion: add a `time_until_worker` variable, which looks across all of the queues to estimate when a worker will be free and will arrive at the current queue.

gradio/blocks.py
aliabid94: was getting type errors elsewhere because `self.concurrency_limit` was being treated as `int | None | str` otherwise

client/js/src/client.ts
aliabid94: pending_stream_messages is only the list of events that beat the race condition specified in the PR (event messages received before the POST request completed). In most cases there are no pending_stream_messages, in which case there is nothing to do.

gradio/queueing.py
aliabid94: wait_so_far is the running estimated wait time. As we iterate through the events in the queue, we add the expected runtime of the each event It gets set to None if we encounter an event who's runtime we do not know (first time running this particular fn). In this case, this event and all events behind in queue will not recieve an estimate ETA.

gradio/queueing.py
aliabid94: If there are no available free workers due to hitting the concurrency limit, the expected amount of time we have to wait for a worker to free up is the expected runtime of the function divided by the concurrency limit divided by 2 (we divide by two assuming that any worker is half way through running the process)

aliabid94:
> (1) How do we ensure that the workers take tasks from each queue, rather than just taking off tasks from a single queue? Imagine you have two tasks: A (with concurrency_limit = None) and B (with concurrency_limit_1). If there are a large number of tasks in each queue (e.g. a viral app), what would prevent the workers from always picking tasks from the frist queue? This would lead to a frustrating experience for users who are waiting on task B.

I think a better approach would be to randomize the order of event queues we check. Otherwise its still possible that a single event queue dominates. WIll implement.

> (2) Are the estimations correct in the case where the number of different concurrency ids exceeds the number of workers (e.g. an app like automatic1111 or other apps with lots of functions). In that case, you need to account for the fact that first, you need to have an available worker which can then pick up a task from the current queue. In other words, you can't just treat all the queues as completely independent.

Accounting for number of workers is actually very complicated. To calculate `time_until_worker`, you'd have to look at all pending tasks and their expected runtimes, as well as every other queue since you are competing with them for workers. Not only that, but as new tasks get added to queues, they would potentially affect the estimated runtimes of all other events as it affects the competition for workers. I actually think this logic is so complex that it should be in another PR.

abidlabs:
Sounds reasonable to me. I'm working on something else atm but will circle back to review this later again

gradio/blocks.py
freddyaboulton: Yea pyright doesn't seem remember about Literals if you access a property from a method

abidlabs:
Thanks @aliabid94 for the explanations. I'd suggest including them as inline comments to make the code more understandable. 

So I tested out the new queue and generally speaking, its working quite well. The only issue I can find is this that events that share a `concurrency_id` are not handled correctly when computing estimations. 

Take this example:

```py
import gradio as gr
import time

def slow(x):
    def sleep_identity(y):
        time.sleep(x)
        return x
    return sleep_identity

with gr.Blocks() as demo:
    with gr.Row():
        t1 = gr.Textbox()
        t2 = gr.Textbox()        
    with gr.Row():
        t5 = gr.Textbox()
        t6 = gr.Textbox()
    
    t1.submit(slow(5), t1, t2, concurrency_id=123)
    t5.submit(slow(15), t5, t6, concurrency_id=123)
    
demo.launch()
```

Run each event once. Then run the bottom (slower) event, and then immediately run the top (faster) event. You'll see that the estimated time is 7.5 seconds -- this is incorrect, it should be 5 + 15/2 = 12.5 seconds. The problem, I believe, is that when computing the expected time, it uses the current function's expected time as the expected time for all function that are in progress, even if the function is different (just shares the same concurrency_id).

gradio/queueing.py
```diff
@@ -143,32 +174,45 @@ def _resolve_concurrency_limit(self, default_concurrency_limit):
         else:
             return 1
 
+    def __len__(self):
+        return sum(
```
freddyaboulton, line 178: nit: Don't think we need to create a list in memory just to throw it out immediately. A for-loop where we `+=` an variable might be easier to understand.

freddyaboulton (approved):
Left some nits beyond @abidlabs current comments. Otherwise looks good to me! Thanks @aliabid94 !

gradio/queueing.py
```diff
@@ -345,81 +372,70 @@ async def clean_events(
                     if job.session_hash == session_hash or job._id == event_id:
                         job.alive = False
 
-        events_to_remove = []
-        for event in self.event_queue:
-            if event.session_hash == session_hash or event._id == event_id:
-                events_to_remove.append(event)
+        async with self.delete_lock:
+            events_to_remove: list[Event] = []
```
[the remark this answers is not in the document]
aliabid94: I didn't want this to be running at the same time as `self.get_events`, because it may be that the first loop flags an event for removal, and then self.get_events pops the event off the queue into the active jobs, and then the second loop runs which actually removes the events from the queue, and it doesn't find the event and errors.

Could also do a "if event in queue: queue,remove(event)" check, if you think that's better.

aliabid94:
> So I tested out the new queue and generally speaking, its working quite well. The only issue I can find is this that events that share a concurrency_id are not handled correctly when computing estimations.

Revamped this logic, now we keep track of all active jobs and when they were started. This way we can be very precise on the ETA of pending jobs. Should be ready for re-review

aliabid94:
actually an issue with the frontend printing, one sec

aliabid94:
ok ready for re-review

abidlabs (approved):
Stress tested with a lot of different setups and the etas were remarkably accurate. Good stuff @aliabid94!

Issue #3801, opened 2023-04-10
Allow users to set upload limits 


abidlabs:
Let users set upload limits to prevent too many large files from being uploaded to the server. This is especially relevant as we release the client as programmatically uploaded files could take up a lot of space on the servers. 

This could be a `max_file_size` parameter to `launch()`

vagxrth:
Hey, @abidlabs I would like to contribute to this issue, but I request a brief description beforehand. What I conclude is the parameter will act as a configurable setting which will enable the user to specify the maximum allowable size for the uploaded files.

vagxrth:
Thank you for the explanation, can I work on the issue?

PinkDraconian:
Hi @abidlabs,

You referenced this issue in my report on Huntr. Could you please review the report there, as it is not the same issue as referenced here. Feel free to ask any further questions on the report to clear up the confusion.

abidlabs:
Closing in favor of https://github.com/gradio-app/gradio/issues/7825

PR #6830, opened 2023-12-18
Set up File Upload Limit

vagxrth:
## Description
Added max_file_size parameter in the launch function that will prevent the user from uploading too many large files on the server.

Closes: https://github.com/gradio-app/gradio/issues/3801


gradio/blocks.py
```diff
@@ -1719,6 +1720,7 @@ def validate_queue_settings(self):
 
     def launch(
         self,
+        upload_file: FileStorage,
```
abidlabs, line 1723: Why is there an `upload_file` parameter?

abidlabs:
Hi @vagxrth thanks for opening this PR. I don't quite follow the logic here. The additional `max_file_size` argument is good, but I don't follow the other changes in `launch()`.

I believe the implementation of this parameter should happen here: https://github.com/gradio-app/gradio/blob/245d58eff788e8d44a59d37a2d9b26d0f08a62b4/gradio/routes.py#L753

vagxrth:
Hey @abidlabs I've gained insights from the review of my PR and I'll proceed to work on it keeping all the points in mind.

PR #6833, opened 2023-12-18
Prevent file traversals

abidlabs:
[Internal thread](https://huggingface.slack.com/archives/C02QZLG8GMN/p1702901295299349)


freddyaboulton (approved):
Thanks @abidlabs !

freddyaboulton (commented):
Would be nice to write a unit test for this!

abidlabs:
I tried to write a unit test but for some reason I can't reproduce this behavior with fastapi's test client. Only curl seems to have this effect (and only if you pass in `--path-as-is` flag)

abidlabs:
Apparently due to this PR https://github.com/urllib3/urllib3/pull/1487. Let me see if I can write a unit test a different way

abidlabs:
Ok added a quick unit test. Thanks for the review @freddyaboulton! Will merge after CI is green

PR #6842, opened 2023-12-19
Fix md highlight

pngwn:
## Description

some weird issues as we were reusing the same marked instance across multiple instances of `gr.Markdown`. The issues only seems to present when more than on markdown component is used.

https://marked.js.org/using_advanced#instance

The quickstart for marked is idiotic in the extreme but this should fix it.

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/js/markdown/shared/utils.ts
+++ b/js/markdown/shared/utils.ts
@@ -1,4 +1,4 @@
-import { marked, type Renderer } from "marked";
+import { type Renderer, Marked } from "marked";
 import { markedHighlight } from "marked-highlight";
 import { gfmHeadingId } from "marked-gfm-heading-id";
 import Prism from "prismjs";
@@ -75,9 +75,7 @@ const renderer: Partial<Omit<Renderer, "constructor" | "options">> = {
 		escaped: boolean
 	) {
 		const lang = (infostring ?? "").match(/\S*/)?.[0] ?? "";
-
 		code = code.replace(/\n$/, "") + "\n";
-
 		if (!lang) {
 			return (
 				'<div class="code_wrap">' +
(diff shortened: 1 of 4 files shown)


js/markdown/shared/utils.ts
```diff
@@ -128,28 +127,26 @@ export function create_marked({
 	);
 
 	if (header_links) {
-		if (header_links) {
```
pngwn, line 131: dno why we had this double if

js/markdown/shared/utils.ts
```diff
@@ -110,6 +107,8 @@ export function create_marked({
 	header_links: boolean;
 	line_breaks: boolean;
 }): typeof marked {
+	const marked = new Marked();
```
pngwn, line 110: Fresh instance per 'create_marked`.

js/markdown/shared/utils.ts
```diff
@@ -128,28 +127,26 @@ export function create_marked({
...
+						const id = "h" + slugger.slug(raw);
+						const level = token.depth;
+						const text = this.parser.parseInline(token.tokens!);
+
+						return `<h${level} id="${id}"><a class="md-header-anchor" href="#${id}">${LINK_ICON_CODE}</a>${text}</h${level}>\n`;
 					}
-				]
-			});
-		}
+				}
```
pngwn, line 147: since these plugins were being applied _again_ for every marked instance, the highlight code was getting passed through another highlighter. I imagine if we added more instances we would get increasingly weird results.

pngwn (commented):
We were basically adding _all_ plugins to the same renderer for every instance of the component. So 2 instances would be 2 highlighters. And i think that basically meant that the highlighted code was being passed through the highlighter again (but at that point it was html code).

abidlabs (approved):
LGTM thanks for the fix @pngwn!

Issue #6744, opened 2023-12-11
A Space loaded with `gr.load()` is not functional


vitaliy-sharandin:
### Describe the bug

While embedding the private space into the public one, UI is transitioned, but is completely unfunctional, although the private space by itself works perfectly. Dropdowns can't be clicked, buttons don't do anything, one can only fill a text field successfully. Also, no error log is shown.
I am embedding space as I currently don't want to share my code, hence private space load from public space.

### Have you searched existing issues?  🔎

- [x] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr
import os

HF_TOKEN = os.environ["HF_TOKEN"]

css = """
.column-frame {
    border: 2px solid #AAA;
    border-radius: 10px;
    padding: 10px;
    margin: 10px;
}
"""

with gr.Blocks(theme=gr.themes.Soft(),css=css) as demo:
    gr.load("private_space_path", src="spaces", hf_token=HF_TOKEN)

demo.launch(debug=True)
```

### System Info

```shell
gradio=4.8.0
env=Hugging Face Spaces
```

### Severity

Blocking usage of gradio

abidlabs:
Hi @vitaliy-sharandin do you see any errors in the logs?

vitaliy-sharandin:
As I've written before, no error logs are shown, it just silently doesn't work, there are dropdowns/video upload/buttons failing and from what I've seen in the HF forums, some users experienced troubles with data in loading spaces as well. Again, when private one works fine. Here is my public space, you can see for yourself.

I've also tried switching my public space to GPU configuration as private one requires it, still no success. Should this be done at all or it's ok to use CPU config in public space while private is GPU dependent?

https://huggingface.co/spaces/vitaliy-sharandin/Video_translation_with_speaker_diarization_and_voice_cloning

abidlabs:
Thanks @vitaliy-sharandin. The public Space does not need a GPU, it calls the private one via API. I'll take a look!

vitaliy-sharandin:
Thanks, @abidlabs, also, please let me know in this thread when you finish looking at project so that I turn off the private space, as GPU is eating money away.

abidlabs:
Oh please don't keep the GPU running. I would at least set the sleep time to a small amount, like 5 minutes (you can do this in the settings).

vitaliy-sharandin:
[...]

Issue #6476, opened 2023-11-18
Errors in Space loading another Space


dwipper:
### Describe the bug

Getting the below error when loading loading one Space in another Space. These Spaces are on 4.4.0.

The app runs just fine in the target space......

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

https://huggingface.co/spaces/cimsai/Compliance

```
import gradio as gr
nili = gr.load(src="spaces",name="dwipper/Compliance_Test")
nili.launch()
```

```
import os
import gradio as gr
import requests

def login_auth(username, password):

    if username == password:

        return {login_row: gr.Row(visible=False), app_row: gr.Row(visible=True)}
    else:
        
        return {login_row: gr.Row(visible=True), app_row: gr.Row(visible=False)}

with gr.Blocks() as iface:
    with gr.Row(visible=False) as app_row:
        with gr.Column():
            with gr.Row():
                with gr.Column():
                    gr.Markdown(value="<H2 style='text-align: center;'>NILI Compliance Desktop</h2>")
    with gr.Row(visible=True) as login_row:
        with gr.Column():
            gr.Markdown(value="<H2 style='text-align: center;'>NILI Login</h2>")
            with gr.Row():
                with gr.Column(scale=2):
                    gr.Markdown("")
                with gr.Column(scale=1, variant='panel'):
                    username_tbox = gr.Textbox(label="User Name", interactive=True)
                    password_tbox = gr.Textbox(label="Password", interactive=True, type='password')
                    submit_btn = gr.Button(value='Submit', variant='primary', size='sm')
                    submit_btn.click(login_auth, inputs=[username_tbox, password_tbox], outputs=[login_row,app_row])
                with gr.Column(scale=2):
                    gr.Markdown("")
    with gr.Row():
        with gr.Column(scale=4):
            gr.HTML('<center><i>© 2023 Collegiate Influencer Marketing Systems, Inc.</i><br>CIMS.AI, CIMS.AI logo, NILI, NILI logo, and EzNIL are trademarks of Collegiate Influencer Marketing Systems, Inc.</center>')

iface.queue()
iface.launch(show_api=False)
```

### Logs

```shell
===== Application Startup at 2023-11-17 23:51:28 =====

Fetching Space from: https://huggingface.co/spaces/dwipper/Compliance_Test
Loaded as API: https://dwipper-compliance-test.hf.space ✔
Running on local URL:  http://0.0.0.0:7860

[...]

Issue #4894, opened 2023-07-12
API page of Spaces doesn't show properly when there's a method with `api_name=''`


hysts:
### Describe the bug

The API page of [this Space](https://huggingface.co/spaces/hysts-debug/api-page-with-empty-api-name) is like this:
![](https://github.com/gradio-app/gradio/assets/25161192/b7235258-3540-4137-b875-f1987e7ce292)
code:
```python
import gradio as gr

with gr.Blocks() as demo:
    text = gr.Text()
    out = gr.Text()
    btn = gr.Button()
    text.submit(fn=lambda x: x, inputs=text, outputs=out, api_name='')
    btn.click(fn=lambda x: x, inputs=text, outputs=out, api_name='run')
demo.queue().launch()
```
The API page is empty and we can't close it.

Also, the API page of [this Space](https://huggingface.co/spaces/hysts-debug/api-page-with-empty-api-name-2) is like this:
![](https://github.com/gradio-app/gradio/assets/25161192/f4d5da11-74a9-4ffe-b70f-27cd9e624ff0)

code:
```python
import gradio as gr

with gr.Blocks() as demo:
    text = gr.Text()
    out = gr.Text()
    btn = gr.Button()
    text.submit(fn=lambda x: x, inputs=text, outputs=out, api_name='')
demo.queue().launch()
```
It says there is one API endpoint, but nothing is shown.

related: https://github.com/gradio-app/gradio/pull/4886#discussion_r1260570816

IMO, empty string should be invalid for `api_name`, and it should raise error when one sets it. But maybe the API page can be fixed instead. 

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

https://huggingface.co/spaces/hysts-debug/api-page-with-empty-api-name
https://huggingface.co/spaces/hysts-debug/api-page-with-empty-api-name-2

### System Info

```shell
gradio==3.36.1
```

### Severity

I can work around it

PR #6846, opened 2023-12-20
Add `show_api` parameter to events, and fix `gr.load()`. Also makes some minor improvements to the "view API" page when running on Spaces

abidlabs:
This PR does several things

1. It adds a `show_api` parameter to each event trigger. This `show_api` parameter controls whether or not the API route is visible in the API docs. Specifically, we set to `False` for internal routes in `gr.Interface()` and `gr.ChatInterface()` instead of setting `api_name=False`, which was causing issues using `gr.load()` with `gr.Interface()` and `gr.ChatInterface()`.

Fixes: #6744
Fixes: #6476 

2. Disallows the `api_name=""`, which was causing issues. If `api_name=""` is passed in, it is treated the same as `api_name=None`, and the function name is used instead. 

Fixes: #4894 

3. Also makes some minor improvements to the View API docs pages. For example, "Named endpoints" is replaced with "endpoints", endpoints are counted correctly, and the whitespace is fixed. 

------

**Testing instructions:**

I've installed this PR in: https://huggingface.co/spaces/gradio/test-gr-load

You can confirm this works by `gr.load()`-in the Space as follows and making sure the loaded Space works fully:

```py
import gradio as gr

io = gr.load("spaces/gradio/test-gr-load")
io.launch()
```


client/python/gradio_client/utils.py
```diff
@@ -856,9 +856,9 @@ def _json_schema_to_python_type(schema: Any, defs) -> str:
     elif type_ == "null":
         return "None"
     elif type_ == "const":
-        return f"Litetal[{schema['const']}]"
+        return f"Literal[{schema['const']}]"
     elif type_ == "enum":
-        return f"Literal[{', '.join([str(v) for v in schema['enum']])}]"
+        return "Literal[" + ", ".join(["'" + str(v) + "'" for v in schema['enum']]) + "]"
```
abidlabs, line 861: Unrelated fix that just adds quotation marks around literals. So instead of the type hint in the API docs being: `Literal[a]`, it will be correctly displayed as `Literal["a"]`

gradio/blocks.py
```diff
@@ -2252,9 +2255,11 @@ def get_api_info(self):
         """
         config = self.config
         api_info = {"named_endpoints": {}, "unnamed_endpoints": {}}
-        mode = config.get("mode", None)
 
-        for d, dependency in enumerate(config["dependencies"]):
+        for dependency in config["dependencies"]:
```
abidlabs, line 2259: Simplified this block of code a bit

abidlabs:
Need to add / fix tests, which I'll do tomorrow, but otherwise should be good for review

hysts (approved):
Thank you! So now `gr.Examples` no longer shows `/load_example` too. Awesome!
LGTM!

abidlabs:
> Thank you! So now gr.Examples no longer shows /load_example too. Awesome!

Yeah that was bugging me as well

abidlabs:
Noticed an issue while adding the tests, will fix

js/app/src/api_docs/ApiDocs.svelte
```diff
@@ -79,9 +79,15 @@
...
-	get_info().then((data) => (info = data));
+	get_info().then((data) => {
+		info = data;
+		console.log("info", info);
+	});
 
-	get_js_info().then((js_api_info) => (js_info = js_api_info));
+	get_js_info().then((js_api_info) => {
+		js_info = js_api_info;
+		console.log("js_info", js_info);
```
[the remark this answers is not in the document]
abidlabs: Ah thanks I'll clean that up

js/app/src/api_docs/ApiDocs.svelte
```diff
@@ -182,11 +187,11 @@
 				<InstallSnippet {current_language} />
 
 				{#if Object.keys(info.named_endpoints).length}
-					<h2 class="header">Named Endpoints</h2>
+					<h2 class="header">Endpoints</h2>
```
freddyaboulton, line 190: The deployed space says "undefined" right underneath the Enpoints header

image

js/app/src/api_docs/ApiDocs.svelte
```diff
@@ -144,13 +148,14 @@
 </script>
 
 {#if info}
-	{#if Object.keys(info.named_endpoints).length + Object.keys(info.unnamed_endpoints).length}
+	{#if dependencies.filter((dependency) => dependency.api_name && dependency.show_api).length}
```
freddyaboulton, line 151: I think we're writing `dependency.api_name && dependency.show_api` enough times in this file that we can write a helper function in the script tag to improve readability

js/app/src/api_docs/ApiDocs.svelte
```diff
@@ -79,9 +79,13 @@
 
 	let js_info: Record<string, any>;
 
-	get_info().then((data) => (info = data));
+	get_info().then((data) => {
```
freddyaboulton, line 82: Unrelated but I think it'd be nice if we fixed the url for deployed spaces. I think there's an issue open

image

client/python/gradio_client/utils.py
```diff
@@ -856,9 +856,11 @@ def _json_schema_to_python_type(schema: Any, defs) -> str:
     elif type_ == "null":
         return "None"
     elif type_ == "const":
-        return f"Litetal[{schema['const']}]"
```
freddyaboulton, line 859: 🤦‍♂️

gradio/blocks.py
```diff
@@ -2252,9 +2276,15 @@ def get_api_info(self):
...
+            if (
+                not dependency["backend_fn"]
```
freddyaboulton, line 2282: I think this change does not account for tabbed interfaces any longer and so the tabbed interface api docs are broken

image


I ran the `stt_or_tts` demo

freddyaboulton (commented):
Nice @abidlabs ! Left some nits. Can be merged once they're addressed - no need to request re-review imo.

js/app/src/api_docs/ApiDocs.svelte
abidlabs: I actually don't see the footer at all in the deployed Space? How are you seeing this?

image

js/app/src/api_docs/ApiDocs.svelte
abidlabs: Let me refactor this

js/app/src/api_docs/ApiDocs.svelte
abidlabs: Yeah I was thinking about fixing this, will do so in a follow-up PR

gradio/blocks.py
abidlabs: Interesting, nice catch will take a look

abidlabs: I can't repro this, perhaps it was fixed in my last change -- would you be able to double check if its still an issue for you @freddyaboulton?

js/app/src/api_docs/ApiDocs.svelte
abidlabs: If you meant the test Space, I just installed the latest version of this branch and its working for me: 
image

freddyaboulton (approved):
Those two issues I noticed were my bad. LGTM!

abidlabs:
Thank you for the detailed review @freddyaboulton!

PR #6864, opened 2023-12-21
Add some playwright tips to the test strategy document

freddyaboulton:
## Description
Will present at retro

## 🎯 PRs Should Target Issues

## Tests


test-strategy.md
```diff
@@ -108,3 +108,83 @@ For instructions on how to write and run tests see the [contributing guide](http
...
+![download_trace](https://github.com/gradio-app/gradio/assets/41651716/20c279a8-9a56-4dcf-8df0-c4711e305515)
+
+### Tip 4 - Playwright can write the test for you
+
+You can write the basic skeleton of the test automatically by just interacting with the UI!
+
+First, start a gradio demo from the command line. Then use the following command and point it to the URL of the running demo:
+
+```bash
+npx playwright codegen <gradio-url>
```
abidlabs, line 183: just to confirm, this is the regular url, right? might be clearer:
```suggestion
npx playwright codegen <url>
```

test-strategy.md
```diff
@@ -108,3 +108,83 @@ For instructions on how to write and run tests see the [contributing guide](http
...
+For this reason, playright ships with some [retrying assertions](https://playwright.dev/docs/test-assertions#auto-retrying-assertions).
+These assertions will retry until they pass or a timeout is reached, by default 5 seconds.
+So even if playwright checks a DOM element before the server is done, it gives the server a chance to finish by retrying.
+
+
+You can increase the timeout manually as well:
+
+```js
+// 5 seconds
+await expect(page.getByText('Hidden text')).toBeAttached({timeout?: 5000});
```
abidlabs, line 135: For clarity, could you include what this test would look like without the retrying assertion?

test-strategy.md
```diff
@@ -108,3 +108,83 @@ For instructions on how to write and run tests see the [contributing guide](http
...
+  const response = await page.request.get('https://api.example.com');
+  expect(response.status()).toBe(200);
+}).toPass();
+```
+
+### Tip 2 - Don't rely on internal network calls to check if something is done
+Internal network calls are not visible to the user, so they can be refactored whenever.
+If we have tests that rely on a request to a given route finishing before moving on, for example, they will fail if we ever change the route name or some other implementation detail.
+
+It's much better to use a retrying assertion that targets a visible DOM element with a larger timeout to check if some work is done.
```
abidlabs, line 152: request: could you provide an example of the right vs. wrong thing to do here?

abidlabs:
This is super helpful @freddyaboulton thanks for sharing this! Some small notes above. A nit: can we put these guidelines in a separate document, as this document is more high-level strategy ("...nor does it contain any best practices that we try to follow when writing tests"). We could perhaps create a `testing-guidelines` folder in the repo with these docs, as well as others we might write in the future

freddyaboulton:
I addressed your comments @abidlabs @hannahblair !

abidlabs (approved):

abidlabs:
Awesome!

freddyaboulton:
Thanks for the reviews!

PR #6868, opened 2023-12-22
Added a max_file_size parameter to the upload route 

vagxrth:
## Description
Added a `max_file_size` parameter that will allow the users to set the upload limit which will prevent the uploading of too many large-sized files to the server

Closes: #3801


abidlabs:
Hi @vagxrth is this ready for review? It looks like the parameter currently has no effect

vagxrth:
> Hi @vagxrth is this ready for review? It looks like the parameter currently has no effect

I've been trying to figure out the possible reason why the parameter is still unknown. Can you guide me in any way if possible to fix the issue?

abidlabs:
Hi @vagxrth I'll go ahead and close this PR since its not really ready for review. If you'd like to work on a solution, I'd suggest a strategy similar to: https://github.com/tiangolo/fastapi/issues/362

Issue #6873, opened 2023-12-22
File upload with file_count='directory' creates error


joshwilson-dev:
### Describe the bug

gr.File(file_count="directory") causes the error: "ModelMetaclass object argument after ** must be a mapping, not list".

I can fix it by changing line 82 in gradio>components>file.py to:

if self.file_count == "multiple" or self.file_count == "directory":

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

def main(dir):
    return [d.name for d in dir]

demo = gr.Interface(
    fn = main,
    inputs = gr.File(file_count="directory"),
    outputs = gr.Textbox(label="File paths"))

demo.queue().launch()
```

### Logs

```shell
Traceback (most recent call last):
  File "env\lib\site-packages\gradio\queueing.py", line 489, in call_prediction
    output = await route_utils.call_process_api(
  File "env\lib\site-packages\gradio\route_utils.py", line 232, in call_process_api
    output = await app.get_blocks().process_api(
  File "env\lib\site-packages\gradio\blocks.py", line 1559, in process_api
    inputs = self.preprocess_data(fn_index, inputs, state)
  File "env\lib\site-packages\gradio\blocks.py", line 1335, in preprocess_data
    inputs_cached = block.data_model(**inputs_cached)  # type: ignore
TypeError: ModelMetaclass object argument after ** must be a mapping, not list
```

### System Info

```shell
Gradio 4.12.0
```

### Severity

I can work around it

abidlabs:
Thanks @joshwilson-dev for catching this! That seems to be the fix indeed. Would you be able to open a PR with this fix?

joshwilson-dev:
Yep, I'll do it now.

PR #6874, opened 2023-12-23
fix issue 6873: File with file_count='directory' bug

joshwilson-dev:
## Description
Specifies that if file_count is set to directory when using File we should expect a list of files rather than a single file.
Closes: #6873

### DIFF
--- a/.changeset/nasty-eagles-do.md
+++ b/.changeset/nasty-eagles-do.md
@@ -1,0 +1,5 @@
+---
+"gradio": patch
+---
+
+fix:fix issue 6873: File with file_count='directory' bug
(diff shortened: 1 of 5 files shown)


joshwilson-dev:
@abidlabs Could you please review this pull request when you have a moment?

abidlabs:
Thanks @joshwilson-dev! This looks good. Would you also be able to make the same change in `upload_button.py`? 

It would be amazing if you could add a test in `test/test_components.py`. Something short, along these lines:

https://github.com/gradio-app/gradio/blob/50496f967f8209032b753912a4379eb9cea66627/test/test_components.py#L949

just to make sure the `.preprocess()` function doesn't raise any errors.

Thanks for your help!

abidlabs:
Made the updates, we should be good to merge. Thanks again @joshwilson-dev for catching this.

PR #6884, opened 2023-12-26
Component Server fix

aliabid94:
See: https://huggingface.slack.com/archives/C02QZLG8GMN/p1703402300689679?thread_ts=1702901295.299349&cid=C02QZLG8GMN

### DIFF
--- a/gradio/components/file_explorer.py
+++ b/gradio/components/file_explorer.py
@@ -139,10 +139,8 @@ class FileExplorer(Component):
         return FileExplorerData(root=root)
 
     @server
-    def ls(self, value=None) -> list[dict[str, str]] | None:
+    def ls(self, _) -> list[dict[str, str]] | None:
         """
-        Parameters:
-            value: file path as a list of strings for each directory level relative to the root.
         Returns:
             tuple of list of files in directory, then list of folders in directory
         """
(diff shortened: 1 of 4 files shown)


gradio/components/file_explorer.py
```diff
@@ -139,10 +139,8 @@ def postprocess(self, value: str | list[str] | None) -> FileExplorerData | None:
         return FileExplorerData(root=root)
 
     @server
-    def ls(self, value=None) -> list[dict[str, str]] | None:
+    def ls(self, _) -> list[dict[str, str]] | None:
```
abidlabs, line 142: This change -- making the second argument requried -- is causing the tests to fail

abidlabs (approved):
LGTM thanks @aliabid94!

gradio/routes.py
```diff
@@ -699,7 +699,12 @@ def component_server(body: ComponentServerBody):
                 block = state[component_id]
             else:
                 block = app.get_blocks().blocks[component_id]
-            fn = getattr(block, body.fn_name)
+            fn = getattr(block, body.fn_name, None)
+            if fn is None or not getattr(fn, "_is_server_fn", False):
```
akx, line 703: `None` wouldn't have the attribute either, so this simplifies to
```suggestion
            if not getattr(fn, "_is_server_fn", False):
```

aliabid94: makes sense, thanks!

aliabid94: actually needed for type check :/

Issue #6872, opened 2023-12-22
gr.Video output showing empty 


yadavj2008:
### Describe the bug

**Problem statment:**
Trying to use gradio Video feature to allow apps to capture video from webcam, when clicking on record and submit video button a function process_video() get trigger, this function takes video as input and process further, looks like the gr.Video is not sending the video bytes to the function, hence, function call return None every time.

**Expected behaviour:**

The user should be able to capture gr.Video output and pass it to the function

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr


def video_flip(video):
    if video is None:
        print("No Video")
    return video


iface = gr.Interface(video_flip, gr.Video(sources="webcam"), "playable_video")

iface.launch(share=False, debug=True)
```

### System Info

```shell
gradio                           4.11.0
gradio_client                    0.7.3

OS: MacOS
Notebook: Colab
```

### Severity

Blocking usage of gradio

yadavj2008:
Hi @abidlabs Can you please help on this?. Thanks

abidlabs:
I can repro this issue, we'll take a look!

yadavj2008:
Hi @dawoodkhan82 thanks a lot for fixing this video issue, I just tested the PR changes and it looks like the video output is working for gr.Interface but not for gr.Blocks. Can you please check and see if there is any issue with block or I'm doing something wrong?.

abidlabs:
Hi @yadavj2008 can you provide a code snippet with gr.Blocks that is not working for you?

yadavj2008:
Hi @abidlabs,
I noticed that the issue was in my application was due to the longer video rendering time, i did a quick test using below sample code and it seems that Gradio is taking approx 1 min to render a 2secs video. Is there any way to optimise the performance?.

```
import gradio as gr

def video_flip(video):
    if video is None:
        print("No Video")
    return video

with gr.Blocks() as demo:
      with gr.Row():
        video_in = gr.Video(sources="webcam",format="mp4",include_audio=True)
        video_out = gr.Video(show_label=False)
        submit = gr.Button("Submit")
        submit.click(fn=video_flip, inputs=video_in, outputs=video_out)

demo.launch(share=False, debug=True)

```

yadavj2008:
Any further update on this?. Also, looks like the patch only works when I have "share=False"

abidlabs:
[...]

PR #6885, opened 2023-12-26
Fix issue with Webcam Recording

dawoodkhan82:
## Description
Fix issue with no video being passed to backend when recording with webcam.

Closes: #6872

## 🎯 PRs Should Target Issues

## Tests


js/image/shared/Webcam.svelte
```diff
@@ -88,15 +91,14 @@
...
+			ReaderObj.onload = async function (e): Promise<void> {
 				if (e.target) {
-					dispatch("capture", {
-						//@ts-ignore
-						data: e.target.result,
-						name: "sample." + mimeType.substring(6),
-						is_example: false,
-						is_file: false
-					});
+					let _video_blob = new File([video_blob], "video.mp4");
```
aliabid94, line 96: previously we had the logic `"sample." + mimeType.substring(6)` to get the file extension. Now it's always an mp4. Is that a safe assumption?

aliabid94 (approved):
Works, thanks for fixing!

js/image/shared/Webcam.svelte
dawoodkhan82: Good point, it can be `.webm` or `.mp4`. Fixed.

PR #6931, opened 2024-01-02
Fix functional tests

aliabid94:
Should make all functional tests pass. Also makes it so if the connection to the backend breaks, the app shows an error instead of pending indefinitely.


demo/hello_world/run.py
abidlabs, line 1: Can we not change this file? I'm pretty sure we use this in our docs

abidlabs:
Nice @aliabid94. What was the issue? That we weren't closing the stream if there was an error? How was this related to the functional tests?

demo/hello_world/run.py
aliabid94: accident, reverted

client/js/src/client.ts
```diff
@@ -1053,6 +1050,17 @@ export function api_factory(
 						pending_stream_messages[event_id].push(_data);
 					}
 				};
+				event_stream.onerror = async function (event) {
```
aliabid94, line 1053: Not completely related, but wanted to have an error if the connection broke. Previously, if you closed the backend while a function was running, the frontend would indefintely show "processing". Now it clearly displays an error message.

demo/cancel_events/run.py
```diff
@@ -5,7 +5,7 @@
 def fake_diffusion(steps):
     for i in range(steps):
         print(f"Current step: {i}")
-        time.sleep(1)
```
aliabid94, line 8: Made timeout shorter cause test was taking a while

js/app/test/cancel_events.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+	page
+}) => {
+	const start_button = await page.locator("button", {
+		hasText: /Start Iterating/
+	});
+	const textbox = await page.getByLabel("Iterative Output");
+
+	let output_values: string[] = [];
+	let last_output_value = "";
+	let interval = setInterval(async () => {
```
aliabid94, line 13: Previously relied on websocket events to check for value changes. Now just store all value changes until 8, and then make sure all values 1-7 are stored in array.

js/fileexplorer/shared/DirectoryExplorer.svelte
```diff
@@ -21,12 +20,14 @@
 	}>();
 	const tree = make_fs_store();
 
-	$: glob,
-		ignore_glob,
-		root,
+	const render_tree = (): void => {
```
aliabid94, line 23: particulary nasty bug where all "object" types are considered changed in svelte due to how we re-render all components. This caused svelte to assume the server object changed on any  re-render, causing an `ls` server function to be fired on every component change, refreshing the tree. Took server out of the reactive statement to fix this.

js/image/Index.svelte
```diff
@@ -65,7 +65,7 @@
 	}>;
 
 	$: url = _value?.url;
-	$: url && gradio.dispatch("change");
+	$: url, gradio.dispatch("change");
```
aliabid94, line 68: Reacts to url being null as well

js/upload/src/Upload.svelte
```diff
@@ -48,14 +48,12 @@
...
 				if (type) {
-					dispatch("load", null);
 					items[i].getType(type).then(async (blob) => {
 						const file = new File(
 							[blob],
 							`clipboard.${type.replace("image/", "")}`
 						);
-						const f = await load_files([file]);
-						dispatch("load", f?.[0]);
+						await load_files([file]);
```
aliabid94, line 56: dispatch already fired within load_files. Double dispatch was causing flaky issues

gradio/blocks.py
```diff
@@ -429,6 +429,8 @@ def postprocess_update_dict(
         else:
             update_dict["value"] = value
     update_dict["__type__"] = "update"
+    if "root" in update_dict:  # reserved keyword
+        update_dict["_root"] = update_dict["root"]
```
abidlabs, line 433: As discussed in retro, I think a better approach would be to:

1. soft deprecate the `root` parameter with a warning
2. add a `root_dir` parameter in `FileExplorer` (this also matches the name of the parameter in `glob.glob()` so that's a nice plus)
3. in the constructor of `FileExplorer`, check to see if `root_dir` is not supplied and if `root` is. If that's the case, do `self.root_dir = root`
4. in the frontend, replace `root` with `root_dir`

I think in that case, you shouldn't have to make any hacky changes here

abidlabs:
Amazing work @aliabid94! Just left one comment above, everything else lgtm

gradio/blocks.py
aliabid94: done

aliabid94:
Should be ready for re-review / merge @abidlabs

abidlabs (approved):
LGTM @aliabid94!

Issue #6930, opened 2024-01-02
`gr.ClearButton()` not working for updating visibility


apolinario:
### Describe the bug

If I try to update the visibility of `gr.ClearButton()` it errors out 

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

def hide_clear():
    return gr.update(visible=False)

with gr.Blocks() as demo:
    files = gr.File(interactive=True)
    clear_button = gr.ClearButton(components=files)
    trigger_hide = gr.Button()
    trigger_hide.click(fn=hide_clear, outputs=clear_button)
demo.launch(debug=True)
```
Errors with 
```shell
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/gradio/queueing.py", line 489, in call_prediction
    output = await route_utils.call_process_api(
  File "/usr/local/lib/python3.10/dist-packages/gradio/route_utils.py", line 232, in call_process_api
    output = await app.get_blocks().process_api(
  File "/usr/local/lib/python3.10/dist-packages/gradio/blocks.py", line 1570, in process_api
    data = self.postprocess_data(fn_index, result["prediction"], state)
  File "/usr/local/lib/python3.10/dist-packages/gradio/blocks.py", line 1443, in postprocess_data
    state[output_id] = self.blocks[output_id].__class__(**kwargs)
  File "/usr/local/lib/python3.10/dist-packages/gradio/component_meta.py", line 157, in wrapper
    return fn(self, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/gradio/components/clear_button.py", line 65, in __init__
    self.add(components)
  File "/usr/local/lib/python3.10/dist-packages/gradio/components/clear_button.py", line 93, in add
    self.click(
  File "/usr/local/lib/python3.10/dist-packages/gradio/events.py", line 276, in event_trigger
    raise AttributeError(
AttributeError: Cannot call click outside of a gradio.Blocks context.
```

### System Info

```shell
gradio==4.12.0
```

### Severity

I can work around it

PR #6932, opened 2024-01-02
Allow `gr.ClearButton` and `gr.DuplicateButton` to be made hidden (and otherwise updated)

abidlabs:
Closes: #6930 and adds a functional test

Repro:

```py
import gradio as gr

def hide_clear():
    return gr.update(visible=False)

with gr.Blocks() as demo:
    files = gr.File(interactive=True)
    clear_button = gr.ClearButton(components=files)
    trigger_hide = gr.Button()
    trigger_hide.click(fn=hide_clear, outputs=clear_button)
demo.launch()
```


aliabid94 (approved):

abidlabs:
Thanks @aliabid94! The problem is that when you have a function that returns an update for `gr.Clearbutton()` or `gr.DuplicateButton()`, it has to construct that class outside of the Blocks context. This was a problem because those constructors attach event listeners, and that throws an error when outside of a Blocks context. I fixed this by checking if we are inside a Blocks context and only then attaching event listeners. (It doesn't matter whether or not event listeners are going to be attached for an update, we're going to throw away everything except the constructor attributes anyways)

Issue #6893, opened 2023-12-27
`ImageEditor` examples don't work properly


merveenoyan:
### Describe the bug

I have an ImageEditor and I feed the example dictionary like following:
```python
    background = Image.open("./background.png")
    layer = Image.open("./layer.png")
    composite = Image.open("./composite.png")

    gr.Examples([[{"background": background, "layers": [layer], "composite":composite}]],
          inputs = [im],
          outputs = [output],
          fn=infer,
          cache_examples=True,
          label='Click on the example to try')
demo.launch(debug=True)
```

My ImageEditor has type of "pil" thus each dictionary value is PIL image.
Although, the examples are cached properly and output seems okay, the example on the UI seems broken and returns an error.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
from huggingface_hub import snapshot_download
import gradio as gr
import numpy as np
import torch
import sys
sys.path.append("..")
from tinysam import sam_model_registry, SamPredictor
from PIL import Image

snapshot_download("merve/tinysam", local_dir="tinysam")

model_type = "vit_t"
sam = sam_model_registry[model_type](checkpoint="./tinysam/tinysam.pth")

predictor = SamPredictor(sam)

def infer(img):
  if img is None:
      gr.Error("Please upload an image and select a point.")
  if img["background"] is None:
      gr.Error("Please upload an image and select a point.")
  # background (original image) layers[0] ( point prompt) composite (total image)
  image = img["background"].convert("RGB")
  point_prompt = img["layers"][0]
  total_image = img["composite"]
  predictor.set_image(np.array(image))

  # get point prompt
  img_arr = np.array(point_prompt)
  if not np.any(img_arr):
    gr.Error("Please select a point on top of the image.")
  else:
    nonzero_indices = np.nonzero(img_arr)
    
    center_x = int(np.mean(nonzero_indices[1]))
    center_y = int(np.mean(nonzero_indices[0]))
    input_point = np.array([[center_x, center_y]])

    input_label = np.array([1])
    masks, scores, logits = predictor.predict(
        point_coords=input_point,
        point_labels=input_label,
    )


    result_label = [(masks[0, :, :], "mask")]
    return image, result_label


with gr.Blocks() as demo:
    gr.Markdown("## TinySAM")
    gr.Markdown("**[TinySAM](https://arxiv.org/abs/2312.13789) is a framework to distill Segment Anything Model.**")
[...]

PR #6933, opened 2024-01-03
Refactor examples so they accept data in the same format as is returned by function, rename `.as_example()` to `.process_example()`

abidlabs:
Changes how values are processed for examples so that (by default), they are processed exactly the same way that regular values are post-processed for that component. In other words, by default, the `.postprocess()` method is used to process example values. This has several advantages:

* It allows users to provide example values in all of the same formats that they provide to the regular component. For example, the `Image` component now accepts examples in numpy format or PIL Images, whereas previously you needed to provide them as string filepaths.
* Reduces backend code. Importantly, this means that one fewer method needs to be implemented for custom components. Custom components still have the option to write a custom processor in case the regular `.postprocess()` method is too computationally heavy or results in a very large payload. 
* Now custom component authors don't need to know about the `move_resource_to_block_cache()` method as this is taken care of automatically as long as they use the `FileData` dataclass.
* Conceptually its more consistent, since there's only way data is being processed in the backend. 

Also, I renamed `as_example()` to `process_example()` as its a bit clearer. Everything is done in a backwards-compatible way. 

Closes: #6893, test that with:

```py
cheetah_img = "cheetah.jpg"

import gradio as gr

background = cheetah_img
layer = cheetah_img
composite = cheetah_img

d = {"background": background, "layers": [layer], "composite":composite}

with gr.Blocks() as demo:
    im = gr.ImageEditor()
    output = gr.Textbox()
    gr.Examples([[d]],
          inputs = [im],
          outputs = [output],
          fn=lambda args:"hi",
          cache_examples=True,
          label='Click on the example to try')


demo.launch()
```


gradio/components/image_editor.py
```diff
@@ -320,7 +320,7 @@ def resolve_path(file_or_url: str | None) -> str | None:
             # If an externally hosted image or a URL, don't convert to absolute path
             if self.proxy_url or client_utils.is_http_url_like(input_data):
                 return input_data
-            return str(utils.abspath(input_data))
+            return self.move_resource_to_block_cache(input_data)
```
pngwn, line 323: This isn't very intuitive for custom component authors. Is there a better abstraction or perhaps just an alternative name that doesn't reference implementation details?

pngwn (approved):
Looks good to me, love the added simplicity of this and `process_example` is a much more sensible name.

I've tested and this work fine as far as I can tell. Do we need to update any of the other `Example` components?

Thanks @abidlabs!

abidlabs:
Thanks for reviewing @pngwn! 

> I've tested and this work fine as far as I can tell. Do we need to update any of the other Example components?

I actually think we might need to fix Video's but that should be the only one

aliabid94 (approved):
LGTM! 
Only thought I had about this was that the we've used the word "process" previously in the context of pre/postprocessing data that is being passed to the inference function. In "process_example", we are only processing for the purpose of displaying, I wonder if "render_example" or something similar is a better name. Not a big deal, just a thought on words

abidlabs:
>  In "process_example", we are only processing for the purpose of displaying, I wonder if "render_example" or something similar is a better name. Not a big deal, just a thought on words

In the case of postprocess, we are also processing the data for the purpose of displaying it in the frontend, so I think its okay. I purposefully wanted to use "process" to indicate that theirs a parallel between .postprocess() and .process_example()

abidlabs:
I'll probably merge this later today after the release btw

wangsr126:
It may not work for gr.Example with `run_on_click=True`.

liyaowei-stu:
When interactive=True, it seems to be ineffective.

liyaowei-stu:
It may not work for gr.ImageEditor with interactive=True.

```
cheetah_img = "composite.png"

import gradio as gr

make_dict = lambda x: {'background':x, 'composite':x, 'layers':[x]}
d = make_dict(cheetah_img)

with gr.Blocks() as demo:
    im = gr.ImageEditor( 
                    label="Target Image (user-drawn mask)",
                    type="pil",
                    brush=gr.Brush(colors=["#FFFFFF"], default_size = 30, color_mode="fixed"),
                    placeholder="Please click here or the icon to upload the image.",
                    sources=["upload"],
                    interactive=True,
                    layers=False,
                    )
    output = gr.Textbox()
    gr.Examples([[d]],
          inputs = [im],
          outputs = [output],
          fn=lambda args:"hi",
          cache_examples=True,
          label='Click on the example to try')


demo.launch(server_name="0.0.0.0", server_port=12345)
```

Issue #6926, opened 2024-01-02
Allow for more UI control on `gr.Interface()` by allowing `gr.Accordion()` to work with it


apolinario:
**Is your feature request related to a problem? Please describe.**  
While `gr.Blocks()` is great and versatile, it also increases the complexity of the app. `gr.Interface()` is simple and allow for streamlined creations, however it has little versatility/compositionally for the layout. While I understand the trade-off between them is exactly that, I think may still be potential for `gr.Interface()` to keep its simplicity while giving a slight bit of control to the user on the design. 

**Describe the solution you'd like**  
To me, the quickest win for the spirit of this issue would be to allow `gr.Accordion()` to work with `gr.Interface()`. I believe allowing people to have a `gr.Interface()` with an `Advanced Settings` accordion that could be collapsed could increase dramatically the scope of `gr.Interface()` as people can expose a very simple UI and leave some parameters available only in the accordion without trading off the simplicity of the interface with user control (and without having to migrate to `gr.Blocks()` just to add an accordion on a very simple UI) 

I've faced it myself here: https://huggingface.co/spaces/multimodalart/Ip-Adapter-FaceID

abidlabs:
Thanks @apolinario, what if we were to support something like this:

```py
gr.Interface(
  fn=...,
  inputs=[
    "textbox",
    gr.Accordion(components=[
      "image",
      "audio"
    ])
  ],
  outputs=...
)

```

apolinario:
Cool API design IMO!

PR #6945, opened 2024-01-03
Add `additional_inputs`, `additional_inputs_accordion` parameters to `gr.Interface`

abidlabs:
You can now put accordions in your `gr.Interface` doing something like this:

```py
gr.Interface(
  fn=...,
  inputs=[
    "textbox",
  ],
  outputs=...
  additional_inputs=[
    "image",
    "audio"
  ])
)

```

Closes: #6926 (cc @apolinario)


gradio/layouts/accordion.py
```diff
@@ -41,13 +58,32 @@ def __init__(
             elem_id: An optional string that is assigned as the id of this component in the HTML DOM. Can be used for targeting CSS styles.
             elem_classes: An optional string or list of strings that are assigned as the class of this component in the HTML DOM. Can be used for targeting CSS styles.
             render: If False, this layout will not be rendered in the Blocks context. Should be used if the intention is to assign event listeners now but render the component later.
+            components: If component(s) are supplied, then they are automatically rendered within the accordion. This provides an alterative to the context-based syntax (`with gr.Accordion...`) that can be used with a `gr.Interface()` to render input or output components within accordions.
```
aliabd, line 61:
```suggestion
            components: If component(s) are supplied, then they are automatically rendered within the accordion. This provides an alternative to the context-based syntax (`with gr.Accordion...`) that can be used with a `gr.Interface()` to render input or output components within accordions.
```

gradio/layouts/accordion.py
```diff
@@ -8,18 +8,34 @@
...
     """
-    Accordion is a layout element which can be toggled to show/hide the contained content.
-    Example:
-        with gr.Accordion("See Details"):
-            gr.Markdown("lorem ipsum")
+    Accordion is a layout element which can be toggled to show/hide the contained content. You can use an Accordion
+    in two ways: as a context manager, or as a component. As a context manager, you define the components inside the
+    accordion by indenting them. As a component, you can pass in a list of components to be rendered inside the
+    accordion using the `components` parameter, which makes it possible to use accordions with `gr.Interface`.
+    Example: (Blocks)
```
aliabd, line 24: This doesn't work. https://gradio-frk61ulb7-hugging-face.vercel.app/main/docs/accordion 

This "Example: (cls)" syntax is meant for cases where a function is used by both Blocks and Interface (for example launch or queue) to show the right example depending on the page the user is in. So that you're not on Blocks and see a Launch example that uses Interface. It doesn't work to show both examples in a class like Accordion. 

I can probably add support for this, but my suggestion if you want both examples to show is to just put them in one code block and add a comment. ie: 

```
 Example:
        # Using Accordion with Blocks 
        with gr.Blocks() as demo:
            with gr.Accordion("See Details"):
                gr.Markdown("lorem ipsum")

        # Using Accordion with Interface 
        demo = gr.Interface(
            fn=image_generator,
            inputs=[gr.Textbox(),
                gr.Accordion("Advanced Options",
                components=[
                    gr.Slider(minimum=1, maximum=10),
                    gr.Checkbox(label="Noise reduction")
                ])
            ],
            outputs=gr.Image())

```
Screen Shot 2024-01-07 at 10 22 23 AM

abidlabs: Thanks @aliabd I made the change.

dawoodkhan82:
@abidlabs I'm getting an `RecursionError: maximum recursion depth exceeded while calling a Python object` when running a simple example like this. 

```
gr.Interface(
  fn=...,
  inputs=[
    "textbox",
    gr.Accordion("More inputs", components=[
      "image",
      "audio"
    ])
  ],
  outputs=...
)
```

aliabid94:
I think we should limit how much flexibility we want for Interface and define the limit at which point we want to direct folks towards using Blocks instead. This feature imo brushes up against that limit. 

I especially don't like that we are splitting the API of Accordion to support a Blocks usage and an Interface usage - Components should be agnostic to what context they are being used in. I remember when we designed Blocks that @pngwn said Interface should be an implementation of Blocks, a design decision that has been very fruitful and I think the suggested API goes against this.

An API I would prefer much more would be:
```python
def generate_img(prompt, seed, neg_prompt):
  # ...

gr.Interface(generate_img, inputs="textbox", additional_inputs=["number", "text], outputs="image")
```

I think this is much simpler and cleaner than creating a second API for Accordion and extracting the components of that.

abidlabs:
> @abidlabs I'm getting an RecursionError: maximum recursion depth exceeded while calling a Python object when running something like this.

@dawoodkhan82 can you share your exact code? I'm not able to reproduce

abidlabs:
> I especially don't like that we are splitting the API of Accordion to support a Blocks usage and an Interface usage - Components should be agnostic to what context they are being used in. I remember when we designed Blocks that @pngwn said Interface should be an implementation of Blocks, a design decision that has been very fruitful and I think the suggested API goes against this.

That's actually not the case because you can use this syntax in `gr.Blocks` as well. So you could do:

```py
with gr.Blocks() as demo:
   gr.Accordion(components=...)
```

and the components would get rendered as if they were defined inside the context manager

aliabid94:
Ok I didn't realize that `components=` worked outside of Interfaces. But we are creating more API surface, meaning more surface for bugs, and ideally there should be only one way to do something. 
The `components=` approach seems to go against every other API format of nesting Components in layouts in the Blocks API format, so even if `components=` works in Blocks, it seems like in practice, the `with` clause is the approach in Blocks, and the kwarg is for Intefaces.
Using `additional_inputs` makes things much cleaner and would reduce the code of this PR by a lot.

aliabid94:
> However, this API is very limiting...

We can't implement everything in Interfaces. If you're thinking about multiple accordions, imo you are already way beyond Interfaces - just use Blocks.

abidlabs:
Ok upon further reflection, I don't feel too strongly, I'll refactor to use the `additional_inputs` parameter, which will at least be consistent with `ChatInterface`

abidlabs:
Much simpler PR ready for review @aliabid94

gradio/interface.py
```diff
@@ -142,13 +146,19 @@ def __init__(
...
             allow_duplication: If True, then will show a 'Duplicate Spaces' button on Hugging Face Spaces.
             concurrency_limit: If set, this is the maximum number of this event that can be running simultaneously. Can be set to None to mean no concurrency_limit (any number of this event can be running simultaneously). Set to "default" to use the default concurrency limit (defined by the `default_concurrency_limit` parameter in `.queue()`, which itself is 1 by default).
+            js: Custom js or path to custom js file to run when demo is first loaded.
+            head: Custom html to insert into the head of the page. This can be used to add custom meta tags, scripts, stylesheets, etc. to the page.
+            additional_inputs: A single Gradio component, or list of Gradio components. Components can either be passed as instantiated objects, or referred to by their string shortcuts. These components will be rendered in an accordion below the main input components. By default, no additional input components will be displayed.
```
aliabid94, line 151: Can we consolidate additional_inputs and additional_inputs_accordion? e.g. additional_inputs can take an Accordion

aliabid94 (approved):
LGTM

gradio/interface.py
abidlabs: How would you specify the components if additional_inputs took an Accordion?

aliabid94: oh right I forgot we deleted the kwarg, nvm

abidlabs:
Thanks @aliabid94

PR #6946, opened 2024-01-03
Fix dropdown warning

jjshoots:
## Description

The dropdown component accepts a list of `(name, value)` tuples for the `choices` argument, the `name` component is displayed in the dropdown, and the `value` item is returned.
When `name != value`, when we select an item from the dropdown, we get the following error:

```
UserWarning: The value passed into gr.Dropdown() is not in the list of choices. Please update the list of choices to include: [<name>]
```

This happens because the `dropdown` component is trying to find `name` in the list of `value`s.
This PR fixes that.

Closes: Can't find an existing issue for this.

### DIFF
--- a/gradio/components/dropdown.py
+++ b/gradio/components/dropdown.py
@@ -158,7 +158,7 @@ class Dropdown(FormComponent):
             )
 
     def _warn_if_invalid_choice(self, value):
-        if self.allow_custom_value or value in [value for _, value in self.choices]:
+        if self.allow_custom_value or value in [v for pair in self.choices for v in pair]:
             return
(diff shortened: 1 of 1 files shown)


abidlabs:
Hi @jjshoots can you provide us example code to reproduce this issue that you’re describing?

jjshoots:
Apologies, it seems like I misinterpreted the bug. The warning goes away with setting `allow_custom_value=True`, but something a bit weirder is going on #6946. Closing this issue since it is no longer relevant.

abidlabs:
Got it, thanks!

Issue #6910, opened 2023-12-30
Gradio cannot be started under window system


Zengyf-CVer:
### Describe the bug

- gradio version 4.12.0
- window 10 system
- python 3.8.0

run python code: 

```python 
import gradio as gr
import time
def my_function(x, progress=gr.Progress()):
    progress(0, desc="Starting...")
    time.sleep(1)
    for i in progress.tqdm(range(100)):
        time.sleep(0.1)
    return x
gr.Interface(my_function, gr.Textbox(), gr.Textbox()).queue().launch()

```

![image](https://github.com/gradio-app/gradio/assets/41098760/a26bf540-4f02-4afa-8fc3-5377e73da44b)

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr
import time
def my_function(x, progress=gr.Progress()):
    progress(0, desc="Starting...")
    time.sleep(1)
    for i in progress.tqdm(range(100)):
        time.sleep(0.1)
    return x
gr.Interface(my_function, gr.Textbox(), gr.Textbox()).queue().launch()

```

### System Info

```shell
- gradio version 4.12.0
- window 10 system
- python 3.8.0
```

### Severity

I can work around it

abidlabs:
Apparently, this was a bug in Python itself: https://www.reddit.com/r/Python/comments/eobtwm/systemerror_no_locals_when_loading_str/, that should be fixed in the latest version of python 3.8

Zengyf-CVer:
@abidlabs 
I upgraded python to the latest version, but the problem is still not solved.

![image](https://github.com/gradio-app/gradio/assets/41098760/7d802c6d-2696-4fa1-b7db-d55474091c25)

abidlabs:
Interesting, can you try a later version of Python, such as Python 3.10?

PR #6948, opened 2024-01-04
Added starlette version restrictions

Zengyf-CVer:
## Description

I upgraded the starlette version to 0.34.0, which solved this bug. If the starlette version is too low, this problem will occur.

Closes: #6910

### DIFF
--- a/.changeset/major-beers-swim.md
+++ b/.changeset/major-beers-swim.md
@@ -1,0 +1,5 @@
+---
+"gradio": patch
+---
+
+fix:Added starlette version restrictions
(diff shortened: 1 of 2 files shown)


abidlabs:
Thanks @Zengyf-CVer. This version constraint is very restrictive and may cause conflicts for users who need to use older versions of starlette for other libraries. Would you be able see what versions of starlette work for you? Specifically what's the oldest version of starlette that works?

abidlabs:
For the reasons mentioned above, I'll close this PR. Happy to reopen once we can relax the constraint a bit.

PR #6953, opened 2024-01-04
added examples label for chatinterface

vlasvlasvlas:
why? because this way you can have a customized label for the block title Examples (translation, other uses, etc)

## Description

The main idea is to have control of the example block label from the chatinterface config

Closes: #(issue)

## 🎯 PRs Should Target Issues

I couldnt find one so i made this pull request.

## Tests

### DIFF
--- a/gradio/chat_interface.py
+++ b/gradio/chat_interface.py
@@ -76,6 +76,7 @@ class ChatInterface(Blocks):
         clear_btn: str | None | Button = "🗑️  Clear",
         autofocus: bool = True,
         concurrency_limit: int | None | Literal["default"] = "default",
+        examples_label:str | None = None,
     ):
         """
         Parameters:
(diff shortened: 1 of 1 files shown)


abidlabs:
Thanks @vlasvlasvlas! Open to this PR conceptually but I think a better approach would be to let the `examples` argument take in a `gr.Examples` object directly. That way, you could customize anything about the examples, without needing an additional argument. 

Would you be open to amending your PR?

vlasvlasvlas:
> Thanks @vlasvlasvlas! Open to this PR conceptually but I think a better approach would be to let the `examples` argument take in a `gr.Examples` object directly. That way, you could customize anything about the examples, without needing an additional argument.
> 
> Would you be open to amending your PR?

ok, i like that! i think its better, ok ill redo this and remake de pr if you think thats fine :)

abidlabs:
Yes, go for it! Thank you @vlasvlasvlas

Issue #6943, opened 2024-01-03
Problem changing Chatbot text size in Blocks


ZQ-Dev8:
### Describe the bug

Reporting a potential bug where lowering the text size via `gr.Blocks(theme = gr.themes.Default(text_size='sm',))` does not affect the size of the text in the chatbot. Oddly, this does work when increasing the text size to `lg`, but values of `sm` (and presumably lower) are ignored. See attached reproduction and screenshots for reference.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```
import gradio as gr
import random
import time

# text_size='sm' is ignored (see screenshot 1) while text_size='lg' is obeyed (see screenshot 2)
with gr.Blocks(theme = gr.themes.Default(text_size='sm',)) as demo:
# with gr.Blocks(theme = gr.themes.Default(text_size='lg',)) as demo:
    chatbot = gr.Chatbot()
    msg = gr.Textbox()
    clear = gr.ClearButton([msg, chatbot])

    def respond(message, chat_history):
        bot_message = random.choice(["How are you?", "I love you", "I'm very hungry"])
        chat_history.append((message, bot_message))
        time.sleep(2)
        return "", chat_history

    msg.submit(respond, [msg, chatbot], [msg, chatbot])

if __name__ == "__main__":
    demo.launch()
    ```

### Screenshot

![ss1](https://github.com/gradio-app/gradio/assets/6372983/b35351c2-516c-4516-873f-d67a02301c39)
![ss2](https://github.com/gradio-app/gradio/assets/6372983/ac1c0b62-2199-49ee-851f-b2284fd72317)

### Logs

```shell
No error message to report.
```

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Linux
gradio version: 4.11.0
gradio_client version: 0.7.3

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.1.2
fastapi: 0.104.1
ffmpy: 0.3.1
gradio-client==0.7.3 is not installed.
httpx: 0.25.0
huggingface-hub: 0.19.4
importlib-resources: 6.1.0
jinja2: 3.1.2
markupsafe: 2.1.3
matplotlib: 3.8.1
numpy: 1.26.1
orjson: 3.9.10
packaging: 23.2
pandas: 2.1.2
pillow: 10.1.0
pydantic: 2.4.2
pydub: 0.25.1
python-multipart: 0.0.6
[...]

PR #6958, opened 2024-01-05
Ensure Chatbot theme text size is set correctly

hannahblair:
## Description

We were hardcoding the font size to the large font variable in `Chatbot.svelte`, so the `text_size` was being ignored when setting the theme. 

Demo:
```
import gradio as gr
import random
import time

# test "sm" "md" and "lg" and check font-size is 13px, 14px and 16px respectively

with gr.Blocks(theme = gr.themes.Default(text_size="lg")) as demo:
    chatbot = gr.Chatbot()
    msg = gr.Textbox()

        time.sleep(2)
        return "", chat_history

if __name__ == "__main__":
    demo.launch()

```
Closes: #6943 

## 🎯 PRs Should Target Issues

## Tests


js/chatbot/shared/ChatBot.svelte
```diff
@@ -306,7 +306,7 @@
 		background: var(--background-fill-secondary);
 		width: calc(100% - var(--spacing-xxl));
 		color: var(--body-text-color);
-		font-size: var(--text-lg);
+		font-size: var(--body-text-size);
```
abidlabs, line 309: I think this will cause a breaking change for demos that use the default theme without specifying a text size, because the default "body-text-size" is medium.

js/chatbot/shared/ChatBot.svelte
```diff
@@ -306,7 +306,7 @@
...
 		line-height: var(--line-lg);
```
abidlabs, line 310: Should we avoid hardcoding this line to `--line-lg` as well?

js/chatbot/shared/ChatBot.svelte
pngwn: Since the previous behaviour was broken, I'd be inclined to call this a bugfix.

abidlabs: Nice! LGTM

> My concern with this fix is that in a demo (with a theme) with a chatbot and multiple other components, the body text sizes won't be consistent throughout each component. Some custom CSS could be used to fix this but I'm wondering whether this might cause developer frustration as well.

That's true, although that is also the case right now with the default theme (chatbot size is larger than main body text) and we haven't heard any issues with it.

abidlabs:
LGTM thanks @hannahblair!

abidlabs (approved):

PR #6477, opened 2023-11-18
Custom component gallery

freddyaboulton:
## Description

Trying something a bit different - embedding the gallery in the website! We should probably do this for the themes too - I think it will help us stay on top of it and make sure it's working.

You can test by going to the `/custom_components/gallery` route in the deployed website URL.

### My screen recorder messed up the colors!
custom_component_gallery_preview

Overview:
* There's a space serving as the backend. Right now it has a curated list of components (manually excluding test components). Going to set up the webhook next week (need help from hub team) so that the backend gets updated as components are published.
* Display all the custom component spaces in a grid kind of like we do for the guides. The grid view uses the color gradient as background like we do in the guides. Use an emoji as a background like spaces on the hub.
* I don't think we should preview the space via iframe because we got feedback that it did not scale well for the theme gallery (theme gallery was only loading like 25 spaces at a time or something).
* Instead, when you click on a space, a sidebar opens up with some info on the space and the iframe where you can view the space. We can't use the <gradio-app> embedding tag because I don't think it can load custom components?

TODO:

- [x] Fix styling
- [ ] Search bar at the top where you can search by name or tag

Out of Scope:
* sort. Not enough custom components yet 😉.
* pagination. Although the backend already supports this.

## 🎯 PRs Should Target Issues

Before your create a PR, please check to see if there is [an existing issue](https://github.com/gradio-app/gradio/issues) for this change. If not, please create an issue before you create this PR, unless the fix is very small. 

Not adhering to this guideline will result in the PR being closed. 

## Tests

1. PRs will only be merged if tests pass on CI. To run the tests locally, please set up [your Gradio environment locally](https://github.com/gradio-app/gradio/blob/main/CONTRIBUTING.md) and run the tests: `bash scripts/run_all_tests.sh`

2. You may need to run the linters: `bash scripts/format_backend.sh` and `bash scripts/format_frontend.sh`

PR #6970, opened 2024-01-08
Style changes to custom components gallery

aliabd:
Made some changes, but open to feedback. The gallery feels a bit wordy to me but I think the descriptions are kind of nice to read. Maybe down the line we can add some screenshots or something. 

I didn't change the actual content in the card too much because I think we're going to replace it with @pngwn's auto docs


abidlabs:
Nice @aliabd! I really like it.

To reiterate some of my feedback from sync:
* I think we should remove the tags
* Would be great to use the HF API to get the custom components (any Space with tag `gradio-custom-component`) as well as the number of likes, and to display the number of likes (the components should be sorted based on # of likes as well)
* 

Some additional feedback:
image
* This header is nice, but the link is incorrect
* As in the Guides, it would be nice if the search bar showed the total number of components as well:

image

aliabd:
thanks @abidlabs! other call feedback (just listing for completeness) 

- replace the content of the card with just the space, which will contain the auto docs 
- remove the name (or at least "gradio_") as its not as useful as the description 
- remove the tags but keep them as searchable.

abidlabs:
Perhaps we should also include what base component a custom component was templated from. This could be helpful for users. I believe that we include this as a tag when a custom component is being created. If not, we should!

aliabd:
Made a lot of changes: 


gallery


- Removed tags 
- Showing number of likes
- Sorting by likes 
- Fixed header 
- Showing number of components in search bar 
- Replaced card with just the space 
- Removed “gradio_” from the titles 
- Showing the templates
- Better hovering
- Showing the author

I didn't switch to using hf api because that's actually what @freddyaboulton is already doing in https://huggingface.co/spaces/gradio/custom-component-gallery-backend/blob/main/app.py exactly like we want. I think it was just easier in python and the space made sense as a backend. I think it's still curated though based on this comment https://github.com/gradio-app/gradio/pull/6477#issue-2000129693

Also currently I'm getting likes in a very dumb slow way because i couldn't figure out how to add it to freddy's space. Probably because webhooks aren't set up? Will need his help with that. But for now results will lag and resort themselves..

Will also refactor a lot of the code and move it to page.server.ts

abidlabs:
Amazing UX @aliabd (love the hover)! Just a few little things from my end:

* In the case that the template is Fallback, I think we should just remove it. That means that the user didn't specify a template when creating the custom component
* The default font size for the description feels a bit large to me, and results in many of the descriptions being cut off at the default zoom in my browser (though the ellipses are a nice UI):

image

* Yeah we need to fix that lag/resorting...

But overall super nice! Excited to release this

js/_website/src/routes/custom-components/gallery/+page.svelte
```diff
@@ -0,0 +1,212 @@
...
+		)
+			.then((response) => response.json())
+			.catch((error) => `Error: ${error}`);
+		define_colors(components);
+		if (!components_length) {
+			components_length = components.length;
+		}
+		for (const component of components) {
+			let info;
+			info = await fetch(`https://hf.co/api/spaces/${component.id}`)
```
freddyaboulton, line 63: The backend returns likes now so we can get rid of this

freddyaboulton (approved):
Great pr @aliabd !! I love the new style.

All of the feedback from @abidlabs has been addressed and I just tested and works great. Combining with the forthcoming ParamViewer is going to be 🔥 

One thing, not everyone will update their components and use the ParamViewer. I wonder if as a fallback we can at least link to the space so that users can look at the code? Thanks!

aliabd:
Sounds good @freddyaboulton I added a link to the space. 

I removed my likes code so now everything is snappy again :) I also now load all the spaces in the beginning and clicking on one just toggles hidden. So now clicking around is much faster.

freddyaboulton:
> I also now load all the spaces in the beginning and clicking on one just toggles hidden.

I'm a little wary about this. That's what we do in the custom components gallery and people have complained that it impacts the performance of the page. Since we load the iframe "in the background" I think in this case it may be ok. Let's give it a shot since the page is very smooth right now.

Thanks @aliabd ! Good to merge from my point of view. Love the go to space button!

aliabd:
Got it will keep this in mind and if there are performance issues we can definitely go back to the original way. 

we should probably wait for autodocs before merging as merging this will launch it on the website

freddyaboulton:
> we should probably wait for autodocs before merging as merging this will launch it on the website

Autodocs is an opt-in feature. After it's released the existing custom components won't update to use it automatically. So I think it's good to release without it?

abidlabs:
Before we make the launch announcement, can we update some of the most popular custom components to include autodocs

pngwn:
Autodocs will be done shortly so we can release that today and get some spaces using it before making noise on twitter.

Issue #6929, opened 2024-01-02
`gr.update(value=[])` for `gr.File()` does not clear it


apolinario:
### Describe the bug

If you return a `gr.update(value=[])` or `gr.File(value=[])` it does not clear and get back to the state where you can upload a file, but rather gets on a state where the component is empty and you still need to click the close button to clear

![image](https://github.com/gradio-app/gradio/assets/788417/9b3e9d80-b104-4d6d-a656-ac506406c52a)

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

def clear_files():
    return gr.update(value=[], interactive=True)

with gr.Blocks() as demo:
    files = gr.Files()
    clear = gr.Button()
    clear.click(fn=clear_files, outputs=files)
```
(this is jus a minimal repro, my use-case is more complex, hence I can't use `gr.ClearButton()` 

### System Info

```shell
gradio==4.12.0
```

### Severity

I can work around it

pngwn:
You should be able to clear it if you set that value to `None`.

PR #6980, opened 2024-01-09
`gr.update(value=[])` for `gr.File()` clears it

dawoodkhan82:
## Description
Setting an empty list as the value to `gr.File` will set the component to a clear state.

Closes: #6929

## 🎯 PRs Should Target Issues

## Tests


js/file/shared/File.svelte
```diff
@@ -20,7 +20,7 @@
 	label={label || "File"}
 />
 
-{#if value}
+{#if value && (Array.isArray(value) ? value.length > 0 : false)}
```
aliabid94, line 23: shouldn't this be:
```suggestion
{#if value && (Array.isArray(value) ? value.length > 0 : true)}
```
because if its not an Array, it is automatically valid?

Issue #6950, opened 2024-01-04
(Reload mode) RuntimeError: Caught handled exception, but response already started


kavincsekaran:
### Describe the bug

Getting a runtime error after upgrading to Gradio 4.0 when app is launched in reload mode and encounter any error.

What is happening:
1. Launch app in reload mode.
2. Open app in browser
3. Have any error in the code.
4. Fix error in code.
5. App reloads in browser
6. Have any interactive component like a button and interact with the app by clicking the button.
7. Get the runtime error.

Cannot use the app until the app is relaunched.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

Working code with an interactive component. Save as app.py:
```python
import gradio as gr

def sum_nums(n1, n2, n3):
    return n1+n2+n3

with gr.Blocks() as demo:
    with gr.Row():
        gr.HTML("""<h1 style = "text-align:center">Reproducing Bug</h1>""")

    with gr.Row():
        num1 = gr.Number(label="Number 1", value = 0.0)
        num2 = gr.Number(label="Number 2", value = 0.0)
        num3 = gr.Number(label="Number 3", value = 0.0)

    with gr.Row():
        sum = gr.Textbox(label = "Sum", value = 0.0)
        calc_bttn = gr.Button(value="Calculate sum")
        calc_bttn.click(sum_nums, [num1, num2, num3], sum)
        


if __name__ == "__main__":
    demo.launch()
```
1) Launch app in reload mode: ```gradio app.py```
2) Open app in browser
3) Comment out a line of code to trigger an error: ```#sum = gr.Textbox(label = "Sum", value = 0.0)```
4) Uncomment the line of code to have a working code: ```sum = gr.Textbox(label = "Sum", value = 0.0)```
5) Enter input for "Number 1", "Number 2", "Number 3" and press the "Calculate sum" button.
6) Error in browser and console

![Screenshot 2024-01-03 231343](https://github.com/gradio-app/gradio/assets/2325146/632938aa-79c4-449a-bb4d-72a45f6a3c22)
![Screenshot 2024-01-04 000659](https://github.com/gradio-app/gradio/assets/2325146/6ea6372c-96ed-4924-8087-15ae529defb2)

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Windows
gradio version: 4.11.0
gradio_client version: 0.7.3

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
[...]

Issue #6878, opened 2023-12-23
Hot reload not working on python 12


SalamanderXing:
### Describe the bug

I ran my demo gradio app and when i make changes to my files i got an error.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import time
import gradio as gr


def slow_echo(message, history):
    for i in range(len(message)):
        time.sleep(0.05)
        yield "You typed: " + message[: i + 1]

css = ""
with gr.Blocks(css=css) as main_block:
    gr.ChatInterface(
        slow_echo,
        theme=gr.themes.Soft(),
        clear_btn=None,
        retry_btn=None,
        undo_btn=None,
        examples=["hey"],
    ).queue()

main_block.launch()
```
Run

 `gradio test.py` 

I first run it and it displays well, but then i make a change and get an error and have to exit and restart gradio.

### Logs

```shell
Warning: Cannot statically find a gradio demo called demo. Reload
work may fail.
Watching: 
'/home/bluesk/venv/lib/python3.12/site-packages/gradio', 
'/home/bluesk/Documents/aikosmo', 
'/home/bluesk/Documents/aikosmo'

Running on local URL:  http://127.0.0.1:7860

To create a public link, set `share=True` in `launch()`.
Changes detected in: /home/bluesk/Documents/aikosmo/test.py
Exception in thread Thread-7:
Traceback (most recent call last):
  File "/usr/lib/python3.12/threading.py", line 1052, in _bootstr
ap_inner
    self.run()
  File "/usr/lib/python3.12/threading.py", line 989, in run
    self._target(*self._args, **self._kwargs)
  File "/home/bluesk/venv/lib/python3.12/site-packages/gradio/uti
ls.py", line 217, in watchfn
    demo = getattr(module, reloader.demo_name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: module 'test' has no attribute 'demo'
```
```


### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Linux
gradio version: 4.12.0
gradio_client version: 0.8.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.2.0
fastapi: 0.100.1
ffmpy: 0.3.1
gradio-client==0.8.0 is not installed.
httpx: 0.25.1
huggingface-hub: 0.19.4
importlib-resources: 6.1.0
[...]

PR #6983, opened 2024-01-09
Fix the reloader

aliabid94:
Our reloader was broken (e.g. `gradio run.py`). Now it should work.

Fixes: #6950
Fixes: https://github.com/gradio-app/gradio/issues/6878


gradio/cli/commands/reload.py
```diff
@@ -56,16 +54,10 @@ def _setup_config(
         relpath = original_path
     module_name = str(relpath.parent / relpath.stem).replace(os.path.sep, ".")
 
-    gradio_folder = Path(inspect.getfile(gradio)).parent
```
aliabid94, line 59: I removed the listener to the gradio folder itself, because there are many parts of the gradio the library that are not compatible with reloading - better to just remove the whole directory.

gradio/queueing.py
```diff
@@ -110,6 +110,14 @@ def __init__(
...
     def start(self):
         self.active_jobs = [None] * self.max_thread_count
+        self.set_event_queue_per_concurrency_id()
+
+        run_coro_in_background(self.start_processing)
+        run_coro_in_background(self.start_progress_updates)
+        if not self.live_updates:
+            run_coro_in_background(self.notify_clients)
+
+    def set_event_queue_per_concurrency_id(self):
```
aliabid94, line 120: Extracted this out because it needs to be run every reload.

gradio/routes.py
```diff
@@ -670,6 +670,8 @@ async def queue_join(
             request: fastapi.Request,
             username: str = Depends(get_current_user),
         ):
+            blocks = app.get_blocks()
```
aliabid94, line 673: we need to use get_blocks, because the blocks reference passed from create_app does not get refreshed on reloaded

gradio/utils.py
```diff
@@ -206,11 +208,11 @@ def iter_py_files() -> Iterator[Path]:
             try:
                 module = importlib.import_module(reloader.watch_module_name)
                 module = importlib.reload(module)
-            except Exception as e:
+            except Exception:
                 print(
                     f"Reloading {reloader.watch_module_name} failed with the following exception: "
                 )
-                traceback.print_exception(None, value=e, tb=None)
+                traceback.print_exc()
```
aliabid94, line 215: want the whole traceback, not just the last line

js/app/src/Blocks.svelte
```diff
@@ -638,6 +638,7 @@
 			}
 		});
 
+		if (render_complete) return;
```
aliabid94, line 641: event listener should only be attached once, otherwise every reload doubles up the listeners.

aliabid94:
> Nice! Does this fix https://github.com/gradio-app/gradio/issues/6950?

Yep!

demo/cancel_events/run.py
```diff
@@ -5,7 +5,7 @@
 def fake_diffusion(steps):
     for i in range(steps):
         print(f"Current step: {i}")
-        time.sleep(0.2)
```
aliabid94, line 8: fix a flaky functional test because this was updating too fast

gradio/cli/commands/reload.py
abidlabs: Wait why are we dropping this? Now the reloader doesn't apply if you are changing gradio source files? I would regularly use this (and it would work fine).

aliabid94: ok reverted

js/app/src/Blocks.svelte
abidlabs: I don't follow what this line is doing

abidlabs (approved):
LGTM! Nice @aliabid94 

Also closes: https://github.com/gradio-app/gradio/issues/6878

PR #6994, opened 2024-01-10
Switch default order for sources for `gr.Video` so that upload is the default

abidlabs:
Switches the default order of sources for the `gr.Video` component so that it now starts in the upload state instead of the webcam state.


aliabd (approved):
nice lgtm!

dawoodkhan82 (approved):

abidlabs:
Thanks guys for reviewing!

PR #6997, opened 2024-01-10
Design changes to Playground

aliabd:
Made a few design changes to the playground to make it more immersive. Removed a lot of the navbar and the text. Feedback welcome

https://github.com/gradio-app/gradio/assets/9021060/4716a759-f8af-4346-b29b-eddaf3094c89


abidlabs (approved):
Love it! Gave some feedback internally

PR #6999, opened 2024-01-10
lite build fix: install build package 

abidlabs:


aliabd (approved):
nice i think this'll do it

Issue #6886, opened 2023-12-27
Visibility and Interactive parameters for gr.Tab() component


dwipper:
- [X] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
Since gr.Tab() is a grouping similar to gr.Row or gr.Column for developing multi-tab UIs where the user experience needs to be controlled, i.e. fill out items in the first tab, then the second tab, then the third tab, etc.

**Describe the solution you'd like**  
Have a 'visible=' parameter that controls whether or not the tab (and underlaying components) are visible or not. 'True' would be the default. If False, the tab label would be hidden.

Have an 'interactive=' parameter. 'True' would be the default. If 'False' and the tab was 'visible=True', the user could not click/navigate to the tab.

**Additional context**  
This is what got me thinking about the issue. I didn't realize gr.Tab didn't have a visibility param.......
(https://discord.com/channels/879548962464493619/1025174734427656283/threads/1189201383765856297)

thiswillbeyourgithub:
Same here. I was expecting Tab to have a visibility param

abidlabs:
Yes that makes sense, we have an existing issue for visibility of `gr.Tab`: https://github.com/gradio-app/gradio/issues/2592

I'll rename this issue to focus on the interactivity bit

Issue #2592, opened 2022-11-02
allow setting `gr.Tab` visibility


vzakharov:
- [x] I have searched to see if a similar issue already exists.

Currently `gr.Tab` [doesn’t](https://www.gradio.app/docs/#tab) seem to have a `visible` parameter, unlike most other components. This makes it impossible to hide a Tab depending on certain conditions

Say, in my [Jukebox Web UI](https://github.com/vzakharov/jukebox-webui) this could be hiding or showing “continue composing” tab depending on whether or not the very first sample has already been generated.

I wonder if it’s possible to add such a parameter?

pngwn:
I don't think this is documented but if you use the full version of Tab instead of the shorthand then you can achieve this behaviour:

```py
import gradio as gr

with gr.Blocks() as demo:
  with gr.Tabs() as tabs:
    with gr.TabItem("Tab 1"):
      ...
    width gr.TabItem("Tab 2"):
      ...
```

in this example `tabs` can be used as an output to an event and you can set the visible state. This will hide the whole tabbed interface, hiding an individual tab is not currently possible due to how tabs actually work. But we could try to make this work.

vzakharov:
It’s good to know the full version, but yeah it doesn’t achieve the intended effect. Thanks for giving this a thought!

hithereai:
Any updates on this one?

abidlabs:
Not yet, the team has other competing priorities so it's likely we won't be able to get to this anytime soon. PRs are welcome!

pseudotensor:
@abidlabs Yes, this has also been issue for me, hard to control gr.TabItem for specific flexible startup modes.  So I just end up with dummy tab that has nothing in it, which is bad looking.

linuxonly801:
> I don't think this is documented but if you use the full version of Tab instead of the shorthand then you can achieve this behaviour:
> 
> ```python
> import gradio as gr
> 
> with gr.Blocks() as demo:
>   with gr.Tabs() as tabs:
>     with gr.TabItem("Tab 1"):
>       ...
>     with gr.TabItem("Tab 2"):
>       ...
> ```
> 
> in this example `tabs` can be used as an output to an event and you can set the visible state. This will hide the whole tabbed interface, hiding an individual tab is not currently possible due to how tabs actually work. But we could try to make this work.

This works for me! 
(I only need to hide a tab, and never need to show it.) the codess are as follows:
 ```python
with gr.Tabs(visible=False) as tabs:
[...]

PR #7018, opened 2024-01-12
Add `visible` and `interactive` params to `gr.Tab()`

hannahblair:
## Description

Allows setting the visibility and interactivity of a single `Tab` component via the `visible` and `interactive` components. 

Closes: #2592 
Closes: #6886 

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/slow-ends-double.md
+++ b/.changeset/slow-ends-double.md
@@ -1,4 +1,5 @@
 ---
+"@gradio/dataframe": minor
 "@gradio/tabitem": minor
 "@gradio/tabs": minor
 "gradio": minor


freddyaboulton (commented):
Nice @hannahblair ! 

Setting interactive and visibility is working well for me when running the audio_debugger demo but updating interactive or visible does not seem to work.

I tried with this demo:

```python
import gradio as gr
import subprocess
import os

audio_file = os.path.join(os.path.dirname(__file__), "cantina.wav")


with gr.Blocks() as demo:
    with gr.Tab("Audio", interactive=False) as a:
        gr.Audio(audio_file)
    with gr.Tab("Interface", visible=False) as b:
        gr.Interface(lambda x:x, "audio", "audio", examples=[audio_file], cache_examples=True)
    with gr.Tab("console", visible=False) as c:
        ip = gr.Textbox(label="User IP Address")
        gr.Interface(lambda cmd:subprocess.run([cmd], capture_output=True, shell=True).stdout.decode('utf-8').strip(), "text", "text")
    
    button = gr.Button(value="Make visible")
    button.click(lambda: (gr.Tab("Interface", visible=True), gr.Tab("console", visible=True)),
                 outputs=[b, c])
    def get_ip(request: gr.Request):
        return request.client.host
    
    demo.load(get_ip, None, ip)
        
if __name__ == "__main__":
    demo.queue()
    demo.launch()
```

I also don't think we need to increase the api of `TabbedInterface`. What's the use case for passing a non interactive or invisible tab?

Otherwise looks great!

freddyaboulton (approved):
Very nice @hannahblair !

abidlabs:
Thanks for tackling the longstanding Tab issues @hannahblair!

Issue #7020, opened 2024-01-13
The load() method cannot update the control method after being updated from 3 to 4


iulkm:
### Describe the bug

In version 3.42.0, this code can be used normally and the control properties can be updated normally. However, after updating to version 4.14.0, the method ran but the control cannot be updated properly.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr
Start = False

def change():
    global Start
    print(Start)
    if Start:
        Start = False
        return gr.Button(interactive=False)
    else:
        Start = True
        return gr.Button(interactive=True)

with gr.Blocks() as demo:
    with gr.Column():
        btn0 = gr.Button("btn0")
    demo.load(change, None, btn0, every=1)
    demo.queue()
    demo.launch(debug=True)
```

### Screenshot

![image](https://github.com/gradio-app/gradio/assets/97347703/a4aa9959-ff42-4429-af47-321d3dadb18c)
![image](https://github.com/gradio-app/gradio/assets/97347703/22b82391-ea30-4bdb-8067-27f00bff635e)

### Logs

```shell
No error reported
```

### System Info

```shell
gradio-4.14.0
```

### Severity

Blocking usage of gradio

Issue #6762, opened 2023-12-12
Info and Warning not shown if executed inside a generator


iosonopersia:
### Describe the bug

Hi, I noticed a strange bug when trying to show Info and Warning messages in the UI of my Gradio app.

It happens only if:
- I use a generator as a function for an event listener;
- the `gr.Info(...)` or `gr.Warning(...)` instruction is placed after at least one `yield` instruction.

Info and warning messages are still correctly shown in the UI when the corresponding Python instructions are executed before any `yield`. You can refer to the reproduction example below.

By means of some debugging, I discovered that in these precise circumstances the following happens: in file `helpers.py`, function `log_message`, lines 1116 and 1117 both result in `None` values, leading to the execution of the if branch at line 1118.
https://github.com/gradio-app/gradio/blob/f3c4d78b710854b94d9a15db78178e504a02c680/gradio/helpers.py#L1113-L1126

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr


def does_not_work():
    yield
    gr.Info("An info message")
    gr.Warning("A warning message")
    yield


def does_work():
    gr.Info("An info message")
    gr.Warning("A warning message")
    yield


with gr.Blocks() as app:
    button_1 = gr.Button(value="This DOES NOT work!")
    button_1.click(does_not_work)

    button_2 = gr.Button(value="This WORKS!")
    button_2.click(does_work)

app.queue().launch()
```

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Linux
gradio version: 4.8.0
gradio_client version: 0.7.1

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.2.0
fastapi: 0.104.1
ffmpy: 0.3.1
gradio-client==0.7.1 is not installed.
httpx: 0.25.2
huggingface-hub: 0.19.4
importlib-resources: 6.1.1
jinja2: 3.1.2
markupsafe: 2.1.3
matplotlib: 3.8.2
numpy: 1.26.2
orjson: 3.9.10
packaging: 23.2
pandas: 2.1.3
pillow: 10.1.0
pydantic: 2.5.2
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0.1
requests: 2.31.0
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.9.0
typing-extensions: 4.8.0
[...]

PR #7029, opened 2024-01-15
Run before_fn and after_fn for each generator iteration

freddyaboulton:
## Description

Closes: #7020
Closes: #6762

The issue is that the `after_fn` which sets `in_event_listener` to False was never run for the iterations of an async generator. This meant the component init run this [line](https://github.com/gradio-app/gradio/blob/f2d69fc7d0c1c3457112e702b53e38a0255fc1b7/gradio/blocks.py#L1443) was not working properly.

To test, you can use the repro from the original issue:

```python
import gradio as gr
Start = False

def change():
    global Start
    print(Start)
    if Start:
        Start = False
        return gr.Button(interactive=False)
    else:
        Start = True
        return gr.Button(interactive=True)

with gr.Blocks() as demo:
    with gr.Column():
        btn0 = gr.Button("btn0", interactive=False)
    demo.load(change, None, btn0, every=1)
    demo.queue()
    demo.launch(debug=True)
```

update_interactive

## 🎯 PRs Should Target Issues

## Tests


gradio/utils.py
```diff
@@ -639,8 +639,8 @@ async def asyncgen_wrapper(*args, **kwargs):
                 before_fn(*before_args)
             async for response in f(*args, **kwargs):
                 yield response
-            if after_fn:
-                after_fn(*after_args)
+                if after_fn:
```
aliabid94, line 642: I don't understand - why should we call the after_fn (which in this case sets in_event_listener to false) right after the first yield, if there are more responses left to be yielded? Why wouldn't we call it after we have yielded all the responses?

freddyaboulton: We need to do this because we create components during the postprocessing (after each iteration in the case of generators) and we need `in_event_listener` to be False for the component to be properly created. 

This is the relevant code -

```python
  state[output_id] = self.blocks[output_id].__class__(**kwargs)

  prediction_value = postprocess_update_dict(
      block=state[output_id],
      update_dict=prediction_value,
      postprocess=block_fn.postprocess,
  )
```

Right now `in_event_listener` is always True because the after function never runs.

aliabid94: Ok this makes sense to me. I believe in this case we should be calling the before_fn before every function call, and calling after_fn right after every function call, but before the yield. Something like (haven't tested this):
```
        @functools.wraps(f)
        async def asyncgen_wrapper(*args, **kwargs):
            while True:
                if before_fn:
                    before_fn(*before_args)
                try:
                    response = next(f(*args, **kwargs))
                except StopAsyncIteration:
                    if after_fn:
                        after_fn(*after_args)
                    break
                if after_fn:
                    after_fn(*after_args)
                yield response
```

This way, in_event_listener is set properly to only be applied while the function is running

freddyaboulton: Nice suggestion! I took a separate approach but the end result is the same, we call before_fn and after_fn for each iteration. I also realized this closes #6762!

aliabid94:
is this ready for re-review?

freddyaboulton:
@aliabid94  Yes !

gradio/utils.py
```diff
@@ -635,12 +635,8 @@ def function_wrapper(
 
         @functools.wraps(f)
         async def asyncgen_wrapper(*args, **kwargs):
```
aliabid94, line 637: why were these lines removed? So it's just not using before_fn and after_fn anymore in generators?

aliabid94 (approved):
LGTM!

freddyaboulton:
Thanks for the review @aliabid94 !

PR #7030, opened 2024-01-15
add autodocs

### LINKED ISSUE #6648: Autogenerate docs for custom components
It would be nice to document a few things for custom components:

* How to import them and use them in a Gradio app (e.g. `from gradio_richtextbox import RichTextbox`)
* What parameters the custom component supports
* What events the custom component supports

This could be included in the Custom Components gallery that @freddyaboulton is building, as well as potentially on the pypi package itself.

### LINKED ISSUE #6684: add command to autodocument custom components
CLI command to autodocument custom components.

- generate nice docs space 
- generate readme (for pypi and github)

pngwn:
## Description

This is the bulk of the work for autodocumenting custom components. To test this you will need to install `gradio_paramviewer` from pip. 

I will be moving this component into gradio core, so we don't need an additional dependency but I haven't quite got round to that.

I also had to remove the readme generation out while i fixed some issues but I'll add that back very shortly.

Closes #6684. Closes #6648.

## 🎯 PRs Should Target Issues

## Tests


pngwn:
I might need help with some of those type errors, I'll swap `toml` for `tomlkit` since that is already being used.

pngwn:
One or two tweaks to come but now reviewable.

pngwn:
Sample markdown readme: https://gist.github.com/pngwn/d53d345e3887de357b5420581a534952

gradio/cli/commands/components/docs.py
```diff
@@ -0,0 +1,101 @@
+import importlib
```
freddyaboulton, line 1: We need to add `from __future__ import annotations` here to support 3.8/3.9.

gradio/cli/commands/components/docs.py
```diff
@@ -0,0 +1,101 @@
...
+from pathlib import Path
+from typing import Optional
+
+import requests
+import toml
```
freddyaboulton, line 6: toml is only included in 3.10. We need to use `tomlkit` here for older versions. It's already installed and AKAIK `toml` is identical to `tomlkit`.

gradio/cli/commands/components/_docs_utils.py
```diff
@@ -0,0 +1,747 @@
...
+                        member_name != "__init__"
+                        and member_name != "preprocess"
+                        and member_name != "postprocess"
+                    ):
+                        continue
+
+                    docs[name]["members"][member_name] = {}
+
+                    member_docstring = inspect.getdoc(member) or ""
+                    hints = typing.get_type_hints(member)
```
freddyaboulton, line 306: I think we need to do `member.__annotations__` here. 

`get_type_hints` is failing for me on python 3.9 because of the use of `|` for unions.

image

gradio/cli/commands/components/_docs_utils.py
```diff
@@ -0,0 +1,747 @@
...
+            return value
+    return None
+
+
+def format(code: str, type: str):
+    """Formats code using ruff."""
+    if type == "value":
+        code = f"value = {code}"
+
+    ruff_args = ["ruff", "format", "-", "--line-length=60"]
```
freddyaboulton, line 22: Ruff may not be installed in a user's system. I think we should make this optional?

pngwn: good point. Or we could just add it as a dep? It really does make everything more readable.

gradio/cli/commands/components/docs.py
```diff
@@ -0,0 +1,101 @@
...
+from typer import Argument, Option
+from typing_extensions import Annotated
+
+from ._docs_utils import extract_docstrings, get_deep, make_markdown, make_space
```
freddyaboulton, line 10: I ran the docs command for all of the components - I only got two failures

- `Fallback` component (e.g. no template) - fails with `KeyError: 'return'`
- `Plot` component - fails with `KeyError: 'value'`

Haven't checked the output yet. That's next.

gradio/cli/commands/components/_docs_utils.py
pngwn: Are these two separate issues? I'm not sure i'm understanding correctly.

If i change `hints = typing.get_type_hints(member)` to `hints = typing.get_type_hints(member.__annotations)` then it fails. Is there some specific issue with an earlier python version?

> get_type_hints is failing for me on python 3.9 because of the use of | for unions.

Is this unions in the docs code, in the component code or somewhere else?

I'll test on 3.9.

gradio/cli/commands/components/docs.py
pngwn: Ah `Fallback` has no hints or docstrings. I'll see if we can fail more gracefully or throw an error if the output isn't helpful.

Strange behaviour for `Plot`, i think it has everything is should. Will check that.

gradio/cli/commands/components/_docs_utils.py
```diff
@@ -0,0 +1,747 @@
...
+    return docs_classes
+
+
+def make_space(docs, name, description, local_version, demo, space, repo, pypi_exists):
+    filtered_keys = [key for key in docs if key != "__meta__"]
+
+    source = f"""
+import gradio as gr
+from app import demo as app
+from {name} import {",".join(filtered_keys)}
```
freddyaboulton, line 657: Are these imported classes actually used? If not we can remove

freddyaboulton (commented):
Awesome PR @pngwn ! I'm still going through this but wanted to comment on some things I've noticed - 

1. In the `space.py` the `from app import demo as app` line will hang forever unless the app file launches the demo in a `if __name__ == '__main__'` clause
2. In the README, `None` shows up for the badges if urls are not specified in the pyproject toml
image

3. I can't launch the space.py for Audio docs because there's a syntax error in the file

image

4. In the README, the types in the param table always show up as empty code blocks?

image

gradio/cli/commands/components/_docs_utils.py
pngwn: Ah no, they won't be used.

pngwn:
@freddyaboulton Can you should be the generated README.md, for the empty code block case?

pngwn:
I have pushed up some fixes.

- I've change the demo generated demo file to contain the `if __name__ == '__main__'` clause, I'm not sure there is anything else we can do.
- I have done some additional cleanup on the Additional classes that we show (this was the issue with the `Audio` component not working)
- Opted for simpler type introspection in version prior to 3.10 which do not have great type introspection capabilities.
- More gracefully handled missing docstring or types (try the `Fallback` component now and it should be much nicer). Likewise plot has a missing type annotation, this should work nicely now.
- Fixed the badges
- Hide parts of the docs if there is no info to show.

gradio/cli/commands/components/_docs_utils.py
```diff
@@ -0,0 +1,892 @@
...
+
+    process = Popen(
+        ruff_args,
+        stdin=PIPE,
+        stdout=PIPE,
+        stderr=PIPE,
+        universal_newlines=True,
+    )
+    formatted_code, err = process.communicate(input=code)
+    print(err)
```
freddyaboulton, line 34: I'm getting that `--line-length` is not a valid ruff parameter name

error: unexpected argument '--line-length' found

image

Just FYI

freddyaboulton:
This is great @pngwn ! Thanks for making the fixes. I can confirm they're working. I noticed the following things:

1. In docs.py I think we need to import tomlkit instead of toml

2. The parameter table no longer displays default values when I use 3.10

image

If I use 3.9 it works lol?

image


In 3.10, the "User Function" always shows `Unknown` for the type hint but not in 3.9

image

I'm seeing the same thing for the READMEs. In 3.9, the type hints get shown in the parameter table but not in 3.10.

gradio/cli/commands/components/_docs_utils.py
pngwn: How are you switching your python version. `Line-length` _is` a valid parameter. I'm guessing that your ruff version for that python version is earlier than 0.1.4 when the param was added or maybe even prior to 0.1.2 when the formatter was added.

I've added a recent version of ruff as a dep now, so this shouldn't be an issue.

pngwn:
@freddyaboulton I think the recent update should address everything, I've also added some prettier output and warnings/ errors for certain cases.

- If a user is on 3.9 or below then they will get a gentle warning at the end of the run that updating python will give them a better experience.
- If there is no `if __name__ == '__main__'"` then the docs command errors but tells the user how to fix the problem, or how to disable the check (`--suppress-demo-check`)

I think CI should be green now.

pngwn:
I have a type error in 3.8 (i thought 3.9 was out min version?). I'll see if i can figure that one out.

freddyaboulton:
Awesome @pngwn ! Can confirm that all of the bugs noted yesterday are now gone. Only have some questions around the content now.

* Capitalize the header titles

image

* Kind of confused by the User function section. It says `As input` for Annotated image but Annotated image is really only used an output component? I think we should explain this section before going into the `As input`

image

* ParamViewer is the only component that has an `As output` section. I think that's a bug?

* Some components do not have a `User function` section (video, annotated image, model3d, etc). If it's expected, can we omit the section? Without context it's a bit confusing what this means. 

image

freddyaboulton:
One tiny bug is that you cannot generate docs for `gr.Button`. Says `TypeError: Object of type type is not JSON serializable`

pngwn:
> Kind of confused by the User function section. It says As input for Annotated image but Annotated image is really only used an output component?

That isn't really true in a blocks world. It might be the intention but all component can be used as inputs or outputs, there is no way to restrict this so they documentation is correct and I'd think we should focus on explaining to authors that components should work as an input or an output even if they don't do much.

> ParamViewer is the only component that has an As output section. I think that's a bug?

I'll check how I'm pulling in the docstrings. We need a docstring for the return of preprocess and the param of postprocess to generate a meaningful entry.

> Some components do not have a User function section (video, annotated image, model3d, etc). If it's expected, can we omit the section? Without context it's a bit confusing what this means.

It's better than nothing imo, this happens because the params / returns  of processors are typed but there is no docstring (or there is a bug in pulling it in). We could use the type when there is no docstring to provide _some_ context.

I'll take a look at Button.

freddyaboulton:
> We need a docstring for the return of preprocess and the param of postprocess to generate a meaningful entry.

Ok I think there may be a bug then, annotatedimage does not have docstrings for either pre or postprocess and it has an `As input` entry. Video does not have docstrings as well and it has no entry.

Regarding the `User function` section, I think we need to add a brief sentence or two explaining what it's trying to communicate and in what context it's relevant. Right now I find it a bit confusing.

gradio/components/paramviewer.py
```diff
@@ -0,0 +1,86 @@
...
+
+class Parameter(TypedDict):
+    type: str
+    description: ParameterOne
+    default: ParameterTwo
+
+
+class ParamViewer(Component):
+    """
+    Displays an interactive table of parameters and their descriptions and default values width syntax highlighting
```
freddyaboulton, line 37: Typo: `width syntax highlighting`

freddyaboulton (commented):
Just noticed these two small things!

gradio/components/paramviewer.py
```diff
@@ -0,0 +1,86 @@
...
+
+from typing import Literal, TypedDict
+
+from gradio.components.base import Component
+from gradio.events import Events
+
+lang_type = Literal["python", "typescript"]
+
+
+class ParameterThree(TypedDict):
```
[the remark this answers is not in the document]
pngwn: Oh sorry, I'll remove these before merging.

pngwn:
The reason the input/ output text was mixed up is because i got them mixed up. Annotated image docs should behave as expected now. Button thing was a bug and should be fixed. 

Just gonna fix that type error in 3.8 then I'll clean up the ParamViewer docs.

freddyaboulton (approved):
Thanks @pngwn !

Only thing I noticed is that the markdown should also have the helpful explanation of the User Function section. Otherwise LGTM once CI is fixed. Thanks!

image

pngwn:
Ah yes ofc. Thanks for catching that!

Issue #6609, opened 2023-11-30
gradio 4 UI locks up when components become active with generators


pseudotensor:
### Describe the bug

In gradio 4 only, for objects like chatbot, once streaming some response, can't navigate anywhere to any other tabs or scroll the scroll bar of the chat window.  Everything gets stuck in UI when the component or backend is busy.

Absolutely no problems in gradio 3 with this.  Never seen before until using gradio 4

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

Working on repro.  Doesn't happen in simplest case of fake streaming.

Happens in h2oGPT, e.g. when doing chatbot, and zephyr is streaming in background in some transformers streaming thread.  Even when nothing being updated in the chatbot text window (say some prompting is off and nothing yielded) still everything locks up in UI.  Can't scroll, click other tabs, etc.

Might be related to this in gradio 3: https://github.com/gradio-app/gradio/issues/6100 and maybe https://github.com/gradio-app/gradio/issues/5914

But this problem is super persistent and reproducible and much worse in terms of not being able to go anywhere.

### Screenshot

You can see me clicking on tabs, can't reach until the chatbot generation is complete.  Even though really nothing is yielded, just the fact that a generator is going is enough to cause issue.

https://github.com/gradio-app/gradio/assets/2249614/b9212c7b-3cf9-4e4c-9d24-9d69a3387ce7

### System Info

```shell
gradio==4.7.1
gradio_client==0.7.0
```

### Severity

Blocking usage of gradio

pseudotensor:
Here's video of the above after entering "Who are you?" and changing to `max_new_tokens=900,` to avoid unrelated problems.

There are many issues:
* Within few seconds, UI locks up.  Scroll, clicking the other tab, etc. all fail to work.  streaming no longer updates, even though if I add a print to gradio code I can see it working.
* A bit later, the gradio server also hangs while streaming
* Right after the gradio server hangs, the web browser complains about hanging



https://github.com/gradio-app/gradio/assets/2249614/ad0156be-b66b-42c2-b966-b9417856d312

pseudotensor:
If one does `max_new_tokens=100,` to avoid too much output, then the UI freezes, but eventually recovers like in the h2oGPT case.

pseudotensor:
[...]

PR #7055, opened 2024-01-18
Fix UI freeze on rapid generators

aliabid94:
Fixes: #6609 

Demo:
```python
import gradio as gr
import time

def bot():
    for i in range(100):
        yield "hello " * i
    print("done")

with gr.Blocks() as demo:
    output = gr.HTML(elem_id="output_html")
    msg = gr.Number()
    msg.submit(bot, None, output)

    n_elements = 5000
    for i in range(n_elements):
        gr.Button()

demo.launch()
```

Will explain fix with code comments

### DIFF
--- a/client/js/src/client.ts
+++ b/client/js/src/client.ts
@@ -290,6 +290,7 @@ export function api_factory(
 			let pending_stream_messages: Record<string, any[]> = {}; // Event messages may be received by the SSE stream before the initial data POST request is complete. To resolve this race condition, we store the messages in a dictionary and process them when the POST request is complete.
 			let event_stream: EventSource | null = null;
 			const event_callbacks: Record<string, () => Promise<void>> = {};
+			const unclosed_events: Set<string> = new Set();
 			let config: Config;
 			let api_map: Record<string, number> = {};
 
@@ -902,9 +903,6 @@ export function api_factory(
 											) {
 												if (event_callbacks[event_id]) {
 													delete event_callbacks[event_id];
-													if (Object.keys(event_callbacks).length === 0) {
-														close_stream();
-													}
 												}
 											}
 										} catch (e) {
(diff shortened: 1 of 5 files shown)


client/js/src/client.ts
```diff
@@ -1042,7 +1041,14 @@ export function api_factory(
...
 					} else if (event_callbacks[event_id]) {
-						await event_callbacks[event_id](_data);
+						if (_data.msg === "process_completed") {
+							unclosed_events.delete(event_id);
+							if (unclosed_events.size === 0) {
+								close_stream();
+							}
+						}
+						let fn = event_callbacks[event_id];
+						window.setTimeout(fn, 0, _data);
```
aliabid94, line 1051: Because SSE onmessage events take high precedence by browsers (higher than websocket onmessage events), they block the browser's rendering if many events are received in quick succession. Moving the event callback to happen after a timeout places the callback in the JS event queue, which reduces its priority to happen after a browser render. 
However, closing the event stream does need to happen immediately if a process_complete is received, otherwise the browser will think the connection broke when the backend closes the stream, so we moved that logic here so it happens in the onmessage immediately and not later in the timeout.

pngwn: Makes sense. Ish.

pngwn, line 1051: Could we add a link to the PR in a comment here, just to provide context since this is slightly strange?

pngwn (approved):
This is awesome. Tested the original repro and everything is working great! Also went through the demos and can't see any changes in behaviour.

Thanks for figuring this one out @aliabid94! Very weird browser behaviour.

freddyaboulton:
Very nice!

client/js/src/client.ts
freddyaboulton: Agreed. I think it would be easy to change this in a refactor inadvertently. How can we test this?

aliabid94: Can't think of any way to test this with our existing framework - the DOM already *does* update, it's the browser that skips the intermediate repaints, which we can't test for with playwright

pseudotensor:
Thanks!  I'm  hopeful I can upgrade to gradio4 soon1

PR #5305, opened 2023-08-23
Rotate axes labels on LinePlot, BarPlot, and ScatterPlot

Faiga91:
## Description

Added two parameters `x_label_angle` and `y_label_angle` to the plot `update` and `init` methods, for native plots: `LinePlot`, `BarPlot`, and `ScatterPlot`.

🎯  Closes: #3160

PR #7056, opened 2024-01-18
fix: don't ignore y_label_angle in bar plots

akx:
## Description

My IDE's linter highlighted that `y_label_angle` wasn't being used.

It's probably never worked since being implemented in #5305.


akx:
Huh, looks like my fork was in a State. Never mind.

PR #7059, opened 2024-01-18
Remove flag_dir from read_from_flag()

akx:
## Description

The argument isn't used by the function, and [it doesn't seem to be used by external code across GitHub either](https://github.com/search?q=read_from_flag+language%3APython&type=code&l=Python&p=1).

This was flagged by Ruff's ARG lint.


abidlabs:
Thank you @akx!!

PR #7061, opened 2024-01-18
Update ruff to 0.1.13, enable more rules, fix issues

akx:
## Description

This PR updates Ruff from 0.1.7 to 0.1.13, enables a handful more rules and fixes the resulting issues.

The `ARG` rules will prevent copy-paste typo bugs and brainfarts, since you'll instantly notice whether you've missed using an argument you might actually want to use.

### DIFF
--- a/.changeset/salty-olives-clean.md
+++ b/.changeset/salty-olives-clean.md
@@ -1,0 +1,8 @@
+---
+"@gradio/preview": minor
+"@gradio/wasm": minor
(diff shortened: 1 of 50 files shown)


akx:
Rebased. Could e.g. @abidlabs take a look?

abidlabs:
Much appreciated @akx. Went through all of the changes and the look good to me!

PR #7066, opened 2024-01-18
Add more reformatting commits to .git-blame-ignore-revs

akx:
## Description

Follows up on #4586 to add more reformatting commits to the ignore file.


pngwn (approved):
Thanks again @akx!

Issue #7086, opened 2024-01-20
gradio client streaming eating too much CPU waiting


pseudotensor:
### Describe the bug

client eating 100% core waiting for server response

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

In client.py's job class in the __next__ while waiting for a response, it loops without any sleep at all and rams against a single core.  Should add some minimal sleep here.

```python
        while True:
            with self.communicator.lock:
                if len(self.communicator.job.outputs) == self._counter + 1:
                    o = self.communicator.job.outputs[self._counter]
                    self._counter += 1
                    return o
                if self.communicator.job.latest_status.code == Status.FINISHED:
                    raise StopIteration()

```

### System Info

```shell
gradio 3.52.0 or 4.15.0


```

### Severity

My only work-around is to change the gradio code itself to add the sleep.

abidlabs:
Ah thanks @pseudotensor! Would you like to open a PR? Happy to review this week

pseudotensor:
Sure, another issue is that this gets caught in an infinite loop if there is an early exception.

PR #7091, opened 2024-01-21
Fixes excessive CPU use by client while it waits for server to send messages during streaming

pseudotensor:
## Description

Without this fix, as long as client is waiting for server messages, the client consumes 100% of 1 core due to rapid looping.  Added small ms sleep that avoids that wasteful use of CPU.

Closes: https://github.com/gradio-app/gradio/issues/7086

## 🎯 PRs Should Target Issues

https://github.com/gradio-app/gradio/issues/7086

## Tests

```
def test_hang():
    import gradio as gr
    import time

    with gr.Blocks() as demo:
        chatbot = gr.Chatbot()
        msg = gr.Textbox()

            time.sleep(10000)
            if not chat_history:
                chat_history = [[message, '']]
            chat_history[-1][1] = message
            for fake in range(0, 1000):
                chat_history[-1][1] += str(fake)
                time.sleep(0.1)
                yield "", chat_history
            return

        msg.submit(respond, [msg, chatbot], [msg, chatbot], api_name='submit')

    demo.queue(api_open=True)
    demo.launch(prevent_thread_lock=True, show_error=True, )

    client = Client('http://localhost:7860')
    job = client.submit('', None, api_name='/submit')

    for res in job:
        print("Reached inside loop: %s" % res)
```

### DIFF
--- a/client/python/gradio_client/client.py
+++ b/client/python/gradio_client/client.py
@@ -1559,6 +1559,7 @@ class Job(Future):
                     return o
                 if self.communicator.job.latest_status.code == Status.FINISHED:
                     raise StopIteration()
+                time.sleep(0.001)
 
     def result(self, timeout: float | None = None) -> Any:
         """


abidlabs (approved):
LGTM. Thanks for raising the issue and contributing the fix @pseudotensor

Issue #7093, opened 2024-01-21
Video supoort show_download_button attribute


tradoff:
- [ yes ] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
Why gr.Video does not have an attribute "show_download_button" just like gr.Image? I think it could be needed in some case... 

**Describe the solution you'd like**  
Add an attribute "show_download_button" attribute for gr.Video.

**Additional context**

Issue #7088, opened 2024-01-20
missing download button for gr.components.Audio


ngaudemet:
### Describe the bug

their should be a download button in my Audio component
but I can't see it anywhere, neither on Safari or Chrome, neither when I upload a mp3 nor when I record from my microphone.
it is important that users are able to download their mic recording. 
my version of gradio is 4.12.0

the PR "Add download button for audio #4977" should have solved this but it seems it doesn't work.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

audio = gr.components.Audio(
                label="Upload an audio"
                type="filepath",
                visible=True,
                interactive=True,
                show_download_button=True,
            )

### Screenshot

<img width="591" alt="image" src="https://github.com/gradio-app/gradio/assets/126974970/c8575f7f-adbc-43fb-9a02-ca449df139bb">

### System Info

```shell
Successfully installed gradio-4.12.0 gradio-client-0.8.0
```

### Severity

I can work around it

abidlabs:
It seems like there may have been a regression here, cc @hannahblair?

abidlabs:
Ah thanks for clarifying @hannahblair. Is it feasible to add the download button to the interactive Audio component? I think its useful as in some cases, an interactive Audio component may serve as both an input and an output, and so it would make sense for a user to be able to download the audio even if `interactive=True`

abidlabs:
Let's do it

ngaudemet:
thanks a lot. it is also useful when a user records an audio with their mic, and simply wants to download the audio.

PR #7104, opened 2024-01-22
Allow download button for interactive Audio and Video components

hannahblair:
## Description

This PR allows showing the download button for interactive Audio and Video components. I've modified the `<ModifyUpload />` component to do this as that seems to be the cleanest approach toward adding the download button (though downloading isn't actually _modifying_ an upload) 

Added storybook tests to show examples 🧪 

We also had the `show_download_button` logic in place in the frontend but wasn't a param in the backend, so this PR adds that param to `video.py`. 

Closes: #7088 
Closes: #7093

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/test/test_components.py
+++ b/test/test_components.py
@@ -860,7 +860,7 @@ class TestAudio:
         assert audio_output.get_config() == {
             "autoplay": False,
             "name": "audio",
-            "show_download_button": False,
+            "show_download_button": None,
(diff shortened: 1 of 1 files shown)


dawoodkhan82 (dismissed):
Nice, works great. Although it seems like the download button is showing by default for Audio input components.

PR #7080, opened 2024-01-19
start cc docs guide

pngwn:
## Description

Please include a concise summary, in clear English, of the changes in this pull request. If it closes an issue, please mention it here.

Closes: #(issue)

## 🎯 PRs Should Target Issues

Before your create a PR, please check to see if there is [an existing issue](https://github.com/gradio-app/gradio/issues) for this change. If not, please create an issue before you create this PR, unless the fix is very small. 

Not adhering to this guideline will result in the PR being closed. 

## Tests

1. PRs will only be merged if tests pass on CI. To run the tests locally, please set up [your Gradio environment locally](https://github.com/gradio-app/gradio/blob/main/CONTRIBUTING.md) and run the tests: `bash scripts/run_all_tests.sh`

2. You may need to run the linters: `bash scripts/format_backend.sh` and `bash scripts/format_frontend.sh`

PR #7109, opened 2024-01-22
generate docs when running `gradio cc build`

pngwn:
## Description

Runs the docs generator as part of the build. This only provides defaults with no options (so as not to bloat the build command). If you users wish to customise the behaviour, they can use the `docs` command.

Added new arg to `build` to optionally disbale the docs gen:

```bash
gradio cc build --no-generate-docs
```

Related to #7080.

Closes: #(issue)

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/gradio/cli/commands/components/docs.py
+++ b/gradio/cli/commands/components/docs.py
@@ -83,20 +83,20 @@ def _docs(
         if not isinstance(name, str):
             raise ValueError("Name not found in pyproject.toml")
 
-    run_command(
-        live=live,
-        name=name,
-        suppress_demo_check=suppress_demo_check,
-        pyproject_toml=data,
-        generate_space=generate_space,
-        generate_readme=generate_readme,
-        type_mode="simple",
-        _demo_path=_demo_path,
-        _demo_dir=_demo_dir,
-        _readme_path=_readme_path,
-        space_url=space_url,
-        _component_dir=_component_dir,
-    )
+        run_command(
+            live=live,
+            name=name,
+            suppress_demo_check=suppress_demo_check,
+            pyproject_toml=data,
+            generate_space=generate_space,
+            generate_readme=generate_readme,
+            type_mode="simple",
+            _demo_path=_demo_path,
+            _demo_dir=_demo_dir,
+            _readme_path=_readme_path,
+            space_url=space_url,
+            _component_dir=_component_dir,
+        )
 
 
 def run_command(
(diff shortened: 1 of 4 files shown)


gradio/cli/commands/components/docs.py
```diff
@@ -77,67 +77,118 @@ def _docs(
...
-        with open(_demo_path) as f:
-            demo = f.read()
 
         name = get_deep(data, ["project", "name"])
 
         if not isinstance(name, str):
             raise ValueError("Name not found in pyproject.toml")
 
-        pypi_exists = requests.get(f"https://pypi.org/pypi/{name}/json").status_code
+    run_command(
```
freddyaboulton, line 86: The warning about python 3.9 needs to be printed with `live` otherwise it's out of order:

image

pngwn: I kinda wanted it to be not in the live panel because there isn't actually an issue with the doc gen, just a note. But I can put it at the bottom with some space. 

Will update the message.

pngwn:
@freddyaboulton fixed (i think)!

gradio/cli/commands/components/docs.py
freddyaboulton: Looks great! My bad, I think we should remove the backticks around --no-generate-docs since they don't change. how it renders

freddyaboulton (approved):
Thanks @pngwn !

pngwn:
done!

gradio/cli/commands/components/docs.py
```diff
@@ -77,69 +77,114 @@ def _docs(
...
+            f.write(readme)
+            if not simple:
                 live.update(
                     f":white_check_mark: README generated in [orange3]{_readme_path}[/]"
                 )
+    if simple:
+        short_readme_path = Path(_readme_path).relative_to(_component_dir)
+        short_demo_path = Path(_demo_path).relative_to(_component_dir)
+        live.update(
+            f":white_check_mark: Documention generated in [orange3]{short_demo_path}[/] and [orange3]{short_readme_path}[/]. Pass --no-generate-docs to disable auto documentation."
```
freddyaboulton, line 184: This should be `space.py` not `demo.py` no?

image

pngwn: fixed!

pngwn:
I've also tweaked the `ParamViewer` so that it does not show a toggle if there is no content.

Issue #7095, opened 2024-01-21
gradio cc dev one CPU core full usage


velaia:
### Describe the bug

I've noticed my MacBook battery draining rapidly while working on a keypoint detection component for gradio. I noticed that there's a Python process that is at 100%.

Under the project issues I found #7086 and #6847 and not understanding the code + project enough I checked whether upgrade to 4.15.0 or addition of the time.sleep(0.001) from #7086 in the corresponding client.py would fix my issue but it didn't.

Under the site-packages I searched for "while True" and had a quick shot at the one occurrence under gradio/cli/commands/components/dev.py but found this was not related.

Maybe someone with more understanding of the code can reproduce or give me a tip on where to look.

I've also found that I can kill the 100% Python process in the Activity Monitor and the "gradio cc dev" will still run.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

When starting to develop a new component. Can easily be reproduced

```
gradio cc create testcomponent --template Image
cd testcomponent
gradio cc dev
```

### System Info

```shell
macOs: 14.1.1 (23B81)
Python 3.12.1
gradio                    4.15.0
gradio_client             0.8.1
```

### Severity

I can work around it

velaia:
I've checked the process ID and started it using the profiler

```
python -m cProfile -o prof1.prof /Users/user/git/gradio_dev/.venv/bin/gradio demo/app.py --watch-dirs /Users/daniel/git/gradio_dev/testcomponent
```

The prof1.prof (renamed to txt so I could upload it) is attached to this comment. [prof1.txt](https://github.com/gradio-app/gradio/files/14002035/prof1.txt)

velaia:
So the profiling file is huge (1.8 MB), even if I only run the program for a couple of seconds. The snakeviz tool to visualise the profiling session hits a limit at a depth of 15 and I have to set the cutoff to 1/100 to even get there. It seems to me that there's an issue with recursion.

<img width="1504" alt="Screenshot 2024-01-22 at 09 39 44" src="https://github.com/gradio-app/gradio/assets/1515904/223b66a3-adad-41f7-9fff-576e185a20e1">

At the same time I've found that the issue occurs even if I run it on the main directory of my project, not the custom-component subdirectory. So it seemingly doesn't have anything to do with the custom-components code.
The gradio interface in the app.py file is from the (beginner) tutorial I believe:

```
import gradio as gr

def greet(name, intensity):
[...]

PR #7113, opened 2024-01-22
Reduce CPU usage of dev mode

freddyaboulton:
## Description

Closes: #7095

We're watching the files of the gradio source code in addition to the demo. This increases the number of files that are polled for changes by an order of magnitude and most users are not installing gradio from source and making changes to the source code. 

Fix:
- Only watch the gradio source code if installed in edit mode from source. There isn't a foolproof way to determine this but I have implemented a heuristic that should work most of the time. This change will only be "breaking" users who install gradio from pypi and who edit the source code in the `site-packages` directory. I think that's a small proportion of users and they can always do an editable install to get around this.
- Introduce a short sleep between watch iterations. I'm using 0.05 seconds which should be fast enough so that users can't notice while freeing up the CPU core running the watch thread a bit.

There are other watch file algorithms we can use (https://pypi.org/project/watchfiles/) that are potentially more efficient but they would require installing other dependencies. With this change CPU usage is about 2% which I think is ok.

### utilization in this PR
Screenshot 2024-01-22 at 6 21 27 PM

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/petite-islands-follow.md
+++ b/.changeset/petite-islands-follow.md
@@ -1,0 +1,5 @@
+---
+"gradio": patch
+---
+
+fix:Reduce CPU usage of dev mode
--- a/gradio/cli/commands/reload.py
+++ b/gradio/cli/commands/reload.py
@@ -10,6 +10,7 @@ from __future__ import annotations
 import inspect
 import os
 import re
+import site
 import subprocess
 import sys
 import threading
@@ -63,9 +64,14 @@ def _setup_config(
 
     watching_dirs = []
     if str(gradio_folder).strip():
-        watching_dirs.append(gradio_folder)
-        message += f" '{gradio_folder}'"
-        message_change_count += 1
+        package_install = any(
+            utils.is_in_or_equal(gradio_folder, d) for d in site.getsitepackages()
+        )
+        if not package_install:
+            # This is a source install
+            watching_dirs.append(gradio_folder)
+            message += f" '{gradio_folder}'"
+            message_change_count += 1
 
     abs_parent = abs_original_path.parent
     if str(abs_parent).strip():
--- a/gradio/utils.py
+++ b/gradio/utils.py
@@ -14,6 +14,7 @@ import os
 import pkgutil
 import re
 import threading
+import time
 import traceback
 import typing
 import urllib.parse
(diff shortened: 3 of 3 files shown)


abidlabs:
Thanks @freddyaboulton! Changes look good to me, but here's the performance I'm getting:

* At baseline, python consumes ~1% of CPU (some background python processes)
* On `main`, running `gradio cc dev`, python consumes ~98% of CPU
* On this branch, running `gradio cc dev`, python consumes ~60% of CPU

I assume that this is because I've installed from source. I haven't been able to test what happens when you install from pypi since this is not published yet. Is this is consistent with what you expect?

freddyaboulton:
Yea I was seeing 50-60% if you do an editable install. You can test what would happen on a pypi install by omitting the `-e` flag when installing from source. So `pip install <gradio-dir>` will place it in `site-packages` and dev mode will no longer watch the gradio lib source.

if you run dev mode in a directory that has hundreds of python files then the CPU usage will be high. That's expected though and that should be the small minority of cases though.

abidlabs:
Ah I see. Well I tested installing from source (but omitting the `-e` flag) and I still get around the same CPU usage: ~50-60% when running `gradio cc dev` in a brand new component directory. Increasing the timeout to 0.1 seconds brings the CPU usage down to 30% but at that point performance of reload mode is degrading. So this PR generally looks good to me (its certainly better than what we have now) but perhaps we can do a quick sync to figure out why there's a discrepancy between the cpu usage numbers we're seeing @freddyaboulton

freddyaboulton:

I looked into it and I understand the discrepancy. I was getting low usage rate running dev mode in a directory with only two python files. If you run dev mode in a custom component directory, lots of files get created, e.g. `frontend`, `frontend/node_modules`. The way dev mode works is that we recursively watch all python and css files in the directory from which its called. So we end up iterating through files in the frontend directory (of which there are many because of `node_modules`). This is technically unnecessary since a separate process is watching the frontend directory but I'm not sure how we can change this without introducing a breaking change to reload mode. 

The current change is an improvement and the CPU utilization we see now in cc dev mode should be ok for the large majority of machines imo since machines are multicore.

abidlabs:
Can we skip watching the `frontend` folder altogether? We shouldn't need the Python reloader to watch those since the frontend server is already watching them.

abidlabs (approved):
Sounds reasonable to me. Thanks @freddyaboulton!

freddyaboulton:
Thanks @abidlabs !

PR #7116, opened 2024-01-23
Document the `gr.ParamViewer` component, and fix component preprocessing/postprocessing docstrings

abidlabs:
This PR does a few separate things:

**ParamViewer**

* Documents the `gr.ParamViewer` component and adds a component demo
* Fixes the typing for `gr.ParamViewer`

**General docstrings**

* Removes "queue must be enabled" from docstrings as the queue is now already enabled by default in Gradio 4.x
* Moves the preprocessing/postprocessing information from the class to the respective methods
* Removes "Examples-format" as examples are now in the same format as postprocessing (mentions this in the docs for examples)
* Improves the website UI for displaying the preprocessing/postprocessing information (thanks @aliabd)
* Fixes many incorrect docstrings 😓


gradio/components/paramviewer.py
```diff
@@ -48,26 +59,20 @@ def __init__(
...
+    def preprocess(self, payload: dict[str, Parameter]) -> dict[str, Parameter]:
         return payload
 
-    def postprocess(self, value: list[Parameter]) -> list[Parameter]:
-        """
-        Parameters:
-            value: A list of dictionaries with keys "type", "description", and "default" for each parameter.
-        Returns:
-            A list of dictionaries with keys "type", "description", and "default" for each parameter.
-        """
```
pngwn, line 66: We should add these docstrings back. I'm not against having this info elsewhere as well, but the docs generator pulls this information from the docstring for the pre and postprocessor return and inputs. 

This is also the correct place for this information from a maintenance PoV rather than in the class docstring.

We should also ensure docstrings in other components are consistent with this pattern, so that cloning templates gives a good experience.

aliabd: I agree that preprocess/postprocess functions are the right place for this to be documented. I could change the website to read from there instead. I think the reason we started doing it in the class docstring was because we show it on the website as one sentence.

Screen Shot 2024-01-23 at 11 42 06 PM

So its easier for the website to read this sentence from the docstrings instead of forming it from param descriptions

abidlabs: Yes that sounds good, go ahead and push directly to this PR @aliabd 

I think we could write this as:

```
Supported data types:

As input: [the "return" part of the docstring of "preprocess()"]
As output: [the docstring for "value" in "postprocess()"]
```

abidlabs:
@aliabd I think some of the changes here made the font size of any code in the docs pages really small, e.g. 

image

aliabd (approved):
Made the changes @abidlabs and now everything looks good! Take another scan and if it looks good to you feel free to merge

abidlabs:
LGTM thanks @aliabd for the fixes and review!

image

Issue #5392, opened 2023-08-31
White background appears while a Gradio app is loading even in dark mode


abidlabs:
### Describe the bug

While a Gradio app is loading, the loading indicator appears on top of a white background. This is particularly disconcerting when a Gradio app is embedded in another page that has dark theme enabled, e.g. 

https://huggingface.co/spaces/multimodalart/LoraTheExplorer

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

Visit https://huggingface.co/spaces/multimodalart/LoraTheExplorer

### System Info

```shell
sdk_version: 3.39.0
```

### Severity

I can work around it

aditya-azad:
I cannot reproduce the issue.

![image](https://github.com/gradio-app/gradio/assets/26451004/35841cec-f23d-4c7b-8660-a8892855a264)

abidlabs:
Thanks for looking into this @aditya-azad! The problem is still there -- it happens when the app is initially loading. There's a brief second where the background is black but the part where the gradio app is supposed to be is white:

<img width="1343" alt="image" src="https://github.com/gradio-app/gradio/assets/1778297/a4aeb129-b58a-40f5-b0e4-88fe88d7e142">

gary149:
Up 👍  if anything is possible as the issue still persists.

abidlabs:
Reopening the issue: https://github.com/gradio-app/gradio/pull/7117#issuecomment-1984146237

gary149:
Up if anyone can help 👍

abidlabs:
This is much less of an issue now since with SSR, Spaces load much faster. Feel free to ping here @gary149 if you still see this

abidlabs:
It looks like we fixed this for non-SSR Spaces:


https://github.com/user-attachments/assets/42719ed8-1c94-467b-a111-0f14f4646959

but not for SSR Spaces, e.g. see: https://huggingface.co/spaces/abidlabs/test-dark

https://github.com/user-attachments/assets/f57c42a8-42b2-4a09-a6af-d198cbce4492

PR #7117, opened 2024-01-23
add background color based on the OS mode

aileenvl:
## Description

There is no background applied to the body and it defaults to white with a function to detect the users os mode we can change that background before gradio loads. there might be a better way to implement but just a draft idea.

Closes: #5392 

## 🎯 PRs Should Target Issues

## Tests


pngwn:
Hey @aileenvl! Thanks for this PR; this is a great idea! 

I have just tweaked it a little so instead of using JS to set the color, we use static CSS instead. This way we can guarantee we never have any flashes by putting it right at the top of the head. This did require some changes to the build config but seems to be working now. I also added a default font until our custom fonts load in. 

There is no flashing now and the font switch is pretty subtle.

Thanks again!

js/app/index.html
```diff
@@ -16,6 +16,21 @@
 			content="width=device-width, initial-scale=1, shrink-to-fit=no"
 		/>
 
+		<style>
+			body {
+				background: white;
+				color: black;
+				font-family: Arial, Helvetica, sans-serif;
```
pngwn, line 23: Prevents the janky switch from a default serif to our sans-serif.

js/app/vite.config.ts
```diff
@@ -115,6 +115,7 @@ export default defineConfig(({ mode }) => {
 							} else if (
 								selector.indexOf(":root") > -1 ||
 								selector.indexOf("dark") > -1 ||
+								selector.indexOf("body") > -1 ||
```
pngwn, line 118: We don't want to add a prefix to `body` selectors, that breaks them.

aliabid94:
Is there any we we can use the theme variables, `--body-background-fill` `--body-background-fill-dark`, `body-text-color`, etc?

pngwn:
We can but we would have to define it right alongside the definition (which seems pointless to me at least) because the styles where we define those things haven't loaded yet, they get loaded in by the gradio app and this flash happens because of that delay.

aliabid94:
but now there will be a flash if those colors differ, correct? 
would it be overkill to pass those theme colors directly to the index.html template at render time so there' sno flash at all?

aliabid94:
e.g. 
```
		<style>
			body {
				background: {{ config['theme-bg'] }};
				color: {{ config['theme-color'] }};
				font-family: Arial, Helvetica, sans-serif;
			}

			@media (prefers-color-scheme: dark) {
				body {
    		         		background: {{ config['theme-bg-dark'] }};
			        	color: {{ config['theme-color-dark'] }};
				}
			}
		</style>
```

might be unnecessary

pngwn:
> but now there will be a flash if those colors differ, correct?

Yes, if they get out of sync.

> would it be overkill to pass those theme colors directly to the index.html template at render time so there's no flash at all?

No, i think that makes perfect sense, would be ideal.

Will take a look!

aliabid94:
I can take a crack at it if its late and you can review in the morning?

pngwn:
@aliabid94 Sounds good! Thansk!

pngwn:
Feel free to merge if you get it working and want to get the release out. The best way to test is just to add some network throttling with the network tab open and there should be no color changes. You'll know where the color flash _should_ come in because you'll see the font change (it is subtle in this branch), it tends to happen around the same time.

aliabid94:
I couldn't figure out how to pass the config values to the CSS, because:
```
		<style>
			body {
				background: {{ config['body_css']['body_background_fill'] }};
				color: {{ config['body_css']['body_text_color'] }};
				font-family: Arial, Helvetica, sans-serif;
			}

			@media (prefers-color-scheme: dark) {
				body {
					background: {{ config['body_css']['body_background_fill_dark'] }};
					color: {{ config['body_css']['body_text_color_dark'] }};
				}
			}
		</style>
```

would cause the compiler to fail parsing the CSS. Can't figure out how to embed the tempate tags in CSS since they're not quote enclosed

pngwn:
@aliabid94 which compiler? The vite build or the templating compiler?

aliabid94:
vite

```
11:07:36 AM [vite-plugin-svelte] you are building for production but compilerOptions.dev is true, forcing it to false
vite v4.3.9 building for production...
✓ 2 modules transformed.
✓ built in 53ms
[vite:css] [postcss] /Users/aliabid/Desktop/gradio/js/app/index.html?html-proxy&inline-css&index=0.css:3:20: Unknown word
file: /Users/aliabid/Desktop/gradio/js/app/index.html?html-proxy&inline-css&index=0.css:3:20
error during build:
CssSyntaxError: [postcss] /Users/aliabid/Desktop/gradio/js/app/index.html?html-proxy&inline-css&index=0.css:3:20: Unknown word
    at Input.error (/Users/aliabid/Desktop/gradio/node_modules/.pnpm/postcss@8.4.27/node_modules/postcss/lib/input.js:106:16)
    at Parser.unknownWord (/Users/aliabid/Desktop/gradio/node_modules/.pnpm/postcss@8.4.27/node_modules/postcss/lib/parser.js:587:22)
    at Parser.other (/Users/aliabid/Desktop/gradio/node_modules/.pnpm/postcss@8.4.27/node_modules/postcss/lib/parser.js:429:12)
    at Parser.parse (/Users/aliabid/Desktop/gradio/node_modules/.pnpm/postcss@8.4.27/node_modules/postcss/lib/parser.js:464:16)
    at parse (/Users/aliabid/Desktop/gradio/node_modules/.pnpm/postcss@8.4.27/node_modules/postcss/lib/parse.js:11:12)
    at new LazyResult (/Users/aliabid/Desktop/gradio/node_modules/.pnpm/postcss@8.4.27/node_modules/postcss/lib/lazy-result.js:133:16)
    at Processor.process (/Users/aliabid/Desktop/gradio/node_modules/.pnpm/postcss@8.4.27/node_modules/postcss/lib/processor.js:53:14)
    at compileCSS (file:///Users/aliabid/Desktop/gradio/node_modules/.pnpm/vite@4.3.9_@types+node@20.3.2_less@4.2.0/node_modules/vite/dist/node/chunks/dep-e8f070e8.js:38481:63)
    at async Object.transform (file:///Users/aliabid/Desktop/gradio/node_modules/.pnpm/vite@4.3.9_@types+node@20.3.2_less@4.2.0/node_modules/vite/dist/node/chunks/dep-e8f070e8.js:37917:56)
    at async transform (file:///Users/aliabid/Desktop/gradio/node_modules/.pnpm/rollup@3.29.4/node_modules/rollup/dist/es/shared/node-entry.js:24449:16)
/Users/aliabid/Desktop/gradio/js/app:
 ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL 
```

aliabid94:
ready for review

gradio/themes/base.py
```diff
@@ -93,6 +94,30 @@ def repl_func(match):
 
         return f"{css_code}\n{dark_css_code}"
 
+    def _get_computed_value(self, property: str, depth=0) -> str:
+        MAX_DEPTH = 100
+        if depth > MAX_DEPTH:
```
aliabid94, line 99: don't want to raise an exception in case there were "broken" themes with circular references before. They didn't error out before and don't want a breaking change.

abidlabs:
Tested locally and works great! Thanks everyone ❤️

pngwn (approved):
Looks (and works) great! Thanks @aileenvl @aliabid94!

freddyaboulton (approved):

abidlabs:
@aliabid94 re: the tests, I ran them all locally and they passed, so I'll go ahead and merge this in.

cc @gary149

gary149:
fyi there's still a white background while loading on Spaces in dark mode:

image

Issue #7078, opened 2024-01-19
exception leads to elements being disabled, e.g. button is disabled, in gradio 4 unlike in gradio 3


pseudotensor:
### Describe the bug

See repro.

I expect that elements do not get disabled or I can avoid that aspect, since just because output to the element was not possible does not mean it is a bad element and should be automatically disabled.

E.g. this case is MRE of a case where I fail the upload only after uploading the file, then checking in .then if they are allowed to add a file to a database.  I raise if it's not allowed, but this disabled the button.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr


def upload_file(files):
    file_paths = [file.name for file in files]
    return file_paths


with gr.Blocks() as demo:
    file_output = gr.File()
    upload_button = gr.UploadButton("Click to Upload a File", file_types=["image", "video", ".pdf"], file_count="multiple")
    event = upload_button.upload(upload_file, upload_button, file_output)

    def check(x):
        if '.png' in str(x):
            raise ValueError("Cannot handle png")
        if '.jpg' in str(x):
            raise ValueError("Cannot handle jpg")
    event.then(check, inputs=upload_button, outputs=upload_button)

demo.launch()



```

### Screenshot

![image](https://github.com/gradio-app/gradio/assets/2249614/3d3bbd97-2609-499f-83d1-b7851a03700b)

### System Info

```shell
gradio 4.15.0
```

### Severity

Blocking usage of gradio

PR #7125, opened 2024-01-23
un-disable output components after exception is raised

abidlabs:
This fixes a regression identified by @pseudotensor, closes: https://github.com/gradio-app/gradio/issues/7078

Test code:

```py
import gradio as gr

def upload_file(files):
    file_paths = [file.name for file in files]
    return file_paths

with gr.Blocks() as demo:
    file_output = gr.File()
    upload_button = gr.UploadButton("Click to Upload a File", file_types=["image", "video", ".pdf"], file_count="multiple")
    event = upload_button.upload(upload_file, upload_button, file_output)

    def check(x):
        raise ValueError("Cannot handle jpg")
        
    event.then(check, inputs=upload_button, outputs=upload_button)

demo.launch()
```

### DIFF
--- a/.changeset/lemon-tips-switch.md
+++ b/.changeset/lemon-tips-switch.md
@@ -1,0 +1,6 @@
+---
+"@gradio/app": minor
+"gradio": minor
+---
+
+feat:re-enable output components after exception is raised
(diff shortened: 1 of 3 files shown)


freddyaboulton (approved):
LGTM thanks @abidlabs

Issue #6869, opened 2023-12-22
Returning `None` to a `gr.Button` causes app to hang


abidlabs:
### Describe the bug

A function that returns `None` to a `gr.Button` causes the app to hang (reported by @hysts)

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr


def fn(text):
    return '', None


with gr.Blocks() as demo:
    text = gr.Text()
    btn = gr.Button()
    text.submit(fn=fn, inputs=text, outputs=[text, btn])
demo.queue().launch()
```

### System Info

```shell
4.11.0
```

### Severity

I can work around it

PR #7126, opened 2024-01-23
Allow buttons to take null value

abidlabs:
Fixes: https://github.com/gradio-app/gradio/issues/6869

Test code in original issue


js/utils/src/utils.ts
```diff
@@ -151,11 +151,6 @@ async function copy_to_clipboard(value: string): Promise<boolean> {
 	return copied;
 }
 
-// import { format } from "svelte-i18n";
```
freddyaboulton, line 154: Is this intentional?

freddyaboulton (approved):
Looks good @abidlabs ! Can we unit test this?

js/utils/src/utils.ts
abidlabs: it seemed like an unused comment so I removed it. am I missing something? cc @hannahblair just in case

pngwn (approved):
Looks great! Thanks @abidlabs!

abidlabs:
Thanks folks for the review! I added a story to catch regressions.

Issue #6578, opened 2023-11-25
Inconsistent dimensions for Audio input


gau-nernst:
### Describe the bug

Depending on the file, audio input might have shape `time, n_channels` or just `time`. This is inconsistent and introduces bugs to the code when one would expect the audio to have certain shape.

The documentation also does not mention anything about the expected shape.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

def predict(audio):
    sr, x = audio
    return f"{x.shape} {sr}"

gr.Interface(fn=predict, inputs="audio", outputs="text").launch()
```

With microphone input: (74880,) 48000
With mono audio (e.g. [here](https://www2.cs.uic.edu/~i101/SoundFiles/CantinaBand3.wav)): (66150,) 22050
With stereo audio (e.g. [here](https://onlinetestcase.com/wp-content/uploads/2023/06/500-KB-WAV.wav)): (128180, 2) 44100

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Darwin
gradio version: 4.7.1
gradio_client version: 0.7.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.1.2
fastapi: 0.104.1
ffmpy: 0.3.1
gradio-client==0.7.0 is not installed.
httpx: 0.25.1
huggingface-hub: 0.16.4
importlib-resources: 6.0.1
jinja2: 3.1.2
markupsafe: 2.1.2
matplotlib: 3.6.2
numpy: 1.23.1
orjson: 3.9.10
packaging: 23.0
pandas: 2.0.3
pillow: 9.2.0
pydantic: 2.5.2
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0
requests: 2.31.0
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.9.0
typing-extensions: 4.8.0
uvicorn: 0.24.0.post1
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2023.6.0
httpx: 0.25.1
huggingface-hub: 0.16.4
packaging: 23.0
requests: 2.31.0
typing-extensions: 4.8.0
websockets: 11.0.3
```

### Severity

I can work around it

Issue #5365, opened 2023-08-29
_js is not controlled by ``every'' argument 


chenjoya:
### Describe the bug

I am not sure if it is a bug. I find _js is not controlled by every argument: 

```python
with gr.Blocks() as app:
    dummy = gr.JSON({}, visible=False)
    app.load(fn=lambda x:print(x), every=1, inputs=[dummy], _js="() => Math.random()")
app.queue()
app.launch(server_port=8899, debug=True, share=True)
```

If we run this, every time we get the same number 
```
0.14488436985087927
0.14488436985087927
0.14488436985087927
0.14488436985087927
0.14488436985087927
0.14488436985087927
0.14488436985087927
```

But I need maybe continuously js random value. Is there any method to run js code and control by every=1?

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

Follow the code in **Describe the bug**.

### System Info

```shell
gradio==3.41.2
```

### Severity

Blocking usage of gradio

abidlabs:
Hi @chenjoya this is intentional, the `js` function is supposed to run once, before the main function is run, which is controlled by the `every` parameter. But we should document this more clearly

chenjoya:
Thank you so much! ;) 

Do you have elegant implementation method to achieve the streaming functionality, ie if there is a video playing in js, how to get its current time and streamingly output in a chatbot box? We recently want to show such a model. 

I implemented one by a json dictionary to receive the metadata of the video, and use .change method to refresh chatbox. But it may not be optimal.

Wish you merry Christmas in advance!

Issue #4665, opened 2023-06-24
Docs for Session state don't (clearly) use Session State


drscotthawley:
### Describe the bug

In documentation for Interface Session state,  https://gradio.app/interface-state/#session-state, 
talks about setting the `state` variable of `Interface`, e.g.,

> add the 'state' input and 'state' output components when creating your Interface

however in the code that follows, neither "Interface" nor "state" appear at all.  In fact they use `with gr.Blocks() as demo:` with no keywords. 

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

Try to read the documentation on Interface Session State. 

### Screenshot

![2023-06-24_16-58](https://github.com/gradio-app/gradio/assets/13925685/1e12bf14-008b-4555-a635-da163ed2641d)

### System Info

```shell
https://status.gradio.app/
```

### Severity

Blocking usage of gradio

PR #7131, opened 2024-01-24
Miscellaneous doc fixes

abidlabs:
* Improve documentation for `gr.Audio` component, particularly the shape of the audio data array. Closes: https://github.com/gradio-app/gradio/issues/6578
* Clarify that `every` only applies to backend functions, not `js` functions. Closes: https://github.com/gradio-app/gradio/issues/5365
* Fix the example for Interface state and improve the docs there generally. Closes: https://github.com/gradio-app/gradio/issues/4665


gradio/components/audio.py
```diff
@@ -49,7 +49,7 @@ class Audio(
 ):
     """
     Creates an audio component that can be used to upload/record audio (as an input) or display audio (as an output).
-    Preprocessing: passes the uploaded audio as a {Tuple(int, numpy.array)} corresponding to (sample rate in Hz, audio data as a 16-bit int array whose values range from -32768 to 32767), or as a {str} filepath, depending on `type`.
+    Preprocessing: depending on `type`, passes the uploaded audio as {str} filepath or a {Tuple(int, numpy.array)} corresponding to (sample rate in Hz, audio data). If the latter, the audio data is a 16-bit int array whose values range from -32768 to 32767 and shape of the audio data array is (samples,) for mono audio or (samples, channels) for multi-channel audio.
```
aliabd, line 52:
```suggestion
    Preprocessing: depending on `type`, passes the uploaded audio as {str} filepath or a {Tuple(int, numpy.array)} corresponding to (sample rate in Hz, audio data). If the latter, the audio data is a 16-bit int array whose values range from -32768 to 32767 and shape of the audio data array is {(samples,)} for mono audio or {(samples, channels)} for multi-channel audio.
```

aliabd (approved):
Thanks for the fixes @abidlabs They all lgtm! Consider adding the new `interface_state` demo as a demo in the State class docstring. so it shows up in the docs too.

gradio/components/audio.py
abidlabs: Why the curly brackets?

aliabd: Adds formatting that makes it easier to parse the text. Feel free to ignore

Screen Shot 2024-01-24 at 11 00 35 AM

Screen Shot 2024-01-24 at 11 03 35 AM

abidlabs: I thought we only used that to designate Python types

abidlabs:
Thanks @aliabd! I’ll add the to docs

Issue #7094, opened 2024-01-21
gr.File only supports Drag and Drop for types of the first element in file_types


tradoff:
### Describe the bug

gr.File( file_types = [ 'image', 'video' ] ) only supports Drag and Drop for image file but not video file. (However, clicking interface to select file to upload is both support for the two. )
gr.File( file_types = [ 'file' ] ) can support all types to Drag and Drop but...

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr
def generate_file(file_obj):
    return None
if __name__ == "__main__":
    inputs = gr.components.File( file_types = [ 'image', 'video' ] )
    outputs = gr.components.File()
    demo = gr.Interface(fn=generate_file, inputs=inputs, outputs=outputs)
    demo.launch(share=True)
```

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Linux
gradio version: 4.15.0
gradio_client version: 0.8.1

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.2.0
fastapi: 0.109.0
ffmpy: 0.3.1
gradio-client==0.8.1 is not installed.
httpx: 0.26.0
huggingface-hub: 0.20.2
importlib-resources: 6.1.1
jinja2: 3.1.3
markupsafe: 2.1.3
matplotlib: 3.8.2
numpy: 1.26.2
orjson: 3.9.10
packaging: 23.2
pandas: 2.1.4
pillow: 10.2.0
pydantic: 2.5.3
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0.1
ruff: 0.1.14
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.9.0
typing-extensions: 4.9.0
uvicorn: 0.25.0
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.

gradio_client dependencies in your environment:

fsspec: 2023.12.2
httpx: 0.26.0
huggingface-hub: 0.20.2
packaging: 23.2
typing-extensions: 4.9.0
websockets: 11.0.3
```

### Severity

I can work around it

abidlabs:
I thought we fixed this? @dawoodkhan82 can we add some functional tests to prevent this issue from reoccurring

dawoodkhan82:
[...]

Issue #7132, opened 2024-01-24
Uploaded file not showing in interface in 4.15.0


snakeninny:
### Describe the bug

Uploaded file not showing in interface, as you can see in the screenshot video.
The weird is that it's reproducible in my Ubuntu server, but not in my macbook air

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr
import os

def test(image_file):
    if image_file is None:
        raise gr.Error(f"Forgot to upload?")

    return image_file

demo = gr.Interface(
    title="Test",
    fn=test,
    inputs=[
        gr.File(file_types=["image"], file_count="single"),
    ],
    outputs=[
        gr.File(file_types=["image"], file_count="single"),
    ],
    allow_flagging="never",
    examples=[
        [os.path.join(os.path.dirname(__file__), "男.png")],
    ]
)

gr_version = gr.__version__
print("Gradio version:", gr_version)
demo.launch(debug=True, server_name="0.0.0.0")
```

### Screenshot

https://github.com/gradio-app/gradio/assets/641869/faf9a51f-bf9c-4027-b264-a093dc0d2995

### Logs

```shell
There's no error and no log
```

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Linux
gradio version: 4.15.0
gradio_client version: 0.8.1

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.2.0
fastapi: 0.109.0
ffmpy: 0.3.1
gradio-client==0.8.1 is not installed.
httpx: 0.26.0
huggingface-hub: 0.20.3
importlib-resources: 6.1.1
jinja2: 3.1.2
markupsafe: 2.1.3
matplotlib: 3.8.2
numpy: 1.26.3
orjson: 3.9.12
packaging: 23.2
pandas: 2.2.0
pillow: 10.2.0
pydantic: 2.5.3
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0.1
ruff: 0.1.14
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.9.0
typing-extensions: 4.9.0
uvicorn: 0.27.0
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2023.12.2
httpx: 0.26.0
huggingface-hub: 0.20.3
[...]

PR #7141, opened 2024-01-24
Few File component drag and drop

dawoodkhan82:
## Description
- Fixes Drag and Drop for types of the first element in file_types
- Added playwright playwright tests for drag & drop

Closes: #7094

## 🎯 PRs Should Target Issues

## Tests


js/file/shared/File.svelte
```diff
@@ -20,7 +20,7 @@
 	label={label || "File"}
 />
 
-{#if value && (Array.isArray(value) ? value.length > 0 : true)}
+{#if (Array.isArray(value) && value.length > 0) || value !== null}
```
abidlabs, line 23: #7132 has already been fixed so I don't think we need to make this change?

abidlabs:
@dawoodkhan82 #7132 is already fixed, so I've removed it from the parent comment, and I think we can reduce some of the changes in this PR as a result

js/file/shared/File.svelte
dawoodkhan82: Either is fine, but I can remove my change.

abidlabs: The original seems slightly simpler to parse and we've already validated it, so maybe keep the original? Not a big deal either way

abidlabs:
Thanks @dawoodkhan82. I tested this and can confirm that the original issue is fixed. However, the drag-and-drop doesn't seem to work with file extensions. Here's the repro:

```py
import gradio as gr
def generate_file(file_obj):
    return None

inputs = gr.components.File( file_types = [ '.jpg', '.mov' ] )
outputs = gr.components.File()
demo = gr.Interface(fn=generate_file, inputs=inputs, outputs=outputs)
demo.launch()
```

If you try dragging-and-dropping `.jpg` or `.mov` files, they won't get uploaded (however, if you click on the file component, they'll still work)

dawoodkhan82:
@abidlabs Should be fixed

abidlabs:
Extensions work but hmm now the original repro isn't working for me:

```py
import gradio as gr
def generate_file(file_obj):
    return None

inputs = gr.components.File( file_types = [ 'image', 'video' ] )
outputs = gr.components.File()
demo = gr.Interface(fn=generate_file, inputs=inputs, outputs=outputs)
demo.launch()
```

dawoodkhan82:
@abidlabs Ah yes I see what's going wrong. Need to pass both the file extension and file type of the uploaded file. Pushing a fix now.

dawoodkhan82:
@abidlabs Can you double check and make sure it works for you now?

abidlabs (approved):
Nice! Works for me now @dawoodkhan82 thanks

Issue #7137, opened 2024-01-24
Unable to upload pictures on iPhone mobile device


aiise117:
### Describe the bug

Clicking the image upload button opens the folder by default instead of the album library. Unable to recognize image formats in the folder directory, and also unable to click to upload them. I hope you will take note of this issue
![IMG_2066](https://github.com/gradio-app/gradio/assets/114309081/d227ff40-bc70-4667-8ed4-572ffd1fbfce)

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

```

### Screenshot

![IMG_2063](https://github.com/gradio-app/gradio/assets/114309081/0b94476d-26e5-404c-b2fb-254512d33775)
![IMG_2064](https://github.com/gradio-app/gradio/assets/114309081/98342a2b-7ae8-4e57-955d-00b9c1a1d090)
![IMG_2065](https://github.com/gradio-app/gradio/assets/114309081/829e6408-5dc6-4392-8d08-dfcb63a36cd4)
![IMG_2066](https://github.com/gradio-app/gradio/assets/114309081/b31c0d36-1f0b-4c46-ae52-378fc60cfd35)

### System Info

```shell
Yes
```

### Severity

Blocking usage of gradio

freddyaboulton:
Hi @aiise117 ! 

The reason that the photo album doesn't open might be because the demo is using `gr.File` instead of `gr.Image`. Will investigate why images in your files are not allowed to be uploaded.

PR #7142, opened 2024-01-24
Remove kwargs from template components

freddyaboulton:
## Description

Closes: #7137

The issue is that only props that are explicitly listed in the __init__ are saved to the config. For this reason we removed **kwargs from the main components. We kept them in the templates by mistake.

I removed kwargs from the templates and made sure every parameter listed in the parent component is also listed in the template component. Also added a unit test to prevent this in the future.

Repro from original issue

```python
import gradio as gr

  

with gr.Blocks() as demo:

    files = gr.Files(interactive=True, file_types=["image"])

  

demo.launch(share=True)
```

Go to share link from your phone and upload a photo. You should be able to select from photo gallery.

https://github.com/gradio-app/gradio/assets/41651716/575c0798-d593-4c47-839a-953a77d88503

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/gradio/templates.py
+++ b/gradio/templates.py
@@ -294,7 +294,7 @@ class PlayableVideo(components.Video):
         | Callable
         | None = None,
         *,
-        format: str | None = "mp4",
+        format: Literal["mp4"] = "mp4",
         sources: list[Literal["upload", "webcam"]] | None = None,
         height: int | str | None = None,
         width: int | str | None = None,
@@ -356,7 +356,7 @@ class Microphone(components.Audio):
         self,
         value: str | Path | tuple[int, np.ndarray] | Callable | None = None,
         *,
-        sources: list[Literal["microphone"]] | None = None,
+        sources: list[Literal["upload", "microphone"]] | None = None,
         type: Literal["numpy", "filepath"] = "numpy",
         label: str | None = None,
         every: float | None = None,
@@ -382,7 +382,7 @@ class Microphone(components.Audio):
         sources = ["microphone"]
         super().__init__(
             value,
-            sources=sources,  # type: ignore
+            sources=sources,
             type=type,
             label=label,
             every=every,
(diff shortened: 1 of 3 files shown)


gradio/templates.py
```diff
@@ -262,31 +288,60 @@ class PlayableVideo(components.Video):
...
+        value: str
+        | Path
+        | tuple[str | Path, str | Path | None]
+        | Callable
+        | None = None,
         *,
-        format: Literal["mp4"] | None = "mp4",
-        sources: list[Literal["upload", "webcam"]] = ["upload"],  # noqa: B006
+        format: str | None = "mp4",
+        sources: list[Literal["upload", "webcam"]] | None = None,
```
freddyaboulton, line 298: switching to none because it's bad practice to have mutable default arguments in constructors

test/test_components.py
```diff
@@ -2955,3 +2956,12 @@ def test_constructor_args():
...
     }
+
+
+def test_template_component_configs(io_components):
+    template_components = [c for c in io_components if getattr(c, "is_template", False)]
+    for component in template_components:
+        component_parent_class = inspect.getmro(component)[1]
+        template_config = component().get_config()
+        parent_config = component_parent_class().get_config()
+        assert set(parent_config.keys()).issubset(set(template_config.keys()))
```
abidlabs, line 2967: Why do we assert for subset and not that the keys are identical?

abidlabs:
Nice PR @freddyaboulton! LGTM, just one question above

abidlabs (approved):

test/test_components.py
freddyaboulton: Because the custom buttons (gr.LogInButton) are templates but they add some keys. I think checking a subset is ok. We just need to make sure that everything in the parent is represented in the child's config.

abidlabs: Ah I didn't realize those were considered templates. Ok sg!

freddyaboulton:
Thanks for the review @abidlabs !!!

PR #7148, opened 2024-01-25
Use Gallery as input component

freddyaboulton:
## Description

Closes: #1379

The Gallery component can now be used as an input! This comes with the following changes:

1. Adding `interactive` to the Gallery constructor
2. Adding `type` to the Gallery constructor. Can be `numpy`, `filepath` or `pil` like image. It controls the data type passed to the prediction function when the component is used as input.
3. Adding the `upload` and `change` events to the gallery. The upload event is only triggered after upload. The change event triggered when the user clears out the interactive gallery or when the value of an output gallery changes.

> The only potential issue I see is that the `preprocess` of the gallery component always strips the captions. This was done so that the input of a prediction function that takes a gallery is just a list of images. But there may be some users who want the captions if they exist so maybe we control that via another parameter?

I added a `keep_captions` parameter to determine whether the captions are kept in the data passed to the prediction function.

### Demo

gallery_as_input

## 🎯 PRs Should Target Issues

## Tests

5. You may need to run the linters: `bash scripts/format_backend.sh` and `bash scripts/format_frontend.sh`


.config/playwright-setup.js
```diff
@@ -62,7 +62,8 @@ function spawn_gradio_app(app, port, verbose) {
 			env: {
 				...process.env,
 				GRADIO_SERVER_PORT: `7879`,
-				PYTHONUNBUFFERED: "true"
+				PYTHONUNBUFFERED: "true",
+				GRADIO_ANALYTICS_ENABLED: "False"
```
freddyaboulton, line 66: lol i think playwright was counting towards analytics

demo/gallery_component/run.py
```diff
@@ -8,6 +8,16 @@
         "https://nationalzoo.si.edu/sites/default/files/animals/cheetah-002.jpg",
         "https://images.theconversation.com/files/375893/original/file-20201218-13-a8h8uq.jpg?ixlib=rb-1.1.0&rect=16%2C407%2C5515%2C2924&q=45&auto=format&w=496&fit=clip",
     ]
-    gr.Gallery(value=cheetahs, columns=4)
```
freddyaboulton, line 11: let me revert this. was a mistake

js/app/src/lang/en.json
```diff
@@ -112,6 +112,7 @@
 		"drop_file": "Drop File Here",
 		"drop_image": "Drop Image Here",
 		"drop_video": "Drop Video Here",
+		"drop_gallery": "Drop Image(s) Here",
```
freddyaboulton, line 115: I can update the other files with mistral/google pre-merge

gradio/components/gallery.py
```diff
@@ -177,10 +186,29 @@ def postprocess(
...
+            return img
+        else:
+            converted_image = _Image.open(img)
+            if type == "numpy":
+                converted_image = np.array(converted_image)
+            return converted_image
+
+    def preprocess(
+        self, payload: GalleryData | None
+    ) -> List[str | _Image.Image | np.ndarray] | None:
```
pngwn, line 201: I think this type is incorrect currently.

pngwn:
> The only potential issue I see is that the preprocess of the gallery component always strips the captions. This was done so that the input of a prediction function that takes a gallery is just a list of images. But there may be some users who want the captions if they exist so maybe we control that via another parameter?

I am pretty strongly against this.

We did this in the Image component and ended up regretting it, which is why we separated out  `Image` and `ImageEditor`, so we could have more appropriate signature for pre + post process. I don't think the shape of the processor signatures should change based on which arguments are passed. It is hard to reason about and even harder to document.

freddyaboulton:
I see your point but 99% of the time (I reckon) gradio developers will not want to deal with a tuple since the caption will be empty. Should we always strip captions? cc @abidlabs @aliabid94 ?

abidlabs:
A list of single-element tuples (images without captions) is not so bad to parse. As long we document it clearly, it should be fine.

gradio/components/gallery.py
```diff
@@ -95,6 +98,9 @@ def __init__(
             object_fit: CSS object-fit property for the thumbnail images in the gallery. Can be "contain", "cover", "fill", "none", or "scale-down".
             show_share_button: If True, will show a share icon in the corner of the component that allows user to share outputs to Hugging Face Spaces Discussions. If False, icon does not appear. If set to None (default behavior), then the icon appears if this Gradio app is launched on Spaces, but not otherwise.
             show_download_button: If True, will show a download button in the corner of the selected image. If False, the icon does not appear. Default is True.
+            interactive: If True, the gallery will be interactive, allowing the user upload images. If False, the gallery will be static. Default is True.
+            type: The format the image is converted before being passed into the prediction function. "numpy" converts the image to a numpy array with shape (height, width, 3) and values from 0 to 255, "pil" converts the image to a PIL image object, "filepath" passes a str path to a temporary file containing the image. If the image is SVG, the `type` is ignored and the filepath of the SVG is returned.
```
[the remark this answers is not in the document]
freddyaboulton: Good catch!

gradio/components/gallery.py
```diff
@@ -95,6 +98,9 @@ def __init__(
             object_fit: CSS object-fit property for the thumbnail images in the gallery. Can be "contain", "cover", "fill", "none", or "scale-down".
             show_share_button: If True, will show a share icon in the corner of the component that allows user to share outputs to Hugging Face Spaces Discussions. If False, icon does not appear. If set to None (default behavior), then the icon appears if this Gradio app is launched on Spaces, but not otherwise.
             show_download_button: If True, will show a download button in the corner of the selected image. If False, the icon does not appear. Default is True.
+            interactive: If True, the gallery will be interactive, allowing the user upload images. If False, the gallery will be static. Default is True.
```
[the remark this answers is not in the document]
freddyaboulton: Good catch!

freddyaboulton:
I removed `keep_captions` and we're always going to pass a tuple. Can I get a re-review?

js/app/test/gallery_component_events.spec.ts
```diff
@@ -29,3 +29,21 @@ test("Gallery select event returns the right value and the download button works
 	const download = await downloadPromise;
 	expect(download.suggestedFilename()).toBe("lite-logo.png");
 });
+
+test("Gallery allows click-to-upload and upload and change events work correctly", async ({
```
pngwn, line 33:
```suggestion
test("Gallery click-to-upload, upload and change events work correctly", async ({
```

just for clarity.

pngwn (commented):
Looks great! Left a question and a nit but good to go otherwise.

Thanks @freddyaboulton!

pngwn (approved):

gradio/components/gallery.py
```diff
@@ -177,10 +183,26 @@ def postprocess(
...
+            return img
+        else:
+            converted_image = _Image.open(img)
+            if type == "numpy":
+                converted_image = np.array(converted_image)
+            return converted_image
+
+    def preprocess(
+        self, payload: GalleryData | None
+    ) -> List[tuple[str | _Image.Image | np.ndarray, str | None]] | None:
```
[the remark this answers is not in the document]
freddyaboulton: You're right that this is more precise. I wish there was a less verbose way to express this with the type hints but I'm not sure there is.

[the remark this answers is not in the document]
freddyaboulton: Just pushed up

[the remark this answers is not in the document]
pngwn: > I wish there was a less verbose way to express this with the type hints but I'm not sure there is.

Yeah same but I can't think of another way to express it, unless there is a type helper out there somewhere.

.config/playwright-setup.js
abidlabs: cc @AK391 lol

abidlabs (approved):
LGTM good stuff @freddyaboulton!

freddyaboulton:

PR #7151, opened 2024-01-25
Fixes #7136 -- gradio client iterator can hit finish event before yield all outputs

### LINKED ISSUE #7136: client incorrectly iterates counter, and standard while out in job loop leaves left over outputs not consumed
### Describe the bug

```
            with self.communicator.lock:
                if len(self.communicator.job.outputs) == self._counter + 1:
                    o = self.communicator.job.outputs[self._counter]
                    self._counter += 1
                    return o
                if self.communicator.job.latest_status.code == Status.FINISHED:
                    raise StopIteration()
```

Issue is that by time code reaches the first conditional, the number of job.outputs may already be 2+ while self._counter=0 at first.

So everything hangs until the end.  I find this is quite common.

Additionally, the job finished can be reached, but there are still extra things to be consumed.  So I have to add extra code after the normal `for out in job` to catch this extra stuff, making the client annoying to code for.

In the end I have to hack together a bunch of external code to work around these issues.

### Reproduction

A detailed repro is not required and the problem is self-evident from the code.

I can put together a PR if desired.

### Screenshot

_No response_

### Logs

_No response_

### System Info

```shell
gradio 4.15.0 or also older gradio too.
```

### Severity

Blocking usage of gradio

pseudotensor:
## Description

Partially fixed iterator here: https://github.com/gradio-app/gradio/commit/287fe6782825479513e79a5cf0ba0fbfe51443d7

But failed to account for fact that finish can be true while have not returned all outputs.

Closes: #7136


client/python/gradio_client/client.py
```diff
@@ -1557,7 +1557,8 @@ def __next__(self) -> tuple | Any:
                     o = self.communicator.job.outputs[self._counter]
                     self._counter += 1
                     return o
-                if self.communicator.job.latest_status.code == Status.FINISHED:
+                if self.communicator.job.latest_status.code == Status.FINISHED and \
```
freddyaboulton, line 1560: Thanks @pseudotensor ! This looks good but it won't fix the following point you raised in the issue right?

> Issue is that by time code reaches the first conditional, the number of job.outputs may already be 2+ while self._counter=0 at first.

I think we can copy over the new entries of `communicator.job.outputs` to a temporary list and then pop from that. Something like this:


```python
        while True:
            with self.communicator.lock:
                if len(self.communicator.job.outputs) > self._counter:
                    for i in range(self._counter, len(self.communicator.job.outputs)):
                        self._outputs.append(self.communicator.job.outputs[i])
                    self._counter = len(self.communicator.job.outputs)
                if self._outputs:
                    return self._outputs.pop(0)
                if self.communicator.job.latest_status.code == Status.FINISHED and not self._outputs:
                    raise StopIteration()
                time.sleep(0.001)
```

pseudotensor: What I mentioned was that your team already fixed the first issue I mentioned in gradio 4 updates.

pseudotensor: To be clear, the only issue with the first conditional was a > vs. >=

freddyaboulton: Ah gotcha - ok looks good to me

freddyaboulton (approved):

client/python/gradio_client/client.py
freddyaboulton, line 1560:
```suggestion
                if (
                    self.communicator.job.latest_status.code == Status.FINISHED
                    and self._counter >= len(self.communicator.job.outputs)
                ):
```

Need this to pass the lint

abidlabs:
Thanks @pseudotensor and @freddyaboulton. 

@pseudotensor we can merge this in once you run `bash scripts/format_backend.sh` or give maintainers the permission to modify this PR.

client/python/gradio_client/client.py
pseudotensor: Did that.  I ignored what that script did to some other random file, and only committed the file I changed.

Issue #6981, opened 2024-01-09
Add possibility to show copy button to gr.Dataframe


mireiarmirapeix:
- [x] I have searched to see if a similar issue already exists.
[#2585 ](https://github.com/gradio-app/gradio/issues/2585) Add copy to clipboard button to other input types - added show_copy_button to gr.Textbox()
[#6086](https://github.com/gradio-app/gradio/issues/6086) Copying cell values from dataframe output - discusses copying individual cells but not entire dataframe

**Is your feature request related to a problem? Please describe.**  
I am building an app that returns tabular data. I am able to display the output as a gr.Dataframe() but there's no way to add a copy button to it (nor to easily select and copy the data manually). I have tried displaying the data as gr.Markdown() which allows me to select and copy manually, but it doesn't allow for a copy button either.

**Describe the solution you'd like**  
The possibility to add a copy button to Dataframes and Markdowns.

**Additional context**

pngwn:
We need to figure out where to put buttons like these. I think we need to adjust the design of the dataframe in order to add a small 'toolbar' or icon bar. This would solve a number of issues and offer a clearer path forwards for some of the other feature requests we have had.

mberco-quandl:
@abidlabs +1 Stumbled on this thread while looking for a workaround to the lack of ability to write dataframes into Gradio's file handling system as a csv https://github.com/gradio-app/gradio/issues/8098#issuecomment-2072370094. Currently I write the dataframe to my working directory, then I pass that full path into the Gr.File object, then use an upload event handler to delete the file in the working directory. In this way I can rely on `delete_cache` to semi-safely manage storage, however there is a risk of my working directory getting polluted in a tempfile is not deleted. If a user could copy-paste an entire dataframe's contents into memory they could simply paste it into an open spreadsheet. _**Even better**_ would be the ability to download a csv representing the Dataframe's contents.

Issue #6086, opened 2023-10-25
Copying cell values from dataframe output


lukasugar:
- [x ] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
It's difficult to copy cell value from output Dataframe. It's very frustrating to have to go "inspect element" to just simply copy text.

**Describe the solution you'd like**  
I'd like to either:
- have "copy" button in the corner of a cell when I hover it
- be able to copy text from cell using triple click

**Additional context**  

https://github.com/gradio-app/gradio/assets/22435018/aae0d627-2152-4660-8fde-262102c73064

lukasugar:
Workaround: instead of using Dataframe, I'm using Markdown, and filling it with tabulated data. Works, but would be cool to support this in Dataframe :)

abidlabs:
Hi @lukasugar which browser are you using?

hysts:
@abidlabs I'm using Chrome and have the same issue with `gradio==3.50.2` and `gradio==4.0.2`. It was possible to copy when using `gradio==3.44.4`, for example.

lukasugar:
I'm also using Chrome, on Mac.

shimizust:
I'm also seeing this issue. I was able to double click a DataFrame cell to copy it in `gradio==3.44.3`, but after upgrading to `gradio==3.50.2` double-clicking does not select the cell contents. This is on Chrome on Mac/Linux.

1713172499:
You can try using css, and I did it:
`.cell-wrap span{
user-select:text !important;
-moz-user-select:text !important;
-webkit-user-select:text !important;
-o-user-select:text !important;
-ms-user-select:text !important;
}
`
then:
with gr.Blocks(css=css) as demo:

one-lithe-rune:
> You can try using css, and I did it: `.cell-wrap span{ user-select:text !important; -moz-user-select:text !important; -webkit-user-select:text !important; -o-user-select:text !important; -ms-user-select:text !important; } ` then: with gr.Blocks(css=css) as demo:

This works about 50% for me. It doesn't amount to a real workaround.

If you manage to create the text selection you want without triggering gradio dataframe cell selection behavior, then you're fine. However gradio seems to use mouseup to decide when you're  selecting a cell as a whole, and doesn't seem to check whether that cell is already selected, so it always re-renders it, and because mouseup is also what you use to mark the end of a text selection, you text selection is immediately wiped out as the gradio cell selection is re-applied

[...]

PR #7154, opened 2024-01-25
Allow selecting texts in dataframe cells

shubhamofbce:
## Description
I have used user-select css property to allow users to select texts in dataframe cells. This will fix the issue where users where not able to copy text from output(non-editable) cells of dataframe.

Closes: #6086 
Related: #6981 

## 🎯 PRs Should Target Issues

https://github.com/gradio-app/gradio/assets/45156638/8c442fb2-5aba-4a13-82e3-7888d39b468a

## Tests


shubhamofbce:
@abidlabs What do you suggest, I should keep the show_copy_button default to false or true.
I was also not able to copy the output cells in the dataframe as reported, so true makes more sense to me. What do you suggest ??

abidlabs:
Hi @shubhamofbce, thanks for this very interesting PR. 

The problem with adding a copy button to each cell is that it becomes very cluttered. The better approach would be to simply allow users to select any cell and copy it using ctrl/cmd+C as they would intuitively try to do. This should not require any new parameters and would keep the UI clean.

Could you refactor this PR so that you can just copy cell values by using the standard copy shortcut?

shubhamofbce:
Sure, I can do that as well. I will make the required changes.

shubhamofbce:
@abidlabs I have updated the PR, now we can select data normally from any dataframe cells.

abidlabs:
LGTM @shubhamofbce! Thanks for this PR.

I agree it closes #6086 as this allows copying from individual cells, but it doesn't close #6981 as that was related to copying entire dataframes. I think the ideal UX for copying the entire dataframe would be an excel-like UX where you could select multiple cells (by holding the shift key) and then then copying them all using the standard keyboard shortcuts.

abidlabs (approved):

shubhamofbce:
@abidlabs Thanks for the quick feedback.

PR #7157, opened 2024-01-25
Defer importing matplotlib

akx:
## Description

The Gradio features that actually require matplotlib are few and far between, and most matplotlib importing is already deferred.

This moves the rest.

### DIFF
--- a/.changeset/old-rice-laugh.md
+++ b/.changeset/old-rice-laugh.md
@@ -1,0 +1,5 @@
+---
+"gradio": minor
+---
+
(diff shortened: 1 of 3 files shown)


akx:
I unfortunately can't tell if the test failures are legitimate, or if it's just the test suite being broken since I'm an external contributor and `secrets` are missing.

abidlabs:
We’ll take a look thanks @akx!

abidlabs:
LGTM! Thanks again @akx

PR #7159, opened 2024-01-25
Ensure `gradio cc publish` uploads the documentation space, if it exists.

pngwn:
## Description

Closes: #(issue)

## 🎯 PRs Should Target Issues

## Tests


gradio/cli/commands/components/publish.py
```diff
@@ -112,8 +115,9 @@ def _publish(
             help="HuggingFace token for uploading demo. Can be omitted if already logged in via huggingface cli."
         ),
     ] = None,
+    prefer_local: Annotated[bool, Option(help="Install")] = False,
```
freddyaboulton, line 118: I think you forgot to finish the help string here

gradio/cli/commands/components/publish.py
```diff
@@ -219,10 +223,22 @@ def _publish(
             source_dir_ = source_dir_ or str(Path("."))
             source_dir = Path(source_dir_).resolve()
     if upload_demo:
+        package_name, version = wheel_file.name.split("-")[:2]
```
freddyaboulton, line 226: might be better to pull these from the pyproject toml

freddyaboulton (approved):
Thanks @pngwn ! Left a couple of nits but works well.

gradio/cli/commands/components/publish.py
pngwn: Yeah, I was thinking that but we don't actually _require_ the source directory here. Should we just assume it is `./pyproject.yaml` if none is passed?

gradio/cli/commands/components/publish.py
pngwn: ops

gradio/cli/commands/components/publish.py
pngwn: `.toml` even

pngwn:
Okay, addressed those comments.

Changed the package_name to use the `source dir`  if it passed, `.` if it isn't, and fall back to the wheel if everything fails.

pngwn:
There was also a rogue backtick on the guide causing some formatting issues. I've fixed that here too.

Issue #7166, opened 2024-01-26
Demo fake_diffusion fails, with slider able to have floats by default


abhishekkr:
### Describe the bug

Demo fake_diffusion fails, with slider able to have floats by default.. the Interface's `fn` passes it to `range()` which gives error if someone tries to tweak input value other than default without making sure the increment is in whole numbers.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

* Run [demo/fake_diffusion](https://github.com/gradio-app/gradio/blob/main/demo/fake_diffusion/run.py) with Gradio.
* Change Input Slider value to have Float
* Submit the Input, Output shall have Error

### Screenshot

![Screenshot from 2024-01-26 11-46-50](https://github.com/gradio-app/gradio/assets/32578/6ef5f56f-9c64-482f-aa2d-837b6120247b)

### Logs

```shell
Traceback (most recent call last):
  File "/tmp/today/.venv/lib64/python3.11/site-packages/gradio/queueing.py", line 495, in call_prediction
    output = await route_utils.call_process_api(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/today/.venv/lib64/python3.11/site-packages/gradio/route_utils.py", line 232, in call_process_api
    output = await app.get_blocks().process_api(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/today/.venv/lib64/python3.11/site-packages/gradio/blocks.py", line 1561, in process_api
    result = await self.call_function(
             ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/today/.venv/lib64/python3.11/site-packages/gradio/blocks.py", line 1191, in call_function
    prediction = await utils.async_iteration(iterator)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/today/.venv/lib64/python3.11/site-packages/gradio/utils.py", line 519, in async_iteration
    return await iterator.__anext__()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/tmp/today/.venv/lib64/python3.11/site-packages/gradio/utils.py", line 512, in __anext__
    return await anyio.to_thread.run_sync(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #7167, opened 2024-01-26
Fix fake_diffusion by slider to only have int so range() doesn't fail

abhishekkr:
## Description

I tried this demo on Gradio Docs and the run failed for me, when I tweaked slider value.

Cause was default increment allows float values and range(steps) fails for float values. Thus solving it in Gradio's idiomatic way... adding step=1 to Slider would only allow steps:int to be used. Avoiding error scenarios.

Closes: #7166


abidlabs:
Nice catch @abhishekkr! LGTM.

abidlabs:
We'll need to update the demo notebooks as well (we have a notebook for each demo). You can do that by doing:

`cd demo && python generate_notebooks.py` 

or you can let maintainers edit this PR and I'm happy to make that change. Then we should be good to merge!

abhishekkr:
> We'll need to update the demo notebooks as well (we have a notebook for each demo). You can do that by doing:
> 
> `cd demo && python generate_notebooks.py`
> 
> or you can let maintainers edit this PR and I'm happy to make that change. Then we should be good to merge!

will do :)

abhishekkr:
@abidlabs have updated the PR's commit for notebook changes

abidlabs:
Thanks @abhishekkr!

freddyaboulton (approved):

PR #7183, opened 2024-01-27
[WIP] Refactor file normalization to be in the backend and remove it from the frontend of each component

abidlabs:
The idea behind this PR is to clean up how we do "file normalization" i.e. how we convert filepaths into URLs like `/file=filepath` or `/proxy=filepath` so that the files can be fetched from the frontend of the Gradio app. 

A filepath is converted into a URL in the following way:

* If its a file on the local machine, it is prepended with `{root_url}/file=`. The root url is the path where the Gradio app is running. Usually, it is just `http://localhost:7860`. However, if the Gradio app is running within a larger FastAPI app on a subpath, or if it is running behind an nginx server, the root_url might be different, e.g. `http://localhost:7860/gradio`. 

* Otherwise, if the filepath is on an external Space (in the words, the component has been loaded using `gr.load`), then the prefix `{root_url}/proxy={space_url}/file=` is prepended to the filepath, which makes it a URL that can be accessed from the frontend. The local Gradio app serves as a proxy to route the request, as this is needed to access files on private Spaces using the local user's HF token

Previously, this file normalization was done in the frontend, in every component's frontend files, which led to code repetition and error-prone code, and was an annoyance for custom component developers. This PR moves it to the backend and does it in a single place, in a backwards-compatible manner for existing custom components. 

Once we have this cleanup in place, it should be easier to tackle issues like: #6744

Todo:
- [x] handle proxy urls
- [x] handle root urls
- [x] fix functional tests
- [x] fix unit tests
- [x] handle streaming
- [x] clean up frontend


client/js/src/upload.ts
```diff
@@ -18,46 +18,18 @@ export function normalise_file(
 	proxy_url: string | null // root_url: string | null
 ): FileData[] | FileData | null;
 
+/*
```
aliabid94, line 21: Is there a need for backwards compatibility? wouldn't previous custom components be version pinned for the client?

pngwn: They should but we would need to make the client version bump minor, to respect semver.

pngwn: (Minor because this is a pre-1.0 version, so minor is breaking).

abidlabs: Good point @aliabid94 @pngwn I'll change

abidlabs: Done

js/image/Index.svelte
```diff
@@ -64,8 +62,12 @@
 		share: ShareData;
 	}>;
 
-	$: url = _value?.url;
-	$: url, gradio.dispatch("change");
+	$: {
```
aliabid94, line 65: why not just listen to the url change as before?

abidlabs: Yeah that would be okay too, I ended up removing quite a few things from Image, including this line, and then I copied this snippet over from the Audio component. Either way is fine I think

aliabid94 (approved):
LGTM, nice cleanup

abidlabs:
Thanks for the review @aliabid94! I'm going to merge this in as it'll make it easier to tackle some other issues I'm looking at.

Issue #7180, opened 2024-01-26
Examples Images on gr.Examples Gallery are distorted


radames:
### Describe the bug

Images should be contain or cover to keep it's orignal aspect ratio

related issue #5067 

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

https://colab.research.google.com/drive/1EyjUNbM0nTa5PCpdnTqC3LpjU8B7pXBk#scrollTo=W6zmau9ccXv_

```python
import gradio as gr
from PIL import Image
import requests

img_s_example = ["https://source.unsplash.com/random/512x512",
                 "https://source.unsplash.com/random/512x1000",
                 "https://source.unsplash.com/random/1000x512"]

pil_imgs = [Image.open(requests.get(url, stream=True).raw) for url in img_s_example]
names = [img.save(f'example_{i}.jpg') or f'example_{i}.jpg' for i, img in enumerate(pil_imgs)]
def predict(text, image):
    return image.rotate(45)

with gr.Blocks() as demo:
    with gr.Row():
      with gr.Column():
        image = gr.Image(label="in")
        text = gr.Text()
      with gr.Column(): 
        image_out = gr.Image(label="ou")
        greet_btn = gr.Button("Rotate")
    greet_btn.click(fn=predict,
                    inputs=[text, image],
                    outputs=[image_out])
    gr.Examples(
        examples=[
            ["hello", "./example_0.jpg"],
            [None, None],
            [None, "./example_1.jpg"],
            [None, "./example_2.jpg"],
            ["dsdsd", None],
        ],
        fn=predict,
        inputs=[text, image],
        outputs=[image_out]
    )

if __name__ == "__main__":
    demo.launch()
```

### Screenshot

![image](https://github.com/gradio-app/gradio/assets/102277/93e357b4-726a-41d1-bf73-aff0010186ce)

### System Info

```shell
gradio 4.16.0
```

### Severity

I can work around it

abidlabs:
Thank you for the comprehensive repro @radames, very helpful

Issue #5067, opened 2023-08-02
Broken/Missing Images in UI when providing Partial Examples for gr.Image()


yvrjsharma:
### Describe the bug

I would only like to provide examples for a subset of input components. I am passing None for the image input in order to exclude it for some of the examples. 
The UI appears to be showing a broken or corrupted image instead and would be cleaner if it was a blank field instead [please refer the attached gif].
![examples issue](https://github.com/gradio-app/gradio/assets/48665385/80f335fd-f692-4fe0-bfc9-6320ccdd8bc7)

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```
with gr.Blocks() as demo:
    img = gr.Image(label="image input")
    tb = gr.Textbox(label="input text")

    gr.Examples(examples = [
        ["/content/lion.jpg", "Lion"],
        [None, "Tiger"],
        ["/content/zebra.jpg", None],
    ],
    inputs = [img, tb]                
                )
demo.launch()
```

### Screenshot

![image](https://github.com/gradio-app/gradio/assets/48665385/3989d2a3-28de-4773-8a1e-c0236d552355)

### System Info

```shell
Ran on Colab and Spaces
```

### Severity

I can work around it

radames:
Another issue is that if the images have a non-square aspect ratio, the Gallery displays distorted images.


<img width="594" alt="image" src="https://github.com/gradio-app/gradio/assets/102277/bea59838-314b-4cf7-84e6-23d608f7b1b3">


reproduction code

https://colab.research.google.com/drive/1EyjUNbM0nTa5PCpdnTqC3LpjU8B7pXBk#scrollTo=W6zmau9ccXv_
```python
import gradio as gr
from PIL import Image
import requests

img_s_example = ["https://source.unsplash.com/random/512x512",
                 "https://source.unsplash.com/random/512x1000",
                 "https://source.unsplash.com/random/1000x512"]

pil_imgs = [Image.open(requests.get(url, stream=True).raw) for url in img_s_example]
names = [img.save(f'example_{i}.jpg') or f'example_{i}.jpg' for i, img in enumerate(pil_imgs)]
def predict(text, image):
    return image.rotate(45)

with gr.Blocks() as demo:
    with gr.Row():
      with gr.Column():
        image = gr.Image(label="in")
        text = gr.Text()
      with gr.Column(): 
        image_out = gr.Image(label="ou")
        greet_btn = gr.Button("Rotate")
[...]

PR #7192, opened 2024-01-27
Handle the case where examples is `null` for all components

abidlabs:
Example values can be `null` but this case wasn't really handled for any of the components. This PR handles the case for all components as part of fixing https://github.com/gradio-app/gradio/issues/5067. Also fixes the image distortion issue for examples mentioned here: https://github.com/gradio-app/gradio/issues/7180

To test, at least for `gr.Textbox` and `gr.Image`, use @radames's test code from https://github.com/gradio-app/gradio/issues/7180:

```py
import gradio as gr
from PIL import Image
import requests

img_s_example = ["https://source.unsplash.com/random/512x512",
                 "https://source.unsplash.com/random/512x1000",
                 "https://source.unsplash.com/random/1000x512"]

pil_imgs = [Image.open(requests.get(url, stream=True).raw) for url in img_s_example]
names = [img.save(f'example_{i}.jpg') or f'example_{i}.jpg' for i, img in enumerate(pil_imgs)]
def predict(text, image):
    return image.rotate(45)

with gr.Blocks() as demo:
    with gr.Row():
      with gr.Column():
        image = gr.Image(label="in")
        text = gr.Text()
      with gr.Column(): 
        image_out = gr.Image(label="ou")
        greet_btn = gr.Button("Rotate")
    greet_btn.click(fn=predict,
                    inputs=[text, image],
                    outputs=[image_out])
    gr.Examples(
        examples=[
            ["hello", "./example_0.jpg"],
            [None, None],
            [None, "./example_1.jpg"],
            [None, "./example_2.jpg"],
            ["dsdsd", None],
        ],
        fn=predict,
        inputs=[text, image],
        outputs=[image_out]
    )

if __name__ == "__main__":
    demo.launch()
```

Closes: https://github.com/gradio-app/gradio/issues/7180
Closes: https://github.com/gradio-app/gradio/issues/5067


abidlabs:
The static tests are failing even though those files are untouched by this PR. @whitphx do you have any idea of what could be going on?

whitphx:
Hmm I can't find what's the problem

pnpm-lock.yaml
```diff
@@ -1361,15 +1361,21 @@ importers:
   js/simpleimage:
     dependencies:
       '@gradio/atoms':
+<<<<<<< HEAD
```
freddyaboulton, line 1364: FYI there's a merge conflict here

js/image/Example.svelte
```diff
@@ -26,10 +29,12 @@
 	.container.selected {
 		border-color: var(--border-color-accent);
 	}
+	.border.table {
+		border: 2px solid var(--border-color-primary);
+	}
 
 	.container.table {
 		margin: 0 auto;
-		border: 2px solid var(--border-color-primary);
```
freddyaboulton, line 32: In the gallery view there's still a border being applied

image

js/video/Example.svelte
```diff
@@ -38,7 +38,7 @@
 		/>
 	</div>
 {:else}
-	<div>{value}</div>
```
freddyaboulton, line 41: In this case we see a black box - I think it would be better to omit it entirely to be consistent with image and the other components

image

freddyaboulton (commented):
Looks good to me but noticed an issue with the empty video. Is it possible to test this via story book? Or is a unit test better here? CC @hannahblair

abidlabs:
> Hmm I can't find what's the problem

Thanks @whitphx. [Should be fixed now, it was this issue (internal link)](https://huggingface.slack.com/archives/C055NMB0S87/p1706405159927249)

abidlabs:
Thanks @freddyaboulton for the review, I'll fix those issues and add visual tests to storybook.

js/video/Example.svelte
abidlabs: Makes sense

pnpm-lock.yaml
abidlabs: Thanks, resolved

abidlabs:
Ok I've added stories for 4 components -- figured that this should be good to catch any regressions without bloating the number of components. 

Should be good for another look @freddyaboulton @hannahblair

freddyaboulton (approved):
Very nice @abidlabs !

js/checkbox/Example.svelte
```diff
@@ -9,7 +9,7 @@
 	class:gallery={type === "gallery"}
 	class:selected
 >
-	{value.toLocaleString()}
+	{value ? value.toLocaleString() : ""}
```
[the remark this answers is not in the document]
abidlabs: Thanks for pointing this out @hannahblair. As I was testing this, I realized that my original approach won't work since `value` for `Checkbox` can be a boolean and if its False, that gets rendered as an empty string. Will revise

abidlabs:
Thanks @freddyaboulton and @hannahblair for reviewing!

abidlabs:
Hi @TonyAssi what issue are you running into? This PR didn't handle displaying images on mobile, it was related to setting examples to `None`.

PR #7205, opened 2024-01-29
Fix SimpleImage package json

abidlabs:
Fixes the package json to look at the versions in the workspace instead of fixed versions. Should fix the static typechecking error that we're seeing.


pngwn (approved):
Awesome! Thanks @abidlabs!

abidlabs:
Thank _you_ @pngwn

Issue #7193, opened 2024-01-27
gr.Row(visible=True) and gr.Tab(visible=False) combination shows component artifacts under Tab, but Tab isn't showing.


dwipper:
### Describe the bug

The below code swaps between a login_row and a tabs_row. The first time through, clicking "Login" hides the login_row, and show the tabs_row. Clicking "Show Tabs" displays the Tabs and underlying components. 

Clicking "Logout" hides the tabs_row, tab_one, and tab_tow, and shows the login_row.

Clicking the "Login" button again hides the login_row and shows the tabs_row. But this time there are component artifacts from the tabs, but no Tabs (there shouldn't be anything on the screen since the Tabs were set to visible=False.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```
import gradio as gr

def change_tab_params():
    return gr.Tab(interactive=True, visible=True)

def change_to_tab_two():
    return gr.Tabs(selected=2)

def login():
    return gr.Row(visible=True), gr.Row(visible=False)

def logout():
    return gr.Row(visible=False), gr.Row(visible=True), gr.Tab(visible=False), gr.Tab(visible=False)

def show_tabs():
    return gr.Tab(visible=True), gr.Tab(visible=True)

with gr.Blocks() as demo:
    with gr.Row(visible=False) as tabs_row:
        with gr.Column():
            with gr.Row():
                with gr.Tabs() as tabs:
                    with gr.Tab(id=1, label="Tab 1", visible=False, interactive=True) as tab_one:
                        tab_one_tbox = gr.Textbox(label="Tab One Textbox")
                        tab_one_btn = gr.Button(value="Change Tab")
                    with gr.Tab(id=2, label="Tab 2", visible=False, interactive=True) as tab_two:
                        tab_two_tbox = gr.Textbox(label="Tab Two Textbox",interactive=False)
                        tab_one_btn.click(fn=change_to_tab_two,inputs=None, outputs=[tabs])
            with gr.Row():
                show_tabs_btn = gr.Button(value="Show Tabs")
                logout_btn = gr.Button(value="Logout")
                show_tabs_btn.click(fn=show_tabs, inputs=None, outputs=[tab_one,tab_two])
    with gr.Row(visible=True) as login_row:
        login_btn = gr.Button(value="Login")
        login_btn.click(fn=login,inputs=None,outputs=[tabs_row, login_row])
[...]

PR #7220, opened 2024-01-30
Add `visible` check to Tab 

hannahblair:
## Description

Ensures that TabItem contents won't be visible if the Tab is `!visible`. 

We initially set the `display` class to `block` if `selected_tab === id` (i.e it's the selected tab), but not checking if `visible` === `true`. 

In the linked demo, the user was programmatically hiding the Tab, but not changing the selected tab, so `display` was still set to `block` in that instance. Thanks for spotting this @dwipper!

Test with the demo in the original issue 🧪 

Closes: #7193

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/true-pears-decide.md
+++ b/.changeset/true-pears-decide.md
@@ -1,0 +1,6 @@
+---
+"@gradio/tabitem": patch
+"gradio": patch
+---
+
+fix:Add `visible` check to Tab 


js/tabitem/shared/TabItem.svelte
```diff
@@ -49,4 +49,8 @@
 		border-bottom-left-radius: var(--container-radius);
 		padding: var(--block-padding);
 	}
+
+	.hide {
```
freddyaboulton, line 53: I don't think we need this since we're setting display: "none" inline?

freddyaboulton: otherwise looks good to me!

PR #7222, opened 2024-01-30
Add mobile Chromatic tests

hannahblair:
## Description

This PR implements viewport tests to several components. Happy to add more - the components I've added these into so far are those I feel are more important to keep an eye on as they may be more susceptible to change in smaller viewports. 

This PR also adds the other layout components as stories. 

`Meta` is now deprecated in svelte-addon-csf, and should be expressed as `export const meta` (see [here](https://github.com/storybookjs/addon-svelte-csf/issues/135)), so I've made a few changes accordingly. 

## 🎯 PRs Should Target Issues

## Tests


js/column/Column.stories.svelte
abidlabs, line 1: Nice to have these tests!

abidlabs:
@hannahblair do you know why the storybook isn't running on ci? or maybe it is, but I don't see the link to the live storybook?

abidlabs:
Where can I see the new mobile story tests @hannahblair? E.g. I do this viewport button, but I also see that button for other components like `Textbox` that don't have any new stories added:

https://github.com/gradio-app/gradio/assets/1778297/e3700adb-8fff-493f-8086-15873dbbc511

abidlabs:
Got it. I agree, I think mobile + desktop should be sufficient!

abidlabs (approved):
One request: the plot components might look different on desktop vs. mobile as well. If we could add stories for some of these as well (currently they are not visually tested at all), that'd be awesome

But otherwise, this looks great! Thanks so much @hannahblair for adding these. Let's make the mobile UX 🚀

js/row/Row.stories.svelte
```diff
@@ -1,10 +1,13 @@
-<script>
-	import { Meta, Template, Story } from "@storybook/addon-svelte-csf";
+<script context="module">
+	import { Template, Story } from "@storybook/addon-svelte-csf";
 	import Row from "./Index.svelte";
 	import Button from "../button/shared/Button.svelte";
-</script>
 
-<Meta title="Layout Elements/Row" component={Row} />
+	export const meta = {
```
freddyaboulton, line 6: Shouldn't we also test the mobile view here?

freddyaboulton (approved):
Very awesome @hannahblair ! I think we should also test the layouts in mobile mode. I agree we can take out the tablet tests. Also, are you going to accept all the new test results in chromatic? I think the job is pending because of that.

abidlabs:
@hannahblair I took a quick look and everything looks good to me. Adding the other plot types would be a nice to have but not critical if you want to get these in

Issue #7021, opened 2024-01-13
No output image when using webcam streaming input  


nassimberrada:
### Describe the bug

Hi! 

I am trying to build a demo that uses the webcam as input feed to perform real-time image editing. Unfortunately, I don't seem to get any output despite the webcam input stream working properly. I am getting the same behavior whether using the example snippet in the doc:

```
demo = gr.Interface(
    fn=flip,
    inputs=gr.Image(sources=["webcam"], streaming=True),
    outputs=gr.Image(),
    live=True
    )
demo.launch(share=True)
```
or the no-loading solution proposed in one of the threads as follows:
```
with gr.Blocks() as demo:
    with gr.Row():
        input_img = gr.Image(sources=["webcam"])
        output_img = gr.Image()
        input_img.stream(flip, input_img, output_img)
demo.launch(share=True)
```
I am encountering this issue when running on a standard google colab instance, both on the Ipython embedded application and the shared link. The issue doesn't seem related to browser settings as I experienced this on different browsers but this seems to work when running locally (albeit very slowly).

The attached screenshot shows a black screen on the input side but this is just due to my camera being covered, I can confirm the input feed is working properly.

Would love to get any pointers to solve this, thanks! 

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python

import gradio as gr
import numpy as np

def flip(im):
    return np.flipud(im)

# No loading 

with gr.Blocks() as demo:
    with gr.Row():
        input_img = gr.Image(sources=["webcam"])
        output_img = gr.Image()
        input_img.stream(flip, input_img, output_img)
demo.launch(share=True)

# Or standard streaming
demo = gr.Interface(
    fn=flip,
    inputs=gr.Image(sources=["webcam"], streaming=True),
    outputs=gr.Image(),
    live=True
    )
demo.launch(share=True)
```

### Screenshot

![gradio_issue](https://github.com/gradio-app/gradio/assets/112006029/a9f76693-abb0-42ab-8ba1-a2f639378504)

### Logs

```shell
No error
```

### System Info

```shell
Gradio version 4.14.0
```

### Severity

Blocking usage of gradio

freddyaboulton:
I can repro the problem when using the `Interface` code snippet but blocks.

HubHop:
same here

donnadulcinea:
Not sure this is related, but I am unable to use gradio webcam at all. With any version matrix of gradio and python.
Getting

[...]

Issue #3642, opened 2023-03-27
Allow Webcam stream to be stopped/resume programmatically 


radames:
As for privacy concerns, it would be useful to enable the webcam stream to be stopped and resumed programmatically. For instance, in the code below, when a user accesses the app, the webcam stream permission would be requested right away, rather than only upon tab activation. Moreover, if the user switches from the webcam tab to the file tab, the webcam remains active.

**Ideas**:

* Allow the webcam stream to be stopped and resumed programmatically.
* Destroy the webcam component when the tab is inactive and remount it when the tab becomes active again.
* Once issue #3306  is resolved, users might be able to update the source dynamically, and the webcam stream will automatically stop and resume as they change the source, e.g., source="webcam" => source="file" => source="webcam".

Code example:

```python
import gradio as gr

with gr.Blocks() as demo:
    with gr.Tab("File"):
        with gr.Row():
          with gr.Column():
            image = gr.Image(label="Input Image")
            run = gr.Button("Run")
          with gr.Column():
            output = gr.Image(label="Input Image")
    with gr.Tab("File"):
        with gr.Row():
          with gr.Column():
            image = gr.Image(label="Input Image", source="webcam")
            run = gr.Button("Run")
          with gr.Column():
            output = gr.Image(label="Input Image")
  
demo.launch()
```

https://colab.research.google.com/drive/1PHohs4lgt-Ty4SHK1ZGdpHwmWadvZXAY

pngwn:
> Destroy the webcam component when the tab is inactive and remount it when the tab becomes active again.

This is what we should be doing really but requires some internal changes. It should be addressed soon though.

innat-asj:
Any update on this? Or, any workaround with the current APIs?

taoari:
Hope to see an update on this.

PR #7228, opened 2024-01-31
Allow start/pause of streaming image input. Only access the webcam while it's needed

freddyaboulton:
## Description

Closes: #3642
Closes: #7230

If the webcam is an allowed image source, gradio will access the webcam as soon as the component is mounted on the page. There is also no way to stop gradio from accessing the webcam. This is #3642

When the webcam is an allowed source, users have to click into the image component to grant webcam access. This is what it looks like in that case:

image

After the webcam stops recording or taking a picture, gradio will stop using the webcam and revert back to the ^previous state.

Additionally, when `gr.Image(streaming=true)`, the webcam component will behave similarly to video recording. There will be a record and stop button at the bottom that users can use to stop the stream of images. 

### Webcam Demo no streaming
webcam_demo

### Webcam Demo with streaming
stream_frames

Test streaming with this demo
```python
import gradio as gr
import numpy as np

def flip(im):
    return np.flipud(im)

with gr.Blocks() as demo:
    with gr.Row():
        input_img = gr.Image(sources=["webcam"])
        output_img = gr.Image()
        input_img.stream(flip, input_img, output_img)
demo.launch(share=True)

if __name__ == "__main__":
    demo.launch()
    
```

EDIT:

Also fixes Interface streaming blocking the output, fixes: #7021 

## 🎯 PRs Should Target Issues

## Tests


js/image/shared/Webcam.svelte
```diff
@@ -180,19 +203,34 @@
 		event.stopPropagation();
 		options_open = false;
 	}
+
+	$: console.log("webcam_accessed", webcam_accessed);
```
freddyaboulton, line 207: will delete 🤦

js/image/shared/Webcam.svelte
```diff
@@ -180,19 +203,34 @@
...
+		class:flip={mirror_webcam}
+		class:invisible={!webcam_accessed}
+	/>
+	{#if !webcam_accessed}
+		<div class="wrap-webcam-access">
+			<span class="icon-wrap">
+				<WebcamIcon />
+			</span>
+			<button on:click={async () => await access_webcam()}
+				>{"Click to access webcam"}</button
```
freddyaboulton, line 222: This should probably be camel case for consistency

aliabid94:
I could've sworn I made the change to set the progress tracker visibility to False when streaming from a webcam

abidlabs:
> I could've sworn I made the change to set the progress tracker visibility to False when streaming from a webcam

I'm pretty sure its a regression. We should do this for any streaming event, e.g. https://github.com/gradio-app/gradio/issues/7230 as well

abidlabs:
This looks great to me @freddyaboulton! My only feedback is that the clickable area to give access to the webcam is very small. I kept clicking the icon and was confused why it wasn't working. I'd suggest making this whole area clickable:

image

aliabid94:
Is the demo supposed to be using an audio component? Btw when I was running locally I did not see the progress bar but also I got this error after a while when streaming a webcam:
`ValueError: Cannot process this value as an Image, it is of type: <class 'tuple'>`

aliabid94:
> Btw when I was running locally I did not see the progress bar

Actually there was a visible progress bar only when using streaming from Interface, fixed in this PR in this commit: https://github.com/gradio-app/gradio/pull/7228/commits/db019e4c2a61f8fe16f68dc915982c060182267d

freddyaboulton:
> Actually there was a visible progress bar only when using streaming from Interface, fixed in this PR in this commit: https://github.com/gradio-app/gradio/commit/db019e4c2a61f8fe16f68dc915982c060182267d

Nice - confirmed it works! 🔥 

> Is the demo supposed to be using an audio component?

Yea that's my bad - copy/pasted the wrong demo. I updated the PR description to include the correct one but also including it here

```python
import gradio as gr
import numpy as np

def flip(im):
    return np.flipud(im)

with gr.Blocks() as demo:
    with gr.Row():
        input_img = gr.Image(sources=["webcam"])
        output_img = gr.Image()
        input_img.stream(flip, input_img, output_img)
demo.launch(share=True)

if __name__ == "__main__":
    demo.launch()
```

> My only feedback is that the clickable area to give access to the webcam is very small. 

Agreed - just pushed up a fix! Will wait for CI to pass before opening it up for broader review.

js/image/shared/Webcam.svelte
```diff
@@ -131,7 +142,20 @@
...
 
-	access_webcam();
+	let webcam_accessed = false;
+
+	function record_video_or_photo(): void {
+		if (mode === "image" && streaming) {
+			recording = !recording;
+		}
+		const func = mode === "image" ? take_picture : take_recording;
+		func();
```
aliabid94, line 152: let's avoid abbreviations like `func`. Personally think this is more readable tbh
```suggestion
		if (mode === "image") {
		  take_picture();
		} else {
		  take_recording();
		}
```

js/image/shared/Webcam.svelte
```diff
@@ -253,6 +282,10 @@
 		height: var(--size-full);
 	}
 
+	.invisible {
```
aliabid94, line 285: nit, but we use the class name `hide` throughout the codebase for `display: none;`, for consistency would be nice

aliabid94 (approved):
LGTM great work

dawoodkhan82:
Looks great @freddyaboulton. Small nit, I notice after taking a picture the input component quickly flashes back to the "click to access webcam" state before showing the taken picture.

abidlabs:
Maybe also fixes: https://github.com/gradio-app/gradio/issues/7230?

freddyaboulton:
Thanks for the reviews everyone! Will merge the comments and flicker have been addressed.

PR #7229, opened 2024-01-31
Fix hyphen-bug in gradio cc publish

### LINKED ISSUE #7213: Custom component feedback
When I was building my custom component I ran into a number of issues. Dumping here as feedback while we figure out what action to take (if any). The component I was working on was bootstrapped back in prerelease so some of these issues may not be present any more. I will be building a new component this week and will check. 

- tooling / configuration 
  - if the `project.readme` isn't set correctly you get a weird error about long description. 
  - I didn't have `gradio` in my dependencies field. 
  - gradio cc publish failed catastrophically when trying to find the latest versions of the wheel in CI meaning I had to publish manually. (Will make reproduction and file an issue)
- runtime
  - some small css bugs. We should all build some stuff to make sure various internal components work correctly when used in different contexts
  - the static / interactive flipping causes many issues. Created a separate issue for this. 
  - custom status tracker was trickier than I expected, might have just been the above issue complicating things though. The props are definitely a bit weird, I think we could clean this up. 
  - I still have no idea what all of our CSS variables are, it is really difficult to figure out what to use. 
  - I don't know what all of our components are (especially atoms) and it isn't obvious how to use them. 

Other than two obvious bugs (that there are issues for) I think some of these have been resolved in recent versions or are simply a case of improving the docs. 

All that said, it still cool how we have the whole workflow pretty self contained. And other than a few specific issues things worked really nicely. A little more refinement and documentation and this will be really slick I think.

freddyaboulton:
## Description

Partially closes #7213 

If there was a hyphen in a parent directory, the publish command would fail.

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/spicy-rabbits-check.md
+++ b/.changeset/spicy-rabbits-check.md
@@ -1,5 +1,5 @@
 ---
-"gradio": minor
+"gradio": patch
 ---
 
 feat:Fix hyphen-bug in gradio cc publish
--- a/.changeset/spicy-rabbits-check.md
+++ b/.changeset/spicy-rabbits-check.md
@@ -1,0 +1,5 @@
+---
+"gradio": minor
+---
+
+feat:Fix hyphen-bug in gradio cc publish
--- a/gradio/cli/commands/components/publish.py
+++ b/gradio/cli/commands/components/publish.py
@@ -141,7 +141,7 @@ def _publish(
     ]
     wheel_file = max(
         (p for p in distribution_files if p.suffix == ".whl"),
-        key=lambda s: semantic_version.Version(str(s).split("-")[1]),
+        key=lambda s: semantic_version.Version(str(s.name).split("-")[1]),
     )
     if not wheel_file:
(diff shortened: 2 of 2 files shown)


pngwn:
Why do we actually perform this check?

freddyaboulton:
This came up in a previous [review](https://github.com/gradio-app/gradio/pull/6436#issuecomment-1814199283) of the CLI command. We should either upload everything in `dist/` or the latest version. I don't think we should clear out the dist directory.  I'm ok with either but given we're already doing the latter, I figure it's best just to fix the tiny bug and keep the behavior rather than change it.

dawoodkhan82 (approved):
Thanks for the fix!

pngwn:
Makes sense. Thanks @freddyaboulton! I havent had chance to test this in CI yet but the fix looks good!

abidlabs:
LGTM @freddyaboulton I'll go ahead and merge this in to unblock publish

PR #7236, opened 2024-01-31
Fix PIL imports

akx:
## Description

Follows up on #7116, which changed some of these imports.

## 🎯 PRs Should Target Issues

See #7116.

### DIFF
--- a/gradio/components/annotated_image.py
+++ b/gradio/components/annotated_image.py
@@ -15,8 +15,6 @@ from gradio.events import Events
 
 set_documentation_group("component")
 
-PIL.Image.init()  # fixes https://github.com/gradio-app/gradio/issues/2843
-
 
 class Annotation(GradioModel):
     image: FileData
--- a/gradio/components/base.py
+++ b/gradio/components/base.py
@@ -15,7 +15,6 @@ from pathlib import Path
 from typing import TYPE_CHECKING, Any, Callable
 
 from gradio_client.documentation import set_documentation_group
-from PIL import Image as _Image  # using _ to minimize namespace pollution
 
 from gradio import utils
 from gradio.blocks import Block, BlockContext
@@ -34,7 +33,6 @@ if TYPE_CHECKING:
 
 
 set_documentation_group("component")
-_Image.init()  # fixes https://github.com/gradio-app/gradio/issues/2843
 
 
 class _Keywords(Enum):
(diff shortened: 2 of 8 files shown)


abidlabs:
Thanks @akx lgtm feel free to open for review and we can get this merged in quite quickly

gradio/components/annotated_image.py
```diff
@@ -15,8 +15,6 @@
 
 set_documentation_group("component")
 
-PIL.Image.init()  # fixes https://github.com/gradio-app/gradio/issues/2843
```
abidlabs, line 18: We should keep this change as older versions of PIL (before 9.4.0) have the reported bug. Gradio pins to PIL 8.0 or higher

gradio/components/base.py
```diff
@@ -34,7 +33,6 @@ class DataframeData(TypedDict):
 
 
 set_documentation_group("component")
-_Image.init()  # fixes https://github.com/gradio-app/gradio/issues/2843
```
abidlabs, line 37: Same

gradio/components/image.py
```diff
@@ -18,7 +18,6 @@
 from gradio.events import Events
 
 set_documentation_group("component")
-PIL.Image.init()  # fixes https://github.com/gradio-app/gradio/issues/2843
```
abidlabs, line 21: Same

gradio/components/image_editor.py
```diff
@@ -18,10 +18,9 @@
 from gradio.events import Events
 
 set_documentation_group("component")
-_Image.init()  # fixes https://github.com/gradio-app/gradio/issues/2843
```
abidlabs, line 21: Same

gradio/image_utils.py
```diff
@@ -4,20 +4,20 @@
 from typing import Literal
 
 import numpy as np
-from PIL import Image as _Image  # using _ to minimize namespace pollution
+import PIL.Image
 
 from gradio import processing_utils
 
-_Image.init()
```
abidlabs, line 11: Same

gradio/components/annotated_image.py
abidlabs, line 18: We may need to keep this one as we don't import image_utils here

abidlabs:
Just one note above, otherwise LGTM @akx

gradio/components/annotated_image.py
akx: There's no way to import `gradio.components.annotated_image` without the parent module `gradio.components` getting imported, which imports `gradio.components.image` and then `image_utils`.

abidlabs (approved):
LGTM

abidlabs:
Merging, thanks again @akx

Issue #6856, opened 2023-12-20
Add diffusers pipelines to `gr.Interface.from_pipeline()`


abidlabs:
Just like we can easily create a demo for a `transformers` pipeline with `gr.Interface.from_pipeline()`, it'd be great to support this with `diffusers` pipelines as well: https://huggingface.co/docs/diffusers/api/pipelines/overview

shubhamofbce:
I am working on this.

PR #7265, opened 2024-02-01
Add support for diffuser pipelines in gr.Interface.from_pipeline().

shubhamofbce:
## Description

I have added support for the following diffuser pipelines in gr.Interface.from_pipeline().

- [x] StableDiffusionPipeline
- [x] StableDiffusionImg2ImgPipeline
- [x] StableDiffusionInpaintPipeline
- [x] StableDiffusionDepth2ImgPipeline
- [x] StableDiffusionImageVariationPipeline
- [x] StableDiffusionInstructPix2PixPipeline
- [x] StableDiffusionUpscalePipeline

Closes: #6856 

## 🎯 PRs Should Target Issues

## Tests


shubhamofbce:
@abidlabs, I've made an initial commit adding support for the Stable Diffusion pipeline, which I've successfully tested locally. Could you or someone from your team review the approach? I plan to apply the same methodology to the remaining pipelines.
Although the PR is still in draft, I'd like to ensure I'm on the right track early on, given the PR's substantial scope.

gradio/pipelines.py
```diff
@@ -19,14 +19,17 @@ def load_from_pipeline(pipeline: pipelines.base.Pipeline) -> dict:
     (dict): a dictionary of kwargs that can be used to construct an Interface object
     """
     try:
-        import transformers
+        import transformers, diffusers
         from transformers import pipelines
+        from diffusers import DiffusionPipeline
+        from diffusers import pipelines as diffuser_pipelines
+
```
abidlabs, line 26: Instead of having these within one block, which would mean that a user needs both `diffusers` and `transformers` installed, we should separate them so that only one is needed

abidlabs:
Thanks @shubhamofbce this is looking good! Just one comment at the top

shubhamofbce:
@aliabd This is ready for review.
I have made the changes such that if the pipeline you are trying to load is a transformer pipeline, only then you need to install transformers and same for diffusers. You don't need to install both of them together.
I can see **test / python / linux** failing, can you please confirm if that is expected or i need to do something as i have formatted my code using format_backend script, but it suggested a few reformatting in some part which were not my change so i didn't committed them. If it is failing due to that I will reformat the file and commit.

gradio/pipelines.py
```diff
@@ -18,253 +18,579 @@ def load_from_pipeline(pipeline: pipelines.base.Pipeline) -> dict:
...
-        }
-    elif hasattr(transformers, "ZeroShotClassificationPipeline") and isinstance(
-        pipeline, pipelines.zero_shot_classification.ZeroShotClassificationPipeline
-    ):
-        pipeline_info = {
-            "inputs": [
-                components.Textbox(label="Input", render=False),
-                components.Textbox(
-                    label="Possible class names (" "comma-separated)", render=False
+    if "transformers.pipelines" in str(type(pipeline)):
```
akx, line 22: Maybe refactor these top-level `if` branches into separate shorter helper functions like `_handle_transformers_pipeline()`?

Also, what if `str(type(pipeline))` is `my_happy_transformers.pipelines.something`? This should probably be `.startswith()`.

gradio/pipelines.py
```diff
@@ -18,253 +18,579 @@ def load_from_pipeline(pipeline: pipelines.base.Pipeline) -> dict:
...
+                            i["label"],
+                        )
+                        for i in r
+                    ],
+                ),
+            }
+        else:
+            raise ValueError(f"Unsupported pipeline type: {type(pipeline)}")
+
+    elif "diffusers.pipelines" in str(type(pipeline)):
```
akx, line 237: Ditto here – `_handle_diffusers_pipeline()`?, and ditto here with using `startswith` instead of substring?

gradio/pipelines.py
```diff
@@ -18,253 +18,579 @@ def load_from_pipeline(pipeline: pipelines.base.Pipeline) -> dict:
...
     interface_info = pipeline_info.copy()
     interface_info["fn"] = fn
     del interface_info["preprocess"]
     del interface_info["postprocess"]
 
     # define the title/description of the Interface
-    interface_info["title"] = pipeline.model.__class__.__name__
+    interface_info["title"] = (
+        pipeline.model.__class__.__name__
+        if "diffusers.pipelines" not in str(type(pipeline))
```
akx, line 592: And here – `startswith`.

abidlabs:
Thanks @shubhamofbce this generally looks good to me. I'll give it a full review after you've refactored based on @akx's comments. 

To get the python tests to run properly, you'll need to use ruff to format the code. Please run `bash scripts/format_backend.sh`

abidlabs:
@shubhamofbce let us know when this is ready for another look!

shubhamofbce:
@akx @abidlabs This is ready for review now.
I have created a new helper file to store the _helper methods and refactored the code to use new helper methods.
I also changed the type of pipeline in

shubhamofbce:
@akx @abidlabs This is ready for review now.
I have created a new helper file to store the _helper methods and refactored the code to use new helper methods.

**One important change that i want to point out**
I have also changed the type of pipeline in the `load_from_pipeline` function from `pipelines.base.Pipelines` to `Any`.
I was left with only this choice as i tried using Union of `pipelines.base.Pipelines` and `diffusers.pipelines.DiffusionPipeline` but then Diffusion pipeline doesn't have a _call method due to which i was not able to use it as a type, as we are calling the class. So i was not able to come up with a common type for both these pipelines. 
Let me know if i can use something better as the input type or if you have any better suggestions.

shubhamofbce:
I have also formatted the code using `format_backend.sh` but still the `test / python / linux` is failing not sure why.

abidlabs:
Thanks @shubhamofbce for making the changes! It looks like we need we need to add `diffusers` as a dependency to `test/requirements.in` and `test/requirements.txt` to make the typechecker aware of the diffusers library Would you be able to add to that? (That'll also allow us to write a test for this integration, which I can take care of)

shubhamofbce:
@abidlabs I have updated the requirements.in file and also updated the requirements.txt file using pip compile.
Now there are some tests failing in `routes.py` and `route_utils.py`, None of them are related to my change, not sure how to handle them.
image

gradio/pipelines.py
```diff
@@ -3,268 +3,77 @@
...
-                    )
-                    for i in r
-                ],
-            ),
-        }
+    if str(type(pipeline)).startswith("<class 'transformers.pipelines"):
+        pipeline_info = _handle_transformers_pipeline(pipeline)
+
+    elif str(type(pipeline)).startswith("<class 'diffusers.pipelines"):
+        pipeline_info = _handle_diffusers_pipeline(pipeline)
```
akx, line 26: Relying on the `repr` seems a bit ugly, c.f. something like
```suggestion
    pipeline_class_name = pipeline.__class__.__name__
    if pipeline_class_name.startswith("transformers.pipelines"):
        pipeline_info = _handle_transformers_pipeline(pipeline)
        
    elif pipeline_class_name.startswith("diffusers.pipelines"):
        pipeline_info = _handle_diffusers_pipeline(pipeline)
```

gradio/pipelines.py
```diff
@@ -3,268 +3,77 @@
...
+            # special case for object-detection
+            # original input image sent to postprocess function
+            if isinstance(
+                pipeline,
+                pipelines.object_detection.ObjectDetectionPipeline,
+            ):
+                output = pipeline_info["postprocess"](data, params[0])
+            else:
+                output = pipeline_info["postprocess"](data)
+            return output
```
akx, line 60: Shouldn't this also be in the pipeline helper module?

gradio/pipelines.py
```diff
@@ -3,268 +3,77 @@
...
-            pipeline,
-            (
-                pipelines.text_classification.TextClassificationPipeline,
-                pipelines.text2text_generation.Text2TextGenerationPipeline,
-                pipelines.text2text_generation.TranslationPipeline,
-            ),
-        ):
-            data = pipeline(*data)
-        else:
+        if "tranformers.pipelines" in str(type(pipeline)):
```
akx, line 36: I doubt this code path was ever exercised since there's a typo here.

gradio/pipelines.py
```diff
@@ -3,268 +3,77 @@
...
     interface_info = pipeline_info.copy()
     interface_info["fn"] = fn
     del interface_info["preprocess"]
     del interface_info["postprocess"]
 
     # define the title/description of the Interface
-    interface_info["title"] = pipeline.model.__class__.__name__
+    interface_info["title"] = (
+        pipeline.model.__class__.__name__
+        if str(type(pipeline)).startswith("<class 'transfomers.pipelines")
```
akx, line 75: Typo here too.

gradio/pipelines_helpers.py
```diff
@@ -0,0 +1,541 @@
...
+    except ImportError as ie:
+        raise ImportError(
+            "transformers not installed. Please try `pip install transformers`"
+        ) from ie
+
+    # Handle the different pipelines. The has_attr() checks to make sure the pipeline exists in the
+    # version of the transformers library that the user has installed.
+    if hasattr(transformers, "AudioClassificationPipeline") and isinstance(
+        pipeline, pipelines.audio_classification.AudioClassificationPipeline
+    ):
```
akx, line 25: This repeated segment could be refactored into a helper (repeated nearly similarly for `diffusers`):

```
def is_transformers_pipeline_type(pipeline, class_name: str):
    cls = getattr(transformers, class_name, None)
    return cls and isinstance(pipeline, cls)
```

gradio/pipelines_helpers.py
```diff
@@ -0,0 +1,541 @@
...
+        pipeline_info = {
```
akx, line 26: You can just `return` the dict from each of these branches.

gradio/pipelines_helpers.py
```diff
@@ -0,0 +1,541 @@
...
+    else:
+        raise ValueError(f"Unsupported pipeline type: {type(pipeline)}")
+    return pipeline_info
+
+
+def _handle_diffusers_pipeline(pipeline: Any) -> dict:
+    try:
+        import diffusers
+        from diffusers import pipelines as diffuser_pipelines
+        from PIL import Image
```
akx, line 232: `Pillow` is a required dependency, you don't need to guard this import.

test/requirements.txt
```diff
@@ -32,77 +39,155 @@ certifi==2022.6.15
...
-gradio_pdf==0.0.3
+fonttools==4.48.1
+    # via matplotlib
+fsspec==2024.2.0
+    # via
+    #   gradio-client
+    #   huggingface-hub
+    #   torch
+gradio==4.18.0
+    # via gradio-pdf
```
akx, line 86: Gradio itself shouldn't be a test requirement...

akx (changes requested):

test/requirements.txt
shubhamofbce: Makes sense, I have moved it out of the if block.

gradio/pipelines.py
shubhamofbce: I know it looks ugly but I tried pipeline.__class__.__name__ and it just returns the class name lilke `TextClassificationPipeline`. It doesn't returns the whole path of class like type(). So I can't use your suggestion. I will still look what we can do.

gradio/pipelines.py
shubhamofbce: We can move.

gradio/pipelines.py
shubhamofbce: Fixed it for now, but i will add tests for this to make sure this does not happen again.

gradio/pipelines.py
shubhamofbce: fixed this, will add tests.

gradio/pipelines.py
shubhamofbce: I found a better way, What do you say abut this ?? I tried multiple things out of which this looks the best
```suggestion
     if str(type(pipeline).__module__).startswith("transformers.pipelines"):
        pipeline_info = _handle_transformers_pipeline(pipeline)

    elif str(type(pipeline).__module__).startswith("diffusers.pipelines"):
        pipeline_info = _handle_diffusers_pipeline(pipeline)
```

gradio/pipelines.py
shubhamofbce: I saw the code, this thing is better at its old place pipelines.py, If we move it to the new file, I will have to refactor a lot of things and pass a lot of things to this function which is not required as of now.
This is not something that i wrote, i just added a few lines in it but still i think this is good at its original place only. So I am not moving it to helper file. Let me know what you think.

gradio/pipelines_helpers.py
shubhamofbce: Updated this

gradio/pipelines_helpers.py
shubhamofbce: done

gradio/pipelines_helpers.py
shubhamofbce: refactored this in the latest commit, thanks for this suggestion.

gradio/pipelines.py
shubhamofbce: Fixed the typo and tried a transformer pipeline and diffuser pipeline locally and they are working fine.

test/requirements.txt
shubhamofbce: I have removed this from the file, but I may need your help here as i am using pip-compile for the first time and it added a lot of dependencies in this file. Can you please take a look if it is correct or not.

shubhamofbce:
@akx I have fixed the code as per your suggestions, Also i have tested locally a transformer pipeline and a diffuser pipeline to make sure everything is working as expected.
I just need your help in fixing test `requirements.txt`, I have updated it using `pip compile` but since it added a lot of dependencies so i am not sure if its fully correct so i will ask you to look it carefully and let me know if there are issues in it or what should i do to make sure its correct. I was expecting a few dependencies as i am adding support for diffusers.
Let me know if you have any other comments or suggestion on this PR.

gradio/pipelines.py
shubhamofbce: Added this

shubhamofbce:
@akx and @abidlabs I have also added unit tests one for load transformers from pipeline and one to load diffusers from pipeline. This is to make sure there are no typos introduced while refactoring code. Please let me know if more such tests are required.

gradio/pipelines_helpers.py
akx: Why assign to `pipeline_info`? Just `return {`... 😄

gradio/pipelines_helpers.py
```diff
@@ -0,0 +1,535 @@
...
+                label="Input",
+                render=False,
+            ),
+            "outputs": components.Label(label="Class", render=False),
+            "preprocess": lambda i: {"inputs": i},
+            "postprocess": lambda r: {i["label"].split(", ")[0]: i["score"] for i in r},
+        }
+        return pipeline_info
+
+    elif is_transformers_pipeline_type(pipeline, "AutomaticSpeechRecognitionPipeline"):
```
akx, line 43: None of these need to be `elif`s since each branch returns directly.

gradio/pipelines_helpers.py
```diff
@@ -0,0 +1,535 @@
...
+        ) from ie
+
+    def is_transformers_pipeline_type(pipeline, class_name: str):
+        cls = getattr(transformers, class_name, None)
+        return cls and isinstance(pipeline, cls)
+
+    # Handle the different pipelines. The has_attr() checks to make sure the pipeline exists in the
+    # version of the transformers library that the user has installed.
+    if is_transformers_pipeline_type(pipeline, "AudioClassificationPipeline"):
+        pipeline_info = {
```
akx, line 30: Why assign to `pipeline_info`? Just `return {`... 😄

gradio/pipelines_helpers.py
```diff
@@ -0,0 +1,535 @@
...
+                        i["label"],
+                    )
+                    for i in r
+                ],
+            ),
+        }
+        return pipeline_info
+
+    else:
+        raise ValueError(f"Unsupported pipeline type: {type(pipeline)}")
```
akx, line 225: This should be unwrapped from the `else`.

```suggestion
    raise ValueError(f"Unsupported pipeline type: {type(pipeline)}")
```

gradio/pipelines_helpers.py
```diff
@@ -0,0 +1,535 @@
...
+            "diffusers not installed. Please try `pip install diffusers`"
+        ) from ie
+
+    def is_diffusers_pipeline_type(pipeline, class_name: str):
+        cls = getattr(diffusers, class_name, None)
+        return cls and isinstance(pipeline, cls)
+
+    # Handle diffuser pipelines
+    if is_diffusers_pipeline_type(pipeline, "StableDiffusionPipeline"):
+        pipeline_info = {
```
akx, line 244: Same here for all these branches: 

* `return {`...
* `if` instead of `elif`

gradio/pipelines_helpers.py
```diff
@@ -0,0 +1,535 @@
...
+                "num_inference_steps": num_inf_steps,
+                "guidance_scale": g_scale,
+                "noise_level": noise_level,
+            },
+            "postprocess": lambda r: r["images"][0],
+        }
+        return pipeline_info
+
+    else:
+        raise ValueError(f"Unsupported pipeline type: {type(pipeline)}")
```
akx, line 535:
```suggestion
    raise ValueError(f"Unsupported pipeline type: {type(pipeline)}")
```

test/requirements.txt
```diff
@@ -4,213 +4,367 @@
 #
 #    pip-compile --output-file=requirements.txt requirements.in
```
akx, line 5: You should just run this command with the new requirement added to `requirements.in`, not with `--upgrade`. Also make sure to restore this `requirements.txt` from the main branch before you do that.

test/test_pipelines.py
```diff
@@ -2,6 +2,7 @@
 import transformers
 
 import gradio as gr
+from gradio.interface import Interface
```
akx, line 5: Everything else in this file is using `gr.Interface`; the new tests should probably too.

test/test_pipelines.py
```diff
@@ -22,3 +23,26 @@ def test_interface_in_blocks():
...
+    assert io.input_components[0].label == "Context"
+    assert io.input_components[1].label == "Question"
+    assert io.output_components[0].label == "Answer"
+    assert io.output_components[1].label == "Score"
+
+
+def test_diffusers_load_from_pipeline():
+    from diffusers import StableDiffusionPipeline
+
+    pipe = StableDiffusionPipeline.from_pretrained("runwayml/stable-diffusion-v1-5")
```
akx, line 42: Does this mean the test will download the weights for SD1.5? If so, that should be avoided.

akx (changes requested):

gradio/pipelines_helpers.py
shubhamofbce: done

gradio/pipelines_helpers.py
shubhamofbce: 😄

test/requirements.txt
shubhamofbce: Followed the exact same steps now and generated the requirements.txt file. Marking this as resolved.

test/test_pipelines.py
shubhamofbce: I have removed this test for now.

shubhamofbce:
@akx @abidlabs Since this is a major refactor now so I have added unit tests for all possible cases that we have in pipelines_helpers.py. You can look at  648e408b07143f5a816580599d85a1441b3c77d3 and let me know if anything else is required from my end.

abidlabs:
Thanks @shubhamofbce I'll review and get this merged in this week!

abidlabs (approved):
Great PR @shubhamofbce! Thank you so much for taking our feedback into account. I made some small tweaks, but overall, this looks good to me.

Will merge in once CI passes

abidlabs:
@shubhamofbce your contribution is now part of the official diffusers docs: https://huggingface.co/docs/diffusers/main/en/api/pipelines/stable_diffusion/overview#create-web-demos-using-gradio

Thanks again for the great PR!

shubhamofbce:
Thanks @abidlabs, looking forward for more such useful additions to gradio.

Issue #7179, opened 2024-01-26
`HH:MM:SS` time code in `gr.Audio`


jjshoots:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
When using the `gr.Audio` feature, for audio files over 1 hour, the timestamp is shown as `MM:SS`, so if the audio track is 1 hour and 15 minutes long, it shows 75:00. Would it be possible for this to be shown as `HH:MM:SS` instead?

**Describe the solution you'd like**  
Show timecode in `gr.Audio` as `HH:MM:SS`.

abidlabs:
I agree that would look a bit nicer. Would you be able to open a PR to fix this? I think we'd need to make changes here:

https://github.com/gradio-app/gradio/blob/d548202d2b5bd8a99e3ebc5bf56820b0282ce0f5/js/video/shared/VideoControls.svelte#L27

and

https://github.com/gradio-app/gradio/blob/ded5256c4a39a84155b9c3d7f4e8e7060d798186/js/audio/recorder/AudioRecorder.svelte#L64

arian81:
@abidlabs I can work on this if @jjshoots doesn't want to

abidlabs:
Go for it @arian81! Thanks

jjshoots:
@arian81 Are you still tackling this? I can do it if need be within the next couple days.

arian81:
> @arian81 Are you still tackling this? I can do it if need be within the next couple days.

I was working on it because I couldn't get the video player time to change format but I think it's because it's using the browser's built in player. I pushed the changes I had.

PR #7270, opened 2024-02-01
Add hours indicator to video and audio

jjshoots:
## Description

This converts the `MM:SS` time indicator to `HH:MM:SS`.

Closes: #7179

### DIFF
--- a/.changeset/eight-webs-spend.md
+++ b/.changeset/eight-webs-spend.md
@@ -1,7 +1,7 @@
 ---
-"@gradio/audio": minor
-"@gradio/video": minor
-"gradio": minor
+"@gradio/audio": patch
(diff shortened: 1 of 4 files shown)


abidlabs:
Ah looks like one more place needs to be fixed @jjshoots: https://github.com/gradio-app/gradio/blob/21cfb0acc309bb1a392f4d8a8e42f6be864c5978/js/video/shared/Player.svelte#L80

(we should really write a single helper method to avoid duplication)

abidlabs (changes requested):
Just the changes mentioned above please

abidlabs:
Let's close this in favor of https://github.com/gradio-app/gradio/pull/7274. @jjshoots I'll credit you in that PR

PR #7275, opened 2024-02-01
improve visual ci config

pngwn:
## Description

Closes: #(issue)

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.github/actions/changes/action.yml
+++ b/.github/actions/changes/action.yml
@@ -148,4 +148,4 @@ runs:
         type: ${{ inputs.type }}
         init: true
         mergeable: ${{ steps.pr.outputs.source_branch == 'main' || steps.pr.outputs.mergeable }}
-        commit_status: ${{ inputs.commit_status }}
+        commit_status: ${{ inputs.type == 'visual' && steps.should_run.outputs.should_run == 'false' || inputs.commit_status }}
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy+test-visual.yml
```diff
@@ -86,18 +86,6 @@ jobs:
...
-        uses: "gradio-app/github/actions/commit-status@main"
-        with:
-          type: "visual"
-          token: ${{ secrets.GITHUB_TOKEN }}
-          pr: ${{ needs.changes.outputs.pr_number }}
-          sha: ${{ needs.changes.outputs.sha }}
-          name: "test / visual"
-          result: ${{ job.status }}
-          job_id: "test-visual"  
-          url: ${{ steps.publish-chromatic.outputs.buildUrl }}
```
pngwn, line 100: We don't need this any more, the chromatic action will take care of everything if the job 'should run'

.github/actions/changes/action.yml
```diff
@@ -148,4 +148,4 @@ runs:
         type: ${{ inputs.type }}
         init: true
         mergeable: ${{ steps.pr.outputs.source_branch == 'main' || steps.pr.outputs.mergeable }}
-        commit_status: ${{ inputs.commit_status }}
+        commit_status: ${{ inputs.type == 'visual' && steps.should_run.outputs.should_run == 'false' || inputs.commit_status }}
```
pngwn, line 151: We have to do this here. If the job doesn't need to run we are making a status with the same name (so we can pass the required check), if it _does_ need to run we don't make a status and let the chromatic action handle it.

Issue #7264, opened 2024-02-01
Width and Fill of .Label Incorrect


codydh:
### Describe the bug

When creating a `.Label()`, both the width and fill are set to the percentage passed in, creating the following effect in Safari:
![Screenshot_2024-01-31_at_3 44 17_PM](https://github.com/gradio-app/gradio/assets/11252/8c4af333-771b-4dc3-93df-927f1fa10851)
The following HTML is an example of one of the bars created by the reproduction included:
```
<meter aria-labelledby="meter-text" class="bar svelte-1pq4gst" min="0" max="100" value="97" aria-label="97%" style="width: 97%; background: var(--stat-background-fill);"></meter>
```
Modifying this so that either (1) the value is set to 100% and the width varies, or (2) the width is set to 100% and the value varies (probably more correct) fixes the issue.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

gr.Interface(
    fn=None,
    inputs=None,
    outputs=gr.Label({
        "Potential": 0.97,
        "Growth": 0.96,
        "Tacos": 0.06
    })
).launch()
```

### Screenshot

Included Above

### Logs

```shell
N/A
```

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Darwin
gradio version: 4.16.0
gradio_client version: 0.8.1

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.2.0
fastapi: 0.109.0
ffmpy: 0.3.1
gradio-client==0.8.1 is not installed.
httpx: 0.26.0
huggingface-hub: 0.20.3
importlib-resources: 6.1.1
jinja2: 3.1.3
markupsafe: 2.1.4
matplotlib: 3.8.2
numpy: 1.26.3
orjson: 3.9.12
packaging: 23.2
pandas: 2.2.0
pillow: 10.2.0
pydantic: 2.6.0
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0.1
ruff: 0.1.15
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.9.0
typing-extensions: 4.9.0
uvicorn: 0.27.0.post1
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2023.10.0
httpx: 0.26.0
[...]

PR #7276, opened 2024-02-01
Adjust Value to Fix .Label Display

codydh:
## Description

This adjusts the way .Label bars are created. Previously, both the `value` and the `width` of the var were set to the appropriate value, causing displays on Firefox and Safari that were wrong (see bug).

Closes: #7264

## 🎯 PRs Should Target Issues

## Tests


dawoodkhan82:
Thanks for the fix @codydh. Although I don't see the original issue, even on safari/firefox. There's no background bar. Maybe @abidlabs can confirm

codydh:
> Thanks for the fix @codydh. Although I don't see the original issue, even on safari/firefox. There's no background bar. Maybe @abidlabs can confirm

Strange! I have included the Firefox screenshot here just for reference. Slightly different appearance than Safari, but same (I believe) bug. For what it's worth, this is a stock Firefox install on macOS Sonoma 14.3.

Screenshot 2024-02-02 at 8 22 14 AM

abidlabs:
I was able to repro actually on Safari! And confirm the fix. Thanks a bunch @codydh

Issue #4001, opened 2023-04-28
Allow `gr.Chatbot` to fill all height of rest of space


liaoweiguo:
- [ * ] I have searched to see if a similar issue already exists.

I can use
`chatbot = gr.Chatbot(elem_id="chatbot").style(height=600)`

but it can not adapt to diff screen size

**Is your feature request related to a problem? Please describe.**  

**Describe the solution you'd like**  

**Additional context**

dawoodkhan82:
If you remove setting a specific height, it will dynamically resize the chatbot to fit screen sizes. Try `chatbot = gr.Chatbot(elem_id="chatbot")` without adding style

liaoweiguo:
> If you remove setting a specific height, it will dynamically resize the chatbot to fit screen sizes. Try `chatbot = gr.Chatbot(elem_id="chatbot")` without adding style


I donot thing so. There is a gap between components and footer

like this
![截图 2023-04-28 18-28-55](https://user-images.githubusercontent.com/22187224/235124271-b012c539-daaf-4ec8-b481-04cf560c4a6e.png)

abidlabs:
I get the same behavior as @liaoweiguo -- it looks like there's a fixed height even if no height is specified @dawoodkhan82

liaoweiguo:
any hints?

dawoodkhan82:
@bent-verbiage @liaoweiguo can you check if this pr fixes your issue? #4041

aliabid94:
We currently do not support a concept of an element taking up the "remaining vertical space". Can you explain the scenario where this is helpful?

Your best bet would be using custom CSS. For example
```python
CSS ="""
.contain { display: flex; flex-direction: column; }
#component-0 { height: 100%; }
#chatbot { flex-grow: 1; }
"""

with gr.Blocks(css=CSS) as demo:
    chatbot = gr.Chatbot(elem_id="chatbot")
    msg = gr.Textbox()
    clear = gr.Button("Clear")
```
allows the chatbot to expand to fill the "remaining vertical space", though it may expand if there are more messages.

joedavis:
> We currently do not support a concept of an element taking up the "remaining vertical space". Can you explain the scenario where this is helpful?

How about accessibility? When the chatbot window grows, other elements can end up moving around, until eventually it settles at a fixed size taking up most of the screen anyway.

It's possible to minimize this motion by setting a fixed large size for the chatbot window, but obviously that doesn't take into account different screen sizes.

[...]

PR #7278, opened 2024-02-02
Full height chatbot (and other visual improvements)

aliabid94:
ChatInterfaces (and other Blocks optionally) take full screen height. Rearranged interface a little:

Recording 2024-02-01 at 16 48 17

Closes: https://github.com/gradio-app/gradio/issues/4001


gradio/blocks.py
```diff
@@ -504,6 +504,7 @@ def __init__(
         css: str | None = None,
         js: str | None = None,
         head: str | None = None,
+        fill_vertical_space: bool = False,
```
abidlabs, line 507: What about just renaming this parameter `full_height`. Its less verbose and clearer than `fill_vertical_space` imo

gradio/chat_interface.py
```diff
@@ -112,6 +112,7 @@ def __init__(
             theme=theme,
             js=js,
             head=head,
+            fill_vertical_space=True,
```
abidlabs, line 115: This should definitely be a parameter in `ChatInterface` to allow users to opt-out

gradio/chat_interface.py
```diff
@@ -73,9 +73,9 @@ def __init__(
         analytics_enabled: bool | None = None,
         submit_btn: str | None | Button = "Submit",
         stop_btn: str | None | Button = "Stop",
-        retry_btn: str | None | Button = "🔄  Retry",
-        undo_btn: str | None | Button = "↩️ Undo",
-        clear_btn: str | None | Button = "🗑️  Clear",
+        retry_btn: str | None | Button = "Retry",
+        undo_btn: str | None | Button = "Undo",
+        clear_btn: str | None | Button = "Clear",
```
abidlabs, line 78: I'm not a huge fan of removing the emojis. For two reasons: now the buttons look too monotonous to me, and we lose the benefits of the buttons being immediately clear and good for i18n. Why remove them?

abidlabs:
Very cool @aliabid94! This was a requested feature a while back so I added it to the issue. I do have some concerns about the changes to `gr.ChatInterface`, see above

gradio/chat_interface.py
pngwn: I don't love the emojis, but I would like us to keep some 'symbols' for the reasons you mentioned. I think some nice clean icons would be better.

pngwn:
This seems to contain some of @freddyaboulton's changes from a previous PR. I think if you rebase/ merge main it should clean itself up.

gradio/chat_interface.py
abidlabs: True now that we support icons in buttons, we could have some cleaner icons instead by default

gradio/blocks.py
aliabid94: full_height is not accurate though. All apps take full height with the footer. This parameter specifically means that top level components, by default, expand to fill the vertical space. Open to other names, cc: @pngwn

gradio/chat_interface.py
aliabid94: ok will revert but I think the emojis are ugly

aliabid94:
> This seems to contain some of @freddyaboulton's changes from a previous PR. I think if you rebase/ merge main it should clean itself up. 

I'm bad at git. Gonna close this PR and try again

Issue #6055, opened 2023-10-22
Show Space ID instead of full URL with replica name in Space API documentation


hysts:
- [x] I have searched to see if a similar issue already exists.

It would be nice if the API documentation showed Space ID instead of the full URL.
([internal discussion](https://huggingface.slack.com/archives/C02QZLG8GMN/p1697858097614789))

Currently, the API documentation of Spaces shows the full URL that includes a replica name. But replica names will be changed when restarted, and it leads to 404 error. Also, I think Space ID is shorter and easier to remember than full URL.

![](https://github.com/gradio-app/gradio/assets/25161192/0385e04a-a256-43e9-8592-5b4081989ddc)

Issue #2569, opened 2022-10-30
Mention that HF token may be needed in the view API Docs page for Gradio apps running on Spaces


freddyaboulton:
- [ ] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  

Currently, the Api Docs page does not mention that you need to pass in your Huggingface token in the header when you are loading a private space. 

Related thread: https://discuss.huggingface.co/t/how-to-test-api-of-a-private-space/25211/2

**Describe the solution you'd like**  
We should modify our api docs page to mention that users may have to pass a HF token when accessing a private space.

PR #7292, opened 2024-02-02
Improvements to API Docs

abidlabs:
Some small improvements to the API docs as I start to look at the clients.

Closes: https://github.com/gradio-app/gradio/issues/2569
Closes: https://github.com/gradio-app/gradio/issues/6055

Here's how it looks locally:

image

-------

Here's how it looks on Spaces (note that we now use Space ID instead of the ugly replica URL):

### DIFF
--- a/js/app/src/api_docs/ApiDocs.svelte
+++ b/js/app/src/api_docs/ApiDocs.svelte
@@ -189,13 +189,13 @@
 					{/each}
 				</div>
 
-				<p style="margin:15px 0px;">
+				<p class="padded">
 					1. Install the client if you don't already have it installed.
 				</p>
 
 				<InstallSnippet {current_language} />
 
-				<p style="margin:15px 0px;">
+				<p class="padded">
 					2. Find the API endpoint below corresponding to your desired function
 					in the app. Copy the code snippet, replacing the placeholder values
 					with your own input data.
@@ -358,5 +358,9 @@
 
 	a.underline {
 		text-decoration: underline;
+	}
+
+	p.padded {
+		padding: 15px 0px;
 	}
 </style>
--- a/js/app/src/api_docs/ApiDocs.svelte
+++ b/js/app/src/api_docs/ApiDocs.svelte
@@ -22,6 +22,16 @@
 	export let dependencies: Dependency[];
 	export let root: string;
 	export let app: Awaited<ReturnType<typeof client>>;
+	export let space_id: string | null;
+
+	const js_spaces_docs =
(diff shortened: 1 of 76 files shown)


js/app/src/api_docs/ApiDocs.svelte
```diff
@@ -148,26 +159,18 @@
...
-				api_count={dependencies.filter((dependency) => dependency.show_api)
-					.length}
-			/>
+			<ApiBanner on:close root={space_id || root} {api_count} />
 		</div>
 		<div class="docs-wrap">
 			<div class="client-doc">
 				<p>
-					Use the <a
-						href="https://gradio.app/docs/#python-client"
```
abidlabs, line 164: These links were broken but it was never reported because I don't think anyone realized that they were links in the first place. The styling does not make it very clear its a link.

js/app/src/api_docs/ApiDocs.svelte
```diff
@@ -183,11 +186,25 @@
 						</li>
 					{/each}
 				</div>
+
+				<p class="padded">
+					1. Install the client if you don't already have it installed.
```
abidlabs, line 191: I think explicitly numbering the steps makes it more inviting to users.

js/app/src/api_docs/ApiDocs.svelte
```diff
@@ -335,4 +352,12 @@
 		padding: var(--size-3);
 		padding-top: 0;
 	}
+
+	a {
+		text-decoration: underline;
+	}
+
+	p.padded {
```
aliabd, line 360: The font sizes are really off: 

Screen Shot 2024-02-05 at 11 38 17 AM

The text looks really small. and the code snippets for the libraries are too large with too much padding on the bottom. it's weird. 

I suggest adding: 

```
p {
		font-size: var(--text-lg);
	}
```
and modifying the .library css to be 
```
.library {
		border: 1px solid var(--border-color-accent);
		border-radius: var(--radius-sm);
		background: var(--color-accent-soft);
		padding: var(--size-1);
		color: var(--color-accent);
		font-size: var(--text-md);
	}
```

looks much better:

Screen Shot 2024-02-05 at 11 45 28 AM

aliabd (approved):
Thanks for the fixes @abidlabs lgtm, made one comment

abidlabs:
Ah nice @aliabd thanks for catching both of those issues. Will fix soon

abidlabs:
Thanks again @aliabd!

Issue #6708, opened 2023-12-08
Ability to remove individual files from `gr.File()` component


apolinario:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
Currently the `gr.File()` component, once you upload files you can only remove all files at once
<img width="328" alt="image" src="https://github.com/gradio-app/gradio/assets/788417/a2853d35-3eef-402c-849e-254fd91d6bbf">

**Describe the solution you'd like**  
I'd like to be able to individually remove each uploaded file.

**Additional context**  
If uploaded something wrong by accident is very frustrating to remove all to reupload the correct ones.

abidlabs:
Makes sense, this would be a nice addition

PR #7299, opened 2024-02-03
Added remove button for every file in file preview, to remove individual file in gr.File()

shubhamofbce:
## Description

I have added a remove file button which can be used to remove individual file from file preview page. We already have a close button which removes all the file, so we will not show the remove file button if the file size is 1 in preview. It will be visible only if the size is greater than 1.

Closes: #6708 

https://github.com/gradio-app/gradio/assets/45156638/f82f360d-2c5d-431a-ba5f-5a0337e85674

## 🎯 PRs Should Target Issues

## Tests


shubhamofbce:
@hannahblair Can you please review the changes, I have just added a clear button in multiple file preview page using which you can clear individual files.

dawoodkhan82:
Looks great! Maybe the clear button for the entire component should be a little bigger. Just to differentiate between it and the individual clear buttons.

abidlabs:
Thanks @shubhamofbce for the great PR! From a UI perspective, I think having all of the "X" icons is starting to look a little bit cluttered. As @dawoodkhan82 mentioned, it would be better if there was a visual differentiation between the "X" icon to clear all of the files and the "X" icon to clear a single file.

I would suggest instead having a more subtle "X" right after the file name to clear the individual file

abidlabs:
Thanks @hannahblair for pushing the UI changes

abidlabs:
One other issue I noticed is that even though the files disappear when you click on the "x", the value of the File component does not actually change, and so you see this behavior:


https://github.com/gradio-app/gradio/assets/1778297/9f8656ba-bc02-4281-af2a-9457e637bd19

@shubhamofbce would you be able to push a fix for this?

abidlabs (changes requested):
See above issue. Repro is just:

```py
import gradio as gr

gr.Interface(lambda x:x, gr.File(file_count="multiple"), gr.File()).launch()
```

shubhamofbce:
@abidlabs Thanks for pointing this out. I will take a look at how we can do this.

shubhamofbce:
@hannahblair You are the best... Thanks for the contribution.


Sorry I was not well so didn't got chance to work on it.

shubhamofbce:
@aliabd This is up for review now. We have fixed the issue that you pointed.
Thanks to @hannahblair for making most of the changes.

abidlabs (approved):
Great PR @shubhamofbce and @hannahblair, LGTM

mberco-quandl:
Hi there team, thanks for your work! @abidlabs 

If a interface user uploads a file and their client connection is terminated for any reason this leads to memory leaks as there is no way to navigate back to the page and delete the files. Not to mention this option only outsources the burden of memory management to a manual action taken by users. Consider a containerized application where it is not easy to configure scheduled delete/cleanup jobs. This leak is especially cumbersome if the files are truly ephemeral and there is no need whatsoever to persist the temporary files. Consider the following example that illustrates what the intent is (note I am unsure of the gr.File.name syntax for referencing the full path):
`csv_upload = gr.File(
    file_count=1,
    file_types=["csv"],
    label="Upload CSV file here:",
    type="filepath",
)
csv_as_json = gr.Text(
    label="JSON representation of uploaded csv after deletion",
    info="to be hidden once implemented",
)

@csv_upload.upload(inputs=[csv_upload], outputs=[csv_as_json])
def convert_csv_to_json_then_delete():
    with open(csv_upload.name, "r") as csv_input:
        csv_reader = csv.DictReader(csv_input)
        data = [row for row in csv_reader]
    os.remove(csv_upload.name)
    json_string = json.dumps(data, indent=4)
    return json_string`
    
Currently thinking the best way to workaround this is to have the user specify a filepath and handle the conversion to json format entirely within python to avoid the gr.File class because of this unavoidable memory leak.

Generally speaking I'm thinking that supporting some kind of custom upload behaviour would be really desirable. It could even be an `after_upload` event listener that listens for the default upload to complete, enabling the app to immediately read and delete the file once it is in memory. This is also leak proof as interruptions to the upload would trigger a container restart and refresh the temp storage anyways.

PR #7304, opened 2024-02-05
fix: Code component not accepting type Callable on value parameter

daviirodrig:
## Description

The Code component can accept a Callable as value, as said on the documentation, but the typing on that parameter wasn't following that.


abidlabs:
Thank you @daviirodrig for the fix!

PR #7313, opened 2024-02-05
Expand chatinterface to full window height

aliabid94:
ChatInterfaces (and other Blocks optionally) take full screen height. Rearranged interface a little:

Recording 2024-02-01 at 16 48 17

Closes: https://github.com/gradio-app/gradio/issues/4001


aliabid94:
PR #7278 got in a weird git state so trying again.

gradio/chat_interface.py
```diff
@@ -170,85 +173,84 @@ def __init__(
             if description:
                 Markdown(description)
```
aliabid94, line 175: looks like a lot of change but just moved the chat buttons before the textbox and got rid of the parent panel column

gradio/blocks.py
```diff
@@ -514,6 +515,7 @@ def __init__(
             css: Custom css as a string or path to a css file. This css will be included in the demo webpage.
             js: Custom js or path to js file to run when demo is first loaded. This javascript will be included in the demo webpage.
             head: Custom html to insert into the head of the demo webpage. This can be used to add custom meta tags, scripts, stylesheets, etc. to the page.
+            expand_to_height: Whether to vertically expand top-level child components to the height of the window. If True, expansion occurs when the scale value of the child components >= 1.
```
abidlabs, line 518: This means that the Gradio app will be the full height of the window, right? I think `full_height` would be a better parameter name

abidlabs: Should we add this parameter to `Interface` and `TabbedInterface` too btw?

abidlabs: So just to understand, the same `scale` parameter is used to determine how much the height of each child component should expand by? If so, we should update the docstring for `scale` for each component to mention this. If not, how does this work if you have multiple child elements whose height can vary?

abidlabs:
@aliabid94 the full-screen `gr.ChatInterface` looks great but I'm confused about how to use the `expand_to_height` parameter more generally. For example, neither of these work:

```py
import gradio as gr

with gr.Blocks(expand_to_height=True) as demo:
    gr.Chatbot(scale=2)
    gr.Chatbot(scale=2)
    
demo.launch()
```

(This^ takes up more than the full height of the page)

Nor this:

```py
import gradio as gr

with gr.Blocks(expand_to_height=True) as demo:
    gr.Textbox(scale=2)
    gr.Textbox(scale=2)
    
demo.launch()
```

(This^ takes up less than the full height)

Some additional explanation of this parameter in the Guides would be useful imo.

gradio/blocks.py
aliabid94: >  I think full_height or fill_height would be a better parameter name

Renamed to `fill_height`. 

> Should we add this parameter to Interface and TabbedInterface too btw?

I don't think it's necessary. 

> If so, we should update the docstring for scale for each component to mention this.

Ok updated docstring of every component.

aliabid94:
> For example, neither of these work:
>
> ```
> import gradio as gr
> 
> with gr.Blocks(expand_to_height=True) as demo:
>     gr.Chatbot(scale=2)
>     gr.Chatbot(scale=2)
>     
> demo.launch()
> ```
> 
>
> (This^ takes up more than the full height of the page)

Chatbots by default have a minimum height of 400 if no height is specified. If an app extends past the window, fill_height does not shrink any element, the height of the elemnts are respected first. To "fix" this, simply set the height to something smaller, and the Components will expand.


> Nor this:
> 
> ```
> import gradio as gr
> 
> with gr.Blocks(expand_to_height=True) as demo:
>     gr.Textbox(scale=2)
>     gr.Textbox(scale=2)
>     
> demo.launch()
> (This^ takes up less than the full height)
> ```

Fixed. Was an issue with the autogenerated Form component not respecting scale.

abidlabs:
Very nice @aliabid94! LGTM

abidlabs (approved):

abidlabs:
Merging so that we can do a release

kostinmax:
Could you please provide a working example of how to use this (fill_height) new property?

PR #7319, opened 2024-02-06
Handle httpx.InvalidURL when setting non-URL article strings

nopperl:
## Description

The `article` parameter of `gradio.Interface` takes both an URL and markdown/HTML text. The constructor first tries to use article string as URL, and if that throws an `httpx.RequestError`, the article string is used directly. This does not work if the article string is longer than `httpx._urlparse.MAX_URL_LENGTH` or contains non-printable characters, in which case `httpx.InvalidURL` is raised. This PR handles this error.

### DIFF
--- a/.changeset/red-turkeys-make.md
+++ b/.changeset/red-turkeys-make.md
@@ -1,0 +1,5 @@
+---
+"gradio": minor
+---
+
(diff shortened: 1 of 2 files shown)


freddyaboulton (approved):
This looks good. Thanks @nopperl !

Issue #7241, opened 2024-01-31
Gallery Postprocessing Blocks the Demo


omerXfaruq:
### Describe the bug

When the function returns the list, gradio takes a very long time to render/postprocess the images. And meanwhile all other operations on other clients' actions are blocked (image upload, example tries).

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

https://huggingface.co/spaces/omerXfaruq/FindYourTwins

Use advanced section with max number of images and try to use the demo from another tab at the same time.

### Screenshot

Blocked image upload on other client:

![image](https://github.com/gradio-app/gradio/assets/44067394/b0f48079-41df-40e7-873a-8804e442ee6f)

### Logs

execution of the function takes 1 second, hence I'm sure that 20-30 seconds of blocking is caused by postprocessing.

Finished embedding: 1706744955.7943506
Finished fetching images: 1706744955.9093854
Finished creating list: 1706744956.575465

```shell
        async def find_similar_faces(image, count):
            inputs = extractor(images=image, return_tensors="pt")
            outputs = model(**inputs)
            embed = outputs.last_hidden_state[0][0]
            import time
            print(f"Finished embedding: {time.time()}")
            result = await index.query(
                vector=embed.tolist(), top_k=max(1, min(MAX_K, int(count)))
            )
            print(f"Finished fetching images: {time.time()}")
            images = [dataset["train"][int(vector.id)]["image"] for vector in result]
            print(f"Finished creating list: {time.time()}")
            return images
```

### System Info

```shell
gradio version: 4.16.0
```

### Severity

Blocking usage of gradio

freddyaboulton:
Hey @omerXfaruq ! Are `extractor` and `model` coroutines? Since they come from transformers, I think they're normal functions that would block the event loop. I think you need to run those in a separate thread or rework the function so that it's not a coroutine.

I would test that out by duplicating your space but I don't have upstash api keys.

omerXfaruq:
> Hey @omerXfaruq ! Are `extractor` and `model` coroutines? Since they come from transformers, I think they're normal functions that would block the event loop. I think you need to run those in a separate thread or rework the function so that it's not a coroutine.
> 
> I would test that out by duplicating your space but I don't have upstash api keys.

Hello,

The extractor and model don't take a long time, please check the timestamps I've shared in logs.

Hmm, for duplicating the space you can check this blog post on how to create an index.
But I think it's related to postprocessing.

[...]

PR #7327, opened 2024-02-06
Run pre/post processing in threadpool

freddyaboulton:
## Description

Closes: #7241

In some cases, the component pre/postprocessing takes a long time (like big galleries or chatbots) and it can block the event loop. We can run the processing in the thread pool to avoid the block. Tested this with the following demo and running two concurrent users:

```python
import gradio as gr
import time

def get(image):
    output = [image] * 30
    return output

gr.Interface(get, gr.Image(type="pil"), gr.Gallery(), concurrency_limit=2).launch()
```

### main - notice how the demo on the right will freeze until the first one finishes

https://github.com/gradio-app/gradio/assets/41651716/dfdf1c0f-6bc0-4525-9cfe-90105fe7a036

### pr - both finish at the same time

https://github.com/gradio-app/gradio/assets/41651716/0ee6d6e4-bf20-41a8-9282-a579ddfdf02c

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/hot-taxis-jump.md
+++ b/.changeset/hot-taxis-jump.md
@@ -1,0 +1,5 @@
+---
+"gradio": patch
+---
+
+fix:Run pre/post processing in threadpool
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -1521,6 +1521,9 @@ Received outputs:
 
         return data
 
+    def run_fn_batch(self, fn, batch, fn_index, state):
+        return [fn(fn_index, list(i), state) for i in zip(*batch)]
+
     async def process_api(
         self,
         fn_index: int,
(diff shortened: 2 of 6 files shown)


abidlabs:
The real issue is that Gallery takes criminally long to postprocess 30 images, which is really not that many. I can't reproduce the 15-second wait, but it does take several seconds on my local computer. Can we do something to improve that?

Otherwise, these changes look good. I tested with `gradio/lite` as well because IIRC threads don't play too nicely with pyodide but there wasn't any issue running the demos that I tried. So LGTM

whitphx:
Thank you @abidlabs , that's so helpful you tested this with Lite.

`anyio.run_sync` is mocked in Lite as below so theoretically it shouldn't have any good and bad effect on Lite. It lgtm as @abidlabs 's test worked well :D
https://github.com/gradio-app/gradio/blob/49d9c48537aa706bf72628e3640389470138bdc6/js/wasm/src/webworker/index.ts#L151-L162

freddyaboulton:
> Can we do something to improve that?

I think the only thing we could do is maybe saving them in parallel. I just profiled running gallery postprocess on 30 images and half the time is spent converting to bytes (for the hash) and the other half is calling PIL.save. Let me see how easy it is to save them in parallel. 

image

freddyaboulton:
@abidlabs I made two speed-ups to the gallery postprocess - 1) only doing one call to `PIL.Image.save` instead of two and 2) running the images concurrently with a ThreadPoolExecutor. 

I stil lhink we should be running pre/postprocess in the threadpool because if for whatever reason we add a new component/custom component whose pre/postprocess takes more than a second the whole page will freeze which is not ideal.

This should be good for another look.

abidlabs (approved):
Looks great. Beautiful speedup @freddyaboulton!

freddyaboulton:
Thanks for the review!

PR #7329, opened 2024-02-06
Remove torch from dependencies

freddyaboulton:
## Description

Don't think it's used anywhere

## 🎯 PRs Should Target Issues

## Tests


freddyaboulton:
But are we using transformers anywhere in our tests? They just passed without it. Let me see if I can force the functional tests to run to be sure.

abidlabs:
We use `transformers` to test `gr.Interface.from_pipeline`:

https://github.com/gradio-app/gradio/blob/287fe6782825479513e79a5cf0ba0fbfe51443d7/test/test_pipelines.py#L7

freddyaboulton:
Why do the tests keep passing then? cc @pngwn 

Will close - thanks @abidlabs !

PR #7332, opened 2024-02-06
Hotfix for: `stt_or_tts` demo

abidlabs:


demo/stt_or_tts/run.py
```diff
@@ -10,13 +10,12 @@
...
     examples=tts_examples,
     description="Give me something to say!",
-    cache_examples=False
 )
 
 stt_demo = gr.load(
     "huggingface/facebook/wav2vec2-base-960h",
     title=None,
-    inputs="mic",
+    inputs=gr.Microphone(type="filepath"),
```
abidlabs, line 18: The input component needs to have `type="filepath"` to preprocess the values correctly. Not sure how/if it was working before

gradio/external.py
```diff
@@ -153,7 +153,7 @@ def custom_post_binary(data):
         fn = custom_post_binary
     # example model: facebook/wav2vec2-base-960h
     elif p == "automatic-speech-recognition":
-        inputs = components.Audio(sources=["upload"], type="filepath", label="Input")
+        inputs = components.Audio(type="filepath", label="Input")
```
abidlabs, line 156: Unrelated, but I think its better not to restrict the sources by default

freddyaboulton (approved):
Nice @abidlabs !

PR #7333, opened 2024-02-06
Stop using deprecated pydantic config class and filter gradio warnings from tests

freddyaboulton:
## Description

We're using a deprecated pydantic api (https://docs.pydantic.dev/2.5/migration/#changes-to-config) so migrating to the new api.

In addition, warnings raised by our own code (gr.LogoutButton, etc) are clogging up the pytest warning log which makes it hard to review/see legit warnings. These warnings are usually raised because we're testing an API we've deprecated but not removed.

## 🎯 PRs Should Target Issues

## Tests


abidlabs:
cool thanks @freddyaboulton!

abidlabs (approved):

freddyaboulton:
Thanks @abidlabs !

Issue #7149, opened 2024-01-25
Updating the root parameter in the fileexplorer component does not change the front-end display of the file directory.


lifeisgoodcdj:
### Describe the bug

The gr.fileexplorer() object changes the root parameter when used as a callback function, but it does not actually change the root directory of the component for file searches.
I want to know how to change the root directory of gr.fileexplorer()  

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

with gr.Blocks() as app:
    objFileExplorer = gr.FileExplorer(root="/home/file_path1/", show_label=True, height=400, interactive=True)
    objButtonClear = gr.Button("clear")

    @objButtonClear.click(inputs=[objFileExplorer], outputs=[objFileExplorer])
    def fnButtonClearClick(listSelectedFiles):
        print(listSelectedFiles)
        return gr.update(root="/home/file_path2/", value="", show_label=True, height=600, interactive=False)

if __name__ == "__main__":
    app.queue().launch(server_name='0.0.0.0',
                       server_port=40123,
                       share=False,
                       debug=True,
                       max_threads=2,
                       )

```

### System Info

```shell
gradio==4.4.0
```

### Severity

Blocking usage of gradio

abidlabs:
Hi @lifeisgoodcdj is this still an issue in the latest version of Gradio (4.15.0)? I believe we’ve fixed it

lifeisgoodcdj:
The issue still exists. I have tried the 4.15.0 version of gradio, but I still cannot update or switch directories actually

abidlabs:
cc @aliabid94

liucw2012:
i also need to update root dir ~

freddyaboulton:
Hi @lifeisgoodcdj - the `root` parameter is deprecated for the FileExplorer. If you use `root_dir` in both the definition of `objFileExplorer` and `gr.update` it should work. If you want to use `root`, you can return a new instance of `gr.FileExplorer` in your update function. 

Going to close as this seems like a corner case with a deprecated API and fixing would require adding some component specific logic to the app-level postprocessing which would not be ideal for maintainability.

Issue #6627, opened 2023-12-02
The `interactive` parameter in `gr.FileExplorer` has no effect. It should either be correctly documented or the feature should be added


rrfaria:
- [X] I have searched to see if a similar issue already exists.

### Is your feature request related to a problem? 

 interactive -  if True, will allow users to upload a file; if False, can only be used to display files. If not provided, this is inferred based on whether the component is used as an input or output.

This functionality apparently does nothing. I used like this:

```python
gr.FileExplorer(root='./voices', height=400, interactive=True, label="voices available", file_count="multiple", glob="**/*.*", ignore_glob="*.db,")
```

Either it should be documented correctly, or the ability to upload files should be added:

This functionality could be like google colab file explorer, enabling users to handle their files without need direct access to server.
And also could allow see empty folders on root folder I defined

![image](https://github.com/gradio-app/gradio/assets/11916455/747cf85c-5000-4b46-b5fe-92046f654641)

pngwn:
I'm torn on this, making it part of the component would be a much slicker UI/ UX for users but also we need to sensible defaults so users don't accidentally give everyone access to the machine. This can technically be accomplished without adding these features to the FileExplorer but the UI/UX won't be as nice.

rrfaria:
@pngwn I believe root param may handle access to files. the way it is today it only depends on the developer make a mistake giving access to wrong path. Adding buttom to upload file only for root folder defined on param and give an option to limit extension types may protect mistakes.

PR #7337, opened 2024-02-07
Improve File Explorer performance

aliabid94:
Previously we would load the entire directory structure at the root directory and send that to the frontend when the fileexplorer component loaded. This could be prohibitively expensive for higher level directories to crawl through all the nested subdirectories, taking on the order of minutes to load. Now the API call only returns the content of a single level, and when opening a folder, we load the content of that folder. See gif below, showing network requests as well.

Recording 2024-02-06 at 21 08 25

Also fixed root_dir update issue.
Fixes: #7149 

Demo code to repro fix (change root_dir values ofc):
```python
import gradio as gr

with gr.Blocks() as app:
    file_explorer = gr.FileExplorer(
        value=["/Users/aliabid/Desktop/gradio/demo/hello_blocks/run.py", "/Users/aliabid/Desktop/gradio/demo/change_vs_input/run.py"],
        root_dir="/Users/aliabid/Desktop/gradio/demo/", 
        file_count="multiple",
        height=200
    )
    select_btn = gr.Button("Select")
    selected_file = gr.Textbox(label="Selected")

    select_btn.click(lambda x:x, inputs=file_explorer, outputs=selected_file)

    change_root_btn = gr.Button("Change Root")
    change_root_btn.click(lambda: gr.FileExplorer(root_dir="/Users/aliabid/Desktop/gradio/"), outputs=file_explorer)


if __name__ == "__main__":
    app.launch()
```


abidlabs:
Nice! There are some bugs related to the `FileExplorer` maybe it would make sense to include fixes in this PR:

* https://github.com/gradio-app/gradio/issues/7149
* https://github.com/gradio-app/gradio/issues/6627

demo/file_explorer_component_events/run.py
```diff
@@ -10,22 +10,18 @@
                         choices=[str(base_root / "dir1"), str(base_root / "dir2"),
                                  str(base_root / "dir3")])
         with gr.Group():
-            dir_only_glob = gr.Checkbox(label="Show only directories", value=False)
-            ignore_dir_in_glob = gr.Checkbox(label="Ignore directories in glob", value=False)
+            txt_only_glob = gr.Checkbox(label="Show only text files", value=False)
+            ignore_txt_in_glob = gr.Checkbox(label="Ignore text files in glob", value=False)
 
     fe = gr.FileExplorer(root=str(base_root / "dir1"),
```
abidlabs, line 16:
```suggestion
    fe = gr.FileExplorer(root_dir=str(base_root / "dir1"),
```

gradio/components/file_explorer.py
```diff
@@ -156,80 +155,40 @@ def postprocess(self, value: str | list[str] | None) -> FileExplorerData | None:
...
-            file = str(f.relative_to(self.root_dir))
-            if f.is_dir():
-                file += os.path.sep
-            files_with_sep.append(file)
-
-        tree = make_tree(files_with_sep)
-        return tree
+        return folders + files
 
     def _safe_join(self, folders):
```
abidlabs, line 193: Didn't notice this before, but we should use `routes.safe_join` instead of reimplementing this method

gradio/components/file_explorer.py
```diff
@@ -59,7 +58,7 @@ def __init__(
...
             file_count: Whether to allow single or multiple files to be selected. If "single", the component will return a single absolute file path as a string. If "multiple", the component will return a list of absolute file paths as a list of strings.
             root_dir: Path to root directory to select files from. If not provided, defaults to current working directory.
-            ignore_glob: The glob-tyle pattern that will be used to exclude files from the list. For example, "*.py" will exclude all .py files from the list. See the Python glob documentation at https://docs.python.org/3/library/glob.html for more information.
+            ignore_glob: The glob-style pattern that will be used to exclude files from the list. For example, "*.py" will exclude all .py files from the list. See the Python glob documentation at https://docs.python.org/3/library/glob.html for more information.
```
abidlabs, line 61: I would add "case-sensitive" here and above, as files on Windows are not case-sensitive, but fnmatch.fnmatch is
```suggestion
            ignore_glob: The glob-style, case-sensitive pattern that will be used to exclude files from the list. For example, "*.py" will exclude all .py files from the list. See the Python glob documentation at https://docs.python.org/3/library/glob.html for more information.
```

gradio/components/file_explorer.py
```diff
@@ -156,80 +155,40 @@ def postprocess(self, value: str | list[str] | None) -> FileExplorerData | None:
         return FileExplorerData(root=root)
 
     @server
-    def ls(self, _=None) -> list[dict[str, str]] | None:
+    def ls(self, subdirectory: list | None = None) -> list[dict[str, str]] | None:
         """
         Returns:
             tuple of list of files in directory, then list of folders in directory
```
abidlabs, line 161:
```suggestion
            a list of dictionaries, where each dictionary represents a file or subdirectory in the given subdirectory
```

demo/file_explorer_component_events/run.py
abidlabs, line 1: I would add this to https://github.com/gradio-app/gradio/blob/1086542f17271d170eb7934aa12ead1defacf8a7/scripts/copy_demos.py#L9, we don't currently have any fileexplorer demo that gets deployed to spaces

abidlabs:
Great PR @aliabid94! Everything looks good to me, just left a few nits on the code above.

On the UI side, I preferred having the checkboxes next to everything, even if they weren't clickable. Now, you have this empty space, which looks odd imo:

image

abidlabs (approved):

gradio/components/file_explorer.py
```diff
@@ -59,7 +58,7 @@ def __init__(
...
             root_dir: Path to root directory to select files from. If not provided, defaults to current working directory.
-            ignore_glob: The glob-tyle pattern that will be used to exclude files from the list. For example, "*.py" will exclude all .py files from the list. See the Python glob documentation at https://docs.python.org/3/library/glob.html for more information.
+            ignore_glob: The glob-style pattern that will be used to exclude files from the list. For example, "*.py" will exclude all .py files from the list. See the Python glob documentation at https://docs.python.org/3/library/glob.html for more information.
             label: The label for this component. Appears above the component and is also used as the header if there are a table of examples for this component. If None and used in a `gr.Interface`, the label will be the name of the parameter this component is assigned to.
             every: If `value` is a callable, run the function 'every' number of seconds while the client connection is open. Has no effect otherwise.sed (e.g. to cancel it) via this component's .load_event attribute.
             show_label: if True, will display label.
```
abidlabs, line 64: Can we properly document `interactive` parameter below and fix https://github.com/gradio-app/gradio/issues/6627?

pngwn:
Yeah I liked the disabled checkboxes, they make the alignment tidier and hint more clearly that selecting the entire folder is explicitly not allowed on this app.

aliabid94:
I think it's hard to tell the checkboxes are disabled when all of them are disabled, e.g.

Screenshot 2024-02-13 at 4 35 08 PM

but don't feel strongly, will revert

gradio/components/file_explorer.py
aliabid94: gonna keep it separate for now, `routes.safe_join` returns HTTPExceptions that don't make sense in this context, and has a different signature.

shimizust:
Thanks so much for the fix @aliabid94 ! I was just running into this issue

PR #7344, opened 2024-02-07
New demo gradio space for Darts timeseries forecasting. Based on the published Prophet demo space.

ivelin:
## Description

Darts is becoming a popular timeseries forecasting library. This is a small contribution to integrating Darts with Gradio for timeseries demo apps.

## 🎯 PRs Should Target Issues

Targets https://github.com/unit8co/darts/issues/2220

## Tests

Backend linting script applied.


pngwn:
I think this would be better as its own space on hugging face spaces, rather than a demo in the gradio repo. These demos aren't that discoverable and we don't really want to add more as they are primarily used to testing purposes and documentation.

abidlabs:
I agree with @pngwn -- for visibility, I'd recommend creating a standalone Hugging Face Spaces, and working on increasing the visibility of that Space.

ivelin:
Sounds good:
https://huggingface.co/spaces/ivelin/darts-demo

Issue #7335, opened 2024-02-07
CSS selectors that begin with `.dark` are not applied


oobabooga:
### Describe the bug

Since https://github.com/gradio-app/gradio/pull/6738, the CSS selectors in custom user styles receive a prefix to increase their specificity. I find that styles starting in `.dark` do not receive that prefix.

cc @pngwn

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

Launch the demo below and access `http://127.0.0.1:7860/?__theme=dark`. The text should be blue, but it's red.

```python
import gradio as gr

css = """
.dark textarea {
   color: blue;
   -webkit-text-fill-color: blue;
}

textarea {
   color: red;
   -webkit-text-fill-color: red;
}
"""

with gr.Blocks(css=css) as demo:
  gr.TextArea(value="This is a test.")

demo.launch()
```

### System Info

```shell
gradio==4.17.0
```

### Severity

Blocking usage of gradio

abidlabs:
cc @aliabid94 @hannahblair as well for visibility

pngwn:
I'm quite surprised this is happening. I'm not surprised it breaks things but I am surprised that a specific class name is being ignored. It isn't supposed to discriminate.

oobabooga:
Thanks for the fix!!

oobabooga:
I have found another related issue and reported it at https://github.com/gradio-app/gradio/issues/7390.

PR #7355, opened 2024-02-08
Ensure CSS `.dark` rule selectors are applied 

hannahblair:
## Description

We're currently generating the CSS selector for custom CSS like so: 

`gradio-app .gradio-container.gradio-container-${version} .contain ${s.trim()}`

(`s` being the selector)

So for the following example:

```
.dark button {
    color: blue
}
```
The following selector is generated: 

`gradio-app .gradio-container.gradio-container-4-17-0 .contain .dark button`

The problem is that this selector expects `.dark` to be a descendant of `.contain`, but `.dark` is actually set at the `<body>` element - which is a higher ancestral element than `gradio-app`, so this selector isn't actually targeting the `button` element. 

So, the solution I've added is to
- Check if the selector is a dark mode selector with `.includes(".dark")`
- Remove it from the original selector
- Add it at the beginning of the string (if it's a dark mode selector)

Closes: #7335

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/js/imageeditor/ImageEditor.stories.svelte
+++ b/js/imageeditor/ImageEditor.stories.svelte
@@ -50,7 +50,7 @@
 />
 
 <Story
-	name="Image Editor Interactions"
+	name="Image Editor Drawing Interactions"
 	args={{
 		value: {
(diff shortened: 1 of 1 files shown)


pngwn (approved):
makes sense! Thanks for taking care of this @hannahblair!

Issue #7359, opened 2024-02-08
`gr.Markdown()` does not render spaces around links correctly


abidlabs:
### Describe the bug

In 4.x, `gr.Markdown()` does not render spaces around links even when they are included in the Markdown string. E.g.

```py
import gradio as gr

with gr.Blocks() as demo:
    gr.Markdown("""
    ## Visit our [HF space](https://huggingface.co/spaces/lmsys/chatbot-arena-leaderboard) for more analysis!
    """
)
    demo.launch()
```

produces:

<img width="471" alt="image" src="https://github.com/gradio-app/gradio/assets/1778297/85f70a38-8dac-4ca1-9029-250740caee00">

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

See above

### System Info

```shell
4.17
```

### Severity

I can work around it

PR #7361, opened 2024-02-08
Fixes gr.Markdown() does not render spaces around links correctly

dawoodkhan82:
## Description
Removed styling that created issues around any children of headers (links, bold, italics, text, etc.)

See `markdown_example` demo to view changes this causes. 
Closes: #7359

## 🎯 PRs Should Target Issues

## Tests


js/markdown/shared/MarkdownCode.svelte
```diff
@@ -160,14 +160,4 @@
 	span.md :global(.md-header-anchor > svg) {
 		color: var(--body-text-color-subdued);
 	}
-
-	span :global(h1),
```
dawoodkhan82, line 164: @pngwn I'm curious why we needed this? It removes any whitespace from the children of headers. If we do need it I could also add margins to those html elements separately.

pngwn: I think it is for the header links + anchors that are generated with certain settings enabled.

abidlabs:
Ah I see the issue that @pngwn is referring to. It happens when you set `header_links=True`. e.g.

```py
import gradio as gr

with gr.Blocks() as demo:
    gr.Markdown("""
    # abc
    def
    """, header_links=True)
    demo.launch()
```

The links get rendered on a separate line now:

image

abidlabs (changes requested):
See above

abidlabs (approved):
LGTM @dawoodkhan82! I added a story to prevent regressions

PR #7373, opened 2024-02-09
fix spaces comment and remove echos

pngwn:


abidlabs:
Thanks for removing the echos @pngwn. I went through the remaining echos and they all look safe.

However, I still don't see the Spaces link or comment with the wheels on this PR

pngwn:
No, we have to merge into main for them to take effect. It is still triggering the workflow_run workflows that exists on main.

Very secure but as you can see, very tedious too.

abidlabs (approved):

Issue #7372, opened 2024-02-09
When running on Spaces or using share links, root_url is set to localhost instead of the full URL breaking css


abidlabs:
On Spaces and with share links, the `config["root"]` is not set correctly (its set to localhost instead of the full url), which means that Gradio looks for the css and other assets in the wrong place, and the demo ends up looking like this:

<img width="1261" alt="image" src="https://github.com/gradio-app/gradio/assets/1778297/d0ba4587-e78a-416d-9293-ea9b90f8eef1">

PR #7374, opened 2024-02-09
Stop caching root url

abidlabs:
Fixes: #7372

The issue was that we were caching the "root_url" the first time the Gradio app was launched. I did this to speed things up, but the problem is that the same Gradio app can be launched on different URLs. For example, when you create a share link, you also might launch the Gradio app locally first (which would set the root_url to "localhost:7860") and then when you open the share link, the "root_url" will remain localhost:7860.

Now, we no longer cache the root url. I profiled this change and it adds pretty negligible latency for most demos (e.g. `2 ms` for `kitchen_sink`). Does add more latency for very complex apps (`150 ms` for the `theme_builder`) so maybe we should think of a better solution long-term (when we work on reducing the launch time).

Also properly handles https & does some minor cleanup of this code.

You can confirm this works by visiting e.g. https://gradio-pr-deploys-pr-7374-all-demos.hf.space/demo/altair_plot/config and checking the "root" propery. Or to test locally, launch a Gradio demo with `share=True` and visit the `/config` route and confirm that the "root" property is what you expect.

### DIFF
--- a/gradio/routes.py
+++ b/gradio/routes.py
@@ -311,7 +311,7 @@ class App(FastAPI):
         def main(request: fastapi.Request, user: str = Depends(get_current_user)):
             mimetypes.add_type("application/javascript", ".js")
             blocks = app.get_blocks()
-            root_path = route_utils.strip_url(str(request.url))
+            root_path = route_utils.get_root_url(request)
             if app.auth is None or user is not None:
                 config = copy.deepcopy(app.get_blocks().config)
                 config["root"] = root_path
@@ -321,7 +321,7 @@ class App(FastAPI):
                     "auth_required": True,
                     "auth_message": blocks.auth_message,
                     "space_id": app.get_blocks().space_id,
-                    "root": route_utils.strip_url(root_path),
+                    "root": root_path,
                 }
 
             try:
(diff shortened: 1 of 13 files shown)


abidlabs:
Actually wait I think the cached examples are off

abidlabs:
Should be ready now!

gradio/routes.py
```diff
@@ -571,8 +570,8 @@ async def predict(
                     content={"error": str(error) if show_error else None},
                     status_code=500,
                 )
-            root_path = app.get_blocks().config.get("root", "")
-            output = add_root_url(output, route_utils.strip_url(root_path))
+            root_path = route_utils.get_root_url(request)[: -len(f"/api/{api_name}")]
```
abidlabs, line 573: Annoying that I have to do this manually, but if you use e.g. `request.url.path`, it also includes the root url if gradio is mounted within another fastapi app, which we don't want. So I think the most reliable solution is to just remove the suffix manually.

freddyaboulton (approved):
Looks good to me! Thanks @abidlabs !

abidlabs:
Thanks so much for the review @freddyaboulton!

PR #7377, opened 2024-02-09
Make set_documentation_group a no-op

freddyaboulton:
## Description

Also strips the `\@document()` call from custom component templates so authors don't get a warning about the documentation group not being found.

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/rare-months-guess.md
+++ b/.changeset/rare-months-guess.md
@@ -1,0 +1,6 @@
+---
+"gradio": minor
+"gradio_client": minor
+---
+
+feat:Make set_documentation_group a no-op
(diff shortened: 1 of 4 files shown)


abidlabs:
Thanks @freddyaboulton! Reviewing this now. 

I was thinking it might be good to have one or two e2e tests involving custom components to ensure that we don't see this situation again

abidlabs:
Nice @freddyaboulton. I can confirm that this allows working with `UnifiedAudio`. However, I see this warning when I do: `from gradio_unifiedaudio import UnifiedAudio`

image

Full repro:

```py
import gradio as gr
from gradio_unifiedaudio import UnifiedAudio

gr.Interface(lambda x:x, UnifiedAudio(), gr.Textbox()).launch()
```

abidlabs:
Fixed the warning, I'll go ahead and merge this in so that we can do the release. We can add an e2e test with custom components later on

Issue #7370, opened 2024-02-09
httpx.InvalidURL: Invalid non-printable ASCII character in URL


YuryYakhno:
### Describe the bug

# Overview

When `Interface` object is created with `article` parameter set to some multiline text, this causes exception at the moment of object creation:
```text
Traceback (most recent call last):
  File "demo.py", line 102, in <module>
    gr_io = gr.Interface(
  File ".venv/lib/python3.10/site-packages/gradio/interface.py", line 300, in __init__
    article = utils.readme_to_html(article)
  File ".venv/lib/python3.10/site-packages/gradio/utils.py", line 292, in readme_to_html
    response = httpx.get(article, timeout=3)
  File ".venv/lib/python3.10/site-packages/httpx/_api.py", line 196, in get
    return request(
  File ".venv/lib/python3.10/site-packages/httpx/_api.py", line 104, in request
    return client.request(
  File ".venv/lib/python3.10/site-packages/httpx/_client.py", line 815, in request
    request = self.build_request(
  File ".venv/lib/python3.10/site-packages/httpx/_client.py", line 347, in build_request
    url = self._merge_url(url)
  File ".venv/lib/python3.10/site-packages/httpx/_client.py", line 377, in _merge_url
    merge_url = URL(url)
  File ".venv/lib/python3.10/site-packages/httpx/_urls.py", line 115, in __init__
    self._uri_reference = urlparse(url, **kwargs)
  File ".venv/lib/python3.10/site-packages/httpx/_urlparse.py", line 160, in urlparse
    raise InvalidURL("Invalid non-printable ASCII character in URL")
httpx.InvalidURL: Invalid non-printable ASCII character in URL
```

It looks like the same problem was mentioned in #6889.

# Possible Reason

1. New version of `httpx` package [does not support](https://github.com/encode/httpx/discussions/2676) ASCII non-printable characters in URL. `\n` in multiline text is also ASCII non-printable character.
1. It looks like code of [readme_to_html() function](https://github.com/gradio-app/gradio/blob/2244059cdbacb713530a3b760205c5464c05491c/gradio/utils.py#L290) uses `httpx.get()` function improperly - for link validation instead of sending `GET` request.

# Proposed Fix

Validate link with suitable package, like [validators](https://validators.readthedocs.io/en/latest/#module-validators.url), or with regular expression. If link is valid, then send `GET` request by `httpx`.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

[...]

PR #7389, opened 2024-02-12
Fix HTTPX package crash for some values of "article" parameter in the interface

YuryYakhno:
## Description

"article" parameter of the interface class contained not documented feature: downloading file by HTTP(S) protocol if value of the parameter is URL. But new version of `httpx` package is started to expect URL only with printable characters, and produce exception if `\n`, `\t` or other characters are in the "article" content. This PR:
* Documents "automatic downloading" feature
* Checks value of "article" for valid URL separately, and use `httpx` package only for validated URLs
* Updates name of downloading procedure, but it is not in API
* Updates tests to cover more cases for "article" value

Closes: #7370


gradio/utils.py
```diff
@@ -286,13 +286,19 @@ def is_zero_gpu_space() -> bool:
...
+        is_url = all([result.scheme, result.netloc, result.path])
+        is_url = is_url and result.scheme in ["http", "https"]
+    except:
+        is_url = False
+
+    if is_url:
         response = httpx.get(article, timeout=3)
         if response.status_code == httpx.codes.OK:  # pylint: disable=no-member
             article = response.text
-    except (httpx.InvalidURL, httpx.RequestError):
```
freddyaboulton, line 294: I think we should still keep the request error exception catching but otherwise this looks good to me. Thanks @YuryYakhno !

abidlabs: I think this is fine because @YuryYakhno has moved the exception handling up and made it more generic

gradio/utils.py
```diff
@@ -286,13 +286,19 @@ def is_zero_gpu_space() -> bool:
     return os.getenv("SPACES_ZERO_GPU") == "true"
 
 
-def readme_to_html(article: str) -> str:
+def download_if_url(article: str) -> str:
     try:
+        result = urllib.parse.urlparse(article)
+        is_url = all([result.scheme, result.netloc, result.path])
+        is_url = is_url and result.scheme in ["http", "https"]
+    except:
```
abidlabs, line 294: Would be better to not use a bare exception here

gradio/utils.py
abidlabs: Yes and if there is an exception, we can just return the original URL string as before

abidlabs:
Thank you @YuryYakhno! Besides the nits pointed out above, everything else LGTM

abidlabs (approved):

freddyaboulton (approved):

YuryYakhno:
Just noticed that initial problem was fixed in #7319 with handling `httpx.InvalidURL`, so our solution should not change the observed behavior. But it is more robust in case of new exception types in `httpx` package (like it was with `InvalidURL`).

PR #7400, opened 2024-02-12
Small PR to move `generate_notebooks.py` to scripts folder

abidlabs:
Small PR, was bothering me for a long time that `generate_notebooks.py` wasn't in `scripts/` folder with the rest of our scripts

### DIFF
--- a/.github/workflows/test-hygiene.yml
+++ b/.github/workflows/test-hygiene.yml
@@ -44,7 +44,7 @@ jobs:
           repository: ${{ needs.changes.outputs.mergeable == 'true' &&  github.repository || needs.changes.outputs.source_repo }}
       - name: Generate Notebooks
         run: |
-          pip install nbformat && cd python scripts/generate_notebooks.py
+          pip install nbformat && python scripts/generate_notebooks.py
       - name: Print Git Status
         run: echo $(git status) && echo $(git diff)
       - name: Assert Notebooks Match 


abidlabs:
The `test/hygiene` is failing because it is running the test using the script from main, which I believe is intentional for security reasons. But after this is merged, everything should pass

freddyaboulton:
Did we get rid of the automated comment that told you where to run the script from if it failed? Otherwise looks good to me!

freddyaboulton (approved):

.github/workflows/test-hygiene.yml
```diff
@@ -44,7 +44,7 @@ jobs:
           repository: ${{ needs.changes.outputs.mergeable == 'true' &&  github.repository || needs.changes.outputs.source_repo }}
       - name: Generate Notebooks
         run: |
-          pip install nbformat && cd demo && python generate_notebooks.py
+          pip install nbformat && cd python scripts/generate_notebooks.py
```
aliabd, line 47: typo: remove cd

aliabd:
@abidlabs this won't work, you'll have to edit the `generate_notebooks.py` file to change where it looks for the demos. right now it just does `os.getcwd()` because it thinks it's in `/demos`. So it's creating an empty list

abidlabs:
Oops thanks @aliabd will fix

abidlabs:
Should be fixed now! I also cleaned up `generate_notebook.py` a bit

.github/workflows/test-hygiene.yml
freddyaboulton: I think typo still needs to be fixed @abidlabs

abidlabs: 😱

abidlabs: fixed now

pngwn (approved):
Looks good! Thanks @abidlabs!

You'll need to abuse your admin powers to merge this one. 

@freddyaboulton i think that got lost as part of the CI rework. I'll add it back later this week.

abidlabs:
Thanks @pngwn @freddyaboulton and @aliabd!

Issue #6853, opened 2023-12-20
Dropdown doesn't retain value if choice name is changed


Donkie:
### Describe the bug

I need to update the names of choices in the dropdown, which I do using the gr.update. The choices change name correctly, but the already selected value in the dropdown is not retained, despite the value in tuple is unchanged.

I have also tried changing the gr.update to also include `value="IdA"` but that doesn't help.

I expect the dropdown to automatically update the name of the selected item, without it becoming gray.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
from dataclasses import dataclass

import gradio as gr

@dataclass
class Data:
    """Some data."""

    name: str
    key: str

my_data: list[Data] = [Data("Choice A", "IdA"), Data("Choice B", "IdB")]

def tb_edit(current_choice: str, new_name: str):
    for item in my_data:
        if item.key == current_choice:
            item.name = new_name

def update_choices():
    return gr.update(choices=[(c.name, c.key) for c in my_data])

with gr.Blocks() as blocks:
    dd = gr.Dropdown(choices=[(c.name, c.key) for c in my_data])
    tb = gr.Textbox(label="Edit Name")
    tb.submit(
        tb_edit,
        inputs=[dd, tb],
    ).then(
        update_choices,
        outputs=[dd],
    )

blocks.launch()
```

### Screenshot

[Screencast from 2023-12-20 11:40:13.webm](https://github.com/gradio-app/gradio/assets/2332094/5e10237a-d3e6-4f83-a75f-8afe0779c890)

### Logs

```shell
No errors.
```

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Linux
gradio version: 4.11.0
gradio_client version: 0.7.3

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.1.2
fastapi: 0.104.1
ffmpy: 0.3.1
gradio-client==0.7.3 is not installed.
httpx: 0.25.1
huggingface-hub: 0.19.4
importlib-resources: 6.1.1
jinja2: 3.1.2
markupsafe: 2.1.3
matplotlib: 3.8.1
numpy: 1.26.2
orjson: 3.9.10
packaging: 23.2
pandas: 2.1.3
pillow: 10.1.0
pydantic: 2.5.1
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0.1
semantic-version: 2.10.0
[...]

PR #7401, opened 2024-02-12
Retain dropdown value if choices have been changed

abidlabs:
Closes: https://github.com/gradio-app/gradio/issues/6853 and adds a unit test.

### DIFF
--- a/js/dropdown/shared/Dropdown.svelte
+++ b/js/dropdown/shared/Dropdown.svelte
@@ -8,8 +8,9 @@
 
 	export let label: string;
 	export let info: string | undefined = undefined;
-	export let value: string | number | (string | number)[] | undefined = [];
-	let old_value: string | number | (string | number)[] | undefined = [];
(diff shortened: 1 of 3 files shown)


js/dropdown/shared/Dropdown.svelte
```diff
@@ -92,9 +93,19 @@
 	}
 
 	$: {
-		if (choices !== old_choices || input_text !== old_input_text) {
-			filtered_indices = handle_filter(choices, input_text);
+		if (choices !== old_choices) {
+			set_input_text();
 			old_choices = choices;
+			filtered_indices = handle_filter(choices, input_text);
+			if (!allow_custom_value && filtered_indices.length > 0) {
```
abidlabs, line 100: A little bit of code duplication from below, but I think its easier to follow this way

abidlabs:
Thanks @hannahblair for the review! The issue was that we weren't calling `set_input_text();` if `choices` was updated.

Issue #7388, opened 2024-02-11
`.change` or `.input` event isn't triggered while typing in `gr.Dropdown` when `filterable=True`


apolinario:
### Describe the bug

Currently, `.change` or `.input` events are only triggered in `gr.Dropdown` if an option is selected and not while being typed (when `filterable=True` allows typing) as it is for the `gr.Textbox()`. 

As the `.select` event already fulfils the purpose of only triggering an event if a valid option is selected - `.change`/`.input` become a bit redundant for such use-case. This is the reason why this is filled as a bug and not a feature request (imo as the `.select` event already behaves for option selection, it is a bug that `.change`/`.input` isn't triggered while the user is typing with `filterable=True`

Addressing this bug allows for cool new use-cases such as changing the options in the dropdown component in real time based on what the user is typing (e.g.: quicksearch). [Internal thread with more context](https://huggingface.slack.com/archives/C02QZLG8GMN/p1707400718463809) 

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr
import requests
import time

def search_huggingface_datasets(query):
    if not query:  # If the query is empty, return an empty list
        return []
    url = f"https://huggingface.co/api/quicksearch?q={query}&type=dataset&limit=20"
    print(url)
    response = requests.get(url)
    if response.status_code == 200:
        data = response.json()
        print(data)
        dataset_names = [d['id'] for d in data['datasets']]
        return dataset_names
    else:
        return ["Error fetching datasets"]

def update_dropdown(query):
    datasets = search_huggingface_datasets(query)
    return gr.update(choices=datasets, visible=True)

with gr.Blocks() as demo:
    with gr.Row():
        dataset_dropdown = gr.Dropdown(label="Datasets", choices=[""], elem_id="dataset_list", interactive=True, filterable=True, allow_custom_value=True)
    dataset_dropdown.change(update_dropdown, inputs=dataset_dropdown, outputs=dataset_dropdown, queue=False, show_progress="hidden")
demo.launch(debug=True)
```

Colab: https://colab.research.google.com/drive/1NynI5ZtQs3F8RAoCq3GhUYMU19ji3iEB?usp=sharing

### Screenshot

https://github.com/gradio-app/gradio/assets/788417/c5d6985b-1433-40e0-b983-be4ef8b5f2f7

### System Info

```shell
gradio==4.18.0
```

### Severity

I can work around it

abidlabs:
[...]

PR #7404, opened 2024-02-12
Add `.key_up` event listener to `gr.Dropdown()`

abidlabs:
Closes: https://github.com/gradio-app/gradio/issues/7388 cc @apolinario 

Note: I had initially called it `.type()` but that conflicts with the `Dropdown.type` attribute. Also, because typing does not automatically update the `value` of the `Dropdown` component, added a new `KeyDownData` class to transmit the relevant data.


.changeset/shaggy-hairs-peel.md
```diff
@@ -0,0 +1,6 @@
+---
+"@gradio/dropdown": minor
+"gradio": minor
+---
+
+fix:Add `.type` event listener to `gr.Dropdown()`
```
aliabid94, line 6:
```suggestion
fix:Add `.key_down` event listener to `gr.Dropdown()`
```

aliabid94:
Keep in mind key_down is triggered on any key press, including arrow keys, shift, etc. which do not affect the actual typed value. Also, key_down is triggered before the value of an input field changes (unlike key up, which triggers after the input value changes)

aliabid94:
This does not work on main the intended use case - allowing dynamically changing choices when a user types in the dropdown text:

```python
import gradio as gr

def test(value, key_down_data: gr.KeyDownData):
    input_val = key_down_data.input_value
    return gr.Dropdown(choices=[f"{input_val} {i}" for i in range(10)])

with gr.Blocks() as demo:
    d = gr.Dropdown(["abc", "def"], allow_custom_value=True)
    t = gr.JSON()
    d.key_down(test, d, d)

if __name__ == "__main__":
    demo.launch()
```

abidlabs:
> Keep in mind key_down is triggered on any key press, including arrow keys, shift, etc. which do not affect the actual typed value. 

Yes, I considered excluding those but figured that it would be better to just provide the `.key` data and let the developer handle it themselves

> Also, key_down is triggered before the value of an input field changes (unlike key up, which triggers after the input value changes)

Yup but again I think this is fine because we provide both the `.input_value` which is the state of the component before the key press, and the key data and the developer can put them together.

aliabid94:
> Also, key_down is triggered before the value of an input field changes (unlike key up, which triggers after the input value changes)
> 
> > Yup but again I think this is fine because we provide both the .input_value which is the state of the component before the key press, and the key data and the developer can put them together.

Why not just use the key up method

abidlabs:
> Why not just use the key up method

Tbh no strong reason, I can change it

.changeset/shaggy-hairs-peel.md
abidlabs: Weird, maybe b/c this wasn't initially targeting `main`, now it is

abidlabs:
Ok I've changed it to use `.key_up()`. The modified version of your demo @aliabid94:

```py
import gradio as gr

def test(value, key_up_data: gr.KeyUpData):
    input_val = key_up_data.input_value
    return gr.Dropdown(value=input_val, choices=[f"{input_val} {i}" for i in range(10)])

with gr.Blocks() as demo:
    d = gr.Dropdown(["abc", "def"], allow_custom_value=True)
    t = gr.JSON()
    d.key_up(test, d, d, show_progress="hidden")

if __name__ == "__main__":
    demo.launch()
```

still doesn't work, but for a different reason. The problem is the issue that @pngwn brought up before which is when a component is used as an output component, [it it set to be static and then converted back to interactive](https://github.com/gradio-app/gradio/blob/d8753edbe0fb444d98ca55d61b1ce65ed1fc3ccf/js/app/src/Blocks.svelte#L535), so it loses its state. In this case, it means that the Dropdown choices are hidden and the Dropdown is blurred. This leads to an unusuable UX.

There are a few different ways we could fix it, the simplest being that we don't do this interactive/static switch when `show_progress="hidden"`. But I believe that @pngwn has a more comprehensive fix planned soon, so I'll leave that out of the scope of this PR.

abidlabs:
Ok I've made a temporary workaround by not disabling output components when the event is of type `key_up` (we do the same thing for `focus` actually) so that we can get this in. We're discussing a more [comprehensive solution internally](https://huggingface.slack.com/archives/C02SPHC1KD1/p1707788541757289).

@aliabid94 the demo I shared above should work now. Look ok to merge?

js/dropdown/shared/Multiselect.svelte
```diff
@@ -259,6 +260,11 @@
 					bind:value={input_text}
 					bind:this={filter_input}
 					on:keydown={handle_key_down}
+					on:keyup={(e) =>
+						dispatch("key_up", {
+							key: e.key,
+							input_value: input_text
+						})}
```
pngwn, line 267: I would just use keypress here, the issue with `keyup` is that it won't fire if you hold the key _down_ even though the text will be displayed which isn't very intuitive.

gradio/events.py
```diff
@@ -516,6 +531,10 @@ class Events:
         "load",
         doc="This listener is triggered when the {{ component }} initially loads in the browser.",
     )
+    key_up = EventListener(
+        "key_up",
+        doc="This listener is triggered when the user presses a key while the {{ component }} is focused.",
+    )
```
pngwn, line 537: I think `keypress` is a little more intuitive, partly because we might want to change precisely _when_ this even is fired (keypresses have multiple stages) but also because 'press' is a little more generic. key press phases are pretty close to an implementation detail.

aliabid94:
I remember users have asked for adding keyboard shortcuts to their apps before, this api could lead to something like:
```python
with gr.Blocks() as demo:
  def keylistener(evt: KeyPressEvent):
    if evt.keyCode == ...

  demo.keypress(keylistener)
```

allowing keypress listeners to a whole app

pngwn:
Sometimes they want the keypress to do some discreet UI thing that isn't really controllable via python atm though. We have had lots of image editor related keyless request which would be almost impossible to implement via a backend function and would lead to a really bad experience anyway.

.changeset/tasty-spies-spend.md
```diff
@@ -3,4 +3,4 @@
 "gradio": patch
 ---
 
-fix:Retain dropdown value if choices have been changed
+fix:Retain dropdown value if choices have been changed
```
aliabid94, line 6: again, not the correct message

aliabid94 (approved):
LGTM, agree that we should be using KeyPress everywhere instead of KeyUp

abidlabs:
Thanks for the reviews @aliabid94 and @pngwn. I'll replace with `.keypress` then.

abidlabs:
Actually @aliabid94 @pngwn the `keypress` event is deprecated and `keydown` is the recommended alternative: https://developer.mozilla.org/en-US/docs/Web/API/Element/keypress_event

Let me know if I'm missing anything!

Issue #7405, opened 2024-02-12
Model3D Gaussian Splatting Support


dylanebert:
- Add official support for Gaussian Splatting in the `Model3D` component
- Reflect functionality in the `Model3DGS` [custom component](https://github.com/dylanebert/gradio-splatting)

PR #7406, opened 2024-02-12
Model3D Gaussian Splatting

dylanebert:
## Description

- Adds support for viewing gaussian splatting scenes using gsplat.js
- This is done by adding a `Canvas3DGS` svelte component that mirrors `Canvas3D`, then using a conditional on the file extension

## Considerations

- If `.ply` support is later added for meshes, then inferring mesh vs. splat from the extension alone may not be sufficient.

Closes: #7405 

cc @abidlabs @hysts


pngwn:
Hmm, is this really something we want in core? This feel specialised enough to be its own custom component.

demo/model3D/files/bonsai.splat
pngwn, line 1: This binary is too large for the repo, we should store it in S3 or something and link to it.

demo/model3D/files/luigi.ply
pngwn, line 1: This is borderline but I personally think it is too large, it should also be stored elsewhere.

js/model3D/shared/Model3D.svelte
```diff
@@ -3,6 +3,7 @@
 	import { BlockLabel, IconButton } from "@gradio/atoms";
 	import { File, Download, Undo } from "@gradio/icons";
 	import Canvas3D from "./Canvas3D.svelte";
+	import Canvas3DGS from "./Canvas3DGS.svelte";
```
pngwn, line 6: We should dynamically import these components only when necessary. They both bring significant dependencies with them.

js/model3D/shared/Model3DUpload.svelte
```diff
@@ -6,6 +6,7 @@
 	import { File } from "@gradio/icons";
 	import type { I18nFormatter } from "@gradio/utils";
 	import Canvas3D from "./Canvas3D.svelte";
+	import Canvas3DGS from "./Canvas3DGS.svelte";
```
pngwn, line 9: Same here.

pngwn (changes requested):

demo/model3D/files/bonsai.splat
dylanebert: Moved to dataset url

js/model3D/shared/Canvas3DGS.svelte
```diff
@@ -0,0 +1,124 @@
...
+		load();
+		frameId = requestAnimationFrame(frame);
+	}
+
+	onMount(() => {
+		if (value != null) {
+			reset_scene();
+		}
+		mounted = true;
+	});
```
pngwn, line 115: We will want to return a clean up function  from `onMount` to remove scene and cancel any pending frames.

dylanebert: Thanks, I have now added this

js/model3D/shared/Model3D.svelte
dylanebert: I'm not sure how to do this, would appreciate help with this if the gradio team decided to go through with this PR.

pngwn: This will probably give some type errors atm but I'd need to check exactly what they are, dynamic imports are a bit weird to type.

pngwn: We could abstract this with a fancy function but it probably isn't worth it atm, the duplication is fine. I should create a utility to make this simpler, we do a lot of dynamic importing.

pngwn:
@dylanebert I think we are good to move ahead with this.

js/model3D/shared/Model3D.svelte
dylanebert: Thanks, I have revised it to use dynamic imports. I had to make use of `any` which I usually avoid in typescript.

js/model3D/shared/Model3DUpload.svelte
dylanebert: Added 👍

dylanebert:
I believe all comments have now been addressed :)

Issue #7331, opened 2024-02-06
Make it easier to call gradio apps programmatically from outside python/js


freddyaboulton:
- [x] I have searched to see if a similar issue already exists.

Moving an internal discussion to github so that we don't forget and hear other perspectives.

Context: https://huggingface.slack.com/archives/C02SPHC1KD1/p1706886135997019

**Is your feature request related to a problem? Please describe.**  

We've added some complexity to the clients and API recently for good reason but at this point it's unreasonable for the community to write clients in their own language or use gradio programmatically outside of python/js.

**Describe the solution you'd like**  
Some things we discussed

> What if we exposed every API endpoint in two ways:
the default "optimized" way which requires the use of the Clients (currently the sse_v2 protocol)
a "basic" way which lets users query an endpoint using basic http sse requests in a simple format that is curl-friendly.

> I think best way to add a simpler route without supporting two completely  pathways is to also allow getting updates to a specific event id
Send data for event and session hash via POST request and receive event id (same as before)
Send GET request with event id to get output of event if complete

> Can we have a separate POST/GET request that returns an event stream you listen to for updates? It's not introducing a whole separate path and should be less burdensome than polling

**Additional context**

divyanshluthra:
are u guys just waiting to see the response of how many people buy API endpoints service that u started?

PR #7407, opened 2024-02-13
Support simplified event api for event only updates

aliabid94:
Partial fix for https://github.com/gradio-app/gradio/issues/7331

--------

Made several updates that make querying the API easier:

- Can listen to the updates of just one event id, instead of streaming updates of all events of a session_hash 
- A session hash is not required and will be given if not assigned. This allows users to make "sessionless" requests, so they don't have to worry about creating a session_hash for one off requests
- Support api_names for `queue/join`

See the following example for making a request via Python:

```python
import httpx
import json

evt = httpx.post('http://localhost:7860/queue/join/greet', json={
    "data": ["john"],
})
evt = dict(evt.json())

async with httpx.AsyncClient() as client:
    async with client.stream('GET', 'http://localhost:7860/queue/data', params=evt) as response:
        async for chunk in response.aiter_text():
            msg = json.loads(chunk[5:])
            if msg["msg"] == "process_completed":
                output = msg["output"]["data"]
                print(output)
```

I did not modify either client because
- the JS client is used by the gradio frontend, and we want that to listen to session_hashs to reduce the amount of connections, due to the browser limit
- all our backend tests use the python client, which should use the same routing as the JS client


abidlabs:
Nice @aliabid94 I agree that these are nice simplifications but I wonder if we can simplify even further. Some thoughts:

1. This approach is not ideal for generators, since the messages returned by the `/queue/data` endpoint will look like:

"data": [[["append", [], "l"]]
"data": [[["append", [], "o"]]
"data": [[["append", [], "v"]]
...

2. Do we need both the `event_id` and the `session_hash` that is returned from `/queue/join`? Is `event_id` not enough?

3. The Python code you've written is nice, but it involves having to parse the returned data in a way that would be annoying to do in cURL / command line.

For those reasons^ instead of re-using the `/queue/join` and `/queue/data` endpoints, I'd suggest creating a separate endpoints for this. Say:

* POST to `/request/{api_name}` and in return, you get a unique URL that points to an [EventSource](https://developer.mozilla.org/en-US/docs/Web/API/EventSource) 
* GET the URL above and you get a stream of outputs -- just the outputs, nothing else to keep it dead simple to use with cURL / command line (ok fine maybe we include some metadata so we can manage errors)

Replicate does something like this fwiw: https://replicate.com/docs/streaming

abidlabs:
> I will need to send heartbeats because the request may wait a while in queue, and there needs to be a message within a certain timeout or the connection will drop. So I will only send yielded output and heartbeats over this connections

Ok then maybe a simple protocol like:

{"status": "pending", result: ""}
{"status": "pending", result: ""}
{"status": "generating", result: "h"}
{"status": "generating", result: "e"}
{"status": "generating", result: "l"}
{"status": "generating", result: "l"}
{"status": "generating", result: "o"}
{"status": "complete", result: "hello"}

abidlabs:
One other thing we should consider is how we want to handle files (i.e. any components that use the FileData type, such as gr.Image). Right now, the clients upload them to the Gradio app server and convert them to FileData objects before sending them to the prediction API. It would be a pain if a user had to manually provide a FileData dict using cURL / command line

freddyaboulton:
Nice the approach @aliabid94 ! Agree with @abidlabs that we should not use the diffed output for generators in this new route. I think simplifying the returned fields like in @abidlabs 's comment [here](https://github.com/gradio-app/gradio/pull/7407#issuecomment-1942934580) would be useful but I think the `result`/`data` field should always be a list so that the API doesn't change depending on the number of output components. 

Also, I think we need `session_hash` for stateful demos like chatbots, no?

aliabid94:
Ok redid this whole PR. New endpoints mimic replicate's structure:
- POST `/call/{api_name}`
upload data here, recieve event id
- GET `/call/{api_name}/{event_id}`
Get stream of events. Supported events: "generating", "complete", "error", "heartbeat"
This stream shows actual values in generating, not diffs

Also refactored all server messages to use pydantic structures, instead of throwing around jsons. Pretty important to typecheck what our API's return, and easier to refactor as well.

Example using curl:
Recording 2024-02-14 at 20 24 42

abidlabs:
This looks awesome @aliabid94. Now that we have this simplified API, could we provide cURL examples in our View API page? Would be great for the visibility of this API:

image

pngwn (approved):
Looks great to me! Only thing to note is that it would be good to bring these classes across to the frontend as typescript interfaces.

gradio/utils.py
```diff
@@ -1035,6 +1035,9 @@ def __setitem__(self, key: K, value: V) -> None:
             self.popitem(last=False)
         super().__setitem__(key, value)
 
+    def __getitem__(self, key: K) -> V:
```
abidlabs, line 1038: Is there any reason you explicitly add this?

abidlabs:
Works great, thanks @aliabid94! 

For anyone else who wants to test, these are the cURL commands:

* `curl -X POST -H "Content-Type: application/json" -d '{"data": ["Gradio", 4]}' http://127.0.0.1:7860/call/predict`
* `curl -X GET -H "Accept: text/event-stream" http://127.0.0.1:7860/call/predict/1897b393c005417e90104dda525d1701CC`

I'm excited to get this working with files and then surfacing this on the view API docs (but that can a separate PR)!

abidlabs (approved):

PR #7410, opened 2024-02-13
remove static while pending behaviour

pngwn:
## Description

Closes: #(issue)

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/demo/chatinterface_streaming_echo/run.py
+++ b/demo/chatinterface_streaming_echo/run.py
@@ -1,10 +1,11 @@
 import time
 import gradio as gr
 
+
 def slow_echo(message, history):
     for i in range(len(message)):
-        time.sleep(0.05)
-        yield "You typed: " + message[: i+1]
+        yield "You typed: " + message[: i + 1]
+
 
 demo = gr.ChatInterface(slow_echo).queue()
 
--- a/js/app/src/Blocks.svelte
+++ b/js/app/src/Blocks.svelte
@@ -525,22 +525,7 @@
 				.on("status", ({ fn_index, ...status }) => {
 					tick().then(() => {
 						const outputs = dependencies[fn_index].outputs;
-						outputs.forEach((id) => {
-							if (
-								instance_map[id].props.interactive &&
-								status.stage === "pending" &&
-								dep.targets[0][1] !== "focus"
-							) {
-								pending_outputs.push(id);
-								instance_map[id].props.interactive = false;
-							} else if (
-								["complete", "error"].includes(status.stage) &&
-								pending_outputs.includes(id) &&
-								!outputs_set_to_non_interactive.includes(id)
-							) {
-								instance_map[id].props.interactive = true;
-							}
-						});
+
 						//@ts-ignore
 						loading_status.update({
 							...status,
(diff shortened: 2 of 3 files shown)


pngwn:
I've tweaked a test here to get rid of some flake. The commented our part does not test anything (it will pass regardless) and causes the flake.

pngwn:
Actually i know how to fix that test. Will amend.

demo/chatinterface_streaming_echo/run.py
abidlabs, line 1: Is there a reason you removed the time.sleep? Without it, its not as clear that this is a streaming demo

pngwn: Clear in what way? It returns an iterator.

js/app/src/Blocks.svelte
```diff
@@ -525,22 +525,7 @@
 				.on("status", ({ fn_index, ...status }) => {
 					tick().then(() => {
 						const outputs = dependencies[fn_index].outputs;
-						outputs.forEach((id) => {
```
abidlabs, line 528: We can also delete the line above (508) which defines `const pending_outputs: number[] = [];`

abidlabs: We also have some additional logic involving `outputs_set_to_non_interactive` that can be deleted

demo/chatinterface_streaming_echo/run.py
abidlabs: Meaning visually when you run the demo, the text appears all at once, instead of the characters appearing one by one

abidlabs: This demo is used in the docs here: https://github.com/gradio-app/gradio/blob/7bb561a294ca41d1044927cb34d8645c4175cae0/guides/04_chatbots/01_creating-a-chatbot-fast.md#L11

pngwn: Ah, we should really have dedicated demo's for tests. WIll fix.

pngwn: I'll create a new demo because i need to tweak the demo to fix the tests.

abidlabs: Maybe we can use the `_test` suffix for designating test demos

js/app/src/Blocks.svelte
pngwn: Ah yeh.

demo/chatinterface_streaming_echo/run.py
pngwn: Maybe a prefix, so they are easier to find in the fs?

pngwn:
Okay I cleaned this up and got the tests passing (i think, CI will tell).

pngwn:
was wrong

js/app/src/Blocks.svelte
```diff
@@ -367,11 +363,6 @@
 				for (const [update_key, update_value] of Object.entries(value)) {
 					if (update_key === "__type__") {
 						continue;
-					} else {
-						output.props[update_key] = update_value;
```
pngwn, line 371: oops

abidlabs (approved):
LGTM nice @pngwn!

Issue #7391, opened 2024-02-12
With gradio==4.18.0, only get loading screen for server for remote connections going through nginx


pseudotensor:
### Describe the bug

With gradio==4.18.0, only get loading screen for server for remote connections

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

No repro yet, but it works locally, but not when using nginx on remote server.

It happens each time I switch back and forth.  So something new broke nginx/proxy use in gradio 4.18.0.

For gradio 4.18.0, using non-ngnix name, just raw ip/port, works fine.  So definitely proxy issue that wasn't there in 4.17.0.

### Screenshot

![image](https://github.com/gradio-app/gradio/assets/2249614/7f88162b-7c90-4fd4-ad96-a98906113a31)

### System Info

```shell
gradio==4.18.0 gradio_client=0.10.0


For exact same code, just switching back to gradio==4.17.0 gradio_client=0.9.0 and no issues.
```

### Severity

Blocking usage of gradio

pseudotensor:
Plausible breaking changes: https://github.com/gradio-app/gradio/pull/7374  

FYI @abidlabs

pseudotensor:
Related of course, but not same issue: https://github.com/gradio-app/gradio/issues/6920

Would be good if both related issues were fixed.  I'm back to gradio 3.50.2 again because of this particular regression.

abidlabs:
Hmm yes it would be great if you could share a repro with the nginx config as well please @pseudotensor

pseudotensor:
Dev console in bad 4.18.0 case:

```
index.js:1181 Mixed Content: The page at 'https://xxxx.ai/' was loaded over HTTPS, but requested an insecure resource 'http://xxxx.ai/info'. This request has been blocked; the content must be served over HTTPS.
view_api @ index.js:1181
config_success @ index.js:441
(anonymous) @ index.js:481
await in (anonymous) (async)
client2 @ index.js:403
(anonymous) @ Index.svelte:254
run @ svelte.js:41
(anonymous) @ svelte.js:3149
flush @ svelte.js:2108
init @ svelte.js:3266
Index @ Index.svelte:357
connectedCallback @ main.ts:109
await in connectedCallback (async)
create_custom_element @ main.ts:211
(anonymous) @ main.ts:214
index.js:443 Could not get api details: Failed to fetch
config_success @ index.js:443
await in config_success (async)
(anonymous) @ index.js:481
[...]

Issue #7317, opened 2024-02-06
Gradio static files do not follow the root_path behind a Jupyter proxy on k8s


louisbrulenaudet:
### Describe the bug

Hi everyone,

I am experiencing a problem when running Gradio 4.16.0 behind a Jupyter proxy on a k8s cluster specifying a root_path, static files do not follow this path and my browser console returns a 404 error.

Below is the code:
```python
demo.launch(
    server_port=5300, 
    root_path="/proxy/absolute/5300/", 
    share=True, 
    inline=False
)
```
The link is still accessible and the components can be viewed without style.

Thank you in advance, I remain at your disposal for any clarification.

Best regards,
Louis

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

def greet(name):
    return "Hello " + name + "!"

demo = gr.Interface(
    fn=greet, 
    inputs="text", 
    outputs="text"
)

demo.launch(
    server_port=5300, 
    root_path="/proxy/absolute/5300/", 
    share=True, 
    inline=False
)
```

### System Info

```shell
gradio==4.16.0 on Python 3.9
```

### Severity

Blocking usage of gradio

abidlabs:
@louisbrulenaudet could you please try installing this version of `gradio` and seeing if the issue is resolved?

```
pip install https://gradio-builds.s3.amazonaws.com/9b8810ff9af4d9a50032752af09cefcf2ef7a7ac/gradio-4.18.0-py3-none-any.whl
```

pseudotensor:
@abidlabs Does it make sense that 3.50.2 had no k8 issue but 4.17.0, while only 4.18.0 had the reverse proxy issue?  I'm worried that the PR solved the recent regression from 4.17.0->4.18.0, but not this one.

abidlabs:
To be honest I'm not totally sure. I have not tested on a k8 cluster. However, I did fix an issue related to `root_path` so I *think* this issue is fixed. Would be awesome if someone can test on a k8 cluster

louisbrulenaudet:
Hi @abidlabs, @pseudotensor,

First of all, I'd like to thank you for your help. I've just tried version 4.19.0 released today and the problem persists. 

![application](https://github.com/gradio-app/gradio/assets/35007448/50e95b64-0f36-4e4e-8bd7-46790355145e)

Looking through the element inspector, it turns out that the index.css and theme.css files don't have the right path. These are respectively :
[...]

Issue #6920, opened 2024-01-02
404 Session Not Found error When accessing gradio via a proxy


lykeven:
### Describe the bug

I am running a Gradio application locally, where there's a requests request to a remote server in the click event function of a button, and the result is returned to the component. 
Everything works fine, but if I turn on a proxy (Shadowsocks) to access the Gradio application, requests with short response time return normally, while requests that take longer return exceptions.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
#!/usr/bin/env python

import gradio as gr
import os
import json
import requests
import base64

URL = os.environ.get("URL")

def image_to_base64(image_path):
    with open(image_path, "rb") as image_file:
        encoded_string = base64.b64encode(image_file.read())
        return encoded_string.decode('utf-8')

def post(
        input_text,
        image_prompt,
        ):
    headers = {
        "Content-Type": "application/json; charset=UTF-8",
        "User-Agent": "Mozilla/5.0 (Windows NT 6.1; WOW64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/67.0.3396.87 Safari/537.36",
    }
    if image_prompt:
        encoded_img = image_to_base64(image_prompt)
    else:
        return "", []

    data = json.dumps({
        'text': input_text,
        'history': [],
        'image': encoded_img
    })
    try:
        response = requests.request("POST", URL, headers=headers, data=data, timeout=(60, 100)).json()
    except Exception as e:
        return "", []
    answer = str(response['result'])
    return "", [[input_text, answer]]


def main():
    gr.close_all()
    with gr.Blocks() as demo:
        with gr.Row():
            with gr.Column(scale=4.5):
                with gr.Group():
                    input_text = gr.Textbox(label='Input Text', placeholder='Please enter text prompt below and press ENTER.')
                    with gr.Row():
                        run_button = gr.Button('Generate')
                    image_prompt = gr.Image(type="filepath", label="Image Prompt", value=None)

            with gr.Column(scale=5.5):
                result_text = gr.components.Chatbot(label='Multi-round conversation History', value=[("", "Hi, What do you want to know about this image?")], height=550)
                
        run_button.click(fn=post,inputs=[input_text, image_prompt],
                         outputs=[input_text, result_text])
    demo.launch(server_port=7862)

if __name__ == '__main__':
    main()

```

### Screenshot

[...]

PR #7411, opened 2024-02-13
Set `root` correctly for Gradio apps that are deployed behind reverse proxies 

abidlabs:
This PR fixes the issues that we're seeing around the `root` url when Gradio apps are deployed behind nginx or other reverse proxies. As far as I can tell, there were two separate issues:

1. The `root_path` parameter in `Blocks` was being ignored. This parameter is needed to tell Gradio (FastAPI really) the subpath where the Gradio app is being served from. E.g. if you are serving Gradio on `www.mydomain.com/demo`, it should be `/demo`. Now, its correctly accounted for. While its hard to write an end-to-end test for this, I have added unit tests for `get_root_url()` and I have tested it by running an nginx reverse proxy on an ec2 instance here: http://35.85.61.147/gradio-demo/

**Main**:

image

**This PR**:

image

2. When serving behind an nginx proxy, the "x-forwarded-proto" is supposed to designate whether the original website is being served on https or on http. We use this to set http/https in the `root` in the config. However, sometimes this header is not set correctly, or it fails to contain the protocol information. This causes a problem on websites that are using https because we try to fetch certain gradio static assets on http, which browsers automatically block. I added another check in the frontend to check to see if current window is being served with https and if so, upgrade the config.root to be https as well

Closes: https://github.com/gradio-app/gradio/issues/7391 (as confirmed by @pseudotensor)
Closes: https://github.com/gradio-app/gradio/issues/7317 

(May close?) https://github.com/gradio-app/gradio/issues/6920


js/app/src/Index.svelte
```diff
@@ -255,6 +255,9 @@
 			status_callback: handle_status
 		});
 		config = app.config;
+		if (window.location.protocol === "https:") {
```
freddyaboulton, line 258: How come we need to do this check if the client already does it?

test/test_routes.py
```diff
@@ -862,3 +862,50 @@ def test_component_server_endpoints(connect):
             },
         )
         assert fail_req.status_code == 404
+
+
+@pytest.mark.parametrize(
+    "request_url, route_path, root_path, expected_root_url",
+    [
+        ("http://localhost:7860/", "/", None, "http://localhost:7860"),
```
freddyaboulton, line 870: I think adding a a test where route_path ends in `/` could be useful. I know we have the `/` test here though.

test/test_routes.py
```diff
@@ -862,3 +862,50 @@ def test_component_server_endpoints(connect):
             },
         )
         assert fail_req.status_code == 404
+
+
+@pytest.mark.parametrize(
```
freddyaboulton, line 867: 👍

freddyaboulton (approved):
Nice fix @abidlabs !

js/app/src/Index.svelte
abidlabs: Actually we might not need this one. I added it here first and then to the client, I'll check with this removed as well.

abidlabs:
Thanks @freddyaboulton for the review! I'll clean this up tomorrow and merge it in.

cc @pngwn and @aliabid94 as well if y'all wanna take a look at this before that

js/app/src/Index.svelte
abidlabs: Tested and it seems to work, thanks @freddyaboulton!

Issue #7412, opened 2024-02-14
Updating to Gradio 4.18.0 causing TypeError: cannot pickle 'dict_keys' object


awssamdwar:
### Describe the bug

It seems Gadrio 4.18.0 is causing an issue where I get the below error, same code works with v4.17.0. 

As I work around I had to pin gradio version to 4.17.0. It seems the issue is related to adding deepcopy in routers.py (see [here](https://github.com/gradio-app/gradio/commit/7f19ba272c5a46d7f2c11ad5c9f0ab15d4e27083))

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

Full code can be found [here](https://github.com/aws-samples/codenator-automatic-code-generation-and-execution-using-llm/blob/main/src/codenator/web_ui/app/main.py)

### Logs

```shell
2024-02-14T00:21:39.815Z ERROR:    Exception in ASGI application
2024-02-14T00:21:39.815Z Traceback (most recent call last):
2024-02-14T00:21:39.815Z   File "/usr/local/lib/python3.10/site-packages/uvicorn/protocols/http/h11_impl.py", line 412, in run_asgi
2024-02-14T00:21:39.815Z     result = await app(  # type: ignore[func-returns-value]
2024-02-14T00:21:39.815Z   File "/usr/local/lib/python3.10/site-packages/uvicorn/middleware/proxy_headers.py", line 84, in __call__
2024-02-14T00:21:39.815Z     return await self.app(scope, receive, send)
2024-02-14T00:21:39.815Z   File "/usr/local/lib/python3.10/site-packages/fastapi/applications.py", line 1054, in __call__
2024-02-14T00:21:39.815Z     await super().__call__(scope, receive, send)
2024-02-14T00:21:39.815Z   File "/usr/local/lib/python3.10/site-packages/starlette/applications.py", line 123, in __call__
2024-02-14T00:21:39.815Z     await self.middleware_stack(scope, receive, send)
2024-02-14T00:21:39.815Z   File "/usr/local/lib/python3.10/site-packages/starlette/middleware/errors.py", line 186, in __call__
2024-02-14T00:21:39.815Z     raise exc
2024-02-14T00:21:39.815Z   File "/usr/local/lib/python3.10/site-packages/starlette/middleware/errors.py", line 164, in __call__
2024-02-14T00:21:39.815Z     await self.app(scope, receive, _send)
[...]

PR #7415, opened 2024-02-14
Allow config to include non-pickle-able values

abidlabs:
Closes: https://github.com/gradio-app/gradio/issues/7412

As part of the [recent changes](https://github.com/gradio-app/gradio/pull/7374) around `root`, I added a step which deep-copied the config. This prevented the config from including any objects that could not be pickled. I've refactored the code so that we do not need to deepcopy the config any more. As an added bonus, this should mostly eliminate the performance issues caused by https://github.com/gradio-app/gradio/pull/7374

Simple repro:

```py
import gradio as gr

my_dict = {'a': 1, 'b': 2, 'c': 3}

with gr.Blocks() as demo:
    j = gr.JSON(my_dict.keys())
    
demo.launch()
```


gradio/routes.py
```diff
@@ -311,17 +310,19 @@ def login(form_data: OAuth2PasswordRequestForm = Depends()):
...
             blocks = app.get_blocks()
-            root_path = route_utils.get_root_url(request)
+            root = route_utils.get_root_url(request)
             if app.auth is None or user is not None:
-                config = copy.deepcopy(app.get_blocks().config)
-                config["root"] = root_path
-                config = add_root_url(config, root_path)
+                config = app.get_blocks().config
+                previous_root = config.get("root", None)
+                if previous_root is None or previous_root != root:
```
abidlabs, line 317: The main change in this PR being that we no longer need to traverse the config if the `root` has not change

gradio/routes.py
```diff
@@ -352,11 +353,13 @@ def api_info():
...
-            config["root"] = root_path
-            config = add_root_url(config, root_path)
-            return config
+            config = app.get_blocks().config
+            root = route_utils.get_root_url(request)[: -len("/config")]
+            previous_root = config.get("root", None)
+            if previous_root is None or previous_root != root:
+                config["root"] = root
+                config = add_root_url(config, root, previous_root)
+            return ORJSONResponse(content=config)
```
abidlabs, line 362: Another change is that explicitly serialize the config using `ORJSONResponse` here. While `ORJSONResponse` is already the default serializer, if you return a dictionary-like from a fastapi route, it first tries to copy/cast it to a dictionary, which causes issues if the dictionary has non-pickle-able values.

So this was previously a bug (if a user tried to access the /config route and the config contained non-pickle-values, it would error out)

gradio/processing_utils.py
```diff
@@ -289,9 +289,11 @@ def _move_to_cache(d: dict):
     return client_utils.traverse(data, _move_to_cache, client_utils.is_file_obj)
 
 
-def add_root_url(data, root_url) -> dict:
+def add_root_url(data: dict, root_url: str, previous_root_url: str | None) -> dict:
     def _add_root_url(file_dict: dict):
         if not client_utils.is_http_url_like(file_dict["url"]):
+            if previous_root_url and file_dict["url"].startswith(previous_root_url):
```
freddyaboulton, line 295: Why do we need this extra step here? If we're running this we know that the root changed so we can just replace the root key entirely. I think we can get rid of the previous_root_url altogether.

abidlabs: Yes but this allows us to robustly identify what part of the `file_dict["url"]` is the previous root and strip it away. Otherwise, we'd have to look for the presence of a prefix before `/file=`, which makes me a bit nervous in case we change that prefix or somehow we have an unusual path

gradio/routes.py
```diff
@@ -311,17 +310,19 @@ def login(form_data: OAuth2PasswordRequestForm = Depends()):
         def main(request: fastapi.Request, user: str = Depends(get_current_user)):
             mimetypes.add_type("application/javascript", ".js")
             blocks = app.get_blocks()
-            root_path = route_utils.get_root_url(request)
+            root = route_utils.get_root_url(request)
             if app.auth is None or user is not None:
-                config = copy.deepcopy(app.get_blocks().config)
-                config["root"] = root_path
-                config = add_root_url(config, root_path)
+                config = app.get_blocks().config
```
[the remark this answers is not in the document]
abidlabs: Can do

gradio/processing_utils.py
freddyaboulton: Gotcha makes sense!

abidlabs:
Thanks @freddyaboulton for the review! I've made the suggested changes and am putting this on auto-merge

PR #7416, opened 2024-02-14
WIP: Optimize /file route

freddyaboulton:
## Description

Current `/file` route will scan over all the temp file sets to check if the file is allowed. This can cause latency in demos that create a lot of files across components, e.g. stable cascade ([context](https://huggingface.slack.com/archives/C02QZLG8GMN/p1707922456777229)). Solution here is to finish the scan early as soon as we find a match.

## 🎯 PRs Should Target Issues

## Tests


freddyaboulton:
Looks like this is helping the stablecascade demo so going to open for review.

abidlabs:
Thanks @freddyaboulton for the fix. Makes sense, with tens of thousands of files, the union operator is linear in the number of files, while this solution should be linear in the number of blocks (which is fixed and small), while the look ups themselves are constant time.

freddyaboulton:

PR #7417, opened 2024-02-14
Fix File Explorer interactivity

### LINKED ISSUE #6627: The `interactive` parameter in `gr.FileExplorer` has no effect. It should either be correctly documented or the feature should be added

aliabid94:
Fixes #6627 

When FileExplorer is not interactive, checkboxes are hidden

### DIFF
--- a/gradio/components/file_explorer.py
+++ b/gradio/components/file_explorer.py
@@ -66,7 +66,7 @@ class FileExplorer(Component):
             scale: relative size compared to adjacent Components. For example if Components A and B are in a Row, and A has scale=2, and B has scale=1, A will be twice as wide as B. Should be an integer. scale applies in Rows, and to top-level Components in Blocks where fill_height=True.
             min_width: minimum pixel width, will wrap if not sufficient screen space to satisfy this value. If a certain scale value results in this Component being narrower than min_width, the min_width parameter will be respected first.
             height: The maximum height of the file component, specified in pixels if a number is passed, or in CSS units if a string is passed. If more files are uploaded than can fit in the height, a scrollbar will appear.
-            interactive: if True, will allow users to upload a file; if False, can only be used to display files. If not provided, this is inferred based on whether the component is used as an input or output.
+            interactive: if True, will allow users to select a file; if False, will only display files. If not provided, this is inferred based on whether the component is used as an input or output.
             visible: If False, component will be hidden.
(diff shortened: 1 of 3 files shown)


gradio/components/file_explorer.py
```diff
@@ -66,7 +66,7 @@ def __init__(
             scale: relative size compared to adjacent Components. For example if Components A and B are in a Row, and A has scale=2, and B has scale=1, A will be twice as wide as B. Should be an integer. scale applies in Rows, and to top-level Components in Blocks where fill_height=True.
             min_width: minimum pixel width, will wrap if not sufficient screen space to satisfy this value. If a certain scale value results in this Component being narrower than min_width, the min_width parameter will be respected first.
             height: The maximum height of the file component, specified in pixels if a number is passed, or in CSS units if a string is passed. If more files are uploaded than can fit in the height, a scrollbar will appear.
-            interactive: if True, will allow users to upload a file; if False, can only be used to display files. If not provided, this is inferred based on whether the component is used as an input or output.
+            interactive: if True, will allow users to select a file; if False, will only display files. If not provided, this is inferred based on whether the component is used as an input or output.
```
abidlabs, line 69:
```suggestion
            interactive: if True, will allow users to select file(s); if False, will only display files. If not provided, this is inferred based on whether the component is used as an input or output.
```

abidlabs:
Thanks @aliabid94! One thing to think about -- usually, the non-interactive version of a component is used to display the `value` of a component. Here, the `value` are the selected files, however, these are not displayed at all when `interactive=False`. Would it make sense to just show the selected files, but not let users modify them?

aliabid94:
Hmm I see your point. Will revert. Visually I still think it's hard to tell a checkbox is disabled.

abidlabs (approved):
Ok cool looks good then.

Issue #6976, opened 2024-01-09
Chat Input Component


dawoodkhan82:
- [X] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  

Proposal UI:
![Group 4](https://github.com/gradio-app/gradio/assets/12725292/af054efd-bc8b-4631-b420-060c6d7fc599)

**Describe the solution you'd like**  

**Additional context**

dawoodkhan82:
@abidlabs Thoughts?

PR #7420, opened 2024-02-14
 Multimodal Textbox (Chat Input Component)

dawoodkhan82:
## Description

`MultimodalTextbox` in `ChatInterface`:

https://github.com/gradio-app/gradio/assets/12725292/e9afec7d-0617-4eec-beaf-43f87e3be720

TODO:

- [x] add submit button
- [x] ability to remove thumbnails
- [x] match API to what chatbot APIs expect
- [x] cleanup code

New multimodal textbox component.

Closes: #6976

## 🎯 PRs Should Target Issues

## Tests


demo/hello_blocks/run.py
```diff
@@ -4,8 +4,8 @@ def greet(name):
     return "Hello " + name + "!"
 
 with gr.Blocks() as demo:
-    name = gr.Textbox(label="Name")
-    output = gr.Textbox(label="Output Box")
+    name = gr.MultimodalTextbox(label="Name")
```
dawoodkhan82, line 7: Will change this back. Was just testing.

aliabid94:
Screenshot 2024-02-27 at 1 03 59 PM

My unsolicited design thoughts:
1. The submit arrow does not appear centered within the circle button
2. the X to delete an image seems to large within the circle
3. Seems strange that the images are flush against the left border but the text is not, due to the upload button being in the way. Can we move the upload button somewhere

dawoodkhan82:
> My unsolicited design thoughts:
> 
> 1. The submit arrow does not appear centered within the circle button
> 2. the X to delete an image seems to large within the circle
> 3. Seems strange that the images are flush against the left border but the text is not, due to the upload button being in the way. Can we move the upload button somewhere

Will fix 1 & 2.

3. I could possibly move the upload and submit buttons under the `textarea`. And extend the `textarea` to the full width.

gradio/components/multimodal_textbox.py
```diff
@@ -0,0 +1,139 @@
...
+        Events.input,
+        Events.select,
+        Events.submit,
+        Events.focus,
+        Events.blur,
+    ]
+
+    def __init__(
+        self,
+        value: list[dict[str, str]] | Callable | None = None,
```
aliabid94, line 33: I think this would be a cleaner type for this object:
```
{
  "text": string,
  "files": list[string]
} 
```
either that, or a tuple:
```
Tuple[str, list[str]]
```

have a slight preference for the latter.

dawoodkhan82: That is definitely a simpler data structure, but I wanted it to be more aligned with chat + vision APIs like OpenAI's (https://platform.openai.com/docs/guides/vision). I'm not super tied to the current data structure tho, so open to changing it.

abidlabs:
Nice @dawoodkhan82 this is a very cool component! Sharing some initial UI/UX feedback first:

1. There is a shadow that appears around the inner textbox which looks a little weird to me. Also the inner textbox has square corners whereas everything else in Gradio has rounded corner (including the `gr.Textbox`).

image

This is more visible on light mode. Dark mode looks pretty good imo, though you can still see some difference in shading if you zoom in:

image


2. As you can see in the first screenshot, at narrower screen widths, the "up arrow" button comes inside the textbox. Here's another example where the arrow overlays the text:

image

3. In light mode, the buttons are blue, which seems off to me, since we never use blue anywhere else. The plus button should be gray (the "secondary" color), and the submit button should be either the secondary color or the primary color. I actually think it should be the primary color to match the Submit button in ChatInterface when multimodal=False. 

4. Speaking of the buttons, I'm not a huge fan of the up arrow. How about instead using an airplane icon like in Slack:

image

This is all for now, I'll look into the backend next

demo/chatbot_multimodal/run.py
abidlabs, line 1: We should update `guides/04_chatbots/02_creating-a-custom-chatbot-with-blocks.md`, which references this demo as the surrounding text is no longer accurate

gradio/chat_interface.py
abidlabs, line 1: It'd be great to add a `chatinterface_multimodal` showing how to use `multimodal=True` as well as to add a section in this Guide: guides/04_chatbots/01_creating-a-chatbot-fast.md

gradio/chat_interface.py
```diff
@@ -81,8 +83,9 @@ def __init__(
...
+            textbox: An instance of the gr.Textbox or gr.MultimodalTextbox component to use for the chat interface, if you would like to customize the textbox properties. If not provided, a default gr.Textbox or gr.MultimodalTextbox component will be created.
             additional_inputs: An instance or list of instances of gradio components (or their string shortcuts) to use as additional inputs to the chatbot. If components are not already rendered in a surrounding Blocks, then the components will be displayed under the chatbot, in an accordion.
             additional_inputs_accordion_name: Deprecated. Will be removed in a future version of Gradio. Use the `additional_inputs_accordion` parameter instead.
             additional_inputs_accordion: If a string is provided, this is the label of the `gr.Accordion` to use to contain additional inputs. A `gr.Accordion` object can be provided as well to configure other properties of the container holding the additional inputs. Defaults to a `gr.Accordion(label="Additional Inputs", open=False)`. This parameter is only used if `additional_inputs` is provided.
```
abidlabs, line 91: We should clarify that the `submit_btn` parameter below is not used if `multimodal=True`

gradio/components/multimodal_textbox.py
```diff
@@ -0,0 +1,139 @@
...
+        interactive: bool | None = None,
+        visible: bool = True,
+        elem_id: str | None = None,
+        autofocus: bool = False,
+        autoscroll: bool = True,
+        elem_classes: list[str] | str | None = None,
+        render: bool = True,
+        text_align: Literal["left", "right"] | None = None,
+        rtl: bool = False,
+        show_copy_button: bool = False,
```
abidlabs, line 55: I think we should add a parameter for whether or not to show the submit button. I can imagine some non-chatbot use cases where it doesn't make sense

gradio/components/multimodal_textbox.py
abidlabs: I looked into this and personally what I would prefer is this format:

```py
{"text": ..., "files": [..., ...]}
```

For several reasons. 

(1) This is very similar to the `ImageEditor` format so we have a precedent.
(2) I originally thought it would make sense to align with `openai`'s format, but having looked at [their format](https://platform.openai.com/docs/guides/vision), they have specific keys like "image_url" etc. that we wouldn't want to bake into our component. So if a developer has to parse the data anyways, we should make it an easy format for them to parse
(3) Explicit keys are easier for developers to remember and for users reading their code than implicit tuple positions

E.g. its would be clear to a user of the code what this does:

import gradio as gr

def echo(x,y):
    x["text"]

gr.ChatInterface(echo, multimodal=True).launch()

abidlabs:
@dawoodkhan82 I was testing the `ChatInterface(multimodal=True)` with the following code:

```py
import gradio as gr

def test(x,y):
    print(x[0]['text'])
    return x[0]['text']

gr.ChatInterface(test, multimodal=True).launch()
```

and it is not working for me. When I run submit, I don't see any response. Am I doing something wrong?

gradio/components/multimodal_textbox.py
dawoodkhan82: Makes sense, thanks for looking into this. making this change now

pngwn: Yeah. Following the discussions and after some thought I think what @abidlabs said above mid probably the best approach.

js/multimodaltextbox/package.json
```diff
@@ -0,0 +1,25 @@
+{
+	"name": "@gradio/multimodaltextbox",
+	"version": "0.4.9",
```
pngwn, line 3: Lets set this to a normal initial value of `0.1.0`

js/multimodaltextbox/shared/MultimodalTextbox.svelte
```diff
@@ -0,0 +1,484 @@
...
+		_el.style.overflowY = "scroll";
+		_el.addEventListener("input", resize);
+
+		if (!_value.trim()) return;
+		resize({ target: _el });
+
+		return {
+			destroy: () => _el.removeEventListener("input", resize)
+		};
+	}
```
pngwn, line 213: Lets move the action and event handler into a different file to make this more readable.

We'll need to tweak it a little:

- The `_value` here is a string but this can be an object, allowing us to pass in whatever values we need here. So lets change that to accept `lines`, `max_lines` and we will also need to pass those into the event handler. We can change the event handler to a normal function that take in an element, lines and max_lines and adjust the usage:

```js
resize(_el, lines, max_lines);

// and 

_el.removeEventListener("input", (e: Event) => resize(e.target, lines, max_lines))
```

js/multimodaltextbox/shared/MultimodalTextbox.svelte
```diff
@@ -0,0 +1,484 @@
...
+		}
+	}
+
+	function copy_feedback(): void {
+		copied = true;
+		if (timer) clearTimeout(timer);
+		timer = setTimeout(() => {
+			copied = false;
+		}, 1000);
+	}
```
pngwn, line 117: We should move this as well but it will need refactoring into a store:

```js
import { writable } from 'svelte/store';

export function make_copy_store() {
  const copied = writeable(false);
  
  async function copy(value): Promise<void> {
    if ("clipboard" in navigator) {
      const text = value.find((item) => item && "text" in item);
      if (text && "text" in text && typeof text["text"] === "string") {
        await navigator.clipboard.writeText(text["text"]);
          copy_feedback();
      }
    }
  }
  
  function copy_feedback(): void { 
    copied.set(true);
    if (timer) clearTimeout(timer);
      timer = setTimeout(() => {
        copied.set(false);
      }, 1000);
  }
  
  return {
    copied,
    copy
  }
}
```

Then we should just able to use it like:

```
import { make_copy_store } from './wherever';

const { copied, copy } = make_copy_store();

// to copy
copy(value)

// the live 'copied' state, we'll need to replace all instances of `copied` with `$copied`.
$copied

**Review by pngwn — COMMENTED** [review]
Few comments but looks great otherwise.


There is also an alignment issues with the text and the `+` and `->` buttons. I found that if we added a padding-top of 12 or 13 pixels to the textbox then it aligned nicely.

**Comment by abidlabs** [comment]
Thanks @dawoodkhan82, the `MultimodalTextbox` component is looking really nice! Just gave this PR another spin, and noting the small bugs that I noticed:

1) I think the `.change()` event is not configured properly. For example, if I run:

```py
import gradio as gr

with gr.Blocks() as demo:
    t1 = gr.MultimodalTextbox()
    t2 = gr.MultimodalTextbox()
    t1.change(lambda x:x, t1, t2)
    
demo.launch()
```

The t2 component gets stuck in an infinite loop

2) The `label` and `info` attributes seem to have no effect. The `gr.MultimodalTextbox` should support these parameters, like the `gr.Textbox` and all the other form-based components:

import gradio as gr

demo = gr.Interface(lambda x:x, gr.MultimodalTextbox(label="name", info="hi"), "multimodaltextbox")

demo.launch()

3) There are some other parameters which also have no effect, like `show_copy_button`. But I think we can just remove these attributes for now. I'd make a pass through all of the parameters to make sure there are no extraneous ones left and the docstrings are all correct for the ones that are included.

4) When `interactive=False`, the files still have an "x" that allows them to be removed. These should not show up. There's also the "+" icon and the submit airplane icon. I think its fine if these icons are there, but they should be gray and disabled. (Or they can be removed. I don't have a strong opinion).

image

5) `gr.Chatbot(multimodal=True)` also looks super nice! And the basic form worked for me as I would have expected:

```py
import gradio as gr

def test(x,y):
    return x['text']

gr.ChatInterface(test, multimodal=True).launch()
```

However, the streaming version does not work:

```py
import gradio as gr

def test(x,y):
    yield x['text']

gr.ChatInterface(test, multimodal=True).launch()
```

6) We are missing the standard demo (the `multimodaltextbox_component` demo) that shows up in the component docs for each demo. We should add this

I tested quite a bit and these are the only issues I found so far. Happy to give it another spin later @dawoodkhan82!

abidlabs:
Also we should definitely add a section on `gr.Chatbot(multimodal=True)` in https://www.gradio.app/guides/creating-a-chatbot-fast

dawoodkhan82:
@abidlabs Thanks for the thorough review. Will fix these asap!

abidlabs:
Thanks @dawoodkhan82 for addressing the feedback I had above! I did some more tests:

> 1) I think the .change() event is not configured properly

I don't see the infinite reloading issue that I saw previously. BUT it looks like the .change() event is triggered before the value has finished updating. So it doesn't include the final character. It also doesn't trigger when I type backspace. I'm attaching a video showing the behavior that I'm seeing:

https://github.com/gradio-app/gradio/assets/1778297/df04e1dc-a9e9-4b15-a818-932f4b9bd7e0


> 2) The label and info attributes seem to have no effect. The gr.MultimodalTextbox should support these parameters, like the gr.Textbox and all the other form-based components:

Fixed, looks great!

> 3) There are some other parameters which also have no effect, like show_copy_button. But I think we can just remove these attributes for now. I'd make a pass through all of the parameters to make sure there are no extraneous ones left and the docstrings are all correct for the ones that are included.

These still need to be removed

> 4) When interactive=False, the files still have an "x" that allows them to be removed

Fixed, looks great!

> 5) However, the streaming version does not work:

This is not completely fixed. For example, this does not work:

```py
import gradio as gr
import time

def test(x,y):
    t = x['text']
    for i in range(len(t)):
        time.sleep(0.5)
        yield t[:i+1]

gr.ChatInterface(test, multimodal=True).launch()
```

Compare with this version, which does work:

```py
import gradio as gr
import time

def test(x,y):
    t = x
    for i in range(len(t)):
        time.sleep(0.5)
        yield t[:i+1]

gr.ChatInterface(test, multimodal=False).launch()
```

> 6) We are missing the standard demo (the multimodaltextbox_component demo) that shows up in the component docs for each demo. We should add this

Still needs to be added.

> 7) Also we should definitely add a section on gr.Chatbot(multimodal=True) in https://www.gradio.app/guides/creating-a-chatbot-fast

Still needs to be added.

-----

Also noticed a few other things:

8) In light mode, the "X" next to the images when the component is interactive is blue. I would suggest making it the "secondary" color like it is in dark mode: 

image

image

9) It seems that when files are provided like this, the image previews don't show up, as you can see the in the second screenshot above:

```py
with gr.Blocks() as demo:
    gr.MultimodalTextbox({"text": "abc", "files": ["cheetah.jpg"]}, interactive=True)
    
demo.launch()
```

10) It seems that *both* the "text" and the "files" field are required, otherwise there is no error raised, but the frontend crashes. I would instead make them optional, so that you can do stuff like this easily:

```py
import gradio as gr

gr.Interface(lambda x:{"text":x}, "textbox", "multimodaltextbox").launch()
```

And if a user provides data in an invalid format (e.g. just a string instead of a dict), the backend should raise an error in `gr.MultimodalTextbox.postprocess()` explaining what the correct format should be)

abidlabs:
Oh @dawoodkhan82 one thing I didn't test was examples. It'd be great if the examples for `gr.MultimodalTextbox` and `gr.ChatInterface(multimodal=True)` included the image files visualized as well

Here's a current demo for a vLLM for reference which includes a nice set of examples: https://huggingface.co/spaces/deepseek-ai/DeepSeek-VL-7B

With `gr.MultimodalTextbox`, the examples could be even more flexible, e.g. including multiple images alongside a single prompt

guides/04_chatbots/01_creating-a-chatbot-fast.md
```diff
@@ -124,6 +124,28 @@ gr.ChatInterface(
 ).launch()
 ```
 
+## Add Multimodal Capability to your chatbot
+
+You may want to add multimodal capability to your chatbot. For example, you may wantusers to be able to easily upload images or files to your chatbot and ask questions about it. This is supported by the `gr.MultimodalTextbox()` component, which supports both text and file inputs. Here's an example of how to use it:
```
abidlabs, line 129:
```suggestion
You may want to add multimodal capability to your chatbot. For example, you may want users to be able to easily upload images or files to your chatbot and ask questions about it. ```

File: guides/04_chatbots/01_creating-a-chatbot-fast.md [hunk 9]
```diff
@@ -124,6 +124,28 @@ gr.ChatInterface(
...
+
+```python
+import gradio as gr
+
+chat_input = gr.MultimodalTextbox(file_types=["image"], placeholder="Enter message or upload file...")
+```
+
+`gr.ChatInterface` also supports multimodality, simply pass in the `multimodal` parameter as `True`:
```
abidlabs, line 137: I would remove this part, since this guide is about `ChatInterface`, not the individual components.

```suggestion
```python
import gradio as gr

chat_input = gr.MultimodalTextbox(file_types=["image"], placeholder="Enter message or upload file...")
```

`gr.ChatInterface` also supports multimodality, simply pass in the `multimodal` parameter as `True`:
```

File: guides/04_chatbots/01_creating-a-chatbot-fast.md [hunk 10]
```diff
@@ -124,6 +124,28 @@ gr.ChatInterface(
...
+
+```python
+import gradio as gr
```
abidlabs, line 140:
```suggestion
import gradio as gr
import time

def echo(message, history):
    t = x['text']
    for i in range(len(t)):
        time.sleep(0.5)
        yield t[:i+1]

File: guides/04_chatbots/01_creating-a-chatbot-fast.md [hunk 11]
```diff
@@ -124,6 +124,28 @@ gr.ChatInterface(
...
+
+demo = gr.ChatInterface(fn=echo, examples=["hello", "hola", "merhaba"], title="Echo Bot", multimodal=True)
+
+demo.launch()
+```
+
+The `MultimodalTextbox` returns a dictionary of the form `{"text": "user input", "file": FileData}`. The FileData object contains the file name, file type, and file data. You can access the file path using the `file.path` attribute of the FileData object.
```
abidlabs, line 147:
```suggestion
When `multimodal=True`, the first parameter of your function should accept a dictionary that looks like this: `{"text": "user input", "file": ["file_path1", "file_path2", ...]}`.
```

abidlabs:
Thanks so much @dawoodkhan82 for addressing all of the comments! I left a few nits on the guide but everything else looks good.

Just one last piece of technical feedback:

We should make sure that the preprocess and postprocess of the "multimodaltextbox" are compatible with each other. In other words, this should work:

```py
import gradio as gr

gr.Interface(lambda x:x, "multimodaltextbox", "multimodaltextbox").launch()
```

Right now, it errors out if you add any files. I think the reason is that the multimodaltextbox preprocesses files, leaving them as a list of file dictionaries, whereas the postprocess expects a list of file *paths*

We should make preprocess also convert the file dictionaries to file paths (that's what we do with all of the other components, and it makes it easier for users to work with these filepaths). I've also suggested changes on the guide to this effect.

gradio/components/multimodal_textbox.py
```diff
@@ -0,0 +1,196 @@
...
+    ) -> dict[str, str | list] | None:
+        """
+        Parameters:
+            payload: the text and list of file(s) entered in the multimodal textbox.
+        Returns:
+            Passes text value and list of file(s) as a {dict} into the function.
+        """
+        return None if payload is None else payload
+
+    def _download_files(self, value: str | list[str]) -> str | list[str]:
```
freddyaboulton, line 133: I don't think we need `_download_files` here. We added this to `File` to get the size but that's not used in the multimodal textbox.

https://github.com/gradio-app/gradio/pull/7620/files#r1515022822

gradio/components/multimodal_textbox.py
```diff
@@ -0,0 +1,196 @@
...
+
+from gradio import processing_utils
+from gradio.components.base import FormComponent
+from gradio.data_classes import FileData, GradioRootModel
+from gradio.events import Events
+
+
+class MultimodalData(GradioRootModel):
+    root: Dict[str, str | List[FileData]]
+
```
freddyaboulton, line 19: I think we can create a better type of the data expected in postprocess

```python
from typing import TypedDict

class MultimodalPostprocess(TypedDict):
    text: str
    files: List[FileData | str]
```

gradio/components/multimodal_textbox.py
```diff
@@ -0,0 +1,196 @@
...
+import gradio_client.utils as client_utils
+from gradio_client.documentation import document
+
+from gradio import processing_utils
+from gradio.components.base import FormComponent
+from gradio.data_classes import FileData, GradioRootModel
+from gradio.events import Events
+
+
+class MultimodalData(GradioRootModel):
```
freddyaboulton, line 17: I think it would be better to name the fields explicitly

```python
from pydantic import Field

class MultimodalData(GradioModel):
    text: Optional[str] = None
    files: Optional[List[FileData]] = Field(default_factory=list)
```

gradio/components/multimodal_textbox.py
```diff
@@ -0,0 +1,196 @@
...
+
+@document()
+class MultimodalTextbox(FormComponent):
+    """
+    Creates a textarea for users to enter string input or display string output and also allows for the uploading of multimedia files.
+
+    Demos: chatbot_multimodal
+    Guides: creating-a-chatbot
+    """
+
```
freddyaboulton, line 29: We should set `data_model = MultimodalData` here so that we can get rid of the `api_info` method.

Right now the return value is incorrect and the View API page does not show the right data
image

gradio/components/multimodal_textbox.py
```diff
@@ -0,0 +1,196 @@
...
+    def preprocess(
+        self, payload: dict[str, str | list] | None
+    ) -> dict[str, str | list] | None:
+        """
+        Parameters:
+            payload: the text and list of file(s) entered in the multimodal textbox.
+        Returns:
+            Passes text value and list of file(s) as a {dict} into the function.
+        """
+        return None if payload is None else payload
```
freddyaboulton, line 131: Rewrite this as follows if you follow the `data_model = ` comment above^

```
        return None if payload is None else payload.model_dump()
```

gradio/chat_interface.py
```diff
@@ -82,8 +84,9 @@ def __init__(
         """
         Parameters:
             fn: The function to wrap the chat interface around. Should accept two parameters: a string input message and list of two-element lists of the form [[user_message, bot_message], ...] representing the chat history, and return a string response. See the Chatbot documentation for more information on the chat history format.
+            multimodal: If True, the chat interface will use a gr.MultimodalTextbox component for the input, which allows for the uploading of multimedia files. If False, the chat interface will use a gr.Textbox component for the input.
```
freddyaboulton, line 87: I think the `Undo` and `Retry` do not work if `multimodal=True`. We're passing in a string instead of a dict. This is the error I got for Retry

```
  File "/Users/freddy/miniconda3/envs/gradio-310/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 807, in run
  File "/Users/freddy/sources/gradio/gradio/utils.py", line 493, in run_sync_iterator_async
    return next(iterator)
  File "/Users/freddy/sources/scratch/multimodal_textbox_demo.py", line 5, in test
    t = x['text']
TypeError: string indices must be integers
```

freddyaboulton (commented):
Nice work @dawoodkhan82 ! Love the look of the component. Left some comments on the python file for the component but otherwise looks good.

abidlabs:
One thought: instead of “show_submit_button: bool =True”, what if we made it “submit_btn: str | Literal[False] = [plane emoji]”. That way we could change the icon e.g. while a chatbot is streaming and if a user clicks on it, it could also stop the streaming. Setting this to False would hide the button as is currently the case.

We don’t have to implement the change in ChatInterface but we should finalize tge parameter in MultimodalTextbox so we don’t make a breaking change.

abidlabs (approved):
LGTM 🔥🔥🔥

abidlabs:
Its just flakes that are blocking this now. I'm going to override so that we can merge this in

nxphi47:
@dawoodkhan82 Many thanks for this great features. However, how to use this with `gr.Examples` ?

I need the specify and `gr.Examples` so that when clicking on an example, it will copy it the Multimodal textbox (both text and images)

Thanks!

dawoodkhan82:
You can do:

`gr.Examples([[{"text": "Hello, how are you?", "files": ["file_path"]}]], [chat_input])`

where "file_path" is a path to your image files, and `chat_input` is a reference to the multimodal textbox.

PR #7422, opened 2024-02-14
Fixes #4016 by using websockets with callback so user can manage freeing up of state objects

### LINKED ISSUE #4016: if user exits browser or tab, gradio not cleaning up process/threads
### Describe the bug

https://huggingface.co/docs/transformers/internal/generation_utils#transformers.TextIteratorStreamer

When using this, if user exits tab or closes browser, generation continues in background indefinitely.

### Reproduction

```
import gradio as gr

with gr.Blocks() as demo:
    chatbot = gr.Chatbot()
    msg = gr.Textbox()
    clear = gr.Button("Clear")

    from transformers import AutoModelForCausalLM, AutoTokenizer, TextIteratorStreamer
    tok = AutoTokenizer.from_pretrained("gpt2")
    model = AutoModelForCausalLM.from_pretrained("gpt2")
    streamer = TextIteratorStreamer(tok)


        from threading import Thread

        # Run the generation in a separate thread, so that we can fetch the generated text in a non-blocking way.
        inputs = tok([message], return_tensors="pt")
        generation_kwargs = dict(inputs, streamer=streamer, max_new_tokens=1000)
        thread = Thread(target=model.generate, kwargs=generation_kwargs)
        thread.start()
        bot_message = ""
        chat_history.append([message, bot_message])
        for new_text in streamer:
            bot_message += new_text
            chat_history[-1][1] = bot_message
            yield chat_history
        return

    msg.submit(respond, [msg, chatbot], chatbot, queue=True)

demo.queue(concurrency_count=1)
demo.launch()
```

Enter "The sky is" and soon after close the tab.  The generation will continue.  For larger models, this is problem since gradio thinks the user is gone, queue is open, but now threads will overlap.

Also, note that adding a `raise StopIteration()` has no effect on the model.generate(), it only terminates the respond generation.  So GPU usage continues in background.  i.e.

```
import gradio as gr

with gr.Blocks() as demo:
    chatbot = gr.Chatbot()
    msg = gr.Textbox()
    clear = gr.Button("Clear")

    from transformers import AutoModelForCausalLM, AutoTokenizer, TextIteratorStreamer
    tok = AutoTokenizer.from_pretrained("gpt2")
    model = AutoModelForCausalLM.from_pretrained("gpt2")
    streamer = TextIteratorStreamer(tok)


        from threading import Thread

        # Run the generation in a separate thread, so that we can fetch the generated text in a non-blocking way.
        inputs = tok([message], return_tensors="pt")
        generation_kwargs = dict(inputs, streamer=streamer, max_new_tokens=1000)
        thread = Thread(target=model.generate, kwargs=generation_kwargs)
        thread.start()
        bot_message = ""
        chat_history.append([message, bot_message])
        for new_text in streamer:
            bot_message += new_text
            chat_history[-1][1] = bot_message
            if len(bot_message) > 50:
                raise StopIteration()
            yield chat_history
        return

    msg.submit(respond, [msg, chatbot], chatbot, queue=True)

demo.queue(concurrency_count=1)
demo.launch()
```

This will lead to a termination of output to gradio chatbot, but continue to use GPU in other thread.  It doesn't help to join the thread, that just holds there until generation finishes.  Some methods for excepting a thread exist, but that doesn't solve first issue and is not standard to do.

### Screenshot

image

### Logs

```shell
None required.
```

### System Info

```shell
latest, chrome browser.
```

### Severity

blocking all usage of gradio

pseudotensor:
## Description

Fixes #4016 by using websockets with callback so user can manage freeing up of state objects.

Closes: #4016 

## Tests

I'd prefer to not add specific tests or other things until there is high likelihood of being merged if tests were added.

Example use cases we could test:
* Chroma database in state, callback removes the database and garbage collects to release the state
* Torch model state is in state on CUDA device, callback moves to cpu(), clears torch cache, then garbage collects to release GPU memory

### DIFF
--- a/client/js/src/client.ts
+++ b/client/js/src/client.ts
@@ -283,6 +283,30 @@ export function api_factory(
 				await process_endpoint(app_reference, hf_token);
 
 			const session_hash = Math.random().toString(36).substring(2);
+
+			// WebSocket setup using the session_hash to check for alive state of client
+			const wsUrl = `${ws_protocol}://${host}/ws?session_hash=${session_hash}`;
+			const ws = new WebSocket(wsUrl);
+
+			ws.onopen = () => {
+						console.log("WebSocket connected with session_hash:", session_hash);
+						// Optionally send a message or perform an action upon connection
+			};
+
+			ws.onmessage = (event) => {
+						console.log("Received message:", event.data);
+						// Handle incoming WebSocket messages
+			};
+
+			ws.onerror = (error) => {
+						console.error("WebSocket error:", error);
+			};
+
+			ws.onclose = () => {
+						console.log("WebSocket connection closed");
+						// Optionally perform cleanup or reconnection logic
+			};
+
 			const last_status: Record<string, Status["stage"]> = {};
 			let stream_open = false;
 			let pending_stream_messages: Record<string, any[]> = {}; // Event messages may be received by the SSE stream before the initial data POST request is complete. To resolve this race condition, we store the messages in a dictionary and process them when the POST request is complete.
--- a/gradio/components/state.py
+++ b/gradio/components/state.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 from copy import deepcopy
-from typing import Any
+from typing import Any, Callable
 
 from gradio_client.documentation import document
 
@@ -27,6 +27,7 @@ class State(Component):
         self,
         value: Any = None,
         render: bool = True,
+        callback: Callable = None,
     ):
         """
         Parameters:
@@ -40,6 +41,7 @@ class State(Component):
             raise TypeError(
                 f"The initial value of `gr.State` must be able to be deepcopied. The initial value of type {type(value)} cannot be deepcopied."
             ) from err
+        self.callback = callback
         super().__init__(value=self.value, render=render)
 
     def preprocess(self, payload: Any) -> Any:
--- a/gradio/routes.py
+++ b/gradio/routes.py
@@ -118,6 +118,8 @@ client = httpx.AsyncClient()
 
 file_upload_statuses = FileUploadProgress()
 
+from fastapi import FastAPI, WebSocket, Query
+from starlette.websockets import WebSocketDisconnect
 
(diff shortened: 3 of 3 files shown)


pseudotensor:
@abidlabs let me know what you think.  As I mentioned, probably needs some love, I'm not web dev.

abidlabs:
Thanks @pseudotensor for this PR! My main hesitation is that this solution requires the use of websockets, which we intentionally removed in the migration from Gradio 3.x -> Gradio 4.x because websockets are not supported on some cloud deployment platforms. I would be loathe to bring them back again. I suppose we could do a similar approach using http long polling instead? 

Will cc @pngwn @aliabid94 @freddyaboulton for their thoughts -- generally speaking, I do think we should come up with a way to free `gr.State` and other temporary files when a user has finished using a Gradio app

pseudotensor:
Maybe after your conversion way from websockets you have some ideas for how to do the equivalent.  But the code is very simple and general via the callback, so it's pretty useful.  I'll use it for my purposes.

freddyaboulton:
Thanks for the PR @pseudotensor ! I think we can accomplish the same with a new SSE stream that's open once per app when it first connects. We can detect when the incoming requests disconnects and clear the state then.

pngwn:
I'm not convinced this is reliable. SSE/ EventSource has retries / reconnections built in, that's one of the big advantages. Clearing up state based on that could easily break things. I think we need real sessions for this.

pngwn:
Thanks for making this PR!

We don't want to go with this particularl implementation as it brings another protocol into the codebase which we would rather avoid but we definitely want to solve this, possibly using a similar approach with SSE.

Thanks again @pseudotensor!

DarrenZhaoFR:
Hi, I was wondering if there have been any updates on this issue?

freddyaboulton:
Hi @DarrenZhaoFR - it has been implemented. gr.State objects are automatically deleted 60 minutes after the connection is closed (to allow for reconnection) and there is a `delete_callback` to allow for custom clean up actions.

PR #7425, opened 2024-02-15
Fixes to the `.key_up()` method to make it usable for a dynamic dropdown autocomplete

abidlabs:
This PR makes two changes:

1. In the Dropdown component, we were resetting the value when the choices were updated. It is somewhat debatable whether we should do this or not. For example, in https://github.com/gradio-app/gradio/issues/6853, users asked for this behavior. However, it makes it impossible to use for the use case in which the Dropdown component's choices are being updated based on the value of the Dropdown itself. 

The way I've implemented it here is that iff when `allow_custom_value=True`, the Dropdown's values are *not* reset, which I think is the desired behavior for users in either case. I've also added a test for this.

2. Since `.key_up()` is an event that is expected to be fired continuously, I've set the default `trigger_mode` for the `.key_up()` event to be `always_last`, just like for `.change()`.

Test the new behavior with this demo:

```py
import gradio as gr
import time

def f(x, k: gr.KeyUpData):
    time.sleep(1)
    return gr.update(choices=[f"{k.input_value} {i}" for i in range(int(5))])    


with gr.Blocks() as demo:
    t = gr.Dropdown(interactive=True, allow_custom_value=True)
    t.key_up(f, t, t, show_progress="hidden", trigger_mode="always_last")
    
demo.launch()
```

cc @apolinario (see [internal thread](https://huggingface.slack.com/archives/C02QZLG8GMN/p1707952018419409))


pngwn:
Isn't this a breaking change?

abidlabs:
Both of the changes are technically breaking changes but they were _both_ made in 4.19.0 (the first one was changed in 4.19.0 and the KeyUp event was introduced in 4.19.0) so I think its ok

pngwn:
I wouldnt list them as breaking changes if we aren't doing a major bump :p that's just confusing. I'd just put it through as a fix.

abidlabs:
The python test is failing due to a typecheck issue on files that I haven't touched, something strange is happening on CI. Will investigate but that should be independent of this PR

abidlabs:
Thanks @hannahblair for reviewing! Merging in changes from `main` to fix the typechecks. And then will merge this in.

PR #7430, opened 2024-02-15
build(deps): 📦 remove ruff as dependencies

onuralpszr:
## Description

Gradio technically for production not require to install ruff as dependencies. It is for linter,formatter etc. It should be install for development purpose but not as main dependency. It may need extra change for certain workflow but I want to start PR for discuss as well. 

Thank you.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -23,4 +23,4 @@ typing_extensions~=4.0
 uvicorn>=0.14.0
 typer[all]>=0.9,<1.0
 tomlkit==0.12.0
-ruff >= 0.1.7
+# ruff >= 0.1.7


pngwn:
We use ruff as part of our custom component tooling which is part of the distributed gradio library (via the CLI), so this dependency is a correct dependency of the library, not just a development dependency.

onuralpszr:
@pngwn thank you for explanation, but could you also show me that usage ?

pngwn:
Sure thing.

It is here: https://github.com/gradio-app/gradio/blob/main/gradio/cli/commands/components/_docs_utils.py#L18-L47

Which is used as part of the automatic docs generation for custom components: https://github.com/gradio-app/gradio/blob/main/gradio/cli/commands/components/docs.py

Issue #7429, opened 2024-02-15
Unable to initialize Dropdown without any preselected value in 4.19.0


yvrjsharma:
### Describe the bug

Working as expected for 4.18.0

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python

import gradio as gr
with gr.Blocks() as demo:
  # Attempt without specifying a default or value
  division = gr.Dropdown(choices=['Division 1', 'Division 2'], label="Division",)
  # Attempt with default=None
  gender = gr.Dropdown(choices=['Male', 'Female', 'Other'], label="Gender", value=None)

demo.launch()
```

### Screenshot

With 4.19.0:
![image](https://github.com/gradio-app/gradio/assets/48665385/b5e01c3f-0800-49c6-b97b-ebc054901a01)

With 4.18.0:
![image](https://github.com/gradio-app/gradio/assets/48665385/e2402306-8088-4797-8cbf-77381db5ba5f)

### System Info

```shell
Running on Google Colab
```

### Severity

I can work around it

PR #7431, opened 2024-02-15
Ensure `gr.Dropdown` can have an empty initial value

hannahblair:
## Description

In PR #7401 we changed the initial value of `value` from `[]` to `undefined`. This means that the `else` check in the following logic was being run because value is now treated as Falsy: 

```
if (value) {
        old_selected_index = choices.map((c) => c[1]).indexOf(value as string);
        selected_index = old_selected_index;
        if (selected_index === -1) {
	        old_value = value;
	        selected_index = null;
        } else {
	        [input_text, old_value] = choices[selected_index];
	        old_input_text = input_text;
        }
} else if (choices.length > 0) {
        old_selected_index = 0;
        selected_index = 0;
        [input_text, value] = choices[selected_index];
        old_value = value;
        old_input_text = input_text;
}
```

when the `else if (choices.length > 0)` expression passes,  we set the selected index to 0, and then set value to 0, resulting in the None value behaviour breaking. So this PR changes the default value back to `[]`.

Closes: #7429 

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/major-years-burn.md
+++ b/.changeset/major-years-burn.md
@@ -3,4 +3,4 @@
 "gradio": patch
 ---
 
-fix:Ensure `gr.Dropdown` can have empty value
+fix:Ensure `gr.Dropdown` can have an empty initial value


abidlabs:
Ah interesting. I hadn't realized that the default behavior before 4.19 was to have no option selected in the Dropdown. This is different than how the `<select>` element works in HTML, where the first element is automatically selected. I slightly prefer the behavior of automatically selecting the first element because that way, the value of the Dropdown is always valid and the user doesn't have to worry about handling the case where the value is None. That being said, since the behavior was previously different, we should not introduce this breaking change. 

Change looks good to me @hannahblair, thanks for adding the tests! I also confirmed that https://github.com/gradio-app/gradio/issues/6853 is still fixed with this change.

abidlabs (approved):
Thanks @hannahblair and @yvrjsharma!

PR #7440, opened 2024-02-15
Prevent timing attacks to guess Gradio passwords

abidlabs:
This way we compare passwords is vulnerable to timing attack to guess the password. This PR uses `hmac` to prevent that, as described here: https://sqreen.github.io/DevelopersSecurityBestPractices/timing-attack/python

### DIFF
--- a/gradio/route_utils.py
+++ b/gradio/route_utils.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import hashlib
+import hmac
 import json
 import shutil
 from collections import deque
@@ -574,3 +575,7 @@ def update_root_in_config(config: dict, root: str) -> dict:
         config["root"] = root
         config = processing_utils.add_root_url(config, root, previous_root)
     return config
+
+
+def compare_passwords_securely(input_password: str, correct_password: str) -> bool:
+    return hmac.compare_digest(input_password, correct_password)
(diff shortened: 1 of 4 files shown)


gradio/route_utils.py
```diff
@@ -574,3 +575,7 @@ def update_root_in_config(config: dict, root: str) -> dict:
         config["root"] = root
         config = processing_utils.add_root_url(config, root, previous_root)
     return config
+
+
+def compare_passwords_securely(input_password: str, correct_password: str) -> bool:
+    return hmac.compare_digest(input_password, correct_password)
```
freddyaboulton, line 581: I think this only works for ASCII characters. We should convert to bytes first something like `input_password.encode("utf-8")`

abidlabs: Ah nice catch

gradio/route_utils.py
```diff
@@ -569,8 +570,12 @@ def update_root_in_config(config: dict, root: str) -> dict:
...
-    previous_root = config.get("root", None)
+    previous_root = config.get("root")
     if previous_root is None or previous_root != root:
         config["root"] = root
         config = processing_utils.add_root_url(config, root, previous_root)
     return config
+
+
+def compare_passwords_securely(input_password: str, correct_password: str) -> bool:
+    return hmac.compare_digest(input_password.encode(), correct_password.encode())
```
abidlabs, line 581: Default encoding is `utf-8`

abidlabs:
Done @freddyaboulton if you could take another look!

freddyaboulton (approved):
Nice and quick fix @abidlabs !

Issue #7432, opened 2024-02-15
FileExplorer change event not working in 4.19.0


shimizust:
### Describe the bug

The FileExplorer change event seems to not be working in 4.19.0. Confirmed it's working as expected in 4.18.0.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr


def on_file_change(file_path):
    print(f"File path changed to {file_path}")


# Set up the Gradio interface
with gr.Blocks(
    title="File explorer value change test"
) as demo:
    file_explorer = gr.FileExplorer(root=".", file_count="single")
    file_explorer.change(fn=on_file_change, inputs=[file_explorer], outputs=[])


demo.launch()
```

In 4.19.0 file path is not printed out, unlike previous versions.

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Darwin
gradio version: 4.19.0
gradio_client version: 0.10.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.2.0
fastapi: 0.108.0
ffmpy: 0.3.1
gradio-client==0.10.0 is not installed.
httpx: 0.26.0
huggingface-hub: 0.20.1
importlib-resources: 6.1.1
jinja2: 3.1.2
markupsafe: 2.1.3
matplotlib: 3.8.2
numpy: 1.26.3
orjson: 3.9.10
packaging: 23.2
pandas: 2.1.4
pillow: 10.2.0
pydantic: 2.5.3
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0.1
ruff: 0.2.1
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.9.0
typing-extensions: 4.9.0
uvicorn: 0.25.0
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2023.12.2
httpx: 0.26.0
huggingface-hub: 0.20.1
packaging: 23.2
typing-extensions: 4.9.0
websockets: 11.0.3
```

### Severity

Blocking usage of gradio

shimizust:
cc @aliabid94 . It seems like this was introduced in https://github.com/gradio-app/gradio/pull/7337/files removing dispatch_change?

shimizust:
Thanks for the quick fix @aliabid94 !

PR #7441, opened 2024-02-15
Dispatch change event for file explorer

aliabid94:
Fixes: #7432


js/fileexplorer/Index.svelte
```diff
@@ -37,6 +38,11 @@
 	export let interactive: boolean;
 
 	$: rerender_key = [root_dir, glob, ignore_glob];
+
+	$: if (JSON.stringify(value) !== JSON.stringify(old_value)) {
```
freddyaboulton, line 42: Let's add a unit test for this - otherwise looks good to me @aliabid94 !

abidlabs (approved):
Agree with @freddyaboulton!

dawoodkhan82 (approved):

js/fileexplorer/Index.svelte
aliabid94: done

pngwn (approved):
Looks good. Just updated the branch to see if the tests pass, they are finae locally so not sure what the issue is. Thanks @aliabid94!

PR #7443, opened 2024-02-15
Update `httpx` to `httpx>=0.24.1` in `requirements.txt`

abidlabs:
We require `httpx>=0.24.1` but this was not explicitly pinned. With lower versions of `httpx`, you get this error:

```
AsyncConnectionPool.__init__() got an unexpected keyword argument 'socket_options'
```

which I just got, and apparently is a [common frustration using Gradio](https://www.google.com/search?q=AsyncConnectionPool.__init__%28%29+got+an+unexpected+keyword+argument+%27socket_options%27&sca_esv=1b487df0b959fb49&sxsrf=ACQVn09lWQMwkTn6CpFGLQH1SvLnz9ugxw%3A1708028583616&ei=p3LOZeSpIo-hur8Pv9256Aw&ved=0ahUKEwikmfWnlq6EAxWPkO4BHb9uDs0Q4dUDCBE&uact=5&oq=AsyncConnectionPool.__init__%28%29+got+an+unexpected+keyword+argument+%27socket_options%27&gs_lp=Egxnd3Mtd2l6LXNlcnAiUkFzeW5jQ29ubmVjdGlvblBvb2wuX19pbml0X18oKSBnb3QgYW4gdW5leHBlY3RlZCBrZXl3b3JkIGFyZ3VtZW50ICdzb2NrZXRfb3B0aW9ucydIAFAAWABwAHgBkAEAmAEAoAEAqgEAuAEDyAEA4gMEGAAgQQ&sclient=gws-wiz-serp)


requirements.txt
```diff
@@ -3,7 +3,7 @@ altair>=4.2.0,<6.0
 fastapi
 ffmpy
 gradio_client==0.10.0
-httpx
+httpx>=0.24.1
```
freddyaboulton, line 6: Let's do so for the python client as well?

abidlabs:
Another good catch. Done!

freddyaboulton (approved):
Nice! 👍

abidlabs:
Don't understand why I'm getting these typecheck errors (?) I busted the gh cache but didn't seem to help

abidlabs:
It seems like the CI is using the `gradio_client` from pypi instead of the one installed locally hmm

abidlabs:
Ok so I've identified the problem. The problem is that we now include `gradio_pdf` as a dependency in `test/requirements.txt`. The problem is that `gradio_pdf` itself lists `gradio` as a dependency. When `pip` looks on pypi, it seems that the latest version of `gradio` itself lists `gradio_client==0.0.9` as a dependency so it uninstalls the local version of `gradio_client` and instead uses `gradio_client` from pypi. 

Seeing if I can change the order and force reinstall `gradio_client`.

requirements.txt
abidlabs: This comment led me down a deep rabbit hole. But the upside is that I know our CI a lot better now. Thanks @freddyaboulton! I've addressed it so I'll go ahead and merge

PR #7444, opened 2024-02-15
Block `/file=` filepaths that could expose credentials on Windows

abidlabs:


test/test_routes.py
```diff
@@ -720,25 +726,6 @@ def test_orjson_serialization():
     demo.close()
 
 
-def test_file_route_does_not_allow_dot_paths(tmp_path):
```
abidlabs, line 723: We don't need a dedicated test suite for this since this is no longer a special case (we simply block all files in the working directory)

gradio/routes.py
```diff
@@ -428,12 +428,18 @@ async def file(path_or_url: str, request: fastapi.Request):
                 return RedirectResponse(
                     url=path_or_url, status_code=status.HTTP_302_FOUND
                 )
+
+            invalid_prefixes = ["//", "file://", "ftp://", "sftp://", "smb://"]
```
freddyaboulton, line 432: I believe checking whether or not `abs_path` exists will also fix the issue and it doesn't involve hardcoding this list of prefixes?

abidlabs: So I'm not sure totally sure but it seems like any file operation on Windows could leak credentials

freddyaboulton: Yikes

freddyaboulton (approved):
Looks good @abidlabs !

gradio/routes.py
pngwn: Do we care about ipfs urls here too?

abidlabs: Probably so I don't know that much about them, but let me rewrite them to prevent any protocols except for http/https

PR #7447, opened 2024-02-16
Add delete_cache parameter to gr.Blocks to delete files created by app on shutdown

### LINKED ISSUE #3950: after prediction, the input and output files remain in the temporary directory.

### LINKED ISSUE #5350: Examples consisting of Video components fill up /tmp
### Describe the bug

Hi,

I am attempting to create an interface that is supposed to show a video output component and thumbnails of available video files. The purpose is to allow the user to select and view a video from a list which is already existing on the machine that is running the application and then allow the user to do further processing of the selected clip.

I will provide a very simple example, stripped of any further logic in the "Reproduction" section. It will fail if `videos/*` contains more data than can fit into `/tmp/`.

What I am seeing is, that when the app is launched, all video files get copied to `/tmp/gradio/[somehash]/original_filename.mp4`.

The issues is that I have about 115 files which in total occupy 18GB, so not only does launching the application takes a significant amount of time due to the copying, at some point it crashes with an `OSError: [Errno 28] No space left on device:` exception, because `/tmp/` gets filled up.

This seems to be somewhat related to https://github.com/gradio-app/gradio/issues/2660 and I understand, that the copying is being done in order not to modify the original video. I wonder however, if this modification is really always necessary or if there could be a user parameter identifying the video as "read only"/"do not copy"?

Are there currently any workarounds which would prevent copying, while not modifying the original video? I figured that it would probably be possible to set `GRADIO_TEMP_DIR` in order to prevent the out of space issue, but the copying is still quite time consuming. In the end it did not work out - `GRADIO_TMP_DIR` got ignored (perhaps related to https://github.com/gradio-app/gradio/issues/4063). I had to set `TMP` to work around it.

I can also see, that after the exception `/tmp/gradio` is not being cleaned up and remains filled with copied videos.

### Reproduction

```
import gradio as gr
import glob

with gr.Blocks() as demo: 
    dirlist = glob.glob("videos/*.mp4") # must contain more than can fit into /tmp
    dirlist.sort()
    gr.Examples(examples = dirlist, inputs = gr.Video(interactive = False), cache_examples = False)

if __name__ == "__main__":
    demo.launch()
```

### Screenshot

_No response_

### Logs

_No response_

### System Info

```shell
$ pip list | grep gradio
gradio                        3.36.1
gradio_client                 0.2.7
```

Running on Fedora 37.
```


### Severity

I can work around it

### PR DESCRIPTION
**Pull request by freddyaboulton**
## Description

Partial fix for #5350. 
Closes #3950

Adds a `delete_cache` parameter which is a tuple of [frequency, age]. Every `frequency` seconds, files in the cache older than `age` seconds will be deleted.

## 🎯 PRs Should Target Issues

## Tests

### CODE REVIEW

**Comment by abidlabs** [comment]
Based on various internal and external questions about the disk space on Spaces filling up (e.g. [here](https://huggingface.slack.com/archives/C039P47V1L5/p1708128081330059)), I think we should provide a mechanism to delete files that does not require a server restart. E.g. all files older than a certain timestamp get deleted

**Comment by freddyaboulton** [comment]
Good point - I think we make `delete_cache: tuple[int, int]`. The first element is the frequency to run the deletion script. The second element is the number of seconds since the last modification that files should be deleted.

**Comment by freddyaboulton** [comment]
@abidlabs I think this is good for review. The tests are failing for an unrelated reason I believe.

File: .changeset/public-hoops-drum.md [hunk 1]
```diff
@@ -0,0 +1,5 @@
+---
+"gradio": patch
```
abidlabs, line 2: should be a `minor`

gradio/route_utils.py
```diff
@@ -583,3 +599,64 @@ def starts_with_protocol(string: str) -> bool:
...
+        to_remove = set()
+        for file in temp_set:
+            try:
+                file_path = Path(file)
+                modified_time = datetime.fromtimestamp(file_path.lstat().st_ctime)
+                print("modified_time", modified_time)
+                print("now", datetime.now())
+                print("age", (datetime.now() - modified_time).seconds)
+                if age is None or (datetime.now() - modified_time).seconds > age:
+                    print("Deleting file", file)
```
abidlabs, line 617: Remove print statements

gradio/route_utils.py
```diff
@@ -583,3 +599,64 @@ def starts_with_protocol(string: str) -> bool:
...
+                    os.remove(file)
+                    to_remove.add(file)
+            except FileNotFoundError:
+                continue
+        temp_set -= to_remove
```
abidlabs, line 622: Why are we storing the unremoved files in `temp_set` here? Are we intending on doing something with these files later?

gradio/route_utils.py
```diff
@@ -583,3 +599,64 @@ def starts_with_protocol(string: str) -> bool:
     """
     pattern = r"^[a-zA-Z][a-zA-Z0-9+\-.]*://"
     return re.match(pattern, string) is not None
+
+
+def delete_files_created_by_app(blocks: Blocks, age: int | None) -> None:
```
abidlabs, line 604: I believe this won't delete uploaded files (which can take up a lot of disk space in viral Spaces

gradio/blocks.py
```diff
@@ -516,6 +517,7 @@ def __init__(
             js: Custom js or path to js file to run when demo is first loaded. This javascript will be included in the demo webpage.
             head: Custom html to insert into the head of the demo webpage. This can be used to add custom meta tags, scripts, stylesheets, etc. to the page.
             fill_height: Whether to vertically expand top-level child components to the height of the window. If True, expansion occurs when the scale value of the child components >= 1.
+            delete_cache: A tuple corresponding [frequency, age] both expressed in number of seconds. Every `frequency` seconds, the contents of GRADIO_CACHE_DIR created by this Blocks instance will be deleted if more than `age` seconds have passed since the file was created. The cache will be deleted entirely when the server restarts. If None, no cache deletion will occur.
```
abidlabs, line 520:
```suggestion
            delete_cache: A tuple corresponding [frequency, age] both expressed in number of seconds. Every `frequency` seconds, the temporary files created by this Blocks instance will be deleted if more than `age` seconds have passed since the file was created. The cache will be deleted entirely when the server restarts. If None, no cache deletion will occur.
```

gradio/chat_interface.py
```diff
@@ -103,6 +104,7 @@ def __init__(
             autofocus: If True, autofocuses to the textbox when the page loads.
             concurrency_limit: If set, this is the maximum number of chatbot submissions that can be running simultaneously. Can be set to None to mean no limit (any number of chatbot submissions can be running simultaneously). Set to "default" to use the default concurrency limit (defined by the `default_concurrency_limit` parameter in `.queue()`, which is 1 by default).
             fill_height: If True, the chat interface will expand to the height of window.
+            delete_cache: A tuple corresponding [frequency, age] both expressed in number of seconds. Every `frequency` seconds, the contents of GRADIO_CACHE_DIR created by this Blocks instance will be deleted if more than `age` seconds have passed since the file was created. The cache will be deleted entirely when the server restarts. If None, no cache deletion will occur.
```
abidlabs, line 107:
```suggestion
            delete_cache: A tuple corresponding [frequency, age] both expressed in number of seconds. Every `frequency` seconds, the contents of GRADIO_CACHE_DIR created by this Interface instance will be deleted if more than `age` seconds have passed since the file was created. The cache will be deleted entirely when the server restarts. If None, no cache deletion will occur.
```

abidlabs: We should add this parameter to `ChatInterface` as well

gradio/blocks.py
abidlabs: I'm a little concerned that users might configure this parameter incorrectly, which could lead to undesired behavior since its not clear what a good default value is. For example, setting it to `(60, 60)` would probably break many demos since temporary files will be deleted before their turn in the queue. 

Some ideas:

1. Provide some guidance on a good value in the docstring, e.g. "e.g. setting this to (86400, 86400)" will delete temporary files every day.

2. Allow users to just set this to `True` which will  still do this: "The cache will be deleted entirely when the server restarts." but not have the background task. Should be much easier for users to configure.

abidlabs:
Left some high-level comments above @freddyaboulton! Happy to do a deeper review if/when these are addressed

gradio/blocks.py
freddyaboulton: Can you explain 2 a bit more? How would removing the background task address this comment:

> I think we should provide a mechanism to delete files that does not require a server restart. E.g. all files older than a certain timestamp get deleted

gradio/route_utils.py
freddyaboulton: You can't modify the contents of a set as you're iterating - you get an exception. So I'm removing from the set at the end.

gradio/chat_interface.py
freddyaboulton: I think it's already in ChatInterface!

gradio/route_utils.py
freddyaboulton: Yes good point! Fixed now

gradio/route_utils.py
freddyaboulton: Done thank you!

freddyaboulton:
Thanks for the review @abidlabs ! I addressed your comments. For this [comment](https://github.com/gradio-app/gradio/pull/7447#discussion_r1496856676), I went with option 1.

guides/01_getting-started/03_sharing-your-app.md
```diff
@@ -283,7 +283,7 @@ Sharing your Gradio app with others (by hosting it on Spaces, on your own server
...
+- **Temporary files created by Gradio.** These are files that are created by Gradio as part of running your prediction function. For example, if your prediction function returns a video file, then Gradio will save that video to a temporary cache on your device and then send the path to the file to the front end. You can customize the location of temporary cache files created by Gradio by setting the environment variable `GRADIO_TEMP_DIR` to an absolute path, such as `/home/usr/scripts/project/temp/`. You can delete the files created by your app when it shutsdown with the `delete_cache` parameter of `gradio.Blocks`, `gradio.Interface`, and `gradio.ChatInterface`. This parameter is a tuple of integers of the form `[frequency, age]` where `frequency` is how often to delete files and `age` is the time in seconds since the file was last modified.
```
abidlabs, line 286:
```suggestion
- **Temporary files created by Gradio.** These are files that are created by Gradio as part of running your prediction function. For example, if your prediction function returns a video file, then Gradio will save that video to a temporary cache on your device and then send the path to the file to the front end. You can customize the location of temporary cache files created by Gradio by setting the environment variable `GRADIO_TEMP_DIR` to an absolute path, such as `/home/usr/scripts/project/temp/`. You can delete the files created by your app when it shuts down with the `delete_cache` parameter of `gradio.Blocks`, `gradio.Interface`, and `gradio.ChatInterface`. This parameter is a tuple of integers of the form `[frequency, age]` where `frequency` is how often to delete files and `age` is the time in seconds since the file was last modified.
```

gradio/blocks.py
```diff
@@ -586,7 +589,8 @@ def __init__(
         self.auth = None
         self.dev_mode = bool(os.getenv("GRADIO_WATCH_DIRS", ""))
         self.app_id = random.getrandbits(64)
-        self.temp_file_sets = []
+        self.upload_file_set = set()
```
abidlabs, line 592: Nice fix

gradio/route_utils.py
abidlabs: Ah got it

gradio/blocks.py
```diff
@@ -516,6 +517,7 @@ def __init__(
             js: Custom js or path to js file to run when demo is first loaded. This javascript will be included in the demo webpage.
             head: Custom html to insert into the head of the demo webpage. This can be used to add custom meta tags, scripts, stylesheets, etc. to the page.
             fill_height: Whether to vertically expand top-level child components to the height of the window. If True, expansion occurs when the scale value of the child components >= 1.
+            delete_cache: A tuple corresponding [frequency, age] both expressed in number of seconds. Every `frequency` seconds, the temporary files created by this Blocks instance will be deleted if more than `age` seconds have passed since the file was created. For example, setting this to (86400, 86400) will delete temporary files every day. The cache will be deleted entirely when the server restarts. If None, no cache deletion will occur.
```
abidlabs, line 520:
```suggestion
            delete_cache: A tuple corresponding [frequency, age] both expressed in number of seconds. Every `frequency` seconds, the temporary files created by this Blocks instance will be deleted if more than `age` seconds have passed since the file was created. For example, setting this to (86400, 86400) will delete all temporary files that are at least one day old, every day. Setting any non-None value will also cause the cache will be deleted entirely when the Gradio app is closed using the .close() method. If None, no cache deletion will occur.
```

abidlabs: Note: as far as I could tell, the temporary files are only deleted when the server is closed using the .close() method, but not if the Python process is terminated with cmd+C for example

abidlabs (approved):
Looks awesome @freddyaboulton! Thanks for the nice PR while you're travling

gradio/blocks.py
freddyaboulton: Just re-tested and the files are deleted (as long as they are not used in examples)

sbarman-mi9:
@abidlabs @freddyaboulton This is awesome. Thank you!

PR #7451, opened 2024-02-16
Add error handling for missing `js/_website/version.json`

hannahblair:
## Description

When running `npm i` from root, you'll see an error related to a missing module `version.json` in `js/_website/src/lib/json/`. This is because we need to run `generate_jsons/generate.py` to get that file, but if devs aren't touching any files related website, they shouldn't have to worry about this. This fix dynamically imports `version.json`, catches the error if it doesn't exist, and uses a fallback version (set to 4.0.0 - should this change?).

### DIFF
--- a/.changeset/silent-corners-mix.md
+++ b/.changeset/silent-corners-mix.md
@@ -1,0 +1,5 @@
+---
+"gradio": minor
+---
+
+feat:Add `/logout` functionality for Gradio auth
--- a/.changeset/thin-beers-worry.md
+++ b/.changeset/thin-beers-worry.md
@@ -1,0 +1,15 @@
+---
+"@gradio/button": patch
+"@gradio/chatbot": patch
+"@gradio/client": patch
+"@gradio/dataframe": patch
+"@gradio/dataset": patch
+"@gradio/image": patch
+"@gradio/imageeditor": patch
+"@gradio/simpleimage": patch
+"@gradio/uploadbutton": patch
+"@gradio/video": patch
+"gradio": patch
+---
+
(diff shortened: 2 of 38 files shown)


pngwn (approved):
Looks good. I think 4.0 is fine as a fallback. Thanks @hannahblair!

aliabd:
the changes look fine to me but i can't figure out why website build is failing

aliabd:
@hannahblair it seems to be related to the dynamic import. you need `.then((module) => module.default)` because it returns `_version`  as `{ default: { version: '4.19.0' } }` 

but that actually didn't solve the problem. version is still undefined when there is a version.json, and i'm not sure why. Try debugging with pnpm dev or pnpm build

pngwn (approved):
Looks good! Thanks @hannahblair!

aliabd:
Nice @hannahblair the build errors are gone! but why did you remove `/3.50.2/docs` and `/3.50.2/guides` from the prerender routes? They cause the 404 error when you toggle to that version in docs and guides.

Bringing those back fixes everything though! nice work, this was an annoying bug.

js/_website/svelte.config.js
```diff
@@ -23,8 +52,6 @@ const config = {
 				`/main/docs/js`,
 				`/main/docs/js/storybook`,
 				`/main/docs/js/`,
```
aliabd, line 54:
```suggestion
				`/main/docs/js/`,
				`/3.50.2/docs`,
				`/3.50.2/guides`,
```

aliabd (approved):
lgtm aside from the 3.50.2 stuff!

PR #7453, opened 2024-02-16
Make fix in #7444 (Block /file= filepaths that could expose credentials on Windows) more general

abidlabs:
Slightly more general fix for #7444 to handle IPFS and other crazy formats

### DIFF
--- a/.changeset/bumpy-wasps-march.md
+++ b/.changeset/bumpy-wasps-march.md
@@ -1,0 +1,5 @@
+---
+"gradio": minor
+---
+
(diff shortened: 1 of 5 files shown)


pngwn (approved):
Looks good! Thanks for hardening this @abidlabs!

abidlabs:
Ok so `python-multipart` just released a new version (0.0.9) which changed some of the type annotations, which was causing our typecheck to fail. I've updated the typechecks, everything should pass now and I'll go ahead and merge.

Issue #7442, opened 2024-02-15
Use `uv` in our CI workflows to speed them up 


abidlabs:
Just tested installing `gradio` in a fresh virtual environment:

`pip`: 30 seconds
`uv`: 8 seconds (note this is after `uv` itself has been installed via `pip` which took a few seconds)

See more: https://github.com/astral-sh/uv. `uv` already supports editable installs and local installs so it should work for our CI

tldr: We should replace `pip install` with `uv pip install` in our CI workflows.

abidlabs:
See comments on PR^

pngwn:
Adding a note for clarity: We are leaving this issue open for visibility but we don’t plan to do this work anytime soon and  PRs addressing it will likely be closed without merging until uv is a little more mature.

PR #7457, opened 2024-02-16
Replace `pip` with `uv` in our CI

abidlabs:
I tested `uv` on a fork of `gradio` and found that when a PR requires refreshing the cache, it shaves ~30 seconds off of the python action in our CI.

Here are the approximate runtimes:

| Package manager | No caching |  Caching | 
| ------------- | ------------- | --- |
| `pip`  | [7 min 31 seconds](https://github.com/gradio-app/gradio/pull/7443)  | [6 min 10 seconds](https://github.com/gradio-app/gradio/pull/7407) |
| `uv`  | [6 min 56 seconds](https://github.com/abidlabs/gradio/pull/17)  | [6 min 6 seconds](https://github.com/abidlabs/gradio/pull/18) |

Closes: https://github.com/gradio-app/gradio/issues/7442

### DIFF
--- a/.github/actions/install-all-deps/action.yml
+++ b/.github/actions/install-all-deps/action.yml
@@ -73,8 +73,8 @@ runs:
       shell: bash
       run: |
         . ${{ env.VENV_ACTIVATE }}
-        python -m uv pip install -e client/python  
-        python -m uv pip install -e ".[oauth]"
+        uv pip install -e client/python  
+        uv pip install -e ".[oauth]"
     - name: Install ffmpeg
       uses: FedericoCarboni/setup-ffmpeg@v2
     - name: Install test dependencies
@@ -82,8 +82,8 @@ runs:
       shell: bash
       run: |
         . ${{ env.VENV_ACTIVATE }}
-        python -m uv pip install -r test/requirements.txt
-        python -m uv pip install -r client/python/test/requirements.txt
+        uv pip install -r test/requirements.txt
+        uv pip install -r client/python/test/requirements.txt
     - name: install-frontend
       uses: "gradio-app/gradio/.github/actions/install-frontend-deps@main"
       with:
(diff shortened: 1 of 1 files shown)


abidlabs:
Ah interesting, thanks @akx. Just to understand your comment, we already cache the `venv` where `uv` installs the python libraries after we create the environment. So I agree I can remove `cache: pip` if it serves no purpose. But otherwise, what cache am I missing? 

```py
    - name: Install Python
      uses: actions/setup-python@v5
      with:
        python-version: ${{ inputs.python_version }}
        cache: pip
        cache-dependency-path: |
          client/python/requirements.txt
          requirements.txt
          test/requirements.txt
    - name: Create env
      shell: bash
      run: |
        python -m pip install --upgrade uv
        uv venv venv
    - uses: actions/cache@v4
      id: cache
      with:
        path: |
          venv/*
          client/python/venv
```

akx:
I would maybe probably just opt to cache `~/.cache/uv` instead, since `uv` creating a venv + installing from cache should be very snappy indeed.

abidlabs:
Agreed, `uv` is quite fast at creating the venv but there's no need to create it again if the requirements haven't changed. I'll open up this PR for review and get @pngwn's thoughts as well

pngwn:
Are we sure this is mature enough to rly on? I agree that it is very interesting but it was released about a week ago and isn't really battle tested.

I for example cannot perform a fresh install of `scikit-image==0.19.3` with `uv` as the wheel fails to build. I have no such issues with `pip`.

This also only save us 30 seconds some of the time, which is nice but hardly game-changing. Most of our CI time is spent elsewhere.

pngwn:
I think we should revisit in a short while when some of the kinks have been worked out. It would be nice to speed things up, both locally and in CI.

abidlabs:
Will leave the issue open but close this for now

Issue #7449, opened 2024-02-16
Tabs stop responding to events when visibility is changed


edgarGracia:
### Describe the bug

If a ``gr.Tab`` visibility is set to ``False``, and then to ``True`` again, the tab stops sending events to the ``.select`` event listener. Also, the last tab triggers all the other tabs' events at once.

I tried with and without the ``gr.Tabs()`` block and setting the tab ``id`` but none works.

I'm using ``gr.update(visible=x)`` to change the visibility.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

Run the following code and steps to reproduce the issue. It will create 10 tabs and a slider to select how many tabs to show. The console and a Textbox displays which tab is selected.

1. Launch the demo and click the different tabs. You will see the Textbox and console display the selected tab.
1. Move the slider to, for example, 5 and see the tabs disappear. Now move it back to 10 and see all the tabs again.
1. Now if you click the tabs 0 to 4 it works well, but if you click 5 to 8, it does  not.
1. Also, look at the console and click the last tab. You will see how it triggers the events of tabs 5 to 9.

```python
import gradio as gr

def update_visibility(slider):
    return [
        (gr.update(visible=True) if i < slider else gr.update(visible=False))
        for i in range(10)
    ]

def on_select(i):
    def fn():
        print(i)
        return i
    return fn

with gr.Blocks() as demo:
    slider = gr.Slider(
        minimum=1,
        maximum=10,
        value=10,
        step=1
    )
    text = gr.Textbox()
    
    tab_list = []
    with gr.Tabs() as tabs:
        for i in range(10):
            with gr.Tab(i) as tab:
                tab.select(on_select(i), None, text)
                tab_list.append(tab)
    
    slider.change(
        update_visibility,
        inputs=slider,
        outputs=tab_list
    )

demo.launch()
```

### System Info

```shell
Version: 4.19.0
```

### Severity

I can work around it

skye0402:
I think this is the same like https://github.com/gradio-app/gradio/issues/7189

aliabid94:
taking a look

Issue #7189, opened 2024-01-27
Clicking on tab (Select event) stacks events and doesn't respond


skye0402:
### Describe the bug

I have 3 tabs in my gradio blocks app. When I select a tab, I want to change the visibility of 2 Accordion elements on the right side of the page. The tabs are on the left side.

```
def genai_tab_selected(state: dict):
    """ Handles the click on the Generative AI Playground """
    return [
        gr.Accordion(visible=True),
        gr.Accordion(visible=True),
    ]
```

This is how I handle the visibility (depending on the tab, visible or not). The first tab doesn't trigger the select event at all. However, when I click on the second or third tab (depending on which tab is currently not active), it would call the according `select `event for the first tab and then immediately after call the event of the clicked tab. So it seems to stack the event of the first tab but executes only with the second event.

If I click around a few times I get the error shown in the log.

Any idea what that could be? 

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr
from gradio import Blocks

TEST_MODELS = [
  {
        "model": "gpt-4-32k", 
        "name": "Azure OpenAI GPT-4 32k", 
        "desc": "GPT-4 is a large multimodal model (accepting text or image inputs and outputting text) that can solve difficult problems with greater accuracy than any of our previous models, thanks to its broader general knowledge and advanced reasoning capabilities.", 
        "platform": "SAP Generative AI Hub",
        "maxToken": 16000,
        "hasImage": False,
        "restricted": False,
        "active": True
    }
]

# SQL Playground functions ---------------------
        
def sql_tab_selected(state: dict):
    """ Handles the click on the SQL Generation tab """
    return [
        gr.Accordion(visible=False),
        gr.Accordion(visible=False),
    ]
        
# User tab functions ----------------------------------------

def user_tab_selected(state: dict):
    """ Set user list dataframe """
    login_data = []
    return [
        login_data,
        gr.Accordion(visible=False),
        gr.Accordion(visible=False)
    ]
    
# General AI Playground functions ---------------------------

def genai_tab_selected(state: dict):
    """ Handles the click on the Generative AI Playground """
    return [
        gr.Accordion(visible=True),
        gr.Accordion(visible=True),
    ]
    
def set_user_data(state: dict):
    """ Sets the tabs visibility """
    return [
        gr.Tab(visible=True),
        gr.Tab(visible=True),
    ]

[...]

PR #7470, opened 2024-02-17
Tab select fix

aliabid94:
Fixes issue with tab.select listener not working when it's visiblity is toggled:

Fixes: #7449 Fixes: #7189


js/tabs/shared/Tabs.svelte
```diff
@@ -31,11 +31,13 @@
 
 	setContext(TABS, {
 		register_tab: (tab: Tab) => {
+			let index: number;
 			let existingTab = tabs.find((t) => t.id === tab.id);
 			if (existingTab) {
 				// update existing tab with newer values
 				let i = tabs.findIndex((t) => t.id === tab.id);
 				tabs[i] = { ...tabs[i], ...tab };
+				index = i;
```
akx, line 40:
```suggestion
				index = tabs.findIndex((t) => t.id === tab.id);
				tabs[index] = { ...tabs[index], ...tab };
```

abidlabs:
Thanks @aliabid94! We've increased what's possible with `gr.Tabs` recently, so it'd be great to add an E2E test for tabs to catch any other regressions that might pop up. Would you be able to create one testing these kinds of workflows:

* Ensuring that tabs interactivity can be toggled on and off
* Ensuring that tabs visibility can be toggled on and off
* Ensuring that toggling visibility does not lose the event listeners
* Ensuring that tabs that are not interactive cannot be navigated to

aliabid94:
ok added all the functional tests @abidlabs

abidlabs:
Cool thanks! We should be good to merge then once they pass

PR #7473, opened 2024-02-18
Fixed typo in example: client -> app

up_the_irons:
I was trying to run this example, but it failed.  predict() is a method of the app object, not client.  Fixed!


pngwn (approved):
Ah nice catch. Thanks for fixing @up_the_irons!

PR #7477, opened 2024-02-19
Enable Ruff S101

akx:
## Description

Inspired by https://github.com/gradio-app/gradio/pull/7465#discussion_r1494139137, this enables the Ruff S101 security lint and fixes non-test code not to use `assert` for codepaths that are actually critical to be run at runtime.

Pure type assertions were noqa'd as such, but they too could be an `if`/`raise`.

The assertions were initially rewritten with `ast-grep`:

```
ast-grep run --pattern 'assert $A' --rewrite 'if not ($A): raise AssertionError("$A")' -l python -i gradio client
ast-grep run --pattern 'assert $A, $B' --rewrite 'if not ($A): raise AssertionError($B)' -l python -i gradio client
```

### DIFF
--- a/.changeset/dirty-carrots-care.md
+++ b/.changeset/dirty-carrots-care.md
@@ -1,0 +1,6 @@
+---
+"gradio": minor
+"gradio_client": minor
+---
+
+feat:Enable Ruff S101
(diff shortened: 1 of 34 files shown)


abidlabs (approved):
Thank you @akx for the great PR. Looks solid

Issue #7482, opened 2024-02-20
Gradio can't handle images in HEIC format


smikhai1:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
I've implemented a simple demo to demonstrate the problem. The demo consists of two image components: for input and output images respectively. The output image is just a copy of the input one for simplicity.

```python
from PIL import Image
import gradio as gr


def identity_fn(img: Image) -> Image:
    return img


def main():
    with gr.Blocks() as demo:
        with gr.Row():
            with gr.Column():
                input_img = gr.Image(label='Input image', sources=['upload', 'clipboard'],
                                     elem_id='input_image', type='pil')
                run_btn = gr.Button(value='Run')

            with gr.Column():
                output_img = gr.Image(label='Output image', interactive=False,
                                      elem_id='output_image', type='pil')

        run_btn.click(fn=identity_fn, inputs=[input_img],
                      outputs=[output_img])

    demo.launch(server_name="0.0.0.0", server_port=7606)


if __name__ == '__main__':
    main()
``` 

When I upload image in HEIC format taken on my iPhone camera, I got the following error:
```python
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/gradio/queueing.py", line 495, in call_prediction
    output = await route_utils.call_process_api(
  File "/usr/local/lib/python3.10/dist-packages/gradio/route_utils.py", line 233, in call_process_api
    output = await app.get_blocks().process_api(
  File "/usr/local/lib/python3.10/dist-packages/gradio/blocks.py", line 1604, in process_api
    inputs = await anyio.to_thread.run_sync(
  File "/usr/local/lib/python3.10/dist-packages/anyio/to_thread.py", line 56, in run_sync
    return await get_async_backend().run_sync_in_worker_thread(
  File "/usr/local/lib/python3.10/dist-packages/anyio/_backends/_asyncio.py", line 2134, in run_sync_in_worker_thread
    return await future
  File "/usr/local/lib/python3.10/dist-packages/anyio/_backends/_asyncio.py", line 851, in run
    result = context.run(func, *args)
  File "/usr/local/lib/python3.10/dist-packages/gradio/blocks.py", line 1335, in preprocess_data
    processed_input.append(block.preprocess(inputs_cached))
  File "/usr/local/lib/python3.10/dist-packages/gradio/components/image.py", line 167, in preprocess
    im = PIL.Image.open(file_path)
[...]

PR #7483, opened 2024-02-20
HEIC support for images

smikhai1:
## Description

Adds support of HEIC images by installing PIL's plugin 

Closes: #7482 

## Tests
All tests passed locally

### DIFF
--- a/gradio/__init__.py
+++ b/gradio/__init__.py
@@ -1,5 +1,7 @@
 import json
 
+from pillow_heif import register_heif_opener
+
 import gradio._simple_templates
 import gradio.image_utils
 import gradio.processing_utils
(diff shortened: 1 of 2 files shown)


abidlabs:
Thank you @smikhai1 for creating this PR. I am not a fan of adding more dependencies to Gradio, so let me respond on the original issue

akx (commented):
Pretty sure this (and the extra dependency) shouldn't be in Gradio's core – if an app requires HEIF, it can run these two lines itself?

pngwn:
Yeah, I don't even think this should be a custom component. Just add the extra lines of code. heif is becoming more popular but it isn't that common atm.

abidlabs:
I'll go ahead and close this PR and issue. @smikhai1 please see the suggestions mentioned above

abidlabs:
I think @smikhai1, having a public demo on [Hugging Face Spaces](https://huggingface.co/spaces) might be better for visibility and to provide guidance for users.

Issue #7242, opened 2024-02-01
Errors in "A streaming example using openai"


robmarkcole:
### Describe the bug

Docs errors in [A streaming example using openai](https://www.gradio.app/guides/creating-a-chatbot-fast#a-streaming-example-using-openai)

1) `ValueError: Queue needs to be enabled!` -> resolved with `gr.ChatInterface(predict).queue().launch()`
2) `You tried to access openai.ChatCompletion, but this is no longer supported in openai>=1.0.0` -> TBC

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

As description

### System Info

```shell
Operating System: Linux
gradio version: 3.50.2
gradio_client version: 0.6.1
```

### Severity

I can work around it

abidlabs:
Thanks @robmarkcole. Would you like to contribute a fix to the openai example?

gBlaku:
> Thanks @robmarkcole. Would you like to contribute a fix to the openai example?

For 2. seems like they forgot to update the code to meet new standards:

from openai import OpenAI
client = OpenAI(
    api_key = 'key goes here'
   //anything else you may need
)
response = client.chat.completions.create(model="gpt-3.5-turbo", 
                                     messages = history_openapi_format,
                                     temperature = 1.0,
                                     stream = True
                                     )

PR #7508, opened 2024-02-22
Streaming example for the updated OpenAI API

triakshunn:
## Description
This PR updates the usage of the OpenAI API in accordance with the changes in version 1.0.0. The deprecated `openai.ChatCompletion` has been replaced with `client.ChatCompletion.create()`.

Additionally, the way we extract the content from the API response has been updated. Previously, we checked if the length of `chunk['choices'][0]['delta']` was not zero. Now, we check if `chunk.choices[0].delta.content` is not `None` before appending it to `partial_message`.

Closes: #7242

### DIFF
--- a/guides/04_chatbots/01_creating-a-chatbot-fast.md
+++ b/guides/04_chatbots/01_creating-a-chatbot-fast.md
@@ -198,10 +198,11 @@ gr.ChatInterface(predict).launch()
 Of course, we could also use the `openai` library directy. Here a similar example, but this time with streaming results as well:
 
 ```python
-import openai
+from openai import OpenAI
 import gradio as gr
 
-openai.api_key = "sk-..."  # Replace with your key
+api_key = "sk-..."  # Replace with your key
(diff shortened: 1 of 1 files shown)


guides/04_chatbots/01_creating-a-chatbot-fast.md
```diff
@@ -198,30 +198,29 @@ gr.ChatInterface(predict).launch()
 Of course, we could also use the `openai` library directy. Here a similar example, but this time with streaming results as well:
 
 ```python
-import openai
+from openai import OpenAI
 import gradio as gr
 
-openai.api_key = "sk-..."  # Replace with your key
+api_key = "sk-..."  # Replace with your key
+client = OpenAI(api_key)
```
abidlabs, line 205:
```suggestion
client = OpenAI(api_key=api_key)
```

abidlabs (approved):
LGTM thanks for the contribution @triakshunn!

Issue #7509, opened 2024-02-22
Meta tags with "og:*" property is duplicated in frontend when using "head" parameter


qkdxorjs1002:
### Describe the bug

When I set open graph meta tags into the "head" parameter, they are not reflected on the web page.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

tts_examples = [
    "I love learning machine learning",
    "How do you do?",
]

tts_demo = gr.load(
    "huggingface/facebook/fastspeech2-en-ljspeech",
    title=None,
    examples=tts_examples,
    description="Give me something to say!",
)

stt_demo = gr.load(
    "huggingface/facebook/wav2vec2-base-960h",
    title=None,
    inputs=gr.Microphone(type="filepath"),
    description="Let me try to guess what you're saying!",
)

head = """
<meta property="og:type" content="website">
<meta property="og:url" content="http://example.com/">
<meta property="og:title" content="example">
<meta property="og:site_name" content="example">
<meta property="og:locale" content="ko_KR">
"""

demo = gr.TabbedInterface([tts_demo, stt_demo], ["Text-to-speech", "Speech-to-text"], head=head)

if __name__ == "__main__":
    demo.launch()
```

### Screenshot

![image](https://github.com/gradio-app/gradio/assets/3241479/ea64bafd-2829-4ed9-a144-500644a08484)

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Darwin
gradio version: 4.19.0
gradio_client version: 0.10.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.2.0
fastapi: 0.109.2
ffmpy: 0.3.2
gradio-client==0.10.0 is not installed.
httpx: 0.26.0
huggingface-hub: 0.20.3
importlib-resources: 6.1.1
jinja2: 3.1.3
markupsafe: 2.1.5
matplotlib: 3.8.3
numpy: 1.26.4
orjson: 3.9.14
packaging: 23.2
pandas: 2.2.0
pillow: 10.2.0
pydantic: 2.6.1
pydub: 0.25.1
python-multipart: 0.0.9
pyyaml: 6.0.1
ruff: 0.2.1
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.9.0
typing-extensions: 4.9.0
uvicorn: 0.27.1
[...]

PR #7510, opened 2024-02-22
when adding custom head html, ensure there are no duplicate meta tags

qkdxorjs1002:
remove old element before append new meta tag to prevent duplication

## Description

I add some code with statement to remove target of duplication before appending new element.

Closes: #7509

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/js/app/src/Index.svelte
+++ b/js/app/src/Index.svelte
@@ -166,10 +166,9 @@
 
 			if (parsed_head_html) {
 				for (let head_element of parsed_head_html) {
-					const head_element_tagName = head_element.tagName.toLowerCase();
 					const head_element_property = head_element.getAttribute("property");
-					
-					if (head_element_tagName == "meta" && head_element_property) {
+
+					if (head_element.tagName == "META" && head_element_property) {
 						const domMetaList = Array.from(
 							document.head.getElementsByTagName("meta") ?? []
 						);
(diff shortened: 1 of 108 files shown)


js/app/src/Index.svelte
```diff
@@ -166,6 +166,20 @@
...
+					if (head_element.tagName == "META" && head_element_property) {
+						const domMetaList = Array.from(
+							document.head.getElementsByTagName("meta") ?? []
+						);
+						const matched = domMetaList.find(
+							(el) => el.getAttribute("property") == head_element_property
+						);
+						if (matched) {
+							document.head.removeChild(matched);
+						}
```
pngwn, line 180: If it matches (i.e. it is functionally the same meta tag) why remove it only to re-add it, can't we just skip the additional work and  `continue;` the loop here instead?

qkdxorjs1002: My intention is to address the issue of duplication that arises when both `default meta tags are predefined in index.html` and `developers add meta tags` with the `head` parameter, which serve the same purpose. 

When rendered in the browser, two tags with similar functionalities exist, albeit with potentially different values or content. This duplication issue prevents the `meta tag specified by the developer` from functioning as intended.

I've written code to remove any existing `meta tags with equivalent functionality but differing values or content` before adding the specified HTML element with the additional `head` parameter. 

Additionally, I've provided further commits outlining improvements.
The improvement entails replacing `meta tags with the same property attribute`, ensuring that tags serving identical functions are substituted when they are not equivalent.

Thank you for your review.

pngwn (approved):
This looks great! Thanks for addressing the feedback @qkdxorjs1002!

Issue #5426, opened 2023-09-06
a `gr.DownloadButton` that lets you download a file with a button click


hyer:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
Is possible to directly download file with a button click, rather using the `gr.File` component. Directly download will be more intuitive, and using the gr. File requires two operations.

**Describe the solution you'd like**  
I don't want to add the `gr.File` component on my app's UI. I'd like to download a file (eg: image) from a http url or local path with a button click.

abidlabs:
Hi @hyer we don't have a `gr.DownloadButton` component -- we could create a component like this in the future. The other possibility would be to use the `link` attribute of the `gr.Button()` to point to the filepath. Something like this:

```py
import gradio as gr

with gr.Blocks() as demo:
    gr.Button(link="/file=cheetah.jpg")
    
demo.launch()
```

cowanAI:
@abidlabs could you do this code: 

On interface too? import gradio as gr

with gr.Blocks() as demo:
    gr.Button(link="/file=cheetah.jpg")
    
demo.launch() thanks for your help on this.

abidlabs:
Not sure I understand the question @cowanAI can you explain what you mean?

cowanAI:
Let’s say I make a gradio interface not blocks, and then add a microphone
component to it, can we use the link button to download the audio in that
way?

abidlabs:
Ah yeah it should be possible I think. Let's say your filepath is `filepath` You can return a `gr.Button.update(link="file/=" + filepath)` from your function

hyer:
> 

Thanks. However, this is viewable only in the browser, not the usual way of saving to local disk. Look forward to seeing the `gr.DownloadButton` component soon

abidlabs:
Hi @pcschreiber1 -- which version of gradio are you using?

pcschreiber1:
Thank you for the quick reply @abidlabs , I'm using gradio 4.7.1

abidlabs:
In Gradio 4.x, files your local system are not exposed by default, so you'll need to explicitly allow the specific file or directory. E.g. 

```py
import gradio as gr

with gr.Blocks() as demo:
...
    
demo.launch(allowed_paths=["file/material/test.txt"])
```

or you could do a containing folder as well, like:

```py
import gradio as gr

with gr.Blocks() as demo:
...
    
demo.launch(allowed_paths=["file/"])
```

[...]

PR #7518, opened 2024-02-23
Adds a `gr.DownloadButton` component

abidlabs:
After https://github.com/gradio-app/gradio/pull/7528, I wanted to go through the process of creating a new component with the new approach of working with the files, i.e. setting the URLs in the backend using `serve_static_file()` to mimic the work that a custom component developer would have to do.

I decided to add the long-requested `gr.DownloadButton` component (fixes: https://github.com/gradio-app/gradio/issues/5426)


abidlabs:
Will probably need to be rebased once https://github.com/gradio-app/gradio/pull/7528 is in

demo/fake_gan/files/cheetah1.jpg
abidlabs, line 1: Unused file

js/uploadbutton/shared/UploadButton.svelte
abidlabs, line 1: Fixing the casing in this file

gradio/components/download_button.py
```diff
@@ -0,0 +1,105 @@
...
+        icon: str | None = None,
+        scale: int | None = None,
+        min_width: int | None = None,
+        interactive: bool = True,
+        elem_id: str | None = None,
+        elem_classes: list[str] | str | None = None,
+        render: bool = True,
+    ):
+        """
+        Parameters:
```
dawoodkhan82, line 43: Missing param for `icon`

dawoodkhan82:
```
from pathlib import Path
import gradio as gr

def upload_file(filepath):
    name = Path(filepath).name
    return [gr.UploadButton(visible=False), gr.DownloadButton(label=f"Download {name}", value=filepath, visible=True, icon="gradio-logo.png")]

def download_file():
    return [gr.UploadButton(visible=True), gr.DownloadButton(visible=False)]

with gr.Blocks() as demo:
    gr.Markdown("First upload a file and and then you'll be able download it (but only once!)")
    with gr.Row():
        u = gr.UploadButton("Upload a file", file_count="single")
        d = gr.DownloadButton("Download a file", visible=False, icon="gradio-logo.png")

    u.upload(upload_file, u, [u, d])
    d.click(download_file, None, [u, d])


if __name__ == "__main__":
    demo.launch()
```

Tested with this code. And seems like updating the `icon` for the `DownloadButton` doesn't work. Gives error: `AttributeError: 'DownloadButton' object has no attribute 'proxy_url'`.

gradio/components/download_button.py
abidlabs: Added

abidlabs:
Nvm the fix was easier than I thought -- I've merged it into this PR. Your example works for me now @dawoodkhan82

dawoodkhan82 (approved):
LGTM, tested updating the icon as well. Thanks for including that fix @abidlabs

abidlabs:
Thanks @dawoodkhan82 for testing thoroughly!

Issue #7471, opened 2024-02-18
regression on gr.File, can't upload files from an iPhone anymore


ngaudemet:
### Describe the bug

thanks for the tremendous work on gradio !

in gradio 4.19.1 (probably since 4.18.0 file enhancement https://github.com/gradio-app/gradio/pull/7299), I can't upload files on iPhone anymore.

In gradio 4.17.0, I could upload from iPhone (in precising specific types). Now it doesn't work anymore

issue #4021 was similar but fixed 

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

files = gr.File(
            label="load_file",
            file_count="multiple",
            file_types=[
                ".csv",
                ".txt",
                ".json",
                ".xlsx",
                ".docx",
                ".pdf",
                ".jpg",
                ".png",
                ".pptx",
            ],
            visible=True,
        )

### Screenshot

![IMG_9375](https://github.com/gradio-app/gradio/assets/126974970/9f5bf89c-69b6-47e3-8f9f-2599de3982ea)

### System Info

```shell
gradio==4.19.1
```

### Severity

I can work around it (in staying with 4.17.0)

abidlabs:
cc @dawoodkhan82

dawoodkhan82:
@ngaudemet seems like this is actually fixed on main. So should be working in the next release.

PR #7525, opened 2024-02-23
Upload files from iPhone fix

dawoodkhan82:
## Description

Skips `mime_type` check when uploading files from iPhone, since mime types are not always guaranteed with iPhone upload. 

Closes: #7471

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/eager-pugs-retire.md
+++ b/.changeset/eager-pugs-retire.md
@@ -1,0 +1,6 @@
+---
+"@gradio/upload": patch
+"gradio": patch
+---
+
+fix:Upload files from iPhone fix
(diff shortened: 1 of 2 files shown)


dawoodkhan82:
Having trouble testing this with browserstack. Also this essentially skips the file type check, so not the best solution. Open to suggestions!

aliabid94:
I am not able to reproduce the original issue - off of main, I set the types to ".pdf" and it did filter to only pdfs. What did you see @dawoodkhan82 ?

dawoodkhan82:
@aliabid94 You tested on iPhone?

aliabid94:
yep. I was able to select only pdfs

PR #7528, opened 2024-02-23
Refactors `get_fetchable_url_or_file()` to remove it from the frontend

abidlabs:
This PR follows up https://github.com/gradio-app/gradio/pull/7183 by removing `get_fetchable_url_or_file()` from the frontend. Instead, all files are normalized in the backend. Now, we have a consistent way of serving files in the backend., and this means the frontend components never have to deal with `proxy_url` or `root`, or `samples_dir`, which simplifies things the frontend quite a bit and also makes things a bit easier for custom component developers. 

All changes should be backwards compatible with existing custom components.

### DIFF
--- a/gradio/components/button.py
+++ b/gradio/components/button.py
@@ -62,7 +62,7 @@ class Button(Component):
             scale=scale,
             min_width=min_width,
         )
-        self.icon = self.move_resource_to_block_cache(icon)
+        self.icon = self.serve_static_file(icon)
         self.variant = variant
         self.size = size
         self.link = link
--- a/js/dataframe/Example.svelte
+++ b/js/dataframe/Example.svelte
@@ -1,10 +1,5 @@
 <script lang="ts">
-	import { csvParseRows, tsvParseRows } from "d3-dsv";
-	import type { Gradio } from "@gradio/utils";
-
-	export let gradio: Gradio;
 	export let value: (string | number)[][] | string;
-	export let samples_dir: string;
 	export let type: "gallery" | "table";
 	export let selected = false;
 	export let index: number;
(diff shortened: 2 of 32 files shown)


gradio/components/button.py
```diff
@@ -62,7 +62,7 @@ def __init__(
             scale=scale,
             min_width=min_width,
         )
-        self.icon = self.move_resource_to_block_cache(icon)
+        self.icon = self.serve_static_file(icon)
```
abidlabs, line 65: I think this is a "friendlier" name

js/dataframe/Example.svelte
```diff
@@ -1,53 +1,12 @@
...
 	export let value: (string | number)[][] | string;
-	export let samples_dir: string;
 	export let type: "gallery" | "table";
 	export let selected = false;
 	export let index: number;
 
 	let hovered = false;
 	let loaded_value: (string | number)[][] | string = value;
 	let loaded = Array.isArray(loaded_value);
-
```
abidlabs, line 15: If a string CSV path is passed for examples for `gr.Dataframe`, the backend reads the file and sends the head as an array (in `gr.Dataframe.process_example()`), so this code is never used.

abidlabs:
@hannahblair it looks like the UI test action is failing -- do you know why that might be the case? I wonder if it has to do with #7513 or #7514 though both looked good to me

pngwn:
Looks like a config issue: https://github.com/gradio-app/gradio/actions/runs/8029101596/job/21935791561#step:7:94

aliabid94 (approved):
LGTM

abidlabs:
Thanks @aliabid94 for the review! I think I'll go ahead and merge this in since its blocking me on other PRs and we can figure out the UI tests issue later (cc @hannahblair)

PR #7544, opened 2024-02-26
Prevent paths beginning with `//` or `\\`

abidlabs:
Path starting with // or \\ on Windows are interpreted as SMB share paths and automatically opened with the open function.


aliabid94 (approved):
LGTM

abidlabs:
Thanks @aliabid94!

Issue #5163, opened 2023-08-10
Request for Adding Logout Functionality


Buddies-as-you-know:
- [x] I have searched to see if a similar issue already exists.
Is your feature request related to a problem? Please describe.
I often find it inconvenient when using Gradio that there isn't a way to log out from the system once I'm done using it. This creates a potential security risk, especially when using shared computers or public networks.

**Describe the solution you'd like**
I would appreciate it if a logout functionality could be implemented in Gradio. This would allow users to securely log out after they have finished using the platform, ensuring that their sensitive information and session data are properly cleared.

**Additional context**
It would be helpful if the logout option could be easily accessible from the user interface, perhaps in the form of a profile or account settings menu. This feature would enhance the overall user experience and provide an added layer of security to the application. Screenshots or mockups demonstrating the proposed logout feature's placement and design might be beneficial for consideration.

ODENSUKI:
Hello.
Any progress on this issue?

yosephinestwn:
> Hi, Modifing the routes.py file in gradio adding the following code , makes the logout button work.
> 
> ```
> @app.get("/logout")
>         def logout():
>             response = RedirectResponse(url="/", status_code=status.HTTP_302_FOUND)
>             response.delete_cookie(key=f"access-token-{app.cookie_id}")
>             response.delete_cookie(key=f"access-token-unsecure-{app.cookie_id}")
>             return response
> ```
> 
> Wonder if theres anyway to do that from main app withou modifing gradio files.

This solution does not work for me. The current user is still logged in if the button is clicked. Is there anything that you do (e.g. to the event listener of the logout button) other than only modifying the routes.py?

cpatrickalves:
> Hi, Modifing the routes.py file in gradio adding the following code , makes the logout button work.
> 
> ```
> @app.get("/logout")
>         def logout():
>             response = RedirectResponse(url="/", status_code=status.HTTP_302_FOUND)
>             response.delete_cookie(key=f"access-token-{app.cookie_id}")
>             response.delete_cookie(key=f"access-token-unsecure-{app.cookie_id}")
>             return response
> ```
> 
> Wonder if theres anyway to do that from main app withou modifing gradio files.

This solution also did not work for me. 
Does anyone have any other solution?

PR #7547, opened 2024-02-26
Add `/logout` functionality for Gradio auth

abidlabs:
Addresses a long-standing request, closes: https://github.com/gradio-app/gradio/issues/5163 as well as tests and documentation

### DIFF
--- a/gradio/routes.py
+++ b/gradio/routes.py
@@ -50,7 +50,7 @@ from gradio import ranged_response, route_utils, utils, wasm_utils
 from gradio.context import Context
 from gradio.data_classes import ComponentServerBody, PredictBody, ResetBody
 from gradio.exceptions import Error
-from gradio.oauth import _redirect_to_target, attach_oauth
+from gradio.oauth import attach_oauth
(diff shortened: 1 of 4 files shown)


abidlabs:
Thanks @hannahblair for confirming!

arieltoledo:
Hi, just upgrade gradio, and added the fix here. 
Solution its not working for me runing auth in blocks launch method like this:
**with gr.Blocks(css=css, js=js) as demo:**
**....**
**demo.launch(auth=(username, password), server_name="0.0.0.0", server_port=5000)**
Think got to be something realted to this line
` if app.blocks is not None and app.blocks.expects_oauth:
            attach_oauth(app)
        else:`
because removing it gets the logout functionality back again. 
Thanks

abidlabs:
Hi @arieltoledo just to confirm, you're installing Gradio from this branch or from main? Secondly, can you please provide a minimal code example that we can use to reproduce the issue above? See: https://stackoverflow.com/help/minimal-reproducible-example

Issue #6717, opened 2023-12-09
file upload failed using gradio_client.Client, when there is an auth


ArslanTu:
### Describe the bug

Setup an gradio server with auth, uploading file through browser works fine. But it didn't work for gradio_client.
When I remove the auth setting, gradio_client works fine, too.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

`server.py`

```python
import gradio as gr

def func(file_obj) -> str:
    return file_obj.name

def auth(username: str, passwd: str) -> bool:
    return username == "tom" and passwd == "123"

def main():
    demo = gr.Interface(
        fn=func,
        inputs=[gr.File()],
        outputs=[gr.Textbox()],
    )
    demo.queue(api_open=True)
    demo.launch(
        server_port=9600,
        auth=auth,
    )

if __name__ == "__main__":
    main()
```

`test_upload.py`

```python
from gradio_client import Client
from time import sleep

client = Client("http://127.0.0.1:9600", auth=("tom", "123"))
job = client.submit(
    "./test_file.json",
    api_name="/predict",
)
while not job.done():
    sleep(1)
print(job.result())
```

### Logs

```shell
Traceback (most recent call last):
  File "/home/tom/anaconda3/envs/referqa/lib/python3.10/site-packages/gradio/queueing.py", line 456, in call_prediction
    output = await route_utils.call_process_api(
  File "/home/tom/anaconda3/envs/referqa/lib/python3.10/site-packages/gradio/route_utils.py", line 232, in call_process_api
    output = await app.get_blocks().process_api(
  File "/home/tom/anaconda3/envs/referqa/lib/python3.10/site-packages/gradio/blocks.py", line 1520, in process_api
    inputs = self.preprocess_data(fn_index, inputs, state)
  File "/home/tom/anaconda3/envs/referqa/lib/python3.10/site-packages/gradio/blocks.py", line 1300, in preprocess_data
    inputs_cached = block.data_model(**inputs_cached)  # type: ignore
TypeError: gradio.data_classes.FileData() argument after ** must be a mapping, not str
Traceback (most recent call last):
  File "/home/tom/anaconda3/envs/referqa/lib/python3.10/site-packages/gradio/queueing.py", line 456, in call_prediction
    output = await route_utils.call_process_api(
  File "/home/tom/anaconda3/envs/referqa/lib/python3.10/site-packages/gradio/route_utils.py", line 232, in call_process_api
[...]

PR #7555, opened 2024-02-27
Allow Python Client to upload/download files when connecting to Gradio apps with auth enabled

abidlabs:
Fixes: https://github.com/gradio-app/gradio/issues/6717

### DIFF
--- a/client/python/gradio_client/client.py
+++ b/client/python/gradio_client/client.py
@@ -32,7 +32,7 @@ from packaging import version
 
 from gradio_client import serializing, utils
 from gradio_client.documentation import document
-from gradio_client.exceptions import SerializationSetupError
+from gradio_client.exceptions import AuthenticationError, SerializationSetupError
 from gradio_client.utils import (
     Communicator,
     JobStatus,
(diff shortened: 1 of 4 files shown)


client/python/gradio_client/client.py
```diff
@@ -731,7 +731,7 @@ def _login(self, auth: tuple[str, str]):
             data={"username": auth[0], "password": auth[1]},
         )
         if not resp.is_success:
-            raise ValueError(f"Could not login to {self.src}")
+            raise AuthenticationError(f"Could not login to {self.src}")
```
aliabd, line 734: just fyi this will raise even when the problem isn't authentication, but the link being down for example because the space is sleeping. don't think it matters but if you want to make it strictly for authentication like you described in exceptions.py you should look at the status code.

aliabd (approved):
lgtm! nice work

abidlabs:
Thanks @aliabd! Appreciate the review

djaym7:
which release has this fix ?

djaym7:
[gradio_client@0.11.0](https://github.com/gradio-app/gradio/releases/tag/gradio_client%400.11.0) found it

Issue #7005, opened 2024-01-11
secure Gradio via FastAPI with external authentication


jayadnure:
### Describe the bug

We are trying to secure Gradeo application via FastAPI by having secure authentication in place. The routes defined by FastAPI are secured but when trying to access the mounted Gradeo application via a route it does not take authentication into consideration.
How can we impose strict authentication on accessing the Gradeo resource via fastapi.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

from fastapi import FastAPI,Request
import gradio as gr
from gradio_ui import demo
import uvicorn
from sap import xssec
from cfenv import AppEnv
from starlette.middleware.sessions import SessionMiddleware
from starlette.responses import HTMLResponse, RedirectResponse

app = FastAPI()
env = AppEnv()
uaa_service = env.get_service(name='gradiouaa').credentials

@app.get('/')
async def root(request: Request):
    print(request.headers)
    if 'authorization' not in request.headers:
         return 'not authenticated', 403
    access_token = request.headers.get('authorization')[7:]
    security_context = xssec.create_security_context(access_token, uaa_service)
    isAuthorized = security_context.check_scope('openid')
    if not isAuthorized:
         return 'not authenticated', 403
    return 'Gradio app is running at /gradio', 200
    
app = gr.mount_gradio_app(app, demo, path='/gradio')

if __name__ == "__main__":
    uvicorn.run(app, host="0.0.0.0", port=port)

### System Info

```shell
gradio version
4.13.0

FastApi
0.108.0
```

### Severity

Blocking usage of gradio

abidlabs:
@jayadnure if you'd like to test this feature, I have a PR ready here: https://github.com/gradio-app/gradio/pull/7557

You can install this Gradio from this PR by doing: 

```bash
pip install !pip install -q https://gradio-builds.s3.amazonaws.com/abf7bda7be2e2375e09514fde0ea8c86491900af/gradio-4.19.2-py3-none-any.whl
```

And you can add e.g. Google OAuth to your Gradio app like this:

```py
import os
from authlib.integrations.starlette_client import OAuth, OAuthError
from fastapi import FastAPI, Depends, Request
from starlette.config import Config
from starlette.responses import RedirectResponse
from starlette.middleware.sessions import SessionMiddleware
import uvicorn
import gradio as gr

app = FastAPI()

# OAuth settings
GOOGLE_CLIENT_ID = "..."
[...]

Issue #2790, opened 2022-12-11
Google Authentication


yuvalkirstain:
In some apps, we want to filter harmful users or bots. I think that having a component that enables google authentication (rather than a username, passowrd authentication) in gradio can be very helpful. 

**Describe the solution you'd like**  
I'd like to have an authentication component that receives the details of the user so it can decide if the user may access the app or not.

yuvalkirstain:
This code is partial (and unfortunately non-elegant) solution. It: 
1. creates a new app that mounts the gradio app
2. the new app is in charge of the google authentication. 
 
A key aspect that is missed here, is that in the case that the user goes straight ahead to the gradio endpoint, there is no redirection to login and the user is stuck there.  

```python
import json
from authlib.integrations.base_client import OAuthError
from fastapi import FastAPI
from starlette.middleware.sessions import SessionMiddleware
from starlette.responses import HTMLResponse, RedirectResponse
from starlette.requests import Request
import gradio as gr
from authlib.integrations.starlette_client import OAuth
from starlette.config import Config

# CODE FOR NEW APP

app = FastAPI()
app.add_middleware(SessionMiddleware, secret_key="!secret")

config = Config('.env')
oauth = OAuth(config)

CONF_URL = 'https://accounts.google.com/.well-known/openid-configuration'
oauth.register(
    name='google',
    server_metadata_url=CONF_URL,
    client_kwargs={
        'scope': 'openid email profile'
    }
)


@app.get('/')
async def homepage(request: Request):
    user = request.session.get('user')
    if user:
        data = json.dumps(user)
        html = (
            f'<pre>{data}</pre>'
            '<a href="/logout">logout</a>'
            '<br>'
            '<a href="/gradio">demo</a>'
        )
        return HTMLResponse(html)
    return HTMLResponse('<a href="/login">login</a>')


@app.get('/login')
async def login(request: Request):
    redirect_uri = request.url_for('auth')
    return await oauth.google.authorize_redirect(request, redirect_uri)


@app.get('/auth')
async def auth(request: Request):
    print(f"before request user {request.session.get('user')}")
    try:
        token = await oauth.google.authorize_access_token(request)
    except OAuthError as error:
        return HTMLResponse(f'<h1>{error.error}</h1>')
    user = token.get('userinfo')
    if user:
        request.session['user'] = dict(user)
    print(f"after request user {request.session.get('user')}")
    return RedirectResponse(url='/')


@app.get('/logout')
async def logout(request: Request):
[...]

PR #7557, opened 2024-02-28
Allow mounted Gradio apps to work with external / arbitrary authentication providers

abidlabs:
Final auth-related PR and after this, I'm working on other stuff! But I think this one is pretty cool as it allows mounted Gradio apps to work with e.g. Google OAuth, which is a long-requested feature.

Closes: https://github.com/gradio-app/gradio/issues/2790
Closes: https://github.com/gradio-app/gradio/issues/7005

Example running on Spaces: https://huggingface.co/spaces/gradio/oauth-example
Example running on AWS: http://35.85.61.147.nip.io/main/ (since this is HTTP, I will need to manually approve your email if you'd like to try. DM me if you would like that)


abidlabs:
Also the recent Hub outage made me realize that we have more flaky tests than we realized. I've marked ~8 additional tests as flaky.

js/app/test/load_space.spec.ts
abidlabs, line 1: I deleted this test because its rather simplistic and makes our functional tests depend on the Hub and therefore flaky. The functionality that this test is testing can be sufficiently tested by a backend python test with a fastapi test client that makes a request to the api endpoint

pievalentin:
I would happily test it in Azure AD once it get merged!

abidlabs:
Thanks @pievalentin if you'd like, you can already test it by installing gradio like this:

```bash
pip install https://gradio-builds.s3.amazonaws.com/f78a94031344a956d40cbff23698188b8a900f9f/gradio-4.19.2-py3-none-any.whl
```

In fact, if you could test it, we'd appreciate it so that we can incorporate your feedback into the PR itself. If it helps, here's an example OAuth running on Spaces: https://huggingface.co/spaces/gradio/oauth-example

pievalentin:
Sounds good! I have a lot on my plate at work but I think I will have time to try it this weekend. Looking at your example it looks straightforward :)

guides/01_getting-started/03_sharing-your-app.md
```diff
@@ -275,38 +307,119 @@ Users can revoke access to their profile at any time in their [settings](https:/
...
 
-## Accessing the Network Request Directly
 
-When a user makes a prediction to your app, you may need the underlying network request, in order to get the request headers (e.g. for advanced authentication), log the client's IP address, getting the query parameters, or for other reasons. Gradio supports this in a similar manner to FastAPI: simply add a function parameter whose type hint is `gr.Request` and Gradio will pass in the network request as that parameter. Here is an example:
+### OAuth (with external providers)
+
+It is also possible to authenticate with external OAuth providers (e.g. Google OAuth) in your Gradio apps. To do this, first mount your Gradio app within a FastAPI app ([as discussed above](#mounting-within-another-fastapi-app)). Then, you must write an *authentication function*, which gets the user's username from the OAuth provider and returns it. This function should be passed to the `auth_dependency` parameter in `gr.mount_gradio_app`.
```
aliabd, line 313: believe it or not, the one missing dash in this link is what's breaking the website build

aliabd:
```suggestion
It is also possible to authenticate with external OAuth providers (e.g. Google OAuth) in your Gradio apps. To do this, first mount your Gradio app within a FastAPI app ([as discussed above](#mounting-within-another-fast-api-app)). Then, you must write an *authentication function*, which gets the user's username from the OAuth provider and returns it. This function should be passed to the `auth_dependency` parameter in `gr.mount_gradio_app`. 
```

aliabid94:
When I test it out using the spaces link, it opens a new tab for google auth, and then after login, the new tab becomes the logged in gradio app. Is this the expected flow?

abidlabs:
Yes because of Spaces IFrames. In non-iframes it should work directly

aliabid94:
Even when opening https://gradio-oauth-example.hf.space/main/ directly, this was the case.

abidlabs:
>  Even when opening https://gradio-oauth-example.hf.space/main/ directly, this was the case.

Yes sorry I hardcoded this behavior to open in a new tab but I can change the AWS EC2 one to open in the same tab

abidlabs:
@aliabid94 see here for an example opening in same page: http://35.85.61.147.nip.io/

freddyaboulton (approved):
This looks good to me @abidlabs ! I don't think passing both `auth` and `auth_dependency` makes sense. Should we disallow that? Can we consolidate `auth` and `auth_dependency`?

abidlabs:
Thanks @freddyaboulton! Good point, I'll raise a ValueError if both are provided in order to not cause any unexpected security issues for users

aliabid94 (approved):
code lgtm

gradio/components/download_button.py
```diff
@@ -25,7 +25,7 @@ class DownloadButton(Component):
     def __init__(
         self,
         label: str = "Download",
-        value: str | list[str] | Callable | None = None,
```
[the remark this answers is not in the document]
abidlabs: No that was a mistake before, download button can only let you download a single file at a time

abidlabs:
Thanks @aliabid94 and @freddyaboulton for the nice reviews!

guides/01_getting-started/03_sharing-your-app.md
```diff
@@ -275,38 +307,119 @@ Users can revoke access to their profile at any time in their [settings](https:/
...
+        return RedirectResponse(url='/login-demo')
+
+@app.route('/logout')
+async def logout(request: Request):
+    request.session.pop('user', None)
+    return RedirectResponse(url='/')
+
+@app.route('/login')
+async def login(request: Request):
+    redirect_uri = request.url_for('auth')
```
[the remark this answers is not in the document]
abidlabs: Yes that's a good idea. I'll make a quick PR to add this

Issue #7531, opened 2024-02-24
gr.Audio blocked by nginx


pseudotensor:
### Describe the bug

gr.Audio playing gets stuck and doesn't play, and then never goes to stop/pause mode after clicking play. However, this only happens for the indirect nginx DNS not the direct port.

https://github.com/gradio-app/gradio/issues/7391#issuecomment-1962284080

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

Just any gr.Audio with some file path.

There are some bad console errors even on startup of app:

![image](https://github.com/gradio-app/gradio/assets/2249614/a507af90-7079-414a-86af-57e3230183e4)

But new errors appear after trying to view the audio element before pressing play:

![image](https://github.com/gradio-app/gradio/assets/2249614/aff97d6b-a005-496f-83ea-20e376f5a16a)

The element also looks off, no time, should be 1s:

![image](https://github.com/gradio-app/gradio/assets/2249614/9e3bb7c1-fde7-44be-8e1e-fdbd303a8675)

Then when clicking play it just gets stuck:

![image](https://github.com/gradio-app/gradio/assets/2249614/abd268fc-db91-4417-999a-58db3e9930c2)

### System Info

```shell
gradio 4.19.1
```

### Severity

Blocking usage of gradio

pseudotensor:
One can see it by going to https://gpt.h2o.ai and uploading small wave file, then doing to document viewer, selecting that file, then clicking play.

One can compare that to real address at http://38.128.232.144:13001/ that has no issues.

pseudotensor:
If useful, you can compare and look at dev console:

http://38.128.232.144:13001/ (plays audio that was uploaded and looked at in doc viewer)
vs.
https://gpt.h2o.ai (fails to play audio that was uploaded and looked at in doc viewer)


working:

![image](https://github.com/gradio-app/gradio/assets/2249614/64c554cf-50b1-4c8c-bf17-d0c15ddad7e6)


not playing:

![image](https://github.com/gradio-app/gradio/assets/2249614/12694c9e-b3b0-4493-9b7b-c910471dc94a)

abidlabs:
Noticed that this happens with share links as well:

```py
import gradio as gr

with gr.Blocks() as demo:
    gr.Audio("https://file-examples.com/storage/fe98e6ab1965df346a899de/2017/11/file_example_MP3_700KB.mp3")
    
demo.launch(share=True)
```

PR #7565, opened 2024-02-29
Fixes method to resolve the `root` URLs

abidlabs:
There was a mistake in how we were updating the `root` url, so that they weren't being updated correctly after the first time, which was leading to issues like this: https://github.com/gradio-app/gradio/issues/7531

Test with:

```py
import gradio as gr

with gr.Blocks() as demo:
    gr.Audio("https://file-examples.com/storage/fe98e6ab1965df346a899de/2017/11/file_example_MP3_700KB.mp3")
    
demo.launch(share=True)
```

Fixes: https://github.com/gradio-app/gradio/issues/7531
Fixes: https://github.com/gradio-app/gradio/pull/7566

### DIFF
--- a/test/test_processing_utils.py
+++ b/test/test_processing_utils.py
@@ -361,7 +361,7 @@ def test_add_root_url():
     new_expected = {
         "file": {
             "path": "path",
-            "url": f"{root_url}/file=path",
+            "url": f"{new_root_url}/file=path",
         },
         "file2": {
             "path": "path2",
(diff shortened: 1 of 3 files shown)


test/test_processing_utils.py
```diff
@@ -361,13 +361,13 @@ def test_add_root_url():
     new_expected = {
```
abidlabs, line 361: This test was wrong which is why we didn't notice this bug sooner (this is what I get for not paying close enough attention to GH Copilot's output)

yvrjsharma:
This fixes #7566 as well. Thanks @abidlabs

yvrjsharma (approved):
This resolves #7566 Thanks @abidlabs !

Issue #7549, opened 2024-02-27
Dropdown with `None` initial value and custom values enabled causes JavaScript error in Gradio v4.19.1+


Paillat-dev:
### Summary
When using a dropdown component in Gradio with both the `allow_custom_value` option set to `True` and its initial value set to `None`, the interface fails to load, and a JavaScript error is displayed in the browser console. This issue has been observed starting from Gradio version 4.19.1 and persists in later versions. The expected behavior is for the dropdown to load with no initial selection and allow for custom values to be entered without causing interface failures.

### Steps to Reproduce
1. Install Gradio version 4.19.1 or later.
2. Execute the following Python script:
    ```python
    import gradio as gr

    text = gr.Dropdown(["a", "b", "c"], label="Select a letter", value=None, allow_custom_value=True)
    iface = gr.Interface(fn=lambda x: print(x), inputs=[text], outputs=[])
    iface.launch()
    ```
3. Open the provided URL in a web browser to view the Gradio interface.
4. Observe that the interface does not load, and a JavaScript error appears in the console.

### Expected Behavior
The dropdown should be displayed without any initial selection, and users should be able to enter custom values.

### Actual Behavior
The interface fails to load, and the JavaScript console displays the following error:
```
Uncaught (in promise) TypeError: input_text.toLowerCase is not a function
```

### System Info
- Gradio Version: 1.19.1 or later
- Operating System: Windows 11
- Browser: Firefox

### Screenshots
![image](https://github.com/gradio-app/gradio/assets/75439456/90b38e25-1e06-47f8-b961-52b8936c89e0)
![image](https://github.com/gradio-app/gradio/assets/75439456/cc87f9cd-4f25-4946-b198-8575b6e01d52)

### Logs
No relevant logs available.

### Severity
This issue is blocking the usage of the Gradio interface with dropdowns that allow custom values, as it prevents the interface from loading entirely.

dawoodkhan82:
Thanks for filing this issue @Paillat-dev, will take a look!

PR #7567, opened 2024-02-29
Quick fix: custom dropdown value

dawoodkhan82:
## Description

Fixes:

```
import gradio as gr

text = gr.Dropdown(["a", "b", "c"], label="Select a letter", value=None, allow_custom_value=True)
iface = gr.Interface(fn=lambda x: print(x), inputs=[text], outputs=[])
iface.launch()
```

Closes: #7549
Closes: https://github.com/gradio-app/gradio/issues/7613

## 🎯 PRs Should Target Issues

## Tests


Paillat-dev:
Btw @dawoodkhan82 That same error happens if you wanna update the dropdown choices, if you wanna take a look:
```python
import gradio as gr

text = gr.Dropdown(["a", "b", "c"], label="Select a letter", value=None, allow_custom_value=True)
iface = gr.Interface(fn=lambda x: gr.Dropdown(choices=["a", "b", "c", x]), inputs=[text], outputs=[text])
iface.launch()
```

Paillat-dev:
By the way, the following is throwing the same error, including with this pr:

```python
import gradio as gr

text = gr.Dropdown(["a", "b", "c"], label="Select a letter", value=None, allow_custom_value=True)
iface = gr.Interface(fn=lambda x: [gr.Dropdown(choices=[], value=None)], inputs=[text], outputs=[text])
iface.launch()
```

dawoodkhan82:
@Paillat-dev @abidlabs thanks for testing, taking a look at the issues.

dawoodkhan82:
```
import gradio as gr

text = gr.Dropdown(["a", "b", "c"], label="Select a letter", value=None, allow_custom_value=True)
iface = gr.Interface(fn=lambda x: gr.Dropdown(choices=[], value=None), inputs=[text], outputs=[text])
iface.launch()

```

@Paillat-dev This works fine for me by the way.

abidlabs:
@dawoodkhan82 did you try submitting the Interface? That's when the error happened for me

Paillat-dev:
> ```
> import gradio as gr
> 
> text = gr.Dropdown(["a", "b", "c"], label="Select a letter", value=None, allow_custom_value=True)
> iface = gr.Interface(fn=lambda x: gr.Dropdown(choices=[], value=None), inputs=[text], outputs=[text])
> iface.launch()
> 
> ```
> 
> @Paillat-dev This works fine for me by the way.

Yeah that's what I was saying, the error only happens when you (lambda) return a list with only one item, it being the drop-down, but not when you return it alone. Maybe it also happens when returning a one item tuple, I should try. Nonetheless, since a list works with multiple items it should also work with one.

dawoodkhan82:
@abidlabs are you testing this code?

```
import gradio as gr

text = gr.Dropdown(["a", "b", "c"], label="Select a letter", value=None, allow_custom_value=True)
iface = gr.Interface(fn=lambda x: gr.Dropdown(choices=[], value=None), inputs=[text], outputs=[text])
iface.launch()
```

https://github.com/gradio-app/gradio/assets/12725292/76d11c01-62ae-440f-bb4e-6d5e6c896d35

Works for me, when _not_ returning the dropdown as a list. I'm not sure returning a list is supposed to work

abidlabs:
Ah that's true, when `gr.Dropdown()` has `multiselect=False`, the return type should be a string, not a list. I'm not sure @Paillat-dev what you mean by this:

> returning multiple items as a list, just not with one. And especially when in the function the outputs are still specified as a list

abidlabs:
Ah good point @dawoodkhan82! 

@Paillat-dev when you have a single output component, you should *not* wrap the return in a list / tuple. You should just return the output as is. The reason we don't support singleton lists is that it becomes impossible to distinguish whether you are intending on returning a singleton list or the item within it in some scenarios. Consider this example:

```py
import gradio as gr

gr.Interface(lambda x:[{"abc": "def"}], "textbox", "json").launch()
```

In this case, both are possible: the user might be intending on returning a list, or just the dict within it. When you have multiple output components, this ambiguity is resolved.

So in that case, this PR looks good to me @dawoodkhan82 if we could just add a js unit test to prevent regressions!

abidlabs (approved):

Paillat-dev:
I did not think of that. Sounds logical now, thanks for the explanation. And thanks for the fix :)

dawoodkhan82:
> In addition to the issue @Paillat-dev pointed out, I noticed that the case of value="" is not handled correctly:
> 
> ```python
> import gradio as gr
> 
> with gr.Blocks() as demo:
>     text = gr.Dropdown(["a", "b", "c"], label="Select a letter", value="", allow_custom_value=True, interactive=True)
> 
> demo.launch()
> ```
> 
> This should initialize the Dropdown with a value of "" but instead it initializes with "a" -- I think there could be a related issue happening.
> 
> Finally, could we add js unit tests for these fixes @dawoodkhan82?

Taking a look at this again, and according to the dropdown tests. When initializing the dropdown with `value=""`, it should pre-select the first dropdown option.

abidlabs:
> Taking a look at this again, and according to the dropdown tests. When initializing the dropdown with value="", it should pre-select the first dropdown option.

Oh weird, is that the case when initializing with any value that is not one of the choices? I guess that makes sense

abidlabs:
LGTM @dawoodkhan82! Thanks for adding the test.

I've confirmed that this fixes: https://github.com/gradio-app/gradio/issues/7613 so I've added it to the parent comment. 

I'll go ahead and merge this in so that we can get it out for the release!

PR #7568, opened 2024-02-29
Add Playground to Lite Custom Element, and Convert Docs Demos to Lite

aliabd:
* Adds the `playground` attribute to the `<gradio-lite>` custom element. Which creates an interactive side-by-side code and demo playground. It hides the demo until it's loaded (but shows the loading messages on the top right). It will take the full width and height of it's container, and switch to rows from columns at the md breakpoint. 

Example: 

```html
<gradio-lite playground>
import gradio as gr
gr.Interface(
	fn=lambda x: x,
	inputs=gr.Textbox(),
	outputs=gr.Textbox()
).launch()
</gradio-lite>
```

https://github.com/gradio-app/gradio/assets/9021060/c821a5b7-c526-42b6-b511-867a179a4dd8

To test this run: 

```bash
pnpm --filter @gradio/app build:lite
python -m http.server --directory js/lite
```

This will start a server at `http://localhost:8000`. You can then edit `js/lite/index.html` to add a gradio-lite element with the playground attribute.

  * Also converts the demos on the docs to use this instead of embedding from spaces. Since the website loads lite from the cdn and these changes haven't been released (I probably should've broken up this PR - maybe I still will), the preview won't work. But locally you can just run the website as normal with `pnpm dev` and separately build and serve the lite files: `pnpm --filter @gradio/app build:lite && python -m http.server --directory js/lite` 

This is what it looks like: 

https://github.com/gradio-app/gradio/assets/9021060/d911dbca-a920-4470-b820-92b528a404ce

I removed/replaced the demos that won't work with lite.

PR still a draft because there's still one issue I'm debugging: page needs to refresh for the lite to fill the custom elements. I think the app is not getting properly destroyed when the user navigates away. 

Also still messing a bit with the styling to get it right.


abidlabs:
Closing in favor of #7660 and #7661

Issue #7362, opened 2024-02-09
Can't load some chatbot demos on huggingface spaces


freddyaboulton:
### Describe the bug

I can't load the following two chatbot demos in the gradio org. The cause is different for each once. Might be fixed by #7350 but just noting here.

1. `gr.load("gradio/chatinterface_streaming_echo", src="spaces")`
```
Fetching Space from: https://huggingface.co/spaces/gradio/chatinterface_streaming_echo
Loaded as API: https://gradio-chatinterface-streaming-echo.hf.space ✔
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[3], line 1
----> 1 gr.load("gradio/chatinterface_streaming_echo", src="spaces")

File ~/miniconda3/envs/dotenv-dev/lib/python3.10/site-packages/gradio/external.py:72, in load(name, src, hf_token, alias, **kwargs)
     48 @document()
     49 def load(
     50     name: str,
   (...)
     54     **kwargs,
     55 ) -> Blocks:
     56     """
     57     Method that constructs a Blocks from a Hugging Face repo. Can accept
     58     model repos (if src is "models") or Space repos (if src is "spaces"). The input
   (...)
     70         demo.launch()
     71     """
---> 72     return load_blocks_from_repo(
     73         name=name, src=src, hf_token=hf_token, alias=alias, **kwargs
     74     )

File ~/miniconda3/envs/dotenv-dev/lib/python3.10/site-packages/gradio/external.py:111, in load_blocks_from_repo(name, src, hf_token, alias, **kwargs)
    106         warnings.warn(
    107             """You are loading a model/Space with a different access token than the one you used to load a previous model/Space. This is not recommended, as it may cause unexpected behavior."""
    108         )
    109     Context.hf_token = hf_token
--> 111 blocks: gradio.Blocks = factory_methods[src](name, hf_token, alias, **kwargs)
    112 return blocks

File ~/miniconda3/envs/dotenv-dev/lib/python3.10/site-packages/gradio/external.py:527, in from_spaces(space_name, hf_token, alias, **kwargs)
    520 if kwargs:
    521     warnings.warn(
    522         "You cannot override parameters for this Space by passing in kwargs. "
    523         "Instead, please load the Space as a function and use it to create a "
    524         "Blocks or Interface locally. You may find this Guide helpful: "
    525         "https://gradio.app/using_blocks_like_functions/"
    526     )
[...]

Issue #7360, opened 2024-02-08
Make detection of files in `gradio_client.Client` more robust


abidlabs:
The `gradio_client.Client` accepts string filepaths which are converted to file objects and uploaded to the server if they correspond to an input component that accepts files. However, this detection is not very robust.

Specifically, if _any_ part of the input component payload can be a file, all parts of the payload are treated as potentially containing files. This leads to issues such as this:

```py
import gradio as gr

with gr.Blocks() as demo:
    chatbot = gr.Chatbot()
    msg = gr.Textbox()
    clear = gr.Button("Clear")
    st = gr.State([1, 2, 3])

    def respond(message, st, chat_history):
        assert st[0] == 1 and st[1] == 2 and st[2] == 3
        bot_message = "I love you"
        chat_history.append((message, bot_message))
        return "", chat_history

    msg.submit(respond, [msg, st, chatbot], [msg, chatbot], api_name="submit")
    clear.click(lambda: None, None, chatbot, queue=False)
    
_, url, _ = demo.launch(inline=False)

from gradio_client import Client

client = Client(url)

initial_history = [["", None]]
message = "Hello"
ret = client.predict(message, initial_history, api_name="/submit")
```

This produces an error because the `""` in the `initial_history` is treated as a potential file and is tried to be uploaded to the server. We should really validate against the API info to figure out which part of the payload could be a file and which format it needs to be in. We should only upload those files, from a security and robustness perspective. Or we need to think of something else altogether

abidlabs:
Ok so I'll work on writing a function to traverse a given value based on an arbitrary type hint and identify all parts of that value that could potentially match a FileData in the type hint. Its probably not going to be perfect, but it'll be better than what we have now. But I'm planning on making one big change. Right now, the process of detecting and uploading files happens like this:

1. The Client decides whether a particular string could potentially be a file
2. If it is, then it uploads it to the server and replaces the string with a FileData object
3. It then sends the FileData object to the prediction API

Instead, I'd like to propose we do this:

1. The Client sends the data as is to the Gradio prediction API
2. As part of `preprocess()`, we figure out whether whether a particular string should really be a FileData object. Because we have access to the input components and their type hints as actual Python Types instead of just strings, its a little bit easier
3. If indeed a particular string should be a FileData, then the Gradio server requests the client to upload the file and a FileData object is generated using the uploaded filepath

[...]

PR #7575, opened 2024-02-29
Files should now be supplied as `file(...)` in the Client, and some fixes to `gr.load()` as well

abidlabs:
This PR makes the following changes:

- [x] Adds a `file` method to `gradio_client` that all files/urls should be wrapped in
- [x] Allows the current behavior (filepath provided as a string), but in this case a warning is raised
- [x] Correctly handles files when using `gr.load()`
- [x] Updates all of the `example_payload()` functions in the components and adds a test for these
- [x] Fixes the View API page to show the correct syntax
- [x] Updates the documentation to show the new syntax

Fixes: https://github.com/gradio-app/gradio/issues/7360

Note: this PR also makes some fixes to `gr.load()` so that Gradio apps with `.then()` events can be loaded correctly. Starting to fix https://github.com/gradio-app/gradio/issues/7362, but a complete fix will come later to not blow this PR up too much


abidlabs:
I still want to do some cleanup of the upload/gather code as its getting quite messy but I can do that in parallel with a review.

js/app/src/api_docs/CodeSnippets.svelte
```diff
@@ -18,19 +27,12 @@
 	let python_code: HTMLElement;
 	let js_code: HTMLElement;
 
+	let has_file_path = endpoint_parameters.some(
+		(param: EndpointParameter) => param.python_type.type === "filepath"
```
freddyaboulton, line 31: What if instead of making this change, we wrap the data in `example_inputs` method with the File dataclass? That way it shows up for Video/File as well. Right now this check does not take those components into account (they have nested file types).

freddyaboulton (commented):
Thanks for the nice PR @abidlabs ! Left some comments.

client/python/gradio_client/client.py
```diff
@@ -989,10 +990,10 @@ def _inner(*data):
             if not self.is_valid:
                 raise utils.InvalidAPIEndpointError()
             data = self.insert_state(*data)
-            if self.client.upload_files:
-                data = self.serialize(*data)
+            data = self.serialize(*data)
             predictions = _predict(*data)
-            predictions = self.process_predictions(*predictions)
+            if self.client.download_files:
+                predictions = self.process_predictions(*predictions)
```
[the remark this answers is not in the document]
abidlabs: Will double check this

client/python/gradio_client/client.py
```diff
@@ -1129,24 +1130,47 @@ def _gather_files(self, *data):
...
+                # Check files to upload
                 # file dict is a corner case but still needed for completeness
-                # most users should be using filepaths
+                # file paths are allowed for backwards compatibility
+                # if the Client is used directly, it should be a FileData object
                 d = utils.traverse(
-                    d, get_file, lambda s: utils.is_file_obj(s) or utils.is_filepath(s)
+                    d,
+                    get_file,
+                    lambda s: utils.is_file_data(s, url_ok=False)
```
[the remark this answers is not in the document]
abidlabs: Totally agreed, will clean this up

client/python/gradio_client/data_classes.py
```diff
@@ -15,3 +17,8 @@ class FileData(TypedDict):
     orig_name: NotRequired[str]  # original filename
     mime_type: NotRequired[str]
     is_stream: NotRequired[bool]
+
+
+@dataclass
+class File:
```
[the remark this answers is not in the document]
abidlabs: I thought about that, but there wouldn't be a clash unless they've actually imported `File` like:

`from gradio import File`, which is [quite uncommon](https://huggingface.co/search/full-text?q=%22from+gradio+import+File%22). I'm open to renaming but I would prefer a shorter name. I guess the other option would be to make it a factory method instead `file()` -- thoughts?

client/python/gradio_client/client.py
```diff
@@ -1129,24 +1130,47 @@ def _gather_files(self, *data):
         def get_file(d):
             if utils.is_file_obj(d):
                 file_list.append(d["path"])
+            elif isinstance(d, File):
+                file_list.append(d.path)
             else:
+                warnings.warn(
+                    f"The Client is treating: {d} as a filepath/URL and uploading it to the remote Gradio app. "
```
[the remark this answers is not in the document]
abidlabs: I think we should keep the warning (but maybe throw it somewhere else) since URLs will also need be wrapped with `File()` (or whatever we call it) starting in 1.0

js/app/src/api_docs/CodeSnippets.svelte
abidlabs: Ah yeah good catch, will that show up correctly in the snippet if I just wrap in a File()? I'll double check

gradio/external.py
```diff
@@ -418,7 +418,7 @@ def from_spaces(
 
 
 def from_spaces_blocks(space: str, hf_token: str | None) -> Blocks:
-    client = Client(space, hf_token=hf_token, download_files=False)
+    client = Client(space, hf_token=hf_token, upload_files=False, download_files=False)
```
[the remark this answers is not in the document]
abidlabs: Nope it works, because files are automatically preprocessed and converted to a FileData by the loaded Gradio app before they are sent to the remote Space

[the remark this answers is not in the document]
freddyaboulton: I got an error with this code (no output image when uploading my own file) 

```python
gr.load("spaces/gradio/image_mod").launch()
```

Maybe my environment was messed up - can you repro?

js/app/src/api_docs/CodeSnippets.svelte
freddyaboulton: I think so but will be good to double check.

client/python/gradio_client/data_classes.py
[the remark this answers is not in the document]
freddyaboulton: I like the `file` factory method suggestion!

abidlabs:
Ok I'm going to to clean everything up and fix the issues we're seeing with `gr.load()` in this PR. Going to convert back to draft status while I do so. Thanks @freddyaboulton for this first review -- will likely ping you for another!

gradio/external.py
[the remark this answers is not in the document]
abidlabs: This should now be fixed

client/js/src/types.ts
```diff
@@ -20,7 +20,7 @@ export interface Config {
 	show_api: boolean;
 	stylesheets: string[];
 	path: string;
-	protocol?: "sse_v2" | "sse_v1" | "sse" | "ws";
+	protocol?: "sse_v2.1" | "sse_v2" | "sse_v1" | "sse" | "ws";
```
abidlabs, line 23: v2.1 changes the format of FileData responses: they now include a `meta` field to make them easier to identify

client/python/gradio_client/compatibility.py
abidlabs, line 1: I kept accidentally editing the `EndpointV3Compatibility` class in `client.py` instead of the `Endpoint` class so moved it out to its own separate file. Makes the core `client.py` a bit smaller and easier to grasp.

client/python/gradio_client/utils.py
```diff
@@ -898,13 +914,18 @@ def get_type(schema: dict):
         raise APIInfoParseError(f"Cannot parse type for {schema}")
 
 
-OLD_FILE_DATA = "Dict(path: str, url: str | None, size: int | None, orig_name: str | None, mime_type: str | None)"
-FILE_DATA = "Dict(path: str, url: str | None, size: int | None, orig_name: str | None, mime_type: str | None, is_stream: bool)"
+FILE_DATA_FORMATS = [
```
abidlabs, line 917: We can mostly remove this in 4.0 when we don't need to use the old data formats to determine if a component has any filedata in its payload

abidlabs:
Ok I did a lot of cleanup of the `client.py` file and now everything should be working! At this point, this PR only fixes: https://github.com/gradio-app/gradio/issues/7360, but it should make other issues related to `gr.load()` a lot easier to fix. Will tackle those in future PRs

freddyaboulton (commented):
Looks great @abidlabs ! Still having an issue with gr.load ( i know more fixes are coming) but thought i'd let you know cause I think it should be working given the changes in this PR.

gradio/data_classes.py
```diff
@@ -164,6 +164,7 @@ class FileData(GradioModel):
     orig_name: Optional[str] = None  # original filename
     mime_type: Optional[str] = None
     is_stream: bool = False
+    meta: dict = {"_type": "gradio.FileData"}
```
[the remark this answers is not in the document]
abidlabs: Yea originally I had it

* meta: "gradio.FileData"

but felt that it didn't signal "internal use" clearly. I also considered

* "_type": "gradio.FileData"

but pydantic ignores props that begin with "_" when dumping to a json.

So I figured this was the clearest and sufficiently "unique" that it would be unlikely to cause issues down the line

abidlabs:
> It's still not working for me
> Looking at the logs in the space, I think the file is not being uploaded

Can you share the code that you are running locally?

client/python/gradio_client/client.py
```diff
@@ -1059,68 +1063,51 @@ def _predict(*data) -> tuple:
...
+        # If self.download_file is True, we assume that that the user is using the Client directly (as opposed
+        # within gr.load) and therefore, download any files generated by the server and skip values for
+        # components that the user likely does not want to see (e.g. gr.State, gr.Tab).
+        if self.client.download_files:
+            predictions = self.download_files(*predictions)
+            predictions = self.remove_skipped_components(*predictions)
+        predictions = self.reduce_singleton_output(*predictions)
+        return predictions
+
+    def download_files(self, *data) -> tuple:
```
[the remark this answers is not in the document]
abidlabs: Hmm we could do this, but there's still a chance of false positives. Let's say that your function returns a json that is of this format: {"path": "/url/to/image"}, which is not *that* unlikely

This will fail because `is_file_obj_new(data)` will parse the json dict {"path": "/url/to/image"} as a file and try to download it.

client/python/gradio_client/client.py
```diff
@@ -1059,68 +1063,51 @@ def _predict(*data) -> tuple:
...
+            elif not self.client.upload_files:
+                d = utils.traverse(d, self._upload_file, utils.is_file_obj_with_meta)
+            data_.append(d)
+        return tuple(data_)
+
+    def process_predictions(self, *predictions):
+        # If self.download_file is True, we assume that that the user is using the Client directly (as opposed
+        # within gr.load) and therefore, download any files generated by the server and skip values for
+        # components that the user likely does not want to see (e.g. gr.State, gr.Tab).
+        if self.client.download_files:
```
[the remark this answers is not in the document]
abidlabs: Ok yeah I can have a separate internal parameter that designates that the Client is being used inside of `gr.load()`

abidlabs:
Ok `gr.load("spaces/gradio/image_mod")` should be fixed now. I also fixed the issue with the example images not showing up. Let me test a few other Spaces and I'll ping you again @freddyaboulton if everything looks good

client/python/gradio_client/client.py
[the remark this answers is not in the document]
abidlabs: Added `_skip_components` parameter

abidlabs:
I tested a bunch of simple Spaces:

* `gr.load("spaces/gradio/image_mod").launch()`
* `gr.load("spaces/gradio/video_identity").launch()`
* `gr.load("spaces/gradio/fake_gan_2_main").launch()`
* `gr.load("spaces/gradio/filter_records").launch()`
* `gr.load("spaces/gradio/diff_texts_main").launch()`

And these all work now. Some more complex demos are broken:

* `gr.load("spaces/gradio/kitchen_sink").launch()`
* `gr.load("spaces/gradio/chatinterface_streaming_echo").launch()`

I'll fix these as part of https://github.com/gradio-app/gradio/issues/7362 next.

gradio/processing_utils.py
```diff
@@ -275,6 +277,8 @@ def _move_to_cache(d: dict):
                         f"File {path} is not in the upload folder and cannot be accessed."
                     )
             temp_file_path = block.move_resource_to_block_cache(payload.path)
+            if keep_in_cache:
```
freddyaboulton, line 280: We should only be adding files to this set, right now we add aanydata (pd.DataFrame, strings etc) if `keep_in_cache` is True

gradio/components/base.py
```diff
@@ -190,9 +190,6 @@ def __init__(
         self.scale = scale
         self.min_width = min_width
         self.interactive = interactive
-        # Keep tracks of files that should not be deleted when the delete_cache parmaeter is set
```
freddyaboulton, line 193: Please add `keep_in_cache=True` to `move_files_to_cache` below

freddyaboulton (approved):
Thanks for the fixes @abidlabs ! Left two comments but otherwise LGTM!

client/python/gradio_client/client.py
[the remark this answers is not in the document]
freddyaboulton: Nice!

gradio/processing_utils.py
abidlabs: Hmm how do you figure that? `_move_to_cache` is only called if the value is a `FileData` dictionary, meaning that `temp_file_path` must be a string path.

gradio/components/base.py
abidlabs: Ah yeah thanks

gradio/processing_utils.py
abidlabs: I moved it to the end of `_move_to_cache` to be more explicit about this

freddyaboulton: Yea you're right I was mistaken! Should the predicate in `traverse` used here be `file_obj_with_meta` though?

abidlabs: Yes, I think so. Let me just make sure that doesn't break older apps

abidlabs: Actually unfortunately it does seem to break `gr.load()` with older Spaces that have examples because the format of file data in those examples doesn't have "meta". We may need to able support both. I'll see if there's a clean way to fix this as part of tackling https://github.com/gradio-app/gradio/issues/7362.

abidlabs:
Thanks @freddyaboulton everything should be addressed now! Will merge in after the CI runs

PR #7589, opened 2024-03-03
The render decorator

aliabid94:
Introducing the render decorator. This PR has the simple functionality of re-rendering a set of code based on some triggers and inputs. Take a look at the example below:

```python
import gradio as gr

with gr.Blocks() as demo:
    input_text = gr.Textbox(label="input")
    split_btn = gr.Button("Split")

    @gr.render(inputs=input_text, triggers=[split_btn.click])
    def show_split(text):
        if len(text) == 0:
            gr.Markdown("## No Input Provided")
        else:
            for letter in text:
                gr.Textbox(letter)

if __name__ == "__main__":
    demo.launch()
```

which generates:
Recording 2024-03-04 at 03 53 13

So far, this only supports the ability of the render being the result of some other components. The render block cannot be used as input, or have any event listeners inside just yet.


abidlabs:
@aliabid94 this is really cool and nice leveraging of existing Blocks API to implement this. I noticed a few small issues:

1) The View API page shows gr.render events but with the incorrect signature. E.g. in your example I see this:

image

I would just set the `api_name=False` for these events so they don't show up in the view API page

2. We should document this in the Blocks Guides! As well as in the docs. We should also update this section: https://www.gradio.app/guides/controlling-layout#variable-number-of-outputs

3. For some reason, file-type components don't seem to work inside a gr.render function. I tried this:

```py
import gradio as gr

with gr.Blocks() as demo:
    input_text = gr.Number(label="input")
    split_btn = gr.Button("Split")

    @gr.render(inputs=input_text, triggers=[split_btn.click, input_text.change])
    def show_split(n):
        if not n:
            gr.Markdown("## No Input Provided")
        else:
            for letter in range(n):
                gr.Image()

if __name__ == "__main__":
    demo.launch()
```

aliabid94:
> 1. The View API page shows gr.render events but with the incorrect signature. E.g. in your example I see this:

Fixed.

> 2. We should document this in the Blocks Guides

Will do all the documenting once all the pieces for render are in, the API might change as I add the different pieces

> 3. For some reason, file-type components don't seem to work inside a gr.render function

Fixed, this was an issue of how add_root_url works. It naively recurses through the output json and looks for any key called "path" and tries to modify it. I made a modification so that render would work, but the logic of add_root_url needs to be fixed properly at some point.

aliabid94:
@pngwn could I get a quick look on the frontend of this as well? I don't know if the approach of importing @gradio/app from @gradio/renderable is correct, as it causes a circular import. It's also casuing a fail in the typecheck

pngwn:
This is very cool!

Regarding the frontend architecture. Why don't we update the existing config (components, dependencies, and layout) to represent the new structure? I think this would be preferable to having a new way to express what is essentially the same thing (a tree of components and their relationship). It keeps things simpler architecturally as we are simply reusing the existing rendering architecture. We will need to change a few things to make sure the confing _can_ be updated and rerender without any side effects but that shouldn't be too much work.

We don't really want to create a new instance of `Blocks` that is really designed to be a singleton and we haven't given any thought to what might happen if there is more than one instance of it (especially a nested instance) we do things like setting context inside blocks which has implications for the whole tree, having another instance doing the same thing could cause issues with certain components. There are also performance implications with creating a new Blocks instance. I think we'd need to do some more refactoring of Blocks to be certain that we wouldn't cause issues (although #7564 is a good first step in that direction).

I though there was a PR sometime ago that would allow the config to be updatable? Or did i imagine that? I think I remember us having a conversation pre-4.0 that we wanted to take that approach instead.

pngwn:
Regarding the type issues, looks like an issue with it not resolving the export properly, can look into it if we go this route. We would also need to type that 'value' property fully (that error isn't showing up yet but I'm guessing it will if we fix the other issue).

abidlabs:
@pngwn I can't comment on the full frontend implications of this PR, but I would disagree on this point:

> We don't really want to create a new instance of Blocks that is really designed to be a singleton and we haven't given any thought to what might happen if there is more than one instance of it (especially a nested instance) we do things like setting context inside blocks which has implications for the whole tree

In the backend, we fully support nesting Blocks inside other Blocks, we've implemented logic to support this use case, e.g. https://github.com/gradio-app/gradio/blob/19e4e04b9e4906a82c1d9b1eb2e29243dadb386d/gradio/blocks.py#L1023

And there are lots of demos that do this, often utilizing multiple `gr.load()`'s (each of which create a `gr.Blocks`) inside a parent `gr.Blocks`

Just wanted to mention that in case it affects the discussion on this PR.

abidlabs:
Couple of minor points as I'm testing this out:

1. `triggers` should be able to take in a single trigger, shouldn't need to be wrapped in a list, to be consistent with how we handle similar parameters in events. E.g. this should work:

```py
        @gr.render(inputs=[], triggers=t.submit)
```

2. It seems like @gr.render automatically places internal elements inside a gr.Column instead of respecting the parent BlocksContext, e.g.:

```py
import gradio as gr
import random

with gr.Blocks() as demo:
    t = gr.Textbox()

    with gr.Row():    
        @gr.render(inputs=[], triggers=[t.submit])
        def generate():
            for _ in range(random.randint(3, 8)):
                gr.Image()
                
demo.launch()
```

abidlabs:
Also, I think uploading isn't working in "rendered" components. E.g. try uploading an audio file in this demo:

```py
import gradio as gr
import random

with gr.Blocks() as demo:
    t = gr.Textbox()

    @gr.render(inputs=[], triggers=[t.submit])
    def generate():
        with gr.Row():    
            for _ in range(random.randint(1, 4)):
                gr.Audio()
                
demo.launch()
```

abidlabs:
One question: is there a way to get acess on the "gr.render" event? Say I want to chain it, or cancel it

pngwn:
> In the backend, we fully support nesting Blocks inside other Blocks, we've implemented logic to support this use case, e.g.

Yeah sorry, to clarify i meant `Blocks.svelte`. Blocks as a concept are a backend only, we simply deal a virtual component data structure in the frontend rather than individual block instances.

PR #7598, opened 2024-03-04
Prevent additional paths that can trigger credential leakage on Windows

abidlabs:


freddyaboulton (approved):
Looks good @abidlabs ! Thanks for the fix. I think this should be a patch instead of a minor though.

abidlabs:
Thanks @freddyaboulton! yup

Issue #7592, opened 2024-03-04
Audio is sped up when trimmed (possible related to #6336)


linoytsaban:
### Describe the bug

It seems the audio is sped up the you trim it, for example: 

https://github.com/gradio-app/gradio/assets/57615435/742c05e1-626d-4ad3-829f-053d10c3172a

possibly related to #6336

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

gr.Interface(fn=lambda x: x, inputs='audio', outputs='audio').queue().launch()

```

### System Info

```shell
gradio==4.19.1
```

### Severity

I can work around it

JackismyShephard:
Could also be related to https://github.com/gradio-app/gradio/issues/7284

PR #7599, opened 2024-03-04
Prevent audio speeding up when trimming

### LINKED ISSUE #6336: Audio get louder when trimmed
### Describe the bug

It seems the audio volume goes up every time you crop it as you can hear in the following video:

https://github.com/gradio-app/gradio/assets/25161192/fdcacb28-c5a6-47e6-bfe6-c516d2385c70

### Reproduction

```py
import gradio as gr

gr.Interface(fn=lambda x: x, inputs='audio', outputs='audio').queue().launch()
```

### Screenshot

_No response_

### Logs

_No response_

### System Info

```shell
gradio==4.1.2
```

### Severity

I can work around it

hannahblair:
## Description

This PR changes the way we convert the audio buffer we get from the audio component to WAV format. The function should now interleave the audio samples correctly so the audio data matches what the audio component expects for stereo or multi-channel WAV files. 

This also fixes the background of the trim region which got lost at some point (if the option isn't provided as a param)

Note, I think this should close #6336 but I can't be sure as I can't repro it

Tested with:

- .ogg
- .mp3
- .wav
- .mp4
- .webp

Closes: #7592

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/great-worms-sip.md
+++ b/.changeset/great-worms-sip.md
@@ -1,0 +1,5 @@
+---
+"@gradio/preview": patch
+---
+
+feat:chore(deps): update dependency stylus to ^0.63.0
--- a/.changeset/icy-teams-act.md
+++ b/.changeset/icy-teams-act.md
@@ -1,0 +1,6 @@
+---
+"@gradio/wasm": minor
+"gradio": minor
+---
+
+feat:Upgrade Pyodide to 0.25.0
--- a/.changeset/weak-windows-itch.md
+++ b/.changeset/weak-windows-itch.md
@@ -1,0 +1,5 @@
+---
+"gradio": patch
+---
+
+feat:Prevent additional paths that can trigger credential leakage on Windows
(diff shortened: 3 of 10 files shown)


abidlabs (approved):
LGTM @hannahblair nice fix!

hysts:
> @hysts I haven't been able to repro the volume getting ticket in your issue https://github.com/gradio-app/gradio/issues/6336 - would you mind giving this PR a test and seeing if it persists for the files you were noticing this on?

@hannahblair Yes, it seems to be fixed by this PR! Thanks!
(FYI, in my environment, #6336 reproduces even with `gradio==4.19.2`, so I wonder why you couldn't reproduce it. Maybe it depends on audio files? I used [these audio files](https://huggingface.co/spaces/facebook/seamless-m4t-v2-large/tree/main/assets) for testing this time. Anyway, I guess it doesn't matter as it's now fixed.)

PR #7612, opened 2024-03-05
Add client wheel url to main spaces 

aliabd:
Adds the missing client wheel url to the requirements for main website spaces.

### DIFF
--- a/scripts/upload_website_demos.py
+++ b/scripts/upload_website_demos.py
@@ -27,7 +27,8 @@ def upload_demo_to_space(
     space_id: str, 
     hf_token: str, 
     gradio_version: str | None, 
-    gradio_wheel_url: str | None = None
+    gradio_wheel_url: str | None = None,
+    gradio_client_url: str | None = None
 ):
     """Upload a demo in the demo directory to a huggingface space.
     Parameters:
@@ -36,6 +37,7 @@ def upload_demo_to_space(
         hf_token: HF api token. Need to have permission to write to space_id for this to work.
         gradio_version: If not None, will set the gradio version in the created Space to the given version.
         gradio_wheel_url: If not None, will install the version of gradio using the wheel url in the created Space.
+        gradio_client_url: If not None, will install the version of gradio client using the wheel url in the created Space.
     """
 
     with tempfile.TemporaryDirectory() as tmpdir:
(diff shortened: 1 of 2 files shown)


scripts/upload_website_demos.py
```diff
@@ -93,13 +95,15 @@ def upload_demo_to_space(
...
     args = parser.parse_args()
     gradio_wheel_url = args.WHEEL_URL + f"gradio-{args.GRADIO_VERSION}-py3-none-any.whl"
+    
     if args.AUTH_TOKEN is not None:
         hello_world_version = str(huggingface_hub.space_info("gradio/hello_world").cardData["sdk_version"])
         for demo in demos:
             if hello_world_version != args.GRADIO_VERSION:
                 upload_demo_to_space(demo_name=demo, space_id="gradio/" + demo, hf_token=args.AUTH_TOKEN, gradio_version=args.GRADIO_VERSION)
-            upload_demo_to_space(demo_name=demo, space_id="gradio/" + demo + "_main", hf_token=args.AUTH_TOKEN, gradio_version=args.GRADIO_VERSION, gradio_wheel_url=gradio_wheel_url)
+            upload_demo_to_space(demo_name=demo, space_id="gradio/" + demo + "_main", hf_token=args.AUTH_TOKEN, gradio_version=args.GRADIO_VERSION, gradio_wheel_url=gradio_wheel_url, gradio_client_url=args.CLIENT_URL)
```
abidlabs, line 109: We modify args.WHEEL_URL by adding "gradio-{args.GRADIO_VERSION}-py3-none-any.whl" to it. Do we need to do the same thing for the `gradio_client`?

abidlabs:
Just one question above^ @aliabd but otherwise lgtm

scripts/upload_website_demos.py
aliabd: Nope, the gradio_client is just installed from github, so there's no actual wheel file. it's just something like 
`
gradio-client @ git+https://github.com/gradio-app/gradio@c0d25fc1df60987793471b4745ab453bea92dcfe#subdirectory=client/python
`
that's also what we do for the pr demos

abidlabs: Ok great! Thanks for clarifying

PR #7614, opened 2024-03-05
Fix the `root_url` logic for streaming files

abidlabs:
Moves the `add_root_url` logic to happen before the stream diffing. This solves the issue we had seen with `fake_diffusion_with_gif` not working on Spaces under a custom root path

Confirm by testing: https://gradio-pr-deploys-pr-7614-all-demos.hf.space/demo/fake_diffusion_with_gif/

Will add a test and fix the current tests, but should otherwise be good for a review


demo/fake_diffusion_with_gif/image.gif
abidlabs, line 1: Unrelated, but I gitignored and deleted this file

test/test_utils.py
```diff
@@ -439,3 +440,21 @@ def test_is_in_or_equal():
 )
 def test_get_extension_from_file_path_or_url(path_or_url, extension):
     assert get_extension_from_file_path_or_url(path_or_url) == extension
+
+
+@pytest.mark.parametrize(
```
abidlabs, line 445: Unrelated, but added some quick tests for the `diff` function

gradio/blocks.py
```diff
@@ -1673,6 +1675,8 @@ async def process_api(
                 state,
                 limiter=self.limiter,
             )
+            if root_path is not None:
```
freddyaboulton, line 1678: This also needs to be applied to the batch case, no?

freddyaboulton (approved):
Looks good to me @abidlabs ! Left a comment about batch mode

gradio/blocks.py
abidlabs: Batch mode is a bit tricky, because we have to add the root_urls to each element, which in theory can be different (though unlikely). For now, I'll just assume the root_urls are the same.

PR #7617, opened 2024-03-06
fix flaky python tests

abidlabs:
hotfixes for two tests

### DIFF
--- a/client/python/test/test_utils.py
+++ b/client/python/test/test_utils.py
@@ -77,7 +77,9 @@ def test_download_private_file(gradio_temp_dir):
         "https://gradio-tests-not-actually-private-spacev4-sse.hf.space/file=lion.jpg"
     )
     file = utils.download_file(
-        url_path=url_path, headers={"Authorization": f"Bearer {HF_TOKEN}"}, save_dir=str(gradio_temp_dir)
+        url_path=url_path,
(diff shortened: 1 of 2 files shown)


client/python/test/test_utils.py
```diff
@@ -77,7 +77,7 @@ def test_download_private_file(gradio_temp_dir):
         "https://gradio-tests-not-actually-private-spacev4-sse.hf.space/file=lion.jpg"
     )
     file = utils.download_file(
-        url_path=url_path, hf_token=HF_TOKEN, dir=str(gradio_temp_dir)
+        url_path=url_path, headers={"Authorization": f"Bearer {HF_TOKEN}"}, save_dir=str(gradio_temp_dir)
```
abidlabs, line 80: Because the API has changed

test/test_external.py
```diff
@@ -252,11 +252,11 @@ def test_conversational_in_blocks(self):
         client = TestClient(app)
         response = client.post(
             "/api/predict/",
-            json={"session_hash": "foo", "data": ["Hi!", None], "fn_index": 0},
+            json={"session_hash": "foo", "data": ["Hi!"], "fn_index": 0},
```
abidlabs, line 255: Looks like the classification of this model has changed, now its no longer a chatbot

PR #7618, opened 2024-03-06
Control which files get moved to cache with gr.set_static_paths

freddyaboulton:
## Description

Closes: #5350

This PR adds `gr.set_static_paths`, which takes a list of directory or filenames represented as strings or `pathlib.Path` objects. Files in `gr.set_static_paths` are not moved to the cache if they are encountered during pre/post-process or examples. 

### Example Usage
Set the GRADIO_TEMP_DIR env variable to manually check nothings gets saved during the cache as the interface launches. In this particular example, output files will be saved to the cache.

```python
import gradio as gr

gr.StaticFiles(paths=["test/test_files/cheetah1.jpg", "test/test_files/bus.png"])

demo = gr.Interface(
    lambda s: s.rotate(45),
    gr.Image(value="test/test_files/cheetah1.jpg", type="pil"),
    gr.Image(),
    examples=["test/test_files/bus.png"],
)

demo.launch()
```

Alternative APIs I considered before choosing `gr.StaticFiles`
* Wrap the value you don't want to move into the cache with a class like `gr.StaticFile`. Similar to (#7575). This adds another possible typehint to the init method which is already complicated enough. Also need to make sure to appropriate wrap/unwrap this type everywhere which is not ideal.
* Add a `static_files` parameter to `gr.Blocks`. This was what I originally thought of doing but won't handle the case of components being instantiated with a value outside of a blocks scope, like in `gr.Interface`.

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/gradio/data_classes.py
+++ b/gradio/data_classes.py
@@ -221,7 +221,7 @@ class _StaticFiles:
 
     def __init__(self, paths: list[str | pathlib.Path]) -> None:
         self.paths = paths
-        self.all_paths.extend([pathlib.Path(p).resolve() for p in paths])
+        self.all_paths = [pathlib.Path(p).resolve() for p in paths]
 
     @classmethod
     def clear(cls):
--- a/gradio/data_classes.py
+++ b/gradio/data_classes.py
@@ -210,3 +210,11 @@ class ListFiles(GradioRootModel):
 
     def __iter__(self):
         return iter(self.root)
+
+
+class StaticFiles:
+    all_paths = []
+
+    def __init__(self, paths: list[str | pathlib.Path]) -> None:
+        self.paths = paths
+        self.all_paths.extend([pathlib.Path(p).resolve() for p in paths])
--- a/gradio/data_classes.py
+++ b/gradio/data_classes.py
@@ -10,7 +10,6 @@ from enum import Enum, auto
 from typing import TYPE_CHECKING, Any, List, Optional, Union
 
 from fastapi import Request
-from gradio_client.documentation import document
 from gradio_client.utils import traverse
 
 from . import wasm_utils
@@ -213,32 +212,9 @@ class ListFiles(GradioRootModel):
         return iter(self.root)
 
 
-@document()
-class StaticFiles:
+class _StaticFiles:
     """
-    Static files to be served by the gradio app.
-
-    Static files are not moved to the gradio cache and are served directly from the file system.
-    This class is useful when you want to serve files that you know will not be modified during the lifetime of the gradio app (like examples).
-    By using StaticFilesm your app will launch faster and it will consume less disk space.
-
-    Example:
-        import gradio as gr
-
-        # Paths can be a list of strings or pathlib.Path objects
-        # corresponding to filenames or directories.
-        gr.StaticFiles(paths=["test/test_files/"])
-
-        # The example files and the default value of the input
-        # will not be copied to the gradio cache and will be served directly.
-        demo = gr.Interface(
-            lambda s: s.rotate(45),
-            gr.Image(value="test/test_files/cheetah1.jpg", type="pil"),
-            gr.Image(),
-            examples=["test/test_files/bus.png"],
-        )
-
-        demo.launch()
+    Class to hold all static files for an app
     """
 
     all_paths = []
@@ -246,3 +222,7 @@ class StaticFiles:
     def __init__(self, paths: list[str | pathlib.Path]) -> None:
         self.paths = paths
         self.all_paths.extend([pathlib.Path(p).resolve() for p in paths])
+
+    @classmethod
+    def clear(cls):
+        cls.all_paths = []
--- a/gradio/data_classes.py
+++ b/gradio/data_classes.py
@@ -10,6 +10,7 @@ from enum import Enum, auto
 from typing import TYPE_CHECKING, Any, List, Optional, Union
 
 from fastapi import Request
+from gradio_client.documentation import document
 from gradio_client.utils import traverse
 
 from . import wasm_utils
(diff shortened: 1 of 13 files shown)


abidlabs:
Hmm instead of introducing additional API could we try the approach of determining this automatically based on whether the files are ever used as an input?

freddyaboulton:
> automatically based on whether the files are ever used as an input

I don't think this will cover the case of gr.Examples or the case where you don't want the default value of the component to be copied over but the component is still used in as input (example code in the PR body) . Also in order to determine if a component is ever used as an input we'd have to wait for all of the events to be defined. At that point the files are already copied to the cache and it's possible you OOM beforehand (rare but possible). 

> I think exposing an API for file management is pretty confusing, and could lead to unintended side effects if done incorrectly.

Can you elaborate?

abidlabs:
> Can you elaborate?

It would allow people to overwrite files that they shouldn't be overwriting. For example, say you have this demo, and you don't want to copy over `video.mp4`, so you make it a static file:

```py
import gradio as gr

gr.StaticFiles(paths=["video.mp4"])

def trim_video(input_file):
    command = [
        'ffmpeg',
        '-i', input_file,     
        '-ss', "00:00:10",    
        '-t', "00:00:20",       
        '-c', 'copy',         
        'video.mp4'           
    ]
    subprocess.run(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
    

demo = gr.Interface(
    trim_video,
    gr.Video()
    gr.Video(),
    examples=["video.mp4"],
)

demo.launch()
```

When this demo is used, because `video.mp4` is also the name of the output file of your function, it wouldn't get moved to cache, and IIUC it would overwrite the example video. Lmk if I'm misunderstanding!

> I don't think this will cover the case of gr.Examples or the case where you don't want the default value of the component to be copied over but the component is still used in as input (example code in the PR body)

Good point, hmm

abidlabs:
What if we instead added a parameter, `copy_files=True` to `gr.Examples` and to every `Component`? Basically it would control whether files generated by that component were copied over to the cache or whether they remained as is (with a warning telling users that reusing files can be problematic if the same file is being modified by the function across runs)

freddyaboulton:
I tried the ffmpeg example but it wouldn't work because ffmpeg can't edit files in place. You're right that it's possible to overwrite an input file but I think you can already do that outside this PR. You can manually save an output file to the same location as the input file. The next time that file is used as input it won't be re-written to the cache so the next run of the function will use the output from the previous run. We haven't heard of this happening in the wild so I think we're ok to not worry about it.

> For the OP in https://github.com/gradio-app/gradio/issues/5350, I think it would be easier than specifying 115 files or even doing a glob
> My other hesitation with this API is that it requires you to get the filenames exactly right in different places

You can pass in directory names to `StaticFiles` so you don't have to specify file names exactly. I don't think the typo risk is greater than with `allowed_paths`, `blocked_paths`.

Interesting point about copy_files. My one hesitation is that it will *always* not copy over the data to the cache. I'm thinking of an example where you give lots of default data to a `gr.Gallery`. It may be fine for the default value to not be copied over to the cache but for user-submitted input/output you may want to copy over to the cache.

About to log off - let's synch tomorrow?

abidlabs:
> Interesting point about copy_files. My one hesitation is that it will always not copy over the data to the cache. I'm thinking of an example where you give lots of default data to a gr.Gallery. It may be fine for the default value to not be copied over to the cache but for user-submitted input/output you may want to copy over to the cache.

That's true.  Another idea, now that you mention `allowed_paths` and `blocked_paths`, what about adding `static_paths` as a parameter in `launch()`? These paths would not get copied over to the cache.

(Sorry to keep suggesting different things, but I think this is more inline with our existing API)

freddyaboulton:
All good @abidlabs ! Yea I thought about that too and was what I was originally going to implement. But two points related to that approach that made me not implement it:

* The parameter should be in the init as opposed to launch because launch happens after all the components are defined and they write to the cache. 
* it won't handle the case of components defined outside of a block scope, like the inputs parameter to a gr.Interface.

abidlabs:
After sleeping on it, I agree @freddyaboulton with your proposed approach. All things considered, it'll probably be the most flexible approach and let us manage static files on an app-level as opposed to a component level like the `copy_files` parameter, which seems more appropriate. 

Nit: I would consider renaming `gr.StaticFiles` to `gr.set_static_paths()`. The former seems like a component to me, whereas the latter indicates to me that we are setting a property that will apply for the rest of the session.

gradio/processing_utils.py
```diff
@@ -262,6 +262,8 @@ def _move_to_cache(d: dict):
         # This makes it so that the URL is not downloaded and speeds up event processing
         if payload.url and postprocess and client_utils.is_http_url_like(payload.url):
             payload.path = payload.url
+        elif utils.is_static_file(payload):
+            payload.path = payload.path
```
abidlabs, line 266:
```suggestion
            pass
```

gradio/utils.py
```diff
@@ -958,6 +960,62 @@ def is_in_or_equal(path_1: str | Path, path_2: str | Path):
     return True
 
 
+@document()
+def set_static_paths(paths: list[str | Path]) -> None:
+    """
+    Set the static paths to be served by the gradio app.
```
abidlabs, line 966: From what I understand (and testing the code), this method sets the static_paths for all Gradio apps defined until that Python session ends (or `set_static_paths` is called again(. For example, if you have multiple Gradio apps in one jupyter notebook, this will affect all of them. We should explicitly describe this behavior here imo.

gradio/data_classes.py
```diff
@@ -210,3 +210,19 @@ def __getitem__(self, index):
...
+class _StaticFiles:
+    """
+    Class to hold all static files for an app
+    """
+
+    all_paths = []
+
+    def __init__(self, paths: list[str | pathlib.Path]) -> None:
+        self.paths = paths
+        self.all_paths.extend([pathlib.Path(p).resolve() for p in paths])
```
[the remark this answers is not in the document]
freddyaboulton: Yea good point! The reason I decided to do it this way is in the case you're nesting various blocks to make a single demo. I think we can have it replace the previous paths and add an api to extend previous paths later based on user request.

gradio/utils.py
freddyaboulton: Yea good call!

gradio/processing_utils.py
freddyaboulton: Done!

freddyaboulton:
Thanks @abidlabs ! Went ahead and made those changes

gradio/utils.py
```diff
@@ -958,6 +960,64 @@ def is_in_or_equal(path_1: str | Path, path_2: str | Path):
...
+    from gradio.data_classes import _StaticFiles
+
+    return _is_static_file(file_path, _StaticFiles.all_paths)
+
+
+def _is_static_file(file_path: Any, static_files: list[Path]) -> bool:
+    """
+    Returns True if the file is a static file (i.e. is is in the static files list).
+    """
+    if not isinstance(file_path, (str, Path, FileData)):
```
abidlabs, line 1010: Unless I missed something, `file_path` is always a `FileData` in all invocations of `_is_static_file()` so the other cases are not necessary.

abidlabs (approved):
LGTM 🔥

PR #7620, opened 2024-03-06
Refactor `example_inputs()`, separating its logic into two separate methods: `example_payload()` and `example_value()`

abidlabs:
As discussed [internally](https://huggingface.slack.com/archives/C02SPHC1KD1/p1709683107328489), our implementation of `example_inputs()` was faulty, because we were relying on it to do two separate things:

1. Generating the docs on the view API page --> i.e. the input to .preprocess()
1. As an example value when generating the placeholder app for custom component --> i.e. the input to .postprocess()

The reason we've gotten away with this so far is that most components can accept the same format for the input and output (e.g. Textbox, Dropdown, etc.). For file-based components, we were previously doing some preprocessing in the client to convert string paths to FileData dictionaries, so this was effectively true for them as well.

But this is NOT true generally. For example, Video expects a dict for .preprocess() and a str/tuple for .postprocess(). Similarly for Label, etc. In some cases, we were implementing the logic for `example_payload()` and in other cases for `example_value()`

This PR fixes that by refactoring `example_inputs()`, separating its logic into two separate methods: `example_payload()` and `example_value()`

Repro of issue: `gradio cc create --template video video2`, uncommenting the example line and running the demo. It will crash

Still need to:

- [ ]  Add tests -- I'm thinking of automatically testing all components to ensure that `component.preprocess(component.example_payload()` and `component.postprocess(component.example_value()` are valid
- [ ] Update the custom components docs to mention these two methods

But should be otherwise ready for review

### DIFF
--- a/gradio/components/annotated_image.py
+++ b/gradio/components/annotated_image.py
@@ -103,7 +103,7 @@ class AnnotatedImage(Component):
     ) -> tuple[str, list[tuple[str, str]]] | None:
         """
         Parameters:
-            payload: Tuple of base image and list of annotations.
+            payload: Dict of base image and list of annotations.
         Returns:
             Passes its value as a `tuple` consisting of a `str` filepath to a base image and `list` of annotations. Each annotation itself is `tuple` of a mask (as a `str` filepath to image) and a `str` label.
         """
@@ -198,5 +198,8 @@ class AnnotatedImage(Component):
             annotations=sections,
         )
 
-    def example_inputs(self) -> Any:
-        return {}
+    def example_payload(self) -> Any:
+        return {"image": "https://raw.githubusercontent.com/gradio-app/gradio/main/test/test_files/bus.png", "annotations": []}
+
+    def example_value(self) -> Any:
+        return ("https://raw.githubusercontent.com/gradio-app/gradio/main/test/test_files/bus.png", [([0, 0, 100, 100], "bus")])
--- a/gradio/components/annotated_image.py
+++ b/gradio/components/annotated_image.py
@@ -199,7 +199,13 @@ class AnnotatedImage(Component):
         )
 
     def example_payload(self) -> Any:
-        return {"image": "https://raw.githubusercontent.com/gradio-app/gradio/main/test/test_files/bus.png", "annotations": []}
+        return {
+            "image": "https://raw.githubusercontent.com/gradio-app/gradio/main/test/test_files/bus.png",
+            "annotations": [],
+        }
 
     def example_value(self) -> Any:
-        return ("https://raw.githubusercontent.com/gradio-app/gradio/main/test/test_files/bus.png", [([0, 0, 100, 100], "bus")])
+        return (
+            "https://raw.githubusercontent.com/gradio-app/gradio/main/test/test_files/bus.png",
+            [([0, 0, 100, 100], "bus")],
+        )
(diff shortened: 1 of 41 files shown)


abidlabs:
This is a prereq for: https://github.com/gradio-app/gradio/pull/7575

gradio/components/file.py
```diff
@@ -139,15 +141,36 @@ def preprocess(
...
+            )
+            return downloaded_file
+        else:
+            return value
+
     def postprocess(self, value: str | list[str] | None) -> ListFiles | FileData | None:
         """
         Parameters:
-            value: Expects a `str` filepath, or a `list[str]` of filepaths.
+            value: Expects a `str` filepath or URL, or a `list[str]` of filepaths/URLs.
```
freddyaboulton, line 167: We should update the `__init__` docstring to say that the value can be a URL now. Neat!

guides/05_custom-components/04_backend.md
```diff
@@ -81,23 +74,34 @@ You do **not** need to implement this yourself if you components specifies a `da
...
+    """
+    pass
+```
+
+### `example_value`
 
-The example inputs for this component displayed in the `View API` page. 
-Must be JSON-serializable.
-If your component expects a file, it is best to use a publicly accessible URL.
+An example value for your component, e.g. something that can be passed into the `.postprocess()` method
```
freddyaboulton, line 100: nit - mention that it's used in the default app used in dev mode?

guides/05_custom-components/06_frequently-asked-questions.md
```diff
@@ -22,7 +22,7 @@ If you would like to share your component with the gradio community, it is recom
 
 ## What methods are mandatory for implementing a custom component in Gradio?
 
-You must implement the `preprocess`, `postprocess`, `api_info`, `example_inputs`, `flag`, and `read_from_flag` methods. Read more in the [backend guide](./backend).
+You must implement the `preprocess`, `postprocess`, `example_payload`, and `example_value` methods. If yourer component does not use a data model, you must also define the `api_info`, `flag`, and `read_from_flag` methods. Read more in the [backend guide](./backend).
```
freddyaboulton, line 25: nit - typo `yourer`

gradio/components/file.py
```diff
@@ -139,15 +141,36 @@ def preprocess(
             return [self._process_single_file(f) for f in payload]  # type: ignore
         return [self._process_single_file(payload)]  # type: ignore
 
+    def _download_files(self, value: str | list[str]) -> str | list[str]:
```
freddyaboulton, line 144: Was debating whether this actually needed to be in the component since `gr.Blocks` can download files for you in the postprocess. But we need this here for the size of the file to be in the payload. LGTM.

freddyaboulton (approved):
Nice @abidlabs ! Looks good to me.

freddyaboulton:
Love the test BTW!

gradio/components/annotated_image.py
```diff
@@ -131,6 +132,10 @@ def postprocess(
             return None
         base_img = value[0]
         if isinstance(base_img, str):
+            if client_utils.is_http_url_like(base_img):
```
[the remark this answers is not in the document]
abidlabs: Yup in the sense that `annotated_image` didn't work with image URLs previously

gradio/components/file.py
abidlabs: Yes exactly!

abidlabs:
Thanks @freddyaboulton, will fix up the issues you found. Thanks for the great review

Issue #7569, opened 2024-02-29
`gr.Markdown` is not updated properly when it has an image tag


hysts:
### Describe the bug

The following works fine,
```py
import gradio as gr


def get_img_tag():
    return 'a'


def get_img_tag2():
    return 'b'


with gr.Blocks() as demo:
    md = gr.Markdown(value=get_img_tag())
    btn = gr.Button()
    btn.click(fn=get_img_tag2, outputs=md)
demo.queue().launch()
```
but the following doesn't.
```python
import gradio as gr


def get_img_tag():
    return '<img src="https://raw.githubusercontent.com/williamyang1991/DualStyleGAN/main/doc_images/cartoon_overview.jpg">'


def get_img_tag2():
    return '<img src="https://raw.githubusercontent.com/williamyang1991/DualStyleGAN/main/doc_images/Reconstruction_arcane_overview.jpg">'


with gr.Blocks() as demo:
    md = gr.Markdown(value=get_img_tag())
    btn = gr.Button()
    btn.click(fn=get_img_tag2, outputs=md)
demo.queue().launch()
```

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

[...]

PR #7623, opened 2024-03-06
Fixes: gr.Markdown is not updated properly when it has an image tag

dawoodkhan82:
## Description

Trying to `render_math_in_element` was causing errors, because the kaTex seemed to be trying to access elements in the DOM that weren't loaded yet. Added a check that only runs the `render_math_in_element` when `latex_delimiters` are found in the markdown `value`.

Closes: #7569

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/plenty-paths-sing.md
+++ b/.changeset/plenty-paths-sing.md
@@ -1,0 +1,6 @@
+---
+"@gradio/markdown": patch
+"gradio": patch
+---
+
+fix:Fixes: gr.Markdown is not updated properly when it has an image tag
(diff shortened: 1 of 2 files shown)


aliabid94 (approved):
LGTM! Didn't understand why the original issue was happening if you can elaborate real quick.

abidlabs:
Thanks @aliabid94 and @dawoodkhan82. Will merge this in so that we can get it out for the release

dawoodkhan82:
@aliabid94 `render_math_in_element` would error out when parsing certain html. So now we only call `render_math_in_element` when the latex delimiters are found.

Issue #7520, opened 2024-02-23
There are regression for root_path after 4.19.0


futurestandard:
### Describe the bug

it seems root_path setting is ignored.
4.18.0 is fine, but 4.19.0/1/2 have broken display

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr
import time
import numpy as np
import cv2

def saveimg(im):
    #print(im["layers"][0])
    background=im["background"]
    composite=im["composite"]
    background=cv2.cvtColor(background, cv2.COLOR_BGR2RGB )
    composite=cv2.cvtColor(composite, cv2.COLOR_BGR2RGB )
    cv2.imwrite("layers0.png",im["layers"][0])
    cv2.imwrite("background.png",background)
    cv2.imwrite("composite.png",composite)
    return gr.ImageMask(value=im,height=600), background

def loadimg(im):
    layers0=cv2.imread("layers0.png",-1)
    background=cv2.imread("background.png")
    composite=cv2.imread("composite.png")
    #layers0=cv2.cvtColor(layers0, cv2.COLOR_RGB2BGR, cv2.IMREAD_UNCHANGED )
    background=cv2.cvtColor(background, cv2.COLOR_RGB2BGR )
    composite=cv2.cvtColor(composite, cv2.COLOR_RGB2BGR )
    
    im["layers"][0]=layers0
    im["background"]=background
    im["composite"]=composite
    return gr.ImageMask(value=im,height=600), background

def upim(im_in,im):
    print(type(im_in))
    height, width, channels = im_in.shape[:3]
    im["background"]=im_in
    im["layers"][0]=np.zeros((height, width, 4), np.uint8)
    #im["layers"][1]=None

    print(f" shape:{im_in.shape}, ndim:{im_in.ndim}, size:{im_in.size}")
    return gr.ImageMask(value=im,height=600)
def getmask(im):
    im.save("i.png")
    #dst = Image.new('RGBA', (im.width + o_img.width, im.height))
    
with gr.Blocks() as demo:
    with gr.Tab("カウント"):
        with gr.Row():
            with gr.Column():
                im_in = gr.Image(sources="upload")
                uploadbtn = gr.Button("image upload")
            with gr.Column():    
                with gr.Row():
                    im = gr.ImageMask(height=600)
                with gr.Row():
                    savebtn = gr.Button("save")
                    loadbtn = gr.Button("load")
                with gr.Accordion():
[...]

PR #7624, opened 2024-03-06
Fixing root path issue with subpath being repeated twice

abidlabs:
We have an issue with the subpath being repeated twice when constructing the `root_url`. I can't quite figure out why this happens, because I don't see the issue when I set up nginx proxy myself, but my guess it has something to do with which headers the nginx proxy attaches to the request. I've modified the `get_root_url` logic to check to see if the subpath is already present and if so, skip adding it. 

Fixes: https://github.com/gradio-app/gradio/issues/7520


abidlabs:
Has been [validated internally](https://huggingface.slack.com/archives/C0517L95ASG/p1709756391688419?thread_ts=1709749802.327169&cid=C0517L95ASG)

pngwn (approved):
lgtm! Tested as much as I can locally, if the test all pass then looks good to merge. Thanks @abidlabs!

abidlabs:
Hmm tests have decided not to pass, and it actually looks like a legit failure. Looking into it some more

PR #7628, opened 2024-03-06
feature detect CSSStylesheet

### LINKED ISSUE #7437: Add guard around `CSSStyleSheet`
### Describe the bug

We use the `CSSStyleSheet` constructore to scope custom CSS but support in Safari is relatively recent (few years) so there are some users who will run into issues with this. In these cases the app will crash completely.

We can do a simple feature detection for this and bail on scoping for those users. It does mean that custom CSS might not behave as expected but the numbers that this bug affects are pretty low, so it doesn't warrant a reimplementation and any negative impact would likewise be limited, and at least the app wouldn't crash. 

### Reproduction

Anything but you need to run Safari prior to 16.2, i think.

### Screenshot

_No response_

### Logs

_No response_

### System Info

```shell
Safari, 16.2 or earlier.
```

### Severity

Blocking usage of gradio

pngwn:
## Description

Closes #7437 

Hopefully we can get rid of this when we change scoping but it works well for now. Tested with lmsys folk who reported the issue.

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/evil-emus-relax.md
+++ b/.changeset/evil-emus-relax.md
@@ -1,0 +1,6 @@
+---
+"@gradio/app": patch
+"gradio": patch
+---
+
+fix:feature detect CSSStylesheet
--- a/.changeset/itchy-wombats-end.md
+++ b/.changeset/itchy-wombats-end.md
@@ -1,0 +1,5 @@
+---
+"gradio": patch
+---
+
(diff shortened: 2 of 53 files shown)


abidlabs (approved):
lgtm @pngwn! This has been validated and I don't see any side effects when testing in normal browsers.

Issue #7621, opened 2024-03-06
`gradio cc create` Command Fails Without Virtual Environment Due to pip Path Issue


MobisParkHeekang:
### Describe the bug

I encountered an issue with the `gradio cc create` command failing when executed in a global Python environment, rather than within a virtual environment. The root cause seems to be how Gradio locates the pip command.

In `gradio/cli/commands/components/install_component.py`, the `_install_command()` methods attempts to locate pip using `shutil.which("pip")`. For global python, pip is typically located at `/usr/bin/pip3`. `pip` usually refers to global python2's pip, unless you use virtual environment which override `pip` to refer to virtual environment python3. On my system, I don't use both global python2 and virtual environment, so `shutil.which("pip")` returns `None` as there's no executable named `pip`.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```bash
gradio cc create MyComponent
```
I'm using global python located at `/usr/bin/python3` and global pip located at `/usr/bin/pip3`. I make aliases `python="/usr/bin/python3"`, `pip="/usr/bin/pip3"` on `~/.bashrc`. Without those aliases, `which pip` command on bash shell fails as there's no executable named `pip` on my `PATH`.

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Linux
gradio version: 4.19.2
gradio_client version: 0.10.1

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.2.0
fastapi: 0.110.0
ffmpy: 0.3.2
gradio-client==0.10.1 is not installed.
httpx: 0.27.0
huggingface-hub: 0.20.3
importlib-resources: 6.1.2
jinja2: 3.1.2
markupsafe: 2.1.3
matplotlib: 3.8.3
numpy: 1.26.4
orjson: 3.9.15
packaging: 23.1
pandas: 2.2.0
pillow: 10.2.0
pydantic: 2.6.1
pydub: 0.25.1
python-multipart: 0.0.9
pyyaml: 6.0.1
ruff: 0.2.2
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.9.0
typing-extensions: 4.8.0
uvicorn: 0.27.1
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2023.10.0
httpx: 0.27.0
[...]

Issue #6821, opened 2023-12-18
Custom Components Guide does not seem to lead to a working example


AlUlkesh:
### Describe the bug

I tried following the guide at https://www.gradio.app/guides/custom-components-in-five-minutes.

I didn't make any code-changes yet, so I'm expecting to see my new component as a SimpleTextbox.

However when I go to the frontend URL I only see a screen with an infinite "Loading" text. The console shows these errors:

```
Uncaught SyntaxError: Unexpected token '{' (at (index):55:35)

svelte.js:3232 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading '$$')
    at init (svelte.js:3232:44)
    at new Blocks (Blocks.svelte:669:31)
    at Array.create_if_block_1 (index.svelte:184:43)
    at Object.p (Index.svelte:394:41)
    at update_slot_base (svelte.js:203:8)
    at Object.p (core.js:231:1)
    at update (svelte.js:2128:30)
    at flush (svelte.js:2088:5)
```

Line 55 at index looks like this:

`		<script>window.gradio_config = {{ config | toorjson }};</script>`

Is this a bug? Am I missing a step?

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

Follow the guide until you enter the frontend URL.

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Windows
gradio version: 4.10.0
gradio_client version: 0.7.3

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.2.0
fastapi: 0.105.0
ffmpy: 0.3.1
gradio-client==0.7.3 is not installed.
httpx: 0.24.1
huggingface-hub: 0.19.4
importlib-resources: 6.1.1
jinja2: 3.1.2
markupsafe: 2.1.3
matplotlib: 3.8.2
numpy: 1.23.5
orjson: 3.9.10
packaging: 23.2
pandas: 2.1.4
pillow: 9.5.0
pydantic: 2.5.2
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0.1
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.9.0
typing-extensions: 4.8.0
uvicorn: 0.24.0.post1
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2023.12.1
[...]

PR #7638, opened 2024-03-07
Add --python-path, --pip-path, --gradio-path CLI arguments to let custom component developers control which executable is used

freddyaboulton:
## Description

Closes: #7621 

Adds CLI arguments to let developers control the path of the python executables used in create, dev, and build commands.

Also adds some debugging information to dev mode in case no custom components are found. I suspect this will help debug issues like #6821 

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/fast-heads-visit.md
+++ b/.changeset/fast-heads-visit.md
@@ -1,0 +1,6 @@
+---
+"@gradio/preview": patch
+"gradio": patch
+---
+
+fix:Add --python-path, --pip-path, --gradio-path CLI arguments to let custom component developers control which executable is used
--- a/gradio/cli/commands/components/build.py
+++ b/gradio/cli/commands/components/build.py
@@ -2,6 +2,7 @@ import importlib
 import shutil
 import subprocess
 from pathlib import Path
+from typing import Optional
 
 import semantic_version
 import typer
@@ -13,6 +14,7 @@ from gradio.cli.commands.components._docs_utils import (
     get_deep,
 )
 from gradio.cli.commands.components.docs import run_command
+from gradio.cli.commands.components.install_component import _get_executable_path
 from gradio.cli.commands.display import LivePanelDisplay
 
 gradio_template_path = Path(gradio.__file__).parent / "templates" / "frontend"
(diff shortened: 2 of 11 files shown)


freddyaboulton:
We used to use `which` to determine the path of the gradio executable so we could launch the demo in reload mode, e.g. `gradio demo/app.py`. But it's possible that it picks up the wrong version of gradio if multiple versions of gradio are installed in the same system. So I'm adding a CLI argument to let developers take control if something goes wrong.

abidlabs:
Thanks @freddyaboulton for this PR! While I think this is a working solution, it seems like a bit of a headache for the user to have to find and pass in the path to their Python for the CLI commands, particularly as they need to do it for each `gradio cc` command. We can add these arguments, but could we also handle the default case better:

What if we resolved `pip` in this order:

* the `pip_path` cli arguments, if provided, otherwise:
* which("pip3") if exists, otherwise 
* which("pip") if exists, otherwise
* fail loudly

Same for `python3` and `python`. I think that should make the default behavior much better for the majority of users

freddyaboulton:
> it seems like a bit of a headache for the user to have to find and pass in the path to their Python for the CLI commands, particularly as they need to do it for each gradio cc command.

I don't follow. Passing in the `-path` parameters is not required and is only needed if whatever heuristic we have in place fails. We can also check for `python3/pip3` though.

abidlabs:
> I don't follow. Passing in the -path parameters is not required and is only needed if whatever heuristic we have in place fails. We can also check for python3/pip3 though.

Sorry my earlier comment may not have been clear. I agree -- let's keep these parameters but check for python3 and pip3 as well

freddyaboulton:
Sounds good @abidlabs ! Just pushed up that change!

abidlabs:
Code changes look good! In my tests, `--pip-path` is working, but `--python-path` is not having any effect.

For example, I can run:

```py
gradio cc dev --python-path abdef/asdfklasdjf
```

after creating a custom component and the dev server launches successfully. That's not expected, right?

freddyaboulton:
@abidlabs I think you needed to built the js for this PR for the error to show up for dev mode. But your comment made me realize we wouldn't raise a helpful error if the executable did not exist so I just pushed that up.

abidlabs:
Ah I see, thanks I'll give this another spin

abidlabs (approved):
Beautiful. LGTM

abidlabs:
Going to merge this in for the release @freddyaboulton

PR #7641, opened 2024-03-08
Use `x-forwarded-host` header to determine the root url, and let users provide a full `root_path` to override the automatically determined root url

abidlabs:
This PR fixes the issue that we're seeing where Gradio apps don't render properly on k8 environments and some other places where the URL of the Gradio app is behind an app router or proxy.

The fix is to look at the `x-forwarded-host` header, if available, and use that in place of the request URL, as this header points to the public URL of the Gradio app.

In addition, we provide users with a fallback in case Gradio is not able to determine the root url automatically. By providing a full `root_path` (either as an env variable or as a parameter in `launch()`), users can set their own root url, which takes precedence over the one Gradio constructs automatically.

Closes: https://github.com/gradio-app/gradio/issues/7317


freddyaboulton (approved):
Nice fix @abidlabs !

abidlabs:
Thanks a bunch for the review @freddyaboulton

koderx:
mount_gradio_app subpath mount with reverse proxy not working properly
@abidlabs 

mount at `/subpath`
```python
app = gradio.mount_gradio_app(app, io, path="/subpath")
```

reverse proxy route `http://xxx.com/subpath` to `localhost:1234/subpath`
and set `x-forwarded-host` = `xxx.com`

The `/subpath` is lost in `route_util.get_root_url`

```python
# gradio.route_util.get_root_url
x_forwarded_host = request.headers.get("x-forwarded-host")
root_url = f"http://{x_forwarded_host}" if x_forwarded_host else str(request.url) 
# ^ only http://xxx.com, /subpath is missing
```

in frontend, `config.root` is `http://xxx.com` 
some resource `config.root + '/theme.css'` is 404
```
await mount_css$1(config.root + "/theme.css", document.head);
```

and if `io` is `Blocks`,  `mount_gradio_app` is no way to set `root_path`, only from environment
```python
       # blocks.py  Blocks.__init__
        self.root_path = os.environ.get("GRADIO_ROOT_PATH", "")
       # routes.py App.configure_app
       self.root_path = blocks.root_path
```

PR #7661, opened 2024-03-11
Convert Docs Demos to Lite

aliabd:
Converts the demos on the docs to use the lite playground element instead of embedding from spaces. 

This is what it looks like: 

https://github.com/gradio-app/gradio/assets/9021060/94431b7e-43c2-4dea-8b9d-9e62f5dacd57

I removed/replaced the demos that won't work with lite. Please test them out on the preview (make sure you're on main!)

I also overided some small styling on the playground element, for example i don't think side by side works with this amount of space, so I used css to force code editor to be on top of preview (where it normally does that at <md breakpoint). I can clean this up on the Playground element in the future once i'm sure about the styling. I actually have a few other things i'd like to change as well.

There are still a lot of demos on the website that embed from spaces (guides, landing page, embedded components in docs, etc) but this is a good place to start. 

Also so that there aren't any broken demos on /docs when this is merged, I kept those embedding from spaces for now. So only /main/docs uses lite. This is because some of the demos on the released version don't work with lite. Right before the next release I can undo the changes in this [commit](https://github.com/gradio-app/gradio/pull/7661/commits/07d2e24fb110f997fb75ea3d87a84338237b3417) and those will be lite too.


abidlabs:
This looks awesome @aliabd! Thanks for through and updating the demos as well

I just noticed two small issues: 

1. When you change the component whose docs you are viewing, gradio/lite has to load all over again. I thought the SPA nature of our website would mean we only have to load it once? If there's a way to enable that, that'd be awesome.

2. While gradio/lite is loading, there's no way for a user to copy-paste the code for a particular demo. Its not as big of a problem when the demo is interactive, because at that point, you can just click inside the code block, and select all of the text, but otherwise its kind of annoying.

Ideally, we include both the copy and download buttons that are present in the regular `Code` component:

image


Neither of these are blockers. In particular, for the 2nd one, perhaps it'd be good to have a separate PR that updates the <gradio-lite playground> tag to include this. 

If so, I'd also suggest a few other things I noticed:
* When pressing CMD+Enter, it also creates a new line in the code. It should only Run the code block
* I think we should add a `layout` attribute that can be "horizontal" or "vertical" so that we don't need the custom css in this PR
* Design nits:

image

  * I think the size of words "app.py", "interactive", and "run" should all be the same. They should all be the size that "app.py" currently is
  * I think the lightning bolt next to Interactive should actually be orange (sorry I said gray before but I think orange is better since the color of the loading circle is also orange)
  * I think the Run button should be orange as well to make it more prominent.
  * Btw love the ⌘+↵

abidlabs (approved):

aliabd:
Thanks @abidlabs for the detailed feedback! 

1. Check now! :) 

2. There was a really annoying styling issue with those buttons that I couldn't figure out. So I temporarily hid them using css. 
https://github.com/gradio-app/gradio/blob/cf98c7ed18c0211b89e8eed81a5fb61c34b705c3/js/app/src/lite/Playground.svelte#L322-L327
Then I kind of forgot about them. I'll add them back for sure. 

I agree with all your suggestions, and had thought about some independently while I was making the PR. Def wanna change the design a bit more. Anyway will make a PR that: 

- brings back code editor buttons
- fixes cmd+enter so there's no newline 
- adds the layout attr
- design nits

abidlabs (approved):
🔥🔥🔥🔥

aliabd:
shared-worker mode! cc @whitphx

Issue #7659, opened 2024-03-11
mount_gradio_app subpath mount with reverse proxy not working properly


koderx:
### Describe the bug

mount at `/subpath`
```python
app = gradio.mount_gradio_app(app, io, path="/subpath")
```

reverse proxy route `http://xxx.com/subpath` to `localhost:1234/subpath`
and set `x-forwarded-host` = `xxx.com`

The `/subpath` is lost in `route_util.get_root_url`
https://github.com/gradio-app/gradio/blob/6683ab2589f9d8658e1f51acc1b7526edce988d3/gradio/route_utils.py#L301
```python
# gradio.route_util.get_root_url
x_forwarded_host = request.headers.get("x-forwarded-host")
root_url = f"http://{x_forwarded_host}" if x_forwarded_host else str(request.url) 
# ^ only http://xxx.com, /subpath is missing
```

in frontend, `config.root` is `http://xxx.com` 
some route are 404
```ts
// dynamic import css
await mount_css$1(config.root + "/theme.css", document.head);
// client.ts /info
response = await fetch_implementation(`${config.root}/info`, {
```

and if `io` is `Blocks`,  `mount_gradio_app` is no way to set `root_path`, only from environment
```python
       # blocks.py  Blocks.__init__
       self.root_path = os.environ.get("GRADIO_ROOT_PATH", "")
       # routes.py App.configure_app
       self.root_path = blocks.root_path
```
or in trick
```python
with gr.Blocks() as io:
    # code

io.root_path = "/subpath" # set root_path here, but i think is so trick
app = gradio.mount_gradio_app(app, io, path="/subpath")
```

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr
app = gr.mount_gradio_app(app, io, path="/subpath")
```
reverse proxy set `x-forwarded-host` = `xxx.com`
http://xxx.com/subpath ->  innerhost:1234/subpath

### System Info

```shell
gradio==4.21.0
```

### Severity

I can work around it

abidlabs:
I'll look into this. Just to clarify, is this a new bug in 4.21 or did you notice this in older versions of Gradio (such as 4.20.1) as well?

abidlabs:
Hi @koderx I have a PR which exposes a `root_path` parameter in `gr.mount_gradio_app` which I think is the best way to solve this, providing maximal flexibility to the developer

[...]

Issue #7629, opened 2024-03-06
Allow developers to use built-in auth when mounting in a FastAPI app


freddyaboulton:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
We recently added third-party auth with the `auth_dependency` parameter. This works when launching gradio as a standalone application or within a larger FastAPI app. In contrast, you can't use the built-in auth when launching within a larger FastAPI app.

There is some user interest in this as noted in this thread: https://discord.com/channels/879548962464493619/1198067704314208338

**Describe the solution you'd like**  
- Add an auth parameter to `mount_gradio_app`
- Add an auth parameter to `__init__` so that developers can set auth prior to running `mount_gradio_app`. 

**Additional context**

Issue #7607, opened 2024-03-05
where the ENTRY_CSS path is not added to the path when Gradio is launched with a specified path.


qyxing:
### Describe the bug

When running with a specified path set, ENTRY_CSS fails to load correctly as the path is not added to the beginning of the path.

app = gr.mount_gradio_app(app, interface1, path="/cao/")

<img width="980" alt="image" src="https://github.com/gradio-app/gradio/assets/3423208/bc26205d-0d10-41ba-8206-415ba1885e17">

Solution:
Modify the mount_css function in ./js/app/src/css.ts to concatenate the path with the specified path.

<img width="659" alt="image" src="https://github.com/gradio-app/gradio/assets/3423208/37c2ee0a-7d16-4d38-877e-adf23ea1d8b3">

The modified path is displayed correctly.
<img width="1028" alt="image" src="https://github.com/gradio-app/gradio/assets/3423208/4bf28847-484f-49a4-a1f8-98c9fd809286">

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr
app = gr.mount_gradio_app(app, interface1, path="/cao/")
```

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Darwin
gradio version: 4.19.2
gradio_client version: 0.10.1

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.2.0
fastapi: 0.109.2
ffmpy: 0.3.1
gradio-client==0.10.1 is not installed.
httpx: 0.26.0
huggingface-hub: 0.20.3
importlib-resources: 6.1.1
jinja2: 3.1.3
markupsafe: 2.1.5
matplotlib: 3.8.2
numpy: 1.26.4
orjson: 3.9.13
packaging: 23.2
pandas: 2.2.0
pillow: 10.2.0
pydantic: 2.6.1
pydub: 0.25.1
python-multipart: 0.0.9
pyyaml: 6.0.1
ruff: 0.3.0
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.9.0
typing-extensions: 4.9.0
uvicorn: 0.27.0.post1
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2024.2.0
httpx: 0.26.0
huggingface-hub: 0.20.3
packaging: 23.2
[...]

PR #7667, opened 2024-03-11
Add `auth`, `auth_message`, and `root_path` parameters to `mount_gradio_app`

abidlabs:
Firstly adds `auth` and `auth_message` to close: https://github.com/gradio-app/gradio/issues/7629

Also adds `root_path` to `mount_gradio_app`. The reason is that we now look at the `x-forwarded-host` header to automatically determine the request url. However, depending on how the developer sets of up the header, it may or may not include the full path to the gradio app. To provide full flexibility to the developer, we add an additional `root_path` to the Gradio app. Closes: https://github.com/gradio-app/gradio/issues/7659, Closes: https://github.com/gradio-app/gradio/issues/7607


freddyaboulton (approved):
Nice LGTM @abidlabs !

abidlabs:
Thanks @freddyaboulton!!

PR #7670, opened 2024-03-12
Benchmarking scripts for load tests

aliabid94:
Added seom scripts to simulate load tests on Gradio. Current results from streaming 500 tokens at 100 tok/s:
| Simultaneous Clients | Gradio 4.0 | Gradio 3.50 | Raw SSE | Raw Websockets |
|----------------------|------------|-------------|---------|----------------|
| 1                    | 6.17       | 5.9         | 5.2     | 5.2            |
| 5                    | 6.5        | 6.5         | 5.4     | 5.6            |
| 25                   | 21.2       | 10.0        | 5.8     | 5.9            |
| 100                  | 98.0       | 32.3        | 6.7     | 16.3           |
| 250                  | 276.9      | 91.4        | 14.5    | 43.3           |


scripts/load_test/app.py
```diff
@@ -0,0 +1,28 @@
...
+async def sse():
+    return StreamingResponse(number_generator(), media_type="text/event-stream")
+
+@app.websocket("/ws")
+async def websocket_endpoint(websocket: WebSocket):
+    await websocket.accept()
+    for number in range(1, 501):
+        message = "Lorem "*(number + 1)
+        await websocket.send_text(message)
+        await asyncio.sleep(0.01)
```
abidlabs, line 23: Nit: reduce duplication
```suggestion
    gen = number_generator()
    while True:
        try:
            message = await gen.__anext__()
            await websocket.send_text(message)
        except StopIteration:
            break
```

abidlabs (approved):

scripts/load_test/chat.py
```diff
@@ -0,0 +1,26 @@
...
+
+    if version == "3":
+        text.submit(respond, text, chatbot)
+    else:
+        text.submit(respond, text, chatbot, concurrency_limit=None)
+
+
+if __name__ == "__main__":
+    if version == "3":
+        demo.queue(concurrency_limit=None).launch(max_threads=250)
```
freddyaboulton, line 24: This doesn't work for gradio 3.50.2 for me -  it's `concurrency_count=250`  (can't be None)

freddyaboulton: When I made that modification, I get `OSError: [Errno 24] Too many open files` when running the 250 concurrent users case for gradio 3.x. 

What version of 3 are you using?

freddyaboulton (commented):
Looks great to me but I can't repro the 3.x numbers for 250 concurrent users. I am getting an error. Did you use nginx?

scripts/load_test/chat.py
aliabid94: I was using 3.50 as well, behind an nginx server

freddyaboulton: Cool, I think we should make a note of that for reproducibility. But I guess this is a limitation of WS to keep in mind.

abidlabs:
Going to merge this in as I'd like to use these scripts in other PRs. Thanks @aliabid94!

Issue #7701, opened 2024-03-14
Flagging does not handle files correctly


abidlabs:
### Describe the bug

When a Gradio Interface that contains file-based components is flagged, files are not saved as filepaths, rather the entire FileData dictionary is saved, e.g.

```
x,y,output 0,output 1,flag,username,timestamp
abc,"{""path"":""flagged/y/ee116ed9c7acc9a98c59/001.jpg"",""url"":""http://127.0.0.1:7860/file=/private/var/folders/kf/ll6fqkys7_s05x4tvb546mdc0000gn/T/gradio/30051634f345dc47162df9144c44b3f68647ee97/001.jpg"",""size"":319979,""orig_name"":""001.jpg"",""mime_type"":""image/jpeg"",""is_stream"":false,""meta"":{""_type"":""gradio.FileData""}}",abc,"{""path"":""flagged/output 1/bbe052c0d4955425f70d/image.png"",""url"":""http://127.0.0.1:7860/file=/private/var/folders/kf/ll6fqkys7_s05x4tvb546mdc0000gn/T/gradio/eef3453ead8c83d394ce7ade4d0eb4b754dbb5b5/image.png"",""size"":null,""orig_name"":""image.png"",""mime_type"":null,""is_stream"":false,""meta"":{""_type"":""gradio.FileData""}}",,,2024-03-13 21:24:47.604304
```

which makes it very hard to parse

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

gr.Interface(lambda x,y:(x,y), ["textbox", "image"], ["textbox", "image"], allow_flagging="auto").launch()
```

### System Info

```shell
4.21
```

### Severity

I can work around it

PR #7704, opened 2024-03-14
Fix flagged files and ensure that flagging_mode="auto" saves output components as well

abidlabs:
Components with `FileData` values were being saved in their entirety in the CSV files. This PR makes it so that only the filepaths are saved. I considered modifying the `flag()` function, but that would lead to incompatibility with custom components and make it more complex for custom component developers to write their own `flag()` functions so instead I do some postprocessing after the `.flag()` function has been called. 

Closes: #7701 

Also fixed a long-standing issue where if `flagging_mode="auto"`, the output was not being saved.

Closes: https://github.com/gradio-app/gradio/issues/3441

I set this to `minor` because although these are technically bugfixes, the behavior has been this way for such a long time, some users might see this as a breaking change.


gradio/components/code.py
```diff
@@ -156,9 +155,6 @@ def postprocess(self, value: tuple[str] | str | None) -> None | str:
         else:
             return value.strip()
 
-    def flag(self, payload: Any, flag_dir: str | Path = "") -> str:
```
abidlabs, line 159: Not needed, the base class takes care of this.

gradio/components/json_component.py
```diff
@@ -94,13 +93,6 @@ def example_payload(self) -> Any:
     def example_value(self) -> Any:
         return {"foo": "bar"}
 
-    def flag(
```
abidlabs, line 97: Not needed anymore either, the base class takes care of this.

gradio/flagging.py
```diff
@@ -127,8 +127,8 @@ def image_classifier(inp):
     Guides: using-flagging
     """
 
-    def __init__(self):
-        pass
+    def __init__(self, simplify_files: bool = True):
```
abidlabs, line 130: We add this parameter because example caching also uses this method, and there, we actually do want to keep the full file data

freddyaboulton (approved):
Looks good. Tested and works well @abidlabs !

abidlabs:
Thanks @freddyaboulton

PR #7712, opened 2024-03-15
More fixes for `gr.load()` as well as a tweaking the `__str__` and `__repr__` methods of components

abidlabs:
I went through and tested the following Spaces using `gr.load()`:

        "audio_debugger_main",
        "blocks_essay_main",
        "blocks_group_main",
        "blocks_js_methods_main",
        "blocks_layout_main",
        "blocks_update_main",
        "calculator_main",
        "kitchen_sink_main",
        "matrix_transpose_main",
        "model3D_main",

These all worked fine. (Make sure to test the "_main" Spaces as some fixes for `gr.load` have already been pushed to `main`)

This one was broken:

        "blocks_multiple_event_triggers"

because in the config, we saved the "type" of a `gr.LinePlot` as just a "plot". We need to do this so that the frontend uses the correct svelte template but we should also save the true component class so that we can reconstruct for `gr.load()` and also so that the api info has the correct value. This PR fixes that.

You can confirm by testing:

```py
import gradio as gr

gr.load("abidlabs/blocks_multiple_event_triggers_main", src="spaces").launch()
```


gradio/components/base.py
```diff
@@ -240,7 +240,7 @@ def __str__(self):
         return self.__repr__()
 
     def __repr__(self):
-        return f"{self.get_block_name()}"
+        return f"<{self.__class__.__name__} object> with config: {self.get_config()}"
```
abidlabs, line 243: Made this change due in part to this internal discussion: https://huggingface.slack.com/archives/C02QZLG8GMN/p1708388337912479?thread_ts=1708385877.550569&cid=C02QZLG8GMN

freddyaboulton: Do we need to implement custom `__str__` and `__repr__`? Dumping the whole config to the terminal can be unwieldy and the built-in methods make it clear it's an object and not just a type

image

gradio/blocks.py
```diff
@@ -2481,7 +2492,7 @@ def get_api_info(self):
                 else:
                     skip_endpoint = True  # if component not found, skip endpoint
                     break
-                type = component["type"]
+                type = component["props"].get("name")
```
freddyaboulton, line 2495: Isn't `name` always present? I'm thinking we can remove the `get`. I'm not sure it makes sense anyways because as-is the call to `capitalize` below will fail if `type` is None.

gradio/components/base.py
abidlabs: Ok I don't feel strongly at all, will just eliminate them

gradio/blocks.py
abidlabs: Sure yep

freddyaboulton (commented):
Fixes look good - I just left a couple of nits!

Issue #5497, opened 2023-09-12
Gradio Client Is Unable to Connect to Locally Served HTTPS API


pourmand1376:
### Describe the bug

When trying to connect to a locally served HTTPS API from a Gradio client, an SSL certification verification error occurs that prevents the client from accessing the API.

> Note that the Gradio web UI works very good in this case but Gradio API client in unusable. 

Specifically:

A simple Gradio app is created that serves an HTTPS API on localhost.
A self-signed SSL certificate is added to enable HTTPS on localhost.
The Gradio client tries to connect to this local HTTPS API, but gets an SSLCertVerificationError.
This is because the Gradio client does not trust the locally generated self-signed certificate used by the server. During the SSL handshake, certificate verification fails with the error:

"[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self-signed certificate"

As a result, the SSL connection between the client and server cannot be established. The client is unable to send any requests to the local HTTPS API.

I do not know how I can fix this. I have tried solutions from [here](https://stackoverflow.com/questions/15445981/how-do-i-disable-the-security-certificate-check-in-python-requests) but none of them worked. 

Here is also the code I am using to connect to gradio local endpoint:

```python
from gradio_client import Client
client =  Client("https://10.2.20.20:8006/")
 result = client.predict(input_parameters,api_name="/run"
  )
 print(result)
```

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

- Generate a simple app which uses gradio. 
- Add Local Signed Certificate to it via [this](https://github.com/gradio-app/gradio/issues/563) or [this](https://discuss.huggingface.co/t/how-to-run-gradio-with-0-0-0-0-and-https/38003). 
- There is no way to connect to that https endpoint. 

### Logs
Loaded as API: https://10.2.20.20:8006/ ✔
```shell
SSLCertVerificationError                  Traceback (most recent call last)
File ~/.pyenv/versions/3.10.12/lib/python3.10/site-packages/urllib3/connectionpool.py:714, in HTTPConnectionPool.urlopen(self, method, url, body, headers, retries, redirect, assert_same_host, timeout, pool_timeout, release_conn, chunked, body_pos, **response_kw)
    713 # Make the request on the httplib connection object.
--> 714 httplib_response = self._make_request(
    715     conn,
    716     method,
    717     url,
    718     timeout=timeout_obj,
    719     body=body,
    720     headers=headers,
[...]

PR #7718, opened 2024-03-15
Add support for python client connecting to gradio apps running with self-signed SSL certificates

abidlabs:
This PR adds an `ssl_verify` parameter to `gradio_client.Client` to match the corresponding parameter in `launch()`

Not easy to write a unit test for this but you can run the following to confirm:

1. `openssl req -x509 -newkey rsa:4096 -keyout key.pem -out cert.pem -sha256 -days 365 -nodes` in terminal
2. In the same directory, launch a gradio app:

```py
import gradio as gr

demo = gr.Interface(lambda x:x, "image", "image")

_, url, _ = demo.launch(
    ssl_certfile="cert.pem",
    ssl_keyfile="key.pem",
    ssl_verify=False
)
```

3. Connect with the client:

```py
from gradio_client import Client, file

client = Client(url, ssl_verify=False)

client.predict(file("cheetah.jpg"))
```

Fixes: https://github.com/gradio-app/gradio/issues/5497

### DIFF
--- a/.changeset/kind-eyes-shake.md
+++ b/.changeset/kind-eyes-shake.md
@@ -1,6 +1,6 @@
 ---
-"gradio": minor
-"gradio_client": minor
+"gradio": patch
+"gradio_client": patch
 ---
 
-feat:Add support for python client connecting to gradio apps running with self-signed SSL certificates
+fix:Add support for python client connecting to gradio apps running with self-signed SSL certificates
--- a/.changeset/kind-eyes-shake.md
+++ b/.changeset/kind-eyes-shake.md
@@ -1,0 +1,6 @@
+---
+"gradio": minor
+"gradio_client": minor
+---
+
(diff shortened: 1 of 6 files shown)


freddyaboulton (approved):
Looks good @abidlabs ! I think we're missing `check_for_cancel` and (`download_byte_stream`, `stream_sse_v0`, `download_temp_copy_of_file`) so that this parameter can work for 3.x spaces

abidlabs:
As part of this PR, I moved the `download_files()` and `upload_files()` from utils.py to client.py since we were passing so many of the client attributes as parameters anyways. I've updated the tests accordingly.

abidlabs:
Thanks @freddyaboulton for the review and for catching those. I'll go through and add

abidlabs:
Should be all good now, will merge in after CI passes

PR #7729, opened 2024-03-18
Revert "Replace httpx with requests library in save_url_to_cache to make it work in wasm mode"

abidlabs:
Reverts gradio-app/gradio#7725:

> One issue is that this PR makes requests a dependency of gradio, which is not ideal. Currently, this passes because requests is a dependency of huggingface_hub but that may change in the future so we would need to include it as an explicit dependency. Let me revert this PR and we can figure this out as part of https://github.com/gradio-app/gradio/pull/7728

### DIFF
--- a/.changeset/spotty-parents-wink.md
+++ b/.changeset/spotty-parents-wink.md
@@ -1,5 +1,0 @@
----
-"gradio": patch
----
-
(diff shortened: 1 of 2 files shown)


freddyaboulton (approved):
Thanks @abidlabs !

abidlabs:
Thanks @freddyaboulton

Issue #5344, opened 2023-08-25
Gradio API kwargs support


taoari:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  

Current gradio API only supports positional arguments, it can become unmanaged if there are a lot of parameters for complex applications.  It would be great if kwargs would be supported. 

For example, the following code 

```python
import random
import gradio as gr

def random_response(message, history):
    return random.choice(["Yes", "No"])

params = [gr.Slider(minimum=0, maximum=1, value=0.7, step=0.1, interactive=True, label="Temperature",
            info="Lower temperator for determined output; hight temperate for randomness"),
    gr.Slider(minimum=0, maximum=1024, value=512, step=64, interactive=True, label="Max new tokens"),]
demo = gr.ChatInterface(random_response, additional_inputs=params)

if __name__ == "__main__":
    demo.launch()
```

generates the following API:

![](https://user-images.githubusercontent.com/3815006/263124397-4465f8d1-bcc7-414d-9829-d5cee29c7671.png)

It would be great if the API can have kwargs support:

```python
from gradio_client import Client

client = Client("http://localhost:7860/")
result = client.predict(
				message="Howdy!",	# str in 'Message' Textbox component
				temperature=0.7,	# int | float (numeric value between 0 and 1)
								in 'Temperature' Slider component
				max_new_tokens=512,	# int | float (numeric value between 0 and 1024)
								in 'Max new tokens' Slider component
				api_name="/chat"
)
print(result)
```

First, the API function kwargs names are generated from "label"; Second, the default values are generated from "value".

**Describe the solution you'd like**  

See above.

**Additional context**

abidlabs:
Related issue: https://github.com/gradio-app/gradio/issues/5049

abidlabs:
Hi @taoari this is a great suggestion and we're working on this in #7732. The PR is currently a WIP but feel free to try it out and share any feedback that you have on the PR as it'll help us make the Client more useful. Thanks!

Issue #4498, opened 2023-06-13
Display default values in the API page of Spaces and `.view_api()`


hysts:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  

I find it very useful that the API page shows the range of `gr.Slider`, but it would be more helpful if the default values were shown in the API page as well.
![](https://github.com/gradio-app/gradio/assets/25161192/a184bc0a-1c7f-4265-9bf1-26f6e2ef3357)
![](https://github.com/gradio-app/gradio/assets/25161192/b9b0ad34-9d64-4a70-b52f-5b5127a7f3e1)

(BTW, it would be super useful if we didn't even have to pass parameters that has default values, but it's not easy, I guess?)

Issue #3540, opened 2023-03-20
Client does not support default parameters in functions


dbalabka:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
Adding new input into `gr.Interface` leads to backward incompatible changes in API and leads to the following error:
```
{
  "error": "list index out of range"
}
```
In my particular case, I've added:
```python
gr.Radio(["vtt", "txt", "srt", "tsv", "json"], value='vtt', label='Caption file format'),
```
This input has a default value and does not require input. Still, it leads to `500 Internal Server Error`

**Describe the solution you'd like**  
The simplest solution is to allow skipping the values with default values.

**Additional context**  
Gradio version is `3.21.0`

dbalabka:
@abidlabs, the default value present in the function's parameter :
```python
def transcribe(
        language = 'auto',
        audio = None,
        audio_file = None,
        output_format = 'vtt',
)
```
Anyway, the API should not return the 500 error. It should be classified as 400 Bad requests instead.

dbalabka:
Thanks @abidlabs

PR #7732, opened 2024-03-18
Adds support for kwargs and default arguments in the python client, and improves how parameter information is displayed in the "view API" page

abidlabs:
This PR adds support for kwargs and default arguments in the Client. In other words, suppose you have the following demo:

```py
    if operation == "add":
        return num1 + num2
    elif operation == "subtract":
        return num1 - num2
    elif operation == "multiply":
        return num1 * num2
    elif operation == "divide":
        if num2 == 0:
            raise gr.Error("Cannot divide by zero!")
        return num1 / num2

demo = gr.Interface(
    calculator,
    ["number", gr.Radio(["add", "subtract", "multiply", "divide"], value="add), "number"],
    "number",
    examples=[
        [5, "add", 3],
        [4, "divide", 2],
        [-4, "multiply", 2.5],
        [0, "subtract", 1.2],
    ],
).launch()
```

When you connect to this app via the client, you can now make queries like this:

```py

client = Client("http://127.0.0.1:7861/")
result = client.predict(
		num1=3,
		operation="add",
		num2=10,
		api_name="/predict"
)
print(result)
```

Note that parameters to `.predict()` are now supplied with key-word arguments. For backwards compatibility, the old approach using positional arguments continues to work. e.g. 

```py
result = client.predict(
		3,
		"add",
		33,
		api_name="/predict"
)
```

but the key-word approach is encouraged because it is more robust and also allows us to take advantage of default arguments. In the example above, we could also do:

```py

client = Client("http://127.0.0.1:7861/")
result = client.predict(
		num1=33,
		num2=33,
		api_name="/predict"
)
print(result)
```

taking advantage of the fact that `operation` is `"add"` by default, as specified in the initial value of the component.

Closes: https://github.com/gradio-app/gradio/issues/3540
Closes: https://github.com/gradio-app/gradio/issues/5344
Closes: https://github.com/gradio-app/gradio/issues/4498 

Finally, I've updated the design of the view API page to showcase all this information about each parameter. (parameter names, whether it is required, or the default fallback value if it is not). Here's a partial screenshot:

image


gradio/blocks.py
```diff
@@ -2482,29 +2482,51 @@ def get_api_info(self):
...
+                # have those names. Hence the final checks.
+                if (
+                    dependency["backend_fn"]
+                    and index < len(fn_info)
+                    and fn_info[index][0]
+                    not in ["api_name", "fn_index", "result_callbacks"]
+                ):
+                    parameter_name = fn_info[index][0]
+                    parameter_has_default = fn_info[index][1]
+                    parameter_default = fn_info[index][2]
```
freddyaboulton, line 2518: The default value may not be json-serializable. We should check against that return `None` in that case.

freddyaboulton:
> I would say that this also closes: https://github.com/gradio-app/gradio/issues/4498 because it would be confusing to show two different "defaults" -- the function default and the slider default. Instead, we just show the function default if it is provided.

I think it would be better if we showed the component default value. That's the standard way gradio app developers specify default values for the app. I don't think function defaults are that common in gradio since they don't really take effect. We also know the component value is serializable since it's been postprocessed by the component and is in the config.

abidlabs:
> I think it would be better if we showed the component default value. That's the standard way gradio app developers specify default values for the app. I don't think function defaults are that common in gradio since they don't really take effect. We also know the component value is json-serializable since it's been postprocessed by the component and is in the config.

Interesting. So just to to understand, you're suggesting using the initial value for a component, if it has one, as the default when accessing via the clients?

We could do that, but what if a developer wants to set a default for the clients but not change their UI?

freddyaboulton:
Yes that's what I was suggesting! The other point about using the python function's default value is that I'm not sure it would work for some file based components (like Image). If you have a python function that is expecting a PIL image, you'd have to set the default value of the parameter to be a URL string for it to be useful in the ViewAPI page, which is awkward. 

I hear you about maybe wanting to specify a different default value for the API as opposed to the UI but I think pulling the defaults from the component as opposed to the function is closer to what an app author would expect (and saves them from having to update defaults in two places if they do want to use the UI default for the API default).

abidlabs:
Yeah makes sense, I'll change that!

Also just noticed this issue where the author suggests exactly your approach so that's good validation too: https://github.com/gradio-app/gradio/issues/5344

abidlabs:
By the way @gary149 if you have feedback on the new design of the view API page, that would be super helpful.

Here's an example of how it looks: https://gradio-pr-deploys-pr-7732-all-demos.hf.space/demo/altair_plot/?__theme=dark&view=api

freddyaboulton:
@abidlabs Yea good point about default value of None. Agree with your proposal!

abidlabs:
Ok I've made the change @freddyaboulton. Still working on the documentation, but should otherwise be good for a review

test/test_utils.py
```diff
@@ -453,3 +454,74 @@ def test_get_extension_from_file_path_or_url(path_or_url, extension):
...
+            ("arg1", False, None),
+            ("arg2", True, 42),
+        ]
+
+    def test_class_method_with_args(self):
+        class MyClass:
+            def method(self, a, *args, b=42):
+                pass
+
+        assert get_function_params(MyClass().method) == [("a", False, None)]
```
freddyaboulton, line 521: Why doesn't the b get picked up in this case?

js/app/src/api_docs/CodeSnippet.svelte
```diff
@@ -49,35 +48,30 @@
...
-			-->	# {python_type.type} {#if python_type.description}({python_type.description})
-								{/if}<!---->in '{label}' <!--
-			-->{component} component<!--
-			--></span
-							><!--
-		-->{/each}<!--
+		{parameter_name
+								? parameter_name + "="
+								: ""}<span class="example-inputs"
+								>{represent_value(
```
freddyaboulton, line 63: nit: there is an extra space added before the comma

Pasted image 20240321162808

abidlabs: It looks like that but I think there's no space actually, maybe its a padding issue

test/test_utils.py
abidlabs: We stop as soon as we hit an *args or **kwargs because we no longer know for sure which component that parameter corresponds to

guides/08_gradio-clients-and-lite/01_getting-started-with-the-python-client.md
```diff
@@ -130,13 +135,49 @@ client.predict(4, "add", 5)
...
+```python
+from gradio_client import Client
+
+client = Client("gradio/calculator")
+client.predict(num1=4, operation="add", num2=5)
+
+>> 9.0
+```
+
+The default value is the initial value of the corresponding Gradio component. If the component does not have an initial value, but if the corresponding argument in the predict function has a default value of `None``, then that parameter is also optional in the client. Of course, if you'd like to override it, you can include it as well:
```
freddyaboulton, line 161: Nit: there's a double ` in after None

client/python/gradio_client/client.py
```diff
@@ -659,15 +671,23 @@ def _render_endpoints_info(
...
-        if endpoints_info["parameters"]:
-            for info in endpoints_info["parameters"]:
+        if parameter_info:
+            for info in parameter_info:
                 desc = (
                     f" ({info['python_type']['description']})"
                     if info["python_type"].get("description")
                     else ""
                 )
+                default_value = info.get("parameter_default")
```
freddyaboulton, line 681: I think we need to traverse the value here to only pull out the actual filepath or url

js/app/src/api_docs/CodeSnippet.svelte
freddyaboulton: Interesting. I think ideal would be no padding but not a big deal

abidlabs:
Ok the docs should be ready as well now!

client/python/gradio_client/client.py
abidlabs: So the view API page takes care of this automatically:

image

but the `.view_api()` function does not

image

```py
import gradio as gr

demo = gr.Interface(
    lambda x:x,
    gr.Audio("test.mp3"),
    "audio"
)

_, url, _ = demo.launch()

client = Client(url)

client.view_api()
```

abidlabs: I'll fix in the `.view_api()`

freddyaboulton (commented):
Very slick PR @abidlabs ! Tested a variety of demos and they all work well!

one thing that would be nice is to have the URL of the file if it's on spaces otherwise the file upload would fail?

image

client/python/gradio_client/client.py
freddyaboulton: Yea sorry that's what I meant!

abidlabs:
Thanks @freddyaboulton! Agree will figure out a better solution for files

abidlabs:
Ok that should address everything in the review:

image

I'll merge in later today unless @aliabd has any other suggestions

aliabd (approved):
Really like this @abidlabs! 

I have two small suggestions that you can make the final call on. 

* I feel like the multiple parameters are blending together visually. I know you have the faint orange separator (in light mode), maybe making it more prominent or increasing the space? 
* I think 'Defaults to _' and 'Required' should occupy the same position. Right now 'Required' is after the name and type, where as the default value is in the end of the description. I think it would be better if you maybe put them both at the end of the first line, on the right.

Everything else looks so clean!

abidlabs:
Thanks for the reviews @freddyaboulton and @aliabd. I agree with your suggestions @aliabd, will make the changes and merge

abidlabs:
Updated. It now looks like this (more margin between the parameters and defaults on the same line as the required): 

image

Issue #7690, opened 2024-03-13
Gradio launch succeeds but gets stuck on loading screen if you use integers instead of strings for the "value" field on gr.Button()


heaversm:
### Describe the bug

Make a button with a value of `1`:

`vote1 = gr.Button(size="sm", value=1)`

Run the app (`python app.py`)

Terminal outputs: `running on local URL: http://127.0.0.1:7860`

Browser show a "loading" screen.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

Make a button with a value of `1`:

`vote1 = gr.Button(size="sm", value=1)`

Run the app (`python app.py`)

Terminal outputs: `running on local URL: http://127.0.0.1:7860`

Browser show a "loading" screen.

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Darwin
gradio version: 4.2.0
gradio_client version: 0.7.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.2.0
fastapi: 0.108.0
ffmpy: 0.3.1
gradio-client==0.7.0 is not installed.
httpx: 0.26.0
huggingface-hub: 0.20.1
importlib-resources: 6.1.1
jinja2: 3.1.2
markupsafe: 2.1.3
matplotlib: 3.8.2
numpy: 1.26.2
orjson: 3.9.10
packaging: 23.2
pandas: 2.1.4
pillow: 10.2.0
pydantic: 2.5.3
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0.1
requests: 2.31.0
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.9.0
typing-extensions: 4.9.0
uvicorn: 0.25.0
websockets: 11.0.3
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2023.12.2
httpx: 0.26.0
huggingface-hub: 0.20.1
packaging: 23.2
requests: 2.31.0
typing-extensions: 4.9.0
websockets: 11.0.3
```

### Severity

I can work around it

abidlabs:
Thanks for reporting @heaversm. I can confirm the issue on the latest version of `gradio`. While the `value` parameter is supposed to be a string, we can handle this case by casting to a string here:

[...]

PR #7733, opened 2024-03-18
Cast button value as string in postprocess

heaversm:
Gradio launch succeeds but gets stuck on loading screen if you use integers instead of strings for the "value" field on gr.Button().

## Description

Add forced casting to string of button value in postprocess function of gradio Button component (`return str(value)`)

Closes: #7690
(https://github.com/gradio-app/gradio/issues/7690#issuecomment-1997718185)

## Tests

### DIFF
--- a/.changeset/tasty-donkeys-warn.md
+++ b/.changeset/tasty-donkeys-warn.md
@@ -1,0 +1,5 @@
+---
+"gradio": patch
+---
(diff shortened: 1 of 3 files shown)


abidlabs (approved):
lgtm thanks @heaversm! added a quick test

Issue #7303, opened 2024-02-04
If gr.Dataset() isn't initialized with data on launch, pagination doesn't work


dwipper:
### Describe the bug

If gr.Dataset() isn't initialized with data on launch, pagination doesn't work. 

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```
import gradio as gr

def get_samples():

    samples = [['1'],['2'],['3'],['4'],['5'],['6'],['7'],['8'],['9'],['10']]
    return samples

with gr.Blocks() as demo:
    with gr.Row(variant='panel'):
        with gr.Column():
            ds_tbox = gr.Textbox(label='Example Name', visible=False)
            samples_ds = gr.Dataset(components=[ds_tbox], headers=['Samples'],label='Sample Questions', samples=get_samples(),samples_per_page=5)
        with gr.Column():
            examples_name_tbox = gr.Textbox(label='Examples Name', visible=False)
            examples_ds = gr.Dataset(components=[examples_name_tbox], headers=['NIL Questions'],
                                     label='Example Questions - click to select', samples='', samples_per_page=5)
    with gr.Row():
        get_examples_btn = gr.Button(value="Get Examples")
        get_examples_btn.click(fn=get_samples,inputs=None, outputs=examples_ds)

if __name__ == "__main__":
  demo.launch()
```

### Screenshot

On load:
![image](https://github.com/gradio-app/gradio/assets/140648761/7605b9aa-5974-41a7-a1df-0af5a6143fe3)
After clicking "Get Examples" button
![image](https://github.com/gradio-app/gradio/assets/140648761/a3b0ee7c-abc4-4fc9-8a6e-82f8877e2f55)

### Logs

```shell
None
```

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Darwin
gradio version: 4.16.0
gradio_client version: 0.8.1

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 22.1.0
altair: 5.1.2
fastapi: 0.104.0
ffmpy: 0.3.1
gradio-client==0.8.1 is not installed.
httpx: 0.25.0
huggingface-hub: 0.19.4
importlib-resources: 6.1.0
jinja2: 3.1.2
markupsafe: 2.1.1
matplotlib: 3.7.2
numpy: 1.24.3
orjson: 3.9.9
packaging: 23.1
pandas: 2.0.3
pillow: 9.4.0
pydantic: 2.4.2
[...]

PR #7761, opened 2024-03-20
Ensure `paginate` updates when samples value changes in `Dataset`

hannahblair:
## Description

We weren't updating the `paginate` value in the Dataset component, so I've we dynamically populated the value of samples, the paginate bool would still be false. This PR adds reactivity logic to the `paginate` value. 

❓ Shall I add an e2e test for Dataframe in this PR?

To test, run the following demo and click "get samples". Check the pagination works as expected in the dynamically updated Dataset. 

```
import gradio as gr

def get_samples():

    samples = [['1'],['2'],['3'],['4'],['5'],['6'],['7'],['8'],['9'],['10']]
    return samples

with gr.Blocks() as demo:
    with gr.Row(variant='panel'):
        with gr.Column():
            ds_tbox = gr.Textbox(label='Example Name', visible=False)
            samples_ds = gr.Dataset(components=[ds_tbox], headers=['Samples'],label='Sample Questions', samples=get_samples(),samples_per_page=5)
        with gr.Column():
            examples_name_tbox = gr.Textbox(label='Examples Name', visible=False)
            examples_ds = gr.Dataset(components=[examples_name_tbox], headers=['NIL Questions'],
                                     label='Example Questions - click to select', samples='', samples_per_page=5)
    with gr.Row():
        get_examples_btn = gr.Button(value="Get Examples")
        get_examples_btn.click(fn=get_samples,inputs=None, outputs=examples_ds)

if __name__ == "__main__":
  demo.launch()
```

Closes: #7303

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/young-ants-build.md
+++ b/.changeset/young-ants-build.md
@@ -1,0 +1,6 @@
+---
+"@gradio/dataset": patch
+"gradio": patch
+---
+
+fix:Ensure `paginate` updates when samples value changes in `Dataset`


freddyaboulton (approved):
The code changes look good to me @hannahblair ! 

I think a test would be nice but imo this is a small bug so not sure we need a whole e2e just for this functionality.

Issue #7757, opened 2024-03-20
Radio component doesn't work if updated.


GaiZhenbiao:
### Describe the bug

I'm encountering an issue where a Gradio app's Radio button component stops responding to user input after its choices are updated dynamically. Here's a breakdown of the problem:

# Steps to reproduce
Create a Gradio app with a Radio button component and some initial choices.
Include a function that displays a notification based on the selected radio option.
Write code to dynamically alter the choices in the Radio button component.
Try selecting a radio option after the choices have been updated.

# Expected behavior

The notification function should execute, displaying the newly selected value.

# Actual behavior

No notification appears, and the Radio button appears unresponsive.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

def select_llm(llm_name):
    gr.Info(f"Selected {llm_name}")

def update_llm_select():
    return gr.Radio(choices=["LLaMA", "Gemma", "Grok", "Phi"])


with gr.Blocks() as demo:
    radio = gr.Radio(choices=["Claude", "DialoGPT", "GPT-3", "MegaM", "Mitsuku", "Replika", "Simsimi", "Sofia", "Zo"])
    update_radio_btn = gr.Button("Update")

    radio.input(select_llm, radio, [])
    update_radio_btn.click(update_llm_select, [], [radio])

demo.queue().launch()
```

### Screenshot

The reproduction code will pop a notification containing the user-selected value.
<img width="917" alt="image" src="https://github.com/gradio-app/gradio/assets/51039745/b36f8e18-7ccc-46b5-abdc-8f98d7aa15d9">
But if you update the choices of the radio group, it breaks; nothing happens after the user's action, nothing is executed.
<img width="917" alt="image" src="https://github.com/gradio-app/gradio/assets/51039745/4d9eebad-bfa3-4689-b235-ebd7b77cce7a">
There were no warnings or errors in either the Python console or the browser's JavaScript console.

### Logs

```shell
No errors or warnings are displayed in either the Python console or the browser's JavaScript console.
```

### System Info

```shell
Gradio version: 4.22.0
Python version: 3.10
Browser: Arc 1.34.0 (Chromium)
Operating system: macOS Sonoma
```

### Severity

Blocking usage of gradio

freddyaboulton:
This should work with the `select` event @GaiZhenbiao - but I think we can fix the issue with the `input` event.

freddyaboulton:
[...]

PR #7762, opened 2024-03-20
Trigger input event for Radio even when radio is output component

freddyaboulton:
## Description

Closes: #7757

The `input` event was not triggered if the Radio was output by an event. But based on the docstring for `input`, it should be triggered for all user input, which makes it equivalent to `select` event.

Based on the discussion in the issue, it seems like this was working in a previous version. Not sure what could have changed (didn't look into it), but I think this refactor makes sense anyways.

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/mighty-bags-begin.md
+++ b/.changeset/mighty-bags-begin.md
@@ -1,0 +1,6 @@
+---
+"@gradio/radio": patch
+"gradio": patch
+---
+
+fix:Trigger input event for Radio even when radio is output component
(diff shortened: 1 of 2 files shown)


abidlabs:
Makes sense. I was able to repro the original issue and confirm that this fixes it.

abidlabs (approved):

freddyaboulton:
Will check @abidlabs ! Thanks for the review!

PR #7772, opened 2024-03-20
node.js support

cocktailpeanut:
## Description

Currently the gradio js client DOES NOT work at all unless it's used in a browser setting.

The only reason this happens is because it assumes the availability of some browser variables such as `EventSource` and `window`.

I've patched the code so that:

1. `window` only kicks in when the variable exists
2. `setTimeout` does not need to use `window.setTimeout` since node.js natively supports setTimeout as well https://nodejs.org/api/timers.html#settimeoutcallback-delay-args
3. adopted `eventsource` NPM package to make it work both in the browser and in node.js https://github.com/EventSource/eventsource


pngwn:
Unfortunately the client is undergoing significant refactoring at the minute (which has been mentioned in multiple issues and PRs about the client), so we’ll need to close this PR. 

Thank you for taking the time to put this together.

PR #7784, opened 2024-03-21
Fix input event for Number, Slider, Checkbox

freddyaboulton:
## Description

Follow-up from #7762 

The input event for some components does not work if they are returned by an event.
Fixing that in this PR.

To test, you can use the repro from #7762 but change out the component. The input event should trigger after the update button is clicked. Example of `Slider`

```python
def select_llm(llm_name):
    gr.Info(f"Selected {llm_name}")

def update_llm_select():
    return gr.Slider(value=14, label="Code code code")


with gr.Blocks() as demo:
    radio = gr.Slider()
    update_radio_btn = gr.Button("Update")

    radio.input(select_llm, radio, [])
    update_radio_btn.click(update_llm_select, [], [radio])

demo.queue().launch()
```

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/evil-nights-wear.md
+++ b/.changeset/evil-nights-wear.md
@@ -1,8 +1,8 @@
 ---
-"@gradio/checkbox": minor
-"@gradio/number": minor
-"@gradio/slider": minor
-"gradio": minor
+"@gradio/checkbox": patch
+"@gradio/number": patch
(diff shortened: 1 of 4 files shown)


freddyaboulton:
Going to close, I think slider is still messed up

Issue #7774, opened 2024-03-21
gradio-4.22.0 broken ngrok tunneling 


FurkanGozukara:
The images are not displayed in 4.21 or in 4.22

however in 4.19 they are working

first image is 4.22

![image](https://github.com/gradio-app/gradio/assets/19240467/2b638346-981a-4fb3-9858-03aedb674553)

second image is 4.19

![image](https://github.com/gradio-app/gradio/assets/19240467/2e35475f-e1bb-4f19-82a8-84b5c1990183)

abidlabs:
@FurkanGozukara what errors do you see in the console log?

FurkanGozukara:
@abidlabs this is the only error

![image](https://github.com/gradio-app/gradio/assets/19240467/5b286ce7-35ce-4d2d-9e45-4117122a2e86)

![image](https://github.com/gradio-app/gradio/assets/19240467/fa20f6c1-f0fd-48bd-9bb9-ad7129c6bdec)

abidlabs:
Hi @FurkanGozukara I just added some logging to Gradio -- if you install this version of Gradio, it should print some stuff to your terminal. Would you be able to please install and let me know what is printed to your terminal when you run your app?

`pip install https://gradio-builds.s3.amazonaws.com/a3b519f2e78b5ac8acd5b3b6c52f357eed59330e/gradio-4.22.0-py3-none-any.whl`

FurkanGozukara:
> Hi @FurkanGozukara I just [added some logging to Gradio](https://github.com/gradio-app/gradio/pull/7787) -- if you install this version of Gradio, it should print some stuff to your terminal. Would you be able to please install and let me know what is printed to your terminal when you run your app?
> 
> ```
> pip install https://gradio-builds.s3.amazonaws.com/ba3ac976918be4aec67b02baca37459e64537c33/gradio-4.22.0-py3-none-any.whl
> ```

hello worked no issues @abidlabs 

![image](https://github.com/gradio-app/gradio/assets/19240467/b0a91185-baf1-4222-9fed-8c6d5d88f07a)

abidlabs:
Nice! Will get the PR out soon but for not you can keep using that version of Gradio

Issue #7724, opened 2024-03-17
Remaining bug from #7317 - file path resolution wrong


skye0402:
### Describe the bug

With the correction of bug #7317 it's possible again from release 4.21.0 to work behind an application router. That's great. I recently discovered that the file path for pictures in a chatbot interface (at least that, maybe others as well) is not correct. Can you fix it?

Running local:
<img width="1580" alt="image" src="https://github.com/gradio-app/gradio/assets/36907475/753c89f4-eca8-421b-beb1-0a5c46e20d1f">
No issues. As you can see the path looks correct.

Running behind approuter:
<img width="1853" alt="image" src="https://github.com/gradio-app/gradio/assets/36907475/9846cebe-a8b8-46ef-addd-65c07aed16f4">

The issue is the URL. 

Used in the first reference:
`https://apachat.c-290ae5b.kyma.shoot.live.k8s-hana./file=/tmp/gradio/8002ab707d333733cb66b8e9241218143955df1d/20220922063344.jpeg`

Used in the second reference:
`https://apachat.c-290ae5b.kyma.shoot.live.k8s-hana.o/file=/tmp/gradio/8002ab707d333733cb66b8e9241218143955df1d/20220922063344.jpeg`

Correct:
`https://apachat.c-290ae5b.kyma.shoot.live.k8s-hana.ondemand.com/file=/tmp/gradio/8002ab707d333733cb66b8e9241218143955df1d/20220922063344.jpeg`

Somehow the last 2 parts of the domain name are cut off. Interesting is that the first is cut after the dot while the second shows the "o" of "ondemand". Can you check that please?

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

```

### System Info

```shell
Gradio 4.21.0 (upgraded from 4.16.0 where it still worked).
```

### Severity

Blocking usage of gradio

abidlabs:
Thanks @skye0402 that's super weird but I'll look into it. Can you share a quick code snippet sharing how you're supplying the images to the chatbot so that I can repro the same on my end?

abidlabs:
I'm having a bit of a hard time reproducing this. Here's the code I tried:

```py
import gradio as gr

with gr.Blocks() as demo:
    gr.Chatbot([
        ("abc", ("panda.jpg", )),
    ])


demo.launch(root_path="/8044")
```

Running behind a proxy here just fine: http://35.85.61.147/8044/

[...]

PR #7787, opened 2024-03-21
Fix root url resolution from `x_forwarded_host` headers

abidlabs:
Oversight on my part -- if we are using `x_forwarded_host` header to construct the root url, we should not be stripping the trailing characters belonging to the route_url.

I've added a test and the fix has been confirmed by users in https://github.com/gradio-app/gradio/issues/7724 and https://github.com/gradio-app/gradio/issues/7774

Closes: https://github.com/gradio-app/gradio/issues/7774
Closes: https://github.com/gradio-app/gradio/issues/7724


freddyaboulton (approved):

abidlabs:
Thanks @freddyaboulton for reviewing!

rwb-truelime:
Thanks for fixing! 🚀

Issue #7755, opened 2024-03-20
Error while uploading file with certain name


deepanshuc:
### Describe the bug

I am trying to upload this video file to gradio but it throws error.
Apparently gradio throws error on every file with this name no matter what file type(jpg, txt etc).

**Filename:** 
> _Like A Royal Bengal Tiger...__ Mamata Banerjee Responds To JP Nadda.mp4

I checked the tmp upload folder and the file is present in there.
I don't think its a corrupt file because it works if I change the name.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr
import os


def video_identity(video):
    return video


demo = gr.Interface(video_identity, 
                    gr.Video(), 
                    "playable_video")

if __name__ == "__main__":
    demo.launch()

```

### Screenshot

<img width="1840" alt="Screenshot 2024-03-20 at 6 02 28 PM" src="https://github.com/gradio-app/gradio/assets/35916748/cbdfee99-bdec-4037-ad64-aee2afba9438">

### Logs

```shell
Traceback (most recent call last):
  File "/home/vscode/.local/lib/python3.10/site-packages/gradio/queueing.py", line 501, in call_prediction
    output = await route_utils.call_process_api(
  File "/home/vscode/.local/lib/python3.10/site-packages/gradio/route_utils.py", line 253, in call_process_api
    output = await app.get_blocks().process_api(
  File "/home/vscode/.local/lib/python3.10/site-packages/gradio/blocks.py", line 1686, in process_api
    inputs = await anyio.to_thread.run_sync(
  File "/home/vscode/.local/lib/python3.10/site-packages/anyio/to_thread.py", line 56, in run_sync
    return await get_async_backend().run_sync_in_worker_thread(
  File "/home/vscode/.local/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 2144, in run_sync_in_worker_thread
    return await future
  File "/home/vscode/.local/lib/python3.10/site-packages/anyio/_backends/_asyncio.py", line 851, in run
    result = context.run(func, *args)
  File "/home/vscode/.local/lib/python3.10/site-packages/gradio/blocks.py", line 1391, in preprocess_data
    inputs_cached = processing_utils.move_files_to_cache(
  File "/home/vscode/.local/lib/python3.10/site-packages/gradio/processing_utils.py", line 303, in move_files_to_cache
[...]

PR #7795, opened 2024-03-22
Bugfix: `..` in filename throwing error while loading in output.

shubhamofbce:
## Description

We were checking if `..` in file path to prevent path traversal, but we  need to exclude filename while checking this otherwise it will return false if filename has dots in it. So I have updated the function to check path without filename.

Closes: #7755 

## 🎯 PRs Should Target Issues


shubhamofbce:
@abidlabs This is a small bugfix, and ready for review. Let me know if you have any questions.

gradio/utils.py
```diff
@@ -1011,7 +1011,10 @@ def is_in_or_equal(path_1: str | Path, path_2: str | Path):
     """
     path_1, path_2 = abspath(path_1), abspath(path_2)
     try:
-        if ".." in str(path_1.relative_to(path_2)):  # prevent path traversal
+        path_1_without_filename = abspath("/".join(str(path_1).split("/")[:-1]))
```
freddyaboulton, line 1014: I think we should use `path_1.parent` here but not sure if there are any unintended security consequences of this change.

abidlabs: Agree with @freddyaboulton. We should use `path1.parent` instead here. As long as `path1` is a file (which we should check for and raise if not), then I believe it should be fine

shubhamofbce: @abidlabs @freddyaboulton Thanks for the suggestion. I have fixed this.

abidlabs:
Hmm let me take a look at these tests

freddyaboulton:
Nice LGTM @abidlabs !

abidlabs:
Thanks @shubhamofbce for the original PR identifying the cause of the issue and @freddyaboulton for reviewing the fix. Will merge in after CI

PR #7796, opened 2024-03-22
Decrease latency: do not run pre and postprocess in threadpool

freddyaboulton:
## Description

In #7327, we run pre-post process in a separate threadpool to not block the event loop in the event a component has an expensive processing function. This introduces some latency for most demos so it's not worth it. We still process the gallery images in parallel so will prevent prevent the issue with galleries from #7327 from coming up again. In the attached gif, we process 30 images of 1.6 mb (left panel) and 1 mb (right panel) and the latency is minimal and the event loop is not blocked.

do_not_run_threadpool

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -1603,15 +1603,6 @@ Received outputs:
 
         return data
 
-    def run_fn_batch(self, fn, batch, fn_index, state, explicit_call=None):
-        output = []
-        for i in zip(*batch):
-            args = [fn_index, list(i), state]
-            if explicit_call is not None:
-                args.append(explicit_call)
-            output.append(fn(*args))
-        return output
-
     async def process_api(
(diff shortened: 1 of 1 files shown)


abidlabs:
So just to understand, we still postprocess gallery images in separate parallel threads due to this change in #7327 :

image

However, the general case, we do not run `self.postprocess()` in a separate threadpool? 

I tested this out with images of different sizes and didn't notice a significant difference compared to `main` using the original repro with the gallery images. So this should be good!

abidlabs (approved):

freddyaboulton:
> However, the general case, we do not run self.postprocess() in a separate threadpool?

Yes!

freddyaboulton:

Issue #7497, opened 2024-02-21
[BUG Regression] Audio streaming in gradio 4.18.0+ does not  work


pseudotensor:
### Describe the bug

Audio streaming does not work in gradio 4.18.0 or 4.19.0 but works fine in gradio 3.50.2 and up to and including 4.17.0.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

Literal example with only `autoplay=True` added so streaming occurs without clicking anything: https://www.gradio.app/guides/reactive-interfaces#streaming-components

Note that this has a typo `format="bytes"` but that has no effect.

Add some wave file, I added example.

[audio.zip](https://github.com/gradio-app/gradio/files/14352948/audio.zip)

1. unzip zip or edit with your own audio
2. run demo
3. click on first example
4. click stream as bytes

You'll see that streaming via bytes fails to work.  The audio component is changed, but no audio appears and is not played.

```python
import gradio as gr
from pydub import AudioSegment
from time import sleep

with gr.Blocks() as demo:
    input_audio = gr.Audio(label="Input Audio", type="filepath", format="mp3")
    with gr.Row():
        with gr.Column():
            stream_as_file_btn = gr.Button("Stream as File")
            format = gr.Radio(["wav", "mp3"], value="wav", label="Format")
            stream_as_file_output = gr.Audio(streaming=True)

            def stream_file(audio_file, format):
                audio = AudioSegment.from_file(audio_file)
                i = 0
                chunk_size = 1000
                while chunk_size * i < len(audio):
                    chunk = audio[chunk_size * i : chunk_size * (i + 1)]
                    i += 1
                    if chunk:
                        file = f"/tmp/{i}.{format}"
                        chunk.export(file, format=format)
                        yield file
                        sleep(0.5)

            stream_as_file_btn.click(
                stream_file, [input_audio, format], stream_as_file_output
            )

            gr.Examples(
                [["audio/speech.wav", "wav"], ["audio/speech.wav", "mp3"]],
                [input_audio, format],
                fn=stream_file,
                outputs=stream_as_file_output,
                cache_examples=True,
            )

        with gr.Column():
            stream_as_bytes_btn = gr.Button("Stream as Bytes")
            stream_as_bytes_output = gr.Audio(streaming=True, autoplay=True)

            def stream_bytes(audio_file):
                chunk_size = 20_000
                with open(audio_file, "rb") as f:
                    while True:
                        chunk = f.read(chunk_size)
[...]

PR #7797, opened 2024-03-22
This fixes issue #7497 on Chromium based browsers

cschin:
This fixes issue #7497 on Chromium based browsers. For Safari, the http response needs more work to make it working.

## Description

This PR is trying to address issue #7497. 

Closes: #7497 (Only "partially", as it still is not working for Safari)

I think the issues was introduced by commit 68a54a7a310 which was attempting to provide differential streaming for chatbot. After that, the audio element in HTML missed the URL attribute which was generated automatically. In the current main branch, the machinery to add URL from audio data request seems to be missing. The patch passes the URL through. 

## Tests

the PR can run this test script properly https://gist.github.com/cschin/a28f97b95823a35aff47cda761589673 on Edge and Chromo.

### DIFF
--- a/gradio/components/audio.py
+++ b/gradio/components/audio.py
@@ -281,7 +281,7 @@ class Audio(
         if value is None:
             return None, output_file
         if isinstance(value, bytes):
-            output_file["url"]=f"stream/{output_id}"
+            output_file["url"] = f"stream/{output_id}"
             return value, output_file
         if client_utils.is_http_url_like(value["path"]):
             response = httpx.get(value["path"])
@@ -303,7 +303,7 @@ class Audio(
                     )
                 else:
                     binary_data = binary_data[44:]
-            output_file["url"]=f"stream/{output_id}"
+            output_file["url"] = f"stream/{output_id}"
         return binary_data, output_file
 
     def process_example(


freddyaboulton:
Hi @cschin ! Thanks for the contribution. I tested that this works when you stream out a file, but streaming bytes still does not work. Test with the `stream_audio_out` demo in the repo. There are two issues still:

1. The `Stream as Bytes` button does not work
2. The `Stream as File` does play but it eventually hits an error in the console that blocks the whole app

```
AudioPlayer.svelte:64 Uncaught (in promise) TypeError: Cannot set properties of null (setting 'innerHTML')
    at $$self.$$.update (AudioPlayer.svelte:64:22)
    at update (svelte.js:2124:6)
    at flush (svelte.js:2088:5)
```

cschin:
when the format is "bytes", we need to add another line to fill in the url. In my test. this works for the `stream_audio_out` demo for the "Stream as Bytes" and "Stream as Files".  I can't reproduce the failure of the "Stream as Byters" even with a longer 43s wav file I uploaded to the demo.

cschin:
@freddyaboulton  I was able to reproduce the `TypeError: Cannot set properties of null (setting 'innerHTML')` error, but only when I try to use `Stream as File`, however, in my test, the audio still plays. I think that is a different UI bug other than the steaming code issue. Perhaps it needs another ticket for its own.

PR #7801, opened 2024-03-23
Refactor CORS Middleware to be much faster

abidlabs:
Further performance improvements Using @aliabid94's test script with 500 workers and no nginx, this brought down the latency from `72.4` seconds on `main` to `33.1` seconds on this branch, which is essentially the same as having no CORS middleware at all. To do this, I mirrored the implementation from [Starlette's CORS Middlware](https://github.com/encode/starlette/blob/89fae174a1ea10f59ae248fe030d9b7e83d0b8a0/starlette/middleware/cors.py#L15), which aside from providing performance benefits also corrected how CORS was being applied for requests including cookies.

---------

To test that CORS is being applied properly, you can start running any Gradio app, e.g. this one, locally:

```py
import gradio as gr

with gr.Blocks() as demo:
    t1 = gr.Textbox()
    demo.load(lambda :"Hello", None, t1)
    
demo.launch()
```

And then open up different websites in your browser and simulate requests from them using the Inspector console, e.g.:

```js
fetch('http://localhost:7860/config')
  .then(response => {
    if (!response.ok) {
      throw new Error('Network response was not ok');
    }
    return response.json(); // or response.text() if the response is not in JSON format
  })
  .then(data => {
    console.log(data); // This will display the response in the console
  })
  .catch(error => {
    console.error('There has been a problem with your fetch operation:', error);
  });
```

If you make the request from a website on localhost, the request should go through, but if you make the request while a different webpage is loaded, it should be blocked.

I also installed this PR in this Space: https://huggingface.co/spaces/abidlabs/test-cors, and confirmed that:

* The Client works with it:

```py

client = Client("abidlabs/test-cors")
client.predict("Hello")
```

* Embedding with an iFrame works:

```html
<iframe
	src="https://abidlabs-test-cors.hf.space"
	frameborder="0"
	width="850"
	height="450"
></iframe>
```

* Embedding with a web component works:

```html
<html>
<script
	type="module"
	src="https://gradio.s3-us-west-2.amazonaws.com/4.19.1/gradio.js"
></script>

<gradio-app src="https://abidlabs-test-cors.hf.space"></gradio-app>
</html>
```


gradio/route_utils.py
```diff
@@ -648,41 +650,86 @@ def get_hostname(url: str) -> str:
...
 
 
-class CustomCORSMiddleware(BaseHTTPMiddleware):
-    async def dispatch(self, request: fastapi.Request, call_next):
-        host: str = request.headers.get("host", "")
-        origin: str = request.headers.get("origin", "")
-        host_name = get_hostname(host)
-        origin_name = get_hostname(origin)
+class CustomCORSMiddleware:
+    # This is a simplified and modified version of the Starlette CORSMiddleware that only enables CORS when running on localhost.
```
aliabid94, line 654: You mean only enables CORS when **not** running on localhost right?

aliabid94 (approved):
Works when tested, LGTM!

gradio/route_utils.py
abidlabs: I meant when **the host** is running on localhost (that's when we prevent external requests from coming in). Let me clarify

PR #7803, opened 2024-03-23
Fix #7282 Dataset fails to click after load

### LINKED ISSUE #7282: Dataset fails to click after load
### Describe the bug

After updating dataset in fn outputs, old content is retrieved when clicking.
If there are too many or too few rows, an error occurs.

### Reproduction

```python
import gradio as gr

def select(c):
    return c[0] + ' selected.'

def load():
    return [['aaa'],['bbb'],['ccc']]

with gr.Blocks() as demo:
    title = gr.Markdown('# dataset fails to click after load')
    info = gr.Markdown()
    col = gr.Textbox()
    load_button = gr.Button(value='load another dataset')
    ds = gr.Dataset(components=[col], samples=[['xxx'],['yyy']])

    ds.click(
        fn=select,
        inputs=[ds],
        outputs=[info]
    )
    load_button.click(
        fn=load,
        outputs=[ds],
    )

demo.launch()
```

1. Click "load another dataset"
2. Click any Example

- aaa => xxx
- bbb => yyy
- ccc => ERROR

### Screenshot

_No response_

### Logs

_No response_

### System Info

```shell
gradio 4.16.0
```

### Severity

Blocking usage of gradio

aka7774:
## Description

Closes: #7282

There may be a more correct way to implement this.

### DIFF
--- a/gradio/components/dataset.py
+++ b/gradio/components/dataset.py
@@ -147,6 +147,7 @@ class Dataset(Component):
         Returns:
             Returns the updated dataset data as a `dict` with the key "samples".
         """
+        self.samples = samples
         return {
             "samples": samples,
             "__type__": "update",


abidlabs:
Hi @aka7774 this approach will unfortunately not work if multiple users access the same Gradio app. You can try this yourself by opening the Gradio app on multiple tabs in your browser. Once the dataset is updated on one tab, all future users on all other tabs will also see the updated values. 

We need to dig deeper to see why the update logic isn't working for the `gr.Dataset` component. I think it will involve this logic: https://github.com/gradio-app/gradio/blob/c62a57e7f8f2f6dad0110d06e915c48e7f628073/gradio/blocks.py#L1509

Let me close this PR, but feel free to open another one if you are able to look more deeply into this issue.

PR #7807, opened 2024-03-23
fix misleading doc on gr.themes

Mon-ius:
As shown on official [doc](https://www.gradio.app/docs/themes), it makes user confused to understand to have a correct and comprehenssive usage of `gr.themes` to `load`, `export`, etc, to build their own themes.

For example, the correct way to load a local theme `json` file

```python
import gradio as gr

theme = gr.themes.Base.load("src/theme.json")

iface = gr.Interface(
    fn = lambda x:x,
    inputs=gr.Text(
        "Fix the doc"
    ),
    outputs = gr.Text(),
    theme=theme
)

iface.launch()
```


js/_website/src/routes/[[version]]/docs/themes/+page.server.ts
```diff
@@ -27,7 +27,7 @@ export async function load({ parent }) {
 
 	let objs = [docs.building.base];
 	let headers = [
-		["Base", "base"],
+		["themes.Base", "base"],
```
Comment (shubhamrmob) [inline js/_website/src/routes/[[version]]/docs/themes/+page.server.ts:30]: Instead of themes.Base or Base we can just use `Theme`
as you can do gr.Theme.push_to_hub and it looks correct as well.

Comment (shubhamofbce) [inline js/_website/src/routes/[[version]]/docs/themes/+page.server.ts:30]: Instead of themes.Base or Base we can just use `Theme`
as you can do gr.Theme.push_to_hub and it looks correct as well.

Mon-ius:
You are right, the `gr.Theme` can perform similar as `gr.themes.Base`, here to fix the Base related issue.

Mon-ius:
Hi, @shubhamofbce

I just checked the source code that `Theme` just an alias of `Base`, See `from gradio.themes import Base as Theme` in https://github.com/gradio-app/gradio/blob/61cb4fd711b214b85b0cc41b510ed0aabfc65491/gradio/__init__.py#L99

js/_website/src/routes/[[version]]/docs/themes/+page.server.ts
Comment (Mon-ius) [reply js/_website/src/routes/[[version]]/docs/themes/+page.server.ts:30]: r u looking for `from gradio.themes import Base as Theme` 🤗

aliabid94:
@aliabd there seem to be a couple of things off about the docs page for themes, can you take a look?

Mon-ius:
@aliabid94 @shubhamofbce @up_the_irons everything works fine, can u merge it?

abidlabs:
As @aliabid94 pointed out, this doesn't really fix the core issue, which is that the documentation should explain that every theme, including `Base`, is a subclass of `ThemeClass`, and therefore inherits the `.push_to_hub()`, .. methods. As it stands, it seems to suggest that `Base` is some sort of special class. 

We should fix this more comprehensively, so converting this to draft. cc @aliabd perhaps you can take this?

aliabd:
@abidlabs @Mon-ius @aliabid94 You're right the themes page has a lot of issues. Will open a separate PR and clean everything up there. 

I'm going to close this PR because I don't think only changing the side-bar header fixes the problem, and will need to restructure the entire page. Will tag you in the new PR @Mon-ius so you can see the changes.

PR #7811, opened 2024-03-24
Lite playground design changes 

aliabd:
Makes a lot of changes to the lite playground element: 

Screen Shot 2024-03-24 at 4 45 57 AM

- Adds an optional layout attribute to the custom element. If horizontal code and preview will always be side by side, if vertical they'll always be on top of each other. If it isn't provided it will switch at the md breakpoint. 
- Adds copy and download buttons to the code editor
- Fixes issue with CMD+Enter adding a new line 
- Cleans up the header, colors and font sizes. Moved things around a bit and removed the barrier in the middle. Not sure if the run button is in the right place though.. 

Having an issue getting the heights to work right, will spend more time on that.

To test just run 

```
sh scripts/build_lite.sh
python -m http.server --directory js/lite
```
and open `http://localhost:8000/`. you can also edit `js/lite/index.html`


js/app/src/lite/Playground.svelte
```diff
@@ -60,37 +60,55 @@
...
 	function shortcut_run(e: KeyboardEvent): void {
-		if (e.key === "Enter" && (e.metaKey || e.ctrlKey)) {
-			e.preventDefault();
+		if (e.key == "Enter" && (e.metaKey || e.ctrlKey)) {
 			dispatch("code", { code });
+			e.preventDefault();
 		}
 	}
 
+	window.addEventListener("keydown", shortcut_run, true);
```
whitphx, line 69: I think this event listener should be attached to each `<Code />` element, instead of the global window.

With the current setting, it behaves like below and I think it's not intuitive.
* `CMD+Enter` triggers a new run even if the editor is not focused.
* When there are multiple Playground elements, a single `CMD+Enter` triggers a new run on all the Playgrounds.

js/app/src/lite/Playground.svelte
```diff
@@ -213,9 +238,12 @@
 
 	.code-header {
 		padding-top: 0.25rem;
```
whitphx, line 240: This `padding-top` should be removed?
The "app.py" text is vertically aligned to the "Run" text.
<table>
<tr>
<th>With it</th>
<th>Without it</th>
</tr>
<tr>
<td>

</td>
<td>

</td>
</tr>
</table>

whitphx:
I also found when the system is set as dark-mode, the Gradio Playground is initialized in light-mode during the initial load, but becomes dark after the initialization finishes.
CleanShot 2024-03-25 at 16 26 20

abidlabs:
Thanks @aliabd! UI/UX looks great. I don't have much to add beyond what @whitphx pointed out in his review. The only other thing I noticed was that copy button was not working for me. Clicking on it has no effect (does not copy anything to clipboard, no console error logs)

Also the padding around the code buttons maybe should be a bit smaller:

image

Otherwise very nice!

aliabd:
@abidlabs navigator clipboard access is blocked by chrome on localhost. But I tested it and it will work fine otherwise. Will fix the padding

abidlabs:
Ah okay thanks for clarifying @aliabd! Btw I realized we don’t document the playground tag anywhere! We should add it to the Lite Guide?

aliabd:
@abidlabs yup will add to the guide as well. Btw, it's prettier that keeps removing the line breaks in index.html, will try to fix that

aliabd:
@abidlabs @whitphx this is ready for another pass. I fixed everything you pointed out. 

Also added dark mode! 

https://github.com/gradio-app/gradio/assets/9021060/69b8c9ef-3ed0-4bcb-9f62-a9f05723358d

abidlabs (approved):
LGTM @aliabd! Dark mode is slick!

My only UX feedback is that I wish that the code editor did not lose focus when I did CMD + Enter (minor but I think it would be an improvement as it would allow me to keep typing and easily seeing the preview)

abidlabs:
Hmm wait actually I'm seeing something weird -- when I open `http://localhost:8000/`, my memory usage jumps ~15GB and other tabs start to freeze up. Do you see this behavior?

aliabd:
@abidlabs i guess this is what happens when you have three gradio-lite elements on one page.. adding shared-worker reduces the memory load though. Just added it to index.html so at least the problem will be gone there. And we use it on the website already (that's why it doesn't reload lite for every page). 

I agree with you losing focus is annoying, but we can't fix that now because currently running the code reloads the whole playground element. we discussed this a bit here: https://github.com/gradio-app/gradio/pull/7660#discussion_r1520786298 but i think we should revisit later.

js/app/src/lite/Playground.svelte
```diff
@@ -60,65 +60,94 @@
...
+
+	onMount(() => {
+		var code_editors = document.getElementsByClassName("code-editor");
+		for (var i = 0; i < code_editors.length; i++) {
+			code_editors[i].addEventListener(
+				"keydown",
+				shortcut_run as EventListener,
+				true
+			);
+		}
```
whitphx, line 131: This still has a problem that the shortcut key pressed on a single editor will trigger re-running all the mounted playgrounds on the page.
The event listener here should be attached only to the child `<Code />` in this component.

BTW, such "Enter+CMD" shortcut can be `<Code />` component's feature (maybe a new event dispatcher like `run_requested`?), WDYT @pngwn 

I think it can be a separate PR.

whitphx:
@aliabd Thank you, I checked the changes and confirmed it has nice look and feel!
I left one comment. I think it's technically a bug but not critical so we can deal with it later in a separate PR if this PR should go.

abidlabs:
> @abidlabs i guess this is what happens when you have three gradio-lite elements on one page.. adding shared-worker reduces the memory load though. Just added it to index.html so at least the problem will be gone there. And we use it on the website already (that's why it doesn't reload lite for every page).

Got it, thanks for adding this. I tested and confirmed that the memory usage is a lot better now. Can we just make this the default? Is there any downside of using shared-worker by default @whitphx @aliabd?

> I agree with you losing focus is annoying, but we can't fix that now because currently running the code reloads the whole playground element. we discussed this a bit here: https://github.com/gradio-app/gradio/pull/7660#discussion_r1520786298 but i think we should revisit later.

I agree sounds good.

Everything else lgtm!

abidlabs (approved):

guides/08_gradio-clients-and-lite/gradio-lite.md
```diff
@@ -148,6 +148,21 @@ demo.launch()
 
 **Try it out**: You can see this example running in [this Hugging Face Static Space](https://huggingface.co/spaces/abidlabs/gradio-lite-classify), which lets you host static (serverless) web applications for free. Visit the page and you'll be able to run a machine learning model without internet access!
 
+### Code and Demo Playground 
+
+If you'd like to see the code side-by-side with the demo just pass in the `playground` attribute to the gradio-lite element. This will create an interactive playground that allows you to change the code and update the demo! If you're using playground, you can also set layout to either 'vertical' or 'horizontal' which will determine if the code editor and preview are side-by-side or on top of each other (by default it's reposnsive with the width of the page).
```
abidlabs, line 153: > layout to either 'vertical' or 'horizontal'

Unclear how to do this, an example would help. Also there are other attributes, like shared-worker, right? Would be good to document them all

whitphx:
@abidlabs 
> Can we just make this the default? Is there any downside of using shared-worker by default?

Yea, that's an option.
On the other hand, these are what I thought when I made the SharedWorker mode optional, not the default:
* The worker logs are printed in the SharedWorker's console, not the normal developer console. So it's a bit cumbersome for users to get the logs to report errors. In Chrome, for example, users have to open chrome://inspect/#workers to access the SharedWorker's console.
* The SharedWorker mode has some things to know like below, so I thought users should be aware of using the SharedWorker mode by explicitly enabling it.
    * The file system is shared among all the Lite apps using the SharedWorker while each has a separated home dirs, so file operations outside the home e.g. on `/` may conflict.
    * Likewise, the Python runtime is shared, so apps may conflict. For example, if one app monkey-patches some module, another app using that module is affected too.

WDYT?

abidlabs:
Thanks for the info, I agree with keeping shared workers optional then. But let’s add this information to the documentation and make the recommendation that shared workers should be enabled if you have multiple playground instances on a webpage.

whitphx:
Yes, definitely.
Will write it.

whitphx:
Added a section about SharedWorker: #7923

aliabd:
Thanks @whitphx for the changes. I also edited the guide to make layout attribute clearer. 

Will merge this, we can make any other changes in a separate PR.

Issue #7813, opened 2024-03-24
Disabled slider is visually no different from enabled one


Pythongor:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
Disabled slider is visually no different from enabled one. The only difference is cursor on hover (only for slider thumb, on slider track cursor is "pointer" and it is unconsistent).

**Describe the solution you'd like**  
Solve it using css for Slider component, of course

shubhamofbce:
Can you please add a comparison screenshot after the fix here, so we can take a look without running the code.

PR #7814, opened 2024-03-24
Improve UX of non-interactive slider

Pythongor:
## Description

This pull request makes the slider look more logical and compliant with UX standards. Now in the inactive state the slider looks more subdued and the cursor on the hover is displayed correctly.

Closes: #7813

### DIFF
--- a/js/slider/Index.svelte
+++ b/js/slider/Index.svelte
@@ -24,7 +24,7 @@
 	export let scale: number | null = null;
 	export let min_width: number | undefined = undefined;
 	export let minimum: number;
-	export let maximum: number;
+	export let maximum = 100;
 	export let step: number;
 	export let show_label: boolean;
 	export let interactive: boolean;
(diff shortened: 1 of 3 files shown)


shubhamofbce:
@Pythongor  Can you please add a comparison screenshot after the fix here, so we can take a look without running the code.

pngwn:
Looks great!

Screenshot 2024-03-25 at 14 34 26

pngwn:
Fixed the stories and also fixed a small bug that mean the background size wasn't being coloured in when the max was not set.

js/slider/Index.svelte
```diff
@@ -61,11 +61,10 @@
...
 	function setSliderRange(): void {
-		rangeInput.style.backgroundSize =
-			((Number(rangeInput.value) - Number(rangeInput.min)) /
-				(Number(rangeInput.max) - Number(rangeInput.min))) *
-				100 +
-			"% 100%";
+		const dividend = Number(rangeInput.value) - Number(rangeInput.min);
+		const divisor = Number(rangeInput.max) - Number(rangeInput.min);
+		const h = divisor === 0 ? 0 : dividend / divisor;
+		rangeInput.style.backgroundSize = h * 100 + "% 100%";
```
pngwn, line 67: Just added some checks so we don't get infinity.

js/slider/Index.svelte
```diff
@@ -24,7 +24,7 @@
 	export let scale: number | null = null;
 	export let min_width: number | undefined = undefined;
 	export let minimum: number;
-	export let maximum: number;
+	export let maximum = 100;
```
pngwn, line 27: Default for browser is 100 anyway, so just set this explicitly, to avoid undefined values.

pngwn (approved):
Looks great. Thanks for tackling this @Pythongor!

Issue #7714, opened 2024-03-15
ChatInterface rendered inside a Block appears very small in size


yvrjsharma:
### Describe the bug

When the ChatInterface is integrated within a Gradio block, the resulting chat window appears to be significantly smaller in size than expected. This can make the chat interface difficult to use and may cause inconvenience to the user.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

While rendering ChatInterface inside a block:

```python
import gradio as gr
def chat(msg, hist):
    return "ok"

ci = gr.ChatInterface(chat)
with gr.Blocks() as demo:
    ci.render()
demo.launch()
```

Implementing ChatInterface directly:

```python
import gradio as gr
def chat(msg, hist):
    return "ok"

ci = gr.ChatInterface(chat)
ci.launch()
```

### Screenshot

While rendering ChatInterface inside a block:

![image](https://github.com/gradio-app/gradio/assets/48665385/05b8e1b7-6a33-4a4b-923f-f880401ffb37)

Implementing ChatInterface directly:

![image](https://github.com/gradio-app/gradio/assets/48665385/310c4f6c-d864-4849-9693-c92e93082a7a)

### System Info

```shell
Testing on Colab
```

### Severity

I can work around it

dawoodkhan82:
@yvrjsharma I think this is expected.

To fix you can set `fill_height=True`

```
import gradio as gr
def chat(msg, hist):
    return "ok"

ci = gr.ChatInterface(chat)
with gr.Blocks(fill_height=True) as demo:
    ci.render()
demo.launch()
```

yvrjsharma:
Ah, this is perfect @dawoodkhan82 ! Thanks for sharing the solution.

yvrjsharma:
Actually, I am still experiencing issues with the chat window size while customizing a ChatInterface using the chatbot parameter.

```
import gradio as gr
def chat(msg, hist):
    return "ok"

chatbot= gr.Chatbot(placeholder='''<img src="https://raw.githubusercontent.com/gradio-app/gradio/main/readme_files/gradio.svg" style="width:30%; opacity:0.5;">''')
ci = gr.ChatInterface(chat, chatbot=chatbot)
with gr.Blocks(fill_height=True) as demo:
    ci.render()
demo.launch()

```

![image](https://github.com/gradio-app/gradio/assets/48665385/817437a1-09e6-4aad-97a9-4adac1cbd2d5)

[...]

PR #7819, opened 2024-03-25
fix: The meaning of fill_height is reversed at ChatInterface

SeongBeomLEE:
## Description

It says fill_height extends to the height of the window when True

But it actually extends to the height of the window when fill_height is False

Therefore, if fill_height is True, it should be None, and if it is False, it should be 200

Thanks!

### DIFF
--- a/gradio/chat_interface.py
+++ b/gradio/chat_interface.py
@@ -184,7 +184,7 @@ class ChatInterface(Blocks):
                 self.chatbot = chatbot.render()
             else:
                 self.chatbot = Chatbot(
-                    label="Chatbot", scale=1, height=200 if fill_height else None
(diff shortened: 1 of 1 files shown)


abidlabs:
Haven't tested but maybe this would fix: https://github.com/gradio-app/gradio/issues/7714

aliabid94:
Actually the current implementation is correct, even though it looks counter-intuitive. Basically, when fill_height=False, we want the chatbot to be fixed to its default size, which when set to None is specified in the frontend as 400px. But when it is True, we want it to be able to shrink even smaller in case of small windows, down to 200px. 
Agree it could be clearer, thanks for flagging, but will close for now.

Issue #7517, opened 2024-02-23
Memory management for State


aliabid94:
Right now, the amount of memory used for gr.State and hard disk used for files uploaded increases in an unbounded manner. If an app has a large amount of traffic, we will hit an OOM or OOD error. We need to manage memory and disk space used and delete data for older sessions to avoid this issue.

abidlabs:
See: https://github.com/gradio-app/gradio/pull/7447 for disk space (it doesn't cover memory though)

PR #7829, opened 2024-03-25
Delete user state when they close the tab. Add an unload event for the demo and a delete_callback on gr.State to let developers control how resources are cleaned up

freddyaboulton:
## Description

This PR adds some features to address the problem of state being accumulated in the gradio server indefinitely.

Changes:
* Once a user leaves closes the tab, state stored for that session in the server is deleted.
* Add a `time_to_live` parameter to `gr.State`. This is how long the state will be held onto in the server after its most recent creation/update. This is for cases where users minimize the tab and forget about it.
* Add a `delete_callback` parameter to `gr.State` to let developers do any additional clean up actions when state is deleted, e.g. delete tensors from GPU memory. This parameter accepts a function that takes one parameter, the state value.
* Add a `Blocks.unload` method to set an "event" that is triggered when a user closes the page. The goal is to perform any additional clean up not pertaining to gr.State when a user leaves the page. The state is automatically deleted even if no unload event is attached to the demo. This event only accepts one parameter, the function to run. No inputs and outputs are accepted because there is no longer a client connection to receive or send data to.

To test, try the `demo/state_cleanup` or `demo/unload_event_test` demos. 

Closes: #7517 

## 🎯 PRs Should Target Issues

## Tests


gradio/routes.py
```diff
@@ -584,6 +584,55 @@ async def reset_iterator(body: ResetBody):
...
+            background_tasks: BackgroundTasks,
+            username: str = Depends(get_current_user),
+        ):
+            """Clients make a persistent connection to this endpoint to keep the session alive.
+            When the client disconnects, the session state is deleted.
+            It is robust to internet cutting out, etc.
+            """
+
+            async def iterator():
+                while True:
```
freddyaboulton, line 599: I did the experiment where I killed my network connection and the state did not get deleted when I reconnected.

network_kill

freddyaboulton:
cc @pseudotensor  - I think you had asked for automatic state deletion.

gradio/routes.py
abidlabs: Nice! How's that possible though? An asyncio.CancelledError should be raised when the network disconnects right?

pseudotensor:
Nice, I can remove my work-around involving a web socket if this works.  

I'm still waiting for the k8 nginx proxy stuff to be fixed before moving forward with gradio 4 in general.  Also the audio streaming regression isn't fixed yet.

client/python/gradio_client/client.py
```diff
@@ -159,6 +159,7 @@ def __init__(
         self.sse_url = urllib.parse.urljoin(
             self.src, utils.SSE_URL_V0 if self.protocol == "sse" else utils.SSE_URL
         )
+        self.heartbeat_url = urllib.parse.urljoin(self.src, utils.HEARTBEAT_URL)
```
abidlabs, line 162: Nice to include this in the clients as well!

gradio/components/state.py
```diff
@@ -27,13 +27,19 @@ def __init__(
         self,
         value: Any = None,
         render: bool = True,
+        *,
+        time_to_live: int | None = None,
```
abidlabs, line 31:
```suggestion
        time_to_live: int | float | None = None,
```

gradio/components/state.py
```diff
@@ -27,13 +27,19 @@ def __init__(
...
     ):
         """
         Parameters:
             value: the initial value (of arbitrary type) of the state. The provided argument is deepcopied. If a callable is provided, the function will be called whenever the app loads to set the initial value of the state.
             render: has no effect, but is included for consistency with other components.
+            time_to_live: The number of seconds the state should be stored for after it is created or updated. If None, the state will be stored indefinitely. Gradio automatically deletes state variables after a user closes the browser tab or refreshes the page, so this is useful for clearing state for potentially long running sessions.
+            delete_callback: A function that is called when the state is deleted. The function should take the state value as an argument.
         """
-        self.stateful = True
+        self.time_to_live = time_to_live or math.inf
```
abidlabs, line 41: This will not work if time_to_live=0
```suggestion
        self.time_to_live = math.inf if time_to_live is None else time_to_live
```

abidlabs:
It looks like the python unit tests were timing out? Rerunning..

gradio/routes.py
freddyaboulton: Maybe chrome keeps the connections alive but doesn't allow new requests to go through?

abidlabs: Yes that's what I was thinking as well

freddyaboulton:
@abidlabs the issue with the tests was actually caused by the addition of the heartbeat connection in the python client. It was preventing the servers from closing because the connection never closed. Should be fixed now!

abidlabs:
Thanks @freddyaboulton! will do a full review in a bit

client/js/src/client.ts
```diff
@@ -357,6 +357,13 @@ export function api_factory(
 				);
 
 				const _config = await config_success(config);
+				// connect to the heartbeat endpoint via GET request
+				const heartbeat = new EventSource(
+					`${config.root}/heartbeat/${session_hash}`
+				);
+				heartbeat.onmessage = function () {
+					// do nothing
+				};
```
abidlabs, line 366: Do we even need to add this? jw

client/python/gradio_client/client.py
```diff
@@ -184,13 +185,43 @@ def __init__(
...
         self.pending_messages_per_event: dict[str, list[Message | None]] = {}
         self.pending_event_ids: set[str] = set()
 
+    def close(self):
+        self._kill_heartbeat.set()
+        self.heartbeat.join(timeout=1)
+
+    def _stream_heartbeat(self):
+        while True:
+            url = self.heartbeat_url.format(session_hash=self.session_hash)
```
abidlabs, line 206: Will test this but this might lead to lots of cpu consumption. We may need to add a sleep

demo/unload_event_test/run.py
abidlabs, line 1: This demo is not working for me.

1. Every time I generate an image, I see this error in the console:

```
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/starlette/responses.py", line 262, in stream_response
    async for chunk in self.body_iterator:
  File "/Users/abidlabs/dev/gradio-repos/gradio/gradio/routes.py", line 617, in iterator
    done, _ = await asyncio.wait(
              ^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/asyncio/tasks.py", line 415, in wait
    raise TypeError("Passing coroutines is forbidden, use tasks explicitly.")
TypeError: Passing coroutines is forbidden, use tasks explicitly.
```

2. The images are _not_ deleted when I close the tab

abidlabs: Sorry this is regarding `demo/state_cleanup/run.py`.

abidlabs: Actually affecting every demo

freddyaboulton: My bad I didn't realize that syntax would not work in 3.11. Should be fixed now

abidlabs: Python should respect semver and not deprecate stuff like that! Luckily we all use different versions of python

abidlabs: Works beautifully now

client/js/src/client.ts
freddyaboulton: Probably not but will remove if so!

demo/unload_event_test/run.py
```diff
@@ -0,0 +1,34 @@
...
+    with open(log_file, "a") as f:
+        f.write(f"incremented {x}\n")
+    return x + 1, x + 1
+
+def delete_fn(v):
+    with log_file.open("a") as f:
+        f.write(f"deleted {v}\n")
+
+def unload_fn(req: gr.Request):
+   assert req.request
```
abidlabs, line 21: What's the purpose of the `gr.Request` here? For simplicity, maybe better to remove it
```suggestion
def unload_fn():
```

gradio/blocks.py
```diff
@@ -878,6 +882,43 @@ def expects_oauth(self):
             for block in self.blocks.values()
         )
 
+    def unload(self, fn: Callable):
```
abidlabs, line 885: For this event, agree that there is no need for outputs. But in theory, `inputs` could be useful. E.g. if you want to log the final value of a component as the user leaves the page. My suggestion would to have all of the regular parameters but simply warn if a user tries to put in outputs that it will unused.

freddyaboulton: The tricky thing here is that when this is triggered, the client closed the connection so we can't get data from the client anymore

abidlabs:
So I tested out whether losing network connectivity would delete state and trigger unload, and it seems that indeed it does:

Repro:

1. https://huggingface.co/spaces/abidlabs/test-state-delete
2. Go to the demo tab and generate some images
3. Go to the console tab and run `ls` to see that the user directory has been created
4. Disconnect wifi
5. Reconnect wifi 
6. `ls` and you'll see that the directory will not be there
7. Also check the logs and confirm that "STATE DELETED" has been printed

I think this is not a huge problem for the `.unload()` event, we should just document to users that this might get triggered if a user loses network. But losing state could lead to issues for existing demos that use state, e.g. the lmsys chat arena. My suggestion would be instead of deleting the state immediately, we should add an expiry to the state (e.g. 1 hour) and if the state does not get updated in that 1 hour, then we delete it (and we already have the mechanism to expire state!).

gradio/blocks.py
abidlabs: Ah right, makes sense nvm then not needed.

demo/unload_event_test/run.py
freddyaboulton: Just to show that you can pass gr.Request to state but can remove

freddyaboulton:
Very smart suggestion @abidlabs , will work on that!

client/python/gradio_client/client.py
freddyaboulton: Just tested and it will have a minimal impact on CPU. As long as the connection is open, the loop will be held up in `response.iter_lines()` which doesn't take up CPU as messages come every 15 seconds. We return to the top of the loop only when a new session is requested (`.reset_session`)

freddyaboulton:
@abidlabs I just pushed up your suggestion (state deleted 60 minutes after client disconnects) and made other minor tweaks (like highlight changelog)!

.changeset/better-tires-shave.md
```diff
@@ -0,0 +1,42 @@
...
+
+highlight:
+
+#### Automatically delete state when a user closes the browser tab
+
+Gradio now automatically deletes `gr.State` variables stored in the server's RAM when users close their browser tab.
+The deletion will happen 60 minutes after the server detected a disconnect from the user's browser.
+If the user connects again in that timeframe, their state will not be deleted.
+
+Additionally, gradio now runs an `unload` event when users disconnect (this does not have a 60 minute delay).
```
abidlabs, line 15:
```suggestion
Additionally, Gradio now includes a `Blocks.unload()` event, allowing you to run cleanup jobs when users disconnect (this does not have a 60 minute delay).
```

.changeset/better-tires-shave.md
```diff
@@ -0,0 +1,42 @@
+---
+"@gradio/client": minor
+"gradio": minor
+"gradio_client": minor
+---
+
+highlight:
+
+#### Automatically delete state when a user closes the browser tab
```
abidlabs, line 9: Super nit: maybe we should consider rewording in light of the fact that the deletion is not immediate
```suggestion
#### Automatically delete state after user has disconnected from the webpage
```

abidlabs (approved):
Super cool PR! LGTM @freddyaboulton

freddyaboulton:
Thanks for the review @abidlabs !!!

DarrenZhaoFR:
> @abidlabs I just pushed up your suggestion (state deleted 60 minutes after client disconnects) and made other minor tweaks (like highlight changelog)!

Thanks for the PR! Is it possible to customize the 60-minute deletion period(for gr.State)? @freddyaboulton 
Possiable use case: 
1. When debugging, set the deletion period to 0.1 minute just to make sure it works

freddyaboulton:
> Thanks for the PR! Is it possible to customize the 60-minute deletion period for gr.State?

Something I've thought about. Yes I think this makes sense.

Issue #4175, opened 2023-05-11
More documentation is needed on how to use `gradio_client` with multi-turn dialogue with history


lychees:
- [x] I have searched to see if a similar issue already exists.

I am trying to use ChatGLM with gradio_client to do some research, but soon found that it doesn't support multi-turn dialogue like the original app. Below is my implementation:

```
def predict(input, history=None):
    if history is None:
        history = []
 
    client = Client('https://multimodalart-chatglm-6b.hf.space/')
    with open(client.predict(input, fn_index=0)) as f: 
        text = process_text(f.read())
        output = json.loads(text)[0]
        history += [output]
        return history, history
```

Since we remove state from the API recently, [4107](https://github.com/gradio-app/gradio/pull/4107), I thought it could be impossible to support multi-turn dialogue, is that true?

abidlabs:
I think we can document this better, so I'll reframe this issue around that

PR #7830, opened 2024-03-25
Add guide on Client state and and fix default values of components

abidlabs:
Fixes: https://github.com/gradio-app/gradio/issues/4175

Also sets the default values of several components to `None` instead of an empty string. This makes no difference in the Gradio app itself, but it makes a difference when connecting from the client as any non-None value means that the parameter is not required, which can be confusing if a user did not intentionally specify a non-None value.

### DIFF
--- a/guides/08_gradio-clients-and-lite/01_getting-started-with-the-python-client.md
+++ b/guides/08_gradio-clients-and-lite/01_getting-started-with-the-python-client.md
@@ -295,3 +295,28 @@ job = client.submit("abcdef")
 time.sleep(3)
 job.cancel()  # job cancels after 2 iterations
 ```
+
+## Demos with Session State
+
+Gradio demos can include [session state](https://www.gradio.app/guides/state-in-blocks), which provides a way for demos to persist information from user interactions within a page session.
+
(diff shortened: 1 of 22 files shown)


guides/08_gradio-clients-and-lite/01_getting-started-with-the-python-client.md
```diff
@@ -295,3 +295,40 @@ job = client.submit("abcdef")
...
+    textbox = gr.Textbox()
+    number = gr.Number()
+    textbox.submit(count, inputs=[textbox, words], outputs=[number, words])
+    
+demo.launch()
+```
+
+If you were to connect this this Gradio app using the Python Client, you would notice that the API information only shows a single input and output:
+
+```text
```
aliabd, line 322: if you want a codeblock with no syntax highlighting (like terminal output) use csv

aliabd:
```suggestion
```csv
```

aliabd (approved):
LGTM with one comment

guides/08_gradio-clients-and-lite/01_getting-started-with-the-python-client.md
abidlabs: Nice

PR #7835, opened 2024-03-26
Stop running iterators when js client disconnects

freddyaboulton:
## Description

If you run an iterator event and close the page the python function will keep running in the backend.

To test, run the following demo, click the button and close the page. Things will still be being printed to the console in main. On this branch, they do not. Also added a playwright test for this.

```python
    def iterate():
        for i in range(8):
            print(f"UPDATE: {i}")
            yield i
            time.sleep(0.5)

    with gr.Blocks() as demo:
        with gr.Row():
            c_time2 = gr.Textbox(label="Iteration Step")
            btn = gr.Button("Start")
            btn.click(iterate, None, c_time2, api_name="start")


   demo.launch()
```

## 🎯 PRs Should Target Issues

## Tests


demo/cancel_events/run.py
abidlabs, line 1: Let's revert these changes?

abidlabs: Ah got it! We don't use this demo in any of our docs so it should be fine

demo/cancel_events/run.py
```diff
@@ -1,10 +1,15 @@
...
+import pathlib
 
+log_file = (pathlib.Path(__file__).parent / "cancel_events_output_log.txt").resolve()
 
 def fake_diffusion(steps):
+    log_file.write_text("")
     for i in range(steps):
         print(f"Current step: {i}")
+        log_file.write_text(f"Current step: {i}\n")
         time.sleep(0.5)
```
abidlabs, line 13: Could we make this faster? So that we could reduce the time that we have to wait in the test
```suggestion
        time.sleep(0.2)
```

gradio/routes.py
```diff
@@ -803,11 +803,11 @@ async def sse_stream(request: fastapi.Request):
                         message=str(e),
                     )
                     response = process_msg(message)
-                    if response is not None:
```
abidlabs, line 806: Ah so we already had the deletion logic in place, but it was never called?

js/app/test/cancel_events.spec.ts
```diff
@@ -44,3 +45,26 @@ test("when using an iterative function it should be possible to cancel the funct
...
+	// wait for the duration of the entire iteration
+	// check that the final value did not get written
+	// to the log file. That's our proof python stopped
+	// running
+	await new Promise((resolve) => setTimeout(resolve, 5000));
+	const data = readFileSync(
+		"../../demo/cancel_events/cancel_events_output_log.txt",
+		"utf-8"
+	);
+	expect(data).not.toContain("Current step: 8");
```
abidlabs, line 69: Let's also ensure that at least one of the steps that should be logged are logged
```suggestion
	expect(data).not.toContain("Current step: 1");
	expect(data).not.toContain("Current step: 8");
```

gradio/routes.py
freddyaboulton: Yea we yielded the response before the clean up logic happened so we never reached that point in the code

abidlabs: Nice catch

abidlabs (approved):
LGTM with a few suggestions. Cool test!

js/app/test/cancel_events.spec.ts
freddyaboulton: Reasonable!

freddyaboulton:
Thanks for the review @abidlabs ! I addressed your comments so will merge once ci passes

PR #7845, opened 2024-03-26
ensure `ImageEditor` events work as expected

### LINKED ISSUE #6757: ImageEditor.upload is not triggered at all.
### Describe the bug

ImageEditor.upload is not triggered at all.

### Reproduction

```python
import gradio as gr
import time


def sleep():
    print('succ')

with gr.Blocks() as demo:
    im = gr.ImageEditor(
    )
    im.upload(sleep, outputs=[], inputs=[])

demo.launch(share=True)

```

### Screenshot

_No response_

### Logs

_No response_

### System Info

```shell
Linux
```

### Severity

Blocking usage of gradio

### LINKED ISSUE #6827: gr.ImageEditor.change(fn,...) can only be triggered one time when the value parameter is passed
### Describe the bug

gr.ImageEditor.change(fn,...) can only be triggered one time when the value parameter is passed. After that, nothing happens. Also, when the event listener is triggered the first time, the drawn mask disappears from the ImageEditor component

### Reproduction

```python
import gradio as gr

def draw_mask(input):

    print('I was triggered')

    return input["layers"][0]

import gradio as gr

with gr.Blocks() as demo:
    img = gr.ImageEditor(
        value="https://people.sc.fsu.edu/~jburkardt/data/png/dragon.png",
        interactive=True
        )
    output_mask = gr.Image()


    img.change(fn=draw_mask,inputs=img,outputs=output_mask)

    demo.launch(height=1324)

```

### Screenshot

_No response_

### Logs

_No response_

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Linux
gradio version: 4.10.0
gradio_client version: 0.7.3

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.1.2
fastapi: 0.103.2
ffmpy: 0.3.1
gradio-client==0.7.3 is not installed.
httpx: 0.25.0
huggingface-hub: 0.19.4
importlib-resources: 6.1.0
jinja2: 3.1.2
markupsafe: 2.1.3
matplotlib: 3.8.0
numpy: 1.26.0
orjson: 3.9.7
packaging: 23.1
pandas: 2.1.1
pillow: 10.0.1
pydantic: 2.4.2
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0.1
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.9.0
typing-extensions: 4.7.1
uvicorn: 0.23.2
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2023.9.2
httpx: 0.25.0
huggingface-hub: 0.19.4
packaging: 23.1
typing-extensions: 4.7.1
websockets: 11.0.3
```

### Severity

Blocking usage of gradio

### LINKED ISSUE #7098: .change() listener not working as intended for ImageEditor (Paint, Image, ImageEditor, Sketchpad)
### Describe the bug

I am trying to use sketchpad and using the .change listener so every time I draw anything and let go, it runs a function. This used to work just fine before (v 3.50.2) but is broken now where I have to press this "save changes" button every single time I want it to run. 

I am not sure if this is a bug or intended and if it's intended, then what listener event can I use to perform a real-time change? I have pasted a basic function where everything i draw gets returned as is but it should be doing it automatically while im drawing which is not happening.

As you can tell in the image, the green checkmark is what im having to press each time.
image

### Reproduction

```python
import gradio as gr

async def gen(image):
    return image["composite"]

with gr.Blocks() as demo:
    with gr.Row(variant="default"):
        with gr.Column():
            #img = gr.Sketchpad(height="512", width="1024", type="pil", scale=1, brush=gr.Brush(color_mode="defaults")) # type: ignore
            img = gr.Paint(image_mode="RGB",type="pil", interactive=True)
        with gr.Column():
            sktout = gr.Image(
                show_label=False,
                type="pil",
                interactive=False,
                height=512,
                width=768,
                elem_id="sktout",
            )
    img.change(fn=gen, inputs=[img],outputs=[sktout], show_progress="hidden")

demo.queue().launch(inline=False, share=False, server_name="0.0.0.0")

```

### Screenshot

_No response_

### Logs

_No response_

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Linux
gradio version: 4.15.0
gradio_client version: 0.8.1

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.0.1
fastapi: 0.102.0
ffmpy: 0.3.1
gradio-client==0.8.1 is not installed.
httpx: 0.24.1
huggingface-hub: 0.20.2
importlib-resources: 6.0.1
jinja2: 3.1.3
markupsafe: 2.1.3
matplotlib: 3.8.2
numpy: 1.24.1
orjson: 3.9.5
packaging: 23.2
pandas: 2.0.3
pillow: 9.3.0
pydantic: 2.2.1
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0
ruff: 0.1.14
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.9.0
typing-extensions: 4.9.0
uvicorn: 0.23.2
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2023.12.2
httpx: 0.24.1
huggingface-hub: 0.20.2
packaging: 23.2
typing-extensions: 4.9.0
websockets: 11.0.3
```

### Severity

Blocking usage of gradio

pngwn:
## Description
This PR reworks some of the `ImageEditor` events:

- `change` now triggers in realtime as expected (every stroke)
- `input` now triggers in realtime as expected (every stroke)
- `upload` now triggers when a background image is added via any method (upload, webcam, paste)
- `apply` is a new event that replaces the old `change` event. When this event is used a tick button is added to the UI, clicking the button triggers the event.

In order to support realtime events some changes were made to the `/component_server` route to support binary data and a server function was added to the `ImageEditor`. I also added some functional tests for the component.

Closes #6757. Closes #6827. Closes #7098.


js/imageeditor/shared/InteractiveImageEditor.svelte
```diff
@@ -102,11 +107,82 @@
...
+			if (layer)
+				images.push([
+					id as string,
+					`layer`,
+					new File([layer], `layer_${i}.png`),
+					i
+				]);
+		});
+
+		await Promise.all(
```
freddyaboulton, line 167: I think you wrote `component_server` to accept multiple files in the same request. Would that be better than uploading separately?

freddyaboulton (commented):
Very nice PR @pngwn ! Left some comments. Big things though are that the webcam view does not work for me

image

and I also can't paste from the clipboard (tried on chrome)

gradio/components/image_editor.py
```diff
@@ -195,6 +222,7 @@ def __init__(
         self.transforms = transforms
         self.eraser = Eraser() if eraser is None else eraser
         self.brush = Brush() if brush is None else brush
+        self.blob_storage: dict[str, EditorDataBlobs] = {}
```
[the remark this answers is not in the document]
pngwn: > Returning a new image editor for an update will cause this to be reset so in-progress sessions will get an error.

That seem correct apart from the in_progress error part. I could add some kind of guard for that if you think it is important but this data does belong to the instance i think.

.changeset/silent-pianos-smoke.md
```diff
@@ -0,0 +1,8 @@
+---
+"@gradio/app": minor
+"@gradio/client": minor
+"@gradio/imageeditor": minor
+"gradio": minor
+---
+
+feat:ensure `ImageEditor` events work as expected
```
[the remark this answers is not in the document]
pngwn: I was going to hopefully get a few image editor improvements in before the next release, and create one highlight for all of them.

gradio/image_utils.py
```diff
@@ -48,18 +48,18 @@ def format_image(
...
+    y: np.ndarray | PIL.Image.Image | str | Path, cache_dir: str, format: str = "webp"
 ):
-    # numpy gets saved to png as default format
-    # PIL gets saved to its original format if possible
     if isinstance(y, np.ndarray):
         path = processing_utils.save_img_array_to_cache(
             y, cache_dir=cache_dir, format=format
         )
     elif isinstance(y, PIL.Image.Image):
+        fmt = y.format if y.format is not None and y.format != "png" else "webp"
```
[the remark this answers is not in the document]
pngwn: I specifically wanted to coerce png to webp because of the performance improvements. If we don't have some kind of handling here then it is will use the existing file format which will always be `png` for images generated in the browser.

js/imageeditor/shared/InteractiveImageEditor.svelte
pngwn: I did but this is faster and simpler since they can be sent in parallell. I just wanted the `component_server` stuff to work with whatever is thrown at it.

freddyaboulton:
@pngwn I think the webcam input display is still messed up. It doesn't remove the upload text.

webcam_switch

And I still can't get the copy from clipboard to work. The input event is triggered even with programmatic updates, like hitting a button to clear the value of the image editor. This makes it identical to change, is that expected?

Other than those points, this looks good to me!!

abidlabs:
I can feel the difference in speed switching to webp! We should document in this PR that webp is now the default format for all image components

abidlabs:
> The input event is triggered even with programmatic updates, like hitting a button to clear the value of the image editor. This makes it identical to change, is that expected?

I don't see this bug @freddyaboulton. For example, in this demo:

```py
import gradio as gr

with gr.Blocks() as demo:
    with gr.Row():
        i1 = gr.ImageEditor()
        i2 = gr.ImageEditor(interactive=True)
        i3 = gr.ImageEditor(interactive=True)   
        i1.input(lambda x:x, i1, i2)
        i2.input(lambda x:x, i2, i3)
        
demo.launch()
```

when I make changes to the first image editor, only the second one is changed, the third one is not:

image

abidlabs:
> And I still can't get the copy from clipboard to work.

Is it working for you in the deployed Space? https://huggingface.co/spaces/gradio-pr-deploys/pr-7845-all-demos

Based on your browser settings, chrome(ium) might be blocking pasting image into localhost. (It is working for me both in the deployed Space and localhost)

abidlabs:
I *am* able to repro that webcam upload looks weird: 

image

abidlabs:
The only other issue that I noticed while testing was that the progress bar overlays the upload to text when the imageeditor is used as an output:
 
image

Otherwise looks good to me @pngwn, thanks for adding the tests!

gradio/components/image_editor.py
```diff
@@ -83,6 +104,10 @@ def __post_init__(self):
             )
 
 
+class EditorId(GradioModel):
```
abidlabs, line 107: Used anywhere?

gradio/components/image_editor.py
```diff
@@ -213,20 +241,27 @@ def __init__(
...
-        im = PIL.Image.open(file.path)
-        if file.orig_name:
+        im = (
+            PIL.Image.open(file.path)
+            if isinstance(file, FileData)
+            else PIL.Image.open(BytesIO(file))
+        )
+        if isinstance(file, (bytes, bytearray, memoryview)):
+            name = "image"
+            suffix = "webp"
```
abidlabs, line 255: This should just use `self.format`?

gradio/components/image_editor.py
```diff
@@ -213,20 +241,27 @@ def __init__(
...
+        elif file.orig_name:
             p = Path(file.orig_name)
             name = p.stem
             suffix = p.suffix.replace(".", "")
             if suffix in ["jpg", "jpeg"]:
                 suffix = "jpeg"
         else:
             name = "image"
-            suffix = "png"
+            suffix = "webp"
```
abidlabs, line 264: Same

gradio/components/image_editor.py
```diff
@@ -95,8 +120,10 @@ class ImageEditor(Component):
     EVENTS = [
         Events.clear,
         Events.change,
+        Events.input,
         Events.select,
         Events.upload,
+        Events.apply,
     ]
     data_model = EditorData
```
abidlabs, line 129: We should add a `format` param instead of hardcoding to webp:
```
        format: str = "webp",
```

gradio/components/image_editor.py
```diff
@@ -289,7 +349,7 @@ def postprocess(self, value: EditorValue | ImageType | None) -> EditorData | Non
                     path=image_utils.save_image(
                         cast(Union[np.ndarray, PIL.Image.Image, str], layer),
                         self.GRADIO_CACHE,
-                        format="png",
+                        format="webp",
```
abidlabs, line 352: Same, and also below

freddyaboulton:
> I don't see this bug @freddyaboulton. For example, in this demo:

What I saw was that clearing the ImageEditor programmatically would also trigger the input event

image_editor_change

I see the same behavior in your demo!

freddyaboulton:
My chrome is being weird, I'm giving it permission and the copy still does not do anything. 

image_editor_copy

I'll try reinstalling in the meantime but if I'm the only one then don't worry about it.

I did notice that the `Layers` button creeps into the other buttons when the screen size changes. I think that should be fixed if you can reproduce that.

image_editor_shrink

gradio/components/image_editor.py
pngwn: I don't know that a format param really makes sense here, the images are all either generated by or used by the image editor internally.

The only time this would ever have any impact is if a user is using the `filepath` type. In all other cases the images are either utilised by the imageeditor frontend internally or converted to np or PIL images before the users sees them.

It will also potentially undo the added performance that webp brings. I personally think we should require this to be webp. In the rare cases where a user definitely needs an image file and that files needs to be something other than webp, it is easier for them to convert it.

gradio/components/image_editor.py
pngwn: No longer! Will remove.

pngwn:
@freddyaboulton 
> The input event is triggered even with programmatic updates, like hitting a button to clear the value of the image editor. This makes it identical to change, is that expected?

I'll double check this. It doesn't necessarily make it identical to change though, `input` is any change in response to a UI action. `change` is any change in data regardless of source.

For example, it is expected that clicking the `x` clear icon in the `ImageEditor` would trigger the input event. The clear button that 'programatically' clears the input is definitely an interesting case. I would actually expect that _not_ to trigger the `input` even so i will check. But i think that is a discrete bug rather than `input` and `change` events triggering at the same time in all cases.

@freddyaboulton 
> I did notice that the Layers button creeps into the other buttons when the screen size changes. I think that should be fixed if you can reproduce that.

I'll fix this in a follow up because I'm changing the icon UI in another PR I have in progress.

@abidlabs 
> the progress bar overlays the upload to text when the imageeditor is used as an output:

I'll probably look at this in a followup. Nothing should have changed in this PR regarding this (although ImageEditor outputs may have been completely broken). I'd like to tackle all of the 'output' issues together in a single PR (and add some tests).

I am seeing the webcam UI issue, will fix that. I'll try to reproduce the paste issue.

pngwn:
@freddyaboulton @abidlabs Webcam thing should be fixed. I can't reproduce the image paste issue in any browser that supports the Clipboard API, even if i reset all permissions.

Still need clarification on the `format` kwarg because I'm not sure it would do anything other than cause problems.

gradio/components/image_editor.py
abidlabs: @pngwn there are two cases I see where a user might want to use the `format` parameter:

(1) as you said, if the `ImageEditor` component is being used as an input component with `type="filepath"`. 

(2) If the `ImageEditor` is being used as an output component. Currently, the `ImageEditor` as an output component is exactly the same as the `Image` component, but in the future, we might want to expose the different layers as well. In this case, users might want to be able to set the format as well.

In both cases, while it is true that other formats are less performant, we should set the default to `webp` and then let users change it if needed -- this would be consistent with other components, such as `Image` or `Video`

abidlabs:
@pngwn I responded above about the format. 

Webcam issue is fixed on my end, although the background color should be white in light mode (currently black):

image

gradio/components/image_editor.py
pngwn: That makes sense. Will add the kwarg.

freddyaboulton (approved):
Thanks @pngwn ! All of my comments have been addressed. Tested and works well.

Issue #7128, opened 2024-01-23
Automatically provide the Client code to reconstruct user interaction in a Gradio app


abidlabs:
As mentioned [here](https://twitter.com/artificialguybr/status/1749891790533013556), the request is to make it possible to automatically convert user interaction in a Gradio app to code using the Clients:

> Gradio could help by automatically providing the API call for performing a specific action (which often involves manipulating multiple text boxes

jvkap:
Yes. Exactly that. It would be very interesting. 

You would act as if you were doing a workflow/action normally using Gradio. And maybe it could show in the console or network f12 how to execute that exact workflow/action using the gradio_client api directly without the need to open the GUI. We have similar things on Replicate, ComfyUI Workflow...

Often the actions are quite complex involving dozens of different buttons, selections. This would make it especially easy for laymen and not hard-devs.

PR #7850, opened 2024-03-27
Adds an "API Recorder" to the view API page, some internal methods have been made async

abidlabs:
This PR adds an "API Recorder" to Gradio's View API page, which should make it easier for developers to use Gradio / Spaces via API.

Closes: https://github.com/gradio-app/gradio/issues/7128


abidlabs:
Still need to add docs, but otherwise should be ready for review. Addressed the comments from sync @aliabd @dawoodkhan82!

dawoodkhan82 (commented):
Works great, with the few demos I tested. One nit in terms of UI, it isn't quite clear how you can view your recorded api calls. Maybe adding "Click here to view the recorded API call" or something along those lines would help.
Screenshot 2024-03-29 at 1 39 50 AM

aliabd (approved):
Works great! nice that you got the python snippets correct. 

I have a few nits on the design: 

- This doesn't really look like a button: 

Screen Shot 2024-03-29 at 5 18 38 AM

We should just copy the design of the gradio buttons.

- I hate this black square... 

Screen Shot 2024-03-29 at 5 18 47 AM

Why not do something like I did for the loading status on lite playground 

https://github.com/gradio-app/gradio/assets/9021060/8953fd77-dbbf-437f-9759-d83a693704f9


But both of these are minor, and if you want I can take another pass on a separate PR after.

freddyaboulton:
Nice PR @abidlabs ! I found a tricky corner case. I tried the chatbot multimodal demo and uploaded a file and submitted it. This is the api call the recorder gave me

```python
from gradio_client import Client, file

client = Client("http://127.0.0.1:7860/")

client.predict(
  history=[],
  message={"text":"Hello","files":[file('http://127.0.0.1:7860/file=/private/var/folders/t1/j7cmtcgd0mx43jh9nj_r9mmw0000gn/T/gradio/c07fe25f20271b76caa3103105d3fb0158760d27/3262_2.png')]},
  api_name="/add_message"
)
```

If I run that, the request will fail and the server will stop raise an httpx timeout error.

I think what's happening is that `move_resource_to_block_cache` is trying to download the file contents from that URL but the `file=` route is actually blocked because `move_resource_to_block_cache` is waiting for it to finish. I wonder if making `move_resource_to_block_cache` async will fix it.

This behavior is not caused by this PR but we should fix as people will run into it.

abidlabs:
Interesting nice catch @freddyaboulton! Let me look into this and see if I can include a fix in this PR

abidlabs:
> This doesn't really look like a button:

> We should just copy the design of the gradio buttons.

Fixed. I used the `<Button>` component, it looks kinda interesting. Not as seamless as I was going for, but at least it'll bring visibility to the feature:

image

> I hate this black square...
> Why not do something like I did for the loading status on lite playground

That'd be very slick! I couldn't figure out how to create that, so I've replaced it with a lame orange circle for now:

image

abidlabs:
> Nice PR @abidlabs ! I found a tricky corner case. I tried the chatbot multimodal demo and uploaded a file and submitted it. This is the api call the recorder gave me

This issue is actually trickier than I thought. I can't think of a way to solve this without something hacky, like including an additional field in the FileData that the client populates telling the gradio server that this file is on the same server as the gradio app, so do not try to download it and move to cache. The problem is that the Gradio app doesn't know its own public url, so it doesn't know when to try downloading & moving an image to cache and when it does not need to. 

I'll implement it this way -- if you can think of a cleaner way @freddyaboulton, maybe you can refactor accordingly when you get a chance.

abidlabs:
Actually going to think about this some more. In the meantime @aliabd @dawoodkhan82 if you would like to make any UI changes, feel free to push directly to this branch.

abidlabs:
Ok so I've converted `processing_utils.save_url_to_cache` to async, which in turn required converting `blocks.preprocess_data` and `blocks.post_process_data` to async as well. This might give us some performance benefits as well (e.g. if the value of a component consists of lots of large urls that need to be downloaded) similar to https://github.com/gradio-app/gradio/pull/7327

Because this involves changing a number of files, would be great if you could give this another look @freddyaboulton. Feel free to merge in if everything looks good

client/python/gradio_client/utils.py
```diff
@@ -841,20 +841,36 @@ def set_space_timeout(
 ########################
 
 
-def synchronize_async(func: Callable, *args, **kwargs) -> Any:
+def synchronize_async(func, *args, **kwargs) -> Any:
```
abidlabs, line 844: I benchmarked performance to see if this change would degrade, but it does not. cc @aliabid94 

image

abidlabs:
Tagging @whitphx for the gradio/lite e2e tests timing out -- all the other tests are passing

whitphx:
The direct reason of the E2E test failure (timeout) is that the Lite app didn't boot up successfully with an error caused by the [change on `synchronize_async()`](https://github.com/gradio-app/gradio/pull/7850/files#diff-5557f3070c89abd91ce70a262a2f7df99867027634a451c96102392e4d81d9d2R844).
In the Wasm/Pyodide env, both `fsspec.asyn.sync()` and `threading` don't work due to it's nature where `threading` is not supported and blocking code is not allowed.
Until now, ALAIK, we have avoided using this method in Lite nicely, so we haven't seen this error.

Just to avoid the error, [this workaround](https://github.com/gradio-app/gradio/pull/7903) works.
However, the E2E test still fails due to another reason:
With this workaround, `synchronize_async` doesn't await the coroutine and its returned value is missed, so some necessary values are discarded. As a result, the button label disappears as the screenshot below and E2E test fails because they can't find the button in the test suites.
CleanShot 2024-04-01 at 19 24 22@2x

@abidlabs If it's urgent, please go ahead without solving the Lite problem, commenting out the E2E test and marking `@gradio/lite` private not to release it until this problem is solved.

I want to propose that we will make all the method async-oriented and get rid of `synchronize_async()`.
Currently, the code around the web server (FastAPI) and the queue is async and the leaf methods are also async like this PR, but some middle methods are synchronous so `synchronize_async` is needed. I think `synchronize_async` can be removed (or at least be avoided to use in the Wasm case) by making such sync methods async-oriented, e.g. pre- and postprocess.
Async code is rather Pyodide-friendly as the Wasm env is eventloop-based, so incresing the amount of async code is a very good thing to Lite as well in principle.

aliabid94:
It broken for me currently with the blocks_xray demo:

Recording 2024-04-01 at 16 01 21

abidlabs:
Yeah sorry I pushed some WIP stuff that broke it. Let me revert to draft

gradio/helpers.py
```diff
@@ -353,8 +353,7 @@ def _start_caching(self):
                 warnings.warn(
                     "Setting `cache_examples=True` is not supported in the Wasm mode. You can set `cache_examples='lazy'` to cache examples after first use."
                 )
-            else:
-                client_utils.synchronize_async(self.cache)
+            await self.cache()
 
     def lazy_cache(self) -> None:
```
freddyaboulton, line 358: I think `lazy_cache` can also be made a co-routine?

freddyaboulton (commented):
This PR is pretty much ready to go @abidlabs ! Noticed a couple of small issues in the generated python calls:

* If the function has no arguments, the call will be `client.predict(, api_name="/...")` which is not valid syntax (empty comma)
* The python docs show the javascript `null` as opposed to `None` so the python calls can't be used as-is.

Would appreciate if @aliabd and @dawoodkhan82 can take a closer look at the UI and js!

gradio/helpers.py
abidlabs: Done

abidlabs:
@freddyaboulton both issues have been fixed now. Will go ahead and merge in this PR if CI passes.

Issue #7860, opened 2024-03-27
DowloadButton callback not triggering


joel-odlund:
### Describe the bug

The DownloadButton is not triggering the callback for the `click` event. 
When clicking on the button,  the code in the callback is not run. 
In the example, expect the label to change when pressed. It does not,  and 
debugging the code indicates that the callback is not run. 

I am not sure if this is intended or a bug.  The examples indicate that this is the correct usage.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

def callback():
    return  gr.DownloadButton(
        label="Download pressed",
        value="file.csv")

with gr.Blocks() as app:
    dl = gr.DownloadButton(label="Download",value = None )
    dl.click(callback, inputs = None, outputs=dl)

app.launch()

```

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Darwin
gradio version: 4.20.0
gradio_client version: 0.11.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.2.0
fastapi: 0.110.0
ffmpy: 0.3.2
gradio-client==0.11.0 is not installed.
httpx: 0.27.0
huggingface-hub: 0.22.1
importlib-resources: 6.4.0
jinja2: 3.1.3
markupsafe: 2.1.5
matplotlib: 3.8.3
numpy: 1.26.4
orjson: 3.9.15
packaging: 24.0
pandas: 2.2.1
pillow: 10.2.0
pydantic: 2.6.4
pydub: 0.25.1
python-multipart: 0.0.9
pyyaml: 6.0.1
ruff: 0.3.4
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.11.0
typing-extensions: 4.10.0
uvicorn: 0.29.0
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.
```

### Severity

Blocking usage of gradio

joel-odlund:
👍

PR #7862, opened 2024-03-27
Trigger click event from `gr.DownloadButton` even when no file is present

abidlabs:
Very quick fix: https://github.com/gradio-app/gradio/issues/7860


freddyaboulton (approved):

abidlabs:
Thanks @freddyaboulton for the ⚡ review!

Issue #7424, opened 2024-02-14
Lazy cache examples


abidlabs:
Instead of caching the examples before the Gradio app is launched, there should be a way to lazily cache the examples, i.e. cache them when after they are first processed. This has several advantages, as @apolinario pointed out:

* App becomes instantly available
* Would cut the startup time of Spaces with cache, as caching examples is the thing that takes the most time in startup of Spaces
* With ZeroGPU, when a GPU isn’t available, it errors out while caching examples

We can make this another option for the `cache_examples` parameter (the string "lazy") and maybe make this the default behavior on Spaces. 

See also: [internal discussion](https://huggingface.slack.com/archives/C02QZLG8GMN/p1707843648515889)

cc @radames @cbensimon as well

shubhamofbce:
@abidlabs I am interested in working on this, I saw the code and I need to make changes in the `Examples class in helpers.py`. I just can't understand what you mean by `cache them when after they are first processed`, like what is the meaning of processed here? I will be thankful if you can answer this or connect me with someone who can help me understand this.

abidlabs:
> @abidlabs could you clarify cache them after they are first processed? Is it about caching only when examples are run by the user and not during the startup in lazy mode?

Yes exactly!

PR #7863, opened 2024-03-27
Add support for lazy caching of examples, as well as add `GRADIO_CACHE_EXAMPLES` env variable

abidlabs:
Adds a new mode of caching examples: "lazy" -- which means that a particular example will only get cached after it is first used (by any user) in the Gradio app. This should help speed up the launch times for Gradio apps on Spaces and is particularly important for ZeroGPU, as GPUs may not be available when the app first launches.

Closes: #7424 

cc @apolinario @cbensimon @hysts


gradio/helpers.py
```diff
@@ -286,32 +288,85 @@ async def load_example(example_id):
...
+            async for item in self.fn(*args):
+                yield item
+        elif inspect.isgeneratorfunction(self.fn):
+            for item in self.fn(*args):
+                yield item
+        else:
+            yield self.fn(*args)
+
+    async def lazy_cache(self, example_index):
+        if Path(self.cached_indices_file).exists():
```
abidlabs, line 345: The reason I did it this way -- saving the indices of examples that have been cached -- in a separate CSV is because I wanted to preserve one nice feature of cached examples currently, that is that if you cache examples locally and then upload the cached examples to Spaces, you don't need to cache them again. This can save a lot of time for demos that take a long time to run.

gradio/helpers.py
```diff
@@ -286,32 +288,85 @@ async def load_example(example_id):
...
+        elif inspect.isasyncgenfunction(self.fn):
+            async for item in self.fn(*args):
+                yield item
+        elif inspect.isgeneratorfunction(self.fn):
+            for item in self.fn(*args):
+                yield item
+        else:
+            yield self.fn(*args)
+
+    async def lazy_cache(self, example_index):
```
abidlabs, line 344: The basic idea here is that lazy caching does two things:
1. Checks to see if the example index is already cached. If so, it loads it from cache
2. Otherwise, it runs the original prediction function and then caches the result. This is why we need the `_handle_callable_as_generator` method -- to handle the various kinds of prediction functions that are possible.

abidlabs:
Still need to add tests but should otherwise be ready for review @freddyaboulton @aliabid94

gradio/helpers.py
```diff
@@ -106,24 +106,47 @@ def __init__(
...
+                    self.cache_examples = True
+                elif cache_examples_env.lower() == "false":
+                    self.cache_examples = False
+                elif cache_examples_env.lower() == "lazy":
+                    self.cache_examples = "lazy"
+                else:
+                    raise ValueError(
+                        "The `GRADIO_CACHE_EXAMPLES` env variable must be one of: 'true', 'false', 'lazy' (case-insensitive)."
+                    )
+            elif utils.get_space() and fn is not None and outputs is not None:
```
freddyaboulton, line 138: I thought we wanted to set "lazy" for zero-GPU spaces?

abidlabs: See [internal discussion](https://huggingface.slack.com/archives/C04ABS7C13K/p1711645269328639?thread_ts=1711606376.711969&cid=C04ABS7C13K)

gradio/helpers.py
```diff
@@ -298,9 +341,56 @@ async def load_example(example_id):
...
+            result = await self.fn(*args)
+            yield result
+        elif inspect.isasyncgenfunction(self.fn):
+            async for item in self.fn(*args):
+                yield item
+        elif inspect.isgeneratorfunction(self.fn):
+            for item in self.fn(*args):
+                yield item
+        else:
+            yield self.fn(*args)
```
freddyaboulton, line 357: This will block the event loop

abidlabs: Ah right I was going to find a better way to do this. Do you know how to handle this?

freddyaboulton: We need to split this into two functions, one a coroutine and the other a normal function and pass the appropriate one to the event depending on the type of the `fn` parameter

freddyaboulton: There may be other solutions ^

abidlabs: Ok let me take a look! Thanks

freddyaboulton: What I suggested abouve is what we do for the iterators in the non-lazy case basically

abidlabs: Right, I was also seeing if it made sense to do this: https://github.com/gradio-app/gradio/blob/e0a8b7fa2c60591f7fc6861707f65bf85f4ad27f/gradio/blocks.py#L1266

abidlabs:
Fixed now, thanks for the suggestion @freddyaboulton! Test with e.g. 

```py
import gradio as gr
import time

def test(x):
    yield x
    time.sleep(1)
    yield x+x
    time.sleep(1)
    yield x+x+x

gr.Interface(test, "textbox", "textbox", examples=[1], cache_examples="lazy").launch()
```

gradio/interface.py
```diff
@@ -96,7 +96,7 @@ def __init__(
         inputs: str | Component | list[str | Component] | None,
         outputs: str | Component | list[str | Component] | None,
         examples: list[Any] | list[list[Any]] | str | None = None,
-        cache_examples: bool | None = None,
+        cache_examples: bool | Literal["lazy"] | None = None,
```
freddyaboulton, line 99: We need to update this typehint and doc for chat_interface!

freddyaboulton (approved):
Thanks for the nice feature and addressing the comments @abidlabs ! This looks good to me

gradio/interface.py
abidlabs: Right! Will add

abidlabs:
Thanks for the review @freddyaboulton! I'll add to ChatInterface and merge in later today

abidlabs:
Added to ChatInterface, added docs, and testing. Will merge after CI runs. Thanks so much for the review @freddyaboulton

PR #7865, opened 2024-03-27
JS functions break entire app if there's no input, fixed

aliabid94:
Basically JS functions break the app if they dont have any inputs, e.g.

```python
import gradio as gr

with gr.Blocks() as demo:
    btn = gr.Button()
    btn.click(None, None, None, js="""() => console.log('Test')""")

    text = gr.Textbox()
    text2 = gr.Textbox()
    text.submit(lambda x:x, text, text2)

demo.launch()
```
After clicking the button, the second event listener stops working.

This as also causing theme builder to break. Fixed now but maybe not correctly, if @pngwn you have a better suggestion.

Closes: https://github.com/gradio-app/gradio/issues/7866


pngwn (approved):
Ah. So that's why this happened. Looks good to me!

abidlabs:
Could we add an E2E test for the theme builder? Its a very important feature that's basically untested, e.g. https://github.com/gradio-app/gradio/issues/7866

PR #7869, opened 2024-03-28
Make buttons in `gr.ChatInterface` more mobile-friendly

abidlabs:
Reduces the default min width of the buttons in `gr.ChatInterface` to make it more mobile-friendly.

Before:

image

After:

image


dawoodkhan82 (approved):
lgtm

abidlabs:
Thanks @dawoodkhan82!

PR #7875, opened 2024-03-28
Fix how files are processed in `gr.ChatInterface` 

abidlabs:
When we added multimodal support, we changed the format of files so that we were just sending a file path instead of a file data dictionary. We forgot to make the corresponding change in `gr.ChatInterface`.

Test with:

```py
import gradio as gr

gr.ChatInterface(lambda x,y: (x["files"][0],), multimodal=True).launch()
```

Made some tweaks to the guide as well.


dawoodkhan82 (approved):
Ah yes makes sense. Thanks for fixing this.

PR #7877, opened 2024-03-28
[ZeroGPU] postMessage `zerogpu-headers`

cbensimon:
Provide a logged-in experience for ZeroGPU (regarding GPU quotas)

ZeroGPU functions are marked with a `zerogpu` attribute by the `@spaces.GPU` decorator
When this attribute is detected, `"zerogpu"` is set to `True` in the dependency config

On JS client side, when `dependency.zerogpu` is `true` we ask parent frame (HF hub) for headers to be included in the API request (through the postMessage API)


gradio/blocks.py
abidlabs, line 1: Changes in this file lgtm. @hannahblair could you review the js client changes?

abidlabs (approved):
Nice thank you @cbensimon!

PR #7881, opened 2024-03-29
Fix audio streaming out

aliabid94:
Fix audio streaming caused by broken URL.

Test with streaming_audio_out demo.

Fixes: #7497


freddyaboulton (approved):
Nice looks good to me @aliabid94 !

freddyaboulton:
We should add an e2e test for this because I think this has regressed before.

abidlabs:
Agreed with @freddyaboulton about the test

Let me tag @pseudotensor who reported this as well, if you'd like to confirm the fix @pseudotensor

pseudotensor:
Yes, it works now!  Thanks!

PR #7886, opened 2024-03-29
logout route deleting cookie fix

### LINKED ISSUE #7885: logout route not deleting cookies
### Describe the bug

using /logout route, the cookies are not deleted properly 

### Reproduction

```python
import gradio as gr

def update_message(request: gr.Request):
    return f"Welcome, {request.username}"

with gr.Blocks() as demo:
    m = gr.Markdown()
    logout_button = gr.Button("Logout", link="/logout")
    demo.load(update_message, None, m)
    
demo.launch(auth=[("Pete", "Pete"), ("Dawood", "Dawood")])
```

### Screenshot

_No response_

### Logs

_No response_

### System Info

```shell
gradio==4.21.0
gradio_client==0.12.0
```

### Severity

I can work around it

MichaelPerger:
Closes #7885 

now deleting cookies properly by calling 
RedirectResponse(url="/", status_code=status.HTTP_302_FOUND) 
earlier in the logout route

### DIFF
--- a/gradio/routes.py
+++ b/gradio/routes.py
@@ -331,6 +331,7 @@ class App(FastAPI):
 
             @app.get("/logout")
             def logout(response: Response, user: str = Depends(get_current_user)):
+                response = RedirectResponse(url="/", status_code=status.HTTP_302_FOUND)
                 response.delete_cookie(key=f"access-token-{app.cookie_id}", path="/")
                 response.delete_cookie(
                     key=f"access-token-unsecure-{app.cookie_id}", path="/"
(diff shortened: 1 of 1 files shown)


MichaelPerger:
should be ready for a review request:  @aliabd @dawoodkhan82
i saw you two the most, hopefully its fine that i link you - i am new to the open source world,
thanks for checking :)

abidlabs:
Thanks @MichaelPerger for pointing out the issue and for the fix PR. I've just removed `response: Response` from the route params since we don't need it if we do it this way (which I agree is the correct way)

abidlabs (approved):

Issue #7882, opened 2024-03-29
[REGRESSION] client hangs reaching server on 4.23.0 and 4.24.0 (ok for 4.22.0 and lower) -- with repro


pseudotensor:
### Describe the bug

Starting with gradio 4.23.0 (all ok in 4.22.0 and lower) instead of the client connecting to h2oGPT, it just hangs, and server shows it trying to get /info many times.

What one sees with gradio 4.22.0 and lower:

server:
```
INFO:     127.0.0.1:36172 - "GET /config HTTP/1.1" 200 OK
```
client:
```
Successfully installed gradio-4.22.0 gradio-client-0.13.0
(h2ogpt) jon@pseudotensor:~/h2ogpt$ python
Python 3.10.13 | packaged by conda-forge | (main, Dec 23 2023, 15:36:39) [GCC 12.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from gradio_client import Client
>>> client = Client('http://localhost:7860')
Loaded as API: http://localhost:7860/ ✔
```
No hang.

But on 4.23.0 and 4.24.0 see instead:

server:
```

INFO:     127.0.0.1:59630 - "GET /config HTTP/1.1" 200 OK
INFO:     127.0.0.1:59646 - "GET /info HTTP/1.1" 200 OK
INFO:     127.0.0.1:59658 - "GET /info HTTP/1.1" 200 OK
INFO:     127.0.0.1:59674 - "GET /info HTTP/1.1" 200 OK
INFO:     127.0.0.1:59688 - "GET /info HTTP/1.1" 200 OK
INFO:     127.0.0.1:59690 - "GET /info HTTP/1.1" 200 OK
INFO:     127.0.0.1:59702 - "GET /info HTTP/1.1" 200 OK
INFO:     127.0.0.1:59718 - "GET /info HTTP/1.1" 200 OK
INFO:     127.0.0.1:52502 - "GET /info HTTP/1.1" 200 OK
INFO:     127.0.0.1:52506 - "GET /info HTTP/1.1" 200 OK
INFO:     127.0.0.1:52514 - "GET /info HTTP/1.1" 200 OK
INFO:     127.0.0.1:52530 - "GET /info HTTP/1.1" 200 OK
INFO:     127.0.0.1:52538 - "GET /info HTTP/1.1" 200 OK
INFO:     127.0.0.1:52554 - "GET /info HTTP/1.1" 200 OK
[...]

PR #7888, opened 2024-03-29
Cache view_api info in server and python client

freddyaboulton:
## Description

Closes: #7882 

The client will call the api info endpoint only once during initialization. The server caches the result because right now it should not change throughout the lifetime of the app. This is for the case where many clients connect to view the api info simultaneously.

## 🎯 PRs Should Target Issues

## Tests


client/python/gradio_client/client.py
```diff
@@ -1029,8 +1028,6 @@ def _get_component_type(self, component_id: int):
         )
 
     def _get_parameters_info(self) -> list[ParameterInfo] | None:
-        if not self.client._info:
-            self._info = self.client._get_api_info()
```
abidlabs, line 1033: Whoops 😭

abidlabs (approved):
LGTM thanks for fixing this @freddyaboulton!

freddyaboulton:
Thank you for the quick review!! And @pseudotensor for the tip about the test

PR #7897, opened 2024-03-31
 Replace pip install with uv pip install in our CI workflows.

hegdeadithyak:
## Description
`uv` is a fast and reliable Python package installer `pip` built with Rust. It aims to be a comprehensive solution for Python package management, similar to Cargo for Rust.

Closes: #7442

### DIFF
--- a/.github/actions/install-all-deps/action.yml
+++ b/.github/actions/install-all-deps/action.yml
@@ -62,7 +62,7 @@ runs:
       shell: bash
       run: |
         python -m pip install --upgrade virtualenv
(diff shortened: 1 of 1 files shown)


pngwn:
Thanks for the PR but uv is a little immature at the minute, we’ll plan on revisiting when it is a bit more stable. See https://github.com/gradio-app/gradio/pull/7457

hegdeadithyak:
Got it. @pngwn

pngwn:
Apologies for the confusion, I’ve added a note to the issue to clarify our position.

Issue #7825, opened 2024-03-25
Limit file upload size


freddyaboulton:
- [ ] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
Gradio app developers should be able to control the max individual file size that can be uploaded. 

**Describe the solution you'd like**  
Add a `max_file_size: int` parameter to media-based components to limit the size of files that can be uploaded. Setting it at the component level will be more flexible for different input file types instead of at the global level.

This should work both in the UI and the clients. Ideally an error should be raised prior to uploading any files so the logic should be replicated across the clients. 

**Additional context**

Issue #6700, opened 2023-12-07
Allow users to close error tag and get back to the component


apolinario:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
Sometimes the same component is an input and an output. If something errors out, the component gets in a state of error and you can't fix it because it was also the input. 

Another scenario where this happens is in case there's an error in a `.upload` from a `gr.Image` component for example: maybe the cropping errored out, or you uploaded an invalid image. But now the component is broken and you can't "close" the error to try again with a different image. 

**Describe the solution you'd like**  
Allow users to "close" the `Error` state of components when they error out, revealing what was underneath

apolinario:
Minimal reproduction
```py
import gradio as gr

def crop_image(image):
  #Simulate error with cropping or the uploaded image
  raise gr.Error("Test")

with gr.Blocks() as demo:
    image = gr.Image(label="Name")
    image.upload(fn=crop_image, inputs=image, outputs=image)
  
if __name__ == "__main__":
    demo.launch()
```

This will render the upload component useless

https://github.com/gradio-app/gradio/assets/788417/84350a73-2008-4c19-b186-16e26b5014da

(Playground video but the same happens in 4.x) 

I think a `x` button to close the error and reveal the component behind would be enough

freddyaboulton:
![max_file_size_upload](https://github.com/gradio-app/gradio/assets/41651716/7547330c-a082-4901-a291-3f150a197e45)

PR #7909, opened 2024-04-01
Add `max_file_size` parameter to `launch()` that limits the size of files that can be uploaded in the Gradio app

freddyaboulton:
## Description

Closes: #7825
Closes: #6700 

`max_file_size` is a block level component set in launch.

Developers can set this parameter as an int (like `1024` to specify 1 kb or as a string like "1kb"). Validation happens before the file is uploaded both for the app and the python client. The `upload` route will return a 413 error code if any of the uploaded files are larger than max_file_size

## 🎯 PRs Should Target Issues

## Tests


pngwn (commented):
I think we should perform the max file size check on the frontend in the client in the upload files utility. This way it will apply to the client and also work for all file uploads, not just those uploaded via the Upload component. This means it will be easy to add this feature to custom components that upload assets using the client rather than the Upload component too.

aliabid94:
Should this logic be performed in the backend for security? As in, a user may not want to allow file sizes greater than X to be uploaded, and a direct API call could circumvent the check?

abidlabs:
Do we need this parameter for every component? It seems to me that this should be an app-level setting, which would be simpler for developers to use (they only need to set it once) and would take care of (1)

pngwn:
In a demo with multiple upload types you may want more granular control. Although in some cases you may care about the raw file size, in others it may be the relationship between size and how that impact inference. I think you'd expect a video with audio to be bigger than a text file or image, for example. But maybe there is a better heuristic for limiting specific file types.

freddyaboulton:
Yea I was thinking along @pngwn 's lines that app authors should have granular control over upload limits for the different modalities their demo can use. 

Maybe for 1) the server can block any upload that's larger than the maximum `max_file_size` used across all the components in the demo?

abidlabs:
I'm fairly certain that app developers who are setting a max_file_size mostly care about making sure their server doesn't run out of disk space -- its less so about controlling the specific files for different components. (Do we have any original user requests from users that we can refer to?)

> Maybe for 1) the server can block any upload that's larger than the maximum max_file_size used across all the components in the demo?

The problem with this, and really with any solution that is component-level, is that if a user forgets to set the limit on any component in a potentially large and complex gradio app, then their app is now exposed to potentially very large files being uploaded. That's why I would prefer an app-level limit -- easy for developers to set / configure and hard to forget

Also, if a Gradio app uses a custom component, then an app-level setting is the only way for it to apply to their app as well.

freddyaboulton:
Ok after sleeping on this, I am coming around to @abidlabs 's argument. Will add a block-level argument and modify the upload route to respect it. Will keep the client-side validation though for better UX.

js/audio/Index.svelte
```diff
@@ -34,6 +34,7 @@
 	export let show_share_button = false;
 	export let editable = true;
 	export let waveform_options: WaveformOptions = {};
+	export let max_file_size: number | null = null;
```
freddyaboulton, line 37: The `max_file_size` is passed in from the app, it's not in the component config. This is done so that the `upload` function in the client can know the max_file_size

freddyaboulton:
Ok I have made the changes above^, this is good for another review!

gradio/utils.py
```diff
@@ -1280,3 +1280,25 @@ async def function_wrapper(*args, **kwargs):
...
+
+
+def _parse_file_size(size: str | int | None) -> int | None:
+    if isinstance(size, int) or size is None:
+        return size
+
+    last_digit_index = next(
+        (i for i, c in enumerate(size) if not c.isdigit()), len(size)
+    )
+    size_int, unit = int(size[:last_digit_index]), size[last_digit_index:].upper()
```
abidlabs, line 1300: Perhaps we should strip these to be a bit more robust in case users do something like "5 mb"
```suggestion
    size_int, unit = int(size[:last_digit_index].strip()), size[last_digit_index:].strip().upper()
```

gradio/routes.py
```diff
@@ -1100,6 +1108,7 @@ def mount_gradio_app(
     blocked_paths: list[str] | None = None,
     favicon_path: str | None = None,
     show_error: bool = True,
+    max_file_size: str | int | None = None,
```
abidlabs, line 1111: Nice to include here as well

gradio/routes.py
```diff
@@ -971,24 +972,31 @@ async def upload_file(
...
             locations: list[str] = []
-            for temp_file in form.getlist("files"):
-                if not isinstance(temp_file, GradioUploadFile):
+
+            uploaded_files: list[GradioUploadFile] = form.getlist("files")  # type: ignore
+            for f in uploaded_files:
+                if not isinstance(f, GradioUploadFile):
                     raise TypeError("File is not an instance of GradioUploadFile")
+
+            for temp_file in uploaded_files:
```
abidlabs, line 999: Not clear to me why this has been refactored to two separate for loops

gradio/route_utils.py
```diff
@@ -357,7 +357,7 @@ def __init__(
         filename: str | None = None,
         headers: Headers | None = None,
     ) -> None:
-        super().__init__(file, size=size, filename=filename, headers=headers)
+        super().__init__(file, size=size or 0, filename=filename, headers=headers)
```
abidlabs, line 360: Is this correct? Meaning if a user sets size to be `None`, then it would be set here to 0?

gradio/__init__.py
```diff
@@ -96,7 +96,7 @@
     TextArea,
 )
 from gradio.themes import Base as Theme
-from gradio.utils import NO_RELOAD, get_package_version, set_static_paths
+from gradio.utils import NO_RELOAD, FileSize, get_package_version, set_static_paths
```
abidlabs, line 99: Why do we export FileSize?

.config/playwright-setup.js
```diff
@@ -115,7 +115,12 @@ ${demos.map((d) => `from demo.${d}.run import demo as ${d}`).join("\n")}
 
 app = FastAPI()
 ${demos
-	.map((d) => `app = gr.mount_gradio_app(app, ${d}, path="/${d}")`)
+	.map(
+		(d) =>
+			`app = gr.mount_gradio_app(app, ${d}, path="/${d}", max_file_size=${
+				d == "upload_file_limit_test" ? 1 : "None"
```
abidlabs, line 121: Might be worth adding a note in `demo/upload_file_limit_test` that the upload limit is set in `.config/playwright-setup.js` because
(1) users might be confused since no limit is explicitly set in that demo
(2) if we were to change the test later, we might not remember where the limit is actually being set

demo/upload_file_limit_test/run.py
```diff
@@ -0,0 +1,24 @@
+import gradio as gr
+
+with gr.Blocks() as demo:
+    gr.Markdown("""
+    # ⬆️📁 max_file_size test
+    Components have a max file size of 1kb (disabling upload). The error modal should pop up when a file is uploaded.
```
abidlabs, line 6:
```suggestion
    This Gradio app limits uploads to be 1b or lower. The error modal should pop up when a file is uploaded larger than that limit          
```

abidlabs: Is it 1b or 1kb? The playwright config suggests 1b

client/python/test/test_client.py
```diff
@@ -85,6 +88,13 @@ def test_raise_error_invalid_state(self):
         with pytest.raises(ValueError, match="invalid state"):
             Client("gradio-tests/paused-space")
 
+    def test_raise_error_max_file_size(self, max_file_size_demo):
+        with connect(max_file_size_demo, max_file_size="1b") as client:
+            with pytest.raises(
+                ValueError, match=f"File {__file__} exceeds the maximum file size"
+            ):
+                client.predict(file(__file__), api_name="/upload_1b")
```
abidlabs, line 96: Is this tryping to upload the test file itself? That's interesting but maybe more robust to upload a sample test file that we know is of a fixed size. We should also test that files that are below the specified upload limit are successfully uploaded

js/app/test/upload_file_limit_test.spec.ts
abidlabs, line 1: Thanks for adding these tests @freddyaboulton! Can we also add at least a couple of tests that ensure that when an upload limit is specified but a file is smaller than that limit, the file is successfully able to be uploaded?

test/test_routes.py
```diff
@@ -1268,3 +1268,14 @@ def test_compare_passwords_securely():
...
     assert starts_with_protocol(string) == expected
+
+
+def test_max_file_size_used_in_upload_route(connect):
+    with gr.Blocks() as demo:
+        gr.Markdown("Max file size demo")
+
+    app, _, _ = demo.launch(prevent_thread_lock=True, max_file_size="1kb")
+    test_client = TestClient(app)
+    with open("test/test_files/bus.png", "rb") as f:
```
abidlabs, line 1279: Again, let's do a check to ensure that files smaller than the specified limit are able to be uploaded

abidlabs:
Super cool PR @freddyaboulton! I left some notes above. Generally speaking, this PR works well and the frontend validation looks great.

One issue that I'm running when testing this is that when the `max_file_size` limit is slightly higher than a file that I'm trying to upload, I see this error: 

image

You should be able to repro by running this demo:

```py
import gradio as gr

gr.Interface(lambda x:x, "file", "file").launch(max_file_size="55kb")
```

and then uploading this image, which is 51kb:

cheetah (1)

gradio/route_utils.py
freddyaboulton: This isn't a public API - I did this to try and make the type hints work but it's not needed anyways so will revert

gradio/__init__.py
freddyaboulton: So users can use ` 5 * FileSize.MB`

abidlabs: got it. that reminds me, we should add documentation for `max_file_size` in one of the guides. Perhaps https://www.gradio.app/guides/sharing-your-app#security-and-file-access

client/python/gradio_client/client.py
```diff
@@ -1228,6 +1230,22 @@ def _upload_file(self, f: str | dict) -> dict[str, str]:
...
+                ),
+                {},
+            )
+            max_file_size = self.client.config.get("max_file_size", None)
+            max_file_size = math.inf if max_file_size is None else max_file_size
+            if os.path.getsize(file_path) > max_file_size:
+                raise ValueError(
+                    f"File {file_path} exceeds the maximum file size of {max_file_size} bytes "
+                    f"set in {component_config.get('label', '') + ''} component."
+                )
```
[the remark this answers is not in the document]
freddyaboulton: I want to keep the component label in the error message to make the error more helpful in the case multiple files are being passed into the prediction

demo/upload_file_limit_test/run.py
freddyaboulton: Fixed!

client/python/test/test_client.py
freddyaboulton: Yea good point!

test/test_routes.py
freddyaboulton: Done!

js/app/test/upload_file_limit_test.spec.ts
freddyaboulton: Yes good point!

gradio/utils.py
freddyaboulton: Added a test for this case too

gradio/routes.py
freddyaboulton: Reverted

gradio/__init__.py
freddyaboulton: Great idea - just pushed up!

freddyaboulton:
Thanks for the very thorough review @abidlabs ! All the comments have been addressed, should be good for another review

gradio/routes.py
```diff
@@ -971,21 +972,25 @@ async def upload_file(
             try:
                 if upload_id:
                     file_upload_statuses.track(upload_id)
+                max_file_size = app.get_blocks().max_file_size or math.inf
```
abidlabs, line 975: A bit of an edge case, but this will convert 0 to math.inf

guides/01_getting-started/03_sharing-your-app.md
```diff
@@ -451,4 +451,16 @@ Gradio DOES NOT ALLOW access to:
 
 - **Any other paths on the host machine**. Users should NOT be able to access other arbitrary paths on the host.
 
+Sharing your Gradio application will also allow users to upload files to your computer or server. You can set a maximum file size for uploads to prevent abuse and to preserve disk space. You can do this with the `max_file_size` parameter of `.launch`. For example, the following two code snippets limit file sizes to 5 megabytes.
```
abidlabs, line 454: Nit: maybe we should clarify its 5 mb per file
```suggestion
Sharing your Gradio application will also allow users to upload files to your computer or server. You can set a maximum file size for uploads to prevent abuse and to preserve disk space. You can do this with the `max_file_size` parameter of `.launch`. For example, the following two code snippets limit file sizes to 5 megabytes per file.
```

abidlabs (approved):
Just a couple of nits, but LGTM! Beautiful PR @freddyaboulton

js/app/src/init.ts
```diff
@@ -160,6 +160,9 @@ export function create_components(): {
 			instance.props.server_fns,
 			app
 		);
+		// Pass the max file size from the app to the component
+		// so we can validate it before files are uploaded
+		instance.props.max_file_size = app.config.max_file_size;
```
pngwn, line 165: Rather than adding this here, i think it might make sense to add it to the `Gradio` object that gets passed into each component. The idea of of that object is that it contains app level configuration (ie globals) where as props should just be specific to the component.

We should be able to pass the value through to the `Render.svelte` via blocks and then add it as an additional arg to `new Gradio()`

pngwn (commented):
Looks great in general, just left a small comment about how we make the value accessible to components.

yvrjsharma:
Hi @freddyaboulton, when we try to upload a file that exceeds the maximum file size limit set in `demo.launch()`, we are prevented from doing so as expected. However, there is no warning message displayed to the user indicating that the file size limit has been exceeded. I think it would be helpful to display a warning message to the user specifying the maximum allowed file size. I tested this with the wheel on a Colab notebook.

Repro:
```python
import gradio as gr
demo = gr.Interface(lambda x: x, "image", "image")
demo.launch(max_file_size="1mb",)
```

https://github.com/gradio-app/gradio/assets/48665385/ff287f5d-97a0-4682-a24b-8954f172f3a7

abidlabs:
+1 to @yvrjsharma's suggestion. It would be particularly annoying if you have many inputs that you've filled out and then you upload a file that is too large, as it would force you to clear everything. 

The other option would be to make error messages close-able as @apolinario has suggested https://github.com/gradio-app/gradio/issues/6700

yvrjsharma:
Thanks @abidlabs and @freddyaboulton for looking into this. 
I am just wondering if we were to allow users to close the error messages, would it still make sense to use `gr.Warnings`? As users would have the option to dismiss the error and proceed, it could potentially blur the distinction between an error and a warning.

freddyaboulton:
I still think `Error` is more appropriate here. To me, a warning means that the program can continue as-is which is not the case if the file is too large. I'm working on making the error close-able!

.config/playwright-setup.js
freddyaboulton: Done!

gradio/routes.py
freddyaboulton: Good catch!

js/app/src/Blocks.svelte
```diff
@@ -452,6 +455,21 @@
 
 	$: set_status($loading_status);
 
+	function update_status(
+		id: int,
```
pngwn, line 459: No `int` type in javascript. Use `number`.

js/app/src/Blocks.svelte
```diff
@@ -18,6 +18,7 @@
 	import logo from "./images/logo.svg";
 	import api_logo from "./api_docs/img/api-logo.svg";
 	import { create_components, AsyncFunction } from "./init";
+	import type { int } from "babylonjs";
```
pngwn, line 21: Mistaken, autoimport.

js/statustracker/static/index.svelte
```diff
@@ -271,6 +277,16 @@
 			<p class="loading">{loading_text}</p>
 		{/if}
 	{:else if status === "error"}
+		<div class="clear-status">
+			<IconButton
+				Icon={Clear}
+				label={i18n("common.clear")}
+				disabled={false}
+				on:click={(event) => {
```
pngwn, line 285: we don't use `event` so we can just make this `()`

pngwn (approved):
Looks great to me. Just one or two small thing and good to go! Thanks @freddyaboulton!

Issue #4268, opened 2023-05-18
gr.Examples Enable for each example to have a label and a value


t-marinos:
- [yes ] I have searched to see if a similar issue already exists.

For example I would like the gr.Examples to display the example as "Person 1" and when it is clicked, to populate the linked component with "John Doe".

abidlabs:
The way I'm thinking to implement this is to add a `labels` parameter to `gr.Examples()` (and an `examples_labels` parameter to `gr.Interface()`) which is a list of the same length as `examples`. If provided, these text labels appear instead of the original examples, but clicking on the labels populates the inputs in the same way as before.

tlabarta:
@abidlabs Is there any update on this? I could need exactly this enhancement for a use case where we have long text examples that we could just abbreviate by a short label.

PR #7912, opened 2024-04-02
Allow displaying gr.Code examples by their filename if value is a tuple

freddyaboulton:
## Description

You can display a filename for the code component example as opposed to the whole contents of the file. Just need to pass in a tuple to denote it's a file (same as postprocess)

Test with

```python
import gradio as gr

demo = gr.Interface(
    lambda x: x,
    gr.Code(),
    gr.Code(),
    examples=[[("/Users/freddy/sources/gradio/demo/code_component/run.py",)], ]
)


if __name__ == "__main__":
    demo.launch()
```

Don't think we need #4268 for this

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/gradio/components/code.py
+++ b/gradio/components/code.py
@@ -163,3 +163,8 @@ class Code(Component):
 
     def example_value(self) -> Any:
         return "print('Hello World')"
+
+    def process_example(self, value: str | tuple[str] | None) -> str | None:
+        if isinstance(value, tuple):
+            return value[0]
+        return super().process_example(value)
--- a/gradio/components/code.py
+++ b/gradio/components/code.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from pathlib import Path
 from typing import Any, Callable, Literal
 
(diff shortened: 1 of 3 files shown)


gradio/components/code.py
```diff
@@ -163,3 +163,8 @@ def example_payload(self) -> Any:
 
     def example_value(self) -> Any:
         return "print('Hello World')"
+
+    def process_example(self, value: str | tuple[str] | None) -> str | None:
+        if isinstance(value, tuple):
+            return value[0]
```
abidlabs, line 169: I would suggest only showing the name of the file, not the whole path, to be consistent with: https://github.com/gradio-app/gradio/blob/c62a57e7f8f2f6dad0110d06e915c48e7f628073/gradio/components/audio.py#L307

abidlabs (approved):

julien-c (commented):
very cool!

dawoodkhan82 (approved):
sweet!

Issue #7910, opened 2024-04-02
Infinite search in dataframe


clefourrier:
### Describe the bug

Hi!
Looking for items in a dataframe of one element makes the table load forever.
This happens in 4.22 and 4.24, but not 4.9.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

I made you a MWE [here](https://huggingface.co/spaces/clefourrier/debug-leaderboard). 

Just search for anything and it will load forever. Search should return the available value if matched, else the full table.
You can duplicate the space and change the version to 4.9 and it will behave properly.

### System Info

```shell
On my local env: 


(tmp_gradio) clementine@multivaclem-HF:~/Desktop/debug-leaderboard$ gradio environment
Gradio Environment Information:
------------------------------
Operating System: Linux
gradio version: 4.24.0
gradio_client version: 0.14.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.3.0
fastapi: 0.110.1
ffmpy: 0.3.2
gradio-client==0.14.0 is not installed.
httpx: 0.27.0
huggingface-hub: 0.22.2
importlib-resources: 6.4.0
jinja2: 3.1.3
markupsafe: 2.1.5
matplotlib: 3.7.1
numpy: 1.24.2
orjson: 3.10.0
packaging: 24.0
pandas: 2.0.0
pillow: 10.3.0
pydantic: 2.6.4
pydub: 0.25.1
python-multipart: 0.0.9
pyyaml: 6.0.1
ruff: 0.3.5
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.12.0
typing-extensions: 4.10.0
uvicorn: 0.29.0
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2023.6.0
httpx: 0.27.0
huggingface-hub: 0.22.2
packaging: 24.0
typing-extensions: 4.10.0
websockets: 11.0.3
```

But best refer to the space.
```


### Severity

I can work around it

PR #7915, opened 2024-04-02
Fix gr.CheckboxGroup change event

freddyaboulton:
## Description

Closes: #7910 

The gr.CheckboxGroup change event was triggering itself. You can trigger with a simple script like this

```python
import gradio as gr

with gr.Blocks() as demo:
    select = gr.CheckboxGroup(choices=['Option 1', 'Option 2', 'Option 3'], value=['Option 1', 'Option 2'],
                            )
    jsout = gr.JSON()
    select.change(lambda c: c, select, jsout)

demo.launch()
```

Also adding a mini llm leaderboard to our deployed demos and e2e tests so that we can better track regressions in leaderboard type demos cc @clefourrier

image

## 🎯 PRs Should Target Issues

## Tests


pngwn (commented):
Hmm, I'd rather we moved away from the equality trick. What happens if we set this in `Index.svelte`:

```svelte
<svelte:options immutable={true} />
```

This needs to be at the top level (not a child of anything).

pngwn (approved):
Never mind, the fix to this general issue needs more thought.

Looks good to merge @freddyaboulton! Thanks!

freddyaboulton:
Thank you @pngwn !

Issue #7726, opened 2024-03-18
Programmatically selecting tabs is broken


FurkanGozukara:
Let's say I have below tabs

![image](https://github.com/gradio-app/gradio/assets/19240467/2784347d-6e36-475b-b49f-aa1a821f1758)

How can I switch to specific tab when user clicks a button?

So when user clicks load Metadata I can switch back to InstantId tab

Tab index 0 or 1 etc?

`gradio==4.21.0`

abidlabs:
Hi @FurkanGozukara this is how its supposed to be done: https://github.com/gradio-app/gradio/issues/2412

However, I just tested this code on the latest version of gradio and its not working. It seems that we've had a regression. Let me categorize the issue appropriately and we'll get it fixed.

cc @hannahblair for visibility

FurkanGozukara:
yes this didnt work for me

def change_tab():
    return gr.Tabs.update(selected=1)

phihung:
Here a quick example to reproduce

Both initial value ` with gr.Tabs(selected="b") as tabs` and button click are ignored
There are some warnings in the UI :  `Tabs.svelte:84 Attempted to select a non-interactive or hidden tab.`

```
import gradio as gr

def select_c():
    return gr.Tabs(selected="c")

with gr.Blocks() as demo:
    with gr.Tabs(selected="b") as tabs:
        with gr.Tab("A", id="a"):
            gr.Markdown("## A")
        with gr.Tab("B", id="b"):
            gr.Markdown("## B")
        with gr.Tab("C", id="c"):
            gr.Markdown("## C")

    btn = gr.Button("Switch to tab C")
    btn.click(select_c, None, tabs)

demo.launch()
```

ThanThoai:
> Here a quick example to reproduce
> 
> Both initial value ` with gr.Tabs(selected="b") as tabs` and button click are ignored There are some warnings in the UI : `Tabs.svelte:84 Attempted to select a non-interactive or hidden tab.`
> 
> ```python
> import gradio as gr
> 
> def select_c():
>     return gr.Tabs(selected="c")
> 
> with gr.Blocks() as demo:
>     with gr.Tabs(selected="b") as tabs:
>         with gr.Tab("A", id="a"):
>             gr.Markdown("## A")
>         with gr.Tab("B", id="b"):
>             gr.Markdown("## B")
>         with gr.Tab("C", id="c"):
>             gr.Markdown("## C")
> 
[...]

Issue #7786, opened 2024-03-21
[REGRESSION again] Accordion close behavior back in gradio 4.22.0


pseudotensor:
Accordion either pops open and while streaming cannot be closed, or other way it closes and cannot be opened, depending upon where the elements are.

Reproduction
pops-open no matter what repro is below. If you need a repro of case where it pops-close no matter, let me know.

```python
import time

import gradio as gr

with gr.Blocks() as demo:
    with gr.Column():
        with gr.Accordion():
            image = gr.Image()
        with gr.Tabs():
            with gr.Tab("Chat"):
                chatbot = gr.Chatbot()
                msg = gr.Textbox()
                clear = gr.Button("Clear")
            with gr.Tab("Other"):
                msg2 = gr.Textbox()

    def user(user_message, history):
        return "", history + [[user_message, None]]

    def bot(history):
        history[-1][1] = ''
        for msg in range(0, 10000):
            history[-1][1] += str(msg)
            if msg % 100 == 0:
                history[-1][1] += '\n'
            time.sleep(0.01)
            yield history

    msg.submit(user, [msg, chatbot], [msg, chatbot], queue=False).then(
        fn=bot, inputs=chatbot, outputs=chatbot, api_name='bot',
    )
    clear.click(lambda: None, None, chatbot, api_name='clear')

demo.queue()
demo.launch()
```

abidlabs:
Agggh that really sucks I thought we had a test for this. cc @hannahblair @pngwn if you have bandwidth could you take a look?

pseudotensor:
I confirmed by going back and forth from 4.20.1 vs. 4.22.0.  I didn't try between, so unsure if really 4.22.0 when hit.  I can bisect.

pseudotensor:
4.21.0 is also ok.  So only 4.22.0 is bad.

pngwn:
Hmm, this is weird. We also _do_ have a test for this.

pngwn:
Clearly not a very reliable one, will take a look.

pseudotensor:
Any progress here?

With https://github.com/gradio-app/gradio/issues/7882 I am stuck on 4.21.0 as all other new releases are DOA.

pseudotensor:
Any progress here?

With https://github.com/gradio-app/gradio/issues/7882 I am stuck on 4.21.0 as all other new releases are DOA.

pseudotensor:
[...]

PR #7916, opened 2024-04-02
Fix programmatic tab selection

aliabid94:
`selected` was not being respected for tabitems, because the `id` property of a tabitem was being overwritten internally. Renamed internal use of `id` to `_id`.

Test with:
```python
import gradio as gr

def select_c():
    return gr.Tabs(selected="c")

with gr.Blocks() as demo:
    with gr.Tabs(selected="b") as tabs:
        with gr.Tab("A", id="a"):
            gr.Markdown("## A")
        with gr.Tab("B", id="b"):
            gr.Markdown("## B")
        with gr.Tab("C", id="c"):
            gr.Markdown("## C")

    btn = gr.Button("Switch to tab C")
    btn.click(select_c, None, tabs)

demo.launch()
```

Fixes: #7726


pngwn (approved):
Ah nice catch. Thanks @aliabid94!

We should probably lint against 'shadowing' variables. I wonder if this fixes #7786 as well?

freddyaboulton (commented):
Nice @aliabid94 ! Can we add a test for this? It's regressed before

aliabid94:
Added tests. While I was adding tests, I caught an error that wouldn't show up in dev mode for some reason: If a gr.Tabs() had `selected=` set to a value in its constructor, than any event listener trigger that would cause a redraw would reset the gr.Tabs to the selected tab. I made changes to Tab.svelte to fix this, @pngwn if you could just take a quick pass at it again.

aliabid94:
actually my solution won't work in the case that an app programmatically tries to set the selected tab to the same tab a second time after a user moved away from that tab. Is there any way for gradio to know if the selected tab being passed is the result of an event listener output, or just a redraw resetting the prop?

pngwn:
What is the actually problem, sorry. In principle we shouldn't need to know the difference.

pngwn:
This feels like the accordion thing tbh. Gonna take a closer look.

aliabid94:
Sorry let me clarify. The issue I was trying to fix with selected, _selected, and old_selected_prop is that event listeners were causing the component to re-render, which was causing the props to be reset to their set values. So the selected tab would get reset to the constructor-set value. So if there was a gr.Tabs(selected="b") setting the default to "b", and a user clicked away to tab "a", when an event listener triggered a re-render, the tab would reset to "b".

pngwn:
I see. That does sound very similar to the accordion issues where it resets to the default.

pngwn:
@aliabid94 I have just opened #7963 to fix the accordion 'reset' issue, and this feels similar. When that is merged in i think if you merge main into this it may fix the remaining issue you are having.

pngwn (approved):
Looks great! Thanks for fixing this and adding those extra tests @aliabid94!

arose26:
So, how do I change the tab programmatically? (not from a button, just from code).
I've tried changing the tab instance's 'selected' property directly but it doesn't do the job. gr.Tabs.update(selected=newid) doesn't work either.

avfranco-br:
> So, how do I change the active tab programmatically? (not from a button, just from code). I've tried changing the tab instance's 'selected' property directly but that only works BEFORE calling launch. After launch, it doesn't do the job. gr.Tabs.update(selected=newid) doesn't work either. I want to change selected tab in real time, from python code.

Hi @arose26 , have you found a fix for this issue yet? I'm having the same issue. Thanks, Alexandre

ben-directai:
Also experiencing this issue

Issue #7878, opened 2024-03-28
Dropdown area should be clickable to open


dawoodkhan82:
The dropdown's arrow and area to the right of it is not clickable and will not open the dropdown.

![image](https://github.com/gradio-app/gradio/assets/12725292/72d7a2c7-3c36-409c-851c-1015d7d6de47)

PR #7918, opened 2024-04-02
ensure entire dropdown is clickable

dawoodkhan82:
## Description

Fixes: The dropdown's arrow and area to the right of it is not clickable and will not open the dropdown.

Closes: #7878

## 🎯 PRs Should Target Issues

## Tests


pngwn (approved):
Looks great! Thanks @dawoodkhan82!

js/dropdown/shared/Dropdown.svelte
```diff
@@ -250,9 +250,12 @@
 
 <style>
 	.icon-wrap {
+		position: absolute;
+		top: var(--size-2);
```
aliabid94, line 254: Nice! My concern is that fixing the top position to be --size-2 will not render perfectly in themes that have a different font size. For example, this theme is set text-md to be 18px and the arrow is not vertically centered.

Screenshot 2024-04-05 at 12 58 16 PM

Would ideally be properly centered vertically

aliabid94: https://chat.openai.com/share/d5a7ada5-93e5-4119-86ae-c2d3bec553b4

PR #7924, opened 2024-04-03
Clarify the streaming response documentation

justalphie:
## Description

You cannot simply yield one new token at a time, you have to yield the new full response.

### DIFF
--- a/guides/04_chatbots/01_creating-a-chatbot-fast.md
+++ b/guides/04_chatbots/01_creating-a-chatbot-fast.md
@@ -73,8 +73,7 @@ gr.ChatInterface(alternatingly_agree).launch()
 
 ## Streaming chatbots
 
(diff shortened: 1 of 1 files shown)


guides/04_chatbots/01_creating-a-chatbot-fast.md
```diff
@@ -74,6 +74,7 @@ gr.ChatInterface(alternatingly_agree).launch()
 ## Streaming chatbots
 
 If in your chat function, you use `yield` to generate a sequence of responses, you'll end up with a streaming chatbot. It's that simple!
+Note that each yielded response replaces the previous one (and is not appended to the previously yielded content).
```
pngwn, line 77: I think we can rephrase the first paragraph instead of adding an additional one.

justalphie: Feedback addressed ;-)

freddyaboulton (approved):
Thanks @justalphie !

Issue #7921, opened 2024-04-03
Button name changes to undefined


kota-iizuka:
### Describe the bug

When you start the server with the script below, there is a button named `Run` at first, but if you click the `Run` button many times in a row, the name changes to `undefined`.

- The number of times it changes to `undefined` is approximately 2 to 10 times, and the number of times varies each time.
- In `debug=True` mode the console does not change and there are no error messages on the console.
- When I restart the browser it goes back to the `Run` button, but after a few clicks it becomes `undefined` again.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

enable_btn = gr.Button(interactive=True)
disable_btn = gr.Button(interactive=False)

def run():
    for _ in range(100):
        yield disable_btn
    yield enable_btn
    return

with gr.Blocks() as demo:
    btn = gr.Button()
    btn.click(run, None, [btn])

demo.launch()
```

### Screenshot

![image](https://github.com/gradio-app/gradio/assets/64062831/1b735cdb-c687-42ba-b690-23291d64fbe9)

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Linux
gradio version: 4.22.0
gradio_client version: 0.13.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.3.0
fastapi: 0.110.1
ffmpy: 0.3.2
gradio-client==0.13.0 is not installed.
httpx: 0.27.0
huggingface-hub: 0.22.2
importlib-resources: 6.4.0
jinja2: 3.1.3
markupsafe: 2.1.5
matplotlib: 3.8.3
numpy: 1.26.4
orjson: 3.10.0
packaging: 24.0
pandas: 2.2.1
pillow: 10.3.0
pydantic: 2.6.4
pydub: 0.25.1
python-multipart: 0.0.9
pyyaml: 6.0.1
ruff: 0.3.5
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.12.0
typing-extensions: 4.10.0
uvicorn: 0.29.0
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2024.3.1
httpx: 0.27.0
huggingface-hub: 0.22.2
packaging: 24.0
typing-extensions: 4.10.0
websockets: 11.0.3
```

### Severity

[...]

PR #7926, opened 2024-04-04
Fixes streaming event race condition

aliabid94:
Happened because of how we were closing event streams. Added a test that blasts lots of events rapidly so we don't run into this again. Test by running demo/rapid_generation on main and this branch.

Fixes: #7921


freddyaboulton (approved):
Nice fix @aliabid94 !

yvrjsharma:
I can confirm that I no longer see the issue in the [duplicated (private atm) mPLUGOwl/Doc](https://huggingface.co/spaces/ysharma/DocOwl) Space when using the gradio wheels from this PR. Thanks @aliabid94 !

pngwn:
Can you explain what the race condition is here? It isn't obvious from the code or tests.

pngwn (approved):
Looks good otherwise! Thanks @aliabid94!

aliabid94:
race condition was that sometimes the client would receive the `close_stream` event from the backend before it had finished processing pending callbacks. This would cause the callback to be called with `close_stream`, but we shouldn't be calling the callback with that message, as would cause issues with apply_diff_stream

dawoodkhan82:
Thanks for fixing @aliabid94

keenranger:
This helped me streaming agent a lot! TY!

PR #7936, opened 2024-04-04
Restore Markdown formatting for Chatbots, MarkdownCode

aliabid94:
The MarkdownCode object does not format several of its Markdown elements, including headers, lists, backquotes because it doesn't use the `prose` class. Added.

Before:
Screenshot 2024-04-04 at 1 28 27 PM

After: 
Screenshot 2024-04-04 at 1 27 56 PM


pngwn (commented):
This messing up the spacing in both the data frame and the chatbot. We'll need to fix that somehow. 

It would also be good to modify one of the chatbot stories to use some lore complex markdown so we can catch any regressions or strangeness since the markdown code component is used in a few places.

aliabid94:
I fixed the extra spacing at the bottom of each markdown blob. The text size has gotten 2 points smaller in the chatbot because we use --text-md in prose (14px), instead of the usual 16px used for buttons. But I think almost all the visual changes make sense, this makes MarkdownCode look consistent with Markdown. 

Can you point to specific issues?

Could add UI test that's more visually complex.

pngwn:
Was just that bottom spacing I think. Other changes look fine to me, I only noticed when reviewing the UI tests.

dawoodkhan82:
lgtm! I don't notice any breaking style changes

dawoodkhan82 (approved):

Issue #7934, opened 2024-04-04
UI not displaying properly because path to theme.css is not using provided root_path


PaulClas:
I have tried all the solutions listed in this issue as well as #3798, #3662, #3482, #3472 and https://stackoverflow.com/questions/61641514/css-not-recognized-when-served-with-fastapi. 

Solutions that I have tried include:

- Mounting the subapp to my FastAPI app
- Providing a root_path to the FastAPI declaration
- Providing a root_path to the mount_gradio_app() method
- Using both the mount function from FastAPI in conjunction with the mount_gradio_app()
- Downgrading and upgrading gradio to 3.50.1, 4.19.0, 4.20.0, 4.21.0, 4.25.0

My issue is the same as stated in the issue title of #3472. When running behind a proxy and load balancer, the gradio app requests the theme.css at the root of the url and not after the url+prefix. I have tried providing the root_path at the app.mount_gradio_app(root_path=) level and at the app=FastAPI(rooth_path=) level.

The gradio app runs at the provided root_path but the theme.css file is requested at the url without the root_path:

<img width="1170" alt="Screenshot 2024-04-04 at 3 29 04 PM" src="https://github.com/gradio-app/gradio/assets/6021481/d501ea55-c1f7-4743-bf01-03fccb71f641">

I have tried a few different releases from gradio and the gradio python client and have the problem persisting.

This is a sample version of the two solutions I have tried in my code that don't provide the right url to get the theme.css. Both ways work to have the gradio app running with the prefix at https//:abc.com/app/gradio but the theme is requested at https//:abc.com/gradio and fails.

Try 1:
```from fastapi import FastAPI
import gradio as gr
import uvicorn

app = FastAPI(root_path="/app")

@app.get('/')
def hello():
    return {"message":"hello"}

g_app = gr.Interface(
    lambda x: f"Hello {x}",
    "textbox",
    "textbox",
)

app = gr.mount_gradio_app(app, g_app, path="/gradio")

uvicorn.run(app,host="0.0.0.0",port=8080)
```

Try 2:
```
from fastapi import FastAPI
import gradio as gr
import uvicorn

app = FastAPI()

@app.get('/')
def hello():
    return {"message":"hello"}

g_app = gr.Interface(
    lambda x: f"Hello {x}",
    "textbox",
    "textbox",
)

app = gr.mount_gradio_app(app, g_app, path="/gradio",root_path="/app")

uvicorn.run(app,host="0.0.0.0",port=8080)
```

[...]

PR #7938, opened 2024-04-04
Handle the case of multiple headers when constructing root url

abidlabs:
While investigating https://github.com/gradio-app/gradio/issues/7934, I learned that the x-forwarded-host and x-forwarded-proto headers can contain multiple values if the request has passed through several proxies. 

E.g. see https://gitlab.com/gitlab-org/gitlab-pages/-/merge_requests/386

This PR modifies the `get_root_url()` method to handle these cases

### DIFF
--- a/gradio/route_utils.py
+++ b/gradio/route_utils.py
@@ -312,9 +312,7 @@ def get_root_url(
     def get_first_header_value(header_name: str):
         header_value = request.headers.get(header_name)
         if header_value:
-            return header_value.split(",")[
-                0
-            ].strip()
+            return header_value.split(",")[0].strip()
         return None
 
     if root_path and client_utils.is_http_url_like(root_path):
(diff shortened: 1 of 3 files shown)


gradio/route_utils.py
```diff
@@ -306,16 +306,26 @@ def get_root_url(
...
     In cases (2) and (3), We also check to see if the x-forwarded-proto header is present, and if so, convert the root url to https.
+    And if there are multiple hosts in the x-forwarded-host or multiple protocols in the x-forwarded-proto, the first one is used.
     """
+
+    def get_first_header_value(header_name: str):
+        header_value = request.headers.get(header_name)
+        if header_value:
+            return header_value.split(",")[
+                0
+            ].strip()  # Splits the value and returns the first element
```
abidlabs, line 317:
```suggestion
            ].strip()
```

freddyaboulton (approved):
Thanks @abidlabs !!

abidlabs:
Thanks @freddyaboulton will get this in.

Issue #7890, opened 2024-03-29
gradio client httpx session to reduce TIME_WAIT tcp connections


dodysw3:
- [x ] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
Calling predict() of simple task but large number of inputs generates large number of TIME_WAIT connections.

**Describe the solution you'd like**  
Have an option for predict() to maintain a single httpx session, or takes a httpx client instance from parameter.

**Additional context**  
 In some public cloud infrastructure, outgoing connection to public internet is done through shared NATed and the number of unique connections are very limited. TIME_WAIT in Linux takes 30s. When reaching limit, predict() hung as new connection waits for released TIME_WAIT connections.

dodysw3:
This is what I came with https://github.com/gradio-app/gradio/compare/main...dodysw3:gradio:http2client?expand=1 which works for my use case; it reuses the httpx client across multiple predict invocations and opportunistically use http2 if gradio server behind h2 reverse proxy -- which further reduce the number of connection needed; 

I don't know if this useful for others.

freddyaboulton:
Hi @dodysw3 , your changes make sense. If you open a PR one of us can review soon. One comment though is that I would not enable http2 connection by default as that adds a new requirement to the gradio client and the http/2 implementation in httpx is not as mature as http/1. 

Instead I think we can first try to instantiate a client with http2 support, if that raises an import error, enable the default http/1 client.

abidlabs:
Going to close this as not planned. If you are still encountering issues in the latest version of `gradio_client` and would like to open a PR to fix, please let us know and I can reopen.

Issue #7581, opened 2024-03-01
Gradio client: Calling sync() from within a running loop


Joystickplays:
### Describe the bug

I currently have a Flask app that uses Gradio API for inferencing. If I run it with normal Flask built-in WSGI server, it'll work.
However, I want to run it with gunicorn and gevent in my server. 

### Have you searched existing issues?  🔎

- [x] I have searched and found no existing issues

### Reproduction

```py
from gevent import monkey
monkey.patch_all()
from gradio_client import Client
from flask import Flask, send_file
import time

app = Flask(__name__)

imageclient = Client("https://diffusers-unofficial-sdxl-turbo-i2i-t2i.hf.space/")
@app.route("/gen")
def gen():
      result = imageclient.predict(
                  None,
                  "A sunny view of the park",	
                  0.9,	
                  2,
                  round(time.time()),	
                  api_name="/predict"
              )
      return send_file(result)

if __name__ == "__main__":
      app.run(host="0.0.0.0", port=5000)
```

Then run it with `gunicorn -k gevent -w 1 <app_name>:app --bind 0.0.0.0:5000`
Access 127.0.0.1:5000/gen

### Logs

```shell
Traceback (most recent call last):
  File "/home/Joystick/.local/lib/python3.10/site-packages/flask/app.py", line 1463, in wsgi_app
    response = self.full_dispatch_request()
  File "/home/Joystick/.local/lib/python3.10/site-packages/flask/app.py", line 872, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/home/Joystick/.local/lib/python3.10/site-packages/flask/app.py", line 870, in full_dispatch_request
    rv = self.dispatch_request()
  File "/home/Joystick/.local/lib/python3.10/site-packages/flask/app.py", line 855, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)  # type: ignore[no-any-return]
  File "/home/Joystick/ConversAItionalasFlask/app.py", line 414, in create_image
    result = imageclient.predict(
  File "/home/Joystick/.local/lib/python3.10/site-packages/gradio_client/client.py", line 389, in predict
    return self.submit(*args, api_name=api_name, fn_index=fn_index).result()
  File "/home/Joystick/.local/lib/python3.10/site-packages/gradio_client/client.py", line 1590, in result
    return super().result(timeout=timeout)
  File "/usr/lib/python3.10/concurrent/futures/_base.py", line 458, in result
[...]

PR #7960, opened 2024-04-08
reuse httpx client to reduce number of TIME_WAIT connections

dodysw3:
## Description

This PR changes gradio python client instance to reuse httpx client connection and uses http2 if possible. The intention is to keep number of TIME_WAIT TCP connections low to avoid hitting outbound connection limit on shared NAT gateway.

Closes: #7890


abidlabs:
Do we need to pass in an instance of a Client as an argument into the constructor? It seems relatively uncommon to need to recycle the httpx client across more than one Gradio client so it should be enough to create one at init time and use it for all calls to predict() and submit()

dodysw3:
> Do we need to pass in an instance of a httpx client as an argument into the constructor? It seems relatively uncommon to need to recycle the httpx client across more than one Gradio client so it should be enough to create one at init time and use it for all calls to predict() and submit()

agree, removing it, better also not exposing so we can switch to different protocol/library later without breaking change.

freddyaboulton:
Can we call `close` in the destructor?

abidlabs:
I don’t think we need to explicitly destroy the httpx async client. It should get garbage collected when all references to the Python Gradio Client end, right? Do you notice any issues if you don’t explicitly close it?

abidlabs:
Converting to draft for now, let us know if ready for review @dodysw3

dodysw3:
> Converting to draft for now, let us know if ready for review @dodysw3
Thanks apologize this taking much time.

> I don’t think we need to explicitly destroy the httpx async client. It should get garbage collected when all references to the Python Gradio Client object end, right? ...

Thanks, I realized my current understanding of how async works in Python is limited. In my use case, actually I did encountered deadlock (infinite loop) inside fsspec/asyn.py, specifically this block:
```
    while True:
        # this loops allows thread to get interrupted
        if event.wait(1):
            break
        if timeout is not None:
            timeout -= 1
            if timeout < 0:
                raise FSTimeoutError
```
when the client process exiting. It took me a while to debug; unfortunately after merging the dev branch with latest gradio main I'm no longer able to reproduce, I will try to revert to old code again later.

>Do you notice any issues if you don’t explicitly close it?

Only that the client TCP connections are still established when nothing references the Python Gradio Client object, I also found destructor method is not called immediately when references are removed (or out of scope), which might be the reason. In my case, it's not an issue as a single Gradio Client object is reused for all predict()s, but leaving connections open after finishes using Gradio Client object seems like a resource leak to me.

dodysw3:
Illustration of the problem, it seemed launching a new thread when the process is exiting will lead to deadlock, this is just Python general behaviour that leads to general advice not to run complex operations on destructors. In my PR case, running utils.synchronize_async() leads to fsspec to potentially (recreate) new thread to run its own event loop -- caused deadlock since the Python process is exiting -- and the event.wait(1) will never be signaled since the threads never starts. If this analysis correct, I don't wish to proceed with this PR until Gradio Python client API is to run it within try..finally or context with-statements to ensure the cleanup happens. Thanks @abidlabs @freddyaboulton for the exercise!

```
import asyncio
import threading
import time


class Client:
    def __del__(self):
        print("If this is called after loop END, it will deadlock")
        th = threading.Thread(target=lambda: time.sleep(1))
        th.start()
        print("this is never reached")


async def main():
    c = Client()
    c.c = c
    # intentionally create cyclical reference so it will not immediately be destroyed

if __name__ == '__main__':
    print("Begin")
    asyncio.run(main())
    print("End")
```

abidlabs:
Hmm I see let us know if you want us to take another look at this PR or if we should mark this as closed for now

dodysw3:
Hi @abidlabs reopening again this for a quick ask of your insight of this PR.

1. Adding optional parameter in gradio python Client() to explicitly enable persistent connection; as a more explicit desire for user to increase performance with compromise of not forgetting to call .close() afterward.
2. Added wrapper classes to httpx.Client that optionally disable context-manager to reused https.Client instance (it will be cleaned up when close() is called); this way existing codes don't have to care if it's in "persistent-mode" or not, the class will reuse httpx.Client if available or revert to old behaviour.

Documentation update can guide users on how to use this correctly. Thanks for looking!

freddyaboulton:
Hi @dodysw3 , we're going to remove the async client shortly so that only the sync client is used (https://github.com/gradio-app/gradio/issues/7581). Then we shouldn't have the deadlock async issue your seeing now. I think we should close this PR and re-open this once that's done.  Would still love your contribution. Thanks for your hard work on this.

Issue #7964, opened 2024-04-08
support custom pipelines in `gr.Interface.from_pipeline` 


not-lain:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
I can't run any custom pipelines on the `gr.Interface.from_pipeline` 
**Describe the solution you'd like**  
check inheritence and support the custom pipelines
**Additional context**  
I made a custom pipeline in this repo that can be used as a reference : https://huggingface.co/not-lain/MyRepo
how to use : 
```python 
# download an image (should have a number in it because our model is trained on mnist in that repo)
!wget https://e7.pngegg.com/pngimages/414/624/png-clipart-brand-black-and-white-angle-number-2-text-monochrome.png -O 2.png
```
```python
from transformers import pipeline
pipe = pipeline("image-classification", model="not-lain/MyRepo", trust_remote_code=True)
pipe("2.png") 
```
a list of other repos that have custom pipelines : 
* https://huggingface.co/briaai/RMBG-1.4 (image segmentation is not a supported task yet in the gr.Interface.from_pipeline) 
* https://huggingface.co/p1atdev/siglip-tagger-test-3

abidlabs:
Thanks @not-lain. Will look into your PR!

abidlabs:
I don't think we'll implement this, see discussion on https://github.com/gradio-app/gradio/pull/7965

PR #7965, opened 2024-04-08
Support custom pipelines in `gr.Interface.from_pipeline`

not-lain:
## Description

this will support custom pipelines

Closes: #7964

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/fresh-worms-cough.md
+++ b/.changeset/fresh-worms-cough.md
@@ -1,0 +1,5 @@
+---
+"gradio": minor
+---
+
+feat:Support custom pipelines in `gr.Interface.from_pipeline`
--- a/.changeset/tiny-friends-stick.md
+++ b/.changeset/tiny-friends-stick.md
@@ -1,0 +1,5 @@
+---
+"gradio": patch
+---
+
+feat:Use asyncio.Event to stop stream in heartbeat route
--- a/.changeset/twenty-paws-poke.md
+++ b/.changeset/twenty-paws-poke.md
@@ -1,0 +1,5 @@
+---
+"gradio": patch
+---
+
+feat:Fix task bug in python 3.9
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -2480,6 +2480,7 @@ Received outputs:
             self._queue.close()
             # set this before closing server to shut down heartbeats
             self.is_running = False
+            self.app.stop_event.set()
             if self.server:
                 self.server.close()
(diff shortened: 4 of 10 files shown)


not-lain:
I was running the following notebook : https://colab.research.google.com/drive/1i08Tw0VpfEJZ5TI6Hkq8h0rJHkcpzAZT?usp=sharing
and I ran into the following problem : 
```
  File "/usr/local/lib/python3.10/dist-packages/gradio/queueing.py", line 527, in process_events
    response = await route_utils.call_process_api(
  File "/usr/local/lib/python3.10/dist-packages/gradio/route_utils.py", line 261, in call_process_api
  File "/usr/local/lib/python3.10/dist-packages/gradio/blocks.py", line 1782, in process_api
    inputs = await self.preprocess_data(
  File "/usr/local/lib/python3.10/dist-packages/gradio/blocks.py", line 1494, in preprocess_data
    inputs_cached = await processing_utils.async_move_files_to_cache(
  File "/usr/local/lib/python3.10/dist-packages/gradio/processing_utils.py", line 410, in async_move_files_to_cache
    return await client_utils.async_traverse(
AttributeError: module 'gradio_client.utils' has no attribute 'async_traverse'
Keyboard interruption in main thread... closing server.
Killing tunnel 127.0.0.1:7860 <> https://c6807a1e4eb8056212.gradio.live/
```
I don't think it's related to this pr but wanted to share it, would be great if i can get a comment on this
Also friendly pinging @abidlabs for review

abidlabs:
Hi @not-lain sorry for the delay in responding. The problem I see is that this doesn't really add support for custom pipelines generally, correct? In the sense that if I were to define a new custom pipeline that does not share its name / task with an existing transformers pipeline, there's no way for gradio to infer what the right input / output components should be. This only adds support for custom pipelines that share their task name with an existing transformers pipeline, which seems fairly limiting?

not-lain:
@abidlabs no you are right on point, when defining a pipeline using the transformers library, a person can define whatever task they want (there are no constraints on the task name), I thought that by specifying the correct task might be a good feature to have.
I also agree with you on every bit of information that you mentioned.

abidlabs:
Thanks for confirming @not-lain -- I would suggest we close this PR then as it could be misleading to users as we're not really adding support for custom pipelines. And simply setting the task name to be the same as an existing pipeline as not a standard spec for pipelines as far as I'm aware.

Issue #7931, opened 2024-04-04
`gr.ChatInterface()` parameter `examples=` doesn't expect a `list[list[str]]`


WH-Yoshi:
### Describe the bug

It's a **minor** problem and actually doesn't cause any.

The parameter `examples=` from the `gr.ChatInterface()` expects a `type 'list[str] | list[dict[str, str | list]] | None'`, but when I give a `list[str]` it throws this ValueError because of additional input added:

ValueError: The parameter `examples` must either be a string directory or a list (if there is only 1 input component) or (more generally), a nested list, where each sublist represents a set of inputs.

As suggested in the Error, a nested list as following works: `list[list[str]` but it isn't expected by the parameter.

I suggest changing the description and expectations(?) of the `gr.ChatInterface()` `examples=` parameter, so no "Quick-fixes" is shown and we can't get confused with these slight problems.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr


def predict(message, history, choice):
    return "Hello " + choice

if __name__ == '__main__':
    # doesn't work
    examples = [
        "Hi",
        "Good morning",
    ]
    # works
    examples = [
        ["Hi"],
        ["Good morning"],
    ]
    chatbot = gr.Chatbot()
    textbox = gr.Textbox()

    with gr.Blocks() as demo:
        dropdown = gr.Dropdown(
            choices=["General", "Sir", "Madam"],
            value="General",
            label="Decoration",
        )
        gr.ChatInterface(
            fn=predict,
            chatbot=chatbot,
            textbox=textbox,
            examples=examples,
            additional_inputs=dropdown,
        )

    demo.launch()
```

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Windows
gradio version: 4.25.0
gradio_client version: 0.15.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.3.0
fastapi: 0.110.1
ffmpy: 0.3.2
gradio-client==0.15.0 is not installed.
httpx: 0.27.0
huggingface-hub: 0.22.2
importlib-resources: 6.4.0
jinja2: 3.1.3
markupsafe: 2.1.5
matplotlib: 3.8.3
numpy: 1.26.4
orjson: 3.10.0
packaging: 23.2
pandas: 2.2.1
pillow: 10.3.0
pydantic: 2.6.4
pydub: 0.25.1
[...]

PR #7966, opened 2024-04-08
Adding hint `list[list]` for example parameter

WH-Yoshi:
## Description

This PR add the hint type `list[list]` in the example parameter of the `gr.ChatInterface()` object, to resolve the "Expected parameter" message when a single or multiple additionnal_inputs are provided to the instance.

Closes: #7931 (closed)


abidlabs (approved):
LGTM thanks @WH-Yoshi!

PR #7967, opened 2024-04-08
Fix handling of single font name in theme

sd109:
## Description

Previous check failed because `str` is an instance of Iterable so was never correctly wrapped in `[...]`. Instead, providing a single font such as "Arial" would result in a font list of "A, r, i, a, l" in the bundled CSS.

## 🎯 PRs Should Target Issues

I think this fix is small enough to not require opening an issue, but happy to do so if desired.


aliabid94:
Makes sense, thanks @sd109!

Issue #7985, opened 2024-04-10
An environment variable to override the share=True/False option when launching


cocktailpeanut:
[o] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  

There are many gradio apps, and some apps launch with the `share=True` option and some don't.

But sometimes we just want to force `share=True`, WITHOUT having to change the code. For example, even if the app launched with `app.launch()` (no `share=True`), if we can pass in an environment variable that forces the gradio.live link that would be very convenient.

**Describe the solution you'd like**  

An environment variable named `GRADIO_SHARE`:

1. `GRADIO_SHARE=1 python app.py`: will launch the gradio.live share link even if the `app.py` launched the server with `app.launch()`.
2. `GRADIO_SHARE=0 python app.py`: will launch WITHOUT the gradio.live share link even if the `app.py` launched with the server with `app.launch(share=True)`

**Additional context**  

This feature should be beneficial to huggingface and gradio adoption since even the users who run the projects 100% locally can easily start using the gradio proxy through the gradio.live links.

abidlabs:
Closed via #7986

PR #7986, opened 2024-04-10
GRADIO_SHARE Environment Variable

cocktailpeanut:
## Description

This feature lets you override the `share` argument passed into the `launch()` method.

There are many gradio apps, and some apps have the `share=True` hardcoded and some don't. But sometimes we want to just take ANY gradio app and make them shareable via the gradio.live URL, even if the original source code does not specify `app.launch(share=True)`.

How it works:

1. `GRADIO_SHARE=1 python app.py`: will launch the gradio.live share link even if the `app.py` launched the server with `app.launch()`.
2. `GRADIO_SHARE=0 python app.py`: will launch WITHOUT the gradio.live share link even if the `app.py` launched with the server with `app.launch(share=True)`

## 🎯 PRs Should Target Issues

I've searched and couldn't find any similar issue. So just created one here: https://github.com/gradio-app/gradio/issues/7985

## Tests

I tested this locally by monkey patching my `site-packages` directly and confirm it's working as explained, but could not test by pulling from my PR fork (since the `share=True` option can only run on the versions released through PyPi).

Happy to do whatever is needed.

### DIFF
--- a/.changeset/free-things-wink.md
+++ b/.changeset/free-things-wink.md
@@ -1,5 +1,5 @@
 ---
-"gradio": minor
+"gradio": patch
 ---
 
 feat:GRADIO_SHARE Environment Variable
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -2246,9 +2246,10 @@ Received outputs:
         else:
             self.share = share
 
-        # GRADIO_SHARE environment variable for forcing 'share=True'
-        if bool(os.getenv("GRADIO_SHARE")):
-            self.share = True
+        # GRADIO_SHARE environment variable for forcing 'share=True' or 'share=False'
+        share_env = os.getenv("GRADIO_SHARE")
+        if share_env is not None:
+            self.share = bool(share_env)
 
         # If running in a colab or not able to access localhost,
         # a shareable link must be created.
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -2247,6 +2247,8 @@ Received outputs:
             self.share = share
 
         # GRADIO_SHARE environment variable for forcing 'share=True' or 'share=False'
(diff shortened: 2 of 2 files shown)


abidlabs:
Thanks @cocktailpeanut for this contribution. In principle, the environment variable looks good but I’d request two changes in order to be consistent with other Gradio envs:

(1) The env should be set to “True” / “False” instead of 1/0 e.g. as is the case with GRADIO_CACHE_EXAMPLES

(2) More importantly, an explicitly provided parameter should always take precedence over an env variable. So this env variable should only take effect if no explicit value is provided for the share parameter. This is the case with every other env variable in Gradio

cocktailpeanut:
@abidlabs Just updated the code so the explicitly provided parameter takes precedence over the env variable. 

By the way just a heads up, I just assumed the convention was to use 0/1 for boolean because I saw `GRADIO_DEBUG` using 0/1 in the same file. Maybe this needs to be changed as well for consistency? https://github.com/gradio-app/gradio/blob/main/gradio/blocks.py#L2395

abidlabs:
Ah okay I missed that! I guess then it’s not a big deal to use 0/1.  

Can you add mention of this env variable in the docstring?

cocktailpeanut:
@abidlabs 

> Ah okay I missed that! I guess then it’s not a big deal to use 0/1.

I'll just leave it as "True"/"False" strings as per your original suggestion, since i've already changed it to work this way, and is more straightforward anyway.

> Can you add mention of this env variable in the docstring?

You mean like this, correct? https://github.com/gradio-app/gradio/pull/7986/commits/299b17805f13de226cc6ecc18e9447219f6e7a77

freddyaboulton (approved):
Thanks @cocktailpeanut and @abidlabs

PR #7998, opened 2024-04-10
Restore chatbot formatting

aliabid94:
When we added the prose to MarkdownCode, some of the chatbot styling changed. We want to preserve the larger text size and the indentation on lists in prose was too large. Lmsys requested a reversion. Fixed.

Before:

Screenshot 2024-04-10 at 3 19 48 PM

After:

Screenshot 2024-04-10 at 3 19 56 PM


abidlabs:
LGTM thanks @aliabid94

abidlabs (approved):

PR #8003, opened 2024-04-11
Update `Client` usage examples

hannahblair:
## Description

Updates any examples of client usage from `await client()` to `await Client.create()`


abidlabs:
Nice, is this ready for review or still draft?

client/js/README.md
```diff
@@ -21,9 +21,9 @@ The client function connects to the API of a hosted Gradio space and returns an
 The simplest example looks like this:
 
 ```ts
-import { client } from "@gradio/client";
+import { Client } from "@gradio/client";
```
abidlabs, line 24: We should update the section header to be `Client` instead of `client`, and in the paragraph above, replace "client function" with e.g. `Client.client` factory function

client/js/README.md
```diff
@@ -21,9 +21,9 @@ The client function connects to the API of a hosted Gradio space and returns an
...
 
-const app = await client("user/space-name");
+const app = await Client.create("user/space-name");
 const result = await app.predict("/predict");
 ```
```
abidlabs, line 29: There's an example below:

```
client("user/space-name");
```

which should be

```
Client.create("user/space-name");
```

client/js/README.md
```diff
@@ -21,9 +21,9 @@ The client function connects to the API of a hosted Gradio space and returns an
 The simplest example looks like this:
```
abidlabs, line 21: "The JavaScript Gradio Client exposes two named imports, client and duplicate" needs to be updated

client/js/README.md
```diff
@@ -64,7 +64,7 @@ Applications hosted on Hugging Face spaces can be in a number of different state
 ```ts
 import { client, type SpaceStatus } from "@gradio/client";
```
abidlabs, line 65:
```suggestion
import { Client, type SpaceStatus } from "@gradio/client";
```

client/js/README.md
```diff
@@ -124,9 +124,9 @@ const result = await app.predict("/predict");
 The `payload` argument is generally optional but this depends on the API itself. If the API endpoint depends on values being passed in then it is required for the API request to succeed. The data that should be passed in is detailed on the "View API" page of a space, or accessible via the `view_api()` method of the client.
```
abidlabs, line 124: I think the original comment is misleading. Most endpoints do require input data.
```suggestion
The `payload` argument is generally required but this depends on the API itself. If the API endpoint depends on values being passed in then the argument is required for the API request to succeed. The data that should be passed in is detailed on the "View API" page of a space, or accessible via the `view_api()` method of the client.
```

client/js/README.md
```diff
@@ -255,9 +255,9 @@ console.log(api_info);
 The `config` property contains the configuration for the gradio application you are connected to. This object may contain useful meta information about the application.
 
 ```ts
-import { client } from "@gradio/client";
+import { Client } from "@gradio/client";
 
-const app = await client("user/space-name");
+const app = await Client.create("user/space-name");
 console.log(app.config);
 ```
```
abidlabs, line 263: At the bottom of this document, we discuss `duplicate`. I believe this has been refactored as well, correct?

client/js/src/types.ts
```diff
@@ -139,6 +139,23 @@ export interface Config {
...
+
+export interface Payload {
+	data: unknown[];
+	fn_index: number;
+	event_data: unknown;
+	time?: Date;
+	trigger_id: any;
+}
+
+export interface Payload {
```
abidlabs, line 153: Unless I'm missing something, isn't this duplicated twice?

guides/08_gradio-clients-and-lite/02_getting-started-with-the-js-client.md
```diff
@@ -66,7 +66,7 @@ The `@gradio/client` exports another function, `duplicate`, to make this process
 `duplicate` is almost identical to `client`, the only difference is under the hood:
```
abidlabs, line 66: Is the duplicating section correct or have we refactored duplicate?

js/app/src/Index.svelte
```diff
@@ -65,6 +65,7 @@
 <script lang="ts">
 	import { onMount, setContext, createEventDispatcher } from "svelte";
 	import type { SpaceStatus } from "@gradio/client";
+	import { Client as ClientClass } from "@gradio/client";
```
abidlabs, line 68: Is ClientClass used anywhere?

abidlabs:
Left a few places in the docs that need to be updated. Should be good to merge once these are addressed

abidlabs (approved):
LGTM @hannahblair! Thanks for updating the docs including the one in Chinese

One tiny nit: many of the code snippets use:

```js
const app = Client.create(...)
```

I feel like it would be slightly clearer to name the object `client` instead of `app`

```js
const client = Client.create(...)
```

and it would be consistent with the python docs. Ofc not a big deal but pointing out in case you feel the same way..

abidlabs:
Did we rename  `Client.create()` to `Client.connect()`?

abidlabs (approved):
LGTM @hannahblair!

abidlabs:
Going to merge this in so that we have a single PR with all of the changes

PR #8009, opened 2024-04-12
Fix gr.Examples in reload mode

freddyaboulton:
## Description

Noticed that if you ran a demo with gr.Examples in reload mode, the app would crash after reloading. The UI would update though. You can test this with `calculator_blocks` demo.

The issue is that the `component_map` in `Dataset` would be empty after reloading. It is a bit strange because I verified that the `Dataset` gets a non-empty map in `init.ts`, but then the size of the map would decrease to 0 and then the error would happen. So I fixed this by storing the initial component map in a new variable.

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/true-teeth-add.md
+++ b/.changeset/true-teeth-add.md
@@ -1,0 +1,6 @@
+---
+"@gradio/dataset": patch
+"gradio": patch
+---
+
+fix:Fix gr.Examples in reload mode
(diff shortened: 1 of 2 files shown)


freddyaboulton:
Will clean up the lint but want to make sure this makes sense @pngwn

freddyaboulton:
@pngwn found the proper fix in #8028

Issue #8013, opened 2024-04-12
Double Output on file Upload


TALLEC-Scott:
### Describe the bug

When you only upload a file in the multimodal interface, an extra empty box gets shown. 

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr
def fn(message, history):
    return "wow"

gr.ChatInterface(
    fn=fn,
    multimodal=True,
).launch()


```

### Screenshot

![image](https://github.com/gradio-app/gradio/assets/63871153/26f97968-7268-47ed-87a7-43462f696a39)

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Windows
gradio version: 4.26.0
gradio_client version: 0.15.1

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.3.0
fastapi: 0.110.1
ffmpy: 0.3.2
gradio-client==0.15.1 is not installed.
httpx: 0.27.0
huggingface-hub: 0.22.2
importlib-resources: 6.4.0
jinja2: 3.1.3
markupsafe: 2.1.5
matplotlib: 3.8.4
numpy: 1.26.4
orjson: 3.10.0
packaging: 24.0
pandas: 2.2.1
pillow: 10.3.0
pydantic: 2.6.4
pydub: 0.25.1
python-multipart: 0.0.9
pyyaml: 6.0.1
ruff: 0.3.5
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.12.0
typing-extensions: 4.10.0
uvicorn: 0.29.0
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2024.3.1
httpx: 0.27.0
huggingface-hub: 0.22.2
packaging: 24.0
typing-extensions: 4.10.0
websockets: 11.0.3
```

### Severity

I can work around it

PR #8014, opened 2024-04-12
Fix multimode interface double box on file upload

TALLEC-Scott:
## Description

When file is provided but string in input is empty it still gets added to message history which renders empty string and generates a senseless box. See issue for example on this.

Closes: #8013


gradio/chat_interface.py
```diff
@@ -202,7 +202,10 @@ def __init__(
             with Group():
                 with Row():
                     if textbox:
-                        textbox.container = False
+                        if self.multimodal:
```
freddyaboulton, line 205: This makes the textbox look squished in the multimodal case

image

I think we just need `submit_btn=None` in this case

TALLEC-Scott: That's strange, I flipped the lines. Good catch. Should be good now.

TALLEC-Scott (commented):
Check  my newest commit

freddyaboulton (approved):
Thanks @TALLEC-Scott !

dawoodkhan82 (approved):
lgtm! Thanks for the fix @TALLEC-Scott

Issue #7769, opened 2024-03-20
The image uploaded in the multimodal chatbot might be too small to be clearly visible.


yvrjsharma:
### Describe the bug

I noticed that the size of the image I uploaded in the multimodal chatbot is quite small. I am not entirely certain if this is an issue or if it was intended to be that way. However, I feel that the image could be more visible and impactful if it was larger.

cc: @abidlabs 

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr
import os
import time

# Chatbot demo with multimodal input (text, markdown, LaTeX, code blocks, image, audio, & video). Plus shows support for streaming text.


def print_like_dislike(x: gr.LikeData):
    print(x.index, x.value, x.liked)

def add_message(history, message):
    for x in message["files"]:
        history.append(((x["path"],), None))
    if message["text"] is not None:
        history.append((message["text"], None))
    return history, gr.MultimodalTextbox(value=None, interactive=False, file_types=["image"])

def bot(history):
    response = "**That's cool!**"
    history[-1][1] = ""
    for character in response:
        history[-1][1] += character
        time.sleep(0.05)
        yield history


with gr.Blocks() as demo:
    chatbot = gr.Chatbot(
        [],
        elem_id="chatbot",
        bubble_full_width=False,
        #avatar_images=(None, (os.path.join(os.path.dirname(__file__), "files/avatar.png"))),
    )

    chat_input = gr.MultimodalTextbox(interactive=True, file_types=["image"], placeholder="Enter message or upload file...", show_label=False)
    chat_msg = chat_input.submit(add_message, [chatbot, chat_input], [chatbot, chat_input], queue=False).then(
        bot, chatbot, chatbot, api_name="bot_response"
    )
    chat_msg.then(lambda: gr.Textbox(interactive=True), None, [chat_input], queue=False)
    chatbot.like(print_like_dislike, None, None)

demo.queue()
if __name__ == "__main__":
    demo.launch()

```

### Screenshot

How it is coming up for us :

https://github.com/gradio-app/gradio/assets/48665385/8038a7c4-c1a9-4864-9e8e-a4766b4f7b4d

How ChatGPT does it atm:

![image](https://github.com/gradio-app/gradio/assets/48665385/c568849f-f56f-4327-831d-0c04770f7822)

### System Info

```shell
Google Colab
```

### Severity

I can work around it

nxphi47:
+1

[...]

PR #8025, opened 2024-04-15
Fixes Chatbot Image Sizing

dawoodkhan82:
## Description
Increases size of uploaded images for chatbot.

Closes: #7769

## 🎯 PRs Should Target Issues

## Tests


abidlabs:
I wonder if this should be configurable?

js/chatbot/shared/ChatBot.svelte
```diff
@@ -325,7 +325,10 @@
 
 	.message-wrap > div :not(.avatar-container) :global(img) {
 		border-radius: 13px;
+		margin: var(--size-2);
+		width: 400px;
 		max-width: 30vw;
+		max-height: auto;
```
pngwn, line 331: Do you think we should have different sized avatars for mobile / desktop?

dawoodkhan82: This actually doesn't affect avatar images, only images uploaded into the chatbot.

Are the avatar images on mobile too large right now?

pngwn: Ah sorry. My mistake. 

No I think avatar images are fine on mobile.

dawoodkhan82:
> I wonder if this should be configurable?

@abidlabs I don't think so, do we offer that much control over sizing on other components? I think once we add the option to click and zoom into any image in the chatbot, that would solve any "sizing" issues.

pngwn (approved):
Looks great to me! Mentioned in person that it might be worth checking the behaviour on very small screens but otherwise good to go. Thanks @dawoodkhan82!

dawoodkhan82:
Thanks for the review @pngwn. Seems like it reacts fine to small screen sizes

Screenshot 2024-04-15 at 4 48 08 PM

freddyaboulton:
Nice test image @dawoodkhan82 hehe

yvrjsharma:
Hello @dawoodkhan82, thanks for looking into this. I have noticed that the image resizing works perfectly for large images and videos. However, I have observed that for smaller images, they appear stretched. Is there any way to prevent this from happening?

https://github.com/gradio-app/gradio/assets/48665385/8cb65bd3-7ef2-4f90-a8d6-1f4c81d7ca9a

PR #8046, opened 2024-04-17
round [x, y, w, h] before cropping to avoid unexpected interpolation on pixel values

ernestchu:
## Description

**round [x, y, w, h] before cropping to avoid unexpected interpolation on pixel values**
  
I'm working on some demo of image watermarking. Cropping the test image using `ImageEditor` ruins the watermark while my watermark should have the ability to withstand cropping.

After some investigation into both front-end and back-end, I figured the problem is that the current implementation of cropping using `CanvasRenderingContext2D: drawImage()` in front-end would produce undesired interpolation on pixel values **if the canvas dimensions are not integers**.

Therefore, a simple fix is to round the cropping parameters before calling `$pixi.get_layers()`

### DIFF
--- a/.changeset/green-keys-invite.md
+++ b/.changeset/green-keys-invite.md
@@ -1,0 +1,6 @@
+---
+"@gradio/imageeditor": patch
+"gradio": patch
+---
+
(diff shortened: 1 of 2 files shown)


freddyaboulton (approved):
Thanks @ernestchu

PR #8050, opened 2024-04-17
Update `typer` to drop `[all]` as it is no longer needed

joennlae:
## Description

Source: https://typer.tiangolo.com/release-notes/#0121


freddyaboulton (approved):
Thanks @joennlae

PR #8056, opened 2024-04-17
Using keys to preserve values between reloads

aliabid94:
Added the concept of a key to a Block, such that any Block that has the same key between re-renders has its value preserved between renders. Will be needed specifically for render PR.

Implemented in dev mode by automatically assigning every Block a generated key and giving similarly positioned blocks have the same key across renders. See below how re-renders do not change the value of the components. 

BEFORE:
Recording 2024-04-17 at 18 14 02

AFTER:
Recording 2024-04-17 at 18 10 29

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -121,6 +121,7 @@ class Block:
         self.state_session_capacity = 10000
         self.temp_files: set[str] = set()
         self.GRADIO_CACHE = get_upload_folder()
+        self.key: int | str | None = None
         # Keep tracks of files that should not be deleted when the delete_cache parmaeter is set
         # These files are the default value of the component and files that are used in examples
         self.keep_in_cache = set()
@@ -1892,6 +1893,7 @@ Received outputs:
             block_config["component_class_id"] = getattr(
                 block, "component_class_id", None
             )
+            block_config["key"] = block.key
 
             if not block.skip_api:
                 block_config["api_info"] = block.api_info()  # type: ignore
@@ -2128,6 +2130,11 @@ Received outputs:
             raise ValueError("`blocked_paths` must be a list of directories.")
 
         self.validate_queue_settings()
+
+        if self.dev_mode:
+            for block in self.blocks.values():
+                if block.key is None:
+                    block.key = f"__{block._id}__"
 
         self.config = self.get_config_file()
         self.max_threads = max_threads
--- a/gradio/utils.py
+++ b/gradio/utils.py
@@ -103,13 +103,6 @@ class BaseReloader(ABC):
     def running_app(self) -> App:
         pass
 
-    def queue_changed(self, demo: Blocks):
-        return (
-            hasattr(self.running_app.blocks, "_queue") and not hasattr(demo, "_queue")
-        ) or (
-            not hasattr(self.running_app.blocks, "_queue") and hasattr(demo, "_queue")
-        )
-
     def swap_blocks(self, demo: Blocks):
         assert self.running_app.blocks  # noqa: S101
         # Copy over the blocks to get new components and events but
@@ -155,7 +148,10 @@ class SourceFileReloader(BaseReloader):
         self.change_event.set()
 
     def swap_blocks(self, demo: Blocks):
+        old_blocks = self.running_app.blocks
         super().swap_blocks(demo)
+        reassign_keys(old_blocks, demo)
+        demo.config = demo.get_config_file()
         self.alert_change()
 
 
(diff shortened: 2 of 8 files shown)


gradio/blocks.py
```diff
@@ -2129,6 +2131,11 @@ def reverse(text):
 
         self.validate_queue_settings()
 
+        if self.dev_mode:
+            for block in self.blocks.values():
+                if block.key is None:
+                    block.key = f"__{block._id}__"
```
freddyaboulton, line 2137: Can we make the id a hash of the component's config? That way it will be unique across re-renders unless something about the component changed and we don't need to do the `reassign_keys` logic.

js/app/src/Blocks.svelte
```diff
@@ -51,6 +52,23 @@
...
+		components.forEach((component) => {
+			if (component.key) {
+				const old_component = component_values_by_key[component.key];
+				if (old_component) {
+					component.props.value = old_component.props.value;
+				}
+			}
+		});
+	};
+
```
pngwn, line 71: Do you think it would be possible to do all of this in the `init` module? I think we could expose an `update_components` function where we could pass in components, and the old / new merging could be handled inside that module with all of the other component updates.

gradio/blocks.py
aliabid94: So what you're suggesting is that we use the hash of the config to decide what components are considered identical across re-renders. While we could use that logic in `utils.reassign_keys` to consider what components are equivalent, it's gonna be problematic because it's very likely two components have the exact same config, e.g.

```python
with gr.Blocks():
  a = gr.Textbox()
  b = gr.Textbox()
```

I think it makes more sense to rely on the position and type of components to assign identity. The logic in `utils.reassign_keys` can always be improved but keep in mind this is only for dev mode refreshes - in actual render blocks, we will require users to explicitly assign keys to maintain identity.

js/app/src/Blocks.svelte
aliabid94: sure, moved to init.ts

gradio/blocks.py
freddyaboulton: I see your point about duplicate keys but is that a concern for reload mode? Even if auto-generated ids collide, the component fetched in the frontend's `restored_keyed_values` would be correct since the config should determine the state of the component. 

The problem with using the position to determine equivalence is that if you insert a component "in between" or "before" already existing components, then the front-end state would get wiped.

This PR is already an improvement of what we currently do but wondering if we can create a better heuristic for equivalence from the get-go.

gradio/utils.py
```diff
@@ -155,7 +148,11 @@ def alert_change(self):
         self.change_event.set()
 
     def swap_blocks(self, demo: Blocks):
+        old_blocks = self.running_app.blocks
         super().swap_blocks(demo)
```
freddyaboulton, line 152: I think we need to take care of adding `max_file_size` from the old blocks here as well:

demo.max_file_size = old_blocks.max_file_size

js/app/src/Blocks.svelte
pngwn: Sorry I didn't mean just move the function but rather the old component state could be tracked in the init logic. 

We already have the previous component state. If we added an `update` method that took the new component / deps / layout, we could easily perform the logic all inside of the update method without needing to track the previous state outside of that module. 

This way we could also perhaps avoid some work on every component change and instead only do what is necessary (this may be minimal if anything).

gradio/blocks.py
aliabid94: > Even if auto-generated ids collide, the component fetched in the frontend's restored_keyed_values would be correct since the config should determine the state of the component.

I don't think this is the case. If initially I have:

```python
with gr.Blocks():
    boxA = gr.Textbox()
    boxB = gr.Textbox()
```
and I type something into boxA, and then I change boxA to:
```python
with gr.Blocks():
    boxA = gr.Textbox(label="A")
    boxB = gr.Textbox()
```
then the value that was in A would get moved to B, since that's the best match by config logic.

> This PR is already an improvement of what we currently do but wondering if we can create a better heuristic for equivalence from the get-go.

Sure, I think we can iterate on `reassign_keys` logic in other PRs, but since its only used in dev mode I think this can be a decent starting point.

gradio/utils.py
aliabid94: ok, handled in BaseReloader.swap_blocks

js/app/src/Blocks.svelte
aliabid94: Gotcha, makes sense! Implemented this in https://github.com/gradio-app/gradio/pull/8110 which targets this PR.

gradio/blocks.py
freddyaboulton: Ok sounds good!

freddyaboulton (commented):
Thanks for the great PR @aliabid94 ! 

One remaining issue I noticed is that this breaks custom component reload mode. I created a new component in `js/preview/test` (needs to be here until the next release because of #8066) and ran dev mode. Changes in the python file were not reflected in the UI. Confirmed it does work on main though.

pngwn:
I can also reproduce what @freddyaboulton is seeing.

aliabid94:
Okay I believe I fixed the maximum 6 refreshes bug and also the bug @freddyaboulton mentioned by simply closing the heartbeat connection on refreshes. Should be ready!

pnpm-lock.yaml
```diff
@@ -1478,6 +1478,25 @@ importers:
         specifier: ^3.28.0
         version: 3.28.0
 
+  js/preview/test/mycomponent/frontend:
```
freddyaboulton, line 1481: Revert this change please

freddyaboulton (approved):
Tested and works well @aliabid94 🔥 Nice

abidlabs:
Didn't do a deep review but it looks like the issues that I noticed are fixed. Nice!!

Issue #8070, opened 2024-04-19
Failure of Hot Reloading with Code Containing Multibyte Characters


zolgear:
### Describe the bug

**Summary:**

The changes made to the file `gradio/utils.py` are causing issues with hot reloading for applications using code containing multibyte characters.

**Details:**

Pull Request #7684 adds the open function to gradio/utils.py without specifying the file encoding, which impacts hot reload functionality when using code that contains multibyte characters.

gradio/utils.py
```python
NO_RELOAD = True


def _remove_no_reload_codeblocks(file_path: str):
    """Parse the file, remove the gr.no_reload code blocks, and write the file back to disk.
    Parameters:
        file_path (str): The path to the file to remove the no_reload code blocks from.
    """

    with open(file_path) as file:
        code = file.read()
```

**Related Links:**

- Related past issues: [#6063](https://github.com/gradio-app/gradio/issues/6063), [#6364](https://github.com/gradio-app/gradio/issues/6364)
- Related past Bug Fix: [#6622](https://github.com/gradio-app/gradio/pull/6622)

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
# File Save with UTF-8 encoding
import gradio as gr

with gr.Blocks() as demo:
    markdown = gr.Markdown('こんにちわ.')

demo.launch()
```

### Logs

```shell
(.env) PS D:\gradio-issue-hotreload> gradio .\demo.py
Watching: 'D:\gradio-issue-hotreload' 'D:\gradio-issue-hotreload'

Running on local URL:  http://127.0.0.1:7860

To create a public link, set `share=True` in `launch()`.
Changes detected in: D:\gradio-issue-hotreload\demo.py
Reloading demo failed with the following exception:
Traceback (most recent call last):
  File "D:\gradio-issue-hotreload\.env\Lib\site-packages\gradio\utils.py", line 266, in watchfn
    changed_in_copy = _remove_no_reload_codeblocks(str(changed))
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "D:\gradio-issue-hotreload\.env\Lib\site-packages\gradio\utils.py", line 173, in _remove_no_reload_codeblocks
    code = file.read()
           ^^^^^^^^^^^
UnicodeDecodeError: 'cp932' codec can't decode byte 0x8f in position 125: illegal multibyte sequence
```

### System Info

```shell
Gradio Environment Information:
[...]

PR #8075, opened 2024-04-19
Fix causing Hot Reload (#8070)

zolgear:
## Description

Fixes bug causing Hot Reload failure in code with non-English languages due to missing encoding specification.

I hardcoded UTF-8 because I couldn't figure out how to specify the encoding, similar to the fix in `gradio\cli\commands\reload.py`.

Closes: #8070


freddyaboulton:
Thank you @zolgear !

PR #8077, opened 2024-04-19
Fix an enconding issue in gradio/components/code.py

3210448723:
## Description

Modified the postprocess function in gradio/components/code.py to handle file encoding properly. Added encoding="utf-8" parameter to the open function when reading code files. I hardcoded UTF-8 because I couldn't figure out how to specify the encoding, similar to the fix in [gradio-app#8075](https://github.com/gradio-app/gradio/pull/8075).

Closes: #(8076)

### DIFF
--- a/gradio/components/code.py
+++ b/gradio/components/code.py
@@ -151,7 +151,7 @@ class Code(Component):
         if value is None:
             return None
         elif isinstance(value, tuple):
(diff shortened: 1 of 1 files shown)


freddyaboulton (approved):

freddyaboulton:
Thanks @3210448723

Issue #8017, opened 2024-04-15
Do not raise the cache example error when `fn` or `outputs` is not set and `GRADIO_CACHE_EXAMPLES` is `lazy`


hysts:
- [x] I have searched to see if a similar issue already exists.

([internal slack](https://huggingface.slack.com/archives/C02QZLG8GMN/p1713149850584069))

When running this code with `GRADIO_CACHE_EXAMPLES=lazy python app.py`,
```
import gradio as gr


with gr.Blocks() as demo:
    text = gr.Textbox()
    out = gr.Textbox()
    gr.Examples(examples=["aaa", "bbb"], inputs=text)
demo.queue().launch()
```
the following error is raised.
```
Traceback (most recent call last):
  File "/tmp/temp/app.py", line 7, in <module>
    gr.Examples(examples=["aaa", "bbb"], inputs=text)
  File "/home/ubuntu/.virtualenvs/py31011/lib/python3.10/site-packages/gradio/helpers.py", line 55, in create_examples
    examples_obj = Examples(
  File "/home/ubuntu/.virtualenvs/py31011/lib/python3.10/site-packages/gradio/helpers.py", line 153, in __init__
    raise ValueError("If caching examples, `fn` and `outputs` must be provided")
ValueError: If caching examples, `fn` and `outputs` must be provided
```

But I think it would be better not to raise this error when `fn` or `outputs` is not provided even if `GRADIO_CACHE_EXAMPLES` is set to `lazy`. The current behavior is problematic on ZeroGPU because `GRADIO_CACHE_EXAMPLES=lazy` is now the default on ZeroGPU environment and it's unexpected for developers to get this error.
(It seemed that one can still avoid this error by explicitly setting `cache_examples=False`, but I find it a bit tedious.)

PR #8087, opened 2024-04-21
Fix `GRADIO_CACHE_EXAMPLES` environment variable to only take effect if `fn` and `outputs` parameters are provided

abidlabs:
Closes: https://github.com/gradio-app/gradio/issues/8017

Repro:

```py
import gradio as gr
import os

os.environ["GRADIO_CACHE_EXAMPLES"] = "lazy"

with gr.Blocks() as demo:
    text = gr.Textbox()
    out = gr.Textbox()
    gr.Examples(examples=["aaa", "bbb"], inputs=text)

demo.launch()
```

which would previously throw an error

@hysts would appreciate if you can confirm the fix for ZeroGPU Spaces

### DIFF
--- a/.changeset/puny-sites-marry.md
+++ b/.changeset/puny-sites-marry.md
@@ -1,0 +1,5 @@
+---
+"gradio": minor
+---
+
+feat:Fix `GRADIO_CACHE_EXAMPLES` environment variable to only take effect if `fn` and `outputs` parameters are provided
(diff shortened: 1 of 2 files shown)


hysts (approved):
Thanks! Tested this on ZeroGPU and it worked. LGTM!

abidlabs:
Thanks for the quick review @hysts!

Issue #8096, opened 2024-04-22
Parameter `auth_dependency` in .launch() has no effect


Blubbaa:
### Describe the bug

Hello, 

I am trying to make use of the new authentication mechanism using the `auth_dependency` parameter. To my understanding this is a custom function which gets the FastAPI request as input and can implement arbitrary logic for authentication. In my case I actually just need to read the username from a forwarded header, i.e. `user = request.headers.get("x-forwarded-user")`. I would then expect this function to at least be called once when loading the app to check if the user is allowed to access.

However the gr.Request does not seem to contain the username afterwards (None), even if though the custom function should be called automatically via `get_current_user`, right? [#main/gradio/routes.py#L249](https://github.com/gradio-app/gradio/blob/ac30e07f9266f84088a0b8677a259a5b64c037c5/gradio/routes.py#L249)

So this might actually be two different things, one is what I am trying to achieve as described above, one is what I have noticed when trying to debug this. Setting the `auth_dependency=get_user` parameter in `demo.launch()` does not really have any effect, even if I return `None`, as indicated in the docs "If the function returns None for a specific request, that user is not authorized to access the app (they will see a 401 Unauthorized response).", I will still be able to access the app. See also repro below.

So my question, am I understanding and/or doing something wrong here, or does the parameter in the launch method have no effect? I am no expert in FastAPI etc. and am trying to avoid making the app more complex than I need to.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr
from fastapi import Request


def get_user(request: Request):
    # get username from forwarded header
    # ... but return None for testing
    return None


def get_user_proxy(request: gr.Request):
    user = get_user(request.request)
    print(f"Hello, {user}")


with gr.Blocks() as demo:
    gr.Button("Greet User").click(get_user_proxy)

# I would expect to not be able to access the app
demo.launch(auth_dependency=get_user)

```

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Windows
gradio version: 4.25.0
gradio_client version: 0.15.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.2.0
fastapi: 0.109.0
ffmpy: 0.3.1
gradio-client==0.15.0 is not installed.
httpx: 0.25.0
huggingface-hub: 0.20.2
importlib-resources: 6.1.1
jinja2: 3.1.3
markupsafe: 2.1.3
[...]

PR #8097, opened 2024-04-22
Respect `auth_dependency` parameter in `launch()`

abidlabs:
We were previously respecting `auth_dependency` only if the gradio app was mounted within a FastAPI app. I refactored the related code a bit to reduce code duplication & added a test.

Closes: https://github.com/gradio-app/gradio/issues/8096


freddyaboulton (approved):
Looks good to me @abidlabs !

abidlabs:
Thanks @freddyaboulton!

PR #8106, opened 2024-04-23
Pass Error status in /dev/reload stream

freddyaboulton:
## Description

Let clients know when an error happened during reload. Also refactored the route to support more than one client connecting simultaneously.

Logic is here and should not conflict with #8056 but will wait for #8056

There are two message types in the thread `reload` and `error`.

reload_status_stream

Edit: also errors are shown in the UI now

image

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/slow-bats-sniff.md
+++ b/.changeset/slow-bats-sniff.md
@@ -1,6 +1,6 @@
 ---
-"@gradio/app": minor
-"gradio": minor
+"@gradio/app": patch
+"gradio": patch
 ---
 
 feat:Pass Error status in /dev/reload stream
--- a/.changeset/slow-bats-sniff.md
+++ b/.changeset/slow-bats-sniff.md
@@ -1,0 +1,6 @@
+---
+"@gradio/app": minor
+"gradio": minor
+---
+
+feat:Pass Error status in /dev/reload stream
--- a/.changeset/slow-bats-sniff.md
+++ b/.changeset/slow-bats-sniff.md
@@ -1,5 +1,4 @@
 ---
-"@gradio/app": patch
 "gradio": patch
 ---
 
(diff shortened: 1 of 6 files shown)


pngwn:
I think it would be nice to show the full trace here but I guess the error toast is too small for that?

freddyaboulton:
I think so and they should have access to the server logs anyways (that's the idea on spaces as well I believe)

pngwn:
Would be cool if it was all viewable in the browser tho, on a single screen you often have to choose 2 of browser, terminal, editor. Can discuss later anyway, this is much better DX!

pngwn (approved):
Tested and works great! Thanks @freddyaboulton!

freddyaboulton:
Just added the python traceback to the console! Thanks for the review @pngwn

image

aliabid94 (approved):
Looks great, thanks freddy! Only thought is if I wonder if the error message should show more than just "Error reloading app", maybe the actual traceback?

EDIT: just saw that was discussed above.

Also, I wish the reload error toast would go away immediately as soon as I fixed the app, but that's probably a bit tricky.

freddyaboulton:
Thanks for the review @aliabid94 ! Will think about how we can close the error toast automatically!

pngwn:
> Also, I wish the reload error toast would go away immediately as soon as I fixed the app, but that's probably a bit tricky.

I think this should be possible. 

If each error message takes an `id` of the `fn_index`, then when we change the status of that `fn_index` to 'not an error' we can call a 'close_message' function on the Toast component (export one) that could filter out toasts with the given `id`. This is probably a good idea for all statuses tbh, because after the fn has run again those status are stale and irrelevant.

pngwn:
We need to refactor the progress stuff tbh it has gotten a bit out of hand, very hard to parse now. One problem @freddyaboulton and I came across is that the loading status centres around components, changing statuses for functions is harder than we felt it should be. Can look into it.

PR #8117, opened 2024-04-24
Add session hash to gr request

freddyaboulton:
## Description

Being able to pull the session_hash has come up twice recently: the command+R demo and I personally wanted build a demo where I use the session hash to save some files to disk and then clean them up with `unload`. To get the session hash, you have to use `await request.body()` which is not ideal since you'd have to make your handler async and it's also not clear you have to parse the body to get the session hash.

I think this should make it easier in the future for others.

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/great-plants-deny.md
+++ b/.changeset/great-plants-deny.md
@@ -1,0 +1,5 @@
+---
+"gradio": minor
+---
+
+feat:Add session hash to gr request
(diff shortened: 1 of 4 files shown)


pngwn (approved):
Ah this is awesome, thanks @freddyaboulton! I can think of loads of cases where this will be useful. This makes things much easier.

freddyaboulton:
Thanks @pngwn !

abidlabs:
Nice!

Issue #7103, opened 2024-01-22
@Gradio/client 0.10.1 Multiple changes to avoid errors in Node.js apps


fffiloni:
### Describe the bug

Hello ! 

I'm a working on a simple Node.js express app, which can call the gradio client to play with space apis.
I encountered several errors about "EventSource" not defined, and "window" not defined, errors i have been able to solve :) 

Here's the workaround for @gradio/client 0.10.1 and  Node.js v18.16.0

#1 First make sure in your package.json file you add
```
"type": "module"
```

#2 About EventSource

`npm install eventsource`

then in your server.js add

```javascript
import { createRequire } from 'node:module'
const require = createRequire(import.meta.url);
global.EventSource = require('eventsource');
```

then in file `/node_modules/@gradio/client/dist/index.js`
line 744 replace:
replace: 
```javascript
eventSource = EventSource_factory(url);
```

by adding .toString() to url arg:
```javascript
eventSource = EventSource_factory(url.toString());
```

—

same operation to line 1060
```javascript
event_stream = EventSource_factory(url);
```
by adding .toString() to url arg:
```javascript
event_stream = EventSource_factory(url.toString());
```

—
#3 About window not defined

It's not finished, now you will get window not defined event, so in file `/node_modules/@gradio/client/dist/index.js:1078`
line 1078

replace: 
```javascript
let fn2 = event_callbacks[event_id];
window.setTimeout(fn2, 0, _data);
```

by : 
```javascript
let fn2 = event_callbacks[event_id];
if (typeof window !== "undefined") {

  window.setTimeout(fn2, 0, _data);
} else {
  setTimeout(fn2, 0, _data);
}
```

—
So now in your main server.js, your code should be like this:

```javascript
import { client } from "@gradio/client";
import { createRequire } from 'node:module'
const require = createRequire(import.meta.url);
global.EventSource = require('eventsource');

async function test_servers(){
  const grapi_test = await client("gradio/hello_world");
  const apitest_result = await grapi_test.predict("/predict", [
    "John",
  ]);
  console.log(apitest_result.data);
}

test_servers()
```

It should now work, as it works on my hand, console should log: 
`[ 'Hello John!' ]`

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```javascript
import { client } from "@gradio/client";

async function test_servers(){
  const grapi_test = await client("gradio/hello_world");
  const apitest_result = await grapi_test.predict("/predict", [
    "John",
[...]

Issue #6352, opened 2023-11-09
`ReferenceError: EventSource is not defined` when API request in Node.js


gibiee:
### Describe the bug

I am testing API request in Node.js with @gradio/client.

But, `ReferenceError: EventSource is not defined` is occured when I execute the code by the command.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

`"type": "module"` must be set in `package.json`

test.js
```js
import { client } from "@gradio/client";

const app = await client("gradio/hello_world");
const result = await app.predict("/predict", [
  "John",
]);
console.log(result.data);
```

Then, execute `node test.js` in the command

### Logs

```shell
file:///C:/source/gradio_test/nodejs_client0.7.2/node_modules/@gradio/client/dist/index.js:733
            eventSource = new EventSource(url);
            ^

ReferenceError: EventSource is not defined
    at file:///C:/source/gradio_test/nodejs_client0.7.2/node_modules/@gradio/client/dist/index.js:733:13
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)

Node.js v20.9.0
```

### System Info

```shell
Node.js v20.9.0
@gradio/client 0.7.2
```

### Severity

I can work around it

gibiee:
I can run the code with simple tricks, but but it doesn't seem to be clean.

1. In `node_modules/@gradio/client/dist/index.js` 733 Line, add `.toString()` to url.
```js
// before
eventSource = new EventSource(url);

// after
eventSource = new EventSource(url.toString());
```

2. Declare `EventSource` globally.
```js
import { client } from "@gradio/client";
import { createRequire } from "module";
const require = createRequire(import.meta.url);
global.EventSource = require('eventsource');

const app = await client("gradio/hello_world");
const result = await app.predict("/predict", [
  "John",
]);
console.log(result.data);
```

Then I can get a output without the error.

PhantHive:
Hello,
I have the same issue in NodeJS with the following versions:
`gradio: 4.4.1`
`@gradio/client 0.8.1`

PS: Work around works fine for me as well

I wish all of you a nice day ☀️

gotDaijobu:
Also got the same issue with Node/Nuxt - without using a command. I had to install eventsource ("npm i eventsource") and use the work around. 

I also fixed the other call to EventSource Factory, line 1071:

`        event_stream = EventSource_factory(url.toString());`

pngwn:
we broke it, will release again early next week!

PR #8118, opened 2024-04-24
Add eventsource polyfill for Node.js and browser environments

hannahblair:
## Description

Replaces the Web API `eventsource`  with [eventsource](https://github.com/EventSource/eventsource) to support Node.js environments. Also addresses `window is not defined` errors in Node.js. 

To test, run the following in a node environment:

```
import { Client } from "@gradio/client";

const test = async () => {
    try {
	const app = await Client.connect("gradio/hello_world");
	const result = await app.predict("/predict", ["John"]);
	console.log("Result:", result);
    } catch (error) {
	console.error("Error:", error);
    }
};

(async () => {
	await test();
})();
```

Closes: #6352
Closes: #7103 

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/cold-spies-fail.md
+++ b/.changeset/cold-spies-fail.md
@@ -1,0 +1,5 @@
+---
+"gradio": patch
+---
+
+fix:fix: handling SIGINT correctly in reload.py, single entrance of block_thread in blocks.py
--- a/.changeset/major-zoos-read.md
+++ b/.changeset/major-zoos-read.md
@@ -1,0 +1,5 @@
+---
+"gradio": patch
+---
+
+fix:run python reload only if python file changed
--- a/.changeset/red-colts-burn.md
+++ b/.changeset/red-colts-burn.md
@@ -1,5 +1,6 @@
 ---
 "@gradio/client": patch
+"@gradio/upload": patch
 "gradio": patch
 ---
 
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -2427,18 +2427,21 @@ Received outputs:
             }
             analytics.launched_analytics(self, data)
 
-        # Block main thread if debug==True
-        if debug or int(os.getenv("GRADIO_DEBUG", "0")) == 1 and not wasm_utils.IS_WASM:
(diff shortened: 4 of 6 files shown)


pngwn (commented):
I think it would be good to run these test in node as well. We can doo this by creating a new script:

```json
"pnpm:test:client:node": "vitest client/js -c path/to/config --mode=test_node"
```

Then in the `vite.config.ts` check the `mode` is `node-test` and set the `test.environment` to `"node"`. We would also need to check anyting that is currently checkng for just `mode == "test"` and also catch the new `node-test`.

client/js/src/client.ts
```diff
@@ -36,6 +36,16 @@ export class NodeBlob extends Blob {
...
+if (typeof window === "undefined") {
+	import("eventsource")
+		.then((EventSourceModule) => {
+			global.EventSource = EventSourceModule.default as any;
+		})
+		.catch((error) =>
+			console.error("Failed to load EventSource module:", error)
+		);
+}
+
```
pngwn, line 48: Thinking about it, I don't think we want to do this. This is why we have the issues with "cannot set window.Websocket" previously. We should just do this in the `eventsource_factory` method and directly return the class.

client/js/src/client.ts
```diff
@@ -73,8 +63,17 @@ export class Client {
 	}
 
 	eventSource_factory(url: URL): EventSource | null {
-		if (typeof window !== undefined && typeof EventSource !== "undefined") {
-			return new EventSource(url.toString());
+		if (typeof window === "undefined" || typeof EventSource === "undefined") {
+			import("eventsource")
+				.then((EventSourceModule) => {
+					global.EventSource = EventSourceModule.default as any;
+					return new EventSource(url.toString());
```
pngwn, line 70: We can just remove the assignment to the global namespace here, and called the module + return directly.

pngwn:
Also closes #5840

pngwn (approved):
Looks great! Thanks for ironing out the kinks with this one @hannahblair! And sorry for the back and forth!

PR #8127, opened 2024-04-25
allow the canvas size to be set on the `ImageEditor`

pngwn:
## Description

Closes: #(issue)

## 🎯 PRs Should Target Issues

## Tests


abidlabs:
>  canvas_size but no crop size (+ sources). canvas_size should apply at load. Adding a new source may alter that. reset should use the canvas_size again.

Just to confirm, source here refers to the background image, right?

abidlabs:
> Just laying out some scenarios that I need to test:

Yes these are the only scenarios I can think of as well @pngwn

abidlabs:
Noticed that the docstrings for `height` and `width` are incorrect:

> height: The height of the displayed images, specified in pixels if a number is passed, or in CSS units if a string is passed.
> width: The width of the displayed images, specified in pixels if a number is passed, or in CSS units if a string is passed.

This is not the behavior, at least when interactive=True. These arguments do not affect the size of the images, but rather of the container of the component as a whole (which is consistent with other components).

e.g. `width=500`

image

e.g. `width=1000`

image

This means we shoudn't worry about these arguments conflicting with `canvas_size` (unless they are too small), but we should fix the docstrings

pngwn:
I'll fix the docstrings as well. That has actually been a source of confusion so it will be good to clean them up.

pngwn:
> Just to confirm, source here refers to the background image, right? So the idea here is that if you set a background image, its dimensions will set the canvas size?

Yes that is correct. The crop is what acts as an actual constraint, canvas_size is just a default / clear value really.

pngwn:
Tested all and works fine. I have also added the ability to set `sources=None` because `sources=[]` was annoying.

Some demo's below if anyone wants to test:

**Scenario 1**
`canvas_size` but `sources=None`

```python
import gradio as gr


def predict(im):
    return im["composite"]

def get_image_size(im):
    return f"{im['background'].shape[0]} x {im['background'].shape[1]}"


with gr.Blocks() as demo:
    with gr.Group():
        with gr.Row():
            im = gr.ImageEditor(
                canvas_size=(1000, 1000),
                sources=None,
            )
            im_preview = gr.Image()
        txt = gr.Textbox(label="image size")
        im.change(predict, outputs=im_preview, inputs=im, show_progress="hidden")
        im.change(get_image_size, inputs=im, outputs=txt)


if __name__ == "__main__":
    demo.launch()
```

**Scenario 2**

`canvas_size` with sources available (default)

```python
import gradio as gr


def predict(im):
    return im["composite"]

def get_image_size(im):
    return f"{im['background'].shape[0]} x {im['background'].shape[1]}"


with gr.Blocks() as demo:
    with gr.Group():
        with gr.Row():
            im = gr.ImageEditor(
                canvas_size=(1000, 1000),
            )
            im_preview = gr.Image()
        txt = gr.Textbox(label="image size")
        im.change(predict, outputs=im_preview, inputs=im, show_progress="hidden")
        im.change(get_image_size, inputs=im, outputs=txt)


if __name__ == "__main__":
    demo.launch()
```


**Scenario 3**

`canvas_size` and `crop_size`. 

```python
import gradio as gr


def predict(im):
    return im["composite"]

def get_image_size(im):
    return f"{im['background'].shape[0]} x {im['background'].shape[1]}"


with gr.Blocks() as demo:
    with gr.Group():
        with gr.Row():
            im = gr.ImageEditor(
                canvas_size=(1000, 1000),
                crop_size=(200, 200),
            )
            im_preview = gr.Image()
        txt = gr.Textbox(label="image size")
        im.change(predict, outputs=im_preview, inputs=im, show_progress="hidden")
        im.change(get_image_size, inputs=im, outputs=txt)


if __name__ == "__main__":
    demo.launch()
```

pngwn:
Also fixed the docstrings. Automated tests are challenging for the image editor, so I'll look into that elsewhere. Existing visual tests for the image editor are always causing flake, so I don't want to add to that right now.

pngwn:
@abidlabs @freddyaboulton should be fine now, silly thing.

gradio/components/image_editor.py
```diff
@@ -160,12 +160,13 @@ def __init__(
         brush: Brush | None | Literal[False] = None,
         format: str = "webp",
         layers: bool = True,
+        canvas_size: Tuple[int, int] | None = None,
```
abidlabs, line 163:
```suggestion
        canvas_size: tuple[int, int] | None = None,
```

abidlabs (approved):
I tried different setups and works beautifully! Just left one nit on the python type hint, thanks @pngwn

Issue #7330, opened 2024-02-06
Add analytics for custom components workflow


abidlabs:
It'd be great to have some analytics for how many people are running each step of the custom component workflow along with some basic metadata:
* `create`
* `dev`
* `build`
* `publish`

pngwn:
I'm not a huge fan of this personally but i'll let others chime in.

freddyaboulton:
I think this will help us keep track of growth in the custom components workflow so I'm for it. I don't think we need separate analytics for dev because that usage is already tracked in our current analytics since we're launching a regular gradio blocks - maybe we just need a flag in the analytics metadata.

PR #8130, opened 2024-04-25
Add Analytics for custom components

freddyaboulton:
## Description

Closes: #7330 

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/gradio/analytics.py
+++ b/gradio/analytics.py
@@ -178,17 +178,12 @@ def launched_analytics(blocks: gradio.Blocks, data: dict[str, Any]) -> None:
         [],
     )
 
-    from gradio.blocks import BlockContext
-
     for x in list(blocks.blocks.values()):
-        blocks_telemetry.append(x.get_block_name()) if isinstance(
-            x, BlockContext
-        ) else blocks_telemetry.append(str(x))
-
+        blocks_telemetry.append(x.get_block_name())
     for x in blocks.fns:
         targets_telemetry = targets_telemetry + [
             # Sometimes the target can be the Blocks object itself, so we need to check if its in blocks.blocks
-            str(blocks.blocks[y[0]])
+            blocks.blocks[y[0]].get_block_name()
             for y in x.targets
             if y[0] in blocks.blocks
         ]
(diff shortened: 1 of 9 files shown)


gradio/analytics.py
```diff
@@ -178,29 +178,34 @@ def launched_analytics(blocks: gradio.Blocks, data: dict[str, Any]) -> None:
...
         events_telemetry = events_telemetry + [
             y[1] for y in x.targets if y[0] in blocks.blocks
         ]
         inputs_telemetry = inputs_telemetry + [
-            str(blocks.blocks[y]) for y in x.inputs if y in blocks.blocks
+            blocks.blocks[y].get_block_name() for y in x.inputs if y in blocks.blocks
         ]
         outputs_telemetry = outputs_telemetry + [
-            str(blocks.blocks[y]) for y in x.outputs if y in blocks.blocks
+            blocks.blocks[y].get_block_name() for y in x.outputs if y in blocks.blocks
```
dawoodkhan82, line 197: does `get_block_name` return a string different from before? If so, when comparing analytics this could break things

gradio/analytics.py
```diff
@@ -178,29 +178,34 @@ def launched_analytics(blocks: gradio.Blocks, data: dict[str, Any]) -> None:
...
     )
 
-    from gradio.blocks import BlockContext
-
     for x in list(blocks.blocks.values()):
-        blocks_telemetry.append(x.get_block_name()) if isinstance(
-            x, BlockContext
-        ) else blocks_telemetry.append(str(x))
-
+        blocks_telemetry.append(x.get_block_name())
```
freddyaboulton, line 182: We were saving the string repr of the instance as opposed to the name, e.g. <gradio.components.textbox.Textbox object at 0x11c035790>

gradio/analytics.py
freddyaboulton: Before we were doing the wrong thing imo. We were storing the full class path `<gradio.components.textbox.Textbox object at 0x11c035790>`.

Change introduced in this PR: https://github.com/gradio-app/gradio/pull/7712 probably

gradio/analytics.py
```diff
@@ -227,6 +232,33 @@ def launched_analytics(blocks: gradio.Blocks, data: dict[str, Any]) -> None:
...
 
+def custom_component_analytics(
+    command: str,
+    template: str | None,
+    upload_pypi: bool | None,
+    upload_demo: bool | None,
+    upload_source: bool | None,
+    generate_docs: bool | None = None,
+    bump_version: bool | None = None,
+) -> None:
```
pngwn, line 243: Do you think it would be worth capturing the arguments being used (as a list) but maybe not the value. Then we could see if people are customising extensively or not?

pngwn (approved):
look great. Thanks @freddyaboulton! This will be super useful.

Just left a small question.

abidlabs:
Everything looks good to me @freddyaboulton! Just about to review the companion PR on the server

abidlabs (approved):

PR #8131, opened 2024-04-25
Gradio components in `gr.Chatbot()`

dawoodkhan82:
## Description

Use Image, Audio, Video, Gallery, and Plot components in the chatbot.

Closes: #(issue)

## 🎯 PRs Should Target Issues

## Tests


gradio/components/chatbot.py
```diff
@@ -184,18 +209,36 @@ def preprocess(
...
-            filepath = str(chat_message[0])
-
+        self, chat_message: str | tuple | list | GradioComponent | None
+    ) -> str | FileMessage | List[FileMessage] | PlotData | GalleryData | None:
+        def create_file_message(chat_message, filepath):
             mime_type = client_utils.get_mimetype(filepath)
             return FileMessage(
                 file=FileData(path=filepath, mime_type=mime_type),
-                alt_text=chat_message[1] if len(chat_message) > 1 else None,
+                alt_text=chat_message[1]
```
freddyaboulton, line 218: Imo we should not support passing a gradio component in a tuple. Makes it so that there are too many ways to do the same thing. I think just supporting the value outside the tuple is enough.

gradio/components/chatbot.py
```diff
@@ -139,24 +151,37 @@ def __init__(
...
             return None
         elif isinstance(chat_message, FileMessage):
             if chat_message.alt_text is not None:
                 return (chat_message.file.path, chat_message.alt_text)
             else:
                 return (chat_message.file.path,)
-        elif isinstance(chat_message, str):
+        elif isinstance(chat_message, (str)):
             return chat_message
+        elif isinstance(chat_message, GalleryData):
```
freddyaboulton, line 171: I think if a user returns `gr.Image` in the conversation, a `gr.Image` should be present in the history. That's not the case now, but it should be possible if we implement one of the api suggestions above.

gradio/components/chatbot.py
```diff
@@ -139,24 +151,37 @@ def __init__(
...
+            value = Gallery().preprocess(chat_message)
```
freddyaboulton, line 172: I would set `value='filepath`` explicitly here. Cause if we ever change the default it would increase the latency a lot by converting to PIL or whatever.

gradio/components/chatbot.py
```diff
@@ -184,18 +209,36 @@ def preprocess(
...
+        elif isinstance(chat_message, (tuple, list)):
+            if isinstance(chat_message[0], GradioComponent):
+                return type(chat_message[0]).postprocess(
+                    chat_message[0], chat_message[0]._constructor_args[1]["value"]
+                )
+            else:
+                filepath = str(chat_message[0])
+                return create_file_message(chat_message, filepath)
+        elif isinstance(chat_message, GradioComponent):
+            if type(chat_message) == Plot or type(chat_message) == Gallery:
```
freddyaboulton, line 235: Imo i think `isinstance(chat_message, (Plot, Gallery))` is better here

freddyaboulton (commented):
🔥 PR @dawoodkhan82 ! Made some suggestions to simplify the api in the backend

demo/chatbot_multimodal/run.py
```diff
@@ -16,18 +20,21 @@ def add_message(history, message):
...
+fig = random_plot()
+
+with gr.Blocks(fill_height=True) as demo:
     chatbot = gr.Chatbot(
-        [],
+        [["Image", gr.Image(value="files/avatar.png", render=False)],
+         ["Video", gr.Video(value="files/world.mp4", render=False)],
+         ["Audio", gr.Audio(value="files/cantina.wav", render=False)],
+         ["Plot", gr.Plot(value=fig, render=False)],
+         ["Gallery", gr.Gallery(value=["files/lion.jpg", "files/cheetah.jpg", "files/zebra.jpg"], render=False)]],
```
dawoodkhan82, line 34: @freddyaboulton Any suggestions on how to simplify this so `render=False` is not necessary? I tried passing it as a constructor arg during postprocessing, but the component would still render.

gradio/components/chatbot.py
```diff
@@ -21,7 +26,12 @@ class FileMessage(GradioModel):
 
 
 class ChatbotData(GradioRootModel):
-    root: List[Tuple[Union[str, FileMessage, None], Union[str, FileMessage, None]]]
+    root: List[
```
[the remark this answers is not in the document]
dawoodkhan82: Nice, I like the first approach better. We would be adding another Model, but it's more readable

andreped:
@dawoodkhan82 Would be neat if there was added support for `gradio.Dataframe`, `gradio.Code`, and `gradio.Button` as well, as our application needs it :] As a minimum, `gradio.Dataframe` would be quite neat.

Could be done in a separate PR I suppose. Just throwing it out there, as you are already adding support for five of the most commonly used ones.

demo/chatbot_multimodal/run.py
freddyaboulton: Can you call `unrender` for each component in the initial value in the chatbot init?

freddyaboulton: Wondering if it would work.

dawoodkhan82:
@andreped For sure, dataframe is already on the roadmap. Will look into adding additional components in a future PR.

aliabid94:
Sorry should've commented earlier, but this seems a lot of custom code for each component. We should have a general approach that works for all components, rather than having to support each component one by one. We should also only import components "dynamically" if needed, rather than making chatbot have every other type of component as a dependency.
Also think visually, Component outputs should not be in a bubble, but stand alone in their usual Block. 
Happy to sync on this to design a solution.

dawoodkhan82:
Ok worked on refactoring the PR to generalize for all components, but reached some areas where I still can't get rid of some custom code. cc @aliabid94 lmk if you want to sync on a solution to see if its possible without having to change component code itself.

gradio/components/chatbot.py
```diff
@@ -139,24 +181,52 @@ def __init__(
...
+                if chat_message.component == "gallery":
+                    chat_message.value = data_model.parse_obj(chat_message.value)
+                    value = [x.image.path for x in chat_message.value.root]
+                    return component(value=value)
+                chat_message.value = data_model.parse_obj(chat_message.value)
+                return component(value=chat_message.value)
+            else:
+                raise ValueError(
+                    f"Invalid component for Chatbot component: {chat_message.component}"
+                )
```
dawoodkhan82, line 222: @aliabid94 here we still need some custom code for some components

js/chatbot/shared/ChatBot.svelte
```diff
@@ -215,37 +252,51 @@
...
-										</Video>
-									{:else if message !== null && message.file?.mime_type?.includes("image")}
-										<Image
-											data-testid="chatbot-image"
-											src={message.file?.url}
-											alt={message.alt_text}
-										/>
-									{:else if message !== null && message.file?.url !== null}
+									{:else if message !== null && "component" in message}
+										{#if message.component == "gallery"}
```
dawoodkhan82, line 256: Similar thing here, accessing the value can be different per component.

js/chatbot/shared/ChatBot.svelte
```diff
@@ -215,37 +261,51 @@
...
+											<BaseGallery
```
dawoodkhan82, line 266: @aliabid94 having trouble dynamically loading and using the different svelte components instead of using this giant if/else block.

gradio/components/chatbot.py
```diff
@@ -139,24 +174,35 @@ def __init__(
...
                 return (chat_message.file.path,)
-        elif isinstance(chat_message, str):
+        elif isinstance(chat_message, (str)):
             return chat_message
+        elif isinstance(chat_message, ComponentMessage):
+            component = import_component_and_data(
+                chat_message.component.capitalize()
+            )
+            if component is not None:
+                return component(**chat_message.constructor_args[0]) # type: ignore
```
freddyaboulton, line 194: I think we need to call preprocess here as well. I was playing around with it and I think the following should work (granted postprocess was modified to what I suggested above)

```python
            if component is not None:
                instance = component()
                if issubclass(instance.data_model, GradioModel):  # type: ignore
                    payload = instance.data_model(**chat_message.value)  # type: ignore
                elif issubclass(instance.data_model, GradioRootModel):  # type: ignore
                    payload = instance.data_model(root=chat_message.value)  # type: ignore
                value = instance.preprocess(payload)
                args = {k:v for k,v in chat_message.constructor_args[0].items() if k != "value"}
                return component(value=value, **args) # type: ignore
```

freddyaboulton: One complication here is that Plot preprocess should be changed to return either a plotly, matplotlib plot. not sure if that is possible actually. 

We may have to simple keep old components in the ComponentData format.

gradio/components/chatbot.py
```diff
@@ -2,26 +2,59 @@
...
 from gradio import utils
+from gradio.components import (
+    Component as GradioComponent,
+)
 from gradio.components.base import Component
 from gradio.data_classes import FileData, GradioModel, GradioRootModel
 from gradio.events import Events
 
 
+def import_component_and_data(component_name: str) -> GradioComponent | None:
```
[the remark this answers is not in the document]
dawoodkhan82: Nice didn't know we had that util function.

gradio/components/chatbot.py
dawoodkhan82: Added preprocess.

gradio/components/chatbot.py
```diff
@@ -184,18 +230,56 @@ def preprocess(
...
             mime_type = client_utils.get_mimetype(filepath)
             return FileMessage(
                 file=FileData(path=filepath, mime_type=mime_type),
-                alt_text=chat_message[1] if len(chat_message) > 1 else None,
+                alt_text=chat_message[1]
+                if not isinstance(chat_message, GradioComponent)
+                and len(chat_message) > 1
+                else None,
             )
+
```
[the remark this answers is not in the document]
dawoodkhan82: This breaks for the `plot` component, so left it as is for now.

freddyaboulton:
Thanks for making the changes @dawoodkhan82 ! 

When I return an image from the chatbot, I get 

```
AttributeError: 'Image' object has no attribute 'GRADIO_CACHE'
```

Everything else seems ok though! Demo code:

```python
import gradio as gr
import os
import plotly.express as px
import random

# Chatbot demo with multimodal input (text, markdown, LaTeX, code blocks, image, audio, & video). Plus shows support for streaming text.

def random_plot():
    df = px.data.iris()
    fig = px.scatter(df, x="sepal_width", y="sepal_length", color="species",
                    size='petal_length', hover_data=['petal_width'])
    return fig

def print_like_dislike(x: gr.LikeData):


def random_matplotlib_plot():
    import numpy as np
    import pandas as pd
    import matplotlib.pyplot as plt

    countries = ["USA", "Canada", "Mexico", "UK"]
    months = ["January", "February", "March", "April", "May"]
    m = months.index("January")
    r = 3.2
    start_day = 30 * m
    final_day = 30 * (m + 1)
    x = np.arange(start_day, final_day + 1)
    pop_count = {"USA": 350, "Canada": 40, "Mexico": 300, "UK": 120}
    df = pd.DataFrame({"day": x})
    for country in countries:
        df[country] = x ** (r) * (pop_count[country] + 1)

    fig = plt.figure()
    plt.plot(df["day"], df[countries].to_numpy())
    plt.title("Outbreak in " + "January")
    plt.ylabel("Cases")
    plt.xlabel("Days since Day 0")
    plt.legend(countries)
    return fig


    for x in message["files"]:
        history.append(((x,), None))
        history.append((message["text"], None))

def bot(history, response_type):
    if response_type == "plot":
        history[-1][1] = gr.Plot(random_plot())
    elif response_type == "matplotlib_plot":
        history[-1][1] = gr.Plot(random_matplotlib_plot())
    elif response_type == "gallery":
        history[-1][1] = gr.Gallery([os.path.join("files", "avatar.png")] * 2)
    elif response_type == "image":
        history[-1][1] = gr.Image(os.path.join("files", "avatar.png"))
    elif response_type == "video":
        history[-1][1] = gr.Video(os.path.join("files", "world.mp4"))
    else:
        history[-1][1] = "Cool!"
    return history

fig = random_plot()

with gr.Blocks(fill_height=True) as demo:
    chatbot = gr.Chatbot(
        elem_id="chatbot",
        bubble_full_width=False,
        scale=1,
    )
    response_type = gr.Radio(["plot", "image", "text", "gallery", "video", "matplotlib_plot"], value="text", label="Response Type")

    chat_input = gr.MultimodalTextbox(interactive=True, placeholder="Enter message or upload file...", show_label=False)

    bot_msg = chat_msg.then(bot, [chatbot, response_type], chatbot, api_name="bot_response")

    chatbot.like(print_like_dislike, None, None)

demo.queue()
if __name__ == "__main__":
    demo.launch()
```

gradio/components/chatbot.py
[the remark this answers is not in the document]
freddyaboulton: Ok I can't repro that if I make this change with the demo code I linked below

chatbot_components_demo

Also I'm thinking that constructor_args does not need to be a list.

freddyaboulton:
Should be good to test with this demo!

[chatbot_multimodal.zip](https://github.com/user-attachments/files/15780118/chatbot_multimodal.zip)

pngwn:
this is the one, i can feel it

Issue #8157, opened 2024-04-28
gradio-cli accept SIGINT only once, leaving the subprocess running


Tiger3018:
### Describe the bug

When using gradio-cli, the SIGINT(2) or Ctrl+C passed to the gradio-cli will call `self.server.close()` and terminate the cli immediately.

However, there are circumstances when the `launch()` can't finish the jobs of thread join, such as #7051(`thread.join()`) and `http_server.watch_thread.join()`, or run into another `block_thread()` when `debug=True`.

When loaded with models, silent subprocess running will cause unnecessary RAM usage.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

def transcribe(stream, new_audio_chunk):
    return "", None

demo = gr.Interface(
    transcribe,
    ["state", gr.Audio(sources=["microphone"], streaming=True, )]
,
    ["state", "text"],
    live=True,
)

demo.launch(share=False, debug=True)

```

```sh
gradio app.py
```

### System Info

```shell
------------------------------
Operating System: Darwin
gradio version: 4.27.0
gradio_client version: 0.15.1
------------------------------------------------
gradio dependencies in your environment:
aiofiles: 23.2.1
altair: 5.3.0
fastapi: 0.110.2
ffmpy: 0.3.2
gradio-client==0.15.1 is not installed.
httpx: 0.27.0
huggingface-hub: 0.22.1
importlib-resources: 6.4.0
jinja2: 3.1.3
markupsafe: 2.1.3
matplotlib: 3.8.0
numpy: 1.26.4
orjson: 3.10.1
packaging: 23.2
pandas: 2.1.4
pillow: 10.2.0
pydantic: 2.5.3
pydub: 0.25.1
python-multipart: 0.0.9
pyyaml: 6.0.1
ruff: 0.4.0
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.12.3
typing-extensions: 4.9.0
urllib3: 2.0.7
uvicorn: 0.29.0
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.
gradio_client dependencies in your environment:
fsspec: 2023.10.0
httpx: 0.27.0
huggingface-hub: 0.22.1
packaging: 23.2
typing-extensions: 4.9.0
websockets: 11.0.3
```

### Severity

I can work around it

PR #8158, opened 2024-04-28
fix: handling SIGINT correctly in reload.py, single entrance of block_thread in blocks.py

Tiger3018:
## Description

* Handle SIGINT correctly in reload.py, to prevent gradio-cli exit when subprocess running 
* Make sure there is only a single entrance of block_thread in blocks.py

Closes: #8157

### DIFF
--- a/gradio/cli/commands/reload.py
+++ b/gradio/cli/commands/reload.py
@@ -130,7 +130,7 @@ def main(
         try:
             popen.wait()
         except (KeyboardInterrupt, OSError):
-            print("gradio-cli: Waiting gradio main thread cleaning.")
+            pass
 
 
 if __name__ == "__main__":
--- a/gradio/cli/commands/reload.py
+++ b/gradio/cli/commands/reload.py
@@ -126,7 +126,11 @@ def main(
             GRADIO_WATCH_DEMO_PATH=str(path),
         ),
     )
-    popen.wait()
+    if popen.poll() is None:
+        try:
+            popen.wait()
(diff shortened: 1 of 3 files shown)


gradio/cli/commands/reload.py
```diff
@@ -126,7 +126,11 @@ def main(
             GRADIO_WATCH_DEMO_PATH=str(path),
         ),
     )
-    popen.wait()
+    if popen.poll() is None:
+        try:
+            popen.wait()
+        except (KeyboardInterrupt, OSError):
+            print("gradio-cli: Waiting gradio main thread cleaning.")
```
freddyaboulton, line 133: Hi @Tiger3018 ! This contribution looks good to me but I don't think we need print statement here

Tiger3018: In the original implementation, the typer module will print `Aborted!` (See [this line](https://github.com/tiangolo/typer/blob/f6ba077dca84dfb567a8ee7e4580b0657d80878a/typer/core.py#L204) in typer module) and silently exit. This PR will remove the `Aborted!` echo and I don't know if an alternative text should be there.

Maybe we should just follow the implementation of gradio module, and use `pass` in this line instead?

freddyaboulton: Yes, `pass` is what I had in mind

freddyaboulton: Pushed that up so we can merge this in. Thanks!

freddyaboulton (approved):

Issue #8022, opened 2024-04-15
Gradio Unable to Load More than 5 iFrames After Package Update to 4.25 Version


marcoquin:
### Describe the bug

thanks for the tremendous work on gradio !

After updating the package from version 4.24 to version 4.25, Gradio encounters a problem where it is unable to load more than 5 iFrames. It is loading indefinitely. (the problem is still present in 4.26)

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

Here is the main app file:
```python
import gradio as gr
from fastapi import FastAPI, Request
from fastapi.responses import FileResponse
from fastapi.staticfiles import StaticFiles

from test import io_1, io_2, io_3, io_4, io_5, io_6
app = FastAPI()

@app.get("/")
async def get_home(request: Request):
    response = FileResponse(f"./folder_front/front.html"    )
    return response

app.mount("/folder_front", StaticFiles(directory="folder_front"), name="folder_front")

io_6_page = gr.mount_gradio_app(app=app, blocks=io_6, path="/6")
io_3_page = gr.mount_gradio_app(
    app=app, blocks=io_3, path="/3"
)
io_1_page = gr.mount_gradio_app(app=app, blocks=io_1, path="/1")
io_5_page = gr.mount_gradio_app(
    app=app, blocks=io_5, path="/5"
)
io_2_page = gr.mount_gradio_app(
    app=app, blocks=io_2, path="/2"
)

io_4_page = gr.mount_gradio_app(
    app=app, blocks=io_4, path="/4"
)
```
the test.py file for the import
```python
import gradio as gr

with gr.Blocks() as io_1:
    title = gr.Markdown("1")
with gr.Blocks() as io_2:
    title = gr.Markdown("2")
with gr.Blocks() as io_3:
    title = gr.Markdown("3")
with gr.Blocks() as io_4:
    title = gr.Markdown("4")
with gr.Blocks() as io_5:
    title = gr.Markdown("5")
with gr.Blocks() as io_6:
    title = gr.Markdown("6")
```
The html front:
```html
<!DOCTYPE html>
<head>
    <meta charset="UTF-8">
</head>
<body class = "light-body">

    <div class="iframe-container">
        <iframe id="1" src="/1/?__theme=light"  style="display:block;" ></iframe>
        <iframe id="2" src="/2/?__theme=light" style="display:block;"></iframe>
        <iframe id="3" src="/3/?__theme=light" style="display:block;"></iframe>
[...]

PR #8169, opened 2024-04-29
Only connect to heartbeat if needed

freddyaboulton:
## Description

Closes: #8022 

Rather than always connecting to the heartbeat route, we should do so only when gr.State variables are used (to clean up state) or when the `unload` event is set. Users may still run into the browser limit, however. 

EDIT: I originally wanted to raise an error or message when the page locked up due to the connection limit. But the browser will block all requests after the limit is reached so we won't be able to fetch the config and progressively load the app.

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/polite-lamps-burn.md
+++ b/.changeset/polite-lamps-burn.md
@@ -1,0 +1,6 @@
+---
+"@gradio/client": patch
+"gradio": patch
+---
+
+fix:Only connect to heartbeat if needed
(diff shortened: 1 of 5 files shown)


abidlabs:
Will this be compatible with gr.render() -- now that we can dynamically create components, a developer might include a workflow that dynamically creates `gr.State()` components? 

The only other idea I can think of is to use long polling instead of a persistent SSE connection for the heartbeat

pngwn (approved):
Looks great! Thanks @freddyaboulton!

Note: this should working fine with the render decorator because we refresh the config, as freddy mentioned.

aliabid94 (approved):
LGTM!

freddyaboulton:
Thanks for the reviews all!

Issue #8168, opened 2024-04-29
Browsers can cache the results of the `/custom_component` route, leading to confusion when new versions are published


freddyaboulton:
### Describe the bug

This happened to me while I was developing the leaderboard component last week. I don't have screenshots but what happened was the following:

* I had published v0.0.5 and it was on pypi and spaces
* Add a new feature
* Release v0.0.6 on pypi and install on spaces
* Space UI looks as if v0.0.5 is used even though space has 0.0.6 wheel installed. (I know because the feature I added in version 0.0.6 was not show in the UI). 
* Scratch head.
* I clear browser cache and everything looks as expected

We should set a `Cache-Control: no-cache`  header so that browsers don't use any cached requests. 

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

-

### System Info

```shell
gradio 4.27.0
```

### Severity

I can work around it

pngwn:
I don't think we should set no cache but we could probably set a max age of 5 mins or so?

pngwn:
We could also look into etags to allow more efficient revalidation.

freddyaboulton:
Yea makes sense!

PR #8170, opened 2024-04-29
Add ETag to `/custom_component` route to control browser caching

freddyaboulton:
## Description

Setting an ETag in the response of `/custom_component` that's equal to the hash of the requested file. I think this is preferable than setting a `max-age` because it should minimize the chance of a stale cache. During development, it's possible to build frequently to see if latest changes work outside the development server ( I do that atleast) so the `max-age` would be hard to set in this case.

Closes: #8168 

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/gradio/routes.py
+++ b/gradio/routes.py
@@ -12,6 +12,7 @@ if sys.version_info >= (3, 9):
     from importlib.resources import files
 else:
     from importlib_resources import files
+import hashlib
 import inspect
 import json
 import mimetypes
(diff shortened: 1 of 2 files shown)


gradio/routes.py
```diff
@@ -444,13 +448,28 @@ def custom_component_path(id: str, type: str, file_name: str):
...
-                )
+            path = safe_join(
+                str(Path(module_path).parent),
+                f"{component_instance.__class__.TEMPLATE_DIR}/{type}/{file_name}",
             )
 
+            key = f"{id}-{type}-{file_name}"
+
+            if key not in app.custom_component_hashes:
+                app.custom_component_hashes[key] = hashlib.md5(
```
freddyaboulton, line 459: We only need to compute the hash once as changing the frontend files is not supported after the server starts. This route is not used for dev mode.

pngwn (approved):
> I think this is preferable than setting a max-age because it should minimize the chance of a stale cache.

This makes sense, cache is always busted when devtools are open but that cant be guaranteed, just an old habit of mine.

Looks good. Thanks for taking care of this @freddyaboulton!

freddyaboulton:
Thanks for the review @pngwn !!

PR #8179, opened 2024-04-30
rework upload to be a class method + pass client into each component

### LINKED ISSUE #8150: APP breaking error with ImageEditor - latest version
### Describe the bug

I am using ImageEditor gradio demo from https://huggingface.co/spaces/yisol/IDM-VTON

gradio==4.28.3

First upload works

But if you don't change image , not modify reupload anything it gives below error

image

if i make a tiny change on the image and even revert it back it works again

**this error do not happen on my local computer but happens on gradio live share or runpod proxy**

Blocking usage of gradio

pngwn:
## Description

Closes #8150.

This PR reworks the `upload` function exported from `@gradio/client` to instead be an instance method of the client. This means we do not need to pass the `upload_fn` in and can instead use `this.upload_files` which uses `this.fetch_implementation`. This simplifies things significantly as the lite client and core client define their own fetch and stream methods.

To further reduce the complexity of the code base, i decide to pass the client into each component on the `gradio` namespace. This allows components to fetch/ stream as needed using the client methods (which are set appropriately for their environment) without us needing to pass individual function using svelte's context.

Wherever an `upload` function is required, the component can simply call it or pass it down.


client/js/src/client.ts
```diff
@@ -60,10 +61,11 @@ export class Client {
...
 	}
 
-	eventSource_factory(url: URL): EventSource | null {
+	eventSource_factory(url: URL): EventSource {
 		if (typeof window !== undefined && typeof EventSource !== "undefined") {
 			return new EventSource(url.toString());
 		}
+		// @ts-ignore
 		return null; // todo: polyfill eventsource for node envs
 	}
```
pngwn, line 70: This tying is temporary but was causing problems for some components. Ignored it until we polyfill.

client/js/src/client.ts
```diff
@@ -108,6 +116,7 @@ export class Client {
 		this.predict = predict.bind(this);
 		this.open_stream = open_stream.bind(this);
 		this.resolve_config = resolve_config.bind(this);
+		this.upload = upload.bind(this);
```
pngwn, line 119: This is the main change.

client/js/src/utils/upload_files.ts
```diff
@@ -1,4 +1,4 @@
-import type { Client } from "..";
+import { Client } from "..";
```
pngwn, line 1:
```suggestion
import type { Client } from "..";
```

js/app/src/lite/index.ts
```diff
@@ -127,13 +127,6 @@ export function create(options: Options): GradioAppController {
 
 	mount_prebuilt_css(document.head);
 
-	const overridden_fetch: typeof fetch = (input, init?) => {
-		return wasm_proxied_fetch(worker_proxy, input, init);
-	};
-	const EventSource_factory = (url: URL): EventSource => {
-		return wasm_proxied_EventSource_factory(worker_proxy, url);
-	};
-
```
pngwn, line 136: We don't need this any more, we just use the client everywhere.

js/app/src/Render.svelte
```diff
@@ -45,12 +47,6 @@
 
 	setContext("BLOCK_KEY", parent);
 
-	function handle_prop_change(e: { detail: Record<string, any> }): void {
-		// for (const k in e.detail) {
-		// 	instance_map[id].props[k] = e.detail[k];
-		// }
-	}
-
 	$: {
```
pngwn, line 50: This was an unused leftover.

js/app/src/Blocks.svelte
```diff
@@ -522,8 +521,6 @@
 	function isCustomEvent(event: Event): event is CustomEvent {
 		return "detail" in event;
 	}
-
-	setContext("upload_files", app.upload_files);
```
pngwn, line 526: ANd we don't need this for the same reason.

pngwn:
storybook should be fine now.

js/gallery/shared/Gallery.svelte
```diff
@@ -166,7 +167,7 @@
 	// and their remote URLs are directly passed to the client as `value[].image.url`.
 	// The `download` attribute of the <a> tag doesn't work for remote URLs (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#download),
 	// so we need to download the image via JS as below.
-	const fetch_implementation = getContext<typeof fetch>("fetch_implementation");
+	// const fetch_implementation = getContext<typeof fetch>("fetch_implementation");
```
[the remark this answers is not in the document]
pngwn: Yep. Missed this one.

js/audio/interactive/InteractiveAudio.svelte
```diff
@@ -98,9 +97,9 @@
 		let _audio_blob = new File(blobs, "audio.wav");
 		const val = await prepare_files([_audio_blob], event === "stream");
 		value = (
-			(
-				await upload(val, root, undefined, max_file_size ?? Infinity, upload_fn)
-			)?.filter(Boolean) as FileData[]
+			(await upload(val, root, undefined, max_file_size ?? Infinity))?.filter(
```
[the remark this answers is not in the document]
pngwn: I swapped it to ` max_file_size || undefined` because the 'falsey' value this cant take is actually `null` and the function doesn't accept `null`

PR #8180, opened 2024-04-30
Refactor analytics to not use api.gradio.app

freddyaboulton:
## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/gradio/analytics.py
+++ b/gradio/analytics.py
@@ -8,9 +8,10 @@ import os
 import threading
 import warnings
 from typing import Any
+from urllib.parse import quote
 
-import anyio
 import httpx
+from huggingface_hub.utils import build_hf_headers
 from huggingface_hub.utils._telemetry import _send_telemetry_in_thread
 from packaging.version import Version
 
@@ -101,14 +102,19 @@ def _do_normal_analytics_request(topic: str, data: dict[str, Any]) -> None:
 async def _do_wasm_analytics_request(topic: str, data: dict[str, Any]) -> None:
     data["ip_address"] = await get_local_ip_address_wasm()
     try:
-        await anyio.to_thread.run_sync(
-            _send_telemetry_in_thread,
-            topic,
-            "gradio",
-            data.get("version"),
-            data,
+        path = "/".join(quote(part) for part in topic.split("/") if len(part) > 0)
+        url = f"https://huggingface.co/api/telemetry/{path}"
+        headers = build_hf_headers(
+            token=False,  # no need to send a token for telemetry
+            library_name="gradio",
+            library_version=data.get("version"),
+            user_agent=data,
         )
-    except (Exception, BaseException):
+        await asyncio.wait_for(
+            pyodide_pyfetch(url, method="HEAD", headers=headers),
+            timeout=5,
+        )
+    except (Exception, BaseException, asyncio.TimeoutError):
         pass
 
 
--- a/gradio/analytics.py
+++ b/gradio/analytics.py
@@ -6,11 +6,12 @@ import asyncio
 import json
 import os
 import threading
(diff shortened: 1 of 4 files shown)


abidlabs:
Instead of using `_send_telemetry_in_thread`, why not use `send_telemetry`? That way we can avoid spinning up our own thread for each analytics request and instead use the single-thread approach that is baked into `huggingface_hub` (since Lite doesn't support threads, I guess we'd still need to use `_send_telemetry_in_thread` in that case)

cc @Wauplin

freddyaboulton:
@abidlabs _send_telemetry_in_thread does not actually launch a new thread as far as I can tell, it just sends the request to HF hub. In contrast, `send_telemetry` will place data in a queue and then start a new thread to pull data off that queue and send to the hub. I think we still need to run the analytics in a separate thread so that the "get ip address" operation does not block, so I think `_send_telemetry_in_thread` is better to not have an additional background thread running.

abidlabs:
@freddyaboulton yes you're right, the `_send_telemetry_in_thread `, but create a thread ourselves, e.g. when we do:

```py
threading.Thread(
            target=_do_normal_analytics_request,
```

it seems like it'd be better to use the documented public method if possible?

gradio/analytics.py
```diff
@@ -29,60 +31,91 @@ async def pyodide_pyfetch(*_args, **_kwargs):
...
+        url = f"https://huggingface.co/api/telemetry/{path}"
+        headers = build_hf_headers(
+            token=False,  # no need to send a token for telemetry
+            library_name="gradio",
+            library_version=data.get("version"),
+            user_agent=data,
+        )
         await asyncio.wait_for(
-            pyodide_pyfetch(url, method="POST", headers=headers, body=body),
+            pyodide_pyfetch(url, method="HEAD", headers=headers),
```
freddyaboulton, line 114: I originally wrote this to use `_send_telemetry_in_thread` but the analytics requests were not showing up in the webworker network tab.

If I use `pyodide_pyfetch`, they show up in the network tab. Still not sure if this works 100% though.

So I think we need this lower-level approach for wasm and can't rely on higher level functions in hub python library. 


image

abidlabs:
Nvm @freddyaboulton you are right we will need to spin our own thread to get the IP so let's stick with `_send_telemetry_in_thread`. @Wauplin is that a safe choice or is there a better option you'd suggest?

gradio/analytics.py
whitphx: If `_send_telemetry_in_thread` uses threading as its name implies, it [doesn't work](https://pyodide.org/en/stable/usage/wasm-constraints.html#included-but-not-working-modules) on Lite/Pyodide.

I think using `pyodide_pyfetch` is a right approach on Lite and it should work.

freddyaboulton: I reverted the changes for wasm since we can't set the user-agent on chrome!

whitphx: Ref: https://stackoverflow.com/a/42815264/13103190

Wauplin:
> we will need to spin our own thread to get the IP so let's stick with `_send_telemetry_in_thread`. @Wauplin is that a safe choice or is there a better option you'd suggest?

It's a reasonably safe choice yes. Don't see a better option that does not involve re-implementing the request.
Could you add a comment in `huggingface_hub`'s `_send_telemetry_in_thread` to reference to gradio-lite so that we don't forget? Little chance we change this internal method anytime soon but better safe that sorry :)

abidlabs (approved):
Code changes LGTM @freddyaboulton! Jw if you checked to confirm that the analytics went through to kibana?

freddyaboulton:
Yes analytics are going through!

PR #8182, opened 2024-05-01
Convert sse calls in client from async to sync

abidlabs:
This converts all of the HTTP / SSE calls in the python client from async to sync. There was no benefit in keeping them async as we would convert them to synchronous calls using `utils.synchronize_async` anyways and the use of this method was preventing the client from being used in environments that have their own async event management like gunicorn/gevent.

Closes: https://github.com/gradio-app/gradio/issues/7581

To test, create a file called `test.py` with the following code:

```py
from gevent import monkey
monkey.patch_all()
from flask import Flask

app = Flask(__name__)

client = Client("abidlabs/en2fr")  # a Space that translates from English to French

@app.route("/gen")
def gen():
      result = client.predict(
                  "hello",	
                  api_name="/predict"
              )
      return result

if __name__ == "__main__":
      app.run(host="0.0.0.0", port=5000)

```

Then run:

`gunicorn -k gevent -w 1 test:app --bind 0.0.0.0:5000`

Then visit `http://localhost:5000/gen` and you should see a valid greeting. 

Note: I didn't change the ws calls since websocket.send/receive are async by nature and so we'd have to do a workaround like `utils.synchronize_async` anyways to support them. So if you use the Client to connect to Gradio apps running 3.x  ("compatibility mode"), it will still use the async client.


client/python/gradio_client/utils.py
```diff
@@ -243,6 +244,7 @@ class Communicator:
     reset_url: str
     should_cancel: bool = False
     event_id: str | None = None
+    prediction_complete: bool = False
```
abidlabs, line 247: This field is added so that we can detect when a prediction is complete in the `stream_sse_v1plus()` thread so that we can check this field from the `check_for_cancel()` thread and have it terminate itself (since threads can't just be canceled like asyncio tasks)

abidlabs: Update: this field has been renamed to `thread_complete`

client/python/test/test_client.py
```diff
@@ -1240,7 +1240,7 @@ def test_download_private_file(self, gradio_temp_dir):
             src="gradio/zip_files",
         )
         url_path = "https://gradio-tests-not-actually-private-spacev4-sse.hf.space/file=lion.jpg"
-        file = client.endpoints[0]._upload_file(url_path)  # type: ignore
+        file = client.endpoints[0]._upload_file(url_path, 0)  # type: ignore
```
abidlabs, line 1243: This test is not related to this but needed to be fixed since we changed the signature of `_upload_file`. It wasn't caught before because this test is marked flaky

freddyaboulton (approved):
This looks good to me @abidlabs! Tested some local demos and everything looks good. One comment I have is whether we can stop raising `CancelledError` and just return now? That should stop the job in the threadpool no?

client/python/gradio_client/utils.py
```diff
@@ -311,7 +313,7 @@ async def get_pred_from_ws(
                             # otherwise will get nasty warning in console
                             task.cancel()
                             await asyncio.gather(task, reset, return_exceptions=True)
-                        raise CancelledError()
+                        raise concurrent.futures.CancelledError()
```
freddyaboulton, line 316: Why this change?

client/python/gradio_client/utils.py
```diff
@@ -400,58 +398,55 @@ async def get_pred_from_sse_v1plus(
...
-    )
-
-    for task in pending:
-        task.cancel()
-        try:
-            await task
-        except asyncio.CancelledError:
-            pass
+    helper.thread_complete = False
+    with concurrent.futures.ThreadPoolExecutor() as executor:
```
freddyaboulton, line 402: Can we use the same one used in the parent `Client` to not have to create another one?

aliabid94 (approved):
tested, LGTM!

abidlabs:
> This looks good to me @abidlabs! Tested some local demos and everything looks good. One comment I have is whether we can stop raising CancelledError and just return now? That should stop the job in the threadpool no?

Yup I think you're right

client/python/gradio_client/utils.py
abidlabs: This is just because I've replaced

```
from concurrent.futures import CancelledError
```

with 

```
import concurrent.futures
```

abidlabs:
> One comment I have is whether we can stop raising CancelledError and just return now? That should stop the job in the threadpool no?

Ok so I spent some time on this, but if we stop raising CancelledError and instead return `None`, then several upstream methods fail. The basic problem is we need to decide what `job.result()` should produce if the `job` is canceled. I think its not a bad idea for it to just raise a `CancelledError` as is currently the behavior? We could potentially change this and I'm open to other ideas, but I think we can make that part of another PR, perhaps as we address: https://github.com/gradio-app/gradio/issues/8188

abidlabs:
I'll go ahead and merge this in but open to suggestions on this @freddyaboulton

abidlabs:
Thanks for the reviews @aliabid94 and @freddyaboulton!

kxbin:
great work :+1:

Issue #7831, opened 2024-03-26
Add support for key-word arguments in JS client


abidlabs:
This is already supported in the Python Client, e.g.

```py
from gradio_client import Client, file

client = Client("abidlabs/whisper")

client.predict(
    audio=file("audio_sample.wav")
)
```
```
>> "This is a test of the whisper speech recognition model."
```

Beyond consistency, it would allow the JS client to take advantage of default arguments

Issue #5462, opened 2023-09-08
better typing for `@gradio/client`


pngwn:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
> gradioApp.predict() has return type Promise<unknown>. It could be something better than unknown (maybe use generics like gradioApp.predict<T>() will return Promise<GradioRes<T>>

Will add more details soon

**Describe the solution you'd like**  

**Additional context**

Ifeanyi55:
I agree. The importance of simplicity and clarity cannot be overemphasized.

Ifeanyi55:
I would like to take this issue. In what part of the doc exactly did you notice this? Is it the Python client or Javascript client? @pngwn

freddyaboulton:
This is specifically for the javascript client

Ifeanyi55:
Nice, thanks @freddyaboulton

DarhkVoyd:
Hey, I would like to take this issue if it's still valid.

pngwn:
@DarhkVoyd definitely still valid! I'll add some additional details in a reply shortly!

pngwn:
So I think the general idea was that it would be good to have response types typed, in order to do this we could need the client or the predict/ submit types to be generic.

Originally I was thinking that something like this would be cool:

```ts
import { client } from '@gradio/client';

const app = client(...)

type Input = ['textbox', 'number', 'radio'];
type Output = ['image', 'textbox'];

const resp = await app.predict<[Input, Output]>(...)
```

> [!NOTE]
> The following types are illustrative, I haven't checked to see if they are correct.

the input would be typed as:

```ts
type Input = [string, number, string]

// type error
app.predict([1, 2, 3])

// correct
app.predict(["hi", 23, "cheese"])
```

response data would be typed as:
```ts
interface ImageData {
  path: string;
  name: string;
  url: string;
}

type TextboxData  = string;

type ResponseData = [ImageData, TextboxData];

// type error
response[2]
response[1].key
response[0].nsme
response[0].name.match(..) // regex method

// correct
response[0].name
response[0].name.toLowerCase()
response[1].substr(...) // string method
```

So essentially the input literals would become keys for specific types (that we would define somewhere) and we would map over those types to dynamically create the types for the inputs + returns values.

Additionally or optionally, I think it would be nice to be able to type the client directly, rather than needing to type every predict or submit call.

```ts
import { client } from '@gradio/client';
 
type PredictInput = ...
[...]

PR #8187, opened 2024-05-01
Add support for keyword args in JS client

hannahblair:
## Description

This PR adds support for passing keyword arguments as well as positional args in the JS client. This is backwards compatible, so the two usages will work: 

```
const result = await app.predict({
    endpoint: "/predict",
    data: ["Chewbacca"]
});

```
```
const result = await app.predict("/predict", ["Chewbacca"]);
```

Closes: #7831

## 🎯 PRs Should Target Issues

## Tests


pngwn:
I don't think was the intention with that issue. I think it is keyword arguments specifically for the data. The endpoint isn't optional in the JS client so this would be something like:

```ts
const result = await app.predict("/predict", {
  audio: audio_file,
  textbox: "Chewbacca"
});
```

What i don't understand is what those key correspond to. @abidlabs could you clarify what the keys respond and how we figure out what they are?

abidlabs:
Thanks @hannahblair @pngwn yes the idea of #7831 is to allow users to provide the data as a dictionary where keys are parameter names and the values are the data. The benefit of this is (a) it would make it clearer which argument is being passed to which parameter, which can help prevent errors, especially with many parameters or similar-looking parameters (b) allow us to utilize default values of parameters i.e. make some keys optional if a developer provides a default value in the function

> What i don't understand is what those key correspond to. @abidlabs could you clarify what the keys respond and how we figure out what they are?

The keys can be obtained using the "/info" route. If you look at the "/info" route, each endpoint has a "parameters" key whose dictionary now looks like this:

```js
"parameters": [
  {
    "label": "x",
    "parameter_name": "x",
    "parameter_has_default": false,
    "parameter_default": null,
    "type": {
    "type": "string"
  },
}
...
],
```

The key should be value of `parameter_name`, whether or not it has a default is indicated by `parameter_has_default` (so if this is true, the key should be considered optional), and the value of that default is indicated by `parameter_default`

pngwn:
The only issue with is that we won't have any types, although we might be able to make it generic and provide a snippet. Maybe as part of #5462.

PR #8189, opened 2024-05-01
Use workspace version for code in _website

aliabd:
Uses workspace for @gradio/code in the _website package.json instead of a pinned version. Removes the circular import from js/preview/test/test/frontend that was causing the issue. Also changes the version of postcss to use pnpm 9.0.0


pngwn (approved):
Ah nice catch. Thanks for figuring this out @aliabd!

PR #8197, opened 2024-05-02
Add support for passing keyword args to `data` in JS client

hannahblair:
## Description

This PR adds support for passing keyword arguments as well as positional args to the `data` param in the JS client. This is backwards compatible, so the two usages will work:

```
const result = await app.predict("/predict", ["Chewbacca"]);
```
```
const result = await app.predict("/predict", { name: "Chewbacca" });
```
Also adds unit tests. 

The `CodeSnippet.svelte` API usage has been updated, which you can see here: https://huggingface.co/spaces/hmb/test_args

Closes: #7831

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/js/app/src/api_docs/ApiDocs.svelte
+++ b/js/app/src/api_docs/ApiDocs.svelte
@@ -192,9 +192,6 @@
 								endpoint_parameters={info.named_endpoints[
 									"/" + dependency.api_name
 								].parameters}
-								js_parameters={js_info.named_endpoints[
-									"/" + dependency.api_name
-								].parameters}
 								{dependency}
 								{dependency_index}
 								{current_language}


client/js/src/helpers/api_info.ts
```diff
@@ -298,3 +298,72 @@ export function handle_message(
...
+	const resolved_data: unknown[] = [];
+	const provided_keys = Object.keys(data);
+
+	parameters.forEach((param, index) => {
+		if (data.hasOwnProperty(param.parameter_name)) {
+			resolved_data[index] = data[param.parameter_name];
+		} else if (param.parameter_has_default) {
+			resolved_data[index] = param.parameter_default;
+		} else {
+			resolved_data[index] = undefined;
```
pngwn, line 348: Should we throw an error in this case? 

@abidlabs @freddyaboulton what does the python client do if no param is given but there is no default?

pngwn (approved):
Just left one small comment but everything else looks great + works as expected!

Thanks @hannahblair!

client/js/src/helpers/api_info.ts
abidlabs: Yes it raises an error saying that a required param is missing

abidlabs:
Nice @hannahblair! Just 2 nits with regards to the code snippets:

1)

image


* The code snippet itself looks good, but in the parameter table below, we should now put the name of the parameter, instead of the `[0]` to designate the positional argument
* We can also remove the comment `// string in...` as that information is included in the parameter table

Matching the Python code snippet and parameter table:

image

2) The RecordingSnippet.svelte code snippet still shows the positional format, would be great if we can update this as well:

image

pngwn (approved):
Tested and looks good, will file another issue for the API recorder because it is broken in the release.

Issue #8124, opened 2024-04-25
[Custom Components] support custom components in gr.load


Col0ring:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
I have a space with custom components, but `gr.load` is not working properly.

**Describe the solution you'd like**  
Support custom components in `gr.load`.

**Additional context**  
```
File "/Users/coloring/anaconda3/envs/gradio_4/lib/python3.12/site-packages/gradio/utils.py", line 525, in component_or_layout_class
    raise ValueError(f"No such component or layout: {cls_name}")
ValueError: No such component or layout: i18neditor
```
This error is caused by `gradio.utils.component_or_layout_class`, maybe `gradio` should expose some context or parameters of this method.

abidlabs:
Makes sense — could you link to the Space you are trying to load or provide us a quick repro?

Col0ring:
> Makes sense — could you link to the Space you are trying to load or provide us a quick repro?
Reproduce repo: https://huggingface.co/spaces/Coloring/gradio_load_test

PR #8200, opened 2024-05-02
Support custom components in gr.load

freddyaboulton:
## Description

Closes: #8124

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/fuzzy-mirrors-scream.md
+++ b/.changeset/fuzzy-mirrors-scream.md
@@ -1,0 +1,6 @@
+---
+"gradio": patch
+"gradio_client": patch
+---
+
+fix:Support custom components in gr.load
--- a/.changeset/fuzzy-mirrors-scream.md
+++ b/.changeset/fuzzy-mirrors-scream.md
@@ -1,9 +1,9 @@
 ---
-"gradio": patch
+"gradio": feat
 "gradio_client": patch
 ---
 
-### Support custom components in gr.load
+#### Support custom components in gr.load
 
 It is now possible to load a demo with a custom component with `gr.load`.
 
--- a/.changeset/fuzzy-mirrors-scream.md
+++ b/.changeset/fuzzy-mirrors-scream.md
@@ -1,5 +1,5 @@
 ---
-"gradio": feat
+"gradio": patch
 "gradio_client": patch
 ---
 
--- a/.changeset/fuzzy-mirrors-scream.md
+++ b/.changeset/fuzzy-mirrors-scream.md
@@ -3,4 +3,21 @@
 "gradio_client": patch
 ---
 
-fix:Support custom components in gr.load
+### Support custom components in gr.load
+
+It is now possible to load a demo with a custom component with `gr.load`.
+
+The custom component must be installed in your system and imported in your python session.
+
+```python
+import gradio as gr
+import gradio_pdf
+
+demo = gr.load("freddyaboulton/gradiopdf", src="spaces")
+
+if __name__ == "__main__":
+    demo.launch()
+```
+
+<img width="1284" alt="image" src="https://github.com/gradio-app/gradio/assets/41651716/9c3e846b-f3f2-4c1c-8cb6-53a6d186aaa0">
+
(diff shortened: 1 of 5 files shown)


client/python/gradio_client/client.py
```diff
@@ -1171,7 +1171,11 @@ def process_input_files(self, *data) -> tuple:
                     or utils.is_http_url_like(f),
                 )
             elif not self.client.upload_files:
-                d = utils.traverse(d, self._upload_file, utils.is_file_obj_with_meta)
+                d = utils.traverse(
```
freddyaboulton, line 1174: This was an pre-existing bug where _upload_file was not being passed the index parameter

gradio/utils.py
```diff
@@ -518,6 +518,32 @@ def resolve_singleton(_list: list[Any] | Any) -> Any:
         return _list
 
 
+def get_all_components() -> list[type[Component] | type[BlockContext]]:
```
freddyaboulton, line 521: We're using this in #8180 so I will refactor when this is in

gradio/utils.py
```diff
@@ -528,33 +554,20 @@ def component_or_layout_class(cls_name: str) -> type[Component] | type[BlockCont
...
+
+    components = get_all_components()
+    for cls in components:
+        if cls.__name__.lower() == cls_name.replace("_", "") and (
+            issubclass(cls, Component) or issubclass(cls, BlockContext)
         ):
             return cls
-    raise ValueError(f"No such component or layout: {cls_name}")
+    raise ValueError(
+        f"No such component or layout: {cls_name}. "
```
freddyaboulton, line 567: I tried to automatically import the package containing the custom component but it's quite tricky. We just have a class name so in order to get the package that it's in we have to import every package installed in the system which is not ideal and cause cause things to break.

abidlabs:
Nice @freddyaboulton! I agree this is the right approach:

> The custom component must be installed in your system and imported in your python session.

I'll review the rest of the PR soon

.changeset/fuzzy-mirrors-scream.md
```diff
@@ -0,0 +1,23 @@
+---
+"gradio": feat
```
abidlabs, line 2: is this supposed to be minor?
```suggestion
"gradio": minor
```

abidlabs: or actually it should be a patch?

gradio/utils.py
```diff
@@ -518,6 +518,32 @@ def resolve_singleton(_list: list[Any] | Any) -> Any:
...
+        gr.components.Component.__subclasses__()
+        + gr.blocks.BlockContext.__subclasses__()  # type: ignore
+    )
+    subclasses = []
+
+    while classes_to_check:
+        subclass = classes_to_check.pop()
+        children = subclass.__subclasses__()
+        if children:
+            classes_to_check.extend(children)
```
abidlabs, line 534: nit: slightly more concise (since if the subclass doesn't have any subclasses, children will be an empty list and extending by an empty list leaves `classes_to_check` unchanged) 
```suggestion
        classes_to_check.extend(subclass.__subclasses__())
```

test/test_external.py
```diff
@@ -514,3 +514,11 @@ def test_use_api_name_in_call_method():
     # app = gr.load(name="spaces/gradio/multiple-api-name-test")
     # assert app(15, api_name="minus_one") == 14
     # assert app(4, api_name="double") == 8
+
+
+def test_load_custom_component():
+    demo = gr.load("spaces/freddyaboulton/gradiopdf")
```
abidlabs, line 520: This test is currently passing on CI because we import gradio_pdf.PDF in a separate test. But it fails if you run it standalone. Would be good to import `gradio_pdf` here so it passes on its own
```suggestion
    from gradio_pdf import PDF
    demo = gr.load("spaces/freddyaboulton/gradiopdf")
```

abidlabs (approved):
Nice implementation @freddyaboulton! Left a couple of small nits but otherwise good to merge

test/test_external.py
freddyaboulton: Good call

Issue #4922, opened 2023-07-14
Documentation page has an error with CheckBox


JElbourne:
### Describe the bug

Documentation page has an error with CheckBox

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

Visit the webpage https://www.gradio.app/docs/chatbot
, Then go to checkbox page or another https://www.gradio.app/docs/checkbox

### Screenshot

![image](https://github.com/gradio-app/gradio/assets/556142/ee636a42-bf8d-4fe6-9bf0-9915778b0379)

### System Info

```shell
Webpage on your site
```

### Severity

I can work around it

freddyaboulton:
There is an error in the console @aliabd 

```js
Uncaught (in promise) DOMException: Failed to execute 'insertBefore' on 'Node': The node before which the new node is to be inserted is not a child of this node.
    at yt (https://www.gradio.app/_app/immutable/chunks/index.a326135e.js:1:2748)
    at Object.m (https://www.gradio.app/_app/immutable/nodes/6.a488eca3.js:1:4854)
    at Object.p (https://www.gradio.app/_app/immutable/nodes/6.a488eca3.js:1:3940)
    at Object.p (https://www.gradio.app/_app/immutable/nodes/6.a488eca3.js:17:3776)
    at $t (https://www.gradio.app/_app/immutable/chunks/index.a326135e.js:1:6906)
    at rt (https://www.gradio.app/_app/immutable/chunks/index.a326135e.js:1:6559)
```

johnson7788:
same issue, gradio-3.45.0, when use chatbox 
Uncaught (in promise) DOMException: Failed to execute 'insertBefore' on 'Node': The node before which the new node is to be inserted is not a child of this node.
    at T (https://gradio.s3-us-west-2.amazonaws.com/3.45.0/assets/index-a9603da9.js:2:12017)
    at Gc.i (https://gradio.s3-us-west-2.amazonaws.com/3.45.0/assets/index-a9603da9.js:2:14895)
    at Gc.p (https://gradio.s3-us-west-2.amazonaws.com/3.45.0/assets/index-a9603da9.js:2:14946)
    at Object.p (https://gradio.s3-us-west-2.amazonaws.com/3.45.0/assets/StaticMarkdown-ae9af4d2.js:339:365)
[...]

PR #8220, opened 2024-05-06
Convert all demos on docs to lite

aliabd:
- Converts the demos on both main and pip docs to lite. 
- Removes the demos for 3.x docs (keeps the code snippets)
- Removes the embedded component for now, will bring it back with lite in a separate PR. 

I will still need to differentiate the main and pip demos, which currently both use the released version of lite. I'll upload the main wheel of lite to aws and add it to our github actions workflow. 

But this currently finally makes the super annoying docs issue go away where embedded spaces were interfering with the page and causing the js to break/users to see previous docs content on a new page. Please test this by navigating really fast between pages. 

Closes: #4922 #5825


abidlabs:
awesome! closes https://github.com/gradio-app/gradio/issues/4922 and maybe closes https://github.com/gradio-app/gradio/issues/5825 as well?

abidlabs:
Very cool @aliabd! Went through a bunch of demos and they all look good to me!

js/_website/src/lib/components/Demos.svelte
```diff
@@ -26,33 +41,11 @@
 				Open in 🎢 ↗
 			</button>
 
-			<gradio-lite playground shared-worker class="p-2">
+			<gradio-lite playground shared-worker layout="vertical" class="p-2">
```
freddyaboulton, line 44: Annoying suggestion, but can the demo go before the code? Users now have to scroll down a lot to see the demo.

playground_lite_docs

freddyaboulton (approved):
Nice PR @aliabd ! Are we also replacing the demos in the guides with lite?

js/_website/src/lib/components/Demos.svelte
aliabd: Good point that's kind of annoying. I'll fix it on the lite playground side and not directly on the website. Opened an issue https://github.com/gradio-app/gradio/issues/8228

aliabd:
@freddyaboulton yes, will do that in a separate PR because it's slightly different. Most of it is written content anyway (ie will just replace `<gradio-app space="...">` with `<gradio-lite>{code}</gradio-lite>`)

Issue #7739, opened 2024-03-19
Individual Button Creation each time a gr.DataFrame has a new row entry


athapapa:
I have got a gr.DataFrame which gets updated with new row entries manually by the user. Ideally, I would like to have another column in the same DataFrame where individual Clear Buttons are created every time a new row is created. Unfortunately, unless I am mistaken, it is not possible to place a gr.Button in a gr.DataFrame cell. Therefore, I would like to iteratively create individual buttons outside the DataFrame each time there is a new entry which will be responsible for the deletion of each row. 

Code below: 

![Screenshot 2024-03-19 112936](https://github.com/gradio-app/gradio/assets/72748167/0cd20276-b5c5-4e87-be93-9c6f9368c18d)

As shown in the snippet, I would basically like a new_delete_btn to be created every time there is a new url_table entry. I can do a workaround by initialising many buttons and playing with their visibility, but that would require to have a maximum predetermined number of rows and the UI result is not great.

Issue #4689, opened 2023-06-26
Declarative and Dynamic APIs for Gradio 4.0


aliabid94:
This issue is to discuss potential API designs of the Gradio 4.0 API, with the goals of:

- Making it easier to build complex interfaces
- Being fully backwards compatible
- Supporting declarative logic (where the UI is a function of state)
- Supporting dynamic logic (where the UI logic can include if/else and loops)

Suggested API below will introduce:
- the `gr.render` decorator, which will be attached to a function that takes State variables as inputs. 
- Connecting function arguments and returns to components via type hints, to make it easier to associate components with function variables.
- The `gr.on` decorator that makes it easier to associate event functions and their triggers

## Example 1: Add Numbers

Let's take a look at a simple demo that adds two numbers:

```python
import gradio as gr

with gr.Blocks() as demo:
    a = gr.Number()
    b = gr.Number()
    btn = gr.Button("Add")
    result = gr.State()

    @gr.render
    def show_result(result: result):
        if result is None:
            gr.Markdown("Click the button to add two numbers.")
        else:
            gr.Number(value=result)

    @gr.on(btn.click)
    def add(a: a, b: b) -> result:
        return a + b
```

The function `show_result` will automatically be re-run on any changes to its inputs, which in this case is the `result` state variable. This state variable is bound to a function argument of the same name. 

When a user clicks the add button, the function `add` is run, using gr.Numbers `a` and `b` as inputs, as implied by the type hint. The values of these components are bound to arguments of the same name. The returned value is set to the `result` State variable, as implied by the function return type hint. The change in `result` triggers a re-render of `show_result`.

## Example 2: Todo List

Let's take a look at a Todo list, which involves dynamically generating an arbitrary number of components.

```python
import gradio as gr

with gr.Blocks() as demo:
    todos_var = gr.State([])

    @gr.render
    def todo_list(todos: todos_var):
        for i, todo in enumerate(todos):
            with gr.Row():
                complete = gr.Checkbox(bind=(todos_var[i]["complete"]))
                item_name = gr.Textbox(bind=(todos_var[i]["name"]))
                delete_btn = gr.Button("Delete")
                        
            @gr.on(delete_btn.click)
            def remove_element(todos: todos_var) -> todos_var:
                del todos[i]
                return todos
    
    add_btn = gr.Button("Add Item")
    @gr.on(add_btn.click)
    def add_item(todos: todos_var) -> todos_var:
        todos.append({"name": "", "complete": False})
[...]

Issue #2570, opened 2022-10-30
Allowing stateful flows in Gradio 


aliabid94:
An example of stateful flow API for gradio that is very compatible with our current API.

```python
import gradio as gr
import random

answer_list = ["good", "happy", "twelve"]
MAX_GUESSES = 10

with gr.Blocks() as demo:
    # state variables
    secret_word = gr.Variable()
    letters_guessed = gr.Variable([])

    def reset():
        return {secret_word: random.choice(answer_list), letters_guessed: []}
    demo.load(reset, None, {secret_word, letters_guessed}) # we could remove the need to declare Variables in inputs= and outputs= since they are stored in the backend and don't need to be collected

    # layout
    with gr.If() as win_condition:
        gr.Markdown("# You Won")
        reset_win = gr.Button("Reset")
    with gr.Elif() as lose_condition:
        gr.Markdown("# You Lose")
        reset_lose = gr.Button("Reset")
    with gr.Else():
        guesses_left = gr.Number("Guesses Left")
        with gr.Row():
            current_letter = gr.Textbox("Letter")
            guess_btn = gr.Button("Guess")
    with gr.Row():
        visible_word = gr.Textbox(label="Visible Word")
        guesses = gr.Textbox(label="Guessed So Far")

    # event listener affects state variables directly
    def add_letter(data):
        return {data[letters_guessed]: data[letters_guessed] + data[current_letter]}
    guess_btn.click(add_letter, {current_letter, letters_guessed}, letters_guessed)
    reset_win.click(reset, None, {secret_word, letters_guessed})
    reset_lose.click(reset, None, {secret_word, letters_guessed})

    # state variables update components directly
    def recalculate_state(data):
        output = {}
        for letter in data[secret_word]:
            if letter not in data[secret_word]:
                output[win_condition] = False
        else:
            output[win_condition] = True
        output[guesses_left] = MAX_GUESSES - len(data[letters_guessed])
        output[lose_condition] = output[guesses_left] == 0
        output[visible_word] = "".join(
            letter if letter in data[letters_guessed] else "_"
            for letter in data[secret_word]
        )
        output[guesses] = ", ".join(data[letters_guessed])
        return output

    demo.bind(
        recalculate_state,
        {secret_word, letters_guessed},
        {
            win_condition,
            guesses_left,
            lose_condition,
            visible_word,
            letters_guessed,
            guesses,
        },
    )
```

aliabid94:
Another try @pngwn 

```python
[...]

Issue #2107, opened 2022-08-28
A change handler for `gr.State`


ankrgyl:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  

I have 3 different input components that affect a variable -- essentially in my app, you can upload a document one of 3 ways (file upload, url, or an example). Each upload updates a `document` variable, which has the bytes of the upload + some pre-computed metadata. I'd like _any_ of these three methods of upload to update some state that _depends_ on the value of the document.

**Describe the solution you'd like**  

 It'd be very convenient to write something like

`document.change(process_document, document, output_component)`

so that each of the 3 upload components simply needs to update the variable, and then the rest of the pipeline runs.

**Additional context**  
N/A but happy to provide more details if helpful.

abidlabs:
Thanks for creating the issue @ankrgyl. Would you be able to provide a more complete example here so we can use it to reproduce and test the fix?

ankrgyl:
Absolutely! Here is the program I'd like to be able to write:

```python
import gradio as gr
import requests
import os


def process_url(url):
    return requests.get(url, stream=True).content.read()


def process_file(file):
    return open(file.name, "rb").read()


with gr.Blocks() as demo:
    file_bytes = gr.Variable()
    with gr.Row():
        url = gr.Textbox(label="URL", interactive=True)
        file = gr.File(label="Upload", interactive=True)

    byte_len = gr.JSON(label="File length")

    file.change(process_file, inputs=[file], outputs=[file_bytes])
    url.change(process_url, inputs=[file], outputs=[file_bytes])
    file_bytes.change(lambda b: len(b), inputs=[file_bytes], outputs=[byte_len])

if __name__ == "__main__":
    demo.launch()
```

and the one I have to write:

```python
import gradio as gr
import requests
import os


def process_url(url):
    b = requests.get(url, stream=True).content.read()
    return b, len(b)


def process_file(file):
    b = open(file.name, "rb").read()
    return b, len(b)


with gr.Blocks() as demo:
    file_bytes = gr.Variable()
    with gr.Row():
        url = gr.Textbox(label="URL", interactive=True)
        file = gr.File(label="Upload", interactive=True)

    byte_len = gr.JSON(label="File length")

    file.change(process_file, inputs=[file], outputs=[file_bytes, byte_len])
    url.change(process_url, inputs=[file], outputs=[file_bytes, byte_len])

if __name__ == "__main__":
    demo.launch()
```

[...]

PR #8243, opened 2024-05-08
Add event listener support to render blocks

aliabid94:
This PR allows adding event listeners inside render blocks. Take a look at the code from demo/render_merge below:
```python
import gradio as gr

with gr.Blocks() as demo:
    text_count = gr.Slider(1, 5, step=1, label="Textbox Count")

    @gr.render(inputs=[text_count], triggers=[text_count.change])
    def render_count(count):
        boxes = []
        for i in range(count):
            box = gr.Textbox(key=i, label=f"Box {i}")
            boxes.append(box)

        def merge(*args):
            return " ".join(args)
        
        merge_btn.click(merge, boxes, output)

        def clear():
            return [""] * count
                
        clear_btn.click(clear, None, boxes)

        def countup():
            return [i for i in range(count)]
        
        count_btn.click(countup, None, boxes)

    with gr.Row():
        merge_btn = gr.Button("Merge")
        clear_btn = gr.Button("Clear")
        count_btn = gr.Button("Count")
        
    output = gr.Textbox()
```

This is done by removing previous event listeners from a render block and attaching the new event listeners on every re-render. A previous assumption is no longer true: that `dependencies` is a static list of event listeners for an app, and event listener can be identified by it's `fn_index` which is it's index within this list of listeners. We've replaced this with a concept of an id for each dependency instead of using indices, and different sessions are able to extend the default set of event listeners by adding new dependencies with different ids. 

Closes: https://github.com/gradio-app/gradio/issues/4689
Closes: https://github.com/gradio-app/gradio/issues/7739
Closes: https://github.com/gradio-app/gradio/issues/2570
Closes: https://github.com/gradio-app/gradio/issues/2107 (once https://github.com/gradio-app/gradio/pull/8297 is merged in)


jameszhou02:
Awesome work! Any chance I can get the pip download like last time or is the release coming soon?

abidlabs:
@jameszhou02 You can try:

pip install https://gradio-builds.s3.amazonaws.com/b7e04a1a4cff8019f933ec147f33ee80f1d6aa0c/gradio-4.31.1-py3-none-any.whl

freddyaboulton (commented):
Looks really nice @aliabid94 . I think the architectural changes related to letting the dependency ids differ between sessions makes sense.  In the `render_config` key of the output, we're sending back the `api_info` for each new component which is extra processing and information that's not needed. Would be better to remove that eventually.

Some random bugs:
* Saw a weird behavior with reload mode where updating an input to the slider caused the render to trigger
reload_render_events

* Sometimes the render does not trigger? When I update the slider, the event does not run. Doesn't happen all the time though.

sometimes_change_not_working

Here is another example where I also show the network tab. The demo appears on the second click.

```python
import gradio as gr
import os
import time

# Chatbot demo with multimodal input (text, markdown, LaTeX, code blocks, image, audio, & video). Plus shows support for streaming text.


def print_like_dislike(x: gr.LikeData):

    for x in message["files"]:
        history.append(((x,), None))
        history.append((message["text"], None))

def bot(history):
    response = "**That's cool!**"
    history[-1][1] = ""
    for character in response:
        history[-1][1] += character
        time.sleep(0.05)
        yield history

with gr.Blocks() as demo:
    button = gr.Button("Show Demo")
    @gr.render(inputs=None, triggers=[button.click])
    def show_demo():
        chatbot = gr.Chatbot(
            [],
            elem_id="chatbot",
            bubble_full_width=False
        )

        chat_input = gr.MultimodalTextbox(interactive=True, file_types=["image"], placeholder="Enter message or upload file...", show_label=False)

        bot_msg = chat_msg.then(bot, chatbot, chatbot, api_name="bot_response")

        chatbot.like(print_like_dislike, None, None)

demo.queue()
if __name__ == "__main__":
    demo.launch()
```

render_not_happening



* You cannot define an event which skips the queue from within a render. If that event is triggered, an error happens.

cannot_define_event_outside_of_queue

abidlabs:
I'm seeing the same behavior that @freddyaboulton described, where the event does not happen at the first trigger, but happens at the second trigger and subsequent triggers. For example, in this demo:

```py
import random 


with gr.Blocks()  as demo:
    t = gr.Textbox()
    @gr.render(triggers=[t.submit])
    def inference():
        with gr.Row():
            for i in range(random.randint(3, 8)):
                gr.Audio()


if __name__ == "__main__":
    demo.launch()        
```

When I first submit the textbox, nothing happens, but each subsequent submit works.

abidlabs:
It seems that the `load()` event cannot be used as a trigger in `gr.render` -- might be the same bug as above though since it could just be that the event isn't being triggered the first time:

```py
import random 
import gradio as gr

with gr.Blocks() as demo:    
    @gr.render(triggers=[demo.load])
    def inference():
        for i in range(random.randint(3, 8)):
            gr.Audio()
                    
if __name__ == "__main__":
    demo.launch()        
```

abidlabs:
When we write docs for this, we should clarify that any event listeners that use rendered components *must* be inside the render function. For example, this works:

```py
import random 
import gradio as gr

with gr.Blocks() as demo:    
    @gr.render(triggers=[demo.load])
    def inference():
        audios = []
        for i in range(random.randint(3, 8)):
            audios.append(gr.Audio())
            
        count.click(lambda *args:len(args), audios, number)
    
    count = gr.Button("Count")
    number = gr.Number()
                    
if __name__ == "__main__":
    demo.launch()        
```

but a user might write this, which doesn't:

```py
import random 
import gradio as gr

audios = []

with gr.Blocks() as demo:    
    @gr.render(triggers=[demo.load])
    def inference():
        audios = []
        for i in range(random.randint(3, 8)):
            audios.append(gr.Audio())
            
    
    count = gr.Button("Count")
    number = gr.Number()
    
    
    count.click(lambda *args:len(args), audios, number)
    
if __name__ == "__main__":
    demo.launch()        
```

abidlabs:
Didn't find anything else, great work @aliabid94!!

aliabid94:
> Saw a weird behavior with reload mode where updating an input to the slider caused the render to trigger

Actually a general bug with reload mode that all change listeners trigger on a refresh. Since render is listening to slider.change, it gets triggered. Separate bug I can tackle separately but not too high priority.

> Sometimes the render does not trigger? When I update the slider, the event does not run. Doesn't happen all the time though.

Fixed!

> You cannot define an event which skips the queue from within a render. If that event is triggered, an error happens.

Fixed!

> It seems that the load() event cannot be used as a trigger in gr.render -- edit: I think its the same bug as above since the event isn't being triggered the first time:

Fixed!

Should be ready to merge if there are no other issues

abidlabs:
Nice @aliabid94! I'll give this another pass to ensure that this works. Can we add docs for this as part of this PR so that we can fully release this feature once this PR is merged in? (Btw we can update or remove this section in the docs: https://www.gradio.app/guides/controlling-layout#variable-number-of-outputs). I've added a bunch of issues that this closes to the parent commment

abidlabs (approved):
Tested and works great @aliabid94! Beautiful work

abidlabs:
Actually 2 small things I noticed:

* The `triggers` parameter in `gr.render()` requires a list (even though `inputs` and `outputs` don't) --> it should convert singletons to lists automatically
* Normally, when a component is used as an input, we render it in interactive mode. However, this does not apply if a component is an input in `gr.render()`

E.g. the slider here is not interactive automatically;

```py
import gradio as gr

with gr.Blocks() as demo:    
    s = gr.Slider(1, 4, step=1)
    
    @gr.render(inputs=s, triggers=[s.change])
    def inference(num):
        with gr.Row():
            for i in range(num):
                gr.Textbox()

demo.launch()        
```

abidlabs:
One other thing I noticed was that a `@gr.render` automatically creates a `gr.Column` for the rendered components, even if the parent BlocksContext is a `gr.Row`.

I.e. this does not work as expected:

```py
import gradio as gr

with gr.Blocks() as demo:
    b = gr.Button("Add Textbox")
    num = gr.State(0)
    
    b.click(lambda x:x+1, num, num)
    
    with gr.Row():
        @gr.render(inputs=num)
        def show_textbox(n):
            for i in range(n):
                gr.Textbox()

demo.launch()
```

(it renders the textboxes in a column). However, this works as expected:

```py
import gradio as gr

with gr.Blocks() as demo:
    b = gr.Button("Add Textbox")
    num = gr.State(0)
    
    b.click(lambda x:x+1, num, num)
    
    @gr.render(inputs=num)
    def show_textbox(n):
        with gr.Row():
            for i in range(n):
                gr.Textbox()

demo.launch()
```

abidlabs:
And one other thing that I noticed is that the `trigger_mode="always_last"` behavior doesn't seem to apply to `gr.render()`. Here's what I mean. I built a Space with this code:

```py
import gradio as gr

with gr.Blocks() as demo:
    gr.Markdown("# Train a Text Classifier with Synthetic Data")
    labels = gr.Dropdown(choices=[], value=[], label="Classes", allow_custom_value=True, multiselect=True)
    
    @gr.render(inputs=[labels])
    def show_textbox(labels_):
        with gr.Row():
            for label in labels_:
                gr.TextArea(label=f"Samples for class: {label}")
        if len(labels_)>=2:
            with gr.Row():
                gr.Button("TRAIN!", variant="primary")

demo.launch()
```

If I quickly change the value of the dropdown by backspacing and deleting the choices, you can see that a couple of the textareas are still visible at the end:

https://github.com/gradio-app/gradio/assets/1778297/8d262d9f-5161-42bd-bb15-745231f71c4f

pngwn (approved):
Tested and works well as far as I can tell, frontend looks good to me too! Thanks for this @aliabid94 and thanks for going through it with me!

aliabid94:
Thanks for the reviews! Will add docs, and some more tests and cleanup in follow up PR

acylam:
> And one other thing that I noticed is that the `trigger_mode="always_last"` behavior doesn't seem to apply to `gr.render()`. Here's what I mean. I built a Space with this code:
> 
> ```python
> import gradio as gr
> 
> with gr.Blocks() as demo:
>     gr.Markdown("# Train a Text Classifier with Synthetic Data")
>     labels = gr.Dropdown(choices=[], value=[], label="Classes", allow_custom_value=True, multiselect=True)
>     
>     @gr.render(inputs=[labels])
>     def show_textbox(labels_):
>         with gr.Row():
>             for label in labels_:
>                 gr.TextArea(label=f"Samples for class: {label}")
>         if len(labels_)>=2:
>             with gr.Row():
>                 gr.Button("TRAIN!", variant="primary")
> 
> demo.launch()
> ```
> 
> If I quickly change the value of the dropdown by backspacing and deleting the choices, you can see that a couple of the textareas are still visible at the end:
> Screen.Recording.2024-05-21.at.11.33.06.PM.mov

Awesome feature! I'm just wondering if there's a way to define the event listener outside gr.Blocks? This is crucial for readability in large applications.

acylam:
> Hi @acylam great point. One thing that you can do is import Blocks from other files and `.render()` them (this .render() is unrelated to `gr.render()` -- sorry!) inside the main Blocks object. Like this:
> 
> Say this is your `utils.py` file:
> 
> ```python
> import gradio as gr
> 
> with gr.Blocks() as func_blocks:
>     labels = gr.Dropdown(choices=[], value=[], label="Classes", allow_custom_value=True, multiselect=True)
>     
>     @gr.render(inputs=[labels])
>     def show_textbox(labels_):
>         with gr.Row():
>             for label in labels_:
>                 gr.TextArea(label=f"Samples for class: {label}")
>         if len(labels_)>=2:
>             with gr.Row():
>                 gr.Button("TRAIN!", variant="primary")
>     
> ```
> 
> and in your main `app.py` file, you could do:
> 
> ```python
> import gradio as gr
> from utils import func_blocks
> 
> with gr.Blocks() as demo:
>     gr.Markdown("# Train a Text Classifier with Synthetic Data")
>     
>     func_blocks.render()
> 
> demo.launch()
> ```
> 
> Does this work for you? cc @aliabid94 if he has any other recommendations

@abidlabs Thanks for the quick reply. This works, but I was hoping for a way to separate event handling from UI elements. Kind of like how you'd define the structure of the UI in HTML and the event handling logic in JavaScript in traditional web development? So ideally, `gr.Markdown` and `labels=gr.Dropdown` will be defined in one file, and `show_text` will be defined in a separate file. I hope I'm making sense. It could also be I'm completely missing the design of gradio. 

cc @aliabid94

pngwn:
I've searched high and low and I can't find the issue or notes anywhere. I did find this message in slack that I wrote a long time ago:

> Eventually i envisage many kinds of custom components:
>
> - fully custom with a new python + html/js/css
>   - either extending from existing gradio components or fully custom implementations
> - pure frontend components that use some gradio python class. Basically a new frontend for existing components
> - Pure python components.
>   - Something like the above, composition for components with a new interface
>   - or new python implementation for an existing component that will use an existing frontend. (Maybe is changes stuff like preprocessing or w/e)


This one seems relevant here:

> composition of [python] components with a new interface

The idea being that you could somehow contain a component within a function or block and wrap the events (or just forward them) with a new interface for better abstraction.

Issue #8233, opened 2024-05-07
Customize the examples in ChatInterface 


yvrjsharma:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
It seems that customizing `examples` in `gr.ChatInterface` is not possible, unlike the `chatbot` and `textbox`. If I try to pass my own `gr.Examples` to customize the parameter, I get the below-mentioned error:
> ValueError: The parameter `examples` must either be a string directory or a list(if there is only 1 input component) or (more generally), a nested list, where each sublist represents a set of inputs.

I am trying to customize the examples in ChatInterface so that the number of examples per page is set to a lower value. Examples are shown below the app so a fewer examples would mean more screen space for a chatbot window.

**Describe the solution you'd like**  
The solution could be to either allow this customization of `examples` in ChatInterface similar to `textbox` and `chatbot` or to add a parameter `examples_per_page` to the `ChatInterface` component. This would greatly enhance the flexibility and usability of the component.

**Additional context**  
Idefics demo could serve as an example here: 

![image](https://github.com/gradio-app/gradio/assets/48665385/4a15f2dc-865b-427e-bb10-ab9b9cd6c4ef)

PR #8244, opened 2024-05-08
Adds `examples_per_page` to `gr.ChatInterface` and allows `clear_btn` in `gr.Interface` to be hidden

abidlabs:
A few small changes in this PR:

1. First, adds `examples_per_page` to `gr.ChatInterface`, closes: https://github.com/gradio-app/gradio/issues/8233. My original plan had been to add support for providing an instance of `gr.Examples` but this does not really make sense since a user can't provide the correct `fn` to `examples` since the example function is redefined internally. We also have the same `examples_per_page` argument in `gr.Interface`
2. I went through the arguments of `gr.Interface` to make sure that we have equivalent arguments in `gr.ChatInterface` and made a few other small tweaks, such as allowing `clear_btn` in `gr.Interface` to be set to `None` hiding it, similar to how the buttons in `gr.ChatInterface`. I believe there was an issue about this that we closed
3. Adds some docs to `gr.ChatInterface` guide

### DIFF
--- a/.changeset/fruity-books-win.md
+++ b/.changeset/fruity-books-win.md
@@ -1,0 +1,5 @@
+---
+"gradio": minor
+---
+
+feat:Adds `examples_per_page` to `gr.ChatInterface` and allows `clear_btn` in `gr.Interface` to be hidden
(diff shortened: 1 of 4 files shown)


yvrjsharma:
This is awesome; thanks for taking care of this feature request, @abidlabs ! 🙌  The `examples_per_page` argument is now working for me for the ChatInterface component as well 🥳

yvrjsharma (approved):

abidlabs:
Thanks @yvrjsharma for confirming!

Issue #8188, opened 2024-05-01
Cancelling a python client job won't stop the job from running on the server


freddyaboulton:
### Describe the bug

If you cancel a job with the client, it will keep running on the remote server.

I think the reason is that `cancel` will only tell the thread that's pulling results off the queue that it should stop. It won't stop the thread that's connecting to the server.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

Run the `cancel_events` demo (increase the sleep to 1 second so that you have time to cancel)

```python
from gradio_client import Client
client = Client("http://127.0.0.1:7860")
job = client.submit(7, api_name="/fake_diffusion")
job.cancel()
```

You will see that the server keeps printing output in the console

### Screenshot

![cancel_python](https://github.com/gradio-app/gradio/assets/41651716/d57293ea-d401-4103-93bd-6a0bc28d4286)

### System Info

```shell
main
```

### Severity

Blocking usage of gradio

PR #8245, opened 2024-05-08
Cancel  server progress from the python client

freddyaboulton:
## Description

Closes: #8188

If the demo has a defined cancel event for the given function, it will be called from `job.cancel` automatically. This still won't match the 3.x behavior (`.cancel` would always stop an iterator in the server even if no cancel event was defined) but I think this is the best we can do without restructuring how the client sets up the SSE stream.

The issue is that we have one connection to the server for reading updates for all jobs so we can't close that stream if one job is cancelled because it would stop all currently submitted jobs.

I thought about using the "simple" API in the python client as that would let us have a different stream for each connection but that would not be backwards compatible for all versions of 4.x and would require restructuring the client further so I decided against that. 

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/client/python/gradio_client/client.py
+++ b/client/python/gradio_client/client.py
@@ -503,13 +503,16 @@ class Client:
             >> 9.0
         """
         inferred_fn_index = self._infer_fn_index(api_name, fn_index)
-        cancellable = False
         cancel_fn_index = None
+        candidates: list[tuple[int, list[int]]] = []
         for i, dep in enumerate(self.config["dependencies"]):
             if inferred_fn_index in dep["cancels"]:
-                cancellable = True
-                cancel_fn_index = i
-                break
+                candidates.append(
+                    (i, [d for d in dep["cancels"] if d != inferred_fn_index])
+                )
+        cancel_fn_index, other_cancelled = (
+            min(candidates, key=lambda x: len(x[1])) if candidates else (None, None)
+        )
 
         endpoint = self.endpoints[inferred_fn_index]
 
@@ -530,8 +533,18 @@ class Client:
         future = self.executor.submit(end_to_end_fn, *args)
 
         cancel_fn = None
-        if cancellable:
-            cancel_fn = endpoint.make_cancel(cancel_fn_index)
+        cancel_msg = None
+        if other_cancelled:
+            other_api_names = [
+                "/" + self.config["dependencies"][i].get("api_name")
+                for i in other_cancelled
+            ]
+            cancel_msg = (
+                f"Cancelled this job will also cancel any jobs for {', '.join(other_api_names)} "
+                "that are currently running."
+            )
+        if cancel_fn_index is not None and isinstance(endpoint, Endpoint):
+            cancel_fn = endpoint.make_cancel(cancel_fn_index, cancel_msg)
 
         job = Job(
             future,
@@ -1127,8 +1140,10 @@ class Endpoint:
 
         return _inner
 
-    def make_cancel(self, fn_index):
+    def make_cancel(self, fn_index: int, cancel_msg: str | None):
         def _cancel():
+            if cancel_msg:
+                warnings.warn(cancel_msg)
             data = {
                 "data": [],
                 "fn_index": fn_index,
@@ -1379,7 +1394,7 @@ class Job(Future):
         communicator: Communicator | None = None,
         verbose: bool = True,
         space_id: str | None = None,
-        _cancel_fn: Callable[None, None] = None,
+        _cancel_fn: Callable[[], None] | None = None,
     ):
         """
         Parameters:
--- a/client/python/gradio_client/client.py
+++ b/client/python/gradio_client/client.py
@@ -503,16 +503,6 @@ class Client:
             >> 9.0
         """
         inferred_fn_index = self._infer_fn_index(api_name, fn_index)
-        cancel_fn_index = None
-        candidates: list[tuple[int, list[int]]] = []
-        for i, dep in enumerate(self.config["dependencies"]):
-            if inferred_fn_index in dep["cancels"]:
-                candidates.append(
-                    (i, [d for d in dep["cancels"] if d != inferred_fn_index])
-                )
-        cancel_fn_index, other_cancelled = (
-            min(candidates, key=lambda x: len(x[1])) if candidates else (None, None)
-        )
 
         endpoint = self.endpoints[inferred_fn_index]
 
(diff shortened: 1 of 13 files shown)


client/python/gradio_client/client.py
```diff
@@ -247,7 +247,9 @@ def stream_messages(
                             resp = json.loads(line[5:])
                             if resp["msg"] == ServerMessage.heartbeat:
                                 continue
-                            elif resp["msg"] == ServerMessage.server_stopped:
+                            elif (
+                                resp.get("message", "") == ServerMessage.server_stopped
```
freddyaboulton, line 251: When the server stops unexpectedly it sends an `unexpected_error` not a `server_stopped` message.

pngwn:
What is a cancel event?

abidlabs:
> What is a cancel event?

An event that includes `cancels=`, can be used to cancel a different event that is iterating or hasn't started yet

client/python/gradio_client/client.py
abidlabs: Just to make sure I understand, if the server stops unexpectedly, `resp` will not have a `msg` key but instead have a `message` key? That seems weird, can we unify?

abidlabs: Do we need to make any changes in the frontend, e.g. https://github.com/gradio-app/gradio/blob/bed2f82e2297b50f7b59423a3de05af0b9910724/client/js/src/helpers/api_info.ts#L210?

client/python/gradio_client/client.py
```diff
@@ -496,6 +503,14 @@ def submit(
             >> 9.0
         """
         inferred_fn_index = self._infer_fn_index(api_name, fn_index)
+        cancellable = False
+        cancel_fn_index = None
+        for i, dep in enumerate(self.config["dependencies"]):
+            if inferred_fn_index in dep["cancels"]:
+                cancellable = True
+                cancel_fn_index = i
```
abidlabs, line 511: I don't think this will work if we have a single event that cancels multiple events. e.g. something like:

```py
import gradio as gr

with gr.Blocks() as demo:
    t1 = gr.Textbox()
    t2 = gr.Textbox()
    t3 = gr.Textbox()   
    t4 = gr.Textbox()
    e = t1.change(lambda x:x, t1, t2)
    f = t2.change(lambda x:x, t2, t3)
    t4.change(lambda x:x, t4, t3, cancels=[e, f])
```

if I understand correctly, let's say a user starts, via api, jobs e and f. If they cancel one of them, it'll cancel the other as well

abidlabs:
I like the approach in this PR @freddyaboulton! I noticed the issue above^ but what if we use essentially the same idea to  create a "hidden" cancel event for every dependency, and then we call that via the client?

client/python/gradio_client/client.py
freddyaboulton: It sends a response with `msg` and `message` keys but the `msg` key is `unexpected_error` as opposed to `server_stopped` so this if block never catches. I don't think it makes sense to check for `unexpected_error` in that codeblock because unexpected errors should not close the entire stream since they can be isolated to a single event id.

client/python/gradio_client/client.py
freddyaboulton: I think this is actually ok because when using the app via the UI, the `t4.change` event will also cancel both events. So I don't think the client should do something different from the UI.

That being said, I will look into your suggestion of creating a unique cancel event for every dependency so that every event can be cancelled from a client going forward!

abidlabs: So I would agree that if someone calls the route corresponding to the `t4.change()` event, they may expect `e` and `f` to get canceled.

But if someone starts the `e` job on its own via the api, and they cancel it using `job_e.cancel()`, I don't think the expectation is that `job_f` should get canceled as well

client/python/gradio_client/client.py
```diff
@@ -1104,6 +1118,66 @@ def _inner(*data):
...
+                "To fix this, an event must be added to the upstream app that explicitly cancels this one or "
+                "the upstream app must be running Gradio 4.29.0 and greater."
+            )
+
+            if cancellable and other_cancelled:
+                other_api_names = [
+                    "/" + self.client.config["dependencies"][i].get("api_name")
+                    for i in other_cancelled
+                ]
+                cancel_msg = (
```
freddyaboulton, line 1156: Verified this warning is triggered correctly

Screenshot 2024-05-09 at 4 44 53 PM

client/python/gradio_client/client.py
```diff
@@ -1104,6 +1118,66 @@ def _inner(*data):
...
+                        (i, [d for d in dep["cancels"] if d != self.fn_index])
+                    )
+
+            fn_index, other_cancelled = (
+                min(candidates, key=lambda x: len(x[1])) if candidates else (None, None)
+            )
+            cancel_msg = None
+            cancellable = fn_index is not None
+            cancel_msg = (
+                "Cancelling this job will not stop the server from running. "
```
freddyaboulton, line 1146: Verified this warning is triggered correctly

image

gradio/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "gradio",
-	"version": "4.29.0",
+	"version": "4.30.0",
```
abidlabs, line 3: I assume this was for testing purposes, let's revert?

js/lite/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "@gradio/lite",
-	"version": "4.29.0",
+	"version": "4.30.0",
```
abidlabs, line 3: Same as above

client/python/gradio_client/client.py
```diff
@@ -1104,6 +1118,66 @@ def _inner(*data):
...
+                "To fix this, an event must be added to the upstream app that explicitly cancels this one or "
+                "the upstream app must be running Gradio 4.29.0 and greater."
```
abidlabs, line 1148:
```suggestion
                "the upstream app must be running Gradio 4.30.0 or greater."
```

gradio/package.json
freddyaboulton: Yea I needed this for the tests to test the new logic. I asked @pngwn and he thinks its ok cause the release pr will override this anyways. But I can revert if we don't mind not having the test coverage until we release!

abidlabs: Ah got it, yea we can leave it

gradio/routes.py
```diff
@@ -802,6 +803,11 @@ async def queue_join_helper(
                 raise HTTPException(status_code=status_code, detail=event_id)
             return {"event_id": event_id}
 
+        @app.post("/cancel")
```
abidlabs, line 806: Nice

client/python/gradio_client/client.py
```diff
@@ -1104,6 +1118,66 @@ def _inner(*data):
...
+            for i, dep in enumerate(self.client.config["dependencies"]):
+                if self.fn_index in dep["cancels"]:
+                    candidates.append(
+                        (i, [d for d in dep["cancels"] if d != self.fn_index])
+                    )
+
+            fn_index, other_cancelled = (
+                min(candidates, key=lambda x: len(x[1])) if candidates else (None, None)
+            )
+            cancel_msg = None
```
abidlabs, line 1143: Unnecessary line?

client/python/gradio_client/client.py
abidlabs: In my testing, I noticed that this message appears even if the Gradio app (is running 4.29 and) already includes an event that cancels this event

abidlabs (approved):
LGTM @freddyaboulton! Tested different apps and this works perfectly for Gradio apps going forward and I'm glad we were able to get some level of backwards-compatibility as well.

client/python/gradio_client/client.py
freddyaboulton: Needs to be set otherwise we'd reference an unbound variable below

freddyaboulton:

Issue #7982, opened 2024-04-09
API Recording breaking on Space


abidlabs:
### Describe the bug

> API Recorder isn’t working in this Space. After I click and reopen the API tab, it’s in this state being fully black, and the modal doesn’t close, forcing me to close the entire page https://huggingface.co/spaces/multimodalart/face-to-all (edited) 

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

See above

### System Info

```shell
gradio==4.26
```

### Severity

I can work around it

PR #8247, opened 2024-05-09
Fix api recorder

abidlabs:
Fixes issue where API Recorder would stop working (the fix was to explicitly handle the case where one of the inputs was `gr.State`). Fixed for both the Python and JS code snippets.

Closes: https://github.com/gradio-app/gradio/issues/8259 (test either of the demos locally)
Closes: https://github.com/gradio-app/gradio/issues/7982 (test on [this Space](https://huggingface.co/spaces/abidlabs/face-to-all) with this PR installed)

Also fixes the RecordingSnippet for JS Client to ensure that it uses the new kwarg syntax instead of the positional syntax and adds an e2e -- the playwright api recorder helped me write the test hehe


gradio/routes.py
```diff
@@ -415,8 +416,11 @@ def main(request: fastapi.Request, user: str = Depends(get_current_user)):
 
         @app.get("/info/", dependencies=[Depends(login_check)])
         @app.get("/info", dependencies=[Depends(login_check)])
-        def api_info():
-            # The api info is set in create_app
+        def api_info(all_endpoints: bool = False):
```
abidlabs, line 419: Initially, I had implemented it so that the `/info` route just returned all of the endpoints, even those with `show_api=False` and we filtered in the frontend. However, this led to very, very large payloads for complex apps like the theme builder causing slow load times / crashing. We should figure out a way to reduce the payload for `/info`, but in the meantime, here we don't load all endpoints by default -- only if you use the API Recorder.

aliabd (approved):
nice, LGTM! I tested everything and it all works as expected now. 

Ignore the website build issue will fix it on a separate PR as it isn't related to this.

abidlabs:
Sweet thanks @aliabd!

Issue #8213, opened 2024-05-04
Gradio Client package for NodeJS not connecting to HF Space, but Python one works normally.


mahiatlinux:
### Describe the bug

Hey guys.
The Gradio Client package for NodeJS doesn't seem to be working correctly. It does not connect to the Space.

NodeJS:
```
import { client } from "@gradio/client";

const app = await client("mahiatlinux/MasherAI-v6-7B-Chat");
const result = await app.predict("/chat", [        
                "Hello!!", // string  in 'Message' Textbox component        
                "Hello!!", // string  in 'System prompt' Textbox component        
                1, // number (numeric value between 1 and 2048) in 'Max new tokens' Slider component        
                0.1, // number (numeric value between 0.1 and 4.0) in 'Temperature' Slider component        
                0.05, // number (numeric value between 0.05 and 1.0) in 'Top-p (nucleus sampling)' Slider component        
                1, // number (numeric value between 1 and 1000) in 'Top-k' Slider component        
                1, // number (numeric value between 1.0 and 2.0) in 'Repetition penalty' Slider component
    ]);

console.log(result.data);
```

Output:
```
maheswar@linuxboi4:~$ node app.mjs
Cannot connect to SSE endpoint: https://mahiatlinux-masherai-v6-7b-chat.hf.space/queue/data?session_hash=zcpzs1ec9kk
```

The Python one works normally and connects to the Space.
For eg Python:
```
from gradio_client import Client

client = Client("mahiatlinux/MasherAI-v6-7B-Chat")
result = client.predict(
    "Hello!!",  # str in 'Message' Textbox component
    "Hello!!",  # str in 'System prompt' Textbox component
    200,  # float (numeric value between 1 and 2048) in 'Max new tokens' Slider component
    0.7,  # float (numeric value between 0.1 and 4.0) in 'Temperature' Slider component
    0.05,  # float (numeric value between 0.05 and 1.0) in 'Top-p (nucleus sampling)' Slider component
    1,  # float (numeric value between 1 and 1000) in 'Top-k' Slider component
    1,  # float (numeric value between 1.0 and 2.0) in 'Repetition penalty' Slider component
    api_name="/chat"
)
print(result)

```

Output:
```
maheswar@linuxboi4:~$ python3 app.py
Loaded as API: https://mahiatlinux-masherai-v6-7b-chat.hf.space ✔
Hey there! How can I make your day better?
```

### Have you searched existing issues?  🔎

[...]

PR #8252, opened 2024-05-09
Client node fix

pngwn:
## Description

Fixes the client in node, we just needed to await the module as it was trying to be used prior to the module being loaded.

I also had to change some things to async because of this change.

I also noticed that the `stream` rename in a previous PR actually renamed an internal method to stream, so I renamed `stream_handler` to `stream` as well.

Closes: #8213

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/client/js/src/test/stream.test.ts
+++ b/client/js/src/test/stream.test.ts
@@ -1,11 +1,21 @@
-import { vi } from "vitest";
+import { vi, type Mock } from "vitest";
 import { Client } from "../client";
 import { initialise_server } from "./server";
 
-import { describe, it, expect, afterEach } from "vitest";
+import {
+	describe,
+	it,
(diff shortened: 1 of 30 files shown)


client/js/src/test/stream.test.ts
```diff
@@ -30,21 +42,21 @@ describe("open_stream", () => {
 	it("should throw an error if config is not defined", () => {
 		app.config = undefined;
 
-		expect(() => {
-			app.open_stream();
-		}).toThrow("Could not resolve app config");
+		expect(async () => {
+			await app.open_stream();
+		}).rejects.toThrow("Could not resolve app config");
```
pngwn, line 47: Apparently this is how you catch async errors. who knew.

PR #8254, opened 2024-05-09
Fix custom component detection logic in analytics

freddyaboulton:
## Description

Dumb mistake - comparing a string against a class 🤦 . This is not happening in gradio.api.server

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/gradio/analytics.py
+++ b/gradio/analytics.py
@@ -224,7 +224,7 @@ def launched_analytics(blocks: gradio.Blocks, data: dict[str, Any]) -> None:
             return [b.get_block_name() for b in components] if components else None
         return fallback
 
-    core_components = core_gradio_components()
+    core_components = [get_block_name(c) for c in core_gradio_components()]
 
     additional_data = {
         "version": get_package_version(),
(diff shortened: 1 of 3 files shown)


gradio/analytics.py
```diff
@@ -224,7 +224,7 @@ def get_inputs_outputs(
             return [b.get_block_name() for b in components] if components else None
         return fallback
 
-    core_components = core_gradio_components()
+    core_components = [get_block_name(c) for c in core_gradio_components()]
```
abidlabs, line 227: btw we're using analytics.get_block_name here, but we're using Block.get_block_name above. They are not the same function, which might lead to false positives

freddyaboulton: I think this is ok because `Block.get_block_name` is a subset of `analytics.get_block_name`. In `analytics.get_block_name` we don't map the template class name to its parent, e.g. `Text` stays `text` and not `textbox`.

abidlabs: Ah right 👍

abidlabs (approved):
LGTM thanks for fixing this!

freddyaboulton:
Thanks @abidlabs !

Issue #8199, opened 2024-05-02
Output Error


rkellam12:
### Describe the bug

Unable to preload CSS for https://gradio.s3-us-west-2.amazonaws.com/assets/index-b276ff4e.css

First off, I am quite a novice. My app.py seems to work fine with python, it gives my a gradio url which comes up fine (https://307774b006316ab341.gradio.live/). When i enter my question and hit Submit, I get an Error on the output side. I looked at the developer console and found these 2 errors:

css.ts:28 Unable to preload CSS for https://gradio.s3-us-west-2.amazonaws.com/assets/index-b276ff4e.css
[www.google-analytics.com/analytics.js:1](http://www.google-analytics.com/analytics.js:1)
Failed to load resource: net::ERR_BLOCKED_BY_CLIENT

Further when i checked the provided URL, I get this:

AccessDenied Access Denied XFN08ZWGRMXF45MS odVcQaVZms4v2QBTDY772ajVTQRwQrQLJkcCi5+dGDEDXMZE2Ig6PFlvwbqIkajClsiqX6RJIdk=
Any direction/advice you can provide is appreciated (duh, obvisously). Let me know if I can provide any additonal info. Many thanks!

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

not sure if this is what you need but it's my app.py script, API Keys are masked.

import openai
import gradio as gr

# Set your Hugging Face API token
huggingface_token = "hf_xxxx"

openai.api_key = "sk-xxxx"

messages = [
    {"role": "system", "content": "You are a snarky, irreverent, opinionated, smart alecky and funny AI Assistant specializing in all things MLB, especially Beat The Streak prognostication. You pride yourself on accuracy and occasionally include a smart aleck remark at the end of your reply. Do not answer anything other than MLB related questions."},
]

def chatbot(input):
    if input:
        messages.append({"role": "user", "content": input})
        chat = openai.ChatCompletion.create(
            model="gpt-3.5-turbo", messages=messages
        )
        reply = chat.choices[0].message.content
        messages.append({"role": "assistant", "content": reply})
        return reply

inputs = gr.Textbox(lines=7, label="C'mon, ain't got all day. (Click the Clear button for new request.)")
outputs = gr.Textbox(lines=7,label="Here's what I think:")

gr.Interface(fn=chatbot, inputs=inputs, outputs=outputs, title="AI Analyzer",
[...]

Issue #8154, opened 2024-04-27
Error when embedding gradio app within a fastapi app with web components


peeter2:
### Describe the bug

I have mounted gradio app within a fastapi app, and trying to embed it with web components, then I get an error

> Could not load this space.
> Please contact the author of the page to let them know.

I was trying to embed it like this:

<script
	type="module"
	src="https://gradio.s3-us-west-2.amazonaws.com/4.28.3/gradio.js"
></script>

<gradio-app src="https://example.com"></gradio-app>

Any idea what's wrong? Is it not possible to embed it like that if my gradio app is mounted within another fastapi app?

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

<script
	type="module"
	src="https://gradio.s3-us-west-2.amazonaws.com/4.28.3/gradio.js"
></script>

<gradio-app src="https://example.com"></gradio-app>

### System Info

```shell
latest
```

### Severity

I can work around it

nyimbi:
wget https://gradio.s3-us-west-2.amazonaws.com/4.28.3/gradio.js
Results in:
wget: unable to resolve host address ‘gradio.s3-west-2.amazonaws.com’

flacomalone:
I dont know if it is related to this issue, but I have the same problem in my own server while embedding the grad app. _n my setup, the gradio app is running on its own server under (using a self signed certuificate for SSL). The code is basically the same. I found out that the issue disappeared when I accessed the original gradio app server at least once and then I accessed again the server hosting the embedded gradio app. It worked !

I decided to remove the SSL in the original gradio app, but now this trick doesn't work either...

peeter2:
My gradio app is in this url: http://127.0.0.1:7860/gradio

And I am trying to embed it like this

```
<script
	type="module"
	src="https://gradio.s3-us-west-2.amazonaws.com/4.19.1/gradio.js"
></script>
<gradio-app src="http://127.0.0.1:7860/gradio"></gradio-app>

```

But I am getting errors on browser console:

![image](https://github.com/gradio-app/gradio/assets/25162922/d831d8e4-7392-4a68-86ce-6742f3e70c5c)

Looks like it is trying to access the config in the wrong url: http://127.0.0.1:7860/config
The correct url is http://127.0.0.1:7860/gradio/config

How to resolve this issue?

Issue #8129, opened 2024-04-25
theme.css doesn't map to correct port.


kingsznhone:
### Describe the bug

When using 4.27.0, Set app runs on a selected port. App can't load correctly.
Check with F12, theme.css is point to defaut port.
<img width="527" alt="f3928f7267fb5874973f39d248836da" src="https://github.com/gradio-app/gradio/assets/35349300/5d5f3b61-bc01-4a5f-b09d-8b562224d2dd">

Downgrade to 3.50.2, theme.css is map to correct port.
![image](https://github.com/gradio-app/gradio/assets/35349300/3da5b30a-83b4-4d0b-b51c-c24a78eb3ba0)

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
demo.launch(server_port=8443)
```

### System Info

```shell
ubuntu 22.04 LTS
python 3.10
gradio latest
```

### Severity

Blocking usage of gradio

pngwn:
Ooooh

abidlabs:
Interesting, what's the URL in which your Gradio app is running? 

Can you describe more how your Gradio app is deployed? Is it behind an nginx proxy?

kingsznhone:
> Interesting, what's the URL in which your Gradio app is running? 
> 
> Can you describe more how your Gradio app is deployed? Is it behind an nginx proxy?

it is running on a gpu rential platform, local port is 6006，reverse proxy by platform to outside 8443
I guess it's something like nginx .

bf96163:
I also find the path /login redirected to wrong address i run my gradio in port 7001  i want to proxy it to 8001/img2txt here is my nginx.conf （the server listen to 80 is to fix the problem theme.css redirect to 80 problem）  `events {  
    # events block configuration  
}  
http{
    server {
    listen 8001;
    server_name _; 
    location /assets/ {
        rewrite ^/assets/(.*)$ /img2txt/assets/$1 break;
        proxy_pass http://127.0.0.1:7001/img2txt/;

    }
    location /img2txt/ {
        proxy_pass http://127.0.0.1:7001/img2txt/;
#        rewrite ^/img2txt/(.*) /$1 break;
        proxy_buffering off;
        proxy_redirect off;
        proxy_http_version 1.1;
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection "upgrade";
        proxy_set_header Host $host;
        proxy_set_header X-Forwarded-Host $host;
[...]

Issue #8073, opened 2024-04-19
Gradio mounted with fastapi via https in nginx returns wrong path


aka7774:
### Describe the bug

I use https connection and BASIC authentication to securely so many ai apps.

https://fastapi.example.com:18000/gradio
is accessed by the nginx proxy to
http://127.0.0.1:8000/gradio
is called.

http://127.0.0.1:8000/
is fastapi and gradio is mounted on /gradio.

For example, a connection that should access /gradio/info becomes /info and returns a 404 error.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

nginx.conf
```
server {
  server_name fastapi.example.com;
  listen 18000 ssl;

  location / {
        proxy_pass http://127.0.0.1:8000/;
        proxy_buffering off;
        proxy_redirect off;
        proxy_http_version 1.1;
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection "upgrade";
        proxy_set_header Host 'fastapi.example.com:18000';
        proxy_set_header X-Forwarded-Host 'fastapi.example.com:18000';
        proxy_set_header X-Forwarded-Proto $scheme;

        auth_basic "Basic Authentication";
        auth_basic_user_file "/etc/nginx/conf.d/.htpasswd";
  }
}
```

main.py
```python
import gradio as gr
from app import demo
app = FastAPI()
...
gr.mount_gradio_app(app, demo, path="/gradio")
```

app.py
```python
import gradio as gr
with gr.Blocks() as demo:
...

if __name__ == '__main__':
    demo.launch()
```

Temporary workaround
```
  location = /info {
    rewrite ^ /gradio/info;
  }
  location = /theme.css {
    rewrite ^ /gradio/theme.css;
  }
  location /assets/ {
    rewrite (.*) /gradio$1;
  }
  location /heartbeat/ {
    rewrite (.*) /gradio$1;
  }
  location /queue/ {
    rewrite (.*) /gradio$1;
  }
```

### Screenshot

![image](https://github.com/gradio-app/gradio/assets/116415100/ab5ea184-d085-4e8a-8a46-e0851259b691)

### System Info

```shell
gradio==4.27.0
fastapi==0.110.2
uvicorn==0.29.0

nginx version: nginx/1.24.0
built by gcc 11.2.0 (Ubuntu 11.2.0-19ubuntu1)
built with OpenSSL 3.0.2 15 Mar 2022
```

### Severity

I can work around it

[...]

PR #8258, opened 2024-05-10
Improve URL handling in JS Client

### LINKED ISSUE #4463: Have to reopen this issue
In the meantime, anyone needs the fix immediately can install gradio from main by following the instructions here: https://gradio.app/docs/main

_Originally posted by @abidlabs in https://github.com/gradio-app/gradio/issues/3472#issuecomment-1542707381_
            
firstly, thx for fixing the missing theme.css problem. It DO works in following two cases!
1、using it directly, say http://mydomain.com/myapp/XXX
2、reference it within iframe, say  <iframe src="http://mydomain.com/myapp/XXX">

but， it FAILED once again in following scenario!!
we using it with wss way, namely <gradio-app src="https://mydomain.com/myapp/XXX"></gradio-app>

the config is missing again!

image

hannahblair:
## Description

This PR ensures that we correctly handle URLs in various formats. E.g. We weren't handling URLs with paths (to an extent) or URLs with trailing slashes. This also resulted in issues around fetching `theme.css`, `/config` and `/info` in some cases. 

The following should now work (providing a gradio space is to be found at these URLs):

- hmb/hello-world
- http://localhost:7860
- http://localhost:7860/
- http://localhost:7860/app/gradio
- hmb-hello-world.hf.space
- hmb-hello-world.hf.space/ 

To test, run this python app and mount the gradio app on various paths:

```
import uvicorn
from fastapi import FastAPI

import gradio as gr


def predict(text):
    return text

with gr.Blocks() as block:
    text_in = gr.Textbox()
    text_out = gr.Textbox()
    btn = gr.Button("Predict")
    btn.click(predict, inputs=[text_in], outputs=[text_out], api_name="predict")

block.queue()
app = FastAPI()

gr.mount_gradio_app(app, block, path="")
# gr.mount_gradio_app(app, block, path="/")

if __name__ == "__main__":
    uvicorn.run(app, host="0.0.0.0", port=7860)
```

then in a separate app (or in serve an app as a static file with fastAPI), test you can connect to the app. 

```
const app = await Client.connect("hmb/test-args");
const app = await Client.connect("http://0.0.0.0:7860/app/gradio");
const app = await Client.connect("http://0.0.0.0:7860/");
// etc 
```

You can also test that it fixes #4463 by testing embedding with the custom path:

```
import uvicorn
from fastapi import FastAPI

import gradio as gr

app = FastAPI()

@app.get('/app')
def hello():
    return {"message":"hello"}

g_app = gr.Interface(
    lambda x: f"Hello {x}",
    "textbox",
    "textbox",
)

app = gr.mount_gradio_app(app, g_app, path="/app/gradio")

uvicorn.run(app,host="0.0.0.0",port=8080)
```

```
<script type="module" src="http://localhost:4321/gradio.js"></script>

<gradio-app src="http://0.0.0.0:8080/app/gradio"></gradio-app>
```

Closes: #4469
Closes: #4463
Closes: #3472
Closes: #8154 
Closes: #8199
Closes: #8073
(probably) Closes: #8129 

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/client/js/src/helpers/api_info.ts
+++ b/client/js/src/helpers/api_info.ts
@@ -67,12 +67,11 @@ export async function process_endpoint(
 
 export const join_urls = (...urls: string[]): string => {
 	try {
-		const parts = urls
-			.map((s) => s.replace(/^\/|\/$/g, ""))
-			.slice(1, urls.length)
-			.join("/");
-
-		return new URL(parts, urls[0]).toString();
+		return urls.reduce((base_url: string, part: string) => {
+			base_url = base_url.replace(/\/+$/, "");
+			part = part.replace(/^\/+/, "");
+			return new URL(part, base_url + "/").toString();
+		});
 	} catch (e) {
 		throw new Error(INVALID_URL_MSG);
 	}
--- a/client/js/src/test/api_info.test.ts
+++ b/client/js/src/test/api_info.test.ts
@@ -500,7 +500,15 @@ describe("join_urls", () => {
 			"http://localhost:7860/app/gradio"
 		);
 		expect(join_urls("http://localhost:7860/", "/app/", "/gradio/")).toBe(
-			"http://localhost:7860/app/gradio"
+			"http://localhost:7860/app/gradio/"
+		);
+
+		expect(join_urls("http://127.0.0.1:8000/app", "/config")).toBe(
+			"http://127.0.0.1:8000/app/config"
+		);
+
+		expect(join_urls("http://127.0.0.1:8000/app/gradio", "/config")).toBe(
+			"http://127.0.0.1:8000/app/gradio/config"
 		);
 	});
 	it("throws an error when the URLs are not valid", () => {


client/js/src/helpers/api_info.ts
```diff
@@ -1,9 +1,9 @@
 import type { Status } from "../types";
-import { QUEUE_FULL_MSG } from "../constants";
+import { HOST_URL, QUEUE_FULL_MSG } from "../constants";
 import type { ApiData, ApiInfo, Config, JsApiData } from "../types";
 import { determine_protocol } from "./init_helpers";
 
-export const RE_SPACE_NAME = /^[^\/]*\/[^\/]*$/;
+export const RE_SPACE_NAME = /^[a-zA-Z0-9_\-]+\/[a-zA-Z0-9_\-]+$/;
```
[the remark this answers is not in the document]
aliabid94: space names can also have a `.` in them, e.g. `https://huggingface.co/spaces/aliabid94/d.3`. I believe that covers all the possible names.

aliabid94 (approved):
Code LGTM!

client/js/src/helpers/api_info.ts
```diff
@@ -60,6 +62,12 @@ export async function process_endpoint(
 	};
 }
 
+export const join_urls = (...urls: string[]): string => {
+	return urls.reduce((base_url, part) => {
+		return `${base_url.replace(/\/+$/, "")}/${part.replace(/^\/+/, "")}`;
+	});
+};
+
```
pngwn, line 70: Does `new URL(relative, base)` not handle this?

pngwn (commented):
One small comment, otherwise looks great.

client/js/src/helpers/api_info.ts
pngwn: Shouldn't we just throw an error in that case? In node we need that base url so I feel like throwing early would be best. Unless I'm missing something.

pngwn: > we don't know what the base path would be at that point

Just to clarify, we will never know the base path in node right? If it is possible to use the URL class and just propagate that error, then I think that would simplest, if that isn't possible then this fn is fine.

abidlabs:
Nice!

autherous:
Thanks, Have been looking for this update.

Getting this error, when trying to install it 
`Unable to create package with no name`

Is this the right way to install it 
```
$ ./poetry add  https://gradio-builds.s3.amazonaws.com/f17855cf0d974ad4076ee8df3e8c3ce60eef0cee/gradio-4.31.5-py3-none-any.whl

Unable to create package with no name
```

PR #8260, opened 2024-05-10
Send ProcessCompleted message when job is cancelled

freddyaboulton:
## Description

If a job is cancelled, the server and the client will keep the `/queue/data` stream open forever.

Notice that in the first image, the multiple heartbeats gets sent after the job is cancelled (indicating the channel was kept open for at least 45 seconds post cancellation) while in the second message the process_completed and close_stream messages are sent after cancellation. This was tested on the `fake_diffusion` demo.

### Main
Screenshot 2024-05-10 at 3 36 47 PM

### This PR
Screenshot 2024-05-10 at 3 35 08 PM

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -490,6 +490,7 @@ class BlockFunction:
         scroll_to_output: bool = False,
         show_api: bool = True,
         renderable: Renderable | None = None,
+        is_cancel_function: bool = False,
     ):
         self.fn = fn
         self.inputs = inputs
@@ -526,6 +527,15 @@ class BlockFunction:
             or bool(self.every)
         )
         self.renderable = renderable
+
+        # We need to keep track of which events are cancel events
+        # in two places:
+        # 1. So that we can skip postprocessing for cancel events.
+        #   They return event_ids that have been cancelled but there
+        #   are no output components
+        # 2. So that we can place the ProcessCompletedMessage in the
+        #   event stream so that clients can close the stream when necessary
+        self.is_cancel_function = is_cancel_function
 
         self.spaces_auto_wrap()
 
@@ -1143,6 +1153,7 @@ class Blocks(BlockContext, BlocksEvents, metaclass=BlocksMeta):
         concurrency_id: str | None = None,
         show_api: bool = True,
         renderable: Renderable | None = None,
+        is_cancel_function: bool = False,
     ) -> tuple[BlockFunction, int]:
         """
         Adds an event to the component's dependencies.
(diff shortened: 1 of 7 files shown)


gradio/blocks.py
```diff
@@ -1656,10 +1669,18 @@ def validate_outputs(self, fn_index: int, predictions: Any | list[Any]):
...
         self, fn_index: int, predictions: list | dict, state: SessionState | None
-    ):
+    ) -> Any:
         state = state or SessionState(self)
         block_fn = self.fns[fn_index]
 
+        # If the function is a cancel function, 'predictions' are the ids of
+        # the event in the queue that has been cancelled. We need these
+        # so that the server can put the ProcessCompleted message in the event stream
+        # Cancel events have no output components, so we need to return early otherise the output
```
aliabid94, line 1679:
```suggestion
        # Cancel events have no output components, so we need to return early otherise the output
```

gradio/routes.py
```diff
@@ -738,6 +738,21 @@ async def predict(
                     fn_index_inferred=fn_index_inferred,
                     root_path=root_path,
                 )
+                if (  # noqa: SIM102
```
aliabid94, line 741: when I ran ruff, it didn't need this noqa statement? what does it do here?

aliabid94 (approved):
Reviewed and tested and LGTM, thanks freddy! I hadn't really explored the cancel logic in the backend before, and found the design a bit confusing. When we cancel, the frontend makes a call to both 
1. `/reset`  and 
2. as a regular event listener whose function is defined by `get_cancel_function`.
Unless I'm missing something, I think we should remove 2 entirely and handle all the logic needed via 1. Not only does it save a request, but some of the logic is done in a weird way in 1 because it's not being handled by gradio itself but some pre-defined logic. For example, using task names to store and track data for cancelling is weird, but necessary right now. But if we used the `/reset` endpoint to handle that logic, gradio can track events directly and I think it would be much simpler and cleaner.

freddyaboulton:
Thanks for the review @aliabid94 ! Let me merge this in and then we can refactor in subsequent PR.

gradio/blocks.py
```diff
@@ -1656,10 +1669,18 @@ def validate_outputs(self, fn_index: int, predictions: Any | list[Any]):
 
     async def postprocess_data(
         self, fn_index: int, predictions: list | dict, state: SessionState | None
-    ):
+    ) -> Any:
```
aliabid94, line 1672: I see why you set it to Any to avoid type hinting entirely with the list | dict, but would be great to fix in a refactor soon

gradio/routes.py
freddyaboulton: It raised an error because the two if statements can be combined into 1 but I think that would make the code harder to read.

PR #8262, opened 2024-05-11
Reorganize Guides in a more logical order

abidlabs:
The ordering of the [Guides currently](https://www.gradio.app/guides/quickstart) doesn't make sense, as we introduce `gr.Interface` in the Quickstart, then cover all sorts of [key features](https://www.gradio.app/guides/key-features) that are quite advanced, and then lots of [details related to sharing](https://www.gradio.app/guides/sharing-your-app), before jumping back to `gr.Interface`.

Now, I think the ordering makes more sense. These are the new sections

1. Quickstart
2. Interface
3. Additional Features (key features and sharing details moved here. I've also separated key features into individual guides since we should add more details to these guides -- will do in a separate pr -- and also for seo purposes)
4. Blocks
5. ChatInterface and Chatbots
6. Custom Components
7. Tabular DS and Plots
8. Clients and Lite
9. Other Tutorials (moved integrations to this section)

I've also fixed some typos, etc. I want to rework the guides some more but I didn't want to mix refactoring & other changes in a single PR.


aliabd (approved):
Nice! I agree this is much better. 

Can you add redirect routes for old links? Not a big deal but can be helpful https://github.com/gradio-app/gradio/blob/main/js/_website/src/routes/redirects.js

abidlabs:
Is it needed, I thought moving the guides to different sections does not affect the url, e.g. https://www.gradio.app/guides/the-interface-class

aliabd:
yeah you're right, i meant for the ones you removed, which i guess is just key-features

abidlabs:
right, will add

abidlabs:
Thanks @aliabd for the review!

Issue #7889, opened 2024-03-29
Better error messages with the Python Client


abidlabs:
If an invalid value is provided using the Python Client, then the error that is raised is not very informative. We should transmit the error from the server to the client to provide better error messages.

E.g.

```py
from gradio_client import Client, file

client = Client("abidlabs/whisper")

client.predict(
    audio="abc"
)
```

Here "abc" needs to be a file not a string, but the Client currently produces:

```
ValueError: None
```

SrijanSahaySrivastava:
Hey, I was able to reproduce this and I was thinking of doing a check of all args passed with `isinstance()` but idk what should I be checking them against

eg.:-
(*if we wanna check if argument is int*)
  for arg in args:
        if not isinstance(arg, int):  
            raise ValueError(f"Invalid argument: {arg}. All arguments must be integers.")

PR #8264, opened 2024-05-11
Make exceptions in the Client more specific

abidlabs:
Previously, If an exception was raised in the upstream app due to an invalid value provided by the Python Client, then the error that is raised is not very informative:

```
ValueError: None
```

My initial thought was to transmit the error from the server to the client to provide better error messages. However, we intentionally do not reveal exceptions to users of Gradio applications as they can reveal security vulnerabilities, unless `show_error=True`.

So this PR only prints the upstream exception if `show_error=True`, otherwise prints a generic error explaining how to print more detailed errors. For other types of errors that the Client can generate, e.g. when an invalid parameter is provided, it changes the errors to be `TypeError` instead of `ValueError` -- that's consistent with the error type that is raised when you call a function in Python with the wrong parameters.

Closes: https://github.com/gradio-app/gradio/issues/7889


.changeset/blue-frogs-fetch.md
```diff
@@ -0,0 +1,6 @@
+---
+"gradio": minor
```
freddyaboulton, line 2: nit: feels like a patch to me

freddyaboulton (approved):
Lovely @abidlabs !

PR #8292, opened 2024-05-14
Ensure JSON component outputs handled properly in postprocess

freddyaboulton:
## Description

* Makes sure json component outputs do not get moved to cache
* `EventData` instances must only reference data in the cache

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/gradio/data_classes.py
+++ b/gradio/data_classes.py
@@ -16,7 +16,7 @@ from gradio_client.utils import traverse
 from . import wasm_utils
 
 if not wasm_utils.IS_WASM or TYPE_CHECKING:
-    from pydantic import BaseModel, RootModel, ValidationError
+    from pydantic import BaseModel, JsonValue, RootModel, ValidationError
 else:
     # XXX: Currently Pyodide V2 is not available on Pyodide,
     # so we install V1 for the Wasm version.
@@ -24,6 +24,8 @@ else:
 
     from pydantic import BaseModel as BaseModelV1
     from pydantic import ValidationError, schema_of
+
+    JsonValue = Any
 
     # Map V2 method calls to V1 implementations.
     # Ref: https://docs.pydantic.dev/latest/migration/#changes-to-pydanticbasemodel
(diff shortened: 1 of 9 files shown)


gradio/data_classes.py
```diff
@@ -25,6 +25,8 @@
     from pydantic import BaseModel as BaseModelV1
     from pydantic import ValidationError, schema_of
 
+    JsonValue = Any
```
abidlabs, line 28: We can be more specific no? Something like
```suggestion
    JsonValue = Union[Dict, List]
```

gradio/processing_utils.py
```diff
@@ -341,6 +342,24 @@ def move_resource_to_block_cache(
     return block.move_resource_to_block_cache(url_or_file_path)
 
 
+def is_file_obj(data: dict | JsonData):
```
abidlabs, line 345: Is this method only added for type-checking purposes? Because `client_utils.is_file_obj` already checks to see if `data` is a `dict`?

test/components/test_json.py
```diff
@@ -95,5 +95,8 @@ def get_avg_age_per_gender(data):
     def test_postprocess_returns_json_serializable_value(self, value, expected):
         json_component = gr.JSON()
         postprocessed_value = json_component.postprocess(value)
-        assert postprocessed_value == expected
-        assert json.loads(json.dumps(postprocessed_value)) == expected
+        if postprocessed_value is None:
+            assert postprocessed_value is None
```
abidlabs, line 99: I don't get this

abidlabs (approved):
Thanks @freddyaboulton for taking care of this! Changes look good to me, left a couple of nits.

test/components/test_json.py
freddyaboulton: Brainfart - fixed!

gradio/processing_utils.py
freddyaboulton: Yea you're right another brain fart hehe

PR #8297, opened 2024-05-15
Add `state.change` listener

aliabid94:
Can now listen to changes to state variables via `state.change`. See demo/state_change for example use, which shows usage with render block as well:

```python
import gradio as gr

with gr.Blocks() as demo:

    with gr.Row():
        state_a = gr.State(0)
        btn_a = gr.Button("Increment A")
        value_a = gr.Number(label="A")
        btn_a.click(lambda x: x+1, state_a, state_a)
        state_a.change(lambda x: x, state_a, value_a)
    with gr.Row():
        state_b = gr.State(0)
        btn_b = gr.Button("Increment B")
        value_b = gr.Number(label="num")
        btn_b.click(lambda x: x+1, state_b, state_b)

        @gr.on(inputs=state_b, outputs=value_b)
        def identity(x):
            return x

    @gr.render(inputs=[state_a, state_b])
    def render(a, b):
        for _ in range(a):
            with gr.Row():
                for _ in range(b):
                    gr.Button(f"Button {a}, {b}")

if __name__ == "__main__":
    demo.launch()
```
Recording 2024-05-15 at 11 45 22


abidlabs (approved):
Tested and works great. LGTM @aliabid94

DoiiarX:
Thank you very much! I tried to solve the problem myself and quickly found that you did an excellent job.

Issue #2266, opened 2022-09-15
How does the login process works


leszekhanusz:
I am trying to connect to a gradio backend which uses authentication (`auth` argument with login and password is set in the `launch` method of my gradio interface).

It works fine when I use my browser to login but I'm now trying to login through another page using javascript and I don't know how to proceed. The [documentation about sharing your app](https://gradio.app/sharing_your_app/) explains how to activate this option but there is no explanation on how it works.

By looking at the network tab on Chrome, I was able to infer that you need to post formdata with `username` and `password` to the `/login` endpoint but if I do programmatically that I receive an html response.

Here is my javascript code right now:

```js
const login_response = await fetch("http://127.0.0.1:7860/login/", {
  method: "POST",
  body: formData,
});                      
                      
console.log(login_response);                                                               
                    
for (var pair of login_response.headers.entries()) { 
  console.log(pair[0]+ ': '+ pair[1]);                                                                        
}
// Only headers received are content-length and content-type (text/html)

const login_response_text = await login_response.text();
console.log(login_response_text);
// response here is an html page without any token or session hash
```

abidlabs:
Thanks for creating this issue @leszekhanusz and for providing the detailed JS code. @aliabid94 can you look at this, and is it possibly related to #1259?

aliabid94:
will take a look!

leszekhanusz:
In the meantime, as a workaround, I simply [added an access code](https://github.com/leszekhanusz/diffusion-ui-backend/releases/tag/v0.2.0) in the gradio inputs.

Fondson:
I was stuck on this problem too and I found this thread https://github.com/gradio-app/gradio/discussions/444#discussioncomment-1993832. It looks like the recommended flow is to login through the UI, call `/token` to get the user-specific token, and use that token in all API requests moving forward.

If this is true, then I wonder if this token ever expires? And if it does, then can we programmatically get another token?

chhatra-thapa:
Hello aliabid94 & @Fondson 
Is there any update regarding this issue?
I could not find any solution regarding this issue.

My gardio client javascript app has been stucked on login process too.

abidlabs:
Not at the moment sorry, we haven't had a chance to look into it. Contributions are welcome!

PR #8299, opened 2024-05-16
Allow JS Client to work with authenticated spaces 🍪

hannahblair:
## Description

This PR allows the connection to or duplication of spaces with authentication enabled. We get the cookie response from `/login` and if the response is `ok` then we parse and add the received access tokens to the the client's `this.fetch` function's `headers.Cookie` so all subsequent fetch requests will be authenticated. 

Note: This won't work with `EventSource` as I haven't added the logic for this, since we're moving away from it anyway. 

Closes: #2266

## 🎯 PRs Should Target Issues

## Tests


pngwn (approved):
Looks great! Thanks for taking care of this @hannahblair!

PR #8301, opened 2024-05-16
URL Safety Validation Function

mvlttt:
## Description

This pull request introduces a new function `is_safe_url()` in the `client_utils` module to validate the safety of URLs before processing them further. Here's an overview of the changes:

- Added `client_utils.is_safe_url()` function to check if a URL is safe for processing.
- The function uses URL parsing (`urlparse`), DNS lookup (`socket.gethostbyname`), and IP address validation (`ipaddress.ip_address`) to determine the safety of the URL.
- If the URL's scheme is not 'http' or 'https', or if the resolved IP address is private or multicast, the function returns `False`.
- Error handling is implemented to catch invalid URL formats, DNS resolution failures, and other potential issues.

This enhancement ensures that URLs are properly validated before any further actions are taken, improving the security and reliability of URL processing within the application.

### DIFF
--- a/client/python/gradio_client/utils.py
+++ b/client/python/gradio_client/utils.py
@@ -13,6 +13,9 @@ import shutil
 import tempfile
 import time
 import warnings
+import ipaddress
+import socket
+from urllib.parse import urlparse
 from dataclasses import dataclass, field
 from datetime import datetime
 from enum import Enum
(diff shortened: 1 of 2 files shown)


abidlabs:
Thanks @mvlttt I'm not totally sure if we want to introduce all these checks as they will likely add some latency any time URLs are being handled. Can you explain the motivation for this a little more?

abidlabs:
I'm mainly concerned about this DNS lookup:

```py
        # Resolve the domain to an IP address using DNS lookup
        resolve = socket.gethostbyname(domain)
```

Would the following simpler approach work:
1. Check to see if its a url string or an ip address
2. If its a url string, it should not start with localhost
3. If its an ip address, it should not be ip.is_private or ip.is_multicast

mvlttt:
The reason I do DNS lookup is to prevent the DNS rebinding vulnerability. I'm not sure how to optimize this specifically, but I'm considering implementing a control to only perform DNS lookup if the domain is different from the IP.

The solutions you suggest have many bypass methods. I'm adding a few references for you.

https://payatu.com/blog/dns-rebinding/
https://cheatsheetseries.owasp.org/cheatsheets/Server_Side_Request_Forgery_Prevention_Cheat_Sheet.html
https://github.com/swisskyrepo/PayloadsAllTheThings/blob/master/Server%20Side%20Request%20Forgery/README.md

missionfloyd:
This is going to break LAN and localhost addresses. Maybe it should be optional.

Or perhaps it should only apply to domain names, not IPs.

one-lithe-rune:
> This is going to break LAN and localhost addresses.

It seems to me a fair few downstream projects of gradio are essentially single user applications where the browser is connecting to an instance running on your localhost or another machine on your LAN - that machine being the one has the AI processing capability. At least that's I've always encountered outside of the actual huggingface site, including the one I've contributed to in the past or (less significant, because only I use them) the ones I've made myself.

I have no idea whether this particular change actually *does* break that, but if it does that would be concerning.

> Maybe it should be optional.

Ability to set a whitelist of ips/domains?

abidlabs:
One of the gradio maintainers, @aliabid94, started working on this here: https://github.com/gradio-app/gradio/pull/8978. I'll close this PR for now

aliabid94:
I believe I implemented the same safety checks in https://github.com/gradio-app/gradio/pull/8978 if you'd like to review @mvlttt !

PR #8308, opened 2024-05-16
Update custom component installation requirements

shimizust:
## Description

- Added pip 21.3+ requirement to be able to build from pyproject.toml otherwise get errors of this type:
```
 👷 Installing python... (/my/path/.venv/bin/pip3 install -e testcomponent[dev])                                                                           │
│ 🟥 Python installation failed                                                                                                                                                                                  │
│ ERROR: File "setup.py" or "setup.cfg" not found. Directory cannot be installed in editable mode: /my/path/gradio-components/testcomponent              │
│ (A "pyproject.toml" file was found, but editable mode currently requires a setuptools-based build.) 
```


abidlabs:
Thank you @shimizust for adding this clarification

PR #8311, opened 2024-05-16
Cleanup markdown styling

aliabid94:
Couple issues wth markdown styling, also lots of repeated css across Chatbot.svelte, MarkdownCode.svelte, and typography.css. I put repeated css in typography.css, removed from other sources. Also fixed:
- nested list indentation didn't work in gr.Markdown
- Text in a list that was longer than a line caused a line break at the start of the line, fixed now.
- chatbot list indentation was too great in gr.Chatbot
- Margin for h1, and for last element not set correctly

BEFORE
Screenshot 2024-05-16 at 3 56 34 PM

AFTER
Screenshot 2024-05-16 at 3 56 55 PM

Demo for above:

```python
import gradio as gr

def welcome(name):
    return f"Welcome to Gradio, {name}!"

with gr.Blocks() as demo:
    md =     """Absolutely, here are 10 tips for maintaining your health from [Harvard Health](https://www.health.harvard.edu/staying-healthy/10-tips-for-better-health):
- Eat a variety of foods including fruits, vegetables, lean proteins, whole grains, and healthy fats. Avoid excessive sugar, salt, and processed foods.
    - nested a
    - nested b
    - lorem ipsum dolor sit amet consectetur adipiscing elit sed do eiusmod tempor incididunt ut labore et dolore magna aliqua

- Aim for at least 150 minutes of moderate aerobic activity or 75 minutes of vigorous activity each week, along with strength training exercises twice a week.

# Table Example
| Header 1 | Header 2 |
|----------|----------|
| Data 1   | Data 2   |
| Data 3   | Data 4   |

- Get at least 7 hours of sleep each night and practice good sleep hygiene.

$$y = mx + b$$

and finally some code:

~~~python
def welcome(name):
    return f"Welcome to Gradio, {name}!"
~~~

    """
    with gr.Row():
        gr.Markdown(md)
        gr.Chatbot([[md, md]], height="100%")

if __name__ == "__main__":
    demo.launch()
```


pngwn (approved):
This is awesome. Thanks you @aliabid94!!

PR #8323, opened 2024-05-17
Fix zero gpu auth ignoring hf_token

Saghen:
## Description

When sending requests to a ZeroGPU endpoint via the JS client on node, the client would attempt to index `window` while getting credentials, causing an error. The logic has been changed to prefer the `hf_token`, falling back to the `post_message` if the `window` is available, and finally falling back to `null`.

I have not included an issue due to the size of the PR, as per the PR template, but will add if needed!

### DIFF
--- a/client/js/src/utils/submit.ts
+++ b/client/js/src/utils/submit.ts
@@ -496,8 +496,12 @@ export function submit(
 					const origin = hostname.includes(".dev.")
 						? `https://moon-${hostname.split(".")[1]}.${hfhubdev}`
 						: `https://huggingface.co`;
+
+					const is_iframe =
+						typeof window !== "undefined" && window.parent != window;
+					const is_zerogpu_space = dependency.zerogpu && config.space_id;
 					const zerogpu_auth_promise =
-						dependency.zerogpu && window.parent != window && config.space_id
+						is_iframe && is_zerogpu_space
 							? post_message<Headers>("zerogpu-headers", origin)
 							: Promise.resolve(null);
 					const post_data_promise = zerogpu_auth_promise.then((headers) => {


abidlabs:
> @abidlabs do you confirm that the JS client should target NodeJS ? Shouldn't the CI somehow check the compatibility with Node then ?

Let me tag @hannahblair who can review this PR better than me

Saghen:
> One point though: isn't the `Authorization` header already set elsewhere in the client ? `zero_gpu_auth` only aims to add additional headers specifically supported by the ZeroGPU Spaces infra.

Yep, thanks! Removed that logic

client/js/src/utils/submit.ts
```diff
@@ -497,7 +497,10 @@ export function submit(
 						? `https://moon-${hostname.split(".")[1]}.${hfhubdev}`
 						: `https://huggingface.co`;
 					const zerogpu_auth_promise =
-						dependency.zerogpu && window.parent != window && config.space_id
+						dependency.zerogpu &&
+						typeof window !== "undefined" &&
+						window.parent != window &&
+						config.space_id
 							? post_message<Headers>("zerogpu-headers", origin)
 							: Promise.resolve(null);
```
pngwn, line 505: This is getting pretty verbose now. Could we break this out into an `if` block instead?

pngwn (approved):
Beautiful, thank you @Saghen!

pngwn:
> @abidlabs do you confirm that the JS client should target NodeJS ? 

The client should (and will eventually) work on all platforms, not just node + browser but also cf workers, services workers, deno, etc.

> Shouldn't the CI somehow check the compatibility with Node then ?

Yes, and it kind of does but we don't have any tests for zero gpu atm.

Issue #8268, opened 2024-05-13
.select Event has abnormal behavior with gr.File


marcoquin:
### Describe the bug

First of all, thanks for the tremendous work on gradio !

I try using gradio.File with `file_count="multiple",`. I also use in my code a `file.select` event. When I delete a file, the event is triggered except when the deleted file is the last in the gradio.File box. 

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

with gr.Blocks() as io:
    title = gr.Markdown("1")
    file = gr.File(
        label="Upload Files",
        file_count="multiple",
        )

    file.select(
        fn=(lambda file: print("Event triggered!")),
        inputs=[file],
        outputs=[],
    )

io.launch()
```
I use the command `uvicorn run:app` (my python file is called run.py) I made a simple event that prints a message in the console, so you need to check the logs to see if the event is triggered or not.

### Screenshot

![Capture d'écran 2024-05-13 100421](https://github.com/gradio-app/gradio/assets/166842689/04bdecd8-70f8-45b5-bc0e-0731e2f52fbc)
Here is the precision of the precise file that does not trigger the select event (it is always the last)

### System Info

```shell
I tried on gradio 4.28.0 too. Result of pip freeze : 

aiofiles==23.2.1
altair==5.3.0
annotated-types==0.6.0
anyio==4.3.0
attrs==23.2.0
certifi==2024.2.2
charset-normalizer==3.3.2
click==8.1.7
colorama==0.4.6
contourpy==1.2.1
cycler==0.12.1
fastapi==0.110.1
ffmpy==0.3.2
filelock==3.13.4
fonttools==4.51.0
fsspec==2024.3.1
gradio==4.29.0
gradio_client==0.16.1
h11==0.14.0
httpcore==1.0.5
httpx==0.27.0
huggingface-hub==0.22.2
idna==3.7
importlib_resources==6.4.0
Jinja2==3.1.3
jsonschema==4.21.1
jsonschema-specifications==2023.12.1
kiwisolver==1.4.5
markdown-it-py==3.0.0
MarkupSafe==2.1.5
matplotlib==3.8.4
mdurl==0.1.2
numpy==1.26.4
orjson==3.10.1
packaging==24.0
pandas==2.2.2
pillow==10.3.0
pydantic==2.7.0
pydantic_core==2.18.1
pydub==0.25.1
Pygments==2.17.2
pyparsing==3.1.2
[...]

PR #8334, opened 2024-05-20
fix: prevent triggering gr.File.select on delete

gtm-nayan:
## Description
the select event shouldn't be triggering on any of the files, but it did because of some dom-bubbling and svelte lifecycle

just adding the key to the each fixes that issue by ensuring the element is removed,
added additional checks just to be sure and so that e.g. clicking the download like doesn't trigger select

Closes: #8268

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/light-rats-exist.md
+++ b/.changeset/light-rats-exist.md
@@ -1,0 +1,6 @@
+---
+"@gradio/file": patch
+"gradio": patch
+---
+
+fix:fix: prevent triggering gr.File.select on delete
--- a/js/file/shared/FilePreview.svelte
+++ b/js/file/shared/FilePreview.svelte
@@ -45,15 +45,20 @@
 >
 	<table class="file-preview">
 		<tbody>
-			{#each normalized_files as file, i}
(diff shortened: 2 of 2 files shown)


pngwn (approved):
Tested this and confirmed that it now works as expected! Thanks @gtm-nayan!

gtm-nayan:
forgot a `.orig_name` [here](https://github.com/gradio-app/gradio/pull/8334/files#diff-98be64edb102f9cff4313f1a346aeacdee035209a1c6629ebf70ed0033928fcbR44)

@pngwn possible to fix that in main directly?

pngwn:
@gtm-nayan Sadly not, can make another PR.

marcoquin:
Hello guys, thanks for the fix and the work on Gradio. However, I used the `.select` event on my app to detect when a file was deleted from a `gr.File` box. The `.change` event is not precise enough, and `.clear` is only triggered when you delete all the files at once. It would be nice to have an event (for example a `.delete`) that is only triggered when you delete a single file from a multi-file box, as we can't use the `.select` event anymore.

Issue #8309, opened 2024-05-16
Unable to run custom component dev server based on DataFrame template


shimizust:
### Describe the bug

I'm trying to create a custom component that extends the existing DataFrame component. While I am able to spin up dev servers for custom components based on `SimpleTextbox` and `Button`, for example, it just hangs when I try using `DataFrame` as template. I haven't made any modifications to the code at this point.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```
gradio cc create DataFrameV2 --template DataFrame
```
This succeeds, but when I go to start up the dev server it just hangs at the loading screen. I see the following errors in the browser dev console (see screenshot):
```
gradio cc dev
```

I also tried explicitly specifying my gradio-path and python-path, but encounter the same issue.

### Screenshot

<img width="589" alt="image" src="https://github.com/gradio-app/gradio/assets/119249469/4ee28999-d39b-4a09-bb0d-ca1b428c2ea9">

<img width="877" alt="image" src="https://github.com/gradio-app/gradio/assets/119249469/3bc9e3ae-ead4-4d06-bac8-e46d0a93c75b">

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Darwin
gradio version: 4.31.3
gradio_client version: 0.16.3

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.0.1
fastapi: 0.109.1
ffmpy: 0.3.0
gradio-client==0.16.3 is not installed.
httpx: 0.26.0
huggingface-hub: 0.20.3
importlib-resources: 6.1.1
jinja2: 3.1.3
markupsafe: 2.1.5
matplotlib: 3.7.2
numpy: 1.25.2
orjson: 3.6.8
packaging: 23.2
pandas: 2.2.0
pillow: 10.2.0
pydantic: 2.5.3
pydub: 0.25.1
python-multipart: 0.0.9
pyyaml: 6.0.1
ruff: 0.2.2
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.12.3
typing-extensions: 4.9.0
urllib3: 2.2.0
uvicorn: 0.27.0.post1
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2023.12.2
httpx: 0.26.0
huggingface-hub: 0.20.3
packaging: 23.2
[...]

PR #8339, opened 2024-05-20
Fix Chatbot, Dataframe, Markdown custom components

freddyaboulton:
## Description

Closes: #8309

You can verify this works by templating off of DataFrame or Chatbot and making the following modifications to `@gradio/markdown` in node_modules.

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/tall-suits-train.md
+++ b/.changeset/tall-suits-train.md
@@ -1,0 +1,6 @@
+---
+"@gradio/markdown": patch
+"gradio": patch
+---
+
+fix:Fix Chatbot, Dataframe, Markdown custom components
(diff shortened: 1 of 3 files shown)


abidlabs:
Do you think it would make sense to add a test to ensure all of our core components can be used as templates for custom components?

pngwn:
I think to test this we would have to run dev or build after generating them and this process is quite slow, so this would add a substantial amount of time to CI. Open to any suggestions to speed it up though, it would be nice to have some kind of check.

freddyaboulton:
Yes a CI test would make sense. I tried to add one in #6715 but it would not work in CI. @pngwn is right though that it would be really slow to test all core components. Maybe we can create a giant custom component directory and launch a kitchen sink style demo. 

@pngwn let me know if this particular change looks good though.

pngwn (approved):
Look great! Thanks @freddyaboulton!

freddyaboulton:
Thank you @pngwn 🙏

PR #8341, opened 2024-05-21
add missing orig_name (follow up to #8334)

gtm-nayan:
## Description

See https://github.com/gradio-app/gradio/pull/8334#issuecomment-2120464108


pngwn (approved):
Thanks @gtm-nayan!

Issue #8354, opened 2024-05-22
Add delete or remove event to `gr.File`


pngwn:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  

We recently fixed an issue where removing a file from a file component would trigger a select event but some people require an event to trigger when a file is removed. 
 
**Describe the solution you'd like**  
A dedicated event, similar to clear, but just for deleting an individual file.

ngaudemet:
thanks, it was missing ! it is not yet on gradio 31.5 but will be in a next version, right ?

pngwn:
I'll make sure we get this out in the next release.

ngaudemet:
Hi @pngwn 
thanks for your help
I still can't upgrade to newer gradio version (I'm still on 4.29.0) because it seems the introduction of the delete event for File components destroyed the select event for File components. But it is useful as well. Thus I open a new issue.
****

PR #8361, opened 2024-05-22
Add file deletion event to gr.File component

pngwn:
Related to #8354

Implements a new event for file deletion in `gr.File` components and updates event handling to support this feature.

- **Adds a new event type** for file deletion within `gr.File` components in `gradio/events.py`. This allows for a dedicated event to trigger when an individual file is removed, addressing the feature request.
- **Implements logic** in `gradio/interface.py` to trigger the new file deletion event when a file is removed from a `gr.File` component. This ensures the backend is notified of the file removal action.
- **Introduces JavaScript logic** in `gradio/static/js/client.js` to emit an event to the backend when a file is removed from a `gr.File` component. This front-end handling complements the backend changes to provide full support for the new file deletion event.

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/gradio-app/gradio/issues/8354?shareId=78f8b63b-b9a1-40db-a454-8ab803f11912).

### DIFF
--- a/.changeset/red-beers-move.md
+++ b/.changeset/red-beers-move.md
@@ -1,0 +1,5 @@
+---
+"gradio": minor
+---
+
+feat:Add file deletion event to gr.File component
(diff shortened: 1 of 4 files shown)


pngwn:
So long.

PR #8363, opened 2024-05-22
Add allow-credentials to preflight header

abidlabs:
I've tested and this fixes the issue that @hannahblair I identified. I don't think this introduces any security risk based on my mental model (as long as the origins are trusted, we do indeed want to pass on cookies and authorization headers).

### DIFF
--- a/.changeset/dark-streets-visit.md
+++ b/.changeset/dark-streets-visit.md
@@ -1,5 +1,5 @@
 ---
-"gradio": minor
+"gradio": patch
 ---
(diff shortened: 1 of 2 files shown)


pngwn (approved):
Looks good to me. We are already trusting all origins so this isn't significantly more risky considering this is just for basic auth anyway. 

Thanks @abidlabs!

abidlabs:
Thanks @hannahblair @pngwn!

Issue #8375, opened 2024-05-24
Right-handed coordinate system in 3D viewer


dylanebert:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
The coordinate system in babylon is currently left-handed, leading to misleading or unexpected results for most 3D researchers.

**Describe the solution you'd like**  
Use right-handed coordinate system in the Canvas3D.

PR #8376, opened 2024-05-24
Model3D right-handed coordinate system

dylanebert:
## Description

Revise Canvas3D viewer to use a right-handed coordinate system, leading to fewer confusing or unexpected results for researchers.

Closes: #8375 

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/chubby-rings-beg.md
+++ b/.changeset/chubby-rings-beg.md
@@ -1,0 +1,6 @@
+---
+"@gradio/model3d": minor
+"gradio": minor
+---
+
+feat:Model3D right-handed coordinate system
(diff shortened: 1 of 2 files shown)


dylanebert:
For more context, this isn't a bug, but many researchers perceive it as a bug, such as in [InstantMesh](https://huggingface.co/spaces/TencentARC/InstantMesh).

Another option is to allow changing the coordinate system in the viewer, but this may also be confusing. For simplicity, a right-handed coordinate system should lead to the least confusion for users :)

EmVSW5AW8AAoDD9

Some additional context here: [forum](https://forum.babylonjs.com/t/obj-export-mirrored/10835)

dawoodkhan82:
Quick question about InstantMesh: 

> Note: Downloaded .obj model will be flipped. Export .glb instead or manually flip it before usage.

Will this PR fix this issue in their space? Why does it only affect the obj and glb files differently if so?

dawoodkhan82 (approved):
Besides that one question, LGTM! Thanks @dylanebert

Issue #6841, opened 2023-12-19
Events created with `every` are put on the queue


freddyaboulton:
### Describe the bug

If you create an event with `every` it is put on the queue. This is not what used to happen in 3.x - see the code below taken from 3.x routes.py

```python
            # Continuous events are not put in the queue  so that they do not
            # occupy the queue's resource as they are expected to run forever
            if blocks.dependencies[event.fn_index].get("every", 0):
                await cancel_tasks({f"{event.session_hash}_{event.fn_index}"})
                await blocks._queue.reset_iterators(event.session_hash, event.fn_index)
                blocks._queue.continuous_tasks.append(event)
                task = run_coro_in_background(
                    blocks._queue.process_events, [event], False
                )
                set_task_name(task, event.session_hash, event.fn_index, batch=False)
            else:
                rank = blocks._queue.push(event)
```

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

Run the `live_dashboard` demo but set the queue max_size=1. This event has multiple load events created with `every` so you will get a Queue full error.

### Screenshot

<img width="1482" alt="image" src="https://github.com/gradio-app/gradio/assets/41651716/a2da1ac4-b99f-4095-9ee3-1611e897afaf">

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Darwin
gradio version: 4.10.0
gradio_client version: 0.7.3

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.2.0
fastapi: 0.105.0
ffmpy: 0.3.1
gradio-client==0.7.3 is not installed.
httpx: 0.25.2
huggingface-hub: 0.19.4
importlib-resources: 6.1.1
jinja2: 3.1.2
markupsafe: 2.1.3
matplotlib: 3.8.2
numpy: 1.26.2
orjson: 3.9.10
packaging: 23.2
pandas: 2.1.4
pillow: 10.1.0
pydantic: 2.5.2
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0.1
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.9.0
typing-extensions: 4.9.0
uvicorn: 0.24.0.post1
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2023.12.2
httpx: 0.25.2
[...]

PR #8382, opened 2024-05-27
fixes to `gr.load()` with `load` events and `every` events

abidlabs:
Fixes: https://github.com/gradio-app/gradio/issues/6072

We're doing some weird stuff with `every` events. Will try to fix as part of this PR. Related: https://github.com/gradio-app/gradio/issues/6841


abidlabs:
Will come back to this later

PR #8385, opened 2024-05-27
Fix bug in reload mode equality check. Better equality conversion for state variables

### LINKED ISSUE #8384: gr.State JSON serializable, when there's a reload
### Describe the bug

Usually I use in my gradio demos  [gr.State()](https://www.gradio.app/guides/state-in-blocks) to define a class with methods and attributes, that can be used at different steps or functionalities within the application. For example, I can define a MainState outside the gradio main Blocks:

```
class MainState:
    """ Class with the main state. Needed to be able to have a persistent state between interactions, as well as unique state for each user.
    """
    def __init__(self):
        self.index_name = "Leg"
```
        
and then instantiate it:

```
with gr.Blocks(theme=theme, css = css_custom) as demo:
    main_state = gr.State(MainState())
```

The issue is that I'm having problems when working in development and there's a reload of the application. I'm getting the error:

`TypeError: Object of type MainState is not JSON serializable
` 

It works at first, but if I edit my source code, save it and gradio reloads it, then it tries to serialize the state and the error occurs. 

### Reproduction

```python
import gradio as gr

class MainState:
    """ Class with the main state. Needed to be able to have a persistent state between interactions, as well as unique state for each user.
    """

    def __init__(self):
        self.index_name = "Leg"


################
## GRADIO APP ##
################

with gr.Blocks() as demo:

    main_state = gr.State(MainState())

    def set_index_name(state,index_name):
        state.index_name = index_name
        print(state.index_name)

    with gr.Row():
        with gr.Column(scale = 30):
            
            message = gr.Textbox(show_label = False,lines =3)
            btn = gr.Button(value="Procurar",size="sm")

            choice = gr.Radio(choices=['Leg', 'IPV'], label="Legislação ou IPVs", value = "Leg")
            choice.change(fn=set_index_name, inputs=[main_state,choice], outputs=None)


demo.queue()
demo.launch()

```

### Screenshot

_No response_

### Logs

_No response_

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Darwin
gradio version: 4.31.5
gradio_client version: 0.16.4

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.2.0
fastapi: 0.108.0
ffmpy: 0.3.2
gradio-client==0.16.4 is not installed.
httpx: 0.27.0
huggingface-hub: 0.21.2
importlib-resources: 6.1.2
jinja2: 3.1.3
markupsafe: 2.1.5
matplotlib: 3.8.3
numpy: 1.26.4
orjson: 3.9.15
packaging: 23.2
pandas: 2.2.1
pillow: 10.2.0
pydantic: 2.6.3
pydub: 0.25.1
python-multipart: 0.0.9
pyyaml: 6.0.1
ruff: 0.2.2
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.12.3
typing-extensions: 4.10.0
urllib3: 2.2.1
uvicorn: 0.25.0
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2024.2.0
httpx: 0.27.0
huggingface-hub: 0.21.2
packaging: 23.2
typing-extensions: 4.10.0
websockets: 11.0.3
```

### Severity

I can work around it

freddyaboulton:
## Description

Closes #8384 

Fixes the linked issue specific reload mode. But the equality check for state, basically `current_state != old_state` would throw an error for some types like numpy/dataframe. So proposing a function to handle that case.

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -1726,7 +1726,9 @@ Received outputs:
 
             if block.stateful:
                 if not utils.is_update(predictions[i]):
-                    if block._id not in state or state[block._id] != predictions[i]:
+                    if block._id not in state or not utils.gradio_deep_equal(
+                        state[block._id], predictions[i]
+                    ):
                         changed_state_ids.append(block._id)
                     state[block._id] = predictions[i]
                 output.append(None)
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -1728,7 +1728,7 @@ Received outputs:
                 if not utils.is_update(predictions[i]):
                     has_change_event = False
                     for dep in state.blocks_config.fns.values():
-                        if block._id in [t[0] for t in dep.targets]:
+                        if block._id in [t[0] for t in dep.targets if t[1] == "change"]:
                             has_change_event = True
                             break
                     if has_change_event and (
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -1726,7 +1726,7 @@ Received outputs:
 
             if block.stateful:
                 if not utils.is_update(predictions[i]):
-                    if block._id not in state or not utils.gradio_deep_equal(
+                    if block._id not in state or not utils.deep_equal(
                         state[block._id], predictions[i]
                     ):
                         changed_state_ids.append(block._id)
--- a/gradio/blocks.py
+++ b/gradio/blocks.py
@@ -1726,8 +1726,14 @@ Received outputs:
 
             if block.stateful:
                 if not utils.is_update(predictions[i]):
-                    if block._id not in state or not utils.deep_equal(
-                        state[block._id], predictions[i]
+                    has_change_event = False
+                    for dep in state.blocks_config.fns.values():
+                        if block._id in [t[0] for t in dep.targets]:
+                            has_change_event = True
+                            break
+                    if has_change_event and (
+                        block._id not in state
+                        or not utils.deep_equal(state[block._id], predictions[i])
                     ):
                         changed_state_ids.append(block._id)
                     state[block._id] = predictions[i]
--- a/gradio/utils.py
+++ b/gradio/utils.py
@@ -291,7 +291,7 @@ def watchfn(reloader: SourceFileReloader):
         time.sleep(0.05)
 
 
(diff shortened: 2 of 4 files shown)


gradio/utils.py
```diff
@@ -290,6 +291,29 @@ def iter_py_files() -> Iterator[Path]:
         time.sleep(0.05)
 
 
+def gradio_deep_equal(a: Any, b: Any) -> bool:
```
abidlabs, line 294: Suggest dropping the `gradio`:
```suggestion
def deep_equal(a: Any, b: Any) -> bool:
```

aliabid94:
Oh I'm hesitant about this one. The reason why reactive code blocks in svelte don't automatically do deep equality checks with lists/objects is that it could be extremely expensive accidentally.
The problem in this PR is that we are going to be doing deep checks on any state change at all, even if there's no change event listener attached to the state object. 

> let developers define their own state equality check function

I think it is very likely that developers will see their performance degrade and not have any idea why this is happening.

I don't mind the deep check in reload mode, it's okay if we add lag on reload, but I'd feel more comfortable using deep checks for state changes only if we also only do state change checks if there is a change event listener for the state variable. That way the deep check is only happening when explicitly requested to.

gradio/blocks.py
```diff
@@ -1726,8 +1726,14 @@ async def postprocess_data(
 
             if block.stateful:
                 if not utils.is_update(predictions[i]):
-                    if block._id not in state or not utils.deep_equal(
-                        state[block._id], predictions[i]
+                    has_change_event = False
```
freddyaboulton, line 1729: @aliabid94 I implemented your suggestion to only compute the state diff of there is a change event defined for that state

aliabid94: thanks, looks good!

aliabid94 (approved):
LGTM!

freddyaboulton:
Thank you @aliabid94 @abidlabs for the reviews!!

PR #8398, opened 2024-05-28
Improve rendering

aliabid94:
Several improvements to gr.render, including:

1. Animation to show rendering
Recording 2024-05-28 at 08 49 22

2. Memory cleanup in frontend, removing elements from previous render
3. Allow rendering in a row 
4. Respond to last trigger, via trigger_mode="always_last"


abidlabs (approved):
Tested and this fixed all the issues that I had found previously. LGTM

gradio/blocks.py
```diff
@@ -233,13 +236,19 @@ def get_config(self):
         for parameter in signature.parameters.values():
             if hasattr(self, parameter.name):
                 value = getattr(self, parameter.name)
-                config[parameter.name] = utils.convert_to_dict_if_dataclass(value)
+                if isinstance(value, Callable):
+                    continue
+                if dataclasses.is_dataclass(value):
```
freddyaboulton, line 241: I think the better fix is to use `default=str` in the `orjson.dumps` call in `process_msg`. There could be non-serializable values beyond functions/dataclasses set in state variables/custom components.

freddyaboulton:
Rendering animation looks great. Two comments beyond what I left on slack:

* I think it should be possible to disable the progress animation for render. It can be distracting for quick renders. Especially if render is triggered based on small changes to a state variable.

too_much_progress

* I think it should be possible to not queue render events. If it's a quick UI update, I don't see the benefit of queueing.

pngwn (approved):
I've only tested a little but everything mentioned in the description seems to work well and the frontend changes look good to me! Thanks for fixing these issues @aliabid94!

pngwn:
Regarding rapid updates, a common approach is to add a threshold of a few hundred MS before starting any animations, that could work here (but is also something we could do in a follow up)

gradio/blocks.py
aliabid94: Thanks, fixed

aliabid94:
> I think it should be possible to disable the progress animation for render. It can be distracting for quick renders. Especially if render is triggered based on small changes to a state variable.

Okay I made this change for now: Previously, the animation would be alternating from 100% opacity to 50% and back to 100%, over 2s infinitely. Now it starts at 0%, goes to 100% over 1s, and then 100% opacity to 50% and back to 100%, over 2s infinitely. Because it now starts at 0%, quick renders don't show any visual change.

> I think it should be possible to not queue render events. If it's a quick UI update, I don't see the benefit of queueing.

Done.

Issue #8399, opened 2024-05-28
Additional parameters in ChatInterface have incorrect names


freddyaboulton:
### Describe the bug

If you launch a ChatInterface with `additional_inputs`, the view api page will show an incorrect name for the additional inputs.

For example, the `system_prompt` is called `request` here:

<img width="896" alt="image" src="https://github.com/gradio-app/gradio/assets/41651716/37315706-7622-4480-99a7-229e78845ff5">

You can repro this by running `demo/chatinterface_system_prompt`

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr
import time


def echo(message, history, system_prompt, tokens):
    response = f"System prompt: {system_prompt}\n Message: {message}."
    for i in range(min(len(response), int(tokens))):
        time.sleep(0.05)
        yield response[: i + 1]


demo = gr.ChatInterface(
    echo,
    additional_inputs=[
        gr.Textbox("You are helpful AI.", label="System Prompt"),
        gr.Slider(10, 100),
    ],
)

if __name__ == "__main__":
    demo.queue().launch()
```

and go to view api page

### System Info

```shell
latest main
```

### Severity

I can work around it

abidlabs:
This actually happens anytime a `gr.Request` is present in the arguments of a function and it is not the last argument, e.g.

```py
import gradio as gr
import time


def echo(message, request: gr.Request, num):
    return message

demo = gr.Interface(
    echo,
    ["textbox", "slider"],
    ["textbox"],
)

demo.launch()
```

PR #8400, opened 2024-05-28
Handle special arguments when extracting parameter names for view API page

abidlabs:
Closes: https://github.com/gradio-app/gradio/issues/8399

Test with:

```py
import gradio as gr
import time


def echo(message, request: gr.Request, num):
    return message

demo = gr.Interface(
    echo,
    ["textbox", "slider"],
    ["textbox"],
)

demo.launch()
```

For the case of `gr.ChatInterface`, we can't get the original parameter names because we construct a wrapper function over the user's defined function, so in such cases, the parameter names are generated automatically:

image


freddyaboulton:
Can we use something like functools.wraps in the chat_interface case to preserve the original signature?

Also, I was wondering why we need to construct the special args within the chat_interface function. Shouldn't that be handled by the regular request life-cycle in Blocks.

abidlabs:
> Can we use something like functools.wraps in the chat_interface case to preserve the original signature?

I think so yes I'll take a look

> Also, I was wondering why we need to construct the special args within the chat_interface function. Shouldn't that be handled by the regular request life-cycle in Blocks.

Sorry what are you referring to by this?

freddyaboulton:
> Sorry what are you referring to by this?

If we can use something like `functools.wraps` to preserve the original signature, then I don't think we need to call `special_args` within `submit_fn` because `gr.Blocks` will know to pass the request as an input if it sees the type hint.

abidlabs:
Thanks @freddyaboulton for the suggestion. Refactored to use `functools.wraps`, so now we handle `gr.ChatInterface` with additional parameters correctly as well. For the example in the original issue, we now have:

image

abidlabs:
Indeed. Thanks @freddyaboulton for the reviews and suggestions.

PR #8403, opened 2024-05-29
Editable Docs

aliabd:
Converts the docs to a template format. 
- To make changes to an existing page, or add a new one, you can just go to the `js/_website/src/lib/templates/`directory and make the changes. All the jsons are still there to use and build the same way as before. Just run `pnpm dev` again and the changes will be there.
- When a new version is released, the entire template dir gets uploaded to aws, and every time the website is built it first loads the template_{version} directory from aws and then reads from it for /docs (and reads from the regular one for /main/docs)
- Refactored a lot of the code and removed 3.50.2 docs 

The build is failing because the current version bucket (4.31.5) on aws doesn't have the right jsons / templates. I can't manually change the format there because it will break the website. But i tested everything with a fake version. 

If you want to test version vs main with the fake version just change /gradio/package.json to have 	"version": "TEST.VERSION" instead of 4.31.5 and run `pnpm dev` 

When we want to merge this, we should do it and release a new version. Alternatively, I can edit the jsons on 4.31.5 and then merge and rebuild the website. 

I will write a proper readme / contributing guide for how to make changes to the website and build locally


pngwn (approved):
Thanks for this @aliabd! tested and working great, and we've been through the PR together a few times now!

PR #8408, opened 2024-05-29
Connect heartbeat if state created in render. Also fix config cleanup bug #8407

freddyaboulton:
## Description

Noticed a bug where the heartbeat was not established if a state variable was created in a render.
In the process of fixing that, I noticed that components were not being removed from the old config. The rendered_in key is in props, not the top level component config.

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/easy-trains-obey.md
+++ b/.changeset/easy-trains-obey.md
@@ -1,6 +1,6 @@
 ---
-"@gradio/client": minor
-"gradio": minor
+"@gradio/client": patch
+"gradio": patch
 ---
 
-feat:Connect heartbeat if state created in render. Also fix config cleanup bug #8407
+fix:Connect heartbeat if state created in render. Also fix config cleanup bug #8407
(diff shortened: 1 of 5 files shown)


aliabid94:
Would it be a bad idea to remove this complexity entirely and just always maintain a heartbeat? Now the configs can change, I feel like it's a lot of complexity to keep track of whethers there's any state variables to track, and it doesn't seem like we'd consume a lot of resources by just always having a heartbeat. (Sorry should've suggested this earlier)

freddyaboulton:
My concern there is running into the 6 concurrent connection limit if you open multiple tabs of the same demo.

aliabid94:
we could use long polling then instead for the connect heartbeat, no? e.g. we send a POST request every 10 min, and if we haven't received anything in 30 min for an app we clear it out.

I think this would cut down complexity a lot, and we still could hit the 6 connection limit on any app that has state.

freddyaboulton:
Thanks @aliabid94 !

PR #8411, opened 2024-05-29
Fix flaky render test

aliabid94:
Render test was flaky because sometimes it would start editing a component while it was being re-rendered. Edited the test to start editing the last component in a render first, which ensures the render is complete before starting to edit.


pngwn (approved):
Seems reasonable. Thanks @aliabid94!

PR #8415, opened 2024-05-30
Fix spaces load error

aliabid94:
Fixes spaces load error on old spaces that don't provide id's in their dependencies.

Test with 
```
<gradio-app
		src="https://gradio-chatinterface-streaming-echo.hf.space"
	></gradio-app>		
```


aliabd (approved):
Looks good thanks @aliabid94

PR #8419, opened 2024-05-30
Fixes to Website Demo Upload Script 

aliabd:
We've been having a recurring issue where the script fails and many of the demos don't upload. It seems like the issue was with the create_repo function unnecessarily running too many times and creating a bunch of 409 errors which lead to 403 errors. 

I made some changes: 
- create_repo will only run if a space url doesn't return 200 
- the client is created once instead of for every demo 
- the create/upload logic is inside a try except so it doesn't stop other demos from getting uploaded. 

I tested this with a list of over 500 demos and there wasn't any errors! Will keep an eye on it over next few days


freddyaboulton (approved):
Thanks @aliabd !

pngwn (approved):
Looks good. Thanks @aliabd!

Issue #7118, opened 2024-01-23
Allow `gr.Chatbot` to accept messages in the openai format


abidlabs:
The `openai` spec for chatbots has become the standard for llms, we should support messages in that format (which is essentially a list of dictionaries) for the `gr.Chatbot`.

abidlabs:
I realized that this is actually not as advantageous as I thought. The main bottleneck when using `gr.ChatInterface` is that the history needs to be converted from the format it is stored by the `gr.Chatbot` to a format that is acceptable by openai/langchain. We could make the switch but then we'd be pretty vendor-specific and it would not be backwards-compatible.

See code examples here: https://www.gradio.app/guides/creating-a-chatbot-fast#a-langchain-example

Let's close this issue as I don't see a way to do a closer integration in a backwards-compatible way.

PR #8422, opened 2024-05-30
Support Message API for chatbot and chatinterface

freddyaboulton:
### Main Changes:
* Adds a `msg_format` parameter to `Chatbot` and `ChatInterface` so that messages can be returned as either the current list of tuples or a dictionary that is a superset of the messages/"OAI" format, e.g. {"role": "user", "content": ""}.
* Adds a ChatMessage dataclass that can be used instead of a dictionary when `msg_format="messages". This is nice because the IDE will autocomplete.
Screenshot 2024-07-02 at 10 51 54 AM

* ~If `msg_format` is "messages", then in `ChatInterface` developers can just yield the next token. They don't have to yield the entire message up to and including that token. I think this makes demos easier to write. And lets developers simply yield from their iterator.~
* Added the ability to parametrize e2e tests. If you create a python file that ends with `_testcase.py` in a demo directory corresponding to an e2e test, that demo will also be loaded to the e2e app. And you can use `go_to_testcase` to navigate to that testcase.

### Messages format overview

The message format is a dict with two required keys `role` and `content`. There is an additional `metadata` key that is not required but it can be used for tools and additional info about the message. Most messages returned from an "openai compatible" client will be compatible with gradio. 

The implementation of the message is below:

```python
class Metadata(GradioModel):
    title: Optional[str] = None

class Message(GradioModel):
    role: str
    metadata: Metadata = Field(default_factory=Metadata)
    content: str | FileData
```

### Examples

**Realistic Inference API Streaming**
```python
from huggingface_hub import InferenceClient
import gradio as gr

"""
For more information on `huggingface_hub` Inference API support, please check the docs: https://huggingface.co/docs/huggingface_hub/v0.22.2/en/guides/inference
"""
client = InferenceClient("HuggingFaceH4/zephyr-7b-beta")


def respond(
    prompt: str,
    history,
):
    if not history:
        history = [{"role": "system", "content": "You are a friendly chatbot"}]
    history.append({"role": "user", "content": prompt})

    yield history

    response = {"role": "assistant", "content": ""}
    for message in client.chat_completion(
        history,
        temperature=0.95,
        top_p=0.9,
        max_tokens=512,
        stream=True,
    ):
        response["content"] += message.choices[0].delta.content or ""

        yield history + [response]


with gr.Blocks() as demo:
    gr.Markdown("# Chat with Hugging Face Zephyr 7b 🤗")
    chatbot = gr.Chatbot(
        label="Agent",
        msg_format="messages",
        avatar_images=(
            None,
            "https://em-content.zobj.net/source/twitter/376/hugging-face_1f917.png",
        ),
    )
    prompt = gr.Textbox(lines=1, label="Chat Message")
    prompt.submit(respond, [prompt, chatbot], [chatbot])


if __name__ == "__main__":
    demo.launch()
```

**ChatInterface Streaming**
```python
import time
import gradio as gr

def slow_echo(message, history):
    yield f"You typed: "
    for i in range(len(message)):
        time.sleep(0.05)
        yield message[i]


demo = gr.ChatInterface(slow_echo, msg_format="messages").queue()

if __name__ == "__main__":
    demo.launch()
```

**ChatInterface Multimodal**
```python
import gradio as gr


def echo(message, history):
    return message["text"]


demo = gr.ChatInterface(
    fn=echo,
    examples=[{"text": "hello"}, {"text": "hola"}, {"text": "merhaba"}],
    title="Echo Bot",
    multimodal=True,
    msg_format="messages",
)
demo.launch()
```

**Chatbot**
```python
import gradio as gr
import random
import time

with gr.Blocks() as demo:
    chatbot = gr.Chatbot(msg_format="openai")
    msg = gr.Textbox()

    def respond(message, chat_history: list):
        chat_history.extend([{"role": "user", "content": message}, {"role": "assistant", "content": bot_message}])
        time.sleep(2)
        return "", chat_history

if __name__ == "__main__":
    demo.launch()
```

**Multimodal Chatbot**

```python
import gradio as gr
import time

    for x in message["files"]:
        history.append({"role": "user", "content": {"path": x}})
        history.append({"role": "user", "content": message["text"]})

def bot(history: list):
    response = "**That's cool!**"
    history.append({"role": "assistant", "content": ""})
    for character in response:
        history[-1]['content'] += character
        time.sleep(0.05)
        yield history

with gr.Blocks() as demo:
    chatbot = gr.Chatbot(
        [],
        elem_id="chatbot",
        bubble_full_width=False,
        msg_format="messages"
    )

    chat_input = gr.MultimodalTextbox(interactive=True, file_types=["image"], placeholder="Enter message or upload file...", show_label=False)

    bot_msg = chat_msg.then(bot, chatbot, chatbot, api_name="bot_response")

demo.queue()
if __name__ == "__main__":
    demo.launch()

```

Closes: https://github.com/gradio-app/gradio/issues/7118


abidlabs:
@freddyaboulton this looks great! Just one quibble: I'd suggest not using "openai" as the name of the format. It might be that they modify their message format in a way that we don't want to track. Also [Anthropic](https://docs.anthropic.com/en/api/messages) and others use the same format as well. What about `"tuples" | "dicts"`?

So thinking how we could add support for components into this: we could modify the `content` key to accept `Component` as well?

```py
class Message(GradioModel):
    role: str
    metadata: Metadata = Field(default_factory=Metadata)
    content: str | FileData | Component
```

wdyt @dawoodkhan82 @freddyaboulton

freddyaboulton:
Yes makes sense regarding renaming. What about "messages" (the name used by tgi/transformers and from the looks of it is actually the industry standard name [transformers docs](https://huggingface.co/docs/text-generation-inference/en/messages_api) [anthropic docs ](https://docs.anthropic.com/en/api/messages)

dawoodkhan82:
This format looks good.

> So thinking how we could add support for components into this: we could modify the content key to accept Component as well?

Regrading this, it would have to be another dict `ComponentMessage` or `ComponentContent` which stores the component name, the processed value, and the constructor_args.

```
class ComponentMessage(GradioModel):
    component: str
    value: Any
    constructor_args: List[Dict[str, Any]]
```

abidlabs:
imo it would be much nicer DX if the `ComponentMessage` class was only used internally and the developer could just pass in a Component object, e.g. `gr.Gallery([..., ...])`

demo/chatbot_core_components_simple/messages_testcase.py
```diff
@@ -0,0 +1,101 @@
...
+        add_message, [chatbot, chat_input], [chatbot, chat_input]
+    )
+    bot_msg = chat_msg.then(
+        bot, [chatbot, response_type], chatbot, api_name="bot_response"
+    )
+    bot_msg.then(lambda: gr.MultimodalTextbox(interactive=True), None, [chat_input])
+
+    chatbot.like(print_like_dislike, None, None)
+
+demo.queue()
```
abidlabs, line 99: nit: we can remove .queue() from all of these demos

gradio/chat_interface.py
```diff
@@ -182,10 +186,18 @@ def __init__(
                 Markdown(description)
 
             if chatbot:
+                if self.msg_format != chatbot.msg_format:
+                    raise ValueError(
```
abidlabs, line 190: I can imagine users passing in a custom `gr.Chatbot` but forgetting to set `msg_format` to to "messages" so alternatively, we could just raise a warning when they mismatch and say that the `msg_format` to `ChatInterface` will be used

gradio/components/chatbot.py
```diff
@@ -252,29 +321,71 @@ def preprocess(
...
+            Passes the messages in the chatbot as a `list[list[str | None | tuple]]`, i.e. a list of lists. The inner list has 2 elements: the user message and the response message. Each message can be (1) a string in valid Markdown, (2) a tuple if there are displayed files: (a filepath or URL to a file, [optional string alt text]), or (3) None, if there is no message displayed.
+        """
+        if payload is None:
+            return payload
+        if self.msg_format == "tuples":
+            if not isinstance(payload, ChatbotDataTuples):
+                raise Error("Data incompatible with tuples format")
+            return self._preprocess_messages_tuples(cast(ChatbotDataTuples, payload))
+        if not isinstance(payload, ChatbotDataMessages):
+            raise Error("Data incompatible with openai format")
```
abidlabs, line 349:
```suggestion
            raise Error("Data incompatible with the `messages` format")
```

gradio/components/chatbot.py
```diff
@@ -252,29 +321,71 @@ def preprocess(
...
+        Parameters:
+            payload: data as a ChatbotData object
+        Returns:
+            Passes the messages in the chatbot as a `list[list[str | None | tuple]]`, i.e. a list of lists. The inner list has 2 elements: the user message and the response message. Each message can be (1) a string in valid Markdown, (2) a tuple if there are displayed files: (a filepath or URL to a file, [optional string alt text]), or (3) None, if there is no message displayed.
+        """
+        if payload is None:
+            return payload
+        if self.msg_format == "tuples":
+            if not isinstance(payload, ChatbotDataTuples):
+                raise Error("Data incompatible with tuples format")
```
abidlabs, line 346:
```suggestion
                raise Error("Data incompatible with the `tuples` format")
```

demo/chatbot_streaming/testcase_messages.py
abidlabs, line 1: This demo isn't working for me. Getting a couple of errors saying "Data incompatible with openai format"

demo/chatbot_with_tools/run.py
abidlabs, line 1: Very cool! Just some comments on the UI:

image

1. Its not clear that you can click on the error message or tool use message and expand to get more details. I would suggest replacing with an accordion-like element where the toggle icon is clear
2. As seen in the screenshot, the color of the tool use message is the "primary" color even though the rest of the bot message is not that color. In fact, its the user message that is the primary color, which is confusing.

abidlabs: Btw it'd be awesome if we can include a real example of tool use with transformers agents or one of the other llm providers in our docs!

gradio/components/chatbot.py
```diff
@@ -252,29 +321,71 @@ def preprocess(
...
+        self,
+        payload: ChatbotDataTuples | ChatbotDataMessages | None,
+    ) -> (
+        list[list[str | tuple[str] | tuple[str, str] | None]] | list[MessageDict] | None
+    ):
+        """
+        Parameters:
+            payload: data as a ChatbotData object
+        Returns:
+            Passes the messages in the chatbot as a `list[list[str | None | tuple]]`, i.e. a list of lists. The inner list has 2 elements: the user message and the response message. Each message can be (1) a string in valid Markdown, (2) a tuple if there are displayed files: (a filepath or URL to a file, [optional string alt text]), or (3) None, if there is no message displayed.
```
abidlabs, line 340: We should update this "Returns" docstring  as it is used to generate the docs

gradio/components/chatbot.py
```diff
@@ -287,54 +398,107 @@ def create_file_message(chat_message, filepath):
...
-                | tuple
-            ]
-            | None
-        ),
-    ) -> ChatbotData:
+        value: TupleFormat | list[MessageDict | Message] | None,
+    ) -> ChatbotDataTuples | ChatbotDataMessages:
         """
         Parameters:
             value: expects a `list[list[str | None | tuple]]`, i.e. a list of lists. The inner list should have 2 elements: the user message and the response message. The individual messages can be (1) strings in valid Markdown, (2) tuples if sending files: (a filepath or URL to a file, [optional string alt text]) -- if the file is image/video/audio, it is displayed in the Chatbot, or (3) None, in which case the message is not displayed.
```
abidlabs, line 473: We should update this "value" docstring as well

gradio/components/chatbot.py
```diff
@@ -287,54 +398,107 @@ def create_file_message(chat_message, filepath):
...
-                )
-            processed_messages.append(
-                [
-                    self._postprocess_chat_messages(message_pair[0]),
-                    self._postprocess_chat_messages(message_pair[1]),
-                ]
-            )
-        return ChatbotData(root=processed_messages)
+            return ChatbotDataTuples(root=[])
+        if self.msg_format == "tuples":
```
abidlabs, line 479: Note: for all of our other components that can accept different types (e.g. `Image` with multiple types such as PIL images, numpy arrays, string filepaths), we don't enforce the `type` in `postprocess()` -- we just automatically figure it out based on `value`. Potentially, we could do the same thing here

abidlabs:
Nits:
-------

1. Not sure if related to this PR, but noticed a UI issue when multiple user messages are one after another:

```py
import gradio as gr

with gr.Blocks() as demo:
    gr.Chatbot([
        gr.ChatMessage(role="user", content="Hello!"),
        gr.ChatMessage(role="user", content="Hello!")        
    ], msg_format="messages")
    
demo.launch()
```

image

2. Any non-"user" role is treated like an "assistant" just fyi:

```py
import gradio as gr

with gr.Blocks() as demo:
    gr.Chatbot([
        gr.ChatMessage(role="abc", content="Hello!"),
    ], msg_format="messages")
    
demo.launch()
```

image

3. Is it expected that this doesn't work? No bot message is printed for me

```py
import gradio as gr

demo = gr.ChatInterface(lambda x,y:x, msg_format="messages")
    
demo.launch()
```

image



@freddyaboulton very nice PR! Made a first pass and left some comments above. Down to do another deeper review again once these comments are addressed

gradio/components/chatbot.py
freddyaboulton: We use the msg_format to set the data model and api info. So I want to make sure it matches in postprocess. Otherwise, it would be possible for the API info to tell you to expect dictionaries but you get tuples.

demo/chatbot_with_tools/run.py
freddyaboulton: Will make the font color more consistent! I like the minimal toolbox though and it matches the hugging face styling. Does anyone else have thoughts?

Agreed about the demo but to make it interesting we need to add api tokens etc so don't want to include it in tests or the repo. Will prepare something for the launch though!

demo/chatbot_streaming/testcase_messages.py
freddyaboulton: Fixed - issue was a typo in the `msg_format` parameter.

freddyaboulton:
Thanks for the review @abidlabs !! I think I got all of the comments (and added some more unit tests because of them :) )

abidlabs (approved):
LGTM @freddyaboulton this looks great! Tested `gr.Chatbot` and `gr.ChatInterface` with the new format and everything is working as expected. 

I just have three small points of feedback, which I'll leave below

abidlabs:
(1) The first concerns this:

> If msg_format is "messages", then in ChatInterface developers can just yield the next token. They don't have to yield the entire message up to and including that token. I think this makes demos easier to write. And lets developers simply yield from their iterator.

Although I agree that this makes demos easier to write, this introduces a different behavior for iterators in the very special case where you are iterating from a ChatInterface with msg_format="messages". This is likely to confuse users who are used to sending the complete message with `yield` in all other cases. It also could lead to bugs. For example, if you run 

```
python demo/chatinterface_streaming_echo/messages_testcase.py
```

and then use it via the client, e.g.

```py

client = Client("http://127.0.0.1:7864/")
result = client.predict(
		message="Hello!!",
		api_name="/chat"
)
print(result)
```

You only get "!" (the final token). But if you run the regular version of this demo (with msg_format="tuples"), you get the entire final string: "You typed: Hello!!". This introduces a discrepancy between what a user would observe if they used the Gradio UI and what you get when you make a prediction with the client.

abidlabs:
(2) Just to reiterate the earlier point about the design of tools, I think we can improve the UI quite a bit

image

* We should make the color of the tool "box" match the color of the messages
* We should provide some visual indication that the tool "box" can be clicked to expand the message

On the second point, I think an accordion would be the best UI. This is one area I like [Streamlit's UI](https://blog.langchain.dev/langchain-streamlit/). If its an accordion, we should keep the accordion open if the accordion is the final message, but then collapse it if there are subsequent messages. 

cc @pngwn @hannahblair on this front

abidlabs:
(3) Let's add some docs for this, perhaps in the chatbot/chatinterface guides. Excited to do some nice comms here!

abidlabs:
Discussed with @freddyaboulton and we don't need this:

> We should make the color of the tool "box" match the color of the messages

if we are bringing the bubbles back. @pngwn perhaps you could review the design of the `chatbot_with_tools` demo after you revert the bubbles back to ensure it looks good in both light and dark mode.

js/chatbot/Index.svelte
```diff
@@ -11,17 +11,19 @@
...
-		type NormalisedMessage,
-		normalise_messages
-	} from "./shared/utils";
+	import { normalise_tuples, normalise_messages } from "./shared/utils";
 
 	export let elem_id = "";
 	export let elem_classes: string[] = [];
 	export let visible = true;
-	export let value: messages = [];
+	export let value: TupleFormat | Message[] = [];
```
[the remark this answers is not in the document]
freddyaboulton: Yes it would be breaking for clients so lets do in 5.0!

js/chatbot/shared/ChatBot.svelte
```diff
@@ -173,45 +172,76 @@
...
+				currentRole = message.role;
+			}
+		}
+
+		if (currentGroup.length > 0) {
+			groupedMessages.push(currentGroup);
+		}
+
+		return groupedMessages;
 	}
```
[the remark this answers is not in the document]
freddyaboulton: So that they would be displayed in the same bubble but we got rid of the bubbles lol

js/chatbot/Index.svelte
```diff
@@ -52,9 +55,12 @@
...
 	export let avatar_images: [FileData | null, FileData | null] = [null, null];
 
-	let _value: [NormalisedMessage, NormalisedMessage][] | null = [];
+	let _value: NormalisedMessage[] | null = [];
 
-	$: _value = normalise_messages(value, root);
+	$: _value =
+		msg_format === "tuples"
+			? normalise_tuples(value as TupleFormat, root)
+			: normalise_messages(value as Message[], root);
```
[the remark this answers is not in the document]
freddyaboulton: There is slightly different processing steps for each case so thought that keeping them separate would be best

dawoodkhan82 (approved):
lgtm. Thanks for working on this @freddyaboulton

js/chatbot/shared/ChatBot.svelte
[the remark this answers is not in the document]
freddyaboulton: I think for sure the bot messages look cleaner in the same bubble. Happy to iterate with you when we bring them back.

PR #8429, opened 2024-05-31
Fix type hints for render and on

freddyaboulton:
## Description

I noticed two things that were off about the type hints for on and render
- We were using `Any` for the triggers of `gr.on`. Which was too broad.
- In `render`, the triggers were not typed correctly so passing a valid trigger, e.g. `slider.change` would be flagged in vs code. 
- In `render`, the type hint did not allow for a single component to be passed as input, but it is allowed.

**Main**
Screenshot 2024-05-31 at 2 09 04 PM
Screenshot 2024-05-31 at 2 03 46 PM

**PR**
Invalid value for `triggers` of `gr.on` is flagged
image

Passing an event listener method for `render` is not flagged
image

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/gradio/renderable.py
+++ b/gradio/renderable.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Callable, List, Literal, Union, cast
+from typing import TYPE_CHECKING, Callable, List, Literal, Sequence, Union, cast
 
 from gradio_client.documentation import document
 
@@ -80,7 +80,7 @@ class Renderable:
 @document()
 def render(
     inputs: list[Component] | Component | None = None,
-    triggers: list[EventListenerCallable] | EventListenerCallable | None = None,
+    triggers: Sequence[EventListenerCallable] | EventListenerCallable | None = None,
     *,
     queue: bool = True,
     trigger_mode: Literal["once", "multiple", "always_last"] | None = "always_last",
--- a/gradio/renderable.py
+++ b/gradio/renderable.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Callable, Literal
+from typing import TYPE_CHECKING, Callable, List, Literal, Union, cast
 
 from gradio_client.documentation import document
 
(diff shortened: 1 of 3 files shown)


gradio/renderable.py
```diff
@@ -76,8 +79,8 @@ def apply(self, *args, **kwargs):
 
 @document()
 def render(
-    inputs: list[Component] | None = None,
-    triggers: list[EventListener] | EventListener | None = None,
+    inputs: list[Component] | Component | None = None,
+    triggers: Sequence[EventListenerCallable] | EventListenerCallable | None = None,
```
abidlabs, line 83: The only issue with putting `Sequence` here is that it'll affect the docs for gr.render(): https://www.gradio.app/docs/gradio/render

abidlabs: And I think many Python users don't know what `Sequence` is

abidlabs: Can we add a step in the docs generation to convert `Sequence` container types to `list` @aliabd?

aliabid94 (approved):
LGTM! I was trying to see if we could actually use EventListenerCallable to typehint EventListener.listener, so that if EventListenerCallable ever diverges from the actual signature of a listener, our typecheker catches it. Cou;dn't figure it out though

gradio/renderable.py
freddyaboulton: I think I can edit `01_render.svx` to do this but maybe we do this once users complain? I would prefer to not introduce a difference between docs and code unless it's absolutely necessary imo.

abidlabs: Ok I think this would be applicable in many other parts of the codebase as well (e.g. anywhere we currently type hint list[Component] should really be Sequence[Component]) but doing it later is fine

abidlabs (approved):
LGTM

gradio/renderable.py
freddyaboulton: Ok I think we can edit the ParamTable svelte component in that case but will hold off for now.

freddyaboulton:
Thanks for the reviews everyone!

Issue #5120, opened 2023-08-07
JS client does not manage state like the python client


freddyaboulton:
### Describe the bug

When calling a stateful demo with the js client, the user is expected to manage the state themselves unlike the python client.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

with gr.Blocks() as demo:
    state = gr.State(0)
    click = gr.Button(value="Click Me")
    n_clicks = gr.Number()
    click.click(lambda state: (state + 1, state + 1), state, [state, n_clicks], api_name="click")

demo.launch()
```

### Python client
<img width="432" alt="image" src="https://github.com/gradio-app/gradio/assets/41651716/9968bdfa-1c26-4d36-b7aa-496da9b39168">

### JS client
<img width="997" alt="image" src="https://github.com/gradio-app/gradio/assets/41651716/9c973f19-a3f6-42f0-b895-8b76b2951226">

Server log:
<img width="803" alt="image" src="https://github.com/gradio-app/gradio/assets/41651716/b2e6762d-7cea-44ad-bbb5-9822d31fa665">

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Darwin
gradio version: 3.39.0
gradio_client version: 0.3.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.1.0
aiohttp: 3.8.4
altair: 4.2.0
fastapi: 0.100.0
ffmpy: 0.3.0
gradio-client: 0.3.0
httpx: 0.23.0
huggingface-hub: 0.16.4
jinja2: 3.1.2
markdown-it-py: 2.2.0
markupsafe: 2.1.3
matplotlib: 3.7.1
mdit-py-plugins: 0.3.3
numpy: 1.22.4
orjson: 3.9.1
packaging: 23.1
pandas: 1.5.3
pillow: 9.5.0
pydantic: 2.0.3
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0
requests: 2.31.0
semantic-version: 2.10.0
typing-extensions: 4.7.1
uvicorn: 0.22.0
websockets: 11.0.3


gradio_client dependencies in your environment:

fsspec: 2023.6.0
httpx: 0.23.0
huggingface-hub: 0.16.4
packaging: 23.1
requests: 2.31.0
typing-extensions: 4.7.1
websockets: 11.0.3
```

### Severity

I can work around it

PR #8439, opened 2024-06-03
Handle gradio apps using `state` in the JS Client

hannahblair:
## Description

We were receiving a param error when a space was using state. This PR adds logic to send `null` for each `state` param in the API, allowing us to receive the expected data. 

Closes: #5120

👻  Note, I haven't added a test for this because we still haven't got tests in place for `submit` and `predict`, which are TBD. I'm not sure where would be worth adding a test until we have those relevant tests. 

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/young-poets-change.md
+++ b/.changeset/young-poets-change.md
@@ -5,4 +5,4 @@
 "gradio": patch
 ---
 
-fix:Handle spaces using `state` in the JS Client
+fix:Handle gradio apps using `state` in the JS Client


abidlabs:
Thanks @hannahblair! We should also make sure to remove any output values that correspond to a `gr.State` output component. 

For example, here:

```py
import gradio as gr

gr.ChatInterface(lambda x,y:x).launch()
```

The docs state that the `.predict()` function should take a single input and return a single output:

image

When I ran this code:

```js
import { Client } from "@gradio/client";

const app = await Client.connect("http://127.0.0.1:7860/");
const result = await app.predict("/chat", {		
	message: "Hello!!",
});

console.log(result.data);
```

the function did indeed accept a single input, but it also returned a null value corresponding to the state output component that should not be there:

```
% node index.js
[ 'Hello!!', null ]
```

As a side note, I needed to do:

```
const app = await Client.connect("http://127.0.0.1:7860/");
```

not

```
const app = await Client.connect("http://localhost:7860/");
```

otherwise I see this error:

```
file:///Users/abidlabs/dev/gradio-repos/gradio-js-client/node_modules/@gradio/client/dist/index.js:1900
      throw Error(e);
            ^

Error: Error: TypeError: fetch failed
    at Client.init (file:///Users/abidlabs/dev/gradio-repos/gradio-js-client/node_modules/@gradio/client/dist/index.js:1900:13)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Client.connect (file:///Users/abidlabs/dev/gradio-repos/gradio-js-client/node_modules/@gradio/client/dist/index.js:1937:5)
    at async file:///Users/abidlabs/dev/gradio-repos/gradio-js-client/index.js:3:13
```

client/js/src/helpers/data.ts
```diff
@@ -118,3 +120,61 @@ export function post_message<Res = any>(
...
+): any[] {
+	let updated_payload: unknown[] = [];
+
+	dependency.inputs.map((input_id: number, index: number): any => {
+		const component = components.find((c: any) => c.id === input_id);
+		if (component?.type === "state") {
+			return null;
+		}
+		updated_payload.push(resolved_payload[index]);
+	});
```
pngwn, line 178: This is the main issue. That return null isn't being returned anywhere, I'd switch this to a `.forEach` and push the `null` value to the `updated_payload` array as well (or return everything and return the map directly).

pngwn:
**problems**

1. While direct client use should strip state `null` values from the output, we need to keep them there for use in gradio directly because the server expects them as do frontend functions etc.
2. We need to be aware of whether or not `null` state values are present in the payload or not when adding them in.

**solutions**
1. I added `input` and `output` variants. Then we could be more granular about which 'phase' we were in. I also added an option to the client decide whether not to keep state values. in gradio we set this to `true` manually but we remove them by default (for raw client use)
2. we compare the `resolved_payload` length with the `dep.inputs` length. If they match state values are present and we don't need to add them.

PR #8444, opened 2024-06-03
Remove deprecated parameters from Python Client

abidlabs:
This PR:

* Removes the `serialize` argument from `Client` which was not used anywhere
* Removes the `upload_files` argument from `Client` effectively setting its value to `False` everywhere. What this means is that now, any files that the Client submits need to be wrapped in the `file()` method.
* Consolidates the `output_dir` and `download_files` arguments in `Client`
* Renames `file()` to `handle_file()` across the codebase to be consistent with upcoming PR for JS Client


client/python/gradio_client/compatibility.py
```diff
@@ -61,8 +61,6 @@ def _inner(*data):
             if not self.is_valid:
                 raise utils.InvalidAPIEndpointError()
             data = self.insert_state(*data)
-            if self.client.upload_files:
```
freddyaboulton, line 64: Maybe we should still `serialize` the data here. I think this is breaking backwards compatibility with 3.x

client/python/gradio_client/client.py
```diff
@@ -122,9 +109,10 @@ def __init__(
         self.ssl_verify = ssl_verify
         self.space_id = None
         self.cookies: dict[str, str] = {}
-        self.output_dir = (
-            str(output_dir) if isinstance(output_dir, Path) else output_dir
-        )
+        if isinstance(self.download_files, (str, Path)):
```
freddyaboulton, line 112: nit we should check for empty string and if the path exists

freddyaboulton (approved):
Looks good @abidlabs thanks! one question about serialize in the compatibility class though

Issue #8436, opened 2024-06-03
How to modify State(list) to trigger State.change()


coolboyqu:
It seems that state.change() can only be triggered when variables' id changes, such as int, str, etc. For a list, if only the elements in the list are modified, state.change() cannot be triggered unless a new copy of the list is copied and returned. However, copying the list every time can waste a lot of time and resources. Is there any way to add elements to the list and trigger state.change() without copying the list?

abidlabs:
We're looking into a solution for this! cc @aliabid94

PR #8446, opened 2024-06-04
state.change listener with deep hash check

aliabid94:
state.change event listeners now do "deep checks", by deep hashing the old value. This makes @gr.render logic much cleaner. See demo/todo_list changes to see how much simpler the code is now because state values can be edited directly to induce a change.

Closes: https://github.com/gradio-app/gradio/issues/8436


pngwn:
How expensive is this?

aliabid94:
> How expensive is this?

Pretty cheap I would think. It runs a hash function for every element in the nested object, and only gets run if there's a state change listener involved. In regular python the same hash function is called whenever you add any element to a dictionary or a set, so hashing is already quite optimized.

gradio/utils.py
```diff
@@ -1402,3 +1403,25 @@ def connect_heartbeat(config: dict[str, Any], blocks) -> bool:
...
+                for k, v in sorted(obj.items(), key=lambda x: hash(x[0]))
+            ]
+        )
+    elif isinstance(obj, (list, tuple)):
+        items = tuple(deep_hash(x) for x in obj)
+    elif isinstance(obj, set):
+        items = tuple(deep_hash(x) for x in sorted(obj, key=hash))
+    else:
+        items = str(id(obj)).encode("utf-8")
+    hasher.update(repr(items).encode("utf-8"))
```
abidlabs, line 1426: Is there a reason you use `repr` here instead of `str`?

abidlabs:
It might be worth adding a note on the `State.change()` event listener docstring that this could get expensive if the `State` object is a sequence with a large number of elements

abidlabs (approved):
Tested and works well, thanks @aliabid94!

gradio/utils.py
aliabid94: its what chatgpt suggested lol

abidlabs: repr is more unique so that makes sense ✔️

aliabid94:
> It might be worth adding a note on the State.change() event listener docstring that this could get expensive if the State object is a sequence with a large number of elements

I don't think it's possible to directly add docs to the event of a single component. Regardless will update docs and guides on gr.State in a follow up PR.

PR #8451, opened 2024-06-04
Change client submit API to be an AsyncIterable and support more platforms

### LINKED ISSUE #4348: `gradio/client` does not work in non-node, non-browser environments - React native
### Describe the bug

After installing the @gradio/client package using `expo install @gradio/client` I get the following error:

```
The package at "node_modules/@gradio/client/dist/index.js" attempted to import the Node standard library module "node:buffer".
It failed because the native React runtime does not include the Node standard library.
```

### Reproduction

1. Install npm -> https://nodejs.org/en
2. Install expo cli : `npm install -g expo-cli`
3. Create a new app : `expo init test-app`
4. Choose the tabs (Typescript) template
5. Install Gradio Client : `expo install @gradio/client`
6. Start the app : "yarn web"

By following these steps you should end up getting the same error as me as seen in the screen shot below only I was using the ios simulator.

### Screenshot

Screenshot 2023-05-27 at 5 33 51 PM

### Logs

```shell
Logs for your project will appear below. Press Ctrl+C to exit.
iOS Bundling failed 681ms
The package at "node_modules/@gradio/client/dist/index.js" attempted to import the Node standard library module "node:buffer".
It failed because the native React runtime does not include the Node standard library.
```

### System Info

```shell
Gradio Version: 0.1.1 | OS: MacOS ventura 13.0.1 | ** using expo (https://expo.dev/) **
```

### Severity

blocking all usage of gradio

### LINKED ISSUE #8207: move away from `EventSource`

`EventSource` has a number of issues:

- it is only available in browsers, we resolve the issue with node by using a polyfill but this does not resolve the issues for non-node, non-browser environments. #4348
- it is not possible to set custom headers, this require us to work around authentication issues using a JWT and a special query param but this is far from ideal as it requires an additional network request, and is another remote API we need to keep up top date with. Additionally, while this works for spaces, it will not work when arbitrary headers are required. #7833 
- 

I think we should switch to the [Web Streams API](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API) which is supported in more environments. More specifically i think we should use [`fetch-event-stream` ](https://github.com/lukeed/fetch-event-stream) because:

- it is tiny (less than 1kb).
- it works in all environments (ish).
- it support worker environments, which may be helpful for `@gradio/lite`.
- I know the author and know it is a high quality library.
- It will allow us to use any HTTP method + set arbitrary headers.

See the following from the repo:

> - Allows any HTTP method
> - Built with native [Web Streams API](https://developer.mozilla.org/en-US/docs/Web/API/Streams_API)
> - Works with browser, Node.js, Cloudflare Workers, Deno, and Bun
> - Supports WebWorker or Service Worker environments
> - Accepts [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) for cancellable streams

I also think we should change the streaming API, but I'll open a new PR for that.

### LINKED ISSUE #8208: Change API for streams

The streaming API for the `@gradio/client` isn't really that nice. I think we should switch to an async iterator, we would go from this:

```js

const app = Client.create();
const job = app.submit("/endpoint", body)

job.on("data", data => {
  // do something with data
})
job.on("status",status => {
  // do something with status
})
```

To this:

```js
const app = Client.create();
const stream = app.submit("/endpoint", body)

for await (let event of stream) {
  console.log('<<', event.type); // 'status' | 'data'
  console.log('<<', event.data);
} 
```

Benefits:

- its just nicer
- fits into a linear `await` flow better which is the standard in JS now

**Other thoughts**

@freddyaboulton @abidlabs Should the python client follow a similar pattern and return an iterator for streams instead? I've seen this pattern quite a bit recently when working with streams and found it really nice.

pngwn:
## Description

Closes #8208.
Closes #8207.
Closes #4348 

## 🎯 PRs Should Target Issues

## Tests


pngwn:
Docs etc will come in a separate PR.

client/js/index.html
```diff
@@ -0,0 +1,39 @@
+<!doctype html>
```
aliabid94, line 1: delete this file?

pngwn: using it to test, will delete just before merge

client/js/src/test/upload_files.test.ts
```diff
@@ -29,7 +29,7 @@ describe("upload_files", () => {
 		expect(response.files[0]).toBe("lion.jpg");
 	});
 
-	it("should handle a server error when connected to a running app and uploading files", async () => {
+	it.skip("should handle a server error when connected to a running app and uploading files", async () => {
```
pngwn, line 32: file issue

client/js/src/types.ts
```diff
@@ -236,25 +227,25 @@ export interface FileData {
...
-};
-export interface LogMessage {
+export type GradioEvent = {
+	[P in EventType]: EventMap[P];
+}[EventType];
+
+// export type EventListener<K extends EventType> = (event: Event<K>) => void;
+// export type ListenerMap<K extends EventType> = {
+// 	[P in K]?: EventListener<K>[];
+// };
```
pngwn, line 243: remove

client/js/src/utils/submit.ts
```diff
@@ -651,6 +628,7 @@ export function submit(
 												endpoint: _endpoint,
 												fn_index
 											});
+											// close();
```
pngwn, line 631: remove

client/js/src/index.ts
```diff
@@ -7,9 +7,12 @@ export { FileData, upload, prepare_files } from "./upload";
...
+	StatusMessage,
 	client_return,
-	UploadResponse
+	UploadResponse,
+	RenderMessage,
+	LogMessage,
+	Payload
 } from "./types";
 
 // todo: remove in @gradio/client v1.0
```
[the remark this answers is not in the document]
pngwn: ah yeh, i'll add these back

client/js/src/client.ts
```diff
@@ -109,12 +120,12 @@ export class Client {
...
 		event_data?: unknown,
 		trigger_id?: number | null
-	) => SubmitReturn;
+	) => SubmitIterable<GradioEvent>;
 	predict: (
 		endpoint: string | number,
 		data: unknown[] | Record<string, unknown>,
 		event_data?: unknown
-	) => Promise<SubmitReturn>;
+	) => Promise<unknown>;
```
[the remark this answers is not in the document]
pngwn: :o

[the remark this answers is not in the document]
pngwn: Oh no, that is correct nice

client/js/src/utils/submit.ts
```diff
@@ -677,8 +655,9 @@ export function submit(
 										time: new Date()
 									});
 									if (["sse_v2", "sse_v2.1"].includes(protocol)) {
```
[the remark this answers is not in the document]
pngwn: Yes! NIce catch!

Issue #8457, opened 2024-06-05
Add API Recorder support for `curl`


abidlabs:
Did not get a chance to add  Add API Recorder support for `curl` in https://github.com/gradio-app/gradio/pull/8445, we'll need to add that later after the launch

PR #8460, opened 2024-06-05
Support Bash in Api Recorder 

aliabd:
Creates API recording snippets for bash like with js and python. There was a bit of a challenge given two requests are required and the event_id is needed for the second. To keep the snippet copy-paste-able without changes I opted for using [jq](https://jqlang.github.io/jq/) but i'm not sure that was the right call since the user may not have it installed already. Open to removing that.

Screen Shot 2024-06-04 at 8 56 49 PM

Also noticed a bug with the copy button on recording snippets. It would always just copy the last of the three languages to mount. Very weird. I fixed it in a bit of a roundabout way but it works as expected now. 

closes: #8457


abidlabs:
Thanks for taking care of this @aliabd! I like the suggestion of saving the event_id to a variable so that the snippet is completely copy-able. I'll make the regular code snippet like that as well. I do think we should avoid `jq` but it should be pretty easy to do this with `awk` which is installed by default in most OS-es

aliabd:
Made the changes if you can take another look @abidlabs

abidlabs (approved):
LGTM @aliabd! I tweaked the code snippets so that each command appears on a separate line

image

but otherwise, this is working great!

PR #8469, opened 2024-06-05
Fixes flaky state change test

aliabid94:
basically needed to have some UI change in response to a click so that the test would wait for the response in an expect. otherwise would "run too fast".


pngwn (approved):
Makes sense! Thanks @aliabid94!

Issue #6480, opened 2023-11-18
Add display time and screen location params to gr.info, gr.error, gr.warning


dwipper:
- [X] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
These pop-ups are great, but in some situations it would be a better user experience if the message cleared sooner and/or stayed longer. It would also be helpful to be able to control where on the screen message pops open.

**Describe the solution you'd like**  
1. Provide a time parameter in seconds that would control the amount of time the message is displayed.
2. Have a param provides the ability to position the message on the screen. Maybe allow a combination of top, middle, bottom, left, right, center. And/or potentially provide a control name reference, and have the message pop near that control. Would be great for field validation messages.

**Additional context**

Gauntlet173:
Is there a hack to set the timeout to infinite in the meantime? I have some long-running things that throw errors for the users and the helpful information is gone when they come back to it.

abidlabs:
@freddyaboulton this potentially related to your idea about controlling visibility of gr.Error

dwipper:
@freddyaboulton Rocks!

PR #8489, opened 2024-06-07
Control Display of Error, Info, Warning

freddyaboulton:
## Description

Adds `duration` property to `gr.Warning`, `gr.Info`, and `gr.Error`.

Setting `duration` to 0 not display the toast in the UI. This is useful for giving useful error messages for clients. For example, the data validation done in the component classes can now raise `gr.Error` instead of `TypeError` or `ValueError` and clients will see the exact error message.

The `duration` parameter is specified in seconds. If it's `None`, the toast will be displayed indefinitely.

Closes: https://github.com/gradio-app/gradio/issues/6480

modal_control

```python
import gradio as gr
from functools import partial

with gr.Blocks() as demo:
    with gr.Row():
        duration = gr.Number(label="Duration", info="Set to -1 for infinite", value=10, minimum=-1)
    with gr.Row():
        error = gr.Button("Error")
        info = gr.Button("Info")
        warning = gr.Button("Warning")

    def display_message(type, msg, duration):
        duration = None if duration < 0 else duration
        if type == "error":
            raise gr.Error(msg, duration=duration)
        elif type == "info":
            gr.Info(msg, duration=duration)
        elif type == "warning":
            gr.Warning(msg,  duration=duration)

    error.click(partial(display_message, "error", "ERROR 💥"),  [duration])
    info.click(partial(display_message, "info", "INFO ℹ️"), [duration])
    warning.click(partial(display_message, "warning", "WARNING ⚠️"), [duration])


demo.launch()
```

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/client/js/src/helpers/api_info.ts
+++ b/client/js/src/helpers/api_info.ts
@@ -318,6 +318,8 @@ export function handle_message(
 					status: {
 						queue,
 						message: data.output.error as string,
+						display: data.output.display as boolean,
+						duration: data.output.duration as number,
 						stage: "error",
 						code: data.code,
 						success: data.success
--- a/client/js/src/helpers/api_info.ts
+++ b/client/js/src/helpers/api_info.ts
@@ -318,7 +318,7 @@ export function handle_message(
 					status: {
 						queue,
 						message: data.output.error as string,
-						display: data.output.display as boolean,
+						visible: data.output.visible as boolean,
 						duration: data.output.duration as number,
 						stage: "error",
 						code: data.code,
--- a/gradio/exceptions.py
+++ b/gradio/exceptions.py
@@ -76,15 +76,14 @@ class Error(Exception):
     def __init__(
         self,
         message: str = "Error raised.",
-        display: bool = True,
-        duration: int | None = 10000,
+        duration: int | None = 10,
     ):
         """
         Parameters:
             message: The error message to be displayed to the user.
+            duration: The duration in seconds to display the error message. If None, the error message will be displayed until the user closes it. If 0, the error message will not be displayed in the UI but will still be raised as an exception.
         """
         self.message = message
-        self.display = display
         self.duration = duration
         super().__init__(self.message)
 
--- a/gradio/exceptions.py
+++ b/gradio/exceptions.py
@@ -73,12 +73,16 @@ class Error(Exception):
     Demos: calculator, blocks_chained_events
     """
 
-    def __init__(self, message: str = "Error raised."):
+    def __init__(self, message: str = "Error raised.",
+                 display: bool = True,
+                 duration: int | None = 10000):
         """
         Parameters:
             message: The error message to be displayed to the user.
         """
         self.message = message
+        self.display = display
+        self.duration = duration
         super().__init__(self.message)
 
     def __str__(self):
--- a/gradio/exceptions.py
+++ b/gradio/exceptions.py
@@ -73,9 +73,12 @@ class Error(Exception):
     Demos: calculator, blocks_chained_events
     """
 
-    def __init__(self, message: str = "Error raised.",
-                 display: bool = True,
-                 duration: int | None = 10000):
+    def __init__(
+        self,
+        message: str = "Error raised.",
+        display: bool = True,
+        duration: int | None = 10000,
+    ):
         """
         Parameters:
             message: The error message to be displayed to the user.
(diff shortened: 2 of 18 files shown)


js/_website/src/lib/templates/gradio/05_modals/error.svx
```diff
@@ -17,13 +17,19 @@
...
+raise gradio.Error("An error occured 💥!", duration=5)
 ```
 
 <!--- Description -->
 ### Description
 ## {@html style_formatted_text(obj.description)}
 
+## You can control for how long the error message is displayed with the `duration` parameter. If it's `None`, the message will be displayed forever until the user closes it. If it's a number, it will be shown for that many seconds. 
+
+## Below is a demo of how different values of duration control the error, info, and warning messages. You can see the code [here](https://huggingface.co/spaces/freddyaboulton/gradio-error-duration/blob/244331cf53f6b5fa2fd406ece3bf55c6ccb9f5f2/app.py#L17).
```
freddyaboulton, line 29: @aliabd The link here is not being formatted as a link. How can I fix it?

image

aliabd: pushed a commit that fixed it. also styling clash. 

what's happening with these issues is that i had overrided some default styling rules to design stuff (for example the side bar is a ul but we don't want bullet points there) and now we need them for the content.

freddyaboulton: Thank you!

abidlabs:
Btw I think this sufficiently closes https://github.com/gradio-app/gradio/issues/6480

> Provide a time parameter in seconds that would control the amount of time the message is displayed

This is complete

> Have a param provides the ability to position the message on the screen. Maybe allow a combination of top, middle, bottom, left, right, center. And/or potentially provide a control name reference, and have the message pop near that control. Would be great for field validation messages.

I don't think we want to do this, especially the second part as we are planning on instead having a dedicated way to do frontend validation

gradio/exceptions.py
```diff
@@ -73,12 +75,18 @@ def divide(numerator, denominator):
     Demos: calculator, blocks_chained_events
     """
 
-    def __init__(self, message: str = "Error raised."):
+    def __init__(
+        self,
+        message: str = "Error raised.",
+        duration: int | None = 10,
```
abidlabs, line 81: nit but 
```suggestion
        duration: int | float | None = 10,
```

pngwn:
With number values I'm used to `0` meaning unlimited, although in this case 0 meaning hidden does make sense I think.

aliabid94:
> With number values I'm used to 0 meaning unlimited, although in this case 0 meaning hidden does make sense I think.

I think it's just a bit confusing that 0 and None have such different meanings, when they're both "falsy". Wonder if another `visible=` kwarg would be more appropriate specifically for display.

client/js/src/helpers/api_info.ts
```diff
@@ -318,6 +318,8 @@ export function handle_message(
 					status: {
 						queue,
 						message: data.output.error as string,
+						display: data.output.display as boolean,
```
aliabid94, line 321: where is `display` passed from?

gradio/exceptions.py
aliabid94: just float is fine, int is a subset of float

gradio/queueing.py
```diff
@@ -376,6 +381,7 @@ def log_message(
         event_id: str,
         log: str,
         level: Literal["info", "warning"],
+        duration: int | None = 10000,
```
aliabid94, line 384: why 10000 by default?

aliabid94 (approved):
just a few comments, otherwise LGTM!

freddyaboulton:
Yes that's fair. Originally had a separate `display` kwarg but then thought that less parameters would be better. Will bring that back and clean up a bit to address the other comments. Thanks for the input all!

gradio/queueing.py
freddyaboulton: Originally had it being set in ms and forgot to change the default. Fixed now.

client/js/src/helpers/api_info.ts
freddyaboulton: Originally had a separate display parameter. Added it back as `visible` so this makes more sense now.

PR #8499, opened 2024-06-07
Cache break themes on change

aliabid94:
We load our theme variables from `/theme.css` but this means if the theme css changes, server or browser caches may not realize a change due to the same URL being used. LMSYS was facing this issue.
Now the CSS is loaded from `/theme.css?v={hash_of_theme_css_content}`.


freddyaboulton (approved):
Looks good @aliabid94 !

Issue #8463, opened 2024-06-05
When `every` set in gr.Button.click, the inner function cannot get live camera images


P2Oileen:
### Describe the bug

I am trying to show the camera image, but when I set button.click(every=xxx), it cannot get any image. This problem doesn't happen when I delete every=xxx and click the button by myself. 

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

def echo(webcam_image):
    if webcam_image is not None:
        gr.Info("webcam_image get")
    else:
        gr.Info("no webcam_image")
    return webcam_image
    
def main():
    with gr.Blocks() as demo:
        webcam = gr.Image(sources=["webcam"], height=300, streaming=True)
        img = gr.Image()
        button = gr.Button("Start!")
        button.click(echo, inputs=[webcam], outputs=[img], every=5)
    demo.launch(server_name="0.0.0.0")
    demo.queue()

if __name__ == "__main__":
    main()
```

### Screenshot

![image](https://github.com/gradio-app/gradio/assets/26021786/1a87d98f-ebb6-4c4b-afc8-05b69b9d8040)

### System Info

```shell
Gradio Environment Information: 
------------------------------
Operating System: Linux
gradio version: 4.32.2
gradio_client version: 0.17.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.3.0
fastapi: 0.110.2
ffmpy: 0.3.2
gradio-client==0.17.0 is not installed.
httpx: 0.27.0
huggingface-hub: 0.22.2
importlib-resources: 6.4.0
jinja2: 3.0.3
markupsafe: 2.1.2
matplotlib: 3.7.1
numpy: 1.24.4
orjson: 3.10.1
packaging: 23.1
pandas: 2.0.1
pillow: 9.5.0
pydantic: 2.7.1
pydub: 0.25.1
python-multipart: 0.0.9
pyyaml: 5.4.1
ruff: 0.4.1
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.12.3
typing-extensions: 4.11.0
urllib3: 2.2.1
uvicorn: 0.29.0
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2024.3.1
httpx: 0.27.0
huggingface-hub: 0.22.2
packaging: 23.1
[...]

Issue #7051, opened 2024-01-18
Keyboard interruption does not work when "every" is used (until demo is closed in browser)


muerrilla:
### Describe the bug

Running the code below, if I don't click the Run button, pressing ctrl+c closes the server as it should. But if I click Run and start the "every" event, pressing ctrl+c will print `Keyboard interruption in main thread... closing server.` but it will not work and the program keeps running. It only works (along with throwing the error below) when I close the demo window in the browser.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

def tick():
    print("Running")

with gr.Blocks() as demo:
    run = gr.Button(value="Run")
    cancel = gr.Button(value="Cancel")
    click_event = run.click(tick, None, None, every=1)
    cancel.click(None, None, None, cancels=[click_event])

demo.queue().launch(server_name="127.0.0.1", server_port=15433, inbrowser=True)
```

### Logs

```shell
Running on local URL:  http://127.0.0.1:15433

To create a public link, set `share=True` in `launch()`.
Running
Running
Running
Keyboard interruption in main thread... closing server.
Running
Running
Traceback (most recent call last):
  File "C:\Users\Muerrilla\anaconda3\envs\diffusers\Lib\site-packages\gradio\blocks.py", line 2233, in block_thread
    time.sleep(0.1)
KeyboardInterrupt

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "E:\diffusers\gr.py", line 12, in <module>
    demo.queue().launch(server_name="127.0.0.1", server_port=15433, inbrowser=True)
  File "C:\Users\Muerrilla\anaconda3\envs\diffusers\Lib\site-packages\gradio\blocks.py", line 2140, in launch
    self.block_thread()
  File "C:\Users\Muerrilla\anaconda3\envs\diffusers\Lib\site-packages\gradio\blocks.py", line 2237, in block_thread
    self.server.close()
  File "C:\Users\Muerrilla\anaconda3\envs\diffusers\Lib\site-packages\gradio\networking.py", line 76, in close
    self.thread.join()
  File "C:\Users\Muerrilla\anaconda3\envs\diffusers\Lib\threading.py", line 1112, in join
    self._wait_for_tstate_lock()
  File "C:\Users\Muerrilla\anaconda3\envs\diffusers\Lib\threading.py", line 1132, in _wait_for_tstate_lock
    if lock.acquire(block, timeout):
[...]

Issue #6577, opened 2023-11-24
`every` does not send the latest values of the input components


abidlabs:
### Describe the bug

If you create an event with `every`, it only sends the original values of the input components, instead of the latest.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```py
import gradio as gr

with gr.Blocks() as demo:
    c1 = gr.Checkbox()
    c2 = gr.Checkbox()
    demo.load(lambda x:x, c1, c2, every=1)
    
demo.launch()
```

### System Info

```shell
gradio==4.7.1
```

### Severity

I can work around it

howardwu1:
I noticed the same thing. What is the work around?

thiswillbeyourgithub:
Have had this issue for a while. This is problematic and I haven't found a workaround so far. I resort to clicking manually on a button to get a "status" value.

redevined:
Any updates on this? I don't really see a workaround either.
```python
import gradio as gr

with gr.Blocks() as demo:
    out = gr.Textbox()

    @demo.load(outputs=out, every=10)
    def periodic_task():
        return read_rows_from_some_database()
    
demo.launch()
```

howardwu1:
My workaround ended up requiring me to cancel the queue events whenever ui elements were interacted with. Not elegant but gets the job done with several lines of code if you have multiple ui.

freddyaboulton:
@redevined Can you explain what your problem is? It seems different from what this issue is tackling as your demo does not have any input components.

@howardwu1 - that's actually the correct approach here. It's shown in the `live_dashboard` demo:  https://github.com/gradio-app/gradio/blob/6be3c2c47a616c904c8497d1fbef7a851c54d488/demo/live_dashboard/run.py#L55 

It's a bit wordy but the alternative would require gradio to launch a different server request for each iteration which I don't think would scale well. Going to think about this some more.

redevined:
Sorry for the confusion, seems like I simplified my example a bit too much.

PR #8505, opened 2024-06-07
Add Timer component

aliabid94:
1. Implemented `gr.Timer`, a Component meant to replace `every`, e.g. `timer = gr.Timer(5)` for a 5 second regular timer. Can be used via:
  -  a `tick` event listener, e.g. `timer.tick(fn=..., inputs=[], outputs=[])`
  - an `every=` that can accept a timer  `gr.Textbox(value=lambda: ..., every=timer)

2. Also added `inputs=` kwarg to make function values more useful, e.g.  `gr.Textbox(value=lambda a, b: ..., inputs=[text1, text2])`. The component will automatically re-update if any of the the input components change.

Can all be combined for maximum "dashboard" shorthand:  `gr.Textbox(value=lambda a, b: ..., inputs=[text1, text2], every=timer)`,  which sets the value of the component as a function of the inputs, updating every timer tick and also when the inputs change.

See demo file that uses all these concepts:
```python
import gradio as gr
import random

countries = [
    "Algeria", "Argentina", "Australia", "Brazil", "Canada", "China", "Democratic Republic of the Congo", "Greenland (Denmark)", "India", "Kazakhstan", "Mexico", "Mongolia", "Peru", "Russia", "Saudi Arabia", "Sudan", "United States"
]

with gr.Blocks() as demo:
    with gr.Row():
        count = gr.Slider(1, 10, step=1, label="Country Count")
        alpha_order = gr.Checkbox(True, label="Alphabetical Order")

    gr.JSON(lambda count, alpha_order: countries[:count] if alpha_order else countries[-count:], inputs=[count, alpha_order])
    timer = gr.Timer(1)
    with gr.Row():
        gr.Textbox(lambda: random.choice(countries), label="Random Country", every=timer)
        gr.Textbox(lambda count: ", ".join(random.sample(countries, count)), inputs=count, label="Random Countries", every=timer)
    with gr.Row():
        gr.Button("Start").click(lambda: gr.Timer(active=True), None, timer)
        gr.Button("Stop").click(lambda: gr.Timer(active=False), None, timer)


if __name__ == "__main__":
    demo.launch()
```

Fixes: #8582 
Fixes: #6577 
Fixes: #8463 
Fixes: #7051 
Fixes: #6841


abidlabs:
> Also added inputs= kwarg to make function values more useful, e.g. gr.Textbox(value=lambda a, b: ..., inputs=[text1, text2])

This addition to the API  does feel useful to me ✔️

freddyaboulton:
>  I would instead make the every argument also be able to take a gr.Timer() instance, which can be started/stopped/adjusted as your PR implements.

I like this suggestion.

Also, I think right now the timer only triggers `tick` event and value functions. Can we make it trigger any event on a schedule?

For a use case where you want to start the schedule based on a button click or tab select, you have to do something like this

```python
with gr.Blocks() as demo:
    timer = gr.Timer(5, active=False)
    plot = gr.LinePlot(value=get_data, every=timer)
    button = gr.Button("Start")
    button.click(lambda: gr.Timer(active=True), None, timer)
```

If we let it trigger any event, the code could be rewritten like the following which feels more natural:

```python
with gr.Blocks() as demo:
    plot = gr.LinePlot()
    button = gr.Button("Start")
    button.click(get_data, None, plot, every=gr.Timer(5))
```

aliabid94:
> don't think it makes sense for gr.Timer() to be a context manager... instead make the every argument also be able to take a gr.Timer() instance

Okay sure. Originally my logic for using context managers for Timer and InputSet had been that in a dashboard, every component is tied to the same timer and set of inputs, and its annoying to have to set the timer and the inputs for each component, but that's something we can design some other time.

> I think right now the timer only triggers tick event and value functions. Can we make it trigger any event on a schedule? ...e.g. `button.click(get_data, None, plot, every=gr.Timer(5))`

A little confused by this syntax. Does the syntax above start the given timer, and then bind this same event to the tick of the Timer? If it gets clicked again, it does nothing?

Without any button control, the syntax is very clear:

```python
plot = gr.Plot()
timer = gr.Timer(5)
timer.tick(get_data, None, plot)
```

or simply:

```python
plot = gr.Plot(get_data, every = gr.Timer(5))
```

If you are adding a button to start the timer, then you probably want to be able to stop it too, so you need access to the timer object. In that case, you are suggesting:

```python
plot = gr.Plot()
timer = gr.Timer(5)
button.click(get_data, None, plot, every=timer)
button2.click(lambda: gr.Timer(active=False), outputs=timer)
```
correct? Imo that's messy compared to:

```python
timer = gr.Timer(5)
plot = gr.Plot(get_data, every=timer)
button.click(lambda: gr.Timer(active=True), outputs=timer)
button2.click(lambda: gr.Timer(active=False), outputs=timer)
```

The second option is much more explicit, takes the same amount of code, and starting and stopping use the same syntax.

I don't think it makes sense to add `every` to event listeners. It's a decent shorthand for components that have function values though.

freddyaboulton:
Ok let's stick with the original proposal to have every only apply to component value functions.

aliabid94:
Ready for review!

aliabd (commented):
Website build is failing because you added a component without creating a page for it in the docs. I can add it to the docs if you'd like @aliabid94

js/timer/Index.svelte
```diff
@@ -0,0 +1,22 @@
...
+	$: if (old_value !== value || active !== old_active) {
+		if (interval) clearInterval(interval);
+		if (active) {
+			interval = setInterval(() => {
+				gradio.dispatch("tick");
+			}, value * 1000);
+		}
+		old_value = value;
+		old_active = active;
+	}
```
pngwn, line 21: We need to cleanup in `onDestroy` just prevent possible memory leaks. A `clearInterval` if the interval exists should do.

aliabd:
@aliabid94 I added the docs, feel free to edit. I also edited some docstrings (just added the types to preprocess and postprocess) 

Theres a separate issue with the website build on main right now.

gradio/_simple_templates/simpledropdown.py
```diff
@@ -21,7 +24,8 @@ def __init__(
         value: str | int | float | Callable | None = None,
         label: str | None = None,
         info: str | None = None,
-        every: float | None = None,
+        every: Timer | None = None,
```
abidlabs, line 27: we should continue to accept both a `gr.Timer()` for a `float` for convenience and backwards-compatibility
```suggestion
        every: float | Timer | None = None,
```

gradio/_simple_templates/simpledropdown.py
```diff
@@ -38,7 +42,7 @@ def __init__(
             value: default value selected in dropdown. If None, no value is selected by default. If callable, the function will be called whenever the app loads to set the initial value of the component.
             label: component name in interface.
             info: additional component description.
-            every: If `value` is a callable, run the function 'every' number of seconds while the client connection is open. Has no effect otherwise. The event can be accessed (e.g. to cancel it) via this component's .load_event attribute.
+            every: If `value` is a callable, run the function every time the provided Timer ticks. Has no effect otherwise.
```
abidlabs, line 45: If `value` is a callable, run the function 'every' number of seconds while the client connection is open. Has no effect otherwise. If a `Timer` component is provided, the the function will be run every time the provided Timer ticks.

gradio/blocks.py
```diff
@@ -702,7 +702,7 @@ def set_event_trigger(
             batch: whether this function takes in a batch of inputs
             max_batch_size: the maximum batch size to send to the function
             cancels: a list of other events to cancel when this event is triggered. For example, setting cancels=[click_event] will cancel the click_event, where click_event is the return value of another components .click method.
-            every: Run this event 'every' number of seconds while the client connection is open. Interpreted in seconds.
+            every: Deprecated, use gr.Timer() instead. Run this event 'every' number of seconds while the client connection is open. Interpreted in seconds.
```
abidlabs, line 705: No it is not deprecated.

gradio/blocks.py
```diff
@@ -740,6 +740,10 @@ def set_event_trigger(
 
         if fn is not None and not cancels:
             check_function_inputs_match(fn, inputs, inputs_as_dict)
+        if isinstance(every, float):
+            warnings.warn(
```
abidlabs, line 744: No it is not deprecated.

gradio/components/timer.py
```diff
@@ -0,0 +1,67 @@
...
+        """
+        Parameters:
+            payload: The interval of the timer as a float or None.
+        Returns:
+            The interval of the timer as a float.
+        """
+        return payload
+
+    def api_info(self) -> dict:
+        return {"type": "string"}
```
abidlabs, line 61:
```suggestion
        return {"type": "number"}
```

guides/04_building-with-blocks/01_blocks-and-event-listeners.md
```diff
@@ -163,12 +163,29 @@ The `.then()` method of an event listener executes the subsequent event regardle
 
 ## Running Events Continuously
 
-You can run events on a fixed schedule using the `every` parameter of the event listener. This will run the event `every` number of seconds while the client connection is open. If the connection is closed, the event will stop running after the following iteration. Note that this does not take into account the runtime of the event itself. So a function with a 1 second runtime running with `every=5`, would actually run every 6 seconds. Also note that this parameter does not apply to the `js` function, only the Python function associated with the event listener.
```
abidlabs, line 166: We should preserve the old behavior because its very easy to use & backwards-compatibility

abidlabs: > Note that this does not take into account the runtime of the event itself. So a function with a 1 second runtime running with `every=5`, would actually run every 6 seconds. Also note that this parameter does not apply to the `js` function, only the Python function associated with the event listener.

This part is still true, right?

abidlabs:
* Changing the timer frequency doesn't work for me. In this demo, clicking the "Slow" button has no effect (though clicking the "Stop" button does work as expected):

```py
import gradio as gr
import random
import time

def slow_random():
    time.sleep(1)
    return random.random()

with gr.Blocks() as demo:
    t1 = gr.Timer(1)
    t2 = gr.Textbox(slow_random)
    t1.tick(slow_random, None, t2)
    b = gr.Button("Stop")
    b2 = gr.Button("Slow")
    b.click(lambda : gr.Timer(active=False), None, t1)
    b2.click(lambda : gr.Timer(3), None, t1)

demo.launch()
```

abidlabs:
> Also added inputs= kwarg to make function values more useful, e.g. gr.Textbox(value=lambda a, b: ..., inputs=[text1, text2]). The component will automatically re-update if any of the the input components change.

The docstrings for `inputs` is missing, and we should also add a section about this in the guides!

abidlabs:
Also, could we fix this issue https://github.com/gradio-app/gradio/issues/6577 in this PR (if it is not already fixed?)

aliabid94:
> Changing the timer frequency doesn't work for me. In this demo, clicking the "Slow" button has no effect (though clicking the "Stop" button does work as expected):

No it works, the only reason you think so is because the default timer has 1 second between ticks, and the function takes approx. 1 second to run. So it will usually feel like 2 seconds between runs because the there is a tick that isn't executing because the process is still running. So going from ~2 seconds to 3 seconds does not feel like a significant difference.

See code below for an exaggerated effect.

```python
with gr.Blocks() as demo:
    t1 = gr.Timer(1)
    t2 = gr.Textbox(slow_random)
    t1.tick(slow_random, None, t2)
    b = gr.Button("Stop")
    b2 = gr.Button("Slow")
    b3 = gr.Button("Fast")
    b.click(lambda : gr.Timer(active=False), None, t1)
    b2.click(lambda : gr.Timer(10), None, t1)
    b3.click(lambda : gr.Timer(0.1), None, t1)
```

aliabid94:
I believe everything has been addressed, could use a re-review cc @abidlabs, also @pngwn if you can look at my approach to avoid calls stacking up when the browser is minimized - this was happening because window.dispatchEvent was triggering while minimized, but window.addEventListener("gradio") was not issuing callbacks until the window was restored. Fixed by only firing ticks if the document is visible

client/python/gradio_client/client.py
```diff
@@ -453,11 +453,7 @@ def predict(
             client.predict(5, "add", 4, api_name="/predict")
             >> 9.0
         """
-        inferred_fn_index = self._infer_fn_index(api_name, fn_index)
-        if self.endpoints[inferred_fn_index].is_continuous:
```
abidlabs, line 457: Don't we still need this when the Client connects to Spaces running older versions of Gradio?

gradio/events.py
```diff
@@ -398,7 +427,7 @@ def event_trigger(
                 preprocess: If False, will not run preprocessing of component data before running 'fn' (e.g. leaving it as a base64 string if this method is called with the `Image` component).
                 postprocess: If False, will not run postprocessing of component data before returning 'fn' output to the browser.
                 cancels: A list of other events to cancel when this listener is triggered. For example, setting cancels=[click_event] will cancel the click_event, where click_event is the return value of another components .click method. Functions that have not yet run (or generators that are iterating) will be cancelled, but functions that are currently running will be allowed to finish.
-                every: Run this event 'every' number of seconds while the client connection is open. Interpreted in seconds.
+                every: Will be deprecated in favor of gr.Timer. Run this event 'every' number of seconds while the client connection is open. Interpreted in seconds.
```
abidlabs, line 430: Why does `every` for events behave differently than in components? We should just have the same behavior here, where you can provide a timer or a float?

guides/06_custom-components/07_pdf-component-example.md
```diff
@@ -570,7 +570,7 @@ class PDF(Component):
                  elem_classes: list[str] | str | None = None,
                  render: bool = True,
                  load_fn: Callable[..., Any] | None = None,
-                 every: float | None = None):
+                 every: Timer | float | None = None):
```
abidlabs, line 573: We'll need to import Timer

gradio/components/timer.py
```diff
@@ -0,0 +1,67 @@
...
+from gradio.events import Events
+
+if TYPE_CHECKING:
+    pass
+
+
+@document()
+class Timer(Component):
+    """
+    Create a timer that ticks at regular intervals when active.
```
abidlabs, line 19:
```suggestion
    A special component that ticks at regular intervals when active. The Timer component does not show up in the UI of your Gradio application. It is only meant to be used as an event listener to trigger events at a regular frequency.
```

abidlabs (approved):
Left a few nits, but can otherwise confirm that this works great. Great stuff @aliabid94!

gradio/events.py
aliabid94: I don't want to encourage the use of `every` added to an event listener which is a very weird API imo, so I'm only supporting floats for the sake of backwards compatibility.

abidlabs: Why not? Its quite an intuitive api -- once the event is triggered, it should rerun every X seconds. Its much easier than creating a gr.Timer object and setting it to active, etc.

aliabid94: This would be the 4th (!!!) way to support a continuous event, we already have:
1. using a direct timer with a tick listener
2. using every=float in a Component, which creates a timer under the hood
3. using every=Timer in a Component with an existing explicit timer

We keep coming up way too many ways to do the same thing and it makes the API messier and messier.

abidlabs: But these are not the same thing. Using `every` in a Component only allows you to rerun a function when the demo first loads. Whereas using `every` in an event listener allows you to rerun a function when an event is triggered.

In fact, I would argue that this would make things *more* consistent. Just like `every` in a Component can be either a float or a Timer, `every` in an event listener can be a float or a Timer

abidlabs: Synced with @aliabid94 and I'm convinced that its not a good idea to add yet another approach.

abidlabs:
all good from my end!

Issue #8494, opened 2024-06-07
Incorrect monitoring URL when using the root_path parameter


ai-marat:
### Describe the bug

If the user is using root_path, the link in the console for monitoring will be incorrect. This can be easily fixed by adding root_path to the URL.

Monitoring URL: http://127.0.0.1:7869/monitoring/sBfG5plI5lk5vbJEGgfAYQ
Monitoring URL: http://127.0.0.1:7869/root_path/monitoring/sBfG5plI5lk5vbJEGgfAYQ

p.s. A big thank you for the monitoring feature, it's lifesaving for me!

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

```

### System Info

```shell
gradio@4.36.0
```

### Severity

I can work around it

PR #8506, opened 2024-06-08
Use root url for monitoring url

abidlabs:
Closes: https://github.com/gradio-app/gradio/issues/8494

Ensures that the correct /monitoring URL is shown if running with root url or running on HF Spaces, etc.

Also ensures that non-logged in users can't hit /monitoring


freddyaboulton (approved):
Thanks @abidlabs - only comment is that I would write a unit test for this!

abidlabs:
Yup will do, thanks @freddyaboulton!

Issue #8495, opened 2024-06-07
Error when using js for event listeners


loganrenaud:
### Describe the bug

When I add a js function to any event listeners, there is an error and the function doesn't work. It worked well on gradio 4.29.0 but it doesn't work since gradio 4.31.
If I remove the js function from the event listener, everything is fine.
I wrote the simplest code for the reproduction.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

js = """
function test_function_js() {}
"""

def function():
    pass

with gr.Blocks() as app:
    button = gr.Button()
    button.click(
        fn=function,
        js=js,
     )
    
app.launch()
```

### Screenshot

![image](https://github.com/gradio-app/gradio/assets/90515810/727be085-6dce-4e71-aad5-ca3187057e00)
![image](https://github.com/gradio-app/gradio/assets/90515810/623db414-75f1-4276-8d67-e1e3d30852b6)

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Windows
gradio version: 4.36.0
gradio_client version: 1.0.1

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.2.0
fastapi: 0.109.0
ffmpy: 0.3.1
gradio-client==1.0.1 is not installed.
httpx: 0.26.0
huggingface-hub: 0.20.3
importlib-resources: 6.1.1
jinja2: 3.1.3
markupsafe: 2.1.4
matplotlib: 3.8.2
numpy: 1.26.3
orjson: 3.10.3
packaging: 23.2
pandas: 2.2.0
pillow: 10.2.0
pydantic: 2.5.3
pydub: 0.25.1
python-multipart: 0.0.9
pyyaml: 6.0.1
ruff: 0.3.0
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.12.3
typing-extensions: 4.9.0
urllib3: 2.1.0
uvicorn: 0.27.0
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2023.12.2
httpx: 0.26.0
huggingface-hub: 0.20.3
packaging: 23.2
typing-extensions: 4.9.0
websockets: 11.0.3
```

### Severity

Blocking usage of gradio

PR #8521, opened 2024-06-10
Ensure frontend functions work when they don't return a value

pngwn:
## Description

We now expect some kind of value to be passed into a 'submit' even if it is an empty array (no data). This normally happens even where there are no component inputs, the exception to this is when the frontend function returns nothing or a falsey value. Since the return value of the frontend function gets passed into the 'submit' call, this was causing an error. 
 
I have added a default value to the problematic function to address this, and added a few unit tests to prevent regressions.

Closes: #8495 

## 🎯 PRs Should Target Issues

## Tests


aliabid94:
The root problem is that `Payload.data` is supposed to be of `unknown[]`, but if the js function does not return anything, `Payload.data` gets set to `undefined`, which is not of type `unknown[]`. This will cause issues downstream, which currently is manifesting in `map_data_to_params`.  I believe the appropriate way to fix the overall problem would be to modify `process_frontend_fn` to return  `[]` when the result is undefined.

pngwn:
Yeah I can change that. I'm not sure what other downstream issues it would cause though. This has been the behaviour forever, it just hasn't been erroring until now.

aliabid94:
Actually its probably enough to just type Payload.data as `unknown[]  | undefined` and keep the parent logic. I just meant the reason that things broke is because the typing was wrong - Payload.data could be undefined as well - and this caused an issue downstream in `map_data_to_params`, but if we simply fix the typing, then the typechecker should also catch anywhere else we assumed it to be a list.

pngwn:
I think it's cleaner to always return an array from the frontend function, so I've added some logic for that. Should be good now.

aliabid94 (approved):
awesome LGTM!

Issue #7287, opened 2024-02-02
OpenAPI docs not working


jpaquim:
### Describe the bug

Getting this error when trying to get the API docs as described in [this comment](https://github.com/gradio-app/gradio/issues/4054#issuecomment-1555871838): `pydantic.errors.PydanticInvalidForJsonSchema: Cannot generate a JsonSchema for core_schema.IsInstanceSchema (<class 'starlette.requests.Request'>)`

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr


def greet(name, intensity):
    return "Hello " * intensity + name + "!"


demo = gr.Interface(fn=greet, inputs=["text", "slider"], outputs=["text"])

demo.launch(app_kwargs={"docs_url": "/docs"})
```

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Darwin
gradio version: 4.16.0
gradio_client version: 0.8.1

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.2.0
fastapi: 0.109.0
ffmpy: 0.3.1
gradio-client==0.8.1 is not installed.
httpx: 0.26.0
huggingface-hub: 0.20.3
importlib-resources: 6.1.1
jinja2: 3.1.3
markupsafe: 2.1.5
matplotlib: 3.8.2
numpy: 1.26.3
orjson: 3.9.12
packaging: 23.2
pandas: 2.2.0
pillow: 10.2.0
pydantic: 2.6.0
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0.1
ruff: 0.2.0
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.9.0
typing-extensions: 4.9.0
uvicorn: 0.27.0.post1
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2023.12.2
httpx: 0.26.0
huggingface-hub: 0.20.3
packaging: 23.2
typing-extensions: 4.9.0
websockets: 11.0.3
```

### Severity

I can work around it

abidlabs:
[...]

PR #8530, opened 2024-06-11
Fix request serialization for fastapi `/docs`

zhzLuke96:
This commit addresses a serialization issue in the `PredictBody` model that has been present since version 4.0. The `request` attribute, being of type `Request`, could not be serialized automatically, affecting FastAPI's auto-generated documentation (/openapi.json).
By adding a custom `__get_pydantic_json_schema__` method, the serialization schema is manually defined, ensuring correct documentation generation.
This is also one of the reasons why applications like "stable-diffusion-webui," which require /docs pages, have not updated to newer versions of Gradio.

Closes: #7287

### DIFF
--- a/gradio/data_classes.py
+++ b/gradio/data_classes.py
@@ -110,6 +110,24 @@ class PredictBody(BaseModel):
         None  # dictionary of request headers, query parameters, url, etc. (used to to pass in request for queuing)
     )
 
+    def __get_pydantic_json_schema__(self, source_type: Any, handler):
+        return {
+            "title": "PredictBody",
+            "type": "object",
+            "properties": {
+                "session_hash": {"type": "string"},
+                "event_id": {"type": "string"},
+                "data": {"type": "array", "items": {"type": "object"}},
+                "event_data": {"type": "object"},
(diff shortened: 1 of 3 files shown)


gradio/data_classes.py
```diff
@@ -110,6 +110,24 @@ class PredictBody(BaseModel):
         None  # dictionary of request headers, query parameters, url, etc. (used to to pass in request for queuing)
     )
 
+    def __get_pydantic_json_schema__(self, source_type: Any, handler):
```
freddyaboulton, line 113: Based on the [docs](https://docs.pydantic.dev/latest/concepts/json_schema/#implementing-__get_pydantic_core_schema__), this should be a class method!

zhzLuke96: Thank you, @freddyaboulton, for pointing that out! I've updated the method to be a class method as per the documentation. Please review the latest commit.

freddyaboulton (approved):

gradio/data_classes.py
freddyaboulton: Thanks - added a unit test

PR #8537, opened 2024-06-12
Many small fixes to website and docs

aliabd:
Makes a lot of small fixes to the docs that have been requested: 
- changes banner to remove the livestream date 
- replaces Sequence with list as a preprocessing step on all parameter docs 
- fixes capitlization issues with functions names on the docs navbar, and shows correct names for components (ImageEditor not Imageeditor) 
-  removes fake_gan as a demo for gr.Examples, and replaces it with calculator_blocks
- adds respective install from main commands for the python client and js client pages 
- fixes the upload_website_demos script to only update versioned demos when a new version is released and not on every commit

### DIFF
--- a/.changeset/true-places-post.md
+++ b/.changeset/true-places-post.md
@@ -1,6 +1,6 @@
 ---
-"gradio": minor
-"website": minor
+"gradio": patch
+"website": patch
 ---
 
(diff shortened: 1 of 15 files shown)


freddyaboulton (approved):
Thanks @aliabd !

PR #8543, opened 2024-06-12
Ability to disable orange progress animation for generators

freddyaboulton:
## Description

The orange animation is always displayed for a generator. You may want to disable this. For example, in a chatbot demo where the text stream is enough indication that progress is being made. In this PR, the orange animation is only shown for generators if `show_progress='full'`. Also added `show_progress` for `Interface` and `ChatInterface`. 

I thought there was an issue for this but I can't find it.

## 🎯 PRs Should Target Issues

## Tests


gradio/interface.py
```diff
@@ -732,6 +735,9 @@ async def cleanup():
                     batch=self.batch,
                     max_batch_size=self.max_batch_size,
                     concurrency_limit=self.concurrency_limit,
+                    show_progress=cast(
```
aliabid94, line 738: why is a cast necessary? isn't show_progress set to be of type `Literal["full", "minimal", "hidden"]`

aliabid94 (approved):
lgtm thanks! technically a visual breaking change but pretty minor

gradio/interface.py
freddyaboulton: pyright infers it as `str` instead of the narrower `Literal` so it complains. I think it's because the property is set in __init__ but it's being used in a separate method here.

freddyaboulton:
Thanks for the review @aliabid94 !

PR #8548, opened 2024-06-13
Fix reload mode by implementing `close` on the client

freddyaboulton:
## Description

Reload mode would break because the heartbeat could not be closed - 

Screenshot 2024-06-13 at 10 53 15 AM

You can test by with a stateful demo like ChatInterface

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/js/app/test/test_chatinterface.reload.spec.ts
+++ b/js/app/test/test_chatinterface.reload.spec.ts
@@ -1,0 +1,81 @@
+import { test, expect } from "@playwright/test";
+import { spawnSync } from "node:child_process";
+import { launch_app_background, kill_process } from "./utils";
+import { join } from "path";
+
+let _process;
+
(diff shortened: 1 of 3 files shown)


js/app/test/test_chatinterface.reload.spec.ts
pngwn, line 1: Don't we have an existing reload test?

freddyaboulton: This specific issue is triggered when there are state components so used a different demo/file for that case

pngwn: Ah cool.

pngwn (approved):
Looks great. Thanks @freddyaboulton!

freddyaboulton:
Thank you @pngwn 🙏

PR #8555, opened 2024-06-14
support html in chatbot

pngwn:
## Description

Closes: #(issue)

## 🎯 PRs Should Target Issues

## Tests


gradio/components/chatbot.py
```diff
@@ -185,10 +187,17 @@ def _preprocess_chat_messages(
         elif isinstance(chat_message, str):
             return chat_message
         elif isinstance(chat_message, ComponentMessage):
-            component = import_component_and_data(chat_message.component.capitalize())
+            capitalized_component = (
+                chat_message.component.upper()
+                if chat_message.component in ("json", "html")
+                else chat_message.component.capitalize()
+            )
+            component = import_component_and_data(capitalized_component)
```
pngwn, line 195: To handle. `gr.HTML` and `gr.JSON` which don't follow the same pattern as others.

gradio/components/chatbot.py
```diff
@@ -185,10 +187,17 @@ def _preprocess_chat_messages(
...
             if component is not None:
                 instance = component()  # type: ignore
-                if issubclass(instance.data_model, GradioModel):
+                if not instance.data_model:
+                    payload = chat_message.value
+                elif issubclass(instance.data_model, GradioModel):
```
pngwn, line 200: `issubclass` requires the first arg to be a class and if there is no `data_model` then this isn't the case, so handled this earlier.

gradio/components/chatbot.py
pngwn: We should probably think of a better way to handle this at some point.

aliabid94 (approved):
looks great, thanks pete!

PR #8565, opened 2024-06-16
Remove duplicated code in `routes.py`

sadra-barikbin:
## Description

Hi there!

To do a tiny refactor in `routes.py`.


abidlabs:
Thank you @sadra-barikbin for the fix!

abidlabs (approved):

Issue #7996, opened 2024-04-10
It is confusing to call render after gr.load


freddyaboulton:
### Describe the bug

If you want to embed a loaded gradio app into a parent Blocks context you HAVE to do the following:

1. Call gr.load outside the parent blocks context
2. Then call `.render()` in the parent context, in the location you want to embedded app to appear

I'm wondering if we can make calling `gr.load(...)` once within the parent context work as expected?

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

Call `gr.load()` inside a parent block without render. The embedded app won't trigger its dependencies because render was not called.

### System Info

```shell
4.26.0
```

### Severity

I can work around it

DeepeshKalura:
## Greetings
Hello @freddyaboulton , I am new open source in general and new to this repo, I would like to work on this issue.  
## Understanding
A way to when we call gr.load() inside a parent block it will trigger the dependencies without explicit tell to render.

Since, I never use gradio, I need some tips to get started and how can we can successfully resolve this bug. 
Currently just install code locally.

PR #8579, opened 2024-06-19
Allow `gr.load` to work inside `gr.Blocks` automatically

abidlabs:
Closes: https://github.com/gradio-app/gradio/issues/7996. Now both of these work:

```py
import gradio as gr

gr.load("abidlabs/en2fr", src="spaces").launch()
```

and 

```py
import gradio as gr

with gr.Blocks() as demo:
    gr.load("abidlabs/en2fr", src="spaces")
    
demo.launch()
```

### DIFF
--- a/.changeset/quick-dolls-admire.md
+++ b/.changeset/quick-dolls-admire.md
@@ -1,0 +1,5 @@
+---
+"gradio": minor
+---
+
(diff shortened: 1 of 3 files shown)


aliabd (approved):
Tested and works great! Thanks @abidlabs

abidlabs:
Thanks @aliabd!

Issue #6850, opened 2023-12-20
`gr.Plot` is rendered smaller in a hidden `gr.Tab`


hysts:
### Describe the bug

When `gr.Plot` is placed in `gr.Tab`, the graph is rendered smaller than expected in the second tab compared to the first tab even though the figures are the same.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

repro Space: https://huggingface.co/spaces/hysts-debug/gradio-plot-render-in-tab

```python
import gradio as gr
import numpy as np
import pandas as pd
import plotly.express as px

np.random.seed(0)

df = pd.DataFrame({"x": np.arange(100), "y": np.random.normal(size=100).cumsum()})
fig = px.line(df, x="x", y="y")

with gr.Blocks() as demo:
    with gr.Tabs():
        with gr.Tab(label="a"):
            gr.Plot(value=fig)
        with gr.Tab(label="b"):
            gr.Plot(value=fig)

if __name__ == "__main__":
    demo.queue().launch()
```

### Screenshot

https://github.com/gradio-app/gradio/assets/25161192/8514591a-8643-4117-97b1-60a57409c315

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Linux
gradio version: 4.11.0
gradio_client version: 0.7.3

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.2.0
fastapi: 0.105.0
ffmpy: 0.3.1
gradio-client==0.7.3 is not installed.
httpx: 0.25.2
huggingface-hub: 0.19.4
importlib-resources: 6.1.1
jinja2: 3.1.2
markupsafe: 2.1.3
matplotlib: 3.8.2
numpy: 1.26.2
orjson: 3.9.10
packaging: 23.2
pandas: 2.1.4
pillow: 10.1.0
pydantic: 2.5.2
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0.1
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.9.0
typing-extensions: 4.9.0
uvicorn: 0.24.0.post1
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2023.10.0
httpx: 0.25.2
huggingface-hub: 0.19.4
packaging: 23.2
typing-extensions: 4.9.0
[...]

Issue #5180, opened 2023-08-11
LinePlot text color


jadehardouin:
### Describe the bug

When I started changing the theme of my app, I encountered an issue with the color of the text that's inside the LinePlot. Because I put a dark background for the app, I put a light grey font color but this didn't apply this specific text. Does someone know how to change it? Or how it's instantiated? 

Thanks in advance!

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```
import gradio as gr
from gradio.themes.base import Base
from gradio.themes.utils.colors import Color
from gradio.themes.utils import colors, fonts, sizes

light_grey = Color(
    name="light_grey",
    c50="#e0e0e0",
    c100="#e0e0e0",
    c200="#e0e0e0",
    c300="#e0e0e0",
    c400="#e0e0e0",
    c500="#e0e0e0",
    c600="#e0e0e0",
    c700="#e0e0e0",
    c800="#e0e0e0",
    c900="#e0e0e0",
    c950="#e0e0e0",
)

class Style(Base):
    def __init__(
        self,
        *,
        primary_hue: colors.Color | str = light_grey,
        secondary_hue: colors.Color | str = light_grey,
        neutral_hue: colors.Color | str = light_grey,
        spacing_size: sizes.Size | str = sizes.spacing_md,
        radius_size: sizes.Size | str = sizes.radius_md,
        text_size: sizes.Size | str = sizes.text_md,
        font: fonts.Font
        | str
        | Iterable[fonts.Font | str] = (fonts.GoogleFont("Sora")),
        font_mono: fonts.Font
        | str
        | Iterable[fonts.Font | str] = (fonts.GoogleFont("Sora")),
    ):
        super().__init__(
            primary_hue=primary_hue,
            secondary_hue=secondary_hue,
            neutral_hue=neutral_hue,
            spacing_size=spacing_size,
            radius_size=radius_size,
            text_size=text_size,
            font=font,
            font_mono=font_mono,
        )
        super().set(
            background_fill_primary="#050f19", #The color of the background of blocks
            background_fill_secondary="#050f19",
            block_background_fill="#050f19", #The color of the background of blocks
            block_background_fill_dark="#050f19",
            
            border_color_primary="#050f19", #The border of a block such as dropdown 
            border_color_primary_dark="#050f19",
            
            link_text_color="#f0ba2d", #The color for links
            link_text_color_dark="#f0ba2d",
            
            block_info_text_color="ffffff",
[...]

Issue #2780, opened 2022-12-08
Add preset color schemes for gradio plots


freddyaboulton:
> We'll probably want to update the color schemes for visualisations in the future with a few pre-baked options for users. Data-viz colour schemes have some specific requirements/ demands that fall outside of typical 'design' issues. We can address that (and the duplication) later.

_Originally posted by @pngwn in https://github.com/gradio-app/gradio/pull/2764#discussion_r1043436605_

pngwn:
Some examples of different types of palette can be found here: https://carbondesignsystem.com/data-visualization/color-palettes/

I'll try to find some more resources to guide us as well. We need to consider color blindness etc when choosing good defaults.

aliabid94:
As of #8550, matches gr.themes for Altair plots. Still should do some matching for plotly.

abidlabs:
This is great! I don't think we want to do this for plotly/matplotlib/bokeh since people generate the plots using those libraries outside of gradio and then gradio simply displays them (we wouldn't want to change the color on the users) -- but it makes sense for our native plot components

PR #8580, opened 2024-06-19
Improved plot rendering

aliabid94:
Improved plot rendering. Main changes:
- Native plots did not fit block. Resizing only "stretched" canvas (distorting text as well), did not actually repaint the plot. Now native plots fit block and repaint to fit on resize.
- Plot uses theme values for styling (font, spacing, colors, etc.) so plots actually look native.
- Dev reload mode did not work with plots, fixed (by removing svelte-vega dependency).
- Cleaned up frontend code for plots.

BEFORE:
Recording 2024-06-19 at 13 00 32

AFTER:
Recording 2024-06-19 at 13 03 42


abidlabs:
Great work @aliabid94! Does this close any of these issues:

* https://github.com/gradio-app/gradio/issues/6850
* https://github.com/gradio-app/gradio/issues/5180
* https://github.com/gradio-app/gradio/issues/2780

requirements.txt
```diff
@@ -1,5 +1,5 @@
 aiofiles>=22.0,<24.0
-altair>=4.2.0,<6.0
+altair>=5.3.0,<6.0
```
abidlabs, line 2: Curious, why upgrade?

abidlabs:
Backend changes look good. Agreed UI looks much nicer this way. This _is_ a visual breaking change, so we should make sure 

* to mention this in the release notes (we can add a highlight to this changelog)
* We should ensure that that users can get the previous behavior (a fixed width) if they want. The obvious way would be to use the `width` parameter, but that parameter seems to have no effect. Note that the `height` parameter is indeed working for me.

gradio/components/bar_plot.py
```diff
@@ -126,6 +126,8 @@ def __init__(
         self.height = height
```
abidlabs, line 126: Lets add to the docstring for `width` that if no `width` is passed, then it takes up the full width of the container.

abidlabs: What happens if no `height` is passed? I'm actually not sure

js/plot/shared/Plot.svelte
```diff
@@ -1,235 +1,39 @@
 <script lang="ts">
 	//@ts-nocheck
-	import Plotly from "plotly.js-dist-min";
+	import PlotlyPlot from "./plot_types/PlotlyPlot.svelte";
+	import BokehPlot from "./plot_types/BokehPlot.svelte";
+	import AltairPlot from "./plot_types/AltairPlot.svelte";
+	import MatplotlibPlot from "./plot_types/MatplotlibPlot.svelte";
```
pngwn, line 6: Since we have split these into separate components, lets lazy load them with a dynamic `import` instead of a static import. That will improve the loading time of apps using `Plot`. No users use all three types of plot at the same time.

pngwn (commented):
Frontend looks good to me apart from my one comment

requirements.txt
aliabid94: actually not necessary, will revert. altair 4.0 has no type hinting which was annoying internally but this makes no difference for the end user and also, altair 5.0 is still allowed

Issue #8312, opened 2024-05-17
Return of older bug. Cannot update Dataset


Solomin0:
### Describe the bug

@abidlabs 
Unable to update the dataset object correctly. It seems like all the threads on this are closed or before 4.0 using the old .update methods. 
#2143 
Any help or workarounds would be great thanks.
### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr
def update(name):

    print(name)
    # new_examples = gr.Examples(["1","2","3"], inputs=[inp],label = "Saved Prompts")
    # return f"Welcome to Gradio, {name}!", new_examples
    return [["a"],["b"]]
with gr.Blocks() as demo:
    with gr.Row():
        inp = gr.Textbox(placeholder="What is your name?")

        saved_examples = gr.Dataset(components = ["text"], samples = [["This is my Favorite Prompt"]],label= "Saved Prompts" )

        saved_examples.click(update, inputs=[saved_examples], outputs=[saved_examples])
    # btn.click(fn=update, inputs=inp, outputs=[ex])

demo.launch()
```

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Windows
gradio version: 4.31.0
gradio_client version: 0.16.2

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.3.0
fastapi: 0.111.0
ffmpy: 0.3.2
gradio-client==0.16.2 is not installed.
httpx: 0.27.0
huggingface-hub: 0.23.0
importlib-resources: 6.4.0
jinja2: 3.1.4
markupsafe: 2.1.5
matplotlib: 3.8.4
numpy: 1.26.4
orjson: 3.10.3
packaging: 23.2
pandas: 2.2.2
pillow: 10.3.0
pydantic: 2.7.1
pydub: 0.25.1
python-multipart: 0.0.9
pyyaml: 6.0.1
ruff: 0.4.4
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.12.3
typing-extensions: 4.11.0
urllib3: 2.2.1
uvicorn: 0.29.0
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2024.3.1
httpx: 0.27.0
huggingface-hub: 0.23.0
packaging: 23.2
typing-extensions: 4.11.0
[...]

Issue #6807, opened 2023-12-15
Problem with Dataset in an Interface when pressing the clear button


uvsq22201695:
### Describe the bug

To be able to realize the bug, copy the code provided and run it. Once on the web page, perform these actions:

1. Enter a name (this is an example) in the TextBox
2. Press the “Submit” button, the Dataset updates.
3. Press the “Clear” button

The Dataset and Textbox will not be deleted. No more actions will be possible (for example, changing the value in Textbox and clicking again on "submit" will not change the dataset).

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

def greet(name):
    return [["Hello " + name + "!"]]


demo = gr.Interface(fn=greet, inputs=gr.Textbox(), outputs=gr.Dataset(components=["text"], samples=[]))

demo.launch(debug=True)
```

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Windows
gradio version: 4.9.0
gradio_client version: 0.7.2

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.2.0
fastapi: 0.105.0
ffmpy: 0.3.1
gradio-client==0.7.2 is not installed.
httpx: 0.25.2
huggingface-hub: 0.19.4
importlib-resources: 6.1.1
jinja2: 3.1.2
markupsafe: 2.1.3
matplotlib: 3.8.2
numpy: 1.26.2
orjson: 3.9.10
packaging: 23.2
pandas: 2.1.4
pillow: 10.1.0
pydantic: 2.5.2
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0.1
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.9.0
typing-extensions: 4.9.0
uvicorn: 0.24.0.post1
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2023.12.2
httpx: 0.25.2
huggingface-hub: 0.19.4
packaging: 23.2
typing-extensions: 4.9.0
websockets: 11.0.3
```

### Severity

I can work around it

abidlabs:
Interesting, perhaps the clear functionality is not supported by `gr.Dataset`

uvsq22201695:
[...]

Issue #6415, opened 2023-11-14
update gr.Dataset gets AttributeError: 'Textbox' object has no attribute 'proxy_url'


Billijk:
### Describe the bug

I tried to update samples of the Dataset component and it failed with the following error:
```
Traceback (most recent call last):
  File "/data1/conda/agents/lib/python3.9/site-packages/gradio/queueing.py", line 427, in call_prediction
    output = await route_utils.call_process_api(
  File "/data1/conda/agents/lib/python3.9/site-packages/gradio/route_utils.py", line 232, in call_process_api
    output = await app.get_blocks().process_api(
  File "/data1/conda/agents/lib/python3.9/site-packages/gradio/blocks.py", line 1525, in process_api
    result = await self.call_function(
  File "/data1/conda/agents/lib/python3.9/site-packages/gradio/blocks.py", line 1147, in call_function
    prediction = await anyio.to_thread.run_sync(
  File "/data1/conda/agents/lib/python3.9/site-packages/anyio/to_thread.py", line 33, in run_sync
    return await get_asynclib().run_sync_in_worker_thread(
  File "/data1/conda/agents/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 877, in run_sync_in_worker_thread
    return await future
  File "/data1/conda/agents/lib/python3.9/site-packages/anyio/_backends/_asyncio.py", line 807, in run
    result = context.run(func, *args)
  File "/data1/conda/agents/lib/python3.9/site-packages/gradio/utils.py", line 672, in wrapper
    response = f(*args, **kwargs)
  File "/data/codes/test/langchain/server/gradio_test.py", line 21, in change_dataset
    return gr.Dataset(components=["textbox"], visible=False)
  File "/data1/conda/agents/lib/python3.9/site-packages/gradio/component_meta.py", line 152, in wrapper
    return fn(self, **kwargs)
  File "/data1/conda/agents/lib/python3.9/site-packages/gradio/components/dataset.py", line 73, in __init__
    self.component_props = [
  File "/data1/conda/agents/lib/python3.9/site-packages/gradio/components/dataset.py", line 75, in <listcomp>
    component.get_config(),
  File "/data1/conda/agents/lib/python3.9/site-packages/gradio/components/base.py", line 212, in get_config
    config = super().get_config()
  File "/data1/conda/agents/lib/python3.9/site-packages/gradio/blocks.py", line 216, in get_config
[...]

PR #8581, opened 2024-06-19
fix dataset update

abidlabs:
Allows the samples in a `gr.Dataset` to be updated. Here's a quick example:

```py
import gradio as gr

def update(value):
    print(value)
    return value, gr.Dataset(samples=[["a"],["b"]])

with gr.Blocks() as demo:
    with gr.Row():
        textbox = gr.Textbox()
        dataset = gr.Dataset(components = ["text"], samples = [["This"]], label="Saved Prompts")
        dataset.click(update, inputs=[dataset], outputs=[textbox, dataset])

demo.launch()
```

Closes: https://github.com/gradio-app/gradio/issues/8312
Closes: https://github.com/gradio-app/gradio/issues/7282
Closes: https://github.com/gradio-app/gradio/issues/6415

Also fixes the issue where clearing an `Interface` that included a `Dataset` would cause the Interface to crash. 

Closes: https://github.com/gradio-app/gradio/issues/6807

### DIFF
--- a/gradio/flagging.py
+++ b/gradio/flagging.py
@@ -164,7 +164,7 @@ class CSVLogger(FlaggingCallback):
             ) / client_utils.strip_invalid_filename_characters(
                 getattr(component, "label", None) or f"component {idx}"
             )
-            if utils.is_update(sample):
+            if utils.is_prop_update(sample):
                 csv_data.append(str(sample))
             else:
                 data = (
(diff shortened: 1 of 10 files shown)


gradio/flagging.py
```diff
@@ -164,7 +164,7 @@ def flag(
             ) / client_utils.strip_invalid_filename_characters(
                 getattr(component, "label", None) or f"component {idx}"
             )
-            if utils.is_update(sample):
```
abidlabs, line 167: just a refactor for clarity

aliabd (approved):
Tested and everything lgtm! Thanks @abidlabs. docs look good too!

abidlabs:
Thanks for the prompt review @aliabd!

Issue #8423, opened 2024-05-31
Missing deploy_space_action.yaml file


danwahl:
### Describe the bug

The [deploy_space_action.yaml](https://github.com/gradio-app/gradio/pull/4033/files#diff-5fb2b972974afbead892dd23fe3bd2cbd90c6137af3af6d8b1909da4a1ff9aee) file appears to be missing in more recent versions. This file is the template used to create the Github Action:

https://github.com/gradio-app/gradio/blob/bc1eff12c1e9ca93bc44348913f06466fb888f9a/gradio/cli/commands/deploy_space.py#L16-L17

Manually adding the file to the specified directory (e.g. `/path/to/env/lib/python3.11/site-packages/gradio/cli/commands/`) seems to resolve the problem.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```
$ gradio deploy
Creating new Spaces Repo in '/path/to'. Collecting metadata, press Enter to accept default 
value.
Enter Spaces app title [spaces-app-title]: 
Enter Gradio app file [app.py]: 
Enter Spaces hardware (cpu-basic, cpu-upgrade, t4-small, t4-medium, zero-a10g, a10g-small, a10g-large, a10g-largex2, a10g-largex4, a100-large) [cpu-basic]: 
Any Spaces secrets (y/n) [n]: n
Create Github Action to automatically update Space on 'git push'? [n]: y
Enter branch to track [main]: 
...
FileNotFoundError: [Errno 2] No such file or directory: 
'/path/to/env/lib/python3.11/site-packages/gradio/cli/commands/deploy_space_action.yaml'
```

### System Info

```shell
$ gradio environment
Gradio Environment Information:
------------------------------
Operating System: Linux
gradio version: 4.31.5
gradio_client version: 0.16.4

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.3.0
fastapi: 0.111.0
ffmpy: 0.3.2
gradio-client==0.16.4 is not installed.
httpx: 0.27.0
huggingface-hub: 0.23.2
importlib-resources: 6.4.0
jinja2: 3.1.4
markupsafe: 2.1.5
matplotlib: 3.9.0
numpy: 1.26.4
orjson: 3.10.3
packaging: 24.0
pandas: 2.2.2
pillow: 10.3.0
pydantic: 2.7.2
pydub: 0.25.1
python-multipart: 0.0.9
[...]

PR #8588, opened 2024-06-20
move `deploy_space_action.yaml` to correct location

abidlabs:
fixes: https://github.com/gradio-app/gradio/issues/8423

test with repro in issue


pngwn (approved):
Ah, that was silly. Thanks for taking care of this @abidlabs!

Issue #5174, opened 2023-08-10
Document the image formats that are acceptable in the `Image` and `Gallery` components


vadimkantorov:
### Describe the bug

I'm interested in having animated small videos at display in Image and in Gallery (in gallery it's also useful to have an option to synchronize all GIFs and pad all GIFs to maximum number of frames).

We can often have a GIF as result of matplotlib FuncAnimation, it's simpler to manage than video. So it would be nice to have these GIFs displayed as video too. I can either provide these GIFs as paths or read the video frames somehow into an array. It can be nice to have the browser display render them as GIFs specifically. If there's many of them, it should create less load for the browser.

I found https://github.com/gradio-app/gradio/issues/577 but could not understand if displaying these GIFs is supported as Videos and whether I can provide a THW3 arrays anywhere to Image/Gallery items.

Thanks!

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

N/A

### System Info

```shell
N/A
```

### Severity

I can work around it

abidlabs:
Hi @vadimkantorov you can already display gifs using the `gr.Image()` and `gr.Gallery()` components. You can verify with a simple demo like this:

```py
import gradio as gr

with gr.Blocks() as demo:
    gr.Gallery([
        "example.gif",
        "example.gif"
    ])
    gr.Image("example.gif")
    
demo.launch()
```

abidlabs:
Am I misunderstanding something?

vadimkantorov:
Will these be rendered as GIFs by the browser? Or converted to videos?

I guess the remaining bit is being able to pass THWC and have it rendered either as GIF or as a video (depending on passed option).

abidlabs:
They will be rendered as gifs. If you'd like to convert to videos, you can use a library like `moviepy` to do so: https://github.com/gradio-app/gradio/issues/5174

Not sure what THWC is.

Yes, you can pass in image captions to the gallery, using a tuple structure: (img_path, img_caption).

vadimkantorov:
THWC meaning numpy array shape Time x Height x Width x Channels (Channels is usually 3)

One unrelated feedback - it would be great if button Enter by default was hooked to the orange default button "Submit"

vadimkantorov:
Maybe docs (https://www.gradio.app/docs/image, https://www.gradio.app/docs/gallery) should explicitly mention the formats that they support. E.g. is SVG supported? Is PDF supported (sometimes used for plots)?

[...]

Issue #1965, opened 2022-08-07
Handle GIFs correctly in Image component


AlicanAKCA:
### Describe the bug

Hello, I had tried to upload GIF. I encountered an issue. I saw that when I wanted to split the GIF file frame by frame, it only took the first frame. This situation occurs in both local and embedded environments. Here's the code:

```python
#fname is a filename giving by input. You can see the output below. 
 def GIF(fname,pixel_size):
    print(fname)
    gif = Image.open(fname)
    frames = []
    for i in range(gif.n_frames):
        gif.seek(i)
        frame = Image.new('RGB', gif.size)
        frame.paste(gif)
        frame = np.array(frame)
        frames.append(frame)
    print(len(frames))
```
```python
inputs_tab1 = [gr.inputs.Image(type='file', label="Image"),
               gr.Slider(4, 100, value=12, step = 2, label="Pixel Size")]
outputs_tab1 = [gr.Image(type="file",label="Front")]

gr.Interface(fn = GIF,
                    inputs = inputs_tab1,
                    outputs = outputs_tab1).launch()
```

Output:
```
'C:\Users\alica\AppData\Local\Temp\tmpk4__6naz.gif' # print(fname)
1 # print(len(frames)) 
```
As you can see, there is only one frame of GIF file. I didn't encounter this issue when I wasn't using the Gradio. But, I want to implement this code on Gradio for GIF processing. Gradio just taking the first frame of GIF.

### Is there an existing issue for this?

- [x] I have searched the existing issues

### Reproduction

You can try to upload GIF on [Huggingface Space](https://huggingface.co/spaces/Alican/pixera). But, It  will not return anything. You have to check out the logs.

### Logs

```shell
To create a public link, set `share=True` in `launch()`.
/tmp/tmp4ze2data.gif
1
```

### System Info

```shell
Gradio 3.1.0
```

### Severity

annoying

JunyaoHu:
hello, have you solved it?

ThePyceIsRight:
I am also curious if you figured this out.

AlicanAKCA:
No, I couldn't.

ThePyceIsRight:
> No, I couldn't. 

What if instead of creating a new image (frame) and pasting the original, you did:

frame = img.seek(i)
frame.convert('RGBA')
frame = np.frame

That's how I would do it. I still don't see why your method failed though.

ThePyceIsRight:
[...]

PR #8589, opened 2024-06-20
Handle GIFs correct in `gr.Image` preprocessing

abidlabs:
We can finally handle GIFs in `gr.Image` closing one of our oldest issues! Fixes: https://github.com/gradio-app/gradio/issues/1965, Fixes: https://github.com/gradio-app/gradio/issues/5174

Screen Recording 2024-06-20 at 3 33 32 PM

Also removes up some old methods that are not used anymore


aliabd (commented):
Hmm i'm still seeing a very weird issue.. 

https://github.com/gradio-app/gradio/assets/9021060/5496d748-a49b-466d-8586-50b0506cab41

vs main 

https://github.com/gradio-app/gradio/assets/9021060/f286220d-fd08-4266-b80b-893bc99dc14a

with just 

```python 
import gradio as gr
gr.Interface(lambda x:x, "image", "image").launch()
```

Maybe i'm not supposed to use defaults for gifs?

abidlabs:
Ok so basically by default, `gr.Image()` converts to a numpy array which is not valid for GIFs. For GIFS, you need to set `type` to either `filepath` or `PIL`. Let me add that to the docstrings

abidlabs:
Yeah hmm let me see

abidlabs:
Oh okay so what's happening is that when a PIL image is passed into a `gr.Image()` output component, the `format` parameter coerces it into the specified format, which is "webp" by default. So you'd need to do:

```py
import gradio as gr

demo = gr.Interface(lambda x:x, gr.Image(type="pil"), gr.Image(format="gif"))
    
demo.launch()
```

or you can keep it as a filepath, like this:

```py
import gradio as gr

demo = gr.Interface(lambda x:x, gr.Image(type="filepath"), "image")
    
demo.launch()
```

Slightly confusing, but I think its best we can do without a breakage

aliabd (approved):
Ok, i'm fine with it as is. Consider changing the docs a bit. You could even add a section on GIFs :) 

Thanks!

abidlabs:
Yeah I'll do that!

PR #8590, opened 2024-06-21
Fix multimodal chat look

aliabid94:
There was too much spacing before between the attach button and the textbox. Also looked in the file and we were using absolute positioning as well as percents for padding - should avoid both of these.

### DIFF
--- a/.changeset/fifty-peaches-hug.md
+++ b/.changeset/fifty-peaches-hug.md
@@ -1,4 +1,5 @@
 ---
+"@gradio/icons": minor
 "@gradio/multimodaltextbox": minor
 "gradio": minor
 ---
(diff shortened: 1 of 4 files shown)


pngwn:
Okay, so i notices two small thins:

- on a small screen (iPhone SE) the two buttons of the multimodal chatbot get squished.
- When using the multimodal textbox alone (no send button) the text has a large gap from the attach icon.

aliabid94:
> on a small screen (iPhone SE) the two buttons of the multimodal chatbot get squished.

Fixed.

> When using the multimodal textbox alone (no send button) the text has a large gap from the attach icon.

How do I remove the send button? `gr.MultimodalTextbox(submit_btn=None)` has no effect, on this branch or main.

abidlabs:
> How do I remove the send button? gr.MultimodalTextbox(submit_btn=None) has no effect, on this branch or main.

It looks like there's a bug, in the `interactive` version, setting `submit_btn=None` has no effect, but in the `static` version, it works:

image

image

PR #8593, opened 2024-06-21
Adding more docs for using components in chatbot

abidlabs:


gradio/components/chatbot.py
```diff
@@ -62,7 +62,7 @@ class Chatbot(Component):
     Also supports audio/video/image files, which are displayed in the Chatbot, and other kinds of files which are displayed as links. This
     component is usually used as an output component.
 
-    Demos: chatbot_simple, chatbot_multimodal
+    Demos: chatbot_simple, chatbot_multimodal, chatbot_core_components
```
aliabd, line 65: we don't currently support demos with any requirements on the docs, I can add support for it soon. But remove this for now as its erroring out on lite.

aliabd (commented):
Looks good! thanks @abidlabs

aliabd (approved):
sorry meant to approve

abidlabs:
Ah good catch. I'll make a simplified version of this demo without dependencies. Just to be clear, the problem is plotly, or is it the files?

aliabd:
Oh the files might be an issue as well, there are workarounds. But there's also another problem that we only used the release version of lite on the website, so this wouldn't work anyways until after release. I would just remove the demo for now and I can add it after. 

Or actually you could just embed it from spaces directly into the svx? Like just add the demo under the the new examples section.

abidlabs:
The new demo I'm using (`chatbot_core_components_simple`) should be fine, right? It has no external dependences or files.

aliabd:
@abidlabs yeah it will work. but only on release when lite gets updated. I think this is good for now

I'll open a PR to start using main wheel of lite on /main

PR #8603, opened 2024-06-25
Fix resizer on altair

aliabid94:
when using altair with gr.Plot, the resizer is not guaranteed to work. It only works when using native plots (`gr.LinePlot`, `gr.BarPlot`, `gr.ScatterPlot`). Made it so resizing logic only applies when the chart type is provided, meaning this is a native plot.

test with `altair_plot` and `native_plots` demos/

### DIFF
--- a/.changeset/few-camels-swim.md
+++ b/.changeset/few-camels-swim.md
@@ -1,0 +1,6 @@
+---
+"@gradio/plot": patch
+"gradio": patch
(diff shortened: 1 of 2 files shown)


abidlabs (approved):
lgtm! thanks for the quick fix

abidlabs:
let's get this in for the release

Issue #7976, opened 2024-04-09
Document `.on` trigger


apolinario:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
I could not find documentation on the `.on` trigger

**Describe the solution you'd like**  
Document the `.on` trigger as it's super helpful

abidlabs:
Yes we need to do this. cc @aliabd 

Related: https://github.com/gradio-app/gradio/issues/5072

Issue #7380, opened 2024-02-10
Document that `gr.EventData` can be used with any event


thiswillbeyourgithub:
### Describe the bug

As described [in another issue](https://github.com/gradio-app/gradio/issues/6278#issuecomment-1937018835)

I have several buttons in my app. By default I want them to be invisible except for the first one. Clicking on the first one called "Load buttons" should update the properties of the other buttons to change their names and make them visible.
The event of all those buttons is the same function. In this function I have `evt: gr.EventData` as arg to know which button was clicked as that has an influence on what the function will do.

When doing this I noticed that although the buttons now were visible and with the right name, evt still returned their old names.

Here's another way I explained it:

> In my example I have a button B that called function f.
Before f is called, B has value "UNDEFINED".
As soon as f is called: it returns gr.update(value="name") direclty to B.
So far so good: for the user the text appearing on B is now "name" after clicking.
But when clicking again on B, f now checks evt.target.value to see the name of the button that called it. I expected the value to be "name" but it was instead "UNDEFINED".
I ended up working around it by setting elem_id of my buttons so that I can keep track of what button called no matter their name but still this is a bug.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

def f(evt: gr.EventData, btn1, btn2):
    print(f"Name of the button clicked: {evt.target.value}")
    print(f"Actual name of the buttons: {btn1}, {btn2}")
    btn1 = gr.update(value="0")
    btn2 = gr.update(value="1")
    return [btn1, btn2]

with gr.Blocks() as demo:
    b1 = gr.Button(value="Undefined 1", elem_id="Undefined 1")
    b2 = gr.Button(value="Undefined 2", elem_id="Undefined 2")

    b1.click(fn=f, inputs=[b1, b2], outputs=[b1, b2])
    b2.click(fn=f, inputs=[b1, b2], outputs=[b1, b2])

demo.launch(inbrowser=True)
```

### System Info

```shell
gradio==4.18.0
```

### Severity

I can work around it

abidlabs:
Hi @thiswillbeyourgithub this issue is not really clear to me. The `gr.Button()` does not have any officially supported event that includes EventData. The `gr.Button` component's only supported event is `.click()` which carries the current value of the buttons. But it does not have any event like `.select()` which carries `EventData`. Passing in an argument of type `EventData` may do something but its not officially supported so the behavior is unpredictable. 

[...]

Issue #7348, opened 2024-02-07
More documentation around event chaining, especially for things like clearing and triggering other components


JohnDuncanScott:
[x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
The documentation is generally really good, but I'm struggling with chaining different events together. The concrete example I have is:
    `
chatbot = gr.Chatbot(height=300, render=False)
    textbox = gr.Textbox(container=False, scale=7, render=False)
    clear_button = gr.ClearButton([textbox, chatbot], render=False)

    system_prompt_dropdown = gr.Dropdown(
        choices=[prompt.value for prompt in SystemPromptType],
        value=default_system_prompt_type_value,
        label="Character type (sets description)",
        render=False,
    )
    system_prompt_dropdown.change(
        gradio_system_prompt_dropdown_change,
        inputs=system_prompt_dropdown,
        outputs=system_prompt_textbox,
    )
`

When the dropdown change event happens (i.e. the user selects a value), I would like to call the click event on the clear_button button programmatically so that it then clears textbox and chatbot. I can't seem to get this to work and I'm not sure what the correct way is of "calling" a component from another component, while also re-using the variables I've defined, if that makes sense?

So basically, on dropdown change, call clear button click. My Python is rusty and I'm just learning Gradio, so apologies if it's something obvious :).

**Describe the solution you'd like**  
More documentation.

dwipper:
@JohnDuncanScott An easy way to accomplish the clearing of the textbox (or anything other controls) is to clear the value of the textbox as part of the return from your gradio_system_prompt_dropdown_change function. it would look something like this:

```
def gradio_system_prompt_dropdown_change():
    return gr.Textbox(value="")
```

This sets the value of the gradio_system_prompt_dropdown_change output referenced in your .change event. You can also ask question like this on the Gradio Discord: https://discord.com/channels/879548962464493619/1025174734427656283

JohnDuncanScott:
Thanks for the response, I saw that, but it's not quite what I want I think. Each of those components have customisations, so the new component would need to match that. You'd need to create a factory method for each component just to initialise it and then re-create it back to being empty and keep the original customisation (unless I'm misunderstanding the documentation) (e.g. in your above example, I'm assuming the new textbox would lose the container=False, scale=7 customisation since it's replacing the original).

[...]

PR #8604, opened 2024-06-25
Add docs for `.on()`, `.then()`, and `.success()`, as well as the subclasses of `gr.EventData`

abidlabs:
Adds several classes and functions to the "Helpers" section of the documentation.

* Documents `.on()`

Closes: #7976 

* Documents the `Dependency` class with its `.then()` and `success()` methods

Closes: https://github.com/gradio-app/gradio/issues/5072
Closes: https://github.com/gradio-app/gradio/issues/7348

* Documents `gr.EventData` better as well as all of its subclasses explicitly

Closes: https://github.com/gradio-app/gradio/issues/7380

image

See: https://gradio-p2wvb6ymp-hugging-face.vercel.app/main/docs/gradio/likedata


js/_website/src/lib/templates/gradio/04_helpers/eventdata.svx
```diff
@@ -29,23 +29,27 @@ gradio.EventData(···)
...
+    textbox.select(on_select, textbox, statement)
+
+demo.launch()
 ```
 {/if}
 
 <!--- Initialization -->
-### Initialization
+### Attributes
 <ParamTable parameters={obj.parameters} />
```
aliabd, line 53: this is coming with 'required' which doesn't make sense here for an attribute. to fix just add  `obj.parameters[0]["kwargs"] = null` in the script section in the top.

aliabd (commented):
Thanks @abidlabs for adding all of these docs and closing long standing issues! 

A few notes: 
- you aren't currently documenting `.then` and `.success`. Did you mean to add them manually as attributes to `dependency.svx`? 
- can you use a demo without external dependencies for `LikeData`? `chatbot_multimodal` is breaking for this reason. You can try it out on https://www.gradio.app/playground first 
- since there are a lot of pages in the helpers section now, maybe ordering them makes sense? nbd but for example `DeletedFileData` mentions `EventData` which comes later. If it makes sense to you, you can prefix the important ones with 01_ , 02_ etc. It's a better experience when we highlight the important ones first.

abidlabs:
Thanks @aliabd for the review!

> you aren't currently documenting .then and .success. Did you mean to add them manually as attributes to dependency.svx?

as discussed in our sync just now, I've left them out because they are attributes but also functions, which make them hard to document automatically. instead, I've just described them in the docstring and included a code example for each

> can you use a demo without external dependencies for LikeData? chatbot_multimodal is breaking for this reason. You can try it out on https://www.gradio.app/playground first

will do

> since there are a lot of pages in the helpers section now, maybe ordering them makes sense? nbd but for example DeletedFileData mentions EventData which comes later. If it makes sense to you, you can prefix the important ones with 01_ , 02_ etc. It's a better experience when we highlight the important ones first

did not know, will do!

js/_website/src/lib/templates/gradio/04_helpers/01_eventdata.svx
```diff
@@ -0,0 +1,72 @@
...
+    import {get_object} from "../../process_json.ts";
+    import ParamTable from "$lib/components/ParamTable.svelte";
+    import ShortcutTable from "$lib/components/ShortcutTable.svelte";
+    import DemosSection from "$lib/components/DemosSection.svelte";
+    import FunctionsSection from "$lib/components/FunctionsSection.svelte";
+    import GuidesSection from "$lib/components/GuidesSection.svelte";
+    import CopyButton from "$lib/components/CopyButton.svelte";
+    import { style_formatted_text } from "$lib/text";
+
+    let obj = get_object("eventdata");
```
aliabd, line 12: I think you forgot to add the kwargs thing here 


Screen Shot 2024-06-25 at 7 59 43 PM

aliabd:
```suggestion
    let obj = get_object("eventdata");
    obj.parameters[0]["kwargs"] = null;
```

aliabd (approved):
Nice @abidlabs everything looks good apart from one comment!

abidlabs:
Thanks @aliabd for the great review!

PR #8608, opened 2024-06-25
Bugfix: Allow multiple files to be uploaded via multimodal textbox

freddyaboulton:
## Description

The docs of multimodal textbox mentions that it "allows for the uploading of multimedia files" but only one file can be uploaded. So I think this is a bugfix as opposed to a new feature. Updating the e2e tests as well as the typehint for `file_count` in the Upload component.

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/js/multimodaltextbox/shared/MultimodalTextbox.svelte
+++ b/js/multimodaltextbox/shared/MultimodalTextbox.svelte
@@ -161,6 +161,7 @@
 			for (let file of detail) {
 				value.files.push(file);
 			}
+			value = value;
 		} else {
 			value.files.push(detail);
 			value = value;
@@ -244,7 +245,7 @@
 		<Upload
 			bind:this={upload_component}
 			on:load={handle_upload}
-			filetype={accept_file_types}
+			file_count={"multiple"}
 			{root}
 			{max_file_size}
 			bind:dragging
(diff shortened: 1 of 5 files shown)


abidlabs:
@freddyaboulton it looks like `file_count` is not a parameter on the Python side? We should add that and make the default "single" to be consistent with the other file components

freddyaboulton:
Ah I thought it should always be "multiple" based on the docs but agreed the consistency would be nice. We can get this in after the release then.

freddyaboulton:
Added `file_count`! cc @dawoodkhan82

js/multimodaltextbox/shared/MultimodalTextbox.svelte
```diff
@@ -244,7 +246,7 @@
 		<Upload
 			bind:this={upload_component}
 			on:load={handle_upload}
-			filetype={accept_file_types}
```
dawoodkhan82, line 247: @freddyaboulton was this removed by accident?

dawoodkhan82:
Tested and works great, thanks for adding @freddyaboulton

js/multimodaltextbox/shared/MultimodalTextbox.svelte
freddyaboulton: Yes, let me add back!

freddyaboulton: Fixed

dawoodkhan82 (approved):

freddyaboulton:
Thanks @dawoodkhan82 !

Issue #8615, opened 2024-06-25
Critical Accessibility Bug: gr.Chatbot and gr.ChatInterface is broken in v4.37.1.


chigkim:
### Describe the bug

Screen reader cannot read the message anymore.
It just says user's message and bot's message without the actual content.
It worked in previous version.
The embedded examples below on the documentation doesn't work either.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

https://www.gradio.app/guides/creating-a-chatbot-fast

https://www.gradio.app/guides/creating-a-custom-chatbot-with-blocks

### Logs

```shell
No log.
```

### System Info

```shell
4.37.1
```

### Severity

Blocking usage of gradio

abidlabs:
@hannahblair could you take a look into this?

PR #8631, opened 2024-06-26
Ensure chatbot messages are visible to screenreaders

hannahblair:
## Description

The logic for checking whether the message was of text type wasn't working and thus the `message.value` of the aria label wasn't being added to the string. I've tweaked/added a bit of logic to specify the file extension if the message is a file and to specify the component type if it's a component. 

Closes: #8615 

## 🎯 PRs Should Target Issues

## Tests


js/chatbot/shared/ChatBot.svelte
```diff
@@ -195,6 +195,23 @@
...
+		if (message.type === "text") {
+			return message.value;
+		} else if (message.type === "component") {
+			return `a component of type ${message.component}`;
+		} else if (message.type === "file") {
+			if (Array.isArray(message.file)) {
+				return `file of extension type: ${message.file[0].orig_name?.split(".").pop()}`;
+			}
+			return (
+				`extension: ${message.file?.orig_name?.split(".").pop()}` +
```
freddyaboulton, line 209: Should this also `file of extension type: ... with name .... `? Just so that it's more consistent with the array case.

freddyaboulton (commented):
Makes sense to me! Will leave the approval to the chatbot masters though

pngwn (approved):
Ah yes, same issue as the copy button thing. Thanks for taking care of this @hannahblair!

abidlabs:
Thanks @hannahblair @pngwn @freddyaboulton, let's merge this in so that we can get a patch release out.

Issue #8644, opened 2024-06-27
FileExplorer.height can be str type


magical-l:
### Describe the bug

`FileExplorer.height` can be str type in fact, but `str` is not in declarelation: `height: int | float | None = None`

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

None

### System Info

```shell
None
```

### Severity

I can work around it

abidlabs:
Thanks @magical-l would you like to open a PR to fix this?

magical-l:
@abidlabs I created a pull request(edit from browser), but I cannot clone my forked repo to my local to execute the script(looks like some network issue, same with yesterday)😅

PR #8645, opened 2024-06-27
Update file_explorer.py for str type `height`

magical-l:
add `str` type to `FileExplorer.height` declaration

Closes: #8644


abidlabs:
Perfect thanks @magical-l!

abidlabs (approved):

PR #8649, opened 2024-06-27
ensure `File` objects are handled in JS client `handle_file`

hannahblair:
## Description

It seems we don't actually need the extra metadata in an object when we pass an object of type `File` to `handle_file`. We can just pass the `File` object itself to the client and it'll get uploaded.

Closes: #(issue)

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/little-falcons-chew.md
+++ b/.changeset/little-falcons-chew.md
@@ -1,6 +1,6 @@
 ---
-"@gradio/client": minor
-"gradio": minor
+"@gradio/client": patch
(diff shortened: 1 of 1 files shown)


freddyaboulton (approved):
Thanks @hannahblair !

pngwn (approved):
This is perfectly fine. As long as we have a blob or file as the data we don't need to process it into a Command because `handle_blibs` deals with that. 

Thanks @hannahblair!

Issue #8201, opened 2024-05-03
Allow Gradio apps to take up more (full screen?) width


abidlabs:
A common request is to increase the width of Gradio apps so that there's more space for the components and less white space around the edges of the app. While this can be done with custom css, it'd be good to have a built-in parameter in Gradio for this:

Quoting @yvrjsharma:

> This is particularly relevant for arena-type demos, where multiple (>= two) chatbots are present in the UI, allowing for easy comparisons. These types of demos have become quite popular recently, with the success of lmsys-arena, and I think we might see an increase in their numbers in the near future. For example: https://huggingface.co/spaces/WildVision/vision-arena

one-lithe-rune:
In case it's useful to anyone trying to do this, the code I'm currently using make things full width for Blocks is similar to this:

```Python
import gradio as gr

# Change the @media query to whatever min-width you need

head_style = """
<style>
@media (min-width: 1536px)
{
    .gradio-container {
        min-width: var(--size-full) !important;
    }
}
</style>
"""

# It has to go in head=

with gr.Blocks(
    title="A chat interface",
    head=head_style,
) as chat_interface:

# ...etc.

```

Note that it has to go in head, as putting it in CSS didn't work on the version of gradio I was working with at the time I was trying to do this, at least with the `@media` query.

Issue #6732, opened 2023-12-10
svelte max-width property should be 100% instead of 1024px


thiswillbeyourgithub:
### Describe the bug

I notice that the max-width of the svelte app seems to be 1024px. I think it should be 100% as it renders much more nicely especially on mobile.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

css = """
.app.svelte-1kyws56.svelte-1kyws56 { max-width: 100%; }
"""


with gr.Blocks(css=css) as demo:
    with gr.Column():
        b = gr.Button("click me")
demo.launch()

```

### System Info

```shell
gradio==4.8.0
python 3.9.10
```

### Severity

I can work around it

pngwn:
I don't think is the correct approach as the screen will be very wide on large screens.

Do you have a screenshot showing the mobile improvements?

cc @hannahblair

thiswillbeyourgithub:
On my computer with 100% : 
![image](https://github.com/gradio-app/gradio/assets/26625900/96c6a285-987f-4dc6-af25-a57480c747cc)

On my computer with default value:
![image](https://github.com/gradio-app/gradio/assets/26625900/294955c1-704f-41a2-97ce-71566594c5ff)

On my tablet with default value:
![image](https://github.com/gradio-app/gradio/assets/26625900/5683e0cb-2eb2-45c7-a643-ea93baa565d3)

On my tablet with 100%:
![image](https://github.com/gradio-app/gradio/assets/26625900/2b033345-471d-4a31-bcb6-2a1af4f7e6e0)


The difference does not seem much but I'm building a pretty complicated UI and losing 25% of the width for no reason is actually problematic, it's squeezing my audio components, making the buttons hard to click etc.

I agree that 100% can be an issue for some screens but a middle ground seems better. Something a bit more adaptative that takes into account vertical screens for example. Or simply using an optionnal argument for the screen width?

thiswillbeyourgithub:
Currently on gradio 4.11 and I think this has gotten worse. My css workaround does not seem to work anymore and I'm stuck with the app appearing "zoomed out" on mobile.

What kind of arguments do you recommend for block.launch()?

PR #8651, opened 2024-06-27
Allow app to fill width

aliabid94:
Useful for dashboards and other complex apps that want to use all the screen space.

Closes: https://github.com/gradio-app/gradio/issues/8201
Closes: https://github.com/gradio-app/gradio/issues/6732


abidlabs:
Niceee I just needed this for the app I'm building 😆

abidlabs:
Two quick comments:

(1) I think `.launch()` would be a better place to put this parameter. I know we put a lot of global styling features in `gr.Blocks` like `css` and `theme` but this doesn't make sense since `Blocks` can be nested within other `Blocks` and only the style parameters in the outer `Blocks` apply.

(2) Not a strong opinion, but instead of a boolean, should we make this a more flexible parameter that can take in a specified width e.g. full width corresponds to 100%. And the default setting could be a tuple of ints corresponding to the current widths at different screen breakpoints

aliabid94:
> (1) I think .launch() would be a better place to put this parameter.

We already have fill_height in gr.Blocks() though

> Not a strong opinion, but instead of a boolean, should we make this a more flexible parameter that can take in a specified width e.g. full width corresponds to 100%. And the default setting could be a tuple of ints corresponding to the current widths at different screen breakpoints

I don't think that's necessary until we see a request for it, seems like overkill.

abidlabs:
> We already have fill_height in gr.Blocks() though

Ah good point, ok then

> I don't think that's necessary until we see a request for it, seems like overkill.

Hmm but would be a breaking change if we do get requests for it.

freddyaboulton (approved):
Nice - also agree Blocks init is the best place for the parameter

abidlabs:
After sleeping on it, I do think the `fill_width` parameter looks fine as a boolean and is symmetric with the `fill_height` parameter

2 things:

* This parameter doesn't seem to have any effect on Spaces: compare https://huggingface.co/spaces/abidlabs/fill-width and https://huggingface.co/spaces/abidlabs/fill-width-2 -- I think its due to the fact that Spaces adds its own margin around the iframe. Can we work with Spaces to fix that?
* Can we add `fill_height` and `fill_width` to `Interface` and `ChatInterface` as well?

pngwn (approved):
Looks great. Thanks @aliabid94!

aliabid94:
> On very wide screens, this will cause Gradio apps to look very stretched out and distorted. Instead of a boolean parameter, do you think we should set a higher max width instead of 100%

There is already a max-width set by gradio's container that's meant to prevent this. When you set full_width to True, you are specifically asking to skip this, so we should take all the width in this case.

> This parameter doesn't seem to have any effect on Spaces: compare https://huggingface.co/spaces/abidlabs/fill-width and https://huggingface.co/spaces/abidlabs/fill-width-2 -- I think its due to the fact that Spaces adds its own margin around the iframe. Can we work with Spaces to fix that?

I'll message Spaces team

> Can we add fill_height and fill_width to Interface and ChatInterface as well?

Done.

Issue #8006, opened 2024-04-11
Chatbot component freeze when using specific LaTeX delimiters


GaiZhenbiao:
### Describe the bug

The issue I'm encountering is reminiscent of the one documented in https://github.com/gradio-app/gradio/issues/7651. Specifically, when setting certain LaTeX delimiters (such as `\[` and `\]`, or `$`) within the Chatbot component, it ceases to function, accompanied by a JavaScript error identical to the one described in the referenced issue.

I think this malfunction likely stems from a conflict between the JavaScript responsible for markdown rendering and that for LaTeX rendering, as both systems utilize similar notation (e.g., `[xxx]` in Markdown). Interestingly, using `$` as a LaTeX delimiter is expected to be conflict-free, I don't know what could go wrong. Besides, messages incorporating `\[` and `\]` consistently trigger this bug 100% of the time, whereas `$` results in the bug approximately 50% of the time.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import json
import time

import gradio as gr

with open("chatbots3.json", "r") as f:
    data3 = json.load(f)

with open("chatbots4.json", "r") as f:
    data4 = json.load(f)

def update_chatbot3():
    for d in data3:
        time.sleep(0.005)
        yield d

def update_chatbot4():
    for d in data4:
        time.sleep(0.005)
        yield d

latex_delimiters_set = [
        # {"left": "$$", "right": "$$", "display": True},
        # {"left": "\\(", "right": "\\)", "display": False},
        {"left": "\\[", "right": "\\]", "display": True},
        {"left": "$", "right": "$", "display": False},
    ]

with gr.Blocks() as demo:
    chatbot = gr.Chatbot(
        latex_delimiters=latex_delimiters_set,
    )

    trigger_btn3 = gr.Button("Trigger Chatbot Bug [] (100% triggers bug)")
    trigger_btn4 = gr.Button("Trigger Chatbot Bug $$ (~50% triggers bug)")
    trigger_btn3.click(update_chatbot3, [], chatbot)
    trigger_btn4.click(update_chatbot4, [], chatbot)

demo.queue().launch()
```
[chatbots3.json](https://github.com/gradio-app/gradio/files/14949005/chatbots3.json)
[chatbots4.json](https://github.com/gradio-app/gradio/files/14949006/chatbots4.json)

### Screenshot

[...]

PR #8658, opened 2024-06-28
Chatbot LaTeX Crash Fix

dawoodkhan82:
## Description
With this fix, it seems to be crashing a lot less. Also doesn't seem like the Latex is being rendered correctly anymore. 

Closes: #8006

## 🎯 PRs Should Target Issues

## Tests


dawoodkhan82:
Can't seem to figure out this bug. This fix helps, but doesn't fully resolve the error: `Failed to execute 'insertBefore' on 'Node': The node before which the new node is to be inserted is not a child of this node.` Definitely something with the order of how things are rendered in the DOM.

@pngwn if you get a chance, could use a second set of eyes.

js/markdown/shared/MarkdownCode.svelte
```diff
@@ -73,10 +73,26 @@
 			}
 		}
 	}
-	afterUpdate(() => render_html(message));
+	onMount(() => {
+		el = document.getElementById("container") as HTMLSpanElement;
+	});
```
pngwn, line 78: This is essentially the same as the binding, does just having the condition in the `afterUpdate` address this issue?

pngwn (approved):
Ah interesting! Just one small comment but otherwise looks great. Thanks @dawoodkhan82!

js/markdown/shared/MarkdownCode.svelte
dawoodkhan82: Nope, it still crashes half the time. Seems like KaTex doesn't play well with svelte since both manipulate the DOM.

js/markdown/shared/MarkdownCode.svelte
```diff
@@ -73,15 +73,18 @@
...
+		}
+	});
 </script>
 
 <span class:chatbot bind:this={el} class="md" class:prose={render_markdown}>
-	{#if render_markdown}
-		{@html html}
-	{:else}
-		{html}
-	{/if}
```
dawoodkhan82, line 84: do we need this conditional? Removing this fixes the error. And we already handle the `render_narkdown` check in `process_message`

pngwn: its actually wrong anyway i think even if we did want to do it. Safe to remove.

Issue #8675, opened 2024-07-01
Error when passing in a custom chatbot to add like/dislike feature to Chatinterface


sajjadmosaheb:
how can I add a like/dislike function to Chatinterface?
this is possible with blocks but I can't add to a Chatinterface.

import gradio as gr

def greet(history, input):
return history + [(input, "Hello, " + input)]

def vote(data: gr.LikeData):
if data.liked:
print("You upvoted this response: " + data.value)
else:
print("You downvoted this response: " + data.value)

with gr.Blocks() as demo:
chatbot = gr.Chatbot()
textbox = gr.Textbox()
textbox.submit(greet, [chatbot, textbox], [chatbot])
chatbot.like(vote, None, None) # Adding this line causes the like/dislike icons to appear in your chatbot

demo.launch()

abidlabs:
Hi @sajjadmosaheb to do this, you can pass in a custom `gr.Chatbot` object into the `chatbot` argument of `gr.ChatInterface` as described here: https://www.gradio.app/guides/creating-a-chatbot-fast#customizing-your-chatbot, and then attach a `.like` event to it.

sajjadmosaheb:
Hi @abidlabs ,
but when I attach the .like() event like below I got this error:

**AttributeError: Cannot call like outside of a gradio.Blocks context.**

```
demo= gr.ChatInterface(main,
    title="test",
    css=css,
    fill_height=True,
    chatbot=gr.Chatbot(rtl=True,likeable=True).like(vote, None, None))

demo.launch() 
```

abidlabs:
Hi @sajjadmosaheb, you'll need to put everything under a `gr.Blocks`:

```py
with gr.Blocks():
  chatbot = gr.Chatbot(rtl=True,likeable=True)
  chatbot.like(vote, None, None)
  demo= gr.ChatInterface(main,
      title="test",
      css=css,
      fill_height=True,
      chatbot=chatbot)

demo.launch() 
```

sajjadmosaheb:
thanks @abidlabs and I appreciate your response,
now I got this error! (I even add the render=False in the first chatbot but not working)
**DuplicateBlockError: A block with id: 174 has already been rendered in the current Blocks.**

abidlabs:
Hmm let me reopen and investigate this

abidlabs:
For now, this should work, but looking into a better way of supporting this:

```py
import gradio as gr

with gr.Blocks() as demo:
[...]

PR #8677, opened 2024-07-01
Allow supplying custom `gr.Chatbot` with events to `gr.ChatInterface`

abidlabs:
Fixes: https://github.com/gradio-app/gradio/issues/8675 so that now you can put a `gr.ChatInterface` inside a `gr.Blocks` with a custom `gr.Chatbot` and attach events, like this:

```py
import gradio as gr

def vote(data: gr.LikeData):
    print(data.value)
    if data.liked:
        print("You upvoted this response: " + data.value["value"])
    else:
        print("You downvoted this response: " + data.value["value"])

with gr.Blocks() as demo:
    chatbot = gr.Chatbot()
    chatbot.like(vote, None, None)
    gr.ChatInterface(fn=lambda x,y:x, chatbot=chatbot)
    
demo.launch()
```


freddyaboulton (approved):
Nice fix @abidlabs !

guides/05_chatbots/01_creating-a-chatbot-fast.md
```diff
@@ -136,6 +136,25 @@ gr.ChatInterface(
 
 The placeholder appears vertically and horizontally centered in the chatbot.
 
+If you would like to attach event listeners to your custom chatbot, wrap the chatbot as well as the `gr.ChatInterface` inside of a `gr.Blocks` like this:
```
aliabid94, line 139: feel like the "creating a chatbot fast" guide should be quick and sweet, this particular usecase doesn't seem common enough to be here.

aliabid94 (approved):
LGTM!

guides/05_chatbots/01_creating-a-chatbot-fast.md
abidlabs: That's fair, this guide is getting quite bloated. It would be better to move this over to the docs page for `gr.ChatInterface`. Let me do that

abidlabs:
Thanks folks!

PR #8683, opened 2024-07-02
Warn against Falsy credentials

Paillat-dev:
Allowing Falsy auth credentials will cause problems later, for example in `hmac.compare_digest(input_password.encode(), correct_password.encode())` as NoneType has no `encode` method. Moreover, allowing empty passwords would make no sense from a security point of view.

Let me know if any changes are required.

### DIFF
--- a/.changeset/chubby-spiders-love.md
+++ b/.changeset/chubby-spiders-love.md
@@ -1,5 +1,5 @@
 ---
-"gradio": minor
+"gradio": patch
 ---
 
 feat:Warn against Falsy credentials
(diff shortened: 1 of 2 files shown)


abidlabs:
Thanks @Paillat-dev I would suggest raising a warning when a none or empty credential is passed in instead of an error -- a lot of times, users are simply prototyping quickly with auth and we don't want to break those workflows

Paillat-dev:
Sure, I changed it.

abidlabs (approved):
LGTM @Paillat-dev thanks for the contribution!

Issue #8668, opened 2024-06-29
Can you explain which library you use to read 'ply' files into a 3D model?


NTNPhim:
- [ ] I have searched to see if a similar issue already exists.

I couldn't find a library that helps display the 'ply' file as a 3D model like a demo. Some parameters, such as mesh or color, cannot be read. If you have any example code to read a 'ply' file, please share it. Thank you so much!

abidlabs:
Hi @NTNPhim can you please provide a minimal code example that we can use to reproduce the issue above, including the file you are trying? See: https://stackoverflow.com/help/minimal-reproducible-example

NTNPhim:
This is my code:

'''
import open3d as o3d
import numpy as np

pcd = o3d.io.read_point_cloud("hand4.ply")
o3d.visualization.draw_geometries([pcd])
 '''
I attach picture of point cloud,
<img width="962" alt="Screenshot 2024-07-01 at 22 08 08" src="https://github.com/gradio-app/gradio/assets/104830358/8c2c024f-c8f0-4f4f-83d6-2def89f972e6">

This is link of ply file: https://drive.google.com/file/d/1bWCP87hFfWlbN3S1RdB-cxGA7W_5ANBq/view?usp=sharing
Thank you!

Issue #8459, opened 2024-06-05
Show Point Cloud in Model3D


buaacyw:
- [√ ] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
People sometimes need to visualize point cloud in 3D generation pipeline.

**Describe the solution you'd like**  
Adding Point Cloud visualization support in Model3D. The file Format might be ply. Support save format like Trimesh.PointCloud from [here](https://trimesh.org/trimesh.html#trimesh.PointCloud).

jytime:
Hi @buaacyw @abidlabs 

I am making a demo which is related to this. May I ask if there is any update on this?

What I am doing now is using trimesh.Scene() to load a point cloud, save it as a .glb file and pass it to gradio.Model3D. The problem here is, the visualised points are too small and usually hard to distinguish. It can be checked at this hugging face demo https://huggingface.co/spaces/facebook/vggsfm.

I attempted to enhance visibility by defining a sphere for each 3D point:
```
        point_size = 0.05  

        for point, color in zip(points3D, points3D_rgb):
            sphere = trimesh.creation.icosphere(subdivisions=2, radius=point_size)
            sphere.apply_translation(point)
            sphere.visual.vertex_colors = [color] * len(sphere.vertices)
            glbscene.add_geometry(sphere)
```
However, this method is too slow, especially for point clouds with over 1M points.

It would be much appreciated if Gradio could support better visualization of point clouds, as it is a common use case in the 3D community. Thank you :)

dawoodkhan82:
@buaacyw @jytime Releasing soon! https://github.com/gradio-app/gradio/pull/8687

Issue #8447, opened 2024-06-04
Showing Mesh Wireframe in Model3D


buaacyw:
- [√] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
Sometimes users need to view the wireframe of mesh

**Describe the solution you'd like**  
Provide an option to show wireframe of mesh in Model3D

**Additional context**  
Recently there are some works focus on generating better mesh wireframes. (https://nihalsid.github.io/mesh-gpt/). Personally, I'm trying to use gradio to show my new work but find it impossible to render mesh wireframe.

abidlabs:
Hi @buaacyw is there a particular file format / file extension for mesh wireframes that you need support for?

buaacyw:
Thanks for your quick reply!
Don't need to be a particular file format. Any mesh file format should be able to easily showing its wireframes.
The result should be like this (a figure from MeshGPT):
![image](https://github.com/gradio-app/gradio/assets/52091468/b86b5979-ea25-40e2-a26c-fa0e7606d24b)

buaacyw:
I believe this is quite easy to implement since you only need to highlight the edges.

abidlabs:
It looks like this is supported in babylonjs, so it should be quite doable: https://www.babylonjs-playground.com/#QCU8DJ#22

We don't have a lot of bandwidth at the moment, but if a community contributor wants to take this on, happy to review PRs

PR #8687, opened 2024-07-03
Model3D point cloud and wireframe display modes

dawoodkhan82:
## Description
Adds `display_mode` param to `gr.model3D()`.  `default`, `point_cloud`, and `wireframe` options.

Screenshot 2024-07-02 at 8 47 49 PM

Closes: #8447
Closes: #8459
Closes: https://github.com/gradio-app/gradio/issues/8668

## 🎯 PRs Should Target Issues

## Tests


gradio/components/model3d.py
```diff
@@ -60,6 +61,7 @@ def __init__(
         """
         Parameters:
             value: path to (.obj, .glb, .stl, .gltf, .splat, or .ply) file to show in model3D viewer. If callable, the function will be called whenever the app loads to set the initial value of the component.
+            display_mode: the display mode of the 3D model in the scene. Can be "default", "point_cloud", or "wireframe".
```
abidlabs, line 64: Maybe instead of "default", we call it "solid" for clarity? I'm not sure if that's the standard terminology (cc @dylanebert). If not, at least we should explain in the docstring:
```suggestion
            display_mode: the display mode of the 3D model in the scene. Can be "default" (which renders the model as a solid object), "point_cloud", or "wireframe".
```

abidlabs (approved):
Nice @dawoodkhan82! LGTM works great. I just left one nit above.

Also, I'm guessing that the "wireframe" and "point_cloud" display modes have parameters that can be used to configure e.g. the density of the rendered objects? If so, we might want to expose some of those parameters, but we can do so later if users request it

gradio/components/model3d.py
dawoodkhan82: Just asked chatgpt, it agrees `solid` is better

Issue #8704, opened 2024-07-05
An environment variable for setting allowed_paths and blocked_paths


cocktailpeanut:
[O] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  

a lot of AI applications use up a lot of disk space and many users end up using symbolic links (or junctions) to link files (such as model files) to another drive. Note that these actions are not hard coded into the app, but each user takes care of the linking outside of the apps. This means, these apps will break when they link files to a path outside of the gradio app.

The existing solution is for the app to declare `allowd_paths` when calling `launch()` so that the specified paths will be allowed despite being outside of the gradio app root.

However, this is not ideal because the user needs to modify the source code to pass the custom `allowed_paths` array.

**Describe the solution you'd like**  

By supporting `GRADIO_ALLOWED_PATHS` (also implemented `GRADIO_BLOCKED_PATHS` for the sake of completeness), the users don't need to touch the code, but simply set the custom `GRADIO_ALLOWED_PATHS` environment variable.

**Additional context**  

Here are some users struggling with this issue, to show you that this is a real issue https://github.com/butaixianran/Stable-Diffusion-Webui-Civitai-Helper/issues/71

As you can see, the Stable Diffusion Web UI project has already kind of implemented a workaround by utilizing launch flags, but we can't expect every gradio app to implement the custom allowed_paths feature in code. By supporting the environment variable in addition to the existing approach, all apps built with gradio can let the end users easily use files outside of the gradio path if they want, WITHOUT having to touch the code.

PR #8705, opened 2024-07-05
GRADIO_ALLOWED_PATHS & GRADIO_BLOCKED_PATHS comma separated environme…

cocktailpeanut:
## Description

Currently gradio only allows for custom `allowed_paths` and `blocked_paths` programmatically via the `launch()` args. But as mentioned in the linked issue, this requires end users to modify the code, which is not convenient.

Supporting an environment variable based alternative should let users use external drives and file paths when using ANY gradio app, WITHOUT having to change the source code.

The approach here is to take a comma separated string (since environment variables are strings only), split them by commas, trim the whitespace, and use the result array as `allowed_paths` and `blocked_paths`.

Closes: https://github.com/gradio-app/gradio/issues/8704


abidlabs:
Thanks so much @cocktailpeanut for this PR. Generally looks good to me. Could you also add documentation for these env variables here: https://github.com/gradio-app/gradio/blob/main/guides/03_additional-features/08_environment-variables.md

abidlabs (approved):
lgtm!

Issue #8718, opened 2024-07-08
Cannot do custom component with MultimodalTextbox


loganrenaud:
### Describe the bug

When I launch gradio cc create MultimodalTextbox --template MultimodalTextbox I have the issue FileNotFoundError: [Errno 2] No such file or directory: 
'/home/logan/mon_env/lib/python3.10/site-packages/gradio/components/m
ultimodaltextbox.py'

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

gradio cc create MultimodalTextbox --template MultimodalTextbox

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Linux
gradio version: 4.37.2
gradio_client version: 1.0.2

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.3.0
fastapi: 0.111.0
ffmpy: 0.3.2
gradio-client==1.0.2 is not installed.
httpx: 0.27.0
huggingface-hub: 0.23.3
importlib-resources: 6.4.0
jinja2: 3.1.4
markupsafe: 2.1.5
matplotlib: 3.9.0
numpy: 1.26.4
orjson: 3.10.4
packaging: 24.1
pandas: 2.2.2
pillow: 10.3.0
pydantic: 2.7.3
pydub: 0.25.1
python-multipart: 0.0.9
pyyaml: 6.0.1
ruff: 0.4.8
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.12.3
typing-extensions: 4.12.2
urllib3: 2.2.1
uvicorn: 0.30.1
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2024.6.0
httpx: 0.27.0
huggingface-hub: 0.23.3
packaging: 24.1
typing-extensions: 4.12.2
websockets: 11.0.3
```

### Severity

Blocking usage of gradio

PR #8719, opened 2024-07-08
Fix multimodal textbox custom components

freddyaboulton:
## Description

Closes: #8718

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/test/test_gradio_component_cli.py
+++ b/test/test_gradio_component_cli.py
@@ -27,7 +27,7 @@ def test_template_override_component(template, tmp_path):
     """When you add a new component this test will likely fail locally
     because the js files have not been moved to the _frontend_code directory.
 
-    Just build the python package to move the latest state of the js directory to _frontend_code.
+    Just build the python package (python -m build -w) to move the latest state of the js directory to _frontend_code.
     """
     _create(
         "MyComponent",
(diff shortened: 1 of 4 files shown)


abidlabs:
Thanks @freddyaboulton! Could we add a test to ensure that all of our core components can be used as templates for custom components? Would be helpful as we add new components (e.g. TimeRange)

freddyaboulton:
Modified the tests @abidlabs ! Thanks for the suggestion

test/test_gradio_component_cli.py
```diff
@@ -4,39 +4,31 @@
...
-        "FileExplorer",
-        "Model3D",
-    ],
+    core,
 )
 def test_template_override_component(template, tmp_path):
+    """When you add a new component this test will likely fail locally
+    because the js files have not been moved to the _frontend_code directory.
+
+    Just build the python package to move the latest state of the js directory to _frontend_code.
```
abidlabs, line 30: What does "build the python package" mean? Do you mean build the frontend?

abidlabs (approved):
lgtm thanks for updating the test @freddyaboulton! just one question above

test/test_gradio_component_cli.py
freddyaboulton: Need to build the wheel so that a script to copy the js files gets run. Let me update comment!

Issue #8700, opened 2024-07-04
API Docs do not support auth


omerXfaruq:
### Describe the bug

Hello, API docs are not aware of demo being authenticated.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```
import gradio as gr

def hello(msg):
    return f"{msg} hello"

with gr.Blocks() as app:
    txt1 = gr.Textbox()
    txt2 = gr.Textbox()
    txt1.change(hello, txt1, txt2)

app.launch(auth=("gradio", "gradio"))
```

### System Info

```shell
macos
gradio == 4.32.2
```

### Severity

Blocking usage of gradio

abidlabs:
Thanks @omerXfaruq just to confirm, you mean the Clients code needs to be updated to include `auth=`?

omerXfaruq:
Yes, but not just for clients, also for curl. I couldn't generate curl documentation with this code snippet. But curl needs an update as well.

Issue #2376, opened 2022-09-30
[docs] document `/config`


pngwn:
- [x] I have searched to see if a similar issue already exists.

From #2364.

We should document at least parts of the `/config` path. As explained in the linked issue, the config is useful for API introspection but it does have other use cases too. The config doesn't actually change that much (at least the general shape) and most changes have not been breaking but rather additive. I'm not sure we should publicly document the whole thing but we could document parts of it. Alternatively we could add some new endpoints to surface the parts that we want users to use and keep everything else internal, to reduce the risk of breaking changes.

Additionally, this is an opportunity to clean up at least the API part, into a state that we'd happy to surface publicly. It probably doesn't need much work but maybe there are ways we can make it friendlier to users if we are going to surface it. I haven't actually looked at it in a long time, so it could be fine. Will look at it soon to see if I have any thoughts.

PR #8720, opened 2024-07-08
Documents auth in the guides, in the view API page, and also types the Blocks.config object 

abidlabs:
This PR does a few different things:

* In the Client Guides, it includes sections explaining how to connect to a Gradio app with `auth`
* In the view API page, if you look at the sample code with the clients, it includes the `auth` parameter if the gradio app includes auth

Closes: #8700 

* It also adds typing for the `Blocks.config` attribute in the backend. I believe should be good to closes: https://github.com/gradio-app/gradio/issues/2376, because (1) the information that the original user was asking for is already available in the `/info` route, in the `.view_api()` method, and in the view API page / docs. (2) in terms of documenting the /config route, I don't think it makes sense to add to our docs / guide, but now we at least have a well-defined typeddict that we point users to.

### DIFF
--- a/.changeset/witty-pugs-beg.md
+++ b/.changeset/witty-pugs-beg.md
@@ -1,0 +1,7 @@
+---
+"@gradio/app": patch
+"@gradio/client": patch
+"gradio": patch
+---
+
+feat:Documents auth in the guides, in the view API page, and also types the Blocks.config object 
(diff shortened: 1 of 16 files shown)


aliabd (approved):
Looks good @abidlabs! 

Though why so many `# type: ignore` ?

abidlabs:
Thanks @aliabd! The `type: ignore`'s are needed because we don't populate all of the keys of the BlocksConfig at once and so there are parts where the typing doesn't exactly match up. We can potentially refactor the code so that we have a valid BlocksConfig from the very beginning but I don't think its a huge priority

Issue #7632, opened 2024-03-07
The rendering area of `gr.Model3D` is collapsed in the second `gr.Tab`


hysts:
### Describe the bug

When `gr.Model3D` is used in a hidden `gr.Tab`, the rendering area is collapsed when opening the tab.

might be related to: #6850 #7485

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

repro Space: https://huggingface.co/spaces/hysts-debug/model3d-rendering-area-collapsed

```python
import gradio as gr

with gr.Blocks() as demo:
    with gr.Tabs():
        with gr.Tab(label="a"):
            out1 = gr.Model3D(value="data.ply")
        with gr.Tab(label="b"):
            out2 = gr.Model3D(value="data.ply")

if __name__ == "__main__":
    demo.queue().launch()
```

### Screenshot

It's rendered correctly in the first tab,
![](https://github.com/gradio-app/gradio/assets/25161192/c17ba70e-7829-4ab5-9e69-dd283ee27c68)
but the rendering area is collapsed in the second tab.
![](https://github.com/gradio-app/gradio/assets/25161192/60544f49-00fa-4ef1-b631-e53ecad0f3e9)

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Linux
gradio version: 4.20.1
gradio_client version: 0.11.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.2.0
fastapi: 0.110.0
ffmpy: 0.3.2
gradio-client==0.11.0 is not installed.
httpx: 0.27.0
huggingface-hub: 0.21.3
importlib-resources: 6.1.2
jinja2: 3.1.3
markupsafe: 2.1.5
matplotlib: 3.8.3
numpy: 1.26.4
orjson: 3.9.15
packaging: 23.2
pandas: 2.2.1
pillow: 10.2.0
pydantic: 2.6.3
pydub: 0.25.1
python-multipart: 0.0.9
pyyaml: 6.0.1
ruff: 0.3.1
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.9.0
typing-extensions: 4.10.0
uvicorn: 0.27.1
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2024.2.0
httpx: 0.27.0
[...]

PR #8740, opened 2024-07-10
Fixes Plotly is rendered smaller in a hidden `gr.Tab`

dawoodkhan82:
## Description

Resizes plotly div. There is a bit of a visual resizing when the tab is first switched. I think this is a bug with tabs itself. 

Closes: #6850

## 🎯 PRs Should Target Issues

## Tests


abidlabs:
If its a problem with Tabs, which I agree with given the other issues https://github.com/gradio-app/gradio/issues/6850
.ply and https://github.com/gradio-app/gradio/issues/7632, then I think we should fix the issue there, instead of in Plot.svelte, no?

dawoodkhan82:
@abidlabs yeah looking into it. But I think this PR still good to add. It basically resizes the plotdiv whenver its parent div changes sizes. This _shouldn't_ break anything.

abidlabs:
Thanks folks! Let's merge this in for the release tomorrow

PR #8743, opened 2024-07-10
Perform CORS validation when the request has a cookie

abidlabs:
Removes the check for when the request has a cookie


freddyaboulton (approved):
Thanks @abidlabs !

abidlabs:
Thanks @freddyaboulton!

PR #8773, opened 2024-07-14
Hide embedded components while loading 

aliabd:
Hides the embedded components from the screen while they're loading. 

Before:


abidlabs (approved):
Nice tysm!

Issue #8736, opened 2024-07-10
Slider Component Demos Not Working in Documentation


lappemic:
### Describe the bug

I've noticed an issue in the Gradio documentation, specifically in the [Slider component demos section](https://www.gradio.app/docs/gradio/slider#demos). The demo tabs for "sentence_builder", "slider_release", "interface_random_slider", and "blocks_random_slider" aren't displaying any content. Additionally, the "Open in ..." button, which is expected to open the demos in a separate environment, is also not functioning.

Could someone confirm if this is a general issue? I'm willing to assist in resolving this if needed.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

Go to [Slider component demos section](https://www.gradio.app/docs/gradio/slider#demos) and try to open in separate window or choose any given tab.

### System Info

```shell
On `main` as well as on `4.37.2` docs
```

### Severity

I can work around it

lappemic:
My bad! Missed this, sorry! Can confirm it happens in Firefox and Safari (interactive demo does also not work on top of page). Works in Chrome.

abidlabs:
Fixed now, confirmed in Firefox!

<img width="782" alt="image" src="https://github.com/user-attachments/assets/afa80e7b-da3f-4400-b163-5c479d675259">

PR #8784, opened 2024-07-15
Fix OS detection for cross-browser compatibility

lappemic:
This PR addresses part of the issue described in #8736, where Slider Component Demos and some UI elements were not working correctly in Firefox and Safari.

Changes made:
- Replaced the use of `navigator.userAgentData` with a more compatible `get_os()` function
- Updated the logic for setting the `meta_key` based on the detected OS

These changes improve compatibility across different browsers, particularly Firefox and Safari. The new implementation:
- Fixes the functionality of tabs and the "Open in..." button in Firefox and Safari
- Resolves the demo loading issue in Safari

Note: The demo in Firefox still requires further investigation.

To test:
1. Check the functionality of tabs and the "Open in..." button in Firefox and Safari
2. Verify that the demo loads correctly in Safari
3. Confirm that the correct keyboard shortcut (⌘+K for Mac/iOS, CTRL+K for others) is displayed

If anyhone knows how to address the Firefox demo issue i would love to fix this as well.


aliabd (approved):
Thanks @lappemic! Changes look good to me. 

I'll continue to debug what's going on with demos on firefox.

Issue #8793, opened 2024-07-15
gr.DateTime documentation issue <- duplicated TextBox, not DateTime. 


robwsinnott:
### Describe the bug

https://www.gradio.app/docs/gradio/datetime <- still appears to be the textBox documentation for the most part 

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

https://www.gradio.app/docs/gradio/datetime

### Screenshot

https://www.gradio.app/docs/gradio/datetime

### System Info

```shell
Not Relevant - on Website gradio.app -> https://www.gradio.app/docs/gradio/datetime
```

### Severity

I can work around it

aliabid94:
fixing, thanks for pointing it out!

robwsinnott:
I’d love to help out with your upcoming pdf annotation component!  I’m
building one now, but assume you all have one in motion too?

abidlabs:
> I’d love to help out with your upcoming pdf annotation component!  I’m building one now, but assume you all have one in motion too?

We don't have a specific component to annotate pdfs on the roadmap. Are you building a custom component? Happy to try it out and share it on our socials (if you are interested)

Issue #2797, opened 2022-12-12
Allow overriding  the 5000 data rows limit for for altair visualizations


freddyaboulton:
- [ ] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
Altair will throw an error if the dataframe has more than 5000 rows: https://altair-viz.github.io/user_guide/faq.html#maxrowserror-how-can-i-plot-large-datasets

I think this limitation is reasonable as every row of the dataset gets mapped to a mark on the chart and visualizing more than 5000 points on a scatterplot, for example, will probably not look good anyways.

That being said, we should decide what should happen when someone inevitably passes more than 5000 rows to our visualization components. 

Options include:

- suppress altair error and proceed anyways
- raise informative error from gradio

**Describe the solution you'd like**  

**Additional context**

abidlabs:
Similar to Altair itself, I think we should raise an error explaining that a very large webpage will be created but also give users a simple way to override if they know what they're doing

venkatavamsidama:
Handling the Altair limitation on the number of rows in a dataframe is indeed important. Given the options you've mentioned, I'd suggest incorporating a combination of both to provide flexibility to users.
1)Suppress Altair Error and Proceed
2)Suppress Altair Error and Proceed

freddyaboulton:
I think you may have mistyped @venkatavamsidama.You said the same thing twice 😅

PR #8807, opened 2024-07-16
Refactor plots to drop `altair` and use `vega.js` directly

aliabid94:
Lots of updates to plotting. The major thing is that Plot Components previously could not recieve any component updates because the entire vega config was generated in the backend via altair using the value and component args in the in the postprocess step. Took out altair and created the vega config in the svelte component directly. This enables many useful features, such as updating ranges of plots based on selections, etc.

Recording 2024-07-16 at 14 24 24

Updated monitoring dashboard:
Recording 2024-07-17 at 12 32 16

Should make plotting much better and enable a lot of cool dashboard usecases!

Full changes:
- Simplified and unified plotting API across native plots (LinePlot, BarPlot, and ScatterPlot). 
- Support all forms of updates being passed to native plots.
- Added .select() listener to all plots
- Added the parameters to all plots:
  - x_bin: Grouping used to cluster x values. If x column is numeric, should be number to bin the x values. If x column is datetime, should be string such as "1h", "15m", "10s", using "s", "m", "h", "d" suffixes.
  - y_aggregate: Aggregation function used to aggregate y values, used if x_bin is provided or x is a string/category. Must be one of "sum", "mean", "median", "min", "max".
  - color_map: Mapping of series to color names or codes. For example, {"success": "red", "fail": "#FF8888"}.
- Removed these visual parameters:
    - "width" (automatically expands to fill width)
    - "stroke_dash" (not very visually configurable) 
    - "overlay_point" (conflicts with the point that exists on hover for line plots)
    - "tooltip" (autmatically generated now)
    - "x_label_angle", "y_label_angle" (unnecessary visual configuration)
    - "interactive" (had already been deprecated, didn't do anythong)
    - "show_actions_button" (more visual noise)
- Updated monitoring dashboard with new plot features
- Fixed gr.datetime (did not update content when used as output)
- Fixed gr.datetime documentation on website, fixes: #8793   
- Since we remove altair, this closes: https://github.com/gradio-app/gradio/issues/2797


abidlabs:
A few quick questions

1. Can we remove `altair` as a dependency of gradio?

2. it looks like you removed specific component types:`gr.LinePlot`, `gr.BarPlot`, why? I think users search for these components and make the charts easy to use. Ofc if we are removing them, we'll need to include this in 5.0

aliabid94:
> Can we remove altair as a dependency of gradio?

done.

> it looks like you removed specific component types:gr.LinePlot, gr.BarPlot, why? I think users search for these components and make the charts easy to use. Ofc if we are removing them, we'll need to include this in 5.0

No these all still exist, just documented the changes I made to these plots in the PR description (includes visual breaking changes)

aliabid94:
This is ready for review

gradio/blocks.py
```diff
@@ -925,6 +925,7 @@ def __init__(
         js: str | None = None,
         head: str | None = None,
         fill_height: bool = False,
+        fill_width: bool = False,
```
abidlabs, line 928: Why is this part of this PR? I thought this was already done in a separate PR?

gradio/blocks.py
```diff
@@ -938,6 +939,7 @@ def __init__(
             js: Custom js as a string or path to a js file. The custom js should be in the form of a single js function. This function will automatically be executed when the page loads. For more flexibility, use the head parameter to insert js inside <script> tags.
             head: Custom html to insert into the head of the demo webpage. This can be used to add custom meta tags, multiple scripts, stylesheets, etc. to the page.
             fill_height: Whether to vertically expand top-level child components to the height of the window. If True, expansion occurs when the scale value of the child components >= 1.
+            fill_width: Whether to horizontally expand to fill container fully. If False, centers and constrains app to a maximum width.
```
abidlabs, line 942:
```suggestion
            fill_width: Whether to horizontally expand to fill container fully. If False, centers and constrains app to a maximum width. Only applies if this is the outermost `Blocks` in your Gradio app.
```

gradio/components/native_plot.py
```diff
@@ -0,0 +1,255 @@
...
+    datatypes: Dict[str, Literal["quantitative", "nominal", "temporal"]]
+    mark: str
+
+
+@document()
+class NativePlot(Component):
+    """
+    Creates a line plot component to display data from a pandas DataFrame.
+
+    Demos: live_dashboard
```
abidlabs, line 30:
```suggestion
    Demos: native_plots
```

gradio/components/native_plot.py
```diff
@@ -0,0 +1,255 @@
...
+    columns: List[str]
+    data: List[List[Any]]
+    datatypes: Dict[str, Literal["quantitative", "nominal", "temporal"]]
+    mark: str
+
+
+@document()
+class NativePlot(Component):
+    """
+    Creates a line plot component to display data from a pandas DataFrame.
```
abidlabs, line 28:
```suggestion
    Creates a native Gradio plot component to display data from a pandas DataFrame. Supports interactivity and updates.
```

js/nativeplot/Example.svelte
```diff
@@ -0,0 +1 @@
+<div></div>
```
abidlabs, line 1: You don't need to add an `Example.svelte` file since this component is not going to be used as an input

requirements.txt
```diff
@@ -1,5 +1,4 @@
 aiofiles>=22.0,<24.0
-altair>=5.0,<6.0
```
abidlabs, line 2: 🥳

abidlabs (approved):
Great, clean work @aliabid94! I tested a lot of different plots and everything is working well. 

Just note that the visual tests need to be fixed and that the website is broken as well (I think the .svx files need to be generated). Otherwise all good!

gradio/blocks.py
aliabid94: https://huggingface.slack.com/archives/C02QNREU24S/p1721340933971549

pngwn (approved):
Took a pass over the frontend, few small nits for consistency but looks great! Thanks @aliabid94! Excited to play more with this.

gradio/components/native_plot.py
```diff
@@ -0,0 +1,254 @@
...
+        caption: str | None = None,
+        sort: Literal["x", "y", "-x", "-y"] | list[str] | None = None,
+        height: int | None = None,
+        label: str | None = None,
+        show_label: bool | None = None,
+        container: bool = True,
+        scale: int | None = None,
+        min_width: int = 160,
+        every: Timer | float | None = None,
+        inputs: Component | list[Component] | set[Component] | None = None,
```
abidlabs, line 60: See https://github.com/gradio-app/gradio/pull/8854
```suggestion
        inputs: Component | Sequence[Component] | AbstractSet[Component] | None = None,
```

gradio/components/native_plot.py
```diff
@@ -0,0 +1,254 @@
+from __future__ import annotations
+
+import json
+import warnings
+from typing import TYPE_CHECKING, Any, Callable, Dict, List, Literal
```
abidlabs, line 5:
```suggestion
from typing import TYPE_CHECKING, AbstractSet, Any, Callable, Dict, List, Literal, Sequence
```

aliabid94:
Thanks for the reviews guys!

jamie0725:
hi @aliabid94, I see that the `x_label_angle` parameter is removed, is there a particular reason for that?

Right now my generated plot is in a mess because texts are overlapping each other on the axis.

image

Previously I can get around this with the angle parameter.

aliabid94:
whoops thanks for flagging @jamie0725, will fix

PR #8809, opened 2024-07-16
Use constant-time comparison when checking user provided analytics key

abidlabs:


aliabid94 (approved):
Tested, lgtm

gradio/routes.py
```diff
@@ -1185,7 +1186,7 @@ async def analytics_login(request: fastapi.Request):
 
         @app.get("/monitoring/{key}")
         async def analytics_dashboard(key: str):
-            if key == app.analytics_key:
```
freddyaboulton, line 1188: can we use `route_utils.compare_passwords_securely` for consistency?

freddyaboulton (approved):
nice ! just left a small suggestion

gradio/routes.py
abidlabs: good call thanks!

PR #8810, opened 2024-07-16
Prevent invalid values from being submitted to dropdown, etc.

abidlabs:
This change prevents values that are not valid choices in `gr.Dropdown`, `gr.Radio`, and `gr.CheckboxGroup` from being submitted. This is a good idea for several reasons:
* It prevents users from using the clients to submit values to functions that developers are not expecting
* It makes behavior of these components consistent with e.g. `gr.Number` that already prevent invalid values from being submitted
* This prevents certain security exploits from happening (or at least being exacerbated) 

targeting `5.0-dev` since technically a breaking change


aliabid94 (approved):
Nice LGTM! One thing that does work, but would be good to have a test for anyway, is if a gr.Dropdown receives an update that changes its choices, the validation check applies to the new set of choices, not the original ones.

abidlabs:
Thanks for the suggestion @aliabid94, I added that test. Will merge once CI passes!

PR #8817, opened 2024-07-17
Allow HTML in alert modals

abidlabs:


pngwn (changes requested):
Is this safe? Is there a scenario in which an error message is being embedded in the returned HTML and a user could manipulate this process and inject something harmful, either directly or via a mitm attack?

At the very least we need to escape the HTML.

abidlabs:
Ok @pngwn I am now sanitizing the HTML, which at least makes this consistent with our other Markdown or HTML-based components

pngwn (approved):
Looks great. Thanks @abidlabs!

Issue #8819, opened 2024-07-17
Wrong Required Param Check in JavaScript Client


JacobLinCool:
### Describe the bug

The `map_data_to_params` checks all parameters on all endpoints and causes a failure if a required parameter from another endpoint is not given.

I am working on a PR to fix it.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

Taking my space as an example: https://huggingface.co/spaces/JacobLinCool/vocal-separation

The suggested code for the `/youtube` endpoint:
```js
import { Client } from "@gradio/client";

const client = await Client.connect("JacobLinCool/vocal-separation");
const result = await client.predict("/youtube", { 		
		url: "Hello!!", 
});

console.log(result.data);
```

Parameter check failed:

```
Submit function encountered an error: Error: No value provided for required parameter: param_0
    at <anonymous> (/Users/jacoblincool/Documents/GitHub/gradio-bot/node_modules/.pnpm/@gradio+client@1.3.0/node_modules/@gradio/client/dist/index.js:476:13)
    at Array.forEach (<anonymous>)
    at map_data_to_params (/Users/jacoblincool/Documents/GitHub/gradio-bot/node_modules/.pnpm/@gradio+client@1.3.0/node_modules/@gradio/client/dist/index.js:470:14)
    at Client.submit (/Users/jacoblincool/Documents/GitHub/gradio-bot/node_modules/.pnpm/@gradio+client@1.3.0/node_modules/@gradio/client/dist/index.js:1583:25)
    at <anonymous> (/Users/jacoblincool/Documents/GitHub/gradio-bot/node_modules/.pnpm/@gradio+client@1.3.0/node_modules/@gradio/client/dist/index.js:940:22)
    at new Promise (<anonymous>)
    at Client.predict (/Users/jacoblincool/Documents/GitHub/gradio-bot/node_modules/.pnpm/@gradio+client@1.3.0/node_modules/@gradio/client/dist/index.js:939:10)
    at <anonymous> (/Users/jacoblincool/Documents/GitHub/gradio-bot/packages/gradio-bot/test.ts:4:29)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
node:internal/process/promises:288
            triggerUncaughtException(err, true /* fromPromise */);
            ^

Error: No value provided for required parameter: param_0
[...]

PR #8820, opened 2024-07-17
fix: wrong named param check for js client

JacobLinCool:
## Description

The `map_data_to_params` function checks all parameters on all endpoints and causes a failure if a required parameter from another endpoint is not provided.

`map_data_to_params` should check parameters based on each endpoint, not the entire API.

Closes: #8819 

## Potential Improvement

Currently, my implementation passes the endpoint information (parameters, returns, etc.) into `map_data_to_params`, but only the parameters are used. It might be better to pass only the parameters.

### DIFF
--- a/client/js/src/helpers/api_info.ts
+++ b/client/js/src/helpers/api_info.ts
@@ -1,4 +1,4 @@
-import type { Status } from "../types";
+import type { EndpointInfo, Status } from "../types";
 import {
 	HOST_URL,
 	INVALID_URL_MSG,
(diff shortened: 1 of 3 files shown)


JacobLinCool:
Hello @hannahblair,

I think the problem is that `get_endpoint_info` returns `endpoint_info` as `undefined` here, which breaks the type guarantee at runtime:

https://github.com/JacobLinCool/gradio/blob/a64afedc60f707793933251d2fa5faf4b3a50198/client/js/src/utils/submit.ts#L57-L64

I would add a check in `map_data_to_params` first, which I think should solve the problem. Maybe it requires submitting another issue to track the problem of `get_endpoint_info`.

pngwn (approved):
Looks good to me. Thanks @JacobLinCool!

PR #8821, opened 2024-07-17
Fix dependency loop with statustracker and markdown 

aliabd:
There's a dependency loop introduced by https://github.com/gradio-app/gradio/pull/8817 where statustracker imports markdown which imports statustracker. 

This PR implements https://github.com/gradio-app/gradio/pull/8817 without using markdown.

### DIFF
--- a/.changeset/swift-ghosts-spend.md
+++ b/.changeset/swift-ghosts-spend.md
@@ -1,6 +1,6 @@
 ---
-"@gradio/statustracker": minor
-"gradio": minor
+"@gradio/statustracker": patch
+"gradio": patch
 ---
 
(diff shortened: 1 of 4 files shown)


abidlabs (approved):
LGTM thanks @aliabd!

Confirmed that sanitization is working correctly, e.g. with:

```py
import gradio as gr
import time

def show():
    gr.Info("Please visit <p>abc<iframe//src=jAva&Tab;script:alert(3)>def</p> for free datasets.", duration=100)
    time.sleep(1)
    return "Hello"


gr.Interface(show, None, "textbox").launch()
```

PR #8823, opened 2024-07-17
Fix DateTime docs

aliabd:
Not sure how this happened but the svx file for datetime references textbox instead. 

Closes: #8793


abidlabs:
Thanks! this closes: https://github.com/gradio-app/gradio/issues/8793

abidlabs (approved):
lgtm thanks @aliabd!

image

PR #8825, opened 2024-07-17
Fix param table rendering 

aliabd:
What's happening is paramviewer's styling references css variables defined in the gradio theme, which is not on the page. The simplest fix i could think of was to just import the stylesheet from here: https://gradio-hello-world.hf.space/theme.css

I already do that for lite demos on the website. Which is why you'll only see this error when directly navigating to a page without demos or embedded components, for example: https://www.gradio.app/docs/gradio/render

### DIFF
--- a/.changeset/huge-pears-like.md
+++ b/.changeset/huge-pears-like.md
@@ -1,5 +1,5 @@
 ---
-"website": minor
+"website": patch
 ---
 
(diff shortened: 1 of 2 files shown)


abidlabs:
Thanks @aliabd! Can confirm that this fixes the issue

abidlabs (approved):

PR #8832, opened 2024-07-18
Fix build for pre-release

pngwn:
There is a discrepancy between the version formatting for pre-released which is causing an issue with the build for lite.

- js: `<x.y.z>-<beta|alpha>-<n>` eg: `5.0.0-beta.0`
- python: `<x.y.x><b|a><n>` eg: `5.0.0b0`

This just reformats for lite. This is tested for the v5 branch, if CI here is good I'll merge.

### DIFF
--- a/.changeset/nasty-clowns-show.md
+++ b/.changeset/nasty-clowns-show.md
@@ -1,0 +1,6 @@
+---
+"@gradio/app": minor
+"gradio": minor
+---
+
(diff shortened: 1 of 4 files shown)


freddyaboulton (approved):
Renaming scheme makes sense to me and watches what's best practice (https://packaging.python.org/en/latest/discussions/versioning/) .

Issue #8831, opened 2024-07-18
`.input()` is missing in `gr.Image` and `gr.Audio`


hysts:
### Describe the bug

[internal Slack](https://huggingface.slack.com/archives/C02QZLG8GMN/p1721291769074839)

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

with gr.Blocks() as demo:
    image = gr.Image(interactive=True)
    image.input(fn=lambda: print("foo"))
demo.launch()
```

### Logs

```
Traceback (most recent call last):
  File "/tmp/app.py", line 5, in <module>
    image.input(fn=lambda: print("foo"))
AttributeError: 'Image' object has no attribute 'input'
```

### System Info

```shell
gradio==4.38.1
```

### Severity

Blocking usage of gradio

ngaudemet:
Hi gradio Team :) 
what would be the difference between gr.Image.upload() dans gr.Image.input()?
I’ve seen you fixed the issue but not sure to understand what would the .input event add compared to an upload event ? the event documentation is not very clear to me, and I can’t read your internal slack where you seem to describe the issue . 
thanks !

abidlabs:
Hi @ngaudemet! The `.input()` event is fired whenever the user changes the value of the `Image` component through the UI. This would mean, for example, if the user clears the image. However, `.upload()` is only fired when a new image is actually uploaded.

ngaudemet:
thanks @abidlabs! It's clearer for me now. Concerning events, the gr.File.select() still doesn't work anymore, though it is very useful when you want to know in the back end that a user selected a file in the front end :) cf issue #8772

Issue #5309, opened 2023-08-23
ensure all components have change events


pngwn:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
Make sure every component that accepts a value dispatches a change event when that value changes. There are currently a few components that do not do this.

**Describe the solution you'd like**  
I think we should change how we handle `change` events. Since `change` events only fire when the main 'data' for a component changes, we should be able to do this at the app level, rather than needing every component to handle it own change event.

**Additional context**  
This will also be nice for custom components, because then we can make sure all custom components have change event this way too.

abidlabs:
The ones that don't and should are:

* `AnnotatedImage`
* `Gallery` (#2123 might have been reverted)

freddyaboulton:
Yea implementing change when the value is an object is a pain. Have to resort to keeping track of the previous value manually and using JSON.stringfy. Would be awesome to remove this!

abidlabs:
And it would be great if we automatically supported this event in custom components so users don't have to worry about it. Technically, in gr.Interface(live=True) assumes that all events have a `.change()` event, but this may not be automatically the case for custom components.

pngwn:
x-posting from #11615 

> There are two main challenges in using data changes as a trigger for `change` events:
>
> - Premature and repeated events (when data hasn't actually changed) are very expensive and problematic for gradio because they cause server request (and frequently expensive computation), additionally we have blocking mechanisms for new requests when requests are still pending. To alleviate this we have component specific checks that decide when a change event should be emitted.
> - For some components `change` event are not exclusively tied to data changes. The `ImageEditor` for example does not update its `value` when things are changed because the `value` of a component needs to be serialisable. For the image editor this is expensive, so data is extracted on demand (when an event is triggered that uses the `ImageEditor` as an input). In this case the pattern is inverted, the change event is the thing that signifies 'data has changed'. Data changing does not trigger the `change` event.
>
> In order to implement the 'cleaner' fix for this general problems we need:
>
> - A better way to signify data has changed, that both components and the core app can understand.
> - To reduce unnecessary data updates (which also has performance implications). Svelte 5 will offer some advantages here but we also need to rethink how we do certain things i think.

PR #8836, opened 2024-07-18
Add `.input()` events to `gr.Audio` and `gr.Image`

abidlabs:
Adds `.input()` events to `gr.Audio` and `gr.Image`. The current implementation is based off of the implementation in `gr.Code` and other components, which is a bit hacky, but the goal is to unblock #8831 and these events will be refactored as part of https://github.com/gradio-app/gradio/issues/5309 anyways

```py
import gradio as gr

with gr.Blocks() as demo:
    with gr.Row():
        i1 = gr.Image()
        i2 = gr.Image()
        i3 = gr.Image()
    with gr.Row():
        a1 = gr.Audio()
        a2 = gr.Audio()
        a3 = gr.Audio()
    
    i1.input(lambda x:x, i1, i2)
    i2.input(lambda x:x, i2, i3)
    a1.input(lambda x:x, a1, a2)
    a2.input(lambda x:x, a2, a3)
    
demo.launch()
```

Closes: #8831


hysts (approved):
Thanks for the quick fix! It worked!

abidlabs:
Sweet thanks for confirming @hysts.

Issue #3419, opened 2023-03-08
Use immutable container types for method parameters


zellski:
### Describe the bug

Types like `List` and `Set` are [invariant](https://mypy.readthedocs.io/en/stable/generics.html#variance-of-generic-types), meaning a method like:
```
    def change(
        self,
        fn: Callable | None,
        inputs: Component | List[Component] | Set[Component] | None = None,
        ...
```
can't be called the way you'd generally expect with e.g. a `List[gr.Textbox]` for `inputs` (or, to be more precise, such calls don't pass static type checking.)

A more intuitive signature would use the abstract, immutable types (which exist largely for this purpose) that have covariant behaviour – we'd use `Sequence` instead of `List`, `Mapping` instead of `Dict`, `AbstractSet` instead of `Set`, and so on.

It matters much less for return types (though it can obviously be nice to signal immutability in cases where the caller shouldn't be mutating the internals of returned collections).

This should be close to a a drop-in replacement. Is it on anybody's radar? I could possibly put together a PR, if there's interest.

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction

This example snippet fails type-checking:
```
    inputs = [gr.Number()]
    outputs = [gr.Text()]
    inputs[0].change(fn=lambda n: f"The number is #{n}", inputs=inputs, outputs=outputs)
```

Pyre's output:
```
Incompatible parameter type [6]: In call `gr.events.Changeable.change`, for argument `inputs`, expected `Union[None, List[Component], Set[Component], Component]` but got `List[Number]`.
Incompatible parameter type [6]: In call `gr.events.Changeable.change`, for argument `outputs`, expected `Union[None, List[Component], Component]` but got `List[Textbox]`.
```

### Logs

```shell
N/A, not a runtime error
```

### System Info

```shell
>>> gradio.__version__
'3.19.1'
```

### Severity

annoying

abidlabs:
Thanks @zellski, I was trying to figure this out when we added typing to the codebase but couldn't fully wrap my head around covariant types. We would definitely want to fix this issue, so a PR for this would be much appreciated.

abidlabs:
Hi @zellski are you still interested in opening a PR for this?

zellski:
@abidlabs I am, as time allows – but admittedly that's a tad scarce these days & I can't make any promises. If anyone else is chomping at the bit to get it, let them!

abidlabs:
No problem,  thanks @zellski!

[...]

PR #8837, opened 2024-07-18
upgrade pyright==1.1.372

abidlabs:
I wanted to get our types in good order before the 5.0 release, so planning on working on https://github.com/gradio-app/gradio/issues/3419. But before that, upgrading pyright to the latest version.


pyproject.toml
```diff
@@ -107,6 +107,9 @@ exclude = [
   "gradio/node/*.py",
   "gradio/_frontend_code/*.py",
 ]
+stubPath = "gradio/stubs"
```
aliabid94, line 110: where is this?

aliabid94 (approved):
lgtm!

pyproject.toml
abidlabs: The typing for anyio seems to be missing / incorrect so I added a stub file that provides typing hint within our project

abidlabs: Oh you're asking _where_ is this folder. Sorry it was gitignored, just added it

abidlabs:

Issue #8812, opened 2024-07-17
No response when using gr.image in chatinterface in second round


tyc333:
### Describe the bug

If I use gr.image or other components as chatbot response in chatinterface, it will just work at the first time, but not work any more then. 

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr


def echo(message, history):
    return gr.Image(
            "https://raw.githubusercontent.com/gradio-app/gradio/main/test/test_files/bus.png"
        )


demo = gr.ChatInterface(
    fn=echo,
    title="Echo Bot",
)
demo.launch()
```

### Screenshot

![image](https://github.com/user-attachments/assets/f5bdec44-6d8e-48b6-a892-b62fa8547f9d)

### System Info

```shell
gradio 4.38
```

### Severity

I can work around it

abidlabs:
Thanks @tyc333 I can repro. We'll take a look at this as soon as we can!

PR #8847, opened 2024-07-19
Refactor Chatinterface to use Chatbot instead of gr.State variables

freddyaboulton:
## Description

Closes: #8812 
Closes: #8847

The issue is that the chatbot preprocess does not run because the input to the event handler is a gr.State variable instead of a gr.Chatbot. 

I refactored chat_interface to use the gr.Chatbot as the input/output component but there may be some unintended consequences.

## 🎯 PRs Should Target Issues

## Tests


freddyaboulton:
Thoughts on this approach @dawoodkhan82 @abidlabs ? Not sure how else to fix

dawoodkhan82:
@freddyaboulton Haven't tested yet, but this should be fine. Basically same change we made to the chatbot demos a while back.

abidlabs:
The reason we used state previously was because it simplified API usage — you didn’t have to provide the history when hitting the /chat endpoint. Haven’t taken a deep look here yet

abidlabs:
Given that we are still using `chatbot_state` in the API functions, this should be okay. Let me test it out

abidlabs:
Generally looks good @freddyaboulton. We need to also update the `_append_multimodal_history()` method as well though I believe, since right now the multimodal chatinterface demos aren't working, e.g. `demo/chatinterface_multimodal/run.py`

freddyaboulton:
Thanks for testing @dawoodkhan82 @abidlabs ! I think the multimodal issue has been fixed!

abidlabs (approved):
lgtm @freddyaboulton nice fix!

PR #8854, opened 2024-07-19
Use covariant container types across the codebase and add typing to our demos

abidlabs:
In many places in the code base, we'd use invariant types in our annotations e.g. `list[Component]`. This is a problem because when a developer of Gradio passes in any subclass of `Component`, it raises an error. I've switched to covariant types, e.g. `Sequence[Component]` to avoid this issue.

As a simple example, the following would raise an error for any user of our library:

```py
import gradio as gr

with gr.Blocks() as demo:
    op1 = gr.Number()
    op2 = gr.Number()
    out = gr.Number()
    op1.change(fn=lambda x,y:x+y, 
               inputs=[op1, op2], # this would cause a type hint error since we are passing in a list[Number] for an argument that expects a list[Component]
               outputs=out)
    
demo.launch()
```

In order to "test" this change and prevent regressions, I've extended pyright type checking for all of our demos. In a follow up PR, I'll add typing to our backend tests as well as a further check.

Closes: https://github.com/gradio-app/gradio/issues/3419

I also went ahead and deleted some demos that were very old and no longer worked.


aliabd (approved):
Although i'm still seeing errors locally when i run the type check they're all demo errors unrelated to gradio and related to external libraries. Like you said @abidlabs it would be better to have some sort of flag that ignores demos since there are a ton of dependencies there. But otherwise not seeing any errors on the backend and also confirmed the examples work correctly. 

Thanks @abidlabs!

abidlabs:
Thanks so much @aliabd! Will merge in so that I can start adding to our tests

Issue #8754, opened 2024-07-12
Forward custom authentication mechanism to httpx


valgai:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
We deployed a Gradio app behind a proxy which requires Kerberos authentication. When accessing the app from the browser, there is no problem, but the trouble begins when using the API with `gradio_client`. There is currently, AFAIK, no way to forward custom authentication mechanism to `httpx` such as [httpx_kerberos](https://github.com/hyprxa/httpx_kerberos) or [httpx-negotiate-sspi](https://github.com/rob-blackbourn/httpx-negotiate-sspi).

**Describe the solution you'd like**  
I wish the `gradio_client.client.Client` constructor would accept an authentication mechanism that would be forwarded to `httpx` calls.

**Additional context**
I wrote a minimal implementation here : c4622ea05. It works well for our usecases, but I don't know about the edge cases.

Thanks for your help !

abidlabs:
Hi @valgai yes we could add this. Would you be able to open a PR with your work so far and we can review it and modify it as needed?

valgai:
Of course, I just created one.

Issue #6049, opened 2023-10-21
Add timeout functionality to `gradio_client.Client`


hysts:
- [x] I have searched to see if a similar issue already exists.

When calling an API of a Space via `gradio_client`, it could take quite long to get the result, so I think it would be nice to be able to set timeout.
For example, the response will be slow in the following cases:
- The Space is being built.
- The queue is too long.
- The Space is in a broken state and stuck, though it's not down.

abidlabs:
Hi @hysts, just to clarify, are you talking about a timeout when you first create a `Client` object, or do you mean when you submit a prediction?

hysts:
Sorry, I wasn't clear. Actually, I was thinking of both. I think the first point is about creating a `Client` object and the rest two are about calling `.predict`.

Regarding the first point, it could take up to 30 minutes to build a Space, so it would be nice to be able to set timeout for creating a `Client` object too.
https://github.com/gradio-app/gradio/blob/8241f9a7bd034256aabb9efa9acb9e36216557ac/client/python/gradio_client/client.py#L109-L114



For more context, I was fixing [this Space](https://huggingface.co/spaces/merve/comparing-captioning-models), and it was creating a `Client` object every time it tried to call it:
```python
def generate_caption_instructblip(image):
    client = Client('RamAnanth1/InstructBLIP')
    result = client.predict(image, 'Describe the image', 1, #min length
                   250, #max length
                   5, #beam size 
                   1, #length penalty
                   1, #rep penalty
                   0.9, #top p
                   'Beam search', fn_index=1)
    return result
```
I think the author of this Space did this because the Spaces it's calling might be down or in sleep when this Space is started, but they could be back up again when users run this Space.

Maybe it would be useful to be able to create a `Client` object in a lazy manner.

MohamedAliRashad:
@abidlabs I keep getting different errors with `httpx` because of timing out

kyrylogy:
> @abidlabs I keep getting different errors with `httpx` because of timing out

dd you manage to solve it?

djaym7:
+1 on this

PR #8862, opened 2024-07-22
Support the use of custom authentication mechanism, timeouts, and other `httpx` parameters in Python Client

valgai:
## Description

Allow users of the `gradio_client` library to forward authentication mechanism to `httpx` calls using a new `httpx_auth` 
parameter in `gradio_client.client.Client` signature.

Closes: #8754 
Closes: https://github.com/gradio-app/gradio/issues/6049

## 🎯 PRs Should Target Issues

## Tests


client/python/gradio_client/client.py
```diff
@@ -79,6 +80,7 @@ def __init__(
         max_workers: int = 40,
         verbose: bool = True,
         auth: tuple[str, str] | None = None,
+        httpx_auth: AuthTypes | None = None,
```
freddyaboulton, line 83: Should the type be `httpx.Auth`?

freddyaboulton (approved):
Looks good @valgai - just left a minor suggestion to not import from private `_auth` module

valgai:
Yes, I agree it's ugly. The `httpx.Auth` type is one member of the union represented by `httpx._types.AuthTypes`.
So we do have:

```
class Auth: ...
```

and

```python
AuthTypes = Union[
    Tuple[Union[str, bytes], Union[str, bytes]],
    Callable[["Request"], "Request"],
    "Auth",
]
``` 

However, through `httpx.BasicAuth` and `httpx.auth_.FunctionAuth` types, instances of `Auth` can be created using a `Tuple[Union[str, bytes], Union[str, bytes]]` or a `Callable[["Request"], "Request"]` as an argument. Only `FunctionAuth` is not exported publicly.

abidlabs:
The `httpx.Client` class has [a lot of kwargs](https://www.python-httpx.org/api/#client) and instead of adding them individually to our `Client` class, what if we just add a `client_kwargs: dict[str, Any]` argument and forward them to the `httpx.Client` constructor?

abidlabs:
@valgai would you be able to update your PR with the suggestion above? This way will better from a maintenance perpsective

valgai:
Sure! As `client_kwargs` is being forwarded to `httpx.stream`, `httpx.post`, `httpx.get` in addition to `httpx.Client`, maybe we could name it `httpx_kwargs` instead?

abidlabs:
sounds good to me, let's change to `httpx_kwargs`

abidlabs (approved):
lgtm thanks @valgai. going to add a test and we should be good to merge

valgai:
Thanks!

PR #8867, opened 2024-07-22
Make `update_root_in_config` atomic

abidlabs:
Making `update_root_in_config` atomic is better from a security perspective. @aliabid94 could you check if this has any performance-related implications?

### DIFF
--- a/gradio/route_utils.py
+++ b/gradio/route_utils.py
@@ -54,6 +54,7 @@ if TYPE_CHECKING:
 
 config_lock = threading.Lock()
 
+
 class Obj:
     """
     Using a class to convert dictionaries into objects. Used by the `Request` class.
--- a/gradio/route_utils.py
+++ b/gradio/route_utils.py
@@ -9,6 +9,7 @@ import os
 import re
 import shutil
 import sys
+import threading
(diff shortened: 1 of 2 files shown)


gradio/route_utils.py
```diff
@@ -51,6 +52,9 @@
     from gradio.routes import App
 
 
+config_lock = threading.Lock()
```
aliabid94, line 55: shouldn't this be an asyncio.Lock?

aliabid94:
> could you check if this has any performance-related implications?

no change in performance

gradio/route_utils.py
aliabid94: yeah I think a threading Lock won't work because jobs are run all on the same thread (same event loop), so this could still run the code simultaneously across different async tasks.

abidlabs:
ah right because this is called from the queue. will update

gradio/route_utils.py
aliabid94: jk this is not run from the queue, thread lock is correct

aliabid94 (approved):
lgtm

PR #8869, opened 2024-07-23
Add proof of concept for watermarking

meg-huggingface:
## Description

Creates a new method in `processing_utils`, `add_watermark`, that adds a watermark image to the bottom right corner of a given video. Tested with a playable .mp4 video and a .png image.


abidlabs:
Thanks @meg-huggingface! My thinking is that we can add a `watermark_path` argument to `gr.Video` (and maybe also `gr.Image`?) which uses this internally. Is that what you had in mind as well?

meg-huggingface:
Precisely what I had in mind, yes! I can give that bit a shot, although ofc I am new to gradio internals so may do it wrong. Let me see!

abidlabs:
Thanks @meg-huggingface, happy to iterate off of your PR! Here's the relevant section of the code: https://github.com/gradio-app/gradio/blob/5622331da724eb2510c74f156a2fd140a932cc55/gradio/components/video.py#L59

meg-huggingface:
Yes, saw that!
It seems this might naturally be called via the postprocess method:

https://github.com/gradio-app/gradio/blob/5622331da724eb2510c74f156a2fd140a932cc55/gradio/components/video.py#L59

Although that would require changing some logic, and looks like it may have ramifications throughout the rest of the code.
Would you prefer me to:
(a) Make a second method that would occur after postprocess
(b) Incorporate `watermark_path` into `postprocess` & address any further code calling to it that might need to change
(c) Something else?

I think I'll go for (b) if I don't hear back. I can extend `postprocess` to optionally _also_ take a triple, in addition to the current options:
`{Tuple[str | pathlib.Path, str | pathlib.Path | None, str | pathlib.Path | None]}`, where the first argument is the video file, the second is the subtitle file, and the third is the watermarking file.

abidlabs:
going to close this as iiuc this is superseded by https://github.com/gradio-app/gradio/pull/8900

Issue #6588, opened 2023-11-27
Add "Confirmation Dialogue" before leaving the Gradio Playground


apolinario:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
If someone is deep into using the playground to prototype/play with possibilities for a gradio design, can be frustrating that a misclick closes a tab or returns to a previous page.

**Describe the solution you'd like**  
As this handles inputs that can be edited for a bit, I'd suggest to add a `Leave site?` confirmation to the Playground ge
![image](https://github.com/gradio-app/gradio/assets/788417/585a3fc4-fa6c-4a9a-addd-c2fd9ce41b35)

**Additional context**  
Obviously this happened to me after spending some time in the playground playing with building the layout of an app 😅

apolinario:
That's cool @light-and-ray, but ideally this is a feature of the playground site itself (vs. the app being developed) imo

delonleo:
Hi there. getting this error `Error TypeError: Cannot convert undefined or null to object`
Would you know how to fix this.

```python
import gradio as gr

def onIfaceLoadedJS():
    return '''() => {
        console.log('JavaScript function loaded');
        window.addEventListener('beforeunload', (event) => {
            event.returnValue = 'Are you sure you want to close?';
        });
    }'''

def hello_world():
    return 'This is a warning message.'

with gr.Blocks() as demo:
    md = gr.Markdown()
    demo.load(fn=hello_world, inputs=[], outputs=[md], js=onIfaceLoadedJS())
demo.queue().launch()
```

PR #8871, opened 2024-07-23
Add confirmation dialogue if leaving playground 

aliabd:
Adds a confirmation dialogue before leaving the Playground if there are changes to a demo. 

https://github.com/user-attachments/assets/97964249-2aba-4439-8937-6216cc9ec3d4

cc @apolinario 

Closes: #6588

### DIFF
--- a/.changeset/cool-buckets-notice.md
+++ b/.changeset/cool-buckets-notice.md
@@ -1,0 +1,5 @@
+---
+"website": patch
+---
+
+feat:Add confirmation dialogue if leaving playground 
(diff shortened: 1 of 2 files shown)


abidlabs:
LGTM @aliabd! One suggestion is that if a user clicks on the "Share your app" button and does not make any further changes to the demo, that should be considered as "saving the changes" and the confirmation dialog should not appear

abidlabs (approved):

aliabd:
@abidlabs i like that suggestion! Can you please test again?

abidlabs (approved):
Nice! Now working as expected

Issue #8547, opened 2024-06-13
Cannot upload files from an iPhone


loupdmer:
### Describe the bug

Related to #7471, the bug still exists.

Contrains to upload to airdrop to my mac

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr
files = gr.File(file_types=[".zip"], visible=True)
```

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Linux
gradio version: 4.36.1
gradio_client version: 1.0.1

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.3.0
fastapi: 0.111.0
ffmpy: 0.3.2
gradio-client==1.0.1 is not installed.
httpx: 0.27.0
huggingface-hub: 0.23.3
importlib-resources: 6.4.0
jinja2: 3.1.4
markupsafe: 2.1.5
matplotlib: 3.9.0
numpy: 1.26.4
orjson: 3.10.3
packaging: 24.1
pandas: 2.2.2
pillow: 10.3.0
pydantic: 2.7.3
pydub: 0.25.1
python-multipart: 0.0.9
pyyaml: 6.0.1
ruff: 0.4.8
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.12.3
typing-extensions: 4.12.2
urllib3: 2.2.1
uvicorn: 0.30.1
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2024.6.0
httpx: 0.27.0
huggingface-hub: 0.23.3
packaging: 24.1
typing-extensions: 4.12.2
websockets: 11.0.3
```

### Severity

I can work around it

PR #8879, opened 2024-07-23
Fix file uploading in iOS

hannahblair:
## Description

For some reason, passing `file/*` to the `input`'s `accept` attribute doesn't work in iOS, and greys out all the options. passing `*`(to indicate all files are allowed) does, so this PR passes * instead of `file/*` if the device is iOS. We could pass undefined, but `*` is just for clarity. 

To test, I ran the file component demo locally, and ran the BrowserStack chrome extension with an iOS device to make sure this works!

Closes: #8547 

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/js/upload/src/Upload.svelte
+++ b/js/upload/src/Upload.svelte
@@ -42,6 +42,9 @@
 		if (type.startsWith(".") || type.endsWith("/*")) {
 			return type;
 		}
+		if (ios && type === "file") {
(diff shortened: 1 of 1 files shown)


pngwn (approved):
Why is safari like this. 

Looks great and works from my testing! Thanks @hannahblair!

PR #8884, opened 2024-07-23
Remove manual ip address check and launch counter

abidlabs:
We no longer calculate the ip address manually and send it to telemetry (instead, we just send an empty string ""). We also no longer count the number of times that gradio has been launched locally.

Targeting 5.0 since it seems like a good time to make this change.


gradio/analytics.py
```diff
@@ -116,51 +118,35 @@ def version_check():
...
-            ).text.strip()
-        except (httpx.ConnectError, httpx.ReadTimeout):
-            ip_address = "No internet connection"
-        Context.ip_address = ip_address
-    else:
-        ip_address = Context.ip_address
-    return ip_address
+def _write_machine_hash() -> str:
+    machine_hash = secrets.token_urlsafe(32)
+    with open(JSON_PATH, "w+", encoding="utf-8") as j:
```
aliabid94, line 123: so just to be clear, this function overwrites the entire json file. because we don't store any other information in this file anymore, just the hash? in which case, why even store in json format. just store the hash

aliabid94 (commented):
This has been raised before, but this is going to overcount a lot, due to every time Gradio is launched in a separate docker container or even virtualenv. If gradio is part of any CI or regular launch process that spins up a new environment, which it certainly is in many many instances, the numbers will be severely cooked. I wonder if this provides a useful guidance at all.

pngwn:
I don't think it matters much really. The metrics are already off in different ways. The only important thing is that they are consistent so we can track changes from a baseline to show general trends. 

We don't have much of an option anyway, we have to do something that doesn't involve IP address.

gradio/analytics.py
abidlabs: Yes, storing as json just in case we want to add more fields later on.

freddyaboulton (approved):
Thanks @abidlabs !

abidlabs:
Thanks @freddyaboulton!

PR #8896, opened 2024-07-24
Checking on generalizing an `if` statement in the Video postprocess function

meg-huggingface:
I'm working on a PR to add the ability to insert watermarks into videos. This shares some logic with adding subtitles and reformatting videos, suggesting that the `postprocess` function for the Video component is a good one for me to add to for the additional functionality.

I'm submitting this PR to see if this one-line change is allowable or if it breaks something that is wholly intentional, such as asserting that the input MUST be a two-item set (tuple or list), and never longer than that. 
If we are a bit more general, allowing longer sets -- and specifically for my case, three items -- then I can pass in a watermark file as well, and we'd also have the flexibility for further additions in the future (metadata? audio? etc.)

An alternative to this PR which would more directly address my use case while also not changing the current logic would be:
`if value is None or value == [None, None] or value == (None, None) or value == [None, None, None] or value == (None, None, None):`
...But that strikes me as a bit nutty, so I thought I would take a step back and check on this more general form.

Would this first change here be okay? If yes, I can finish up the rest to submit the full PR.

### DIFF
--- a/gradio/components/video.py
+++ b/gradio/components/video.py
@@ -247,7 +247,7 @@ class Video(Component):
         Returns:
             VideoData object containing the video and subtitle files.
         """
-        if value is None or all(x is None for x in value):
+        if value is None or (isinstance(value, (list, tuple)) and all(x is None for x in value)):
             return None
         if isinstance(value, (str, Path)):
             processed_files = (self._format_video(value), None)
--- a/gradio/components/video.py
+++ b/gradio/components/video.py
@@ -247,7 +247,7 @@ class Video(Component):
         Returns:
             VideoData object containing the video and subtitle files.
         """
-        if value is None or value == [None, None] or value == (None, None):
+        if value is None or all(x is None for x in value):
             return None
         if isinstance(value, (str, Path)):
             processed_files = (self._format_video(value), None)
--- a/.changeset/cold-streets-clap.md
+++ b/.changeset/cold-streets-clap.md
@@ -1,0 +1,5 @@
+---
+"gradio": minor
+---
+
+feat:Checking on generalizing an `if` statement in the Video postprocess function


gradio/components/video.py
```diff
@@ -247,7 +247,7 @@ def postprocess(
         Returns:
             VideoData object containing the video and subtitle files.
         """
-        if value is None or value == [None, None] or value == (None, None):
+        if value is None or all(x is None for x in value):
```
abidlabs, line 250: this is fine, but nit: we should typecheck that `x` is a list/tuple before iterating over it like this

abidlabs:
@meg-huggingface thanks for the PR! This change is fine in principle, but I think the approach of having a `value` that is a tuple of three elements is not ideal for two reasons:

1. Having multiple values in a tuple is less readable (and easier for users to mix up) than having a separate parameter in the constructor to `gr.Video`

2. I imagine that the typical use case is that the developer specifies a watermark ahead of time, and this should be applied to all videos output-ted in this Video component. In that case, it makes sense to have a fixed `watermark_path` that is independent of the `value`

So my suggested implementation would be to add a parameter in the constructor of `gr.Video` and then use this parameter in the `postprocess()` method to add a watermark before outputing the final file.

meg-huggingface:
> @meg-huggingface thanks for the PR! This change is fine in principle, but I think the approach of having a `value` that is a tuple of three elements is not ideal for two reasons:
> 
> 1. Having multiple values in a tuple is less readable (and easier for users to mix up) than having a separate parameter in the constructor to `gr.Video`
> 2. I imagine that the typical use case is that the developer specifies a watermark ahead of time, and this should be applied to all videos output-ted in this Video component. In that case, it makes sense to have a fixed `watermark_path` that is independent of the `value`
> 
> So my suggested implementation would be to add a parameter in the constructor of `gr.Video` and then use this parameter in the `postprocess()` method to add a watermark before outputing the final file.

Makes sense.
On (1), I'd set it so the order of elements 2 and 3 don't matter, what matters is the extension (echoing some of the subtitle extension-checking logic). I was nervous about adding a new parameter in the constructor, but happy to if OK by you, thanks.
On (2), that use case makes sense to me. I was echoing subtitling here so as to mirror/echo the code already in-place, but happy to make something more independent. Will resubmit.

abidlabs:
Thanks @meg-huggingface, yeah no problem adding parameters to the constructor of `gr.Video`. I think this use case (watermarks) is a bit different than subtitles since watermarks are independent of the video file while subtitles are not

abidlabs:
Feel free to close this pr if not needed

abidlabs:
going to close this as iiuc this is superseded by https://github.com/gradio-app/gradio/pull/8900

PR #8900, opened 2024-07-24
Adds ability to watermark videos via a `watermark` parameter in Video component

meg-huggingface:
Providing a draft of watermarking for video components.
There were a few decisions I had to make that I'm not sure are preferred.
For example, I've repurposed `_format_video` to also include watermarking. I could see another option to have a separate function. My main goal was to not have a video be processed more than it needs to be. 
Note that there will still be double-processing of the video if the format is changed ([processing #1](https://github.com/meg-huggingface/gradio/blob/patch-3/gradio/components/video.py#L332)) and then the watermark is added ([processing #2](https://github.com/meg-huggingface/gradio/blob/patch-3/gradio/components/video.py#L332)). It seems that that double-processing was already there.
This can be avoided at least on the watermark side by adding the watermark functionality to processing #1, but this would require  even more changes to the code and logic.
I figure this is a reasonable starting point to iterate on what you all would rather like to have.


gradio/components/video.py
```diff
@@ -120,6 +121,7 @@ def __init__(
             min_length: The minimum length of video (in seconds) that the user can pass into the prediction function. If None, there is no minimum length.
             max_length: The maximum length of video (in seconds) that the user can pass into the prediction function. If None, there is no maximum length.
             loop: If True, the video will loop when it reaches the end and continue playing from the beginning.
+            watermark_file: An image file to be included as a watermark on the video.
```
abidlabs, line 124: Let's add a description of how the watermark appears in the final video. Also if we are restricting images to specific file types, would be good to let users know here

gradio/components/video.py
```diff
@@ -199,7 +202,20 @@ def preprocess(self, payload: VideoData | None) -> str | None:
                 raise gr.Error(
                     f"Video is too long, and must be at most {self.max_length} seconds"
                 )
-
+        # TODO: Check other image extensions to see if they work.
+        valid_watermark_extensions = [".png", ".jpg", ".jpeg"]
+        if self.watermark_file is not None:
+            if not isinstance(self.watermark_file, (str, Path)):
+                raise gr.Error(
```
abidlabs, line 209: By default, `gr.Error` are displayed to end users of the app, which should only be done if the error is a result of user error. Here, I'd say its probably the developer error, so better to keep it restricted to the error trace in the console:
```suggestion
                raise ValueError(
```

gradio/components/video.py
```diff
@@ -199,7 +202,20 @@ def preprocess(self, payload: VideoData | None) -> str | None:
...
+                    f"Provided watermark file not an expected file type. "
+                    f"Received: {self.watermark_file}"
+                )
+            elif Path(self.watermark_file).suffix not in valid_watermark_extensions:
+                raise gr.Error(
```
abidlabs, line 214:
```suggestion
                raise ValueError(
```

gradio/components/video.py
```diff
@@ -292,11 +309,13 @@ def _format_video(self, video: str | Path | None) -> FileData | None:
 
         is_url = client_utils.is_http_url_like(video)
 
-        # For cases where the video is a URL and does not need to be converted to another format, we can just return the URL
-        if is_url and not (conversion_needed):
+        # For cases where the video is a URL and does not need to be converted
+        # to another format or have a watermark added, we can just return the URL
+        if not self.watermark_file or (is_url and not conversion_needed):
```
abidlabs, line 314:
```suggestion
        # For cases where the video is a URL and does not need to be converted
        # to another format and have a watermark added, we can just return the URL
        if not self.watermark_file and (is_url and not conversion_needed):
```

gradio/components/video.py
```diff
@@ -306,21 +325,40 @@ def _format_video(self, video: str | Path | None) -> FileData | None:
...
+        if self.watermark_file or (
+            self.format is not None and returned_format != self.format
+        ):
+            global_option_list = ["-y"]
+            output_file_name = video[0 : video.rindex(".") + 1]
+            if self.format is not None:
+                if returned_format != self.format and wasm_utils.IS_WASM:
+                    raise wasm_utils.WasmUnsupportedError(
+                        "Returning a video in a different format is not supported in the Wasm mode."
+                    )
```
abidlabs, line 343: wasm doesn't support ffmpeg generally, so we can't add a watermark in ffmpeg, so I'd move this back to line 337 at the top of this if block. We can modify this message to say something like "Modifying a video is not supported in the Wasm mode"

abidlabs:
Thanks @meg-huggingface! I left some nits above. Just wanted to ask if the demo is working for you? I'm seeing a blank video output:

image

I see this in the error console if its helpful:

> [vost#0:0/libx264 @ 0x12f00c5c0] No filtered frames for output stream, trying to initialize anyway.

gradio/components/video.py
meg-huggingface: I see, interesting. Thanks!

meg-huggingface:
> Thanks @meg-huggingface! I left some nits above. Just wanted to ask if the demo is working for you? I'm seeing a blank video output:
> 
> image
> I see this in the error console if its helpful:
> 
> > [vost#0:0/libx264 @ 0x12f00c5c0] No filtered frames for output stream, trying to initialize anyway.

Thanks so much for your time on this!
Yes, the demo works for me locally, but messaged within Slack about being unclear how it's meant to be viewed remotely.

- When trying to run gradio apps via a fork, you get errors that the UI needs to be recompiled
- Using the automagic results in demos where it's assumed you're working in the main repo. E.g., for the auto-magic generation of the colab, it produces the following imports:
`
!wget -q -O files/a.mp4 https://github.com/gradio-app/gradio/raw/main/demo/video_watermark/files/a.mp4
!wget -q -O files/b.mp4 https://github.com/gradio-app/gradio/raw/main/demo/video_watermark/files/b.mp4
!wget -q -O files/w1.jpg https://github.com/gradio-app/gradio/raw/main/demo/video_watermark/files/w1.jpg
!wget -q -O files/w2.png https://github.com/gradio-app/gradio/raw/main/demo/video_watermark/files/w2.png
`
Which obviously won't be there...

- How to compile a UI specifically on a Space or a Colab notebook is unknown territory to me. Not saying it's impossible, just that it would require more digging/unclear to me immediately how that would work/thought perhaps something else was preferred.

So I followed your advice about where the files could be accessed remotely based on this PR (https://gradio-builds.s3.amazonaws.com/e5bdb87f7ed88db292837e352d4b7b3afcc9d7a4/gradio-4.39.0-py3-none-any.whl), and Pete's advice of just deleting app.py in the auto-generated Space and copy+pasting in the app.py I have on my machine. (That also requires changing where the files are pulled from).

It resulted in this, which appears to work:

https://huggingface.co/spaces/meg/pr-8900-all-demos

The only funny thing I see there is that the videos are rendering as file directories, rather than videos themselves; again, unclear if this is just a matter of something on the gradio-auto-magic end that could be updated, or something on my end.

gradio/components/video.py
meg-huggingface: I guess you mean "we can't add a watermark in WASM"?
Addressed.

meg-huggingface:
> Thanks @meg-huggingface! I left some nits above. Just wanted to ask if the demo is working for you? I'm seeing a blank video output:
> 
> image
> I see this in the error console if its helpful:
> 
> > [vost#0:0/libx264 @ 0x12f00c5c0] No filtered frames for output stream, trying to initialize anyway.

Thanks for this!!
It works for me locally, but was not sure how to access the demo via the PR (mentioned in thread) in channel.
- The colab notebook generation uses the main repo, not a fork:
`
!wget -q -O files/a.mp4 https://github.com/gradio-app/gradio/raw/main/demo/video_watermark/files/a.mp4
!wget -q -O files/b.mp4 https://github.com/gradio-app/gradio/raw/main/demo/video_watermark/files/b.mp4
!wget -q -O files/w1.jpg https://github.com/gradio-app/gradio/raw/main/demo/video_watermark/files/w1.jpg
!wget -q -O files/w2.png https://github.com/gradio-app/gradio/raw/main/demo/video_watermark/files/w2.png
`
- Using the fork results in the error that gradio must be recompiled, which I wasn't sure was a normal thing to do for a Space or for a colab notebook and wasn't immediately sure how to do (thought perhaps this was not the right route to take).

abidlabs:
Awesome @meg-huggingface!

> It resulted in this, which appears to work:

Nice yep working for me

> The only funny thing I see there is that the videos in the example list at the bottom are rendering as file directories, rather than videos themselves; again, unclear if this is just a matter of something on the gradio-auto-magic end, or something on my end. The watermark on the video appears fine to me either way when the example is selected. 

The reason for that is that the demo uses a `gr.Textbox` to supply the videos. If you replace that with a `gr.Video` component, the examples will be rendered as videos as well.

Is this PR ready for another pass?

meg-huggingface:
> Awesome @meg-huggingface!
> 
> > It resulted in this, which appears to work:
> 
> Nice yep working for me
> 
> > The only funny thing I see there is that the videos in the example list at the bottom are rendering as file directories, rather than videos themselves; again, unclear if this is just a matter of something on the gradio-auto-magic end, or something on my end. The watermark on the video appears fine to me either way when the example is selected.
> 
> The reason for that is that the demo uses a `gr.Textbox` to supply the videos. If you replace that with a `gr.Video` component, the examples will be rendered as videos as well.
> 
> Is this PR ready for another pass?

Cool, yeah I realized it was probably this.

Let me do a pass on the formatting; the linter is saying it "would reformat" (without specifics), so I'll do some reformatting.

abidlabs:
> Let me do a pass on the formatting; the linter is saying it "would reformat" (without specifics), so I'll do some reformatting.

Yes, you can just run `bash scripts/format_backend.sh` and it should take care of it for you

abidlabs:
Small nit: I'm going to rename `watermark_file` -> `watermark` because in the future, we might want to extend this parameter to accept a more complex object, like a `gr.Watermark()` with information about positioning, opacity, etc.

meg-huggingface:
Agreed, was thinking this as well; figure the single `watermark_file` (and default format string) was the right thing to do for a minimal first pass. Then we can fancy it up. =)

Thank you!!!

abidlabs:
LGTM. just on issue I noticed is that `scale2ref` in deprecated in ffmpeg 7.0, which explains why I was seeing the empty video locally. I've removed the scaling for now -- so a developer will need to ensure that the watermark image is the appropriate size for the video. Couldn't figure out how to use `scale` (which is the suggested replacement) to scale the logo based on the _video_'s height, but I don't think this is a big issue for now, if needed we can add the logic for scaling an image later

abidlabs (approved):

Issue #7930, opened 2024-04-04
.las file cann't be uploaded in Gradio app in iPhone OS


shafiqul-chowa:
### Describe the bug

I created a gradio app. After running the app I accessed the app in iPhone browser (Safari/Chrome/Opera). In my app i need to upload a las file or csv file from phone storage. I can browse and upload csv file but the las file is greyed and can not be selected. Is this a gradio issue?

I saw a similar type of issue : https://github.com/gradio-app/gradio/issues/4021.
If this is a gradio issue then can you please fix it? Your cooperation is highly appreciated.

LAS file : https://laspy.readthedocs.io/en/latest/intro.html

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr
file = gr.components.File(file_types=['.csv', '.las'], label="Input File (.csv or .las)", height=200)

```

### Screenshot

<img width="614" alt="Screenshot 2024-04-04 at 8 26 28 PM" src="https://github.com/gradio-app/gradio/assets/126661128/258ae4ff-e735-4e20-908d-a07b31b54bed">

### System Info

```shell
gradio version : 4.25.0
iOS version : 17.3.1
iPhone model : iPhone 12 Pro
```

### Severity

Blocking usage of gradio

PR #8905, opened 2024-07-25
Allow use of file extensions in gr.File in iOS

hannahblair:
## Description

This one is a yet another annoying iOS issue. We accept file extensions (alongside `image/*` etc, this is fine) in the `file_types` attribute in `gr.File`, but in iOS these file extensions won't actually work, and will grey out all the options. We _could_ process each and every possible file extension that might be used and get the corresponding mime type but I don't like this approach.

I've opted to validate the file _after_ the file has been selected, showing an error with the expected file type. I'd hope that the expected file types would be displayed either in the info text by the user or some point soon in the actual upload area which I think we've discussed before (I should raise an issue for this, it would improve UX). 

Here's how the overall logic looks in the File component now:

filetype Input | accept_file_types | File Picker Behavior | Post-Upload Validation
-- | -- | -- | --
['.csv', '.doc'] | ".csv, .doc" | Allows all files | Yes, filters to .csv and .doc
'.csv' | ".csv" | Allows all files | Yes, filters to .csv
'file/*' | "*" | Allows all files | No
['image/*', 'text/*'] | "image/, text/" | Shows image and text files | No
'image/*' | "image/*" | Shows image files | No

Test example:
 
Uploading a .las file when `file_types=['.las']`

|  |   |
|--------|--------|
|  |  | 

Closes: #7930 

Caniuse reference
Screenshot 2024-07-25 at 18 44 40

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/real-buckets-ring.md
+++ b/.changeset/real-buckets-ring.md
@@ -1,0 +1,6 @@
+---
+"@gradio/upload": patch
+"gradio": patch
+---
+
+fix:Allow use of file extensions in gr.File in iOS


dawoodkhan82 (approved):
The code lgtm! Thanks for adding this. I couldn't test on iOS, but trusting it works.I think having the error for now should be good enough until we refactor the file upload Ui to include more context.

PR #8907, opened 2024-07-25
Update guides esp plots

aliabid94:
Updated all guides to remove incorrect content and simplify stuff. Improved and cleaned up demos. And added all new guides and demos on plots. 

Most of the changes in demos are removing unnecessary whitespace. to see the actual changes, go through the guides, and then see if the linked demos have actual changes.


aliabd (approved):
Thanks @aliabid94! I fixed some mismatching notebooks. 

Also the generate_english_german demo which is referenced in [using blocks like functions guide](https://www.gradio.app/guides/using-blocks-like-functions) is [broken](https://huggingface.co/spaces/gradio/generate_english_german). Unrelated to your pr, but if you know the fix maybe you can add it here. otherwise i'll look into it on a separate pr

guides/03_building-with-blocks/05_more-blocks-features
```diff
@@ -0,0 +1,72 @@
+# More Blocks Features
+
+## Examples
+
+Just like with `gr.Interface`, you can also add examples for your functions when you are working with `gr.Blocks`. In this case, instantiate a `gr.Examples` similar to how you would instantiate any other component. The constructor of `gr.Examples` takes two required arguments:
+
+* `examples`: a nested list of examples, in which the outer list consists of examples and each inner list consists of an input corresponding to each input component
+* `inputs`: the component or list of components that should be populated when the examples are clicked
+
+You can also set `cache_examples=True` similar to `gr.Interface`, in which case two additional arguments must be provided:
```
freddyaboulton, line 10: Would be good to mention lazy example caching as well

freddyaboulton (approved):
Love the new plot guides ! We should probably add a ruff rule or something to catch whitespace in the demos otherwise we'll eventually end up back where we started.

PR #8906, opened 2024-07-25
Use HTTP Livestreaming for audio/video streaming out

freddyaboulton:
## Description

Reworks our approach for streaming media to use http live streaming. This is widely supported across browsers, mobile devices and gives us the following benefits:

* More control of when playback starts via "low latency mode". This lets the playback for audio start as soon as the first chunk is received
* Can playback/rewind streams

Additional changes:
* Added ability to stream video in chunks. Each chunk should be a `.ts` file with `h.264`  codec. The strictness of the encoding is set by http live streaming. The `h.264` codec is pretty standard from what I read but I added the ability to convert `.mp4` files into the expect file type for streaming to make a bit easier for developers. 
* Ability to download generated audio/video streams

## 🎯 PRs Should Target Issues

Closes #8185 
Closes #6570 
Closes: #5187 
Closes #8177

## Tests

1. PRs will only be merged if tests pass on CI. To run the tests locally, please set up [your Gradio environment locally](https://github.com/gradio-app/gradio/blob/main/CONTRIBUTING.md) and run the tests: `bash scripts/run_all_tests.sh`

2. You may need to run the linters: `bash scripts/format_backend.sh` and `bash scripts/format_frontend.sh`

PR #8919, opened 2024-07-26
Run streaming e2e tests on firefox

freddyaboulton:
## Description

Chromium has [issues](https://issues.chromium.org/issues/40057824) with http live streaming (#8906) which was causing the e2e tests in #8906 to fail. I think this is just a chromium thing since I tested #8906 on chrome, safari, firefox, and ios and it works.

Long story short:
* run the streaming e2e tests on firefox since it's open source and supports http live streaming
* run the other tests in chromium because we the microphone recording, copy/paste permissions are only allowed in chromium by playwright.

## 🎯 PRs Should Target Issues

## Tests


.config/playwright.config.js
```diff
@@ -46,6 +59,4 @@ const lite = defineConfig(base, {
 	timeout: 60000
 });
 
-lite.projects = undefined; // Explicitly unset this field due to https://github.com/microsoft/playwright/issues/28795
```
freddyaboulton, line 49: this issue has been fixed and shipped

abidlabs (approved):
code changes look good to me @freddyaboulton. The only thing is that the e2e tests are starting to take a long time to run and installing firefox will only increase that so we should figure out a way to speed this up

PR #8926, opened 2024-07-29
fix typescript issue with custom components

### LINKED ISSUE #8897: gradio cc dev not working for Model3D template
### Describe the bug

Starting from yesterday I cannot get gradio cc working for Model3D template. After creating the custom component using the template, when I enter the URL returned by gradio cc dev, it will report the following error. It worked fine last week. I've tried on the same environment I used last week and with different environments/different versions of gradio but still got the same error. I'm not very familiar with svelte, is this a problem on my side or a possible bug related to some updates happened on the server side?

### Reproduction

gradio cc create mycomponent --template Model3D
gradio cc dev
(Open the frontend URL)

### Screenshot

Screenshot from 2024-07-24 12-28-13

### Logs

_No response_

### System Info

```shell
Ubuntu 22.04 (also tried on Ubuntu 24.04)
Gradio 4.39.0 (also tried 4.38.0 and earlier version)
```

### Severity

Blocking usage of gradio

pngwn:
## Description

There is a weird issue due to the tsconfig of `@gradio/wasm` this fixes it.

Closes #8897

### DIFF
--- a/.changeset/weak-bottles-stick.md
+++ b/.changeset/weak-bottles-stick.md
@@ -1,0 +1,5 @@
+---
+"@gradio/preview": patch
+---
+
+fix:fix typescript issue with custom components
--- a/js/preview/src/plugins.ts
+++ b/js/preview/src/plugins.ts
@@ -48,7 +48,14 @@ export function plugins(config: ComponentConfig): PluginOption[] {
 			},
 			extensions: _svelte_extensions,
 			preprocess: [
-				preprocess(),
+				preprocess({
+					typescript: {
+						compilerOptions: {
+							declaration: false,
+							declarationMap: false
+						}
+					}
+				}),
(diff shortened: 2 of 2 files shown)


freddyaboulton:
I believe this closes https://github.com/gradio-app/gradio/issues/8897 ?

freddyaboulton (approved):
Not sure how to test this - I tried creating a model3d component in `js/preview/test` on main but that works fine which makes me think the `js/preview` environment is doing something different than what would happen in the wild. We should talk about the best way to test un-released changes to `@gradio/preview`. 

Approving because it fixes the problem according to users though

pngwn:
Yeah. I don't know how to test it either. I tested it by editing my node modules with this change in a live project but slightly annoying.

pngwn:
It does indeed fix that issue.

PR #8927, opened 2024-07-29
Fix `Could not resolve "virtual:component-loader"` in gradio/utils package

benzler:
## Description

Fixes 

```
Could not resolve "virtual:component-loader"

    node_modules/@gradio/utils/src/utils.ts:237:10:
      237 │       import("virtual:component-loader").then((module) ...
          ╵              ~~~~~~~~~~~~~~~~~~~~~~~~~~

  You can mark the path "virtual:component-loader" as external to
  exclude it from the bundle, which will remove this error and leave
  the unresolved path in the bundle. You can also add ".catch()" here
  to handle this failure at run-time instead of bundle-time.

```

which occurs when installing any gradio component with the newest version from npm in a typescript svelte project.

This bug prevents us from switching from an old version of @gradio/audio from 0.11.4 to 0.13.0 and of @gradio/atoms from 0.7.3 to 0.7.7.

### DIFF
--- a/.changeset/rude-women-fly.md
+++ b/.changeset/rude-women-fly.md
@@ -1,6 +1,6 @@
 ---
-"@gradio/utils": minor
-"gradio": minor
+"@gradio/utils": patch
+"gradio": patch
 ---
 
-feat:Fix `Could not resolve "virtual:component-loader"` in gradio/utils package
+fix:Fix `Could not resolve "virtual:component-loader"` in gradio/utils package
--- a/.changeset/rude-women-fly.md
+++ b/.changeset/rude-women-fly.md
@@ -1,0 +1,6 @@
+---
+"@gradio/utils": minor
+"gradio": minor
+---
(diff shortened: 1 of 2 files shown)


benzler:
@pngwn Could you please help me? I think the workflow (https://github.com/gradio-app/gradio/actions/runs/10144390110/job/28047813456) is just randomly failing to upload the screenshots. However I have no permissions to restart the workflow nor to request a review.
From my side the fix is finished and the pr is ready for a review.

freddyaboulton:
@pngwn can you please review?

pngwn:
Will do. Need to pull this down and look at it, to see what that failure is about.

I don't think it the ideal fix for the issue in question but it will handle it better than it is currently handled.

pngwn:
Ah ok, just flake.

pngwn (approved):
Thanks @benzler!

Issue #1175, opened 2022-05-06
Collapse/expand {}[]


pngwn:

Nancis1130:
Hi, is there currently a way to collapse/expand JSON results? Perhaps I haven't found any relevant updates, I really need this function.

PR #8929, opened 2024-07-29
Add line numbering and collapse/expand logic to gr.JSON

hannahblair:
## Description

Adds line numbers and collapse and expand toggles to `gr.JSON`.

- Toggle and expand via toggle button
- Expand via preview ( `Object(3) text` )
- Line numbers cant be highlighted

~~💭 By default all key vals are expanded. WDYT of adding a `collapsed` param to gr.JSON to allow devs to collapse certain paths?~~

Now added an `open` param to expand all nodes and collapsed all nodes with a depth of >=3 by default

Demo here: https://huggingface.co/spaces/hmb/json_test

Closes: #1175 

## 🎯 PRs Should Target Issues

## Tests


dawoodkhan82 (approved):
Nice! lgtm. Also I like the idea of a `collapsed` param

dawoodkhan82:
Also, I realized we're using custom css for the json font colors. Should we import the codemirror themes, like in `gr.Code()`

abidlabs:
I think a global `open` parameter (similar to `gr.Accordion`) that sets the initial state of `gr.JSON` could be useful. Not sure if we need it to be path-specific.

abidlabs (approved):
Wow looks really nice!

Issue #3511, opened 2023-03-19
Support `info` with `gr.Image`


hysts:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  

I think the `info` feature is awesome, but currently `gr.Image` doesn't seem to support it. It would be nice if we could add `info` to the image component as well.

**Additional context**  
When I made [this Space](https://huggingface.co/spaces/ELITE-library/ELITE), I was initially thinking of adding instruction with `info`, but I realized that it wasn't supported, so I used `gr.Markdown` instead. Using `gr.Markdown` is OK, but `info` feature would be useful for image component as well, especially when using sketchpad with input image, because it would probably require instruction on what to draw with the sketchpad function.
![](https://user-images.githubusercontent.com/25161192/226161491-d3e1a02a-2bce-4f67-b6ce-40fcabf60e36.png)

Also, other than sketchpad, sometimes there might be some assumption on input images, like it has to be a human face, etc., so `info` field would be useful for such cases too.

PR #8930, opened 2024-07-29
Add `placeholder` param to Image and ImageEditor to replace upload image text

hannahblair:
## Description

Adds the option to add a `placeholder` parameter to the image components, to allow replacing the upload text on the upload area. 

Screenshot 2024-08-20 at 12 35 54

Closes: #3511

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/js/atoms/src/UploadText.svelte
+++ b/js/atoms/src/UploadText.svelte
@@ -58,7 +58,7 @@
 
 <style>
 	h2 {
-		font-size: var(--text-xl);
+		font-size: var(--text-xl) !important;
 	}
 
 	p,


freddyaboulton:
@hannahblair the functionality looks good but two comments:

* I think the text is a bit hard to read against the gray background
Screenshot 2024-08-05 at 12 53 35 PM

* Should the title be within the <Block> so that it gets covered by the status tracker like for the other form components (Number, Info)

aliabid94:
I like the second design but now it covers up too much of the image I feel.

abidlabs:
Nice @hannahblair! I think this servers the original purpose quite nicely. Let me cc @hysts if he has any thoughts on this PR.

Just a couple of nits:

1. It would be really great if new lines were supported in the placeholder, e.g.

image

2. For `gr.ImageEditor`, I think the "or select the draw tool to start" should be replaced as well?

image

hysts:
Looks great! This is different from what I was initially thinking of when I opened the issue, but this approach works perfectly and is much cleaner. Thanks @hannahblair !

abidlabs:
Thanks @hannahblair! The `placeholder` argument works great for `gr.Image`, but for `gr.ImageEditor`, its not rendering the markdown for me:

image

pngwn:
Yeah, sorry I haven't looked at this sooner. I've done a bad job of keeping up with PRs lately.

I'm concerned with this becoming a pattern as it introduces additional dependencies that may not always be needed, and quite frankly seem like a pretty blunt hammer for a very niche problem. I'm also concerned that this pattern of using kwargs for all configurability is unnecessarily complicating the class signatures for only moderate benefit but having a huge impact on DX for general users.

I think in cases like these we should look for less noisy, more optional patterns that have minimal API impact for most users, and do not introduce additional runtime performance issues, while still allowing for the necessary level of flexibility. I think composition is probably the best pattern we have at our disposal in this regard.

The issue @hannahblair linked about slots is relevant here but so too is the undocumented issue of `gr.html.Something()`.

abidlabs:
Yeah makes sense, I think Markdown may be overkill for a placeholder. But perhaps we can still support `\n`? I think we would just need to convert them to `<br/>` and then render the result as html?

gradio/components/image.py
```diff
@@ -102,6 +103,7 @@ def __init__(
             key: if assigned, will be used to assume identity across a re-render. Components that have the same key across a re-render will have their value preserved.
             mirror_webcam: If True webcam will be mirrored. Default is True.
             show_share_button: If True, will show a share icon in the corner of the component that allows user to share outputs to Hugging Face Spaces Discussions. If False, icon does not appear. If set to None (default behavior), then the icon appears if this Gradio app is launched on Spaces, but not otherwise.
+            placeholder: Custom text for the upload area. Overrides default upload messages when provided. Accepts plain text or markdown strings.
```
abidlabs, line 106:
```suggestion
            placeholder: Custom text for the upload area. Overrides default upload messages when provided. Accepts new lines and <h1> tags.
```

gradio/components/image_editor.py
```diff
@@ -205,6 +206,7 @@ def __init__(
             elem_classes: An optional list of strings that are assigned as the classes of this component in the HTML DOM. Can be used for targeting CSS styles.
             render: If False, component will not render be rendered in the Blocks context. Should be used if the intention is to assign event listeners now but render the component later.
             key: if assigned, will be used to assume identity across a re-render. Components that have the same key across a re-render will have their value preserved.
+            placeholder: Custom text for the upload area. Overrides default upload messages when provided. Accepts plain text or markdown strings.
```
abidlabs, line 209:
```suggestion
            placeholder: Custom text for the upload area. Overrides default upload messages when provided. Accepts new lines and <h1> tags.
```

js/imageeditor/shared/utils/parse_placeholder.ts
```diff
@@ -0,0 +1,21 @@
+export function parse_placeholder(text: string): string {
+	if (!text) return "";
```
abidlabs, line 2: We'll need to sanitize the the HTML first to prevent arbitrary HTML/JS from sneaking in here, right? We use DOMPurify elsewhere in the frontend

abidlabs:
Thanks @hannahblair! I tested this out, and I'm seeing:

1. as mentioned above, arbitrary HTML/JS can be injected because we don't sanitize the HTML prior to rendering. 
2. Support for headings isn't working for me, not sure why. Am I testing it wrong?

image

js/imageeditor/shared/utils/parse_placeholder.ts
pngwn: This also seems like unnecessary weight, i don't think we need to treat this as raw html, we are only supporting a subset.

js/imageeditor/shared/utils/parse_placeholder.ts
```diff
@@ -0,0 +1,21 @@
...
+
+		if (line) {
+			return `<p>${line}</p>`;
+		}
+
+		return "";
+	});
+
+	return processed_lines.join("");
+}
```
pngwn, line 21: Instead of this, I think we can just extract any headings and preserve the whitespace in the paragraph to avoid DOM injections.

I made a REPL cos that was easier: https://svelte.dev/repl/fb4577b0993a47579a981af6be3d6b36?version=4.2.18

abidlabs:
Nice @hannahblair. I noticed a couple of things with the heading:

(1) If the placeholder is *only* a heading, it will repeat twice, like this:

image

(2) The difference between the heading font size and the regular text in `gr.Image` is not that noticeable. For ImageEditor, the differences are more noticeable:

image

image

Nit: although even for `gr.ImageEditor`, they both look rather small. I would have expected the size of the regular placeholder text to the around the current size for the heading, and the heading to be slightly larger. (This is just personal preference so feel free to ignore)

js/atoms/src/utils/parse_placeholder.ts
```diff
@@ -0,0 +1,21 @@
+const RE_HEADING = /^#\s*(.+)$/m;
+
+export function inject(text: string): [string | false, string | false] {
+	const trimmed_text = text.trim();
+
+	const heading_match = trimmed_text.match(RE_HEADING);
+	const _heading = heading_match ? heading_match[1].trim() : false;
+
+	const heading_end_index = heading_match
+		? trimmed_text.indexOf("\n", heading_match.index)
```
pngwn, line 10: It is safer to use the actual length of the matched text (plus the two characters for the hash and space) because we cannot guarantee that there will be a new line character.

pngwn:
Left a comment about the above issue.

abidlabs:
> @abidlabs fixed the dupe heading text. wdyt of this heading text size? I've changed them from lg to xl:

Looks good to me! Let me do another round of testing

abidlabs:
I'm seeing headings are not rendered in `gr.Image` -- they *are* rendered in `gr.ImageEditor`:

image

Did you push all of the changes?

abidlabs (approved):
LGTM @hannahblair, thanks for accommodating all of the feedback!

Issue #6949, opened 2024-01-04
JSON index bug


wdlm520:
### Describe the bug

The following effect is the output of the JSON component of Gradio
How can I remove the Index such as 0,1
<img width="317" alt="截屏2024-01-04 11 46 33" src="https://github.com/gradio-app/gradio/assets/84497598/6f3847b5-66dc-4310-b0c9-21159bd6a20d">

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

with gr.Blocks() as demo:
    res = {'uri': '/QueryAnalyseService/query_analyse', 'da_res_list': [{'norm': [{'query': '故宫', 'prob': 1}, {'query': '故宫', 'prob': 1}]}]}
    gr.JSON(res)

if __name__ == '__main__':
    demo.launch()

```

### System Info

```shell
4.12.0
```

### Severity

I can work around it

abidlabs:
Hi @wdlm520, thanks for opening this issue. Lists are rendered as dictionaries by the `gr.JSON()` component with the index representing the key. I do feel like we should have option to keep the as lists, although we might not be able to get around to this soon. If you'd like to open a PR, we're happy to review!

As another option, we've made it possible for Gradio users to create their own custom components -- meaning that you can write some Python and JavaScript (Svelte), and publish it as a Gradio component. You can use it in your own Gradio apps, or share it so that anyone can use it in their Gradio apps. 

Here's an example of a Gradio custom component: https://github.com/PhyscalX/gradio-image-prompter -- it lets you upload images and process points or draw boxes on top of the image.

If you'd like to create your own, we've put together a Guide: https://www.gradio.app/guides/five-minute-guide, and we're happy to help.

PR #8932, opened 2024-07-29
Allow viewing JSON as list or dict with `show_indices` param

hannahblair:
## Description

Shows JSON as a list instead of an indexed dict. 

Demo
```
import json

import gradio as gr

JSONOBJ = """{"items":{"item":[{"id": "0001","type": null,"is_good": false,"ppu": 0.55,"batters":{"batter":[{ "id": "1001", "type": "Regular" },{ "id": "1002", "type": "Chocolate" },{ "id": "1003", "type": "Blueberry" },{ "id": "1004", "type": "Devil's Food" }]},"topping":[{ "id": "5001", "type": "None" },{ "id": "5002", "type": "Glazed" },{ "id": "5005", "type": "Sugar" },{ "id": "5007", "type": "Powdered Sugar" },{ "id": "5006", "type": "Chocolate with Sprinkles" },{ "id": "5003", "type": "Chocolate" },{ "id": "5004", "type": "Maple" }]}]}}"""

def fn(inp):
    return [json.loads(JSONOBJ), json.loads(JSONOBJ)]

with gr.Blocks() as demo:
    out = gr.JSON(label="Hide indices", show_indices=False)
    out2 = gr.JSON(label="Show indices", show_indices=True)

    btn = gr.Button(value="submit")
    btn.click(fn, inputs=out, outputs=[out, out2])



if __name__ == "__main__":
    demo.launch()
```

Closes: #6949

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/test/components/test_json.py
+++ b/test/components/test_json.py
@@ -28,7 +28,7 @@ class TestJSON:
             "_selectable": False,
             "open": False,
             "key": None,
-            "show_indices": False
+            "show_indices": False,
         }
         js_component = gr.Json(value={"a": 1, "b": 2})
         assert js_component.get_config()["value"] == {"a": 1, "b": 2}


abidlabs:
I don't think we even need a parameter for this. If the value that is passed into `gr.JSON` is a list, it should be rendered as a list, whereas if its a dictionary, it should be rendered as a dictionary. I've never seen a list being rendered a dict with numerical keys, unless it has been converted to that format. (i.e. I would treat this as a bug fix)

pngwn:
Screenshot 2024-07-30 at 13 00 24

This is the browser console. I agree that we shouldn't have a param but I do think we should render arrays like this for readability.

abidlabs:
Interesting, like the OP, I find the numerical indices distracting and would prefer if they were not shown (also makes the list look like a dictionary). Since it feels like developers might have different opinions about this, then let's keep the parameter (sorry @hannahblair!). I would probably call it `show_indices`. If its False (default), then we render lists as they are rendered now:

image

and if set to False, then something like:

image

freddyaboulton:
Makes sense given different developer preferences

gradio/components/json_component.py
```diff
@@ -60,6 +66,7 @@ def __init__(
             render: If False, component will not render be rendered in the Blocks context. Should be used if the intention is to assign event listeners now but render the component later.
             key: if assigned, will be used to assume identity across a re-render. Components that have the same key across a re-render will have their value preserved.
             open: If True, all JSON nodes will be expanded when rendered. By default, node levels deeper than 3 are collapsed.
+            show_indices: The view mode of the JSON component. If False (default), the JSON will be displayed as a dictionary. If True, the JSON will be displayed as a list.
```
abidlabs, line 69:
```suggestion
            show_indices: Whether to show numerical indices when displaying the elements of a list within the JSON object.
```

abidlabs (approved):
LGTM @hannahblair! Very nice. Small nit: the closing bracket of a list sometimes is green instead of red to match the opening bracket, e.g. in your example in line 29:

image

nanshenjiang:
I use the latest version of gradio and the sample code, but the parameter show_indices does not take effect.

- local pip3 version:
  '''
  pip3 list|grep gradio
  gradio              4.44.1
  gradio_client       1.3.0
  '''
- screenshot:
  截屏2025-01-06 00 57 17

Issue #8916, opened 2024-07-26
Grdio JSON component is not rendering dict object correctly


jamie0725:
### Describe the bug

Hi Gradio developers, I am using the latest version of Gradio (4.39.0), and noticed that the `gr.JSON` component is not rendering JSON objects correctly.

For example, given a value of `dict[str, str]`, Gradio will render it as a string starting with "root=". However, in previous versions of Gradio (e.g. 4.17.0), the component will render the JSON objects prettily.

This is also reproducible in the official Gradio docs (as attached below).

Is it expected? Or is it a bug?

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

```

### Screenshot

<img width="851" alt="image" src="https://github.com/user-attachments/assets/63e06ae5-1b59-4e3c-9beb-20bceb6c4dd7">

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Darwin
gradio version: 4.39.0
gradio_client version: 1.1.1

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
anyio: 4.4.0
fastapi: 0.110.3
ffmpy: 0.3.2
gradio-client==1.1.1 is not installed.
httpx: 0.27.0
huggingface-hub: 0.24.1
importlib-resources: 6.4.0
jinja2: 3.0.1
markupsafe: 2.1.5
matplotlib: 3.5.0
numpy: 1.26.4
orjson: 3.10.6
packaging: 21.3
pandas: 1.4.3
pillow: 10.4.0
pydantic: 2.8.2
pydub: 0.25.1
python-multipart: 0.0.9
pyyaml: 6.0
ruff: 0.5.4
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.12.3
typing-extensions: 4.12.2
urllib3: 2.2.2
uvicorn: 0.17.6
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2024.6.1
httpx: 0.27.0
huggingface-hub: 0.24.1
packaging: 21.3
typing-extensions: 4.12.2
websockets: 11.0.3
```

### Severity

Blocking usage of gradio

PR #8934, opened 2024-07-29
Fix Json component serialization bug

freddyaboulton:
## Description

Closes: #8916

You can test with `demo/json_component/run.py`

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/sad-sides-grow.md
+++ b/.changeset/sad-sides-grow.md
@@ -1,0 +1,5 @@
+---
+"gradio": patch
+---
+
+fix:Fix Json component serialization bug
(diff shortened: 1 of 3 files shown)


abidlabs:
Thanks @freddyaboulton can you explain the change quickly? Trying to understand how this change fixed the json serialization

freddyaboulton:
We changed the return type of the json component's postprocess to be a pydantic model for security reasons. Pydantic models get automatically serialized in the queue/api functions. But the `/config` route uses a custom `ORJsonResponse` that does not serialize pydantic models. So manually serializing the `value` prop so that it is already serialized when we call `get_config`.

abidlabs (approved):
Ah got it, thanks!

freddyaboulton:
Thanks for the reviews everyone!

Issue #5873, opened 2023-10-11
Use WebRTC for streaming media files


freddyaboulton:
- [ ] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
@pngwn brought this up in a conversation. Writing down so we don't forget. Definitely post 4.0. Details/implementation to be filled in later.

**Describe the solution you'd like**  

**Additional context**

freddyaboulton:
Fixed in #8941 to be released in 5.0 branch

Need to install both gradio and the client from these urls

pip install https://gradio-pypi-previews.s3.amazonaws.com/a0fdc968ef8eca168fb3a1713952420992de6d37/gradio-4.41.0-py3-none-any.whl

pip install "gradio-client @ git+https://github.com/gradio-app/gradio@a0fdc968ef8eca168fb3a1713952420992de6d37#subdirectory=client/python"

Issue #5848, opened 2023-10-09
Support setting sample rate and channels for mic capture


virajkarandikar:
- [x]  I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
I am building a app for real time streaming ASR use case. App is hosted on a webserver. Streaming mic data to app happens at 48khz, 2 channel. This seems to adding latency. Can this format be changed to 16khz, 1 channel?

**Describe the solution you'd like**  
Support setting audio format of the mic audio stream - sample rate, channel, sample width 

**Additional context**

Archit-Kohli:
Hey @virajkarandikar are you still working on this issue? I might have some idea how to do this, and would like to work on it

virajkarandikar:
I am not working on fixing this issue

mikejgray:
https://github.com/gradio-app/gradio/blob/37dd335e5f04a8e689dd7f23ae24ad1934ea08d8/gradio/processing_utils.py#L330

It appears that the `AudioSegment.from_file()` method does accept kwargs for sample_width, frame_rate, and channels (https://github.com/jiaaro/pydub/blob/996cec42e9621701edb83354232b2c0ca0121560/pydub/audio_segment.py#L179), but it isn't exposed in `gradio.processing_utils.audio_from_file()`.  Should be a pretty straightforward passthrough.

Will do some testing on a local fork and possibly PR, if no one beats me to it. 🙂

virajkarandikar:
Hi @mikejgray did you get time to submit PR for this?

virajkarandikar:
@mikejgray the code you mentioned above is for reading audio from file. My request was to allow setting sample rate for the mic capture.

qianhuiliu:
> Hey @virajkarandikar are you still working on this issue? I might have some idea how to do this, and would like to work on it

Hello, I have the same question. Can you share the ideas of how to figure it out with me? Thx!

qianhuiliu:
@Archit-Kohli
Hello, I have the same question. Can you share the ideas of how to figure it out with me? Thx!

Archit-Kohli:
> @Archit-Kohli
> Hello, I have the same question. Can you share the ideas of how to figure it out with me? Thx!

I too thought the same as @mikejgray but it seems @virajkarandikar wants this in the mic input, @virajkarandikar can you send the link to the files where mic input is being preprocessed, I looked (not hard enough though) but couldn't find them

virajkarandikar:
[...]

PR #8941, opened 2024-07-30
Streaming inputs for 5.0

freddyaboulton:
## Description
Reworks how input streams are handled by the queue. All the data for a stream is processed as the same queue event as opposed to separate events for each chunk of data. This is a smoother experience as the queue won't switch between streams until the stream is closed.

Since streams can theoretically run forever, streaming input events now have a `time_limit` parameter to control for how long a given stream will be processed by the queue. So `time_limit=10` means that a user will "own" the queue worker processing that stream for 10 seconds. All others will have to wait. To handle streams concurrently, the `concurrency_count` in that event needs to be set to be greater than 1.

Also added a `stream_frequency` parameter to `Image` and `Audio` components that controls the stream sampling rate. Its unit is Hz (so `stream_frequency=10` means 10 chunks per second)/

### Demo (stream image in)

```python
import gradio as gr

import gradio as gr
import numpy as np
import cv2


def transform_cv2(frame, transform):
    if transform == "cartoon":
        # prepare color
        img_color = cv2.pyrDown(cv2.pyrDown(frame))
        for _ in range(6):
            img_color = cv2.bilateralFilter(img_color, 9, 9, 7)
        img_color = cv2.pyrUp(cv2.pyrUp(img_color))

        # prepare edges
        img_edges = cv2.cvtColor(frame, cv2.COLOR_RGB2GRAY)
        img_edges = cv2.adaptiveThreshold(
            cv2.medianBlur(img_edges, 7),
            255,
            cv2.ADAPTIVE_THRESH_MEAN_C,
            cv2.THRESH_BINARY,
            9,
            2,
        )
        img_edges = cv2.cvtColor(img_edges, cv2.COLOR_GRAY2RGB)
        # combine color and edges
        img = cv2.bitwise_and(img_color, img_edges)
        return img
    elif transform == "edges":
        # perform edge detection
        img = cv2.cvtColor(cv2.Canny(frame, 100, 200), cv2.COLOR_GRAY2BGR)
        return img
    else:
        return np.flipud(frame)

with gr.Blocks() as demo:
    gr.Markdown("# New Image Streaming In/Out 📸")
    with gr.Row():
        with gr.Column():
            transform = gr.Dropdown(choices=["cartoon", "edges", "flip"],
                                    value="flip", label="Transformation")
            input_img = gr.Image(sources=["webcam"], stream_frequency=10, type="numpy")
        with gr.Column():
            output_img = gr.Image(format="jpeg")
        dep = input_img.stream(transform_cv2, [input_img, transform], [output_img], time_limit=25)

if __name__ == "__main__":
    demo.launch()
```
stream_image_in_out

### Demo (audio in)

```python
import gradio as gr

from dotenv import load_dotenv
load_dotenv()

from openai import OpenAI

client = OpenAI()

def transcribe(audio: str, state: str):
    next_chunk = client.audio.transcriptions.create(
        model="whisper-1",
        file=open(audio, "rb")
    ).text
    state += " " + next_chunk
    return state, state

with gr.Blocks() as demo:
    gr.Markdown("# New Audio Streaming In 🎤")
    inp = gr.Audio(sources=["microphone"], stream_frequency=1, type="filepath")
    out = gr.Text()
    transcription = gr.State(value="")
    clear = gr.Button("Clear")

    inp.stream(transcribe, [inp, transcription], [out, transcription], time_limit=10)

if __name__ == "__main__":
    demo.launch()
```

## 🎯 PRs Should Target Issues

Closes: #5873 
Closes: #5848

## Tests


js/app/src/Blocks.svelte
```diff
@@ -254,25 +255,38 @@
...
+						make_prediction(payload, dep.connection == "stream");
 					} else {
 						dep.final_event = payload;
 					}
 				}
 			}
 		}
 
-		async function make_prediction(payload: Payload): Promise<void> {
+		async function make_prediction(
```
freddyaboulton, line 272: If we're streaming, `make_prediction` will send the latest stream event data to the queue as opposed to creating a whole new event.

js/app/src/Blocks.svelte
```diff
@@ -518,6 +535,17 @@
 				messages = [new_message(data, -1, event), ...messages];
 			} else if (event == "clear_status") {
 				update_status(id, "complete", data);
+			} else if (event == "close_stream") {
```
freddyaboulton, line 538: You'll see `close_stream` in two places in the front-end - 

1. The first is here where I've added a `close_stream` event that can be emitted by the svelte components. This signals to the queue that the stream has been closed client-side and so we can just terminate the event running on the queue
2. The second is the `close_stream` function in `init.ts`. If the stream is closed server-side (time limit reached), we need to signal to the components that they should stop streaming. Each streaming component exposes a `close_stream` callback that the app will call once the stream is closed server-side

js/image/shared/ImageUploader.svelte
```diff
@@ -47,7 +50,7 @@
 	async function handle_save(img_blob: Blob | any): Promise<void> {
 		pending = true;
 		const f = await upload_input.load_files([
-			new File([img_blob], `webcam.png`)
+			new File([img_blob], `image/${streaming ? "jpeg" : "png"}`)
```
freddyaboulton, line 53: We should use `jpeg` for streaming since they're smaller

gradio/events.py
```diff
@@ -348,6 +348,7 @@ def __init__(
         trigger_after: int | None = None,
         trigger_only_on_success: bool = False,
         doc: str = "",
+        connection: Literal["sse", "stream"] = "sse",
```
freddyaboulton, line 351: `connection` is maybe a confusing name here, but basically we need to be able to tell which events are streaming events without relying on the event name

gradio/exceptions.py
```diff
@@ -59,6 +59,12 @@ class GradioVersionIncompatibleError(Exception):
     pass
 
 
+class StreamingTimeoutError(Exception):
```
freddyaboulton, line 62: I think I can remove this

gradio/processing_utils.py
```diff
@@ -489,6 +489,8 @@ async def _move_to_cache(d: dict):
 
 def add_root_url(data: dict | list, root_url: str, previous_root_url: str | None):
     def _add_root_url(file_dict: dict):
+        if file_dict.get("data") is not None:
```
freddyaboulton, line 492: Will remove - typo

gradio/queueing.py
```diff
@@ -532,15 +597,25 @@ async def process_events(
...
                     gr_request=gr_request,
                     fn=fn,
                     root_path=root_path,
                 )
+                end = time.monotonic()
+                first_iteration = end - start
                 err = None
+                for event in awake_events:
+                    event.run_time += end - start
+                    event.n_calls += 1
```
freddyaboulton, line 615: Will remove this n_calls line

gradio/events.py
abidlabs: Why not just a boolean `streaming` parameter?

gradio/components/audio.py
```diff
@@ -109,6 +109,7 @@ def __init__(
         max_length: int | None = None,
         waveform_options: WaveformOptions | dict | None = None,
         loop: bool = False,
+        stream_frequency: float = 2,
```
abidlabs, line 112: Nitty nit, but we could call this `stream_every` to make it consistent with our other `every` parameter

abidlabs: Could we consolidate this and the `streaming` arguments?

gradio/components/image.py
```diff
@@ -101,6 +102,7 @@ def __init__(
             key: if assigned, will be used to assume identity across a re-render. Components that have the same key across a re-render will have their value preserved.
             mirror_webcam: If True webcam will be mirrored. Default is True.
             show_share_button: If True, will show a share icon in the corner of the component that allows user to share outputs to Hugging Face Spaces Discussions. If False, icon does not appear. If set to None (default behavior), then the icon appears if this Gradio app is launched on Spaces, but not otherwise.
+            stream_frequency: The number of times per second to stream the webcam feed. Only valid if streaming is True.
```
abidlabs, line 105: Is it true that this argument is valid only if streaming is True? In your example demo, you have this line:

>             input_img = gr.Image(sources=["webcam"], stream_frequency=10, type="numpy")

abidlabs: I'm a little confused about the `streaming` argument actually. Is it used at all for input streaming? If I understand correctly, it is only used to set the behavior when a component is being used as output. We should make this clearer in the docs and docstrings.

abidlabs:
Very nice @freddyaboulton! I took it for a spin and the streaming works very nicely. I left a couple of minor points above, but my main feedback is around the `time_limit` experience. 

(1) I mentioned this before, but when you're using the app locally (or in a private Space), the "Time limit reached!" modal is jarring as it seems like the Gradio app is putting an artificial limit in my ability to interact with my own model / demo. I think the simplest way would be to automatically resend the stream event when it is complete to the end of the queue, which would work no matter how many people are using the app. However, this might produce an interruption between consecutive events, so there might be a better way to manage the single-user use case

(2) I like the orange bar that partially disappears to indicate the streaming time. However, I'm not sure if it makes sense to put it in the output component, since the `time_limit` is actually more connected to the input component. In some cases, you might have multiple output components (or _no_ output components): 

image

I'd suggest moving this indicator over to the input component. Perhaps you could have a red border around the record button which partially disappears to indicate the time limit.

gradio/components/image.py
freddyaboulton: Yes `streaming` parameter is only used for outputs. Will make a note of that in docstring.

abidlabs: We do have this logic:

```
            self.sources = (
                ["webcam"] if streaming else ["upload", "webcam", "clipboard"]
            )
```

but I think we can change this to: if an image component only has .stream() events, then the default source is webcam

abidlabs: Perhaps we should rename the `streaming` parameter to `stream_out` to help users know which flags they need to set for which use case (if we are worried about breaking, then we could at least deprecate the former and assign its value to the latter)

gradio/components/audio.py
abidlabs: Actually could we make this a parameter of .stream()? That way (a) different events could have different frequencies (b) this parameter and the time_limit() parameter are in the same place (c) if someone isn’t using streaming, they don’t have to think about this parameter

freddyaboulton: Done

abidlabs: > Actually could we make this a parameter of .stream()? That way (a) different events could have different frequencies (b) this parameter and the time_limit() parameter are in the same place (c) if someone isn’t using streaming, they don’t have to think about this parameter

Thoughts on this @freddyaboulton?

abidlabs:
Thanks @freddyaboulton changes look good! The single-user experience is quite good, but a couple of suggestions to improve the streaming input UX when there are multiple users using the demo at the same time:

1. When you are waiting for your turn in the queue, there is no progress indicator in the top right of the output to let you know your position in the queue, eta, etc. like there normally is for events:

image

2. Part of the progress indicator is hidden behind the label, which can be misleading. I'd suggest either putting it on the bottom of the component (closer to the record button), or making the progress indicator a circular border around the record button which gradually disappears (I think this UX would look really nice). Nit: regardless of which way we go with, can we color it red? Visually it would be more connected to the record button that way

image



This is the demo that I used to test (opened the demo in multiple tabs):

```py
import gradio as gr

import gradio as gr
import numpy as np
import cv2


def transform_cv2(frame, transform):
    if transform == "cartoon":
        # prepare color
        img_color = cv2.pyrDown(cv2.pyrDown(frame))
        for _ in range(6):
            img_color = cv2.bilateralFilter(img_color, 9, 9, 7)
        img_color = cv2.pyrUp(cv2.pyrUp(img_color))

        # prepare edges
        img_edges = cv2.cvtColor(frame, cv2.COLOR_RGB2GRAY)
        img_edges = cv2.adaptiveThreshold(
            cv2.medianBlur(img_edges, 7),
            255,
            cv2.ADAPTIVE_THRESH_MEAN_C,
            cv2.THRESH_BINARY,
            9,
            2,
        )
        img_edges = cv2.cvtColor(img_edges, cv2.COLOR_GRAY2RGB)
        # combine color and edges
        img = cv2.bitwise_and(img_color, img_edges)
        return img
    elif transform == "edges":
        # perform edge detection
        img = cv2.cvtColor(cv2.Canny(frame, 100, 200), cv2.COLOR_GRAY2BGR)
        return img
    else:
        return np.flipud(frame)

with gr.Blocks() as demo:
    gr.Markdown("# New Image Streaming In/Out 📸")
    with gr.Row():
        with gr.Column():
            transform = gr.Dropdown(choices=["cartoon", "edges", "flip"],
                                    value="flip", label="Transformation")
            input_img = gr.Image(sources=["webcam"], stream_every=0.1, type="numpy")
        with gr.Column():
            output_img = gr.Image(format="jpeg")
        dep = input_img.stream(transform_cv2, [input_img, transform], [output_img], time_limit=25)

if __name__ == "__main__":
    demo.launch()
```

abidlabs:
Btw similar to (1), when your turn is up in the queue, it should again show the progress indicator with the queue position and eta to let you know when it would be your turn again, as otherwise you're in this intermediate state where it seems like the demo has just stopped working:

image

gradio/components/audio.py
freddyaboulton: Yes it's something I thought about but it's a bit awkward since we don't have a way of passing dependency props into components but regardless I will implement.

aliabid94:
I tried running the audio streaming demo drom PR desc and got:
```
  File "/Users/aliabid/Desktop/gradio/demo/hello_world/run.py", line 1, in <module>
    import gradio as gr
  File "/Users/aliabid/Desktop/gradio/gradio/__init__.py", line 3, in <module>
    import gradio._simple_templates
  File "/Users/aliabid/Desktop/gradio/gradio/_simple_templates/__init__.py", line 1, in <module>
    from .simpledropdown import SimpleDropdown
  File "/Users/aliabid/Desktop/gradio/gradio/_simple_templates/simpledropdown.py", line 6, in <module>
    from gradio.components.base import Component, FormComponent
  File "/Users/aliabid/Desktop/gradio/gradio/components/__init__.py", line 1, in <module>
    from gradio.components.annotated_image import AnnotatedImage
  File "/Users/aliabid/Desktop/gradio/gradio/components/annotated_image.py", line 14, in <module>
    from gradio.components.base import Component
  File "/Users/aliabid/Desktop/gradio/gradio/components/base.py", line 20, in <module>
    from gradio.blocks import Block, BlockContext
  File "/Users/aliabid/Desktop/gradio/gradio/blocks.py", line 39, in <module>
    from gradio import (
  File "/Users/aliabid/Desktop/gradio/gradio/networking.py", line 14, in <module>
    from gradio.routes import App  # HACK: to avoid circular import # noqa: F401
  File "/Users/aliabid/Desktop/gradio/gradio/routes.py", line 89, in <module>
    from gradio.server_messages import (
  File "/Users/aliabid/Desktop/gradio/gradio/server_messages.py", line 51, in <module>
    class ProcessGeneratingMessage(BaseMessage):
  File "/Users/aliabid/Desktop/gradio/gradio/server_messages.py", line 52, in ProcessGeneratingMessage
    msg: Literal[ServerMessage.process_generating, ServerMessage.process_streaming] = (  # type: ignore
  File "/Library/Developer/CommandLineTools/Library/Frameworks/Python3.framework/Versions/3.9/lib/python3.9/enum.py", line 429, in __getattr__
    raise AttributeError(name) from None
AttributeError: process_streaming
```

abidlabs:
@aliabid94 make sure you're installing the client from this PR as well. I had the same issue

`pip install -e client/python`

freddyaboulton:
@abidlabs I believe I've addressed all of your comments:

* stream_every moved to event definition as opposed to component constructor
* time limit indicator moved to the bottom of the component
* queue status displayed now for streaming events

Additionally, the record button icon for images now goes off of the primary theme color to match the audio component!

abidlabs:
@freddyaboulton I'm seeing a weird issue which is that when I run this demo:

```py
import gradio as gr

import gradio as gr
import numpy as np
import cv2


def transform_cv2(frame, transform):
    if transform == "cartoon":
        # prepare color
        img_color = cv2.pyrDown(cv2.pyrDown(frame))
        for _ in range(6):
            img_color = cv2.bilateralFilter(img_color, 9, 9, 7)
        img_color = cv2.pyrUp(cv2.pyrUp(img_color))

        # prepare edges
        img_edges = cv2.cvtColor(frame, cv2.COLOR_RGB2GRAY)
        img_edges = cv2.adaptiveThreshold(
            cv2.medianBlur(img_edges, 7),
            255,
            cv2.ADAPTIVE_THRESH_MEAN_C,
            cv2.THRESH_BINARY,
            9,
            2,
        )
        img_edges = cv2.cvtColor(img_edges, cv2.COLOR_GRAY2RGB)
        # combine color and edges
        img = cv2.bitwise_and(img_color, img_edges)
        return img
    elif transform == "edges":
        # perform edge detection
        img = cv2.cvtColor(cv2.Canny(frame, 100, 200), cv2.COLOR_GRAY2BGR)
        return img
    else:
        return np.flipud(frame)

with gr.Blocks() as demo:
    gr.Markdown("# New Image Streaming In/Out 📸")
    with gr.Row():
        with gr.Column():
            transform = gr.Dropdown(choices=["cartoon", "edges", "flip"],
                                    value="flip", label="Transformation")
            input_img = gr.Image(sources=["webcam"], stream_every=0.1, type="numpy")
        with gr.Column():
            output_img = gr.Image(format="jpeg")
        dep = input_img.stream(transform_cv2, [input_img, transform], [output_img], time_limit=25)

if __name__ == "__main__":
    demo.launch()
```

on three separate tabs, the third demo never runs. It just freezes up in this state:

image

Sometimes the demo will run, but it will be very laggy. If you switch back to the first and second demos, they also start to freeze up or not work. The error-prone behavior varies a little bit, but I consistently get some sort of error when I open up the demo on three tabs (but not on just two). 

I also see a bunch of errors like this in the console:


image

Besides that, two small suggestions:
* Can we set `show_progress="minimal"` as the default for `.stream()` events? I believe we do the same thing for `.change()`
* The queue calculation seems slightly off. When I run two events with `stream_every=10`, the queue calculation should show an ETA of 10 seconds, but it actually shows 20, because its adding the total streaming time of the current job, which it shouldn't since users would want to know the time until the *start* of the stream

abidlabs:
Discussed with @freddyaboulton and the reason I'm seeing the error is because the browser is throttling the messages. Won't happen with different browsers so I think its ok

gradio/component_meta.py
```diff
@@ -57,6 +59,8 @@ def {{ event }}(self,
             concurrency_limit: if set, this is the maximum number of this event that can be running simultaneously. Can be set to None to mean no concurrency_limit (any number of this event can be running simultaneously). Set to "default" to use the default concurrency limit (defined by the `default_concurrency_limit` parameter in `Blocks.queue()`, which itself is 1 by default).
             concurrency_id: if set, this is the id of the concurrency group. Events with the same concurrency_id will be limited by the lowest set concurrency_limit.
             show_api: whether to show this event in the "view API" page of the Gradio app, or in the ".view_api()" method of the Gradio clients. Unlike setting api_name to False, setting show_api to False will still allow downstream apps as well as the Clients to use this event. If fn is None, show_api will automatically be set to False.
+            time_limit: The time limit for the function to run. Only applies if connection is "stream".
+            stream_every: The latency (in seconds) at which stream chunks are sent to the backend. Defaults to 0.5 seconds. Only applies for "stream" events.
```
abidlabs, line 63:
```suggestion
            stream_every: The latency (in seconds) at which stream chunks are sent to the backend. Defaults to 0.5 seconds. Parameter only used for the `.stream()` event.
```

gradio/component_meta.py
```diff
@@ -57,6 +59,8 @@ def {{ event }}(self,
             concurrency_limit: if set, this is the maximum number of this event that can be running simultaneously. Can be set to None to mean no concurrency_limit (any number of this event can be running simultaneously). Set to "default" to use the default concurrency limit (defined by the `default_concurrency_limit` parameter in `Blocks.queue()`, which itself is 1 by default).
             concurrency_id: if set, this is the id of the concurrency group. Events with the same concurrency_id will be limited by the lowest set concurrency_limit.
             show_api: whether to show this event in the "view API" page of the Gradio app, or in the ".view_api()" method of the Gradio clients. Unlike setting api_name to False, setting show_api to False will still allow downstream apps as well as the Clients to use this event. If fn is None, show_api will automatically be set to False.
+            time_limit: The time limit for the function to run. Only applies if connection is "stream".
```
abidlabs, line 62:
```suggestion
            time_limit: The time limit for the function to run. Parameter only used for the `.stream()` event.
```

gradio/component_meta.py
abidlabs: wording nit, applies to the other docstrings as well

abidlabs:
Thanks @freddyaboulton, gave it another review. I think everything looks good to me, except for three small things that I noticed. Two I've mentioned above, will copy here:

* Can we set show_progress="minimal" as the default for .stream() events? I believe we do the same thing for .change()


* The queue calculation seems slightly off. When I run two events with stream_every=10, the queue calculation should show an ETA of 10 seconds, but it actually shows 20, because its adding the total streaming time of the current job, which it shouldn't since users would want to know the time until the start of the stream


The last thing I noticed is that when using `gr.Audio` as an input with streaming, it seems that even though the stream event gets restarted after the streaming is done, you don't see the red bar reappear. Here's the demo I am running:

```py
import gradio as gr

t = "the big blue " * 5
i = 0

def transcribe(audio: str, state: str):
    global i
    next_chunk = t[i:i+1]
    i += 1
    state += " " + next_chunk
    return state, state

with gr.Blocks() as demo:
    gr.Markdown("# New Audio Streaming In 🎤")
    inp = gr.Audio(sources=["microphone"], type="filepath")
    out = gr.Text()
    transcription = gr.State(value="")
    clear = gr.Button("Clear")

    inp.stream(transcribe, [inp, transcription], [out, transcription], time_limit=10, stream_every=1)

if __name__ == "__main__":
    demo.launch()
```

and you can see after 10 seconds, it restarts and continues streaming, but there's no red bar:

image

freddyaboulton:
Thanks @abidlabs ! I made those changes. We don't set `show_progress='minimal'` for change events but I did it for streaming here anyways. I modified the queue eta for streaming and fixed progress tracker issue with the audio component.

abidlabs (approved):
Awesome PR @freddyaboulton, and thanks for addressing all the suggestions. Everything looks good to me. 

It'd be great to add some documentation with real-world examples, but we can do that in a subsequent PR

freddyaboulton:
Thank you so much for the reviews!

PR #8948, opened 2024-07-31
Bump websockets version max for gradio-client

evanscho:
## Description

gradio-client has had a pinned version maximum for websockets of 12.0 that was created before websockets 12.0 was released.

websockets 12.0 has since been released, and the only backwards-incompatible change was ending support for python 3.7, which gradio-client and gradio don't support anymore either. 

To keep with the spirit of the original file, I've bumped up the version requirement to 13.0, which may in the future present other breaking changes.


abidlabs (approved):
Thanks @evanscho!

PR #8957, opened 2024-07-31
Add information about `FORWARDED_ALLOW_IPS` environment variable

abidlabs:
Per ToB's recommendation, adds information about `FORWARDED_ALLOW_IPS` environment variable


aliabid94 (approved):
LGTM but all this content needs to be moved to docs imo, cc @aliabd

abidlabs:
I don't think its super necessary to add to docs since these env variables show up in search, which covers both guides and docs

PR #8959, opened 2024-07-31
Adds `strict_cors` parameter to `launch()`

abidlabs:
Added tests to ensure that CSRF attacks cannot happen. To test this change with regards to embedding, create a Gradio app, and then embed it with web components in a static file. Follow these steps:

1. create a gradio app:

```py
import gradio as gr

gr.Interface(lambda x:x, "text", "text").launch(strict_cors=False)
```

2. create an index.html file with

```
<script
	type="module"
	src="https://gradio.s3-us-west-2.amazonaws.com/4.40.0/gradio.js"
></script>

<gradio-app src="http://localhost:7860"></gradio-app>
```

If you open the file in a browser, the embedded app should work as long as you include the `strict_cors=False` parameter above.

### DIFF
--- a/.changeset/hungry-tips-sin.md
+++ b/.changeset/hungry-tips-sin.md
@@ -1,0 +1,5 @@
+---
+"gradio": minor
+---
+
+feat:Adds `strict_cors` parameter to `launch()`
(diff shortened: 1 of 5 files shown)


aliabd (approved):
Nice! tested and works great

not sure what's going on with functional tests?

abidlabs:
Yeah that's something we need to figure out on the `5.0-dev` branch but unrelated to this PR

Issue #6939, opened 2024-01-03
How to make Gallery to pop up or display image in full resolution on click


FurkanGozukara:
### Describe the bug

I am looking options here but can't find it
When image is clicked I want it to be displayed in full size or like pop up and display full resolution etc

is that possible?

https://www.gradio.app/docs/gallery

![image](https://github.com/gradio-app/gradio/assets/19240467/1ae942c5-bb52-4064-92e7-dbfb316304ad)

output_gallery = gr.Gallery(label="Generated Images")

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

output_gallery = gr.Gallery(label="Generated Images")

### System Info

```shell
(venv) G:\IP-Adapter-FaceID\venv\Scripts>pip freeze
accelerate==0.25.0
aiofiles==23.2.1
albumentations==1.3.1
altair==5.2.0
annotated-types==0.6.0
anyio==4.2.0
attrs==23.2.0
certifi==2022.12.7
charset-normalizer==2.1.1
click==8.1.7
colorama==0.4.6
coloredlogs==15.0.1
contourpy==1.2.0
cycler==0.12.1
Cython==3.0.7
diffusers==0.25.0
easydict==1.11
einops==0.7.0
exceptiongroup==1.2.0
fastapi==0.108.0
ffmpy==0.3.1
filelock==3.9.0
flatbuffers==23.5.26
fonttools==4.47.0
fsspec==2023.12.2
gradio==4.12.0
gradio_client==0.8.0
h11==0.14.0
httpcore==1.0.2
httpx==0.26.0
huggingface-hub==0.20.1
humanfriendly==10.0
idna==3.4
imageio==2.33.1
importlib-metadata==7.0.1
importlib-resources==6.1.1
insightface==0.7.3
ip-adapter @ git+https://github.com/tencent-ailab/IP-Adapter.git@9d8960cbe87977df258a0fb441fd3eba1e5efca8
Jinja2==3.1.2
joblib==1.3.2
jsonschema==4.20.0
jsonschema-specifications==2023.12.1
kiwisolver==1.4.5
lazy_loader==0.3
markdown-it-py==3.0.0
MarkupSafe==2.1.3
matplotlib==3.8.2
mdurl==0.1.2
mpmath==1.3.0
networkx==3.0
numpy==1.24.1
onnx==1.15.0
onnxruntime-gpu==1.16.3
opencv-python-headless==4.9.0.80
[...]

Issue #4908, opened 2023-07-13
Add a fullscreen icon for the Image component


taoari:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  

For the Image component, there are already **"close"** and **"edit"** buttons. It is very much desired if there could be an additional **"fullscreen"** button. The current Image component displays an image in a fixed size. The "fullscreen" button is of great help to allow a user to examine the details of an image without changing the app layout.

![image](https://github.com/gradio-app/gradio/assets/3815006/a1febec5-bf51-427d-9afa-97042e0d4e21)

**Describe the solution you'd like**  

Add an additional "fullscreen" button along with existing "edit" and "close" buttons on the top right corner of the Image component.

**Additional context**

thiswillbeyourgithub:
I think this should be added to both Image and Gallery.

It can actually be impossible for a user to differentiate some pictures in the gallery in some situation without this. My use case involves having many gallery components stacked on top of another in a column.

PR #8964, opened 2024-08-01
Add min/max-imize button to gr.Image and gr.Gallery

hannahblair:
## Description

Allows users to maximise and minimise an image of an output component to full screen.

It could also be good to include zoom in and out with panning outside of this PR to improve the UX of zoom functionality - I did initially include zoom in/out but its bad UX without the panning. 

~❓ question! should this be a param? if so, what should be the default? ❓~
~❓ another question! any other components we should apply this to? annotated image?❓~

Closes: #6939
Closes: #4908

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.github/workflows/previews-deploy.yml
+++ b/.github/workflows/previews-deploy.yml
@@ -62,6 +62,7 @@ jobs:
         uses: actions/download-artifact@v4
         with:
           run-id: ${{ github.event.workflow_run.id }}
+          github-token: ${{ secrets.GITHUB_TOKEN }}
       - name: list artifacts
         run: ls -R .
 
--- a/.github/workflows/website-deploy.yml
+++ b/.github/workflows/website-deploy.yml
@@ -10,11 +10,6 @@ concurrency:
   group: "${{ github.event.workflow_run.head_repository.full_name }}-${{ github.event.workflow_run.head_branch }}-${{ github.workflow_ref }}"
   cancel-in-progress: true
 
-# these variables are not secret and are used to identify the Vercel project
-env:
-  VERCEL_ORG_ID: ${{ secrets.VERCEL_ORG_ID }}
-  VERCEL_PROJECT_ID: ${{ secrets.VERCEL_PROJECT_ID }}
-
 jobs:
   changes:
     name: "changes"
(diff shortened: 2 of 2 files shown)


abidlabs:
This is awesome @hannahblair! UX looks good to me

> ❓ question! should this be a param? if so, what should be the default? ❓

Yes, I think so. Similar to e.g. show_share_button, you could have a `show_fullscreen_button` (which is by default `True` to help with discoverability)

> ❓ another question! any other components we should apply this to? annotated image?❓

Yes, I think this should be part of both `gr.AnnotatedImage` and `gr.ImageEditor` when `interactive=False`

js/annotatedimage/Index.svelte
```diff
@@ -38,6 +39,24 @@
...
+	let image_container: HTMLElement;
+
+	onMount(() => {
+		document.addEventListener("fullscreenchange", () => {
+			is_full_screen = !!document.fullscreenElement;
+		});
+	});
+
+	const toggle_full_screen = async (): Promise<void> => {
+		if (!document.fullscreenElement) {
```
freddyaboulton, line 54: nit: use `is_full_screen` here for consistency

freddyaboulton (commented):
Looks good @hannahblair - tested demos locally and in the deployed space and the functionality works as expected.

I think we should use the same minimize icon in the chatbot full screen mode for consistency. Right now an `X` is used.

image

Actually, the chatbot already has some custom behavior for having images/gallery take up the entire chatbot area. I think we should remove and use this behavior in this PR. Thoughts @dawoodkhan82 ?

Also, should we also add this parameter for video?

freddyaboulton:
> This would actually be a nice idea! I think its worth doing this in a separate PR, alongside the updated minimise icon

Agreed!

abidlabs (approved):
lgtm great implementation @hannahblair!

pngwn:
rude

FurkanGozukara:
finally full screen arrived nice

do we also have no full screen image slider?

abidlabs:
Full screen would be really useful for the ImageSlider agreed, I’ll make a note on the issue

Issue #8912, opened 2024-07-26
Displaying examples inside the chatbot window


yvrjsharma:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
Improving the UI for displaying examples in Gradio's Chatbot and ChatInterface components. Currently, examples are shown outside (below) the chat window, which works but could be enhanced for a better user experience, especially for multimodal demos. This could help in below mentioned benefits imo:
- Improved UX, especially for multimodal demos (Audio, Video, Image, Docs)
- More screen space for chatbot window
- Closer alignment with common practices in chat interfaces

**Describe the solution you'd like**  
- Implement a way to display example conversations directly within the Chatbot window.
- Maybe creating a new parameter for Chatbot/ChatInterface to specify these in-chat examples.
- Would be great if we can ensure compatibility with the existing `placeholder` argument of the chatbot.
- If both placeholder and new examples are provided, we can consider overlaying examples on top of the placeholder (cc: @abidlabs, from Slack conversation)

**Additional context**  
- Consider an API that could potentially support both examples and custom buttons within the chat interface (Ref: #8698; from Slack discussion)

abidlabs:
Closed via https://github.com/gradio-app/gradio/pull/8966

PR #8843, opened 2024-07-19
`5.0 dev` -> `main`

abidlabs:
Opening this PR so we can easily see the diff between `5.0-dev` and `main`, and also easily bring `5.0-dev` up-to-date with `main`. This should NOT be merged until we are ready to release 5.0

PR #8966, opened 2024-08-01
Chatbot Examples

dawoodkhan82:
## Description
Chatbot Examples

Screenshot 2024-08-28 at 2 35 41 PM

Closes: #8912, #9231

## 🎯 PRs Should Target Issues

## Tests


abidlabs:
Very nice @dawoodkhan82! I'm thinking that when you click on the examples, it populates the `Chatbot` with that value automatically? And we should also add an event to `Chatbot` that is triggered when an example is clicked. We can use that in `gr.ChatInterface` to trigger the bot response

gradio/components/chatbot.py
```diff
@@ -88,14 +88,16 @@ class Message(GradioModel):
     metadata: Metadata = Field(default_factory=Metadata)
     content: Union[str, FileMessage, ComponentMessage]
 
+class ExampleMessage(TypedDict):
```
dawoodkhan82, line 91: @abidlabs thoughts on this Data structure for examples? If we want to support multimodal examples, that is essentially 2 Messages, the text and the file.

abidlabs: I've seen multimodal chatbots where the examples are two images, etc. so I think for the general case, I would make the change below:

gradio/components/chatbot.py
```diff
@@ -88,14 +88,16 @@ class Message(GradioModel):
...
+    text: str
+    file: NotRequired[FileData]
```
abidlabs, line 93:
```suggestion
    files: NotRequired[list[FileData]]
```

gradio/components/chatbot.py
abidlabs: Also, I looked at a few example chat applications like ChatGPT and HuggingChat, and its often the case that the display text is not the same as the message that is submitted to the chatbot. For example, you can see ChatGPT's starters look like this:

image

But if you click on them, you get much more text as the initial prompt.

I suggest adding an optional argument called `label` which takes in arbitrary Markdown and if it is provided, it is rendered in the starter bubble. If its not provided, then the text and files are rendered as you currently have

gradio/components/chatbot.py
```diff
@@ -172,6 +174,7 @@ def __init__(
         likeable: bool = False,
         layout: Literal["panel", "bubble"] | None = None,
         placeholder: str | None = None,
+        examples: list[ExampleMessage] | None = None,
```
abidlabs, line 177: I know we discussed what to call this, but now I'm thinking `examples` might be confusing to users since this does not follow the same behavior as examples typically do for components. What do you think about `suggestions`?

gradio/components/chatbot.py
```diff
@@ -202,6 +205,7 @@ def __init__(
             likeable: Whether the chat messages display a like or dislike button. Set automatically by the .like method but has to be present in the signature for it to show up in the config.
             layout: If "panel", will display the chatbot in a llm style layout. If "bubble", will display the chatbot with message bubbles, with the user and bot messages on alterating sides. Will default to "bubble".
             placeholder: a placeholder message to display in the chatbot when it is empty. Centered vertically and horizontally in the Chatbot. Supports Markdown and HTML. If None, no placeholder is displayed.
+            examples: A list of example messages to display in the chatbot. The `main_text` field is required, and the `additional_input` field is optional. The `additional_input` field can be a string or a `FileMessage` object.
```
abidlabs, line 208: Docstring seems outdated

gradio/components/chatbot.py
abidlabs: Sorry one more thing, why are we expecting users to provide FileData instead of just the path directly? FileData is mostly an implementation details that users should not worry about. So I actually think it should be like this^

```py
examples=[{"text": "Try this example with this audio.", "file": {"path": "files/cantina.wav"}}, {"text": "Try this example with this image.", "file": {"path": "files/avatar.png"}}, {"text": "This is just text, no files!"}],
```

cc @freddyaboulton as well since `FileMessage` has the same issue

abidlabs:
Left some suggestions above.

The UI looks good to me, however a few nits:

(1) I'd recommend setting a `min-height` so that the height of all of the suggestion boxes is the same unless the content is particularly long vertically

(2) When there are a large number of suggestion boxes, I can scroll to see more boxes. However, the whole background of the chatbot scrolls, including the placeholder -- only the suggestion boxes should scroll:

image

(3) The suggestion boxes are not legible in light mode:

image

demo/chatbot_multimodal/run.py
abidlabs, line 1: I would revert the changes in this demo since this demo is used in our guides to explain multimodal, and these additional changes complicate the demo a bit. We can create a new demo to showcase suggestions.

demo/chatinterface_multimodal/run.py
abidlabs, line 1: Same, I would revert the changes in this demo since this demo is used in our guides to explain multimodal, and these additional changes complicate the demo a bit. We can create a new demo to showcase suggestions.

demo/html_component/run.py
```diff
@@ -1,6 +1,6 @@
 import gradio as gr
 
 with gr.Blocks() as demo:
-    gr.HTML(value="<p style='margin-top: 1rem, margin-bottom: 1rem'>This <em>example</em> was <strong>written</strong> in <a href='https://en.wikipedia.org/wiki/HTML' _target='blank'>HTML</a> </p>")
+    gr.HTML(label="testqwfasdf", value="<p style='margin-top: 1rem, margin-bottom: 1rem'>This <em>example</em> was <strong>written</strong> in <a href='https://en.wikipedia.org/wiki/HTML' _target='blank'>HTML</a> </p>", show_label=True)
```
abidlabs, line 4: Seems unrelated?

gradio/chat_interface.py
```diff
@@ -143,8 +146,9 @@ def __init__(
         self.buttons: list[Button | None] = []
 
         self.examples = examples
-        self.cache_examples = cache_examples
 
+        if cache_examples is not None:
+            raise NotImplementedError()
```
abidlabs, line 151:
```suggestion
            raise NotImplementedError(). # TODO: needs to be implemented
```

gradio/chat_interface.py
```diff
@@ -66,9 +71,8 @@ def __init__(
         textbox: Textbox | MultimodalTextbox | None = None,
         additional_inputs: str | Component | list[str | Component] | None = None,
         additional_inputs_accordion: str | Accordion | None = None,
-        examples: list[str] | list[dict[str, str | list]] | list[list] | None = None,
         cache_examples: bool | Literal["lazy"] | None = None,
-        examples_per_page: int = 10,
+        examples: list[SuggestionMessage] | None = None,
```
abidlabs, line 75: We should make sure to document the `SuggestionMessage` class

gradio/components/chatbot.py
```diff
@@ -84,6 +84,13 @@ class Message(GradioModel):
     content: Union[str, FileMessage, ComponentMessage]
 
 
+class SuggestionMessage(TypedDict):
+    icon: NotRequired[str]
+    display_text: NotRequired[str]
+    text: str
```
abidlabs, line 90: Should text be required? What if the suggestionmessage is just an image for example?

gradio/components/chatbot.py
```diff
@@ -198,6 +206,7 @@ def __init__(
             likeable: Whether the chat messages display a like or dislike button. Set automatically by the .like method but has to be present in the signature for it to show up in the config.
             layout: If "panel", will display the chatbot in a llm style layout. If "bubble", will display the chatbot with message bubbles, with the user and bot messages on alterating sides. Will default to "bubble".
             placeholder: a placeholder message to display in the chatbot when it is empty. Centered vertically and horizontally in the Chatbot. Supports Markdown and HTML. If None, no placeholder is displayed.
+            suggestions: A list of suggestion messages to display in the chatbot. The `main_text` field is required, and the `additional_input` field is optional. The `additional_input` field can be a string or a `FileMessage` object.
```
abidlabs, line 209: I think this documentation is incorrect. Also, I think it would be easier for users to understand if it was worded like this:
```suggestion
            suggestions: A list of suggestion messages to display in the chatbot before any user/assistant messages are shown. Each suggestion should be a dictionary with a required "text" key representing the message that should be populated in the Chatbot when clicked, an optional "files" key, whose value should be a list of files to populate in the Chatbot, ...
```

abidlabs:
Looks really nice @dawoodkhan82! I'll do some more testing tomorrow, but in the meantime, just one UI suggestion: can we add a maximum width and maximum height to the suggestion boxes, and if the text overflows those bounds, then we just cut it off with an "..."? To prevent extremely wide suggestions like this:

image

abidlabs:
Other than that, let's also update the ChatInterface guide to talk about examples and how they should be provided.

gradio/chat_interface.py
```diff
@@ -96,9 +100,8 @@ def __init__(
...
             additional_inputs_accordion: if a string is provided, this is the label of the `gr.Accordion` to use to contain additional inputs. A `gr.Accordion` object can be provided as well to configure other properties of the container holding the additional inputs. Defaults to a `gr.Accordion(label="Additional Inputs", open=False)`. This parameter is only used if `additional_inputs` is provided.
-            examples: sample inputs for the function; if provided, appear below the chatbot and can be clicked to populate the chatbot input. Should be a list of strings if `multimodal` is False, and a list of dictionaries (with keys `text` and `files`) if `multimodal` is True.
             cache_examples: if True, caches examples in the server for fast runtime in examples. The default option in HuggingFace Spaces is True. The default option elsewhere is False.
-            examples_per_page: if examples are provided, how many to display per page.
+            examples: A list of examples messages to display in the chatbot. The `main_text` field is required, and the `additional_input` field is optional. The `additional_input` field can be a string or a `FileMessage` object.
```
abidlabs, line 104: This documentation looks incorrect. (See my comment below on chatbot.py). For backwards compatibility, can we also accept a list of strings, and just convert them to a list of dicts with the "text" keys?

demo/chatinterface_multimodal/run.py
```diff
@@ -5,7 +5,7 @@ def echo(message, history):
 
 demo = gr.ChatInterface(
     fn=echo,
-    examples=[{"text": "hello"}, {"text": "hola"}, {"text": "merhaba"}],
+    examples=[{"icon": "files/avatar.png", "display_text": "Display Text Here!", "text": "Try this example with this audio.", "files": ["files/cantina.wav"]}, {"text": "Try this example with this image.", "files": ["files/avatar.png"]}, {"text": "This is just text, no files!"}, {"text": "Try this example with this image.", "files": ["files/avatar.png", "files/avatar.png"]}, {"text": "Try this example with this Audio.", "files": ["files/cantina.wav"]}],
```
freddyaboulton, line 8: nit but I would use the full path here otherwise you have to run the demo from this directory. or make a note that it must be run for this directory

gradio/components/chatbot.py
freddyaboulton: Why not call these `examples` to be consistent with our other examples?

gradio/components/chatbot.py
```diff
@@ -250,6 +259,30 @@ def __init__(
...
+                if "icon" in suggestion:
+                    suggestion["icon"] = self.serve_static_file(suggestion["icon"])
+                file_info = suggestion.get("files")
+                if file_info is not None and not isinstance(file_info, list):
+                    raise Error(
+                        "Data incompatible with files format. The 'files' passed should be a list of file paths or URLs."
+                    )
+                files = []
+                if file_info is not None:
+                    for x, file in enumerate(file_info):
```
freddyaboulton, line 274: mega nit: x is a uncommon name for an integer index

gradio/components/chatbot.py
```diff
@@ -250,6 +259,30 @@ def __init__(
...
+                        if file is not None:
```
freddyaboulton, line 275: What's the use case for allowing `None` files? I don't think we should allow. Seems confusing.

gradio/chat_interface.py
freddyaboulton: I think it would be good to only accept dicts as it would prep users for when "messages" is the only allowed type

js/chatbot/shared/ChatBot.svelte
```diff
@@ -178,6 +180,18 @@
 	}
 
 	$: groupedMessages = value && group_messages(value);
+	$: if (value !== null && value.length > 0) suggestionsVisible = false;
```
freddyaboulton, line 183: I believe `$: suggestionsVisible = !(value !== null && value.length > 0)` would be better as it would automatically be set to True if someone clears out the chat history with the new clear icon being added in #9201

freddyaboulton (commented):
Nice PR @dawoodkhan82 ! Agree with most of @abidlabs 's comments and left some small nits/suggestions. I wonder if rather than having each suggestion be a box of a different size, it would be cleaner if each box was the same size?

If there are more than about 5 suggestions, it looks a bit off to have some squish

image

freddyaboulton:
Overall should mesh well with #9201

gradio/chat_interface.py
dawoodkhan82: Agreed, not sure we still want to support just strings anymore.

abidlabs: Ok I don't feel too strongly, but I would then just have a special error if a user passes in a list of strings telling them that format has been deprecated and they have to now pass in a list of dictionaries with key "text". Because I'm sure this will trip up users who are migrating to 5.0

gradio/components/chatbot.py
dawoodkhan82: We want to eventually use this same API for showing alternative suggestion messages under bot messages

dawoodkhan82:
> Nice PR @dawoodkhan82 ! Agree with most of @abidlabs 's comments and left some small nits/suggestions. I wonder if rather than having each suggestion be a box of a different size, it would be cleaner if each box was the same size?

Added a width to the suggestions. Looks a lot better imo.

abidlabs (approved):
LGTM @dawoodkhan82! 

Just a very tiny nit. In light mode, the border color is orange while the interior is gray. This looks off to me -- either we should make the interior a subdued orange or the border color a darker gray:

image

Otherwise, looks ready to go! Good stuff

abidlabs:
Let's make sure to create an issue to handle cached examples so we don't forget it before launch

js/chatbot/shared/ChatBot.svelte
```diff
@@ -492,19 +506,136 @@
...
+								<Image
+									class="suggestion-image"
+									src={suggestion.files[0].url}
+									alt="suggestion-image"
+								/>
+							{:else if suggestion.files !== undefined && suggestion.files[0] !== undefined}
+								<span class="suggestion-file"
+									><em>{suggestion.files[0].orig_name}</em></span
+								>
+							{/if}
```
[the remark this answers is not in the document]
dawoodkhan82: fixed

aliabid94:
Great work @dawoodkhan82! Two minor opinions:
- Clicking an example should not require a backend transaction to load the example into the Chatbot, it should happen automatically that that example clicked becomes the value of the chatbot.
- Users should be allowed to pass just a string as a suggestion, along with the Suggestion dictionary type.

I've been playing around with it locally, I might push the changes above to this branch of that's okay. Also might tweak the visuals a bit.

Also, I think we should call it ChatExample instead of Suggestion, and `chatbot.load_example` instead of `chatbot.suggestion_select`? We use the term Example for examples and I think it should be consistent.

aliabid94:
Screen issues, I'll fix with flexbox and push directly.
Recording 2024-09-11 at 16 26 19

dawoodkhan82:
> Clicking an example should not require a backend transaction to load the example into the Chatbot, it should happen automatically that that example clicked becomes the value of the chatbot.

@aliabid94 We want to automatically trigger a submit and get a response when an example is clicked. 

> Also, I think we should call it ChatExample instead of Suggestion

We landed on `suggestion` because we want to use the same API to show buttons in the chatbot for alternate "suggestions"

dawoodkhan82:
The tests failing are also failing on `5.0-dev` for me.

abidlabs:
Which tests? Python backend tests seem to be passing here: https://github.com/gradio-app/gradio/pull/8843, although I haven't taken a deep look

dawoodkhan82:
`test_chat_interface` tests. Atlease they're failing for me locally on `5.0-dev`

aliabid94:
> We landed on suggestion because we want to use the same API to show buttons in the chatbot for alternate "suggestions"

As in a pre-generated "quick response"? Why would those have an icon, or a display text? It's actually very unlikely they'd even have a file. A quick response would almost certainly be pure text and nothing else

dawoodkhan82:
@aliabid94 Yeah I guess it made more sense before I added display icon and display text. Not opposed to renaming this from  suggestions -> examples. We can rethink the chatbot button API later.

abidlabs:
Yeah ok sounds good to me too

aliabid94:
I've been playing with the suggestions PR all day, and now I'm wondering why it has a different object type than the value of the Chatbot component in the first place. Why can we use the same type as an item in chatbots value? So for example:
```python
gr.Chatbot(value=[["what is your name?", None]])
gr.Chatbot(value=[[gr.Image("lion.jpg"), None], ["what is the animal in the pic?", None]])
gr.Chatbot(value=[[gr.File("a.json"), None], [gr.File("b.json"), None], ["diff these two jsons", None]])
```
correspond to a chatbot with these suggestions
```python
gr.Chatbot(suggestions=[
  "what is your name?",
  [gr.Image("lion.jpg"), "what is the animal in the pic?"]
  [gr.File("a.json"), gr.File("b.json"), "diff these two jsons"]
])
```
Now we don't have to introduce a new API, the suggestions have the user messages in intended order, and the file and component types of the suggestions can be directly inferred, saving us a lot of logic!

dawoodkhan82:
@aliabid94 Makes sense. Agreed that we can support components and that would simplify the logic. Would still be great to have display icon and text support too. Are you working on these changes? Or should I

aliabid94:
Pushed the following visual changes to this branch directly in https://github.com/gradio-app/gradio/pull/8966/commits/bafc39e5b9ba3c93cb156b1aa6cbe8a9f1b02e34:

- Suggestions fit in chatbot at any screen size
- Suggestions are arranged slightly differently, take more horizontal space when possible
- resize() height calculation of textbox and multimodal textbox corrected to use actual theme size vales rather than hardcoded numbers. also made them slightly smaller in height when with a submit button.
- Made multimodal block border fainter to match all other blocks

Screenshot 2024-09-16 at 10 15 07 AM
Screenshot 2024-09-16 at 10 14 54 AM

js/chatbot/shared/ChatBot.svelte
```diff
@@ -562,51 +534,63 @@
...
+
+	.placeholder {
+		align-items: center;
+		display: flex;
+		justify-content: center;
+		height: 100%;
+		flex-grow: 1;
 	}
 
-	.suggestions :global(img) {
```
dawoodkhan82: Adding this back, otherwise clicking on an image in examples makes it fullscreen.

abidlabs (approved):
Tested a fixed a few things I noticed, but otherwise LGTM @dawoodkhan82 @aliabid94

Let's write some docs as well as parametrized tests next week: multimodal X message_format X whether example labels/icons are provided X caching=True/False

PR #8967, opened 2024-08-01
Set `content_disposition_type` and `media_type` on downloaded files

aliabid94:


abidlabs:
I think we also need to add to:

* RangedFileResponse in the /file route
* StreamingResponse in the /proxy route

abidlabs (approved):
LGTM thanks @aliabid94!

PR #8974, opened 2024-08-02
Allow empty value in gr.Number 

### LINKED ISSUE #6728: Setting the value of the Number component in Gradio to None results in it being displayed as 0 in the frontend.

The Gradio version I am currently using is 3.50.2. I have noticed that the value variable of the Number component in Gradio can accept float, Callable, and None. However, when I set the value to None, the frontend always forcibly changes the value to 0. I'm not sure if this is the intended behavior of Gradio. In some cases, None is necessary.

```py
class Number(...):
...
    def __init__(
        self,
        value: float | Callable | None = None,
```

For example, I use the Gradio Number component to set the model's parameter `length_penalty`, and when this value is set to `None`, it means that this `parameter is not used`. If the Gradio Number component is set to `0`, it will cause the model to set the `length_penalty` parameter to `0`, which may affect the results of the model.

Currently, I can set the value to Callable, where Callable returns None. I have tested this, and the frontend displays it as blank, which aligns with my requirements.

```py
gr.Number(label="Length Penalty", value=getLengthPenaltyValue)
def getLengthPenaltyValue():
    if config.length_penalty is None:
        return None
    else:
        return float(config.length_penalty)
```

Although the above solution has improved my issue, it is still somewhat inconvenient. I wonder if Gradio's official stance on setting the value of the Number component to None can be changed to accept None instead of converting it to 0. Thank you.

**Additional context**  
N/A

hannahblair:
## Description

We should allow empty values to be shown on the frontend in gr.Number when its None/null. I've not changed the data type of `value` itself but rather created a new variable which is used for displaying the value as a empty string when the value is null, and it updates whenever value updates. 

Demo:

```
import gradio as gr

with gr.Blocks() as demo:
    gr.Number(value=None)

demo.launch()
```

Closes: #6728

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/gradio/components/number.py
+++ b/gradio/components/number.py
@@ -121,17 +121,13 @@ class Number(FormComponent):
         """
         if payload is None:
             return None
-        try:
-            value = float(payload)
-        except ValueError:
-            return None
 
-        if self.minimum is not None and value < self.minimum:
-            raise Error(f"Value {value} is less than minimum value {self.minimum}.")
-        elif self.maximum is not None and value > self.maximum:
-            raise Error(f"Value {value} is greater than maximum value {self.maximum}.")
+        if self.minimum is not None and payload < self.minimum:
+            raise Error(f"Value {payload} is less than minimum value {self.minimum}.")
+        elif self.maximum is not None and payload > self.maximum:
+            raise Error(f"Value {payload} is greater than maximum value {self.maximum}.")
 
-        return self._round_to_precision(value, self.precision)
+        return self._round_to_precision(payload, self.precision)
 
     def postprocess(self, value: float | int | None) -> float | int | None:
         """
@@ -141,10 +137,6 @@ class Number(FormComponent):
             The (optionally rounded) field value as a `float` or `int` depending on `precision`.
         """
         if value is None:
-            return None
-        try:
-            value = float(value)
-        except ValueError:
             return None
         return self._round_to_precision(value, self.precision)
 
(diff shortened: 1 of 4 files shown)


gradio/components/number.py
```diff
@@ -121,12 +121,14 @@ def preprocess(self, payload: float | None) -> float | int | None:
...
             return None
-        elif self.minimum is not None and payload < self.minimum:
+
+        if self.minimum is not None and payload < self.minimum:
             raise Error(f"Value {payload} is less than minimum value {self.minimum}.")
         elif self.maximum is not None and payload > self.maximum:
             raise Error(
                 f"Value {payload} is greater than maximum value {self.maximum}."
             )
+
```
abidlabs, line 131: IIUC this change should have no effect since the first branch terminates in a return statement, am I misunderstanding?

js/number/Index.svelte
```diff
@@ -96,7 +102,7 @@
 		border: var(--input-border-width) solid var(--input-border-color);
 		border-radius: var(--input-radius);
 	}
-	input[type="number"] {
+	input[type="text"] {
```
abidlabs, line 105: This means that browsers won't treat this field as numeric. On Desktop it means we'll lose these:

image

On mobile, I think it will lead to a ux degradation since the numeric keyboard won't pop up anymore

abidlabs: I don't think this change should be needed since at least according to the original issue, it is in theory possible to set a numeric field blank:

> Currently, I can set the value to Callable, where Callable returns None. I have tested this, and the frontend displays it as blank, which aligns with my requirements.

PR #8975, opened 2024-08-02
Prevent overflow in Model3D

hannahblair:
## Description

Ensures that if a background is set, the border radius of the Model3D component is still respected, and prevents any overflow.

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/js/model3D/shared/Model3DUpload.svelte
+++ b/js/model3D/shared/Model3DUpload.svelte
@@ -140,6 +140,8 @@
 		align-items: center;
 		width: var(--size-full);
 		height: var(--size-full);
+		border-radius: var(--block-radius);
+		overflow: hidden;
 	}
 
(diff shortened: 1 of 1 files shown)


abidlabs:
I actually don't see the issue in the original Space anymore: 

image

Were you able to confirm the issue locally @hannahblair? If so can you share a repro?

abidlabs:
Oh this is good to fix too, but I thought the original issue was referring to the image icons being cut off (the icons under the cat image here):

image

aliabd (commented):
I still see overflow in the input 

Screen Shot 2024-08-02 at 5 47 26 PM

I think you need to add the same rules to `.input-model` in `Model3DUpload.svelte`

aliabd (approved):
LGTM! :)

PR #8976, opened 2024-08-02
Streaming out tweaks

freddyaboulton:
## Description

Some minor tweaks to streaming out case I overlooked:
- Adaptive max chunk length
- Need to `end_stream` so that the client can stop requesting the playlist file

## 🎯 PRs Should Target Issues

## Tests


abidlabs:
Thanks @freddyaboulton. What can I do to test these changes?

gradio/route_utils.py
```diff
@@ -867,13 +867,15 @@ def __init__(self):
         self.ended = False
         self.segment_index = 0
         self.playlist = "#EXTM3U\n#EXT-X-PLAYLIST-TYPE:EVENT\n#EXT-X-TARGETDURATION:10\n#EXT-X-VERSION:4\n#EXT-X-MEDIA-SEQUENCE:0\n"
+        self.max_length = 5
```
abidlabs, line 870: Why do we hardcode a max length here?

freddyaboulton: It can't be undefined in the http spec but this will be overwritten as soon as we start receiving data

freddyaboulton:
Just saw this @abidlabs  - you can run a demo in which the audio chunk size is larger than 3 seconds. On `5.0-dev`, the the audio player will skip the first chunk.

Sample demo

```python
import gradio as gr
import math
import time
import numpy as np
from pydub import AudioSegment
import io


def numpy_to_mp3(audio_array, sampling_rate):
    # Normalize audio_array if it's floating-point
    if np.issubdtype(audio_array.dtype, np.floating):
        max_val = np.max(np.abs(audio_array))
        audio_array = (audio_array / max_val) * 32767  # Normalize to 16-bit range
        audio_array = audio_array.astype(np.int16)

    # Create an audio segment from the numpy array
    audio_segment = AudioSegment(
        audio_array.tobytes(),
        frame_rate=sampling_rate,
        sample_width=audio_array.dtype.itemsize,
        channels=1
    )

    # Export the audio segment to MP3 bytes - use a high bitrate to maximise quality
    mp3_io = io.BytesIO()
    audio_segment.export(mp3_io, format="mp3", bitrate="320k")

    # Get the MP3 bytes
    mp3_bytes = mp3_io.getvalue()
    mp3_io.close()

    return mp3_bytes

def stream(audio, chunk_length_s):
    start_time = time.time()
    sampling_rate, array = audio
    chunk_length = int(chunk_length_s * sampling_rate)
    time_length = chunk_length_s / 2  # always stream outputs faster than it takes to process
    audio_length = len(array)
    num_batches = math.ceil(audio_length / chunk_length)

    for idx in range(num_batches):
        start_pos = idx * chunk_length
        end_pos = min((idx + 1) * chunk_length, audio_length)
        chunk = array[start_pos : end_pos]
        chunk_mp3 = numpy_to_mp3(chunk, sampling_rate=sampling_rate)
        if idx == 0:
            first_time = round(time.time() - start_time, 2)
        run_time = round(time.time() - start_time, 2)
        yield chunk_mp3, first_time, run_time, idx

with gr.Blocks() as demo:
    gr.Markdown("# New Audio Streaming Out 🔊")
    with gr.Row():
        with gr.Column():
            audio_in = gr.Audio(value="librispeech.wav", sources=["upload"], type="numpy")
            chunk_length = gr.Slider(minimum=2, maximum=10, value=2, step=2, label="Chunk length (s)")
            run_button = gr.Button("Stream audio")
        with gr.Column():
            audio_out = gr.Audio(streaming=True, autoplay=True, format="wav", label="mp3")
            with gr.Group():
                with gr.Row():
                    first_time = gr.Textbox(label="Time to first chunk (s)")
                    run_time = gr.Textbox(label="Time to current chunk (s)")
                    chunk_index = gr.Number(value=0, label="Chunk Index")

    run_button.click(fn=stream, inputs=[audio_in, chunk_length], outputs=[audio_out, first_time, run_time, chunk_index])

demo.launch()
```

Sample mp3 file (unzip first)
[librispeech.mp3.zip](https://github.com/user-attachments/files/16475491/librispeech.mp3.zip)

gradio/route_utils.py
abidlabs: Interesting

abidlabs:
Makes sense! Can repro the original issue and fix, thanks @freddyaboulton

abidlabs (approved):

freddyaboulton:
Thank you @abidlabs !

aidevmin:
@freddyaboulton 
Thank you for pull request. I have a experiment. I want to recording audio from my mic and I want to save audio after each 5 minutes. As my understanding, the default chunk size from streaming is 5 ms. It means that after each 5 ms, prediction function will be called and check whether accumulated audio length is greater than 5 minutes and save audio if the condition is met. I run experiments many time, but sometimes, for example, after 100 s, prediction function is not called, it take some time for prediction function is called again, for example 20 s. The audio length is 120 s until prediction function is called again, but actually the audio length is 100 s ==> loss audio.

How this problem is happened?

freddyaboulton:
Hi @aidevmin ! Can you please verify by installing the latest version of 5.0 from this pull request: https://github.com/gradio-app/gradio/pull/8843?

image

> As my understanding, the default chunk size from streaming is 5 ms.

The default is 0.5 seconds (500 ms). You can configure this with the `stream_frequency` parameter in 5.0, e.g. `audio.stream(..., stream_every=<number of seconds>)`

PR #8978, opened 2024-08-02
Improve url downloads for file objects

aliabid94:


freddyaboulton (commented):
Proposed changes look good to me (left a minor suggestion for the url safety function) but I think this should target 5.0-dev

gradio/processing_utils.py
```diff
@@ -259,21 +268,33 @@ def save_file_to_cache(file_path: str | Path, cache_dir: str) -> str:
 
     return full_temp_file_path
 
+def check_public_url(url: str):
+    hostname = urlparse(url).hostname
+    if not hostname:
+        raise httpx.RequestError(f"Invalid URL: {url}")
+    ip = socket.gethostbyname(hostname)
```
abidlabs, line 275: I wasn't familiar with this function, so looked it up: https://docs.python.org/3/library/socket.html

> Translate a host name to IPv4 address format. The IPv4 address is returned as a string, such as '100.50.200.5'. If the host name is an IPv4 address itself it is returned unchanged. See [gethostbyname_ex()](https://docs.python.org/3/library/socket.html#socket.gethostbyname_ex) for a more complete interface. [gethostbyname()](https://docs.python.org/3/library/socket.html#socket.gethostbyname) does not support IPv6 name resolution, and [getaddrinfo()](https://docs.python.org/3/library/socket.html#socket.getaddrinfo) should be used instead for IPv4/v6 dual stack support.

So I think we'll need to switch to `getaddrinfo` instead

gradio/processing_utils.py
```diff
@@ -176,9 +180,11 @@ def encode_pil_to_bytes(pil_image, format="png"):
             pil_image.save(output_bytes, format, **params)
         return output_bytes.getvalue()
 
+hash_seed = str(uuid.uuid4()).encode("utf-8")
```
abidlabs, line 183: Right now, `hash_seed` is refreshed every time someone restarts their Python session and reimports Gradio. I think this could lead to a lot of unnecessary files created locally. Imagine someone has Automatic1111 running locally and they launch the app with a bunch of images. New temporary copies will get saved each time the app is launched.

Instead, what about creating a `hash_seed` when Gradio is first installed and saving it in `launches.json` (or a different file name but the point is that it doesn't change for the user while Gradio remains installed)

gradio/processing_utils.py
```diff
@@ -259,21 +268,33 @@ def save_file_to_cache(file_path: str | Path, cache_dir: str) -> str:
 
     return full_temp_file_path
 
+def check_public_url(url: str):
+    hostname = urlparse(url).hostname
```
abidlabs, line 272: Btw sometimes Gradio is served on a url that is a raw ip address. Will this method work properly in that case? We should write unit tests for this function

abidlabs:
Btw @aliabid94, there's another PR by a contributor working on this same issue: https://github.com/gradio-app/gradio/pull/8301/files. I'll close that PR, but if its possible to credit @mvlttt as part of this PR, that'd be great

abidlabs:
> Proposed changes look good to me (left a minor suggestion for the url safety function) but I think this should target 5.0-dev

Agreed, there's some discussion here: https://github.com/gradio-app/gradio/pull/8301 about potential breaks this change could have

aliabid94:
addressed everything, ready for re-review

.github/workflows/previews-deploy.yml
abidlabs, line 1: We should remove these workflow changes. I think they got picked up from when this PR was targeting `main`

test/test_processing_utils.py
```diff
@@ -404,3 +405,23 @@ async def test_json_data_not_moved_to_cache():
...
+        "https://172.16.0.1",
+    ],
+)
+def test_various_local_urls(url):
+    with pytest.raises(httpx.RequestError, match="Non-public IP address found"):
+        processing_utils.check_public_url(url)
+
+
+def test_valid_public_url():
+    assert processing_utils.check_public_url("https://www.example.com")
```
abidlabs, line 427: Can we also test a valid public IP address works?

abidlabs: imo would be cleaner to combine the two unit tests by adding a second argument to the test which is a boolean argument indicating whether the url should be allowed or not.

aliabid94: what do you mean? this is an example of a domain that points to a public ip address that works. If you're saying a domain that points to a private address (so that this fails) that's very hard to set up, we'd need to set up a domain for that.

gradio/processing_utils.py
```diff
@@ -260,20 +270,51 @@ def save_file_to_cache(file_path: str | Path, cache_dir: str) -> str:
...
+    if parsed_url.scheme not in ["http", "https"]:
+        raise httpx.RequestError(f"Invalid URL: {url}")
+    hostname = parsed_url.hostname
+    if not hostname:
+        raise httpx.RequestError(f"Invalid URL: {url}")
+
+    try:
+        addrinfo = socket.getaddrinfo(hostname, None)
+    except socket.gaierror:
+        raise httpx.RequestError(f"Cannot resolve hostname: {hostname}") from None
```
abidlabs, line 284: I was thinking about whether this would cause any issues if there's no internet connection, but I believe socket.getaddrinfo works for localhost addresses even without internet connections so this should be fine

abidlabs (approved):
LGTM @aliabid94, nice tests!

Issue #9002, opened 2024-08-05
Pass gr.Request if type hint is `Request | None`


catwell:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  

Currently the `request` object is passed [only if the type hint is *exactly* `Request`](https://github.com/gradio-app/gradio/blob/a043b0c985855d415d9692dcbaee6271bbe9e9cc/gradio/helpers.py#L905). However when computing cached examples `None` will be passed (with a warning, which you can fix by passing `None` explicitly). This means the typing of your function *must* be incorrect for this to work (i.e. it must be `gr.Request` but you must handle the case where it is actually `None`).

**Describe the solution you'd like**  

The `gr.Request` object should be passed when the type hint is `Request | None`. Ideally it would support `request: gr.Request | None = None` and using normal cached examples would not warn in this case.

**Additional context**  

```py
import gradio as gr

def predict(request: gr.Request):
    if request:
        return request.client.host
    else:
        return "No request"

with gr.Blocks() as block:
    text2 = gr.Textbox(interactive=False)
    btn = gr.Button("What is my IP?")

    btn.click(predict, inputs=[], outputs=[text2])

    ex = gr.Examples(
        examples=[[]],
        example_labels=["click me"],
        inputs=[],
        outputs=[text2],
        fn=predict,
        cache_examples=True,
    )

block.queue()
block.launch()
```

The code that prints "No request" should be unreachable, and yet it is called.

![Screenshot from 2024-08-05 18-27-08](https://github.com/user-attachments/assets/35d6d6b1-5e6c-46a8-86a5-2d44e7566c7f)

abidlabs:
Makes sense, will update this

abidlabs:
PR here if you'd like to give it a test (you can install `gradio` from the PR using the automatically-generated links in the PR body)

catwell:
@abidlabs Thanks! I don't see the links you're talking about in the PR body but maybe that's because a runner looks stuck?

![Screenshot from 2024-08-06 09-52-29](https://github.com/user-attachments/assets/8f7839f5-2dba-4570-b23c-744abca25e81)

catwell:
I have tried applying the same patch on my local Gradio copy in my venv and it works, both on my toy sample and my original use case.

PR #9011, opened 2024-08-05
Passes `gr.Request` if type hint is `Request | None`

abidlabs:
Closes: https://github.com/gradio-app/gradio/issues/9002

Test with repro from original issue:

```py
import gradio as gr


def predict(request: gr.Request):
    if request:
        return request.client.host
    else:
        return "No request"


with gr.Blocks() as block:
    text = gr.Textbox(interactive=True)
    text2 = gr.Textbox(interactive=False)
    btn = gr.Button("What is my IP?")

    btn.click(predict, inputs=[], outputs=[text2])

    ex = gr.Examples(
        examples=[["foo", None]],
        example_labels=["click me"],
        inputs=[text],
        outputs=[text2],
        fn=predict,
        cache_examples=False,
    )

block.queue()
block.launch()
```


aliabid94:
Did you mean to make the demo have the function signature:
`def predict(request: gr.Request | None):`
the above works, but
`def predict(request: None | gr.Request):`
doesn't.

Should there be a more general solution?

abidlabs:
Really? Interesting.. what version of Python are you using?

aliabid94:
Python 3.9.6,
code to repro:
```python

import gradio as gr


def predict(request: None | gr.Request):
    if request:
        return request.client.host
    else:
        return "No request"


with gr.Blocks() as block:
    text = gr.Textbox(interactive=True)
    text2 = gr.Textbox(interactive=False)
    btn = gr.Button("What is my IP?")

    btn.click(predict, inputs=[], outputs=[text2])

    ex = gr.Examples(
        examples=[["foo", None]],
        example_labels=["click me"],
        inputs=[text],
        outputs=[text2],
        fn=predict,
        cache_examples=False,
    )

block.queue()
block.launch()
```

abidlabs:
I believe the problem should go away on Python 3.10 or higher, and we’re planning on dropping support for Python 3.8 and 3.9 anyways, but I’ll include a fix here

abidlabs:
Fixed now, agreed its hacky but we can remove all this when upgrade to 5.0

freddyaboulton (approved):
LGTM @abidlabs - thanks for the fix.

abidlabs:
Thanks @freddyaboulton and @aliabid94

Issue #7446, opened 2024-02-15
Chatbot copy all messages


mberco-quandl:
for gradio.Chatbot(...) create a 'show_copy_all_messages' button. It would copy the entire chat history onto clipboard.  This is a useful feature because copying all the messages manually can be difficult given the scrolling nature of the interface. Currently there are two arguments 'show_share_button' and 'show_copy_button' that are close but not quite right. Copying a single message is trivial and omits context, the majority of the time a user wants to send more than a single message. The 'show_share_button' has the right idea, but simply copying to clipboard would be superior to sharing to HugginFaces space. Thank you.

abidlabs:
Thanks @mberco-quandl for creating this issue. Yes we've heard versions of this issue quite a bit, and I think what you're describing makes sense. cc @dawoodkhan82!

awonglk:
I would love to see this enhancement added as well.

PR #9013, opened 2024-08-05
Add copy all messages button to chatbot

hannahblair:
## Description

Adds a copy all button (via a `show_copy_all_button` param) similar to gr.JSON and other components. I'm curious to know if we think this should look more unique to the other copy buttons in Chatbot 🤔 

~Having other messages with copy buttons will look confusing, and scrolling will look terrible if users have copy buttons as well (which I've asked about in Slack)~

We could use a clipboard icon: 

Screenshot 2024-08-06 at 11 03 35

but then we're losing out on consistency between other components. 

Also, should default be true or false?

Screenshot 2024-08-06 at 10 59 47

Test:

```
import gradio as gr

with gr.Blocks() as demo:
    gr.Chatbot(show_copy_all_button=True, show_copy_button=True, value=[["Hello World","Hey Gradio!"],["❤️","😍"],["🔥","🤗"]])

demo.launch()
```

Clipboard output:

```
user: Hello World

assistant: Hey Gradio!

user: ❤️

assistant: 😍

user: 🔥

assistant: 🤗
```

Closes: #7446 

## 🎯 PRs Should Target Issues

## Tests


gradio/components/chatbot.py
```diff
@@ -172,6 +172,7 @@ def __init__(
         likeable: bool = False,
         layout: Literal["panel", "bubble"] | None = None,
         placeholder: str | None = None,
+        show_copy_all_button=False,
```
abidlabs, line 175:
```suggestion
        show_copy_all_button: bool = False,
```

js/chatbot/shared/CopyAll.svelte
```diff
@@ -0,0 +1,78 @@
...
+		if (timer) clearTimeout(timer);
+		timer = setTimeout(() => {
+			copied = false;
+		}, 1000);
+	}
+
+	const copy_conversation = (): void => {
+		if (value) {
+			const conversation_text = value
+				.map((message) => `${message.role}: ${message.content}`)
```
freddyaboulton, line 22: This is going to render images/videos as `[Object Object]`. I think the file url should be used instead. Or maybe some placeholder like `Generated File`

PR #9020, opened 2024-08-05
Some tweaks to is_in_or_equal

freddyaboulton:
## Description

* Consolidates usages of "safe_join" across the code base to use a common util function
* Introduces `DeveloperPath` and `UserProvidedPath` type aliases to make it clearer what each argument of safe join should be
* Adds a test fuzzer for "safe_join"

## 🎯 PRs Should Target Issues

## Tests


gradio/utils.py
```diff
@@ -1466,3 +1459,23 @@ def __len__(self):
 
     def as_list(self):
         return [v for _, v in self.data]
+
+
+def safe_join(directory: DeveloperPath, path: UserProvidedPath) -> str:
```
freddyaboulton, line 1464: Using these custom types so that in the future user provided paths are not accidentally sanitized before being joined

test/test_utils.py
```diff
@@ -369,6 +371,69 @@ def test_is_in_or_equal():
     assert not is_in_or_equal("/safe_dir/subdir/../../unsafe_file.txt", "/safe_dir/")
 
 
+def create_path_string():
```
abidlabs, line 374: very interesting

test/test_utils.py
```diff
@@ -369,6 +371,69 @@ def test_is_in_or_equal():
...
+    except ValueError:
+        return False
+
+
+@given(
+    path_1=create_path_string(),
+    path_2=create_path_string(),
+)
+def test_is_in_or_equal_fuzzer(path_1, path_2):
+    print(path_1, path_2)
```
abidlabs, line 404:
```suggestion
```

test/test_utils.py
abidlabs: Is this deterministic or can we make it so with a seed? Otherwise it'll lead to flaky tests

gradio/routes.py
```diff
@@ -1218,32 +1241,21 @@ async def analytics_dashboard(key: str):
...
 
-def safe_join(directory: str, path: str) -> str:
-    """Safely path to a base directory to avoid escaping the base directory.
-    Borrowed from: werkzeug.security.safe_join"""
-    _os_alt_seps: List[str] = [
-        sep for sep in [os.path.sep, os.path.altsep] if sep is not None and sep != "/"
-    ]
-
+def routes_safe_join(directory: DeveloperPath, path: UserProvidedPath) -> str:
+    """Safely join the user path to the directory while performing some additional http-related checks."""
```
abidlabs, line 1245: Nice this makes sense. Perhaps worth mentioning the checks explicitly here:
```suggestion
    """Safely join the user path to the directory while performing some additional http-related checks, e.g. ensuring that the full path exists on the local file system and is not a directory."""
```

gradio/routes.py
```diff
@@ -543,7 +559,7 @@ async def file(path_or_url: str, request: fastapi.Request):
 
             is_dir = abs_path.is_dir()
 
-            if in_blocklist or is_dir:
+            if is_dir or in_blocklist:
```
abidlabs, line 562: curious why :)

gradio/routes.py
```diff
@@ -1142,20 +1158,27 @@ async def upload_file(
...
+                        DeveloperPath(str(directory)), UserProvidedPath(name)
+                    )
+                except InvalidPathError as err:
+                    raise HTTPException(
+                        status_code=400, detail=f"Invalid file name: {name}"
+                    ) from err
                 temp_file.file.close()
                 # we need to move the temp file to the cache directory
-                # but that's possibly blocking and we're in an async function
+                # but that's possibl y blocking and we're in an async function
```
abidlabs, line 1171:
```suggestion
                # but that's possibly blocking and we're in an async function
```

gradio/routes.py
freddyaboulton: Typo hehe. Originally moved `is_blocklist` to be a separate check but realized we don't have to do that

test/test_routes.py
```diff
@@ -1335,3 +1335,48 @@ def test_docs_url():
...
     finally:
         demo.close()
+
+
+def test_file_access():
+    with gr.Blocks() as demo:
+        gr.Markdown("Test")
+
+    import tempfile
+    from pathlib import Path
```
abidlabs, line 1345: nit: might as well put these imports at the top of the file?

test/test_utils.py
freddyaboulton: I think we can seed in CI but I think the benefit of the random fuzzing is that it can eventually find a case we have overlooked so I won't seed when running locally

gradio/utils.py
```diff
@@ -1090,7 +1083,7 @@ def is_in_or_equal(path_1: str | Path, path_2: str | Path):
         return ".." not in str(relative_path)
     except ValueError:
         return False
-    return True
+    return False
```
abidlabs, line 1086: When is this ever used? It seems like the function should terminate by lines 1083 or 105 at the latest

gradio/utils.py
```diff
@@ -1056,24 +1063,10 @@ def tex2svg(formula, *_args):
 
 
 def abspath(path: str | Path) -> Path:
-    """Returns absolute path of a str or Path path, but does not resolve symlinks."""
-    path = Path(path)
-
-    if path.is_absolute():
-        return path
+    return Path(os.path.abspath(str(path)))
```
abidlabs, line 1066: Ok so the reason we used the [convoluted symlink logic](https://github.com/gradio-app/gradio/pull/3037) is because of this issue: https://github.com/gradio-app/gradio/issues/3024

Basically, we don't want to resolve symlinks but we do want to make them absolute. This fix might work, let me check

test/test_utils.py
abidlabs: 👍 and if we ever find a failure, we can add it to the known edge cases below

gradio/components/file_explorer.py
```diff
@@ -202,11 +203,9 @@ def ls(self, subdirectory: list | None = None) -> list[dict[str, str]] | None:
...
-        absolute_path = os.path.abspath(combined_path)
-        if os.path.commonprefix([self.root_dir, absolute_path]) != os.path.abspath(
-            self.root_dir
-        ):
-            raise ValueError("Attempted to navigate outside of root directory")
-        return absolute_path
+    def _safe_join(self, folders: list[str]):
+        if not folders or len(folders) == 0:
+            return self.root_dir
+        head, tail = folders[0], folders[1:]
```
aliabid94, line 209: What's the purpose of splitting head and tail? Why can't we do `combined_path = UserProvidedPath(os.path.join(*folders))` directly

gradio/components/file_explorer.py
```diff
@@ -85,7 +86,7 @@ def __init__(
             )
             root_dir = root
             self._constructor_args[0]["root_dir"] = root
-        self.root_dir = os.path.abspath(root_dir)
+        self.root_dir = cast(DeveloperPath, os.path.abspath(root_dir))
```
aliabid94, line 89: could we do `DeveloperPath(os.path.abspath(root_dir))` instead of using casting?

gradio/utils.py
abidlabs: Ok so this should work, nice! I would just add a test case in test_utils.py to confirm that symlinks are made absolute but not resolved.

abidlabs (approved):
Awesome PR @freddyaboulton! I went through and reviewed it as thoroughly as I good and everything lgtm. Just a few nits above

gradio/components/file_explorer.py
aliabid94: Also applies to casting in routes.py, can we replace all casting with just directly creating the DeveloperPath with a string?

gradio/components/file_explorer.py
freddyaboulton: It was originally like that but then I ran into errors in the functional tests. I thought it was because `os.path.join` couldn't handle a single argument but it's actually because the list was empty.

So long story short, I think we can use the line you're suggesting since I added the empty list check right above.

gradio/utils.py
freddyaboulton: Yes this is right - I think we can remove but just kept because I figured it was there for a reason lol

aliabid94 (approved):
LGTM, some small comments but great work freddy

freddyaboulton:
Thanks for the reviews! I've addressed all of the comments

PR #9021, opened 2024-08-06
Minor fixes

aliabid94:
Minor issues fixed from here: https://github.com/orgs/gradio-app/projects/5/views/1?filterQuery=assignee%3Aaliabid94

1. Remove html tags function did not work with new lines, fixed now
2. If `enable_monitoring` is explicitly set to False, the traffic monitoring dashboard can not be accessed.


gradio/hash_seed.txt
```diff
@@ -0,0 +1 @@
+d4fd4f26b5c14e5fa534cf5f660155c4
```
freddyaboulton, line 1: Where did this come from?

gradio/hash_seed.txt
abidlabs, line 1: remove

gradio/routes.py
```diff
@@ -1184,6 +1184,10 @@ def robots_txt():
 
         @app.get("/monitoring", dependencies=[Depends(login_check)])
         async def analytics_login(request: fastapi.Request):
+            if not blocks.enable_monitoring:
```
abidlabs, line 1187: can we add a unit test for this?

gradio/blocks.py
```diff
@@ -2180,7 +2181,7 @@ def launch(
         auth_dependency: Callable[[fastapi.Request], str | None] | None = None,
         max_file_size: str | int | None = None,
         _frontend: bool = True,
-        enable_monitoring: bool = False,
+        enable_monitoring: bool | None = None,
```
abidlabs, line 2184: why not just 
```suggestion
        enable_monitoring: bool = True,
```

gradio/blocks.py
```diff
@@ -1018,6 +1018,7 @@ def __init__(
                 t.start()
         else:
             os.environ["HF_HUB_DISABLE_TELEMETRY"] = "True"
+        self.enable_monitoring: bool | None = None
```
abidlabs, line 1021: do we need None? see my comment below

gradio/blocks.py
```diff
@@ -2434,6 +2436,7 @@ def reverse(text):
             print(
                 f"Monitoring URL: {self.local_url}monitoring/{self.app.analytics_key}"
             )
+        self.enable_monitoring = enable_monitoring in [True, None]
```
freddyaboulton, line 2439: I think we can just have this parameter be a bool and default to `True`? Might be clearer than `bool | None` - the only difference with having default `None` is that the link won't get printed to the console automatically but I don't think it's a bad thing if it does.

freddyaboulton (approved):
LGTM thanks @aliabid94 !

abidlabs (approved):
lgtm, see minor comments

abidlabs:
nice @freddyaboulton ;)

gradio/hash_seed.txt
aliabid94: whoops, I added to gitignore but it still somehow got in, will remove

gradio/routes.py
aliabid94: to test what? that the route is blocked?

gradio/blocks.py
aliabid94: True: monitoring enabled, link printed to console
None (default): monitoring enabled, nothing printed
False: monitoring disabled.

The reason why I wanted three options is because by default, we want there to be monitoring. Users are not going to remember to add monitoring, and if for example their app goes viral, they will not have access to any traffic data. However, I don't want True to be default because there's no need to print the monitoring URL every launch, we already have too much text printed on launch. So None was the default option .

gradio/blocks.py
aliabid94: See comment above

gradio/routes.py
abidlabs: Yea

gradio/blocks.py
abidlabs: Ok this should be added to docstring if not there already

Issue #8982, opened 2024-08-04
The gr.JSON requires the lines parameter.


universetao:
- [ Yes] I have searched to see if a similar issue already exists.

Is it possible to configure gr.JSON() with lines to ensure that content overflow does not occur? Previously, I tried using gr.Textbox(inputs=gr.JSON()), but doing this loses the interactive functionality of JSON.

freddyaboulton:
I think this is fixed in version 4.40.0 - the json component will no longer overflow horizontally or vertically

universetao:
My current version is already 4.40. Currently, the JSON is too long and automatically adapts to be particularly long (as shown in the first image). However, I would like to restrict it within a specified box size (like Textbox) while retaining the formatted indentation of the JSON for interaction.

![image](https://github.com/user-attachments/assets/1577ddef-691c-4aa6-8e1b-af7e611090c6)

![image](https://github.com/user-attachments/assets/afdbed3c-378b-4a86-a8cf-65579fc80b3a)

PR #9023, opened 2024-08-06
Add height param to gr.JSON

hannahblair:
## Description

I initially implemented `height` as a param but we have a `lines` param in gr.Code, so it makes some sense to keep them fairly consistent. 

I also added the lines param description in gr.Code that was missing, and added a tiny css tweak to handle big line numbers.

Closes: #8982

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/gradio/components/json_component.py
+++ b/gradio/components/json_component.py
@@ -68,7 +68,7 @@ class JSON(Component):
             key: if assigned, will be used to assume identity across a re-render. Components that have the same key across a re-render will have their value preserved.
             open: If True, all JSON nodes will be expanded when rendered. By default, node levels deeper than 3 are collapsed.
             show_indices: Whether to show numerical indices when displaying the elements of a list within the JSON object.
-            height: Height of the JSON component in pixels. Overflow will be scrollable. If None, the height will be automatically adjusted to fit the content.
+            height: Height of the JSON component in pixels if a number is passed, or in CSS units if a string is passed. Overflow will be scrollable. If None, the height will be automatically adjusted to fit the content.
         """
         super().__init__(
             label=label,


gradio/components/code.py
```diff
@@ -122,6 +122,7 @@ def __init__(
             elem_classes: An optional list of strings that are assigned as the classes of this component in the HTML DOM. Can be used for targeting CSS styles.
             render: If False, component will not render be rendered in the Blocks context. Should be used if the intention is to assign event listeners now but render the component later.
             key: if assigned, will be used to assume identity across a re-render. Components that have the same key across a re-render will have their value preserved.
+            lines: Maximum number of visible lines to show in the code editor. Overflow will be scrollable.
```
abidlabs, line 125: Just tested and it looks like the `lines` parameter actually behaves differently than this. Its more like the `lines` parameter in `gr.Textbox`, which is the *minimum* number of visible lines to show. If the code takes up more than this number of lines, the full code will be shown. We *could* add a `max_lines` parameter, but that's a separate issues

gradio/components/json_component.py
```diff
@@ -67,6 +68,7 @@ def __init__(
             key: if assigned, will be used to assume identity across a re-render. Components that have the same key across a re-render will have their value preserved.
             open: If True, all JSON nodes will be expanded when rendered. By default, node levels deeper than 3 are collapsed.
             show_indices: Whether to show numerical indices when displaying the elements of a list within the JSON object.
+            lines: Maximum number of visible lines to show in the component. Overflow will be scrollable.
```
abidlabs, line 71: In light of the previous comment, I would suggest naming this parameter `max_lines` to be consistent with `gr.Textbox`

abidlabs:
This works well @hannahblair! But I'm wondering if `height` might actually be better for two reasons:

(1) The behavior of `lines` is based on the number of lines of code that the json value takes up when it is rendered in the component, which may not necessarily be the same as the line numbers and could be confusing to users. For example, here I set `lines` to `10` but the json is only displayed to line `7`, since line `2` takes up three lines in the display: 

image


(2) `lines` in `gr.Code` actually behaves differently as commented above. So this parameter would be more like the `max_lines` parameter in `gr.Textbox` or the `height` parameter that we have in most of our other components

gradio/components/json_component.py
```diff
@@ -49,6 +49,7 @@ def __init__(
         key: int | str | None = None,
         open: bool = False,
         show_indices: bool = False,
+        height: int | None = None,
```
abidlabs, line 52: I think we support height as a string (in CSS units) as well, right?

abidlabs (approved):
LGTM @hannahblair! The `height` parameter looks good.

One thing I noticed while testing is that the JSON component, when empty, has its icon pinned to the top:

image

Unlike the other components, which center the icon vertically:

image

Although its not related to this PR, we could perhaps fix it here? Also fine to handle separately

PR #9043, opened 2024-08-07
Filter out type ignore comments from demos on website

aliabd:
simple pr that hides the `# type: ignore` comments from demos on the website in docs, guides, and the playground.


abidlabs:
Looks good to me. 

image

But it looks like there are a couple of other demos that have `#type: ignore` instead of `# type: ignore`:

image

aliabd:
@abidlabs ah, i did site wide search and saw unified demo but not translation (not on website) 

anyway, fixed!

Issue #2298, opened 2022-09-20
Video Gallery Outputs


fffiloni:
[*] I have searched to see if a similar issue already exists.
YES

**Is your feature request related to a problem? Please describe.**  
NO

**Describe the solution you'd like**  
I'd like to be able to display a list of videos outputs as a Gallery, that would work as the images gallery works, but for videos.

**Additional context**  
Context: I made a PySceneDetect demo that can divide a video sequence into every shots it can detect.
Link: [Scene Edit Detection on HuggingFace](https://huggingface.co/spaces/nanomenta/scene-edit-detection)
I use the `gr.List` instead, and display an image Gallery for every first frame of each shot.

abidlabs:
Hey @fffiloni! We've now made it possible for Gradio users to create their own custom components -- meaning that you can write some Python and JavaScript (Svelte), and publish it as a Gradio component. You can use it in your own Gradio apps, or share it so that anyone can use it in their Gradio apps. Here are some examples of custom Gradio components:

* A "Rich Textbox" that [allows you to write bold/italics/colored text](https://huggingface.co/spaces/abidlabs/gradio_rich_textbox)
* A "Folium Map Viewer" component that [allows you to use interactive maps](https://huggingface.co/spaces/freddyaboulton/gradio_folium)

You can see the source code for those components by clicking the "Files" icon and then clicking "src". The complete source code for the backend and frontend is visible. **In particular, its very fast if you want to build off an existing component**. We've put together a Guide: https://www.gradio.app/guides/five-minute-guide, and we're happy to help if you'd like to put together a custom component for this issue

dawoodkhan82:
@fffiloni I published a `video_gallery` component using our new custom component workflow. Feel free to try it out (https://pypi.org/project/gradio-videogallery/)

dawoodkhan82:
@abidlabs After seeing how much thumbs up the original issue has, maybe we should add this functionality to core.

abidlabs:
Instead of adding a new component, could we modify `Gallery` to accept images or videos or a combination?

dawoodkhan82:
@abidlabs yeah for sure, wasn't clear but that's what I meant.

abidlabs:
Let's do it

fffiloni:
yay 🎉

fffiloni:
[...]

PR #9052, opened 2024-08-07
Video gallery

dawoodkhan82:
## Description
Screenshot 2024-08-07 at 12 48 40 PM
Screenshot 2024-08-07 at 12 48 45 PM

Closes: #2298

## 🎯 PRs Should Target Issues

## Tests


dawoodkhan82:
Will add a param, to limit gallery uploads to only images or videos.

gradio/components/gallery.py
```diff
@@ -31,24 +32,30 @@
...
 
 
 class GalleryImage(GradioModel):
     image: FileData
     caption: Optional[str] = None
 
 
+class GalleryVideo(GradioModel):
+    video: FileData
+    caption: Optional[str] = None
```
abidlabs, line 46: Do we support captions in videos in a Gallery? How would a user pass that in?

gradio/components/gallery.py
```diff
@@ -31,24 +32,30 @@
...
+    subtitles: Optional[FileData] = None
```
abidlabs, line 47: Do we support subtitles in videos in a Gallery? How would a user pass that in?

abidlabs:
Works quite well @dawoodkhan82! Just a few things I noticed:

(1) When I would click on a video, it would briefly show a circular loading icon in the center of the video. It feels unnecessary as nothing happens when the video finishes playing, can we just remove it?

https://github.com/user-attachments/assets/059fe291-1bf2-49ef-8435-a8d8b350bba1

(2) Captions don't show up for videos in the gallery mode:

image

They do show up when the video is selected though:

image

It'd be great if we can add the captions to the preview mode as well

> Will add a param, to limit gallery uploads to only images or videos.

Yeah, I think we can have a `file_types` parameter similar to `gr.File` that can be used to control images/videos or even more fine-grained control.

Happy to give it another pass and test events, etc. when ready

gradio/components/gallery.py
dawoodkhan82: Oh yeah will remove, there's no way to pass it in as of now. We can look into adding them later.

dawoodkhan82:
@abidlabs 
- Removed subtitles (it didn't work anyways)
- Added `file_types` param. Defaults to allow both image & video. Maybe it should default to just image. thoughts?
- Fixed upload event bug
- Fixed caption on preview

> (1) When I would click on a video, it would briefly show a circular loading icon in the center of the video. It feels unnecessary as nothing happens when the video finishes playing, can we just remove it?

Not sure we can remove this, it's part of the `htmlvideoelement` controls. Also it only shows for a second or two while the video loads for me, then gets removed automatically.

gradio/components/gallery.py
```diff
@@ -95,8 +102,9 @@ def __init__(
     ):
         """
         Parameters:
-            value: List of images to display in the gallery by default. If callable, the function will be called whenever the app loads to set the initial value of the component.
+            value: List of images or videos to display in the gallery by default. If callable, the function will be called whenever the app loads to set the initial value of the component.
             format: Format to save images before they are returned to the frontend, such as 'jpeg' or 'png'. This parameter only applies to images that are returned from the prediction function as numpy arrays or PIL Images. The format should be supported by the PIL library.
+            file_types: List of file extensions or types of files to be uploaded (e.g. ['image', 'video']). "image" allows only image files to be uploaded, "video" allows only video files to be uploaded, "video" allows only video files to be uploaded, ".mp4" allows only mp4 files to be uploaded, etc. If None, both image and video files types are allowed.
```
abidlabs, line 107:
```suggestion
            file_types: List of file extensions or types of files to be uploaded (e.g. ['image', 'video']). "image" allows only image files to be uploaded, "video" allows only video files to be uploaded, ".mp4" allows only mp4 files to be uploaded, etc. If None, any image and video files types are allowed.
```

gradio/components/gallery.py
```diff
@@ -170,27 +179,31 @@ def preprocess(
     ):
         """
         Parameters:
-            payload: a list of images, or list of (image, caption) tuples
+            payload: a list of images or videos, or list of (media, caption) tuples
         Returns:
-            Passes the list of images as a list of (image, caption) tuples, or a list of (image, None) tuples if no captions are provided (which is usually the case). The image can be a `str` file path, a `numpy` array, or a `PIL.Image` object depending on `type`.
+            Passes the list of images as a list of (media, caption) tuples, or a list of (image, None) tuples if no captions are provided (which is usually the case). The image can be a `str` file path, a `numpy` array, or a `PIL.Image` object depending on `type`. The video can be a `str` file path.
```
abidlabs, line 184:
```suggestion
            Passes the list of images or videos as a list of (media, caption) tuples, or a list of (media, None) tuples if no captions are provided (which is usually the case). Images can be a `str` file path, a `numpy` array, or a `PIL.Image` object depending on `type`.  Videos are always `str` file path.
```

abidlabs (approved):
LGTM, nice @dawoodkhan82! Reviewed the backend + functionality and everything looks good to me. Perhaps @aliabid94 @hannahblair or @pngwn can review the frontend code as well

js/gallery/shared/Gallery.svelte
```diff
@@ -54,16 +56,30 @@
...
-		value == null
-			? null
-			: value.map((data) => ({
-					image: data.image as FileData,
-					caption: data.caption
-				}));
-
-	let prev_value: GalleryData | null = value;
+	let resolved_value: GalleryData[] | null = null;
+	let thumbnailsGenerated = false;
```
freddyaboulton, line 60: @dawoodkhan82 this implementation delivers but why do we need to manually generate thumbnails?

I think we can display videos inside the buttons like so

```svelte
<Video
  src={media.video.url}
  title={media.caption || null}
  is_stream={false}
  data-testid={"thumbnail " + (i + 1)}
  alt=""
  loading="lazy"
  loop={false}
/>
```

dawoodkhan82: Yeah originally had it this way, but thought loading videos would be too intensive. Looking at it again, i think you're right. Let me revert

freddyaboulton (commented):
@dawoodkhan82 thanks for making the change! Code looks good - one last night is that instead of having the `<Play>` button overlayed for videos, we can start playing the video on a mouseover? Like we do for video examples?

Tests are failing but it may be because of recent changes to 5.0-dev. Can take a look again once they pass.

freddyaboulton (approved):
LGTM once tests pass thanks @dawoodkhan82 !

Issue #9042, opened 2024-08-07
Interface() show_progress doesn't work


missionfloyd:
### Describe the bug

Progress in `gr.Interface()` is always shown regardless of what `show_progress` is set to.

`"minimal"`, `"hidden"`, and `False` have no effect.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

demo = gr.Interface(
    fn=lambda x: x,
    inputs=["textbox"],
    outputs=["textbox"],
    show_progress="hidden"
)

demo.launch()
```

### Screenshot

https://github.com/user-attachments/assets/2985f01d-10e2-4c60-9c68-a5e592bbb09b

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Windows
gradio version: 4.40.0
gradio_client version: 1.2.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
anyio: 4.4.0
fastapi: 0.112.0
ffmpy: 0.4.0
gradio-client==1.2.0 is not installed.
httpx: 0.27.0
huggingface-hub: 0.24.5
importlib-resources: 6.4.0
jinja2: 3.1.3
markupsafe: 2.1.5
matplotlib: 3.9.1.post1
numpy: 1.26.3
orjson: 3.10.6
packaging: 24.1
pandas: 2.2.2
pillow: 10.2.0
pydantic: 2.8.2
pydub: 0.25.1
python-multipart: 0.0.9
pyyaml: 6.0.2
ruff: 0.5.6
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.12.3
typing-extensions: 4.9.0
urllib3: 2.2.2
uvicorn: 0.30.5
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2024.2.0
httpx: 0.27.0
huggingface-hub: 0.24.5
packaging: 24.1
typing-extensions: 4.9.0
websockets: 12.0
```

### Severity

I can work around it

PR #9053, opened 2024-08-07
Fix `show_progress` in `gr.Interface`

abidlabs:
Fixes: https://github.com/gradio-app/gradio/issues/9042


freddyaboulton (approved):

PR #9064, opened 2024-08-07
Improve plot guide, add double clicking to plots

aliabid94:
Improved the plot guides a bit. Also added the double_click listener to plots so users can add intuitive zoom outs.


abidlabs:
Haven't taken a deep look at all but a nitty nit: could we use `double_click()` instead of `dbl_click()`? We don't use any abbreviations in any event listener method names, looks stylistically off

aliabd (commented):
changes look good, I uploaded the new demos to spaces. of course some won't work until release because of the new changes (double_click). 

One thing, the double click event doesn't show up in the event listeners section of the plot docs. it does show up in the components introduction page though. 

is it possible to add it to all the plots here? https://github.com/gradio-app/gradio/blob/main/gradio/components/plot.py#L41 seems like its what we do for other events

aliabid94:
> is it possible to add it to all the plots here?

That's for gr.Plot, which does not have double click. LinePlot, BarPlot, and ScatterPlot are NativePlots, which are different.

aliabd:
Oh my bad. Guess I should read these guides

aliabd (approved):

PR #9065, opened 2024-08-07
Have gr.on set value at start as well

aliabid94:
If you have a demo like this:
```python
with gr.Blocks() as demo:
    a = gr.Number(1)
    b = gr.Number(1)
    c = gr.Number()

    @gr.on(inputs=[a,b], outputs=c)
    def add(x,y):
        return x + y

demo.launch()
```

c will only be set to the value of a + b when they change, not initially when the demo runs. Usually users would want the initial value to also be a function of the inputs (when no trigger is specified).

This is also consistent with the behaviour of gr.render and value=fn, e.g. both of these would set c to 2 intitially:

```python
with gr.Blocks() as demo:
    a = gr.Number(1)
    b = gr.Number(1)

    @gr.render(inputs=[a, b])
    def renderfn(x, y):
        gr.Number(x + y)

demo.launch()
```

```python
with gr.Blocks() as demo:
    a = gr.Number(1)
    b = gr.Number(1)
    c = gr.Number(lambda x, y: x + y, inputs=[a, b])

demo.launch()
```
Technically breaking so going into 5.0

### DIFF
--- a/gradio/events.py
+++ b/gradio/events.py
@@ -646,7 +646,7 @@ def on(
             [EventListenerMethod(input, "change") for input in inputs]
             if inputs is not None
             else []
-        )  # type: ignore
+        ) + [EventListenerMethod(root_block, "load")]
     else:
         methods = [
             EventListenerMethod(t.__self__ if t.has_trigger else None, t.event_name)  # type: ignore
--- a/gradio/events.py
+++ b/gradio/events.py
@@ -564,7 +564,7 @@ def on(
     for all events in the triggers list.
 
     Parameters:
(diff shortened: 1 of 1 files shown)


gradio/events.py
```diff
@@ -646,7 +646,7 @@ def inner(*args, **kwargs):
             [EventListenerMethod(input, "change") for input in inputs]
             if inputs is not None
             else []
-        )  # type: ignore
+        ) + [EventListenerMethod(root_block, "load")]  # type: ignore
```
abidlabs, line 649: Cool let's just update the docstring of `triggers` as well

abidlabs:
lgtm one note above

PR #9069, opened 2024-08-08
No token passed by default in `gr.load()`

abidlabs:


aliabd (approved):
if its just the docs change lgtm! you need to lint tho

abidlabs:
And also the parameter default has changed

abidlabs:
The lint and typing issues are actually already on the `5.0-dev` branch, not originating from this PR. They are not hard to fix but I'll wait until @freddyaboulton is back in case there are underlying issues that need addressing

abidlabs:
Will go ahead and merge this in

Issue #8424, opened 2024-05-31
Got stuck when using ffmpeg to covert mp3 to wav with gr.Audio inputs


vicfullstack:
### Describe the bug

I deployed gradio 4.32.0 in Debian 11/12 server. I used gr.Audio as input. When I upload a mp3 file, it got stuck and consumed all the server's resource which result in the hanging of server. 
`inputs=[gr.Audio(sources=["upload", "microphone"], value="samples/sample_audio.mp3", type="filepath")],`
It seems when I upload mp3 file, it will convert into wav automatically by ffmpeg, but the command of ffmpeg running in the background will keep running:
`ffmpeg -y -i xxx.mp3 -acodec pcm_s16le -vn -f wav -`

When I am running gradio 4.32.0 in my MacOS 14.5, it works well. 

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

```

### Screenshot

![image](https://github.com/gradio-app/gradio/assets/48231513/fe098eb3-dbf3-44b8-8842-a988fc4129ff)

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Linux
gradio version: 4.32.0
gradio_client version: 0.17.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.3.0
fastapi: 0.111.0
ffmpy: 0.3.2
gradio-client==0.17.0 is not installed.
httpx: 0.27.0
huggingface-hub: 0.23.2
importlib-resources: 6.4.0
jinja2: 3.1.4
markupsafe: 2.1.5
matplotlib: 3.9.0
numpy: 1.26.4
orjson: 3.10.3
packaging: 24.0
pandas: 2.2.2
pillow: 10.3.0
pydantic: 2.7.2
pydub: 0.25.1
python-multipart: 0.0.9
pyyaml: 6.0.1
ruff: 0.4.6
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.12.3
typing-extensions: 4.12.0
urllib3: 2.2.1
uvicorn: 0.30.0
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2024.5.0
httpx: 0.27.0
huggingface-hub: 0.23.2
packaging: 24.0
typing-extensions: 4.12.0
websockets: 11.0.3
```

[...]

Issue #7378, opened 2024-02-10
Interface crashes with MP3 file for audio


jjshoots:
### Describe the bug

The interface crashes with a SIGILL for longer MP3 files, it works for WAV files (which, of the same length, are like 10x the filesize).

I'm using the Brave Browser v1.62.162.

The audio file in question (this link doesn't immediately download the file): http://jjshoots.ddns.net:8090/share/0ikEkQG0

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

with gr.Blocks() as app:
    audio = gr.Audio(value="AUDIO_FILE_HERE", interactive=False, autoplay=False)

app.launch()
```

### Screenshot

![image](https://github.com/gradio-app/gradio/assets/38184875/2edc7014-69da-4de2-800b-5faa8bfb5dba)

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Linux
gradio version: 4.17.0
gradio_client version: 0.9.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
altair: 5.1.2
fastapi: 0.104.1
ffmpy: 0.3.1
gradio-client==0.9.0 is not installed.
httpx: 0.25.1
huggingface-hub: 0.20.3
importlib-resources: 6.1.1
jinja2: 3.1.2
markupsafe: 2.1.3
matplotlib: 3.8.1
numpy: 1.26.2
orjson: 3.9.10
packaging: 23.2
pandas: 2.1.3
pillow: 10.1.0
pydantic: 2.5.0
pydub: 0.25.1
python-multipart: 0.0.6
pyyaml: 6.0.1
ruff: 0.2.1
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.9.0
typing-extensions: 4.8.0
uvicorn: 0.24.0.post1
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2023.10.0
httpx: 0.25.1
huggingface-hub: 0.20.3
packaging: 23.2
typing-extensions: 4.8.0
websockets: 11.0.3
```

### Severity

I can work around it

abidlabs:
Is still an issue with the latest version of `gradio`?

jjshoots:
@abidlabs Just tried it, yes. Going to update the issue with the new version.

[...]

PR #9073, opened 2024-08-09
Set default `format` in `gr.Audio` to be `None` to avoid unnecessary preprocessing

abidlabs:
In 4.x, the default `format` of `gr.Audio` is `"wav"`, which means that any `.mp3` that is uploaded to `gr.Audio` is converted to ".wav" format. This is not really necessary as far as I know, and has led to some issues since this conversion relies on ffmpeg. For example, see https://github.com/gradio-app/gradio/issues/8424 and https://github.com/gradio-app/gradio/issues/7378

In this PR, I'm changing the default audio format to `None`, which means that no conversion is performed by default. Aside from the issues above, this also makes the `gr.Audio` component consistent with the `gr.Video` component's `format` parameter. I've also improved the performance of `gr.Audio` preprocessing by skipping audio extraction when not needed and fixed docstrings. 

I believe this PR closes: https://github.com/gradio-app/gradio/issues/8424 and closes: https://github.com/gradio-app/gradio/issues/7378


gradio/components/video.py
```diff
@@ -99,41 +99,41 @@ def __init__(
...
-            inputs: Components that are used as inputs to calculate `value` if `value` is a function (has no effect otherwise). `value` is recalculated any time the inputs change.
+            value: path or URL for the default value that Video component is going to take. Can also be a tuple consisting of (video filepath, subtitle filepath). If a subtitle file is provided, it should be of type .srt or .vtt. Or can be callable, in which case the function will be called whenever the app loads to set the initial value of the component.
+            format: the file extension with which to save video, such as 'avi' or 'mp4'. This parameter applies both when this component is used as an input to determine which file format to convert user-provided video to, and when this component is used as an output to determine the format of video returned to the user. If None, no file format conversion is done and the vidoe is kept as is. Use 'mp4' to ensure browser playability.
```
freddyaboulton, line 103: type: `vidoe`

gradio/components/audio.py
```diff
@@ -166,7 +166,12 @@ def __init__(
             raise ValueError(
                 "Audio streaming only available if sources includes 'microphone'."
             )
-        self.format = format
+        valid_formats = ["wav", "mp3"]
+        if format is not None and format.lower() not in valid_formats:
+            raise ValueError(
+                f"Invalid value for parameter `format`: {format}. Please choose from one of: {valid_formats}"
```
freddyaboulton, line 172: nit: "Please choose from one of: {', '.join(valid_formats)}"

freddyaboulton (approved):
Nice @abidlabs thanks!

gradio/components/audio.py
```diff
@@ -224,16 +229,23 @@ def preprocess(
...
-            temp_file_path.with_name(f"{temp_file_path.stem}{temp_file_path.suffix}")
-        )
+        needs_conversion = False
+        original_suffix = Path(payload.path).suffix.lower()
+        if self.format is not None and original_suffix != f".{self.format}":
+            needs_conversion = True
 
-        sample_rate, data = processing_utils.audio_from_file(temp_file_path)
+        # To avoid possibly-unbound errors
+        duration, sample_rate, data = 0, 0, np.array([])
```
aliabid94, line 238: A suggesiton, could we refactor the code so that the logic that uses duration, sample_rate, and data is inside the following if block where they are actually set? Would be a better approach than assigning placeholder values above to avoid unbound-errors, bc this will be prone to accidentally using the incorrect placeholder values

aliabid94 (approved):
Tested, works well thanks. Left a comment otherwise LGTM!

gradio/components/audio.py
abidlabs: makes sense

abidlabs:
Done, thanks for the review folks!

PR #9075, opened 2024-08-09
Add warning to guides and change styling of tip

aliabd:
The ability to add a stylized warning to guides and modifies the styling of tip to be green (warning is red). Also replaces emojis with svgs so they look nicer. 

Screenshot 2024-08-09 at 1 37 05 PM

This pr doesn't actually add a warning anywhere though, so maybe @abidlabs you can? just push to the pr if its easier.


abidlabs:
LGTM @aliabd! I'll add warnings in https://github.com/gradio-app/gradio/pull/9074

PR #9091, opened 2024-08-12
Be able to set optimizeDeps options in gradio.config.js

freddyaboulton:
## Description

Fixes a corner case in pdf component ( https://github.com/freddyaboulton/gradio-pdf/issues/9 ) but I think it would be useful in general.

I tested by copying the build preview index.js file to the pdf components node_modules. And I also tested that not modifying gradio.config.js works as expected.

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/js/preview/src/dev.ts
+++ b/js/preview/src/dev.ts
@@ -60,6 +60,7 @@ export async function create_server({
 			build: {
 				target: config.build.target
 			},
+			optimizeDeps: config.optimizeDeps,
 			plugins: [
 				...plugins(config),
 				make_gradio_plugin({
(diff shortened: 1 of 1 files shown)


js/preview/src/build.ts
```diff
@@ -4,6 +4,7 @@ import { build } from "vite";
 import { plugins, make_gradio_plugin } from "./plugins";
 import type { PreRenderedChunk } from "rollup";
 import { examine_module } from "./index";
+import { optimizeDeps } from "vite";
```
pngwn, line 7: Is this a mistake?

pngwn (approved):
Small comment but looks great otherwise. If we add any more options we might want to do a smarter merge.

Thanks @freddyaboulton!

js/preview/src/build.ts
freddyaboulton: Yes it was autocomplete!

PR #9104, opened 2024-08-13
Fix chatinterface e2e test

freddyaboulton:
## Description

Minor tweak to one of our e2e tests. If the `chatinterface works with streaming functions and all buttons behave as expected` test fails, all subsequent retries will fail because the backend global variable is not reset.

also fixes image in chatbot docs that's outdated 
## 🎯 PRs Should Target Issues

## Tests


demo/test_chatinterface_streaming_echo/run.py
```diff
@@ -2,13 +2,25 @@
...
 
-demo = gr.ChatInterface(slow_echo)
+chat = gr.ChatInterface(slow_echo, fill_height=True)
+
+with gr.Blocks() as demo:
+    chat.render()
+    # We reset the global variable to minimize flakes
+    # this works because CI runs only one test at at time
+    # need to use gr.State if we want to parallelize this test
+    # currently chatinterface does not support that
```
abidlabs, line 22: We could pass in `gr.State` as `additional_inputs` no?

abidlabs (approved):
changes lgtm but I left a note with an alternative proposal

demo/test_chatinterface_streaming_echo/run.py
freddyaboulton: But we can't update it after each run, no?

PR #9106, opened 2024-08-14
fix: gradio will use internal address to load static resource in client in proxy case

ganisback:
## Description
the access flow looks like following, the problem is that gradio will load css or js with internal proxy server address,
the correct behavior is gradio should use public server to load static resources in the client. 

image

img_v3_02dm_a17bb8ad-664b-4095-a4d2-24cbb0446d7g

Closes: #(issue)

## 🎯 PRs Should Target Issues

## Tests

Tested in my cs arch

### DIFF
--- a/.changeset/plenty-ideas-yell.md
+++ b/.changeset/plenty-ideas-yell.md
@@ -1,0 +1,6 @@
+---
+"@gradio/client": minor
+"gradio": minor
+---
+
+feat:fix: gradio will use internal address to load static resource in client in proxy case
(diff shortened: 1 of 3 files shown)


abidlabs:
Hi @ganisback have you set the `x-forwarded-host` header in your proxy? Typically, this should take care of the issue above

abidlabs:
let us know if that suffices because otherwise this PR will have unintended consequences and I'm inclined to close this

ganisback:
@abidlabs I tried nginx settings, it does not work, so I think this is a valid fix

Issue #9107, opened 2024-08-14
Got Unexpected datatype


avcode-exe:
### Describe the bug

I got datatype of `message` to be `<class 'gradio.components.multimodal_textbox.MultimodalData'>` instead of `<class 'dict'>`, hence `messages["text"]` breaks.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import time
import gradio as gr

class response_test:
    def __init__(self):
        self.text = "test"

def chatbot_nightly(message, history):
    print(type(message))  # Got <class 'gradio.components.multimodal_textbox.MultimodalData'>
    print(message)  # Return text='hi' files=[]
    message_text = message["text"]  # Breaks here
    message_files = message["files"]
    
    response_nightly = response_test()

    # Suspect the culprit is the for loop below
    for i in range(len(response_nightly.text)):
        time.sleep(0.05)
        yield response_nightly.text[:i + 1]

gr.ChatInterface(
    chatbot_nightly,
    title='name',
    multimodal=True
).queue().launch(debug=True, show_error=True, share=True)
```

### Screenshot

![image](https://github.com/user-attachments/assets/c832e1ee-d6e8-48f0-be1d-354c9c27bcb8)

### Logs

```shell
To create a public link, set `share=True` in `launch()`.
<class 'gradio.components.multimodal_textbox.MultimodalData'>
text='hi' files=[]
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/gradio/queueing.py", line 536, in process_events
    response = await route_utils.call_process_api(
  File "/usr/local/lib/python3.10/site-packages/gradio/route_utils.py", line 288, in call_process_api
    output = await app.get_blocks().process_api(
  File "/usr/local/lib/python3.10/site-packages/gradio/blocks.py", line 1931, in process_api
    result = await self.call_function(
  File "/usr/local/lib/python3.10/site-packages/gradio/blocks.py", line 1528, in call_function
    prediction = await utils.async_iteration(iterator)
  File "/usr/local/lib/python3.10/site-packages/gradio/utils.py", line 671, in async_iteration
    return await iterator.__anext__()
  File "/usr/local/lib/python3.10/site-packages/gradio/utils.py", line 776, in asyncgen_wrapper
    response = await iterator.__anext__()
  File "/usr/local/lib/python3.10/site-packages/gradio/chat_interface.py", line 651, in _stream_fn
    first_response = await async_iteration(generator)
[...]

PR #9119, opened 2024-08-14
Fix chatinterface multimodal bug

freddyaboulton:
## Description

Closes: #9107

Consolidates some logic across stream and non-stream code-paths to make the code more maintainable.
Also adds more e2e tests for multimodal cases

## 🎯 PRs Should Target Issues

## Tests


gradio/chat_interface.py
```diff
@@ -579,14 +579,12 @@ def response_as_dict(self, response: MessageDict | Message | str) -> MessageDict
...
         self,
         message: str | MultimodalData,
         history_with_input: TupleFormat | list[MessageDict],
-        request: Request,
-        *args,
-    ) -> tuple[TupleFormat, TupleFormat] | tuple[list[MessageDict], list[MessageDict]]:
-        if self.multimodal and isinstance(message, MultimodalData):
+    ) -> tuple[str | dict, TupleFormat | list[MessageDict]]:
+        if isinstance(message, MultimodalData):
             remove_input = (
```
abidlabs, line 588: nit: could refactor to `remove_input = len(message.files) + int(message.text is not None)`

gradio/chat_interface.py
```diff
@@ -579,14 +579,12 @@ def response_as_dict(self, response: MessageDict | Message | str) -> MessageDict
             new_response = response
         return cast(MessageDict, new_response)
 
-    async def _submit_fn(
+    def _format_input(
```
abidlabs, line 582: nit: perhaps we should this method a more descriptive name? it took me a while to understand what its used for. Maybe `process_message_and_trim_history`?

gradio/chat_interface.py
```diff
@@ -597,7 +595,26 @@ async def _submit_fn(
         else:
             history = history_with_input[:-1]
             message_serialized = message
+        return message_serialized, history
 
+    def _append_history(self, history, message, first_response=True):
+        if self.type == "messages":
+            message = self.response_as_dict(message)
+        if self.type == "tuples":
+            history[-1][1] = message  # type: ignore
```
abidlabs, line 604: Wouldn't it be cleaner just to have 
```
        if self.type == "tuples":
            history[-1][1] = message  # type: ignore
       else:
            message = self.response_as_dict(message)
            if first_response: 
               ..
```

abidlabs (approved):
Looks good @freddyaboulton! I can confirm this fixes the issue, and thanks for cleaning up the code. 

That being said, I think we have a lot of message formats for Chatbot now and I'm quite lost myself about how to use everything. For example, how would I modify the chat function [in this issue](https://github.com/gradio-app/gradio/issues/9107) if I wanted to include a file being streamed. Is that possible?

freddyaboulton:
It doesn't have to be just text, it can be any valid Chatbot response, component instances included. 

I think the issue is that chatinterface only allows the bot to send one response at a time. For sending files, it's best to return a component (e.g. gr.Image), but you can't send both a component and text in the same message. It's possible with a regular Blocks app though. So maybe we can make ChatInterface return more than one message at a time?

abidlabs:
Ah right, I was confused, I thought you might be able to return `{"text": ..., "files": []}` as well if the chatbot is multimodal. I just need to have a clearer mental picture / we might need to improve documentation here. 

> I think the issue is that chatinterface only allows the bot to send one response at a time. For sending files, it's best to return a component (e.g. gr.Image), but you can't send both a component and text in the same message. It's possible with a regular Blocks app though. So maybe we can make ChatInterface return more than one message at a time?

This _has_ been requested, although I'm not sure if its common enough to warrant adding yet another acceptable return type. Anyways, for a separate PR. This PR lgtm

freddyaboulton:
Agreed re documentation revamp.

Issue #8909, opened 2024-07-26
Sharing in chatbot is broken


abidlabs:
### Describe the bug

Not sure if this is a widespread issue, but as @osanseviero reported, sharing in https://huggingface.co/spaces/gokaygokay/Gemma-2-llamacpp is broken.

> I tried https://huggingface.co/spaces/gokaygokay/Gemma-2-llamacpp
> Wrote "hello", got an answer
> Clicked share button
> Ended up in https://huggingface.co/spaces/gokaygokay/Gemma-2-llamacpp/discussions/new?description=

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

See above

### System Info

```shell
sdk_version: 4.37.2
```

### Severity

I can work around it

ps24601:
Space link: https://huggingface.co/spaces/huggingface-projects/gemma-2-2b-it
When you click on share:
![share failed](https://github.com/user-attachments/assets/7d033fcb-6e4b-46ef-9d85-9727d9ff11e7)

kghamilton89:
Hi @abidlabs, I've raised a simple PR that populates the `Description` parameter with a default value to resolve this.

abidlabs:
Also I'm seeing a double share icon:

![Image](https://github.com/user-attachments/assets/ad4e14c0-694b-478b-a017-8ecdce79e91d)

PR #9121, opened 2024-08-14
add default description to share button

kghamilton89:
## Description

Added the following default `description` to Gradio UI to resolve `400: Description is required` on Hugging Face Spaces.

```python
	const defaultDescription = "Check out this awesome project on Gradio!";

<IconButton
	Icon={Community}
	label={i18n("common.share")}
	{pending}
	on:click={async () => {
		try {
			pending = true;
			let formatted = await formatter(value);

			if (!formatted || formatted.trim() === "") {
				formatted = defaultDescription;
			}
```

Closes: #8909

### DIFF
--- a/js/atoms/src/ShareButton.svelte
+++ b/js/atoms/src/ShareButton.svelte
@@ -15,6 +15,8 @@
 	export let value: any;
 	export let i18n: I18nFormatter;
 	let pending = false;
+
+	const defaultDescription = "Check out this awesome project on Gradio!";
 </script>
 
 <IconButton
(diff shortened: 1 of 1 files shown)


freddyaboulton (commented):
I don't think this fixes the core issue @kghamilton89 - I think the problem is that `formatter` for the chatbot treats the value as a list of tuples but it's actually a list of message dicts

kghamilton89:
> I don't think this fixes the core issue @kghamilton89 - I think the problem is that `formatter` for the chatbot treats the value as a list of tuples but it's actually a list of message dicts

@freddyaboulton, makes sense. I'll rework this.

PR #9122, opened 2024-08-14
Postprocess hardening

freddyaboulton:
## Description

Ensures only files located in the current working directory or tempdir are able to be moved to the cache

## 🎯 PRs Should Target Issues

## Tests


gradio/routes.py
```diff
@@ -553,44 +553,22 @@ async def file(path_or_url: str, request: fastapi.Request):
...
+                allowed_paths=blocks.allowed_paths
+                + [app.uploaded_file_dir, utils.get_cache_folder()]
+                + _StaticFiles.all_paths,
+                file_sets=blocks.temp_file_sets,
+            )
+            if not allowed:
                 raise HTTPException(403, f"File not allowed: {path_or_url}.")
 
-            if not abs_path.exists():
-                raise HTTPException(404, f"File not found: {path_or_url}.")
```
freddyaboulton, line 592: According to the unit tests, we actually don't raise 404 if the file does not exist

gradio/blocks.py
```diff
@@ -1910,6 +1909,10 @@ async def process_api(
         state_ids_to_track, hashed_values = self.get_state_ids_to_track(block_fn, state)
         changed_state_ids = []
 
+        from gradio.context import LocalContext
+
+        LocalContext.blocks.set(self)
```
abidlabs, line 1914: Interesting, why do we set it here and not in the init of Blocks?

gradio/processing_utils.py
```diff
@@ -462,6 +457,55 @@ def _move_to_cache(d: dict):
     return client_utils.traverse(data, _move_to_cache, client_utils.is_file_obj)
 
 
+def _check_allowed(path: str | Path, check_in_upload_folder: bool):
+    blocks = LocalContext.blocks.get()
```
abidlabs, line 461: is not sufficient to use Context.root_block?

gradio/processing_utils.py
```diff
@@ -462,6 +457,55 @@ def _move_to_cache(d: dict):
...
+    if blocks is None or not blocks.is_running:
+        return
+
+    import tempfile
```
abidlabs, line 465: suggest putting this import at top of file

gradio/processing_utils.py
```diff
@@ -462,6 +457,55 @@ def _move_to_cache(d: dict):
...
+        if reason == "in_blocklist":
+            msg += f"it is located in one of the blocked_paths ({', '.join(blocks.blocked_paths)})."
+        elif check_in_upload_folder:
+            msg += "it was not uploaded by a user."
+        else:
+            msg += "it was not created by the application or it is not "
+            msg += "located in either the current working directory or your system's temp directory. "
+            msg += "To fix this error, please ensure your function returns files located in either "
+            msg += f"the current working directory ({os.getcwd()}), your system's temp directory ({tempfile.gettempdir()}) "
+            msg += f"or add {str(abs_path.parent)} to the allowed_paths parameter."
```
abidlabs, line 494: I would add something like, "if you'd like to specifically allow this file to be served, you can add it to the `allowed_paths` parameter of `launch()`"

gradio/processing_utils.py
```diff
@@ -462,6 +457,55 @@ def _move_to_cache(d: dict):
...
+    if (
+        utils.is_in_or_equal(abs_path, os.getcwd())
+        and abs_path.name.startswith(".")
+        and not any(
+            is_in_or_equal(path, allowed_path) for allowed_path in blocks.allowed_paths
+        )
+    ):
+        raise InvalidPathError(
+            "Dotfiles located in the temporary directory cannot be moved to the cache for security reasons. "
+            "Please check whether this file is valid and rename it if so."
```
abidlabs, line 505: same^

test/test_utils.py
```diff
@@ -414,6 +424,52 @@ def test_is_in_or_equal_fuzzer(path_1, path_2):
...
 
 
+@settings(derandomize=os.getenv("CI") is not None)
+@given(
+    path=create_path_string(),
+    blocked_paths=create_path_list(),
+    allowed_paths=create_path_list(),
+    file_sets=st.lists(create_path_set(), min_size=0, max_size=3),
+)
+def test_is_allowed_file_fuzzer(
```
abidlabs, line 434: Cool idea to use the fuzzer here. But if I'm understanding correctly, randomly generating paths is very, very unlikely to produce any overlaps between the parameters, right? So this is actually unlikely to test what happens if a `path` is in `blocked_paths` or `allowed_paths` or `file_sets`? Perhaps part of the path should be manually specified or the values should be chosen from a set with smaller cardinality

gradio/routes.py
```diff
@@ -547,44 +547,22 @@ async def file(path_or_url: str, request: fastapi.Request):
...
-            in_blocklist = any(
-                utils.is_in_or_equal(abs_path, blocked_path)
-                for blocked_path in blocks.blocked_paths
-            )
-
-            is_dir = abs_path.is_dir()
-
-            if is_dir or in_blocklist:
+            if abs_path.is_dir() or not abs_path.exists():
                 raise HTTPException(403, f"File not allowed: {path_or_url}.")
```
abidlabs, line 551: nit: it would be better to set an `allowed` flag here, and raise a single 403 later, after the `utils.is_allowed_file` has completed so that malicious attackers cannot using timing attacks to infer whether a particular filepath even exists (that itself can be insecure in certain settings)

abidlabs:
Note: I believe the `gr.Code` component is still vulnerable to attacks since in `postprocess()`, we simply read the file and return its contents to users. I think we just should just deprecate letting `gr.Code` taking in a tuple value

abidlabs:
Left some comments but overall this works great. Good stuff @freddyaboulton

abidlabs (approved):

gradio/processing_utils.py
freddyaboulton: I think we set Context.root_block to `None` when we exit the Blocks context and launch

gradio/blocks.py
freddyaboulton: I thought this would be more "future proof" as it's the most up to date version of the Blocks before any sort of processing happens. But I don't think we need this actually.

gradio/routes.py
freddyaboulton: Yea good call.

gradio/processing_utils.py
abidlabs: Ah I see

test/test_utils.py
freddyaboulton: I greatly reduced the cardinality of the allowed possible paths set and also added some corner cases manually in a different test.

gradio/processing_utils.py
freddyaboulton: done

gradio/processing_utils.py
freddyaboulton: Done

gradio/blocks.py
freddyaboulton: We need this to be done after init for `is_running` to be set. I think actually this is a good place for it.

Issue #7601, opened 2024-03-05
gr.Dataframe not sending sorted dataframe to callback


ross-mitchell:
### Describe the bug

The gr.Dataframe method allows the user to display and sort data tables. The user can sort the displayed data by clicking on the triangle in the column title cell. See attached screen shot. 

If the user sorts the data by a column then clicks on a data cell, then the data passed to the select call back function is incongruent.

The gr.SelectData event contains the correct row and column indices, and the correct value of the selected cell.

However, the unsorted dataframe is passed into the select callback function for the widget.

The value in the passed dataframe at the passed row and column indices does not match the passed value in the gr.SelectData event.

This means the callback can not access other data in a sorted dataframe, outside the selected cell, reliably. 

This issue is present in Gradio v4.19.2 but not in v3.36.1

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr
import pandas as pd

with gr.Blocks() as demo:

    sample_data = {
                    'A': ['A0', 'A1', 'A2', 'A3'],
                    'B': ['B0', 'B1', 'B2', 'B3'],
                    'C': ['C0', 'C1', 'C2', 'C3']
                   }
    df = pd.DataFrame(sample_data)
    print(f"Dataframe as initialized:\n{df}")

    df_widget = gr.Dataframe(df, interactive=False)

    def df_select_callback(df: pd.DataFrame, evt: gr.SelectData):
        row_index = evt.index[0]
        col_index = evt.index[1]
        value = evt.value
        print(f"\n(Row, Column) = ({row_index}, {col_index}). Value = {value}")
        print(f"Dataframe in callback:\n{df}")
        return

    df_widget.select(df_select_callback, inputs=[df_widget])

if __name__ == "__main__":
    demo.launch()

```

### Screenshot

The user has sorted column A by descending value, then clicked on the cell containing "B0" in row 3, column 1.

![CleanShot 2024-03-04 at 17 42 28@2x](https://github.com/gradio-app/gradio/assets/13511772/1dd816e9-4479-4dbc-8e61-ea0f3a7b1397)

### Logs

```shell
Dataframe as initialized:
    A   B   C
0  A0  B0  C0
1  A1  B1  C1
2  A2  B2  C2
3  A3  B3  C3
Running on local URL:  [edited out]
To create a public link, set `share=True` in `launch()`.

(Row, Column) = (3, 1). Value = B0
Dataframe in callback:
[...]

Issue #7127, opened 2024-01-23
Entire Row Data in Dataframe Select Event


derek-pyne:
- [x] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
I have a Dataframe Component with a row for every datapoint and multiple columns with aspects of the data.... one of the columns being an ID column.

I have my app instrumented such that when someone clicks a row in the Dataframe, several other components are updated to showcase data for that particular ID.

However, the evt send when a Dataframe is selected only contains the value and index of the cell that was clicked. If someone has clicked a cell in the ID column.... this works great. But if they clicked a different cell in the row I have no way of figuring out what ID that corresponds to from the evt.

The index value in the event is not useful as it gives the current index of wherever the user clicked... and does not take into account any column sorting the user may have done (ie. if you click the first row and first column.. the index will always return `[0,0]` even if you have sorted the columns such that the 100th row is now shown in the first row spot).

**Describe the solution you'd like**  
Either:
- Entire row sent with dataframe select along with everything else
- An index that references the underlying data regardless of sorting
- A way to get the current view of the Dataframe so that I can deduce what row a particular index refers to... this might exist but struggling

jonnyjohnson1:
You can kind of make this happen in gr.Dataset, but it's not as pretty

wangyiav:
I have exact same issue and feature requirement here. Any progress on this one?

skye0402:
Same here - would be great to know the row index of what the user clicked in the original data.

abidlabs:
Working on a PR that should provide a solution for this: https://github.com/gradio-app/gradio/pull/9128 if anyone would like to give feedback (installation instructions are in the PR)

PR #9128, opened 2024-08-15
Allow accessing the entire row of selected values in `gr.DataFrame`

abidlabs:
Closes: https://github.com/gradio-app/gradio/issues/7601
Closes: https://github.com/gradio-app/gradio/issues/7127

This PR adds a `.row_value` parameters to `gr.SelectData` for `gr.DataFrame`. 

Example:

```py
import gradio as gr
import pandas as pd

with gr.Blocks() as demo:

    sample_data = {
                    'A': ['A0', 'A1', 'A2', 'A3'],
                    'B': ['B0', 'B1', 'B2', 'B3'],
                    'C': ['C0', 'C1', 'C2', 'C3']
                   }
    df = pd.DataFrame(sample_data)
    df_widget = gr.Dataframe(df, interactive=False)

    def df_select_callback(df: pd.DataFrame, evt: gr.SelectData):
        print("index", evt.index)
        print("value", evt.value)
        print("row_value", evt.row_value)
        return

    df_widget.select(df_select_callback, inputs=[df_widget])

if __name__ == "__main__":
    demo.launch()
```


dwipper:
@abidlabs Based on my testing, works great! Thanks!

freddyaboulton:
@abidlabs - seeing this weird behavior when testing locally - I think the `original_indices` are being refreshed on every sort so the `original_index` is not the actual original index. 

https://www.loom.com/share/f8e799ef89c24e578979cf4e7703ccd9?sid=45e1e768-a361-4991-bad4-667684df436b


I would be more in favor of passing the entire row value as `value` in the select event. Seems like that would be more useful in the event you don't have access to the original dataframe (it was created in an event). Of course, that's breaking but we could do for 5.0?

abidlabs:
> I would be more in favor of passing the entire row value as value in the select event. Seems like that would be more useful in the event you don't have access to the original dataframe (it was created in an event). Of course, that's breaking but we could do for 5.0?

Can you explain what you mean by if you don't have access to the original dataframe? You can always get the value of the dataframe by passing it is as an input component. 

I like the idea of passing in the original indices over passing in a row in case we later introduce functionality that allows re-ordering the columns as well

freddyaboulton:
> You can always get the value of the dataframe by passing it is as an input component.

Yea that's true!

abidlabs:
> @abidlabs - seeing this weird behavior when sorting more than once - I think the original_indices are being refreshed on every sort so the original_index is not the actual original index.

I'll check to see what the issue is, thanks

pngwn:
> Can you explain what you mean by if you don't have access to the original dataframe? You can always get the value of the dataframe by passing it is as an input component.

This does generally create an enormous payload tho, far from ideal imo.

abidlabs:
In cases where a user doesn't have access to the original dataframe yes, but otherwise its just sending the tuple of indices. If we were to send the entire row, it'd be a bigger payload than just sending the tuple in all cases, *and* it might not be future proof in case we introduce mechanisms to reorder columns at some point

abidlabs:
After thinking about it some more, I think that @freddyaboulton's suggestion of sending the updated row values makes more sense than sending the original index. Reason being that the the concept of "original index" is not well-defined for interactive dataframes (e.g. if a user has modified the value of a cell, or has inserted/deleted rows/columns, what should the original index reflect?). I think the concept of a `row_value` and `column_value` should be well-defined. 

@dwipper if you have access to the sorted `row_value` and `column_value`, will that work for your use case?

dwipper:
@abidlabs It's interesting. In 4.41 it appears the index_num = evt.index[0] is returning the original index value. This would ha have worked for my use case when I created a shadow DF that had the original DF, and was looking up a unique reference id in that shadow DF. Since that didn't work, I moved the reference id into a column of the DF, and now the `record_id=df.iat[index_num, 6]` is returning the original index vs the DF record that was clicked....

abidlabs:
hmm @dwipper I can't think of any recent change to the dataframe that would have caused this.

dwipper:
@abidlabs To clarify, on the df.select event, the index_num = evt.index[0] is providing the sorted row index. But when I use the df.ait(index_num, 2) function to try get a cell value in the row of sorted df, the lookup is seems to reference the unsorted index values, and returns the cell value in the unsorted row. I'm thinking this isn't the expected/desired behavior?

Running this modification your above code in 4.41 will show you the issue:

`import gradio as gr
import pandas as pd

with gr.Blocks() as demo:
    sample_data = {
                    'A': ['A0', 'A1', 'A2', 'A3'],
                    'B': ['B0', 'B1', 'B2', 'B3'],
                    'C': ['C0', 'C1', 'C2', 'C3']
                   }
    df = pd.DataFrame(sample_data)
    df_widget = gr.Dataframe(df, interactive=False)

    def df_select_callback(df: pd.DataFrame, evt: gr.SelectData):
        print(evt.index)
        index_num = evt.index[0]
        print(df.iat[index_num, 2])
        return

    df_widget.select(df_select_callback, inputs=[df_widget])

if __name__ == "__main__":
    demo.launch()`

dwipper:
@abidlabs My bigger picture use case is this: 

I have a list of data in a database table. The dataframe shows a subset of the columns in the table. When the user clicks on a row in the dataframe, I need to go back to the table and get additional fields to display in the UI for the row the user clicked on. In order to do that, I have a hidden record_id column in the dataframe to get back to the record in the table. Due to the above behavior of the .ait() function, if the dataframe is sorted, the incorrect record_id value is returned and the wrong record and data from the table is returned and displayed in the UI.

freddyaboulton (approved):
This works well @abidlabs ! Two comments:
- Should we send the current headers in a separate key? In the event columns are reordered.
- Tables are typically taller than they are wide. I'm a bit worried sending the entire column over is a bit too much data. Perhaps that can be made opt-in?

This fixes the given issues so will approve

abidlabs:
> Should we send the current headers in a separate key? In the event columns are reordered.
> Tables are typically taller than they are wide. I'm a bit worried sending the entire column over is a bit too much data. Perhaps that can be made opt-in?

Good points @freddyaboulton, for now I'll only send `row_value` and we can think about how to handle column values if/when we introduce mechanisms for changing columns. We might need to modify `.select()` to add a flag or something else

abidlabs:
Thanks for the feedback everyone!

dwipper:
@abidlabs The new function works great! 

Based on some related testing, I figured out that the root issue here is that when the gr.Dataframe gets sorted by the user by clicking on the column header arrow, that sort isn't reflected in the gr.Dataframe instance. So when the gr.Dataframe instance is passed to a function, it has the original index order, not the sorted order that is represented in the UI. This is apparently why, as mentioned above, the df.iat() function uses the original index order, not the sorted index order.

Is this a bug or the expected behavior?

dwipper:
@abidlabs Thanks for the clarification. Any way you can think around that, i.e. getting the sorted DF? Does a gr.Dataset() function differently? 

While getting the row the user clicked on is really helpful, in my app, I want the user to be able to scroll through the list with a VCR control (see following image). It works fine on the unsorted list, but if the user sorts the list, the scrolling doesn't work properly since it's based on the original index.

image
image

PR #9150, opened 2024-08-19
DNS resolver on ip check

aliabid94:


gradio/processing_utils.py
```diff
@@ -270,7 +270,20 @@ def save_file_to_cache(file_path: str | Path, cache_dir: str) -> str:
...
 
-def check_public_url(url: str):
+def resolve_with_google_dns(hostname: str) -> str | None:
+    google_dns = "8.8.8.8"
+    try:
+        resolver_socket = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+        resolver_socket.settimeout(3)
+        resolver_socket.connect((google_dns, 53))
+        return socket.gethostbyname(hostname)
+    except OSError:
```
abidlabs, line 280:
```suggestion
    except (OSError, socket.gaierror):
```

gradio/processing_utils.py
```diff
@@ -283,23 +296,35 @@ def check_public_url(url: str):
...
     for family, _, _, _, sockaddr in addrinfo:
         ip = sockaddr[0]
         if family == socket.AF_INET6:
             ip = ip.split("%")[0]  # Remove scope ID if present
 
-        if not ipaddress.ip_address(ip).is_global:
-            raise httpx.RequestError(
-                f"Non-public IP address found: {ip} for URL: {url}"
+        if ipaddress.ip_address(ip).is_global:
+            is_public = True
```
abidlabs, line 306: Why not just return the URL right here instead of creating the `is_public` flag

gradio/processing_utils.py
abidlabs: also maybe socket timeout error

test/test_processing_utils.py
```diff
@@ -421,7 +421,7 @@ async def test_json_data_not_moved_to_cache():
 )
 def test_local_urls_fail(url):
     with pytest.raises(httpx.RequestError, match="Non-public IP address found"):
```
abidlabs, line 423: Looks like you'll need to update these tests to just check if `get_public_url` returns `None` instead since we no longer raise errors in the method

gradio/processing_utils.py
```diff
@@ -284,23 +297,35 @@ def check_public_url(url: str):
...
+        return url
+
+    google_resolved_ip = resolve_with_google_dns(hostname)
+    if google_resolved_ip and ipaddress.ip_address(google_resolved_ip).is_global:
+        new_parsed = parsed_url._replace(netloc=google_resolved_ip)
+        if parsed_url.port:
+            new_parsed = new_parsed._replace(
+                netloc=f"{google_resolved_ip}:{parsed_url.port}"
             )
+        return urlunparse(new_parsed)
```
freddyaboulton, line 320: If we've verified that the url maps to a public domain, why can't we use the original url?

freddyaboulton (approved):
LGTM @aliabid94 - I left a question but it's more for my understanding

gradio/processing_utils.py
aliabid94: https://huggingface.slack.com/archives/C06V0AAK7QB/p1724065700937259?thread_ts=1723778316.424279&cid=C06V0AAK7QB

gradio/processing_utils.py
```diff
@@ -271,7 +272,19 @@ def save_file_to_cache(file_path: str | Path, cache_dir: str) -> str:
     return full_temp_file_path
 
 
-def check_public_url(url: str):
+def resolve_with_google_dns(hostname: str) -> str | None:
+    url = f"https://dns.google/resolve?name={hostname}&type=A"
```
abidlabs, line 276: It was hard to find concrete numbers, but it looks like the rate limit for this API is 100 queries/second: https://groups.google.com/g/public-dns-discuss/c/p1o62SJElck, which I'm worried we'll hit from Spaces infra.

E.g. imagine a single popular Space that has a `gr.Examples` that includes a file from a Hugging Face dataset, now multiplied across multiple viral Spaces that share the same IP address

abidlabs (approved):

gradio/processing_utils.py
```diff
@@ -271,36 +273,71 @@ def save_file_to_cache(file_path: str | Path, cache_dir: str) -> str:
...
+        for answer in data["Answer"]:
+            if answer["type"] == 1:
+                return answer["data"]
+
+
+# Always return these URLs as is, without checking to see if they resolve
+# to an internal IP address. This is because Hugging Face uses DNS splitting,
+# which means that requests from HF Spaces to HF Datasets, Models, or Spaces
+# may resolve to internal IP addresses even if they are publicly accessible.
+PUBLIC_URL_WHITELIST = ["hf.co", "huggingface.co", "hf.space"]
```
freddyaboulton, line 293: Is `hf.space` actually expected? The hostname of a space is `<user>-<space-id>.hf.space` which contains `hf.space` but would not be in the whitelist.

freddyaboulton (approved):
LGTM @abidlabs ! Just a small question on the whitelist for spaces.

gradio/processing_utils.py
abidlabs: Good point! I'll remove that one, I actually don't even think we need to whitespace Spaces, its just models or datasets

abidlabs:
Thanks @freddyaboulton @aliabid94!

Issue #9137, opened 2024-08-17
Can't open images in new tab anymore


missionfloyd:
### Describe the bug

As of Gradio 4.41.0, images can no longer be opened in a new tab. Trying to do so, either from the context menu or by dragging, saves the image instead.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr
from PIL import Image, ImageDraw, ImageFont

def image():
    font = ImageFont.load_default(72)
    img = Image.new("RGB", (200, 100), (255, 255, 255))
    draw = ImageDraw.Draw(img)
    draw.text((100, 50), gr.__version__, anchor="mm", font=font, fill=(0, 0, 0))
    return img

demo = gr.Interface(
    fn=image,
    inputs=None,
    outputs=["image"],
)

demo.launch()

```

### Screenshot

https://github.com/user-attachments/assets/65b756d3-a389-42d0-9fd3-68540ce72a79

### System Info

```shell
Gradio Environment Information:
------------------------------
Operating System: Windows
gradio version: 4.41.0
gradio_client version: 1.3.0

------------------------------------------------
gradio dependencies in your environment:

aiofiles: 23.2.1
anyio: 4.4.0
fastapi: 0.112.1
ffmpy: 0.4.0
gradio-client==1.3.0 is not installed.
httpx: 0.27.0
huggingface-hub: 0.24.5
importlib-resources: 6.4.2
jinja2: 3.1.4
markupsafe: 2.1.5
matplotlib: 3.9.2
numpy: 2.0.1
orjson: 3.10.7
packaging: 24.1
pandas: 2.2.2
pillow: 10.4.0
pydantic: 2.8.2
pydub: 0.25.1
python-multipart: 0.0.9
pyyaml: 6.0.2
ruff: 0.6.1
semantic-version: 2.10.0
tomlkit==0.12.0 is not installed.
typer: 0.12.3
typing-extensions: 4.12.2
urllib3: 2.2.2
uvicorn: 0.30.6
authlib; extra == 'oauth' is not installed.
itsdangerous; extra == 'oauth' is not installed.


gradio_client dependencies in your environment:

fsspec: 2024.6.1
httpx: 0.27.0
huggingface-hub: 0.24.5
packaging: 24.1
typing-extensions: 4.12.2
websockets: 12.0
```

### Severity

I can work around it

PR #9151, opened 2024-08-20
Open media type files in browser

aliabid94:
We made a change the all `file=` endpoints loaded as attachments. However, it is fine to allow all media types (except SVG, which could contain JS) to load in the browser. This PR makes that change. 

Closes: https://github.com/gradio-app/gradio/issues/9137

### DIFF
--- a/test/test_routes.py
+++ b/test/test_routes.py
@@ -275,9 +275,11 @@ class TestRoutes:
 
         file_response = client.get(f"/file={image_file.name}")
         assert file_response.headers["Content-Type"] == "image/png"
+        assert "inline" in file_response.headers["Content-Disposition"]
 
         file_response = client.get(f"/file={html_file.name}")
         assert file_response.headers["Content-Type"] == "application/octet-stream"
+        assert "attachment" in file_response.headers["Content-Disposition"]
 
     def test_allowed_and_blocked_paths(self):
         with tempfile.NamedTemporaryFile(suffix=".jpg", delete=False) as tmp_file:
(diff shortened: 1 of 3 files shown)


freddyaboulton (approved):
Thanks @aliabid94 !

test/test_routes.py
```diff
@@ -254,6 +254,31 @@ def test_get_allowed_paths(self):
...
+                os.path.dirname(html_file.name),
+            ],
+        )
+        client = TestClient(app)
+
+        file_response = client.get(f"/file={image_file.name}")
+        assert file_response.headers["Content-Type"] == "image/png"
+
+        file_response = client.get(f"/file={html_file.name}")
+        assert file_response.headers["Content-Type"] == "application/octet-stream"
```
abidlabs, line 280: Can we also test the `content_disposition_type` in these tests?

abidlabs (approved):
LGTM with one comment!

test/test_routes.py
aliabid94: Done. This was giving me issues earlier cause apparently the `content_disposition_type` was being dropped since there was no explicit filename in the response. I also had to set the filename in FileResponse to be the filename of the file requested. It seems like setting the media_type to `octet-stream` was enough to make it download as an attachment previously, but good to have an explicit "attachment" disposition as well.

abidlabs: Ah interesting. Nice!

abidlabs:
Merging ahead of the release, thanks @aliabid94!

PR #9160, opened 2024-08-21
Fix native plot lite demos 

aliabd:
Just creates new demos for each where all the code is in one file. Lite playground element doesn't support multiple files yet.


demo/scatter_plot_demo/run.py
```diff
@@ -0,0 +1,78 @@
...
+        "price": [randint(10, 50) + 4 * (i % 3) for i in range(100)],
+        "wait": [random() for i in range(100)],
+    }
+)
+
+with gr.Blocks() as scatter_plots:
+    with gr.Row():
+        start = gr.DateTime("2021-01-01 00:00:00", label="Start")
+        end = gr.DateTime("2021-01-05 00:00:00", label="End")
+        apply_btn = gr.Button("Apply", scale=0)
```
hysts, line 28: Looks like code for the event triggers for `apply_btn` and `rescale_evt` was removed in this demo because it's commented out in the original code:
https://github.com/gradio-app/gradio/blob/41d5ab987ba9728753be4509490c79041655809b/demo/native_plots/scatter_plot_demo.py#L37-L44
But without this, `start`, `end` and `apply_btn` are not used and become unnecessary, so maybe we can just remove this part to avoid confusion.
I wonder why it's commented out in the first place, though. It might be better to re-add the missing event triggers.

aliabd:
Nice catch @hysts not sure why it was commented out in the first place. but I added the code back and it's working fine. it's also necessary for selecting a timespan in the graph.

Issue #8791, opened 2024-07-15
Better data capture story?


pngwn:
Argilla integration, dataset integration etc.

detail to follow.

abidlabs:
Very interesting, would love to be more involved in these discussions

freddyaboulton:
I think the solution is to write guides/tutorials on this (could have a dedicated section for this) as opposed to directly integrating anything into core gradio. The HF Datasaver abstractions have proved brittle and users always want to do something custom anyways ("I actually want each chatbot response to be a separate row in the dataset").

pngwn:
I agree that the current abstractions aren't very useful but I don't agree with just delegating to users via tutorials, will edit the OP with some context / thoughts.

davidberenstein1957:
I agree with @pngwn, from a user persona that recently wanted to implement this. It required 3 manual steps to actively and continuously capture input and output from Gradio apps while it could be really powerful to make this more integrated out of the box, like the Rest-API availability. Especially for iterating on your data and models. 

```
from transformers import pipeline

import gradio as gr

pipe = pipeline("text-classification", model="mrm8488/bert-tiny-finetuned-sms-spam-detection")
demo = gr.Interface.from_pipeline(pipe)

demo.sync_with_hub(
    repo_id="davidberenstein1957/bert-tiny-finetuned-sms-spam-detection",
    every=1
)
demo.launch()
``` 

or 

```
from transformers import pipeline

import gradio as gr

pipe = pipeline("text-classification", model="mrm8488/bert-tiny-finetuned-sms-spam-detection")
demo = gr.Interface.from_pipeline(pipe)

demo.sync_with_argilla(
    dataset="davidberenstein1957/bert-tiny-finetuned-sms-spam-detection",
    every=1
)
``` 

At Argilla we also used to have these [monitors that functions as pipeline wrappers](https://docs.argilla.io/en/latest/reference/python/python_monitoring.html#python-ref-framework-monitors).

```python
from argilla import monitor
from transformers import pipeline

pipe = pipeline("text-classification", model="mrm8488/bert-tiny-finetuned-sms-spam-detection")

monitored_pipe = monitor(model=pipeline, dataset="my_dataset", log_interval=1, sample_rate=0.3)
```

I also created 2 example spaces based on the ChatInterface
- https://huggingface.co/spaces/davidberenstein1957/llm-human-feedback-collector-chat-interface-dpo
[...]

PR #9166, opened 2024-08-21
Minor changes to flagging for 5.0

abidlabs:
This PR makes some ergonomic changes to flagging for 5.0. 

* Renames the `allow_flagging` parameter in `gr.Interface` to `flagging_mode` (deprecates old parameter)
* Flagging function is now concurrent-safe and some small perf improvements
* The default directory for flagged data is now `.gradio/flagged` instead of `flagged` and the name of the file is "datasetXX.csv" instead of "log.csv" where the XX suffix is used to create new datasets if the developer changes the components in `Interface
* The default directory for cached examples is now `.gradio/cached_examples` (these direcotry changes make it easier for developers to identify gradio-related files)

Originally, I was planning on changing the default format of the flagged data will change from CSV to JSONL, but I realized that this is not necessary as CSV files are completely compatible with the HF Datasets Hub, so for now, the changes are pretty minor. But this PR is also an opportunity to consider other breaking changes that we want to make, some of which have been discussed here: https://github.com/gradio-app/gradio/issues/8791

cc @davidberenstein1957


gradio/flagging.py
```diff
@@ -188,6 +190,153 @@ def flag(
...
+        elif self.verbose:
+            print("Using existing dataset file at:", self.dataset_filepath)
+
+    def flag(
+        self,
+        flag_data: list[Any],
+        flag_option: str | None = None,
+        username: str | None = None,
+    ) -> int:
+        if self.first_time:
```
freddyaboulton, line 292: I think we need to set `self.first_time = False` after this if block

freddyaboulton (commented):
Proposed changes look good to me @abidlabs . Automatically adding + 1 to the filename is nice but I feel like we should have a fallback in case it goes wrong. 

I feel like the last piece is a guide on synching to the hub and also implementing a custom flag method for a chatbot or something like that

abidlabs:
> Proposed changes look good to me @abidlabs . Automatically adding + 1 to the filename is nice but I feel like we should have a fallback in case it goes wrong.

Have added a try/catch 

> I feel like the last piece is a guide on synching to the hub and also implementing a custom flag method for a chatbot or something like that

I'll do this in a separate PR so that we can get this PR in for beta testing

PR #9178, opened 2024-08-22
Remove `ruff` from runtime dependencies

akx:
## Description

`ruff` should probably not be a runtime dependency for all non-wasm users of Gradio.

This was changed in https://github.com/gradio-app/gradio/pull/7744 so cc @whitphx.

### DIFF
--- a/.changeset/plain-rules-cross.md
+++ b/.changeset/plain-rules-cross.md
@@ -1,0 +1,5 @@
+---
+"gradio": patch
+---
+
+feat:Remove `ruff` from runtime dependencies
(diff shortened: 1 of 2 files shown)


abidlabs:
I believe we use `ruff` in the custom components workflow, right @freddyaboulton @pngwn?

akx:
It looks like `components/_docs_utils.py` uses `ruff format`, but it looks like it could also just not do that if `ruff` is not available.

On that note, that implies custom components will not or should not work on wasm – is that the case?

pngwn:
Yeah, we are gonna leave this in for now it is a relatively important part of the CC CLI. Maybe we could make custom components an optional thing at some point but I don't think that is a priority atm.

See the previous PR for more details: https://github.com/gradio-app/gradio/pull/7430

Issue #9180, opened 2024-08-22
Audio stop_recording/start_recording events triggered multiple times when you pass in custom js


freddyaboulton:
### Describe the bug

Run this demo

```python
import gradio as gr


custom_js = """
function customJS(num) {
    console.log("calling js");
    return num;
}
"""

def add_one(num):
    print(f"calling python {num}")
    return num + 1

with gr.Blocks() as demo:
    n_calls = gr.Number(value=0, label="Number of calls")
    audio = gr.Audio(sources="microphone", label="Audio")

    audio.stop_recording(add_one,  n_calls, n_calls, js=custom_js)

demo.launch()
```

After you stop recording you'll see multiple lines printed to the console like

```bash
calling python 0
calling python 0
calling python 0
calling python 0
calling python 0
```

And the "calling js" line was printed multiple times

<img width="593" alt="image" src="https://github.com/user-attachments/assets/76f0a6b0-5c6a-4b68-913b-2207f33da2fd">

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

Above^

### Screenshot

![record_stop_bug](https://github.com/user-attachments/assets/69b8a336-b9b7-4c81-b499-38a6279da086)

### System Info

```shell
main
```

### Severity

Blocking usage of gradio

freddyaboulton:
If you remove the js, the python function is only called once

PR #9188, opened 2024-08-23
Fix multiple trigger bug when function has js

freddyaboulton:
## Description

Closes: #9180

Changes the audio component so that multiple events are not triggered but also fixed the bigger issue that the app was not respecting `trigger_mode` if the function ran `js`

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/blue-rooms-run.md
+++ b/.changeset/blue-rooms-run.md
@@ -1,0 +1,7 @@
+---
+"@gradio/audio": patch
+"@gradio/core": patch
+"gradio": patch
+---
+
+fix:Fix multiple trigger bug when function has js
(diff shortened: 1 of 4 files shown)


aliabid94 (approved):
Thanks freddy! Confirmed original issue is resolved. Can you explain the effect of the change in Blocks.svelte? Looks like you just moved some code into a function, wondering why that would change anything

freddyaboulton:
@aliabid94 `pnpm lint` (or `ts:check` can't remember) complained that the function reached a max "complexity limit" if I just used an if statement to add the new code. So wrapped it into a function.  Perhaps that function should be broken up a bit but didn't think it was necessary for this pr. 

Thanks for the review!

Issue #8450, opened 2024-06-04
Sliders on Firefox look different than Chrome


freddyaboulton:
### Describe the bug

The slider in firefox looks different than the chrome one

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Firefox
![firefox](https://github.com/gradio-app/gradio/assets/41651716/12eea65e-01ee-4228-b05c-f4279aec776b)

### Chrome
![chrome](https://github.com/gradio-app/gradio/assets/41651716/625b72e9-5e74-4168-af91-84f3386b2486)

### System Info

```shell
Latest
```

### Severity

I can work around it

pngwn:
I don't think those images show up @freddyaboulton.

freddyaboulton:
Fixed - thanks @pngwn

abidlabs:
Fixed in `5.0-dev` branch, thanks @hannahblair!

Issue #937, opened 2022-04-06
Reset to default on gradio.inputs.Slider


apolinario:
**Is your feature request related to a problem? Please describe.**
If the users mess with the parameters they may get confused and want to reload the whole page just to get all the defaults back

**Describe the solution you'd like**
A way for the user to reset to the default value if they messed with the slider only, maybe a button on the top right of the component

abidlabs:
Maybe all of our components should have an individual clear button, but it also might lead to a cluttered UI. We can think about this.

apolinario:
Maybe a very discrete gray round arrow wouldn't hurt, if that's clear enough 
![image](https://user-images.githubusercontent.com/788417/162062891-ab1d34bb-42d3-4cea-b9ab-c9029c87e4dd.png)

abidlabs:
@apolinario circling back, do you still feel that this is an important addition to the `Slider` component?

dawoodkhan82:
Looking into this now. What about adding a tick on the slider to show the default value. 

![slider](https://github.com/gradio-app/gradio/assets/12725292/bbf2f449-d5b3-4cbb-a48f-7661c564e807)

aliabid94:
Don't think the tick would be very obvious. Also think we could just have a separate button to restore defaults in most use cases

dawoodkhan82:
@aliabid94 as in one button to restore defaults for all components in a demo?

abidlabs:
Closed via https://github.com/gradio-app/gradio/pull/9197, thanks @hannahblair!

PR #9197, opened 2024-08-27
Redesign `gr.Slider()`

hannahblair:
## Description

Closes: #8450
Closes: #937 

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/js/slider/Index.svelte
+++ b/js/slider/Index.svelte
@@ -216,31 +216,34 @@
 
 	/* firefox */
 	input[type="range"]::-moz-range-track {
-		height: var(--size-1);
+		height: var(--size-2);
 		background: var(--neutral-200);
 		border-radius: var(--radius-xl);
 	}
 
 	input[type="range"]::-moz-range-thumb {
+		appearance: none;
 		height: var(--size-4);
 		width: var(--size-4);
-		background-color: var(--color-accent);
+		background-color: white;
 		border-radius: 50%;
-		border: var(--size-0-5) solid var(--color-accent);
-		margin-top: calc(-1 * (var(--size-4) - var(--size-1)) / 2);
-		box-shadow: 0 0 0 var(--size-0-5) var(--neutral-50);
+		border: none;
+		margin-top: calc(-1 * (var(--size-4) - var(--size-2)) / 2);
+		box-shadow:
+			0 0 0 1px rgba(247, 246, 246, 0.739),
+			1px 1px 4px rgba(0, 0, 0, 0.1);
 	}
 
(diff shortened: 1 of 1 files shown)


abidlabs:
Thanks @hannahblair. I agree that this slider feels a bit too bulky (the gray part that you slide the slider button across takes up too much vertical space). Also I'm not sure if its a good idea to put the number on the right slide as it reduces the horizontal space that the slider can slide across. I actually prefer the general layout of our current slider on those two points. 

My main UI/UX concerns with our current slider are:

* The slider color is blue instead of the "primary" color
* There's no way to reset to the original value of the design (https://github.com/gradio-app/gradio/issues/937)
* (Most annoyingly to me), the number input looks ugly and adds another border. Maybe the border should only appear if its being actively edited:

image

I think the `gr.Slider` doesn't need a significant design change if we are able to accommodate these issues, wdyt @hannahblair @dawoodkhan82 @pngwn?

abidlabs (approved):
Beautiful LGTM

abidlabs:
Very small design nit: adding the minimum and maximum numbers is a nice touch, but it does end up eating into the real estate of the slider. Perhaps we reduce some of the horizontal padding between the sliders and numbers?

Screenshot 2024-08-29 at 1 00 01 PM

abidlabs:
> @abidlabs do you reckon show_minmax would a useful param?

Yes, let's not add the parameter right now, but we could consider it if we get lots of user requests. 

>  alternatively they could do something like .min, .max { display: none } in some custom css

👍

abidlabs:
Nice good call!

rsxdalv:
Something got overlooked, I will see if I can find the solution to the visuals:

image

Maybe dark mode compatibility issue?

js/slider/Index.svelte
```diff
@@ -128,106 +162,202 @@
...
+		background-color: var(--body-text-color-subdued);
 		cursor: not-allowed;
 	}
 
-	input[type="range"] {
-		-webkit-appearance: none;
-		appearance: none;
-		width: 100%;
-		accent-color: var(--slider-color);
-		height: 4px;
```
rsxdalv, line 175: height: 4px;
appears to fix it on Firefox
image

without:
image

So I am not getting a layout shift when it's not present. 

To be fair, a probably better option (perhaps alongside height) would be background: transparent or background that follows darkmode

rsxdalv: With the latest gradio 5.3.0 it seemed to be ok even on Firefox dark mode.

Issue #8787, opened 2024-07-15
Redesign tabs


hannahblair:
- [ ] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  

**Describe the solution you'd like**  

**Additional context**

abidlabs:
Closed via #9199!

PR #9199, opened 2024-08-27
Redesign `gr.Tabs()`

hannahblair:
## Description

This PR implements a new design to the tabs to feel a bit cleaner and more modern. I'm wondering if there is enough separation around the nesting. I do like the simplicity and it looks so much better without the borders, though. 

Screenshot 2024-08-28 at 17 51 54

Screenshot 2024-08-28 at 17 51 49

Closes: #8787 

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/js/tabs/shared/Tabs.svelte
+++ b/js/tabs/shared/Tabs.svelte
@@ -43,6 +43,7 @@
 	}>();
 
 	let is_overflowing = false;
+	let overflow_has_selected_tab = false;
 
 	setContext(TABS, {
 		register_tab: (tab: Tab) => {
@@ -146,6 +147,12 @@
 
 		nav_items.forEach((item) => tab_nav_el.appendChild(item));
 		overflow_items.forEach((item) => overflow_nav.appendChild(item));
+
+		overflow_has_selected_tab = tabs.some(
+			(t) =>
+				t.id === $selected_tab &&
+				overflow_nav.contains(document.querySelector(`[data-tab-id="${t.id}"]`))
+		);
 	}
 </script>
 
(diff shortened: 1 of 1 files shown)


abidlabs:
🔥

abidlabs:
The new tabs look great to me @hannahblair, love the slight animation. Just a couple of thoughts:

(1) The border underneath the tabs does not fully extend all the way to the right side, which means that some of the content underneath the tab can peek out, like this:

image

I would extend the border all the way to the right

(2) I notice that you add a small padding to the left of the content within a tab. I think that makes sense, especially in the context of nested tabs to help users figure out what content belongs to which tab:

image

Just wanted to confirm, do you also apply a small padding to the right of the content to balance the layout?

(3) I think this tab design is good as a default, but it maybe for some layouts that a border around the tab content is necessary to help users figure out what content belongs to which tab. Down the line, if we get requests, we may want to add a parameter that sets a tab border -- but this is is a much nicer, cleaner default. GG!

abidlabs:
> thanks for reviewing @abidlabs! I've extended that width, i agree it looks better like that. also yes - there is equal padding all around the container's content. do you feel there's an imbalance on the right hand side?

All looks great to me! Just was confirming.

abidlabs (approved):

missionfloyd:
When an overflow tab is selected, and the menu is closed, does it just look like no tab is selected?

abidlabs:
^good point

abidlabs:
Yeah you're right @hannahblair. In fact, our UI is more informative because as you mentioned, when open up the overflow menu you can see that the tab is selected, whereas that does not happen with GH tabs:

image

That being said, I wonder if we should color the "..." orange when one of the overflow tabs is active, in order to inform users who might not realize that an overflow tab is selected. wdyt?

image

Either way, I don't think its a blocker to getting this in, we can tweak the UI as needed while beta testing.

Issue #9141, opened 2024-08-17
Replace buttons in `gr.ChatInterface` with icons on the messages instead


abidlabs:
Remove the buttons and instead replace them with the retry/delete icons next to individual messages that appear when you over the message  for a cleaner, more standard UI.

[internal discussion](https://huggingface.slack.com/archives/C03BHNJCBC4/p1723763151696169)

Kathanaboina:
![123](https://github.com/user-attachments/assets/a2ab9887-27c5-4e56-ab63-e96dda3dc580)
how to remove that share icon

abidlabs:
Closed via https://github.com/gradio-app/gradio/pull/9201

PR #9201, opened 2024-08-27
Move buttons from chat_interface into Chatbot

freddyaboulton:
## Description

Closes: #9141 

The retry, undo, and clear buttons of the chatbot have been moved to the chatbot window.

Clicking each button will dispatch the `retry`, `undo`, and `clear` events of the chatbot respectively. It is the developer's responsibility to implement the correct behavior. The `clear` button will automatically set the value to an empty list because I don't think it makes sense to do a round trip to the server in that case.

chat_interface_buttons

## 🎯 PRs Should Target Issues

## Tests


freddyaboulton:
Should be good for review. There is an issue with e2e tests that I think is unrelated to the changes in the PR.

abidlabs:
This looks really great @freddyaboulton! A *much* cleaner UI. 

A few minor design nits:

1. If you set `show_copy_all_button=True, show_share_button=True`, the other two buttons don't mesh with the delete button. I prefer the borderless design of the delete button, so can we modify the other buttons to look the same?

```py
import random
import gradio as gr

def random_response(message, history):
    return random.choice(["Yes", "No"])

demo = gr.ChatInterface(random_response, chatbot=gr.Chatbot(show_copy_all_button=True, show_copy_button=True, show_share_button=True))

if __name__ == "__main__":
    demo.launch()
```

image

abidlabs:
That's all actually, and code changes look good.

--------

But one thing I noticed is that now that we have removed the buttons, the "Submit" button looks kind of ugly and I'm wondering if we can remove that as well. My suggestion would be to use the `gr.MultimodalTextbox` component because it actually looks a lot nicer than our `gr.Textbox` (nicer padding, built-in submit button), but add a parameter that enables/disables the ability to send a file (in which case the payload is just the text). 

Compare:

image

with:

image

Latter looks much nicer! And it would be great to unify the UI of ChatInterface for the cases when multimodal=True/False. One thing I noticed, however, is that we don't currently have a way to stop generation when `multimodal=True`, so it would be really nice if we can change the submit icon of `gr.MultimodalTextbox` (which seems like its already supported) to a stop icon and attach the stop event to that click. 

Anyways, all this can be done in a separate PR, but wanted to leave the feedback here b/c I think this'll make our ChatInterface UI *really* nice. cc @dawoodkhan82

freddyaboulton:
Thanks for the review @abidlabs ! I'll fix the button look. I'll open a separate PR for the multimodal textbox!

freddyaboulton:
@abidlabs I've made the buttons look consistent now! I tried moving the buttons lower so that they're not blocked by the status tracker but it looked weird and I don't think we should move the status tracker position for consistency. The status tracker will now cover all buttons (the share button was not covered in your screenshot) and I think that's expected - not sure clearing the chat while waiting makes sense. 

Also made a tweak where the retry/undo buttons are disabled while the response is generating.

Should be good for another review!

abidlabs:
Looks better @freddyaboulton but I'm still seeing a faint border around the copy all button:

image

And very small nit, but the buttons overlap with the messages underneath (I think we need to add a solid fill background to the buttons to prevent this):

image

freddyaboulton:
Good catch @abidlabs ! Just addressed that:

image

Screenshot 2024-08-28 at 3 02 38 PM

abidlabs:
Sorry @freddyaboulton but now I'm noticing that if you *only* have the delete icon, its centering is off:

image

everything else looks good though

abidlabs (approved):
LGTM @freddyaboulton great PR!

dawoodkhan82:
Works great! thanks for adding this @freddyaboulton.
Small nit: I don't love the retry icon. maybe something more like this would be good. Also I think the icons can be a bit smaller.
Screenshot 2024-08-28 at 5 37 41 PM

dawoodkhan82:
Also for the `chatinterface_multimodal` demo, redo and undo don't show when I submit a multimodal message. Is this done on purpose?

freddyaboulton:
I can't repro the issue with `demo/chatinterface_multimodal/run.py` @dawoodkhan82 !

For the icons, I can definitely make them a bit smaller. How about we use the same icon as now but it's turning to the left? It's consistent with hugginchat/claude.

### Claude
image


### huggingchat
image

dawoodkhan82:
Looks way better @freddyaboulton !

dawoodkhan82:
> I can't repro the issue with demo/chatinterface_multimodal/run.py @dawoodkhan82 !

See how the undo/retry buttons don't show? When submitting an image/text.

https://github.com/user-attachments/assets/b5472d4d-7bdf-49a0-ba36-ceb7ceade0cc

abidlabs:
I can repro the issue @dawoodkhan82 pointed out. In addition, we had a few design suggestions when we were going over this PR at retro:

1. The icons are not super intuitive. In particular, the retry icon looks very similar to an undo icon, and the "X" icon doesn't feel like an undo icon. I like @dawoodkhan82's suggestion for the retry icon, and @hannahblair suggested an icon like this: https://www.flaticon.com/free-icon/undo_1239227 for undo

2. Can we confirm that the buttons are big enough to click on mobile?

3.  There is some unnecessary vertical space between messages. It looks like that might the placeholder for the icons, but it appears even when no icons are there. Can we reduce this a little?

Screenshot 2024-08-29 at 12 04 18 PM

4. Probably unrelated to this PR, but we noticed that the user message does not wrap automatically, although the bot message does. See:

image

freddyaboulton:
Thanks for the reviews @hannahblair @dawoodkhan82 @abidlabs !

1. I have changed the icons to be the ones suggested and I verified that they're clickable on mobile

https://github.com/user-attachments/assets/41b097fb-d5cf-4c26-aaf2-183de098730f

2. The vertical space between the messages is because we currently apply a margin of `calc(var(--spacing-xl) * 3)` to the top. I changed it to `calc(var(--spacing-xl) * 2)` so that there is less space.  

### Before
margin-3

### After

margin-2


3. I applied the wrap to the user message as well as you can see above.


I can't repro the issue about `chatinterface_multimodal` not working though. Are we running the same one?

chat_interface_buttons_repro

abidlabs:
Very strange, I can't repro the issue anymore either! @dawoodkhan82 are you still seeing it on the latest version of the branch? Icons look great @freddyaboulton! But a few small things I noticed while testing:

1. If you undo and then retry, then it uses the wrong input value for the retry. See here:

https://github.com/user-attachments/assets/365d0bd8-925b-4a21-a9e1-19b1be92e4f8

2. Nit: the icons underneath the bot message are no longer aligned to the left of the bot message. They seem to have some padding. 

image

whereas it should be this:

image

3. Nit: in light mode, if you upload an image and write text at the same time, in the chat bot message, they both have a weird gray border around them:

image


4. Nit: In light mode, I notice that the trash can icon in the top right has a border or box shadow:

image

It looks strange to me, I think we should just remove it so that the background is a plain white

freddyaboulton:
@abidlabs thanks for the review! I fixed the styling issues you've noticed. The behavior about undo + retry is currently present in chatinterface. It has to do with the usage of the `saved_input` `gr.State` variable. So I think we should fix in a separate PR as refactoring the backend might be more involved.

abidlabs (approved):
Beautiful work @freddyaboulton! Thanks for iterating on the style issues

abidlabs:
I'm going to merge this into `5.0-dev` as @freddyaboulton is OOO and so that we can resolve any conflicts with any other PRs

Issue #9222, opened 2024-08-30
I don't want to convert uploaded image to image_mode forcedly when using Gradio.Image


GeeMoose:
Gradio.Image component has image_mode parameters to force uploaded image to convert RGB\RGBA and etc, if I want to keep raw image mode, how should I do? if image_mode can add a option, I think it is useful.

abidlabs:
Thanks @GeeMoose, this makes sense to me. We could allow `image_mode` to take a `None` value in which case its inferred from the file itself. Would you like to open a PR for this?

GeeMoose:
if gradio organization allows, I am glad to do it.

abidlabs:
Yes, please go for it! Happy to review once the PR is open

abidlabs:
The main file to edit would be: https://github.com/gradio-app/gradio/blob/41d5ab987ba9728753be4509490c79041655809b/gradio/components/image.py

GeeMoose:
I found that the change did not pass the rule check.

PR #9225, opened 2024-08-30
 Add a 'None' option to the gradio.Image component to disable image_m…

GeeMoose:
Add a 'None' option to the gradio.Image component to disable image_mode conversion. This commit will avoid convert behavior in prepress phase forcefully.

Closes: https://github.com/gradio-app/gradio/issues/9222

### DIFF
--- a/gradio/components/image.py
+++ b/gradio/components/image.py
@@ -52,7 +52,7 @@ class Image(StreamingInput, Component):
         width: int | str | None = None,
         image_mode: Literal[
             "1", "L", "P", "RGB", "RGBA", "CMYK", "YCbCr", "LAB", "HSV", "I", "F"
-        ] | None = None,
+        ] | None = "RGB",
         sources: list[Literal["upload", "webcam", "clipboard"]]
         | Literal["upload", "webcam", "clipboard"]
         | None = None,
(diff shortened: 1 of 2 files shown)


gradio/components/image.py
```diff
@@ -51,7 +51,7 @@ def __init__(
         height: int | str | None = None,
         width: int | str | None = None,
         image_mode: Literal[
-            "1", "L", "P", "RGB", "RGBA", "CMYK", "YCbCr", "LAB", "HSV", "I", "F"
+            "1", "L", "P", "RGB", "RGBA", "CMYK", "YCbCr", "LAB", "HSV", "I", "F", "None"
```
abidlabs, line 54: This should be `Literal["1", "L", "P", "RGB", "RGBA", "CMYK", "YCbCr", "LAB", "HSV", "I", "F"] | None`, instead of the string "None"

gradio/components/image.py
```diff
@@ -199,7 +199,8 @@ def preprocess(
         if suffix.lower() != "gif" and im is not None:
             with warnings.catch_warnings():
                 warnings.simplefilter("ignore")
-                im = im.convert(self.image_mode)
+                if self.image_mode != "None":
```
abidlabs, line 202:
```suggestion
                if self.image_mode is not None:
```

abidlabs:
Thanks @GeeMoose, this looks good with a minor tweak suggested above^

GeeMoose:
I I made some changes and it seems to be able to meet needs. @abidlabs

abidlabs (approved):
Great thanks @GeeMoose. I made a tweak -- setting the default back to "RGB" to avoid a breaking change -- and modifying the docstring. LGTM

GeeMoose:
OK. Although I think it should be None by default. I have seen the default value in the document. This may be a major change for Gradio. Thanks @abidlabs

PR #9229, opened 2024-08-31
Update analytics.py

311-code:
Security and privacy concerns. IP address should not be sent.

## Description

There are significant security and privacy concerns in this file. A user's IP address is sent and this should not be the case. Telemetry should be more general than this. In addition, this could easily be modified to send str for a clip text encode in kwargs to my own ip address instead if I were to redistribute my own custom Gradio. The IP address code should not be there in the first place.

Closes: #(issue)

## 🎯 PRs Should Target Issues

I don't see anything specified regarding sending str from AI image and video users or hijacking this file.

## Tests

I have successfully modified this locally to sent info to a test server containing model state_dict layers information and str strings, k in keys, and extrapolatied information (on myself only) what I had been generating, and forwarded the IP info. In my opinion the IP address code should not be here and be completely removed.

### DIFF
--- a/gradio/analytics.py
+++ b/gradio/analytics.py
@@ -67,7 +67,7 @@ def _do_analytics_request(topic: str, data: dict[str, Any]) -> None:
 
 
 def _do_normal_analytics_request(topic: str, data: dict[str, Any]) -> None:
-    data["ip_address"] = get_local_ip_address()
+    # data["ip_address"] = get_local_ip_address()
     try:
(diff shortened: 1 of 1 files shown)


abidlabs:
Hi @311-code, thank you for this PR. We have already made this change in the `5.0-dev` branch that we will be merging into `main` for the 5.0 release in the coming weeks. Please see here: https://github.com/gradio-app/gradio/pull/8884

PR #9233, opened 2024-09-01
Adds LLM to the Playground

aliabd:
There's a different cloudflare page that's interacting with the inference api and saving the queries/responses. 

Use this link to try it out:  [aliabd-playground-llm.gradio-website.pages.dev/playground](https://aliabd-playground-llm.gradio-website.pages.dev/playground)


abidlabs:
@aliabd I think the S3 bucket's cors permissions need to be updated, the playground isn't working for me on the test link above, and I'm seeing a bunch of errors in the console:

image

aliabd:
@abidlabs can you try now? the CORS error should be gone and the playground should load as expected. That other gradio.js is unrelated I'll fix that separately

abidlabs:
Very nice work @aliabd. From a UI/UX perspective, works super well. As I've tested it, I've mainly noticed two frustrating things:

(1) Use of incorrect Gradio syntax. I know this is a general issue but if I could suggest something, I *really* think we should ensure that **streaming** inputs and outputs works well, since this is something we're changing up quite a bit in 5.0. Currently, it does quite poorly for streaming, adding a non-existent `streaming` parameter to `Interface`:

image


(2) Use of 3rd party libraries. I know that you say not to import 3rd party libraries in the prompt (twice), but it regularly imports them in the output code, e.g.
 
image

Perhaps this is because we import 3rd party libraries in some of the demos?  

That being said, I think we should go ahead and merge this in since we'll continue playing around with LLM/tinkering with the prompt afterwards. The UI/UX is in great shape!

abidlabs (approved):

Issue #8102, opened 2024-04-23
[ImageEditor] - Brush is limited to 4 colors


maggieNeedsZzz:
### Describe the bug

When providing a brush to ImageEditor it doesn't display all brush colors only the first 4. Tried using both hex and rgb colors but the result was the same. Setting the brush' color_mode to 'fixed' or 'defaults' also doesn't make any difference.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

with gr.Blocks() as demo:
    with gr.Column():
        with gr.Tab("Semantic Synthesis") as semTab:
            myhexBrush = gr.Brush(
                colors=['#ff0000','#00ff00','#0000ff','#00ffff',
                        '#ffff00','#3333ff','#803380','#808080','#ffffff','#ff8000'],
                color_mode='fixed'
            )
    
            myrgbBrush = gr.Brush(
                colors=["rgb(255, 50, 255)","rgb(255,0,0)","rgb(204, 50, 50)","rgb(255,255,0)", "rgb(0,0,0)"],
                color_mode='fixed'
            )
            segmentation = gr.ImageEditor(
                sources=[],
                transforms=[],
                width=512,
                height=512,
                image_mode='RGB',
                brush=myhexBrush
            )
demo.launch()

```

### Screenshot

![bug](https://github.com/gradio-app/gradio/assets/72552221/3bfafe1e-977a-40a2-b7cb-2a31d97cf8f7)

### Logs

```shell
None
```

### System Info

```shell
Using clean conda environment created with:

conda create --name gradioenv python=3.8
conda activate gradioenv
pip install gradio


Versions:
gradio=4.27.0 
python=3.8.19
```

### Severity

Blocking usage of gradio

PR #9234, opened 2024-09-01
Fix limited brush colors in image editor

maggieNeedsZzz:
## Description

Related to: #8102

Removes the hardcoded limit of 4 colors in the brush color swatch in the image editor. It now supports displaying all available colors which automatically wrap to new rows by using flex-wrap: wrap.

new_brush


pngwn:
I'm not sure about this. I understand the desire but don't know if this is the right UX. at the very least the CSS needs some tweaking to clean up the padding. 

What do you think @hannahblair?

maggieNeedsZzz:
Sorry, UX is not something I usually do. I can match the bottom margin to the top one. Is there anything else I should do?

maggieNeedsZzz:
I know this is a small change, so I'll provide some context. 

I was trying to use the image editor for drawing a semantic label image.
I have a model trained for semantic synthesis, a task which can be described as generating photorealistic images from semantic layouts. Typically, these layouts are user-drawn, similar to what can be seen in [NVIDIA Canvas](https://www.nvidia.com/en-us/studio/canvas/).  

Popular datasets used to train models for this task like COCO-Stuff, Cityscapes, and ADE20K, make use of more than 4 labels in the semantic image, meaning more than 4 colors.

Therefore, the current limit on the number of brush colors makes using this component for this particular task very difficult.

I've looked for potential workarounds but couldn't find any. Apologies if I've missed something.

pngwn:
Hey! Totally understand the use case and sorry it's taken us so long to get back to you! We've been a little busy planning the upcoming release. I think @hannahblair is gonna take a closer look at this as soon as possible!

Thanks again for submitting the PR and sorry for the delay!

abidlabs:
Going to close this, while we work on a separate PR to fix the underlying issue

Issue #9241, opened 2024-09-02
create_or_modify_pyi Encoding Selection Incorrect in Windows


SmirkingKitsune:
### Describe the bug

This issue is likely related or a duplicate of #8316, but that issue seems to have been closed due to original poster not providing enough information to diagnose or troubleshoot issue. 

Sometimes the `read_text()` call in `create_or_modify_pyi` will not use expected encoding when reading the `source_file`. 

In Unix environments (and presumably in MacOS too though I have not tested), this will use the 'utf-8' encoding, which is what the `source_file` are encoded in. 

In Windows, it seems to default to a different encoding when 'utf-8' should be selected. I think it was trying to use the 'CP-1252' encoding, a legacy encoding format in Windows. This issue seems to be an edge case with Windows Python encoding interpretation. This seems to happen when encoding is not explicit and the interpreter misinterprets the correct encoding or uses the system's default. When this error does happen, it breaks the script, but it seems to only happen if encoding selection by upstream python was incorrect the first time the code was run. 

Reproducibility might be an issue, because I encountered this error three times in a row. After I implemented a fix, then reverted the fix, I no longer see the error. 

This is the code that caused the issue:
``` python
    source_code = source_file.read_text()
```
This fixed the problem, but I was unable to reproduce the problem upon reverting to the original state:
``` python
    source_code = source_file.read_text(encoding='utf-8')
```
Reverting the fix back to original state, no longer caused the issue:
``` python
    source_code = source_file.read_text()
```

I suspect when encoding is not explicitly declared, the `read_text` function might be unstable in Windows. So it should be safer to declare the encoding when calling the `read_text` function, like how I showed above. 

### Have you searched existing issues?  🔎

- [ ] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr
from gradio.components import Component

class CustomChatbot(gr.Chatbot):
    def __init__(self, *args, **kwargs):
        self.branchable = False

def  create_chat_interface():
    with gr.Blocks() as demo:
        with gr.Tabs():
            with gr.TabItem("Chatbot"):
                with gr.Row():
                    chatbot = CustomChatbot()
    return demo


if __name__ == "__main__":
    demo = create_chat_interface()
    demo.launch(debug=True)
```

### Logs

```shell
Traceback (most recent call last):
  File "C:\Users\SK\AI Share\SCAR\app.py", line 6, in <module>
    from ui import create_chat_interface
[...]

Issue #8316, opened 2024-05-17
Gradio has a conflict with Python file cp1252


MichelangeloAmorusoManzari:
### Describe the bug

When running gradio file.py, the code compiles. When opening the provided link for the gradio interface, the code crashes, giving the following error:

```
  File "C:\Users\User\AppData\Local\Programs\Python\Python38\lib\encodings\cp1252.py", line 23, in decode
    return codecs.charmap_decode(input,self.errors,decoding_table)[0]
UnicodeDecodeError: 'charmap' codec can't decode byte 0x8f in position 309911: character maps to <undefine>
```

After hours of trial and error, switching Python versions, using a new virtual environment, changing gradio version and more, I found a temporary solution: comment the code of file: cp1252.

This solves the issue for now.

Regarding the error I found that "0x8f" is part of the charmap of Python, but it's an undefined character, I cannot find more information on what is triggering this error.

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

I cannot share the code as it is under NDA unfortunately, I know you will hate me, but I'm just showing an issue.

### System Info

```shell
Gradio 4.31.3

Python 3.8 or 3.11 gave the same error.
```

### Severity

I can work around it

abidlabs:
Can you please provide a minimal repro with code @MichelangeloAmorusoManzari?

MichelangeloAmorusoManzari:
Hello, I cannot for an NDA agreement.

The versions of gradio and Python on Windows 11 made me reproduce this problem.

abidlabs:
Hi @MichelangeloAmorusoManzari you don't need to share proprietary code but we'll need some sort of minimal example that produces this error, otherwise we are not sure if it's a gradio issue or something else in your code. I'll close this for now, until we have a suitable repro.

SmirkingKitsune:
I have encountered this bug, I will post the full traceback from my console:
```python
Traceback (most recent call last):
  File "C:\Users\SK\AI Share\SCAR\app.py", line 6, in <module>
    from ui import create_chat_interface
  File "C:\Users\SK\AI Share\SCAR\ui.py", line 47, in <module>
    class CustomChatbot(gr.Chatbot):
  File "C:\Users\SK\AI Share\SCAR\venv\lib\site-packages\gradio\component_meta.py", line 202, in __new__
    create_or_modify_pyi(component_class, name, events)
[...]

PR #9242, opened 2024-09-02
Fix for create_or_modify_pyi read_text Windows Issue in component_meta.py

SmirkingKitsune:
Sometimes the `read_text()` call in `create_or_modify_pyi` will not use expected encoding when reading the `source_file`. 

## Description

This will declare the encoding that `read_text` is supposed to use. It is reading files that are encoded in 'utf-8'. Windows seems to sometimes default to 'CP-1252', but not always. Issue was not present when tested in Unix. I think it is safer to explicitly declare the encoding to be 'utf-8' since the gradio source files that are being read are encoded in 'utf-8'.

Closes: #8316 and Closes: #9241

### DIFF
--- a/.changeset/eighty-jeans-unite.md
+++ b/.changeset/eighty-jeans-unite.md
@@ -1,0 +1,5 @@
+---
+"gradio": patch
+---
+
+fix:Fix for create_or_modify_pyi read_text Windows Issue in component_meta.py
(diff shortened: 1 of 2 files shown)


abidlabs:
Makes sense, thanks for the contribution @SmirkingKitsune!

abidlabs (approved):

Issue #9249, opened 2024-09-03
Improve consistency of Icon Buttons


hannahblair:
- [ ] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
We have some icon buttons which are floating in the top right, one or two buttons attached to the component border. It would be nice if these could all be kept consistent. 

Thanks @abidlabs for the suggestion! 

<img width="50" alt="Screenshot 2024-09-03 at 16 50 16" src="https://github.com/user-attachments/assets/b0fa54e5-f3e2-4321-94d3-47a4c7406fa6">

<img width="105" alt="Screenshot 2024-09-03 at 16 50 01" src="https://github.com/user-attachments/assets/04f52ad7-2af8-4084-a796-dadd4134a6d2">

**Describe the solution you'd like**  

**Additional context**

abidlabs:
Closed via https://github.com/gradio-app/gradio/pull/9250, thanks @hannahblair!

PR #9250, opened 2024-09-03
Improve Icon Button consistency 

hannahblair:
## Description

`IconButton`s are now pushed to the top right and look a bit cleaner when aligned and place together. This PR introduces the `IconButtonWrapper` component which takes a <slot />, and is to be used in tandem with `IconButton`. A. bunch of styles can be removed as a result. 

I've added a background on hover which uses `background-fill-secondary`, but now background of the buttons is now block-background-fill, so we can use that as the hover colour and will have a tad more contrast. 

Screenshot 2024-09-03 at 17 56 59

---

I've also used these changes to improve gr.Video's edit buttons which I think look a lot nicer now! 

base
Screenshot 2024-09-03 at 17 38 01

new 
Screenshot 2024-09-03 at 17 37 18

To achieve this, I've amended `ModifyUpload` to accept a <slot /> so that we can add more `IconButton`s to the `IconButtonWrapper` inside `ModifyUpload`.

Closes: #9249

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/calm-monkeys-argue.md
+++ b/.changeset/calm-monkeys-argue.md
@@ -11,6 +11,7 @@
 "@gradio/simpleimage": minor
 "@gradio/upload": minor
 "@gradio/video": minor
+"@gradio/wasm": minor
 "gradio": minor
 ---
 


abidlabs:
Looks very nice @hannahblair! Some small things I noticed:

(1) The light gray border doesn't always appear between icons. For example, the `gr.Video` component (when `interactive=False`) on Spaces doesn't have the gray border between the download and share icons

image


I noticed in the kitchen sink demo on Spaces, but I think its true for all components that have a share icon

(2) In the case of `gr.Video` component when `interactive=True`, I think the "undo" icon should only appear after you've made a trim (edit: actually, it should be shaded out until a trim has been made, like is the case for `gr.ImageEditor`):

image

(3) Nit: if possible, same for `gr.Model3D`'s undo icon

(4) Wdyt about increasing the height of the icons to make it the same height as the label? Now that the icon buttons are grouped together, it might be nice to give a bit of extra height to be able to click on the buttons more easily, particularly on mobile. I'm not sure if they would be too large, but its worth a quick check. It would also be nicer in the case where you have multiple components grouped together, like `image_editor_events`

image

abidlabs (approved):
Beautiful!

Issue #6219, opened 2023-11-01
Add ability to block” event listeners (e.g. file uploads, submit button)


freddyaboulton:
From the 4.0 tracking issue. Turned it into an issue in the 4.x milestone

vidthehuman:
I am uploading the  video everything fine  when i am  using local url  but when  i am  using public url  video uploading getting taking time and my code  is  running in backend why this happening so i ma using gradio 4.21.0 and when i am using the version 3.50 then getting connection error

abidlabs:
or just make the event pending until the upload etc. is ready

abidlabs:
Closed via #9253

PR #9253, opened 2024-09-03
Adds ability to wait for file uploads before triggering event

dawoodkhan82:
## Description

Adds `input_ready` param to wait on triggering events when file is still uploading.

Closes: #6219

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/js/core/src/Blocks.svelte
+++ b/js/core/src/Blocks.svelte
@@ -227,13 +227,45 @@
 		return get_data(component_id);
 	}
 
+	function waitForInputReady(dependency: Dependency): Promise<void> {
+		return new Promise((resolve) => {
+			const checkInputsReady = (): boolean => {
+				return dependency.inputs.every((inputId) => {
+					const component = components.find(
+						(comp) => comp.id === Number(inputId)
+					);
+					return component?.props["input_ready"] === true;
+				});
+			};
+
+			if (checkInputsReady()) {
+				resolve();
+				return;
+			}
+
+			const onPropChange = (e: Event): void => {
+				if (!isCustomEvent(e)) throw new Error("not a custom event");
+				const { id, prop, value } = e.detail;
+
+				if (prop === "input_ready" && value === true) {
+					if (checkInputsReady()) {
+						target.removeEventListener("prop_change", onPropChange);
+						resolve();
+					}
+				}
+			};
+
+			target.addEventListener("prop_change", onPropChange);
+		});
+	}
+
 	async function trigger_api_call(
 		dep_index: number,
 		trigger_id: number | null = null,
 		event_data: unknown = null
 	): Promise<void> {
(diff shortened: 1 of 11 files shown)


js/core/src/Blocks.svelte
```diff
@@ -227,13 +227,45 @@
...
 	}
 
+	function waitForInputReady(dependency: Dependency): Promise<void> {
+		return new Promise((resolve) => {
+			const checkInputsReady = (): boolean => {
+				return dependency.inputs.every((inputId) => {
+					const component = components.find(
+						(comp) => comp.id === Number(inputId)
+					);
+					return component?.props["input_ready"] === true;
```
dawoodkhan82, line 237: @pngwn Is there a better way to check for a components props? This doesn't seem to work as expected.

js/core/src/Blocks.svelte
```diff
@@ -227,13 +228,23 @@
...
+	}
+
 	async function trigger_api_call(
 		dep_index: number,
 		trigger_id: number | null = null,
 		event_data: unknown = null
 	): Promise<void> {
 		let dep = dependencies.find((dep) => dep.id === dep_index)!;
-
+		if (check_inputs_ready() === false) {
```
freddyaboulton, line 244: I haven't checked but I think we need to take the dep_index into account in this case. Otherwise a pending upload for event B can block event A.

js/core/src/Blocks.svelte
```diff
@@ -227,13 +228,23 @@
 		return get_data(component_id);
 	}
 
+	function check_inputs_ready(): boolean {
```
abidlabs, line 231: Nit: we don't need a function for this, can just use `inputs_waiting.length > 0` in the code directly

abidlabs (approved):
tested, lgtm!

js/core/src/Blocks.svelte
dawoodkhan82: Ah yes, fixed!

js/core/src/lang/en.json
```diff
@@ -22,7 +22,8 @@
 	"blocks": {
 		"connection_can_break": "On mobile, the connection can break if this tab is unfocused or the device sleeps, losing your position in queue.",
 		"long_requests_queue": "There is a long queue of requests pending. Duplicate this Space to skip.",
-		"lost_connection": "Lost connection due to leaving page. Rejoining queue..."
+		"lost_connection": "Lost connection due to leaving page. Rejoining queue...",
+		"waiting_for_inputs": "Waiting for file(s) to finish uploading, please retry."
```
freddyaboulton, line 26: I would say `please retry when upload is complete`

dawoodkhan82:
> Would be good to add an e2e test for the multiple events case

Will do in a separate PR, having trouble running e2e tests on 5.0-dev atm.

PR #9254, opened 2024-09-03
Adds a "huggingface" button variant, and makes it the default for `gr.LoginButton` and `gr.DuplicateButton`

abidlabs:
For more consistency with HF Hub branding, this PR adds a "huggingface" button variant, and makes it the default for `gr.LoginButton`

Here are the "primary", "secondary", and "clear" variants:

image

image

Compare with HF Hub branding: https://huggingface.co/docs/hub/en/oauth#branding

image

Closes: https://github.com/gradio-app/gradio/issues/5628

wdyt @hannahblair @Wauplin @gary149 @julien-c

### DIFF
--- a/.changeset/few-clowns-notice.md
+++ b/.changeset/few-clowns-notice.md
@@ -1,0 +1,6 @@
+---
+"@gradio/button": minor
+"gradio": minor
+---
+
+feat:Adds a "clear" button variant, and makes it the default for `gr.LoginButton`
--- a/.changeset/few-clowns-notice.md
+++ b/.changeset/few-clowns-notice.md
@@ -3,4 +3,4 @@
 "gradio": minor
 ---
 
-feat:Adds a "clear" button variant, and makes it the default for `gr.LoginButton` and `gr.DuplicateButton`
+feat:Adds a "huggingface" button variant, and makes it the default for `gr.LoginButton` and `gr.DuplicateButton`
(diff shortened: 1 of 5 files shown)


abidlabs:
Note that the original issue suggested dark buttons on a light theme and vice versa but that is jarring, especially in dark mode (and imo defeats the point of a light/dark setting) so instead, I've gone with light-on-light and dark-on-dark.

abidlabs:
Also I've opted not to change the border radius, as that would break the ability of the button to be grouped with other components, e.g. 

```py
with gr.Group():
  gr.Textbox()
  gr.LoginButton()
```

or

```py
with gr.Group():
  gr.DuplicateButton()
  gr.LoginButton()
```

Wauplin:
For reference for other reviewers, here is the current style

image

image

julien-c:
maybe we could add a `rounded=True ` option?

abidlabs:
>  maybe we could add a rounded=True  option?

@julien-c I'm a little hesitant with rounded borders because they look inconsistent with other buttons (or other components) generally in the Gradio app which all have a very small border radius, and it wouldn't play well with `gr.Group`, which groups together different components (e.g. if you want to stick a textbox next to a button, its much nicer if they have less rounded borders:

```py
import gradio as gr

with gr.Blocks() as demo:
    with gr.Group():
        with gr.Row():
          gr.Textbox(show_label=False, scale=2)
          gr.LoginButton()

demo.launch()
```

image

> Instead of "clear" I would maybe renamed it "huggingface" to emphasize it's HF branding. The advantage of doing so is been able to update the style in the future if we want (compared to "clear" which can only mean "no-background"). For example for more rounded border or to have a different font.

👍

Wauplin (approved):
Thank you!

abidlabs:
Thanks all!

PR #9260, opened 2024-09-04
Fix overflowing markdown in Chatbot

hannahblair:
## Description

Not sure what caused the issue, but the markdown tables were overflowing outside the chatbot messages. This PR allows the markdown table to break the word in order to let it wrap. 

Closes: #(issue)

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/.changeset/long-donkeys-hang.md
+++ b/.changeset/long-donkeys-hang.md
@@ -1,5 +1,4 @@
 ---
-"@gradio/icons": minor
 "@gradio/markdown": minor
 "gradio": minor
 ---


freddyaboulton:
Do you have a repro of the original issue @hannahblair ?

js/markdown/shared/MarkdownCode.svelte
```diff
@@ -188,4 +188,8 @@
 	span.md :global(.md-header-anchor > svg) {
 		color: var(--body-text-color-subdued);
 	}
+
+	span :global(table) {
```
abidlabs, line 192: My understanding is that this means that the the rule applies to <table> elements globally, not just within the Markdown component. Is that what we want?

abidlabs (approved):
Works well, thanks @hannahblair!

Here was my test code @freddyaboulton to repro the original issue and confirm the fix:

```py
import gradio as gr

md = """
| Column 1  | Column 2  | Column 3  | Column 4  | Column 5  |
|-----------|-----------|-----------|-----------|-----------|
| Row 1     | Data 1    | Data 2    | Data 3    | Data 4    |
| Row 2     | Data 1    | Data 2    | Data 3    | Data 4    |
| Row 3     | Data 1    | Data 2    | Data 3    | Data 4    |
| Row 4     | Data 1    | Data 2    | Data 3    | Data 4    |
"""


with gr.Blocks() as demo:
    gr.Markdown(md)
    gr.Chatbot(
        [("Hello", md)]
    )


demo.launch()
```

(view on a narrow screen)

PR #9261, opened 2024-09-04
Move icons into `IconButtonWrapper`

hannahblair:
## Description

Updates the chatbot icons UI. A lot of the changes are removing CSS styling and ensuring that all icons are wrapped in an IconButton and IconButtonWrapper, so any button styling and placement is consistent. 

@freddyaboulton found that any icons in the top right will hide the status tracker text in this, and other components. To fix this, I've moved the status tracker progress text to the bottom right of our components where there isn't going to be any obstruction. Let me know if you see any issues with this!

Screenshot 2024-09-09 at 16 09 55

base
image

new
Screenshot 2024-09-04 at 13 13 21

Closes: #(issue)

## 🎯 PRs Should Target Issues

## Tests


freddyaboulton (commented):
Nice the new style looks so good! And glad that it's consistent with other components now.

Two comments: the buttons now block the status tracker which I think is unintended cause it blocks the eta. 

image

The undo/retry buttons on chatinterface change color when hovered but these new buttons change background color instead. I think we should be consistent here so I guess that means changing the background color for undo/retry buttons.

icon_hover

freddyaboulton:
Thanks for the changes @hannahblair ! I was thinking that we'd place the status tracker over the buttons since I think using those buttons makes sense while the component is being updated. In principle ok with this change but would be good to get others thoughts as well.

abidlabs:
I quite like it actually! Feels cleaner to me having the status tracker in the bottom right

abidlabs (approved):

abidlabs:
I'll go ahead and merge this in @hannahblair as we're getting close to 5.0 beta release, and this might conflict with changes in https://github.com/gradio-app/gradio/pull/8966

PR #9268, opened 2024-09-04
Raise error instead of warning if checksums for binary do not match

abidlabs:
Context: [internal conversation](https://huggingface.slack.com/archives/C06V0AAK7QB/p1725461274060749)


freddyaboulton (approved):

abidlabs:
Thanks @freddyaboulton!

PR #9276, opened 2024-09-05
Fix scrollbars everywhere

aliabid94:
In the JSON component we added scrollbars to all .blocks, adding scrollbars everywhere. So if any component had a JSON component, we got scrollbars everywhere. Removed this CSS.

Screenshot 2024-09-06 at 7 28 35 AM

### DIFF
--- a/js/json/Index.svelte
+++ b/js/json/Index.svelte
@@ -70,9 +70,3 @@
 
 	<JSON {value} {open} {theme_mode} {show_indices} />
 </Block>
-
-<style>
-	:global(.block) {
-		overflow-y: scroll !important;
-	}
-</style>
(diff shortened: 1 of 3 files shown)


js/json/Index.svelte
```diff
@@ -72,7 +72,8 @@
 </Block>
 
 <style>
-	:global(.block) {
-		overflow-y: scroll !important;
+	:global(.json) {
```
aliabid94, line 75: where is this class? I couldn't find it

aliabid94:
LGTM @hannahblair, can confirm this solves the scrollbar everywhere issue. Can you explain when the JSON component will actually show the scroll bar? couldn't make sense of that from the PR.

aliabid94:
Oh I see what you're trying to do with the heights and scrolling. Looks good! Should also be possible with pure CSS and flexbox, but no need to rewrite.

abidlabs:
Cool I'll go ahead and merge this in @aliabid94 @hannahblair, since it looks like all issues are resolved, and this is a good one to get out

kexul:
When the pip package will be updated? this is quite annoying.

PR #9282, opened 2024-09-06
Further tweak to is_in_or_equal

freddyaboulton:
## Description

* Handle a corner case in is_in_or_equal
* Tweak is_running in move to cache check

## 🎯 PRs Should Target Issues

## Tests

### DIFF
--- a/gradio/utils.py
+++ b/gradio/utils.py
@@ -113,7 +113,7 @@ class BaseReloader(ABC):
         # Copy over the blocks to get new components and events but
         # not a new queue
         demo._queue = self.running_app.blocks._queue
-        demo.is_launching = False
+        demo.has_launched = True
         demo.max_file_size = self.running_app.blocks.max_file_size
         demo.is_running = True
         self.running_app.state_holder.reset(demo)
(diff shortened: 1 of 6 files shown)


abidlabs:
`Path.resolve()` can be a problem because it resolves symlinks, see: https://github.com/gradio-app/gradio/pull/3037

But I think in this case, its okay since we're only resolving the symlinks to check subpaths, we're not actually replacing filepaths with their resolved versions.

freddyaboulton:
I think `.resolve` would be ok but just refactored to not resolve symlinks at all

gradio/utils.py
```diff
@@ -113,6 +113,7 @@ def swap_blocks(self, demo: Blocks):
         # Copy over the blocks to get new components and events but
         # not a new queue
         demo._queue = self.running_app.blocks._queue
+        demo.is_launching = False
```
abidlabs, line 116: Just to confirm, `demo.is_launching()` should already be False here, right?

abidlabs (approved):
lgtm @freddyaboulton. I went through the PR and everything makes sense to me. I just had one question above. 

Also, a nit, but I find the name `is_launching` a bit unclear (it sounds like its in the process of launching()`. I think renaming it `has_launched` and inverting it would be clearer. Feel free to change if you agree

freddyaboulton:
Nice suggestion @abidlabs - just implemented and will merge once CI passes!

PR #9302, opened 2024-09-09
Fileformat whitelist

aliabid94:
Use a allowlist instead of a blocklist for XSS.


gradio/routes.py
```diff
@@ -125,17 +125,21 @@
...
+    "image/gif",
+    "image/webp",
+    "audio/mpeg",
+    "audio/wav",
+    "audio/ogg",
+    "video/mp4",
+    "video/webm",
+    "video/ogg",
+    "text/plain",
+    "application/pdf",
```
freddyaboulton, line 140: pdf files can run js so best to keep out of this list for extra safety.

Otherwise looks good to me!

PR #9306, opened 2024-09-10
Fixes race condition in `update_root_in_config`

abidlabs:
Addresses: [internal link](https://huggingface.slack.com/archives/C06V0AAK7QB/p1725632365262559)

The easiest way to test may be to change `safe_deepcopy` to a no-op and confirm that the new `test_attacker_cannot_change_root_in_config` test fails.

### DIFF
--- a/gradio/utils.py
+++ b/gradio/utils.py
@@ -465,13 +465,14 @@ def get_default_args(func: Callable) -> list[Any]:
     ]
 
 
-
 def safe_deepcopy(obj: Any) -> Any:
     try:
         return copy.deepcopy(obj)
     except Exception:
         if isinstance(obj, dict):
-            return {safe_deepcopy(key): safe_deepcopy(value) for key, value in obj.items()}
+            return {
+                safe_deepcopy(key): safe_deepcopy(value) for key, value in obj.items()
+            }
         elif isinstance(obj, list):
             return [safe_deepcopy(item) for item in obj]
         elif isinstance(obj, tuple):
--- a/gradio/utils.py
+++ b/gradio/utils.py
@@ -465,6 +465,23 @@ def get_default_args(func: Callable) -> list[Any]:
     ]
 
 
+
+def safe_deepcopy(obj: Any) -> Any:
+    try:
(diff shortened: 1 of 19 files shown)


test/test_routes.py
```diff
@@ -1548,3 +1550,58 @@ def test_bash_api_multiple_inputs_outputs():
         assert response.status_code == 200
         assert "event: complete\ndata:" in response.text
         assert json.dumps([123, "abc"]) in response.text
+
+
+def test_attacker_cannot_change_root_in_config(
```
freddyaboulton, line 1555: Great test! But it takes about a minute to run. I wonder if it can be optimized to run faster or only run on release prs

gradio/utils.py
```diff
@@ -465,6 +465,24 @@ def get_default_args(func: Callable) -> list[Any]:
     ]
 
 
+def safe_deepcopy(obj: Any) -> Any:
+    try:
+        return copy.deepcopy(obj)
+    except Exception:
+        if isinstance(obj, dict):
```
freddyaboulton, line 472: We really only care about deep-copying the `props` dict right? I think we can skip the deep-copy for all other cases

freddyaboulton (approved):
Great stuff! Left some suggestions but lgtm

test/test_routes.py
abidlabs: Really, are you sure? I was worried about this, so I ran this a few times, and it takes about 5 seconds on my machine

gradio/utils.py
abidlabs: Ah yeah we can probably optimize, let me check

abidlabs: I considered changing this, but I think its better to leave as is. There are other fields of the config that we want to copy, like `username` and `root`. Currently, they are primitives so we don't have to worry about deepcopying them, but in case they change in the future, then the safer approach is to deepcopy the whole config. Deepcopying the config only takes about 4 ms for a relatively complex demo like `kitchen_sink`, so performance should be ok

Issue #9202, opened 2024-08-27
Add a max height to `gr.Row`


elismasilva:
- [yes ] I have searched to see if a similar issue already exists.

**Is your feature request related to a problem? Please describe.**  
I am creating dynamic app where components are rendered at runtime, so i am doing something this image:
![image](https://github.com/user-attachments/assets/3ae47499-ff01-4be9-b042-f385fc16a3dc)

where i am plotting cards for each file in folder.

**Describe the solution you'd like**  
I want paginate or do scrolling down on cards because i dont want these many cards take all page area. I want to add other components below this tab session and I don't want to have to roll the whole pagin to get to these components, so it would be good that I could limit the view to a fixed size within this Tab and that the components rendered inside it are paid or rolled vertically.

**Additional context**  
this is my tab code 

``` python
with gr.TabItem("Lora's", id=1):                                        
      with gr.Row():
          with gr.Column(scale=3, elem_id="lora_tab"):
              lora_count = gr.State(len(LORA_FILES))
              use_this_loras=[]
              loras = []
              loras_scales = []                                                
              @gr.render(inputs=lora_count)
              def render_tracks(count):                                                    
                  with gr.Row():
                      for i in range(len(LORA_FILES)):
                          with gr.Column(variant="panel", min_width=300):
                              use_this_lora = gr.Checkbox(label="Apply Lora", value=False)
                              lora = gr.Textbox(label="Lora Name", value=f"{LORA_FILES[i]}", key=f"{LORA_FILES[i]}", show_label=True)                                                                
                              lora_scale = gr.Slider(
                                  interactive=False,
                                  minimum=0.1,
                                  maximum=20.0,
                                  step=0.1,
                                  value=0.1,
                                  label="Lora Scale", key=f"scale-{i}")
                              use_this_loras.append(use_this_lora)

                              def change_this_lora(use_this_lora):
                                  if use_this_lora:
                                      return gr.update(interactive=True)
                                  else:
                                      return gr.update(interactive=False)
                              use_this_lora.select(fn=change_this_lora, inputs=use_this_lora, outputs=[lora_scale])

                              loras.append(lora)
                              loras_scales.append(lora_scale)

```
I with something like this image (my drawing):
![image](https://github.com/user-attachments/assets/c76b64be-4475-47f8-a969-20a9c7db51aa)

abidlabs:
Thanks @elismasilva -- if we added a max height for `gr.Row`, it would solve the issue, right?

[...]

PR #9313, opened 2024-09-10
Standardize `height` across components and add `max_height` and `min_height` parameters where appropriate

abidlabs:
Currently, we have a `height` parameter for many of the components. However, in some components, it actually acts like the max-height parameter in CSS. See: [internal link](https://huggingface.slack.com/archives/C02SPHC1KD1/p1725429269431509?thread_ts=1725429150.525409&cid=C02SPHC1KD1)

This PR standardizes the behavior of "height" across components and add `max_height` and `min_height` parameters where appropriate

----------

**As a result of this PR, the behavior of `height`/`max_height`/`min_height` is as follows:**

(1) The `height` parameter is used in components where the `value` of a component *does not affect the height*, e.g. `gr.Image`. Therefore, the height is fixed to the value specified in `height` at all times
(2) The `max_height` and `min_height` parameters are added to those components where the value can affect the height. In those cases, the `height` parameter is left to indicate the default height, similar to how in CSS, there are separate attributes for `height`, `max-height`, and `min-height`. 

**Note**: In the case of (2), there is now a default value of `height` or `max_height` for every component. For components that look better when they have a fixed height regardless of how little content is present, like `gr.Chatbot`, the `height` has been set to a default value of `400`. For components that look better when they adjust to be smaller with less content, like `gr.FileExplorer` or `gr.DataFrame`, the `max_height` has been set to `500`. As a result, components should no longer get extremely tall if they are populated with a lot of content. The only exceptions to this is `gr.Markdown` and `gr.Row`, which I think developers would expect to grow along with the content no matter how large (but of course, they can configure this with the `height`, `max_height`, and `min_height` parameters)

Docstrings have been updated to reflect this.

-----------

Example test code:

```py
import gradio as gr
import random

with gr.Blocks() as demo:
    chatbot = gr.Chatbot(type="messages", min_height=200, max_height=300, height=None)
    msg = gr.Textbox()

        bot_message = random.choice(["How are you?", "Today is a great day", "I'm very hungry"])
        chat_history.append({"role": "user", "content": message})
        chat_history.append({"role": "assistant", "content": bot_message})
        return "", chat_history

if __name__ == "__main__":
    demo.launch()
```

Closes: https://github.com/gradio-app/gradio/issues/9202


abidlabs:
Still need to fix tests, but should otherwise be good for review

abidlabs:
Thanks so much for reviewing this @hannahblair!

Issue #9257, opened 2024-09-03
For some reason it generates a copy of the file when the image is used as the input parameter


peeter2:
### Describe the bug

For some reason it generates a copy of the file when the image is used as the input parameter. If the image component type is already a filepath then there will be two files in the temp folder. One is the original file that I uploaded and the other is a generated file that has the same file name but extension is jpeg instead of jpg, and the image file size is also different.
Why does it generate the additional file? Why can't it use the file that was already uploaded?

### Have you searched existing issues?  🔎

- [x] I have searched and found no existing issues

### Reproduction

```
def function(image):
    return gr.Button(interactive=False)
with gr.Blocks() as demo:
    image = gr.Image(sources="upload", type="filepath")
    btn = gr.Button()
    image.change(fn=function, inputs=image, outputs=[btn])

```

### System Info

```shell
Latest
```

### Severity

I can work around it

abidlabs:
Hi @peeter2 can you please provide a minimal code example that we can use to reproduce the issue above? See: https://stackoverflow.com/help/minimal-reproducible-example

abidlabs:
Hi @peeter2 the reason a new file is created is because Gradio saves all uploaded files in a temporary directory that it uses to serve files. Only files in this location are accessible to users of your application for security reasons. You can read more here: https://www.gradio.app/guides/sharing-your-app#security-and-file-access

peeter2:
the problem is that for some reason it generates two files. Let me try to be more clear:
1) The first file is uploaded to the temp directory when I drag an image to the image component.
2) The second file is generated when I have this line in my code: image.change(fn=function, inputs=image

So why does it generate the second file? Wouldn't it make sense to save resources by not creating a duplicate image file in the temp directory?

abidlabs:
Ah I see what you mean, I think yes we can reduce this redundancy

abidlabs:
Closed via https://github.com/gradio-app/gradio/pull/9314

PR #9314, opened 2024-09-10
Make `gr.Image` preprocessing more efficient

abidlabs:
This PR changes image preprocessing so it no longer makes a copy of the input image file if its not necessary to do so. In particular, if `type="filepath"` and the developer does not choose a specific image mode, then it just returns the filepath as is. Note that this file will already be in the cache since it was uploaded by a user. 

Closes: https://github.com/gradio-app/gradio/issues/9257


freddyaboulton (commented):
Is the problem that the uploaded image gets converted to `webp`? I think that was originally added for the image editor demos as saving to webp is fast and the UI could open it faster than png. 

Just want to make sure this doesn't cause an unintended change. WDYT of this pr @pngwn ?

abidlabs:
Not really, the previous logic in `.preprocess()` was that if a FileData with an `.orig_name` was passed into `preprocess()`, the original suffix for the image was respected. That was this logic:

```py
        if payload.orig_name:
            p = Path(payload.orig_name)
            name = p.stem
            suffix = p.suffix.replace(".", "")
            if suffix in ["jpg", "jpeg"]:
                suffix = "jpeg"
````

However, even then, the original image file was opened via PIL and saved. Particularly in the case of `.jpg` files, it was problematic because a new `.jpeg` file was created with the original contents. That's what's happening here:

```py
        return image_utils.format_image(
            im,
            cast(Literal["numpy", "pil", "filepath"], self.type),
            self.GRADIO_CACHE,
            name=name,
            format=suffix,
        )
```

So this fix basically avoids this saving of a new file in cases where the image file does not need to be opened. 

I think the case that you are thinking about is when the `Image` component is used as an output, in which case the format is still "webp" by default, e.g. see this test:

https://github.com/gradio-app/gradio/blob/41d5ab987ba9728753be4509490c79041655809b/test/components/test_image.py#L81

aliabid94 (approved):
LGTM! Unrelatedly, I didn't realize we accept SVGs but that forcese the format to be filepath. I would imagine this would lead to a broken demo for any image demo that was expected a numpy array or PIL object but recieved an svg filepath. We should not allow svgs unless the format is filepath.

abidlabs:
> We should not allow svgs unless the format is filepath.

Good point, I'll make that change in this PR as well

abidlabs:
Thanks folks!

Issue #9227, opened 2024-08-31
Undo + retry behavior bug in `gr.ChatInterface`


abidlabs:
@abidlabs thanks for the review! I fixed the styling issues you've noticed.  The behavior about undo + retry is currently present in chatinterface. It has to do with the usage of the `saved_input` `gr.State` variable. So I think we should fix in a separate PR as refactoring the backend might be more involved.

_Originally posted by @freddyaboulton in https://github.com/gradio-app/gradio/issues/9201#issuecomment-2322517450_

abidlabs:
Closed via https://github.com/gradio-app/gradio/pull/9316

PR #9316, opened 2024-09-10
9227 chatinterface retry bug

freddyaboulton:
## Description

Closes: #9227

In order to let developers build demos with `.undo` and `.retry` events more easily, this PR introduces `gr.UndoData` and `gr.RetryData`. They carry the index of the last user message so that developers know immediately where to trim the history. It should work across "tuples" and "messages" but the guides only cover "messages".

## 🎯 PRs Should Target Issues

## Tests


gradio/chat_interface.py
```diff
@@ -267,14 +268,12 @@ def _setup_events(self) -> None:
...
-                submit_triggers,
+            self.textbox.submit(
                 self._clear_and_save_textbox,
-                [self.textbox],
-                [self.textbox, self.saved_input],
+                [self.textbox, self.previous_input],
+                [self.textbox, self.saved_input, self.previous_input],
                 show_api=False,
                 queue=False,
-                preprocess=False,
```
freddyaboulton, line 277: Used `preprocess=False` in the past so that the data would be of `MultimodalData` instance. But for some reason, state variables do not get passed to events that have `preprocess=False`. Something to look into but I just added back preprocessing.

gradio/chat_interface.py
```diff
@@ -258,6 +258,7 @@ def __init__(
             self.chatbot_state = (
                 State(self.chatbot.value) if self.chatbot.value else State([])
             )
+            self.previous_input = State(value=[])
```
freddyaboulton, line 261: The cause of #9227 is that hitting `undo` will update `saved_input` to be the undone user message. But then that will be used as the input of `retry` which is not what we want. 

We want `saved_input` to be the latest submitted user message. In order to keep track of that, I'm adding `previous_input` variable.

gradio/chat_interface.py
```diff
@@ -328,22 +346,10 @@ def _setup_events(self) -> None:
...
-            else:
-                return data
-
         self.chatbot.undo(
-            self._delete_prev_fn,
-            [self.saved_input, self.chatbot],
-            [self.chatbot, self.saved_input],
-            show_api=False,
-            queue=False,
-        ).then(
```
freddyaboulton, line 343: Condensing the `undo` handler into one function to save a trip to the server.

gradio/events.py
```diff
@@ -306,6 +306,63 @@ def __init__(self, target: Block | None, data: Any):
...
+    def __init__(self, target: Block | None, data: Any):
+        super().__init__(target, data)
+        self.index: int | tuple[int, int] = data["index"]
+        """
+        The index of the user message that should be retried.
+        """
+
+
+@document()
+class UndoData(EventData):
```
freddyaboulton, line 340: I think it's best if these classes are separate because `gr.RetryUndoData` feels like a clunky name. But it comes with a lot of code duplication.

js/spa/test/test_chatinterface_streaming_echo.spec.ts
```diff
@@ -45,7 +45,7 @@ for (const test_case of cases) {
 
 		await page.getByLabel("retry button").click();
 		const expected_text_el_2 = page.locator(".bot p", {
-			hasText: "Run 3 - You typed: hi"
+			hasText: "Run 3 - You typed: hello"
```
freddyaboulton, line 48: Test was checking the bug lol

demo/chatbot_retry_undo/run.py
abidlabs, line 1: Very clean demo!

gradio/events.py
```diff
@@ -306,6 +306,63 @@ def __init__(self, target: Block | None, data: Any):
         """
 
 
+@document()
+class RetryData(EventData):
```
abidlabs, line 310: Consider adding a `.value` attribute to `gr.RetryData` and `gr.UndoData` carrying the value of the message, in order to be consistent with `gr.LikeData`

guides/05_chatbots/04_the-retry-and-undo-events.md
```diff
@@ -0,0 +1,115 @@
+# Retrying and Undoing Messages
```
abidlabs, line 1: Suggestion: I would rename this Guide something like: "Chatbot Events: Retrying, Undoing, and Liking/Disliking Messages"

and move the section "Liking / Disliking Chat Messages" from the `creating-a-custom-chatbot-with-blocks` guide here. It makes sense to include the `.like` event in the same guide. If you want, you could keep the `.like` section in the original guide, but then also repeat it here, because I think its nice to have them all in one place for easy reference (and you could include a link to this Guide from that one)

js/chatbot/shared/ChatBot.svelte
```diff
@@ -197,16 +201,13 @@
...
 		const last_index = messages[messages.length - 1].index;
-		let is_last;
-		if (Array.isArray(last_index)) {
-			is_last = 2 * last_index[0] + last_index[1] === total_length - 1;
-		} else {
-			is_last = last_index === total_length - 1;
-		}
+		const is_last =
+			JSON.stringify(last_index) ===
+			JSON.stringify(all_messages[all_messages.length - 1].index);
```
abidlabs, line 210: Sorry can you explain what's happening here @freddyaboulton? Why are we stringifying what should just be numerical indices?

js/chatbot/shared/ChatBot.svelte
```diff
@@ -222,11 +223,24 @@
...
 			return;
 		} else if (selected === "retry") {
-			dispatch("retry");
+			let last_index = val_.length - 1;
+			while (val_[last_index].role === "assistant") {
+				last_index--;
+			}
+			dispatch("retry", {
+				index: val_[last_index].index
+			});
```
abidlabs, line 243: And what's happening here? Why do we need a while loop, can't we just look at the index of the message that the retry/undo button is attached to?

abidlabs:
Very nice demo & docs @freddyaboulton, I do have some nits and questions, if you can help me understand the PR a bit more, I'd appreciate it!

js/chatbot/shared/ChatBot.svelte
freddyaboulton: In the tuples case the index is an array of the form (message_index, 0 if user or 1 if bot). And in javascript comparing two arrays will be false even if they have the same elements. So the code in the PR will work for either case.

Can't wait for 6.0 😅

guides/05_chatbots/04_the-retry-and-undo-events.md
freddyaboulton: Good idea - I moved the like section from the blocks guide to this new guide

js/chatbot/shared/ChatBot.svelte
freddyaboulton: The message that got clicked is the bot message. But you want to retry/undo the user message. So we have to find the last user message that got submitted and send that index to the server. That's what the while loop is doing but there could be a more idiomatic functional programming approach to do that.

js/chatbot/shared/utils.ts
```diff
@@ -64,6 +64,10 @@ export const format_chat_for_sharing = async (
 		.join("\n");
 };
 
+export interface RetryData {
```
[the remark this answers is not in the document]
freddyaboulton: It's a good idea

freddyaboulton:
@abidlabs I've addressed your comments and added some explanations to the code itself as well for future reference.

abidlabs (approved):
LGTM!

abidlabs:
I'll go ahead and merge this in @freddyaboulton as we're getting close to 5.0 beta release, and this might conflict with changes in https://github.com/gradio-app/gradio/pull/8966

PR #9321, opened 2024-09-11
Remove two dependencies: `importlib_resources` and `urllib3` (if not in Wasm)

abidlabs:
`importlib_resources` is not needed anymore since `importlib.resources.files` exists in Python 3.10 and higher: https://docs.python.org/3.10/library/importlib.html?highlight=importlib#importlib.resources.files

`urllib3` is only needed in Gradio-lite so I've updated the requirements.txt accordingly

### DIFF
--- a/.changeset/cute-rules-write.md
+++ b/.changeset/cute-rules-write.md
@@ -2,4 +2,4 @@
 "gradio": minor
 ---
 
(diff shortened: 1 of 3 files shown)


freddyaboulton (approved):
LGTM @abidlabs !

Issue #8813, opened 2024-07-17
Disable like/dislike from input message


sglbl:
- [y] I have searched to see if a similar issue already exists.
```gr.__version__: '4.38.1'```

**Is your feature request related to a problem? Please describe.**  
I don't think it's necessary to put like/dislike button in user message. Previous gradio versions didn't have this problem. 

**Describe the solution you'd like**  
Don't show like/dislike button in user message. Show only for the bot message. (or at least a parameter for this)

```python
import gradio as gr
def test(value, like_data: gr.LikeData):
    return {
        "liked_message": like_data.value,
        "input": value
    }
    
with gr.Blocks() as demo:
    input = gr.Chatbot([("question", "answer")])
    output = gr.Json()
    input.like(test, input, output)
demo.launch()
```

**Additional context**  
![image](https://github.com/user-attachments/assets/ec40269e-609e-4500-9b9c-c434e68aed85)

abidlabs:
Hi @sglbl we've seen use cases where users need to be able to upvote/downvote both the "user" messages as well as the "bot" messages, so we've included both for the sake of completion. If you really need to remove the like/dislike buttons from the user messages, I think this would be a great case for a custom component!

The tldr we've made it possible for Gradio users to create their own custom components -- meaning that you can write some Python and JavaScript (Svelte), and publish it as a Gradio component. You can use it in your own Gradio apps, or share it so that anyone can use it in their Gradio apps. 

Here's an example of a Gradio custom component: https://github.com/PhyscalX/gradio-image-prompter -- it lets you upload images and process points or draw boxes on top of the image.

If you'd like to create your own, we've put together a Guide: https://www.gradio.app/guides/five-minute-guide, and we're happy to help. The idea here would be to use the current gr.Chatbot as a template and then just remove the lines that add the like/dislike to the user message

remiconnesson:
> Hi @sglbl we've seen use cases where users need to be able to upvote/downvote both the "user" messages as well as the "bot" messages, so we've included both for the sake of completion. If you really need to remove the like/dislike buttons from the user messages, I think this would be a great case for a custom component!

Honestly, more often than not, we don't want to let the user upvote / downvote their own messages though.

> The idea here would be to use the current gr.Chatbot as a template and then just remove the lines that add the like/dislike to the user message

For anybody wanting to attempt this, then modifiying maybe this line
[...]

PR #9323, opened 2024-09-11
Disable liking user message in chatbot by default but make it configurable

freddyaboulton:
## Description

Closes: #8813

Also does some clean-up in the front-end to make the logic for setting props from dependencies a bit more general.

## 🎯 PRs Should Target Issues

## Tests


gradio/events.py
```diff
@@ -778,11 +792,31 @@ class Events:
...
+                "doc": "The time limit for the function to run. Parameter only used for the `.stream()` event.",
+                "component_prop": "false",
+            },
+        ],
     )
     like = EventListener(
         "like",
         config_data=lambda: {"likeable": False},
         callback=lambda block: setattr(block, "likeable", True),
+        event_specific_args=[
```
freddyaboulton, line 813: event_specific_args is used in two places:
1. In the `.pyi` template so that `like_user_message` is only shown in the IDE hint of the like event

Screenshot 2024-09-11 at 12 23 29 PM

2. In the `init.ts` file so that we know which props must be set from the events

js/core/src/Blocks.svelte
```diff
@@ -290,12 +290,13 @@
 
 		function trigger_prediction(dep: Dependency, payload: Payload): void {
 			if (dep.trigger_mode === "once") {
-				if (!dep.pending_request) make_prediction(payload);
+				if (!dep.pending_request)
```
freddyaboulton, line 293: Merging in main accidentally modified these lines

gradio/components/chatbot.py
```diff
@@ -199,12 +198,10 @@ def __init__(
             render_markdown: If False, will disable Markdown rendering for chatbot messages.
             bubble_full_width: If False, the chat bubble will fit to the content of the message. If True (default), the chat bubble will be the full width of the component.
             line_breaks: If True (default), will enable Github-flavored Markdown line breaks in chatbot messages. If False, single new lines will be ignored. Only applies if `render_markdown` is True.
-            likeable: Whether the chat messages display a like or dislike button. Set automatically by the .like method but has to be present in the signature for it to show up in the config.
```
freddyaboulton, line 202: This is actually not needed. We can treat this the same as the selectable prop

aliabid94:
Is there any case we'd want a user to be able to like their own message? Doesn't make sense to me tbh. There exists the ability to .select() any message for complex interactions involving user messages, but I don't get why a user would specifically .like() their own message.

freddyaboulton:
I agree it's a bit odd. Maybe in a multi-agent scenario where you have two bots speaking to each other? Decided not to get rid of it entirely since it was added due to users asking for it.

js/core/src/Blocks.svelte
whitphx: I found these were revered in
https://github.com/gradio-app/gradio/pull/9318/files#diff-b1efd869b69b090780297c2060a3e07431340720a850ce6819598716c456ff4f.
@freddyaboulton Looks like other lines were also affected in the PR above.
Like https://github.com/gradio-app/gradio/pull/9318/files#diff-b1efd869b69b090780297c2060a3e07431340720a850ce6819598716c456ff4f and 
https://github.com/gradio-app/gradio/pull/9318/files#diff-b1efd869b69b090780297c2060a3e07431340720a850ce6819598716c456ff4fR287
Can you check if it's ok, while they are not directly related to this PR.

js/core/src/init.ts
```diff
@@ -67,12 +67,14 @@ export function create_components(): {
...
-	function set_stream_every(dependencies: Dependency[]): void {
+	function set_event_specific_args(dependencies: Dependency[]): void {
 		dependencies.forEach((dep) => {
 			dep.targets.forEach((target) => {
 				const instance = instance_map[target[0]];
-				if (instance && dep.connection == "stream") {
-					instance.props.stream_every = dep.stream_every;
+				if (instance && dep.event_specific_args?.length > 0) {
+					dep.event_specific_args?.forEach((arg: string) => {
+						(instance.props as any)[arg] = (dep as any)[arg];
```
whitphx, line 76:
```suggestion
						instance.props[arg] = dep[arg as keyof Dependency];
```

js/core/src/init.ts
```diff
@@ -67,12 +67,14 @@ export function create_components(): {
...
 	let _rootNode: ComponentMeta;
 
-	function set_stream_every(dependencies: Dependency[]): void {
+	function set_event_specific_args(dependencies: Dependency[]): void {
 		dependencies.forEach((dep) => {
 			dep.targets.forEach((target) => {
 				const instance = instance_map[target[0]];
-				if (instance && dep.connection == "stream") {
-					instance.props.stream_every = dep.stream_every;
+				if (instance && dep.event_specific_args?.length > 0) {
```
whitphx, line 74:
```suggestion
				if (instance) {
```

whitphx (approved):
LGTM! Just left nit suggestions.

Also plz check https://github.com/gradio-app/gradio/pull/9323/files#r1766469891

js/chatbot/shared/ChatBot.svelte
```diff
@@ -482,13 +483,15 @@
...
-					show={likeable ||
+					show={show_like ||
 						(_retryable && is_last_bot_message(messages, value)) ||
 						(_undoable && is_last_bot_message(messages, value)) ||
 						show_copy_button}
 					handle_action={(selected) => handle_like(i, messages[0], selected)}
-					{likeable}
+					likeable={show_like}
 					_retryable={_retryable && is_last_bot_message(messages, value)}
 					_undoable={_undoable && is_last_bot_message(messages, value)}
```
whitphx, line 496: Just thought that `_retryable` and `_undoable` can also be defined as variables like `@const show_like` to remove the duplications of `_retryable && is_last_bot_message(messages, value)` and `_undoable && is_last_bot_message(messages, value)` respectively as refactoring, though it's not related to this PR.

js/core/src/Blocks.svelte
freddyaboulton: Thank you! I think it's looking good now

js/chatbot/shared/ChatBot.svelte
freddyaboulton: Yes great idea!

freddyaboulton:
Thank you for the review @whitphx ! We can still get rid of liking the user message entirely before the official 5.0 release so going to merge this in.

abidlabs:
Sorry should have commented earlier,  but instead of `like_user_message`, can we make this parameter: able to like user, bot, or both?, e.g.

```
likeable_messages: Literal["user", "bot", "both"]
```

freddyaboulton:
I think liking only user message is taking it too far hehe what's the use case?

PR #9327, opened 2024-09-12
Add `matplotlib` requirements to several demos

abidlabs:
Because `matplotlib` is no longer a dependency of `gradio` itself. This should fix all of the demos in the deployed Space PR


pngwn (approved):
Nice. Thanks for taking care of this @abidlabs!

Issue #8911, opened 2024-07-26
Disable typing in Chatinterface textbox while Chatbot is generating


abidlabs:
e.g. here: https://huggingface.co/spaces/gokaygokay/Gemma-2-llamacpp

While the Chatbot is generating, a user can submit multiple messages, which is not good from a ui perspective and could lead to a weird state. Better to disable typing while the user is generating (cc @osanseviero)

abidlabs:
Closed via https://github.com/gradio-app/gradio/pull/9328

Issue #8643, opened 2024-06-26
Placeholder for `gr.Textbox` is black instead of gray if the `gr.Textbox` is not interactive


abidlabs:
### Describe the bug

Setting the `interactive=True` seems to affect the color of the placeholder text, see:

<img width="585" alt="image" src="https://github.com/gradio-app/gradio/assets/1778297/62ade22c-c733-4d9e-a137-a08c59ffff67">

### Have you searched existing issues?  🔎

- [X] I have searched and found no existing issues

### Reproduction

```python
import gradio as gr

gr.Textbox(placeholder="Logs...", lines=2)

gr.Textbox(placeholder="Logs...", lines=2, interactive=True)
```

### System Info

```shell
gradio==4.37.1
```

### Severity

I can work around it

abidlabs:
Closed via https://github.com/gradio-app/gradio/pull/9328

PR #9328, opened 2024-09-12
Set the color of placeholder in a disabled textbox to gray instead of black, and disable typing while a response is generating in `gr.ChatInterface`, allow `gr.MultimodalTextbox` to accept string values

abidlabs:
This PR does three things (apologies for bundling together):

1. Makes the placeholder for gr.Textbox gray even if the gr.Textbox is not interactive, fixes: https://github.com/gradio-app/gradio/issues/8643

2. While the Chatbot is generating, a user could submit multiple messages. Now we disable typing while the user is generating (instead displaying a grayed-out placeholder with their original message), fixes: https://github.com/gradio-app/gradio/issues/8911 

3. Allows `gr.MultimodalTextbox` to accept a `str` value, which gets converted to a `dict` internally


gradio/chat_interface.py
```diff
@@ -430,11 +436,16 @@ async def api_fn(message, history, *args, **kwargs):
...
+    ) -> tuple[Textbox | MultimodalTextbox, str | MultimodalData]:
+        placeholder = message if isinstance(message, str) else message.get("text", "")
         if self.multimodal:
-            return {"text": "", "files": []}, MultimodalData(**cast(dict, message))
+            return MultimodalTextbox(
+                {"text": "", "files": []}, interactive=False, placeholder=placeholder
+            ), MultimodalData(**cast(dict, message))
         else:
-            return "", cast(str, message)
+            return Textbox("", interactive=False, placeholder=placeholder), cast(
```
freddyaboulton, line 446: I don't really like the placeholder because we've already cleared the textbox and the user message is displayed in the chatbot.

Minor point though.

abidlabs: Yeah okay I can just get rid of it, no strong feelings on my part

abidlabs:
Ah right, I'll address both points, thanks @freddyaboulton

gradio/components/multimodal_textbox.py
```diff
@@ -172,10 +172,12 @@ def postprocess(self, value: MultimodalValue | None) -> MultimodalData:
         """
         if value is None:
             return MultimodalData(text="", files=[])
-        if not isinstance(value, dict):
+        if not isinstance(value, (dict, str)):
             raise ValueError(
-                f"MultimodalTextbox expects a dictionary with optional keys 'text' and 'files'. Received {value.__class__.__name__}"
+                f"MultimodalTextbox expects a string or a dictionary with optional keys 'text' and 'files'. Received {value.__class__.__name__}"
             )
+        if isinstance(value, str):
```
freddyaboulton, line 179: LGTM!

abidlabs:
Ok everything should be addressed ✅

abidlabs:
Thanks again @freddyaboulton!

